PR #965, opened 2019-05-01
odemisd: increase timeout during instantiation

pieleric:


src/odemis/model/_core.py
```diff
@@ -433,17 +438,29 @@ def createInNewContainer(container_name, klass, kwargs):
     """
     container = createNewContainer(container_name, validate=False)
     try:
```
Sabrina85: Would it not be possible to just call `createInContainer` here and then do all the try/excepts in the `createInContainer` method? Now there are three nested try/excepts and it is ugly to read. Thera and I think it would be possible to skip one then.

pieleric: Actually, that's what I wanted to do... but we cannot stop the container if it was created for other component previously. So we need to handle it differently.... so that's probably the best we can do: 3 try/except :-/

PR #966, opened 2019-05-01
util linalg: duplicate private frunction from scipy

pieleric:


src/odemis/util/linalg.py
```diff
@@ -42,11 +42,25 @@
...
 import numpy
 from scipy.linalg.lapack import get_lapack_funcs
-from scipy.linalg.misc import LinAlgError, _datacopied
+from scipy.linalg.misc import LinAlgError
 
 __all__ = ['tri_inv']
 
 
+# Duplicate from scipy.linalg.misc
+def _datacopied(arr, original):
```
tepals: I somehow can't place comments on unchanged lines so I'll do it here. The module docstring mentions on line 7 the `solve_triangular` function but not this one. Shouldn't it also mention this function?

pieleric: Good catch. Just added a link to that original code too.

PR #967, opened 2019-05-01
driver semcomedi: restart acquisition thread if it failed

pieleric:


src/odemis/driver/semcomedi.py
```diff
@@ -3168,7 +3169,9 @@ def _update_raw_scan_array(self, shape, scale, translation, margin):
             center = (lim[0] + lim[1]) / 2
             width = lim[1] - lim[0]
             ratio = (shape[i] * scale[i]) / area_shape[i]
-            assert ratio <= 1 # cannot be bigger than the whole area
+            if ratio > 1:  # cannot be bigger than the whole area
+                # TODO: handle floating point error, and check for 1.0000001?
+                raise ValueError("Scan area too big: %s * %s > %s", shape, scale, area_shape)
```
Sabrina85: Do we need brackets? Quickly checked in Odemis and there you use `percentage + brackets` and no `comma`.

pieleric: You're right. I'm just really good at making this mistake....

pieleric:
Updated.

PR #969, opened 2019-05-02
SmarPod Driver

amuskens:


src/odemis/driver/smartpod.py
```diff
@@ -0,0 +1,835 @@
...
+    def moveAbs(self, pos):
+        """
+        API call to absolute move
+        """
+        if not pos:
+            return model.InstantaneousFuture()
+        
+        self._checkMoveAbs(pos)
+        if not self.IsPoseReachable(pos):
+            raise ValueError("Pose %s is not reachable by the SmartPod controller" %pos)
```
pieleric: `% pos` -> `% (pos,)`: it's a little safer.

src/odemis/driver/smartpod.py
```diff
@@ -0,0 +1,835 @@
...
+                if now - last_upd > 0.1:
+                    self._updatePosition()
+                    last_upd = time.time()
+
+                # Wait half of the time left (maximum 0.1 s)
+                left = end - time.time()
+                sleept = max(0.001, min(left / 2, 0.1))
+                future._must_stop.wait(sleept)
+
+                time.sleep(0.05)
```
pieleric: What is this extra sleep? You've just slept the right amount of time above.

src/odemis/driver/smartpod.py
```diff
@@ -0,0 +1,835 @@
...
+        f = self._createMoveFuture()
+        f = self._executor.submitf(f, self._doMoveRel, f, shift)
+        return f
+    
+    def _doMoveRel(self, future, shift):
+        """
+        Do a relative move by converting it into an absolute move
+        """
+        pos = add_coord(self.position.value, shift)
+        self._doMoveAbs(future, pos)
```
pieleric: Check the pose before calling `_doMoveAbs()`

src/odemis/driver/smartpod.py
```diff
@@ -0,0 +1,835 @@
...
+import copy
+import threading
+
+from odemis import model
+from odemis.util import driver
+from odemis.model import HwError, CancellableFuture, CancellableThreadPoolExecutor, isasync
+from operator import pos
+
+
+def add_coord(pos1, pos2):
```
pieleric: This code works well for what MoveRel() does, but it has limitations, which should be documented (before someone copies it and gets confused).
In particular, if pos1 and pos2 don't contain the same entries, it can do unexpected things. For example `{"x": 0, "y": 0}` + `{"x": 1, "z": 1}. What happens? y is dropped, and then the function fails because pos1 doesn't contain z.

At least indicate that: all axes in pos2 should be also in pos1.
The return value *only* contains the axes defined in pos2.

src/odemis/driver/smartpod.py
```diff
@@ -0,0 +1,835 @@
...
+
+    def Smarpod_SetSensorMode(self, id, mode):
+        return SmartPodDLL.SMARPOD_OK.value
+
+    def Smarpod_FindReferenceMarks(self, id):
+        self.stopping.clear()
+        time.sleep(0.5)
+        if self.stopping.is_set():
+            self.referenced = False
+            return SmartPodDLL.SMARPOD_STOPPED_ERROR.value
```
pieleric: These functions simulate the DLL, including the errcheck wrapping. So instead of returning an error, it should raise a SmartPodError.

src/odemis/driver/smartpod.py
```diff
@@ -0,0 +1,835 @@
...
+    def Smarpod_GetAcceleration(self, id, p_accel_control, p_accel):
+        accel = _deref(p_accel, c_double)
+        accel.value = self.accel.value
+        accel_control = _deref(p_accel_control, c_int)
+        accel_control.value = 1
+        return SmartPodDLL.SMARPOD_OK.value
+
+    def Smarpod_GetDLLVersion(self, p_major, p_minor, p_update):
+        major = _deref(p_major, c_uint)
+        major.value = 0
```
pieleric: Put different number for each version. eg 1.2.3 .

src/odemis/driver/smartpod.py
```diff
@@ -0,0 +1,835 @@
...
+        minor = _deref(p_minor, c_uint)
+        minor.value = 0
+        update = _deref(p_update, c_uint)
+        update.value = 0
+        return SmartPodDLL.SMARPOD_OK.value
+
+    def Smarpod_Set_ui(self, id, prop, value):
```
pieleric: Not used anymore?

src/odemis/driver/test/smartpod_test.py
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+CONFIG = {"name": "SmartPod",
+        "role": "",
+        "locator": "usb:ix:0",
+        # "locator": "fake",
+        "options":"",
+        "axes": {
+            'x': {
+                'range': [-1, 1],
```
pieleric: Could you put more realistic ranges? Based on the hardware we have, a couple of cm would sound more reasonable...

src/odemis/driver/test/smartpod_test.py
```diff
@@ -0,0 +1,168 @@
...
+    def test_out_of_range(self):
+        """
+        Test sending a position that is out of range.
+        """
+        pos = {'x': 1.5, 'y': 20, 'z': 0, 'theta_x': 0, 'theta_y': 0, 'theta_z': 0.0005}
+        with self.assertRaises(ValueError):
+            self.dev.moveAbs(pos).result()
+
+    def test_move_abs(self):
+        self.dev.SetSpeed(0.1)
```
pieleric: Why do you set the speed? Is that necessary before being able to move? If so, it should probably in the init. Either hard-code the speed for now, or get it as argument.

src/odemis/driver/test/smartpod_test.py
```diff
@@ -0,0 +1,168 @@
...
+    Tests cases for the SmartPod actuator driver
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        cls.dev = smartpod.SmartPod(**CONFIG)
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.dev.terminate()  # free up socket.
```
pieleric: Socket?! It's over USB, isn't it?

src/odemis/driver/smartpod.py
```diff
@@ -38,11 +42,17 @@
...
     Adds two coordinate dictionaries together and returns a new coordinate dictionary.
+
+    NOTE: pos1 and pos2 MUST contain the same axes values
+
     pos1: dict (axis name str) -> (float)
     pos2: dict (axis name str) -> (float)
     Returns ret
         dict (axis name str) -> (float)
     """
+    if pos1.keys() != pos2.keys():
```
pieleric: Now this function doesn't fit `doMoveRel` anymore... For instance, it's allowed to call `dev.moveRel({"x", 0.1})`. As the pos1 is the current position, the keys are not the same. => failure. I think the previous behaviour was fine. It should just be documented.

src/odemis/driver/smartpod.py
```diff
@@ -119,38 +140,40 @@ class SmartPodDLL(CDLL):
     all the functions automatically.
     """
     
+    hwModel = c_long(10001)  # specifies the SmarPod 110.45 S (nano)
```
pieleric: Let's name that constant based on what it's for -> SMARPOD_110_45_S.
You also don't need to put in the DLL class. If you put it at the top of the file, it's easier to write afterwards.

src/odemis/driver/smartpod.py
```diff
@@ -231,41 +254,48 @@ def __getitem__(self, name):
...
+            following locator syntax:
+                usb:id:<id>
+            where <id> is the first part of a USB devices serial number which
+            is printed on the MCS controller.
+            If the controller has a TCP/IP connection, use:
+                network:<ip>:<port>
         axes: dict str (axis name) -> dict (axis parameters)
             axis parameters: {
                 range: [float, float], default is -1 -> 1
+                unit: (str)
```
pieleric: Default is "m".

src/odemis/driver/test/smartpod_test.py
```diff
@@ -42,18 +42,17 @@
...
-                'range': [-1, 1],
+                'range': [-0.2, 0.2],
                 'unit': 'm',
             },
             'z': {
-                'range': [-1, 1],
+                'range': [-0.1, 0.1],
                 'unit': 'm',
             },
             'theta_x': {
```
pieleric: Do you have better range of the rotation axes too? In particular, can't they go to negative value? I'd tend to think that 0 is the default position, and it can go in both direction? The range also seems a lot less than 180°.

src/odemis/driver/smartpod.py
```diff
@@ -644,6 +671,8 @@ def moveRel(self, shift):
         if not shift:
             return model.InstantaneousFuture()
```
pieleric: For `moveRel()` and `moveAbs()`, you also want to call `shift = self._applyInversion(shift)`, in order to support the "inverted" argument. It's also needed on the `updatePosition()`, just before setting `.position`. This is a very handy argument in case the actuator vendor had a different convention than us on the "direction" of each axis (which is quite likely, as it's mostly arbitrary)

pieleric:
BTW, that device is called "SmarPod", without "t", so let's rename the files appropriately.

src/odemis/driver/smartpod.py
```diff
@@ -0,0 +1,840 @@
...
+        """
+        Move to pose command. Non-blocking
+        pos: (dict str -> float): axis name -> position
+            This is converted to the pose C-struct which is sent to the SmartPod DLL
+        Raises: SmartPodError if a problem occurs
+        """
+        # convert into a smartpad pose
+        newPose = dict_to_pose(pos)
+
+        self.core.Smarpod_Move(self._id, byref(newPose), c_uint(1000), c_int(0))
```
pieleric: Please document the 1000 (ms holding) and 0 (non-blocking).
How about using HOLDTIME_INFINITE instead of 1s? We need to find out what happens when it's not holding (does it move due to gravity?), and when it holds (does it vibrate due to closed-loop?).

src/odemis/driver/smartpod.py
```diff
@@ -0,0 +1,840 @@
...
+    """
+    Convert a Pose (C structure) to a coordinate dictionary (str) -> (double)
+        pose: a Pose (C struct)
+    returns: Coordinate dictionary (str) -> (double) of axis name to value
+    """
+    pos = {}
+    pos['x'] = pose.positionX
+    pos['y'] = pose.positionY
+    pos['z'] = pose.positionZ
+    pos['theta_x'] = pose.rotationX
```
pieleric: In Odemis, rotation is in radiants, but the API returns values in degrees => you need to convert back & forth.

src/odemis/driver/smartpod.py
```diff
@@ -0,0 +1,840 @@
...
+        self._speed_control = c_int()
+        self._accel = c_double(0)
+        self.referenced = False
+
+        # Sepcify ranges
+        self._range = {}
+        self._range['x'] = (-1, 1)
+        self._range['y'] = (-1, 1)
+        self._range['z'] = (-1, 1)
+        self._range['theta_x'] = (-3.14, 3.14)
```
pieleric: The simulator should also contain the rotation in degrees.

src/odemis/driver/smartpod.py
```diff
@@ -0,0 +1,840 @@
...
+    """
+    pose = Pose()
+    for an, v in pos.items():
+        if an == "x":
+            pose.positionX = v
+        elif an == "y":
+            pose.positionY = v
+        elif an == "z":
+            pose.positionZ = v
+        elif an == "theta_x":
```
pieleric: The convention in Odemis is to call the rotation axes "rx/ry/rz". So I'd suggest to follow the same naming convention here. Alternatively, the user would be able to override the names via the "axes" parameter, but let's not support that yet.

src/odemis/driver/smartpod.py
```diff
@@ -0,0 +1,840 @@
...
+        self._speed_control = speed_control
+
+    def Smarpod_GetSpeed(self, id, p_speed_control, p_speed):
+        speed = _deref(p_speed, c_double)
+        speed.value = self._speed.value
+        speed_control = _deref(p_speed_control, c_int)
+        speed_control.value = self._speed_control.value
+
+    def Smarpod_SetAcceleration(self, id, accel_control, accel):
+        self._accel = accel
```
pieleric: Let's store the accel_control too, to make it "feature complete".

src/odemis/driver/smarpod.py
```diff
@@ -412,8 +416,8 @@ def Move(self, pos):
         """
         Move to pose command. Non-blocking
         pos: (dict str -> float): axis name -> position
```
pieleric: I'd suggest to make that function has all the same parameters than the wrapped function `Smapod_Move()`. So the signature could be `Move(self, pos, hold_time=1, block=False)`. `hold_time` would be in s, with `float("inf")` mapped to infinite.

This has two advantages:
 * the "magical" constants are not magical and not constant anymore.
 * you don't reduce the functionality of this function.

src/odemis/driver/smarpod.py
```diff
@@ -350,29 +349,34 @@ def __init__(self, name, role, locator, axes=None, **kwargs):
...
 
-        # Check referencing, and reference if not referenced
+        axes_ref = {a: False for a, i in self.axes.items()}
+        # VA dict str(axis) -> bool
         self.referenced = model.VigilantAttribute(axes_ref, readonly=True)
         referenced = c_int()
         self.core.Smarpod_IsReferenced(self._id, byref(referenced))
         if not referenced.value:
-            logging.debug("SmartPod is not referenced. The device will not function until referencing occurs.")
+            logging.debug("SmarPod is not referenced. The device will not function until referencing occurs.")
```
pieleric: Move this log in an `else` block of the `if`. Also, make it a little higher level: "warning".

src/odemis/driver/smarpod.py
```diff
@@ -350,29 +349,34 @@ def __init__(self, name, role, locator, axes=None, **kwargs):
...
         self._speed = self.GetSpeed()
         self._accel = self.GetAcceleration()
         try:
             self._updatePosition()    # will fail if not referenced.
-        except SmartPodError as ex:
-            if ex.errno == SmartPodDLL.SMARPOD_NOT_REFERENCED_ERROR:
-                logging.info("Must reference SmartPod controller before use.")
+        except SmarPodError as ex:
+            if ex.errno == SmarPodDLL.SMARPOD_NOT_REFERENCED_ERROR:
+                logging.info("Must reference SmarPod controller before use.")
```
pieleric: That's the second log about the same thing. Just indicate that the position is unknown until the device is referenced.
Please also make sure that in this case, `.position` contains still a dict, with all the axes. I'd suggest to just put `0` as the position for each axis (as it'll probably be the position once it's referenced).

src/odemis/driver/smarpod.py
```diff
@@ -350,29 +349,34 @@ def __init__(self, name, role, locator, axes=None, **kwargs):
         # will take care of executing axis move asynchronously
         self._executor = CancellableThreadPoolExecutor(1)  # one task at a time
 
-        # Check referencing, and reference if not referenced
+        axes_ref = {a: False for a, i in self.axes.items()}
```
pieleric: Set it to `referenced.value` at init.

src/odemis/driver/smarpod.py
```diff
@@ -114,19 +119,21 @@ def dict_to_pose(pos):
...
         elif an == "z":
             pose.positionZ = v
-        elif an == "theta_x":
-            pose.rotationX = v
-        elif an == "theta_y":
-            pose.rotationY = v
-        elif an == "theta_z":
-            pose.rotationZ = v
+        elif an == "rx":
+            pose.rotationX = v * (180.0 / math.pi)
```
pieleric: `math.radians()`

src/odemis/driver/smarpod.py
```diff
@@ -101,9 +104,11 @@ def pose_to_dict(pose):
     pos['x'] = pose.positionX
     pos['y'] = pose.positionY
     pos['z'] = pose.positionZ
-    pos['theta_x'] = pose.rotationX
-    pos['theta_y'] = pose.rotationY
-    pos['theta_z'] = pose.rotationZ
+
+    # Note: internally, the system uses metres and degrees for rotation
+    pos['rx'] = pose.rotationX * (math.pi / 180.0)
```
pieleric: `math.degrees()`

src/odemis/driver/smarpod.py
amuskens: Should hold time be in seconds rather than ms? And as a float then?

src/odemis/driver/smarpod.py
amuskens: Although referenced.value is defined from this...?

src/odemis/driver/smarpod.py
pieleric: Yes, it should be a float.

src/odemis/driver/smarpod.py
pieleric: It's confusing... I don't mean `self.referenced.value`, I mean the output argument  to `SmarPod_IsReferenced()`, which is also called `referenced`.

amuskens:
Made the changes - hopefully it's ready?

src/odemis/driver/smarpod.py
```diff
@@ -0,0 +1,873 @@
...
+
+
+def add_coord(pos1, pos2):
+    """
+    Adds two coordinate dictionaries together and returns a new coordinate dictionary.
+
+    NOTE: pos1 and pos2 MUST contain the same axes values
+
+    pos1: dict (axis name str) -> (float)
+    pos2: dict (axis name str) -> (float)
```
pieleric: Please update the docstrings.

pieleric: Please, state in the docstring that all pos2 keys must be present in pos1, and that the ret keys are the same as in pos2.

src/odemis/driver/smarpod.py
```diff
@@ -0,0 +1,872 @@
...
+import time
+import math
+from ctypes import *
+import copy
+import threading
+
+from odemis import model
+from odemis.util import driver
+from odemis.model import HwError, CancellableFuture, CancellableThreadPoolExecutor, isasync
+from operator import pos
```
pieleric: Is that used?

amuskens:
I think it's ready now?

PR #970, opened 2019-05-02
Autofocus of optical path mbsem

tepals:


src/odemis/acq/align/autofocus.py
```diff
@@ -501,7 +506,7 @@ def _DoExhaustiveFocus(future, detector, emt, focus, dfbkg, good_focus, rng_focu
         # we know that upper_bound is excluded but: 1. realistically the best focus
         # position would not be there 2. the upper_bound - orig_pos range is not
         # expected to be precisely a multiple of the step anyway
-        for next_pos in numpy.arange(orig_pos, upper_bound, step):
+        for next_pos in numpy.linspace(orig_pos, upper_bound, (upper_bound - orig_pos) / step, endpoint=False):
```
pieleric: You should update the comment about this new behaviour. That said, I think it's simpler if you include the end_point (upper_bound). The reason it wasn't included was because arange would typically skip it. Now, with linspace, it's easy to include it.

src/odemis/acq/align/autofocus.py
```diff
@@ -520,8 +525,9 @@ def _DoExhaustiveFocus(future, detector, emt, focus, dfbkg, good_focus, rng_focu
 
         # if nothing was found return to original position and start going
         # downwards
-        focus.moveAbsSync({"z": orig_pos})
-        for next_pos in numpy.arange(orig_pos - step, lower_bound, -step):
+        focus.moveAbsSync({"z": orig_pos})  # TODO does this do anything?
```
pieleric: I think so... it goes back to the original position? Or is it already supposed to be there?

tepals: Yeah it goes back to the original position but does not do any measurements at that position. It moves to `next_pos` on line 552 before measuring anything, or if it does not go into the for loop it moves to `best_pos` on line 570.

src/odemis/acq/align/autofocus.py
```diff
@@ -549,7 +545,7 @@ def _DoExhaustiveFocus(future, detector, emt, focus, dfbkg, good_focus, rng_focu
         # downwards
         focus.moveAbsSync({"z": orig_pos})  # TODO does this do anything?
         num = max((orig_pos - lower_bound) / step, 0)  # Take 0 steps if orig_pos is too close to lower_bound
-        for next_pos in numpy.linspace(orig_pos - step, lower_bound, num, endpoint=False):
+        for next_pos in numpy.linspace(orig_pos - step, lower_bound, num, endpoint=True):
```
pieleric: You can just drop `endpoint=True`, as it's the default.

src/odemis/acq/align/autofocus.py
pieleric: Indeed... making unnecessary moves is not a great idea, so I'd suggest to drop that line.

src/odemis/acq/align/autofocus.py
```diff
@@ -977,6 +1004,45 @@ def AutoFocusSpectrometer(spectrograph, focuser, detectors, selector=None, strea
     return f
 
 
+def CLSpotsAutoFocus(detector, focus, good_focus=None, rng_focus=None, method=MTD_EXHAUSTIVE):
```
pieleric: This is a complete method, which only changes the exposure time of the detector, and then does exactly as `AutoFocus()`. I'd suggest that the caller just does this. If in the future, more things needs to be done, then we could keep that function, but it might be simpler for now to just write:
```
detector.exposureTime.value = 0.01
return AutoFocus(...)
```

pieleric:
BTW, the `dependencies` changes have also modified simcam, so you'll need to have a look at fixing that conflict.

tepals:
updated with your last comments and resolved the conflict, I think it can be merged.

PR #972, opened 2019-05-02
acq: improve optical path during alignment and afterwards

pieleric:


src/odemis/acq/path.py
```diff
@@ -566,30 +567,27 @@ def _doSetPath(self, path, detector):
                                 # if there is a grating stored use this one
                                 # otherwise find the non-mirror grating
                                 if axis in self._stored:
```
tepals: I think this is broken and it should change to `if (comp_role, axis) in self._stored`

src/odemis/acq/path.py
```diff
@@ -566,30 +567,27 @@ def _doSetPath(self, path, detector):
...
-                                    pos = self._stored[axis]
+                                    pos = self._stored[comp_role, axis]
                                 else:
                                     pos = self.findNonMirror(choices)
                                 if 'wavelength' in self._stored:
```
tepals: same here

pieleric:
@tepals Good catch! Just pushed a fixed version.

PR #974, opened 2019-05-06
quickcl: bugfix, filter control

philipwinklerdelmic:


plugins/quickcl.py
```diff
@@ -482,6 +469,22 @@ def _setup_sbar_cont(self):
...
+        if main_data.cl_filter and "band" in main_data.cl_filter.axes:
+            # find the "pass-through"
+            bdef = main_data.cl_filter.axes["band"]
+            for b, bn in bdef.choices.items():
+                if bn == "pass-through":
+                    main_data.cl_filter.moveAbs({"band": b})
+                    break
+            else:
+                logging.debug("Pass-through not found in the CL-filter")
+            for s_cont in self._dlg.streambar_controller.stream_controllers:
```
pieleric: Please add a comment to explain that this block adds the "filter" axis to the stream.

plugins/quickcl.py
```diff
@@ -482,6 +469,22 @@ def _setup_sbar_cont(self):
...
+                if isinstance(s_cont.stream, LiveCLStream):
+                    s_cont.add_axis_entry("band", main_data.cl_filter)
```
pieleric: I'd be nice to use the "stream_configs", as in `show_mn_axes()`. You'd need to use the CLSettingsStream. In particular, this renames the axis to "Filter" instead of "Band".

philipwinklerdelmic:
I changed the show_mn_axes function now to make it more generic.
Does this look more like what you had in mind?

pieleric:
That looks even better than what I had in mind :-)

PR #976, opened 2019-05-07
Migration python 3: updated dictionaries

philipwinklerdelmic:


src/odemis/driver/picoquant.py
```diff
@@ -726,7 +726,7 @@ def _acquire(self):
                 self._acq_wait_start()
 
                 # Open protection shutters
-                self._toggle_shutters(self._shutters.keys(), True)
+                self._toggle_shutters(list(self._shutters.keys()), True)
```
pieleric: That shouldn't be necessary for `_toggle_shutters()`.

src/odemis/driver/spectrometer.py
```diff
@@ -130,7 +130,7 @@ def __init__(self, name, role, dependencies, **kwargs):
 
         # duplicate every other VA and Event from the detector
         # that includes required VAs like .pixelSize and .exposureTime
-        for aname, value in model.getVAs(dt).items() + model.getEvents(dt).items():
+        for aname, value in list(model.getVAs(dt).items()) + list(model.getEvents(dt).items()):
```
pieleric: It might be a little bit more fancy/right to write:
```
from itertools import chain
:
for aname, value in chain(iter(model.getVAs(dt).items()), iter(model.getEvents(dt).items())):
```

src/odemis/acq/stream/_static.py
```diff
@@ -422,12 +422,12 @@ def dis_bbtl(v):
 
         # polarization VA
         # check if any polarization analyzer data, (None) == no analyzer data (pol)
-        if self._pos.keys()[0][-1]:
+        if list(self._pos.keys())[0][-1]:
```
pieleric: This one is a little annoying, as _pos might be a little big. Anyway, it should be possible to just replace by `polpositions != {None}` and in the next line `list(polpositions)[0]`.

src/odemis/acq/stream/_static.py
```diff
@@ -422,12 +422,12 @@ def dis_bbtl(v):
...
             # use first entry in acquisition to populate VA (acq could have 1 or 6 pol pos)
-            self.polarization = model.VAEnumerated(self._pos.keys()[0][-1],
+            self.polarization = model.VAEnumerated(list(self._pos.keys())[0][-1],
                                 choices=polpositions)
 
-        if self._pos.keys()[0][-1]:
+        if list(self._pos.keys())[0][-1]:
```
pieleric: Drop that `if`, which is identical to the `if` just above, and merge the two blocks.

src/odemis/odemisd/modelgen.py
```diff
@@ -845,7 +845,7 @@ def instantiate_component(self, name):
         comp = self._instantiate_comp(name)
 
         # Add to the microscope all the new components that should be child
-        mchildren = self._microscope_ast["children"].values()
+        mchildren = list(self._microscope_ast["children"].values())
```
pieleric: That one shouldn't be needed, as `in` works on a `dict_values`.

src/odemis/odemisd/modelgen.py
```diff
@@ -644,7 +644,7 @@ def get_required_components(self, name):
             pass  # no power supplier
 
         # Support legacy code
-        children = attrs.get("children", {}).values()
+        children = list(attrs.get("children", {}).values())
```
pieleric: That one shouldn't be needed.

src/odemis/odemisd/modelgen.py
```diff
@@ -633,7 +633,7 @@ def get_required_components(self, name):
         ret = set()
 
         attrs = self.ast[name]
-        dependencies = attrs.get("dependencies", {}).values()
+        dependencies = list(attrs.get("dependencies", {}).values())
```
pieleric: Actually, this shouldn't be needed.
However, the code just below is broken (due to the changes to dependencies). I think that instead of appending to `dependencies`, it should add to `ret`.

src/odemis/model/_futures.py
```diff
@@ -146,7 +146,7 @@ def _schedule_work(self):
                 for dep in self._sets_in_progress.values():
                     if dep & dependences:  # intersects?
                         logging.debug("Waiting for scheduling task with dep %s, conflicting with current %s",
-                                      dependences, set.union(*self._sets_in_progress.values()))
+                                      dependences, set.union(*list(self._sets_in_progress.values())))
```
pieleric: I think that `*` works on the `dict_value`, so it shouldn't be needed.

src/odemis/driver/andorshrk.py
```diff
@@ -2141,7 +2141,7 @@ def __init__(self, name, role, children=None, daemon=None, **kwargs):
 
         # duplicate every other VA and Event from the detector
         # that includes required VAs like .exposureTime
-        for aname, value in model.getVAs(dt).items() + model.getEvents(dt).items():
+        for aname, value in list(model.getVAs(dt).items()) + list(model.getEvents(dt).items()):
```
pieleric: Please use `itertools.chain` here too.

src/odemis/util/__init__.py
```diff
@@ -57,7 +57,7 @@ def index_closest(val, l):
     Works also with dict and tuples.
     """
     if isinstance(l, dict):
-        return min(l.items(), key=lambda x:abs(x[1] - val))[0]
+        return min(list(l.items()), key=lambda x:abs(x[1] - val))[0]
```
pieleric: That shouldn't be needed (`min()` is happy to get an iterator).

philipwinklerdelmic:
Done!

PR #977, opened 2019-05-07
acq path: bugfix addendum

philipwinklerdelmic:
Previous bugfix wasn't working correctly. For each future, we must remember the axis in order to check if it has been referenced.

### DIFF
--- a/src/odemis/acq/path.py
+++ b/src/odemis/acq/path.py
@@ -479,7 +479,7 @@ class OpticalPathManager(object):
             elif self.quality == ACQ_QUALITY_BEST:
                 self._setCCDFan(target.role == "ccd")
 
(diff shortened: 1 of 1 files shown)


src/odemis/acq/path.py
```diff
@@ -601,7 +602,9 @@ def _doSetPath(self, path, detector):
 
             try:
                 # move actuator
-                fmoves.append(comp.moveAbs(mv))
```
pieleric: I'd suggest more something like:
`fmoves.append(comp.moveAbs(mv), comp, mv)`.

Then, in the loop waiting for the end of the moves you can write:
`for f, comp, mv in fmoves:`

This avoids having two variables containing the same info *and*, `comp` will be the right value ;-)

PR #992, opened 2019-05-13
bug fix simulator: estimate of acquisition time was off

Sabrina85:


Sabrina85:
It is updated. The simcam testcases are running fine, however not the simstreakcam...

PR #996, opened 2019-05-14
added functionality to test simulating a 1 line ccd autofocus

tepals:
- added 1 line autofocus tests to autofocus_test
- testing for slit image and initial simulator image
- Measure1d measures the focus for a 1 line ccd based on a gaussian fit
- initial sigma estimate from 100 to 10 because sigma will be small for the good focus position

### DIFF
--- a/src/odemis/acq/align/autofocus.py
+++ b/src/odemis/acq/align/autofocus.py
@@ -322,6 +322,7 @@ def _DoBinaryFocus(future, detector, emt, focus, dfbkg, good_focus, rng_focus):
                 logging.debug("Using Spot method to estimate focus")
                 Measure = MeasureSpotsFocus
             elif detector.resolution.value[1] == 1:
+                logging.debug("Using 1d method to estimate focus")
                 Measure = Measure1d
             else:
                 logging.debug("Using Optical method to estimate focus")
@@ -531,6 +532,7 @@ def _DoExhaustiveFocus(future, detector, emt, focus, dfbkg, good_focus, rng_focu
                 logging.debug("Using Spot method to estimate focus")
                 Measure = MeasureSpotsFocus
             elif detector.resolution.value[1] == 1:
+                logging.debug("Using 1d method to estimate focus")
                 Measure = Measure1d
             else:
                 logging.debug("Using Optical method to estimate focus")
--- a/src/odemis/acq/align/autofocus.py
+++ b/src/odemis/acq/align/autofocus.py
@@ -35,6 +35,7 @@ from odemis.util import executeAsyncTask, almost_equal
 from odemis.util.img import Subtract
 from scipy import ndimage
 from scipy.optimize import curve_fit
+from scipy.signal import medfilt
 import threading
 import time
 
@@ -120,20 +121,28 @@ def Measure1d(image):
     Given an image of a 1 line ccd, measure the focus based on the inverse of the width of a gaussian fit of the data.
     It is assumed that the signal is in focus when the width of the signal is smallest and therefore sigma is smallest.
     image (model.DataArray): 1D image from 1 line ccd.
-    returns (float): The focus level of the image.
+    returns (float): The focus level of the image, based on the inverse of the width of a gaussian fitted on the image.
     """
-    def gauss(x, *p):
-        a, b, c, d = p
-        y = a * numpy.exp(-numpy.power((x - b), 2.) / (2. * c ** 2.)) + d
+    # Use the gauss function to fit a gaussian to the 1 line image.
+    def gauss(x, amplitude, pos, width, base):
+        y = amplitude * numpy.exp(-(x - pos) ** 2 / (2 * width ** 2)) + base
         return y
     # squeeze to make sure the image array is 1d.
     signal = numpy.squeeze(image)
+    # Apply a median filter with a kernel of 5, to handle noise with up to 2 neighbouring pixels with a very high value,
+    # resembling a peak, which sometimes happens on CCDs.
+    signal = medfilt(signal, 5)
     x = numpy.arange(len(signal))
-    # give an initial estimate for the parameters of the gaussian fit: [max, mu, sigma, y-offset]
-    p_initial = [max(signal), numpy.argmax(signal), 10, min(signal)]
+    width = max(3.0, 0.01 * len(signal))
+    # determine the indices and the values of the 1% highest points in the signal.
+    max_ids = signal.argsort()[-int(width):]
+    max_sig = signal[max_ids]
+    med_sig = numpy.median(signal)
+    # give an initial estimate for the parameters of the gaussian fit: [amplitude, expected position, width, base]
+    p_initial = [numpy.median(max_sig) - med_sig, numpy.median(max_ids), width, med_sig]
     # Use curve_fit to fit the gauss function to the data. Use p_initial as our initial guess.
     popt, pcov = curve_fit(gauss, x, signal, p0=p_initial)
-    # The focus metric is the inverse of sigma of the gaussian fit; a smaller sigma is a higher focus level.
+    # The focus metric is the inverse of width of the gaussian fit (a smaller width is a higher focus level).
     return 1 / abs(popt[2])
 
 
--- a/src/odemis/acq/align/autofocus.py
+++ b/src/odemis/acq/align/autofocus.py
@@ -34,6 +34,7 @@ from odemis.model import InstantaneousFuture
 from odemis.util import executeAsyncTask, almost_equal
 from odemis.util.img import Subtract
 from scipy import ndimage
+from scipy.optimize import curve_fit
 import threading
 import time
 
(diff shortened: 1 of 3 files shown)


src/odemis/driver/spectrometer.py
```diff
@@ -121,7 +121,7 @@ def __init__(self, name, role, dependencies, **kwargs):
         max_bin = (dt.binning.range[1][0], dt.resolution.range[1][1])
         # Initial binning is minimum binning horizontally, and maximum vertically
         self._binning = (1, dt.resolution.range[1][1])
-        self._max_det_vbin = dt.binning.range[1][1]
```
pieleric: That was only for debugging. Please drop it.

src/odemis/acq/align/test/autofocus_test.py
```diff
@@ -602,5 +603,76 @@ def test_multi_det(self):
...
+        if self.backend_was_running:
+            self.skipTest("Running backend found")
+
+    @timeout(1000)
+    def test_autofocus_spect(self):
+        """
+        Test AutoFocus on 1 line CCD for example spectrum.
+        """
+        self.focus.moveAbs({"z": self._good_focus - 200e-6}).result()
+        self.spectrometer.depthOfField._set_value(1e-6, force_write=True)
```
pieleric: The right way to change the `depthOfField` VA is to update the LENS_* metadata on the detector. See `odemis.model._component.DigitalCamera` .

src/odemis/acq/align/autofocus.py
```diff
@@ -114,6 +115,21 @@ def MeasureOpticalFocus(image):
     return cv2.Laplacian(image, cv2.CV_64F).var()
 
 
+def Measure1d(image):
+    def gauss(x, *p):
```
pieleric: Please add docstrings.

src/odemis/acq/align/autofocus.py
```diff
@@ -114,6 +115,21 @@ def MeasureOpticalFocus(image):
...
+        a, b, c, d = p
+        y = a * numpy.exp(-numpy.power((x - b), 2.) / (2. * c ** 2.)) + d
+        return y
+
+    signal = numpy.squeeze(image)
```
pieleric: It'd be nice to explain the rough idea of the algorithm. It fits a Gaussian? And then report the inverse of the gaussian width?

src/odemis/acq/align/autofocus.py
```diff
@@ -114,6 +115,21 @@ def MeasureOpticalFocus(image):
...
+    x = numpy.arange(len(signal))
+    p_initial = [max(signal), numpy.argmax(signal), 10, min(signal)]
```
pieleric: Please add a comment explain what are these 4 parameters.

src/odemis/acq/align/test/autofocus_test.py
```diff
@@ -602,5 +603,76 @@ def test_multi_det(self):
...
+            logging.info("A running backend is already found, skipping tests")
+            cls.backend_was_running = True
+            return
+        except IOError as exp:
+            logging.error(str(exp))
+            raise
+
+        # find components by their role
+        cls.ccd = model.getComponent(role="ccd")
+        cls.spectrometer = model.getComponent(role="spectrometer-integrated")
```
pieleric: If removing the special code to prevent hardware binning, it might work less well. Please check whether the test cases pass (probably), or you need to introduce extra tricks to simulate a 1 line CCD properly.

tepals:
updated

src/odemis/acq/align/autofocus.py
```diff
@@ -114,6 +115,28 @@ def MeasureOpticalFocus(image):
...
 
+def Measure1d(image):
+    """
+    Given an image of a 1 line ccd, measure the focus based on the inverse of the width of a gaussian fit of the data.
+    It is assumed that the signal is in focus when the width of the signal is smallest and therefore sigma is smallest.
+    image (model.DataArray): 1D image from 1 line ccd.
+    returns (float): The focus level of the image.
+    """
+    def gauss(x, *p):
+        a, b, c, d = p
```
pieleric: Isn't it possible to just write `def gauss(x, a,b,c,d):`?
Also, "a,b,c,d" are really obfuscated names. I'd suggest to use the same names as in `util.peak.GaussianFit()`: pos, width, amplitude, offset (not sure of the order though).
It might also be interesting to compare your formula to the one used in `GaussianFit()` and check if they can learn from each other...
At least, it should be possible to drop the `.`, and replace `numpy.power(..., 2) by `(...)**2`.

src/odemis/acq/align/autofocus.py
```diff
@@ -114,6 +115,28 @@ def MeasureOpticalFocus(image):
...
+        y = a * numpy.exp(-numpy.power((x - b), 2.) / (2. * c ** 2.)) + d
+        return y
+    # squeeze to make sure the image array is 1d.
+    signal = numpy.squeeze(image)
+    x = numpy.arange(len(signal))
+    # give an initial estimate for the parameters of the gaussian fit: [max, mu, sigma, y-offset]
+    p_initial = [max(signal), numpy.argmax(signal), 10, min(signal)]
```
pieleric: BTW, use `signal.max()` and `signal.min()` instead of the standard Python functions. This avoids the conversion to Python values for each elements... and makes it 10x faster.

src/odemis/acq/align/autofocus.py
tepals: Updated the pull request, removed the *p. I think the biggest difference between the two functions is the base offset I'm using. That is necessary here because the base can be at 800 or something and the curve fit returns garbage without that offset, I don't know if that would improve the other function.

src/odemis/acq/align/autofocus.py
```diff
@@ -120,21 +121,31 @@ def Measure1d(image):
...
-    def gauss(x, *p):
-        a, b, c, d = p
-        y = a * numpy.exp(-numpy.power((x - b), 2.) / (2. * c ** 2.)) + d
+    # Use the gauss function to fit a gaussian to the 1 line image.
+    def gauss(x, amplitude, pos, width, base):
+        y = amplitude * numpy.exp(-(x - pos) ** 2 / (2 * width ** 2)) + base
         return y
     # squeeze to make sure the image array is 1d.
     signal = numpy.squeeze(image)
+    # Apply a median filter with a kernel of 5, to filter out (at the most 2 neighbouring) outliers.
```
pieleric: This comments explains exactly what the code below does. Please explain _why_ it does it. IE, to handle noise with up to 2 pixels with very large values, that would look like a peak, which sometimes happens on CCDs.

src/odemis/acq/align/autofocus.py
```diff
@@ -120,21 +121,31 @@ def Measure1d(image):
...
+    p_initial = [numpy.median(max_sig) - numpy.median(signal), numpy.median(max_ids), 0.01 * len(signal),
+                 numpy.median(signal)]
     # Use curve_fit to fit the gauss function to the data. Use p_initial as our initial guess.
     popt, pcov = curve_fit(gauss, x, signal, p0=p_initial)
-    # The focus metric is the inverse of sigma of the gaussian fit; a smaller sigma is a higher focus level.
-    return 1 / abs(popt[2])
+    # Determine the goodness of the fit. It is expected that in the area of focus the goodness of fit is high.
+    y_fit = gauss(x, *popt)
+    r2 = 1 - (((signal - y_fit) ** 2).sum() / ((signal - signal.mean()) ** 2).sum())
+    # The focus metric is a combination of the inverse of sigma of the gaussian fit (a smaller sigma is a higher focus
```
pieleric: Have you tried using the amplitude (compared to the base)? It's very typical that focused image have brighter pixels, so it should be a quite reliable indicator.
In addition, I'm a little worried by the gaussian fit, because I wonder how well it performs if the signal happens to not really be a gaussian (eg, a square wave).

tepals: The problem I'm seeing with using the amplitude (for the simulated focus test) is that there is a spread around the focus point where the amplitude is equal to that of the amplitude in the focus point. 

I tried combining the amplitude with the width, however I can't think of a way to scale them so I can add them together since the amplitude has a much higher value than the width.

I also tried what you suggested in the email to take the sum of the X% highest pixels. That has the following problem: if the in focus peak has a very small width and you go slightly out of focus there are more bright pixels thus a the sum of the X% highest pixels is higher when slightly out of focus then when in focus.

I agree that the goodness of fit is probably not the best metric, especially for a square wave.

tepals:
updated the pull request. Left the focus metric to be 1 /  width of the gaussian fit

src/odemis/acq/align/test/autofocus_test.py
```diff
@@ -602,5 +603,76 @@ def test_multi_det(self):
...
+        if self.backend_was_running:
+            self.skipTest("Running backend found")
+
+    @timeout(1000)
+    def test_autofocus_spect(self):
+        """
+        Test AutoFocus on 1 line CCD for example spectrum.
+        """
+        self.focus.moveAbs({"z": self._good_focus - 200e-6}).result()
+        self.ccd.updateMetadata({model.MD_LENS_RI: 1.0e-2})
```
pieleric: If they all need this, I'd recommend to move this to the `setUpClass()`, or even in the microscope file. 
BTW, what is this for? IIRC, that's for depthOfField, which influences the minimum step? A RI of 0.01 seems really weird. I wonder if instead we could just increase the NA, or maybe increase the focus step size...

src/odemis/acq/align/test/autofocus_test.py
```diff
@@ -602,5 +603,76 @@ def test_multi_det(self):
...
+        f = align.AutoFocus(self.spectrometer, None, self.focus, method=MTD_BINARY)
+        foc_pos, foc_lev = f.result(timeout=900)
+        logging.debug("Found focus at {} good focus at {}".format(foc_pos, self._good_focus))
+        numpy.testing.assert_allclose(foc_pos, self._good_focus, atol=2.5e-5)
```
pieleric: I guess this 2.5e-5 is not coming completely out of nowhere, but just big enough to work?
As the focus step size is set (in the microscope file) to 10.9e-6m, that's essentially 2 focus steps. That sounds fine, but could be documented.

In test_autofocus_slit(), that sounds a little bit too much optimistic, as it'll fail if it's not exactly the right step.

src/odemis/acq/align/test/autofocus_test.py
```diff
@@ -602,5 +603,76 @@ def test_multi_det(self):
...
+
+    @timeout(1000)
+    def test_autofocus_slit(self):
+        """
+        Test AutoFocus on 1 line CCD for an image of a slit.
+        """
+        # Change image to slit image.
+        data = tiff.read_data(os.path.join(TEST_IMAGE_PATH, "brightlight-on-slit-spccd-simple.ome.tiff"))
```
pieleric: Be aware that the tests are _not_ run in any order (as methods on a object are actually contained in a dictionary, and Python dictionaries do not maintain order). So the ccd could have the image set before runnning `test_autofocus_spect()`, which would run with that same image.
The simplest might be to explicitly set the image also in `test_autofocus_spect()`.

tepals:
updated PR. Removed the changing of  ```model.MD_LENS_RI``` to change the depth of field, and just increased the absolute tolerance two ```2.5e-5``` everywhere and added a comment about it.

PR #998, opened 2019-05-14
odemisd modelgen: bugfix 'is_leaf' function after children refactoring

philipwinklerdelmic:


src/odemis/odemisd/modelgen.py
```diff
@@ -548,9 +552,11 @@ def _get_container(self, name):
                 deps_cont.add(cont)
 
         if len(deps_cont) == 1:
+            logging.debug("Found one good container: %s", deps_cont)
```
pieleric: Please drop these log messages. They were useful only for debugging.

philipwinklerdelmic: done

PR #1001, opened 2019-05-15
do not show warning when time or wl list is empty

Sabrina85:


Sabrina85:
Closed as we modify drivers to not append an empty WL_LIST to the metadata ever.

PR #1005, opened 2019-05-20
python3-migration-bugfix

philipwinklerdelmic:


src/odemis/driver/pigcs.py
```diff
@@ -4024,7 +4024,7 @@ def _processCommand(self, com):
 
         com = m.group("com") # also removes the \n at the end if it's there
         # split into arguments separated by spaces (not including empty strings)
-        args = [bool(a) for a in com.split(" ")]
+        args = [a for a in com.split(" ") if a]
```
pieleric: Just fixed in PR#1004 -> drop

philipwinklerdelmic: done

PR #1015, opened 2019-05-22
acq stream: increase timeout for SEMTemporalMDStream

philipwinklerdelmic:


philipwinklerdelmic:
Ok, let's do it properly and implement synchronization support. I'll put it on my list for next week.

philipwinklerdelmic:
I just realized: the dwell time of 1e-6 is the dwell time for the drift correction. The actual dwell time is 1. I think these values are fine.

philipwinklerdelmic:
changed 1.5 to 2.5

PR #1018, opened 2019-05-22
sparc2 autofocus: better handling of cancellation

pieleric:


src/odemis/gui/cont/tabs.py
```diff
@@ -3903,6 +3903,8 @@ def _onAutofocus(self, active):
     def _on_autofocus_done(self, future):
         try:
             future.result()
+        except CancelledError:
```
VictoriaDel: We do need this don't we? Or we assume that any "cancellation" is checked by the big future?

pieleric: Yes, we do need this... that's why I added it ;-)
To be clear, in case of cancellation, the following happens:
1. the GUI calls bigFuture.cancel()
2. _CancelSparc2Autofocus() is called
3. future._state is set to CANCELLED
4. _running_subf is cancelled()
5. In _DoSparc2AutoFocus(), two slightly different cases can happen depending on when the cancellation happened: either CancelledError is raised from the running_subf.result(), or  it finds out that the state is CANCELLED and raises CancelledError.
6. _DoSparc2AutoFocus() ends early, with a CancelledError raised.
7. In the GUI, _on_focus_done() is called
8. future.result() raises the CancelledError which was raised.

PR #1022, opened 2019-05-23
Bug fix for export: Temporal spectrum data was flipped on horizontal axis

Sabrina85:


src/odemis/gui/util/img.py
```diff
@@ -1197,30 +1197,29 @@ def clip_data_window(hrange, vrange, xd, yd):
...
-        raise LookupError("No pixel selected to pick a spectrum")
-    spectrum_range, unit = spectrum.get_spectrum_range(spec)
-
-    if raw:
-        if unit == "m":
-            spec.metadata[model.MD_WL_LIST] = spectrum_range
-        spec.metadata[model.MD_ACQ_TYPE] = model.MD_AT_SPECTRUM
-        return spec
-    else:
+    if raw:  # csv
```
pieleric: In the case of the point spectrum/chronogram, the `.image` is the same as `.projectAsRaw()`. So that previous code wasn't incorrect. However, I think it's a little better now, as it makes all the 4 functions follow the same pattern.

Sabrina85: ok good :)

PR #1023, opened 2019-05-24
Bug in andorshrk: WL_LIST was not updated on the detector in AndorSpec

Sabrina85:


Sabrina85:
Please have a look at the descriptive text. Not sure it is correct...

PR #1025, opened 2019-05-24
Flipped mirror:

VictoriaDel:


src/odemis/gui/comp/overlay/world.py
```diff
@@ -1521,19 +1522,21 @@ def set_mirror_dimensions(self, parabola_f, cut_x, cut_offset_y, hole_diam):
...
+        self.cut_offset_y = abs(cut_offset_y)
 
         # The radius of the circle shaped edge facing the detector
         # We don't care about cut_x, but "cut_y"
         # Use the formula y = x²/4f
         self.parabole_cut_radius = 2 * math.sqrt(parabola_f * cut_x)
 
-        # The mirror is cut horizontally just above the symmetry line
-        self.cut_offset_y = cut_offset_y
+        # # The mirror is cut horizontally just above the symmetry line
```
pieleric: Don't comment out. Just move the comment to the new place, and remove the old lines.

src/odemis/gui/comp/overlay/world.py
```diff
@@ -1521,19 +1522,21 @@ def set_mirror_dimensions(self, parabola_f, cut_x, cut_offset_y, hole_diam):
...
+        # self.cut_offset_y = cut_offset_y
 
+        # self.mirror_height = self.parabole_cut_radius - cut_offset_y
         self.mirror_height = self.parabole_cut_radius - self.cut_offset_y
         # The number of radians to remove from the left and right of the semi-circle
+        # self.rad_offset = math.atan2(cut_offset_y, self.parabole_cut_radius)
```
pieleric: Drop the old line.

src/odemis/gui/cont/tabs.py
```diff
@@ -2985,6 +2985,11 @@ def __init__(self, name, button, panel, main_frame, main_data):
         panel.html_moi_doc.SetBorders(0)
         panel.html_moi_doc.LoadPage(self.doc_path)
 
+        if main_data.lens:  # TODO do we need this or is lens not always there?
```
pieleric: Indeed, you can assume that the lens is always present.

src/odemis/gui/cont/tabs.py
```diff
@@ -3310,21 +3315,17 @@ def __init__(self, name, button, panel, main_frame, main_data):
         else:
             self.panel.btn_manualfocus.Bind(wx.EVT_BUTTON, self._onManualFocus)
 
-        if "center-align" in tab_data.align_mode.choices:
+        if "center-align" not in tab_data.align_mode.choices:
```
pieleric: I'd recommend to inverse the `if`, because it is a little easier to read when the condition is not negated. (so move the negative into the else block).

src/odemis/gui/cont/tabs.py
```diff
@@ -4010,6 +4016,15 @@ def _onSEMMag(self, mag):
             self._moi_stream.is_active.value = True
         self.panel.vp_moi.canvas.fit_view_to_next_image = True
 
+    def _onMirrorDimensions(self, focusDistance):
```
pieleric: Run in main thread, with `@call_in_wx_main`

src/odemis/gui/main_xrc.py
```diff
@@ -623,7 +624,7 @@ def __init__(self, parent):
 
 def __init_resources():
     global __res
-    __res = xrc.XmlResource()
+    __res = xrc.EmptyXmlResource()
```
pieleric: This one shouldn't be changed. Maybe you're still using wxpython 3, instead of wxpython 4?

src/odemis/util/angleres.py
```diff
@@ -26,7 +26,9 @@
 from scipy.interpolate import LinearNDInterpolator
 import numpy
 from odemis import model
-# import matplotlib.pyplot as plt
+import matplotlib.pyplot as plt
```
pieleric: Leave the original line as-is.

src/odemis/util/angleres.py
```diff
@@ -69,11 +71,16 @@ def _ExtractAngleInformation(data, hole):
         pixel_size = data.metadata[model.MD_PIXEL_SIZE]
         pole_x, pole_y = data.metadata[model.MD_AR_POLE]
         parabola_f = data.metadata.get(model.MD_AR_PARABOLA_F, AR_PARABOLA_F)
+        focus_distance = data.metadata.get(model.MD_AR_FOCUS_DISTANCE, AR_FOCUS_DISTANCE)
+        # TODO new MD for inverted mirror or not?
```
pieleric: No. The "new metadata" is that AR_FOCUS_DISTANCE can be negative.

src/odemis/driver/test/static_test.py
```diff
@@ -69,6 +69,10 @@ def test_mag_choices(self):
             comp.magnification.value = 2.0
         comp.terminate()
 
+    def test_configurations(self):
+        configurations = {"Mirror up": {pole_pos: [458, 519], focus_dist: 0.5e-3},
+                          "Mirror down": {pole_pos: [634, 652], focus_dist: -0.5e-3}}
+        comp = static.OpticalLens("test", "lens", 1, configurations_choices=configurations)
```
pieleric: Extend the test case. In particular, you need to change the configuration, and check that the `.polePosition` and `.focusDistance` VAs are updated.

src/odemis/driver/static.py
```diff
@@ -45,7 +45,8 @@ class OpticalLens(model.HwComponent):
 
     def __init__(self, name, role, mag, mag_choices=None, na=0.95, ri=1,
                  pole_pos=None, x_max=None, hole_diam=None,
-                 focus_dist=None, parabola_f=None, rotation=None, **kwargs):
+                 focus_dist=None, parabola_f=None, rotation=None,
+                 configurations=None, **kwargs):
         """
         name (string): should be the name of the product (for metadata)
         mag (float > 0): magnification ratio
```
pieleric: Update the docstring, to explain the new "configurations" parameter.

src/odemis/driver/static.py
```diff
@@ -110,7 +111,21 @@ def __init__(self, name, role, mag, mag_choices=None, na=0.95, ri=1,
         if rotation is not None:
             self.rotation = model.FloatContinuous(rotation, (0, 2 * math.pi),
                                                   unit="rad")
-
+        if configurations is not None:
+            self._configurations = configurations
+            conf = list(configurations.keys())[0]
```
pieleric: Ideally, that initial conf would be picked based on the current values of the VAs.

install/linux/usr/share/odemis/sim/sparc2-sim.odm.yaml
```diff
@@ -240,12 +227,15 @@ SPARC2: {
     },
     affects: ["Camera", "Spectral Camera", "Spectrometer IR"],
 }
-
 "Optical Path Properties": {
     class: static.OpticalLens,
     role: lens,
     # Standard mirror config
     init: {
+       configurations: {
```
pieleric: Only these lines are to be added. All the other changes in this file should be reverted.

src/odemis/gui/cont/tabs.py
VictoriaDel: Done ! I did it that way because the condition about "lens-align" was done by checking the negative first. Maybe we could change this one to be clearer.

PR #1027, opened 2019-05-27
LongExpTime

Sabrina85:


Sabrina85:
Code requires the bug fix regarding temporal spectrum data assembly to be merged first.

src/odemis/acq/stream/_helper.py
```diff
@@ -93,6 +93,25 @@ def __init__(self, name, detector, dataflow, emitter, scanner=None, sstage=None,
         #  * Rep set: Rep (as requested) + ROI (current) → PxS (updated)
         #    The repetition is adjusted to fit the hardware limits
 
+        # If "exposureTime" not detvas and detector has VA "exposureTime",
```
pieleric: It should be possible to make that comment less identical to the code it's supposed to document. How about: "If no local exposureTime VA is requested, automatically provide an integrationTime VA, which allows longer exposure by doing image integration."

src/odemis/acq/stream/_helper.py
```diff
@@ -93,6 +93,25 @@ def __init__(self, name, detector, dataflow, emitter, scanner=None, sstage=None,
...
+        if "detvas" in kwargs.keys() and "exposureTime" not in kwargs["detvas"] and model.hasVA(detector, "exposureTime"):
+
+            self.hwExpTime = self.detector.exposureTime.value
+            # Number of images that need to be acquired for the requested exposure time.
+            # If not integration time, default is 1 image.
+            self.integrationCounts = model.VigilantAttribute(1, readonly=True)
+
+            # increase exposure time range to perform image integration
+            # TODO: for now we specify a max integration time by using a fixed multiple of the exp time
+            # TODO: for now only used in Braunschweig which has max exp time of 10s -> so now 1h
```
pieleric: Indeed, this "6*60" constant is a little arbitrary. Maybe we could just "round" it to 1000, and limit the max time to 1h (either here, or in the gui.data.conf).

src/odemis/acq/stream/_helper.py
```diff
@@ -149,6 +168,175 @@ def __init__(self, name, detector, dataflow, emitter, scanner=None, sstage=None,
...
+        """
+        Called when a new image has arrived from the detector.
+        Usually the dataflow subscribes to onNewData.
+        :param dataflow: (model.Dataflow) The dataflow.
+        :param data: (model.DataArray). The new image that has arrived from the detector.
+        """
+        if hasattr(self, "integrationTime"):
+            self.raw = (self.raw + [data])[-self.integrationCounts.value:]
+        else:
+            self.raw = (self.raw + [data])
```
pieleric: This is appending (forever). I think you want to just write `self.raw = [data]`.

src/odemis/acq/stream/_helper.py
```diff
@@ -149,6 +168,175 @@ def __init__(self, name, detector, dataflow, emitter, scanner=None, sstage=None,
...
+
+        # TODO both methods call integrateImage -> make easier?
```
pieleric: Yes, but not at the same frequency. I'd just leave it as-is. If the integration count is very large, the main issue might be memory, before the CPU (as, anyway, the new data wouldn't arrive fast).

src/odemis/acq/leech.py
```diff
@@ -294,14 +310,15 @@ def start(self, dt, shape):
 #             self.series_start()
 
         super(AnchorDriftCorrector, self).start(dt, shape)
-        self._next_px = self._dc_estimator.estimateCorrectionPeriod(self.period.value, dt, shape)
+        self._next_acq = self._dc_estimator.estimateCorrectionPeriod(self.period.value, dt, shape[-2:])
```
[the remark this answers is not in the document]
Sabrina85: `self._num_acq_until_next_corr ` :D

Sabrina85:
Btw: I get following message in the log, whenever recording an image: 

`Fast conversion cannot run: Optimised version only works on uint16 (got uint32)`

Do we need to handle this?

pieleric:
That message just indicates that the "fast path" cannot be used to convert the greyscale image to RGB. That's normal, as indeed the data is now 32 bits, and we don't have dedicated fast function that support this type. Currently, the hardware we have, when we need fast display, only generates 8 bit or 16 bits data. So that's fine.

src/odemis/acq/stream/_sync.py
```diff
@@ -812,25 +823,45 @@ def _estimateRawAcquisitionTime(self):
...
-        Read the SEM and CCD stream settings and adapt the SEM scanner
-        accordingly.
-        return (float): estimated time for a whole CCD image
+        Read the SEM and CCD stream settings and adapt the SEM scanner accordingly.
+        :returns: exp + readout (float): Estimated time for a whole (but not integrated) CCD image.
         """
-        exp = self._sccd._getDetectorVA("exposureTime").value  # s
+        if model.hasVA(self._sccd, "integrationTime"):
+            # calculate exposure time to be set on detector
+            exp = self._sccd.integrationTime.value / self._sccd.integrationCounts.value
```
pieleric: I'm not very convinced by the idea of using `self._sccd.integrationCounts.value`. Primarily, because it's coming from a separate class, which has this VA mostly only for display purpose (so, not so important to have correct), and here we rely on it to have the right acquisition parameters. On the other you could argue that this information to the user should always be correct, and if it's not, it should be fixed, and not do different things in the acquisition.

However, I'd really argue against our own `self._integrationCounts`, and instead, I'd suggest to return from this function "exp_time, integ_count". That second parameter would be then passed to the functions which care about it (so it makes the code more explicit).

src/odemis/acq/stream/_sync.py
```diff
@@ -812,25 +823,45 @@ def _estimateRawAcquisitionTime(self):
             res = self._sccd._getDetectorVA("resolution").value
             readout = numpy.prod(res) / ro_rate
 
-            exp = self._sccd._getDetectorVA("exposureTime").value
+            if hasattr(self._sccd, "integrationTime"):
+                exp = self._sccd.integrationTime.value
+                # calculate number of images to integrate
+                hwExpTimeMax = self._sccd.detector.exposureTime.range[1]
+                numImages2integrate = int(numpy.ceil(exp / hwExpTimeMax))
+                readout *= numImages2integrate
```
pieleric: Use the same code as in `_adjustHardwareSettings()`

src/odemis/acq/stream/_sync.py
```diff
@@ -812,25 +823,45 @@ def _estimateRawAcquisitionTime(self):
...
+        Read the SEM and CCD stream settings and adapt the SEM scanner accordingly.
+        :returns: exp + readout (float): Estimated time for a whole (but not integrated) CCD image.
         """
-        exp = self._sccd._getDetectorVA("exposureTime").value  # s
+        if model.hasVA(self._sccd, "integrationTime"):
+            # calculate exposure time to be set on detector
+            exp = self._sccd.integrationTime.value / self._sccd.integrationCounts.value
+            # TODO should we have a try/except here in case of value out of range
+            #  or should that already be handled somewhere else...?
+            self._sccd.detector.exposureTime.value = exp
```
pieleric: Actually, the hardware may pick a different value (for instance, because it always rounds the value to 0.1s). So you might want to play it super safe, which such (seemingly redundant) code:
```
exp = self._sccd.integrationTime.value / self._sccd.integrationCounts.value
self._ccd.exposureTime.value = self._ccd.exposureTime.clip(exp)
integ_count = int(math.ceil(self._sccd.integrationTime.value / self._ccd.exposureTime.value))
```
(and maybe add an extra log if `integ_count != self._sccd.integrationCounts.value`.)

src/odemis/acq/stream/_sync.py
```diff
@@ -327,7 +337,7 @@ def acquire(self):
         executeAsyncTask(f, self._runAcquisition, args=(f,))
         return f
 
-    def _updateProgress(self, future, dur, current, tot, bonus=0):
+    def _updateProgress(self, future, dur, current, tot, bonus=0):  # TODO?
```
pieleric: For now `tot` is probably incorrect when integrationCount > 1. Maybe it's just a matter of updating the caller.

src/odemis/acq/stream/_helper.py
```diff
@@ -93,6 +93,21 @@ def __init__(self, name, detector, dataflow, emitter, scanner=None, sstage=None,
...
+        # which allows longer exposure by doing image integration.
+        if "exposureTime" not in kwargs.get("detvas", {}) and model.hasVA(detector, "exposureTime"):
+            self._hwExpTime = self.detector.exposureTime.value
+            # Number of images that need to be acquired for the requested exposure time.
+            # If not integration time, default is 1 image.
+            self.integrationCounts = model.VigilantAttribute(1, readonly=True)
+
+            # increase exposure time range to perform image integration
+            # TODO: for now we specify a max integration time by using a fixed multiple of the exp time, but max 1h
+            longExpTimeRange = (detector.exposureTime.range[0], numpy.max((detector.exposureTime.range[1]*1000, 3600)))
```
pieleric: `numpy.max` -> `max`

pieleric: Rename the variable `integrationTimeRange` ?

src/odemis/acq/stream/_helper.py
```diff
@@ -149,6 +164,172 @@ def __init__(self, name, detector, dataflow, emitter, scanner=None, sstage=None,
...
+        # First update the value, before the intensityRange subscribers are called...
+        self.histogram._value = chist
+
+        if self.auto_bc.value:
+            self._recomputeIntensityRange()
+
+        # Notify last, so intensityRange is correct when subscribers get the new histogram.
+        self.histogram.notify(chist)
+
+    def _setCamExpTime(self, value):
```
pieleric: rename to `_setIntegrationTime()`, to have the same name as the VA.

src/odemis/acq/stream/_sync.py
```diff
@@ -915,22 +947,23 @@ def _runAcquisitionEbeam(self, future):
...
+            img_time = self._adjustHardwareSettings()
             dwell_time = self._emitter.dwellTime.value
             sem_time = dwell_time * numpy.prod(self._emitter.resolution.value)
             spot_pos = self._getSpotPositions()  # list of center positions for each point of the ROI
             logging.debug("Generating %dx%d spots for %g (dt=%g) s",
-                          spot_pos.shape[1], spot_pos.shape[0], px_time, dwell_time)
+                          spot_pos.shape[1], spot_pos.shape[0], img_time, dwell_time)
             rep = self.repetition.value  # (int, int): number of pixels in the ROI (X, Y)
-            tot_num = numpy.prod(rep)  # total number of images to acquire
+            tot_num = numpy.prod(rep) # total number of images to acquire
```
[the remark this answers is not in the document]
Sabrina85: Sorry this happened when I changed this code a couple of times by multiplying with the integrationCount and then ending up doing it at a different place.

src/odemis/acq/stream/_sync.py
Sabrina85: This should work? Would also simplify the line with the .exposureTime.

`            if hasattr(self._sccd, "integrationTime"):
                exp = self._sccd.integrationTime.value
                readout *= self._sccd.integrationCounts.value
            else:
                exp = self._sccd.exposureTime.value`

src/odemis/acq/stream/_sync.py
```diff
@@ -1088,20 +1137,64 @@ def _runAcquisitionEbeam(self, future):
...
             self._acq_done.set()
 
-    def _waitForImage(self, px_time):
+    def _integrateImages(self, n):
+        """
+        Integrate/sum the last n acquired images in self._acq_data and replace by the
+        integrated image. If values overflow, clip the image.
+        :param n: (int) Number of images that need to be integrated (summed).
+        """
+        # TODO is there an sem_idx or similar? How to handle if more than 1 sem det -> for loop?
```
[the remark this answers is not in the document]
Sabrina85: Should the fast path be here or where the `integrateImages` is called from? I think in the polarimetry pr, you suggested to put it in the caller?

Should the `MD_INTEGRATION_COUNT` also be appended when =1?

Is there always only one SEM stream and one detector stream? 
Is there a `sem_idx` similar to `self._ccd_idx`

src/odemis/acq/stream/_sync.py
```diff
@@ -1088,20 +1137,64 @@ def _runAcquisitionEbeam(self, future):
...
+        # get the SEM images corresponding to the last acquisition and average
+        data_sem = numpy.mean(self._acq_data[0][-n:], axis=0)
+        # TODO anything need to consider regarding type?
```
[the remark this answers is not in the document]
Sabrina85: Why update the dwell time? We average, don't we? So isn't the dwell time the same per image as before?

src/odemis/acq/stream/_sync.py
Sabrina85: I am confused. We already have a self._integrationCounts VA here. And we set it default to 1 also if no integrationVA. Can we quickly discuss please, also in regard to the next comments?

src/odemis/acq/stream/_sync.py
Sabrina85: Ah didn't process the comment on the code in `adjustHardwareSettings` yet.

src/odemis/acq/stream/_sync.py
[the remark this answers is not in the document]
Sabrina85: At this place `tot_num` is indeed the number of ebeam positions to be acquired. It excludes any information on how many images per ebeam pos (polarizer) or info on image integration.

src/odemis/util/__init__.py
```diff
@@ -43,6 +44,27 @@
...
+def get_best_dtype_for_acc(idtype, count):
+    """
+    Computes the smallest dtype that allows to integrate the number of inputs without overflowing.
+    :param idtype: (dtype) dtype of the input (the raw data that is integrated).
+    :param count: (int) Number of values/images to be accumulated/integrated.
+    :returns: (dtype) The best fitting dtype.
+    """
+    maxval = numpy.iinfo(idtype).max * count
+
+    if maxval <= numpy.iinfo(numpy.uint64).max:
```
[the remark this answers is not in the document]
Sabrina85: Mhm I think I had uint32 or something for the first if...

src/odemis/acq/stream/_sync.py
pieleric: I think this code would work fine for the time estimation. (Moreover, you cannot really set the exposureTime on the hardware, so you may not be able to have exactly the same code as `adjustHardwareSettings()`.

src/odemis/acq/stream/_sync.py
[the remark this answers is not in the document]
pieleric: Whether the `if` is inside the function or in the caller, of course, eventually it's the same behaviour. So it's a little bit a matter of taste and I think also a matter of the "semantic". Personally, I'd say that here it's just a "fast path" (because the standard path would still work fine with n == 1), so it'd feel more logical to hide this implementation detail inside the function. In the case of the polarimetry export, it'd read more like `draw_logo(None)`, where the code can simply not work with that special parameter value.

The default value of MD_INTEGRATION_COUNT is 1. So I don't think you need to put it explicitly. I'd suggest to only put it if that simplifies the code (which I don't think is the case).

There is no `sem_idx`, because the idea is that every stream which is not `_ccd_idx` should be considered a SEM stream. BTW, in that function, I think that if you use MD_DET_TYPE, you don't even need to care about the SEM vs CCD.

src/odemis/acq/stream/_sync.py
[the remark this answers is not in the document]
pieleric: No, the dwell time is the total time over which the beam stayed on the pixel. If the beam stayed 3 x 10 s, or 6 x 5s, in both cases the final data is equivalent to a dwell time of 30 s.

src/odemis/acq/stream/_sync.py
pieleric: self._integrationCounts is only used in one function: the caller of `_adjustHardwareSettings()`. However, it's computed by a separate class, and there is a slight chance that it's incorrect in case the hardware actually accepts a different exposure time. So I'd suggest that this function returns the "definitive" integ_counts, and that value would be used by the caller.

(And as a tiny bonus, that allows you to remove the couple of lines needed to initialise properly this VA in `__init__`!)

src/odemis/acq/stream/_sync.py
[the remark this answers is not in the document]
Sabrina85: ok. One more question for the total estimated time: If I have to integrate 3 images each 1s for the optical (total = 3s), I will average 3 sem images each 10us (total = 30us). So the total should be something not much bigger than 3.03 s (plus some overhead, readout etc.). Or does the dwell time also include the time the ebeam stays at the position until also the opitcal image was acquired? Now I get something around 7s, which I think is too much.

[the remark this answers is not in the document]
Sabrina85: I ran a few tests with different number of pixelsizes and it looks actually quite good now. I adjusted the dwelltime *= integrationCount. Still surprised that for 1 px the sem and optical time was similar.

Sabrina85:
Not 100% sure about the estimated acq time.

src/odemis/acq/stream/_sync.py
```diff
@@ -1140,46 +1143,55 @@ def _runAcquisitionEbeam(self, future):
...
+        integrated image. If values overflow, clip the image. If number of images = 1,
+        no image integration necessary.
         :param n: (int) Number of images that need to be integrated (summed).
         """
-        # TODO is there an sem_idx or similar? How to handle if more than 1 sem det -> for loop?
-        # get the SEM images corresponding to the last acquisition and average
-        data_sem = numpy.mean(self._acq_data[0][-n:], axis=0)
-        # TODO anything need to consider regarding type?
+        # check if we need to integrate images
+        if not n == 1:
```
pieleric: Invert the test case, and immediately return if n == 1. This way, you gain one indentation level ;-)

src/odemis/acq/stream/_sync.py
```diff
@@ -1140,46 +1143,55 @@ def _runAcquisitionEbeam(self, future):
...
+            # SEM
+            for stream_idx in range(len(self._acq_data)):
```
pieleric: I think you can avoid using explicitly the index, and just manipulate the lists of DataArrays, like:
`for das in self._acq_data:`

src/odemis/acq/stream/_sync.py
```diff
@@ -1140,46 +1143,55 @@ def _runAcquisitionEbeam(self, future):
...
+                det_type = self._acq_data[stream_idx][-1].metadata[model.MD_DET_TYPE]
```
pieleric: Don't rely on the metadata to be present. => `das[-1].metadata.get(model.MD_DET_TYPE, model.MD_DT_INTEGRATING)`

src/odemis/acq/stream/_sync.py
```diff
@@ -1140,46 +1143,55 @@ def _runAcquisitionEbeam(self, future):
...
+        no image integration necessary.
         :param n: (int) Number of images that need to be integrated (summed).
         """
-        # TODO is there an sem_idx or similar? How to handle if more than 1 sem det -> for loop?
-        # get the SEM images corresponding to the last acquisition and average
-        data_sem = numpy.mean(self._acq_data[0][-n:], axis=0)
-        # TODO anything need to consider regarding type?
+        # check if we need to integrate images
+        if not n == 1:
+            # SEM
```
pieleric: You can drop this "SEM" comment, as it's just after.

src/odemis/acq/stream/_sync.py
```diff
@@ -1140,46 +1143,55 @@ def _runAcquisitionEbeam(self, future):
...
 
-        logging.debug("Integrate %s images", n)
-        orig_dtype = data2integrate[0].dtype
-        best_dtype = get_best_dtype_for_acc(orig_dtype, n)  # avoid saturation and overflow
-        data = numpy.sum(data2integrate, axis=0, dtype=best_dtype)
+                    if model.MD_BASELINE in md_optical:
+                        baseline = md_optical[model.MD_BASELINE]
+                        # Subtract baseline (aka black level) to avoid it from being multiplied
+                        # remove baseline from n-1 images, keep one baseline as bg level
+                        data = data - (n - 1) * baseline
```
pieleric: Yes... but maybe first double check that you're not going to get negative values, because they would underflow. So I'd suggest to use the same code as in`driver.spectrometer`.

src/odemis/acq/stream/_sync.py
```diff
@@ -1140,46 +1143,55 @@ def _runAcquisitionEbeam(self, future):
...
 
-        # Subtract baseline (aka black level) to avoid it from being multiplied
-        # TODO images do not have a baseline...
+                    # replace images, which needed integration, with integrated image
+                    self._acq_data[stream_idx][-n:] = [model.DataArray(data, md_optical)]
 
-        # append integrated image back to list of data
-        self._acq_data[self._ccd_idx].append(model.DataArray(data, md_optical))
+                else:
+                    # TODO should we raise a ValueError?
```
pieleric: Maybe just do as if it's `model.MD_DT_INTEGRATING`, because it's still more likely to be the right behaviour that completely failing. (ie, just put an else). If you want to be super careful, add a warning in such case.

src/odemis/acq/test/stream_test.py
```diff
@@ -3076,6 +3076,226 @@ def test_arpol_ss(self): # TODO move to SettingStreamTest cases?
...
+        md = ar_da[1].metadata
+        self.assertEqual(md[model.MD_EXP_TIME], ars.integrationTime.value)
+        self.assertIn(model.MD_INTEGRATION_COUNT, md)
+        # check that HW exp time * numberOfImages = integration time
+        self.assertEqual(self.ccd.exposureTime.value * md[model.MD_INTEGRATION_COUNT],
+                         ars.integrationTime.value)
+
+        # TODO need a check that images are actually summed up and have the correct values - how to do??
+        # -> check the i values increase by using different integration times and compare acq?
+        # TODO need to check overflow and baseline are handled correctly
```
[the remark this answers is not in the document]
Sabrina85: How can I add the MD_BASELINE to the md in the testcase, so it triggers the baseline substraction in `integrateImages`?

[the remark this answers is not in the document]
pieleric: You can do `streakccd.updateMetadata({MD_BASELINE: 100})`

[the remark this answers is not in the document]
Sabrina85: thanks!

src/odemis/acq/test/stream_test.py
```diff
@@ -2687,10 +2686,13 @@ def test_streak_acq_integrated_images(self):
...
         sems.emtDwellTime.value = 1e-06
 
+        # set a baseline, which does not effect data, but needed later to verify baseline is handled correctly
+        self.streak_ccd.updateMetadata({model.MD_BASELINE: 0})
+
         # set stream VAs
         streaks.integrationTime.value = 2  # s
         # TODO use fixed repetition value -> set ROI?
-        streaks.repetition.value = (3, 5)
+        streaks.repetition.value = (2, 2)
```
pieleric: Usually, I avoid the same repetition in X and Y in the test, as if for some reason X&Y get swapped, it's easier to detect. => 2x3?

Sabrina85: Agree. I had 2x3, but then this `self._roiToPhys` made it 2x2. I think, that is also why I put the TODO there at some stage. I can try another combination and see if the code keeps it, which is different in x,y.

src/odemis/acq/stream/_sync.py
```diff
@@ -240,17 +240,17 @@ def estimateAcquisitionTime(self):
...
         # Estimate time spent for the leeches
         for l in self.leeches:
+            shape = (len(pol_pos), rep[1], rep[0])
             # estimate acq time for leeches is based on two fastest axis
             if hasattr(self._sccd, "integrationTime"):
                 # get the exposure time directly from the hardware (e.g. hardware rounds value) to calc counts
                 integration_count = int(math.ceil(self._sccd.integrationTime.value / self._ccd.exposureTime.value))
                 if integration_count != self._sccd.integrationCounts.value:
-                    logging.debug("Integration count of %f, does not match integration count of %f as expected",
+                    logging.debug("Integration count of %d, does not match integration count of %f as expected",
```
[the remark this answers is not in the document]
Sabrina85: oje, I need holidays ... sorry

src/odemis/acq/stream/_sync.py
```diff
@@ -1148,50 +1147,60 @@ def _integrateImages(self, n):
...
+                orig_dtype = data2integrate[0].dtype
+                best_dtype = get_best_dtype_for_acc(orig_dtype, n)  # avoid saturation and overflow
+                data = numpy.sum(data2integrate, axis=0, dtype=best_dtype)
+
+                if model.MD_BASELINE in md_optical:
+                    baseline = md_optical[model.MD_BASELINE]
+                    # Subtract baseline (aka black level) to avoid it from being multiplied.
+                    # Remove baseline from n-1 images, keep one baseline as bg level.
+
+                    # If the baseline is too high compared to the actual black, we
```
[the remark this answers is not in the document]
Sabrina85: oh true. Changed it, but do the extra_bl calculation a bit differently (with the same result of course), as I think it is a bit better to read.

Sabrina85:
update 4 ready for review :)

Sabrina85:
Commits are squashed. Changed the max integration time to 24h (see sync and streams).

PR #1028, opened 2019-05-27
odemisd: terminate components in parallel and in the right order

philipwinklerdelmic:


src/odemis/odemisd/main.py
```diff
@@ -328,71 +329,98 @@ def _terminate_all_alive(self):
...
-                del parents[c]
-                for p in parents.values():
-                    p.discard(c)
-                alive.discard(c)
-                try:
-                    mic.alive.value = alive
-                except Exception:
-                    logging.warning("Failed to update the alive VA", exc_info=True)
+                                self._parents)
+                parent_less = (tuple(self._parents.keys())[0],)
```
pieleric: You could write slightly more concisely: `tuple(self._parents.keys())[:1]`

src/odemis/odemisd/main.py
```diff
@@ -328,71 +329,98 @@ def _terminate_all_alive(self):
...
+
+            for comp in parent_less:
+                terminating_comps.append(comp)
+                f = executor.submit(self._terminate_parentless, comp)
+                fs.add(f)
+
+            # Block until one future is completed, then continue looping
+            done, not_done = futures.wait(fs, return_when=futures.FIRST_COMPLETED)
```
pieleric: If you named `fs` as `fs_running`, you should be able to write directly (and understandably) as a single line:
`done, fs_running = futures.wait(fs_running, return_when=futures.FIRST_COMPLETED)`

src/odemis/odemisd/main.py
```diff
@@ -328,71 +329,98 @@ def _terminate_all_alive(self):
...
+            fs = not_done
+
+        futures.wait(fs, return_when=futures.ALL_COMPLETED)
+
+    def _terminate_parentless(self, comp):
+        # First terminate the children
+        children = comp.children.value or []
```
pieleric: The `or []` shouldn't be necessary. `.children` should always be a set.

src/odemis/odemisd/main.py
```diff
@@ -328,71 +329,98 @@ def _terminate_all_alive(self):
...
+            for comp in parent_less:
+                terminating_comps.append(comp)
+                f = executor.submit(self._terminate_parentless, comp)
+                fs.add(f)
+
+            # Block until one future is completed, then continue looping
+            done, not_done = futures.wait(fs, return_when=futures.FIRST_COMPLETED)
+            fs = not_done
+
+        futures.wait(fs, return_when=futures.ALL_COMPLETED)
```
pieleric: Let's add a log here, saying that it's done requesting the termination of all components, and just still waiting for N to be done.

src/odemis/odemisd/main.py
```diff
@@ -328,71 +329,98 @@ def _terminate_all_alive(self):
...
-        alive = set(mic.alive.value)
-
-        # create a "graph" child->parents
-        parents = {c: set() for c in alive}  # comp -> set of comps (all its parents)
-        for comp in alive:
-            # TODO: if a component was created by delegation, use its creator
-            # instead of itself as "used" by the other components, to ensure
-            # they might not be terminated too early.
+
+        # create a "graph" dependency->parents
```
pieleric: The "parent" term only matches for the children. Here we are talking of dependencies. So it should be something like "dependents".  So I'd suggest to rename the `_parents` variable as `_dependents`.
Moreover, this is not a graph (my fault). So I'd update the comment as something like:
`# alive component -> components which depends on it`

src/odemis/odemisd/main.py
```diff
@@ -328,71 +329,98 @@ def _terminate_all_alive(self):
         It also stops the containers, once no component is running in them.
```
pieleric: Please update the docstring. It shouldn't talk about parents, but "dependents". "Instantiated children" are now called "dependencies", and "Delegated children" are "children".

src/odemis/odemisd/main.py
```diff
@@ -328,71 +329,98 @@ def _terminate_all_alive(self):
...
+        for comp in self._parents:
             try:
-                for child in comp.children.value:
-                    # TODO: do not add the creator
-                    parents[child].add(comp)
+                for dep in comp.dependencies.value:
+                    # if a component was created by delegation, use its creator
+                    # instead of itself as "used" by the other components, to ensure
+                    # they might not be terminated too early.
+                    for c in self._parents:
```
pieleric: It should be possible to just use `dep.parent.value`, and if it's not None, it's the right parent.
(but double that it actually works, and maybe it's safer to not rely on this "back link" to work)

src/odemis/odemisd/main.py
```diff
@@ -328,71 +329,98 @@ def _terminate_all_alive(self):
...
+
+                    self._parents[dep].add(comp)
             except Exception:
                 logging.warning("Failed to find children of component %s when terminating",
                                 comp.name, exc_info=True)
 
-        # TODO: for the components created by delegation, either terminate them
-        # before their parents, or don't do it at all (as the parent normally
-        # is in charge of it).
+            for c in self._parents:
```
pieleric: I think it should be possible to remove the children in just one pass, after creating the complete "dependents". Loop through each component, and remove all its children from "dependents". You just need to be careful to not modify the set you loop through, so you need to do a copy:
```
for comp in list(self._dependents.keys()):
    for child in comp.children.value:
        self._dependents.pop(child, None)
```

src/odemis/odemisd/main.py
```diff
@@ -321,78 +322,95 @@ def _terminate_all_alive(self):
...
+        for comp in list(self._dependents.keys()):  # ._dependents will be modified, so we need a copy
+            for child in comp.children.value:
+                self._dependents.pop(child, None)
+        for comp in self._dependents:
             try:
-                for child in comp.children.value:
-                    # TODO: do not add the creator
-                    parents[child].add(comp)
+                deps = comp.dependencies.value
             except Exception:
```
philipwinklerdelmic: Do we really need the try-except here? If so, we should probably also do it in line 336 with `comp.children`.

src/odemis/odemisd/main.py
```diff
@@ -321,78 +322,95 @@ def _terminate_all_alive(self):
...
+            done, fs_running = futures.wait(fs_running, return_when=futures.FIRST_COMPLETED)
+
+        logging.debug("Finished requesting termination of all components. Waiting for %s components to terminate." % len(fs_running))
+        futures.wait(fs_running, return_when=futures.ALL_COMPLETED)
+
+    def _terminate_independent(self, comp):
+        # First terminate the children
+        children = comp.children.value
+        for child in children:
+            self._terminate_component(child)
```
philipwinklerdelmic: should we do this here explicitly or is it better to just let the parent deal with the children

philipwinklerdelmic:
I deleted the change to odemis-stop (3s->5s). Turns out it's not needed. Stopping in parallel also seems to pay off, all components are now stopped within 3s even for very long yaml files.

src/odemis/odemisd/main.py
pieleric: I think you could let the parent do it ... but you'd need to make sure to update .alive properly. So it might be easier to leave it as is and just say we explicitly terminate every component. Somehow, this keeps the rule simple, which is nice too.

philipwinklerdelmic:
I looked into the try-except thing and decided it's indeed necessary. I improved the code a bit, so it's clearer what's going on.

philipwinklerdelmic:
attributeerror -> exception

src/odemis/odemisd/main.py
```diff
@@ -321,78 +322,96 @@ def _terminate_all_alive(self):
...
+                # if a component was created by delegation, use its creator
+                # instead of itself as "used" by the other components, to ensure
+                # they might not be terminated too early.
                 try:
-                    # FIXME: don't try to terminate the children created by
-                    # delegation after terminating their parent (and container)
-                    # For now it just causes a warning message, but it'd be
-                    # better to not have the warning at all.
-                    c.terminate()
+                    d = dep.parent.value if dep.parent else dep
```
pieleric: `.parent` is not a VA. It's a (special) attribute. So it should be accessed just as `dep.parent`. I think it should be possible to write simply `d = dep.parent or dep`. 
That said, I'm wondering whether what the current code does. Would it just always raise an exception for components with parents? If so, you'd see many of the warnings in the log. Or maybe it returned a `Pyro4.RemoteMethod` in this case (instead of a component).

src/odemis/odemisd/main.py
```diff
@@ -321,78 +322,96 @@ def _terminate_all_alive(self):
...
-                del parents[c]
-                for p in parents.values():
-                    p.discard(c)
-                alive.discard(c)
-                try:
-                    mic.alive.value = alive
-                except Exception:
-                    logging.warning("Failed to update the alive VA", exc_info=True)
+                    # if the dependency has been terminated already, dep.parent will be None, so trying
+                    # to get the value will result in an AttributeError
```
pieleric: That comment doesn't sound correct. If `dep.parent` is None, it's evaluated as false, so the `.value` is never attempted.
Please check more what's going on. Also, you can mention that this code is mostly to handle components which died "early" due to accidents (eg, segfaults).

src/odemis/odemisd/main.py
```diff
@@ -321,78 +322,96 @@ def _terminate_all_alive(self):
...
+        del self._dependents[comp]  # children are already deleted from ._parents
+
+    def _terminate_component(self, c):
+        # remove from the graph
+        for p in self._dependents.values():
+            p.discard(c)
+
+        cname = c.name
+        logging.debug("Stopping comp %s", cname)
+        # TODO: update the .alive VA every time a component is stopped?
```
pieleric: It seems that this is now done, so drop this TODO :-)

src/odemis/odemisd/main.py
philipwinklerdelmic: You're right, I didn't test this enough. `dep.parent.value` always returns a `Pyro4.RemoteMethod`. I'll change it. Do we still need the try/except block? The only thing that could happen that I can think of is that if `dep` is destroyed, accessing `dep.parent` will raise an AttributeError, but I couldn't reproduce it.

philipwinklerdelmic: never mind, your next comment already answers my question :)

philipwinklerdelmic:
pushed changes

PR #1029, opened 2019-05-28
Way to force software binning for spectrometer

pieleric:


src/odemis/driver/spectrometer.py
```diff
@@ -295,16 +295,20 @@ def _newFrame(self, df, data):
...
                 # could end up subtracting too much, and values would underflow
                 # => be extra careful and never subtract more than min value.
-                extra_bl = min(int(data.min()), baseline_sum) - baseline
-                if extra_bl > 0:
-                    data -= extra_bl
-                else:
-                    logging.info("Baseline reported at %d, but lower values found, so not compensating",
-                                 baseline)
+                minv = float(data.min())
+                extra_bl = baseline_sum - baseline
```
Sabrina85: I don't get this part. So `baseline_sum` is just the baseline value times the shape of the data, so you get an 1dim array with values of baseline. Then you subtract from this array values again exactly this value, which would be 0 everywhere. What do I miss?

src/odemis/driver/spectrometer.py
```diff
@@ -147,6 +147,9 @@ def __init__(self, name, role, dependencies, **kwargs):
         self.binning.subscribe(self._onResBinning)
         self._updateWavelengthList()
 
+        # Indicate the data contains spectrum on the "fast" dimension
+        self._metadata[model.MD_DIMS] = "XC"
+
```
Sabrina85: order correct?

install/linux/usr/share/odemis/hwtest/sr193-andor3.odm.yaml
```diff
@@ -0,0 +1,128 @@
...
+    },
+}
+
+"Spectral Camera": {
+    class: andorcam3.AndorCam3,
+    role: ccd,
+    init: {
+           device: 0,  # if a real camera is plugged in, it's usually 0
+           bitflow_install_dirs: "/usr/share/bitflow/", # Put the path where Bitflow data is installed
+           #max_bin: [640, 16], # (px), to limit the maximum binning on X, Y
```
Sabrina85: Is it correct, that it is a comment? As you introduced the max_bin in the first commit here. Btw, is there a test case for testing the correct bg level subtraction?

pieleric: Yes, because for the testing in general, we don't want to use this limited binning. This limited binning is only used on some cameras, for certain systems.

pieleric: I've created a test case for the software binning (and baseline level subtraction)

src/odemis/driver/spectrometer.py
pieleric: No `baseline_sum` is just one value: the baseline value after binning (summing) the vertical axis. We want to reduce this "black level" back to the standard baseline => so extra_bl contains a single value, which is how much we should subtract to get back to the standard baseline.

src/odemis/driver/spectrometer.py
pieleric: Yes, it's actually of shape (1, C).

pieleric:
Just added a test case.

src/odemis/driver/test/spectrometer_test.py
```diff
@@ -466,6 +467,47 @@ def test_ccd_separated(self):
...
+
+        orig_res = self.spectrometer.resolution.value
+        self.spectrometer.data.subscribe(self.receive_spec_data)
+        try:
+            # Force the detector's vertical binning to be smaller
+            self.detector.binning.value = (binning[0], 1)
+
+            d = self._data.get()  # First image, with original FVB binning
+            self.assertEqual(d.shape, orig_res[::-1])
+            d = self._data.get()  # Second image, with the software binning
```
Sabrina85: How did the binning change?

src/odemis/driver/spectrometer.py
Sabrina85: Ah ok: So `extra_bl = (orig_shape[0] - 1) * baseline`.
And data is already binned?
I still don't get why you subtract  `(orig_shape[0] - 1) * baseline` instead of  `orig_shape[0] * baseline`.

install/linux/usr/share/odemis/hwtest/sr193-andor3.odm.yaml
Sabrina85: > Yes, because for the testing in general, we don't want to use this limited binning. This limited binning is only used on some cameras, for certain systems.

So maybe good to mention that here, so the next reader knows why it is commented code?

src/odemis/driver/test/spectrometer_test.py
pieleric: I've added a comment explaining this.

pieleric:
Just pushed new version of the commits with extra comments.

PR #1030, opened 2019-05-29
python3 migration: metaclass

philipwinklerdelmic:


philipwinklerdelmic:
remove inheritance from abc

PR #1035, opened 2019-06-03
model core: bugfix basestring import

philipwinklerdelmic:


pieleric:
Thank. I tried to write the code to be Python 3 compatible... and failed miserably ;-)

PR #1049, opened 2019-06-17
doc: Update the INSTALL.txt document

pieleric:


doc/INSTALL.txt
```diff
@@ -24,83 +23,29 @@ sudo add-apt-repository ppa:delmic-soft/odemis
...
-cd comedi_calibrate
-./autogen.sh
-./configure
-make -j
-sudo make install
-
-== Loading the comedi simulator driver ==
-Needs to be done after every boot.
+
+To Loading the comedi simulator driver, you can either type the following after every boot:
```
tepals: "To loading" does not make sense, should be To load

doc/INSTALL.txt
```diff
@@ -128,107 +73,21 @@ cd odemis
 python ./setup.py build
 sudo ./setup.py install --root=/ --install-scripts=/usr/local/bin/ --install-data=/usr/ --single-version-externally-managed
 
-To get the fluorophores database, you can run this command:
+To get the fluorophores database (for the SECOM/DELPHI), you can run this command:
 ./util/getfluodb.py
 Then copy all the fluodb directory to /usr/share/odemis/fluodb/
 
 = Installing the vendor device drivers =
```
tepals: The following paragraph now doesn't really say anything anymore about installing the vendor device drivers. Maybe add something like, see manufacturers for installation instructions.

pieleric:
Second commit with the suggested fixes.

PR #1054, opened 2019-06-19
Flipped mirror: new VA, update YAML file

VictoriaDel:


src/odemis/driver/test/static_test.py
```diff
@@ -69,6 +69,22 @@ def test_mag_choices(self):
             comp.magnification.value = 2.0
         comp.terminate()
 
+    def test_configurations(self):
+        configurations = {"Mirror up": {"pole_pos": (458, 519),  "focus_dist": 0.5e-3},
+                          "Mirror down": {"pole_pos": (634, 652),  "focus_dist": -0.5e-3}}
```
pieleric: Maybe add an extra configuration, with a non existing attribute (eg, "booo"), to test the handling (=> it shouldn't fail, but just log a warning).
PS: no need to check that it logged a warning, because it's actually really hard, and not so important behaviour-wise.

src/odemis/driver/test/static_test.py
```diff
@@ -69,6 +68,42 @@ def test_mag_choices(self):
...
+
+        comp.configuration.value = "Mirror up"
+        conf = configurations["Mirror up"]
+        self.assertEqual(comp.polePosition.value, conf["pole_pos"])
+        self.assertEqual(comp.focusDistance.value, conf["focus_dist"])
+
+        comp.terminate()
+
+    def test_badconfigurations(self):
+        configurations= {"conf missing": {"x_max": 43e-5, "focus_dist": 6e-3}}
```
pieleric: That's twice the same `configurations`, so the second one passes easily ;-)
I'd suggest to replace the first test with `configurations = {"conf unknown": {"booo": 43e-5, "focus_dist": 6e-3}}`

PR #1057, opened 2019-06-19
python3-migration-diverse-changes

philipwinklerdelmic:


src/odemis/acq/stream/__init__.py
```diff
@@ -40,7 +40,7 @@
 if sys.version_info >= (3, 4):
     ABC = abc.ABC
 else:
-    ABC = abc.ABCMeta('ABC', (object,), {'__slots__': ()})
+    ABC = abc.ABCMeta(str('ABC'), (object,), {'__slots__': ()})
```
pieleric: `'ABC'` == `str('ABC')`, right? So what's the difference?

src/odemis/gui/comp/text.py
```diff
@@ -275,7 +275,7 @@ def SetChoices(self, choices):
         self.dropdownlistbox.SetWindowStyleFlag(flags)
         if not isinstance(choices, list):
             self._choices = list(choices)
-        self._choices.sort(cmp=locale.strcoll)
+        self._choices.sort()
```
pieleric: That used to sort *case insensitive*. It's still useful, so to keep doing the same: `self._choices.sort(key=str.lower)`

src/odemis/model/test/remote_test.py
```diff
@@ -39,6 +38,10 @@
 import time
 import unittest
 
+if sys.version_info[0] >= 3:
+    from multiprocessing import Process
+else:
+    from multiprocessing.process import Process
```
pieleric: Actually, you can just write `from multiprocessing import Process`. That also works with Python 2.7.

src/odemis/acq/stream/__init__.py
philipwinklerdelmic: You're right, it doesn't change anything. In a different commit that I haven't pushed yet I import unicode_literals in every file in odemis. I'm not sure if that's a good idea. On the one hand it makes sense to be consistent across the entire codebase (we do need to import it in some files), but on the other hand it might also introduce a lot of unintended bugs.

What do you think?

If unicode_literals is imported in this file, we need to do this conversion to string here.

src/odemis/gui/comp/text.py
philipwinklerdelmic: It seems to me that str.lower and locale.strcoll do opposite things:
```
In [6]: x=['a','C',"B",'t']

In [7]: x.sort(cmp=locale.strcoll)

In [8]: x
Out[8]: ['B', 'C', 'a', 't']

In [9]: x.sort(key=str.lower)

In [10]: x
Out[10]: ['a', 'B', 'C', 't']
```

philipwinklerdelmic: that's in python2

philipwinklerdelmic: `x.sort()` in this case does the same as `x.sort(cmp=locale.strcoll)`

src/odemis/acq/stream/__init__.py
pieleric: Actually, originally, I had in mind that we would add an "import unicode_literals" to every file which we have tested both with Python 2 and 3... however, I hadn't foreseen such issue, where external code behaves differently with str and unicode (on Python 2). As shown here, using the unicode_literals forces  us to write ugly code, which wouldn't be necessary otherwise.   

So maybe we should adjust this idea, and say: if possible, use "import unicode_literals", but if that makes the code too ugly, just test on python 2 & 3, and add a comment just before the `import __future__` to indicate the code has been tested, but unicode_literals will break it.

Here, I'd suggest to drop this change.

src/odemis/gui/comp/text.py
pieleric: On my machine, `strcoll` is case-insensitive. Anyway, the original goal was to be case-insensitive, so let's use `str.lower`.

philipwinklerdelmic:
done

src/odemis/model/test/remote_test.py
pieleric: Please just convert that change to a single line. Or is there a reason this cannot be done?

philipwinklerdelmic: sorry, must have overlooked the comment, will pushed update in a minute

philipwinklerdelmic: done

PR #1059, opened 2019-06-19
python3 migration: unicode --> str

philipwinklerdelmic:


pieleric:
I *think* it's fine... but I have a feeling that it can have a lot of side effects, which will need a little bit of time to double check. So I'd rather wait until we branch Odemis 2.10 before merging this change (hopefully < 1 week).

philipwinklerdelmic:
Not ready to merge yet. We decided to not import `unicode_literals` in some files, so it doesn't mess up external functions. However, if we want to use `from builtins import str` we actually need the `unicode_literals` import. Needs some more consideration.

philipwinklerdelmic:
Actually, we don't need the `unicode_literals` import here, `from builtins import str` should be enough. Should I change it? This might decrease the chance of side effects.

pieleric:
Well, on the one hand, `unicode()` *has* to go away, to be compatible with Python 3. The best way is to use the "new" `str()`. I think this is kind of compulsory.

Concerning `unicode_literals`, it's independent of this. It just means that everywhere we wrote `""`, we meant `u""`. Maybe you can make two separate PR about this. In general, it should be possible to switch all the `gui` and  `cli` modules happily. The `util`, `odemisd`, and `acq` modules should also be fine. The `dataio` and `driver` modules will need a lot more care, and likely, quite a few strings will need to be annotated `b""`.

philipwinklerdelmic:
got rid of unicode_literals import for now.

src/odemis/dataio/tiff.py
```diff
@@ -1808,8 +1809,8 @@ def _genResizedShapes(data):
 
 
 def _ensure_fs_encoding(filename):
```
philipwinklerdelmic: that doesn't make any sense

philipwinklerdelmic: I don't think we need this. Libtiff seems to encode the filename properly internally, so we might as well pass unicode.

PR #1060, opened 2019-06-19
python3 migration: string_escape

philipwinklerdelmic:


pieleric:
I don't know if this makes much sense, as it is. In almost all these places, the "string" is actually containing bytes (that's why we encode it with escape when printing it). So on Python3, it'll either already be a bytes (eg, from serial, or from struct), or it should be bytes.  "Of course", as it's bytes, you cannot use `encode()`... you need to use `decode()`. On Python3, with bytes you need: `b.decode("ascii", "backslashreplace")`.

Too bad, it doesn't work on Python 2 (on a string).
However, it seems that `b.decode("latin1", "backslashreplace")` always works... but you want to double-check.

philipwinklerdelmic:
Indeed, the changes here don't make much sense. I will close this pr and make the appropriate changes when updating the drivers individually.

PR #1065, opened 2019-06-21
FLipped mirror - Angle resolved image

VictoriaDel:


src/odemis/util/test/angleres_test.py
```diff
@@ -90,6 +93,25 @@ def setUp(self):
         white_data_2500.metadata[model.MD_PIXEL_SIZE] = white_pxs_2500
         self.white_data_2500 = white_data_2500
 
+    def test_invertedMirror_AR2polar(self):
+        data_invMirror = self.data_invMir
```
pieleric: If you don't need the dimensions, you can just write `data_invMirror = ensure2DImage(self.data_invMir[0])`

src/odemis/util/angleres.py
```diff
@@ -26,7 +26,8 @@
 from scipy.interpolate import LinearNDInterpolator
 import numpy
 from odemis import model
-# import matplotlib.pyplot as plt
+import matplotlib.pyplot as plt
```
pieleric: Please, drop this change. I guess it was only for debugging.

src/odemis/util/test/angleres_test.py
VictoriaDel: I need the dimensions of the image so I didn't change this part.

src/odemis/util/angleres.py
```diff
@@ -161,6 +160,16 @@ def _FindAngle(x_array, y_array, pixel_size, parabola_f):
     return theta, phi, omega
 
 
+def _flipDataIfMirrorFlipped(data):
+    # invert data and adjust metadata for inverted mirror
```
pieleric: Convert this comment into a docstring.

src/odemis/util/angleres.py
```diff
@@ -161,6 +160,16 @@ def _FindAngle(x_array, y_array, pixel_size, parabola_f):
...
+    data = data[::-1, :]
```
pieleric: Please move the `if focus_distance ...` block here too.

src/odemis/util/angleres.py
```diff
@@ -161,6 +160,16 @@ def _FindAngle(x_array, y_array, pixel_size, parabola_f):
...
+    data.metadata = data.metadata.copy()
+    data.metadata[model.MD_AR_FOCUS_DISTANCE] *= -1  # invert the focus distance for inverted mirror
+    # put new y pole coordinate
+    arpole = data.metadata[model.MD_AR_POLE]
+    data.metadata[model.MD_AR_POLE] = (arpole[0], data.shape[0] - 1 - arpole[1])
+
```
pieleric: You need to *return* data. Otherwise, this has no effect. In general, anyway, it's nicer to have the input and output arguments decoupled.

src/odemis/util/test/angleres_test.py
pieleric: The proposed code replaces the 3 lines. Your code really doesn't look like it needs the Y,X values afterwards, so please double check.

src/odemis/util/angleres.py
VictoriaDel: Done. I also thought of it but it seems quite redundant to me. In case the condition is not true, the whole function _flipDataIfMirrorFlipped just checks the condition without any functionality.

src/odemis/util/test/angleres_test.py
VictoriaDel: in _flipDataIfMirrorFlipped
    data = data[::-1, :]
IndexError: too many indices for array

PR #1070, opened 2019-07-01
python3 migration: str.translate() -->str.replace()

philipwinklerdelmic:


pieleric:
Let's just use `str.replace()` in these cases. I think that would work too, right?

philipwinklerdelmic:
done

PR #1072, opened 2019-07-01
python3 migration: tiff strings

philipwinklerdelmic:


src/odemis/dataio/tiff.py
```diff
@@ -207,20 +208,20 @@ def _readTiffTag(tfile):
     # informative metadata
     val = tfile.GetField(T.TIFFTAG_PAGENAME)
     if val is not None:
-        md[model.MD_DESCRIPTION] = val
+        md[model.MD_DESCRIPTION] = val.decode('utf-8')
```
pieleric: For all the `decode()`, please add a "ignore" error handler, so that if the file metadata is a little bit unexpected, we don't entirely fail to read the file.

src/odemis/dataio/test/tiff_test.py
```diff
@@ -168,7 +169,7 @@ def testExportThumbnail(self):
 
         # check OME-TIFF metadata
         imo = libtiff.TIFF.open(FILENAME)
-        omemd = imo.GetField("ImageDescription")
+        omemd = imo.GetField("ImageDescription").decode('utf-8')
```
pieleric: Let's do the same as in the actual code, and keep this variable as bytes.

src/odemis/dataio/tiff.py
philipwinklerdelmic: Should I also add "ignore" to the encode statements or do we actually want it to fail in this case (e.g. because we don't support Chinese characters in the description)?

philipwinklerdelmic:
done

PR #1073, opened 2019-07-02
python3 migration: str in gui

philipwinklerdelmic:


src/odemis/gui/conf/file.py
```diff
@@ -138,10 +138,26 @@ def get(self, section, option):
...
+                return ret
         except (ConfigParser.NoOptionError, ConfigParser.NoSectionError):
             return self.default.get(section, option)
 
+    def _ensure_str_format(self, s):
+        """
+        The value argument of ConfigParser requires a unicode str in python3 and a byte str
+        in python2. This function makes sure a string is in the right format.
+        """
+        if sys.version_info[0] >= 3 and isinstance(s, bytes):
```
tepals: Just a readability comment no need to update. For readability I usually try to avoid having to many returns. So I would write this as:
```
if sys.version_info[0] >= 3 and isinstance(s, bytes):
    s = s.decode("utf-8")
elif sys.version_info[0] < 3 and isinstance(s, unicode):
    s = s.encode("utf-8")
return s
```

philipwinklerdelmic: done

PR #1075, opened 2019-07-03
python3 migration: tlfw driver

philipwinklerdelmic:


src/odemis/driver/tlfw.py
```diff
@@ -186,24 +187,25 @@ def _flushInput(self):
 
             # Shouldn't be necessary, but just in case
             skipped = self._serial.read(1000) # More than 1000 chars => give up
-            logging.debug("Skipping input %s", skipped.encode('string_escape'))
```
philipwinklerdelmic: Is it important to use string_escape here?

src/odemis/driver/tlfw.py
```diff
@@ -214,30 +216,30 @@ def _sendQuery(self, com):
...
             err = m.group(1)
             raise TLFWError("Device rejected command '%s': %s" % (com, err))
 
-        return line
+        return line.decode('utf-8')
 
     def _sendCommand(self, com):
         """
         Send a command which does not expect any answer
-        com (string): command to send (not including the ? and the \r)
```
philipwinklerdelmic: I decided to force the com argument to be bytes. Alternatively, we could also allow unicode and convert in this function. I thought it's a good idea to be more explicit. What do you think?

pieleric: Yes, I agree with you.
That said, I'd go even further, and would also return bytes. The function sends bytes, and returns bytes. Up the caller to do the right thing with bytes.
(In practice, that probably doesn't change much).

src/odemis/driver/tlfw.py
pieleric: Yes, kind-of. At the very least, you *cannot* use `.decode("utf-8")`. You can only do that if you know that the data is UTF-8 encoded. So far, I've never seen any serial device using such encoding. In general, they are all limited to ascii (with some exception for some Japanese SEMs which also support some old japanese encodings).

Here, we want to display the received data. When the protocol is really some "random" bytes (eg, the tmcm), we need a dedicated function to show any useful. As this protocol is "human readeable", the shortcut was to use `encode('string_escape')`. However, that's not really supported in Python 3. As discussed in PR #1060, it should be possible to use `.decode("latin1", "backslashreplace")`.

philipwinklerdelmic:
I pushed the new changes.

PR #1077, opened 2019-07-05
gui cont: move mirror type combox just below the mode selector buttons

pieleric:


pieleric:
Indeed, we could update the description of centering to mention this "Mirror type"... but this should be done only when the combo box is shown.
For now, this is shown in 0 systems, so let's not spent too much time on this.

PR #1080, opened 2019-07-05
driver focus tracker

tepals:
- driver that uses CANopen for communication with the focus tracker
- added class that simulates focus tracker communication through CANopen
- added tests
- added FocusTracker.eds file that that contains the object dictionary of the CANopen network
- added minimal backend yaml files


src/odemis/driver/test/focustracker_test.py
```diff
@@ -0,0 +1,187 @@
...
+        # SDOCommunicationError when we try to send a message to an object in the object dictionary.
+        with self.assertRaises(SdoCommunicationError):
+            FocusTrackerCO(name="Focus Tracker", role="focus_tracker", channel='can0', node_idx=0x12)
+
+    def test_object_dictionary(self):
+        """Verify that all relevant objects are in the object dictionary."""
+        object_names = ["AI Input PV", "CO Set Point W", "CO Proportional Band Xp1",
+                        "CO Integral Action Time Tn1", "CO Derivative Action Time Tv1"]
+        object_dictionary = self.focus_tracker.get_object_dictionary()
+        self.assertTrue(all(n in object_dictionary.names for n in object_names))
```
pieleric: I don't really think this test is necessary. That's all the internal business of the driver. As long as it behaves properly, it the actual names of the canOpen dict shoudn't matter.

tepals:
updated the PR

src/odemis/driver/focustracker.py
```diff
@@ -39,164 +43,162 @@ def __init__(self, name, role, channel, node_idx, **kwargs):
...
         if model.MD_GAIN_P in md:
-            md[model.MD_GAIN_P] = self.set_proportional(md[model.MD_GAIN_P])
+            md[model.MD_GAIN_P] = self._set_proportional(md[model.MD_GAIN_P])
         if model.MD_GAIN_I in md:
-            md[model.MD_GAIN_I] = self.set_integral(md[model.MD_GAIN_I])
+            md[model.MD_GAIN_I] = self._set_integral(md[model.MD_GAIN_I])
         if model.MD_GAIN_D in md:
-            md[model.MD_GAIN_D] = self.set_derivative(md[model.MD_GAIN_D])
+            md[model.MD_GAIN_D] = self._set_derivative(md[model.MD_GAIN_D])
         super(FocusTrackerCO, self).updateMetadata(md)
```
pieleric: A little annoying corner case is when the user request a new set of PID values, and one of the value is rejected. For example, let's say that D is a negative value. Then, P and I will be set, then updateMetadata will raise an error. The P&I will be updated on the device, but the metadata left as-is.

Maybe the simplest way to deal with it would be to immediately update the metadata after setting each gain.

src/odemis/driver/focustracker.py
```diff
@@ -39,164 +43,162 @@ def __init__(self, name, role, channel, node_idx, **kwargs):
...
+                raise
         self.network.check()
-        self.node = canopen.RemoteNode(node_idx, os.path.dirname(odemis.__file__) + "/driver/FocusTracker.eds")
+        object_dictionary = pkg_resources.resource_filename("odemis.driver", "FocusTracker.eds")
+        self.node = canopen.RemoteNode(node_idx, object_dictionary)
         self.network.add_node(self.node)
+        if channel == 'fake':
+            # overwrite the node so we can get the fake sdo's
+            self.node = FakeRemoteFocusTracker(node_idx, object_dictionary)
+        self._hwVersion = self.node.sdo["Identity Object"]["Revision number"].raw
```
pieleric: Does this work with the simulator? It has no such SDO defined.

src/odemis/driver/focustracker.py
```diff
@@ -39,164 +43,162 @@ def __init__(self, name, role, channel, node_idx, **kwargs):
...
+            else:
+                raise
         self.network.check()
-        self.node = canopen.RemoteNode(node_idx, os.path.dirname(odemis.__file__) + "/driver/FocusTracker.eds")
+        object_dictionary = pkg_resources.resource_filename("odemis.driver", "FocusTracker.eds")
+        self.node = canopen.RemoteNode(node_idx, object_dictionary)
         self.network.add_node(self.node)
+        if channel == 'fake':
+            # overwrite the node so we can get the fake sdo's
+            self.node = FakeRemoteFocusTracker(node_idx, object_dictionary)
```
pieleric: Do we need to create a "real" node  when the channel is "fake"? It's odd that you create a non-existent node on a virtual network, and then create another node. Couldn't use the same node all the time?

src/odemis/driver/focustracker.py
```diff
@@ -0,0 +1,215 @@
...
+        channel (str): channel name of can bus
+        node_idx (int): node index of focus tracker
+        """
+        model.HwComponent.__init__(self, name, role, **kwargs)
+        # Connect to the CANbus and the CANopen network.
+        self.network = canopen.Network()
+        self.network.connect(bustype='socketcan', channel=channel)
+        self.network.check()
+        self.node = canopen.RemoteNode(node_idx, os.path.dirname(odemis.__file__) + "/driver/FocusTracker.eds")
+        self.network.add_node(self.node)
```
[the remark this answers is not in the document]
pieleric: In `swVersion`, you could indicate the python-canopen/python-can versions. 
If you have some way to get information on the USB can adapter, it would be nice. Maybe `self.network.bus.channel_info` reports something non-boring?

src/odemis/driver/focustracker.py
tepals: It does not

src/odemis/driver/focustracker.py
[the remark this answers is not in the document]
tepals: `self.network.bus.channel_info`  reports `'socketcan channel \'can0\''` so nothing interesting, I also can't find anything interesting. For now I've indicated the python-canopen/python-can versions in `swVersion`.

src/odemis/driver/focustracker.py
```diff
@@ -19,13 +19,17 @@
...
+import numbers
+import pkg_resources
 
 import canopen  # TODO package canopen and add to project requirements.
 
-import odemis
 from odemis import model
+from odemis.model import HwError
+
+TARGET_POSITION_RANGE = (0, 100)
```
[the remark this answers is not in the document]
tepals: I asked Wouter before my vacation and he was relatively sure that at least it should not be negative/does not make sense to be negative

src/odemis/driver/focustracker.py
```diff
@@ -39,164 +43,162 @@ def __init__(self, name, role, channel, node_idx, **kwargs):
...
 
-    def get_target_pos(self):
+    def _set_target_pos(self, value):
         """
+        value (float, int): The target position in meters.
         return (float): The target position of the laser on the linear ccd.
         """
-        return self._target_pos_sdo.raw * 1e-6
+        self._target_pos_sdo.raw = value * 1e6
+        return value
```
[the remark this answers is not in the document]
tepals: On the hardware it is possible to set it to negative values and whatever range. An `IndexError` is raised when you try to set is with a value out of range, because that's what the `FloatContinuous` VA does

tepals:
updated

src/odemis/driver/focustracker.py
```diff
@@ -39,171 +44,177 @@ def __init__(self, name, role, channel, node_idx, **kwargs):
...
         self.network.check()
-        self.node = canopen.RemoteNode(node_idx, os.path.dirname(odemis.__file__) + "/driver/FocusTracker.eds")
+        object_dict = pkg_resources.resource_filename("odemis.driver", "FocusTracker.eds")
+        if channel == 'fake':
+            self.node = FakeRemoteNode(node_idx, object_dict)
+        else:
+            self.node = canopen.RemoteNode(node_idx, object_dict)
         self.network.add_node(self.node)
+        self._swVersion = "python-canopen v%s , python-can v%s" % (canopen.__version__, can.__version__)
+        self._hwVersion = self.node.sdo["Identity Object"]["Revision number"].raw
```
pieleric: Revision number is an int, but hwVersion should be a string. Also, i wonder if there is some official encoding of the revision number, because it's currently set to 65537, which seems an awful lot... maybe it's "1.1"?

src/odemis/driver/test/focustracker_test.py
```diff
@@ -26,162 +26,82 @@
...
+        self.focus_tracker = FocusTrackerCO(**KWARGS)
+        self.kwargs = KWARGS
 
     def test_connection(self):
         """Test that connection to the network is successful when connected, and unsuccessful when disconnected."""
-        # If the network scanner does not see a node at index 0x10, the focus tracker is not connected.
-        self.assertTrue(0x10 in self.focus_tracker.get_available_nodes())
+        if TEST_NOHW:
+            # If the network scanner does not see a node at index 0x10, the focus tracker is not connected.
+            self.assertFalse(0x10 in self.focus_tracker.network.scanner.nodes)
```
pieleric: I don't think it's a really good proof that the code is broken if the simulator reports that 0x10 is present. I'd suggest to just not test (so the simulator may or may not report the presence of 0x10).

src/odemis/driver/focustracker.py
tepals: 65537 is 0x10001 so I guess it's similar to 1.1, I'll set hwVersion as a hex string; `hwVersion = '0x10001'`

tepals:
updated

src/odemis/driver/focustracker.py
```diff
@@ -39,171 +44,180 @@ def __init__(self, name, role, channel, node_idx, **kwargs):
...
+        if channel == 'fake':
+            self.node = FakeRemoteNode(node_idx, object_dict)
+        else:
+            self.node = canopen.RemoteNode(node_idx, object_dict)
         self.network.add_node(self.node)
+        self._swVersion = "python-canopen v%s , python-can v%s" % (canopen.__version__, can.__version__)
+        rev_num = hex(self.node.sdo["Identity Object"]["Revision number"].raw).split('x')[-1]
+        major = int(rev_num[:-4])
+        minor = int(rev_num[-4:])
+        self._hwVersion = "{}.{}".format(major, minor)
```
pieleric: That's not going to work well if there is any number above 9 (as 10 is expressed as "a"). You could use `int(...,16)`. However, you can also directly convert to the right number with such code:
```
rev_num = self.node.sdo["Identity Object"]["Revision number"].raw
minor = rev_num & 0xffff
major = (rev_num >> 16) & 0xffff
```

PR #1081, opened 2019-07-08
secom-cl plugin:

Sabrina85:


Sabrina85:
Please mainly review on the methods `....2`, which might be directly be modified in _sync in Odemis. Or if there is any other clever way to make the code nice. Doc strings are still incomplete.

plugins/secom_cl.py
```diff
@@ -14,54 +15,69 @@
...
 
     return md
 
+
 def get_ccd_pxs(ccd):
     """
-    Returns the (theoretical) pixelsize of the CCD (projected on the sample).
+    Calculates the (theoretical) pixel size of the optical detector (projected on the sample).
+    :param ccd: (model.Component????) The optical detector.
+    :returns: () The theoretical pixel size of the optical detector.  TODO @Eric isn't that the effective px size?
```
pieleric: I think that "theoretical" here meant that PIXEL_SIZE_COR is not taken into account. You could drop the word, and make the sentence less arguable.

plugins/secom_cl.py
```diff
@@ -72,54 +88,59 @@ def get_ccd_pxs(ccd):
 
     return pxs
 
+
 def get_ccd_fov(ccd):
     """
-    Returns the (theoretical) field of view of the CCD.
-    returns (tuple of 4 floats): position in physical coordinates m (l, t, b, r)
+    Calculates the (theoretical) field of view of the optical detector.
+    :param ccd: (model.Component????) The optical detector.
```
pieleric: if you want to be precise: "Component???" -> DigitalCamera

plugins/secom_cl.py
```diff
@@ -196,322 +221,463 @@ def sem_roi_to_ccd(escan, ccd, roi, margin=0):
...
+class SECOMCLSettingsStream(acqstream.CCDSettingsStream):
+    """
+    A cl settings stream, for a set of points (on the SEM).
+    The live view is just the raw CCD image.
+    """
+    def __init__(self, name, detector, dataflow, emitter, **kwargs):
+        if "acq_type" not in kwargs:
+            kwargs["acq_type"] = model.MD_AT_CL
+
+        super(SECOMCLSettingsStream, self).__init__(name, detector, dataflow, emitter, **kwargs)
```
pieleric: If I understand correctly, you cannot override the original methods because the `__init__` would fail?
If so, don't call the `__init__` of RepetitionStream, and instead, just copy all the code (needed). In practice, this is going to be barely more than you have now. Note, you still need to call the init of the LiveStream, and you can do it with `LiveStream.__init__(self, ....)`

plugins/secom_cl.py
```diff
@@ -196,322 +221,463 @@ def sem_roi_to_ccd(escan, ccd, roi, margin=0):
...
+        roi, rep, pxs = self._updateROIAndPixelSize2(roi, pxs)
+        # update roi and pixel size without going through the checks
+        self.roi._value = roi
+        self.roi.notify(roi)
+        self.pixelSize._value = pxs
+        self.pixelSize.notify(pxs)
+
+        return rep
+
+    # TODO: instead of caring about the pixel size in m, just use the scale, which
```
pieleric: Drop that TODO, that's only interesting for the RepetitionStream class.

plugins/secom_cl.py
```diff
@@ -196,322 +221,463 @@ def sem_roi_to_ccd(escan, ccd, roi, margin=0):
...
+        roi_size = (rep[0] * pxs[0] / phy_size[0],
+                    rep[1] * pxs[1] / phy_size[1])
+        roi = (roi_center[0] - roi_size[0] / 2,
+               roi_center[1] - roi_size[1] / 2,
+               roi_center[0] + roi_size[0] / 2,
+               roi_center[1] + roi_size[1] / 2)
+
+        return roi
+
+    def _fitROI2(self, roi):
```
pieleric: I don't think you need to duplicate that one, do you?

plugins/secom_cl.py
```diff
@@ -196,322 +221,463 @@ def sem_roi_to_ccd(escan, ccd, roi, margin=0):
...
+        Computes the pixel size (based on the repetition, roi and FoV of the
+        e-beam). The RepetitionStream does provide a .pixelSize VA, which
+        should contain the same value, but that VA is for use by the GUI.
+        :returns: (float) The pixel size in m.
+        """
+        epxs = self._emitter.pixelSize.value
+        rep = self.repetition.value
+        roi = self.roi.value
+        eshape = self._emitter.shape
+        phy_size_x = (roi[2] - roi[0]) * epxs[0] * eshape[0]  # one dim is enough
```
pieleric: Remove the "one dim is enough" comment, as you are using 2 dims.

plugins/secom_cl.py
```diff
@@ -196,322 +221,463 @@ def sem_roi_to_ccd(escan, ccd, roi, margin=0):
...
-    def acquire_ccd(self, x, y, ccd_roi_idx):
-        """
-        Acquire an image from the CCD while having the e-beam at a spot position
-        x, y (floats): spot position in the ebeam coordinates
-        ccd_roi_idx: slice to crop the CCD image
-        return (model.DataArray of shape (Y,X), model.DataArray of shape (1,1)):
-          the CCD image and the SEM data (at the spot)
-        """
-        self.move_spot(x, y)
+        fn = self.filename.value
```
pieleric: Just before this part of the code, you need to put a BIG comment explaining this gigantic hack.

plugins/secom_cl.py
```diff
@@ -196,322 +221,463 @@ def sem_roi_to_ccd(escan, ccd, roi, margin=0):
...
-        self._ccd_data_received.clear()
-        self.ccd.softwareTrigger.notify()
+        self.save_data(data,
+                       prefix=fn_prefix,
+                       xres=self.repetition.value[0],
+                       yres=self.repetition.value[1],
+                       xstepsize=self._getPixelSize()[0] * 1e9,
+                       ystepsize=self._getPixelSize()[1] * 1e9,
+                       xpos=i[1]+1,  # start counting with 1
+                       ypos=i[0]+1
```
pieleric: Maybe just pass "type" as "optical" from here too.

plugins/secom_cl.py
```diff
@@ -196,322 +221,463 @@ def sem_roi_to_ccd(escan, ccd, roi, margin=0):
...
+        if n != self._ccd_idx:
+            return super(SECOMCLSEMMDStream, self)._onCompletedData(n, raw_das)
 
-        if not self._sem_data_received.wait(3):
-            logging.warning("No SEM data received, 3s after the CCD data")
-        if len(self._sem_data) > 1:
-            logging.warning("Received %d SEM data, while expected just 1", len(self._sem_data))
+        memu = acqstream.udriver.readMemoryUsage()
+        # Too bad, need to use VmSize to get any good value
+        logging.info("Memory usage is %d. ", memu)
```
pieleric: This memory check can go away now that we've confirmed the memory doesn't go crazy.

src/odemis/acq/stream/_sync.py
```diff
@@ -576,8 +576,12 @@ def _get_center_pxs(self, rep, sub_shape, datatl):
         # Note: we don't rely on the MD_PIXEL_SIZE, because if the e-beam was in
         # spot mode (res 1x1), the scale is not always correct, which gives an
         # incorrect metadata.
-        pxsx = self._getPixelSize()
-        pxs = pxsx / sub_shape[0], pxsx / sub_shape[1]
+        pxs = self._getPixelSize()
+        if not isinstance(pxs, tuple):
+            pxs = pxs, pxs
+
+        pxs = pxs[0] / sub_shape[0], pxs[1] / sub_shape[1]
```
pieleric: Let's just update `_getPixelSize()` to always return a tuple, and adjust "all" the callers. You've found 2 of them, but there is a third one in the file.

plugins/secom_cl.py
```diff
@@ -651,71 +923,108 @@ def start(self):
...
-
-        dlg.Close()
+    def save_hw_settings(self):
+        """
+        TODO
+        """
+        res = self.secom_cl_stream._emitter.resolution.value
+        scale = self.secom_cl_stream._emitter.scale.value
+        trans = self.secom_cl_stream._emitter.translation.value
+        dt = self.secom_cl_stream._emitter.dwellTime.value
```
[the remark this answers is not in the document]
Sabrina85: Pycharm is of the same opinion ;)

plugins/secom_cl.py
Sabrina85: Yes I think we need it, because `roi, rep, pxs = self._updateROIAndPixelSize2(roi, pxs)` is called somewhere in this method.

Sabrina85:
Something is off regarding the estimated acq time. After the first acq (aka opening again the plugin) it is hours...same for closing GUI only and reopen.

src/odemis/acq/stream/_sync.py
```diff
@@ -851,22 +850,25 @@ def _adjustHardwareSettings(self):
...
-            ts = max(1, min(max_tile_shape_dt, max_tile_shape_scale))
-            tile_shape = (ts, ts)
-            subpxs = pxs / ts
-            dt = (exp / numpy.prod(tile_shape)) / 2
-            scale = (subpxs / sem_pxs[0], subpxs / sem_pxs[1])
-
-            # Double check fuzzing would work (and make sense)
-            if ts == 1 or not (rng[0] <= dt <= rng[1]) or scale[0] < 1 or scale[1] < 1:
-                logging.info("Disabled fuzzing because SEM wouldn't support it")
+            if pxs[0] != pxs[1]:  # TODO: support fuzzing for rectangular pxs
```
pieleric: Please, in this case, just log a warning that it's not supported, and continue with the usual code. (I'm particularly worried with floating point errors).

PR #1082, opened 2019-07-08
python3 migration: nfpm driver

philipwinklerdelmic:
* send bytes to driver
* attributes should be unicode strings
* copy .items into list instead of using iterator because elements can be deleted inside loop
* use correct python interpreter

### DIFF
--- a/src/odemis/driver/nfpm.py
+++ b/src/odemis/driver/nfpm.py
@@ -40,6 +40,7 @@ from subprocess import CalledProcessError
 import subprocess
 import threading
 import time
+import sys
 
 
 class NewFocusError(Exception):
@@ -159,10 +160,10 @@ class PM8742(model.Actuator):
             self._accesser.terminate()
             self._accesser = None
 
-    def _sendOrderCommand(self, cmd, val="", axis=None):
+    def _sendOrderCommand(self, cmd, val=b"", axis=None):
         return self._accesser.sendOrderCommand(cmd, val, axis)
 
-    def _sendQueryCommand(self, cmd, val="", axis=None):
+    def _sendQueryCommand(self, cmd, val=b"", axis=None):
         """
         Same as accesser's sendQueryCommand, but with error recovery
         """
@@ -203,15 +204,15 @@ class PM8742(model.Actuator):
              Firmware version (and date)
              serial number
         """
-        resp = self._sendQueryCommand("*IDN")
-        # expects something like this:
-        # New_Focus 8742 v2.2 08/01/13 11511
+        resp = self._sendQueryCommand(b"*IDN")
         try:
-            m = re.match("\w+ (?P<model>\w+) (?P<fw>v\S+ \S+) (?P<sn>\d+)", resp)
+            resp_str = resp.decode("ascii")
+            # expects something like this:
+            # New_Focus 8742 v2.2 08/01/13 11511
+            m = re.match("\w+ (?P<model>\w+) (?P<fw>v\S+ \S+) (?P<sn>\d+)", resp_str)
             modl, fw, sn = m.groups()
         except Exception:
-            raise IOError("Failed to decode firmware answer '%s'" %
-                          resp.encode('string_escape'))
+            raise IOError("Failed to decode firmware answer '%s'" % resp.decode("latin1", "backslashreplace"))
 
         return modl, fw, sn
 
@@ -223,7 +224,7 @@ class PM8742(model.Actuator):
         axis (1<=int<=4): axis number
         return (0<=int<=3): the motor type
         """
-        resp = self._sendQueryCommand("QM", axis=axis)
+        resp = self._sendQueryCommand(b"QM", axis=axis)
         return int(resp)
 
     def GetVelocity(self, axis):
@@ -232,7 +233,7 @@ class PM8742(model.Actuator):
         axis (1<=int<=4): axis number
         return (0<=int<=2000): the speed in step/s
         """
-        resp = self._sendQueryCommand("VA", axis=axis)
+        resp = self._sendQueryCommand(b"VA", axis=axis)
         return int(resp)
 
     def SetVelocity(self, axis, val):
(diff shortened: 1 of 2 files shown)


src/odemis/driver/npmc.py
```diff
@@ -710,7 +710,7 @@ def _updateCurrentPosition(self):
             self._pos = cur_position
 
     def _sendAnswer(self, ans):
-        self._output_buf += "%s\r" % (ans,)
+        self._output_buf += ("%s\r" % (ans,)).encode()
```
pieleric: I don't think `.encode()` should be necessary here. It receives bytes (or should), and append them to bytes.

src/odemis/driver/npmc.py
```diff
@@ -795,26 +795,26 @@ def _parseMessage(self, msg):
...
+        elif re.match(br'\dPA', msg):
+            axis = int(msg[:1])
             new_pos = float(msg[3:])
             self._doMove(axis - 1, new_pos)
 
         # Move to a relative position
-        elif re.match(r'\dPR', msg):
-            axis = int(msg[0])
+        elif re.match(br'\dPR', msg):
+            axis = int(msg[:1])
```
pieleric: Good catch! We have to be super careful with this in the conversion to bytes.

src/odemis/driver/npmc.py
philipwinklerdelmic: I sometimes passed ints and floats to the function, which is why I needed the encode, but, indeed, it's better to only send bytes. I made the changes and created a separate pull request. I will delete the commit from this PR, so the 2 drivers are separate.

src/odemis/driver/nfpm.py
```diff
@@ -695,7 +696,11 @@ def _scanOverIP():
         # Run the separate program via authbind
         try:
             exc = os.path.join(os.path.dirname(__file__), "nfpm_netscan.py")
-            out = subprocess.check_output(["authbind", "python", exc])
+            if sys.version_info[0] >= 3:
+                py_interpreter = "python3"
+            else:
+                py_interpreter = "python"
+            out = subprocess.check_output(["authbind", py_interpreter, exc])
```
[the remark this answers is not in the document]
philipwinklerdelmic: done

src/odemis/driver/nfpm.py
```diff
@@ -203,15 +205,14 @@ def GetIdentification(self):
              Firmware version (and date)
              serial number
         """
-        resp = self._sendQueryCommand("*IDN")
+        resp = self._sendQueryCommand(b"*IDN").decode("latin1", "backslashreplace")
```
pieleric: You shouldn't use "backslashreplace" for normal bytes->string conversion. This is only used for pretty-printing bytes. In order to actually interpret "resp", you should use `.decode("ascii")`. Move it into the try/except block, so that if there are weird characters, it fails with IOError (and store the decoded string into a separate variable, for instance `resp_str`)

src/odemis/driver/nfpm.py
```diff
@@ -344,14 +345,14 @@ def GetError(self):
...
         # Note: there is another one "TE" which only returns the number, and so
         # is faster, but then there is no way to get the message
-        resp = self._sendQueryCommand("TB")
+        resp = self._sendQueryCommand(b"TB")
         # returns something like "108, MOTOR NOT CONNECTED"
         try:
-            m = re.match("(?P<no>\d+), (?P<msg>.+)", resp)
-            no, msg = int(m.group("no")), m.group("msg")
+            m = re.match(b"(?P<no>\d+), (?P<msg>.+)", resp)
+            no, msg = int(m.group("no")), m.group("msg")  # group takes unicode str even on byte str input
```
pieleric: "msg" should be converted to string (from bytes), using `.decode("ascii")`.

src/odemis/driver/nfpm_netscan.py
```diff
@@ -48,16 +48,19 @@ def scan():
...
-                        end_hn = data.index("\x00", 19)
+                        end_hn = data.index(b"\x00", 19)
                         hn = data[19:end_hn]
-                        print("%s\t%s\t%d" % (hn, addr, p))
+                        print(b"%s\t%s\t%d" % (hn, addr, p))
                     except Exception:
-                        logging.exception("Failed to decode packet %r from %s", data, addr)
+                        logging.exception("Failed to decode packet %r from %s",
+                                          data.decode("latin1", "backslashreplace"),
+                                          addr.decode("latin1", "backslashreplace"))
```
pieleric: I think that `addr`, as returned by `recvfrom()` is a string (also on Python 3). So there shouldn't be any `.decode()` needed... but you might want to double check (assuming that it's using IPv4, aka IF_INET.).

src/odemis/driver/nfpm.py
```diff
@@ -91,8 +92,9 @@ def __init__(self, name, role, address, axes, stepsize, sn=None, **kwargs):
                 raise ValueError("stepsize should be in meter, but got %g" % (sz,))
         self._stepsize = stepsize
 
-        self._address = address
-        self._sn = sn
+        # Convert python2 byte str to unicode str
+        self._address = address.decode() if isinstance(address, bytes) else address
+        self._sn = sn.decode() if isinstance(sn, bytes) else sn
```
pieleric: Why would the user pass these parameters as bytes? I think you can just assume these are strings. Or are there specific cases you want to handle?

philipwinklerdelmic: I thought it would be best if the class also worked with python2 strings (=bytes) as arguments. Is this unnecessary?

pieleric: On Python2, the user passes a string, and it should stay a string, to be used later as an IP address, and as comparison. It works fine for now, no need to make the code more complex by converting the strings to unicode.

philipwinklerdelmic: ok

philipwinklerdelmic:
done

philipwinklerdelmic:
all good now, ready to merge

PR #1086, opened 2019-07-09
python3 migration: tlfw driver

philipwinklerdelmic:


philipwinklerdelmic:
I worked on this twice without noticing... The PR which is already merged is slightly better (returns a unicode string in `GetIdentification`), so I will close this one.

PR #1089, opened 2019-07-09
python3 migration: tlaptmf driver

philipwinklerdelmic:


src/odemis/driver/tlaptmf.py
```diff
@@ -416,11 +419,11 @@ def GetInfo(self):
...
 
         # remove trailing 0's
-        modl = modl.rstrip("\x00")
-        notes = notes.rstrip("\x00")
+        modl = modl.rstrip(b"\x00")
+        notes = notes.rstrip(b"\x00")
 
         # Convert firmware version to a string
-        fmvs = "%d.%d.%d" % ((fmv & 0xff0000) >> 16,
+        fmvs = b"%d.%d.%d" % ((fmv & 0xff0000) >> 16,
```
pieleric: No, please convert to a *str*, as it'll be shown to the user. 
You need to also make sure that the other str are *str* (and not bytes, on Python 3). That's mostly important for the `modl`, as it's used after.

src/odemis/driver/tlaptmf.py
```diff
@@ -303,7 +303,10 @@ def SendMessage(self, msg, dest=0x50, src=1, p1=None, p2=None, data=None):
             while True:
                 trials += 1
                 try:
-                    logging.debug("Sending: '%s'", ", ".join("%02X" % ord(c) for c in com))
+                    # Iterate through com.decode('latin1', 'backslashreplace') to allow compatibility with python2 and python3.
+                    # Iteration through a byte str in python3 automatically returns the code point,
+                    # while python2 requires explicit conversion with ord().
+                    logging.debug("Sending: '%s'", ", ".join("%02X" % ord(c) for c in com.decode('latin1', 'backslashreplace')))
```
pieleric: With "weird" characters, it'll go wrong, as it'll show each value of the escaped string. I guess you could just decode to latin1 without special handler, or, prettier, convert to a `bytearray`, like:
`"%02X" % ord(c) for c in bytearray(com)`

You could also make a dedicated function, which behaves differently depending on the python version. Something like `_msg_to_str()`. You'd write 2 separate functions `_msg_to_str_py2()` and `_msg_to_str_py3()`, and assign one of them at init, based on the Python version.

src/odemis/driver/tlaptmf.py
```diff
@@ -374,27 +377,27 @@ def _ReadMessage(self, timeout=None):
...
+                    raise IOError("Controller timed out, after receiving '%s'" % msg.decode('latin1', 'backslashreplace'))
 
                 msg += char
         finally:
             self._serial.timeout = old_timeout
 
         mid = struct.unpack("<H", msg[0:2])[0]
-        if not (ord(msg[4]) & 0x80): # short message
-            logging.debug("Received: '%s'", ", ".join("%02X" % ord(c) for c in msg))
+        if not (ord(msg.decode('latin1', 'backslashreplace')[4]) & 0x80): # short message
```
pieleric: Use [4:5], instead of the `decode()`, or convert `msg` to a bytearray. (Also below, in other similar patterns).

src/odemis/driver/tlaptmf.py
philipwinklerdelmic: bytearray requires a second argument for the encoding in python3, so it's quite similar to .decode('latin1'). I think I'm going to stick with this and remove the backslashreplace handler

philipwinklerdelmic: actually, I think bytearray wouldn't even work in python3 because com is already bytes.

philipwinklerdelmic:
pushed changes

src/odemis/driver/tlaptmf.py
pieleric: Sorry, actually I meant. `"%02X" % c for c in bytearray(com)`.
It does work on Python 2 and Python 3 (I tried!). com is a string on python2 and bytes on python3. In both cases, this creates the right bytearray, which is immediately iterated as ints.
Please update all the logging which use `ord()` .

pieleric:
This requires `to_str_escape()` update.

philipwinklerdelmic:
updated

src/odemis/driver/tlaptmf.py
```diff
@@ -374,27 +374,27 @@ def _ReadMessage(self, timeout=None):
...
 
                 msg += char
         finally:
             self._serial.timeout = old_timeout
 
         mid = struct.unpack("<H", msg[0:2])[0]
-        if not (ord(msg[4]) & 0x80): # short message
-            logging.debug("Received: '%s'", ", ".join("%02X" % ord(c) for c in msg))
+        if not (ord(msg[4:5]) & 0x80): # short message
+            logging.debug("Received: '%s'", ", ".join("%02X" % ord(c) for c in to_str_escape(msg)))
```
pieleric: All these `join()`s are for a different output. So you need to write this way:
`", ".join("%02X" % c for c in bytearray(msg))`

You could consider putting it in a separate function, like `to_str_escape()` but that's not really needed I think.

src/odemis/driver/tlaptmf.py
```diff
@@ -415,6 +415,11 @@ def GetInfo(self):
         values = struct.unpack('<I8sHI48s12xHHH', res)
         sn, modl, typ, fmv, notes, hwv, state, nc = values
 
+        if sys.version_info[0] >= 3:
+            # convert bytes to str
+            modl = modl.decode('latin1')
+            notes = notes.decode('latin1')
+
```
pieleric: Don't worry about doing it all the time. I'd suggest to do it immediately after the rstrip(), like that:
`modl = modl.rstrip("\x00").decode('latin1')`

src/odemis/driver/tlaptmf.py
```diff
@@ -805,9 +810,9 @@ def _processMessage(self, msg):
                 self._end_motion = time.time() + 1 # 1s move
                 # no output
             else:
-                logging.warning("Message '%X' unknown", mid)
+                logging.warning("Message '%X' unknown", to_str_escape(mid))
         except Exception:
-            logging.exception("Simulator failed on message %X", mid)
+            logging.exception("Simulator failed on message %X", to_str_escape(mid))
```
pieleric: These 2 last changes are wrong. `%X` expects an int, which `mid` is. So no string escaping needed.

PR #1091, opened 2019-07-10
python3 migration: comedi simple import

philipwinklerdelmic:


src/odemis/driver/comedi_simple.py
```diff
@@ -27,7 +27,10 @@
 
 
 # first, add everything from comedi as is
-import comedi as _comedi
+try:
+    from comedi import comedi as _comedi  # python3
```
pieleric: cf https://bugs.launchpad.net/ubuntu/+source/comedilib/+bug/1797258

I think we'll have to support this too, but IMHO, that's a packaging bug.

philipwinklerdelmic: Should I repackage comedi for python3?

pieleric: Let's not spend too much time on this. As long as it's only this issue, I'd rather not have to maintain our own version of comedi. So, for now, let's just leave this in place, but with a different comment, explaining this is just to handle some mistake in packaging on the Python3 version of comedi in some Debian/Ubuntu versions.

philipwinklerdelmic: done

PR #1092, opened 2019-07-10
python3 migration: pmtctrl driver

philipwinklerdelmic:


src/odemis/driver/pmtctrl.py
```diff
@@ -323,21 +323,21 @@ def _updateSupplied(self):
         self.supplied.notify(self.supplied.value)
 
     def _getIdentification(self):
-        return self._sendCommand("*IDN?")
+        return self._sendCommand(b"*IDN?")
```
pieleric: The return should be converted back to a string.

src/odemis/driver/pmtctrl.py
```diff
@@ -605,88 +605,88 @@ def _processMessage(self, msg):
...
                 if self._powerSupply:
-                    res = "1" + '\n'
+                    res = b"1" + b'\n'
                 else:
-                    res = "0" + '\n'
-            elif tokens[0] == "VOLT?":
-                res = str(self._gain) + '\n'
-            elif tokens[0] == "PCURR?":
-                res = str(self._prot_curr) + '\n'
+                    res = b"0" + b'\n'
```
pieleric: Maybe simplify to `b"0\n"`? (if so, do the same for all the similar lines).

philipwinklerdelmic:
done

PR #1093, opened 2019-07-10
python3 migration: rigol driver

philipwinklerdelmic:


amuskens:
I don't think there are any extra changes since the Rigol DG4000 API is actually slightly different

src/odemis/driver/rigol.py
```diff
@@ -248,67 +248,67 @@ def terminate(self):
...
+        msg = b"%s %s\n" % (cmd, val)
 
         with self._net_access:
             self.socket.sendall(msg)
 
     def sendQueryCommand(self, cmd):
         """
         Sends one command, and expect a reply
-        cmd (str): command to send, including the ?
+        cmd (byte str): command to send, including the ?
```
pieleric: I think it'd keep the code easier to read if cmd (and val) would be just "str" (basestring on Python2). It would then be converted to bytes by `sendQueryCommand()` and `sendOrderCommand()` (using ascii). The response seems to always be "sane" (ie, no weird character), so it should also be fine to decode it back immediately into a string. 

BTW, could you update the docstring here to indicate the type of the return value.

philipwinklerdelmic: done

src/odemis/driver/rigol.py
```diff
@@ -259,56 +256,59 @@ def sendOrderCommand(self, cmd, val=""):
         if not self._is_connected:
             raise IOError("Device %s not connected." % (self._host,))
 
-        msg = "%s %s\n" % (cmd, val)
-
+        msg = ("%s %s\n" % (cmd, val)).encode('acii')
```
pieleric: -> "ascii".

src/odemis/driver/rigol.py
```diff
@@ -259,56 +256,59 @@ def sendOrderCommand(self, cmd, val=""):
...
-        logging.debug("Received: %s", ans.encode('string_escape'))
+        logging.debug("Received: %s", ans.decode('latin1', 'backslashreplace'))
 
-        ans, left = ans.split("\n", 1)  # remove the end of line characters
+        ans, left = ans.split(b"\n", 1)  # remove the end of line characters
         if left:
             logging.error("Received too much data, will discard the end: %s",
-                          left.encode('string_escape'))
+                          left.decode('latin1', 'backslashreplace'))
+        ans = ans.decode('latin1') if sys.version_info[0] >= 3 else ans
```
pieleric: Please, drop the `if`. On Python 2 it's a no-op, and that's fine.
(and then you can remove the `import sys`)

src/odemis/driver/rigol.py
```diff
@@ -441,7 +441,7 @@ def _decodeMessage(self, msg):
             except TypeError:   # could not unpack message
                 self._error_state = True
                 logging.exception('%s: Error Setting duty cycle %s', self.name,
-                                  cmd.decode('latin1', 'backslashreplace'))
+                                  msg.decode('latin1', 'backslashreplace'))
```
pieleric: I agree that it's more useful to use `msg` than `cmd` in case the command couldn't be parsed. That said, I don't think `msg.split()` should never fail due to TypeError as it's always the right type (bytes). I understand that during the migration you've hit such case, but now that it's fixed, there is no reason it'd happen again. This `except` was actually put there for the `float()`.

So the change looks good, but I'm not very convinced by the commit message. 
Could you apply the same change to the previous block (APPL:SQU) too?

src/odemis/driver/rigol.py
philipwinklerdelmic: I think it is necessary.
`decode` should return a unicode in python2 and not a str.
`isinstance(u'a', str)` will return False

philipwinklerdelmic:
should be good now

PR #1095, opened 2019-07-10
python3 migration: picoquant driver

philipwinklerdelmic:


philipwinklerdelmic:
Should we make a distinction depending on whether we're on python2 (don't decode) or python3 (decode) or just decode all the time?
The same goes for the andorcam driver.

pieleric:
In general, I'd suggest to decode all the time. So once we move to Python 3-only code, it'll be immediately the right code. Anyway, on Python 2 `.decode("latin1")` does nothing.

philipwinklerdelmic:
I believe `.decode("latin1")` returns a unicode in python2 (so it's not a str). But it probably doesn't matter.

philipwinklerdelmic:
done

philipwinklerdelmic:
done

PR #1096, opened 2019-07-10
python3 migration: andorcam2 driver

philipwinklerdelmic:


philipwinklerdelmic:
should be working now

PR #1097, opened 2019-07-10
python3 migration: andorshrk driver

philipwinklerdelmic:


philipwinklerdelmic:
should be good now

pieleric:
I think that (almost?) every argument to the library which is instantiated with `create_string_buffer()` should be converted back-and-forth between bytes and (unicode) string on Python 3.
On the simulator side, just initialise the values as bytes, and you don't need to `.encode()` them.

philipwinklerdelmic:
Updated the pull request, but I'm not entirely convinced yet. The test cases always seem to pass, maybe I need to test some more things.

philipwinklerdelmic:
done

PR #1106, opened 2019-07-16
python3 migration: pigcs driver

philipwinklerdelmic:


philipwinklerdelmic:
Tested on the hardware and seems to be working (all the testcases that passed before are passing now), both in python2 and python3.

philipwinklerdelmic:
I made a couple of changes in the force push. The main change is that `_sendQuery` now takes str as input and not bytes. That simplifies the code a lot. The tests still pass on the hardware.

philipwinklerdelmic:
@pieleric Do you think it might be a good idea to return unicode instead of bytes in the `_sendQuery` function in python3 and bytes in python2 (in other words return str)? That would simplify the interface even further. We could decode with latin1, so we can deal with funny characters.

src/odemis/driver/pigcs.py
```diff
@@ -511,7 +511,7 @@ def recoverTimeout(self):
         try:
             for i in range(2):
                 self.busacc.flushInput()
-                resp = self.busacc.sendQueryCommand(self.address, "ERR?\n")
+                resp = self.busacc.sendQueryCommand(self.address, "ERR?\n").decode('latin1', 'backslashreplace')
```
pieleric: Don't decode to a string using "backslashreplace". That's just useful when outputting logs, to make it readable. However, for data processing, you never want to have this.

src/odemis/driver/pigcs.py
```diff
@@ -580,7 +580,8 @@ def GetAxes(self):
         # SAI? (Get List Of Current Axis Identifiers)
         # SAI? ALL: list all axes (included disabled ones), one per line
         answer = self._sendQueryCommand("SAI? ALL\n")
-        axes = set(a for a in answer)
+        answer = list([answer]) if not isinstance(answer, list) else answer
+        axes = set([a.decode('ascii') for a in answer])
```
pieleric: Use "latin1" instead of "ascii", to be sure there is no issue with unusual characters.

src/odemis/driver/pigcs.py
philipwinklerdelmic: What about functions like `GetIdentification` or `GetStageName`? Should I use backslashreplace there or not?

philipwinklerdelmic:
should I also update this to return str in the sendcommand functions, instead of bytes?

src/odemis/driver/pigcs.py
```diff
@@ -511,7 +511,7 @@ def recoverTimeout(self):
         try:
             for i in range(2):
                 self.busacc.flushInput()
-                resp = self.busacc.sendQueryCommand(self.address, "ERR?\n")
+                resp = self.busacc.sendQueryCommand(self.address, "ERR?\n").decode('latin1')
 
                 if isinstance(resp, list):
```
pieleric: This line is a sign that resp could sometimes be a `list`. In such case `decode()` will not work on it. So you need to either delay. Either just after this check, of maybe after the `re.match()`?

src/odemis/driver/pigcs.py
```diff
@@ -638,18 +641,18 @@ def GetAvailableParameters(self):
 
     def GetParameter(self, axis, param):
         """
-        axis (bytes): axis number
+        axis (str): axis number
         param (0<int): parameter id (cf p.35)
         returns (str): the string representing this parameter
         """
         # SPA? (Get Volatile Memory Parameters)
-        assert(isinstance(axis, bytes) and 1 <= len(axis) <= 8)
```
pieleric: Please don't drop this check (and all the other ones). Just update it bytes -> str.

src/odemis/driver/pigcs.py
```diff
@@ -696,19 +699,18 @@ def SetParameter(self, axis, param, val, check=True):
...
         # SEP? (Get Non-Volatile Memory Parameters)
-        assert(isinstance(axis, bytes) and 1 <= len(axis) <= 8)
         assert 0 <= param
         if hasattr(self, "_avail_params") and param not in self._avail_params:
             raise ValueError("Parameter %s %d not available" % (axis, param))
 
         answer = self._sendQueryCommand("SEP? %s %d\n" % (axis, param))
         try:
-            value = answer.split("=")[1]
+            value = answer.split(b"=")[1]
```
pieleric: Also `decode("latin1")` here. At least, the returned value should be a str.

src/odemis/driver/pigcs.py
```diff
@@ -739,7 +741,7 @@ def _readAxisValue(self, com, axis):
 
         resp = self._sendQueryCommand("%s %s\n" % (com, axis))
         try:
-            value_str = resp.split("=")[1]
+            value_str = resp.split(b"=")[1]
         except IndexError:
             raise ValueError("Failed to parse answer from %s %s: '%s'" % (com, axis, resp))
         try:
```
pieleric: Need to `decode("latin1")` at least when it's neither an int nor a float.

src/odemis/driver/pigcs.py
```diff
@@ -867,11 +869,12 @@ def IsOnTarget(self, axis, check=True):
...
                 raise PIGCSError(err)
             r = lresp[1]
-            ss = r.split("=")
+            ss = r.split(b"=")
             if len(ss) != 2:
                 raise ValueError("Failed to parse answer from %s: '%s'" %
-                                 (com, lresp))
-            return ss[1] == "1"
+                                 ([c.decode('latin1', 'backslashreplace') for c in com],
+                                  lresp.decode('latin1', 'backslashreplace')))
```
pieleric: `lresp` is also a list. => also a list comprehension. Alternatively, you could just use "%r" and be done ;-) We don't need too much beauty in this very unlikely error message.

src/odemis/driver/pigcs.py
```diff
@@ -3539,16 +3541,16 @@ def sendOrderCommand(self, addr, com):
         else:
             full_com = "%d %s" % (addr, com)
         with self.ser_access:
-            logging.debug("Sending: '%s'", full_com.encode('string_escape'))
-            self.serial.write(full_com)
+            logging.debug("Sending: '%s'", full_com)
```
pieleric: We actually should escape `full_com`. Again, maybe use `%r` to avoid making it too complex.

src/odemis/driver/pigcs.py
```diff
@@ -3572,19 +3574,19 @@ def sendQueryCommand(self, addr, com):
...
+            prefix = b""
         else:
             full_com = "".join("%d %s" % (addr, c) for c in com)
-            prefix = "0 %d " % addr
+            prefix = b"0 %d " % addr
 
         with self.ser_access:
-            logging.debug("Sending: '%s'", full_com.encode('string_escape'))
-            self.serial.write(full_com)
+            logging.debug("Sending: '%s'", full_com)
```
pieleric: `%r`

src/odemis/driver/pigcs.py
```diff
@@ -3594,11 +3596,12 @@ def sendQueryCommand(self, addr, com):
                                         "plugged in and turned on." % addr)
                 ans += char
 
-                anssplited = ans.split("\n")
+                anssplited = ans.split(b"\n")
                 # if the answer finishes with \n, last split is empty
                 anssplited, ans = anssplited[:-1], anssplited[-1]
                 if anssplited:
-                    logging.debug("Received: '%s'", "\n".join(anssplited).encode('string_escape'))
+                    anssplited_str = [i.decode('latin1') for i in anssplited]
```
pieleric: This one needs a "backslashreplace" (as it's for a log).

src/odemis/driver/pigcs.py
philipwinklerdelmic: It seems to me that we don't support passing lists to an order command, only to query commands. Otherwise, the next line `serial.write(full_com)` would probably fail.

src/odemis/driver/pigcs.py
philipwinklerdelmic: full_com should be a string by now, if the argument was a list, it's converted by `"".join()`.

src/odemis/driver/pigcs.py
philipwinklerdelmic: len(axis) is always 1, so it doesn't really make sense to test for that

philipwinklerdelmic: also we should use basestring, if we decode senquerycommand all the time, then some commands can be unicode

src/odemis/driver/pigcs.py
```diff
@@ -478,21 +478,19 @@ def _sendQueryCommand(self, com):
...
+                if isinstance(com, list):
                     full_com = "".join(com)
 
                 if success:
-                    logging.warning("Controller %s timeout after '%s', but recovered.",
-                                    self.address, full_com.encode('string_escape'))
+                    logging.warning("Controller %s timeout after '%r', but recovered.",
+                                    self.address, full_com)
                     # try one more time (and it has to work this time)
                     lines = self.busacc.sendQueryCommand(self.address, com)
```
philipwinklerdelmic: I think this should be full_com instead of com

philipwinklerdelmic: also %s should suffice since full_com is always a string

philipwinklerdelmic:
I changed quite a bit, mainly the return type of `sendQuery` is now a (unicode) str. Also, I don't agree with using %r in all the places (if the argument is a string, it shouldn't be necessary, right?). I think there might have been a bug in line 488, where we send `com` a second time instead of `full_com`.

src/odemis/driver/pigcs.py
pieleric: Normally, for the wrapping functions, you shouldn't escape the returned data. Either detect there is something wrong in it, or pass it as-is (just do that, it's easier ;-) ).

src/odemis/driver/pigcs.py
pieleric: Let's escape it with `to_str_escape()`.

pieleric:
Sorry, due to my misunderstanding of "backslashreplace", all of these need to be replaced with `to_str_escape()`. In general, every string/bytes received from the hardware must be escaped before being shown in a message, unless it's absolutely certain that it doesn't contain control characters (or other weird values).

src/odemis/driver/pigcs.py
pieleric: What makes you think that `axis` always of length 1? I'm pretty sure that the user could pass any string as axis. Also, I think that the hardware supports axes names of multiple characters. See `GetAxes()`.

philipwinklerdelmic:
done

philipwinklerdelmic:
updated

src/odemis/driver/pigcs.py
```diff
@@ -2843,8 +2840,8 @@ def __init__(self, name, role, port, axes, baudrate=38400,
         controllers = {} # address -> kwargs (axes, dist_to_steps, min_dist, vpms...)
         for axis, (add, channel, isCL) in axes.items():
             if isinstance(channel, int):
-                channel = str(channel)
-            channel = channel.encode("ascii")  # bytes
+                channel = "%d" % channel
+            channel = channel  # str
```
pieleric: That's a little redundant code. Just move the `# str` comment to the line above, and drop this line.

src/odemis/driver/pigcs.py
pieleric: Still needs an escape or a `%r`

src/odemis/driver/pigcs.py
```diff
@@ -478,21 +478,19 @@ def _sendQueryCommand(self, com):
                     raise
 
                 success = self.recoverTimeout()
-                if isinstance(com, basestring):
-                    full_com = com
-                else:
+                if isinstance(com, list):
                     full_com = "".join(com)
```
pieleric:
```
else:
    full_com = com
```

src/odemis/driver/pigcs.py
```diff
@@ -3728,14 +3725,14 @@ def sendQueryCommand(self, addr, com):
...
         if addr is None:
-            full_com = "".join(com)
-            prefix = ""
+            full_com = b"".join(com)
+            prefix = b""
         else:
-            full_com = "".join("%d %s" % (addr, c) for c in com)
-            prefix = "0 %d " % addr
+            full_com = b"".join(b"%d %s" % (addr, c) for c in com)
+            prefix = b"0 %d " % addr
```
pieleric: `com` is a string or a list of string. So you cannot directly get bytes out of them.
I think that the easiest would be to compute first `full_com` as a string (in both sides of the if), then convert it to bytes with `.encode("latin1")`

philipwinklerdelmic:
updated

PR #1107, opened 2019-07-16
driver: pigcs test

philipwinklerdelmic:


philipwinklerdelmic:
This didn't happen before the python3 migration, don't know why it happens now. Might just be some concurrency issue.

philipwinklerdelmic:
i couldn't reproduce it anymore, so maybe we can close this PR and investigate more another time in case it happens again.

PR #1108, opened 2019-07-16
streak cam bug fix: handle WL_LIST if numpy.array

Sabrina85:


src/odemis/acq/stream/_static.py
```diff
@@ -564,7 +564,8 @@ def __init__(self, name, image, *args, **kwargs):
 
         # This is for "average spectrum" projection
         # cached list of wavelength for each pixel pos
-        self._wl_px_values, unit_bw = spectrum.get_spectrum_range(image)
+        wl, unit_bw = spectrum.get_spectrum_range(image)
+        self._wl_px_values = numpy.array(wl) if not isinstance(wl, numpy.ndarray) else wl
```
pieleric: I finally found the "right" command to force an object to be a ndarray without copying it if it's already an ndarray:
`self._wl_px_values = numpy.array(wl, copy=False)  # Force it to be a numpy array`

Sabrina85: nice!

Sabrina85:
Force updated commit

PR #1111, opened 2019-07-17
python3-migration spectrapro driver

philipwinklerdelmic:


src/odemis/driver/spectrapro.py
```diff
@@ -339,8 +340,8 @@ def _sendQuery(self, com, timeout=1):
 
         assert(1 < len(com) <= 100) # commands cannot be long
         com += "\r"
-
```
philipwinklerdelmic: It seems like we send and receive weird characters. Is 'latin1' going to work?

src/odemis/driver/spectrapro.py
```diff
@@ -353,26 +354,26 @@ def _sendQuery(self, com, timeout=1):
...
+               not (response.endswith(b" ok\r\n") or response.endswith(b"? \r\n"))):
             self._serial.timeout = max(0.1, timeend - time.time())
             char = self._serial.read()
             if not char: # timeout
                 break
             response += char
 
-        logging.debug("Received: %s", response.encode('string_escape'))
-        if response.endswith(" ok\r\n"):
-            return response[:-5]
```
philipwinklerdelmic: I changed the interface of the sendQuery function compared to the other drivers, here I return unicode. I think that I don't need any changes in the functions that call sendQuery if I do it like this. Do you think it's a good idea?

philipwinklerdelmic: Alternatively, we could decode in python3 and do nothing in python2, so we always return str. I think that might be even better.

src/odemis/driver/spectrapro.py
```diff
@@ -51,10 +51,11 @@ def hextof(s):
     if len(s) != 23:
         raise ValueError("Cannot convert %s to a float" % (s,))
 
-    d = b""
+    d = ""
     try:
         for c in s.split(","):
             d += chr(int(c, 16))
+        d = d.encode('latin1') if not isinstance(d, bytes) else d
         return struct.unpack("<d", d)[0]
```
pieleric: I think you can write:
```
d = bytearray(int(c, 16) for c in s.split(","))
return struct.unpack("<d", d)[0]
```
(in the try/catch)

src/odemis/driver/spectrapro.py
pieleric: I'm fine with the simple `decode("latin1")`.
Eventually, we will only run on Python 3, so the more the code looks like the final code, the better.

src/odemis/driver/spectrapro.py
```diff
@@ -381,7 +382,7 @@ def _sendQuery(self, com, timeout=1):
                     raise IOError("Device keeps sending data")
                 response += garbage
                 raise SPError("Sent '%s' and received error: '%s'" %
-                              (com.encode('string_escape'), response.encode('string_escape')))
+                              (com, response.decode('latin1', 'backslashreplace')))
```
pieleric: You should also escape  `com`, which is bytes.

philipwinklerdelmic:
done

PR #1115, opened 2019-07-18
Newport ESP301 identifier fix

amuskens:
Not tested yet.

Checks the identifier of the Newport to determine if it is the right device.

### DIFF
--- a/src/odemis/driver/npmc.py
+++ b/src/odemis/driver/npmc.py
@@ -294,6 +294,9 @@ class ESP(model.Actuator):
 
                 try:
                     ve = self.GetVersion()
+                    if not re.match(r"ESP301 Version \d\.\d\.\d (1[0-2]|[1-9])\/(3[01]|[12][0-9]|[1-9])\/(99|[0-9]?[0-9])", ve):
+                        raise IOError("Device at %s is not an ESP301 controller" % ports)
+
                 except ESPError as e:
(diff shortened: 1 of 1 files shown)


src/odemis/driver/npmc.py
```diff
@@ -294,6 +294,9 @@ def _findDevice(self, ports, baudrate=19200):
 
                 try:
                     ve = self.GetVersion()
+                    if not re.match(r"ESP301 Version \d\.\d\.\d (1[0-2]|[1-9])\/(3[01]|[12][0-9]|[1-9])\/(99|[0-9]?[0-9])", ve):
```
pieleric: Let's be a little bit more relaxed on the match. I'd say that if `ve.upper()` contains ESP301, it's fine.

src/odemis/driver/npmc.py
```diff
@@ -294,6 +294,9 @@ def _findDevice(self, ports, baudrate=19200):
...
+                        raise IOError("Device at %s is not an ESP301 controller" % ports)
```
pieleric: ports -> n.
Please also include the content of `ve` inside the error message.

src/odemis/driver/npmc.py
amuskens: Haha... I got a bit carried away with the Regex I think.

PR #1116, opened 2019-07-18
driver zeiss: be more picky in connecting to the device

pieleric:


pieleric:
This is inspired by PR #1115.

PR #1117, opened 2019-07-18
Bug hamamatsu driver: list can not be encoded with string_escape

Sabrina85:


Sabrina85:
It's updated. I put `tuple of str`, as it is not always 3.

PR #1118, opened 2019-07-18
Bug fix streakcam: bug in MD_TIME_RANGE_TO_DELAY handling

Sabrina85:


src/odemis/driver/test/hamamatsurx_test.py
```diff
@@ -325,12 +325,12 @@ def test_MCPGain(self):
...
         """Test time range VA for sweeping of streak unit."""
-        for timeRange in self.streakunit.timeRange.choices:
+        for timeRange in self.streakunit.timeRange.choices:  # values different from yaml due to floating point issues
             # change timeRange VA to value in range
-            self.streakunit.timeRange.value = util.find_closest(timeRange, self.streakunit.timeRange.choices)
+            self.streakunit.timeRange.value = timeRange
             self.assertAlmostEqual(self.streakunit.timeRange.value, timeRange)
 
-            tr2d = self.delaybox._metadata.get(model.MD_TIME_RANGE_TO_DELAY)
+            tr2d = self.delaybox._metadata.get(model.MD_TIME_RANGE_TO_DELAY)  # TODO MD not updated when HW VA changes?
```
pieleric: Yes, that's expected. It's up to the "client" to update the calibration.

src/odemis/gui/cont/settings.py
```diff
@@ -838,7 +838,24 @@ def _onUpdateTriggerDelayMD(self, evt):
...
+            trigger2delay_MD[key] = requested_triggerDelay
+        else:
+            trigger2delay_MD[cur_timeRange] = requested_triggerDelay
+            logging.warning("A new time range %s was added to the MD, which is not consistent with the timeRanges"
+                            " specified in the yaml file.", cur_timeRange)
+
+        # check the number of keys in the dict is same as choices for VA
+        if len(trigger2delay_MD.keys()) != len(self.streak_unit.timeRange.choices):
+            logging.warning("The number of %s time ranges in the trigger delay MD does not match with the number of %s "
+                            "choices specified in the yaml file.",
```
pieleric: In general, I'd suggest to avoid talking of "the yaml file" in the GUI. Firstly, it's not very precise (as YAML is a just file format), so "microscope file" would be more adapted. More theoretically, I think it's better to keep it generic, and not have expection from the GUI on how the back-end is started.
Anyway, here I think here the time ranges are from the `timeRange.choices`, so you could write something like:
`"MD_TIME_RANGE_TO_DELAY has %d entries, while the device .timeRange has %d choices."`

(also update the message just before).

src/odemis/driver/test/hamamatsurx_test.py
Sabrina85: yes, forgot to remove this comment after we discussed

PR #1119, opened 2019-07-19
driver simcam: add support for stage connection and multi FM channel

ihebdelmic:
By passing a stage and cropping the simulated image, with the resolution argument,
it's possible to simulate the movement of the stage.

It will also look at the MD_WL_IN metadata to pick the best channel if the
input image has several wavelength bands.

### DIFF
--- a/src/odemis/driver/simcam.py
+++ b/src/odemis/driver/simcam.py
@@ -2,21 +2,21 @@
 '''
 Created on 19 Jun 2014
 
-@author: Éric Piel
+@author: Éric Piel, Iheb Zaabouti
 
 Copyright © 2014 Éric Piel, Kimon Tsitsikas, Delmic
 
 This file is part of Odemis.
 
-Odemis is free software: you can redistribute it and/or modify it under the terms 
-of the GNU General Public License version 2 as published by the Free Software 
+Odemis is free software: you can redistribute it and/or modify it under the terms
+of the GNU General Public License version 2 as published by the Free Software
 Foundation.
 
-Odemis is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; 
-without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR 
+Odemis is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
 PURPOSE. See the GNU General Public License for more details.
 
-You should have received a copy of the GNU General Public License along with 
+You should have received a copy of the GNU General Public License along with
 Odemis. If not, see http://www.gnu.org/licenses/.
 '''
 from __future__ import division
@@ -38,7 +38,7 @@ class Camera(model.DigitalCamera):
     given at initialisation.
     '''
 
-    def __init__(self, name, role, image, dependencies=None, daemon=None, blur_factor=1e4, **kwargs):
+    def __init__(self, name, role, image, dependencies=None, daemon=None, blur_factor=1e4, resolution=None, **kwargs):
         '''
         dependencies (dict string->Component): If "focus" is passed, and it's an
             actuator with a z axis, the image will be blurred based on the
@@ -46,6 +46,7 @@ class Camera(model.DigitalCamera):
         image (str or None): path to a file to use as fake image (relative to
          the directory of this class)
         '''
+
         # TODO: support transpose? If not, warn that it's not accepted
         # fake image setup
         image = unicode(image)
@@ -53,33 +54,41 @@ class Camera(model.DigitalCamera):
         if not os.path.isabs(image):
             image = os.path.join(os.path.dirname(__file__), image)
         converter = dataio.find_fittest_converter(image, mode=os.O_RDONLY)
-        self._img = converter.read_data(image)[0]  # can be RGB or greyscale
-
+        self._imgs = converter.read_data(image)      # can be RGB or greyscale
         model.DigitalCamera.__init__(self, name, role, dependencies=dependencies, daemon=daemon, **kwargs)
 
-        if self._img.ndim > 3:  # remove dims of length 1
-            self._img = numpy.squeeze(self._img)
+        for i, img in enumerate(self._imgs):
+            if img.ndim > 3:  # remove dims of length 1
+                self._imgs[i] = numpy.squeeze(img)
+            imshp = img.shape
+            if len(imshp) == 3 and imshp[i] in {3, 4}:
+                # CYX, change it to YXC, to simulate a RGB detector
+                self._imgs[i] = util.img.ensureYXC(img)
 
-        imshp = self._img.shape
-        if len(imshp) == 3 and imshp[0] in {3, 4}:
-            # CYX, change it to YXC, to simulate a RGB detector
-            self._img = numpy.rollaxis(self._img, 2) # XCY
-            self._img = numpy.rollaxis(self._img, 2) # YXC
-            imshp = self._img.shape
+        for img in self._imgs[1:]:
+            if self._imgs[0].shape != self._imgs[i].shape:
+                raise ValueError("all images must have the same resolution")
+        imshp = self._imgs[0].shape
 
         # For RGB, the colour is last dim, but we still indicate it as higher
         # dimension to ensure shape always starts with X, Y
         if len(imshp) == 3 and imshp[-1] in {3, 4}:
             # resolution doesn't affect RGB dim
-            res = imshp[-2::-1]
+            if resolution:
+                if resolution >= imshp[-2::-1]:
+                    res = tuple(resolution)
+            else:
+                res = imshp[-2::-1]
             self._shape = res + imshp[-1::] # X, Y, C
-            # indicate it's RGB pixel-per-pixel ordered
-            self._img.metadata[model.MD_DIMS] = "YXC"
         else:
-            res = imshp[::-1]
+            if resolution:
+                res = tuple(resolution)
+            else:
+                res = imshp[::-1]
+
             self._shape = res # X, Y,...
         # TODO: handle non integer dtypes
-        depth = 2 ** (self._img.dtype.itemsize * 8)
+        depth = 2 ** (self._imgs[0].dtype.itemsize * 8)
         self._shape += (depth,)
 
         self._resolution = res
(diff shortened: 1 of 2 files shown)


src/odemis/driver/simcam.py
```diff
@@ -289,19 +329,69 @@ def _simulate(self):
...
+                if abs(current_inwl_c - c) < closest:
+                    _img = im
+                    closest = abs(current_inwl_c - c)
+
+            # _img = min(self._imgs, key=lambda im: abs(util.fluo.get_center(im.metadata[model.MD_IN_WL] - current_inwl_c)))
+
+            current_inwl = self.getMetadata().get(model.MD_IN_WL, None)
+            # logging.debug('current_inwl %s', current_inwl)
+            # logging.debug('im.metadata[model.MD_IN_WL] %s', self._imgs[0].metadata[model.MD_IN_WL])
+            # logging.debug('im.metadata[model.MD_IN_WL]1111111111111 %s', self._imgs[1].metadata[model.MD_IN_WL])
```
pieleric: Drop all these 5 lines, they are for debug.

src/odemis/driver/simcam.py
```diff
@@ -289,19 +329,69 @@ def _simulate(self):
...
+            Y = exshift[1] / pxs[1]
+            exs = (X,Y)
+
+            npos=numpy.array(pxs_pos) + numpy.array(exs)
+            pxs_pos=tuple(npos)
+
+            logging.debug('the position is %s',pxs_pos)
+
+        # Pick the image based on the current excitation wl
+
```
pieleric: Drop this empty line.

src/odemis/driver/simcam.py
```diff
@@ -289,19 +329,69 @@ def _simulate(self):
...
+        try:
+            current_inwl = self.getMetadata()[model.MD_IN_WL]   # (350e-9, 360e-9)
+            logging.debug('current_inwl %s',current_inwl)
```
pieleric: Better move this log message after the selection, indicating the current wl and the wavelength picked.

src/odemis/driver/simcam.py
```diff
@@ -289,19 +329,69 @@ def _simulate(self):
...
+
+            closest = float('inf')
+            _img = self._imgs[0]
+            for im in self._imgs:
+                c = util.fluo.get_center(im.metadata[model.MD_IN_WL])
+                if abs(current_inwl_c - c) < closest:
+                    _img = im
+                    closest = abs(current_inwl_c - c)
+
+            # _img = min(self._imgs, key=lambda im: abs(util.fluo.get_center(im.metadata[model.MD_IN_WL] - current_inwl_c)))
```
pieleric: I think you can mix between these 2 writings, and get something more compact and clearer:
```
def dist_wl(img):
  try:
      c = util.fluo.get_center(im.metadata[model.MD_IN_WL])
      return abs(current_inwl_c - c)
  except KeyError:
      return float('inf')

img = min(self._imgs, key=dist_wl)
```

src/odemis/driver/simcam.py
```diff
@@ -289,19 +329,69 @@ def _simulate(self):
...
+
+            current_inwl = self.getMetadata().get(model.MD_IN_WL, None)
+            # logging.debug('current_inwl %s', current_inwl)
+            # logging.debug('im.metadata[model.MD_IN_WL] %s', self._imgs[0].metadata[model.MD_IN_WL])
+            # logging.debug('im.metadata[model.MD_IN_WL]1111111111111 %s', self._imgs[1].metadata[model.MD_IN_WL])
+
+        except KeyError:
+            _img = self._imgs[0]
+
+        shape = _img.shape
```
pieleric: `_img` -> `img`.
Underscores are to indicate a class attribute is private. For a variable inside a function, you never need to start with _.

src/odemis/driver/simcam.py
```diff
@@ -289,19 +329,69 @@ def _simulate(self):
...
         # compute each row and column that will be included
         # TODO: Could use something more hardwarish like that:
         # data0 = data0.reshape(shape[0]//b0, b0, shape[1]//b1, b1).mean(3).mean(1)
         # (or use sum, to simulate binning)
         # Alternatively, it could use just [lt:lt+res:binning]
         coord = ([int(round(lt[0] + i * binning[0])) for i in range(res[0])],
                  [int(round(lt[1] + i * binning[1])) for i in range(res[1])])
-        sim_img = self._img[numpy.ix_(coord[1], coord[0])]  # copy
+
+        sim_img = _img[numpy.ix_(coord[1], coord[0])]
```
pieleric: No need to add extra empty lines.

src/odemis/driver/simcam.py
```diff
@@ -289,19 +329,69 @@ def _simulate(self):
...
+        shape = _img.shape
+
         center = (shape[1] / 2, shape[0] / 2)
-        lt = (center[0] + pxs_pos[0] - (res[0] / 2) * binning[0],
+
+        lt = (-center[0] - pxs_pos[0] - (res[0] / 2) * binning[0],
               center[1] + pxs_pos[1] - (res[1] / 2) * binning[1])
-        assert(lt[0] >= 0 and lt[1] >= 0)
+
+        # assert(lt[0] >= 0 and lt[1] >= 0)
```
pieleric: Why disabling the assert? If it's not valid anymore, remove it, otherwise, put it back.
BTW, how is it handled when the stage is causing the image area outside of the bounds (eg, <0 or >= res)? Does it wrap around? Is it clipped? I think there should be a comment explaining. Also, if the image returned is smaller than  the expected resolution, that's not good. It should always be the size requested by the user.
Maybe the use of `% res[0]` in the computation of `coord` would make it wrap properly.

src/odemis/driver/test/simcam_test.py
```diff
@@ -31,18 +31,28 @@
...
 
 # TODO focus
 
 
 class TestSimCam(unittest.TestCase):
 
+    # @classmethod
+    # def setUpClass(cls):
+    #     cls.focus = simulated.Stage(**KWARGS_FOCUS)
+    #     cls.camera = CLASS(children={"focus": cls.focus}, **KWARGS)
```
pieleric: Drop all these lines.

src/odemis/driver/test/simcam_test.py
```diff
@@ -31,18 +31,28 @@
 
 CLASS = simcam.Camera
 KWARGS_FOCUS = {"name": "focus", "role": "overview-focus", "axes": ["z"], "ranges": {"z": [0, 0.012]}}
-KWARGS = dict(name="camera", role="overview", image="simcam-fake-overview.h5")
+
+KWARGS_STAGE = {"name": "stage", "role": "stage", "axes": ["x","y"], "ranges": {"x": [0, 0.013],"y": [0, 0.013]}}
+#KWARGS = dict(name="camera", role="overview", image="/home/zaabouti/Downloads/FM-grid/tile.tif", resolution=[128, 128])
+KWARGS = dict(name="camera", role="overview", image="9.tif", resolution=[512, 512])
```
pieleric: Drop the commented line.
Also, "9.tif" doesn't exists. I'd rather not introduce even more test images in the Odemis source code. Maybe you can just use "songbird-sim-ccd.h5", which has MD_WL_IN.

src/odemis/driver/test/simcam_test.py
```diff
@@ -31,18 +31,28 @@
 
 CLASS = simcam.Camera
 KWARGS_FOCUS = {"name": "focus", "role": "overview-focus", "axes": ["z"], "ranges": {"z": [0, 0.012]}}
-KWARGS = dict(name="camera", role="overview", image="simcam-fake-overview.h5")
```
pieleric: Please still test at list a little bit with this RGB image.
I'd suggest to have a separate class `TestSimCamRGB`, which uses these arguments.

src/odemis/driver/test/simcam_test.py
```diff
@@ -108,12 +118,14 @@ def test_roi(self):
...
         self.camera.translation.value = (0, 0)
 
-#     @unittest.skip("simple")
+    # @unittest.skip("simple")
     def test_acquire(self):
         self.assertGreaterEqual(len(self.camera.shape), 3)
         exposure = 0.1
         self._ensureExp(exposure)
 
+        self.camera.updateMetadata({model.MD_IN_WL: (500e-9, 700e-9)})
```
pieleric: You need to try with different metadata. First try when there is no MD_IN_WL, then with this value, and then with values very far, eg, `(900e-9, 950e-9)`.

src/odemis/driver/test/simcam_test.py
```diff
@@ -303,6 +317,40 @@ def test_focus(self):
...
+        # restore original position
+        f = self.stage.moveAbs(pos)
+        logging.debug('pos= %s',pos)
+        f.result()
+        self.assertEqual(self.stage.position.value, pos)
+        d0n =  self.camera.data.get()
+        numpy.testing.assert_array_almost_equal(d0, d0n)
+
+
+
```
pieleric: No need for all these empty lines. Please put maximum 1 empty line between blocks.

src/odemis/driver/test/simcam_test.py
```diff
@@ -303,6 +317,40 @@ def test_focus(self):
...
+        self.assertNotEqual(self.stage.position.value, pos)
+        d1 = self.camera.data.get()
+        self.assertTrue(numpy.any(d0 != d1))
+
+
+        f = self.stage.moveRel({"y": 1e-3}) # 1 mm
+        f.result()
+        self.assertNotEqual(self.stage.position.value, pos)
+        self.camera.data.get()
+
```
pieleric: Please check that the image returned is the right shape.

src/odemis/driver/simcam.py
```diff
@@ -38,48 +38,59 @@ class Camera(model.DigitalCamera):
...
+
+        imshp = self._imgs[0].shape
 
-        imshp = self._img.shape
         if len(imshp) == 3 and imshp[0] in {3, 4}:
             # CYX, change it to YXC, to simulate a RGB detector
-            self._img = numpy.rollaxis(self._img, 2) # XCY
-            self._img = numpy.rollaxis(self._img, 2) # YXC
-            imshp = self._img.shape
+            self._imgs[0] = numpy.rollaxis(self._imgs[0], 2) # XCY
```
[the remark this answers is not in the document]
ihebdelmic: I am not able to find img in util.

src/odemis/driver/simcam.py
```diff
@@ -38,48 +38,59 @@ class Camera(model.DigitalCamera):
...
+        self._imgs = converter.read_data(image)
 
         model.DigitalCamera.__init__(self, name, role, dependencies=dependencies, daemon=daemon, **kwargs)
 
-        if self._img.ndim > 3:  # remove dims of length 1
-            self._img = numpy.squeeze(self._img)
+        if self._imgs[0].ndim > 3:  # remove dims of length 1
+            self._imgs[0] = numpy.squeeze(self._imgs[0])
+
+        imshp = self._imgs[0].shape
```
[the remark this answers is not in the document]
ihebdelmic: i actually make it rising an error but i prefer if we  drop different resolution images

src/odemis/driver/simcam.py
ihebdelmic: actually outside the bounds its black, this can cause problems if the acqusition goes over bounds as the detector does not receive any images, i will change it, what is the best option ?

pieleric: The modulo funciton, as proposed, should work fine, and it's very easy to implement.

pieleric:
You've "resolved" the comments, but haven't yet pushed your changes to the branch. Please push, to get to the next (and hopefully) final round of review :-)

ihebdelmic:
i still have a s;al changes to the test file but the driver is normally done

src/odemis/driver/simcam.py
```diff
@@ -289,22 +317,50 @@ def _simulate(self):
...
+
+            phys_shift = self._orig_stage_pos[0] - pos[0], self._orig_stage_pos[1] - pos[1]
+            phys_shift = tuple(phys_shift)
+
+            # #convert to pxl
+            pxs = self._metadata[model.MD_PIXEL_SIZE]
+            px_shift = phys_shift[0] / pxs[0], phys_shift[1] / pxs[1]
+
+            npos=pxs_pos[0]+px_shift[0], pxs_pos[1]+px_shift[1]
+            pxs_pos=tuple(npos)
```
pieleric: Just write `pxs_pos = pxs_pos[0] + px_shift[0], pxs_pos[1] + px_shift[1]`
BTW, please make sure to leave spaces, as it's the official formatting. (PEP8)

src/odemis/driver/simcam.py
pieleric: Please either put back the assert, if it's still correct, or entirely drop the line.

effting:
Latest update on this pull request is by @pieleric on Sep 18, 2019. Is this still relevant? Should we close this request?

pieleric:
I think this feature could be useful, but no one is working on it now. I'm parking it on separate branch `simcam-multiple-colour`.

PR #1120, opened 2019-07-19
gui comp overlay: allow the RepetitionSelectOverlay to not use a rela…

ihebdelmic:
…tive ROA

Sometimes the ROA is generic, for the physical coordinates.
In such case, it shouldn't be between 0->1, relative to a scanner,
but just in meter (relative to the stage).

### DIFF
--- a/src/odemis/gui/comp/overlay/world.py
+++ b/src/odemis/gui/comp/overlay/world.py
@@ -266,19 +266,20 @@ class RepetitionSelectOverlay(WorldSelectOverlay):
         roa (None or VA of 4 floats): If not None, it's linked to the rectangle
           displayed (ie, when the user changes the rectangle, its value is
           updated, and when its value changes, the rectangle is redrawn
-          accordingly). Value is relative to the scanner.
+          accordingly). Value is relative to the scanner (if passed), and otherwise it's absolute (in m).
         scanner (None or HwComponent): The scanner component to which the relative
-          ROA values refer to. Required if a roa argument is passed.
+         ROA. If None, the roa argument is interpreted as absolute physical coordinates (m). If it's a HwComponent, the roa will be interpreted as a ratio of its fielf of viewd.
+
+
         """
         WorldSelectOverlay.__init__(self, cnvs, colour)
 
         self._fill = self.FILL_NONE
         self._repetition = (0, 0)
+
         self._roa = roa
         self._scanner = scanner
         if roa:
-            if not scanner:
-                raise ValueError("scanner is required when roa VA passed")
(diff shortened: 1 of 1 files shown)


ihebdelmic:
Changes are made .

src/odemis/gui/comp/overlay/world.py
```diff
@@ -268,17 +268,18 @@ def __init__(self, cnvs, roa=None, scanner=None, colour=gui.SELECTION_COLOUR):
           updated, and when its value changes, the rectangle is redrawn
           accordingly). Value is relative to the scanner.
         scanner (None or HwComponent): The scanner component to which the relative
-          ROA values refer to. Required if a roa argument is passed.
+          ROA values refer to. Required if a roa argument is passed and if the ROI is a ratio it should not be passed.
```
pieleric: Now, it not "required", as it's fine if it's None. So the following would be clearer: "If None, the roa argument is interpreted as absolute physical coordinates (m). If it's a HwComponent, the roa will be interpreted as a ratio of its fielf of view".
Also, update the docstring of `roa` too, to indicate "Value is relative to the scanner (if passed), and otherwise it's absolute (in m)".

src/odemis/gui/comp/overlay/world.py
```diff
@@ -390,18 +391,26 @@ def on_roa(self, roa):
...
-
         if self._roa:
             if self.active:
                 if self.get_size() != (None, None):
                     phys_rect = self.get_physical_sel()
-                    rel_rect = self.convert_roi_phys_to_ratio(phys_rect)
+                    if self._scanner:
+                        rel_rect = self.convert_roi_phys_to_ratio(phys_rect)
+                    else:
+                        rel_rect = phys_rect
```
pieleric: Actually, it's a little confusing to have the variable still named `rel_rect` (even if it contains physical coordinates). So I'd suggest to rename it to `rect` or `roa`.

src/odemis/gui/comp/overlay/world.py
```diff
@@ -390,18 +391,26 @@ def on_roa(self, roa):
         roi (tuple of 4 floats): left, top, right, bottom position relative to the SEM image
 
         """
-        phys_rect = self.convert_roi_ratio_to_phys(roa)
+
+        if self._scanner:
+            phys_rect = self.convert_roi_phys_to_ratio(roa)
+        else:
+            phys_rect = roa
+
```
pieleric: Please don't add double empty lines. Just one line is enough. Also, avoid starting the function with an empty line.

ihebdelmic:
Changes made.

PR #1122, opened 2019-07-22
python3 migration: powerctrl driver

philipwinklerdelmic:


philipwinklerdelmic:
@pieleric Could you have a look at this? I this interface for the `sendCommand` function which accepts `str` and returns `str` might be ideal. The amount of changes that we have to make becomes minimal.

Can we use latin1-encoding every time?

src/odemis/driver/powerctrl.py
```diff
@@ -263,23 +263,24 @@ def _sendCommand(self, cmd):
         """
         cmd = cmd + "\n"
         with self._ser_access:
-            logging.debug("Sending command %s", cmd.encode('string_escape'))
-            self._serial.write(cmd)
+            logging.debug("Sending command %s" % cmd)
```
philipwinklerdelmic: I should move up the encoding of cmd and then decode here with backslashreplace

pieleric: Maybe is it clearer if indeed you move the encoding to the very beginning of the function. Then, the logging handler would be the same as usual `ans.decode('latin1', 'backslashreplace')`.

src/odemis/driver/powerctrl.py
```diff
@@ -263,23 +263,24 @@ def _sendCommand(self, cmd):
...
                     # TODO: See how you should handle a timeout before you raise
                     # an HWError
                     raise HwError("Power Control Unit connection timeout. "
                                   "Please turn off and on the power to the box.")
                 # Handle ERROR coming from Power control unit firmware
                 ans += char
 
-            logging.debug("Received answer %s", ans.encode('string_escape'))
+            logging.debug("Received answer %s", ans.decode('latin1', 'backslashreplace'))
+            ans = ans.decode('latin1') if sys.version_info[0] >= 3 else ans
```
pieleric: The `if` shouldn't be necessary. On Python 2, it just becomes a no-op.

philipwinklerdelmic:
done

pieleric:
I think it's alright now... after replacing the `decode()` into a `to_str_escape()`.

philipwinklerdelmic:
done

PR #1123, opened 2019-07-22
driver: use simulator in pi_test if no hardware is available

philipwinklerdelmic:


philipwinklerdelmic:
Actually, I might have been too quick to skip the tests. We do have a simulator, so we should use that instead.

philipwinklerdelmic:
done

PR #1124, opened 2019-07-22
python3 migration: driver pi

philipwinklerdelmic:


philipwinklerdelmic:
This should be tested on the hardware (not a lot of simulation available).

src/odemis/driver/pi.py
```diff
@@ -185,8 +186,8 @@ def _sendGetCommand(self, com, prefix="", suffix="\r\n\x03"):
         return (string): the report without prefix nor newline
         """
         assert(len(com) <= 10)
-        assert(len(prefix) <= 2)
```
pieleric: Why removing this check?

src/odemis/driver/pi.py
```diff
@@ -743,6 +745,7 @@ def read(self):
         return self.file.read()
 
     def write(self, arg):
+        arg = arg.decode('latin1') if sys.version_info[0] >= 3 else arg
```
pieleric: If the file is opened in binanry mode, it should accept bytes (and return bytes), which would be more realistic, and would avoid this extra conversion. => `open(..., "bw+")`

philipwinklerdelmic:
done

PR #1127, opened 2019-07-23
dataio tiff: remove ensure_fn_encoding function

philipwinklerdelmic:


pieleric:
Indeed, the latest libtiff already takes care of it, with Python 3 support (actually I wrote that code, but forgot about it ;-) ). So the Odemis' code is not needed anymore.

PR #1130, opened 2019-07-23
driver andorcam2: support disabling the cooler also on cameras which only accept temperatures < 20°C

pieleric:


src/odemis/driver/andorcam2.py
```diff
@@ -538,12 +538,15 @@ def __init__(self, name, role, device=None, emgains=None, shutter_times=None,
...
-                ranges = (-275, 100)
-            self.targetTemperature = model.FloatContinuous(ranges[0], ranges, unit=u"°C",
+                trange = (-275, 25)
+            self._hw_temp_range = trange
+            # Always support 25°C, to disable the cooling
+            trange = (trange[0], max(trange[1], 25))
+            self.targetTemperature = model.FloatContinuous(trange[0], trange, unit=u"°C",
                                                            setter=self._setTargetTemperature)
-            self._setTargetTemperature(ranges[0], force=True)
+            self._setTargetTemperature(trange[0], force=True)
```
tepals: Is this still necessary? I thought the value is set when you create the FloatContinuous VA

src/odemis/driver/andorcam2.py
```diff
@@ -2679,7 +2691,7 @@ def GetTemperatureRange(self, p_mint, p_maxt):
         mint = _deref(p_mint, c_int)
         maxt = _deref(p_maxt, c_int)
         mint.value = -200
-        maxt.value = 50
+        maxt.value = -10
```
tepals: why did you change this?

pieleric: To be simulate more properly real hardware. Typically, the warmest cooling temperature is below 0... not 50°C. That was the reason that we didn't see with the simulator the bug that this PR fixes.

src/odemis/driver/andorcam2.py
pieleric: It is. Actually, the initial value of a VA *is not* passed to the setter. That's on purpose, because in many cases, the VA is set with the current value (of the hardware), so it doesn't make sense to call the setter immediately.

PR #1132, opened 2019-07-23
python3 migration: util img

philipwinklerdelmic:


src/odemis/gui/util/img.py
```diff
@@ -1091,7 +1091,7 @@ def val_y_to_pos_y(val_y, client_size, range_y):
 def bar_plot(ctx, data, range_x, range_y, client_size, fill_colour):
     """ Do a bar plot of the current `_data` """
 
-    if len(data) < 2:
+    if len(list(data)) < 2:
```
pieleric: I think you'd better indicate that `data` should be a list (or ndarray), and fix the only caller which passes an iterator (in `chronogram_to_export_data()`).

philipwinklerdelmic:
done

PR #1133, opened 2019-07-23
fix small dependencies handling

pieleric:


philipwinklerdelmic:
Looks good, ready to merge.

However, I still get some exceptions when stopping odemis, like this:
```
  File "/usr/lib/python2.7/dist-packages/concurrent/futures/_base.py", line 300, in _invoke_callbacks
    callback(self)
  File "/home/winkler/development/Pyro4/src/Pyro4/core.py", line 980, in on_future_completion
    client_future.set_exception(exc_value)
  File "/home/winkler/development/Pyro4/src/Pyro4/core.py", line 165, in __call__
    return self.__send(self.__name, args, kwargs)
  File "/home/winkler/development/Pyro4/src/Pyro4/core.py", line 299, in _pyroInvoke
    self.__pyroCreateConnection()
  File "/home/winkler/development/Pyro4/src/Pyro4/core.py", line 405, in __pyroCreateConnection
    raise errors.CommunicationError(err)
```

This also seems to be the reason why the integration tests always fail.

Do you have any idea on how to fix it?

pieleric:
I don't know where these extra errors are coming from. It's worth investigating (outside of this PR). It might be related to the power_supplier which should be considered as a dependency too. It might be something different.

PR #1134, opened 2019-07-24
Spectrum zoom fixes

amuskens:
Fixed:
* For the vertical legend, I'd suggest to change a little bit the rule
for displaying the "...". The problem currently is that it makes the
text really wide, and especially it tends to change when the character
is present or not (so the whole graph shifts). Instead of
appending/prepending the character, I'd suggest to just show that text
(and drop the original tick text).
 * As soon as an axis is zoomed-in, the scale-lock should be activated.
Currently, if you zoom on a live stream, pretty much nothing happens,
because the scale automatically unzooms on the next data received. You
can notice it for instance with a sparc backend, when you play the live
spectrum (in the acquisition tab). If you zoom in, it automatically
zooms out.
 * When dragging the mouse with the left button, the marking line
overlay doesn't follow the mouse. The line is only updated when the
mouse button is released.

Added one more fix: 
* the ticklist for the y display goes underneath the lock button, so added an offset to make sure that it displays correctly.

### DIFF
--- a/src/odemis/gui/comp/legend.py
+++ b/src/odemis/gui/comp/legend.py
@@ -418,12 +418,15 @@ class AxisLegend(wx.Panel):
             return
 
         # shared function with the export method
-        self._tick_list, self._vtp_ratio = calculate_ticks(self._value_range, self.ClientSize, self._orientation, self._tick_spacing)
         csize = self.ClientSize
+        self._tick_list, self._vtp_ratio = calculate_ticks(self._value_range, csize, self._orientation, self._tick_spacing)
 
-        if self.lock_va is not None and self.lock_va.value == False:
+        if self.lock_va is not None:
             if self._orientation == wx.HORIZONTAL:
                 csize.x -= 24
+
+            elif self._orientation == wx.VERTICAL:
+                self._tick_list = [(pos, val) for (pos, val) in self._tick_list if pos > 24]
 
         # If min and max are very close, we need more significant numbers to
         # ensure the values displayed are different (ex 17999 -> 18003)
--- a/src/odemis/gui/comp/legend.py
+++ b/src/odemis/gui/comp/legend.py
@@ -451,10 +451,10 @@ class AxisLegend(wx.Panel):
             label = units.readable_str(val, self.unit, sig)
 
             if i == 0 and self._lo_ellipsis:
-                label = u"…" + label
+                label = u"…"
 
             if i == len(self._tick_list) - 1 and self._hi_ellipsis:
-                label = label + u"…"
+                label = u"…"
 
             _, _, lbl_width, lbl_height, _, _ = ctx.text_extents(label)
 
--- a/src/odemis/gui/comp/viewport.py
+++ b/src/odemis/gui/comp/viewport.py
@@ -1255,6 +1255,10 @@ class NavigablePlotViewport(PlotViewport):
             self.bottom_legend.lo_ellipsis = self.canvas.display_xrange[0] > self.canvas.data_xrange[0]
             self.bottom_legend.hi_ellipsis = self.canvas.display_xrange[1] < self.canvas.data_xrange[1]
 
+        # automatically disable autoscaling (by locking) when the user zooms in
+        if self.canvas.display_xrange != self.canvas.data_xrange:
+            self.hrange_lock.value = True
+
     def set_vrange(self, vrange):
         """
         Setter for VA vrange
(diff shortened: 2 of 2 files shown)


src/odemis/gui/comp/viewport.py
```diff
@@ -1255,6 +1255,11 @@ def on_hrange(self, hrange):
             self.bottom_legend.lo_ellipsis = self.canvas.display_xrange[0] > self.canvas.data_xrange[0]
             self.bottom_legend.hi_ellipsis = self.canvas.display_xrange[1] < self.canvas.data_xrange[1]
 
+        if self.canvas.display_xrange != self.canvas.data_xrange:
+            self.hrange_lock.value = True
+        else:
+            self.hrange_lock.value = False
```
pieleric: Please add a comment to explain the goal of this code. Something like "Disable auto-scale as soon as the user manually zooms in".

More importantly, I don't think this does the right thing. It now also automatically re-enable auto-scale when zooming out. I don't think we should do that. Just let the user unlock it explicitly. 
(to be done both here, and in the vrange)

src/odemis/gui/comp/legend.py
```diff
@@ -415,13 +415,16 @@ def on_paint(self, _):
         if self._value_range is None:
             return
 
-        # shared function with the export method
-        self._tick_list, self._vtp_ratio = calculate_ticks(self._value_range, self.ClientSize, self._orientation, self._tick_spacing)
+        offset_y = 0
```
pieleric: It used to be 10, now you're setting it to 0 by default. I think that's wrong.

src/odemis/gui/comp/legend.py
```diff
@@ -415,13 +415,16 @@ def on_paint(self, _):
...
         csize = self.ClientSize
-
-        if self.lock_va is not None and self.lock_va.value == False:
+        if self.lock_va is not None and self.lock_va.value == True:
             if self._orientation == wx.HORIZONTAL:
                 csize.x -= 24
+            elif self._orientation == wx.VERTICAL:
+                offset_y = 24
```
pieleric: Shouldn't csize also be reduced by 24?

src/odemis/gui/comp/legend.py
```diff
@@ -415,13 +415,16 @@ def on_paint(self, _):
...
+
+        # shared function with the export method
+        self._tick_list, self._vtp_ratio = calculate_ticks(self._value_range, self.ClientSize, self._orientation, self._tick_spacing, start_y=offset_y)
```
pieleric: self.ClientSize -> csize? (I'm not sure, check)

src/odemis/gui/comp/viewport.py
```diff
@@ -1193,7 +1193,7 @@ def __init__(self, *args, **kwargs):
 
         self.canvas.Bind(wx.EVT_MIDDLE_DOWN, self.on_drag_start)
         self.canvas.Bind(wx.EVT_MIDDLE_UP, self.on_drag_end)
-        # self.canvas.Bind(wx.EVT_MOTION, self.on_canvas_motion)
+        self.canvas.Bind(wx.EVT_MOTION, self.on_canvas_motion)
```
pieleric: Would you mind not commenting out this line in the previous commit, so that this commit doesn't have to uncomment it? It'll be easier to backport....

src/odemis/gui/comp/legend.py
amuskens: It gets set now in the default argument for the offset

src/odemis/gui/comp/legend.py
amuskens: Yup. It is.

src/odemis/gui/comp/viewport.py
amuskens: Hmm..... how do I do this.... I think the commented-out line was a mistake. Maybe it is easier to back-port all of these changes at once?

src/odemis/gui/comp/legend.py
pieleric: Then pass `csize`?

BTW, I think we should change a little bit the code to read like `csize = self.GetClientSize()`. This is exactly equivalent. However, I find it confusing to read, as setting the .ClientSize actually changes the size of the widget. It's not clear that `csize.x += 24` does *not* change the size of the widget.

src/odemis/gui/comp/viewport.py
pieleric: I can do it with you, and show you. Essentially, you'd discard the commit history, and recreate each commit, using "git add -i".

PR #1136, opened 2019-07-25
bug fix: overflow of images displayed in gui and export

Sabrina85:


pieleric:
Merged, with an updated description which indicates that the overflow was not in the display, but in the *acquisition*.

PR #1137, opened 2019-07-25
introduce to_str_escape

pieleric:


src/odemis/util/__init__.py
```diff
@@ -102,6 +102,40 @@ def almost_equal(a, b, atol=1e-18, rtol=1e-7):
     return False
 
 
+if sys.version_info[0] < 3:
+    def to_str_escape(s):
+        """
+        Escapes the given string (or unicode string) in such a way that all the
+        non-displayable characters converted to "\\??". It's possible to use that string
```
philipwinklerdelmic: Does this include characters like "äßé"? I thought that we actually want to display them in the logging message and **not** escape them

In python3 with bytes
```
In [12]: s = "äßé".encode("latin1")

In [13]: s.decode("latin1", "unicode_escape")
Out[13]: 'äßé'

In [15]: s.decode("latin1").encode("unicode_escape").decode("ascii")
Out[15]: '\\xe4\\xdf\\xe9'
```

pieleric: Honestly, both is fine. For the current use, in the drivers, escaping these characters make more sense because the device is very unlikely to generate such characters as part of an actual (latin-characters-based) language, so it's more likely that looking at the hexadecimal is what we need.

PR #1139, opened 2019-07-29
acq stitching: weaver bugfix

philipwinklerdelmic:


src/odemis/acq/stitching/_weaver.py
```diff
@@ -331,19 +331,10 @@ def getFullImage(self):
...
-                x = numpy.arange(-hw, hw, 1)
-            else:
-                x = numpy.arange(-hw, hw + 1, 1)
-
-            if sz[0] % 2 == 0:
-                y = numpy.arange(-hh, hh, 1)
-            else:
-                y = numpy.arange(-hh, hh + 1, 1)
-
+            hh, hw = numpy.divide(sz, 2)  # half-height, half-width
```
pieleric: As `numpy.divide()` gives different values on Python 2 vs 3. That still makes me a little nervous. I'd suggest to use this code, to ensure it always do the same thing:
```
sz = roi.shape
hh, hw = sz[0] / 2, sz[1] / 2
```
or
```
sz = numpy.array(roi.shape)
hh, hw = sz / 2
```

philipwinklerdelmic:
updated

PR #1141, opened 2019-07-29
opencv3 compatibility

philipwinklerdelmic:


src/odemis/acq/align/keypoint.py
```diff
@@ -30,6 +30,11 @@
 from odemis import model
 from odemis.util import img
 
+if int(cv2.__version__[0]) <= 2:
```
pieleric: For info, this code will break starting from OpenCV v20.0 ... but hopefully by that time this code will have long been gone!
Officially, the right way to compare version is something like:
`pkg_resources.parse_version(cv2.__version__) < pkg_resources.parse_version("3.0")`

PR #1143, opened 2019-07-29
driver tlfw: to_str_escape

philipwinklerdelmic:
it shouldn't be necessary to decode the string when using to_str_escape

### DIFF
--- a/src/odemis/driver/tlfw.py
+++ b/src/odemis/driver/tlfw.py
@@ -253,7 +253,7 @@ class FW102c(model.Actuator):
         return (str): model name as reported by the device
         """
         # answer is like "THORLABS FW102C/FW212C Filter Wheel version 1.04"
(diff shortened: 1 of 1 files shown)


src/odemis/driver/tlfw.py
```diff
@@ -253,7 +253,7 @@ def GetIdentification(self):
         return (str): model name as reported by the device
         """
         # answer is like "THORLABS FW102C/FW212C Filter Wheel version 1.04"
-        return to_str_escape(self._sendQuery(b"*idn?").decode("latin1")
+        return to_str_escape(self._sendQuery(b"*idn?"))
```
pieleric: No, that's a wrapper to the code, you want to return the value, as received. So leave the `decode("latin1")`, and remove `to_str_escape()`.

PR #1149, opened 2019-07-30
python3 migration: driver scpi

philipwinklerdelmic:


src/odemis/driver/scpi.py
```diff
@@ -219,22 +219,22 @@ def _sendQuery(self, cmd, timeout=1):
...
                 if not char:
-                    raise IOError("Timeout after receiving %s" % ans.encode('string_escape'))
+                    raise IOError("Timeout after receiving %s" % to_str_escape(ans))
                 ans += char
 
-            logging.debug("Received answer %s", ans.encode('string_escape'))
+            logging.debug("Received answer %s", to_str_escape(ans))
 
-            return ans.rstrip()
+            return to_str_escape(ans.rstrip())
```
pieleric: No, `to_str_escape()` is only to be used for displaying string to the user. Use `decode("latin1")` here, to get a (unicode) string.

philipwinklerdelmic:
updated

PR #1150, opened 2019-07-31
python3 migration: cli main

philipwinklerdelmic:


src/odemis/cli/main.py
```diff
@@ -1015,7 +1015,7 @@ def main(args):
                 dataflows = ["data"]
             else:
                 dataflows = options.acquire[1:]
-            filename = options.output.decode(sys.getfilesystemencoding())
+            filename = to_str_escape(options.output)
```
pieleric: You *cannot* escape the string. It's going to add `\` at every non-ascii character. What's wrong with the original code? Is it just because it doesn't work with Python 3? If so, I think you should do the same as (I've implemented) in the pylibtiff.open() code:
https://github.com/pearu/pylibtiff/blob/master/libtiff/libtiff_ctypes.py#L500

It sounds also like there should be a test case that use this code, to check it work. For example, request to save a file as `TÈßt.tiff", and check that a file named like this is created.

philipwinklerdelmic: * I updated test_acquire to use TÈßt.tiff as picture name instead of test.tiff
* for the filename I decode based on the whether it's unicode
```
            if isinstance(options.output, unicode):  # python3
                filename = options.output
            else:  # python2
                filename = options.output.decode(sys.getfilesystemencoding())
```

src/odemis/cli/main.py
```diff
@@ -26,6 +26,7 @@
 
 from past.builtins import basestring
 from builtins import str
+from past.builtins import unicode
```
pieleric: Please merge with the other line of `past.builtins`.

src/odemis/cli/test/main_test.py
```diff
@@ -313,7 +309,7 @@ def test_stop(self):
         self.assertEqual(ret, 0, "trying to run '%s'" % cmdline)
     
     def test_acquire(self):
-        picture_name = "test.tiff"
+        picture_name = "TÈßt.tiff"
```
pieleric: Could you add a comment like "utf-8 bytes on Python 2, unicode on Python 3", to make the trick clear.

philipwinklerdelmic:
updated

PR #1152, opened 2019-07-31
skip tests that are known to fail

philipwinklerdelmic:


src/odemis/acq/align/test/keypoint_test.py
```diff
@@ -78,8 +78,10 @@ def test_image_pair(self):
         sem_img = preprocess(sem_img, image_pair[1][1], image_pair[1][2],
                              image_pair[1][3], image_pair[1][4], True)
         # execute the algorithm to find the transform between the images
-        tmat, _, _, _, _ = keypoint.FindTransform(tem_img, sem_img)
-
+        try:
```
philipwinklerdelmic: We might also use the unittest.expectedfailure here. However, I thought it might be a good idea to only skip the test in case of a ValueError, so if anything else goes wrong, we will be notified.

src/odemis/acq/align/test/keypoint_test.py
```diff
@@ -78,8 +78,10 @@ def test_image_pair(self):
...
+            tmat, _, _, _, _ = keypoint.FindTransform(tem_img, sem_img)
+        except ValueError:
+            self.skipTest("Test only works with SIFT, not with ORB.")
```
pieleric: I think that in that exception, you still need to check whether SIFT is available (using `hasattr(cv2....)`), and *only if* SIFT is not available, then skip (ortherwise, raise an assertError).

src/odemis/dataio/test/catmaid_test.py
```diff
@@ -56,7 +57,11 @@ def test_open_data_virtualflybrain(self):
         """
         # test for a url with a specified project id and stack id
         url = 'catmaids://fafb.catmaid.virtualflybrain.org/?pid=2&sid0=2'
-        acquisition = open_data(url)
+        try:
+            acquisition = open_data(url)
+        except ValueError:
+            logging.error("No stack info at url %s." % url)
+            self.skipTest("No stack info at url %s." % url)
```
pieleric: This looks like the URL is a dead link. Can you check with @tepals  for another server to use for testing?

philipwinklerdelmic:
I updated the keypoint test. The gui failures are hard to reproduce, they only occur when running the runtests script (running them from the command line seems fine). I'll leave them out for now.

PR #1154, opened 2019-07-31
python3 migration: diverse fixes

philipwinklerdelmic:


src/odemis/util/__init__.py
```diff
@@ -148,7 +148,7 @@ def recursive_dict_update(d, other):
 
     """
 
-    for k, v in other.iteritems():
+    for k, v in iter(other.items()):
```
pieleric: Drop the `iter()` entirely.

src/odemis/gui/util/test/updater_test.py
```diff
@@ -39,7 +39,7 @@ def test_version(self):
         rv = u.get_remote_version()
         # Note: if no internet access, it would be acceptable to get None
 
-        self.assertIn(len(rv.split('.')), (3, 4))
+        self.assertIn(len(rv.split(b'.')), (3, 4))
```
pieleric: I think here the test case is correct, but `get_remote_version()` should return a (unicode) string.

philipwinklerdelmic:
updated

PR #1164, opened 2019-08-01
config: update/fix dependencies

pieleric:


requirements.txt
```diff
@@ -5,11 +5,11 @@ futures==3.0.5
 h5py==2.5.0
 libtiff==0.4.2
 matplotlib==1.4.3
-numpy==1.9.2
+numpy==1.11.3
 opencv-python==2.4.11
```
philipwinklerdelmic: Should we change opencv to 3.2?

pieleric: Eventually, yes. For now OpenCV 2.4 is still the one known to work fine, so let's keep it as-is. Once we have tested a little bit more with OpenCV 3.2 (also on Python 2), we can update. It still has to work with OpenCV 2.4 for a while, as it's the one present in Ubuntu 16.04...

PR #1174, opened 2019-08-06
Background correction temporal spectrum

Sabrina85:


src/odemis/gui/cont/tabs.py
```diff
@@ -1840,7 +1841,21 @@ def display_new_data(self, filename, data, extend=False):
...
+        self.spectrum_streams = []
+        self.timecorrelator_streams = []
+        # sort spec streams
+        for s in spec_streams:
+            if hasattr(s, "selected_wavelength") and hasattr(s, "selected_time"):
+                self.temporalspectrum_streams.append(s)
+            elif hasattr(s, "selected_wavelength") and not hasattr(s, "selected_time"):
+                self.spectrum_streams.append(s)
+            elif not hasattr(s, "selected_wavelength") and hasattr(s, "selected_time"):
+                self.timecorrelator_streams.append(s)
```
pieleric: Currently, selected_wavelength is always created, so all the timecorrelator streams are concidered temporal spectrum, right?
Maybe just add a TODO for now, and later change the SpectrumStaticStream.

src/odemis/gui/cont/tabs.py
```diff
@@ -1840,7 +1841,21 @@ def display_new_data(self, filename, data, extend=False):
         all_streams = streams + self.tab_data_model.streams.value
 
         # Spectrum and AR streams are, for now, considered mutually exclusive
+
+        # all spectrum streams including spectrum, temporal spectrum and time correlator (chronograph)
         spec_streams = [s for s in all_streams if isinstance(s, acqstream.SpectrumStream)]
+        self.temporalspectrum_streams = []
+        self.spectrum_streams = []
+        self.timecorrelator_streams = []
```
pieleric: No, please, don't store the stream in variable. It's kind of handy, but now, the streams have another point of reference, so they will never stay in memory until the lists are reset (by calling this method again).
It's safer to just separate all the 3 types of streams every time. If you need to do it often, you could put the code in a method called like `_get_time_spectrum_streams()`, which returns the 3 lists.

src/odemis/util/dataio.py
```diff
@@ -65,12 +65,11 @@ def data_to_static_streams(data):
...
             (ci >= 0 and d.shape[ci] >= 5)  # No metadata, but looks like a spectrum
            ):
-            if (MD_TIME_LIST in d.metadata and
-                (ti >= 0 and d.shape[ti] > 1)):
+            if (MD_TIME_LIST in d.metadata and (ti >= 0 and d.shape[ti] > 1)):
                 # Streak camera data. Create a temporal spectrum
                 name = d.metadata.get(model.MD_DESCRIPTION, "Temporal Spectrum")
                 klass = stream.StaticSpectrumStream
-            else:
+            else:  # TODO this is also temporal spectrum data acquired with mirror and focus mode (so no time/wl info)
```
pieleric: Indeed, we might want to drop the check for the presence of MD_TIME_LIST and/or MD_WL_LIST. Anyway, when we have data of CT1YX, I guess the temporal-spectrum views are the best to use.

Sabrina85: Could we not drop it now? It looks like the only difference is the MD_DESCRIPTION...

Sabrina85:
TODO:

- move testcases
- stored background corr widgets, but file, path info etc is duplicated

src/odemis/acq/test/stream_test.py
```diff
@@ -4497,155 +4497,330 @@ def test_spec_1d(self):
...
+        # Check it's a RGB DataArray
+        self.assertEqual(im2d_effcorr.shape, temporalspectrum.shape[-2:] + (3,))
+        # check image different from previous image after efficiency correction
+        self.assertTrue(numpy.any(im2d_effcorr != prev_im2d))
+
+        # apply bg correction
+        tss.background.value = bckg
+
+        time.sleep(0.5)
+        im2d_bgcorr = proj_spatial.image.value
```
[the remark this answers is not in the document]
Sabrina85: So moving all the background and calibration file tests? That would mean, we should also move test regarding this for spectrum, ar etc...?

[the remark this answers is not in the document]
Sabrina85: I spitted the tests into 2, as I would always need to recreate the data and the stream. The first one checks for calibration file and bg file and also checks they do something different when applied. So don't want to split that test. The second one is testing for invalid files, which I think is ok to have in one test. Or would you still split them up?

src/odemis/gui/cont/settings.py
```diff
@@ -680,23 +722,60 @@ def _on_ar_cal(self, val):
...
+            self._temporalspec_bckfile_ctrl.Bind(EVT_FILE_SELECT, self._on_temporalspec_bck_file_select)
+            self.tab_data.temporalspec_bck_cal.subscribe(self._on_temporalspec_bck_cal, init=True)
+
+        if spectrum or temporalspectrum:
+            self._specfile_ctrl = self._pnl_acqfile.add_file_button(
+                "Spectrum correction",
+                tooltip="Spectrum efficiency correction file",
+                clearlabel="None", wildcard=self.wildcards).value_ctrl
+            self._specfile_ctrl.Bind(EVT_FILE_SELECT, self._on_spec_file_select)
+            self.tab_data.spec_cal.subscribe(self._on_spec_cal, init=True)
```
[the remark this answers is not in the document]
Sabrina85: I tried this, however, the problem is, that `self._pnl_acqfile.panel.clear_all()` is called whenever a new file is loaded (`on_acqfile_change`), which results in the destruction of the controls. That makes sense for the File name etc. but not for the bg and calib controls. So I now create a separate panel for the bg and calib files. But I create the controls in the init (`_create_controls`) as suggested and then just show and hide them as needed. What do you think?

src/odemis/acq/calibration.py
```diff
@@ -179,47 +237,101 @@ def get_spectrum_efficiency(das):
     return ret
 
 
-def compensate_spectrum_efficiency(data, bckg=None, coef=None):
+def spectrum_efficiency_and_bg_correction(data, bckg=None, coef=None):
```
pieleric: It's not really a great name. Firstly, functions should start with a verb, and secondly it's a little too long. However, I don't have great idea either... Maybe just `adjust_spectrum` or `compensate_spectrum`? I don't think the name absolutely needs to indicate that it uses the efficiency coefficient and the background, because these are in the arguments of the function.

src/odemis/gui/cont/settings.py
```diff
@@ -558,35 +559,57 @@ def _create_controls(self):
...
-        self._pnl_arfile = FileInfoSettingsController(self.tab_panel.fp_fileinfo, "")
-        self._arfile_ctrl = self._pnl_arfile.add_file_button(
+        self._arfile_ctrl = self._pnl_calibration.add_file_button(
             "AR background",
             tooltip="Angle-resolved background acquisition file",
             clearlabel="None", wildcard=wildcards).value_ctrl
-        self._pnl_arfile.hide_panel()
+        self._ar_bckfile_entry = self._pnl_calibration.entries[0]
+        self._ar_bckfile_entry.lbl_ctrl.Show(False)
+        self._ar_bckfile_entry.value_ctrl.Show(False)
```
pieleric: The function `add_file_button()` returns the entry. So no need to guess try to find it again via the `_pnl_calibration.entries`. You should be able to get rid of `._arfile_ctrl`, and replace it everywhere by `._ar_bckfile_entry.value_ctrl`.

(Same thing for the next three file buttons).

src/odemis/gui/cont/settings.py
```diff
@@ -656,6 +680,24 @@ def _on_spec_bck_file_select(self, evt):
 
         self.tab_data.spec_bck_cal.value = fn
 
+    def _on_temporalspec_bck_file_select(self, evt):
+        """ Pass the selected spec background file on to the VA """
+        logging.debug("Temoral spectrum background file selected by user")
```
pieleric: Temoral -> Temporal

src/odemis/acq/calibration.py
Sabrina85: How about `apply_corrections`?

src/odemis/gui/cont/settings.py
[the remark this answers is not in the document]
Sabrina85: Can I mark this resolved?

src/odemis/util/dataio.py
Sabrina85: Should I implement it in a separate pr or leave as it is for now?

src/odemis/gui/cont/settings.py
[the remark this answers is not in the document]
pieleric: Yes. I think this is still essentially what you've implemented. At least, a new widget is not created for every new file :-)

src/odemis/util/dataio.py
pieleric: Leave this as-is, with the TODO.

src/odemis/acq/calibration.py
```diff
@@ -237,7 +237,7 @@ def get_spectrum_efficiency(das):
     return ret
 
 
-def spectrum_efficiency_and_bg_correction(data, bckg=None, coef=None):
+def apply_corrections(data, bckg=None, coef=None):
```
pieleric: Well, the name should indicate it processes spectra. => `apply_spectrum_corrections()`?

src/odemis/gui/cont/settings.py
```diff
@@ -637,8 +628,8 @@ def on_acqfile_change(self, file_info):
                 self._pnl_acqfile.add_metadata(key, value)
 
             # Change default dir for the calibration files
-            for file_ctrl in (self._arfile_ctrl, self._spec_bckfile_ctrl,
-                              self._temporalspec_bckfile_ctrl, self._specfile_ctrl):
+            for file_ctrl in (self._ar_bckfile_entry.value_ctrl, self._spec_bckfile_entry.value_ctrl,
+                              self._temporalspec_bckfile_entry.value_ctrl, self._specfile_entry.value_ctrl):
                 file_ctrl.default_dir = file_info.file_path
```
pieleric: The point of this loop is to avoid repeating the code, so you could write more concisely:
```
            for file_entry in (self._ar_bckfile_entry., self._spec_bckfile_entry,
                              self._temporalspec_bckfile_entry, self._specfile_entry):
                file_entry.value_ctrl.default_dir = file_info.file_path
```

src/odemis/acq/calibration.py
Sabrina85: My intention was to keep it generic in case we support time correlator corrections, but as we do not for now I changed it  accordingly.

PR #1179, opened 2019-08-12
Smaract API testing version (for review)

amuskens:


src/odemis/driver/smarpod.py
```diff
@@ -400,15 +647,15 @@ def IsReferenced(self):
...
+
     def GetMoveStatus(self):
         """
         Gets the move status from the controller.
         Returns:
             SmarPodDLL.SMARPOD_MOVING is returned if moving
             SmarPodDLL.SMARPOD_STOPPED when stopped
             SmarPodDLL.SMARPOD_HOLDING when holding between moves
-            SmarPodDLL.SMARPOD_CALIBRATING when calibrating
+            SmarPodDLL.SMARPOD_CALIBRATING wsmarpod.SmarAct_MC(**CONFIG)hen calibrating
```
tepals: copy paste error

src/odemis/driver/smarpod.py
```diff
@@ -668,7 +915,7 @@ def _cancelCurrentMove(self, future):
         # The difficulty is to synchronise correctly when:
         #  * the task is just starting (not finished requesting axes to move)
         #  * the task is finishing (about to say that it finished successfully)
-        logging.debug("Canceling current move")
+        logging.debug("Canceling current smarpod.SmarAct_MC(**CONFIG)move")
```
tepals: copy paste error, you might want to check if there are more

src/odemis/driver/smarpod.py
```diff
@@ -279,9 +279,256 @@ def __init__(self, error_code):
...
+    Convert a coordinate dictionary (str) -> (double) to a SA_MC Pose C struct
+        pos: Coordinate dictionary (str) -> (double) of axis name to value
+    returns: a Pose (C struct)
+    raises ValueError if an unsupported axis name is input
+    """
+    pose = SA_MC_Pose()
+
+    # Note: internally, the system uses metres and degrees for rotation
+    for an, v in pos.items():
+        if an == "x":
```
tepals: couldn't you rewrite this whole for loop to:
```
pose.x = pos["x"]
pose.y = pos["y"]
pose.z = pos["z"]
pose.rx = math.degrees(pos["rx"])
pose.ry = math.degrees(pos["ry"])
pose.rz = math.degrees(pos["rz"])
```

amuskens: I think I did it this way because it is possible to have a pos dict with not all of the axes i.e. pos = {'x': 0.5}

scripts/acquire_cryo.py
```diff
@@ -0,0 +1,121 @@
...
+    return (int): value to return to the OS as program exit code
+    """
+    # Get components
+    # stage = model.getComponent(role="scan-stage")
+    #scanner = model.getComponent(role="laser-mirror")
+    """
+
+    # parse arguments
+    parser = argparse.ArgumentParser(description="Run an acquisition "
+                                     "with Cryo hardware for testing")
```
pieleric: I don't quite understand what does this script (better than just `odemis-cli --acquire se-detector ...`), but in any case it should be in a separate PR from the extension of the SmarAct driver.

src/odemis/driver/smarpod.py
```diff
@@ -279,9 +279,256 @@ def __init__(self, error_code):
...
+        except Exception:
+            raise AttributeError("Failed to find %s" % (name,))
+        func.__name__ = name
+        func.errcheck = self.sp_errcheck
+        return func
+
+    @staticmethod
+    def sp_errcheck(result, func, args):
+        """
+        Analyse the retuhwModelrn value of a call and raise an exception in case of
```
pieleric: `hwModel` got dropped in the middle of the string.

src/odemis/driver/mcs2.py
```diff
@@ -0,0 +1,999 @@
...
+    def __init__(self, error_code):
+        self.errno = error_code
+        super(SA_CTLError, self).__init__("Error %d. %s" % (error_code, SA_CTLDLL.err_code.get(error_code, "")))
+
+
+class SmarAct_MCS2(model.Actuator):
+
+    def __init__(self, name, role, locator, ref_on_init=False,axes=None, **kwargs):
+        """
+        A driver for a SmarAct MCS2 Actuator.
```
pieleric: Please describe more here what is the "MCS2" actuator. I'd say it's "the standard SmarAct motion controller".

src/odemis/driver/smarpod.py
```diff
@@ -870,3 +1117,568 @@ def Smarpod_GetDLLVersion(self, p_major, p_minor, p_update):
         minor.value = 2
         update = _deref(p_update, c_uint)
         update.value = 3
+
+
+class SmarAct_MC(model.Actuator):
```
pieleric: Let's call this class "MCTriGlide".

src/odemis/driver/smarpod.py
```diff
@@ -870,3 +1117,568 @@ def Smarpod_GetDLLVersion(self, p_major, p_minor, p_update):
...
+
+    def __init__(self, name, role, locator, ref_on_init=False, linear_speed=0.01,
+                 rotary_speed=1, axes=None, **kwargs):
+        """
+        A driver for a SmarAct SA_MC Actuator.
```
pieleric: SA_MC -> MC. Please explain what the actuator is. That's the "Motion Controller" which formally was called "TriGlide", and is a 5DoF custom-designed stage.

src/odemis/driver/mcs2.py
```diff
@@ -0,0 +1,999 @@
...
+    def _updateSpeed(self):
+        """
+        update the speeds
+        """
+        s = {}
+        try:
+            for axis_name, axis_channel in self._axis_map.items():
+                s[axis_name] = self.self.GetSpeed(axis_channel)
+                
+        except SA_CTLError as ex:
```
pieleric: Drop this try/except if you do nothing with the exception. (same for all the similar functions)

src/odemis/driver/mcs2.py
```diff
@@ -0,0 +1,999 @@
...
+        self._checkMoveAbs(pos)
+        pos = self._applyInversion(pos)
+
+        f = self._createMoveFuture()
+        f = self._executor.submitf(f, self._doMoveAbs, f, pos)
+        return f
+
+    @isasync
+    def moveRel(self, shift):
+        self._checkMoveRel(shift)
```
pieleric: Make it like `moveAbs()`: immediately return if shift is empty.

src/odemis/driver/smarpod.py
pieleric: What happens to the axes where the pose is not defined? Do they move to a random position? Or back to 0? That's all sounds fishy...

src/odemis/driver/smarpod.py
```diff
@@ -279,9 +279,256 @@ def __init__(self, error_code):
...
+        }
+
+    def __init__(self):
+        if os.name == "nt":
+            raise NotImplemented("Windows not yet supported")
+            # WinDLL.__init__(self, "libSA_MC.dll")  # TODO check it works
+            # atmcd64d.dll on 64 bits
+        else:
+            # Global so that its sub-libraries can access it
+            CDLL.__init__(self, "libsmaractmc.so.0.9.1", RTLD_GLOBAL)
```
pieleric: It should be just `libsmaractmc.so.0` (which should be a link to the latest version of the library).

src/odemis/driver/smarpod.py
```diff
@@ -279,9 +279,256 @@ def __init__(self, error_code):
...
+
+    def __init__(self, error_code):
+        self.errno = error_code
+        super(SA_MCError, self).__init__("Error %d. %s" % (error_code, SA_MCDLL.err_code.get(error_code, "")))
+
+
 class SmarPod(model.Actuator):
     
-    def __init__(self, name, role, locator, ref_on_init=False, actuator_speed=0.1,
+    def __init__(self, name, role, locator, daemon=None, ref_on_init=False, actuator_speed=0.1,
```
pieleric: Don't get `daemon` here. It is normally a hidden data directly passed to the parent class (via kwargs). If you grab it here, the parent will never get it and be very sad.

src/odemis/driver/mcs2.py
```diff
@@ -0,0 +1,999 @@
...
+
+        options = c_char_p(option_string.encode("ascii"))
+
+        self.core.SA_CTL_Open(byref(self._id), options)
+        logging.debug("Successfully connected to SA_CTL Controller ID %d", self._id.value)
+        model.Actuator.__init__(self, name, role, axes=axes_def, **kwargs)
+
+        # Add metadata
+        # TODO: Fix getting software version with a supported function
+        self._swVersion = 0  # self.GetSwVersion()
```
pieleric: Use `SA_CTL_GetFullVersionString()`?

src/odemis/driver/mcs2.py
```diff
@@ -0,0 +1,999 @@
...
+
+        self._metadata[model.MD_SW_VERSION] = self._swVersion
+        logging.debug("Using SA_CTL library version %s", self._swVersion)
```
pieleric: Please also report the hwVersion, using SA_CTL_PKEY_FIRMWARE_VERSION.

src/odemis/driver/smarpod.py
```diff
@@ -870,3 +1117,568 @@ def Smarpod_GetDLLVersion(self, p_major, p_minor, p_update):
...
+        self.core.SA_MC_Open(byref(self._id), options)
+        logging.debug("Successfully connected to SA_MC Controller ID %d", self._id.value)
+        model.Actuator.__init__(self, name, role, axes=axes_def, **kwargs)
+
+        # Add metadata
+        # TODO: Fix getting software version with a supported function
+        self._swVersion = 0  # self.GetSwVersion()
+
+        self._metadata[model.MD_SW_VERSION] = self._swVersion
+        logging.debug("Using SA_MC library version %s", self._swVersion)
```
pieleric: Let's ask SmarAct to get this info. For now, just disable entirely the code, so that the default value ("unknown") is used.

pieleric:
BTW, to control the pivot point on the MCTriGlide class, the client shouldn't rely on directly calling the internal SetPivot()/GetPivot() methods. The class should provide a dedicated metadata, something like `MD_PIVOT_POS = {"x": ..., "y":..., "z":...)`. To support it, it should be mostly a matter of calling GetPivot() at init, and then override `updateMetadata()` to handle a change of value, by calling SetPivot().

src/odemis/driver/mcs2.py
```diff
@@ -0,0 +1,999 @@
...
+        self._updatePosition()
+
+    def _updatePosition(self):
+        """
+        update the position VA
+        """
+        p = {}
+        try:
+            for axis_name, axis_channel in self._axis_map.items():
+                p[axis_name] = self.self.GetPosition(axis_channel)
```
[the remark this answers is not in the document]
amuskens: Haha... luckily this is fixed now...

src/odemis/driver/smarpod.py
amuskens: Apparently, it's actually not a triglide. Any alternative ideas for the name?

src/odemis/driver/smarpod.py
amuskens: I think it will move to 0 as that would be the value initialized. This could be problematic. Maybe this needs to be accounted for?

src/odemis/driver/test/smatpod_test.py
```diff
@@ -42,8 +42,8 @@
         "role": "",
         "ref_on_init": True,
         "actuator_speed": 0.1,  # m/s
-        "locator": "usb:ix:0",
-        # "locator": "fake",
+        # "locator": "usb:ix:0",
+        "locator": "fake",
```
pieleric: Please use "usb:ix:0", and if `TEST_NOHW` is set then override "locator" to "fake".

PR #1181, opened 2019-08-21
resize spot grid live viewer 

tepals:
and image when the image from the camera is larger than the screen.

Before this change when the image was larger than the screen you would only see part of the image.


scripts/spot-grid.py
```diff
@@ -81,6 +81,18 @@ def new_image(self, data):
...
         self.app.img = NDImage2wxImage(rgb)
+        disp_size = wx.Display(0).GetGeometry().GetSize()
+        if disp_size[0] < rgb.shape[1] or disp_size[1] < rgb.shape[0]:
+            asp_ratio = rgb.shape[1] / rgb.shape[0]
+            if disp_size[0] / disp_size[1] < asp_ratio:
+                self.app.img.Rescale(disp_size[0], disp_size[0] / asp_ratio, rgb.shape[2])
+                self.app.magn = disp_size[0] / rgb.shape[1]
+            else:
+                self.app.img.Rescale(disp_size[1] * asp_ratio, disp_size[1], rgb.shape[2])
+                self.app.magn = disp_size[1] / rgb.shape[0]
```
pieleric: I'd suggest to avoid having duplicate call to `.Rescale()` by reodering the code like that:
```
if disp_size[0] / disp_size[1] < asp_ratio:
    self.app.magn = disp_size[0] / rgb.shape[1]
else:
    self.app.magn = disp_size[1] / rgb.shape[0]

self.app.img.Rescale(rgb_shape[1]*self.app.magn, rgb_shape[0]*self.app.magn, rgb.shape[2])
```

tepals:
updated

PR #1182, opened 2019-08-26
MBSEM: mppc to multiprobe alignment

Sabrina85:


scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,242 @@
...
+        logging.getLogger().setLevel(logging.WARNING)
+
+    if not (options.file_offset and options.file_rotation):
+        logging.error("Please specify an offset image and a rotation image for running the mppc to "
+                      "multiprobe alignment procedure.")
+        return
+
+    # read image file
+    try:
+        image_off = hdf5.read_data(options.file_offset)
```
tepals: the mppc outputs tiff files

scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,242 @@
...
+    try:
+        # TODO @Job/Andries are these steps independent of each other or not? If not how depending on each other?
+        #  Would need to run the script first on one correction and then do a new scan and run script for second corr..
+        #  If offset not corrected, would the spot position be additionally affected for rotation correction?
+
+        # 1. offset correction
+        # TODO pass only the D4 mppc cell image, all others need to be set black. How do I know where the D4 cell is?
+        # Later only the D4 cell should be active so nothing to worry about
+        mppc_offset = get_mppc_offset(image_offset)
+        print("Distance from image center is x={:.2f} and y={:.2f}. Apply corresponding voltage to the galvo mirror "
```
tepals: Since we're not converting to a voltage yet add the unit here (pixels I think?).
`...is x={:.2f} pixels and y={:.2f} pixels. ...`

scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,242 @@
...
+        If possible to activate all 4 cells in same scan, need multi spot detection.
+    Note: Image needs to be acquired on homogeneous non-sample area.
+    :param image: (ndarray) The full mppc detector image to run the calibration for.
+    :returns: (tuple of 2 float) The difference of the corner cells A8 to A1 in x and the difference
+                of H1 to A1 in y.
+    """
+    # Get the 4 corner mppc cell images: A1, H1, A8, H8
+    # Size of one mppc cell image is (800, 800) TODO @Job correct?
+    # Shape of full mppc is (8, 8)
+    image_A1 = image[0:800, 0:800]
```
tepals: To make it independent of the amount of pixels in a cell, I would change this to something like this:

```
cell_size_x = image.shape[0] / 8
cell_size_y = image.shape[1] / 8
image_A1 = image[0:cell_size_x, 0:cell_size_y]
...
...
...
```

This works because you know that there are only 8 cells

scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,242 @@
...
+    image_H1 = image[-801:-1, 0:800]
+    image_A8 = image[0:800, -801:-1]
+    image_H8 = image[-801:-1, -801:-1]
```
tepals: I think you made a mistake in the above 3 lines with the `[-801:-1, ...]` and it should be `[-800:, ...]`. If you do `:-1` it excludes the last element of the array

```
In: b = numpy.array([1, 2, 2, 3, 4, 5])
In: b[-2:]
Out[9]: array([4, 5])
In: b[-2:-1]
Out[10]: array([4])
```

scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,242 @@
...
+    delta_x1 = mppc_cell_spotcenter_A1[0] - mppc_cell_spotcenter_A8[0]  # pos: CCW rotation, neg: CW rotation
+    # compare H1 with H8: should have same x
+    delta_x2 = mppc_cell_spotcenter_H1[0] - mppc_cell_spotcenter_H8[0]
+
+    # check that values are the same
+    if not almost_equal(delta_x1, delta_x2):
+        raise ValueError
+
+    # calculate angle for rotation
+    l = 800 * 8  # length in one dimension of the full mppc image of all 8 cells
```
tepals: Nit-picky: a single l as a variable name reads really annoying. The next line looks like it could also be delta_y1 divided by one. Also isn't the length of the full mppc image just image.shape[0]

scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,242 @@
...
+    angle = numpy.arcsin(delta_y1/l)  # [rad]
+    mppc_rotation = angle * 360 / (2 * numpy.pi)  # convert to degree
```
tepals: `math.degrees(angle)`

scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,242 @@
...
+    # compare H1 with H8: should have same x
+    delta_x2 = mppc_cell_spotcenter_H1[0] - mppc_cell_spotcenter_H8[0]
+
+    # check that values are the same
+    if not almost_equal(delta_x1, delta_x2):
+        raise ValueError
+
+    # calculate angle for rotation
+    l = 800 * 8  # length in one dimension of the full mppc image of all 8 cells
+    angle = numpy.arcsin(delta_y1/l)  # [rad]
```
tepals: I think this should be:
```
delta_x3 = mppc_cell_spotcenter_A1[0] - mppc_cell_spotcenter_H1[0]
angle = numpy.arctan(delta_y1 / delta_x3)  # [rad]
```

scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,242 @@
...
+    mppc_rotation = angle * 360 / (2 * numpy.pi)  # convert to degree
+    # control
+    angle_cont = numpy.arccos(delta_x1/l)  # [rad]
```
tepals: And I think this should be:
```
delta_y3 = mppc_cell_spotcenter_A1[1] - mppc_cell_spotcenter_A8[1]
angle = numpy.arctan(delta_x1 / delta_y3)  # [rad]
```

Sabrina85: Is there a reason you choose this one? How about this one:

```suggestion
    delta_x4 = mppc_cell_spotcenter_A8[0] - mppc_cell_spotcenter_H8[0]
    angle_control = numpy.arctan(delta_y2 / delta_x4)  # [rad]
```

scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,247 @@
...
+              "to correct for the offset.".format(mppc_offset[0], mppc_offset[1]))
+
+        input('Press enter after finishing the rotation correction\n')
+
+        # calculate mppc rotation
+        mppc_rotation = get_mppc_rotation(corner_images)
+        print("Apply {:.3f}° to the MPPC rotation using the manual rotational axes of the MPPC"
+              " detector.".format(mppc_rotation))
+
+        input('Press enter after finishing the rotation correction\n')
```
tepals: is this necessary?

scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,247 @@
...
+        image = tiff.read_data(options.file)[0]
+        corner_images, image_D4 = preprocess_images(image)
+
+        # calculate mppc offset
+        mppc_offset = get_mppc_offset(image_D4)
+        print("Distance from image center is y={:.2f} pixels and x={:.2f} pixels. "
+              "Apply corresponding voltage to the galvo mirror "
+              "to correct for the offset.".format(mppc_offset[0], mppc_offset[1]))
+
+        input('Press enter after finishing the rotation correction\n')
```
tepals: this should be offset correction

scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,247 @@
...
+def preprocess_images(image):
+    """
+    Pre-processing the data to calculate the corrections. The size of the input image is well defined (6400px x 6400px).
+    The image is acquired with one beamlet (D4/(5,4)) and contains 64 cell images (8x8; 800px x 800px; 30um x 30um).
+    Each cell image shows the corresponding active mppc cell image as a bright square. The cell image covers a field
+    of view larger than all mppc cells. The mppc cell image (bright square) consists of a fixed size of about
+    3um x 3um (80px x 80px), which corresponds to about 1% of the area of a single cell image.
+    Depending on the cell image, the bright square is located in a different position of the cell image.
+    The 64 cell images are stiched together. A projection of all single cell images onto each other would result
+    in one image with all mppc cell images being bright. As the image is recorded using the D4 beamlet,
```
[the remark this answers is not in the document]
Sabrina85: In this type of acquisition, we scan with one beam over all cells. So when the first cell image (size which covers more then the full mppc chip) is recorded, only the top left corner mppc cell records some signal. Then when the beam moves onto the location of the next mppc cell, the second cell image is recorded. There the top left corner mppc cell is dark, but the second records a signal. All other mppc cells are also dark. By that the bright square in each cell image shifts its position.

scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,247 @@
...
+
+
+def preprocess_images(image):
+    """
+    Pre-processing the data to calculate the corrections. The size of the input image is well defined (6400px x 6400px).
+    The image is acquired with one beamlet (D4/(5,4)) and contains 64 cell images (8x8; 800px x 800px; 30um x 30um).
+    Each cell image shows the corresponding active mppc cell image as a bright square. The cell image covers a field
+    of view larger than all mppc cells. The mppc cell image (bright square) consists of a fixed size of about
+    3um x 3um (80px x 80px), which corresponds to about 1% of the area of a single cell image.
+    Depending on the cell image, the bright square is located in a different position of the cell image.
```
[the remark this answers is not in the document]
Sabrina85: I tried to rewrite it. It is a bit difficult, as the term single cell image as used during multibeam mode is a bit differently defined. We scan in this procedure a field of view that is larger than the mppc detector and also with one beam. Usually we would only scan with each beam the specified pitch size (3.2um).

scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,298 @@
...
+
+    # thresholding
+    # Known:
+    #   - size of single image (800px x 800px => 30um x 30um = 900 um2)
+    #   - size of active mppc cell image (bright square) approximately the same at all times
+    #     (3um x 3um = 9um2 => 1% of image)
+    # set 99% of the pixels with the lowest values to 0, set the rest of the pixels to 1
+    for num_img, img in enumerate(images):
+        shape = img.shape
+        img_histogram = numpy.bincount(numpy.ravel(img))
```
pieleric: Note that this only works if the dtype is uint8 or uint16. This is probably always the case, but you might want to add a comment about this. Otherwise, you'd need to use some more fancy function. see odemis.util.img.histogram()

BTW, you can just write: `numpy.bincount(img.flat)`.

scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,298 @@
...
+        index_array = numpy.cumsum(img_histogram) >= 0.99 * shape[0] * shape[1]  # bool array: True if 99% accumulated
+        threshold = index_array.argmax()  # find the first occurrence of True
```
pieleric: It should be possible to write a little bit more explicitly like this:
```
cum_hist = numpy.cumsum(img_histogram)
threshold = numpy.searchsorted(cum_hist, 0.99 * shape[0] * shape[1], side="left")
```

scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,298 @@
...
+        img[img <= threshold] = 0  # set 99% of px with the lowest values to 0
+        img[img > threshold] = 1  # set the remaining 1% to 1
```
pieleric: Maybe it's not so important, but it might be more efficient to just write `img = img > threshold`, which directly returns a black & white (boolean) image or `img = (img > threshold).astype(numpy.uint8)`, which returns a 8-bit greyscale image.

scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,298 @@
...
+        if num_labels != 1:
+            logging.info("There should be only one active area in the image. However, found {:f} labelled"
+                         "objects.".format(num_labels))
+        # Labelled object should have the size of about 1% of total image in px. Otherwise active area cut of or
+        # outside of the scanning area e.g. due to too much rotation of the mppc.
+        # Use a threshold of 90% of the expected size to allow for variation due to dilation/erosion operation.
+        active_area_found = False
+        # find the connected component
+        # Note: num_labels does not include any background values
+        for label in range(num_labels):
```
pieleric: Maybe write `range(1, num_labels + 1)`,  to avoid writing `label + 1` when you actually care about the label number.

scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,298 @@
...
+            # check that the size of the connected component is at least 10% of the size
+            # of the active area (about 1% of the total area)
+            # If not, it is either noise or only part of the active area was captured due to too much rotation
+            # (active area a edge and thus cut-off).
+            if len(img_cc[img_cc == label + 1]) < 0.1 * 0.01 * cell_size_y * cell_size_x:  # as 0 is background
```
pieleric: Maybe this is also more explicit:
`len(img_cc[img_cc == label + 1])` -> `numpy.count_nonzero(img_cc == label + 1)`

scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,298 @@
...
+                # mark connected component as background as it cannot be the active area and
+                # was missed in the other pre-processing steps
+                img_cc[img_cc == label + 1] = 0
+            else:
+                active_area_found = True
+
+        if active_area_found:
+            images[num_img] = img_cc
+        else:
+            raise ValueError("Active area in individual cell image is cut off or not existing. Please check size of"
```
pieleric: A space is missing after "of".
Maybe also add the `num_img` in the error message to hint more at where the issue is when it happens.

scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,298 @@
...
+            # check that the size of the connected component is at least 10% of the size
+            # of the active area (about 1% of the total area)
+            # If not, it is either noise or only part of the active area was captured due to too much rotation
+            # (active area a edge and thus cut-off).
+            if len(img_cc[img_cc == label + 1]) < 0.1 * 0.01 * cell_size_y * cell_size_x:  # as 0 is background
+                # mark connected component as background as it cannot be the active area and
+                # was missed in the other pre-processing steps
+                img_cc[img_cc == label + 1] = 0
+            else:
+                active_area_found = True
```
pieleric: Note that theoretically several areas could be larger than the minimum area. So img_cc could contain more than just 2 values. Maybe in such case report it, or as soon as you find a big-enough area, mark all the other (ie, next) areas to background?

scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,298 @@
...
+    Depending on which of the 64 cell images, the bright square is located in a different position of the image.
+    The 64 cell images are stitched together. A projection of all single cell images onto each other would result
+    in one image with all mppc cell images being bright. As the image is recorded using the E4 beamlet,
+    the E4 cell image should show the mppc cell image at the center of the image if correctly aligned.
+    The corner images A1, A8, H1 and H8 are used to correct for rotation, the E4 image is used to correct for an offset.
+    :param image: (DataArray) Image of size 6400px x 6400px made up by 64 cell images (800px x 800px), each containing
+                the corresponding mppc cell image (bright square) at a given position, which are shifted by the size
+                of the square in respect to each other.
+    :returns: (list of 4 ndarrays) The preprocessed corner images. Shape is (y, x).
+              (ndarray) The E4 cell image. Shape is (y, x).
```
pieleric: Let's be a little bit more specific on the data type. It's an uint8, which contains 0 for the "background" and a number (typically 1, but not necessarily) for the area. Right?

scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,298 @@
...
+    spotcenter_H8 = center_of_mass(images[3])  # tuple [px]
+
+    # compare A1 with H1: should have same y
+    delta_y_A1_H1 = spotcenter_A1[0] - spotcenter_H1[0]  # pos: CCW rotation, neg: CW rotation
+    # compare A8 with H8: should have same y
+    delta_y_A8_H8 = spotcenter_A8[0] - spotcenter_H8[0]
+
+    # check that values are the same, allow a 10px margin
+    if not almost_equal(delta_y_A1_H1, delta_y_A8_H8, atol=10):
+        raise ValueError("Difference in y position for spots found in mppc cells A1 and H1 should"
```
pieleric: Missing space after "should".

scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,298 @@
...
+
+    return images[0:4], images[4]
+
+
+def get_mppc_rotation(images):
+    """
+    Calculates the rotation of the MPPC detector in space. MPPC detector should be horizontally positioned.
+    The rotation angle is calculated based on the 4 corner cell images by evaluating trigonometric quantities.
+    :param images: (list of 4 ndarrays) Preprocessed corner cell images.
+    :returns: (float) The angle that should be applied to the mppc stage. This is the opposite (negative)
```
pieleric: I'd suggest to return the value in radians, as everywhere else in Odemis. For the *display*, you can convert it to degree.

scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,298 @@
...
+        # calculate mppc offset
+        mppc_offset = get_mppc_offset(image_E4, options.phi)
+        print("Distance of the E4 mppc cell from image center is y={:.2f} pixels and x={:.2f} pixels. "
+              "Apply corresponding voltage to the galvo mirror "
+              "to correct for the offset.".format(mppc_offset[0], mppc_offset[1]))
+
+        # calculate mppc rotation
+        mppc_rotation = get_mppc_rotation(corner_images)
+        print("Apply {:.2f}° to correct the MPPC position using the manual rotational mount of the MPPC"
+              " detector.".format(mppc_rotation))
```
pieleric: Convert to degrees only here.

scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,298 @@
...
+def get_mppc_offset(image, phi):
+    """
+    Calculates the offset of the MPPC detector in space. MPPC detector should be centered in regard to the scanning
+    beamlet, when in single beam mode.
+    Calculates the distance of the active MPPC cell image (bright square) in regard to the center of the cell image.
+    Note: Return value is based on top/left corner (0,0).
+    :param image: (ndarray) Preprocessed E4 cell images.
+    :param phi: (float) Rotation angle to transform the mppc image coordinate system to the multiprobe
+                coordinate system.
+    :returns: (tuple of 2 float) The distance to the E4 cell image center in y and x in px.
```
pieleric: Typically for positions, we use tuples of x,y (and not y,x). Please follow this convention, so that we don't get confused later on.

scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,298 @@
...
+
+
+def get_mppc_offset(image, phi):
+    """
+    Calculates the offset of the MPPC detector in space. MPPC detector should be centered in regard to the scanning
+    beamlet, when in single beam mode.
+    Calculates the distance of the active MPPC cell image (bright square) in regard to the center of the cell image.
+    Note: Return value is based on top/left corner (0,0).
+    :param image: (ndarray) Preprocessed E4 cell images.
+    :param phi: (float) Rotation angle to transform the mppc image coordinate system to the multiprobe
```
pieleric: Unit? rad?

scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,298 @@
...
+    return mppc_rotation  # angle
+
+
+def get_mppc_offset(image, phi):
+    """
+    Calculates the offset of the MPPC detector in space. MPPC detector should be centered in regard to the scanning
+    beamlet, when in single beam mode.
+    Calculates the distance of the active MPPC cell image (bright square) in regard to the center of the cell image.
+    Note: Return value is based on top/left corner (0,0).
+    :param image: (ndarray) Preprocessed E4 cell images.
```
pieleric: Shape of ndarray? Y,X of uint8?

scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,298 @@
...
+
+    return mppc_rotation  # angle
+
+
+def get_mppc_offset(image, phi):
+    """
+    Calculates the offset of the MPPC detector in space. MPPC detector should be centered in regard to the scanning
+    beamlet, when in single beam mode.
+    Calculates the distance of the active MPPC cell image (bright square) in regard to the center of the cell image.
+    Note: Return value is based on top/left corner (0,0).
```
pieleric: Just move this note to the return value description.

scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,298 @@
...
+    """
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--debug', action="store_true", default=False)
+    parser.add_argument('--file', dest="file",
+                        help="Image file to run the mppc to multiprobe alignment for.")
+    parser.add_argument('--angle', dest="phi", required=True,
+                        help="Rotation angle to match the mppc image coordinate system"
+                             "with the coordinate system of the multiprobe."
+                             "Can be automatically extracted with galvo_scan_to_multiprobe script.")
```
pieleric: Spaces are missing at the end of the lines.

scripts/mppc_to_multiprobe.py
```diff
@@ -194,48 +203,50 @@ def get_mppc_rotation(images):
...
 
-    # calculate center of mass (y, x) of the bright spot
-    spotcenter_E4 = center_of_mass(image)
+    # calculate center of mass (x, y) of the bright spot
+    spotcenter_E4 = center_of_mass(image)[::-1]  # invert to follow convention (x, y) in px
     spotcenter_E4_rot = numpy.dot(spotcenter_E4, rot_matrix)  # coordinate transform into system of multiprobe
 
     # center of image
-    image_center = (image.shape[0]/2., image.shape[1]/2.)  # (y, x) [px] is at 400 x 400
+    image_center = (image.shape[0]/2., image.shape[1]/2.)[::-1]  # (x, y) in px is at 400 x 400
```
pieleric: You can simply write `image_center = image.shape[1] / 2, image.shape[0] / 2`

scripts/mppc_to_multiprobe.py
```diff
@@ -250,8 +261,8 @@ def main(args):
     parser.add_argument('--file', dest="file",
                         help="Image file to run the mppc to multiprobe alignment for.")
     parser.add_argument('--angle', dest="phi", required=True,
```
pieleric: Add `type=float` to force it to be a float (and then you won't have to convert it to float).

scripts/mppc_to_multiprobe.py
```diff
@@ -194,48 +203,50 @@ def get_mppc_rotation(images):
...
 def get_mppc_offset(image, phi):
     """
     Calculates the offset of the MPPC detector in space. MPPC detector should be centered in regard to the scanning
     beamlet, when in single beam mode.
     Calculates the distance of the active MPPC cell image (bright square) in regard to the center of the cell image.
-    Note: Return value is based on top/left corner (0,0).
-    :param image: (ndarray) Preprocessed E4 cell images.
-    :param phi: (float) Rotation angle to transform the mppc image coordinate system to the multiprobe
+    :param image: (ndarray, dtype: int64) Preprocessed E4 cell images. Shape is (y, x).
+    :param phi: (float) Rotation angle in degrees to transform the mppc image coordinate system to the multiprobe
```
pieleric: Let's get this this param phi in radians too. To follow the convention.

scripts/mppc_to_multiprobe.py
```diff
@@ -194,48 +203,50 @@ def get_mppc_rotation(images):
...
                 coordinate system.
-    :returns: (tuple of 2 float) The distance to the E4 cell image center in y and x in px.
+    :returns: (tuple of 2 float) The distance to the E4 cell image center in x and y in px.
+              Return value is based on top/left corner (0,0).
     """
 
     rot_matrix = numpy.array(((numpy.cos(phi), - numpy.sin(phi)), (numpy.sin(phi), numpy.cos(phi))))
```
pieleric: Note: sin/cos functions expect angles in *radians*. So currently the rotation matrix is incorrectly computed.

scripts/mppc_to_multiprobe.py
```diff
@@ -276,18 +287,18 @@ def main(args):
...
               "Apply corresponding voltage to the galvo mirror "
               "to correct for the offset.".format(mppc_offset[0], mppc_offset[1]))
 
         # calculate mppc rotation
         mppc_rotation = get_mppc_rotation(corner_images)
         print("Apply {:.2f}° to correct the MPPC position using the manual rotational mount of the MPPC"
-              " detector.".format(mppc_rotation))
+              " detector.".format(math.degrees(mppc_rotation)))
 
     except Exception as exp:
```
pieleric: You can drop the `as exp`.

scripts/mppc_to_multiprobe.py
```diff
@@ -194,48 +203,50 @@ def get_mppc_rotation(images):
...
 
     # control
     delta_x_A8_H8 = spotcenter_H8[1] - spotcenter_A8[1]
     angle_control = numpy.arctan2(delta_y_A8_H8, delta_x_A8_H8)  # [rad]
-    angle_control_degree = -math.degrees(angle_control)  # convert to degree and change sign for correcting the rotation
+    angle_control_degree = -angle_control  # change sign for correcting the rotation
 
-    # check that angles are the same, allow a 1 degree margin
+    # check that angles are the same, allow a 0.5 degree margin
     if not almost_equal(mppc_rotation, angle_control_degree, atol=0.5):
```
pieleric: You need to convert 0.5 into radians.

scripts/mppc_to_multiprobe.py
Sabrina85: I didn't do it to have it similar to the calculation of the spotcenter, but changed it now as suggested.

scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,339 @@
...
+    parser.add_argument('--gain', dest="gain_diagcam", required=True, nargs='+', type=float,
+                        help="The distance (gain) on the diagnostic camera per Volt applied on the AWGs"
+                             "in x and y direction. "
+                             "Can be automatically extracted with galvo_scan_to_multiprobe script "
+                             "(galvo gain in px/volt).")
+
+    parser.add_argument('--px_size_diagcam', dest="px_size_diagcam", type=int, default=75,
+                        help="The pixel size in nm on the diagnostic camera.")  # TODO depending on??
+    # TODO this needs to be an input arg or using the magnification to calc this pixel sizes!
+    parser.add_argument('--px_size_diagimg', dest="px_size_scanimg", type=int, default=40,
```
tepals: same

scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,339 @@
...
+        last_scan = max(scans, key=os.path.getctime)
+        image = TIFF.open(os.path.join(scan_path,
+                                       "{}/field_000_000_0_raw2.tiff".format(os.path.basename(last_scan)))).read_image()
+
+    try:
+        # pre-process image
+        corner_images, image_E4 = preprocess_images(image)
+
+        # calculate mppc offset, Note: accuracy on AWGs is 0.001 volt
+        mppc_offset_px, mppc_offset_volt = \
```
tepals: Personally I think it looks better for mulit-line code not to have to use the "\". I would write this as:
```
mppc_offset_px, mppc_offset_volt = get_mppc_offset(
    image_E4, math.radians(options.phi), tuple(options.gain_diagcam), options.px_size_diagcam, options.px_size_scanimg)
```

scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,339 @@
...
+import cv2
+import glob
+import os
+
+from skimage import measure
+
+from odemis.util.img import RGB2Greyscale
+
+from odemis.dataio import tiff
+from libtiff import TIFF
```
tepals: crtl+alt+o in pycharm orders the imports alphabetically and based on local and outside imports.

scripts/mppc_to_multiprobe.py
```diff
@@ -0,0 +1,339 @@
...
+        images[num_img] = img_close
+
+        # find connected component in thresholded image
+        img_cc, num_labels = measure.label(img_close, background=0, connectivity=1, return_num=True)
+
+        # there should be only one label in num_labels
+        if num_labels != 1:
+            logging.info("There should be only one active area in the image. However, found {:f} labelled"
+                         "objects.".format(num_labels))
+        # Labelled object should have the size of about 1% of total image in px. Otherwise active area cut of or
```
tepals: I think a typo: cut of -> cut off

Sabrina85:
Close pr as we decided to collect all alignment scripts on bitbucket.

PR #1183, opened 2019-08-28
Bug fix for image integration:

Sabrina85:


src/odemis/acq/stream/_sync.py
```diff
@@ -151,6 +151,13 @@ def __init__(self, name, streams):
                 self._polarization = s.polarization
                 self._acquireAllPol = s.acquireAllPol
 
+        # Get integration time on detector if found in optical substream
```
pieleric: Let's more document _what_ integrationTime is supposed to mean here. Something like "Pick integrationTime if found on a stream (typically, optical). In this case, for each ebeam pos, the (short) acquisition will be repeated until the integration time is reached".

src/odemis/acq/stream/_sync.py
```diff
@@ -151,6 +151,13 @@ def __init__(self, name, streams):
...
+        self._integrationTime = None
+        for s in streams:
+            if hasattr(s, "integrationTime"):
+                # get the integration time VA
+                self._integrationTime = s.integrationTime
```
pieleric: I'd suggest that for clarity, `integrationCounts` should also be directly copied too, and they go together.

src/odemis/acq/stream/_sync.py
```diff
@@ -241,7 +248,7 @@ def estimateAcquisitionTime(self):
         for l in self.leeches:
             shape = (len(pol_pos), rep[1], rep[0])
             # estimate acq time for leeches is based on two fastest axis
-            if hasattr(self._sccd, "integrationTime"):
+            if self._integrationTime:
```
pieleric: Let's update every place with such code to use `._integrationTime` and `._integrationCounts`. This ensure that we don't get confused in the long term with some code using the wrong variable.

PR #1184, opened 2019-08-28
chronograph/time correlator: only add selected_wavelength if actually spectrum data available

Sabrina85:


src/odemis/gui/comp/viewport.py
```diff
@@ -1807,7 +1807,9 @@ def connect_stream(self, projs):
         super(TemporalSpectrumViewport, self).connect_stream(projs)
         stream, proj = self._stream, self._projection
 
-        if hasattr(stream, "selected_time"):
+        if hasattr(stream, "selected_time") and hasattr(stream, "selected_wavelength"):
+            # TODO @Anders: not sure what this is for. Think the view for the raw temporal spectrum data in
```
pieleric: I think this is the right way to do it. If the two dimensions are not there, the projection (`PixelTemporalSpectrumProjection`) will return None, and the viewport will be cleared up anyway.

pieleric:
@amuskens Please review.

Sabrina85:
I removed the comment, code is updated.

PR #1186, opened 2019-08-29
ebeam scan to multiprobe script changed to always need a backend to run.

tepals:
- In practice you always run this script with the backend running to get the live view of the diagnostic ccd.
- Added diagnostic-ccd as a default value for the role.
- Cleaned up doc strings.

### DIFF
--- a/scripts/ebeam_scan_to_multiprobe.py
+++ b/scripts/ebeam_scan_to_multiprobe.py
@@ -3,9 +3,11 @@
 """
 Created on 30 Jul 2019
 
-@author: Thera Pals
+@author: Andries Effting, Thera Pals
 
-This script provides a command line interface for aligning the ebeam scan to multiprobe
+Copyright © 2019 Thera Pals, Delmic
+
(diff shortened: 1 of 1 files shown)


scripts/ebeam_scan_to_multiprobe.py
```diff
@@ -83,13 +84,15 @@ def get_sem_rotation(ccd):
...
+    Handles the command line arguments.
+
+    args: The list of arguments passed.
+    return:
+        (int) value to return to the OS as program exit code.
     """
     parser = argparse.ArgumentParser()
     parser.add_argument('--debug', action="store_true", default=False)
-    parser.add_argument("--role", dest="role", metavar="<component>",
+    parser.add_argument("--role", dest="role", default="diagnostic-ccd", metavar="<component>",
```
Sabrina85: Isn't it `Role of the camera to connect to via the Odemis back-end`

tepals: that is correct, I'll update it

PR #1188, opened 2019-09-04
gui comp: flush ctx in slider

philipwinklerdelmic:


philipwinklerdelmic:
self.IsDoubleBuffered did not work in ubuntu! The gui crashed with a segmentation error.

PR #1189, opened 2019-09-04
viewer.spec: updated for python3 installation

philipwinklerdelmic:


install/windows/viewer.spec
```diff
@@ -148,7 +181,7 @@ a = Analysis(
     datas=None,
     hiddenimports=[
         'cairo',
-        'Queue',
+        #'Queue',  # not required for python3 installation
         'queue',
```
pieleric: Please make it still compatible with Python2.
I'd suggest that you move all the "queue" related imports to a separate function. You can detect the python version from the current interpreter.

install/windows/viewer.spec
```diff
@@ -193,7 +226,8 @@ coll = COLLECT(
     a.zipfiles,
     a.datas,
     get_lib_tiff(),
-    get_cairo_dlls(),
+    # get_cairo_dlls(),  # not required for python3 installation
```
pieleric: Please make it still work on Python2.
Be aware that maybe it works now, not because of Python3, but because you are using a new PyInstaller or PyCairo.
So it might be fine to completely drop it. Otherwise, we need to just return nothing from the function if it's Python 3.

philipwinklerdelmic: I'm still using the old cairo version with the python2 installation (py2cairo instead of pycairo). I think this is the only way it's going to work. 

I did not update numpy with python2 (1.9.5 -> 1.16.5). It might also work with the newer numpy version, but there is a good chance it's going to break something.

philipwinklerdelmic:
updated

PR #1191, opened 2019-09-05
driver hamamatsu: python 3 string migration and TODO

pieleric:


src/odemis/driver/hamamatsurx.py
```diff
@@ -1068,6 +1068,10 @@ def sendCommand(self, func, *args, **kwargs):
                         # log the last error code received before timeout
                         logging.error("Latest response before timeout was '%s'",
                                       latest_response)
+                    # TODO: try to close/reopen the connection. However, maybe not resent
```
tepals: typo resent -> resend, or are you resenting the command :)

pieleric:
Let's keep the full Python 3 migration (and the simulator) for later....

PR #1192, opened 2019-09-05
cli video_displayer: when data is a single value, show the value as a number

pieleric:


src/odemis/gui/util/wx_adapter.py
```diff
@@ -45,7 +45,10 @@ def __new__(cls, *args, **kwargs):
             if isinstance(args[0], (file, io.IOBase)):
                 return wx.ImageFromStream(*args)
             elif len(args) >= 2 and isinstance(args[0], int) and isinstance(args[1], int):
-                return wx.EmptyImage(*args, **kwargs)
+                if "data" in kwargs and "alpha" in kwargs:
+                    return wx.ImageFromDataWithAlpha(*args, **kwargs)
```
tepals: Since `wx.ImageFromDataWithAlpha` reports a deprecation warning, shouldn't you use `wx.Image` here, like you did above?

pieleric: That's the code of the wrapper, which converts the new wxPython 4 code to the wxPython 3 version. It'll be only run with the old wxPython, and so on that version, it'll not be deprecated.... but actually the only way to do it.

PR #1193, opened 2019-09-05
plugins timelapse: store data in parallel, to save time

pieleric:


plugins/timelapse.py
```diff
@@ -267,6 +273,57 @@ def start(self):
...
+
+        for i in range(n):
+            t = threading.Thread(target=self._saving_thread, args=(i,))
+            t.start()
+            self._sthreads.append(t)
+
+    def _stop_saving_threads(self):
+        """
+        Blocks until all the data has been stored
+        Can be called multiple times in a raw
```
tepals: on a raw or in a row?

plugins/timelapse.py
```diff
@@ -267,6 +273,57 @@ def start(self):
...
+        """
+        # Indicate to all the threads that they should stop
+        for t in self._sthreads:
+            self._to_store.put((None, None))  # Special "quit" message
```
tepals: it feels like you should use the `t` here? Otherwise comment on why this stops all threads in `self._sthreads`.

pieleric: Indeed, it's not obvious. We send one message per thread. We actually just send N messages, and the queue will ensure that each thread gets one.

plugins/timelapse.py
```diff
@@ -267,6 +273,57 @@ def start(self):
...
+                if fn is None:
+                    self._to_store.task_done()
+                    return
+                logging.info("Saving data %s in thread %d", fn, i)
+                self._exporter.export(fn, das)
+                self._to_store.task_done()
+        except Exception:
+            logging.exception("Failure in the saving thread")
+        finally:
+            logging.debug("Saving thread %d done", i)
```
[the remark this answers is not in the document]
pieleric: Yes, it's handy to know that a thread is done. Technically, you could indeed guess that the thread is done if you see the exception log, but it's easier if it's always shown. Especially, if we have an extra exception handler which shows a different log. I have similar code around almost every thread code.

PR #1198, opened 2019-09-09
python3 dataio hdf5: use to_str_convert when reading strings

philipwinklerdelmic:


philipwinklerdelmic:
I could reproduce the issues, by saving in hdf5 format with python2 odemis and opening with python3. See commit message for details.

pieleric:
`to_str_escape()` is only to be used in log messages, as it converts all non-ascii characters to an escaped representation (eg, "\0xa5"). That is not to be used to "unpickle" a string. 
Instead, I'd suggest to use `codecs.decode`, or to be absolutely correct `lambda s: codecs.decode(s, "utf8")`, as the default encoding on Python 2 is "ascii" (but that metadata should always be ascii anyway).

philipwinklerdelmic:
updated

PR #1199, opened 2019-09-09
python3 odemis-relay

philipwinklerdelmic:


philipwinklerdelmic:
needs to be tested
Is it easily possible to test on a fake port?

philipwinklerdelmic:
updated

PR #1207, opened 2019-09-18
ueye gige camera output to 12 bit

tepals:
For SENSOR_UI527xSE_M the camera output is now 12 bit, because this is the highest possible for that sensor. For all other sensors the camera output is still 8 bit.

### DIFF
--- a/src/odemis/driver/ueye.py
+++ b/src/odemis/driver/ueye.py
@@ -693,11 +693,21 @@ class Camera(model.DigitalCamera):
             self._set_static_settings()
 
             # Set the format
-            # TODO: color mode based on the maximum BPP
-            # cf is_DeviceFeature( IS_DEVICE_FEATURE_CMD_GET_SUPPORTED_SENSOR_BIT_DEPTHS)
-            self._dll.is_SetColorMode(self._hcam, CM_MONO8)
-            self._dtype = numpy.uint8
-            self._metadata[model.MD_BPP] = 8
+            # Loop through possible bit depths and pick the highest one that does not raise an error.
+            # We tried GET_SUPPORTED_SENSOR_BIT_DEPTHS, to find the appropriate bit depth per camera,
+            # however it is not supported on any of the camera's we tried.
+            for bpp, bit_depth in ((16, CM_MONO16), (12, CM_MONO12), (10, CM_MONO10), (8, CM_MONO8)):
+                try:
+                    self._dll.is_SetColorMode(self._hcam, bit_depth)
+                    break  # found a good color mode
+                except UEyeError as err:
+                    if err.errno == 174 and bpp > 8:  # INVALID_COLOR_FORMAT
+                        logging.debug("Set color mode to {} bits not possible, trying a lower value.".format(bpp))
+                    else:
+                        raise
+            self._dtype = numpy.uint16 if bpp > 8 else numpy.uint8
+            self._metadata[model.MD_BPP] = bpp
+            logging.info("Set color mode to {} bits.".format(bpp))
             self._shape = res + (numpy.iinfo(self._dtype).max + 1,)
 
             # TODO: binning? frameRate? resolution + translation = AOI?


src/odemis/driver/ueye.py
```diff
@@ -695,9 +695,14 @@ def __init__(self, name, role, device=None, **kwargs):
             # Set the format
             # TODO: color mode based on the maximum BPP
             # cf is_DeviceFeature( IS_DEVICE_FEATURE_CMD_GET_SUPPORTED_SENSOR_BIT_DEPTHS)
```
tepals: I tried to run this command but got a `NOT_SUPPORTED` error. Also looked through the manual but could not find another way to get the supported bit depth.

pieleric:
I agree with Sabrina. Actually, I guess this is why Thera tried to use `CMD_GET_SUPPORTED_SENSOR_BIT_DEPTHS` . I'd suggest to try a little bti more, Maybe try with code that just follow the example of the manual:
https://en.ids-imaging.com/manuals/uEye_SDK/EN/uEye_Manual_4.92.3/index.html?is_devicefeaturebitdepth.html

That said, this command seems related to changing the *bit depth*, with IS_DEVICE_FEATURE_CMD_SET_SENSOR_BIT_DEPTH. I'm not sure what's relationship with the *color mode* (of `is_SetColorMode()`). I wonder if the later one maybe just changes the data format, even if the camera actually acquire still with 8 bits?! Or maybe the camera always acquires with 12 bits, but the default color mode discards some data to make it 8 bit? 

Maybe you could also just have a code that tries to set first 16 bit, and if it fails, tries 12 bits, etc, until it finds a color mode accepted?

tepals:
@Sabrina85 & @pieleric updated, I changed the code so that it tries 16 bits first, if that fails, tries 12 bits etc., so that it will end up with the highest possible value

tepals:
> Is it needed to handle the unlikely case that 8bit can also not be set?

handled that now

src/odemis/driver/ueye.py
```diff
@@ -695,9 +695,19 @@ def __init__(self, name, role, device=None, **kwargs):
...
+                    self._dll.is_SetColorMode(self._hcam, bit_depth)
+                    self._dtype = numpy.uint8 if bpp == 8 else numpy.uint16
+                    self._metadata[model.MD_BPP] = bpp
+                    logging.info("Set color mode to {} bits.".format(bpp))
+                    break
+                except UEyeError as err:
+                    if err.errno != 174 or (err.errno == 174 and bpp == 8):  # INVALID_COLOR_FORMAT
+                        raise
+                    else:
+                        logging.info("Set color mode to {} bits not possible, trying a lower value.".format(bpp))
```
pieleric: To make it clearer that the exception is coming from `SetColorMode()`, I'd suggest to put only that call inside the try/except block. Moreover I'd suggest to invert the if, to make it also more readable.
Finally, you can omit the zip, and directly write as it'd be generated.
```suggestion
            for bpp, bit_depth in ((16, CM_MONO16), (12, CM_MONO12), (10, CM_MONO10), (8, CM_MONO8)):
                try:
                    self._dll.is_SetColorMode(self._hcam, bit_depth)
                    break  # found a good color mode
                except UEyeError as err:
                    if err.errno == 174 and bpp > 8:  # INVALID_COLOR_FORMAT
                        logging.debug("Set color mode to {} bits not possible, trying a lower value.".format(bpp))
                    else:
                        raise
           self._dtype = numpy.uint16 if bpp > 8 else numpy.uint8 
           self._metadata[model.MD_BPP] = bpp
           logging.info("Set color mode to {} bits.".format(bpp))
```

src/odemis/driver/ueye.py
```diff
@@ -695,9 +695,19 @@ def __init__(self, name, role, device=None, **kwargs):
             # Set the format
             # TODO: color mode based on the maximum BPP
             # cf is_DeviceFeature( IS_DEVICE_FEATURE_CMD_GET_SUPPORTED_SENSOR_BIT_DEPTHS)
```
pieleric: Please update that comment. The `TODO` can go, and indicate that we tried GET_SUPPORTED_SENSOR_BIT_DEPTHS, but it's actually not supported on any camera we've tried.

tepals:
@pieleric updated

PR #1208, opened 2019-09-18
debian: python3 packaging

philipwinklerdelmic:


philipwinklerdelmic:
not ready to merge yet

philipwinklerdelmic:
It seems to be almost working now, just one more problem (didn't we look into that already?):
https://github.com/pyenv/pyenv/issues/691

I could add a patch and repackage. Is there a better way?

pieleric:
Sorry, I don't really understand the relationship with this wxPython issue. We don't use pyenv, we actually package "system" libraries, so it shouldn't be related?

debian/control
```diff
@@ -9,9 +9,15 @@ Build-Depends: debhelper (>= 9),
...
                inkscape,
                cython,
+               cython3,
                python-numpy,
-               python-all-dev (>= 2.7.1)
+               python-all-dev (>= 2.7.1),
+               python3-setuptools,
+               python3-sphinx (>= 1.1.3),
+               python3-numpy,
+               python3-all-dev (>= 2.7.1)
```
pieleric: That's really a small version number for Python 3. I'd suggest to use 3.5 as a minimum, as that's the oldest one we use for testing (on Ubuntu 16.04).

debian/control
```diff
@@ -9,9 +9,15 @@ Build-Depends: debhelper (>= 9),
...
 X-Python-Version: >= 2.7
+X-Python3-Version: >= 3.4
```
pieleric: Let's make it 3.5 too.

debian/control
```diff
@@ -64,7 +70,36 @@ Depends: python (>= 2.7.1),
...
+         python3-evdev,
+         python3-cairo,
+         python3-numpy,
+         python3-scipy,
+         python3-libtiff (>= 0.4.0),
+         python3-h5py (>= 2.1),
+         python3-xdg,
+         python3-suds,
+         python3-matplotlib,
+         python3-opencv (>= 2.4),
```
pieleric: That's also probably worth a little bump, to 3.2 at least.

setup.py
```diff
@@ -49,7 +49,7 @@ def _get_version_git():
 # Check version
 try:
     gver = _get_version_git()
-    if "-" in gver:
+    if b"-" in gver:
```
pieleric: Typically, in Python, the versions are "strings". So I'd suggest to just adjust _get_version_git() to return a string. => .decode("utf-8")  should do it.

debian/rules
```diff
@@ -3,7 +3,11 @@
 VER := $(shell dpkg-parsechangelog | sed -n 's/^Version: *\([^-]\+\)-.\+/\1/p')
 
 %:
-	dh $@ --with python2 --with sphinxdoc
+	dh $@ --with python2,python3 --with sphinxdoc
+
+override_dh_auto_install:
+	python setup.py install --root debian/python-mote --install-layout=deb
+	python3 setup.py install --root debian/python3-mote --install-layout=deb
```
pieleric: I wonder what is this `python-mote`. Did you copy-paste from https://github.com/pimoroni/mote/blob/master/python/packaging/debian/rules ?
If so, I guess it should be changed to "odemis".
However, I think that for Odemis this is a little special. I'd rather avoid having two packages "python-odemis" and "python3-odemis". I think we should still only have a single "odemis" package. Maybe that one includes both the python 2 and python 3 files (so that's it really easy to switch, by changing just the `PYTHON_INTERPRETER` variable. But all the files should go to the same package, so probably `--root debian/odemis`. 

I think there are other alternatives:
* Always package only for one Python version. I'd be fine with only packaging for Python 2. (but still add the Python 3 dependencies, so that it's easy on a computer to switch to the development version on Python 3).
* Have 3 packages: odemis, which contains the generic files and depends on one of the next packages. python-odemis, which contains the Python files for Python 2, and lastly the python3-odemis, which is equivalent for Python 3.

I think this last alternative is nice, but a lot of work for not much usefulness. So I'd recommend to attempt packaging as you're planning now, and if it doesn't work, just switch to Python 2-only.

PR #1209, opened 2019-09-20
Live update acquistion 

amuskens:
For spectrum, AR, and temporal spectrum

### DIFF
--- a/src/odemis/acq/stream/_sync.py
+++ b/src/odemis/acq/stream/_sync.py
@@ -1107,7 +1107,8 @@ class SEMCCDMDStream(MultipleDetectorStream):
 
                     # integrate/sum images
                     self._integrateImages(integration_count)
-
+                    # Live update the setting stream with the new data
+                    self._sccd._onNewData(self._ccd_df, self._acq_data[self._ccd_idx][-1])
                     logging.debug("Done acquiring image number %s out of %s." % (n, tot_num))
 
             # acquisition done!
@@ -1590,6 +1591,9 @@ class SEMCCDMDStream(MultipleDetectorStream):
                             if self._acq_state == CANCELLED:
                                 raise CancelledError()
 
+                    # Live update the setting stream with the new data
+                    self._sccd._onNewData(self._ccd_df, self._acq_data[self._ccd_idx][-1])
+
                     # Since we reached this point means everything went fine, so
                     # no need to retry
                     break
@@ -2052,6 +2056,9 @@ class SEMTemporalMDStream(MultipleDetectorStream):
                     logging.warning("No dwell time metadata in SEM data")
                 s = model.DataArray(pxsum, se_md)
                 se_data.append(s)
+
+                # Live update the setting stream with the new data
+                self._tc_stream._onNewData(self._tc_stream._dataflow, tc_data)
 
             self._onCompletedData(0, se_data)
             self._onCompletedData(1, tc_data)
--- a/src/odemis/acq/test/stream_test.py
+++ b/src/odemis/acq/test/stream_test.py
@@ -1751,6 +1751,56 @@ class SPARC2TestCase(unittest.TestCase):
         numpy.testing.assert_allclose(cl_md[model.MD_POS], exp_pos)
         numpy.testing.assert_allclose(cl_md[model.MD_PIXEL_SIZE], exp_pxs)
 
+    def test_acq_spec_live_update(self):
+        """
+        Test the live feedback acquisition in the spectrum stream
+        """
+        # Check that it works even when not at 0,0 of the sample stage
+        f = self.stage.moveRel({"x":-1e-3, "y": 2e-3})
+        f.result()
+
+        # Zoom in to make sure the ROI is not too big physically
+        self.ebeam.horizontalFoV.value = 200e-6
+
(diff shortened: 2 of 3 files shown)


src/odemis/acq/stream/_sync.py
```diff
@@ -2053,6 +2054,10 @@ def _runAcquisition(self, future):
                 s = model.DataArray(pxsum, se_md)
                 se_data.append(s)
 
+                # Live update the setting stream with the new data
+                self._se_stream._onNewData(None, se_data)
+                self._tc_stream._onNewData(None, tc_data)
```
pieleric: I don't think you should update the "se" (=e-beam) data, as for now this stream is not shown, and if you pass it as-is, it'd show a single pixel.
For the `_tc_stream`, I'd suggest to pass the dataflow as well, just to make sure we don't break anything if the function needs it (in the future). You can use `self._tc_stream._dataflow`.

src/odemis/acq/test/stream_test.py
```diff
@@ -44,6 +44,8 @@
 import unittest
 from unittest.case import skip
 import weakref
+from odemis.model._metadata import MD_ACQ_DATE
+import operator
```
pieleric: Drop this non-used import.

src/odemis/acq/test/stream_test.py
```diff
@@ -44,6 +44,8 @@
 import unittest
 from unittest.case import skip
 import weakref
+from odemis.model._metadata import MD_ACQ_DATE
```
pieleric: Just write in the code `model.MD_ACQ_DATE` and drop this import.

src/odemis/acq/stream/_sync.py
```diff
@@ -1107,7 +1107,8 @@ def _runAcquisitionEbeam(self, future):
 
                     # integrate/sum images
                     self._integrateImages(integration_count)
-
+                    # Live update the setting stream with the new data
+                    self._sccd._onNewData(self._ccd_df, self._acq_data[self._ccd_idx][-1])
```
pieleric: Please add similar code to the `_runAcquisitionScanStage()`. Somewhere just before the `break` (l 1595) should be equivalent. As the two functions are very similar, it should work well too.

src/odemis/driver/simstreakcam.py
```diff
@@ -274,7 +274,7 @@ def _getNewImage(self):
         """
         self._img_counter = (self._img_counter + 1) % len(self._img_list)
         image = self._img_list[self._img_counter]
-
+        image += (numpy.random.rand(*image.shape) * 1000 - 500).astype(numpy.uint16)
```
pieleric: Let's not add this extra debug code here. This driver already supports sending multiple images.

src/odemis/driver/andorcam2.py
```diff
@@ -2954,6 +2954,8 @@ def GetMostRecentImage16(self, cbuffer, size):
         ndbuffer[...] = self._data[self.roi[2] - 1:self.roi[3]:self.binning[1],
                                    self.roi[0] - 1:self.roi[1]:self.binning[0]]
 
+        ndbuffer += (numpy.random.rand(res[1], res[0]) * 1000 - 500).astype(numpy.uint16)
```
pieleric: Please check that all the test cases still pass (andorcam2_test). I don't think any test case expects that images are always the same, but you need to double check.

Also, I'd suggest to tune down a little bit the values to around 200, which is the typical noise level on these cameras. Also make them only positive, to simplify a little bit. Moreover `randint()` seems to do just what you need. So you could write:
`ndbuffer += numpy.random.randint(0, 200, ndbuffer.shape, dtype=ndbuffer.dtype)`

src/odemis/acq/test/stream_test.py
```diff
@@ -2919,6 +3014,57 @@ def _roiToPhys(self, repst):
...
+            time.sleep(3.0)
+            im2 = ars.image.value
+
+            # wait until it's over
+            data = f.result(timeout)
+
+            # Check if the image changed (live update is working)
+            self.assertRaises(AssertionError, numpy.testing.utils.assert_array_equal, im1, im2)
+
     def test_acq_arpol(self):
```
pieleric: Please also add a test for the `SEMTemporalMDStream`.

src/odemis/driver/andorcam2.py
amuskens: The Andorcam tests run fine.

src/odemis/acq/test/stream_test.py
amuskens: Somehow the backend was not starting so I couldn't verify this test yet.

amuskens:
Ok. Added back the change

pieleric:
I think you forgot to push. I don't see any change.

src/odemis/driver/andorcam2.py
amuskens: I force-pushed it now.

pieleric:
Well, now I *think* the code is correct, but it's back to 5 commits! Please squash the commits into just 2, as discussed previously. There should also be no change to the simstreakcam.

amuskens:
Ahhh... noo.. I will see if  can fix it.

PR #1210, opened 2019-09-23
XTclient add hwcomponents

tepals:
@pieleric could you have a look to see if what I'm doing with the SEM, Scanner and Stage objects is correct?

### DIFF
--- a/install/linux/usr/share/odemis/hwtest/clspots-sem-diagnostic.odm.yaml
+++ b/install/linux/usr/share/odemis/hwtest/clspots-sem-diagnostic.odm.yaml
@@ -12,6 +12,7 @@ MultiBeamSEM: {
   role: diagnostic-ccd,
   init: {
     device: null,
+    transp: [-1, -2],
   },
   children: {focus: "DiagnosticCam Focus"}
 }
@@ -32,7 +33,7 @@ MultiBeamSEM: {
 #  class: pigcs.Bus,
 #  role: stage,
 #  init: {
-#    port: "/dev/fake",
+#    port: "/dev/ttyUSB0",
 #    # axis -> controller, channel, closed-loop?
 #    axes: {
 #      "z": [null, "Z", True]
@@ -54,8 +55,7 @@ MultiBeamSEM: {
   class: xt_client.SEM,
   role: sem,
   init: {
-    address: "tcp://127.0.0.1:4242",
-    timeout: 30,
+    address: "PYRO:Microscope@localhost:4242",
   },
   children: {
     scanner: "EBeam control",
--- a/install/linux/usr/share/odemis/hwtest/clspots-sem-diagnostic.odm.yaml
+++ b/install/linux/usr/share/odemis/hwtest/clspots-sem-diagnostic.odm.yaml
@@ -1,0 +1,86 @@
+MultiBeamSEM: {
+  class: Microscope,
+  role: mbsem,
+  children: ["SEM", "EBeam Focus", "EBeam Stage", "EBeam control",
+             "DiagnosticCam", "DiagnosticCam Focus", "Sample Stage"],
+}
+
+# Axes: X is horizontal on screen (going left->right), physical: far->close when looking at the door
+#       Y is vertical on screen (going bottom->top), physical: left->right when looking at the door
+"DiagnosticCam": {
+  class: ueye.Camera,
+  role: diagnostic-ccd,
+  init: {
+    device: null,
+  },
+  children: {focus: "DiagnosticCam Focus"}
+}
+
+"DiagnosticCam Focus": {
+  class: actuator.MultiplexActuator,
+  role: diagnostic-cam-focus,
+  children: {"z": "Sample Stage"},
+  init: {
+    axes_map: {"z": "z"}
+  },
+  properties: {
+    speed: {'z': 0.001} # m/s
+  }
+}
+
+#"Sample Stage": {
+#  class: pigcs.Bus,
+#  role: stage,
+#  init: {
+#    port: "/dev/fake",
+#    # axis -> controller, channel, closed-loop?
+#    axes: {
+#      "z": [null, "Z", True]
+#    },
+#  },
+#}
+
+"Sample Stage": {
+    class: simulated.Stage,
+    role: stage,
+    init: {
+        axes: ["z"],
+        ranges: {"z": [0, 100.e-6]}
+    },
+}
+
+
+"SEM": {
+  class: xt_client.SEM,
+  role: sem,
+  init: {
+    address: "tcp://127.0.0.1:4242",
+    timeout: 30,
+  },
+  children: {
+    scanner: "EBeam control",
+    focus: "EBeam Focus",
+    stage: "EBeam Stage",
+  }
+}
+
+"EBeam control": {
+  creator: "SEM",
+  role: ebeam-control,
+  init: {
+    hfw_nomag: 0.114,
+  },
+}
+
+"EBeam Focus": {
+  # Internal child of SUIPSEM, so no class
+  role: ebeam-focus,
+}
+
+"EBeam Stage": {
+  # Internal child of SUIPSEM, so no class
+  role: ebeam-stage,
+  init: {
+    inverted: ["x", "y"],
+  }
+}
--- a/src/odemis/driver/xt_client.py
+++ b/src/odemis/driver/xt_client.py
@@ -21,19 +21,19 @@ http://www.gnu.org/licenses/.
 """
 from __future__ import division, print_function
 
+import base64
 import logging
+import threading
 import time
 from concurrent.futures import CancelledError
 
-import msgpack_numpy
-import zerorpc
+import msgpack
+import msgpack_numpy as m
+from Pyro5.api import Proxy
 
 from odemis import model
 from odemis import util
-from odemis.model import CancellableThreadPoolExecutor, HwError
-
-# allow to pass numpy arrays over msgpack
-msgpack_numpy.patch()
+from odemis.model import CancellableThreadPoolExecutor, HwError, isasync, CancellableFuture
 
 XT_RUN = "run"
 XT_STOP = "stop"
@@ -45,33 +45,28 @@ class SEM(model.HwComponent):
     Class to communicate with a Microscope server via the ZeroRPC protocol.
     """
 
-    def __init__(self, name, role, children, address, timeout, daemon=None,
+    def __init__(self, name, role, children, address, daemon=None,
                  **kwargs):
         """
         Parameters
         ----------
         address: str
-            server address and port of the Microscope server, e.g. tcp://192.168.1.1:4242
+            server address and port of the Microscope server, e.g. "PYRO:Microscope@localhost:4242"
         timeout: float
             Time in seconds the client should wait for a response from the server.
         """
 
         model.HwComponent.__init__(self, name, role, daemon=daemon, **kwargs)
-        # set heartbeat to None on client and server side, otherwise after two missed heartbeats the client thinks the
-        # connection is lost. A heartbeat happens every 5 seconds, when a function takes longer than 5 seconds to
-        # respond a heartbeat is skipped. timeout controls how long a call can take to respond, the default is 30
-        # seconds.
+        self._proxy_access = threading.Lock()
         try:
-            # If the client fails to connect to the server, sometimes the connect call will raise an error. Other times
-            # the error will be raised when trying to call a method on the server (i.e. get_software_version).
-            self.server = zerorpc.Client(heartbeat=None, timeout=timeout)
-            self.server.connect(address)
+            self.server = Proxy(address)
+            self.server._pyroTimeout = 30  # seconds
             self._swVersion = self.server.get_software_version()
             self._hwVersion = self.server.get_hardware_version()
-        except Exception:
+        except Exception as err:
             raise HwError("Failed to connect to XT server '%s'. Check that the "
-                          "server address and port are correct and XT server is"
-                          " connected to the network." % (address,))
+                          "uri is correct and XT server is"
+                          " connected to the network. %s" % (address, err))
 
         # create the scanner child
         try:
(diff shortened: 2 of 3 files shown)


src/odemis/driver/xt_client.py
```diff
@@ -366,3 +365,362 @@ def set_beam_shift(self, x_shift, y_shift):
...
+            getter=self.get_scanning_size,
+            setter=self.set_scanning_size)
+
+        mag = self._hfw_nomag / fov
+        self.magnification = model.VigilantAttribute(mag, unit="",
+                                                     readonly=True)
+
+        selection_range = self.selected_area_info()["range"]
+        self._shape = selection_range[0][1], selection_range[1][1]
+        self.resolution = self.get_selected_area()[2:]
```
pieleric: `.resolution` should be a VA.
Be careful if you want to support it: there are typically a strong relationship between resolution/scale (and translation). For instance, if the user changes the scale, the resolution is automatically adjusted to keep the same FoV. See phenom or semcomedi.

src/odemis/driver/xt_client.py
```diff
@@ -366,3 +365,362 @@ def set_beam_shift(self, x_shift, y_shift):
...
+        ----------
+        address: str
+            server address and port of the Microscope server.
+        """
+
+        model.HwComponent.__init__(self, name, role, daemon=daemon, **kwargs)
+        self._swVersion = "xtlib"  # TODO figure out sw version
+        self._hwVersion = "Unknown"  # TODO add get_InstrumentID to xt adapter
+        # connect to the microscope pc using zerorpc
+        self.client = zerorpc.Client(address, heartbeat=None, timeout=timeout)
```
pieleric: I didn't think of it in this way. I thought you'd create a "MicroscopeClient", and use it afterwards. That said, as almost every function is a direct mapping, you're right that it's probably as easy to directly call the remote methods. That said, I'd suggest then to (eventually) delete the `MicroscopeClient` class, to avoid code duplication.
Note that a third alternative would be to move all the methods of `MicroscopeClient` onto `SEM` (ie, essentially merging the two classes). Then the scanner/stage could do `self.parent.method()`. The main advantage of this, is that if you want to debug, it's actually possible to call the method on the `SEM` from a terminal, which is the not the case if you have a sub-component.

src/odemis/driver/xt_client.py
```diff
@@ -366,3 +365,362 @@ def set_beam_shift(self, x_shift, y_shift):
     def beam_shift_info(self):
         """Returns: (dict) the unit and xy-range of the beam shift."""
         return self.client.beam_shift_info()
+
+
+class SEM(model.HwComponent):
+    """
+    Connects to a Zeiss SEM via the RemCon interface (over RS-232).
+    At initialisation, the SEM software should already be running, and the
+    RemCon option active (might require an extra license).
```
pieleric: Don't forget to update the docstring eventually.

src/odemis/driver/xt_client.py
```diff
@@ -366,3 +365,362 @@ def set_beam_shift(self, x_shift, y_shift):
...
+
+        selection_range = self.selected_area_info()["range"]
+        self._shape = selection_range[0][1], selection_range[1][1]
+        self.resolution = self.get_selected_area()[2:]
+        pxs = self.horizontalFoV.value / self.resolution[0]
+        self.pixelSize = model.VigilantAttribute(pxs, unit="m", readonly=True)
+
+        self._scale = (self._shape[0] / self.resolution[0],
+                       self._shape[1] / self.resolution[1])
+
```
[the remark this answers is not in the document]
tepals: yes, I will need that. Figured I would first do everything I already implemented on the microscope side :)

src/odemis/driver/xt_client.py
tepals: I was in doubt between what I implemented and creating a `MicroscopeClient` and use that afterwards. But now I actually like the last option, so I'll merge the `SEM` and the `MicroscopeClient` classes

src/odemis/driver/xt_client.py
```diff
@@ -79,6 +111,7 @@ def move_stage(self, position, rel=False):
             position. If False the stage is moved to the absolute position.
         """
         self.client.move_stage(position, rel)
+        return self.get_stage_position()
```
pieleric: I wouldn't recommend putting the `get*()` into the `set*()`. It's handy to have direct wrappers. Inside the Scanner component, you'd create special setter methods, which do both the call to the set() and get().

src/odemis/driver/xt_client.py
```diff
@@ -164,7 +198,10 @@ def set_scanning_size(self, x, y):
         y: (float)
             size for y in meters.
         """
+        if y is None:
+            _, y = self.get_scanning_size()
         self.client.set_scanning_size(x, y)
+        return self.get_scanning_size()
```
pieleric: Similarly, here, just make that wrapper direct, and the caller has to take care of the doing the extra work.
That said, the `set_scanning_size()` of the server only takes one argument (`x`). The y is hard-coded.

src/odemis/driver/xt_client.py
```diff
@@ -23,36 +22,69 @@
...
+        model.HwComponent.__init__(self, name, role, daemon=daemon, **kwargs)
         # set heartbeat to None on client and server side, otherwise after two missed heartbeats the client thinks the
         # connection is lost. A heartbeat happens every 5 seconds, when a function takes longer than 5 seconds to
         # respond a heartbeat is skipped. timeout controls how long a call can take to respond, the default is 30
         # seconds.
-        self.client = zerorpc.Client(server_address, heartbeat=None, timeout=timeout)
+        self.client = zerorpc.Client(heartbeat=None, timeout=timeout)
+        self.client.connect(address)
+        self._swVersion = self.client.get_software_version()
+        self._hwVersion = self.client.get_hardware_version()
```
pieleric: Do these function really exists? I can't see them...

src/odemis/driver/xt_client.py
```diff
@@ -23,36 +22,69 @@
...
         """
+
+        model.HwComponent.__init__(self, name, role, daemon=daemon, **kwargs)
         # set heartbeat to None on client and server side, otherwise after two missed heartbeats the client thinks the
         # connection is lost. A heartbeat happens every 5 seconds, when a function takes longer than 5 seconds to
         # respond a heartbeat is skipped. timeout controls how long a call can take to respond, the default is 30
         # seconds.
-        self.client = zerorpc.Client(server_address, heartbeat=None, timeout=timeout)
+        self.client = zerorpc.Client(heartbeat=None, timeout=timeout)
+        self.client.connect(address)
```
pieleric: I'd advise to rename this `client` attribute. When reading the code it's the bit confusing. Here, it's actually  the object that we use to talk to the _server_. So `.server` would be more appropriate. Alternatively, you could call it `.remote`, or `.connection`.

pieleric: Please catch exceptions, so that if the server address is not responding it raises a `HwError`. This allows to automatically display a nice message in the back-end start-up, and automatically retry.

src/odemis/driver/xt_client.py
tepals: They're in the remote-protocol branch

tepals: They're in the remote-protocol branch

tepals:
@pieleric updated the PR, could you have another look?

install/linux/usr/share/odemis/hwtest/clspots-sem-diagnostic.odm.yaml
```diff
@@ -0,0 +1,86 @@
...
+}
+
+# Axes: X is horizontal on screen (going left->right), physical: far->close when looking at the door
+#       Y is vertical on screen (going bottom->top), physical: left->right when looking at the door
+"DiagnosticCam": {
+  class: ueye.Camera,
+  role: diagnostic-ccd,
+  init: {
+    device: null,
+  },
```
Sabrina85: We might need to check with Andries again. I think they inverted now the ueye in the yaml to swap the coordinate system from left handed to right handed.

src/odemis/driver/xt_client.py
```diff
@@ -22,37 +21,72 @@
...
 
 
-class MicroscopeClient:
+class SEM(model.HwComponent):
     """
     Class to communicate with a Microscope server via the ZeroRPC protocol.
     """
 
-    def __init__(self, server_address="tcp://192.168.1.1:4242", timeout=30):
+    def __init__(self, name, role, children, address, timeout, daemon=None,
```
pieleric: Please provide a default value for the timeout (30s?), as it'll hint on what is a good value. I think we could even argue that the user shouldn't need to set it, and it's up to you to get a good constant.

src/odemis/driver/xt_client.py
```diff
@@ -164,17 +225,23 @@ def set_scanning_size(self, x, y):
         y: (float)
             size for y in meters.
```
pieleric: Drop this part of the docstring.

src/odemis/driver/xt_client.py
```diff
@@ -349,20 +468,579 @@ def set_autofocusing(self, channel, state):
...
+
+        voltage_info = self.parent.ht_voltage_info()
+        self.accelVoltage = model.FloatContinuous(
+            self.parent.get_ht_voltage(),
+            voltage_info["range"],
+            unit=voltage_info["unit"],
+            setter=self._setVoltage
+        )
+
+        self.blanker = model.VAEnumerated(
```
pieleric: You can use `BooleanVA`, and then you can drop the `choices` argument.

src/odemis/driver/xt_client.py
```diff
@@ -349,20 +468,579 @@ def set_autofocusing(self, channel, state):
...
+                self.beamShift._value = beam_shift
+                self.beamShift.notify(beam_shift)
+            rotation = self.parent.get_rotation()
+            if rotation != self.rotation.value:
+                self.rotation._value = rotation
+                self.rotation.notify(rotation)
+            fov = self.parent.get_scanning_size()[0]
+            if fov != self.horizontalFoV.value:
+                self.horizontalFoV._value = fov
+                self.horizontalFoV.notify(fov)
```
pieleric: It'd slightly better to only notify after the magnification value has been set. This way, if a subscriber reads the magnification value, it'll already be correct.

src/odemis/driver/xt_client.py
```diff
@@ -349,20 +468,579 @@ def set_autofocusing(self, channel, state):
...
+    def __init__(self, name, role, parent, rng=None, **kwargs):
+        if rng is None:
+            rng = {}
+        stage_info = parent.stage_info()
+        if "x" not in rng:
+            rng["x"] = stage_info["range"].get("x", (0, 100e-6))
+        if "y" not in rng:
+            rng["y"] = stage_info["range"].get("y", (0, 100e-6))
+        if "z" not in rng:
+            rng["z"] = stage_info["range"].get("z", (0, 100e-6))
```
pieleric: Are there any chance that `stage_info` doesn't contain any of these axes? I'd guess that if there are missing, something is really odd, and it's better to fail, instead of guessing 100µm. => just write like ` rng["x"] = stage_info["range"]["x"]`

src/odemis/driver/xt_client.py
```diff
@@ -349,20 +468,579 @@ def set_autofocusing(self, channel, state):
...
+            logging.exception("Unexpected failure when updating position")
+
+    def _moveTo(self, future, pos, timeout=60):
+        with future._moving_lock:
+            try:
+                if future._must_stop.is_set():
+                    raise CancelledError()
+                logging.debug("Moving to position (%s, %s, %s)", pos["x"], pos["y"], pos["z"])
+                self.parent.move_stage(pos, rel=False)
+                time.sleep(1)
```
pieleric: Please explain why this `sleep()` is needed. 1s sounds quite a long time.

src/odemis/driver/xt_client.py
```diff
@@ -349,20 +468,579 @@ def set_autofocusing(self, channel, state):
...
+                    raise ValueError("Unsupported position %s for axis %s"
+                                     % (val, axis))
+                elif (hasattr(axis_def, "range") and not
+                      axis_def.range[0] <= val <= axis_def.range[1]):
+                    # TODO: if not referenced, double the range
+                    rng = axis_def.range
+                    raise ValueError("Position %s for axis %s outside of range %f->%f"
+                                     % (val, axis, rng[0], rng[1]))
+            else:
+                raise ValueError("Unknown axis %s" % (axis,))
```
pieleric: Any reason that you overriding `_checkMoveAbs()` and `_checkMoveRel()`? They are part of `Actuator`, so they should be already available (and they look very similar).

src/odemis/driver/xt_client.py
```diff
@@ -349,20 +468,579 @@ def set_autofocusing(self, channel, state):
...
+        f = self._executor.submit(self._doMoveAbs, foc)
+        return f
+
+    def stop(self, axes=None):
+        """
+        Stop the last command
+        """
+        # Empty the queue (and already stop the stage if a future is running)
+        self._executor.cancel()
+        logging.debug("Stopping all axes: %s", ", ".join(self.axes))
```
pieleric: This message is not really accurate, as there is only 1 axis, and there is no way to stop the current move (which is pretty much instantaneous anyway). I'd suggest something like "Cancelled all ebeam focus moves".

src/odemis/driver/xt_client.py
```diff
@@ -349,20 +468,579 @@ def set_autofocusing(self, channel, state):
...
+    This is an extension of the model.Actuator class. It provides functions for
+    moving the SEM focus (as it's considered an axis in Odemis)
+    """
+
+    def __init__(self, name, role, parent, **kwargs):
+        """
+        axes (set of string): names of the axes
+        """
+
+        self.parent = parent
```
pieleric: You don't need to store it explicitly, as the super class will take care of it. Just use `parent` instead of `self.parent` in the line below.

src/odemis/driver/xt_client.py
tepals: No this was a mistake, I'll remove them.

tepals:
@pieleric I updated the pull request.

PR #1216, opened 2019-09-27
gui evt: Fix Knob Events with wxPython 4

pieleric:
We had a "super clever" way to generate events.
So clever that the new "super clever" functions of wxPython 4 got completely confused,
and the attributes were lost.

=> Make everything explicit. It makes it work on wxPython 4, and it's also more obvious
so that you don't have to be "super clever" at the time that you read the code in order to understand it.

It should still be compatible with wxPython 3 (but untested)

### DIFF
--- a/src/odemis/gui/evt/__init__.py
+++ b/src/odemis/gui/evt/__init__.py
@@ -21,28 +21,44 @@ from __future__ import division
 
 import wx
 
-
-def new_knob_event():
-    """ Generate new (CmdEvent, Binder) tuple e.g. MooCmdEvent, EVT_MOO = NewCommandEvent() """
-    evttype = wx.NewEventType()
-
-    class _Event(wx.PyCommandEvent):
-        def __init__(self, id, **kw):
-            wx.PyCommandEvent.__init__(self, evttype, id)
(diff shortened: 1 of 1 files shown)


pieleric:
I've tried to test it on wxPython 3, but it turns out that the pycairo update broke it. That said, even if we back port to 2.10, normally every system will use wxPython 4, as the dependency has been bumped.

philipwinklerdelmic:
> I've tried to test it on wxPython 3, but it turns out that the pycairo update broke it. That said, even if we back port to 2.10, normally every system will use wxPython 4, as the dependency has been bumped.

Should we just ignore it and expect everyone to use wxpython 3? It seems that wxpython4 requires the new version of cairo and wxpython3 the old version, so it's going to hard to repackage it in a way that works for both.

pieleric:
Yes, I think we can now (as of 2.10) expect that everyone uses wxPython 4, as it's now the new dependency in our debian package.
As a matter of fact, we could probably even remove the support for wxPython 3 (with the wx_adapter) from the development version... The main reason I haven't done it yet is that it's pretty easy to not break ;-)

PR #1218, opened 2019-10-02
bug fix: left over in test case after renaming of method see commit d4feb0f868c03de1c153cc37a68a66a5ebe970d6

Sabrina85:


Sabrina85:
Yes indeed. Should have checked that... I ran now all the testcases in calibration_test.py and all testcases in stream_test.py related to spectrum/temporal spectrum/chronograph. I updated the commit.

PR #1222, opened 2019-10-04
driver actuator LinearActuator: don't move at init if already referenced

pieleric:
If the axis was already referenced, it would try to move to the "current" position,
but that was without including the offset. It would cause an actual move, unneeded,
and sometimes even prevent initialisation, if the current position of the child is outside the
range.

=> Don't move if there is no need to move.

Also clean-up a little the log messages.

### DIFF
--- a/src/odemis/driver/actuator.py
+++ b/src/odemis/driver/actuator.py
@@ -1006,17 +1006,16 @@ class LinearActuator(model.Actuator):
         logging.debug("Subscribing to position of dependency %s", dep.name)
         dep.position.subscribe(self._update_dep_position, init=True)
 
-        referenced = False
         if model.hasVA(dep, "referenced") and axis_name in dep.referenced.value:
             referenced = dep.referenced.value[axis_name]
             self.referenced = model.VigilantAttribute({self._axis: referenced}, readonly=True)
             dep.referenced.subscribe(self._update_dep_ref)
-        if not referenced:
-            # The initialisation will not fail if the referencing fails
-            f = self.reference({axis})
-            f.add_done_callback(self._on_referenced)
-        else:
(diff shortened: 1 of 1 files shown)


src/odemis/driver/actuator.py
```diff
@@ -1006,17 +1006,16 @@ def __init__(self, name, role, dependencies, axis_name, offset=0,
...
         if model.hasVA(dep, "referenced") and axis_name in dep.referenced.value:
             referenced = dep.referenced.value[axis_name]
             self.referenced = model.VigilantAttribute({self._axis: referenced}, readonly=True)
             dep.referenced.subscribe(self._update_dep_ref)
-        if not referenced:
-            # The initialisation will not fail if the referencing fails
-            f = self.reference({axis})
-            f.add_done_callback(self._on_referenced)
-        else:
-            self.moveAbs({self._axis: self._dependency.position.value[self._caxis]}).result()
```
philipwinklerdelmic: Are you sure we can drop this completely? Shouldn't we try to move to the right position instead, including the offset?

pieleric: Here the "right" position should be the "current" position, so in general no move should happen. 

We could argue on what happens if the current position is out of the range. Should we explicitly request a move to reach the nearest range? In general, at this level (of abstraction), I think it's preferable to never automatically move. This is the initialisation code of (a wrapper of) a hardware component. It's can be really dangerous do decide that because the back-end starts, the hardware has to move. So far we *only* do a move if it's a referencing move and the user indicated it's fine in the arguments (ie, YAML file). Up to the higher level (eg, the Optical path manager) to decide where to move things.

philipwinklerdelmic: ok

PR #1224, opened 2019-10-04
gui cont streams: Fix failure to show acquisition date on foreign locale

pieleric:
Same fix as done in commit 1003ae787bd (gui cont settings...)

This should be the last code with strftime("%c"...).


Sabrina85:
Also couldn't find any other places of strftime("%c"...).

PR #1225, opened 2019-10-07
integration-tests

philipwinklerdelmic:


util/run_intg_tests.py
```diff
@@ -205,7 +205,7 @@ def test_config(sim_conf, path_root, logpath, interpreter):
         logging.error("Back-end still running after requesting to stop")
         backend.kill()
 
-    if gui and gui.returncode != 143:  # SIGTERM return code
+    if gui and gui.returncode != -15:  # SIGTERM return code
```
pieleric: The point here is that when we call "odemis-stop" after 10s, the GUI ends with a error code, and we don't want to catch it as an "Error". At least with Python 2, in practice it's 143. You can try by starting the GUI from the command-line (`odemis-gui`), and later call `odemis-stop`. Use `echo $?` to check the return code.

philipwinklerdelmic: I don't get it. When I print the returncode inside the script, I get -15 and the GUI starts fine. Then I try to reproduce it in a python shell and the return value is 143. How is this possible? The 2 numbers might be related, 15 seems to be signal.SIGTERM value, and it seems that sometimes it's added to 128 to produce the final error code (=143).

philipwinklerdelmic: All testcases that pass return this `ERROR:root:GUI failed to start, with return code -15`

philipwinklerdelmic:
There were more issues with python2/python3. I pushed another commit.

util/runtests.sh
```diff
@@ -144,21 +144,23 @@ run_unittests python3
...
 TESTLOG="$(readlink -m "$TESTLOG")"
-INTEGLOGDIR="./integtest-$DATE/"
-mkdir -p "$INTEGLOGDIR"
+INTEGLOGDIR="./integtest-$DATE"
+mkdir -p "$INTEGLOGDIR-python2/"
+mkdir -p "$INTEGLOGDIR-python3/"
 
 # only echo ERRORs in the output
 touch "$TESTLOG" # To make sure tail doesn't fail
 tail -f "$TESTLOG" | grep --line-buffered "ERROR:" &
```
philipwinklerdelmic: What does this do? Isn't $TESTLOG empty at this point?

philipwinklerdelmic:
updated, also made some more changes to properly save the integration test results

util/runtests.sh
```diff
@@ -144,21 +144,31 @@ run_unittests python3
...
```
pieleric: Surprisingly, `tail -f` does quite different thing from `tail`. It opens the file, and keep checking if there is more line. Every new line is output. It never exits (unless you kill it).
So it's used for the terminal output while the tests are running.

src/odemis/odemisd/start.py
```diff
@@ -298,7 +298,7 @@ def start_backend(self, modelfile):
                            "Try starting Odemis again, and use the \"Report a "
                            "problem\" function if the issue continues.\n\n"
                            "The full message is the following:\n %s" %
-                           (p.returncode, err[-1], "".join(err)))
+                           (p.returncode, err[-1], b"".join(err)))
```
pieleric: If err is a list of bytes, that is not what we expect. We want to show the message, so it should be *strings*. As it is, it will look strange (as the bytes are shown as `b'...'`). I'd suggest to `decode("utf-8", "ignore_error")` on out and err. It *could* be directly inside `proc_communicate()`.

philipwinklerdelmic: updated

PR #1228, opened 2019-10-09
Distance tool in progress

VictoriaDel:
A new functionality is added in the GUI which allows to measure the distance between 2 points on an image.


src/odemis/gui/comp/overlay/world.py
```diff
@@ -832,6 +833,566 @@ def deactivate(self):
...
+                self.ruler_label.colour = self.colour
+
+            Label.draw(self.ruler_label, ctx)
+            # self._debug_ruler_edges(ctx, True)
+
+
+MODE_CREATE_RULER = 1
+MODE_SHOW_RULER = 2
+
+class RulerSelectOverlay(WorldOverlay, DragMixin):
```
pieleric: Maybe just `RulerOverlay`

src/odemis/gui/comp/overlay/world.py
```diff
@@ -832,6 +833,566 @@ def deactivate(self):
...
+
+        if not self.active:
+            return super(RulerSelectOverlay, self).on_left_down(evt)
+
+        if self.right_dragging:
+            evt.Skip()
+            return
+
+        vpos = evt.Position
+        self.drag_v_start_pos = self.drag_v_end_pos = Vec(vpos)
```
pieleric: `self.drag_v_start_pos` and `self.drag_v_end_pos` don't seem used outside of this method. With DragMixin, it's a little more confusing, but I have a feeling that it's still not used. Actually, I have a feeling this class doesn't really need to inherit from DragMixin, as it just needs a couple of similar code.

src/odemis/gui/comp/overlay/world.py
```diff
@@ -832,6 +833,566 @@ def deactivate(self):
...
+            pos=(0, 0),
+            font_size=12,
+            flip=True,
+            align=wx.ALIGN_CENTRE_HORIZONTAL,
+            colour=self.colour,
+            opacity=1.0,
+            deg=None
+        )
+
+        self.drag_v_start_pos = drag_v_start_pos
```
pieleric: Add a comment like: "positions (in view coordinates) where the drag started, and where it is now".

src/odemis/gui/comp/overlay/world.py
```diff
@@ -832,6 +833,566 @@ def deactivate(self):
...
+NONE_RULER_MODE = 0
+MOVE_RULER_MODE = 1
+EDIT_START_RULER_MODE = 2
+EDIT_END_RULER_MODE = 3
+
+class Ruler(object):
+
+    hover_margin = 10  # pixels
+
+    def __init__(self, cnvs, drag_v_start_pos, drag_v_end_pos):
```
pieleric: I think you shouldn't need drag_v_start_pos and drag_v_end_pos. 
However, what you *could* have is p_start_pos and p_stop_pos (defaulting to None). If they are defined, the v_start_pos/v_stop_pos would be immediately computed too. With them, you could pass what you currently set via `v_start_pos` and, more interestingly, you could create some Rulers from a test case.
So this would allow you to do a basic test case where you "manually" add a few Rulers to the overlay, and check that the .draw() doesn't fail.

src/odemis/gui/comp/overlay/view.py
```diff
@@ -1304,7 +1303,8 @@ def draw(self, ctx):
                 y += 40
 
             self.intensity_label.pos = (x, y)
-            self._write_label(ctx, self.intensity_label)
+            Label.draw(self.intensity_label, ctx)
```
pieleric: This label also needs to be fixed.

src/odemis/gui/util/img.py
```diff
@@ -2313,6 +2342,13 @@ def images_to_export_data(streams, view_hfw, view_pos,
 
     # Create legend for print-ready
     if not raw:  # png, tiff
+        surface = cairo.ImageSurface.create_for_data(
+            data_to_draw, cairo.FORMAT_ARGB32, buffer_size[0], buffer_size[1])
+        ctx = cairo.Context(surface)
+        # In raw is False, use a fake canvas and ask the ruler overlay to draw on it.
+        fake_canvas = FakeCanvas(ctx, buffer_size, buffer_center, (1 / buffer_scale[0], 1/buffer_scale[1]))
+        fake_canvas.draw_overlay(orig_canvas.ruler_overlay)
```
pieleric: Put it in a `if orig_canvas.ruler_overlay:`

src/odemis/gui/comp/overlay/world.py
```diff
@@ -832,6 +833,604 @@ def deactivate(self):
...
+                self._update_moving()
+            self.cnvs.Refresh()
+            self._view_to_phys()
+
+    def _update_editing(self, round_angle):
+        """Update view coordinates while editing the ruler. If round_angle(boolean) is True,
+        the ruler is forced to be at one angle multiple of 45 degrees"""
+        current_pos = Vec(self.cnvs.clip_to_viewport(self.drag_v_end_pos))
+
+        if self.edit_hover == gui.HOVER_START:
```
pieleric: Is it possible to just check whether the mode is EDIT_START_RULER_MODE vs EDIT_END_RULER_MODE, and drop the use of `edit_hover`?

src/odemis/gui/comp/overlay/world.py
```diff
@@ -832,6 +833,604 @@ def deactivate(self):
...
+        line segment is kept.
+        Args:
+            v_pos: (v_start_pos or v_end_pos) the view coordinates of the fixed endpoint while dragging
+            current_pos: the view coordinates of the endpoint that is being edited.
+
+        Returns: the view coordinates of the edited endpoint (either the start or the end coordinates of the ruler)
+        """
+        # unit vector for view coordinates
+        dx, dy = (current_pos[0] - v_pos[0],
+                  current_pos[1] - v_pos[1])
```
pieleric: Put it on one line, and drop the parenthesis.

src/odemis/gui/comp/overlay/world.py
```diff
@@ -832,6 +833,604 @@ def deactivate(self):
...
+            ctx.move_to(*b_start)
+            ctx.line_to(*b_end)
+            ctx.stroke()
+            self.pixel_ruler = False
+
+        # Label and ruler have the same colour, highlighted or not
+        self._label.colour = self.highlight if selected else self.colour
+
+        self._label.draw(ctx)
+        # self._debug_edges(ctx, True)
```
pieleric: Please make the `True` clearer by writing `convert_to_buffer=True`.

src/odemis/gui/comp/overlay/world.py
```diff
@@ -832,6 +833,604 @@ def deactivate(self):
...
+                                                   self.v_end_pos,
+                                                   vpos)
+                if dist < self.HOVER_MARGIN:
+                    return gui.HOVER_LINE
+
+        return gui.HOVER_NONE
+
+    def draw(self, ctx, selected, shift=(0, 0), scale=1.0, canvas=None):
+        """Draw a ruler and display the size in meters next to it. If the ruler is selected,
+        highlight it and make it thicker"""
```
pieleric: Add a docstring on what `canvas` is for.

src/odemis/gui/comp/overlay/world.py
```diff
@@ -832,6 +833,604 @@ def deactivate(self):
...
+    def get_hover(self, vpos):
+        """
+        Check if the given position is on/near a selection edge or inside the selection.
+        It returns a "gui.HOVER_*"
+        """
+
+        if self._edges:
+
+            if self.pixel_ruler:
+                return gui.HOVER_LINE
```
pieleric: This doesn't make sense. The mouse has to be near the ruler. It cannot happen for *any* position.
=> Check if its within the outer bounding box, and if not, return `HOVER_NONE`.

src/odemis/gui/util/img.py
```diff
@@ -2313,6 +2346,9 @@ def images_to_export_data(streams, view_hfw, view_pos,
 
     # Create legend for print-ready
     if not raw:  # png, tiff
+        # In print-ready export, a fake canvas is used by the ruler overlay
+        if orig_canvas.ruler_overlay and fake_canvas:
```
pieleric: I don't think you need to check for `fake_canvas` here. It's created when i == 0, so all the time.
Or if you want to be extra careful, I'd put it in a sub `if` block and in the else log a big warning to indicate something unexpected happened.

src/odemis/gui/comp/overlay/world.py
```diff
@@ -832,6 +833,606 @@ def deactivate(self):
...
+
+            ctx.set_source_rgba(0.3, 1, 0.3, 1)
+            ctx.rectangle(*start_rect)
+            ctx.stroke()
+
+            ctx.set_source_rgba(0.6, 1, 0.6, 1)
+            ctx.rectangle(*end_rect)
+            ctx.stroke()
+
+    def _edges_to_rect(self, x1, y1, x2, y2, ):
```
pieleric: The last comma in the parameter list is not very pretty.

src/odemis/gui/comp/overlay/world.py
```diff
@@ -832,6 +833,606 @@ def deactivate(self):
...
+                self._label.font_size = 13
+                self._label.weight = cairo.FONT_WEIGHT_BOLD
+            ctx.move_to(*b_start)
+            ctx.line_to(*b_end)
+            ctx.stroke()
+
+        # Label and ruler have the same colour, highlighted or not
+        self._label.colour = self.highlight if selected else self.colour
+        self._label.draw(ctx)
+        self._debug_edges(ctx)
```
pieleric: Please comment out this line.

src/odemis/gui/comp/overlay/world.py
```diff
@@ -832,6 +833,606 @@ def deactivate(self):
...
+        self._selected_tool_va = tool_va
+        if self._selected_tool_va:
+            self._selected_tool_va.subscribe(self._on_tool, init=True)
+
+        self._selected_ruler = None
+        self._rulers = []  # Rulers and all
+
+        # Indicate whether a mouse drag is in progress
+        self._left_dragging = False
+        self._right_dragging = False
```
pieleric: This `_right_dragging` is always False => drop the 3 places where it's used.

src/odemis/gui/comp/overlay/world.py
```diff
@@ -832,6 +833,606 @@ def deactivate(self):
...
+        if self._left_dragging:
+            if self._selected_ruler:
+                self._selected_ruler.stop_updating_ruler()
+                self.cnvs.update_drawing()
+
+            if self.ruler_mode == MODE_CREATE_RULER:
+                # Revert to the standard (NONE) tool
+                self.ruler_mode = MODE_SHOW_RULER
+                self._selected_tool_va.value = TOOL_NONE
+                self.active = True
```
pieleric: If self.active is False, the line will never be reached (due to the `if` at the beginning of this function). => drop this line.

src/odemis/gui/test/comp_overlay_test.py
```diff
@@ -726,6 +727,84 @@ def on_key(evt):
...
+        tab_mod = self.create_simple_tab_model()
+        view = tab_mod.focussedView.value
+
+        self.add_control(cnvs, wx.EXPAND, proportion=1, clear=True)
+        cnvs.setView(view, tab_mod)
+        cnvs.current_mode = TOOL_RULER
+
+        rol = wol.RulerOverlay(cnvs)
+        rol.activate()
+        rol.enabled = True
```
pieleric: You could replace all these lines by the same thing you did for the export test cases: `tab_mod.tool.choices |= {TOOL_RULER}`. Then `rol = cnvs.ruler_overlay`.
Actually, I don't think that changing the "cnvs.current_mode" has any effect, does it?

PR #1229, opened 2019-10-09
debian-control: add wxgtk3 dependency

philipwinklerdelmic:


philipwinklerdelmic:
Actually, there was an issue on ubuntu 18.04. We need to install this extra dependency for wx to work.

pieleric:
Can you explain more the issue? It sounds like an horrible dependency. What is missing in wxgtk4? Wxversion ?

philipwinklerdelmic:
Sorry, I forgot again about this pythonpath issue. Actually, it works with wxgtk4, but we have to add `/usr/lib/python2.7/dist-packages/wxPython-4.0.1-py2.7-linux-amd64.egg` to the pythonpath. I can add this to bashrc in the ubuntu-delmic install script.

philipwinklerdelmic:
I pushed some changes to ubuntu-delmic. Could you have a look?

PR #1230, opened 2019-10-09
driver smartact: Fixed Python 3 support and added one hold time define

amuskens:


src/odemis/driver/smaract.py
```diff
@@ -1217,8 +1217,8 @@ def __init__(self, name, role, locator, axes, ref_on_init=False, linear_speed=0.
         # Connect to the device
         self._id = c_uint32(MC_5DOF_DLL.SA_MC_INVALID_HANDLE)
 
-        option_string = b"model " + str(MC_5DOF_DLL.hwModel) + "\n" + \
-                        b"locator " + locator
+        option_string = b"model " + str(MC_5DOF_DLL.hwModel).encode('ascii') + b"\n" + \
+                        b"locator " + locator.encode('ascii')
```
philipwinklerdelmic: I would suggest this:
```
        option_string = "model " + str(MC_5DOF_DLL.hwModel) + "\n" + \
                        "locator " + locator
                        
        options = c_char_p(option_string.encode("ascii"))
```

src/odemis/driver/smaract.py
```diff
@@ -2274,10 +2276,6 @@ def __init__(self, name, role, locator, ref_on_init=False, axes=None, speed=1e-3
         # Connect to the device
         self._id = c_uint32(0)
 
-        option_string = b"model " + str(SA_CTLDLL.hwModel) + "\n"
-        # options = c_char_p(option_string.encode("ascii"))
-        options = c_char_p("".encode("ascii"))
-
         logging.debug("Connecting to locator %s", locator)
         self.core.SA_CTL_Open(byref(self._id), c_char_p(locator.encode("ascii")), c_char_p(""))
```
philipwinklerdelmic: The second string also needs to be encoded:
`self.core.SA_CTL_Open(byref(self._id), c_char_p(locator.encode("ascii")), c_char_p("".encode('ascii'))`

pieleric: I concur.
I'd just remark that the second string can be more easily written `c_char_p(b"")`.

src/odemis/driver/smaract.py
pieleric: Maybe this make it even more clear:
`option_string = "model %d\n locator %s" % (MC_5DOF_DLL.hwModel, locator)`

amuskens:
Made changes.

PR #1231, opened 2019-10-10
Ebeam scan to multiprobe script automated.

tepals:
- added auto option to cli, if set to True alignment will be done automatically
- added code to automatically align ebeam scan to multiprobe

### DIFF
--- a/scripts/ebeam_scan_to_multiprobe.py
+++ b/scripts/ebeam_scan_to_multiprobe.py
@@ -7,7 +7,22 @@ Created on 30 Jul 2019
 
 Copyright © 2019 Thera Pals, Delmic
 
-This script provides a command line interface for aligning the ebeam scan to multiprobe.
+This script provides a command line interface for aligning the ebeam scan to
+multiprobe.
+
+This script implements the following steps:
+* Acquire image of multiprobe to determine angle of multiprobe.
+* Move multiprobe on single axis with deflectors (by changing DC offset on scan
+deflector AWG) to obtain angle of movement.
+* Calculate angle of movement relative to angle of multiprobe.
+* Match angle of movement to angle of multiprobe by changing scan rotation in
+XT/SEM PC.
+
+Information on how to run this can be found in the sonic-engineering repo:
+sonic-engineering/doc/useful commands.txt
+
+To use this script first run the following line or add it at the end of .bashrc.
+export PYTHONPATH="$HOME/development/sonic-engineering/:$PYTHONPATH"
 
 """
 from __future__ import division
@@ -15,9 +30,12 @@ from __future__ import division
 import argparse
 import logging
 import math
-import numpy
 import sys
 from builtins import input
+
+import numpy
+
+from scanwaveform import dcoffset
 
 from odemis import model
 from odemis.acq.align.spot import FindGridSpots
@@ -26,17 +44,24 @@ from odemis.util.driver import get_backend_status, BACKEND_RUNNING
 
 def get_deflection_angle(coordinates):
     """
-    Fit a line through a set of coordinates and calculate the angle between that line and the x-axis. This angle is
-    called the deflection angle.
+    Fit a line through a set of coordinates and calculate the angle between that
+    line and the x-axis. This angle is called the deflection angle.
 
-    coordinates (ndarray of shape nx2): (x, y) coordinates of center positions of the grid of spots.
-    return:
-        phi (float): deflection angle in radians.
+    Parameters
+    ----------
+    coordinates: (ndarray of shape nx2)
+        (x, y) center coordinates of the moved spot grids.
+
+    Returns
+    -------
+    phi: (float)
+        Deflection angle in radians.
     """
     # Construct a system of linear equations to fit the line a*x + b*y + c = 0
     n = coordinates.shape[0]
     if n < 2:
-        raise ValueError("Need 2 or more coordinates, cannot find deflection angle for {} coordinates.".format(n))
+        raise ValueError("Need 2 or more coordinates, cannot find deflection "
+                         "angle for {} coordinates.".format(n))
     elif n == 2:
         x = coordinates[1, :] - coordinates[0, :]
         # Determine the angle of this line to the x-axis, -pi/2 < phi < pi/2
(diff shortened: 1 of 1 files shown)


scripts/ebeam_scan_to_multiprobe.py
```diff
@@ -7,36 +7,56 @@
...
-    coordinates (ndarray of shape nx2): (x, y) coordinates of center positions of the grid of spots.
-    return:
-        phi (float): deflection angle in radians.
+    Fit a line through a set of coordinates and calculate the angle between that
+    line and the x-axis. This angle is called the deflection angle.
+
+    Parameters
+    ----------
+    coordinates: (ndarray of shape nx2)
+        (x, y) coordinates of center positions of the grid of spots.
```
Sabrina85: When I first read it I thought it is the center of the grid, but that didn't match with the type. Maybe:
`(x, y) coordinates of the center positions of each spot in the grid.`

scripts/ebeam_scan_to_multiprobe.py
```diff
@@ -55,27 +75,49 @@ def get_deflection_angle(coordinates):
...
-        image = ccd.data.get(asap=False)
-        spot_coordinates, translation, scaling, rotation = FindGridSpots(image, n_spots)
-        coordinates.append(translation)
-        inp = input('Applied x-shift to electron beam again? y/n \n')
-        while inp.lower() not in ['y', 'n']:
-            inp = input('{} not an option please choose from y/n \n'.format(inp))
-        if inp.lower() == "n":
-            break
+    if auto:
+        for offset in [-4, -2, 0.0, 2, 4]:
```
Sabrina85: Maybe add a comment explaining why these values are chosen.

scripts/ebeam_scan_to_multiprobe.py
```diff
@@ -55,27 +75,49 @@ def get_deflection_angle(coordinates):
...
 
-def get_sem_rotation(ccd):
+def get_sem_rotation(ccd, auto=True, channel=1):
     """
-
-    ccd: a camera object
-    return:
-    phi: rotation that needs to be added to the scan rotation in the SEM UI.
+    Find the angle of the EBeam-Deflector-x relative to the diagnostic camera.
+
```
Sabrina85: Would be nice to have some info how the procedures differ (auto vs. manual).

scripts/ebeam_scan_to_multiprobe.py
```diff
@@ -86,14 +128,29 @@ def main(args):
...
+                        help="Role of the camera to connect to via the Odemis "
+                             "back-end. Ex: 'ccd'.")
+    parser.add_argument("--channel", dest="channel", default=1,
+                        metavar="<component>",
+                        help="Number of the AWG output channel.")
+    parser.add_argument("--auto", dest="auto", default=True,
+                        metavar="<component>",
+                        help="If True automatically align the ebeam scan to"
+                             "multiprobe. To do this a server must be running "
+                             "on the microscope PC.")
```
Sabrina85: Would that need some more info here? Explaining which server? I am not sure I know what to do, but maybe one would assume the regular user would know.

scripts/ebeam_scan_to_multiprobe.py
```diff
@@ -105,10 +162,20 @@ def main(args):
...
+        if options.auto:
+            val = scanner.rotation.value
+            scanner.rotation.value = val + sem_rotation
+            print("Added {:.3f}° to the scan rotation using SEM PC."
+                  "Rotation is now set to: {:.3f}°".format(
+                math.degrees(sem_rotation),
+                math.degrees(scanner.rotation.value)))
+        else:
+            print("Add {:.3f}° to the scan rotation using SEM PC".format(
+                math.degrees(sem_rotation)))
```
Sabrina85: How is adding the angle to the scan rotation done at the SEM PC. Is there a specific name in the control software. If yes maybe nice to add this info. 
Btw. Is the angle calculation including if the correction needs to be CW or CCW. I guess it does but didn't check explicitly.

scripts/ebeam_scan_to_multiprobe.py
tepals: It's the center positions of the moved grid. I will change the text, is this clear:
`(x, y) coordinates of the center positions of the moved grid of spots.`

scripts/ebeam_scan_to_multiprobe.py
Sabrina85: Nice! Btw what is the accuracy needed for the output?

scripts/ebeam_scan_to_multiprobe.py
Sabrina85: Where can I find this document?

scripts/ebeam_scan_to_multiprobe.py
tepals: you can enter the rotation in the SEM software with 1 value after the point

scripts/ebeam_scan_to_multiprobe.py
tepals: for now it's only on the desktop of the sonic control PC, but I was planning on adding it somewhere on git or dropbox today (just need to decide where to put it).

scripts/ebeam_scan_to_multiprobe.py
```diff
@@ -7,36 +7,61 @@
...
+deflector AWG) to obtain angle of movement.
+* Calculate angle of movement relative to angle of multiprobe.
+* Match angle of movement to angle of multiprobe by changing scan rotation in
+XT/SEM PC.
+
+Information on how to run this can be found in the sonic-engineering repo:
+sonic-engineering/doc/useful commands.txt
+
+To use this script first run the following line or add it at the end of .bashrc.
+export PYTHONPATH="$HOME/development/sonic-engineering/scanwaveform/:$PYTHONPATH"
```
pieleric: For readability, I'd recommend to even make `scanwaveform` a package (by adding an empty `__init__.py` file), and then you can type `from scanwaveform import dcoffset`.

scripts/ebeam_scan_to_multiprobe.py
```diff
@@ -55,27 +80,54 @@ def get_deflection_angle(coordinates):
...
+    Returns
+    -------
+    phi: (float)
+        The angle of the EBeam-Deflector-x relative to the diagnostic camera.
     """
     image = ccd.data.get(asap=False)
     n_spots = (8, 8)
-    spot_coordinates, translation, scaling, rotation = FindGridSpots(image, n_spots)
+    spot_coordinates, translation, scaling, rotation = FindGridSpots(image,
+                                                                     n_spots)
```
pieleric: I'd argue that this change doesn't improve readability. In general, I think that the maximum line length rule should not be applied strictly.

PR #1232, opened 2019-10-11
driver hamamatsurx: accept repetition triggers less than 1

pieleric:
Otherwise we get errors like:

ERROR: Hamamatsu streak camera TCP/IP image thread failed.
  File "/home/sparc/development/odemis/src/odemis/driver/hamamatsurx.py", line 544, in _getDataFromBuffer
    self.parent._delaybox._updateTriggerRate()
  File "/home/sparc/development/odemis/src/odemis/driver/hamamatsurx.py", line 874, in _updateTriggerRate
    triggerRate = int(triggerRate_raw[0])
ValueError: invalid literal for int() with base 10: '0.001'

Also update a little bit the log message to be clearer on which image we received

### DIFF
--- a/src/odemis/driver/hamamatsurx.py
+++ b/src/odemis/driver/hamamatsurx.py
@@ -510,8 +510,8 @@ class ReadoutCamera(model.DigitalCamera):
                 image = numpy.frombuffer(img, dtype=numpy.uint16)  # convert to array
                 image.shape = (int(img_info[1]), int(img_info[0]))
 
-                logging.debug("Requested image number %s, received image number %s from buffer.",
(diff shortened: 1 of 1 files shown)


pieleric:
Indeed, I guess they've changed it manually. Looking at the code, it looks like it should be allowed, otherwise we wouldn't be reading the rate after every image, right?

Sabrina85:
I think the original idea was not to change it and it is a read-only VA, but also think it makes sense to make it more generic. Only wondering what other issues that can produce when they change the rate. Should we check with them if they changed the rate manually?

PR #1237, opened 2019-10-14
Fix middle mouse button dragging in Windows

amuskens:
Note: Requires further testing in Windows.

### DIFF
--- a/src/odemis/gui/comp/viewport.py
+++ b/src/odemis/gui/comp/viewport.py
@@ -1193,6 +1193,7 @@ class NavigablePlotViewport(PlotViewport):
 
         self.canvas.Bind(wx.EVT_MIDDLE_DOWN, self.on_drag_start)
         self.canvas.Bind(wx.EVT_MIDDLE_UP, self.on_drag_end)
+        self.canvas.Bind(wx.EVT_LEAVE_WINDOW, self.on_drag_end)
         self.canvas.Bind(wx.EVT_MOTION, self.on_canvas_motion)
         self.canvas.Bind(miccanvas.EVT_FIT_VIEW_TO_CONTENT, self.on_fit_view)
 


src/odemis/gui/comp/viewport.py
```diff
@@ -1193,6 +1193,7 @@ def __init__(self, *args, **kwargs):
 
         self.canvas.Bind(wx.EVT_MIDDLE_DOWN, self.on_drag_start)
         self.canvas.Bind(wx.EVT_MIDDLE_UP, self.on_drag_end)
+        self.canvas.Bind(wx.EVT_LEAVE_WINDOW, self.on_drag_end)
```
pieleric: Shouldn't it be more like in `DragMixin`?
`self.cnvs.Bind(wx.EVT_KILL_FOCUS, self._on_focus_lost)`

amuskens: Should this one bind too to the drag end function you think?

It seems to work

pieleric:
Yes, it kinds-of-works, but looking at it more into details, I think the more proper way would be do do the same as what BufferedCanvas does on mouse_down/up:
```
    def on_mouse_down(self):
:
        if not self.HasCapture():
            self.CaptureMouse()
:

    def on_mouse_up(self):
:
        if self.HasCapture():
            self.ReleaseMouse()
:
```
The difference is that the canvas still moves until you release the mouse button, even if the mouse goes out of the canvas.

src/odemis/gui/comp/viewport.py
amuskens: Seems like this breaks the view port from loading?

amuskens:
The mouse capturing and releasing seemed to be a bit dangerous as it doesn't seem to always release properly and then it is impossible to click on anything in Ubuntu. Maybe I didn't use it in the right way?

pieleric:
I've implemented the mouse capture in PR#1282

PR #1245, opened 2019-10-22
driver pigcs: make sure we don't connect to hardware when we want to simulate

philipwinklerdelmic:


pieleric:
Hi, I think I have found a better way (and fixed the deeper issue too). See PR #1247 .

PR #1248, opened 2019-10-23
driver pmtctrl: handle disconnecting device

philipwinklerdelmic:


philipwinklerdelmic:
still needs testing on HW, waiting for review first

philipwinklerdelmic:
should be working now

src/odemis/driver/pmtctrl.py
```diff
@@ -447,6 +465,36 @@ def _sendCommand(self, cmd):
...
+            try:
+                logging.debug("retrying to open port %s", self._port)
+                self._port = self._findDevice(self._portname)
+            except IOError:
+                time.sleep(2)
+            except Exception:
+                logging.exception("Unexpected error while trying to recover device")
+                raise
+            else:
+                break
```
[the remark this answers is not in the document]
philipwinklerdelmic: I think we might need it. Somehow, we need to exit the while loop once we found the right port. If we don't use else, it's just going to continue the loop forever.

src/odemis/driver/pmtctrl.py
```diff
@@ -420,18 +422,34 @@ def _sendCommand(self, cmd):
...
+                try:
+                    char = self._serial.read()
+                except IOError:
+                    logging.warn("Failed to read from PMT Control firmware, " +
+                                 "trying to reconnect.")
+                    if self._recovering:
+                        raise
+                    else:
+                        self._tryRecover()
+                        return
```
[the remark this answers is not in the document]
philipwinklerdelmic: Do you think this might work?
```
                else:
                    self._tryRecover()
                    # send command again
                    return self._sendCommand(cmd)
```

src/odemis/driver/pmtctrl.py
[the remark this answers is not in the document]
pieleric: I'm not suggesting to remove the `break`, but only the `else`. It's mostly a matter of taste, but typically I think of the "else" as the special part of the code which should happen if no exception happen, even if some exceptions handler "recover" and continue further. Outside of the else is the all the code that should always happen after the code in the try/except was executed. Here, actually, both semantic fit, so it's up to you, I mostly care about adding a big comment next to the break indicating that it's a good thing to break.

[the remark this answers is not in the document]
philipwinklerdelmic: That makes sense! But then the break has to go in the try block, right? Otherwise we will break after an IOError, even though we want to continue again. Or we should re-raise the IOError if we want to use the break afterwards.
```
                try:
                    logging.debug("retrying to open port %s", self._port)
                    self._port = self._findDevice(self._portname)
                    # We found it back!
                    break
                except IOError:
                    time.sleep(2)
```

src/odemis/driver/pmtctrl.py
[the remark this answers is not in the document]
pieleric: The main things to consider is what happens if actually the device receives the command a second time, and what happens to the original caller if we raise an error.
In this driver, I think most of the cases sending the command twice in a row is very safe (just set a value to the same value), and the only place it could be a tiny bit dangerous would be when resetting the protection. My understanding is that this recovery part is already quite unlikely to happen (as it would more likely happen at the write before). So it should be fine to do.
Technically speaking, the RLock allows you to do that. You could imagine reaching some kind of recursion limit if it keeps calling itself, but that's probably the best thing to do in such bad case anyway.

src/odemis/driver/pmtctrl.py
```diff
@@ -447,6 +470,36 @@ def _sendCommand(self, cmd):
...
+        # again in the same thread
+        with self._ser_access:
+            while True:
+                try:
+                    self._serial.close()
+                    self._serial = None
+                except Exception:
+                    pass
+                try:
+                    logging.debug("retrying to open port %s", self._port)
```
pieleric: As typically `port` is a pattern (eg, /dev/PMT*), this message is a little odd. I'd suggest something like "Searching for the device on port %s".

src/odemis/driver/pmtctrl.py
```diff
@@ -447,6 +470,36 @@ def _sendCommand(self, cmd):
...
+                    self._port = self._findDevice(self._portname)
+                except IOError:
+                    time.sleep(2)
+                except Exception:
+                    logging.exception("Unexpected error while trying to recover device")
+                    raise
+                else:
+                    # We found it back!
+                    break
```
pieleric: Indeed, as the IOError actually continues (which I hadn't noticed), the break would need to go inside the try/except, which would be less readable. So the `else:` is good :-)

src/odemis/driver/pmtctrl.py
```diff
@@ -447,6 +470,36 @@ def _sendCommand(self, cmd):
...
+
+        # it now should be accessible again
+        self.state._set_value(model.ST_RUNNING, force_write=True)
+        logging.info("Recovered device on port %s", self._port)
+        self._recovering = False
```
pieleric: Probably this should be done always when quitting this function, even if an exception is raised. => Move this into a try/finally?

philipwinklerdelmic: I think the _recovering flag should be true as long as we try to find the device to avoid infinite recursion (sendCommand calls tryrecover calls sendcommand calls tryrecover...). Only once there are no exceptions anymore, we want to continue as usual (flat reset to false).

philipwinklerdelmic:
updated

src/odemis/driver/pmtctrl.py
pieleric: I'm worried about the exception happening line 496. If it ever gets triggered, recovering will never happen again.

src/odemis/driver/pmtctrl.py
```diff
@@ -447,6 +473,36 @@ def _sendCommand(self, cmd):
 
             return ans.rstrip()
 
+    def _tryRecover(self):
+        self._recovering = True
+        self.state._set_value(HwError("PMTControl disconnected"), force_write=True)
```
pieleric: -> "USB connection lost"

src/odemis/driver/pmtctrl.py
```diff
@@ -421,17 +423,41 @@ def _sendCommand(self, cmd):
...
+                try:
+                    char = self._serial.read()
+                except IOError:
+                    logging.warn("Failed to read from PMT Control firmware, "
+                                 "trying to reconnect.")
+                    if self._recovering:
+                        raise
+                    else:
+                        self._tryRecover()
+                        # don't send command again
```
amuskens: What is the rationale behind not sending the command again?

philipwinklerdelmic: At this stage, we have already sent a command to the hardware and the connection failed afterwards. Sending the command a second time could potentially be harmful (like doing a move twice). In the PMT, there shouldn't be any commands that break something when sent twice, but it's better to be safe.

src/odemis/driver/pmtctrl.py
philipwinklerdelmic: Indeed, I didn't see that. I would suggest to also add a `self._recovering = False` to the except.

philipwinklerdelmic:
updated

philipwinklerdelmic:
updated

PR #1249, opened 2019-10-23
driver tmcm: autoreconnect

philipwinklerdelmic:


philipwinklerdelmic:
It's working!

src/odemis/driver/tmcm.py
```diff
@@ -275,7 +275,7 @@ def __init__(self, name, role, port, axes, ustepsize, address=None,
         self._refproc_cancelled = {}  # axis number -> event
         self._refproc_lock = {}  # axis number -> lock
 
-        self._ser_access = threading.Lock()
+        self._ser_access = threading.RLock()
         self._serial, ra = self._findDevice(port, address)
         self._target = ra  # same as address, but always the actual one
         self._port = port  # or self._serial.name ?
```
pieleric: Do as in PMTctrl: call it "_portpattern", and you can drop this comment.

src/odemis/driver/tmcm.py
```diff
@@ -843,11 +843,24 @@ def SendInstruction(self, n, typ=0, mot=0, val=0):
         msg[-1] = numpy.sum(msg[:-1], dtype=numpy.uint8)
         with self._ser_access:
             logging.debug("Sending %s", self._instr_to_str(msg))
-            self._serial.write(msg)
+            try:
+                self._serial.write(msg)
+            except IOError:
+                logging.warn("Failed to send command to TMCM, trying to reconnect.")
+                self._tryRecover()
+                # send instruction again
```
pieleric: I'd suggest to make the comment more explicit: "Failure here should mean that the device didn't get the (complete) instruction, so it's safe to send the command again."

src/odemis/driver/tmcm.py
```diff
@@ -875,6 +888,34 @@ def SendInstruction(self, n, typ=0, mot=0, val=0):
...
+        # again in the same thread
+        with self._ser_access:
+            while True:
+                try:
+                    self._serial.close()
+                    self._serial = None
+                except Exception:
+                    pass
+                try:
+                    logging.debug("retrying to open port %s", self._port)
```
pieleric: Could change the log to something similar as for the PMTCtrl:
`logging.debug("Searching for the device on port %s", self._portpattern)`

src/odemis/driver/tmcm.py
```diff
@@ -875,6 +888,34 @@ def SendInstruction(self, n, typ=0, mot=0, val=0):
 
                 return rval
 
+    def _tryRecover(self):
+        self.state._set_value(HwError("TMCM disconnected"), force_write=True)
```
pieleric: As it's the state of the TMCM component it's probably not so useful to indicate "TMCM". Probably for the user, it would be more useful to indicate something like "USB connection lost" (so that he knows it's an issue with the cable or hub).

philipwinklerdelmic:
updated

PR #1257, opened 2019-11-04
bugfix opening files with non-ascii characters in filename

philipwinklerdelmic:


philipwinklerdelmic:
I was wrong, the filename problem was not related to the space, it was due to the non-ascii characters.

philipwinklerdelmic:
updated

PR #1258, opened 2019-11-04
PIGCS driver: add "master" argument to class Bus

Sabrina85:


src/odemis/driver/pigcs.py
```diff
@@ -2808,8 +2808,8 @@ class Bus(model.Actuator):
     def __init__(self, name, role, port, axes, baudrate=38400,
                  dist_to_steps=None, min_dist=None,
                  vmin=None, speed_base=None, auto_suspend=None,
-                 suspend_mode=None,
-                 _addresses=None, **kwargs):
+                 suspend_mode=None, master=254,
+                 _addresses=None,  **kwargs):
```
pieleric: Please don't add an extra space here.

src/odemis/driver/pigcs.py
```diff
@@ -2825,6 +2825,8 @@ def __init__(self, name, role, port, axes, baudrate=38400,
           Default is 10 s.
         suspend_mode (dict str -> "read" or "full"): see CLRelController.
           Default is "read".
+        master: In _some_ TCP/IP configuration, there is a master controller at address 254.
+            Default value of 254 is standard value for IPAccesser with multiple controllers.
```
pieleric: I'd suggest a slightly modified version:
```
master (0<=int<=255): The address of the "master" controller when connecting over TCP/IP to multiple controllers.
    It is unused when connecting over serial port.
```

src/odemis/driver/pigcs.py
```diff
@@ -2866,9 +2868,7 @@ def __init__(self, name, role, port, axes, baudrate=38400,
 
         # Special support for no address
         if len(controllers) == 1 and None in controllers:
-            master = None
-        else:
-            master = 254  # Standard value for IPAccesser with multiple controllers
+            master = None  # overwrite master default value if no address
```
pieleric: I'd suggest: "direct connection to the controller" (the rest is already explained by the comment before the `if`)

Sabrina85:
I found two more places where the default value of master is 254:

_openPort() in class Bus
__init_PBusAccesser

I think I need to change this also to master=master as it is now a param in the init of the Bus class. Is that correct?

pieleric:
Looking at the code, it seems that all the callers of these functions already pass `master`. So there shouldn't be any extra change needed.

PR #1259, opened 2019-11-04
config: Don't force exact version in requirements.txt

pieleric:
We use it to indicate the versions that are known to work, so let's be more generic.


requirements.txt
```diff
@@ -1,24 +1,19 @@
...
+Cython>=0.23.5
+decorator>=4.0.9
+future>=0.16.0
+futures>=3.0.5
 h5py==2.5.0
-libtiff==0.4.2
-matplotlib==1.4.3
+libtiff>=0.4.2
+matplotlib>=1.4.3,<3.0
 numpy==1.11.3
```
tepals: just out of curiosity, why do we keep numpy at 1.11.3?

PR #1260, opened 2019-11-04
powerctrl: turn power off/on on termination

philipwinklerdelmic:


src/odemis/odemisd/main.py
```diff
@@ -335,6 +335,8 @@ def _terminate_all_alive(self):
         for comp in list(self._dependents.keys()):  # ._dependents will be modified, so we need a copy
             for child in comp.children.value:
                 self._dependents.pop(child, None)
+
+        power_supply = model.getComponent(role='power-control')
```
pieleric: Please don't rely on a specific role in the backend.

src/odemis/odemisd/main.py
```diff
@@ -349,6 +351,13 @@ def _terminate_all_alive(self):
                     # access dep.parent
                     logging.warning("Failed to find dependency %s of component %s when terminating.",
                                     dep.name, comp.name, exc_info=True)
+            # Add power supply unit as dependency, so it's not terminated before any of the components
+            # that depend on it. The reason is that on termination, the power supply unit turns off
+            # the power of some components and this should only happen if the components are already
+            # properly terminated.
+            if model.hasVA(comp, 'powerSupply') and power_supply:
+                self._dependents[power_supply].add(comp)
+
```
pieleric: In the back-end, we should rely only on the "power_supplier" option of the components. See the modelgen, which is already aware of this special dependency when starting the components.

philipwinklerdelmic:
updated

PR #1263, opened 2019-11-06
ebeam_scan to multiprobe, galvo_scan to multiprobe updated to calcula…

tepals:
…te the gain (pixels per volt)

- Removed manual function from both scripts.
- get_deflection_angle now calculates a transform between voltages and pixel coordinates, from this transformation the rotation and scale (gain) is obtained.
- changed dcoffset call, because new dcoffset function does not require to pass a channel only 'x' or 'y'

### DIFF
--- a/scripts/ebeam_scan_to_multiprobe.py
+++ b/scripts/ebeam_scan_to_multiprobe.py
@@ -31,105 +31,79 @@ import argparse
 import logging
 import math
 import sys
-from builtins import input
-
-import numpy
 
 from scanwaveform import dcoffset
 
 from odemis import model
 from odemis.acq.align.spot import FindGridSpots
 from odemis.util.driver import get_backend_status, BACKEND_RUNNING
+from odemis.util import transform
 
 
-def get_deflection_angle(coordinates):
+def get_scan_transform(coordinates, voltages):
     """
-    Fit a line through a set of coordinates and calculate the angle between that
-    line and the x-axis. This angle is called the deflection angle.
+    Determine the scaling transform between voltages and coordinates. The scaling transform consists of a translation,
+    rotation and scaling transformation.
 
     Parameters
     ----------
-    coordinates: (ndarray of shape nx2)
+    coordinates: (list of tuples)
         (x, y) center coordinates of the moved spot grids.
+    voltages: (list of tuples)
+        x and y offset applied to the deflectors.
 
     Returns
     -------
     phi: (float)
-        Deflection angle in radians.
+        Deflection angle in radians. Angle is counterclockwise.
+    scale: tuple of floats
+        The x and y scale factor between the voltages and coordinates, in pixels per volt.
     """
-    # Construct a system of linear equations to fit the line a*x + b*y + c = 0
-    n = coordinates.shape[0]
-    if n < 2:
-        raise ValueError("Need 2 or more coordinates, cannot find deflection "
-                         "angle for {} coordinates.".format(n))
-    elif n == 2:
-        x = coordinates[1, :] - coordinates[0, :]
-        # Determine the angle of this line to the x-axis, -pi/2 < phi < pi/2
-        phi = numpy.arctan2(x[0], x[1])
-    else:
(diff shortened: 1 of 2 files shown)


scripts/ebeam_scan_to_multiprobe.py
```diff
@@ -31,105 +31,79 @@
...
+    coordinates: (list of tuples)
         (x, y) center coordinates of the moved spot grids.
+    voltages: (list of tuples)
+        x and y offset applied to the deflectors.
 
     Returns
     -------
     phi: (float)
-        Deflection angle in radians.
+        Deflection angle in radians. Angle is counterclockwise.
```
Sabrina85: Can you also add where 0 is? In Odemis is 0 at 12 o'clock, which is not matching with the mathematical convention.

scripts/ebeam_scan_to_multiprobe.py
```diff
@@ -31,105 +31,79 @@
...
-        # Solve the equation A*x = 0; i.e. find the null space of A. The solution is
-        # the eigenvector corresponding to the smallest singular value.
-        U, s, V = numpy.linalg.svd(A, full_matrices=False)
-        x = numpy.transpose(V[-1])
-
-        # Determine the angle of this line to the x-axis, -pi/2 < phi < pi/2
-        phi = numpy.arctan2(-x[0], x[1])
-    phi = (phi + math.pi / 2) % math.pi - math.pi / 2
-    return phi
+    t = transform.ScalingTransform()
```
Sabrina85: add a comment what this line does

scripts/ebeam_scan_to_multiprobe.py
```diff
@@ -171,20 +140,19 @@ def main(args):
...
+            print("Added {:.2f}° to the scan rotation using SEM PC. Rotation is now set to: {:.2f}°".format(
+                    math.degrees(sem_rotation),
+                    math.degrees(scanner.rotation.value)))
         else:
-            print("Add {:.2f}° to the scan rotation using SEM PC. If negative"
-                  "subtract this value from the scan rotation.".format(
-                math.degrees(sem_rotation)))
+            print("Add {:.2f}° to the scan rotation using SEM PC.".format(
+                    math.degrees(sem_rotation)))
+        print("Gain of the e-beam deflectors is {} pixels per volt in x and {} pixels per volt in y".format(
```
Sabrina85: Maybe we should add somewhere what the definition of gain is. I find it still confusing, as I know gain as an amplification value.

scripts/galvo_scan_to_multiprobe.py
```diff
@@ -40,56 +40,44 @@
...
+    coordinates: (list of tuples)
+        (x, y) center coordinates of the moved spot grids.
+    voltages: (list of tuples)
+        x and y offset applied to the deflectors.
 
     Returns
     -------
     phi: (float)
-        Angle between a line fitted through the coordinates and the y-axis. Angle is clockwise and in radians.
+        Deflection angle in radians. Angle is counterclockwise.
```
Sabrina85: same here: would add the 0 pos info.

scripts/galvo_scan_to_multiprobe.py
```diff
@@ -150,47 +132,41 @@ def get_galvo_rotation(ccd, channel=None, galvo_offset=3):
...
+        Experimentally determined offset of the DeScan-X galvo in Volt.
+    galvo_x_offset: (float)
         Experimentally determined offset of the DeScan-Y galvo in Volt.
 
     Returns
     -------
     galvo_rotation: (float)
         The clockwise angle, in radians, of the DeScan-Y galvo relative to the y-axis of the diagnostic camera.
+    scale: (tuple of floats)
+        The x and y scale factor between the voltages and coordinates, in pixels per volt.
```
Sabrina85: Check if gain and scale are depicting the same, that only one term is used. Otherwise it gets confusing.

tepals:
@Sabrina85 Updated, hope this clarifies the gain a bit.

PR #1264, opened 2019-11-06
driver tmcm: ._expected_do_pos initialized too late

philipwinklerdelmic:


philipwinklerdelmic:
The hardware tester didn't start because of a bug in the tmcm.
It seems like we should have caught this a long time ago. Were the leds never turned on during initialization of the tmcm in any of our systems?

philipwinklerdelmic:
Can we do it like that or do we have to be more clever? I don't really remember what the code does.

pieleric:
Indeed, I think this is enough. Typically, the hardware never has the reference switches turned on at init, so that's probably the reason it never got triggered.

PR #1267, opened 2019-11-07
xt client: added stigmator and ACB functionality

tepals:


pieleric:
Looks alright to me. Maybe you want to already provide the stigmator value as a VA (containing 2 floats) on the e-beam component?

As briefly discussed previously, we need to provide a proper API for these "auto" functionalities.
For the auto B/C, we already have the API as follow: ".applyAutoContrast" on the *detector* used to assess the brightness/contrast. It returns a Future. The detector also has a "contrast" (=gain) and "brightness" (=offset) FloatVAs with a range between 0->1.
However, currently the driver doesn't provide any Detector. It could be extended to provide such component. Possibly, but *not necessarily*, it would also be used to acquire the data via the API.

For the auto-focus (and auto-stigmator), I think they should also be dedicated methods returning a Future. However I'm a little bit confused on which component should have it. Maybe  it should go on the focus for the AutoFocus, and on the ebeam for the auto-stigmator. They could take one argument, which is the Detector to use... of course that would require to have Detectors. Also, it's not clear how this could be used if the detectors seen by the clients are only wrappers to the Detectors of this driver... Food for thought.

PR #1268, opened 2019-11-07
driver hamamatsurx: handle errors better

pieleric:


src/odemis/driver/hamamatsurx.py
```diff
@@ -1122,6 +1122,10 @@ def readCommandResponse(self):
                     # when socket timed out (receiving no response)
                     logging.debug("Timeout on the socket, will wait for more data packets.")
                     continue
+                if not returnValue:
```
Sabrina85: Maybe add a comment why we do this trick and a TODO to actually follow up on this issue.

src/odemis/driver/hamamatsurx.py
```diff
@@ -932,29 +930,36 @@ def __init__(self, name, role, port, host, children=None, dependencies=None, dae
...
+                ckwargs = children["delaybox"]
+            except Exception:
+                raise ValueError("Required child delaybox not provided")
+            self._delaybox = DelayGenerator(parent=self, daemon=daemon, **ckwargs)
+            self.children.value.add(self._delaybox)  # add delaybox to children-VA
         except Exception:
-            raise ValueError("Required child delaybox not provided")
-        self._delaybox = DelayGenerator(parent=self, daemon=daemon, **kwargs)
-        self.children.value.add(self._delaybox)  # add delaybox to children-VA
+            try:
```
Sabrina85: Add a comment why you try to end the app here. To prevent having started the App, failing to connect for some reason and then having the app still running when trying again to start odemis.

src/odemis/driver/hamamatsurx.py
```diff
@@ -109,6 +105,8 @@ def __init__(self, name, role, parent, spectrograph=None, daemon=None, **kwargs)
             # or the camera not being turned on.
             raise model.HwError("Failed to find readout camera, check it is powered. If powered, restart the Hamamatsu PC")
 
+        super(ReadoutCamera, self).__init__(name, role, parent=parent, **kwargs)  # init HwComponent
```
Sabrina85: Add a comment why we call the super method here. Raise HwError before calling super method - why did we want to do that? I think it was to ensure Odemis tries again to connect to RemoteEx. Then we had the chance to restart the Hamamatsu pc, which usually ensures that the frame grabber is working correctly again.

PR #1273, opened 2019-11-11
PIGCS driver: referencing + testcases for closed loop

Sabrina85:


src/odemis/driver/pigcs.py
```diff
@@ -1971,6 +1967,24 @@ def __init__(self, busacc, address=None, axes=None, auto_suspend=10, suspend_mod
...
+            # TODO Eric: Is there a situation where we explicitly want to pass False here even
+            #  if we have a reference switch/limits?
+            #  Is there any controller operated in closed-loop but without ref sensor?
+            self.SetReferenceMode(a, self._hasLimitSwitches[a] or self._hasRefSwitch[a])
+            if self._hasLimitSwitches[a] or self._hasRefSwitch[a]:
+                # TODO problem is that we now call it without a lock
+                #  -> we dont't wait until finished and bus gets reported that it it not referenced
+                #  cannot call doReference here as method on Bus
+                self.startReferencing(a)
+
```
Sabrina85: This doesn't work properly as we do not wait until referencing finished. Any suggestions how to implement or call at a different place?

src/odemis/driver/test/pigcs_test.py
```diff
@@ -74,6 +76,10 @@
 
 KWARGS_E725 = {"name": "test", "role": "stage", "port": "autoip", "axes": CONFIG_BUS_E725}
 
+# testing with real HW on SONIC PIGCS-C-867
+KWARGS_TWO_CL_REAL = {"name": "test_real_pigcs", "role": "stage2d", "port": "autoip", "axes": CONFIG_BUS_TWO_CL_REAL,
```
tepals: you created this but are not using it anywhere?

src/odemis/driver/test/pigcs_test.py
```diff
@@ -601,6 +607,44 @@ def setUp(self):
...
+        Test referencing for 2 axes.
+        """
+        stage = CLASS(**self.kwargs_two)
+
+        # Test proper referencing
+        self.assertTrue(stage.referenced.value['x'])
+        self.assertTrue(stage.referenced.value['y'])
+        pos = stage.position.value
+        # self.assertFalse(stage.referenced.value['x'])
+        # stage.moveRel({'x': 0.0001}).result()  # move to position different from 0
```
tepals: left over from debug?

src/odemis/driver/pigcs.py
```diff
@@ -1971,6 +1967,25 @@ def __init__(self, busacc, address=None, axes=None, auto_suspend=10, suspend_mod
             self._encoder_mng[a] = t
             t.start()
 
+            # TODO: moved ref at init to Bus class for now as a hack to work around the below mentioned problem
+            # TODO: Check if there was any need to send the RON command via SetReferenceMode in the startServo method.
+            #  @Eric was there any reason for this? It seems to work fine in the tests without.
```
pieleric: IIRC, turning off the servo automatically sets the referenced state in the controller to unreferenced (which is fair, as from there on, the hardware might physically move without the controller noticing, which would cause a shift in the referencing position). So that's why every time we restart the servo we have to convince the hardware that it's referenced (in relative moves). That said, now, we also support suspending with just setting the PID to 0, and leaving the servo on. We could do this unconditionally if the hardware has been *really* referenced.

src/odemis/driver/pigcs.py
pieleric: Anyway, as a basic rule, we never move the hardware without the user explicitly requesting it. So normally we never do any referencing at driver initialisation. (In the case of this microscope, if it's actually always safe to do, we can still do it early via the GUI or a new special init_on_ref argument).

I'd suggest the following. The driver distinguishes between two kinds of hardware: the one which can reference (ie, has a Limit or Ref switch), and the one which cannot. For the later case, the code should behave exactly as before (eg, RON False all the time, and reports it cannot be referenced), and therefore we should be confident not breaking behaviour on existing hardware (see note below though). In the case the hardware is referenceable:
* at init, and until it's been successfully referenced, it should behave like non-referenceable hardware. When isReferenced() is called, it should return False (so you might need to *not* call FRF?, but just store the status internally)
* After being referenced, it should use "RON True" (or maybe not needed, as referencing already does it?) and isReferenced() should report it's been referenced.
Maybe at init it's possible to check whether the hardware has already been referenced or not, using a command like "RON?" . In the worse case, I'd suggest to assume at init that it's not referenced, and hopefully in reality, the backe-end is sufficiently rarely restarted that's it'd become an issue.

Note: some of the current hardware might report the presence of reference or limit switches *even if* they don't support referencing (because physically they have the switch, it just doesn't work). We need to check. If so, then we'll need to bring some extra cleverness...

src/odemis/driver/pigcs.py
```diff
@@ -2037,8 +2052,6 @@ def _startServo(self, axis):
                 self.SetParameter(axis, 0x56, 1, check=False)  # 1 = on
                 time.sleep(2)  # 2 s seems long enough for the encoder to initialise
             self.SetServo(axis, True)
-            # To allow (relative) moves, even if it's not actually referenced
-            self.SetReferenceMode(axis, False)
```
pieleric: Without this, it'll be impossible to move again the hardware. So it should stay as-is. In your case, is this code actually called (ie, is `self._servo_suspend` True or False?). If so, we probably want to adjust the computation of `_servo_suspend` to be always False if the axis is referenceable. 
Note that this code might still be called in case some error happened during a move (eg, reached a limit, or it failed to reach the target position). I'd suggest that for now we reset the referencing state to False, for safety. If it turns out some some error can be handled without resetting the referencing, we could handle them explicitly, when we face them.

src/odemis/driver/pigcs.py
```diff
@@ -2878,9 +2894,11 @@ def __init__(self, name, role, port, axes, baudrate=38400,
         # TODO also a rangesRel : min and max of a step
         speed = {}
         referenced = {}
+        controller_list = []
```
pieleric: Unused? (only written)

src/odemis/driver/pigcs.py
```diff
@@ -4224,9 +4299,23 @@ def _processCommand(self, com):
                 axis = int(args[1])
                 if axis != 1:
                     raise SimulatedError(15)
+
+                # simulate moving to reference position
+                ref_pos = self._parameters[0x16] * 1e3  # value at reference
+                # TODO I have not idea why it is necessary to multiply by *1e3,
+                #  but saw it also happen in the MOV commands that values were in mm instead of m.
```
pieleric: Yes, exactly, the parameters are stored in mm, but internally the computations are done in m.

src/odemis/driver/pigcs.py
```diff
@@ -4224,9 +4299,23 @@ def _processCommand(self, com):
...
+                logging.debug("Simulating a referencing move of %f s", duration)
+                self._start_move = time.time()
+                self._end_move = self._start_move + duration
+                self._position = cur_pos
+                self._target = ref_pos
                 self._referenced = 1
-                self._end_move = 0
-                self._position = self._parameters[0x16] # value at reference
+
+                # self._target = self._parameters[0x16] * 1e3
```
pieleric: drop?

src/odemis/driver/test/pigcs_test.py
```diff
@@ -601,6 +602,47 @@ def setUp(self):
...
+        stage.moveAbs({'x': 0.015}).result()
+        stage.moveAbs({'y': 0.005}).result()
+        self.assertAlmostEqual(stage.position.value['x'], 0.015, places=3)  # 0.001 times out
+        self.assertAlmostEqual(stage.position.value['y'], 0.005, places=3)
+        self.assertNotAlmostEqual(stage.position.value['x'], pos["x"], places=3)
+
+        stage.reference({'x'}).result()
+        self.assertTrue(stage.referenced.value['x'])
+        # The reference position should be 12.5 mm for simulator
+        self.assertAlmostEqual(stage.position.value["x"], 0.0125, places=3)
```
pieleric: The problem with such check is that it only works with the simulator. We could try to read the parameter, and convert it using the *right* conversion factor. Alternatively, you could just do it twice (with a move in between) and check it's the same value.

src/odemis/driver/pigcs.py
Sabrina85: For now the `_servo_suspend` is always False if we are in "read" mode and True if in "full" mode. So would you suggest to have a differentiation in "full" mode between referenceable and not referenceable?

pieleric: I think we should just always call `self.SetReferenceMode(axis, False)` here. If we just started the servo, either it's because we just initialised, so the axis is not referenced yet, or there was a stop servo, so it's not referenced. The `startReference()` function will switch the mode when needed.

src/odemis/driver/pigcs.py
```diff
@@ -1865,6 +1860,8 @@ def __init__(self, busacc, address=None, axes=None, auto_suspend=10, suspend_mod
         self._pos_lock = {}  # acquire to read/write position
         self._slew_rate = {}  # in s, copy of 0x7000002: slew rate, for E-861
 
+        self._referenced = False   # tracking if the axis is actually referenced and not only pretending to be
```
pieleric: Please add a note that RON causes FRF? to always return True.

src/odemis/driver/pigcs.py
```diff
@@ -3070,16 +3101,57 @@ def moveAbs(self, pos, update=False):
...
+                    self.referenced._value[a] = False
+                    controller.startReferencing(channel)
+                    self._waitEndMove(future, (a,), time.time() + 100)  # block until it's over
+                    self.referenced._value[a] = True
+            except CancelledError:
+                # FIXME: if the referencing is stopped, the device refuses to
+                # move until referencing is run (and successful).
+                # => Need to put back the device into a mode where at least
+                # relative moves work.
+                logging.warning("Referencing cancelled, device will not move until another referencing")
```
pieleric: That looks like a left-over from copy-pasting, but I don't think this is correct for the PI controllers?

src/odemis/driver/test/pigcs_test.py
```diff
@@ -39,7 +40,7 @@
 # Export TEST_NOHW=1 to force using only the simulator and skipping test cases
 # needing real hardware
 TEST_NOHW = (os.environ.get("TEST_NOHW", 0) != 0)  # Default to Hw testing
-
+TEST_NOHW=1
```
pieleric: Please drop eventually.

src/odemis/driver/test/pigcs_test.py
```diff
@@ -601,6 +602,48 @@ def setUp(self):
...
+        self.assertAlmostEqual(stage.position.value['x'], ref_pos_x + stage.axes["x"].range[1] * 0.1, places=3)
+        self.assertAlmostEqual(stage.position.value['y'], ref_pos_y + stage.axes["y"].range[1] * 0.2, places=3)
+
+        # reference again
+        stage.reference({'x'}).result()
+        stage.reference({'y'}).result()
+        self.assertTrue(stage.referenced.value['x'])
+        self.assertTrue(stage.referenced.value['y'])
+        self.assertAlmostEqual(stage.position.value["x"], ref_pos_x, places=3)
+        self.assertAlmostEqual(stage.position.value["y"], ref_pos_y, places=3)
```
pieleric: Please also check the referencing of multiple axes. Like: `stage.reference({"x", "y"})`.

src/odemis/driver/pigcs.py
```diff
@@ -2396,24 +2413,31 @@ def stopMotion(self):
...
+        #  exactly would I expect to not work?
+
+        # So referencing could look like:  TODO remove these lines when ref implemented?
         # ron 1 1
         # frf -> go home and now know the position officially
         # orig_pos = pos?
 
         if self._hasRefSwitch[axis]:
-            self.SetReferenceMode(axis, True)
+            # self.SetReferenceMode(axis, True)  # TODO think it is unnecessary
```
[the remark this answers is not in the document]
Sabrina85: We already do this in `_startServo`. And it is only necessary to do it once. That's why I think this line is not necessary again. What do you think?

src/odemis/driver/pigcs.py
```diff
@@ -2396,24 +2413,31 @@ def stopMotion(self):
...
             self.ReferenceToSwitch(axis)
+            # FIXME: If referencing fails, this flag might be True though it should be not.
+            while self.isMoving(): # wait until move is done  @Eric does this work? or use future for ReferencetoSwitch call?
+                time.sleep(0.01)
```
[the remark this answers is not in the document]
Sabrina85: Also remove the FIXME commend?

src/odemis/driver/pigcs.py
```diff
@@ -2435,7 +2460,13 @@ def isReferenced(self, axis):
         if not self._hasRefSwitch[axis] and not self._hasLimitSwitches[axis]:
             return None
         else:
-            return self.IsReferenced(axis)
+            # Check flag referenced and HW report the same status for referencing
+            if self.IsReferenced(axis) == self._referenced:
+                return self._referenced
+            else:
+                logging.warning("Axis reports being referenced though it is not.")
+                self._referenced = False  # make sure this variable is False if this happens
```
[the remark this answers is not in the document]
Sabrina85: Can we discuss this?

src/odemis/driver/pigcs.py
```diff
@@ -2933,8 +2936,8 @@ def __init__(self, name, role, port, axes, baudrate=38400,
         else:
             self._pos_updater = None
 
-        # RO VA dict axis -> bool: True if the axis has been referenced
-        # Only axes which can be referenced are listed
+        # # RO VA dict axis -> bool: True if the axis has been referenced
+        # # Only axes which can be referenced are listed
```
pieleric: Don't double comment -> restore to the previous lines.

src/odemis/driver/pigcs.py
```diff
@@ -2396,24 +2389,22 @@ def stopMotion(self):
...
         # with TMN/TMX and some correct moves will fail.
-        # So referencing could look like:
-        # ron 1 1
-        # frf -> go home and now know the position officially
-        # orig_pos = pos?
 
         if self._hasRefSwitch[axis]:
             self.SetReferenceMode(axis, True)
             self.ReferenceToSwitch(axis)
+            # FIXME: If referencing fails, this flag might be True though it should be not.
```
pieleric: That's true, but as it's also compared with IsReferenced(), I don't think it's too big an issue. At least, as long as the startServo() (and so RON) is not called, that should be detected anyway, and the `_referenced` flag reset to False.

Sabrina85:
@pieleric ready to merge?

pieleric:
Almost ready to merge, there are still 3 comments unresolved. Could you have a look at them (check the github page)

PR #1274, opened 2019-11-12
Andor usb compliance

philipwinklerdelmic:
Still a work in progress. Main issue is that if an error occurs during startup, it seems like we connect to the wrong andor hardware (serial number 'x').

The reconnect takes 2 minutes, which is a bit excessive.

### DIFF
--- a/src/odemis/driver/andorshrk.py
+++ b/src/odemis/driver/andorshrk.py
@@ -81,6 +81,28 @@ SHUTTER_OPEN = 1
 SHUTTER_BNC = 2  # = driven by external signal
 
 
+def callWithReconnect(fun):
+    """
+    Wrapper for functions that communicate with the shamrock.
+    In case the connection is lost (e.g. due to em-interference), we
+    try to reconnect and send the command again.
+    """
+    def wrapper(self, *args, **kwargs):
+        try:
+            return fun(self, *args, **kwargs)
+        except ShamrockError as ex:
+            if self._reconnecting:
+                raise
+            else:
+                logging.error("Communication with shamrock failed with "
+                              "exception '%s', trying to reconnect." % ex)
+                if ex.errno in (20201, 20275):  # COMMUNICATION_ERROR / NOT_INITIALIZED
+                    self._reconnect()
+                    return fun(self, *args, **kwargs)
+                else:
+                    raise
+    return wrapper
+
 class ShamrockError(IOError):
     def __init__(self, errno, strerror, *args, **kwargs):
         super(ShamrockError, self).__init__(errno, strerror, *args, **kwargs)
@@ -114,8 +136,6 @@ class ShamrockDLL(CDLL):
                 #  are still valid)
                 CDLL.__init__(self, "libatspectrograph.so.2", RTLD_GLOBAL)
 
-        self.device_lock = threading.RLock()
-
     def at_errcheck(self, result, func, args):
         """
         Analyse the return value of a call and raise an exception in case of
@@ -124,11 +144,11 @@ class ShamrockDLL(CDLL):
         """
         # everything returns SHAMROCK_SUCCESS on correct usage
         if result not in ShamrockDLL.ok_code:
-            errmsg = create_string_buffer(ERRORLENGTH).decode('latin1')
+            errmsg = create_string_buffer(ERRORLENGTH)
             self.ShamrockGetFunctionReturnDescription(result, errmsg, len(errmsg))
             raise ShamrockError(result,
                                 "Call to %s failed with error %d: %s" %
-                                (func.__name__, result, errmsg.value))
+                                (func.__name__, result, errmsg.value.decode('latin1')))
         return result
 
     def __getitem__(self, name):
@@ -335,8 +355,7 @@ class Shamrock(model.Actuator):
         self._slit_names.update(slits)
 
         self.Initialize()
-
-        self._dev_id = device
+        self._reconnecting = False
 
         try:
             if isinstance(device, basestring):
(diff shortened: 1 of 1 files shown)


src/odemis/driver/andorshrk.py
```diff
@@ -623,6 +654,54 @@ def Initialize(self):
...
+        """
+        self.state._set_value(HwError("Shamrock disconnected"), force_write=True)
+        self._reconnecting = True
+        logging.debug("Reconnecting Shamrock...")
+        self.Close()
+
+        # In order to make reestablish the connection, we need to turn the power off and on again and then
+        # reinitialize.
+        logging.debug("Cycling power...")
+        if self._power_supplier:
```
philipwinklerdelmic: @pieleric Do you think this is a good idea to implement power cycling? Maybe we can also to it for the pmtctrl and tmcm instead of uhubctl. For this driver, it seems necessary to do this.

philipwinklerdelmic:
OK, the problem with the connection on startup is solved now. Everything seems to be working. Still there are some minor issues:
* time of reconnection: the spectrograph seems to be ready after about a minute, but we wait for 150s in `Initialize()`.
* If we move with `odemis-cli` and we have to reconnect, we get an error message telling us that the move failed, even though it worked after reconnecting

I don't think this is a big issue, considering that the reconnection only happens very rarely (up to now only during the emc test and when we willfully try to break it with sparks).

Another point to discuss might be the way we do the power cycling. I implemented it now with the power control unit, not with `uhubctl`. This means we still rely on odemis code and expect that the power control is not going to break during the emc test. On the other hand, we don't need any third-party software and this seems like a nice and easy solution.

We might even consider implementing this for the reconnect-function of the tmcm and pmt if the current implementation doesn't work (sometimes, very rarely, the reconnect implementation didn't work and we had to do a power cycle even on the tmcm and pmt).

src/odemis/driver/andorshrk.py
pieleric: The main problem with the power cycling is that it takes so much time (90 s in the bad cases). So it would be good to find a faster alternative. Did you check if using the uhubctl it is possible to find the device again? If so, it would be worth implementing. There are other types of USB disconnection in Linux (eg via /sys/), they might also be worth checking too as might be easier to implement.
I think that you could also have several of the approaches implented, and they would be use as fallback for one another.

philipwinklerdelmic: Even if we do it with `uhubctl`, we need to reinitialize the spectrograph otherwise we get the 'device not initialized' error back from the hardware. This takes quite a long time anyway.

src/odemis/driver/andorshrk.py
```diff
@@ -114,6 +137,8 @@ def __init__(self):
                 #  are still valid)
                 CDLL.__init__(self, "libatspectrograph.so.2", RTLD_GLOBAL)
 
+        self.device_lock = threading.RLock()
```
pieleric: Unused

src/odemis/driver/andorshrk.py
```diff
@@ -122,11 +147,11 @@ def at_errcheck(self, result, func, args):
         """
         # everything returns SHAMROCK_SUCCESS on correct usage
         if result not in ShamrockDLL.ok_code:
-            errmsg = create_string_buffer(ERRORLENGTH).decode('latin1')
+            errmsg = create_string_buffer(ERRORLENGTH)
             self.ShamrockGetFunctionReturnDescription(result, errmsg, len(errmsg))
             raise ShamrockError(result,
                                 "Call to %s failed with error %d: %s" %
-                                (func.__name__, result, errmsg.value))
+                                (func.__name__, result, to_str_escape(errmsg.value)))
```
pieleric: That looked better before. The error message should be a english message, so it shouldn't be escaped. I think it'd typically be ascii, but using latin1 is a safe bet.

src/odemis/driver/andorshrk.py
```diff
@@ -333,27 +358,27 @@ def __init__(self, name, role, device, camera=None, accessory=None,
...
+                    self._power_supplier = kwargs['power_supplier']
+                    self._reconnect(on_startup=True)
+                else:
+                    # Hope everything is fine and the spectrograph just has a weird model number
+                    logging.info("Cannot reconnect because spectrograph doesn't have a power supplier.")
+                # Try again, if we still don't recognize the model, the spectrograph is probably fine and
+                # we just don't recognize the model number.
+                self._model, fl = self._get_model(name, device)
+                if not self._model:
+                    logging.warning("Untested spectrograph with focus length %d mm", fl * 1000)
```
pieleric: I'd rather avoid extra turning on/off at startup. If it's messed-up at start-up, let the user handle it. There might be a lot of other reason than this issue.

src/odemis/driver/andorshrk.py
pieleric: Every device which as a power_supplier has a `powerSupply` VA, which can be set False/True. => Use that VA.

src/odemis/driver/andorshrk.py
```diff
@@ -623,6 +677,45 @@ def Initialize(self):
...
+            self._power_supplier.supply({'Spectrograph': False}).result()
+            time.sleep(2)  # wait a bit, otherwise the system doesn't notice
+            self._power_supplier.supply({'Spectrograph': True})  # this takes 90s, don't wait for .result()
+            time.sleep(2)
+        else:
+            raise ValueError("Spectrograph doesn't have a power supplier, aborting reconnect.")
+
+        # Initialization
+        logging.debug("Turned power off and on, now initializing...")
+        self.Initialize()  # blocking, takes 2 min
```
pieleric: If the power has been on for 90s, hopefully the `Initialize()` should be fast.

src/odemis/driver/andorshrk.py
```diff
@@ -623,6 +677,45 @@ def Initialize(self):
...
+        # If the function is called on startup, we are not ready to call ._updatePosition yet
+        if not on_startup:
+            try:
+                self._updatePosition()
+                logging.debug("Restarting spectrograph after power cycling was successful.")
+            except ShamrockError as ex:
+                logging.error("Unable to restart spectrograph. Try to turn the power off and on again. Failed "
+                              "with exception %s." % ex)
+            self.state._set_value(model.ST_RUNNING, force_write=True)
+        self._reconnecting = False
```
pieleric: Put this in a try/finally.

src/odemis/driver/andorshrk.py
```diff
@@ -757,6 +859,8 @@ def GetGratingInfo(self, grating):
 
         return Lines.value * 1e3, Blaze.value.decode('latin1'), Home.value, Offset.value
 
+    # don't call with reconnect, it sometimes might raise a 20201 error that's not related
+    # to a lost connection
```
pieleric: I think it should still try to reconnect, especially as it's one of the most common function. Maybe only call reconnect if failed 5 times.

src/odemis/driver/andorshrk.py
```diff
@@ -777,14 +850,15 @@ def SetWavelength(self, wavelength):
                     self._dll.ShamrockSetWavelength(self._device, c_float(wavelength * 1e9))
                 except ShamrockError as ex:
                     if ex.errno != 20201 or retry >= 5: # SHAMROCK_COMMUNICATION_ERROR
-                        raise
+                        self._reconnect()
```
pieleric: Only if errno is  in `(20201, 20275)`, reconnect, otherwise raise.

philipwinklerdelmic: I'm not sure if this is going to work. I think it sometimes also raises the 20201 error that has nothing to do with reconnect.

philipwinklerdelmic:
updated

PR #1275, opened 2019-11-13
symphotime python3 bug

philipwinklerdelmic:


philipwinklerdelmic:
self.data is added to a byte at some point, so we shouldn't convert it to a (unicode) string. There are probably more issues in this driver, but we can only test it when the server is on (which seems to be mostly during nighttime).
@amuskens Can you maybe figure out the times when the server is up? How did you do the testing with python 2?

pieleric:
Philip, the symphotime (fake) server has to be run by you, on your computer (the picoquant server). My guess is that your automatic test script takes care of running and stopping the server. As you run it during the evening, you only see during the evening!

philipwinklerdelmic:
Ok, that makes more sense! I didn't connect the dots here.

I'll have another look and will fix it.

philipwinklerdelmic:
All the testcases are passing now.

src/odemis/driver/symphotime.py
```diff
@@ -233,7 +233,7 @@ def to_bytes(self):
             for val in self.data:
                 output_string += struct.pack('I', val)
         elif self.typ == PQ_OPT_DATATYPE_FIXED_LENGTH_STRING:
-            output_string += struct.pack('<BH', self.typ, len(self.data)) + self.data
+            output_string += struct.pack('<BH', self.typ, len(self.data)) + self.data.encode('ascii')
```
pieleric: This may be used to encode non-ascii characters (eg, filename). I am not sure what's the expected encoding on the Symphotime server side, but I think utf-8 would be a good bet.
So, encode in "utf-8" and pass the length of the *encoded* data.

src/odemis/driver/symphotime.py
```diff
@@ -856,7 +856,7 @@ def _actOnMessage(self,decoded_msg):
                 logging.warning("Measurement data frame received while not in measuring state. ")
 
             # convert the OptionalDataRecord type to a regular python dictionary for logging
-            data = {r.name : r.data for r in decoded_msg.odOptional.values()}
+            data = {r.name.decode('latin1') : r.data for r in decoded_msg.odOptional.values()}
```
pieleric: Instead of decoding here, I'd suggest to decode in `from_bytes()`. Maybe use "utf-8" with replace error handler, to be symmetrical with the encoding used in `to_bytes()`

philipwinklerdelmic:
updated

PR #1276, opened 2019-11-18
galvo scan to multiprobe fixed, to properly deal with galvo offsets a…

tepals:
…nd switching of the AWG cables.

- galvo offsets are converted to floats from terminal input and added instead of subtracted.
- removed channel input argument, that is now handled inside the scanwaveform code.
- sem gain is now also printed.


scripts/galvo_scan_to_multiprobe.py
```diff
@@ -213,11 +213,13 @@ def main(args):
         raise ValueError("Backend is not running.")
 
     ccd = model.getComponent(role=options.role)
+    galvo_x_offset = float(options.galvo_x_offset)
```
Sabrina85: I would move these lines into the `try` statement.

PR #1282, opened 2019-11-28
gui comp: better mouse capture when dragging

pieleric:


src/odemis/gui/comp/viewport.py
```diff
@@ -1366,6 +1370,9 @@ def on_drag_start(self, evt):
     def on_drag_end(self, evt):
         """ End the dragging procedure """
         self._dragging = False
+        # Make sure we stop tracking the mouse
+        if evt.EventObject.HasCapture():
+            evt.EventObject.ReleaseMouse()
```
amuskens: I think i tried this in Linux and lost control of the mouse completely somehow - but maybe I did it in the wrong sequence.

src/odemis/gui/util/__init__.py
```diff
@@ -144,6 +144,42 @@ def wxwrapper(f):
...
+    def release_mouse_on_drag(obj):
+        """
+        Releases the mouse when acquired via capture_mouse_on_drag()
+        """
+        if not obj.HasCapture():
+            logging.info("Mouse was not captured, so not releasing it")
+            return
+
+        obj.ReleaseMouse()
+
```
amuskens: Ah good. This addresses my last comment. Perfect.

PR #1283, opened 2019-11-28
config: fix windows installation when using anaconda

pieleric:


doc/develop/extending.rst
```diff
@@ -160,121 +158,88 @@ Setting up the development environment on Windows
...
+Download and install `Microsoft Visual C++ Compiler for Python 2.7 <https://www.microsoft.com/download/details.aspx?id=44266>`_.
+If you use Python 3.6, download and install `Build Tools for Visual Studio 2019 <https://www.visualstudio.com/downloads/#build-tools-for-visual-studio-2019>`_. 
 
-Now copy these files:
-``rc.exe`` and
-``rc.dll``
+You can finally install pylibtiff using::
 
-from ``C:\Program Files (x86)\Windows Kits\8.1\bin\x86``
+   pip install libtiff
```
tepals: It might not be completely clear that you need to do this in the activated conda env.

doc/develop/extending.rst
```diff
@@ -313,7 +278,7 @@ scripts, or via a command-line interface.
 #. Install using ``pip install libtiff-0.4.2-cp27-cp27m-win_amd64.whl`` (or ``-win32``),
    downloaded from http://www.lfd.uci.edu/~gohlke/pythonlibs/#pylibtiff
 
-#.  Install OpenCV using ``pip install opencv_python-2.4.13.5-cp27-cp27m-win_amd64.whl`` (or ``-win32``),
+#.  Install OpenCV using ``conda install opencv -c conda-forge`` (or ``-win32``),
```
tepals: What does the (or `-win32`) mean here? Should you change it to ``conda install opencv -c conda-forge -win32`` or ``conda install opencv -win32``

requirements.txt
```diff
@@ -6,14 +6,14 @@ h5py==2.5.0
...
 pycairo>=1.10.0
 PyInstaller==3.5
 psutil>=4.3.0
-Pyro4==4.18
 PyYAML>=3.11
 pyzmq>=15.2.0
 scipy>=0.15.1
 wxPython>=4.0.4
 setuptools==39.2.0
+Pyro4==4.18
```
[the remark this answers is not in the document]
pieleric: It's a bit of a hack, but the problem is that we need a special version of Pyro4. So when using anaconda, the user has to remove this line first... that's why I've moved it explicitly at the end.
Maybe installing our own version of Pyro4 first, because installing from requirements.txt works too... I'll check next time we do an installation on Windows.

doc/develop/extending.rst
pieleric: `-win32` is just some meaningless left-over.

PR #1285, opened 2019-12-03
acq stream StaticARStream: warn if a combination of pos+polarization is missing

pieleric:
Normally for every polarization, we acquire every positions. However,
there are two reason this doesn't happen:
 * Something went wrong the acquisition file
 * Some positions are slightly different (due to floating point error),
   and not detected as similar.

The stream doesn't really mind, but if a non-existing combination is
requested, it's obviously going to fail. So let's try to warn it early,
and if that due to the second case, we might eventually try to make a
more clever code.

### DIFF
--- a/src/odemis/acq/stream/_static.py
+++ b/src/odemis/acq/stream/_static.py
@@ -27,18 +27,18 @@ from __future__ import division
 
 from past.builtins import basestring, long
 import collections
-import logging
+import copy
 import gc
+import logging
 import math
 import numpy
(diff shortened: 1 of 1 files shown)


pieleric:
Actually, I've seen an error most probably due to this in a system log, but that wasn't related to such a bug and the customer didn't attach the acquisition file. So I'm not exactly what caused it. 
At least, with this check, if this happens again, we should find out more easily.

PR #1288, opened 2019-12-04
Save all metadata for every acquisition

philipwinklerdelmic:


src/odemis/acq/__init__.py
```diff
@@ -396,6 +398,14 @@ def _adjust_metadata(self, raw_data):
                 if model.MD_DESCRIPTION not in d.metadata:
                     d.metadata[model.MD_DESCRIPTION] = s.name.value
 
+        # Add extra settings to metadata
+        for s, data in raw_data.items():
+            for d in data:
+                if hasattr(s, '_md_manager') and s._md_manager:
+                    d.metadata[model.MD_EXTRA_SETTINGS] = s._md_manager.all_metadata
+                else:
+                    logging.warning("Stream %s has no metadata manager, not saving extra metadata.", s)
```
pieleric: This is done only after the acquisition of all the streams. It should be done just after the stream acquisition, and *before* the acquisition of the next stream.

src/odemis/acq/__init__.py
```diff
@@ -38,6 +38,7 @@
     ScannedRemoteTCStream, ScannedTCSettingsStream
 from odemis.util import img, fluo, executeAsyncTask
 import time
+import resource
```
pieleric: Not needed?

src/odemis/acq/stream/_sync.py
```diff
@@ -80,7 +80,7 @@ class MultipleDetectorStream(with_metaclass(ABCMeta, Stream)):
     the .raw from the sub-streams.
     """
 
-    def __init__(self, name, streams):
+    def __init__(self, name, streams, md_manager=None):
```
pieleric: If you use the stream.acquire() function to store the metadata, you don't need to pass the md_manager to every stream (which would be a good thing), and instead just pass it to the `acquire()` function.

src/odemis/acq/__init__.py
philipwinklerdelmic: We actually do need it (see l. 456). On ubuntu, there is a limit on the number of files that can be opened at the same time. This also seems to apply to VAs. This limit is by default 1024, which is not enough if we want to listen to all VAs, so the gui doesn't start and I get an error message. Typing `ulimit -n 2048` in the command line works and the python resource function does the same.

src/odemis/acq/stream/_sync.py
philipwinklerdelmic: indeed, that's much better!

philipwinklerdelmic:
updated

src/odemis/gui/cont/streams.py
```diff
@@ -544,6 +548,28 @@ def _on_metadata_btn(self, evt):
...
                     text += u"%s: %s\n" % (key, v)
 
+        # only display extra settings once
+        if model.MD_EXTRA_SETTINGS in raw[0].metadata:
+            text += u"\n======================================\nHardware Settings\n"
+            text += u"======================================\n"
+            for comp, vas in md[model.MD_EXTRA_SETTINGS].items():
+                if vas != {}:
+                    text += u"Component %s:\n" % comp
+                for name, value in vas.items():
```
pieleric: I think it's a little confusing that a "value" is a "data" + "unit". Maybe rename "value" to "content" and "data" to "value"?

src/odemis/gui/cont/streams.py
```diff
@@ -544,6 +548,28 @@ def _on_metadata_btn(self, evt):
...
+                    data = value[0]
+                    unit = value[1] or ""  # don't display 'None'
+                    unit = unit if data else ""  # don't display unit if data is None (None Hz doesn't make sense)
+                    if isinstance(data, dict):
+                        if value != {}:
```
pieleric: Probably `value` -> `data` ? (not if you rename the variable ;-) )
Also you should be able to write `if data:`.

src/odemis/gui/cont/acquisition.py
```diff
@@ -742,7 +742,7 @@ def on_acquisition(self, evt):
         self._tab_panel.Layout()  # to put the gauge at the right place
 
         # start acquisition + connect events to callback
-        self.acq_future = acq.acquire(self._tab_data_model.acquisitionStreams)
+        self.acq_future = acq.acquire(self._tab_data_model.acquisitionStreams, self._main_data_model.md_manager)
```
pieleric: You also need to update the code for the SECOM. See the other callers of acq.acquire (I think it's in win.acquisition).
The plugins should also be updated. That's a bit of a pain. (and maybe it's a sign we should have merged the MetadataManager into a large "acquisition manager"... but might be better done later, in a refactoring task)

src/odemis/model/_metadata.py
```diff
@@ -194,3 +194,5 @@
 MD_GAIN_P = "Proportional gain"  # float
 MD_GAIN_I = "Integral gain"  # float
 MD_GAIN_D = "Derivative gain"  # float
+
+MD_EXTRA_SETTINGS = "Extra settings"
```
pieleric: Indicate the purpose and type: "string containing a dict encoded in JSON, which represents all the known state of all the hardware used during an acquisition"

src/odemis/acq/__init__.py
philipwinklerdelmic: I'm not really sure where to put it. Is the `Instantiator` in `modelgen` a good place? Or in `model.core`? What do you suggest?

philipwinklerdelmic:
updated

src/odemis/acq/__init__.py
```diff
@@ -315,6 +321,15 @@ def run(self):
...
                     logging.warning("Future of %s didn't return a list of DataArrays, but %s", s, das)
                     das = []
+
+                # Add extra settings to metadata
+                if self._settings_obs:
+                    settings = self._settings_obs.get_all_settings()
+                    for da in das:
+                        da.metadata[model.MD_EXTRA_SETTINGS] = copy.deepcopy(settings)
+                else:
+                    logging.warning("Acquisition task for stream %s has no metadata manager " % s,
```
pieleric: This message needs to be updated. Anyway, it's now the same for all the streams, so probably should just be put at the beginning (or the end), only once.

src/odemis/gui/model/__init__.py
```diff
@@ -262,13 +262,15 @@ def __init__(self, microscope):
         # The microscope object will be probed for common detectors, actuators, emitters etc.
         if microscope:
             self.role = microscope.role
+            self._comps_with_role = []
```
pieleric: Don't store it in an attribute, but just in a variable `comps_with_role`.

philipwinklerdelmic:
updated

src/odemis/gui/cont/streams.py
```diff
@@ -544,6 +548,28 @@ def _on_metadata_btn(self, evt):
...
+        if model.MD_EXTRA_SETTINGS in raw[0].metadata:
+            text += u"\n======================================\nHardware Settings\n"
+            text += u"======================================\n"
+            for comp, vas in md[model.MD_EXTRA_SETTINGS].items():
+                if vas:
+                    text += u"Component %s:\n" % comp
+                for name, content in vas.items():
+                    value = content[0]
+                    unit = content[1] or ""  # don't display 'None'
+                    unit = unit if value else ""  # don't display unit if data is None (None Hz doesn't make sense)
```
pieleric: If value == 0, we should still show a unit. => check for `not None` explicitly.

src/odemis/gui/cont/streams.py
```diff
@@ -544,6 +548,28 @@ def _on_metadata_btn(self, evt):
...
                     text += u"%s: %s\n" % (key, v)
 
+        # only display extra settings once
+        if model.MD_EXTRA_SETTINGS in raw[0].metadata:
+            text += u"\n======================================\nHardware Settings\n"
+            text += u"======================================\n"
+            for comp, vas in md[model.MD_EXTRA_SETTINGS].items():
+                if vas:
+                    text += u"Component %s:\n" % comp
+                for name, content in vas.items():
```
pieleric: You could write `for name, (value, unit) in ...`, to simplify the lines just below.

pieleric: Please put this entire block in a try/except so that if the metadata is not the expected format it will just skip showing the component, and move on.

src/odemis/acq/__init__.py
```diff
@@ -38,6 +38,9 @@
     ScannedRemoteTCStream, ScannedTCSettingsStream
 from odemis.util import img, fluo, executeAsyncTask
 import time
+import resource
```
pieleric: Drop this import.

src/odemis/acq/__init__.py
```diff
@@ -433,3 +448,35 @@ def cancel(self, future):
...
+
+HIDDEN_VAS = ['children', 'dependencies', 'affects', 'alive', 'state', 'ghosts']
+class SettingsObserver(object):
+    """
+    Class that listens to all settings, so they can be easily stored as metadata
+    at the end of an acquisition.
+    """
+
+    def __init__(self, components):
+        self._components = components
```
pieleric: I don't think you need to hold a reference to `components`. => Drop this line.

philipwinklerdelmic:
updated

src/odemis/gui/cont/streams.py
```diff
@@ -544,6 +548,31 @@ def _on_metadata_btn(self, evt):
...
+                                text += u"\t%s:\n" % name
+                                for key, val in value.items():
+                                    text += u"\t\t%s: %s %s\n" % (key, val, unit)
+                            else:
+                                # still display the VA, might be interesting (e.g. that no axis was referenced)
+                                text += u"\t%s: {}\n" % name
+                        else:
+                            text += u"\t%s: %s %s\n" % (name, value, unit)
+                except Exception as ex:
+                    logging.warning("Couldn't display metadata for component %s, failed with exception %s" % (comp, ex))
```
pieleric: You can keep the message a little shorter with just "Couldn't display metadata for component %s: %s".

philipwinklerdelmic:
updated

PR #1289, opened 2019-12-05
acq stream test: fix exposure time setting

pieleric:
Now that there is the special integrationTime VA on the SPARC streams,
we have to be more careful on how the exposure time is set.
When exposureTime is not passed as detvas, integrationTime is
automatically created and used. So the raw detector .exposureTime is
*never* used, as we used to assume previously.
This caused many test cases to use the default exposureTime of 1s,
instead of way smaller values. So the test cases were very long to run.

Also merge/fix some of the live update test cases.
Also fix some of the integrationTime test cases.

Note: there is still a bug when live update is used at the same time at
integrationTime (on the metadata)

### DIFF
--- a/src/odemis/acq/test/stream_test.py
+++ b/src/odemis/acq/test/stream_test.py
@@ -952,14 +952,15 @@ class SPARCTestCase(unittest.TestCase):
 
         # Create the stream
         sems = stream.SEMStream("test sem", self.sed, self.sed.data, self.ebeam)
-        ars = stream.ARSettingsStream("test ar", self.ccd, self.ccd.data, self.ebeam)
+        ars = stream.ARSettingsStream("test ar", self.ccd, self.ccd.data, self.ebeam,
+                                      detvas={"exposureTime"})
         sas = stream.SEMARMDStream("test sem-ar", [sems, ars])
 
         ars.roi.value = (0.1, 0.1, 0.8, 0.8)
         self.ccd.binning.value = (4, 4) # hopefully always supported
 
         # Long acquisition
-        self.ccd.exposureTime.value = 0.2  # s
+        ars.detExposureTime.value = 0.2  # s
         ars.repetition.value = (2, 3)
         exp_shape = ars.repetition.value[::-1]
         num_ar = numpy.prod(ars.repetition.value)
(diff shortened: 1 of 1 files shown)


src/odemis/acq/test/stream_test.py
```diff
@@ -2973,6 +2941,7 @@ def tearDownClass(cls):
         if cls.backend_was_running:
             return
         test.stop_backend()
+        pass
```
Sabrina85: left over from debugging?

src/odemis/acq/test/stream_test.py
```diff
@@ -3172,6 +3109,7 @@ def test_acq_arpol_leech(self):
 
         ars.integrationTime.value = 1  # s
         ars.repetition.value = (2, 3)  # TODO use fixed repetition value -> set ROI?
+        assert ars.repetition.value == (2, 3)
```
Sabrina85: It happened that the repetition changed after calling self._roiToPhys. Why do you put the assert here?

src/odemis/acq/test/stream_test.py
```diff
@@ -3099,8 +3015,30 @@ def test_acq_arpol(self):
             # estimated acquisition time should be accurate with less than 50% margin + 1 extra second
             timeout = 1 + 1.5 * sas.estimateAcquisitionTime()
             start = time.time()
+            im0 = ars.image.value
             f = sas.acquire()
 
+            if pos == "acquireAllPol":
+                # Check if there is a live update in the setting stream.
+                # Only checked with multiple polarization, as otherwise, there is
+                # only one image anyway.
```
Sabrina85: This is only true if one polarization position and no integration time correct? If there is 1 pol pos and integration count >1 then there should be a lso a live update with different content?

pieleric: Yes, but as it's a 1x1 acquisition, it's not really possible to check that the live image "updates", as there is only one. So we only check where it's easy: when the polarization changes.

src/odemis/acq/test/stream_test.py
pieleric: Yes, but here we care that the repetition is the one we've set.

src/odemis/acq/test/stream_test.py
pieleric: fixed

PR #1293, opened 2019-12-06
util test: new assert_array_not_equal() helper function

pieleric:
Unfortunately, numpy doesn't provide such function, and it's not really
obvious/easy to implement. We have at least 3 open-coded way to do it
now, so let's consolidate it.


Sabrina85:
Nice! There are a couple of `assert not numpy.array_equal()` in the test cases. Shouldn't they also be replaced?
Can we also add the same for `assertListEqual`? There is no `assertListNotEqual`, but would be also useful to have, wouldn't it?

PR #1295, opened 2019-12-09
Store EXTRA_SETTINGS metadata for tiff and hdf5

philipwinklerdelmic:


src/odemis/dataio/tiff.py
```diff
@@ -1243,6 +1252,11 @@ def _addImageElement(root, das, ifd, rois, fname=None, fuuid=None):
             for j in range(3):
                 trane.attrib["A%d%d" % (i, j)] = "%.15f" % trans_mat[i][j]
 
+    if model.MD_EXTRA_SETTINGS in globalMD:
+        sett = globalMD[model.MD_EXTRA_SETTINGS]
+        sett = json.dumps(sett)  # serialize hw settings
+        extrase = ET.SubElement(ime, "ExtraSettings", attrib={"Settings": sett})
```
pieleric: Do not use an attribute inside the element, but the *content* of the element, aka `.text`.

src/odemis/dataio/tiff.py
```diff
@@ -350,6 +351,7 @@ def _convertToOMEMD(images, multiple_files=False, findex=None, fname=None, uuids
 #        + ARData ({0, 1})
 #        + POLData ({0, 1})
 #        + StreakCamData ({0, 1})
+#        + ExtraSettings   # To store all hw settings
```
pieleric: An alternative to inventing new extension to the OME standard would be to try to fit it to something of OME. I think the most fitting would be https://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2016-06/ome_xsd.html#CommentAnnotation .... which could be ok, if using a specific description or namespace. Still, it's probably not that exciting, because it's quite a bit more complex (as you need to use a AnnotationRef). So long story short, I agree with you to create a special attribute ;-)

philipwinklerdelmic:
updated

philipwinklerdelmic:
updated

src/odemis/dataio/test/hdf5_test.py
```diff
@@ -426,13 +429,14 @@ def testMetadata(self):
...
         self.assertEqual(len(rdata), 1)
-
         for im in rdata:
             self.assertEqual(im.metadata[model.MD_DESCRIPTION], metadata[model.MD_DESCRIPTION])
             self.assertEqual(im.metadata[model.MD_POS], metadata[model.MD_POS])
             self.assertEqual(im.metadata[model.MD_PIXEL_SIZE], metadata[model.MD_PIXEL_SIZE])
             self.assertEqual(im.metadata[model.MD_ACQ_DATE], metadata[model.MD_ACQ_DATE])
             self.assertEqual(im.metadata[model.MD_EXP_TIME], metadata[model.MD_EXP_TIME])
+            self.assertEqual(sorted(im.metadata[model.MD_EXTRA_SETTINGS].items()),
+                             sorted(metadata[model.MD_EXTRA_SETTINGS].items()))
```
pieleric: I don't understand why you need so use `sorted(...items())`. It's a dict, so there is no order anyway. Or is there a more hidden reason?

philipwinklerdelmic: I didn't realize `assertEqual` could also compare dictionaries (because they are not ordered), but actually it works just fine. I'll change it.

philipwinklerdelmic:
updated

src/odemis/dataio/test/hdf5_test.py
```diff
@@ -370,6 +370,8 @@ def testMetadata(self):
         """
         size = (512, 256, 1)
         dtype = numpy.dtype("uint16")
+        # list instead of tuple for binning because json only uses lists
+        extra_md = {"Camera" : {'binning' : [0, 0]}, u"¤³ß": {'</Image>': '</Image>'}}
```
pieleric: I think that the binning value should be stored as a *tuple*, and you would compute a "expected" metadata with the list instead of tuple. Either explicitly change it, or compute it by just `json.loads(json.dumps(extra_md))`.

Also, please add units, with some `None`, and some strings.

src/odemis/dataio/test/tiff_test.py
```diff
@@ -559,6 +565,8 @@ def testReadAndSaveMDSpec(self):
         for i in range(1, 171):
             cot.append((cot[0][0] + i, i * 1e-12))
 
+        # use list instead of tuple for binning because json converts tuples to lists anyway
+        extra_md = {"Camera" : {'binning' : [0, 0]}, u"¤³ß": {'</Image>': '</Image>'}}
```
pieleric: Same remark here on the content.

src/odemis/dataio/tiff.py
```diff
@@ -1243,6 +1251,11 @@ def _addImageElement(root, das, ifd, rois, fname=None, fuuid=None):
             for j in range(3):
                 trane.attrib["A%d%d" % (i, j)] = "%.15f" % trans_mat[i][j]
 
+    if model.MD_EXTRA_SETTINGS in globalMD:
+        sett = globalMD[model.MD_EXTRA_SETTINGS]
+        extrase = ET.SubElement(ime, "ExtraSettings")
+        extrase.text = json.dumps(sett)  # serialize hw settings
```
pieleric: For safety, I'd suggest to protect the `dumps()` in a try/except. So that if there is some weird object in the dictionary, it doesn't prevent from entirely saving the data.

philipwinklerdelmic:
updated

src/odemis/dataio/hdf5.py
```diff
@@ -1005,6 +1008,12 @@ def _add_image_metadata(group, image, mds):
         state = [ST_INVALID if v is None else ST_REPORTED for v in cots]
         _h5svi_set_state(gp["EmissionCurrentOverTime"], state)
 
+    sett = [md.get(model.MD_EXTRA_SETTINGS) for md in mds]
+    if any(sett):
+        value = [json.dumps(s) for s in sett]
```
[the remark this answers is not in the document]
philipwinklerdelmic: done!

PR #1296, opened 2019-12-10
gui: bugfix enabling entries

philipwinklerdelmic:


philipwinklerdelmic:
updated

pieleric:
Looks good. Especially, calling several time in a row `.pause()` and then `.resume()` should still work fine.

PR #1302, opened 2019-12-13
acq path: fix the check for pass-through filter

pieleric:
We tried to check for the presence of the pass-through filter, but a
typo prevented from ever finding it (choices -> choices.values()).
Anyway, the function needs to also account for the possible different
role of the ccd (eg, ccd0).

So instead of having a dedicated function, just detect the case *in* the
code that is in charge of the move to the right filter. This avoids
having to code twice the same thing in two different ways.

### DIFF
--- a/src/odemis/acq/path.py
+++ b/src/odemis/acq/path.py
@@ -489,9 +489,6 @@ class OpticalPathManager(object):
                 # path is correct anyway, just potentially more vibrations.
                 logging.exception("Failed to change CCD fan")
 
-        if mode == "mirror-align":
-            self._check_pass_through_available()
-
         fmoves = []  # moves in progress, list of (future, Component, dict(axis->pos) tuples
 
         # Restore the spectrometer focus before any other move, as (on the SR193),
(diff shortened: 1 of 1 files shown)


src/odemis/acq/path.py
```diff
@@ -552,7 +549,13 @@ def _doSetPath(self, path, detector):
                                     self._stored[comp_role, axis] = comp.position.value[axis]
                                 break
                         else:
-                            logging.debug("Choice %s is not present in %s axis", pos, axis)
+                            if mode == "mirror-align" and pos == "pass-through":
+                                # That's typically a sign that the microscope file is incorrect
+                                # eg, a typo in the filter name, as if there is a filter wheel
```
Sabrina85: I think the code is correct, but the comment is really confusing. Make it 2 sentences -> `If there is a filter wheel and a ccd .....`

src/odemis/acq/path.py
```diff
@@ -552,7 +549,13 @@ def _doSetPath(self, path, detector):
...
+                                # and a ccd, there should always be a pass-through position
+                                logging.warning("No 'pass-through' filter available, alignment might be harder due to limited signal")
+                            else:
+                                logging.debug("Choice %s is not present in %s axis", pos, axis)
```
Sabrina85: What will happen when this choice is not available. Stick with the old one? Mention this in the debug message?

Sabrina85: If the choice is not present, is that an error in the yaml file? Should we hint that in the message?

src/odemis/acq/path.py
Sabrina85: Line 542: I cannot comment on this line, but the comment is not really helping as it describes what the code does. Would be more helpful to say something like `Handle the filter wheel in a special way. Search for the fw position that corresponds to the requested position (band). `

PR #1303, opened 2019-12-16
gui comp canvas: Also update the display if the canvas is disabled

pieleric:
It was an optimisation because typically if the canvas is disabled,
the user doesn't need to see the images... however that's not true.
For instance, in the SPARC we are going to get live update during
acquisition. During the acquisition the canvas is disabled, but we
still want to see the image.
=> Don't optimise, and always show images when the streams provide
new data.

### DIFF
--- a/src/odemis/gui/comp/canvas.py
+++ b/src/odemis/gui/comp/canvas.py
@@ -876,9 +876,12 @@ class BitmapCanvas(BufferedCanvas):
         :param interpolate_data: (boolean) Apply interpolation if True
 
         """
-
-        # Don't draw anything if the canvas is disabled, leave the current buffer intact.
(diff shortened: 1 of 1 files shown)


src/odemis/gui/comp/canvas.py
```diff
@@ -1957,7 +1956,10 @@ def draw(self):
         #     logging.warn("No buffer created yet, ignoring draw request")
         #     return
 
-        if not self or not self.IsEnabled():
+        # Even if not IsEnabled() or not IsShownOnScreen() or or 0 in ClientSize,
```
Sabrina85: 2x `or`

pieleric:
The comments should be clearer now...

PR #1313, opened 2019-12-24
live update of acquisition

Kleijwegt:
provides live image
works for: CL, SEM, ARMDstream, streak camera
orange next pixels work
however has a blue NON-transparant background
Still 9 points to discuss with Eric [See TODO K.K. ERIC]

### DIFF
--- a/plugins/cliccd.py
+++ b/plugins/cliccd.py
@@ -29,6 +29,7 @@ from odemis import model
 from odemis.acq.stream import SEMCCDMDStream, ARSettingsStream
 from odemis.gui.conf.data import get_local_vas
 from odemis.gui.plugin import Plugin
+from odemis.model import MD_DESCRIPTION
 
 
 class SEMCLCCDStream(SEMCCDMDStream):
@@ -60,18 +61,21 @@ class SEMCLCCDStream(SEMCCDMDStream):
         # positions scanned).
         return data.mean()
 
-    def _onCompletedData(self, n, raw_das):
+    def _assembleLiveData(self, n, raw_data, px_idx, rep, pol_idx):
+        if n != self._ccd_idx:
+            return super(SEMCLCCDStream, self)._assembleLiveData(n, raw_data, px_idx, rep, pol_idx)
+
+        raw_data.metadata[MD_DESCRIPTION] = self._streams[n].name.value
+        self._live_data[n].append(raw_data)
+
+    def _assembleFinalData(self, n, data):
         # Only override for the CCD data
         if n < len(self._streams) - 1:
-            r = super(SEMCLCCDStream, self)._onCompletedData(n, raw_das)
-            return r
+            return super(SEMCLCCDStream, self)._assembleFinalData(n, data)
 
         # Same as sem data, but without computing the data position from the
         # CCD metadata
         md = self._ccd_md.copy()
-        sem_data = self._raw[0]  # _onCompletedData() should be called in order
-        md[model.MD_POS] = sem_data.metadata[model.MD_POS]
-        md[model.MD_DESCRIPTION] = self._streams[n].name.value
         # Make sure it doesn't contain metadata related to AR
         for k in (model.MD_AR_POLE, model.MD_AR_FOCUS_DISTANCE,
                   model.MD_AR_HOLE_DIAMETER, model.MD_AR_PARABOLA_F,
@@ -80,6 +84,7 @@ class SEMCLCCDStream(SEMCCDMDStream):
 
         try:
             # handle sub-pixels (aka fuzzing)
+            sem_data = self._raw[0]
             sem_shape = sem_data.shape[-1:-3:-1]  # 1,1,1,Y,X -> X, Y
             rep = self.repetition.value
             tile_shape = (sem_shape[0] / rep[0], sem_shape[1] / rep[1])
@@ -90,7 +95,7 @@ class SEMCLCCDStream(SEMCCDMDStream):
             logging.warning("Metadata missing from the SEM data")
 
         # concatenate data into one big array of (number of pixels,1)
-        flat_list = [ar.flatten() for ar in raw_das]
+        flat_list = [ar.flatten() for ar in data]
         rep_one = numpy.concatenate(flat_list)
         # reshape to (Y, X)
         rep_one.shape = rep[::-1]
--- a/src/odemis/acq/stream/_base.py
+++ b/src/odemis/acq/stream/_base.py
@@ -268,12 +268,12 @@ class Stream(object):
 
         self.intensityRange.subscribe(self._onIntensityRange)
 
-    def _init_thread(self):
+    def _init_thread(self, period=0.1):
         """ Initialize the thread that updates the image
         """
         self._imthread = threading.Thread(target=self._image_thread,
-                                          args=(weakref.ref(self),),
-                                          name="Image computation")
+                                          args=(weakref.ref(self), period),
+                                          name="Image computation of %s" % self.name.value)
         self._imthread.daemon = True
         self._imthread.start()
 
@@ -848,15 +848,14 @@ class Stream(object):
         self._im_needs_recompute.set()
 
     @staticmethod
-    def _image_thread(wstream):
+    def _image_thread(wstream, period=0.1):
         """ Called as a separate thread, and recomputes the image whenever it receives an event
         asking for it.
 
         Args:
             wstream (Weakref to a Stream): the stream to follow
-
+            period ( float > 0) (Seconds): Minimum time in second between two image updates
         """
-
         try:
             stream = wstream()
             name = stream.name.value
@@ -864,8 +863,6 @@ class Stream(object):
             # Only hold a weakref to allow the stream to be garbage collected
             # On GC, trigger im_needs_recompute so that the thread can end too
             wstream = weakref.ref(stream, lambda o: im_needs_recompute.set())
-
-            tnext = 0
             while True:
                 del stream
                 im_needs_recompute.wait()  # wait until a new image is available
@@ -875,16 +872,15 @@ class Stream(object):
                     logging.debug("Stream %s disappeared so ending image update thread", name)
                     break
 
-                tnow = time.time()
+                tnext = time.time() + period  #running with a period of max "period", optional arg standard is 10 Hz
+                im_needs_recompute.clear()
+                stream._updateImage()
 
+                tnow = time.time()
                 # sleep a bit to avoid refreshing too fast
                 tsleep = tnext - tnow
                 if tsleep > 0.0001:
                     time.sleep(tsleep)
-
-                tnext = time.time() + 0.1  # max 10 Hz
-                im_needs_recompute.clear()
-                stream._updateImage()
         except Exception:
             logging.exception("Image update thread failed")
 
@@ -917,7 +913,6 @@ class Stream(object):
     def _updateImage(self):
         """ Recomputes the image with all the raw data available
         """
-        # logging.debug("Updating image")
         if not self.raw:
             return
 
--- a/src/odemis/acq/stream/_sync.py
+++ b/src/odemis/acq/stream/_sync.py
@@ -42,14 +42,16 @@ from odemis.acq.stream._live import LiveStream
 from odemis.model import MD_POS, MD_DESCRIPTION, MD_PIXEL_SIZE, MD_ACQ_DATE, MD_AD_LIST, \
     MD_DWELL_TIME, MD_EXP_TIME, MD_DIMS
 from odemis.model import hasVA
-from odemis.util import units, executeAsyncTask, almost_equal, get_best_dtype_for_acc
+from odemis.util import units, executeAsyncTask, almost_equal, get_best_dtype_for_acc, img
 import queue
 import threading
 import time
 
 import odemis.util.driver as udriver
+from . import MonochromatorSettingsStream
 
 from ._base import Stream, POL_POSITIONS, POL_MOVE_TIME
+import weakref
 
 # On the SPARC, it's possible that both the AR and Spectrum are acquired in the
 # same acquisition, but it doesn't make much sense to acquire them
@@ -70,7 +72,8 @@ from ._base import Stream, POL_POSITIONS, POL_MOVE_TIME
 # with a scanner/emitter.
 EBEAM_DETECTORS = ("se-detector", "bs-detector", "cl-detector", "monochromator",
                    "ebic-detector")
-
+GUI_BLUE = (47, 167, 212) # FG_COLOUR_EDIT - from src/odemis/gui/__init__.py
+GUI_ORANGE = (255, 163, 0) # FG_COLOUR_HIGHLIGHT - from src/odemis/gui/__init__.py
 
 class MultipleDetectorStream(with_metaclass(ABCMeta, Stream)):
     """
@@ -160,8 +163,13 @@ class MultipleDetectorStream(with_metaclass(ABCMeta, Stream)):
                 self._integrationTime = s.integrationTime
                 self._integrationCounts = s.integrationCounts
 
-        # acquisition end event
+        # Acquisition end event
         self._acq_done = threading.Event()
+
+        # Start threading event for live update overlay
+        self._live_update_period = 0.33
+        self._im_needs_recompute = threading.Event()
+        self._init_thread(self._live_update_period)
 
         # For the acquisition
         self._acq_lock = threading.Lock()
@@ -172,9 +180,9 @@ class MultipleDetectorStream(with_metaclass(ABCMeta, Stream)):
         self._acq_rep_n = 0  # number of acquisitions so far
         self._prog_sum = 0  # s, for progress time estimation
 
-        # all the data received, in order, for each stream
-        self._acq_data = [[] for _ in streams]
-
+        # the data received, in order, for each stream
+        self._acq_data = [[] for _ in streams] # latest acquired data
+        self._live_data = [[] for _ in streams] # all acquired data in live format, reshaped to the final shape by _assembleFinalData
         self._acq_min_date = None  # minimum acquisition time for the data to be acceptable
 
         # Special subscriber function for each stream dataflow
@@ -189,8 +197,8 @@ class MultipleDetectorStream(with_metaclass(ABCMeta, Stream)):
         self.should_update = model.BooleanVA(False)
         self.is_active = model.BooleanVA(False)
 
-#     def __del__(self):
-#         logging.debug("MDStream %s unreferenced", self.name.value)
+    #     def __del__(self):
+    #         logging.debug("MDStream %s unreferenced", self.name.value)
 
     @property
     def streams(self):
@@ -559,6 +567,7 @@ class MultipleDetectorStream(with_metaclass(ABCMeta, Stream)):
 
     def _onCompletedData(self, n, raw_das):
         """
+        OLD-METHOD WHICH IS STILL USED FOR SEMTemporalMDStream, ScannedFluoMDStream
         Called at the end of an entire acquisition. It should assemble the data
         and append it to ._raw .
         Override if you need to process the data in a different way.
@@ -679,10 +688,11 @@ class MultipleDetectorStream(with_metaclass(ABCMeta, Stream)):
             if numpy.prod(rep) == arr.shape[0]:
                 arr.shape = rep[::-1]
             else:
-                im_px = int(arr.shape[0]/numpy.prod(rep))  # number of images per pixel
+                im_px = int(arr.shape[0] / numpy.prod(rep))  # number of images per pixel
                 arr.shape = rep[::-1] + (im_px,)
                 # average images
                 arr = numpy.mean(arr, 2).astype(data_list[0].dtype)
+
         else:
             # need to reorder data by tiles
             # change N to Y, X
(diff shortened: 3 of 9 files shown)


src/odemis/acq/stream/_sync.py
```diff
@@ -774,6 +779,149 @@ def _stopLeeches(self):
...
+        rgbim[numpy.invert(self._acq_mask), 1] = 121
+        rgbim[numpy.invert(self._acq_mask), 2] = 192
+
+        if hasattr(self, "_npixels2scan"):
+            size_preview = self._npixels2scan
+        else:
+            size_preview = 1
+
+        # Orange progress pixels
+        idx = numpy.where(self._acq_mask == False)
```
pieleric: I don't think this is very reliable. In particular, I think in case of "fuzzy", this is not going to work. An alternative which might work could to have a _scanning_area which is basically a left/top/rigth/bottom (either as 4 int, or a numpy slice), which would directly fit on the SEM data. It would be computed/updated by the runAcquisition function.

src/odemis/acq/stream/_sync.py
```diff
@@ -774,6 +779,149 @@ def _stopLeeches(self):
...
+        rgbim[idx[0][0:size_preview],idx[1][0:size_preview],0] = 204
+        rgbim[idx[0][0:size_preview], idx[1][0:size_preview], 1] = 130
+        rgbim[idx[0][0:size_preview], idx[1][0:size_preview], 2] = 0
```
pieleric: You can write directly the last dimension, so:
`rgbim[idx[0][0:size_preview],idx[1][0:size_preview]] = (204, 130, 0)`
That magical number should be moved to a constant like `SCANNING_AREA_COLOUR`.

src/odemis/acq/stream/_sync.py
```diff
@@ -1062,18 +1206,32 @@ def _runAcquisitionEbeam(self, future):
...
+                                                     args = (weakref.ref(self), self._live_update_period),
+                                                     name = "Live update of the acquisition")
+            self._liveImageThread.daemon = True
+            self._liveImageThread.start()
+
+
+            for pol_idx, pol_pos in enumerate(pos_polarizations):
                 if pol_pos is not None:
-                    logging.debug("Acquiring with the polarization position %s", pol_pos)
+                    logging.debug("Acquire with the following polarization position: %s" % pol_pos)
```
pieleric: Please "revert", as for the other logging lines.

src/odemis/acq/stream/_sync.py
```diff
@@ -1062,18 +1206,32 @@ def _runAcquisitionEbeam(self, future):
...
                     f = self._analyzer.moveAbs({"pol": pol_pos})
                     f.result()
                     time_move_pol_left -= time_move_pol_once
 
-                # iterate over pixel positions for scanning
+                # iterate over pixel positions for scanning.
                 for px_idx in numpy.ndindex(*rep[::-1]):  # last dim (X) iterates first
                     trans = tuple(spot_pos[px_idx])  # spot position
 
+                    self._current_scan_area = (px_idx[1] * tile_size[0],
```
pieleric: Let's define it in the `__init__`, for documentation, as a tuple `(0,0,0,0)`.

src/odemis/acq/stream/_sync.py
```diff
@@ -1118,7 +1275,19 @@ def _runAcquisitionEbeam(self, future):
...
+                    for i, das in enumerate(self._acq_data):
+                        self._assembleLiveData(i, das[-1], px_idx, rep, pol_idx)
+
+                    # Activate _updateImage thread
+                    self._shouldUpdateImage()
+
+                    # Live update the setting stream with the new data
+                    self._sccd._onNewData(self._ccd_df, self._acq_data[self._ccd_idx][-1])
+
+                    logging.debug("Done acquiring image number %s out of %s." % (n, tot_num))
```
pieleric: Typically, in logging, it's considered nicer to read to have each variable passed as a separate parameter (as it used to be for this line). So no need to change to using `%`.

src/odemis/acq/stream/_sync.py
```diff
@@ -1130,8 +1299,11 @@ def _runAcquisitionEbeam(self, future):
                     raise CancelledError()
                 self._acq_state = FINISHED
 
-            for n, das in enumerate(self._acq_data):
-                self._onCompletedData(n, das)
+            self._im_needs_recompute.clear()
+            # Broadcast all the self._LiveData into self._raw and do post-processing
```
pieleric: I don't think "broadcast" might is the right term. Maybe rephrase as "Adjust the (intermediary) ._live_data to the right shape/format for the final ._raw".

src/odemis/acq/stream/_sync.py
```diff
@@ -1130,8 +1299,11 @@ def _runAcquisitionEbeam(self, future):
...
+            self._raw = [] # reset the ._raw variable
```
pieleric: That shouldn't be needed, as it's done at the beginning of the function.

src/odemis/acq/stream/_sync.py
```diff
@@ -1741,30 +1913,48 @@ def _runAcquisition(self, future):
...
             leech_np, leech_time_ppx = self._startLeeches(px_time, tot_num, (rep[1], rep[0]))
 
+            self._im_needs_recompute = threading.Event()
+            self._live_update_period = 0.33 # s
+            self._liveImageThread = threading.Thread(target = self._image_thread,
+                                                     args = (weakref.ref(self), self._live_update_period),
+                                                     name = "Live update of the acquisition")
+            self._liveImageThread.daemon = True
+            self._liveImageThread.start()
+
```
pieleric: Can be moved to the `__init__`.

src/odemis/acq/stream/_sync.py
```diff
@@ -1741,30 +1913,48 @@ def _runAcquisition(self, future):
...
             # number of spots scanned so far
             spots_sum = 0
             while spots_sum < tot_num:
+                px_idx = (spots_sum//rep[0], spots_sum%rep[0]) #current pixel index
```
pieleric: Move this line to just before `_current_scan_area`.
BTW, please add spaces around the operators.

src/odemis/acq/stream/_sync.py
```diff
@@ -1741,30 +1913,48 @@ def _runAcquisition(self, future):
...
                 # Acquire the maximum amount of pixels until next leech
-                npixels = min([np for np in leech_np if np is not None] +
-                              [tot_num - spots_sum])  # max, in case of no leech
-                n_y, n_x = leech.get_next_rectangle((rep[1], rep[0]), spots_sum, npixels)
-                npixels = n_x * n_y
+                self._npixels2scan = min([np for np in leech_np if np is not None] +
```
pieleric: No need (anymore) to store it on self. => You can revert to the original name `npixels`.

src/odemis/acq/stream/_sync.py
```diff
@@ -1741,30 +1913,48 @@ def _runAcquisition(self, future):
...
+                                         [tot_num - spots_sum] + [int(self._live_update_period // px_time + 1)]) # max, in case of no leech
+                #TODO K.K. update unit tests to test for:
```
pieleric: Please do this TODO, or drop this line if it's done now.

src/odemis/acq/stream/_sync.py
```diff
@@ -1741,30 +1913,48 @@ def _runAcquisition(self, future):
...
+                # cornercase where leech is set far after update_interval and therefore a non rectangular block may be scanned
+                # case where leech and update interval run semi in phase/out of phase
+                n_y, n_x = leech.get_next_rectangle((rep[1], rep[0]), spots_sum, self._npixels2scan)
+                self._npixels2scan = n_x * n_y
```
pieleric: Indicate *why* you recompute it afterwards (ie, it might be smaller, to stay a rectangle)

src/odemis/acq/stream/_sync.py
```diff
@@ -1741,30 +1913,48 @@ def _runAcquisition(self, future):
...
+
+                self._current_scan_area = (px_idx[1],
+                                           px_idx[0],
+                                           px_idx[1] + self._npixels2scan,
+                                           px_idx[0])
```
pieleric: That looks incorrect. -> `px_idx[1] + ny, px_idx[0]+nx`?

src/odemis/acq/stream/_sync.py
```diff
@@ -1806,6 +1994,17 @@ def _runAcquisition(self, future):
                         raise CancelledError()
                     s._dataflow.unsubscribe(self._subscribers[i])
 
+                px_x, px_y = divmod(spots_sum, rep[0])
+                px_idx = (px_x, px_y)
```
pieleric: That should be merged with px_idx computed at the beginning. (You could just drop these two lines, or use divmod up).

src/odemis/acq/stream/_sync.py
```diff
@@ -1833,17 +2032,19 @@ def _runAcquisition(self, future):
...
                 if self._acq_state == CANCELLED:
                     raise CancelledError()
                 self._acq_state = FINISHED
 
-            # TODO: ideally, we would directly assemble the _acq_data into the final
-            # raw, which would avoid temporarily holding twice all the data.
-
-            for n, das in enumerate(self._acq_data):
-                self._onCompletedData(n, das)
+            self._im_needs_recompute.clear()
```
pieleric: Drop this line, I don't think it's needed.

src/odemis/acq/stream/_sync.py
```diff
@@ -1806,6 +1994,17 @@ def _runAcquisition(self, future):
                         raise CancelledError()
                     s._dataflow.unsubscribe(self._subscribers[i])
 
+                px_x, px_y = divmod(spots_sum, rep[0])
+                px_idx = (px_x, px_y)
+
+                for i, das in enumerate(self._acq_data):
+                    self._assembleLiveData2D(i, das[-1], px_idx, rep, 0)
+
+                # Activate _updateImage thread
```
pieleric: That's pretty much the exact description of what `shouldUpdateImage()` does. Indicate *why* now and what are the concequences. Maybe something like "Update the live image of each stream".
That said, how does it work for the other streams than the first one? I don't see where is the code...

src/odemis/acq/stream/_sync.py
```diff
@@ -2178,6 +2411,33 @@ class SEMTemporalSpectrumMDStream(SEMCCDMDStream):
...
+            md[MD_POS] = self._live_data[0][pol_idx].metadata[MD_POS]
+            md[MD_DESCRIPTION] = self._streams[n].name.value
+            md[MD_DIMS] = "CTZYX"
+
+            self._live_data[n] = model.DataArray(
+                                 numpy.zeros(shape = (spec_res, temp_res, 1, rep[1], rep[0]), dtype = raw_data.dtype), md)
+
+        self._live_data[n][:, :, 0, px_idx[0], px_idx[1]] = raw_data.reshape(spec_res, temp_res)
+
     def _onCompletedData(self, n, raw_das):
```
pieleric: All the _onCompletedData() methods and their helper functions need to go away.

src/odemis/acq/stream/_sync.py
```diff
@@ -773,6 +778,143 @@ def _stopLeeches(self):
...
+                       px_idx[1]: px_idx[1] + tile_shape[1]] = True
+        self._live_data[n][pol_idx][
+                           px_idx[0]: px_idx[0] + tile_shape[0],
+                           px_idx[1]: px_idx[1] + tile_shape[1]] = raw_data
+
+    def _assembleFinalData(self, data, pol_idx, pos_polarizations, stream_idx):
+        '''
+        :param data: all acquired data of the streams
+        :param pol_idx: (int) polarisation index related to name as defined in pos_polarizations variable
+        :param pos_polarizations: (list of str) number of possible polarization possibilities
```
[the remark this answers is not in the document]
Kleijwegt: pos_polarizations we need to not average if there are not multiple polarisations.
pol_idx is very nice if stuff goes wrong for debugging. Since it does not make the function unreadable or messy I propose to leave it in place.

src/odemis/acq/stream/_sync.py
```diff
@@ -1037,17 +1181,17 @@ def _runAcquisitionEbeam(self, future):
...
-                logging.debug("Will acquire the following polarization positions: %s",
-                              pos_polarizations)
+                    logging.debug("Will acquire the following polarization position: %s" % pos_polarizations)
                 # extra time to move pol analyzer for each pos requested (value is very approximate)
                 time_move_pol_once = POL_MOVE_TIME  # s
+                logging.debug("Add %s extra sec to move polarization analyzer for all positions requested."
+                              % time_move_pol_left)
                 time_move_pol_left = time_move_pol_once * len(pos_polarizations)
-                logging.debug("Add %s extra sec to move polarization analyzer for all positions requested.",
-                              time_move_pol_left)
```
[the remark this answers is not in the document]
Kleijwegt: do you want to lose these logging messages of just change them back as they where?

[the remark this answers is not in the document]
pieleric: Change them back (to the version which is currently in master/HEAD).

src/odemis/acq/stream/_sync.py
```diff
@@ -773,6 +778,143 @@ def _stopLeeches(self):
...
+                       px_idx[1] * tile_shape[1]:(px_idx[1] + 1) * tile_shape[1]] = True
+        self._live_data[n][pol_idx][
+                       px_idx[0] * tile_shape[0]:(px_idx[0] + 1) * tile_shape[0],
+                       px_idx[1] * tile_shape[1]:(px_idx[1] + 1) * tile_shape[1]] = raw_data
+
+    def _assembleLiveData2D(self, n, raw_data, px_idx, rep, pol_idx):
+        """
+        :param n: (int) number of the current stream
+        :param raw_data: acquired data of stream
+        :param px_idx: (tuple of int) pixel index: y, x
```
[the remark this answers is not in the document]
Kleijwegt: add to docstring that pxidx is first pixel of last aquisition of CL since CL works different

src/odemis/acq/stream/_sync.py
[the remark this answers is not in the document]
Kleijwegt: ok remove

src/odemis/acq/stream/_sync.py
```diff
@@ -1062,18 +1206,32 @@ def _runAcquisitionEbeam(self, future):
...
             n = 0  # number of images acquired so far
 
-            for pol_pos in pos_polarizations:
+            self._im_needs_recompute = threading.Event()
+            self._live_update_period = 0.33  # s
+            self._liveImageThread = threading.Thread(target = self._image_thread,
+                                                     args = (weakref.ref(self), self._live_update_period),
+                                                     name = "Live update of the acquisition")
+            self._liveImageThread.daemon = True
+            self._liveImageThread.start()
```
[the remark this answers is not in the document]
Kleijwegt: fix as in todo

src/odemis/acq/stream/_sync.py
Kleijwegt: remove K.K.

src/odemis/acq/stream/_sync.py
```diff
@@ -834,62 +859,48 @@ def _assembleLiveData2D(self, n, raw_data, px_idx, rep, pol_idx):
...
+        Create RGB projection of over of live SEM data, also adds a blue background of non-scanned pixels and orange
+        pixels for the pixels which are currently being scanned.
         """
         # Here we need to pass a *copy* of mask, because it might change *while* numpy is doing
         # the indexing. It seems that numpy first allocates the size of data_acq before
         # actually copying the values, so if there are eventually more values, it can overflow
         # the array index.
-        data_acq = data[self._acq_mask.copy()]
-        logging.debug("Got %s data for histogram", data_acq.shape)
+        acq_mask = self._acq_mask.copy() #because of threading issues this variable needs to be copied
```
pieleric: Please drop this comment, and just update the explanation just above.

src/odemis/acq/stream/_sync.py
```diff
@@ -899,21 +910,35 @@ def _projectXY2RGB(self, data, tint=(255, 255, 255)):
...
-        self.streams[0].raw = [self._live_data[0][-1]]  # For GetBoundingBox()
-        no_acq_mask = numpy.invert(self._acq_mask)
-        scan_area = self._current_scan_area
-
-        rgbim = self._projectXY2RGB(self.streams[0].raw[0])
-        rgbim = self._progressPixels(rgbim, scan_area, no_acq_mask)
+        try:
+            raw_data = self._live_data[0][-1]
+        except IndexError:  # Can happen if the acquisition has just finished
+            if self._acq_done.isSet():
```
pieleric: Please use the "new" name which is `is_set()`.

src/odemis/acq/stream/_sync.py
```diff
@@ -899,21 +910,35 @@ def _projectXY2RGB(self, data, tint=(255, 255, 255)):
...
+                logging.debug("Not updating live image, as acquisition is over")
+                return
+            else:
+                raise
 
-        rgbim.flags.writeable = False
+        self.streams[0].raw = [raw_data]  # For GetBoundingBox()
+        rgbim = self._projectXY2RGB(raw_data)
+        # Don't update if the acquisition is already over
+        if self._acq_done.isSet():
```
pieleric: -> `is_set()`

src/odemis/acq/stream/_sync.py
```diff
@@ -1636,6 +1659,12 @@ def _runAcquisitionScanStage(self, future):
 
             n = 0  # number of points acquired so far
             for px_idx in numpy.ndindex(*rep[::-1]):  # last dim (X) iterates first
+                #TODO K.K. what is both CL and SEM-scanstage are used simoultounously what is displayed? what is called?
```
pieleric: I'm not exactly sure I understand the question, but essentially, the same place is scanned as in the equivalent code in `runAcquisitionEbeam()`. So that code looks fine.

src/odemis/acq/stream/_sync.py
```diff
@@ -1790,8 +1826,12 @@ def _runAcquisitionScanStage(self, future):
...
                 self._acq_state = FINISHED
 
-            for n, das in enumerate(self._acq_data):
-                self._onCompletedData(n, das)
+            # Process all the (intermediary) ._live_data to the right shape/format for the final ._raw
+            for stream_idx, das in enumerate(self._live_data):
+                self._assembleFinalData(stream_idx, das)
+
+            # for n, das in enumerate(self._acq_data):
+            #     self._onCompletedData(n, das)
```
pieleric: Drop these lines.

src/odemis/acq/stream/_sync.py
```diff
@@ -1923,38 +1967,29 @@ def _runAcquisition(self, future):
...
+                npixels2scan = min([np for np in leech_np if np is not None] +
                                          [tot_num - spots_sum] + [int(self._live_update_period // px_time + 1)]) # max, in case of no leech
-                #TODO K.K. update unit tests to test for:
+                #TODO update unit tests to test for:
                 # cornercase where leech is set far after update_interval and therefore a non rectangular block may be scanned
                 # case where leech and update interval run semi in phase/out of phase
-                n_y, n_x = leech.get_next_rectangle((rep[1], rep[0]), spots_sum, self._npixels2scan)
-                self._npixels2scan = n_x * n_y
+                n_y, n_x = leech.get_next_rectangle((rep[1], rep[0]), spots_sum, npixels2scan) #method to only scan
+                # rectangular blocks of pixels which start at the beginning of a row
```
pieleric: Either put the long comment in a separate line *before* the function call, or make the comment fit on the line. Maybe you can just write `# also ensure it's a rectangle`

src/odemis/acq/stream/_sync.py
```diff
@@ -2072,10 +2102,35 @@ def _runAcquisition(self, future):
...
             self._dc_estimator = None
             self._current_future = None
             self._acq_done.set()
 
+    def _updateImage(self):
+        """
+        Function called by image update thread which handles updating the overlay of the SEM live update image
+        """
+        # Display only the SEM image taken in the last polarization and convert to RGB + display
+        for stream in range(2):
```
pieleric: I'd suggest to write it like `for i, stream in enumerate(self._stream)`.
The, you can replace the `stream` -> `i`, and `self.streams[stream]` -> `stream`.

src/odemis/gui/cont/acquisition.py
```diff
@@ -815,9 +817,11 @@ def on_acquisition_done(self, future):
             logging.error("Acquisition failed (after %d streams): %s",
                           len(data), exp)
 
-        # Remove live update
+        #  REMOVE the the overlay (live_update) in the SEM window which displays the SEM measurements of the current
+        #  acquisition if a  stream is added and acquisition is started.
+        # TODO K.K. fix that also the CL stream is removed
```
pieleric: No, I think the CL stream should stay, as it's a setting stream. That said, you might want to make sure the last data displayed corresponds to the complete acquisition (without any orange/blue indicators).... but that not a big deal if it's not done and just left as a TODO.

src/odemis/acq/stitching/_registrar.py
```diff
@@ -206,8 +206,11 @@ def getPositions(self):
         tile_positions = []
         dep_tile_positions = []
 
+        # TODO K.K. remove this or implement differently
+        self.shift_wilco = [] #Made for Wilco, currenlty a debug value
         for ti in self.acqOrder:
             shift = self.registered_positions[ti[0]][ti[1]]
+            self.shift_wilco.append(shift)
```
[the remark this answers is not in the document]
Kleijwegt: shit these are changes for Sonic - mixed up commits

src/odemis/acq/stream/_sync.py
```diff
@@ -789,6 +802,10 @@ def _assembleLiveDataTiles(self, n, raw_data, px_idx, rep, pol_idx):
         :param rep: (tuple of int) repetition frame/ size of entire frame
         :param pol_idx: (int) polarisation index related to name as defined in pos_polarizations variable
         uses fuzzing
+        One full SEM scan is stored for each polarisation, at the end assembleFinalData() will take care of
+        integrating (aka averaging) them into a single one. To do this "live" would be diffucult and prone to error's since
+        an weighted average over the total number of polarisations in combination with the previous done
+        polarisations need te be made.
```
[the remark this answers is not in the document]
Kleijwegt: It feels a bit odd to start the function with a comment since this is a general comment. Where do you suggest to put this comment?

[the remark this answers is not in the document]
pieleric: I think that it's fine to be at the start of the function. BTW, I'd suggest adjust the punctuation to read more easily:
" One full SEM scan is stored for each polarisation. At the end, assembleFinalData() will take care of..."

src/odemis/acq/stream/_base.py
```diff
@@ -848,43 +848,40 @@ def _shouldUpdateImage(self):
...
             name = stream.name.value
             im_needs_recompute = stream._im_needs_recompute
             # Only hold a weakref to allow the stream to be garbage collected
             # On GC, trigger im_needs_recompute so that the thread can end too
             wstream = weakref.ref(stream, lambda o: im_needs_recompute.set())
-
-            tnext = 0
             while True:
                 del stream
+                #TODO K.K. Eric does this go alright with waiting/pointers etc?
```
pieleric: That looks fine. Especially, assuming it was fine, the changes shouldn't bring more issues.

src/odemis/acq/stream/_sync.py
```diff
@@ -1183,7 +1375,7 @@ def _integrateImages(self, n):
             orig_dtype = data2process[0].dtype
 
             if det_type == model.MD_DT_NORMAL:  # SEM
-                logging.debug("Averaging %s images", n)
+                # logging.debug("Average %s images", n)
```
pieleric: Please leave these log as they were (l 1378 and 1387).

src/odemis/acq/stream/_sync.py
```diff
@@ -1600,6 +1802,13 @@ def _runAcquisitionScanStage(self, future):
                             leech_np[li] = np
                             if self._acq_state == CANCELLED:
                                 raise CancelledError()
+                    #TODO K.K. just added code here for scan stage, check if placed right
```
pieleric: That looks alright. => remove this comment.

Kleijwegt:
> Just added a few extra comments, mostly concerning the "TODO k.k." that you had left. Please have a look at them, and the few other comments left, and let's try to get this merged soon.

hjfgjyjy

src/odemis/acq/stream/_sync.py
Kleijwegt: oeps that was actually an comment for my-future-selfto review my-past-self

Kleijwegt:
@pieleric I have implemented a monochrometer fix and a fix for the ugly orange line/dot in the CL update on the last run (this was easy).
Do you like this implementation?
Do you know how to propely wait on the thread (which updates the live view) to finish so instead of using time.sleep()?

src/odemis/acq/stream/_sync.py
```diff
@@ -1833,17 +2068,31 @@ def _runAcquisition(self, future):
...
+                if isinstance(self._streams[stream_idx], MonochromatorSettingsStream):
+                    # The Monochromator stream uses a chronogram view, which is not compatible with the RGB spatial image
+                    continue
+                elif stream_idx >= 1:
+                    # Only update the CL stream
+                    self._streams[1].raw = [self._live_data[1][-1]]  # For GetBoundingBox()
+                    rgbim = self._projectXY2RGB(self._live_data[1][-1])
+                    self._streams[1].image.value = rgbim
+                    #TODO K.K. what is better? Above or below?
+                    # self._streams[stream_idx]._onNewData(self._streams[stream_idx]._dataflow, self._raw[-1])
```
Kleijwegt: @pieleric I am not sure if my implementation is actually that much better, which one do you prefer?

Kleijwegt: after this I will remove all the _Oncompletedata, remove all the unnecacerry changes etc. and then this should be ready to be merged.

pieleric: I think that if you end-up still needing this "isinstance()" (because you cannot call `updateImage()` directly, then just go for the shorter version, with `onNewData()`.
BTW, I'd suggest to put this block concerning the image update into a try/except block, and only log the exceptions, because we don't want any issue here to cause a complete failure of the acquisition.

pieleric:
Once you're done adjusting this small bit, I think the code should be ready. You'll still need to rebase with upstream (to fix the conflict), remove the onCompleteData() functions, put back the testing code, and remove the extra changes related to the stitching.
Finally, please squash you changes, and put a nice commit text explanation.

plugins/cliccd.py
```diff
@@ -60,16 +61,22 @@ def _preprocessData(self, n, data, i):
...
-            return r
+            return super(SEMCLCCDStream, self)._assembleFinalData(n, data)
 
         # Same as sem data, but without computing the data position from the
         # CCD metadata
         md = self._ccd_md.copy()
-        sem_data = self._raw[0]  # _onCompletedData() should be called in order
+        sem_data = self._raw[0]  # _assembleFinalData() should be called in order
         md[model.MD_POS] = sem_data.metadata[model.MD_POS]
         md[model.MD_DESCRIPTION] = self._streams[n].name.value
```
pieleric: You can drop this line, as you've put it in `_assembleLiveData()`.
I think that could also move the MD_POS and removal of the MD_AR* to `_assembleLiveData()`.

plugins/cliccd.py
```diff
@@ -90,7 +97,7 @@ def _onCompletedData(self, n, raw_das):
             logging.warning("Metadata missing from the SEM data")
```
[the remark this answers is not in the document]
pieleric: It looks alright. It essentially does the same as `get_center_pxs()`.
BTW, I guess in theory, we should move more of this code to the "assembleLiveData()", as every new pixel (ie, CCD image) could immediately be stored in the final array. Anyway, this is good enough...

Kleijwegt:
@pieleric Ready to be merged

Kleijwegt:
@pieleric ready to be merged

PR #1314, opened 2020-01-08
driver simstreakcam/hamamatsurx: support for transp

pieleric:


pieleric:
No, the transp argument reverses the image, because the camera is physically in the "other" direction. That's somehow precisely to match the MD_WL_LIST, so this metadata should stay as-is.

Sabrina85:
ok and the same for the TIME_LIST?

PR #1318, opened 2020-01-10
gui cont tabs SPARC2AlignTab: when leaving the tab, always ensure the calibration light is off

pieleric:
The calibration light is turned during focusing.
It would normally be turned off automatically when cancelling the
autofocus. However, on the streak cam, if the manual focus was active,
the light would not be turned off.

=> When leaving the tab, reset the manual focus (so that when coming back to the tab
it doesn't look like we are in this mode), and turn off the light.

### DIFF
--- a/src/odemis/gui/cont/tabs.py
+++ b/src/odemis/gui/cont/tabs.py
@@ -4235,6 +4235,16 @@ class Sparc2AlignTab(Tab):
             # Cancel autofocus (if it happens to run)
             self.tab_data_model.autofocus_active.value = False
 
+            # Cancel manual focus: just reset the button, as the rest is just
+            # optical-path moves.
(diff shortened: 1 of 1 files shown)


src/odemis/gui/cont/tabs.py
```diff
@@ -4235,6 +4235,16 @@ def Show(self, show=True):
             # Cancel autofocus (if it happens to run)
             self.tab_data_model.autofocus_active.value = False
 
+            # Cancel manual focus: just reset the button, as the rest is just
```
Sabrina85: What about the slit of the spectrograph? For manual focus we also automatically close it. Or is this always taken care of by the optical path settings?

pieleric: Yes, the optical path manager takes care of the slit.

PR #1323, opened 2020-01-15
Os error msg pop up

Kleijwegt:
shows an pop up message if an error occurred or if an recovery from an error is made.

TODO is change all the error messages for subsystems to a (maybe) more meaningful message, which includes the next actions to be taken. So for example notify user that a recovery is made.

### DIFF
--- a/src/odemis/driver/simcam.py
+++ b/src/odemis/driver/simcam.py
@@ -21,18 +21,20 @@ Odemis. If not, see http://www.gnu.org/licenses/.
 '''
 from __future__ import division
 
+import threading
 from past.builtins import long
 from builtins import str
 import queue
 import logging
 import numpy
 from odemis import model, util, dataio
-from odemis.model import oneway
+from odemis.model import BASE_DIRECTORY, oneway
 import os
 from scipy import ndimage
 import time
 from PIL import Image, ImageDraw, ImageFont
 
+ERROR_STATE_FILE = "simcam-hw.error"
 
 class Camera(model.DigitalCamera):
     '''
@@ -142,6 +144,14 @@ class Camera(model.DigitalCamera):
         # Convenience event for the user to connect and fire
         self.softwareTrigger = model.Event()
 
+        # Include a thread which creates or fixes an hardware error in the simcam on the basis of the presence of the
+        # file ERROR_STATE_FILE in model.BASE_DIRECTORY
+        self._is_running = True
+        self._error_creation_thread = threading.Thread(target=self._state_error_run,
+                                                       name="Creating and state error")
+        self._error_creation_thread.daemon = True
+        self._error_creation_thread.start()
+
     def _setBinning(self, value):
         """
         value (2-tuple int)
@@ -196,6 +206,7 @@ class Camera(model.DigitalCamera):
         Must be called at the end of the usage. Can be called multiple times,
         but the component shouldn't be used afterwards.
         """
+        self._is_running = False #Stop error creation thread
         self._stop_generate()
 
     def _start_generate(self):
@@ -316,6 +327,28 @@ class Camera(model.DigitalCamera):
 
         return sim_img
 
+    def _state_error_run(self):
+        '''
+        Creates or fixes an hardware error in the simcam if an file is present.
+        Creating the file 'ERROR_STATE_FILE' (or by deleting this file) in the folder 'model.BASE_DIRECTORY' the
+        state of the simcam is changed to and error (or fixed by putting the state back to the standard 'running').
+        '''
+        try:
+            while self._is_running:
+                time.sleep(0.3)  # max 3 Hz
+
+                # Check if an error file is present
+                error_file_present = os.path.isfile(os.path.join(model.BASE_DIRECTORY, ERROR_STATE_FILE))
+                if error_file_present and not isinstance(self.state.value, model.HwError):
+                    self.state._set_value(
+                            model.HwError("Camera disconnected due to forced testing error."),
+                            force_write=True)
+
+                elif not error_file_present and isinstance(self.state.value, model.HwError):
+                    self.state._set_value(model.ST_RUNNING, force_write=True)
+
+        except Exception as e:
+            logging.warning("In changing states in SimCam the error '%s' occurred", e)
 
 class SimpleDataFlow(model.DataFlow):
     def __init__(self, ccd):
--- a/src/odemis/gui/log.py
+++ b/src/odemis/gui/log.py
@@ -21,9 +21,12 @@ from __future__ import division, print_function
 
 import collections
 import logging
+import time
 from logging.handlers import RotatingFileHandler
 from odemis.gui import FG_COLOUR_ERROR, FG_COLOUR_WARNING, FG_COLOUR_DIS, FG_COLOUR_MAIN
 from odemis.gui.util import wxlimit_invocation, get_home_folder
+from odemis.gui.comp.popup import show_message
+from odemis.model import ST_RUNNING, HwError
 import os.path
 import sys
 import threading
@@ -202,3 +205,29 @@ class TextFieldHandler(logging.Handler):
                 self.textfield.Remove(0, first_new)
 
         self.textfield.Refresh()
+
+# List for passing component.name to the function stage_change_pop_up
+state_subscribers = []
+# List with reference to components observed for a state change
+observed_components = []
+def observe_comp_state(comps):
+    '''
+    Function which produces an warning/information pop-up in the OS if an error or recovery of an error occurs
+    :param comps: list with all the components and their data
+    '''
+    global observed_components
+    observed_components = comps
+    for component in comps:
+        def state_change_pop_up(component_state_value, component_name=component.name):
+            if component_state_value == ST_RUNNING:
+                show_message(wx.GetApp().main_frame, 'Recovered ' + component_name,
+                             'Functionality of the "' + component_name + '" is recovered successfully.',
+                             timeout=3.0, level=logging.INFO)
+
+            elif isinstance(component_state_value, HwError):
+                show_message(wx.GetApp().main_frame, 'Error in ' + component_name, str(component_state_value),
+                             timeout=5.0, level=logging.WARNING)
+
+        # Keep a reference to each subscriber function so they won't get dereferenced (because VA's use weakrefs)
+        state_subscribers.append(state_change_pop_up)
+        component.state.subscribe(state_change_pop_up)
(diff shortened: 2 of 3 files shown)


src/odemis/driver/simcam.py
```diff
@@ -142,6 +143,48 @@ def __init__(self, name, role, image, dependencies=None, daemon=None, blur_facto
...
+
+        tnext = 0
+        while True:
+            tnow = time.time()
+
+            # sleep a bit to avoid refreshing too fast
+            tsleep = tnext - tnow
+            if tsleep > 0.001:
+                time.sleep(tsleep)
+            tnext = time.time() + 0.1  # max 10 Hz
```
pieleric: No need to make it so complex, as the rest of the block has a (approximately) fixed duration.
=> Replace by `time.sleep(0.1) # max 10 Hz`.

src/odemis/driver/simcam.py
```diff
@@ -142,6 +143,48 @@ def __init__(self, name, role, image, dependencies=None, daemon=None, blur_facto
...
+        After the error is produced the files should be deleted
+        :return:
+        '''
+        def attempt_to_delete_file(del_file):
+            try:
+                os.remove(del_file)
+            except:
+                pass
+
+        path = '/var/run/odemisd'
```
pieleric: Don't hard code it here, just use `model.BASE_DIRECTORY`.

src/odemis/driver/simcam.py
```diff
@@ -142,6 +143,48 @@ def __init__(self, name, role, image, dependencies=None, daemon=None, blur_facto
...
+        [attempt_to_delete_file(del_file) for del_file in [path + '/simcam.HW.error', path + '/simcam.HW.good']]
```
pieleric: Please put the filename in a constant (defined at the top of the file). Eg:
`ERROR_STATE_FILE = "simcam-hw-error"`
Also, note I'd suggest a different name. In particular, it's best to have at most one . in the filename, for the extension. As this file contains *nothing*, it's fine to have not extension.

src/odemis/driver/simcam.py
```diff
@@ -142,6 +143,48 @@ def __init__(self, name, role, image, dependencies=None, daemon=None, blur_facto
...
+        by creating the files 'simcam.HW.error' or 'simcam.HW.good' in the folder '/var/run/odemisd' and hardware
+        error can be produced in the Sim Cam or the error can be 'fixed'
+        After the error is produced the files should be deleted
+        :return:
+        '''
+        def attempt_to_delete_file(del_file):
+            try:
+                os.remove(del_file)
+            except:
+                pass
```
pieleric: Don't try to delete the file, as you'll enter a whole rat nest.
I'd suggest a slightly different behaviour:
* only one single file, /var/run/odemisd/simcam-hw-error`
* if it's not present, the state is "good"
* if it's present, the state is "bad"
* Let the user add/remove the file.

src/odemis/driver/simcam.py
```diff
@@ -142,6 +143,48 @@ def __init__(self, name, role, image, dependencies=None, daemon=None, blur_facto
...
+        self._errorCreationThread.daemon = True
+        self._errorCreationThread.start()
+
+    def _state_error_creation(self):
+        '''
+        by creating the files 'simcam.HW.error' or 'simcam.HW.good' in the folder '/var/run/odemisd' and hardware
+        error can be produced in the Sim Cam or the error can be 'fixed'
+        After the error is produced the files should be deleted
+        :return:
+        '''
```
pieleric: For the main function of thread, I'd recommend to surround the entire function by a try/except, so that if an exception happens, you can log the exception (instead of it being sent to a black hole).

src/odemis/driver/simcam.py
```diff
@@ -142,6 +143,48 @@ def __init__(self, name, role, image, dependencies=None, daemon=None, blur_facto
...
+                                                     args = (),
+                                                     name = "Creating and state error")
+        self._errorCreationThread.daemon = True
+        self._errorCreationThread.start()
+
+    def _state_error_creation(self):
+        '''
+        by creating the files 'simcam.HW.error' or 'simcam.HW.good' in the folder '/var/run/odemisd' and hardware
+        error can be produced in the Sim Cam or the error can be 'fixed'
+        After the error is produced the files should be deleted
```
pieleric: Please change to start by first describing what this function is for.

src/odemis/driver/simcam.py
```diff
@@ -142,6 +143,48 @@ def __init__(self, name, role, image, dependencies=None, daemon=None, blur_facto
...
         self.softwareTrigger = model.Event()
 
+        # show_message(None, 'starting and error wise thread', 'test error', timeout = 1.0)
+        self._errorCreationThread = threading.Thread(target = self._state_error_creation,
+                                                     args = (),
+                                                     name = "Creating and state error")
+        self._errorCreationThread.daemon = True
+        self._errorCreationThread.start()
+
+    def _state_error_creation(self):
```
pieleric: "creation" doesn't reality fit what it does, and the name doesn't indicate it's for a thread.
Maybe rename to `_state_error_run`.

src/odemis/driver/simcam.py
```diff
@@ -142,6 +143,48 @@ def __init__(self, name, role, image, dependencies=None, daemon=None, blur_facto
         # Convenience event for the user to connect and fire
         self.softwareTrigger = model.Event()
 
+        # show_message(None, 'starting and error wise thread', 'test error', timeout = 1.0)
+        self._errorCreationThread = threading.Thread(target = self._state_error_creation,
+                                                     args = (),
+                                                     name = "Creating and state error")
```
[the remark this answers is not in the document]
Kleijwegt: code format convention (in the file) <-- whats file?

src/odemis/driver/simcam.py
Kleijwegt: do you mean to include the part:

`            if os.path.isfile(path + '/simcam.HW.error'):
                self.state._set_value(model.HwError("Camera disconnected. Now recovering..."), force_write = True)
                os.remove(path + '/simcam.HW.error')

            elif os.path.isfile(path + '/simcam.HW.good'):
                self.state._set_value('running', force_write = True)
                os.remove(path + '/simcam.HW.good')`
in a try expect block?

src/odemis/driver/simcam.py
Kleijwegt: That would mean there is now feedback to the user whether or not the function executed correctly. If a user now sees that /var/run/odemisd/simcam-hw-error is not deleted this means that the thread did not execute correctly.
Plus this way ensures that the simulation will not start with a bunch of error's blocking the start up of the simulator in total.

> Because of the try statement the rats in the rats nest maybe won't bite that fast?

src/odemis/driver/simcam.py
Kleijwegt: changed it to one dot file extension is now .error or .good

src/odemis/driver/simcam.py
pieleric: I mean that the first line of the function should be `try:`, and the *entire* current code should be within that block.

src/odemis/driver/simcam.py
pieleric: The function should work fine. Also, the user's feedback is via the .state of the component.

src/odemis/driver/simcam.py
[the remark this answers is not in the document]
pieleric: I meant the "code style". There is a code style convention for Odemis. It's pep8 + some few tweaks. One of the few tweaks is "look at the code style" of the current file you're editing, and follow it ;-)

src/odemis/driver/simcam.py
```diff
@@ -142,6 +145,41 @@ def __init__(self, name, role, image, dependencies=None, daemon=None, blur_facto
...
+        def attempt_to_delete_file(del_file):
+            try:
+                os.remove(del_file)
+            except:
+                pass
+
+        [attempt_to_delete_file(del_file) for del_file in [model.BASE_DIRECTORY + '/' + ERROR_STATE_FILE,
+                                                           model.BASE_DIRECTORY + '/' + GOOD_STATE_FILE]]
+
+        while True:
```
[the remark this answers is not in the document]
Kleijwegt: @pieleric  is this not easier solved by making this thread a non-deamon?

Kleijwegt:
Ready to be merged

src/odemis/gui/log.py
```diff
@@ -202,3 +205,26 @@ def write_to_field(self):
...
+state_change_function_list = [] #List for passing component.name to the function stage_change_pop_up
+def observe_comp_state (comps_with_role):
+    '''
+    Function which produces an warning/information pop-up in the OS if an error or recovery of an error occurs
+    :param comps_with_role: list with all the components and their data
+    '''
+    for component in comps_with_role:
+        def state_change_pop_up(component_state_value, component_name=component.name):
+            if component_state_value == ST_RUNNING:
+                show_message(None, 'Recovered ' + component_name,
```
pieleric, line 218: Don't pass `None` as parent, but the main frame. `wx.GetApp().main_frame`.

src/odemis/gui/log.py
```diff
@@ -202,3 +205,26 @@ def write_to_field(self):
...
+                             'Functionality of the "' + component_name + '" is recovered successfully',
+                             timeout=3.0, level=logging.INFO)
+
+            elif isinstance(component_state_value, HwError):
+                show_message(None, 'Error in ' + component_name, str(component_state_value),
+                             timeout=5.0, level=logging.WARNING)
+
+        state_change_function_list.append(state_change_pop_up)
```
pieleric, line 226: Add a comment explaining why you keep a reference to them -> so that they are not dereferenced (as VAs use weakrefs).

src/odemis/gui/model/__init__.py
```diff
@@ -288,6 +289,9 @@ def __init__(self, microscope):
 
                 # Otherwise, just not interested by this component
 
+            #If the state of an HW component changes create an OS pop up message
+            observe_comp_state(comps_with_role)
```
pieleric, line 293: As mentioned before, it's probably best to listen to *every* component. As the ones with roles can be some simple wrapper on top of the real hardware, which will have an error.

You probably can reuse the `model.getComponents()` and instead of using it directly in the loop, save it in a variable.
Also, note that the function will need to hold a reference to the VA (or the component itself), as they will be dereferenced if you don't. (I guess it works for now because the SettingsObserver already holds a reference to each of the components with role).

PR #1325, opened 2020-01-16
driver: fix some more bytes/string confusion

pieleric:


pieleric:
Note that this is probably much easier to review when looking at each commit separately.

PR #1327, opened 2020-01-17
driver hamamatsurx: Migrate to using bytes for Python 3

pieleric:
In Python 3, the network sockets send and receive bytes.

Also adjust a couple of extra places in the message handling.
In particular, the CameraInfo now behaves a little bit less special, and
each line is return as a separate return value instead of one value
containing a list.

These changes haven't be tested on the hardware yet.

### DIFF
--- a/src/odemis/driver/hamamatsurx.py
+++ b/src/odemis/driver/hamamatsurx.py
@@ -112,13 +112,13 @@ class ReadoutCamera(model.DigitalCamera):
         super(ReadoutCamera, self).__init__(name, role, parent=parent, **kwargs)
 
         try:
-            self._hwVersion = cam_info[0][0] + ", " + cam_info[0][1] + ", " + cam_info[0][2]   # needs to be a string
+            self._hwVersion = cam_info[0] + ", " + cam_info[1] + ", " + cam_info[2]  # needs to be a string
         except:
             self._hwVersion = "N.A."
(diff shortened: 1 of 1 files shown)


src/odemis/driver/hamamatsurx.py
```diff
@@ -1157,10 +1160,10 @@ def readCommandResponse(self):
                         continue  # return to try-statement and start receiving again
 
                     if self._getReadoutCamInfo:
-                        # command parent.CamParamGet("Setup", "CameraInfo") behaves differently then all other commands.
+                        # HACK: command CamParamGet("Setup", "CameraInfo") behaves differently then all other commands.
```
Sabrina85: I think I made a language error here. Think it should be `...from all the other commands`

Sabrina85:
Will Iheb test these changes?

src/odemis/driver/hamamatsurx.py
tepals: just a very small note: different than is not always incorrect as long as you spell it with an "a" and not an "e". https://www.grammar.com/different-from-vs-different-than/

pieleric:
Fixed the spelling.

pieleric:
> Will Iheb test these changes?
I'll let him know that next update should be double-checked. However, I'm not in a rush to test the Python 3 support. For now, I'd just check that it still works in Python 2.

PR #1328, opened 2020-01-17
Xt client extra functionality

tepals:


src/odemis/driver/xt_client.py
```diff
@@ -21,20 +21,21 @@
...
-import msgpack_numpy as m
-from Pyro5.api import Proxy
+import msgpack_numpy
+from Pyro5.api import Proxy, config
 
 from odemis import model
 from odemis import util
 from odemis.model import CancellableThreadPoolExecutor, HwError, isasync, CancellableFuture
 
+config.SERIALIZER = 'msgpack'
```
pieleric: `config` is a very generic term. It's not very convenient to have it in the main namespace. Moreover, it's only used here. So I'd suggest to explicitly write `Pyro5.api.config`.

PR #1330, opened 2020-01-20
bugfix tiff: don't fail saving image if an element of the metadata is bytes

philipwinklerdelmic:


src/odemis/dataio/tiff.py
```diff
@@ -445,6 +445,14 @@ def _convertToOMEMD(images, multiple_files=False, findex=None, fname=None, uuids
             if model.MD_LENS_NAME in da0.metadata:
                 obj.attrib["Model"] = da0.metadata[model.MD_LENS_NAME]
 
+        # Make sure all string metadata values are unicode strings (not bytes),
+        # otherwise serialization will fail later on
+        for da in g:
+            for key, value in da.metadata.items():
+                print(key, value)
```
pieleric: Drop the `print()`.

src/odemis/dataio/tiff.py
```diff
@@ -445,6 +445,14 @@ def _convertToOMEMD(images, multiple_files=False, findex=None, fname=None, uuids
...
+                if isinstance(value, bytes):
+                    da.metadata[key] = value.decode('utf-8')
```
pieleric: Please put this in a `try/except Exception`. On error: log a warning and replace the value by an empty string.

philipwinklerdelmic:
updated

PR #1331, opened 2020-01-21
update spot grid: flip translation and rotation

tepals:
- flip the displayed rotation and translation, so the values
correspond to what they would be if the origin of the coordinate system is
in the bottom left of the image.
- update default spots to 8, 8 because that is what we now always use.

### DIFF
--- a/scripts/spot-grid.py
+++ b/scripts/spot-grid.py
@@ -43,7 +43,7 @@ class VideoDisplayerGrid(VideoDisplayer):
         coming pictures
         """
         self.app = ImageWindowApp(title, size)
-        self.gridsize = (14, 14) if gridsize is None else gridsize
+        self.gridsize = (8, 8) if gridsize is None else gridsize
 
     def new_image(self, data):
         """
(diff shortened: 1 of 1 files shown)


scripts/spot-grid.py
```diff
@@ -52,6 +52,9 @@ def new_image(self, data):
         data (numpy.ndarray): an 2D array containing the image (can be 3D if in RGB)
         """
         self.app.spots, self.app.translation, self.app.scaling, self.app.rotation = FindGridSpots(data, self.gridsize)
+        self.app.translation[1] = data.shape[0] - self.app.translation[1]
+        self.app.rotation = -self.app.rotation
```
pieleric: That's becoming a little weird to read. I'd suggest:
```
self.app.spots, trans, scale, rot = FindGridSpots(data, self.gridsize)
self.app.translation = trans[0], data.shape[0] - trans[1]
self.app.scale = scale
self.app.rotation = -rot
```

tepals: good suggestion, fixed!

PR #1332, opened 2020-01-21
model vattribute: also accepts numpy integers in IntVA and ResolutionVA

pieleric:
Numpy integers easly creep everywhere. Normally that's fine because they
quack like an int (except for the over- under-flows of course, but that
typically doesn't happen in the number we handle).
So let's accept them also in the VAs.

Note: in Python 2, numpy.int* were a subclass of int, so it worked
fine (not for numpy.uint* though). In Python 3, this "link" has been removed,
that's why we have to explicitly add it

### DIFF
--- a/src/odemis/model/_vattributes.py
+++ b/src/odemis/model/_vattributes.py
@@ -636,7 +636,7 @@ class IntVA(VigilantAttribute):
 
     def _check(self, value):
         # we really accept only int, to avoid hiding lose of precision
-        if not isinstance(value, (int, long)):
+        if not isinstance(value, (int, long, numpy.integer)):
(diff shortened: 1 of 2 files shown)


pieleric:
Indeed, sometimes numpy ints crop in. I've tried to fix it with commit 2a27c083c7b9eca19c9aff580db94351b20f7bb9 , but it seems it still didn't catch some other cases (cf test case failures).
We could hunt all of the places, but I had a feeling that it'd get quite ugly with lots of `int()` sparkled around the code, which doesn't make the code more readable at all. In addition. it could reappear from time to time when adding new code.
So my conclusion is that if the rest of the code is happy with a numpy.int, let's just be fine with it in the VA.

PR #1333, opened 2020-01-21
plugins: clean-up

pieleric:


plugins/secom_cl.py
```diff
@@ -1,24 +1,39 @@
...
+e-beam spots on the sample along a grid.
 
-
-run as:
-./secom_cl --xrep 45 --yrep 5 --prefix filename-prefix
-
---prefix indicates the beginning of the filename.
 The files are saved in TIFF, with the y, x positions of the ebeam (in nm) in the name,
-the total number of ebeam positions in x and y, the physical distance between positions in x and y
-and the type of the acquisition.
```
Kleijwegt: Why is the part on how to run the script + the options deleted here?

pieleric: That explained (briefly) in the commit text. This functionality was removed 8 months ago. So it makes no sense to still document it.

PR #1336, opened 2020-01-22
gui cont export: only enable "Export" button when there is something to export

pieleric:
When pressing "Export" with a view which has no stream (shown), an error
would be raised. That's was a bug added by the polarimetry export, which
assumed there is always at least one stream.

=> Check whether there is a stream.
=> Instead checking if the export button should be shown based on the
the global streams, check the streams of the view.

### DIFF
--- a/src/odemis/gui/cont/export.py
+++ b/src/odemis/gui/cont/export.py
@@ -63,39 +63,42 @@ class ExportController(object):
         tab_data: MicroscopyGUIData -- the representation of the microscope GUI
         main_frame: (wx.Frame): the whole GUI frame
         """
-
         self._data_model = tab_data
-        self._main_data_model = tab_data.main
         self._main_frame = main_frame
         self._tab_panel = tab_panel
+        self._viewports = list(viewports.keys())
         self._conf = get_acqui_conf()
 
-        # Listen to "acquire image" button
+        # Listen to "export" button and menu
         self._tab_panel.btn_secom_export.Bind(wx.EVT_BUTTON, self.on_export)
-
-        self._viewports = list(viewports.keys())
-
         self._main_frame.Bind(wx.EVT_MENU, self.on_export, id=self._main_frame.menu_item_export_as.GetId())
-
         self._main_frame.menu_item_export_as.Enable(False)
 
         # subscribe to get notified about tab changes
         self._prev_streams = None # To unsubscribe afterwards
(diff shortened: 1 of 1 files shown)


src/odemis/gui/cont/export.py
```diff
@@ -63,40 +63,44 @@ def __init__(self, tab_data, main_frame, tab_panel, viewports):
...
-        if self._prev_streams:
-            self._prev_streams.unsubscribe(self.on_streams_change)
-            self._prev_streams = None
-        if tab is not None and tab.name == 'analysis':
+        if tab is not None and tab.tab_data_model is self._data_model:
             # Only let export item to be enabled in AnalysisTab
-            tab.tab_data_model.streams.subscribe(self.on_streams_change, init=True)
-            self._prev_streams = tab.tab_data_model.streams
+            #tab.tab_data_model.streams.subscribe(self.on_streams_change, init=True)
+            #self._prev_streams = tab.tab_data_model.streams
```
pieleric: Drop these lines.

src/odemis/gui/cont/export.py
```diff
@@ -63,40 +63,44 @@ def __init__(self, tab_data, main_frame, tab_panel, viewports):
...
 
     @call_in_wx_main
     def on_tab_change(self, tab):
         """ Subscribe to the the current tab """
-        if self._prev_streams:
-            self._prev_streams.unsubscribe(self.on_streams_change)
-            self._prev_streams = None
-        if tab is not None and tab.name == 'analysis':
+        if tab is not None and tab.tab_data_model is self._data_model:
             # Only let export item to be enabled in AnalysisTab
```
pieleric: Update to read "in "our" tab, ie the Analysis tab"

src/odemis/gui/cont/export.py
```diff
@@ -63,40 +63,44 @@ def __init__(self, tab_data, main_frame, tab_panel, viewports):
...
-            tab.tab_data_model.streams.subscribe(self.on_streams_change, init=True)
-            self._prev_streams = tab.tab_data_model.streams
+            #tab.tab_data_model.streams.subscribe(self.on_streams_change, init=True)
+            #self._prev_streams = tab.tab_data_model.streams
+            self._data_model.focussedView.subscribe(self.on_view_change, init=True)
```
pieleric: This keeps subscribing without ever unsubscribing. Either subscribe once at the beginning, or unsubscribe whenever the tab changes.

PR #1337, opened 2020-01-22
projection: modified mean projection to use floating point values

amuskens:
I think this is the fix.

Tested:
- Export of PNG and CSV
- Test cases


pieleric:
No, that's not the correct fix. You can check that by exporting to CSV, and noticing that the "intensity" is still always a integer. Also, note that `numpy.mean` *always* returns a float64, so your change has no effect. What should be done is to edit the line 1652, by removing the `.astype(...)`. 

Note that the spectrum plot rounds the values to 3 significant numbers, so it's a little hard to see the difference, but the graph should adjust properly.

amuskens:
Just tested it - seems to work without any problems - also in export.

PR #1338, opened 2020-01-23
acq stream ARProjection: small fixes

pieleric:


src/odemis/acq/stream/_projection.py
```diff
@@ -379,23 +379,19 @@ def _project2Polar(self, ebeam_pos, pol_pos):
...
         try:
-            polar_data = polar_cache[ebeam_pos][pol_pos]
+            polar_data = polar_cache.setdefault(ebeam_pos, {})[pol_pos]
         except KeyError:
             # Compute the polar representation
             data = self.stream._pos[ebeam_pos + (pol_pos,)]
-            # TODO stream._pos can be then also be ordered first ebeam_pos, then pol_pos
-            # TODO would also simplify the check for the correct bg image etc.
+            # TODO stream._pos can be then also be structured ebeam_pos/pol_pos.
+            # That would also simplify the check for the correct bg image etc.
```
Sabrina85: I recently found out, that if a TODO is multiple lines it will appear also in yellow if there is an extra space in front of the extra lines. So just add another space before `That....` and it will be also part of the TODO.

src/odemis/acq/stream/_projection.py
```diff
@@ -379,23 +379,19 @@ def _project2Polar(self, ebeam_pos, pol_pos):
...
         :returns: (2D DataArray) The polar projection.
         """
-
         # Note: Need a copy of the link to the dict. If self._polar_cache is reset while
         # still running this method, the dict might get new entries again, though it should be empty.
         polar_cache = self._polar_cache
 
         try:
-            polar_data = polar_cache[ebeam_pos][pol_pos]
+            polar_data = polar_cache.setdefault(ebeam_pos, {})[pol_pos]
```
Sabrina85: Nice! Didn't know this method before.

src/odemis/acq/stream/_projection.py
```diff
@@ -540,10 +528,7 @@ def projectAsVis(self):
 
         if hasattr(self, "polarization"):
             for pol_pos in self.polarization.choices:
-                try:
-                    data = self._polar_cache[ebeam_pos][pol_pos]
-                except KeyError:
-                    data = self._project2Polar(ebeam_pos, pol_pos)
+                data = self._project2Polar(ebeam_pos, pol_pos)
```
Sabrina85: good catch

src/odemis/acq/stream/_projection.py
Kleijwegt: Is this also in eclipse? @philipwinklerdelmic @pieleric

philipwinklerdelmic: I don't think the TODO appears in any special color in eclipse (even a single line TODO), it just looks like a normal comment and on the scrollbar we have a blue marker.

pieleric: Added space.

Issue #319, opened 2017-04-19
hdf5 file_format


thomasaarholt:
Currently, there is nothing that clearly distinguishes a `.h5` file exported by odemis from any other hdf5 file. In order to import such a file into other software, it is very helpful to include some marker that implies that "this file is a format created by odemis", so that the file can be imported in the appropriate way. (A "hacky way" to do this would be to check if the name of the first group in the file is "Acquisition0", or something similar).

Over at [Hyperspy](https://github.com/hyperspy/hyperspy), we save our hdf5 files with `file_format = Hyperspy` and `file_format_version = 2.2` metadata. This enables any file reader to quickly recognise the file type and treat the rest of the data as such. 

The [code ](https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/io_plugins/hdf5.py#L709)is simple enough - I'll be happy to make a PR.

```python
def file_writer(filename,
                signal,
                *args, **kwds):
    with h5py.File(filename, mode='w') as f:
        f.attrs['file_format'] = "HyperSpy"
        f.attrs['file_format_version'] = version
```

pieleric:
What do you suggest to implement? Extending Odemis HDF5 loader to support also HyperSpy format, or to extend the HDF5 writer to also write HyperSpy compatible metadata?

In the first case, any patch is welcome. In the second case, I'd rather like to avoid having two HDF5 sub-formats, between which the user would need to choose when saving the file. However, if it's possible to save both the current metadata and the HyperSpy metadata in the same file, there wouldn't be any issue to accept such an extension.

thomasaarholt:
Sorry, I was clearly distracted whilst writing. I've edited the post clarifying the following:

I would like to read the odemis hdf5 file using Hyperspy. However, in order to be able to distinguish the odemis file from any other generic `somefile.h5 `, it needs some metadata telling the file reader what the file is. Instead of checking for some pecularity within the format, it's much nicer to have something like: `file_format = odemis` in the metadata of the root of the file.

thomasaarholt:
To answer your final sentence - The addition of the metadata (referred to as an "attribute") should not cause any issue. It is simply a line in addition to the currently existing metadata, which (found by opening my odemis `.h5` in [HDFView](https://support.hdfgroup.org/products/java/hdfview/)) is currently:
```
group_size = 4
Number of attributes = 0
```

pieleric:
[...]

PR #1343, opened 2020-02-05
dataio hdf5: store the odemis version directly in the metadata

pieleric:
Normally all future Odemis versions should support reading back the old
format, and we should use just the file structure and the
FileSpecificationVersion.

Nevertheless, it's useful for debugging to know which exact version of
Odemis generated the file.

### DIFF
--- a/src/odemis/dataio/hdf5.py
+++ b/src/odemis/dataio/hdf5.py
@@ -26,6 +26,7 @@ import collections
 import h5py
 import logging
 import numpy
+import odemis
 from odemis import model
 from odemis.util import spectrum, img, fluo
 import os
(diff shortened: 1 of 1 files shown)


pieleric:
"Fully importing `odemis`" might sound a lot of imports, but it is actually a very tiny module, which only defines some constants. That's the standard way to get the odemis version.

Concerning the order, I agree with you that it makes more sense to have it before the `from odemis`. The Eclipse code formatter has a different idea, but I'll change it.

Kleijwegt:
@pieleric After merging this pull request should issue "hdf5 file_format #319" be closed?

pieleric:
@Kleijwegt, good remark. I had forgotten about this issue. Maybe we should add *another* tag, which is "file_format", as described in issue #319 (directly at the root for the file).

PR #1347, opened 2020-02-07
Setup Python Dev in Windows

amuskens:
Updated for new instructions on building Odemis in Windows with python 3.6.8

### DIFF
--- a/src/odemis/model/_core.py
+++ b/src/odemis/model/_core.py
@@ -31,7 +31,6 @@ import os
 import threading
 from future.moves.urllib.parse import quote
 from odemis.util import inspect_getmembers
-import resource
 
 
 # Pyro4.config.COMMTIMEOUT = 30.0 # a bit of timeout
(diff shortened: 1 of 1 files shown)


requirements-win-py3.6.txt
```diff
@@ -0,0 +1,18 @@
...
+decorator>=4.0.9
+future>=0.16.0
+h5py
+libtiff>=0.4.2
+matplotlib>=1.4.3,<3.0
+numpy>=1.11.0,<1.17
+opencv>=2.4.11,<4.0
+Pillow>=3.2.0
+pycairo>=1.10.0
+PyInstaller
```
tepals: Why don't you specify the version of h5py and PyInstaller?

src/odemis/model/_core.py
```diff
@@ -31,7 +31,9 @@
 import threading
 from future.moves.urllib.parse import quote
 from odemis.util import inspect_getmembers
-import resource
+
+if os.name != 'nt':
+    import resource
```
pieleric: Let's move these lines together with the (only) function that uses it. Then you can put a if block around the entire import + function, and if it's windows, just define the function as doing nothing.

pieleric:
I've updated the doc in PR #1350. So I'd suggest to drop that part of your PR, and just fix the `ressource` module issue here.

src/odemis/model/_core.py
```diff
@@ -219,24 +218,35 @@ def load_roattributes(self, roattributes):
...
+        """
+
+        cur_limit = resource.getrlimit(resource.RLIMIT_NOFILE)
+        try:
+            resource.setrlimit(resource.RLIMIT_NOFILE, (cur_limit[0] + inc, cur_limit[1]))
+        except ValueError:
+            # this happens when starting odemis from eclipse
+            logging.warning("Maximum number of open files is already at its limit %s." % cur_limit[0])
+
+    # Container management functions and class
```
pieleric: Please leave this comment out of the `if`. It should be just before the next `class`.

amuskens:
Should be ready now!

PR #1349, opened 2020-02-11
Improvement of peak fitting

VictoriaDel:


src/odemis/util/peak.py
```diff
@@ -33,34 +33,58 @@
...
+    return gau
+
 
+def GaussianEnergyFit(data, *peaks):
+    """
+    Applies gaussian fitting to data given the "peaks" parameters.
+    data (1d array of floats): the energy list for each N pixel
+    peaks (list of floats): series of pos, width, amplitude and initial offset (currently given in the energy domain)
+    """
+    gau = peaks[-1] * constant / data**2
```
pieleric: Please add a comment on why the offset is done differently (ie, it's a constant in space domain, so converted to energy here).

src/odemis/util/peak.py
```diff
@@ -237,7 +274,8 @@ def _DoFit(self, future, spectrum, wavelength, type='gaussian'):
             logging.debug("Starting peak detection on data (len = %d) with window = %d",
                           len(wavelength), window_size)
             try:
-                width = PEAK_WIDTHS[type]
+                initial_width = PEAK_WIDTHS[type]
+                width = calc_width(initial_width, wavelength)
```
pieleric: I don't think it's useful to put it in a separate function. I'd find it more clear to directly see `width = wl_rng * WIDTH_RATIO`

src/odemis/util/peak.py
```diff
@@ -255,11 +293,15 @@ def _DoFit(self, future, spectrum, wavelength, type='gaussian'):
...
-                    fit_list.append(amplitude)
-                # Initialize offset to 0
-                fit_list.append(0)
+                    if type in {'gaussian_energy', 'lorentzian_energy'}:
+                        fit_list.extend(peak_to_energy(pos, width, amplitude))
+                    else:
+                        peak_params = [pos, width, amplitude]
+                        fit_list.extend(peak_params)
+
+                # Initialize offset to zero
```
pieleric: It's pretty clear that it's initialised to 0. A more interesting comment would be to indicate why not other values. In particular, we could mention that we've tried to use the minimum value, but that tended to lead the fitting to prefer having a large offset instead of smaller peaks, especially in energy domain.

src/odemis/util/peak.py
```diff
@@ -255,11 +293,15 @@ def _DoFit(self, future, spectrum, wavelength, type='gaussian'):
...
-                    fit_list.append(pos)
-                    fit_list.append(width)
-                    fit_list.append(amplitude)
-                # Initialize offset to 0
-                fit_list.append(0)
+                    if type in {'gaussian_energy', 'lorentzian_energy'}:
+                        fit_list.extend(peak_to_energy(pos, width, amplitude))
+                    else:
+                        peak_params = [pos, width, amplitude]
+                        fit_list.extend(peak_params)
```
pieleric: Just write `fit_list.extend([pos, width, amplitude])`.

src/odemis/gui/comp/viewport.py
```diff
@@ -1520,7 +1520,7 @@ def _on_new_data(self, data):
         data (1D DataArray)
         """
         if data is not None and data.size:
-            wll, unit_x = spectrum.get_spectrum_range(data)
+            wll, unit_x = spectrum.get_spectrum_range(data) # unit_x in 'm' or 'px'
```
Kleijwegt: description of variable unit_x is given, but maybe it is nice to also comment on wll?

src/odemis/util/peak.py
```diff
@@ -318,14 +368,110 @@ def estimateFitTime(self, data):
...
+        wavelength: the list of wavelengths in "m" or "px
+
+    Returns: peak width estimation
+
+    """
+    peak_width_ratio = (wavelength[-1] - wavelength[0]) / 10
+    return initial_width * peak_width_ratio
+
+
+def applyJacobian_x(wavelength):
```
Kleijwegt: non consequent naming with functions above, first ones are snake_case then this one uses camelCase**_x**. Functions after this have the same

src/odemis/util/peak.py
```diff
@@ -318,14 +368,110 @@ def estimateFitTime(self, data):
...
+    Returns:
+        curve_space (1d array of floats): dataset of points representing the curve in space domain
+
+    """
+    wavelength = numpy.asarray(wavelength)
+    curve_space = (curve_energy * constant) / (wavelength ** 2)
+    return curve_space
+
+
+def Curve(wavelength, peak_parameters, offset, type='gaussian_space'):
```
Kleijwegt: start with a capital, while being a non protected member? Maybe I am being a bit hard on method naming (I ain't a saint in that either)

src/odemis/util/peak.py
```diff
@@ -318,14 +372,96 @@ def estimateFitTime(self, data):
...
+    Converts the peaks to energy domain.
+    Args:
+        pos: the center position of the peak in 'm'
+        width: the width of the peak in 'm'
+        amplitude: the amplitude of the peak in space domain
+
+    Returns: peak parameters as (pos, width, amplitude) in energy domain
+
+    """
+    amplitude_e = (amplitude * pos ** 2) / WL_TO_ENERGY
```
pieleric: If you sort them in the same order as they are defined in the function signature, it would be a little bit easier to read.

src/odemis/util/peak.py
```diff
@@ -254,12 +285,27 @@ def _DoFit(self, future, spectrum, wavelength, type='gaussian'):
...
-                    fit_list.append(pos)
-                    fit_list.append(width)
-                    fit_list.append(amplitude)
-                # Initialize offset to 0
-                fit_list.append(0)
+                    if type in {'gaussian_energy', 'lorentzian_energy'}:
+                        fit_list.extend(peak_to_energy(pos, width, amplitude))
+                        # lower & upper bounds for center position, width, amplitude in energy domain
+                        lower_bounds.extend([0, 0, 0])
+                        upper_bounds.extend([100, 10, numpy.inf])
```
pieleric: These constants are not explained enough. You could mention that 100 nm ~= 12 and 1000nm ~= 0.12, so they are very large bounds.
Alternatively, you could use the same formula as for the space domain (but add a `WL_TO_ENERGY / `)
Another way could be to move the space domain formula up, and to call `peak_to_energy()` for the lower and upper bounds too.

PR #1350, opened 2020-02-13
doc: add instructions for Python 3.6 (on Windows)

pieleric:
Create a different requirements.txt, to avoid the user having to editing
too much the file.

### DIFF
--- a/doc/develop/extending.rst
+++ b/doc/develop/extending.rst
@@ -173,13 +173,19 @@ right-click and select *Git Bash here*. Then type::
 Creating the Odemis environment
 -------------------------------
 
-Install `Anaconda <https://www.anaconda.com/distribution/>`_ with Python 2.7 32 bits
-(or Python 3.6, but you will have to adjust some of the commands).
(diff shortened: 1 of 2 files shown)


doc/develop/extending.rst
```diff
@@ -188,23 +194,42 @@ Open the *Anaconda prompt* and type::
...
+For Python 3, open the *Anaconda prompt* and type::
+
+   cd Documents\odemis
+   conda create -y --name odemisdev python==3.6.8
+   conda activate odemisdev
+   conda config --append channels conda-forge
+   conda install --name odemisdev --file requirements-py3.txt
+   conda develop src
+   pip install git+https://github.com/delmic/Pyro4.git
+
```
Kleijwegt: Maybe nice as a task in the back-log wouldn't it be nice to write an automatic .exe/installer which does all of this for the user?

pieleric: Even better would be to have Odemis packaged in the anaconda repo (or pypi)... I've just added the task to the backlog.

PR #1353, opened 2020-02-14
allow changing the tint of all the 2D static streams

pieleric:


src/odemis/acq/stream/_projection.py
```diff
@@ -1014,9 +1014,15 @@ def _projectTile(self, tile):
         dims = tile.metadata.get(model.MD_DIMS, "CTZYX"[-tile.ndim::])
         ci = dims.find("C")  # -1 if not found
         # is RGB
+        tint = tuple(self.stream.tint.value)
```
Kleijwegt: I assume that tint converts a stream from grayscale to (for example) a tinted "blue"scale? Is that right?
Would be nice to add somewhere a comment which more explicitly describes the function to tint

pieleric: Indeed, I forgot to update the comments in this function. Now done.

src/odemis/acq/stream/_projection.py
```diff
@@ -1013,10 +1013,15 @@ def _projectTile(self, tile):
         """
         dims = tile.metadata.get(model.MD_DIMS, "CTZYX"[-tile.ndim::])
         ci = dims.find("C")  # -1 if not found
-        # is RGB
-        if dims in ("CYX", "YXC") and tile.shape[ci] in (3, 4):
-            # Just pass the RGB data on
+        tint = tuple(self.stream.tint.value)
```
Kleijwegt: Altough tint is a normal word, I (and Bassim) had problems understanding that it changed somthing from gray scale to a scale based on antother color. I keep thinking it is probably a 't'-type integer (t-int) so it would help to add behind this line the comment:
" value to change the tint/colorscale"

But again, tint is a normal English word.

pieleric: I understand the confusion with "int". That said, this file (and acq.stream._base) has many places with "tint". So I'd rather leave it like this here.

Kleijwegt: Ok, I understand that.
Ready to Merge! (should I now press the button since I last reviewed or do you(@pieleric ) still do that?

pieleric: Usually, I take care of merging. Although everyone can do it.

PR #1354, opened 2020-02-18
gui cont tabs: small clean-ups and improvements

pieleric:


src/odemis/gui/cont/tabs.py
```diff
@@ -4352,8 +4353,14 @@ def _create_needed_tabs(self, tab_defs, main_frame, main_data):
...
-            else:
-                tab_def["button"].Hide()  # Hide the tab button
+                assert tab.button not in buttons
+                buttons.add(tab.button)
+
+        # Hides the buttons which are not used
+        for tab_def in tab_defs:
+            b = tab_def["button"]
+            if b not in buttons:
+                b.Hide()
```
tepals: Why did you move this outside the other loop?

src/odemis/gui/cont/tabs.py
```diff
@@ -1716,6 +1716,16 @@ def __init__(self, name, button, panel, main_frame, main_data):
         self._settings_controller.setter_temporalspec_bck_file = self.set_temporalspec_background
         self._settings_controller.setter_spec_file = self.set_spec_comp
 
+        if main_data.role is None:
+            # HACK: Move the logo to inside the FileInfo bar, as the tab buttons
+            # are hidden when only one tab (ie, no backend).
```
tepals: when only one tab is selected or when there is only one tab?

src/odemis/gui/cont/tabs.py
```diff
@@ -1716,6 +1716,16 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
+            capbar = panel.fp_fileinfo._caption_bar
+            logo = main_frame.logo
+            assert main_frame.pnl_tabbuttons.GetSizer().Detach(logo)
+            logo.Reparent(capbar)
+            # Move to the right of the bar (before the icon)
+            logo.SetPosition((capbar.Size[0] - logo.Size[0] - 20 - 16 - 6, 8))
```
tepals: You could add a small comment where the  - 20 - 16 - 6 numbers are coming from

src/odemis/gui/cont/tabs.py
pieleric: Only one tab shown. -> Fixed comment.

src/odemis/gui/cont/tabs.py
pieleric: Because now buttons can be shared (see commit message). So we cannot immediately hide a button if the tab is not shown. We have to check that no tab uses this button.

src/odemis/gui/cont/tabs.py
pieleric: Indeed, that looks a little too magical... -> added extra comment.

PR #1357, opened 2020-02-24
Label tool

VictoriaDel:


src/odemis/gui/comp/overlay/world.py
```diff
@@ -833,48 +835,61 @@ def deactivate(self):
...
+    This abstract GadgetToolInterface class forms the base for a series of classes that
+    refer to gadgets tools and their functionality.
     """
 
+    def __init__(self, cnvs):
+        """ Args: (cnvs) canvas passed by the GadgetOverlay and used to draw the gadgets """
+        self.cnvs = cnvs
+
+    def start_dragging(self, drag, vpos, tool_mode):
+        """ The user can start dragging the tool when the left mouse button is pressed down """
```
pieleric: Please put the full docstring, with the argument and return descriptions here. In the subclasses, you can keep the docstring empty, or much shorter.

src/odemis/gui/comp/overlay/world.py
```diff
@@ -1231,46 +1314,350 @@ def draw(self, ctx, selected, shift=(0, 0), scale=1.0, canvas=None):
...
+            vpos: the view coordinates of the mouse cursor once left click mouse event is fired
+
+        """
+        super(LabelTool, self).start_dragging(drag, vpos, tool_mode)
+
+        if drag == gui.HOVER_START:
+            self.mode = EDIT_START_LABEL_LINE
+        elif drag == gui.HOVER_END:
+            if tool_mode == MODE_CREATE_LABEL:
+                self.mode = CREATE_LABEL
```
pieleric: This is the only place that you need the `tool_mode`. Moreover the `CREATE_LABEL` mode is there also just for the creation of the label. I think a simpler solution would be to just use `EDIT_END_LABEL_LINE` always, and detect that the text hasn't ever been set. For instance you could set `.text = None` at init, or you could just have a dedicated flag like `._ask_user_for_text = True`. So in `stop_updating_tool()`, you could check if that flag is set, and call `_edit_text()`.

src/odemis/gui/comp/overlay/world.py
```diff
@@ -1231,46 +1314,350 @@ def draw(self, ctx, selected, shift=(0, 0), scale=1.0, canvas=None):
...
+                dist = util.perpendicular_distance(self.v_start_pos, self.v_end_pos, vpos)
+                if dist < self.HOVER_MARGIN:
+                    return gui.HOVER_LINE
+
+        return gui.HOVER_NONE
+
+    def debug_edges(self, ctx):
+        super(LabelTool, self).debug_edges(ctx)
+
+        if self._edges and self._label.render_pos:
```
pieleric: Looking for `self._label.render_pos` is a little weird here, as you don't use it. Do you still need that? Or is that to check whether `t_*` are in `_edges`? If so, I'd suggest to explicitly write `'t_l' in self._edges`.

src/odemis/gui/comp/overlay/world.py
```diff
@@ -1231,46 +1314,350 @@ def draw(self, ctx, selected, shift=(0, 0), scale=1.0, canvas=None):
...
+            ctx.set_source_rgba(*self.highlight)
+            ctx.set_line_width(2)
+        else:
+            ctx.set_source_rgba(*self.colour)
+            ctx.set_line_width(1)
+        ctx.set_line_join(cairo.LINE_JOIN_MITER)
+
+        self._label.weight = cairo.FONT_WEIGHT_NORMAL
+        self._label.colour = self.highlight if selected else self.colour
+
```
pieleric: No need for empty line here.

src/odemis/gui/comp/overlay/world.py
```diff
@@ -1231,46 +1314,350 @@ def draw(self, ctx, selected, shift=(0, 0), scale=1.0, canvas=None):
...
+        elif phi > 7 * math.pi / 4:
+            self._label.align = wx.ALIGN_BOTTOM | wx.ALIGN_CENTER_HORIZONTAL
+        elif 3 * math.pi / 4 < phi < 5 * math.pi / 4:
+            self._label.align = wx.ALIGN_CENTER_HORIZONTAL
+        elif math.pi / 4 < phi < 3 * math.pi / 4:
+            self._label.align = wx.ALIGN_CENTRE_VERTICAL
+        else:  # math.pi < phi < 2 * math.pi
+            self._label.align = wx.ALIGN_RIGHT | wx.ALIGN_CENTER_VERTICAL
+
+        self._label.font_size = 14
```
pieleric: Move it to the else of the `if`... maybe a tiny be clearer.

src/odemis/gui/comp/overlay/world.py
```diff
@@ -1231,46 +1314,350 @@ def draw(self, ctx, selected, shift=(0, 0), scale=1.0, canvas=None):
...
+        if pixel_length <= 1:
+            ctx.move_to(*b_start)
+            ctx.line_to(b_start[0] + 1, b_start[1] + 1)
+        else:
+            ctx.move_to(*b_start)
+            ctx.line_to(*b_end)
+        ctx.stroke()
+
+        self._label.draw(ctx)
+        self._update_view_positions(b_start, b_end)
```
pieleric: You shouldn't call `_update_view_positions()` all the time. Maybe just move the `if` from the top to here? Or does the should the update happen before the drawing? I have a feeling this is only used for the mouse interaction, but I might be wrong.
An alternative could be to put the burden on the caller. You could have a 6th public method `sync_with_canvas()` (or `on_canvas_moved()`), which would do what `_update_view_positions()` does. GagdetOverlay would be in charge of calling it (and for now, you could just do it after every draw). One of the main advantage would be that you could go away with this weird mix-up with `canvas` argument.

src/odemis/gui/comp/overlay/world.py
```diff
@@ -1231,46 +1314,350 @@ def draw(self, ctx, selected, shift=(0, 0), scale=1.0, canvas=None):
...
+    def _edit_text(self):
+        """ A dialog box pops up and the user can edit the text """
+        dlg = wx.TextEntryDialog(None, 'Enter the label', 'Text entry', value=self.text)
+
+        if dlg.ShowModal() == wx.ID_OK:
+            self.text = dlg.GetValue()
+        else:
+            logging.debug("Dialog cancelled")
+        dlg.Destroy()
+        self.calc_edges()
```
pieleric: I don't understand how this works. The `.text` has just been changed, but not yet `._label.text`, so `_label.render_pos` will be incorrect.
My approach to solve this would be:
* do not have a separate `.text`, but instead always directly set `._label.text` (or `.text` would be a short-cut to  `._label.text`).
* Change `Label`to have `render_pos` always correct. One optimised way would be to provide it via a getter, which caches the value if it's already computed.

src/odemis/gui/comp/overlay/world.py
```diff
@@ -1148,16 +1076,171 @@ def _points_to_rect(left, top, right, bottom):
...
+        pass
+
+
+NONE_RULER_MODE = 0
+MOVE_RULER_MODE = 1
+EDIT_START_RULER_MODE = 2
+EDIT_END_RULER_MODE = 3
+
+
+class Ruler(GenericGadgetLine):
```
pieleric: Let's call this class `RulerGadget` and `LabelTool` -> `LabelGadget`. This way, everything connected to the `GadgetOverlay` has the same keyword.

src/odemis/gui/comp/overlay/world.py
```diff
@@ -1231,46 +1314,350 @@ def draw(self, ctx, selected, shift=(0, 0), scale=1.0, canvas=None):
...
+    and also handle the mouse interaction when dragging/moving the label.
+    """
+
+    def __init__(self, cnvs, p_start_pos, p_end_pos):
+        super(LabelTool, self).__init__(cnvs, p_start_pos, p_end_pos)
+
+        self._label = Label(
+            "",
+            pos=(0, 0),
+            font_size=12,
```
pieleric: Put 14, as this will be the value for (most of) the rest of the time.

src/odemis/gui/comp/overlay/world.py
```diff
@@ -1231,46 +1314,350 @@ def draw(self, ctx, selected, shift=(0, 0), scale=1.0, canvas=None):
...
+        # unit vector for buffer (pixel) coordinates
+        dpx, dpy = b_end[0] - b_start[0], b_end[1] - b_start[1]
+
+        phi = math.atan2(dx, dy) % (2 * math.pi)  # phi angle  of the label line in radians
+
+        # Find the label line length by calculating the Euclidean distance
+        pixel_length = math.hypot(dpx, dpy)  # label lime length in pixels
+
+        # text is placed horizontal
+        self._label.deg = math.degrees(math.pi)
```
pieleric: If it's a constant, just move it to the init.

src/odemis/gui/comp/overlay/world.py
```diff
@@ -1231,46 +1314,350 @@ def draw(self, ctx, selected, shift=(0, 0), scale=1.0, canvas=None):
...
+
+        # Draws a black background for the ruler
+        ctx.set_line_width(2)
+        ctx.set_source_rgba(0, 0, 0, 0.5)
+        ctx.move_to(*b_start)
+        ctx.line_to(*b_end)
+
+        # Ruler of 1 pixel width. Highlight the selected ruler and make it slightly thicker (2 pixels)
```
pieleric: That comment is not very clear. I'd just say "Make the line thicker and a different colour if it's selected"

src/odemis/gui/comp/overlay/world.py
```diff
@@ -1231,46 +1314,350 @@ def draw(self, ctx, selected, shift=(0, 0), scale=1.0, canvas=None):
...
+
+            ctx.set_line_width(0.5)
+            ctx.set_dash([])
+
+            ctx.set_source_rgba(0.6, 1, 0.6, 1)
+            ctx.rectangle(*text_rect)
+            ctx.stroke()
+
+    def draw(self, ctx, selected, shift=(0, 0), scale=1.0, canvas=None, font_size=None):
+        """ Draw a ruler and display the size in meters next to it. If the ruler is selected,
```
pieleric: That docstring should be adjusted to the "Label".

src/odemis/gui/comp/overlay/world.py
```diff
@@ -833,48 +835,61 @@ def deactivate(self):
...
+        """ Stop dragging the tool """
+        pass
+
+    def get_hover(self, vpos):
+        """ Check if the given position is on/near a selection edge or inside the selection of a tool.
+        It returns a "gui.HOVER_*" """
+        return gui.HOVER_NONE
+
+    def draw(self, ctx, selected, shift=(0, 0), scale=1.0, canvas=None, font_size=None):
+        pass
```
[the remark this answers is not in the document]
VictoriaDel: Most of draw methods in world.py have no docstring. I explained in detail what draw does in Ruler and LabelTool classes. I will add a generic comment

src/odemis/gui/comp/overlay/world.py
```diff
@@ -833,48 +835,61 @@ def deactivate(self):
...
-    and also handle the mouse interaction when dragging/moving the ruler.
+    This abstract GadgetToolInterface class forms the base for a series of classes that
+    refer to gadgets tools and their functionality.
     """
 
+    def __init__(self, cnvs):
+        """ Args: (cnvs) canvas passed by the GadgetOverlay and used to draw the gadgets """
+        self.cnvs = cnvs
+
+    def start_dragging(self, drag, vpos, tool_mode):
```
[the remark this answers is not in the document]
VictoriaDel: Thanks for letting me know. In abstract class Overlay the methods are not marked as abstract.

src/odemis/gui/comp/overlay/world.py
```diff
@@ -1231,46 +1314,350 @@ def draw(self, ctx, selected, shift=(0, 0), scale=1.0, canvas=None):
...
+            vx, vy = vpos
+
+            if self._edges:
+                vx, vy = vpos
+
+                if (
+                        self._edges["t_l"] < vx < self._edges["t_r"] and
+                        self._edges["t_t"] < vy < self._edges["t_b"]
+                ):
+                    return gui.HOVER_TEXT
```
Kleijwegt: add an extra tab to align with the condition block above

src/odemis/gui/comp/overlay/world.py
```diff
@@ -1231,46 +1314,350 @@ def draw(self, ctx, selected, shift=(0, 0), scale=1.0, canvas=None):
...
+                return gui.HOVER_NONE
+
+            # if position inside inner box
+            if (
+                    self._edges["i_l"] < vx < self._edges["i_r"] and
+                    self._edges["i_t"] < vy < self._edges["i_b"]
+            ):
+                dist = util.perpendicular_distance(self.v_start_pos, self.v_end_pos, vpos)
+                if dist < self.HOVER_MARGIN:
+                     return gui.HOVER_LINE
```
Kleijwegt: add an extra tab to align with the condition block above

src/odemis/gui/comp/overlay/world.py
```diff
@@ -1231,46 +1314,350 @@ def draw(self, ctx, selected, shift=(0, 0), scale=1.0, canvas=None):
...
+        self._update_view_positions(b_start, b_end)
+        self.debug_edges(ctx)
+
+
+MODE_CREATE_LABEL = 1
+MODE_CREATE_RULER = 2
+MODE_SHOW_TOOLS = 3
+
+
+class GadgetOverlay(WorldOverlay):
```
Kleijwegt: Gadget overlay, love it. The more gadgets, the better!

src/odemis/gui/main_xrc.py
```diff
@@ -3935,7 +3935,7 @@ def __init_resources():
             </object>
             <object class="sizeritem">
               <object class="ToolBar" name="sparc_acq_toolbar">
```
pieleric: I'm not sure it makes sense to have it in the SPARC acquisition, as these labels will not be stored in the acquisition. And if it's on the SPARC, we probably should also have it on the SECOM. Let's discuss it with Toon or Sangeetha.

src/odemis/gui/comp/overlay/world.py
```diff
@@ -1147,27 +1109,195 @@ def _points_to_rect(left, top, right, bottom):
...
+        Draw the line to given context
+        Args:
+            ctx: cairo context to draw on
+            selected: if the line is selected, it gets highlighted and thicker
+            canvas: canvas on which the tools are drawn. In case of print-ready export a fake canvas is passed and the
+            gadget overlay draws on it.
+            font_size: fontsize is given in case of print-ready export
+        """
+        pass
+
```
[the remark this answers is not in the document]
VictoriaDel: This class inherits from the abstract class GadgetToolInterface. I just deleted the above methods. ( Do you mean to mark it by using the "with_metaclass(ABCMeta, object)"??). Please check my latest version and let me know.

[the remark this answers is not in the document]
pieleric: It's good to remove these methods. You should *also* mark the class as "abstract", because it wouldn't work properly if you were directly creating an instance of it. So, mark with with `with_metaclass(ABCMeta, GadgetToolInterface)`.

pieleric:
Please add also the updated `images/tool_buttons.xcf` file to the commit.

src/odemis/gui/cont/tools.py
```diff
@@ -258,8 +276,21 @@ def _add_button(self, cls, img_prefix, tooltip=None):
...
-        self.btn_sizer.Add(btn, border=b, flag=f)
-        self.btn_panel.Layout()
+        if self.orientation == (wx.VERTICAL | wx.HORIZONTAL):
+            if self.tool == TOOL_RULER:
+                self.btn_sizer.Add(btn, (0, 0), border=b, flag=wx.BOTTOM | wx.ALIGN_CENTRE_HORIZONTAL)
+            elif self.tool == TOOL_LABEL:
+                self.btn_sizer.Add(btn, (1, 0), border=b, flag=wx.BOTTOM | wx.ALIGN_CENTRE_HORIZONTAL)
+            elif self.tool == TOOL_POINT:
+                self.btn_sizer.Add(btn, (0, 1), border=b, flag=wx.BOTTOM | wx.ALIGN_CENTRE_HORIZONTAL | wx.LEFT)
+            elif self.tool == TOOL_LINE:
```
pieleric: Is the `wx.LEFT` flag really necessary? It'd be simpler it it wasn't needed. If you need margin, you might be able to add it to the sizer itself.

VictoriaDel:
Is it necessary? In this file, there are tools missing like the line spectra. I didn't save the label tool, I will create it again.

pieleric:
Yes, the idea is that *all* the tools are there. The fact that some are missing is not a good reason to forgetting even more of them ;-)

src/odemis/gui/cont/tools.py
```diff
@@ -258,8 +276,21 @@ def _add_button(self, cls, img_prefix, tooltip=None):
...
+                self.btn_sizer.Add(btn, (1, 1), border=b, flag=wx.BOTTOM | wx.ALIGN_CENTRE_HORIZONTAL | wx.LEFT)
+            elif self.tool == TOOL_ACT_ZOOM_FIT:
+                self.btn_sizer.Add(btn, (2, 1), border=b, flag=wx.BOTTOM | wx.ALIGN_CENTRE_HORIZONTAL | wx.LEFT)
```
pieleric: There should be a else, which handles properly unknown tool ID. => Just pick a random place available.
I think you could try something like this:
```
# Ideal position for the known tools
pos = {TOOL_RULER: (0,0), TOOL_LABEL:(1,0)...}.get(self.tool)
# Unknown tool, or position already used => pick the first position available
if not pos or self.btn_sizer.FindItemAtPosition(pos):
   for j in range(8): # max 2x8
      for i in range(2):
        if not self.btn_sizer.FindItemAtPosition((i,j)):
           pos = (i,j)
           break
   else:
       raise ValueError("No more space in toolbar")
self.btn_sizer.Add(btn, pos, border=b, flag=wx.BOTTOM | wx.ALIGN_CENTRE_HORIZONTAL)
```

src/odemis/gui/comp/overlay/world.py
```diff
@@ -1147,27 +1098,164 @@ def _points_to_rect(left, top, right, bottom):
...
+            self.v_end_pos = Vec(self.cnvs.buffer_to_view(b_end))
+            self._calc_edges()
+            self.last_shiftscale = shiftscale
+
+    def draw(self, ctx, selected, canvas=None, font_size=None):
+        """ Draw a ruler and display the size in meters next to it. If the ruler is selected,
         highlight it and make it thicker. A canvas is passed in case of print-ready export
-        and the ruler overlay draws on it"""
+        and the gadget overlay draws on it """
+        super(RulerGadget, self).draw(ctx, selected, canvas=None, font_size=None)
```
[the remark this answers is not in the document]
pieleric: Sorry, but you need to also change `canvas=canvas`.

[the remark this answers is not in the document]
VictoriaDel: Sorry, it's my bad. I pushed all the changes.

PR #1358, opened 2020-02-24
acq-stitching: bugfix GlobalShiftRegistrar

philipwinklerdelmic:


src/odemis/acq/stitching/_registrar.py
```diff
@@ -746,4 +746,8 @@ def _assemble_mosaic(self):
             if i not in positions:
                 positions[i] = [None, None]
 
-        return numpy.array(list(positions.values())).reshape([num_rows, num_cols, 2])
+        # Sort positions by key
+        positions = numpy.array(sorted(positions.items(), key=lambda x: x[0]))
```
pieleric: You don't need to store it as a numpy array here.

philipwinklerdelmic: indeed, it was actually also failing on python3 (it didn't like converting a mulitdimensional list to a numpy.array)

philipwinklerdelmic:
updated

Kleijwegt:
> Actually dictionaries have never been supposed to be sorted. I don't know why it worked....

Maybe it is a good idea to do a software knowledge transfer meeting on dictionaries, sorting and lambdas?

Sabrina85:
@Kornee We have a software presentation on dictionaries in the dropbox folder. Btw also one on the stitching from Philip in case you haven't seen it yet.

src/odemis/acq/stitching/_registrar.py
```diff
@@ -746,4 +746,8 @@ def _assemble_mosaic(self):
             if i not in positions:
                 positions[i] = [None, None]
 
-        return numpy.array(list(positions.values())).reshape([num_rows, num_cols, 2])
+        # Sort positions by key
```
pieleric: I think you could entirely do without the dict, and that it would clarify what's happening. `positions` should be there for each int 0->len(tree) -1. So it's actually an array.
Also, you have this special code to handle missing positions, but I think that the caller doesn't handle `[None, None] ` afterwards anyway. So I'd suggest to start with the "final array", initialised at [0,0], and update it on-the-fly as you're following the graph.
```
positions = numpy.zeros((num_rows, num_cols, 2)) # start with no shift for each tile
positions_flat = positions.reshape((num_rows * num_cols, 2))  # the tiles in the "idx" notation
:
: the while loop as before, using positions_flat
: 
# No special handling of the missing tiles or dict sorting!
return positions
```

philipwinklerdelmic: I added a new commit for this, not sure though if it's making it much better. It seems that I used the dictionary, so I can easily test if a position has already been considered. Now, I have to keep an additional list for this.

philipwinklerdelmic: To make it even better, we could consider cleaning up adj_positions a bit earlier.

philipwinklerdelmic: But maybe it's good enough already, I don't really see the need to improve it any further.

src/odemis/acq/stitching/_registrar.py
```diff
@@ -726,28 +726,21 @@ def _assemble_mosaic(self):
                     adj.append([j, shifts[j][i]])
             adj_list[i] = adj
 
-        positions = {0: (0, 0)}
+        positions = numpy.zeros((num_rows, num_cols, 2))  # start with no shift for each tile
+        positions_flat = positions.reshape((num_rows * num_cols, 2))  # the tiles in the "idx" notation
         idx_queue = deque([0])  # start with index 0
+        upd_positions = []  # list of positions that have already been updated
```
pieleric: I'd suggest to name it something like `idx_done`, to make it a little bit more symetrical with `idx_queue`. Also, make it a set, as it's a little faster, and conveys the fact it the order doesn't matter.

philipwinklerdelmic: Good idea! I've updated the commit.

PR #1359, opened 2020-02-25
readme: use markdown style

philipwinklerdelmic:


philipwinklerdelmic:
I just created a readme file for the jolt software and realized how ugly the readme for odemis looks.
If we use the markdown style, github will display it much more beautifully
Screenshot from 2020-02-25 15-11-09

philipwinklerdelmic:
I also changed the python version to 3.5 and the ubuntu version to 18.04 instead of 12.04

README.md
```diff
@@ -0,0 +1,73 @@
+<p align="right">
+  <img  src="./src/odemis/gui/img/logo_h30.png">
+</p>
+
+# Odemis
+*Open Delmic Microscope Software*
+
+## Requirements
+* Linux (tested on Ubuntu 16.04 and 18.04 x86 32-bits and 64-bits)
+* Python (v3.5)
```
Kleijwegt: Isn't python 3.6 the preferred version since this is the standard in Ubuntu 18?

pieleric: Well, we still support Python 3.5 (and 2.7!). So that's good enough. I guess that ideally it would indicate all the Python versions supported.
This readme file could be a lot further improved, but that's already a good step :-)

PR #1361, opened 2020-02-28
acq align delphi_man_calib: add --keep-loaded argument to avoid loading/unloading

pieleric:
The overview image is only used for the stage alignment, and still only
really useful for the very first time. In many cases, this causes an
extra 2 minutes to the calibration.

=> allow to not do that, at least on the command line

### DIFF
--- a/src/odemis/acq/align/delphi_man_calib.py
+++ b/src/odemis/acq/align/delphi_man_calib.py
@@ -182,7 +182,7 @@ class ArrowFocus(object):
             self._focus_thread.join(10)
 
 
-def man_calib(logpath):
+def man_calib(logpath, keep_loaded=False):
     escan = None
     detector = None
     ccd = None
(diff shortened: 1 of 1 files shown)


src/odemis/acq/align/delphi_man_calib.py
```diff
@@ -262,45 +263,47 @@ def man_calib(logpath):
                   "                    rotation: " + str(irot) + "\n"
                   "                    scale-xy: " + str(iscale_xy) + "\n"
                   "                    shear: " + str(ishear))
-        print_col(ANSI_YELLOW,
+        print_col(ANSI_YELLOW, -
```
tepals: what does the - do here?

pieleric: It makes sure that Python cannot parse this file :-P

Thanks for catching this. Just pushed a fixed version.

tepals: haha here I was, thinking it was some fancy trick :)

PR #1362, opened 2020-02-28
delphi: fix minor bugs

pieleric:


src/odemis/acq/stream/_live.py
```diff
@@ -475,6 +475,10 @@ def _DoPrepare(self):
                 self._emitter.shift.value = (0, 0)
                 shift = FindEbeamCenter(self._ccd, self._detector, self._emitter)
                 logging.debug("Spot shift is %s m,m", shift)
+                shift_clipped = self._emitter.shift.clip(shift)
+                if shift_clipped != shift:
+                    shift = shift_clipped
+                    logging.info("Limiting spot shift to %s m,m due to hardware constraints", shift)
```
Kleijwegt: What is the added value of creating the variable shift_clipped? I think it is still readable if you write:
```
if shift_clipped != self._emitter.shift.clip(shift)
   shift = self._emitter.shift.clip(shift)
   logging.info("Limiting spot shift to %s m,m due to hardware constraints", shift)
```

Especially since the logging message is almost a comment explaining what is happening

pieleric: The goal of using `shift_clipped` is to only call `.clip()` once. Admittedly, it's not very costly, but that's a remote call, so I like to be a little careful.
That said, this `if` is quite unlikely, so the extra (small) cost is very limited, and the advantage of your suggestion would be to keep the code a tiny be more readable.

pieleric: I've just tried it. Looking at the new code, I just too much have a feeling "why do you have to call that function twice!?".
So I'd rather keep the code as-is.

PR #1363, opened 2020-02-28
acq path: don't queue OPM setPath() calls

pieleric:
Sometimes, for instance when changing quickly between chamber and
alignment tab, setPath() is called several times in a row.
Currently, every request is full-field, which can end up being very
slow. However, we only care about the latest requested position.

=> Discard move requests which haven't started yet.

We could be even more clever, and immediately stop the current moves,
but that's quite a lot more complicated, espcially because that means
all the callers need to handle CancelledError.

### DIFF
--- a/src/odemis/acq/path.py
+++ b/src/odemis/acq/path.py
@@ -31,6 +31,7 @@ from odemis import model, util
 from odemis.acq import stream
 import time
 from odemis.util import TimeoutError
+import queue
 
 GRATING_NOT_MIRROR = ("NOTMIRROR",)  # A tuple, so that no grating position can be like this
 
@@ -259,6 +260,28 @@ SECOM_MODES = {
 ALIGN_MODES = {'mirror-align', 'chamber-view', 'fiber-align', 'streak-align', 'spec-focus', 'spec-fiber-focus', 'streak-focus'}
 
 
+class OneTaskExecutor(ThreadPoolExecutor):
+    """
+    Only one task at a time is executed. If several tasks are queued, only the
+    last one is executed.
+    """
+
+    def __init__(self, thread_name_prefix=''):
+        super(OneTaskExecutor, self).__init__(max_workers=1, thread_name_prefix=thread_name_prefix)
+
+    def submit(self, fn, *args, **kwargs):
+        # Cancels all the previous tasks which didn't start yet
+        # Note: in practice, there should be only one task at most queued.
+        while True:
+            try:
+                self._work_queue.get(block=False)
+                logging.debug("Canceling 1 task before starting the next one")
+            except queue.Empty:
+                break
+
+        return ThreadPoolExecutor.submit(self, fn, *args, **kwargs)
+
+
 # TODO: Could be moved to util
 def affectsGraph(microscope):
     """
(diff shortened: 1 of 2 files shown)


src/odemis/acq/test/path_test.py
```diff
@@ -476,6 +477,24 @@ def test_wrong_mode(self):
...
+        Test changing path multiple times without waiting for it to be complete
+        """
+        tstart = time.time()
+        self.optmngr.setPath("cli")
+
+        # All these ones should get discarded
+        for i in range(5):
+            self.optmngr.setPath("ar")
+            self.optmngr.setPath("mirror-align")
+            self.optmngr.setPath("cli")
```
Sabrina85: These ones are not started as you do not call `.result()` and therefore only put on the queue?

src/odemis/acq/path.py
```diff
@@ -259,6 +260,28 @@
...
+class OneTaskExecutor(ThreadPoolExecutor):
+    """
+    Only one task at a time is executed. If several tasks are queued, only the
+    last one is executed.
+    """
+
+    def __init__(self, thread_name_prefix=''):
+        super(OneTaskExecutor, self).__init__(max_workers=1, thread_name_prefix=thread_name_prefix)
+
+    def submit(self, fn, *args, **kwargs):
```
Sabrina85: what is param fn? doc string?i

src/odemis/acq/path.py
```diff
@@ -259,6 +260,28 @@
...
+        # Cancels all the previous tasks which didn't start yet
+        # Note: in practice, there should be only one task at most queued.
+        while True:
+            try:
+                self._work_queue.get(block=False)
+                logging.debug("Canceling 1 task before starting the next one")
+            except queue.Empty:
+                break
```
Sabrina85: You empty the queue correct? How do you then execute the last one on the queue?

src/odemis/acq/path.py
```diff
@@ -259,6 +260,28 @@
...
 
 
+class OneTaskExecutor(ThreadPoolExecutor):
+    """
+    Only one task at a time is executed. If several tasks are queued, only the
+    last one is executed.
+    """
+
+    def __init__(self, thread_name_prefix=''):
+        super(OneTaskExecutor, self).__init__(max_workers=1, thread_name_prefix=thread_name_prefix)
```
Sabrina85: doc string?

src/odemis/acq/test/path_test.py
pieleric: Yes. Because there is already a setPath() going on, and that we immediately ask for more, they end up being queued (and now, immediately discarded).

src/odemis/acq/path.py
pieleric: This is overriding the parent method. So, same signature. That method actually has also no docstring, because it also overrides the method from its parents, but eventually you can find the right docstring. Typically you don't want to duplicate the docstring when overriding methods (because duplication can easily lead to out of date information), unless you change the signature.

src/odemis/acq/path.py
pieleric: Yes, first empty the queue (so only the task which is already running is left), and then the call to the super() will add this new function to the work_queue.
I've added a comment now (and change to user `super()`).

src/odemis/acq/path.py
pieleric: It's just the same as the overriding method (- `max_workers`).

src/odemis/acq/path.py
Sabrina85: ok

PR #1364, opened 2020-03-06
Made axes local VA's so that a different filter can be selected per CL stream

amuskens:
Preliminary working version - still didn't update the new test cases.

Please review - since I am not sure if I did everything correctly. 

Issues
- Filter does not immediately select when the stream is created. Could be related to the callback function for the setting entry and combobox.


src/odemis/acq/stream/_base.py
```diff
@@ -191,6 +193,8 @@ def __init__(self, name, detector, dataflow, emitter, focuser=None, opm=None,
         self._det_vas = self._duplicateVAs(detector, "det", detvas or set())
         self._emt_vas = self._duplicateVAs(emitter, "emt", emtvas or set())
 
+        self._init_axis_vas(axis_map)
```
pieleric: It'd fit better the code if it could look like:
`self._axis_vas = self._duplicateAxes(...).`

Also, IMHO, it's nice if the VAs are available directly as attributes of the stream. So, for each axis, also provide the VA as an attribute `.axisNameOfTheAxis`.

src/odemis/acq/stream/_base.py
```diff
@@ -300,6 +320,10 @@ def det_vas(self):
     def emt_vas(self):
         return self._emt_vas
```
Kleijwegt: This function seems like a waste of code/characters, calling the function cost more than the actual implementation of the function?

src/odemis/model/_futures.py
```diff
@@ -364,6 +364,105 @@ def set_exception(self, exception):
...
 
 
+class CompositeFuture(futures.Future):
+    """
+    A future that checks the running state of multiple futures and joins them
+    """
+    MAX_FUTURES = 99
+
+    def __init__(self, list_of_futures):
+        # list of futures (list): a list of futures that will all be executed simultaneously.
```
Kleijwegt: use docstring instead of # comment

src/odemis/acq/stream/_base.py
pieleric: This part is already in the code (so it's not Anders' fault).
The goal of such "property" is that it indicates the attribute is read-only. A call in Python is not very costly, but surely, if you need an extremely fast code path, this would need an optimization (or to rewrite everything in cython ;-) ).

src/odemis/acq/stream/_base.py
```diff
@@ -561,6 +636,65 @@ def _getDetectorVA(self, vaname):
...
+
+    def _update_linked_axis(self, ax_name, pos):
+        """ Update the value of a linked hardware axis VA
+            when the stream is active
+        """
+        if not self.is_active.value:
+            return
+        if hasattr(self, "_axis_map") and self._axis_map:
+            try:
+                act = self._axis_map[ax_name][1]
```
pieleric: You also need the "actuator axis name", as `ax_name` is for the stream.
=> `real_ax_name, act = self._axis_map[ax_name]`

src/odemis/acq/stream/_base.py
```diff
@@ -561,6 +636,65 @@ def _getDetectorVA(self, vaname):
...
+                logging.info("Moving actuator %s to position %s.", act.name, pos)
+                f = act.moveAbs({ax_name: pos})
+                # f.result()
```
pieleric: Please add a comment here why you'd want this here: this would allow the caller (ie, the GUI) to know when the axis is done moving. However, this would require the GUI to be adjusted so that it sets the value in a separate thread.

src/odemis/acq/stream/_base.py
```diff
@@ -561,6 +636,65 @@ def _getDetectorVA(self, vaname):
...
+        """ Update the value of a linked hardware axis VA
+            when the stream is active
+        """
+        if not self.is_active.value:
+            return
+        if hasattr(self, "_axis_map") and self._axis_map:
+            try:
+                act = self._axis_map[ax_name][1]
+                logging.info("Moving actuator %s to position %s.", act.name, pos)
+                f = act.moveAbs({ax_name: pos})
```
pieleric: Please add a callback to the future, to just check the result of the move and log an error in case of error/exception. Otherwise failing moves will be hidden. (See polarizer code for an example).

src/odemis/acq/stream/_base.py
```diff
@@ -561,6 +636,65 @@ def _getDetectorVA(self, vaname):
...
+    def _update_linked_axis(self, ax_name, pos):
+        """ Update the value of a linked hardware axis VA
+            when the stream is active
+        """
+        if not self.is_active.value:
+            return
+        if hasattr(self, "_axis_map") and self._axis_map:
+            try:
+                act = self._axis_map[ax_name][1]
+                logging.info("Moving actuator %s to position %s.", act.name, pos)
```
pieleric: Please also log the `real_ax_name`.

src/odemis/acq/stream/_base.py
```diff
@@ -561,6 +636,65 @@ def _getDetectorVA(self, vaname):
...
+                f = act.moveAbs({ax_name: pos})
+                # f.result()
+            except Exception:
+                logging.exception("Failed to move axis.")
```
pieleric: Please log the actuator and axis name.

src/odemis/acq/stream/_base.py
```diff
@@ -561,6 +636,65 @@ def _getDetectorVA(self, vaname):
...
+        return pos
+
+    def _unlinkHwAxes(self):
+        """
+        Unlink the axes to the hardware components
+        """
+        if hasattr(self, "_axis_map") and self._axis_map is not None:
+            for ax_name in self._axis_map:
+                local_pos_va = self._axis_vas[ax_name]
+                local_pos_va.unsubscribe(self._update_linked_axis)
```
pieleric: That doesn't work: it does nothing because `self._update_linked_axis` is not a subscriber (only some partials made of it). You need to unsubscribe the `self._axisvaupdaters[ax_name]`.

src/odemis/acq/stream/_helper.py
```diff
@@ -1208,7 +1208,6 @@ def _onNewData(self, dataflow, data):
         # And update the stream status if protection was triggered
         super(CLSettingsStream, self)._onNewData(dataflow, data)
 
-
```
pieleric: Please don't touch this file, as you're not modifying it.

src/odemis/acq/test/stream_test.py
```diff
@@ -3824,9 +3833,14 @@ def test_ar_ss(self):
 
     def test_cl_ss(self):
         """ Test CLSettingsStream """
+
+        # create axes
+        axes = {"band": ("band", self.filter)}
```
pieleric: Please call this one "filter", to make it a little bit harder.

src/odemis/acq/test/stream_test.py
```diff
@@ -3837,11 +3851,23 @@ def test_cl_ss(self):
 
         cls.emtDwellTime.value = 10e-6 # s
         cls.pixelSize.value *= 10
+        cls.axis_band.value = 0
```
pieleric: Please don't hard code the values for the band. They are enumerated, and might change. It's better to just pick 2 "random" values, from the choices. Something like:
`b0, b1 = list(cls.axisFilter.choices)[:2]`

src/odemis/gui/conf/util.py
```diff
@@ -929,6 +929,165 @@ def cb_get(ctrl=value_ctrl, va=va, u=unit):
     return setting_entry
 
 
+def create_local_axis_entry(container, name, va, comp, conf=None):
```
pieleric: That's a lot of new code, almost identical to the other ones. I *really* would like to avoid this. It should be possible to use `create_setting_entry()`, with `hw_comp=None`. If it doesn't work out-of-the-box, only minor tweaks should be needed (and with `hw_comp=None`, it's easy to distinguish).

src/odemis/gui/cont/streams.py
```diff
@@ -148,6 +150,9 @@ def __init__(self, stream_bar, stream, tab_data_model, show_panel=True, view=Non
 
         self._add_stream_setting_controls()
 
+        self._axes = None
```
pieleric: It shouldn't be needed, just write `self.stream.axis_vas` when you need it.

src/odemis/gui/cont/streams.py
```diff
@@ -331,6 +336,40 @@ def _add_stream_setting_controls(self):
...
             self.add_setting_entry(vaname, va, hw_comp=None, conf=conf)
 
+    def _add_axis_controls(self):
+        """
+        Add controls for the axes that are connected to the stream
+        """
+        
+        self._axes = self.stream.axis_map
+
+        if self._axes is not None:
```
pieleric: It shouldn't need to use `axis_map`, just use `axis_vas`, and add each VA as an entry. Done.

src/odemis/gui/cont/streams.py
```diff
@@ -331,6 +336,38 @@ def _add_stream_setting_controls(self):
...
+                    # If the axis is configured with local axis VA's, use a setting entry
+                    # instead of an axis entry.
+                    if conf is None:
+                        conf = {"axis_va": True}
+                    else:
+                        conf["axis_va"] = True
+
+                    self.add_setting_entry(axisname, self.stream.axis_vas[axisname], comp, conf)
+                else:
+                    self.add_axis_entry(axisname, comp, conf)
```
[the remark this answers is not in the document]
pieleric: => Drop this line.

src/odemis/gui/cont/streams.py
```diff
@@ -331,6 +336,40 @@ def _add_stream_setting_controls(self):
...
+            for axisname in axes_names:
+                comp = self._axes[axisname][1]
+                if comp is None:
+                    logging.debug("Skipping axis %s for non existent component",
+                                  axisname)
+                    continue
+                if axisname not in comp.axes or self._axes[axisname][0] not in comp.axes:
+                    logging.debug("Skipping non existent axis %s on component %s",
+                                  axisname, comp.name)
+                    continue
```
pieleric: This shouldn't be checked here. If the axis is in axis_vas, then it exists.

src/odemis/gui/cont/streams.py
```diff
@@ -331,6 +336,40 @@ def _add_stream_setting_controls(self):
...
+                conf = self._stream_config.get(axisname)
+
+                if hasattr(self.stream, "_axis_vas"):
+                    # If the axis is configured with local axis VA's, use a setting entry
+                    # instead of an axis entry.
+                    if conf is None:
+                        conf = {"axis_va": True}
+                    else:
+                        conf["axis_va"] = True
```
pieleric: No need to pass such flag.

src/odemis/gui/cont/streams.py
```diff
@@ -2554,29 +2610,12 @@ def _addRepStream(self, stream, mdstream, axes, **kwargs):
         if model.hasVA(stream, "useScanStage"):
             stream.useScanStage.value = self._tab_data_model.useScanStage.value
 
-        stream_cont = self._add_stream(stream, add_to_view=True, **kwargs)
+        stream_cont = self._add_stream(stream, add_to_view=True, ** kwargs)
```
pieleric: Don't change that.

src/odemis/gui/cont/streams.py
```diff
@@ -2621,6 +2660,15 @@ def addCLIntensity(self):
...
 
         main_data = self._main_data_model
+
+        # Need to pick the right filter wheel (if there is one)
+        axes = {}
+        for fw in (main_data.cl_filter, main_data.light_filter, main_data.tc_od_filter, main_data.tc_filter):
+            if fw is None:
+                continue
+            if main_data.cld.name in fw.affects.value:
+                axes["band"] = ("band", fw)
```
pieleric: `axes["band"]` -> `axes["filter"]`
(for all the streams)
Then you can update gui.conf.data, and drop all `label` line in the config like:
```
            ("band", {  # from filter
                "label": "Filter",
            }),
```

src/odemis/gui/cont/streams.py
```diff
@@ -2678,39 +2719,39 @@ def addSpectrum(self, name=None, detector=None):
...
+                "grating": ("grating", spg),
+                "slit-in": ("slit-in", spg),
+               }
+
+        # Also add light filter for the spectrum stream if it affects the detector
+        for fw in (main_data.cl_filter, main_data.light_filter):
+            if fw is None:
+                continue
+            if detector.name in fw.affects.value:
+                axes["band"] = ("band", fw)
```
pieleric: Also use the name `filter` here.

src/odemis/gui/cont/streams.py
```diff
@@ -2678,39 +2719,39 @@ def addSpectrum(self, name=None, detector=None):
         logging.debug("Adding spectrum stream for %s", detector.name)
 
         spg = self._getAffectingSpectrograph(detector)
+
+        # band should be set to the pass-through by default
+        axes = {"wavelength": ("wavelength", spg),
+                "grating": ("grating", spg),
+                "slit-in": ("slit-in", spg),
+               }
+
```
pieleric: You need to drop these axes if there are not available on the hardware, here, before creating the stream. 
Maybe you could put this functionality into a common method named like `._filter_axes(axes)`. It would take the `axes` parameter passed to the stream, and return the same dict, but without the invalid entries (ie, component is None or component doesn't have this axis).

src/odemis/acq/stream/_base.py
```diff
@@ -561,6 +636,65 @@ def _getDetectorVA(self, vaname):
                 raise AttributeError(u"Detector has not VA %s" % (vaname,))
             return hwva
 
+    def _linkHwAxes(self):
```
pieleric: BTW, there are 2 streams which already have this method. You need to check that they don't break, now they are overloading it (might be just a matter of calling also the `super()`).
```
> grep -Ir linkHwAxes src/odemis/ plugins/
plugins/la_spec.py:            # self._linkHwAxes()
plugins/la_spec.py:            self._unlinkHwAxes()
plugins/la_spec.py:        self._linkHwAxes()
plugins/la_spec.py:    def _linkHwAxes(self, _=None):
plugins/la_spec.py:    def _unlinkHwAxes(self):
plugins/la_spec.py:        self._sccd._linkHwAxes()
plugins/la_spec.py:            self._sccd._unlinkHwAxes()
src/odemis/acq/stream/_helper.py:            self._linkHwAxes()
src/odemis/acq/stream/_helper.py:            self._unlinkHwAxes()
src/odemis/acq/stream/_helper.py:    def _linkHwAxes(self):
src/odemis/acq/stream/_helper.py:    def _unlinkHwAxes(self):
```

src/odemis/gui/cont/streams.py
amuskens: You mean universal for all of the times you add axes?

src/odemis/gui/conf/util.py
amuskens: I tried this at first, but there are a lot of differences in the axis that are slightly different, so the result is merging the "create_axis_entry" and "create_setting_entry" if this is ok. The very large function before became rather confusing, and I was worried I'd break something.

src/odemis/acq/stream/_base.py
amuskens: Seems like the fix is simple for this. Added it.

src/odemis/gui/conf/util.py
pieleric: Leave `create_axis_entry()` as-is. However, `create_local_axis_entry()` and `create_setting_entry()` should be merged. I'm pretty convinced that the result is just the current `create_setting_entry()`, with maybe a few tweaks, if I'm not mistaken.

src/odemis/gui/cont/streams.py
pieleric: All the time that the axes are added in the SPARCStreamController.

src/odemis/acq/stream/_base.py
```diff
@@ -404,6 +435,25 @@ def _va_sync_from_hw(self, lva, v):
         if lva._value != v:
             lva._value = v  # TODO: works with ListVA?
             lva.notify(v)
+            
+    # TODO: move to odemis.util ?
```
pieleric, line 439: No. => Drop this TODO.

src/odemis/acq/stream/_base.py
```diff
@@ -257,6 +263,27 @@ def __init__(self, name, detector, dataflow, emitter, focuser=None, opm=None,
...
+        axis_map (dict of axis_name -> Actuator): map of an axis name to an Actuator component
+        """
+        if not axis_map:
+            return {}
+
+        # Add axis position VA's to the list of hardware VA's
+        axis_vas = {}  # dict of axis_name to duplicated position VA
+        for va_name in axis_map:
+            axis_name = axis_map[va_name][0]
+            actuator = axis_map[va_name][1]
```
pieleric, line 279: Please write `for va_name, (axis_name, actuator) in axis_map.item():`

src/odemis/acq/stream/_base.py
```diff
@@ -561,6 +611,77 @@ def _getDetectorVA(self, vaname):
...
+        If local axis Vigilant Attributes's are specified, write the values of the local axis VA
+        to the real hardware
+        """
+
+        if hasattr(self, "axis_map") and self.axis_map:
+
+            logging.info("Moving linked HW axes")
+
+            moving_axes = []
+            for ax_name in self.axis_map:
```
pieleric, line 626: You should be able to simplify by writing: `for va_name, (axis_name, actuator) in self.axis_map.item():`

src/odemis/gui/cont/streams.py
```diff
@@ -137,6 +138,7 @@ def __init__(self, stream_bar, stream, tab_data_model, show_panel=True, view=Non
             self._display_metadata()
 
         # Add local hardware settings to the stream panel
+
```
pieleric, line 141: Drop this extra line.

src/odemis/gui/cont/streams.py
```diff
@@ -331,6 +335,20 @@ def _add_stream_setting_controls(self):
             conf = stream_config.get(vaname)
             self.add_setting_entry(vaname, va, hw_comp=None, conf=conf)
 
+    def _add_axis_controls(self):
+        """
+        Add controls for the axes that are connected to the stream
+        """
+        if hasattr(self.stream, "axis_map") and self.stream.axis_map:
+            # Add Axes (in same order as config)
+            axes_names = util.sorted_according_to(list(self.stream.axis_map.keys()), list(self._stream_config.keys()))
```
pieleric, line 344: This whole function should only use `axis_vas`, not `axis_map` and `_axis_vas`, which should both be private.

src/odemis/gui/cont/streams.py
```diff
@@ -331,6 +335,20 @@ def _add_stream_setting_controls(self):
...
+            for axisname in axes_names:
+                real_ax_name, comp = self.stream.axis_map[axisname]
```
pieleric, line 346: You don't need real_ax_name, just the va_name, and you don't need the comp => drop this line.

src/odemis/gui/cont/streams.py
```diff
@@ -348,6 +366,23 @@ def add_setting_entry(self, name, va, hw_comp, conf=None):
 
         return se
 
+    def add_local_axis_entry(self, name, va, hw_comp, conf=None):
+        """ Add a name/value pair to the settings panel.
+
+        :param name: (string): name of the value
+        :param va: (VigilantAttribute)
+        :param hw_comp: (Component): the component that contains this VigilantAttribute
```
pieleric, line 374: You don't need to know the component containing the axis. => drop it.

src/odemis/gui/cont/streams.py
```diff
@@ -331,6 +335,20 @@ def _add_stream_setting_controls(self):
...
+        Add controls for the axes that are connected to the stream
+        """
+        if hasattr(self.stream, "axis_map") and self.stream.axis_map:
+            # Add Axes (in same order as config)
+            axes_names = util.sorted_according_to(list(self.stream.axis_map.keys()), list(self._stream_config.keys()))
+            for axisname in axes_names:
+                real_ax_name, comp = self.stream.axis_map[axisname]
+                conf = self._stream_config.get(axisname)
+
+                if hasattr(self.stream, "_axis_vas"):
```
pieleric, line 349: You should assume it's there. => drop this `if`.

src/odemis/gui/cont/streams.py
```diff
@@ -331,6 +335,20 @@ def _add_stream_setting_controls(self):
             conf = stream_config.get(vaname)
             self.add_setting_entry(vaname, va, hw_comp=None, conf=conf)
 
+    def _add_axis_controls(self):
+        """
+        Add controls for the axes that are connected to the stream
+        """
+        if hasattr(self.stream, "axis_map") and self.stream.axis_map:
```
pieleric, line 342: You can revert this `if`, to avoid extra indentation:
```
if not hasattr(self.stream, "axis_vas"):
    return
```

src/odemis/gui/cont/streams.py
```diff
@@ -2621,6 +2639,15 @@ def addCLIntensity(self):
...
 
         main_data = self._main_data_model
+
+        # Need to pick the right filter wheel (if there is one)
+        axes = {}
+        for fw in (main_data.cl_filter, main_data.light_filter, main_data.tc_od_filter, main_data.tc_filter):
+            if fw is None:
+                continue
+            if main_data.cld.name in fw.affects.value:
+                axes["filter"] = ("band", fw)
```
pieleric, line 2649: Add a `break` here (was a bug before).

src/odemis/gui/cont/streams.py
amuskens: Doesn't it need the real names from axis_map?

amuskens: Woops - nvm.

src/odemis/gui/cont/streams.py
amuskens: So now.. the display reads "Filter" rather than "band" since I didn't use the real name. Is that ok?

src/odemis/gui/conf/util.py
amuskens: Maybe... although there are a lot of tweaks it seems for the axis. I will see if I can merge it again. I think i have to change some of the ways it handles arguments - there wouldn't be any settings entries with  hw_comp=None that don't use the axes?

amuskens: Ok - while testing - apparently there are cases where hw_comp is None that are not related to local axis (_add_hw_setting_controls, with CameraStream.emitter) so is there another way to determine when the SettingEntry should default to localAxis mode?

amuskens: Also the resolution VA

src/odemis/gui/cont/streams.py
pieleric: That's the whole point. In the GUI, we don't care of the name on the hardware. The important name is the name on the stream.

src/odemis/gui/conf/util.py
pieleric: Could you point me to where is it important for the code to know that it's a "local axis" instead of a standard VA?
I'd expect that at this point, the GUI doesn't care. It should all be hidden away by the stream. Eventually it's a "setting". Some settings are properties of the hardware, some involve moving an axis, but that doesn't matter.

src/odemis/acq/stream/_base.py
```diff
@@ -561,6 +608,76 @@ def _getDetectorVA(self, vaname):
...
+            logging.exception("Failed to move axis.")
+        return pos
+
+    def _unlinkHwAxes(self):
+        """
+        Unlink the axes to the hardware components
+        """
+        if hasattr(self, "axis_map") and self.axis_map is not None:
+            for ax_name in self.axis_map:
+                va = self._axis_vas[ax_name]
```
pieleric, line 677: You don't need axis_map for this. Just write:
`for va_name, va in self._axis_vas.items():`
This replaces the two lines.

src/odemis/acq/stream/_base.py
```diff
@@ -561,6 +608,76 @@ def _getDetectorVA(self, vaname):
...
+        """
+         Callback method, which checks that the move is actually finished.
+        :param f: (future)
+        """
+        try:
+            f.result()
+        except Exception:
+            logging.exception("Failed to move axis.")
+
+    def _update_linked_axis(self, ax_name, pos):
```
pieleric, line 654: Please rename `ax_name` -> `va_name`, to keep the same variable name as in `_linkHwAxes()`.

src/odemis/acq/stream/_base.py
```diff
@@ -561,6 +608,76 @@ def _getDetectorVA(self, vaname):
...
+            when the stream is active
+        """
+        if not self.is_active.value:
+            return
+        try:
+            act = self.axis_map[ax_name][1]
+            logging.info("Moving actuator %s to position %s.", act.name, pos)
+            f = act.moveAbs({ax_name: pos})
+            # TODO: Move this to a separate thread to make sure it doens't block the GUI
+            # it would be good to have the caller know when the axis is done moving.
```
pieleric, line 665: To clarify, I'd suggest to write something more like:
```
TODO: ideally, it would block, so that the the caller knows when the move is complete.
   However, this requires that the GUI calls this function is a separate thread.
```

src/odemis/acq/stream/_base.py
```diff
@@ -586,6 +703,8 @@ def _prepare(self):
         logging.debug(u"Preparing stream %s ...", self.name.value)
         # actually indicate that preparation has been triggered, don't wait for
         # it to be completed
+        self._linkHwAxes()
+
```
pieleric, line 707: No. You shouldn't call it twice. Either call it in `prepare()`, in which case, ideally it'd be part of the future which is returned. Or you call it in `_is_active_setter()`. That second option is probably a little bit less nice, but easier, as you don't need to combine the futures.

src/odemis/gui/conf/util.py
amuskens: I think it is just related to the way the axes are formatted

src/odemis/gui/conf/util.py
```diff
@@ -498,43 +498,52 @@ def format_choices(choices):
...
-        choices = axis_def.choices
-    except AttributeError:
-        return None
-
     if not choices:
         return None
 
-    unit = axis_def.unit
-
     if isinstance(choices, dict):
```
pieleric, line 541: Please revert to the original version.

src/odemis/gui/conf/util.py
```diff
@@ -1010,7 +1018,7 @@ def create_axis_entry(container, name, comp, conf=None):
         # that's always the case for combo-boxes anyway)
         lbl_ctrl, value_ctrl = container.add_combobox_control(label_text, conf={"style": wx.CB_READONLY})
 
-        choices_fmt = format_axis_choices(name, ad)
+        choices_fmt = format_axis_choices(name, ad.choices, ad.unit)
```
pieleric, line 1021: This one should also be reset.

src/odemis/gui/cont/streams.py
```diff
@@ -2725,6 +2741,23 @@ def addTemporalSpectrum(self):
             # remove exposureTime from local (GUI) VAs to use a new one, which allows to integrate images
             detvas.remove("exposureTime")
 
+        spg = self._getAffectingSpectrograph(main_data.streak_ccd)
+
+        # band should be set to the pass-through by default
```
pieleric, line 2746: Mark this line as a TODO too.

src/odemis/gui/cont/streams.py
```diff
@@ -2734,42 +2767,47 @@ def addTemporalSpectrum(self):
...
+        return self._addRepStream(ts_stream, sem_ts_stream)
+
+    def addMonochromator(self):
+        """ Create a Monochromator stream and add to to all compatible viewports """
+
+        main_data = self._main_data_model
+        spg = self._getAffectingSpectrograph(main_data.spectrometer)
 
-        # TODO: all the axes, including the filter band should be local. The
         # band should be set to the pass-through by default
```
pieleric, line 2786: Please mark this comment as a `TODO`. (now that the axis is local, it should be easy to implement actually, but let's do it separately).

src/odemis/gui/conf/util.py
amuskens: Although now - we don't have axis_def anymore due to the changes of the function

src/odemis/acq/stream/_base.py
```diff
@@ -444,7 +444,7 @@ def _duplicateAxis(self, axis_name, actuator):
         pos = actuator.position.value[axis_name]
 
         if hasattr(axis, "choices"):
-            return model.FloatEnumerated(pos, choices=axis.choices, unit=axis.unit)
+            return model.VAEnumerated(pos, choices=axis.choices, unit=axis.unit)
```
pieleric, line 447: Why did you have to use a `VAEnumerated` instead of a `FloatEnumerated`? Note that the `FloatEnumerated` class accepts integers (or at least it should!). I don't think we have any axis which is not a float. So it should be compatible with everything...

src/odemis/acq/stream/_base.py
```diff
@@ -257,6 +265,22 @@ def __init__(self, name, detector, dataflow, emitter, focuser=None, opm=None,
             self._updateHistogram(drange_raw)
             self._onNewData(None, self.raw[0])
 
+    def _duplicateAxes(self, axis_map):
```
pieleric, line 268: Please move this function to just before `_duplicateAxis()`.

src/odemis/gui/conf/util.py
pieleric: No, I mean you should revert the function signature and the internal code to be like it used to be (look it up in the git history, or on the github diff page). The only caller, which is in `create_axis_entry()` should also be reverted. 
The idea is that there is no need to change code that worked fine. This new version probably also works, but it's no necessary to mix it with the new feature you are adding.

src/odemis/gui/cont/streams.py
```diff
@@ -331,6 +333,21 @@ def _add_stream_setting_controls(self):
             conf = stream_config.get(vaname)
             self.add_setting_entry(vaname, va, hw_comp=None, conf=conf)
 
+    def _add_axis_controls(self):
+        """
+        Add controls for the axes that are connected to the stream
+        """
+        stream_config = self._stream_config.get(type(self.stream), {})
+        if not hasattr(self.stream, "axis_vas"):
```
pieleric, line 341: Please, move this `if` into the caller. It's the same behaviour, but the rest of the functions work like this, so let's follow the flow.

src/odemis/acq/stream/_base.py
amuskens: So this is what caused the problem with the actuators because there was one axis which was an int. Changing this back fixed the issue. But maybe I need to determine if I should make an IntEnumerated?

amuskens: So this is what caused the problem with the actuators because there was one axis which was an int. Changing this back fixed the issue. But maybe I need to determine if I should make an IntEnumerated?

amuskens: So this is what caused the problem with the actuators because there was one axis which was an int. Changing this back fixed the issue. But maybe I need to determine if I should make an IntEnumerated?

PR #1368, opened 2020-03-11
gui cont Sparc2AlignTab: Handle better cancelation of manual focus

pieleric:
Two (weird) user behaviours were not well handled:
1. The user quickly toggle/untoggle the manual focus button (a couple of
times). In such case, every click would be executed, at the same time.
Eventually the right (last) thing would happen, but for a while
the gauge would get crazy.
=> Cancel the previous manul focus action before starting a new one.

2. The user starts manual focus, and immediately after presses auto
focus. It would result in both running, then the auto focus taking
control.
=> Immediately disable the auto focus button when the user toggles
the manual focus. It's now possible to also immediately change the
grating and detector. That's fine.

### DIFF
--- a/src/odemis/gui/cont/tabs.py
+++ b/src/odemis/gui/cont/tabs.py
@@ -3979,7 +3979,8 @@ class Sparc2AlignTab(Tab):
             for st in focusedview.getStreams():
                 if st in self._focus_streams and st is not should_update_stream:
                     focusedview.removeStream(st)
-            focusedview.addStream(should_update_stream)
+            if not should_update_stream in focusedview.stream_tree:
+                focusedview.addStream(should_update_stream)
             # Set the focus detectors combobox selection
             try:
                 istream = self._focus_streams.index(should_update_stream)
@@ -4039,6 +4040,7 @@ class Sparc2AlignTab(Tab):
         bl = main.brightlight
         align_mode = self.tab_data_model.align_mode.value
         gauge = self.panel.gauge_autofocus
+        self._mf_future.cancel()  # In case it's still changing, immediately stop (the gauge)
 
         if event.GetEventObject().GetValue():  # manual focus btn toggled
             # Set the optical path according to the align mode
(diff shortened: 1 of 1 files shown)


src/odemis/gui/cont/tabs.py
```diff
@@ -4087,14 +4090,13 @@ def _onManualFocusReady(self, future):
...
 
-        align_mode = self.tab_data_model.align_mode.value
-        if align_mode == "lens-align":
-            self._enableFocusComponents(manual=True, ccd_stream=False)
-
     def _onManualFocusFinished(self, future):
         """
         Called when finishing manual focus is done
         """
+        if future.cancelled():
```
lazem: Thanks for catching these erratic cases. 
We talked also about the gauge when cancelled, being stopped in its track. 
Maybe it can be resetted here with:  
`self.panel.gauge_autofocus.SetValue(0)`

pieleric: It sounds good. So I've tried... and it didn't work. I think it's because the ProgressiveFutureConnector also sees the end of the Future, and updates one last time the gauge, just a tiny bit later.
So I've tried to put it into `onAlignMode()`, that works, but then the gauge is *always* reset, even after successfully stopping the manual focus, or completion of the autofocus. That's a little be too often.
So I've tried to put it into Show(True), and that seems to work :-)

lazem: Nice, it works on switching the tabs, though won't work on switching align mode (Mirror, Centering), but I think it's fine to see it stopped.

pieleric: Indeed, I had forgotten that changing the align mode would also cancel the focus. I guess if we really wanted to fix this case, one way would be to adjust the ProgressiveFutureConnector, with a special parameter indicating that when the future is cancelled we want the gauge to be set to 0 (instead of leaving it as-is).  For now, let's just leave it as-is.

PR #1371, opened 2020-03-16
Add the shear of the grid to the spot grid live visualization.

tepals:
To see if the shape of the grid is perfectly square it is important to see the shear.
To keep the FindGridSpots function compatible with scripts outside odemis I decided to only return the shear if specifically requested.

### DIFF
--- a/scripts/spot-grid.py
+++ b/scripts/spot-grid.py
@@ -51,10 +51,11 @@ class VideoDisplayerGrid(VideoDisplayer):
         at ratio 1:1)
         data (numpy.ndarray): an 2D array containing the image (can be 3D if in RGB)
         """
-        self.app.spots, trans, scale, rot = FindGridSpots(data, self.gridsize)
+        self.app.spots, trans, scale, rot, shear = FindGridSpots(data, self.gridsize)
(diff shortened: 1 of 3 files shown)


pieleric:
Indeed, it makes sense to return the shear as well, especially as it's already computed.
However, such type of method signature, which changes the number of return values based on the arguments, is really annoying (in general). Python allows you to do that, but it's ugly and confusing. So I'd suggest to try avoiding it as much as possible. If I understand correctly, in addition to the (1) usage in Odemis, there are a couple of scripts outside (3?). 
Knowing it's really a small number and it's easy to for you to update all the code, I'd suggest you do just that. In particular, in Python 3, if you want to have the caller be compatible with both old and newer versions of the function you can write it this way:
`spot_coordinates, t, s, r, *_ = FindGridSpots(ccd_image, (8, 8))`
or, if you only care about the first argument:
`spot_coordinates, *_ = FindGridSpots(ccd_image, (8, 8))`

A second note: please update the unit test, to check the shear too (no need for very fancy code, but at least read this new return value).

tepals:
Updated. I agree, it does look better to not change the return values. I'll also update the scripts outside odemis.

PR #1375, opened 2020-03-17
Light Power Refactoring

lazem:


src/odemis/driver/test/pwrcomedi_test.py
```diff
@@ -65,61 +65,55 @@ def tearDown(self):
...
+            # value so small that it's == 0 for the hardware
+            self.dev.power.value[i] *= 1e-8
             time.sleep(1)
-            self.assertEqual(self.dev.emissions.value, em)
-#             # value so small that it's == 0 for the hardware
-#             self.dev.emissions.value[i] = 1e-8
-#             em[i] = 0
-#             self.assertEqual(self.dev.emissions.value, em)
+            self.assertGreater(self.dev.power.value[i], 0)
+            # self.assertEquals(self.dev.power.value[i], 0)
```
lazem: How can we check it's so small that it's 0 for the driver? Currently it will always be greater by fraction.

src/odemis/driver/pwrcomedi.py
```diff
@@ -211,30 +209,30 @@ def getHwName(self):
...
+    #     intensities (list of N floats [0..1]): intensity of each source
+    #     """
+    #     if len(intensities) != len(self._channels):
+    #         raise ValueError("Emission must be an array of %d floats." % len(self._channels))
+    #
+    #     # clamp intensities which cannot reach the maximum power
+    #     cl_intens = []
+    #     for crv, intens in zip(self._pwr_curve, intensities):
+    #         cl_intens.append(min(max(0, intens), crv[-1][1] / self.power.range[1]))
+    #         # TODO: if the intensity is so small that the power would be 0, force
```
lazem: How to do this?

src/odemis/driver/pwrcomedi.py
```diff
@@ -124,33 +124,31 @@ def __init__(self, name, role, device, channels, spectra, pwr_curve, **kwargs):
...
+                                                        ".".join("%s" % v for v in lnx_ver))
         self._metadata[model.MD_SW_VERSION] = self._swVersion
-        self._metadata[model.MD_HW_VERSION] = self._hwVersion # unknown
+        self._metadata[model.MD_HW_VERSION] = self._hwVersion  # unknown
 
     def terminate(self):
         if self._device:
             # Make sure everything is powered off
-            self._updateIntensities(0, [0.] * len(self._channels))
+            self._updateIntensities([0.] * len(self._channels))
```
pieleric: You could also write `self._updateIntensities(self.power.range[0])`, which is slightly more explicit.

src/odemis/driver/pwrcomedi.py
```diff
@@ -124,33 +124,31 @@ def __init__(self, name, role, device, channels, spectra, pwr_curve, **kwargs):
...
             spect.append(tuple(wls))
 
-        max_power = max(crv[-1][1] for crv in self._pwr_curve)
-        # power of the whole device (=> max power of the device with max power)
-        self.power = model.FloatContinuous(0., (0., max_power), unit="W")
+        # Maximum power for channel to be used as a range for power
+        max_power = [crv[-1][1] for crv in self._pwr_curve]
+        # Power value for each channel of the device
+        self.power = model.ListContinuous(value=[0.] * len(self._channels),
+                                          range=([0.] * len(self._channels), max_power,), unit="W")
```
pieleric: The range should be tuples, as a way to convey it shouldn't be changed.

src/odemis/driver/pwrcomedi.py
pieleric: You'd check if the return value of `self._volt_to_data(v, c, r)` == `self._volt_to_data(crv[0][0], c, r)` (that's the minimum value).
Ideally, you'd even compute back for any value: convert the requested power (float) to the "data" (ie, int), and then convert back from the int to the actual power. This would indicate to the user the actual value set. That said, no one really cares much, excepted potentially for the "tiny power is no power" case (discussed above), where it might sometimes be useful.
Note that this wasn't handle so far, so you don't need to handle it now either.

src/odemis/driver/test/pwrcomedi_test.py
pieleric: Indeed, the driver would need to reverse back the actual data into a power value to realise it's actually 0. Not a big deal though.

src/odemis/driver/pwrcomedi.py
```diff
@@ -59,7 +59,7 @@ def __init__(self, name, role, device, channels, spectra, pwr_curve, **kwargs):
 
         try:
             self._device = comedi.open(device)
-#             self._fileno = comedi.fileno(self._device)
+        #             self._fileno = comedi.fileno(self._device)
```
Kleijwegt: What is this change about?

src/odemis/driver/pwrcomedi.py
```diff
@@ -105,7 +105,7 @@ def __init__(self, name, role, device, channels, spectra, pwr_curve, **kwargs):
             # Find the best range to use
             try:
                 ri = comedi.find_range(self._device, self._ao_subd,
-                                      c, comedi.UNIT_volt, crv[0][0], crv[-1][0])
+                                       c, comedi.UNIT_volt, crv[0][0], crv[-1][0])
```
Kleijwegt: Also seems like nothing is changed here

src/odemis/model/_vattributes.py
```diff
@@ -1123,6 +1123,33 @@ def _check(self, value):
...
+            raise ValueError("value (%s) is not a list" % (value,))
+        self._cls = cls or value[0].__class__
+        Continuous.__init__(self, range)
+        ListVA.__init__(self, value, unit=unit, **kwargs)
+
+    def _set_value(self, value, **kwargs):
+        # force list
+        value = list(value)
+        ListVA._set_value(self, value, **kwargs)
+    # need to overwrite the whole property
```
Kleijwegt: Can you add in this comment why you need to overwrite the whole property?

lazem:
Thanks @pieleric @Kleijwegt for the review. I updated the rest of drivers that use power/emission in their logic. Appreciating your review on them.

Thanks!

src/odemis/driver/lle.py
```diff
@@ -148,16 +148,15 @@ def __init__(self, name, role, port, sources, _serial=None, **kwargs):
         model.Emitter.__init__(self, name, role, **kwargs)
 
         self._shape = ()
-        self._max_power = 0.4 # W (According to doc: ~400mW)
-        self.power = model.FloatContinuous(0., (0., self._max_power), unit="W")
+        # Maximum power taken from the manual file 4487-LUM.54-10009B.SPECTRA.X.pdf
+        self._max_power = (0.231, 0.260, 0.196, 0.295, 0.310, 0.256, 0.62)
```
pieleric: The idea here is that the device might actually have less than 7 sources installed. This depends on the sources. In that case the spectra and power could be shorter. So you cannot directly hardcode it here. 

As a matter of fact, most of the systems using this device only have 4 sources. See https://lumencor.com/products/filters-for-spectra-x-light-engines/ for the power (which is dependent) on the sources. Ideally, the `sources` parameter would also have the max power. We could do that, but we'd have to update the YAML file of every single system. An alternative could be to support the power as 6th value in the spectra, eg `[379.e-9, 384.e-9, 390.e-9, 396.e-9, 401.e-9, 0.225]`, but *also* to have a table with known power for the one listed in the website, and use that as a fallback when there are only 5 values.

src/odemis/driver/lle.py
```diff
@@ -372,91 +371,78 @@ def _getIntensityGY(self, intensities):
...
         # Yellow has precedence over green
         if yellow_i is not None and intensities[yellow_i]: # don't use if None or 0
-            return intensities[yellow_i]
+            return intensities[yellow_i], yellow_i
         elif green_i is not None:
-            return intensities[green_i]
+            return intensities[green_i], green_i
         else:
-            return 0
+            return (0, 0)
```
pieleric: The second 0 is not really correct. It should still be one of the `yellow_i` or `green_i`. Of course, based on how it's used afterwards (multiply it by 0), that doesn't matter much. It'd just feel less odd to use a "random" channel.

src/odemis/driver/lle.py
lazem: So instead of hardcoding `max_power` directly, I used `DEFAULT_SOURCES_POWERS` dict with these default values and filled `max_power` with only the existing channels. Is that correct?

src/odemis/odemisd/mdupdater.py
```diff
@@ -262,12 +262,11 @@ def updateMDFromVA(val, md_key=md_key, comp_affected=comp_affected):
         return True
 
     def observeLight(self, light, comp_affected):
-
-        def updateInputWL(emissions, light=light, comp_affected=comp_affected):
+        def updateLightPower(power, light=light, comp_affected=comp_affected):
```
lazem, line 265: I removed `updateInputWL` which include emissions update, and moved updating `model.MD_IN_WL` to `updateLightPower` 
@pieleric

src/odemis/acq/stream/_live.py
```diff
@@ -724,6 +721,17 @@ class BrightfieldStream(CameraStream):
       all the spectrum), or a "light" emitter (with multiple channels, for
       various spectra). To activate the light, it just turns on all the channels.
     """
+    def __init__(self, name, detector, dataflow, emitter, emtvas=None, **kwargs):
+
+        if emitter is not None:
+            # Current power VA representing power for one channel only
+            cp_range = (emitter.power.range[0][0], emitter.power.range[1][0])
```
lazem, line 728: Could a `BrightfieldStream` emitter have something other than range in its power?  In other words should I check if range exists here?
@pieleric

plugins/sracq.py
```diff
@@ -113,12 +113,8 @@ def __init__(self, microscope, main_app):
...
         self.addMenu("Acquisition/Super-resolution...", self.start)
 
-        # light.power has a maximum range corresponding to the maximum source,
-        # and other sources are actually proportionally scaled, which is very
-        # confusing. So make it a bit more obvious that it's just a ratio.
-        pwr = 100 * self.light.power.value / self.light.power.range[1]
-        self.power = model.FloatContinuous(pwr, range=(0, 100), unit="%",
-                                           setter=self._set_power)
+        # light.power has a maximum range corresponding to each source
+        self.power = self.light.power
```
lazem, line 117: I removed this power conversion part but I'm not sure what is the use of self.power in this plugin now. Should it be removed or the ratio is used somewhere?
@pieleric

lazem:
Hi @pieleric  
I have updated this pull request with 4 additional commits:
- Power API driver refactoring 
- Power API Stream Refactoring
- Power API Refactoring - Update scripts and plugins
- Power API refactoring - Adjust the GUI

Appreciating your review on this.  
Thanks

src/odemis/driver/lle.py
```diff
@@ -32,6 +32,7 @@
 import serial
 import threading
 import time
+import random
```
pieleric, line 35: No need for this module.

src/odemis/driver/lle.py
```diff
@@ -372,91 +380,78 @@ def _getIntensityGY(self, intensities):
...
-                    intensities[self._gy[0]] = 0
+                    intensities[self._gy[1]] = 0.
+                else:  # second is the strongest
+                    intensities[self._gy[0]] = 0.
 
         # set the actual values
         for i, intensity in enumerate(intensities):
             # clip + indicate minimum step
-            if intensity < 1 / 256:
+            if intensity/self._max_power[i] < 1 / 256 :
```
pieleric, line 447: Format: drop the space before `:`, and add space around `/`.

src/odemis/driver/lle.py
```diff
@@ -372,91 +380,78 @@ def _getIntensityGY(self, intensities):
...
                 logging.debug("Clipping intensity from %f to 0", intensity)
-                intensity = 0
-            elif intensity > 255 / 256:
-                intensity = 1
-            intensities[i] = intensity
-            self._intensities[i] = intensity * self.power.value
+                intensity = 0.
+            elif intensity/self._max_power[i] > 255 / 256:
```
pieleric, line 450: Add space around `/`.

src/odemis/driver/pwrcomedi.py
```diff
@@ -124,33 +124,31 @@ def __init__(self, name, role, device, channels, spectra, pwr_curve, **kwargs):
...
+                                                        ".".join("%s" % v for v in lnx_ver))
         self._metadata[model.MD_SW_VERSION] = self._swVersion
-        self._metadata[model.MD_HW_VERSION] = self._hwVersion # unknown
+        self._metadata[model.MD_HW_VERSION] = self._hwVersion  # unknown
 
     def terminate(self):
         if self._device:
             # Make sure everything is powered off
-            self._updateIntensities(0, [0.] * len(self._channels))
+            self._updateIntensities(self.power.range[0])
```
pieleric, line 151: Let's write explicitly `self.power.value = self.power.range[0]`

src/odemis/driver/pwrcomedi.py
```diff
@@ -211,30 +209,13 @@ def getHwName(self):
...
-        # clamp intensities which cannot reach the maximum power
-        cl_intens = []
-        for crv, intens in zip(self._pwr_curve, intensities):
-            cl_intens.append(min(max(0, intens), crv[-1][1] / self.power.range[1]))
-            # TODO: if the intensity is so small that the power would be 0, force
-            # it to 0.
-
-        self._updateIntensities(self.power.value, cl_intens)
-        return cl_intens
+        self._updateIntensities(value)
```
pieleric, line 221: You can move `_updateIntensities()` into `_updatePower()`, as it's the only caller.

src/odemis/driver/test/emitter_test.py
```diff
@@ -73,61 +73,50 @@ def tearDown(self):
...
-
-        # turn on second source to 90%
-        self.dev.power.value = self.dev.power.range[1]
-        em[0:2] = [0, 0.9]
-        self.dev.emissions.value = em
-        self.assertGreater(self.dev.emissions.value[1], 0)
+        self.dev.power.value[0] = self.dev.power.range[1][0] * 0.5
+        self.assertGreater(self.dev.power.value[0], 0)
+
+        # turn on second source (first channel) to 90%
```
pieleric, line 84: That's the "third" source. => update the comment.

src/odemis/model/_vattributes.py
```diff
@@ -1123,6 +1123,33 @@ def _check(self, value):
...
+        """
+        if not isinstance(value, list):
+            raise ValueError("value (%s) is not a list" % (value,))
+        self._cls = cls or value[0].__class__
+        Continuous.__init__(self, range)
+        ListVA.__init__(self, value, unit=unit, **kwargs)
+
+    def _set_value(self, value, **kwargs):
+        # force list
+        value = list(value)
```
pieleric, line 1141: No need for this, as `ListVA._set_value()` does this too.

src/odemis/driver/blinkstick.py
```diff
@@ -65,15 +65,11 @@ def __init__(self, name, role, sn=None, max_power=0.1, inversed=False, **kwargs)
...
-        # TODO: allow to change the power also via emissions? Or just make it a choice
-        self.emissions = model.ListVA([0], unit="", setter=self._setEmissions)
-        self.emissions.subscribe(self._updatePower)
         # list of 5-tuples of floats
         self.spectra = model.ListVA([(380e-9, 390e-9, 560e-9, 730e-9, 740e-9)],
                                     unit="m", readonly=True)
 
-        self.power = model.FloatContinuous(0., (0., self._max_power), unit="W",
-                                           setter=self._setPower)
+        self.power = model.ListContinuous([0.,], ((0.,), (max_power,)), unit="W", setter=self._setPower)
```
pieleric, line 72: For *all* these ListContinuous, you need to indicate that the cls can be an int too. Otherwise, it's really annoying in practice, to have to put a magical `.` after 0 or 1. I think you can just write: `cls=numbers.Real`. Otherwise, you'll have to do like what we've done for a lot of TupleContinous: `(int, long, float)`, and import `long` from builtins.past, to be compatible with Python 2 and 3.

src/odemis/model/_vattributes.py
```diff
@@ -1123,6 +1123,33 @@ def _check(self, value):
                     (axis, v, self._range[0], self._range[1]))
 
 
+class ListContinuous(ListVA, Continuous):
+    def __init__(self, value, range, unit="", cls=None, **kwargs):
+        """
+        range (2 x list): minimum and maximum size for each dimension
```
pieleric, line 1129: "list" -> "tuple".

src/odemis/acq/stream/_live.py
```diff
@@ -724,6 +721,17 @@ class BrightfieldStream(CameraStream):
...
+            self.current_power = model.FloatContinuous(emitter.power.value[0],
```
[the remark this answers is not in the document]
lazem: Thanks for the review @pieleric. The reason to name it `current_power` is not to confuse it with the emitter `power` which is a list of values, while the duplicated VA on the stream is just one channel value corresponding to the currently activated channel, but from the GUI side, it doesn't have to differentiate between list and float.

src/odemis/gui/cont/streams.py
```diff
@@ -808,6 +825,22 @@ def _add_dye_ctrl(self):
...
 
+    def _onExcitationChannelChange(self, _):
+        """
+        Event handler for the Excitation channel combobox selection.
+        Update the power slider range with the current power VA's
+        """
+        power_ent = next((spe for spe in self.entries if spe.name == "power"), None)
+        if not power_ent:
+            return
+        # Set the slider with min value first
```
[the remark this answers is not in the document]
lazem: That's right, I will mention that the slider complains when the value is out of range.

src/odemis/gui/cont/streams.py
```diff
@@ -808,6 +825,22 @@ def _add_dye_ctrl(self):
...
+        power_ent.value_ctrl.SetValue(power_ent.vigilattr.min)
+        power_ent.value_ctrl.value = power_ent.vigilattr.min
```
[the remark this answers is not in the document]
lazem: Mmmh it didn't work with `SetRange` alone, but it seems that was a side effect from setting the VA's range and value. It indeed works fine now without changing `value` as well.

src/odemis/acq/stream/_live.py
```diff
@@ -905,13 +920,19 @@ def __init__(self, name, detector, dataflow, emitter, em_filter, **kwargs):
...
+        self.current_channel_idx = emitter.spectra.value.index(current_exc)
         # The wavelength band on the out path (set when emission changes)
         self.emission = model.VAEnumerated(current_em, choices=set(em_choices.values()),
                                            unit="m")
         self.emission.subscribe(self.onEmission)
 
-        # colouration of the image
+        # Current power VA representing power for one 'currently selected' channel only
+        cp_range = (emitter.power.range[0][self.current_channel_idx], emitter.power.range[1][self.current_channel_idx])
+        self.current_power = model.FloatContinuous(emitter.power.value[self.current_channel_idx], readonly=emitter.power.readonly, range=cp_range,
```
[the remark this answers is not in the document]
lazem: Ahah right, I think I was inspired by the way `_duplicateVA` recreated the objects, but sure here it has always to be True.

plugins/sracq.py
lazem: Yeah I noticed the plugin doesn't use the stream directly, so should I remove `power` attribute from the plugin entirely @pieleric ? 

And also switching off the light power?
```
# Switch off laser
self.light.power.value = self.light.power.range[0]
```

pieleric: Yes, you should remove `.power` from the plugin class.
To switch off the power (l 322), your proposal seems fine.

src/odemis/driver/test/emitter_test.py
lazem: It's correct, second source, first channel -> multiplex power idx 2.

src/odemis/model/_vattributes.py
lazem: It made an issue with me while terminating the multiplex tests. The dependencies `terminate()` sets power with range[0] and it stayed tuple.

src/odemis/driver/emitter.py
```diff
@@ -45,93 +45,26 @@ def __init__(self, name, role, dependencies, **kwargs):
...
-            if pwr > 0:
-                pratio = child.power.value / pwr
-            else:
-                pratio = child.power.range[1] / self.power.range[1]
-            for i, e in enumerate(child.emissions.value):
-                em[idx + i] = e * pratio
-                logging.debug("Read em %d as %s * %g W => %s * %g W",
-                              idx + i, e, child.power.value, em[idx + i], pwr)
-        return pwr, em
-
```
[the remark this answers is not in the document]
lazem: Thanks for clarification. I implemented these two behaviours and added a test function for them in `TestMultiplexLight`. The second behaviour (updating child -> update multiplex) made me subscribe to each children power va. Can you please check if the logic is correct?

lazem:
Hi @pieleric  I added another commit to handle the errors I got after running the full test. 
Appreciating your review. 
Thanks

src/odemis/dataio/test/hdf5_test.py
```diff
@@ -1126,7 +1126,7 @@ def testReadMDOutWlBands(self):
                      model.MD_IN_WL: (500e-9, 520e-9),  # m
                      model.MD_OUT_WL: ((630e-9, 660e-9), (675e-9, 690e-9)),  # m
                      model.MD_USER_TINT: (255, 0, 65),  # purple
-                     model.MD_LIGHT_POWER: 100e-3  # W
+                     model.MD_LIGHT_POWER: [100e-3]  # W
```
pieleric, line 1129: No. MD_LIGHT_POWER shouldn't change of format. It's still a float. So please revert these changes in all the dataio module.
In order to make sure it stays the same, you will have to adjust odemisd.mdupdater. Now MD_LIGHT_POWER is slightly more complex to compute: it's not the `power.value`, but the `sum(power.value)`.

src/odemis/driver/emitter.py
```diff
@@ -57,14 +58,42 @@ def __init__(self, name, role, dependencies, **kwargs):
...
+        # To avoid infinite loop from alternating subscriber calls
+        child_power = []
+        for child in self._child_idx.keys():
+            child_power.extend(child.power.value)
+        if power == child_power:
+            return
+        # Update child powers with new values
+        power = copy.deepcopy(power)  # Not to be attached to self.power.value
+        for child, idx in self._child_idx.items():
+            child.power.value[:] = power[idx:idx+len(child.power.value)]
```
pieleric, line 87: Please drop the `[:]`. `NotifyingList` is actually clever enough to detect it too, but it's easier to just set it explicitly.

src/odemis/driver/emitter.py
```diff
@@ -57,14 +58,42 @@ def __init__(self, name, role, dependencies, **kwargs):
...
+        """
+        # Prevent re-updating child power if power values are the same
+        # To avoid infinite loop from alternating subscriber calls
+        child_power = []
+        for child in self._child_idx.keys():
+            child_power.extend(child.power.value)
+        if power == child_power:
+            return
+        # Update child powers with new values
+        power = copy.deepcopy(power)  # Not to be attached to self.power.value
```
pieleric, line 85: `deepcopy()` sounds little over the top as actually it's just a list of float (not recursivity). -> `copy.copy()`.

src/odemis/odemisd/mdupdater.py
```diff
@@ -277,25 +276,12 @@ def updateInputWL(emissions, light=light, comp_affected=comp_affected):
...
-            p = power * numpy.sum(emissions)
-
-            md = {model.MD_IN_WL: wl_range,
-                  model.MD_LIGHT_POWER: p}
-            comp_affected.updateMetadata(md)
-
-        def updateLightPower(power, light=light, comp_affected=comp_affected):
-            p = power * numpy.sum(light.emissions.value)
-            md = {model.MD_LIGHT_POWER: p}
+            md = {model.MD_IN_WL: wl_range, model.MD_LIGHT_POWER: power}
```
pieleric, line 279: `power` -> `sum(power)`

pieleric:
BTW, note that there are some (minor) conflicts now. Please rebase on upstream/master too.

lazem:
Hi @pieleric 

Thanks for your review! I have fixed the issue with MD_LIGHT_POWER update, and rebased the branch on upstream/master. Can you please have a look for any further issue, then I willl squash the commits for merging.

Thanks!

scripts/demo_overlay.py
```diff
@@ -145,18 +145,18 @@ def _DoAcquisition(future, escan, ccd, detector, light):
         logging.debug("Acquiring CCD images...")
 
         # Turn on light for CCD acquisition
-        emissions = [1, 0, 0, 0, 0, 0, 0]
-        light.emissions.value = emissions
+        intensities = [1, 0, 0, 0, 0, 0, 0]
+        light.power.value = [ints * pw for ints,pw in zip(intensities, light.power.value)]
```
pieleric, line 149: I think that you need to change `light.power.value` -> `light.power.range[1]`

scripts/demo_overlay.py
```diff
@@ -167,8 +167,8 @@ def _DoAcquisition(future, escan, ccd, detector, light):
             future._acq_state = FINISHED
 
         # Turn off light for CCD acquisition
-        emissions = [0] * len(light.emissions.value)
-        light.emissions.value = emissions
+        intensities = [0] * len(light.power.value)
+        light.power.value = intensities
```
pieleric, line 171: Just write `light.power.value = light.power.range[0]`.

scripts/hracq.py
```diff
@@ -142,7 +141,7 @@ def acquire(self):
         logging.info("Finished with average %g fps", fps)
 
         # Switch off laser
-        self.light.power.value = 0
+        self.light.power.value = [0.] * len(self.light.power.value)
```
pieleric, line 144: `light.power.value = light.power.range[0]`

src/odemis/acq/align/autofocus.py
```diff
@@ -979,7 +979,7 @@ def updateProgress(subf, start, end):
         future.set_progress(end=time.time() + sum(future._actions_time))
 
         logging.debug("Turning off the light")
-        bl.power.value = bl.power.range[0]
+        bl.power.value = list(bl.power.range[0])
```
pieleric, line 982: It's not needed to convert to list, right? If so, please revert this line and the l.997.

src/odemis/acq/align/light.py
```diff
@@ -66,7 +66,7 @@ def _doTurnOnLight(f, bl, ccd):
         # We need the light to be off, so that we can notice a difference when
         # it turns on.
         # In case it's already turned on, just assume everything is fine.
-        if bl.emissions.value[0] * bl.power.value != 0:
+        if bl.power.value[0] != bl.power.range[0][0]:
```
pieleric, line 69: I think the right way to check if the light is on is `sum(bl.power.value) > 0`.

lazem:
Hi @pieleric 

I updated the code with your suggested changes, rechecked all the changes and fixed minor issues. I also squashed the commits into a single big commit. Would you please have a look?
Thanks alot!

scripts/hracq.py
```diff
@@ -88,6 +88,9 @@ def set_hardware_settings(self, wavelength, power=None):
         """
         if power is None:
             power = self.light.power.range[-1]
+        elif type(power) == float:
+            # Convert the power commandline argument float value into a list (to match light power type)
+            power = [power]
```
lazem, line 93: Does this make sense? @pieleric

pieleric: Good catch, `power` should indeed be a number. That said, to check for a number, it's better to write something like: if `isinstance(power, numbers.Number)`. This catches integers too.

lazem:
Amended the number check in `hracq` module. 
Thanks for all the guidance in this @pieleric Immensely appreciated. :)

PR #1376, opened 2020-03-19
Anamorphosis

effting:
I wrote a new Transform class that supports the 'Anamorphosis transform'. Useful to model stigmator effects.

### DIFF
--- a/src/odemis/util/spot.py
+++ b/src/odemis/util/spot.py
@@ -201,9 +201,10 @@ def FindCenterCoordinates(image, smoothing=True):
     w = numpy.sqrt(dI2 / numpy.hypot(ik - i0, jk - j0))
 
     # Solve the linear set of equations in a least-squares sense.
-    # Note: rcond set to -1 to suppress warnings on numpy 1.11 -> 1.14. It maybe
-    # could be set to None with newer numpy (or entirely removed).
-    ic, jc = numpy.linalg.lstsq(numpy.vstack((w * a, w * b)).T, w * c, rcond=-1)[0]
+    # Note: rcond is set explicitly to have correct and consistent behavior,
+    #       independent on numpy version.
+    rcond = numpy.finfo(numpy.float64).eps * max(m, n)
+    ic, jc = numpy.linalg.lstsq(numpy.vstack((w * a, w * b)).T, w * c, rcond)[0]
 
     # Convert from index (top-left) to (center) position information.
     xc = jc - 0.5 * float(m) + 0.5
@@ -255,17 +256,19 @@ def MaximaFind(image, qty, len_object=18):
 
     # dilation
     structure = _CreateSEDisk(len_object // 2)
-    dilated = cv2.dilate(filtered, structure, iterations=1)
+    dilated = cv2.dilate(filtered, structure)
 
     # find local maxima
-    binary = (filtered == dilated)
+    binary = (filtered == dilated).astype(numpy.uint8)
 
     # thresholding
     qradius = max(len_object // 4, 1)
     structure = _CreateSEDisk(qradius)
-    BWdil = cv2.dilate(numpy.array(binary).astype(numpy.uint8), structure)
+    binary_dilated = cv2.dilate(binary, structure)
 
-    labels, num_features = ndimage.label(BWdil)
+    # Identify all features and then select qty of features with highest mean
+    # value in the filtered image
+    labels, num_features = ndimage.label(binary_dilated)
     index = numpy.arange(1, num_features + 1)
     mean = ndimage.mean(filtered, labels=labels, index=index)
     si = numpy.argsort(mean)[::-1][:qty]
--- a/src/odemis/util/spot.py
+++ b/src/odemis/util/spot.py
@@ -201,10 +201,9 @@ def FindCenterCoordinates(image, smoothing=True):
     w = numpy.sqrt(dI2 / numpy.hypot(ik - i0, jk - j0))
 
     # Solve the linear set of equations in a least-squares sense.
-    # Note: rcond is set explicitly to have correct and consistent behavior,
-    #       independent on numpy version.
-    rcond = numpy.finfo(numpy.float64).eps * max(m, n)
-    ic, jc = numpy.linalg.lstsq(numpy.vstack((w * a, w * b)).T, w * c, rcond)[0]
+    # Note: rcond set to -1 to suppress warnings on numpy 1.11 -> 1.14. It maybe
+    # could be set to None with newer numpy (or entirely removed).
+    ic, jc = numpy.linalg.lstsq(numpy.vstack((w * a, w * b)).T, w * c, rcond=-1)[0]
 
     # Convert from index (top-left) to (center) position information.
     xc = jc - 0.5 * float(m) + 0.5
@@ -256,19 +255,17 @@ def MaximaFind(image, qty, len_object=18):
 
     # dilation
     structure = _CreateSEDisk(len_object // 2)
-    dilated = cv2.dilate(filtered, structure)
+    dilated = cv2.dilate(filtered, structure, iterations=1)
 
     # find local maxima
-    binary = (filtered == dilated).astype(numpy.uint8)
+    binary = (filtered == dilated)
 
     # thresholding
     qradius = max(len_object // 4, 1)
     structure = _CreateSEDisk(qradius)
-    binary_dilated = cv2.dilate(binary, structure)
+    BWdil = cv2.dilate(numpy.array(binary).astype(numpy.uint8), structure)
 
-    # Identify all features and then select qty of features with highest mean
-    # value in the filtered image
-    labels, num_features = ndimage.label(binary_dilated)
+    labels, num_features = ndimage.label(BWdil)
     index = numpy.arange(1, num_features + 1)
     mean = ndimage.mean(filtered, labels=labels, index=index)
     si = numpy.argsort(mean)[::-1][:qty]
--- a/src/odemis/util/test/transform_test.py
+++ b/src/odemis/util/test/transform_test.py
@@ -27,6 +27,7 @@ import numpy
 from numpy.linalg import LinAlgError
 import unittest
 
+from odemis.util.spot import GridPoints
 from odemis.util.transform import (_rotation_matrix_from_angle,
                                    _rotation_matrix_to_angle,
                                    RigidTransform, SimilarityTransform,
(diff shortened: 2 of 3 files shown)


src/odemis/util/transform.py
```diff
@@ -818,3 +818,72 @@ def inverse(self):
...
+    def params(self, params):
+        self._params = params
+        a, b1, b2 = params[:3]
+        p, q = (b1.real, b1.imag)
+        r, s = (b2.real, b2.imag)
+        matrix = numpy.array([(p + r, -q + s), (q + s, p - r)])
+        R, S = qrp(matrix)
+        self._rotation = _rotation_matrix_to_angle(R)
+        self._scale = numpy.diag(S)
+        self._shear = S[0, 1] / S[0, 0]
```
pieleric: All these lines could be avoided by just using the setter from AffineTransform: `self.transformation_matrix = matrix` (and you'd just need to drop the property from this class).

src/odemis/util/transform.py
```diff
@@ -818,3 +818,72 @@ def inverse(self):
         Ainv = numpy.dot(tri_inv(SL), numpy.transpose(self.rotation_matrix))
         tinv = -numpy.dot(Ainv, self.translation)
         return self.__class__(matrix=Ainv, translation=tinv)
+
+
+class AnamorphosisTransform(AffineTransform):
+
+    def __init__(self, params):
+        self.params = params
```
pieleric: I'd highly suggest to not use such setter functionality. Firstly, it's hiding that it does a lot of things. Secondly, if the user wants to change the params, it'd better just create a new transformation object. 
So I'd suggest to leave just the `.params` getter (maybe with a better name?), and put the code of the setter into a function named like `_set_from_params()` or `_compute_attributes()`.

src/odemis/util/test/transform_test.py
```diff
@@ -649,5 +651,32 @@ def test_affine_transform_inverse_known_values(self):
             numpy.testing.assert_array_almost_equal(src, tform(dst))
 
 
+class AnamorphosisTransformKnownValues(unittest.TestCase):
+
+    @staticmethod
+    def _grid(n):
```
pieleric: Please explain what kind of grid it generates. The shape is (n x n), but the content is not absolutely obvious.

src/odemis/util/test/transform_test.py
```diff
@@ -649,5 +651,32 @@ def test_affine_transform_inverse_known_values(self):
...
+        xv = yv = numpy.arange(n, dtype=float) - 0.5 * float(n - 1)
```
pieleric: The `float(...)` is not needed. `0.5 * (n -1)` is a float. BTW, maybe it's clearer to write `(n - 1) / 2`?

src/odemis/util/spot.py
```diff
@@ -201,9 +201,10 @@ def FindCenterCoordinates(image, smoothing=True):
     w = numpy.sqrt(dI2 / numpy.hypot(ik - i0, jk - j0))
 
     # Solve the linear set of equations in a least-squares sense.
-    # Note: rcond set to -1 to suppress warnings on numpy 1.11 -> 1.14. It maybe
-    # could be set to None with newer numpy (or entirely removed).
-    ic, jc = numpy.linalg.lstsq(numpy.vstack((w * a, w * b)).T, w * c, rcond=-1)[0]
+    # Note: rcond is set explicitly to have correct and consistent behavior,
+    #       independent on numpy version.
+    rcond = numpy.finfo(numpy.float64).eps * max(m, n)
+    ic, jc = numpy.linalg.lstsq(numpy.vstack((w * a, w * b)).T, w * c, rcond)[0]
```
pieleric: The changes related to this file don't seem to have anything to do with anamorphosis. Please drop them (and submit them via another PR).

src/odemis/util/test/transform_test.py
tepals: It looks like this function is equal to `odemis.util.spot.GridPoints()`

src/odemis/util/transform.py
```diff
@@ -818,3 +818,72 @@ def inverse(self):
         Ainv = numpy.dot(tri_inv(SL), numpy.transpose(self.rotation_matrix))
         tinv = -numpy.dot(Ainv, self.translation)
         return self.__class__(matrix=Ainv, translation=tinv)
+
+
+class AnamorphosisTransform(AffineTransform):
```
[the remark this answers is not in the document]
tepals: Wikipedia indeed gives different options. Is this still a geometric transform? If yes, maybe there's a less ambiguous name for it [https://en.wikipedia.org/wiki/Geometric_transformation](https://en.wikipedia.org/wiki/Geometric_transformation)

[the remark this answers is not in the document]
effting: I added a docstring

src/odemis/util/transform.py
```diff
@@ -818,3 +818,72 @@ def inverse(self):
...
+
+    def __init__(self, params):
+        self.params = params
+
+    @classmethod
+    def from_pointset(cls, x, y):
```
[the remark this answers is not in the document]
effting: I added a docstring

src/odemis/util/transform.py
```diff
@@ -818,3 +818,72 @@ def inverse(self):
         Ainv = numpy.dot(tri_inv(SL), numpy.transpose(self.rotation_matrix))
         tinv = -numpy.dot(Ainv, self.translation)
         return self.__class__(matrix=Ainv, translation=tinv)
+
+
+class AnamorphosisTransform(AffineTransform):
+
+    def __init__(self, params):
```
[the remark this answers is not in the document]
effting: I updated the constructor. Let me know if you think this is OK.

src/odemis/util/transform.py
effting: I reworked the code. It now separately handles the linear (affine) and non-linear coefficients of the transform. For all linear coefficients it now uses the functions inherited from AffineTransform. As a bonus, the attributes (properties) are no longer read-only.

src/odemis/util/transform.py
```diff
@@ -818,3 +818,176 @@ def inverse(self):
...
+    electron optical system, and is described in more detail in [1].
+
+    The anamorphosis transform has the following form:
+
+        wᵢ = A + B₁ w + B₂ w̄ + C₁ w² + C₂ ww̄ + C₃ w̄² +
+                                       D₁ w²w̄ + D₂ ww̄² + E₁ w³w̄² + E₂ w²w̄³,
+
+    where A, B₁, B₂, C₁, C₂, C₃, D₁, D₂, E₁, and E₂ are the (complex)
+    coefficients of the transform. The in- and output coordinates are the
+    complex numbers w = x + j*y. Note that A, B₁, B₂ define an ordinary affine
```
tepals: It would be good to note here that `w̄ ` is the complex conjugate of `w`

src/odemis/util/transform.py
```diff
@@ -818,3 +818,176 @@ def inverse(self):
...
         tinv = -numpy.dot(Ainv, self.translation)
         return self.__class__(matrix=Ainv, translation=tinv)
+
+
+class AnamorphosisTransform(AffineTransform):
+    """
+    AnamorphosisTransform
+
+    The anamorphosis transform is a polynomial model of the distortions in an
+    electron optical system, and is described in more detail in [1].
```
tepals: Would it be possible to add a one or two sentences with a bit more explanation? 

If I understand correctly it's like an affine transform and the higher order coefficients allow for inter-coordinate distortion?

src/odemis/util/transform.py
```diff
@@ -818,3 +818,176 @@ def inverse(self):
...
+        a, b1, b2, c1, c2, c3, d1, d2, e1, e2 = coeffs
+        p, q = (b1.real, b1.imag)
+        r, s = (b2.real, b2.imag)
+        matrix = numpy.array([(p + r, -q + s), (q + s, p - r)])
+        self.transformation_matrix = matrix
+        self.nlcoeffs = (c1, c2, c3, d1, d2, e1, e2)
+        self.translation = (a.real, a.imag)
+
+    def inverse(self):
+        raise NotImplementedError()
```
tepals: Is it possible to calculate the inverse? 

Could you add an explanation to the `NotImplementedError`?
`NotImplementedError("Not implemented because......")`

src/odemis/util/transform.py
```diff
@@ -818,3 +818,176 @@ def inverse(self):
...
+        Constructor for AnamorphosisTransform that determines the best
+        coordinate transformation from two point sets `x` and `y` in a
+        least-squares sense. For more information see [1].
+
+        Parameters
+        ----------
+        x : (n, 2) array
+            Coordinates in the source reference frame.
+        y : (n, 2) array
+            Coordinates in the destination reference frame. Must be of same
```
tepals: The order of the points in in x and y also need to correspond right? So if x = [point1, point5, point2, ....] then y = [point1, point5, point2, ....]

src/odemis/util/transform.py
effting: OK; added message to the exception

src/odemis/util/transform.py
effting: Added note to docstring

src/odemis/util/test/transform_test.py
tepals: This function is equal to `odemis.util.spot.GridPoints()`, please use that function.

src/odemis/util/test/transform_test.py
```diff
@@ -649,5 +651,90 @@ def test_affine_transform_inverse_known_values(self):
...
+            self.assertAlmostEqual(0., _angle_diff(rotation, tform.rotation))
+            numpy.testing.assert_array_almost_equal(scale, tform.scale)
+            self.assertAlmostEqual(shear, tform.shear)
+            numpy.testing.assert_array_almost_equal(translation, tform.translation)
+            numpy.testing.assert_array_almost_equal(0., tform.coeffs[3:])
+
+    def test_anamorphosis_transform_rotation(self):
+        """
+        The rotation component of an AnamorphosisTransform should be equal to
+        the argument of coefficient b1 if the transform is has zero shear and
```
tepals: typo: is has -> has

src/odemis/util/test/transform_test.py
```diff
@@ -649,5 +651,90 @@ def test_affine_transform_inverse_known_values(self):
...
+        for rotation, scale, translation in iterator:
+            tform = AnamorphosisTransform(rotation=rotation, scale=scale,
+                                          translation=translation)
+            b1 = tform.coeffs[1]
+            self.assertAlmostEqual(0., _angle_diff(rotation, numpy.angle(b1)))
+
+    def test_anamorphosis_transform_scale(self):
+        """
+        The scale component of an AnamorphosisTransform should be equal to
+        the absolute value of coefficient b1 if the transform is has zero
```
tepals: typo: is has -> has

src/odemis/util/transform.py
pieleric: Isn't this the case for every `from_pointset()`? If so, I'd say it's fine. Especially it's indicated as "a set of *corresponding* points".

effting:
Yes; that would be the case for all from_pointset() constructors.

PR #1378, opened 2020-03-20
1D and 2D plot export fixes and clean-ups

pieleric:


pieleric:
I'd highly suggest to review the commits one-by-one, as they overlay each other. The couple of first commits are bug fixes and short. The last ones are refactoring (ie, the behaviour of the code shouldn't change) and are quite big.

src/odemis/dataio/csv.py
```diff
@@ -98,7 +96,7 @@ def export(filename, data):
                 unit = "nm"
 
             # attach distance as first row
-            line_length = data.shape[1] * data.metadata[model.MD_PIXEL_SIZE][1]
+            line_length = data.shape[0] * data.metadata[model.MD_PIXEL_SIZE][1]
```
Kleijwegt: Double check:
Probably this change was made on purpose but is the shape of the multiplication correct?
Is **0** meant to be multiplied with **1**?

pieleric: Yes, that was the bug. Actually, even for standard images, that's typically the case, because numpy stores the dimensions in reverse order.

src/odemis/gui/util/img.py
```diff
@@ -917,29 +917,38 @@ def calculate_ticks(value_range, client_size, orientation, tick_spacing):
...
     if value_range is None:
         logging.info("Not drawing scale bar, as range is None")
         return
 
+    # TODO: Instead of just the min/max value, support a whole list of values,
+    # so that if they are not linearly distributed, the ticks are shown at the
+    # right place.
+
+    im = numpy.empty((client_size[1], client_size[0], 4), dtype=numpy.uint8)
+    im.fill(255)
```
tepals: I think this can be replaced with `im = numpy.full((client_size[1], client_size[0], 4), 255, dtype=numpy.uint8)`
Or is this faster?

src/odemis/gui/util/img.py
```diff
@@ -937,13 +937,9 @@ def draw_scale(ctx, value_range, client_size, orientation, tick_spacing,
...
+        logging.info("Not drawing scale bar, as range is None")
         return
 
-    if value_range[0] > value_range[1]:
-        mirror = True
-        v = value_range
-        value_range = (v[1], v[0])
-
     tick_list, _ = calculate_ticks(value_range, client_size, orientation, tick_spacing)
     # TODO use always client_size instead of scale_width
```
tepals: I think you now resolved this todo?

src/odemis/gui/util/img.py
```diff
@@ -1031,8 +1027,8 @@ def draw_scale(ctx, value_range, client_size, orientation, tick_spacing,
 
             if prev_lpos >= lpos + 20 or i == 0 or i == len(tick_list):
                 if mirror:
-                    ctx.move_to(scale_width - lbl_width - 9, client_size[1] - lpos)
-                    ctx.show_text(label)
+#                     ctx.move_to(scale_width - lbl_width - 9, client_size[1] - lpos)
+#                     ctx.show_text(label)
```
tepals: why did you leave this here?

src/odemis/gui/util/img.py
pieleric: Indeed, it would be clearer this way. Actually, I've been looking for such function in numpy for a long time, and never found it!

src/odemis/gui/util/img.py
pieleric: indeed.
Dropped it now.

src/odemis/gui/util/img.py
pieleric: Because currently the function shows the text or not based on whether the scale is shown at the top or bottom (parameter `mirror`). I don't like such arbitrary behaviour. So I have in mind that one day there would be a separate parameter to show or not the text. In such case, this code will be used.

src/odemis/acq/stream/_projection.py
```diff
@@ -1300,19 +1300,21 @@ def _on_selected_line(self, _):
...
-            where spec1d is the line spectrum (1D array)
-            and md is the metadata structure
+        return DataArray of of shape XC: the line spectrum
+           the distance increases in X and the wavelength increases in C
         """
 
         if ((None, None) in self.stream.selected_line.value or
                 self.stream.calibrated.value.shape[0] == 1):
-            return None, None
+            return None
```
amuskens: If the expected return value is a tuple, will this work?

pieleric: I've updated all the (2) callers, and changed all the places with a return. So it should work.

PR #1380, opened 2020-03-21
gui comp popup: fix popup message not shown on Windows

pieleric:


pieleric:
Well, maybe the names are confusing, but the title is shown in big (and supposed to be relatively short), while the message is shown in small (and can be somewhat longer). So actually the most important is the title. The message is just extra information. So the title is required.

PR #1381, opened 2020-03-22
Small changes to spot.py to improve readability and have consistent numpy behavior

effting:
Created a separate pull-request for this as requested.

### DIFF
--- a/src/odemis/util/spot.py
+++ b/src/odemis/util/spot.py
@@ -201,9 +201,10 @@ def FindCenterCoordinates(image, smoothing=True):
     w = numpy.sqrt(dI2 / numpy.hypot(ik - i0, jk - j0))
 
     # Solve the linear set of equations in a least-squares sense.
-    # Note: rcond set to -1 to suppress warnings on numpy 1.11 -> 1.14. It maybe
-    # could be set to None with newer numpy (or entirely removed).
-    ic, jc = numpy.linalg.lstsq(numpy.vstack((w * a, w * b)).T, w * c, rcond=-1)[0]
+    # Note: rcond is set explicitly to have correct and consistent behavior,
+    #       independent on numpy version.
+    rcond = numpy.finfo(numpy.float64).eps * max(m, n)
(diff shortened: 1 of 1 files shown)


src/odemis/util/spot.py
```diff
@@ -255,17 +256,19 @@ def MaximaFind(image, qty, len_object=18):
 
     # dilation
     structure = _CreateSEDisk(len_object // 2)
-    dilated = cv2.dilate(filtered, structure, iterations=1)
+    dilated = cv2.dilate(filtered, structure)
 
     # find local maxima
-    binary = (filtered == dilated)
+    binary = (filtered == dilated).astype(numpy.uint8)
```
pieleric: Could you explain what is this needed? Normally it returns an array of booleans. Maybe that is not compatible with cv2? (I guess that booleans are stored as uint8 anyway, so that's why it'd work anyway).

effting: Yes; I think this has to do with cv2. Note that I didn't added the .astype() call; I just moved it up by a few lines. As you say it might indeed just work without but I think it is better to explicitly cast the array to what is expected by cv2.

src/odemis/util/spot.py
```diff
@@ -255,17 +256,19 @@ def MaximaFind(image, qty, len_object=18):
 
     # dilation
     structure = _CreateSEDisk(len_object // 2)
-    dilated = cv2.dilate(filtered, structure, iterations=1)
+    dilated = cv2.dilate(filtered, structure)
```
tepals: Why did you remove this, is `iterations=1` the default?

src/odemis/util/spot.py
tepals: Yes, this is because of cv2. It doesn't accept it without the .astype() call

src/odemis/util/spot.py
effting: Yes; default is iterations=1; see [link](https://docs.opencv.org/2.4/modules/imgproc/doc/filtering.html#void%20dilate(InputArray%20src,%20OutputArray%20dst,%20InputArray%20kernel,%20Point%20anchor,%20int%20iterations,%20int%20borderType,%20const%20Scalar&%20borderValue))

PR #1382, opened 2020-03-25
acq stream AlignedSEMStream: fix preparation with MTD_MD_UPD

pieleric:
Commit 5c839bba4f31 (don't try to shift more than the hardware supports)
put extra checks for MTD_EBEAM_SHIFT. However, it requires having .shift
VA.

For some bad reason, related to Pyro and proxies, the code previously
didn't break even if the emitter didn't have a .shift VA. It now got
officially broken.
=> Don't access .shift if not using MTD_EBEAM_SHIFT.

### DIFF
--- a/src/odemis/acq/stream/_live.py
+++ b/src/odemis/acq/stream/_live.py
@@ -430,12 +430,12 @@ class AlignedSEMStream(SEMStream):
         """
         res, shift = self._computeROISettings(self.roi.value)
 
-        if (self._shiftebeam == MTD_EBEAM_SHIFT) and (self._beamshift is not None):
-            shift = tuple(s + c for s, c in zip(shift, self._beamshift))
-
-        # always in this order
+        # always in this order: resolution, then shift
         self._emitter.resolution.value = res
-        self._emitter.shift.value = shift
+        if self._shiftebeam == MTD_EBEAM_SHIFT:
+            if self._beamshift is not None:
+                shift = tuple(s + c for s, c in zip(shift, self._beamshift))
+            self._emitter.shift.value = shift
 
     def _prepare(self):
         """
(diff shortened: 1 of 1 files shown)


src/odemis/acq/stream/_live.py
```diff
@@ -472,14 +472,16 @@ def _DoPrepare(self):
...
+                    self._emitter.shift.value = (0, 0)
                 shift = FindEbeamCenter(self._ccd, self._detector, self._emitter)
-                logging.debug("Spot shift is %s m,m", shift)
-                shift_clipped = self._emitter.shift.clip(shift)
-                if shift_clipped != shift:
-                    shift = shift_clipped
-                    logging.info("Limiting spot shift to %s m,m due to hardware constraints", shift)
-                self._beamshift = shift
+                if self._shiftebeam == MTD_EBEAM_SHIFT:
+                    logging.debug("Spot shift is %s m,m", shift)
```
pieleric: Let's move this back outside of the `if`.

src/odemis/acq/stream/_live.py
```diff
@@ -430,12 +430,12 @@ def _applyROI(self):
...
-        if (self._shiftebeam == MTD_EBEAM_SHIFT) and (self._beamshift is not None):
-            shift = tuple(s + c for s, c in zip(shift, self._beamshift))
-
-        # always in this order
+        # always in this order: resolution, then shift
         self._emitter.resolution.value = res
-        self._emitter.shift.value = shift
+        if self._shiftebeam == MTD_EBEAM_SHIFT:
+            if self._beamshift is not None:
+                shift = tuple(s + c for s, c in zip(shift, self._beamshift))
```
Kleijwegt: this is always a typle of size 2 isn't it?
wouldn't :
shift = (shift[0] + self._beamshift[0], shift[1] + self._beamshift[1])
be more readable?

src/odemis/acq/stream/_live.py
```diff
@@ -430,12 +430,12 @@ def _applyROI(self):
...
 
-        if (self._shiftebeam == MTD_EBEAM_SHIFT) and (self._beamshift is not None):
-            shift = tuple(s + c for s, c in zip(shift, self._beamshift))
-
-        # always in this order
+        # always in this order: resolution, then shift
         self._emitter.resolution.value = res
-        self._emitter.shift.value = shift
+        if self._shiftebeam == MTD_EBEAM_SHIFT:
+            if self._beamshift is not None:
```
Kleijwegt: What happens if self._beamshift == None? Shouldn't there be an error defined?

src/odemis/acq/stream/_live.py
pieleric: I agree with you. It was a little bit of over-engineering. However, for this change, I didn't want to muddy the water. This line is still correct. So I'm leaving it as-is.
I'll push a second commit, with this clean-up.

src/odemis/acq/stream/_live.py
pieleric: No, beamshift == None is not sign of an error. It just means that there is no beamshift (known). That said, this could be as simply convey by setting beamshift to `(0, 0)`.
=> Also in my clean-up commit.

PR #1384, opened 2020-03-25
Sonic Stage Driver (WIP)

philipwinklerdelmic:
First version of the stage driver. Definitely needs a lot of testing on the real hardware before it can be merged.

I'm not entirely convinced by how I deal with the separate axes. I define them as separate components, which might be unnecessary. What do you think?


scripts/piezomotor_config.py
```diff
@@ -0,0 +1,68 @@
...
+    stage = PMD401("PMD", "stage", "ttyUSB*", {"x": axis}, {"x": 0})
+    print("Please connect the controller board for the x-axis and press enter. " +
+          "Don't connect any of the other boards yet.\n")
+    while True:
+        input()
+        try:
+            res = stage.set_axis_address(1)
+            break
+        except Exception as ex:
+            print("\nConfiguration failed, please try again\nException: %s" % ex)
```
Kleijwegt: Don't you want to use some timeout in here?
Maybe have:
```
startup_time = time.time()
while not timeout(10, startup_time):
```
with:
```
def timeout(a):
  return [True if startup_time - time.time() < -a else False][0]
```

(But than not so ugly)

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,453 @@
...
+WFM_STEPSIZE = 5e-6  # m / step
+USTEPS_PER_WFM = 8192
+ENCODER_STEPSIZE = 1e-9  # m / step
+
+EOL = b'\r'  # 0x0D, carriage return
+
+# Waveforms
+WAVEFORM_RHOMB = 1  # fast max speed
+WAVEFORM_DELTA = 2  # preferred, higher accuracy
+WAVEFORM_PARK = 4  # power off
```
Kleijwegt: Why is power off not zero?

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,453 @@
...
+        except:
+            raise IOError("Failed to find a TMCM controller on port '%s'. "
+                          "Check that the device is turned on and "
+                          "connected to the computer." % (port,))
+
+        return ser
+
+    def set_axis_address(self, new_address, old_address=0):
+        """
+        Set the address of the axis. The factory default is 0 for all boards. Don't use this
```
pieleric: I think this command should take current the axis number, the new address. Or maybe just make it generic setParam(axis, param, value).

scripts/piezomotor_config.py
```diff
@@ -0,0 +1,68 @@
...
+    print("Piezomotor Stage Configuration\n=====================================\n\n")
+
+    axis = Axis()
+    stage = PMD401("PMD", "stage", "ttyUSB*", {"x": axis}, {"x": 0})
+    print("Please connect the controller board for the x-axis and press enter. " +
+          "Don't connect any of the other boards yet.\n")
+    while True:
+        input()
+        try:
+            res = stage.set_axis_address(1)
```
pieleric: The documentation seems to imply that this should be explicitly stored to flash, using Y32.

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,453 @@
...
+    def _doMoveAbs(self, pos):
+        if not self.closed_loop:
+            raise ValueError("Cannot perform absolute move in open loop mode.")
+
+        targets = {}
+        for axis, val in pos.items():
+            encoder_cnts = round(val / ENCODER_STEPSIZE)
+            speed_usteps = round(self._speed / WFM_STEPSIZE * USTEPS_PER_WFM)  # steps / second
+            self._sendCommand(b'X%dT%d,%d' % (self._ax_numbers[axis], encoder_cnts, speed_usteps))
+            targets[axis] = encoder_cnts
```
pieleric: Regularly, it'll need to check the status of the hardware (U0). At least, it should be done after each move (or while waiting for the move to finish).

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,453 @@
...
+        self._executor = CancellableThreadPoolExecutor(max_workers=1)  # one task at a time
+        self._ser_access = threading.RLock()
+
+        self._ax_numbers = ax_numbers
+        self.closed_loop = closed_loop
+        self._speed = speed
+
+        # Connect to hardware
+        self._serial = self._findDevice(port)
+
```
pieleric: You need to provide the `.position` VA, updated at least after each move.
You should also provide a `.referenced` VA, which at least should be empty.

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,453 @@
...
+    The class provides only part of the functionality of the model.Actuator interface. It fully
+    implements the functions moveRel and stop. However, there is no reference function and
+    moveAbs is only possible if we are operating in closed loop, i.e. an encoder (stage position sensor)
+    is present.
+    """
+
+    def __init__(self, name, role, port, axes, ax_numbers, inverted=None, closed_loop=True, speed=1e-6, **kwargs):
+        """
+        :arg ax_numbers (dict: str --> int): axis name --> axis number
```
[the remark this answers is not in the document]
philipwinklerdelmic: @pieleric My idea was to define the axes with all parameters separately in the yaml file, a little bit similar to what we do in the TMCM. This is also what I was trying to do with the awkward `PMAxis` class (which is just and Axis that inherits from HwComponent, so it can be instantiated when starting odemis, at least that's what I was trying to accomplish...).

Do you think this is worth pursuing (maybe in a cleaner way), or is it better to just pass all arguments directly to the PMD401 class and instantiate the axes internally?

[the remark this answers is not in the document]
pieleric: I think I understand what you wanted to do with the PMAxis class, but then it's not the same as with the TMCM class. Anyway, I'd suggest to not follow into having separate component per axis. I'd suggest to go with the technic we've used for the smaract. I think this is the cleanest and most friendly with extension. (ie, a dict {axis name -> dict {property name -> value}})

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,453 @@
...
+
+
+class PMD401(Actuator):
+    """
+    This represents the PMD401 motor controller for the PiezoMotor LEGS motors.
+    The specification for the hardware interface can be found in the document
+     "PiezoMotor_PMD401_Technical_Manual.pdf".
+
+    The class provides only part of the functionality of the model.Actuator interface. It fully
+    implements the functions moveRel and stop. However, there is no reference function and
```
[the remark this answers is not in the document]
philipwinklerdelmic: I didn't really understand the section on index mode in the manual.
> The I command works the same (as J) but will only run if index mode is activated and index has not been detected.

I would have thought that index mode allows you to do an absolute move (go to position x), but it seems to work the same J, so with relative moves. I have no clue what "index has not been detected" means.

It seems to be possible to get the expected position with the read command `N?`, so then for an absolute move we can just subtract this from the target and send the relative move command. However, it is still not clear to me how to implement referencing. Just move in one direction until we get to the limit switch and then reset the index?

Referencing is not needed for closed-loop mode, right?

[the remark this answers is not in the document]
philipwinklerdelmic: Are we ever planning to use open-loop mode anyway? Since we have encoders for all axes, I thought it wasn't really necessary.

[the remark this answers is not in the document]
pieleric: Reading the chapter 4.5 of the documentation, I understand that this is a quite low-level referencing procedure. The `Nx` command set the device for reference search, and the `I` command actually tells the device to search. You have to build your referencing procedure based on this commands. Depending on the hardware the referencing procedure might defer. In particular, the physical position of the index matters. An "easy" hardware is were the index is at the end of the axis (let's say at the lowest position). In this case, you could do something like:
1. start the index mode for search of index
2. jog in reverse direction (because you know that the index is "before" the current position
3. wait until the index is found, or stop after too much time has passed.

That would translate into something like:
`N4`, `I-16`, `N`..... `S`

Note that this is a really basic referencing procedure. You typically want to do a fast search, and then do a second search at slow speed, to locate the index more precisely. Also, if the index is somewhere in the middle, you'll need to handle the fact you might have picked the wrong direction (in which case, after while you'll bump into the limit switch) and in such case, continue by going in the other direction.

After the referencing procedure, typically what we want is that the index position is considered position 0, for the absolute moves. Also, the axis will be physically at this position (but it doesn't really have to be). 

The driver could provide different type of referencing procedure, defined as a parameter per axis (as we do in the TMCM). That said, hopefully for the hardware we have we can use the same procedure for all the axes... so only one procedure would need to be implemented for now.
I'd suggest to start roughly implementing the simple procedure that I've described, and once the hardware is sufficiently define, adjust the code.

philipwinklerdelmic:
I updated the PR, though I still have to go through it and review it myself to see if I missed anything. @pieleric Could you at least have a look at the init and how I deal with the axes parameters. Is that more or less what you had in mind?

The pmconfig script is far from being ready. We still need to figure out which parameters need configuration. I think this is best done when we have the hardware.

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,681 @@
...
+from threading import Thread
+import copy
+from odemis.model import HwError
+from odemis.util.driver import getSerialDriver
+
+BAUDRATE = 115200
+# TODO: Is this always the same for all encoders? Otherwise, it needs to be a parameters in the yaml file.
+WFM_STEPSIZE = 5e-6  # m / step
+USTEPS_PER_WFM = 8192
+ENCODER_STEPSIZE = 1e-9  # m / step
```
philipwinklerdelmic, line 42: The encoder stepsize probably needs to be configured manually (run a few wfm steps and get the count, then divide) and put in the yaml file as well. I was hoping the stepsizes are the same for each encoder of the same type, but they probably vary a bit for every model. Maybe we should test this first on the hardware.

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,681 @@
...
+     "PiezoMotor_PMD401_Technical_Manual.pdf".
+
+    The class provides only part of the functionality of the model.Actuator interface. It fully
+    implements the functions moveRel and stop. However, there is no reference function and
+    moveAbs is only possible if we are operating in closed loop, i.e. an encoder (stage position sensor)
+    is present.
+    """
+
+    def __init__(self, name, role, port, axes, inverted=None, **kwargs):
+        """
```
philipwinklerdelmic, line 84: TODO: update docstring

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,681 @@
...
+            except KeyError:
+                # m if linear, "rad" otherwise
+                self._speed = DEFAULT_AXIS_SPEED
+                logging.info("Axis %s was not given a speed value. Assuming %s", axis_name, self._speed)
+
+            # Axis mode (closed loop/ open loop)
+            try:
+                mode = axis_par['mode']
+            except KeyError:
+                logging.info("Axis mode (closed/open loop) not specified for axis %s. Assuming %s",
```
philipwinklerdelmic, line 146: put default value for axis mode

src/odemis/driver/piezomotor.py
philipwinklerdelmic: actually, no, probably that's given by the encoder (need to check which encoder model we're using).

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,453 @@
...
+import os
+import threading
+import time
+import serial
+from threading import Thread
+import copy
+from odemis.model import Axis, HwComponent
+
+BAUDRATE = 115200
+WFM_STEPSIZE = 5e-6  # m / step
```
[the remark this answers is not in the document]
philipwinklerdelmic: I'm wondering whether we need to have the accurate value in the code at all. We need to configure the stepsize (encoder counts --> motor counts) on the hardware and this value will be stored in EEPROM. In case of closed loop control, we will only pass the encoder counts and the conversion will be done internally. On the other hand, if we use open loop control, the amount of steps is approximate anyway, so an approximate stepsize might be good enough. This would save quite a lot of trouble, because we don't need to set this parameter accurately in every yaml file (which might be prone to mistakes).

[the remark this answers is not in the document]
pieleric: I think that for now, all the hardware will be closed-loop, so we just don't need to worry about open-loop. For closed-loop, this info will indeed probably not be needed, as it's just the encoder which returns the information.

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,681 @@
...
+            # it now should be accessible again
+            self.state._set_value(model.ST_RUNNING, force_write=True)
+            logging.info("Recovered device on port %s", self._port)
+        finally:
+            self._recovering = False
+
+    def terminate(self):
+        # called in __del__ (see HwComponent)
+        self.setWaveform(WAVEFORM_PARK)  # power off
+        self._serial.close()
```
Kleijwegt, line 426: Maybe it is nice to add a logging.info message holding "succefully terminated connection with the sonic stage driver"?

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,681 @@
...
+                if t is None:
+                    raise ValueError("No target provided for closed-loop move on axis %s." % ax_name)
+                query_cmd = b'X%dE' % self._axis_map[ax_name]
+                resp_finished = b'%d' % t  # wait till encoder position reports target position
+            else:
+                query_cmd = b'X%dJ' % self._axis_map[ax_name]
+                resp_finished = b'1'  # 1 if finished, 0 while moving
+
+            moving = True
+            end_time = time.time() + 5  # 5 s until timeout
```
pieleric, line 284: That timeout is really crude. In addition, 5 s could be really short for some long moves, with closed loop. If you don't know better use 30s.
See the tmcm driver. Based on the distance and speed of each axis, it computes a rough estimation of how long it takes for each axis, and use the longest time.

src/odemis/driver/test/piezomotor_test.py
```diff
@@ -0,0 +1,109 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+'''
+Created on 1 mar 2012
+
+@author: Éric Piel
+Testing class for pi.py .
+
+Copyright © 2012 Éric Piel, Delmic
+
```
pieleric, line 10: Please update this header ;-)

src/odemis/driver/test/piezomotor_test.py
```diff
@@ -0,0 +1,109 @@
...
+logging.getLogger().setLevel(logging.DEBUG)
+
+# Export TEST_NOHW=1 to force using only the simulator and skipping test cases
+# needing real hardware
+TEST_NOHW = (os.environ.get("TEST_NOHW", 0) != 0)  # Default to Hw testing
+
+ax1 = Axis(range=[0, 100e-3])
+ax2 = Axis(range=[0, 100e-3])
+ax3 = Axis(range=[0, 10e-3])
+axes = {'x': ax1, 'y': ax2, 'z': ax3}
```
pieleric, line 42: That doesn't look used.

src/odemis/driver/test/piezomotor_test.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+class TestPMD401(unittest.TestCase):
+    """
+    Test directly the PIRedStone class.
```
pieleric, line 47: Please update this docstring.

util/pmconfig.py
```diff
@@ -0,0 +1,98 @@
...
+                                     description="Read/write parameters in a PM stage controller")
+
+    parser.add_argument("--log-level", dest="loglev", metavar="<level>", type=int,
+                        default=1, help="set verbosity level (0-2, default = 1)")
+
+    parser.add_argument("--set-address", dest="address", metavar="<level>", type=int,
+                        help="set address of stage controller.")
+
+    parser.add_argument("--auto-config", dest="autoconf", metavar="<level>", type=int,
+                        help="Automatically configure axis.")
```
pieleric, line 65: "axis" -> "encoder's step per count", to make it clearer.

util/pmconfig.py
```diff
@@ -0,0 +1,98 @@
...
+    parser = argparse.ArgumentParser(prog="pmconfig",
+                                     description="Read/write parameters in a PM stage controller")
+
+    parser.add_argument("--log-level", dest="loglev", metavar="<level>", type=int,
+                        default=1, help="set verbosity level (0-2, default = 1)")
+
+    parser.add_argument("--set-address", dest="address", metavar="<level>", type=int,
+                        help="set address of stage controller.")
+
+    parser.add_argument("--auto-config", dest="autoconf", metavar="<level>", type=int,
```
pieleric, line 64: Drop the "metavar".

util/pmconfig.py
```diff
@@ -0,0 +1,98 @@
...
+
+    # Set up logging before everything else
+    if options.loglev < 0:
+        logging.error("Log-level must be positive.")
+        return 127
+    loglev_names = (logging.WARNING, logging.INFO, logging.DEBUG)
+    loglev = loglev_names[min(len(loglev_names) - 1, options.loglev)]
+    logging.getLogger().setLevel(loglev)
+
+    port = "/dev/ttyUSB*"
```
pieleric, line 82: That should be an argument too. The extra advantage is that it allows you to test your script by passing `--port /dev/fake`.

util/pmconfig.py
```diff
@@ -0,0 +1,98 @@
...
+    axes = {"x": {"axis_number": 0,
```
pieleric, line 83: Axis number should be an argument of the script, with 0 as default value.
For auto conf, it could then become a flag (cf "store_true"). You'd write a command like `pmconfig --port /dev/ttyPM0 --address 1 --auto-config`

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,681 @@
...
+                mode = axis_par['mode']
+            except KeyError:
+                logging.info("Axis mode (closed/open loop) not specified for axis %s. Assuming %s",
+                             axis_name, mode)
+            if mode not in (OPEN_LOOP, CLOSED_LOOP):
+                raise ValueError("Axis mode %s not available, needs to be 0 (closed loop) " % mode +
+                                 "or 1 (open loop)")
+            self._axis_modes[axis_name] = mode
+
+            ad = model.Axis(canAbs=True, unit=axis_unit, range=axis_range)
```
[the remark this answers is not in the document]
philipwinklerdelmic: I tried to implement the moveAbs functionality even in open loop (by subtracting the current position from the requested position and calling a relative move). Should I drop this and put canAbs to false?

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,681 @@
...
+                self._speed = DEFAULT_AXIS_SPEED
+                logging.info("Axis %s was not given a speed value. Assuming %s", axis_name, self._speed)
+
+            # Axis mode (closed loop/ open loop)
+            try:
+                mode = axis_par['mode']
+            except KeyError:
+                logging.info("Axis mode (closed/open loop) not specified for axis %s. Assuming %s",
+                             axis_name, mode)
+            if mode not in (OPEN_LOOP, CLOSED_LOOP):
```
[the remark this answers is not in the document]
philipwinklerdelmic: That makes sense. What about the internal dictionary, do you think self._axis_modes should be self._closed_loop as well?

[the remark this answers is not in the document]
pieleric: Yes, it'll also make the code a little bit more easy to read (because you can write `if self._closed_loop[axis]` .

src/odemis/driver/piezomotor.py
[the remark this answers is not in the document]
pieleric: The meaning of "canAbs" is essentially to indicate that the axis uses closed-loop. It indicates the precision. 
Still whether it's closed-loop or open-loop, moveAbs() must be implemented. How it's implemented doesn't really matter, but please pick the way which is giving the best expected behaviour. So I'd suggest in this case, if the hardware still have some "absolute move" in open-loop directly use this function.

philipwinklerdelmic:
Open issues:
* install/linux/lib/udev/rules.d/.
* wfm encoder stepsize (need to check everything related to encoder step/ wfm step/ meter conversion)

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,733 @@
...
+        """
+        :param axis (int): axis number
+        :param target (float): target position for axes
+        :returns (bool): True if moving, False otherwise
+        """
+        resp = self._sendCommand(b'X%dE' % axis)
+        if abs(float(target) - float(resp)) < 1e-9:  # 1 nm accuracy
+            return True
+        else:
+            return False
```
pieleric, line 374: I don't think you can check that it's moving this way. Firstly the accuracy is far too high (it'll stop more likely before being within 1 nm), but more importantly there are other reason that the move can stop (eg, an error). Looking at the doc, there is no very clear way to check, but I'd say that looking at U0 targetReached will be a good start. You probably also want to check it's still in targetMode (and that targetLimit is not reached). Maybe just looking at the running status is enough. Another possibility again, could be to use Y23.  Most likely, the only way to be really sure is to test.

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,733 @@
...
+
+    def getSerialNumber(self, axis):
+        """
+        :returns (str): serial number
+        """
+        return self._sendCommand(b'X%dY42' % axis)
+
+    def getEncoderPosition(self, axis):
+        """
+        :returns (dict: str --> int): current position of axes as reported by encoders
```
pieleric, line 333: The code seems to imply it'll return a single float. Also, please indicate that it's already converted to the "user unit".

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,733 @@
...
+            if status[0] == 8:
+                raise PMDError(1, "Communication Error (wrong baudrate, data collision, or buffer overflow)")
+            elif status[0] == 4:
+                raise PMDError(2, "Encoder error(serial communication or reported error from serial encoder)")
+            elif status[0] == 2:
+                raise PMDError(3, "Supply voltage or motor fault was detected.")
+            elif status[1] == 1:
+                raise PMDError(4, "Command timeout occurred or a syntax error was detected when response was not allowed.")
+            elif status[1] == 8:
+                raise PMDError(5, "Power-on/reset has occurred.")
```
pieleric, line 304: You need to use bit comparison, not equality, because there might be several errors at the same time. => Replace `==` by `&`.

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,733 @@
...
+        :arg wf (waveform): waveform to set
+        """
+        if wf not in (WAVEFORM_DELTA, WAVEFORM_RHOMB, WAVEFORM_PARK):
+            raise ValueError("wf %s not a valid waveform" % wf)
+
+        self._sendCommand(b'X%dM%d' % (axis, wf))
+
+    def getStatus(self, axis):
+        """
+        :returns (str): 4-number status code
```
pieleric, line 348: Please, don't return a string, but 4 ints, which is easier to handle later (d1 to d4).

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,733 @@
...
+        self._waitEndMotion(targets)
+        self._updatePosition()
+
+    def _waitEndMotion(self, targets):
+        """
+        Wait until move is done
+        :arg targets (dict: str --> int): target (for closed-loop), None for open loop
+        """
+        # Expected time for move
+        move_length = sum([abs(self.position[ax] - targets[ax]) for ax in targets.keys()])
```
pieleric, line 272: You can drop the `[` as the sum accepts directly an iterator. You could also use `.items()` instead of `.keys()` to avoid the duplication of `targets`.
BTW, I guess that all the axes move simultaneously? In that case, it's not "sum" but "max" that you need to use.

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,733 @@
...
+        self._check_hw_error()
+        targets = {}
+        for axis, val in pos.items():
+            if self._closed_loop[axis]:
+                encoder_cnts = round(val / ENCODER_STEPSIZE)
+                speed_usteps = round(self._speed / self._wfm_stepsizes[axis])  # wfm steps / second
+                self._sendCommand(b'X%dT%d,%d' % (self._axis_map[axis], encoder_cnts, speed_usteps))
+                targets[axis] = encoder_cnts
+            else:
+                self._updatePosition()
```
pieleric, line 237: Updating the position of *every* axis before moving one axis is a little too much. Either read only that axis position, or, just trust the latest position read. In open loop, there is no reason that the position changes without requesting a move, so the latest known position should always be correct.

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,733 @@
...
+        #  the other direction.
+
+    def _doMoveAbs(self, pos):
+        self._check_hw_error()
+        targets = {}
+        for axis, val in pos.items():
+            if self._closed_loop[axis]:
+                encoder_cnts = round(val / ENCODER_STEPSIZE)
+                speed_usteps = round(self._speed / self._wfm_stepsizes[axis])  # wfm steps / second
+                self._sendCommand(b'X%dT%d,%d' % (self._axis_map[axis], encoder_cnts, speed_usteps))
```
pieleric, line 234: Please, for each of this 4 commands, create a dedicated function. It'll be easier to read.

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,733 @@
...
+    def reference(self, axes):
+        f = self._executor.submit(self._doReference, axes)
+        return f
+
+    def _doReference(self, axes):
+        self._check_hw_error()
+        for axname in axes:
+            self.startIndexMode(self._axis_map[axname])
+            self.moveToIndex(self._axis_map[axname])
+            if 'index' not in self.getIndexStatus(self._axis_map[axname]):
```
pieleric, line 216: You'll probably need to wait (with a timeout) here, as moveToIndex is not blocking. Typically a reference procedure can takes several seconds.

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,733 @@
...
+        return f
+
+    def moveAbs(self, pos):
+        self._checkMoveAbs(pos)
+        pos = self._applyInversion(pos)
+        f = self._executor.submit(self._doMoveAbs, pos)
+        return f
+
+    def reference(self, axes):
+        f = self._executor.submit(self._doReference, axes)
```
pieleric, line 208: Call `_checkReference()` first.

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,733 @@
...
+            if b"!" in resp:
+                raise PMDError(0, resp)
+            # Format:
+            #    * for query with response: <cmd>:<ret><EOL> (will return <ret>)
+            #    * for set command without response: <cmd><EOL> (will return b"")
+            return resp[len(cmd) + 1 - len(EOL):-len(EOL)].decode("latin1")
+
+    def _tryRecover(self):
+        self._recovering = True
+        self.state._set_value(HwError("USB connection lost"), force_write=True)
```
pieleric, line 452: As this message is shown to the user (nowadays), it's good to let the user she/he shouldn't worry yet. So I'd recommend something like "Connection lost, reconnecting...".

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,733 @@
...
+                timeout=0.3,  # s
+            )
+        except IOError:
+            raise HwError("Failed to find a PMD controller on port '%s'. "
+                          "Check that the device is turned on and "
+                          "connected to the computer." % (port,))
+
+        return ser
+
+    def setAxisAddress(self, current_address, new_address):
```
pieleric, line 553: Please move these 4 methods next to the other "wrapper" methods.

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,733 @@
...
+        self._output_buf = ""
+
+        self.waveform = {1: WAVEFORM_PARK, 2: WAVEFORM_PARK, 3: WAVEFORM_PARK}
+        self.target_pos = {1: 0, 2: 0, 3: 0}
+        self.current_pos = {1: 0, 2: 0, 3: 0}
+        self.speed = 0
+        self.is_moving = False
+
+    def write(self, data):
+        self._input_buf += data.decode('utf-8')
```
pieleric, line 589: I doubt the actual hardware handles UTF8 ;-) Let's just write "ascii".

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,733 @@
...
+        self._parseMessage(msg)
+
+    def read(self, size=1):
+        ret = self._output_buf[:size]
+        self._output_buf = self._output_buf[len(ret):]
+
+        if len(ret) < size:
+            # simulate timeout
+            time.sleep(self.timeout)
+        return ret.encode('utf-8')
```
pieleric, line 605: "ascii"

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,733 @@
...
+        self._sendCommand("X%dY25" % axis)
+
+    def writeParamsToFlash(self, axis):
+        self._sendCommand("X%dY32" % axis)
+
+    def setParam(self, axis, param, value):
+        self._sendCommand("X%dY%d,%d" % (axis, param, value))
+
+class PMDSimulator(object):
+
```
pieleric, line 576: Please indicate that this simulates 3 axes, with addresses 1,2, and 3.

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,733 @@
...
+    def isMovingOpenLoop(self, axis):
+        """
+        :param axis (int): axis number
+        :returns (bool): True if moving, False otherwise
+        """
+        resp = self._sendCommand(b'X%dJ' % axis)
+        if int(resp) == 0:  # 1 if finished, 0 while moving
+            return True
+        else:
+            return False
```
pieleric, line 385: You can write it `return int(resp) == 0`.

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,733 @@
...
+    implements the functions moveRel and stop. However, there is no reference function and
+    moveAbs is only possible if we are operating in closed loop, i.e. an encoder (stage position sensor)
+    is present.
+    """
+
+    def __init__(self, name, role, port, axes, inverted=None, **kwargs):
+        """
+        :param axes (dict: {"x", "y", "z"} --> dict): axis name --> axis parameters (axis_number, range, unit, closed_loop, speed, wfm_stepsize)
+            axis_number (0 <= int <= 127), required (typically 1-3 for x-z)
+            wfm stepsize (int), stepsize (needs to be
```
pieleric, line 83: That docstring needs some love ;-) I don't think it's an int. Also indicate what "wfm" stands for ("waveform").

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,733 @@
...
+    The class provides only part of the functionality of the model.Actuator interface. It fully
+    implements the functions moveRel and stop. However, there is no reference function and
+    moveAbs is only possible if we are operating in closed loop, i.e. an encoder (stage position sensor)
+    is present.
+    """
+
+    def __init__(self, name, role, port, axes, inverted=None, **kwargs):
+        """
+        :param axes (dict: {"x", "y", "z"} --> dict): axis name --> axis parameters (axis_number, range, unit, closed_loop, speed, wfm_stepsize)
+            axis_number (0 <= int <= 127), required (typically 1-3 for x-z)
```
pieleric, line 82: The name is a little bit verbose. I don't think you need to indicate it's about the "axis". Maybe just call it "number" or "address".

src/odemis/driver/test/piezomotor_test.py
```diff
@@ -0,0 +1,102 @@
...
+from odemis.driver.piezomotor import PMD401Bus
+from collections import deque
+
+logging.getLogger().setLevel(logging.DEBUG)
+
+# Export TEST_NOHW=1 to force using only the simulator and skipping test cases
+# needing real hardware
+TEST_NOHW = (os.environ.get("TEST_NOHW", 0) != 0)  # Default to Hw testing
+
+class TestPMD401(unittest.TestCase):
```
pieleric, line 38: You need to extend the test cases. See the other actuator tests for building up you imagination (eg, tmcm_test, or actuator_test). At least, you want to check that .position is correctly updated, that referenced is properly updated, and that cancelling a move works as expected.

With the simulator, it'd be good to also check both the closed loop and open loop modes.

src/odemis/driver/test/piezomotor_test.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+
+    def setUp(self):
+        if TEST_NOHW:
+            port = "/dev/fake"
+        else:
+            port = "/dev/ttyUSB*"
+
+        axes = {"x": {"axis_number": 1,
+                      "mode": 1,
```
pieleric, line 50: I don't think "mode" does anything. It could be good to raise an error at instantiation if this happens (to do so easily, use `.pop()` on the dict, and raise an error if it's not empty at the end of the parsing).

src/odemis/driver/piezomotor.py
philipwinklerdelmic: Do we have some code in odemis that converts a si-value to the axis.unit? I have the value in meter, how to I get to whatever the user specified (nm, µm, etc.)? Check all possible cases of the unit that I can think of in if-statements?

src/odemis/driver/piezomotor.py
philipwinklerdelmic: Do you mean the 4 commands 1) encoder steps closed loop, 2) encoder steps open loop, 3) usteps open loop, 4) usteps closed loop?

src/odemis/driver/piezomotor.py
pieleric: We have some code to do that... but that should be only for the GUI. The driver shouldn't try to do anything too clever, it uses the stepsize to convert from the internal unit to the "user unit". That's it. Up to the "user" (actually, the installation engineer) to pick the right unit. Also, by convention the unit never contains SI-prefix. So you'd essentially only have "m" or "rad".

src/odemis/driver/piezomotor.py
pieleric: Yes.

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,733 @@
...
+        self.position = model.VigilantAttribute({}, unit="m", readonly=True)
+        self.referenced = model.VigilantAttribute({}, unit="m", readonly=True)
+
+        # Configuration
+        for axis in self._axis_map.values():
+            self.setWaveform(axis, WAVEFORM_DELTA)
+
+        driver_name = getSerialDriver(self._port)
+        self._swVersion = "serial driver: %s" % (driver_name,)
+
```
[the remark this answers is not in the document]
philipwinklerdelmic: I thought we only need referencing for open-loop? Doesn't the encoder deal with the absolute position in closed-loop?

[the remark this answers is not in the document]
pieleric: That's kind of the opposite. Note that in theory, referencing is orthogonal to closed-/open-loop. Referencing  just means that you're able to have a fixed origin, by determining the position of a "reference switch" (or "index").
In practice, however, with open-loop this has little use because after a move, due to imprecision, this origin will shift randomly a little bit. That's why referencing is typically used *only* with closed-loop systems.

src/odemis/driver/test/piezomotor_test.py
```diff
@@ -0,0 +1,111 @@
...
+        self.stage.moveRel(move)
+        # don't wait
+        move = {'x': -0.01}
+        self.stage.moveRel(move)
+        move = {'x': 0}
+        self.stage.moveAbs(move)
+
+        self.stage.stop()
+        self.assertEqual(self.stage._executor._queue, deque([]))
+
```
pieleric, line 108: You should test referencing too.

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,871 @@
...
+        :param limit_type: (0 <= int <= 3) 0 no limit, 1 active high, 2 active low
+        """
+        self._sendCommand(b'X%dY2,%d' % (axis, limit_type))
+
+    def getEncoderPosition(self, axis):
+        """
+        :returns (float): current position of the axis as reported by encoders (in m)
+        """
+        axname = [name for name, num in self._axis_map.items() if num == axis][0]  # get axis name from number
+        return int(self._sendCommand(b'X%dE' % axis)) / self._counts_per_meter[axname]
```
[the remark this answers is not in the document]
philipwinklerdelmic: There are other instances where it is easier to call it with the axis name (e.g. `doMoveRel`). Also the other attributes work with the axis name and not the number, so I thought it would be easier and more consistent to leave it like this.

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,871 @@
...
+EOL = b'\r'  # 0x0D, carriage return
+sEOL = '\r'  # string version of EOL
+
+# Waveforms
+WAVEFORM_RHOMB = 1  # fast max speed
+WAVEFORM_DELTA = 2  # preferred, higher accuracy
+WAVEFORM_PARK = 4  # power off
+
+# Length of the rod available for moves
+STROKE_RANGE = 23.1e-3  # m
```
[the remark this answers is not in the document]
philipwinklerdelmic: This is the exact length of axis as specified in the datasheet of the motor. If we use a different kind of axis later, we should change it. I'm not sure if we're going to axes of different lengths. If not, it would be better to just put the precise value here, so we don't have to put the same value in all the yaml files. A move of 10 cm on the current axis will completely move the motor off the rod.

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,871 @@
...
+
+            if 'range' in axis_par.keys():
+                axis_range = axis_par['range']
+            else:
+                axis_range = (0, STROKE_RANGE)
+                logging.info("Axis %s has no range. Assuming %s", axis_name, axis_range)
+
+            if 'spc' in axis_par.keys():
+                self._steps_per_count[axis_name] = axis_par['spc']  # approximately 5e-6 m / step
+                self._steps_per_meter[axis_name] = axis_par['spc'] * self._counts_per_meter[axis_name]
```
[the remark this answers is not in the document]
philipwinklerdelmic: My idea was to still have the parameter in the yaml file in case we want to override the value in flash memory. Maybe it's not necessary. What do you think?

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,871 @@
...
+            logging.debug("Finished referencing axis %s." % axname)
+            self.stopAxis(self._axis_map[axname])
+            self.referenced.value[axname] = True
+
+        # TODO: Note that this is a really basic referencing procedure. You typically want to do a fast search,
+        #  and then do a second search at slow speed, to locate the index more precisely. Also, if the index is
+        #  somewhere in the middle, you'll need to handle the fact you might have picked the wrong direction
+        #  (in which case, after while you'll bump into the limit switch) and in such case, continue by going in
+        #  the other direction.
+        #  Referencing currently doesn't work if we already start at the target position.
```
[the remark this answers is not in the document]
philipwinklerdelmic: I'm not really sure how to fix it properly. I now make a very small move away from the limit (10 µm). It works, but it also seems a bit ugly.

src/odemis/driver/piezomotor.py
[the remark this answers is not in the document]
pieleric: I'm only talking of the computation of `_steps_per_meter`. The "spc" parameter is fine.

src/odemis/driver/piezomotor.py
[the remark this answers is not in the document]
pieleric: Please discuss with Andries how to implement a reliable referencing procedure.

src/odemis/driver/piezomotor.py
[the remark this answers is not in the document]
philipwinklerdelmic: Ok, but then we need to have this code here. Later, I check if _steps_per_count is None and if so, I read the value from flash. If it's already set, I don't read from flash.

src/odemis/driver/piezomotor.py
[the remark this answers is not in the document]
philipwinklerdelmic: Andries thinks the referencing procedure is fine. We're going to have to test again once we have the real hardware (right now we only have the starter kit, might be a bit different).

src/odemis/driver/piezomotor.py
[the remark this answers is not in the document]
pieleric: As said in the previous comment 😉... I'm only talking of the computation of _steps_per_meter .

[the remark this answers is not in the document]
philipwinklerdelmic: ah, sorry, I didn't read carefully :)

src/odemis/driver/piezomotor.py
```diff
@@ -0,0 +1,871 @@
...
+    def runRelTargetMove(self, axis, encoder_cnts):
+        """
+
+        """
+        # There are two possibilities: move relative to current position (XC) and move relative to
+        # target position (XR). We are moving relative to the current position (might be more intuitive
+        # if something goes wrong and we're stuck in the wrong position).
+        self._sendCommand(b'X%dC%d,%d' % (axis, encoder_cnts, self._speed_steps['x']))
+
+    def runMotorJog(self, axis, wfm_steps, usteps, speed):
```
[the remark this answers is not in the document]
philipwinklerdelmic: I started changing it, but then it seemed more confusing to me. As a `steps` argument you expect an integer, if you don't read the docstring carefully, you don't know that you have microstep precision. On the other hand, it's annoying to do the conversion every time. Not sure what the best way to handle it is, I guess I'll just leave it as is.

philipwinklerdelmic:
Pushed new changes, ready to merge?

src/odemis/driver/piezomotor.py
```diff
@@ -502,14 +542,15 @@ def moveToIndex(self, axis):
...
+        # Move towards fixed end (negative direction)
+        maxdist = int(-self._axes[axname].range[1] * self._counts_per_meter[axname] * DEFAULT_SPC)  # complete rodlength
+        self._sendCommand(b'X%dI%d,0,%d' % (axis, maxdist, self._speed_steps[axname]))
 
     def getIndexStatus(self, axis):
         """
         Returns a description of the index status.
         :returns (tuple of 4):
-            string! mode (int): index mode (1 if position has been reset at index)
+            mode (int): index mode (if position has been reset at index, can be)
```
pieleric, line 553: It feels like the sentence is not complete. Can be...

src/odemis/driver/piezomotor.py
```diff
@@ -470,15 +509,16 @@ def getStatus(self, axis):
...
         """
         :param axis: (int) axis number
         :param target:  (float) target position for axes
-        :returns (bool): True if moving, False otherwise
+        :returns: (bool) True if moving, False otherwise
         """
-        axname = [name for name, num in self._axis_map.items() if num == axis][0]  # get axis name from number
-        position = float(self._sendCommand(b'X%dE' % axis)) / self._counts_per_meter[axname]
-        if abs(position - target) < 2 / self._counts_per_meter[axname]:  # 1 encoder count precision (difference less than 2 counts)
+        # Check d3 (third status value) bit 4 (targetLimit) and bit 1 (targetReached)
```
pieleric, line 518: It's bit *values* 4 (=2²) & 1(=2⁰). The bit number are 2 and 0. Please update this comment.
In practice, that means you could even write `if d3 & 5`, or more clearly `if d3 & 0b0101`.

util/pmconfig.py
```diff
@@ -41,24 +46,42 @@ def set_address(old_address, new_address, stage):
...
+    time.sleep(3)  # wait for autoconf to finish
     stage.writeParamsToFlash(address)
     logging.info("SPC value %s saved to flash.", stage.readParam(address, 11))
 
 
-def manual_conf(address, stage):
+def spc_manual_conf(address, stage, steps=200):
+    """
+    Set spc parameter with manual configuration procedure.
+    This procedure uses a larger move to determine the parameter and is move accurate than the
```
pieleric, line 63: "is move" -> "is more"

util/pmconfig.py
```diff
@@ -41,24 +46,42 @@ def set_address(old_address, new_address, stage):
...
+    automatic configuration procedure.
+    """
     axname = [name for name, num in stage._axis_map.items() if num == address][0]  # get axis name from number
+
+    # Move to a position in the middle of the axis
+    stage.reference({'x'}).result()
+    stage.moveAbsSync({'x': 0.005})
```
pieleric, line 70: That's really ad-hoc to this hardware. I think that alternatively, you could do nothing, and expect the user to have move to a good position already. For now, I'm fine with this, if this simplifies the usage, but please put a big TODO or WARNING here.
Also, if the device is not yet configured, I wonder how well the referencing and move by 5 mm will be reliable.

util/pmconfig.py
```diff
@@ -41,24 +46,42 @@ def set_address(old_address, new_address, stage):
...
     stage._updatePosition()
     startpos = stage.position.value[axname]
-    stage.runMotorJog(address, -200, 0, 200)
+    stage.runMotorJog(address, -steps, 0, stage._speed_steps[axname])
     time.sleep(3)
     stage._updatePosition()
     endpos = stage.position.value['x']
-    spc = 200 / abs(endpos - startpos) / stage._counts_per_meter[axname]
+    encoder_cnts = (abs(endpos - startpos) * stage._counts_per_meter[axname])
+    spc = steps / encoder_cnts
```
pieleric, line 81: Log this value with `log.info()`

src/odemis/driver/piezomotor.py
```diff
@@ -532,21 +577,22 @@ def getIndexStatus(self, axis):
                 indexed = False
             return mode, position, logged, indexed
         except Exception as ex:
-            logging.error("Failed to parse index status %s: %s" % (ret, ex))
+            logging.exception("Failed to parse index status %s: %s" % (ret, ex))
```
pieleric, line 580: `logging.exception()` automatically logs the exception, so you don't need to put it in the message. 
BTW, it's recommended to pass the variables as arguments when using `logging`, though it doesn't really matter.

PR #1386, opened 2020-03-27
acq stream test: test_arpolarimetry warn if arpolarimetry module is missing

pieleric:
Without this module, the projection cannot be computed, so it'll
obviously fail. However, in such case it fails with such generic error:
  File "/home/philip/development/odemis/src/odemis/acq/test/stream_test.py", line 4287, in test_arpolarimetry
    assert e.is_set()
AssertionError

=> Detect the module is missing and indicate how to install it.

### DIFF
--- a/src/odemis/acq/test/stream_test.py
+++ b/src/odemis/acq/test/stream_test.py
@@ -4261,6 +4261,12 @@ class StaticStreamsTestCase(unittest.TestCase):
     def test_arpolarimetry(self):
         """Test StaticARStream with ARPolarimetryProjection projection."""
 
+        # Check whether the arpolarimetry package is available
+        try:
+            import arpolarimetry
+        except ImportError:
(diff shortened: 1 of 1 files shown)


src/odemis/acq/test/stream_test.py
```diff
@@ -4261,6 +4261,12 @@ def on_im(im):
     def test_arpolarimetry(self):
         """Test StaticARStream with ARPolarimetryProjection projection."""
 
+        # Check whether the arpolarimetry package is available
+        try:
+            import arpolarimetryDEBUG
```
Sabrina85: Why `DEBUG`? I only have `import arpolarimetry` in my code in Odemis. Did something change?

pieleric: Sorry about this. I had added it temporarily to test the behaviour works fine when the module is missing (bogus name).
I've just pushed a new version with the proper module name back.

PR #1389, opened 2020-03-31
gui comp overlay LabelGadget: fix large movement when draging from the text

pieleric:
The dragging of the end or start of the line used to always associate the cursor
position to the line position. On the ruler (a simple line), it doesn't
differ too much from the original position of the cursor, and that's
kind of expected anyway.

On the label, if someone starts the dragging from the text area,
this can mean a shift of 100 or 200 px (if the text is long).
When trying to adjust slightly the position of the label, or trying to
click on the text but moving by 1 or 2 px, it's quite irritating.
=> Do not use directly the cursor position as a new position, but
shift that original position by the amount of dragging.
This way, a move of 1 px by the mouse always corresponds to the move
of one pixel in the end point.

Also fix a tiny bug related to the selection. Now that the tools were
ordered in the reverse order, this loop also needs to be inverted.

### DIFF
--- a/src/odemis/gui/comp/overlay/world.py
+++ b/src/odemis/gui/comp/overlay/world.py
@@ -959,6 +959,10 @@ class GenericGadgetLine(with_metaclass(ABCMeta, GadgetToolInterface)):
             vpos: the view coordinates of the mouse cursor once left click mouse event is fired
         """
         self.drag_v_start_pos = self.drag_v_end_pos = Vec(vpos)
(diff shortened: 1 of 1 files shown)


pieleric:
Yes, indeed, for the start point, it's not really necessary. I actually hesitated, but then thought it'd be a little less confusing in the code if both are handled the same way. That said, you could say that the ruler is handled differently, so it's also a little confusing...

PR #1390, opened 2020-04-01
Weaver Brightness Adjustment

philipwinklerdelmic:


src/odemis/acq/stitching/_weaver.py
```diff
@@ -30,24 +31,22 @@
 # directly copy the image already transformed.
 # TODO: handle higher dimensions by just copying them as-is
 
-class CollageWeaver(object):
+class Weaver(object):
```
pieleric, line 34: If you use `@abstractmethod`, you should mark this class as a metaclass.

src/odemis/acq/stitching/_weaver.py
```diff
@@ -30,24 +31,22 @@
...
-    border (None or value): if there is a value, it's used around each image, to
-     highlight the position
-    return (2D DataArray): same dtype as the tiles, with shape corresponding to
-      the bounding box.
+    Abstract class representing a weaver.
+    A weaver assembles a set of small images with MD_POS metadata (tiles) into one large image.
     """
 
-    def __init__(self):
+    def __init__(self, adjust_brt=True):
```
pieleric, line 40: Let's just name this `adjust_brightness`, it's clearer.

src/odemis/acq/stitching/_weaver.py
```diff
@@ -57,12 +56,45 @@ def addTile(self, tile):
...
+        """
+        Adjusts the brightness of a list of tiles, so they all have the same mean value.
+        tiles (2D DataArray): input tiles
+        return (2D DataArray): tiles with adjusted brightness
+        """
+        im_brt = numpy.mean(tiles)
+        for tile in tiles:
+            tile_brt = numpy.mean(tile)
+            diff = numpy.array(im_brt - tile_brt).astype(tile.dtype)  # single value, use array for type cast
+            tile += diff
```
pieleric, line 77: There are two issues with this code:
* The original data (`tile`) is changed. In general, you don't want such side effect in any function. It's a little annoying here because that means you'll essentially need 2x more memory. You could avoid this though by adjusting the brightness of the tile a little later, only at the very moment it's used in the copy.
* When summing, you can cause over- and under-flows. You want to do clipping. However I don't know the best way to handle clipping generically in numpy. Probably you could just handle it for integers, and for floats it'll typically work fine. For inspiration, see odemis.util.img.Subtract() and odemis.driver.spectrometer l319. Maybe if you write a nice function, you can put it in odemis.util.img too.

src/odemis/acq/stitching/_weaver.py
```diff
@@ -57,12 +56,45 @@ def addTile(self, tile):
...
+        pass
 
-        tiles = self.tiles
+    def _adjust_brightness(self, tiles):
+        """
+        Adjusts the brightness of a list of tiles, so they all have the same mean value.
+        tiles (2D DataArray): input tiles
+        return (2D DataArray): tiles with adjusted brightness
+        """
+        im_brt = numpy.mean(tiles)
```
pieleric, line 73: Please add a comment that this is not the best algorithm ever (because in reality, not every tile has the same average intensity). And one possibility would be to have local brightness adjustments, for instance by comparing the overlapping region (but with bleaching, it's not so easy).

philipwinklerdelmic:
@pieleric This is now more than a year old. I think we should either merge or close it.

pieleric:
Indeed... time flies... I'll give a second review.

PR #1393, opened 2020-04-03
install windows: clean-ups

pieleric:


install/windows/build.py
```diff
@@ -45,10 +47,9 @@ def build_odemisviewer_exe():
     run_command(cpy_command)
     run_command(pyi_command, "odemis")
 
-
-def build_delphiviewer_exe():
-    run_command(cpy_command)
-    run_command(pyi_command, "delphi")
+# def build_delphiviewer_exe():
+#     run_command(cpy_command)
+#     run_command(pyi_command, "delphi")
```
Kleijwegt, line 52: Either udelete this function or add a comment why it should remain commented and not deleted

install/windows/build.py
```diff
@@ -61,34 +62,26 @@ def build_odemisviewer_inst():
...
-    run_command(nsis_cmd, "delphi")
+# def build_delphiviewer_inst():
+#     info = [
+#         "/DPRODUCT_NAME=DelphiViewer",
+#         "/DPRODUCT_HNAME=Delphi Viewer",
+#         "/DIMAGE=install_delphi.bmp",
+#         "/DWEBSITE=http://www.delphimicroscope.com",
+#     ]
+#     nsis_cmd = nsis_command[:-1] + info + [nsis_command[-1]]
+#     run_command(nsis_cmd, "delphi")
```
Kleijwegt, line 73: Either undelete this function or add a comment why it should remain commented and not deleted

install/windows/build.py
```diff
@@ -101,34 +94,35 @@ def build_delphiviewer_inst():
...
+#         add_size_to_version()
+#     elif i == 7:
+#         build_odemisviewer_exe()
+#         build_odemisviewer_inst()
+#         add_size_to_version()
+#     elif i == 8:
+#         build_delphiviewer_exe()
+#         build_delphiviewer_inst()
+#         add_size_to_version()
     elif i == 9:
```
Kleijwegt, line 117: Either undelete this function or add a comment why it should remain commented and not deleted

pieleric: Indeed, keeping all the code was a little bit over the top. I've kept the most useful code, so that it's easy to build again the Delphi Viewer if we ever need it, and added a comment about this.

PR #1397, opened 2020-04-07
Modify analysis tab to show 1x1 when all views are the same.

amuskens:
…ame in all views and to only show the first non-empty view in full screen

### DIFF
--- a/src/odemis/gui/cont/tabs.py
+++ b/src/odemis/gui/cont/tabs.py
@@ -2034,8 +2034,55 @@ class AnalysisTab(Tab):
                                 self.tab_data_model.ar_cal.value)
                 self.tab_data_model.ar_cal.value = u""  # remove the calibration
 
-        # TODO: if all the views are either empty or contain the same streams,
+        # if all the views are either empty or contain the same streams,
         # display in full screen by default (with the first view which has streams)
+        # Show the 1 x 1 view if:
+        # - the streams are all the same in every view 
+        # - views are empty with no projection
+        
+        def projection_list_eq(list1, list2):
+            """ Returns true if two projection lists are the same, 
+            i.e. they are projecting the same streams, even if the objects are different
+            list1 (list of Projections)
+            list2 (list of Projections) to be
+            """
+
+            if len(list1) != len(list2):
+                return False
+            
+            # since it is impossible to order the lists, use O(n²) comparison
+            # each proj in list1 should have at least one match in list2
+            for proj1 in list1:
+                # there should be at least one corresponding match in the second list
+                for proj2 in list2:
+                    # matches are of the same type projecting the same stream reference
+                    if type(proj1) == type(proj2) and proj1.stream == proj2.stream:
+                        break
+                else:  # no projection identical to proj1 found
+                    return False
+                
+            # all projections identical
+            return True
+
+        same_projections = []
+        # Are there at least two views different (not including the empty ones)?
+        for view in self.tab_data_model.visible_views.value:
+            view_projections = view.getProjections()
+            # if no projections, the view is empty
+            if not view_projections:
+                pass
+            elif not same_projections:
+                same_projections = view_projections
+            elif  not projection_list_eq(same_projections, view_projections):
+                break  # At least two views are different => leave the display as-is.
+
+        else:  # All views are identical (or empty)
+            self.tab_data_model.viewLayout.value = guimod.VIEW_LAYOUT_ONE
+
+        # Change the focused view to the first non empty view, then display fullscreen
+        # set the focused view to the view with the most streams.
+        self.tab_data_model.focussedView.value = max(self.tab_data_model.visible_views.value,
+                                                         key=lambda v:len(v.getStreams()))
 
         if not extend:
             # Force the canvases to fit to the content


src/odemis/gui/cont/tabs.py
```diff
@@ -2034,8 +2034,68 @@ def display_new_data(self, filename, data, extend=False):
...
-        # TODO: if all the views are either empty or contain the same streams,
+        # if all the views are either empty or contain the same streams,
         # display in full screen by default (with the first view which has streams)
+        # Show the 1 x 1 view if:
+        # - the streams are all the same in every view 
+        # - views are empty with no projection
+        
+        def compare_projection_lists(list1, list2):
+            """ Returns true if two projection lists are the same, 
+            i.e. they are projecting the same streams, even if the objects are different """
```
Kleijwegt, line 2045: Maybe I dumb question,but if you only check if two streams are a projection from the same stream would simply checking the stream name not be sufficient?

src/odemis/gui/cont/tabs.py
```diff
@@ -2034,8 +2034,68 @@ def display_new_data(self, filename, data, extend=False):
...
+            if len(list1) != len(list2):
+                return False
+            
+            matches = 0
```
Kleijwegt, line 2049: Maybe define this as False or None since this will be a bool value and not an int

src/odemis/gui/cont/tabs.py
```diff
@@ -2034,8 +2034,68 @@ def display_new_data(self, filename, data, extend=False):
...
+                    same_projections = view_projections
+                    same_views += 1
+                    first_non_empty_view = view
+                else:
+                    # if there is any difference, then there are different projections and we can abort early
+                    if compare_projection_lists(same_projections, view_projections):
+                        same_views += 1
+                        continue
+                    else:
+                        break
```
Kleijwegt, line 2092: Did you check what happens if you have two streams are duplicated twice?

src/odemis/gui/cont/tabs.py
```diff
@@ -2034,8 +2034,68 @@ def display_new_data(self, filename, data, extend=False):
...
                 self.tab_data_model.ar_cal.value = u""  # remove the calibration
 
-        # TODO: if all the views are either empty or contain the same streams,
+        # if all the views are either empty or contain the same streams,
         # display in full screen by default (with the first view which has streams)
+        # Show the 1 x 1 view if:
+        # - the streams are all the same in every view 
+        # - views are empty with no projection
+        
+        def compare_projection_lists(list1, list2):
```
pieleric, line 2043: Usually (by convention), "compare" functions return -1, 0, or 1 ,depending whether the object 1 is less, equal, or more than object 2.
Functions which just returns a boolean for equal or not equal, are usual called "equal" (or "eq" in Python). So I'd suggest to rename this to `projection_list_eq()`.

src/odemis/gui/cont/tabs.py
amuskens: The projections could be different, so in fact it wouldn't give the desired effect. For instance, one spectrum stream could be projected in several different ways (line spectrum, point spectrum, etc.)

src/odemis/gui/cont/tabs.py
amuskens: I think it will just focus on the first, which is ok and gives the intended effect.

src/odemis/gui/cont/tabs.py
```diff
@@ -2034,9 +2034,57 @@ def display_new_data(self, filename, data, extend=False):
...
+            view_projections = view.getProjections()
+            # if no projections, the view is empty
+            if not view_projections:
+                pass
+            elif not same_projections:
+                same_projections = view_projections
+            elif  not rojection_list_eq(same_projections, view_projections):
+                break  # At least two views are different => leave the display as-is.
+            
+        # use the count to determine if we aborted early.
```
pieleric, line 2078: Drop this comment (and the empty line).

src/odemis/gui/cont/tabs.py
```diff
@@ -2034,9 +2034,57 @@ def display_new_data(self, filename, data, extend=False):
...
+        else:  # All views are identical (or empty)
+            # Change the focused view to the first non empty view, then display fullscreen
+            # set the focused view to the view with the most streams.
+            self.tab_data_model.focussedView.value = max(self.tab_data_model.visible_views.value,
+                                                         key=lambda v:len(v.getStreams()))
```
pieleric, line 2083: *Always* do this code (even if in 2x2 layout).

src/odemis/gui/cont/tabs.py
```diff
@@ -2034,9 +2034,57 @@ def display_new_data(self, filename, data, extend=False):
...
 
-        # TODO: if all the views are either empty or contain the same streams,
+        # if all the views are either empty or contain the same streams,
         # display in full screen by default (with the first view which has streams)
-
+        # Show the 1 x 1 view if:
+        # - the streams are all the same in every view 
+        # - views are empty with no projection
+        
+        def rojection_list_eq(list1, list2):
```
pieleric, line 2043: typo: rojection -> projection.

src/odemis/gui/cont/tabs.py
```diff
@@ -2034,9 +2034,57 @@ def display_new_data(self, filename, data, extend=False):
...
+                    # matches are of the same type projecting the same stream reference
+                    if type(proj1) == type(proj2) and proj1.stream == proj2.stream:
+                        break
+                else:  # no projection identical to proj1 found
+                    return False
+                
+            # all projections identical
+            return True
+
+        same_projections = []
```
pieleric, line 2067: Add a comment here, explaining what you're looking for: Are there at least two views different (not including the empty ones)?

amuskens:
Done!

PR #1398, opened 2020-04-07
acq stream test: Fix test_spectrum_1d complaining the projection is inverted

pieleric:
commit d214e000f4a (acq stream projection: LineSpectrumProjection
clean-up) changed the behaviour of the projection to return (always) the
closest distance as first index.
The users in the GUI were updated, but not the test case, which got
worried.

### DIFF
--- a/src/odemis/acq/test/stream_test.py
+++ b/src/odemis/acq/test/stream_test.py
@@ -4484,7 +4484,8 @@ class StaticStreamsTestCase(unittest.TestCase):
         specs = stream.StaticSpectrumStream("test", spec)
         proj_line_spectrum = LineSpectrumProjection(specs)
 
-        # Check 1d spectrum on corner-case: parallel to the X axis
+        # Check 1d spectrum on corner-case: parallel to the X axis.
+        # We select line #3, from px 7 until px 65 (included)
         specs.selected_line.value = [(3, 7), (3, 65)]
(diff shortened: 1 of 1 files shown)


src/odemis/acq/test/stream_test.py
```diff
@@ -4498,7 +4498,7 @@ def test_spectrum_1d(self):
                          spec.metadata[model.MD_PIXEL_SIZE][0])
 
         # compare to doing it manually, by cutting the band at 3
```
Kleijwegt, line 4500: Cutting the band at 3 what [units?]

src/odemis/acq/test/stream_test.py
```diff
@@ -4498,7 +4498,7 @@ def test_spectrum_1d(self):
...
-        sp1d_raw_ex = spec[:, 0, 0, 65:6:-1, 3]
+        sp1d_raw_ex = spec[:, 0, 0, 7:66, 3]
```
Kleijwegt, line 4501: What do these numbers mean? It seems a bit arbitrary maybe a comment explaining them would be nice?
Or is this just a test set which did work quite good?

pieleric: It's to "manually" do the same thing as done above, using the "fancy" stream interface `.selected_line`.
I've now added some comments to clarify it.

PR #1402, opened 2020-04-09
gui comp foldpanelbar: fix logo transparency on Windows Viewer

pieleric:


pieleric:
Yes, I checked that it's now working better. Strangely the purpose of these functions is to avoid flicker, and I guess in some old version of wxPython or GTK it did help (otherwise we wouldn't have added them). With the latest version (on Ubuntu 18.04) it causes flicker!

PR #1403, opened 2020-04-17
Fixes for the local axes

pieleric:


plugins/la_spec.py
```diff
@@ -83,18 +85,6 @@ def __init__(self, name, detector, dataflow, emitter, l2=None, analyzer=None, **
             self.acquireAllPol = model.BooleanVA(False)
 
     # Copy from ARSettingsStream (to only change the axis when playing)
```
Kleijwegt, line 87: Is this copying still done? I can't see where this comment is referring towards currently

src/odemis/gui/cont/streams.py
```diff
@@ -160,6 +164,9 @@ def __init__(self, stream_bar, stream, tab_data_model, show_panel=True, view=Non
         elif hasattr(stream, "emission"):  # only emission
             self._add_emission_ctrl()
 
+        if len(self.entries) > num_settings_entries:  # TODO: also only if other controls after
```
Kleijwegt, line 167: If TODO's are not done in the near future a more elaborate description would be nice so other people also know the details

plugins/la_spec.py
pieleric: Most of the code below this comment is from the ARSettingsStream (note, this is *not* the SEMARMDStream, which you are familiar with).

src/odemis/gui/cont/streams.py
pieleric: This refers to the TODO which is just above, line 152. I think that comment is sufficiently clear that someone would understand (at least, if seeing the GUI at the same time).

PR #1407, opened 2020-04-23
Registered positions for field corrections (+ comments for debugging)

Kleijwegt:
The registered positions are saved in the global shift for use in the field correction code for calibration of FUMO Beta.
The comments for debugging hint that one can select only a part of the input image (if for example the last column is black but should not be black the measured shift is wrong, by not selecting the last column this can be fixed).

The last change made was removal of unnecessary spaces. PyCharm has a strong tendency to delete those at the end of the line (I have tried multiple times to stop doing that), but maybe this is also a nice cleanup of the code?

### DIFF
--- a/src/odemis/acq/stitching/_registrar.py
+++ b/src/odemis/acq/stitching/_registrar.py
@@ -460,6 +460,9 @@ class GlobalShiftRegistrar(object):
         self.shifts_hor = [[None]]
         self.shifts_ver = [[None]]
 
+        # Calculated position of each tile relative to the upper left (first) tile in pixels as a 3D array
+        self.registered_positions = None
+
         # List of 2D indices for grid positions in order of acquisition
         self.acq_order = []
 
@@ -487,8 +490,9 @@ class GlobalShiftRegistrar(object):
 
     def getPositions(self):
         """
-        Returns the registered positions.
-
+        Updates the registered positions (found using cross correlation and a min spanning tree) and returns the 
+        adjusted tile_positions & dep_tile_positions. When calling this function .registered_positions is updated with
+        the calculated position of each tile relative to the upper left (first) tile in pixels as a 3D array.
         :returns tile_positions: (list of N tuples) the adjusted position in X/Y for each tile, in the
         order they were added
         :returns dep_tile_positions: (list of N tuples of K tuples of 2 floats) for each tile, it returns
@@ -499,9 +503,9 @@ class GlobalShiftRegistrar(object):
         tile_positions = []
         dep_tile_positions = []
 
-        registered_positions = self._assemble_mosaic()
+        self.registered_positions = self._assemble_mosaic()
         for ti in self.acq_order:
-            shift = registered_positions[ti[0]][ti[1]]
+            shift = self.registered_positions[ti[0]][ti[1]]
             tile_positions.append(((shift[0] + firstPosition[0]) * px_size[0],
                                    (firstPosition[1] - shift[1]) * px_size[1]))
 
@@ -618,7 +622,8 @@ class GlobalShiftRegistrar(object):
         prev_tile_roi = numpy.array(prev_tile)[t1:b1, l1:r1]
         tile_roi = numpy.array(tile)[t2:b2, l2:r2]
 
-        # Calculate the shift
+        # If you need to crop the tile without changing the output shift,
+        # you can do it here with the pattern tile_roi[t:-b, l:-r]
         shift = MeasureShift(tile_roi, prev_tile_roi)
         shift_total = numpy.subtract(exp_shift, shift)
 
@@ -681,7 +686,8 @@ class GlobalShiftRegistrar(object):
         Performs a global optimization to find the best path through the tile grid using 
         a minimum spanning tree.
 
-        :returns: (DataArray with shape num_rows x num_cols x 2) registered positions
+        :returns: (numpy array with shape: num_rows x num_cols x 2) registered positions relative to the upper left
+        tile in pixels
         """
         # Transform shifts and errors to adjacency matrices
         # The normalized cross correlation value needs to be transformed, so it can be


tepals:
Small comment with respect to PyCharm. In Settings>Editor>General at the bottom you can specify if you want to strip trailing spaces. You can switch between stripping all trailing spaces in the file your working, or only on lines you modified. 

I do agree that it's a nice cleanup, so do whatever you want with my comment :)

src/odemis/acq/stitching/_registrar.py
```diff
@@ -53,9 +53,9 @@ def __init__(self):
     def addTile(self, tile, dependent_tiles=None):
         """
         Extends grid by one tile.
-        tile (DataArray of shape YX): each image must have at least MD_POS and MD_PIXEL_SIZE metadata. 
+        tile (DataArray of shape YX): each image must have at least MD_POS and MD_PIXEL_SIZE metadata.
         They should all have the same dtype.
-        dependent_tiles (list of DataArray or None): each of the dependent tile, where their position 
+        dependent_tiles (list of DataArray or None): each of the dependent tile, where their position
```
Sabrina85, line 58: What is " each of the dependent tile". I know that you did not write this and only removed the space. However, I think this should be rephrased. Is that a list of the tiles that depend on the "tile" (the param above)? Or the adjacent tiles? What is the "main tile"? The one that is in the tile param? At least it needs to be plural: "each of the dependent tiles"?

src/odemis/acq/stitching/_registrar.py
```diff
@@ -68,17 +68,17 @@ def getPositions(self):
...
         """
         return self.tile_pos, self.dep_tiles_pos
 
 
 class ShiftRegistrar(object):
     """
-    Locates the position of the image relative to the previous top tiles (horizontally and vertically) 
-    by using cross-correlation. The cross-correlation is done using just the part of the images which are 
-    supposed to be overlapping. In case the cross-correlation doesn't work (based on a couple of simple tests), 
+    Locates the position of the image relative to the previous top tiles (horizontally and vertically)
```
Sabrina85, line 79: Is an image not made up of the tiles? I am not familiar with the code, so do you want to locate the position of a tile (and not an image) relative to the previous tiles?

src/odemis/acq/stitching/_registrar.py
```diff
@@ -114,14 +114,14 @@ def __init__(self):
     def addTile(self, tile, dependent_tiles=None):
         """
         Extends grid by one tile.
-        tile (DataArray of shape YX): each image must have at least MD_POS and MD_PIXEL_SIZE metadata. 
+        tile (DataArray of shape YX): each image must have at least MD_POS and MD_PIXEL_SIZE metadata.
         They should all have the same dtype.
-        dependent_tiles (list of K DataArray or None): each of the dependent tile, where their position 
+        dependent_tiles (list of K DataArray or None): each of the dependent tile, where their position
```
Sabrina85, line 119: See my comment above.

src/odemis/acq/stitching/_registrar.py
```diff
@@ -282,7 +282,7 @@ def _estimateROI(self, shift):
     def _estimateMatch(self, imageA, imageB, shift):
         """
         Returns an estimation of the similarity between the given images
-        when the second is shifted by the shift value. It is used to assess 
+        when the second is shifted by the shift value. It is used to assess
         the quality of a shift measurement by giving the shifted image.
```
Sabrina85, line 286: I don't understand how the quality of the shift measurement is assessed by "giving the shifted image"?

src/odemis/acq/stitching/_registrar.py
```diff
@@ -618,8 +620,8 @@ def _get_shift(self, prev_tile, tile):
         prev_tile_roi = numpy.array(prev_tile)[t1:b1, l1:r1]
         tile_roi = numpy.array(tile)[t2:b2, l2:r2]
 
-        # Calculate the shift
-        shift = MeasureShift(tile_roi, prev_tile_roi)
+        # Calculate the shift ([:,:] is a hint how to specify column/row for debugging purposes)
+        shift = MeasureShift(tile_roi[:, :], prev_tile_roi[:, :])
```
Sabrina85, line 624: I am not sure if we should do that. Maybe it would be better to specify the shape of `tile_roi` and `prev_tile_roi` at the place where you create the array? @pieleric What do you think?

src/odemis/acq/stitching/_registrar.py
```diff
@@ -460,6 +460,8 @@ def __init__(self):
         self.shifts_hor = [[None]]
         self.shifts_ver = [[None]]
 
+        self.registered_positions = [[(None, None)]]  #list of calculated positions (tuple of 2 floats/None)
+
```
Sabrina85, line 464: Is there a need to create the list here for the field corrections code? Otherwise I think it is enough to create it in line 504 where you overwrite it.

Kleijwegt: I thought it is nice to define this list at the init because the convention is more or less to declare it there and explain what the variable holds.

src/odemis/acq/stitching/_registrar.py
Kleijwegt: This is just there fore debugging purposes. I had a black line in my samples and therefore needed to exclude the last column of every part on which the shift was measured. I like to leave it here so that for a future user looking and going through the code it is easy to find where one should implement this easily (for debuging).

Kleijwegt:
> I added also some comments for the doc strings. I know that this is old text, but some sentences just make no sense to me. So it might be worse to look into those if you anyways do formatting. However, in general I would at least make it two commits. One for the actual code changes and one for the formatting/clean-up.

True, maybe we should/could do an update of the doc strings in a different PR then, @philipwinklerdelmic  what do you think?

src/odemis/acq/stitching/_registrar.py
Sabrina85: Then add the shape (x, y).

src/odemis/acq/stitching/_registrar.py
Sabrina85: I would say how to access a column and row in an array does not need to be explicitly specified in the code. This I would place under common knowledge. What you are describing that there are black lines that need to be handled, is something you don't mention in you comment - so it does not provide additional information as it is now.

src/odemis/acq/stitching/_registrar.py
pieleric: Here "image" and "tile" are used interchangeably. Admittedly, it'd be less confusing to use always "tile". 
@Kleijwegt You see why you shouldn't fix the formatting at the same time as the code: people start talking about things completely unrelated ;-)
From my point of view, I'm fine whether you want to also start fixing/improving docstrings... but then do it in a separate commit!

src/odemis/acq/stitching/_registrar.py
pieleric: I also think it's a good idea to initialise all the attributes in the init. However, here I'd suggest to just initialise it to `None`, as it just would be a bad idea to read it before calling `getPositions()`.
Also, please fix the type. It's *not* a list. See the docstring of `_assemble_mosaic()`.

src/odemis/acq/stitching/_registrar.py
```diff
@@ -499,9 +501,9 @@ def getPositions(self):
         tile_positions = []
         dep_tile_positions = []
 
-        registered_positions = self._assemble_mosaic()
+        self.registered_positions = self._assemble_mosaic()
```
pieleric, line 504: Please update the docsting for `.getPositions()` to indicate that .registered_positions will be updated after the call.

src/odemis/acq/stitching/_registrar.py
pieleric: I also don't think it's worthy to modify the code itself. It's probably not going to slow it down much, but that's more confusing than helpful.

Honestly, even if you've spent quite some time looking for this trick, the use case is so unlikely to be used again that I'm not convinced it's really worthy to to mention here. Actually, if the tiles need to be crop, you should just crop them before sending passing them to the registrar. If you still think it could be useful, I'd suggest to just add a separate comment reading like:
`# If you need to crop the tiles, you can do it here with the pattern tile_roi[t:-b, l:-r]`

src/odemis/acq/stitching/_registrar.py
Kleijwegt: Changed it to numpy.array (Docstring of _assemble_mosaic is actually wrong currently)

Kleijwegt:
Removed all trailing spaces, and changed comments of @pieleric 
@Sabrina85 I looks better to update the docstring update in a separate PR later.

src/odemis/acq/stitching/_registrar.py
```diff
@@ -460,6 +460,8 @@ def __init__(self):
         self.shifts_hor = [[None]]
         self.shifts_ver = [[None]]
 
+        self.registered_positions = numpy.array(None)  #list of calculated positions (tuple of 2 floats/None)
```
pieleric, line 463: I think `None` would be fine, no need to have an array, if anyway all you want to indicate is that the data is not available.
However, the main thing to update is the comment: It's *not* the calculated positions, and it's *not* a tuple! It's the calculated shifts, as a 3D array...

src/odemis/acq/stitching/_registrar.py
```diff
@@ -487,7 +489,8 @@ def addTile(self, tile, dependent_tiles=None):
 
     def getPositions(self):
         """
-        Returns the registered positions.
+        Updates the registered positions (found using cross correlation and a min spanning tree) and returns the 
+        adjusted tile_positions & dep_tile_positions.
```
pieleric, line 493: Please add a line about `.registered_positions` being updated when this function is called.

src/odemis/acq/stitching/_registrar.py
```diff
@@ -499,9 +502,9 @@ def getPositions(self):
         tile_positions = []
         dep_tile_positions = []
 
-        registered_positions = self._assemble_mosaic()
+        self.registered_positions = self._assemble_mosaic()
```
philipwinklerdelmic, line 505: `registered_positions` doesn't really contain the positions, but the shifts (that's my mistake). Could you rename all occurrences (also in the other registrars) to something like `self.shifts`?

PR #1409, opened 2020-04-24
Bug fixed: Simple click with label tool creates label

VictoriaDel:


src/odemis/gui/comp/overlay/world.py
```diff
@@ -1465,6 +1464,10 @@ def stop_updating_tool(self):
             if self._ask_user_for_text or not self._edit_label_end:
                 self._edit_text()
                 self._ask_user_for_text = False
+            # In case no line is drawn, the user can still type some text
+            if self.v_end_pos is None:
+                offset = self.cnvs.get_half_buffer_size()
+                self.p_start_pos = self.p_end_pos = self.cnvs.view_to_phys(self.v_start_pos, offset)
```
pieleric, line 1470: I don't understand why this is needed. The `GenericGadgetLine.start_dragging()` method (now) takes care of initialising the v_start/v_end and they are immediately used to compute p_start/p_end. So this `if` should never be triggered. Or am I missing a corner case?

src/odemis/gui/comp/overlay/world.py
```diff
@@ -1520,7 +1523,7 @@ def sync_with_canvas(self, shift=(0, 0), scale=1.0):
         """ Given that the canvas has been shifted or rescaled, update the view positions of the label """
         shiftscale = (shift, scale)
         update_view = self.last_shiftscale != shiftscale
-        if update_view:
+        if update_view and self.p_start_pos:
```
pieleric, line 1526: Here as well, I don't understand `p_start_pos` could be None?

VictoriaDel:
I forgot to use the latest version in this specific branch and v_pos_end was None so that caused a lot of problems. It also seemed weird to me but I didn't want to make big changes in the code. Once I fetched, it works fine with the only last change. I am sorry and thank you for the remark !

PR #1411, opened 2020-04-28
Show pixel raw value

VictoriaDel:


src/odemis/acq/stream/_live.py
```diff
@@ -77,6 +77,32 @@ def _find_metadata(self, md):
...
+        if self.raw:
+            md = self.raw[0].metadata
+            pxs = md.get(model.MD_PIXEL_SIZE, (1e-6, 1e-6))
+            # (l, t, r, b): bounding box
+            bounding_box = self.getBoundingBox()
+            # Calculate the distance to the left bottom in physical units
+            if bounding_box[0] <= p_pos[0] <= bounding_box[2] and bounding_box[1] <= p_pos[1] <= bounding_box[3]:
+                dist = (p_pos[0] - bounding_box[0],
+                        - (p_pos[1] - bounding_box[3]))
+                return int(dist[0] / pxs[0]), int(dist[1] / pxs[1])
```
pieleric, line 98: Sorry, I realise just now that this is not going to be sufficient for data which is rotated. On the SECOM, the images can have MD_SHEAR and MD_ROTATION. In getBoundingBox, we incorrectly don't take them into account. As they are usually small, it's probably not a big deal. However, to know precisely which pixel corresponds a specific position, you *must* take them into account. You can use the (new) 20180703-MPI-pond-01.ome.tiff in our test data set for checking. I'd also suggest to play with the "manual overlay" plugin to increase the rotation and shear, so that it's even more obvious.

I *think* that the position of the pixels in physical coordinates is computed using a transformation matrix applied with the center of the image as referential, with rotation * shear*scale + translation. The exact way is coded in the gui.comp.canvas.draw() function. It might be best to first confirm the exact formula by trying to plot the top left corner (or any corner) of the image using an overlay and only the metadata. Once this matches the top-left corner, you "just" need to inverse the formula to compute the physical coordinates in the pixel coordinates.

src/odemis/gui/comp/overlay/base.py
```diff
@@ -49,14 +49,15 @@
 class Label(object):
     """ Small helper class that stores label properties """
 
-    def __init__(self, text, pos, font_size, flip, align, colour, opacity, deg):
+    def __init__(self, text, pos, font_size, flip, align, colour, opacity, background, deg):
```
pieleric, line 52: Please add a docstring to indicate the type of background (None or some colour??).
I'd also recommend putting it last, with a default value `=None` so that you don't need to update every code using the Label.

src/odemis/acq/stream/_projection.py
```diff
@@ -898,6 +899,40 @@ def _projectXY2RGB(self, data, tint=(255, 255, 255)):
...
+    def getPixelCoordinates(self, p_pos):
+        """
+        Translate physical coordinates into data pixel coordinates
+        Args:
+            p_pos: the position in physical coordinates
+
+        Returns: the position in pixel coordinates
+
+        """
+        md = self.stream.raw[0].metadata
```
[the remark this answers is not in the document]
VictoriaDel: I am not sure if I implemented it the way you proposed. Please check my last changes and let me know.

src/odemis/acq/stream/_base.py
```diff
@@ -1049,30 +1052,86 @@ def _onNewData(self, dataflow, data):
...
+        pxs = md.get(model.MD_PIXEL_SIZE, (1e-6, 1e-6))
+        rotation = md.get(model.MD_ROTATION, 0)
+        shear = md.get(model.MD_SHEAR, 0)
+        translation = md.get(model.MD_POS, (0, 0))
+        if rotation or shear:
+            size = raw.shape[-1], raw.shape[-2]
+            tform = AffineTransform(rotation=rotation, scale=pxs, shear=shear, translation=translation).inverse()
+            pixel_pos = tform(p_pos)
+            if 0 <= p_pos[0] <= size[0] and 0 <= p_pos[1] <= size[1]:
+                return int(pixel_pos[0] + size[0] / 2), - int(pixel_pos[1] - size[1] / 2)
```
pieleric, line 1079: Please explicitly have a `else: return None`. That's the default in Python, but it's better to be explicit about what we return.

src/odemis/acq/stream/_projection.py
```diff
@@ -898,11 +900,69 @@ def _projectXY2RGB(self, data, tint=(255, 255, 255)):
...
+        """
+        Translate physical coordinates into data pixel coordinates
+        Args:
+            p_pos(tuple float, float): the position in physical coordinates
+            im: (None or DataArray(Shadow)): None or the data of the image
+
+        Returns(None or tuple int, int): the position in pixel coordinates
+
+        """
+        md = self.stream.raw[0].metadata.copy()
```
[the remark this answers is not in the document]
pieleric: Please replace this entire function by just `return self.stream.getPixelCoodinates(p_pos)`.
Or is there a reason this wouldn't work?

PR #1415, opened 2020-05-02
util conversion: new JsonExtraEncoder to export numpy types as JSON

pieleric:
When exporting some Metadata ExtraSettings, if it contains a numpy
value, the whole data fails to be exported.
Example:
Failed to save ExtraSettings metadata, exception: Object of type 'int64' is not JSON serializable

=> Properly convert the numpy scalar/array into the right JSON
representation.

### DIFF
--- a/src/odemis/util/test/conversion_test.py
+++ b/src/odemis/util/test/conversion_test.py
@@ -22,6 +22,7 @@ Odemis. If not, see http://www.gnu.org/licenses/.
 '''
 from __future__ import division
 
+import json
 import math
 import numpy
 from odemis import model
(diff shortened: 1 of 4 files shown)


src/odemis/util/test/conversion_test.py
```diff
@@ -314,6 +315,23 @@ def test_get_tile_md_pos(self):
...
+
+        s = json.dumps(obj, cls=JsonExtraEncoder)
+
+        # decoding it back is not going to bring back the numpy types, but
+        # something equivalent
+        dec_obj = json.loads(s)
+        self.assertEqual(dec_obj["int"], 65)
+        self.assertEqual(dec_obj["float"], -64.1)
+        self.assertEqual(dec_obj["2d"], [[1, 2], [3, 4]])
+
```
Kleijwegt, line 334: It looks as if tests for complex values, sets, and bytes are not implemented in this PR. Maybe this should also be build since the class "JsonExtraEncoder" does add json.dumps functionality for these types.

pieleric:
Added extra test cases for all the other types.

PR #1416, opened 2020-05-04
plugins: new export_spec_3d_csv

pieleric:
Allows to export to a CSV a (3D) spectrum data.

As requested by some users.

### DIFF
--- a/plugins/export_spec_3d_csv.py
+++ b/plugins/export_spec_3d_csv.py
@@ -1,0 +1,120 @@
+# -*- coding: utf-8 -*-
+"""
+Created on 24 Apr 2020
+
+@author: Éric Piel
+
+Copyright © 2020 Éric Piel, Delmic
+
+This file is part of Odemis.
+
+Odemis is free software: you can redistribute it and/or modify it under the terms of the GNU
+General Public License version 2 as published by the Free Software Foundation.
+
+Odemis is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even
+the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General
+Public License for more details.
+
+You should have received a copy of the GNU General Public License along with Odemis. If not,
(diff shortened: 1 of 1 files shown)


plugins/export_spec_3d_csv.py
```diff
@@ -0,0 +1,125 @@
...
+            # Y
+            for iy in range(da.shape[-2]):
+                # X
+                for ix in range(da.shape[-1]):
+                    # Wavelength
+                    for wl, data in zip(spectrum_range, da[:, iy, ix]):
+                        csv_writer.writerow([wl, ix * pxs[0], iy * pxs[1], data])
+                        l += 1
+
+        logging.debug("Exported %d rows", l)
```
tepals, line 124: Isn't `l` equal to `da.shape[-2] * da.shape[-1] * len(spectrum_range)`? It looks strange that you have to do the `l+=1` to get the number of rows exported.

plugins/export_spec_3d_csv.py
```diff
@@ -0,0 +1,125 @@
+# -*- coding: utf-8 -*-
+'''
```
tepals, line 2: According to PEP these should actually also be triple double quotes. https://www.python.org/dev/peps/pep-0257/#what-is-a-docstring

It looks like most plugins in odemis have triple single quotes, so you can also leave it like this.

plugins/export_spec_3d_csv.py
pieleric: Indeed, it *should* be prod(da.shape). During development I wanted to check that it's indeed the case. 
This code is not needed anymore, so I'll remove it.

plugins/export_spec_3d_csv.py
pieleric: I agree totally. Actually I think that strings should just always use the double quotes. For consistency I'd like to have no string with single quote. This actually comes from a copy-paste of another plugin. I'll fix this one at least....

PR #1417, opened 2020-05-07
acq stream sync: fixes related to live update

pieleric:


src/odemis/acq/stream/_sync.py
```diff
@@ -2149,35 +2157,35 @@ def _updateImage(self):
...
                 #   there is no SEM image anyway.
                 continue
 
+            # Display only the image taken in the last polarization and convert to RGB + display
             try:
-                raw_data = self._live_data[stream_number][-1]
+                raw_data = self._live_data[stream_idx][-1]
             except IndexError:  # Can happen if the acquisition has just finished
-                if self._acq_done.is_set():
+                if  self._current_scan_area is None:
```
Kleijwegt, line 2172: double space

src/odemis/acq/stream/_sync.py
```diff
@@ -167,6 +167,7 @@ def __init__(self, name, streams):
         self._acq_done = threading.Event()
 
         # Start threading event for live update overlay
+        self._current_scan_area = (0, 0, 0, 0)  # currently scanned area location based on px_idx
```
Kleijwegt, line 170: Would be nice to add here what happens if current_scan_area is defined as None (no live update will be given because of the if statement in _updateImage).

Kleijwegt, line 170: Furthermore why not define _current_scan_area not as (None,None,None,None)? Since at init no area is defined? (just out of curriosity I don't think this is essential to implement)

pieleric: I've changed it to `None` by default. I don't think it's really useful to write `(None,None,None,None)`. Either it contains something invalid (`None`) or it contains a tuple with ints. No need to be half-way.

pieleric: Updated the comment.

PR #1421, opened 2020-05-11
Sonic beam offset controller

philipwinklerdelmic:
First attempt at offset controller driver.

Do you think it makes sense to write a simulator and testcases for this driver? If so, what would you test for? We don't get any values back from the hardware if I understand it correctly, so there's not much we can do.

### DIFF
--- a/src/odemis/driver/test/tfsbc_test.py
+++ b/src/odemis/driver/test/tfsbc_test.py
@@ -1,0 +1,127 @@
+# -*- coding: utf-8 -*-
+"""
+Created on 11 May 2020
+
+@author: Philip Winkler
+
+Copyright © 2020 Philip Winkler, Delmic
+
+This file is part of Odemis.
+
+Odemis is free software: you can redistribute it and/or modify it under the terms
+of the GNU General Public License version 2 as published by the Free Software
+Foundation.
+
+Odemis is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+PURPOSE. See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with
+Odemis. If not, see http://www.gnu.org/licenses/.
+"""
+from __future__ import division
+
+from odemis.driver import tfsbc
+from odemis import model
+import unittest
+import logging
+import os
+import time
+import numpy
+
+TEST_NOHW = (os.environ.get("TEST_NOHW", 0) != 0)  # Default to Hw testing
+
+logging.getLogger().setLevel(logging.DEBUG)
+logging.basicConfig(format="%(asctime)s  %(levelname)-7s %(module)s:%(lineno)d %(message)s")
+
+if TEST_NOHW:
+    PORT = "/dev/fake"
+else:
+    PORT = "/dev/ttyUSB*"
+
+
+# @skip("skip")
+class TestBeamShiftController(unittest.TestCase):
+    """
+    Tests the beam controller driver.
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        cls.bc = tfsbc.BeamShiftController("DC Offset", None, PORT, "RS485")
+        # Values found during testing on the hardware
+        md = ((-0.00027788219369730165, 0.0013604844623992785),
+              (-0.0013604844623992785, -0.00027788219369730165),
+              (0.00012699407486043473, -0.0006217507619527259),
+              (0.0006217507619527259, 0.00012699407486043473))
+        cls.bc.updateMetadata({model.MD_CALIB_BEAMSHIFT: md})
+
+    @classmethod
+    def tearDownClass(cls):
+        pass
+
+    def assertTupleAlmostEqual(self, first, second, places=None, msg=None, delta=None):
+        """
+        check two tuples are almost equal (value by value)
+        """
+        for f, s in zip(first, second):
+            self.assertAlmostEqual(f, s, places=places, msg=msg, delta=delta)
+
+    def test_read_write(self):
+        vals = [27000, 37000, 20000, 44000]
+        self.bc._write_registers(vals)
+        ret = self.bc._read_registers()
+        self.assertTupleAlmostEqual(vals, list(ret), places=1)
+
+    def test_shifts(self):
+        """
+        Move to different shifts, test if .shift VA is updated correctly. Wait in between,
+        so the effect can be seen on the hardware.
+        """
+        shift = (0, 0)
+        self.bc.shift.value = shift
+        self.assertTupleAlmostEqual(self.bc.shift.value, shift, places=9)
+        time.sleep(1)
+
+        shift = (-5e-6, 0)
+        self.bc.shift.value = shift
+        self.assertTupleAlmostEqual(numpy.subtract(self.bc.shift.value, shift), (0, 0), places=7)
+        time.sleep(1)
+
+        shift = (-5e-6, -5e-6)
+        self.bc.shift.value = shift
+        self.assertTupleAlmostEqual(numpy.subtract(self.bc.shift.value, shift), (0, 0), places=7)
+        time.sleep(1)
+
+        shift = (0, -5e-6)
+        self.bc.shift.value = shift
+        self.assertTupleAlmostEqual(numpy.subtract(self.bc.shift.value, shift), (0, 0), places=7)
+
+        shift = (0, 0)
+        self.bc.shift.value = shift
+        self.assertTupleAlmostEqual(numpy.subtract(self.bc.shift.value, shift), (0, 0), places=7)
+
+    def test_write_time(self):
+        startt = time.time()
+        shift = (-3e-6, 1e-6)
+        self.bc.shift.value = shift
+        self.assertLess(time.time() - startt, 0.03, "Reading/writing took more than 30 ms.")
+        logging.debug("Shift value set to %s", self.bc.shift.value)
+        self.assertTupleAlmostEqual(numpy.subtract(self.bc.shift.value, shift), (0, 0), places=7)
+
+    def test_transform_coordinates(self):
+        val = (0, 4e-6)
+        md = ((-0.00027788219369730165, 0.0013604844623992785),
+              (-0.0013604844623992785, -0.00027788219369730165),
+              (0.00012699407486043473, -0.0006217507619527259),
+              (0.0006217507619527259, 0.00012699407486043473))
+        expected = [0x6f7e, 0x835f, 0x7874, 0x7e75]  # from testing with example script
+        ret = tfsbc.transform_coordinates(val, *md)
+        self.assertTupleAlmostEqual(ret, expected, places=5)
+        rev = tfsbc.transform_coordinates_reverse(ret, *md)
+        self.assertTupleAlmostEqual(val, rev, places=5)
+
+
+if __name__ == "__main__":
+    unittest.main()
+
--- a/src/odemis/driver/test/tfsbc_test.py
+++ b/src/odemis/driver/test/tfsbc_test.py
@@ -48,7 +48,7 @@ class TestBeamShiftController(unittest.TestCase):
 
     @classmethod
     def setUpClass(cls):
-        cls.bc = tfsbc.BeamShiftController("DC Offset", None, PORT, "RS485")
+        cls.bc = tfsbc.BeamShiftController("DC Offset", None, PORT, "FT2OMDD5")
         # Values found during testing on the hardware
         md = ((-0.00027788219369730165, 0.0013604844623992785),
               (-0.0013604844623992785, -0.00027788219369730165),
--- a/src/odemis/driver/test/tfsbc_test.py
+++ b/src/odemis/driver/test/tfsbc_test.py
@@ -80,26 +80,35 @@ class TestBeamShiftController(unittest.TestCase):
         """
         shift = (0, 0)
         self.bc.shift.value = shift
-        self.assertTupleAlmostEqual(self.bc.shift.value, shift, places=9)
+        self.assertTupleAlmostEqual(self.bc.shift.value, shift, places=7)
         time.sleep(1)
 
         shift = (-5e-6, 0)
         self.bc.shift.value = shift
-        self.assertTupleAlmostEqual(numpy.subtract(self.bc.shift.value, shift), (0, 0), places=7)
+        self.assertTupleAlmostEqual(self.bc.shift.value, shift, places=7)
         time.sleep(1)
 
         shift = (-5e-6, -5e-6)
         self.bc.shift.value = shift
-        self.assertTupleAlmostEqual(numpy.subtract(self.bc.shift.value, shift), (0, 0), places=7)
+        self.assertTupleAlmostEqual(self.bc.shift.value, shift, places=7)
         time.sleep(1)
 
         shift = (0, -5e-6)
         self.bc.shift.value = shift
-        self.assertTupleAlmostEqual(numpy.subtract(self.bc.shift.value, shift), (0, 0), places=7)
+        self.assertTupleAlmostEqual(self.bc.shift.value, shift, places=7)
 
         shift = (0, 0)
         self.bc.shift.value = shift
-        self.assertTupleAlmostEqual(numpy.subtract(self.bc.shift.value, shift), (0, 0), places=7)
+        self.assertTupleAlmostEqual(self.bc.shift.value, shift, places=7)
+
+        # Large shift
+        shift = (500e-6, 0)
+        with self.assertRaises(ValueError):
+            self.bc.shift.value = shift
+
+        shift = (0, 500e-6)
+        with self.assertRaises(ValueError):
+            self.bc.shift.value = shift
 
     def test_write_time(self):
         startt = time.time()
@@ -107,7 +116,7 @@ class TestBeamShiftController(unittest.TestCase):
         self.bc.shift.value = shift
         self.assertLess(time.time() - startt, 0.03, "Reading/writing took more than 30 ms.")
         logging.debug("Shift value set to %s", self.bc.shift.value)
-        self.assertTupleAlmostEqual(numpy.subtract(self.bc.shift.value, shift), (0, 0), places=7)
+        self.assertTupleAlmostEqual(self.bc.shift.value, shift, places=7)
 
     def test_transform_coordinates(self):
         val = (0, 4e-6)
@@ -117,9 +126,23 @@ class TestBeamShiftController(unittest.TestCase):
               (0.0006217507619527259, 0.00012699407486043473))
         expected = [0x6f7e, 0x835f, 0x7874, 0x7e75]  # from testing with example script
         ret = tfsbc.transform_coordinates(val, *md)
-        self.assertTupleAlmostEqual(ret, expected, places=5)
+        self.assertEqual(ret, expected)
         rev = tfsbc.transform_coordinates_reverse(ret, *md)
         self.assertTupleAlmostEqual(val, rev, places=5)
+
+    def test_update_incorrect_md(self):
+        bs = (-0.00027788219369730165, 0.0013604844623992785)
+        self.assertRaises(ValueError, self.bc.updateMetadata, {model.MD_CALIB_BEAMSHIFT: bs})
+        bs = None
+        self.assertRaises(ValueError, self.bc.updateMetadata, {model.MD_CALIB_BEAMSHIFT: bs})
+        bs = ((-0.00027788219369730165, 0.0013604844623992785))
+        self.assertRaises(ValueError, self.bc.updateMetadata, {model.MD_CALIB_BEAMSHIFT: bs})
+        bs = ((-0.00027788219369730165), (-0.00027788219369730165, 0.0013604844623992785),
+              (-0.00027788219369730165, 0.0013604844623992785), (-0.00027788219369730165, 0.0013604844623992785))
+        self.assertRaises(ValueError, self.bc.updateMetadata, {model.MD_CALIB_BEAMSHIFT: bs})
+        bs = (("-0.00027788219369730165", "0.0013604844623992785"), (-0.00027788219369730165, 0.0013604844623992785),
+              (-0.00027788219369730165, 0.0013604844623992785), (-0.00027788219369730165, 0.0013604844623992785))
+        self.assertRaises(ValueError, self.bc.updateMetadata, {model.MD_CALIB_BEAMSHIFT: bs})
 
 
 if __name__ == "__main__":
--- a/src/odemis/driver/test/tfsbc_test.py
+++ b/src/odemis/driver/test/tfsbc_test.py
@@ -54,7 +54,7 @@ class TestBeamShiftController(unittest.TestCase):
               (-0.0013604844623992785, -0.00027788219369730165),
               (0.00012699407486043473, -0.0006217507619527259),
               (0.0006217507619527259, 0.00012699407486043473))
-        cls.bc.updateMetadata({model.MD_CALIB_BEAMSHIFT: md})
+        cls.bc.updateMetadata({model.MD_CALIB: md})
 
     @classmethod
     def tearDownClass(cls):
@@ -132,17 +132,17 @@ class TestBeamShiftController(unittest.TestCase):
 
     def test_update_incorrect_md(self):
         bs = (-0.00027788219369730165, 0.0013604844623992785)
-        self.assertRaises(ValueError, self.bc.updateMetadata, {model.MD_CALIB_BEAMSHIFT: bs})
+        self.assertRaises(ValueError, self.bc.updateMetadata, {model.MD_CALIB: bs})
         bs = None
-        self.assertRaises(ValueError, self.bc.updateMetadata, {model.MD_CALIB_BEAMSHIFT: bs})
+        self.assertRaises(ValueError, self.bc.updateMetadata, {model.MD_CALIB: bs})
         bs = ((-0.00027788219369730165, 0.0013604844623992785))
-        self.assertRaises(ValueError, self.bc.updateMetadata, {model.MD_CALIB_BEAMSHIFT: bs})
+        self.assertRaises(ValueError, self.bc.updateMetadata, {model.MD_CALIB: bs})
         bs = ((-0.00027788219369730165), (-0.00027788219369730165, 0.0013604844623992785),
               (-0.00027788219369730165, 0.0013604844623992785), (-0.00027788219369730165, 0.0013604844623992785))
-        self.assertRaises(ValueError, self.bc.updateMetadata, {model.MD_CALIB_BEAMSHIFT: bs})
+        self.assertRaises(ValueError, self.bc.updateMetadata, {model.MD_CALIB: bs})
         bs = (("-0.00027788219369730165", "0.0013604844623992785"), (-0.00027788219369730165, 0.0013604844623992785),
               (-0.00027788219369730165, 0.0013604844623992785), (-0.00027788219369730165, 0.0013604844623992785))
-        self.assertRaises(ValueError, self.bc.updateMetadata, {model.MD_CALIB_BEAMSHIFT: bs})
+        self.assertRaises(ValueError, self.bc.updateMetadata, {model.MD_CALIB: bs})
 
 
 if __name__ == "__main__":
--- a/src/odemis/driver/tfsbc.py
+++ b/src/odemis/driver/tfsbc.py
@@ -1,0 +1,357 @@
+# -*- coding: utf-8 -*-
+"""
+Created on 11 May 2020
+
+@author: Philip Winkler
+
+Copyright © 2020 Philip Winkler, Delmic
+
+This file is part of Odemis.
+
+Odemis is free software: you can redistribute it and/or modify it under the terms
+of the GNU General Public License version 2 as published by the Free Software
+Foundation.
+
+Odemis is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+PURPOSE. See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with
+Odemis. If not, see http://www.gnu.org/licenses/.
+"""
+from __future__ import division
+
+import logging
+
+import numpy
+import serial
+import serial.tools.list_ports
+from pymodbus.client.sync import ModbusSerialClient
+
+from odemis import model
+from odemis.model import HwError
+
+# Parameters for connection
+BAUDRATE = 230400
+TIMEOUT = 1
+BYTESIZE = 8
+PARITY = serial.PARITY_NONE
+STOPBITS = 1
+
+# Modbus level addresses
+SLAVE_UNIT = 2
+
+# Modbus registers
+BEAMDEFL_LX = 0  # lower x beam deflection element control
+BEAMDEFL_LY = 1  # lower y beam deflection element control
+BEAMDEFL_UX = 2  # upper x beam deflection element control
+BEAMDEFL_UY = 3  # upper y beam deflection element control
+
+# Transformation constants
+DCREVERSCOEF = -1
+DCROTUPPXX = 1
+DCROTUPPYY = 1
+DCROTLOWXX = -1
+DCROTLOWYY = 1
+
+# Conversion ranges
+C_MIN_RAW_SHIFT = 0
+C_MAX_RAW_SHIFT = 0xFFFF
+C_MIN_DBL_SHIFT = -42.2e-3
+C_MAX_DBL_SHIFT = 42.2e-3
+
+
+def current_to_raw(current):
+    """
+    Helper function for coordinate transform (from Thermofischer example code).
+    :param current: (float)
+    """
+    k = (C_MAX_RAW_SHIFT - C_MIN_RAW_SHIFT) / (C_MAX_DBL_SHIFT - C_MIN_DBL_SHIFT)
+    return int((current - C_MIN_DBL_SHIFT) * k + C_MIN_RAW_SHIFT + 0.5)
+
+
+def raw_to_current(raw):
+    """
+    Inverse of current_to_raw.
+    :param raw: (int)
+    """
+    k = (C_MAX_RAW_SHIFT - C_MIN_RAW_SHIFT) / (C_MAX_DBL_SHIFT - C_MIN_DBL_SHIFT)
+    return (raw - 0.5 - C_MIN_RAW_SHIFT) / k + C_MIN_DBL_SHIFT
+
+
+def transform_coordinates(value, xlower, ylower, xupper, yupper):
+    """
+    Transform x, y coordinates to register values of beamshift hardware.
+    :param value: (float, float) x, y value in the source coordinate system
+    :param xlower: (float, float) xlower metadata
+    :param ylower: (float, float) ylower metadata
+    :param xupper: (float, float) xupper metadata
+    :param yupper: (float, float) yupper metadata
+    :return (int, int, int, int): register values: x lower, y lower, x upper, y upper
+    """
+    value = (value[0] * 1e6, value[1] * 1e6)  # value in µm
+
+    # This transformation was provided as example code from Thermofischer (the variable names are slightly modified
+    # to fit the coding style of this driver, the rest of the calculation is identical).
+    dc_xupper = value[0] * xupper[0] + value[1] * yupper[0]
+    dc_xlower = value[0] * xlower[0] + value[1] * ylower[0]
+    dc_yupper = value[0] * xupper[1] + value[1] * yupper[1]
+    dc_ylower = value[0] * xlower[1] + value[1] * ylower[1]
+
+    currUX = DCREVERSCOEF * DCROTUPPXX * dc_xupper
+    currLX = DCREVERSCOEF * DCROTLOWXX * dc_xlower
+    currUY = DCREVERSCOEF * DCROTUPPYY * dc_yupper
+    currLY = DCREVERSCOEF * DCROTLOWYY * dc_ylower
+
+    for current in [currLX, currLY, currUX, currUY]:
+        if current < C_MIN_DBL_SHIFT or current > C_MAX_DBL_SHIFT:
+            raise ValueError("Beam deflection %s exceeds limits (%s, %s) of DC coils."
+                             % (current, C_MIN_DBL_SHIFT, C_MAX_DBL_SHIFT))
+
+    rawLX = current_to_raw(currLX)
+    rawUX = current_to_raw(currUX)
+    rawLY = current_to_raw(currLY)
+    rawUY = current_to_raw(currUY)
+
+    return [rawLX, rawLY, rawUX, rawUY]
+
+
+def transform_coordinates_reverse(register_values, xlower, ylower, xupper, yupper):
+    """
+    Transform register values back to x, y position in source coordindate system.
+    :param register_values: (int, int, int, int) register values
+    :param xlower: (float, float) xlower metadata
+    :param xupper: (float, float) xupper metadata
+    :param ylower: (float, float) ylower metadata
+    :param yupper: (float, float) yupper metadata
+    :return (int, int): x, y position
+    """
+    rawLX, rawLY, rawUX, rawUY = register_values
+
+    currLX = raw_to_current(rawLX)
+    currLY = raw_to_current(rawLY)
+    currUX = raw_to_current(rawUX)
+    currUY = raw_to_current(rawUY)
+
+    dc_xupper = currUX / (DCREVERSCOEF * DCROTUPPXX)
+    dc_xlower = currLX / (DCREVERSCOEF * DCROTLOWXX)
+    dc_yupper = currUY / (DCREVERSCOEF * DCROTUPPYY)
+    dc_ylower = currLY / (DCREVERSCOEF * DCROTLOWYY)
+
+    # Now we have to solve an overdetermined linear system of four equations with two variables.
+    A = numpy.array([[xupper[0], yupper[0]], [xlower[0], ylower[0]],
+                     [xupper[1], yupper[1]], [xlower[1], ylower[1]]])
+    b = numpy.array([dc_xupper, dc_xlower, dc_yupper, dc_ylower])
+    value, *_ = numpy.linalg.lstsq(A, b)
+
+    # A = numpy.array([[xupper[0], yupper[0]], [xlower[0], ylower[0]]])
+    # b = numpy.array([dc_xupper, dc_xlower])
+    # value = numpy.linalg.solve(A, b)
+    value = (value[0] * 1e-6, value[1] * 1e-6)  # µm --> m
+    return value
+
+
+class BeamShiftController(model.HwComponent):
+    """
+    Driver for the Thermofischer beam deflection controller.
+    This class provides the .shift VA containing a tuple of two floats which describe
+    the x and y beam offset in m in the stage coordinate system.
+
+    The conversion to internal ampere values (including scaling and rotation) is specified
+    through the MD_CALIB_BEAMSHIFT metadata (a 4x2 tuple, 4x (float, float)).
+    """
+
+    def __init__(self, name, role, port=None, serialnum=None, **kwargs):
+        """
+        :param port (str): port (e.g. /dev/ttyUSB0)
+        :param serialnum (str): serial number of RS485 adapter
+        """
+        # .hwVersion, .swVersion not available
+        model.HwComponent.__init__(self, name, role, **kwargs)
+
+        # Find port by RS485 adapter serial number
+        self._port = self._findDevice(port, serialnum)
+        self._serial = self._openSerialPort(self._port)
+
+        # Shift VA
+        # Range depends on metadata and will be checked in ._write_registers
+        self.shift = model.TupleContinuous((0, 0), range=((-1, -1), (1, 1)),
+                                           cls=(int, float), unit="m",
+                                           setter=self._setShift)
+
+    def _findDevice(self, ports=None, serialnum=None):
+        """
+        Look for a compatible device. Requires at least one of the arguments ports and serialnum.
+        ports (str): port (e.g. "/dev/ttyUSB0") or pattern for port ("/dev/ttyUSB*"), "/dev/fake" will start the simulator
+        serialnum (str): serial number
+        return (str): the name of the port used
+        raises:
+            ValueError: if no device on the ports with the given serial number is found
+        """
+        # At least one of the arguments ports and serialnum must be specified
+        if not ports and not serialnum:
+            raise ValueError("At least one of the arguments 'ports' and 'serialnum' must be specified.")
+
+        # For debugging purpose
+        if ports == "/dev/fake":
+            return ports
+
+        # If no ports specified, check all available ports
+        if ports:
+            names = list(serial.tools.list_ports.grep(ports))
+        else:
+            names = serial.tools.list_ports.comports()  # search all serial ports
+
+        # Look for serial number if available, otherwise make sure only one port matches the port pattern.
+        if serialnum:
+            for port in names:
+                if serialnum in port.description or serialnum in port.hwid:
+                    # "RS485" is in port.description, .hwid presumably contains serial number, TODO: check this!
+                    return port.device  # Found it!
+            else:
+                raise HwError("Beam controller device with serial number %s not found for port %s. " % (serialnum, names) +
+                              "Check the connection.")
+        else:
+            if len(names) == 1:
+                port = names[0]
+                return port.device
+            elif len(names) > 1:
+                raise HwError("Multiple ports detected for beam controller. Please specify a serial number.")
+            else:
+                raise HwError("Beam controller device not found for port %s. Check the connection." % ports)
+
+    def _openSerialPort(self, port):
+        if self._port == "/dev/fake":
+            return BeamShiftControllerSimulator()
+        else:
+            return ModbusSerialClient(method='rtu', port=port,
+                                      baudrate=BAUDRATE, timeout=TIMEOUT,
+                                      stopbits=STOPBITS, parity=PARITY,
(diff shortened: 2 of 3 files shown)


src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,129 @@
...
+        # Modbus connection
+        self._modbus = ModbusSerialClient(method='rtu', port=port,
+                                          baudrate=BAUDRATE, timeout=TIMEOUT,
+                                          stopbits=STOPBITS, parity=PARITY,
+                                          bytesize=BYTESIZE)
+
+        # Calculate range
+        range_amps = numpy.subtract(RANGE_AMPS, zero_offset_amps)
+        range_m = numpy.divide(range_amps, amps_per_m)
+        self.shift = model.TupleContinuous((0, 0), range_m,
```
philipwinklerdelmic, line 87: default should be at zero_offset / amps_per_meter

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,129 @@
...
+        self._zero_offset_amps = zero_offset_amps
+
+        # Find port by RS485 adapter serial number
+        for p in serial.tools.list_ports.comports():
+            if rs485_serial_num in p.hwid:
+                # Found it!
+                port = p
+                break
+        else:
+            raise ValueError("RS484 adapter with serial number %s not found!" % rs485_serial_num)
```
philipwinklerdelmic, line 76: probably better raise ioerror

philipwinklerdelmic:
We probably also need a configuration script to get the amps per meter and the zero offset parameters. Not sure how this will work though, needs some discussion.

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,129 @@
...
+        """
+        # .hwVersion, .swVersion not available
+        model.HwComponent.__init__(self, name, role, **kwargs)
+
+        self._amps_per_m = amps_per_m
+        self._zero_offset_amps = zero_offset_amps
+
+        # Find port by RS485 adapter serial number
+        for p in serial.tools.list_ports.comports():
+            if rs485_serial_num in p.hwid:
```
philipwinklerdelmic, line 71: Does that work? Need to test with a real rs485 cable.

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,129 @@
...
+    This class provides the .shift VA containing a tuple of two floats which describe
+    the x and y beam offset in m.
+    """
+
+    def __init__(self, name, role, amps_per_m, zero_offset_amps,
+                 rs485_serial_num, **kwargs):
+        """
+        :param port (str): port (e.g. /dev/ttyUSB0)
+        :param amps_per_m (float): conversion factor amplitude/meter
+        :param zero_offset_amps: amplitude for which beam deflection is 0
```
Kleijwegt, line 60: is this an int or and float? or are both ok?

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,129 @@
...
+        range_amps = numpy.subtract(RANGE_AMPS, zero_offset_amps)
+        range_m = numpy.divide(range_amps, amps_per_m)
+        self.shift = model.TupleContinuous((0, 0), range_m,
+                                           cls=(int, float), unit="m",
+                                           setter=self._setShift)
+
+    def _setShift(self, value):
+        """
+        value (float, float): shift from the center (in m).
+          It will always ensure that the shift is within the hardware limits.
```
philipwinklerdelmic, line 94: update docstring

tepals:
> First attempt at offset controller driver.
> 
> Do you think it makes sense to write a simulator and testcases for this driver? If so, what would you test for? We don't get any values back from the hardware if I understand it correctly, so there's not much we can do.

I thought it was possible to read values, I'm pretty sure someone mentioned this at some point. But I cannot find it in the documentation. Otherwise the simulator would just be the response you get after writing a register, not sure if that's useful.

philipwinklerdelmic:
> > First attempt at offset controller driver.
> > Do you think it makes sense to write a simulator and testcases for this driver? If so, what would you test for? We don't get any values back from the hardware if I understand it correctly, so there's not much we can do.
> 
> I thought it was possible to read values, I'm pretty sure someone mentioned this at some point. But I cannot find it in the documentation. Otherwise the simulator would just be the response you get after writing a register, not sure if that's useful.

I'm not even sure if we get any response, it looks like we just write the values to the 4 registers and that's it. It might be possible to read out the same registers with a different modbus command. That doesn't seem very useful to me though, if we test it on the real hardware we might as well just look at the beam.

Sabrina85:
> > > First attempt at offset controller driver.
> > > Do you think it makes sense to write a simulator and testcases for this driver? If so, what would you test for? We don't get any values back from the hardware if I understand it correctly, so there's not much we can do.
> > 
> > 
> > I thought it was possible to read values, I'm pretty sure someone mentioned this at some point. But I cannot find it in the documentation. Otherwise the simulator would just be the response you get after writing a register, not sure if that's useful.
> 
> I'm not even sure if we get any response, it looks like we just write the values to the 4 registers and that's it. It might be possible to read out the same registers with a different modbus command. That doesn't seem very useful to me though, if we test it on the real hardware we might as well just look at the beam.

I am not sure just looking on the beam is enough. How would you do that? Don't you want to confirm the values you wrote can also be read correctly?

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def _write_registers(self, payload):
+        """
+        Write to all four registers. Try to reconnect to device in case connection was lost.
+        :payload (list of 4 ints): register values (-x, -y, x, y)
+        """
+        if len(payload) != 4:
+            raise ValueError("write_registers received payload of invalid length %s != 4." % len(payload))
+        try:
+            # write all registers together (starting at lx)
```
tepals, line 121: I would suggest writing out lower x in this comment

src/odemis/driver/tfsbc.py
tepals, line 87: I think I'm still more a fan of naming this VA beamShift, but it's up to you :)

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,129 @@
...
+        returns (float, float): actual shift accepted
+        """
+        # Convert meter to amps
+        value = numpy.multiply(value, self._amps_per_m)
+
+        # Subtract offset (center amplitude)
+        value = numpy.subtract(value, self._zero_offset_amps)
+
+        # Convert amps to 16-bit int
+        value = numpy.subtract(value, RANGE_AMPS[0])
```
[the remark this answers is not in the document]
philipwinklerdelmic: It's part of the conversion to 16 bit.
Original range (-42e-3, 42e3)
New range (0, 65000) (approximately)
Conversion: (value + 42e3) * 65000 / (84e3)
* first map to (0, 84e3)
* then to (0, 65000)

philipwinklerdelmic:
> > > > First attempt at offset controller driver.
> > > > Do you think it makes sense to write a simulator and testcases for this driver? If so, what would you test for? We don't get any values back from the hardware if I understand it correctly, so there's not much we can do.
> > > 
> > > 
> > > I thought it was possible to read values, I'm pretty sure someone mentioned this at some point. But I cannot find it in the documentation. Otherwise the simulator would just be the response you get after writing a register, not sure if that's useful.
> > 
> > 
> > I'm not even sure if we get any response, it looks like we just write the values to the 4 registers and that's it. It might be possible to read out the same registers with a different modbus command. That doesn't seem very useful to me though, if we test it on the real hardware we might as well just look at the beam.
> 
> I am not sure just looking on the beam is enough. How would you do that? Don't you want to confirm the values you wrote can also be read correctly?

If the values are written properly to the registers, we should see that the beam changes (actually Andries already confirmed that with the example script that was provided).

Theoretically it could happen that the wrong values are written, but I think that if something goes wrong it's a different issue. We could run this
```write(registers, params)
ret = read(registers)
assert ret == params
``` 
Again, I'm not even certain if this works, but I can try. I would do it in the testcases though, not every time we run the command. Still, writing a simulator just for that seems a little unnecessary.

pieleric:
It's worthy to try to read the 4 values. Firstly, because then it lets you confirm you're probably talking to the right device (as devices with other protocols would probably not return precisely these 4 bytes). It might also be nice also to be able to read back the offset at start-up, although I'm not entirely sure whether that's useful. Maybe if there are other software which can affect these values, then it will come in handy.

Concerning the simulator, you definitely want one! Especially, as you said, it'd be very easy to simulate, so it's an even better reason to do it. With a simulator, you can immediately check you code, and check the test cases. In addition, that allows you to use this code into the full HW simulator.

The main last comment is about the calibration. Indeed, this will  be required. Probably the procedure will look a little like the calibration of the ASM voltage (ie, look at a grid of point in the calibration cam, with different voltage values, and measure the shifts). This also means that this will have to be done online. So the correction values cannot be passed as parameters, but should be metadata (cf the specification). So `updateMetadata()` would update the conversion. That has quite a few implications in what can be fixed at init. Also, you'll have to decide whether updating the metadata changes the hardware values, or the .shift content (and keeping the hardware value). I'd rather go for the later behaviour, as I like metadata change to actually not have any effect on the hardware (the code would need to write .shift in order for the device to actually move the beam).

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,130 @@
...
+        self._zero_offset_amps = zero_offset_amps
+
+        # Find port by RS485 adapter serial number
+        for p in serial.tools.list_ports.comports():
+            if rs485_serial_num in p.hwid:
+                # Found it!
+                port = p
+                break
+        else:
+            raise IOError("RS484 adapter with serial number %s not found!" % rs485_serial_num)
```
pieleric, line 76: This should be actually a HwError, to indicate to the backend, that it should try again.
Concerning the message string. RS484 -> RS485. Also, remove the "!", as it's not that much worrying for the user.

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,130 @@
...
+    """
+    Driver for the Thermofischer beam controller.
+    This class provides the .shift VA containing a tuple of two floats which describe
+    the x and y beam offset in m.
+    """
+
+    def __init__(self, name, role, amps_per_m, zero_offset_amps,
+                 rs485_serial_num, **kwargs):
+        """
+        :param port (str): port (e.g. /dev/ttyUSB0)
```
pieleric, line 58: `port` is not even used. I think the best would be to require this argument, and it can be a glob ("/dev/ttyACM*"). The serial number would be optional, and used to pick among the different ports. This would allow to not rely on the serial number if that's not useful. You might want to support also passing only the serial number and no "port", in which case it'd use the special function from `serial` (which is essentially a big glob).

BTW, please add a log line with the serial number of the port, so that if the user didn't specify it, it's easy to find out, and copy into the YAML file.

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,130 @@
...
+                                          stopbits=STOPBITS, parity=PARITY,
+                                          bytesize=BYTESIZE)
+
+        # Calculate range
+        range_amps = numpy.subtract(RANGE_AMPS, zero_offset_amps)
+        range_m = numpy.divide(range_amps, amps_per_m)
+
+        # Default value: zero offset (in m)
+        default_shift = tuple(numpy.divide(zero_offset_amps, amps_per_m))
+        self.beamShift = model.TupleContinuous(default_shift, range_m,
```
pieleric, line 90: I'm the one who had suggested just naming it "shift". There are two (linked) reasons:
* That's the name we already use in the phenom driver (and it's nice that all VAs with the same behaviour have the same name)
* It's on a component a role like "ebeam-control", so it should be relatively obvious that we are talking for *beam* shift.

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,130 @@
...
+
+        # Convert amps to 16-bit int
+        value = numpy.subtract(value, RANGE_AMPS[0])
+        value = int(numpy.multiply(value, 0xFFFF / (RANGE_AMPS[1] - RANGE_AMPS[0])))
+        # Note: in the example script a constant of 0.5 is added before converting
+        # to int to always round up (why?)
+
+        payload = [-value[0], -value[1], value[0], value[1]]
+
+        self._write_registers(payload)
```
pieleric, line 112: A setter *must* return the accepted value. Here that's useful here as it's not the original value requested, but the value rounded to an int (and converted back to meters).

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,130 @@
...
+                                          baudrate=BAUDRATE, timeout=TIMEOUT,
+                                          stopbits=STOPBITS, parity=PARITY,
+                                          bytesize=BYTESIZE)
+
+        # Calculate range
+        range_amps = numpy.subtract(RANGE_AMPS, zero_offset_amps)
+        range_m = numpy.divide(range_amps, amps_per_m)
+
+        # Default value: zero offset (in m)
+        default_shift = tuple(numpy.divide(zero_offset_amps, amps_per_m))
```
pieleric, line 89: All these "numpy" functions are a bit of an eyesore. => Convert the two values to a numpy array at first, and then use "+" or "/" as professional mathematicians do ;-)

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,130 @@
...
+        if len(payload) != 4:
+            raise ValueError("write_registers received payload of invalid length %s != 4." % len(payload))
+        try:
+            # write all registers together (starting at lower x register (=0x01))
+            self._modbus.write_registers(BEAMDEFL_LX, payload, unit=SLAVE_UNIT)
+        except IOError:
+            logging.warning("Failed to write registers, trying to reconnect...")
+            self._modbus.close()
+            self._modbus.connect()
+            logging.info("Recovered device.")
```
pieleric, line 128: If you want to handle reconnection, let's do it completely. Please move this to a separate function, and inside:
* Update the .state to indicate there is a connection error
* Try reconnecting *in a loop*
* Put back the state when it's reconnected.

src/odemis/driver/tfsbc.py
philipwinklerdelmic: I thought it was impossible to connect to the device without a serial number because there is no getVersion() command. This is why I deleted port from the argument list because I didn't see the point in keeping it as an argument if we always provide the serial number.

Is there a way not to rely on the serial number? As discussed, there is at least one other device (stage driver) using RS485.

pieleric: Well, there is another device using RS485 *in the first system that will use this driver*. It's best to keep it a little generic too. Especially for testing or debugging, it'll be easier if it's possible to just force the port. If you want to keep the logic simple, accept either a port name *or* a serial number.

philipwinklerdelmic:
Updated PR, @pieleric could you have a look tomorrow?

src/odemis/driver/test/tfsbc_test.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def test_rotation(self):
+        """
+        Test if setting shift works with non-zero rotation.
+        """
+        self.bc._metadata[model.MD_CALIB_ROTATION] = math.pi / 4  # rad
+        shift = (5e-9, 8e-9)
+        self.bc.shift.value = shift
+        self.assertTupleAlmostEqual(self.bc.shift.value, shift, places=9)
+        self.bc._metadata[model.MD_CALIB_ROTATION] = 0  # rad
```
tepals, line 87: If you change back the rotation, does the shift than also update? If yes, you could also test that.

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,275 @@
...
+            rotation = 0
+
+        # Convert to numpy.array to simplify calculation
+        value = numpy.array(value, dtype=numpy.float64)
+        gain = numpy.array(gain, dtype=numpy.float64)  # amps / m
+        offset = numpy.array(offset, dtype=numpy.float64)  # amps
+        amp_range = numpy.array(RANGE_AMPS, dtype=numpy.float64)
+
+        # Convert to beam coordinate system (rotation)
+        value = value / math.cos(rotation)
```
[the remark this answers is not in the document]
philipwinklerdelmic: you're completely right, that's not how you do a rotation... I'll fix it

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,275 @@
...
+
+        # Convert to amperes (with gain)
+        value *= gain
+
+        # Add offset
+        value -= offset  # subtract offset (center amplitude)
```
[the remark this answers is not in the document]
philipwinklerdelmic: I don't know, I thought the offset was just a fixed value which needs to be added. We still need to figure out what it's supposed to do.

src/odemis/driver/test/tfsbc_test.py
philipwinklerdelmic: No, it shouldn't update. It's just to put it back for the next testcase.

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,277 @@
...
+                             "first.")
+        offset = self._metadata.get(model.MD_CALIB_TRANSLATION)
+        if not offset:
+            logging.warning("No deflection offset specified, assuming 0 amps.")
+            offset = (0, 0)
+
+        rotation = self._metadata.get(model.MD_CALIB_ROTATION)
+        if not offset:
+            logging.warning("No rotation specified, assuming 0 rad.")
+            rotation = 0
```
pieleric, line 150: You could initialise this metadata to the default value, so that if nothing is done, it'd be visible in the "odemis-cli" (ie .getMetadata). 
The fact that you cannot set the value if there is no calibration is a little bit arguable. One issue I imagine is that if the VA is set as persistent (which it shouldn't) in the backend, then the whole component cannot start (because VAs are set before metadata). That said, maybe that's what should happen, as it's kind of meaningly. It's also weird that there is a value, but you cannot set it. Maybe don't raise an error if someone set the same value? I do agree it's also weird to attempt changing the value without any calibration.
Hopefully, in reality, we'll have a pretty good value by default in the YAML file, and this code will never be triggered anyway.

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,277 @@
...
+
+        # Convert to 16-bit integer
+        conversion_factor = 0xFFFF / (amp_range[1] - amp_range[0])
+        valuePos = (value - amp_range[0]) * conversion_factor
+        valueNeg = (-value - amp_range[0]) * conversion_factor
+        # Note: in the example script a constant of 0.5 is added before converting
+        # to int to always round up (why?)
+        register_values = [int(valueNeg[0]), int(valueNeg[1]), int(valuePos[0]), int(valuePos[1])]
+
+        # Read previous value of registers
```
pieleric, line 176: Please indicate in the comment that it's just for debugging purpose.

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,277 @@
...
+
+        # Add offset
+        value -= offset  # subtract offset (center amplitude)
+
+        # Convert to 16-bit integer
+        conversion_factor = 0xFFFF / (amp_range[1] - amp_range[0])
+        valuePos = (value - amp_range[0]) * conversion_factor
+        valueNeg = (-value - amp_range[0]) * conversion_factor
+        # Note: in the example script a constant of 0.5 is added before converting
+        # to int to always round up (why?)
```
pieleric, line 173: No, adding a constant of 0.5 doesn't always round *up*, it rounds to the closest int (as C int() rounds down). In Python there is a function for that, cleverly named `round()`.
I'd suggest to use it too. Note that you still need `int` to force it being an int: `int(round(x))`.

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,277 @@
...
+
+        self._write_registers(register_values)
+
+        # Read back values to see if everything went fine
+        ret = self._read_registers()
+        if ret != register_values:
+            logging.error("Register values after writing %s != %s." % (ret, register_values))
+
+        # Convert rounded shift back to meter for return value
+        value = (numpy.array(ret[2:]) / conversion_factor + amp_range[0] + offset) / gain * math.cos(rotation)
```
pieleric, line 188: Are you sure you can do that with rotation?! I think you need to use matrix multiplication (if you want to write on a single line).

src/odemis/driver/tfsbc.py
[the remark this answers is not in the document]
pieleric: Before or after, both are possible as long as you document it. I also have hard time to imagine what's the easiest for the user. In the inverted operation you'll have to do it in the opposite order, that's the important part. BTW, I'd find it easier (ie, less surprising) to have the offset *added*.

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,277 @@
...
+        value = numpy.array(value, dtype=numpy.float64)
+        gain = numpy.array(gain, dtype=numpy.float64)  # amps / m
+        offset = numpy.array(offset, dtype=numpy.float64)  # amps
+        amp_range = numpy.array(RANGE_AMPS, dtype=numpy.float64)
+
+        # Convert to amperes (with gain)
+        value *= gain
+
+        # Add offset
+        value -= offset  # subtract offset (center amplitude)
```
pieleric, line 166: Put both on a single line. Maybe even including the rotation.

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,277 @@
...
+    def _write_registers(self, values):
+        """
+        Write to all four registers. Try to reconnect to device in case connection was lost.
+        :values (list of 4 ints): register values (-x, -y, x, y)
+        """
+        if len(values) != 4:
+            raise ValueError("write_registers received payload of invalid length %s != 4." % len(values))
+
+        # Check if values are in allowed range
+        if not all([0 < val < 0xFFFF for val in values]):
```
pieleric, line 200: You actually can drop the square brackets.

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,277 @@
...
+        self.r1 = values[1]
+        self.r2 = values[2]
+        self.r3 = values[3]
+        return SimplifiedModbusObject([])
+
+    def read_holding_registers(self, start_register, unit=None):
+        return SimplifiedModbusObject([self.r0, self.r1, self.r2, self.r3])
+
+
+class SimplifiedModbusObject():
```
pieleric, line 271: Please still inherit from "object", to be compatible with Python 2.

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,277 @@
...
+                logging.info("Recovered device.")
+                break
+            except IOError:
+                continue
+        else:
+            raise IOError("Failed to reconnect to beam deflection controller.")
+        self.state._set_value(model.ST_RUNNING, force_write=True)
+
+
+class BeamControllerSimulator:
```
pieleric, line 249: Please still inherit from "object", to be compatible with Python 2.

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,277 @@
...
+# Modbus level addresses
+SLAVE_UNIT = 2
+
+# Modbus registers
+BEAMDEFL_LX = 0  # lower x beam deflection element control
+BEAMDEFL_LY = 1  # lower y beam deflection element control
+BEAMDEFL_UX = 2  # upper x beam deflection element control
+BEAMDEFL_UY = 3  # upper y beam deflection element control
+
+RANGE_AMPS = [-42.2e-3, 42.2e-3]
```
pieleric, line 50: Add a small comment explaining what are these values.

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,277 @@
...
+Foundation.
+
+Odemis is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+PURPOSE. See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with
+Odemis. If not, see http://www.gnu.org/licenses/.
+"""
+
```
pieleric, line 22: Please import division from the future, for Python 2 compatibility.

src/odemis/driver/test/tfsbc_test.py
```diff
@@ -0,0 +1,92 @@
...
+Foundation.
+
+Odemis is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+PURPOSE. See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with
+Odemis. If not, see http://www.gnu.org/licenses/.
+"""
+
```
pieleric, line 22: Please import division from the future.

src/odemis/driver/test/tfsbc_test.py
```diff
@@ -0,0 +1,92 @@
...
+    """
+    Tests the beam controller driver.
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        cls.bc = tfsbc.BeamController("DC Offset", None, PORT)
+        cls.bc._metadata[model.MD_CALIB_SCALE] = [3e6, 3e6]  # 3e6 amp/m = 3 mamp / nm
+        cls.bc._metadata[model.MD_CALIB_TRANSLATION] = [0, 0]
+        cls.bc._metadata[model.MD_CALIB_ROTATION] = 0  # rad
```
pieleric, line 52: Use updateMetadata() for that.

src/odemis/driver/test/tfsbc_test.py
```diff
@@ -0,0 +1,92 @@
...
+    def test_read_write(self):
+        vals = [27000, 37000, 20000, 44000]
+        self.bc._write_registers(vals)
+        ret = self.bc._read_registers()
+        self.assertTupleAlmostEqual(vals, list(ret), places=1)
+
+    def test_set_shift(self):
+        startt = time.time()
+        shift = (5e-9, 8e-9)
+        self.bc.shift.value = shift
```
pieleric, line 74: Check that the value in .shift is almost equal to the value requested.

src/odemis/driver/test/tfsbc_test.py
```diff
@@ -0,0 +1,92 @@
...
+        self.assertLess(time.time() - startt, 0.01, "Reading/writing took more than 10 ms.")
+        logging.debug("Shift value set to %s", self.bc.shift.value)
+        self.assertTupleAlmostEqual(self.bc.shift.value, shift, places=9)
+
```
pieleric, line 78: Please add a test to check nothing crazy happens when you put very large shift value. It should be clipped to the value corresponding to the device maximum.

src/odemis/driver/test/tfsbc_test.py
philipwinklerdelmic: Are you sure that's a good idea? I think the .shift VA should reflect the state of the hardware. I would suggest do nothing when updating the metadata and only apply the rotation at the next change of .shift.
What do you think?

pieleric: Well, at least we agree that .shift should reflect the state of the hardware. This "reflection" takes into account the calibration values. So if you update the calibration values, the .shift value should change. If you don't do this, you get some weird behaviour (which I have experienced in some other software). Like, you set the shift to a value X, => the hardware goes to position X'. Then you update the calibration (nothing else happens). You read the shift, it's still value X. Finally, you set the shift to value X *again*, and suddenly the hardware moves to position X" ! Reading an attribute and immediately setting it to the same value shouldn't cause any hardware change.

So, for the driver to behave nicely, I think that after the calibration is updated, the shift value should be computed back based on the hardware value + the calibration.

philipwinklerdelmic: I see your point! I'll make the change.

Actually, how do I do this? I guess I'll have to overwrite updateMetadata and from there change the VA. But I can't call the setter, otherwise the value will be written to the hardware registers. Is there a way to modify the value without triggering the setter function?

philipwinklerdelmic: Just to make sure I understand completely: When we write to shift.value, we enter the value in the stage coordinate system without offset. Then, when reading the value back, it's going to be in the beam coordinate system (rotation), and with the offset (and modified slightly to account for the rounding when converting to int16).
So we might have something like
```
controller.shift.value = (5e-9, 3e-9)
print(controller.shift.value) --> will output (8.32e-9, 2.4e-9)
```
(completely made up numbers)

philipwinklerdelmic: If you read the previous comment I just deleted, please ignore, I misunderstood.

src/odemis/driver/tfsbc.py
philipwinklerdelmic: How about something like this:
```
        gain = self._metadata.get(model.MD_CALIB_SCALE)
        if not gain:
            gain = DEFAULT_GAIN
            logging.error("No gain specified, assuming default of %s." % gain +
                          "Calibration is required to set shift in a meaningful way.")
            self.updateMetadata({model.MD_CALIB_SCALE: gain})
```

philipwinklerdelmic: ok, without the last line, that doesn't make sense (otherwise the error will not be triggered the next time)

philipwinklerdelmic:
The code was tested today on the hardware, the communications works!

There's still a couple of things that need attention:
* RS485 serial number --> we need to check that looking it up in .hwid works (or if that's a different number)
* Logging previous register value: reading takes about 20 ms, writing about 14 ms, might be too long?
* Offset parameter is still not completely clear (should it be rotated? etc)

I also still need to make sure that (see chats above)
* after changing rotation, the .shift VA is updated
* we're happy with the behaviour if no calibration has been performed

Since I made quite a few changes after testing, the code needs to be tested again before merging.

src/odemis/driver/tfsbc.py
```diff
@@ -209,8 +219,7 @@ def _setShift(self, value):
         self._write_registers(register_values)
 
         # Scale back to multiprobe units and convert back to multiprobe coordinates
-        value = (value[0] / gain[0], value[1] / gain[1])
-        value = transform_coordinates(value, -rotation, (-offset[0], -offset[1]), LH, RH)
+        value = transform_coordinates(value, rotation, numpy.array(gain), offset, LH, RH, inverse=True)
```
pieleric, line 222: I have a hard time to convince myself it's not an overkill. You call this function twice, once with arguments RH,LH,inverse=False and once with LH,RH,inverse=True. It feels that the inverse is only doing half of its work.
Moreover, assuming that the rotation,gain, offset have been specifically computed for this component, I don't see why you'd need to switch between RH and LH. Isn't it like having a gain with a -Y? 
Also, I feel that your original idea of using SimilarityTransform(rotation,gain,offset) should work. Did you try calling `.inverse()` of this transformation, and this new transform didn't allow you to get the original values? That'd be a bug in that code then.

philipwinklerdelmic: **Scaling Transform**
It turned out that the DC coordinate system is not completely orthogonal. The angle between the x and y axes is less than 90 degrees. This means we need a nonlinear transform. There is a plugin that calculates this transform in the external software. The goal is now to use this transform. This means that I have to change the driver code again. I don't think that the alignment code is necessary anymore.
Probably this nonlinearity was the reason that the ScalingTransform was not working. Doing the rotation + scaling manually gave much better results for some reason (not sure why though, the parameters of the ScalingTransform were the same). I don't think there is a bug.

**transform_parameters**
Why do you think the inverse is doing half of the work? The main problem is, that the order of the scaling and rotation matters. We always scale in the dc frame, so for the inverse we need to do scaling first, then rotation, for the regular transformation it's the other way around.
We need to switch between RH and LH because the value which is given to the driver is in RH, but the dc hardware operates in LH.

pieleric: **Scaling Transform**
So let's see, hopefully this transform will really be available soon. BTW, I don't get how "manually" you can call ScalingTransform with the same parameters and get different results. The transform is very precisely defined mathematically. There is no reason for a computer to interpret it differently than you.

**transform_parameters**
The way that Transform classes are defined is such that if you use the inverse() version, it directly works. The rotation/scaling order is taken care by the inverse() computation.
Concerning the RH/LH, I understand that they are different, but I don't understand why you have it to do so. My feeling is that the calibration is specifically computed for a given hardware, so the numbers should directly fit. To switch between LH and RH, isn't it just a matter of putting a negative sign on the Y scale? I think it'd be better that the calibration algorithm knows (or detects) that an axis is inverted, instead of hard-coding it in the driver.

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,357 @@
...
+    dc_yupper = value[0] * xupper[1] + value[1] * yupper[1]
+    dc_ylower = value[0] * xlower[1] + value[1] * ylower[1]
+
+    currUX = DCREVERSCOEF * DCROTUPPXX * dc_xupper
+    currLX = DCREVERSCOEF * DCROTLOWXX * dc_xlower
+    currUY = DCREVERSCOEF * DCROTUPPYY * dc_yupper
+    currLY = DCREVERSCOEF * DCROTLOWYY * dc_ylower
+
+    for current in [currLX, currLY, currUX, currUY]:
+        if current < C_MIN_DBL_SHIFT or current > C_MAX_DBL_SHIFT:
```
pieleric, line 107: You can write slightly more clearly:
`if not C_MIN_DBL_SHIFT <= current <= C_MAX_DBL_SHIFT:`

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,357 @@
...
+BEAMDEFL_UX = 2  # upper x beam deflection element control
+BEAMDEFL_UY = 3  # upper y beam deflection element control
+
+# Transformation constants
+DCREVERSCOEF = -1
+DCROTUPPXX = 1
+DCROTUPPYY = 1
+DCROTLOWXX = -1
+DCROTLOWYY = 1
+
```
pieleric, line 56: Is there a reason you write it this way? Does it come from other code? All the values are always multiplied by DCREVERSCOEF, so at least it could be simplified by just multiplying them by -1 here.

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,357 @@
...
+    This class provides the .shift VA containing a tuple of two floats which describe
+    the x and y beam offset in m in the stage coordinate system.
+
+    The conversion to internal ampere values (including scaling and rotation) is specified
+    through the MD_CALIB_BEAMSHIFT metadata (a 4x2 tuple, 4x (float, float)).
+    """
+
+    def __init__(self, name, role, port=None, serialnum=None, **kwargs):
+        """
+        :param port (str): port (e.g. /dev/ttyUSB0)
```
pieleric, line 166: Please indicate here the docstring from findDevice (about the special value `/dev/fake` and pattern).

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,357 @@
...
+                                      bytesize=BYTESIZE)
+
+    def _setShift(self, value):
+        """
+        :param value (float, float): x, y shift from the center (in m)
+        """
+        xlower, ylower, xupper, yupper = self._metadata.get(model.MD_CALIB_BEAMSHIFT)
+
+        if None in (xlower, ylower, xupper, yupper):
+            raise ValueError("Cannot set shift, MD_CALIB_BEAMSHIFT metadata not specified.")
```
pieleric, line 239: I don't think this works. `dict.get()` would return None, which cannot be mapped to a 4-tuple. You'll not even reach here.
I'd suggest to surround the above line with a try/except KeyError.

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,357 @@
...
+                logging.info("Recovered device.")
+                break
+            except IOError:
+                continue
+        else:
+            raise IOError("Failed to reconnect to beam deflection controller.")
+        self.state._set_value(model.ST_RUNNING, force_write=True)
+
+    def updateMetadata(self, md):
+        logging.debug("Updating metadata %s." % md)
```
pieleric, line 313: Drop this log here, it's a little too much. If you want to know the change of beamshift, put it in the if.

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,357 @@
...
+        if model.MD_CALIB_BEAMSHIFT in md and self._metadata.get(model.MD_CALIB_BEAMSHIFT):
```
pieleric, line 314: Even if there is not metadata yet, there should be a check that the new metadata is the correct format.
=> if it's not of length 4, and each element is not of length 2, each sub-element being a number => raise a ValueError.

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,357 @@
...
+        # .hwVersion, .swVersion not available
+        model.HwComponent.__init__(self, name, role, **kwargs)
+
+        # Find port by RS485 adapter serial number
+        self._port = self._findDevice(port, serialnum)
+        self._serial = self._openSerialPort(self._port)
+
+        # Shift VA
+        # Range depends on metadata and will be checked in ._write_registers
+        self.shift = model.TupleContinuous((0, 0), range=((-1, -1), (1, 1)),
```
pieleric, line 178: Please add a comment that for now it's not correct, until the metadata is set.

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,357 @@
...
+        else:
+            raise IOError("Failed to reconnect to beam deflection controller.")
+        self.state._set_value(model.ST_RUNNING, force_write=True)
+
+    def updateMetadata(self, md):
+        logging.debug("Updating metadata %s." % md)
+        if model.MD_CALIB_BEAMSHIFT in md and self._metadata.get(model.MD_CALIB_BEAMSHIFT):
+            # Transform to register values with old metadata
+            xlower, ylower, xupper, yupper = self._metadata.get(model.MD_CALIB_BEAMSHIFT)
+            vals = transform_coordinates(self.shift.value, xlower, ylower, xupper, yupper)
```
pieleric, line 317: On the first time, the shift value is initialised at 0,0, but that's not the actual value. => In this case, read the values from the hardware. A simple alternative would be to always read the values from the hardware. (so no need to use the current metadata).

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,357 @@
...
+        Look for a compatible device. Requires at least one of the arguments ports and serialnum.
+        ports (str): port (e.g. "/dev/ttyUSB0") or pattern for port ("/dev/ttyUSB*"), "/dev/fake" will start the simulator
+        serialnum (str): serial number
+        return (str): the name of the port used
+        raises:
+            ValueError: if no device on the ports with the given serial number is found
+        """
+        # At least one of the arguments ports and serialnum must be specified
+        if not ports and not serialnum:
+            raise ValueError("At least one of the arguments 'ports' and 'serialnum' must be specified.")
```
pieleric, line 193: The main class is going to raise this error if the user hasn't specified `port` or `serialnum`. As here it's spelled `ports`, this message is going to be confusing. I think that the easiest would be to just name the argument here also `port`.

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,357 @@
...
+        :param port (str): port (e.g. /dev/ttyUSB0)
+        :param serialnum (str): serial number of RS485 adapter
+        """
+        # .hwVersion, .swVersion not available
+        model.HwComponent.__init__(self, name, role, **kwargs)
+
+        # Find port by RS485 adapter serial number
+        self._port = self._findDevice(port, serialnum)
+        self._serial = self._openSerialPort(self._port)
+
```
pieleric, line 175: Please, try to communicate to the device here in the init, to double check that it's actually present and responding. You could just read the registers (and discard the result). If this read fails, raise an HwError.
Even better would be if the device actually support "Read Device Identification" and there was an easy way to query this, but that doesn't seem the case with pymodbus.

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,357 @@
...
+        if ports:
+            names = list(serial.tools.list_ports.grep(ports))
+        else:
+            names = serial.tools.list_ports.comports()  # search all serial ports
+
+        # Look for serial number if available, otherwise make sure only one port matches the port pattern.
+        if serialnum:
+            for port in names:
+                if serialnum in port.description or serialnum in port.hwid:
+                    # "RS485" is in port.description, .hwid presumably contains serial number, TODO: check this!
```
pieleric, line 209: Yes, please check. Having both is just weird. Shouldn't it be `serial_number`?

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,357 @@
...
+        still needs the settings to be applied.
+        """
+        num_it = 5
+        self.state._set_value(model.HwError("Beam deflection controller disconnected"), force_write=True)
+        logging.warning("Failed to write registers, trying to reconnect...")
+        for i in range(num_it):
+            try:
+                self._serial.close()
+                self._serial.connect()
+                logging.info("Recovered device.")
```
pieleric, line 304: That seems a very optimistic code to *reconnect* and check that the device is answering again...
Please implement and test this function.

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,357 @@
...
+
+    # Now we have to solve an overdetermined linear system of four equations with two variables.
+    A = numpy.array([[xupper[0], yupper[0]], [xlower[0], ylower[0]],
+                     [xupper[1], yupper[1]], [xlower[1], ylower[1]]])
+    b = numpy.array([dc_xupper, dc_xlower, dc_yupper, dc_ylower])
+    value, *_ = numpy.linalg.lstsq(A, b)
+
+    # A = numpy.array([[xupper[0], yupper[0]], [xlower[0], ylower[0]]])
+    # b = numpy.array([dc_xupper, dc_xlower])
+    # value = numpy.linalg.solve(A, b)
```
pieleric, line 149: Non-used code? => drop.

src/odemis/driver/test/tfsbc_test.py
```diff
@@ -0,0 +1,127 @@
...
+# @skip("skip")
+class TestBeamShiftController(unittest.TestCase):
+    """
+    Tests the beam controller driver.
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        cls.bc = tfsbc.BeamShiftController("DC Offset", None, PORT, "RS485")
```
pieleric, line 51: RS485 is not a the serial number ;-)

src/odemis/driver/test/tfsbc_test.py
```diff
@@ -0,0 +1,127 @@
...
+        so the effect can be seen on the hardware.
+        """
+        shift = (0, 0)
+        self.bc.shift.value = shift
+        self.assertTupleAlmostEqual(self.bc.shift.value, shift, places=9)
+        time.sleep(1)
+
+        shift = (-5e-6, 0)
+        self.bc.shift.value = shift
+        self.assertTupleAlmostEqual(numpy.subtract(self.bc.shift.value, shift), (0, 0), places=7)
```
pieleric, line 88: Why do you have to do a subtract here? `assertTupleAlmostEqual()` is a *you* function defined here, so make sure it works as you need.

src/odemis/driver/test/tfsbc_test.py
```diff
@@ -0,0 +1,127 @@
...
+        time.sleep(1)
+
+        shift = (0, -5e-6)
+        self.bc.shift.value = shift
+        self.assertTupleAlmostEqual(numpy.subtract(self.bc.shift.value, shift), (0, 0), places=7)
+
+        shift = (0, 0)
+        self.bc.shift.value = shift
+        self.assertTupleAlmostEqual(numpy.subtract(self.bc.shift.value, shift), (0, 0), places=7)
+
```
pieleric, line 103: Please add a test for when setting a X or Y shift too large.

pieleric (changes requested):
I don't have any major remarks. However many small remarks, especially about handling corner cases.

src/odemis/driver/tfsbc.py
philipwinklerdelmic: I took it from the other code. I also don't understand why they do it that way, but there must be a reason, so I just left it.

src/odemis/driver/test/tfsbc_test.py
```diff
@@ -0,0 +1,127 @@
...
+        val = (0, 4e-6)
+        md = ((-0.00027788219369730165, 0.0013604844623992785),
+              (-0.0013604844623992785, -0.00027788219369730165),
+              (0.00012699407486043473, -0.0006217507619527259),
+              (0.0006217507619527259, 0.00012699407486043473))
+        expected = [0x6f7e, 0x835f, 0x7874, 0x7e75]  # from testing with example script
+        ret = tfsbc.transform_coordinates(val, *md)
+        self.assertTupleAlmostEqual(ret, expected, places=5)
+        rev = tfsbc.transform_coordinates_reverse(ret, *md)
+        self.assertTupleAlmostEqual(val, rev, places=5)
```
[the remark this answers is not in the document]
philipwinklerdelmic: Indeed, the first assertion (with the conversion to integers) can be exact. The reverse conversion, however, is not exact. I'm solving an overdetermined system of linear equations in the reverse function, so there are some inaccuracies.

src/odemis/driver/tfsbc.py
pieleric: If you copied it from the "recommended code", then it's fine to leave it as-is. Especially, if they update their code, it should be easy to update it here too.

src/odemis/driver/test/tfsbc_test.py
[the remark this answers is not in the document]
pieleric: Oh yes, sure. I now realise that `val` contains float, so that's fine.

src/odemis/model/_metadata.py
```diff
@@ -178,6 +178,9 @@
 # dict (int (resolution X) -> dict (float (dwell time) -> tuple of 4 floats (correction factors)))
 MD_SHIFT_LOOKUP = "Pixel shift compensation table"
 
+# The following metadata is used for the calibration of the tfsbc.BeamShiftController
+MD_CALIB_BEAMSHIFT = "Beamshift calibration"  # 4x2 tuple, 4x (float, float), xlower, ylower, xupper, yupper
```
pieleric, line 182: We now have a generic `MD_CALIB`, which you can use in your class. Please document well xlower, ylower, xupper, yupper in the docstring of the class.
You should be able to drop entirely the changes on this file.

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,376 @@
...
+                    raise ValueError("Invalid MD_CALIB_BEAMSHIFT metadata %s: 4 tuples required." % (bs,))
+                if not all(len(val) == 2 for val in bs):  # each of the 4 values is a tuple of 2
+                    raise ValueError("Invalid MD_CALIB_BEAMSHIFT metadata %s: Two values per tuple required." % (bs,))
+                if not all(all([type(val) in (int, float) for val in tup]) for tup in bs):  # each element is a number
+                    raise ValueError("Invalid MD_CALIB_BEAMSHIFT metadata %s: Values must be numbers." % (bs,))
+            except Exception as ex:
+                raise ValueError("Invalid MD_CALIB_BEAMSHIFT metadata %s, ex: %s" % (bs, ex,))
+
+            # If MD_CALIB_BEAMSHIFT was already available, transform current shift to fit new metadata value.
+            if self._metadata.get(model.MD_CALIB_BEAMSHIFT):
```
pieleric, line 334: You could write more explicitly `if model.MD_CALIB in self._metadata`.
However, I don't understand why doing it only if there was already some calibration. On the first time that the component gets a calibration, the hardware already has some values, so there is already a shift => directly use `md[model.MD_CALIB]`.

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,376 @@
...
+    def updateMetadata(self, md):
+        if model.MD_CALIB_BEAMSHIFT in md:
+            # Check format
+            bs = md[model.MD_CALIB_BEAMSHIFT]
+            try:
+                if not len(bs) == 4:  # 4 tuples required
+                    raise ValueError("Invalid MD_CALIB_BEAMSHIFT metadata %s: 4 tuples required." % (bs,))
+                if not all(len(val) == 2 for val in bs):  # each of the 4 values is a tuple of 2
+                    raise ValueError("Invalid MD_CALIB_BEAMSHIFT metadata %s: Two values per tuple required." % (bs,))
+                if not all(all([type(val) in (int, float) for val in tup]) for tup in bs):  # each element is a number
```
pieleric, line 328: You don't need to put a `[]` around the generator.
You shouldn't use `type` to check the class, but `isinstance()`, which also handles the case of sub-classes (eg, numpy.float).
So you could write something like:
`if not all(all(isinstance(val, (int, float)) for val in tup) for tup in bs):`

Note that you could even write, but it's not readable:
`if not all(isinstance(val, (int, float)) for tup in bs for val in tup):`

src/odemis/driver/tfsbc.py
```diff
@@ -0,0 +1,277 @@
...
+    three metadata:
+        * MD_CALIB_SCALE (float, float): conversion factor ampere/meter (gain)
+        * MD_CALIB_TRANSLATION (float, float): offset in amps
+        * MD_CALIB_ROTATION (float): angle in radians
+    """
+
+    def __init__(self, name, role, port=None, serialnum=None, **kwargs):
+        """
+        :param port (str): port (e.g. /dev/ttyUSB0)
+        :param serialnum (str): serial number of RS485 adapter
```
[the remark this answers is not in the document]
pieleric: You should also indicate that just `serialnum` or just `port` is fine, and both is also fine.

pieleric (changes requested):
Just minor comments now.

philipwinklerdelmic:
Updated

PR #1424, opened 2020-05-22
bugreporter: also log memory and disk usage

pieleric:
It's handy to diagnose issues related to out-of-memory, or disk full.

### DIFF
--- a/src/bugreporter/odemis_bugreporter.py
+++ b/src/bugreporter/odemis_bugreporter.py
@@ -310,6 +310,8 @@ class OdemisBugreporter(object):
             outputs["ps.txt"] = self._get_future_output(['/bin/ps', 'aux'], timeout=60)
             outputs["uname.txt"] = self._get_future_output(['/bin/uname', '-a'], timeout=5)
             outputs["ip.txt"] = self._get_future_output(['/bin/ip', 'address'], timeout=30)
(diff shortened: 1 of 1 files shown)


src/bugreporter/odemis_bugreporter.py
```diff
@@ -310,6 +310,8 @@ def compress_files(self):
             outputs["ps.txt"] = self._get_future_output(['/bin/ps', 'aux'], timeout=60)
             outputs["uname.txt"] = self._get_future_output(['/bin/uname', '-a'], timeout=5)
             outputs["ip.txt"] = self._get_future_output(['/bin/ip', 'address'], timeout=30)
+            outputs["memory.txt"] = self._get_future_output(['/usr/bin/free'], timeout=5)
+            outputs["df.txt"] = self._get_future_output(['/bin/df'], timeout=10)
```
Kleijwegt, line 314: Maybe also storing the dmesg output could be useful?

pieleric: Yes, we actually do this indirectly, by saving "/var/log/syslog", which contains the dmesg output.

Kleijwegt: Aaah i didn't spot that line, great!

PR #1425, opened 2020-05-27
driver andorshrk: don't autoreconnect for ZeroSlitPosition commands

pieleric:
The old firmwares used to report a COMMUNICATION_ERROR when calling this
function. This means that with the new auto-reconnction the hardware would
be rebooted every time this function is called. Chaos ensues.


pieleric:
For this kind of auto reconnection on error, it's pretty hard to write test cases. It would involve a relatively large extension of the simulator, and still would only be semi-convincing. So all this code is really best tested manually by physically torturing the hardware.

PR #1429, opened 2020-05-28
Color map feature

amuskens:


src/odemis/acq/stream/_base.py
```diff
@@ -227,7 +227,11 @@ def __init__(self, name, detector, dataflow, emitter, focuser=None, opm=None,
         self.auto_bc_outliers = model.FloatContinuous(100 / 256, range=(0, 40))
         self.auto_bc_outliers.subscribe(self._onOutliers)
 
-        self.tint = model.ListVA((255, 255, 255), unit="RGB")  # 3-int R,G,B
+        # The tint VA could be either:
+        # - a list tuple RGB value (for a tint) or
+        # - a matplotlib.colors.Colormap object for a custom color map
+        # - a string of value "fitrgb" to indicate fit RGB color mapping
+        self.tint = model.VigilantAttribute((255, 255, 255))
```
Kleijwegt, line 234: Why don't you use a TupleContinious to also include a max value for this VA?

amuskens: So the reason is so that the VA could contain either:
- a list tuple RGB value (for a tint) or
- a matplotlib.colors.Colormap object for a custom color map
- a string of value "fitrgb" to indicate fit RGB color mapping
so it needs to be generic

src/odemis/gui/conf/data.py
```diff
@@ -896,6 +897,19 @@ def get_hw_config(hw_comp, hw_settings):
...
+COLORMAPS = OrderedDict([
+    ("viridis", cm.get_cmap("viridis")),
+    ("inferno", cm.get_cmap("inferno")),
+    ("plasma", cm.get_cmap("plasma")),
+    ('magma', cm.get_cmap('magma')),
+    ('spring', cm.get_cmap('spring')),
+    ('summer', cm.get_cmap('summer')),
+    ('autumn', cm.get_cmap('autumn')),
+    ('winter', cm.get_cmap('winter')),
+])
```
pieleric, line 910: Let's add upper cases to the names, to make it nice to look at.

src/odemis/gui/util/img.py
```diff
@@ -1756,17 +1759,49 @@ def draw_legend_multi_streams(images, buffer_size, buffer_scale,
...
-                                 tint_box_size, tint_box_size)
-            legend_ctx.fill()
+            
+            if isinstance(tint, colors.Colormap) or tint == FIT_TO_RGB and tint != (255, 255, 255):
+
+                # convert tint to a color map
+                if isinstance(tint, tuple):
+                    # swap r and b channels
+                    b, g, r = tuple(tint)
+                    tint = r, g, b
```
pieleric, line 1772: I think that the reason you do this is because the internal cairo format is BGRA. So instead, you should convert the `gradient` (numpy array into a BGRA format), using `format_rgba_darray(gradient)`.

amuskens:
Done

src/odemis/gui/comp/combo.py
```diff
@@ -32,7 +32,15 @@
...
 import odemis.gui
 from odemis.gui.comp.buttons import ImageButton, darken_image
+from odemis.gui.conf.data import COLORMAPS
+import wx
 import wx.adv
+from matplotlib import cm
+import matplotlib.colors as colors
+import numpy
+from odemis.util.img import getColorbar, tintToColormap
+import odemis.gui.util.img
```
pieleric, line 42: I don't think this import is needed. => drop.

src/odemis/gui/comp/combo.py
```diff
@@ -32,7 +32,15 @@
...
+from PIL.ImageColor import getcolor
```
pieleric, line 43: Not used? => Drop

src/odemis/gui/comp/combo.py
```diff
@@ -32,7 +32,15 @@
 from odemis.gui import img
 import odemis.gui
 from odemis.gui.comp.buttons import ImageButton, darken_image
+from odemis.gui.conf.data import COLORMAPS
+import wx
 import wx.adv
+from matplotlib import cm
+import matplotlib.colors as colors
+import numpy
```
pieleric, line 40: Unused? => drop.

src/odemis/gui/comp/combo.py
```diff
@@ -189,15 +189,18 @@ class ColorMapComboBox(ComboBox):
...
+        labels: list of str that are the names of color maps
+        choices: list of matplotlib colormap. List of corresponding color map objects
         """
-        self.colormap_dict = kwargs['colormap_dict']
-        self.choices = self.colormap_dict.keys()
-        kwargs['choices'] = self.choices
+        labels = kwargs['labels']
+        colormaps = kwargs['choices']
+        self.choices = labels
+        kwargs['choices'] = labels
```
pieleric, line 198: I don't think that makes sense. The super normally set as client data the "choices". So if you were not changing "choices", it should work out-of-the-box, and you wouldn't need to call `.SetClientData()` here.

src/odemis/gui/comp/combo.py
```diff
@@ -167,3 +175,76 @@ def draw(self, dc):
...
+
+    def SetCustomTintValue(self, value):
+        self.Refresh(eraseBackground=True)
+
+    def OnDrawItem(self, dc, rect, item, flags):
+
+        r = wx.Rect(*rect)  # make a copy
+
+        # Draw a rectangle of the color
+        item_name = self.choices[item]
```
pieleric, line 219: You should be able to write `... = self.Strings[item]`.

src/odemis/gui/comp/combo.py
```diff
@@ -167,3 +175,76 @@ def draw(self, dc):
...
+    """
+
+    def __init__(self, *args, **kwargs):
+        """
+        labels: list of str that are the names of color maps
+        choices: list of matplotlib colormap. List of corresponding color map objects
+        """
+        labels = kwargs['labels']
+        colormaps = kwargs['choices']
+        self.choices = labels
```
pieleric, line 197: I don't think you need this, instead use `.Strings` or `.GetItems()`. Also, it's confusing to put the content of `labels` in `.choices`!

src/odemis/gui/util/img.py
```diff
@@ -1775,13 +1775,7 @@ def draw_legend_multi_streams(images, buffer_size, buffer_scale,
...
             tint = s.tint.value
             
-            if isinstance(tint, colors.Colormap) or tint == FIT_TO_RGB and tint != (255, 255, 255):
-
-                # convert tint to a color map
-                if isinstance(tint, tuple):
-                    # swap r and b channels
-                    b, g, r = tuple(tint)
-                    tint = r, g, b
+            if isinstance(tint, colors.Colormap) or tint == FIT_TO_RGB or tint != RGB_AS_IS:
```
pieleric, line 1778: There is something fishy here.
`... or tint != RGB_AS_IS` means that it'll always go inside, unless `tint == RGB_AS_IS`. So the previous tests are not needed. I doubt this is actually what you meant.

src/odemis/util/img.py
```diff
@@ -39,6 +39,7 @@
 from matplotlib import cm
 
 FIT_TO_RGB = "fitrgb"
+RGB_AS_IS = "rgbasis"
```
pieleric: Please don't create duplicate constants. I'd suggest to import them from `acq.stream`. (an alternative would be to define them here, and import this module in stream. You could even put them in odemis.model.metadata, as TINT_*.

pieleric (changes requested):
It looks way better!
I just have small comments. I'd suggest you fix them, and next week we can have a demo session :-)

pieleric (changes requested):
It now looks really good. I only have minor comments mostly concerning the odemis.util part, which could get a little bit of love.

src/odemis/util/img.py
```diff
@@ -89,6 +92,29 @@
...
+    """
+    Given a tint of a stream, which coudl be an RGB tup,le or colormap object,
+    put it into the format for metadata storage
+    """
+    if isinstance(tint, tuple) or isinstance(tint, list):
+        return tint
+    elif isinstance(tint, colors.Colormap):
+        return tint.name
+    else:
+        return tint
```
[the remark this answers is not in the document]
amuskens: I think this is designed so that the user's selected colormaps will be displayed when they reload the file. Should this not happen?

src/odemis/acq/stream/_projection.py
```diff
@@ -1290,11 +1296,11 @@ def _updateImage(self):
 
             irange = self.stream._getDisplayIRange()  # will update histogram if not yet present
 
-            if not hasattr(self.stream, "fitToRGB") or not self.stream.fitToRGB.value:
+            if self.stream.tint.value != TINT_FIT_TO_RGB:
```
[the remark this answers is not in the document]
amuskens: Do you know which test data file is like this? I tested the files in "good-data" and they seem to work. You can apply fit-to-rgb but it doesn't do anything of course. Maybe I will add an extra check to remove it.

src/odemis/util/img.py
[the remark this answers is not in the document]
pieleric: I meant about the `else` block. It makes the function pass back everything you through at it. I think it's a good place to validate the tint, so I'd extend it this way:
```python
elif tint in (TINT_FIT_TO_RGB, TINT_RGB_AS_IS):
     return tint
else:
    raise ValueError("Unexpected tint %s" % (tint,))
```

(untested code)

src/odemis/util/img.py
```diff
@@ -89,6 +92,29 @@
...
+
+
+def md_format_to_tint(user_tint):
```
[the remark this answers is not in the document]
pieleric: Please, also add info about what happens if you pass the wrong colormap name (eg, "boo"). It's going to call `cm.get_camp("boo")`. What does it return/raise?

src/odemis/util/img.py
```diff
@@ -407,6 +449,40 @@ def DataArray2RGB(data, irange=None, tint=(255, 255, 255)):
...
+    if not alpha:
+        gradient = gradient[:, :, :3]  # discard alpha channel  # convert to rgb
+    gradient = numpy.multiply(gradient, 255)  # convert to rgb
+    return gradient.astype(numpy.uint8)
+
+
+def tintToColormap(tint, name=""):
+    """
+    If a tint is an RGB tuple, or TINT_FIT_TO_RGB string convert it to a matplotlib.colors.Colormap object
+    """
```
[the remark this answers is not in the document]
pieleric: Please, check what happens if you always to `LinearSegmentedColormap.from_list()` `name=""`. If it works fine (I think it does), then let's just entirely drop the `name` argument from your function, and always pass `""`.
If it doesn't, then you need to document what kind of string is allowed.

pieleric (changes requested):
I've followed up on your docstring changes. Also, I think there was an issue with the StreamPanel construction. I've directly put a patch in github... but not tested it. Could you have a look?
BTW to try the temporal only data, see Dropbox/T&D/Software/Test data/good-data/temporal . See also /temporal-spectrum for data which should have fit_to_rgb.

amuskens:
Seems like FIt-toRGB is not working for temporal spectrum data now...

amuskens:
FIT RGB seems to be broken for all the streams now.... I will try fix it.

src/odemis/acq/stream/_static.py
```diff
@@ -337,7 +337,7 @@ def __init__(self, name, raw, *args, **kwargs):
 
         # colouration of the image
         tint = raw.metadata.get(model.MD_USER_TINT, default_tint)
-        self.tint.value = tint
+        self.tint.value = img.md_format_to_tint(tint)
```
pieleric, line 340: Please catch `ValueError` here, and in such case, use default_tint as tint. This ensures that if it's a colormap name, and for some reason matplotlib doesn't know this name, it's still possible to open the file. maybe you could write it this way:
```python
tint = raw.metadata.get(model.MD_USER_TINT, default_tint)
try:
     tint  = img.md_format_to_tint(tint)
except ValueError as ex:
   logging.warning("Failed to find tint: %s", ex)
```

amuskens:
I believe all features are done now!

PR #1431, opened 2020-05-28
CryoSECOM Stage High level control

lazem:


src/odemis/driver/actuator.py
```diff
@@ -2179,3 +2182,516 @@ def terminate(self):
...
+
+    def updateMetadata(self, md):
+        self._metadata.update(md)
+
+    def _updatePosition(self):
+        """
+        update the position VA from the underlying dependency
+        """
+        self._position = self._stage.position.value
+        self.position._set_value(self._position, force_write=True)
```
lazem, line 2276: I found this private use of `self.position._set_value` in other actuator so I kept it as is. This I suppose because the position read only is True? @pieleric

src/odemis/driver/actuator.py
```diff
@@ -2179,3 +2182,516 @@ def terminate(self):
...
+        if self.isInFocus():
+            self.updateMetadata({model.MD_FAV_POS_ACTIVE: self.position.value['z']})
+
+    def isInFocus(self):
+        """
+        A helper function to check if current position is in focus range
+        (the method is public to be used from the parent)
+        :return: True if position in focus range, None otherwise
+        """
+        if self.range[0] <= self.position.value['z'] <= self.range[1]:
```
lazem, line 2564: Is there a case where we need to deal with position being greater than the max range? @pieleric

src/odemis/driver/actuator.py
```diff
@@ -2179,3 +2182,516 @@ def terminate(self):
...
+
+        model.Actuator.__init__(self, name, role, parent=parent, axes=axes_def, **kwargs)
+        self._metadata[model.MD_HW_NAME] = "LinkedHeightFocus"
+        # Get the shared executor from the parent
+        self._executor = parent.executor
+
+        self.position = model.VigilantAttribute({},
+                                                unit="m", readonly=True)
+
+        self.updateMetadata({model.MD_POS_COR: -0.005})
```
lazem, line 2525: I added `MD_POS_COR` in here to use it `updatePosition`, but let me know if there's a better place for it. @pieleric

lazem:
Hi @pieleric I added the code for the `LinkedHeightActuator` wrapper. Would appreciate your feedback on it so far. Meanwhile I will check the yaml file, the lens stage simulator and add few additional test cases.

Thank you!

src/odemis/driver/actuator.py
```diff
@@ -22,19 +22,22 @@
...
+
 import numpy
-import itertools
+from past.builtins import basestring
+
 from odemis import model, util
 from odemis.model import (CancellableThreadPoolExecutor, CancellableFuture,
-                          isasync, MD_PIXEL_SIZE_COR, MD_ROTATION_COR, MD_POS_COR)
-import threading
+                          isasync, MD_PIXEL_SIZE_COR, MD_ROTATION_COR, MD_POS_COR, roattribute)
```
pieleric, line 40: I think you can drop `roattribute` now that executor is not shared with everyone.

src/odemis/driver/actuator.py
```diff
@@ -2179,3 +2182,516 @@ def terminate(self):
...
+        Move the sample stage to the defined position in m for each axis given, and adjust focus if needed. This is an
+        asynchronous method.
+        pos dict(string-> float): name of the axis and new position in m
+        returns (Future): object to control the move request
+        """
+        if not pos:
+            pos = self.position.value
+        self._checkMoveAbs(pos)
+
+        f = self._executor.submit(self._doMoveAbs, pos)
```
pieleric, line 2290: All moves need to be cancellable. When cancelling happens, the current move should stop, and if there were still some (internal) moves planned, they shouldn't be executed.

src/odemis/driver/test/actuator_test.py
```diff
@@ -1274,6 +1275,225 @@ def tearDown(self):
...
+                       'channel': 2,
+                   },
+               },
+               }
+
+if TEST_NOHW:
+    CONFIG_3DOF['locator'] = 'fake'
+
+class TestLinkedHeightActuator(unittest.TestCase):
+    backend_was_running = False
```
pieleric, line 1340: This variable is not needed.

src/odemis/driver/test/actuator_test.py
```diff
@@ -43,7 +44,7 @@
 
 CONFIG_PATH = os.path.dirname(odemis.__file__) + "/../../install/linux/usr/share/odemis/"
 DELPHI_CONFIG = CONFIG_PATH + "sim/delphi-sim.odm.yaml"
-
+# CRYO_CONFIG = CONFIG_PATH + "sim/cryosecom-sim.yaml"
```
pieleric, line 47: Not needed => drop

src/odemis/driver/test/actuator_test.py
```diff
@@ -1274,6 +1275,225 @@ def tearDown(self):
...
+                   },
+                   'rz': {
+                       'range': [-0.785, 0.785],
+                       'unit': 'rad',
+                   },
+               },
+               }
+TEST_NOHW = True
+if TEST_NOHW:
+    CONFIG_5DOF['locator'] = 'fake'
```
pieleric, line 1308: I think that here, for testing the *wrapper*, it's fine to just always have it use the simulator. So you can drop this part.

src/odemis/driver/test/actuator_test.py
```diff
@@ -1274,6 +1275,225 @@ def tearDown(self):
...
+                       'unit': 'm',
+                       'channel': 2,
+                   },
+               },
+               }
+
+if TEST_NOHW:
+    CONFIG_3DOF['locator'] = 'fake'
+
+class TestLinkedHeightActuator(unittest.TestCase):
```
pieleric, line 1339: The test cases need to test more (corner) cases:
 * cancelling a move
 * cancelling a reference
 * reference stage with lens Z is active
 * Try to move Rx with lens Z active
 * Try to set lens Z to active with Rx set
 * Change MD_POS_COR => focus position should immediately be updated
 * Change focus DEACTIVE/ACTIVE => fails
 * Move lens near DEACTIVE, => focus should report it's DEACTIVE
 * Move focus to min/max of range => Check that ACTIVE is updated
 * On the edges of focus range, try to move relative a tiny bit too far => it should fail
 * Move focus to DEACTIVE => check that ACTIVE is not updated

BTW, for checking failures, do you know you can use a construct like `with self.assertRaise(ValueError):`? It's quite handy.

src/odemis/driver/actuator.py
lazem: Thanks for the wholesome amazing review! I made most of the changes but I'm having trouble with the movement cancellation. The movement stops when it's cancelled but the `_was_stopped` attribute always return `False`. Can you please just have a look if I'm indeed on the right way on the cancellation, so I can investigate this issue further.. :)

src/odemis/driver/test/actuator_test.py
lazem: I will remove it after finishing testing with the real hardware..

src/odemis/driver/actuator.py
```diff
@@ -2179,3 +2182,516 @@ def terminate(self):
...
+            return self.ZDirection.NoZMove
+
+    def _checkRxMovement(self, pos, rel=False):
+        """
+        Check if target position contains a movement in Rx axis and compare its value to Rx = 0
+        :param pos: The target position
+        :param rel: whether it's a relative movement or absolute
+        :return: Direction of movement in Rx value
+        """
+        current_rx = self._position['rx']
```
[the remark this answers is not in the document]
lazem: That's right, I added the two CurrentRx and TargetRx if there's a need to distinguish between them in the future. But if there's no need for that, it surely can be simplified as you stated.

src/odemis/driver/actuator.py
lazem: So if I got you right, marking it as out of range would be enough, or it needs more special handling? @pieleric

src/odemis/driver/test/actuator_test.py
```diff
@@ -1274,6 +1276,289 @@ def tearDown(self):
...
+        # Move focus to active position so it can move with the stage
+        f = focus.moveAbs(focus.getMetadata()[model.MD_FAV_POS_ACTIVE])
+        f.result()
+        initial_foc = focus.position.value
+        f = stage.moveAbs(target_pos)
+        time.sleep(3)
+        cancelled = f.cancel()
+        # self.assertTrue(cancelled)
+        # self.assertTrue(f.cancelled())
+        with self.assertRaises(AssertionError):
```
lazem, line 1480: So these commented asserts fail at the moment, while the movements stops and doesn't reach the target position as intended. @pieleric

src/odemis/driver/actuator.py
```diff
@@ -2179,3 +2182,601 @@ def terminate(self):
...
+
+        # If the new position is in focus range, set the MD_FAV_POS_ACTIVE with this new value
+        if self._isInRange():
+            self._metadata[model.MD_FAV_POS_ACTIVE] = {'z': self.position.value['z']}
+            logging.info("%s changed to %s", model.MD_FAV_POS_ACTIVE, self._metadata[model.MD_FAV_POS_ACTIVE])
+
+    def _isInRange(self, pos=None):
+        """
+        A helper function to check if current position is in focus range
+        :return: True if position in focus range, None otherwise
```
pieleric, line 2633: No it must return a *boolean*. So False otherwise.

src/odemis/driver/actuator.py
```diff
@@ -2179,3 +2182,601 @@ def terminate(self):
...
+        """
+        pos = self.position.value['z'] if pos is None else pos
+        # Add 1% margin for hardware slight errors
+        margin = (self._range[1] - self._range[0]) * 0.01
+        if (self._range[0] - margin) <= pos <= (self._range[1] + margin):
+            return True
```
pieleric, line 2639: You can write `return self._range[0] - margin) <= pos <= (self._range[1] + margin)`

src/odemis/driver/test/actuator_test.py
```diff
@@ -1274,6 +1276,289 @@ def tearDown(self):
...
+        f.result()
+        target_pos = {'rx': .350}
+        with self.assertRaises(ValueError):
+            f = stage.moveAbs(target_pos)
+            f.result()
+        f = focus.moveAbs(focus.getMetadata()[model.MD_FAV_POS_DEACTIVE])
+        f.result()
+        f = stage.moveAbs(target_pos)
+        f.result()
+        with self.assertRaises(ValueError):
```
pieleric, line 1500: Before each of the assertRaise, explain why this should fail.

src/odemis/driver/test/actuator_test.py
```diff
@@ -1274,6 +1276,289 @@ def tearDown(self):
...
+        cls.sample_stage = smaract.MC_5DOF(**CONFIG_5DOF)
+        # cls.lens_stage = smaract.MCS2(**CONFIG_3DOF)
+        cls.lens_stage = tmcm.TMCLController(name="test", role="test",
+                                             port="/dev/fake3",
+                                             axes=["x", "y", "z"],
+                                             ustepsize=[5.9e-9, 5.8e-9, 5.8e-9],
+                                             rng=[[-3e-3, 3e-3], [-3e-3, 3e-3], [-3e-3, 3e-3], ],
+                                             refproc="Standard")
+        # cls.lens_stage.updateMetadata({model.MD_FAV_POS_DEACTIVE: -0.006})
+        cls.lens_stage._metadata[model.MD_FAV_POS_DEACTIVE] = {'z': -3e-3}
```
pieleric, line 1353: Please use `updateMetadata()` (it's ok on the lens).

src/odemis/driver/actuator.py
```diff
@@ -2179,3 +2182,601 @@ def terminate(self):
...
+        Call the list of move functions in respect to their ordered execution
+        :param ordered_moves: List of ordered moves functions
+        """
+        for i, ordered_move in enumerate(ordered_moves):
+            try:
+                move_func, *args = ordered_move
+                move_func(*args)
+                if i == 1:
+                    # Re-update the focus position after the second move
+                    self._focus._updatePosition(self._lensz.position.value)
```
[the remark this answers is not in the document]
lazem: Sorry I wasn't clear about this. Basically there are two cases:
1. When the focus is the initial movement, it would need re-updating after the stage is subsequently moved. 
2. The second case is when the stage is moved first. If the focus is updated immediately, and the focus move tried to get the lens position from `_getLensZValue` (which depends on focus position), it returns the same lens position where it is now so it doesn't move. That's why it should wait on updating the focus until it actually moves, so updating the focus anytime the parent stage moves doesn't work. 

I moved it after the loop as you suggested.

src/odemis/driver/actuator.py
```diff
@@ -2179,3 +2182,601 @@ def terminate(self):
...
+        """
+        logging.debug("Cancelling current move")
+        future._must_stop.set()  # tell the thread taking care of the move it's over
+        # Stop both dependencies
+        self._stage.stop()
+        self._lensz.stop()
+        with future._moving_lock:
+            if not future._was_stopped:
+                logging.debug("Cancelling failed")
+            return future._was_stopped
```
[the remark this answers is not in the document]
lazem: I have added the changes as explained here, and tested it with TMCLController for both stages. Something that maybe I didn't get what you meant is, what would be the difference between passing the future to `adjustFocus()` (which subsequently calls the lensz moves functions) and doing it directly from the parent stage?  I think it'd be the same logic moved from the focus to the stage, it'd be just repeated, or I'm wrong on this?

lazem:
Thanks a ton for clarifying the movement cancellation procedure, @pieleric it makes a lot of sense now. I added the requested changes and unittest, plus some minor fixes. Appreciating your feedback.

Thanks!

src/odemis/driver/actuator.py
```diff
@@ -2566,82 +2568,83 @@ def __init__(self, name, role, parent, dependencies, rng, executor, **kwargs):
...
         """
         # Only update position when MD_POS_COR is already configured
         if model.MD_POS_COR not in self._metadata:
-            logging.error("%s not in metadata" % model.MD_POS_COR)
+            logging.error("Focus POS_COR is not found in metadata.")
             return
-        focus_val = self._getFocusValue(target_lensz=lens_pos['z'])
-        logging.debug("Reporting focus position %s", focus_val)
+        if util.almost_equal(lens_pos['z'], self._lensz.getMetadata()[model.MD_FAV_POS_DEACTIVE]['z'], atol=1e-3):
+            # Set focus FAV_POS_DEACTIVE when the lens pos is FAV_POS_DEACTIVE
```
lazem, line 2627: Had to add this special check to set focus `FAV_POS_DEACTIVE` when the lens pos is `FAV_POS_DEACTIVE` (as the reverse is done on when move is initialed) so it's wrong value to get it from `_getFocusValue`

src/odemis/driver/actuator.py
```diff
@@ -2179,3 +2185,631 @@ def terminate(self):
...
+        Move the focus to the defined position for the Z axis. This is an
+        asynchronous method.
+        pos dict(string-> float): name of the axis and new position in m
+        returns (Future): object to control the move request
+        """
+        if not pos:
+            return model.InstantaneousFuture()
+        # check pos value is in range (except when it's = deactive)
+        if not self._isInRange(pos['z']) and not util.almost_equal(pos['z'],
+                                                                   self._metadata[model.MD_FAV_POS_DEACTIVE]['z'],
```
pieleric, line 2656: You have several places checking for a position being (almost) equal to the deactive. I think it'd make it a little clearer if you had a dedicated function, something like `_isDeactivated()` (or `_isParked()` ?). Quite similar to `_isInRange()`.

src/odemis/driver/test/actuator_test.py
```diff
@@ -21,15 +21,23 @@
 """
 
 from __future__ import division
+import sys
+sys.path.insert(0, "/home/cryo-poc/development/odemis/src/odemis/driver/test")
+sys.path.insert(0, "/home/cryo-poc/development/odemis/src")
+# sys.path = [s for s in sys.path if 'packages' not in s]
```
pieleric, line 27: Please drop this before we merge.

src/odemis/util/test.py
```diff
@@ -108,11 +108,11 @@ def stop_backend():
         raise IOError("Backend failed to stop, now %s" % status)
 
 
-def assert_pos_almost_equal(actual, expected, *args, **kwargs):
+def assert_pos_almost_equal(actual, expected, ignore_mismatch=False, *args, **kwargs):
```
pieleric, line 111: It's usually confusing to have parameter/variable names with negation. Let's call this new param something like `match_all=True`. 
Please, also add a docstring about this parameter ("if False, only the expected keys are checked, and actual can have more keys").

src/odemis/driver/actuator.py
```diff
@@ -2179,3 +2182,601 @@ def terminate(self):
...
+
+    def adjustFocus(self, future, vector_value, rel):
+        """
+        Acts as an interface for the parent stage to call when it moves, and focus needs to be adjusted accordingly
+        :param vector_value: name of the axis 'z' and new position value
+        :param rel: whether it's a relative movement or absolute
+        """
+        # Only move focus when it's in active range
+        if not self._isInRange():
+            return
```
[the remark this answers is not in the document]
lazem: Added this check again to prevent movement when lens is parked, I believe it makes sense to drop the movement than throwing exceptions of either:
- Relative movement is out of range (deactive + value)
- Absolute second movement after parent stage rx changed. 

Both of these shouldn't throw exceptions, they just shouldn't happen.

lazem:
Hi @pieleric I added the requested changes, would you please have a look when you have time. 
Thank you!

src/odemis/driver/actuator.py
```diff
@@ -2359,7 +2355,7 @@ def _executeMoves(self, ordered_moves):
         Call the list of move functions in respect to their ordered execution
         :param ordered_moves: List of ordered moves functions and their arguments
         """
-        for i, (move_func, *args) in enumerate(ordered_moves):
+        for move_func, *args in ordered_moves:
```
pieleric, line 2358: This `*args` is an invalid syntax on Python 2. So this prevent loading this module, which would be an issue. I'd suggest to pass the arguments for the function as (sub) tuple, so that you can write here just `move_func, args`.

In general, please check that the test cases pass with Python 2 too. That's not needed for new modules (because we are using Python 3 for any new system installation), but still needed on existing modules.

src/odemis/driver/actuator.py
```diff
@@ -2188,6 +2188,7 @@ def terminate(self):
 
 # Possible movements of the stage Z axis
 MOVE_DOWN, NO_ZMOVE, MOVE_UP = -1, 0, 1
+ATOL_LENS_POS = 100e-6
```
pieleric, line 2191: Please indicate the unit (m) in a comment. Also, you should have a separate value for the rotation, for checking rx/rz. Maybe ATOL_LINEAR_LENS_POS and ATOL_ROTATION_LENS_POS. The second on in rad. 1mrad (1e-3) should probably be already quite small (<0.1°)

src/odemis/driver/actuator.py
```diff
@@ -2405,27 +2401,18 @@ def _checkRxMovement(self, pos, rel=False):
...
-            return True  # Both current and target is 0 => Save to move
-        else:
-            # Only allow a movement in Rx when focus is not in active range
-            if self._focus._isInRange():
-                raise ValueError("Movement in Rx is not allowed while focus is in active position.")
-            else:
-                return False  # Target Rx is not 0 => Do not move focus
+        if util.almost_equal(current_rx, target_rx, atol=ATOL_LENS_POS):
+            return  # Not moving
+        if self._focus._isInRange():
```
pieleric, line 2414: I think it'd be a tiny be safer to use `_isParked()`, and raise a ValueError() whenever it's not officially at the parked position.

src/odemis/driver/actuator.py
```diff
@@ -2779,7 +2778,7 @@ def _checkParentRxRotation(self):
         """
         Throws exception if the parent stage is rotated around the X axis (Rx !=0)
         """
-        if not util.almost_equal(self.parent.position.value['rx'], 0, atol=1e-3):
+        if not util.almost_equal(self.parent.position.value['rx'], 0, atol=ATOL_LENS_POS):
```
pieleric, line 2781: Use ATOL_ROTATION_LENS_POS

src/odemis/driver/actuator.py
```diff
@@ -2179,3 +2185,631 @@ def terminate(self):
...
+
+        self.position = model.VigilantAttribute({}, unit="m", readonly=True)
+        # Position will be updated whenever the underlying lens stage is changed
+        self._lensz.position.subscribe(self._updatePosition, init=True)
+
+        # Check if underlying lens has FAV_POS_DEACTIVE metadata
+        if model.MD_FAV_POS_DEACTIVE not in self._lensz.getMetadata():
+            raise ValueError("The underlying lens stage is missing a FAV_POS_DEACTIVE metadata.")
+        # Check if underlying lens FAV_POS_DEACTIVE is not within its active range
+        lens_deactive = self._lensz.getMetadata()[model.MD_FAV_POS_DEACTIVE]['z']
```
pieleric, line 2565: Preferably, call `.getMetadata()` only once, and reuse the return value. Or use a `try... except KeyError` to do the same.

lazem:
Hi @pieleric Many thanks for your review. I added the suggested changes, can you have a look, and should I squash the commits? 
Thank you!

PR #1433, opened 2020-05-28
eSRM script - bleaching

VictoriaDel:


scripts/sem_bleaching_map.py
```diff
@@ -80,31 +104,36 @@ def __init__(self, dt, roi, pxs, subpx=1, lpower=0.2):
...
+        self.ccd_acq_complete = threading.Event()
+        self._spotsimage = None
+
+        # scale is the distance between the SR pixels in SEM pixels at ebeam.scale=1
+        self.scale = (pxs / self.ebeam.pixelSize.value[0], pxs / self.ebeam.pixelSize.value[1])
+
+        self.power_index = lemissions
         lprng = self.light.power.range
-        if not (lprng[0][self.max_power_index] < lpower < lprng[1][self.max_power_index]):
+        if not (lprng[0][self.power_index] < lpower < lprng[1][self.power_index]):
```
lazem, line 115: Here you are excluding `lpower` from being either 0 or range maximum. Is that intended or these values can be added ( ie replace < with <=  ) ?

scripts/sem_bleaching_map.py
```diff
@@ -311,137 +353,322 @@ def configure_sem_for_tile(self):
...
         # compensate for cosmic rays)
-        
+
+        if lpower is None:
+            lpower = self.lpower
+
+        if exptime is None:
+            exptime = self.exptime
+
+        # Set exposure time and emissions
```
lazem, line 610: Here it's written `#Set exposure time and emissions` but only exposure time is set

scripts/sem_bleaching_map.py
```diff
@@ -1,42 +1,47 @@
...
+--subpx the number of subpixels (must be a square of an integer).
       For each pixel acquired by the CCD, each subpixel is scanned by the ebeam.
---output indicates the name of the file which will contain all the output. It 
+--output indicates the name of the file which will contain all the output. It
          should finish by .h5 (for HDF5) or .tiff (for TIFF).
---lpower defines the excitation light power on the first fluorescence pixel. The value can be 
-		 between 0.004 and 0.4 (in W). Power will be gradually scaled up to the maximum (0.4 W) 
-         on the last pixel to compensate photobleaching effect in the signal drop.
+--lpower defines the excitation light power in Watt
+--lemissions gives the emission source (numbered from 0 to n-1).
```
lazem, line 25: Maybe it'd be clearer if you renamed `lemissions` to be `power_index` or `channel_index` or even `lemission_index `

scripts/sem_bleaching_map.py
```diff
@@ -1,42 +1,47 @@
...
+--exptime gives the exposure time in seconds
+-olp is the excitation light power used for overview image
+-ole is the emission source (numbered from 0 to n-1) used for overview image
+-oet is the exposure time that is used for overview image
```
lazem, line 29: I believe these 3 parameters should be lists not a singular value. Can you make sure of this? They are handled in the code as lists..

scripts/sem_bleaching_map.py
```diff
@@ -311,137 +353,322 @@ def configure_sem_for_tile(self):
...
+            exptime = self.exptime
+
+        # Set exposure time and emissions
+        self.ccd.exposureTime.value = exptime
+
         # Turn on the light
-        # TODO: allow user to specify power (instead of using the maximum)
-        self.light.power.value[self.max_power_index] = self.lpower
-        
+        self.light.power.value[self.power_index] = lpower
```
lazem, line 614: You can add to the comment above the light is turned on by the specified value

scripts/sem_bleaching_map.py
```diff
@@ -311,137 +353,322 @@ def configure_sem_for_tile(self):
...
-        # We use a view to compute the mean to ensure to get a float (and not
-        # a DataArray of one value)
-        # TODO: divide by physical area size, in order to have normalised values
-        # between acquisitions?
-#        pxs = data.metadata[model.MD_PIXEL_SIZE]
-#        area = numpy.prod(data.shape) * numpy.prod(pxs) # m²
-        datar = data[roi[1]:roi[3] + 1, roi[0]:roi[2] + 1]
+        fn_base, fn_ext = os.path.splitext(fn)
+
+        if not self.olpower is None:
```
lazem, line 632: This can be rewritten as:
`if self.olpower:`

scripts/sem_bleaching_map.py
```diff
@@ -465,101 +704,208 @@ def estimate_acq_time(self, shape):
...
+                    for i in numpy.ndindex(subsetshape):
+                        # logging.info("Bleaching pixel (%d,%d) + (%d,%d)*(%d,%d)=(%d,%d)", g[1],g[0],self.gridpitch[1],self.gridpitch[0],i[1], i[0],g[1]+self.gridpitch[1]*i[1],g[0]+self.gridpitch[0]*i[0])
+                        sem_data.append(self.bleach_spot(spotssubset[i].tolist()))
+
+                    self.ccd_acq_complete.wait()
+                    spotsimage_to_save = self._spotsimage[ccd_roi_idx]
+                    self.save_data(spotsimage_to_save, "%s%02d-spots-%05d%s" % (fn_base, ii, n, fn_ext))
+
+                    ccd_data = self.get_fluo_image()
+                    # ccd_data_to_save = ccd_data
```
lazem, line 856: Delete this line

scripts/sem_bleaching_map.py
```diff
@@ -571,37 +917,72 @@ def main(args):
...
+    parser.add_argument("--overviewlightpower", "-olp", dest="olpower", type=float, nargs="+", default=[],
+                        help="excitation light power used for overview image"
+                             "that is acquired before and after the scan.")
+    parser.add_argument("--overviewlightemissions", "-ole", dest="olemissions", type=int, nargs="+", default=[],
+                        help="emissions source (numbered from 0 to n-1) used "
+                             "for overview image that is acquired before and "
+                             "after the scan.")
+    parser.add_argument("--overviewexposuretime", "-oet", dest="oexptime", type=float, nargs="+", default=[],
+                        help="exposure time that is used for overview image"
+                             "that is acquired before and after the scan.")
```
lazem, line 957: As I mentioned above these parameters should be lists not float. Please make this is correct so there'd not be confusion between them and the 3 above.

scripts/sem_bleaching_map.py
```diff
@@ -311,137 +349,315 @@ def configure_sem_for_tile(self):
...
+        sem_pxs = self.ebeam.pixelSize.value
+        ccd_pxs = self.get_ccd_pxs()
+
+        full_width = self.ebeam.resolution.range[1]
+
+        # calculate the width of the border in rel coordinates
+        b_rel = [BORDER * ccd_pxs[0] / (full_width[0] * sem_pxs[0]),
+                 BORDER * ccd_pxs[1] / (full_width[1] * sem_pxs[1])]
+
+        self.ccd.fanSpeed.value = 0
```
pieleric, line 370: Add a comment that this is turning off the cooling and fan to reduce vibrations.

scripts/sem_bleaching_map.py
```diff
@@ -311,137 +349,315 @@ def configure_sem_for_tile(self):
...
+                 full_width[1] * (rel_center[1] - 0.5) - self.drift[1]]
+
+        resolution = [int(full_width[0] * rel_width[0]),
+                      int(full_width[1] * rel_width[1])]
+
+        # configure the ebeam location
+        self.ebeam.translation.value = trans  # px (0 = center)
+        self.ebeam.resolution.value = self.ebeam.resolution.clip(resolution)
+
+    def configure_and_acquire_sem_survey(self, roi):
```
pieleric, line 387: Let's just call this "acquire_sem_survey". It should be a given that the SEM gets configured.

scripts/sem_bleaching_map.py
```diff
@@ -246,33 +292,27 @@ def configure_ccd(self, roi):
...
 
         logging.info("CCD res = %s, binning = %s",
-                      self.ccd.resolution.value,
-                      self.ccd.binning.value)
+                     self.ccd.resolution.value,
+                     self.ccd.binning.value)
+
+        # turn on cooling of the scmos
+        self.ccd.fanSpeed.value = 1
+        self.ccd.targetTemperature = 0
```
pieleric, line 314: This target temperature is only correct for air-cooled Zylas. The proper way would be to store the fanSpeed and targetTemperature at the beginning (or just before turning the fan off), and use it as the "CCD acquisition" values. Also, instead of setting them up here, they should be immediately restored after the survey acquisition (at the end of `acquire_sem_survey()`).

scripts/sem_bleaching_map.py
```diff
@@ -465,101 +693,205 @@ def estimate_acq_time(self, shape):
...
+                        # Time to do drift-correction?
+                        n_drift += 1
+                        if n_drift >= next_dc:
+                            de.acquire()  # take a new
+                            d = de.estimate()
+
+                            self.drift = (self.drift[0] + d[0], self.drift[1] + d[1])
+                            logging.info("Drift estimated to ({:.1f}, {:.1f})".format(*self.drift))
+                            n_drift = 0
+                            next_dc = px_iter.next()
```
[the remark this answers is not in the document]
VictoriaDel: I should still have a condition to check if the drift-correction is performed. Should I replace the condition with if  n > = next_dc, or should I just drop the condition as now next_dc = n + px_iter.next(). I am not sure.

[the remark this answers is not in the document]
pieleric: Sorry for not being very clear. Indeed, you still need to check whether the drift-correction needs to take place, but now, as you suggested you can write `if n > = next_dc:` .

PR #1438, opened 2020-06-05
driver semcomedi: support scanning_ttl high_enabled option

pieleric:
When the TTL has to be inverted for the scanning signal that doesn't
mean it should also be inverted when forced to True/False. It entirely
depends on what the signal means. For example, for an "external" signal,
we want it during for scanning, but still high when set to True.

=> Add another optional value in scanning_ttl

### DIFF
--- a/src/odemis/driver/semcomedi.py
+++ b/src/odemis/driver/semcomedi.py
@@ -2594,13 +2594,16 @@ class Scanner(model.Emitter):
           (lower/upper limit in X) if magnification is 1 (in m)
         park (None or 2-tuple of (0<=float)): voltage of resting position,
           if None, it will default to top-left corner.
-        scanning_ttl (None or dict of int -> (bool or bool, str)): list of
-          digital output ports to indicate the ebeam is scanning or not.
-          If True, it is set to high when scanning, with False, the output is
-          inverted.
+        scanning_ttl (None or dict of int -> (bool or (bool, str), or (bool, str, bool))):
+          List of digital output ports to indicate the ebeam is scanning or not.
+          First boolean is "high_auto": if True, it is set to high when scanning,
+            with False, the output is inverted.
           If a str is provided, a VA with that name will be created, and will
-          allow to force the TTL to enabled (True) or disabled (False), with
-          None (auto) being the default.
+            allow to force the TTL to enabled (True) or disabled (False), in
+            addition to the automatic behaviour (None) which is the default.
+          The second boolean is "high_enabled": if True (default), it will be set
+            to high when VA is True. Otherwise, the TTL will be set to low when
+            the VA is True.
         fastpart (boolean): if True, will park immediately after finishing a scan
           (otherwise, it will wait a few ms to check there if a new scan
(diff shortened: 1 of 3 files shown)


src/odemis/driver/semcomedi.py
```diff
@@ -2594,13 +2594,16 @@ def __init__(self, name, role, parent, channels, limits, settle_time,
...
-          inverted.
+        scanning_ttl (None or dict of int -> (bool or (bool, str), or (bool, str, bool))):
+          List of digital output ports to indicate the ebeam is scanning or not.
+          First boolean is "high_auto": if True, it is set to high when scanning,
+            with False, the output is inverted.
           If a str is provided, a VA with that name will be created, and will
-          allow to force the TTL to enabled (True) or disabled (False), with
-          None (auto) being the default.
+            allow to force the TTL to enabled (True) or disabled (False), with
+            None (auto) being the default.
```
VictoriaDel, line 2603: It would be nice if None is into parentheses, like the other 2 options (True, False).

src/odemis/driver/test/semcomedi_test.py
```diff
@@ -56,7 +56,9 @@
 CONFIG_CNT = {"name": "cnt", "role": "cnt", "source":0}
 CONFIG_SCANNER = {"name": "scanner", "role": "ebeam", "limits": [[-5, 5], [3, -3]],
                   "channels": [0, 1], "settle_time": 10e-6, "hfw_nomag": 10e-3,
-                  "park": [8, 8]}
+                  "park": [8, 8],
+                  "scanning_ttl": {4: True, 2: [True, "external"], 3: [False, "blanker", True]}
+                  }
```
VictoriaDel, line 61: we could add an extra comment here about digital port mapping, or maybe it's too much just for a test file.

src/odemis/driver/test/semcomedi_test.py
```diff
@@ -203,6 +205,12 @@ def compute_expected_duration(self):
         size = self.scanner.resolution.value
         return size[0] * size[1] * dwell + size[1] * settle
 
+    def test_ttl(self):
+        # Just check the VA are created and support changing the values
+        for v in (True, False, None):
+            self.scanner.external.value = v
+            self.scanner.blanker.value = v
```
VictoriaDel, line 212: I am not sure how it is finally tested. If it wasn't a test case, you would raise errors or indicate that VA is not created.

pieleric: Yes, here it's a little hard to check the behaviour, as the test case cannot directly check the behaviour (ie reading the digital output). So it's just a matter of checking the VAs are there, and that they accept 3 values and that when setting it, nothing really bad happens. At least it runs the code.

src/odemis/driver/test/semcomedi_test.py
pieleric: Yes, for the test case, I expect that the reader should be able to read the docstring, so no need to add too much extra comments ;-)

src/odemis/driver/semcomedi.py
pieleric: Updated.

PR #1440, opened 2020-06-08
show pixel value - updated shear calculation for rotated images

VictoriaDel:


src/odemis/acq/stream/_base.py
```diff
@@ -1222,8 +1222,10 @@ def getPixelCoordinates(self, p_pos):
         shear = md.get(model.MD_SHEAR, 0)
         translation = md.get(model.MD_POS, (0, 0))
         size = raw.shape[-1], raw.shape[-2]
-        tform = AffineTransform(rotation=rotation, scale=pxs, shear=shear, translation=translation).inverse()
-        pixel_pos_c = tform(p_pos)
+        tform = AffineTransform(rotation=rotation, scale=pxs, translation=translation)
```
pieleric, line 1225: That needs a comment. Explain that we cannot use the "shear" argument of AffineTransform" because it's a horizontal shear while the MD_SHEAR defines a vertical shear. So it's implemented afterwards by updating the matrix.

PR #1442, opened 2020-06-08
remove choices for the stream

VictoriaDel:


pieleric:
The content is fine. The title of the commit was a little bit confusing, as it sounds like it removes something from Odemis. I've updated the comment to clarify the goal of the change.

PR #1446, opened 2020-06-15
gui cont tabs Sparc2AlignTab: ensure the last autofocus stream has the ccd

pieleric:
Now that we have several focus steams, the order matters.
Otherwise, we have 50% chance of ending with the stream corresponding to
the "other" detector, which also shows the slit, but not in the same way
as the standard image. Therefore the slit line doesn't match perfectly.

=> Always place the stream with the ccd as the first stream.

### DIFF
--- a/src/odemis/gui/cont/tabs.py
+++ b/src/odemis/gui/cont/tabs.py
@@ -4018,6 +4018,14 @@ class Sparc2AlignTab(Tab):
         """
         focus_streams = []
         dets = GetSpectrometerFocusingDetectors(focuser)
+
+        # Sort to have the first stream corresponding to the same detector as the
+        # stream in the alignment view. As this stream will be used for showing
+        # the slit line after the autofocus.
+        if self._ccd_stream:
+            ccd = self._ccd_stream.detector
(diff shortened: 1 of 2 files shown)


src/odemis/gui/cont/tabs.py
```diff
@@ -4018,6 +4018,14 @@ def _createFocusStreams(self, focuser, hw_settings_config):
...
         focus_streams = []
         dets = GetSpectrometerFocusingDetectors(focuser)
+
+        # Sort to have the first stream corresponding to the same detector as the
+        # stream in the alignment view. As this stream will be used for showing
+        # the slit line after the autofocus.
+        if self._ccd_stream:
+            ccd = self._ccd_stream.detector
+            dets = sorted(dets, key=lambda d: d.name == ccd.name, reversed=True)
+
```
lazem, line 4028: Just a small thing here, it's `reverse` not `reversed` :)

pieleric: So strange, I really had tested the code.... Thanks a lot for catching this!
It's now fixed.

PR #1453, opened 2020-06-25
driver: fix ctypes converting strings to wchar*

pieleric:


src/odemis/driver/andorcam2.py
```diff
@@ -867,7 +867,10 @@ def Initialize(self):
 
         logging.debug("Initialising with path %s", self._initpath)
         try:
-            self.atcore.Initialize(self._initpath)
+            path = self._initpath
+            if not isinstance(path, bytes):  # Python 3
```
tepals, line 871: `fsencode` doesn't exist in Python 2, are you sure `not isinstance(path, bytes)` will always return `False` for Python 2?

src/odemis/driver/andorshrk.py
```diff
@@ -631,6 +630,9 @@ def Initialize(self):
         else:
             path = ""
 
+        if not isinstance(path, bytes):  # Python 3
+            path = os.fsencode(path)
```
tepals, line 634: Same comment here: `fsencode` doesn't exist in Python 2, are you sure `not isinstance(path, bytes)` will always return `False` for Python 2?

src/odemis/driver/andorcam2.py
pieleric: I can't think of a way for the path to be unicode. If this happens, the rest will fail anyway. Still, I agree that the code is a little confusing. I've changed it to make it explicitly check for Python 3. The advantage is that once we drop Python 2 support, it's clear that we can also drop the if.

PR #1454, opened 2020-06-26
model vattributes: check the content of the value at init

pieleric:


pieleric:
Added test cases for IntContinuous and FloatContinuous with wrong initial values.

PR #1455, opened 2020-06-30
CryoSECOM overview map acquisition

lazem:
The goal of this procedure is to build a complete view of the TEM grid for both SEM and FM streams, covering the entire sample. Needed tiles are first acquired for each stream, then the complete view is created by stitching the tiles. Most of the acquisition and stitching code is reused from the tile plugin. The objective is to reduce the user input to the minimum. Eventually the function will be called from a single button in the GUI, after the user has configured the streams to be acquired.


src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -0,0 +1,551 @@
...
+            else:
+                if hasattr(self, '_optical_focus'):
+                    try:
+                        current_focus = MeasureOpticalFocus(das[self._streams.index(self._fm_focus_stream)])
+                        # Run the autofocus if current focus got worse than permitted deviation
+                        if abs(current_focus - self._optical_focus) / self._optical_focus > self._focus_fidelity:
+                            self._future.running_subf = AutoFocus(self._fm_focus_stream.detector,
+                                                                  self._fm_focus_stream.emitter,
+                                                                  self._fm_focus_stream.focuser,
+                                                                  good_focus=self._optical_focus)
```
lazem, line 418: Here is missing the small range as mentioned in the spec, just how small it should be?

src/odemis/acq/stitching/test/tiledacq_test.py
```diff
@@ -0,0 +1,230 @@
...
+import odemis.acq.stream as stream
+from odemis import model
+from odemis.acq.stitching._tiledacq import TiledAcquisitionTask, acquireOverviewMap
+from odemis.util import test
+from odemis.util.test import assert_pos_almost_equal
+
+logging.getLogger().setLevel(logging.DEBUG)
+
+CONFIG_PATH = os.path.dirname(odemis.__file__) + "/../../install/linux/usr/share/odemis/"
+SECOM_CONFIG = CONFIG_PATH + "sim/secom-focus-test.odm.yaml"
```
lazem, line 20: Using this for now, though `cryosecom-sim` is using a different fake camera `andorcam2.AndorCam2` than the one tested in here `simcam.Camera`

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -0,0 +1,551 @@
...
+        except Exception as ex:
+            logging.exception("Acquisition failed.")
+            self._future.running_subf.cancel()
+        finally:
+            logging.info("Tiled acquisition ended")
+            self._stage.moveAbs(self._orig_pos)
+        return st_data
+
+
+def acquireOverviewMap(streams, stage, orig_pos=(0, 0), total_area=(0.002, 0.002), overlap=0.2, settings_obs=None,
```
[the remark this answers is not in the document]
pieleric: Also, this function (and the other estimating functions) should be available directly from the acq.stitching module, so you should add an import there.

amuskens:
Is the overview acquisition ready to merge? I am curious to test it with the overview stream GUI

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -0,0 +1,551 @@
...
+    The goal of this task is to acquire a set of tiles then stitch them together
+    """
+
+    def __init__(self, future, streams, stage, orig_pos, total_area, overlap, settings_obs=None, debug_path=None):
+        self._future = future
+        # TODO: order the 'FM' streams for optimal acquisition
+        self._streams = streams
+        # Change the SEM stream horizontalFoV VA to the max if it's found
+        for stream in self._streams:
+            if model.hasVA(stream, "horizontalFoV"):
```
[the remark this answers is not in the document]
lazem: Can you please elaborate on which maximum to limit the fov?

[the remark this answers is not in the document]
pieleric: Use the area size. For example, if the user requests 1mm view, no need to set the FoV to 1.5mm. You can use `horizontalFoV.clip(...)` for that.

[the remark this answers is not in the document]
lazem: Thanks for clarification, done as instructed.

src/odemis/driver/simcam.py
```diff
@@ -42,14 +42,14 @@ class Camera(model.DigitalCamera):
...
+    def __init__(self, name, role, image, dependencies=None, daemon=None, blur_factor=1e4, max_res=None, **kwargs):
+        """
         dependencies (dict string->Component): If "focus" is passed, and it's an
             actuator with a z axis, the image will be blurred based on the
             position, to simulate a focus axis.
-        image (str or None): path to a file to use as fake image (relative to
-         the directory of this class)
-        '''
+        image (str or None): path to a file to use as fake image (relative to the directory of this class)
+        max_res (tuple of (float, float) or None): maximum resolution to clip simulated image, if None whole image shape will be used
```
pieleric, line 51: It'd better be a (int, int), instead of (float, float). Also, please indicate that the simulated image will be a part of the original image based on the MD_POS metadata.

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -0,0 +1,607 @@
...
+
+        return mem_sufficient, mem_est
+
+    STITCH_SPEED = 1e-8  # px/s
+    MOVE_SPEED = 100e-6  # m/s
+
+    def estimateTime(self, remaining=None):
+        """
+        Estimates duration for acquisition and stitching.
+        :returns: (float) estimated required time
```
pieleric, line 366: Please add a docstring about `remaining`. (BTW, what's the type? `int >= 0` ?)

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -0,0 +1,607 @@
...
+from odemis.acq import acqmng
+from odemis.acq import stitching
+from odemis.acq.align.autofocus import MeasureOpticalFocus, AutoFocus, MTD_EXHAUSTIVE
+from odemis.acq.stream import Stream, SEMStream, CameraStream, RepetitionStream, EMStream, ARStream, \
+    SpectrumStream, FluoStream, MultipleDetectorStream, util, executeAsyncTask, \
+    CLStream
+from odemis.util import dataio as udataio
+from odemis.util.comp import compute_scanner_fov, compute_camera_fov
+
+WEAVER_COLLAGE_REVERSE = 2
```
pieleric, line 21: Please import it, don't duplicate.

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -0,0 +1,607 @@
...
+# TODO: Find a value that works fine with cryo-secom
+# Percentage of the allowed difference of tile focus from good focus
```
pieleric, line 23: That's not a "percentage", but a "ratio". I'd also add "The bigger, the less likely the autofocus will be triggered".

install/linux/usr/share/odemis/sim/cryosecom-sim.yaml
```diff
@@ -231,8 +232,9 @@ CryoSECOM: {
     POS_ACTIVE_RANGE: {'x':  [0.00, 0.003], 'y':  [-0.003, -0.001], 'z':  [0.00, 0.003], } , # stage Z – lens Z, when the lens Z is the closest (highest) from the stage
     FAV_POS_ACTIVE: {'rx': 0.00, 'rz': 0.0, 'x':  0.0, 'y':  -0.002, 'z':  0.002},
     FAV_POS_DEACTIVE:  {'rx': 0.00, 'rz': 0.0, 'x':  0.01, 'y':  0.00, 'z':  -0.001},
+    FAV_AREA:  [0, 0, 0.002, -0.002]  # left, top, right, bottom
```
pieleric, line 235: Please change to `[0, -0.002, 0.002, 0]`... or wouldn't it make more sense for the simulator to have something centered around 0? => `[-0.001, -0.001, 0.001, 0.001]` ?

pieleric (changes requested):
Looks good. I've only minor comments.

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -0,0 +1,607 @@
...
+        self._future = future
+        self._streams = streams
+
+        # Get total area as a tuple of width, height from ltrb area points
+        if len(area) != 4:
+            raise ValueError("4 points required (left, top, right, bottom) for total acquisition area.")
+        if not (area[1] > area[3]):
+            raise ValueError("Top point {0} should be larger than bottom point {1}.".format(area[3], area[1]))
+        if not (area[2] > area[0]):
+            raise ValueError("Right point {0} should be larger than left point {1}.".format(area[2], area[0]))
```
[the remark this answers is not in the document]
lazem: Thanks for clarification, I wasn't sure about this.

src/odemis/acq/stitching/_tiledacq.py
lazem: The reason for adding it is that `WEAVER_COLLAGE_REVERSE` is defined in the same `__init__` file that imports the `_tiledacq` functions. What do you recommend in order to avoid circular imports?

https://github.com/delmic/odemis/pull/1455/files#diff-2e25268b4d6583980658f17ba41cc30f

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -48,13 +48,8 @@ def __init__(self, streams, stage, area, overlap, settings_obs=None, log_path=No
...
         # Get total area as a tuple of width, height from ltrb area points
-        if len(area) != 4:
-            raise ValueError("4 points required (left, top, right, bottom) for total acquisition area.")
-        if not (area[1] > area[3]):
-            raise ValueError("Top point {0} should be larger than bottom point {1}.".format(area[3], area[1]))
-        if not (area[2] > area[0]):
-            raise ValueError("Right point {0} should be larger than left point {1}.".format(area[2], area[0]))
-        height = area[1] - area[3]
+        area = util.normalize_rect(area)
+        height = area[3] - area[1]
```
lazem, line 52: I replaced these checks with `util.normalize_rect()` since it already checks for the `len(area)` and reorder the parameters.

src/odemis/acq/stitching/_tiledacq.py
pieleric: I see... yes, this kind of issue creeps in from time to time. Definitely, you must use an import here.

In general, I feel it's a tricky part in Python. You want to have the constants and abstract classes in the root of the module, for everyone to see, so you put them in the `__init__.py` file... but then the internal modules also need them and you end-up with circular deps.

I'd suggest to move the constants to a separate private submodule. You can call it `_constants.py`, or `_interface.py` (but if you have a better name, go for it!)

src/odemis/acq/stitching/_constants.py
```diff
@@ -0,0 +1,19 @@
...
+RIGHT_TO_LEFT = -1
+
+# TODO: Find a value that works fine with cryo-secom
+# Ratio of the allowed difference of tile focus from good focus
+# The bigger, the less likely the autofocus will be triggered
+FOCUS_FIDELITY = 0.3
+# Limit focus range, half the margin will be used on each side of initial focus
+FOCUS_RANGE_MARGIN = 100e-6  # m
+# Indicate the number of tiles to skip during focus adjustment
+SKIP_TILES = 3
```
pieleric, line 19: No, this shouldn't be all the constants. It should be only the constants used by external callers (ie, constants used in the module interface). 
So the constants that are used to tweak the internal module behaviour, or just to make the code more readable shouldn't appear here, but just be in the submodule that they help. (ie, every constant starting from l. 8 should go back).

lazem: Thanks for the clarification, returned these constants to their submodules.

pieleric (approved):
Hi Bassim, it looks good! Would you mind squashing the commits into just *2* commits:
* One commit for the simcam extension (ie, simcam + test cases + yaml file)
* One commit for the stitching extension (ie, all the rest)

lazem:
> Hi Bassim, it looks good! Would you mind squashing the commits into just _2_ commits:
> 
> * One commit for the simcam extension (ie, simcam + test cases + yaml file)
> * One commit for the stitching extension (ie, all the rest)

Hi @pieleric Done as requested.

PR #1456, opened 2020-06-30
Improve appearance spectral export

VictoriaDel:


src/odemis/gui/util/img.py
```diff
@@ -1226,8 +1228,8 @@ def spectrum_to_export_data(proj, raw, vp=None):
         text_colour = (0, 0, 0)
 
         # Draw bottom horizontal scale legend
-        tick_spacing = client_size[0] // 4
-        font_size = client_size[0] * SPEC_FONT_SIZE
+        tick_spacing = client_size[0] // 6
+        font_size = client_size[0] * 0.035
```
pieleric, line 1232: Please use a constant. Actually, I'd recommend updating this constant to 0.035. And if it's too big for the other usage (like in AR display, a different constant should be used there, eg AR_FONT_SIZE)

src/odemis/gui/util/img.py
```diff
@@ -1208,6 +1210,8 @@ def spectrum_to_export_data(proj, raw, vp=None):
             # Limit to the displayed ranges
             range_x = vp.hrange.value
             range_y = vp.vrange.value
+            if min(range_y) < 10:
+                range_y = (0, max(range_y))
```
pieleric, line 1214: What is this for? range is either in m or pixels. Why 10 m? Why 10 px? 
Typically, when the range is in px, half of the range is negative (eg, -256 -> 256). So cutting half of it doesn't make sense.
This line is really fishy, please explain.

src/odemis/gui/util/img.py
```diff
@@ -1205,9 +1207,19 @@ def spectrum_to_export_data(proj, raw, vp=None):
...
             range_y = vp.vrange.value
+            # extend the range_y by 5% from both sides (range extended by 10% in total)
+            ext = (range_y[1] - range_y[0]) * 0.05
+            # Check if the range includes 0, after the extension, and compare 0 with the initial minimum and maximum
+            # values to replace them if needed.
+            if range_y[0] - ext <= 0 <= range_y[1] + ext:
+                if 0 < min(range_y):
+                    range_y = (0, max(range_y))
+                elif 0 > max(range_y):
+                    range_y = (min(range_y), 0)
```
pieleric, line 1222: That's almost good. A corner-case that this doesn't handle is the "middle-case": if 0 is actually inside the range already (eg, values from -100 to 500). In such case, it would actually reduce the range. I think you could rewrite it this way:
```python
if range_y[0] - ext <= 0 <= range_y[1] + ext:
    range_y_and_0 = sorted([range_y[0], 0, range_y[1]])
    range_y = range_y_and_0[0], range_y_and_0[-1]

```

src/odemis/gui/util/img.py
```diff
@@ -1205,9 +1207,19 @@ def spectrum_to_export_data(proj, raw, vp=None):
...
 
         if vp is not None:
+
             # Limit to the displayed ranges
             range_x = vp.hrange.value
             range_y = vp.vrange.value
+            # extend the range_y by 5% from both sides (range extended by 10% in total)
+            ext = (range_y[1] - range_y[0]) * 0.05
+            # Check if the range includes 0, after the extension, and compare 0 with the initial minimum and maximum
+            # values to replace them if needed.
```
pieleric, line 1217: I think this could be a single comment, as it's all about the same thing. Something like:
"Check if the range is near 0, and if so, extend it slightly to include it.
We define "near" as 5% of the range."

BTW, I think that 10% would be very acceptable (eg, if values between 1 and 9 => extending to 0 and 9 would be acceptable).

src/odemis/gui/util/img.py
```diff
@@ -1205,9 +1207,19 @@ def spectrum_to_export_data(proj, raw, vp=None):
         ctx = cairo.Context(surface)
 
         if vp is not None:
+
```
pieleric, line 1210: No need for an empty line here.

PR #1460, opened 2020-07-09
Bug fixed: Long acquisitions (using integration time) use much memory

VictoriaDel:


src/odemis/acq/stream/_sync.py
```diff
@@ -1402,36 +1415,57 @@ def _integrateImages(self, n):
...
             if MD_EXP_TIME in md:
                 md[model.MD_EXP_TIME] *= n
             md[model.MD_INTEGRATION_COUNT] = n
 
+    def _subtractBaseline(self, n):
+        """
+       Substract the baseline, if exists, from the final integrated image which is saved in self._acq_data
+       Args:
+           n: (int) Number of images that need to be integrated (summed).
+        """
```
pieleric, line 1441: Move this function to a dedicated function in `odemis.util.img`. It should look like: `subtract_baseline(img, n)`, and return a DataArray. Indicate in the docstring that the original data is modified (but not the metadata).

src/odemis/acq/stream/_sync.py
```diff
@@ -1402,36 +1415,57 @@ def _integrateImages(self, n):
...
-                    md[model.MD_BASELINE] = baseline_sum - extra_bl
+            # replace images, which needed to be averaged/integrated, with averaged/integrated image
+            self._acq_data[stream_idx] = [model.DataArray(data, md)]
 
+    def _updateMetadata(self, n):
+        """
+        Update the metadata for the final integrated image which is saved in self._acq_data
+        Args:
+            n: (int) Number of images that need to be integrated (summed).
+        """
```
pieleric, line 1426: Move it to `odemis.util.img`, with a more specific name, like `update_integration_metadata(md, n)`.

src/odemis/acq/stream/_sync.py
```diff
@@ -1371,22 +1378,28 @@ def _runAcquisitionEbeam(self, future):
...
-        integrated image. If values overflow, clip the image. If number of images = 1,
-        no image integration necessary.
-        :param n: (int) Number of images that need to be integrated (summed).
+        Integrate/sum the images in self._acq_data. If number of images = 1, no image integration necessary.
+        If number of images > 10, integrate one image after another and replace by the integrated image.
+        It's about integrating the new acquired image in self._acq_data with the last integrated image if exists.
+        If number of images <= 10, integrate/sum the last n acquired images in self._acq_data and replace by the
+        integrated image. If values overflow, clip the image.
+        Args:
+            n: (int) Number of images that need to be integrated (summed).
```
pieleric, line 1387: I think it should be simplified and moved to `odemis.util.img` as `sum_img(imgs, init=None)`. imgs is a list of images. init is a single image. It returns a single image. If there is no init, it computes the right dtype to ensure it cannot overflow. If init is present, it uses it to store the result, and so its dtype should fit, and it will be also the return value.

In the stream, here you "only" have to call this function correctly. When receiving the first image, compute the right dtype with `get_best_dtype_for_acc(orig_dtype, n)`, and convert the image. Then, for every other image, call `sum_img()`.

src/odemis/acq/stream/_sync.py
```diff
@@ -1371,22 +1378,28 @@ def _runAcquisitionEbeam(self, future):
...
         for stream_idx, das in enumerate(self._acq_data):
             # get the metadata from last image (should be the same for all images)
             md = das[-1].metadata
             det_type = md.get(model.MD_DET_TYPE, model.MD_DT_INTEGRATING)
             # get the images, which need to be averaged/integrated
             data2process = das[-n:]
-            orig_dtype = data2process[0].dtype
+            orig_dtype = data2process[-1].dtype if n == 2 else data2process[0].dtype
 
             if det_type == model.MD_DT_NORMAL:  # SEM
```
pieleric, line 1404: Average is not going to work as expected if you do it one image after one image (because the latest images would have a bigger weight than the first ones). So you need to always sum, and only after the end of the acquisition, if the detector is DT_NORMAL, then divide and restore the original dtype.

src/odemis/acq/stream/_sync.py
```diff
@@ -1301,13 +1301,20 @@ def _runAcquisitionEbeam(self, future):
...
                             logging.exception("Failed to update CCD live view")
 
                         n += 1  # number of images acquired so far
 
-                    # integrate/sum images
-                    self._integrateImages(integration_count)
+                    if integration_count <= 10:
+                        # integrate/sum images
+                        logging.debug("Integrate the %s acquired images", n)
+                        self._integrateImages(integration_count)
```
pieleric, line 1315: Don't do anything special for integ count < 10. You're not really saving time, and the code is twice more complex. In addition, to test the difficult code path, you now have to have at least 10 images.
The only special case "allowed" is integration_count == 1.

src/odemis/acq/stream/_sync.py
```diff
@@ -1301,13 +1301,20 @@ def _runAcquisitionEbeam(self, future):
                         # Live update the setting stream with the new data
                         try:
                             self._sccd._onNewData(self._ccd_df, self._acq_data[self._ccd_idx][-1])
+                            if integration_count > 10:
+                                logging.debug("Sum the acquired images one after another")
+                                self._integrateImages(integration_count)
```
pieleric, line 1306: That shouldn't be inside this block. It's reserved for `_onNewData()` which is "allowed" to fail because it's just an extra feature.

Also drop this special case for 10. Always do the code path.

src/odemis/acq/stream/_sync.py
Kleijwegt: @VictoriaDel I think you could do the division part in the method assemble_final_data if you saved to sum in live_data. (Though I haven't read the specifications for this PR/do not know all the details). In live_data you want to also save the latests SEM image to be displayed on screen.

src/odemis/acq/stream/_sync.py
VictoriaDel: I did it in the _sync.py because I wanted to replace the images saved in self._acq_data so as the self._acq_data is not getting too big.

src/odemis/util/img.py
```diff
@@ -811,3 +813,107 @@ def getBoundingBox(content):
 
     return rect
 
+
+def integrateImages(imgs, n, integration_step, best_dtype=None):
```
pieleric, line 817: Honestly, this function usage is terrible. I'm sorry suggesting you to do something like that. You pass a lot of parameters, some of them are used only when other parameters are as special value. It also constantly returns a dtype, but you have to store it anyway. I think that we can do much better for the use cases by creating a class. The main difference is that the `best_dtype` and `integration_step` don't need to be stored apart. It'd look like this.
``` python
class ImageIntegrator(...):
    def __init__(steps):
         self.steps = steps  # can be changed by the caller, on the fly
         self._step = 0
         self._img = None
         self._best_dtype = None

    def append(img):
        """
        Integrate the image with the previous images and return the complete image.
        It will automatically reset the integrated image after reaching the number of integration count.
        """
       self._step += 1
        # if no image yet, do fist image code
        # else, add to _img
        # if reached the number of steps, do final integration, and reset all the values

      return img
```

This way, it should be possible to only store a "ImageIntegrator" in the two callers above, and nothing else really.

src/odemis/acq/stream/_sync.py
```diff
@@ -1303,12 +1309,17 @@ def _runAcquisitionEbeam(self, future):
...
                         except Exception:
                             logging.exception("Failed to update CCD live view")
+                        # check if we need to integrate images (integration count > 1)
+                        if integration_count > 1:
+                            # integrate the acquired images one after another during the acquisition
+                            self.integration_step += 1
+                            for stream_idx, das in enumerate(self._acq_data):
+                                self._acq_data[stream_idx], self.best_dtype = \
+                                    img.integrateImages(das, integration_count, self.integration_step,
+                                                        best_dtype=self.best_dtype)
```
pieleric, line 1319: You cannot have a single best_dtype for all the streams, it's different for each of them (in theory). Same thing for the image integrator. I think you need to init with `self._img_intor = [None for s in streams]`, and in this loop here, create an image integrator if there is none yet. After the acquisition, it must be reset to `None` (both to save memory, and to make sure it still works the same on the next acquisition).

src/odemis/acq/stream/_helper.py
```diff
@@ -168,13 +172,24 @@ def __init__(self, name, detector, dataflow, emitter, scanner=None, sstage=None,
...
         :param dataflow: (model.Dataflow) The dataflow.
         :param data: (model.DataArray). The new image that has arrived from the detector.
         """
-        if hasattr(self, "integrationTime"):
-            self.raw = ([data] + self.raw)[:self.integrationCounts.value]
+        if hasattr(self, "integrationTime") and self.integrationCounts.value > 1:
+            self.integration_step += 1
+            # Once the value of integration counts is reached, prepare the raw data to store the
+            # first acquired image for the next ebeam position and update the integration step.
+            if self.integration_step > self.integrationCounts.value:
```
pieleric, line 185: You need to handle the fact that `integrationCounts` can change while playing. The simplest way to handle it is to reset everything if the count is different from the image integrator steps. The more fancy way is that when increasing the integration count (compared to before), then the previous accumulated images are kept. However, as it's quite a bit of a corner case, I think it's fine to leave this fancy way as a TODO.

src/odemis/util/img.py
```diff
@@ -811,3 +813,107 @@ def getBoundingBox(content):
...
+        orig_dtype = imgs[-1].dtype
+        best_dtype = get_best_dtype_for_acc(orig_dtype, n)
+        return imgs, best_dtype
+
+    # get the metadata from the last image (should be the same for all images)
+    md = imgs[-1].metadata
+    logging.debug("Integrate one image after another")
+    data = numpy.sum(imgs[:2], axis=0, dtype=best_dtype)
+    # update the metadata of the integrated image in every integration step
+    md = update_integration_metadada(md, integration_step)
```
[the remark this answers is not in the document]
VictoriaDel: I multiply by ._step and not by steps (which is the total number of integration counts) so I thought multiplication is correct and it's good to update the metadata after every integration.

src/odemis/util/img.py
```diff
@@ -811,3 +813,105 @@ def getBoundingBox(content):
...
+            logging.debug("Integrate one image after another")
+            data = numpy.add(img[0], img[1], dtype=self._best_dtype)
+            # update the metadata of the integrated image in every integration step
+            md = self.update_integration_metadada(md)
+            self._img = [model.DataArray(data, md)]
+
+            # At the end of the acquisition, check if the detector type is DT_NORMAL and then take the average by
+            # dividing with the number of acquired images (integration count) for every pixel position and restoring
+            # the original dtype.
+            if self._step == self.steps:
```
pieleric, line 859: Make it `>=`, so that if the user changed steps to a smaller number, it doesn't go forever.

src/odemis/util/img.py
```diff
@@ -811,3 +813,105 @@ def getBoundingBox(content):
...
+        """
+        baseline = model.MD_BASELINE
+        # Subtract the baseline (aka black level) from the final integrated image.
+        # Remove the baseline from n-1 images, keep one baseline as bg level.
+        minv = float(data.min())
+
+        # If the baseline is too high compared to the actual black, we could end up subtracting too much,
+        # and values would underflow => be extra careful and never subtract more than the min value.
+        baseline_sum = self.steps * baseline  # sum of baselines for n images.
+        extra_bl = (self.steps - 1) * baseline  # sum of baselines for n-1 images, keep one baseline as bg level.
```
pieleric, line 905: Use self._step, which is the actual number of integrations. In general it should be the same value, but if .steps is suddenly decreased, ._steps will still contain the right value.

src/odemis/util/img.py
```diff
@@ -811,3 +813,105 @@ def getBoundingBox(content):
...
+            self._img = [model.DataArray(data, md)]
+
+            # At the end of the acquisition, check if the detector type is DT_NORMAL and then take the average by
+            # dividing with the number of acquired images (integration count) for every pixel position and restoring
+            # the original dtype.
+            if self._step == self.steps:
+                det_type = md.get(model.MD_DET_TYPE, model.MD_DT_INTEGRATING)
+                if det_type == model.MD_DT_NORMAL:  # SEM
+                    orig_dtype = img[-1].dtype
+                    data = numpy.divide(data, self.steps, dtype=orig_dtype)
```
pieleric, line 863: This will not work as orig_dtype hasn't been stored.
Please write a test case for this code path.

src/odemis/util/img.py
```diff
@@ -811,3 +813,105 @@ def getBoundingBox(content):
...
+
+        if self._img is None:
+            orig_dtype = img[-1].dtype
+            self._best_dtype = get_best_dtype_for_acc(orig_dtype, self.steps)
+            self._img = img
+            return img
+        else:
+            # get the metadata from the last acquired image (should be the same for all images)
+            md = img[-1].metadata
+            logging.debug("Integrate one image after another")
```
pieleric, line 850: That's a lot of debug messages, as this will happen for every image received. I'd recommend putting a message only at the end of the integration.

src/odemis/util/img.py
pieleric, line 859: Move it outside of the if, so that if .steps == 1, it works too.

src/odemis/util/img.py
```diff
@@ -811,3 +813,105 @@ def getBoundingBox(content):
...
+        Returns:
+            model.DataArray: the integrated image with the updated metadata
+        """
+        self._step += 1
+
+        if self._img is None:
+            orig_dtype = img[-1].dtype
+            self._best_dtype = get_best_dtype_for_acc(orig_dtype, self.steps)
+            self._img = img
+            return img
```
pieleric, line 846: Don't return inside the `if` block, but outside. It should be possible to return the same variable in both cases of the if (name it something like `integ_img`.

src/odemis/util/img.py
```diff
@@ -811,3 +813,105 @@ def getBoundingBox(content):
...
+        else:
+            # get the metadata from the last acquired image (should be the same for all images)
+            md = img[-1].metadata
```
pieleric, line 849: `img` is a single DataArray, so drop `[-1]`. This works here because the DataArray also supports indexing and returns a smaller DataArray.

src/odemis/util/img.py
```diff
@@ -811,3 +813,105 @@ def getBoundingBox(content):
...
+
+    def update_integration_metadada(self, md):
+        """
+        Update the metadata of each integrated image.
+        """
+        if MD_DWELL_TIME in md:
+            md[model.MD_DWELL_TIME] *= self._step  # total time ebeam stayed on same pixel/position
+        if MD_EXP_TIME in md:
+            md[model.MD_EXP_TIME] *= self._step
+        md[model.MD_INTEGRATION_COUNT] = self._step
```
pieleric, line 884: Yes, this works, because you pass the latest metadata, and normally all the images are acquired with the same settings. However I feel it'd be easier to understand and it would work in more cases if you were doing the addition of two metadata. I'd recommend something like that:
``` python
def add_integration_metadata(mda, mdb):
   """
   add mdb to mda, and update mda with the result
   returns dict: mda, which has been updated
   """
   if MD_DWELL_TIME in mda:
       mda[model.MD_DWELL_TIME] += mdb.get(model.MD_DWELL_TIME, 0)
   :
   mda[model.MD_INTEGRATION_COUNT] = mda.get(model.MD_INTEGRATION_COUNT, 1) + mdb.get(model.MD_INTEGRATION_COUNT,1)

   return mda
```

src/odemis/acq/stream/_sync.py
```diff
@@ -1303,19 +1307,23 @@ def _runAcquisitionEbeam(self, future):
                             self._sccd._onNewData(self._ccd_df, self._acq_data[self._ccd_idx][-1])
                         except Exception:
                             logging.exception("Failed to update CCD live view")
+                        # check if we need to integrate images (integration count > 1)
+                        if integration_count > 1:
+                            for stream_idx, das in enumerate(self._acq_data):
+                                if self._img_intor[stream_idx] is None:
+                                    self._img_intor[stream_idx] = img.ImageIntegrator(integration_count)
+                                self._acq_data[stream_idx] = self._img_intor[stream_idx].append(das)
```
pieleric, line 1315: Just pass the "last" and only DataArray in das => `self._acq_data[stream_idx][-1] = ... .append(das[-1])`.

src/odemis/acq/stream/_helper.py
```diff
@@ -168,13 +170,38 @@ def __init__(self, name, detector, dataflow, emitter, scanner=None, sstage=None,
...
+                self._img_intor = img.ImageIntegrator(self.integrationCounts.value)
+            # Catch the exception when the user changes some of the hardware settings (eg, resolution), while playing
+            # the stream. To do so, restart everything and keep the last image received.
+            try:
+                self.raw = self._img_intor.append(self.raw)
+            except Exception:
+                logging.warning("Failure when updating the hardware settings")
+                self._img_intor = None
+                self.raw = [data]
+                self._integration_step = 1
```
pieleric, line 204: The warning message is not proper. I'd suggest something like `"Failed to integrate image: %s", ex`.
In this case, immediately replace the integrator, and set it with the first image:
```python
self._img_intor = img.ImageIntegrator(...)
self._img_intor.append(data)
```
No need to have a `._integration_step`

pieleric (changes requested):
I think that with the class it now starts to be nice to look at. The API can be slightly clarified.
Also, now you can add test cases... with all sorts of different corner cases (eg, steps == 1, DET_TYPE=DET_INTEGRATING, .steps is reduced during the integration...)

src/odemis/util/img.py
```diff
@@ -811,3 +813,114 @@ def getBoundingBox(content):
...
+    def subtract_baseline(self, data, md):
+        """
+        Subtract accumulated baselines from the data so that the data only has one.
+        Args:
+            data: the data after the integration of all images
+            md: metadata of the integrated image
+        Returns:
+            data, md: the updated data and metadata after the subtraction of the baseline
+        """
+        baseline = model.MD_BASELINE
```
pieleric, line 906: Something went wrong here! `baseline = md[model.MD_BASELINE]`

Please add a BASELINE in one of the test cases. You can use like `data.metadata[model.MD_BASELINE] = 10`.

src/odemis/util/test/img_test.py
```diff
@@ -837,6 +837,105 @@ def test_25d(self):
         self.assertEqual(255, out[128, 256, 3])
 
 
+class TestImageIntegrator(unittest.TestCase):
+
+    def setUp(self):
+        dtype = numpy.dtype("uint16")
```
pieleric, line 843: Please just write `numpy.uint16`.

src/odemis/util/test/img_test.py
```diff
@@ -837,6 +837,105 @@ def test_25d(self):
...
+        self.integrated_data = None
+        self.img_intor = None
+        self.integrationCounts = 3
+
+    def test_simple(self):
+        self.img_intor = img.ImageIntegrator(self.integrationCounts)
+        self.assertEqual(self.img_intor.steps, self.integrationCounts)
+
+        for i in range(self.integrationCounts):
+            self.integrated_data = self.img_intor.append(self.data[i])
```
pieleric, line 864: I'd recommend to test passing data of 2D. (data[i] is only one dimension). As data contains only 1's, I'd suggest to just use `data`, and it's already 2D.

pieleric (changes requested):
It looks good, but there are a couple of minor things to adjust.

PR #1463, opened 2020-07-10
driver static/spectrapro: fix initial VA value not fitting

pieleric:
Commit e608787e8 (model vattributes: check the content of the value at
init) "broke" these places which were incorrect.

### DIFF
--- a/src/odemis/driver/spectrapro.py
+++ b/src/odemis/driver/spectrapro.py
@@ -199,8 +199,8 @@ class SpectraPro(model.Actuator):
             self._turret = self.GetTurret()
 
         # for now, it's fixed (and it's unlikely to be useful to allow less than the max)
-        max_speed = 1000e-9 / 10 # about 1000 nm takes 10s => max speed in m/s
-        self.speed = model.MultiSpeedVA(max_speed, range=[max_speed, max_speed], unit="m/s",
+        wl_speed = 1000e-9 / 10  # about 1000 nm takes 10s => max speed in m/s
+        self.speed = model.MultiSpeedVA({"wavelength": wl_speed}, range=[wl_speed, wl_speed], unit="m/s",
                                         readonly=True)
 
         gchoices = self.GetGratingChoices()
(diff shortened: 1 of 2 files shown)


src/odemis/driver/spectrapro.py
```diff
@@ -200,7 +200,7 @@ def __init__(self, name, role, port, turret=None, calib=None,
 
         # for now, it's fixed (and it's unlikely to be useful to allow less than the max)
         max_speed = 1000e-9 / 10 # about 1000 nm takes 10s => max speed in m/s
-        self.speed = model.MultiSpeedVA(max_speed, range=[max_speed, max_speed], unit="m/s",
+        self.speed = model.MultiSpeedVA({"wavelength": max_speed}, range=[max_speed, max_speed], unit="m/s",
```
Kleijwegt, line 203: I think the range is set incorrectly the min and maximum is the same now.
I would either expect "range=[0, max_speed]' or something a like

src/odemis/driver/static.py
```diff
@@ -103,9 +104,10 @@ def __init__(self, name, role, mag, mag_choices=None, na=0.95, ri=1,
             if (not isinstance(pole_pos, collections.Iterable) or
                 len(pole_pos) != 2 or any(not 0 < v < 1e6 for v in pole_pos)):
                 raise ValueError("pole_pos must be 2 positive values, got %s" % pole_pos)
-            self.polePosition = model.ResolutionVA(tuple(pole_pos),
-                                                   rng=((0, 0), (1e6, 1e6)),
-                                                   unit="px")
+            self.polePosition = model.TupleContinuous(tuple(pole_pos),
```
Kleijwegt, line 107: Do we want to add the range to the docstring? Since this input is directly used from the __ init__ maybe it is usefull information when looking at the signature of the constructor. Or is this something we choose to not add to the docstring?

src/odemis/driver/spectrapro.py
pieleric: Actually it's read-only, so it's not possible to be changed. That's why all the values are the same. That said, I think the name is weird. It should just be "speed".

src/odemis/driver/static.py
pieleric: Which docstring are you talking about? The one of OpticalLens? In the docstring, we've already indicated that it should be > 0, so only the max is needed. (actually, it's >=0, I'll fix that). However, here the reason I didn't specify the max is because I picked "1e6" arbitrarily as just a big number which no CCD should have more pixels than that.

pieleric:
Updated, as per the comments.

PR #1464, opened 2020-07-13
doc dataio: add formula to convert from pixel coordinates to physic coordinates

pieleric:
Now that we've written code that does the conversion, and realise how
unclear it was... so let's document it.

### DIFF
--- a/doc/develop/dataio.rst
+++ b/doc/develop/dataio.rst
@@ -79,7 +79,8 @@ provides four functions:
     :returns: an opened file
     :rtype: AcquisitionData
 
-.. TODO: describe the helper functions of dataio
+.. TODO: describe the helper functions of dataio and util.dataio
+.. TODO: describe AcquisitionData class
 
 Example usage
 -------------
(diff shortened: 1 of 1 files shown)


doc/develop/dataio.rst
```diff
@@ -104,6 +105,48 @@ To read data from an OME-TIFF file, one could write:
    print(das[0].metadata[model.MD_LENS_MAG])
    print(das[0].metadata)
 
+Image position metadata
+=======================
+Most of the data store correspond to "spatial" data, i.e., an image representing
```
tepals, line 110: typo: corresponds

doc/develop/dataio.rst
```diff
@@ -104,6 +105,48 @@ To read data from an OME-TIFF file, one could write:
...
+the with axes in X and Y. In order to ensure a perfect overlay between data from
```
tepals, line 111: I think there's a word missing after "the". What is the image representing?

doc/develop/dataio.rst
```diff
@@ -104,6 +105,48 @@ To read data from an OME-TIFF file, one could write:
...
+different acquisition type, the metadata describes precisely how the image should
```
tepals, line 112: typo: types

doc/develop/dataio.rst
```diff
@@ -104,6 +105,48 @@ To read data from an OME-TIFF file, one could write:
...
+
+Note that during acquisition, all these metadata have a twin-brother named with an extra
+`_COR` in order to record the correction on the image display. The function ``util.img.mergeMetadata()``
+can be used to merge these corrections into the main metadata. Before saving data
+to a file, the correction is automatically merged. So typically, after opening a
+file, the data will not have any of these extra correction metadata.
+
+For a given pixel situated at coordinates *p = i, j* in an image, its "physical" position
+*P = x, y* (in meters) can be computed with the following formula *P = RSLp + T*,
+where R is the rotation, L is the scale, and S is the shear.
```
tepals, line 130: You didn't mention "and T is the translation" at the end. Translation is also nowhere mentioned in this description. Is there a MD_TRANSLATION?

tepals, line 130: According to the docstring and code of the  `AffineTransform` class S is scale and L is shear. However I think the order should be `P = Rotation*Shear*Scale*p+Translation`. There might be a bug in the code.

doc/develop/dataio.rst
```diff
@@ -104,6 +105,48 @@ To read data from an OME-TIFF file, one could write:
...
+   from odemis import model
+   from odemis.util.transform import AffineTransform
+   pxs = md.get[model.MD_PIXEL_SIZE]
+   translation = md.get(model.MD_POS, (0, 0))
+   rotation = md.get(model.MD_ROTATION, 0)
+   shear = md.get(model.MD_SHEAR, 0)
+
+   tform = AffineTransform(rotation=rotation, scale=pxs, translation=translation)
+   L = numpy.array([(1, 0), (-shear, 1)])
+   tform.transformation_matrix = numpy.dot(tform.transformation_matrix, L)
```
tepals, line 145: why add the negative shear, and not just add the shear as an argument to the class?

pieleric: That's what took a full day of debugging when we implemented it in Odemis! The AffineTransform implements an horizontal shear, however, in Odemis (overlay calibration), it turns out that we had a vertical shear (and I think that the reason is that we had just picked something random, although for the typical case, it would have made sense to use a horizontal shear).
The simplest way we've found (with Victoria) to work-around this discrepancy is this separate shear afterwards.

doc/develop/dataio.rst
```diff
@@ -104,6 +105,48 @@ To read data from an OME-TIFF file, one could write:
...
+
+   P = tform(p)
+   p = tform.inverse()(P)
```
tepals, line 148: I decided to try this code, and I thought small p would stay equal. However it doesn't:

```
tform = AffineTransform(rotation=1.2, scale=2, translation=2)
P = tform(((1, 1), (2, 2)))
p = tform.inverse()(P)
p
 
 ([[1.93203909, -0.50172042],
 [4.22643593,  1.06796091]])
```

I expected it to return ((1, 1), (2, 2)).

pieleric: Translation should be 2 floats, not a single value. If you use (2,2), or (2,3), that code works.
That said, I've found extra issues in the example code, and I'll update it.

Also, I don't understand what should mean translation = 2, but I would expect that either it does the same as `(2,2)`, as the "worst acceptable behaviour", and not cause the inversion to break. Anyway, I think it should just raise a ValueError.

tepals: Indeed, I meant to do `translation = (2, 2)`. And I would also expect `translation=2` to do the same as `(2, 2)` or raise an error. 

The same could actually be said for scale. The docstring says scale is the scale in x and in y. However it doesn't raise an error when you pass a float, it will then use it as `(2, 2)`.

pieleric: I'll add the type checks in that code.

doc/develop/dataio.rst
pieleric: Well, the code is `numpy.dot(numpy.dot(R, S), L)`, so it's "RSL". Why do you think this would be wrong? It's all about convention anyway. In practice, we've tried a lot of images on the display, and it seems to always match perfectly, using this formula.

doc/develop/dataio.rst
pieleric: I'll send a separate PR for checking the arguments of the transformations.

doc/develop/dataio.rst
tepals: It is indeed all about convention, and if you've tried with a lot of images it should be correct.

My comment that S is scale and L is shear in `AffineTransform`, is still true. So please update your explanation.

pieleric: Oooohhh... of course. Sorry, I didn't realise.

pieleric:
Updated.

PR #1466, opened 2020-07-13
util transform: check the type of the init arguments

pieleric:
There was very little check on the arguments passed.
In particular, scale should be two values, excepted for the SimilarityTransform
where it's a single value. On the AffineTransform, passing a single
value kind-of work, but the inverse returns something wrong.

Also update the test cases to always pass double value when required.


src/odemis/util/transform.py
```diff
@@ -919,7 +936,7 @@ def __call__(self, x):
         v = numpy.dot(M, self.coeffs)
 
         if x.ndim == 1:
-            return np.array((v.real, v.imag))
+            return numpy.array((v.real, v.imag))
```
tepals, line 939: Good catch! I guess this isn't tested anywhere.

src/odemis/util/transform.py
```diff
@@ -886,11 +903,11 @@ def __init__(self, coeffs=None, rotation=None, scale=None, shear=None,
...
         else:
-            self.rotation = 0. if rotation is None else rotation
-            self.scale = (1., 1.) if scale is None else scale
-            self.shear = 0. if shear is None else shear
+            GeometricTransform.__init__(self, rotation=rotation, scale=scale,
+                                        shear=shear, translation=translation)
             self.nlcoeffs = (0., 0., 0., 0., 0., 0., 0.) if nlcoeffs is None else nlcoeffs
-            self.translation = (0., 0.) if translation is None else translation
+            if len(self.nlcoeffs) != 7 or not all(isinstance(a, numbers.Real) for a in self.nlcoeffs):
+                raise ValueError("Scale should be 7 floats, but got %s" % (self.nlcoeffs,))
```
tepals, line 910: Copy paste error? I think "Scale ....." should be "Coefficients ...." .

src/odemis/util/test/transform_test.py
```diff
@@ -51,65 +51,65 @@
...
-    (ROT90, 2., 0., -TX, [(-1., 0.), (-1., 2.), (-3., 0.)]),
-    (ROT90, 2., 0., -TY, [(0., -1.), (0., 1.), (-2., -1.)]),
-    (ROT90, 2., 0., TXY, [(2., 3.), (2., 5.), (0., 3.)]),
-    (ROT180, 2., 0., T0, [(0., 0.), (-2., 0.), (0., -2.)]),
-    (ROT180, 2., 0., TX, [(1., 0.), (-1., 0.), (1., -2.)]),
-    (ROT180, 2., 0., TY, [(0., 1.), (-2., 1.), (0., -1.)]),
-    (ROT180, 2., 0., -TX, [(-1., 0.), (-3., 0.), (-1., -2.)]),
-    (ROT180, 2., 0., -TY, [(0., -1.), (-2., -1.), (0., -3.)]),
-    (ROT180, 2., 0., TXY, [(2., 3.), (0., 3.), (2., 1.)]),
+    (ROT45, (SQ2, SQ2), 0., T0, [(0., 0.), (1., 1.), (-1., 1.)]),
```
tepals, line 83: Why did you decide to make the scale a tuple for the `SIMILARITY_KNOWN_VALUES`?

src/odemis/util/test/transform_test.py
```diff
@@ -51,65 +51,65 @@
...
-    (ROT90, 1., 0., -TX, [(-1., 0.), (-1., 1.), (-2., 0.)]),
-    (ROT90, 1., 0., -TY, [(0., -1.), (0., 0.), (-1., -1.)]),
-    (ROT90, 1., 0., TXY, [(2., 3.), (2., 4.), (1., 3.)]),
-    (ROT180, 1., 0., T0, [(0., 0.), (-1., 0.), (0., -1.)]),
-    (ROT180, 1., 0., TX, [(1., 0.), (0., 0.), (1., -1.)]),
-    (ROT180, 1., 0., TY, [(0., 1.), (-1., 1.), (0., 0.)]),
-    (ROT180, 1., 0., -TX, [(-1., 0.), (-2., 0.), (-1., -1.)]),
-    (ROT180, 1., 0., -TY, [(0., -1.), (-1., -1.), (0., -2.)]),
-    (ROT180, 1., 0., TXY, [(2., 3.), (1., 3.), (2., 2.)]),
+    (0., (1., 1.), 0., T0, [(0., 0.), (1., 0.), (0., 1.)]),
```
tepals, line 54: Shouldn't all the translations also be changed to tuples?

src/odemis/util/test/transform_test.py
```diff
@@ -149,60 +149,60 @@
 
 AFFINE_KNOWN_VALUES = copy.copy(SIMILARITY_KNOWN_VALUES)
 AFFINE_KNOWN_VALUES.extend([
```
tepals, line 151: Can you also add the commented explanation line here?

`# (rotation, scale, shear, translation, [(x0, y0), (x1, y1), ...])`

src/odemis/util/transform.py
```diff
@@ -56,13 +56,12 @@
...
-
-import numpy
-import scipy.optimize
 from future.utils import with_metaclass
+import numbers
+import numpy
 from numpy.linalg import LinAlgError
-
 from odemis.util.linalg import qrp, tri_inv
+import scipy.optimize
```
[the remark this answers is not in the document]
pieleric: I didn't really mean it. Pydev decided :-D
I guess it's not so clever. I'll slightly move them back around.

src/odemis/util/transform.py
pieleric: Yeah... Pydev found it (so it's a little bit clever).
That said, if you want to add a test case, that'd be welcome ;-)

src/odemis/util/transform.py
pieleric: Yep

src/odemis/util/test/transform_test.py
pieleric: Because it's also used by the other transformations too. That's why in the actual similarity tests I explicitly change the tuple to a single value.

src/odemis/util/test/transform_test.py
pieleric: They are tuples.... All these constants are of length 2. Yeah, so not easy to read.
Actually, SQ2 is a single value, and S23 is a tuple :-/

pieleric:
Updated with all your comments

PR #1472, opened 2020-08-10
CryoSECOM position switching procedures

lazem:
The following is an implementation for the procedures to switch between loading/imaging/tilting position for the CryoSECOM stage. 
Two functions in `odemis.acq.move module`:
**cryoTiltSample**: Allows to switch between imaging ↔ tilted. 
**cryoLoadSample**: Allows to switch between imaging ↔ loading.

### DIFF
--- a/src/odemis/acq/move.py
+++ b/src/odemis/acq/move.py
@@ -1,0 +1,317 @@
+# -*- coding: utf-8 -*-
+"""
+Created on 27 July 2020
+
+@author: Éric Piel, Bassim Lazem
+
+Copyright © 2020 Delmic
+
+This file is part of Odemis.
+
+Odemis is free software: you can redistribute it and/or modify it under the
+terms  of the GNU General Public License version 2 as published by the Free
+Software  Foundation.
+
+Odemis is distributed in the hope that it will be useful, but WITHOUT ANY
+WARRANTY;  without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+PARTICULAR  PURPOSE. See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with
+Odemis. If not, see http://www.gnu.org/licenses/.
+"""
+
+from __future__ import division
+
+import logging
+import math
+import threading
+from asyncio import CancelledError
+from concurrent.futures._base import CANCELLED, RUNNING, FINISHED
+
+import numpy
+import scipy
+
+from odemis import model, util
+from odemis.util import executeAsyncTask
+
+MAX_SUBMOVE_DURATION = 60  # s
+LOADING, IMAGING = 0, 1
+ATOL_LINEAR_POS = 100e-6  # m
+ATOL_ROTATION_POS = 1e-3  # rad (~0.5°)
+RTOL_PROGRESS = 0.1
+
+
+def getLoadingProgress(current_pos, start_pos, end_pos):
+    """
+    Computes the position on the path between LOADING and IMAGING.
+    If it’s too far from the line between LOADING → IMAGING, then it’s considered out of the path.
+    :param current_pos: (dict str->float) Current position of the stage
+    :param start_pos: (dict str->float) Either LOADING or IMAGING position to start the movement from
+    :param end_pos: (dict str->float)Either LOADING or IMAGING position to end the movement to
+    :return:(0<=float<=1, or None) Ratio of the progress, None if it's far away from of the path
+    """
+
+    def get_distance(start, end):
+        # Calculate the euclidean distance between two 3D points
+        sp = numpy.array([start['x'], start['y'], start['z']])
+        ep = numpy.array([end['x'], end['y'], end['z']])
+        return scipy.spatial.distance.euclidean(ep, sp)
+
+    def check_axes(pos):
+        if not {'x', 'y', 'z'}.issubset(set(pos.keys())):
+            raise ValueError("Missing x,y,z axes in {} for correct distance measurement.".format(pos))
+
+    # Check we have the x,y,z axes in all points
+    check_axes(current_pos)
+    check_axes(start_pos)
+    check_axes(end_pos)
+    # Get distance for current point in respect to start and end
+    from_start = get_distance(start_pos, current_pos)
+    to_end = get_distance(current_pos, end_pos)
+    total_length = get_distance(start_pos, end_pos)
+    # Check if current position is on the line from start to end position
+    # That would happen if start_to_current +  current_to_start = total_distance from start to end
+    if util.almost_equal((from_start + to_end), total_length, rtol=RTOL_PROGRESS):
+        return min(from_start / total_length, 1.0)  # Clip in case from_start slightly > total_length
+    else:
+        return None
+
+
+def _isInRange(stage, axis):
+    """
+    Check if current position is within active range
+    :return: True if position in active range, False otherwise
+    """
+    if axis not in stage.axes.keys():
+        raise ValueError("Axis %s is not found in stage axes", axis)
+    pos = stage.position.value[axis]
+    active_range = [r for r in stage.getMetadata()[model.MD_POS_ACTIVE_RANGE][axis]]
+    # Add 1% margin for hardware slight errors
+    margin = (active_range[1] - active_range[0]) * 0.01
+    return (active_range[0] - margin) <= pos <= (active_range[1] + margin)
+
+
+def _isNearDeactive(current_pos, stage_deactive, axis):
+    """
+    Check whether given axis is near stage's FAV_POS_DEACTIVE position
+    :return: True if the axis is near (within 1 mm or 0.1 rad) FAV_POS_DEACTIVE, False otherwise
+    :raises ValueError if axis is unknown
+    """
+    current_value = current_pos[axis]
+    deactive_value = stage_deactive[axis]
+    if axis in {'x', 'y', 'z'}:
+        return abs(deactive_value - current_value) < ATOL_LINEAR_POS
+    elif axis in {'rx', 'rz'}:
+        return util.rot_almost_equal(current_value, deactive_value, atol=ATOL_ROTATION_POS)
+    else:
+        raise ValueError("Unknown axis value %s.", axis)
+
+
+def cryoLoadSample(target):
+    """
+    Provide the ability to switch between loading position and imaging position, without bumping into anything.
+    :param target: (int) target position either one of the constants LOADING or IMAGING
+    :return (CancellableFuture -> None): cancellable future of the move to observe the progress, and control the
+   raise ValueError exception
+    """
+    # Get the stage and focus components from the backend components
+    stage = model.getComponent(role='stage')
+    focus = model.getComponent(role='focus')
+
+    f = model.CancellableFuture()
+    f.task_canceller = _cancelCryoMoveSample
+    f._task_state = RUNNING
+    f._task_lock = threading.Lock()
+    f._running_subf = model.InstantaneousFuture()
+    # Run in separate thread
+    executeAsyncTask(f, _doCryoLoadSample, args=(f, stage, focus, target))
+    return f
+
+
+def _doCryoLoadSample(future, stage, focus, target):
+    """
+    Do the actual switching procedure for the Cryo sample stage between loading and imaging
+    :param future: cancellable future of the move
+    :param stage: sample stage that's being controlled
+    :param focus: focus for optical lens
+    :param target: target position either one of the constants LOADING or IMAGING
+    """
+    try:
+        stage_md = stage.getMetadata()
+        focus_md = focus.getMetadata()
+        stage_active = stage_md[model.MD_FAV_POS_ACTIVE]
+        stage_deactive = stage_md[model.MD_FAV_POS_DEACTIVE]
+        focus_deactive = focus_md[model.MD_FAV_POS_DEACTIVE]
+        # Create axis->pos dict from target position given smaller number of axes
+        filter_dict = lambda keys, d: {key: d[key] for key in keys}
+
+        if target == LOADING:
+            # Check that the stage X,Y,Z are within the limits
+            for axis in {'x', 'y', 'z'}:
+                if not _isInRange(stage, axis):
+                    logging.warning("Axis %s position is out of active range.", axis)
+            # Check that current position is near the ACTIVE to DEACTIVE line
+            # (ie, in case it was previously stopped half-way through loading/unloading)
+            if getLoadingProgress(stage.position.value, stage_active, stage_deactive) is None:
+                logging.warning("Stage position is not near the ACTIVE to DEACTIVE line.")
+
+            if abs(stage_deactive['rx']) > ATOL_ROTATION_POS:
+                raise ValueError(
+                    "Absolute value of rx for FAV_POS_DEACTIVE is greater than {}".format(ATOL_ROTATION_POS))
+
+            # Create an ordered sub moves list to perform the loading move
+            sub_moves = [
+                (focus, focus_deactive),
+                (stage, filter_dict({'rx', 'rz'}, stage_deactive)),
+                (stage, filter_dict({'x', 'y'}, stage_deactive)),
+                (stage, filter_dict({'z'}, stage_deactive)),
+            ]
+        elif target == IMAGING:
+            current_pos = stage.position.value
+            for axis in stage.axes:
+                # Check if the current position is near DEACTIVE position (within 1 mm or 0.1 rad)
+                if not _isNearDeactive(current_pos, stage_deactive, axis):
+                    raise ValueError("Axis %s position is not near FAV_POS_DEACTIVE position.", axis)
+
+            focus_active = focus_md[model.MD_FAV_POS_ACTIVE]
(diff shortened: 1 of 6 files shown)


src/odemis/acq/move.py
```diff
@@ -0,0 +1,314 @@
...
+            if not getLoadingProgress(stage.position.value, stage_active, stage_deactive) <= 0.1:
+                logging.warning("Stage position is not near the ACTIVE to DEACTIVE line.")
+
+            if abs(stage.getMetadata()[model.MD_FAV_POS_DEACTIVE]['rx']) > 0.1:
+                raise ValueError("Absolute value of rx for FAV_POS_DEACTIVE is greater than 0.1")
+
+            # Create an ordered sub moves list to perform the loading move
+            sub_moves = [
+                (focus, focus_deactive),
+                (stage, axis_dict({'z'}, stage_deactive)),
```
lazem, line 200: Added a move in Z axis here though it was not mentioned in the spec. document. The reason is that the pretesting values (which I used) had two different values for Z for imaging and loading. 

```
loading = {'rx': 0.00, 'rz': 0.0, 'x':  0.01, 'y':  0.00, 'z':  -0.001}
imaging = {'rx': 0.00, 'rz': 0.0, 'x':  0.0, 'y':  -0.002, 'z':  0.002}
```

src/odemis/acq/move.py
```diff
@@ -0,0 +1,314 @@
...
+    :param future: cancellable future of the whole move
+    :param driver: Either the stage of the focus driver
+    :param sub_move_dict: the sub_move axis->pos dict
+    :raises TimeoutError: if the sub move timed out
+    :raises CancelledError: if the sub move is cancelled
+    """
+    try:
+        logging.debug("Performing sub move {} -> {}".format(driver.name, sub_move_dict))
+        future._running_subf = driver.moveAbs(sub_move_dict)
+        future._running_subf.result(timeout=TIMEOUT_DURATION)
```
lazem, line 307: Added a timeout though it's inspired by the autofocus function and may not be needed here.

src/odemis/acq/move.py
```diff
@@ -0,0 +1,314 @@
...
+        logging.exception(str(exp))
+        raise
+    finally:
+        with future._task_lock:
+            if future._task_state == CANCELLED:
+                raise CancelledError()
+            future._task_state = FINISHED
+
+
+def run_sub_move(future, driver, sub_move_dict):
```
pieleric, line 295: driver -> comp

src/odemis/acq/move.py
```diff
@@ -0,0 +1,314 @@
...
+            if rx is not None:
+                sub_moves.append((stage, {'rx': rx}))
+            if rz is not None:
+                sub_moves.append((stage, {'rz': rz}))
+
+        if future._task_state == CANCELLED:
+            logging.info("CryoTiltSample procedure cancelled before any sub move.")
+            raise CancelledError()
+
+        for driver, sub_move_dict in sub_moves:
```
pieleric, line 281: driver -> comp

src/odemis/acq/move.py
```diff
@@ -0,0 +1,314 @@
...
+            run_sub_move(future, driver, sub_move_dict)
+    except CancelledError:
+        logging.info("_doCryoTiltSample cancelled.")
+    except Exception as exp:
+        logging.exception(str(exp))
```
pieleric, line 286: Please put a message indicating to which tilting position it was requested

src/odemis/acq/move.py
```diff
@@ -0,0 +1,314 @@
...
+    :param end_pos: Either LOADING or IMAGING position to end the movement to
+    :return: Percentage of the progress, None if it's out of the path
+    """
+
+    def get_distance(start, end):
+        # Calculate the euclidean distance between two 3D points
+        sp = np.array([start['x'], start['y'], start['z']])
+        ep = np.array([end['x'], end['y'], end['z']])
+        squared_dist = np.sum((ep - sp) ** 2, axis=0)
+        dist = np.sqrt(squared_dist)
```
tepals, line 56: I think `numpy.linalg.norm(ep-sp)` should do the same as these two lines.

https://numpy.org/doc/stable/reference/generated/numpy.linalg.norm.html

src/odemis/acq/move.py
```diff
@@ -0,0 +1,314 @@
...
+"""
+
+from __future__ import division
+
+import logging
+import threading
+from asyncio import CancelledError
+from concurrent.futures._base import CANCELLED, RUNNING, FINISHED
+
+import numpy as np
```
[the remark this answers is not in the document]
tepals: Out of context of this PR: why is this the convention in Odemis? Or is it just because it's always been like this.

src/odemis/acq/test/move_test.py
```diff
@@ -0,0 +1,188 @@
...
+        if self.backend_was_running:
+            self.skipTest("Running backend found")
+
+    def test_loading(self):
+        """
+        Test moving the sample stage to loading position while it's in imaging position
+        """
+        stage = self.stage
+        # Get the stage to imaging position
+        f = stage.moveAbs(stage.getMetadata()[model.MD_FAV_POS_ACTIVE])
```
tepals, line 75: you're calling `stage.getMetadata()[model.MD_FAV_POS_ACTIVE]` in pretty much every test, wouldn't it be better to move it to the setup class?

src/odemis/acq/test/move_test.py
```diff
@@ -0,0 +1,188 @@
...
+        """
+        stage = self.stage
+        f = stage.moveAbs(stage.getMetadata()[model.MD_FAV_POS_ACTIVE])
+        f.result()
+        f = cryoLoadSample(LOADING)
+        time.sleep(2)
+        cancelled = f.cancel()
+        self.assertTrue(cancelled)
+        with self.assertRaises(AssertionError):
+            test.assert_pos_almost_equal(stage.position.value, stage.getMetadata()[model.MD_FAV_POS_DEACTIVE],
```
tepals, line 152: Just a note, and it's fine to leave you're code like it is: this feels like we're missing a `test.assert_pos_not_equal()` function.

src/odemis/acq/move.py
tepals: or `scipy.spatial.distance.euclidean(ep, sp)`

https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.distance.euclidean.html

pieleric:
> Out of context of this PR: why is this the convention in Odemis? Or is it just because it's always been like this.

It's a conscious decision I made. `numpy` is not really special nor really a long name, so no need to rename it as `np`. From this time on, we only use `numpy` in Odemis.

src/odemis/acq/move.py
pieleric: To continue on alternatives, you could also write: `math.hypot(*(ep - sp))`.... but that only works on Python 3.8 with more than 2 dimensions... so not really an alternative ;-)

src/odemis/acq/test/move_test.py
pieleric: Indeed, we've recently added `assert_array_not_equal()`, and a `assert_pos_not_almost_equal()` could be handy. In particular, instead of this implementation, if the keys are different, it should (still) fail.

src/odemis/acq/move.py
```diff
@@ -0,0 +1,314 @@
...
+            if getLoadingProgress(stage.position.value, stage_active, stage_deactive) is None:
+                logging.warning("Stage position is not near the ACTIVE to DEACTIVE line.")
+
+            if abs(stage_deactive['rx']) > ATOL_ROTATION_POS:
+                raise ValueError("Absolute value of rx for FAV_POS_DEACTIVE is greater than {}".format(ATOL_ROTATION_POS))
+
+            # Create an ordered sub moves list to perform the loading move
+            sub_moves = [
+                (focus, focus_deactive),
+                (stage, filter_dict({'z'}, stage_deactive)),
```
pieleric, line 202: Sorry, now that I think of it, I think it's probably safer to move the "z" near the loading position (because there are less things that could collide there). So you should do this move last, and for the "IMAGING", you should do this move first.
I'll update the spec (for us to remember).

src/odemis/acq/test/move_test.py
lazem: Cool, I will include `assert_pos_not_almost_equal` to the `test` module.

src/odemis/acq/test/move_test.py
```diff
@@ -0,0 +1,193 @@
...
+import odemis
+from odemis import model
+from odemis.acq.move import LOADING, IMAGING, getLoadingProgress
+from odemis.acq.move import cryoTiltSample, cryoLoadSample
+from odemis.util import test
+
+logging.getLogger().setLevel(logging.DEBUG)
+
+CONFIG_PATH = os.path.dirname(odemis.__file__) + "/../../install/linux/usr/share/odemis/"
+CRYO_SECOM_CONFIG = CONFIG_PATH + "sim/cryosecom-sim.yaml"
```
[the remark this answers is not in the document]
lazem: Yes, I have actually put down another file `cryosecom-poc-sim.odm.yaml` replacing the camera with `sim_cam`. I can merge both together but should it be included in the PR? @pieleric

src/odemis/acq/move.py
lazem: Makes sense, I will do as such.

src/odemis/acq/test/move_test.py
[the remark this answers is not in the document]
pieleric: Yes, please, just replace cryosecom-sim.yaml by the content of your new simulator (using simcam). Let's not have 2 almost identical simulators. One should be enough.
Preferably, in this PR, push  a version which works (eg, with the simcam, but without maxres), and in the overview acq PR, you'd push the changes to the simulator (adding maxres).

lazem:
Hi @pieleric I added the suggested changes and rebased with master. Can you please check? 
Thank you

src/odemis/acq/move.py
```diff
@@ -0,0 +1,318 @@
...
+    :return: True if the axis is near (within 1 mm or 0.1 rad) FAV_POS_DEACTIVE, False otherwise
+    :raises ValueError if axis is unknown
+    """
+    current_value = current_pos[axis]
+    deactive_value = stage_deactive[axis]
+    if axis in {'x', 'y', 'z'}:
+        return abs(deactive_value - current_value) < 0.001
+    elif axis in {'rx', 'rz'}:
+        diff = math.degrees(abs(deactive_value - current_value))
+        return math.radians(diff % 360) < ATOL_ROTATION_POS
```
pieleric, line 106: No, this is still incorrect. For example, if deactive = 1° and current = 359°. `diff` is 358°, which is still 358° after the modulo.

The formula is:
`abs(((deactive_value - current_value) + math.pi) % (2 * math.pi) - math.pi) < ATOL_ROTATION_POS`

Admittedly, it's not an absolutely obvious formula to read or write, and we already have 3 or 4 duplicates. We should maybe think of making a dedicated function as `odemis.util.rot_almost_equal()`...

src/odemis/acq/move.py
```diff
@@ -0,0 +1,318 @@
...
+
+
+def getLoadingProgress(current_pos, start_pos, end_pos):
+    """
+    Computes the position on the path between LOADING and IMAGING.
+    If it’s too far from the line between LOADING → IMAGING, then it’s considered out of the path.
+    :param current_pos: Current position of the stage
+    :param start_pos: Either LOADING or IMAGING position to start the movement from
+    :param end_pos: Either LOADING or IMAGING position to end the movement to
+    :return: Percentage of the progress, None if it's far away from of the path
```
pieleric, line 51: Please indicate the types of the arguments. the positions are `dict str->float`. The return is `0<=float<=1, or None`. Also, the return value is not a "percentage" (ie, between 0 and 100), but a "ratio".

src/odemis/acq/test/move_test.py
```diff
@@ -0,0 +1,192 @@
...
+        if self.backend_was_running:
+            self.skipTest("Running backend found")
+
+    def test_loading(self):
+        """
+        Test moving the sample stage to loading position while it's in imaging position
+        """
+        stage = self.stage
+        # Get the stage to imaging position
+        f = stage.moveAbs(self.stage_active)
```
pieleric, line 79: On the real hardware, that's kind of dangerous to do.... and it's exactly why we are designing a special function ;-)
I'd recommend to merge `test_loading` and `test_imaging_from_loading`. In it, start with a call to cryoLoadSample(LOADING), do some basic tests, call cryoLoadSample(IMAGING), more checks, and then  cryoLoadSample(LOADING) with full checks.

src/odemis/acq/test/move_test.py
```diff
@@ -0,0 +1,192 @@
...
+        """
+        Test moving the sample stage to imaging position while it's in tilting position
+        """
+        stage = self.stage
+        focus = self.focus
+        # Get the stage to tilting position and park focus
+        f = stage.moveAbs(self.stage_active)
+        f.result()
+        f = focus.moveAbs(self.focus_deactive)
+        f.result()
```
pieleric, line 108: Use `cryoLoadSample(IMAGING)`

src/odemis/acq/test/move_test.py
```diff
@@ -0,0 +1,192 @@
...
+        """
+        Test tilting the sample stage while the stage is in Imaging position
+        """
+        stage = self.stage
+        focus = self.focus
+        # Get the stage to imaging position and park focus
+        f = stage.moveAbs(self.stage_active)
+        f.result()
+        f = focus.moveAbs(self.focus_deactive)
+        f.result()
```
pieleric, line 126: Use `cryoLoadSample(IMAGING)`

src/odemis/acq/move.py
tepals, line 106: Why are you converting between degrees and radians? This should do the same:
`return abs(deactive_value - current_value) % (2 * math.pi) < ATOL_ROTATION_POS`

src/odemis/acq/test/move_test.py
```diff
@@ -0,0 +1,192 @@
...
+            return
+        except IOError as exp:
+            logging.error(str(exp))
+            raise
+
+        # find components by their role
+        cls.stage = model.getComponent(role="stage")
+        cls.focus = model.getComponent(role="focus")
+
+        cls.stage_active = cls.stage.getMetadata()[model.MD_FAV_POS_ACTIVE]
```
tepals, line 60: Why not add `stage_deactive` here as well?

src/odemis/acq/test/move_test.py
```diff
@@ -0,0 +1,198 @@
...
+                                     atol=ATOL_STAGE)
+        # Get the stage to imaging position
+        f = cryoLoadSample(IMAGING)
+        f.result()
+        test.assert_pos_almost_equal(stage.position.value, self.stage_active, atol=ATOL_STAGE)
+        # Switch back to loading position
+        f = cryoLoadSample(LOADING)
+        f.result()
+        test.assert_pos_almost_equal(stage.position.value, self.stage_deactive, atol=ATOL_STAGE)
+        test.assert_pos_almost_equal(focus.position.value, self.focus_deactive, atol=ATOL_STAGE)
```
lazem, line 93: Any more tests you had in mind other than these, @pieleric

src/odemis/acq/move.py
```diff
@@ -0,0 +1,317 @@
...
+    Check whether given axis is near stage's FAV_POS_DEACTIVE position
+    :return: True if the axis is near (within 1 mm or 0.1 rad) FAV_POS_DEACTIVE, False otherwise
+    :raises ValueError if axis is unknown
+    """
+    current_value = current_pos[axis]
+    deactive_value = stage_deactive[axis]
+    if axis in {'x', 'y', 'z'}:
+        return abs(deactive_value - current_value) < ATOL_LINEAR_POS
+    elif axis in {'rx', 'rz'}:
+        return util.rot_almost_equal(current_value, deactive_value, atol=ATOL_ROTATION_POS)
```
lazem, line 105: Thanks for clarifying this, added `rot_almost_equal` can you please check? @pieleric

lazem:
> It looks good to me. Could you squash it into 3 commits:
> 
> * one for the util.test
> * one for util.rot_almost_equal
> * one for "the rest"

Done as requested @pieleric

src/odemis/util/__init__.py
```diff
@@ -102,6 +102,26 @@ def almost_equal(a, b, atol=1e-18, rtol=1e-7):
...
+    atol (float): absolute tolerance
+    rtol (float): relative tolerance
+    returns (bool): True if a and b rotation is within a margin
+    """
+    if a == b:
+        return True
+
+    tau = 2 * math.pi
+    tol = max(atol, max(abs(a % tau), abs(b % tau)) * rtol)
+    if abs(((a - b) + math.pi) % (2 * math.pi) - math.pi) <= tol:
```
tepals, line 119: Small comment: you can replace the `(2 * math.pi)` here by tau

tepals (approved):

src/odemis/util/__init__.py
lazem: Good catch! Replaced it.

PR #1473, opened 2020-08-12
driver-tmcm: canopen driver

philipwinklerdelmic:


src/odemis/driver/tmcm.py
```diff
@@ -2502,3 +2553,871 @@ def _parseMessage(self, msg):
...
+        """
+        return (int, int, int):
+             Controller ID
+             Firmware major version number
+             Firmware minor version number
+        """
+        cont = self.GetAxisParam(axis, DEVICE_NAME)
+        vmaj = self.GetAxisParam(axis, HW_VERSION)  # e.g. '1.0'
+        vmin = self.GetAxisParam(axis, IDENTITY, 3)
+        return cont, int(float(vmaj)), int(vmin)
```
Kleijwegt, line 2749: Why do you use int(float(vmaj))) and not just int(vmaj)?

src/odemis/driver/tmcm.py
```diff
@@ -2502,3 +2553,871 @@ def _parseMessage(self, msg):
...
+
+    def GetAxisParam(self, axis, param, idx=None):
+        """
+        Read the axis/parameter setting from the RAM
+        axis (0<=int<=5): axis number
+        param (0<=int<=255): parameter number
```
Kleijwegt, line 2755: Would param_nmr not be a more descriptive name?

src/odemis/driver/tmcm.py
philipwinklerdelmic: vmaj is "1.0", directly casting it to int doesn't work. I will add a comment to explain.

src/odemis/driver/tmcm.py
philipwinklerdelmic: I use the same convention as in the previous tmcm driver (corresponding function has exact same parameters).
Usually, "axis" always refers to the axis number, "axname" or "axis_name" to the name (like 'x'). That avoids long variable names.

src/odemis/driver/tmcm.py
```diff
@@ -2502,3 +2553,871 @@ def _parseMessage(self, msg):
...
+                        homingstatus = key
+                if homingstatus in ('INTERRUPTED', 'ERROR VELOCITY IS NOT ZERO', 'ERROR VELOCITY IS ZERO'):
+                    raise RuntimeError('Unable to home. Reason: {0}'.format(homingstatus))
+                time.sleep(0.001)
+                if time.time() > t:
+                    raise RuntimeError('Unable to home, timeout reached')
+            if set_new_home:
+                actual_position = self._node.sdo[ACTUAL_POSITION].raw
+                self._node.sdo[HOME_OFFSET].raw = actual_position  # home offset (0x607C)
+                logging.info('Homing offset set to {0}'.format(actual_position))
```
pieleric, line 3258: I have the feeling this works only for the referencing of the first axis. Or is there something special with referencing which make all the axes move at the same time? For now, we can't test with several axes, so unless it's clear to you how to deal with multiple axes, I'd suggest to put a big TODO at the top of this function. Maybe even pass an "axis" parameter and raise an Error if axis != 0.

src/odemis/driver/test/tmcm_test.py
```diff
@@ -581,6 +581,338 @@ def test_open_shutter(self):
...
+
+    def setUp(self):
+        if TEST_NOHW:
+            address = "fake"
+        else:
+            address = "can0"
+
+        current_dir = os.path.abspath(os.path.dirname(__file__))
+        parent_dir = os.path.abspath(current_dir + "/../")
+        datasheet = os.path.join(parent_dir, 'TMCM-1240_CANopen_V322.dcf')
```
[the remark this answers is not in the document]
philipwinklerdelmic: Do you mean `current_dir` (instead of `parent_dir`?

src/odemis/driver/tmcm.py
```diff
@@ -2502,3 +2553,871 @@ def _parseMessage(self, msg):
...
+          Note: If the axis is inverted, the values provided will be inverted too.
+        unit (None or list of str): The unit of each axis. When it's None, it
+          defaults to "rad" for all the axes.
+        refproc (str or None): referencing (aka homing) procedure type. Use
+          None to indicate it's not possible (no reference/limit switch) or the
+          name of the procedure. For now only "Standard" is accepted.
+        """
+
+        if ustepsize is None:
+            ustepsize = [DEFAULT_STEPSIZE] * len(axes)
```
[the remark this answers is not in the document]
philipwinklerdelmic: Ok. It was quite hard though to figure out the stepsize for this device. I'll use the value now in the testcases. Should I also make an example yaml file or put the stepsize value somewhere else, so we can easily find it in the future?

src/odemis/driver/test/tmcm_test.py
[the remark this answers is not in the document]
pieleric: Indeed! `datasheet = os.path.join(current_dir, "..","TMCM-1240_CANopen_V322.dcf")`

src/odemis/driver/tmcm.py
```diff
@@ -2502,3 +2553,871 @@ def _parseMessage(self, msg):
...
+        Requests a move to a relative position. This is non-blocking.
+        axis (0<=int<=5): axis number
+        offset (-2**31 <= int 2*31-1): relative position
+        """
+        self.SetAxisParam(axis, TARGET_POSITION, offset)
+        # Bit 6 of controlword: absolute (0) or relative (1)
+        # Bits 0-3 to 1: switch on and enable (see p 71 of canopen firmware manual)
+        # Bit 4: start positioning
+        self._node.controlword = 0xf  # switch on (transition to state 3)
+        self._node.controlword = 0b1011111
```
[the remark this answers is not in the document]
philipwinklerdelmic: Yes, the way it works is that there are different states that the hardware can be in. We have to write the controlword twice, first, to get into "switched on" state, then, to get into "operation enabled" state. I'll add a comment to make it more clear.

src/odemis/driver/tmcm.py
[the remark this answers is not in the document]
pieleric: Yes, of course, the stepsize should be in a YAML file, which is in _some_ git repo. Typically, it's good to have at least one YAML file in Odemis with the driver, so it's easy to find an example. Either add a YAML file with the simulator, or add a special YAML with almost only the hardware support, as we've done for `install/linux/usr/share/odemis/hwtest/pigcs-1-ctrl.odm.yaml`.

src/odemis/driver/tmcm.py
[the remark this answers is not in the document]
pieleric: I see. IMHO, that's one case where Python properties should not be used. If it was a function like ".setControlState()" it would look a lot less suspicious to call it twice with different values. Anyway, let's leave this API as it is ;-)

src/odemis/driver/tmcm.py
```diff
@@ -2502,3 +2553,871 @@ def _parseMessage(self, msg):
...
+        # ._node.statusword is not updated frequently enough, so directly ask for the status via .sdo
+        return bool((self._node.sdo[STATUS_WORD].raw & 0x400))
+
+    def _checkErrorFlag(self, axis):
+        """
+        Raises an HWError if the axis error flag reports an issue
+        """
+        stat = self._node.statusword
+        if stat & 0b1000:
+            raise TMCLError(stat, "Fault detected.")
```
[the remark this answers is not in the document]
philipwinklerdelmic: How about using HwError? Or is it clearer to create an error class called TMCMCANError?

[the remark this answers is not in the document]
pieleric: Yes, raise a HwError, as the version in the TMCLController does (and the docstring says).

src/odemis/driver/tmcm.py
```diff
@@ -2502,3 +2553,871 @@ def _parseMessage(self, msg):
...
+        # Start with creating a network representing one CAN bus
+        network = canopen.Network()
+
+        # Connect to the CAN bus
+        network.connect(bustype='socketcan', channel=channel)
+        network.check()
+
+        # Add some nodes with corresponding Object Dictionaries
+        node = canopen.BaseNode402(nodeid, datasheet)
+        network.add_node(node)
```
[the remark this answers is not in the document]
philipwinklerdelmic: Like this?
```
        # Connect to the CAN bus
        try:
            network.connect(bustype='socketcan', channel=channel)
            network.check()
        except CanError:
            raise HwError("Can adapter not found.")

        # Add some nodes with corresponding Object Dictionaries
        node = canopen.BaseNode402(nodeid, datasheet)
        network.add_node(node)

        # Reset network
        node.nmt.state = 'RESET COMMUNICATION'
        node.nmt.wait_for_bootup(15)
        logging.debug('Device state after reset = {0}'.format(node.nmt.state))

        # Transmit SYNC every 100 ms
        network.sync.start(0.1)

        try:
            node.load_configuration()
            node.setup_402_state_machine()
        except ValueError:
            raise HwError("Node not present on the bus.")
```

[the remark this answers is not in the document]
pieleric: Yes... I trust you've tested it with both cases and that's indeed the exceptions that happen in both cases :-)

Just a small request, please add a bit more info in the error messages. Indicate the "channel" in the CAN adapter error (and also spell "CAN" in uppercase, to avoid confusion). For the node missing, please add both the channel and nodeid.

src/odemis/driver/tmcm.py
```diff
@@ -2502,3 +2553,871 @@ def _parseMessage(self, msg):
...
+        self.SetAxisParam(axis, PULLUP_RESISTORS, 0x1)  # Enable pull-up resistor for HOME input
+        # Encoder: 0 internal magnetic encoder, -1 (default) no encoder, -2 external encoder
+        self.SetAxisParam(axis, SENSOR_SELECTION, 0)  # Use Internal magnetic encoder
+        self.SetAxisParam(axis, POSITION_WINDOW, 0)  # enable position window 0 (target = demand position)
+        self.SetAxisParam(axis, POSITION_WINDOW_TIME, 100)  # remain in window for 100 ms before setting status bit
+
+        self._node.op_mode = HOMING_MODE
+        self.SetAxisParam(axis, HOMING_METHOD, Homing.HM_ON_POSITIVE_INDEX_PULSE)
+        self._node.state = OPERATION_ENABLED
+        self._node.op_mode = PP_MODE
```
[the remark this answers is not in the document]
philipwinklerdelmic: The specification is not really clear, the parameter is only defined in the "homing mode" section, not the "profiled position mode" section. However, I just tested it on the hardware and it also seems to be working when set from profiled position mode.

install/linux/usr/share/odemis/hwtest/tmcm-can-ctrl.odm.yaml
```diff
@@ -0,0 +1,20 @@
...
+    role: brightfield,
+    children: [Light, TMCM_CAN],
+}
+
+Light: {
+    class: simulated.Light,
+    role: brightlight,
+}
+
+TMCM_CAN: {
```
pieleric, line 12: Let's call this component with an easier name to spell "TMCMCan" would be OK, or maybe "Rotator".

install/linux/usr/share/odemis/hwtest/tmcm-can-ctrl.odm.yaml
```diff
@@ -0,0 +1,20 @@
...
+    class: tmcm.CANController,
+    role: null,
```
pieleric, line 14: Let's put a role, just for the sake of it. Maybe "rotator" ? Or what role will this hardware have in the Sonic?

install/linux/usr/share/odemis/hwtest/tmcm-can-ctrl.odm.yaml
```diff
@@ -0,0 +1,20 @@
+SimOptical: {
+    class: Microscope,
+    role: brightfield,
+    children: [Light, TMCM_CAN],
+}
+
+Light: {
+    class: simulated.Light,
+    role: brightlight,
+}
```
pieleric, line 10: Actually, you can drop this. It was there to be able to start the GUI, but as there is no stage and no CCD, that's a bit hopeless anyway.

install/linux/usr/share/odemis/hwtest/tmcm-can-ctrl.odm.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+TMCM_CAN: {
+    class: tmcm.CANController,
+    role: null,
+    init: {
+        channel: "can0", node_id: 1, datasheet: "TMCM-1240_CANopen_V322.dcf",
+        axes: {"x"}, refproc: "Standard", param_file: "test/tmcm-pd1240.tmcm.tsv",
+        ustepsize: [0.0001227184630308513],  # 2 * math.pi / (200 * 2 ** 8)
```
pieleric, line 18: Please, put each key on a separate line.
Also, add the unit as "rad". Otherwise a stepsize depending on PI sounds a little fancy ;-)

src/odemis/driver/test/tmcm-pd1240.tmcm.tsv
```diff
@@ -0,0 +1,12 @@
+# Parameters from PD-1240
+# Bank/Axis	Address	Value	Description
```
pieleric, line 2: Drop the "bank" from this comment.

src/odemis/driver/test/tmcm-pd1240.tmcm.tsv
```diff
@@ -0,0 +1,12 @@
...
+# Axis 0 = S
```
pieleric, line 3: S == ???

src/odemis/driver/test/tmcm-pd1240.tmcm.tsv
```diff
@@ -0,0 +1,12 @@
...
+A0	0x2005	0x03	# Limit switch inputs
+A0	0x2710	0x01	# Pull-up resistors
+A0	0x606a	0x00	# Sensor selection
+A0	0x6067	0x00	# Position window
+A0	0x6068	0x64	# Position window time
+A0	0x6098	0x22	# Homing method
+A0	0x607C	0x00	# Home offset
+A0	0x605A	0x06	# Quick stop option
```
pieleric, line 11: Drop the "A". It's always an axis. You could even drop this whole column, and let the user do +0x800 for the other axes (but that's less user-friendly).

src/odemis/driver/tmcm.py
```diff
@@ -2703,50 +2723,33 @@ def terminate(self):
...
              Firmware minor version number
         """
         cont = self.GetAxisParam(axis, DEVICE_NAME)
-        vmaj = self.GetAxisParam(axis, HW_VERSION)  # e.g. '1.0'
-        vmin = self.GetAxisParam(axis, IDENTITY, 3)
-        return cont, int(float(vmaj)), int(vmin)
+        hw_version = self.GetAxisParam(axis, HW_VERSION)  # e.g. '1.0'
+        sw_version = self.GetAxisParam(axis, SW_VERSION)
+        id = int(self.GetAxisParam(axis, IDENTITY, 3))
+        vmaj = 0xFF00 & id  # first 16 bits
```
pieleric, line 2750: `(0xFF00 & id) >> 16`, to not have gigantic numbers.

src/odemis/driver/tmcm.py
```diff
@@ -2778,12 +2781,12 @@ def MoveAbsPos(self, axis, pos):
...
         pos (-2**31 <= int 2*31-1): position
         """
-        self._node.controlword = 0x0f  # switch on
         self.SetAxisParam(axis, TARGET_POSITION, pos)
         # Bit 6 of controlword: absolute (0) or relative (1)
         # Bits 0-3 to 1: switch on and enable (see p 71 of canopen firmware manual)
         # Bit 4: start positioning
-        self._node.controlword = 0x1f
+        self._node.controlword = 0b0001111  # switch on
+        self._node.controlword = 0b0011111
```
pieleric, line 2789: Please add the comment about double setting the value also here (or just put a "cf MoveRelPos").

src/odemis/driver/tmcm.py
```diff
@@ -3226,28 +3228,32 @@ def homing(self, future, timeout=30, set_new_home=True):
         object (0x607C) to the current position after the homing procedure (default: true)
         :return bool: If the homing was complete with success
         """
+        # TODO: Homing currently only works with one axis
+        if axis != 0:
+            raise ValueError("Homing not supported for axis %s != 0." % axis)
```
pieleric, line 3233: -> `NotImplementedError`

src/odemis/driver/tmcm.py
```diff
@@ -3280,25 +3287,68 @@ def _openCanNode(channel, nodeid, datasheet):
...
+        for l in f:
+            # comment or empty line?
+            mc = re.match(r"\s*(#|$)", l)
+            if mc:
+                logging.debug("Comment line skipped: '%s'", l.rstrip("\n\r"))
+                continue
+            m = re.match(r"(?P<type>[AGO])(?P<num>[0-9]+)\t(?P<param>0x[0-9a-fA-F]+)\t(?P<value>0x[0-9a-fA-F]+)\s*(#.*)?$", l)
+            if not m:
+                raise ValueError("Failed to parse line '%s'" % l.rstrip("\n\r"))
+            typ, num, add, val = m.group("type"), int(m.group("num")), int(m.group("param"), 16), int(m.group("value"), 16)
```
pieleric, line 3344: I think you can drop this "type". It's not useful here.

src/odemis/driver/test/tmcm-pd1240.tmcm.tsv
```diff
@@ -0,0 +1,12 @@
+# Parameters from PD-1240
```
pieleric, line 1: About the file extension, I'd suggest to use a different one than "tmcm.tsv", as it's a different format that these files. I had suggested "tmcc.tsv" in the spec (but feel free to pick something better).

src/odemis/driver/test/tmcm_test.py
```diff
@@ -585,14 +585,16 @@ class TestCanActuator(unittest.TestCase):
...
 
         current_dir = os.path.abspath(os.path.dirname(__file__))
-        parent_dir = os.path.abspath(current_dir + "/../")
-        datasheet = os.path.join(parent_dir, 'TMCM-1240_CANopen_V322.dcf')
-        self.dev = tmcm.CANController("Can Actuator", None, address, 1, datasheet, ['x'], refproc=tmcm.REFPROC_STD)
+        datasheet = os.path.join(current_dir, "..", 'TMCM-1240_CANopen_V322.dcf')
+        # 200 steps / cycle and 2 ** 8 µsteps per step
+        sz = [2 * math.pi / (200 * 2 ** 8)]  # µstep size in rad
+        self.dev = tmcm.CANController("Can Actuator", None, channel, 1, datasheet, ['x'], refproc="Standard",
+                                      param_file=ABS_PATH + "/tmcm-pd1240.tmcm.tsv", ustepsize=sz)
```
pieleric, line 597: Please pass put the name of every parameter. Otherwise it's not very clear what they mean. (eg: 1). Also, please add `rng` and `unit`, at least for the sake of code coverage.

PR #1478, opened 2020-08-17
config: switch to Python 3 by default

pieleric:
Change the shebang header to explicitly call Python 3.

And for cases where the PYTHON_INTERPRETER variable is used, default to
Python 3.

We still support running Odemis on Python 2, but this is now deprecated.

Also remove an old test script, which has been broken for a few years,
and has even less chances to work on Python 3.


src/odemis/acq/align/keypoint.py
```diff
@@ -1,4 +1,3 @@
-#!/usr/bin/env python
```
tepals, line 1: why not replace by python3 here?

src/odemis/model/_dataio.py
```diff
@@ -1,4 +1,3 @@
-#!/usr/bin/env python
```
tepals, line 1: same here

src/odemis/dataio/catmaid.py
```diff
@@ -1,4 +1,3 @@
-#!/usr/bin/python
```
tepals, line 1: same here

src/odemis/acq/stream/_projection.py
```diff
@@ -1,4 +1,3 @@
-#!/usr/bin/env python
```
tepals, line 1: same here

pieleric:
Kornee also noticed this actually. Sorry for not clarifying in the commit message. All these files are not supposed to be executed, they are just Python modules. The convention in Odemis is to **not** have a shebang on such files.

PR #1480, opened 2020-08-19
Xt autofunctionality API

Kleijwegt:


src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +282,93 @@ def test_set_beam_shift(self):
...
+        """
+        # Create sets holding stigmator values to check if the outcome is consistent
+        unique_stigmator_values_x = set()  # Set holding only the unique stigmator values in x direction
+        unique_stigmator_values_y = set()  # Set holding only the unique stigmator values in y direction
+
+        # Run apply_autostigmator multiple times and afterwards check if the outcome is the same for each run
+        for i in range(0, 5):
+            autostigmator_return = self.scanner.apply_autostigmator("electron1", "run")
+            time.sleep(0.2)  # Give microscope/simulator the time to update the state
+            # direct_call = self.microscope.set_autostigmator("electron1", "run")
```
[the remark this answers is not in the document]
Kleijwegt: Yes should have been deleted indeed

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +282,93 @@ def test_set_beam_shift(self):
...
+
+            unique_stigmator_values_x.add(self.microscope.get_stigmator()[0])
+            unique_stigmator_values_y.add(self.microscope.get_stigmator()[1])
+
+        # Check if all the stigmator values are the same for each run
+        self.assertEqual(len(unique_stigmator_values_x), 1,
+                         "'apply_autostigamtor' does not provide the same stigmator values each time in x direction.")
+        self.assertEqual(len(unique_stigmator_values_y), 1,
+                         "'apply_autostigamtor' does not provide the same stigmator values each time in y direction.")
+
```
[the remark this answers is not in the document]
Kleijwegt: No they need to be up to 14 decimals the same.
Do you want to include some extra margin? An if yes how much?
With the current check if these values are exactly the same it works.

I found that set check up to 14 decimals by using:
foo = set()
i=30
while len(foo)<2:
     foo.add(1 + 10**-i)
     i = i - 1# Counting down from 30 decimals up to the point an extra element is added to the set.
print("Precision of set is %s" % i)

src/odemis/driver/xt_client.py
```diff
@@ -997,6 +1042,26 @@ def __init__(self, name, role, parent, **kwargs):
         self._pos_poll = util.RepeatingTimer(5, self._refreshPosition, "Position polling")
         self._pos_poll.start()
 
+    @isasync
+    def apply_autofocus(self, channel_name, state):
+        """
+        Wrapper for running the autofocus functionality asynchronously. It sets the state of autofocus,
+        beam must be turned on. This call is non-blocking.
```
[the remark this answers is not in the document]
Kleijwegt: Is this ok?
```
Also a a reasonable manual focus is needed. When the image is to far out of focus not correct focus can be found using the autofocus functionality.
```

src/odemis/driver/xt_client.py
```diff
@@ -676,6 +680,47 @@ def __init__(self, name, role, parent, hfw_nomag, **kwargs):
         self._va_poll = util.RepeatingTimer(5, self._updateSettings, "Settings polling")
         self._va_poll.start()
 
+    @isasync
+    def apply_autostigmator(self, channel_name, state):
```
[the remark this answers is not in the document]
Kleijwegt: Also, as a minor note, please follow the naming convention for the functions in the API, it should be spelled applyAutoStigmator.
> Fixed this

src/odemis/driver/xt_client.py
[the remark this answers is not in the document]
tepals: Yes with minor tweaks:

`Also a reasonable manual focus is needed. When the image is too far out of focus, an incorrect focus can be found using the autofocus functionality.`

src/odemis/driver/test/xt_client_test.py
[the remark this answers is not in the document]
tepals: I wonder if exactly the same value is always found for autostigmation, if you would test on the hardware. But I think this is fine for now.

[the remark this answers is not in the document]
pieleric: There is no way that you're going to run 5 times the auto stigmator on a real hardware and get twice the same value. Let alone 5 times the same value! I guess you get always the same value because you're using a simulator. 
In general, I don't think we need to test that the stigmator behaviour is good anyway. This is up to TFS, if their auto stigmator is good, then it's nice, if their stigmator doesn't always work, it's a pity.... but we can't do much about it, so we don't need to care  in the test case.
So, here, I'd suggest to just check that the values are roughly the same (eg, 20% -> `almost_equal(...,rtol=0.2)`)

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,160 @@ def test_set_beam_shift(self):
...
+            stigmator_values_x.append(self.microscope.get_stigmator()[0])
+            stigmator_values_y.append(self.microscope.get_stigmator()[1])
+
+        # Check if all the stigmator values are within the required range each run
+        stigmator_range = self.microscope.stigmator_info()['range']
+        min_x_stig_values = stigmator_range['x'][0]
+        max_x_stig_values = stigmator_range['x'][1]
+        min_y_stig_values = stigmator_range['y'][0]
+        max_y_stig_values = stigmator_range['y'][1]
```
tepals, line 340: You could shorten these 4 lines:

```
min_x_stig_values, max_x_stig_values = stigmator_range['x']
min_y_stig_values, max_y_stig_values = stigmator_range['y']
```

src/odemis/driver/xt_client.py
```diff
@@ -676,6 +683,114 @@ def __init__(self, name, role, parent, hfw_nomag, **kwargs):
         self._va_poll = util.RepeatingTimer(5, self._updateSettings, "Settings polling")
         self._va_poll.start()
 
+    @isasync
+    def applyAutoStigmator(self, detector):
```
tepals, line 687: shouldn't this function also be commented?

src/odemis/driver/xt_client.py
```diff
@@ -676,6 +683,114 @@ def __init__(self, name, role, parent, hfw_nomag, **kwargs):
...
+        Starts applying auto contrast brightness and checks if the process is finished for the ProgressiveFuture object.
+        :param future (Future): the future to stop. Unused, only one future must be running at a time.
+        """
+        channel_name = future.channel_name
+        with future._auto_contrast_brighness_lock_lock:
+            if future._must_stop.is_set():
+                raise CancelledError()
+            self.parent.set_auto_contrast_brightness(channel_name, XT_RUN)
+            time.sleep(0.5)  # Wait for the autofocussing to start
+        # Wait until the microscope is no longer autofocussing
```
tepals, line 774: this comment is in a weird spot

src/odemis/driver/xt_client.py
```diff
@@ -676,6 +683,114 @@ def __init__(self, name, role, parent, hfw_nomag, **kwargs):
...
+
+        :param channel_name (str): Name of one of the electron channels, the channel must be running.
+        :return: Future object
+
+        """
+        # Create ProgressiveFuture and update its state
+        est_start = time.time() + 0.1
+        f = ProgressiveFuture(start=est_start,
+                              end=est_start + 20)  # rough time estimation
+        f._auto_contrast_brighness_lock_lock = threading.Lock()
```
tepals, line 757: is there a reason why this is named `...._lock_lock`?

src/odemis/driver/xt_client.py
```diff
@@ -997,6 +1112,58 @@ def __init__(self, name, role, parent, **kwargs):
...
 
+    @isasync
+    def applyAutofocus(self, detector):
+        """
+        Wrapper for running the autofocus functionality asynchronously. It sets the state of autofocus,
+        the beam must be turned on unblanked. Also a a reasonable manual focus is needed. When the image is to far
+        out of focus, an incorrect focus can be found using the autofocus functionality.
+        This call is non-blocking.
+
+        :param channel_name (str): Name of one of the electron channels, the channel must be running.
```
tepals, line 1123: update this docstring to detector

src/odemis/driver/xt_client.py
```diff
@@ -676,6 +683,114 @@ def __init__(self, name, role, parent, hfw_nomag, **kwargs):
...
+            return True
+
+    @isasync
+    def applyAutoContrastBrightness(self, detector):
+        """
+        Wrapper for running the automatic setting of the contrast brightness functionality asynchronously. It sets
+        the state of auto_contrast_brightness, the beam must be turned on unblanked. Auto contrast brightness works
+        best if there is a feature visible in the image. This call is non-blocking.
+
+        :param channel_name (str): Name of one of the electron channels, the channel must be running.
```
tepals, line 749: update this docstring to detector

tepals (commented):
No major comments from me

Sabrina85 (approved):
The latest changes (commit 7) look good to me.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,158 @@ def test_set_beam_shift(self):
...
+        # Test if an error is raised if an invalid detector is provided
+        with self.assertRaises(KeyError):
+            self.scanner.applyAutoStigmator("error_expected")
+        time.sleep(2.5)  # Give microscope/simulator the time to update the state
+        autostigmator_state = self.microscope.is_autostigmating(detector2ChannelName[detector])
+        self.assertEqual(autostigmator_state, False)  # Check if state remained unchanged
+
+    def test_apply_auto_contrast_brightness(self):
+        """
+        Test for the auto functionality of the apply_auto_contrast_brightness.
```
Sabrina85, line 361: Same as for auto stigmation.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,158 @@ def test_set_beam_shift(self):
...
+        Only checks the returned state, the return object and if an error is raised for an incorrect input state.
+        """
+        for detector in detector2ChannelName:
+            auto_contrast_brightness_return = self.scanner.applyAutoContrastBrightness(detector)
+            time.sleep(2.5)  # Give microscope/simulator the time to update the state
+            auto_contrast_brightness_state = self.microscope.is_running_auto_contrast_brightness(detector2ChannelName[detector])
```
Sabrina85, line 367: Add comments the same way as for the auto stigmation test.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,158 @@ def test_set_beam_shift(self):
...
+            self.assertEqual(auto_contrast_brightness_state, True)
+            self.assertIsInstance(auto_contrast_brightness_return, ProgressiveFuture)
+
```
Sabrina85, line 370: same here: add comment

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,158 @@ def test_set_beam_shift(self):
...
+            auto_contrast_brightness_return.cancel()
+            time.sleep(5.0)  # Give microscope/simulator the time to update the state
+            auto_contrast_brightness_state = self.microscope.is_running_auto_contrast_brightness(detector2ChannelName[detector])
+            self.assertEqual(auto_contrast_brightness_state, False)
+            self.assertIsInstance(auto_contrast_brightness_return, ProgressiveFuture)
+
```
Sabrina85, line 376: Add comment what you are testing in this block.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,158 @@ def test_set_beam_shift(self):
...
+            self.scanner.applyAutoContrastBrightness(detector)
+            starting_time = time.time()
+            time.sleep(0.5)  # Give microscope/simulator the time to update the state
+            while self.microscope.is_running_auto_contrast_brightness(detector2ChannelName[detector]):
+                time.sleep(0.2)  # Wait until the autofocus is finished
+                if time.time() - starting_time > 30:
+                    break
+            # Check if the time to perform the autofocus is not to long
```
Sabrina85, line 384: to -> too (This is a mistake I see very often. Please look up how to use "to" and "too").

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,158 @@ def test_set_beam_shift(self):
...
+            self.assertLess(time.time() - starting_time, 30, "Execution auto-contrast brightness is slow.")
+            # Line to inspect the execution time to update the expected time
+            print("Execution time was %s seconds " % (time.time() - starting_time))
+            auto_contrast_brightness_state = self.microscope.is_running_auto_contrast_brightness(detector2ChannelName[detector])
+            self.assertEqual(auto_contrast_brightness_state, False)
+
+        # Test if an error is raised if an invalid detector is provided
```
Sabrina85, line 391: missing: "or"

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,158 @@ def test_set_beam_shift(self):
...
+            while self.microscope.is_running_auto_contrast_brightness(detector2ChannelName[detector]):
+                time.sleep(0.2)  # Wait until the autofocus is finished
+                if time.time() - starting_time > 30:
+                    break
+            # Check if the time to perform the autofocus is not to long
+            self.assertLess(time.time() - starting_time, 30, "Execution auto-contrast brightness is slow.")
+            # Line to inspect the execution time to update the expected time
+            print("Execution time was %s seconds " % (time.time() - starting_time))
+            auto_contrast_brightness_state = self.microscope.is_running_auto_contrast_brightness(detector2ChannelName[detector])
+            self.assertEqual(auto_contrast_brightness_state, False)
```
Sabrina85, line 389: I don't understand this test. What is the definition of "too long"? Should the auto contrast brightness procedure be cancelled/aborted if it is "too long"?

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,158 @@ def test_set_beam_shift(self):
...
+
+        # Test if an error is raised if an invalid detector is provided
+        with self.assertRaises(KeyError):
```
Sabrina85, line 392: Add comment what this block tests.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,158 @@ def test_set_beam_shift(self):
...
+            self.scanner.applyAutoContrastBrightness("error_expected")
+        time.sleep(2.5)  # Give microscope/simulator the time to update the state
+        autofocus_state = self.microscope.is_autofocusing(detector2ChannelName[detector])
+        self.assertEqual(autofocus_state, False)
+
+    def test_apply_autofocus(self):
+        """
+        Test for the auto functionality of the autofocus.
+        Only checks the returned state, the return object and if an error is raised for an incorrect input state.
```
Sabrina85, line 401: Same here as above.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,158 @@ def test_set_beam_shift(self):
...
+        with self.assertRaises(KeyError):
+            self.scanner.applyAutoStigmator("error_expected")
+        time.sleep(2.5)  # Give microscope/simulator the time to update the state
+        autostigmator_state = self.microscope.is_autostigmating(detector2ChannelName[detector])
+        self.assertEqual(autostigmator_state, False)  # Check if state remained unchanged
+
+    def test_apply_auto_contrast_brightness(self):
+        """
+        Test for the auto functionality of the apply_auto_contrast_brightness.
+        Only checks the returned state, the return object and if an error is raised for an incorrect input state.
```
Sabrina85, line 362: I think it is ok to skip this sentence.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,158 @@ def test_set_beam_shift(self):
         self.scanner.beamShift.value = (new_beam_shift_x, new_beam_shift_y)
         self.assertAlmostEqual((new_beam_shift_x, new_beam_shift_y), self.scanner.beamShift.value)
 
+    @unittest.skip("Currently this functionality is not supported by the XT client and the code is commented out")
+    def test_apply_autostigmator(self):
+        """
+        Test for the auto functionality of the apply_autostigmator.
+        Runs the apply_autostigmator multiple times and checks (using "get_stigmator()) if the found stigmator values
```
Sabrina85, line 290: You can skip these 2 sentences.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,158 @@ def test_set_beam_shift(self):
...
+        are equal each run. Checks the returned state, the return object and if an error is raised for an incorrect input state.
+        """
+        detector = 'se-detector'  # Test autostigmator only for se-detector, channel=electron1
```
Sabrina85, line 293: Why do you not use the detector from `detector2ChannelName` same way you do for auto focusing?

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,158 @@ def test_set_beam_shift(self):
...
+        time.sleep(2.5)  # Give microscope/simulator the time to update the state
+        autofocus_state = self.microscope.is_autofocusing(detector2ChannelName[detector])
+        self.assertEqual(autofocus_state, False)
+
+    def test_apply_autofocus(self):
+        """
+        Test for the auto functionality of the autofocus.
+        Only checks the returned state, the return object and if an error is raised for an incorrect input state.
+        """
+        # Check all the different detector types by looping over them
```
Sabrina85, line 403: Add this comment also in the auto brightness contrast test.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,158 @@ def test_set_beam_shift(self):
...
+        for detector in detector2ChannelName:
+            autofocus_return = self.efocus.applyAutofocus(detector)
```
Sabrina85, line 405: Add comment what this block tests.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,158 @@ def test_set_beam_shift(self):
...
+            time.sleep(2.5)  # Give microscope/simulator the time to update the state
+            autofocus_state = self.microscope.is_autofocusing(detector2ChannelName[detector])
+            self.assertEqual(autofocus_state, True)
+            self.assertIsInstance(autofocus_return, ProgressiveFuture)
+
```
Sabrina85, line 410: Add comment what this block tests.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,158 @@ def test_set_beam_shift(self):
...
+            autofocus_return = self.efocus.applyAutofocus(detector)
+            starting_time = time.time()
+            time.sleep(0.5)  # Give microscope/simulator the time to update the state
+            while self.microscope.is_autofocusing(detector2ChannelName[detector]):
+                time.sleep(0.2)  # Wait until the autofocus is finished
+                if time.time() - starting_time > 40:
+                    break
+            # Check if the time to perform the autofocus is not to long
```
Sabrina85, line 424: to -> too

Sabrina85: See my comment on auto BC regarding definition of "too long" and if you abort the procedure.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,158 @@ def test_set_beam_shift(self):
...
+            self.assertLess(time.time() - starting_time, 40)
+            # Line to inspect the execution time to update the expected time
+            print("Execution time was %s seconds " % (time.time() - starting_time))
+            autofocus_state = self.microscope.is_autofocusing(detector2ChannelName[detector])
+            self.assertEqual(autofocus_state, False)
+
+        # Test if an error is raised if an invalid detector is provided
```
Sabrina85, line 431: if -> when

Sabrina85: Add this comment the same way to the other 2 tests.

Sabrina85 (changes requested):
I looked at the complete code. What I noticed is, that the test cases are very inconsistent in respect to each other. They are very similar, so you can basically use the same comments for all tests. Now sometimes you add a comment what you test for and in another test at exactly the same block you don't add it.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,158 @@ def test_set_beam_shift(self):
...
+        autostigmator_state = self.microscope.is_autostigmating(detector2ChannelName[detector])
+        self.assertEqual(autostigmator_state, False)
+
+        ## Test if autostigmating gives repeatable results and check time it takes to auto stigmate.
+
+        # Create sets holding stigmator values to check if the outcome is consistent
+        stigmator_values_x = []  # List holding only the stigmator values in x direction
+        stigmator_values_y = []  # List holding only the stigmator values in y direction
+
+        # Run apply_autostigmator multiple times and afterwards check if the outcome is the same for each run
```
Sabrina85, line 312: Can you actually be sure that you will find exactly the same values whenever you run auto stigmation? Maybe check this with Wilco or Job.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,158 @@ def test_set_beam_shift(self):
...
+            self.assertIsInstance(autostigmator_return, ProgressiveFuture)
+
+            starting_time = time.time()
+            time.sleep(0.5)  # Give microscope/simulator the time to update the state
+            while self.microscope.is_autostigmating(detector2ChannelName[detector]):
+                time.sleep(0.2)  # Wait until the autofocus is finished
+                if time.time() - starting_time > 30:
+                    break
+
+            # Check if the time to perform the autofocus is not to long
```
Sabrina85, line 327: to -> too

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,158 @@ def test_set_beam_shift(self):
...
+
+        autostigmator_return.cancel()
+        time.sleep(5.0)  # Give microscope/simulator the time to update the state
+        autostigmator_state = self.microscope.is_autostigmating(detector2ChannelName[detector])
+        self.assertEqual(autostigmator_state, False)
+
+        ## Test if autostigmating gives repeatable results and check time it takes to auto stigmate.
+
+        # Create sets holding stigmator values to check if the outcome is consistent
+        stigmator_values_x = []  # List holding only the stigmator values in x direction
```
Sabrina85, line 309: -> `List of stigmator values in x direction`

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,158 @@ def test_set_beam_shift(self):
...
+        self.assertIsInstance(autostigmator_return, ProgressiveFuture)
+
+        autostigmator_return.cancel()
+        time.sleep(5.0)  # Give microscope/simulator the time to update the state
+        autostigmator_state = self.microscope.is_autostigmating(detector2ChannelName[detector])
+        self.assertEqual(autostigmator_state, False)
+
+        ## Test if autostigmating gives repeatable results and check time it takes to auto stigmate.
+
+        # Create sets holding stigmator values to check if the outcome is consistent
```
Sabrina85, line 308: Create sets of stigmator values to check if the results are reproducible.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,158 @@ def test_set_beam_shift(self):
...
+        # Check if all the stigmator values are within the required range each run
+        stigmator_range = self.microscope.stigmator_info()['range']
+        min_x_stig_values, max_x_stig_values = stigmator_range['x']
+        min_y_stig_values, max_y_stig_values = stigmator_range['y']
+        self.assertGreaterEqual(min(stigmator_values_x), min_x_stig_values)
+        self.assertLessEqual(max(stigmator_values_x), max_x_stig_values)
+        self.assertGreaterEqual(min(stigmator_values_y), min_y_stig_values)
+        self.assertLessEqual(max(stigmator_values_y), max_y_stig_values)
+
+        # TODO K.K. remove debug values after testing
```
Sabrina85, line 344: I don't think that should be merged.

src/odemis/driver/xt_client.py
```diff
@@ -676,6 +683,114 @@ def __init__(self, name, role, parent, hfw_nomag, **kwargs):
...
+    #         return True
+
+    @isasync
+    def applyAutoContrastBrightness(self, detector):
+        """
+        Wrapper for running the automatic setting of the contrast brightness functionality asynchronously. It sets
+        the state of auto_contrast_brightness, the beam must be turned on unblanked. Auto contrast brightness works
+        best if there is a feature visible in the image. This call is non-blocking.
+
+        :param detector (str): Name of one of the detector.
```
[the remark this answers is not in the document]
Kleijwegt: Actually I think this is not plural, it sets for one detector the autofunctionality settings at a time.

src/odemis/driver/test/xt_client_test.py
Kleijwegt: Because per definition this functionality will never be available on other detectors. In theory if other detectors are added, those could be present on other detectors.

src/odemis/driver/test/xt_client_test.py
Kleijwegt: This test is currenlty not used because stigmator values cannot be applied yet. But indeed it is clean to remove this.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,158 @@ def test_set_beam_shift(self):
...
+        print("X list is")
+        print(stigmator_values_x)
+        print("Y list is")
+        print(stigmator_values_y)
+        print("With a range of:")
+        print(self.microscope.stigmator_info()['range'])
+
+        # Test if an error is raised if an invalid detector is provided
```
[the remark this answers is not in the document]
Kleijwegt: Is this ok?
> Test if an error is raised when an invalid detector name is provided

src/odemis/driver/test/xt_client_test.py
Kleijwegt: Changed this to:
 > Test if an error is raised when an invalid detector is provided
 Is that clearer?

src/odemis/driver/test/xt_client_test.py
Kleijwegt: That is what the comment above says.

src/odemis/driver/test/xt_client_test.py
Kleijwegt: The stigmation test if slightly different (it first tests basic functionality and then after it repeats the test and compaires the outcome)
If I look at the other tests in this test file I don't see this extensive commenting. And to me the line "auto_contrast_brightness_future = self.scanner.applyAutoContrastBrightness(detector)" suggests that the "AutoContrastBrightness" function is applied and some lines down that:  "auto_contrast_brightness_future.cancel()" this cancels it.
But maybe that is just me not liking extra comments what do you guys think @Sabrina85  and @pieleric ?

src/odemis/driver/test/xt_client_test.py
Kleijwegt: If just a single time of autofunctionality takes to long we don't want that A. we don't have test cases getting stuck B. have the test case fail because autofunctionality should not take 3 times longer than usually, most likely something is wrong with than functionality.

src/odemis/driver/xt_client.py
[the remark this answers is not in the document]
Sabrina85: If you say "one of something" you pick one thing out of a collection. The collection would need to be detectors (plural). If it is just one detector you may write simply: `Name of the detector.`

src/odemis/driver/test/xt_client_test.py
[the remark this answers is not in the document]
Sabrina85: yes :)

src/odemis/driver/test/xt_client_test.py
Sabrina85: Ok makes sense and it is important info. Maybe make your comment even more explicit saying that autostigmation only works with the se-detector.

src/odemis/driver/test/xt_client_test.py
Sabrina85: Indeed - my bad.

src/odemis/driver/test/xt_client_test.py
Sabrina85: I asked this for consistency. If you do it in one test, do it also in the other ones. The reader easily can follow the structure then. The reader will also immediately notice the blocks that are different in the tests. Alternatively, don't do it in any test and only comment the "special part" in the autostigmator.

src/odemis/driver/test/xt_client_test.py
Sabrina85: That does not answer how you define "too long". Do you have a feeling what the typical time auto BC should take is and how did you decide that "30" is too long? 
When auto BC was taking longer than 30 [unit?] then you will raise an error. The error message just says that auto BC is slow. But actually the test failed and we do not know was it slow and didn't finish or was there something going wrong.  
I suggest to first add when you decide stop waiting for auto BC to finish. And second say in the assertion message that the auto BC exceeded that limit or did not finish in that limit.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -324,7 +325,7 @@ def test_apply_autostigmator(self):
                 if time.time() - starting_time > 30:
                     break
 
-            # Check if the time to perform the autofocus is not to long
+            # Check if the time too perform the autofocus is not to long
```
Sabrina85, line 328: autofocus -> autostigmation
**to** perform .... **too** long.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -286,36 +286,37 @@ def test_set_beam_shift(self):
...
         """
-        Test for the auto functionality of the apply_autostigmator.
-        Runs the apply_autostigmator multiple times and checks (using "get_stigmator()) if the found stigmator values
-        are equal each run. Checks the returned state, the return object and if an error is raised for an incorrect input state.
+        Test for the auto stigmation functionality.
         """
         detector = 'se-detector'  # Test autostigmator only for se-detector, channel=electron1
-        ## Basic testing of the functionality of the wrapper: can it be started and canceled
-        autostigmator_return = self.scanner.applyAutoStigmator(detector)
+        ## Start with basic check on starting and stopping the auto stigmation functionality
```
Sabrina85, line 292: This comment can be skipped as this is now split in between line 293 and 300.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -286,36 +286,37 @@ def test_set_beam_shift(self):
...
 
-        # Create sets holding stigmator values to check if the outcome is consistent
-        stigmator_values_x = []  # List holding only the stigmator values in x direction
-        stigmator_values_y = []  # List holding only the stigmator values in y direction
+        # Create lists holding stigmator values to check if the outcome is within the allowed range.
+        stigmator_values_x = []  # List of stigmator values in x direction
+        stigmator_values_y = []  # List of stigmator values in y direction
 
-        # Run apply_autostigmator multiple times and afterwards check if the outcome is the same for each run
+        # Run apply_autostigmator multiple times and afterwards check if the outcome is within the allowed range
```
Sabrina85, line 313: I don't understand this test. I see in your code, that your run auto stigmation 5 times. You collect the values set on the stigmator HW after auto stigmation was finished. Then you check that these values are in the range of the allowed values for the stigmator HW. 
This should never be a problem I think. The values that you get from the stigmator are already set. So I would assume that if somehow a value is out of range, the autostigmation would already fail internally in xt and you will never reach the point to retrieve a value from the stigmator that is out of range.
Do I miss something?

src/odemis/driver/test/xt_client_test.py
```diff
@@ -358,21 +351,21 @@ def test_apply_autostigmator(self):
 
     def test_apply_auto_contrast_brightness(self):
         """
-        Test for the auto functionality of the apply_auto_contrast_brightness.
-        Only checks the returned state, the return object and if an error is raised for an incorrect input state.
+        Test for the auto contrast brightness functionlaity.
```
Sabrina85, line 354: Spelling mistake in "functionality".

Sabrina85 (changes requested):

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,150 @@ def test_set_beam_shift(self):
...
+            autostigmator_future = self.scanner.applyAutoStigmator(detector)
+            time.sleep(0.5)  # Give microscope/simulator the time to update the state
+            autostigmator_state = self.microscope.is_autostigmating(detector2ChannelName[detector])
+            self.assertEqual(autostigmator_state, True)
+            self.assertIsInstance(autostigmator_future, ProgressiveFuture)
+
+            starting_time = time.time()
+            time.sleep(0.5)  # Give microscope/simulator the time to update the state
+            while self.microscope.is_autostigmating(detector2ChannelName[detector]):
+                time.sleep(0.2)  # Wait until the autofocus is finished
```
Sabrina85, line 324: autofocus -> autostigmation

src/odemis/driver/test/xt_client_test.py
Kleijwegt: Actually this test isn't perfec, because the whole functionality is not implemented on the XT side.
Ideally we wanted to see if the function would give reproducible values from autostigmating which are more or less in the same range. But that test failed because the autostigmating didn't work. Thera suggested to at least check if the range information was given right and the correct values where returned. But as you say that is a really basic check which on second thought is not even needed to do. Note that currently the entire test cannot be runned because the autostigmation call doesn't work at all.
We can keep the list in there to allow easy implementation of a more elaborate test case when autostigmation works again. Do you think we should replace the assertLess/GreaterEqual statments with just a TODO?

src/odemis/driver/test/xt_client_test.py
Kleijwegt: I see and indeed I agree, lets discuss briefly during the STUP.

src/odemis/driver/test/xt_client_test.py
Sabrina85: As discussed in the STUP:

- Either remove the check if "value set on stigmator is in range that is allowed by the stigmator HW" or if you leave it put a TODO to check if this can ever happen as you suggested.

- Add a TODO that running the autostigmation procedure should theoretically result in similar values set on the HW, but that the accuracy for this needs to be defined.

src/odemis/driver/test/xt_client_test.py
Kleijwegt: I think the current implementation is what you suggested, but please check if this is ok

src/odemis/driver/test/xt_client_test.py
```diff
@@ -351,33 +352,38 @@ def test_apply_autostigmator(self):
...
+            max_execution_time = 30
             self.scanner.applyAutoContrastBrightness(detector)
             starting_time = time.time()
             time.sleep(0.5)  # Give microscope/simulator the time to update the state
             while self.microscope.is_running_auto_contrast_brightness(detector2ChannelName[detector]):
                 time.sleep(0.2)  # Wait until the autofocus is finished
-                if time.time() - starting_time > 30:
+                if time.time() - starting_time > max_execution_time:
                     break
             # Check if the time to perform the autofocus is not too long
```
Sabrina85, line 382: autofocus -> auto BC

src/odemis/driver/test/xt_client_test.py
```diff
@@ -394,29 +400,34 @@ def test_apply_autofocus(self):
...
 
+            # Stop auto focus and check if it stopped running.
             autofocus_future.cancel()
             time.sleep(5.0)  # Give microscope/simulator the time to update the state
             autofocus_state = self.microscope.is_autofocusing(detector2ChannelName[detector])
             self.assertEqual(autofocus_state, False)
             self.assertIsInstance(autofocus_future, ProgressiveFuture)
 
+            # Start autofocus
+            max_execution_time = 40
```
Sabrina85, line 418: Same here as in auto BC.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -394,29 +400,34 @@ def test_apply_autofocus(self):
...
                 time.sleep(0.2)  # Wait until the autofocus is finished
-                if time.time() - starting_time > 40:
+                if time.time() - starting_time > max_execution_time:
                     break
             # Check if the time to perform the autofocus is not too long
-            self.assertLess(time.time() - starting_time, 40)
-            # Line to inspect the execution time to update the expected time
-            print("Execution time was %s seconds " % (time.time() - starting_time))
+            self.assertLess(time.time() - starting_time, max_execution_time,
+                            "Execution auto-contrast brightness was stopped because it took more than %s seconds."
```
Sabrina85, line 428: auto-BC -> autofocus?

Sabrina85 (changes requested):
Small changes only. Please be more thorough when doing copy/paste between test cases. I still found mix ups between for example autofocus and auto-BC in the messages etc.

src/odemis/driver/test/xt_client_test.py
Sabrina85: Indeed my mistake. Your implementation is correct!

Sabrina85 (approved):
Looks good!

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,162 @@ def test_set_beam_shift(self):
...
+        time.sleep(2.5)  # Give microscope/simulator the time to update the state
+        autofocus_state = self.microscope.is_autofocusing(detector2ChannelName[detector])
+        self.assertEqual(autofocus_state, False)
+
+    def test_apply_autofocus(self):
+        """
+        Test for the auto functionality of the autofocus.
+        """
+        # Check all the different detector types by looping over them
+        for detector in detector2ChannelName:
```
pieleric, line 402: Let's write more something like: `for role, channel in detector2ChannelName.items()`. Then you can drop the use of `detector2ChannelName[detector]`.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,162 @@ def test_set_beam_shift(self):
...
+
+            # Stop auto focus and check if it stopped running.
+            autofocus_future.cancel()
+            time.sleep(5.0)  # Give microscope/simulator the time to update the state
+            autofocus_state = self.microscope.is_autofocusing(detector2ChannelName[detector])
+            self.assertEqual(autofocus_state, False)
+            self.assertIsInstance(autofocus_future, ProgressiveFuture)
+
+            # Start autofocus
+            max_execution_time = 40    # Approximately 3 times the normal expected execution time
```
pieleric, line 418: Please indicate the unit in the comment (ie, "s").

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,162 @@ def test_set_beam_shift(self):
...
+            starting_time = time.time()
+            autofocus_future = self.efocus.applyAutofocus(detector)
+            time.sleep(0.5)  # Give microscope/simulator the time to update the state
+            while self.microscope.is_autofocusing(detector2ChannelName[detector]):
+                time.sleep(0.2)  # Wait until the autofocus is finished
+                if time.time() - starting_time > max_execution_time:
+                    break  # If autofocussing takes too long stop waiting
```
pieleric, line 425: Please only use the standard API. I think you can do the same with such code:
``` python
autofocus_future.result(timeout=max_execution_time)
```
This will raise an error on Timeout or if another issue happened.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,162 @@ def test_set_beam_shift(self):
...
+            self.assertEqual(autofocus_state, True)
+            self.assertIsInstance(autofocus_future, ProgressiveFuture)
+
+            # Stop auto focus and check if it stopped running.
+            autofocus_future.cancel()
+            time.sleep(5.0)  # Give microscope/simulator the time to update the state
+            autofocus_state = self.microscope.is_autofocusing(detector2ChannelName[detector])
+            self.assertEqual(autofocus_state, False)
+            self.assertIsInstance(autofocus_future, ProgressiveFuture)
+
```
pieleric, line 416: Please add another test where the future is cancelled immediately after starting the autofocus. It should also work.

pieleric (changes requested):
Looks alright. My main recommendation is to limit as much as possible the usage of the internal API in the test cases. Please try to only use the API provided by the component under test. See my comments on the autofocus test, they are also valid for the other test-cases.

I have a couple of small extra minor comments for improving the readability of the code.

src/odemis/driver/xt_client.py
```diff
@@ -676,6 +683,115 @@ def __init__(self, name, role, parent, hfw_nomag, **kwargs):
...
+            if future._must_stop.is_set():
+                raise CancelledError()
+
+    def _cancelAutoContrastBrightness(self, future):
+        """
+        Cancels the autofocussing. Non-blocking.
+        :param future (Future): the future to stop.
+        :return (bool): True if it successfully cancelled (stopped) the move.
+        """
+        logging.debug("Cancelling current move")
```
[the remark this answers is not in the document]
Kleijwegt: I should have changed that message indeed.
Do you think the logging message is moved to the correct place now?

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,171 @@ def test_set_beam_shift(self):
...
+        """
+        # Check all the different detector types by looping over them
+        for role, channel in DETECTOR2CHANNELNAME.items():
+            # Start auto contrast brightness and check if it is running.
+            auto_contrast_brightness_future = self.scanner.applyAutoContrastBrightness(role)
+            auto_contrast_brightness_state = self.microscope.is_running_auto_contrast_brightness(channel)
+            self.assertEqual(auto_contrast_brightness_state, True)
+            self.assertIsInstance(auto_contrast_brightness_future, ProgressiveFuture)
+            # Give microscope/simulator to properly start before canceling, to prevent an error.
+            time.sleep(2.5)
```
Kleijwegt, line 362: Without sleeping the error:
`Error
  File "/usr/lib/python3.6/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/usr/lib/python3.6/unittest/case.py", line 605, in run
    testMethod()
  File "/home/sonic/development/odemis/src/odemis/driver/test/xt_client_test.py", line 421, in test_apply_auto_contrast_brightness
    auto_contrast_brightness_future.cancel()
  File "/home/sonic/development/odemis/src/odemis/model/_futures.py", line 311, in cancel
    if not (canceller and canceller(self)):
  File "/home/sonic/development/odemis/src/odemis/driver/xt_client.py", line 982, in _cancelAutoContrastBrightness
    self.parent.set_auto_contrast_brightness(future._channel_name, XT_STOP)
  File "/home/sonic/development/odemis/src/odemis/driver/xt_client.py", line 516, in set_auto_contrast_brightness
    self.server.set_auto_contrast_brightness(name, state)
  File "/home/sonic/.local/lib/python3.6/site-packages/Pyro5/client.py", line 476, in __call__
    return self.__send(self.__name, args, kwargs)
  File "/home/sonic/.local/lib/python3.6/site-packages/Pyro5/client.py", line 243, in _pyroInvoke
    raise data  # if you see this in your traceback, you should probably inspect the remote traceback as well
OSError: [Errno -2147219405] Set auto contrast brightness state failed. Error message: None`

Is raised, having this waiting time here fixing this.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,171 @@ def test_set_beam_shift(self):
...
+
+    def test_apply_autofocus(self):
+        """
+        Test for the auto functionality of the autofocus.
+        """
+        # Check all the different detector types by looping over them
+        for role, channel in DETECTOR2CHANNELNAME.items():
+            # Start auto focus and check if it is running.
+            autofocus_future = self.efocus.applyAutofocus(role)
+            autofocus_state = self.microscope.is_autofocusing(channel)
```
pieleric, line 410: Wait a tiny bit before checking.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,171 @@ def test_set_beam_shift(self):
...
+        max_execution_time = 30  # Approximately 3 times the normal expected execution time (s)
+        for i in range(0, 5):
+            starting_time = time.time()
+            autostigmator_future = self.scanner.applyAutoStigmator(detector)
+            time.sleep(0.5)  # Give microscope/simulator the time to update the state
+            autostigmator_state = self.microscope.is_autostigmating(channel)
+            self.assertEqual(autostigmator_state, True)
+            self.assertIsInstance(autostigmator_future, ProgressiveFuture)
+
+            time.sleep(0.5)  # Give microscope/simulator the time to update the state
```
pieleric, line 323: Drop this sleep, you've already just slept before.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,171 @@ def test_set_beam_shift(self):
...
+
+            stigmator_values_x.append(self.microscope.get_stigmator()[0])
+            stigmator_values_y.append(self.microscope.get_stigmator()[1])
+
+        # Check if all the stigmator values are within the required range each run
+        stigmator_range = self.microscope.stigmator_info()['range']
+        min_x_stig_values, max_x_stig_values = stigmator_range['x']
+        min_y_stig_values, max_y_stig_values = stigmator_range['y']
+        # TODO When auto stigmation works implement a check on the reproducibility of the auto stigmation values and
+        #  define the range with which they may differ from each other. In theory similar values should be returned.
```
pieleric, line 341: No need for that. When auto stigmation works, this driver will allow to call it. No need to add tests to check that it works well and reliably, because if it doesn't, there is no change in Odemis that will fix it anyway.

pieleric (changes requested):
Yes, the test cases look better. There are still a few things that could be improved/clarified.

src/odemis/driver/test/xt_client_test.py
Kleijwegt: If already in focus or with a correct brighness/contrast these methods can run really fast. For auto contrast brigthness this can be < 1 sec for autofocus this can be up to around 1 sec. We can implement a wait but we then have the chance of a false fail of the test because of the wait. Currently the status update is fast enough and the tests will pas.
What do you think I should do?

src/odemis/driver/xt_client.py
```diff
@@ -676,6 +684,131 @@ def __init__(self, name, role, parent, hfw_nomag, **kwargs):
...
+                    time.sleep(1.0)  # Give microscope/simulator time to update the state
+                    # Check if auto contrast brightness stopped
+                    if not self.parent.is_running_auto_contrast_brightness(future._channel_name):
+                        return False
+                    else:
+                        # Auto contrast brightness is not stopped correctly, then do so.
+                        time.sleep(2.0)  # Wait a bit so we can properly cancel auto contrast brightness.
+                        # Stop auto contrast brightness properly
+                        self.parent.set_auto_contrast_brightness(future._channel_name, XT_STOP)
+                        return True
```
Kleijwegt, line 808: @pieleric Maybe this part is a bit over the top since we don't expect this to happen. But maybe because of threading the auto functionality is not started yet and it is already tried to cancel it. Then this error is raised but the functionality is not actually stopped. Very much a corner case and we could replace these lines with just "raise" but I don't know what you think is the nicest solution?

src/odemis/driver/test/xt_client_test.py
```diff
@@ -281,6 +283,154 @@ def test_set_beam_shift(self):
...
+
+    def test_apply_auto_contrast_brightness(self):
+        """
+        Test for the auto contrast brightness functionality.
+        """
+        # Check all the different detector types by looping over them
+        for role, channel in DETECTOR2CHANNELNAME.items():
+            # Start auto contrast brightness and check if it is running.
+            auto_contrast_brightness_future = self.scanner.applyAutoContrastBrightness(role)
+            auto_contrast_brightness_state = self.microscope.is_running_auto_contrast_brightness(channel)
```
pieleric, line 343: Add a tiny sleep here too.

pieleric (changes requested):
I think you can keep the special except block in the cancel simple.
Also, let's have some very tiny sleeps in the test cases, and hopefully that will ensure the test cases always pass.

src/odemis/driver/test/xt_client_test.py
Kleijwegt: That works indeed.

pieleric (approved):
Looks good to me :-)

PR #1481, opened 2020-08-19
Added clarifying comments to FindGridSpots and related functions.

tepals:
Added comments and updated docstrings, I hope this clarifies things.

### DIFF
--- a/src/odemis/acq/align/spot.py
+++ b/src/odemis/acq/align/spot.py
@@ -39,7 +39,6 @@ from scipy.spatial import cKDTree as KDTree
 import threading
 import time
 
-
 ROUGH_MOVE = 1  # Number of max steps to reach the center in rough move
 FINE_MOVE = 10  # Number of max steps to reach the center in fine move
 FOV_MARGIN = 250  # pixels
@@ -345,7 +344,9 @@ def FindSpot(image, sensitivity_limit=100):
 
 def FindGridSpots(image, repetition):
     """
-    Find a grid of spots in an image.
+    Find the coordinates of a grid of spots in an image. And find the
+    corresponding transformation to transform a grid centered around the origin
+    to the spots in an image.
 
     Parameters
     ----------
@@ -357,22 +358,36 @@ def FindGridSpots(image, repetition):
     Returns
     -------
     spot_coordinates : array like
-        A 2D array of shape (N, 2) containing the coordinates of the spots.
+        A 2D array of shape (N, 2) containing the coordinates of the spots,
+        in respect to the top left of the image.
     translation : tuple of two floats
+        Translation from the origin to the center of the grid in image space,
+        origin is top left of the image. Primary axis points down and the
+        secondary axis points to the right.
     scaling : tuple of two floats
+        Scaling factors for primary and secondary axis.
     rotation : float
+        Rotation in image space, positive rotation is clockwise.
     shear : float
+        Horizontal shear factor. A positive shear factor transforms a coordinate
+        in the positive x direction parallel to the x axis.
 
     """
+    # Find the center coordinates of the spots in the image.
     spot_positions = MaximaFind(image, repetition[0] * repetition[1])
     if len(spot_positions) < repetition[0] * repetition[1]:
         logging.warning('Not enough spots found, returning only the found spots.')
         return spot_positions, None, None, None
-    # Estimate transformation
+    # Estimate the two most common directions in the grid of spots, these form the transformation matrix.
     lattice_constants = EstimateLatticeConstant(spot_positions)
+    # Transpose to put the primary axis on the first row instead of the first column. In a transformation matrix, the
+    # first row should represent the transform for the primary axis, second row for the secondary, etc..
     transformation_matrix = numpy.transpose(lattice_constants)
     if numpy.linalg.det(lattice_constants) < 0.:
+        # If the determinant is smaller than zero flip the transformation matrix to make it a right handed
+        # transformation matrix.
         transformation_matrix = numpy.fliplr(transformation_matrix)
+    # Translation is the mean of the spots, which is the distance from the origin to the center of the grid of spots.
     translation = numpy.mean(spot_positions, axis=0)
     transform_to_spot_positions = AffineTransform(matrix=transformation_matrix, translation=translation)
     # Iterative closest point algorithm - single iteration, to fit a grid to the found spot positions
@@ -380,8 +395,9 @@ def FindGridSpots(image, repetition):
     spot_grid = transform_to_spot_positions(grid)
     tree = KDTree(spot_positions)
     dd, ii = tree.query(spot_grid, k=1)
-
+    # Sort the original spot positions to know which spot is which.
     pos_sorted = spot_positions[ii.ravel(), :]
+    # Find the transformation from a grid centered around the origin to the sorted positions.
     transformation = AffineTransform.from_pointset(grid, pos_sorted)
     spot_coordinates = transformation(grid)
     return spot_coordinates, translation, transformation.scale, transformation.rotation, transformation.shear
--- a/src/odemis/acq/align/spot.py
+++ b/src/odemis/acq/align/spot.py
@@ -378,15 +378,12 @@ def FindGridSpots(image, repetition):
     if len(spot_positions) < repetition[0] * repetition[1]:
         logging.warning('Not enough spots found, returning only the found spots.')
         return spot_positions, None, None, None
-    # Estimate the two most common directions in the grid of spots, these form the transformation matrix.
+    # Estimate the two most common (orthogonal) directions in the grid of spots, defined in the image coordinate system.
     lattice_constants = EstimateLatticeConstant(spot_positions)
-    # Transpose to put the primary axis on the first row instead of the first column. In a transformation matrix, the
-    # first row should represent the transform for the primary axis, second row for the secondary, etc..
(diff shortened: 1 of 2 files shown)


src/odemis/acq/align/spot.py
```diff
@@ -357,31 +358,44 @@ def FindGridSpots(image, repetition):
...
         logging.warning('Not enough spots found, returning only the found spots.')
         return spot_positions, None, None, None
-    # Estimate transformation
+    # Estimate the two most common directions in the grid of spots, these form the transformation matrix.
     lattice_constants = EstimateLatticeConstant(spot_positions)
+    # Transpose to put the primary axis on the first row instead of the first column. In a transformation matrix, the
+    # first row should represent the transform for the primary axis, second row for the secondary, etc..
     transformation_matrix = numpy.transpose(lattice_constants)
     if numpy.linalg.det(lattice_constants) < 0.:
         transformation_matrix = numpy.fliplr(transformation_matrix)
```
tepals, line 387: this still needs a comment

src/odemis/acq/align/spot.py
```diff
@@ -357,31 +358,46 @@ def FindGridSpots(image, repetition):
...
         return spot_positions, None, None, None
-    # Estimate transformation
+    # Estimate the two most common directions in the grid of spots, these form the transformation matrix.
     lattice_constants = EstimateLatticeConstant(spot_positions)
+    # Transpose to put the primary axis on the first row instead of the first column. In a transformation matrix, the
+    # first row should represent the transform for the primary axis, second row for the secondary, etc..
     transformation_matrix = numpy.transpose(lattice_constants)
     if numpy.linalg.det(lattice_constants) < 0.:
+        # If the determinant is smaller than zero flip the transformation matrix to make it a right handed
+        # transformation matrix.
```
pieleric, line 388: You don't really need to say "If the determinant is smaller than zero" because that's what the code just above says. Either don't mention it, or "translate" it into something more high-level (eg, "A negative determinant means that the first axis goes blabla compared to the second axis"). I'm not convinced that saying "right handed transformation matrix" is really clear. Maybe you can say that "we assume that the image is not inverted and not too rotated, so the horizontal axis should be the first axis (X) and the vertical axis the second axis (Y). I'm not sure that's true though... I leave it up to you ;-)

pieleric (changes requested):
That's definitely helping. I only have one comment (about one comment).

src/odemis/acq/align/spot.py
tepals: I will think of something better! 

I also wonder if it makes sense to move this if statement inside `EstimateLatticeConstant`, then `fliplr` should be changed to `flipud`. Moving it inside `EstimateLatticeConstant` will make the output consistent. Now the output of `EstimateLatticeConstant` depends on the "strongest" mean, found by the kmeans algorithm.

tepals: @Sabrina85 what do you think of moving `fliplr` inside `EstimateLatticeConstant`? That way the output of `EstimateLatticeConstant` has the horizontal axis first and vertical axis second, corresponding to the axes of `spot_positions`; the input of `EstimateLatticeConstant`.

src/odemis/acq/align/spot.py
```diff
@@ -357,31 +358,46 @@ def FindGridSpots(image, repetition):
...
+    # Translation is the mean of the spots, which is the distance from the origin to the center of the grid of spots.
     translation = numpy.mean(spot_positions, axis=0)
     transform_to_spot_positions = AffineTransform(matrix=transformation_matrix, translation=translation)
     # Iterative closest point algorithm - single iteration, to fit a grid to the found spot positions
     grid = GridPoints(*repetition)
     spot_grid = transform_to_spot_positions(grid)
     tree = KDTree(spot_positions)
     dd, ii = tree.query(spot_grid, k=1)
-
+    # Sort the original spot positions to know which spot is which.
```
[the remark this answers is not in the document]
tepals: The spot positions are mapped to the positions of the grid, based on the iterative closest point algorithm. So, after sorting, the order of the spot positions corresponds to the order of the points in the grid calculated with `GridPoints`. I will clarify this comment.

src/odemis/acq/align/spot.py
```diff
@@ -357,31 +358,46 @@ def FindGridSpots(image, repetition):
...
+        A 2D array of shape (N, 2) containing the coordinates of the spots,
+        in respect to the top left of the image.
     translation : tuple of two floats
+        Translation from the origin to the center of the grid in image space,
+        origin is top left of the image. Primary axis points down and the
+        secondary axis points to the right.
     scaling : tuple of two floats
+        Scaling factors for primary and secondary axis.
     rotation : float
+        Rotation in image space, positive rotation is clockwise.
```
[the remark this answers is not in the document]
tepals: After thinking about it for a bit longer, I think it does make sense the primary axis is horizontal and the secondary is vertical. Which is a left handed system, so positive rotation is clockwise. I will update the explanation of translation to:
`Primary axis to the right and the axis points down.`

src/odemis/acq/align/spot.py
```diff
@@ -357,31 +358,46 @@ def FindGridSpots(image, repetition):
...
+    # Find the center coordinates of the spots in the image.
     spot_positions = MaximaFind(image, repetition[0] * repetition[1])
     if len(spot_positions) < repetition[0] * repetition[1]:
         logging.warning('Not enough spots found, returning only the found spots.')
         return spot_positions, None, None, None
-    # Estimate transformation
+    # Estimate the two most common directions in the grid of spots, these form the transformation matrix.
     lattice_constants = EstimateLatticeConstant(spot_positions)
+    # Transpose to put the primary axis on the first row instead of the first column. In a transformation matrix, the
+    # first row should represent the transform for the primary axis, second row for the secondary, etc..
```
Sabrina85, line 384: That is another way of saying it as in line 388?
`        # If the determinant is smaller than zero flip the transformation matrix to make it a right handed
        # transformation matrix.`
If yes maybe you could add here that is a RH transformation matrix then and in line 388 you could add the description with the primary and secondary axis.

src/odemis/acq/align/spot.py
Sabrina85: That makes sense! What about the `transformation_matrix = numpy.transpose(lattice_constants)`? That would need to be moved as well as it does the transform as described in the comment?

src/odemis/acq/align/spot.py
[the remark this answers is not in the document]
Sabrina85: There is no absolute spot position information right? If the pattern is flipped by 180 degrees we wouldn't know.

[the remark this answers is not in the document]
tepals: That is correct.

src/odemis/acq/align/spot.py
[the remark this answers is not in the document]
Sabrina85: Did that also solve your example?

src/odemis/acq/align/spot.py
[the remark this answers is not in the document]
Sabrina85: Maybe that is also worse to mention.

src/odemis/acq/align/spot.py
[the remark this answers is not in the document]
tepals: yes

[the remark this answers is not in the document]
Sabrina85: ok :)

src/odemis/acq/align/spot.py
```diff
@@ -357,31 +358,43 @@ def FindGridSpots(image, repetition):
...
     spot_positions = MaximaFind(image, repetition[0] * repetition[1])
     if len(spot_positions) < repetition[0] * repetition[1]:
         logging.warning('Not enough spots found, returning only the found spots.')
         return spot_positions, None, None, None
-    # Estimate transformation
+    # Estimate the two most common (orthogonal) directions in the grid of spots, defined in the image coordinate system.
     lattice_constants = EstimateLatticeConstant(spot_positions)
+    # The two directions form the transformation matrix. Transpose because we want to describe a grid in the coordinate
+    # system of the lattice, in the image coordinate system.
     transformation_matrix = numpy.transpose(lattice_constants)
```
tepals, line 385: This step is still super funky to me. I will ask Andries if this comment is correct.

src/odemis/util/spot.py
```diff
@@ -116,7 +116,7 @@ def SpotIntensity(data, background=None):
               min(int(center[1]), data0.shape[0] - 3))
     # Take the 3x3 image around the center
     neighborhood = data0[(center[1] - 1):(center[1] + 2),
-                         (center[0] - 1):(center[0] + 2)]
+                   (center[0] - 1):(center[0] + 2)]
```
Sabrina85, line 119: Wasn't this indentation correct before?

tepals: yes, will fix it.

tepals:
I talked to Andries and updated the comment. @Sabrina85 can you have one last look?

Sabrina85 (approved):
Looks good to me!
At this point you cannot connect the vectors in the columns to primary and secondary axes, can you?

tepals:
> Looks good to me!
> At this point you cannot connect the vectors in the columns to primary and secondary axes, can you?

No, you cannot know if for instance the top left spot corresponds to the first beam and therefore you don't know the orientation of the grid.

tepals:
@pieleric if you're fine with the PR you can merge it.

PR #1484, opened 2020-08-21
gui util img: Fix calculate_ticks() sometimes doing infinite loop

pieleric:


src/odemis/gui/util/img.py
```diff
@@ -860,24 +860,28 @@ def calculate_ticks(value_range, client_size, orientation, tick_spacing):
     vtp_ratio = abs(pixel_space / value_space)  # px/unit
     epsilon = (1 / vtp_ratio) / 10  # "tiny" value: a 10th of a pixel
 
-    num_ticks = pixel_space // tick_spacing
+    # Find the ticks to show so that it looks good for the user. It must have a
+    # constant spacing between the ticks, and this spacing should be a "easy
+    # number": 1, 2, 5, 7.5, or any of these values multiplied by a power of 10.
```
Kleijwegt, line 865: Why is 7.5 easy and 1.5 or 2.5 not easy? What does easy mean?

src/odemis/gui/util/img.py
```diff
@@ -860,24 +860,28 @@ def calculate_ticks(value_range, client_size, orientation, tick_spacing):
...
+    # For example, 150 is not good, but 200 is fine.
+
+    # minimum number of ticks that we want
+    num_ticks = max(2, pixel_space // tick_spacing + 1)
```
Kleijwegt, line 869: Is there a testcase where you test for 'uneasy' values? in case the coding changes in the future and someone messes up this might be handy.

src/odemis/gui/util/img.py
```diff
@@ -860,24 +860,28 @@ def calculate_ticks(value_range, client_size, orientation, tick_spacing):
...
 
-    while value_step and abs(value_space)/value_step > num_ticks:
-        # round down to a power of 10
-        power = int(math.log10(value_step))
-        for step in spacing_values:
-            value_step = step * 10 ** power
-            if abs(value_space) / value_step < num_ticks:
-                break
-    logging.debug("Value step is %s with range %s", value_step, value_space)
+    for step in [1, 2, 5, 7.5, 10]:
```
Kleijwegt, line 877: Maybe it is handy to define these 'easy' numbers as a global variable so that a testcase can use the same.
Btw 10 equals 1 * 10 so I don't think you need to also include that number in the list.

src/odemis/gui/util/img.py
Kleijwegt, line 865: Why is 7.5  "easy" and 2.5 not? Is 250 also not fine? or 300?
And what does "easy" mean? Why is it easy?
Can you maybe add what easy means? Probably it has something to do with the layout?

Kleijwegt (changes requested):

src/odemis/gui/util/img.py
pieleric: To be honest, I'm not exactly sure on what basis these values where picked, but they were defined by Toon as the simplest distances to have in a graph. I guess the idea is to have only a few distances possible, and they should divide 10. Indeed, 7.5 is a little odd, but I think that jumping from 5 to 10 brings a whole set of graphs with too many ticks. So it's the less bad option. "Easy" here means "easy to the eyes of someone who has a PhD in Physics" ;-) It might sound a little exclusive, but I think that actually defines most of our users ;-) Honestly... it's very arguable... but I'd rather not change it now.

src/odemis/gui/util/img.py
pieleric: 10 is necessary. That was actually one of the errors in the previous code. As we round down to the lower power of 10 first, if the value is above 7.5, it wouldn't find anything and loop infinitely (and now, it wouldn't loop infinitely, but find 7.5). With 10, we ensure that any value above 7.5 returns 10, which is a nice multiple of 10 :-)

src/odemis/gui/util/img.py
pieleric: The test case "test_simple" would trigger the infinite loop bug. What do you mean by "uneasy" values? Maybe the name of the test case is not so well chosen, but it contains quite a few different inputs, which could trigger different issues if not handled properly.

src/odemis/gui/util/img.py
VictoriaDel: Indeed, Toon picked those values. Initially, this code was implemented to start with spacing 10 (multiplied by a power of 10). This spacing would be divided by 2 until the optimal spacing is reached and consequently the final values for spacing would be one of the following: 1.25, 2.5, 5, 10 (multiplied by a power of 10), which are not the desirable ones. I think it's not a big issue, we could again ask for his feedback and change them anytime as far as they remain hardcoded.

src/odemis/gui/util/img.py
pieleric: Yes, in the previous version, the code kept increasing the `power` (with a while loop) until the value was sufficiently big. So 10 was equivalent to 1, with just the next power. However, I've removed this while loop, as it's safer (makes it impossible to loop infinitely), and it's possible to compute immediately the right `power` anyway. Now, 10 is necessary.

PR #1487, opened 2020-08-27
Add specific error and todo for diagonally shifted tiles in registrar

Kleijwegt:


src/odemis/acq/stitching/_registrar.py
```diff
@@ -580,6 +581,11 @@ def _insert_tile_to_grid(self, tile):
                     self.tiles[i].append(None)
                     self.shifts_hor[i].append(None)
                     self.shifts_ver[i].append(None)
+        elif almost_equal(abs(ver_diff), abs(hor_diff)) and abs(hor_diff) > 0 and abs(ver_diff) > 0:
```
philipwinklerdelmic, line 584: I don't think almost_equal will work here. `ver_diff` and `hor_diff` might still be a few pixels apart in this case, so you want to test whether the magnitude is approximately the same. I think that's a bit difficult to do and might be more confusing than helpful.

The docstring of `.addTile` explicitly says that diagonal moves are not supported, so we probably don't need to check it here.

Kleijwegt: I increased the tolerance to 1%. I don't actually think it would be such a big problem because in any other case when ver_diff != hor_diff the code would go into the other if statements.

pieleric:
I'm curious on what kind of acquisition does diagonal acquisition? Is that some kind of hexagonal pattern? Or is it just to catch when by mistake the caller forgets to pass a tile?

src/odemis/acq/stitching/_registrar.py
```diff
@@ -580,6 +581,11 @@ def _insert_tile_to_grid(self, tile):
                     self.tiles[i].append(None)
                     self.shifts_hor[i].append(None)
                     self.shifts_ver[i].append(None)
+        elif almost_equal(abs(ver_diff), abs(hor_diff), atol=0, rtol=0.01) and abs(hor_diff) > 0 and abs(ver_diff) > 0:
+            # TODO: Also implement registrar for tiles which are purely diagonally shifted. Currently this option is
+            #  not implemented.
+            raise NotImplementedError("Cannot insert tiles which have a pure diagonal shift, this functionality is "
+                                      "not yet implemented.")
         else:
             raise ValueError("Cannot insert multiple tiles at the same position.")
```
Kleijwegt, line 590: Result discussion       

 elif abs(ver_diff) ```
== 0 and abs(hor_diff) == 0:
            raise ValueError("Cannot insert multiple tiles at the same position.")
        elif ver_diff == hor_diff:
            # this only happens when ver_diff and ho_diff are exactly equal and not 0, so
            #   the shift is perfectly diagonal
            # Inpratice would only happen in a test case situation.
            raise NotImplementeIerror()
        else:
            # this should never happen
            raise ValueError
```

File: src/odemis/acq/stitching/_registrar.py [hunk 3]
```diff
@@ -580,8 +581,18 @@ def _insert_tile_to_grid(self, tile):
                     self.tiles[i].append(None)
                     self.shifts_hor[i].append(None)
                     self.shifts_ver[i].append(None)
-        else:
+        elif abs(hor_diff) == 0 and abs(ver_diff) == 0:
             raise ValueError("Cannot insert multiple tiles at the same position.")
+        elif abs(ver_diff) == abs(hor_diff):
+            # Pure diagonal shift. Via the previous if statements is implied that hor_diff and ver_diff are not zero.
```
philipwinklerdelmic, line 587: Looks good. Maybe we can make the comments a bit clearer. For example here, I would put something like
```# Purely diagonal shift. In practice, this will never happen because the smallest difference will be picked up by one of the previous if-statements. This exception is likely to be triggered only in artificial tests.```

TODO is not needed in my opinion. The only way to deal with it I can think of is to add a small value to one of the shifts, so they are not exactly equal anymore. However, we probably prefer to get an error in this case (especially since this registrar is not supposed to handle diagonal moves anyway).

**Review by philipwinklerdelmic — APPROVED** [review]

File: src/odemis/acq/stitching/_registrar.py [hunk 4]
```diff
@@ -580,8 +581,19 @@ def _insert_tile_to_grid(self, tile):
...
+        elif abs(hor_diff) == 0 and abs(ver_diff) == 0:
             raise ValueError("Cannot insert multiple tiles at the same position.")
+        elif abs(ver_diff) == abs(hor_diff):
+            # Purely diagonal shift. In practice, this will never happen because the smallest difference will be
+            # picked up by one of the previous if-statements. This exception is likely to be triggered only in
+            # artificial tests.
+            # TODO: Implement registrar for tiles which are purely diagonally shifted.
+            raise NotImplementedError("Cannot insert tiles which have a pure diagonal shift, this functionality is "
+                                      "not yet implemented. This error statement is only expected in test cases "
+                                      "since the horizontal shift needs to be exactly equal to the vertical shift.")
```
pieleric, line 593: If that only happens in some test cases, I'd suggest to fix the test cases. If for some reason you really need that, then make the error a ValueError, with a simpler message like "Horizontal and vertical shift show pure diagonal shift which is a sign of missing tile".
Also, I don't see any reason to implement code for such artificial corner case. => remove the TODO.

Kleijwegt:
reaction to @pieleric
> I'm curious on what kind of acquisition does diagonal acquisition? Is that some kind of hexagonal pattern? Or is it just to catch when by mistake the caller forgets to pass a tile?

I encountered it during the stage to multiprobe calibration for sonic. Here we want to take two images which are shifted in X direction (but maybe by accident move also a lot in Y direction). For the testcases I extended this into the extreme where a diagonal shift was present. I got an unexpected and wrong error, hence the fixed by a new more specific error.
Although probably this error won't be raised in practice it is better and cleaner to have an error pointing to the right problem instead of sending you into the wrong direction.
In case we do want to also register the shift for diagonally shifted images, for calibration purposes for example, we need to fix this problem but for now that won't be any time soon.

src/odemis/acq/stitching/_registrar.py
```diff
@@ -580,8 +581,19 @@ def _insert_tile_to_grid(self, tile):
...
+        else:
+            # This should never happen
+            raise ValueError("Unexpected error.")
```
pieleric, line 596: For now this code cannot be reached. In case the code above changes, this could be still handy to have, but make the error more descriptive. Add the ver_diff and hor_diff values in the message.

pieleric (changes requested):
I'm now understanding better what you're trying to fix. Adding a test for hor_diff=ver_diff=0 did help. Just a few extra comment on the error messages.

src/odemis/acq/stitching/_registrar.py
Kleijwegt: I fixed the changes and force pushed on the previous commit. Because it is only a couple of lines I don't think this would make the review process a lot harder plus this way it can be merged directly without merging the commits before finishing the PR

src/odemis/acq/stitching/_registrar.py
```diff
@@ -33,6 +33,7 @@
 import numpy
 import math
 from odemis import model
+from odemis.util import almost_equal
```
pieleric, line 36: You can drop this import, and this should be fine to merge :-)

pieleric (changes requested):

src/odemis/acq/stitching/_registrar.py
Kleijwegt: Done

src/odemis/acq/stitching/_registrar.py
Sabrina85: Note: When a pr in github is merged, there is a selection, where commits are automatically squashed before the pr is merged. See "squash and merge".

Sabrina85:
> reaction to @pieleric
> 
> > I'm curious on what kind of acquisition does diagonal acquisition? Is that some kind of hexagonal pattern? Or is it just to catch when by mistake the caller forgets to pass a tile?
> 
> I encountered it during the stage to multiprobe calibration for sonic. Here we want to take two images which are shifted in X direction (but maybe by accident move also a lot in Y direction). For the testcases I extended this into the extreme where a diagonal shift was present. I got an unexpected and wrong error, hence the fixed by a new more specific error.
> Although probably this error won't be raised in practice it is better and cleaner to have an error pointing to the right problem instead of sending you into the wrong direction.
> In case we do want to also register the shift for diagonally shifted images, for calibration purposes for example, we need to fix this problem but for now that won't be any time soon.

How can it happen that the image is shifted by accident into the y direction? That sounds strange. As we do the stage to multiprobe alignment I can only imagine that you are already somehow accidentally in the other coordinate system where the move in x direction is actually a move in x and y. Did you check that there is not my some chance a mistake in the code of the alignment script where the error occurred?

PR #1490, opened 2020-09-01
model component: forbid overriding VAs on a Proxy

pieleric:
When writing Python code, due to the weird syntax of VAs, it's easy to
override a VA instead of actually setting a value. This should hopefully
catch a few bugs a bit earlier.

We could also do a similar behaviour on a Component, but there might be
some specific cases where this is acceptable. Maybe we could acccept
overriding a VA only with another VA... but for now I don't dare trying.

### DIFF
--- a/src/odemis/model/_components.py
+++ b/src/odemis/model/_components.py
@@ -247,6 +247,15 @@ class ComponentProxy(ComponentBase, Pyro4.Proxy):
         _vattributes.load_vigilant_attributes(self, vas)
         _dataflow.load_events(self, events)
 
+    def __setattr__(self, name, value):
+        # Detect that the user is trying to replace a VigilantAttribute, which is
+        # most likely a typo of forgetting VA.value .
+        # On a Component, it's fishy, but on a Proxy, there is just no reason to
+        # do that.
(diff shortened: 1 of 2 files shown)


src/odemis/model/_components.py
```diff
@@ -247,6 +247,15 @@ def __setstate__(self, state):
         _vattributes.load_vigilant_attributes(self, vas)
         _dataflow.load_events(self, events)
 
+    def __setattr__(self, name, value):
+        # Detect that the user is trying to replace a VigilantAttribute, which is
+        # most likely a typo of forgetting VA.value .
+        # On a Component, it's fishy, but on a Proxy, there is just no reason to
+        # do that.
+        if hasVA(self, name):
+            raise AttributeError("Cannot override VigilantAttribute")
```
tepals, line 256: Is it maybe useful here to add the name you're trying to set, in this AttributeError?

pieleric: Indeed. I somehow assumed that from the backtrace it'd be possible to find out which VA is overridden. However, it's really simple and simpler to be more explicit.

PR #1492, opened 2020-09-08
CryoSECOM lens aligner active position

lazem:
The goal is to ensure that the optical lens is positioned as close as possible from the last known aligned position after a referencing (or a move to a safe position during 5DoF referencing).

### DIFF
--- a/src/odemis/gui/cont/tabs.py
+++ b/src/odemis/gui/cont/tabs.py
@@ -2588,8 +2588,11 @@ class SecomAlignTab(Tab):
             # as we expect no acquisition active when changing tab, it will always
             # lead to subscriptions to VA
             main_data.is_acquiring.subscribe(self._on_acquisition, init=True)
+            # Subscribe to lens aligner movement to update its FAV_POS_ACTIVE metadata
+            self._aligner_xy.position.subscribe(self._onAlignPos)
         else:
             main_data.is_acquiring.unsubscribe(self._on_acquisition)
+            self._aligner_xy.position.unsubscribe(self._onAlignPos)
 
     def terminate(self):
         super(SecomAlignTab, self).terminate()
@@ -2825,9 +2828,27 @@ class SecomAlignTab(Tab):
         best_mpp = self._sem_view.mpp.clip(best_mpp)
         self._sem_view.mpp.value = best_mpp
 
+    def _onAlignPos(self, pos):
+        """
+        Called when the aligner is moved (and the tab is shown)
+        :param pos: (dict str->float) updated position of the aligner
+        """
+        # Return if aligner doesn't have MD_FAV_POS_ACTIVE and MD_FAV_POS_DEACTIVE metadata
+        md = self._aligner_xy.getMetadata()
+        if (model.MD_FAV_POS_ACTIVE and model.MD_FAV_POS_DEACTIVE) not in md:
+            return
+        # Check if updated position is close to FAV_POS_DEACTIVE
+        dist_deactive = math.hypot(pos["x"] - md[model.MD_FAV_POS_DEACTIVE]["x"],
+                                   pos["y"] - md[model.MD_FAV_POS_DEACTIVE]["y"])
+        if dist_deactive <= 0.1e-3:  # within 0.1mm of deactive is considered deactivated.
+            logging.warning("Aligner seems parked, not updating FAV_POS_ACTIVE")
+            return
+        # Save aligner position as the "calibrated" one
+        self._aligner_xy.updateMetadata({model.MD_FAV_POS_ACTIVE: pos})
+
     @classmethod
     def get_display_priority(cls, main_data):
-        if main_data.role == "secom":
+        if main_data.role in ("secom", 'cryo-secom'):
             return 1
         else:
             return None
--- a/src/odemis/gui/cont/tabs.py
+++ b/src/odemis/gui/cont/tabs.py
@@ -2544,6 +2544,8 @@ class SecomAlignTab(Tab):
 
         self.tab_data_model.tool.subscribe(self._onTool, init=True)
         main_data.chamberState.subscribe(self.on_chamber_state, init=True)
+        # To check that aligner move to FAV_POS_ACTIVE position is initiated
+        self._aligner_activated = False
 
     def _on_ccd_should_update(self, update):
         """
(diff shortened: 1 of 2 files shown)


lazem:
Hi @pieleric I added the part of updating the `FAV_POS_ACTIVE` of the lens aligner. Appreciating your review. 
Thanks!

src/odemis/gui/cont/tabs.py
```diff
@@ -2825,9 +2829,28 @@ def _onSEMpxs(self, pixel_size):
...
+        Called when the aligner is moved (and the tab is shown)
+        :param pos: (dict str->float) updated position of the aligner
+        """
+        # Return if aligner doesn't have MD_FAV_POS_ACTIVE and MD_FAV_POS_DEACTIVE metadata
+        md = self._aligner_xy.getMetadata()
+        if model.MD_FAV_POS_ACTIVE not in md and model.MD_FAV_POS_DEACTIVE in md:
+            return
+        # Check if updated position is close to FAV_POS_DEACTIVE
+        for axis in self._aligner_xy.axes:
+            if not _isNearDeactive(pos, md[model.MD_FAV_POS_DEACTIVE], axis):
```
pieleric, line 2843: Please don't use private methods.
I'd actually not recommend using this method anyway.
You could just write a check like this:
``` python
dist_deactive = math.hypot(pos["x"] - MD_FAV_POS_DEACTIVE["x"],
               pos["y"] - MD_FAV_POS_DEACTIVE["y"])
if dist_deactive <= 0.1e-3: # within 0.1mm of deactive is considered deactivated.
       ...warning...
      return
```

pieleric (changes requested):
For this part of the feature, it looks pretty fine. Just a few comments.

However, there is a second part in the specification to automatically move to the ACTIVE position whenever the localization or alignment tab is selected. Will you implement it in the same PR, or later? A little tricky part is that if the aligner moves on tab showing, it might update the position while moving (at a regular interval). These "transition positions" from DEACTIVE to ACTIVE shouldn't be used to update the metadata. So it'll be best to only subscribe to the position only at the end of this first move.
Another thing to pay attention is that (as defined in the specification) this move shouldn't be done if the aligner is not referenced yet. In such case, it might be referencing and so will finish soon. In that case, you could listen to the position update to detect the referencing is complete and the the move to ACTIVE should be initiated.

lazem:
> For this part of the feature, it looks pretty fine. Just a few comments.
> 
> However, there is a second part in the specification to automatically move to the ACTIVE position whenever the localization or alignment tab is selected. Will you implement it in the same PR, or later? A little tricky part is that if the aligner moves on tab showing, it might update the position while moving (at a regular interval). These "transition positions" from DEACTIVE to ACTIVE shouldn't be used to update the metadata. So it'll be best to only subscribe to the position only at the end of this first move.
> Another thing to pay attention is that (as defined in the specification) this move shouldn't be done if the aligner is not referenced yet. In such case, it might be referencing and so will finish soon. In that case, you could listen to the position update to detect the referencing is complete and the the move to ACTIVE should be initiated.

Thanks for the review @pieleric I guess I will continue the second part in this same repo. I implemented the active move in the alignment tab with the precautions you outlined. Would appreciate to have a look.
Thanks

src/odemis/gui/cont/tabs.py
```diff
@@ -2831,12 +2836,19 @@ def _onSEMpxs(self, pixel_size):
...
         md = self._aligner_xy.getMetadata()
         if (model.MD_FAV_POS_ACTIVE and model.MD_FAV_POS_DEACTIVE) not in md:
             return
+        # Move aligner on tab showing to FAV_POS_ACTIVE position (if all axes are referenced)
+        if not self._aligner_activated:
+            if all(self._aligner_xy.referenced.value.values()):
+                self._aligner_xy.moveAbs(md[model.MD_FAV_POS_ACTIVE]).result()  # Wait till movement done
+                self._aligner_activated = True
+            return
+
```
pieleric, line 2851: Firstly, I don't see why this code should be part of this function. This function is to update the metadata when the aligner is moved. Let's only do that. I think that moving the aligner should be handled by a separate function.

Also, from the experience I recently got from the SPARC alignment tab, I'd recommend the following:
* On show(), if there is any metadata and it's referenced, start the aligner move, and disable the X/Y buttons. Add a callback for the end of the move.
* At the end of the move, enable the X/Y buttons and subscribe the position to _onAlignPos(), if the tab is still shown.

(You need to do it in a callback, because you don't want the GUI to block while the aligner is moving)

You can have a look at odemis.gui.cont.tabs.Sparc2AlignTab, code related to "lens_mover".
You should be able to get rid of `_aligner_activated`.

src/odemis/gui/cont/tabs.py
```diff
@@ -2825,9 +2833,34 @@ def _onSEMpxs(self, pixel_size):
...
+        if dist_deactive <= 0.1e-3:  # within 0.1mm of deactive is considered deactivated.
+            logging.warning("Aligner seems parked, not updating FAV_POS_ACTIVE")
+            return
+        # Save aligner position as the "calibrated" one
+        self._aligner_xy.updateMetadata({model.MD_FAV_POS_ACTIVE: pos})
+
     @classmethod
     def get_display_priority(cls, main_data):
-        if main_data.role == "secom":
+        if main_data.role in ("secom", 'cryo-secom'):
```
pieleric, line 2863: Please use double-quotes. I usually only have a small preference for them, but here, I'd be quite picky, because the other string already uses that.

src/odemis/gui/cont/tabs.py
```diff
@@ -2825,9 +2833,34 @@ def _onSEMpxs(self, pixel_size):
         best_mpp = self._sem_view.mpp.clip(best_mpp)
         self._sem_view.mpp.value = best_mpp
 
+    def _onAlignPos(self, pos):
+        """
+        Called when the aligner is moved (and the tab is shown)
+        :param pos: (dict str->float or None) updated position of the aligner
+        """
+        # Return if aligner doesn't have both MD_FAV_POS_ACTIVE and MD_FAV_POS_DEACTIVE metadata
```
pieleric, line 2841: Please add a check that the tab is currently shown. That's an extra safety, mostly in case the user moved to the tab, and immediately left to another tab...  while later the aligner goes to "active" position, and subscribes to this.

pieleric (changes requested):
I'd recommend to follow the same way as we did in the SPARC2 aligner. Two separate functions, for moving to the aligned position vs updating the metadata. Also the GUI shouldn't block when the hardware moves, as it's often a little slow, and sometimes very slow (eg, broken hardware/connection).

PR #1496, opened 2020-09-09
Lakeshore 335 driver

amuskens:


src/odemis/driver/test/lakeshore_test.py
```diff
@@ -0,0 +1,100 @@
...
+        """
+        Test some simple functions
+        """
+
+        self.assertNotEqual(self.dev.temperature.value, 0)
+
+        temps = range(1, 25)
+        for temp in temps:
+            self.dev.targetTemperature.value = temp
+            self.assertEqual(self.dev.targetTemperature.value, temp)
```
pieleric, line 77: Sounds like dangerous on the real hardware. 1K ?! Also, maybe not needed to try 24 values.... just pick 3 manually.

After the test, revert to the previous value, for safety, if the hardware is in some kind of use.

src/odemis/driver/lakeshore.py
```diff
@@ -0,0 +1,520 @@
...
+        elif re.match(b'RANGE\?', msg):
+            self._sendAnswer(b"%d" % (self._heating,))
+        # set heating range
+        elif re.match(b"RANGE", msg):
+            vals = msg[5:].split(',')
+            self._heating = int(vals[1])
+        # Query temperature
+        elif re.match(b'SRDG\?', msg):
+            self._sendAnswer(b"%f" % (self._temperature,))
+            self._temperature += 1
```
pieleric, line 517: Please change to a random value around the setpoint. Otherwise, in the simulated system, the temperature will get very, confusingly, hot after a couple of minutes.

src/odemis/driver/test/lakeshore_test.py
```diff
@@ -0,0 +1,100 @@
...
+logging.basicConfig(format="%(asctime)s  %(levelname)-7s %(module)s:%(lineno)d %(message)s")
+
+# Export TEST_NOHW=1 to force using only the simulator and skipping test cases
+# needing real hardware
+TEST_NOHW = (os.environ.get("TEST_NOHW", 0) != 0)  # Default to Hw testing
+
+# arguments used for the creation of basic components
+CONFIG = {"name": "Lakeshore Test",
+          "role": "temperature-controller",
+          "port": "/dev/fake",
```
pieleric, line 44: Set to "/dev/ttyUSB*".

src/odemis/driver/lakeshore.py
```diff
@@ -0,0 +1,520 @@
...
+import glob
+from odemis import model
+from odemis.model import HwError
+from odemis.util import to_str_escape
+import os
+import re
+import serial
+import threading
+import time
+from odemis.model._vattributes import FloatVA, IntEnumerated
```
pieleric, line 36: Don't import from the private module, but instead just from `odemis.model`.

src/odemis/driver/lakeshore.py
```diff
@@ -0,0 +1,520 @@
...
+from _ast import Pass
```
pieleric, line 37: Drop.

src/odemis/driver/lakeshore.py
```diff
@@ -0,0 +1,520 @@
...
+        # Clear errors at start
+        try:
+            self.checkError()
+        except LakeshoreError:
+            pass
+
+        manufacturer, md, serial = self.GetIdentifier()
+
+        self._hwVersion = str(md)
+        # self._swVersion = self._version.decode(serial)
```
pieleric, line 101: You can use the same as the other drivers, which just show the linux kernel driver version:
```python
        driver_name = driver.getSerialDriver(self._port)
        self._swVersion = "serial driver: %s" % (driver_name,)
```

pieleric (changes requested):
Overall it looks good. I've got mostly only minor remarks.
Also, could you make sure this is tested on Python 2 & 3?

src/odemis/driver/lakeshore.py
```diff
@@ -0,0 +1,520 @@
...
+                    current_or_power):
+        """
+        Send a heater setup command to the device
+
+        output_type (int): Output type (Output 2 only): 0=Current, 1=Voltage
+        heater_resistance (int): Heater Resistance Setting: 1 = 25 ), 2 = 50 ).
+        max_current (int): Specifies the maximum heater output current: 
+            0 = User Specified, 1 = 0.707 A, 2 = 1 A, 3 = 1.141 A,
+            4 = 1.732 A
+        max_user_current (int): Specifies the maximum heater output current if 
```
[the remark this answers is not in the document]
amuskens: I think it is amps actually. It's a heater after all.

src/odemis/driver/lakeshore.py
```diff
@@ -0,0 +1,520 @@
...
+            pass
+
+        manufacturer, md, serial = self.GetIdentifier()
+
+        self._hwVersion = str(md)
+        # self._swVersion = self._version.decode(serial)
+
+        # Vigilant attributes of the controller
+        self.temperature = FloatVA(unit='K', value=self.GetSensorTemperature(), readonly=True)
+        self.targetTemperature = FloatVA(unit='K', setter=self._set_targetTemperature, value=self.GetSetpoint())
```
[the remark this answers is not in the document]
amuskens: What range should be used?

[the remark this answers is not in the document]
pieleric: I have not much idea. I already know it's bounded between the absolute 0K and 100°C. I guess it's actually a much smaller bound. If you cannot find in the documentation, please ask Daan or Mart about the values they think would fit.

src/odemis/driver/lakeshore.py
```diff
@@ -0,0 +1,510 @@
...
+
+        # Vigilant attributes of the controller.
+        self.temperature = FloatVA(unit=u"°C", value=self.GetSensorTemperature(), readonly=True)
+        self.targetTemperature = FloatContinuous(value=self.GetSetpoint(), unit=u"°C", range=[-273, 50], setter=self._set_targetTemperature)
+        self.heating = IntEnumerated(value=self.GetHeaterRange(), choices={0: "Off", 1: "Low", 2: "Medium", 3: "High"},
+                                     setter=self._set_heating)
+
+
+        self._poll_timer = RepeatingTimer(POLL_INTERVAL, self._poll,
+                                                "Lakeshore Polling update")
```
pieleric, line 120: Please fix the indentation. Probably it would fit on a single line.
BTW, the name of the timer could be "Lakeshore temperature update".

src/odemis/driver/lakeshore.py
```diff
@@ -0,0 +1,510 @@
...
+            if status_byte & err:
+                errors.append(STATUS_BYTE[err])
+
+        if errors:
+            raise LakeshoreError("Lakeshore %s" % (str(errors)), status_byte)
+
+    def GetIdentifier(self):
+        """
+        Get the identifier from the controller
+        Returns 3 strings: manufacturer, model number, and serial number
```
pieleric, line 275: It now returns 4 strings => update docstring

src/odemis/driver/lakeshore.py
```diff
@@ -0,0 +1,510 @@
...
+        Set the temperature setpoint
+        temp (float): the temperature to set, in Celsius
+        """
+        self._sendOrder(b"SETP %d,%.2f" % (self._output_channel, temp + KELVIN_CONVERT))
+
+    def GetSetpoint(self):
+        """
+        Get the temperature setpoint. Returns a float in Celsius
+        """
+        val = self._sendQuery(b"SETP? %d" % (self._output_channel,)).strip('+')
```
pieleric, line 298: Is the "string(+)" really needed? float() accepts + (if there is no space between the + and the number). So maybe you can drop it to keep the code simpler (and the next function too).

src/odemis/driver/lakeshore.py
```diff
@@ -0,0 +1,510 @@
...
+        heater_resistance (int): Heater Resistance Setting: 1 = 25 Ohm, 2 = 50 Ohm.
+        max_current (int): Specifies the maximum heater output current: 
+            0 = User Specified, 1 = 0.707 A, 2 = 1 A, 3 = 1.141 A,
+            4 = 1.732 A
+        max_user_current (int): Specifies the maximum heater output current if 
+            max current is set to User Specified.
+        current_or_power (int): Specifies whether the heater output displays in current or 
+            power (current mode only). Valid entries: 1 = current, 2 = power.
+        """
+        self._sendOrder(b"HTRSET %d,%d,%d,%d,%d,%d" % (self._output_channel,
```
pieleric, line 336: From the doc max_user_current should be converted with `%f`.

src/odemis/driver/lakeshore.py
```diff
@@ -0,0 +1,510 @@
...
+        (output_type,
+            heater_resistance,
+            max_current,
+            max_user_current,
+            current_or_power) = htr_setup.split(',')
+
+        return (int(output_type),
+            int(heater_resistance),
+            int(max_current),
+            int(max_user_current),
```
pieleric, line 359: Float?

src/odemis/driver/lakeshore.py
```diff
@@ -0,0 +1,510 @@
...
+        This method runs in a separate thread and polls the device for the temperature
+        '''
+        try:
+            temp = self.GetSensorTemperature()
+            self.temperature._set_value(temp, force_write=True)
+            logging.debug(u"Lakeshore temperature: %f °C", self.temperature.value)
+            self.checkError()
+        except Exception as e:
+            # another exception. End running.
+            logging.exception(e)
```
pieleric, line 419: The comment says it ends running, but it actually catches the exception and keeps running => update the comment.

Also, `logging.exception()` already logs the exception string, so no need to pass it. Instead, say something like "Failed to read sensor temperature".

src/odemis/driver/lakeshore.py
```diff
@@ -0,0 +1,510 @@
...
+        logging.debug("SIM: parsing %s", to_str_escape(msg))
+        msg = msg.strip()  # remove leading and trailing whitespace
+        msg = b"".join(msg.split())  # remove all space characters
+
+        if msg == b"*STB??":
+            self._sendAnswer(b"%d" % (self._status_byte,))
+        elif msg == b"*CLS":
+            self._status_byte = 0
+        elif msg == b"*IDN?":
+            self._sendAnswer(b"SIM,335,fake")
```
pieleric, line 487: Now the driver expects 4 values => Please make it look more like the real hardware.

src/odemis/driver/lakeshore.py
```diff
@@ -0,0 +1,510 @@
...
+                # Ensure no one will talk to it simultaneously, and we don't talk to devices already in use
+                self._file = open(n)  # Open in RO, just to check for lock
+                try:
+                    fcntl.flock(self._file.fileno(), fcntl.LOCK_EX | fcntl.LOCK_NB)  # Raises IOError if cannot lock
+                except IOError:
+                    logging.info("Port %s is busy, will not use", n)
+                    continue
+
+                self._serial = self._openSerialPort(n)
+                return n
```
pieleric, line 201: Please check that the device answers something that looks fine from "*IDN?". If not, skip it, and continue searching. There can be so many devices connected via a serial port that it's needed. For an example see piezomotor.py/_findDevice()

pieleric (changes requested):
It now looks almost ready to merge. There are a couple of comments which you haven't handled yet, and I've just added a few more. All should be pretty minor.

Please also make sure the simulator still works.

src/odemis/driver/test/lakeshore_test.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def test_heating_test(self):
+        """
+        Test heating
+        """
+        old = self.dev.targetTemperature.value
+        self.dev.targetTemperature.value = self.dev.temperature.value + 1
+        self.dev.heating.value = 3  # fast heating
+        time.sleep(30)
+        self.assertAlmostEqual(self.dev.temperature.value, self.dev.targetTemperature.value)
```
pieleric, line 117: That's not going to work with the simulator, which returns the value +- 5. So either you change the simulator to make it more stable after a while, or increase the temperature change to a larger value (eg +6), and convince the almost equal to have a absolute tolerance of 5°.

src/odemis/driver/lakeshore.py
amuskens: Yes, I think so. Somehow float() didn't like it and it was breaking till I added the strip..

amuskens:
Should be nearly ready to merge!

src/odemis/driver/lakeshore.py
```diff
@@ -198,6 +195,7 @@ def _findDevice(self, ports):
                     continue
 
                 self._serial = self._openSerialPort(n)
+                self.GetIdentifier()  # if value is incorrect, will throw an exception wile unpacking
```
pieleric, line 198: That's no enough, because it's going to raise a TypeError, which you don't handle.
I think that the neatest would be to handle such TypeError inside GetIdentifier (when it's due to no having enough commas in the returned string), and raise IOError in such case.

Also, this is technically just checking that the identifier string returns at least 3 commas, it's not so specific to this device. Please also check the the `manufacter.lower() == "lakeshore"`.

src/odemis/driver/lakeshore.py
```diff
@@ -233,15 +231,15 @@ def _sendQuery(self, cmd):
 
             self._serial.timeout = 1
             ans = b''
-            while ans[-1:] != b'\r':
```
pieleric: BTW, as the device only accepts LF, could you update also the docstring of sendOrder/sendQuery to say "LF" instead of "CR".

src/odemis/driver/lakeshore.py
```diff
@@ -432,13 +430,14 @@ def __init__(self, timeout=1):
         self._input_buf = b""  # what we receive from the "host computer"
         self._status_byte = POWER_ON
 
-        self._setpoint = 10
-        self._temperature = 50
-        self._heating = 1
+        self._setpoint = 150
+        self._temperature = 100
```
pieleric, line 434: Please indicate in comments the unit of these values ("K", right ?).

src/odemis/driver/lakeshore.py
```diff
@@ -432,13 +430,14 @@ def __init__(self, timeout=1):
...
+        self._stable_temperature = 50
```
pieleric, line 435: That's just a constant? Then make it a class constant and use upper case: `STABLE_TEMPERATURE`. Also add a small comment like "K, temperature reached without heating".

src/odemis/driver/lakeshore.py
```diff
@@ -502,9 +501,18 @@ def _parseMessage(self, msg):
             vals = msg[5:].split(',')
             self._heating = int(vals[1])
         # Query temperature
-        elif re.match(b'SRDG\?', msg):
-            self._sendAnswer(b"%f" % (self._temperature,))
-            self._temperature += random.randint(-5, 5)
+        elif re.match(b'KRDG\?', msg):
+            # send temperature with some noise
+            self._sendAnswer(b"+%.3f" % (self._temperature + random.random() / 1000,))
```
pieleric, line 506: `random.random() / 1000` -> `random.uniform(-0.1, 01)`.
(I think that 0.1° is quite small already compared to the hardware precision, so that should be enough).

src/odemis/driver/lakeshore.py
```diff
@@ -502,9 +501,18 @@ def _parseMessage(self, msg):
...
+            if self._heating:
+                if self._temperature < self._setpoint:  # heating is enabled
+                    self._temperature += 0.05 * self._heating  # simulate heating
+                else:
+                    self._temperature -= 0.01
+            else:  # no heating so no temperature control
+                # maintain stable temperature
+                if self._temperature > self._stable_temperature:
+                    self._temperature -= 0.1  # cool off with no heating
```
pieleric, line 515: This is a little odd with the simulator as the temperature rate depends on how frequently you're reading it. That said, for now that should be enough... so let's keep it like this.

src/odemis/driver/test/lakeshore_test.py
```diff
@@ -111,12 +111,15 @@ def test_heating_test(self):
         Test heating
         """
         old = self.dev.targetTemperature.value
+        # set a target temperature 1 degree higher and start heating
         self.dev.targetTemperature.value = self.dev.temperature.value + 1
         self.dev.heating.value = 3  # fast heating
+        # wait to heat up and then check if the temperature rose by a degree
         time.sleep(30)
-        self.assertAlmostEqual(self.dev.temperature.value, self.dev.targetTemperature.value)
+        self.assertAlmostEqual(self.dev.temperature.value, self.dev.targetTemperature.value, places=0)
```
pieleric, line 119: "places=0" is not very readable/clear. I'd suggest to replace with `delta=0.5` .

pieleric (changes requested):
Indeed, looks pretty much ready. You still need to improve the check for the correct device. I've also suggested a few more tweaks.

src/odemis/driver/lakeshore.py
amuskens: Should be fixed now.

src/odemis/driver/lakeshore.py
```diff
@@ -273,7 +273,16 @@ def GetIdentifier(self):
...
-        manufacturer, md, serialn, firmware = identity.decode("latin1").split(',')
+        try:
+            manufacturer, md, serialn, firmware = identity.decode("latin1").split(',')
+        except TypeError:
+            raise IOError("Invalid identifier received")
+
+        if manufacturer != "LSCI":
+            raise IOError("Invalid device manufacturer")
+        if md != "MODEL335":
+            raise IOError("The model is not 335")
```
pieleric, line 284: This part goes into `_findDevice()`

pieleric (changes requested):
I have just a tiny tiny remark about the identifier check, and it's done :-)

PR #1497, opened 2020-09-11
xtclient: new functionalities required for multibeam

tepals:
- pitch, stigmators, dc coils en switch multi/single beam, get scan mode

### DIFF
--- a/src/odemis/driver/test/xt_client_test.py
+++ b/src/odemis/driver/test/xt_client_test.py
@@ -296,10 +296,10 @@ class TestMicroscope(unittest.TestCase):
         stig_x, stig_y = (init_stig[0] + 1e-3, init_stig[1] + 1e-3)
         stig_x = stig_x if stig_range['x'][0] < stig_x < stig_range['x'][1] else init_stig[0] - 1e-3
         stig_y = stig_y if stig_range['y'][0] < stig_y < stig_range['y'][1] else init_stig[1] - 1e-3
-        new_stig = (stig_x, stig_y)
-        self.microscope.set_primary_stigmator(*new_stig)
-        self.assertEqual(self.microscope.get_primary_stigmator()[0], new_stig[0])
-        self.assertEqual(self.microscope.get_primary_stigmator()[1], new_stig[1])
+        self.microscope.set_primary_stigmator(stig_x, stig_y)
+        self.assertEqual(self.microscope.get_primary_stigmator()[0], stig_x)
+        self.assertEqual(self.microscope.get_primary_stigmator()[1], stig_y)
+        # Set back to the initial stigmator value, so the system is not misaligned after finishing the test.
         self.microscope.set_primary_stigmator(*init_stig)
 
     def test_secondary_stigmator(self):
@@ -311,10 +311,10 @@ class TestMicroscope(unittest.TestCase):
         stig_x, stig_y = (init_stig[0] + 1e-3, init_stig[1] + 1e-3)
         stig_x = stig_x if stig_range['x'][0] < stig_x < stig_range['x'][1] else init_stig[0] - 1e-3
         stig_y = stig_y if stig_range['y'][0] < stig_y < stig_range['y'][1] else init_stig[1] - 1e-3
-        new_stig = (stig_x, stig_y)
-        self.microscope.set_secondary_stigmator(*new_stig)
-        self.assertEqual(self.microscope.get_secondary_stigmator()[0], new_stig[0])
-        self.assertEqual(self.microscope.get_secondary_stigmator()[1], new_stig[1])
+        self.microscope.set_secondary_stigmator(stig_x, stig_y)
+        self.assertEqual(self.microscope.get_secondary_stigmator()[0], stig_x)
+        self.assertEqual(self.microscope.get_secondary_stigmator()[1], stig_y)
+        # Set back to the initial stigmator value, so the system is not misaligned after finishing the test.
         self.microscope.set_secondary_stigmator(*init_stig)
 
     def test_dc_coils(self):
--- a/src/odemis/driver/test/xt_client_test.py
+++ b/src/odemis/driver/test/xt_client_test.py
@@ -70,6 +70,7 @@ class TestMicroscope(unittest.TestCase):
             self.skipTest("No hardware available.")
         if self.microscope.get_vacuum_state() != 'vacuum':
             self.skipTest("Chamber needs to be in vacuum, please pump.")
+        self.xt_type = "xttoolkit" if "xttoolkit" in self.microscope.swVersion.lower() else "xtlib"
 
     def test_acquisition(self):
         """Test acquiring an image."""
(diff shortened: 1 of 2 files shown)


pieleric (changes requested):
Per se, this code looks alright. However, this is only 50% of the implementation. None of these new functions should be accessed directly by any Odemis client. There should be "official" API on the components which translate to these internal functions. I think many of them could be provided via a new VA on the e-beam (Scanner) component. For instance `.pitch` as a FloatContinuous. The multibeam selection could also be a VA, something like `beamlets`=EnumeratedVA({1, 64}).  Maybe the stigmators could be represented as extra axes on the Focus component. The DC coils info is a little harder to map. It could be a VA with an explicit getter, or just an special function.

tepals:
> Per se, this code looks alright. However, this is only 50% of the implementation. None of these new functions should be accessed directly by any Odemis client. There should be "official" API on the components which translate to these internal functions. I think many of them could be provided via a new VA on the e-beam (Scanner) component. For instance `.pitch` as a FloatContinuous. The multibeam selection could also be a VA, something like `beamlets`=EnumeratedVA({1, 64}). Maybe the stigmators could be represented as extra axes on the Focus component. The DC coils info is a little harder to map. It could be a VA with an explicit getter, or just an special function.

Yes, that's true. I don't know from where and how this functionality will be accessed. @Sabrina85 have you thought about that? Could it also be useful to create a Multibeam component, next to the Scanner, Stage, Focus etc., or GridScanner or something?

Sabrina85:
> > Per se, this code looks alright. However, this is only 50% of the implementation. None of these new functions should be accessed directly by any Odemis client. There should be "official" API on the components which translate to these internal functions. I think many of them could be provided via a new VA on the e-beam (Scanner) component. For instance `.pitch` as a FloatContinuous. The multibeam selection could also be a VA, something like `beamlets`=EnumeratedVA({1, 64}). Maybe the stigmators could be represented as extra axes on the Focus component. The DC coils info is a little harder to map. It could be a VA with an explicit getter, or just an special function.
> 
> Yes, that's true. I don't know from where and how this functionality will be accessed. @Sabrina85 have you thought about that? Could it also be useful to create a Multibeam component, next to the Scanner, Stage, Focus etc., or GridScanner or something?

Good point! So far we haven't discussed any API yet. So far we everywhere directly call the respective methods. 
As far as I understood the xt client should be also used for other TFS microscopes @pieleric ? Usually we have already SEM component with certain children. Would it be an option to make it a child there. Or alternatively, I could imagine a wrapper component for the SEM. But practically speaking I don't know if that would work.

src/odemis/driver/xt_client.py
```diff
@@ -174,6 +174,18 @@ def set_scan_mode(self, mode):
             self.server._pyroClaimOwnership()
             self.server.set_scan_mode(mode)
 
+    def get_scan_mode(self):
+        """
+        Get the scan mode.
+        Returns
+        -------
+        mode: str
+            Name of desired scan mode, one of: unknown, external, full_frame, spot, or line.
```
Sabrina85, line 183: desired -> set? As you don't want to set a "desired" mode, but want to know which one is set.

src/odemis/driver/xt_client.py
```diff
@@ -612,6 +624,144 @@ def set_autostigmator(self, channel_name, state):
...
+        Returns
+        -------
+        list, len 4
+            These 4 items describe 4x2 transformation matrix for 1 um beam shift using DC coils.
+        """
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            return self.server.get_dc_coils()
+
+    def get_use_case(self):
```
Sabrina85, line 740: This name does not directly reveal, what actually is meant with use case. However, if we do implement a proper API for all the xt functionality, then we could just do it there, and leave here the name the same as the xt call. Otherwise, I would suggest something like "get_sem_mode".

src/odemis/driver/xt_client.py
```diff
@@ -612,6 +624,144 @@ def set_autostigmator(self, channel_name, state):
...
+        """
+        Get the current state of the use case.
```
Sabrina85, line 742: Here you could add one more sentence saying what is meant with use case. I only understood when reading the return value description.

src/odemis/driver/xt_client.py
```diff
@@ -612,6 +624,144 @@ def set_autostigmator(self, channel_name, state):
...
+            return self.server.secondary_stigmator_info()
+
+    def get_dc_coils(self):
+        """
+        Get the four values of the dc coils.
+
+        Returns
+        -------
+        list, len 4
+            These 4 items describe 4x2 transformation matrix for 1 um beam shift using DC coils.
```
Sabrina85, line 734: I don't understand why these 4 values describe a 1um shift. I thought they are just the current position values of the beam shift and can be modified in any way.

src/odemis/driver/xt_client.py
```diff
@@ -612,6 +624,144 @@ def set_autostigmator(self, channel_name, state):
...
+        state: 'MultiBeamTile' or 'SingleBeamlet'
+
+        """
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            return self.server.get_use_case()
+
+    def set_use_case(self, state):
+        """
+
```
Sabrina85, line 755: Description missing.

src/odemis/driver/xt_client.py
```diff
@@ -612,6 +624,144 @@ def set_autostigmator(self, channel_name, state):
...
+            self.server._pyroClaimOwnership()
+            return self.server.get_dc_coils()
+
+    def get_use_case(self):
+        """
+        Get the current state of the use case.
+
+        Returns
+        -------
+        state: 'MultiBeamTile' or 'SingleBeamlet'
```
Sabrina85, line 746: Add type, though it is kind of obvious, but for completion.

src/odemis/driver/xt_client.py
```diff
@@ -612,6 +624,144 @@ def set_autostigmator(self, channel_name, state):
...
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            return self.server.get_use_case()
+
+    def set_use_case(self, state):
+        """
+
+        Parameters
+        ----------
+        state: 'MultiBeamTile' or 'SingleBeamlet'
```
Sabrina85, line 758: Add type.

src/odemis/driver/xt_client.py
```diff
@@ -612,6 +624,144 @@ def set_autostigmator(self, channel_name, state):
...
+        -------
+        pitch: float, [um]
+            The distance between two beams of the multiprobe pattern.
+        """
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            return self.server.set_pitch(pitch)
+
+    def pitch_info(self):
+        """"Returns the unit and range of the pitch."""
```
Sabrina85, line 654: In the tests, you explicitly call `self.microscope.pitch_info()["range"]`. So is it a dict that is returned with specific keys "unit" and "range"? Might be worse to extend the doc string. Same for the other "info" methods.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -276,6 +277,66 @@ def test_set_beam_shift(self):
...
+        stig_range = self.microscope.primary_stigmator_info()["range"]
+        init_stig = self.microscope.get_primary_stigmator()
+        stig_x, stig_y = (init_stig[0] + 1e-3, init_stig[1] + 1e-3)
+        stig_x = stig_x if stig_range['x'][0] < stig_x < stig_range['x'][1] else init_stig[0] - 1e-3
+        stig_y = stig_y if stig_range['y'][0] < stig_y < stig_range['y'][1] else init_stig[1] - 1e-3
+        new_stig = (stig_x, stig_y)
+        self.microscope.set_primary_stigmator(*new_stig)
+        self.assertEqual(self.microscope.get_primary_stigmator()[0], new_stig[0])
+        self.assertEqual(self.microscope.get_primary_stigmator()[1], new_stig[1])
+        self.microscope.set_primary_stigmator(*init_stig)
```
Sabrina85, line 303: Maybe mention that this is important to set back the old values so the system is not suddenly misaligned. Same for secondary stigmator.

Sabrina85 (changes requested):
Looks good! I have only some minor comments for clarifying parameters and doc strings.

src/odemis/driver/xt_client.py
pieleric: Actually, the doctring is not clear here. If I understand correctly, the return value is 4 tuples containing each 2 floats. So that's 8 values in total.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -276,6 +277,66 @@ def test_set_beam_shift(self):
...
+    def test_secondary_stigmator(self):
+        """Test getting and setting the secondary stigmator."""
+        if self.xt_type != 'xttoolkit':
+            self.skipTest("This test needs XTToolkit to run.")
+        stig_range = self.microscope.secondary_stigmator_info()["range"]
+        init_stig = self.microscope.get_secondary_stigmator()
+        stig_x, stig_y = (init_stig[0] + 1e-3, init_stig[1] + 1e-3)
+        stig_x = stig_x if stig_range['x'][0] < stig_x < stig_range['x'][1] else init_stig[0] - 1e-3
+        stig_y = stig_y if stig_range['y'][0] < stig_y < stig_range['y'][1] else init_stig[1] - 1e-3
+        new_stig = (stig_x, stig_y)
```
Kleijwegt, line 314: What is the advantage of packing these values in a tuple of in the next lines you will aways unpack these values or use only one of them by indexing?

src/odemis/driver/xt_client.py
```diff
@@ -612,6 +624,144 @@ def set_autostigmator(self, channel_name, state):
             self.server._pyroClaimOwnership()
             return self.server.set_autostigmator(channel_name, state)
 
+    def get_pitch(self):
+        """
+        Get the pitch between two beams within the multiprobe pattern.
```
Kleijwegt, line 629: Could be nice to add that it are two directly neighbouring beams (at least that is what I expect).

src/odemis/driver/xt_client.py
```diff
@@ -612,6 +624,144 @@ def set_autostigmator(self, channel_name, state):
...
+
+        Parameters
+        -------
+        (x, y) control values of primary stigmator, unitless.
+        """
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            return self.server.set_primary_stigmator(x, y)
+
+    def primary_stigmator_info(self):
```
Kleijwegt, line 687: For the order of the code I would expect first the info and then the getter/setter.
While reading the code if you wonder what the range is you know that if could be found in the method you just read. But maybe that doesn't make sense to you.

src/odemis/driver/xt_client.py
```diff
@@ -612,6 +624,144 @@ def set_autostigmator(self, channel_name, state):
...
+        Parameters
+        -------
+        (x, y) control values of primary stigmator, unitless.
+        """
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            return self.server.set_secondary_stigmator(x, y)
+
+    def secondary_stigmator_info(self):
+        """"Returns the unit and range of the secondary stigmator."""
```
Kleijwegt, line 722: Same as for the primary stigmator

Kleijwegt (changes requested):

src/odemis/driver/xt_client.py
tepals: Lets leave the name as it is for now, and give it a proper name in the API. I will update the docstring to better explain the method.

pieleric: I think that the names should be just the same as the names used by the xtLib/xtToolkit. The xtadapter is not really try to hide the low-level API and provide a high-level view, but just to pass the messages as-is.
With all this said... I agree with Sabrina that here "usecase" is a really bad name. Such a bad name that it'd be fine to bend the rule, and pick a better name ;-)

src/odemis/driver/xt_client.py
tepals: It is indeed a list of 4 tuples each containing 2 floats.

I copied this explanation line from the document you linked to in the specification. Now that I look at it again, the 1 um doesn't make sense. Although I am also not sure what these values represent. Or is it good enough if I change this to:
`These 4 items describe 4x2 transformation matrix for a required beam shift using DC coils. `

src/odemis/driver/xt_client.py
tepals: I think it's is out of scope for this PR to do it for all info methods. I will do it for pitch and the stigmators and update the others in a new PR.

src/odemis/driver/test/xt_client_test.py
tepals: There is no advantage, I will change it.

src/odemis/driver/xt_client.py
tepals: It does make more sense, however in the rest of this class the order is first the methods then the info. So for consistency I will leave it like this. If you ever feel like updating the whole class and changing the order go ahead :grin:.

tepals:
Updated the PR. Not sure about my explanation of the dc coils in the docstring, because I'm also not completely sure what these values represent.

Sabrina85 (approved):
It looks all good to me! 

The dc coils description is indeed not completely correct. As far as I understand is this method purely about getting the current settings on the dc coils and doesn't know anything about the beam shift. That is handled in the dc coils driver. I would suggest that we simplify the description.

@pieleric How should we proceed. Will you first merge the pr as it is as Thera is on holiday and I can make a small update for the description or is there a better way to do so (pull this branch and make a commit on top of this - don't know if that actually works :D )

The renaming of the get_use_case is still open. Should we leave it as it is @pieleric ? Otherwise, I could include it in the update with the dc coils description.

src/odemis/driver/xt_client.py
```diff
@@ -174,6 +174,18 @@ def set_scan_mode(self, mode):
...
+        """
+        Get the scan mode.
+        Returns
+        -------
+        mode: str
+            Name of set scan mode, one of: unknown, external, full_frame, spot, or line.
+        """
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            return self.server.get_scan_mode()
```
Kleijwegt, line 187: issue so that I can reference to a part of code

PR #1498, opened 2020-09-11
Xtclient add missing unittests

tepals:
Make sure almost all code is covered by unittests and at the end of each test the values are set back to their initial value.

### DIFF
--- a/src/odemis/driver/test/xt_client_test.py
+++ b/src/odemis/driver/test/xt_client_test.py
@@ -77,22 +77,23 @@ class TestMicroscope(unittest.TestCase):
         image = self.microscope.get_latest_image(channel_name='electron1')
         self.assertEqual(len(image.shape), 2)
 
-    @unittest.skip("Skip because the microscope stage is not used, an external stage is used.")
     def test_move_stage(self):
         """
         Test that moving the microscope stage to a certain x, y, z position moves it to that position.
         """
-        pos = self.stage.position.value.copy()
+        if self.xt_type == 'xttoolkit':
+            self.skipTest("Skip because the microscope stage is not used, an external stage is used.")
+        init_pos = self.stage.position.value.copy()
         f = self.stage.moveRel({"x": 2e-6, "y": 3e-6})
         f.result()
-        self.assertNotEqual(self.stage.position.value, pos)
+        self.assertNotEqual(self.stage.position.value, init_pos)
 
-        init_pos = self.stage.position.value.copy()  # returns [x, y, z, r, t]  in [m]
+        new_pos = self.stage.position.value.copy()  # returns [x, y, z, r, t]  in [m]
         # move stage to a different position
-        position = {'x': init_pos['x'] - 1e-6, 'y': init_pos['y'] - 2e-6, 'z': 10e-6}  # [m]
+        position = {'x': new_pos['x'] - 1e-6, 'y': new_pos['y'] - 2e-6, 'z': 10e-6}  # [m]
         f = self.stage.moveAbs(position)
         f.result()
-        self.assertNotEqual(self.stage.position.value, pos)
+        self.assertNotEqual(self.stage.position.value, new_pos)
 
         stage_position = self.stage.position.value.copy()
         # test move_stage method actually moves HW by the requested value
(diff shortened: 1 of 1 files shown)


tepals:
https://github.com/delmic/odemis/pull/1497 Needs to be merged first, because this PR uses `get_scan_mode` from that PR.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -70,56 +70,75 @@ def setUp(self):
...
         self.assertEqual(len(image.shape), 2)
 
-    @unittest.skip("Skip because the microscope stage is not used, an external stage is used.")
     def test_move_stage(self):
         """
         Test that moving the microscope stage to a certain x, y, z position moves it to that position.
         """
-        pos = self.stage.position.value.copy()
+        if self.xt_type == 'xttoolkit':
+            self.skipTest("Skip because the microscope stage is not used, an external stage is used.")
```
pieleric, line 85: I don't get the relationship between stage and xttoolkit. Or is it because the current system with xttoolkit has an external stage? 
This might be the case on many systems with a xtlib connection too. Isn't there a way to detect that the stage is missing in xtlib (and xttoolkit)?  In general, it would be handy to fail at initialisation if the user requests a stage component but there is no such hardware. 
Here, if there is no way to check the stage is available, maybe check if the stage is referenced (aka homed)? You could also just leave the unconditional skip, but with a simpler message: "Microscope stage not tested, too dangerous".

src/odemis/driver/test/xt_client_test.py
```diff
@@ -70,56 +70,75 @@ def setUp(self):
...
+        init_pos = self.microscope.get_stage_position()  # returns [x, y, z, r, t]  in [m]
+        # move stage to a different position
+        position = {'x': init_pos['x'] - 1e-6, 'y': init_pos['y'] - 2e-6, 'z': 10e-6}  # [m]
+        self.microscope.move_stage(position)
+        self.microscope.stop_stage_movement()
+        time.sleep(1)
+        self.assertFalse(self.microscope.stage_is_moving())
+        # Move stage back to initial position
+        self.microscope.move_stage(init_pos)
```
pieleric, line 134: I'm not a fan of using these internal functions. The testing should only test the functions provided by the Odemis API. Here that'd be `moveRel()`, `moveAbs()`, `future.cancel()`, `.stop()`, and `.position`.

If you really want to test these internal functions, maybe make it really clear, by moving this test to a separate TestCase class, or appending to the function name `_internal`.
There are several other testing functions in the same vein that could be adjusted the same way.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -226,9 +260,10 @@ def test_vent_and_pump(self):
         self.microscope.pump()
         self.assertEqual(self.microscope.get_vacuum_state(), 'vacuum')
```
pieleric, line 261: These vacuum related function should be accessed via an "Chamber" actuator. See the Phenom driver API for an example.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -226,9 +260,10 @@ def test_vent_and_pump(self):
...
 
-    @unittest.skip("Skip because the microscope stage is not used, an external stage is used.")
     def test_home_stage(self):
         """Test that the stage is homed after home_stage is called."""
+        if self.xt_type == 'xttoolkit':
+            self.skipTest("Skip because the microscope stage is not used, an external stage is used.")
         self.microscope.home_stage()
```
pieleric, line 267: This should be provided via `stage.reference()`, and `.referenced`.

pieleric (changes requested):
Most of the test cases are testing internal functions that no one should use directly. Normally you don't need to test such internal functions directly. You should test the public API, which will indirectly trigger the internal funtion. However, here I can imagine it might be handy in order to test early the support for new functions. I'd recommend to move all this internal function test to a separate TestCase class.

Kleijwegt (approved):

tepals:
I moved all the functions that test internal functionality to a separate class. You are correct, the "internal tests" come in useful to be able to test new functionality early.

I have created a new issue for your comments about the API: https://delmic.atlassian.net/browse/BEN-25 
There are probably more functions that need support from a proper API, I will gather those on that issue.

src/odemis/driver/test/xt_client_test.py
tepals: For now I've gone with "Microscope stage not tested, too dangerous". I did add this point to https://delmic.atlassian.net/browse/BEN-25

PR #1504, opened 2020-09-18
gui cont tabs SPARC2Align: only update L1 FAV_POS_ACTIVE when optical path is ready

pieleric:
When activating the tab, if L1 is not far from the FAV_POS (eg, it just
referenced), and the optical path is still being moved from a previous
tab change, then it could have some time to update the position, which
the tab would use as new FAV_POS, until it finally asks to move the
optical path. In such case, the FAV_POS would stay at an incorrect
position.

=> Do as for the fiber aligner: only update when the optical path is
fully ready. So we know that by that time the L1 position is not
updating due to a move from the OPM.

Also fix a potential rare bug causing the GUI to crash when switching to
fiber align mode.


lazem (approved):
Looks good.

PR #1505, opened 2020-09-19
Smaract - Poll posiiton on 5dof and added speed VA

amuskens:


src/odemis/driver/smaract.py
```diff
@@ -1590,9 +1601,42 @@ def _doMoveAbs(self, future, pos):
         max_dur = dur * 2 + 1
         logging.debug("Expecting a move of %f s, will wait up to %g s", dur, max_dur)
 
+<<<<<<< HEAD
```
Kleijwegt, line 1604: @amuskens I think this does not belong here and is a result of a not completed pull/rebase/merge?

src/odemis/driver/smaract.py
```diff
@@ -2747,6 +2794,7 @@ def _doMoveRel(self, future, pos):
             CancelledError: if cancelled before the end of the move
         """
         with future._moving_lock:
+<<<<<<< HEAD
```
Kleijwegt, line 2797: I think this is a result of the same rebase/merge/pull which is not yet completed

src/odemis/driver/smaract.py
```diff
@@ -2778,6 +2843,7 @@ def _doMoveAbs(self, future, pos):
...
```
Kleijwegt, line 2846: Last mergeconflict

src/odemis/driver/smaract.py
```diff
@@ -1590,9 +1601,42 @@ def _doMoveAbs(self, future, pos):
...
+                                           (max_dur, dur))
+
+                    # Update the position from time to time (10 Hz)
+                    if now - last_upd > 0.1:
+                        self._updatePosition()
+                        last_upd = time.time()
+
+                    # Wait half of the time left (maximum 0.1 s)
+                    left = end - time.time()
+                    sleept = max(0.001, min(left / 2, 0.1))
```
Kleijwegt, line 1634: Maybe I miss something but to me the naming sleep_time or just sleep would be more logical than sleept

pieleric (changes requested):
The merge conflicts makes this PR quite of a mess.
Apart from this, we still need to check on hardware that indeed the controller updates the position while it's moving (probably it does, but if it doesn't the whole code is pointless so we need to check!). For now, the simulator doesn't update the position. You could look at the TMCM code to get this feature. This would be handy both for doing some testing, but also for the simulation in the GUI.

I think that we should also reduce the update frequency when it's not moving, but that's a minor point compared to the rest.

src/odemis/driver/smaract.py
amuskens: Uh oh hahhaa.... let me fix that...

amuskens:
Oops.. looks like messed up with the rebase. Anyways I fixed it and made a new commit.

src/odemis/driver/smaract.py
```diff
@@ -1254,8 +1254,18 @@ def __init__(self, name, role, locator, axes, ref_on_init=False, linear_speed=0.
...
+
+        self.speed = VigilantAttribute({'x': self.linear_speed,
+                                   'y': self.linear_speed,
+                                   'z': self.linear_speed,
+                                   'rx': self.rotary_speed,
+                                   'rz': self.rotary_speed},
+                                   readonly=True)
+
+        # create a timer thread that will be used to update the position while waiting for events
+        self.update_position_timer = RepeatingTimer(0.1, self._updatePosition)
```
pieleric, line 1266: No, I'd say, start with a period of 1s (ie, 1Hz), and during a move, switch to 0.05s (20Hz). You can do that by changing the RepeatingTimer.period at the beginning and end of the move (in a `finally` block).

src/odemis/driver/smaract.py
```diff
@@ -1814,7 +1824,19 @@ def SA_MC_GetProperty_s(self, id, property_key, val, ioArraySize):
...
+        for i in range(n_steps):
+            intermediate = SA_MC_Pose()
+            intermediate.x = self.pose.x + (self.target.x - self.pose.x) * i / n_steps
+            intermediate.y = self.pose.y + (self.target.y - self.pose.y) * i / n_steps
+            intermediate.z = self.pose.z + (self.target.z - self.pose.z) * i / n_steps
+            intermediate.rx = self.pose.rx + (self.target.rx - self.pose.rx) * i / n_steps
+            intermediate.rz = self.pose.rz + (self.target.rz - self.pose.rz) * i / n_steps
+            self.pose = copy.copy(intermediate)
+            time.sleep(0.05)
+
```
pieleric, line 1839: That doesn't properly simulate the move! You have to take into account the speed. I'd highly suggest to not put the simulation code in the WaitForEvent(), but in the GetPose(). See in the TMCM simulator how this is done.

src/odemis/driver/test/smaract_test.py
```diff
@@ -274,7 +274,7 @@ def test_move_cancel(self):
         self.dev.moveAbs({'x': 0, 'y': 0, 'z': 0, 'rx': 0, 'rz': 0}).result()
         new_pos = {'x':0.001, 'y': 0, 'z': 0.0007, 'rx': 0.001, 'rz': 0.002}
         f = self.dev.moveAbs(new_pos)
-        time.sleep(0.05)
+        time.sleep(0.01)
         f.cancel()
 
         difference = new_pos['x'] - self.dev.position.value['x']
```
pieleric, line 280: Let's add a test case for the position update during move. It should be just a matter of regularly reading the position for eg 1s after asking the move, and checking that you've got at least one position which is different from the initial and target positions.

pieleric (changes requested):
It's much better without conflicts ;-)
Still, I'd like to have a different frequency update during move and when not moving. Also, the simulator is too crude as it is.

pieleric (changes requested):
I'm not convinced by this new feature "safe_move" both at the implementation level and at the higher level on how the whole system is organized. I'm wondering whether it'd better that it's the code which request the reference which is in charge to move the hardware to a safe position. It'd be more explicit. I think that as this hardware has a little odd behaviour of referencing (ie, it can do it anywhere), there is some merit to it.
Please move this new "safe pos after referencing"  feature to a separate PR.
In this PR, please just upgrade the 5DoF simulator, and you should be done.

src/odemis/driver/smaract.py
```diff
@@ -1554,6 +1557,12 @@ def _doReference(self, future):
             if self._is_referenced():
                 self.referenced._value = {a: True for a in self.axes.keys()}
                 self._updatePosition()
+                if self._safe_pos is not None:
+                    logging.info("Moving axes to safe position")
+
+                    self._checkMoveAbs(self._safe_pos)
+                    f = self._createMoveFuture()
+                    self._doMoveAbs(f, self._safe_pos)
```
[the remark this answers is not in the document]
amuskens: What do you think is the best way to address the issue that it can't be cancelled?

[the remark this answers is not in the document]
pieleric: See the other PR.

src/odemis/driver/smaract.py
```diff
@@ -1747,27 +1759,101 @@ def SA_MC_Move(self, id, p_pose):
...
+            # calculate intermediate positions
+            dx = self.target.x - self.pose.x
+            if dx >= 0:
+                self.pose.x = self.pose.x + lin_speed * dt
+                if self.pose.x >= self.target.x:
+                    self.pose.x = self.target.x
+            elif dx < 0:
+                self.pose.x = self.pose.x - lin_speed * dt
+                if self.pose.x < self.target.x:
+                    self.pose.x = self.target.x
```
pieleric, line 1805: This is 5 times the same code, with slightly different variable names. Please move to a separate function, and call it 5 times.

src/odemis/driver/smaract.py
```diff
@@ -1814,7 +1900,12 @@ def SA_MC_GetProperty_s(self, id, property_key, val, ioArraySize):
 
     def SA_MC_WaitForEvent(self, id, p_ev, timeout):
         ev = _deref(p_ev, SA_MC_Event)
-        time.sleep(0.25)
+        start_time = time.time()
+        
+        while time.time() < self._current_move_finish \
+            and time.time() < start_time + timeout.value:
+            time.sleep(0.05)
```
pieleric, line 1907: I don't think it behaves properly if there is a timeout passed. It will conclude that the movement is completed. Maybe with the rest of the driver it's fine... but if it's not breaking other things, I'd write this loop like this:
``` python
while time.time() < self._current_move_finish:
     if  time.time() < start_time + timeout.value:
         return    # Special timeout value?
     time.sleep(0.05)
```

src/odemis/driver/smaract.py
```diff
@@ -2303,6 +2394,7 @@ def __str__(self):
         return self.strerror
 
 
+
```
pieleric, line 2397: No extra line needed here.

pieleric (changes requested):
It now looks almost ready to merge. I just have some small suggestions for the new simulator code.

src/odemis/driver/smaract.py
```diff
@@ -900,6 +900,11 @@ class SA_MC_Event(Structure):
         ("u", SA_MC_EventData),
         ]
 
+    def __str__(self):
+        return "SA_MC_EVenet {type: %s, i32: %s}" % \
```
pieleric, line 904: I'm not sure why you need it to be pretty, but if you want to have such a function, please spell correctly the class name, otherwise it's confusing: SA_MC_EVenet -> SA_MC_Event

src/odemis/driver/smaract.py
```diff
@@ -900,6 +900,11 @@ class SA_MC_Event(Structure):
...
+            (MC_5DOF_DLL.event_name.get(self.type),
+             MC_5DOF_DLL.err_code.get(self.i32))
```
pieleric, line 906: if the event name or the err_code are unknow, they will be shown as None, which is even less useful than the raw value. => write like `MC_5DOF_DLL.event_name.get(self.type, self.type)`, so that it defaults to showing the raw value.

src/odemis/driver/smaract.py
```diff
@@ -1062,6 +1067,10 @@ class MC_5DOF_DLL(CDLL):
     # events
     SA_MC_EVENT_MOVEMENT_FINISHED = 0x0001
 
+    event_name = {
+        0x0001: "MOVEMENT_FINISHED"
```
pieleric, line 1071: The link is clearer if you write:
`SA_MC_EVENT_MOVEMENT_FINISHED: "MOVEMENT_FINISHED"`

pieleric (changes requested):
I think the simulator is now good... but I have remarks about the extension to Even.

src/odemis/driver/smaract.py
amuskens: Just to make the logs nicer.

PR #1506, opened 2020-09-21
driver andorshrk: option to workaround slit reference issue

pieleric:
One spectrograph's slit had the reference switch not working.
As we don't need a very precise slit opening (because the user can
visually check), we can work around it by forcing the slit to go to the
maximum opening before every move. This has a similar effect as the
referencing as it allows to know where is the slit.

### DIFF
--- a/src/odemis/driver/andorshrk.py
+++ b/src/odemis/driver/andorshrk.py
@@ -300,8 +300,11 @@ class Shamrock(model.Actuator):
         camera (None or AndorCam2): Needed if the connection is done via the
           I²C connector of the camera.
         inverted (None): it is not allowed to invert the axes
-        slits (None or dict int -> str): names of each slit,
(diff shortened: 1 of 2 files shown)


src/odemis/driver/andorshrk.py
```diff
@@ -347,12 +350,20 @@ def __init__(self, name, role, device, camera=None, accessory=None,
         dependencies = dependencies or {}
         rng = rng or {}
 
+        self._slit_names = SLIT_NAMES.copy()
+        self._force_slit_max = set()
         slits = slits or {}
-        for i in slits:
+        for i, slitn in slits.items():
```
Kleijwegt, line 356: Does there needs to be an underscore (or capital) between slit and n? so slit_n to make it distinct that n refers to name?

Kleijwegt (approved):

PR #1508, opened 2020-09-22
odemis-mic-selector: dedicated error message if no EEPROM IDs found

pieleric:
There is typically always at least one optical module, so one EEPROM ID on a SPARC.
If no ID is found, it's a sign of something fishy, so it's best to
distinguish this case explicitly from the standard message "no matching
hardware found".

### DIFF
--- a/install/linux/usr/bin/odemis-mic-selector
+++ b/install/linux/usr/bin/odemis-mic-selector
@@ -126,6 +126,11 @@ def guess_hw(config):
             else:
                 logging.warning("Failed to read EEPROM IDs, will pretend no ID is connected", exc_info=True)
 
+        if not eids and all(c.eeprom for c in config):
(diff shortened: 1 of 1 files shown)


Kleijwegt (commented):
Don't we want to also have a testcase for this change/error?

pieleric:
There are couple of limitations for testing. Firstly, there is no test cases yet for this module (admittedly, that's more a sign that we should add some testing) and for the test it would need to be forced to use a simulator of the PCU, which cannot yet be done easily. Secondly, it's hard to test hardware error handling (in general). We could have some special trick to activate misbehaviour of the simulator to simulate such hardware error, but that's quite a bit of extra work.
So for now, it's been only tested "manually" by modifying the simulator code, I'd suggest that if we want to add test cases, this is implemented in a separate PR.

PR #1516, opened 2020-09-29
gui plugin: add log when pressing a button

pieleric:
As the button callbacks run in a separate thread, it can easily crash
the GUI if it ever calls a wxPython function. So it's handy to know if
a callback has started if a crash is followed.


VictoriaDel (approved):
It's quite useful !!

PR #1517, opened 2020-09-30
Smaract Picoscale Driver

philipwinklerdelmic:


src/odemis/driver/smaract.py
```diff
@@ -3084,3 +3085,1139 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+        channels: (str --> int) dictionary mapping channel names to channel numbers
+        ref_on_init: (bool) determines if the controller should automatically reference
+            on initialization
+        precision_mode: (0 <= int <= 5) strength of digital lowpass filter, a higher level corresponds to higher
+            precision, but lower velocity. Not available on all systems.
+        """
+        model.HwComponent.__init__(self, name, role, *args, **kwargs)
+
+        if not channels:
+            raise ValueError("Needs at least 1 axis.")
```
pieleric, line 3451: Please check more thoroughly `channels`. At least, it should check the channel numbers are within 0 and num_ch -1.

src/odemis/driver/smaract.py
```diff
@@ -3084,3 +3085,1139 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+            where <id> is the first part of a USB devices serial number which
+            is printed on the Picoscale controller.
+            If the controller has a TCP/IP connection, use:
+                network:<ip>:<port>
+            The device can also be addressed by its serial number:
+                usb:sn:<serial_number>
+                network:sn:<serial_number>
+        channels: (str --> int) dictionary mapping channel names to channel numbers
+        ref_on_init: (bool) determines if the controller should automatically reference
+            on initialization
```
pieleric, line 3444: You can make it clearer by saying "If True, will run full referencing just after initialization".

src/odemis/driver/smaract.py
```diff
@@ -3084,3 +3085,1139 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+    # API functions
+    def GetFiberLength(self):
+        """
+        returns: (float) fiber length in mm
+        """
+        return self.GetProperty_f64(SA_SIDLL.SA_PS_SYS_FIBERLENGTH_HEAD_PROP)
+
+    def GetExtensionFiberLength(self):
+        """
+        returns: (float) extension fiber length
```
pieleric, line 3600: Please indicate the unit (mm?)

src/odemis/driver/smaract.py
```diff
@@ -3084,3 +3085,1139 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+        """
+        return self.GetProperty_f64(SA_SIDLL.SA_PS_SYS_FIBERLENGTH_EXTENSION_PROP)
+
+    def GetWorkingDistance(self):
+        """
+        returns: (float, float) minimum and maximum working distance
```
pieleric, line 3606: Please indicate the unit.

src/odemis/driver/smaract.py
```diff
@@ -3084,3 +3085,1139 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+        self._polling_thread.start()
+
+        # Precision mode
+        # The precision mode is a special feature that needs to be purchased separately, so it is not available
+        # by default.
+        # TODO: this functionality has not yet been tested
+        if precision_mode != 0:
+            if not self.IsFeatureEnabled(SA_SIDLL.FEATURE_PRECISION_MODE):
+                raise ValueError("Device does not have precision mode. This feature needs to be purchased separately.")
+            self.SetPrecisionMode(precision_mode)
```
pieleric, line 3497: You probably should set it to 0, if it's requested to be 0. Otherwise it won't be possible to disable it,

src/odemis/driver/smaract.py
```diff
@@ -3084,3 +3085,1139 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+        ev = SA_SI_Event()
+
+        state = None
+        while state != end_state:
+            # Don't check event type, it's not always accurate.
+            # Instead, only one event is active and all other events are disabled, so it is clear
+            # which event we are waiting for.
+            self.core.SA_SI_WaitForEvent(self._id, byref(ev), t)
+            state = ev.devEventParameter & 0xffff
+            time.sleep(0.1)
```
pieleric, line 3762: I don't think you need to have a sleep here, as WaitForEvent already blocks.
However, you need to check if the timeout is passed.

src/odemis/driver/smaract.py
```diff
@@ -3084,3 +3085,1139 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+        """
+        encoded_key = SA_SIDLL.SA_SI_EVENT_NOTIFICATION_ENABLED_PROP << 16 | event_type
+        self.core.SA_SI_SetProperty_i32(self._id, c_uint32(encoded_key), c_int32(SA_SIDLL.SA_SI_DISABLED))
+
+    def WaitForEvent(self, end_state, timeout=float("inf")):
+        """
+        Blocks until event is triggered or timeout.
+        end_state (int): state of event variable to wait for
+        timeout (float): maximum time to wait in s. If inf, it will wait forever.
+        """
```
pieleric, line 3748: I think that you're making it too fancy. You should have a function which just wraps the underlying library function. So it would return whichever event.
*In addition* you could have a second function `_wait_for_event(state)` which actually does what this function does.
Please note that currently, the timeout doesn't work as expected. The proper way to do that is to compute at the beginning the maximum time `tend` this function can wait until. Based on that value and the current time, the timeout to WaitForEvent() should be adjusted.

src/odemis/driver/smaract.py
```diff
@@ -3084,3 +3085,1139 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+        Saves the configuration parameters (cf LoadConfiguration) and internal parameters of the
+        referencing procedure. This function is blocking.
+        """
+        self.SetProperty_i32(SA_SIDLL.SA_PS_AF_ADJUSTMENT_RESULT_SAVE_PROP, SA_SIDLL.SA_SI_ENABLED)
+        # There is no documentation on loading/saving the configuration in the manual, even though
+        # the functions are available in the header file. It is not clear how much time saving the
+        # parameters takes. It can be assumed that it's fast --> wait 1 s.
+        time.sleep(1)
+
+    def ValidateChannels(self):
```
pieleric, line 3704: This is not a wrapper function. Let's name it `_validate_channels()`.

src/odemis/driver/smaract.py
```diff
@@ -3084,3 +3085,1139 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+        ver = self.core.SA_SI_GetFullVersionString()
+        return ver.decode("latin1")
+
+    def EnableFullAccess(self):
+        """
+        Get full access to all device functionality (e.g. for adjustment procedure, loading properties, etc).
+        """
+        self.SetProperty_i32(SA_SIDLL.SA_PS_SYS_FULL_ACCESS_CONNECTION_PROP, SA_SIDLL.SA_SI_ENABLED)
+
+    def LoadConfiguration(self):
```
pieleric, line 3674: This is not a wrapper function => name it `_load_configuration()`

src/odemis/driver/smaract.py
```diff
@@ -3084,3 +3085,1139 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+            except Exception:
+                logging.exception("Referencing failure.")
+                raise
+            finally:
+                # If the state of the device is stable and all channels are valid, the referencing procedure
+                # succeeded. Typically, all channels become valid at the same time during the reference procedure.
+                # We still check all of them inidividually to be sure.
+                if self.IsStable():
+                    for ch, num in self.channels.items():
+                        self.referenced._value[ch] = self.IsValid(num)
```
pieleric, line 3940: Could we set the initial referencing to this? If not, it probably shouldn't be in the "finally".

src/odemis/driver/smaract.py
```diff
@@ -3084,3 +3085,1139 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+        Actually runs the validation code.
+        After the device was just turned on, it might take ~5 minutes, otherwise it is very fast.
+        future (Future): the future it handles
+        raise:
+            IOError: if validation failed due to hardware
+            CancelledError if was cancelled
+        """
+        with future._moving_lock:
+            try:
+                self.ValidateChannels()
```
pieleric, line 3978: You could put this function inline here? It would look more similar to the referencing procedure.

src/odemis/driver/smaract.py
```diff
@@ -3084,3 +3085,1139 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+        """
+        logging.debug("Cancelling referencing...")
+        future._must_stop = True  # tell the thread taking care of the referencing it's over
+        self.core.SA_SI_Cancel(self._id)
+
+        # Synchronise with the ending of the future
+        with future._moving_lock:
+            if not future._was_stopped:
+                logging.debug("Cancelling failed.")
+            return future._was_stopped
```
pieleric, line 4063: The main point of the lock is to make sure you call `Cancel()` only when a function is started.
I think that in the cancel() you could write:
``` python
        future._must_stop = True  # tell the thread taking care of the referencing it's over

        # Synchronise with the ending of the future
        with future._moving_lock:
            self.core.SA_SI_Cancel(self._id)
            if not future._was_stopped:
                logging.debug("Cancelling failed.")
            return future._was_stopped
```

The actual move functions would look like:
```python
with future._moving_lock:
    if future._must_stop:
         future._was_stopped = True
    starting_movement_function()

wait_for_event(long timeout)
:
# you can put more start (in locks), followed by waits out of locks
```

src/odemis/driver/smaract.py
```diff
@@ -3084,3 +3085,1139 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+            SA_SIDLL.SA_PS_AF_ADJUSTMENT_STATE_PROP: 0,
+            SA_SIDLL.SA_SI_NUMBER_OF_CHANNELS_PROP: 3,
+            SA_SIDLL.SA_PS_SYS_IS_STABLE_PROP: 1,
+            SA_SIDLL.SA_PS_CH_IS_VALID_PROP: 1,
+            SA_SIDLL.SA_PS_SYS_WORKING_DISTANCE_ACTIVATE_PROP: 0,
+        }
+
+        self.position0 = 10e-6
+        self.position1 = 20e-6
+        self.position2 = 30e-6
```
pieleric, line 4093: I think this is perfect to put in a list ;-)

src/odemis/driver/smaract.py
```diff
@@ -3084,3 +3085,1139 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+            self.SetPrecisionMode(precision_mode)
+
+        # State: starting until first referencing/validation procedure is done
+        # Position requests can be made during startup, however, the values are not going to be accurate until the
+        # initialization process is complete. The position thread is polling from the start.
+        # If during startup a call to .reference is made, the referencing future will be added to the
+        # executor and be carried out right after the initialization referencing is done.
+        # In conclusion, no special handling is required during startup, although it does not make much sense
+        # to use the driver until the startup is complete. The position getter will log a warning in this case.
+        self.state._set_value(model.ST_STARTING, force_write=True)
```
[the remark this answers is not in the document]
philipwinklerdelmic: I think it's nice to have some indication in the software of whether or not the device has finished its initial referencing procedure. It's particularly useful for the testcases, where I just wait until the state is "running". Waiting for "referenced" doesn't work because we have this distinction between validation and referencing and in case of validation I decided not to consider the channel referenced after the procedure is finished.

src/odemis/driver/smaract.py
```diff
@@ -3084,3 +3085,1139 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+        # it is recommended to reference (i.e. adjust) the system frequently (at least during startup)
+        # since even small changes in the experimental setup or environment might cause the old
+        # adjustment parameters to become non-ideal. Therefore, the values for all channels in .referenced
+        # will be False, until the full referencing routine has been performed.
+        # These procedure can take a while (up to 10 minutes), especially after the power of the system
+        # has just been turned on. Therefore, we don't wait until the procedure is complete.
+        # Starting from standby is generally much faster.
+        self._executor = CancellableThreadPoolExecutor(1)  # one task at a time
+        channel_ref = {ch: False for ch in self.channels}
+        self.referenced = model.VigilantAttribute(channel_ref, readonly=True)  # VA dict str(channel) -> bool
```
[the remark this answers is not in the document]
philipwinklerdelmic: No, I don't think it's possible to detect that.

src/odemis/driver/smaract.py
```diff
@@ -3084,3 +3085,1139 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
         self.stopping.set()
+
+
+class SA_SIDLL(CDLL):
+    """
+    Subclass of CDLL specific to SA_SI library, which handles error codes for
+    all the functions automatically.
+    """
+    SA_SI_TIMEOUT_INFINITE = 0xffffffff
+    SA_SI_STRING_MAX_LENGTH = 63
```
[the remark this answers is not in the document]
philipwinklerdelmic: On the other hand, the functions SA_SI_Cancel, SA_SI_EPK,... all have this prefix and the prefix also appears for these attributes in the documentation. There is also some events that are a bit different and use the "SA_PS" prefix. I think it might be clearer to keep all prefixes. What do you think?

src/odemis/driver/smaract.py
philipwinklerdelmic: I think I put it there for a purpose, just waiting for the end_state was sometimes not enough. But I should check that again.

src/odemis/driver/smaract.py
philipwinklerdelmic: Is that such a good idea? Doesn't this make it impossible to stop the 5 minute-long referencing procedure?

pieleric: It should still be possible to stop the procedure at almost any time. If that's not the case it's a bug.
The goal of the changes I suggest is only to ensure that it's not possible to cancel and then the procedure starts. (Due to race condition)

philipwinklerdelmic: You're right, I didn't realize that you put `wait_for_event` outside the lock. Then it makes complete sense.

philipwinklerdelmic:
Pushed changes.
The code should be tested again on the hardware (especially the scan function).

src/odemis/driver/smaract.py
pieleric: What I mean, is that for now, `.referenced` is *always* set to `False` at init. If the hardware has already been referenced + validated, and then backend restarted, it'd be nice to immediately set .referenced to `True`. Can we do that by using `IsValid()`?

src/odemis/driver/smaract.py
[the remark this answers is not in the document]
pieleric: Indeed. Let's use this code. We might want to make this behaviour more officially supported in the back-end, and then use it in other drivers which do referencing at init.

src/odemis/driver/smaract.py
```diff
@@ -3173,3 +3174,1157 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+        """
+        core = SA_SIDLL()
+        b_len = 1024
+        buf = create_string_buffer(b_len)
+        core.SA_SI_EnumerateDevices(c_char_p(b""), buf, byref(c_size_t(b_len)))
+        # sometimes the same locator is returned multiple times, convert to set for unique values
+        locators = set(buf.value.decode('latin1').split("\n"))
+        devices = []
+        for locator in locators:
+            id = Picoscale._openConnection(locator)
```
pieleric, line 3640: You probably should surround this call by a try/except, and in case of HwError, just continue to the next locator.

src/odemis/driver/smaract.py
```diff
@@ -3084,3 +3085,1139 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+        returns: (set of str) locator strings
+        """
+        core = SA_SIDLL()
+        b_len = 1024
+        buf = create_string_buffer(b_len)
+        core.SA_SI_EnumerateDevices(c_char_p(b""), buf, byref(c_size_t(b_len)))
+        locators = buf.value.decode('latin1').split("\n")
+        return set(locators)  # sometimes the same locator is returned multiple times, convert to set for unique values
+
+    @staticmethod
```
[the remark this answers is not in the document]
pieleric: Fine. Note that an alternative would be to have a subclass of Picoscale (eg _PicoscaleScanned) with a much simpler __init__(), so that you directly read the name and channel numbers using the standard wrapper functions.

src/odemis/driver/smaract.py
```diff
@@ -3084,3 +3085,1139 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+                usb:sn:<serial_number>
+                network:sn:<serial_number>
+        returns: (int) device handle for API functions
+        """
+        if locator == "fake":
+            return 0
+
+        # TODO: try multiple times, first time sometimes doesn't work
+        logging.debug("Connecting to locator %s.", locator)
+        core = SA_SIDLL()
```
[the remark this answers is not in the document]
pieleric: Still, I'd rather try to reuse the same core as in the init, firstly to make sure the simulator is also using this code, and secondly to avoid opening twice the DLL at init.

pieleric (changes requested):
Mostly follow-up on the previous discussions.

src/odemis/driver/smaract.py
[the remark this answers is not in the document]
philipwinklerdelmic: How about this: I could move the "core" part into the `_openConnection` function and return it with the id:
``` self.core, self._id = self._openConnection(locator)```
And in `_openConnection(locator)`:
```
        if locator == "fake":
            core = FakePicoscale_DLL()
            return core, 0
        else:
            core = SA_SIDLL()
         ...
         return core, id
```

**Comment by philipwinklerdelmic** [comment]
@pieleric I added the _PicoscaleScanned class. Can you have a look if it's what you had in mind? I still need to open an additional `core` in the `scan` function to get all the locator strings, which is a little ugly, but I don't think it's a big deal.

File: src/odemis/driver/smaract.py [hunk 20]
```diff
@@ -3173,3 +2961,1175 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+                # Activate working distance
+                self.SetProperty_i32(SA_PS_SYS_WORKING_DISTANCE_ACTIVATE_PROP,
+                                     SA_PS_WORKING_DISTANCE_SHRINK_MODE_LEFT_RIGHT)
+            # attribute is write-only and there is no corresponding event type (yet some waiting time is necessary)
+            # --> wait 1 s for command to be processed
+            startt = time.time()
+            while time.time() - startt < 1:
+                if future._must_stop:
+                    future._was_stopped = True
+                time.sleep(0.01)
```
pieleric, line 3765: If it's just 1 second, just put `time.sleep(1) in the lock, and be done.

src/odemis/driver/smaract.py
```diff
@@ -3173,3 +2961,1175 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+        """
+        if timeout == float("inf"):
+            tend = None
+        else:
+            tend = time.time() + timeout
+        ev = SA_SI_Event()
+
+        state = None
+        while state != end_state:
+            if tend:
```
pieleric, line 3708: For debugging, it might be useful to log here that an event was received but discarded. I'd suggest something like `if state != None: logging.debug("Skipped event 0x%x")` .

src/odemis/driver/smaract.py
```diff
@@ -3173,3 +2961,1175 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+                t = int((tend - time.time()) * 1000)  # SA_SI_WaitForEvent accepts timeout in ms
+                if t < 0 or ev.error == SA_SI_ERROR_TIMEOUT:
+                    raise TimeoutError()
```
pieleric, line 3711: I'd just move the `ev.error` check just after the call to SA_SI_WaitForEvent(), to make it clearer when this error comes from, and also avoid duplicate event log on timeout.

src/odemis/driver/smaract.py
```diff
@@ -3173,3 +2961,1175 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+            # receives updates both values are already updated.
+            self.referenced.notify(self.referenced.value)
+
+            # The adjustment state should be "disabled" after automatic adjustment. If the referencing
+            # procedure was stopped prematurely, right after it was set to "manual", it might be in the
+            # wrong state --> make sure it's set to disabled here.
+            if self.GetAdjustmentState() != SA_PS_ADJUSTMENT_STATE_DISABLED:
+                self.SetProperty_i32(SA_PS_AF_ADJUSTMENT_STATE_PROP,
+                                     SA_PS_ADJUSTMENT_STATE_DISABLED)
+                self._wait_for_event(SA_PS_ADJUSTMENT_STATE_DISABLED)  # state will be DISABLED when done
```
pieleric, line 3818: For safety, I'd suggest to always put some timeout. It can be large values (eg, 10s, 5 min)... but they end up always much smaller than infinity.

src/odemis/driver/smaract.py
```diff
@@ -3173,3 +2961,1175 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+        self.SetProperty_i32(SA_PS_AF_ADJUSTMENT_RESULT_SAVE_PROP, SA_SI_ENABLED)
+        # There is no documentation on loading/saving the configuration in the manual, even though
+        # the functions are available in the header file. It is not clear how much time saving the
+        # parameters takes. It can be assumed that it's fast --> wait 1 s.
+        time.sleep(1)
+
+    def _wait_for_event(self, end_state, timeout=float("inf")):
+        """
+        Blocks until event is triggered or timeout.
+        end_state (int): state of event variable to wait for
```
pieleric, line 3697: It's a little odd to wait only for the state, without check the event type. Pretty much all the states are worth either 0 or 1, so you might end up matching the wrong event (which has a state with the same constant value). Maybe also add a event type?

src/odemis/driver/smaract.py
```diff
@@ -3173,3 +2961,1175 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+                if future._must_stop:
+                    future._was_stopped = True
+                # Reset reference so that if it fails, it states the axes are not referenced (anymore)
+                self.referenced._value = {a: False for a in self._channels.keys()}
+                logging.debug("Starting referencing.")
+                # Cannot go immediately to automatic adjustment --> first switch to manual adjustment
+                self.EnableEventNotification(SA_PS_AF_ADJUSTMENT_PROGRESS_EVENT)
+                self.SetProperty_i32(SA_PS_AF_ADJUSTMENT_STATE_PROP,
+                                     SA_PS_ADJUSTMENT_STATE_MANUAL_ADJUST)
+            self._wait_for_event(SA_PS_ADJUSTMENT_STATE_MANUAL_ADJUST)
```
pieleric, line 3752: Does it return if Cancel() was called during the wait? I guess not.... It should either return or raise a CancelledError (if there is a special event corresponding).

pieleric (changes requested):
The scanning looks fine now.
I still have comments, mostly wrt the referencing.

src/odemis/driver/smaract.py
philipwinklerdelmic: I set it up, so we're only ever waiting for one event.  There was a problem with the event type, I think sometimes the hardware returned event type 0 even though it was a completely different event.

src/odemis/driver/smaract.py
philipwinklerdelmic: It depends on what was canceled. If the event is the IS_STABLE event and I'm canceling a referencing procedure, then the waitforevent function will block forever.

src/odemis/driver/smaract.py
philipwinklerdelmic: I wanted to keep it as similar as possible to the waitForEvent functions: call the function in the moving_lock and wait outside (also allow canceling without having to wait a second)

src/odemis/driver/smaract.py
philipwinklerdelmic: It is now working the way you requested. The problem was that I didn't raise CancelledError in enough places.

philipwinklerdelmic:
I think it should be ready to merge now. Just tested on the hw and it seems ok.

src/odemis/driver/smaract.py
```diff
@@ -3699,26 +3926,38 @@ def _wait_for_event(self, end_state, timeout=float("inf")):
...
+            if t < 0:
+                raise TimeoutError("Timeout limit of %s s exceeded." % timeout)
+            t = c_uint(t)
 
             # Don't check event type, it's not always accurate.
             # Instead, only one event is active and all other events are disabled, so it is clear
             # which event we are waiting for.
             self.core.SA_SI_WaitForEvent(self._id, byref(ev), t)
             state = ev.devEventParameter & 0xffff
+            if ev.error != SA_SI_ERROR_NONE:
```
pieleric, line 3949: I think you got confused by the use of union. Check the section 7 of the programmers' guide.
*Only* if the ev.type is `SA_SI_STREAM_ABORTED_EVENT`, then ev.error is meaningful.
For the SA_PS_AF_ADJUSTMENT_PROGRESS_EVENT, the state is lowest 16-bits of the devEventParameter. That's really specific of this event.
I'd suggest to handle each event type explicitly (with a generic else where you log of the event type which is not handled).

Right now, your function is tailored to wait for the ADJUSTMENT_PROGRESS_EVENT, so you could even rename it `wait_for_progress_event()`. In this case, I would highly suggest to move the small part corresponding to wrapping `SA_SI_WaitForEvent()` to a separate function, which converts the timeout, and return the SA_SI_Event received, or raise a TimeoutError.

src/odemis/driver/smaract.py
```diff
@@ -4043,8 +4282,11 @@ def SA_SI_WaitForEvent(self, handle, event, timeout):
         ev = _deref(event, SA_SI_Event)
         if self.cancel_event:
             self.cancel_event = False
-            ev.error = SA_SI_ERROR_CANCELLED
             raise SA_SIError(SA_SI_ERROR_CANCELLED, "CANCELLED")
+
+        # It's not possible to add ev.error here. Somehow, changing one anonymous parameter of
+        # SA_SI_Event (such as ev.devEventParameter) causes all other anonymous parameters
+        # (ev.error) to store the same value.
```
pieleric, line 4289: That's because it's an "anonymous" parameter, but a "union". In C, a union means that each member/attribute shares the same memory space. So, of course, changing one changes the other ones. So I don't think this comment is useful here, because it'd be impossible for the C library to change both. Either the error is set with the event type STREAM_ABORTED_EVENT, or a different type is set and the `devEventParameter` is used.

src/odemis/driver/smaract.py
```diff
@@ -4043,8 +4282,11 @@ def SA_SI_WaitForEvent(self, handle, event, timeout):
         ev = _deref(event, SA_SI_Event)
```
pieleric, line 4282: Note that this function actually doesn't really behaves as the real C library function, as it doesn't block. That said, for the current behaviour that might be enough. Maybe add a tiny sleep(0.01) in case there no cancellation to make sure we don't use the CPU at full throttle during simulated calibration.

src/odemis/driver/smaract.py
```diff
@@ -4018,6 +4255,8 @@ def __init__(self):
             SA_PS_STABLE_STATE_CHANGED_EVENT: SA_PS_SYS_IS_STABLE_PROP,
                                   }
 
+        self.err_code = {}
+
```
pieleric, line 4259: What is this for? I feel it's some kind of hack for the simulator to work. At least add a comment how this works, or fix the simulator to behave more like the normal library.

src/odemis/driver/test/actuator_test.py
```diff
@@ -1592,5 +1593,82 @@ def test_cancel_reference(self):
...
+        self.dev.position.unsubscribe(on_rotation)
+
+    def test_calculation(self):
+        """
+        Test the calculation of the position and angle.
+        """
+        # Force sensor positions in simulator
+        self.dev.sensor.core.position0 = 0
+        self.dev.sensor.core.position1 = 1e-6
+        self.dev.sensor.core.position2 = 0
```
pieleric, line 1653: That's probably broken now that .position is a list. If your test still passes, there is also a problem with you test checks.

src/odemis/driver/test/smaract_test.py
```diff
@@ -500,5 +502,101 @@ def test_reference(self):
...
+        time.sleep(0.1)
+        self.assertTrue(f._was_stopped)
+
+        # Test queued futures
+        f1 = self.dev.reference()
+        f2 = self.dev.reference()
+        f1.cancel()
+        f2.cancel()
+        time.sleep(0.1)
+        self.assertEqual(self.dev._executor._queue, deque([]))
```
pieleric, line 589: Let's just check that `len(self.dev._executor._queue) == 0`, or is there some extra magic of comparing to an empty deque?

src/odemis/driver/test/smaract_test.py
```diff
@@ -500,5 +502,101 @@ def test_reference(self):
...
+        """
+        while self.dev.state == model.ST_STARTING:
+            time.sleep(0.1)
+        self.pos_update = False
+
+        def pos_listener(_):
+            self.pos_update = True
+
+        self.dev.position.subscribe(pos_listener)
+        if CONFIG_Picoscale['locator'] == 'fake':
```
pieleric, line 553: Let's just explicitly write `if TEST_NOHW`.

pieleric (changes requested):
It looks better. I think that you got tricked by how the events should be handled. If you handle then explicitly .type -> union then the code would make a lot more sense and be more reliable.
See my comments for more explanation.

src/odemis/driver/smaract.py
```diff
@@ -3699,26 +3926,38 @@ def _wait_for_event(self, end_state, timeout=float("inf")):
         timeout (float): maximum time to wait in s. If inf, it will wait forever.
         """
         if timeout == float("inf"):
-            tend = None
-        else:
-            tend = time.time() + timeout
+            timeout = 600  # 10 minutes
```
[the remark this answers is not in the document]
philipwinklerdelmic: One of your last comments was to change this from infinity to a value like 10 minutes!

src/odemis/driver/smaract.py
```diff
@@ -3173,3 +3174,1209 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+                    raise TimeoutError("Picoscale reported timeout error.")
+                elif ev.error == SA_SI_ERROR_CANCELLED:
+                    raise CancelledError("WaitForEvent was cancelled.")
+                else:
+                    if ev.error in self.core.err_code:
+                        # Don't raise HwError, most of the time, the error is not meaningful
+                        logging.error("Exception %s during WaitForEvent: %s."
+                                      % (ev.error, self.core.err_code[ev.error]))
+                    else:
+                        logging.error("Unknown error code '%s'." % ev.error)
```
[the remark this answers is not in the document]
philipwinklerdelmic: In the wait function I get errors all the time. I'm not sure why, maybe I set up the class that's passed by reference in a wrong way, or maybe it's a hardware issue. If that bothers you, we have to check with Smaract.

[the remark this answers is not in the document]
philipwinklerdelmic: Wait, I just saw your comment below. ev.error is probably not meaningful.

src/odemis/driver/smaract.py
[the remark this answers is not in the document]
pieleric: Yes, but I meant, that the caller should set the timeout to 10 min. If the timeout argument is infinity, it should mean infinity, not 10 min! => the function behaviour shouldn't change, just the callers.

philipwinklerdelmic:
@pieleric Can you have another look? The biggest change is that I don't allow position requests before/during referencing. I hope that's working properly.
I also haven't tested the last commit (error passing in WaitForEvent) on the hardware, first wanted to see if you had any other comments.

src/odemis/driver/smaract.py
```diff
@@ -3173,3 +3174,1239 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+        time.sleep(1)
+
+    def _wait_for_progress_event(self, end_state, timeout=600):
+        """
+        Blocks until progress event is triggered or timeout.
+        It is assumed that only one event type is enabled. The function will wait until the
+        .devEventParameter contains the end_state. This can mean different things depending on the event.
+        end_state (int): state of event variable to wait for
+        timeout (float): maximum time to wait in s. If inf, it will wait forever.
+        """
```
pieleric, line 3960: Add info about what kind of exceptions can be raised.

src/odemis/driver/smaract.py
```diff
@@ -3173,3 +3174,1239 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+                t = tend - time.time()
+                if t < 0:
+                    raise TimeoutError("Timeout limit of %s s exceeded." % timeout)
+            else:
+                t = float("inf")
+            ev = self.WaitForEvent(t)
+            # Don't check event type, it's not always accurate.
+            # Instead, only one event is active and all other events are disabled, so it is clear
+            # which event we are waiting for.
+            state = ev.devEventParameter & 0xffff
```
pieleric, line 3982: You *have* to check the event type. The rest of the content of `ev` has completely different meaning depending on the event type. So "state" has only a meaning when the even type is for progress event.
I'd suggest to write something like:
```python
ev = self.WaitForEvent(t)  # raise TimeOut/CancelledError by itself
if ev.type == SA_PS_AF_ADJUSTMENT_PROGRESS_EVENT:
      # lowest 16-bits is the adjustment state
      state = ev.devEventParameter & 0xffff
elif ev.type == SA_SI_STREAM_ABORTED_EVENT:
    raise SA_SIError(ev.error, ...)
else:
    logging.debug("Skipped event 0x%x" % ev.type)
```

src/odemis/driver/smaract.py
```diff
@@ -3173,3 +3174,1239 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+            CancelledError: function was cancelled by SA_SI_Cancel
+            SA_SI_Error: something went wrong
+        """
+        if timeout == float("inf"):
+            t = SA_SI_TIMEOUT_INFINITE
+        else:
+            t = c_uint(int(timeout * 1000))  # SA_SI_WaitForEvent accepts timeout in ms
+
+        ev = SA_SI_Event()
+        ret = self.core.SA_SI_WaitForEvent(self._id, byref(ev), t)
```
pieleric, line 3909: That's not going to work, because `sp_errcheck()` already checks the return code and will raise an exception. So instead, your code should look like:
```python
try:
   self.core.SA_SI_WaitForEvent(self._id, byref(ev), t)
except SA_SIError as ex:
    if ex.errno == SA_SI_ERROR_TIMEOUT:
       raise TimeoutError(...)
    elif ex.errno == SA_SI_ERROR_CANCELLED:
       raise CancelledError(...)
    raise
```

src/odemis/driver/smaract.py
```diff
@@ -3173,3 +3174,1239 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+        """
+        property_key (int32): property key symbol
+        data_source_idx:
+        returns (float) the value
+        """
+        ret_val = c_double()
+        self.core.SA_SI_GetValue_f64(self._id, channel, data_source_idx, byref(ret_val))
+        return ret_val.value
+
+    def WaitForEvent(self, timeout=600):
```
pieleric, line 3893: 600 as a default is odd. I'd say either don't put any default, or put infinity.

src/odemis/driver/smaract.py
```diff
@@ -3173,3 +3174,1239 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+        Saves the configuration parameters (cf _load_configuration) and internal parameters of the
+        referencing procedure. This function is blocking.
+        """
+        self.SetProperty_i32(SA_PS_AF_ADJUSTMENT_RESULT_SAVE_PROP, SA_SI_ENABLED)
+        # There is no documentation on loading/saving the configuration in the manual, even though
+        # the functions are available in the header file. It is not clear how much time saving the
+        # parameters takes. It can be assumed that it's fast --> wait 1 s.
+        time.sleep(1)
+
+    def _wait_for_progress_event(self, end_state, timeout=600):
```
pieleric, line 3953: I'd suggest to put the default timeout to infinity, as you used to. 10min is really arbitrary, so it'd be much more readable to have it explicitly written in every place you call this function (and you want a 10 min timeout). Otherwise, it's really hard to guess when you read the caller's code that this function times out after 10 min.

src/odemis/driver/smaract.py
```diff
@@ -3173,3 +3174,1239 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+        It is assumed that only one event type is enabled. The function will wait until the
+        .devEventParameter contains the end_state. This can mean different things depending on the event.
+        end_state (int): state of event variable to wait for
+        timeout (float): maximum time to wait in s. If inf, it will wait forever.
+        """
+        if timeout == float("inf"):
+            tend = None
+        else:
+            tend = time.time() + timeout
+        ev = SA_SI_Event()
```
pieleric, line 3965: You can drop this line.

src/odemis/driver/test/smaract_test.py
```diff
@@ -21,6 +21,8 @@
 '''
 import logging
 import os
+from collections import deque
+
```
pieleric, line 25: Unused => drop

pieleric (changes requested):
It looks quite better. I still have some comment on the way the events are handled, but hopefully we can sort this out easily now.

src/odemis/driver/smaract.py
philipwinklerdelmic: I really don't understand your previous comment then. You told me to **not** put infinity, but a value like 600. What do you mean with that?

src/odemis/driver/smaract.py
philipwinklerdelmic: I also use the function to check the STABLE_STATE_CHANGED event. Most events seem to work in the same way (cf Table 7, p 223). I think I tried to use the type in the beginning, but then it turned out that it wasn't always correct with all events. Now I'm making much less use of the events, so maybe it will work, but I need to check on the hardware. If it doesn't I should probably report it to Smaract.

src/odemis/driver/smaract.py
pieleric: In the function *signatures*, ie, here and line 3953, you set the default to infinity (or no default).
In the function *callers*, l 4018, 4037, 4083, you explicitly indicate a timeout. That timeout could be around 10 min, or less now that we saw a properly configured hardware is faster.

src/odemis/driver/smaract.py
```diff
@@ -3173,3 +3174,1244 @@ def SA_CTL_Move(self, handle, ch, pos_pm, _):
...
+
+            ev = self.WaitForEvent(t)  # raise TimeOut/CancelledError by itself
+            if ev.type == SA_PS_AF_ADJUSTMENT_PROGRESS_EVENT:
+                # lowest 16-bits is the adjustment state
+                state = ev.devEventParameter & 0xffff
+            elif ev.type == SA_PS_STABLE_STATE_CHANGED_EVENT:
+                # new state, 0 means unstable, 1 means stable
+                state = ev.devEventParameter
+            elif ev.type == SA_SI_STREAM_ABORTED_EVENT:
+                raise SA_SIError(ev.error, self.core.err_code[ev.error])
```
pieleric, line 3985: You *might* want to raise a CancelledError in some cases... but let's check, if it's ever needed anyway.

pieleric (changes requested):
Yes, this way of handling events looks way better!

PR #1518, opened 2020-09-30
Smaract handle deactive position after referencing

amuskens:


src/odemis/driver/smaract.py
```diff
@@ -1541,9 +1548,16 @@ def _doReference(self, future):
...
-            if self._is_referenced():
+            # only move if no exception was raised.
+            if self._is_referenced() and not raised:
                 self.referenced._value = {a: True for a in self.axes.keys()}
                 self._updatePosition()
+                if self._deactive_pos_after_ref is not None:
+                    logging.info("Moving axes to deactivated position %s after referencing", self._deactive_pos_after_ref)
+
+                    self._checkMoveAbs(self._deactive_pos_after_ref)
+                    f = self._createMoveFuture()
```
pieleric, line 1559: BTW, could you remove the "ref" argument from `_createMoveFuture()`, as it's unused.

src/odemis/driver/smaract.py
```diff
@@ -2931,6 +2955,15 @@ def _doReference(self, future, axes):
                         logging.warning("Axis %s not referenced after the end of referencing", a)
                         # TODO: Raise some error here
 
+                    # if referenced, move to the safe posiiton (if defined)
+                    if self._deactive_pos_after_ref is not None and self._deactive_pos_after_ref.get(a) is not None:
```
pieleric, line 2959: `self._deactive_pos_after_ref.get(a) is not None` -> `a in self._deactive_pos_after_ref`

src/odemis/driver/smaract.py
```diff
@@ -2931,6 +2955,15 @@ def _doReference(self, future, axes):
...
+                        deactive_pos = self._deactive_pos_after_ref[a]
+                        logging.info("Moving axis %s to deactivated position %f", a, deactive_pos)
+                        f = self._createMoveFuture()
+                        self._doMoveAbs(f, {a: deactive_pos})
```
pieleric, line 2963: Probably you can also reuse the future here.

src/odemis/driver/smaract.py
```diff
@@ -1240,6 +1240,7 @@ def __init__(self, name, role, locator, axes, ref_on_init=False, linear_speed=0.
         # VA dict str(axis) -> bool
         self.referenced = model.VigilantAttribute(axes_ref, readonly=True)
         # If ref_on_init, referenced immediately.
+        self._deactive_pos_after_ref = None
```
pieleric, line 1243: Please, don't make it automatic. There should still be an argument in `__init__` called something like `pos_deactive_after_ref` to indicate the user wants this behaviour. 
That's needed for both drivers.

pieleric (changes requested):
I think using the DEACTIVE metadata makes indeed more sense. I've given some hints on how to support the cancelling, and a few extra recommendations.

src/odemis/driver/smaract.py
amuskens: Hmm.. when I tested it, it gets stuck here if you use the same future

amuskens:
Done

src/odemis/driver/test/smaract_test.py
```diff
@@ -332,6 +333,21 @@ def test_pivot_set(self):
             self.dev.updateMetadata({model.MD_PIVOT_POS: old_pivot})
             self.dev.moveRelSync({"x": 0})
 
+    def test_reference_and_deactivate_move(self):
+        # Set a deactive position and check to be sure that the controller moves to this location
+        # after a reference move
+        # TODO: UIse real physical deactiva position
+        de_pos = {'x': 3.5801e-4, 'y': 0, 'z': 1e-3, 'rx':-1.2e-6, 'rz':-1.453e-6}
```
pieleric, line 340: Let's drop rz, to check that it works even if not all axes are present.

src/odemis/driver/test/smaract_test.py
```diff
@@ -332,6 +333,21 @@ def test_pivot_set(self):
             self.dev.updateMetadata({model.MD_PIVOT_POS: old_pivot})
             self.dev.moveRelSync({"x": 0})
 
+    def test_reference_and_deactivate_move(self):
+        # Set a deactive position and check to be sure that the controller moves to this location
+        # after a reference move
+        # TODO: UIse real physical deactiva position
```
pieleric, line 339: It'd be difficult to find a position which works on any system any way. Let's drop this TODO.

src/odemis/driver/smaract.py
```diff
@@ -2931,6 +2956,16 @@ def _doReference(self, future, axes):
                         logging.warning("Axis %s not referenced after the end of referencing", a)
                         # TODO: Raise some error here
 
+                    # if referenced, move to the safe posiiton (if defined)
```
pieleric, line 2959: posiition - > position

src/odemis/driver/smaract.py
```diff
@@ -2931,6 +2956,16 @@ def _doReference(self, future, axes):
...
+                    if self._pos_deactive_after_ref and self._deactive_pos_after_ref is not None \
+                        and a in self._deactive_pos_after_ref:
+                        deactive_pos = self._deactive_pos_after_ref[a]
+                        logging.info("Moving axis %s to deactivated position %f", a, deactive_pos)
+                        f = self._createMoveFuture()
+                        self._doMoveAbs(f, {a: deactive_pos})
```
pieleric, line 2965: As it is, it's going to move each axis serially. It's quite slow, and will have a strange global movement if there are several axes. So I'd recommend to only call _doMoveAbs(), with all the axes which should be moved. => Copy the deactivate pos, with only the axes present in the referencing.

src/odemis/driver/smaract.py
```diff
@@ -2931,6 +2956,16 @@ def _doReference(self, future, axes):
                         logging.warning("Axis %s not referenced after the end of referencing", a)
                         # TODO: Raise some error here
 
+                    # if referenced, move to the safe posiiton (if defined)
+                    if self._pos_deactive_after_ref and self._deactive_pos_after_ref is not None \
+                        and a in self._deactive_pos_after_ref:
+                        deactive_pos = self._deactive_pos_after_ref[a]
+                        logging.info("Moving axis %s to deactivated position %f", a, deactive_pos)
+                        f = self._createMoveFuture()
```
pieleric, line 2964: It's not a good idea to create a separate Future as it prevents the cancellation from working. I guess it doesn't work to use the same Future only due to the lock. If you convert _moving_lock from Lock() to RLock(), this should be fine.

src/odemis/driver/smaract.py
```diff
@@ -2931,6 +2956,16 @@ def _doReference(self, future, axes):
                         logging.warning("Axis %s not referenced after the end of referencing", a)
                         # TODO: Raise some error here
 
+                    # if referenced, move to the safe posiiton (if defined)
+                    if self._pos_deactive_after_ref and self._deactive_pos_after_ref is not None \
```
pieleric, line 2960: See my comment on the 5DoF concerning using the metadata directly + one variable.

src/odemis/driver/smaract.py
```diff
@@ -2345,7 +2359,9 @@ def __init__(self, name, role, locator, ref_on_init=False, axes=None, speed=1e-3
         # Not to be mistaken with axes which is a simple public view
         self._axis_map = {}  # axis name -> axis number used by controller
         axes_def = {}  # axis name -> Axis object
+        self._axis_safe_pos = {}  # axis_name -> float (position to move after reference)
```
pieleric, line 2362: Variable not used => drop.

src/odemis/driver/smaract.py
```diff
@@ -1541,6 +1552,7 @@ def _doReference(self, future):
         finally:
             # We only notify after updating the position so that when a listener
             # receives updates both values are already updated.
+            # only move if no exception was raised.
```
pieleric, line 1555: Drop/move this comment.

src/odemis/driver/smaract.py
```diff
@@ -1522,6 +1528,11 @@ def _doReference(self, future):
 
                 logging.info("Referencing successful.")
 
+            if self._pos_deactive_after_ref and self._deactive_pos_after_ref is not None and self._is_referenced():
```
pieleric, line 1531: Having two variables with almost the same names is really confusing. I'd recommend to not have a dedicated `_deactive_pos_after_ref` variable. Instead, directly read it in the metadata:
`if self._pos_deactive_after_ref and model.MD_FAV_POS_DEACTIVE in self._metadata ...`

src/odemis/driver/smaract.py
```diff
@@ -1272,7 +1275,11 @@ def updateMetadata(self, md):
             #   or update the current position, based on the new pivot point.
             logging.debug("Updating pivot point to %s.", pivot)
             self.SetPivot(pivot)
-
+        if model.MD_FAV_POS_DEACTIVE in md:
+            deactive_pos = md[model.MD_FAV_POS_DEACTIVE]
+            if not isinstance(deactive_pos, dict) or not set(deactive_pos.keys()) <= set(self.axes.keys()):
+                raise ValueError("Invalid metadata, should be a coordinate dictionary but got %s." % (deactive_pos,))
+            self._deactive_pos_after_ref = deactive_pos
```
pieleric, line 1282: You should be able to drop this line.

src/odemis/driver/smaract.py
```diff
@@ -1255,7 +1259,6 @@ def __init__(self, name, role, locator, axes, ref_on_init=False, linear_speed=0.
         self.rotary_speed = rotary_speed
         self.set_rotary_speed(math.degrees(self.rotary_speed))
         self.set_hold_time(hold_time)
-        self._updatePosition()
```
pieleric, line 1258: Why dropping this line? Please put it back.

pieleric (changes requested):

src/odemis/driver/test/smaract_test.py
amuskens: So... it doesn't work, but this is because of the assert_pos_almost_equal requiring the dict keys to match.

src/odemis/driver/smaract.py
```diff
@@ -1481,7 +1485,7 @@ def reference(self, _=None):
         reference usually takes axes as an argument. However, the SA_MC references all
         axes together so this argument is extraneous.
         """
-        f = self._createMoveFuture(ref=True)
+        f = self._createMoveFuture()
         self._executor.submitf(f, self._doReference, f)
         return f
```
pieleric, line 1490: Isn't this the same as in your other PR? It's going to conflict...

src/odemis/driver/smaract.py
```diff
@@ -1215,6 +1215,7 @@ def __init__(self, name, role, locator, axes, ref_on_init=False, linear_speed=0.
                 raise model.HwError("Failed to find device, check it is connected and turned on")
             raise
         logging.debug("Successfully connected to SA_MC Controller ID %d", self._id.value)
+        self._deactive_pos_after_ref = None
```
pieleric, line 1218: I meant the *opposite*. Keep the `pos_deactive_after_ref` boolean flag + the argument. However, you don't need to store the DEACTIVE value when the metadata changes, because it's already in the metadata.

pieleric (changes requested):

PR #1520, opened 2020-10-05
small tweaks to unit display

pieleric:


src/odemis/util/test/units_test.py
```diff
@@ -99,6 +99,8 @@ def test_readable_str(self):
                   ((999.4, "V", 3), "999 V"),
                   ((200e-6, "m"), u"200 µm"),
                   ((0.0, "m"), "0 m"),
+                  ((0, "rad"), "0 rad"),
+                  ((0, "rad", 3), "0 rad"),
```
Sabrina85, line 103: You could add a test with a decimal instead of `0`. Or is this something we never use? What about `mrad` - do we display this?

src/odemis/gui/comp/text.py
```diff
@@ -896,7 +896,9 @@ def _display_pretty(self):
         else:
             str_val = units.readable_str(self._number_value, self.unit, self.accuracy)
         # Get the length of the number (string length, minus the unit length)
-        number_length = len(str_val.rstrip(string.ascii_letters + u" µ"))
+        number_length = str_val.find(" ")
```
Sabrina85, line 899: I suggest to add a very short comment explaining that number and unit are usually separated by a space. Or update the comment in the line above. 
`# number and unit usually separated by space`

Sabrina85 (changes requested):
Minor comments.

src/odemis/util/test/units_test.py
pieleric: We have plenty of tests with decimals. I wanted more tests with 0, as it's a bit special (ie, very tiny value, but no unit prefix).
The GUI does disaplay the prefix in some cases (and that's why I originally wrote this test case), but this is directly handled into the number-text widget.

pieleric:
Updated the PR.

Sabrina85 (approved):
Looks good!

PR #1526, opened 2020-10-08
Plugin: Merging RGB channels

VictoriaDel:


pieleric (changes requested):
Looks good. Just a few minor comments.

plugins/merge_RGB.py
```diff
@@ -0,0 +1,314 @@
...
+        """
+        Update the stream after shifting it by the given value.
+        Args:
+            stream(StaticFluoStream): stream to be shifted
+            i(int): index to show at which dimension the stream is to be shifted
+            value(ContinuousVA): shift values in meters
+        """
+        logging.debug("New shift = %f on stream %s", value, stream.name.value)
+        poscor = stream.raw[0].metadata.get(model.MD_POS_COR, (0, 0))
+        px_size = stream.raw[0].metadata.get(model.MD_PIXEL_SIZE)
```
pieleric, line 272: You have to provide some default value for the PIXEL_SIZE. Otherwise, you might get `None`, which will mess up the rest.
I'd suggest to extend `_ensureRGB()` so that it detects there is no such metadata, and in such case add 1e-9x1e-9 (as before). Then, here you can just write `px_size = stream.raw[0].metadata[model.MD_PIXEL_SIZE]`.

plugins/merge_RGB.py
```diff
@@ -0,0 +1,314 @@
+# -*- coding: utf-8 -*-
+'''
+Created on 5 October 2020
+
+@author: Victoria Mavrikopoulou
+
+Gives ability to manually change the overlay-metadata.
```
pieleric, line 7: Please update this description.

plugins/merge_RGB.py
```diff
@@ -0,0 +1,314 @@
...
+
+This is free and unencumbered software released into the public domain.
```
pieleric, line 9: Drop this line, this is left-over from the public domain license.

plugins/merge_RGB.py
```diff
@@ -0,0 +1,314 @@
...
+        self._stream_blue = self._openImage(filenameB, TINT_BLUE, self.blueShiftX, self.blueShiftY)
+        self._storeDir(filenameB)
+
+    def _storeDir(self, fn):
+        """Store the directory of the given filename so as the next filename is in the same place"""
+        path, bn = os.path.split(fn)
+        files = [self.filenameR, self.filenameG, self.filenameB]
+        for se in self._dlg.setting_controller.entries:
+            if se.vigilattr in files:
+                logging.debug("Found file browser %s", se.value_ctrl)
```
pieleric, line 183: Let's drop this debug line, now that we know it works.

pieleric (changes requested):
Looks fine. Just very minor comments.

plugins/merge_RGB.py
```diff
@@ -0,0 +1,316 @@
...
+        va_on_shift = functools.partial(self._onShift, stream, index)
+        self._subscribers.append(va_on_shift)
+        vashift.subscribe(va_on_shift)
+
+    def _removeStream(self, st):
+        """Remove the given stream since another one is loaded from the user for display"""
+        sconts = self._dlg.streambar_controller.stream_controllers
+        for sc in sconts:
+            if sc.stream is st:
+                sc.stream_panel.on_remove_btn(st)
```
pieleric, line 254: Also need to `del self._raw_orig[st]`.

plugins/merge_RGB.py
```diff
@@ -0,0 +1,316 @@
...
+        self.filenameB.value = " "
+        self.redShiftX.value = 0
+        self.redShiftY.value = 0
+        self.greenShiftX.value = 0
+        self.greenShiftY.value = 0
+        self.blueShiftX.value = 0
+        self.blueShiftY.value = 0
+        self.cropBottom.value = 0
+        self._subscribers = []
+        self._dlg = None
```
pieleric, line 149: Also need to reset raw_orig.

pieleric (changes requested):
Sorry, I just noticed one more small issue: raw_orig is never reset, so it's going to "leak memory" by holding a reference to every stream and raw data ever opened.
Otherwise, everything else looks fine 🙂

PR #1527, opened 2020-10-08
Bounding box checks if image is present

VictoriaDel:


src/odemis/acq/stream/_base.py
```diff
@@ -1261,7 +1261,10 @@ def getBoundingBox(self, im=None):
             ValueError: If the stream has no (spatial) data and stream's image is not defined
         """
         if im is None:
-            im = self.image.value
+            try:
+                im = self.image.value
+            except AttributeError:
+                im = None
```
Kleijwegt, line 1267: Maybe a logging message here that says : "no data is provided and the raw data of the stream is used" can be useful when debugging.

pieleric: Let's just keep it simple. This functions as multiple ways to get to the (meta)data, and it's rarely important to know how, as eventually it'll get the right one.

PR #1528, opened 2020-10-12
Handle AR background with multiple polarization modes

VictoriaDel:


src/odemis/acq/calibration.py
```diff
@@ -43,29 +43,35 @@ def get_ar_data(das):
     # TODO: also allow to pass an expected resolution, in order to support
     # a calibration file with multiple calibrations resolution?
 
-    # expect the worse: multiple AR data
-    ar_data = []
+    # expect the worse: multiple AR data (per polarization mode)
```
Sabrina85, line 46: worse -> worst

Sabrina85 (approved):
Looks good! 
Did you also check if we have a test case for this?
Reading my code here again, I see that there was a mistake in it (I guess that is why you updated the code :D ), which I think could have been caught by a proper test.

pieleric:
Victoria, could you have a look at the test cases, as suggested by Sabrina? I had a quick look at them, and I guess we broke the arpol test case, as it specifically checks that the returned arrays are in the same order (which is not anymore the case with the user of the dict, and not needed). I guess that changing to `self.assertEqual(set, set)` should solve this.

Also, could you add another test case, which check that only one AR data is returned when passing several AR data (without polarization mode).

src/odemis/acq/test/calibration_test.py
```diff
@@ -145,15 +145,21 @@ def test_load_simple_arpol(self):
...
+            bg_list.append(calib.tolist())
 
         # Give one DA, the correct one, so expect to get it back
         out = calibration.get_ar_data(bg_data)
-        numpy.testing.assert_equal(out, bg_data)  # all bg images should have same shape
+        out_list = []
+        for o_data in out:
+            out_list.append(o_data.tolist())
+
+        self.assertEquals(bg_list, out_list, msg="compare lists ignoring the order")
```
pieleric, line 162: I don't understand how this works, as this is still a list, so it still has a problem with the order. 
I'd recommend something like:
```python
# Check the expected and return lists contain the same DataArrays
self.assertEqual(len(bg_data), len(out))
for da in out:
    self.assertIn(da, bg_data)
```

src/odemis/acq/test/calibration_test.py
```diff
@@ -24,7 +24,7 @@
 import os
 import time
 import unittest
-from odemis.acq.stream import POL_POSITIONS
+from odemis.acq.stream import POL_POSITIONS, MD_POL_VERTICAL
```
pieleric, line 27: `MD_POL_VERTICAL ` doesn't seem used, is it?

pieleric (changes requested):
I'd recommend to do the checks a different way.

PR #1530, opened 2020-10-15
fix exceptions formatting

pieleric:
The regular batch of fixes for exceptions with a , instead of % for the
message formating.


Sabrina85 (approved):
Looks good!

PR #1533, opened 2020-10-20
util filename: fix test on dates where the day==year

pieleric:
On the 20-10-20, dateshrt was incorrectly guessed. Let's just use
dshrtrev, which is the first guess on such date.

### DIFF
--- a/src/odemis/util/test/filename_test.py
+++ b/src/odemis/util/test/filename_test.py
@@ -36,7 +36,11 @@ timelng = time.strftime("%H%M%S")
 # date_sl = time.strftime("%Y/%m/%d")
 # timelng_sl = time.strftime("%H/%M/%S")
 current_year = time.strftime("%Y")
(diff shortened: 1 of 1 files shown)


src/odemis/util/test/filename_test.py
```diff
@@ -36,6 +36,10 @@
 # date_sl = time.strftime("%Y/%m/%d")
 # timelng_sl = time.strftime("%H/%M/%S")
 current_year = time.strftime("%Y")
+dshrtrev = time.strftime("%d%m%y")
+# Note: cannot be used in tests, as on days similar to the year, such as 20-10-20,
+# it's not clear what is the day and what is the year, but the code convention is to
+# guess DDMMYY.
 dateshrt = time.strftime("%y%m%d")
```
tepals, line 43: This is now not used at all, so why not remove it?

pieleric: Fair enough... done.

PR #1534, opened 2020-10-20
plugin merge_RGB: small fixes

pieleric:


VictoriaDel (approved):
Good idea to add this exception. Thank you for fixing it!

PR #1535, opened 2020-10-21
Smaract feature update

amuskens:
driver smaract: updated cancellation of all running moves and state change on failed reference on init

### DIFF
--- a/src/odemis/driver/smaract.py
+++ b/src/odemis/driver/smaract.py
@@ -1254,7 +1254,7 @@ class MC_5DOF(model.Actuator):
             logging.debug("SA_MC is referenced")
         else:
             if ref_on_init:
-                self.reference().result()
+                self.reference()  # will reference now in background.
 
         # Use a default actuator speed
         self.linear_speed = linear_speed
@@ -2506,7 +2506,7 @@ class MCS2(model.Actuator):
             logging.debug("SA_CTL is referenced")
         else:
             if ref_on_init:
-                self.reference()
+                self.reference()  # will reference in background
             else:
                 logging.warning("SA_CTL is not referenced. The device will not function until referencing occurs.")
 
--- a/src/odemis/driver/smaract.py
+++ b/src/odemis/driver/smaract.py
@@ -1255,10 +1255,7 @@ class MC_5DOF(model.Actuator):
             logging.debug("SA_MC is referenced")
         else:
             if ref_on_init:
-                try:
-                    self.reference().result()
-                except Exception as e:
-                    self.state._set_value(e, force_write=True)
+                self.reference().result()
             else:
                 logging.warning("SA_MC is not referenced. The device will not function until referencing occurs.")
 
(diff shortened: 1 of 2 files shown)


pieleric (changes requested):

src/odemis/driver/smaract.py
```diff
@@ -1254,7 +1254,10 @@ def __init__(self, name, role, locator, axes, ref_on_init=False, linear_speed=0.
             logging.debug("SA_MC is referenced")
         else:
             if ref_on_init:
-                self.reference().result()
+                try:
+                    self.reference().result()
+                except Exception as e:
+                    self.state._set_value(e, force_write=True)
```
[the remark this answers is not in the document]
amuskens: Ok. Will this cause problems though? Because any operations after init will fail if the device is not referenced.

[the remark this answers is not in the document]
pieleric: The code in _updatePosition() seems to already handle this case properly.
Well, actually, looking at that code, I'd suggest to return an *empty* dict, instead of a dict with all positions at 0. It's clearer that there is no data, instead of pretending we know that it's at 0 (although it's probably wrong).

src/odemis/driver/smaract.py
```diff
@@ -1278,7 +1282,11 @@ def __init__(self, name, role, locator, axes, ref_on_init=False, linear_speed=0.
 
     def terminate(self):
         # should be safe to close the device multiple times if terminate is called more than once.
-        self.core.SA_MC_Close(self._id)
+        if not self._shutdown_flag:
+            self.stop()
+            self._executor.shutdown(cancel_futures=True)
+            self.core.SA_MC_Close(self._id)
+            self._shutdown_flag = True
```
pieleric, line 1289: I think you should be able to to away with the `_shutdown_flag`, by using `_executor is not None`. After shutting down the executor, set it to `._executor = None`.
(Same thing for both classes).

pieleric (changes requested):
A few more comments... and still need to be non-blocking at init.

src/odemis/driver/smaract.py
```diff
@@ -1255,8 +1255,6 @@ def __init__(self, name, role, locator, axes, ref_on_init=False, linear_speed=0.
         else:
             if ref_on_init:
                 self.reference().result()
```
pieleric, line 1257: Referencing shouldn't be blocking.
If referencing fails, I'm fine with the current implementation about changing the state directly inside the referencing function. So this means you don't need to put a callback. Just call `self.reference()` and add a comment that it's going to run in the background.

src/odemis/driver/smaract.py
```diff
@@ -2503,7 +2506,7 @@ def __init__(self, name, role, locator, ref_on_init=False, axes=None, speed=1e-3
             logging.debug("SA_CTL is referenced")
         else:
             if ref_on_init:
-                self.reference().result()
+                self.reference()
```
pieleric, line 2509: Add a comment that it'll run in the background.

pieleric (changes requested):

PR #1536, opened 2020-10-22
gui: fix support for time-correlator filters

pieleric:
With the local axes changes, the tc-filter and tc-od-filter were not
anymore properly handled.
Especially, the CL intensity can also be affected by the tc-od-filter,
but it doesn't contain a band axis (it contains a "density" axis), so
this would prevent the stream from being created.


amuskens (approved):

amuskens:
Let me know if the new version fixes the issue

PR #1537, opened 2020-10-22
plugins zstack: also handle if the speed VA doesn't know about Z

pieleric:
Some hardware provide a .speed VA, but it doesn't contain the info about
all the axes. If "z" is missing, then it's the same as the whole .speed
missing.

### DIFF
--- a/plugins/zstack.py
+++ b/plugins/zstack.py
@@ -125,10 +125,12 @@ class ZStackPlugin(Plugin):
         return (float > 0): estimated time (in s) that it takes to move the focus
           by one step.
         """
+        speed = None
         if model.hasVA(self.focus, "speed"):
-            speed = self.focus.speed.value['z']
-        else:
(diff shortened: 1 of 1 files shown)


plugins/zstack.py
```diff
@@ -125,10 +125,12 @@ def _estimate_step_duration(self):
         return (float > 0): estimated time (in s) that it takes to move the focus
           by one step.
         """
+        speed = None
         if model.hasVA(self.focus, "speed"):
-            speed = self.focus.speed.value['z']
-        else:
+            speed = self.focus.speed.value.get('z', None)
+        if speed is None:
```
Kleijwegt, line 131: Maybe it could be logged that None is replaced by a (very small) number?

pieleric: That's only used for the estimated acquisition time. I don't think it's a big deal that we need to put log messages.

amuskens (approved):

PR #1539, opened 2020-10-23
driver simsem: don't fail to generate an image if the shift is large

pieleric:
The shift is also taken into account to move the generated image within the whole raw image.
While the translation VA can never be set to go out of bound, the shift
VA can. In such case, we shouldn't fail to simulate, but just clip the
image.


lazem (approved):
Looking good.

PR #1551, opened 2020-11-03
model vattribute: don't fail on __del__ if __init__ failed

pieleric:
__del__() is called even if __init__() didn't finish.
If it failed, for instance because the VA value is incorrect, then the
new code unregistering from the remote listeners would fail.
=> do the same as for the other attributes.


tepals (approved):
Looks good

PR #1552, opened 2020-11-03
odemis acq calibration: fix reading/writing streakcam calibration on Python 3

pieleric:
On Python 2 it was recommended to open the file in binary, but it
actually worked as text. On Python 3, the csv module needs to use the
text mode.
=> Use the text mode.

Move the whole writing to a separate function, so that it can be
tested... and add a test.

### DIFF
--- a/src/odemis/acq/calibration.py
+++ b/src/odemis/acq/calibration.py
@@ -19,6 +19,7 @@ You should have received a copy of the GNU General Public License along with Ode
 
 from __future__ import division
 
+import csv
 import logging
 import numpy
 from odemis import model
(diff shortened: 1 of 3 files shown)


src/odemis/acq/test/calibration_test.py
```diff
@@ -352,5 +353,49 @@ def test_compensate_out(self):
...
+                          1e-07: 1.69e-07,
+                          2e-07: 3.02e-07,
+                          5e-07: 7.31e-07,
+                          0.0001: 0.000161,
+                          0.0002: 0.00032,
+                          0.0005: 0.000798,
+                          0.01: 0.0154}
+
+
+class TimeRangeDelayTest(unittest.TestCase):
```
tepals, line 371: This could use a small explanation of what this test class does. I don't know much about this code and wouldn't expect something with the name `TimeRangeDelayTest` to test writing and reading a csv.

tepals (commented):
one small comment, the rest looks good

Sabrina85 (approved):
Looks good!

pieleric:
A new version. I was unhappy of asymmetry between the reading and writing function. Now, they both take a filename as input, and take care of opening it.
Also updated the test case to test in case of error, and to indicate what the test case is about.

src/odemis/acq/calibration.py
```diff
@@ -356,48 +357,59 @@ def apply_spectrum_corrections(data, bckg=None, coef=None):
...
+
+
+def read_trigger_delay_csv(filename, time_choices, trigger_delay_range):
+    """
+    Read the MD_TIME_RANGE_TO_DELAY from a CSV file, and check its validity based on the hardware
+    filename (str): the path to file
+    trig_delays (dict float -> float): time range to trigger delay info
+    time_choices (set): choices possible for timeRange VA
+    return (dict float -> float): new dictionary containing the loaded time range to trigger delay info
+    raise ValueError: if the data of the CSV file cannot be parsed or doesn't fit the hardware
```
tepals, line 380: What happens if the file doesn't exist?

src/odemis/acq/calibration.py
```diff
@@ -356,48 +357,59 @@ def apply_spectrum_corrections(data, bckg=None, coef=None):
...
+        for time_range, trig_delay in trig_delays.items():
+            calibFile.writerow([time_range, trig_delay])
+
+
+def read_trigger_delay_csv(filename, time_choices, trigger_delay_range):
+    """
+    Read the MD_TIME_RANGE_TO_DELAY from a CSV file, and check its validity based on the hardware
+    filename (str): the path to file
+    trig_delays (dict float -> float): time range to trigger delay info
+    time_choices (set): choices possible for timeRange VA
```
tepals, line 378: It's cleaner if these are in the order of the input arguments, and `trig_delays` is named `trigger_delay_range` in the input arguments.

tepals (commented):
small comments

src/odemis/acq/calibration.py
pieleric: IOError. I've added it to the docstring.

src/odemis/acq/calibration.py
pieleric: Indeed, I copy-pasted the wrong line. Fixed.

pieleric:
Updated docstring.

PR #1553, opened 2020-11-05
Added to_pixel_index function

effting:
Utility function to convert from physical coordinates to pixel indices.

### DIFF
--- a/src/odemis/util/test/transform_test.py
+++ b/src/odemis/util/test/transform_test.py
@@ -350,8 +350,8 @@ class ToPhysicalSpaceKnownValues(PixelIndexCoordinateTransformBase):
         # tuple
         for ji, xy, xy2, xy23 in zip(self._ji, self._xy, self._xy2, self._xy23):
             res = to_physical_space(ji, self._shape)
-            res2 = to_physical_space(ji, self._shape, pixel_size=2.)
-            res23 = to_physical_space(ji, self._shape, pixel_size=(2., 3.))
+            res2 = to_physical_space(ji, self._shape, pixel_size=2)
+            res23 = to_physical_space(ji, self._shape, pixel_size=(2, 3))
 
             numpy.testing.assert_array_almost_equal(xy, res)
             numpy.testing.assert_array_almost_equal(xy2, res2)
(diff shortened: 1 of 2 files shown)


src/odemis/util/transform.py
```diff
@@ -288,6 +288,65 @@ def to_physical_space(ji, shape, pixel_size=None):
...
+        entry is the number of columns in the image.
+    pixel_size : tuple of 2 floats, float (optional)
+        Pixel size in (x, y). For square pixels, a single float can be
+        provided.
+
+    Returns
+    -------
+    ji : ndarray
+        Pixel indices. Same shape as `xy`. For each index the first entry is
+        the row-index `j` and the second entry is the column-index `i`.
```
pieleric, line 318: Could you indicate that it's floats => *not* rounded. For pixel indices, it's a bit odd, so it's worth highlighting. (And I agree that it's fine to *not* round to an integer)

src/odemis/util/test/transform_test.py
```diff
@@ -415,6 +418,64 @@ def test_to_physical_space_raises_value_error(self):
...
+        numpy.testing.assert_array_almost_equal(self._ji, res2)
+        numpy.testing.assert_array_almost_equal(self._ji, res23)
+
+        # ndarray
+        xy = numpy.array(self._xy)
+        xy2 = numpy.array(self._xy2)
+        xy23 = numpy.array(self._xy23)
+        res = to_pixel_index(xy, self._shape)
+        res2 = to_pixel_index(xy2, self._shape, pixel_size=2.)
+        res23 = to_pixel_index(xy23, self._shape, pixel_size=(2., 3.))
```
pieleric, line 452: Please, stop adding `.` after the numbers, it makes it look odd (and old ;-) ).

src/odemis/util/transform.py
```diff
@@ -288,6 +288,65 @@ def to_physical_space(ji, shape, pixel_size=None):
...
+    xy : tuple, list of tuples, ndarray
+        Physical coordinates, list of coordinates, or array of coordinates. For
+        each coordinate the first entry is the `x`-coordinate and the second
+        entry is the `y`-coordinate.
+    shape : tuple of ints
+        Shape of the image. The first entry is the number of rows, the second
+        entry is the number of columns in the image.
+    pixel_size : tuple of 2 floats, float (optional)
+        Pixel size in (x, y). For square pixels, a single float can be
+        provided.
```
pieleric, line 312: Let's mention that if not specified, it defaults to 1? Or maybe instead of using `None` as default value, you could use `1`, and have a special/fast case for that value.

pieleric (changes requested):
Looks fine. Just a few comments.

effting:
implemented requested changes

pieleric (approved):

PR #1554, opened 2020-11-05
Added DeprecationWarning to transform.__call__ method

effting:
Usage of transform.apply() is now preferred.


pieleric (approved):
Looks fine. I like it better with .apply() too.

PR #1555, opened 2020-11-05
driver tmcm: adjust simulator to not fail on 1-axis hardware

pieleric:
The user config contained 3 axes, so on 1-axis simulated hardware, it
would fail.
=> Make a new user config with only 1 axis. It ought to work on any
simulated hardware!

### DIFF
--- a/src/odemis/driver/tmcm.py
+++ b/src/odemis/driver/tmcm.py
@@ -2302,8 +2302,7 @@ class TMCMSimulator(object):
         # 4 * dict(int -> int: param number -> value)
         self._gstate = [{}, {},
                         # Bank 2: example user config v1
-                        {0: 353304838, 1: 50462723, 2: 17104896, 3: 33555201,
-                         4: 83886082, 5: 196869, 6: 0},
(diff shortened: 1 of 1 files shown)


src/odemis/driver/tmcm.py
```diff
@@ -2302,8 +2302,7 @@ def __init__(self, timeout=0, naxes=3, *args, **kwargs):
         # 4 * dict(int -> int: param number -> value)
         self._gstate = [{}, {},
                         # Bank 2: example user config v1
-                        {0: 353304838, 1: 50462723, 2: 17104896, 3: 33555201,
-                         4: 83886082, 5: 196869, 6: 0},
+                        {0: 168034562, 1: 50462723, 2: 17104896},
```
tepals, line 2305: Should the default for the kwarg `naxes` now also be `1`?

pieleric: No, it's not directly related. It comes with configuration for 1 axis, but works fine with more axes.

tepals (approved):

PR #1556, opened 2020-11-05
gui win dialog: fix DELPHI calibration window on Python 3

pieleric:
The XRC python wrapper wasn't Python 3 ready => rebuilt it.


lazem (approved):
This change entails that the `dialog.xrc` shouldn't be generated in a similar way as the` main.xrc`, it has to be modified in `dialog_xrc.py` file, right? 
Otherwise it's fine.

lazem:
Nevermind, I regenerate it using `generate_xrc.py` :+1:

PR #1557, opened 2020-11-05
gui cont microscope: DELPHI first calibration dialog automatically closed when sample is ejected

pieleric:
When inserting a sample holder which is not calibrated, a dialog box
appear to either eject the sample or calibrate it.
This happens on the first time the CLEM sample holder is inserted.
However, if the user also use a standard SEM sample holder, with the
standard Phenom GUI, while Odemis is running, this also happens.
Then, the dialog appear *every* time the sample holder is inserted.
This can cause many time the same dialog box to be shown. None of them
useful.

=> Automatically close the dialog window when the sample is ejected by
an other mean than the Odemis GUI. This way, it never shows more than
one dialog.

Can be tested on the simulator by removing the calibration file, and
then doing:
odemis-cli --set-attr chamber opened True
odemis-cli --set-attr chamber opened False

### DIFF
--- a/src/odemis/gui/cont/microscope.py
+++ b/src/odemis/gui/cont/microscope.py
@@ -648,6 +648,8 @@ class DelphiStateController(SecomStateController):
         self._in_overview = threading.Event()
         self._phenom_load_done = True
 
+        self._first_calib_dlg = None
+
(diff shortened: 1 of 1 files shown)


lazem (approved):
Looks fine.

PR #1559, opened 2020-11-06
gui util img: fix test cases

pieleric:
Some test cases had left-over debug code which broke them.
Some other test cases where affected by the small changes in the
changes of calc_img_buffer_rect(). So update them with the new expected
values (all +/- 1).

### DIFF
--- a/src/odemis/gui/util/test/img_test.py
+++ b/src/odemis/gui/util/test/img_test.py
@@ -970,7 +970,7 @@ class TestSpatialExport(test.GuiTestCase):
         view_pos = [-0.001211588332679978, -0.00028726176273402186]
         draw_merge_ratio = 0.3
         exp_data = img.images_to_export_data([self.streams[0]], view_hfw, view_pos, draw_merge_ratio, False)
-        self.assertEqual(exp_data[0].shape, (2340, 2560, 4))  # RGB
+        self.assertEqual(exp_data[0].shape, (2339, 2560, 4))  # RGB
 
     def test_crop_and_interpolation_need(self):
         """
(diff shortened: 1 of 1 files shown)


src/odemis/gui/util/test/img_test.py
```diff
@@ -1311,17 +1311,14 @@ def test_image_on_ovv(self):
...
+        tile = model.DataArray(255 * numpy.ones((3, 3, 3), dtype=numpy.uint8))
+        tile.metadata[model.MD_POS] = (-3, 3)
+        tile.metadata[model.MD_PIXEL_SIZE] = (1, 1)
         ovv_im_new = insert_tile_to_image(tile, ovv_im)
-        print(ovv_im_new[:, :, 0])  # DEBUG
+        # print(ovv_im_new[:, :, 0])  # DEBUG
         # rectangle with edges (-4,4), (-2,4), (-4,2), (-4,2) should be white now
         # (6,6) is the center, so this corresponds to a rectangle with top left at (2,2) (=(1,1) when counting from 0)
-        numpy.testing.assert_array_equal(ovv_im_new[1:4, 1:4, :], tile)
+        numpy.testing.assert_array_equal(ovv_im_new[1:4, 1:4, :], 255)
```
philipwinklerdelmic, line 1321: shouldn't the second argument be an iterable?

pieleric: Well, I've just found out that not for this numpy test. It actually does something like `a == b`, and in numpy it's allowed that a and b are not the same dimension (in which case there some kind "dimension unfolding").

PR #1560, opened 2020-11-06
driver xtclient fixed test cases for beam shift and blank beam

tepals:
- In test_set_beam_shift a tuple was compared in assertAlmostEqual, I split it up in two comparisons for each value.
- For blank beam to work the scan mode must be full frame and the channel state unpaused.

### DIFF
--- a/src/odemis/driver/test/xt_client_test.py
+++ b/src/odemis/driver/test/xt_client_test.py
@@ -42,7 +42,7 @@ CONFIG_STAGE = {"name": "stage", "role": "stage",
                 "inverted": ["x"],
                 }
 CONFIG_FOCUS = {"name": "focuser", "role": "ebeam-focus"}
-CONFIG_SEM = {"name": "sem", "role": "sem", "address": "PYRO:Microscope@localhost:4242",
+CONFIG_SEM = {"name": "sem", "role": "sem", "address": "PYRO:Microscope@192.168.31.130:4242",
               "children": {"scanner": CONFIG_SCANNER,
                            "focus": CONFIG_FOCUS,
                            "stage": CONFIG_STAGE,
@@ -191,7 +191,7 @@ class TestMicroscope(unittest.TestCase):
     def test_blank_beam(self):
         """Test that the beam is unblanked after unblank beam is called, and blanked after blank is called."""
         self.scanner.blanker.value = False
-        self.assertFalse(self.scanner.blanker.value)
+        self.assertIsInstance(self.scanner.blanker.value, bool)
         self.scanner.blanker.value = True
         self.assertTrue(self.scanner.blanker.value)
 
--- a/src/odemis/driver/test/xt_client_test.py
+++ b/src/odemis/driver/test/xt_client_test.py
@@ -30,6 +30,7 @@ import unittest
 from odemis.driver import xt_client
 from odemis.driver.xt_client import DETECTOR2CHANNELNAME
 from odemis.model import ProgressiveFuture
+from odemis.util import test
 
 logging.basicConfig(level=logging.INFO)
 
@@ -189,13 +190,10 @@ class TestMicroscope(unittest.TestCase):
 
     def test_blank_beam(self):
         """Test that the beam is unblanked after unblank beam is called, and blanked after blank is called."""
-        self.microscope.set_scan_mode("full_frame")
-        self.microscope.set_channel_state('electron1', True)
         self.scanner.blanker.value = False
         self.assertFalse(self.scanner.blanker.value)
         self.scanner.blanker.value = True
         self.assertTrue(self.scanner.blanker.value)
-        self.microscope.set_channel_state('electron1', False)
 
     def test_rotation(self):
         """Test setting the rotation."""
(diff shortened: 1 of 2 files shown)


src/odemis/driver/test/xt_client_test.py
```diff
@@ -340,12 +343,16 @@ def test_set_beam_shift(self):
         new_beam_shift_x = beam_shift_range[1][0] - 1e-6
         new_beam_shift_y = beam_shift_range[0][1] + 1e-6
         self.scanner.beamShift.value = (new_beam_shift_x, new_beam_shift_y)
-        self.assertAlmostEqual((new_beam_shift_x, new_beam_shift_y), self.scanner.beamShift.value)
+        current_shift = self.scanner.beamShift.value
+        self.assertAlmostEqual(new_beam_shift_x, current_shift[0])
+        self.assertAlmostEqual(new_beam_shift_y, current_shift[1])
```
tepals, line 348: @pieleric I found a few cases in odemis where `assertTupleAlmostEqual` has been defined in the test cases as a function. Does it make sense to move that function (in a separate PR) to `odemis.util.test`?

pieleric: Yes. The only reason I haven't done it is because I'm too lazy, but the more we wait, the more the work it is to change! So go ahead :-)
I think that it should then be called `assert_tuple_almost_equal()`.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -189,10 +189,13 @@ def test_set_ht_voltage(self):
 
     def test_blank_beam(self):
         """Test that the beam is unblanked after unblank beam is called, and blanked after blank is called."""
+        self.microscope.set_scan_mode("full_frame")
+        self.microscope.set_channel_state('electron1', True)
         self.scanner.blanker.value = False
         self.assertFalse(self.scanner.blanker.value)
         self.scanner.blanker.value = True
         self.assertTrue(self.scanner.blanker.value)
+        self.microscope.set_channel_state('electron1', False)
```
pieleric, line 198: No. These extra functions are not part of the Odemis API. It means that to turn off the blanker the user wouldn't be able to do it from the command line with `odemis-cli --set-attr e-beam blanker False`, and even the GUI code would be unlikely to do the right thing (as it's a generic code for all types of SEM).

I think that the proper approach is to call these functions in the VA setter. You can assume that the user is not using/touching the standard SEM software.

pieleric (changes requested):

src/odemis/driver/test/xt_client_test.py
tepals: I'm implementing this and realized the easiest way to do it is to use `numpy.testing.assert_allclose()`

```
def assert_tuple_almost_equal(first, second, places=None, msg=None, delta=None):
    """
    check two tuples are almost equal (value by value)
    """
    numpy.testing.assert_allclose(first, second, rtol=10**-places, atol=delta, err_msg=msg)
```

The old functions do this:

```
def assertTupleAlmostEqual(self, first, second, places=None, msg=None, delta=None):
    """
    check two tuples are almost equal (value by value)
    """
    for f, s in zip(first, second):
        self.assertAlmostEqual(f, s, places=places, msg=msg, delta=delta)
```
But in odemis.util.test I cannot use `self.assertAlmostEqual`, because that's part of the testing class. Should I use the implementation with `assert_allclose`, or everywhere where `assertTupleAlmostEqual` is used should I replace it with `numpy.testing.assert_allclose`?

pieleric: It's not exactly identical. Because numpy has this "automatic dimension folding" mechanism, which shouldn't happen here. Ex: `numpy.testing.assert_allclose([1,1], 1)`  => Fine

So, I'd suggest to use your first proposal as `assert_tuple_almost_equal()`, but also checks that the two tuples are the same length. Note that the old version didn't exactly do that, as it only compares the elements present in both tuples (that's the behaviour of `zip()`, but that wasn't on purpose, so let's not reproduce that.

tepals:
@pieleric updated based on your comments

src/odemis/driver/xt_client.py
```diff
@@ -1041,10 +1041,12 @@ def _setVoltage(self, voltage):
 
     def _setBlanker(self, blank):
         """True if the the electron beam should blank, False if it should be unblanked."""
+        self.parent.set_channel_state(DETECTOR2CHANNELNAME["se-detector"], True)
```
tepals, line 1044: @pieleric The problem is actually not in setting the blanker, but in getting the blanker. To get the true state of the beam blank, the stream needs to run. I can add a `_getBlanker` that looks like this:

``` python
def _getBlanker(self, blank):
    self.parent.set_channel_state(DETECTOR2CHANNELNAME["se-detector"], True)
    blanked = self.parent.beam_is_blanked()
    self.parent.set_channel_state(DETECTOR2CHANNELNAME["se-detector"], False)
    return blanked
```

The only thing I'm a bit hesitant about is that if we add this to `_updateSettings` the channel will then be unpaused/paused every 5 seconds when the Delmic XTAdapter is running. I'm not sure if users will like how that looks?

pieleric: Definitely, we don't want to do that in the getter. 
The change to the setter are not useful then? If so, remove them, and let's give up on this function properly working.
I'd suggest to just adjust the test cases to not complain if the blanking VA is not updated.

tepals: The changes I made to the setter were not useful, I've reverted them. I also updated the test, so now at least it passes.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -190,7 +191,7 @@ def test_set_ht_voltage(self):
     def test_blank_beam(self):
         """Test that the beam is unblanked after unblank beam is called, and blanked after blank is called."""
         self.scanner.blanker.value = False
-        self.assertFalse(self.scanner.blanker.value)
+        self.assertIsInstance(self.scanner.blanker.value, bool)
```
pieleric, line 194: Please add a comment here why you do such a week check.

pieleric (changes requested):
Looks fine, but please add a big comment in the code that update the blanker VA that it actually doesn't really work, and describe a bit the current state of the knowledge about this issue (and why we don't fix it for now).

PR #1561, opened 2020-11-06
XT client: When the beam is off, it can be that the voltage is …

tepals:
… outside of the specified XT range. If the value is out of range, clip it to be inside range.

### DIFF
--- a/src/odemis/driver/xt_client.py
+++ b/src/odemis/driver/xt_client.py
@@ -28,6 +28,7 @@ from concurrent.futures import CancelledError
 
 import Pyro5.api
 import msgpack_numpy
+import numpy
 
 from odemis import model
 from odemis import util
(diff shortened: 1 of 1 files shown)


src/odemis/driver/xt_client.py
```diff
@@ -996,7 +998,13 @@ def _updateSettings(self):
...
                 self.dwellTime.notify(dwell_time)
             voltage = self.parent.get_ht_voltage()
-            if voltage != self.accelVoltage.value:
+            v_range = self.accelVoltage.range
+            if voltage != self.accelVoltage.value and v_range[0] <= voltage <= v_range[1]:
+                self.accelVoltage._value = voltage
+                self.accelVoltage.notify(voltage)
+            elif voltage != self.accelVoltage.value:
+                logging.info("Voltage {} V is outside of range {}, clipping to nearest value.".format(voltage, v_range))
+                voltage = numpy.clip(self.parent.get_ht_voltage(), v_range[0], v_range[1])
```
pieleric, line 1007: It should be possible to write this in simpler way, such as:
```python
v_range = self.accelVoltage.range
if not v_range[0] <= voltage <= v_range[1]:
    logging.info(...)
    voltage = self.accelVoltage.clip(voltage)
if voltage != self.accelVoltage.value:
    self.accelVoltage._value = voltage
    self.accelVoltage.notify(voltage)
```
Note that ContinuousVA has a .clip() function.
Also, it's a little better to clip before checking for the value to be different, because otherwise, when the value is out of range, the value will be notified every time the settings are updated (5s). It's a little annoying to get "new" values from a VA which actually hasn't changed.

pieleric (changes requested):

tepals:
@pieleric I updated with an amend to the previous commit

PR #1562, opened 2020-11-08
Added method on GeometricTransform to calculate the fiducial registration error

effting:


src/odemis/util/transform.py
```diff
@@ -431,6 +431,30 @@ def __call__(self, x):
         return self.apply(x)
     __call__.__doc__ = apply.__doc__
 
+    def fre(self, x, y):
+        """
+        Returns the RMS value of the fiducial error registration (FRE).
```
tepals, line 436: Why use FRE and not the target registration error? If I google FRE, I get multiple results saying to use TRE.

> Fiducial Registration Error (FRE) is often used as a surrogate for the Target Registration Error (TRE) that is the quantity of clinical interest [4]. Unfortunately, it has been shown [5] that TRE and FRE are uncorrelated. FRE is thus an unreliable predictor of registration accuracy. (https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4423533/)

Could you also add another sentence briefly explaining when to use this function. Something like:
>This returns a measure for the registration error between source and target frame.

src/odemis/util/transform.py
```diff
@@ -431,6 +431,30 @@ def __call__(self, x):
...
+        x : (n, 2) array
+            Coordinates in the source reference frame.
+        y : (n, 2) array
+            Coordinates in the destination reference frame. Must be of same
+            dimensions as `x`.
+
+        Returns
+        -------
+        fre : float
+            The root mean squared fiducial registration error.
```
pieleric, line 449: Following on Thera's comment, please indicate that  the float >= 0, and that "the smaller, the better the fitting".

src/odemis/util/test/transform_test.py
```diff
@@ -601,6 +601,7 @@ def test_rigid_transform_from_pointset_known_values(self):
             tform = RigidTransform.from_pointset(src, dst)
             self.assertAlmostEqual(0., _angle_diff(rotation, tform.rotation))
             numpy.testing.assert_array_almost_equal(translation, tform.translation)
+            self.assertAlmostEqual(tform.fre(src, dst), 0)
```
pieleric, line 604: Could you add a test case where .fre() doesn't return 0? 
Maybe it's just a matter of extending src & dst with some non matching values, and checking that FRE increases?

pieleric (changes requested):

src/odemis/util/transform.py
effting: I updated the docstring to clarify

src/odemis/util/transform.py
effting: I updated the docstring.

src/odemis/util/test/transform_test.py
effting: I added a test in `SimilarityTransformKnownValues.test_similarity_transform_from_pointset_umeyama`

pieleric (approved):

tepals (approved):

PR #1563, opened 2020-11-09
driver-piezomotor: bugfix and yaml file

philipwinklerdelmic:
* fixed bug in init (probably introduced during review process)
* added testcase for bug (was not caught when testing with single axis, so test with 2 axes, at least in simulator)
* added yaml file to hwtest because the driver parameters are sufficiently complicated

### DIFF
--- a/install/linux/usr/share/odemis/hwtest/piezomotor.odm.yaml
+++ b/install/linux/usr/share/odemis/hwtest/piezomotor.odm.yaml
@@ -1,6 +1,6 @@
 MultiBeamSEM: {
   class: Microscope,
-  role: mbsem,
+  role: brightfield,
   children: ["Piezomotor"],
 }
 
--- a/install/linux/usr/share/odemis/hwtest/piezomotor.odm.yaml
+++ b/install/linux/usr/share/odemis/hwtest/piezomotor.odm.yaml
@@ -1,0 +1,30 @@
+MultiBeamSEM: {
+  class: Microscope,
+  role: mbsem,
+  children: ["Piezomotor"],
+}
+
+Piezomotor: {
+    class: piezomotor.PMD401Bus,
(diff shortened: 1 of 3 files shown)


src/odemis/driver/test/piezomotor_test.py
```diff
@@ -35,9 +35,11 @@
 # Export TEST_NOHW=1 to force using only the simulator and skipping test cases
 # needing real hardware
 TEST_NOHW = (os.environ.get("TEST_NOHW", 0) != 0)  # Default to Hw testing
+MULTIPLE_AXES = (os.environ.get("MULTIPLE_AXES", 0) != 0)  # if available, test 2 axes
 
 TEST_SPEED = 0.001  # m / s
 AXIS_NUM = 1
+AXIS_NUM2 = 2
```
pieleric, line 42: Is that really necessary? In practice, for a lot of drivers, when testing with a specific hardware you have  to adjust the test case configuration. I'd suggest to do the same as in most other drivers: have a KWARGS variable, instead of TEST_SPEED, AXIS_NUM and AXIS_NUM2. And have it configured for 2 axes. If at some point we need to test on a hardware with a single axis, we'll disable the second axis config.

install/linux/usr/share/odemis/hwtest/piezomotor.odm.yaml
```diff
@@ -0,0 +1,30 @@
+MultiBeamSEM: {
+  class: Microscope,
+  role: mbsem,
```
pieleric, line 3: Isn't that a little presumptuous for a microscope with just a stage? ;-) Let's use just "brightfield" as role.

pieleric (changes requested):

philipwinklerdelmic:
updated

PR #1567, opened 2020-11-12
Backend yaml to startup the technolution/asm wrapper

Kleijwegt:


install/linux/usr/share/odemis/hwtest/technolution-asm.odm.yaml
```diff
@@ -0,0 +1,37 @@
...
+    role: mbsem,
+    children: ["ASM", "EBeamScanner", "MirrorDescanner", "MPPC",],
+}
+
+"ASM": {
+  class: technolution.AcquisitionServer,
+  role: "main",
+  init: {
+    # Use following host url for real system instead of simulator: "http://192.168.10.10:8080/v2"
+    host: "http://localhost:8080/v2",
```
Kleijwegt, line 12: Do we want as default the simulator or the real system url?

src/odemis/driver/technolution.py
```diff
@@ -175,21 +175,21 @@ def __init__(self, name, role, host, children, externalStorage, **kwargs):
             ckwargs = children["EBeamScanner"]
         except Exception:
             raise ValueError("Required child EBeamScanner not provided")
-        self._ebeam_scanner = EBeamScanner(parent=self, **ckwargs)
+        self._ebeam_scanner = EBeamScanner(parent=self, daemon=kwargs.get("daemon"), **ckwargs)
```
Kleijwegt, line 178: Changes made to to the technolution wrapper to have the yaml startup the wrapper correctly.

install/linux/usr/share/odemis/hwtest/technolution-asm.odm.yaml
pieleric: For such file, it'd be the real system IP.

pieleric (approved):

install/linux/usr/share/odemis/hwtest/technolution-asm.odm.yaml
Sabrina85: I agree it should be the real system IP as this yaml file is placed in the hwtest folder. All simulator yaml files are in the sim folder.

install/linux/usr/share/odemis/hwtest/technolution-asm.odm.yaml
```diff
@@ -0,0 +1,37 @@
...
+}
+
+"EBeamScanner": {
+  # Internal child of ASM, so no class
+  role: "multibeam",
+}
+
+"MirrorDescanner": {
+  # Internal child of ASM, so no class
+  role: "galvo",
```
Sabrina85, line 31: "galvo" does not look like a good name to me. The corresponding child is named "MirrorDescanner". It is not obvious that the mirror descanner is a mirror galvonometer. I suggest some roles that are inspired by the class names as we usually stick with them "forever". "mppc" and "multibeam" are fine, for the galvo I suggest maybe "multibeam-descanner" or just "descanner" to keep it short?

Sabrina85 (changes requested):

install/linux/usr/share/odemis/hwtest/technolution-asm.odm.yaml
Kleijwegt: Let's go for descanner then.

Sabrina85 (approved):
Looks good!

install/linux/usr/share/odemis/hwtest/technolution-asm.odm.yaml
```diff
@@ -0,0 +1,37 @@
+MultiBeamSEM: {
+    class: Microscope,
+    role: mbsem,
+    children: ["ASM", "EBeamScanner", "MirrorDescanner", "MPPC",],
+}
+
+"ASM": {
+  class: technolution.AcquisitionServer,
+  role: "main",
```
pieleric, line 9: If I understand correctly, this component is also useful for the client, right? (eg, to change the external storage). So it should indeed have a role. However, "main" doesn't sound good. What's the role that will be used for the fastem? Maybe "asm" would be fine?

Sabrina85: Indeed overlook this one. "asm" is good.

Kleijwegt: Ok great, @pieleric I think it can be merged then.

PR #1570, opened 2020-11-13
2D plots: proper display of time values

VictoriaDel:


src/odemis/gui/util/img.py
```diff
@@ -848,16 +848,21 @@ def calculate_ticks(value_range, client_size, orientation, tick_spacing):
...
     # Get the horizontal/vertical space in pixels
     if orientation == wx.HORIZONTAL:
         pixel_space = client_size[0]
         # Don't display ticks too close from the left border
         min_pixel = 10
     else:
+        # Don't display ticks too close from the border
         pixel_space = client_size[1]
-        min_pixel = 0
+        min_pixel = 10
```
Kleijwegt, line 865: If in both if and else the min_pixel is set to 10 shoudln't it be set outside the if statement?
Especially since the comment:
>  Don't display ticks too close from the border
Applies to both the horizontal and vertical case.

Kleijwegt, line 865: If in both if and else the min_pixel is set to 10 shoudln't it be set outside the if statement?
Especially since the comment:
>  Don't display ticks too close from the border
Applies to both the horizontal and vertical case.

src/odemis/gui/util/img.py
```diff
@@ -848,16 +848,21 @@ def calculate_ticks(value_range, client_size, orientation, tick_spacing):
         logging.info("Trying to compute legend tick without range")
         return None
 
-    min_val, max_val = value_range[0], value_range[-1]
+    reverse = value_range[0] > value_range[-1]
+    if reverse:
+        min_val, max_val = value_range[-1], value_range[0]
+    else:
+        min_val, max_val = value_range[0], value_range[-1]
```
Kleijwegt, line 855: I think range is useally just a tuple of two isn't it?
Can this not be shortened into:
> min_val, max_val = min(value_range), max (value_range)

or maybe better:
> min_val = min(value_range)
> max_val = max (value_range)

To me this is clearer and uses less code but maybe value_range is more complicated than I think/see

Kleijwegt (changes requested):

src/odemis/gui/util/img.py
VictoriaDel: Yes, I first want to be sure that 'Eric is ok with defining min_pixel for both vertical and horizontal and I will merge it.

src/odemis/gui/util/img.py
pieleric: Actually, that was the main previous change by Victoria, now "range_value" can be a list. I agree with you that we could avoid using the if to make the code a little bit easier to read. Maybe something like this:
```python
min_val = min(value_range[0], value_range[-1])
max_val = max(value_range[0], value_range[-1])
```

VictoriaDel: yes you are right, as long as I don't use reverse again, I can immediately do it with min and max

src/odemis/gui/comp/legend.py
```diff
@@ -478,7 +481,8 @@ def on_paint(self, _):
                 lpos = pos + (lbl_height // 2)
                 lpos = max(min(lpos, csize.y), 2)
 
-                if prev_lpos >= lpos + 20 or i == 0:
+                if (rng[0] < rng[-1] and prev_lpos >= lpos + 20) or (rng[0] > rng[-1] and prev_lpos <= lpos+20)\
+                        or i == 0:
```
pieleric, line 485: I'd just write:
```python
if i == 0 or  abs(prev_lpos - lpos) >= 20:  
```

pieleric (changes requested):
Looks fine. I just have (long!) minor comments, most related to improving the readability.

PR #1571, opened 2020-11-16
config: fix windows icon having black background

pieleric:
Switch to 32 bits, with transparency. It's much prettier, and Windows
supports this for a long time now.


philipwinklerdelmic (approved):
Haven't tested it. I'm a bit afraid of making Windows changes because something always breaks, but let's see...

pieleric:
I have already tested the icons on Windows. They work fine, so the chance of issues should be low. Anyway, the best way with issues is not to freeze the code, but to be able to quickly find and fix the issues 😉

PR #1572, opened 2020-11-17
CryoSECOM Chamber Tab

lazem:
The main purpose of this tab is to handle the movement of the stage between the different position. During the workflow, this is the first tab displayed, and it’s used every time to switch between loading <> FM/SEM imaging <> milling <> GIS Coating. In addition, it’s used to specify the place to store the data acquired by the user.


src/odemis/gui/cont/tabs.py
```diff
@@ -1602,6 +1611,419 @@ def get_display_priority(cls, main_data):
...
+        self.ion_to_sample = stage_metadata[model.MD_ION_BEAM_TO_SAMPLE_ANGLE]
+        # Default value for milling angle, will be used to store the angle value out of milling position
+        self._prev_milling_angle = math.radians(10)
+        # Define axis connector to link milling angle to UI float ctrl
+        self.milling_connector = AxisConnector('rx', stage, panel.ctrl_milling, pos_2_ctrl=self._milling_angle_changed,
+                                                  ctrl_2_pos=self._milling_ctrl_changed, events=wx.EVT_COMMAND_ENTER)
+        panel.ctrl_milling.SetValueRange(5, 25)
+        panel.ctrl_milling.key_step = 0.1
+        panel.ctrl_milling.Bind(wx.EVT_CHAR, panel.ctrl_milling.on_char)
+        panel.ctrl_milling.accuracy = 4
```
lazem, line 1653: I couldn't set these parameters from the xrc file. Not sure exactly why?

src/odemis/gui/cont/tabs.py
```diff
@@ -1602,6 +1611,419 @@ def get_display_priority(cls, main_data):
...
+        """
+        self.panel.ctrl_milling.Enable(enable)
+        self.panel.stage_align_slider_aligner.Enable(enable)
+        for button in self.btn_aligner_axes.keys():
+            button.Enable(enable)
+
+    def _pause_axis_connectors(self, pause=True):
+        """
+        Pause angle axis connectors (when leaving/entering milling mode)
+        :param pause: (bool) True to pause, False to resume
```
lazem, line 1794: Currently the axis connector is not paused out of milling, the intention is to give a reading of the angle while it's disabled, and it's enabled only in milling. So probably no reason to pause it?

src/odemis/gui/cont/tabs.py
```diff
@@ -1602,6 +1611,419 @@ def get_display_priority(cls, main_data):
...
+        """
+        self.milling_connector.pause() if pause else self.milling_connector.resume()
+
+    def _on_show_advanced(self, evt):
+        """
+        Event handler for the Advanced button to show/hide stage advanced panel
+        """
+        self.panel.pnl_advanced_align.Show(evt.isDown)
```
lazem, line 1802: I tried to set this line with `Show(False)` in the tab's init to start it with this panel hidden, but it's not shown again if I did that! Gonna investigate this further but maybe you've a pointer on why this happens?

src/odemis/gui/cont/tabs.py
```diff
@@ -1602,6 +1611,419 @@ def get_display_priority(cls, main_data):
...
+
+        if angle_type == self.MILL:
+            try:
+                milling_angle = get_angle_value(self.panel.ctrl_milling)
+            except ValueError as error:
+                logging.error(error)
+                return
+            return milling_angle
+        elif angle_type == self.ROTATE:
+            raise NotImplementedError("Rotation angle control is not implemented.")
```
lazem, line 1866: Removed the code for handling RZ control but it's the same handling as the milling angle.

src/odemis/acq/move.py
```diff
@@ -292,24 +297,25 @@ def _doCryoTiltSample(future, stage, focus, rx, rz):
...
         # Otherwise stop if it's not already parked
         if not _isNearPosition(focus.position.value, focus_deactive, {'z'}):
-            if _isNearPosition(current_pos,  {'rx': 0, 'rz': 0}, {'rx', 'rz'}):
+            if _isNearPosition(current_pos, {'rx': 0}, {'rx'}):
                 sub_moves.append((focus, focus_deactive))
             else:
                 raise ValueError("Cannot proceed with tilting while focus is not near FAV_POS_DEACTIVE position.")
 
-        if rx == 0 and rz == 0:  # Imaging
+        if rx == 0:  # Imaging
```
lazem, line 314: Now that rz = 0 is not an indication of imaging, left it out this check but kept it as a parameter still for `cryoTiltSample` function.

src/odemis/gui/cont/tabs.py
```diff
@@ -1602,6 +1611,419 @@ def get_display_priority(cls, main_data):
...
+        new_name = ShowChamberFileDialog(self._tab_panel, np)
+        if new_name is not None:
+            # Create project directory and if it exists warn for existing files
+            if not os.path.isdir(new_name):
+                os.mkdir(new_name)
+            elif os.listdir(new_name):
+                logging.warning("Selected directory {} already contains files.".format(new_name))
+            # If the sample is already loaded, move files that exist in previous folder
+            prev_dir = os.path.join(self.conf.last_path, self.conf.fn_count)
+            if self.current_position != LOADING and os.path.isdir(prev_dir):
```
lazem, line 1715: Is this enough check that the sample is already loaded?

src/odemis/model/_metadata.py
```diff
@@ -184,7 +184,9 @@
 MD_FAV_POS_DEACTIVE = "Favourite position deactive"  # dict of str -> float representing a good position for being "deactive" (eg, mirror parked, lens not in use)
 MD_FAV_POS_COATING = "Favourite position coating"  # dict of str -> float representing a good position for GIS coating
 MD_FAV_AREA = "Favourite acquisition area"  # (float, float, float, float) minX, minY, maxX, maxY of the acquisition area
-MD_POS_ACTIVE_RANGE = "Range for active position"  # dict str → (float, float): axis name → (min,max): the range of the axes within which can be used imaging.
+MD_FAV_Z_RANGE = "Favourite acquisition range for Z axis"  # (float, float): (min,max): the range of the Z axis in milling position
```
lazem, line 187: Added this metadata for advanced Z axis range, not sure if that's what you had in mind though.

lazem:
Hi @pieleric I added two commits, one for the move module modifications, and the other for the chamber tab. I'll continue with the remaining specs to finish the project, sample and advanced panels. Appreciating your review meanwhile. 
Thank you!

src/odemis/gui/cont/tabs.py
pieleric: I'm not convinced by the logic. I don't think it should behave differently if it's in LOADING position. It's only when starting the GUI, or unloading that it should create a new directory, immediately. That's probably a separate (simpler function).
If the previous directory exists (which is most likely the case), use `os.rename()`. If rename fails, I think there are mostly two reasons:
* It's in a different partition => do a complete directory move
* It's inside a the current directory => If there are already files in it, tell the user it's not possible. If the prev_dir is empty, create a directory inside it.
* For all other reasons, log the exception, tell the user it failed, and do not change the folder name.


Also, avoid using the logging infrastructure to indicate an issue to the user. If it goes wrong, it should show an error dialogbox. If it's about to do something really dangerous, ask the user in a dialogbox.

pieleric (changes requested):
First comments (I didn't have time to review all the code yet).

src/odemis/acq/move.py
```diff
@@ -351,12 +362,33 @@ def _cancelCryoMoveSample(future):
...
+    try:
+        with future._task_lock:
+            if future._task_state == CANCELLED:
+                logging.info("Reference procedure is cancelled.")
+                raise CancelledError()
+            logging.debug("Performing stage referencing.")
+            future._running_subf = component.reference(set(component.axes.keys()))
+        future._running_subf.result()
+    except Exception as error:
+        logging.error(error)
```
pieleric, line 381: Typically, it's better to `logging.exception("Description of the high level task which failed")`, because the advantage is that if there is an error, the backtrace will help you to understand better the issue.

pieleric (changes requested):
Looks almost fine. I think the behaviour of the milling angle in some corner cases needs to be checked.

Just a few more remarks, but otherwise should be ready to merge.

src/odemis/acq/move.py
```diff
@@ -179,26 +180,34 @@ def cryoLoadSample(target):
...
         focus_md = focus.getMetadata()
         stage_active = stage_md[model.MD_FAV_POS_ACTIVE]
         stage_deactive = stage_md[model.MD_FAV_POS_DEACTIVE]
         stage_coating = stage_md[model.MD_FAV_POS_COATING]
-        stage_active_range = stage_md[model.MD_POS_ACTIVE_RANGE]
         focus_deactive = focus_md[model.MD_FAV_POS_DEACTIVE]
+        # Some sanity checks for rotation axes
+        for rotation_axis in {'rx', 'rz'}:
+            if rotation_axis in stage_active:
+                logging.warning("Imaging position should not contain a value for %s angle." % rotation_axis)
```
lazem, line 205: Changed this to check for both rotation angels, because `rx` is missing in the stage `FAV_POS_ACTIVE` on enzel-tudelft yaml files

src/odemis/acq/move.py
```diff
@@ -179,26 +180,34 @@ def cryoLoadSample(target):
...
+            if rotation_axis not in stage_deactive:
+                logging.warning("Loading position should contain a value for %s angle." % rotation_axis)
```
lazem, line 207: Maybe it'd be better to raise an error here instead of warning?

src/odemis/gui/cont/tabs.py
```diff
@@ -1899,6 +1922,543 @@ def get_display_priority(cls, main_data):
...
+        Get the value of advanced button for _show_advanced_va_connector ctrl_2_va
+        """
+        return self.panel.btn_switch_advanced.GetValue()
+
+    def _on_show_advanced(self, evt):
+        """
+        Event handler for the Advanced button to show/hide stage advanced panel
+        """
+        self.panel.pnl_advanced_align.Show(self.tab_data_model.show_advaned.value)
+        self.panel.Layout()
```
lazem, line 2222: This manages to show the panel albeit the button text is trimmed.
Screenshot from 2021-01-27 20-37-47

src/odemis/acq/move.py
```diff
@@ -91,22 +97,17 @@ def getMovementProgress(current_pos, start_pos, end_pos):
...
-    # Check we have the x,y,z axes in all points
-    check_axes(current_pos)
-    check_axes(start_pos)
-    check_axes(end_pos)
     # Get distance for current point in respect to start and end
     from_start = get_distance(start_pos, current_pos)
     to_end = get_distance(current_pos, end_pos)
     total_length = get_distance(start_pos, end_pos)
+    if total_length == 0: # same value
+        return None
```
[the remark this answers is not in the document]
lazem: Happened while testing acquiring overview map on localization tab. Returned 1 as advised.

src/odemis/gui/cont/tabs.py
pieleric: That looks like a known bug in wxWidgets :-( 
I have a workaround for it in gui.cont.tabs.TabBarController._on_tab_change(). However, it needs to be called when the text label is shown. So that's why it doesn't work here out-of-the-box, as it's hidden by default. You'll need to call the function explicitly on the first time the "Advanced" button is pressed.

src/odemis/acq/move.py
pieleric: Let's just keep the whole code simple. Every metadata has to have every axis. If an axis is missing, it raises an error.

install/linux/usr/share/odemis/sim/cryosecom-sim.yaml
```diff
@@ -200,10 +200,12 @@ CryoSECOM: {
   },
   metadata: {
     POS_ACTIVE_RANGE: {'x':  [0.00, 0.003], 'y':  [-0.003, -0.001], 'z':  [0.00, 0.003], } , # stage Z – lens Z, when the lens Z is the closest (highest) from the stage
-    FAV_POS_ACTIVE: {'rx': 0.00, 'rz': 0.0, 'x':  0.0, 'y':  -0.002, 'z':  0.002},
-    FAV_POS_DEACTIVE:  {'rx': 0.00, 'rz': 0.0, 'x':  0.002, 'y':  0.00, 'z':  -0.001},
+    FAV_POS_ACTIVE: {'rx': 0.00, 'x':  0.0, 'y':  -0.002, 'z':  0.002},
+    FAV_POS_DEACTIVE:  {'rx': 0.00, 'rz': 0.1, 'x':  0.002, 'y':  0.00, 'z':  -0.001},
     FAV_AREA:  [-0.001, -0.001, 0.001, 0.001],  # left, top, right, bottom
-    FAV_POS_COATING: {'rx': 0.4363, 'rz': 0.0, 'x':  0.0, 'y':  -0.003, 'z':  0.001},
+    FAV_Z_RANGE: [0.001, 0.003],
```
pieleric, line 206: FAV_AREA and FAV_Z_RANGE are now merged into  `POS_ACTIVE_RANGE: {'x':  [-23.e-6, 23.e-6], 'y':  [-33.e-6, 33.e-6], 'z':  [0.001, 0.003], } ,`

pieleric (changes requested):
Please note that I already have a much improved simulator in the queue. So you can either drop your changes to the simulator, or just don't worry about the current one.

src/odemis/model/_metadata.py
pieleric: No need for FAV_Z_RANGE. We'll use POS_ACTIVE_RANGE.

src/odemis/gui/cont/tabs.py
```diff
@@ -1899,6 +1922,547 @@ def get_display_priority(cls, main_data):
...
+            raise ValueError('The stage is missing an rx axis.')
+        panel.ctrl_milling.Bind(wx.EVT_CHAR, panel.ctrl_milling.on_char)
+
+        # Create new project directory on starting the GUI
+        self._create_new_dir()
+
+        self.position_btns = {LOADING: self.panel.btn_switch_loading, IMAGING: self.panel.btn_switch_imaging,
+                              MILLING: self.panel.btn_switch_milling, COATING: self.panel.btn_switch_coating}
+        if not {model.MD_FAV_AREA, model.MD_FAV_Z_RANGE}.issubset(stage_metadata):
+            raise ValueError('The stage is missing FAV_AREA and MD_FAV_Z_RANGE metadata.')
```
pieleric, line 1986: Use POS_ACTIVE_RANGE

pieleric (changes requested):

src/odemis/gui/cont/tabs.py
lazem: I was meaning to call `_on_tab_change` explicitly but I didn't find a straightforward way to access the `TabBarController` object from the tab. And looking at it I think it'd be better to just make `_fix_big_static_text` usable to both classes. Where it'd be a good place to put it? gui.util? Thanks

pieleric: Let's put it in odemis.gui.util.wx_adapter (it's not a super logical place, but that's where another code to workaround this issue is already situated).
Then, I'd suggest you create a new function with the content of the _on_tab_change():
```python
def fix_static_text_clipping(...):
           _fix_big_static_text(tab.panel)
            # Eventually, update the size of the parent, based on everything inside it
            wx.CallLater(100, _update_layout_big_text, tab.panel)  # Quickly
            wx.CallLater(500, _update_layout_big_text, tab.panel)  # Later, in case the first time was too early
```

PR #1574, opened 2020-11-19
XT client add support for rotation (r) and tilt (t).

tepals:


src/odemis/driver/xt_client.py
```diff
@@ -1116,18 +1122,9 @@ def __init__(self, name, role, parent, rng=None, **kwargs):
     def _updatePosition(self, raw_pos=None):
         """
         update the position VA
-        raw_pos (dict str -> float): the position in mm (as received from the SEM)
+        raw_pos (dict str -> float): the position (as received from the SEM)
```
pieleric, line 1125: Could you indicate that if None is passed, it'll ask the current position to the SEM?

src/odemis/driver/xt_client.py
```diff
@@ -1209,7 +1206,8 @@ def moveRel(self, shift):
         Parameters
         ----------
         shift: dict(string->float)
-            Relative shift to move the stage to per axes in m. Axes are 'x' and 'y'.
+            Relative shift to move the stage to per axes in m for 'x', 'y', 'z' in rad for 't', 'r'.
+            Axes are 'x', 'y', 'z', 't' and 'r'.
```
pieleric, line 1210: For moveRel, I think there is a little bit more than nothing to do. In case the rotation supports a full turn, then the absolute value computed should be with modulo 2pi. Otherwise if you're at 5°, and you move -10°, it'll fail.

pieleric (changes requested):
Could you add/extend at least one test case for moveRel and one for moveAbs with these new axes?

src/odemis/driver/xt_client.py
tepals: updated

pieleric:
Please add/extend at least one test case for moveRel and one for moveAbs with these new axes.

src/odemis/driver/xt_client.py
```diff
@@ -1116,18 +1123,10 @@ def __init__(self, name, role, parent, rng=None, **kwargs):
...
-            position = self.parent.get_stage_position()
-            x, y, z = position["x"], position["y"], position["z"]
-        else:
-            x, y, z = raw_pos["x"], raw_pos["y"], raw_pos["z"]
-
-        pos = {"x": x,
-               "y": y,
-               "z": z,
-               }
+        pos = self.parent.get_stage_position() if not raw_pos else raw_pos
```
Kleijwegt, line 1129: Does the other way around not make more sense?
> pos = raw_pos if raw_pos else self.parent.get_stage_position()

tepals: yes, that looks better. Switched!

tepals:
@pieleric added a test case

src/odemis/driver/xt_client.py
```diff
@@ -1092,12 +1094,17 @@ def __init__(self, name, role, parent, rng=None, **kwargs):
...
         axes_def = {
-            # Ranges are from the documentation
-            "x": model.Axis(unit="m", range=rng["x"]),
-            "y": model.Axis(unit="m", range=rng["y"]),
-            "z": model.Axis(unit="m", range=rng["z"]),
+            "x": model.Axis(unit=stage_info["unit"]["x"], range=rng["x"]),
+            "y": model.Axis(unit=stage_info["unit"]["y"], range=rng["y"]),
+            "z": model.Axis(unit=stage_info["unit"]["z"], range=rng["z"]),
+            "t": model.Axis(unit=stage_info["unit"]["t"], range=rng["t"]),
+            "r": model.Axis(unit=stage_info["unit"]["r"], range=rng["r"]),
```
pieleric, line 1107: Sorry, I realise only now: in Odemis the rotation axes are normally called rx (for "t") and "rz" (for "r"). So you need translate the axes names.

src/odemis/driver/xt_client.py
```diff
@@ -1194,6 +1193,9 @@ def _doMoveRel(self, future, shift):
         # Check range (for the axes we are moving)
         for an in shift.keys():
             rng = self.axes[an].range
+            if rng == (0, 2 * math.pi):
```
pieleric, line 1196: I'm a little bit picky, but let's add `and an in ("rx", "rz")`. Otherwise, if for some (very unlikely) reason a linear axis has limits exactly between (0 -> 2 pi), it'll start behaving strange.

pieleric (changes requested):

src/odemis/driver/xt_client.py
tepals: done

src/odemis/driver/test/xt_client_test.py
```diff
@@ -106,7 +106,7 @@ def test_move_stage(self):
 
         stage_position = self.stage.position.value.copy()
         # test move_stage method actually moves HW by the requested value
-        self.assertAlmostEqual(stage_position['x'], relative_position['x'] + new_pos['x'])
+        self.assertAlmostEqual(stage_position['x'], relative_position['x'] - new_pos['x'])
```
pieleric, line 109: That's not right, is it? It seems an unrelated change, concerning the "inverted" issue. In any case, I'm quite convince the original code was fine (it should be a + everywhere).

src/odemis/driver/xt_client.py
```diff
@@ -1157,7 +1168,7 @@ def _moveTo(self, future, pos, timeout=60):
                 moving = True
                 tstart = time.time()
                 while moving:
-                    pos = self.parent.get_stage_position()
+                    pos = self._getPosition()
                     moving = self.parent.stage_is_moving()
                     # Take the opportunity to update .position
                     self._updatePosition(pos)
```
pieleric, line 1174: If you pass the result of getPosition, which already has the axes inverted to updatePosition(), the axes are going to be inverted twice. => not inverted.

src/odemis/driver/xt_client.py
```diff
@@ -1126,7 +1126,7 @@ def _updatePosition(self, raw_pos=None):
         raw_pos (dict str -> float): the position (as received from the SEM). If None is passed the current position is
             requested from the SEM.
         """
-        pos = raw_pos if raw_pos else self.parent.get_stage_position()
+        pos = raw_pos if raw_pos else self._getPosition()
         self.position._set_value(self._applyInversion(pos), force_write=True)
```
pieleric: You're inverting the axes twice.

src/odemis/driver/xt_client.py
```diff
@@ -1113,7 +1113,7 @@ def __init__(self, name, role, parent, rng=None, **kwargs):
         self._executor = CancellableThreadPoolExecutor(max_workers=1)  # one task at a time
 
         self.position = model.VigilantAttribute({}, unit=stage_info["unit"],
-                                                readonly=True)
+                                                readonly=True, getter=self._getPosition)
```
pieleric, line 1116: Let's not do that. A getter is called *every* time the VA is read. That means it's going to be very costly (in time), and if for some reason the connection is down, it'll be blocking, really annoying.
That's why I'd recommend what we normally do: update during/after a move, and otherwise, from time to time. It's enough for most use-cases.

pieleric (changes requested):
The axis name translation looks fine. However, I'm not convinced by the getter, and I think that this is where you've introduced the bug related to axis inversion.

tepals:
@pieleric fixed your last comments

PR #1578, opened 2020-11-20
Update tescan driver

VictoriaDel:
Use FetchImage instead of FetchArray
Handle flush
Update testcases

### DIFF
--- a/src/odemis/driver/tescan.py
+++ b/src/odemis/driver/tescan.py
@@ -421,6 +421,28 @@ class SEM(model.HwComponent):
             dt = self._scanner.dwellTime.value * 1e9
             # with self._acquisition_init_lock:
             logging.debug("Acquiring SEM image of %s with dwell time %f ns", res, dt)
+
+            # make sure socket settings are always set
+            self._device.connection.socket_c.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
+            self._device.connection.socket_d.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
+            # Check if spot mode is required
+            if res == (1, 1):
+                if ((self._scaled_shape != scaled_shape) or
+                        (self._roi != (l, t, r, b)) or
+                        (self._dt != dt) or
+                        (self.pre_res != res)):
+                    self._device.ScStopScan()
+                    # flush remaining data in data buffer
+                    self.flush()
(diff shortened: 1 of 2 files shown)


src/odemis/driver/tescan.py
```diff
@@ -421,6 +421,28 @@ def _single_acquisition(self, channel):
...
+                        (self._dt != dt) or
+                        (self.pre_res != res)):
+                    self._device.ScStopScan()
+                    # flush remaining data in data buffer
+                    self.flush()
+                    # need to reset
+                    self._device.connection.socket_c.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
+                    self._device.connection.socket_d.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
+            else:
+                if self.pre_res == (1, 1) or True:  # debug
```
pieleric, line 441: We cannot leave this `or True` as is. Either remove it,  or remove the whole if, and the inside block should always be executed. Hopefully we can leave the if. Can you check?

src/odemis/driver/tescan.py
```diff
@@ -432,45 +454,39 @@ def _single_acquisition(self, channel):
...
                 # fetch the image (blocking operation), ndarray is returned
                 if res == (1, 1):
-                    sem_pxs = self._device.FetchArray(channel, TESCAN_PXL_LIMIT)
+                    sem_pxs = self._device.FetchImage(channel, TESCAN_PXL_LIMIT)
                     # Since we acquired TESCAN_PXL_LIMIT integrations of
                     # dt/TESCAN_PXL_LIMIT we now get the mean signal and return
                     # it as the result
+                    sem_pxs = numpy.frombuffer(sem_pxs, dtype=">u2")
                     sem_img = numpy.array([sem_pxs.mean()])
+                    logging.debug("Acquiring chamber image of %s", sem_img.shape)
```
pieleric, line 474: This log and the one just a few lines below needs to be polished. Here you could log the value directly.
In the log below, show the length instead of shape.

pieleric (changes requested):
Just a few minor cleanups needed.

src/odemis/driver/tescan.py
VictoriaDel: It doesn't work if I just delete "or True". So I will remove the whole if.

PR #1579, opened 2020-11-23
Update role in testcases of the technolution wrapper

Kleijwegt:


Sabrina85 (approved):

pieleric:
Be aware that role (should) have no effect on the behaviour of the code, similarly to the name. It's there for the client. So this change should not have any effect, excepted for cosmetics.

PR #1580, opened 2020-11-23
Add colormap support for time spectrum and line-spectrum projections

pieleric:


src/odemis/gui/util/img.py
```diff
@@ -112,12 +108,60 @@ def format_rgba_darray(im_darray, alpha=None):
...
     else:
         raise ValueError("Unsupported colour depth!")
 
+    new_darray = model.DataArray(rgba)
+    new_darray.metadata['byteswapped'] = True
+    return new_darray
+
+
+def format_bgra_to_rgb(im_darray, keepalpha=True, inplace=False):
+    """ Reshape the given numpy.ndarray from RGB to BGRA format
```
Kleijwegt, line 120: Is this function converting to rgb (as the name suggests) or to BGRA (as the first line of the docstring suggests)

src/odemis/gui/util/img.py
```diff
@@ -112,12 +108,60 @@ def format_rgba_darray(im_darray, alpha=None):
...
+        else:
+            shape = im_darray.shape[:2] + (3,)
+            if inplace:
+                raise ValueError("Cannot drop alpha channel in-place")
+    else:
+        raise ValueError("Unsupported colour depth!")
+
+    if inplace:
+        rgba = im_darray
+        rgba[:,:, [2, 0]] = rgba[:,:, [0, 2]]  # Just switch R <> B
```
Kleijwegt, line 148: Just a thought on this parameter/part of the code:
If the image is modified "in_place" and you don't use the output value (as in a part of your current code), do you still want to return the array?

pieleric: Yes, I did this on purpose. I agree this is arguable, but my idea is that I want the function to always have the same signature. It avoids surprises. That's actually what numpy functions do as well (when you pass a `out` parameter).

src/odemis/gui/util/img.py
pieleric: Indeed. I messed up the docstring. It's now fixed.

amuskens (approved):

PR #1581, opened 2020-11-24
driver tescan: small fixes

pieleric:
Disable the flushing as it doesn't seem to help, and only cause extra
issues.

Also do not change the blanker value if it's set to auto.

### DIFF
--- a/src/odemis/driver/tescan.py
+++ b/src/odemis/driver/tescan.py
@@ -327,17 +327,20 @@ class SEM(model.HwComponent):
             self._acquisition_thread = None
 
     def flush(self):
-        self._device.Disconnect()
-        self._device = sem.Sem()
-        result = self._device.Connect(self._host, 8300)
-        if result < 0:
-            raise HwError("Failed to connect to TESCAN server '%s'. "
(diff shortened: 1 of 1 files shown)


src/odemis/driver/tescan.py
```diff
@@ -873,11 +871,13 @@ def _pollVAs(self):
                         self.probeCurrent._value = new_pc
                         self.probeCurrent.notify(new_pc)
 
-                    bmode = self.parent._device.ScGetBlanker(1)
-                    blanked = (bmode != 0)
-                    if blanked != self.blanker._value:
-                        self.blanker._value = blanked
-                        self.blanker.notify(blanked)
+                    # if blanker is in auto, don't change its value
+                    if not self.blanker.value is None:
```
tepals, line 875: is this different then: `if self.blanker.value is not None:`? If it is not different, I think `if self.blanker.value is not None:` is more readable.

pieleric: Although in my mind the version I wrote feels more natural, I agree that the one you propose is generally more readable (as it's closer to the English grammar).
I'll change that.

tepals (approved):

PR #1582, opened 2020-11-24
Kryoz Cooler: basic skeleton driver to test the first version of he Cryolab cooler firmware

amuskens:


src/odemis/driver/kryoz.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def __init__(self, name, role, host, port=DEFAULT_PORT, daemon=None, **kwargs):
+        """
+        host: (string) the TCP/IP hostname of the server
+        port: (int) the TCP/IP port of the server.
+
+        Raises:
+            ValueError if no scanner child is present
+        """
+        #super(Cooler, self).__init__(name, role, daemon=daemon, **kwargs)
```
Kleijwegt, line 44: Why is this line commented out/still here? This class has no parent so inheriting would be hard anyway isn't it?

src/odemis/driver/kryoz.py
```diff
@@ -0,0 +1,159 @@
...
+    def _sendQuery(self, cmd):
+        """
+        cmd (byte str): command to be sent to device
+        returns (byte str): answer received from the device
+        raise:
+            IOError if no answer is returned in time
+        """
+        cmd = bytes(cmd, 'utf-8') + b"\r"
+        with self._net_access:
+            #logging.debug("Sending: %s", to_str_escape(cmd))
```
Kleijwegt, line 77: Maybe this line can also be deleted?

src/odemis/driver/kryoz.py
```diff
@@ -0,0 +1,159 @@
...
+            self._socket.sendall(cmd + b'\r\n')
+
+            ans = b''
+            while ans[-2:] != b'\r' and ans[-1:] != b'\n':
+                try:
+                    ans += self._socket.recv(4096)
+                except socket.timeout:
+                    # this is ok. Just means the server didn't send anything.
+                    # Keep listening
+                    pass
```
Kleijwegt, line 87: Don't you want to raise an error if after 10 attemps the server still doesn't respond?
I don't know the code/configuration of the server that well so maybe that doesn't make any sense.

src/odemis/driver/kryoz.py
```diff
@@ -0,0 +1,159 @@
...
+
+DEFAULT_PORT = 5041
+
+
+class CoolerBasic():
+    '''
+    Basic cooler driver
+    '''
+
+    def __init__(self, name, role, host, port=DEFAULT_PORT, daemon=None, **kwargs):
```
pieleric, line 36: No need to use a constant for the default port. Just put the default port as default value.
Drop the deamon argument, as it's not directly used (it'll go into the kwargs).

src/odemis/driver/kryoz.py
pieleric: Actually, please put this line. That's needed to have the component properly registered, and shared over Pyro.
It should look like:
`super(Cooler, self).__init__(name, role, **kwargs)`

src/odemis/driver/kryoz.py
```diff
@@ -0,0 +1,159 @@
...
+        super(CoolerBasic, self).terminate()
+
+    def _sendQuery(self, cmd):
+        """
+        cmd (byte str): command to be sent to device
+        returns (byte str): answer received from the device
+        raise:
+            IOError if no answer is returned in time
+        """
+        cmd = bytes(cmd, 'utf-8') + b"\r"
```
pieleric, line 75: Usually, we write `cmd.encode(...)`. It has the same behaviour. Could you follow the usual way?
BTW, does the device really accepts UTF-8? Often, for such low-level protocol, the hardware just accepts "ascii". If that's the case, please write it so.
However, the docstring say cmd is bytes. Actually that would make things easier and more flexible. => Change the callers, and remove the conversion to bytes.

src/odemis/driver/kryoz.py
pieleric: No, just uncomment it. Seeing every command sent to the device is super useful for debugging.

src/odemis/driver/kryoz.py
```diff
@@ -0,0 +1,159 @@
...
+        with self._net_access:
+            #logging.debug("Sending: %s", to_str_escape(cmd))
+            self._socket.sendall(cmd + b'\r\n')
```
pieleric, line 78: I doubt you need to also add "\r\n" as you already added "\r". Please pick only one place.

src/odemis/driver/kryoz.py
```diff
@@ -0,0 +1,159 @@
...
+
+            ans = b''
+            while ans[-2:] != b'\r' and ans[-1:] != b'\n':
```
pieleric, line 81: I honestly don't get how this works. How could `ans[-2:] == b'\r'`? Only if it receives a single bytes "\r" ? That's odd.
Shouldn't it be `while ans[-2:] != b"\r\n":` ?

src/odemis/driver/kryoz.py
```diff
@@ -0,0 +1,159 @@
...
+    def terminate(self):
+        self._socket.shutdown(socket.SHUT_RDWR)
+        self._socket.close()
+
+        super(CoolerBasic, self).terminate()
+
+    def _sendQuery(self, cmd):
+        """
+        cmd (byte str): command to be sent to device
+        returns (byte str): answer received from the device
```
pieleric, line 71: Usually in such code we return unicode strings (as long as the return value is always a english-like text). That's simpler as otherwise every caller has to convert to unicode.

src/odemis/driver/kryoz.py
pieleric: Indeed, at least put a logging.debug(). For the current implementation, as we are not aiming high, let's just put a log and be done.

src/odemis/driver/kryoz.py
```diff
@@ -0,0 +1,159 @@
...
+                try:
+                    ans += self._socket.recv(4096)
+                except socket.timeout:
+                    # this is ok. Just means the server didn't send anything.
+                    # Keep listening
+                    pass
+
+            logging.debug("Received: %s", to_str_escape(ans))
+
+            return ans.strip()
```
pieleric, line 91: Please write `ans.strip().decode("latin1")`, to immediately return a unicode string.

src/odemis/driver/kryoz.py
```diff
@@ -0,0 +1,159 @@
...
+    def getStatus(self):
+        """
+        Get the status from the device
+
+        returns: list of bool in the order:
+            Connection , Cooling , Executing program , Executing service function,
+            Reporting errors
+        """
+        ans = self._sendQuery("STATUS")
+        # The output is in the format: "SENSORS:100,000|293,554|-2|119,873|0,000|1,237E3"
```
pieleric, line 124: Please update this line, I guess it should be about "STATUS".

pieleric (changes requested):
That's mostly fine for the goal we have. Let's just fix a few minor things and then it should be ready to merge.

src/odemis/driver/kryoz.py
amuskens: I think in the docs, it does actually say [CR] [LF] to terminate.

src/odemis/driver/kryoz.py
amuskens: Ah.. maybe.... it worked like that, but indeed the terminator is actually \r\n

amuskens:
Changes made!

src/odemis/driver/kryoz.py
pieleric: Ok. Then drop line 73.

src/odemis/driver/kryoz.py
```diff
@@ -63,32 +61,32 @@ def terminate(self):
...
-        cmd = bytes(cmd, 'utf-8') + b"\r"
+        cmd = cmd + b"\r"
         with self._net_access:
-            #logging.debug("Sending: %s", to_str_escape(cmd))
+            logging.debug("Sending: %s", to_str_escape(cmd))
             self._socket.sendall(cmd + b'\r\n')
 
             ans = b''
-            while ans[-2:] != b'\r' and ans[-1:] != b'\n':
+            while ans[-2:] != b"\r\n":
```
pieleric, line 79: Be aware however that on such network connection, there might be several responses merged into a single packet. In this case, there could be several messages, or even a command followed by another half one. For handling properly such case you need a buffer and to use .split().
That said, for now, let's assume the device doesn't send messages frequently enough for this to happen. So that's probably fine.

pieleric (changes requested):
Looks almost ready to merge. Just need to drop a line.

amuskens:
Fixed

PR #1584, opened 2020-12-01
plugins cli_rgb: fix filter not applied

pieleric:
Since that the filter is a local axis, the plugin didn't work, as any
change of the filter was reverted by the local axis.

=> Use the local axis to change filter.

Also use user-friendly display of the filter name.

### DIFF
--- a/plugins/cli_rgb.py
+++ b/plugins/cli_rgb.py
@@ -30,35 +30,40 @@ other dealings in the software.
 """
 
 from __future__ import division
+
 from collections import OrderedDict
 from concurrent.futures import CancelledError
 import logging
 from odemis import dataio, model
 from odemis.acq import stream, drift, acqmng
 from odemis.acq.stream import UNDEFINED_ROI
+from odemis.gui.conf import util
 import odemis.gui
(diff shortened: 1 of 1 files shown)


plugins/cli_rgb.py
```diff
@@ -30,35 +30,40 @@
...
 import wx
-from odemis.gui.plugin import Plugin, AcquisitionDialog
 
 
 class RGBCLIntensity(Plugin):
     name = "RGB CL-intensity"
-    __version__ = "1.1"
-    __author__ = "Toon Coenen"
+    __version__ = "1.2"
+    __author__ = "Toon Coenen & Éric Piel"
```
Kleijwegt, line 52: Place holder to ask a question via Slack.

Kleijwegt (commented):
Looks good but since there is some hardware related stuff which I don't understand (regarding the filters) I think someone else should approve this.

plugins/cli_rgb.py
pieleric: Indeed, this is incorrect, on Python2. Fixed.

amuskens (approved):

PR #1585, opened 2020-12-02
driver-tfsbc: log requested shift

philipwinklerdelmic:
It is quite useful for debugging purposes to also log the requested shift in the scan coordinate system (and not only the register values).

### DIFF
--- a/src/odemis/driver/tfsbc.py
+++ b/src/odemis/driver/tfsbc.py
@@ -237,6 +237,7 @@ class BeamShiftController(model.HwComponent):
         """
         :param value (float, float): x, y shift from the center (in m)
         """
+        logging.debug("Requesting shift of %s m." % value)
         try:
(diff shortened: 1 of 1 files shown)


philipwinklerdelmic:
sorry for the crazy branch name

Sabrina85 (approved):

pieleric (commented):
Looks fine.
However, in my experience one even more useful log is to indicate everything we send to the device. I don't know if modbus already does that. If not, I suggest to log inside write_registers(), and possibly also in read_registers().

philipwinklerdelmic:
We are already doing that, it's indeed useful. However, the values that are sent to the device are very different from the values requested. There are a lot of transformations in between, so it's impossible to tell which actual value (in m in the scan coordinate system) was requested. So a second logging message can help a lot.

pieleric:
Good. Just to be clear, I am not arguing against this extra logging, I was just saying we need both 🙂

PR #1587, opened 2020-12-03
driver npmc: fix the Python 3 fix not working on Python 3

pieleric:
commit 59ed1a8f45 (fix instantiation on Python 3) tried to
fix the code for Python 3, but it didn't even work at init.
I guess I mistakenly only tested on Python 2...
Anyway, now it should be properly working.


tepals (approved):
I assume you now tested it on Python 3 and are sure everything works

PR #1590, opened 2020-12-10
gui: temperature display

amuskens:


src/odemis/gui/cont/temperature.py
```diff
@@ -0,0 +1,119 @@
...
+            logging.exception("Could not compute the rate of change of temperature.")
+            return
+
+        # logging.debug(u"Temperature: %f °C, speed is %f °C / s", new_temperature, speed)
+
+        # determine setting the warning
+        if safe_range is not None and (new_temperature > safe_range[1] or new_temperature < safe_range[0]):
+            # temperature is out of range
+            # change colour to red
+            wx.CallAfter(self.temperature_label.SetForegroundColour, wx.RED)
```
pieleric, line 92: Use `gui.FG_COLOUR_ERROR` instead of wx.RED.

src/odemis/gui/cont/temperature.py
```diff
@@ -0,0 +1,119 @@
...
+            logging.warning(text)
+            # display messagebox warning, but only once per minute
+            if time.time() - self._time_last_warning > 60:
+                wx.CallAfter(wx.MessageBox, text, "Warning", wx.ICON_WARNING)
+                self._time_last_warning = time.time()
+
+        elif safe_speed is not None and (speed > safe_speed[1] or speed < safe_speed[0]):
+            # temperature is changing at an unsafe speed
+            # change colour to yellow
+            wx.CallAfter(self.temperature_label.SetForegroundColour, wx.YELLOW)
```
pieleric, line 103: use `gui.FG_COLOUR_WARNING` instead of wx.YELLOW.

src/odemis/gui/cont/temperature.py
```diff
@@ -0,0 +1,119 @@
...
+            text = u"Temperature {temp:.2f}°C is changing at an unsafe rate of {speed:.2f}°C/s!".format(temp=new_temperature, speed=speed)
+            logging.warning(text)
+            # display messagebox warning, but only once per minute
+            if time.time() - self._time_last_speed_warning > 60:
+                wx.CallAfter(wx.MessageBox, text, "Warning", wx.ICON_WARNING)
```
pieleric, line 108: Instead of MessageBox, use notification messages. In Odemis we have the simple `gui.comp.popup.show_message()`

src/odemis/gui/cont/temperature.py
```diff
@@ -0,0 +1,119 @@
...
+                self._time_last_speed_warning = time.time()
+
+        else:
+            # temperature is normal. Set the colour back to normal
+            wx.CallAfter(self.temperature_label.SetForegroundColour, wx.Colour(20, 20, 20))
```
pieleric, line 113: Do not hard-code the colour here. You could use a constant, but I think it's easier/better to store the foreground at init, and restore it here.

src/odemis/gui/cont/temperature.py
```diff
@@ -0,0 +1,119 @@
...
+            # change colour to red
+            wx.CallAfter(self.temperature_label.SetForegroundColour, wx.RED)
+            text = u"Temperature {temp:.2f}°C is outside of the safe operating range!".format(temp=new_temperature)
+            logging.warning(text)
+            # display messagebox warning, but only once per minute
+            if time.time() - self._time_last_warning > 60:
+                wx.CallAfter(wx.MessageBox, text, "Warning", wx.ICON_WARNING)
+                self._time_last_warning = time.time()
+
+        elif safe_speed is not None and (speed > safe_speed[1] or speed < safe_speed[0]):
```
pieleric, line 100: `(speed > safe_speed[1] or speed < safe_speed[0])` -> `not safe_speed[0] <= speed <= safe_speed[1] `

Same for the temperature code.

src/odemis/gui/cont/temperature.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def __init__(self, main_frame, thermostat, sampling_period=1.0, len_history=5):
+        '''
+        main_frame: (wx) main xrc frame of the GUI
+        thermostat: (HwComponent) a component with the VA "temperature"
+        sampling_period: (float) the sampling period of the controller, in seconds.
+            Will update the temperature at this period
+        len_history: (positive int) the length of the temperature history queue.
+            This queue contains the last sampled measurements of temperature and is
+            used to calculate the rate of change as a linear regression
```
pieleric, line 49: Instead of a sampling period + length, you should use a duration.

src/odemis/gui/cont/temperature.py
```diff
@@ -0,0 +1,119 @@
...
+        self._time_last_warning = 0  # time of the last warning
+        self._time_last_speed_warning = 0
+
+        val = self.thermostat.temperature.value
+        self.temperature_history = collections.deque(self._len_temp_q * [val], self._len_temp_q)
+        self._update_temperature_timer = RepeatingTimer(self._temperature_sample_period, self._updateTemperature)
+        self._update_temperature_timer.start()
+
+    def _updateTemperature(self):
+        new_temperature = self.thermostat.temperature.value
```
pieleric, line 71: Let's call this just `temperature`, and everyone should assume this is the current temperature.

src/odemis/gui/cont/temperature.py
```diff
@@ -0,0 +1,119 @@
...
+        self.temperature_history.append(new_temperature)
```
pieleric, line 72: For an example of updating a list while deleting the oldest elements, see the code in `acq.streams._helper.MonochromatorSettingsStream._append()`

src/odemis/gui/model/__init__.py
```diff
@@ -582,7 +584,7 @@ def __init__(self, main):
 
         # VA for autofocus procedure mode
         self.autofocus_active = BooleanVA(False)
-
+
```
pieleric, line 587: Drop this change please ;-)

pieleric (changes requested):

src/odemis/gui/cont/temperature.py
```diff
@@ -0,0 +1,124 @@
...
+                break
+
+        # Get boundaries
+        safe_range = self.thermostat.getMetadata().get(md.MD_FAV_POS_RANGE, None)
+        safe_speed = self.thermostat.getMetadata().get(md.MD_SAFE_SPEED_RANGE, None)
+
+        # Calculate a linear regression of the temperature history to determine the speed
+        l = list(zip(*self.temperature_history))  # convert [(time, temp), ...] to [[time, ...], [temp...]]
+        try:
+            speed, _ = numpy.polyfit(l[0], l[1], 1)
```
pieleric, line 88: This fails for simple cases where the len(l) < 2. It would be relatively easy to trigger if for some reason the temperature hasn't been updated in a while. Such case should *not* show up as an error in the GUI. So I'd suggest to explicitly check for that first, and return immediately in such case.

src/odemis/gui/cont/temperature.py
```diff
@@ -0,0 +1,124 @@
...
+
+        # determine setting the warning
+        if safe_range is not None and not safe_range[0] <= temperature <= safe_range[1] :
+            # temperature is out of range
+            # change colour to red
+            wx.CallAfter(self.temperature_label.SetForegroundColour, gui.FG_COLOUR_ERROR)
+            text = u"Temperature {temp:.2f}°C is outside of the safe operating range!".format(temp=temperature)
+            logging.warning(text)
+            # display messagebox warning, but only once per minute
+            if time.time() - self._time_last_warning > 60:
```
pieleric, line 103: Please put the 60 into a constant named something like `MIN_PERIOD_WARNING`, at the top of the file.

src/odemis/gui/cont/temperature.py
```diff
@@ -0,0 +1,124 @@
...
+
+        elif safe_speed is not None and not safe_speed[0] <= speed <= safe_speed[1] :
+            # temperature is changing at an unsafe speed
+            # change colour to yellow
+            wx.CallAfter(self.temperature_label.SetForegroundColour, gui.FG_COLOUR_WARNING)
+            text = u"Temperature {temp:.2f}°C is changing at an unsafe rate of {speed:.2f}°C/s!".format(temp=temperature, speed=speed)
+            logging.warning(text)
+            # display messagebox warning, but only once per minute
+            if time.time() - self._time_last_speed_warning > 60:
+                comp.popup.show_message(self.main_frame, "Waning", message=text, level=logging.WARNING)
```
pieleric, line 115: "comp.popup" -> "popup". No need to put half of the path.

src/odemis/gui/cont/temperature.py
```diff
@@ -0,0 +1,124 @@
...
+
+        # Calculate a linear regression of the temperature history to determine the speed
+        l = list(zip(*self.temperature_history))  # convert [(time, temp), ...] to [[time, ...], [temp...]]
+        try:
+            speed, _ = numpy.polyfit(l[0], l[1], 1)
+        except:
+            logging.exception("Could not compute the rate of change of temperature.")
+            return
+
+        # logging.debug(u"Temperature: %f °C, speed is %f °C / s", temperature, speed)
```
pieleric, line 93: Let's drop this line.

pieleric (changes requested):

src/odemis/gui/cont/temperature.py
```diff
@@ -27,10 +27,31 @@
...
+    """
+    Set a wx.StaticText foreground colour and font style to bold/unbold
+    label: wx.StaticText object
+    colour: wx.Colour object
+    bold: (bool) true to set to bold
+    """
+
+    label.SetForegroundColour(colour)
+    if bold:
+        font = label.GetFont()
```
pieleric, line 48: You can move this line out of the if block.
You'd recommend also putting the .SetFont() afterwards.

src/odemis/gui/cont/temperature.py
```diff
@@ -84,11 +105,14 @@ def _updateTemperature(self, temperature):
...
-            logging.exception("Could not compute the rate of change of temperature.")
-            return
+        speed = 0
+        # calculate speed (only possible if there is data in the history
+        if len(self.temperature_history) > 1:
+            try:
+                speed, _ = numpy.polyfit(l[0], l[1], 1)
+            except:
+                logging.exception("Could not compute the rate of change of temperature.")
+                return
```
pieleric, line 120: Drop this return (so that the temperature warning can still be shown).

src/odemis/gui/cont/temperature.py
```diff
@@ -27,10 +27,31 @@
 import numpy
 import time
 import logging
+from odemis.gui.util import call_in_wx_main
 import odemis.gui as gui
-from odemis.gui import comp
+from odemis.gui.comp import popup
 import odemis.model as md
 
+MIN_PERIOD_WARNING = 60
```
pieleric, line 35: Add a as a comment `#  s`

src/odemis/gui/cont/temperature.py
```diff
@@ -114,13 +119,11 @@ def _updateTemperature(self, temperature):
...
                 return
 
-        # logging.debug(u"Temperature: %f °C, speed is %f °C / s", temperature, speed)
-
         # determine setting the warning
         if safe_range is not None and not safe_range[0] <= temperature <= safe_range[1] :
             # temperature is out of range
             # change colour to red
-            wx.CallAfter(self.temperature_label.SetForegroundColour, gui.FG_COLOUR_ERROR)
+            set_label_colour(self.temperature_label, gui.FG_COLOUR_ERROR, True)
```
pieleric, line 126: Please, in all these calls write explicitly `bold=True/False`. A lone True/False is really hard to read.

src/odemis/gui/cont/temperature.py
```diff
@@ -71,13 +72,17 @@ def __init__(self, main_frame, thermostat, max_duration=5):
...
+
         self.temperature_label = main_frame.temperature_display
         self._fg_color = self.temperature_label.GetForegroundColour()
         self.temperature_label.Show()
 
         self._time_last_warning = 0  # time of the last warning
         self._time_last_speed_warning = 0
 
+        self.temperature = 5
+
```
pieleric, line 85: What? Some left-over code from debugging?

src/odemis/driver/lakeshore.py
```diff
@@ -451,6 +451,11 @@ def __init__(self, timeout=1):
         self._setpoint = 150  # K
         self._temperature = 100  # K
         self._heating = 0  # enum int 0,1,2, or 3
+        self._temperature_increase = False
+        
+        if os.path.exists("temp_increase.txt"):
```
pieleric, line 456: Don't put it in the `__init__()`, but directly at the `if`. So the user can add/remove the file at any time and the driver follows the requested behaviour.

For the path, currently it checks in the "current directory", which is not super obvious (it depends how you start the back-end). So I'd suggest using `os.path.join(model.BASE_DIRECTORY, ...)`, as we did for the simcam.

src/odemis/gui/cont/temperature.py
```diff
@@ -0,0 +1,151 @@
...
+    Set a wx.StaticText foreground colour and font style to bold/unbold
+    label: wx.StaticText object
+    colour: wx.Colour object
+    bold: (bool) true to set to bold
+    """
+    label.SetForegroundColour(colour)
+    if bold:
+        font = label.GetFont()
+        font.SetWeight(wx.BOLD)
+        label.SetFont(font)
```
pieleric, line 50: You can even write `font.SetWeight(wx.BOLD if bold else wx.NORMAL)`, and make the whole thing just 3 lines ;-)

pieleric (changes requested):
Just a few comments.
Once there are fixed, it should be ready to merge. However, please reshuffle the commits so that the lakeshore changes are in one commit, while the GUI changes are in another commit.

src/odemis/gui/cont/temperature.py
```diff
@@ -0,0 +1,143 @@
...
+
+import wx
+import collections
+import numpy
+import time
+import logging
+from odemis.gui.util import call_in_wx_main
+import odemis.gui as gui
+from odemis.gui.comp import popup
+import odemis.model as md
```
pieleric, line 33: Please just import it as `model`. It makes it easier to copy-paste code.

src/odemis/gui/cont/temperature.py
```diff
@@ -0,0 +1,143 @@
...
+
+MIN_PERIOD_WARNING = 60  # in seconds, represents the time before notifications are shown again
+
+
+@call_in_wx_main
+def set_label_colour(label, colour, bold=False):
```
pieleric, line 39: Now that I think of it, you could just have an argument `weight`, instead of `bold`. Then the caller would call with either wx.BOLD or wx.NORMAL. You wouldn't need the if, and you could avoid the explicit argument name in the callers as it'd be more obvious what we are talking about. (Granted, this is absolute nitpicking!)

pieleric (changes requested):
The code looks mostly fine. However, based on early feedback, a few requests, more at the behavioural level:
 * Both warnings should be red (because orange is too hard to read)
 * Instead of having an "absolute" temperature range, it should be a relative range compared to the .targetTemperature. So I'd suggest to rename MD_FAV_POS_RANGE -> MD_SAFE_REL_RANGE . Then, at init, also check that the component has a .targetTemperature VA.

src/odemis/gui/cont/temperature.py
```diff
@@ -30,22 +30,22 @@
 from odemis.gui.util import call_in_wx_main
 import odemis.gui as gui
 from odemis.gui.comp import popup
-import odemis.model as md
+import odemis.model as model
```
pieleric, line 33: Let's just write `from odemis import model`.

src/odemis/gui/cont/temperature.py
```diff
@@ -112,11 +112,16 @@ def _updateTemperature(self, temperature):
                 logging.exception("Could not compute the rate of change of temperature.")
 
         # determine setting the warning
-        if safe_range is not None and not safe_range[0] <= temperature <= safe_range[1] :
+        if safe_range is not None and hasattr(self.thermostat, "targetTemperature") and \
+            not self.thermostat.targetTemperature.value + safe_range[0] <= temperature <= \
+                self.thermostat.targetTemperature.value + safe_range[1] :
```
pieleric, line 117: Let's read targetTemperature only once, as it can be a little slow.

PR #1593, opened 2020-12-10
driver andorcam2: use recursive lock

pieleric:
It seems it's never called recursively, but with the recovering code,
in the andorshrk, it's best that locks are recursive.

### DIFF
--- a/src/odemis/driver/andorcam2.py
+++ b/src/odemis/driver/andorcam2.py
@@ -743,7 +743,7 @@ class AndorCam2(model.DigitalCamera):
 
         # For temporary stopping the acquisition (kludge for the andorshrk
         # SR303i which cannot communicate during acquisition)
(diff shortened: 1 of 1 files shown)


src/odemis/driver/andorcam2.py
```diff
@@ -743,7 +743,7 @@ def __init__(self, name, role, device=None, emgains=None, shutter_times=None,
 
         # For temporary stopping the acquisition (kludge for the andorshrk
         # SR303i which cannot communicate during acquisition)
-        self.hw_lock = threading.Lock() # to be held during DRV_ACQUIRING (or shrk communicating)
+        self.hw_lock = threading.RLock()  # to be held during DRV_ACQUIRING (or shrk communicating)
```
Kleijwegt, line 746: Maybe a comment on which methods would use the recursive functionality would be nice?
But that is not essential.

Kleijwegt (approved):

PR #1595, opened 2020-12-10
gui: improve file patterns in plugin file selectors

pieleric:


philipwinklerdelmic (approved):
Maybe we could have a constant somewhere like
DEFAULT_WILDCARD = "Any file (*.*)|*.*"

That might make changes in the future easier.

pieleric:
I've been thinking quite hard about the different ways to avoid duplication. I'm not found of the constant, because it'd need to be defined somewhere, and I can't find a good place (maybe just in gui.__init__ with other constants?).
We could also pass `None` and replace `None` by the default wildcards in the actual function. It's not too bad, but it's not very clear in the highest functions what the default is.
We could also just use **kwargs, and pass it on, but it's even more hidden.
I think I'll go with the `None`...

PR #1599, opened 2020-12-11
Fixes to CryoSECOM Overview Map after HW testing

lazem:
- Store starting position as left, top before calling `normalize_rect()`
- Make sure to begin acquisition from starting position
- Use Identity registration in case global shift failed
- Handle ENZEL vibration effects temporarily using `time.sleep()`


src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -67,11 +67,15 @@ def __init__(self, streams, stage, area, overlap, settings_obs=None, log_path=No
         """
         self._future = future
         self._streams = streams
-
+        self._stage = stage
+        self._starting_pos = {'x': area[0], 'y': area[1]}  # left, top
```
pieleric, line 71: Two problems with this:
* You should use `normalized_area` if you want to be sure it's the left,top
* It's not needed to start at the very corner, but only close enough to make sure that the top left of the image will contain this point. I think this is a more general limitation of the current procedure though. To rephrase it, `area` should indicate the area imaged. As the stream has a FoV, the movement of the stage can be smaller. You can think of it as a border around of the area. So the movement area should be "area - (smallest_FoV/2 - overlap_distance)". The overlap distance is removed to represents, pessimistically, the precision of the stage. For small FoVs (eg FM), that is not going to have much effect, but for large FoV (eg EM at low mag), it may end-up going for example from 4x4 -> 3x3 tiles, which could be quite a save in time.

No need to address this point now. Let's do this in a separate task.

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -222,13 +224,15 @@ def _moveToTile(self, idx, prev_idx, tile_size):
...
+        self._future.running_subf = self._stage.moveAbs(m)
         try:
             speed = min(self._stage.speed.value.values()) if model.hasVA(self._stage, "speed") else 10e-6
             # add 1 to make sure it doesn't time out in case of a very small move
             t = math.hypot(abs(idx_change[0]) * tile_size[0] * overlap,
                            abs(idx_change[1]) * tile_size[1] * overlap) / speed + 1
-            f.result(t)
+            self._future.running_subf.result(t)
+            # TODO: instead of sleeping, handle vibration effects
+            time.sleep(1)
```
pieleric, line 235: Well, it should be a FIXME instead of a TODO, because it's really bad.
I think that it should be either handled in the stage driver (eg, add the sleep() in smaract.5DoF), or the stage configuration (ie, change some parameters in the microscope file or the controller). First, more testing is needed to understand what are these vibrations exactly, and how can they be handled most properly.

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -508,7 +515,11 @@ def _stitchTiles(self, da_list):
         st_data = []
         logging.info("Computing big image out of %d images", len(da_list))
         # TODO: Do this registration step in a separate thread while acquiring
-        das_registered = stitching.register(da_list)
+        try:
+            das_registered = stitching.register(da_list, method=REGISTER_GLOBAL_SHIFT)
+        except ValueError:
+            logging.warning("Registration with global_shift failed, retrying with identity registrar..")
```
pieleric, line 521: Please indicate the exception message in the warning. Also change the ".." to either one or three periods.

pieleric (changes requested):

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -441,6 +445,9 @@ def _acquireTiles(self):
         da_list = []  # for each position, a list of DataArrays
         prev_idx = [0, 0]
         i = 0
+        # Make sure to begin from starting position
+        self._stage.moveAbs(self._starting_pos)
```
pieleric, line 449: This should wait for the move to complete.
Ideally, it'd be the same code as in _moveToTile(), but otherwise you could just use moveAbsSync()

src/odemis/acq/stitching/_tiledacq.py
pieleric, line 71: This is left, *bottom*, as y goes from small (bottom) to big (top). I guess if you want to start at left/top, you should use area[3] (or actually normalized_area[3]).

lazem: I got `starting_pos` from `util.normalize_rec` This should return ltrb, right?

pieleric: To clarify, normalize_rec returns minx, miny, maxx, maxy. In physical coordinates, the Y axis goes from bottom to top. So minx is the bottom.

PR #1600, opened 2020-12-11
driver simsem: adjust ltrb to not fail generated image

lazem:


src/odemis/driver/simsem.py
```diff
@@ -464,12 +464,12 @@ def _simulate_image(self):
...
-            elif ltrb[2] > shape[1] - 1:
-                ltrb[0] -= ltrb[2] - (shape[1] - 1)
+            elif ltrb[2] > shape[1]:
+                ltrb[0] -= ltrb[2] - shape[1]
             if ltrb[1] < 0:
                 ltrb[1] = 0
-            elif ltrb[3] >= shape[0] - 1:
-                ltrb[1] -= ltrb[3] - (shape[0] - 1)
+            elif ltrb[3] >= shape[0]:
+                ltrb[1] -= ltrb[3] - shape[0]
```
lazem, line 472: Just a tiny modification without it left and top are equal -1. Tested on all simulators using simsem.

pieleric (changes requested):
No. Firstly, there is no test case showing the issue.
More importantly, it seems that the problem is that ltrb was supposed to be the first and last index. So for an image of res 256, it should generate 0 and 255. However it generates 0 and 256. So it's not a matter of updating the borders, but more about computing better the last indices (r and b).

I'll send a new PR.

pieleric:
I made PR #1602, which should deal more correctly with the issue.

PR #1601, opened 2020-12-14
Updates TMCM Canopen driver

philipwinklerdelmic:
* new adapter needs lower baudrate
* larger timeout required for referencing
* external sensor available --> change homing procedure
* new stepsize


src/odemis/driver/tmcm.py
```diff
@@ -3232,7 +3232,7 @@ def _cancelCurrentMove(self, future):
                 logging.debug("Cancelling failed")
             return future._was_stopped
 
-    def _doReference(self, future, axes, timeout=30):
+    def _doReference(self, future, axes, timeout=300):
```
pieleric, line 3235: It's a bit of sneaky way to put a constant. I don't especially mind. However, please note that you *could* make it a little bit more clever, by using the same  way we did for the USB version: use the axis range * referencing speed to compute the maximum duration time. You can see it in `._waitReferencingStd()`.

PR #1602, opened 2020-12-15
driver simsem: fix acquisition failure when no drift

pieleric:
Commit 215b61b52 fixed a bug but introduced another one.
Instead of calculating the last index, we were computing the
pixel just after. This caused the "protection" code to do the
wrong thing.

Also add a test triggering the issue.


src/odemis/driver/test/simsem_test.py
```diff
@@ -456,5 +443,51 @@ def test_focus(self):
...
+                cls.sed = child
+            elif child.name == CONFIG_SCANNER["name"]:
+                cls.scanner = child
+            elif child.name == CONFIG_FOCUS["name"]:
+                cls.focus = child
+
+    def test_shift(self):
+        """
+        check that .shift works
+        This only works o nthe "drifting" version because on this version the
```
lazem, line 467: only works "on the"

lazem (changes requested):
There's a small typo. 
But worked fine on all the simulators I tried.

PR #1603, opened 2020-12-15
Kryoz fixes

amuskens:


pieleric (changes requested):
No.
`IOError()` only takes one argument, a string. So you need to build that string before passing it, for example like this `IOError("...%s" % (ans,))`

PR #1604, opened 2020-12-15
gui util img: fix merge_screen() only returning original im

pieleric:
In commit e7789884b3 (add format_bgra_to_rgb() function), merge_screen()
was updated to make use of the new function, but that was incorrectly
done, as the original code was a little confusing.

It needs to return the update *background*, not the input im.

Also update the docstring and test cases to indicate now it returns a
RGBA (instead of RGB or RGBA). The actual calling code in Odemis doesn't
care.

### DIFF
--- a/src/odemis/gui/util/test/img_test.py
+++ b/src/odemis/gui/util/test/img_test.py
@@ -1403,19 +1403,20 @@ class TestOverviewFunctions(unittest.TestCase):
         opt = model.DataArray(200 * numpy.ones((10, 10, 3), dtype=numpy.uint8), md)
         sem = model.DataArray(10 * numpy.ones((10, 10, 3), dtype=numpy.uint8), md)
         merged = merge_screen(opt, sem)
+        self.assertEqual(merged.shape, (10, 10, 4))
         # Test if at least one element of ovv image is different from original images
-        # self.assertEqual(merged.shape, (10, 10, 3))
-        self.assertTrue(numpy.all(merged >= opt))
-        self.assertTrue(numpy.any(merged != opt))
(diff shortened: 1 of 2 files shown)


src/odemis/gui/util/test/img_test.py
```diff
@@ -1403,19 +1403,20 @@ def test_merge(self):
...
+        self.assertTrue(numpy.any(merged_rgb != opt))
+        self.assertTrue(numpy.all(merged_rgb >= sem))
+        self.assertTrue(numpy.any(merged_rgb != sem))
 
         # Two very bright images should give a complete white
-        opt = model.DataArray(250 * numpy.ones((10, 10, 3), dtype=numpy.uint8), md)
-        sem = model.DataArray(250 * numpy.ones((10, 10, 3), dtype=numpy.uint8), md)
+        opt = model.DataArray(250 * numpy.ones((10, 12, 3), dtype=numpy.uint8), md)
+        sem = model.DataArray(250 * numpy.ones((10, 12, 3), dtype=numpy.uint8), md)
         merged = merge_screen(opt, sem)
```
Kleijwegt, line 1417: Don't we want to also test what happens when two images with transparency 0 are merged? --> Dark image or white image

pieleric: The idea of the function is to do "background += im" So if the background image is transparent, it should stay transparent. Maybe we could also check if background is opaque but im is fully transparent. I guess this is going to work, however, we do not use this behaviour currently in the code, and the main goal of this PR is to fix a regression. So I'd rather not introduce new test case in this PR.
We can add more test cases in a separate PR.

PR #1608, opened 2020-12-22
Update README.md

Kleijwegt:


README.md
```diff
@@ -1,78 +1,79 @@
...
-## Testing
-To test the software, there are several unit-test classes in each directory (in their `test/` sub-directory). There are also a few example microscope configuration file in the `install/linux/usr/share/odemis/`.
-
-To run all the tests, you can call `util/runtests.sh`.
+<p align="right">
+  <img  src="./src/odemis/gui/img/logo_h30.png">
+</p>
+
+# Odemis
+Odemis (_Open Delmic Microscope Software_) is the open-source microscopy software of [Delmic B.V](https://www.delmic.com) . Odemis is used for controlling microscopes of Delmic and the Odemis viewer allows to load previous experimental data for visualization, analysis and export.
```
tepals, line 6: According to the user manual it stands for Open Delmic Microscopy Suite. Although I do see Open Delmic Microscope Software in other places.

tepals: I also found this quote by googling Odemis: 

> These systems are also accompanied by ODEMIS, open-source software designed to navigate the integrated microscopy systems and to optimize image overlay.

You could change your second sentence to something like:

> Odemis is designed to navigate and control Delmic's integrated microscopy systems and to optimize image overlay. The Odemis viewer allows to load previous experimental data for visualization, analysis and export.

Kleijwegt: Maybe this is a good option for the first two sentences?

> Odemis (_Open Delmic Microscope Software_) is the open-source microscopy software of [Delmic B.V](https://www.delmic.com), which is accompanied with every microscope system of Delmic. Odemis is designed to navigate and control Delmic's integrated microscopy systems and to optimize image overlay. The Odemis viewer allows to load previous experimental data for visualization, analysis and export.

tepals: Sounds good

pieleric: > According to the user manual it stands for Open Delmic Microscopy Suite. Although I do see Open Delmic Microscope Software in other places.

I think that originally I had though of the grand "Suite". However, as it turned out that "Software" made more sense and was more correct, we used this instead. I think the most important is to be consistent. So keep "Software".

pieleric (changes requested):
The new text is much better.
Could you just adjust your commit to keep the same line ending. For now, it looks like every line has been modified.

tepals:
> Could you just adjust your commit to keep the same line ending. For now, it looks like every line has been modified.

@pieleric Why does this README have CRLF line ending, while most files in the repo have LF line ending?

pieleric:
> @pieleric Why does this README have CRLF line ending, while most files in the repo have LF line ending?
The line ending is defined by whoever committed the file first. There is no good reason for CRLF vs LF. Typically in Linux we have LF endings, so that's what most of the files have. Nowadays, pretty much all software handle everything fine. So I don't really care.

PR #1609, opened 2020-12-23
Add new authors to setup.py

tepals:


setup.py
```diff
@@ -116,8 +116,8 @@ def _get_version_git():
 dist = setup(name='Odemis',
              version=VERSION,
              description='Open Delmic Microscope Software',
-             author=u'Éric Piel, Rinze de Laat, Kimon Tsitsikas, Philip Winkler, Anders Muskens, Sabrina Rossberger, Thera Pals, Victoria Mavrikopoulou',
-             author_email='piel@delmic.com, laat@delmic.com, tsitsikas@delmic.com, winkler@delmic.com, muskens@delmic.com, rossberger@delmic.com, pals@delmic.com, mavrikopoulou@delmic.com',
+             author=u'Éric Piel, Rinze de Laat, Kimon Tsitsikas, Philip Winkler, Anders Muskens, Sabrina Rossberger, Thera Pals, Victoria Mavrikopoulou, Kornee Kleijwegt, Bassim Lazem',
+             author_email='piel@delmic.com, laat@delmic.com, tsitsikas@delmic.com, winkler@delmic.com, muskens@delmic.com, rossberger@delmic.com, pals@delmic.com, mavrikopoulou@delmic.com, kleijwegt@delmic.com, lazem@delmic.com',
```
Kleijwegt, line 120: @pieleric Do the mail adresses of Rinze, Kimon (and Anders in the near future) still work? And if no is it maybe not better to replace those with something else? (info@delmic.com for example)

pieleric: Indeed, they don't work. That said, one would wonder what's the point of such addresses anyway. Either it's to be able to link some passed work (eg, commits) to a author, in which case, keeping the address would make sense, or it's to be able to contact an author directly by email, which I wonder why we would want to do that. In this second case, it'd make more sense to put *only* info@ and support@...  In general, anyway it feels more like "please, spam me" than anything useful. ;-)

PR #1610, opened 2020-12-30
gui comp: Also show CLStreams using "screen" operator

pieleric:
They are also typically shown using tint, several of them together (eg,
using some RGB filters), and so it makes more sense to add the pixel
values, like for the FluoStreams, instead of averaging.

Note that in case there is only one CLStream, it's always shown first,
so it doesn't affect the behaviour in this case.

### DIFF
--- a/src/odemis/gui/comp/viewport.py
+++ b/src/odemis/gui/comp/viewport.py
@@ -442,45 +442,29 @@ class MicroscopeViewport(ViewPort):
                 len(self._view.stream_tree) >= 2
         ):
             streams = self._view.getStreams()
-            all_opt = all(isinstance(s, (FluoStream, StaticFluoStream)) for s in streams)
+            all_opt = all(isinstance(s, (FluoStream, StaticFluoStream, CLStream)) for s in streams)
 
             # If all images are optical, assume they are merged using screen blending and no
             # merge ratio is required
             if all_opt:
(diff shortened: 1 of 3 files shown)


src/odemis/gui/comp/viewport.py
```diff
@@ -442,45 +442,29 @@ def _checkMergeSliderDisplay(self):
...
-                    self.bottom_legend.set_stream_type(wx.LEFT, streams[0].acquisitionType.value)
-                    self.bottom_legend.set_stream_type(wx.RIGHT, streams[1].acquisitionType.value)
-
+                def get_stream_prio(s):
+                    if isinstance(s, (FluoStream, StaticFluoStream, CLStream)):
+                        return 0
+                    if isinstance(s, EMStream):
+                        return 1
+                    if isinstance(s, SpectrumStream):
+                        return 2
```
Kleijwegt, line 463: Why are multiple if statements preferred here over elif statements?

pieleric: No specific reason. Technically, this works, as there are "returns", but `elif` are more explicit. Moreover, there should have been an `else`, to handle the case the stream is nothing defined.
I've updated the commit to do this properly.

Kleijwegt (approved):

PR #1617, opened 2021-01-07
doc: update installation instructions on Windows

pieleric:


doc/develop/extending.rst
```diff
@@ -175,9 +175,9 @@ Creating the Odemis environment
 
 Install `Anaconda <https://www.anaconda.com/distribution/>`_ with Python 2.7 32 bits.
 It's also possible to use the version with Python 3.7, 32 or 64 bits, in which case
-you should follow the second set of instructions. Also note that Odemis doesn't
-work with Python 3.7 yet (only Python 3.5 or 3.6) but that it's fine as anaconda
-supports multiple versions of Python simultaneously.
+you should follow the second set of instructions. Also note that typically we validate
+Odemis on Python 3.6, and that why the instructions use the fact that Anaconda
+supports multiple versions of Python simultaneously to install Python 3.6.
```
Kleijwegt, line 180: So basically we support both 3.6 and 3.7 isn't it? For me it is a bit unclear whether I should upgrade if I have python 3.6 currently installed or not.
Maybe it is better if we focus on 3.7 but write a last sentence that there is currently still (nonofficial) support for 3.6?
The way is is currently written is not very clear.

doc/develop/extending.rst
```diff
@@ -175,9 +175,9 @@ Creating the Odemis environment
 
 Install `Anaconda <https://www.anaconda.com/distribution/>`_ with Python 2.7 32 bits.
 It's also possible to use the version with Python 3.7, 32 or 64 bits, in which case
```
Kleijwegt, line 177: Shouldn't this line:
"conda create -y --name odemisdev python==3.6.8" @ linenumber 202
Also be updated/changed to python 3.7?

Kleijwegt (changes requested):

doc/develop/extending.rst
pieleric: I didn't write very clearly. At the time of the writing the Anaconda installer came with Python 3.7 by default (and now with v3.8). However, we've tested fully only on v3.6, so it's best to use v3.6. 
I've update the text to hopefully make it clearer.

doc/develop/extending.rst
pieleric: As described in my previous response, that's actually on purpose, to change the Python interpreter to a version for which Odemis has been well tested.

Kleijwegt (approved):

PR #1618, opened 2021-01-07
gui comp: allow changing merge ratio between two Fluo or CL streams

pieleric:
For simplification, we had said that such image is always shown
all together (ie, merge ratio == 0.5). It might still be handy in some
cases to be able to change the display ratio finely between two images.

Note that if the ratio != 0.5 and the background is not black, the
background will be seen a little bit. As this only happens in some
internal debug mode, that shouldn't matter.

### DIFF
--- a/src/odemis/gui/comp/canvas.py
+++ b/src/odemis/gui/comp/canvas.py
@@ -935,6 +935,12 @@ class BitmapCanvas(BufferedCanvas):
 
         if images:
             n = len(images)
+
+            im_last = images[-1]
+            if isinstance(im_last, tuple):
+                im_last = im_last[0][0]  # first tile
+            bm_last = im_last.metadata["blend_mode"]
+
             for i, im in enumerate(images):
                 if isinstance(im, tuple):
                     first_tile = im[0][0]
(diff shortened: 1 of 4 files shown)


src/odemis/gui/comp/canvas.py
```diff
@@ -945,15 +951,39 @@ def _draw_merged_images(self, ctx, interpolate_data=False):
...
-                        merge_ratio = self.merge_ratio
                 else:
-                    merge_ratio = 1 - i / n
+                    # If there are all "screen" (= last one is screen):
+                    # merge ratio   im0   im1
+                    #     0         1      0
+                    #    0.25       1      0.5
+                    #    0.5        1      1
+                    #    0.75       0.5    1
+                    #     1         0      1
```
VictoriaDel, line 964: Is this correct?

VictoriaDel (changes requested):
In the _draw_merged_images() method the docstring, as well as the comments explaining the idea of merging, should be updated.

src/odemis/gui/comp/canvas.py
pieleric: Yes, at least, this is what the code does, and this gives a "intuitive" behaviour. That's a little odd, that's why I've described it in a long comment ;-)

PR #1619, opened 2021-01-08
Drop MD_WL_POLYNOMIAL : convert it to a MD_WL_LIST when loading a file

VictoriaDel:


src/odemis/acq/calibration.py
```diff
@@ -316,7 +315,7 @@ def apply_spectrum_corrections(data, bckg=None, coef=None):
             try:
                 wl_bckg = spectrum.get_wavelength_per_pixel(bckg)
             except KeyError:
-                raise ValueError("Found no MD_WL_* metadata in background image.")
+                raise ValueError("Found no MD_WL_LIST metadata in background image.")
```
Kleijwegt, line 318: Maybe adding a reason why that is bad to the error message would be nice.
I don't know the code/context but something like:
> wavelenght information is required to do XXXXX"?

pieleric: That's quite a little bit out of scope from this task.
The error message is not really intended to the end-user. The GUI could catch it, and report the background data is not fitting. That said, I don't think it's handled for now. You could make the message indeed a little less cryptic by writing "Found no spectrum metadata (MD_WL_LIST) in the background image".

src/odemis/dataio/hdf5.py
```diff
@@ -541,11 +533,11 @@ def _read_image_info(group):
...
     # (bandwidth of the monochromator), which is a tuple of (wl offset, wl offset + bandwidth).
 
-    # Spectrograph: The metadata includes MD_WL_POLYNOMIAL (conversion from px to wl), where
+    # Spectrograph: The metadata includes metadata in polynomial form (conversion from px to wl), where
     # the first entry corresponds to the wl offset "a" and the second to the coefficient "b" for the
     # polynomial of first order (wl = a + bx).
-    # Based on polynomial the wavelength values for all px can be calculated elsewhere
-    # (see odemis.util.spectrum.get_wavelength_per_pixel).
+    # Based on polynomial the wavelength values for all px can be calculated, converted to a list
+    # and saved in MD_WL_LIST.
```
pieleric, line 540: I'd suggest to make the text clearer that polynomial form is rare. Normally it's stored in list form. Something like:
```
Spectrograph: The metadata is normally stored in list form (of the same length as the C dimension). For old files, it sometimes had metadata in polynomial form, where the first entry corresponds to the wl offset "a" and the second to the coefficient "b" for the polynomial of first order (wl = a + bx). In such case, based on the polynomial, the wavelength values for all px are calculated, and saved in MD_WL_LIST.
```

src/odemis/dataio/test/tiff_test.py
```diff
@@ -204,7 +204,6 @@ def testExportCube(self):
                     model.MD_BPP: 12,
                     model.MD_BINNING: (1, 1), # px, px
                     model.MD_PIXEL_SIZE: (1e-6, 2e-5), # m/px
-                    model.MD_WL_POLYNOMIAL: [500e-9, 1e-9], # m, m/px: wl polynomial
```
pieleric, line 207: Please replace it with MD_WL_LIST.

src/odemis/dataio/test/tiff_test.py
```diff
@@ -272,7 +271,6 @@ def testExportSpatialCube(self):
                     model.MD_BPP: 12,
                     model.MD_BINNING: (1, 1),  # px, px
                     model.MD_PIXEL_SIZE: (1e-6, 2e-5, 3e-4),  # m/px
-                    model.MD_WL_POLYNOMIAL: [500e-9, 1e-9],  # m, m/px: wl polynomial
```
pieleric, line 275: Please replace it with MD_WL_LIST.

src/odemis/dataio/test/tiff_test.py
```diff
@@ -355,7 +353,6 @@ def testExportNoWL(self):
                     model.MD_BPP: 12,
                     model.MD_BINNING: (1, 1), # px, px
                     model.MD_PIXEL_SIZE: (1e-6, 2e-5), # m/px
-                    model.MD_WL_POLYNOMIAL: [0], # m, m/px: missing polynomial
```
pieleric, line 358: It should be `MD_WL_LIST = []`. That's what the hardware does when the wavelength list is unknown.

src/odemis/dataio/test/hdf5_test.py
```diff
@@ -522,8 +516,6 @@ def testReadAndSaveMDSpec(self):
                      model.MD_BPP: 12,
                      model.MD_BINNING: (1, 1), # px, px
                      model.MD_PIXEL_SIZE: (1e-6, 1e-6),  # m/px
-                     #model.MD_WL_POLYNOMIAL: [500e-9, 1e-9], # m, m/px: wl polynomial
-                     model.MD_WL_LIST: [500e-9 + i * 1e-9 for i in range(sizes[1][-1])],
```
pieleric, line 526: That MD_WL_LIST should stay, right?

src/odemis/dataio/test/hdf5_test.py
```diff
@@ -191,14 +190,14 @@ def testExportCube(self):
...
         size = (512, 256)
+        sizes = [(512, 256), (100, 110, 1, 1, 200)]
         metadata3d = {model.MD_SW_VERSION: "1.0-test",
                     model.MD_HW_NAME: "fake spec",
                     model.MD_DESCRIPTION: "test3d",
                     model.MD_ACQ_DATE: time.time(),
                     model.MD_BPP: 12,
                     model.MD_BINNING: (1, 1), # px, px
                     model.MD_PIXEL_SIZE: (1e-6, 2e-5), # m/px
-                    model.MD_WL_POLYNOMIAL: [500e-9, 1e-9], # m, m/px: wl polynomial
```
pieleric, line 201: Please replace it with MD_WL_LIST.

src/odemis/dataio/test/hdf5_test.py
```diff
@@ -191,14 +190,14 @@ def testExportCube(self):
         dtype = numpy.uint16
         size3d = (512, 256, 220) # X, Y, C
         size = (512, 256)
+        sizes = [(512, 256), (100, 110, 1, 1, 200)]
```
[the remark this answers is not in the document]
VictoriaDel: It' s needed to define MD_WL_LIST. So I will just replace MD_WL_POLYNOMIAL with MD_WL_LIST. That's what I initially did but I wasn't sure if we need to keep the MD_WL in the metadata if it's not used. Anyway, I will add MD_WL_LIST !!

src/odemis/dataio/hdf5.py
```diff
@@ -375,22 +376,13 @@ def _add_image_info(group, dataset, image):
                                 "it will not be saved: %s", ex)
 
         # Wavelength (for spectrograms)
-        if ("C" in dims and
-            set(image.metadata.keys()) & {model.MD_WL_LIST, model.MD_WL_POLYNOMIAL}):
+        if "C" in dims and set(image.metadata.keys()) & {model.MD_WL_LIST}:
```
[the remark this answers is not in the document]
VictoriaDel: yes that would look indeed much better. I agree with you!

PR #1620, opened 2021-01-11
driver-tmcm: improved interface for configuration files of canopen driver

philipwinklerdelmic:
* allow relative path for datasheet
* make param_file optional

### DIFF
--- a/src/odemis/driver/tmcm.py
+++ b/src/odemis/driver/tmcm.py
@@ -2590,12 +2590,12 @@ class CANController(model.Actuator):
     """
 
     def __init__(self, name, role, channel, node_id, datasheet, axes, ustepsize,
-                 param_file, rng=None, unit=None, refproc=None, **kwargs):
+                 param_file=None, rng=None, unit=None, refproc=None, **kwargs):
         """
         channel (str): can port name, on linux, this is typically "can0". For testing
             with the simulator, use "fake".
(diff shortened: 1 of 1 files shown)


src/odemis/driver/tmcm.py
```diff
@@ -2647,6 +2647,12 @@ def __init__(self, name, role, channel, node_id, datasheet, axes, ustepsize,
         if not (refproc == REFPROC_STD or refproc is None):
             raise ValueError("Reference procedure %s unknown" % (refproc,))
 
+        # Get path of datasheet
+        if datasheet[0] != "/":
+            # it's a relative path
+            filedir = os.path.dirname(__file__)
+            datasheet = os.path.join(filedir, datasheet)
```
pieleric, line 2654: I'd suggest the same code as we use already in the andorcam2 (and other drivers):
```python
if not os.path.isabs(datasheet):
    datasheet = os.path.join(os.path.dirname(__file__), datasheet)
```

pieleric (changes requested):
Looks good. I have only a minor comment about using "os.path.isabs".

src/odemis/driver/tmcm.py
philipwinklerdelmic: That's indeed cleaner. I didn't know about the isabs function, very useful!
I pushed the change.

PR #1621, opened 2021-01-11
canopen driver: move example tsv file

philipwinklerdelmic:
* file goes in install/linux/usr/share, automatically install to /usr/share in setup.py


pieleric (commented):
No, I don't think it should be moved to this directory. /usr/share/odemis is the standard folder where we put the installation files. So the goal is to keep it empty in the package, so that every file there corresponds to the actual system.
If you want to have this file installed on every system to a fixed location, then move it to /usr/share/odemis/example/, or .../hwtest/. Or is it because you want to avoid having to copy this file on every FASTEM installation?

philipwinklerdelmic:
I understood that we wanted this file to end up in /usr/share/odemis on every FastEM installation. We could also do it manually or in the ubuntu-delmic setup script, but I thought that would be the easiest way.
@Sabrina85 Maybe I misunderstood? What do you think?

pieleric:
Oh, I see. Indeed, we want this file eventually in /usr/share/odemis. However, not by being packaged inside Odemis. Instead it should be copied there at the same time as we copy and configure the microscope files. So a copy of this file will go to the sonic-odm repo. That said I think it's still a good idea to move it to the example or hwtest folder. It's less hidden than in the test folder.

philipwinklerdelmic:
Ok, I'll just move it to one of these folders without changing the setup.py

philipwinklerdelmic:
Removed changes to setup.py

pieleric (changes requested):
Please, move the TSV file to a subdirectory. Either in hwtest/, or sim/ .

philipwinklerdelmic:
moved it to the subdirectory

PR #1623, opened 2021-01-12
XT client add support for acquiring SEM images

tepals:
- added time estimation for image acquisition
- added get and set resolution from xtadapter
- added Detector and Dataflow
- xt image acquisition with threads
- moved commented out code in Scanner to end of class

### DIFF
--- a/src/odemis/driver/test/xt_client_test.py
+++ b/src/odemis/driver/test/xt_client_test.py
@@ -28,6 +28,8 @@ import os
 import time
 import unittest
 
+from odemis import model
+
 from odemis.driver import xt_client
 from odemis.driver.xt_client import DETECTOR2CHANNELNAME
 from odemis.model import ProgressiveFuture
@@ -43,10 +45,12 @@ CONFIG_STAGE = {"name": "stage", "role": "stage",
                 "inverted": ["x"],
                 }
 CONFIG_FOCUS = {"name": "focuser", "role": "ebeam-focus"}
+CONFIG_DETECTOR = {"name": "detector", "role": "se-detector", "channel_name": "electron1"}
 CONFIG_SEM = {"name": "sem", "role": "sem", "address": "PYRO:Microscope@192.168.31.130:4242",
               "children": {"scanner": CONFIG_SCANNER,
                            "focus": CONFIG_FOCUS,
                            "stage": CONFIG_STAGE,
+                           "detector": CONFIG_DETECTOR,
                            }
               }
 
@@ -68,6 +72,8 @@ class TestMicroscope(unittest.TestCase):
                 cls.efocus = child
             elif child.name == CONFIG_STAGE["name"]:
                 cls.stage = child
+            elif child.name == CONFIG_DETECTOR["name"]:
+                cls.detector = child
 
     def setUp(self):
         if TEST_NOHW:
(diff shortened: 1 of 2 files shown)


Sabrina85:
Do you also have test cases to check the acquisition and dataflow?

src/odemis/driver/xt_client.py
```diff
@@ -1339,8 +1691,8 @@ def __init__(self, name, role, parent, **kwargs):
         self._updatePosition()
 
         # Refresh regularly the position
-        self._pos_poll = util.RepeatingTimer(5, self._refreshPosition, "Position polling")
-        self._pos_poll.start()
+        # self._pos_poll = util.RepeatingTimer(5, self._refreshPosition, "Position polling")
+        # self._pos_poll.start()
```
tepals, line 1695: @pieleric What should I do with these lines?

tepals:
> Do you also have test cases to check the acquisition and dataflow?

I just added them.

src/odemis/driver/xt_client.py
```diff
@@ -870,125 +909,45 @@ def __init__(self, name, role, parent, hfw_nomag, **kwargs):
...
+        self.scale = model.TupleContinuous(self._scale,
+                                           [(1, 1), self._shape],
+                                           unit="",
+                                           cls=(int, float),  # int; when setting scale the GUI returns a tuple of ints.
+                                           setter=self._setScale)
+        self.scale.subscribe(self._onScale, init=True)  # to update metadata
+
+        # If scaled up, the pixels are bigger
+        pxs_scaled = (pxs[0] * self.scale.value[0], pxs[1] * self.scale.value[1])
+        self.parent._metadata[model.MD_PIXEL_SIZE] = pxs_scaled
```
pieleric, line 945: BTW, I've noticed that the API provides a "get_DeltaFWD()" function, which should return the value for a "depthOfField" VA. It's not necessary, but it'd good to keep in mind, especially that is helpful for manual focus. It'd probably would have to be updated every time the horizontalFoV is changed.

pieleric (changes requested):
Looks alright. I have added a comments. Also the unit tests are a little underwhelming. You could have a look (aka "steal") some of the unit tests for the semcomedi. I'd suggest to *at least* test a start/stop/start/stop cycle. Something like: 
* set a big resolution
* start acquiring
* stop before the acquisition is complete
* change to a smaller resolution
* start acquiring
* get data
* stop
* check that the data is the small res

src/odemis/driver/xt_client.py
tepals: Since we turned off event registration to fix the channel problems we had, I don't think listening to the FWDActualValueChanged event works. 

I will test your suggestions.

tepals: I was reading the logs wrong. I saw there was 5 seconds between this call and the next, however that was just because there was nothing called in between. The actual call takes less than 0.5 seconds.

Regarding the events, Michal specifically discouraged the use of events because they are unreliable.

tepals:
@pieleric updated the PR with your comments

src/odemis/driver/test/xt_client_test.py
```diff
@@ -245,16 +249,60 @@ def _compute_expected_duration(self):
...
+        self.scanner.resolution.value = (768, 512)  # px
+        im = self.detector.data.get()
+        # Check that the acquired image has the last set resolution.
+        self.assertEqual(im.shape, self.scanner.resolution.value[::-1])
+        self.assertIn(model.MD_DWELL_TIME, im.metadata)
+        # Set back resolution and dwell time to initial values
+        self.scanner.dwellTime.value = init_dwell_time
+        self.scanner.resolution.value = init_resolution
+
+    def test_live_chance(self):
```
pieleric, line 283: typo? chance -> change

pieleric (approved):
Looks fine. I only could find a tiny typo. Otherwise, I think it's fine to merge.

src/odemis/driver/test/xt_client_test.py
tepals: It's to indicate there's only a small chance this test passes :grinning:. Just kidding I'll update it.

PR #1626, opened 2021-01-13
first implementation xt_client update 2.0

Kleijwegt:


src/odemis/driver/test/xt_client_test.py
```diff
@@ -43,10 +43,10 @@
                 "inverted": ["x"],
                 }
 CONFIG_FOCUS = {"name": "focuser", "role": "ebeam-focus"}
-CONFIG_SEM = {"name": "sem", "role": "sem", "address": "PYRO:Microscope@192.168.31.130:4242",
+CONFIG_SEM = {"name": "sem", "role": "sem", "address": "PYRO:Microscope@192.168.10.2:4242",
               "children": {"scanner": CONFIG_SCANNER,
                            "focus": CONFIG_FOCUS,
-                           "stage": CONFIG_STAGE,
+                           # "stage": CONFIG_STAGE,
```
Kleijwegt, line 49: I will remove these changes as soon as I have performed the last testing (after a first round of review)

src/odemis/driver/test/xt_client_test.py
```diff
@@ -246,8 +245,8 @@ def setUpClass(cls):
                 cls.scanner = child
             elif child.name == CONFIG_FOCUS["name"]:
                 cls.efocus = child
-            elif child.name == CONFIG_STAGE["name"]:
-                cls.stage = child
+            # elif child.name == CONFIG_STAGE["name"]:
+            #     cls.stage = child
```
Kleijwegt, line 249: I will remove these changes as soon as I have performed the last testing (after a first round of review)

src/odemis/driver/xt_client.py
```diff
@@ -790,6 +790,77 @@ def set_use_case(self, state):
...
+        """
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            return self.server.get_aperture_index()
+
+    def set_aperture_index(self, aperture_idx):
+        """
+        Set the current aperture index.
+
+        :param aperture_idx:  (float): Aperture index (range 0 - 14)
```
Sabrina85, line 817: Very picky, but one colon is enough :)

src/odemis/driver/xt_client.py
```diff
@@ -790,6 +790,77 @@ def set_use_case(self, state):
...
+
+        :return (dict with a list of lenght 2): The range of the aperture index for the keyword "range".
+        """
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            return self.server.aperture_index_info()
+
+    def get_beamlet_index(self):
+        """
+        Get the current beamlet index.
```
Sabrina85, line 835: Here you could maybe add some info that the beamlet index is characterized by two values in a grid.

src/odemis/driver/xt_client.py
```diff
@@ -790,6 +790,77 @@ def set_use_case(self, state):
...
+
+        :return (tuple of floats): Beamlet index (range 1 - 8)
+        """
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            # Convert to tuple so the output could be directly compaired (or used as input) for the set method.
```
Sabrina85, line 841: You can drop the comment. It is ok to just convert it.

src/odemis/driver/xt_client.py
```diff
@@ -790,6 +790,77 @@ def set_use_case(self, state):
...
+            return tuple(self.server.get_beamlet_index())
+
+    def set_beamlet_index(self, beamlet_idx):
+        """
+        Set the current beamlet index.
+
+        :param beamlet_idx:  (tuple of floats): Beamlet index (range 1 - 8)
```
Sabrina85, line 848: Remove one space after colon.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -712,6 +711,72 @@ def test_apply_autofocus(self):
         autofocus_state = self.microscope.is_autofocusing(channel)
         self.assertEqual(autofocus_state, False)
 
+    def test_get_mpp_orientation(self):
+        """
+        Simple test, checks if no error occur during calling the method and the return value.
```
Sabrina85, line 716: Rephrase: `Test getting the multiprobe pattern orientation.`

Do rephrase also all the other test doc strings the same way. Then they are all consistent with the rest.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -712,6 +711,72 @@ def test_apply_autofocus(self):
...
+        # TODO K.K. check range with Wilco
+        self.assertTrue(0 < mpp_orientation < 360)
+
+    def test_get_aperture_index(self):
+        """
+        Simple test, checks if no error occur during calling the method and the return value.
+        """
+        aperture_index = self.microscope.get_aperture_index()
+        self.assertIsInstance(aperture_index, float)
+        self.assertTrue(0 < aperture_index < 14)
```
Sabrina85, line 729: The aperture index can be 0 and 14, so you need `<=`.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -712,6 +711,72 @@ def test_apply_autofocus(self):
...
+
+    def test_set_aperture_index(self):
+        """
+        Simple test, checks if no error occur during calling the method.
+        """
+        # TODO Check if setting of the value actually happens and if an int value is also allowed.
```
Sabrina85, line 735: You can remove this TODO here. This indeed needs to be checked. But does not need to be noted here. Radim is working on the TFS implementation to actually return an error in case an aperture cannot be set.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -712,6 +711,72 @@ def test_apply_autofocus(self):
...
+        self.microscope.set_aperture_index(10)
+        aperture_index = self.microscope.get_aperture_index()
+        self.assertEqual(aperture_index, 10)
```
Sabrina85, line 738: I suggest to set another one (0 should work as this is the one for multibeam). Then you can check the aperture actually changed compared to the previous one. In case 10 was already set before you run the test, you would not notice if the set call did not work properly.

Also I suggest you do it the same way as the other tests: 

- start with getting the currently set value
- then change to a different value and check it is set
- then put back the originally set value

src/odemis/driver/test/xt_client_test.py
```diff
@@ -712,6 +711,72 @@ def test_apply_autofocus(self):
...
+        self.assertEqual(len(aperture_index_info["range"]), 2)
+
+    def test_get_beamlet_index(self):
+        """
+        Simple test on checks if no error occur during calling the method and the return value.
+        """
+        beamlet_index = self.microscope.get_beamlet_index()
+        self.assertIsInstance(beamlet_index, tuple)
+        self.assertEqual(len(beamlet_index), 2)
+        self.assertTrue(1 < beamlet_index[0] < 8)
```
Sabrina85, line 756: Same here; you need `<=`

src/odemis/driver/test/xt_client_test.py
```diff
@@ -712,6 +711,72 @@ def test_apply_autofocus(self):
...
+        self.assertTrue(1 < beamlet_index[1] < 8)
+
+    def test_set_beamlet_index(self):
+        """
+        Simple test on checks if no error occur during calling the method and the return value.
+        """
+        # TODO Check if setting of the value actually happens and if an int value is also allowed.
```
Sabrina85, line 763: Same as above, you can remove the TODO.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -712,6 +711,72 @@ def test_apply_autofocus(self):
...
+        self.microscope.set_beamlet_index((4, 4))
+        beamlet_index = self.microscope.get_beamlet_index()
+        self.assertEqual(beamlet_index, (4, 4))
```
Sabrina85, line 766: Also here I suggest you do it the same way as the other tests: 

- start with getting the currently set value
- then change to a different value and check it is set
- then put back the originally set value

src/odemis/driver/test/xt_client_test.py
```diff
@@ -712,6 +711,72 @@ def test_apply_autofocus(self):
         autofocus_state = self.microscope.is_autofocusing(channel)
         self.assertEqual(autofocus_state, False)
 
+    def test_get_mpp_orientation(self):
+        """
+        Simple test, checks if no error occur during calling the method and the return value.
+        """
```
Sabrina85, line 717: For all these tests you need to have these lines first:
```
if self.xt_type != 'xttoolkit':
            self.skipTest("This test needs XTToolkit to run.")

This is needed, as some systems might only run xtlib, but not xttoolkit. However, these methods need xttoolkit. Check for an example the use case test.
```

src/odemis/driver/test/xt_client_test.py
Sabrina85: Still don't push these changes please. These things get easily forgotten to change back.

Sabrina85 (changes requested):
Overall it looks good! 
My main suggestion is to modify the test for setting the values, so that you first get the current value, then change this value to a different one (from the range allowed), check it is correctly set and then switch back to the original one. This is important for following test cases or for continue working on the system. E.g. when single beam mode is set, you want to have back beamlet index (4,4) and aperture 10. You can get some inspiration from the other tests.

src/odemis/driver/test/xt_client_test.py
Kleijwegt: I get your comment, and I will watch it closely. Testing the code is this way much easier since I can just fetch the exact code I am working in.

src/odemis/driver/xt_client.py
```diff
@@ -790,6 +790,77 @@ def set_use_case(self, state):
...
+        """
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            return self.server.get_mpp_orientation()
+
+    def get_aperture_index(self):
+        """
+        Get the current aperture index.
+
+        :return (float): Aperture index (range 0 - 14)
```
pieleric, line 807: Sounds like a int to me... are you sure it's a float?
Also, if there is a "..._get_info()" function that's because the range is not fixed, right? So maybe just say "typical range: 0 - 14), see aperture_index_info.

src/odemis/driver/xt_client.py
```diff
@@ -790,6 +790,77 @@ def set_use_case(self, state):
...
+        """
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            self.server.set_aperture_index(aperture_idx)
+
+    def aperture_index_info(self):
+        """
+        Get the current aperture index information, contains the range (0 - 14).
+
+        :return (dict with a list of lenght 2): The range of the aperture index for the keyword "range".
```
pieleric, line 827: You can write `dict str -> list (of length 2)`

src/odemis/driver/xt_client.py
```diff
@@ -790,6 +790,77 @@ def set_use_case(self, state):
             self.server._pyroClaimOwnership()
             return self.server.set_use_case(state)
 
+    def get_mpp_orientation(self):
+        """
+        Get the current multi probe orientation in degrees.
+
+        :return (float): Multi orientation in degrees
```
pieleric, line 797: For the stage rotation, the values reported are in radians. Are you sure it's here in degrees? 
Also, here, the range would be useful.

src/odemis/driver/xt_client.py
```diff
@@ -790,6 +790,77 @@ def set_use_case(self, state):
...
+        """
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            self.server.set_beamlet_index(*beamlet_idx)
+
+    def beamlet_index_info(self):
+        """
+        Get the current beamlet index information, contains the range in x and y direction (1 - 8).
+
+        :return (dict with a sub-dict which has lists of lenght 2): The range of the beamlet index for the keyword
```
pieleric, line 858: lenght -> length

pieleric (changes requested):
Looks alright. I only have very minor comments. However, just to clarify, you're not planning to use this interface as-is, right? Eventually all these settings should be provided as VAs on the scanner or detectors.

Sabrina85:
> However, just to clarify, you're not planning to use this interface as-is, right? Eventually all these settings should be provided as VAs on the scanner or detectors.

Yes, that is handled in a separate specs and pr to make prs smaller and quicker to merge. This interface is already in progress.

src/odemis/driver/xt_client.py
```diff
@@ -832,20 +832,19 @@ def aperture_index_info(self):
...
-            # Convert to tuple so the output could be directly compaired (or used as input) for the set method.
             return tuple(self.server.get_beamlet_index())
 
     def set_beamlet_index(self, beamlet_idx):
         """
-        Set the current beamlet index.
+        Set the current beamlet index which is represented by the two values in a grid (x,y).
 
-        :param beamlet_idx:  (tuple of floats): Beamlet index (range 1 - 8)
+        :param beamlet_idx: (tuple of floats): Beamlet index (range 1 - 8)
```
Sabrina85, line 847: remove one colon

src/odemis/driver/xt_client.py
```diff
@@ -814,7 +814,7 @@ def set_aperture_index(self, aperture_idx):
         """
         Set the current aperture index.
 
-        :param aperture_idx:  (float): Aperture index (range 0 - 14)
+        :param aperture_idx  (float): Aperture index (range 0 - 14)
```
Sabrina85, line 817: remove one space

src/odemis/driver/test/xt_client_test.py
```diff
@@ -713,62 +713,102 @@ def test_apply_autofocus(self):
...
+        Test setting the aperture index.
         """
-        # TODO Check if setting of the value actually happens and if an int value is also allowed.
-        self.microscope.set_aperture_index(10)
+        if self.xt_type != 'xttoolkit':
+            self.skipTest("This test needs XTToolkit to run.")
+
+        range = tuple(self.microscope.aperture_index_info()["range"])
+        current_aperture_index = self.microscope.get_aperture_index()
+        new_aperture_index = current_aperture_index + 1 if range[0] <=current_aperture_index < range[1] else \
```
Sabrina85, line 746: missing space after <=

src/odemis/driver/test/xt_client_test.py
```diff
@@ -713,62 +713,102 @@ def test_apply_autofocus(self):
...
+                             current_aperture_index -1
```
Sabrina85, line 747: missing space after "-"

src/odemis/driver/test/xt_client_test.py
Sabrina85: I think it is enough if you only check `current_aperture_index < range[1]` as you add `+1`

src/odemis/driver/test/xt_client_test.py
```diff
@@ -713,62 +713,102 @@ def test_apply_autofocus(self):
...
-        Simple test on checks if no error occur during calling the method and the return value.
+        Test setting the beamlet index.
         """
-        # TODO Check if setting of the value actually happens and if an int value is also allowed.
+        if self.xt_type != 'xttoolkit':
+            self.skipTest("This test needs XTToolkit to run.")
+
         self.microscope.set_beamlet_index((4, 4))
         beamlet_index = self.microscope.get_beamlet_index()
         self.assertEqual(beamlet_index, (4, 4))
```
Sabrina85, line 791: These 3 lines look like old code -> remove?

src/odemis/driver/test/xt_client_test.py
```diff
@@ -713,62 +713,102 @@ def test_apply_autofocus(self):
...
 
+        current_index = self.microscope.get_beamlet_index()
+        # Make sure both the x and the y value are changed.
+        new_beamlet_index = (3,3) if 3 not in current_index else (5,5) if 5 not in current_index else (6,6)
```
Sabrina85, line 795: You could also make this one a bit more generic same way as for the aperture. I see that you now selected apertures that also work currently on EA1, but on another system that could be different again. From the client point of view all apertures should work for now. As soon as TFS fixed their code and actually propagate the message that a certain aperture cannot be selected, then we can adapt our code.

Sabrina85 (changes requested):
Only very minor comments.

src/odemis/driver/test/xt_client_test.py
Kleijwegt: That wouldn't work if the current index is exactly the max of the current range.

src/odemis/driver/test/xt_client_test.py
Kleijwegt: Do you have a proposal on how to do that nicely?
I want to make sure that both the x and y beamlet index is changed. So that it works for the aperature (1, 8), (8,8), (3,8) (3,1) etc.

src/odemis/driver/xt_client.py
Kleijwegt: I think the get_info is also there for ranges which are fixed (or should be).
But adding the typically is indeed an improvement.

Kleijwegt: The aperture value is actually send and received as a float. But changing it to an int is indeed easy/clearer for the user.

src/odemis/driver/test/xt_client_test.py
Kleijwegt: It was about removing "range[0] <=" this probably can be removed isn't it? @Kleijwegt

src/odemis/driver/test/xt_client_test.py
```diff
@@ -230,7 +230,6 @@ def test_rotation(self):
         self.assertEqual(self.scanner.rotation.value, init_rotation + 0.01)
         self.scanner.rotation.value = init_rotation
 
-
```
tepals, line 233: Please keep two empty lines between classes. https://www.python.org/dev/peps/pep-0008/#blank-lines

src/odemis/driver/test/xt_client_test.py
```diff
@@ -712,6 +711,122 @@ def test_apply_autofocus(self):
...
+        self.assertTrue(0 <= aperture_index <= 14)
+
+    def test_set_aperture_index(self):
+        """
+        Test setting the aperture index.
+        """
+        if self.xt_type != 'xttoolkit':
+            self.skipTest("This test needs XTToolkit to run.")
+
+        range = tuple(self.microscope.aperture_index_info()["range"])
```
tepals, line 756: Because `range` is a builtin python function, I wouldn't use it as a variable name.

pieleric: Indeed, typically, we use "rng".

src/odemis/driver/xt_client.py
Kleijwegt: Yes TFS returns it in degrees, this should be a by designed defined and set value.
The range is kind of hard to add here. It should be -18 to 18 degrees.
However, due to hardware constrictions it is currently 45 degrees turned, an now defined at 45 degrees. This will stay this  value as it cannot change (unless the internal hardware of the microscope changes).

Kleijwegt:
Sorry in hindsight I should have used a separate development and PR branch. I didn't think this PR would receive so much quic and ad rem reviews.
Thanks for that!

src/odemis/driver/test/xt_client_test.py
```diff
@@ -712,6 +712,120 @@ def test_apply_autofocus(self):
...
+    def test_set_beamlet_index(self):
+        """
+        Test setting the beamlet index.
+        """
+        if self.xt_type != 'xttoolkit':
+            self.skipTest("This test needs XTToolkit to run.")
+
+        current_index = self.microscope.get_beamlet_index()
+        # Make sure both the x and the y value are changed.
+        new_beamlet_index = (3, 3) if 3 not in current_index else (5, 5) if 5 not in current_index else (6, 6)
```
pieleric, line 804: Is there a reason you use the same index in x and y? I'd tend to prefer different ones, in a test case, so that if for some reasons x and y are inverted, it's more likely you notice it. If yes, please a comment. If not, maybe just using either 3,5 or 3,6 would be fine?

src/odemis/driver/test/xt_client_test.py
Sabrina85: You could try something like this:
```
new_index = []
for idx,item in enumerate(list(current_index)):
   if item  < 8:
      item += 1
   else:
      item += -1
   new_index.append(item)
if new_index[0] == new_index[1]:
   new_index[0] += 2 if new_index[0] < 7 else -2
new_index = tuple(new_index)
```
This should cover all cases and makes sure you do have different values on x and y. If you want to allow also same values for x and y just remove the last` if `statement. Feel free to make the code more fancy.

src/odemis/driver/xt_client.py
```diff
@@ -800,31 +800,42 @@ def get_mpp_orientation(self):
             self.server._pyroClaimOwnership()
             return self.server.get_mpp_orientation()
 
+    def mpp_orientation_info(self):
```
Sabrina85, line 803: Oh nice. Didn't know that we also have this method available.

src/odemis/driver/xt_client.py
```diff
@@ -800,31 +800,42 @@ def get_mpp_orientation(self):
...
             self.server._pyroClaimOwnership()
-            return self.server.get_aperture_index()
+            return int(self.server.get_aperture_index())
 
     def set_aperture_index(self, aperture_idx):
         """
         Set the current aperture index.
 
-        :param aperture_idx  (float): Aperture index (range 0 - 14)
+        :param aperture_idx (float): Aperture index (typical range 0 - 14)
```
Sabrina85, line 828: Should the input here not be an int then as well as for the return value of the corresponding get method? It seems to be more logic to me that the return value of the get method and the input value of the set method are of the same type.

src/odemis/driver/xt_client.py
```diff
@@ -800,31 +800,42 @@ def get_mpp_orientation(self):
...
         """
         with self._proxy_access:
             self.server._pyroClaimOwnership()
-            self.server.set_aperture_index(aperture_idx)
+            self.server.set_aperture_index(float(aperture_idx))
 
     def aperture_index_info(self):
         """
         Get the current aperture index information, contains the range (0 - 14).
```
Sabrina85, line 836: Also add  here `typical range` for consistency.

Sabrina85 (changes requested):
Looks mostly fine now. I made a suggestion for the test case where you check the beamlet index.

src/odemis/driver/xt_client.py
```diff
@@ -790,6 +790,88 @@ def set_use_case(self, state):
             self.server._pyroClaimOwnership()
             return self.server.set_use_case(state)
 
+    def get_mpp_orientation(self):
+        """
+        Get the current multi probe orientation in degrees.
```
tepals, line 795: Minor comment: could you update this to say "multi probe pattern orientation". Same for the mpp_orientation_info docstring

src/odemis/driver/xt_client.py
Kleijwegt: Correct, fixed!

src/odemis/driver/xt_client.py
Sabrina85: The -18 to 18 degrees range was on FumoBeta where no k-mirror was available. On EA it should be -45 to 45 degrees to keep the beam index in the pattern well defined (due to the 4-fold symmetry).  What values did xt return for the range of the mmp? I see that in your tests you check that the angle is -90 < angle < 90 degrees. If it is indeed 90 degrees I need to check with TFS.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -821,10 +823,10 @@ def test_beamlet_index_info(self):
         beamlet_index_info = self.microscope.beamlet_index_info()
         self.assertIsInstance(beamlet_index_info, dict)
         self.assertIsInstance(beamlet_index_info["range"], dict)
-        self.assertIsInstance(beamlet_index_info["range"]["x-direction"], list)
-        self.assertIsInstance(beamlet_index_info["range"]["y-direction"], list)
-        self.assertEqual(len(beamlet_index_info["range"]["x-direction"]), 2)
-        self.assertEqual(len(beamlet_index_info["range"]["y-direction"]), 2)
+        self.assertIsInstance(beamlet_index_info["range"]["x"], list)
```
Sabrina85, line 826: Also update the doc string of get_beamlet_index() then accordingly. There it still says, keys are x-direction and y-direction.

Sabrina85 (changes requested):
I think the test case for the beam indexing does not work. Please check one more time.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -801,7 +801,9 @@ def test_set_beamlet_index(self):
 
         current_index = self.microscope.get_beamlet_index()
         # Make sure both the x and the y value are changed.
-        new_beamlet_index = (3, 3) if 3 not in current_index else (5, 5) if 5 not in current_index else (6, 6)
+        new_x_beamlet_index = 3 if current_index[0] != 3 else 5
+        new_y_beamlet_index = 3 if current_index[1] != 3 else 5
```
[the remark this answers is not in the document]
Kleijwegt: Let's do this then that seems a lot cleaner. I thought the most important part was that both x and y change value but changing them both to a different value is also possible.
        
```
        # Make sure both the x and the y value are changed.
        new_x_beamlet_index = 3 if current_index[0] != 3 else 6
        new_y_beamlet_index = 5 if current_index[1] != 5 else 7
        new_beamlet_index = (new_x_beamlet_index, new_y_beamlet_index)
```

pieleric (approved):

pieleric:
Looks fine to me now. Could you do a rebase with master to fix the conflicts in the test cases? Then I guess it should be ready to merge.

Kleijwegt:
> Looks fine to me now. Could you do a rebase with master to fix the conflicts in the test cases? Then I guess it should be ready to merge.

I will do this after the approval of @Sabrina85 then I will also squash all the commits.

Sabrina85 (approved):
Looks fine.

PR #1629, opened 2021-01-15
gui: add new icons for Cryo chamber tab

pieleric:
For representing the different positions of the stage in the chamber.

Designed by Maxim Houdijk, with the feedback from Caspar, Bassim and
myself.


Kleijwegt:
I have now idea how to review this? To watch these images should I just fetch this remote branch? Or are these already on dropbox or something?

pieleric:
@Kleijwegt , you can see the icons from within github, by clicking on the "display rich diff" icon (which looks like paper sheet) at the very right of the filename.

lazem (approved):
Looking good!

PR #1630, opened 2021-01-21
driver simcam: small fixes

pieleric:


src/odemis/driver/simcam.py
```diff
@@ -309,41 +316,47 @@ def set_image(self, new_img):
...
         """
         binning = self.binning.value
         res = self.resolution.value
-        pxs_pos = self.translation.value
-        center = self._get_center()
-        lt = (center[0] + pxs_pos[0] - (res[0] / 2) * binning[0],
-              center[1] + pxs_pos[1] - (res[1] / 2) * binning[1])
-        assert(lt[0] >= 0 and lt[1] >= 0)
+        trans = self.translation.value
+        # center = self._get_center()
```
lazem, line 327: Maybe this line is unnecessary if `_get_center()` is omitted?

src/odemis/driver/simcam.py
```diff
@@ -155,6 +155,7 @@ def clip_max_res(img_res):
         self._generator = None
         # Convenience event for the user to connect and fire
         self.softwareTrigger = model.Event()
+        self._last_acq_time = 0  # Time of latest acquisition
```
Kleijwegt, line 158: This var will be a float isn't it (time.time() is a float).
Should it be initialized as a float also? So 0.0
Its not that important since this is python but it is more consequent.

src/odemis/driver/test/simcam_test.py
```diff
@@ -302,29 +302,54 @@ def test_focus(self):
...
+        self.camera.updateMetadata({model.MD_POS: (0, 0)})
+        im0 = self.camera.data.get()
+        self.camera.translation.value = (-1, 1)
+        self.assertEqual(self.camera.resolution.value[::-1], im0.shape[:2])
+
+        # Move a little bit => should have a different image
+        self.camera.updateMetadata({model.MD_POS: (100e-6, 200e-6)})
+        im_move1 = self.camera.data.get()
+        # Note: images are always different, due to synthetic noise
+        test.assert_array_not_equal(im0, im_move1)
```
Kleijwegt, line 336: Using MeasureShift it would be possible to verify the amount of movement of the stage.
But probably you thought about that and that is not the goal of this unit test.

src/odemis/driver/simcam.py
pieleric: Technically Python still differentiates between floats and ints. However, since Python 2.7, it now behaves in pretty much exactly the same with a float or an int. I like to think of int as a subtype of float (in an object-oriented view), and so 0 is fine. 

So I'd rather keep `0`, it's just easier to read, and not showing any low-level details.

pieleric:
The code now seems to work both when testing it manually, and with the new fancy test cases.

src/odemis/driver/simcam.py
```diff
@@ -258,8 +265,8 @@ def _generate(self):
             gen_img = self._write_txt_image(gen_img, txt)
 
         # update fake output metadata
-        exp = timer.period
-        metadata[model.MD_ACQ_DATE] = time.time() - exp
+        self._last_acq_time = time.time() - exp
```
Kleijwegt, line 268: I thought the odemis-style was to define the ACQ_DATE at the start of the acquisition. Now it is done more or less at the end with some steps in between.
This could be prone to errors when the code is changed in later (e.g. acquisitions taken in the past because the sleep is not sufficient.
I don't think it is a big issue, unless it doesn't comply with the general odemis-style but that is up to you @pieleric.

src/odemis/driver/test/simcam_test.py
Kleijwegt: Looks good!

Kleijwegt (approved):

src/odemis/driver/simcam.py
pieleric: Note that this commit doesn't change the current behaviour. 
In any case, the convention is that the ACQ_DATE indicates the time at the beginning of the acquisition. That's the case here, as it reads the time at the end and subtracts the exposure time. The reason that it's done this way is that in the simulator the exposure time is simulated by using the `RepeatingTimer` which already takes care of sleeping so it'd be difficult to store the time before the sleep.

lazem (approved):
Very nice!

PR #1632, opened 2021-01-21
Prevent stopping the calibration loop while it is inactive.

Kleijwegt:
Please merge this one first in the row of PR's for the tech_wrapper.

### DIFF
--- a/src/odemis/driver/technolution.py
+++ b/src/odemis/driver/technolution.py
@@ -505,13 +505,18 @@ class AcquisitionServer(model.HwComponent):
         :param mode (bool): starting calibration mode (True) or stops the calibration mode (False)
         :return (bool): starting calibration mode (True) or stops the calibration mode (False)
         """
+        prev_state_calibration_mode = self.calibrationMode.value
+
         if mode:
             if not self._mppc.acq_queue.empty():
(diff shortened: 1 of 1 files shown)


src/odemis/driver/technolution.py
```diff
@@ -511,17 +511,24 @@ def _setCalibrationMode(self, mode):
                               "started yet.")
                 return False
 
-            self.asmApiPostCall("/scan/stop_calibration_loop", 204)  # Stop the calibration before (re) starting it
+            # If the calibration loop was already running stop it so it can be restarted with new parameters.
+            if self.calibrationMode.value:
```
pieleric, line 515: Because we are in the setter, this value actually represents the previous value of calibrationMode, right? This should be indicated in a comment. Maybe even clearer would be to put at the top of the function `prev_mode = self.calibrationMode.value`.

pieleric (changes requested):
The logic looks fine, but let's be a little more clear.

src/odemis/driver/technolution.py
Kleijwegt: Fixed, the fix is added in the single commit the PR contains so that is is still ready to be merged.

pieleric (approved):

Sabrina85 (approved):
Looks fine.

PR #1635, opened 2021-01-21
acq stream TemporalSpectrumSettingsStream: avoid errors when .raw is empty

pieleric:
During acquisition, the helper stream might update the "live" image in a
background thread while the acquisition thread has just reset the .raw
data, because a new e-beam position is used.

In such case, find_metadata() would fail. However, it's still updating
the projection based on the previous raw data, and the metadata is still
available. So use the metadata passed, instead of searching for it in
.raw again.


VictoriaDel (approved):
Thank you for fixing all the bugs of this code !

PR #1636, opened 2021-01-23
small improvements to the ENZEL simulator

pieleric:


src/odemis/driver/test/actuator_test.py
```diff
@@ -1282,39 +1281,107 @@ def tearDown(self):
...
+}
+
+KWARGS_FOCUS = {
+    "name": "Focus",
+    "role": "focus",
+    "rng": [0, 4.2e-3],  # min/max positions in m
+}
+
+INIT_MD_FOCUS = {
+    model.MD_POS_COR: {'z':-0.003},
```
lazem, line 1358: Just a space needed here  `{'z':- 0.003}`

lazem (changes requested):
Looks good overall, nice improvements. I had 2 issues running the `TestLinkedHeightActuator` tests:

1- In `test_cancel_reference` at line# 1672, replacing sleep time with `time.sleep(0.2)` fixed it.

2- In `test_cancel_move` changing sleep time at line# 1636 precisely to `time.sleep(1.05)` fixed it. Less than that it's not moved, and more it's not cancelled. 

I also suggest to add an additional test case for moving the stage first with 
`target_pos = {'z': self.target_value}  # To move stage first`
and repeat the` # 2. Cancel during movement` part of the test.
Let me know, I can add it myself to #1655 PR

src/odemis/driver/test/actuator_test.py
pieleric: Sorry, I won't do it, because Eclipse/PyDev doesn't allow me. It's part of its (broken?) auto formatting. I could work around it here, but I'm not ready to fight this fight in general.

pieleric:
@lazem , I've fixed the two test cases. It turned out that the issue was that the SmarAct simulators weren't so good. So I had to improve them.

I haven't added the test case. Mostly because it's getting really out of scope of this PR. I just want a nice simulator! Please create the test case as you have in mind, and propose it via a separate PR.

src/odemis/driver/test/actuator_test.py
lazem: I will fix it with the added test case.

lazem (approved):
Looks good!

PR #1639, opened 2021-01-23
driver simcam: remove blur_factor argument

pieleric:
This was used by a single simulator, to make the blur "stronger" when
moving the focus.
In addition, every DigitalCamera has a .depthOfField VA, which is automatically
computed based on the LENS metadata. This VA has a similar meaning. So
it makes sense that the simulated camera follows this VA to adjust the
blur factor.

=> Use depthOfField, and update the simulator with metadata which makes
the blur equivalent.

### DIFF
--- a/src/odemis/driver/simcam.py
+++ b/src/odemis/driver/simcam.py
@@ -42,7 +42,7 @@ class Camera(model.DigitalCamera):
     given at initialisation.
     '''
 
-    def __init__(self, name, role, image, dependencies=None, daemon=None, blur_factor=1e4, max_res=None, **kwargs):
+    def __init__(self, name, role, image, dependencies=None, daemon=None, max_res=None, **kwargs):
         """
         dependencies (dict string->Component): If "focus" is passed, and it's an
             actuator with a z axis, the image will be blurred based on the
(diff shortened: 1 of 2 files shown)


src/odemis/driver/simcam.py
```diff
@@ -266,8 +263,8 @@ def _generate(self):
         if self._focus:
             # apply the defocus
             pos = self._focus.position.value['z']
-            dist = abs(pos - self._metadata[model.MD_FAV_POS_ACTIVE]["z"]) * self._blur_factor
-            logging.debug("Focus dist = %g", dist)
+            dist = abs(pos - self._metadata[model.MD_FAV_POS_ACTIVE]["z"]) / (self.depthOfField.value * 10)
```
tepals, line 266: If I remember correctly the blur factor needed to be quite high to give enough blur. Did you verify that the values you are using now give the right blur?

pieleric: Yes, the 3 LENS_*  metadata were adjusted to get almost exactly the same value (to 2 digits).

tepals (approved):

PR #1640, opened 2021-01-23
driver simulated stage: fail to move outside of axis range

pieleric:
When doing a relative move, it would only indicate by a warning that it
goes outside of the range. That's different from other real drivers,
which typically raise a Exception in such case.

=> Simulate better by raising an exception too.

### DIFF
--- a/src/odemis/driver/simulated.py
+++ b/src/odemis/driver/simulated.py
@@ -114,15 +114,15 @@ class Stage(model.Actuator):
     def _doMoveRel(self, shift):
         maxtime = 0
         for axis, change in shift.items():
-            self._position[axis] += change
             rng = self.axes[axis].range
             if axis in self._inverted:
                 rng = (-rng[1], -rng[0])  # user -> internal range
-            if not rng[0] < self._position[axis] < rng[1]:
-                logging.warning("moving axis %s to %f, outside of range %r",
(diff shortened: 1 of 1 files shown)


src/odemis/driver/simulated.py
```diff
@@ -114,15 +114,15 @@ def _updatePosition(self):
...
-                logging.warning("moving axis %s to %f, outside of range %r",
-                                axis, self._position[axis], rng)
-            else:
-                logging.info("moving axis %s to %f", axis, self._position[axis])
+            if not rng[0] <= self._position[axis] + change <= rng[1]:
+                raise ValueError("moving axis %s to %f, outside of range %r" %
+                                 (axis, self._position[axis] + change, rng))
+
+            self._position[axis] += change
+            logging.info("moving axis %s to %f", axis, self._position[axis])
```
Kleijwegt, line 125: It could be nice to add the source of the error in the logging message (simulated stage).
But that is not essential to me.

Kleijwegt (approved):

src/odemis/driver/simulated.py
pieleric: The logging infrastructure already automatically include this info:
`2021-01-23 23:41:52,045 INFO    simulated:125:  moving axis z to 0.000100`

PR #1643, opened 2021-01-27
Enable Cryo tabs on Imaging mode

lazem:
Both alignment and localization tabs are used only when the stage is in imaging position, otherwise it'd better to disable them to avoid further issues from the enables controls inside them.


pieleric (changes requested):
I think the code to enable/disable the tabs has to be somewhere else. Either inside a stage position subscriber (per tab), or on the chamber tab.

PR #1646, opened 2021-01-28
Updating the scanOffset and scanGain default values

Kleijwegt:


pieleric (commented):
How "stable" are these values going to be? If that's the typical values for all the FAST-EMs, why not. Otherwise, I'd suggest to leave basic values (maybe 0.1 for the scanGain?), and use the microscope file to have default "better" values. You can do it by adding a section to the component that looks like:
```
properties: {
    scanOffset: [0.1172, 0.0989],
    :
}
```

Kleijwegt:
Approved number by Wilco Zuidema

pieleric:
My question is more: have these numbers any chance to stay approved for the rest of the FASTEMs, or they only make sense (with such precision) on the first/test system?

Kleijwegt:
I have set easier default values.
I will make a PR on bitbucket for the precise values in the yaml file

Sabrina85 (approved):

PR #1648, opened 2021-01-29
plugin zstack: in case of filter-wheel, acquire stream separately

pieleric:
Changing the filter-wheel for the emission can take several seconds.
That's typically way longer than a FM acquisition and moving the focus.
So if there are several streams, acquired serialy for each focus
position, the Z stack can get very slow.

=> Detect that there is a filter wheel, and in this case acquire each
stream seprately for the entire focus range.

### DIFF
--- a/plugins/zstack.py
+++ b/plugins/zstack.py
@@ -45,16 +45,13 @@ import wx
 
 class ZStackPlugin(Plugin):
     name = "Z Stack"
-    __version__ = "1.3"
-    __author__ = u"Anders Muskens"
+    __version__ = "1.4"
+    __author__ = u"Anders Muskens, Éric Piel"
     __license__ = "GPLv2"
 
     # Describe how the values should be displayed
     # See odemis.gui.conf.data for all the possibilities
     vaconf = OrderedDict((
-        ("numberOfAcquisitions", {
-            "control_type": odemis.gui.CONTROL_INT,  # no slider
-        }),
         ("filename", {
             "control_type": odemis.gui.CONTROL_SAVE_FILE,
             "wildcard": formats_to_wildcards(get_available_formats(os.O_WRONLY))[0],
@@ -65,6 +62,10 @@ class ZStackPlugin(Plugin):
         ("zstart", {
             "control_type": odemis.gui.CONTROL_FLT,
         }),
+        ("numberOfAcquisitions", {
+            "control_type": odemis.gui.CONTROL_INT,  # no slider
+        }),
+        ("expectedDuration", {}),
     ))
 
     def __init__(self, microscope, main_app):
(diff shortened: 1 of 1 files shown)


plugins/zstack.py
```diff
@@ -82,14 +83,22 @@ def __init__(self, microscope, main_app):
...
 
         self.filename = model.StringVA("a.h5")
         self.expectedDuration = model.VigilantAttribute(1, unit="s", readonly=True)
 
         self.zstep.subscribe(self._update_exp_dur)
-        self.numberofAcquisitions.subscribe(self._update_exp_dur)
+        self.numberOfAcquisitions.subscribe(self._update_exp_dur)
         
+        # Two acquisition order possible:
+        # * for each Z, all the streams (aka itertwined): Z exactly the same for each stream
```
tepals, line 95: typo: itertwined -> intertwined

plugins/zstack.py
```diff
@@ -137,14 +146,15 @@ def _update_exp_dur(self, _=None):
         """
         Called when VA that affects the expected duration is changed
         """
-        nsteps = self.numberofAcquisitions.value
+        nsteps = self.numberOfAcquisitions.value
         step_time = self._estimate_step_duration()
         ss = self._get_acq_streams()
+        nb_acqs = 1 if self._streams_intertwined else len(ss)
```
tepals, line 152: what is `nb_acqs`?

plugins/zstack.py
```diff
@@ -332,14 +342,21 @@ def initAcquisition(self):
...
+
+    def postStepAcquisition(self, i, images):
+        """
+        Called after the ith step of the acquisition
+        i (0 <= int): the step number
         images []: A list of images as DataArrays
         """
-        self.focus.moveRel({'z': self.zstep.value}).result()
-        
+        pass
```
tepals, line 358: Can you add a comment why you're  not doing anything here?

plugins/zstack.py
```diff
@@ -372,64 +389,87 @@ def acquire(self, dlg):
...
+            # Streams are slowest changed: for each stream, do all steps together
+            all_ss = [[s] for s in ss]
         
-        completed = False
-
         try:
-            step_time = self.initAcquisition()
-            logging.debug("Acquisition streams: %s", ss)
+            # list of list of list of DataArray: for each acquisition, for each stream, for each step, the data acquired
+            all_images = [[] for _ in all_ss]
```
tepals, line 411: Is this the same as `[[],] * len(all_ss) `?

tepals (changes requested):

plugins/zstack.py
pieleric: It's not. It's actually an easy mistake in Python. `l = [[]] * 3` returns a list containing 3 times the same reference to the same (sub)list. This means that if you do `l[1].append(0)`, you'll get `[[0], [0], [0]]`. 
That's not the case with the for loop.

plugins/zstack.py
pieleric: Yes, it's a bad name, as there is a VA called "numberOfAcquisitions" which is stored in nsteps. I just basically ran out of words to describe the 3 levels of loop.
I've refactored the code to just not use a name ;-)

plugins/zstack.py
pieleric: Yes, indeed it's a little odd. I've added a comment.
The main reason that this function is there is that the plugin has been built in a way that it should be generic to handle any axis move, by just inheriting the class. It hasn't been used yet, but the function is there in case another type of acquisition would need it....

tepals (approved):

PR #1651, opened 2021-02-01
Reduce the dwell time during check megafield to increase the testing speed

Kleijwegt:
Currently there is an error at the ASM side for long dwell times.
This was found because the check megafield (accidentally uses a long dwell time) however this test should not test a long scan but only check if the functionality of check_mega_field exists.
Therefore the dwell time for this test is set to the minimal allowed value.

### DIFF
--- a/src/odemis/driver/test/technolution_test.py
+++ b/src/odemis/driver/test/technolution_test.py
@@ -326,6 +326,8 @@ class TestAcquisitionServer(unittest.TestCase):
         Testing basics of checkMegaFieldExists functionality.
         """
         ASM = self.ASM_manager
+        # Set dwell time to minimum so scanning of an image is as fast.
+        self.EBeamScanner.dwellTime.value = self.EBeamScanner.dwellTime.range[0]
         # In the "setUp" method the megafield id is (re)set everytime when te test is stared to a string which contains
         # the current time i.e. time.strftime("testing_megafield_id-%Y-%m-%d-%H-%M-%S").
         mega_field_id = self.MPPC.filename.value


Sabrina85:
Will you add a test, that captures this error with the long dwell time so we can make sure it works after Technolution fixed it?

Kleijwegt:
> Will you add a test, that captures this error with the long dwell time so we can make sure it works after Technolution fixed it?

We can make a test in another PR. That test would be not fitting in this commit/PR.
However that test would be very time intensive (since the long dwell time). What do you @Sabrina85 think?

Sabrina85:
> > Will you add a test, that captures this error with the long dwell time so we can make sure it works after Technolution fixed it?
> 
> We can make a test in another PR. That test would be not fitting in this commit/PR.
> However that test would be very time intensive (since the long dwell time). What do you @Sabrina85 think?

Yes I agree it should not be part of this pr. As you found this issue I think there should be a test, even if that needs slightly longer (the maximum dwell time is still fast and maybe you can combine it with a smaller field image size). You can then use the test when Technolution provides an update and you want to make sure they fixed the problem. However, it is a discussion outside of the scope of this pr.

Kleijwegt:
> > > Will you add a test, that captures this error with the long dwell time so we can make sure it works after Technolution fixed it?
> > 
> > 
> > We can make a test in another PR. That test would be not fitting in this commit/PR.
> > However that test would be very time intensive (since the long dwell time). What do you @Sabrina85 think?
> 
> Yes I agree it should not be part of this pr. As you found this issue I think there should be a test, even if that needs slightly longer (the maximum dwell time is still fast and maybe you can combine it with a smaller field image size). You can then use the test when Technolution provides an update and you want to make sure they fixed the problem. However, it is a discussion outside of the scope of this pr.

@Sabrina85 Can you then approve this PR?

Sabrina85 (approved):

PR #1653, opened 2021-02-01
plugins ar_spectral: fix spectrum being inverted on some systems

pieleric:
We use the raw CCD, although normally it's the spectrometer, which wraps
the CCD that has an axis invertion if the wavelengths are the opposite
direction. Due to difference between hardware and possible extra mirror,
there is no convention of wavelength on the CCD. In practice this means
that there is 50% chance the wavelengths are inverted.

=> Automatically look at the spectrometer, and invert the wavelengths if
it does that.

Also remove some incorrect metadata which caused the data to be
sometimes not properly be displayed on the spacial image.

### DIFF
--- a/plugins/ar_spectral.py
+++ b/plugins/ar_spectral.py
@@ -837,6 +837,8 @@ class ARspectral(Plugin):
             try:
                 wl_inverted = (spec.transpose[0] == -1)
             except Exception as ex:
+                # Just in case spec has no .transpose or it's not a tuple
+                # (very unlikely as all Detectors have it)
                 logging.warning("%s: expect that the wavelengths are not inverted", ex)
 
         # the SEM survey stream (will be updated when showing the window)
@@ -975,14 +977,14 @@ class ARspectral(Plugin):
         raise LookupError: if nothing found.
         """
         for spec in self.main_app.main_data.spectrometers:
-            if not model.hasVA(spec, "dependencies"):
-                continue
             # Check by name as the components are actually Pyro proxies, which
             # might not be equal even if they point to the same component.
-            if detector.name in (d.name for d in spec.dependencies.value):
+            if (model.hasVA(spec, "dependencies") and
+                detector.name in (d.name for d in spec.dependencies.value)
+               ):
                 return spec
 
-        raise LookupError("Failed to find spectrometer corresponding to CCD %s" % (detector.name,))
+        raise LookupError("No spectrometer corresponding to %s found" % (detector.name,))
 
     def start(self):
         # get region and dwelltime for drift correction
--- a/plugins/ar_spectral.py
+++ b/plugins/ar_spectral.py
@@ -60,6 +60,11 @@ import time
 
 import odemis.util.driver as udriver
 
+# This is the type of metadata that we could get on the CCD, but wouldn't be correct
+# for the type of EK acquisition. Note that we do keep the AR metadata, which
+# is correct in Y, and can be used to compute the angle.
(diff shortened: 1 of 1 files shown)


plugins/ar_spectral.py
```diff
@@ -803,15 +819,35 @@ def __init__(self, microscope, main_app):
...
+        # in order to make sure the wavelength is is the correct direction (ie,
+        # lowest pixel = lowest wavelength). So we need to do the same on the
+        # raw image. However, there is no "official" way to connect the
+        # spectrometer(s) to their raw CCD. So we rely on the fact that
+        # typically this is a wrapper, so we can check using the .dependencies.
+        wl_inverted = False
+        try:
+            spec = self._find_spectrometer(self.ccd)
+        except LookupError as ex:
+            logging.warning("%s, expect that the wavelengths are not inverted", ex)
```
VictoriaDel, line 834: This LookupError should be "Failed to find spectrometer... "

plugins/ar_spectral.py
```diff
@@ -932,6 +968,22 @@ def _get_sem_survey(self):
         logging.warning("No SEM survey stream found")
         return None
 
+    def _find_spectrometer(self, detector):
+        """
+        Find a spectrometer which wraps the given detector
+        return (Detector): the spectrometer
+        raise LookupError: if nothing found.
+        """
+        for spec in self.main_app.main_data.spectrometers:
```
VictoriaDel, line 977: You could replace main_app.main_data with main_data, as you did earlier for getting self._acqui_tab.

plugins/ar_spectral.py
```diff
@@ -932,6 +968,22 @@ def _get_sem_survey(self):
...
+        Find a spectrometer which wraps the given detector
+        return (Detector): the spectrometer
+        raise LookupError: if nothing found.
+        """
+        for spec in self.main_app.main_data.spectrometers:
+            if not model.hasVA(spec, "dependencies"):
+                continue
+            # Check by name as the components are actually Pyro proxies, which
+            # might not be equal even if they point to the same component.
+            if detector.name in (d.name for d in spec.dependencies.value):
```
VictoriaDel, line 982: Only one if condition with "and" would be clearer to me :) But you know better.

plugins/ar_spectral.py
```diff
@@ -803,15 +819,35 @@ def __init__(self, microscope, main_app):
...
+        try:
+            spec = self._find_spectrometer(self.ccd)
+        except LookupError as ex:
+            logging.warning("%s, expect that the wavelengths are not inverted", ex)
+        else:
+            # Found spec => check transpose in X (1 or -1), and invert if it's inverted (-1)
+            try:
+                wl_inverted = (spec.transpose[0] == -1)
+            except Exception as ex:
+                logging.warning("%s: expect that the wavelengths are not inverted", ex)
```
VictoriaDel, line 840: We need to know if wl is inverted or not. Why would we get an exception? (just for warning the user?)

VictoriaDel (changes requested):

plugins/ar_spectral.py
pieleric: Yes, indeed. So the warning will look like "Failed to find spectrometer... , expect that the wavelengths...". It's fine, right?

pieleric (commented):
I've fixed some of the comments, and answered back for the ones I couldn't fix.

VictoriaDel (approved):
I think only the warning log of the LookupError remains to be updated, and it's fine :)

plugins/ar_spectral.py
pieleric: Yes, that's fine I think.

PR #1654, opened 2021-02-02
ASM wrapper: update MD every time a new field image is acquired

Sabrina85:
If we update the MD only in the start of a megafield, each dataarray (field) returned would have the same stage position. Update the MD every time a new field is acquired, so that the dataarray returned has the correct stage position.

TODO: A proper fix would be to have a settingsMD which includes MD such as the dwell time. And in the next call we would call mergeMD and add the settingsMD via updateMD call. Thus, we make sure that the generic settings such as the dwell time is for all fields the one that was set in the beginning of the megafield acquisition, but the stage position is updated correctly per field.

### DIFF
--- a/src/odemis/driver/technolution.py
+++ b/src/odemis/driver/technolution.py
@@ -1135,7 +1135,6 @@ class MPPC(model.Detector):
 
                     acquisition_in_progress = True
                     megafield_metadata = args[0]
-                    self._metadata = self._mergeMetadata()
                     self.parent.asmApiPostCall("/scan/start_mega_field", 204, megafield_metadata.to_dict())
 
                 elif command == "next":
@@ -1143,6 +1142,7 @@ class MPPC(model.Detector):
                         logging.warning("Start ASM acquisition before request to acquire field images.")
                         continue
 
+                    self._metadata = self._mergeMetadata()
                     field_data = args[0]  # Field metadata for the specific position of the field to scan
                     dataContent = args[1]  # Specifies the type of image to return (empty, thumbnail or full)
                     notifier_func = args[2]  # Return function (usually, dataflow.notify or acquire_single_field queue)


src/odemis/driver/technolution.py
```diff
@@ -1135,14 +1135,14 @@ def _acquire(self):
...
                     megafield_metadata = args[0]
-                    self._metadata = self._mergeMetadata()
                     self.parent.asmApiPostCall("/scan/start_mega_field", 204, megafield_metadata.to_dict())
 
                 elif command == "next":
                     if not acquisition_in_progress:
                         logging.warning("Start ASM acquisition before request to acquire field images.")
                         continue
 
+                    self._metadata = self._mergeMetadata()
```
Kleijwegt, line 1145: Let's check this one together via Zoom i think that the stage shouldn't/isn't allowed to move during the next call.
I think in one way or another the placement of this line of code was discussed before, but maybe I'm wrong.

Kleijwegt (changes requested):

src/odemis/driver/technolution.py
pieleric: I don't know if you already had a discussion with Sabrina. Just in case, you hadn't, the point here is not to handle the if the stage moves during a call to next(), but to handle the fact that typically every `next()` will be called at a different stage position. So the stage is moved just before calling `next()`.

Kleijwegt: But the stage position send to the ASM is only updated at the start of a megafield with the call "/scan/start_mega_field" isn't it?
The call "/scan/scan_field" does include a fieldnumber position noted as position_x/y and not stage position. That position is only included in the start megafield call.

pieleric: That's correct. But that's not what the metadata is about ;-)
In case the data is sent back to the client (thumbnail or full data) via the DataFlow, the DataArray should have the right metadata. Without this change, all data returned directly to the client would look like it's acquired at the first field position.
In practice, this change is needed to properly display the live stream (with the thumbnails) during an acquisition.

Kleijwegt: Can we maybe discuss this via a video call @pieleric  I believe you are right but I don't completly understand how/what is done with the stage_position in the ASM and how the stage position on the metadata is updated correctly. (The stage_position send to the ASM is loaded from the metadata which is now not updated before sending it to the ASM)

Sabrina85: The stage position on the megafield metadata is only send once to the ASM (with start_mega_field: stage_position_x, stage_position_y) and defines the top left field image (its center). This metadata is only saved once per megafield (in the megafield MD) and not for each field. Each field only gets a number (position_x and position_y) which can be used to calculate the relative position. This metadata is attached per field when you do the next call (or scan_field). 
The problem discussed here is about the array that is returned in Odemis to a listener. That one should know the stage position per field. That's why this line was moved in the code.

Kleijwegt (approved):

PR #1655, opened 2021-02-02
Fix out of range issue of Focus Relative Move

lazem:


src/odemis/driver/actuator.py
```diff
@@ -2697,11 +2699,15 @@ def adjustFocus(self, future, vector_value, rel):
         :param rel: whether it's a relative movement or absolute
         """
         # Drop focus adjustment if lens is parked
-        if self._isParked():
-            logging.warning("Focus adjust movement is dropped as lens is parked.")
+        # if self._isParked():
+        #     logging.warning("Focus adjust movement is dropped as lens is parked.")
+        #     return
```
pieleric, line 2704: Well, let's drop this code completely. => remove the lines.

src/odemis/util/comp.py
```diff
@@ -108,3 +110,19 @@ def get_fov_rect(comp, fov):
             center[1] - fov[1] / 2,  # top
             center[0] + fov[0] / 2,  # right
             center[1] + fov[1] / 2)  # bottom
+
+def execute_function_within_subfuture(future, fn, args):
```
pieleric, line 114: Let's not put it in util.comp, yet. It relies on a lot of non-standard behaviour of the futures. Especially, the moving_lock and _must_stop are special attributes placed by the actuator code. 
So I'd suggest to place this function somewhere inside the actuator.LinkedHeightActuator.

That said, in the long term, I agree that it'd make sense to share this code as we are starting to have many places in Odemis which do the same. Maybe it would be possible to extend the `odemis.model._future.CancellableFuture` class to provide such mechanism. In such case, it'd handle the cancellation, and also the lock, and cancelling the sub futures.

src/odemis/driver/actuator.py
```diff
@@ -2697,11 +2699,15 @@ def adjustFocus(self, future, vector_value, rel):
...
+        if not self._isInRange():
+            logging.warning("Focus adjust movement is dropped as lens is not in active range.")
             return
         if rel:
-            self._doMoveRel(future, {'z': vector_value})
+            # Move the underlying lens with the relative shift value
+            execute_function_within_subfuture(future, self._lensz.moveRel, {'z': vector_value})
         else:
             self._doMoveAbs(future, vector_value, adjust=True)
```
pieleric, line 2712: Please, do the same with absolute moves. It should be "symmetric". This way, you can drop the "adjust" argument of doMoveAbs() and the function will be much simpler.

pieleric (changes requested):
Please, add a test case which fails with the current code!
Otherwise, the changes look fine. Just small comments.

PR #1656, opened 2021-02-03
xt client driver: small fix

Sabrina85:


src/odemis/driver/test/xt_client_test.py
```diff
@@ -820,7 +820,7 @@ def test_mpp_orientation_info(self):
         mpp_orientation_info = self.microscope.mpp_orientation_info()
         self.assertIsInstance(mpp_orientation_info, dict)
         self.assertTrue("unit" in mpp_orientation_info)
-        self.assertIsInstance(mpp_orientation_info["range"], tuple)
+        self.assertIsInstance(mpp_orientation_info["range"], list)
```
Kleijwegt, line 823: What changed that this value returns a list instead of tuple? Looking at the code all other info calls return a tuple .
These two tests failed when Sabrina tested it.
If the others all return a tuple, then it is not correctly handled in the respective adapter methods?

Kornee: We need to check/fix it on the adapter side probably

Kleijwegt (changes requested):

src/odemis/driver/test/xt_client_test.py
Sabrina85: I checked for the client and adapter. It looks like the adapter call indeed returns the tuple, but the client call always returns a list. This happened also for other info calls, which I tested.
@tepals has this something to do with how Pyro handles tuples?

Sabrina85: I tested this again. As we do not modify the dictionaries returned by any info call, it will be always a list when called on the client. As we discussed yesterday, we will make sure that all calls returning a tuple on the adapter should also return a tuple on the client. If a dictionary is returned, which happens to contain a tuple, will still be of type list. So we can either modify the test case here, or do not check the type here at all. If we would not check the type here at all, we would keep the tests the same on the adapter and client. @Kleijwegt What do you think?

pieleric: I've looked a little at the xtadapter code, and most probably the fact it's a list instead of a tuple is due to the code getting a list. (Or it's due to Pyro5 changing from tuple -> list). Whichever way, my feeling is that the type is not important. It will not be on the user visible part of the driver, and we only use the info to copy it once at init on a VA.
So I'd suggest to just drop this test line, the check on the length is sufficient.

PR #1657, opened 2021-02-03
Raise value error when there is no MD_WL

VictoriaDel:


src/odemis/util/spectrum.py
```diff
@@ -72,7 +72,7 @@ def get_wavelength_per_pixel(da):
             raise ValueError("Length of wavelength list does not match length of wavelength data.")
         return wl
     except KeyError:
-        logging.error("No MD_WL_LIST metadata available")
+        raise ValueError("No MD_WL_LIST metadata available")
```
pieleric, line 75: Please, change to KeyError, as it used to behave, and as it is described in the docstring.

pieleric (changes requested):
Good catch.
However, in addition to my comment about the exception type, please explain a bit more the reason why this commit fixes things. I had to look at the git history and the current code to understand. That shouldn't be needed.

So in the commit message, you should indicate:
* That this is a fix for another commit (with the commit ID and title)
* Why it was wrong. (eg, "The code used to always return a KeyError, but in the removal of MD_POLYNOMIAL, we also removed the exception and replaced it by a simple log.")

PR #1658, opened 2021-02-03
piezomotor: referencing, bugfixes, scan function

philipwinklerdelmic:
**This is not ready to be merged**
I made a few changes to the stage driver, including
* a new referencing procedure (works differently with encoder, see comments in code)
* the use of a .tsv file to pass the parameters (it turns out that there are more of them then I thought).
* explicit passing of both the motorstep and encoder resolution (the indirect conversion with spc caused problems)
* a timeout factor for the z axis (z axis is attached through a different component, see docstring)
* a scan function and more explicit messages in pmconfig (sometimes it's hard to figure out which axis numbers there are)
* removed referencing in pmconfig, it's too dangerous (especially if there are problems with the limit switch)

I have not tested all the changes on the hardware yet. I expect that I have to make some more adaptations (especially with the scan function).
@pieleric , it would be nice if you could have a first look already and check if you see some major problems. We can wait with the full code review until I tested it next week.

### DIFF
--- a/src/odemis/driver/piezomotor.py
+++ b/src/odemis/driver/piezomotor.py
@@ -23,6 +23,7 @@ from __future__ import division
 
 import glob
 import logging
+from concurrent.futures import CancelledError
 from odemis import model
 from odemis.model import Actuator, CancellableThreadPoolExecutor, CancellableFuture, isasync
 from odemis.util import to_str_escape
@@ -30,13 +31,14 @@ import os
 import threading
 import time
 import serial
+import re
 from threading import Thread
 import copy
 from odemis.model import HwError
 from odemis.util.driver import getSerialDriver
 
 BAUDRATE = 115200
-DEFAULT_AXIS_SPEED = 0.01  # m / s
+DEFAULT_AXIS_SPEED = 0.001  # m / s
 
 EOL = b'\r'  # 0x0D, carriage return
 sEOL = '\r'  # string version of EOL
@@ -47,17 +49,17 @@ WAVEFORM_DELTA = 2  # preferred, higher accuracy
 WAVEFORM_PARK = 4  # power off
 
 # Length of the rod available for moves
-STROKE_RANGE = 23.1e-3  # m
+STROKE_RANGE = [-25e-3, 25e-3]  # m, referencing point is in the middle at position 0
 
 # There are two different units for a move: motor (waveform) steps and encoder counts.
 # The encoder counts are fixed and given by the encoder resolution. The motor steps
 # vary depending on the load of the motor. Therefore, a calibration is required before
 # the first use of the driver to determine the conversion factor between encoder counts
-# and motor steps. This conversion factor can be stored in the controller flash memory.
+# and motor steps (SPC value). This conversion factor is passed in the .tsv file.
 # A motor step is equivalent to 8192 microsteps.
-DEFAULT_COUNTS_PER_METER = 1 / 1.25e-6  # TODO: update default value (current value is for starter kit)
-DEFAULT_SPC = 1.25e-6 / 4.5e-6  # steps per count (encoder steplength / motor steplength)
 USTEPS_PER_STEP = 8192
+DEFAULT_MOTORSTEP_RESOLUTION = 4.5e-6  # LL06 motor
+DEFAULT_ENCODER_RESOLUTION = 1.22e-9  # Mercury II encoder
 
 
 class PMDError(Exception):
@@ -78,9 +80,10 @@ class PMD401Bus(Actuator):
     controller is directly connected to the computer.
     The specification for the hardware interface can be found in the document
      "PiezoMotor_PMD401_Technical_Manual.pdf".
+    This driver works equally well with the PMD301 controller.
     """
 
-    def __init__(self, name, role, port, axes, inverted=None, **kwargs):
+    def __init__(self, name, role, port, axes, inverted=None, param_file=None, **kwargs):
         """
         :param axes (dict: {"x", "y", "z"} --> dict): axis name --> axis parameters
             Each axis is specified by a set of parameters.
@@ -94,26 +97,20 @@ class PMD401Bus(Actuator):
             Axis parameters:
                 axis_number (0 <= int <= 127): typically 1-3 for x-z, required
                 closed_loop (bool): True for closed loop (with encoder), default to True
-                encoder_resolution (float): encoder resolution in m/step
-                spc (float): motor steps per encoder count, default to value in non-volatile memory
-                limit_type (0 <= int <= 2): type of limit switch, 0: no limit, 1: active high, 2: active low, default 0
-                range (tuple of float): in m, default to (0, STROKE_RANGE)
+                encoder_resolution (float): number of encoder counts per meter, default to 1.22e-9
+                motorstep_resolution (float): number of motor steps per m, default to 5e-6
+                range (tuple of float): in m, default to STROKE_RANGE
                 speed (float): speed in m/s
                 unit (str), default to m
+        :param param_file (str or None): (absolute or relative) path to a tmcm.tsv file which will be used to initialise
+            the axis parameters.
         """
         self._axis_map = {}  # axis name -> axis number used by controller
         self._closed_loop = {}  # axis name (str) -> bool (True if closed loop)
         self._speed_steps = {}  # axis name (str) -> int, speed in steps per meter
-        self._portpattern = port
-
-        # Conversion factors
-        # Count refers to encoder counts, step refers to motor steps. The encoder counts are fixed and given
-        # as a parameter to the axis, the motor counts are determined during configuration and are usually
-        # stored in flash memory.
-        # ._steps_per_meter is redundant, but convenient
-        self._steps_per_count = {}  # axis name (str) -> float
-        self._steps_per_meter = {}  # axis name (str) -> float
         self._counts_per_meter = {}  # axis name (str) -> float
+        self._steps_per_meter = {}  # axis name (str) -> float
+        self._portpattern = port
 
         # Parse axis parameters and create axis
         axes_def = {}  # axis name -> Axis object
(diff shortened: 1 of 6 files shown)


src/odemis/driver/piezomotor.py
```diff
@@ -94,26 +96,26 @@ def __init__(self, name, role, port, axes, inverted=None, **kwargs):
...
+                range (tuple of float): in m, default to STROKE_RANGE
                 speed (float): speed in m/s
+                timeout_factor (int): the z axis typically behaves differently since it's not attached directly
+                    to the controller. Instead, there is a connection to a different component, and the movement
+                    of this different component causes the movement of the z axis. A larger movement in the component
+                    only causes a smaller movement in the z axis. For the most part, this is dealt with by adjusting
+                    the encoder resolution. However, after adjusting the encoder resolution, the timeouts are
+                    no longer useful (because we're assuming to move only a small amount on the axis, but actually,
+                    the movement of the second component is much larger and therefore also takes longer).
+                    This parameter allows to multiply the timeouts by a constant factor (1 by default).
```
pieleric, line 110: I don't get that. The default timeout is computed based on the range and the speed. Both of these should be the "user-visible". The time to move around this should also be based on this. It doesn't matter if the motor is directly connected to the stage or connected via a leaver. We care about the final stage movement. 
I don't exactly understand how everything is connected, but it seems that just multiplying encoder_resolution and motorstep_resolution by the factor would suffice. 

In any case, as a separate note, it's best to avoid describing the direct issue you're trying to solve in the docstring, but instead just describe the effect of the parameter. For instance here, the fact that timeout_factor is going to be used for the z axis in the current system has relationship with the actual effect. This parameter can be used on any axis. You should explain the issue and the use of the parameters in the microscope file.

install/linux/usr/share/odemis/hwtest/pmd401.piezomotor.tsv
```diff
@@ -0,0 +1,19 @@
+# Parameters for PMD301/PMD401
+# Bank/Axis	Address	Value	Description
```
pieleric, line 2: I don't think there is such "bank" notion in the controller, so let's remove any reference to banks from this file, and the parser function.

src/odemis/driver/piezomotor.py
```diff
@@ -94,26 +96,26 @@ def __init__(self, name, role, port, axes, inverted=None, **kwargs):
...
                 unit (str), default to m
         """
```
pieleric, line 112: Please add a docstring about `param_file`.

src/odemis/driver/test/piezomotor_test.py
```diff
@@ -44,12 +44,14 @@
 
 KWARGS_OPEN = dict(name="test", role="test", port=PORT,
               axes={'x': {'axis_number': 1, 'speed': 0.001, 'closed_loop': False},
-                    'y': {'axis_number': 2, 'speed': 0.001, 'closed_loop': False}}
+                    'y': {'axis_number': 2, 'speed': 0.001, 'closed_loop': False}},
+                param_file="../../../../install/linux/usr/share/odemis/hwtest/pmd401.piezomotor.tsv"
```
pieleric, line 48: A bit of nitpicking, but ".piezomotor.tsv" is a little longish. How about ".pmd.tsv", it's shorter, and possibly more precise.

pieleric (changes requested):

pieleric:
@philipwinklerdelmic , you've marked the comments resolved, but not pushed any changes. Are you still working on the changes, or could you push the new version for review?

src/odemis/driver/piezomotor.py
```diff
@@ -210,33 +204,24 @@ def __init__(self, name, role, port, axes, inverted=None, **kwargs):
...
+            try:
+                axis_params = self.parse_tsv_config(f)
+            except Exception as ex:
+                raise ValueError("Failed to parse file %s: %s" % (param_file, ex))
+            f.close()
+            logging.debug("Extracted param file config: %s", axis_params)
+            self.apply_params(axis_params)
+            # TODO: flash values? Should not be necessary, but might be convenient (tsv file not required the next time)
+            # for axnum in self._axis_map.values():
+            #     self.writeParamsToFlash(axnum)
```
pieleric, line 224: No need. Next time, the param file will still be there.

src/odemis/driver/piezomotor.py
```diff
@@ -283,38 +268,94 @@ def reference(self, axes):
...
+            if time.time() > end_time:
+                self.stopAxis(axis)  # exit index mode
+                raise IOError("Timeout while waiting for end of motion on axis %s" % axis)
+
+            # Check if limit is reached
+            try:
+                self._check_hw_error()
+            except PMDError as ex:
+                if ex.errno == 6:  # external limit reached
+                    self.stopAxis(axis)  # that seems to be necessary after reaching the limit
```
pieleric, line 341: Add a `logging.debug("Axis %d limit reached during referencing" ....)`

src/odemis/driver/piezomotor.py
```diff
@@ -283,38 +268,94 @@ def reference(self, axes):
...
-                    self.stop()  # exit index mode
-                    raise ValueError("Referencing axis %s failed." % self._axis_map[axname])
-                time.sleep(0.1)
+
+            # First, search for the index in negative direction.
+            idx_found = self._search_index(f, axname, direction=-1)
+
+            # If it wasn't found, try again in positive direction.
+            if not idx_found:
+                idx_found = self._search_index(f, axname, direction=1)
```
pieleric, line 294: Add a `logging.debug("Referencing axis %s in the positive direction"...)`

src/odemis/driver/piezomotor.py
```diff
@@ -326,17 +367,24 @@ def _doMoveAbs(self, f, pos):
...
                 self.runMotorJog(self._axis_map[axname], steps, usteps, self._speed_steps[axname])
                 targets[axname] = None
-        self._waitEndMotion(f, targets)
-        # Leave target mode in case of closed-loop move
-        for ax in pos:
-            self.stopAxis(self._axis_map[ax])
-        self._updatePosition()
+        try:
+            self._waitEndMotion(f, targets)
+        except:
```
pieleric, line 372: Please, prefer `except Exception`. 
However, the whole block is not needed if you just do a raise. Maybe also add a logging.warning?

src/odemis/driver/piezomotor.py
```diff
@@ -347,11 +395,15 @@ def _doMoveRel(self, f, shift):
...
                 targets[axname] = None
-        self._waitEndMotion(f, targets)
-        # Leave target mode in case of closed-loop move
-        for ax in shift:
-            self.stopAxis(self._axis_map[ax])
-        self._updatePosition()
+        try:
+            self._waitEndMotion(f, targets)
+        except:
+            raise
```
pieleric, line 401: Same as for moveAbs()

src/odemis/driver/piezomotor.py
```diff
@@ -365,9 +417,9 @@ def _waitEndMotion(self, f, targets):
             if target is not None:
                 move_length = max(abs(self.position.value[ax] - target), move_length)
         if move_length == 0:  # open loop
-            move_length = STROKE_RANGE
+            move_length = STROKE_RANGE[1] - STROKE_RANGE[0]
         dur = move_length / self._speed
-        max_dur = max(dur * 2, 0.1)  # wait at least 0.1 s
+        max_dur = max(dur * 2, 0.5)  # wait at least 0.5 s
```
pieleric, line 422: Usually, I just do `dur * 2 + 1`, so it's really pessimistic, and don't risk timing out by mistake.

src/odemis/driver/piezomotor.py
```diff
@@ -650,13 +712,26 @@ def _sendCommand(self, cmd):
         cmd += EOL
         with self._ser_access:
             logging.debug("Sending command %s", to_str_escape(cmd))
-            self._serial.write(cmd)
+            try:
+                self._serial.write(cmd)
+            # TODO: what kind of exception is raised? Needs to be more specific.
```
pieleric, line 717: Indeed... please pick an exception.

src/odemis/driver/piezomotor.py
```diff
@@ -674,11 +749,11 @@ def _sendCommand(self, cmd):
 
             # Check response (command should be echoed back)
             if not resp.startswith(cmd[:-len(EOL)-1]):
-                raise IOError("Response starts with %s != %s" % (resp[:len(cmd)], cmd))
+                raise IOError("Response starts with %s != %s" % (resp[:len(cmd)].decode("latin1"), cmd))
```
pieleric, line 752: Instead of using `decode("latin1")`, use `to_str_escape()`

src/odemis/driver/piezomotor.py
```diff
@@ -787,6 +861,80 @@ def _openSerialPort(port):
 
         return ser
 
+    @classmethod
+    def scan(cls):
+        """
+        returns (list of 2-tuple): name, kwargs
+        Note: it's obviously not advised to call this function if a device is already under use
+        """
+        # TODO: use serial.tools.list_ports.comports() (but only availabe in pySerial 2.6)
```
pieleric, line 870: You could definitely use this function now. We support Ubuntu 18.04+, which come with pySerial 3.4.

pieleric (changes requested):

src/odemis/driver/piezomotor.py
philipwinklerdelmic: I want the finally to be executed no matter what. So what it should be doing in case of an exception is just raise it, and then afterwards call the stop function on all axes to go back to a normal state. Is there a better way of doing it?

src/odemis/driver/piezomotor.py
philipwinklerdelmic: Problem is, I don't know which exception is raised. IOError is not the right one, I think it's something very specific, we need to test this with someone in the lab.

src/odemis/driver/piezomotor.py
pieleric: If all you care about is the finally, it's possible to just write:
```python
try:
    blabla
finally:
    do it
```
(You don't need an except)

src/odemis/driver/piezomotor.py
pieleric: The tmcm driver uses `IOError`, and it works fine. So I'd go for that too. It might be possible to be even more specific, but I don't think is really necessary.

philipwinklerdelmic: I tried IOError, it didn't work, but I didn't have time to check what it was in the end.

src/odemis/driver/piezomotor.py
philipwinklerdelmic: What's the advantage of using this function over glob.glob? I just copied this comment from somewhere else.

philipwinklerdelmic:
pushed changes

src/odemis/driver/piezomotor.py
```diff
@@ -787,6 +856,79 @@ def _openSerialPort(port):
...
+    @classmethod
+    def scan(cls):
+        """
+        returns (list of 2-tuple): name, kwargs
+        Note: it's obviously not advised to call this function if a device is already under use
+        """
+        if os.name == "nt":
+            raise NotImplementedError("OS not yet supported")
+        else:
+            ports = list(serial.tools.list_ports.grep('/dev/ttyUSB*'))
```
pieleric, line 868: What I have in mind is that you can replace all this by:
```python
ports = [p.device for p in serial.tools.list_ports.comports()]
```
Is there a problem with using `comports()`?

src/odemis/driver/piezomotor.py
```diff
@@ -674,11 +744,11 @@ def _sendCommand(self, cmd):
...
             # Check response (command should be echoed back)
             if not resp.startswith(cmd[:-len(EOL)-1]):
-                raise IOError("Response starts with %s != %s" % (resp[:len(cmd)], cmd))
+                raise IOError("Response starts with %s != %s" % (to_str_escape(resp[:len(cmd)]), cmd))
             if b"_??_" in resp:
-                raise ValueError("Received response %s, command %s not understood." % (resp, cmd))
+                raise ValueError("Received response %s, command %s not understood." % (resp.decode("latin1"), cmd))
             if b"!" in resp:
-                raise PMDError(0, resp)
+                raise PMDError(0, resp.decode("latin1"))
```
pieleric, line 751: Use `to_str_escape()` here too. Any user output needs to be escaped.

pieleric (changes requested):

PR #1659, opened 2021-02-03
focustracker

philipwinklerdelmic:
**Work in progress, not ready to merge**
* Still needs to be tested on the hardware.
* Decisions on the interface need to be made (functions or VAs or both), see specification
* .eds needs to be updated again after firmware changes
* Testcases need to be written

@pieleric Sorry for spamming the pull requests with half-finished code. I thought this might be useful when you are reviewing the focustracker specification if something is not clear. I don't expect a review of the code yet, I'll first test on the hardware next week.


pieleric:
There is now a different pull request, tested on the hardware.

PR #1660, opened 2021-02-03
plugins: new plugin sparc_engage_warn to show message box before moving mirror

pieleric:
For systems where it can be dangerous (eg, other detectors possibly in
the path), we can add this plugin.
On every press of the "Mirror" button in the chamber a confirmation is
asked.

### DIFF
--- a/plugins/sparc_engage_warn.py
+++ b/plugins/sparc_engage_warn.py
@@ -1,0 +1,89 @@
+# -*- coding: utf-8 -*-
+# Shows a warning message before engaging/parking the SPARC mirror
+'''
+Created on 3 Feb 2021
+
+@author: Éric Piel
+Copyright © 2021 Éric Piel, Delmic
+
+This file is part of Odemis.
+
+Odemis is free software: you can redistribute it and/or modify it under the terms of the GNU
+General Public License version 2 as published by the Free Software Foundation.
+
+Odemis is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even
+the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General
+Public License for more details.
+
+You should have received a copy of the GNU General Public License along with Odemis. If not,
+see http://www.gnu.org/licenses/.
+'''
+
+from __future__ import division
+
+import logging
+from odemis.gui.cont.tabs import MIRROR_PARKED
+from odemis.gui.plugin import Plugin
+import wx
+
+ENGAGE_MSG = """Are you sure you want to engage the mirror?
+* Ensure the stage is at a safe working distance.
+* Ensure all detectors are retracted.
+"""
+
+PARK_MSG = """Are you sure you want to park the mirror?
+* Ensure all detectors are retracted.
+"""
+
+
+class SparcEngageWarnPlugin(Plugin):
+    name = "SPARC mirror engage warning"
+    __version__ = "1.0"
+    __author__ = u"Éric Piel"
+    __license__ = "GPLv2"
+
+    def __init__(self, microscope, main_app):
+        super(SparcEngageWarnPlugin, self).__init__(microscope, main_app)
+
+        # It only makes sense if the SPARC chamber tab is present
+        try:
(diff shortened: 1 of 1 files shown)


plugins/sparc_engage_warn.py
```diff
@@ -0,0 +1,86 @@
...
+
+from __future__ import division
+
+import logging
+from odemis.gui.cont.tabs import MIRROR_PARKED
+from odemis.gui.plugin import Plugin
+import wx
+
+ENGAGE_MSG = """Are you sure you want to engage the mirror?
+* Ensure stage is at a working distance of at least 18 mm.
```
Kleijwegt, line 30: Is there now way of checking what the current working distance is and including that in the message?

pieleric: Nope. We often don't have access to the stage, and even more rarely have the information on the working distance.

plugins/sparc_engage_warn.py
```diff
@@ -0,0 +1,85 @@
...
+    __author__ = u"Éric Piel"
+    __license__ = "GPLv2"
+
+    def __init__(self, microscope, main_app):
+        super(SparcEngageWarnPlugin, self).__init__(microscope, main_app)
+        # Allow to run it on pretty much anything (excepted the viewer)
+        if not microscope or microscope.role != "sparc2":
+            return
+
+        # For reading the ROA and anchor ROI
```
pieleric, line 51: Comment needs to be fixed.

plugins/sparc_engage_warn.py
```diff
@@ -0,0 +1,85 @@
...
+class SparcEngageWarnPlugin(Plugin):
+    name = "SPARC engage warn"
+    __version__ = "1.0"
+    __author__ = u"Éric Piel"
+    __license__ = "GPLv2"
+
+    def __init__(self, microscope, main_app):
+        super(SparcEngageWarnPlugin, self).__init__(microscope, main_app)
+        # Allow to run it on pretty much anything (excepted the viewer)
+        if not microscope or microscope.role != "sparc2":
```
pieleric, line 48: Better just check if the chamber tab is shown? Or use the same check as the chamber tab (the actual mirror stage is present).

plugins/sparc_engage_warn.py
```diff
@@ -0,0 +1,85 @@
...
+
+class SparcEngageWarnPlugin(Plugin):
+    name = "SPARC engage warn"
+    __version__ = "1.0"
+    __author__ = u"Éric Piel"
+    __license__ = "GPLv2"
+
+    def __init__(self, microscope, main_app):
+        super(SparcEngageWarnPlugin, self).__init__(microscope, main_app)
+        # Allow to run it on pretty much anything (excepted the viewer)
```
pieleric, line 47: Wrong comment.

plugins/sparc_engage_warn.py
```diff
@@ -0,0 +1,85 @@
...
+* Ensure all detectors are retracted.
+"""
+
+PARK_MSG = """Are you sure you want to park the mirror?
+* Ensure all detectors are retracted.
+"""
+
+
+class SparcEngageWarnPlugin(Plugin):
+    name = "SPARC engage warn"
```
pieleric, line 40: Quite a lame name.

plugins/sparc_engage_warn.py
```diff
@@ -0,0 +1,85 @@
...
+            # => Would engage
+            box = wx.MessageDialog(self.main_app.main_frame,
+                       msg,
+                       caption=action + " mirror",
+                       style=wx.OK | wx.CANCEL | wx.CANCEL_DEFAULT | wx.ICON_QUESTION | wx.CENTER)
+            box.SetOKLabel("&" + action)
+            ans = box.ShowModal()  # Waits for the window to be closed
+            if ans == wx.ID_CANCEL:
+                logging.info("Cancelled moving mirror for action %s", action)
+                self._chamber_tab.panel.btn_switch_mirror.SetValue(False)
```
pieleric, line 81: Add a comment why this is necessary.

PR #1661, opened 2021-02-04
Adding VA's for the XTToolkit

Kleijwegt:


src/odemis/driver/xt_client.py
```diff
@@ -980,6 +980,71 @@ def __init__(self, name, role, parent, hfw_nomag, **kwargs):
...
+        beamlet_index_range = tuple((int(i), int(j)) for i, j in zip(beamlet_index_range_x, beamlet_index_range_y))
+        beamlet_index = self.parent.get_beamlet_index()
+        self.beamletIndex = model.TupleContinuous(
+            tuple(int(i) for i in beamlet_index),  # convert tuple values to integers.,
+            unit=beamlet_index_info["unit"],
+            range=beamlet_index_range,
+            setter=self._setBeamletIndex
+        )
+
+        beam_mode = True if self.parent.get_use_case() == 'MultiBeamTile' else False
```
pieleric, line 1042: You can directly write `beam_mode = (self.parent.get_use_case() == 'MultiBeamTile')`.
(please update also update_settings() function)

BTW, could you rename `beam_mode` to `mbeam_mode` or `multibeam_mode`? Just "beam_mode" sounds like it should be some kind of type of beam, not really just a boolean.

src/odemis/driver/xt_client.py
```diff
@@ -980,6 +980,71 @@ def __init__(self, name, role, parent, hfw_nomag, **kwargs):
...
+            setter=self._setPatternStigmator)
+
+        self.beamShiftTransformtionMatrix = model.ListVA(
+            self.parent.get_dc_coils(),
+            unit=None,
+            readonly=True)
+
+        self.multiprobeRotation = model.FloatVA(
+            numpy.deg2rad(self.parent.get_mpp_orientation()),
+            unit="radians",
```
pieleric, line 1020: In Odemis, for radians, we use "rad".

src/odemis/driver/xt_client.py
```diff
@@ -1070,6 +1135,38 @@ def _updateSettings(self):
...
+            if mpp_rotation != self.multiprobeRotation.value:
+                self.multiprobeRotation._value = mpp_rotation
+                self.multiprobeRotation.notify(mpp_rotation)
+            aperture_index = self.parent.get_aperture_index()
+            if aperture_index != self.apertureIndex.value:
+                self.apertureIndex._value = aperture_index
+                self.apertureIndex.notify(aperture_index)
+            beamlet_index = self.parent.get_beamlet_index()
+            if beamlet_index != self.beamletIndex.value:
+                self.beamletIndex._value = tuple(int(i) for i in beamlet_index)
```
pieleric, line 1164: Could you convert to a tuple of ints *before* the if? Assuming the float is precisely equivalent to the int, it'd work fine as is, but if for some reason it's a tiny bit smaller or bigger (due to some floating error), you would end up updating the beamletIndex every time.

src/odemis/driver/xt_client.py
```diff
@@ -980,6 +980,71 @@ def __init__(self, name, role, parent, hfw_nomag, **kwargs):
...
+            range=pattern_stigmator_range,
+            setter=self._setPatternStigmator)
+
+        self.beamShiftTransformtionMatrix = model.ListVA(
+            self.parent.get_dc_coils(),
+            unit=None,
+            readonly=True)
+
+        self.multiprobeRotation = model.FloatVA(
+            numpy.deg2rad(self.parent.get_mpp_orientation()),
```
pieleric, line 1019: No need to use numpy. You can use just `math.radians(...)`.

src/odemis/driver/xt_client.py
```diff
@@ -1070,6 +1135,38 @@ def _updateSettings(self):
...
+                self.beamStigmator.notify(beam_stigmator)
+            pattern_stigmator = self.parent.get_secondary_stigmator()
+            if pattern_stigmator != self.patternStigmator.value:
+                self.patternStigmator._value = pattern_stigmator
+                self.patternStigmator.notify(pattern_stigmator)
+            beam_shift_transformtion_matrix = self.parent.get_dc_coils()
+            if beam_shift_transformtion_matrix != self.beamShiftTransformtionMatrix.value:
+                self.beamShiftTransformtionMatrix._value = beam_shift_transformtion_matrix
+                self.beamShiftTransformtionMatrix.notify(beam_shift_transformtion_matrix)
+            mpp_rotation = self.parent.get_mpp_orientation()
```
pieleric, line 1154: Convert to radians here too.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -915,6 +916,181 @@ def test_beamlet_index_info(self):
...
+        # Check if VA is read only
+        with self.assertRaises(NotSettableError):
+            self.scanner.beamShiftTransformtionMatrix.value = beamShiftTransformtionMatrix
+
+    def test_multiprobe_rotation_VA(self):
+        mpp_rotation = self.scanner.multiprobeRotation.value
+        self.assertIsInstance(mpp_rotation, float)
+        # Currently the range of the value can be quite big due to different designs for microscopes.
+        self.assertGreaterEqual(mpp_rotation,  - numpy.rad2deg(90))
+        self.assertLessEqual(mpp_rotation, numpy.rad2deg(90))
```
pieleric, line 1025: Please use `math.radians()` here too. No need to use numpy (and get special numpy floats) when standard functions can do the job ;-)

pieleric (changes requested):
Be aware that as it is now, the xtclient will break for "standard" SEMs which don't provide all these special attributes. But I guess this is the reason it's tagged `WIP` :-D I think the plan is to move this to either a separate subclass, or just catch NotImplementedError, and not provide VAs in this case, right?

Otherwise, I only have minor comments.

Kleijwegt:
> Be aware that as it is now, the xtclient will break for "standard" SEMs which don't provide all these special attributes. But I guess this is the reason it's tagged WIP :-D I think the plan is to move this to either a separate subclass, or just catch NotImplementedError, and not provide VAs in this case, right?
> 
> Otherwise, I only have minor comments.

I have indeed made other classes.
@pieleric your comments should be resolved, but they will show outdated since they code is moved.
Can you do another review?
Do we want to squash these two commits before merging? I don't think the history of the move of that code is particularly important/interesting.

src/odemis/driver/xt_toolkit_client.py
```diff
@@ -0,0 +1,233 @@
...
+        # TODO When the new approach of adding update function to a list is implemented in xt_client.py instead of
+        #  overwriting the method updateSettings the method _updateMBSettings can be added to the list of
+        #  functions to be updated.
+
+        # Polling XT client settings
+        super(Scanner, self)._updateSettings()
+        # Polling XTtoolkit settings
+        # TODO K.K. how to remove waiting but not have error because of inheritance.
+        import time
+        time.sleep(1)
```
pieleric, line 151: Why do you need a sleep() here? The code should be executed one after the other, and all be fine, no?

src/odemis/driver/xt_toolkit_client.py
```diff
@@ -0,0 +1,233 @@
...
+        self._updateMBSettings()
```
pieleric, line 152: You can put that whole function here. As hopefully that should be the only code, in addition to the call to super().

pieleric (changes requested):
It makes sense to have the whole "multi" scanner in a separate class, however I'm not convinced by having the classes in a separate module. All this code is for xtclient/xtadapter, so it should all be in the xt_client module. 

Moreover, I'd not differentiate per-se over xtlib vs xttoolkit but more about single-beam vs multi-beam. This second point is a little bit more nitpicking as essentially we currently use the xttoolkit exclusively for multi-beam ;-) However, in the long term, we could imagine having a single-beam scanner with control of the power (only available via xt_toolkit), or controlling the multi-beam via a different interface again (eg, autoscript). For the user (writing the microscope file), that's what matters: have extra functions to control the multi-beam, or fail if they are not available (eg, because xtadapter was incorrectly configured).

I'd suggest the following organization:
 * Put the `xt_toolkit_client.Scanner` as `xt_client.MultiBeamScanner`
 * Adjust the `xt_client.SEM` class to accept a child "mb-scanner" (or any better name that you can find ;-) ) and if there is such child (instead of "scanner"), then instantiate a MultiBeamScanner.

src/odemis/driver/xt_toolkit_client.py
Kleijwegt: This is an error (witht he TODO) I should have fixed. I think I found a better solution than the sleep but I forgot to implement it.
Because of the update thread in the super the polling cannot start directly, because the multibeam part isn't initialized by then. I should change something in the __init__ of the scanner with the order of the VA's and the super.

src/odemis/driver/xt_toolkit_client.py
```diff
@@ -0,0 +1,233 @@
...
+        Parameters
+        ----------
+        address: str
+            server address and port of the Microscope server, e.g. "PYRO:Microscope@localhost:4242"
+        timeout: float
+            Time in seconds the client should wait for a response from the server.
+        """
+        super(SEM, self).__init__(name, role, children, address, daemon=daemon, **kwargs)
+
+        # Overwrite the scanner child created in the parten __init__
```
Kleijwegt, line 50: fix type parten

src/odemis/driver/xt_client.py
```diff
@@ -1909,3 +1921,178 @@ def stop(self, axes=None):
...
+        multibeam_mode = (self.parent.get_use_case() == 'MultiBeamTile')
+        self.multiBeamMode = model.BooleanVA(
+                multibeam_mode,
+                setter=self._setMultiBeamMode
+        )
+
+        # TODO If _updateSettings is updated move this to the top of the __init__ function. For now the update thread
+        #  can only be started after the MB VA's are initialized.
+        # Instantiate the super scanner class with the update thread
+        super(MultiBeamScanner, self).__init__(name, role, parent, hfw_nomag, **kwargs)
```
pieleric, line 2009: Now, I understand why it'd be really nice to have a list of functions to call for updating the settings.
It's fine as it is now though. It's a little odd, but instantiating the VAs before the rest shouldn't cause any issues.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -916,5 +925,189 @@ def test_beamlet_index_info(self):
...
+
+        self.scanner.beamStigmator.value = current_value
+
+    def test_pattern_stigmator_VA(self):
+        current_value = self.scanner.patternStigmator.value
+
+        # Test if directly changing it via the VA works
+        for test_stigmator_value in self.scanner.patternStigmator.range:
+            self.scanner.patternStigmator.value = test_stigmator_value
+            self.assertEqual(test_stigmator_value, tuple(self.microscope.get_secondary_stigmator()))
```
pieleric, line 1009: use `pattern_stigmator` here too.

pieleric (changes requested):
Overall it looks fine.
I see you're still using the secondary_stigmator, but that is deprecated, and patternStigmator should be using pattern_stigmator, right? Or you had a reason to use this "old" call?

src/odemis/driver/test/xt_client_test.py
```diff
@@ -916,5 +925,189 @@ def test_beamlet_index_info(self):
...
+                cls.efocus = child
+            elif child.name == CONFIG_STAGE["name"]:
+                cls.stage = child
+
+    def setUp(self):
+        if TEST_NOHW:
+            self.skipTest("No hardware available.")
+        if self.microscope.get_vacuum_state() != 'vacuum':
+            self.skipTest("Chamber needs to be in vacuum, please pump.")
+        self.xt_type = "xttoolkit" if "xttoolkit" in self.microscope.swVersion.lower() else "xtlib"
```
tepals, line 951: Don't you completely want to fail if the xt_type is xtlib?

src/odemis/driver/test/xt_client_test.py
```diff
@@ -916,5 +925,189 @@ def test_beamlet_index_info(self):
...
+
+        cls.microscope = xt_client.SEM(**CONFIG_MB_SEM)
+
+        for child in cls.microscope.children.value:
+            if child.name == CONFIG_MB_SCANNER["name"]:
+                cls.scanner = child
+            elif child.name == CONFIG_FOCUS["name"]:
+                cls.efocus = child
+            elif child.name == CONFIG_STAGE["name"]:
+                cls.stage = child
```
tepals, line 944: you don't use the stage right?

src/odemis/driver/test/xt_client_test.py
```diff
@@ -54,6 +54,15 @@
                            }
               }
 
+CONFIG_MB_SCANNER = {"name": "mb-scanner", "role": "ebeam", "hfw_nomag": 1}
+CONFIG_MB_SEM = {"name": "sem", "role": "sem", "address": "PYRO:Microscope@192.168.31.138:4242",
+              "children": {"mb-scanner": CONFIG_MB_SCANNER,
+                           "focus": CONFIG_FOCUS,
+                           "stage": CONFIG_STAGE,
```
tepals, line 61: you can also remove the stage here

src/odemis/driver/test/xt_client_test.py
```diff
@@ -953,5 +962,192 @@ def test_beamlet_index_info(self):
...
+        self.assertEqual(current_value, self.scanner.beamletIndex.value)
+
+    def test_multiprobe_mode_VA(self):
+        current_beam_mode = self.scanner.multiBeamMode.value
+        current_aperture_index = self.scanner.apertureIndex.value
+        current_beamlet_index = self.scanner.beamletIndex.value
+
+        for multi_beam_boolean in [True, False, True, False, True]:
+            self.scanner.multiBeamMode.value = multi_beam_boolean
+            time.sleep(6)
```
tepals, line 1137: Is it really necessary to change the boolean this often with a 6 second sleep in between? Including the sleep at the bottom of this test, there is 36 seconds of sleeping in this test.

src/odemis/driver/xt_client.py
```diff
@@ -59,6 +59,12 @@ class SEM(model.HwComponent):
     Driver to communicate with XT software on TFS microscopes. XT is the software TFS uses to control their microscopes.
     To use this driver the XT adapter developed by Delmic should be running on the TFS PC. Communication to the
     Microscope server is done via Pyro5.
+
+    By adding the using 'mb-scanner' child instead of a 'scanner' child this driver extends the class with the addition
+    of the XTtoolkit functionality.XTtoolkit provides extra functionality for the FAST-EM project which xtlib does not
```
tepals, line 64: Missing space before Xttoolkit: functionality.XTtoolkit

src/odemis/driver/xt_client.py
```diff
@@ -1990,3 +2002,178 @@ def stop(self, axes=None):
...
+    rotation and dwell time. This class adds XTtoolkit functionality via the Vigilant Attributes for the pitch,
+    beam stigmator, pattern stigmator, the beam shift transformation matrix (read-only),
+    multiprobe rotation (read-only), aperture index, beamlet index, beam mode (multi/single beam)
+    Whenever one of these attributes is changed, its setter also updates another value if needed.
+    """
+
+    def __init__(self, name, role, parent, hfw_nomag, **kwargs):
+        # First instantiate the xt_toolkit VA's then call the __init__ of the super for the update thread.
+        self.parent = parent
+
```
tepals, line 2020: I wonder if it makes sense to have a check if XTToolkit is running in this constructor. Right now if the type on the adapter side is set to xtlib and you try to start a MultiBeamScanner it will fail because of a NotImplementedError when calling `self.parent.pitch_info()` on line 2022. 
It might be useful to have a check for XTToolkit and give a descriptive error message. What do you think?

src/odemis/driver/xt_client.py
```diff
@@ -1990,3 +2002,178 @@ def stop(self, axes=None):
...
+                self.beamletIndex.notify(beamlet_index)
+            multibeam_mode = (self.parent.get_use_case() == 'MultiBeamTile')
+            if multibeam_mode != self.multiBeamMode.value:
+                self.multiBeamMode._value = multibeam_mode
+                self.multiBeamMode.notify(multibeam_mode)
+        except Exception:
+            logging.exception("Unexpected failure when polling XTtoolkit settings")
+
+    def _setPitch(self, pitch):
+        self.parent.set_pitch(pitch * 1e6)  # Convert from micrometer to meters.
```
tepals, line 2140: Comment is wrong way around, you're converting from meters to micrometers when calling set_pitch

src/odemis/driver/xt_client.py
```diff
@@ -1990,3 +2002,178 @@ def stop(self, axes=None):
...
+        return int(self.parent.get_aperture_index())
+
+    def _setBeamletIndex(self, beamlet_index):
+        self.parent.set_beamlet_index(beamlet_index)
+        new_beamlet_index = self.parent.get_beamlet_index()
+        return tuple(int(i) for i in new_beamlet_index)  # convert tuple values to integers.
+
+    def _setMultiBeamMode(self, multi_beam_mode):
+        # TODO: When changing the beam mode of the microscope we don't want to also change the aperture and beamlet
+        #  index. However, it this is the current implementation direction of TFS. Therefore in the future code like
```
tepals, line 2162: "it this is ...." doesn't make sense

src/odemis/driver/xt_client.py
```diff
@@ -1990,3 +2002,178 @@ def stop(self, axes=None):
...
+        else:
+            self.parent.set_use_case('SingleBeamlet')
+
+        # TODO: Example of an implementation to make sure aperture is unchanged when switching modes, see TODO above.
+        # if self.parent.get_aperture_index() != current_aperture or self.parent.get_beamlet_index() != current_beamlet:
+        #     time.sleep(3)
+        #     self.apertureIndex.value = current_aperture
+        #     self.beamletIndex.value = current_beamlet
+        return (self.parent.get_use_case() == 'MultiBeamTile')
+
```
tepals, line 2178: One empty line at the end of the file is sufficient

tepals (commented):
A few more comments

src/odemis/driver/xt_client.py
Kleijwegt: Would you do that simply via:

```
        if "xttoolkit" in self.microscope.swVersion.lower():
            self.xt_type = "xttoolkit"
        else:
            raise TypeError("Xttoolkit must be running for this test.")
```
            
 And then implement it in the scanner of SEM constructor? (SEM part in the if statement where we instatiate the MB-scanner part)

pieleric (changes requested):

src/odemis/driver/xt_client.py
```diff
@@ -1990,3 +2002,178 @@ def stop(self, axes=None):
...
+            setter=self._setApertureIndex)
+
+        beamlet_index_info = self.parent.beamlet_index_info()
+        beamlet_index_range_x = beamlet_index_info["range"]["x"]
+        beamlet_index_range_y = beamlet_index_info["range"]["y"]
+        beamlet_index_range = tuple((int(i), int(j)) for i, j in zip(beamlet_index_range_x, beamlet_index_range_y))
+        beamlet_index = self.parent.get_beamlet_index()
+        self.beamletIndex = model.TupleContinuous(
+            tuple(int(i) for i in beamlet_index),  # convert tuple values to integers.,
+            unit=beamlet_index_info["unit"],
```
pieleric, line 2076: Please set the unit to `None` here too.

src/odemis/driver/xt_client.py
```diff
@@ -1990,3 +2002,178 @@ def stop(self, axes=None):
...
+    Whenever one of these attributes is changed, its setter also updates another value if needed.
+    """
+
+    def __init__(self, name, role, parent, hfw_nomag, **kwargs):
+        # First instantiate the xt_toolkit VA's then call the __init__ of the super for the update thread.
+        self.parent = parent
+
+        # Add XTtoolkit specific VA's
+        pitch_info = self.parent.pitch_info()
+        # TODO change VA and method names with pitch to delta pitch
```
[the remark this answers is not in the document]
Kleijwegt: I agree but the plan was to implement that with the update in whihc we also change the subcalls to deltaPitch.
So then also the call get/setPitch will be changed to get/setDeltaPitch.
That is planned directly after the merge of this PR.

src/odemis/driver/xt_client.py
tepals: I think the SEM constructor makes the most sense. Then it fails the soonest, which I guess saves time. In the SEM constructor you will need `self._swVersion`. And don't forget to update the text in the raised error.

```
        if "xttoolkit" in self._swVersion.lower():
            self.xt_type = "xttoolkit"
        else:
            raise TypeError("UPDATE THIS TEXT.")
```

src/odemis/driver/test/xt_client_test.py
Kleijwegt: Yes it is unfortunolty, you want to also test if the change from True --> False and the other way around succeeds also if it is preceded by switch in beam mode via the VA. This did fail before.
I have added a comment.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -1132,18 +1131,21 @@ def test_multiprobe_mode_VA(self):
         current_aperture_index = self.scanner.apertureIndex.value
         current_beamlet_index = self.scanner.beamletIndex.value
 
+        # Test multiple time to see if the change from True --> False and the other way around succeeds also if it is
+        # preceded by switch in beam mode via the VA.
         for multi_beam_boolean in [True, False, True, False, True]:
             self.scanner.multiBeamMode.value = multi_beam_boolean
-            time.sleep(6)
+            time.sleep(8)
```
Kleijwegt, line 1138: Upon another check with Wilco he thought that waiting 8 seconds would be better.

tepals (approved):

src/odemis/driver/xt_client.py
```diff
@@ -2158,10 +2163,10 @@ def _setBeamletIndex(self, beamlet_index):
...
 
     def _setMultiBeamMode(self, multi_beam_mode):
-        # TODO: When changing the beam mode of the microscope we don't want to also change the aperture and beamlet
-        #  index. However, it this is the current implementation direction of TFS. Therefore in the future code like
-        #  the uncommented parts below need to be implemented. Note this code still needs to be tested and is
-        #  therefore now just an example of an implementation.
+        # TODO: When changing the beam mode of the microscope changes we don't want to also change the aperture and
+        #  beamlet index. However, it seems that this is the way TFS will be implementing and supporting XTtoolkit.
+        #  Therefore in the future code like the uncommented parts below need to be implemented.
+        #  Note: this code still needs to be tested and is therefore now just an example of an implementation.
```
pieleric, line 2169: BTW, I've just learned that TFS has new plans for the API. At least, when going back to single beam mode, the next version  of their software will restore the previous beamlet index. Aperture might also be automatic.
Let's see...

PR #1662, opened 2021-02-04
Enable Cryo tabs on Imaging Mode

lazem:
…he tab components based on stage position (imaging, loading..)

Subscribe to stage position on alignment tab and call enable_cryo_tab_on_stage_position

### DIFF
--- a/src/odemis/gui/cont/tabs.py
+++ b/src/odemis/gui/cont/tabs.py
@@ -2855,6 +2855,8 @@ class SecomAlignTab(Tab):
 
         self.tab_data_model.tool.subscribe(self._onTool, init=True)
         main_data.chamberState.subscribe(self.on_chamber_state, init=True)
+        # stage will be used to listen to position changes (to enable/disable the tab in the right positions)
+        self.stage = self.tab_data_model.main.stage
 
     def _on_ccd_should_update(self, update):
         """
@@ -2866,14 +2868,6 @@ class SecomAlignTab(Tab):
         Tab.Show(self, show=show)
 
         main_data = self.tab_data_model.main
-        if main_data.role == "cryo-secom":
-            # Enable the tab if the stage is in imaging position, disable it otherwise
-            stage = self.tab_data_model.main.stage
-            if getCurrentPositionLabel(stage.position.value, stage) is IMAGING:
-                self.panel.Enable()
-            else:
-                self.panel.Disable()
-
         # Store/restore previous confocal settings when entering/leaving the tab
         lm = main_data.laser_mirror
         if show and lm:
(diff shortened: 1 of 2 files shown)


lazem:
I was trying `gh pr` tool and it seems I messed up the PR, created a new one! :upside_down_face:

Anyhow, can you please check, if that's fine, I will apply it to the localization tab. 
Thanks @pieleric

src/odemis/gui/util/__init__.py
```diff
@@ -282,3 +283,17 @@ class AttrDict(dict):
...
+    Enable the given tab if the stage is in imaging position, disable it otherwise
+    :param tab:
+    :param stage:
+    :param pos:
+    :return:
+    """
+    if getCurrentPositionLabel(pos, stage) == IMAGING:
+        tab.panel.Enable()
+    else:
+        tab.panel.Disable()
```
lazem, line 299: I'm getting used to creating helper functions to be used by different tabs. Let me know if that's the right place..

Also I saw your suggestion on Jira to disable the streams instead of the whole tab, what is the right way to do that? Thanks

src/odemis/gui/cont/tabs.py
```diff
@@ -2915,9 +2909,12 @@ def Show(self, show=True):
                 f = self._aligner_xy.moveAbs(md[model.MD_FAV_POS_ACTIVE])
                 self._actuator_controller._enable_buttons(False)
                 f.add_done_callback(self._on_align_move_done)
+            if main_data.role == "cryo-secom":
+                self.stage.position.subscribe(self._on_stage_pos, init=True)
         else:
             main_data.is_acquiring.unsubscribe(self._on_acquisition)
             self._aligner_xy.position.unsubscribe(self._on_align_pos)
+            self.stage.position.unsubscribe(self._on_stage_pos)
```
pieleric, line 2917: No, it shouldn't be in the Show(), as this function is called when the user has just pressed the tab to display it (or hide it). Just put the subscribe in the `__init__()` and the unsubscribe in the `terminate()`.

src/odemis/gui/cont/tabs.py
```diff
@@ -3153,6 +3150,15 @@ def _onSEMpxs(self, pixel_size):
         best_mpp = self._sem_view.mpp.clip(best_mpp)
         self._sem_view.mpp.value = best_mpp
 
+    def _on_stage_pos(self, pos):
+        """
+        Called when the stage is moved (and the tab is shown)
+        :param pos: (dict str->float or None) updated position of the stage
+        """
+        if not self.IsShown():
+            return
```
pieleric, line 3159: It's independent of the tab being currently selected ("Shown") or not. => Remove this check.

src/odemis/gui/util/__init__.py
```diff
@@ -282,3 +283,17 @@ class AttrDict(dict):
     def __init__(self, *args, **kwargs):
         super(AttrDict, self).__init__(*args, **kwargs)
         self.__dict__ = self
+
+
+def enable_cryo_tab_on_stage_position(tab, stage, pos):
```
pieleric, line 288: Let's name this function something a little bit more specific, like `enable_tab_on_imaging_position`. Alternatively, we could accept an extra parameter "target", which would be typically `IMAGING`, but could be another position, and name it  `enable_tab_on_stage_position`.

src/odemis/gui/util/__init__.py
pieleric: Let's not take care of the streams for now. It's a lot more work, and only useful for the localization tab. For the alignment tab, it should just be disabled, unless the stage is at IMAGING.

pieleric (changes requested):
My main comment is that the position check should be typically when the tab is not shown. So it doesn't make sense to only subscribed to the position when the stage is shown.

lazem:
Thanks for the `@call_in_wx_main` tip @pieleric!

PR #1663, opened 2021-02-08
gui overlay: make .active a BooleanVA

philipwinklerdelmic:
For FastEM, it's easier to use a VA, so the ROI controllers can subscribe to it.

### DIFF
--- a/src/odemis/gui/comp/overlay/base.py
+++ b/src/odemis/gui/comp/overlay/base.py
@@ -43,7 +43,7 @@ import odemis.gui as gui
 import odemis.util as util
 import odemis.util.conversion as conversion
 from odemis.gui import EVT_BUFFER_SIZE
-from odemis.model import TupleVA
+from odemis.model import TupleVA, BooleanVA
 
 
 class Label(object):
@@ -313,9 +313,8 @@ class Overlay(with_metaclass(ABCMeta, object)):
             self.add_label(label)
 
         # When an overlay is active, it will process mouse events
-        # So, check for this attribute if the sub class needs to process an event only if it's
-        # active.
-        self.active = False
+        # So, check for this VA if the sub class needs to process an event only if it's active.
+        self.active = BooleanVA(False)
 
         # This attribute can be used to determine if the overlay needs to be drawn or not
         self.show = True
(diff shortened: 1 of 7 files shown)


src/odemis/gui/comp/overlay/base.py
```diff
@@ -340,12 +339,12 @@ def __init__(self, cnvs, label=None):
...
         """ Process user generated mouse events """
-        self.active = True
+        self.active.value = True
         self.cnvs.Refresh()
 
     def deactivate(self):
         """ Stop processing user generated mouse events """
-        self.active = False
+        self.active.value = False
         self.cnvs.Refresh()
```
pieleric, line 348: With a VA, typically these functions shouldn't be public. Instead the VA should be written, and a subscriber would call the `.Refresh()`. 
The only drawback is that it means changing every code which calls any of these two functions... which is quite a bit. I'd rather do it now than have this weird interface left over. And, with a clever search and replace, it should be fast!

src/odemis/gui/test/comp_overlay_test.py
```diff
@@ -950,7 +950,7 @@ def test_line_select_overlay(self):
         cnvs.add_view_overlay(tol)
 
         def toggle(evt):
-            if lsol.active:
+            if lsol.active.value:
                 lsol.deactivate()
             else:
                 lsol.activate()
```
pieleric, line 956: You should be able to write `lsol.active.value = not lsol.active.value`

pieleric (changes requested):

philipwinklerdelmic:
I added a commit to replace the .activate() function with a callback. I feel like it's quite a dangerous change, so we should make sure to test a lot before putting it in a future release. @pieleric Can you have a look? I'm not sure if the private ._activate / ._deactivate functions are the best way of handling it.

src/odemis/gui/comp/overlay/base.py
```diff
@@ -313,9 +313,8 @@ def __init__(self, cnvs, label=None):
             self.add_label(label)
 
         # When an overlay is active, it will process mouse events
-        # So, check for this attribute if the sub class needs to process an event only if it's
-        # active.
-        self.active = False
+        # So, check for this VA if the sub class needs to process an event only if it's active.
+        self.active = BooleanVA(False, setter=self._on_active_va)
```
pieleric, line 317: Instead of using the setter, use a subscription. The difference is that the setter is called before the value of the VA is changed. It can decide what should be the actual value set, or raise an error. Here, it's better to call the (de)activate() function after the value has changed.
```python 
self.active = BooleanVA(False)
self.active.subscribe(self._on_active_va)
```

pieleric (changes requested):
It seems that it could work. Just a small remark on the setter.
I'm now curious to see the code using it.

philipwinklerdelmic:
done

PR #1664, opened 2021-02-09
Support "compressed Z-stack" acquisition in TiledAcquisitionTask

lazem:
- Add an extra parameter zsteps for the focus.z positions
- Create zstack for the given focus levels
- Compress all the zstack into a single image, using “maximum intensity projection”

### DIFF
--- a/src/odemis/acq/stitching/_tiledacq.py
+++ b/src/odemis/acq/stitching/_tiledacq.py
@@ -19,6 +19,7 @@ Odemis. If not, see http://www.gnu.org/licenses/.
 """
 from __future__ import division
 
+import copy
 from concurrent.futures._base import RUNNING, FINISHED, CANCELLED, CancelledError
 import logging
 import math
@@ -31,6 +32,7 @@ from odemis.acq.stitching._constants import WEAVER_COLLAGE_REVERSE, REGISTER_IDE
 from odemis.acq.stream import Stream, SEMStream, CameraStream, RepetitionStream, EMStream, ARStream, \
     SpectrumStream, FluoStream, MultipleDetectorStream, util, executeAsyncTask, \
     CLStream
+from odemis.model import DataArray
 from odemis.util import dataio as udataio
 from odemis.util.comp import compute_scanner_fov, compute_camera_fov
 import os
@@ -46,13 +48,12 @@ FOCUS_RANGE_MARGIN = 100e-6
 # Indicate the number of tiles to skip during focus adjustment
 SKIP_TILES = 3
 
-
 class TiledAcquisitionTask(object):
     """
     The goal of this task is to acquire a set of tiles then stitch them together
     """
 
-    def __init__(self, streams, stage, area, overlap, settings_obs=None, log_path=None, future=None):
+    def __init__(self, streams, stage, area, overlap, settings_obs=None, log_path=None, future=None, zlevels=None):
         """
         :param streams: (Stream) the streams to acquire
         :param stage: (Actuator) the sample stage to move to the possible tiles locations
@@ -62,6 +63,7 @@ class TiledAcquisitionTask(object):
             that should be saved as metadata
         :param log_path: (string) directory and filename pattern to save acquired images for debugging
         :param future: (ProgressiveFuture or None) future to track progress, pass None for estimation only
+        :param zlevels: (list(float) or None) focus z positions required zstack acquisition
         """
         self._future = future
         self._streams = streams
@@ -91,6 +93,7 @@ class TiledAcquisitionTask(object):
         self._nx, self._ny = self._getNumberOfTiles()
 
         # To use in re-focusing acquired images in case they fell out of focus
+        # TODO: make adjust focus optional
         self._focus_stream = next((sd for sd in self._streams if sd.focuser is not None), None)
         if self._focus_stream:
             # save initial focus value to be used in the AutoFocus function
@@ -113,6 +116,13 @@ class TiledAcquisitionTask(object):
             self._exporter = dataio.find_fittest_converter(filename)
             self._fn_bs, self._fn_ext = udataio.splitext(filename)
             self._log_dir = os.path.dirname(self._log_path)
+
+        if zlevels:
+            if self._focus_stream is None:
+                logging.warning("No focuser found in any of the streams, only one acquisition will be performed.")
+            self._zlevels = zlevels
+        else:
+            self._zlevels = []
 
     def _getFov(self, sd):
         """
@@ -385,7 +395,14 @@ class TiledAcquisitionTask(object):
         """
         if remaining is None:
             remaining = self._nx * self._ny
-        acq_time = acqmng.estimateTime(self._streams)
+
+        acq_time = 0
+        for stream in self._streams:
+            acq_stream_time = acqmng.estimateTime([stream])
+            if stream.focuser is not None and len(self._zlevels) > 1:
+                # Acquisition time for each stream will be multiplied by the number of zstack levels
+                acq_stream_time *= len(self._zlevels)
+            acq_time += acq_stream_time
 
         # Estimate stitching time based on number of pixels in the overlapping part
         max_pxs = 0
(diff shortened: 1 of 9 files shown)


lazem:
First version, I'll add time estimation and loggin subsequently. Appreciating your feedback @pieleric 
Thanks

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -413,22 +434,105 @@ def save_tile(ix, iy, das):
         # Run in a separate thread
         threading.Thread(target=save_tile, args=(ix, iy, das), ).start()
 
-    def _acquireTile(self, i, ix, iy):
+    def _constructCube(self, images):
+        """
+        Construct xyz cube from a  z stack of images
+        :param images: (list) list of z ordered images
+        :return: (DataArray) the da of the xyz cube
```
pieleric, line 441: Please indicate more precisely the types:
* images (list of DataArray of shape YX)
* return (DataArray of shape ZYX)

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -413,22 +434,105 @@ def save_tile(ix, iy, das):
...
+        for image in images:
+            stack = numpy.dstack(image)
+            stack = numpy.swapaxes(stack, 1, 2)
+            ret.append(stack[0])
+
+        # Add back metadata
+        metadata3d = copy.copy(images[0].metadata)
+        # Extend pixel size to 3D
+        ps_x, ps_y = metadata3d[model.MD_PIXEL_SIZE]
+        ps_z = ZSTEP
```
pieleric, line 454: Let's use the average step size. Something like: `(zlevels[-1] - zlevels[0]) / (len(zlevels) - 1) if len(zlevels) > 1 else 1e-6`

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -413,22 +434,105 @@ def save_tile(ix, iy, das):
...
+
+        # Computer cube centre
+        c_x, c_y = metadata3d[model.MD_POS]
+        c_z = self._zlevels[len(self._zlevels)//2]  # Assuming zlevels are ordered
```
pieleric, line 458: Let's use the center of zlevels (it's going to be different when there is an even number): `(zlevels[0] + zlevels[-1]) / 2`

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -413,22 +434,105 @@ def save_tile(ix, iy, das):
...
+        metadata3d[model.MD_POS] = (c_x, c_y, c_z)
+
+        # For a negative pixel size, convert to a positive and flip the z axis
+        if ps_z < 0:
+            ret = numpy.flipud(ret)
+            ps_z = -ps_z
+
+        metadata3d[model.MD_PIXEL_SIZE] = (ps_x, ps_y, abs(ps_z))
```
pieleric, line 466: The `abs()` is not needed, as that's just taken cared of  by the line above.

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -112,6 +120,19 @@ def __init__(self, streams, stage, area, overlap, settings_obs=None, log_path=No
             self._fn_bs, self._fn_ext = udataio.splitext(filename)
             self._log_dir = os.path.dirname(self._log_path)
 
+        if zsteps:
+            try:
+                self._focus = model.getComponent(role='focus')
```
pieleric, line 125: That's fine, but another way, possibly a little bit more generic would be to use the stream `.focuser` (if it's not `None`). If the focuser is None, then the stream is acquired only once, without using this parameter. It make it easy to indicate which stream should have this effect and which stream not (for instance, we probably want to avoid this for SEMStreams).

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -413,22 +434,105 @@ def save_tile(ix, iy, das):
...
+        Construct xyz cube from a  z stack of images
+        :param images: (list) list of z ordered images
+        :return: (DataArray) the da of the xyz cube
+        """
+        # images is a list of 3 dim data arrays.
+        ret = []
+        for image in images:
+            stack = numpy.dstack(image)
+            stack = numpy.swapaxes(stack, 1, 2)
+            ret.append(stack[0])
```
pieleric, line 448: I think this is an extremely convoluted way to do that. I know this code is not from you, but from the zstack plugin, but now that see it, it's scary! I think that something like this should work (untested!):
```python
# Will fail on purpose if the images contain more than 2 dimensions
ret = numpy.array([im.reshape(im.shape[-2:]) for im in images])
```

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -413,22 +434,105 @@ def save_tile(ix, iy, das):
...
+            logging.debug(f"Moving focus for tile {ix}x{iy} to {z}.")
+            self._focus.moveAbsSync({'z': z})
+            da = self._acquireStreamTile(i, ix, iy, stream)
+            zstack.append(da)
+
+        if self._future._task_state == CANCELLED:
+            raise CancelledError()
+        logging.debug(f"Zstack acquisition for tile {ix}x{iy}, stream {stream.name} finished, compressing data into a single image.")
+        # Convert zstack into a cube
+        fm_cube = self._constructCube(zstack)
```
pieleric, line 494: Please "log" this fm_cube to a separate file. At least at the beginning, it can be handy to check the whole zstack.

src/odemis/acq/stitching/test/tiledacq_test.py
```diff
@@ -184,6 +184,24 @@ def test_get_fov(self):
...
+        # With fm streams
+        settings_obs = SettingsObserver([self.stage])
+        fm_fov = compute_camera_fov(self.ccd)
+        # Using "songbird-sim-ccd.h5" in simcam with tile max_res: (260, 348)
+        area = (0, 0, fm_fov[0] * 2, fm_fov[1] * 2)  # left, top, right, bottom
+        overlap = 0.2
+        zsteps = 21
+        future = acquireTiledArea(self.fm_streams, self.stage, area=area, overlap=overlap, settings_obs=settings_obs, zsteps=zsteps)
+        data = future.result()
+        self.assertEqual(future._state, FINISHED)
```
pieleric, line 200: Please refrain from checking internal state. This could be written `assertTrue(future.done())`

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -45,14 +47,18 @@
 FOCUS_RANGE_MARGIN = 10e-5
 # Indicate the number of tiles to skip during focus adjustment
 SKIP_TILES = 3
-
+# Step value for z stack levels
+ZSTEP = 1e-6  # m
+# Allowed zsteps range for zstack levels
+MINZ = 1
+MAXZ = 51
```
pieleric, line 54: Let's move this to the overview acquisition window.

pieleric (changes requested):
I think it's pretty much in the right direction. My main comment is about the exact argument that the function takes.

src/odemis/gui/win/acquisition.py
```diff
@@ -874,6 +886,21 @@ def _resume_settings(self):
...
+        :returns (list(float) or None) zstack levels for zstack acquisition
+        """
+        focus_value = self._main_data_model.focus.position.value['z']
+        # Clip zsteps value to allowed range
+        zsteps = numpy.clip(self.zstack_steps.current_value, MINZ, MAXZ)
+        if zsteps == 1:
+            return None
+        # Create focus zlevels from the given zsteps number
+        zlevels = list(numpy.linspace(focus_value - (zsteps / 2 * ZSTEP), focus_value + (zsteps / 2 * ZSTEP), zsteps,
+                                      endpoint=True))
```
pieleric, line 901: Default value of endpoint is True, so let's drop that, because it's just the normal behaviour.

src/odemis/gui/win/acquisition.py
```diff
@@ -874,6 +886,21 @@ def _resume_settings(self):
...
+        Calculate the zstack levels from the current focus position and zsteps value
+        :returns (list(float) or None) zstack levels for zstack acquisition
+        """
+        focus_value = self._main_data_model.focus.position.value['z']
+        # Clip zsteps value to allowed range
+        zsteps = numpy.clip(self.zstack_steps.current_value, MINZ, MAXZ)
+        if zsteps == 1:
+            return None
+        # Create focus zlevels from the given zsteps number
+        zlevels = list(numpy.linspace(focus_value - (zsteps / 2 * ZSTEP), focus_value + (zsteps / 2 * ZSTEP), zsteps,
```
pieleric, line 900: You can write `numpy.linspace(...).tolist()`

src/odemis/gui/win/acquisition.py
```diff
@@ -695,7 +703,11 @@ def __init__(self, parent, orig_tab_data):
         # Set parameters for tiled acq
         self.overlap = 0.2
         try:
-            stage_rng = self._main_data_model.stage.getMetadata()[model.MD_POS_ACTIVE_RANGE]
+            stage_md = self._main_data_model.stage.getMetadata()
+            # Hack: Limit active range with overview map range if it's in metadata, until it's provided in the GUI
+            if model.MD_OVERVIEW_RANGE in stage_md and {'x', 'y'}.issubset(stage_md[model.MD_OVERVIEW_RANGE].keys()):
+                stage_md[model.MD_POS_ACTIVE_RANGE].update(stage_md[model.MD_OVERVIEW_RANGE])
+            stage_rng = stage_md[model.MD_POS_ACTIVE_RANGE]
```
pieleric, line 710: This should work fine, but it's a little odd to update the stage_md, as it feels like the stage metadata will be changed. (It will not, because the dictionary is duplicated when passed via Pyro). So let's write it this way:
```python
stage_md = self._main_data_model.stage.getMetadata()
stage_rng = stage_md[model.MD_POS_ACTIVE_RANGE].copy()
if model.MD_OVERVIEW_RANGE in stage_md and {'x', 'y'}.issubset(stage_md[model.MD_OVERVIEW_RANGE].keys()):
    stage_rng.update(stage_md[model.MD_OVERVIEW_RANGE])
```

src/odemis/gui/win/acquisition.py
```diff
@@ -610,6 +613,11 @@ def on_acquisition_done(self, future):
         # Make sure the file is not overridden
         self.btn_secom_acquire.Enable()
 
+# Step value for z stack levels
+ZSTEP = 1e-6  # m
+# Allowed zsteps range for zstack levels
+MINZ = 1
+MAXZ = 51
```
pieleric, line 620: It's already in the XRC file, so it should be fine, right? Let's drop the extra check.

src/odemis/gui/main.xrc
```diff
@@ -6037,7 +6037,44 @@
...
+									<font>
+										<size>12</size>
+										<encoding>UTF-8</encoding>
+									</font>
+								</object>
+								<flag>wxTOP</flag>
+								<border>25</border>
+							</object>
+							<object class="sizeritem">
+								 <object class="UnitFloatSlider" name="zstack_steps">
```
pieleric, line 6058: Let's use just a `UnitIntegerSlider` . It forces it to be an int.

src/odemis/gui/win/acquisition.py
```diff
@@ -874,6 +886,21 @@ def _resume_settings(self):
...
         self.streambar_controller.resume()
 
+    def _get_zstack_levels(self):
+        """
+        Calculate the zstack levels from the current focus position and zsteps value
+        :returns (list(float) or None) zstack levels for zstack acquisition
+        """
+        focus_value = self._main_data_model.focus.position.value['z']
+        # Clip zsteps value to allowed range
+        zsteps = numpy.clip(self.zstack_steps.current_value, MINZ, MAXZ)
```
pieleric, line 896: Do not use `current_value`, it's an internal attribute (with a bad name!). You should use just the standard wxPython `.GetValue()`.

Also, no need to clip, let's trust the GUI.

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -421,49 +417,51 @@ def estimateTime(self, remaining=None):
...
 
-    def _save_tiles(self, ix, iy, das):
+    def _save_tiles(self, ix, iy, das, stream_cube_id=None):
         """
         Save the acquired data array to disk (for debugging)
         """
 
-        def save_tile(ix, iy, das):
-            fn_tile = "%s-%.5dx%.5d%s" % (self._fn_bs, ix, iy, self._fn_ext)
+        def save_tile(ix, iy, das, stream_cube_id):
```
pieleric, line 425: Please mark `stream_cube_id=None` so that it's clearer that it's typically not needed.

src/odemis/acq/stitching/test/tiledacq_test.py
```diff
@@ -194,10 +196,15 @@ def test_compressed_stack(self):
         # Using "songbird-sim-ccd.h5" in simcam with tile max_res: (260, 348)
         area = (0, 0, fm_fov[0] * 2, fm_fov[1] * 2)  # left, top, right, bottom
         overlap = 0.2
+        focus_value = self.focus.position.value['z']
         zsteps = 21
-        future = acquireTiledArea(self.fm_streams, self.stage, area=area, overlap=overlap, settings_obs=settings_obs, zsteps=zsteps)
+        # Create focus zlevels from the given zsteps number
+        zlevels = list(numpy.linspace(focus_value - (zsteps / 2 * 1e-6), focus_value + (zsteps / 2 * 1e-6), zsteps,
+                                       endpoint=True))
```
pieleric, line 203: Please drop `endpoint=True` as it's the standard behaviour.

pieleric (changes requested):
The acquisition side looks fine now. Just a few tweaks needed on the GUI.

src/odemis/gui/main_xrc.py
```diff
@@ -6797,7 +6798,42 @@ def __init_resources():
...
+										<size>12</size>
+										<encoding>UTF-8</encoding>
+									</font>
+								</object>
+								<flag>wxTOP</flag>
+								<border>25</border>
+							</object>
+							<object class="sizeritem">
+								 <object class="UnitIntegerSlider" name="zstack_steps">
+										<size>-1,20</size>
```
lazem, line 6820: It's working fine now but it behaves as `UnitIntegerCtrl` even though in the test case in `comp_slider_test` shows an integer slider.

install/linux/usr/share/odemis/sim/enzel-sim.yaml
```diff
@@ -235,7 +235,7 @@ ENZEL-Sim: {
     },
     metadata: {
         # To change what the "good" focus position is on the simulator
-        # FAV_POS_ACTIVE: {'z': 0.0012},  # focus position at init
+         FAV_POS_ACTIVE: {'z': 0.002},  # focus position at init
```
pieleric, line 238: Why changing it? Is it to start out of focus?

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -400,35 +417,123 @@ def estimateTime(self, remaining=None):
...
+        # Will fail on purpose if the images contain more than 2 dimensions
+        ret = numpy.array([im.reshape(im.shape[-2:]) for im in images])
+
+        # Add back metadata
+        metadata3d = copy.copy(images[0].metadata)
+        # Extend pixel size to 3D
+        ps_x, ps_y = metadata3d[model.MD_PIXEL_SIZE]
+        ps_z = (zlevels[-1] - zlevels[0]) / (len(zlevels) - 1) if len(zlevels) > 1 else 1e-6
+
+        # Computer cube centre
```
pieleric, line 454: Computer -> Compute

src/odemis/gui/main_xrc.py
pieleric: That's odd....

src/odemis/gui/win/acquisition.py
```diff
@@ -695,8 +700,13 @@ def __init__(self, parent, orig_tab_data):
...
         try:
-            stage_rng = self._main_data_model.stage.getMetadata()[model.MD_POS_ACTIVE_RANGE]
-            # left, top, right, bottom
+            stage_md = self._main_data_model.stage.getMetadata()
+            stage_rng = stage_md[model.MD_POS_ACTIVE_RANGE].copy()
+            # Hack: Limit active range with overview map range if it's in metadata, until it's provided in the GUI
+            if model.MD_OVERVIEW_RANGE in stage_md and {'x', 'y'}.issubset(stage_md[model.MD_OVERVIEW_RANGE].keys()):
+                stage_rng.update(stage_md[model.MD_OVERVIEW_RANGE])
+
+            # left, top, right, bottom expected
```
pieleric, line 709: It's more like "left, bottom, right, top" (and that's the right order to store them).

pieleric (approved):
Just very minor comments.
To merge, it'd be good to have 3 commits:
* fixes for the int slider
* changes to the acquisition
* changes to the acquisition GUI window

install/linux/usr/share/odemis/sim/enzel-sim.yaml
lazem: I chose a value in the middle of focus range, but uncommenting this line is more important than the value. Without it, simcam takes the good focus as the current value -> deactive position.

pieleric: Ok, then please also update the comment above to explain that it's needed for not using the initial value, which is at deactive position.

It works actually fine here, and I wonder a little why this is not necessary on my computer. Might be a timing issue between the time it's referenced, and the time the simcam is instantiated.

pieleric (approved):

PR #1665, opened 2021-02-09
xtPatternStigmator

Sabrina85:


tepals (approved):
Looks good, if you end up changing one of the docstrings on the adapter side please also update them here.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -616,6 +616,21 @@ def test_secondary_stigmator(self):
...
+        init_stig = self.microscope.get_pattern_stigmator()
+        stig_x, stig_y = (init_stig[0] + 1e-3, init_stig[1] + 1e-3)
+        stig_x = stig_x if stig_range['x'][0] < stig_x < stig_range['x'][1] else init_stig[0] - 1e-3
+        stig_y = stig_y if stig_range['y'][0] < stig_y < stig_range['y'][1] else init_stig[1] - 1e-3
+        self.microscope.set_pattern_stigmator(stig_x, stig_y)
+        self.assertEqual(self.microscope.get_pattern_stigmator()[0], stig_x)
+        self.assertEqual(self.microscope.get_pattern_stigmator()[1], stig_y)
+        # Set back to the initial stigmator value, so the system is not misaligned after finishing the test.
+        self.microscope.set_pattern_stigmator(*init_stig)
+
```
Kleijwegt, line 633: Why didn't/don't we add something like this in any of the tests?

with self.assertRaises(OSError):
               self.microscope.set_pattern_stigmator(10e10, 10e10)
# Check if values didn't change:
self.assertEqual(self.microscope.get_pattern_stigmator()[0], stig_x)
self.assertEqual(self.microscope.get_pattern_stigmator()[1], stig_y)

Kleijwegt:
Apart from comment: https://github.com/delmic/odemis/pull/1665#discussion_r573752440

I approve, I don't know if we do want these kind of tests included but if we do they should be added.
@tepals What do you think?

src/odemis/driver/test/xt_client_test.py
tepals: That's a good idea. There are a few tests that check if an exception is raised (for instance `test_set_scan_field_size`). I think it was not added in other tests because we simply forgot.

Sabrina85: I will add it for the client and adapter.
@Kleijwegt Is there a reason to specifically check for OSError, or is an Exception enough. We have in some places such a check in the client, but then check generically for an Exception.

tepals: On the adapter side we're raising an IOError, which Pyro5 will turn into an OSError. So we're expecting an OSError. It's cleaner to check for the specific error instead of a generic error. If something else breaks on the adapter side and for instance a ValueError is raised, while we check for a generic error, the test case will pass. If we a ValueError is raised and we check for a OSError, this testcase will fail.

The other test cases should probably also check for OSErrors, and not Exceptions

Sabrina85: Ok then let's change it.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -607,15 +616,38 @@ def test_secondary_stigmator(self):
...
+        self.assertIsInstance(init_stig, tuple)
         stig_x, stig_y = (init_stig[0] + 1e-3, init_stig[1] + 1e-3)
         stig_x = stig_x if stig_range['x'][0] < stig_x < stig_range['x'][1] else init_stig[0] - 1e-3
         stig_y = stig_y if stig_range['y'][0] < stig_y < stig_range['y'][1] else init_stig[1] - 1e-3
         self.microscope.set_secondary_stigmator(stig_x, stig_y)
         self.assertEqual(self.microscope.get_secondary_stigmator()[0], stig_x)
         self.assertEqual(self.microscope.get_secondary_stigmator()[1], stig_y)
+        # Try to set value outside of range
+        with self.assertRaises(OSError):
+            self.microscope.set_secodary_stigmator(stig_range['x'][1] + 1e-3, stig_range['y'][1] + 1e-3)
```
pieleric, line 628: typo: set_secodary_stigmator -> set_secondary_stigmator

src/odemis/driver/test/xt_client_test.py
```diff
@@ -607,15 +616,38 @@ def test_secondary_stigmator(self):
...
+    def test_pattern_stigmator(self):
+        """Test getting and setting the secondary stigmator."""
+        if self.xt_type != 'xttoolkit':
+            self.skipTest("This test needs XTToolkit to run.")
+        stig_range = self.microscope.pattern_stigmator_info()["range"]
+        init_stig = self.microscope.get_pattern_stigmator()
+        self.assertIsInstance(init_stig, tuple)
+        stig_x, stig_y = (init_stig[0] + 1e-3, init_stig[1] + 1e-3)
+        stig_x = stig_x if stig_range['x'][0] < stig_x < stig_range['x'][1] else init_stig[0] - 1e-3
+        stig_y = stig_y if stig_range['y'][0] < stig_y < stig_range['y'][1] else init_stig[1] - 1e-3
```
pieleric, line 641: Let's use `<=` to be exact. Also, technically, as you have just increased the value, you could just write `stig_x <= stig_range['x'][1]`.

pieleric (changes requested):
Just very minor comments.

pieleric (approved):

src/odemis/driver/test/xt_client_test.py
```diff
@@ -563,41 +563,50 @@ def test_fwd_follows_z(self):
...
+        stig_x = stig_x if stig_range['x'][0] < stig_x <= stig_range['x'][1] else init_stig[0] - 1e-6
+        stig_y = stig_y if stig_range['y'][0] < stig_y <= stig_range['y'][1] else init_stig[1] - 1e-6
         new_stig = (stig_x, stig_y)
         self.microscope.set_stigmator(*new_stig)
         self.assertEqual(self.microscope.get_stigmator()[0], new_stig[0])
         self.assertEqual(self.microscope.get_stigmator()[1], new_stig[1])
+        # Try to set value outside of range
+        with self.assertRaises(OSError):
+            self.microscope.set_stigmator(stig_range['x'][1] + 1e-3, stig_range['y'][1] + 1e-3)
+        # Set back to the initial stigmator value, so the system is not misaligned after finishing the test.
```
Kleijwegt, line 591: I think it is nice to check after this call (which raise an error) if the values are still the same as before.
This would catch if somehow and incorrect value might reset something to another value or do something else which is weird.

So here that would be repeading these two lines from above:

> self.assertEqual(self.microscope.get_stigmator()[0], new_stig[0])
> self.assertEqual(self.microscope.get_stigmator()[1], new_stig[1])

Kleijwegt (changes requested):

pieleric (approved):

Kleijwegt (approved):

PR #1670, opened 2021-02-15
driver-tfsbc: bugfix logging

philipwinklerdelmic:
Use comma instead of '%' in logging messages to avoid problems with logging tuples.

It was working before via pyro, maybe because the values were passed as lists? Testing today in a script (without pyro) caused a logging message to fail.


pieleric (approved):
The "standard" way to call logging is to pass as argument anyway, so it's fine.

As to why it did work before, and not anymore, I'm not certain. I don't think Pyro would have any relation with that. The % only match *tuples*, so if before a variable was a list and it was changed to a tuple, that might be the reason. (Pyro does preserve the types, so that shouldn't be directly the cause)

PR #1673, opened 2021-02-17
gui comp combo: getColorbar() only accepts ints

pieleric:
The new versions of numpy only accepts ints for linspace().
So better be clear and force having width and height as int.

### DIFF
--- a/src/odemis/gui/comp/combo.py
+++ b/src/odemis/gui/comp/combo.py
@@ -223,7 +223,7 @@ class ColorMapComboBox(ComboBox):
         else:
             # for painting the items in the popup
             # Draw color map
-            colorbar_width = r.width * COLOBAR_WITH_RATIO
+            colorbar_width = int(round(r.width * COLOBAR_WITH_RATIO))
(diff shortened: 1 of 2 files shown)


src/odemis/gui/comp/combo.py
```diff
@@ -223,7 +223,7 @@ def OnDrawItem(self, dc, rect, item, flags):
         else:
             # for painting the items in the popup
             # Draw color map
-            colorbar_width = r.width * COLOBAR_WITH_RATIO
+            colorbar_width = int(round(r.width * COLOBAR_WITH_RATIO))
```
Kleijwegt, line 226: if width/height is checked to be bigger than 0 why do you need "round" here? int does the same isn't it?

VictoriaDel: @Kleijwegt "int" always rounds down because it cuts off the decimals, but "round" rounds both up and down depending on the decimal. Only "int" returns an integer, "round" returns a floating point number.

VictoriaDel (approved):

PR #1676, opened 2021-02-18
gui util img: clarify merge_tiles()

pieleric:
No behaviour change, just rename variables to clarify that there aren't
really any differences between "image" and "background", they are just
image A and image B.

### DIFF
--- a/src/odemis/gui/util/img.py
+++ b/src/odemis/gui/util/img.py
@@ -2621,37 +2621,38 @@ def insert_tile_to_image(tile, ovv):
     return ovv
 
 
-def merge_screen(im, background):
+def merge_screen(ima, imb):
     """ 
-    Merges two images (im and background) into one using the "screen" operator:
+    Merges two images into one using the "screen" operator:
     f(xA,xB) = xA + xB − xA·xB (with values between 0 and 1, with each channel independent)
-    im, background: DataArray with im.shape = background.shape (either RGB or RGBA)
-    returns RGBA DataArray (of the same YX as im, but with always depth=4)
+    ima, imb: DataArray with ima.shape = imb.shape (either RGB or RGBA)
+    returns RGBA DataArray (of the same YX as ima, but with always depth=4)
     """
-    assert im.shape == background.shape, "Images have different shapes."
-    if im.shape[-1] != 3 and im.shape[-1] != 4:
-        raise ValueError("Ovv images have an invalid number of channels: %d" % (im.shape[-1]))
+    if ima.shape[-1] != 3 and ima.shape[-1] != 4:
+        raise ValueError("Ovv images have an invalid number of channels: %d" % (ima.shape[-1]))
+    if ima.shape[:-1] != imb.shape[:-1]:
+        raise ValueError("Images have different shapes: %s != %s" % (ima.shape, imb.shape))
 
-    md = background.metadata.copy()
-    im = format_rgba_darray(im, 255)  # convert to BGRA
-    im.metadata["dc_keepalpha"] = True
-    background = format_rgba_darray(background, 255)
+    md = imb.metadata.copy()
+    ima = format_rgba_darray(ima, 255)  # convert to BGRA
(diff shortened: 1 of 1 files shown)


src/odemis/gui/util/img.py
```diff
@@ -2621,37 +2621,38 @@ def insert_tile_to_image(tile, ovv):
...
+        raise ValueError("Ovv images have an invalid number of channels: %d" % (ima.shape[-1]))
+    if ima.shape[:-1] != imb.shape[:-1]:
+        raise ValueError("Images have different shapes: %s != %s" % (ima.shape, imb.shape))
 
-    md = background.metadata.copy()
-    im = format_rgba_darray(im, 255)  # convert to BGRA
-    im.metadata["dc_keepalpha"] = True
-    background = format_rgba_darray(background, 255)
+    md = imb.metadata.copy()
+    ima = format_rgba_darray(ima, 255)  # convert to BGRA
```
Kleijwegt, line 2637: Why is the function 'format_rgba_darray' not called 'format_rgba_**to**_darray'?

src/odemis/gui/util/img.py
```diff
@@ -2621,37 +2621,38 @@ def insert_tile_to_image(tile, ovv):
...
-def merge_screen(im, background):
+def merge_screen(ima, imb):
     """ 
-    Merges two images (im and background) into one using the "screen" operator:
+    Merges two images into one using the "screen" operator:
     f(xA,xB) = xA + xB − xA·xB (with values between 0 and 1, with each channel independent)
-    im, background: DataArray with im.shape = background.shape (either RGB or RGBA)
-    returns RGBA DataArray (of the same YX as im, but with always depth=4)
+    ima, imb: DataArray with ima.shape = imb.shape (either RGB or RGBA)
+    returns RGBA DataArray (of the same YX as ima, but with always depth=4)
```
Kleijwegt, line 2629: > but with always depth=4
Maybe changing this to: 'but with always RGBA format' that is slightly clearer than depth 4 I think.

src/odemis/gui/util/img.py
```diff
@@ -2621,37 +2621,38 @@ def insert_tile_to_image(tile, ovv):
     return ovv
 
 
-def merge_screen(im, background):
+def merge_screen(ima, imb):
     """ 
-    Merges two images (im and background) into one using the "screen" operator:
+    Merges two images into one using the "screen" operator:
```
Kleijwegt, line 2626: Maybe you could add in the first sentence how they are merged. Before it seemed that _im_ would be infront of _background_.
Now maybe now the first sentence can be changed to something like:
> Merges two images by averiging between the values into one using the "screen" operator: f(xA,xB) = xA + xB − xA·xB (with values between 0 and 1, with each channel independent)

I'm never used a "screen" operator but since you are clarifying the doc string it would be nice if you can directly have an intuitive idea of how the function works if you the docstring (without going into maths etc.)

Kleijwegt (changes requested):

src/odemis/gui/util/img.py
pieleric: I don't have a very good "intuitive description. "Averaging" would be wrong. After searching online, I saw 
"soft plus", which is the less bad, I've found. I'll update, unless you have a better suggestion.

src/odemis/gui/util/img.py
pieleric: I respectfully disagree. We already just stated that it returns  a RGBA array. So it's not useful to indicate that it's RGBA to explain it. Maybe I could write "of shape YXC, with YX the same as ima, and C=4)".

src/odemis/gui/util/img.py
pieleric: Just bad naming. The function actually has a "TODO: get a better name". One day someone should take care of that... It's actually not "to darray", it's "to bgra"!

Kleijwegt (approved):

PR #1680, opened 2021-02-19
Fix metadata upload of xt_client.py

Kleijwegt:
The metadata should be updated via the respective components (children) and not via the parents. This way the affects in the yaml file actually works correctly. Otherwise this should go via the parent object indirectly.

@pieleric EA is currenlty not available for testing I only tested this on the simple simulator on my computer, is that ok or do we want this also to be testen on real Hw?

### DIFF
--- a/src/odemis/driver/xt_client.py
+++ b/src/odemis/driver/xt_client.py
@@ -1022,7 +1022,7 @@ class Scanner(model.Emitter):
 
         # If scaled up, the pixels are bigger
         pxs_scaled = (pxs[0] * self.scale.value[0], pxs[1] * self.scale.value[1])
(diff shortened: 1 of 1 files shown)


tepals:
> The metadata should be updated via the respective components (children) and not via the parents. This way the affects in the yaml file actually works correctly. Otherwise this should go via the parent object indirectly.
> 
> @pieleric EA is currenlty not available for testing I only tested this on the simple simulator on my computer, is that ok or do we want this also to be testen on real Hw?

You should be able to test this on fumo beta right?

pieleric:
For such simple change, unrelated to the communication with the hardware, I don't think it's required to test on the real hardware. I've seen it working with the basic simulator, which is enough to convince me the code is fine :-)

PR #1681, opened 2021-02-22
acq/gui: fix huge time estimation causing chaos in progress bar

pieleric:


src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -736,7 +736,7 @@ def acquireTiledArea(streams, stage, area, overlap=0.2, settings_obs=None, log_p
     if not mem_sufficient:
         raise IOError("Not enough RAM to safely acquire the overview: %g GB needed" % (mem_est / 1024 ** 3,))
 
-    future.set_progress(end=mem_est + time.time())
+    future.set_progress(end=task.estimateTime() + time.time())
```
Kleijwegt, line 739: Do futures have a timeout? Or maybe some function which you can connect to a timeout? For example if updating the live image takes longer than the actual scan you want the live update future to go into timeout at the moment the scan is finished. (I think for the live update itself this wouldn't be useful but it is the first example which came to my mind)
Not related to this PR but maybe we want such a feature integrated in the future in the future(class).

src/odemis/gui/util/widgets.py
```diff
@@ -334,13 +334,13 @@ def _update_progress(self):
         past = now - self._start
         left = max(0, self._end - now)
         prev_left = self._prev_left
+        ratio = past / (past + left)
```
Kleijwegt, line 337: What happens if (past + left) = 0 and this code is ran?
It would create a ZeroDivisionError isn't it? Shouldn't that be prevented or catched?

src/odemis/acq/stitching/_tiledacq.py
pieleric: Yes, there is such a thing. You can call `future.result(timeout)`, in which case it'll raise an error if the tasks takes more than you expect. With CancellableFutures (special on Odemis), you then can decide to stop the task. (Otherwise, the only thing you can really do is to leave the task running in the background while you're not waiting for it).

src/odemis/gui/util/widgets.py
pieleric: Indeed, good remark. This is the kind of code where it should never happen... but I'm sure we'll manage to hit it one day anyway! I've updated the code to handle gracely such case.

Kleijwegt (approved):
I think it is good if also someone else (Bassim) approves this PR before merging.
I only spotted the ratio issue.

lazem (approved):
Good catch for the` future.set_progress` and nice solution for the progress bar range computation. 
Believe it looks good now.

PR #1682, opened 2021-02-23
add software binning to ueye camera

pieleric:


pieleric:
The ueye code is untested on the real hardware. It should done before merging this pull-request.

tepals (approved):
Didn't see anything strange

Mahmood-B (commented):
@pieleric  the ueye code is not tested on hardware, but also not on software (unit tests)? I mean there are no testing for the additions in the file of ueye?

Helsloot (changes requested):
I agree with Mahmood's comment about division-by-zero. It would be nice to have that check. Though I'm also wondering how necessary it is to fool proof the inner workings of Odemis like that. On the other hand, adding such a check definitely won't hurt.

pieleric:
> @pieleric the ueye code is not tested on hardware, but also not on software (unit tests)? I mean there are no testing for the additions in the file of ueye?
Actually, the ueye test cases inherit from the "cam_test_abs" which automatically tests binning if the driver provides it. So it should be tested now. That said, there is no uEye simulator, and I had access to the hardware for such a short period that I didn't have time to run the test cases. So there should be test cases, but they haven't been run yet. I'll try to run them next time I have access to such CCD.

Helsloot (approved):

PR #1683, opened 2021-02-23
gui comp text SuggestTextCtrl: make the dropdown at least 300px width

pieleric:
With the addition of the combo-box to select the colourmap, the text
got quite small. It's pretty much impossible to read anything in the
dropdown, especially as it has a bigger text.

=> Force to have min width. 300 px works well in test.

### DIFF
--- a/src/odemis/gui/comp/text.py
+++ b/src/odemis/gui/comp/text.py
@@ -68,6 +68,9 @@ class ChoiceListCtrl(wx.ListCtrl, listmix.ListCtrlAutoWidthMixin):
         listmix.ListCtrlAutoWidthMixin.__init__(self)
 
 
+LIST_MIN_WIDTH = 300  # px
+
+
 class SuggestTextCtrl(wx.TextCtrl, listmix.ColumnSorterMixin):
     def __init__(self, parent, choices=None, drop_down_click=True,
                  col_fetch=-1, col_search=0, hide_on_no_match=True,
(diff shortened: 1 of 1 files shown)


src/odemis/gui/comp/text.py
```diff
@@ -328,18 +332,21 @@ def _showDropDown(self, show=True):
         Either display the drop down list (show = True) or hide it (show = False).
         """
         if show:
-            size = self.dropdown.GetSize()
+            dwidth, dheight = self.dropdown.GetSize()
+            # Use the width of the text control
             width, height = self.GetSize()
+            width = max(width, LIST_MIN_WIDTH)
```
Kleijwegt, line 338: Wouldn't it make more sense to add this check/line of code to line 99/100?
So:
>self._screen_size = (max(wx.SystemSettings.GetMetric(wx.SYS_SCREEN_X),LIST_MIN_WIDTH),
                                wx.SystemSettings.GetMetric(wx.SYS_SCREEN_Y))

pieleric: No, the screen size is only there to limit the position of the dropdown list, so that it's always fully within the screen limits if opening the dropdown very near the border.
The LIST_MIN_WIDTH is the minimum size of the widget. It's typically way smaller than the screen size (unless the user has a tiny tiny screen ;-) )

philipwinklerdelmic (approved):

PR #1684, opened 2021-02-23
gui model dye: Also fix the other HTML characters in the fluodb

pieleric:
Now the data also contains some &amp; and &quot;.
Let's convert them too.

### DIFF
--- a/src/odemis/gui/model/dye.py
+++ b/src/odemis/gui/model/dye.py
@@ -45,8 +45,8 @@ def _clean_up_name(name):
     name = name.strip()
 
     # Some names are HTML escaped, we could use HTMLParser().unescape, but for
-    # now, we keep it simple, as there is only one character escaped
-    name = name.replace("&#39;", "'")
+    # now, we keep it simple, as there are only a few characters escaped.
+    name = name.replace("&#39;", "'").replace("&amp;", "&").replace("&quot;", "\"")
 
     # first letter upper-case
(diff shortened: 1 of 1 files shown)


src/odemis/gui/model/dye.py
```diff
@@ -45,8 +45,8 @@ def _clean_up_name(name):
     name = name.strip()
 
     # Some names are HTML escaped, we could use HTMLParser().unescape, but for
-    # now, we keep it simple, as there is only one character escaped
-    name = name.replace("&#39;", "'")
+    # now, we keep it simple, as there are only a few characters escaped.
+    name = name.replace("&#39;", "'").replace("&amp;", "&").replace("&quot;", "\"")
```
Kleijwegt, line 49: Is this really simpler than:
>		parser = html.parser.HTMLParser()
>		name = (parser.unescape(text))
		
Or is using the HTMLParser more complicated than this?

pieleric: The problem is that I don't know if all the database is actually properly encoded in HTML. Especially, the database is regularly updated. It might work now, but not in the future. So we need to handle the potential case when unescaping HTML fails. Probably just another 3 easy lines though...
Let's switch to it when the current code doesn't fit on a single line anymore :-D

Kleijwegt: That sounds fair ;)

Kleijwegt (approved):

PR #1689, opened 2021-02-28
plugins cli_rgb: store filter settings in config file

pieleric:
Avoids the user to reselect the filter every time Odemis is restarted.

This also gives a good example of how to use a config file in a plugin.


Helsloot (approved):

Mahmood-B (commented):
Don't have specific comments

PR #1692, opened 2021-03-03
ENZEL: Feature Bookmarking

lazem:
The overview image (acquired in SEM and FM) shows the entire sample (possibly with some gaps). The user needs to navigate through this “map” to (1) investigate whether a “site” is interesting, by using FM Z-stacks and (2) go back and forth between sites during milling. So the main goal is to allow the user to mark these sites on top of the  map, and be able to store the settings for each of this sites.

### DIFF
--- a/src/odemis/gui/comp/overlay/view.py
+++ b/src/odemis/gui/comp/overlay/view.py
@@ -61,11 +61,10 @@ class CrossHairOverlay(base.ViewOverlay):
         self.colour = conversion.hex_to_frgba(colour)
         self.size = size
 
-    def draw(self, ctx):
-        """ Draw a cross hair to the Cairo context """
-
-        center = self.cnvs.get_half_view_size()
-
+    def _draw_crosshair(self, ctx, center):
+        """
+        Draw cross hair given Cairo context and center position
+        """
         tl = (center[0] - self.size, center[1] - self.size)
         br = (center[0] + self.size, center[1] + self.size)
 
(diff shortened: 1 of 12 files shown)


src/odemis/gui/comp/overlay/view.py
```diff
@@ -1464,6 +1464,37 @@ def on_left_up(self, evt):
     def draw(self, ctx):
         pass
 
+class StagePointSelectOverlay(PointSelectOverlay):
+    """ Overlay for moving the stage (in physical coordinates) upon the selection of canvas points"""
+
+    def __init__(self, cnvs):
+        PointSelectOverlay.__init__(self, cnvs)
```
Kleijwegt, line 1471: Why isn't chosen to use super here? Of because PointSelectOverlay is already the parent I think the entire init can be omited

Kleijwegt (approved):

src/odemis/gui/cont/tabs.py
```diff
@@ -2039,6 +2078,20 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
+        Load the overview view with the given list of acquired static streams
+        :param streams: (list of StaticStream) the newly acquired static streams from the localization tab
+        """
+        # Replace the old streams with the newly acquired ones in the view
+        overview_view = self._tab_data_model.focussedView.value
+        existing_streams = overview_view.getStreams()
+        for stream in streams:
+            ex_st = next((ex_st for ex_st in existing_streams if ex_st.acquisitionType.value == stream.acquisitionType.value and ex_st.name.value == stream.name.value), None)
+            if ex_st:
+                overview_view.removeStream(ex_st)
```
pieleric, line 2092: See my comments on the "clear_data" PR. I think we could simplify by checking just for the name.

src/odemis/gui/cont/tabs.py
```diff
@@ -2039,6 +2078,20 @@ def __init__(self, name, button, panel, main_frame, main_data):
         panel.stage_align_btn_m_aligner_z.Bind(wx.EVT_BUTTON, self._on_aligner_btn)
         panel.btn_cancel.Bind(wx.EVT_BUTTON, self._on_cancel)
 
+    def load_data(self, streams):
```
pieleric, line 2081: Let's call it `load_overview_data`, to clarify.

src/odemis/gui/cont/tabs.py
```diff
@@ -2039,6 +2078,20 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
+        """
+        Load the overview view with the given list of acquired static streams
+        :param streams: (list of StaticStream) the newly acquired static streams from the localization tab
+        """
+        # Replace the old streams with the newly acquired ones in the view
+        overview_view = self._tab_data_model.focussedView.value
+        existing_streams = overview_view.getStreams()
```
pieleric, line 2088: It'd be best to hard-code the view. For now there is only one, so you end-up always with the right one. However, the tab should support eventually other view (eg, chamber view).

src/odemis/gui/cont/tabs.py
```diff
@@ -515,6 +537,10 @@ def load_data(self, data):
             scont = self._overview_stream_controller.addStream(s, add_to_view=True)
             scont.stream_panel.show_remove_btn(True)
```
pieleric, line 539: I'd suggest to have a dedicated `load_overview_data()` function, as loading the output of an acquisition will be a little different. 

For now add_to_view=True` means it'll be added to all compatible views, so both top-left and top-right views. It should only be shown on the top-left view. A simple way would be to pass as argument `add_to_view=overview_view`. It'd still allow the user to display the overview stream in the top-right view, but that's maybe nice for the user anyway.

BTW, Mahmoud is going to have to do something very similar with the output of the acquisition, so you might want to synchronize with him.

pieleric (changes requested):

src/odemis/gui/cont/tabs.py
```diff
@@ -2075,16 +2082,20 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
-        overview_view = self._tab_data_model.focussedView.value
+
+        overview_view = next((view for view in self._tab_data_model.views.value if type(view) == guimod.FeatureOverviewView), None)
+        if not overview_view:
+            logging.warning("Could not find view of type FeatureOverviewView.")
+            return
         existing_streams = overview_view.getStreams()
         for stream in streams:
-            ex_st = next((ex_st for ex_st in existing_streams if ex_st.acquisitionType.value == stream.acquisitionType.value and ex_st.name.value == stream.name.value), None)
+            ex_st = next((ex_st for ex_st in existing_streams if ex_st.name.value == stream.name.value), None)
```
lazem: You can check this function, loading acquired streams should be similar to this. @Mahmood-B

Mahmood-B: Sure, I'll have a look

Mahmood-B (approved):
It looks good overall

pieleric (approved):

PR #1701, opened 2021-03-15
gui comp text SuggestTextCtrl: try harder finding the text

pieleric:
If the text search is not found at the beginning of a the choices,
try to find the first choices which contains the text *somewhere*.

### DIFF
--- a/src/odemis/gui/comp/text.py
+++ b/src/odemis/gui/comp/text.py
@@ -95,7 +95,7 @@ class SuggestTextCtrl(wx.TextCtrl, listmix.ColumnSorterMixin):
         self._hide_on_no_match = hide_on_no_match
         self._select_callback = select_callback
         self._entry_callback = entry_callback
-        self._match_function = match_function
+        self._match_function = self._match_start_text if match_function is None else match_function
         self._screen_size = (wx.SystemSettings.GetMetric(wx.SYS_SCREEN_X),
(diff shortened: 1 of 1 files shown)


src/odemis/gui/comp/text.py
```diff
@@ -194,22 +194,32 @@ def onEnteredText(self, event):
...
-                toSel = item.GetId()
-                self.dropdownlistbox.Select(toSel)
+                found = numCh
                 break
-        if not found:
+
+        # Try to match any part of the text if standard matching didn't return anything
+        if found is None:
+            for numCh, choice in enumerate(choices):
+                if text.lower() in choice.lower():
```
Mahmood-B, line 213: You could move this if statement as a third `elif` statement in the for loop. I think then it would be more organized.

Mahmood-B (commented):
The code would be more readable if you could add comments and/or doc string

src/odemis/gui/comp/text.py
pieleric: As an `elif`, it'd have a different behaviour: it would find any text that matches anywhere, all the time, but we want it to happen only as a fallback. I've changed a little the code to make it clearer by use a for/else.

pieleric:
Refactored the code to make it easier to read, and added comments.

Mahmood-B (approved):

PR #1704, opened 2021-03-16
util filename: fix {year} being replaced by %Y

pieleric:
We had forgotten to convert to the time, so the year would be litteraly
replaced as %Y.


src/odemis/util/filename.py
```diff
@@ -170,7 +170,7 @@ def generate_fn(ptn, count):
                           dshrtrev=time.strftime("%d%m%y"),
                           dateshrt_hyphen=time.strftime("%y-%m-%d"),
                           dshrtrev_hyphen=time.strftime("%d-%m-%y"),
-                          year="%Y",
```
Mahmood-B, line 173: Now it makes sense why it used to crush

Mahmood-B (approved):

PR #1706, opened 2021-03-17
xtclient added VAs for scan mode and beam power

tepals:


Sabrina85 (approved):
Looks good!

src/odemis/driver/test/xt_client_test.py
```diff
@@ -248,6 +248,20 @@ def test_rotation(self):
...
+        init_scan_mode = self.scanner.scanMode.value
+        self.assertTrue(init_scan_mode in possible_modes)
+        for mode in possible_modes:
+            self.scanner.scanMode.value = mode
+            self.assertEqual(self.microscope.get_scan_mode(), mode)
+
+        with self.assertRaises(IndexError):
+            self.scanner.scanMode.value = "fake_scan_mode"
+
+        self.scanner.scanMode.value = init_scan_mode
```
Kleijwegt, line 263: Don't you want to also add the scan_mode and beam_power to the update_settings method to ensure that if these settings are updated on the microscope itself the VA's are also updated?
If yes then this should also be tested in this test case.

pieleric (changes requested):
We already have standard interfaces for these settings in Odemis, and they are differently named/handled.
The beam power is just called "power" (as it's on the "e-beam", we don't need to add "beam").

For the scanning mode it's more complicated. For the "external" mode, we have a BooleanVA,  called "external". When it's True, the external mode is active.
The other scanning modes are controlled via the resolution. If you set a resolution of 1x1, and start the DataFlow, then the spot mode is activated. If you can retrieve the data, you should return a DataArray of shape 1x1 containing the detector reading (though, typically that would work only with long dwell time). I've never encountered "line mode", but in the continuity, you could say that if the resolution is "Nx1", then it's line mode. That assumes that line mode is always horizontal.

tepals:
> For the scanning mode it's more complicated. For the "external" mode, we have a BooleanVA, called "external". When it's True, the external mode is active.
> The other scanning modes are controlled via the resolution. If you set a resolution of 1x1, and start the DataFlow, then the spot mode is activated. If you can retrieve the data, you should return a DataArray of shape 1x1 containing the detector reading (though, typically that would work only with long dwell time). I've never encountered "line mode", but in the continuity, you could say that if the resolution is "Nx1", then it's line mode. That assumes that line mode is always horizontal.

I had a look at the external VAs in the tescan and zeiss drivers. When the BooleanVA is True the external mode is active. What should I set the scan mode to when someone wants to do: `setExternal(False)`? Should it then go to full frame?

pieleric:
> What should I set the scan mode to when someone wants to do: `setExternal(False)`? Should it then go to full frame?
If it's acquiring, it should go to whichever mode would be selected when the acquisition is going to start, based on the resolution. Otherwise, if it's not acquiring, indeed going to full_frame would probably be fine.

tepals:
@pieleric I talked to Job and he said it's not possible to get a read out in spot mode. Is it then still necessary to support spot mode? For FAST-EM it is not needed. He also said we won't need line mode.

pieleric:
If we don't need it now, I'm fine if we don't support it. I prefer to not support it at all than to have a hacky way to support it. Because if we have a hacky way, someone might write code that use this, and eventually we'll break it.
Note that it's not necessarily an issue if we don't get the data back in spot mode. That's quite common actually. In such case, we just return an empty DataArray.

src/odemis/driver/xt_client.py
```diff
@@ -1120,11 +1120,8 @@ def __init__(self, name, role, parent, hfw_nomag, **kwargs):
         pxs_scaled = (pxs[0] * self.scale.value[0], pxs[1] * self.scale.value[1])
         self._metadata[model.MD_PIXEL_SIZE] = pxs_scaled
 
-        self.scanMode = model.VAEnumerated(
-                self.parent.get_scan_mode(),
-                setter=self.parent.set_scan_mode,
-                choices={"external", "full_frame", "spot", "line"}
-        )
+        emode = self._getExternal()
+        self.external = model.BooleanVA(bool(emode), setter=self._setExternal)
```
pieleric, line 1124: Let's drop the `bool()`, as that the only type that self._getExternal() returns. Maybe you could rename is to `_isExternal()` to avoid confusing with the VA getter. (But that's a little nitpicking)

pieleric (changes requested):

src/odemis/driver/xt_client.py
tepals: I don't understand your sentence. What are you suggesting to rename to `_isExternal()`?

pieleric: Indeed, I forgot an import word in my sentence :see_no_evil: . I meant rename `_getExternal()` to ` _isExternal()`. At least, it'd make it even clearer that it returns a bool.

tepals:
@pieleric I renamed `_getExternal` to `_isExternal` and removed the `bool()`. I tested it on the system and everything is working.

pieleric (approved):

PR #1707, opened 2021-03-18
Modify 3dof Stage referencing procedure

lazem:
This is an updated calibration procedure for referencing the 3dof stage. It entails:

- Separate autozero referencing mode from normal mode.
- Set zero position directly to automatically adjust the logical scale offset.

### DIFF
--- a/src/odemis/driver/smaract.py
+++ b/src/odemis/driver/smaract.py
@@ -2549,15 +2549,12 @@ class MCS2(model.Actuator):
             self._set_accel(channel, accel)
             self._set_hold_time(channel, hold_time)
 
-            try:
-                # Log referencing mode, and warn if it's not normal (autozero)
-                ref_mode = self.GetProperty_i32(SA_CTLDLL.SA_CTL_PKEY_POSITIONER_TYPE, channel)
-                log_lvl = logging.INFO
-                if ref_mode & SA_CTLDLL.SA_CTL_REF_OPT_BIT_AUTO_ZERO:
-                    log_lvl = logging.WARNING
-                logging.log(log_lvl, "Current referencing mode = {}.".format(ref_mode))
-            except SA_CTLError:
-                    raise
+            # Log referencing mode, and warn if it's not normal (autozero)
+            ref_mode = self.GetProperty_i32(SA_CTLDLL.SA_CTL_PKEY_POSITIONER_TYPE, channel)
+            log_lvl = logging.INFO
+            if ref_mode & SA_CTLDLL.SA_CTL_REF_OPT_BIT_AUTO_ZERO:
+                log_lvl = logging.WARNING
+            logging.log(log_lvl, "Current referencing mode = {}.".format(ref_mode))
 
         self.position = model.VigilantAttribute({}, readonly=True)
 
--- a/src/odemis/driver/smaract.py
+++ b/src/odemis/driver/smaract.py
@@ -2549,6 +2549,17 @@ class MCS2(model.Actuator):
             self._set_accel(channel, accel)
             self._set_hold_time(channel, hold_time)
 
+            try:
+                # Log referencing mode, and warn if it's not normal
+                ref_mode = self.GetProperty_i32(SA_CTLDLL.SA_CTL_PKEY_POSITIONER_TYPE, channel)
+                log_lvl = logging.INFO
+                if ref_mode != SA_CTLDLL.SA_CTL_REF_OPT_BIT_NORMAL:
+                    log_lvl = logging.WARNING
+                logging.log(log_lvl, "Current referencing mode = {}.".format(ref_mode))
+            except SA_CTLError:
+                if locator != "fake":
+                    raise
+
         self.position = model.VigilantAttribute({}, readonly=True)
 
         # Indicates moving to a deactive position after referencing.
@@ -3081,9 +3092,6 @@ class MCS2(model.Actuator):
                     channel = self._axis_map[a]
                     moving_channels.add(channel)
                     self.referenced._value[a] = False
-                    # set property key for normal referencing.
-                    self.SetProperty_i32(SA_CTLDLL.SA_CTL_PKEY_REFERENCING_OPTIONS,
-                            channel, SA_CTLDLL.SA_CTL_REF_OPT_BIT_NORMAL)
                     self.Reference(channel)  # search for the negative limit signal to set an origin
 
                 self._waitEndMove(future, moving_channels, time.time() + 100)  # block until it's over
--- a/src/odemis/driver/smaract.py
+++ b/src/odemis/driver/smaract.py
@@ -2550,14 +2550,13 @@ class MCS2(model.Actuator):
             self._set_hold_time(channel, hold_time)
 
             try:
-                # Log referencing mode, and warn if it's not normal
+                # Log referencing mode, and warn if it's not normal (autozero)
                 ref_mode = self.GetProperty_i32(SA_CTLDLL.SA_CTL_PKEY_POSITIONER_TYPE, channel)
                 log_lvl = logging.INFO
-                if ref_mode != SA_CTLDLL.SA_CTL_REF_OPT_BIT_NORMAL:
+                if ref_mode & SA_CTLDLL.SA_CTL_REF_OPT_BIT_AUTO_ZERO:
                     log_lvl = logging.WARNING
                 logging.log(log_lvl, "Current referencing mode = {}.".format(ref_mode))
             except SA_CTLError:
-                if locator != "fake":
                     raise
 
         self.position = model.VigilantAttribute({}, readonly=True)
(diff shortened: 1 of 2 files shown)


util/saconfig
```diff
@@ -96,39 +104,45 @@ def set_zero(dev, channels):
...
-    logging.info("Setting scale offsets. Current position: %s", dev.position.value)
+    # logging.info("Setting scale offsets. Current position: %s", dev.position.value)
+    # for channel in channels:
+    #     logging.info("Setting the scale offset for channel %d", channel)
+    #     scale_offset = dev.GetProperty_i64(smaract.SA_CTLDLL.SA_CTL_PKEY_LOGICAL_SCALE_OFFSET, channel)
+    #     position = dev.GetProperty_i64(smaract.SA_CTLDLL.SA_CTL_PKEY_POSITION, channel)
+    #     # set the new scale offset
+    #     dev.SetProperty_i64(smaract.SA_CTLDLL.SA_CTL_PKEY_LOGICAL_SCALE_OFFSET, channel, scale_offset + position)
 
+    logging.info("Setting current position to 0. This automatically adjusts the logical scale offset.")
```
pieleric, line 115: Could you still log the current position? It can be handy to know by how much it shifted.

pieleric (changes requested):

util/saconfig
```diff
@@ -61,74 +62,95 @@ def calibrate(dev, channels):
...
 
+def reference(dev, channels, mode=NORMAL_REF):
+    # Set referencing mode, currently either normal or autozero
+    if mode == NORMAL_REF:
+        set_normal_reference(dev, channels)
+    elif mode == AUTO_ZERO_REF:
+        set_autozero_reference(dev, channels)
+    # Run the referencing
+    logging.info("Starting reference...")
+    dev.reference().result()
```
pieleric, line 102: The driver now needs an "axes" argument. (This got recently broken).
You need to pass a set of the "channels" as strings (because we named the axis as the string version of the channel numbers).

util/saconfig
```diff
@@ -141,13 +163,14 @@ def main(args):
...
+    axes = {str(c): {"range": [-1, 1], "unit": "m", "channel": c} for c in channels}
 
     dev = smaract.MCS2(locator=locator, axes=axes, **CONFIG_3DOF)
-    
+
     calibrate(dev, options.calibrate)
-    autozero(dev, options.autozero)
+    reference(dev, options.autozero, mode=AUTO_ZERO_REF)
+    reference(dev, options.reference, mode=NORMAL_REF)
     set_zero(dev, options.setzero)
```
pieleric, line 174: Please, protect the calls to each of these function with a `if options.xxx:`.
Otherwise, we end-up calling each function. The functions eventually should do nothing, as no axes is passed, but it'd feel "safer".

pieleric (changes requested):

util/saconfig
```diff
@@ -61,74 +62,95 @@ def calibrate(dev, channels):
...
     parser.add_argument("--log-level", dest="loglev", metavar="<level>", type=int,
                         default=1, help="set verbosity level (0-2, default = 1)")
 
     parser.add_argument('--calibrate', metavar='N', type=int, nargs='+', default=[],
-                    help="Run the calibration command. Specify the channel to calibrate (0, 1, 2)")
+                        help="Run the calibration command. Specify the channel to calibrate (0, 1, 2)")
 
     parser.add_argument('--autozero', metavar='N', type=int, nargs='+', default=[],
-                    help="Run referencing with autozero functionality. Specify the channel to calibrate (0, 1, 2)")
+                        help="Run referencing with set_autozero_reference functionality. Specify the channel to calibrate (0, 1, 2)")
```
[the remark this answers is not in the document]
lazem: Autozero mode is still there, though we are not currently using it. Renamed the help text abit to be clearer.

util/saconfig
```diff
@@ -61,74 +62,95 @@ def calibrate(dev, channels):
...
         dev.SetProperty_i32(smaract.SA_CTLDLL.SA_CTL_PKEY_REFERENCING_OPTIONS,
                             channel, smaract.SA_CTLDLL.SA_CTL_REF_OPT_BIT_NORMAL)  # normal
 
 
+def reference(dev, channels, mode=NORMAL_REF):
+    # Set referencing mode, currently either normal or autozero
+    if mode == NORMAL_REF:
+        set_normal_reference(dev, channels)
+    elif mode == AUTO_ZERO_REF:
+        set_autozero_reference(dev, channels)
```
pieleric, line 99: The more I'm looking at the code, the less it makes sense. 
`dev.reference()` calls `_doReference()`, which... sets the referencing to normal! So setting autozero would have no effect here. I'm really confused, because I'm pretty sure we've run this code, and it did have an effect!

Also, I think that this PKEY_REFERENCING_OPTIONS is normally made to contain other info too, which we might want to retain. That's why it's called OPT_**BIT**_AUTO_ZERO. We should just set/unset this bit (3rd bit). With the constant OPT_BIT_NORMAL = 0, it's confusing. From the developer's doc, it seems bits 1,2,3 can be combined. The value OPT_BIT_NORMAL is just there as a helper constant to reset the behaviour.

So I'd suggest to do three changes.
* Firstly, just have a single function `set_reference_autozero(channel, active: bool)`. It reads the current referencing option,  set/unset the bit SA_CTL_REF_OPT_BIT_AUTO_ZERO (based on active flag), and store back the value.
* Secondly, after a referencing, *always* reset to non-auto-zero mode.
* Thirdly, change the smaract MCS2 driver to not change the referencing options, ever. For extra safety, you could read the option at init, and log them + log a warning if AUTO_ZERO is set.

pieleric (changes requested):

util/saconfig
```diff
@@ -61,74 +62,102 @@ def calibrate(dev, channels):
...
 
 
+def reference(dev, channels):
+    # Run the referencing
+    logging.info("Starting reference...")
+    dev.reference(axes=set([str(ch) for ch in channels])).result()
+    time.sleep(0.1)
+    logging.info("Referencing complete.")
+    # Always set reference to normal after referencing
+    set_reference_normal(dev, channels)
```
lazem, line 116: Is that what you had in mind? @pieleric  `reference()` is called from `set_reference_autozero()` which means the reference mode will return back to normal.

src/odemis/driver/smaract.py
```diff
@@ -2549,6 +2549,17 @@ def __init__(self, name, role, locator, ref_on_init=False, axes=None, speed=1e-3
             self._set_accel(channel, accel)
             self._set_hold_time(channel, hold_time)
 
+            try:
+                # Log referencing mode, and warn if it's not normal
+                ref_mode = self.GetProperty_i32(SA_CTLDLL.SA_CTL_PKEY_POSITIONER_TYPE, channel)
+                log_lvl = logging.INFO
+                if ref_mode != SA_CTLDLL.SA_CTL_REF_OPT_BIT_NORMAL:
```
pieleric, line 2556: BIT_NORMAL is a shortcut for 0. However, there are some bits which are completely fine. The one to pay attention to is `SA_CTL_REF_OPT_BIT_AUTO_ZERO`.
=> `if ref_mode & SA_CTLDLL.SA_CTL_REF_OPT_BIT_AUTO_ZERO:`

util/saconfig
pieleric: Not exactly. Right now it's confusing because set_reference_autozero() does the referencing, while set_reference_normal() just change the flag. We end up with set_reference_autozero() calling indirectly set_reference_normal().
I'd suggest to add a flag to `reference():` `autozero=False`. For each channel, it would set/unset the autozero bit. Note that it should just change that bit, not change the whole value. You need to use bitwise and/or! So it could rely on a helper subfunction set_reference_autozero(autozero) to do it. Something like this:
```python
def reference(dev, channels, autozero=False):
    for channel in channels:
         set_reference_autozero(autozero)
    # Run the referencing
    logging.info("Starting reference...")
    try:
        dev.reference(axes=set(str(ch) for ch in channels)).result()
        time.sleep(0.1)
        logging.info("Referencing complete.")
    finally:
        # Always set reference to normal after referencing
    for channel in channels:
         set_reference_autozero(False)


def set_reference_autozero(dev, channel, autozero):
    logging.info(f"Setting the reference autozero mode to {autozero} for channel {channel}")
    ref_opt = dev.GetProperty_i32(smaract.SA_CTLDLL.SA_CTL_PKEY_REFERENCING_OPTIONS, channel)
    if autozero:
        ref_opt |= smaract.SA_CTLDLL.SA_CTL_REF_OPT_BIT_AUTO_ZERO
    else:
        ref_opt &= ~smaract.SA_CTLDLL.SA_CTL_REF_OPT_BIT_AUTO_ZERO

```

util/saconfig
```diff
@@ -61,74 +62,102 @@ def calibrate(dev, channels):
...
 
     parser.add_argument("--log-level", dest="loglev", metavar="<level>", type=int,
                         default=1, help="set verbosity level (0-2, default = 1)")
 
     parser.add_argument('--calibrate', metavar='N', type=int, nargs='+', default=[],
-                    help="Run the calibration command. Specify the channel to calibrate (0, 1, 2)")
+                        help="Run the calibration command. Specify the channel to calibrate (0, 1, 2)")
+
+    parser.add_argument('--postype', metavar='N', type=int, nargs='+', default=[],
+                        help="Set positioner type to distance coded mark. Specify the channel to calibrate (0, 1, 2)")
```
pieleric, line 151: "Specify the channel to calibrate" -> "Specify the channel to change"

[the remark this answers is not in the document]
pieleric: I think that if we want it more refined, we should go with something like the tmcmconfig, which reads/write values from a tsv file. For now, this should be good enough.

pieleric (changes requested):

lazem:
Thanks @pieleric for explaining setting the autozero bit!

PR #1708, opened 2021-03-19
add the class LocalizationGUIData 

Mahmood-B:


src/odemis/gui/cont/tabs.py
```diff
@@ -404,7 +405,7 @@ def __init__(self, name, button, panel, main_frame, main_data):
             static=True,
         )
 
-        self._overview_stream_controller.add_overview_action(self._on_acquire)
+        # self._overview_stream_controller.add_overview_action(self._on_acquire)
```
lazem, line 408: Please uncomment this.

src/odemis/gui/model/__init__.py
```diff
@@ -595,6 +595,38 @@ def __init__(self, main):
...
+        self.autofocus_active = BooleanVA(False)
+
+        self.Zmin = model.FloatContinuous(
+            value=-10e-6, range=(-1000e-6, 0), unit="m")
+        self.Zmax = model.FloatContinuous(
+            value=10e-6, range=(0, 1000e-6), unit="m")
+        self.Zstep = model.FloatContinuous(
+            value=1e-6, range=(-100e-6, 100e-6), unit="m")
+        self.zStackActive = model.BooleanVA(value=False)
+        self.acquisitionStreams = model.ListVA()
```
lazem, line 627: Maybe add a comment on what these parameters represent?

lazem (changes requested):

src/odemis/gui/model/__init__.py
```diff
@@ -595,6 +595,38 @@ def __init__(self, main):
...
+
+        # VA for autofocus procedure mode
+        self.autofocus_active = BooleanVA(False)
+
+        self.Zmin = model.FloatContinuous(
+            value=-10e-6, range=(-1000e-6, 0), unit="m")
+        self.Zmax = model.FloatContinuous(
+            value=10e-6, range=(0, 1000e-6), unit="m")
+        self.Zstep = model.FloatContinuous(
+            value=1e-6, range=(-100e-6, 100e-6), unit="m")
```
pieleric, line 625: The "tradition" is to name the VAs like "firstSecond". We already have "zPos". So we should name these attributes: zMin, zMax, zStep.

src/odemis/gui/model/__init__.py
pieleric: That's right. For the z attributes, you can have a single comment "Z-Stack acquisition parameters".
For the acquisition streams: "the streams to use in the acquisition, among all the .streams"

pieleric (changes requested):

src/odemis/gui/model/__init__.py
```diff
@@ -595,6 +595,40 @@ def __init__(self, main):
...
+class LocalizationGUIData(MicroscopyGUIData):
+    # TODO: update the doc string
+    """ Represent an interface used to only show the current data from the microscope.
+
+    It should be able to handle SEM-only, optical-only, SECOM and DELPHI systems.
+
+    """
+
+    def __init__(self, main):
+        assert main.microscope is not None
```
Kleijwegt, line 608: Maybe it is nice to add a message if the assertion fails so that the user of the class knows what went wrong. Something like:
> assert main.microscope is not None, "Class .... can only be instantiated with the main.microscope is defined"

src/odemis/gui/model/__init__.py
```diff
@@ -595,6 +595,40 @@ def __init__(self, main):
         # VA for autofocus procedure mode
         self.autofocus_active = BooleanVA(False)
 
+
+class LocalizationGUIData(MicroscopyGUIData):
+    # TODO: update the doc string
+    """ Represent an interface used to only show the current data from the microscope.
+
+    It should be able to handle SEM-only, optical-only, SECOM and DELPHI systems.
```
pieleric, line 603: Please do this TODO now. (And drop the `# TODO...`).
You can say it's for the CryoSECOM.

src/odemis/gui/model/__init__.py
pieleric: To be fair to Mahmoud, this is just copy-paste from the other similar classes.

The reason it's like this on the other classes is that I'm personally not a fan of adding such a message to `assert`. I feel that if you really want to care about some states, then you should use a proper if + Exception. I like to use `assert` as a very short reminder for the reader about the code's expectations.

pieleric (changes requested):

src/odemis/gui/model/__init__.py
Mahmood-B: Do you mean if + raise? 
What kind of error should be raised in the situation of missing/unconnected component?

pieleric: Yes, if + raise ValueError().

src/odemis/gui/model/__init__.py
```diff
@@ -595,6 +597,46 @@ def __init__(self, main):
...
+class LocalizationGUIData(MicroscopyGUIData):
+    """ Represent an interface used to only show the current data from the microscope.
+
+    It it used for handling CryoSECOM systems.
+
+    """
+
+    def __init__(self, main):
+        if main.role != "cryo-secom":
+            raise ValueError("The role of the microscope is not correct")
```
lazem, line 610: The original assert was for checking there is a microscope object in main, not the role is cryo-secom. So it should be:
`if not hasattr(main, 'microscope'):`
`raise ValueError`

src/odemis/gui/model/__init__.py
```diff
@@ -595,6 +597,46 @@ def __init__(self, main):
...
+
+        # Current tool selected (from the toolbar)
+        tools = {TOOL_NONE, TOOL_RULER}  # TOOL_ZOOM, TOOL_ROI}
+        # Update the tool selection with the new tool list
+        self.tool.choices = tools
+        # VA for autofocus procedure mode
+        self.autofocus_active = BooleanVA(False)
+        # the relative focus position below the current position
+        self.zMin = model.FloatContinuous(
+            value=-10e-6, range=(-1000e-6, 0), unit="m")
```
lazem, line 621: I believe the comment should be the zstack minimum range below current focus position.

src/odemis/gui/model/__init__.py
```diff
@@ -595,6 +597,46 @@ def __init__(self, main):
...
+        # the relative focus position above the current position
+        self.zMax = model.FloatContinuous(
+            value=10e-6, range=(0, 1000e-6), unit="m")
```
lazem, line 624: zstack maximum  range above current focus position.

lazem (changes requested):

src/odemis/gui/model/__init__.py
Mahmood-B: Yes it was checking if there microscope exists. But the passed argument, main, is of type MainGUIData. At the init function of this class, there are several checks on the components of the microscope such as emitters and detectors. If any of these components does not exist, then an error is raised. So there is no point from checking if there is microscope object. It would make more sense to check if the role of this microscope object is cryo secom since this class is only for cryo secom (if i'm not wrong)

src/odemis/gui/model/__init__.py
Mahmood-B: maximum or minimum _range_ indicates interval. But zMax and zMin are not intervals. They are scalars, and they are measured relative to the current focus position.

src/odemis/gui/model/__init__.py
lazem: I see what you mean, but you won't know if these microscope components are indeed there by checking its role. Besides it'd be redundant as checking for the role is already done on the tab itself in `get_display_priority`.

src/odemis/gui/model/__init__.py
lazem: In the specification: 

> Zmin and Zmax define the range below and above the current focus position (Z) to acquire.

Both define the min and max values for the range of acquisition. The range is indeed an interval where we'll take images by zsteps from the minimum value till we reach the maximum.

lazem (changes requested):

src/odemis/gui/model/__init__.py
pieleric: First, to be exact, main also has a microscope. It's sometimes `None`. So it'd be:
`assert main.microscope is not None`
This is me would suggested Mahmoud to change it this way. Typically, if microscope is None, then the role is None too. So this check is normally just a little be extra strict.
@lazem , I don't agree with your argument that this is not needed because we already check for it. If that was valid, then all the checks are unnecessary! Of course, right now the code works, all the callers are well behaved. The checks are here for two other reasons:
* Remind the reader about the expectation of the code
* Detect issues in new code, early

So I think this is fine. Mahmoud could add back the original assert, just to make things even clearer.
The only thing that I thing could be improved is the error message. It should include the actual role. Something like `"Microscope role is %s, while expected 'cryo-secom'"`

pieleric (approved):

lazem (approved):

PR #1709, opened 2021-03-23
Updated spot-grid.py script

effting:
- Supports to load image from command line
- Indicator of first spot in the image
- 'Snake' to indicate order of spots

### DIFF
--- a/scripts/spot-grid.py
+++ b/scripts/spot-grid.py
@@ -14,6 +14,7 @@ import argparse
 import cairo
 import logging
 import numpy
+import skimage.io
 import sys
 import threading
 import wx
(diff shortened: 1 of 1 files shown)


scripts/spot-grid.py
```diff
@@ -168,6 +191,27 @@ def image_update(imp, window):
         logging.debug("Display thread ended")
 
 
+class StaticImageComponent(model.DigitalCamera):
+
+    def __init__(self, filename):
```
pieleric, line 196: Please take a DataArray as input. Let's separate reading away.

scripts/spot-grid.py
```diff
@@ -168,6 +191,27 @@ def image_update(imp, window):
...
+        super(StaticImageComponent, self).__init__(name=filename, role='ccd')
```
pieleric, line 197: As name, just hard code to something like: "FakeCCD".

scripts/spot-grid.py
```diff
@@ -224,7 +271,11 @@ def main(args):
     handler.setFormatter(logging.Formatter('%(asctime)s (%(module)s) %(levelname)s: %(message)s'))
     logging.getLogger().addHandler(handler)
 
-    if options.role:
+    if options.filename:
+        logging.info("Will process image file %s" % options.filename)
+        image = StaticImageComponent(options.filename)
```
pieleric, line 276: Let's rename image -> fakeccd.

scripts/spot-grid.py
```diff
@@ -224,7 +271,11 @@ def main(args):
     handler.setFormatter(logging.Formatter('%(asctime)s (%(module)s) %(levelname)s: %(message)s'))
     logging.getLogger().addHandler(handler)
 
-    if options.role:
+    if options.filename:
+        logging.info("Will process image file %s" % options.filename)
```
pieleric, line 275: To open the file, you can do it without skimage, just with what we already have in Odemis. You can write:
``` python
    converter = dataio.find_fittest_converter(fn, default=None, mode=os.O_RDONLY)
    data = converter.read_data(filename)
    image = data[0]
```

scripts/spot-grid.py
```diff
@@ -168,6 +191,27 @@ def image_update(imp, window):
         logging.debug("Display thread ended")
 
 
+class StaticImageComponent(model.DigitalCamera):
```
pieleric, line 194: Can you rename it to StaticCCD?

scripts/spot-grid.py
```diff
@@ -14,6 +14,7 @@
 import cairo
 import logging
 import numpy
+import skimage.io
```
pieleric, line 17: So far we don't depend on skimage. Let's try to not depend on it. If you need more than TIFF + HDF5, you could have a look at Pillow (which we already depend on).

pieleric (changes requested):

pieleric (approved):

tepals (approved):

PR #1710, opened 2021-03-23
driver picoquant: add support for HydraHarp 400

pieleric:
Provided by Eric Liu and Jacob Ng from ORNL.

### DIFF
--- a/src/odemis/driver/picoquant.py
+++ b/src/odemis/driver/picoquant.py
@@ -1,10 +1,15 @@
 # -*- coding: utf-8 -*-
-'''
+"""
 Created on 14 Apr 2016
 
 @author: Éric Piel
 
-Copyright © 2016 Éric Piel, Delmic
+Edited on 9 Nov 2020
+
+@editor: Eric Liu
+@editor: Jacob Ng
+
+Copyright © 2016-2021 Éric Piel, Delmic
 
 This file is part of Odemis.
 
@@ -13,7 +18,10 @@ Odemis is free software: you can redistribute it and/or modify it under the term
 Odemis is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
 
 You should have received a copy of the GNU General Public License along with Odemis. If not, see http://www.gnu.org/licenses/.
-'''
+"""
+# Support for the PicoQuant time-correlators: PicoHarp 300 and HydraHarp 400.
+# Both are connected via USB. They require a dedicated library to be installed
+# on the computer, libph300.so or libhh400.so .
 
 from __future__ import division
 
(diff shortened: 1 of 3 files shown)


src/odemis/driver/picoquant.py
```diff
@@ -91,6 +167,17 @@ def __str__(self):
         return self.args[1]
 
 
+class HHError(Exception):
+    def __init__(self, errno, strerror, *args, **kwargs):
```
Kleijwegt, line 171: Can you add a description when/why a HHError should be raised?

src/odemis/driver/picoquant.py
```diff
@@ -32,52 +37,123 @@
...
+HH_ACQTMAX = 360000000  # ms  (100*60*60*1000ms = 100h)
+
+HH_STOPCNTMIN = 1
+HH_STOPCNTMAX = 4294967295  # 32 bit is mem max
+
+HH_HOLDOFFMIN = 0  # ns
+HH_HOLDOFFMAX = 524296  # ns
 
 
 class PHError(Exception):
```
Kleijwegt, line 159: What is a PHError, is that when the pH (https://en.wikipedia.org/wiki/PH?oldformat=true) is to low/high?
Maybe an description when/why such an error is raised can be added?

src/odemis/driver/picoquant.py
Kleijwegt: Probably it is when the HydraHarp fails (because of the PR title) but that won't be in the file so just one sentence for what fail case this error is defined is enough I think.

src/odemis/driver/picoquant.py
```diff
@@ -1055,8 +2393,8 @@ def PH_CTCStatus(self, i, p_ctcstatus):
         if self._acq_end > time.time():
             ctcstatus.value = 0  # 0 if still running
             # DEBUG
-#             if random.randint(0, 10) == 0:
-#                 raise PHError(-37, PHDLL.err_code[-37])  # bad luck
+            # if random.randint(0, 10) == 0:
+            #    raise PHError(-37, PHDLL.err_code[-37])  # bad luck
```
Kleijwegt, line 2397: Where these lines supposed to stay in this PR?

src/odemis/driver/picoquant.py
```diff
@@ -1069,7 +2407,184 @@ def PH_GetElapsedMeasTime(self, i, p_elapsed):
...
+    def HH_GetModuleIndex(self, i, channel, modidx):
+        # Not needed?
+        pass
+
+    def HH_GetHardwareDebugInfo(self, i, debuginfo):
+        debuginfo.value = b"Fake hardware debug info"
+        # Not needed?
+
+    def HH_Calibrate(self, i):
+        pass
```
Kleijwegt, line 2518: Shouldn't be there some type of logging saying "not implemented function nameOfFunction is called" ?

src/odemis/driver/picoquant.py
```diff
@@ -453,9 +695,9 @@ def SetInputCFD(self, channel, level, zc):
         level (int) CFD discriminator level in millivolts
         zc (0<=int): CFD zero cross in millivolts
         """
-        assert(channel in {0, 1})
-        assert(DISCRMIN <= level <= DISCRMAX)
-        assert(ZCMIN <= zc <= ZCMAX)
+        assert channel in {0, 1}
+        assert PH_DISCRMIN <= level <= PH_DISCRMAX
+        assert PH_ZCMIN <= zc <= PH_ZCMAX
```
Kleijwegt, line 700: Shouldn't asserts like these have a logging message in case they fail? Just so that if a retry isn't succeding multiple times or some setting isn't within limits and therfore you eventually get into a timeout or other error you can find the trace/origin in the log.
There are quite some asserts in this PR but I don't see many logging messages (or custom error) accompanying them.

src/odemis/driver/picoquant.py
pieleric: Yes, that's extra code in case we want to test the error handling when the device communication breaks.

src/odemis/driver/picoquant.py
pieleric: I'm really not a fan of extra message in asserts. One of the main purpose of the asserts is to give some context to the reader. The short asserts allow to keep the line very simple and readable. The other purpose to stop early possible errors. On such case, the log will already indicate the line where it fails and all the backtrace. So for instance here you'd know that channel is not in {0,1}. Arguably, it'd be even nicer if we knew what the channel was, but I don't think the time and risk to make a mistake in the message is worthy. The chances that such assert trigger are very low.

PR #1711, opened 2021-03-24
MultibeamDetector added to support image acquistion through…

tepals:
… XTToolkit.


src/odemis/driver/xt_client.py
```diff
@@ -2215,3 +2237,74 @@ def _setMultiBeamMode(self, multi_beam_mode):
...
+
+        Detector.__init__(self, name, role, parent=parent, channel_name=channel_name, **kwargs)
+
+    def stop_generate(self):
+        logging.debug("Stopping image acquisition")
+        # FIXME there has to be a better way to do this :(
+        # This doesn't work:
+        # self.parent.set_channel_state(self._channel_name, False)
+        server = Pyro5.api.Proxy("PYRO:Microscope@192.168.31.158:4242")
+        server.set_channel_state(self._channel_name, False)
```
tepals, line 2267: @pieleric I tried stopping the scan by using `self.parent.set_channel_state(self._channel_name, False)`. This just waited for the scan_image call to finish before making the call. 

It has something to do with the `threading.Lock()` we use in the calls to the adapter. Because the `scan_image` call uses the `threading.Lock()` I cannot also run the `set_channel_state` call. Can I somehow unlock the Lock?

As shown on line 2266 and 2267 if I create a new server object I can stop the channel, and stop the scan image call. However this is definitely very hacky.

tepals:
Still a work in progress, Éric can you have a look at my comment?

src/odemis/driver/xt_client.py
```diff
@@ -2215,3 +2237,74 @@ def _setMultiBeamMode(self, multi_beam_mode):
...
+        Managed via the ._genmsg Queue
+        """
+        try:
+            while True:
+                # Wait until we have a start (or terminate) message
+                self._acq_wait_start()
+                logging.debug("Preparing acquisition")
+                while True:
+                    if self._acq_should_stop():
+                        break
```
Kleijwegt, line 2284: Can this not be put into one line? like:
> While not self._acq_should_stop():

src/odemis/driver/xt_client.py
```diff
@@ -2215,3 +2237,74 @@ def _setMultiBeamMode(self, multi_beam_mode):
...
+                    image = self.parent.get_latest_image(self._channel_name)
+
+                    md = self.parent._scanner._metadata.copy()
+                    md.update(self._metadata)
+                    md[model.MD_DWELL_TIME] = self.parent._scanner.dwellTime.value
+                    md[model.MD_ROTATION] = self.parent._scanner.rotation.value
+
+                    da = DataArray(image, md)
+                    logging.debug("Notify dataflow with new image.")
+                    self.data.notify(da)
```
Kleijwegt, line 2303: Then this can also go into one if statement like:

> if self._acq_should_stop():
>                     # Acquire the image
>                     # ======================================
>                     image = self.parent.get_latest_image(self._channel_name)

I don't particularly like the break statements in this context.

src/odemis/driver/xt_client.py
pieleric: That's one of the main difference between Pyro4 and Pyro5. Pyro4 is able to automatically create a new connection, if the current one is in use.
Definitely, you cannot just "unlock the lock". You could just not use the Lock at all, and use the same connection, but then mayhem would happen ;-)

I'd suggest to create multiple connections, as you have started to do. Here, it's a little hacky because you create the connection on the fly. But you could create a dedicated "cancel_connection", with its own lock, and use it only for cancelling. If we ever need more than 2 connections, probably we'll need to implement something fancy as in Pyro4 which allow you to get the "first connection not in use". Let's hope we don't need to go there though....

src/odemis/driver/xt_client.py
tepals: This is now the same as in the regular Detector._acquire function, so I'd rather leave it like this.

tepals:
I've updated the PR. I tested it using the VM simulator, I will test it on a real system on Friday.
@pieleric can you do another review?

src/odemis/driver/xt_client.py
```diff
@@ -1010,6 +1017,21 @@ def beamlet_index_info(self):
             self.server._pyroClaimOwnership()
             return self.server.beamlet_index_info()
 
+    def scan_image(self, channel_name='electron1'):
+        """
+        Start the scan of a single image and block until the image is done scanning.
+        This function does not return the image, to get the image call get_latest_image after this function.
+
```
pieleric, line 1024: Does it work also on xtlib, or only on xttoolkit? If only on xttoolkit, it'd be worth mentioning.

src/odemis/driver/xt_client.py
```diff
@@ -2215,3 +2237,90 @@ def _setMultiBeamMode(self, multi_beam_mode):
...
+        """
+        channel_name (str): Name of one of the electron channels.
+        """
+        # The acquisition is based on a FSM that roughly looks like this:
+        # Event\State |    Stopped    |   Acquiring    | Receiving data |
+        #    START    | Ready for acq |        .       |       .        |
+        #    DATA     |       .       | Receiving data |       .        |
+        #    STOP     |       .       |     Stopped    |    Stopped     |
+        #    TERM     |     Final     |      Final     |     Final      |
+        self._proxy_access = threading.Lock()
```
pieleric, line 2258: This lock is to protect "cancel_connection", so name it more similarly. Like `_cancel_access` or `_cancel_connection_lock`...

src/odemis/driver/xt_client.py
```diff
@@ -2215,3 +2237,90 @@ def _setMultiBeamMode(self, multi_beam_mode):
...
+        try:
+            self.cancel_connection = Pyro5.api.Proxy(address)
```
pieleric, line 2260: Please add a tiny bit of explanation why you need a second connection.

src/odemis/driver/xt_client.py
```diff
@@ -2215,3 +2237,90 @@ def _setMultiBeamMode(self, multi_beam_mode):
...
+                          "uri is correct and XT server is"
+                          " connected to the network. %s" % (address, err))
+        Detector.__init__(self, name, role, parent=parent, channel_name=channel_name, **kwargs)
+        self._shape = (2**16,)  # Depth of the image
+
+    def stop_generate(self):
+        logging.debug("Stopping image acquisition")
+        with self._proxy_access:
+            self.cancel_connection._pyroClaimOwnership()
+            # Directly calling set_channel_state does not work with XT. Therefore
```
pieleric, line 2273: "XT" -> "XTToolKit"?

src/odemis/driver/xt_client.py
```diff
@@ -2215,3 +2237,90 @@ def _setMultiBeamMode(self, multi_beam_mode):
...
+                logging.debug("Preparing acquisition")
+                while True:
+                    if self._acq_should_stop():
+                        break
+
+                    # Start a complete scan.
+                    try:
+                        self.parent.scan_image()
+                    except OSError as err:
+                        if err.errno == -2147467260:
```
pieleric, line 2301: Please translate. What does this errno correspond to?

src/odemis/driver/xt_client.py
```diff
@@ -2215,3 +2237,90 @@ def _setMultiBeamMode(self, multi_beam_mode):
...
+                    # =================
+                    image = self.parent.get_latest_image(self._channel_name)
+
+                    md = self.parent._scanner._metadata.copy()
+                    md.update(self._metadata)
+                    md[model.MD_DWELL_TIME] = self.parent._scanner.dwellTime.value
+                    md[model.MD_ROTATION] = self.parent._scanner.rotation.value
+
+                    da = DataArray(image, md)
+                    logging.debug("Notify dataflow with new image.")
```
pieleric, line 2318: Could you add the shape of the image in the log? It's cheap, and can give a lot of clues if something goes wrong.

src/odemis/driver/xt_client.py
```diff
@@ -2215,3 +2237,90 @@ def _setMultiBeamMode(self, multi_beam_mode):
...
+                    self.data.notify(da)
+            logging.debug("Acquisition stopped")
+        except TerminationRequested:
+            logging.debug("Acquisition thread requested to terminate")
+        except Exception as err:
+            logging.exception("Failure in acquisition thread: {}".format(err))
```
pieleric, line 2324: In logging, please just pass the values as argument `"Failure in acquisition thread: %s", err`. Mostly to keep the code homogeneous.

pieleric (changes requested):

tepals:
@pieleric I processed your comments and rebased on the latest master.

pieleric (approved):

pieleric (changes requested):
Sorry, I only realise now what you had already mentioned about the name, a few days ago. I got confused between the mb-scanner and, here, the mb-detector. Indeed, it doesn't make sense to "mirror" the mb-scanner behaviour. 
The idea of explicitly naming "mb-scanner" is that the user *expects* more from it than a normal "scanner": to control the multibeam attributes.

That's not the case for the detector. We had a class working with the xtlib, and we just want one that also works with xttoolkit, but the user shouldn't have to care what's the exact xt_type. So I'd suggest to just handle one "detector", and depending on the actual xt_type, the init code automatically picks the right class to initialize.

The class should also not be named `MultiBeamDetector`, but more something in line with the fact it's to support XTToolKit, like `XTLDetector`, or `XTToolKitDetector.`

tepals:
No problem. I updated the code, and went for the name `XTTKDetector`.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -1179,5 +1179,104 @@ def test_beam_powerVA(self):
         self.scanner.power.value = init_beam_power
 
 
+class TestXTTKDetector(unittest.TestCase):
+    """
+    Test image acquisition using the XTToolkit Detector class.
+    """
```
pieleric, line 1185: Aren't all these tests already done in `TestMicroscope` (now that there no special "md-detector")? 
I quickly search for differences, and couldn't see any. From a theoretical point of view, I think that the test for the XTToolKit and the XTLib should be the same, as the code is supposed to behave the same.

pieleric (changes requested):

src/odemis/driver/test/xt_client_test.py
tepals: Good catch, yes they're all the same. I'll remove this class.

PR #1713, opened 2021-03-31
xtclient driver adjusted delta pitch to add the new value to the old …

tepals:
…delta pitch.

With this implementation changing the delta pitch VA actually adds the difference to the current value instead of the difference to the initial value.

### DIFF
--- a/src/odemis/driver/test/xt_client_test.py
+++ b/src/odemis/driver/test/xt_client_test.py
@@ -1014,26 +1014,28 @@ class TestMBScanner(unittest.TestCase):
             self.assertIsNot(type(child), xt_client.Scanner)
 
     def test_delta_pitch_VA(self):
-        current_value = self.scanner.deltaPitch.value
+        init_value = self.scanner.deltaPitch.value
         # Test if directly changing the value via the VA works. Not always will the entirety of the range be
         # allowed. Negative delta pitch is limited by the voltage it can apply. Therefore the max range and the 0
         # value is tested.
-        for test_delta_pitch in (0.0, self.scanner.deltaPitch.range[1]):
+        current_value = init_value
+        for test_delta_pitch in (0.0, 0.5e-6, self.scanner.deltaPitch.range[1]):
             self.scanner.deltaPitch.value = test_delta_pitch
-            self.assertEqual(test_delta_pitch, self.scanner.deltaPitch.value)
-            self.assertEqual(test_delta_pitch, self.microscope.get_delta_pitch() * 1e-6)
+            current_value += test_delta_pitch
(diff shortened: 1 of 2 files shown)


src/odemis/driver/test/xt_client_test.py
```diff
@@ -1014,26 +1014,28 @@ def test_type_scanner_child(self):
...
         # Test if directly changing the value via the VA works. Not always will the entirety of the range be
         # allowed. Negative delta pitch is limited by the voltage it can apply. Therefore the max range and the 0
         # value is tested.
-        for test_delta_pitch in (0.0, self.scanner.deltaPitch.range[1]):
+        current_value = init_value
+        for test_delta_pitch in (0.0, 0.5e-6, self.scanner.deltaPitch.range[1]):
             self.scanner.deltaPitch.value = test_delta_pitch
-            self.assertEqual(test_delta_pitch, self.scanner.deltaPitch.value)
-            self.assertEqual(test_delta_pitch, self.microscope.get_delta_pitch() * 1e-6)
+            current_value += test_delta_pitch
```
Sabrina85, line 1024: If the current value was 0.5e-6 and you add the self.scanner.deltaPitch.value[1] (so the largest value allowed), it will raise an error, won't it?

src/odemis/driver/xt_client.py
```diff
@@ -2177,7 +2177,10 @@ def _updateSettings(self):
             logging.exception("Unexpected failure when polling XTtoolkit settings")
 
     def _setDeltaPitch(self, delta_pitch):
-        self.parent.set_delta_pitch(delta_pitch * 1e6)  # Convert from meters to micrometers.
+        """Add the new delta pitch to the current delta pitch value and set it on the microscope."""
+        current_delta_pitch = self.deltaPitch.value
+        new_pitch = (current_delta_pitch + delta_pitch) * 1e6  # Convert from meters to micrometers.
```
Sabrina85, line 2182: What happens if the new value is out-of-range? Is that handled?

src/odemis/driver/test/xt_client_test.py
Kleijwegt: That is correct, it would go at of the allowed range.

pieleric (changes requested):
No, we should really never have such behaviour. You're changing the code so that "=" means "+=". It's very confusing. We need to adjust this code, so that it calls the xt_adapter the right way in order that a "=" means "=". Possibly, get the current value first, and pass the difference between the current value and the requested value.

tepals:
Closed for now, until we know what we want to do.

PR #1715, opened 2021-03-31
ENZEL: Fix disappearing live streams on acquisition dialog 

lazem:
Localization tab `clear_data` used the overview stream controller's `clear` function to remove the 'old' streams from the views, but that resulted in removing them from the acquisition dialog as well. To fix this, `removeStream` is used instead.

Note that the old static streams still show on the panel, and it could be required to remove them from it to alleviate memory consumption.

### DIFF
--- a/src/odemis/gui/cont/tabs.py
+++ b/src/odemis/gui/cont/tabs.py
@@ -70,7 +70,7 @@ from odemis.acq.stream import OpticalStream, SpectrumStream, TemporalSpectrumStr
     ScannedTCSettingsStream, SinglePointSpectrumProjection, LineSpectrumProjection, \
     PixelTemporalSpectrumProjection, SinglePointTemporalProjection, \
     ScannedTemporalSettingsStream, \
-    ARRawProjection, ARPolarimetryProjection, StaticStream
+    ARRawProjection, ARPolarimetryProjection, StaticStream, LiveStream
 from odemis.acq.move import LOADING, IMAGING, MILLING, COATING, UNKNOWN, LOADING_PATH, target_pos_str
 from odemis.acq.move import cryoSwitchSamplePosition, cryoTiltSample, getMovementProgress, getCurrentPositionLabel
 from odemis.util.units import decompose_si_prefix, readable_str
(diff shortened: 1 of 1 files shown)


src/odemis/gui/cont/tabs.py
```diff
@@ -508,20 +508,30 @@ def load_data(self, data):
...
         """
         Clear the tab data upon resetting the project:
         - Clear overview map streams
-        - Clear live streams data
-        """
-        self._overview_stream_controller.clear()
+        @:param new_stream: (StaticStream or None) the newly acquired overview stream
+        """
+        # Clear old static streams from the view
+        # TODO: Remove old streams from the panel as well, currently it's hidden
```
pieleric, line 523: Can you explain the issue more? I don't get what's still happening wrong.

src/odemis/gui/cont/tabs.py
```diff
@@ -508,20 +508,30 @@ def load_data(self, data):
...
+    def clear_data(self, new_stream=None):
         """
         Clear the tab data upon resetting the project:
         - Clear overview map streams
-        - Clear live streams data
-        """
-        self._overview_stream_controller.clear()
+        @:param new_stream: (StaticStream or None) the newly acquired overview stream
+        """
+        # Clear old static streams from the view
```
pieleric, line 522: I think that the interface is fine, but the naming is confusing. I find it hard to guess what will happen when calling `clear_data(my_shiny_new_stream)`. It feels like it's going to remove that new stream. 
I'd suggest something like `clear_static_data(filter=None)`. `filter` could be a "similar stream", but I'd go for an even simpler behaviour and just pass a string: the name of the stream(s) to remove. However, `filter` is not a great name, as it's a builtin of Python. Also, when changing of selected Feature, we'll need a similar behaviour which removes all the static streams from the previous feature. So maybe we need to extend the signature to handle such behaviour, or alternatively we'd have a `clear_overview_data(...)` and a `clear_acquired_data(...)`.
In the docstring, please indicate what happens when passing `None`.

Also, we should have a counter-part `clear_live_data()`, which takes care of resetting the live streams when a new project is started. It should just empty the content of the live streams (so that the last image of the previous sample are not shown at the beginning)

src/odemis/gui/cont/tabs.py
```diff
@@ -508,20 +508,30 @@ def load_data(self, data):
...
+        # TODO: Remove old streams from the panel as well, currently it's hidden
+        if new_stream:
+            # Replace only the stream matching the new stream in acquisition type and name
+            old_stream = next((stream for stream in self.tab_data_model.streams.value if isinstance(stream, StaticStream)
+                               and stream.acquisitionType.value == new_stream.acquisitionType.value and stream.name.value == new_stream.name.value), None)
+            if old_stream:
+                self._overview_stream_controller.removeStream(old_stream)
```
pieleric, line 529: You could make it a little simpler, by merging it into the "all streams" code. Assuming that you just are filtering on the name, it'd look like this (but if you pass the stream, it's very similar):
```python
static_streams = [stream for stream in self.tab_data_model.streams.value if isinstance(stream, StaticStream)]
for stream in static_streams:
    if filter and stream.name.value != filter:
        continue
    self._overview_stream_controller.removeStream(stream)
```
Note that it'd remove all the streams matching the filter. I think this is fine. Typically, you'd have only one stream with the same name.

pieleric (changes requested):
I think that part of the issue here is that we have 2 StreamBarContollers, one for the live streams, one for the static streams. However, the StreamBarController class assumes that it's the only one on the tab. So it has a tendency to take some shortcuts. We might need to look at this more closely, especially once we handle the display of the static streams associated to the selected Feature.

src/odemis/gui/cont/tabs.py
lazem: What's happening wrong is that the old image is removed from the view, but it still stays on the panel hidden (see attached image)
Screenshot from 2021-04-09 15-34-57 png

src/odemis/gui/cont/tabs.py
```diff
@@ -508,20 +508,31 @@ def load_data(self, data):
         # Create streams from data
         streams = data_to_static_streams(data)
 
-        # TODO: Clear previous overview streams
-        self.clear_data()
-
         for s in streams:
+            self.clear_overview_data(s)
```
pieleric, line 512: It should pass the stream **name**, right? Then `s.name.value`.

src/odemis/gui/cont/tabs.py
```diff
@@ -508,20 +508,31 @@ def load_data(self, data):
...
             scont.stream_panel.show_remove_btn(True)
 
-    def clear_data(self):
+    def clear_live_streams(self):
+        """
+        Clear the content of the live streams
+        """
+        live_streams = [stream for stream in self.tab_data_model.streams.value if isinstance(stream, LiveStream)]
+        for stream in live_streams:
+            stream.image.value = None
```
pieleric, line 522: That's not enough to reset the data. For instance, if you change the brightness/contrast, the stream will update the .image, and it'll be back from the dead!
See how to do it in gui.cont.microscope._reset_streams(). Essentially:
```python
            if s.raw:
                s.raw = []
                s.image.value = None
                s.histogram._value = numpy.empty(0)
                s.histogram.notify(s.histogram._value)
```

pieleric (changes requested):

lazem:
Thanks for the prompt review! @pieleric

PR #1717, opened 2021-03-31
Enable Cryo tab buttons on imaging only

lazem:
Previously `enable_tab_on_stage_position` was taking the tab as parameter and enable/disable its panel on imaging position. But what was actually required is to enable/disable the tab buttons.

### DIFF
--- a/src/odemis/gui/cont/tabs.py
+++ b/src/odemis/gui/cont/tabs.py
@@ -641,13 +641,6 @@ class LocalizationTab(Tab):
         assert (show != self.IsShown())  # we assume it's only called when changed
         super(LocalizationTab, self).Show(show)
 
-        # Enable the tab if the stage is in imaging position, disable it otherwise
-        stage = self.tab_data_model.main.stage
-        if getCurrentPositionLabel(stage.position.value, stage) is IMAGING:
-            self.panel.Enable()
-        else:
-            self.panel.Disable()
-
         # pause streams when not displayed
         if not show:
             self._streambar_controller.pauseStreams()
--- a/src/odemis/gui/cont/tabs.py
+++ b/src/odemis/gui/cont/tabs.py
@@ -3416,6 +3416,8 @@ class SecomAlignTab(Tab):
 
         self.tab_data_model.tool.subscribe(self._onTool, init=True)
         main_data.chamberState.subscribe(self.on_chamber_state, init=True)
+        # stage will be used to listen to position changes (to enable/disable the tab in the right positions)
+        self.stage = self.tab_data_model.main.stage
 
     def _on_ccd_should_update(self, update):
         """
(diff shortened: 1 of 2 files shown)


lazem:
Hi @pieleric I modified this to enable/disable the tab buttons. But now if the localization tab is currently opened and the stage moved away from imaging, the tab button will be disabled but the tab will still be opened.

pieleric:
If the stage goes out of imaging position while the tab is opened, that's not the GUI's fault. There is pretty much no nice way to handle it, and then anyway, that's probably an "advanced" user, who knows what she/he is doing!

pieleric (approved):

pieleric:
Hi, this PR is messed up. It currently has 168 commits, with some conflicts.

src/odemis/gui/util/__init__.py
```diff
@@ -282,3 +283,18 @@ class AttrDict(dict):
...
+    Enable the given tab button if the stage is in target position, disable it otherwise
+    :param button: (Button) the Tab button to enable/disable
+    :param stage: (Actuator) the stage driver
+    :param pos: (dict str->float) current position to check its label
+    :param target: (int) target position label (IMAGING, LOADING..etc)
+    """
+    if getCurrentPositionLabel(pos, stage) == target:
+        button.Enable(enable=True)
+    else:
+        button.Disable()
```
pieleric, line 300: The point of Enable accepting a boolean is that you can write:
`button.Enable(getCurrentPositionLabel(pos, stage) == target)`
If you find it too obscure, then you can leave the code as it is, but drop the `enable=True` (because the asymmetry brings suspicion)

src/odemis/gui/util/__init__.py
```diff
@@ -39,6 +39,7 @@
 # Decorators & Wrappers
 # They are almost the same but the decorators assume that they are decorating
 # a method of a wx.Object (ie, it has a "self" argument)
+from odemis.acq.move import getCurrentPositionLabel, IMAGING
```
pieleric, line 42: Drop IMAGING, as it's (now) unused.

src/odemis/gui/cont/tabs.py
```diff
@@ -628,6 +641,13 @@ def Show(self, show=True):
         assert (show != self.IsShown())  # we assume it's only called when changed
         super(LocalizationTab, self).Show(show)
 
+        # Enable the tab if the stage is in imaging position, disable it otherwise
+        stage = self.tab_data_model.main.stage
+        if getCurrentPositionLabel(stage.position.value, stage) is IMAGING:
+            self.panel.Enable()
+        else:
+            self.panel.Disable()
+
```
pieleric, line 650: This whole part is not needed. Show is called when the user has pressed on the tab button (or pressed another tab). It's already too late. We just want the tab button to be disabled.

pieleric (changes requested):

src/odemis/gui/util/__init__.py
lazem: I agree it's asymmetric, it's just button.Enable() alone won't work, needs parameter. 
I changed it as you instructed.

PR #1724, opened 2021-04-01
Remove Python 2 support

pieleric:


setup.py
```diff
@@ -3,7 +3,7 @@
 
 # To rebuild just the cython modules, use these commands:
 # sudo apt-get install python-setuptools cython
-# python setup.py build_ext --inplace
+# python3 setup.py build_ext --inplace
 from __future__ import division, print_function
```
Kleijwegt, line 7: With the upgrade from python2 --> python3 these import can maybe be removed?

src/odemis/gui/cont/tabs.py
```diff
@@ -25,10 +25,10 @@
 from __future__ import division
```
Kleijwegt, line 25: This import can also be removed isn't it?

Kleijwegt (changes requested):

setup.py
pieleric: Indeed. I'll make a separate PR with removal of all these imports, for all the files.

philipwinklerdelmic (approved):

PR #1727, opened 2021-04-12
Bug fix: Delay pmt protection to avoid activating it during the acquisition

VictoriaDel:
* During the live update, horizontal stripes are observed due to the activation of pmt protection while acquiring.
    * By delaying a little bit the protection, it is not set when the stream stops and starts quickly. The pmt protection is activated when the acquisition is completed.

### DIFF
--- a/src/odemis/driver/test/pmtctrl_test.py
+++ b/src/odemis/driver/test/pmtctrl_test.py
@@ -174,6 +174,7 @@ class TestPMT(unittest.TestCase):
                 cls.bsd = child
             elif child.name == CONFIG_SCANNER["name"]:
                 cls.scanner = child
+
         cls.pmt = CLASS_PMT(name="test", role="detector",
                                  dependencies={"detector": cls.bsd,
                                            "pmt-control": cls.control})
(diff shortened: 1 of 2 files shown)


src/odemis/driver/test/pmtctrl_test.py
```diff
@@ -261,6 +267,54 @@ def receive_image(self, dataflow, image):
...
+        # Protection should be off while acquiring
+        self.assertEqual(self.control.protection.value, False)
+        # wait the last point is fully acquired
+        self.acq_complete.wait()
+        self.assertEqual(self.acq_left, 0)
+        # the protection remains off during the acquisition
+        self.assertEqual(self.pmt_protection_recorded, self.pmt_protection_expected)
+
+        # the protection remains off after settle time (10e-6) because delay is 0.1
+        time.sleep(self.scanner.settleTime)
```
pieleric, line 295: Well, settle time and protection delay have no links, so let's drop this.
If you want to say "wait a tiny bit", then just write `time.sleep(1e-3)`, to make it clear it's just an arbitrary small number

src/odemis/driver/test/pmtctrl_test.py
```diff
@@ -261,6 +267,54 @@ def receive_image(self, dataflow, image):
...
+        self.assertEqual(self.control.protection.value, False)
+
+        logging.debug("Acquiring again...")
+        self.pmt.data.subscribe(self.acquire_data)
```
pieleric, line 299: Please write `self.acq_left = 1` to make clearer what should happen here.

pieleric (changes requested):

PR #1729, opened 2021-04-13
xtclient: removed primary stigmator, because it controls the same val…

tepals:
…ue as the regular stigmator.


Kleijwegt:
Is the primary stigmator also/already deprecated on FUMO Beta?

tepals:
> Is the primary stigmator also/already deprecated on FUMO Beta?

No but the primary stigmator is the same as the regular stigmator from the base adapter.

Kleijwegt (approved):

pieleric (approved):

PR #1731, opened 2021-04-14
VA: update TupleVA to accept list as input

Sabrina85:


src/odemis/model/_vattributes.py
```diff
@@ -794,12 +794,21 @@ class TupleVA(VigilantAttribute):
     def __init__(self, value=None, *args, **kwargs):
         VigilantAttribute.__init__(self, value, *args, **kwargs)
 
+    def _set_value(self, value, **kwargs):
+        if isinstance(value, list):
+            # force tuple
```
Kleijwegt, line 799: Maybe if you change the type some logging to inform a user (during debugging) could be nice, but I don't think that is essential.

Kleijwegt (approved):
Looks good to me

PR #1732, opened 2021-04-14
ASM API wrapper: allow lists as input for cell VAs

Sabrina85:
* cellTranslation, cellDigitalGain, cellDarkOffset should also accept lists as input type as this is the format that needs to be provided in the yaml files
* simplify setters for readability
* cellDigitalGain: int can be considered a special case of float; simulator accepts ints and floats
* some reformatting and adding docstring to tests

### DIFF
--- a/src/odemis/driver/technolution.py
+++ b/src/odemis/driver/technolution.py
@@ -4,7 +4,7 @@ Created on 11 May 2020
 
 @author: Sabrina Rossberger, Kornee Kleijwegt
 
-Copyright © 2019-2020 Kornee Kleijwegt, Delmic
+Copyright © 2019-2021 Kornee Kleijwegt, Delmic
 
 This file is part of Odemis.
 
@@ -836,6 +836,7 @@ class EBeamScanner(model.Emitter):
         resolution = eff_cell_size * self.parent._mppc.shape[0:2]
 
         return tuple(resolution.astype(int))
+        return tuple(int(i) for i in resolution)
 
 
 class MirrorDescanner(model.Emitter):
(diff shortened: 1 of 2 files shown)


src/odemis/driver/technolution.py
```diff
@@ -836,6 +836,7 @@ def _setResolution(self, resolution):
         resolution = eff_cell_size * self.parent._mppc.shape[0:2]
 
         return tuple(resolution.astype(int))
+        return tuple(int(i) for i in resolution)
```
Kleijwegt, line 839: I think the second return is never reached the way it is written down now. Probably you forgot to delete the first one?

src/odemis/driver/technolution.py
```diff
@@ -1431,34 +1432,29 @@ def _setCellDigitalGain(self, cellDigitalGain):
                                  "Digital gain parameters value remain unchanged.")
 
             for column, DigitalGain in enumerate(cellDigitalGain_row):
-                if isinstance(DigitalGain, int):
-                    # Convert all input values to floats.
-                    logging.warning("Input integer values for the digital gain are converted to floats.")
-                    cellDigitalGain = tuple(tuple(float(cellDigitalGain[i][j]) for j in range(0, self.shape[0]))
-                                            for i in range(0, self.shape[0]))
-                    # Call the setter again with all int values converted to floats and return the output
-                    return self._setCellDigitalGain(cellDigitalGain)
```
Kleijwegt, line 1440: With deleting this we no longer accept integers as an input for the digital gain, for example if the number 1 is entered and somewhere rounded an error could be raised (due to the wrong type being send to the ASM or the VA complaining about containing different types, I don't know which of these was the problem out of my head).
But maybe some things changed and it is now possible to do this though.

p.s. The reason I return at line 1440 with itself is that after updating it to the right format you want to properly perform all checks again.

src/odemis/driver/technolution.py
```diff
@@ -1467,7 +1463,8 @@ def _setCellDarkOffset(self, cellDarkOffset):
             raise ValueError("An incorrect shape of the dark offset parameters is provided.\n"
                              "Please change the shape of the dark offset parameters according to the shape of the mppc "
                              "detector.\n "
-                             "Dark offset parameters value remain unchanged.")
+                             "Dark offset parameters value remain unchanged."
+                             "%s, %s, %s", cellDarkOffset, len(cellDarkOffset), self._shape)
```
Kleijwegt, line 1467: I think a comment (+ enter) on what these output numbers are would be essential, otherwise it is even hard to keep these numbers apart since this will be a very big output so something like:
"Dark offset parameters value remain unchanged. dark offset values: %s \n length of the parameters %s and shape of these parameters saved %s", cellDarkOffset, len(cellDarkOffset), self._shape)

Note: that cellDarkOffset is quite a big variable (64 numbers) so the output error message will also be big/hard to read.

Kleijwegt (changes requested):

src/odemis/driver/technolution.py
Sabrina85: This one is here by accident. I will put this in a separate pr.

src/odemis/driver/technolution.py
Sabrina85: A left over from my debugging. I remove it.

src/odemis/driver/technolution.py
Sabrina85: We still accept integers as input (see modification in line 1435). I tested all the testcases against the simulator (so I specifically uploaded the values after each change in the test) and all worked fine also with integers. I didn't find that the VA complains about different types.

Sabrina85: > p.s. The reason I return at line 1440 with itself is that after updating it to the right format you want to properly perform all checks again.
For readability, instead of returning, I converted all the elif statements to individual if statements. Thus all checks are still performed.

Sabrina85: A side note: When you run this test, it will not detect whether you actually converted to a float or not. It will just see that 0 is 0.0 and passes. If you want to check the type was converted, you need to explicitly check for the type.
```
# Test int as type for input value
self.MPPC.cellDigitalGain.value = tuple(
        tuple(int(0) for i in range(0, self.MPPC.shape[0])) for i in range(0, self.MPPC.shape[1]))
self.assertEqual(self.MPPC.cellDigitalGain.value,
                 tuple(tuple(0.0 for i in range(0, self.MPPC.shape[0])) for i in range(0, self.MPPC.shape[1])))
```
You can check it with this line:
```
self.assertEqual(int(0), 0.0)  # will pass
self.assertEqual(type(int(0)), type(0.0))  # will fail
```

pieleric (approved):

Kleijwegt (approved):

PR #1736, opened 2021-04-15
plugins ar_spectral: check the Acquisition tab is selected

pieleric:
It wouldn't work when called from another tab (eg, analysis).
Fixed it to probably work from any tab, but it gets confusing anyway,
and untested, so request the user to select the acqusition tab.


VictoriaDel (approved):
I also liked the naming "_acqui_tab" in terms of readability but I understand that having "_acqui_tab" and "acqui_tab_data" makes the names too long.

PR #1741, opened 2021-04-16
Bug fix: Pause the streams before opening the slit in Manual Focus

VictoriaDel:
* Untoggled Manual Focus button --> pause the streams before calling Sparc2ManualFocus


VictoriaDel (commented):
Thanks for updating!

PR #1742, opened 2021-04-16
ENZEL: Fix checking for same directory partition in Chamber tab 

lazem:
Before, checking for same partition used `os.stat()` on both previous and new directories. But as the new directory wasn't created yet, `os.stat(new_dir)` throws Not_Found exception. => Get the mount point of prev_dir only, then check that new_dir starts with it.

### DIFF
--- a/src/odemis/gui/cont/tabs.py
+++ b/src/odemis/gui/cont/tabs.py
@@ -2062,11 +2062,27 @@ class CryoChamberTab(Tab):
         # Get previous project dir from configuration
         prev_dir = self.conf.pj_last_path
 
+        def get_mount_point(pathname):
+            """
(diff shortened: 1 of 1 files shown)


pieleric (changes requested):
I think the entire _on_change_project_folder() is now doing too much. It should just create a new folder. Done.
So we don't need all this extra cleverness.
As a side note, it seems that shutil.move() would be able to take care of this automatically.

So, I'd suggest to not do this change, but instead rename  `_on_change_project_folder()`  to  `_on_new_project_folder()`. Then, replace the entire folder move/renaming code to just create a new folder.

pieleric:
I've just made the PR #1747 which would do the simplest alternative that I was proposing.

PR #1744, opened 2021-04-19
Orsay vacuum driver

Helsloot:


src/odemis/driver/orsay.py
```diff
@@ -0,0 +1,766 @@
...
+
+    def _updateLevel(self, parameter=None, attributeName="Actual"):
+        """
+        Reads the battery level of the UPS from the Orsay server and saves it in the level VA
+        """
+        parameter = self._system.UPScontroller.BatteryLevel if (parameter is None) else parameter
+        if parameter is not self._system.UPScontroller.BatteryLevel:
+            raise Exception("Incorrect parameter passed to _updateLevel. Parameter should be "
+                            "datamodel.HybridPlatform.UPS.UPScontroller.BatteryLevel")
+        elif attributeName == "Actual":
```
pieleric, line 757: Note that the `elif` could be replaced by just a `if`, as the previous if always raise an exception. Also, you could even avoid having the main code into a sub-block by turning around the if: 
```
if attributeName != "Actual":
   return
```

src/odemis/driver/test/orsay_test.py
```diff
@@ -0,0 +1,560 @@
...
+                             "ups": CONFIG_UPS}
+                }
+
+
+class TestOrsayStatic(unittest.TestCase):
+    """
+    Tests which don't need an Orsay component ready
+    """
+
+    def test_creation(self):
```
pieleric, line 51: You could also add a test case for when connecting to a non existent host: it should raise a HwError.

src/odemis/driver/test/orsay_test.py
```diff
@@ -0,0 +1,560 @@
...
+    def test_reconnection(self):
+        """
+        Checks that after reconnection things still work
+        """
+        self.oserver._device.HttpConnection.close()  # close the connection
+        self.oserver._device.MessageConnection.Connection.close()
+        self.oserver._device.DataConnection.Connection.close()
+        self.oserver._device.MessageConnection.dataConnection.Connection.close()
+        while not self.oserver.connected.is_set():
+            sleep(1)  # wait for the reconnection
```
pieleric, line 117: Don't rely on non official API in test cases.
I think it'd be fine to just wait "long enough" and try again.

Possibly, you could first check that the .state indicates a HwError, and that afterwards, it goes back to ST_RUNNING. (However, that's not super easy as probably the reconnection will be very fast, with the test case, so you'd need to subscribe to the state, and record all the values passed to the subscriber).

src/odemis/driver/orsay.py
```diff
@@ -0,0 +1,766 @@
...
+
+        self.update_VAs()
+
+    def update_VAs(self):
+        """
+        Update the VA's. Should be called after reconnection to the server
+        """
+        self._updateProcessInfo()
+
+    def _check_connection_thread(self):
```
pieleric, line 128: "connection_monitor" might be a simpler name.

src/odemis/driver/orsay.py
```diff
@@ -0,0 +1,766 @@
...
+        """
+        Once in a while, check the connection to the Orsay server, reconnect if needed and update all VA's
+        """
+        while not self._stop_connection_check.is_set():
+            if self._device.HttpConnection._HTTPConnection__response is None or \
+                    self._device.MessageConnection.Connection._HTTPConnection__response is None:
+                self.connected.clear()  # report the connection is broken
```
pieleric, line 135: Do this via the .state VA. Something like HwError("Connection to FIB server lost, trying to reconnect...")

src/odemis/driver/test/orsay_test.py
```diff
@@ -0,0 +1,560 @@
...
+
+        # perform some test to check writing and reading still works
+        self.psus._valve.Target = 1
+        sleep(1)
+        self.assertTrue(self.psus.power.value)
+        self.psus._valve.Target = 2
```
pieleric, line 123: Try also directly changing a VA.

src/odemis/driver/test/orsay_test.py
```diff
@@ -0,0 +1,560 @@
...
+        """
+        Check that the processInfo VA is updated properly and an exception is raised when the wrong parameter is passed
+        """
+        self.assertRaises(Exception, self.oserver._updateProcessInfo,
+                          {'parameter': self.oserver.datamodel.HybridPlatform.Cancel})
+        if not TEST_NOHW:
+            self.skipTest("TEST_NOHW is not set, cannot force data on Actual parameters of Orsay server outside of "
+                          "simulation")
+        test_string = "Some process information"
+        self.oserver.datamodel.HybridPlatform.ProcessInfo.Actual = test_string
```
pieleric, line 104: Instead of always writing `self.oserver.datamodel`, store the `datamodel` directly on the self in the setUpClass()

src/odemis/driver/test/orsay_test.py
```diff
@@ -0,0 +1,560 @@
...
+
+        oserver.terminate()
+
+
+class TestOrsay(unittest.TestCase):
+    """
+    Tests to run on the main Orsay component
+    """
+
+    def setUp(self):
```
pieleric, line 69: This start/stop the components before every test cases. That works, but might be quite slow. It'd usually nicer to use 
 ```
@classmethod
def setUpClass(cls):
```
(and tearDownClass)

File: src/odemis/driver/test/orsay_test.py [hunk 9]
```diff
@@ -0,0 +1,560 @@
...
+        sleep(0.5)
+        self.assertIn("ValvePneumaticSuspension could not be contacted", self.psus.state.value)
+        self.psus._valve.Target = 1
+
+    def test_updatePower(self):
+        """
+        Check that the power VA is updated correctly and an exception is raised when the wrong parameter is passed
+        """
+        self.assertRaises(Exception, self.psus._updatePower,
+                          {'parameter': self.oserver.datamodel.HybridPlatform.Cancel})
```
pieleric, line 205: This type of code is really hard to read. You can write like this:
```
with self.assertRaises(Exception):
    self.psus._updatePower(parameter=self.oserver.datamodel.HybridPlatform.Cancel)
```
Update also the exception type here.

(Please do it everywhere)

src/odemis/driver/orsay.py
```diff
@@ -0,0 +1,766 @@
...
+                self._device = None  # destroy the current connection object
+                try:
+                    self._device = Connection(self._host)
+                    time.sleep(1)
+                    self.on_connect()
+                    for child in self.children.value:
+                        child.on_connect()
+                    self.connected.set()  # report the connection is fixed
+                except Exception:
+                    logging.warning("Trying to reconnect to Orsay server")
```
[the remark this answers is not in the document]
Helsloot: The exception I'm trying to catch here is the one raised by Connection(self._host) in case the connection is still not there. The Orsay API, however, only raises a simple exception in that case, no specific type, no message attached to it. I'll see what I can do to change their API so it raises proper exceptions and make another merge request to Orsay's GitLab repo.

src/odemis/driver/orsay.py
```diff
@@ -0,0 +1,880 @@
...
+        return (bool): goal position of the gate as set to the server: (True: "open", False: "closed")
+
+        Opens ValveP5 on the Orsay server if argument goal is True. Closes it otherwise.
+        """
+        logging.debug("Setting gate to %s." % ("open" if goal else "closed"))
+        self._gate.IsOpen.Target = VALVE_OPEN if goal else VALVE_CLOSED
+        return self._gate.IsOpen.Target == VALVE_OPEN
+
+    @isasync
+    def moveAbs(self, pos, wait=True):
```
pieleric, line 550: Drop the wait. It cannot be anything but True, as this is the behaviour of the Future: it should report it's finished when the move is finished.

src/odemis/driver/orsay.py
```diff
@@ -0,0 +1,880 @@
...
+        """
+        self._checkMoveAbs(pos)
+        return self._executor.submit(self._changeVacuum, goal=pos["vacuum"], wait=wait)
+
+    @isasync
+    def moveRel(self, shift):
+        """
+        Move the axis of this actuator by shift.
+        """
+        pass
```
pieleric, line 562: Raise an error if trying to move vacuum.

src/odemis/driver/orsay.py
```diff
@@ -0,0 +1,880 @@
...
+                             "datamodel.HybridPlatform.AnalysisChamber.Pressure. Parameter passed is %s"
+                             % parameter.Name)
+        if attributeName != "Actual":
+            return
+        self.pressure._set_value(float(parameter.Actual), force_write=True)
+
+    def _changeVacuum(self, goal, wait=True):
+        """
+        goal (int): goal state of the vacuum: (0: "vented", 1: "primary vacuum", 2: "high vacuum")
+        wait (bool): if True, the function will block until the goal vacuum state is reached
```
pieleric, line 525: You don't see to ever need to not wait => drop this argument, and always wait.

src/odemis/driver/orsay.py
```diff
@@ -0,0 +1,880 @@
...
+        return (int): actual state of the vacuum at the end of this function: (0: "vented", 1: "primary vacuum",
+                      2: "high vacuum")
+
+        Sets the vacuum status on the Orsay server to argument goal and waits until it is reached.
+        Then returns the reached vacuum status.
+        """
+        logging.debug("Setting vacuum status to %s." % self.axes["vacuum"].choices[goal])
+        self._chamber.VacuumStatus.Target = goal
```
pieleric, line 533: To be extra careful, you should set `_vacuumStatusReached.clear()` before changing the target. Otherwise, if this code execute really fast, it might end up with the event still left in set() , and not wait.

src/odemis/driver/orsay.py
```diff
@@ -0,0 +1,880 @@
...
+        if wait:
+            self._vacuumStatusReached.wait()
```
pieleric, line 535: Waiting forever is always dangerous. So I'd recommend to put a timeout. Pick something large, like absolutely unlikely to happen when using the system (1 m, 5m, 1h? I don't know how long it normally takes, but just pick something ~10x longer).
If the timeout is triggered, you should raise a TimeoutError.

src/odemis/driver/orsay.py
```diff
@@ -0,0 +1,880 @@
...
+        """
+        self._gate.ErrorState.Unsubscribe(self._updateErrorState)
+        self._chamber.VacuumStatus.Unsubscribe(self._updatePosition)
+        self._chamber.Pressure.Unsubscribe(self._updatePressure)
+        self._gate.IsOpen.Unsubscribe(self._updateGateOpen)
+        if self._executor:
+            self._executor.shutdown()
+            self._executor = None
+        _gate = None
+        _chamber = None
```
pieleric, line 586: `self.` missing?
How come it's not picked up in your test cases? Do you call `.terminate()`?

src/odemis/driver/orsay.py
```diff
@@ -0,0 +1,880 @@
...
+        """
+
+        axes = {"vacuum": model.Axis(unit=None, choices={0: "vented", 1: "primary vacuum", 2: "high vacuum"})}
+
+        model.Actuator.__init__(self, name, role, parent=parent, axes=axes, **kwargs)
+
+        self._gate = None
+        self._chamber = None
+
+        self.pressure = model.FloatContinuous(0.0, range=(0, 110000), readonly=True, unit="Pa")
```
pieleric, line 387: Doest the parameter provide a .Min and .Max? If so, please use it in the range. Otherwise, please move the range to a constant at the top of the file, like:
`VACUUM_CHAMBER_PRESSURE_RNG = (0, 110000)  # Pa`

src/odemis/driver/orsay.py
```diff
@@ -0,0 +1,880 @@
...
+        • pressure (FloatContinuous, range=(0, 110000), read-only, unit is "Pa", value is _gauge.Actual)
+        """
+
+        # we will fill the set of children with Components later in ._children
+        model.HwComponent.__init__(self, name, role, parent=parent, **kwargs)
+
+        self._valve = None
+        self._gauge = None
+
+        self.pressure = model.FloatContinuous(0.0, range=(0, 110000), readonly=True, unit="Pa")
```
pieleric, line 229: Same comment as for vacuum chamber range. Is expected that they have the same maximum and minimum? I assume that the pneumatic table would go above the normal atmospheric pressure, so anything below would be very odd, especially 0...

src/odemis/driver/orsay.py
```diff
@@ -0,0 +1,880 @@
...
+                             % parameter.Name)
+        if attributeName == "Actual":
+            return
+        self.nitrogenPressure._set_value(float(parameter.Actual), force_write=True)
+
+    def terminate(self):
+        """
+        Called when Odemis is closed
+        """
+        self._system.Manometer1.ErrorState.Unsubscribe(self._updateErrorState)
```
pieleric, line 809: We have the convention that it's always safe to call .terminate() multiple times in a row. So please, protect with a `if self._system`.

src/odemis/driver/orsay.py
```diff
@@ -0,0 +1,880 @@
...
+
+class UPS(model.HwComponent):
+    """
+    This represents the uniterupted power supply from Orsay Physics
+    """
+
+    def __init__(self, name, role, parent, **kwargs):
+        """
+        Defines the following VA's and links them to the callbacks from the Orsay server:
+        • level (FloatContinuous, range=(0.0, 1.0), read-only, unit is "",
```
pieleric, line 829: Instead of saying "unit is """, I'd suggest to clarify that this is the ratio between empty and fully charged.

src/odemis/driver/orsay.py
```diff
@@ -0,0 +1,880 @@
...
+        if attributeName != "Actual":
+            return
+        currentLevel = float(parameter.Actual)
+        self.level._set_value(currentLevel / 100, force_write=True)
+
+    def terminate(self):
+        """
+        Called when Odemis is closed
+        """
+        self._blevel.Unsubscribe(self._updateLevel)
```
pieleric, line 879: Please protect with a `if self._blevel`.

src/odemis/driver/orsay.py
```diff
@@ -0,0 +1,880 @@
...
+        _blevel = None
```
pieleric, line 880: Missing `self`.

src/odemis/driver/test/orsay_test.py
```diff
@@ -0,0 +1,652 @@
...
+
+        f = self.pressure.moveAbs({"vacuum": 0})
+        f.wait()
+        self.assertEqual(self.pressure.position.value["vacuum"], 0)
+        self.assertAlmostEqual(self.pressure.pressure.value, 100000, delta=10000)  # tune the goal and alowed difference
+
+        self.pressure.moveAbs({"vacuum": 1})
+        f = self.pressure.moveAbs({"vacuum": 0})
+        f.wait()
+        self.assertEqual(self.pressure.position.value["vacuum"], 0)
```
pieleric, line 367: All these "0 Pa" are probably not going to be acceptable on the real system (it's unreachable). We should either find typical pressure values now, or review all these numbers once we have access to the real hardware. If you want to wait, it's fine, but make it a big "TODO".

src/odemis/driver/test/orsay_test.py
```diff
@@ -0,0 +1,652 @@
...
+        self.pressure._gate.IsOpen.Target = 1
+        sleep(1)
+        self.assertTrue(self.pressure.gateOpen.value)
+
+        self.pressure._gate.IsOpen.Target = 2
+        sleep(1)
+        self.assertFalse(self.pressure.gateOpen.value)
+
+        self.pressure.gateOpen.value = True
+        self.assertEqual(self.pressure._gate.IsOpen.Target, 1)
```
pieleric, line 318: You should also check that reading self.pressure.gateOpen.value after 1 sec still reports the right value.

src/odemis/driver/test/orsay_test.py
```diff
@@ -0,0 +1,652 @@
...
+    def setUpClass(cls):
+        """
+        Setup the Orsay client
+        """
+        if TEST_NOHW == 1:
+            raise unittest.SkipTest("TEST_NOHW is set. No server to contact.")
+        try:
+            cls.oserver = orsay.OrsayComponent(**CONFIG_ORSAY)
+        except Exception as e:
+            raise HwError(str(e))
```
pieleric, line 172: No need to translate the exceptions in the test case. Any exception will make the test case fail, and it's clearer if you know what is the type of exception.

src/odemis/driver/test/orsay_test.py
```diff
@@ -0,0 +1,652 @@
...
+        sleep(1)
+        while not self.oserver.state.value == model.ST_RUNNING:
+            sleep(2)  # wait for the reconnection
+
+        # perform some test to check writing and reading still works
+        self.psus._valve.Target = 1
+        sleep(1)
+        self.assertTrue(self.psus.power.value)
+
+        self.psus._valve.Target = 2
```
pieleric, line 144: For all these valve.Target = 1/2 , could you replace them by VALVE_OPEN and VALVE_CLOSE? Otherwise, it's very hard to read the test cases.

src/odemis/driver/test/orsay_test.py
```diff
@@ -0,0 +1,652 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+Created on 6 April 2021
+
+@author: Arthur Helsloot
+
+Copyright © 2021-2023 Arthur Helsloot, Delmic
```
pieleric, line 8: That's very looking forward the future! For now you can just write "2021" ;-)

src/odemis/driver/orsay.py
```diff
@@ -0,0 +1,880 @@
...
+
+class vacuumChamber(model.Actuator):
+    """
+    This represents the vacuum chamber from Orsay Physics
+    """
+
+    def __init__(self, name, role, parent, **kwargs):
+        """
+        Has axes:
+        • "vacuum": unit is "None", choices is {0 : "vented", 1 : "primary vacuum", 2 : "high vacuum"}
```
pieleric, line 372: Don't mention the unit, if there is no unit.

pieleric (changes requested):

src/odemis/driver/orsay.py
```diff
@@ -0,0 +1,880 @@
...
+                    self.on_connect()
+                    for child in self.children.value:
+                        child.on_connect()
+                    self.state._set_value(model.ST_RUNNING, force_write=True)
+                except Exception as ex:
+                    logging.exception("Trying to reconnect to Orsay server: %s." % ex)
+            else:
+                self.update_VAs()
+                for child in self.children.value:
+                    child.update_VAs()
```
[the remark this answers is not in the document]
Helsloot: I just realised this will actually never fail due to connection errors. The update_VAs methods read parameters from the ConsoleClient component, which will stay in existence until the orsayComponent detects that there is no connection. Additionally the update_VAs will never be called when the ConsoleClient component does not exist. So if the connection would (forcefully) disappear during a call to an update_VAs method, the update would succeed just fine, but potentially with outdated information.
I added a try block anyway to catch other exceptions, might they occur.

src/odemis/driver/orsay.py
```diff
@@ -135,31 +140,40 @@ def _connection_monitor(self):
...
+                        self.state._set_value(model.ST_RUNNING, force_write=True)
+                    except Exception as ex:
+                        logging.exception("Trying to reconnect to Orsay server: %s." % ex)
+                else:
+                    try:
+                        self.update_VAs()
+                        for child in self.children.value:
+                            child.update_VAs()
+                    except Exception as e:
+                        logging.exception(e)
```
pieleric, line 170: `logging.exception()` is a little special, as it has special code to output the exception. So you don't need to also show the exception. You can just write like:
```python
except Exception:
     logging.exception("Failure while updating VAs")
```
Same thing for the catch just above,  l 163, and just below, l 173.

pieleric (changes requested):

src/odemis/odemisd/main.py
```diff
@@ -662,7 +662,7 @@ def main(args):
     # The settings file is opened here because root privileges are dropped at some point after
```
pieleric, line 662: These changes seem to be from a different commit in Odemis upstream. Probably rebasing against upstream would drop this from the PR, which would be handy, as to merge your PR, I'll do a squash, which would bring conflicts.

pieleric (changes requested):

PR #1747, opened 2021-04-20
gui cont tabs CryoChamberTab: on new project, just create a new folder

pieleric:
Don't try to rename the previous folder, just create a new one.
The only exception is if the previous folder is empty.

Also fix the issue that the automatically generated folder name was
inside the previous folder.

### DIFF
--- a/src/odemis/gui/cont/tabs.py
+++ b/src/odemis/gui/cont/tabs.py
@@ -2044,6 +2044,7 @@ class CryoChamberTab(Tab):
         Shows a dialog to change the path and name of the project directory.
         returns nothing, but updates .conf and project path text control
         """
+        # TODO: do not warn if there is no data (eg, at init)
         box = wx.MessageDialog(self.main_frame,
                                "This will clear the current project data from Odemis",
                                caption="Reset Project", style=wx.YES_NO | wx.ICON_QUESTION | wx.CENTER)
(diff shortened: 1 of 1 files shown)


src/odemis/gui/cont/tabs.py
```diff
@@ -2052,68 +2053,57 @@ def _on_change_project_folder(self, evt):
...
-            """
-            Move all files found in source directory to destination
-            """
-            file_names = os.listdir(source)
-            for file_name in file_names:
-                shutil.move(os.path.join(source, file_name), destination)
+        # Three possibilities:
+        # * The folder doesn't exists yet => create it
+        # * The folder already exists and is empty => nothing else to do
+        # * The folder already exists and has files in it => Error (because we might override files)
```
lazem, line 2071: Though with the current filedialog, there is no way to choose an existing folder . The Save button will always open the existing folder if it's selected or added on the Name text field.

lazem (commented):
I think it looks good, fixes the folder inside folder issue, yet it seems it still needs more testing as some cases (choosing an existing folder) won't be possible with the file dialog.

src/odemis/gui/cont/tabs.py
pieleric: Yes, I'm aware that with the current GTK3 version, it's not possible. Still, I wanted to handle this (theoretical) case, so that if the file selection dialog changes behaviour, Odemis handles it fine too.

PR #1748, opened 2021-04-21
Spot radial symmetry center

effting:
Improves spot center location algorithm:
- change of function name to better describe what it's doing
- update docstring
- returns index in `(j, i)` index format to conform to guideline
- update unit tests


src/odemis/util/spot.py
```diff
@@ -132,6 +133,15 @@ def SpotIntensity(data, background=None):
 
 
 def FindCenterCoordinates(image, smoothing=True):
+    warnings.warn(
+        "FindCenterCoordinates will become deprecated, use RadialSymmetryCenter instead.",
+        PendingDeprecationWarning
+    )
+    jc, ic = RadialSymmetryCenter(image, smoothing)
+    return ic, jc
+
```
pieleric, line 142: You're now causing all SECOM/DELPHI code to generate a warning.
I'd rather have all the callers be updated.

src/odemis/util/spot.py
```diff
@@ -147,70 +157,75 @@ def FindCenterCoordinates(image, smoothing=True):
     Returns
     -------
     pos : tuple
-        Position of the radial symmetry center in px from the center of the
-        image.
+        Position of the radial symmetry center `(dj, di)` in pixels relative to
```
pieleric, line 160: In Odemis, the convention is to report positions as (x,y). I don't see the point of having this function behave differently.
Admittedly, we have some exceptions for values which could be directly used as index in a numpy array, but that's it. Here, the position is relative to the center *and* floats, so it's never going to be directly used an array index. So stick to x,y.

src/odemis/util/spot.py
```diff
@@ -147,70 +157,75 @@ def FindCenterCoordinates(image, smoothing=True):
...
 
     # Solve the linear set of equations in a least-squares sense.
     # Note: rcond is set explicitly to have correct and consistent behavior,
     #       independent on numpy version.
     rcond = numpy.finfo(numpy.float64).eps * max(m, n)
-    ic, jc = numpy.linalg.lstsq(numpy.vstack((w * a, w * b)).T, w * c, rcond)[0]
-
-    # Convert from index (top-left) to (center) position information.
-    xc = jc - 0.5 * float(m) + 0.5
-    yc = ic - 0.5 * float(n) + 0.5
```
pieleric, line 211: Where is this gone? I guess this is now in the meshgrid call, right?

src/odemis/util/test/spot_test.py
```diff
@@ -24,11 +25,21 @@
...
 
+def ignore_pending_deprecation_warning(func):
+    @functools.wraps(func)
+    def decorated(self, *args, **kwargs):
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", PendingDeprecationWarning)
+            func(self, *args, **kwargs)
+    return decorated
+
+
```
pieleric, line 42: Instead of having all this fancy/complicated code hiding warning in the test cases, it'd be best update all the callers, and not have warnings.

pieleric (changes requested):
I'm a little confused by what is the actual purpose of all this change? It seems the only new code is a test case. Do you use this also in a outside script?

src/odemis/util/spot.py
effting: The problem is that `FindCenterCoordinates` returns coordinates as `(x, -y)`. That's confusing as hell.
Since this function doesn't care about the physical size I thought it would make sense to use the pixel coordinates (following the discussions we had in September last year), and then convert to physical coordinates using `to_physical_space` when appropriate.

src/odemis/util/spot.py
effting: Yes; that's absorbed in the meshgrid call

src/odemis/util/synthetic.py
```diff
@@ -0,0 +1,123 @@
...
+You should have received a copy of the GNU General Public License
+along with this program; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301,
+USA.
+
+"""
+import math
+from typing import List, Tuple, Union
+
+import numpy as np
```
pieleric, line 24: Please write "numpy" in the rest of the code. No need to rename it to "np". This is the convention in Odemis.

src/odemis/util/synthetic.py
```diff
@@ -0,0 +1,123 @@
...
+        raise ValueError("The wavelength should be positive.")
+    if numerical_aperture >= refractive_index:
+        raise ValueError(
+            "The numerical aperture should be less than the refractive index."
+        )
+
+    k = 2 * math.pi / wavelength
+    nk = refractive_index * k
+    sa = numerical_aperture / refractive_index
+    ca = math.sqrt(1 - sa * sa)
```
pieleric, line 78: can you write `sa * sa` -> `sa ** 2`

src/odemis/util/synthetic.py
```diff
@@ -0,0 +1,123 @@
...
+    t = pow(ca, 1.5)
```
pieleric, line 79: You can replace `pow()` by `ca ** 1.5`. (and also just below `ca ** 3.5`)

src/odemis/util/synthetic.py
```diff
@@ -0,0 +1,123 @@
...
+) -> np.ndarray:
+    """
+    Return a synthetic spot image of a point-spread function (PSF) approximated
+    by a 2-dimensional Gaussian function.
+
+    Parameters
+    ----------
+    shape : tuple of ints
+        Shape of the array, e.g. ``(9, 9)``.
+    loc : tuple of floats
```
pieleric, line 95: "tuple of floats, or list of tuple of floats", right?

src/odemis/util/synthetic.py
```diff
@@ -0,0 +1,123 @@
...
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301,
+USA.
+
+"""
+import math
```
pieleric, line 21: Could you add here a comment of 1 or 2 lines explaining what sort of code goes in this module? "synthetic" has a lot of meanings, and here I guess you mean "fake based on theory". So indicate it's to simulated data based on theoretical behaviour, or some explanation of this sort.

src/odemis/util/synthetic.py
```diff
@@ -0,0 +1,123 @@
...
+    microscope point spread function.
+
+    Parameters
+    ----------
+    refractive_index : float, >= 1
+        Refractive index
+    numerical_aperture: float, positive
+        Numerical aperture of the optical system
+    wavelength : float
+        Wavelength.
```
pieleric, line 49: What's the unit? Is that µm? If so, please change to m.

src/odemis/util/test/spot_test.py
```diff
@@ -218,5 +218,83 @@ def test_sanity(self):
...
+    @classmethod
+    def setUpClass(cls):
+        """Create a synthetic dataset of 1000 spot images."""
+        n = 1000
+        shape = (9, 9)
+        refractive_index = 1
+        numerical_aperture = 0.95
+        wavelength = 0.55  # [um]
+        magnification = 40
+        pixel_size = 6.5  # [um]
```
pieleric, line 236: Would be nicer if you could write this in m (eg 6.5e-6 m)

pieleric (changes requested):

src/odemis/util/synthetic.py
effting: OK

src/odemis/util/synthetic.py
effting: OK

src/odemis/util/synthetic.py
effting: OK

src/odemis/util/synthetic.py
effting: Yes, indeed. Updated.

src/odemis/util/synthetic.py
effting: OK

src/odemis/util/synthetic.py
effting: The units of `wavelength` and `sigma` are the same and the user is free to choose. Updated the docstring to reflect this.

src/odemis/util/test/spot_test.py
effting: OK

src/odemis/util/spot.py
```diff
@@ -131,7 +134,25 @@ def SpotIntensity(data, background=None):
...
+    """
+    Returns the radial symmetry center of the image with sub-pixel resolution.
+    The position is returned as a tuple `(dx, -dy)` in pixels relative to the
+    center of the image.
+
+    For more information see radial_symmetry_center().
+
+    """
+    ji = radial_symmetry_center(image, smoothing)
+    xc, yc = to_physical_space(ji, image.shape)
```
tepals, line 149: Is `to_physical_space` now only used to move the origin to the center of the image?

src/odemis/util/test/spot_test.py
```diff
@@ -218,5 +218,83 @@ def test_sanity(self):
...
+
+    def test_sanity(self):
+        """
+        Create an image consisting of all zeros and a single pixel with value
+        one. radial_symmetry_center should return the pixel index of this pixel.
+        """
+        for n in range(5, 12):
+            for m in range(5, 12):
+                for j in range(2, n - 2):
+                    for i in range(2, m - 2):
```
tepals, line 291: Is it really necessary to run this, this often?

src/odemis/util/spot.py
effting: The idea here is to keep `FindCenterCoordinates()` for backwards compatibility with existing functions.

The code shows that `radial_symmetry_center()` returns values in pixel coordinates. And that `FindCenterCoordinates` returns in a confusing `x, -y` reference system.

src/odemis/util/test/spot_test.py
effting: This test case runs in 260 ms on my laptop. I guess we're fine?

tepals: that should indeed be fine!

pieleric (approved):

tepals (approved):

PR #1749, opened 2021-04-21
ASM API wrapper: extend VAs to check dependencies

Sabrina85:
The mbscanner.resolution VA, the mppc.cellTranslation VA and the mppc.cellCompleteResolution VA depend on each other.
So far only when starting to acquire a megafield and by then uploading the settings in the respective VAs, an error was raised if certain dependendies were not met.
Modified setters to check directly if dependencies are met or not and don't wait until acquisition is started.

rule: cell complete size >= cell effective size + max cell translation value

### DIFF
--- a/src/odemis/driver/technolution.py
+++ b/src/odemis/driver/technolution.py
@@ -823,6 +823,7 @@ class EBeamScanner(model.Emitter):
         number of cell images (detector shape) multiplied with the size of a cell image (effective cell size). Method
         sets the resolution of a single field image, so that the effective cell size is a multiple of 4 (in both x
         and y direction).
+        The following requirements need to be met: effective cell size <= cell complete size - max cell translation
 
         :param resolution (int, int): requested resolution for a single field image. It is the product of the number of
         cell images multiplied by the effective cell size.
@@ -830,9 +831,24 @@ class EBeamScanner(model.Emitter):
         """
         COMMON_DIVISOR = 4
 
-        req_eff_cell_size = numpy.array(resolution) / self.parent._mppc.shape[0:2]  # Requested effective cell size
+        eff_cell_size = numpy.array(resolution) / self.parent._mppc.shape[0:2]  # effective cell size
         # Round to closest multiple of COMMON_DIVISOR (x.5 is rounded down)
-        eff_cell_size = numpy.round(req_eff_cell_size / COMMON_DIVISOR) * COMMON_DIVISOR
+        eff_cell_size = numpy.round(eff_cell_size / COMMON_DIVISOR) * COMMON_DIVISOR
+
+        # check: effective cell size <= cell complete size - max cell translation = minimal effective cell size;
+        max_cell_translation = max(max(self.parent._mppc.cellTranslation.value))
+        min_eff_cell_size = (self.parent._mppc.cellCompleteResolution.value[0] - max_cell_translation[0],
+                             self.parent._mppc.cellCompleteResolution.value[1] - max_cell_translation[1])
+
+        if tuple(eff_cell_size) > min_eff_cell_size:
+            raise ValueError("(Requested resolution of %s is too big. The resolution is the effective cell size "
+                             "multiplied by the detector shape. The effective cell size must be equal or smaller than "
+                             "the complete cell size minus the maximum cell translation value. "
+                             "The minimal resolution possible with the currently settings for cell complete resolution "
+                             "and cell translation is %s.\n "
+                             "Resolution values remain unchanged."
+                             % (resolution, min_eff_cell_size))
+
         resolution = eff_cell_size * self.parent._mppc.shape[0:2]
 
         return tuple(int(i) for i in resolution)
@@ -1017,7 +1033,8 @@ class MPPC(model.Detector):
                 setter=self._setCellDigitalGain)
 
         # The minimum of the cell resolution cannot be lower than the minimum effective cell size.
-        self.cellCompleteResolution = model.ResolutionVA((900, 900), ((12, 12), (1000, 1000)))
+        self.cellCompleteResolution = model.ResolutionVA((900, 900), ((12, 12), (1000, 1000)),
+                                                         setter=self._setCellCompleteResolution)
 
         # Setup hw and sw version
         self._swVersion = self.parent.swVersion
@@ -1366,6 +1383,7 @@ class MPPC(model.Detector):
         And finally, all cell translations per row are nested in another tuple representing the full cell image.
         (((x0-int, y0-int), (x1-int, y1-int)....), ((x8-int, x8-int), .....), ......) etc.
         This setter checks the correct shape of the nested tuples, the type and minimum value.
+        The following requirement needs to be met: max cell translation <= cell complete size - effective cell size
 
         :param cellTranslation: (nested tuple of ints)
         :return: cell translation: (nested tuple of ints)
@@ -1401,6 +1419,21 @@ class MPPC(model.Detector):
                     raise ValueError("Please use a minimum of 0 cell translation coordinates of cell (%s, %s).\n"
                                      "Cell translation parameter values remain unchanged." %
                                      (row, column))
+
+        # check: max cell translation <= cell complete size - effective cell size = max possible cell translation
+        max_cell_translation = max(max(cellTranslation))
+        eff_cell_res = (int(self.parent._ebeam_scanner.resolution.value[0] / self.shape[0]),
+                        int(self.parent._ebeam_scanner.resolution.value[1] / self.shape[1]))
+        max_possible_cell_translation = (self.cellCompleteResolution.value[0] - eff_cell_res[0],
+                                         self.cellCompleteResolution.value[1] - eff_cell_res[1])
+
+        if max_cell_translation > max_possible_cell_translation:
+            raise ValueError("Requested maximum cell translation value of %s is too large. The cell translation values "
+                             "must be equal or smaller than the complete cell size minus the effective cell size. "
+                             "The maximal cell translation value possible with the currently set effective cell size "
+                             "and cell complete size is %s.\n "
+                             "Cell translation values remain unchanged."
+                             % (max_cell_translation, max_possible_cell_translation))
 
         # force items to be a tuple
         cellTranslation = tuple(tuple(tuple(item) for item in row) for row in cellTranslation)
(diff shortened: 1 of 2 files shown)


Sabrina85:
@pieleric I found a dependency that should always be met so I did not implement a hierarchy as I didn't find a good reasoning to order the VAs. So now I always raise an error when the dependency is not met. Let me know what you think.

src/odemis/driver/technolution.py
```diff
@@ -823,16 +823,32 @@ def _setResolution(self, resolution):
...
 
-        req_eff_cell_size = numpy.array(resolution) / self.parent._mppc.shape[0:2]  # Requested effective cell size
+        eff_cell_size = numpy.array(resolution) / self.parent._mppc.shape[0:2]  # effective cell size
         # Round to closest multiple of COMMON_DIVISOR (x.5 is rounded down)
-        eff_cell_size = numpy.round(req_eff_cell_size / COMMON_DIVISOR) * COMMON_DIVISOR
+        eff_cell_size = numpy.round(eff_cell_size / COMMON_DIVISOR) * COMMON_DIVISOR
+
+        # check: effective cell size <= cell complete size - max cell translation = minimal effective cell size;
+        max_cell_translation = max(max(self.parent._mppc.cellTranslation.value))
+        min_eff_cell_size = (self.parent._mppc.cellCompleteResolution.value[0] - max_cell_translation[0],
```
pieleric, line 840: Isn't this the max_eff_cell_size, as in "the maximum value possible"?

src/odemis/driver/technolution.py
```diff
@@ -823,16 +823,32 @@ def _setResolution(self, resolution):
...
+                             self.parent._mppc.cellCompleteResolution.value[1] - max_cell_translation[1])
+
+        if tuple(eff_cell_size) > min_eff_cell_size:
+            raise ValueError("(Requested resolution of %s is too big. The resolution is the effective cell size "
+                             "multiplied by the detector shape. The effective cell size must be equal or smaller than "
+                             "the complete cell size minus the maximum cell translation value. "
+                             "The minimal resolution possible with the currently settings for cell complete resolution "
+                             "and cell translation is %s.\n "
+                             "Resolution values remain unchanged."
```
pieleric, line 849: That's a really long error message. Can you make it more compact? At least, the last not about unchanged value can go away: it's the expected behaviour if a "=" fails, and it's what happens with all the VAs.
Maybe replace the english by just mathematical formula/pseudo code?

src/odemis/driver/technolution.py
```diff
@@ -1402,6 +1420,21 @@ def _setCellTranslation(self, cellTranslation):
                                      "Cell translation parameter values remain unchanged." %
                                      (row, column))
 
+        # check: max cell translation <= cell complete size - effective cell size = max possible cell translation
+        max_cell_translation = max(max(cellTranslation))
```
pieleric, line 1424: If `cellTranslation` was a numpy array, you could directly write `max(cellTranslation)`, otherwise, you have to explicitly iterate over each sub-tuple before taking the whole max.

src/odemis/driver/test/technolution_test.py
```diff
@@ -432,6 +432,58 @@ def test_resolution_VA(self):
...
+
+        # rule: max cell translation <= cell complete size - effective cell size
+        self.MPPC.cellTranslation.value = \
+            tuple(tuple((0, 0) for i in range(0, self.MPPC.shape[0])) for i in range(0, self.MPPC.shape[1]))
+        self.EBeamScanner.resolution.value = (6400, 6400)  # -> cell eff res = 800
+        self.MPPC.cellCompleteResolution.value = (810, 810)
+        with self.assertRaises(ValueError):
+            # not max cell translation <= cell complete size - effective cell size  (! 50 <= 810 - 800)
+            self.MPPC.cellTranslation.value = \
+                tuple(tuple((50, 50) for i in range(0, self.MPPC.shape[0])) for i in range(0, self.MPPC.shape[1]))
```
pieleric, line 447: Please add a test case which fails with the current max(max()) implementation. I think it's just a matter of pass good X values, but incorrect Y values.

pieleric (changes requested):
The main worry I have about this API change is that it now can happen that for a new set of (cell complete size, cell effective size, cell translation), depending on the current VA values, it's not possible to change the values if done always in the same order. So the code changing the values (coming from calibration) needs to be extra smart, and either adjust the order it sets the VA based on the previous value, or put some intermediary values.

So please document it. Possibly with some example code (in the test cases?).

src/odemis/driver/technolution.py
```diff
@@ -823,16 +823,32 @@ def _setResolution(self, resolution):
...
         COMMON_DIVISOR = 4
 
-        req_eff_cell_size = numpy.array(resolution) / self.parent._mppc.shape[0:2]  # Requested effective cell size
+        eff_cell_size = numpy.array(resolution) / self.parent._mppc.shape[0:2]  # effective cell size
         # Round to closest multiple of COMMON_DIVISOR (x.5 is rounded down)
-        eff_cell_size = numpy.round(req_eff_cell_size / COMMON_DIVISOR) * COMMON_DIVISOR
+        eff_cell_size = numpy.round(eff_cell_size / COMMON_DIVISOR) * COMMON_DIVISOR
+
+        # check: effective cell size <= cell complete size - max cell translation = minimal effective cell size;
+        max_cell_translation = max(max(self.parent._mppc.cellTranslation.value))
```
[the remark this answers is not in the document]
Sabrina85: This is indeed incorrect. I wanted to have the max value per axis. So (max_x, max_y). I will fix it.

src/odemis/driver/technolution.py
Sabrina85: yes

src/odemis/driver/technolution.py
```diff
@@ -1402,6 +1418,18 @@ def _setCellTranslation(self, cellTranslation):
                                      "Cell translation parameter values remain unchanged." %
                                      (row, column))
 
+        # check: max cell translation <= cell complete size - effective cell size = max possible cell translation
+        max_cell_translation = numpy.max(numpy.array(cellTranslation), axis=(0, 1))  # (max_x, max_y)
+        eff_cell_size = numpy.array(numpy.array(self.parent._ebeam_scanner.resolution.value) /
+                                   numpy.array(self.shape[0:2]), dtype=int)
```
pieleric, line 1424: Why isn't this done the same way as in setResolution()? It'd be easier to read if it was exactly the same code. I think that the code `numpy.array(resolution) / self.parent._mppc.shape[0:2]` would work fine here. If you really want an int, you could use `//`.

src/odemis/driver/technolution.py
```diff
@@ -1490,6 +1518,30 @@ def _setCellDarkOffset(self, cellDarkOffset):
...
+        The following requirement needs to be met: cell complete size >= effective cell size + max cell translation
+
+        :param cellCompleteResolution: (int, int) The cell complete resolution to be set.
+        :return: (int, int) The new cell complete resolution.
+        """
+
+        # check: cell complete size >= effective cell size + max cell translation = minimal cell complete size
+        max_cell_translation = numpy.max(numpy.array(self.cellTranslation.value), axis=(0, 1))  # (max_x, max_y)
+        eff_cell_size = numpy.array(numpy.array(self.parent._ebeam_scanner.resolution.value) /
+                                    numpy.array(self.shape[0:2]), dtype=int)
```
pieleric, line 1534: Use the same as in setResolution() too.

pieleric (changes requested):

src/odemis/driver/test/technolution_test.py
```diff
@@ -432,6 +428,95 @@ def test_resolution_VA(self):
...
+            tuple(tuple((50, 50) for i in range(0, self.MPPC.shape[0])) for i in range(0, self.MPPC.shape[1]))
+        self.MPPC.cellCompleteResolution.value = (850, 850)
+
+        # Third example increase sizes 2:
+        # first cellCompleteResolution, then reduce cellTranslation, then increase resolution
+        # (1000, 1000) >= (980, 980) + (20, 20)
+        self.MPPC.cellCompleteResolution.value = (1000, 1000)
+        self.MPPC.cellTranslation.value = \
+            tuple(tuple((20, 20) for i in range(0, self.MPPC.shape[0])) for i in range(0, self.MPPC.shape[1]))
+        self.EBeamScanner.resolution.value = (7840, 7840)  # -> eff cell size = (980, 980)
```
pieleric, line 466: The more I'm looking at it, the less I think it's going to be easy to write reliable client code. What is the rule? Are you able to write a function which takes are arguments resolution, cell translation, and cell complete resolution, and always succeeds in writing it?

Let's say it's currently at 7280, 1000 , 80 (like example 1), and you want to do example 3: 7840, 1000, 20. If you do it in the order of example 3, it works. But if you do in the order of example 1, it fails when attempting to set the resolution (because it'd see 1000 < 980 + 80).

Even more difficult is the case where things are not square (and admittedly, a little unlikely to happen in reality). Imagine it starts at 6400x7280, 900x1000, 50x80, and want to go to 7280x6400, 1000x900, 80x50. 
Do you need to follow the code that increase or decrease? I guess the only way would be to use intermediary values. Maybe first set the cell complete resolution to the max, then change the rest, and then apply the requested cell complete resolution.

pieleric (changes requested):
Please add a example function which can always set the values, whichever are the current values.

Seeing all these difficulties, I'm starting to appreciate more the behaviour of resolution/binning/translation on the DigitalCamera, which automatically adjust the values. I'd suggest to retry to consider doing this here too. Especially, I think it'd be possible to have a simple rule that wouldn't feel too weird: if the cell complete resolution would be too small, automatically increase it, to just fit.
With such rule, I _think_, it'd be possible to always set the values by following this order:
 resolution > cell translation > cell complete resolution.
If you wanted, you could even still put some check up code: in the cell complete resolution setter, it would raise an error if the value is not fitting (instead of picking a compatible value behind the back of the user).

What do you think?

Sabrina85:
> Please add a example function which can always set the values, whichever are the current values.
> 
> Seeing all these difficulties, I'm starting to appreciate more the behaviour of resolution/binning/translation on the DigitalCamera, which automatically adjust the values. I'd suggest to retry to consider doing this here too. Especially, I think it'd be possible to have a simple rule that wouldn't feel too weird: if the cell complete resolution would be too small, automatically increase it, to just fit.
> With such rule, I _think_, it'd be possible to always set the values by following this order:
> resolution > cell translation > cell complete resolution.
> If you wanted, you could even still put some check up code: in the cell complete resolution setter, it would raise an error if the value is not fitting (instead of picking a compatible value behind the back of the user).
> 
> What do you think?

Yes I agree. I am also not convinced by the current implementation. I thought of it more as an intermediate. I will rework it and try to implement it as you suggested.

PR #1750, opened 2021-04-21
FastEM acquisition model and fake acquire function

philipwinklerdelmic:
Specification of the gui/acqmng interface

### DIFF
--- a/src/odemis/acq/fastem.py
+++ b/src/odemis/acq/fastem.py
@@ -1,0 +1,78 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+Created on 19 Apr 2021
+
+Copyright © 2021 Philip Winkler, Delmic
+
+This file is part of Odemis.
+
+Odemis is free software: you can redistribute it and/or modify it under the terms
+of the GNU General Public License version 2 as published by the Free Software
+Foundation.
+
+Odemis is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+PURPOSE. See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with
+Odemis. If not, see http://www.gnu.org/licenses/.
+"""
+from odemis import model
+
+
+class FastEMROA(object):
+    """ Representation of a FastEM ROA (region of acquisition). """
+
+    def __init__(self, name, coordinates, roc):
+        """
+        :param name: (str) name of the ROA
+        :param coordinates: (float, float, float, float) l, t, r, b coordinates in m
+        :param roc: (FastEMROC) corresponding region of calibration
+        """
+        self.name = model.StringVA(name)
+        self.coordinates = model.TupleContinuous(coordinates,
+                                                 range=((-1, -1, -1, -1), (1, 1, 1, 1)),
+                                                 cls=(int, float),
+                                                 unit='m')
+        self.roc = model.VigilantAttribute(roc)
+
+
+class FastEMROC(object):
+    """ Representation of a FastEM ROC (region of calibration). """
+
+    def __init__(self, name, coordinates):
+        """
+        :param name: (str) name of the ROC
+        :param coordinates: (float, float, float, float) l, t, r, b coordinates in m
+        """
+        self.name = model.StringVA(name)
+        self.coordinates = model.TupleContinuous(coordinates,
+                                                 range=((-1, -1, -1, -1), (1, 1, 1, 1)),
+                                                 cls=(int, float),
+                                                 unit='m')
+        self.parameters = None  # calibration object with all relevant parameters
+
+
+# TODO: replace fake testing functions with actual acquisition
+import time
+# The executor is a single object, independent of how many times the module is loaded.
+_executor = model.CancellableThreadPoolExecutor(max_workers=1)
+def acquire(roa, path):
+    """
+    :param roa: (FastEMROA) acquisition region to be acquired
+    :param path: (str) path and filename of the acquisition on the server
+    :returns: (ProgressiveFuture): acquisition future
+    """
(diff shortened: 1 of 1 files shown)


src/odemis/acq/fastem.py
```diff
@@ -0,0 +1,65 @@
...
+                                                 range=((-1, -1, -1, -1), (1, 1, 1, 1)),
+                                                 cls=(int, float))
+        self.calibration = model.VigilantAttribute(calibration)  # ROC
+
+
+class FastEMROC(object):
+    """ Representation of a FastEM parameters region. """
+
+    def __init__(self, number, coordinates):
+        self.number = model.IntVA(number, readonly=True)  # cannot be changed after initialization
```
pieleric, line 39: What is this number for? Is that some sort of name? If so, I'd suggest to use .name = StringVA() and make it read-write. The GUI can enforce extra requirements on it, if you only want a number.

src/odemis/acq/fastem.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+class FastEMROA(object):
+    """ Representation of a FastEM acquisition region. """
+
+    def __init__(self, name, coordinates, calibration):
+        self.name = model.StringVA(name)
+        self.coordinates = model.TupleContinuous(coordinates,
+                                                 range=((-1, -1, -1, -1), (1, 1, 1, 1)),
+                                                 cls=(int, float))
```
pieleric, line 31: Is that physical units, or a ratio? If physical (which is the most likely), add `unit="m"`.

src/odemis/acq/fastem.py
```diff
@@ -0,0 +1,65 @@
...
+You should have received a copy of the GNU General Public License along with
+Odemis. If not, see http://www.gnu.org/licenses/.
+"""
+from odemis import model
+
+
+class FastEMROA(object):
+    """ Representation of a FastEM acquisition region. """
+
+    def __init__(self, name, coordinates, calibration):
```
pieleric, line 27: Add a docstring with the type and meanings. In particular, define the coordinates (left,top,bottom,right)?

Calibration: maybe better call it "roc".

src/odemis/acq/fastem.py
```diff
@@ -0,0 +1,65 @@
...
+PURPOSE. See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with
+Odemis. If not, see http://www.gnu.org/licenses/.
+"""
+from odemis import model
+
+
+class FastEMROA(object):
+    """ Representation of a FastEM acquisition region. """
```
pieleric, line 25: I think the class could be called just "ROA" as it's already in fastem. (But FastEMROA has the advantage of making it a little more explicit in the client code).

In the docstring, at least indicate what "ROA" stands for.

src/odemis/acq/fastem.py
```diff
@@ -0,0 +1,65 @@
...
+    def __init__(self, name, coordinates, calibration):
+        self.name = model.StringVA(name)
+        self.coordinates = model.TupleContinuous(coordinates,
+                                                 range=((-1, -1, -1, -1), (1, 1, 1, 1)),
+                                                 cls=(int, float))
+        self.calibration = model.VigilantAttribute(calibration)  # ROC
+
+
+class FastEMROC(object):
+    """ Representation of a FastEM parameters region. """
```
pieleric, line 36: Indicate what ROC stands for.

src/odemis/acq/fastem.py
```diff
@@ -0,0 +1,65 @@
...
+        self.parameters = None  # Calibration object with all relevant parameters
+
+
+# TODO: replace fake testing functions with actual acquisition
+import time
+executor = model.CancellableThreadPoolExecutor(max_workers=1)
+def acquire(roas, path):
+    """
+    :param roas: (list of FastEMROA) list of acquisition regions to be acquired
+    :param path: (str) path and filename of the acquisition on the server
```
pieleric, line 52: filename of which acquisition? They are not going to have all the same name, right? 
Shouldn't this be an attribute on the RoA?

src/odemis/acq/fastem.py
```diff
@@ -0,0 +1,65 @@
...
+        self.number = model.IntVA(number, readonly=True)  # cannot be changed after initialization
+        self.coordinates = model.TupleContinuous(coordinates,
+                                                 range=((-1, -1, -1, -1), (1, 1, 1, 1)),
+                                                 cls=(int, float))
+        self.parameters = None  # Calibration object with all relevant parameters
+
+
+# TODO: replace fake testing functions with actual acquisition
+import time
+executor = model.CancellableThreadPoolExecutor(max_workers=1)
```
pieleric, line 48: Maybe mark it as private as `_executor`. I'd also mention in a comment that it's single object, independently of how many times the module is loaded.

src/odemis/acq/fastem.py
```diff
@@ -0,0 +1,65 @@
...
+        self.coordinates = model.TupleContinuous(coordinates,
+                                                 range=((-1, -1, -1, -1), (1, 1, 1, 1)),
+                                                 cls=(int, float))
+        self.calibration = model.VigilantAttribute(calibration)  # ROC
+
+
+class FastEMROC(object):
+    """ Representation of a FastEM parameters region. """
+
+    def __init__(self, number, coordinates):
```
pieleric, line 38: Please add a docstring about the parameters type and meaning.

pieleric (changes requested):

src/odemis/acq/fastem.py
philipwinklerdelmic: Sorry, I forgot to update this. The function should take a single ROA, not a list. 

Indeed, we could consider making the path an attribute. The folder structure is <date>/<project_name>/<ROA name>. The ROA name is already an attribute. We need some extra information for the date and project, it could be stored in a second VA `.path` or `.filename`. Do you think that's better than passing it as an additional argument?

Sabrina85 (approved):

PR #1751, opened 2021-04-21
Enzel acquisition panel

Mahmood-B:
This pull request consists of two commits. One includes the changes in the xrc files, which are related to adding the widgets to the GUI. And the second commit includes the changes related to the control of those widgets. 
The acquisition panel is still under development, because this PR does not include the zstack acquisition capability.


src/odemis/gui/cont/acquisition.py
```diff
@@ -450,6 +450,455 @@ def open_acquisition_dialog(self):
...
+            # update the counter of the filename
+            self._config.fn_count = update_counter(self._config.fn_count)
+            # reset the storing directory to the project directory
+            self._config.last_path = self._config.pj_last_path
+        elif state == "FAILED":
+            self._panel.txt_cryosecom_est_time.Show()
+            self._panel.txt_cryosecom_est_time.SetLabel(text)
+        elif state == "CANCELED":
+            self._panel.txt_cryosecom_est_time.Show()
+            self._update_acquisition_time()
```
pieleric, line 633: Add a `else: raise ValueError(...)`. This way, if you make a typo in the state, it's easier to notice.
Also, to make the state even less prone to typos, you could use constants like `ST_FINISHED = "FINISHED"`.

pieleric (changes requested):

src/odemis/gui/cont/acquisition.py
```diff
@@ -450,6 +450,455 @@ def open_acquisition_dialog(self):
...
+        Called when the acquisition process is
+        done, failed or canceled
+        """
+        self._acq_future = None
+        self._gauge_future_conn = None
+        self._tab_data.main.is_acquiring.value = False
+
+        # try to get the acquisition data
+        try:
+            data, exp = future.result()
```
lazem, line 587: No use for `data, exp` -> `_, _`

src/odemis/gui/cont/acquisition.py
```diff
@@ -450,6 +450,455 @@ def open_acquisition_dialog(self):
...
+        self._panel.btn_cryosecom_acquire.Enable()
+        self._panel.btn_cryosecom_change_file.Enable()
+        self._panel.btn_acquire_overview.Enable()
+        self._panel.z_stack_chkbox.Enable()
+        self._panel.streams_chk_list.Enable()
+        self._panel.txt_cryosecom_left_time.Hide()
+        self._tab.streambar_controller.resume()
+        self._panel.Layout()
+
+        if state == "FINISHED":
```
lazem: As Eric suggested, add the strings `"FINISHED", "FAILED"`..etc as constants and use the constants instead.

src/odemis/gui/cont/tabs.py
```diff
@@ -639,6 +629,16 @@ def get_display_priority(cls, main_data):
         else:
             return None
 
+    @call_in_wx_main
+    def display_acquired_data(self, data):
```
lazem, line 635: Please add a docstring for the type of `data` variable.

src/odemis/gui/cont/acquisition.py
```diff
@@ -450,6 +450,455 @@ def open_acquisition_dialog(self):
...
+            "histogram",
+            "is_active",
+            "should_update",
+            "status",
+            "name",
+            "tint",
+        )
+        self._acq_future = None
+        self._executor = futures.ThreadPoolExecutor(max_workers=2)
+        self._scheduled_future = None
```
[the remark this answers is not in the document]
Mahmood-B: indeed i didn't pay attention that it's not required. Following the same logic, we also do not need reference to the thread pool executor

src/odemis/util/comp.py
```diff
@@ -108,3 +108,4 @@ def get_fov_rect(comp, fov):
             center[1] - fov[1] / 2,  # top
             center[0] + fov[0] / 2,  # right
             center[1] + fov[1] / 2)  # bottom
+
```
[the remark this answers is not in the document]
Mahmood-B: There was the signature of the function of generating zlevels. But then i removed it. I thought once it's fully implemented i should add it

src/odemis/gui/main.xrc
```diff
@@ -2970,667 +2971,635 @@
     <bg>#333333</bg>
     <style>wxWANTS_CHARS</style>
   </object>
-  <object class="wxPanel"
```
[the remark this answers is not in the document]
Mahmood-B: No actually, not because of XRCed. I reformatted the file. There are gaps in code, and there are several parts of code which are improperly formatted. Now I have put back the older version (i hope i did so correctly), but i would still suggest to nicely format it, maybe from the master branch directly.

src/odemis/gui/cont/acquisition.py
```diff
@@ -449,6 +453,432 @@ def open_acquisition_dialog(self):
...
+
+    @call_in_wx_main
+    def _on_streams(self, streams):
+        """
+        a VA callback that is called when the .streams VA is changed
+        """
+        # for every entry in the list, is it also present in the .streams?
+        for i in range(self._panel.streams_chk_list.GetCount() - 1, -1, -1):
+            item_stream = self._panel.streams_chk_list.GetClientData(i)
+            if item_stream not in streams:
```
lazem, line 688: `streams` could be the overview map streams and iterating over them would mess up acquisition panel list. So make sure it's the _right_ streams.

src/odemis/gui/cont/tabs.py
```diff
@@ -639,6 +629,16 @@ def get_display_priority(cls, main_data):
...
+        # get the top right view port
+        views = self.tab_data_model.views.value[1]
+        for s in views.getStreams():
+            views.removeStream(s)
+        streams = data_to_static_streams(data)
+        for s in streams:
+            views.addStream(s)
```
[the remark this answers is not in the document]
Mahmood-B: It doesn't work this way. Because, the function "addStream" of the StreamBarController class adds the passed stream to the .streams VA. This leads to adding every stream twice in the checklist. I looked at the implementation of the function addStream , and reused only the line of code that adds the steam to the ACQUIRED panel, which is creating ab object of StreamController. So like this:
`stream_cont = StreamController(self._panel.pnl_cryosecom_acquired, s, self._tab_data_model,
                                           show_panel=True, view=view, sb_ctrl=self._overview_stream_controller)
stream_cont.stream_panel.collapse(True)`

lazem (changes requested):
Looks good overall so far. I have few comments in addition to Eric's, the main issue is that acquiring overview could mess up the acquisition panel list. I also believe that me and you should coordinate on how to handle the overview map streams properly.

src/odemis/gui/cont/acquisition.py
```diff
@@ -450,6 +450,455 @@ def open_acquisition_dialog(self):
...
+        self._panel.btn_cryosecom_acqui_cancel.Hide()
+        self._panel.Layout()
+        # limit number of characters
+        self._panel.param_Zmin.SetMaxLength(8)
+        self._panel.param_Zmax.SetMaxLength(8)
+        self._panel.param_Zstep.SetMaxLength(8)
+        # set the default values
+        self._panel.param_Zmax.SetValue("10 µm")
+        self._panel.param_Zmin.SetValue("-10 µm")
+        self._panel.param_Zstep.SetValue("1 µm")
```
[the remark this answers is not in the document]
Mahmood-B: it's kinda a bit too late to tell me this now. I already spent like 2 days last week on filtering the input text from the user and got something working :(  . It is not mentioned in the specs though, nor you mentioned it in our meetings.

src/odemis/gui/cont/acquisition.py
```diff
@@ -450,6 +450,455 @@ def open_acquisition_dialog(self):
...
+        else:  # the user changed the wavelength
+            counts = self._panel.streams_chk_list.GetCount()
+            streams = [
+                self._panel.streams_chk_list.GetClientData(i) for i in range(counts)
+            ]
+            for i, s in zip(range(counts), acqmng.sortStreams(streams)):
+                # replace the unsorted streams with the sorted streams one by one
+                self._panel.streams_chk_list.Delete(i)
+                self._panel.streams_chk_list.Insert(s.name.value, i, s)
+                self._panel.streams_chk_list.Check(i)
```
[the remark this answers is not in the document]
Mahmood-B: i thought because changing the stream name leads to changing the wavelengths, I thought it's more compact to combine the similar parts of code in one function.

src/odemis/gui/cont/tabs.py
```diff
@@ -632,13 +634,11 @@ def get_display_priority(cls, main_data):
...
-        for s in views.getStreams():
-            views.removeStream(s)
-        streams = data_to_static_streams(data)
-        for s in streams:
-            views.addStream(s)
-
+        view = self.tab_data_model.views.value[1]
+        for s in data_to_static_streams(data):
+            stream_cont = StreamController(self._panel.pnl_cryosecom_acquired, s, self._tab_data_model,
+                                           show_panel=True, view=view, sb_ctrl=self._overview_stream_controller)
```
pieleric, line 640: What's the reason you're explicitly creating the StreamController, instead of calling the _overview_stream_controller.addStream() ? Is that the StaticStream issue that Bassim mentioned? Or is there more reason? Ideally, using the addStream() function would be better.

src/odemis/gui/cont/tabs.py
```diff
@@ -345,6 +345,8 @@ def __init__(self, name, button, panel, main_frame, main_data):
         """
 
         tab_data = guimod.LocalizationGUIData(main_data)
+        self._tab_data_model = tab_data
+        self._panel = panel
```
pieleric, line 349: That shouldn't be needed. The super() already stores these values in .tab_data_model and .panel.

src/odemis/gui/cont/acquisition.py
```diff
@@ -761,34 +733,39 @@ def _update_acquisition_time(self):
...
+        """
+        counts = self._panel.streams_chk_list.GetCount()
+        streams = [
+            self._panel.streams_chk_list.GetClientData(i) for i in range(counts)
+        ]
+        for i, s in zip(range(counts), acqmng.sortStreams(streams)):
+            # replace the unsorted streams with the sorted streams one by one
+            self._panel.streams_chk_list.Delete(i)
+            self._panel.streams_chk_list.Insert(s.name.value, i, s)
+            self._panel.streams_chk_list.Check(i)
```
pieleric, line 749: What does "Check()"? Does it toggle the stream? If so, that shouldn't always be the case. Only if it's in acquisitionStreams.

src/odemis/gui/cont/acquisition.py
```diff
@@ -761,34 +733,39 @@ def _update_acquisition_time(self):
...
+        streams = [
+            self._panel.streams_chk_list.GetClientData(i) for i in range(counts)
+        ]
+        stream_names = [s.name.value for s in streams]
+        for name, i, s in zip(chcklist_names, range(counts), streams):
+            if name not in stream_names:
                 self._panel.streams_chk_list.Delete(i)
-                self._panel.streams_chk_list.Insert(s.name.value, i, s)
+                self._panel.streams_chk_list.Insert(stream_name, i, s)
                 self._panel.streams_chk_list.Check(i)
```
pieleric: I think you can make it a lot clearer by just iterating over all the streams once, and checking one-by-one that it's still the right name. Also, I think it's possible to use SetString() to just change the name, instead of deleting+inserting.
```python
counts = self._panel.streams_chk_list.GetCount()
for i in range(counts):
    sname = self._panel.streams_chk_list.GetString(i)
    s = self._panel.streams_chk_list.GetClientData(i)
    if s.name.value != sname:
           self._panel.streams_chk_list.SetString(i, s.name.value)
```

pieleric (changes requested):

src/odemis/gui/cont/tabs.py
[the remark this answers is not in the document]
pieleric: I now understand. Thanks for the explanation. As Bassim said, it's better to handle it by distinguishing StaticStreams from the "live" streams in .streams. Then you can call .addStream() normally. 

Concerning the behaviour of the view, for now, let's keep it simple: every new stream is added to the view and shown. 
Later, part of the "feature bookmark" work by Bassim, the behaviour will be fine-tuned: the view will only show the streams acquired at the current location. From the point of view of the acquisition manager, that doesn't change anything, so you don't need to worry.

src/odemis/gui/cont/tabs.py
pieleric: I now see your previous answer explained it already.

src/odemis/gui/cont/acquisition.py
```diff
@@ -449,6 +453,432 @@ def open_acquisition_dialog(self):
...
+        thumb = acqmng.computeThumbnail(st, future)
+        data, _ = future.result()
+        filename = self._filename.value
+        if data:
+            exporter = dataio.get_converter(self._config.last_format)
+            exporter.export(filename, data, thumb)
+            logging.info(u"Acquisition saved as file '%s'.", filename)
+            # update the filename
+            self._filename.value = create_filename(
+                self._config.last_path,
```
pieleric, line 670: The filename should be generated the same way as in the model: with `pj_last_path`.

src/odemis/gui/cont/acquisition.py
pieleric: Actually, you should use it.
If exp is not None, it means not all acquisition succeeded. At least, you should log a error. Possibly, you'd change the Label to "Acquisition partially failed" instead of "Saving file...". 
You could also use `data`: instead of passing `future` to the `export_data`, you'd pass directly `data` and the thumbnail. I think it'd actually help to make the code clearer too.

pieleric (changes requested):

src/odemis/gui/cont/acquisition.py
Mahmood-B: Good catch

src/odemis/gui/cont/acquisition.py
Mahmood-B: indeed simpler and more compact

src/odemis/gui/cont/acquisition.py
```diff
@@ -491,46 +487,52 @@ def __init__(self, tab_data, panel, tab):
...
+        _ = VigilantAttributeConnector(
+            va=self._tab_data.zMax,
+            value_ctrl=self._panel.param_Zmax,
+            events=wx.EVT_COMMAND_ENTER,
+        )
+        _ = VigilantAttributeConnector(
+            va=self._tab_data.zStep,
+            value_ctrl=self._panel.param_Zstep,
+            events=wx.EVT_COMMAND_ENTER,
+        )
```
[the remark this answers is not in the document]
Mahmood-B: I'm really not sure if this might happen. The connections are done in the constructor of the VigilantAttributeConnector class, using references to tab_data and panel which are both members of the Cryo acqui controller class, so the connections should stay working in the scope of the cryo acqui controller. In addition to that, it has been tested several times so far, and the VA connections seem to work fine. So i don't think the python interpreter will destroy these connections.

Mahmood-B:
There is still issue related more to the overview but has effect on the acquisition panel. After the overivew image is acquired, any attempt of removing one of the existing streams will cause the following error:
```
16:33:33 ERROR   main           : Traceback (most recent call last):
  File "/home/barazi/development/odemis/src/odemis/gui/comp/stream.py", line 1485, in on_stream_remove
    pub.sendMessage("stream.remove", stream=st)
  File "/usr/lib/python3/dist-packages/wx/lib/pubsub/core/kwargs/publisher.py", line 27, in sendMessage
    topicObj.publish(**kwargs)
  File "/usr/lib/python3/dist-packages/wx/lib/pubsub/core/kwargs/publishermixin.py", line 24, in publish
    self._publish(msgKwargs)
  File "/usr/lib/python3/dist-packages/wx/lib/pubsub/core/topicobj.py", line 376, in _publish
    self.__sendMessage(data, self, iterState)
  File "/usr/lib/python3/dist-packages/wx/lib/pubsub/core/topicobj.py", line 397, in __sendMessage
    self._mix_callListener(listener, data, iterState)
  File "/usr/lib/python3/dist-packages/wx/lib/pubsub/core/kwargs/publishermixin.py", line 64, in _mix_callListener
    listener(iterState.filteredArgs, self, msgKwargs)
  File "/usr/lib/python3/dist-packages/wx/lib/pubsub/core/kwargs/listenerimpl.py", line 42, in __call__
    self._calledWhenDead()
  File "/usr/lib/python3/dist-packages/wx/lib/pubsub/core/listenerbase.py", line 91, in _calledWhenDead
    raise RuntimeError('BUG: Dead Listener called, still subscribed!')
RuntimeError: BUG: Dead Listener called, still subscribed!
```

lazem:
> There is still issue related more to the overview but has effect on the acquisition panel. After the overivew image is acquired, any attempt of removing one of the existing streams will cause the following error:
> 
> ```
> 16:33:33 ERROR   main           : Traceback (most recent call last):
>   File "/home/barazi/development/odemis/src/odemis/gui/comp/stream.py", line 1485, in on_stream_remove
>     pub.sendMessage("stream.remove", stream=st)
>   File "/usr/lib/python3/dist-packages/wx/lib/pubsub/core/kwargs/publisher.py", line 27, in sendMessage
>     topicObj.publish(**kwargs)
>   File "/usr/lib/python3/dist-packages/wx/lib/pubsub/core/kwargs/publishermixin.py", line 24, in publish
>     self._publish(msgKwargs)
>   File "/usr/lib/python3/dist-packages/wx/lib/pubsub/core/topicobj.py", line 376, in _publish
>     self.__sendMessage(data, self, iterState)
>   File "/usr/lib/python3/dist-packages/wx/lib/pubsub/core/topicobj.py", line 397, in __sendMessage
>     self._mix_callListener(listener, data, iterState)
>   File "/usr/lib/python3/dist-packages/wx/lib/pubsub/core/kwargs/publishermixin.py", line 64, in _mix_callListener
>     listener(iterState.filteredArgs, self, msgKwargs)
>   File "/usr/lib/python3/dist-packages/wx/lib/pubsub/core/kwargs/listenerimpl.py", line 42, in __call__
>     self._calledWhenDead()
>   File "/usr/lib/python3/dist-packages/wx/lib/pubsub/core/listenerbase.py", line 91, in _calledWhenDead
>     raise RuntimeError('BUG: Dead Listener called, still subscribed!')
> RuntimeError: BUG: Dead Listener called, still subscribed!
> ```

This should be solved with this PR now:
https://github.com/delmic/odemis/pull/1763

src/odemis/gui/main.xrc
```diff
@@ -5483,19 +5451,362 @@
                         <bg>#555555</bg>
                       </object>
                       <object class="FoldPanelItem">
-                        <object class="StreamBar" name="pnl_overview_streams">
```
Mahmood-B, line 5486: Just for your convenience, the xrc file changes related to the acquisition panel start form this line till line 5809

src/odemis/gui/cont/acquisition.py
```diff
@@ -449,6 +453,433 @@ def open_acquisition_dialog(self):
...
+        acq_time = acqmng.estimateTime(self._acquiStreams.value)
+        acq_time = math.ceil(acq_time)  # round a bit pessimistic
+        # TODO take into account the time of acquiring the zlevels
+        # as well as the time required for moving the actuator
+        txt = u"Estimated time: {}.".format(units.readable_time(acq_time, full=False))
+        self._panel.txt_cryosecom_est_time.SetLabel(txt)
+
+    @call_in_wx_main
+    def _on_stream_wavelength(self, _):
+        """
```
lazem, line 744: Maybe you can call this when you add a new stream to the checklist, that way it would be put in its right place?

src/odemis/gui/cont/acquisition.py
```diff
@@ -449,6 +453,433 @@ def open_acquisition_dialog(self):
...
+        counts = self._panel.streams_chk_list.GetCount()
+        streams = [
+            self._panel.streams_chk_list.GetClientData(i) for i in range(counts)
+        ]
+        for i, s in zip(range(counts), acqmng.sortStreams(streams)):
+            # replace the unsorted streams with the sorted streams one by one
+            self._panel.streams_chk_list.Delete(i)
+            self._panel.streams_chk_list.Insert(s.name.value, i, s)
+            if s in self._acquiStreams.value:
+                self._panel.streams_chk_list.Check(i)
```
lazem, line 757: It seems the function clears the checklist and repopulates it with the sorted live streams, so maybe it'd be more readable to write it like this:

        self._panel.streams_chk_list.Clear()
        live_streams = [s for s in self._tab_data.streams.value if isinstance(s, LiveStream)]
        for i, stream in enumerate(acqmng.sortStreams(live_streams)):
	        self._panel.streams_chk_list.Insert(stream.name.value, i, stream)
                if stream in self._acquiStreams.value:
		        self._panel.streams_chk_list.Check(i)

src/odemis/gui/cont/acquisition.py
```diff
@@ -647,7 +647,10 @@ def _display_acquired_data(self, data):
         Shows the acquired image on the view
         """
         # get the localization tab
-        local_tab = self._tab_data.main.getTabByName("cryosecom-localization")
+        try:
+            local_tab = self._tab_data.main.getTabByName("cryosecom-localization")
+        except LookupError:
+            logging.warning("Localization tab was not found")
```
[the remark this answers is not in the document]
Mahmood-B: Aha, i haven't noticed that

src/odemis/gui/cont/acquisition.py
Mahmood-B: Good point :ok_hand:

src/odemis/gui/cont/acquisition.py
Mahmood-B: Using `GetClientData` is interesting because it returns the stream itself, not only its name as a string. In addition, if we use `GetClientData`, we do not need to check if the stream is `LiveStream`

src/odemis/gui/cont/acquisition.py
lazem: But don't call `_on_stream_wavelength`, create a function `sort_streams` or something, and call it from here and on addition.

Mahmood-B: Makes sense

Mahmood-B:
@Kleijwegt, please focus only on the last commit, and ignore the others as they are already reviewed.

src/odemis/acq/acqmng.py
```diff
@@ -83,6 +84,197 @@ def acquire(streams, settings_obs=None):
     return future
 
 
+def acquireZStack(streams, zlevels, settings_obs=None):
```
Kleijwegt, line 87: I think the docstring format should include some empty lines. You are using the reStructuredText format here  (my favorite). Can you follow that format here?

https://queirozf.com/entries/python-docstrings-reference-examples

lazem (changes requested):

src/odemis/gui/cont/tabs.py
```diff
@@ -665,6 +655,19 @@ def get_display_priority(cls, main_data):
...
+        """
+        Displays the acquired streams on the top right view
+        data (DataArray): the images/data acquired 
+        """
+        # get the top right view port
+        view = self.tab_data_model.views.value[1]
+        for s in data_to_static_streams(data):
+            stream_cont = StreamController(self.panel.pnl_cryosecom_acquired, s, self.tab_data_model,
+                                           show_panel=True, view=view, sb_ctrl=self._overview_stream_controller)
+            stream_cont.stream_panel.collapse(True)
```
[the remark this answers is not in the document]
Mahmood-B: This is confusing. It might be correct what you are saying, but Eric told me to only add the acquired stream to the acquired streams panel as you will be working here further. So either you take care of this, or tell me the big picture as well as details of what you want to do here, because this part is all not mentioned in the specifications.

src/odemis/gui/cont/acquisition.py
```diff
@@ -693,7 +693,7 @@ def _on_streams_change(self, streams):
         """
         # removing a stream 
         # for every entry in the list, is it also present in the .streams?
-        for i in range(self._panel.streams_chk_list.GetCount() - 1, -1, -1):
+        for i in range(self._panel.streams_chk_list.GetCount() - 1):
```
pieleric: No, the reverse was doing something. It ensures that the index you have and the current list index match.
For example If you have a list with 5 streams [A, B, C, D, E], and stream C and D disappeared, when you iterate in normal order, you first delete item 2 (OK, the list becomes [A, B, D, E]), and then item 3... but item 3 is E!

If you delete in reverse order, you don't have this issue.
=> Put back the reverse order, and add a comment explaining this.
Note that typically, when the streams are removed by the user, only one stream is removed at a time, so such error is not triggered. This will happen when the streams are deleted by some code, for instance when changing project.

src/odemis/gui/cont/acquisition.py
```diff
@@ -449,6 +455,444 @@ def open_acquisition_dialog(self):
...
+        new_filename = ShowAcquisitionFileDialog(self._panel, current_filename)
+        if new_filename is not None:
+            self._filename.value = new_filename
+            self._config.fn_ptn, self._config.fn_count = guess_pattern(new_filename)
+            logging.debug("Generated filename pattern '%s'", self._config.fn_ptn)
+
+    def _get_wavelength_vas(self, stream_item):
+        """
+        Gets the excitation and emission VAs of a stream.
+        return (set of VAs)
```
pieleric, line 867: Add a docstring indicating the "stream_item"/"item_stream"/"st" parameter type.

src/odemis/gui/cont/acquisition.py
```diff
@@ -449,6 +453,433 @@ def open_acquisition_dialog(self):
...
+        """
+        current_filename = self._filename.value
+        new_filename = ShowAcquisitionFileDialog(self._panel, current_filename)
+        if new_filename is not None:
+            self._filename.value = new_filename
+            self._config.fn_ptn, self._config.fn_count = guess_pattern(new_filename)
+            logging.debug("Generated filename pattern '%s'", self._config.fn_ptn)
+
+    def _get_wavelength_vas(self, stream_item):
+        """
```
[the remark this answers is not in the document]
pieleric: I'd say that "item" is not really necessary. You could just name it "stream" (like the next function, but that overrides a module name, which is not great), or maybe just "st".

src/odemis/gui/cont/tabs.py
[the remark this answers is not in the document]
pieleric: Sorry for badly communicating it. @lazem , my idea is that Mahmoud just puts "rough" code, and you'll later put all the proper code. It's too much fluctuating right now.
@Mahmood-B , could you just add a `# TODO adjust this code to fit the Feature behaviour` at the top.

pieleric (changes requested):
I think there are just some minor things to fix, and then we can merge it.
Please have a look at the comments and then rebase against master (to fix the conflicts against the XRC file). It should then be ready for merge.

PR #1759, opened 2021-05-03
plugins ar_spectral: take into account readout rate for time estimation

pieleric:
The other streams already do it, but in this plugin we were taking shortcuts
and assumed that the CCD acquisition == exposure time.
However, with slow readout rates, and small exposure times, the readout
can take many time the exposure time. This caused the acquisition to
fail, because it timed out too early.

=> Use the same formula as the other streams.

### DIFF
--- a/plugins/ar_spectral.py
+++ b/plugins/ar_spectral.py
@@ -202,6 +202,26 @@ class SpectralARScanStream(stream.Stream):
             yres = 1
         return int(xres), int(yres)
 
+    def _estimateCCDTime(self):
+        """
+        Estimate the time it will take for measurement at one ebeam position
+        """
+        # Each pixel = the exposure time (of the detector) + readout time +
+        # 10ms overhead + 20% overhead
+
+        binning = (self.binninghorz.value, self.binningvert.value)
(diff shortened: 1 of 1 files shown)


plugins/ar_spectral.py
```diff
@@ -202,24 +202,45 @@ def get_scan_res(self):
...
 
+    def _estimateCCDTime(self):
+        """
+        Estimate the time it will take for measurement at one ebeam position
+        """
+        # Each pixel = the exposure time (of the detector) + readout time +
+        # 10ms overhead + 20% overhead
+
+        binning = (self.binninghorz.value, self.binningvert.value)
+        res = [s / b for s, b in zip(self._detector.resolution.range[1], binning)]
```
VictoriaDel, line 213: Is res the resolution value of the detector? Why don't we get self._detector.resolution.value? I thought it's calculated from the zip(detector.shape/binning). I know estimateAcquisitionTime is called before run acquisition but why don't we calculate the self._detector.resolution.value earlier? and if needed to be updated every time the binning changes, we can take care of it by subscribing to update the resolution.

VictoriaDel (commented):
Thanks for the explanation

PR #1760, opened 2021-05-04
Testcase for pmt: update time delay before checking that the pmt protection is activated

VictoriaDel:


src/odemis/driver/test/pmtctrl_test.py
```diff
@@ -301,8 +301,8 @@ def test_while_acquiring(self):
 
         # Immediately after the acquisition is done, PMT protection remains off
         self.assertEqual(self.control.protection.value, False)
-        # Protection is reset with some delay after acquisition is done
-        time.sleep(0.2)
+        # Protection is reset with some delay after all (10) acquisitions are completed
```
pieleric, line 304: No, the protection should be set after 0.1s, so 0.2 s is enough.
The problem is more that `self.acq_complete.wait()` doesn't work, because it's already set. What you need to do is to `        self.acq_complete.clear()` just before subscribing, line 297.

pieleric (changes requested):

PR #1761, opened 2021-05-05
FastEM Acquisition tab

philipwinklerdelmic:
A first version of the FastEM acquisition tab.

To get an idea of the functionality, I would suggest to pull the code and try it out yourself. You can start the GUI with the yaml file that I added.

The specification can be found in dropbox: sonic/software/specifications/SW-specification_gui_acquisition_tab_v01-02.docx. It contains a few remarks and motivations which might be helpful while reviewing the code


src/odemis/gui/comp/stream.py
```diff
@@ -1582,3 +1582,534 @@ def _set_warning(self):
...
+
+        # Calibration Grid
+        nrows = len(layout)
+        ncols = max([len(row) for row in layout])
+        calgrid_sz = wx.GridBagSizer(nrows, ncols)
+        for row_idx, row in enumerate(layout):
+            for col_idx, elem in enumerate(row):
+                subsz = wx.BoxSizer(wx.HORIZONTAL)
+                btn = wx.Button(self, wx.ALL | wx.ALIGN_CENTER, label="?", size=(30, 30))
+                btn.SetBackgroundColour("#999999")
```
pieleric, line 2103: It seems this colour keeps being reused. I think we should add this colour to the constants in odemis.gui. Or, alternatively, use one from there. It seems it could be called something like FG_COLOUR_....

src/odemis/gui/cont/acquisition.py
```diff
@@ -927,6 +928,203 @@ def on_file_export_done(self, future):
...
+
+    def check_acquire_button(self):
+        self.btn_acquire.Enable(self.roa_count and not self._get_undefined_calibrations())
+
+    @wxlimit_invocation(1)  # max 1/s
+    def update_acquisition_time(self):
+        lvl = None  # icon status shown
+        if self.roa_count == 0:
+            lvl = logging.WARN
+            txt = "No region of interest selected."
```
pieleric, line 1009: -> "region of acquisition"?

src/odemis/gui/cont/acquisition.py
```diff
@@ -927,6 +928,203 @@ def on_file_export_done(self, future):
...
+        elif self._get_undefined_calibrations():
+            lvl = logging.WARN
+            txt = "Calibration regions %s missing." % (self._get_undefined_calibrations(),)
```
pieleric, line 1012: That will look like "Calibration regions [2, 7] missing.", right? It's a little ugly. You could either use `", ".join()`, or just indicate the number of regions missing?

src/odemis/gui/cont/acquisition.py
```diff
@@ -927,6 +928,203 @@ def on_file_export_done(self, future):
...
+        """
+        undefined = set()
+        for p in self._tab_data_model.projects.value:
+            for roa in p.roas.value:
+                roc = roa.roc.value
+                if roc.coordinates.value == stream.UNDEFINED_ROI:
+                    undefined.add(roc.name.value)
+        return sorted(undefined)
+
+    def update_roa_count(self):
```
pieleric, line 1040: Please mark it as private, like the other methods.

src/odemis/gui/cont/acquisition.py
```diff
@@ -927,6 +928,203 @@ def on_file_export_done(self, future):
...
+        self._show_status_icons(None)
+
+        self.gauge_acq.Range = self.roa_count
+        self.gauge_acq.Value = 0
+
+        # Acquire ROAs for all projects
+        for p in self._tab_data_model.projects.value:
+            ppath = os.path.join(self.path, p.name.value)
+            for roa in p.roas.value:
+                fn = os.path.join(ppath, roa.name.value)
```
pieleric, line 1087: Is there a place where you ensure that projects have different names and that RoAs of a project have different names? Otherwise the user could end up with loosing data if mistakenly picking up the same name.

install/linux/usr/share/odemis/sim/fastem-sim.yaml
```diff
@@ -0,0 +1,59 @@
+# Minimal configuration file to start the Fast-EM GUI
+FASTEM-sim: {
+    class: Microscope,
+    role: fast-em,
```
pieleric, line 4: I'm fine with fast-em, but I think "mbsem" was used so far. Can you change, or was it agreed with Sabrina that the role will change?

src/odemis/gui/cont/streams.py
```diff
@@ -2990,3 +2992,356 @@ def _updateScanStage(self, use):
...
+
+        # For ROA creation process
+        self._current_roa_ctrl = None
+
+    def _on_text(self, evt):
+        txt = evt.GetString()
+        logging.debug("Renaming project from %s to %s.", self.model.name.value, txt)
+        self.model.name.value = txt
+
+    def _on_btn_roi(self, _):
```
pieleric, line 3102: roi -> roa

src/odemis/gui/cont/streams.py
```diff
@@ -2990,3 +2992,356 @@ def _updateScanStage(self, use):
...
+        # In the callback to the ROI, the ROI creation will be completed or aborted.
+        self._project_bar.enable_buttons(False)
+
+        # Deactivate all ROAs
+        for roa_ctrl in self.roa_ctrls.values():
+            roa_ctrl.overlay.active.value = False
+
+        # Minimum index that has not yet been deleted
+        num = next(idx for idx, n in enumerate(sorted(self.roa_ctrls.keys()) + [0], 1) if idx != n)
+        name = "ROI %s" % num
```
pieleric, line 3114: Can the user rename ROAs? If so, you need to check that no ROA has such name.

pieleric, line 3114: ROI -> ROA.

src/odemis/gui/cont/streams.py
```diff
@@ -2990,3 +2992,356 @@ def _updateScanStage(self, use):
...
+        # further processing. The process can still be aborted by clicking in the viewport without dragging.
+        # In the callback to the ROI, the ROI creation will be completed or aborted.
+        self._project_bar.enable_buttons(False)
+
+        # Deactivate all ROAs
+        for roa_ctrl in self.roa_ctrls.values():
+            roa_ctrl.overlay.active.value = False
+
+        # Minimum index that has not yet been deleted
+        num = next(idx for idx, n in enumerate(sorted(self.roa_ctrls.keys()) + [0], 1) if idx != n)
```
pieleric, line 3113: I think I see how it works, but it's quite hard to read. Maybe add "find the first index which is not in the existing indices".

src/odemis/gui/cont/streams.py
```diff
@@ -2990,3 +2992,356 @@ def _updateScanStage(self, use):
...
+    def _find_closest_scintillator(self, coordinates):
+        """
+        Given coordinates coords, find the closest scintillator.
+        coordinates (float, float, float, float): l, t, r, b coordinates in m
+        """
+        roi_x, roi_y = (coordinates[2] + coordinates[0]) / 2, (coordinates[1] + coordinates[3]) / 2
+        mindist = 1  # distances always lower 1
+        closest = None
+        for num, (sc_x, sc_y) in self._tab_data.scintillator_positions.items():
+            # scintillators are quadratic, use maximum instead of euclidean distance
```
pieleric, line 3180: quadratic -> rectangular?

src/odemis/gui/cont/streams.py
```diff
@@ -2990,3 +2992,356 @@ def _updateScanStage(self, use):
...
+        # Remove model
+        self.model.roas.value.remove(roa_ctrl.model)
+
+        # Destroy ROAController object
+        del roa_ctrl
+
+    def _find_closest_scintillator(self, coordinates):
+        """
+        Given coordinates coords, find the closest scintillator.
+        coordinates (float, float, float, float): l, t, r, b coordinates in m
```
pieleric, line 3174: Indicate the return value/type

src/odemis/gui/cont/streams.py
```diff
@@ -2990,3 +2992,356 @@ def _updateScanStage(self, use):
...
+            # Add the ROA model to project model
+            self.model.roas.value.append(roa_ctrl.model)
+
+            # Callback to ROI remove button
+            roa_ctrl.panel.btn_remove.Bind(wx.EVT_BUTTON, lambda evt: self._on_btn_remove(evt, roa_ctrl))
+
+        # Enable buttons of project bar
+        self._project_bar.enable_buttons(True)
+
+    def remove_roa_ctrl(self, roa_ctrl):
```
pieleric, line 3152: Indicate that it should be called from the main GUI thread.

src/odemis/gui/cont/streams.py
```diff
@@ -2990,3 +2992,356 @@ def _updateScanStage(self, use):
...
+
+        # Get ROC model (exists already in tab data) and change coordinates
+        self.model = tab_data.calibration_regions.value[number]
+        self.model.coordinates.value = (l, t, r, b)
+        self.model.coordinates.subscribe(self._on_coordinates)
+
+        # Add overlay
+        self.overlay = view_ctrl.viewports[0].canvas.add_calibration_overlay(self.model.coordinates, number)
+
+    def zoom_in(self):
```
pieleric, line 3334: We have other similar behaving functions which are called "fitViewToContent". So I'd suggest to name it the same or something similar, like "fit_view_to_bbox".

pieleric (changes requested):

src/odemis/gui/comp/overlay/world.py
```diff
@@ -2662,3 +2663,253 @@ def draw(self, ctx, shift=(0, 0), scale=1.0):
...
+            return False
+
+    def draw(self, ctx, shift=(0, 0), scale=1.0, line_width=None, dash=True):
+        """ Draw the selection as a rectangle. Exactly the same as parent function except that
+         it has an adaptive line width (wider if the overlay is active). """
+        line_width = 5 if self.active.value else 2
+        super(FastEMROAOverlay, self).draw(ctx, shift, scale, line_width, dash)
+
+
+class FastEMROCOverlay(WorldSelectOverlay):
```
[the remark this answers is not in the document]
philipwinklerdelmic: I'm not sure about that. There are sufficient differences between the classes. I could have this interface

```
FastEMROIOverlay(cnvs, coordinates, colour=gui.SELECTION_COLOUR, allow_edit=True, dash=True, label=None, activation_margin=0)
"""
allow_edit: don't allow changing region boundaries, don't change cursor when hovering on boundaries
dash: True if overlay box should be dashed
label: label of overlay box
activation_margin: margin outside box that triggers activation (for roc to allow selection from far zoom level)
"""
```

For me, it feels clearer to have two classes than all these parameters. Maybe in the future the classes will diverge more and we'll have to add more arguments.

Inheritance is an option, but doesn't change much. The common functions are init and the ._coordinates VA callback. Possibly also `_on_left_up`, the ROA has some special code here for aborting the process, but it would also not hurt the ROC (although in this case it might be confusing).

What's your preferred choice?

src/odemis/gui/cont/acquisition.py
philipwinklerdelmic: That is a very good question. For now, Sabrina and I agreed to worry about this in the second iteration. There's two potential ways we can get a conflict:
* using the same name during the same session --> that's easy to check, we just need to go through all the projects/rois in the model
* using the same name after restarting odemis on the same day (will use same external folder) --> more complicated, we need to check external storage, only possible if that is mounted

Then, there is also the question when to check. We could do it just before the acquisition and just add a counter if necessary. Another option would be to check immediately after typing in the GUI and give the user some feedback that the name has changed (if we do it automatically)/ that it has to be changed (if we require the user to do it).

src/odemis/gui/comp/overlay/world.py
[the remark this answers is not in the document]
pieleric: Indeed, I think it's enough differences that it'd feel a little odd in a single class. I'd suggest to go with inheritance.

install/linux/usr/share/odemis/sim/fastem-sim.yaml
philipwinklerdelmic: I was told that it should be fast-em, but maybe that is outdated information. I think we should check with Sabrina when she comes back.

src/odemis/gui/cont/acquisition.py
philipwinklerdelmic: I'm implementing a simple check now every time the user changes the ROI/project name and adding a counter if necessary.

src/odemis/gui/comp/overlay/world.py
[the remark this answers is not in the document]
philipwinklerdelmic: Ok. I created a superclass `FastEMSelectOverlay`, but to be honest, I don't think it's making the code any easier. Maybe you have some ideas on how to improve the inheritance.

src/odemis/util/__init__.py
```diff
@@ -350,6 +350,33 @@ def normalize_rect(rect):
...
+def expand_rect(rect, margin):
+    """
+    Expand a rectangle by a fixed margin.
+
+    rect (tuple of 4 floats): l, t, r, b positions of rectangle in m
+    margin (float): margin to increase rectangle by in m
+    return (tuple of 4 floats): l, t, r, b positions of adjusted rectangle in m
+    """
+    l, t, r, b = rect
+    return l - margin, t + margin, r + margin, b - margin
```
pieleric, line 377: I realise that this is different from what we normally do: l & t are the small ones and r & b the big ones. IOW, t < b. The names top/bottom are correct in pixel coordinates, but not physical coordinates (as Y is inverted, per convention). See the normalize_rect() function.
So, please follow this convention. You can also drop the "in m" in your docstring, as the rectangle computation is unit independent.
Please also change the `is_point_in_rect()` function.

src/odemis/util/filename.py
```diff
@@ -198,6 +198,27 @@ def generate_fn(ptn, count):
...
+        return fn
+
+    # TODO: Detect if filename has counter already
+
+    # Add counter and count up until unique filename is found
+    cnt = 1
+    while fn + "-%s" % cnt in names:
+        cnt += 1
+
+    return fn + "-%s" % cnt
```
pieleric, line 230: Could you add 1 or 2 test cases for this function? It doesn't have to be fancy, but running the code goes a long way already. Maybe one call with an empty list, one with a conflicting filename, and one without conflicts.

src/odemis/util/filename.py
```diff
@@ -198,6 +198,27 @@ def generate_fn(ptn, count):
 
     return os.path.join(path, fn + ext)
 
+def individualize_filename(fn, names):
```
pieleric: I'm not super convinced by "individualize", it was quite unclear what it'd do, when reading the caller text. Maybe `make_unique_filename()`. 

Also, I'd recommend "names" -> "existing".

src/odemis/util/filename.py
```diff
@@ -198,6 +198,27 @@ def generate_fn(ptn, count):
...
+    """
+    Creates a filename based on the input filename which is unique in a list of filenames by adding a counter.
+    input
+        fn (str): proposed filename
```
pieleric, line 205: Is that really a filename, or just a name? If it's a filename, maybe the function should take care of inserting the counter before the extension?

pieleric (changes requested):
Indeed the inheritance is a little underwhelming. It's kind of odd, because the code looks very similar but as it's different in the middle of the lowest if block, it's not possible to share anything (easily). I'd say that it's still a tiny better than before, so let's leave it like this. In particular, it highlights which methods are different.

install/linux/usr/share/odemis/sim/fastem-sim.yaml
```diff
@@ -0,0 +1,64 @@
...
+                axis_number: 2,
+                speed: 0.001, # m/s
+                closed_loop: True,
+            },
+        },
+    },
+    metadata: {
+        POS_ACTIVE_RANGE: {"x": [-60e-3, 60e-3], "y": [-60e-3, 60e-3]},
+    },
+}
```
pieleric, line 64: I don't know how much it's a good "mbsem" simulator. I think it's missing a lot of components. Anyway, it's good to start somewhere. I think that you typically still want to directly add a `affects: ["EBeam Scanner"]` on the stage, so that moving the stage affects the position of the sem image.

install/linux/usr/share/odemis/sim/fastem-sim.yaml
```diff
@@ -0,0 +1,64 @@
...
+    class: Microscope,
+    role: mbsem,
+    children: ["SEM", "EBeam Scanner", "EBeam Focus", "SED",
+               "Sample Stage"],
+}
+
+"SEM": {
+    class: simsem.SimSEM,
+    # TODO remove role when children components are implemented (chamber camera, chamber and sed)
+    role: sem,
```
pieleric, line 12: You can remove the role (ie, set it to "null"). And also remove the "SEM" from the Microscope children.

install/linux/usr/share/odemis/sim/fastem-sim.yaml
```diff
@@ -0,0 +1,64 @@
...
+    init: {
+           image: "simsem-fake-output.h5", # any large 16 bit image is fine
+           drift_period: 5, # sec
```
pieleric, line 15: If you remove this argument, the image will be larger, and not drift, which is probably closer to the reality.

pieleric (changes requested):

src/odemis/util/filename.py
```diff
@@ -211,10 +211,19 @@ def individualize_filename(fn, names):
...
-    # TODO: Detect if filename has counter already
+    # Detect if filename has counter already, otherwise use 1
+    cnt_ptn = r'-\d{1,5}'
+    cnt_m = None
+    for m in re.finditer(cnt_ptn, fn):
+        cnt_m = m
+    # if multiple numbers are present, use last
+    if cnt_m:
+        cnt = int(cnt_m.group()[1:])
+        fn = fn[:cnt_m.start()]
```
pieleric, line 222: I think you want the same as in the original code `+ "{cnt}" + fn_ptn[cnt_m.end():]`. Otherwise, as soon as the filename has -number, everything after it is removed. 
Maybe in the else block you can do `fn = fn + "{cnt}"`

pieleric:
I've added extra test cases to make_unique_name. They all pass.
That said, for now the behaviour is `make_unique_name('abc-2', ['abc', 'abc-2'])` -> `'abc-3'` . The previous code you had written was giving `'abc-1'`. I don't really mind. But you might disagree and want to adjust the code.

PR #1762, opened 2021-05-06
Implement FIB scanner for both dual as single mode operation with test cases

Kleijwegt:
Still needs to be tested and needs another self review before going into proper review.


src/odemis/driver/test/xt_client_test.py
```diff
@@ -955,6 +974,116 @@ def test_beamlet_index_info(self):
...
+        if TEST_NOHW:
+            self.skipTest("No hardware available.")
+
+        if self.microscope.get_vacuum_state() != 'vacuum':
+            self.skipTest("Chamber needs to be in vacuum, please pump.")
+
+        if "xttoolkit" in self.microscope.swVersion.lower():
+            self.xt_type = "xttoolkit"
+        else:
+            raise TypeError("Xttoolkit must be running for this test.")
```
pieleric, line 1038: No, it's OK to only have xtlib

src/odemis/driver/xt_client.py
```diff
@@ -1469,7 +1498,19 @@ def __init__(self, name, role, parent, channel_name, **kwargs):
...
+                                                        setter=self._set_active_scanner_type)
+        elif hasattr(self.parent, "_scanner"):
+            self._channel_name = channel_name
+            self._active_scanner = self.parent._scanner
+        elif hasattr(self.parent, "_fib_scanner"):
+            self._channel_name = channel_name
+            self._active_scanner = self.parent._fib_scanner
+        else:
+            raise ValueError("Non of the required scanner children are provided. Provide at least an ebeam scanner ("
+                             "single of multi beam) or a FIB scanner.")
```
pieleric, line 1512: The parent already checks for that normally. So you can keep the message  simple, like "No scanner available".

src/odemis/driver/xt_client.py
```diff
@@ -1469,7 +1498,19 @@ def __init__(self, name, role, parent, channel_name, **kwargs):
         self._shape = (256,)  # Depth of the image
         self.data = SEMDataFlow(self)
 
-        self._channel_name = channel_name
+        if hasattr(self.parent, "_scanner") and hasattr(self.parent, "_fib_scanner"):
+            self.active_scanner_type = StringEnumerated("electron", choices={"electron", "ion"},
+                                                        setter=self._set_active_scanner_type)
```
pieleric, line 1503: Let's call this VA just "scanner".
Also, the content should be the **name** of the components. That's how we link the components in Odemis in general, so let's keep the same type of interface.

src/odemis/driver/xt_client.py
```diff
@@ -1447,6 +1458,24 @@ def _cancelAutoContrastBrightness(self, future):
...
+
+    def __init__(self, name, role, parent, **kwargs):
+        model.Emitter.__init__(self, name, role, parent=parent, **kwargs)
+
+        # Set unknown values to None or 0
+        self.dwellTime = model.VigilantAttribute(None, readonly=True)
+        self._shape = (0,)
+        self.updateMetadata({model.MD_DWELL_TIME: None,
+                             model.MD_ROTATION  : None,
+                             model.MD_PIXEL_SIZE: None})
```
pieleric, line 1476: If you don't know, just don't put the metadata.

src/odemis/driver/xt_client.py
```diff
@@ -1447,6 +1458,24 @@ def _cancelAutoContrastBrightness(self, future):
...
+    of the FIB is possible, which is limited to pause/unpausing the image beam, and getting the latest image (via the
+    detector with the appropriate channel).
+    """
+
+    def __init__(self, name, role, parent, **kwargs):
+        model.Emitter.__init__(self, name, role, parent=parent, **kwargs)
+
+        # Set unknown values to None or 0
+        self.dwellTime = model.VigilantAttribute(None, readonly=True)
+        self._shape = (0,)
```
pieleric, line 1473: We don't know the shape, but something like 4096,4096 would be more reasonable.
We might want to have a read-only .resolution VA with the same values. There is a lot of code that expects to get this info.

src/odemis/driver/xt_client.py
```diff
@@ -1447,6 +1458,24 @@ def _cancelAutoContrastBrightness(self, future):
...
+    This is an extension of the model.Emitter class for controlling the FIB. Currently via XTLib only minimal control
+    of the FIB is possible, which is limited to pause/unpausing the image beam, and getting the latest image (via the
+    detector with the appropriate channel).
+    """
+
+    def __init__(self, name, role, parent, **kwargs):
+        model.Emitter.__init__(self, name, role, parent=parent, **kwargs)
+
+        # Set unknown values to None or 0
+        self.dwellTime = model.VigilantAttribute(None, readonly=True)
```
pieleric, line 1472: Drop, as we don't know it, and cannot change it either.

src/odemis/driver/xt_client.py
```diff
@@ -1447,6 +1458,24 @@ def _cancelAutoContrastBrightness(self, future):
...
 
+class FibScanner(model.Emitter):
+    """
+    This is an extension of the model.Emitter class for controlling the FIB. Currently via XTLib only minimal control
+    of the FIB is possible, which is limited to pause/unpausing the image beam, and getting the latest image (via the
+    detector with the appropriate channel).
+    """
+
+    def __init__(self, name, role, parent, **kwargs):
+        model.Emitter.__init__(self, name, role, parent=parent, **kwargs)
```
pieleric, line 1469: I'd suggest to have a argument "channel", with the default value "ion2".
This value should be store on the .channel (or .channel_name) attribute .

src/odemis/driver/xt_client.py
```diff
@@ -1469,7 +1498,19 @@ def __init__(self, name, role, parent, channel_name, **kwargs):
         self._shape = (256,)  # Depth of the image
         self.data = SEMDataFlow(self)
 
-        self._channel_name = channel_name
+        if hasattr(self.parent, "_scanner") and hasattr(self.parent, "_fib_scanner"):
+            self.active_scanner_type = StringEnumerated("electron", choices={"electron", "ion"},
+                                                        setter=self._set_active_scanner_type)
+        elif hasattr(self.parent, "_scanner"):
+            self._channel_name = channel_name
```
pieleric, line 1505: Drop the channel_name argument, and always use the scanner.channel attribute.

src/odemis/driver/xt_client.py
```diff
@@ -1507,14 +1549,23 @@ def _acquire(self):
...
+                    md = self._active_scanner._metadata.copy()
 
                     # Estimated time for an acquisition is the dwell time times the total amount of pixels in the image.
-                    n_pixels = self.parent._scanner.shape[0] * self.parent._scanner.shape[1]
-                    est_acq_time = self.parent._scanner.dwellTime.value * n_pixels
+                    if self._active_scanner.dwellTime.value and self._active_scanner.shape:
+                        n_pixels = self._active_scanner.shape[0] * self._active_scanner.shape[1]
+                        est_acq_time = self._active_scanner.dwellTime.value * n_pixels
+                        md[model.MD_DWELL_TIME] = self._active_scanner.dwellTime.value
+                        md[model.MD_ROTATION] = self._active_scanner.rotation.value
```
pieleric, line 1564: Please move it to a separate `if hasattr(...)`.

src/odemis/driver/xt_client.py
```diff
@@ -1507,14 +1549,23 @@ def _acquire(self):
...
                     self.parent.set_channel_state(self._channel_name, False)
+                    md = self._active_scanner._metadata.copy()
 
                     # Estimated time for an acquisition is the dwell time times the total amount of pixels in the image.
-                    n_pixels = self.parent._scanner.shape[0] * self.parent._scanner.shape[1]
-                    est_acq_time = self.parent._scanner.dwellTime.value * n_pixels
+                    if self._active_scanner.dwellTime.value and self._active_scanner.shape:
+                        n_pixels = self._active_scanner.shape[0] * self._active_scanner.shape[1]
+                        est_acq_time = self._active_scanner.dwellTime.value * n_pixels
+                        md[model.MD_DWELL_TIME] = self._active_scanner.dwellTime.value
```
pieleric, line 1563: Let's just protect with a `if hasattr(...,"dwellTime")`.

src/odemis/driver/xt_client.py
```diff
@@ -1507,14 +1549,23 @@ def _acquire(self):
...
+                            self.parent.unblank_beam()
                     # The channel needs to be stopped to acquire an image, therefore immediately stop the channel.
                     self.parent.set_channel_state(self._channel_name, False)
+                    md = self._active_scanner._metadata.copy()
 
                     # Estimated time for an acquisition is the dwell time times the total amount of pixels in the image.
-                    n_pixels = self.parent._scanner.shape[0] * self.parent._scanner.shape[1]
-                    est_acq_time = self.parent._scanner.dwellTime.value * n_pixels
+                    if self._active_scanner.dwellTime.value and self._active_scanner.shape:
+                        n_pixels = self._active_scanner.shape[0] * self._active_scanner.shape[1]
```
pieleric, line 1561: BTW, I think this is not really correct. It should use .resolution.value

pieleric (changes requested):

src/odemis/driver/xt_client.py
Kleijwegt: with name of the component you mean Scanner/FibScanner (as in name of the class) or do you mean the role of the component?

pieleric: I mean _scanner.name or _fib_scanner.name . These are the names picked by the "user" (ie, set in the microscope file).

Kleijwegt:
I have updated the code. Probably it is best that after testing (on Monday) @tepals checks if none of the proposed changes have an effect on the XTTKDetector/mb-scanner and optical channels.

Kleijwegt:
I have updated and tested all the code but I need to do a review to check for ugly parts which slipped into it. After this it is ready for review

src/odemis/driver/test/xt_client_test.py
```diff
@@ -31,7 +31,7 @@
 from odemis import model
 
 from odemis.driver import xt_client
-from odemis.driver.xt_client import DETECTOR2CHANNELNAME, XTTKDetector
+from odemis.driver.xt_client import XTTKDetector
```
Kleijwegt, line 34: It seems as if XTTKDetector remains unused, however I didn't put it there/neither delete anything related to it. So I'll leave it.

tepals: You can remove it, it was left over from a test that got removed. I first created a separate test for the XTTKDetector, however in the constructor of the SEM class depending on if XTToolkit is used the correct Detector is picked. Therefore the tests regular acquisition tests in TestMicroscope also apply to the XTTKDetector.

Kleijwegt:
It's ready for review

src/odemis/driver/xt_client.py
```diff
@@ -1506,15 +1536,26 @@ def _acquire(self):
                 while True:
                     if self._acq_should_stop():
                         break
-                    self.parent.set_channel_state(self._channel_name, True)
-                    if self.parent._scanner.blanker.value is None:
-                        self.parent.unblank_beam()
+                    self.parent.set_channel_state(self._scanner.channel, True)
+                    if hasattr(self._scanner, "blanker"):
+                        if self._scanner.blanker.value is None:
+                            self.parent.unblank_beam()
```
pieleric, line 1542: Note that if the _scanner has changed since last image, we should blank the *previous* scanner. 
So you should keep a reference to the scanner, so blank the beam if it is changed. 
That said, it seems nothing of this is currently handled. So maybe  you could just add a TODO, and let it be done when someone fixes the blanker handling.

src/odemis/driver/xt_client.py
```diff
@@ -1920,15 +1969,13 @@ def __init__(self, name, role, parent, **kwargs):
...
-    def applyAutofocus(self, detector):
+    def applyAutofocus(self):
         """
         Wrapper for running the autofocus functionality asynchronously. It sets the state of autofocus,
         the beam must be turned on and unblanked. Also a a reasonable manual focus is needed. When the image is too far
         out of focus, an incorrect focus can be found using the autofocus functionality.
         This call is non-blocking.
 
-        :param detector (str): Role of the detector.
-        :param state (str):  "run", or "stop"
```
pieleric, line 1931: Leave the detector param (although it's not yet used).

pieleric (changes requested):

src/odemis/driver/test/xt_client_test.py
```diff
@@ -955,6 +954,106 @@ def test_beamlet_index_info(self):
...
+        if TEST_NOHW:
+            self.skipTest("No hardware available.")
+        if self.microscope.get_vacuum_state() != 'vacuum':
+            self.skipTest("Chamber needs to be in vacuum, please pump.")
+        self.xt_type = "xttoolkit" if "xttoolkit" in self.microscope.swVersion.lower() else "xtlib"
+
+    def test_acquire(self):
+        """Test acquiring an image from the FIB/ion2 channel."""
+        image = self.detector.data.get()
+        self.assertEqual(len(image.shape), 2)
```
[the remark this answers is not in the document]
Kleijwegt: I think we didn't define a pixel size because it is unknown for the FIB, wasn't that the plan?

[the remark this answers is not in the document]
Kleijwegt: I didn't know that, thanks!

src/odemis/driver/xt_client.py
```diff
@@ -90,23 +85,34 @@ def __init__(self, name, role, children, address, daemon=None,
...
+            self._fib_scanner = FibScanner(parent=self, daemon=daemon, **kwargs)
+            self.children.value.add(self._fib_scanner)
+
         if "mb-scanner" in children:
+            if "scanner" in children:
+                raise NotImplementedError("The combination of both an multi beam scanner and single beam scanner at "
+                                          "the same time is not supported")
+            kwargs = children["mb-scanner"]
+            if "xttoolkit" in self._swVersion.lower():
+                self.xt_type = "xttoolkit"
```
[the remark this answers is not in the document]
Kleijwegt: @tepals this seems like a part from previous code which is now included in my PR because I needed to reorder the init. Do you think it is ok to delete the .xt_type safely?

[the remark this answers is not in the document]
tepals: I think you can remove it, just don't forget to re-run all the test cases if you do.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -955,6 +954,106 @@ def test_beamlet_index_info(self):
...
+        self.scanner.dwellTime.value = 25e-9  # s
+        expected_duration = self._compute_expected_duration()
+        start = time.time()
+        im = self.detector.data.get()
+        duration = time.time() - start
+        self.assertEqual(im.shape, self.scanner.resolution.value[::-1])
+        self.assertGreaterEqual(duration, expected_duration,
+                                "Error execution took %f s, less than exposure time %d." % (
+                                    duration, expected_duration))
+        self.assertIn(model.MD_DWELL_TIME, im.metadata)
```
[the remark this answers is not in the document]
Kleijwegt: copied from "test_acquire"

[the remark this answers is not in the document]
Kleijwegt: copied from "test_acquire"

Kleijwegt (commented):
Before merging I would like to test everything on Helios. I did test it on FUMO-Beta and the simulator. I don't expect any changes from testing it as Helios

src/odemis/driver/test/xt_client_test.py
```diff
@@ -956,6 +956,143 @@ def test_beamlet_index_info(self):
...
+                self.assertGreaterEqual(self.images_received, 1)
+
+    def test_acquire_FIB_image(self):
+        """Test acquiring an image from the FIB/ion channel."""
+        # Switch to ion mode
+        self.detector.scanner.value = self.fib_scanner.name
+        image = self.detector.data.get()
+        self.assertEqual(image.ndim, 2)
+        # Check that image size is at least 200*200 pixels
+        self.assertGreaterEqual(image.shape[0],200)
```
tepals, line 1070: Add space after `,`

tepals (changes requested):

src/odemis/driver/xt_client.py
```diff
@@ -1506,35 +1536,44 @@ def _acquire(self):
...
-                    md = self.parent._scanner._metadata.copy()
+                    # try:
+                    #     if self._acq_wait_data(est_acq_time + 20):
+                    #         logging.debug("Stopping measurement early")
+                    #         self.stop_acquisition()
+                    #         break
+                    # except TimeoutError as err:
+                    #     logging.error(err)
+                    #     self.stop_acquisition()
+                    #     break
```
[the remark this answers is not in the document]
Kleijwegt: That was for testing on the simulator, that should have been uncommetted.

src/odemis/driver/xt_client.py
```diff
@@ -1638,6 +1677,19 @@ def _get_acq_msg(self, **kwargs):
             logging.warning("Acq received unexpected message %s", msg)
         return msg
 
+    def _set_scanner(self, scan_mode):
```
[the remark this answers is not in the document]
Kleijwegt: @pieleric thought the VA should be called .scanner. so therefore I named the setter "_set_scanner". It used to be actually "set_active_scanner"

src/odemis/driver/xt_client.py
```diff
@@ -1447,17 +1452,31 @@ def _cancelAutoContrastBrightness(self, future):
...
+
+    def __init__(self, name, role, parent, channel="ion2", **kwargs):
+        model.Emitter.__init__(self, name, role, parent=parent, **kwargs)
+
+        self.channel = channel  # Name of the ion channel used.
+
+        # Fictional values for the interface. Xtlib doesn't support reading/controlling these values
+        self._shape = (4096, 4096)
+        res = self._shape[:2]
+        self.resolution = model.ResolutionVA(res, (res, res), readonly=True)
```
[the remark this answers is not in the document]
Kleijwegt: I coppied the range part from picoquant line 567. That seems to work without any problems since 2016

src/odemis/driver/test/xt_client_test.py
```diff
@@ -724,108 +742,90 @@ def test_apply_autostigmator(self):
...
+        # Start auto contrast brightness and check if it is running.
+        auto_contrast_brightness_future = self.scanner.applyAutoContrastBrightness(self.detector.role)
+        auto_contrast_brightness_state = self.microscope.is_running_auto_contrast_brightness(self.scanner.channel)
+        time.sleep(0.01)
+        self.assertEqual(auto_contrast_brightness_state, True)
+        self.assertIsInstance(auto_contrast_brightness_future, ProgressiveFuture)
+
+        # Stop auto auto contrast brightness and check if it stopped running.
+        auto_contrast_brightness_future.cancel()
+        time.sleep(5.0)  # Give microscope/simulator the time to update the state
```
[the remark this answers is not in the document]
Kleijwegt: These changes are just changes in the indent because of the removal of the for loop and the DETECTOR2CHANNELNAME.
I suppose that in the previous merge (last year) I did check that 5 seconds is needed.

src/odemis/driver/xt_client.py
```diff
@@ -1506,35 +1536,44 @@ def _acquire(self):
...
+                    # except TimeoutError as err:
+                    #     logging.error(err)
+                    #     self.stop_acquisition()
+                    #     break
+
+                    # Retrieve the image
+                    image = self.parent.get_latest_image(self._scanner.channel)
                     md.update(self._metadata)
-                    md[model.MD_DWELL_TIME] = self.parent._scanner.dwellTime.value
-                    md[model.MD_ROTATION] = self.parent._scanner.rotation.value
```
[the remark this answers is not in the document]
Kleijwegt: I tried getting the metadata after the image is pauzed but isn't yet taken (as early as possible but not to soon so that values might have changed. Plus there wasn't already a if statement for the dwelltime/rotation below.
Maybe it is to spread out though and it should be brought back down? I don't know.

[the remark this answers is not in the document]
tepals: Ok, then just leave it as you have it now.

src/odemis/driver/xt_client.py
```diff
@@ -1506,35 +1536,44 @@ def _acquire(self):
                 while True:
                     if self._acq_should_stop():
                         break
-                    self.parent.set_channel_state(self._channel_name, True)
-                    if self.parent._scanner.blanker.value is None:
-                        self.parent.unblank_beam()
+                    self.parent.set_channel_state(self._scanner.channel, True)
+                    if hasattr(self._scanner, "blanker"):
+                        # TODO In case of switching the scanner type the previous scanner should be blanked.
+                        #  Currently this kind of blanker handling isn't supported by XT.
```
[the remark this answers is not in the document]
pieleric: I think this is indirectly coming from me. It is supported by XT. What is not supported is the auto blanking at the end of an acquisition. If you look at the code, if blanker = None, it disables the blanker before acquiring, but never sets it back after acquiring, which it should. That's what this comment should be about.

src/odemis/driver/xt_client.py
```diff
@@ -1506,35 +1536,44 @@ def _acquire(self):
...
-                    n_pixels = self.parent._scanner.shape[0] * self.parent._scanner.shape[1]
-                    est_acq_time = self.parent._scanner.dwellTime.value * n_pixels
+                    if hasattr(self._scanner, "dwellTime") and hasattr(self._scanner, "shape"):
+                        n_pixels = self._scanner.resolution.value[0] * self._scanner.resolution.value[1]
+                        est_acq_time = self._scanner.dwellTime.value * n_pixels
+                        md[model.MD_DWELL_TIME] = self._scanner.dwellTime.value
+                    else:
+                        # Estimated acquisition time used the timeout if not all info about this acquisition time is
+                        # known.
+                        est_acq_time = 60
```
pieleric, line 1557: There are cases where an acquisition could take several minutes. It'd be a pity to fail consistently on such high level acquisition. So let's raise this timeout to 5*60 s

src/odemis/driver/xt_client.py
```diff
@@ -1506,35 +1536,44 @@ def _acquire(self):
...
                     # Estimated time for an acquisition is the dwell time times the total amount of pixels in the image.
-                    n_pixels = self.parent._scanner.shape[0] * self.parent._scanner.shape[1]
-                    est_acq_time = self.parent._scanner.dwellTime.value * n_pixels
+                    if hasattr(self._scanner, "dwellTime") and hasattr(self._scanner, "shape"):
+                        n_pixels = self._scanner.resolution.value[0] * self._scanner.resolution.value[1]
+                        est_acq_time = self._scanner.dwellTime.value * n_pixels
+                        md[model.MD_DWELL_TIME] = self._scanner.dwellTime.value
+                    else:
+                        # Estimated acquisition time used the timeout if not all info about this acquisition time is
+                        # known.
```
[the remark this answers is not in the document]
pieleric: I think it could just be "No idea about acquisition time => assume it will be long"

src/odemis/driver/xt_client.py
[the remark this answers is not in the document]
pieleric: It's the setter of the .scanner, so I find it clearer to have it named "_set_scanner".

src/odemis/driver/xt_client.py
```diff
@@ -1638,6 +1677,19 @@ def _get_acq_msg(self, **kwargs):
             logging.warning("Acq received unexpected message %s", msg)
         return msg
 
+    def _set_scanner(self, scan_mode):
+        """
+        Setter for changing the current active scanner type. The correct scanner object is also updated in ._scanner.
+        :param scan_mode (string): contains mode, can be either 'Scanner' or 'FibScanner'
```
pieleric, line 1683: Please update the docstring. I'd recommend renaming "scan_mode" -> "name" or "scanner_name".

src/odemis/driver/xt_client.py
```diff
@@ -1938,7 +1989,7 @@ def applyAutofocus(self, detector):
         f._autofocus_lock = threading.Lock()
         f._must_stop = threading.Event()  # cancel of the current future requested
         f.task_canceller = self._cancelAutoFocus
-        f._channel_name = DETECTOR2CHANNELNAME[detector]
+        f._channel_name = self._scanner.channel
```
pieleric, line 1992: I don't see a `._scanner` attribute. That should be something like `self.parent._scanner.channel`, right?
I think the focus is hard coded to be for the ebeam in the xtlib API. So I'd also recommend to check at init that there is a _scanner (ie, e-beam "scanner" or "mb-scanner").

pieleric (changes requested):

src/odemis/driver/test/xt_client_test.py
```diff
@@ -956,6 +956,143 @@ def test_beamlet_index_info(self):
...
+        # Check that image size is at least 200*200 pixels
+        self.assertGreaterEqual(image.shape[0], 200)
+        self.assertGreaterEqual(image.shape[1], 200)
+        self.images_received += 1
+
+    def test_scanner_VA(self):
+        scanner_modes = {self.scanner.name: self.scanner, self.fib_scanner.name: self.fib_scanner}
+        self.assertEqual(set(self.detector.scanner.choices), set(scanner_modes))
+
+        for _ in range(2):
```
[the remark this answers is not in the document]
tepals: yes

Kleijwegt:
After approval I will squash all the commits, rebase and fix the conflicts

src/odemis/driver/test/xt_client_test.py
```diff
@@ -956,6 +955,151 @@ def test_beamlet_index_info(self):
...
+
+        # Check if mode remains unchanged if any non existing mode is tried to be set
+        with self.assertRaises(IndexError):
+            self.detector.scanner.value = "non existing mode"
+        # The mode should be equal to the last set correct mode.
+        self.assertEqual(self.detector.scanner.value, scanner_mode)
+
+    def test_subscribing_while_changing_scanner(self):
+        """
+        Subscribes to the dataflow then switches between both scanner types and then checks if an images is received
```
tepals, line 1047: typo: images -> image

src/odemis/driver/test/xt_client_test.py
```diff
@@ -956,6 +955,151 @@ def test_beamlet_index_info(self):
...
+                # Switch active scanner mode
+                self.detector.scanner.value = scanner_mode
+                # Set image counter to zero to check that new images are received after switching
+                self.images_received = 0
+                self.assertEqual(self.detector.scanner.value, scanner_mode)
+                self.assertIsInstance(self.detector._scanner, type(scanner_modes[scanner_mode]))
+                time.sleep(2)
+                self.assertGreaterEqual(self.images_received, 1)
+
+        self.detector.data.unsubscribe(self._callback_counter)  # Unsubscribe so to stop refreshing the image
```
tepals, line 1068: you can remove "so" here.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -956,6 +955,151 @@ def test_beamlet_index_info(self):
...
+        """Test acquiring an image using the Detector."""
+        # Switch to electron mode
+        self.detector.scanner.value = self.scanner.name
+        init_dwell_time = self.scanner.dwellTime.value
+        self.scanner.dwellTime.value = 25e-9  # s
+        expected_duration = self._compute_expected_duration()
+        start = time.time()
+        im = self.detector.data.get()
+        duration = time.time() - start
+        # self.assertEqual(im.shape, self.scanner.resolution.value[::-1])
```
tepals, line 1090: why is this commented?

Kleijwegt: No idea, I will test this again on Helios but I shouldn't have been commented.

tepals (approved):

src/odemis/driver/xt_client.py
```diff
@@ -1915,6 +1980,11 @@ def __init__(self, name, role, parent, **kwargs):
         self.position = model.VigilantAttribute({}, unit="m", readonly=True)
         self._updatePosition()
 
+        if not hasattr(self.parent, "_scanner"):
+            if not hasattr(self.parent, "_mb-scanner"):
```
pieleric, line 1984: There is no such `"_mb-scanner"` (or `"_mb_scanner"`). It's always called `_scanner`. So, you can drop this if.

src/odemis/driver/xt_client.py
```diff
@@ -2268,19 +2337,20 @@ def __init__(self, name, role, parent, channel_name, address, **kwargs):
             raise HwError("Failed to connect to XT server '%s'. Check that the "
                           "uri is correct and XT server is"
                           " connected to the network. %s" % (address, err))
-        Detector.__init__(self, name, role, parent=parent, channel_name=channel_name, **kwargs)
+        Detector.__init__(self, name, role, parent=parent, **kwargs)
         self._shape = (2**16,)  # Depth of the image
+        self._scanner = self.parent._scanner
```
pieleric, line 2342: The parent class already takes care of setting this variable. So you shouldn't change it. => drop thsi line.

src/odemis/driver/xt_client.py
```diff
@@ -1447,17 +1455,41 @@ def _cancelAutoContrastBrightness(self, future):
...
+    def __init__(self, name, role, parent, channel="ion2", **kwargs):
+        """
+
+        :param name (str):
+        :param role (str):
+        :param parent (SEM object):
+        :param channel (str): Specifies the type of scanner (alphabetic part) and the quadrant it is displayed in (
+        nummerical part) on which the scanning feed is displayed on the Microscope PC in the Microscope control window
+        of TFS. The quadrants are numbered 1 (top left) - 4 (right bottom).
+        ion2 is set as default since the top left quadrant is used as default for the ebeam (electron1)
```
pieleric, line 1473: "top left quadrant is used as default for the ebeam (electron1)" ->  "top right quadrant is used as default for the FIB image" (or adjust to whatever is the TFS software behaviour)

src/odemis/driver/xt_client.py
```diff
@@ -1506,15 +1550,28 @@ def _acquire(self):
...
                     if self._acq_should_stop():
                         break
-                    self.parent.set_channel_state(self._channel_name, True)
-                    if self.parent._scanner.blanker.value is None:
-                        self.parent.unblank_beam()
+                    self.parent.set_channel_state(self._scanner.channel, True)
+                    if hasattr(self._scanner, "blanker"):
+                        # TODO: When the acquisition ends, if blanker is set to automatic (None), we need to activate
+                        #  the blanker again. That should also happen when switching e-beam <--> FIB, but currently
+                        #  the FIB blanker is not support via the XT interface.
```
pieleric, line 1557: support -> supported.

src/odemis/driver/xt_client.py
```diff
@@ -1506,15 +1550,28 @@ def _acquire(self):
...
+                            self.parent.unblank_beam()
                     # The channel needs to be stopped to acquire an image, therefore immediately stop the channel.
-                    self.parent.set_channel_state(self._channel_name, False)
+                    self.parent.set_channel_state(self._scanner.channel, False)
+                    md = self._scanner._metadata.copy()
 
                     # Estimated time for an acquisition is the dwell time times the total amount of pixels in the image.
-                    n_pixels = self.parent._scanner.shape[0] * self.parent._scanner.shape[1]
-                    est_acq_time = self.parent._scanner.dwellTime.value * n_pixels
+                    if hasattr(self._scanner, "dwellTime") and hasattr(self._scanner, "shape"):
```
pieleric, line 1565: `hasattr(self._scanner, "shape")` is confusing. You're not using `shape`, and anyway, all scanners have a .shape. 
-> `hasattr(self._scanner, "resolution")` ?
Maybe even clearer to move the metadata update in a separate if, just like rotation. Otherwise, in the (theoretical) case that there is dwellTime, but no resolution, the MD_DWELL_TIME will not be set, but it should.

pieleric (changes requested):

PR #1763, opened 2021-05-07
ENZEL: Handle clearing stream panel

lazem:
Fix the dead listener bug on removing overview streams.
Also fix the issue of removing the stream panel when replacing old overview streams.


Mahmood-B (approved):
Okay good. Then once this PR is merged, i will rebase again and test

PR #1764, opened 2021-05-07
ENZEL: Fix 5DOF Invalid/Unreachable error

lazem:
After referencing the sample stage, it tends to move to a position near the maximum axes range, and trying to move it back to the loading position using our safe method _(rx&rz followed by x&y followed by z)_ may result in an invalid/reachable position error -> For this special case move all linear axes together.

### DIFF
--- a/src/odemis/acq/move.py
+++ b/src/odemis/acq/move.py
@@ -204,6 +204,7 @@ def _doCryoSwitchSamplePosition(future, stage, focus, align, target):
         focus_deactive = focus_md[model.MD_FAV_POS_DEACTIVE]
         align_deactive = align_md[model.MD_FAV_POS_DEACTIVE]
         align_active = align_md[model.MD_FAV_POS_ACTIVE]
+        stage_referenced = all(stage.referenced.value.values())
         # Fail early when required axes are not found on the positions metadata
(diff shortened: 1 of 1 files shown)


src/odemis/acq/move.py
```diff
@@ -243,8 +244,14 @@ def _doCryoSwitchSamplePosition(future, stage, focus, align, target):
 
             # Add the sub moves to perform the loading move
             sub_moves.append((stage, filter_dict({'rx', 'rz'}, stage_deactive)))
-            sub_moves.append((stage, filter_dict({'x', 'y'}, stage_deactive)))
-            sub_moves.append((stage, filter_dict({'z'}, stage_deactive)))
+            if current_label is UNKNOWN and not stage_referenced:
+                # After referencing the stage could move near the maximum axes range,
+                # and moving single axes may result in an invalid/reachable position error,
+                # so all axes will moved together for this special case.
+                sub_moves.append((stage, filter_dict({'x', 'y', 'z'}, stage_deactive)))
```
pieleric, line 251: Should this also be done in lines 271-272, if the stage has just come from LOADING (so it has run_reference())?

lazem: If the stage is in LOADING position, then it's already referenced, so moving from there step by step should be fine.

PR #1765, opened 2021-05-12
Move to target (imaging/coating) position when current is loading

lazem:


lazem:
My bad on this! @pieleric

PR #1770, opened 2021-05-20
acq stitching tiledacq: add registrar and weaver arguments

pieleric:
These argument allow to adjust the exact behaviour of the tiling.
In particular, in some case we want to tile without overlap (because the
stage precisition is good enough). In such case, the default registrar
would get confused/sad. It'd fallback to IDENTITY, but generate a lot of
warnings on the way.

### DIFF
--- a/src/odemis/acq/stitching/test/tiledacq_test.py
+++ b/src/odemis/acq/stitching/test/tiledacq_test.py
@@ -20,22 +20,23 @@ Odemis. If not, see http://www.gnu.org/licenses/.
 
 from __future__ import division
 
-import logging
-import os
-import time
-import unittest
 from concurrent.futures._base import CancelledError, FINISHED
-
+import logging
 import numpy
-
 import odemis
 import odemis.acq.stream as stream
 from odemis import model
 from odemis.acq.acqmng import SettingsObserver
 from odemis.acq.stitching._tiledacq import TiledAcquisitionTask, acquireTiledArea
 from odemis.util import test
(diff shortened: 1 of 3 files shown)


lazem (approved):

pieleric:
I hadn't properly tested the changes. This new version seems to really work, at least on the simulator. Please @lazem, have a second look.

lazem:
> I hadn't properly tested the changes. This new version seems to really work, at least on the simulator. Please @lazem, have a second look.

Sure @pieleric I will have a deeper look.

src/odemis/acq/stitching/test/tiledacq_test.py
```diff
@@ -244,6 +247,19 @@ def test_whole_procedure(self):
         self.assertIsInstance(data[0], model.DataArray)
         self.assertEqual(len(data[0].shape), 2)
 
+    def test_registrar_weaver(self):
+
+        overlap = 0.05  # Little overlap, no registration
+        area = (0, 0, 0.00001, 0.00001)
```
lazem, line 253: This area is smaller than the fov, so it only stitches 1 image. Suggest to change it to:
```
sem_fov = compute_scanner_fov(self.ebeam)
area = (0, 0, sem_fov[0], sem_fov[1])
```

lazem (changes requested):
Looks good, I suggest one small change to the test.

lazem (approved):

PR #1773, opened 2021-05-21
ENZEL: Cryo Feature Overlay

lazem:
Extend the `StagePointSelectOverlay` to handle showing and editing the interesting features of the cryo project. The `CryoFeatureOverlay` moves the stage to the feature position upon double click. When the feature tool is enabled, drag the selected feature to a new position, or create a new feature on that position.


src/odemis/gui/cont/tabs.py
```diff
@@ -2117,13 +2114,14 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
+            ctrl_rng = max(actual_rng[0], guimod.MILLING_ANGLE_RANGE[0]), min(actual_rng[1], guimod.MILLING_ANGLE_RANGE[1])
+            if not ctrl_rng[0] <= guimod.DEFAULT_MILLING_ANGLE <= ctrl_rng[1]:
+                raise ValueError("Default milling angle %s should be within calculated milling range %s" % (
+                guimod.DEFAULT_MILLING_ANGLE, ctrl_rng))
             panel.ctrl_milling.SetValueRange(*(math.degrees(r) for r in ctrl_rng))
             # Default value for milling angle, will be used to store the angle value out of milling position
-            self._prev_milling_angle = DEFAULT_MILLING_ANGLE
-            self.panel.ctrl_milling.Value = readable_str(math.degrees(DEFAULT_MILLING_ANGLE), unit="°", sig=3)
+            self._prev_milling_angle = guimod.DEFAULT_MILLING_ANGLE
+            self.panel.ctrl_milling.Value = readable_str(math.degrees(guimod.DEFAULT_MILLING_ANGLE), unit="°", sig=3)
```
Mahmood-B, line 2124: Why did you define the variable `self._tab_panel` if you do not use it? You use `self.panel` instead. You could use of them only, to reduce redundancy.

src/odemis/gui/comp/overlay/world.py
```diff
@@ -111,6 +112,170 @@ def on_left_up(self, evt):
...
+            if selected_tool == TOOL_FEATURE:
+                self._mode = MODE_EDIT_FEATURES
+            else:
+                self._mode = MODE_SHOW_FEATURES
+
+    def _on_features_changes(self, _):
+        # refresh the canvas on _features_va list change
+        self.cnvs.update_drawing()
+
+    def on_dbl_click(self, evt):
```
Mahmood-B, line 159: The functions `on_dbl_click()`, `on_left_down()`, `on_left_up()` and `on_motion()` are callbacks right? But i do not  see connections between those functions and VA's or connections using the wx function `Bind()`. So how does this work?

src/odemis/gui/comp/overlay/world.py
```diff
@@ -111,6 +112,170 @@ def on_left_up(self, evt):
...
+        :return: (CryoFeature or None) Found feature, None if not found
+        """
+        def in_radius(c_x, c_y, r, x, y):
+            return math.hypot(c_x - x, c_y - y) <= r
+
+        offset = self.cnvs.get_half_buffer_size()  # to convert physical feature positions to pixels
+        for feature in self._features_va.value:
+            pos = feature.pos.value
+            fvsp = self.cnvs.phys_to_view(pos, offset)
+            if in_radius(fvsp[0], fvsp[1], FEATURE_DIAMETER, v_pos[0], v_pos[1]):
```
Mahmood-B, line 233: I think you could directly use the function `math.hypot` in the if statement here.

Mahmood-B (changes requested):

src/odemis/gui/model/__init__.py
```diff
@@ -596,7 +603,34 @@ def __init__(self, main):
...
+        Create a new feature and add it to the features list
+        """
+        if not f_name:
+            f_name = new_feature_name(len(self.features.value))
+        if not pos_z:
+            pos_z = self.main.focus.position.value['z']
+        if not milling_angle:
+            milling_angle = DEFAULT_MILLING_ANGLE
+        feature = CryoFeature(f_name, pos_x, pos_y, pos_z, milling_angle)
+        self.features.value.append(feature)
```
Mahmood-B, line 630: Let me first ask: is the features bookmarking functionality only accessible in the localization tab? If yes, you could define the features VA list in the localizationGUIData. Similarly for the function "add_new_feature". Because, the chamber class, for example, would have access to this function, although the chamber tab is not relevant to the feature selection and features bookmaking, if i am not mistaken.

lazem: That's right, the chamber tab was meant to have the overview map view with the features shown. That's why I created this `CryoGUIData` class so any of the tabs would have the features list. My plan was to check on tab creation if any of the other cryo tabs is created -> then point to its features list directly. But it seems that we may not need to add this view to the chamber tab after all and have only the chamber cam view.

src/odemis/gui/comp/overlay/world.py
lazem: The `Overlay` abstract base class contains these bindings.

src/odemis/gui/model/__init__.py
```diff
@@ -596,7 +603,56 @@ def __init__(self, main):
...
+    def __init__(self, main):
+        if "cryo" not in main.role:
+            raise ValueError(
+                "Expected a cryo microscope role but found it to be %s." % main.role)
+        MicroscopyGUIData.__init__(self, main)
+        # List VA contains all the CryoFeatures (for the current project).
+        self.features = FeaturesListVA()
+        # VA for the currently selected feature
+        # TODO: Make it also a singleton
+        self.currentFeature = model.VigilantAttribute(None)
```
pieleric, line 639: The code to make it a singleton is quite tricky (especially with the listeners), and it doesn't even support the case of currentFeature.
Let's just move these 2 VAs to MainGUIData. I see there are already some shared VA like `fineAlignDwellTime`, and it's all fine. If you want, you could put it in a `if "cryo" not in main.role:` block, so that they only appear on cryo.

pieleric (changes requested):

src/odemis/gui/cont/features.py
```diff
@@ -0,0 +1,155 @@
...
+        """
+        feature = self._tab_data_model.currentFeature.value
+        if not feature:
+            return
+        pos = feature.pos.value
+        logging.info(f"Moving to position: {pos}")
+        self._main_data_model.stage.moveAbs({'x': pos[0], 'y': pos[1]})
+        self._main_data_model.focus.moveAbs({'z': pos[2]})
+
+    def _on_features_changes(self, features):
```
Mahmood-B, line 71: This function makes changes on the GUI, so you could add the decorator `call_in_wx_main`

src/odemis/gui/cont/features.py
```diff
@@ -0,0 +1,155 @@
...
+        """
+        repopulate the feature list dropdown with the modified features
+        :param features: list(CryoFeature) list of modified features
+        """
+        self._panel.cmb_features.Clear()
+        for i, feature in enumerate(features):
+            self._panel.cmb_features.Insert(feature.name.value, i, feature)
+
+    def _on_current_feature_changes(self, feature):
```
Mahmood-B, line 80: Decorator?

src/odemis/gui/cont/features.py
```diff
@@ -0,0 +1,155 @@
...
+
+        if self._feature_z_va_connector:
+            self._feature_z_va_connector.pause()
+        self._feature_z_va_connector = VigilantAttributeConnector(feature.pos,
+                                                                  self._panel.ctrl_feature_z,
+                                                                  events=wx.EVT_TEXT_ENTER,
+                                                                  ctrl_2_va=self._on_ctrl_feature_z_change,
+                                                                  va_2_ctrl=self._on_feature_pos)
+
+    def _on_feature_pos(self, feature_pos):
```
Mahmood-B, line 113: Decorator?

src/odemis/gui/cont/features.py
```diff
@@ -0,0 +1,155 @@
...
+        # Set the feature Z ctrl with the 3rd (focus) element of the feature position
+        self._panel.ctrl_feature_z.SetValue(feature_pos[2])
+
+    def _on_feature_status(self, feature_status):
```
Mahmood-B, line 117: decorator?

src/odemis/gui/cont/features.py
```diff
@@ -0,0 +1,155 @@
...
+        Update the feature status dropdown with the feature status
+        :param feature_status: (string) the updated feature status
+        """
+        self._panel.cmb_feature_status.Clear()
+        self._panel.cmb_feature_status.Append(FEATURE_ACTIVE)
+        self._panel.cmb_feature_status.Append(FEATURE_MILLED)
+        self._panel.cmb_feature_status.Append(FEATURE_DEACTIVE)
+        self._panel.cmb_feature_status.SetValue(feature_status)
+
+    def _on_cmb_features_change(self, evt):
```
Mahmood-B, line 128: decorator ?

src/odemis/gui/cont/features.py
```diff
@@ -0,0 +1,155 @@
...
+        """
+        Change the current feature based on the feature dropdown selection
+        """
+        index = self._panel.cmb_features.GetSelection()
+        if index == -1:
+            return
+        selected_feature = self._panel.cmb_features.GetClientData(index)
+        self._tab_data_model.currentFeature.value = selected_feature
+
+    def _on_cmb_feature_status_change(self):
```
Mahmood-B, line 138: decorator?

src/odemis/gui/cont/features.py
```diff
@@ -0,0 +1,155 @@
...
+        """
+        Get current feature status dropdown value
+        :return: (string) feature status dropdown value
+        """
+        feature = self._tab_data_model.currentFeature.value
+        if feature:
+            return self._panel.cmb_feature_status.GetValue()
+
+    def _on_ctrl_feature_z_change(self):
```
Mahmood-B, line 147: Decorator?

Mahmood-B (changes requested):

src/odemis/gui/comp/overlay/world.py
```diff
@@ -111,6 +113,200 @@ def on_left_up(self, evt):
...
```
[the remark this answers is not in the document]
lazem: It doesn't have to be a big block, @Mahmood-B  I made it as a function here because I'm not sure yet this is what is only needed to detect a feature. So in case there was a requirement to change/add more criteria, it'd be added in this small function.

src/odemis/gui/model/__init__.py
```diff
@@ -708,10 +746,10 @@ def __init__(self, main):
 #         self.autofocus_active = BooleanVA(False)
 
 
-class CryoChamberGUIData(MicroscopyGUIData):
+class CryoChamberGUIData(CryoGUIData):
```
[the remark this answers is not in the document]
lazem: Yes, I have discussed this with Éric and for now we'll keep showing the overview map (with the features) in the chamber tab.

src/odemis/gui/cont/features.py
lazem: The `call_in_wx_main` decorator is called from inside the VA connector. Check the implementation of `VigilantAttributeConnector` class.

src/odemis/acq/feature.py
```diff
@@ -0,0 +1,32 @@
+from numpy import long
+
```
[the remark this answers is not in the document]
lazem: That's to distinguish between: python standard lib import, then external libs imports, then internal modules ones.

lazem:
Hi @pieleric I have committed the remaining parts for feature bookmarking, including handling when there's no selected feature, and adding the toggle button to all localization tab views. Appreciating your review.
Thanks!

src/odemis/gui/cont/streams.py
```diff
@@ -2185,6 +2185,21 @@ def removeStream(self, stream):
...
+    def clear_panel(self):
+        """
+        Clear the stream bar from its GUI controls, and remove each stream controllers stream from its view
+        Must be called in the main GUI thread
+        """
+        # clear the graphical part
+        self._stream_bar.clear()
+        # Clear the stream from the stream controller view
+        for sc in self.stream_controllers:
+            if hasattr(sc, "view") and hasattr(sc.view, "removeStream"):
```
pieleric, line 2197: Why wouldn't the stream have a .view attribute? It's set at init (possibly with None, but seems to always be set).

pieleric (changes requested):

src/odemis/gui/comp/legend.py
```diff
@@ -308,6 +309,16 @@ def __init__(self, parent, wid=wx.ID_ANY, pos=(0, 0), size=wx.DefaultSize,
...
+        # Todo: replace with actual icons
+        bmp = img.getBitmap("menu/btn_view_pick.png")
+        bmpa = img.getBitmap("menu/btn_view_pick_a.png")
+        bmph = img.getBitmap("menu/btn_view_pick_h.png")
+
+        self.featureBtn = ImageToggleButton(self, pos=(0, 0), bitmap=bmp, size=(24, 24))
+        self.featureBtn.bmpSelected = bmpa
+        self.featureBtn.bmpHover = bmph
+        self.featureBtn.Show(False)
+
```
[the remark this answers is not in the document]
lazem: I tried moving the code to the info legend, but it didn't work, somewhere the legend is reinitialized and setting the `show_features` va is lost. Any clue why this is happening with `InfoLegend`?

src/odemis/gui/cont/features.py
```diff
@@ -0,0 +1,188 @@
...
+        :param feature: (CryoFeature or None) the newly selected current feature
+        """
+        if self._feature_name_va_connector:
+            self._feature_name_va_connector.pause()
+
+        if self._feature_status_va_connector:
+            self._feature_status_va_connector.pause()
+
+        if self._feature_z_va_connector:
+            self._feature_z_va_connector.pause()
```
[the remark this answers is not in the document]
lazem: I tried using disconnect, but the `_on_ctrl_value_change` on the va connector still get called, even though the va has no listeners! Not sure why would this behavior happen..

src/odemis/gui/cont/features.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def _on_ctrl_feature_z_change(self):
+        """
+        Get the current feature Z ctrl value to set feature Z position
+        :return: (tuple of 3 floats) the full position including the Z ctrl value
+        """
+        feature = self._tab_data_model.currentFeature.value
+        if feature:
+            pos = feature.pos.value
+            return pos[0], pos[1], float(self._panel.ctrl_feature_z.Value)
```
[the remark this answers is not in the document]
lazem: The self._panel.ctrl_feature_z value is text not float, so I kept the convert.

src/odemis/gui/model/__init__.py
```diff
@@ -595,8 +603,57 @@ def __init__(self, main):
...
+        # VA for the currently selected feature
+        self.currentFeature = main.currentFeature
+
+    def add_new_feature(self, pos_x, pos_y, pos_z=None, f_name=None, milling_angle=None):
+        """
+        Create a new feature and add it to the features list
+        """
+        if not f_name:
+            existing_names = [f.name.value for f in self.features.value]
+            f_name = make_unique_name("Feature-1", existing_names)
```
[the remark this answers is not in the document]
lazem: I tried "Feature 1" at first but make_unique_name function made the 2nd one Feature 1-1, so I kept the dash.

Mahmood-B:
LGTM!

Mahmood-B (commented):
Didn't see odd things

src/odemis/gui/cont/features.py
```diff
@@ -0,0 +1,193 @@
...
+        Change the current feature based on the feature dropdown selection
+        """
+        index = self._panel.cmb_features.GetSelection()
+        if index == -1:
+            logging.warning("cmb_features selection = -1.")
+            return
+        selected_feature = self._panel.cmb_features.GetClientData(index)
+        self._tab_data_model.main.currentFeature.value = selected_feature
+
+    def _on_cmb_feature_status_change(self):
```
pieleric, line 174: Shouldn't it need an `evt` argument?

src/odemis/gui/cont/tabs.py
```diff
@@ -502,26 +506,23 @@ def _create_views(self, main_data, viewports):
...
         return vpv
 
-    def load_data(self, data):
+    def load_overview_data(self, data):
         # Create streams from data
         streams = data_to_static_streams(data)
 
+        view = self.tab_data_model.views.value[0]  # overview map view
         for s in streams:
+            s.name.value = "Overview" + " " + s.name.value
```
pieleric, line 515: `"Overview" + " "` -> `"Overview "`

src/odemis/gui/cont/tabs.py
```diff
@@ -666,27 +665,83 @@ def Show(self, show=True):
...
+        # show the feature streams on the acquired view
+        view = self.tab_data_model.views.value[1]
+        acquired_streams = feature.streams.value if feature else []
+        for stream in acquired_streams:
+            self._show_acquired_stream(stream, view)
+
+        # show the overview maps (if any) on the overview view
+        view = self.tab_data_model.views.value[0]
+        for stream in self.tab_data_model.overviewStreams.value:
+            self._show_acquired_stream(stream, view)
```
pieleric, line 697: I wonder if that's necessary. I had in mind that there are just a few overview streams, which are always visible (unless the user explicitly hide them).

src/odemis/gui/cont/tabs.py
```diff
@@ -666,27 +665,83 @@ def Show(self, show=True):
...
+    def _on_acquired_streams(self, streams):
+        """
+        Filter out deleted acquired streams (features and overview) from their respective origin
+        :param streams: list(Stream) updated list of tab streams
+        """
+        # Get all acquired streams from features list and overview streams
+        acquired_streams = set()
+        for feature in self.tab_data_model.main.features.value:
+            [acquired_streams.add(s) for s in feature.streams.value]
+        [acquired_streams.add(s) for s in self.tab_data_model.overviewStreams.value]
```
pieleric, line 708: I wouldn't recommend list comprehension if you don't use the list, it feels odd to create a list of `None`, and discard it, just for the side-effect.
Here, there is probably a more explicit way to write it:
```
        for feature in self.tab_data_model.main.features.value:
            acquired_streams.update(feature.streams.value)
        acquired_streams.update(self.tab_data_model.overviewStreams.value)
```

src/odemis/gui/cont/tabs.py
```diff
@@ -666,27 +665,83 @@ def Show(self, show=True):
...
+        for st in filter_streams:
+            if st in self.tab_data_model.overviewStreams.value:
+                self.tab_data_model.overviewStreams.value.remove(st)
+                # Remove from chamber tab too
+                chamber_tab = self.main_data.getTabByName("cryosecom_chamber")
+                chamber_tab.remove_overview_stream(st)
+            else:
+                feature = next((f for f in self.tab_data_model.main.features.value if st in f.streams.value), None)
+                if feature:
+                    feature.streams.value.remove(st)
```
pieleric, line 721: You actually want to remove the stream from *all* the features... although, normally it's only present in one of them.
You can write simpler code if you think of "remove the stream from all the features":
```
for feature in self.tab_data_model.main.features.value:
    if st in feature.streams.value:
        feature.streams.value.remove(st)
```

src/odemis/gui/cont/tabs.py
```diff
@@ -666,27 +665,83 @@ def Show(self, show=True):
...
+        Filter out deleted acquired streams (features and overview) from their respective origin
+        :param streams: list(Stream) updated list of tab streams
+        """
+        # Get all acquired streams from features list and overview streams
+        acquired_streams = set()
+        for feature in self.tab_data_model.main.features.value:
+            [acquired_streams.add(s) for s in feature.streams.value]
+        [acquired_streams.add(s) for s in self.tab_data_model.overviewStreams.value]
+
+        filter_streams = acquired_streams.difference(set(streams))
```
pieleric, line 710: Maybe "filter_steams" to "unused_streams"? It might make it clearer that these are the streams we want to remove.

src/odemis/gui/model/__init__.py
```diff
@@ -595,8 +602,51 @@ def __init__(self, main):
...
+        self.main.features.value.append(feature)
+        return feature
+
+    # Todo: find the right margin
+    ATOL_FEATURE_POS = 0.1e-3  # m
+
+    def select_current_position_feature(self):
+        """
+        Given current stage position, either select one of the features closest to the position or create a new one with the position
+        :return: (CryoFeature) the feature selected for the current position
```
pieleric, line 634: This function doesn't return anything. Instead, it updates .currentFeature. Please update the docstring.

src/odemis/gui/model/__init__.py
```diff
@@ -595,8 +602,51 @@ def __init__(self, main):
...
+            feature_dist = math.hypot(feature.pos.value[0] - current_position["x"],
+                                      feature.pos.value[1] - current_position["y"])
+            if feature_dist <= self.ATOL_FEATURE_POS:
+                self.currentFeature.value = feature
+                break
+        else:
+            # create new feature if no close feature found
+            feature = self.add_new_feature(current_position["x"], current_position["y"],
+                                                          self.main.focus.position.value["z"])
+            self.main.currentFeature.value = feature
```
pieleric, line 647: The block just above uses `self.currentFeature`, while here it's `self.main.currentFeature`. That's fishy.

src/odemis/gui/model/__init__.py
```diff
@@ -595,8 +602,51 @@ def __init__(self, main):
...
+        """
+        current_position = self.main.stage.position.value
+        for feature in self.main.features.value:
+            feature_dist = math.hypot(feature.pos.value[0] - current_position["x"],
+                                      feature.pos.value[1] - current_position["y"])
+            if feature_dist <= self.ATOL_FEATURE_POS:
+                self.currentFeature.value = feature
+                break
+        else:
+            # create new feature if no close feature found
```
pieleric, line 644: Please add a logging.debug() indicating a new feature is created at the specific position, because none are close by.

pieleric (changes requested):

src/odemis/gui/cont/features.py
lazem: Indeed but I removed it because` ctrl_2_va()` is called without passing the argument `evt`, didn't want to change it there in case it breaks somewhere else.
https://github.com/delmic/odemis/blob/44afe722ebba577bf2693a76a15eefeb17f2faab/src/odemis/gui/util/widgets.py#L101

PR #1774, opened 2021-05-25
driver-piezomotor: document behaviour of open-loop simulator

philipwinklerdelmic:
Currently, the position returned by the simulator is always 0 in open-loop. This might be the same behaviour as on the hardware if no encoder is connected (should be checked!)

### DIFF
--- a/src/odemis/driver/piezomotor.py
+++ b/src/odemis/driver/piezomotor.py
@@ -1053,6 +1053,8 @@ class PMDSimulator(object):
                 else:
                     raise ValueError()
             elif cmd == "J":
+                # TODO: check hardware behaviour in open loop. What does the E command return if no encoder is
+                #  connected? Probably 0, so don't update .current_pos here.
                 if not args:
                     if self.is_moving:
                         self._output_buf += ":222"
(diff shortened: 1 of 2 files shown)


tepals (commented):
I think it's fine like this, do you think it makes sense to add a closed loop test that does check the position?
Just a simple:
```
pos_x = 0.01
f = self.stage.moveAbs({'x': pos_x})
f.result()
self.assertEqual(self.stage.position["x"], pos_x)
```

pieleric (changes requested):
I'm not very convinced by these changes: it's mostly asking questions. It'd be more useful to put these questions in a Jira task, then to a 1 hour remote connection to the system, and find out. Then you can actually put the answers in the code ;-)

philipwinklerdelmic:
@tepals Do you think it's important enough to investigate? Since it's just a question for the simulator and only for the open-loop (which we don't use anyway), I think we can just drop the issue. Otherwise, I can make a jira-task (not sure though when anybody has time to look into this).

tepals:
I think we can drop it

PR #1775, opened 2021-05-25
Enzel acquisition panel with zstack acquisition manager 

Mahmood-B:
This PR contains all the acquisition panel including the zstack capability. All the code related to the zstack acquisition is only in the second commit (3951cfa).

### DIFF
--- a/src/odemis/acq/acqmng.py
+++ b/src/odemis/acq/acqmng.py
@@ -40,8 +40,8 @@ from odemis.util import img, fluo, executeAsyncTask
 import time
 import copy
 from odemis.model import prepare_to_listen_to_more_vas
-from concurrent.futures._base import CANCELLED, FINISHED
-from odemis.util.driver import estimate_focuser_move_duration 
+from concurrent.futures._base import CANCELLED, FINISHED, RUNNING
+from odemis.util.driver import guessActuatorMoveDuration 
 from odemis.util.img import assembleZCube
 
 # This is the "manager" of an acquisition. The basic idea is that you give it
@@ -129,19 +129,17 @@ class ZStackAcquisitionTask(object):
         self._future_lock = threading.Lock()
         self._actuator_f = None 
         self._single_acqui_f = None 
-        self._future_state = None
+        self._future_state = RUNNING 
         self._streams = sortStreams(streams)
         self._zlevels = zlevels
         self._settings_obs = settings_obs
-        self._zpos = list(zlevels.values())[0] if zlevels.values() else None
-        zstep = abs(self._zpos[0]-self._zpos[1]) if self._zpos else 0
-        # get the focuser object of any of the FM streams
-        focuser = next((s.focuser for s in self._streams if isinstance(s, FluoStream)), None)
-        self._zstep_duration = estimate_focuser_move_duration(focuser, zstep)
+        self._zstep_duration = {}
+        for s,z in zlevels.items():
+            self._zstep_duration[s] = guessActuatorMoveDuration(s.focuser, axis="z", distance=abs(z[0]-z[1]))
 
     def cancel(self, _):
         """
-        Called when the user presses the button "cancel" and the future is still running.
+        Cancels the acquisition task
         (but there might  be some delay (not instantly called) )
         return (Boolean): True if canceled, False otherwise
         """
@@ -151,8 +149,10 @@ class ZStackAcquisitionTask(object):
         self._future_state = CANCELLED
         self._future_lock.release()
         # cancel the sub-futures
-        self._actuator_f.cancel()
-        self._single_acqui_f.cancel()
+        if self._actuator_f:
+            self._actuator_f.cancel()
+        if self._single_acqui_f:
+            self._single_acqui_f.cancel()
         return True
 
     def estimate_total_duration(self):
@@ -168,11 +168,10 @@ class ZStackAcquisitionTask(object):
         """
         remaining_t = self.estimate_total_duration()
         acquired_data = []
-        exp = None
         # iterate through streams
         for stream in self._streams:
             zstack = []
-            if stream not in self._zlevels.keys(): # stream can't be zstack imaged 
+            if stream not in self._zlevels: 
                 try:
                     # acquire this single stream, and get the data
                     self._single_acqui_f = acquire([stream], self._settings_obs)
@@ -189,20 +188,20 @@ class ZStackAcquisitionTask(object):
                         logging.info("The acquisition for stream %s is done", stream)
 
                 except CancelledError:
-                    raise CancelledError()
+                    raise 
                 except Exception as e:
-                    logging.exception("The acquisition failed because of %s" %e)
+                    logging.exception("The acquisition failed because %s" %e)
                     # return what was acquired so far  
                     return acquired_data, e
-                finally:
-                    acquired_data.append(data[0])
-                    # update the remaining time
-                    remaining_t -= stream.estimateAcquisitionTime()
-                    self._main_future.set_end_time(time.time() + remaining_t)
+              
+                acquired_data.append(data[0])
+                # update the remaining time
+                remaining_t -= stream.estimateAcquisitionTime()
+                self._main_future.set_end_time(time.time() + remaining_t)
 
-            else: # stream can be zstack imaged 
-                # for each stream, iterate through zlevels (if exist)
-                for i, z in enumerate(self._zpos):
+            else: 
+                # for each stream, iterate through zlevels
+                for i, z in enumerate(self._zlevels[stream]):
                     # move the focuser
                     self._actuator_f = stream.focuser.moveAbs({"z": z})
                     self._actuator_f.result()
@@ -212,8 +211,8 @@ class ZStackAcquisitionTask(object):
                         raise CancelledError()
 
                     # subtract one zstep time
-                    if i != len(self._zpos) - 1:
-                        remaining_t -= self._zstep_duration
+                    if i != len(self._zlevels[stream]) - 1:
+                        remaining_t -= self._zstep_duration[stream]
                         self._main_future.set_end_time(time.time() + remaining_t)
 
                     try:
@@ -232,20 +231,18 @@ class ZStackAcquisitionTask(object):
                             logging.info("The acquisition for stream %s is done", stream)
 
                     except CancelledError:
-                        raise CancelledError()
-                    except Exception as exp:
-                        logging.exception("The acquisition failed at the %s-th zlevel of streams %s because of %s" %(i+1, stream, exp))
-                        # assemble what was acquired so far, and return 
-                        zcube = assembleZCube(zstack, self._zpos)
-                        acquired_data.append(zcube)
-                        return acquired_data, exp
-                    finally:
-                        zstack.append(data[0])
-                        # update the remaining time
-                        remaining_t -= stream.estimateAcquisitionTime()
-                        self._main_future.set_end_time(time.time() + remaining_t)
+                        raise
+                    except Exception as e:
+                        logging.exception("The acquisition failed at the %s-th zlevel of the stream %s, because %s" %(i+1, stream, e))
+                        # TODO handle zstack assembling in case of error 
+                        return acquired_data, e
+                    
+                    zstack.append(data[0])
+                    # update the remaining time
+                    remaining_t -= stream.estimateAcquisitionTime()
+                    self._main_future.set_end_time(time.time() + remaining_t)
 
-                zcube = assembleZCube(zstack, self._zpos)
+                zcube = assembleZCube(zstack, self._zlevels[stream])
                 acquired_data.append(zcube)
 
         # state that the future has finished
@@ -278,7 +275,7 @@ def estimateZStackAcquisitionTime(streams, zlevels):
         # get the focuser of any of the FM streams
         focuser = next((s.focuser for s in streams if isinstance(s, FluoStream)), None)
         tot_zstep_time = (len(z_pos) - 1) * len(zlevels) * \
-            estimate_focuser_move_duration(focuser, zstep)
+            guessActuatorMoveDuration(focuser, axis="z", distance=zstep)
         acq_time += tot_zstep_time
 
     return acq_time
--- a/src/odemis/acq/acqmng.py
+++ b/src/odemis/acq/acqmng.py
@@ -133,7 +133,7 @@ class ZStackAcquisitionTask(object):
         self._streams = sortStreams(streams)
         self._zlevels = zlevels
         self._settings_obs = settings_obs
-        self._zpos = list(zlevels.values())[0]
+        self._zpos = list(zlevels.values())[0] if zlevels.values() else None
         zstep = abs(self._zpos[0]-self._zpos[1]) if self._zpos else 0
         # get the focuser object of any of the FM streams
         focuser = next((s.focuser for s in self._streams if isinstance(s, FluoStream)), None)
--- a/src/odemis/acq/acqmng.py
+++ b/src/odemis/acq/acqmng.py
@@ -25,7 +25,6 @@ Odemis. If not, see http://www.gnu.org/licenses/.
 
 
 from __future__ import division
-from numpy import flipud
 from collections import OrderedDict
 import collections
 from concurrent.futures import CancelledError
--- a/src/odemis/acq/acqmng.py
+++ b/src/odemis/acq/acqmng.py
@@ -42,6 +42,7 @@ import copy
 from odemis.model import prepare_to_listen_to_more_vas
 from concurrent.futures._base import CANCELLED, FINISHED
 from odemis.util.driver import estimate_focuser_move_duration 
+from odemis.gui.util.img import assemble_z_cube
 
 # This is the "manager" of an acquisition. The basic idea is that you give it
 # a list of streams to acquire, and it will acquire them in the best way in the
@@ -162,35 +163,6 @@ class ZStackAcquisitionTask(object):
         """
         return estimateZStackAcquisitionTime(self._streams, self._zlevels)
 
-    def _construct_zcube(self, data_arrays):
-        """
-        Create 3D data array from the 2D acquired data arrays 
-        data_arrays (list of DataArray): a list of acquired images at different z levels.
-        return (DataArray): a data array embedding all the data acquired at all z levels.
-        """
-        # pixel size along z axis
-        metadata = data_arrays[0].metadata
-        px, py = metadata[model.MD_PIXEL_SIZE]
-        pz = abs(self._zpos[-1] - self._zpos[0]) / (len(self._zpos) - 1)
-
-        # upgrade the metadata with the z pixel size
-        metadata[model.MD_PIXEL_SIZE] = (px, py, round(pz, ndigits=7))
-
-        # in case the zlevels are ordered from positive to negative,
-        # flip the acquired data so that they are always ordered from
-        # lowest zlevel to highest zlevel 
-        if self._zpos:
-            if self._zpos[-1] < self._zpos[0]:
-                data_arrays = flipud(data_arrays)
-
-        # z coordinate of center
-        cx, cy = metadata[model.MD_POS]
-        cz = (self._zpos[0] + self._zpos[-1]) / 2
-        # upgrade the metadata with the z center position
-        metadata[model.MD_POS] = (cx, cy, round(cz, ndigits=5))
-
-        return model.DataArray(data_arrays, metadata)
-
     def run(self):
         """
         The main function of the task class, which will be called by the future asynchronously
@@ -228,14 +200,12 @@ class ZStackAcquisitionTask(object):
 
                     # check on the acquired data
                     if not data:
-                        logging.warning(
-                            "The acquired data array for stream %s is empty", stream
-                        )
+                        logging.warning("The acquired data array for stream %s is empty", stream)
                     else:
                         logging.info("The acquisition for stream %s is done", stream)
 
                 except CancelledError:
-                    logging.exception("The acquisition was canceled")
+                    raise CancelledError()
                 except Exception:
                     logging.exception("The acquisition failed")
                 finally:
@@ -245,7 +215,7 @@ class ZStackAcquisitionTask(object):
                     self._main_future.set_end_time(time.time() + remaining_t)
 
             if isinstance(stream, FluoStream):
-                zcube = self._construct_zcube(zstack)
+                zcube = assemble_z_cube(zstack, self._zpos)
                 acquired_data.append(zcube)
             elif isinstance(stream, EMStream):
                 acquired_data.append(zstack[0])
--- a/src/odemis/acq/acqmng.py
+++ b/src/odemis/acq/acqmng.py
@@ -41,7 +41,7 @@ import copy
 from odemis.model import prepare_to_listen_to_more_vas
 from concurrent.futures._base import CANCELLED, FINISHED
 from odemis.util.driver import estimate_focuser_move_duration 
-from odemis.gui.util.img import assemble_z_cube
+from odemis.util.img import assembleZCube
 
 # This is the "manager" of an acquisition. The basic idea is that you give it
 # a list of streams to acquire, and it will acquire them in the best way in the
@@ -232,7 +232,7 @@ class ZStackAcquisitionTask(object):
                     except Exception as exp:
                         logging.exception("The acquisition failed at the %s-th zlevel of streams %s because of %s" %(i+1, stream, exp))
                         # assemble what was acquired so far, and return 
-                        zcube = assemble_z_cube(zstack, self._zpos)
+                        zcube = assembleZCube(zstack, self._zpos)
                         acquired_data.append(zcube)
                         return acquired_data, exp
                     finally:
@@ -241,7 +241,7 @@ class ZStackAcquisitionTask(object):
                         remaining_t -= stream.estimateAcquisitionTime()
                         self._main_future.set_end_time(time.time() + remaining_t)
 
-                zcube = assemble_z_cube(zstack, self._zpos)
+                zcube = assembleZCube(zstack, self._zpos)
                 acquired_data.append(zcube)
 
         # state that the future has finished 
--- a/src/odemis/acq/acqmng.py
+++ b/src/odemis/acq/acqmng.py
@@ -29,6 +29,7 @@ from collections import OrderedDict
 import collections
 from concurrent.futures import CancelledError
 import logging
+import threading
 
 from odemis import model
 from odemis.acq import _futures
@@ -125,6 +126,7 @@ class ZStackAcquisitionTask(object):
         settings_obs (SettingsObserver): VAs to be integrated in the acquired data as metadata
         """
         self._main_future = future
+        self._future_lock = threading.Lock()
         self._actuator_f = None 
         self._single_acqui_f = None 
         self._future_state = None
@@ -143,9 +145,11 @@ class ZStackAcquisitionTask(object):
         (but there might  be some delay (not instantly called) )
         return (Boolean): True if canceled, False otherwise
         """
+        self._future_lock.acquire()
         if self._future_state is FINISHED: 
             return False
         self._future_state = CANCELLED
+        self._future_lock.release()
         # cancel the sub-futures
         self._actuator_f.cancel()
         self._single_acqui_f.cancel()
@@ -244,8 +248,10 @@ class ZStackAcquisitionTask(object):
                 zcube = assembleZCube(zstack, self._zpos)
                 acquired_data.append(zcube)
 
-        # state that the future has finished 
+        # state that the future has finished
+        self._future_lock.acquire() 
         self._future_state = FINISHED
+        self._future_lock.release()
 
         return acquired_data, exp
 
--- a/src/odemis/acq/acqmng.py
+++ b/src/odemis/acq/acqmng.py
@@ -134,8 +134,9 @@ class ZStackAcquisitionTask(object):
         self._zlevels = zlevels
         self._settings_obs = settings_obs
         self._zstep_duration = {}
-        for s,z in zlevels.items():
-            self._zstep_duration[s] = guessActuatorMoveDuration(s.focuser, axis="z", distance=abs(z[0]-z[1]))
+        for s, z in zlevels.items():
+            if len(z) > 1:
+                self._zstep_duration[s] = guessActuatorMoveDuration(s.focuser, axis="z", distance=abs(z[0] - z[1]))
 
     def cancel(self, _):
         """
@@ -143,11 +144,8 @@ class ZStackAcquisitionTask(object):
         (but there might  be some delay (not instantly called) )
         return (Boolean): True if canceled, False otherwise
         """
-        self._future_lock.acquire()
-        if self._future_state is FINISHED: 
-            return False
-        self._future_state = CANCELLED
-        self._future_lock.release()
+        with self._future_lock:
+            self._future_state = CANCELLED
         # cancel the sub-futures
         if self._actuator_f:
             self._actuator_f.cancel()
@@ -176,7 +174,8 @@ class ZStackAcquisitionTask(object):
                     # acquire this single stream, and get the data
                     self._single_acqui_f = acquire([stream], self._settings_obs)
                     data, exp = self._single_acqui_f.result()
-
+                    if exp:
+                        return acquired_data, exp
                     # check if cancellation happened while the acquiring future is working
                     if self._future_state == CANCELLED:
                         raise CancelledError()
@@ -190,11 +189,12 @@ class ZStackAcquisitionTask(object):
                 except CancelledError:
                     raise 
                 except Exception as e:
-                    logging.exception("The acquisition failed because %s" %e)
+                    logging.exception("The acquisition of stream %s failed", stream.name.value)
                     # return what was acquired so far  
                     return acquired_data, e
-              
-                acquired_data.append(data[0])
+
+                if data:
+                    acquired_data.append(data[0])
                 # update the remaining time
                 remaining_t -= stream.estimateAcquisitionTime()
                 self._main_future.set_end_time(time.time() + remaining_t)
@@ -219,7 +219,8 @@ class ZStackAcquisitionTask(object):
                         # acquire this single stream, and get the data
                         self._single_acqui_f = acquire([stream], self._settings_obs)
                         data, exp = self._single_acqui_f.result()
-
+                        if exp:
+                            return acquired_data, exp
                         # check if cancellation happened while the acquiring future is working
                         if self._future_state == CANCELLED:
                             raise CancelledError()
@@ -236,8 +237,10 @@ class ZStackAcquisitionTask(object):
                         logging.exception("The acquisition failed at the %s-th zlevel of the stream %s, because %s" %(i+1, stream, e))
                         # TODO handle zstack assembling in case of error 
                         return acquired_data, e
-                    
-                    zstack.append(data[0])
+                        
+                    # only if there is data acquired
+                    if data:
+                        zstack.append(data[0])
                     # update the remaining time
                     remaining_t -= stream.estimateAcquisitionTime()
                     self._main_future.set_end_time(time.time() + remaining_t)
@@ -246,9 +249,8 @@ class ZStackAcquisitionTask(object):
                 acquired_data.append(zcube)
 
         # state that the future has finished
-        self._future_lock.acquire() 
-        self._future_state = FINISHED
-        self._future_lock.release()
+        with self._future_lock:
+            self._future_state = FINISHED
 
         return acquired_data, exp
 
@@ -268,15 +270,12 @@ def estimateZStackAcquisitionTime(streams, zlevels):
             acq_time += s.estimateAcquisitionTime() * len(zlevels[s])
         else:
             acq_time += s.estimateAcquisitionTime() 
-    # find the zstep, if applicable
-    z_pos = next((l for l in zlevels.values() if len(l) > 1), None)
-    if z_pos:
-        zstep = abs(z_pos[0]-z_pos[1])
-        # get the focuser of any of the FM streams
-        focuser = next((s.focuser for s in streams if isinstance(s, FluoStream)), None)
-        tot_zstep_time = (len(z_pos) - 1) * len(zlevels) * \
-            guessActuatorMoveDuration(focuser, axis="z", distance=zstep)
-        acq_time += tot_zstep_time
+    for s in streams:
+        zs = zlevels.get(s, [0])  # concider that streams without zlevels are acquired at a single z
+        if len(zs) > 1 and s.focuser:
+            zstep = abs(zs[0] - zs[1])
+            tot_zstep_time = (len(zs) - 1) *  guessActuatorMoveDuration(s.focuser, axis="z", distance=zstep)
+            acq_time += tot_zstep_time
 
     return acq_time
 
--- a/src/odemis/acq/acqmng.py
+++ b/src/odemis/acq/acqmng.py
@@ -25,7 +25,7 @@ Odemis. If not, see http://www.gnu.org/licenses/.
 
 
 from __future__ import division
-
+import numpy
 from collections import OrderedDict
 import collections
 from concurrent.futures import CancelledError
@@ -40,7 +40,8 @@ from odemis.util import img, fluo, executeAsyncTask
 import time
 import copy
 from odemis.model import prepare_to_listen_to_more_vas
-
+from concurrent.futures._base import CANCELLED
+from odemis.util.driver import estimateMoveDuration
 
 # This is the "manager" of an acquisition. The basic idea is that you give it
 # a list of streams to acquire, and it will acquire them in the best way in the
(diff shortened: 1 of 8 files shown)


src/odemis/acq/acqmng.py
```diff
@@ -83,6 +84,189 @@ def acquire(streams, settings_obs=None):
...
+        future (ProgressiveFuture): the future that will execute the task asynchronously
+        streams (list of Stream): the streams to be acquired
+        zlevels (dict Stream -> list of floats): a dictionary containing the streams and the
+        corresponding lists of actuator z positions
+        settings_obs (SettingsObserver):
+        """
+        self._main_future = future
+        self._future_state = None
+        self._streams = sortStreams(streams)
+        self._zpos = next((l for l in zlevels.values() if len(l) > 1), None)
```
Kleijwegt, line 131: The combination of both the _next_ , _if_, _list comprehension/generator_ and with the _None_ don't make this line extremely readable. Maybe you can split it up into two lines?
Without running the code I can't figure out what is happening exactly here.

src/odemis/acq/acqmng.py
```diff
@@ -83,6 +84,189 @@ def acquire(streams, settings_obs=None):
...
+def acquireZStack(streams, zlevels, settings_obs=None):
+    """
+    The acquisition manager of a zstack
+    streams (list of Stream): the streams to be acquired
+    zlevels (dict Stream -> list of floats): a dictionary containing the streams and the
+    corresponding lists of actuator z positions
+    settings_obs (SettingsObserver):
+    return (ProgressiveFuture): the future that will be executing the task
+    """
+    # TODO settings observer
```
Kleijwegt, line 96: Do you mean create settings observer? Test settings observer? Update settings of the settings observer? One verb would help the next reader of this code.

src/odemis/acq/acqmng.py
```diff
@@ -83,6 +84,189 @@ def acquire(streams, settings_obs=None):
...
+        """
+        Called when the user presses the button "cancel" and the future is still running.
+        (but there might  be some delay (not instantly called) )
+        return (Boolean): True if canceled, False otherwise
+        """
+        self._future_state = CANCELLED
+        # cancel the sub-futures
+        self._actuator_f.cancel()
+        self._single_acqui_f.cancel()
+        return True
```
Kleijwegt, line 146: I think this part usually also allows to return a False if the future already finished. Like in "autofocus.py" - _cancelSparc2ManualFocus line 767.
But maybe that is a different part of the future (I didn't have time to check that extensively)

src/odemis/acq/acqmng.py
```diff
@@ -83,6 +84,189 @@ def acquire(streams, settings_obs=None):
...
+        data_arrays (list of DataArray): a list of acquired images at different z levels.
+        return (DataArray): a data array embedding all the data acquired at all z levels.
+        """
+        # pixel size along z axis
+        metadata = data_arrays[0].metadata
+        px, py = metadata[model.MD_PIXEL_SIZE]
+        pz = (self._zpos[-1] - self._zpos[0]) / (len(self._zpos) - 1)
+
+        if pz < 0:
+            data_arrays = numpy.flipud(data_arrays)
```
Kleijwegt, line 196: Maybe add a comment why the arrays are flipped/mirrored. Maybe this is also a particular part of the functionality which should be written in the docstring?

src/odemis/gui/cont/acquisition.py
```diff
@@ -735,19 +743,57 @@ def _on_streams_change(self, streams):
...
+            nr_FM_streams = 0
+            for s in self._acquiStreams.value:
+                acq_time += s.estimateAcquisitionTime() * len(self._zlevels[s])  
+                acq_time = math.ceil(acq_time)              
+                if isinstance(s, FluoStream):
+                    nr_FM_streams += 1
+
+            z_pos = next((l for l in self._zlevels.values() if len(l) > 1), None)
+            if z_pos:
+                zstep = abs(z_pos[0]-z_pos[1])
```
lazem, line 771: Wouldn't `zstep = abs(self._tab_data.zStep.value)` be more readable?

lazem, line 771: A comment about what the code is trying to do would be nice. Specially that this estimation is for FM streams only.

src/odemis/acq/acqmng.py
```diff
@@ -83,6 +84,189 @@ def acquire(streams, settings_obs=None):
...
+        acqui_time = 0.0
+        zstep_time = 0
+        nr_FM_streams = 0
+        for s in self._streams:
+            acqui_time += s.estimateAcquisitionTime() * len(self._zlevels[s])
+            if isinstance(s, FluoStream):
+                nr_FM_streams += 1
+        if self._zpos:
+            zstep_time = self._zstep_duration * (len(self._zpos) - 1) * nr_FM_streams
+        return acqui_time + zstep_time
```
lazem, line 182: This code is nearly duplicated in acquisition.py. I suggest you use this code for both cases, you can take inspiration from the tiled acquisition task, please have a look at `estimateTiledAcquisitionTime()` function and its usage in the OverviewAcquisitionDialog.

src/odemis/acq/test/acq_test.py
```diff
@@ -486,5 +484,302 @@ def on_progress_update(self, future, start, end):
...
+            name="test_SEM_focus",
+            role="focus",
+            port="/dev/fake3",
+            axes=["", "", "z"],
+            ustepsize=[0, 0, 0.5e-4],
+            rng=[
+                [-0.5, 0.5],
+            ],
+            refproc="Standard",
+        )
```
lazem, line 531: Why defining additional components If you are using the enzel simulator? It already has focuser for both FM & SEM.

src/odemis/acq/acqmng.py
```diff
@@ -83,6 +84,189 @@ def acquire(streams, settings_obs=None):
...
+    def cancel(self, _):
+        """
+        Called when the user presses the button "cancel" and the future is still running.
+        (but there might  be some delay (not instantly called) )
+        return (Boolean): True if canceled, False otherwise
+        """
+        self._future_state = CANCELLED
+        # cancel the sub-futures
+        self._actuator_f.cancel()
+        self._single_acqui_f.cancel()
```
lazem, line 145: Please initialize these at `__init__` so it won't give an exception if cancel is called immediately.

src/odemis/acq/acqmng.py
lazem: Indeed, you can also check `_cancelAcquisition` of the `TiledAcquisitionTask`.

lazem (changes requested):
Initial comments, looks good overall. I like the amount of tests you have added for the generate_zlevels and ZStackAcquisitionTask!
Also note there's an issue with the UI controls for Zmin, Zstep and Zmax, they get stuck when the arrows are used.

src/odemis/acq/acqmng.py
Mahmood-B: I had a look at the implementations that you, both, mentioned. Also in the implementation of the future class in the concurrent library they do this. But why actually? I don't totally get why we return False or True? Since we do not use the return value in the acquisition controller. Do you have explanation maybe?

src/odemis/util/comp.py
```diff
@@ -107,4 +108,39 @@ def get_fov_rect(comp, fov):
...
+    n = (zrange[1] - zrange[0]) / abs(zstep) + 1
+    step_f = (zrange[1] - zrange[0]) / (floor(n) - 1)
+    step_c = (zrange[1] - zrange[0]) / (ceil(n) - 1)
+    ef = abs(step_f - abs(zstep))
+    ec = abs(step_c - abs(zstep))
+    n = floor(n) if ef < ec else ceil(n)
+
+    if zstep > 0:
+        zlevels = linspace(zrange[0], zrange[1], n)
+        return round(zlevels + actuator.position.value["z"], decimals=8)
```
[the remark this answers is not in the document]
Mahmood-B: If you want to round an extremely small number, then yes it is necessary. Because the default value of the `decimals` argument is 0. Which means that if you have a value in micrometers then it will be rounded to zero. So we need to include the main number which will be after 5 digits. However, the rounding itself is not essential. It is for better readability, especially when debugging.

Mahmood-B:
> Also note there's an issue with the UI controls for Zmin, Zstep and Zmax, they get stuck when the arrows are used.

@lazem What do you mean precisely? The arrows on the keyboard? I press them while the cursor is in the z parameters widgets, nothing wrong happens

src/odemis/gui/cont/acquisition.py
```diff
@@ -735,19 +743,57 @@ def _on_streams_change(self, streams):
...
+        """
+        return (float > 0): estimated time (in s) that it takes to move the focus actuator
+          by one step.
+        """
+        # get the focuser of any of the FM streams
+        focuser = next(
+            (s.focuser for s in self._acquiStreams.value if isinstance(s, FluoStream)), None
+        )
+        speed = None
+        if focuser is not None:
```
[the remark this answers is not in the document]
Mahmood-B: Can you elaborate? I did not get the point

src/odemis/gui/cont/acquisition.py
Mahmood-B: No actually that would not be accurate. Because we do not always use exactly the same zstep requested by the user. Eventually we use zstep that is closest to the requested step and that gives smaller error. So the estimated time should reflect the time that would be actually taken during acquisition, not the ideal time

src/odemis/acq/acqmng.py
Mahmood-B: That is good point. I looked into it again, we actually do not need to flip the arrays here. The function that generate the zlevels, takes care of flipping the zlevels list in case the zstep is negative, and then in the acquisition manager, we acquire based on that order. So the flipping is already handled

src/odemis/acq/test/acq_test.py
Mahmood-B: Indeed good point

src/odemis/acq/acqmng.py
Mahmood-B: I just talked with Eric about this point. It turned out that we always need to sequence the acquired data from negative zlevel to positive zlevel. So in case the list is reversed, we flip the data array

lazem:
> > Also note there's an issue with the UI controls for Zmin, Zstep and Zmax, they get stuck when the arrows are used.
> 
> @lazem What do you mean precisely? The arrows on the keyboard? I press them while the cursor is in the z parameters widgets, nothing wrong happens

I checked it again. Zmin is always 0, and the arrows typically increase the value of the widget by a small amount defined in the key_step argument. Check usage of other UnitFloatCtrl in Odemis.

src/odemis/gui/cont/acquisition.py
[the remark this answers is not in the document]
lazem: If focser is None, speed will also be None, and in that case it would take 10e-6, even though there is no component for this speed!

src/odemis/acq/acqmng.py
lazem: The return value is for confirming cancellation. If the movement is finished, then cancel didn't do anything and it shouldn't say it's  cancelled (ie True), right?

Mahmood-B:
> > > Also note there's an issue with the UI controls for Zmin, Zstep and Zmax, they get stuck when the arrows are used.
> > 
> > 
> > @lazem What do you mean precisely? The arrows on the keyboard? I press them while the cursor is in the z parameters widgets, nothing wrong happens
> 
> I checked it again. Zmin is always 0, and the arrows typically increase the value of the widget by a small amount defined in the key_step argument. Check usage of other UnitFloatCtrl in Odemis. @Mahmood-B

I see. Indeed there is issue about this. I will try to solve it

src/odemis/gui/cont/acquisition.py
```diff
@@ -850,12 +829,16 @@ def _on_zstack(self):
         self._zlevels = {}
         for s in self._acquiStreams.value:
             if isinstance(s, FluoStream):
+                if not self._tab_data.main.focus:
+                    raise LookupError("Focuser was not found in the tab data")
```
pieleric, line 833: Having a FM acquisition without focus would be really odd. Actually, I'll make a PR for a extra check at init that will complain if this component is missing. So no need to check for it.

src/odemis/gui/main.xrc
```diff
@@ -5582,7 +5582,7 @@
                                                     <object class="UnitFloatCtrl" name="param_Zmax">
                                                         <size>-1,15</size>
                                                         <value>1</value>
-                                                        <key_step>0.1</key_step>
+                                                        <key_step>0.000001</key_step>
                                                         <min>0</min>
                                                         <max>1000</max>
```
pieleric, line 5587: The min/max of these controls don't match the VAs. You could either hard-code the same value here (as in the gui.model.__init__), or when connecting them, update the range of the UnitFloatCtrl using `.SetValueRange()`.

src/odemis/util/comp.py
```diff
@@ -124,23 +124,31 @@ def generate_zlevels(actuator, zrange, zstep):
         raise ZeroDivisionError("The step size 'Zstep' can not be zero")
     if zrange[0] == zrange[1] == 0:
         raise ValueError("'zmax' and 'zmin' can not be both zeros")
+    if "z" not in focuser.axes.keys():
+        raise KeyError("The focus actuator %s does not have z axis", focuser)
```
pieleric, line 128: In an Exception, the error message must be a single argument => use `%` instead of `,`.

src/odemis/acq/acqmng.py
```diff
@@ -25,7 +25,7 @@
 
 
 from __future__ import division
-import numpy
+from numpy import flipud
```
pieleric: I'd rather have just `numpy` imported, and write explicitly `numpy.flipud()`, otherwise the function seems a little coming out of a magical place.

src/odemis/acq/acqmng.py
```diff
@@ -83,6 +84,189 @@ def acquire(streams, settings_obs=None):
...
+        if self._zpos:
+            zstep_time = self._zstep_duration * (len(self._zpos) - 1) * nr_FM_streams
+        return acqui_time + zstep_time
+
+    def _construct_zcube(self, data_arrays):
+        """
+        Create 3D data array from the 2D acquired data arrays 
+        data_arrays (list of DataArray): a list of acquired images at different z levels.
+        return (DataArray): a data array embedding all the data acquired at all z levels.
+        """
```
pieleric, line 170: This code is nearly identical as `_tiledacq._assembleZCube()`. Please move the function to `util.img.assembleZCube()`. In general, I think the code from tiledacq is safer, so I'd suggest to use it as basis.

src/odemis/acq/acqmng.py
```diff
@@ -83,6 +84,189 @@ def acquire(streams, settings_obs=None):
...
+            data_arrays = numpy.flipud(data_arrays)
+            pz = -pz
+        # upgrade the metadata with the z pixel size
+        metadata[model.MD_PIXEL_SIZE] = (px, py, round(pz, ndigits=7))
+
+        # z coordinate of center
+        cx, cy = metadata[model.MD_POS]
+        cz = (self._zpos[0] + self._zpos[-1]) / 2
+        # upgrade the metadata with the z center position
+        metadata[model.MD_POS] = (cx, cy, round(cz, ndigits=5))
```
pieleric, line 190: These `round()` are very fishy. We really don't want this normally. What's the purpose? I guess it's because you'd like to get the same metadata for all the (FM) streams? If so, this is going to only work *sometimes*, and it's going to back hunt us for many other cases! If you want the same metadata for all the streams, I'd suggest to just do it explicitly in a separate function, after generating the cube DataArrays. That said, I don't think that would be nice anyway, because in some cases that precision might be needed. Possibly, check with an application specialist.

src/odemis/acq/acqmng.py
```diff
@@ -83,6 +84,189 @@ def acquire(streams, settings_obs=None):
...
+
+                    # check if cancellation happened while the acquiring future is working
+                    if self._future_state == CANCELLED:
+                        raise CancelledError()
+
+                    # check on the acquired data
+                    if not data:
+                        logging.warning(
+                            "The acquired data array for stream %s is empty", stream
+                        )
```
pieleric, line 233: A single line for this would be more readable, even if it slightly goes over the recommended line width.

src/odemis/acq/acqmng.py
```diff
@@ -83,6 +84,189 @@ def acquire(streams, settings_obs=None):
...
+        """
+        The main function of the task class, which will be called by the future asynchronously
+        """
+        remaining_t = self.estimate_total_duration()
+        acquired_data = []
+        # iterate through streams
+        for stream in self._streams:
+            zstack = []
+            # for each stream, iterate through zlevels (if exist)
+            for i, z in enumerate(self._zlevels[stream]):
```
pieleric, line 205: The specification was much more fancy in terms of zlevels meaning. At least, I'd say that it should support the stream not being in zlevels. In this case, the stream is acquired just once.

src/odemis/acq/acqmng.py
```diff
@@ -83,6 +84,189 @@ def acquire(streams, settings_obs=None):
...
+                    zstack.append(data[0])
+                    # update the remaining time
+                    remaining_t -= stream.estimateAcquisitionTime()
+                    self._main_future.set_end_time(time.time() + remaining_t)
+
+            if isinstance(stream, FluoStream):
+                zcube = self._construct_zcube(zstack)
+                acquired_data.append(zcube)
+            elif isinstance(stream, EMStream):
+                acquired_data.append(zstack[0])
```
pieleric, line 251: That's way too specific. It should handle every type of stream the same way. If it has more than one zlevel -> construct Z cube.

src/odemis/acq/acqmng.py
```diff
@@ -83,6 +84,189 @@ def acquire(streams, settings_obs=None):
...
+                # if FM stream, then subtract one zstep time
+                if isinstance(stream, FluoStream):
+                    if i != len(self._zpos) - 1:
+                        remaining_t -= self._zstep_duration
+                        self._main_future.set_end_time(time.time() + remaining_t)
+
+                try:
+                    # acquire this single stream, and get the data
+                    self._single_acqui_f = acquire([stream], self._settings_obs)
+                    data, exp = self._single_acqui_f.result()
```
pieleric, line 223: You need to do something with the exception `exp`. Basically, if it's not None, you should stop now, pack the data which has already been acquired and return the data (so far) + exp.

src/odemis/acq/acqmng.py
```diff
@@ -83,6 +84,189 @@ def acquire(streams, settings_obs=None):
...
+                        logging.warning(
+                            "The acquired data array for stream %s is empty", stream
+                        )
+                    else:
+                        logging.info("The acquisition for stream %s is done", stream)
+
+                except CancelledError:
+                    logging.exception("The acquisition was canceled")
+                except Exception:
+                    logging.exception("The acquisition failed")
```
pieleric, line 240: Store the exception in exp.

src/odemis/acq/acqmng.py
```diff
@@ -83,6 +84,189 @@ def acquire(streams, settings_obs=None):
...
+                    # check on the acquired data
+                    if not data:
+                        logging.warning(
+                            "The acquired data array for stream %s is empty", stream
+                        )
+                    else:
+                        logging.info("The acquisition for stream %s is done", stream)
+
+                except CancelledError:
+                    logging.exception("The acquisition was canceled")
```
pieleric, line 238: If it's cancelled, not need to log an exception, it's just something normal which has happened. However, the function shouldn't return, instead, the function should raise a CancelledError -> add a `raise` here.

pieleric (changes requested):
(not a complete review, just looked at the main z stack acquisition task)

src/odemis/gui/cont/acquisition.py
```diff
@@ -850,12 +829,16 @@ def _on_zstack(self):
...
                 levels = generate_zlevels(
                     self._tab_data.main.focus,
                     [self._tab_data.zMin.value, self._tab_data.zMax.value],
                     self._tab_data.zStep.value)
                 self._zlevels[s] = list(levels)
             elif isinstance(s, EMStream):
+                if not s.focuser:
+                    raise LookupError("The stream %s does not have SEM focuser object", s)
```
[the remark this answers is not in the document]
Mahmood-B: I'm wondering, would it fail just because there is SEM stream? The SEM stream also has a focuser component. Or is that only in simulation? Technically, this is not essentially important, but i added these 2 if statements for readability and for understanding that the zstack is only for FM

src/odemis/acq/acqmng.py
Mahmood-B: Okay, i will make separate PR for this change, as the changes in the _tiledacq.py is not relevant to enzel acquisition panel

src/odemis/acq/acqmng.py
Mahmood-B: The only purpose is to have more readable report of the metadata after acquisition. Please look at the image attached. It is a snippet of the meta data of an acquired stream. In my opinion, the pixel size is much more compact than the depth of field.
Anyway, it is now removed, because i will reuse the function assemble z cube from _tiledacq.py

meta data

src/odemis/acq/acqmng.py
Mahmood-B: I did not get the point. Could you please elaborate?

pieleric: The spec says "zlevels (dict Actuator → list of floats, or dict Stream → list of floats): In both cases, the list of float is the positions of the z, in order. They must be spread regularly.  If Actuator → list, there should be only one entry. The actuator is used to move the focus, it should have a “z” axis. In such case, for each zlevel in the list, all the streams are acquired. If it’s Stream → list, each stream is acquired serially. For the streams which are in the zlevels dict, the .focuser of the stream is used to change to each zlevel designated. **The streams which are not in the zlevels dict are acquired only once.** (In other words, the FM streams are acquired at each zlevel, and the SEM streams are acquired only once). "

So you have to be extra careful at this line, as `zelevels` may not contain the stream.

src/odemis/acq/acqmng.py
Mahmood-B: I'm wondering, if an error happened during the waiting for the result, the exception statement will be executed, right? So i return the data and exception in that statement. Is that what you meant? Please check it in the latest commits.

src/odemis/gui/cont/acquisition.py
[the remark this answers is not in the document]
Mahmood-B: Anyway, this was removed, because the SEM stream is not included in the zlevels dictionary anymore

src/odemis/acq/acqmng.py
Mahmood-B: Indeed, i made some changes. Now the SEM are not included in the zlevels dictionary. Please have a look at the latest commits

src/odemis/acq/acqmng.py
Mahmood-B: Also in the latest commits

Mahmood-B:
@pieleric After i looked into concurrency for my presentation, i thought i should do these changes:
1. Add lock object for the future state
2. do the enabling/disabling/showing/hiding of widgets after start acquisition, to make use of concurrency and really run in parallel

src/odemis/acq/acqmng.py
```diff
@@ -83,6 +85,205 @@ def acquire(streams, settings_obs=None):
...
+                            logging.warning("The acquired data array for stream %s is empty", stream)
+                        else:
+                            logging.info("The acquisition for stream %s is done", stream)
+
+                    except CancelledError:
+                        raise CancelledError()
+                    except Exception as exp:
+                        logging.exception("The acquisition failed at the %s-th zlevel of streams %s because of %s" %(i+1, stream, exp))
+                        # assemble what was acquired so far, and return 
+                        zcube = assembleZCube(zstack, self._zpos)
```
pieleric, line 239: That probably won't work as it probably failed before the very last step. So zlevels (zpos here) will not be the same length as ztack. So you'd need to pass a shorten zlevels. Also you need to handle the case of zstack being empty (not sure if assembleZCube already handles this).
An alternative  would be to not support the ztack reconstruction in case of error for now (and instead just put a TODO).

src/odemis/acq/acqmng.py
```diff
@@ -83,6 +85,205 @@ def acquire(streams, settings_obs=None):
...
+        streams (list of Stream): the streams to be acquired
+        zlevels (dict Stream -> list of floats): a dictionary containing the streams and the
+        corresponding lists of actuator z positions
+        settings_obs (SettingsObserver): VAs to be integrated in the acquired data as metadata
+        """
+        self._main_future = future
+        self._future_lock = threading.Lock()
+        self._actuator_f = None 
+        self._single_acqui_f = None 
+        self._future_state = None
```
pieleric, line 132: None -> RUNNING.

src/odemis/acq/acqmng.py
```diff
@@ -83,6 +85,205 @@ def acquire(streams, settings_obs=None):
...
+        self._streams = sortStreams(streams)
+        self._zlevels = zlevels
+        self._settings_obs = settings_obs
+        self._zpos = list(zlevels.values())[0] if zlevels.values() else None
+        zstep = abs(self._zpos[0]-self._zpos[1]) if self._zpos else 0
+        # get the focuser object of any of the FM streams
+        focuser = next((s.focuser for s in self._streams if isinstance(s, FluoStream)), None)
+        self._zstep_duration = estimate_focuser_move_duration(focuser, zstep)
```
pieleric, line 140: Both the zstep and the focuser might be different per stream. So it would be better to compute a zstep duration per stream. => Create a dict Stream -> duration.

src/odemis/acq/acqmng.py
```diff
@@ -83,6 +85,205 @@ def acquire(streams, settings_obs=None):
...
+
+    def cancel(self, _):
+        """
+        Called when the user presses the button "cancel" and the future is still running.
```
pieleric, line 144: It's better to not write the docstring about how currently it's used, but what it does and when it should be called (in general). So here I'd say "Cancels the acquisition task". You can mention it's the canceller of the ProgressiveFuture, as it's fixed by the class, but let's not mention the GUI.

pieleric (changes requested):

src/odemis/acq/acqmng.py
```diff
@@ -83,6 +85,205 @@ def acquire(streams, settings_obs=None):
...
+        future (ProgressiveFuture): the future that will execute the task asynchronously
+        streams (list of Stream): the streams to be acquired
+        zlevels (dict Stream -> list of floats): a dictionary containing the streams and the
+        corresponding lists of actuator z positions
+        settings_obs (SettingsObserver): VAs to be integrated in the acquired data as metadata
+        """
+        self._main_future = future
+        self._future_lock = threading.Lock()
+        self._actuator_f = None 
+        self._single_acqui_f = None
```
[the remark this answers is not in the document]
Mahmood-B: I would choose the second option, because the first one is not so accurate. The type of the future that will be given to self._single_acqui_f is actually ProgressiveFuture, not InstantaneousFuture.

src/odemis/util/comp.py
```diff
@@ -107,4 +108,47 @@ def get_fov_rect(comp, fov):
...
+        raise KeyError("The focus actuator %s does not have z axis" %focuser)
+
+    focuser_pos = focuser.position.value["z"]
+    focuser_rng = focuser.axes["z"].range
+    # clip the zMax and zMin to the actuator limits if necessary 
+    if zrange[1] + focuser_pos > focuser_rng[1]:
+        zrange[1] = focuser_rng[1] - focuser_pos
+    elif focuser_pos + zrange[0] < focuser_rng[0]:
+        zrange[0] = focuser_rng[0] - focuser_pos
+
```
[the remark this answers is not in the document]
Mahmood-B: i got a bit confused about this comment and the other one above

src/odemis/util/test/comp_test.py
```diff
@@ -113,6 +115,123 @@ def test_camera_fov(self):
...
+        self.focus.position.value["z"] = 1000e-6
+        self.zMax.value = 100e-6
+        self.zMin.value = -250e-6
+        self.zrange = [self.zMin.value, self.zMax.value]
+        self.zstep.value = -50e-6
+        expected = round(asarray([-250e-6, -200e-6, -150e-6, -100e-6, -50e-6,
+                           0e-6, 50e-6, 100e-6]) + self.focus.position.value["z"], decimals=8)
+        actual = generate_zlevels(self.focus, self.zrange, self.zstep.value)
+        self.assertListEqual(
+            sorted(list(expected), reverse=True), list(actual))
```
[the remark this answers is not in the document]
Mahmood-B: Yes true, for floating point errors. But indeed it makes more sense to use rounding in the testing, not in the main function implementation.

src/odemis/util/comp.py
```diff
@@ -107,4 +108,47 @@ def get_fov_rect(comp, fov):
...
\ No newline at end of file
+            center[1] + fov[1] / 2)  # bottom
+
+
+def generate_zlevels(focuser, zrange, zstep):
+    """
+    Calculates the zlevels for a zstack acquisition, using the zmax, zmin
+    and zstep obtained from the GUI, as well as the current focus z position
+    focuser (Actuator): focus component
+    zrange (list or tuple of floats): contains the zMin and zMax, respectively
```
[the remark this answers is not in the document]
Mahmood-B: No i don't. Now i added a check on this, and it raises error if they are not (min, max).

src/odemis/acq/acqmng.py
```diff
@@ -83,6 +85,205 @@ def acquire(streams, settings_obs=None):
...
+                    return acquired_data, e
+                finally:
+                    acquired_data.append(data[0])
+                    # update the remaining time
+                    remaining_t -= stream.estimateAcquisitionTime()
+                    self._main_future.set_end_time(time.time() + remaining_t)
+
+            else: # stream can be zstack imaged 
+                # for each stream, iterate through zlevels (if exist)
+                for i, z in enumerate(self._zpos):
```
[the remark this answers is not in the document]
Mahmood-B: Okay, i could write it in this way. However, the zlevels can not be different for each stream. They are always the same for all streams. You set fixed zmax, zmin, zstep, you choose 1 or multiple streams, and then you acquire. All the streams are acquired on the same zlevels.

src/odemis/acq/acqmng.py
```diff
@@ -83,6 +85,205 @@ def acquire(streams, settings_obs=None):
...
+        """
+        self._main_future = future
+        self._future_lock = threading.Lock()
+        self._actuator_f = None 
+        self._single_acqui_f = None 
+        self._future_state = None
+        self._streams = sortStreams(streams)
+        self._zlevels = zlevels
+        self._settings_obs = settings_obs
+        self._zpos = list(zlevels.values())[0] if zlevels.values() else None
```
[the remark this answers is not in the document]
Mahmood-B: Again, streams can not have different zlevels.

src/odemis/acq/acqmng.py
Mahmood-B: But the task did not start running yet at this point. It should be the beginning of the run() function, if i'm not mistaken?

Mahmood-B:
@pieleric I finished most of the changes except two which just need a bit clarification as in the comments above. 

One more thing. while i was testing for those final changes, i noticed that when the zstep is too small, MemoryError is raised which leads the gui to get stuck and close. The reason is, too small zstep (~zero) leads to too large number of levels when we use numpy.linspace. The memory is filled (i can see it in top right of my desktop). My question is should we handle this case? And how? Or should we log warning only?

pieleric:
>One more thing. while i was testing for those final changes, i noticed that when the zstep is too small, MemoryError is raised which leads the gui to get stuck and close. The reason is, too small zstep (~zero) leads to too large number of levels when we use numpy.linspace. The memory is filled (i can see it in top right of my desktop). My question is should we handle this case? And how? Or should we log warning only?

Indeed, that's something which is worth handling. In practice, I don't think the number of z-levels should ever be really big. A ztack with 100 levels is already really big. I'd suggest that for now, we raise a IndexError in `generate_zlevels()` if there would me more than 500 zlevels. (Put it in a constant, like MAX_ZLEVELS = 500, and add a note to the docstring).
In practice, with a large number of zlevels, before reaching out of memory in linspace, we'd first hit a memory issue when acquiring each image. The current zstack acquisition function assumes everything will fit in memory. If each zlevel takes 5Mb, we could get in trouble above 500 levels. If we needed to handle more levels, we'd need to do the same as for the tile acquisition, which estimates the memory needed, and compares to the memory available on the computer. But for now, let's not worry about this.

src/odemis/acq/acqmng.py
pieleric: This state is there essentially just for checking it's not cancelled/finished. So you can use RUNNING all the time. It's better than None, which is not of the correct type.

src/odemis/acq/acqmng.py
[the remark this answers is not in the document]
pieleric: That's not correct for this function. The parameters allow to have a different zlevels per stream. Right now, the ENZEL GUI doesn't provide such flexibility, so indeed, in practice all the zlevels for now would be the same.
Please write this function without thinking how the GUI will use it, just how the parameters allow to use it. In practice, it should be pretty much as simple to deal with separate zlevels for each stream, so let's fully embrace this.

src/odemis/acq/acqmng.py
```diff
@@ -83,6 +85,202 @@ def acquire(streams, settings_obs=None):
...
+
+                    # subtract one zstep time
+                    if i != len(self._zlevels[stream]) - 1:
+                        remaining_t -= self._zstep_duration[stream]
+                        self._main_future.set_end_time(time.time() + remaining_t)
+
+                    try:
+                        # acquire this single stream, and get the data
+                        self._single_acqui_f = acquire([stream], self._settings_obs)
+                        data, exp = self._single_acqui_f.result()
```
pieleric, line 221: If exp, return with exception (same comment as above).

src/odemis/acq/acqmng.py
```diff
@@ -83,6 +85,202 @@ def acquire(streams, settings_obs=None):
...
+
+                        # check if cancellation happened while the acquiring future is working
+                        if self._future_state == CANCELLED:
+                            raise CancelledError()
+
+                        # check on the acquired data
+                        if not data:
+                            logging.warning("The acquired data array for stream %s is empty", stream)
```
pieleric, line 229: You show a warning here, but in such case the line `zstack.append(data[0])` will completely fail. => don't make that line fail.

src/odemis/acq/acqmng.py
```diff
@@ -83,6 +85,202 @@ def acquire(streams, settings_obs=None):
...
+        self._future_lock = threading.Lock()
+        self._actuator_f = None 
+        self._single_acqui_f = None 
+        self._future_state = RUNNING 
+        self._streams = sortStreams(streams)
+        self._zlevels = zlevels
+        self._settings_obs = settings_obs
+        self._zstep_duration = {}
+        for s,z in zlevels.items():
+            self._zstep_duration[s] = guessActuatorMoveDuration(s.focuser, axis="z", distance=abs(z[0]-z[1]))
```
pieleric, line 138: Please, mind the spaces. 
For commas: 1 space after, such as `a, b` .
For operators: space before & after, such as `a - b`.

src/odemis/acq/acqmng.py
```diff
@@ -83,6 +85,202 @@ def acquire(streams, settings_obs=None):
...
+            acq_time += s.estimateAcquisitionTime() 
+    # find the zstep, if applicable
+    z_pos = next((l for l in zlevels.values() if len(l) > 1), None)
+    if z_pos:
+        zstep = abs(z_pos[0]-z_pos[1])
+        # get the focuser of any of the FM streams
+        focuser = next((s.focuser for s in streams if isinstance(s, FluoStream)), None)
+        tot_zstep_time = (len(z_pos) - 1) * len(zlevels) * \
+            guessActuatorMoveDuration(focuser, axis="z", distance=zstep)
+        acq_time += tot_zstep_time
```
pieleric, line 279: This doesn't match exactly the actual acquisition code now we use a different focuser for each stream.

In practice, it's probably possible to even simplify the code. Something that would follow more the structure of the acquisition code:
```
for s in streams:
   zs = zlevels.get(s, [0])  # concider that streams without zlevels are acquired at a single z
   acq_time += s.estimateAcquisitionTime() * len(zs)
   if len(zs) > 1 and s.focuser:
       zstep = abs(zs[0] - zs[1])
       tot_zstep_time = (len(zs) - 1) *  guessActuatorMoveDuration(s.focuser, axis="z", distance=zstep)
       acq_time += tot_zstep_time
```

src/odemis/acq/test/acq_test.py
```diff
@@ -36,14 +36,16 @@
...
+from odemis.driver.tmcm import TMCLController
+from odemis.util.comp import generate_zlevels
 
 logging.getLogger().setLevel(logging.DEBUG)
 
 CONFIG_PATH = os.path.dirname(odemis.__file__) + "/../../install/linux/usr/share/odemis/"
 SPARC_CONFIG = CONFIG_PATH + "sim/sparc-pmts-sim.odm.yaml"
 SECOM_CONFIG = CONFIG_PATH + "sim/secom-sim.odm.yaml"
-
+ENZEL_CONFIG = CONFIG_PATH + "sim/enzel-sim.odm.yaml"
```
pieleric, line 49: The convention is to leave 2 empty lines before a class.

src/odemis/acq/test/acq_test.py
```diff
@@ -486,5 +484,306 @@ def on_progress_update(self, future, start, end):
...
+        cls.ccd = model.getComponent(role="ccd")
+        cls.light = model.getComponent(role="light")
+        cls.light_filter = model.getComponent(role="filter")
+        cls.ebeam = model.getComponent(role="e-beam")
+        cls.sed = model.getComponent(role="se-detector")
+        cls.SEM_focuser = model.getComponent(role="ebeam-focus")
+        cls.FM_focuser = model.getComponent(role="focus")
+
+        cls.FM_focus_pos = 0.5e-6  # arbitrary current focus position
+        cls.SEM_focus_pos = 0.1  # arbitrary current position
```
pieleric, line 513: please use lower case only for the names of attributes.

src/odemis/acq/test/acq_test.py
```diff
@@ -486,5 +484,306 @@ def on_progress_update(self, future, start, end):
...
+
+    @classmethod
+    def tearDownClass(cls):
+        if cls.backend_was_running:
+            return
+        del cls.light
```
pieleric, line 519: That line shouldn't be needed. Does it help in any way?

src/odemis/acq/test/acq_test.py
```diff
@@ -486,5 +484,306 @@ def on_progress_update(self, future, start, end):
...
+        self.FM_focuser.moveAbs({"z": self.FM_focus_pos}).result()
+        self.SEM_focuser.moveAbs({"z": self.SEM_focus_pos}).result()
+
+    @unittest.skip("Only zstack")
+    def test_only_FM_streams_without_zstack(self):
+        # create streams
+        s1 = stream.FluoStream(
+            "fluo1", self.ccd, self.ccd.data, self.light, self.light_filter
+        )
+        s1._focuser = self.FM_focuser
```
pieleric, line 538: You should set the focuser by passing `focuser = self.fm_focuser` in the initialisation.

src/odemis/acq/test/acq_test.py
```diff
@@ -486,5 +484,306 @@ def on_progress_update(self, future, start, end):
...
+    def test_FM_and_SEM_without_zstack(self):
+        # create streams
+        s1 = stream.FluoStream(
+            "fluo1", self.ccd, self.ccd.data, self.light, self.light_filter
+        )
+        s1._focuser = self.FM_focuser
+        s1.excitation.value = sorted(s1.excitation.choices)[0]
+
+        sems = stream.SEMStream("sem", self.sed, self.sed.data, self.ebeam)
+        sems._focuser = self.SEM_focuser
```
pieleric, line 619: pass focuser=self.sem_focuser above.

src/odemis/acq/test/acq_test.py
```diff
@@ -486,5 +484,306 @@ def on_progress_update(self, future, start, end):
...
+        # create streams
+        sems = stream.SEMStream("sem", self.sed, self.sed.data, self.ebeam)
+        sems._focuser = self.SEM_focuser
+        self.streams = [sems]
+
+        # mock the zlevels dictionary
+        zlevels = {}
+        for s in self.streams:
+            zlevels[s] = [s.focuser.position.value["z"]]
+
```
pieleric, line 587: Please test `estimateZStackAcquisitionTime()` too, in the different test cases. You can just check that it returns a value above an arbitrary value (eg, > 2s).

src/odemis/acq/test/acq_test.py
```diff
@@ -486,5 +484,306 @@ def on_progress_update(self, future, start, end):
...
+            # since no zstack, the center has 2 components
+            self.assertEqual(len(d.metadata[model.MD_POS]), 2)
+            # since no zstack, the pixel size has 2 components
+            self.assertEqual(len(d.metadata[model.MD_PIXEL_SIZE]), 2)
+
+        # 2 streams, so 2 acquisitions
+        self.assertEqual(len(data), 2)
+
+        # 2 streams, 2 updates per stream, so 2 updates
+        self.assertGreaterEqual(self._nb_updates, 1)
```
pieleric, line 647: According to your comment, you should check it's >= 2, right?

src/odemis/gui/cont/acquisition.py
pieleric: Please, drop these two lines. There is now a check that "focus" exists.

src/odemis/util/comp.py
```diff
@@ -107,4 +109,57 @@ def get_fov_rect(comp, fov):
...
+        zrange[1] = focuser_rng[1] - focuser_pos
+    if focuser_pos + zrange[0] < focuser_rng[0]:
+        zrange[0] = focuser_rng[0] - focuser_pos
+    if zrange[0] == zrange[1]:
+        return focuser_pos + zrange[0]
+
+    # find number of samples 
+    n = (zrange[1] - zrange[0]) / abs(zstep) + 1
+    if n > MAX_ZLEVELS:
+        raise IndexError("The number of zlevels is too large. Reduce the zstep value.")
```
pieleric, line 152: Please indicate in the error message: the number of zlevels, and the maximum number allowed. In case a user triggers this error, it'll be much easier to find out why, and how to fix it.

src/odemis/util/comp.py
```diff
@@ -107,4 +109,57 @@ def get_fov_rect(comp, fov):
...
+    # try the floor and ceil values for the number of samples, and 
+    # take the one that give smaller error
+    step_f = (zrange[1] - zrange[0]) / (numpy.floor(n) - 1)
+    step_c = (zrange[1] - zrange[0]) / (numpy.ceil(n) - 1)
```
pieleric, line 156: In general, avoid using numpy functions on non numpy object. Here, you can use the standard `math.floor()` and `math.ceil()`. (Same a few lines below).

src/odemis/util/driver.py
```diff
@@ -143,6 +143,28 @@ def estimateMoveDuration(distance, speed, accel):
...
+def guessActuatorMoveDuration(actuator, axis, distance, accel=DEFAULT_ACCELERATION):
+    """
+    actuator (Actuator): actuator object 
+    axis (str): indicates along which axis the movement is. 
+    distance (float): the move for which the time to be estimated  
+    return (float >= 0): the estimated time (in s)
+    """
+    speed = None
+    if not (hasattr(actuator, "axes") and isinstance(actuator.axes, dict)):
+        raise ValueError("The component %s is expected to have a dictionary of axes, but it does not." % actuator)
```
pieleric, line 158: You can just write "The component %s should be an actuator, but it is not".

pieleric (changes requested):

PR #1776, opened 2021-05-25
driver xt_client: add immersion mode support

pieleric:
With XTToolKit, it's possible to change the immersion mode using the
compound_lens_focusing_mode.

### DIFF
--- a/src/odemis/driver/test/xt_client_test.py
+++ b/src/odemis/driver/test/xt_client_test.py
@@ -1121,8 +1121,9 @@ class TestMBScanner(unittest.TestCase):
         self.assertEqual(current_value, self.scanner.beamletIndex.value)
 
     def test_immersion_VA(self):
-        current_value = self.scanner.immersion.value
+        current_value = self.scanner.immersion.value  # type: bool
 
+        # Set current_value as last, so that it goes back to current value after the test
         for new_val in (not current_value, current_value):
             # Test if directly changing it via the VA works
             self.scanner.immersion.value = new_val
--- a/src/odemis/driver/test/xt_client_test.py
+++ b/src/odemis/driver/test/xt_client_test.py
@@ -1120,6 +1120,18 @@ class TestMBScanner(unittest.TestCase):
         self.assertEqual(current_value, self.microscope.get_beamlet_index())
         self.assertEqual(current_value, self.scanner.beamletIndex.value)
 
+    def test_immersion_VA(self):
+        current_value = self.scanner.immersion.value
+
+        for new_val in (not current_value, current_value):
+            # Test if directly changing it via the VA works
+            self.scanner.immersion.value = new_val
+            self.assertEqual(self.scanner.immersion.value, new_val)
+
+            # Check it's still correct after updating the settings
+            time.sleep(6)
+            self.assertEqual(self.scanner.immersion.value, new_val)
+
     def test_multiprobe_mode_VA(self):
         current_beam_mode = self.scanner.multiBeamMode.value
         current_aperture_index = self.scanner.apertureIndex.value
--- a/src/odemis/driver/xt_client.py
+++ b/src/odemis/driver/xt_client.py
@@ -53,6 +53,9 @@ DETECTOR2CHANNELNAME = {
     "se-detector": "electron1",
 }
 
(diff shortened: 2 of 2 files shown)


src/odemis/driver/test/xt_client_test.py
```diff
@@ -1120,6 +1120,18 @@ def test_beamlet_index_VA(self):
         self.assertEqual(current_value, self.microscope.get_beamlet_index())
         self.assertEqual(current_value, self.scanner.beamletIndex.value)
 
+    def test_immersion_VA(self):
+        current_value = self.scanner.immersion.value
+
+        for new_val in (not current_value, current_value):
```
Kleijwegt, line 1126: Nice implementation to set back the original value after trying the other than original value.

However, after the presentation of Thera (and via `import this`) in the Zen of python i found rule 2 stating: _Explicit is better than implicit._ I don't know if setting back the original value is done very explicitly here.
On the other had rule 1 states: _Beautiful is better than ugly._ And since beauty is a matter of taste....

Anyhow if @tepals doesn't mind this implementation is ok

pieleric: I can add a comment to explicit state the trick.

src/odemis/driver/xt_client.py
```diff
@@ -982,6 +982,34 @@ def beamlet_index_info(self):
...
+        """
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            self.server.set_compound_lens_focusing_mode(mode)
+
+    def compound_lens_focusing_mode_info(self):
+        """
+        Get the min/max values of the compound lens focus mode.
+
+        :return (dict str -> list (of length 2)): The range of the focusing mode for the keyword "range".
```
tepals, line 1007: Even though the unit is None, it also returns the unit.

src/odemis/driver/xt_client.py
```diff
@@ -2228,6 +2277,10 @@ def _setBeamletIndex(self, beamlet_index):
         new_beamlet_index = self.parent.get_beamlet_index()
         return tuple(int(i) for i in new_beamlet_index)  # convert tuple values to integers.
 
+    def _setImmersion(self, immersion):
+        self.parent.set_compound_lens_focusing_mode(COMPOUND_LENS_FOCUS_IMMERSION if immersion else 0)
```
tepals, line 2281: Can you add a small comment to explain this `COMPOUND_LENS_FOCUS_IMMERSION if immersion else 0`?

src/odemis/driver/xt_client.py
```diff
@@ -2065,6 +2093,10 @@ def stop(self, axes=None):
             logging.exception("Unexpected failure when updating position")
 
 
+# Value to use on the FASTEM to activate the immersion mode
+COMPOUND_LENS_FOCUS_IMMERSION = 2.0
```
tepals, line 2097: I would usually put constants at the top of the module, why put it here?

src/odemis/driver/xt_client.py
```diff
@@ -2140,6 +2172,19 @@ def __init__(self, name, role, parent, hfw_nomag, **kwargs):
             setter=self._setBeamletIndex
         )
 
+        # The compound lens focusing mode allows to switch between immersion mode or not.
+        # So instead of passing a float, we just provide a boolean, with the immersion
+        # mode always set to the same (hard-coded) value. When reading, anything above
+        # 0 is considered in immersion.
```
tepals, line 2178: From this comment it's not clear to me what compound lens focusing values 1.0, and 3.0-9.0 are. Can you clarify that? 

In addition to this, does the following comment mean that the compound lens focusing mode (in the adapter) allows some sort of yes no switch, or does that actually allow a certain level of compound lens focusing?
`# The compound lens focusing mode allows to switch between immersion mode or not.`

src/odemis/driver/xt_client.py
pieleric: Sometimes when a constant is only used by one class, I place it just on top of that class. I don't have a strong preference though, so I'll move it higher.

tepals (approved):

Kleijwegt (approved):

PR #1778, opened 2021-05-28
odemisd start: do not close the GUI window if it's already opened

pieleric:
Calling Odemis-start when Odemis is already started doesn’t restart the
backend. However, it restart the GUI.

That’s bad behaviour, as some customers have multiple icons of odemis.
In such case, it’s easy to call odemis-start by mistake, which kills the
GUI, and brings a new one. Really not fun if the GUI was in the process
of a long acquisition.

=> If GUI is already started, do not restart it.


philipwinklerdelmic (approved):
Looks good to me!

PR #1779, opened 2021-05-28
gui cont tabs SPARC2Align: also update the optical path when changing the detector

pieleric:
The new manual focus option allow to switch the detector. For now it
just select which stream plays. That's not enough as each detector is
selected by a mirror to switch the optical path.

=> Call the optical path manager to update move the mirror(s) to the
right position.


lazem (approved):
Sounds good.

PR #1780, opened 2021-06-01
doc: add information about the Optical Path Manager

pieleric:


doc/develop/acq.rst
```diff
@@ -372,3 +372,70 @@ SPARC alignment:
...
+
+.. py:method:: setPath(mode, detector=None)
+
+   Sets all the necessary components of the optical path to the positions
+   required to achieve the requested mode.
+
+   :param path: The stream or the optical path mode
+   :type path: stream.Stream or str
+   :param detector:
+      When the `path` is a str, this allows
```
Sabrina85, line 394: I think the rule is:
allow + object + infinitive
or:
allow + gerund
So here `this allows defining`...

doc/develop/acq.rst
```diff
@@ -372,3 +372,70 @@ SPARC alignment:
...
+Currently, in Odemis, the Optical Path Manager is called from two places:
+
+ * In the ``Stream.prepare()``, which is itself called just before a stream acquires
+   data.
+ * In the SPARC alignment tab, whenever the alignment mode is changed.
+
+CCD fan control
+---------------
+As a extra function, the Optical Path Manager also takes care of disabling the
+fan of the CCDs during an SEM acquisition, the SECOM and DELPHI platforms.
```
Sabrina85, line 426: This sentence seems incomplete.

doc/develop/acq.rst
```diff
@@ -372,3 +372,70 @@ SPARC alignment:
...
+This ensures that there are no vibration during high quality SEM image acquisition.
```
Sabrina85, line 427: Either use plural acquisitions or an undefined article:
during **a** high quality SEM image acquisition
or
acquisition**s**

doc/develop/acq.rst
```diff
@@ -372,3 +372,70 @@ SPARC alignment:
...
+To allow this feature, the manager has this extra method:
+
+.. py:method:: setAcqQuality(quality)
+
+   Updates the acquisition quality expected during the following acquisitions.
```
Sabrina85, line 432: Is the quality expected or wanted? So can a user select it?

doc/develop/acq.rst
```diff
@@ -372,3 +372,70 @@ SPARC alignment:
...
+   :rtype: Future
+   :raise: ValueError if the given mode does not exist
+           IOError if a detector is missing
+
+The list of supported `paths` (ie, optical modes) can be found in the ``acq.path`` module.
+At runtime, they are limited by the microscope type and the available components.
+
+In terms of implementation, for each optical mode, there are two types of actuators:
+
+ * The path selectors which, typically, are mirrors with 2 positions. Their positions
```
Sabrina85, line 410: I think this "affects" property and the way the optical path is calculated could need some more information. If you have heard about it before, it is fine, but if you never have heard about it, it will be difficult to understand how the optical path is derived. You explained this nicely yesterday in your talk. 
Maybe it is not the right place here, but somewhere in this section you could add information, that the affects needs to be defined in the microscope file, and that you start from the detector towards the actuators (if I understood this correctly).

doc/develop/acq.rst
```diff
@@ -372,3 +372,70 @@ SPARC alignment:
...
+
+
+It has one important method:
+
+.. py:method:: setPath(mode, detector=None)
+
+   Sets all the necessary components of the optical path to the positions
+   required to achieve the requested mode.
+
+   :param path: The stream or the optical path mode
```
Sabrina85, line 391: Where would I find those modes if I have never heard about it before? Can you add this info somewhere?

Sabrina85 (changes requested):
Some small comments. The rest looks good!

doc/develop/acq.rst
```diff
@@ -372,3 +372,70 @@ SPARC alignment:
...
+   :raise: ValueError if the given mode does not exist
+           IOError if a detector is missing
+
+The list of supported `paths` (ie, optical modes) can be found in the ``acq.path`` module.
+At runtime, they are limited by the microscope type and the available components.
+
+In terms of implementation, for each optical mode, there are two types of actuators:
+
+ * The path selectors which, typically, are mirrors with 2 positions. Their positions
+   is derived from the "affects" property of each component, by computing a graph
```
tepals, line 411: Positions is multiple, so "are derived", or "the positions property is derived".

doc/develop/acq.rst
```diff
@@ -372,3 +372,70 @@ SPARC alignment:
...
+   joining the actuators to the detectors.
+ * The acquisition type actuators which configure how the light can be seen. For
+   instance in AR mode, the lens 2 needs to be activated, while it should be removed
+   during spectrum mode.
+
+Currently, in Odemis, the Optical Path Manager is called from two places:
+
+ * In the ``Stream.prepare()``, which is itself called just before a stream acquires
```
tepals, line 419: I don't understand this sentence, especially "which is itself".

doc/develop/acq.rst
```diff
@@ -372,3 +372,70 @@ SPARC alignment:
...
+   data.
+ * In the SPARC alignment tab, whenever the alignment mode is changed.
+
+CCD fan control
+---------------
+As a extra function, the Optical Path Manager also takes care of disabling the
```
tepals, line 425: "an extra"

doc/develop/acq.rst
tepals: Maybe it's supposed to be **"on** the SECOM and DELPHI platforms."?

doc/develop/acq.rst
tepals: Also: "there is no vibration" or "there are no vibrations".

tepals (commented):
A few extra comments, in addition to Sabrina's review.

doc/develop/acq.rst
pieleric: I'm not sure what is precisely your question. Do you mean "expect" as "think it'll happen"? I meant "expect" as "want it to be done this way". I can change to "required", if that makes it clearer?

doc/develop/acq.rst
pieleric: I've extended the line a little below: "The list of supported `paths` (ie, optical modes) can be found in the ``acq.path`` module in the ``SPARC_MODES`` and ``SPARC2_MODES`` dictionaries."

doc/develop/acq.rst
pieleric: I mean that "Stream.prepare()" calls the optical path manager. Then, the reader could immediately ask "when is stream.prepare() called?". It's called just before a stream acquisition starts.
Do you understand what I mean? How would you express it better?
How about:
"In ``Stream.prepare()``. This function is in charge of preparing all the
   microscope components prior to the stream acquisition.
   In Odemis, it is called explicitly in some cases. It is also called implicitly,
   every time a stream is about to start acquiring data."
Is that clearer?

doc/develop/acq.rst
Sabrina85: Yes "think it''ll happen just by some magic" is kind of what I would associate with it as it is now. Indeed "required" or maybe even "requested"? As it is still a user who decides on the quality.

Sabrina85 (approved):
Looks good to me :)

doc/develop/acq.rst
tepals: That's more clear.

tepals (approved):

doc/develop/acq.rst
pieleric: I'll leave it to "required" :-)

PR #1781, opened 2021-06-01
driver semcomedi: fix non-square pixel size on init

pieleric:
When the max_res wasn't square, the pixel size would be initialised
incorrectly to non-square value. On the first magnification update,
it would be set properly. As in almost every case the user has to change
the magnification, this is rarely noticeable. Still, it's good that
the pixelSize is always correct.

### DIFF
--- a/src/odemis/driver/test/semcomedi_test.py
+++ b/src/odemis/driver/test/semcomedi_test.py
@@ -55,8 +55,12 @@ TEST_NOHW = (os.environ.get("TEST_NOHW", 0) != 0)  # Default to Hw testing
 CONFIG_SED = {"name": "sed", "role": "sed", "channel":5, "limits": [-3, 3]}
 CONFIG_BSD = {"name": "bsd", "role": "bsd", "channel":6, "limits": [0.2, -0.1]}
 CONFIG_CNT = {"name": "cnt", "role": "cnt", "source":0}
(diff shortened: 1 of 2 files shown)


src/odemis/driver/test/semcomedi_test.py
```diff
@@ -205,6 +209,21 @@ def test_ttl(self):
...
+        mag_orig = self.scanner.magnification.value
+
+        self.assertEqual(pxs_orig[0], pxs_orig[1])
+
+        # Mag x 2 => pixel size / 2
+        self.scanner.magnification.value *= 2
+        self.assertAlmostEqual(mag_orig * 2, self.scanner.magnification.value)
+        new_pxs = self.scanner.pixelSize.value
+        self.assertAlmostEqual(pxs_orig[1] / 2, new_pxs[0])
+        self.assertEqual(pxs_orig[0], pxs_orig[1])
```
VictoriaDel, line 223: Why don´t you do self.assertEqual(new_pxs[0], new_pxs[1]) ?

VictoriaDel (approved):

src/odemis/driver/test/semcomedi_test.py
pieleric: that's a copy-paste mistake. Now fixed.

PR #1782, opened 2021-06-02
FastEM GUI: scintillator arrangement, bugfix coordinates

philipwinklerdelmic:
* The arrangement of the scintillators on the sample carrier differs from the one which was previously hardcoded
* There was some confusion related to the ltrb coordinates ("top" means minx (not maxx) in physical coordinates)

I decided to set the minx, miny origin at the bottom left of the entire sample carrier (not the first scintillator) to avoid negative coordinates for the background overlay.

The distance between scintillators has been confirmed to be 18 mm, not 18.5.

### DIFF
--- a/src/odemis/gui/model/__init__.py
+++ b/src/odemis/gui/model/__init__.py
@@ -1050,26 +1050,25 @@ class FastEMAcquisitionGUIData(MicroscopyGUIData):
     """
     # Parameters of the scintillators according to the technical drawing of the sample carrier
     # (positions are given relative to top left of sample carrier)
-    # Scintillator arrangement on the sample carrier (9 is top left, 1 is bottom right):
-    # 9 8 7
-    # 6 5 4
-    # 3 2 1
+    # Scintillator arrangement on the sample carrier:
+    # 3 6 9
+    # 2 5 8
+    # 1 4 7
     # The format looks like YAML, because eventually it should be possible to pass it in the microscope (YAML) file.
     SAMPLE_CARRIER_3x3 = {
         "dims": (120e-3, 120e-3),  # m
-        "layout": [[9, 8, 7],
-                   [6, 5, 4],
-                   [3, 2, 1]],  # grid positions of scintillators
+        "layout": [[3, 6, 9],
+                   [2, 5, 8],
+                   [1, 4, 7]],  # grid positions of scintillators
         "scintillator_offsets": {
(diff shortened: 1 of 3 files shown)


src/odemis/gui/model/__init__.py
```diff
@@ -1051,25 +1051,24 @@ class FastEMAcquisitionGUIData(MicroscopyGUIData):
     # Parameters of the scintillators according to the technical drawing of the sample carrier
     # (positions are given relative to top left of sample carrier)
     # Scintillator arrangement on the sample carrier (9 is top left, 1 is bottom right):
-    # 9 8 7
-    # 6 5 4
-    # 3 2 1
+    # 3 6 9
```
Kleijwegt, line 1054: These lines seem to contradict. 9 is top right now isn't it? and 1 is bottom left. Or is this a matter of perspective? (like looking from behind my screen, than maybe that should also be commented.)

src/odemis/gui/model/__init__.py
```diff
@@ -1084,7 +1083,7 @@ def __init__(self, main):
         if model.MD_POS_ACTIVE_RANGE not in md:
             raise KeyError("Stage has no MD_POS_ACTIVE_RANGE metadata.")
         carrier_range = md[model.MD_POS_ACTIVE_RANGE]
-        tlx, tly = float(carrier_range["x"][0]), float(carrier_range["y"][0])  # top left carrier position in m
+        minx, miny = float(carrier_range["x"][0]), float(carrier_range["y"][0])  # top left carrier position in m
```
Kleijwegt, line 1086: Why did you choose to not use an underscore between min_y? (and min_x?)

src/odemis/gui/model/__init__.py
pieleric: Indeed, I think the comment just above should be updated too.

src/odemis/gui/model/__init__.py
philipwinklerdelmic: We use "minx" and "miny" at other places in odemis in the same context, so I wanted to be consistent. My preference would have been for "xmin" and "ymin", but it doesn't really change much.

src/odemis/gui/model/__init__.py
philipwinklerdelmic: Indeed, that is confusing. I deleted it.

PR #1784, opened 2021-06-03
Addition of the !include and !extend tags to merge yamls (including testcases and corresponding test files)

Kleijwegt:


src/odemis/odemisd/test/yaml-merger-relative-path-test/yaml-merger-include-complete-init-relative-path-CL-Detector-test.odm.yaml
```diff
@@ -0,0 +1,4 @@
+{
+    settle_time: !include /home/kleijwegt/development/odemis/src/odemis/odemisd/test/yaml-merger-include-init-val-CL-Detector-test.odm.yaml,
```
tepals, line 2: Did not do a review, however just saw this: /home/kleijwegt/

src/odemis/odemisd/test/yaml-merger-combination-include-extend-expected-result.yaml
```diff
@@ -0,0 +1,80 @@
...
+       detector0: "CL PMT",
+       scanner: "SEM E-beam",
+    }
+}
+
+"CL Detector": {
+    class: pmtctrl.PMT,
+    role: cl-detector,
+    init: {
+        # Add via !inlcude
```
Kleijwegt, line 30: I could add an include call here in the test cases but I don't know if that is needed

src/odemis/odemisd/modelgen.py
```diff
@@ -392,6 +475,11 @@ def _parse_instantiation_model(self, inst_file):
         """
         try:
             # The standard PyYAML loader is dangerous as it can create any python object
+
+            # Allow to include external yaml files via the keyword '!include' by adding a constructor.
+            SafeLoader.add_constructor('!include', SafeLoader.include)
+            # Overwrite construct_yaml_map with custom mapper
+            SafeLoader.yaml_constructors['tag:yaml.org,2002:map'] = SafeLoader.construct_yaml_map
```
pieleric, line 482: I think it's possible to move this just after the SafeLoader class. This way, it's done just once (on importing the module), and the caller code doesn't have to "baby-sit" the class.

src/odemis/odemisd/test/yaml-merger-combination-realistic-startup-file.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+"SEM Scan Interface": !include yaml-merger-combination-SEM-Scan-Interface.yaml
+
+"CL Detector": {
+    class: pmtctrl.PMT,
+    role: cl-detector,
+    init: {
+        # Add via !inlcude
+        settle_time: [0.1, 1, 3],
+        test_input: "a"
```
pieleric, line 19: Shouldn't this file be filled by !include and !extend? It's identical to the -expected version, so it's obviously going to pass...

src/odemis/odemisd/test/modelget_test.py
```diff
@@ -0,0 +1,148 @@
+import unittest
```
pieleric, line 1: That file should probably be called "modelgen_test.py" (not modelge**t**_test).

src/odemis/odemisd/test/modelget_test.py
```diff
@@ -0,0 +1,148 @@
...
+import yaml
```
pieleric, line 2: Please add the standard copyright header to the this file.

src/odemis/odemisd/test/modelget_test.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+class SafeLoader_extensions_test(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        SafeLoader.add_constructor('!include', SafeLoader.include)
+        # Overwrite construct_yaml_map with custom mapper
+        SafeLoader.yaml_constructors['tag:yaml.org,2002:map'] = SafeLoader.construct_yaml_map
+        cls.SafeLoader = SafeLoader
```
pieleric, line 14: It should be possible to drop all this code once SafeLoader is updated at import.

pieleric (changes requested):

src/odemis/odemisd/test/yaml-merger-combination-realistic-startup-file.yaml
Kleijwegt: That is what is meant with [this comment](https://github.com/delmic/odemis/pull/1784/files#r645353416) (though it was on the wrong place)
> I could add an include call here in the test cases but I don't know if that is needed

I think that would be a bit over the top because the SEM Scan Interface already tests multiple nested calls. Or do you think that will add a lot of value?

src/odemis/odemisd/modelgen.py
Kleijwegt: So you would implement it more or less like:

```
class SafeLoader(yaml.SafeLoader):
  def __init__(self, stream):
  .....
  def other_methods()
  ......
  def last_methods_class()
  ......

# Add keys to the Safeloader class
# Allow to include external yaml files via the key '!include' by adding a constructor.
SafeLoader.add_constructor('!include', SafeLoader.include)
# Overwrite construct_yaml_map with custom mapper which adds support for the "!extend" key
SafeLoader.yaml_constructors['tag:yaml.org,2002:map'] = SafeLoader.construct_yaml_map
```

pieleric: Yes, that's how it's also done in the YAML code.

src/odemis/odemisd/test/yaml-merger-combination-realistic-startup-file.yaml
pieleric: Sorry, I hadn't noticed there is one `!include` already. I think this "realistic" file should show more of the usage of `!include` and `!extend`. Anyway, there shouldn't be any comment like "Add via !include" or "add via !extend", and actually it's just normal text. => really do it.

src/odemis/odemisd/modelgen.py
```diff
@@ -898,7 +986,7 @@ def get_instantiables(self, instantiated=None):
                 comps.add(n)
 
         return comps
-            
+
```
Kleijwegt, line 989: Same error as in https://github.com/delmic/odemis/pull/1803#discussion_r654189045.
I need to fix this, but I have already updated the PR because this seems like a minor change.

src/odemis/odemisd/modelgen.py
```diff
@@ -922,7 +1010,7 @@ def read_yaml(self, f):
             logging.warning("Settings file reset after loading failed with Exception %s." % ex)
             data = {}
         return data
-        
+
```
Kleijwegt, line 1013: Same error as in https://github.com/delmic/odemis/pull/1803#discussion_r654189045.
I need to fix this, but I have already updated the PR because this seems like a minor change.

src/odemis/odemisd/test/modelgen_test.py
```diff
@@ -0,0 +1,165 @@
...
+You should have received a copy of the GNU General Public License along with Delmic Acquisition Software. If not, see
+http://www.gnu.org/licenses/.
+"""
+import unittest
+import yaml
+
+from odemis.odemisd.modelgen import SafeLoader, ParseError
+
+
+class SafeLoader_extensions_test(unittest.TestCase):
```
tepals, line 29: This mix of cases looks strange. I would go for `SafeLoaderExtensionsTest`, this is also the convention in the rest of odemis.

src/odemis/odemisd/test/modelgen_test.py
```diff
@@ -0,0 +1,165 @@
...
+
+        # Full CL-detector component
+        with open("yaml-merger-include-full-CL-Detector-component-test.odm.yaml", "r") as f:
+            data_found = yaml.load(f, SafeLoader)
+
+        # Compare with expected results
+        expected_result = {"CL Detector": self.expected_full_result["CL Detector"]}
+        self.assertEqual(expected_result, data_found)
+
+        # __init__ contains a setting which is overwritten of the CL-detector
```
tepals, line 56: I think you mean "overwritten by the CL-detector"

src/odemis/odemisd/test/modelgen_test.py
```diff
@@ -0,0 +1,165 @@
...
+                "r") as f:
+            data_found = yaml.load(f, SafeLoader)
+
+        # Compare with expected results
+        expected_result = self.expected_full_result["CL Detector"]["init"]
+        self.assertEqual(expected_result, data_found)
+
+        # __init__ of the CL-detector using an absolute reference in the !include
+        with open(
+            "yaml-merger-relative-path-test/yaml-merger-include-complete-init-relative-path-CL-Detector-test.odm.yaml",
```
tepals, line 77: What's the difference between this line and line 67?

tepals: You could consider adding a `path = "yaml-mer......" ` variable to make this a bit more readable.

src/odemis/odemisd/test/modelgen_test.py
```diff
@@ -0,0 +1,165 @@
...
+
+        # Full SEM Scan Interface component
+        with open("yaml-merger-extend-full-SEM-Scan-Interface-component-test.odm.yaml", "r") as f:
+            data_found = yaml.load(f, SafeLoader)
+
+        # Compare with expected results
+        expected_result = {"SEM Scan Interface": self.expected_full_result["SEM Scan Interface"]}
+        self.assertEqual(expected_result, data_found)
+
+        # __init__ contains a entries which are overwritten of the CL-detector
```
tepals, line 113: typo: "contains an entry" or "contains entries"
And some comment as above, I think you mean "overwritten by"

tepals (commented):
A few comments. I did not pull the code and test it myself, let me know if you want me to do that!

pieleric (changes requested):
Looks pretty fine to me. If you fix my single comment and the few comments from Thera, it should be fine to merge.

Kleijwegt:
@pieleric I think it is ready to be merged, maybe we should then also present/explain this to the Ralou/Iheb so it will actually be used.

PR #1785, opened 2021-06-03
ENZEL: Handle Smaract invalid unreachable position error

lazem:
Due to the physical constraints on the 5dof stage design, sometimes a move on one of its axes fails because the stage is close to the maximum range of its positions. This typically happens if a move or reference is cancelled (resulting in unknown position) and the stage is trying to retract back to its loading position. To mitigate this issue, an ordered fallback_submoves list is added, moving linear axes at the same time first, followed by rotational axes in case the typical submoves failed with this particular smaract exception.

Also the 5dof FakeMC driver is modified to simulate this behaviour.

### DIFF
--- a/src/odemis/acq/move.py
+++ b/src/odemis/acq/move.py
@@ -213,12 +213,16 @@ def _doCryoSwitchSamplePosition(future, stage, focus, align, target):
         current_pos = stage.position.value
         # To hold the ordered sub moves list
         sub_moves = []
+        # To hold the sub moves to run if the ordered sub moves failed
+        fallback_submoves = []
         # Create axis->pos dict from target position given smaller number of axes
         filter_dict = lambda keys, d: {key: d[key] for key in keys}
 
-        # Initial submove for all procedures is to park focus if it's not already parked
-        if not _isNearPosition(focus.position.value, focus_deactive, {'z'}):
+        # Initial submoves for all procedures is to park focus & aligner to safe position
+        if not _isNearPosition(focus.position.value, focus_deactive, focus.axes):
             sub_moves.append((focus, focus_deactive))
+        if not _isNearPosition(align.position.value, align_deactive, align.axes):
+            run_sub_move(future, align, align_deactive)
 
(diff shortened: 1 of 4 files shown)


src/odemis/acq/move.py
```diff
@@ -285,8 +294,18 @@ def _doCryoSwitchSamplePosition(future, stage, focus, align, target):
...
-        for component, sub_move in sub_moves:
-            run_sub_move(future, component, sub_move)
+        try:
+            for component, sub_move in sub_moves:
+                run_sub_move(future, component, sub_move)
+        except SA_MCError as sa_mcerror:
+            # In case the required movement is invalid/unreachable with the smaract 5dof stage
+            # Move all linear axes first then rotational ones using the fallback_submoves
+            if sa_mcerror.errno != MC_5DOF_DLL.SA_MC_ERROR_POSE_UNREACHABLE:
+                raise
```
pieleric, line 304: Let's really avoid having the function depend on some specific driver. I'm mostly not so happy with the imports that it causes. I'd suggest one of these ways:
* Just detect an Exception, and always retry with the fallback moves. In case this doesn't help, it'd probably fail again the same way.
* Let the 5DoF driver catch this specific error and raise a IndexError. Here, catch the IndexError.
* Create a new type of exception like "UnreachablePositionError", which you'd add to model/_components.py. Then you can let the 5DoF driver catch the specific error and raise UnreachablePositionError.

I'd recommend the 2nd option, but it's up to you.

pieleric (changes requested):

pieleric (approved):
It looks good.
Could you squash the commits so that we have two logical commits:
* Changes to driver/smaract (add support for unreachable position error)
* Changes to acq/move (detect move error and fallback to more aggressive move)

lazem:
> It looks good.
> Could you squash the commits so that we have two logical commits:
> 
> * Changes to driver/smaract (add support for unreachable position error)
> * Changes to acq/move (detect move error and fallback to more aggressive move)

Hi @pieleric Done as requested. 

Thanks for the review.

PR #1786, opened 2021-06-04
driver xt_client: change .resolution and .scale to VAEnumerated

pieleric:


tepals (approved):

philipwinklerdelmic:
We should also update the bitbucket xtlib-adapter repository to use one of the valid resolutions in the simulator, otherwise, the driver will complain that the simulator resolution is not a valid choice.

pieleric:
Yes, I've made a PR in parallel onto xtlib-adapter to extend the simulator.

Kleijwegt:
@tepals this PR corrupted the test cases with the server_sim/adapter_sim simulator. It had stated as default resolution (512, 756) while this isn't one of the allowed choices of the VA.
Do you think we can just edit this in the simulator or not?

philipwinklerdelmic:
I think you can just update the simulator testcases with one of the allowed values.

tepals:
It's fine to edit this value in the adapter_sim simulator

Kleijwegt:
My bad, I wasn't running the latest version of the simulator. It was already updated in the newer one.

PR #1787, opened 2021-06-04
driver xt_client test: extra option to run with the adapter simulator

pieleric:
Use the same was as the orsay test case: add an extra state to
TEST_NOHW=sim. This indicate we are testing agains the simulator.

In such case, all test cases are safe to run, so run them all.
Also fix a few test cases to be compatible with the simulator.

### DIFF
--- a/src/odemis/driver/test/xt_client_test.py
+++ b/src/odemis/driver/test/xt_client_test.py
@@ -31,13 +31,25 @@ import unittest
 from odemis import model
 
 from odemis.driver import xt_client
-from odemis.driver.xt_client import DETECTOR2CHANNELNAME, XTTKDetector
+from odemis.driver.xt_client import DETECTOR2CHANNELNAME
 from odemis.model import ProgressiveFuture, NotSettableError
 from odemis.util import test
 
 logging.basicConfig(level=logging.INFO)
 
-TEST_NOHW = (os.environ.get("TEST_NOHW", 0) != 0)
+# Accept three values for TEST_NOHW
+# * TEST_NOHW = 1: not connected to anything => skip most of the tests
+# * TEST_NOHW = sim: xtadapter/server_sim.py running on localhost
+# * TEST_NOHW = 0 (or anything else): connected to the real hardware
+TEST_NOHW = os.environ.get("TEST_NOHW", "0")  # Default to Hw testing
+if TEST_NOHW == "sim":
+    pass
+elif TEST_NOHW == "0":
+    TEST_NOHW = False
+elif TEST_NOHW == "1":
+    TEST_NOHW = True
+else:
+    raise ValueError("Unknown value of environment variable TEST_NOHW=%s", (TEST_NOHW,))
 
 # arguments used for the creation of basic components
 CONFIG_SCANNER = {"name": "scanner", "role": "ebeam", "hfw_nomag": 1}
@@ -62,6 +74,10 @@ CONFIG_MB_SEM = {"name": "sem", "role": "sem", "address": "PYRO:Microscope@192.1
                               }
                  }
 
+if TEST_NOHW == "sim":
+    CONFIG_SEM["address"] = "PYRO:Microscope@localhost:4242"
+    CONFIG_MB_SEM["address"] = "PYRO:Microscope@localhost:4242"
+
 
 class TestMicroscope(unittest.TestCase):
     """
@@ -70,6 +86,8 @@ class TestMicroscope(unittest.TestCase):
 
     @classmethod
     def setUpClass(cls):
+        if TEST_NOHW is True:
+            raise unittest.SkipTest("No hardware available.")
 
         cls.microscope = xt_client.SEM(**CONFIG_SEM)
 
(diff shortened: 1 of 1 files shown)


src/odemis/driver/test/xt_client_test.py
```diff
@@ -609,7 +616,7 @@ def test_stigmator(self):
         self.assertEqual(self.microscope.get_stigmator()[0], stig_x)
         self.assertEqual(self.microscope.get_stigmator()[1], stig_y)
         # Try to set value outside of range
-        with self.assertRaises(OSError):
+        with self.assertRaises((OSError, ValueError)):
```
tepals, line 619: I thought Pyro5 did not transfer ValueErrors, are you sure you need this?

I'm pretty sure the adapter only returns OSErrors. So if this is for the simulator, maybe we should update the simulator to always return OSErrors.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -31,13 +31,19 @@
...
 logging.basicConfig(level=logging.INFO)
 
-TEST_NOHW = (os.environ.get("TEST_NOHW", 0) != 0)
+# Accept three values for TEST_NOHW
+# * TEST_NOHW = 1: not connected to anything => skip most of the tests
+# * TEST_NOHW = sim: xtadapter/server_sim.py running on localhost
+# * TEST_NOHW = 0 (or anything else): connected to the real hardware
+TEST_NOHW = os.environ.get("TEST_NOHW", 0)  # Default to Hw testing
+if TEST_NOHW != "sim":
+    TEST_NOHW = (TEST_NOHW == "1")  # convert values other than "sim" to False/True
```
tepals, line 46: This feels like your hiding things. If someone accidentally has typed "sym", `TEST_NOHW` will be set to False, which implies hardware testing.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -381,11 +389,11 @@ def test_set_scan_field_size(self):
         init_scan_size = self.microscope.get_scanning_size()[0]
         new_scanfield_x = scanfield_range['x'][1]
         self.microscope.set_scanning_size(new_scanfield_x)
-        self.assertEqual(self.microscope.get_scanning_size()[0], new_scanfield_x)
+        self.assertAlmostEqual(self.microscope.get_scanning_size()[0], new_scanfield_x)
         # Test it still works for different values.
         new_scanfield_x = scanfield_range['x'][0]
         self.microscope.set_scanning_size(new_scanfield_x)
-        self.assertEqual(self.microscope.get_scanning_size()[0], new_scanfield_x)
+        self.assertAlmostEqual(self.microscope.get_scanning_size()[0], new_scanfield_x)
```
tepals, line 396: By default assertAlmostEqual rounds to 7 decimal places. The minimum range for the scanfield is 1e-8, so I think you want to test at least for 8 decimal places.

https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertAlmostEqual

src/odemis/driver/test/xt_client_test.py
pieleric: Yes, that's for the simulator. I noticed that the current normal code either raise ValueError when the Python code checks for the bounds or OSError when it's XTLib/XTOM which checks. I concluded that both should be supported in the client, as it's hard to guess. So I've added it here, and decided to use ValueError in the simulator, as we check in the Python code, and didn't feel it's better to simulate a OSError. I can switch the simulator to OSError and here to just OSError if you see a good reason for it.

src/odemis/driver/test/xt_client_test.py
pieleric: So, you mean it'd be better to say "anything different from 0" is NOHW testing? Or it should raise an error when it's neither sim 0 or 1? So far, in the other test cases, we consider anything different from 0 is NOHW testing.

src/odemis/driver/test/xt_client_test.py
pieleric: Good catch. Indeed, with microscopes, small is sometimes still very big!
Our own method almost_equal uses 1e-18, which is a little safer.

src/odemis/driver/test/xt_client_test.py
tepals: Let's leave the ValueError then.

src/odemis/driver/test/xt_client_test.py
tepals: Maybe the double negations are messing with my mind, but the way I read this is that anything different from 0 is HW testing.
```
TEST_NOHW = "something-different"
TEST_NOHW = (TEST_NOHW == "1")  = ("something-different" == "1") = False
```

TEST_NOHW = False => TEST_HW = True

Anyway, I would say it should raise an error for anything different than sim, 0 or 1.

pieleric: Yes. Right now it says "anything different from 1 is normal hardware testing". That's opposite to every other test cases. I'll change it to raise an error if it's different from the 3 values.

tepals (approved):

PR #1789, opened 2021-06-04
cli: allow to pass the "action" without "--"

pieleric:
Make the command line easier to type and read by ommitting the "--"
before an action, à la git.
The "--" are still allowed.

So now, instead of writting "odemis-cli --list", one can write
"odemis-cli list".

This only works for "actions", not the options, like "--machine".

### DIFF
--- a/src/odemis/cli/main.py
+++ b/src/odemis/cli/main.py
@@ -53,6 +53,11 @@ status_to_xtcode = {BACKEND_RUNNING: 0,
 VAS_COMPS = {"alive", "dependencies"}
 VAS_HIDDEN = {"children", "affects"}
 
+# Command line arguments which can have "--" omitted
+ACTION_NAMES = ("kill", "check", "list", "list-prop", "set-attr", "update-metadata",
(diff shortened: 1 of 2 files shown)


src/odemis/cli/main.py
```diff
@@ -53,6 +53,10 @@
 VAS_COMPS = {"alive", "dependencies"}
 VAS_HIDDEN = {"children", "affects"}
 
+# Command line arguments which can have "--" omitted
+ACTION_NAMES = ("kill", "check", "list", "list-prop", "set-attr", "update-metadata",
+                "move", "position", "stop", "reference", "acquire", "live", "scan")
```
lazem, line 58: I know it's an optional argument, but what would be the side effect of including `help` as well?

lazem (approved):
Nice update! I have a question about the `help` argument, but otherwise it looks good.

src/odemis/cli/main.py
pieleric: That's an oversight. "git help" works, so "odemis-cli help" would be logical that it works too. I've also added "version".

PR #1790, opened 2021-06-07
driver-xt_client: wrapper for flash functions

philipwinklerdelmic:
I'd like to make a call to the flash functions from the GUI, however, it's not very nice to use the blocking functions of the SEM directly. Can we extend the driver with these wrapper functions?

### DIFF
--- a/src/odemis/driver/test/xt_client_test.py
+++ b/src/odemis/driver/test/xt_client_test.py
@@ -678,54 +678,6 @@ class TestMicroscopeInternal(unittest.TestCase):
         self.assertEqual(use_case, "SingleBeamlet")
         self.microscope.set_use_case(init_use_case)
 
-    @unittest.skip("Currently the auto stigmator functionality is not supported by the XT client and the code is "
-                   "commented out")
-    # TODO Update test_apply_autostigmator code and make test work on XT client when auto stigmator functionality
-    #  works again.
-    def test_apply_autostigmator(self):
-        """
-        Test for the auto stigmation functionality.
-        """
-        detector = 'se-detector'  # autostigmator only works for the se-detector with channel=electron1
-        channel = DETECTOR2CHANNELNAME[detector]
-
-        # Start auto stigmation and check if it is running.
-        autostigmator_future = self.scanner.applyAutoStigmator(detector)
-        time.sleep(2.5)  # Give microscope/simulator the time to update the state
-        autostigmator_state = self.microscope.is_autostigmating(channel)
-        self.assertEqual(autostigmator_state, True)
-        self.assertIsInstance(autostigmator_future, ProgressiveFuture)
-
-        # Stop auto stigmation and check if it stopped running.
-        autostigmator_future.cancel()
-        time.sleep(5.0)  # Give microscope/simulator the time to update the state
-        autostigmator_state = self.microscope.is_autostigmating(channel)
-        self.assertEqual(autostigmator_state, False)
-
-        max_execution_time = 30  # Approximately 3 times the normal expected execution time (s)
-        for i in range(0, 5):
-            starting_time = time.time()
-            autostigmator_future = self.scanner.applyAutoStigmator(detector)
-            time.sleep(0.5)  # Give microscope/simulator the time to update the state
-            autostigmator_state = self.microscope.is_autostigmating(channel)
-            self.assertEqual(autostigmator_state, True)
-            self.assertIsInstance(autostigmator_future, ProgressiveFuture)
-
-            autostigmator_future.result(timeout=max_execution_time)
-
-            # Check if the time to perform the auto stigmation is not too long
-            self.assertLess(time.time() - starting_time, 30,
-                            "Execution of auto stigmation was stopped because it took more than %s seconds." %
-                            max_execution_time)
-            # Line to inspect the execution time to update the expected time
-            print("Execution time was %s seconds " % (time.time() - starting_time))
-
-        # Test if an error is raised when an invalid detector name is provided
-        with self.assertRaises(KeyError):
-            self.scanner.applyAutoStigmator("error_expected")
-        time.sleep(2.5)  # Give microscope/simulator the time to update the state
-        autostigmator_state = self.microscope.is_autostigmating(channel)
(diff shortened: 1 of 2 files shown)


src/odemis/driver/xt_client.py
```diff
@@ -2202,6 +2206,17 @@ def __init__(self, name, role, parent, hfw_nomag, **kwargs):
         # Instantiate the super scanner class with the update thread
         super(MultiBeamScanner, self).__init__(name, role, parent, hfw_nomag, **kwargs)
 
+    @isasync
+    def applyAutoStigmator(self, detector):
```
tepals, line 2210: You don't use the detector, so why pass it as an argument?

src/odemis/driver/test/xt_client_test.py
```diff
@@ -1170,6 +1122,50 @@ def test_beam_powerVA(self):
...
+        """
+        Test for the auto stigmation functionality.
+        """
+        detector = 'se-detector'  # autostigmator only works for the se-detector with channel=electron1
+        channel = DETECTOR2CHANNELNAME[detector]
+
+        # Run autostigmator flash script
+        autostigmator_future = self.scanner.applyAutoStigmator(detector)
+        # Don't check .is_autostigmating, function uses the same lock as .applyAutoStigmator, so it's only
+        # ever called after the autoStigmator function finished. TODO: is that intended?
```
tepals, line 1135: It is known that all functions use the same lock, so that if you have a blocking call you cannot call anything else in parallel. It is hard to work around this. If you are interested: in `XTTKDetector.stop_generate()` there is a workaround, to be able to stop an image acquisition. 

Also `is_autostigmating` does not work for the flash functions. I think we should just remove `is_autostigmating` and `set_autostigmator`, since the `set_autostigmator` is not good enough in autostigamting.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -1170,6 +1122,50 @@ def test_beam_powerVA(self):
...
+
+        # TODO: Commented out, because cancelling is not an option for now.
+        # # Stop auto stigmation and check if it stopped running.
+        # autostigmator_future.cancel()
+        # time.sleep(5.0)  # Give microscope/simulator the time to update the state
+        # autostigmator_state = self.microscope.is_autostigmating(channel)
+        # self.assertEqual(autostigmator_state, False)
+
+        max_execution_time = 30  # Approximately 3 times the normal expected execution time (s)
+        for i in range(0, 5):
```
tepals, line 1149: Why repeat this 5 times? 

Also: you can just write `for i in range(5):`

src/odemis/driver/xt_client.py
```diff
@@ -2392,3 +2407,21 @@ def _acquire(self):
...
+
+
+class XTTKFocus(Focus):
+    """
+    This is an extension of xt_client.Focus class. It overwrites the autofocus function to use the one
+    provided by the flash script.
+    """
+
+    @isasync
+    def applyAutofocus(self):
```
tepals, line 2419: `xt_client.Focus.applyAutofocus()` does have a detector argument, so I think it's cleanest if you also have the detector as an argument here.

src/odemis/driver/xt_client.py
philipwinklerdelmic: The other autofunctions use the detector as an argument, as did the code you commented out, so I thought it would be more consistent. But indeed if we are not using it and it's not going to be used in the future, I could remove it.

src/odemis/driver/test/xt_client_test.py
philipwinklerdelmic: Indeed, I thought that the is_autostigmating function wouldn't work for the flash function. Do you really want to delete it? It's not going to be used anymore? What about the set_autofocus? Delete also?

src/odemis/driver/test/xt_client_test.py
philipwinklerdelmic: I don't know, I copied the testcase from you... I thought there was a reason that it might be a good idea to test multiple times, but maybe we can delete this entire part.

src/odemis/driver/xt_client.py
philipwinklerdelmic: Yes, you're completely right!

src/odemis/driver/test/xt_client_test.py
tepals: Kornee originally wrote this code, @Kleijwegt can we delete this?

src/odemis/driver/test/xt_client_test.py
tepals: We will never use set_autostigmator and is_autostigmating, so it's safe to delete them. 

set_autofocus also has an xtlib implementation, so we should not delete that. Systems without xttoolkit can use that function, instead of the flash function.

tepals (approved):

src/odemis/driver/test/xt_client_test.py
Kleijwegt: I assume I put it there for a reason, I think the reason is that the autostigmation process wasn't reliable and sometimes didn't properly finish etc. I don't now what the current state is at the microscope. (of course on the simulator this won't be a problem)

philipwinklerdelmic: It looked to me that it was mostly there to get an idea how long the function takes on average. I deleted it in the new testcase.

philipwinklerdelmic:
@pieleric I think we can merge this PR (unless you have any more comments)

pieleric (approved):

src/odemis/driver/test/xt_client_test.py
```diff
@@ -1170,6 +1107,14 @@ def test_beam_powerVA(self):
         self.assertTrue(self.scanner.power.value)
         self.scanner.power.value = init_beam_power
 
+    def test_apply_autostigmator(self):
+        """
+        Test for the auto stigmation functionality.
+        """
+        autostigmator_future = self.scanner.applyAutoStigmator()
+        self.assertIsInstance(autostigmator_future, ProgressiveFuture)
+        autostigmator_future.result()
```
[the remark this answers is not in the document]
philipwinklerdelmic: done

PR #1794, opened 2021-06-09
sonic overview tab

philipwinklerdelmic:


src/odemis/acq/fastem.py
```diff
@@ -76,3 +77,26 @@ def _run_fake_acquisition():
...
     return len(roas) * 2
+
+
+# Overview acquisition
+def acquire_overview(coords, sem_stream):
+    """
+    :param coords: (float, float, float, float) minx, miny, maxx, maxy coordinates of overview region
+    :param sem_stream (SEMStream):
+    :returns: (ProgressiveFuture): acquisition future
+    """
```
pieleric, line 88: Note that my vision of the interface is now `acq.fastem.acquireTiledArea(stream, stage, area, live_stream=None) → Future`. But as this file is going to be changed anyway, I don't think it matters much.

src/odemis/gui/comp/miccanvas.py
```diff
@@ -1820,11 +1820,11 @@ def zoom_out(self):
...
-            sz = self._tab_data_model.scintillator_size
-            l = min(list(self._tab_data_model.scintillator_positions.values()), key=lambda item: item[0])[0] - sz[0]
-            t = max(list(self._tab_data_model.scintillator_positions.values()), key=lambda item: item[1])[1] + sz[1]
-            r = max(list(self._tab_data_model.scintillator_positions.values()), key=lambda item: item[0])[0] + sz[0]
-            b = min(list(self._tab_data_model.scintillator_positions.values()), key=lambda item: item[1])[1] - sz[1]
+            sz = self._tab_data_model.main.scintillator_size
+            l = min(list(self._tab_data_model.main.scintillator_positions.values()), key=lambda item: item[0])[0] - sz[0]
+            t = max(list(self._tab_data_model.main.scintillator_positions.values()), key=lambda item: item[1])[1] + sz[1]
+            r = max(list(self._tab_data_model.main.scintillator_positions.values()), key=lambda item: item[0])[0] + sz[0]
+            b = min(list(self._tab_data_model.main.scintillator_positions.values()), key=lambda item: item[1])[1] - sz[1]
```
pieleric, line 1827: Shouldn't we just use stage.POS_ACTIVE_RANGE? I feel it's exactly what it's supposed to represent.

src/odemis/gui/comp/miccanvas.py
```diff
@@ -1865,6 +1865,21 @@ def on_dbl_click(self, evt):
...
+        for s in tab_streams:
+            if isinstance(s, FastEMOverviewStream) and s not in ovv_streams:
+                tab_streams.remove(s)
+                self.view.removeStream(s)
+
+        # Add stream to view if it's not already there
+        for s in ovv_streams:
+            if isinstance(s, FastEMOverviewStream) and s not in tab_streams:
+                tab_streams.append(s)
+                self.view.addStream(s)
```
pieleric, line 1885: It's really, really, not the canvas who should take care of this!
The canvas is simple: it looks at the streams in the view.stream_tree, and show them.
I'd say it's up to the streams controller (or the tab controller) to decide which stream is shown on which view.

src/odemis/gui/comp/stream.py
```diff
@@ -1351,6 +1351,19 @@ def add_specselwidth_ctrl(self):
 
         return lbl_selection_width, sld_selection_width
 
+    @control_bookkeeper
+    def add_run_btn(self, label):
```
pieleric, line 1355: Could you move it near the other similar methods. For instance just after `add_text_field()`.
Also the other functions have the parameter called "label_text", so it'd be simpler to use the same name.

src/odemis/gui/comp/stream.py
```diff
@@ -1351,6 +1351,19 @@ def add_specselwidth_ctrl(self):
...
+        """
+        Add a generic run button and the corresponding side label to the gridbag sizer.
+        :param label: (str) side label describing button functionality
+        :returns: (wx.StaticText, ImageTextButton) side label and run button
+        """
+        lbl_ctrl = self._add_side_label(label)
+        metadata_btn = ImageTextButton(self._panel, label="Run...", height=16, style=wx.ALIGN_CENTER)
+        self.gb_sizer.Add(metadata_btn, (self.num_rows, 2), span=(1, 1),
```
pieleric, line 1363: "metadata_btn" -> "action_btn" or "run_btn"

src/odemis/gui/comp/stream.py
```diff
@@ -2111,3 +2124,42 @@ def __init__(self, parent, layout,
...
+        """
+        wx.Panel.__init__(self, parent, wid, pos, size, style, name)
+        self.buttons = {}  # int --> wx.Button
+
+        self._panel_sizer = wx.BoxSizer(wx.VERTICAL)
+        self.SetSizer(self._panel_sizer)
+
+    def create_controls(self, layout):
+        nrows = len(layout)
+        ncols = max([len(row) for row in layout])
```
pieleric, line 2149: Drop the `[]`.

src/odemis/gui/comp/viewport.py
```diff
@@ -1907,3 +1907,23 @@ class FastEMAcquisitionViewport(MicroscopeViewport):
...
+
+
+class FastEMOverviewViewport(FeatureOverviewViewport):
+    """ FeatureOverviewViewport (viewport with a CurrentPosCrossHairOverlay) with specialized FastEMCanvas
+    (which adds a background overlay). """
+
+    canvas_class = miccanvas.FastEMAcquisitionCanvas
+
+    def setView(self, view, tab_data):
+        super(FastEMOverviewViewport, self).setView(view, tab_data)
```
pieleric, line 1923: You can drop the arguments of `super()` now that we only use Python 3.

src/odemis/gui/model/__init__.py
```diff
@@ -1112,6 +1102,39 @@ def __init__(self, main):
         for rect in carrier_params["background"]:
             self.background.append((minx + rect[0], miny + rect[1], minx + rect[2], miny + rect[3]))
 
+        # Overview streams
+        self.overview_streams = model.VigilantAttribute({})  # dict: int --> stream or None
```
pieleric, line 1106: Typically we avoid dicts in VAs, because it's "mutable". So it's possible to change its content, without the VA noticing. As typically the int are 1->9, I think you could do with a ListVA. (Lists are mutable too, but the ListVA uses extra tricks).

src/odemis/gui/cont/tabs.py
```diff
@@ -1706,6 +1706,101 @@ def get_display_priority(cls, main_data):
...
+            self._stream_controller.enable(True)
+
+    def Show(self, show=True):
+        super(FastEMOverviewTab, self).Show(show)
+
+    @classmethod
+    def get_display_priority(cls, main_data):
+        # Tab is used only for FastEM
+        if main_data.role in ("mbsem",):
+            return 1
```
pieleric, line 1799: Which tab should be shown by default? Overview? => use a bigger priority (eg, 2)

src/odemis/gui/cont/tabs.py
```diff
@@ -1706,6 +1706,101 @@ def get_display_priority(cls, main_data):
...
+              "cls": guimod.StreamView,
+              "stream_classes": EMStream,
+              "zPos": self.tab_data_model.zPos,
+              }),
+        ])
+        self.view_controller = viewcont.ViewPortController(tab_data, panel, vpv)
+
+        # Single-beam SEM stream
+        vanames = ("resolution", "dwellTime", "horizontalFoV", "pixelSize")
+        main_data.hw_settings_config["e-beam"]["pixelSize"]["control_type"] = odemis.gui.CONTROL_READONLY
```
pieleric, line 1744: conf.data already has way to override it per type of microscope, see `HW_SETTINGS_CONFIG_PER_ROLE`. If it should be just on this tab, you should make a copy of the dict, not change it on main.

Also, I wonder if we really want to show the pixel size? We never do that, partly because it's not the pixel size that the user expects: it's the pixel size *if scale == 1*.

src/odemis/gui/cont/tabs.py
```diff
@@ -1706,6 +1706,101 @@ def get_display_priority(cls, main_data):
...
+              "stage": main_data.stage,
+              "cls": guimod.StreamView,
+              "stream_classes": EMStream,
+              "zPos": self.tab_data_model.zPos,
+              }),
+        ])
+        self.view_controller = viewcont.ViewPortController(tab_data, panel, vpv)
+
+        # Single-beam SEM stream
+        vanames = ("resolution", "dwellTime", "horizontalFoV", "pixelSize")
```
pieleric, line 1743: BTW, you might want to double check it still works fine with the latest update on xt_client, now that resolution is a read-only EnumeratedVA.
If you want to allow the user to change the resolution, I think you'd need a special control, which shows the resolution, but actually changes the scale.

src/odemis/gui/cont/streams.py
```diff
@@ -217,6 +217,10 @@ def __init__(self, stream_bar, stream, tab_data_model, show_panel=True, view=Non
         if hasattr(stream, "repetition"):
             self._add_repetition_ctrl()
 
+        if self.stream.emitter and hasattr(self.stream.emitter, "applyAutoStigmator"):
```
pieleric, line 220: Due to the way Pyro works, `hasattr(self.stream.emitter, "applyAutoStigmator")` is always True! You'd have to call it to find out it actually doesn't exist :-/ 
I can't think of an easy way. On the Delphi, I relied on the fact that .contrast and .brightness existed. Here, that's not even the case. So maybe the best is to look at main.role and check if it's "mbsem".

pieleric (changes requested):
Only a partial review. I'll look further at it later this week.

src/odemis/gui/comp/miccanvas.py
philipwinklerdelmic: I have thought quite a lot about where to put it. I know that this is not a controller class, so it shouldn't contain functionality like that. On the other hand, there are other classes in `miccanvas.py` doing things like stage moves, so I thought we don't mind so much about MVC separation here. The advantage is that the same canvas is used for the acquisition and overview tab, so we don't need to duplicate any code. My other idea was to have this code in `viewport.py` and write a superclass for `FastEMAcquisitionViewport` and `FastEMAcquisitionViewport` (with multiple inheritance) to avoid duplication, but in the end I didn't like that too much either. 

Another option is to write a new subclass `FastEMStreamsController` of `StreamBarController` and add it to both FastEM tabs. I think this might be the best option, I'll see if I can implement this.

src/odemis/gui/comp/miccanvas.py
philipwinklerdelmic: We could in principle do that. There is still the question, what position POS_ACTIVE_RANGE should be. I think you originally wanted it to indicate the corner scintillator positions. Now I use it as the corner sample carrier position (scintillator + border) because that made the background rectangle coordinates a bit easier. Still, when zooming out, we don't want to show the sample carrier border, only the scintillators. So the question is, should we change the meaning of POS_ACTIVE_RANGE, update the backgorund coordinates (border has negative values) and make the change here? Not sure if it's worth the effort to be honest. On the other hand, using the scintillator positions in POS_ACTIVE_RANGE might be more intuitive and easier to understand? What do you think?

src/odemis/gui/cont/tabs.py
philipwinklerdelmic: It was requested, but we were not aware that it's the pixel size if scale==1. I guess it's more confusing than helpful. It's definitely not essential, but I'll double check with Job if he still wants to have it.

src/odemis/gui/model/__init__.py
philipwinklerdelmic: The dict also has some advantages. It's definitely easier to index. If I want to insert the image for scinitillator 1, I just use .overview_streams[1]. For the ListVA, we need to index from 0. For now that's not such a big deal, but what if we have different scinitllator names in the future, maybe "A", "B", "C" instead of 1-9? For me the dict seemed less confusing, although I agree that it's annoying that we have to explicitly notify. If you insist, I'll change it.

src/odemis/gui/cont/tabs.py
philipwinklerdelmic: Could we just use the normal scale control instead and display the resolution as read-only?

pieleric: I think that scale + resolution read-only would be completely fine. Of course, in the longer run we might want to have something more polished...

src/odemis/gui/comp/miccanvas.py
pieleric: Yes, looks much better now!

src/odemis/gui/comp/miccanvas.py
pieleric: Right, indeed, I forgot about the extra borders that POS_ACTIVE_RANGE include.

I would note that this metadata is supposed to represent the maximum area the user can go with the stage. Soon the GUI will automatically prevent the user from going outside of it. So maybe it'd make sense to have the range limited to the scintillator's area anyway. If that's the case, then you could use it here. For now, let's leave it as-is.
Could you check with Job what's the maximum area we want the user to be able to move to?

pieleric (changes requested):

src/odemis/gui/comp/miccanvas.py
philipwinklerdelmic: The stage moves are limited by the "range" parameter in the "axes" argument of the piezomotor driver. By default, it's set to the scintillator area, so the movements are already limited to that are in the GUI.

philipwinklerdelmic:
Pushed changes.

pieleric (approved):

philipwinklerdelmic:
I think we can merge it. There is still the question what stream controls to show/ which ones to modify (pixelsize, magnification, etc), but we can also make a separate PR for that.

PR #1796, opened 2021-06-11
Bugfix updating of range of TupleContinuous

Helsloot:


pieleric:
Typically, when you fix a bug you should add also a unit test that would trigger the bug before. It's also good to explain in the commit message when the bug got trigger, and what kind of error would happen (eg, copy-paste an example backtrace). 
That said, this bug is so tiny, and I'm convinced it does exist, that I don't think we need to spend even more time on it... so I'll merge anyway.

pieleric:
@Helsloot , I've updated the commit description with more info, if you want to check.

PR #1797, opened 2021-06-15
Bug fix: Pixel value overlay blocked focussing & gadget overlay

VictoriaDel:


src/odemis/gui/comp/overlay/view.py
```diff
@@ -235,18 +234,19 @@ def on_motion(self, evt):
...
+            # the control is passed to the gadget overlay, drawing and editing of tools is possible
             evt.Skip()
 
-        else:
-            vpos = evt.Position
-            self._v_pos = Vec(vpos)
-            self.cnvs.request_drawing_update()
+        vpos = evt.Position
+        self._v_pos = Vec(vpos)
+        self.cnvs.request_drawing_update()
```
pieleric, line 249: I think you can simplify a lot by always "skipping" the event.
The only special case which is worthy to deal explicitly with is the dragging, as the position of the image is not directly accessible.
So I'd suggest to simplify the code to this (and it seems to work):
```python
        # Whatever happens, we don't keep the event, and pass it to any other interested listener.
        evt.Skip()

        # If the canvas is being dragged, the image position cannot be directly queried,
        # and anyway the cursor is above the same pixel all the time, so no update.
        if hasattr(self.cnvs, "left_dragging") and self.cnvs.left_dragging:
            return

        self._v_pos = Vec(evt.Position)
        self.cnvs.request_drawing_update()

```

pieleric (changes requested):

PR #1798, opened 2021-06-17
Move assemble zcube function 

Mahmood-B:
I made this PR, because the files changed are not relevant to the enzel acquisition panel PR. Also because they need to be merged before that PR as that PR will not work without the function assemble z cube


src/odemis/gui/util/img.py
```diff
@@ -2658,3 +2658,37 @@ def merge_screen(ima, imb):
     out.metadata = md
     return out
 
+
+def assemble_z_cube(images, zlevels):
```
pieleric, line 2662: Please move it to `odemis.util.img`, not in the GUI. One of the goal is that the "acq" module doesn't depend on the GUI. So the GUI util only contains code that is related to the GUI itself (like wxpython). 
I think the name should then be `assembleZCube`, to follow the convention in that module.

pieleric (changes requested):
Could you make the commit title clearer? It should be possible to understand it when reading the history of Odemis. So, something like "Move assembleZCub to util.img" would be nicer.
In the commit description you can also add some text to explain that it's going to be used by other code later.

pieleric (approved):
Looks good now.
Please squash the first and last commit together, and then we can merge your changes.

PR #1800, opened 2021-06-17
acq stitching tiledacq: adjust tile positions to fit better requested area

pieleric:


lazem (changes requested):
Looks good. Few comments on the test file.

lazem (approved):

PR #1803, opened 2021-06-18
Limit imaging area in GUI

Kleijwegt:


src/odemis/gui/model/__init__.py
```diff
@@ -1238,7 +1238,7 @@ def __init__(self, name, stage=None, stream_classes=None, fov_hw=None, projectio
         else:
             self.stream_classes = stream_classes
         self._stage = stage
-        
+
```
Kleijwegt, line 1241: Somehow I cannot change these lines back to the original. Maybe someone added tabs here instead of 4 spaces?

src/odemis/gui/model/__init__.py
```diff
@@ -1453,7 +1453,7 @@ def moveFocusRel(self, shift):
         # FIXME: "stop all axes" should also clear the queue
 
         # If streams have a z-level, we calculate the shift differently.
-        
+
```
Kleijwegt, line 1456: Somehow I cannot change these lines back to the original. Maybe someone added tabs here instead of 4 spaces?

tepals: This means there were 8 empty spaces on that line and you removed those spaces. So now it's a completely empty line.

Kleijwegt: I tried again to add 8 spaces instead of leaving the line empty and it still pop-up....

Kleijwegt: Probably this is a pycharm setting which converts these things during committing, I will check those settings.

tepals: Maybe other people disagree, but I don't think it's bad to do this. Especially since it's only on two lines in your PR, it doesn't mess up the entire PR. And it's cleaner to have completely empty lines, which you will now have. 

In Pycharm settings you can go to Editor > General, then all the way at the bottom there is a section On Save. I think if that is set to Remove trailing spaces on all lines, it will also remove spaces on empty lines.

src/odemis/gui/model/__init__.py
```diff
@@ -1562,23 +1577,70 @@ def moveStageTo(self, pos):
...
 
         logging.debug("Requesting stage to move to %s m", move)
         f = self._stage.moveAbs(move)
         self._fstage_move = f
         f.add_done_callback(self._on_stage_move_done)
         return f
 
+    def _getStageLimitsXY(self):
+        """
+        Based on the physical stage limit and the area of the image used for imaging the stage limits are return in
```
tepals, line 1600: typo: are return**ed**

src/odemis/gui/model/__init__.py
```diff
@@ -1562,23 +1577,70 @@ def moveStageTo(self, pos):
...
+        a dict. (MD_POS_ACTIVE_RANGE defines the area which can be used for imaging)
+        If no stage limits are defined an empty dict is returned.
+
+        :return (dictionary): dict which contains the stage limits in x and y direction
+        """
+        stage_limits = {}
+        # Physical stage limits
+        if "x" in self._stage.axes and hasattr(self._stage.axes["x"], "range"):
+            stage_limits.update({"x": [self._stage.axes["x"].range[0], self._stage.axes["x"].range[1]]})
```
tepals, line 1609: Can't you just do:
```
stage_limits.update({"x": self._stage.axes["x"].range})
```

src/odemis/gui/model/__init__.py
pieleric: Indeed, please don't spend too much time on that. It's better to not fix whitespace in the same commit, but it's even worse to spend 1h to put back whitespace ;-)
Your PR is still readable, so let's not worry too much.

src/odemis/gui/model/__init__.py
```diff
@@ -1562,23 +1577,70 @@ def moveStageTo(self, pos):
...
+        :param new_limits (dict): Contains new limits for the stage for one or multiple axis
+        :return (dict): Contains the updated stage limits
+        """
+        for key in new_limits:
+            if key in stage_limits:
+                # Update the stage limits with the intersection of both the existing and new limits.
+                stage_limits.update({key: [max(stage_limits[key][0], new_limits[key][0]),
+                                           min(stage_limits[key][1], new_limits[key][1])]})
+            else:
+                stage_limits.update({key: [stage_limits[key][0], new_limits[key][1]]})
```
tepals, line 1640: I think this doesn't work if key is not in stage_limits. Shouldn't this be:
```
stage_limits.update({key: [new_limits[key][0], new_limits[key][1]]})
```
Which can then also be rewritten to:
```
stage_limits.update({key: new_limits[key]})
```

tepals (changes requested):

src/odemis/gui/model/__init__.py
pieleric, line 1609: `update()` doesn't make it more readable here. You can simply write `stage_limits["x"] = ...`

src/odemis/gui/model/__init__.py
```diff
@@ -1562,23 +1577,70 @@ def moveStageTo(self, pos):
...
-            if (hasattr(axis_def, "range") and
-                not axis_def.range[0] <= p <= axis_def.range[1]
-               ):
-                p = max(axis_def.range[0], min(p, axis_def.range[1]))
-                move[ax] = p
-                logging.info("Restricting stage axis %s move to %g mm due to stage limit",
-                             ax, p * 1e3)
+        # If needed clip current movements in x/y direction to the maximum allowed stage limits
+        stage_limits = self._getStageLimitsXY()
+        if "x" in stage_limits and not stage_limits["x"][0] <= move["x"] <= stage_limits["x"][1]:
```
pieleric, line 1582: In theory stage_limits may not contain all the axes. So you'd need to check this too.
=> move the block to a subfunction?

src/odemis/gui/model/__init__.py
```diff
@@ -1562,23 +1577,70 @@ def moveStageTo(self, pos):
...
+            move["x"] = max(stage_limits["x"][0], min(move["x"], stage_limits["x"][1]))
+            logging.info("Movements of the stage in x direction are limited by the stage limit, the movement is "
+                         "restricted to %s mm." % (move["x"] * 1e3))
```
pieleric, line 1585: You can make the log a little less verbose, and still say more, like:
`"Movements of the stage in x limited to %s, restricting move to %s mm." % (stage_limits["x"], move["x"] * 1e3)`

pieleric (changes requested):

src/odemis/gui/model/__init__.py
Kleijwegt: >And it's cleaner to have completely empty lines, which you will now have. [From @tepals]
>but it's even worse to spend 1h to put back whitespace ;-) [ From @pieleric]
Great, I wasn't particularly looking forward to spending more time on trying to fix this.

src/odemis/gui/model/__init__.py
```diff
@@ -1562,23 +1577,70 @@ def moveStageTo(self, pos):
...
+        stage_limits = {}
+        # Physical stage limits
+        if "x" in self._stage.axes and hasattr(self._stage.axes["x"], "range"):
+            stage_limits.update({"x": [self._stage.axes["x"].range[0], self._stage.axes["x"].range[1]]})
+        if "y" in self._stage.axes and hasattr(self._stage.axes["y"], "range"):
+            stage_limits.update({"y": [self._stage.axes["y"].range[0], self._stage.axes["y"].range[1]]})
+
+        # Area which can be used for imaging
+        if model.MD_POS_ACTIVE_RANGE in self._stage.getMetadata():
+            pos_active_range = self._stage.getMetadata()[model.MD_POS_ACTIVE_RANGE]
```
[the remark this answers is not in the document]
Kleijwegt: I really like this one I didn't know it was possible to supply the argument which is returned if the metadata is not found

src/odemis/gui/model/__init__.py
Kleijwegt: That had to do with a needed conversion from tuple to list. But I fixed that.

tepals: I also mean that your first argument is **stage**_limits[0] and your second argument is **new**_limits[1]

src/odemis/gui/model/__init__.py
pieleric: Discard this comment. This was about `move`. But actually, it always contains both axes.

src/odemis/gui/comp/canvas.py
```diff
@@ -1548,8 +1548,17 @@ def on_dbl_click(self, evt):
...
+            new_pos = [self.p_buffer_center[0] + offset[0],
+                       self.p_buffer_center[1] + offset[1]]
+
+            # Update the position within the allowed limits. The Streamview class logs if the movement is limited.
+            stage_limits = self.view._getStageLimitsXY()
+            if "x" in stage_limits and not stage_limits["x"][0] <= new_pos[0] <= stage_limits["x"][1]:
+                new_pos[0] = max(stage_limits["x"][0], min(new_pos[0], stage_limits["x"][1]))
+
+            if "y" in stage_limits and not stage_limits["y"][0] <= new_pos[1] <= stage_limits["y"][1]:
+                new_pos[1] = max(stage_limits["y"][0], min(new_pos[1], stage_limits["y"][1]))
```
pieleric, line 1560: Let's avoid using a "private" method from the StreamView.
I'd suggest to provide directly a dedicated method that just does what we need: `.clipToStageLimits(pos)` (adjust the name to your taste). This method could even be also used by `.moveStageTo()`.

src/odemis/gui/comp/canvas.py
```diff
@@ -1548,8 +1548,17 @@ def on_dbl_click(self, evt):
 
             # recompute the view
             offset = (-shift[0] / self.scale, shift[1] / self.scale)
-            new_pos = (self.p_buffer_center[0] + offset[0],
-                       self.p_buffer_center[1] + offset[1])
+            new_pos = [self.p_buffer_center[0] + offset[0],
+                       self.p_buffer_center[1] + offset[1]]
+
+            # Update the position within the allowed limits. The Streamview class logs if the movement is limited.
+            stage_limits = self.view._getStageLimitsXY()
```
pieleric, line 1555: You cannot put this code in this class, as it doesn't have a `.view` attribute. You should put this on miccanvas.DblMicroscopeCanvas. I think `.recenter_buffer()` would be even better anyway, as it'd also catch cases of moving the canvas in different ways, like using the keyboard.

src/odemis/gui/model/__init__.py
```diff
@@ -1606,17 +1606,16 @@ def _getStageLimitsXY(self):
         stage_limits = {}
         # Physical stage limits
         if "x" in self._stage.axes and hasattr(self._stage.axes["x"], "range"):
-            stage_limits.update({"x": [self._stage.axes["x"].range[0], self._stage.axes["x"].range[1]]})
+            stage_limits["x"] = list(self._stage.axes["x"].range)
         if "y" in self._stage.axes and hasattr(self._stage.axes["y"], "range"):
-            stage_limits.update({"y": [self._stage.axes["y"].range[0], self._stage.axes["y"].range[1]]})
+            stage_limits["y"] = list(self._stage.axes["x"].range)
```
pieleric: After reading again the code, I realise that `move` always contains x and y. In general, a stage without x and y is not supported. So you can just assume that x and y are present on the stage as well. It'll simplify a bit the code.

src/odemis/gui/model/__init__.py
```diff
@@ -1606,17 +1606,16 @@ def _getStageLimitsXY(self):
...
-        if model.MD_POS_ACTIVE_RANGE in self._stage.getMetadata():
-            pos_active_range = self._stage.getMetadata()[model.MD_POS_ACTIVE_RANGE]
-            if "x" in pos_active_range and "x" in stage_limits:
-                stage_limits = self._updateStageLimits(stage_limits, {"x": pos_active_range["x"]})
-            if "y" in pos_active_range and "y" in stage_limits:
-                stage_limits = self._updateStageLimits(stage_limits, {"y": pos_active_range["y"]})
+        pos_active_range = self._stage.getMetadata().get(model.MD_POS_ACTIVE_RANGE, {})
+        if "x" in pos_active_range and "x" in stage_limits:
+            stage_limits = self._updateStageLimits(stage_limits, {"x": pos_active_range["x"]})
+        if "y" in pos_active_range and "y" in stage_limits:
```
pieleric, line 1617: Just assume that x and y are in stage_limits here too.

pieleric (changes requested):

src/odemis/gui/model/__init__.py
```diff
@@ -1554,31 +1568,86 @@ def moveStageToView(self):
...
+        logging.debug("Requesting stage to move to %s mm in x direction and %s mm in y direction",
+                      move["x"] * 1e3, move["y"] * 1e3)
         f = self._stage.moveAbs(move)
         self._fstage_move = f
         f.add_done_callback(self._on_stage_move_done)
         return f
 
+    def clipToStageLimits(self, pos):
+        """
+        Clip current a position in x/y direction to the maximum allowed stage limits.
```
tepals, line 1589: remove the "a" from "clip current a postion"

tepals (approved):
one tiny comment

src/odemis/gui/model/__init__.py
```diff
@@ -1554,31 +1568,86 @@ def moveStageToView(self):
...
+
+        :param pos (dict): Position to be clipped with keys "x" and "y"
+        :return(dict): Position clipped to the stage limits with keys "x" and "y"
+        """
+        stage_limits = self._getStageLimitsXY()
+        if not stage_limits["x"][0] <= pos["x"] <= stage_limits["x"][1]:
+            pos["x"] = max(stage_limits["x"][0], min(pos["x"], stage_limits["x"][1]))
+            logging.info("Movements of the stage in x limited to %s mm, restricting movement to %s mm.",
+                         stage_limits["x"], pos["x"] * 1e3)
```
pieleric, line 1598: `stage_limits["x"]` must also be `* 1e3` as you're indicating "mm" in the message. (same thing for y). However, this is a list. So it's not very convenient. I expect that this type of message is mostly only for debugging or understanding unexpected behaviour of the software. So you don't have to make it too pretty. It'd fine to report the position in meters.

src/odemis/gui/model/__init__.py
```diff
@@ -1554,31 +1568,86 @@ def moveStageToView(self):
...
+            stage_limits["x"] = list(self._stage.axes["x"].range)
+        if hasattr(self._stage.axes["y"], "range"):
+            stage_limits["y"] = list(self._stage.axes["x"].range)
+
+        # Area which can be used for imaging
+        pos_active_range = self._stage.getMetadata().get(model.MD_POS_ACTIVE_RANGE, {})
+        if "x" in pos_active_range:
+            stage_limits = self._updateStageLimits(stage_limits, {"x": pos_active_range["x"]})
+        if "y" in pos_active_range:
+            stage_limits = self._updateStageLimits(stage_limits, {"y": pos_active_range["y"]})
```
pieleric, line 1626: `pos_active_range` is the right format already. So you can just write:
`stage_limits = self._updateStageLimits(stage_limits, pos_active_range)`

pieleric (changes requested):

src/odemis/gui/model/__init__.py
Kleijwegt: I actually only want to update the stage limits in _x/y_ if somehow a limit in _z_ is included in MD_POS_ACTIVE_RANGE it should not update the stage limits with that. This implementation should limit the area in which you can scroll through the image not the focus for example.
By passing the entire dict "_pos_active_range_" that might happen. (It could include more than just those two axis isn't it?)

If we don't care about updating also other axis the implementation should become: 
```
        if "x" in pos_active_range or "y" in pos_active_range:
            stage_limits = self._updateStageLimits(stage_limits, pos_active_range)
```
To prevent updating the same limits twice.

PR #1804, opened 2021-06-18
technolution driver: add hack to skip the first field image

Sabrina85:
The current ASM HW does not scan the very first field image correctly. This issue needs to be fixed in HW. However, until this is done, we need to "through away" the first field image and scan it a second time to receive a good first field image. To do so, just always scan the first image twice and overwrite on disc (external storage).

### DIFF
--- a/src/odemis/driver/technolution.py
+++ b/src/odemis/driver/technolution.py
@@ -1145,6 +1145,13 @@ class MPPC(model.Detector):
                     dataContent = args[1]  # Specifies the type of image to return (empty, thumbnail or full)
                     notifier_func = args[2]  # Return function (usually, dataflow.notify or acquire_single_field queue)
 
+                    # FIXME: Hack: The current ASM HW does not scan the very first field image correctly. This issue
+                    #  needs to be fixed in HW. However, until this is done, we need to "through away" the first field
+                    #  image and scan it a second time to receive a good first field image. To do so, just always scan
+                    #  the first image twice.
+                    if field_data.position_x == 0 and field_data.position_y == 0:
+                        self.parent.asmApiPostCall("/scan/scan_field", 204, field_data.to_dict())
+
                     self.parent.asmApiPostCall("/scan/scan_field", 204, field_data.to_dict())
 
                     if DATA_CONTENT_TO_ASM[dataContent] is None:
--- a/src/odemis/driver/technolution.py
+++ b/src/odemis/driver/technolution.py
@@ -1146,7 +1146,7 @@ class MPPC(model.Detector):
                     notifier_func = args[2]  # Return function (usually, dataflow.notify or acquire_single_field queue)
 
                     # FIXME: Hack: The current ASM HW does not scan the very first field image correctly. This issue
-                    #  needs to be fixed in HW. However, until this is done, we need to "through away" the first field
+                    #  needs to be fixed in HW. However, until this is done, we need to "throw away" the first field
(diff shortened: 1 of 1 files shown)


src/odemis/driver/technolution.py
```diff
@@ -1145,6 +1145,13 @@ def _acquire(self):
                     dataContent = args[1]  # Specifies the type of image to return (empty, thumbnail or full)
                     notifier_func = args[2]  # Return function (usually, dataflow.notify or acquire_single_field queue)
 
+                    # FIXME: Hack: The current ASM HW does not scan the very first field image correctly. This issue
+                    #  needs to be fixed in HW. However, until this is done, we need to "through away" the first field
```
tepals, line 1149: typo: through -> throw

src/odemis/driver/technolution.py
```diff
@@ -1145,6 +1145,13 @@ def _acquire(self):
...
+                    #  image and scan it a second time to receive a good first field image. To do so, just always scan
+                    #  the first image twice.
+                    if field_data.position_x == 0 and field_data.position_y == 0:
+                        self.parent.asmApiPostCall("/scan/scan_field", 204, field_data.to_dict())
```
tepals, line 1153: Why not put this in the if statement of line 1129. Then you know for sure you're starting the megafield and acquiring the first image, or am I mistaken?

src/odemis/driver/technolution.py
Sabrina85: indeed thanks

src/odemis/driver/technolution.py
Sabrina85: In the "start" part, you don't have any information on the field position available. Only the settings or MD are received with the "start" argument. However, you could argue that we could hardcode the position. So in that respect there is no reason to not to do so. 
However, a reason to leave it here could be my other open pr. There I modified the code to catch errors during the start or during a scan of a field. If you would then mix start and scan of the first field, it might be less clean in terms of what caused the error. But not sure. 

My preference would be still to keep start and scanning separate and do not mix the two.

Kleijwegt (approved):

src/odemis/driver/technolution.py
tepals: Then you can leave it here.

tepals (approved):

src/odemis/driver/technolution.py
```diff
@@ -1145,6 +1145,13 @@ def _acquire(self):
                     dataContent = args[1]  # Specifies the type of image to return (empty, thumbnail or full)
                     notifier_func = args[2]  # Return function (usually, dataflow.notify or acquire_single_field queue)
 
+                    # FIXME: Hack: The current ASM HW does not scan the very first field image correctly. This issue
+                    #  needs to be fixed in HW. However, until this is done, we need to "throw away" the first field
+                    #  image and scan it a second time to receive a good first field image. To do so, just always scan
+                    #  the first image twice.
+                    if field_data.position_x == 0 and field_data.position_y == 0:
+                        self.parent.asmApiPostCall("/scan/scan_field", 204, field_data.to_dict())
```
pieleric, line 1153: Please add something like `logging.debug("Rescanning first field to workaround hardware limitation")` . This way, it'd be less weird when we look at the log and see twice the same call.

pieleric (changes requested):

PR #1805, opened 2021-06-18
technolution-errorpropagation

Sabrina85:


tepals:
I don't really know enough about this to do a proper review. Just one comment: you have a lot of stylistic and grammar changes, could you put those in a separate commit next time? That will help make your pull request a lot more readable and easier to review.

Sabrina85:
> I don't really know enough about this to do a proper review. Just one comment: you have a lot of stylistic and grammar changes, could you put those in a separate commit next time? That will help make your pull request a lot more readable and easier to review.

Yes I know that this is not ideal.

src/odemis/driver/technolution.py
```diff
@@ -1131,9 +1131,17 @@ def _acquire(self):
...
+                    notifier_func = args[1]  # Return function (usually, dataflow.notify or acquire_single_field queue)
+                    try:
+                        self.parent.asmApiPostCall("/scan/start_mega_field", 204, megafield_metadata.to_dict())
+                        notifier_func(None)  # return None if all went fine
+                    except Exception as ex:
+                        logging.error("During the start of the acquisition an error has occurred: %s.", ex)
+                        notifier_func(ex)
+                        continue  # let the caller decide on what to do next
+
+                    acquisition_in_progress = True  # acquisition was successfully started
```
Kleijwegt, line 1144: If an error is found the var acquisition_in_progress is set to True while actually the acquisition isn't really started/in progress. That seems a bit odd.

src/odemis/driver/technolution.py
```diff
@@ -1131,9 +1131,17 @@ def _acquire(self):
...
                         continue
 
-                    acquisition_in_progress = True
                     megafield_metadata = args[0]
-                    self.parent.asmApiPostCall("/scan/start_mega_field", 204, megafield_metadata.to_dict())
+                    notifier_func = args[1]  # Return function (usually, dataflow.notify or acquire_single_field queue)
+                    try:
+                        self.parent.asmApiPostCall("/scan/start_mega_field", 204, megafield_metadata.to_dict())
+                        notifier_func(None)  # return None if all went fine
+                    except Exception as ex:
```
Kleijwegt, line 1139: The only expected exception here would be the AsmApiException isn't it from asmApiPostCall? Should we not limit the catch to just that exception? Instead of catching all the exceptions? Even the error is already logged in that method.

src/odemis/driver/technolution.py
```diff
@@ -1145,23 +1153,29 @@ def _acquire(self):
...
+                            resp = self.parent.asmApiGetCall(
+                                    "/scan/field?x=%d&y=%d&thumbnail=%s" %
+                                    (field_data.position_x, field_data.position_y,
+                                     str(DATA_CONTENT_TO_ASM[dataContent]).lower()),
+                                    200, raw_response=True, stream=True)
+                            resp.raw.decode_content = True  # handle spurious Content-Encoding
+                            img = Image.open(BytesIO(base64.b64decode(resp.raw.data)))
+
+                            da = model.DataArray(img, metadata=self._metadata)
+                    except Exception as ex:
```
Kleijwegt, line 1174: The same as above:
>The only expected exception here would be the AsmApiException isn't it from asmApiPostCall? Should we not limit the catch to just that exception? Instead of catching all the exceptions? Even the error is already logged in that method.

src/odemis/driver/test/technolution_test.py
```diff
@@ -1517,5 +1486,73 @@ def test_get_and_subscribe(self):
...
+        self.MPPC.filename.value = time.strftime("testing_megafield_id-%Y-%m-%d-%H-%M-%S")
+        dataflow = self.MPPC.data
+
+        init_cellCompleteRes = self.MPPC.cellCompleteResolution.value
+        # force a value on a VA which will raise an error on the ASM
+        # use ._value in order to circumvent the setter checks
+        self.MPPC.cellCompleteResolution._value = (1100, 1100)  # max value allowed on HW is 1000; force it!
+
+        # raise exception due to wrong value in settings uploaded to ASM during the start of the acquisition
+        with self.assertRaises(Exception):
```
Kleijwegt, line 1501: In these test cases I have the feeling that the exact expected error should be specified to make it really succeed on that error only. In this case even when the connection isn't perfect on the moment of the test (or some other arbitary error) the testcase might succeed. While we actually only want the test case to succeed if the exact error we expect is raised.

Kleijwegt (changes requested):

src/odemis/driver/technolution.py
pieleric: I think that's fine. Normally, only AsmApiExceptions are raised. But if there are other exceptions (ex, the connection failed, division by zero, index out of bound...) the mega field is most likely not started. It's still worth reporting it.

src/odemis/driver/technolution.py
Sabrina85: If an exception is raised, the continue will be executed and this line is not reached. So the variable `acquisition_in_progress` will still be false. Only if the start succeeded (so the try statement was successfully executed) this line will be reached.

src/odemis/driver/test/technolution_test.py
```diff
@@ -1517,5 +1486,73 @@ def test_get_and_subscribe(self):
...
+
+        self.MPPC.filename.value = time.strftime("testing_megafield_id-%Y-%m-%d-%H-%M-%S")
+        dataflow = self.MPPC.data
+
+        img_queue_1 = queue.Queue()
+        img_queue_2 = queue.Queue()
+
+        def image_received_1(dataflow, da):
+            img_queue_1.put(da)
+            logging.debug("image 1 received")
```
pieleric, line 1521: The message is a little bit confusing. I'd suggest "Image received by subscriber 1".

src/odemis/model/test/dataflow_test.py
```diff
@@ -159,22 +164,42 @@ def synchronizedOn(self, event):
         if self._sync_event:
             self._sync_event.subscribe(self)
 
-        
+
+class DataFlowRaiseError(model.DataFlow):
+    """Very simple dataflow which raises an error during the start of the data generation."""
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
pieleric, line 171: You can drop this, if there is nothing special to do.

pieleric (changes requested):

src/odemis/driver/technolution.py
```diff
@@ -1296,17 +1323,41 @@ def acquireSingleField(self, dataContent="thumbnail", field_num=(0, 0)):
 
         self._ensure_acquisition_thread()
 
-        self.acq_queue.put(("start", mega_field_data))
+        # request to start the acquisition
+        self.acq_queue.put(("start", mega_field_data, return_queue.put))
+        try:
+            status_start = return_queue.get(timeout=30)
```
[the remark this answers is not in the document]
Sabrina85: 30s is a bit long. However, it is also not necessarily < 0.1s as the upload for the settings sometimes is a bit slow (at least at FumoBeta; for EA1 it was supposed to be faster if I remember correctly). I think we can go for 5s.

src/odemis/driver/technolution.py
```diff
@@ -1296,17 +1323,41 @@ def acquireSingleField(self, dataContent="thumbnail", field_num=(0, 0)):
...
 
+        # request to scan a single field image
         self.acq_queue.put(("next", field_data, dataContent, return_queue.put))
+        try:
+            status_next = return_queue.get(timeout=600)
+        except queue.Empty as ex:
+            logging.error("Acquisition of the single field image timed out.")
+            raise ex
+        if isinstance(status_next, Exception):
+            logging.error("Received an exception from the acquisition thread.")
```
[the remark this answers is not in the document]
Sabrina85: Yes this was supposed to be a debug.

src/odemis/model/test/dataflow_test.py
Sabrina85: If I drop it, I receive the following error when running the test:
```
Error
  File "/home/sabrina/development/odemis/src/odemis/model/test/dataflow_test.py", line 329, in test_df_raise_error
    im = self.df.get()
  File "/home/sabrina/development/odemis/src/odemis/model/_dataflow.py", line 326, in get
    self.subscribe(receive_one_image)
  File "/home/sabrina/development/odemis/src/odemis/model/_dataflow.py", line 335, in subscribe
    with self._lock:
AttributeError: 'DataFlowRaiseError' object has no attribute '_lock'
```
So it looks like it is still necessary in order to have the lock attribute. Same for `_listeners` and` _remote_listeners`. So either I need to leave the super or I need to instantiate these attributes in the init I think. Is it ok to leave the` super`?

src/odemis/driver/technolution.py
```diff
@@ -1340,21 +1340,20 @@ def acquireSingleField(self, dataContent="thumbnail", field_num=(0, 0)):
...
             logging.error("Acquisition of the single field image timed out.")
-            raise ex
+            raise TimeoutError("Acquisition of the single field image timed out after 600s.")
+
         if isinstance(status_next, Exception):
-            logging.error("Received an exception from the acquisition thread.")
+            logging.debug("Received an exception from the acquisition thread.")
             raise status_next
+        elif not isinstance(status_next, model.DataArray):
+            raise ValueError("Did not receive image data but the following: %s.", status_next)  # should never happen
```
pieleric, line 1351: Here, you have to use `%`.

pieleric (changes requested):

src/odemis/model/test/dataflow_test.py
pieleric: Sorry, I didn't mean to say to drop *just* that line, but the whole `def __init__()`. In this case, the function will not be overwritten, and so the "super" will be called instead, automatically.

pieleric (approved):
Looks good to me.
Could you squash back everything into two commits (model / technolution)? Then it should be ready to merge :-)

Sabrina85:
Done!

PR #1808, opened 2021-06-22
Orsay fib device driver

Helsloot:


install/linux/usr/share/odemis/hwtest/orsay.odm.yaml
```diff
@@ -50,3 +54,18 @@ Orsay: {
     role: gis-reservoir,
     init: {}
 }
+
+"FIB-device": {
+    role: fib-device,
+    init: {}
```
Kleijwegt, line 60: I think these init: {} are not required in the yaml. That said I think it's not in the way and maybe even looks better.

src/odemis/driver/orsay.py
```diff
@@ -148,6 +164,33 @@ def __init__(self, name, role, children, host, daemon=None, **kwargs):
...
+            logging.info("Orsay was not given a 'fib-device' child")
+        else:
+            self._fib_device = FIBDevice(parent=self, daemon=daemon, **kwargs)
+            self.children.value.add(self._fib_device)
+
+        # create the FIB source child
+        try:
+            kwargs = children["fib-source"]
+        except (KeyError, TypeError):
+            logging.info("Orsay was not given a 'fib-source' child")
```
Kleijwegt, line 180: I'm not that into the naming/branding of Orsay but I thought Orsay was a brand isn't it? Then maybe it's nicer to rewrite this log to:
> "The Orsay microscope was not given a 'fib-source' child"

If you think that makes any sense.

Helsloot: That does make sense, I've updated it to 'The Orsay component'...

src/odemis/driver/orsay.py
```diff
@@ -94,11 +110,13 @@ def __init__(self, name, role, children, host, daemon=None, **kwargs):
                                                            daemon=True)
         self._connection_monitor_thread.start()
 
+        no_child_str = "The Orsay component was not given a '%s' child"
```
Kleijwegt, line 113: Maybe change the naming to "no_child_msg". I didn't understand this setup at first. But it is really nice actually

src/odemis/driver/orsay.py
```diff
@@ -584,7 +635,7 @@ def _changeVacuum(self, goal):
         logging.debug("Setting vacuum status to %s." % self.axes["vacuum"].choices[goal])
         self._vacuumStatusReached.clear()  # to make sure it will wait
         self._chamber.VacuumStatus.Target = goal
-        if not self._vacuumStatusReached.wait(18000):  # wait maximally 5 hours
+        if not self._vacuumStatusReached.wait(18000):  # wait maximally 5 hours  TODO: TUNE THIS?
```
Kleijwegt, line 638: As a user if I need to wait 5 hours on something I would like to have a least a regular feedback in the log that the whole system isn't freezed but actively pumping or something.
Is such a message/pop-up or something provided if pumping takes more than 10 minutes?

src/odemis/driver/orsay.py
```diff
@@ -1389,3 +1440,1097 @@ def terminate(self):
...
+    """
+    Attribute that is connected to a VA and a parameter on the Orsay server.
+    If VA is readonly, the VA will be kept up to date of the changes of the Orsay parameter, but force writing to the VA
+    will not update the Orsay parameter.
+    If VA is not readonly, writing to the VA will write this value to the Orsay parameter's Target attribute.
+    """
+
+    def __init__(self, va, parameter, attributeName="Actual", conversion=None, factor=None, minpar=None, maxpar=None):
+        """
+        va is the vigilant attribute this Orsay parameter connector should be connected to. This VA should not have a
```
Kleijwegt, line 1455: This docstring does not follow the default convention of docstings. In odemis we use two styles I would suggest to choose one of those.

https://www.python.org/dev/peps/pep-0257/

src/odemis/driver/orsay.py
```diff
@@ -1389,3 +1440,1097 @@ def terminate(self):
...
+        self._va_value_type = None
+        self._conversion = conversion
+        self._factor = factor
+        self._minpar = minpar
+        self._maxpar = maxpar
+        self.connect(va, parameter, attributeName)
+
+    def connect(self, va, parameter, attributeName="Actual"):
+        """
+        va is the vigilant attribute this Orsay parameter connector should be connected to. This VA should not have a
```
Kleijwegt, line 1483: Also incorrect formatting of the docstring

src/odemis/driver/orsay.py
```diff
@@ -1389,3 +1440,1097 @@ def terminate(self):
...
+            self._parameters = None
+            self._attributeName = None
+            self._va._setter = WeakMethod(self._va._VigilantAttribute__default_setter)
+            self._va = None
+            self._va_type_name = None
+            self._conversion = None
+
+    def update_VA(self, parameter=None, attributeName=None):
+        """
+        parameter (Orsay Parameter): the parameter on the Orsay server that calls this callback
```
Kleijwegt, line 1596: Actually this is almost correct form of formatting. I you would like to explain the function you can add a line at the top so like this:
```
        Copies the value of the parameter to the VA
        parameter (Orsay Parameter): the parameter on the Orsay server that calls this callback
        attributeName (str): the name of the attribute of parameter which was changed
```

Kleijwegt (changes requested):

src/odemis/driver/orsay.py
Helsloot: This issue is solved in another pending pull request, with all results from the hardware tests done. Basically, it generally only takes 10 minutes, so it waits maximally 30 minutes. In the mean time values like chamber pressure, pumping speed etc. are kept up to date, so the user should be able to see those values change and so know the program did not freeze.

src/odemis/driver/orsay.py
```diff
@@ -1389,3 +1440,1097 @@ def terminate(self):
...
+        • compressedAirPressure: FloatContinuous, readonly, unit="Pa", range=(0, 5e6)
+        """
+
+        super().__init__(name, role, parent=parent, **kwargs)
+
+        self._gunPump = None
+        self._columnPump = None
+        self._interlockHVPS = None
+        self._interlockChamber = None
+        self._valve = None
```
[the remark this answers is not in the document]
Helsloot: I'll add a comment. I define them here, because otherwise Pycharm complains that I am defining attributes in a place other than init and that's not nice. Related to what you said: If someone looks for the source, they'll start searching in init.

src/odemis/driver/orsay.py
```diff
@@ -1389,3 +1440,1097 @@ def terminate(self):
...
+        """
+        Update the VA's. Should be called after reconnection to the server
+        """
+        self._updateErrorState()
+        self._updateInterlockTriggered()
+        self._updateValveOpen()
+        for obj_name in self._connectorList:
+            getattr(self, obj_name).update_VA()
+
+    def _updateErrorState(self, parameter=None, attributeName="Actual"):
```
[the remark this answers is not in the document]
Helsloot: This is a thing from Orsay. Each of their parameters has an Actual and a Target attribute, which each contain a value. Generally when you want to access the value of a parameter, for example one called Pressure, you'd do that by using Pressure.Actual. But when subscribing to a parameter, the callback function needs to accept two arguments: the parameter (as an object) and the attribute name (as a string). This attribute name could then be "Actual" or "Target" (there are some other options, but those are irrelevant for now.) The main advantages of using a constant to store the "Actual" string would be that it only has to be changed in one place if Orsay decides to not call it Actual anymore, and that it decreases the risk of typos in the use of the "Actual" string. For the former argument, since the parameter.Actual notation is used whenever possible, there are many places this would have to be changed anyway (Ctrl+H will help). The latter argument does not weigh heavily enough I think, since any typo I would find through unittests. Do you agree?

src/odemis/driver/orsay.py
```diff
@@ -1389,3 +1440,1097 @@ def terminate(self):
...
+        • extractorVoltage: FloatContinuous, unit="V", range=(0, 12e3)
+        """
+
+        super().__init__(name, role, parent=parent, **kwargs)
+
+        self._hvps = None
+        self._ionColumn = None
+
+        self.gunOn = model.BooleanVA(False)
+        self.gunOnConnector = None
```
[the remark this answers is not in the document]
Helsloot: You're right, they should be private.
The connector indeed also works if the gun is off, but, to be consistent with the notation I've used for other VA's and their connectors, this is the connector of the gunOn VA.

src/odemis/driver/orsay.py
```diff
@@ -1389,3 +1440,1097 @@ def terminate(self):
...
+
+class FIBBeam(model.HwComponent):
+    """
+    Represents the beam of the Focused Ion Beam (FIB) from Orsay Physics. It contains many beam optics settings.
+    """
+
+    def __init__(self, name, role, parent, **kwargs):
+        """
+        Defines the following VA's and links them to the callbacks from the Orsay server:
+        • blanker: VAEnumerated, choices={True: "blanking", False: "no blanking", None: "imaging"}
```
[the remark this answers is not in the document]
Helsloot: Fair point. Especially since, if the range is defined on the Orsay software, the range specified on the VA will be overwritten by the range defined on the Orsay software. So it might not be correct anyway.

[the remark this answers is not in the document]
Helsloot: Btw, because of the ridiculous amount of parameters to control, we wanted to use something like the Connector class, to prevent repeating a LOT of similar code. And this part of the driver does not actually offer much functionality yet. The Detector and Scanner classes will offer the actual functionality, like imaging, and use the FIBBeam, FIBSource and FIBDevice for communication with the Orsay software.

src/odemis/driver/test/orsay_test.py
```diff
@@ -64,7 +72,7 @@ def test_creation(self):
             oserver = orsay.OrsayComponent(**CONFIG_ORSAY)
         except Exception as e:
             self.fail(e)
-        self.assertEqual(len(oserver.children.value), 6)
+        self.assertEqual(len(oserver.children.value), 9)
```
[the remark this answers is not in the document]
Helsloot: How would you test this?

src/odemis/driver/orsay.py
```diff
@@ -1389,3 +1440,1247 @@ def terminate(self):
...
+        self.rotation = model.FloatContinuous(0.0, unit="rad", range=(-pi, pi))
+        self._rotationConnector = None
+        self.dwellTime = model.FloatEnumerated(1e-7, unit="s",
+                                               choices={1e-3, 5e-4, 1e-4, 5e-5, 1e-5, 5e-6, 1e-6, 5e-7, 2e-7, 1e-7})
+        self._dwellTimeConnector = None
+        self.contrast = model.FloatContinuous(1.0, unit="", range=(0, 1))
+        self._contrastConnector = None
+        self.brightness = model.FloatContinuous(1.0, unit="", range=(0, 1))
+        self._brightnessConnector = None
+        self.operatingMode = model.BooleanVA(False)
```
pieleric, line 2329: What is that? It feels strange that the "mode" can be True of False. Is that the "power"?

src/odemis/driver/orsay.py
```diff
@@ -1389,3 +1440,1247 @@ def terminate(self):
...
+        elif new_translation[1] > tran_limit_1:
+            new_translation[1] = tran_limit_1
+
+        target_translation = [0, 0]  # keep centre where it was, move target_trans from centre to upper left corner
+        target_translation[0] = int(self.imageFormat.value[0] / 2 + new_translation[0] - self.resolution.value[0] / 2)
+        target_translation[1] = int(self.imageFormat.value[1] / 2 - new_translation[1] - self.resolution.value[1] / 2)
+
+        target = map(str, target_translation + list(self.resolution.value))
+        target = " ".join(target)
+        self._ionColumn.ImageArea.Target = target
```
pieleric, line 2589: All the code which is identical to the change of resolution should be shared.
Maybe have a function which get the new translation and resolution requested, and return the accepted translation and resolution, as they are set.

src/odemis/driver/orsay.py
```diff
@@ -1389,3 +1440,1247 @@ def terminate(self):
...
+            in IMAGEFORMAT_OPTIONS.
+        :return (tuple (int, int)): The actual image format set.
+        """
+        if value not in IMAGEFORMAT_OPTIONS:  # get the closest option available in IMAGEFORMAT_OPTIONS
+            value = min(IMAGEFORMAT_OPTIONS, key=lambda x: abs(x[0] - value[0]) + abs(x[1] - value[1]))
+        self.imageFormatUpdatedResolutionTranslation.clear()  # let it be known that image format is updating...
+        # ...resolution and translation
+
+        # get the old image format and determine the scale change
+        state = self._ionColumn.ImageSize.Actual
```
pieleric, line 2485: Just look at the old value of the VA, no need to ask the server. As the VA hasn't been updated, you can just do `prev_value = self.imageFormat.value`

src/odemis/driver/orsay.py
```diff
@@ -1389,3 +1440,1247 @@ def terminate(self):
...
+        self.brightness = model.FloatContinuous(1.0, unit="", range=(0, 1))
+        self._brightnessConnector = None
+        self.operatingMode = model.BooleanVA(False)
+        self._operatingModeConnector = None
+        self.imageFormat = model.TupleContinuous((1024, 1024), unit="px", range=[(512, 480), (1024, 1024)],
+                                                 setter=self._imageFormat_setter)
+        self.translation = model.TupleContinuous((0.0, 0.0), unit="px", range=[(-512.0, -512.0), (512.0, 512.0)],
+                                                 setter=self._translation_setter)
+        self.resolution = model.TupleContinuous((1024, 1024), unit="px", range=[(1, 1), (1024, 1024)],
+                                                setter=self._resolution_setter)
```
pieleric, line 2336: Please before this part, add a (long) comment how these 3 VAs are supposed to interact with each other.

src/odemis/driver/orsay.py
```diff
@@ -1389,3 +1440,1247 @@ def terminate(self):
...
+        self.imageFormatUpdatedResolutionTranslation = threading.Event()
+        self.imageFormatUpdatedResolutionTranslation.set()
+        self.updatingImageArea = threading.Lock()
```
pieleric, line 2339: Please explain a little how this event and lock are used.

src/odemis/driver/orsay.py
```diff
@@ -1389,3 +1440,1247 @@ def terminate(self):
...
+
+        :param (tuple (float, float)) value: Target translation of the area to image
+        :return (tuple (float, float)): The actual translation set
+
+        The translation VA marks the centre of the image area with respect to the centre of the field of view. This
+        setter transforms the coordinates of the centre of the image area to the coordinates of the top left corner of
+        the image area, which is the format the Orsay server takes. The setter also adjusts the size of the image area
+        (resolution VA) to prevent the new translation from placing part of the image area outside of the image format.
+        """
+        self.updatingImageArea.acquire()  # translation and resolution cannot be updated simultaniously
```
pieleric, line 2561: This raises a big flag "troubles ahead": you set explicitly a lock, and only release it into another function.
Firstly, if any line in this function fails, the lock will be there forever. Secondly, if the other function is never called, the lock will also stay there forever. For instance, is this function called if by coincidence we request the current value?

I think that one possibility would be to use `with self.updatingImageArea` around the VA setter functions. They would then wait for an event (imageFormatUpdatedResolutionTranslation?), which the parameter update function would send when it's completed.

I'm not completely sure it'd work out of the box though. Maybe it'd be simpler/safer to have a single function which is called to change the image area/size. Then this function would take care of blocking until the request has been applied (or a timeout happens).

src/odemis/driver/orsay.py
```diff
@@ -1389,3 +1440,1247 @@ def terminate(self):
...
+        # Subscribe to the parameter on the Orsay server
+        self._ionColumn.ImageSize.Subscribe(self._updateImageFormat)
+        self._ionColumn.ImageArea.Subscribe(self._updateTranslationResolution)
+
+        self._connectorList = [x for (x, _) in  # save only the names of the returned members
+                               inspect.getmembers(self,  # get all members of this FIB_source object
+                                                  lambda obj: type(obj) == OrsayParameterConnector  # get only the
+                                                  # OrsayParameterConnectors from all members of this FIB_source object
+                                                  )
+                               ]
```
pieleric, line 2460: Why don't you call update_VAs() from here, as it's done in the other components? Do the connectors already take care of setting the VA's value at init? It'd still require to update the translation, resolution VAs, right?

src/odemis/driver/orsay.py
```diff
@@ -1389,3 +1440,1247 @@ def terminate(self):
...
+    def _imageFormat_setter(self, value):
+        """
+        Setter of the imageFormat VA
+
+        :param (tuple (int, int)) value: The goal format of the image. Will be corrected to the closest available value
+            in IMAGEFORMAT_OPTIONS.
+        :return (tuple (int, int)): The actual image format set.
+        """
+        if value not in IMAGEFORMAT_OPTIONS:  # get the closest option available in IMAGEFORMAT_OPTIONS
+            value = min(IMAGEFORMAT_OPTIONS, key=lambda x: abs(x[0] - value[0]) + abs(x[1] - value[1]))
```
pieleric, line 2480: Just make this VA a EnumeratedVA. We don't need to pretend it can accept any value, and only select two.

pieleric (changes requested):
I didn't check too much the test cases.
The main thing for now is to fine tune the OrsayParameterConnector. I think it's possible to simplify it quite a bit.

src/odemis/driver/orsay.py
```diff
@@ -1389,3 +1440,1247 @@ def terminate(self):
...
+        self._va = va
+        self._va_type_name = va.__class__.__name__
+        if self._va_type_name.startswith("Tuple"):
+            self._va_is_tuple = True
+            self._va_value_type = type(self._va.value[0])
+        else:
+            self._va_is_tuple = False
+            self._va_value_type = type(self._va.value)
+        if not self._va.readonly:  # only overwrite the VA's setter if the VA is not read-only
+            self._va._setter = WeakMethod(self._update_parameter)
```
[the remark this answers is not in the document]
Helsloot: Subscribing actually cannot work here, because that causes callback loops.
Since _update_va is subscribed to the parameter, the va will change value when the parameter changes value. If then we subscribe _update_parameter to the va, then the va value will again be written to the parameter and so on. So here we need some way that the parameter will be updated if the va changes value, except for when the va changes value because the parameter changed value. That is now done through the setter. There are of course other ways, but I can't think of a nice one.

src/odemis/driver/orsay.py
```diff
@@ -1389,3 +1440,1247 @@ def terminate(self):
...
+        """Unsubscribes the VA from the parameter"""
+        if self._va is not None and self._parameters is not None:
+            for p in self._parameters:
+                p.Unsubscribe(self.update_VA)
+            self._parameters = None
+            self._attributeName = None
+            self._va._setter = WeakMethod(self._va._VigilantAttribute__default_setter)
+            self._va = None
+            self._va_type_name = None
+            self._conversion = None
```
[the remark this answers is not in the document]
Helsloot: I don't understand your comment. As you mentioned earlier, every other component reconnects by making new OrsayParameterConnector objects, so I have, as you suggested, removed the connect() method.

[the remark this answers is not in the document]
pieleric: I wrote this comment before realizing that the connect() method wasn't used.
Anyway, my comment still stand for the `_va_type_name` and `_conversion`. I don't think you need to discard the attributes when disconnecting. You just need to disconnect from the OrsayParameter and VA. The rest can stay as-is. They will get garbage collected by Python as soon as the connector is dereferenced.

src/odemis/driver/orsay.py
```diff
@@ -1389,3 +1440,1247 @@ def terminate(self):
...
+                                             "Contact Orsay Physics about this!" % (p.Name, self._maxpar[i].Name))
+                if upperbound is not None:  # if an upperbound is defined in the server
+                    new_range[1][i] = self._parameter_to_VA_value(upperbound)  # copy it to the va
+
+            if len(new_range[0]) == 1:
+                new_range = (new_range[0][0], new_range[1][0])
+            else:
+                new_range = (tuple(new_range[0]), tuple(new_range[1]))
+            # Set the range of the VA
+            self._va._value = new_range[0]
```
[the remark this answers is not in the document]
pieleric: Note that the Continuous VAs have a `.clip_on_range` attribute to handle it. If it's set to True (at init), then every time the range is updated, it'll automatically adjust the value, instead of complaining it's not within range. You could put this in the Connector init. Alternatively, leave this code as-is, but add a comment to explain why you put the wrong value and it's not a problem.

src/odemis/driver/orsay.py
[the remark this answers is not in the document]
pieleric: I understand now. In other cases, to avoid such loop, we unsubscribe, set the value, (re)subscribe. However, it's quite ugly, and admittedly here, the use of setter is more correct. We could try to have a setter for the setter in the VA class, but let's not do that now... so leave this code as-is.

src/odemis/driver/orsay.py
```diff
@@ -1389,3 +1440,1247 @@ def terminate(self):
...
+                                                             [self._ionColumn.CondensorSteerer1TiltX,
+                                                              self._ionColumn.CondensorSteerer1TiltY],
+                                                             minpar=[self._ionColumn.CondensorSteerer1TiltX_Minvalue,
+                                                                     self._ionColumn.CondensorSteerer1TiltY_Minvalue],
+                                                             maxpar=[self._ionColumn.CondensorSteerer1TiltX_Maxvalue,
+                                                                     self._ionColumn.CondensorSteerer1TiltY_Maxvalue])
+        self._orthogonalityConnector = OrsayParameterConnector(self.orthogonality,
+                                                               self._ionColumn.ObjectiveOrthogonality)
+        self._objectiveRotationOffsetConnector = OrsayParameterConnector(self.objectiveRotationOffset,
+                                                                         self._ionColumn.ObjectiveRotationOffset)
```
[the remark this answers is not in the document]
Helsloot: I double checked the objectiveRotationOffset and objectiveStageRotationOffset. They are both on a scale of -pi to pi on the Orsay server.

src/odemis/driver/orsay.py
Helsloot: Because I forgot... The connectors do indeed take care of it at initialisation, but the call is indeed needed for imageFormat, translation and resolution. I've changed it in all three FIB devices.

src/odemis/driver/orsay.py
```diff
@@ -1262,3 +1326,1229 @@ def terminate(self):
...
+                        upperbound = self._maxpar[i].Target
+                    if p.Max is not None and p.Max != upperbound:
+                        logging.warning("%s.Max and %s contain different, non-None values."
+                                        "Contact Orsay Physics about this!" % (p.Name, self._maxpar[i].Name))
+                else:
+                    upperbound = p.Max
+                if upperbound is not None:  # if an upperbound is defined in the server
+                    new_range[1][i] = self._parameter_to_VA_value(upperbound)  # copy it to the va
+
+            if len(new_range[0]) == 1:
```
pieleric, line 1460: Shouldn't it be `if self._va_is_tuple:`, the same as above. So that it's more explicit.

src/odemis/driver/orsay.py
```diff
@@ -1262,3 +1326,1229 @@ def terminate(self):
...
+        Converts a value of the parameter to its corresponding value for the VA. Uses the dictionary in self._conversion
+        or the factor in self._factor to do so.
+
+        :param (any) par_value: The value of the parameter. Its type depends on the parameter type. Often a string
+        :return (any): The corresponding value of the VA. Type depends on the VA type
+        """
+        if self._conversion is not None:  # if a conversion dict is supplied
+            for key, value in self._conversion.items():
+                if value == type(value)(par_value):
+                    return key
```
pieleric, line 1577: Please, add a `else` to the for loop and log a warning in case no value matched.

pieleric (changes requested):

pieleric (changes requested):

src/odemis/driver/orsay.py
```diff
@@ -1286,3 +1348,1228 @@ def terminate(self):
...
+        changing the VA's value, and if the reset was not succesful, the VA's value should not change.
+        """
+        if not value:
+            self._interlockOutChamber.Reset.Target = 0
+            logging.debug("Attempting to reset the HybridInterlockOutChamberVac interlock.")
+        return self.interlockOutChamberTriggered.value
+
+    def _setInterlockOutHVPS(self, value):
+        """
+        setter for interlockOutHVPSTriggered VA
```
Kleijwegt, line 1879: Can you spell out the abbreviation HVPS in the docstring?

src/odemis/driver/orsay.py
```diff
@@ -1286,3 +1348,1228 @@ def terminate(self):
...
+        + interlockOutSEDTriggered: BooleanVA
+        + columnPumpOn: BooleanVA
+        + gunPressure: FloatContinuous, readonly, unit="Pa", range=(0, 11e4)
+        + columnPressure: FloatContinuous, readonly, unit="Pa", range=(0, 11e4)
+        + compressedAirPressure: FloatContinuous, readonly, unit="Pa", range=(0, 5e6)
+        """
+
+        super().__init__(name, role, parent=parent, **kwargs)
+
+        self._columnPump = None
```
Kleijwegt, line 1636: As piel said in: https://github.com/delmic/odemis/pull/1808/files#r679277936
> Can you add the type as a comment?

install/linux/usr/share/odemis/hwtest/orsay.odm.yaml
```diff
@@ -52,3 +56,18 @@ Orsay: {
...
     init: {}
 }
+
+"FIB-vacuum": {
+    role: fib-vacuum,
+    init: {}
+}
+
+"FIB-source": {
+    role: fib-source,
```
Kleijwegt, line 66: @pieleric These roles are currently not defined in odemis/gui/model/__init__.py.
Do we want to make the roles already similar as for the FIB of Enzel?
And should this PR already these roles to __init__.py? (I don't think the latter necessarily)

pieleric: Most probably this role will go away, once we have extra wrappers. For now it's fine. It's known that the GUI cannot use this backend. We'll care about the GUI way later ;-)

pieleric (approved):

Kleijwegt (approved):

PR #1809, opened 2021-06-22
driver: fix Exception non-unpicklable

pieleric:
We had managed to make sure all the exception were pickable, which is
necessary for the backend to transfer them to the client (via Pyro4).
However, it turns out that an object can be pickable, but not
unpickable... and we had not cheked for it.
This is because the unpickling mechanism is essentially a new
initialisation. So the arguments passed to it must be the same as the
one passed to the super (which stores them for picking).

In addition, with Futures, when an exception is raised, it's sent
back to the client as "one-way" call. So the backend wouldn't know
that unpickling the exception failed. That mained that not only the call
failed, but the client thought it was taking a very long time... and
block.

=> Fix all the exceptions which had this problem.
=> Add test cases to detect such errors.

### DIFF
--- a/src/odemis/driver/pigcs.py
+++ b/src/odemis/driver/pigcs.py
@@ -136,10 +136,13 @@ class PIGCSError(Exception):
 
     def __init__(self, errno, *args, **kwargs):
         # Needed for pickling, cf https://bugs.python.org/issue1692335 (fixed in Python 3.3)
-        desc = self._errordict.get(errno, "Unknown error")
-        strerror = "PIGCS error %d: %s" % (errno, desc)
-        Exception.__init__(self, strerror, *args, **kwargs)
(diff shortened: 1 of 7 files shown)


tepals (commented):
I tested it on Fumo beta and it now immediately prints the error in the terminal. However it still hangs after reporting the error, and it doesn't stop after the timeout of 2 minutes. Is that how it's supposed to work?

pieleric:
The blocking is coming from something else. I've already noticed it for a long time, but actually thought it happens "from time to time". From here, it seems pretty clear that it happens only when the move fails. I've dug a little further and it's related to stopping the special daemon threads which handle the communication for the futures. However, I don't see what's special with receiving an exception instead of a normal result. It'll require more digging, but in Pyro. So it's fine for this PR.

tepals (approved):

PR #1810, opened 2021-06-25
driver actuator RotationActuator: add argument ref_on_move_count to override the default 5

pieleric:
Allows to set different (less frequent) referencing, even disable it.

### DIFF
--- a/src/odemis/driver/actuator.py
+++ b/src/odemis/driver/actuator.py
@@ -1916,7 +1916,7 @@ class RotationActuator(model.Actuator):
     """
 
     def __init__(self, name, role, dependencies, axis_name, cycle=2 * math.pi,
-                 ref_start=None, inverted=None, **kwargs):
+                 ref_start=None, ref_frequency=5, inverted=None, **kwargs):
         """
         name (string)
         role (string)
@@ -1926,6 +1926,8 @@ class RotationActuator(model.Actuator):
         ref_start (float or None): Value usually chosen close to reference switch from where to start referencing.
                                     Used to optimize runtime for referencing.
                                     If None, value will be 5% of value of cycle.
+        ref_frequency (None or 1 <= int): automatically re-reference the axis
+          after this many moves have been executed. Use None (null) to disable.
         """
         if inverted:
             raise ValueError("Axes shouldn't be inverted")
@@ -1934,11 +1936,13 @@ class RotationActuator(model.Actuator):
             raise ValueError("RotationActuator needs precisely one dependency")
 
         self._cycle = cycle
-        # counter to check when current position has overrun cycle
-        # and is close to zero again (pos and neg direction)
-        self._move_sum = 0
         # check when a specified number of rotations was performed
+        if ref_frequency is None or ref_frequency >= 1:
+            self._ref_frequency = ref_frequency
+        else:
+            raise ValueError("ref_on_move_count is %s, but must be >= 1 or None" % (ref_frequency,))
         self._move_num_total = 0
+
         axis, dep = list(dependencies.items())[0]
         self._axis = axis
         self._dependency = dep
@@ -2098,8 +2102,8 @@ class RotationActuator(model.Actuator):
         final_pos = cur_pos + move
         pass_ref = (cur_pos // self._cycle) != (final_pos // self._cycle)
 
-        # do referencing after i=5 moves or when passing the reference switch anyways
-        if pass_ref or self._move_num_total >= 5:
+        # Reference if passing by the reference switch, or after N moves (if not disabled)
+        if pass_ref or (self._ref_frequency is not None and self._move_num_total >= self._ref_frequency):
             # Move to pos close to ref switch
             move, cur_pos = self._findShortestMove(self._ref_start)
                 
--- a/src/odemis/driver/test/actuator_test.py
+++ b/src/odemis/driver/test/actuator_test.py
@@ -1110,13 +1110,10 @@ class TestRotationActuator(unittest.TestCase):
         # check absolute difference is smaller half the ustepsize
         self.assertLess(abs(self.dev_cycle.position.value[axis_name] - shift), self.dependency1._ustepsize[0] / 2.)
 
-    def test_offset_moveAbs(self):
+    def test_auto_ref_on_frequency(self):
         """
-        test if offset is correctly used
-        when accumulation of angles is overrunning 2pi (pos or neg) do referencing to zero
-        only works for cycle = 2pi
+        Test referencing is executed after a certain number of moves as specified in the reference frequency.
         """
-
         axis_name = list(self.dev_cycle.axes.keys())[0]
 
         f = self.dev_cycle.moveAbs({axis_name: 0.0})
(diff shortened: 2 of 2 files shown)


src/odemis/driver/actuator.py
```diff
@@ -1934,11 +1936,13 @@ def __init__(self, name, role, dependencies, axis_name, cycle=2 * math.pi,
...
 
         self._cycle = cycle
-        # counter to check when current position has overrun cycle
-        # and is close to zero again (pos and neg direction)
-        self._move_sum = 0
         # check when a specified number of rotations was performed
+        if ref_on_move_count is None or ref_on_move_count >= 1:
+            self._ref_on_move_count = ref_on_move_count
+        else:
+            raise ValueError("ref_on_move_count is %s, but must be >= 1 or None" % (ref_on_move_count,))
```
Sabrina85, line 1943: I don't think it is nice to have variable names in an error message. I suggest to say something like `"Referencing after %s moves requested, but must be ....".`

src/odemis/driver/actuator.py
```diff
@@ -1916,7 +1916,7 @@ class RotationActuator(model.Actuator):
     """
 
     def __init__(self, name, role, dependencies, axis_name, cycle=2 * math.pi,
-                 ref_start=None, inverted=None, **kwargs):
+                 ref_start=None, ref_on_move_count=5, inverted=None, **kwargs):
```
Sabrina85, line 1919: I think it is a good idea to add this parameter, but I do not particularly like the name as it is hard to understand what it should mean. It should indicate the frequency of the referencing procedure - not based on a time, but on the number of moves executed. So I suggest something like: `ref_freq_move` or better just `ref_frequency`.

src/odemis/driver/actuator.py
```diff
@@ -2098,8 +2102,8 @@ def _doMoveAbs(self, pos):
         final_pos = cur_pos + move
         pass_ref = (cur_pos // self._cycle) != (final_pos // self._cycle)
 
-        # do referencing after i=5 moves or when passing the reference switch anyways
-        if pass_ref or self._move_num_total >= 5:
+        # do referencing after N moves or when passing the reference switch anyways
```
Sabrina85, line 2105: Add in the comment `... or when passing the reference switch anyways (if referencing is not disabled)` or similar.

src/odemis/driver/test/actuator_test.py
```diff
@@ -1042,7 +1042,12 @@ def setUp(self):
                                           refproc="Standard",
                                           )
 
-        self.dev_cycle = RotationActuator("stage", "stage", {self.axis_name: self.dependency1}, self.axis, ref_start=1)
+        self.dev_cycle = RotationActuator("stage", "stage", 
+                                          dependencies={self.axis_name: self.dependency1},
+                                          axis_name=self.axis,
+                                          ref_start=1,
+                                          ref_on_move_count=7
+                                         )
```
Sabrina85, line 1050: Did you check the testcases where the referencing is checked? There were certain things checked if they are true after 5 moves, as we expected the referencing there. Now that it is not hardcoded anymore, we might want to update the tests as well. See `test_offset_moveAbs`.

Sabrina85 (changes requested):
Some minor comments.

src/odemis/driver/actuator.py
pieleric: I respectfully disagree! That's the argument name in the microscope file, it's much clearer to know which argument is causing the issue.

Sabrina85: ok, then it is fine.

src/odemis/driver/actuator.py
pieleric: I renamed it to `ref_frequency`

src/odemis/driver/test/actuator_test.py
pieleric: I hadn't noticed the test cases. Thanks for noticing it. I've now updated them.

pieleric:
@Sabrina85 , I've updated the test cases again. Do you want to give a (hopefully last) look?

src/odemis/driver/test/actuator_test.py
```diff
@@ -1117,30 +1122,50 @@ def test_offset_moveAbs(self):
...
+            f = self.dev_cycle.moveRel({axis_name: shift})
             f.result()
-            if i == 5:
-                # if referencing was correct position should not differ more than by
-                # half of the ustepsize from the wanted position
-                self.assertLess(abs(self.dev_cycle.position.value[axis_name] - new_pos),
-                                self.dependency1._ustepsize[0] / 2.)
+        exp_pos = (shift * 5) % (2 * math.pi)  # = pi/2
+
+        self.assertAlmostEqual(self.dev_cycle.position.value[axis_name], exp_pos,
```
Sabrina85, line 1148: You check here the position once reported by the wrapper class (RotationActuator) and once by the underlying HW (tmcm) right? I suggest to add a comment what these 2 tests do as it is not immediately obvious. 

`# Check positions reported by rotational axis is as expected after referencing and within the accuracy of the HW`
`# Check positions reported by linear axis (HW) is as expected after referencing and within the accuracy of the HW`

Q: Maybe I remember this incorrectly from our discussion, but I thought you mentioned that the controller does not know about "mistakes" in the physical position. That's why we regularly reference. If this is true, should we not remove the `_ustepsize` as accuracy measure? What do I miss?

src/odemis/driver/test/actuator_test.py
```diff
@@ -1117,30 +1122,50 @@ def test_offset_moveAbs(self):
...
+            f = self.dev_cycle.moveRel({axis_name:-shift})
             f.result()
-            if i == 5:
-                # if referencing was correct position should not differ more than by
-                # half of the ustepsize from the wanted position
-                self.assertLess(abs(self.dev_cycle.position.value[axis_name] - new_pos),
-                                self.dependency1._ustepsize[0] / 2.)
+
+        exp_pos = (-shift * 5) % (2 * math.pi)  #  = 3/2 pi ~= 4.712
+        exp_pos_dep = -shift
```
Sabrina85, line 1163: Add comment: 
`# Expected position of the linear axis (HW)` or something similar

Sabrina85 (changes requested):
The test themselves look fine! 

I think some more comments are needed to read them easier and quicker especially when you are not familiar with the code and the HW specifications. 

The doc-string of the test should be also updated, as it does not match the tests.

`Test referencing is executed after a certain number of moves as specified in the reference frequency. Test referencing is automatically executed when passing the reference switch clockwise and counterclockwise.`

What your test does not cover yet, is your new feature to disable the referencing completely. 

Regarding the behavior: `# Reference if passing by the reference switch, or after N moves (if not disabled)` means that there is currently no way to force the HW to only reference after a specified number (so if I want to only reference every 10 moves, but my moves are quite big, I will always end up referencing more often). Is this a realistic use case or will this never happen?

src/odemis/driver/test/actuator_test.py
```diff
@@ -1117,30 +1122,50 @@ def test_offset_moveAbs(self):
...
-            f = self.dev_cycle.moveRel({axis_name: new_pos})
+        # Check it runs referencing also after many small moves, by checking the time
+        durations = []
+        for i in range(self.ref_freq):
+            start_t = time.time()
+            f = self.dev_cycle.moveRelSync({axis_name: 0.1})
+            durations.append(time.time() - start_t)
+
+        # It should have 1 move which takes quite a lot longer (+1s)
+        dur_outlier = max(durations)
```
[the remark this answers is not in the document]
pieleric: I'd rather not call it "dur_ref_move", because I don't know if it's actually due to referencing. That's what we are testing, and hoping. I'll call it just "dur_longer".

src/odemis/driver/test/actuator_test.py
pieleric: The ustepsize accuracy has nothing to do with the checked behaviour of the wrapper. However, the TMCM (simulator) still will not exactly move to where it was requested, it'll go there within a ustepsize, so it's important to use the ustepsize.

pieleric:
> Regarding the behavior: # Reference if passing by the reference switch, or after N moves (if not disabled) means that there is currently no way to force the HW to only reference after a specified number (so if I want to only reference every 10 moves, but my moves are quite big, I will always end up referencing more often). Is this a realistic use case or will this never happen?
Indeed, it's not currently supported to forbid entirely the referencing. It might be one day useful, but for now it's not needed, and out of scope of this PR.

Sabrina85 (approved):
Very nice now :)

PR #1815, opened 2021-07-01
gui util updater: fix download on Python 3

pieleric:
The way to get the headers of the HTTPResponse have change in Python 3.
(It wasn't really anything official in Python 2).

Also modify a little bit the function to make it testable (on Linux)...
and add a test case.

### DIFF
--- a/src/odemis/gui/util/test/updater_test.py
+++ b/src/odemis/gui/util/test/updater_test.py
@@ -24,9 +24,11 @@ Odemis. If not, see http://www.gnu.org/licenses/.
 
 # test the functions of the gui.util.updater module
 from __future__ import division
-from odemis.gui.util import updater
+
 import logging
 import unittest
+import wx
+from odemis.gui.util import updater
 
 logging.getLogger().setLevel(logging.DEBUG)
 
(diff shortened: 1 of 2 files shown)


src/odemis/gui/util/test/updater_test.py
```diff
@@ -24,10 +24,13 @@
...
 # test the functions of the gui.util.updater module
 from __future__ import division
-from odemis.gui.util import updater
+
 import logging
+from odemis.gui.util import updater
 import unittest
 
+import wx
+
```
Kleijwegt, line 33: I don't know if that is a PEP8 rule but if I reorder the import I like to group and separate the odemis, external and future imports.
That isn't what happened with this PR though. Personally I would prefer something like::
```

import unittest
import wx
import logging

from odemis.gui.util import updater
```

tepals: https://www.python.org/dev/peps/pep-0008/#imports

> Imports should be grouped in the following order:
> 
>     1. Standard library imports.
>     2. Related third party imports.
>     3. Local application/library specific imports.

VictoriaDel (approved):

pieleric:
I've fixed the order of the imports, and a couple of issues when handling errors.

PR #1816, opened 2021-07-05
driver piezomotor: fix same speed for all axes

pieleric:
Each axis has a different speed assigned. However, the speed assumed for
the move was shared between all axes. This meant that some of the
time-out values were incorrectly computed.

Also report the speed in the .speed VA, which is usefule for some time
estimations in client code.

### DIFF
--- a/src/odemis/driver/piezomotor.py
+++ b/src/odemis/driver/piezomotor.py
@@ -97,7 +97,7 @@ class PMD401Bus(Actuator):
 
             Axis parameters:
                 axis_number (0 <= int <= 127): typically 1-3 for x-z, required
-                closed_loop (bool): True for closed loop (with encoder), default to True
+                closed_loop (bool): True for closed loop (with encoder), default to False
                 encoder_resolution (float): number of encoder counts per meter, default to 1.22e-9
                 motorstep_resolution (float): number of motor steps per m, default to 5e-6
                 range (tuple of float): in m, default to STROKE_RANGE
@@ -133,7 +133,7 @@ class PMD401Bus(Actuator):
                 closed_loop = axis_par['closed_loop']
             else:
                 closed_loop = False
-                logging.info("Axis mode (closed/open loop) not specified for axis %s. Assuming closed loop.", axis_name)
+                logging.info("Axis parameter \"closed_loop\" not specified for axis %s. Assuming open-loop.", axis_name)
             self._closed_loop[axis_name] = closed_loop
 
             if 'motorstep_resolution' in axis_par:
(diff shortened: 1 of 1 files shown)


src/odemis/driver/piezomotor.py
```diff
@@ -97,7 +97,7 @@ def __init__(self, name, role, port, axes, inverted=None, param_file=None, **kwa
 
             Axis parameters:
                 axis_number (0 <= int <= 127): typically 1-3 for x-z, required
-                closed_loop (bool): True for closed loop (with encoder), default to True
+                closed_loop (bool): True for closed loop (with encoder), default to False
```
tepals, line 100: I can't comment on unchanged lines, but I think the logging on line 136 is also written as if closed loop is the default. Even though on line 135 closed loop is set to false.

src/odemis/driver/piezomotor.py
```diff
@@ -108,6 +108,7 @@ def __init__(self, name, role, port, axes, inverted=None, param_file=None, **kwa
         """
         self._axis_map = {}  # axis name -> axis number used by controller
         self._closed_loop = {}  # axis name (str) -> bool (True if closed loop)
+        self._speed = {}  # axis name (str) -> speed in unit/s
```
tepals, line 111: Isn't this always [m/s]?

src/odemis/driver/piezomotor.py
```diff
@@ -511,17 +518,17 @@ def getEncoderPosition(self, axis):
         axname = [name for name, num in self._axis_map.items() if num == axis][0]  # get axis name from number
         return int(self._sendCommand(b'X%dE' % axis)) / self._counts_per_meter[axname]
 
-    def runRelTargetMove(self, axis, encoder_cnts):
+    def runRelTargetMove(self, axis, encoder_cnts, speed):
```
tepals, line 522: I would rename this to speed_steps, because in the rest of the module speed is used for speeds in [m/s] and speed_steps is used for speed in [steps/s]

philipwinklerdelmic (approved):

src/odemis/driver/piezomotor.py
pieleric: Good catch!

src/odemis/driver/piezomotor.py
pieleric: There is a parameter to specify the unit (defaults to "m"). So the speed will be in this unit. Admittedly, as in practice so far it's always in "m", I've left the .speed unit to be "m/s", but it's not formally correct. (just as incorrect as in the other drivers) ;-)

src/odemis/driver/piezomotor.py
pieleric: Well, the functions `runAbsTargetMove()` and `runMotorJog()` next to it already have the "speed" argument. So I think it's still clearer to use the same name.

tepals (approved):

PR #1817, opened 2021-07-06
driver andorcam3: improve temperature control

pieleric:
We sort of pretend that the temperature can be changed between -10 and
25°C, but in practice on some hardware, like the Zyla, we end up picking either -10 or 25.
With the latest SDK (3.15.30045.0), it's also possible to select 0°C.
=> At least, log that only some temperatures are actually working.

Also make sure we return the actual target temperature.

Also indicate in the automatic temperature logging the cooling status.

### DIFF
--- a/src/odemis/driver/andorcam3.py
+++ b/src/odemis/driver/andorcam3.py
@@ -179,6 +179,9 @@ LENGTH_FIELD_SIZE = 4
 CID_FIELD_SIZE = 4
 CID_TIMESTAMP = 1
 
+TEMP_NO_COOLING = 25  # °C, fake target temperature indicating that cooling should be disabled
+
+
 class AndorCam3(model.DigitalCamera):
     """
     Represents one Andor camera and provides all the basic interfaces typical of
@@ -509,8 +512,8 @@ class AndorCam3(model.DigitalCamera):
                 sn_paths = glob.glob('/sys/bus/usb/devices/*/serial')
                 for p in sn_paths:
                     try:
-                        f = open(p)
-                        snp = f.read().strip()
+                        with open(p) as f:
+                            snp = f.read().strip()
                     except (IOError, UnicodeDecodeError):
                         logging.debug("Failed to read %s, skipping device", p)
                     if snp == sn:
@@ -520,8 +523,8 @@ class AndorCam3(model.DigitalCamera):
                     return
                 # .../3-1.2/serial => .../3-1.2/3-1.2:1.0/ttyUSB1
                 sys_path = os.path.dirname(p)
-                f = open(sys_path + "/version")
-                usbv = float(f.read().strip())
+                with open(sys_path + "/version") as f:
+                    usbv = float(f.read().strip())
             except Exception:
                 logging.info("Failed to check USB version for device %s", self.name, exc_info=True)
                 return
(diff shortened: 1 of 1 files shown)


Kleijwegt:
I just had a really quick look  and didn't check the details.
But does such a change not also require some updates in the test file? (and maybe even the simulator?)

pieleric:
> But does such a change not also require some updates in the test file? (and maybe even the simulator?)

It doesn't really direct testable behaviour, so the test cases don't need to be changed. The target temperature accepted *might* be different from the requested one, but that depends on the hardware, so it'd be difficult to generically test.

However, you're right for the simulator. We use the simulator provided by Andor, so I assumed that if the code works on real hardware, it'd also work on the Andor simulator... but that was actually incorrect. The simulator doesn't provide temperature control the same way as the real hardware (conveniently!).
I've updated the code to handle such special "hardware".

src/odemis/driver/andorcam3.py
```diff
@@ -831,10 +834,10 @@ def _getTargetTemperatureRange(self):
                 tmps = [float(t) for t in tmps_str if tmps_str is not None]
                 if not self.isWritable(u"TemperatureControl"):
                     # Still allow 25°C to disable the cooling
-                    return min(tmps), 25
+                    return min(tmps), TEMP_NO_COOLING
                 else:
                     # TODO: just return a set
-                    return min(tmps), max(tmps)
+                    return min(tmps), max(tmps + [TEMP_NO_COOLING])
```
VictoriaDel, line 840: why is the TEMP_NO_COOLING added to tmps to get the maximum?

src/odemis/driver/andorcam3.py
```diff
@@ -858,15 +861,24 @@ def setTargetTemperature(self, temp):
...
+        else:  # Fallback to old (deprecated) property
+            if self.isWritable(u"TargetSensorTemperature"):
+                self.SetFloat(u"TargetSensorTemperature", temp)
+            else:
+                if temp != TEMP_NO_COOLING:
+                    temp = self.GetFloat(u"TargetSensorTemperature")
+                logging.warning("Temperature cannot be controlled, using %g °C", temp)
 
-        if temp >= 20:
+        if temp == TEMP_NO_COOLING:
```
VictoriaDel, line 881: Should the docstring of the method be updated as well, to be clear which values are expected?

VictoriaDel (approved):
It took me some time to understand what was the implementation about. I am just not familiar with this driver's code. Maybe a couple of comments would make it clearer :)

src/odemis/driver/andorcam3.py
pieleric: Actually, we want to return  TEMP_NO_COOLING, but in case the maximum temperature is higher than this, we'd return it. 

It should really never happen because the TEMP_NO_COOLING was picked to be higher than any useful cooling temperature. So I've now simplified the code, and just log a warning if this very unlikely situation happens.

src/odemis/driver/andorcam3.py
pieleric: The docstring is correct (it's a float representing a temperature). TEMP_NO_COOLING is just a special temperature which can be used to disabling the cooling.

VictoriaDel (approved):

PR #1818, opened 2021-07-06
driver actuator ConvertStage: add .speed and .referenced

pieleric:
When the underlying actuator provide these VAs, it's handy to have them.
Also fix the .reference() function to pass the right axes names.


tepals (approved):
small comment

philipwinklerdelmic (approved):

pieleric:
Fixed the comment, and also the fact that the speed wasn't updated after a metadata update.

tepals (approved):

PR #1821, opened 2021-07-12
Orsay hardware test results

Helsloot:
All changes made to code that was already merged based on the hardware tests

### DIFF
--- a/src/odemis/driver/orsay.py
+++ b/src/odemis/driver/orsay.py
@@ -419,7 +419,6 @@ class vacuumChamber(model.Actuator):
         • "vacuum": choices is {0 : "vented", 1 : "primary vacuum", 2 : "high vacuum"}
 
         Defines the following VA's and links them to the callbacks from the Orsay server:
-        • gateOpen (BooleanVA, set to True to open/start and False to close/stop)
         • position (VA, read-only, value is {"vacuum" : _chamber.VacuumStatus.Actual})
         • pressure (FloatContinuous, range=VACUUM_CHAMBER_PRESSURE_RNG, read-only, unit is "Pa",
                     value is _chamber.Pressure.Actual)
@@ -429,13 +428,11 @@ class vacuumChamber(model.Actuator):
 
         model.Actuator.__init__(self, name, role, parent=parent, axes=axes, **kwargs)
 
-        self._gate = None
         self._chamber = None
 
+        self.position = model.VigilantAttribute({"vacuum": 0}, readonly=True)
         self.pressure = model.FloatContinuous(VACUUM_CHAMBER_PRESSURE_RNG[0], range=VACUUM_CHAMBER_PRESSURE_RNG,
                                               readonly=True, unit="Pa")
-        self.gateOpen = model.BooleanVA(False, setter=self._changeGateOpen)
-        self.position = model.VigilantAttribute({"vacuum": 0}, readonly=True)
 
         self._vacuumStatusReached = threading.Event()
         self._vacuumStatusReached.set()
@@ -449,13 +446,10 @@ class vacuumChamber(model.Actuator):
         Defines direct pointers to server components and connects parameter callbacks for the Orsay server.
         Needs to be called after connection and reconnection to the server.
         """
-        self._gate = self.parent.datamodel.HybridPlatform.ValveP5
         self._chamber = self.parent.datamodel.HybridPlatform.AnalysisChamber
 
-        self._gate.ErrorState.Subscribe(self._updateErrorState)
         self._chamber.VacuumStatus.Subscribe(self._updatePosition)
         self._chamber.Pressure.Subscribe(self._updatePressure)
-        self._gate.IsOpen.Subscribe(self._updateGateOpen)
 
         self.update_VAs()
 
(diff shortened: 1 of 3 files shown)


src/odemis/driver/orsay.py
```diff
@@ -1021,86 +935,76 @@ def _updatePosition(self, parameter=None, attributeName="Actual"):
...
+                    logging.warning("Moving GIS while gas flow is on.")
+                self._armPositionReached.clear()  # to assure it waits
+                if goal["arm"]:
+                    logging.debug("Moving GIS to working position.")
+                    self._positionPar.Target = STR_WORK
+                else:
+                    logging.debug("Moving GIS to parking position.")
+                    self._positionPar.Target = STR_PARK
+        except KeyError:  # in case "arm" is nog present in goal
+            pass
```
pieleric, line 982: Avoid using exceptions for things are happen normally. One of the problem is that there might be other part of your code that fails due to KeyError, and this will hidden now.
Better just write `if "arm" in goal`...

src/odemis/driver/orsay.py
```diff
@@ -1021,86 +935,76 @@ def _updatePosition(self, parameter=None, attributeName="Actual"):
...
-            goal = False
-        if goal:
-            logging.debug("Starting gas flow.")
-            self._reservoirPar.Target = STR_OPEN
-        else:
-            logging.debug("Stopping gas flow.")
-            self._reservoirPar.Target = STR_CLOSED
-        return self._reservoirPar.Target == STR_OPEN
+        try:
+            if not goal["arm"] == self.position.value["arm"]:  # if the arm needs to move
```
pieleric, line 971: It's nicer to write `!=`

src/odemis/driver/orsay.py
```diff
@@ -1021,86 +935,76 @@ def _updatePosition(self, parameter=None, attributeName="Actual"):
...
+                    self._positionPar.Target = STR_PARK
+        except KeyError:  # in case "arm" is nog present in goal
+            pass
 
-    def _setArm(self, goal):
-        """
-        goal (bool): the goal state of the GIS position: (True: "engaged", False: "parked")
-        return (bool): the state the GIS position is in after finishing moving: (True: STR_WORK, False: STR_PARK)
+        try:
+            if not goal["injectingGas"] == self.position.value["injectingGas"]:  # if the gas flow needs to change
```
pieleric, line 985: Use `!=`

src/odemis/driver/orsay.py
```diff
@@ -1146,8 +1047,7 @@ def __init__(self, name, role, parent, **kwargs):
         Defines the following VA's and links them to the callbacks from the Orsay server:
         • temperatureTarget: FloatContinuous, unit="°C", range=(-273.15, 10^3), setter=_setTemperatureTarget
         • temperature: FloatContinuous, readonly, unit="°C", range=(-273.15, 10^3)
-        • temperatureRegulation: EnumeratedVA, unit=None, choices={0: "off", 1: "on", 2: "rush"},
-                                 setter=_setTemperatureRegulation
+        • temperatureRegulation: BooleanVA, True: "on", False: "off", setter=_setTemperatureRegulation
```
pieleric, line 1050: BTW, in docstrings, you normally shouldn't discuss about the internals. So no need to indicate the setters.

src/odemis/driver/orsay.py
```diff
@@ -1160,8 +1060,7 @@ def __init__(self, name, role, parent, **kwargs):
         self.temperatureTarget = model.FloatContinuous(0, unit="°C", range=(-273.15, 10 ** 3),
                                                        setter=self._setTemperatureTarget)
         self.temperature = model.FloatContinuous(0, unit="°C", range=(-273.15, 10 ** 3), readonly=True)
-        self.temperatureRegulation = model.VAEnumerated(0, unit=None, choices={0: "off", 1: "on", 2: "rush"},
-                                                        setter=self._setTemperatureRegulation)
+        self.temperatureRegulation = model.BooleanVA(False, setter=self._setTemperatureRegulation)
         self.age = model.FloatContinuous(0, unit="s", readonly=True, range=(0, 10 ** 12))
```
pieleric, line 1064: BTW, the normal way to write 10^12 is `1e12`.

src/odemis/driver/orsay.py
```diff
@@ -1356,23 +1242,12 @@ def _setTemperatureTarget(self, goal):
...
-            reg = True
-        elif goal == 1:
-            logging.debug("Turning on temperature regulation.")
-            reg = True
-        else:
-            logging.debug("Turning off temperature regulation.")
-        self._gis.RegulationOn.Target = reg
-        self._gis.RegulationRushOn.Target = rush
+        logging.debug("Turning temperature regulation %s." % "on" if goal else "off")
+        self._gis.RegulationOn.Target = goal
```
pieleric, line 1250: The setters *must* always return the new value that the VA will hold. So, here, you should return "goal".

src/odemis/driver/orsay.py
```diff
@@ -1280,33 +1178,21 @@ def _updateTemperatureRegulation(self, parameter=None, attributeName="Actual"):
...
         temperatureRegulation VA
         """
-        if parameter not in (self._gis.RegulationOn, self._gis.RegulationRushOn, None):
+        if parameter not in (self._gis.RegulationOn, None):
             raise ValueError("Incorrect parameter passed to _updateTemperatureRegulation. Parameter should be "
-                             "datamodel.HybridGIS.RegulationOn, datamodel.HybridGIS.RegulationRushOn, or None. "
+                             "datamodel.HybridGIS.RegulationOn, or None. "
                              "Parameter passed is %s." % parameter.Name)
         if not attributeName == "Actual":
             return
```
pieleric, line 1187: Please leave a comment explaining that RegulationRushOn is also available, but it actually doesn't work (yet), and we don't need it.

src/odemis/driver/test/orsay_test.py
```diff
@@ -91,7 +93,7 @@ def setUpClass(cls):
         if TEST_NOHW == 1:
             raise unittest.SkipTest("TEST_NOHW is set. No server to contact.")
         cls.oserver = orsay.OrsayComponent(**CONFIG_ORSAY)
-
+        sleep(1)
```
pieleric, line 96: Why? If that's needed to have the code work, put it in the `__init__()`. The client shouldn't have to know that there is special timing needed.

src/odemis/driver/test/orsay_test.py
```diff
@@ -138,26 +141,23 @@ def test_reconnection(self):
...
 
-        self.psus._valve.Target = orsay.VALVE_CLOSED
+        test_value = 27
+        self.gis_res.temperatureTarget.value = test_value
         sleep(1)
-        self.assertFalse(self.psus.power.value)
+        self.assertEqual(int(self.gis_res._temperaturePar.Target), test_value)
 
-        self.psus.power.value = True
+        self.gis_res.temperatureTarget.value = 0
```
pieleric, line 156: Sorry, I only realise now that this VA is incorrectly named. In the rest of Odemis, we call it `targetTemperature`. Please update the name.

pieleric (changes requested):

src/odemis/driver/orsay.py
```diff
@@ -948,13 +861,14 @@ class GIS(model.Actuator):
...
+        • "injectingGas": unit is None, choices is {True: "open", False: "closed"}
 
         Defines the following VA's and links them to the callbacks from the Orsay server:
-        • position (VA, read-only, value is {"arm" : _positionPar.Actual})
-        • injectingGas (BooleanVA, set to True to open/start the gas flow and False to close/stop the gas flow)
+        • position (VA, read-only, value is {"arm": _positionPar.Actual, "injectingGas": _reservoirPar.Actual})
         """
-        axes = {"arm": model.Axis(unit=None, choices={True: "engaged", False: "parked"})}
+        axes = {"arm": model.Axis(unit=None, choices={True: "engaged", False: "parked"}),
+                "injectingGas": model.Axis(unit=None, choices={True: "open", False: "closed"})}
```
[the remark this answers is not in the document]
Helsloot: In the orsay client it's called 'reservoir'. Would that work? Otherwise I'll ask Wilco if he has other suggestions.

src/odemis/driver/test/orsay_test.py
Helsloot: Riiight, my mistake. It's actually already there in __init__, so this sleep isn't needed at all. I'll delete it

src/odemis/driver/test/orsay_test.py
```diff
@@ -32,13 +32,17 @@
 if not TEST_NOHW == "sim":
     TEST_NOHW = TEST_NOHW == "1"  # make sure values other than "sim", 0 and 1 are converted to 0
 
+TEST_NOHW = "sim"
+
```
pieleric, line 36: Drop this. (I guess it was for debug)

src/odemis/driver/test/orsay_test.py
```diff
@@ -712,104 +655,153 @@ def test_errorstate(self):
...
+    def test_updatePositionReservoir(self):
         """
-        Check that the injectingGas VA is updated correctly and an exception is raised when the wrong parameter is passed
+        Check that the "reservoir" part of the position VA is updated correctly
         """
-        with self.assertRaises(ValueError):
-            self.gis._updateInjectingGas(self.datamodel.HybridPlatform.Cancel)
-
         if not TEST_NOHW == "sim":
             self.skipTest("TEST_NOHW is not set to sim, data isn't copied from Target to Actual outside of simulation.")
```
pieleric, line 692: I don't think that's the reason you skip this test with the real hardware. I guess the test case would actually pass. But it's not a great idea to open the GIS randomly during tests, right?

pieleric (changes requested):
I have only 2 very minor comments (in the test cases). Once processed, it should be ready to merge.

src/odemis/driver/test/orsay_test.py
Helsloot: No, this is actually the reason and the testcase would fail on the real system. This holds for test_updatePositionReservoir and test_updatePositionArm. They would fail outside of simulation, because I'm writing "MOVING" to the Target, which is a value you should in reality never write to Target. I'm doing so here to test what the position VA does when the Actual has value "MOVING". These two test cases are only there to check that the VA has the correct value for every possible value of the Orsay parameter Actual values. Testing whether moving the arm and opening/closing the reservoir work as they should is done by test_moveAbs. I've clarified on this in the docstring of the two testcases.

PR #1822, opened 2021-07-12
EK1 acquisition

VictoriaDel:
* Odemis supports EK imaging where the slit is close and focused on the center of the mirror
	* Creation of a new stream called AR Spectrum
	* Support for both horizontal and vertical binning
	* New metadata, MD_THETA_LIST, and new dimension "A" added in the MD_DIMS
	* HDF5 supports the extra dimension "A"
	* Support for csv export
	* Creation of 2 new Viewports and 2 new projections for displaying the new stream in Analysis Tab


src/odemis/gui/comp/viewport.py
```diff
@@ -1667,6 +1667,79 @@ def _on_new_data(self, data):
...
+                filtered = [(xv, yv) for xv, yv in zip(x, y) if not math.isnan(xv)]
+                return [f[0] for f in filtered], [f[1] for f in filtered]
+
+            angles, data = remove_nan(angle_range, data)
+
+            range_x = min(angles), max(angles)
+            if not self.hrange_lock.value or self.hrange.value is None:
+                # If there is a known period, use it
+                if hasattr(self._stream, "windowPeriod"):
+                    display_xrange = (-self._stream.windowPeriod.value, range_x[1])
```
pieleric, line 1713: That cannot happen. It's just for time dimension.

src/odemis/gui/cont/streams.py
```diff
@@ -661,16 +661,25 @@ def _link_resolution(self):
                 self._binva = getattr(self.stream, fn)
                 break
         else:
-            logging.warning("Stream has resolution VA but no binning/scale, "
-                            "so it will not be updated.")
-            return
+            if hasattr(self.stream, "spectrum_binning"):
```
pieleric, line 664: Also check that "angular_binning" is present.

src/odemis/gui/cont/streams.py
```diff
@@ -682,6 +691,8 @@ def _update_resolution(self, scale, crop=1.0):
         # if the stream is not playing, the hardware should take care of it
         if self.stream.is_active.value:
             return
+        if hasattr(self.stream, "spectrum_binning"):
```
pieleric, line 694: `and  hasattr(self.stream, "angular_binning")`

src/odemis/gui/cont/streams.py
```diff
@@ -692,7 +703,10 @@ def _on_resolution(self, res, crop=1.0):
         # if the stream is not playing, the hardware should take care of it
         if self.stream.is_active.value:
             return
-        scale = self._binva.value
+        if hasattr(self.stream, "spectrum_binning"):
```
pieleric, line 706: `and  hasattr(self.stream, "angular_binning")`

src/odemis/gui/util/img.py
```diff
@@ -1414,6 +1414,106 @@ def chronogram_to_export_data(proj, raw, vp=None):
...
+    else:  # tiff, png
+        spec = proj.image.value
+        if spec is None:
+            raise LookupError("No pixel selected to pick a theta")
+        angle_range, unit = spectrum.get_angle_range(spec)
+
+        def remove_nan(x, y):
+            """Removes NaN values from the angle list and updates the spectrum to have the same length"""
+            filtered = [(xv, yv) for xv, yv in zip(x, y) if not math.isnan(xv)]
+            return [f[0] for f in filtered], [f[1] for f in filtered]
```
pieleric, line 1460: I think that eventually this could be done in the stream, on the .calibrated for the StaticStream, and possibly in the projection of the SettingsStream. Then this code wouldn't been needed here or in the Viewport.

src/odemis/util/angleres.py
```diff
@@ -608,3 +608,68 @@ def _figure2data(figure):
 
     return image
 
+
+def ExtractThetaList(data):
+    """
+    Extracts the list of theta values so as to update the metadata MD_THETA_LIST.
+    Given the mirror parameters the theta list is calculated for ek imaging to be applied.
+    """
```
pieleric, line 616: Please add info about the input and output values. In particular, what are the minimum and maximum values returned, what is the length?

src/odemis/util/spectrum.py
```diff
@@ -176,6 +179,78 @@ def get_time_range(data):
...
+        dims = data.metadata.get(model.MD_DIMS, "CTZYX"[-data.ndim:])
+
+        if len(dims) == 3 and dims == "YXC" and data.shape[2] in (3, 4):  # RGB?
+            # This is a hack that works the same way as in get_wavelength_per_pixel
+            thetai = 0
+        else:
+            try:
+                thetai = dims.index("A")  # get index of dimension A
+            except ValueError:
+                raise ValueError("Dimension 'A' not in dimensions, so skip computing time list.")
```
pieleric, line 203: "so skip computing time list" -> "so skip computing angle list"

src/odemis/util/spectrum.py
```diff
@@ -176,6 +179,78 @@ def get_time_range(data):
...
+    """ Returns the list of theta values, if exists, in degrees.
+
+    :param data: (model.DataArray of shape A...): the DataArray with metadata
+        MD_THETA_LIST
+
+    :return: (list of numbers or None), unit string:
+    """
+
+    try:
+        return get_angle_per_pixel(data), "°"  # degree symbol
```
pieleric, line 192: report "rad" as unit.

src/odemis/util/spectrum.py
```diff
@@ -176,6 +179,78 @@ def get_time_range(data):
...
+            try:
+                thetai = dims.index("A")  # get index of dimension A
+            except ValueError:
+                raise ValueError("Dimension 'A' not in dimensions, so skip computing theta list.")
+
+        if len(thetal) != da.shape[thetai]:
+            raise ValueError("Length of theta list does not match length of theta data.")
+
+        # Converts the theta values from radians to degrees wherever there is not a NaN value
+        theta_list = [math.degrees(theta) if str(theta) != 'nan' else numpy.nan for theta in thetal]
```
pieleric, line 248: No, this function should return the THETA_LIST in rad. In the Viewport (GUI), you can convert from rad to degrees. You can probably just move this line to just after `get_angle_range()`.
BTW, math.degrees(NaN) returns NaN, so you can keep it simple:
theta_list = [math.degrees(theta) for theta in theta_list]

pieleric (changes requested):
I

src/odemis/gui/cont/streams.py
VictoriaDel: is it needed? If the "spectrum_binning" is present, the "angular binning" should be too.

pieleric: Right now, indeed it's redundant. However, it would make the code more future proof if we extend some stream with a "spectrum_binning" without adding "angular binning". I also find it less "weird" when reading the code, because otherwise asymmetrical (and it's not explained why).

src/odemis/acq/stream/_projection.py
```diff
@@ -1676,8 +1782,11 @@ def _computeSpec(self):
             return None
 
         x, y = self.stream.selected_pixel.value
+        # t represents either the selected time or selected angle pixel value
         if model.hasVA(self.stream, "selected_time"):
             t = numpy.searchsorted(self.stream._tl_px_values, self.stream.selected_time.value)
+        elif model.hasVA(self.stream, "selected_angle"):
+            t = numpy.searchsorted(self.stream._thetal_px_values, self.stream.selected_angle.value)
```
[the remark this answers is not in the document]
pieleric: Now that you went with the second option, this should work iff `_thetal_px_values` doesn't contain NaN.

src/odemis/acq/stream/_static.py
```diff
@@ -607,19 +609,33 @@ def __init__(self, name, image, *args, **kwargs):
...
+                # setter will automatically "snap" it to the closest existing timestamp
+                self.selected_time = model.FloatContinuous(self._tl_px_values[0],
+                                                           range=(min_t, max_t),
+                                                           unit=unit_t,
+                                                           setter=self._setTime)
+            else:  # MD_THETA_LIST is present
+                # cached list of theta stamps for each position in the theta dimension
+                theta_list, unit_theta = spectrum.get_angle_range(image)
+                # Converts radians to degrees and removes NaN values from the theta list
+                self._thetal_px_values = [math.degrees(theta) for theta in theta_list]
```
pieleric, line 630: I think it'd be simpler to not convert to degrees (yet). Only do the conversion at display in the GUI (if the unit is "rad").
Also, here you are not removing the NaNs anymore. It seems to work fine for min/max (they discard NaN).
However for searchsorted() it might not work. Now that the calibrated image has no NaN, so probably you need to remove NaN also here.

src/odemis/dataio/csv.py
```diff
@@ -132,6 +132,26 @@ def export(filename, data):
...
+
+        spectrum_range, unit_c = spectrum.get_spectrum_range(data)
+        if unit_c == "m":
+            spectrum_range = [s * 1e9 for s in spectrum_range]
+            unit_c = "nm"
+
+        angle_range, _ = spectrum.get_angle_range(data)
+        unit_a = "°"
+        angles = [math.degrees(theta) for theta in angle_range if not math.isnan(theta)]  # Convert radians to degrees
+
```
pieleric, line 146: Please add a `assert len(angles) == len(data)`.
Because of the `zip()`, if they are not the same length, the shortest length will be used and the rest of the data discarded. That should never happen!

src/odemis/dataio/csv.py
```diff
@@ -132,6 +132,26 @@ def export(filename, data):
...
+        logging.debug("Exporting angular spectrum data to CSV")
+
+        spectrum_range, unit_c = spectrum.get_spectrum_range(data)
+        if unit_c == "m":
+            spectrum_range = [s * 1e9 for s in spectrum_range]
+            unit_c = "nm"
+
+        angle_range, _ = spectrum.get_angle_range(data)
+        unit_a = "°"
+        angles = [math.degrees(theta) for theta in angle_range if not math.isnan(theta)]  # Convert radians to degrees
```
pieleric, line 145: I don't think you can (and need to) drop the NaN here. Normally the NaN should have already be discarded. In addition, if you drop the angles but not the data, it's going to not be matching.

src/odemis/dataio/csv.py
```diff
@@ -132,6 +132,26 @@ def export(filename, data):
...
+    elif data.metadata.get(model.MD_ACQ_TYPE, None) == model.MD_AT_EK:
+        logging.debug("Exporting angular spectrum data to CSV")
+
+        spectrum_range, unit_c = spectrum.get_spectrum_range(data)
+        if unit_c == "m":
+            spectrum_range = [s * 1e9 for s in spectrum_range]
+            unit_c = "nm"
+
+        angle_range, _ = spectrum.get_angle_range(data)
+        unit_a = "°"
```
pieleric, line 144: Similarly as above, add a `if unit_a == "rad": `... 
Because if it's px, it's not a good idea to convert to °.

src/odemis/gui/comp/viewport.py
```diff
@@ -1852,6 +1917,62 @@ def _on_new_data(self, data):
...
+        """
+        Called whenever the markline position of the overlay changes, to set
+        the selected wavelength & angle in the stream (so that the SinglePoint
+        projections are updated).
+        """
+        if self._stream:
+            if hasattr(self._stream, "selected_wavelength"):
+                self._stream.selected_wavelength.value = pos[0]
+            if hasattr(self._stream, "selected_angle"):
+                self._stream.selected_angle.value = pos[1]
```
pieleric, line 1940: If the data is now (still) in rad, then you'd need to convert from deg to radians. ie: `math.radians(pos[1])`. Possibly, you'd need to check that selected_angle.unit is "rad" (because if it's "px" it should not be touched).

src/odemis/gui/comp/viewport.py
```diff
@@ -1852,6 +1917,62 @@ def _on_new_data(self, data):
...
+        wx.CallAfter(self.bottom_legend.SetToolTip, "Wavelength")
+        wx.CallAfter(self.left_legend.SetToolTip, "Angle")
+
+    def connect_stream(self, projs):
+        super(AngularSpectrumViewport, self).connect_stream(projs)
+        stream, proj = self._stream, self._projection
+
+        if hasattr(stream, "selected_angle") and hasattr(stream, "selected_wavelength"):
+            pos = self.canvas.markline_overlay.val
+            pos.value = (stream.selected_wavelength.value, stream.selected_angle.value)
```
pieleric, line 1953: You'd need to convert from rad to deg here.

src/odemis/gui/util/img.py
```diff
@@ -1414,6 +1414,101 @@ def chronogram_to_export_data(proj, raw, vp=None):
...
+            raise LookupError("No pixel selected to pick a theta")
+        data.metadata[model.MD_ACQ_TYPE] = model.MD_AT_EK
+        return data
+    else:  # tiff, png
+        spec = proj.image.value
+        if spec is None:
+            raise LookupError("No pixel selected to pick a theta")
+        angle_range, _ = spectrum.get_angle_range(spec)
+        unit = "°"
+        angles = [math.degrees(angle) for angle in angle_range if not math.isnan(angle)]  # Converts radians to degrees
```
pieleric, line 1457: No need to drop the NaN, it should be already done.

src/odemis/gui/util/img.py
```diff
@@ -1414,6 +1414,101 @@ def chronogram_to_export_data(proj, raw, vp=None):
...
+        if data is None:
+            raise LookupError("No pixel selected to pick a theta")
+        data.metadata[model.MD_ACQ_TYPE] = model.MD_AT_EK
+        return data
+    else:  # tiff, png
+        spec = proj.image.value
+        if spec is None:
+            raise LookupError("No pixel selected to pick a theta")
+        angle_range, _ = spectrum.get_angle_range(spec)
+        unit = "°"
```
pieleric, line 1456: Only if unit is "rad".

src/odemis/gui/util/img.py
```diff
@@ -1494,6 +1589,40 @@ def temporal_spectrum_to_export_data(proj, raw):
...
+        data.metadata[model.MD_ACQ_TYPE] = model.MD_AT_EK
+        return data
+    else:  # tiff, png
+        spec = proj.image.value
+        if spec is None:
+            raise LookupError("No pixel selected to pick an angular-spectrum")
+        spectrum_range, wl_unit = spectrum.get_spectrum_range(spec)
+        angle_range, _ = spectrum.get_angle_range(spec)
+        angle_unit = "°"
+        angle_range = [math.degrees(angle) for angle in angle_range if not math.isnan(angle)]  # Convert radians to degrees
```
pieleric, line 1613: Don't drop the NaN.

src/odemis/gui/comp/viewport.py
```diff
@@ -1852,6 +1917,62 @@ def _on_new_data(self, data):
...
+
+    def _on_new_data(self, data):
+        if data is not None and data.size:
+            wl, unit_x = spectrum.get_spectrum_range(data)
+            spectrum_range = (min(wl), max(wl))
+
+            angles, _ = spectrum.get_angle_range(data)
+            unit_y = "°"
+            angles = [math.degrees(angle) for angle in angles if not math.isnan(angle)]  # Converts radians to degrees
```
pieleric, line 1962: No need to drop the NaN anyamore, as it's already done.

src/odemis/gui/comp/viewport.py
```diff
@@ -1852,6 +1917,62 @@ def _on_new_data(self, data):
...
+            pos = self.canvas.markline_overlay.val
+            pos.value = (stream.selected_wavelength.value, stream.selected_angle.value)
+
+    def _on_new_data(self, data):
+        if data is not None and data.size:
+            wl, unit_x = spectrum.get_spectrum_range(data)
+            spectrum_range = (min(wl), max(wl))
+
+            angles, _ = spectrum.get_angle_range(data)
+            unit_y = "°"
```
pieleric, line 1961: Only convert to deg if unit == "rad".

src/odemis/gui/comp/viewport.py
```diff
@@ -1667,6 +1667,71 @@ def _on_new_data(self, data):
...
+            if hasattr(self._stream, "windowPeriod"):
+                self.hrange_lock.value = True
+                self.hrange.value = (-self._stream.windowPeriod.value, 0)
+
+    def _on_new_data(self, data):
+        if data is not None and data.size:
+            angle_range, _ = spectrum.get_angle_range(data)
+            unit_x = "°"
+            # Converts radians to degrees & removes NaN values
+            angles = [math.degrees(angle) for angle in angle_range if not math.isnan(angle)]
```
pieleric, line 1703: Only convert if unit is deg, and don't drop the NaN.

pieleric (changes requested):

pieleric:
The entire PR is now in the "EK1 Alignment" PR.

PR #1825, opened 2021-07-14
fastem overview tab: show scale and scaled pixelsize

philipwinklerdelmic:
Show a combobox control to select the scale (and indirectly the resolution). Display the pixelsize, which is adjusted with the scale.

### DIFF
--- a/src/odemis/acq/stream/_live.py
+++ b/src/odemis/acq/stream/_live.py
@@ -534,16 +534,17 @@ class FastEMSEMStream(SEMStream):
     """
 
     def __init__(self, name, detector, dataflow, emitter, blanker=None, **kwargs):
-        super().__init__(name, detector, dataflow, emitter, blanker=None, **kwargs)
-        self.pixelSize = model.VigilantAttribute(0, unit="m", readonly=True)
-        emitter.pixelSize.subscribe(self._on_pxsize, init=True)
+        super().__init__(name, detector, dataflow, emitter, blanker=blanker, **kwargs)
+        pxs = (emitter.pixelSize.value[0] * emitter.scale.value[0],
+               emitter.pixelSize.value[1] * emitter.scale.value[1])
+        self.pixelSize = model.VigilantAttribute(pxs, unit="m", readonly=True)
+        emitter.pixelSize.subscribe(self._on_pxsize)
         emitter.scale.subscribe(self._on_pxsize, init=True)
 
     def _on_pxsize(self, _):
         pxs = (self.emitter.pixelSize.value[0] * self.emitter.scale.value[0],
                self.emitter.pixelSize.value[1] * self.emitter.scale.value[1])
         self.pixelSize._set_value(pxs, force_write=True)
-        self.pixelSize.notify(pxs)
 
 
 class SpotSEMStream(LiveStream):
--- a/src/odemis/acq/stream/_live.py
+++ b/src/odemis/acq/stream/_live.py
@@ -528,6 +528,24 @@ class AlignedSEMStream(SEMStream):
 #         super(AlignedSEMStream, self)._onActive(active)
 
 
+class FastEMSEMStream(SEMStream):
+    """
+    SEM stream with special pixelsize VA (driver value is adjusted with scale).
+    """
+
+    def __init__(self, name, detector, dataflow, emitter, blanker=None, **kwargs):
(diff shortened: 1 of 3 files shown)


src/odemis/driver/xt_client.py
```diff
@@ -1156,6 +1156,7 @@ def __init__(self, name, role, parent, hfw_nomag, has_detector=False, **kwargs):
             pxs = (fov / self._shape[0],
                    fov / self._shape[0])
             self.pixelSize = model.VigilantAttribute(pxs, unit="m", readonly=True)
+            self.pixelSizeScaled = model.VigilantAttribute(pxs, unit="m", readonly=True)
```
pieleric, line 1159: No, please, do not add a VA to the driver, just because the GUI would like to display something slightly differently.
I think the simplest is create subclass of SEMStream, with a .pixelSize VA, which listens to ebeam.pixelSize and ebeam.scale.
Alternatively, you could add some code to StreamController to display such label... but it's a little tricky to display it only in this specific case.
Another alternative would be to simply add a label from the tab controller into the stream. We've done something similar in the SPARC2 alignment tab. See the function _addMoIEntries().

Moreover, it should be labelled as "pixel size". That's easy: in gui/conf/data.py, add an entry for this VA and specify the "label".

src/odemis/gui/conf/data.py
```diff
@@ -648,6 +648,14 @@
             },
         },
     },
+    "mbsem": {
+        "e-beam": {
+            "scale": {
```
pieleric, line 653: Please add comment, why is that necessary? What happens normally, if you don't add these lines?

pieleric (changes requested):

philipwinklerdelmic:
pushed changes

src/odemis/acq/stream/_live.py
```diff
@@ -528,6 +528,24 @@ def _DoPrepare(self):
...
 
 
+class FastEMSEMStream(SEMStream):
+    """
+    SEM stream with special pixelsize VA (driver value is adjusted with scale).
+    """
+
+    def __init__(self, name, detector, dataflow, emitter, blanker=None, **kwargs):
+        super().__init__(name, detector, dataflow, emitter, blanker=None, **kwargs)
+        self.pixelSize = model.VigilantAttribute(0, unit="m", readonly=True)
```
pieleric, line 538: pixelSize is normally a tuple. It'll be confusing to have in one place one VA which doesn't follow the same convention. So I'd suggest to either make it a tuple, and change the way the GUI displays it, or rename it (maybe "pixelSize1D", pixelSizeOne, pixelSizeSingle, or pixelSizeUI). You can then change the "label" in the conf.data.

src/odemis/acq/stream/_live.py
```diff
@@ -528,6 +528,24 @@ def _DoPrepare(self):
...
+        emitter.pixelSize.subscribe(self._on_pxsize, init=True)
+        emitter.scale.subscribe(self._on_pxsize, init=True)
+
+    def _on_pxsize(self, _):
+        pxs = (self.emitter.pixelSize.value[0] * self.emitter.scale.value[0],
+               self.emitter.pixelSize.value[1] * self.emitter.scale.value[1])
+        self.pixelSize._set_value(pxs, force_write=True)
+        self.pixelSize.notify(pxs)
```
pieleric, line 546: No need for this line. The previous one already takes care of notifying.

pieleric (changes requested):

philipwinklerdelmic:
Pushed again. The pixelsize is actually also a tuple here, that was my mistake

src/odemis/gui/conf/data.py
```diff
@@ -691,6 +701,14 @@
                 "tooltip": u"Current temperature of the spectrometer CCD",
             }),
         )),
+    stream.FastEMSEMStream:
+        OrderedDict((
+            # Display the adjusted pixelsize (pixelsize * scale) with the hw VAs.
+            ("pixelSize", {
+                "label": "Pixel size",
```
pieleric, line 708: Could you drop this line? It's not necessary, as it should be the default label. Or, if I'm guessing wrong, please add a comment explaining the label otherwise.

pieleric (changes requested):

PR #1827, opened 2021-07-16
cli: show angles also in degrees, and support movement in degrees

pieleric:
Some users find that thinking in degrees is easier than radians... so
let's accomodate them.

### DIFF
--- a/src/odemis/cli/main.py
+++ b/src/odemis/cli/main.py
@@ -24,14 +24,14 @@ Odemis. If not, see http://www.gnu.org/licenses/.
 
 from __future__ import division, print_function
 
-from past.builtins import basestring, unicode
-from builtins import str
 import argparse
+from builtins import str
 import codecs
 import collections
 import importlib
 import inspect
 import logging
+import math
 import numbers
 from odemis import model, dataio, util
 import odemis
@@ -39,6 +39,7 @@ from odemis.util import units, inspect_getmembers
 from odemis.util.conversion import convert_to_object
 from odemis.util.driver import BACKEND_RUNNING, \
     BACKEND_DEAD, BACKEND_STOPPED, get_backend_status, BACKEND_STARTING
+from past.builtins import basestring, unicode
 import sys
 import threading
 
@@ -282,7 +283,7 @@ def print_events(component, pretty):
         print_event(name, value, pretty)
 
 
-def print_vattribute(name, va, pretty):
+def print_vattribute(component, name, va, pretty):
     if va.unit:
         if pretty:
             unit = u" (unit: %s)" % va.unit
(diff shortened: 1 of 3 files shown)


src/odemis/cli/main.py
```diff
@@ -24,21 +24,22 @@
 
 from __future__ import division, print_function
 
-from past.builtins import basestring, unicode
-from builtins import str
 import argparse
+from builtins import str
```
philipwinklerdelmic, line 28: Do we still need the builtins/past.builtins imports?

philipwinklerdelmic (approved):
Looks good.

src/odemis/cli/main.py
```diff
@@ -331,6 +332,32 @@ def print_vattribute(name, va, pretty):
                 logging.info("Failed to convert %s to component names")
                 # Leave the value as-is
 
+        # Convert to nicer unit for user
+        if va.unit and va.unit == "rad" and isinstance(val, numbers.Real):
+            try:
+                val_converted = u" = %s°" % (math.degrees(val),)
```
Mahmood-B, line 338: I'm wondering. Why actually do you make it unicode? Why not string?

src/odemis/cli/main.py
```diff
@@ -282,7 +283,7 @@ def print_events(component, pretty):
         print_event(name, value, pretty)
 
 
-def print_vattribute(name, va, pretty):
+def print_vattribute(component, name, va, pretty):
```
Mahmood-B, line 286: docstring?

src/odemis/cli/main.py
```diff
@@ -331,6 +332,32 @@ def print_vattribute(name, va, pretty):
...
+        # For position, it's trickier, as the unit is on .axes
+        if (name == "position" and isinstance(va.value, dict) and
+            hasattr(component, "axes") and isinstance(component.axes, dict)
+           ):
+            pos_deg = {}
+            for an, pos in va.value.items():
+                try:
+                    axis_def = component.axes[an]
+                except KeyError:
+                    continue
```
Mahmood-B, line 354: It's better to add a statement showing what axis missing: `logging.error("The axis %s is missing" %an)`. 
Such messages are not for the user, but for the developer if wants to change the code in the future. This was proven to me after adapting enzel code for meteor.

src/odemis/cli/main.py
pieleric: Not really, but to remove them, we need to update the code to Python 3-only, which I'm not going to do in this PR.
I'll put the lines back to where they were.

src/odemis/cli/main.py
pieleric: It's only for Python 2. In Python 3, all strings are unicode. We don't really need this anymore, as we now only support Python 3, but to follow the style of the rest of the code, I keep unicode here too. One day, someone will convert this code to Python 3 only, but that day is not today ;-)

src/odemis/cli/main.py
pieleric: Adding log statements affect the user, so if it's not for the user, I'd suggest to pass the information a different way. That said, I guess it's fine to tell the user that .position has axes not in .axes. I'll add some warnings.

Mahmood-B (approved):

PR #1828, opened 2021-07-16
Fix Smaract 5DOF stage timeout issue internally

lazem:
A timeout error can happen if the movement of some axes pulled others slightly out of their position, and because they are not expected to move by the kinematics (e.g. the x-positioners in  case of a pure y movement), they go back to their position very slowly, causing the movement event to time out. As a fix for this, the same movement will be tried again.

### DIFF
--- a/src/odemis/driver/smaract.py
+++ b/src/odemis/driver/smaract.py
@@ -1249,8 +1249,6 @@ class MC_5DOF(model.Actuator):
 
         # will take care of executing axis move asynchronously
         self._executor = CancellableThreadPoolExecutor(1)  # one task at a time
-        # A boolean to retry the movement in case of timeout error
-        self.movement_retired = False
 
         # Reference tilted positioners towards the negative position
         # Normally this is not needed. Some old version of the controller needed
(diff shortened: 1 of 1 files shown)


src/odemis/driver/smaract.py
```diff
@@ -1718,8 +1721,18 @@ def _doMoveAbs(self, future, pos):
...
+                # A timeout error can happen if the movement of some axes pulled others slightly out of their
+                # position, and because they are not expected to move by the kinematics (e.g. the x-positioners in
+                # case of a pure y movement), they go back to their position very slowly, causing the movement event
+                # to time out. As a fix for this, the same movement will be tried again.
+                if self.movement_retired:
+                    self.movement_retired = False
+                    logging.error("Move timed out after %g s: %s", max_dur, ex)
+                    raise TimeoutError("Move timed out after %g s" % (max_dur,))
+                else:
+                    self.movement_retired = True
```
pieleric, line 1733: Instead of using `self.movement_retired`, an attribute of the class, I think it should work as well by passing an argument `retrial=True` to `_doMoveAbs()`. By default it's set to False, but in case of timeout, you'd call explicitly `self._doMoveAbs(future, pos, retrial=True)`.
This way, there is no need for a dedicated class attribute.

pieleric (changes requested):
This is the first time that I see an explanation for this issue that makes sense. Thanks for finding this out and explaining clearly!

src/odemis/driver/smaract.py
lazem: I put it as a class attribute to run it once and avoid a possible infinite regress if the timeout error stayed for long. Or do you think it'd probably be fine even if it's called multiple times it will eventually arrive?

pieleric: I agree with you that it shouldn't be called multiple times. My suggestion wouldn't change the behaviour. If `retrial=True`, the timeout is immediately raised, as you are doing now.

lazem: Indeed, thanks for pointing that out.

PR #1833, opened 2021-07-21
fastem gui: add alignment button to acquisition tab

philipwinklerdelmic:


philipwinklerdelmic:
Updated to show label on the right

src/odemis/gui/model/__init__.py
```diff
@@ -1121,6 +1121,9 @@ def __init__(self, microscope):
         # Overview streams
         self.overview_streams = model.VigilantAttribute({})  # dict: int --> stream or None
 
+        # Alignment status
+        self.is_aligned = model.BooleanVA(False)
```
pieleric, line 1125: Eventually, this should it automatically be reset whenever the user pumps the chamber. Even turning off/on the e-beam? Other cases? Do you have already plans to add these? Maybe you can already do some of it in this PR.

src/odemis/gui/cont/acquisition.py
```diff
@@ -2141,3 +2147,47 @@ def _on_ac_done(self, future):
...
+
+        # Change button to "cancel", show progress bar
+        self._panel.align_gauge_progress.Show()
+        # Label is displayed on the right, but during the acquisition, we want the gauge bar to occupy the
+        # full available space --> use a panel instead of a spacer which can be hidden (spacers cannot) to
+        # make it look nice.
+        self._panel.align_spacer_panel.Hide()
+        self._panel.align_lbl_gauge.Hide()
+        self._panel.btn_align.SetLabel("Cancel")
+        self._panel.Parent.Layout()
```
pieleric, line 2176: Are you sure that `self._panel.Layout()` is not sufficient? I don't get why you have you force the layout of the parent (but I know that wxWidgets is sometimes strange about this).

src/odemis/gui/cont/acquisition.py
```diff
@@ -2141,3 +2147,47 @@ def _on_ac_done(self, future):
...
+        self._panel = panel
+
+        self._panel.align_gauge_progress.Hide()
+        panel.btn_align.Bind(wx.EVT_BUTTON, self._on_btn_align)
+
+        self._is_aligning = False
+
+    def _on_btn_align(self, evt):
+        if self._is_aligning:
+            logging.error("Cancelling alignment currently not supported.")
```
pieleric, line 2164: Let's add a TODO. (or otherwise, don't call this button "Cancel")

src/odemis/gui/cont/acquisition.py
```diff
@@ -2141,3 +2147,47 @@ def _on_ac_done(self, future):
...
+            return
+        self._is_aligning = True
```
pieleric, line 2166: Shouldn't more things be disabled in the GUI? I guess it wouldn't be a good idea to move the stage, or change the tab and start an overview acquisition during alignment. => Set the `is_acquiring` VA of the MainGUIData?

pieleric (changes requested):

philipwinklerdelmic:
Pushed changes.

pieleric (approved):

PR #1834, opened 2021-07-21
spot_grid.py: options to select magnification

philipwinklerdelmic:
* New cli option --magnification
* If no magnification is specified, try to use lens magnification
* If lens cannot be found, fall back to 50
* In case a magnification is specified which is NOT the lens specification, a warning will be logged (but we will still use the cli-specified value)

### DIFF
--- a/scripts/spot-grid.py
+++ b/scripts/spot-grid.py
@@ -27,7 +27,7 @@ from odemis.driver import ueye
 from odemis.util.driver import get_backend_status, BACKEND_RUNNING
 
 MAX_WIDTH = 2000  # px
-PIXEL_SIZE_SAMPLE_PLANE = 3.45
+PIXEL_SIZE_SAMPLE_PLANE = 3.45e-6  # m
 DEFAULT_MAGNIFICATION = 50
 
 
@@ -42,7 +42,7 @@ class VideoDisplayerGrid(VideoDisplayer):
         """
         Displays the window on the screen
         size (2-tuple int,int): X and Y size of the window at initialisation
-        px_size (float): pixelsize in µm
+        px_size (float): pixelsize in m
         Note that the size of the window automatically adapts afterwards to the
         coming pictures
         """
@@ -223,7 +223,7 @@ def live_display(ccd, dataflow, kill_ccd=True, gridsize=None, pxsize=PIXEL_SIZE_
     dataflow_name: name of the dataflow to access
     kill_ccd: True if it is required to terminate the ccd after closing the window
     gridsize: size of the grid of spots.
-    px_size (float): pixelsize in µm
+    px_size (float): pixelsize in m
     """
     # create a window
     window = VideoDisplayerGrid("Live from %s.%s" % (ccd.role, "data"), ccd.resolution.value, gridsize, pxsize)
@@ -264,7 +264,7 @@ def main(args):
     parser.add_argument("--gridsize", dest="gridsize", nargs=2, metavar="<gridsize>", type=int, default=None,
                         help="size of the grid of spots in x y, default 8 8")
     parser.add_argument("--magnification", dest="magnification", type=float,
-                        help="magnification in µm (typically 40 or 50)")
+                        help="magnification (typically 40 or 50)")
     parser.add_argument("--log-level", dest="loglev", metavar="<level>", type=int, choices=[0, 1, 2],
                         default=0, help="set verbosity level (0-2, default = 0)")
     options = parser.parse_args(args[1:])
@@ -289,14 +289,14 @@ def main(args):
     if options.magnification:
         magnification = options.magnification
         if lens_mag and lens_mag != magnification:
-            logging.warning("Requested magnification %s µm differs from lens magnification %s µm.",
+            logging.warning("Requested magnification %s differs from lens magnification %s.",
                             magnification, lens_mag)
     elif lens_mag:
         magnification = lens_mag
-        logging.debug("No magnification specified, using lens magnification %s µm.", lens_mag)
+        logging.debug("No magnification specified, using lens magnification %s.", lens_mag)
     else:
         magnification = DEFAULT_MAGNIFICATION
-        logging.warning("No magnification specified, falling back to %s µm", magnification)
+        logging.warning("No magnification specified, falling back to %s.", magnification)
     pxsize = PIXEL_SIZE_SAMPLE_PLANE / magnification
 
     if options.filename:
--- a/scripts/spot-grid.py
+++ b/scripts/spot-grid.py
@@ -29,6 +29,7 @@ from odemis.util.driver import get_backend_status, BACKEND_RUNNING
 MAX_WIDTH = 2000  # px
 PIXEL_SIZE_SAMPLE_PLANE = 3.45e-6  # m
 DEFAULT_MAGNIFICATION = 50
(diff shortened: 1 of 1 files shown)


scripts/spot-grid.py
```diff
@@ -27,7 +27,8 @@
 from odemis.util.driver import get_backend_status, BACKEND_RUNNING
 
 MAX_WIDTH = 2000  # px
-PIXEL_SIZE_UM = 3.45 / 50.
+PIXEL_SIZE_SAMPLE_PLANE = 3.45
```
pieleric, line 30: Please add a unit as comment (µm?). Possibly, seeing the rest of the changes, it'd be better to store in m here too. (3.45e-6)

scripts/spot-grid.py
```diff
@@ -211,16 +216,17 @@ def start_generate(self):
...
 
-def live_display(ccd, dataflow, kill_ccd=True, gridsize=None):
+def live_display(ccd, dataflow, kill_ccd=True, gridsize=None, pxsize=PIXEL_SIZE_SAMPLE_PLANE / DEFAULT_MAGNIFICATION):
     """
     Acquire an image from one (or more) dataflow and display it with a spot grid overlay.
     ccd: a camera object
     dataflow_name: name of the dataflow to access
     kill_ccd: True if it is required to terminate the ccd after closing the window
     gridsize: size of the grid of spots.
+    px_size (float): pixelsize in µm
```
pieleric, line 226: No, please, follow the convention of values in meters. You can convert to µm in the display.

scripts/spot-grid.py
```diff
@@ -270,21 +278,42 @@ def main(args):
...
+        lens = model.getComponent(role="lens")
+        lens_mag = lens.magnification.value
+    except Exception as ex:
+        logging.debug("Failed to read magnification from lens, ex: %s", ex)
+        lens_mag = None
+
+    if options.magnification:
+        magnification = options.magnification
+        if lens_mag and lens_mag != magnification:
+            logging.warning("Requested magnification %s µm differs from lens magnification %s µm.",
```
pieleric, line 292: Drop the "µm" from the message.

scripts/spot-grid.py
```diff
@@ -270,21 +278,42 @@ def main(args):
...
+                            magnification, lens_mag)
+    elif lens_mag:
+        magnification = lens_mag
+        logging.debug("No magnification specified, using lens magnification %s µm.", lens_mag)
+    else:
+        magnification = DEFAULT_MAGNIFICATION
+        logging.warning("No magnification specified, falling back to %s µm", magnification)
```
pieleric, line 299: Drop µm from the messages.

scripts/spot-grid.py
```diff
@@ -36,14 +37,16 @@ class VideoDisplayerGrid(VideoDisplayer):
     It should be pretty much platform independent.
     """
 
-    def __init__(self, title="Live image", size=(640, 480), gridsize=None):
+    def __init__(self, title="Live image", size=(640, 480), gridsize=None,
+                 pxsize=PIXEL_SIZE_SAMPLE_PLANE / DEFAULT_MAGNIFICATION):
         """
         Displays the window on the screen
         size (2-tuple int,int): X and Y size of the window at initialisation
+        px_size (float): pixelsize in µm
```
pieleric, line 45: in "m".

pieleric (changes requested):

philipwinklerdelmic:
Pushed changes.

scripts/spot-grid.py
```diff
@@ -36,14 +37,16 @@ class VideoDisplayerGrid(VideoDisplayer):
     It should be pretty much platform independent.
     """
 
-    def __init__(self, title="Live image", size=(640, 480), gridsize=None):
+    def __init__(self, title="Live image", size=(640, 480), gridsize=None,
+                 pxsize=PIXEL_SIZE_SAMPLE_PLANE / DEFAULT_MAGNIFICATION):
```
tepals, line 41: I'm not a fan of having calculations as default arguments, it is a bit hard to read. Why not create a variable PIXEL_SIZE at the top?

scripts/spot-grid.py
```diff
@@ -135,10 +140,10 @@ def update_view(self):
...
         info = [
             "rotation: {:.1f} deg".format(numpy.rad2deg(self.rotation)),
-            "pitch-x: {:.2f} um".format(PIXEL_SIZE_UM * self.scale[0]),
-            "pitch-y: {:.2f} um".format(PIXEL_SIZE_UM * self.scale[1]),
-            "translation-x: {:.1f} um".format(PIXEL_SIZE_UM * self.translation[0]),
-            "translation-y: {:.1f} um".format(PIXEL_SIZE_UM * self.translation[1]),
+            "pitch-x: {:.2f} um".format(self.pixelsize * self.scale[0]),
+            "pitch-y: {:.2f} um".format(self.pixelsize * self.scale[1]),
+            "translation-x: {:.1f} um".format(self.pixelsize * self.translation[0]),
+            "translation-y: {:.1f} um".format(self.pixelsize * self.translation[1]),
```
tepals, line 146: We do want to display these values in micrometers, I think this is now also in meters right?

scripts/spot-grid.py
tepals: Now that I've read the rest of the code, you could also have pxsize as a required argument, and not give it a default value.

tepals (changes requested):
Could you also run it on EA1 or FUMO-beta and put a screenshot on the chat? Then it's immediately clear if the pitch and translation are displayed in the correct format. 

Let me know if you need help with running the code on the system.

scripts/spot-grid.py
```diff
@@ -211,16 +216,17 @@ def start_generate(self):
         self.notify(self._detector.array)
 
 
-def live_display(ccd, dataflow, kill_ccd=True, gridsize=None):
+def live_display(ccd, dataflow, kill_ccd=True, gridsize=None, pxsize=PIXEL_SIZE_SAMPLE_PLANE / DEFAULT_MAGNIFICATION):
```
pieleric, line 219: I agree with Thera, it'd be fine to make the pxsize argument required.

pieleric (changes requested):

scripts/spot-grid.py
philipwinklerdelmic: You mean the 3.45 µm (pixel size sample plane) as an additional argument? Or have the pixelsize argument instead of the magnification (so the user would directly enter 3.45e-6/40 for the pixelsize)?

pieleric: I think she means: "leave pxsize, but drop the default value". At least, that's what I'd suggest to do.

tepals: Indeed I meant "leave pxsize, but drop the default value". The signature would become:
```
    def __init__(self, pxsize, title="Live image", size=(640, 480), gridsize=None):
```

tepals (approved):
looks good!

scripts/spot-grid.py
philipwinklerdelmic: @pieleric Just discussed with Thera, my issue with that was that if we do it like this, the hardware argument pxsize would come before the gui arguments (title, size) and that might be confusing? Maybe it's not such a big deal. What do you think? Another option is to do the division in another constant, so it looks a little cleaner.

PR #1837, opened 2021-07-27
Adjust enzel gui for meteor

Mahmood-B:
I made this PR again only to remove the commits of the enzel branch.


src/odemis/gui/cont/tabs.py
```diff
@@ -2201,60 +2202,130 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
-        self.position_btns = {LOADING: self.panel.btn_switch_loading, IMAGING: self.panel.btn_switch_imaging,
+        if main_data.role == 'enzel':  
+            try:
+                # start and end position are used for the gauge progress bar
+                self._start_pos = main_data.stage.position.value
+                self._end_pos = self._start_pos
+                # Show current position of the stage via the progress bar
+                main_data.stage.position.subscribe(self._update_progress_bar, init=False)
+                # get the stage and its meta data 
+                stage = self.tab_data_model.main.stage
```
lazem, line 2213: I think it's better to have the stage as attribute of the tab `self.stage` and point it to either the main stage in case of `enzel` or the underlying stage in case of `meteor`. After that you can use self.stage in subsequent code without having to check which microscope.

src/odemis/gui/cont/tabs.py
```diff
@@ -2201,60 +2202,130 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
+            panel.stage_align_btn_m_aligner_z.Bind(wx.EVT_BUTTON, self._on_aligner_btn)
+            panel.btn_cancel.Bind(wx.EVT_BUTTON, self._on_cancel)
+
+        elif main_data.role == 'meteor':
+            stage  = model.getComponent(role='stage-bare')
+            # start and end position are used for the gauge progress bar
+            self._start_pos = stage.position.value
+            self._end_pos = self._start_pos
+            # Show current position of the stage via the progress bar
+            self.tab_data_model.main.stage.position.subscribe(self._update_progress_bar, init=False)
```
lazem, line 2296: These lines are identical to enzel, let's put them outside.

src/odemis/gui/cont/tabs.py
```diff
@@ -2429,36 +2477,69 @@ def _enable_movement_controls(self, cancelled=False):
...
+            stage = self.tab_data_model.main.stage
+            # Get current movement (including unknown and on the path)
+            self._current_position = getCurrentPositionLabel(stage.position.value)
+            # Enable stage advanced controls on milling
+            self._enable_advanced_controls(True) if self._current_position is MILLING else self._enable_advanced_controls(
+                False)
+
+        elif self.tab_data_model.main.role == 'meteor':
+            self._current_position = getCurrentPositionLabel(model.getComponent(role='stage-bare').position.value)
+        self._enable_position_controls(self._current_position, cancelled)
```
lazem, line 2490: If you have self.stage as attribute, this can be rewritten as:

```
# Get current movement (would work for both enzel and metero)
self._current_position = getCurrentPositionLabel(stage.position.value)
self._enable_position_controls(self._current_position, cancelled)

# Enable stage advanced controls (you could put the enzel check inside this function to return immediately if it's not enzel)
 self._enable_advanced_controls(True) if self._current_position is MILLING else self._enable_advanced_controls(
                False)
```

src/odemis/gui/cont/tabs.py
lazem: You could also save the `self.tab_data_model.main.stage` as `self.role` as you're using it alot along the tab functions.

src/odemis/gui/cont/tabs.py
```diff
@@ -2429,36 +2477,69 @@ def _enable_movement_controls(self, cancelled=False):
...
+                    button.Enable()
+        elif self.tab_data_model.main.role == 'meteor':
+            # The move button should turn green only if current move is known and not cancelled
+            if current_position in self.position_btns.keys() and not cancelled:
+                currently_pressed = self.position_btns[current_position]
+                self._toggle_switch_buttons(currently_pressed)
+                currently_pressed.icon_on = img.getBitmap(self.btn_toggle_icons[currently_pressed][1])
+                currently_pressed.Refresh()
+            else:
+                self._toggle_switch_buttons(currently_pressed=None)
```
lazem, line 2523: That part is identical to enzel -> put it before the check..

src/odemis/gui/cont/tabs.py
```diff
@@ -2622,34 +2703,50 @@ def _perform_switch_position_movement(self, target_button):
...
+                if self._current_position is MILLING:
+                    milling_angle = self._get_milling_angle_value()
+                    if milling_angle is not None:
+                        self._prev_milling_angle = milling_angle
+                self._end_pos = self.target_position_metadata[self._target_position]
+                return cryoSwitchSamplePosition(self._target_position)
+            else:
+                # Target position is milling, get rx value from saved milling angle
+                rx_angle_value = self.ion_to_sample + self._prev_milling_angle
+                return cryoTiltSample(rx=rx_angle_value)
```
lazem, line 2736: This milling part will be removed on the updated tilted objective specs. With that in mind, performing the switch movement would be the same for enzel and meteor. I suggest you remove these lines from the function, then the code would much be simplified as:

```
self._start_pos = self.stage.position.value
 # target_position metadata has the end positions for all movements except milling
if self._current_position is LOADING:
    box = wx.MessageDialog(self.main_frame, "The sample will be loaded. Please make sure that the sample is properly set and the insertion stick is removed.",
                                    caption="Loading sample", style=wx.YES_NO | wx.ICON_QUESTION| wx.CENTER)

     box.SetYesNoLabels("&Load", "&Cancel")
     ans = box.ShowModal()  # Waits for the window to be closed
     if ans == wx.ID_NO:
          return
self._end_pos = self.target_position_metadata[self._target_position]
return cryoSwitchSamplePosition(self._target_position)
```

No need to check for enzel or meteor..

lazem (changes requested):
I have added some suggestions to have a common interface and avoid repeating the code in the chamber tab and move module between enzel and meteor.

src/odemis/gui/cont/tabs.py
Mahmood-B: For the first part, okay. 
But it would be really odd to put the enzel check inside the function `self._enable_advanced_controls()`. If this function is made solely for enzel, why would i check inside it if the model is enzel or meteor?

src/odemis/gui/cont/tabs.py
Mahmood-B: Not anymore actually. Once i push, you will see it got different.

install/linux/usr/share/odemis/sim/meteor-sim.odm.yaml
```diff
@@ -0,0 +1,198 @@
...
+        rng: [[-0.1, 0.1], [-0.05, 0.05], [0, 0.1], [-2, 2], [0, 6.28]],
+        unit: ["m", "m", "m", "rad", "rad"],
+        refproc: "Standard",
+    },
+    metadata: {
+        FAV_POS_DEACTIVE: {'rx': 0, 'rz': 0, 'x':  0.035, 'y':  0.0042, 'z':  -0.0068},
+        SEM_IMAGING_RANGE: {"x": [0.04, 0.06], "y": [-0.001, 0.003], "z": [0.000, 0.004]},
+        FM_IMAGING_RANGE: {"x": [0.07, 0.09], "y": [0.004, 0.008], "z": [0.0, 0.004]},
+        SAMPLE_CENTERS: {"GRID 1": {'x':  0.05, 'y':  0.001, 'z': 0.002}, "GRID 2": {'x': 0.0534, 'y': 0.001, 'z': 0.002}},
+        GRID_SHIFT: {"x": 0.0034},
```
pieleric, line 63: We don't need GRID_SHIFT. When switching to another grid, use the SAMPLE_CENTERS info.

src/odemis/acq/move.py
```diff
@@ -194,125 +240,253 @@ def _doCryoSwitchSamplePosition(future, stage, focus, align, target):
...
+    if not axis in stage.position.value.keys():
+        raise KeyError("The stage misses the %s axis" % axis)
+    stage_md = stage.getMetadata()
+    switched_pos = stage.position.value.copy()
+    for a in stage.position.value:
+        if a == axis:
+            if to_grid == meteor_labels[GRID_1]:
+                switched_pos[axis] = stage.position.value[axis] - stage_md[model.MD_GRID_SHIFT]["x"]
+            elif to_grid == meteor_labels[GRID_2]:
+                switched_pos[axis] = stage.position.value[axis] + stage_md[model.MD_GRID_SHIFT]["x"]
```
pieleric, line 488: Use SAMPLE_CENTERS.

src/odemis/gui/cont/tabs.py
```diff
@@ -448,9 +447,10 @@ def __init__(self, name, button, panel, main_frame, main_data):
             detvas=get_local_vas(main_data.sed, main_data.hw_settings_config)
         )
 
-        # The sem stream is always visible, so add it by default
-        sem_stream_cont = self._streambar_controller.addStream(sem_stream, add_to_view=True)
-        sem_stream_cont.stream_panel.show_remove_btn(False)
+        # The sem stream is visible for enzel, but not for meteor 
+        if self.main_data.role == 'enzel':
```
pieleric, line 451: Put the entire creation of the sem_stream in this `if` block.

src/odemis/gui/cont/tabs.py
```diff
@@ -2417,48 +2472,151 @@ def _update_progress_bar(self, pos):
...
         Toggle currently pressed button (if any) and untoggle rest of switch buttons
         """
-        for button in self.position_btns.values():
-            button.SetValue(1) if button == currently_pressed else button.SetValue(0)
+        if self._role == 'enzel':
+            for button in self.position_btns.values():
+                button.SetValue(1) if button == currently_pressed else button.SetValue(0)
+        elif self._role == "meteor":
+            # inital case 
+            if self.position_btns[GRID_1].GetValue() == 0 and self.position_btns[GRID_2].GetValue() == 0:
```
pieleric, line 2512: This is a really complex flow. I think it'd be clearer if you had a separate function for Grid buttons. These buttons are mostly independent from the "position" buttons.

src/odemis/gui/cont/tabs.py
```diff
@@ -2417,48 +2472,151 @@ def _update_progress_bar(self, pos):
...
+                self.panel.btn_switch_grid2.Disable()
+                self.panel.btn_switch_grid1.Disable()
+            elif current_pos_label != self._current_position and current_pos_label != UNKNOWN:
+                current_grid_label = getCurrentGridLabel(pos)
+                if current_grid_label == GRID_1:
+                    self.position_btns[GRID_1].SetValue(True)
+                    self.position_btns[GRID_2].SetValue(False)
+                elif current_grid_label == GRID_2:
+                    self.position_btns[GRID_2].SetValue(True)
+                    self.position_btns[GRID_1].SetValue(False)
```
pieleric, line 2491: This sort of code, you can write without `if`, more simply like this:
```python
                    self.position_btns[GRID_1].SetValue(current_grid_label == GRID_1)
                    self.position_btns[GRID_2].SetValue(current_grid_label == GRID_2)
```

pieleric (changes requested):

lazem (changes requested):

src/odemis/gui/cont/tabs.py
Mahmood-B: First of all,  i think it is not accurate to say that the grid buttons are independent of the position buttons. They clearly have some dependency on each other. 
For making separate functions for the grid buttons, I have looked into that. It actually will not make the code simpler. I would have to copy paste the code existing in the current button callback to the new grid buttons' callback. So that would literally be repeated code. Additionally, it will not change this code above. We need to distinguish between the initial case and the normal, and handle them separately. However, i do think that this code of the toggle function was not written in a readable way. I refactored the code. So now it is a bit simpler.

src/odemis/gui/cont/tabs.py
```diff
@@ -2417,48 +2472,151 @@ def _update_progress_bar(self, pos):
...
+                # if user switches imaging modes, keep grids the same. 
+                elif currently_pressed.Label == meteor_labels[SEM_IMAGING]:
+                    self.position_btns[SEM_IMAGING].SetValue(True)
+                    self.position_btns[FM_IMAGING].SetValue(False)
+                # if user switches imaging modes, keep grids the same. 
+                elif currently_pressed.Label == meteor_labels[FM_IMAGING]:
+                    self.position_btns[SEM_IMAGING].SetValue(False)
+                    self.position_btns[FM_IMAGING].SetValue(True)
```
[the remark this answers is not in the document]
Mahmood-B: It can not replace all the code. We have to detect and handle the initial case separately as it is different from the normal cases of pressing a button. But a part of the code seemed not needed after i had another look. So now it is less complicated. However, note that with this implementation if you press a button, it will not be immediately toggled, and stay toggled during the whole move. For example, assume currently the stage is in the sem area. If we press fm button, then it will not toggle fm button directly, because the stage has not totally moved out of the sem area yet. It will take some time until to goes out the sem area, then it will toggle the fm button. I have tried hard to make it similar to enzel buttons behavior, but i think it can not be done in the same way because of the listener of the stage position va. I could handle the toggling properly in this function using the button object itself, not the labels, and then it can be similar to enzel buttons behavior. But in the listener we do not have the button object, and so we have to use the labels which means that we can not have full control on how and when the buttons are toggled (as i mentioned mentioned in the example above).

src/odemis/gui/cont/tabs.py
```diff
@@ -2417,48 +2472,151 @@ def _update_progress_bar(self, pos):
...
+                self.panel.btn_switch_sem_imaging.Enable()
+                self.panel.btn_switch_fm_imaging.Enable()
+                self.panel.btn_switch_grid2.Disable()
+                self.panel.btn_switch_grid1.Disable()
+            elif current_position in [SEM_IMAGING, FM_IMAGING]:
+                # at SEM/FM position: G1=1, G2=1, FM=1, Loc=1 
+                self.panel.btn_switch_grid2.Enable()
+                self.panel.btn_switch_grid1.Enable()
+                self.panel.btn_switch_fm_imaging.Enable()
+                self.panel.btn_switch_sem_imaging.Enable()
```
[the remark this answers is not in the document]
Mahmood-B: What would happen if the user presses grid1 or grid2 after loading? Here we also need to define an imaging mode by default. Lets say it is sem imaging mode is the default. So if user presses grid1, then the stage would go to sem area, grid1. If grid2 is pressed, then stage would go to sem area, grid2. 

For the simplification, that doesn't work. Assume the label is loading, this means all buttons will be set to True (or toggled/pushed) which is not the correct behavior.

src/odemis/gui/cont/tabs.py
```diff
@@ -2622,34 +2745,74 @@ def _perform_switch_position_movement(self, target_button):
...
+                    self._end_pos = sem_grid1_pos
+                elif target_button.Label == meteor_labels[FM_IMAGING]:
+                    fm_grid1 = transformFromSEMToMeteor(sem_grid1_pos, self._stage.axes)
+                    self._end_pos = fm_grid1
+                    if not self._display_meteor_pos_warning_msg(self._end_pos):
+                        return 
+                elif target_button.Label == meteor_labels[GRID_1]:
+                    self._end_pos = stage_md[model.MD_SAMPLE_CENTERS][meteor_labels[GRID_1]] 
+                elif target_button.Label == meteor_labels[GRID_2]:
+                    self._end_pos = stage_md[model.MD_SAMPLE_CENTERS][meteor_labels[GRID_2]]
```
lazem, line 2781: I think now this part is almost identical to going to SEM part, so let's include them together, the only difference is that in Loading-FM move there's no position to convert to Meteor so we take grid 1. 
Btw, we also should display the message whenever we go to FM, not just from loading.

lazem (changes requested):

src/odemis/gui/cont/tabs.py
```diff
@@ -2417,48 +2478,104 @@ def _update_progress_bar(self, pos):
...
+                    self.position_btns[GRID_1].SetValue(True)
+                elif currently_pressed.Label == meteor_labels[SEM_IMAGING]:
+                    self.position_btns[SEM_IMAGING].SetValue(True) 
+                    self.position_btns[GRID_1].SetValue(True)
+                elif currently_pressed.Label == meteor_labels[GRID_1]:
+                    self.position_btns[SEM_IMAGING].SetValue(True)
+                    self.position_btns[GRID_1].SetValue(True)
+                elif currently_pressed.Label == meteor_labels[GRID_2]:
+                    self.position_btns[SEM_IMAGING].SetValue(True)
+                    self.position_btns[GRID_2].SetValue(True)
```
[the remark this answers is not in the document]
Mahmood-B: Considering that the blinking behavior of the buttons is accepted, then yes this part can be removed. This part was added for the attempt to keep the buttons toggled during the whole movement, similar to the enzel's buttons behavior.

src/odemis/acq/test/move_test.py
```diff
@@ -300,5 +303,216 @@ def test_smaract_stage_fallback_movement(self):
...
+    def move_to_loading_position(self):
+        f = model.CancellableFuture()
+        f._task_lock = threading.Lock()
+        f._task_state = RUNNING
+        f._running_subf = model.InstantaneousFuture()
+        filter_dict = lambda keys, d: {key: d[key] for key in keys}
+        sub_moves = []
+        sub_moves.append((self.stage, filter_dict({'x', 'y', 'z'}, self.stage_loading)))
+        for component, sub_move in sub_moves:
+            run_sub_move(f, component, sub_move)
```
lazem, line 343: It's ok here to just do `self.stage.moveAbs(self.stage_loading).result()`

src/odemis/gui/cont/tabs.py
```diff
@@ -2622,34 +2715,70 @@ def _perform_switch_position_movement(self, target_button):
...
+                    self._end_pos = transformFromSEMToMeteor(self._stage.position.value, self._stage.axes)
+                    if not self._display_meteor_pos_warning_msg(self._end_pos):
+                        return 
+                elif self._target_position == GRID_1:
+                    self._end_pos = stage_md[model.MD_SAMPLE_CENTERS][meteor_labels[GRID_1]]
+                elif self._target_position == GRID_2:
+                    self._end_pos = stage_md[model.MD_SAMPLE_CENTERS][meteor_labels[GRID_2]]
+            elif current_position == FM_IMAGING:
+                if self._target_position == GRID_1:
+                    sem_grid1 = stage_md[model.MD_SAMPLE_CENTERS][meteor_labels[GRID_1]]
```
lazem, line 2758: Use `sem_grid1_pos `

lazem (changes requested):
A few more comments and hopefully we merge soon..

lazem (changes requested):

Mahmood-B:
@lazem  By the way, i'm wondering, shouldn't you add the correct formulas for the transformations? Also the correct order of moves? How would you merge while the those are not correct?

src/odemis/acq/move.py
```diff
@@ -194,125 +294,221 @@ def _doCryoSwitchSamplePosition(future, stage, focus, align, target):
...
+    transformed_pos['y']=pos["y"]+0.005
+    return transformed_pos
+
+
+def transformFromMeteorToSEM(pos, axes):
+    # TODO add docstring
+    # TODO check for all the required axes
+    if not {"x", "y"}.issubset(axes):
+        raise KeyError("The stage %s misses 'x' and 'y' axes")
+    transformed_pos = pos.copy()
```
lazem, line 506: This line is duplicated.

src/odemis/acq/move.py
```diff
@@ -34,20 +34,75 @@
...
 
-def getCurrentPositionLabel(current_pos, stage):
+def getTargetPosition(target_pos_lbl, stage):
     """
-    Determine where lies the current stage position
-    :param current_pos: (dict str->float) Current position of the stage
-    :param stage: Sample stage that's being controlled
-    :return: (int) a value representing stage position from the constants LOADING, IMAGING, TILTED, COATING..etc
+    
+    """
```
lazem, line 49: If you put doc string quotes, then don't leave it empty! ;)

src/odemis/acq/move.py
```diff
@@ -85,27 +140,69 @@ def getCurrentPositionLabel(current_pos, stage):
...
+        return _getCurrentMeteorPositionLabel(current_pos, stage)
 
-    def get_distance(start, end):
+
+def _getDistance(start, end):
         # Calculate the euclidean distance between two 3D points
         axes = start.keys() & end.keys()  # only the axes found on both points
         sp = numpy.array([start[a] for a in axes])
         ep = numpy.array([end[a] for a in axes])
         return scipy.spatial.distance.euclidean(ep, sp)
```
lazem, line 190: Fix the indentation in here.

src/odemis/acq/move.py
```diff
@@ -194,125 +294,221 @@ def _doCryoSwitchSamplePosition(future, stage, focus, align, target):
...
+            # Create axis->pos dict from target position given smaller number of axes
+            filter_dict = lambda keys, d: {key: d[key] for key in keys}
+            # get the current label 
+            current_pos_label = getCurrentPositionLabel(stage.position.value, stage)
+            # get the set point position
+            target_pos = getTargetPosition(target, stage)
+            # determine the order of moves for the stage and focuser
+            if current_pos_label in [LOADING, SEM_IMAGING]:
+                if target in [SEM_IMAGING, GRID_1]:  
+                    # TODO put the correct order of moves
```
lazem, line 438: If we are doing a loading to sem move, then we'll need to make sure the focus is parked.

src/odemis/model/_metadata.py
```diff
@@ -186,6 +186,11 @@
 MD_ION_BEAM_TO_SAMPLE_ANGLE = "Ion beam to sample angle"  # (float) angle between ion beam and sample stage
 MD_SAFE_REL_RANGE = "Safe relative range"  # (float, float) +/- safe range relative to a value
 MD_SAFE_SPEED_RANGE = "Safe speed range"  # (float, float) min, max of the safe speed range
+MD_SAMPLE_CENTERS = "Centers position of grids"
+MD_GRID_SHIFT = "Shift value between grids"
```
lazem, line 190: Remove MD_GRID_SHIFT it's not used anymore.

src/odemis/acq/move.py
```diff
@@ -194,125 +294,221 @@ def _doCryoSwitchSamplePosition(future, stage, focus, align, target):
...
+                    # park focus for safety 
+                    if not _isNearPosition(focus.position.value, focus_deactive, focus.axes):
+                        sub_moves.append((focus, focus_deactive))
+                    sub_moves.append((stage, filter_dict({'x', 'y', 'z'}, target_pos)))
+                    sub_moves.append((stage, filter_dict({'rx', 'rz'}, target_pos)))  
+            else:
+                return
+            # run the moves 
+            try:
+                for component, sub_move in sub_moves:
```
lazem, line 471: You could add a log for the movement from current to target position.

src/odemis/gui/cont/tabs.py
```diff
@@ -631,7 +631,12 @@ def _on_stage_pos(self, pos):
         Called when the stage is moved, enable the tab if position is imaging mode, disable otherwise
         :param pos: (dict str->float or None) updated position of the stage
         """
-        guiutil.enable_tab_on_stage_position(self.button, self.stage, pos, target=IMAGING)
+        if self.main_data.role == 'meteor':
+            pos = self.main_data.stage_bare.position.value 
+            stage = self.tab_data_model.main.stage_bare
+        elif self.main_data.role == "enzel":
+            stage = self.stage
+        guiutil.enable_tab_on_stage_position(self.button, stage, pos, target=[IMAGING, FM_IMAGING])
```
lazem, line 641: Don't leave blank lines.

src/odemis/gui/cont/tabs.py
```diff
@@ -2201,91 +2207,152 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
+            # Determine and show current position of the stage
+            self._current_position, self._target_position = None, None
+            self._enable_movement_controls()
+            if self._current_position in self.position_btns.keys():
+                pos_button = next(button for pos, button in self.position_btns.items() if pos == self._current_position)
+                self._toggle_switch_buttons(pos_button)
+            self._show_cancel_warning_msg(None)
+
+        elif self._role == 'meteor':
+            self._stage  = self.tab_data_model.main.stage_bare
```
lazem, line 2304: Remove extra spaces before =

src/odemis/gui/cont/tabs.py
```diff
@@ -2201,91 +2207,152 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
+            if not model.MD_SEM_IMAGING_RANGE in self._stage.getMetadata():
+                raise ValueError("The stage misses the 'SEM_IMAGING_RANGE' metadata.")
+            if not model.MD_FM_IMAGING_RANGE in self._stage.getMetadata():
+                raise ValueError("The stage misses the 'FM_IMAGING_RANGE' metadata.")
+            # Fail early when required axes are not found on the positions metadata
+            focuser = self.tab_data_model.main.focus
+            focus_md = focuser.getMetadata()
+            required_axes = {'z'}
+            for focuser_position in focus_md.values():
+                if not required_axes.issubset(focuser_position.keys()):
```
lazem, line 2325: Not all focuser metadata has to have z axis. Only POS_ACTIVE & POS_DEACTIVE positions. So make sure to only check for these metadata.

src/odemis/acq/move.py
```diff
@@ -34,20 +34,75 @@
...
+    else:
+        end_pos = None 
+    return end_pos
+
+
+def getCurrentGridLabel(current_pos, stage):
+    """
+    Detects which grid on the sample shuttle of meteor being viewed
+    current_pos (dict str->float): position of the stage
+    stage (Actuator): the stage component
```
lazem, line 83: Should be written as `stage: (Actuator)`

src/odemis/gui/cont/tabs.py
```diff
@@ -2622,34 +2722,46 @@ def _perform_switch_position_movement(self, target_button):
...
+        box.SetYesNoLabels("&Load", "&Cancel")
+        ans = box.ShowModal()  # Waits for the window to be closed
+        return False if ans == wx.ID_NO else True 
+
+    def _display_meteor_pos_warning_msg(self, end_pos):
+        x, y, z = end_pos.get('x'), end_pos.get('y'), end_pos.get('z')
+        box = wx.MessageDialog(self.main_frame, "the sample stage will move to x_meteor = {x}, y_meteor = {y}, z_meteor = {z}, is this safe?".format(x=x, y=y, z=z),
+                            caption="Moving to METEOR", style=wx.YES_NO | wx.ICON_QUESTION | wx.CENTER)
+        ans = box.ShowModal()  # Waits for the window to be closed
+        return False if ans == wx.ID_NO else True
```
lazem, line 2764: This could be rewritten as `return ans != wx.ID_NO`

lazem (changes requested):

src/odemis/gui/cont/tabs.py
```diff
@@ -2301,91 +2304,158 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
+                raise ValueError("The stage misses the 'FM_IMAGING_RANGE' metadata.")
+            if not model.MD_POS_COR in self._stage.getMetadata():
+                raise ValueError("The stage misses the 'POS_COR' metadata.")
+            # Fail early when required axes are not found on the positions metadata
+            focuser = self.tab_data_model.main.focus
+            focus_md = focuser.getMetadata()
+            # take only the active and deactive position of the focuser
+            for md in focus_md.keys():
+                if md not in [model.MD_FAV_POS_DEACTIVE, model.MD_FAV_POS_ACTIVE]:
+                    focus_md.pop(md)
```
lazem, line 2425: I don't think you can remove from the dictionary during iteration (try it). I believe you could iterate the dict items so you have both the keys & value in this manner:
```
for f_key, f_value in focus_md.items():
      if f_key in [model.MD_FAV_POS_DEACTIVE, model.MD_FAV_POS_ACTIVE] & not required_axis.issubset(f_value.keys()):
```

src/odemis/gui/cont/tabs.py
```diff
@@ -74,7 +73,7 @@
     PixelTemporalSpectrumProjection, SinglePointTemporalProjection, \
     ScannedTemporalSettingsStream, \
     ARRawProjection, ARPolarimetryProjection, StaticStream, LiveStream
-from odemis.acq.move import LOADING, IMAGING, MILLING, COATING, UNKNOWN, LOADING_PATH, target_pos_str
+from odemis.acq.move import GRID_1, LOADING, IMAGING, COATING, UNKNOWN, LOADING_PATH, getCurrentGridLabel, target_pos_str, meteor_labels, FM_IMAGING, SEM_IMAGING, GRID_2, getTargetPosition
```
lazem, line 76: Please keep the MILLING constant as it's used by the enzel part of chamber tab.

src/odemis/acq/move.py
```diff
@@ -34,20 +34,80 @@
...
+def getTargetPosition(target_pos_lbl, stage):
     """
-    Determine where lies the current stage position
-    :param current_pos: (dict str->float) Current position of the stage
-    :param stage: Sample stage that's being controlled
-    :return: (int) a value representing stage position from the constants LOADING, IMAGING, TILTED, COATING..etc
+    Returns the position that the stage would go to.
+    target_pos_lbl (int): a lable representing a position (SEM_IMAGING, FM_IMAGING, GRID_1 or GRID_2)
+    stage (Actuator): a stage component 
+    returns (dict->float): the end position of the stage
```
lazem, line 51: Indicate also that it could  return None.

src/odemis/model/_metadata.py
```diff
@@ -188,6 +188,10 @@
 MD_ION_BEAM_TO_SAMPLE_ANGLE = "Ion beam to sample angle"  # (float) angle between ion beam and sample stage
 MD_SAFE_REL_RANGE = "Safe relative range"  # (float, float) +/- safe range relative to a value
 MD_SAFE_SPEED_RANGE = "Safe speed range"  # (float, float) min, max of the safe speed range
+MD_SAMPLE_CENTERS = "Centers position of grids"
+MD_SEM_IMAGING_RANGE = "SEM imaging range"
+MD_FM_IMAGING_RANGE = "FM imaging range"
```
lazem, line 193: You could add a comment about what these metadata represent and what are their types.

lazem (changes requested):

src/odemis/gui/cont/tabs.py
lazem: Mmmmh, I added an additional metadata without the z axis, it gave me this error:
`RuntimeError: dictionary changed size during iteration`

Mahmood-B: I see. When i tested i changed the axis from z to x. This is the error i get which is the expected. 
error

Mahmood-B: Apparently, i only tested the part that checks for the required axis, not the part that pops/removes the unwanted metadata.

Mahmood-B:
@lazem In the last commit (ff9f079) i had to make changes that probably have affected enzel cancel button behavior. Would you kindly fix that? Because, otherwise, i will have to go through the new workflow of enzel to be able to fix it.

Mahmood-B:
@pieleric , the issue of the gauge doesn't get filled completely still happens. The refresh() function doesn't always work. Confusing. 
@lazem , Does the issue of incomplete bar exist also in enzel?

src/odemis/acq/move.py
```diff
@@ -234,12 +239,40 @@ def getCurrentAlignerPositionLabel(current_pos, align):
...
+            Rz_end = Rotation.from_euler("z", end['rz']).as_matrix()
+            Rz_start = Rotation.from_euler('z', start["rz"], degrees=False).as_matrix()
+    # find the total rotations
+    R_start = numpy.matmul(Rz_start, Rx_start)
+    R_end = numpy.matmul(Rz_end, Rx_end)
+    # find the rotation error matrix
+    R_diff = numpy.matmul(numpy.transpose(R_start), R_end)
+    # map to scalar error
+    rot_error = 0.03*numpy.trace(numpy.eye(3)-R_diff)
+    return lin_error+rot_error
```
Mahmood-B, line 275: My eye came on this function again when i was reviewing the code for the last time before committing and pushing. The Euclidean distance can be calculated only between Cartesian points. It can not be calculated between angles. This explains the weird behavior of the gauge (the gauge does not progress if the stage moves along x, y or z, but progresses if the stage moves around rx or rz). Therefore the linear error and the rotation error must be calculated separately and then summed. The rotation error can be calculated as follows:

1. Find the elemental rotations of the stage.
2. Find the total rotation of the stage which is the multiplication of the elemental rotations
R
Luckily, scipy contains interesting classes for rotations which makes the calculations of rotation matrices easy.
3. Find the rotation error matrix which is the relative rotation between the given point1 and point2.
Rerr
4. Map/convert the rotation error matrix into a scalar rotation error
scalar

I have tested the gauge movement after that, and it looks better now. There is "_continuity_" in the gauge movement (the gauge moves for x, y, z movements and continues moving for rx and rz). 

@Bassim Would you please test enzel gauge, and see how it looks? If not so good, I need to tune the parameter multiplied by the trace to a value that is good for both meteor and enzel movements.

src/odemis/acq/move.py
```diff
@@ -275,22 +275,41 @@ def _getDistance(start, end):
...
+    if axis == "rx":
+        Rx = numpy.array([[1, 0, 0], [0, numpy.cos(angle), -numpy.sin(angle)], [0, numpy.sin(angle), numpy.cos(angle)]])
+        return Rx
+    elif axis == "ry":
+        Ry = numpy.array([[numpy.cos(angle), 0, numpy.sin(angle)], [0, 1, 0], [-numpy.sin(angle), 0, numpy.cos(angle)]])
+        return Ry
+    elif axis == "rz":
+        Rz = numpy.array([[numpy.cos(angle), -numpy.sin(angle), 0], [numpy.sin(angle), numpy.cos(angle), 0], [0, 0, 1]])
+        return Rz
+
```
pieleric, line 311: You need here a code handling the fallback, so that it's clear that the axis argument was wrong (instead of returning implicitly `None`). 
=> Add `raise ValueError(f"Unknown axis name {axis}")

src/odemis/acq/move.py
```diff
@@ -275,22 +275,41 @@ def _getDistance(start, end):
...
-        rot_error = SCALING_FACTOR*abs(numpy.trace(numpy.eye(3)-R_diff))
-    return lin_error+rot_error
+        rot_error = SCALING_FACTOR * abs(numpy.trace(numpy.eye(3) - R_diff))
+    return lin_error + rot_error
     
 
+def getRotationMatrix(axis, angle):
+    """
+    Computes the rotation matrix of the given angle around the given axis. 
+    The axis can be 'rx', 'ry' or 'rz'. The angle must be in radians.
```
pieleric, line 297: You can move these descriptions to the respective argument descriptions just below.

src/odemis/acq/move.py
```diff
@@ -275,22 +275,41 @@ def _getDistance(start, end):
...
+    axis (str): the axis around which the rotation matrix is to be calculated.
+    angle (float): the angle of rotation.
+    return (numpy.ndarray): the rotation matrix.
```
pieleric, line 300: Please indicate the shape and type: 3x3 of float.

install/linux/usr/share/odemis/sim/meteor-sim.odm.yaml
```diff
@@ -0,0 +1,203 @@
...
+        # Matrix to convert from X/Y to X/Y/Z (on the underlying stage)
+        # xd = POS_COR[0] + CALIB[0,0] * x + CALIB[0,1] * y
+        # yd = POS_COR[1] + CALIB[1,0] * x + CALIB[1,1] * y
+        # zd = POS_COR[2] + CALIB[2,0] * x + CALIB[2,1] * y
+        CALIB: [[1, 0],  [0, 0.6941955197522921], [0, -0.7197864824764667]],
+        POS_COR: [0, -0.0007363771884364807, 0.004945901888925916],
+        # Allowed range when in FM imaging mode. The X/Y ranges are also used
+        # as default area for the overview acquisition. The values must be in the
+        # stage coordinates, so converted from the SEM stage (in Y).
+        POS_ACTIVE_RANGE: {"x":  [50.92e-3, 53.45e-3], "y":  [-4.3e-3, -1.82e-3]},
```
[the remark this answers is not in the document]
pieleric: ACTIVE_RANGE is used by the GUI to limit the area where the user is allowed to go when dragging/moving the image. It has to be on the component with role=`stage`. 
The FM_IMAGING_RANGE is on the stage-bare. I think you could also use ACTIVE_RANGE there, as it would make sense to have both simultaneously. I'm also fine with leaving FM_IMAGING_RANGE as it's clearer what it is about.

pieleric (changes requested):

pieleric:
Only some very minor comments on the docstring of your new function. It should be very straightforward to fix.
Please also rebase against master, as now some of the tab code has already been merged, there are currently some conflicts.

PR #1838, opened 2021-07-28
fastem-gui-overview: fitToContent tool

philipwinklerdelmic:


src/odemis/gui/comp/miccanvas.py
```diff
@@ -1867,3 +1867,9 @@ def on_dbl_click(self, evt):
     def setView(self, view, tab_data):
         super(FastEMAcquisitionCanvas, self).setView(view, tab_data)
         view.show_crosshair.value = False
+
+    def fit_view_to_content(self, recenter=None):
+        # When the stream is playing for the first time, the view should be recentered to show the full image
+        if recenter is None:
+            recenter = True
+        super().fit_view_to_content(recenter)
```
pieleric, line 1875: Instead of adding even more exception to the canvas, let's make it generic.
See my PR #1841 : it now handles CAN_MOVE_STAGE explicitly in the standard fit_view_to_content(). This way, the ENZEL GUI is also fixed :-) 

=> you can drop this part of the PR. (the addition of the tool button is still good).

pieleric (changes requested):

src/odemis/gui/comp/miccanvas.py
philipwinklerdelmic: done

PR #1839, opened 2021-07-28
FindGridSpots: added transformation method selection.

tepals:
- Sometimes it is necessary that the returned grid is a perfect grid, with equal scaling in x and y and no shear. With the possibility to select the similarity method this is now possible.
- Refactored the transformation calls to not use the deprecated call method, but use the apply method.

### DIFF
--- a/src/odemis/acq/align/spot.py
+++ b/src/odemis/acq/align/spot.py
@@ -46,6 +46,9 @@ FOV_MARGIN = 250  # pixels
 STAGE_MOVE = "Stage move"
 BEAM_SHIFT = "Beam shift"
 OBJECTIVE_MOVE = "Objective lens move"
+# Constants for selecting the correct method in FindGridSpots
+GRID_AFFINE = "affine"
+GRID_SIMILARITY = "similarity"
 
 
 def MeasureSNR(image):
@@ -342,7 +345,7 @@ def FindSpot(image, sensitivity_limit=100):
     return max_pos
 
 
-def FindGridSpots(image, repetition, method="affine"):
+def FindGridSpots(image, repetition, spot_size=18, method=GRID_AFFINE):
     """
     Find the coordinates of a grid of spots in an image. And find the
     corresponding transformation to transform a grid centered around the origin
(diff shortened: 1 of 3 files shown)


src/odemis/acq/align/test/spot_test.py
```diff
@@ -211,6 +211,33 @@ def test_find_grid(self):
...
+        """
+        Create an image with a grid of 8 by 8 spots. Then test if the spots are found in the correct coordinates and
+        that the rotation, scaling and translation are correct when using the similarity transformation method.
+        """
+        image = numpy.zeros((256, 256))
+        # set a grid of 8 by 8 points to 1
+        image[54:150:12, 54:150:12] = 1
+        spot_coordinates, translation, scaling, rotation, shear = spot.FindGridSpots(image, (8, 8), method='similarity')
+        self.assertAlmostEqual(rotation, 0, places=4)
+        self.assertEqual(shear, None)
```
Sabrina85, line 224: There is an assertIsNone function that you can use :)

Sabrina85 (changes requested):
I have only two small comments on the tests. The rest looks fine!

src/odemis/acq/align/spot.py
```diff
@@ -354,6 +354,11 @@ def FindGridSpots(image, repetition):
         Data array containing the greyscale image.
     repetition : tuple of ints
         Number of expected spots in (X, Y).
+    method : str, "affine" or "similarity"
```
pieleric, line 357: Instead of a "str", please create two constants (like `GRID_AFFINE`, `GRID_SIMILARITY`) and use them here.
This way, it's clearer that a it's not *any* string, and also you'll reduce chances of typos in the name.

src/odemis/acq/align/test/spot_test.py
```diff
@@ -211,6 +211,33 @@ def test_find_grid(self):
...
 
+    def test_find_grid_similarity(self):
+        """
+        Create an image with a grid of 8 by 8 spots. Then test if the spots are found in the correct coordinates and
+        that the rotation, scaling and translation are correct when using the similarity transformation method.
+        """
+        image = numpy.zeros((256, 256))
+        # set a grid of 8 by 8 points to 1
+        image[54:150:12, 54:150:12] = 1
+        spot_coordinates, translation, scaling, rotation, shear = spot.FindGridSpots(image, (8, 8), method='similarity')
```
pieleric, line 222: So the two methods return... exactly the same thing (excepted that shear = None)? That feels a little weak as test case. Could you find a different input where the output is really different between the 2 methods?

pieleric (approved):

tepals:
I added another test case. In the process of creating, I found out that the minimum amount of spots in the grid is 6, therefore I added checks for that as well.

PR #1841, opened 2021-07-30
gui comp viewport/miccanvas: Fold view.has_stage() into CAN_MOVE_STAGE

pieleric:
Now that we have the CAN_MOVE_STAGE ability, we can use it also
where we used to check if the view has a stage.

At init, if the view doesn't have a stage, drop the CAN_MOVE_STAGE.
Everywhere else, use CAN_MOVE_STAGE.

This is more than cosmetic: now if we have explicitly dropped
CAN_MOVE_STAGE, even if there is a stage, the canvas behaves
really as if there is no stage. That's important in fit_to_content()
and also allowing to drag when paused.

### DIFF
--- a/src/odemis/gui/comp/viewport.py
+++ b/src/odemis/gui/comp/viewport.py
@@ -698,7 +698,7 @@ class FixedOverviewViewport(MicroscopeViewport):
         # we have to allow (and in the worst case the user will be able to move
         # while the chamber is opened)
         if (chamber_state in {CHAMBER_VACUUM, CHAMBER_UNKNOWN} and
-                self._view.has_stage()):
+                CAN_MOVE_STAGE in self.canvas.abilities):
             self.canvas.point_select_overlay.active.value = True
         else:
             self.canvas.point_select_overlay.active.value = False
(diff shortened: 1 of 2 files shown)


src/odemis/gui/comp/viewport.py
```diff
@@ -1923,7 +1925,3 @@ def setView(self, view, tab_data):
         super().setView(view, tab_data)
         self.canvas.add_background_overlay(self._tab_data_model.main.background)
 
-    def _on_stream_play(self, is_playing):
```
philipwinklerdelmic, line 1926: I think we still need that code. In the superclass, dragging is disabled, so if we pause the stream, we cannot move around the overview images anymore. Maybe instead of this, we could remove CAN_MOVE_STAGE and add CAN_DRAG to the abilities in the init of the class? This is not done automatically since we have a stage in this viewport for the crosshairoverlay.

pieleric: This is precisely what the `FeatureOverviewViewport.__init__()` does. It removes CAN_MOVE_STAGE (and CAN_DRAG is set by default). 
I did try the code with the FASTEM, in the overview tab. It's (still) possible to drag the view all the time.

PR #1842, opened 2021-08-03
Spot peak

effting:
Include skimage's peak_local_max function as it allows more options for filtering spots: intensity (absolute and relative), distance, etc. And it's faster as well.


src/odemis/util/peak_local_max.py
```diff
@@ -0,0 +1,482 @@
...
+    -------
+    output : ndarray
+        A subset of coord where a minimum spacing is guaranteed.
+
+    """
+    # Use KDtree to find the peaks that are too close to each other
+    tree = cKDTree(coord)
+
+    indices = tree.query_ball_point(coord, r=spacing, p=p_norm)
+    rejected_peaks_indices: Set[int] = set()
```
tepals, line 82: I've never seen this notation outside the function signature. What's the advantage of using it here? I checked and it doesn't force the set to be ints.

effting: It's a feature added to Python 3.5, see: https://docs.python.org/3/library/typing.html
It is not enforced at run-time, so it doesn't slow down a program. In my view there are two benefits:
- you can check the expected argument and return types directly from the function declaration
- you can have your code checked by a static type checker (find bugs)

This line is a type hint to indicate to the type checker that the variable `rejected_peaks_indices` is a set of ints.

tepals (commented):
one more small comment

src/odemis/util/peak_local_max.py
```diff
@@ -0,0 +1,482 @@
...
+    Return the coordinates of the peaks with the highest intensity.
+
+    Parameters
+    ----------
+    image : ndarray
+        Input image.
+    mask : ndarray
+        Binary array defining the region of `image` to evaluate.
+    num_peaks : int, optional
+        Maximum number of peaks. When the number of peaks exceeds `num_peaks`,
```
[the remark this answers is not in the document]
effting: Thanks; updated.

This is actually an example of where I used a static type checker. In the original skimage code the default value for num_peaks was a float (numpy.inf).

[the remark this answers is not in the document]
tepals: yes, I have seen it in docstrings and function arguments. I had just never seen it inside the code of a function and did not know it also worked there, nice to learn something new :)

tepals (approved):

src/odemis/util/peak_local_max.py
pieleric: If you have mypy activated in your IDE, then it will complain if the set doesn't contain ints (or if it's not a set).

pieleric (commented):
I don't have any specific remark on the code. However, I wonder why should we have duplicate code from scikit? Shouldn't we directly use skimage if it has this code already? It seems a pretty safe dependency (as it's maintained and already packaged in Ubuntu).

effting:
> I don't have any specific remark on the code. However, I wonder why should we have duplicate code from scikit? Shouldn't we directly use skimage if it has this code already? It seems a pretty safe dependency (as it's maintained and already packaged in Ubuntu).

If I recall correctly, last time this was discussed we didn't wan't to have skimage as a dependency. If you are OK with adding skimage as a dependency that would be identical.

pieleric:
> If I recall correctly, last time this was discussed we didn't wan't to have skimage as a dependency. If you are OK with adding skimage as a dependency that would be identical.
I would rather have a dependency on skimage than start copying code from it. That's assuming that the python3-skimage package from Ubuntu 18.04 is sufficiently new for you (ie, we don't need to package anything).

BTW, for now there is no code using this function. Is there a plan to add soon such user, or for now such user would live in a separate tree (eg, calibration repo). In case of the last one, I'd suggest to move the dependency to that tree.

effting:
> I would rather have a dependency on skimage than start copying code from it. That's assuming that the python3-skimage package from Ubuntu 18.04 is sufficiently new for you (ie, we don't need to package anything).
> 
> BTW, for now there is no code using this function. Is there a plan to add soon such user, or for now such user would live in a separate tree (eg, calibration repo). In case of the last one, I'd suggest to move the dependency to that tree.

The package python3-skimage in Ubuntu 18.04 LTS (bionic) is version v0.13.1-2. Current version on GitHub is v0.18.2 so the Ubuntu version is quite old. I think it will work though, but it will be slower (the filtering is much improved in v0.18.2).

The updated pattern calibration relies on this function, for which I didn't create a PR yet. That procedure is dependent / extends the spot / grid / transform / etc. functions already in odemis.util so I thought it would make more sense to have it merged there.

pieleric:
Alright, let's merge it for now, and maybe one day in the future we can depend on skimage, and remove it.

PR #1843, opened 2021-08-03
re-standardize the chamber axis to "vacuum"

pieleric:
This is a double edge PR. Firstly, it makes all the drivers and code use the axis "vacuum" on the chamber to control the vacuum level. Until now some drivers (orsay, xt_client) used "vacuum" while others (phenom, tescan, simulated) used "pressure". As they all only support a very limited set of positions, it makes more sense to call the axis "vacuum". There is a VA "pressure" to read the actual pressure.

Secondly, it extends the FASTEM simulator and fixes the GUI to work again together.

### DIFF
--- a/src/odemis/driver/orsay.py
+++ b/src/odemis/driver/orsay.py
@@ -20,13 +20,15 @@ You should have received a copy of the GNU General Public License along with
 Odemis. If not, see http://www.gnu.org/licenses/.
 """
 
-from odemis import model
-from odemis.model import isasync, CancellableThreadPoolExecutor, HwError
-from ConsoleClient.Communication.Connection import Connection
-
+import logging
+from odemis import model, util
+from odemis.model import isasync, CancellableThreadPoolExecutor, HwError, \
+    InstantaneousFuture
 import threading
 import time
-import logging
+
+from ConsoleClient.Communication.Connection import Connection
+
 
 VALVE_UNDEF = -1
 VALVE_TRANSIT = 0
@@ -409,6 +411,14 @@ class pneumaticSuspension(model.HwComponent):
             self._gauge = None
 
 
+# Very approximate values
+PRESSURE_VENTED = 120e3  # Pa
+PRESSURE_VACUUM = 20  # Pa
+PRESSURE_HV = 50e-3  # Pa
+
+VACUUM_STATUS_TO_PRESSURE = {0: PRESSURE_VENTED, 1: PRESSURE_VACUUM, 2: PRESSURE_HV}
+
+
 class vacuumChamber(model.Actuator):
     """
     This represents the vacuum chamber from Orsay Physics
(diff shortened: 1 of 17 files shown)


src/odemis/driver/test/orsay_test.py
```diff
@@ -372,10 +365,12 @@ def test_updatePosition(self):
...
         init_state = self.pressure._chamber.VacuumStatus.Actual
-        test_value = 1
-        self.pressure._chamber.VacuumStatus.Actual = test_value
+
+        vac_choices = self.pressure.axes["vacuum"].choices
+        pvac_pos = next(p for p, n in vac_choices.items() if n == "primary vacuum")
+        self.pressure._chamber.VacuumStatus.Actual = 1
         sleep(1)
-        self.assertEqual(int(self.pressure.position.value['vacuum']), test_value)
+        self.assertEqual(int(self.pressure.position.value['vacuum']), pvac_pos)
```
Helsloot, line 373: Here the int( should be float(. Then it passes.

src/odemis/driver/orsay.py
```diff
@@ -524,7 +543,12 @@ def moveAbs(self, pos):
         Move the axis of this actuator to pos.
         """
         self._checkMoveAbs(pos)
-        return self._executor.submit(self._changeVacuum, goal=pos["vacuum"])
+
+        if "vacuum" in pos:
+            vac_status = util.index_closest(pos["vacuum"], VACUUM_STATUS_TO_PRESSURE)
+            return self._executor.submit(self._changeVacuum, goal=vac_status)
+        else:
+            return InstantaneousFuture()
```
Helsloot, line 551: In the logging message on line 533 'choices[goal]' should be 'choices[VACUUM_STATUS_TO_PRESSURE[goal]]'
I can't add this comment at that line, so I placed it here.

src/odemis/driver/xt_client.py
```diff
@@ -1694,19 +1694,27 @@ def _set_scanner(self, scanner_name):
         return scanner_name
 
 
+# Very approximate values
+PRESSURE_VENTED = 100e3  # Pa
```
philipwinklerdelmic, line 1698: Isn't it confusing to have a pressure value here? Especially since we also have a .pressure VA. I have the feeling that it would be more intuitive to just use an on/off or True/False value on this axis.

philipwinklerdelmic (approved):
Looks fine. My only comment is that I'm not really convinced that setting the "vacuum" axis to a numerical Pascal value is the best idea. But if you think it's the right way to go, you can merge it.

src/odemis/driver/xt_client.py
pieleric: Normally the vacuum axis linked to the pressure. It's sort of the "target" pressure. That's the idea behind it.

PR #1844, opened 2021-08-06
ENZEL: Tilted Objective Back-end

lazem:
Having the optical objective lens at a fixed (~28 °) angle allows for simultaneous and coincident 3-beam imaging. That makes life easier than rotating back and forth between flat imaging angle and FIB milling one. This PR includes the back-end modifications needed for the modified tilted objective design. Most notably, the fixed lens rotational angle will be compensated for by wrapping the lens 3DOF stage. Moreover, having the 3 beams aligned and looking at roughly the same point in 3D space also means all moves for positioning the sample should be done with the sample stage (with focus being 5DOF stage Z). 
The updated safe movement procedures are also included.


install/linux/usr/share/odemis/sim/enzel-sim.odm.yaml
```diff
@@ -1,416 +1,414 @@
...
+    output_channel: 2,
+  },
+  properties: {
+    targetTemperature: -173, # °C
+  },
+  metadata: {
+    # TODO: Change values from test to real values
+    SAFE_REL_RANGE: [ -5, 5 ],  # °C, safe operating range relative to targetTemperature
+    SAFE_SPEED_RANGE: [ -0.5, 0.5 ] # °C/s, safe operating speed
+  },
```
Kleijwegt, line 32: Because of the style/tab changes made through this file it is hard to review. It would be nice if that is a seperate PR>
That said I'm in favor of these changes and the file does look ok to me.

Kleijwegt (changes requested):

src/odemis/acq/move.py
```diff
@@ -54,23 +54,31 @@ def getCurrentPositionLabel(current_pos, stage):
     stage_active = stage_md[model.MD_FAV_POS_ACTIVE]
     stage_active_range = stage_md[model.MD_POS_ACTIVE_RANGE]
     stage_coating = stage_md[model.MD_FAV_POS_COATING]
+    stage_alignment = stage_md[model.MD_FAV_POS_ALIGN]
+    stage_sem_imaging = stage_md[model.MD_FAV_POS_SEM_IMAGING]
```
Mahmood-B, line 58: Maybe check if these metadata exist before using them. I think there is already check for the active, deactive and coating positions in the cryo chamber tab class. You could check for the new two

src/odemis/acq/move.py
```diff
@@ -81,6 +89,48 @@ def getCurrentPositionLabel(current_pos, stage):
...
+
+
+def getCurrentAlignerPositionLabel(current_pos, align):
+    """
+    Determine where lies the current stage position
+    :param current_pos: (dict str->float) Current position of the stage
+    :param align: Sample stage that's being controlled
+    :return: (int) a value representing stage position from the constants LOADING, IMAGING, TILTED, COATING..etc
+    """
+    align_md = align.getMetadata()
```
Mahmood-B, line 107: A bit confused. Which component is "align"? Stage or objective aligner? I guess it is the objective aligner. If so, then i think it should not be "sample stage" in the docstring...

Mahmood-B (commented):
I reviewed the pr. Looks good generally. Only had few comments/questions

src/odemis/acq/move.py
lazem: In `CryoSwitchSamplePosition` there's check that all axes exist for these metadata.

src/odemis/acq/test/move_test.py
```diff
@@ -119,11 +120,44 @@ def test_loading_procedures(self):
...
+        f.result()
+        test.assert_pos_almost_equal(align.position.value, self.align_active,
+                                     atol=ATOL_LINEAR_POS)
+
+        # Get the stage to imaging position
+        f = cryoSwitchAlignPosition(ALIGNMENT)
+        f.result()
+        test.assert_pos_almost_equal(align.position.value, self.align_alignment,
+                                     atol=ATOL_LINEAR_POS)
+
```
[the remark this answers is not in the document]
lazem: Switching from alignment to loading and vice versa should be fine. Added test for unknown position.

src/odemis/acq/test/move_test.py
```diff
@@ -234,6 +261,39 @@ def test_get_progress(self):
         progress = getMovementProgress(current_point, start_point, end_point)
         self.assertIsNone(progress)
 
+    def test_get_current_aligner_position(self):
+        """
+        Test getCurrentPositionLabel function behaves as expected
+        """
```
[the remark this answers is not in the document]
lazem: When the system is started and the axes are not referenced, the position is rendered unknown. As the stages are referenced in the setup of the unit test class, the stage position could be either unknown or on the loading path.

src/odemis/acq/move.py
```diff
@@ -218,84 +347,64 @@ def _doCryoSwitchSamplePosition(future, stage, focus, align, target):
...
-            # 5. reference stage
+            # park aligner then reference the stage
             if not stage_referenced:
-                if not all(focus.referenced.value.values()):
-                    run_reference(future, focus)
-                run_sub_move(future, focus, focus_deactive)
-                if not all(align.referenced.value.values()):
-                    run_reference(future, align)
-                run_sub_move(future, align, align_deactive)
+                cryoSwitchAlignPosition(LOADING).result()
```
Kleijwegt, line 361: A small comment on what is happening here would be nice (put to loading position and wait until task is finished)?

src/odemis/acq/test/move_test.py
```diff
@@ -234,6 +262,59 @@ def test_get_progress(self):
...
+        aligner = self.aligner
+        # at start the aligner wouldn't be in one of the predefined positions
+        pos_label = getCurrentAlignerPositionLabel(aligner.position.value, aligner)
+        self.assertTrue(pos_label in (LOADING_PATH, UNKNOWN))
+        # Move to loading position
+        f = cryoSwitchAlignPosition(LOADING)
+        f.result()
+        pos_label = getCurrentAlignerPositionLabel(aligner.position.value, aligner)
+        self.assertEqual(pos_label, LOADING)
```
Kleijwegt, line 277: Wouldn't it be shorter and better to see what happens to put these test in a function and call the function for each part. You have now written more or less the same lines of code 5 times which makes it hard to follow the actual test (which is the Loading-->Immaging etc.)

src/odemis/acq/test/move_test.py
```diff
@@ -234,6 +262,59 @@ def test_get_progress(self):
...
+        pos_label = getCurrentAlignerPositionLabel(aligner.position.value, aligner)
+        self.assertEqual(pos_label, LOADING_PATH)
+
+        # simulate moving to unknown position by moving in opposite to deactive-active line
+        unknown_pos = copy.copy(self.align_active)
+        unknown_pos['y'] += 0.005
+        unknown_pos['z'] += 0.005
+        self.aligner.moveAbs(unknown_pos).result()
+        # moving to either imaging/alignment positions shouldn't be allowed
+        with self.assertRaises(ValueError):
```
Kleijwegt, line 292: Should you now also test that:

> pos_label = getCurrentAlignerPositionLabel(aligner.position.value, aligner)
> self.assertEqual(pos_label, UNKOWN)

src/odemis/acq/test/move_test.py
```diff
@@ -234,6 +262,59 @@ def test_get_progress(self):
...
+            f = cryoSwitchAlignPosition(IMAGING)
+            f.result()
```
Kleijwegt, line 294: Good test!

Kleijwegt (approved):

install/linux/usr/share/odemis/sim/enzel-sim.odm.yaml
```diff
@@ -242,175 +240,171 @@ ENZEL-Sim: {
...
-    },
-    metadata: {
-        POS_COR: {'z':  -0.003} , # stage Z - lens Z, when the lens Z is the closest (highest) from the stage (ie, at rng max)
-        # The "good focus" of the simulator is at z = 0 on the 3DoF stage.
-        # So on the focus, that's z = 0.0012.
-    },
-    persistent: {
-         metadata: [FAV_POS_ACTIVE],  # Store "good focus" position between run
-    },
+  class: actuator.MultiplexActuator,
```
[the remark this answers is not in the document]
lazem: Typically each component has a class to instantiate it by delegation. That said this is optional as some components are provided from another component. That's usually indicated by `children` field in the parent component. See an example for this in enzel's microscope "SEM XT connection" component, where it provide the ebeam control and focus as children.

install/linux/usr/share/odemis/sim/enzel-sim.odm.yaml
```diff
@@ -242,175 +240,171 @@ ENZEL-Sim: {
...
-        inverted: ["z"],
-    },
-    metadata: {
-        # "Safe" position to go to so that the 5DoF stage cannot hit the objective lens.
-        # Used after referencing.
-        # Z is used for the focus, when FM is not in use.
-        # X/Y can be ommitted, but should typically be the same as in Objective Aligner.
-        FAV_POS_DEACTIVE: {'x': -0.01, 'y': -0.01, 'z': -6.e-3},
+  class: smaract.MCS2,
+  role: null,
```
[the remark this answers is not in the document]
lazem: he role indicates the function of each component in the system. Some components don't need to have a role as they would be used (wrapped) by other components. For example, the 3DOF stage in enzel doesn't have a role as the lens aligner (which does have a role) wraps it and provides its axes in the proper manner.

PR #1846, opened 2021-08-09
driver-tmcm: bugfix canopen driver

philipwinklerdelmic:
Sometimes, moves did not start. The reason was that the "target reached" needs some time to be reset after a move has been requested. Otherwise, we would start a move and assume it's immediately finished because the bit was still set from the last move.


philipwinklerdelmic:
@Sabrina85 I think we can merge this ourselves if you want a clean master on EA1

Sabrina85 (approved):

PR #1847, opened 2021-08-09
fastem-gui: chamber tab improvements

philipwinklerdelmic:
* better icon for ebeam button
* chamber button: don't show pressure on the button, it's confusing

### DIFF
--- a/src/odemis/gui/cont/microscope.py
+++ b/src/odemis/gui/cont/microscope.py
@@ -94,13 +94,14 @@ class HardwareButtonController(object):
 class ChamberButtonController(HardwareButtonController):
     """ Controller that allows for the more complex state updates required by the chamber button """
 
-    def __init__(self, btn_ctrl, va, main_data, show_pressure=True):
+    def __init__(self, btn_ctrl, va, main_data, pressure_ctrl=None):
         """
 
         :param btn_ctrl: (ImageTextToggleButton) Button that controls and displays the chamber state
         :param va: (VigillantAttribute) The chamber state
         :param main_data: (MainGUIData) GUI microscope model
-        :param show_pressure (bool): whether to show the numerical pressure value on the button
+        :param pressure_ctrl (wxStaticText, None): control for showing the numerical pressure value.
+            If None, show the pressure on the button.
 
         """
         self.main_data = main_data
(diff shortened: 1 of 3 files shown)


src/odemis/gui/cont/microscope.py
```diff
@@ -138,6 +139,9 @@ def __init__(self, btn_ctrl, va, main_data):
         if model.hasVA(main_data.chamber, "pressure"):
             main_data.chamber.pressure.subscribe(self._on_pressure_change, init=True)
 
+        # TODO: support alternative pressure label (e.g. TextControl below button)?
```
Sabrina85, line 142: How do you show the pressure now, if not on the button?

philipwinklerdelmic: I don't show it, it's either pumped (green) or vented (black). Do you think it's a good idea to also show the numerical value?

Sabrina85: Ok. From the TODO I thought there is still somewhere the pressure shown, as it says "alternative pressure label". I don't think we show the pressure in the SPARC, but not sure. Easy check you can do with the simulator. Job would be the right person to say if we should display it or not. Maybe you can rephrase your TODO or delete it if Job says it is not needed.

philipwinklerdelmic: I updated the PR to show the pressure in a label below the buttons.

src/odemis/gui/cont/microscope.py
```diff
@@ -94,13 +94,14 @@ def Enable(self, enabled=True):
...
-    def __init__(self, btn_ctrl, va, main_data, show_pressure=True):
+    def __init__(self, btn_ctrl, va, main_data, pressure_ctrl=None):
         """
 
         :param btn_ctrl: (ImageTextToggleButton) Button that controls and displays the chamber state
         :param va: (VigillantAttribute) The chamber state
         :param main_data: (MainGUIData) GUI microscope model
-        :param show_pressure (bool): whether to show the numerical pressure value on the button
+        :param pressure_ctrl (wxStaticText, None): control for showing the numerical pressure value.
+            If None, show the pressure on the button.
```
Sabrina85, line 104: Shouldn't it be shown at all, when it is None? For the SecomStateController you had before that the pressure is simply not shown at all if I understood correctly. And now you will show the pressure on top of the button, which is not the standard behavior (just pump/vent labels).

We have now defined two behaviors:
1. The pressure is shown as text below the pump/vent button.
2. No pressure is shown, just the pump/vent button as usual.

philipwinklerdelmic: The behaviour before (for secom I guess?) was to show the pressure on top of the button, so I didn't want to change that and only added functionality for the FastEM without modifying the behaviour in other systems.

Sabrina85: Ok, I didn't know this. In your first version you had show_pressure = False, that is why I thought it should not be shown for the SecomStateController.

Sabrina85: I am now wondering how it works in the SPARC. There we also have a chamber...

philipwinklerdelmic:
Resolved conflicts, @pieleric can you review this?

PR #1848, opened 2021-08-11
fastem-gui-overview: bugfix time estimate label

philipwinklerdelmic:
If an acquisition is started successfully after a failed acquisition, the label still showed "Acquisition failed" during the entire process. Instead, after pressing the acquire button, it should return to showing the estimated time.

### DIFF
--- a/src/odemis/gui/cont/acquisition.py
+++ b/src/odemis/gui/cont/acquisition.py
@@ -1649,7 +1649,12 @@ class FastEMOverviewAcquiController(object):
     def check_acquire_button(self):
         self.btn_acquire.Enable(True if self._tab_data_model.selected_scintillators.value else False)
 
-    @wxlimit_invocation(1)  # max 1/s
+    # Don't call wxlimit_invocation, it causes problems with resetting the label after an acquisition failed
+    # in .on_acquisition(). This happens when the acquisition fails immediately (e.g. stage not referenced).
+    # Then, the failure message will not be shown since not enough time has passed from the previous label
+    # update (which in turn is necessary to make sure that the previous failure message disappears when a new
+    # acquisition is successfully started).
+    #@wxlimit_invocation(1)  # max 1/s
     def update_acquisition_time(self):
         lvl = None  # icon status shown
         if not self._main_data_model.active_scintillators.value:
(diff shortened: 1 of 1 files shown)


Sabrina85 (approved):
Looks fine to me.

src/odemis/gui/cont/acquisition.py
```diff
@@ -1649,7 +1649,12 @@ def _on_active_scintillators(self, evt):
     def check_acquire_button(self):
         self.btn_acquire.Enable(True if self._tab_data_model.selected_scintillators.value else False)
 
-    @wxlimit_invocation(1)  # max 1/s
+    # Don't call wxlimit_invocation, it causes problems with resetting the label after an acquisition failed
+    # in .on_acquisition(). This happens when the acquisition fails immediately (e.g. stage not referenced).
+    # Then, the failure message will not be shown since not enough time has passed from the previous label
+    # update (which in turn is necessary to make sure that the previous failure message disappears when a new
+    # acquisition is successfully started).
+    #@wxlimit_invocation(1)  # max 1/s
```
pieleric, line 1657: That sounds strange, because I don't think it matches how wxlimit_invocation should work. In particular, it should always eventually call the function with the last event (if there is no call within a second). But you seem to say that as the last call is too soon, so it's skipped.
For example, let's say there are calls at these times:
: (nothing for a while)
0
0.1
0.2
: (nothing for a while)

That should cause the call @0 to be immediately called, and @0.2 to be called at 1s. Only call @0.1 is skipped.

I'm a little careful with removing the `@wxlimit_invocation`, because that means that any setting change will trigger a GUI update, which may be a little slow. So that could bring a slow down in the whole GUI, whenever the user is interacting. Also, as a side note, be aware that the decorator also implies `@call_in_wx_main`, which is now also gone.

philipwinklerdelmic: I think what is happening is this:
1) on_acquisition starts
2) update_acquisition_time is started in a thread with limit_invocation decorator to display estimated time
3) acquisition fails
4) reset_acquisition_gui displays "Acquisition failed" message
5) the update_acquisition_time thread from earlier now displays the estimated time

So, the user doesn't see the "Acquisition failed" message.

A solution is to create a wrapper for the setLabel function which uses the wxlimit_invocation decorator. This wrapper is called both from update_acquisition time and reset_acquisition_gui.

PR #1849, opened 2021-08-17
fastem chamber controller: don't fail if chamber or ebeam power are n…

philipwinklerdelmic:
…ot available, remove buttons instead

* When the backend is started with a simsem simulator, there is no "power" VA on the ebeam, so the GUI would not start
* The same goes for yaml files without chamber components
* This functionality is not required for the majority of GUI operations, so we still want to be able to start it, especially for the simulator

### DIFF
--- a/src/odemis/gui/cont/microscope.py
+++ b/src/odemis/gui/cont/microscope.py
@@ -1459,23 +1459,31 @@ class FastEMStateController(object):
         self._tab_panel = tab_panel
 
         # E-beam management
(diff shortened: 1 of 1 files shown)


Sabrina85 (approved):

pieleric (approved):
There is a conflict, so it should be resolved before merging. Otherwise looks fine.

philipwinklerdelmic:
Resolved conflicts, ready to merge

PR #1850, opened 2021-08-18
driver-simcam: increase range of exposure time

philipwinklerdelmic:
The image "andorcam2-fake-spots.h5" which can be used for the simulator has an integration time of 19.6 s. This causes an error to be raised since the range only accepts values up to 10. This is a simulator, so it shouldn't be a problem to increase the range to 20s even though such long exposure times are maybe not common in practice.

### DIFF
--- a/src/odemis/driver/simcam.py
+++ b/src/odemis/driver/simcam.py
@@ -118,7 +118,7 @@ class Camera(model.DigitalCamera):
                                               cls=(int, long), setter=self._setTranslation)
 
         self._orig_exp = self._img.metadata.get(model.MD_EXP_TIME, 0.1)  # s
-        self.exposureTime = model.FloatContinuous(self._orig_exp, (1e-3, 10), unit="s")
+        self.exposureTime = model.FloatContinuous(self._orig_exp, (1e-3, 20), unit="s")
 
         # Some code care about the readout rate to know how long an acquisition will take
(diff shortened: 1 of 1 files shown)


Sabrina85 (approved):

src/odemis/driver/simcam.py
```diff
@@ -118,7 +118,7 @@ def clip_max_res(img_res):
                                               cls=(int, long), setter=self._setTranslation)
 
         self._orig_exp = self._img.metadata.get(model.MD_EXP_TIME, 0.1)  # s
-        self.exposureTime = model.FloatContinuous(self._orig_exp, (1e-3, 10), unit="s")
+        self.exposureTime = model.FloatContinuous(self._orig_exp, (1e-3, 20), unit="s")
```
pieleric, line 121: Instead of adjusting the code to fit a specific data, I'd suggest to make the code automatically fit, based on the current exposure time of the image. Something like this:
`self.exposureTime = model.FloatContinuous(self._orig_exp, range=(1e-3, max(10, self._orig_exp * 2)), unit="s")`

pieleric (changes requested):

src/odemis/driver/simcam.py
philipwinklerdelmic: that's a good one, i made the change

PR #1851, opened 2021-08-18
fastem: use different rotations for single and multi beam acqusition

philipwinklerdelmic:
The rotations will be specified as metadata to the e-beam component.

Is it ok to use default values or should we make the metadata required?

### DIFF
--- a/src/odemis/acq/fastem.py
+++ b/src/odemis/acq/fastem.py
@@ -27,6 +27,9 @@ from odemis.acq.stream import SEMStream
 import time
 
 
+SINGLE_BEAM_ROTATION_DEFAULT = 0  # 0 is typically a good guess (better than multibeam rotation of ~0.0157 rad)
+MULTI_BEAM_ROTATION_DEFAULT = 0.015707963  # 0.9 degrees
+
 # The executor is a single object, independent of how many times the module is loaded.
 _executor = model.CancellableThreadPoolExecutor(max_workers=1)
 
@@ -75,6 +78,16 @@ def acquire(roa, path):
     f = model.ProgressiveFuture()
     _executor.submitf(f, _run_fake_acquisition)
     return f
+
+
+def _configure_multibeam_hw(scanner):
+    md = scanner.getMetadata()
+    if model.MD_MULTI_BEAM_ROTATION in md:
+        scanner.rotation.value = md[model.MD_MULTI_BEAM_ROTATION]
+    else:
+        scanner.rotation.value = MULTI_BEAM_ROTATION_DEFAULT
+        logging.warning("Scanner doesn't have MULTI_BEAM_ROTATION metadata, using %s rad.",
+                        scanner.rotation.value)
 
 
 def _run_fake_acquisition():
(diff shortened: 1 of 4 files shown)


src/odemis/acq/fastem.py
```diff
@@ -197,6 +210,14 @@ def _configure_overview_hw(scanner):
...
                         scanner.resolution.value, TILE_RES)
 
+    md = scanner.getMetadata()
+    if model.MD_SINGLE_BEAM_ROTATION in md:
+        scanner.rotation.value = md[model.MD_SINGLE_BEAM_ROTATION]
+    else:
+        scanner.rotation.value = SINGLE_BEAM_ROTATION_DEFAULT
+        logging.warning("Scanner doesn't have SINGLE_BEAM_ROTATION metadata, using %s rad.",
+                        scanner.rotation.value)
+
```
pieleric, line 220: For the case of single beam, I am pretty sure you should also set scanner metadata MD_ROTATION_COR to -rot so that the image is still shown aligned in Odemis GUI.

philipwinklerdelmic: done

src/odemis/acq/fastem.py
```diff
@@ -27,6 +27,9 @@
 import time
 
 
+SINGLE_BEAM_ROTATION_DEFAULT = 0  # 0 is typically a good guess (better than multibeam rotation of ~0.0157 rad)
+MULTI_BEAM_ROTATION_DEFAULT = 0.015707963  # 0.9 degrees
```
pieleric, line 31: Is there any chance that this value is generic? If any system can have a sort of "random" number, then I'd suggest to just default to 0 for all the rotations.

src/odemis/acq/fastem.py
```diff
@@ -77,6 +80,16 @@ def acquire(roa, path):
...
 
+def _configure_multibeam_hw(scanner):
+    md = scanner.getMetadata()
+    if model.MD_MULTI_BEAM_ROTATION in md:
+        scanner.rotation.value = md[model.MD_MULTI_BEAM_ROTATION]
+    else:
+        scanner.rotation.value = MULTI_BEAM_ROTATION_DEFAULT
+        logging.warning("Scanner doesn't have MULTI_BEAM_ROTATION metadata, using %s rad.",
+                        scanner.rotation.value)
+
```
pieleric, line 91: I think that to display properly the thumbnails coming from the ASM, we should add a matching MD_ROTATION. This would either be done by the mdupdater (eg, the e-beam affects the multibeam, and so the rotation VA is copied into the MD_ROTATION of the multibeam), or just explicitly here: `multibeam.updateMetadata({MD_ROTATION: scanner.rotation.value}).`

pieleric (changes requested):

src/odemis/acq/fastem.py
philipwinklerdelmic: How do we access the multibeam here? I think it's not in the stream, so it would have to be an additional argument. I think it might be better to do it in the mdupdater, what do you think?

pieleric: Yes, that sounds like a good idea. The minimum requirement for the mdupdater is that there is explicitly on the "e-beam" an affects -> multibeam. I don't think that's currently the case, but it would make sense to have it anyway.

Then, it's just a matter of adding this special case in the mdupdater class.

philipwinklerdelmic:
I added the mdupdater code to update the multibeam metadata

src/odemis/odemisd/mdupdater.py
```diff
@@ -372,6 +374,22 @@ def updateMagnification(mag, comp_affected=comp_affected):
 
         return True
 
+    def observeEbeam(self, ebeam, comp_affected):
+        """Add ebeam rotation to multibeam metadata to make sure that the thumbnails
+        are displayed correctly."""
+
+        if not comp_affected.role == "multibeam":
```
pieleric, line 381: Let's just write `!=` :-)

pieleric (changes requested):
Just a very minor comment but otherwise it looks good to me. 
Also, it's a pity we don't have a ASM simulator... but let's not solve this here!

src/odemis/odemisd/mdupdater.py
philipwinklerdelmic: done

PR #1852, opened 2021-08-23
Fastem acquisition tab: diverse improvements

philipwinklerdelmic:
* bugfix ROC: couldn't be selected at high zoom level
* only enable calibration buttons for scinitallators selected in chamber tab
* use storage path from asm

### DIFF
--- a/src/odemis/gui/cont/acquisition.py
+++ b/src/odemis/gui/cont/acquisition.py
@@ -1396,10 +1396,8 @@ class FastEMAcquiController(object):
         self._calibration_ctrl = calibration_ctrl
 
         # Path to the acquisition
-        # <External storage>/<date>/<project name>/<roa name>
-        # TODO: display proper path to external storage (if mounted)
         self.path = datetime.today().strftime('%Y-%m-%d')
-        self._tab_panel.txt_destination.SetValue("<External Storage>/%s" % self.path)
+        self._tab_panel.txt_destination.SetValue(self.path)
 
         # ROA count
         self.roa_count = 0
@@ -1547,8 +1545,7 @@ class FastEMAcquiController(object):
         for p in self._tab_data_model.projects.value:
             ppath = os.path.join(self.path, p.name.value)
             for roa in p.roas.value:
-                fn = os.path.join(ppath, roa.name.value)
-                f = fastem.acquire(roa, fn)
+                f = fastem.acquire(roa, ppath)
                 f.add_done_callback(self.increase_acq_progress)
 
         f.add_done_callback(self.on_acquisition_done)
--- a/src/odemis/gui/cont/streams.py
+++ b/src/odemis/gui/cont/streams.py
@@ -3366,6 +3366,10 @@ class FastEMCalibrationController(object):
 
         for btn in self.panel.buttons.values():
             btn.Bind(wx.EVT_BUTTON, self._on_button)
+            btn.Enable(False)  # disabled by default, need to select scintillator in chamber tab first
+
+        # Only enable buttons for scintillators which have been selected in the chamber tab
+        tab_data.main.active_scintillators.subscribe(self._on_active_scintillators)
 
     def _on_button(self, evt):
         btn = evt.GetEventObject()
@@ -3382,6 +3386,14 @@ class FastEMCalibrationController(object):
             # Zoom to existing calibration region
             roc_ctrl = self.roc_ctrls[num]
         roc_ctrl.fit_view_to_bbox()
+
+    @call_in_wx_main
+    def _on_active_scintillators(self, evt):
+        for num, b in self.panel.buttons.items():
+            if num in self._tab_data.main.active_scintillators.value:
+                b.Enable(True)
+            else:
+                b.Enable(False)
 
 
 class FastEMROCController(object):
--- a/src/odemis/gui/cont/streams.py
+++ b/src/odemis/gui/cont/streams.py
@@ -3388,9 +3388,9 @@ class FastEMCalibrationController(object):
         roc_ctrl.fit_view_to_bbox()
 
     @call_in_wx_main
-    def _on_active_scintillators(self, evt):
+    def _on_active_scintillators(self, scintillators):
         for num, b in self.panel.buttons.items():
-            if num in self._tab_data.main.active_scintillators.value:
+            if num in scintillators:
                 b.Enable(True)
             else:
                 b.Enable(False)
--- a/src/odemis/gui/cont/streams.py
+++ b/src/odemis/gui/cont/streams.py
@@ -307,6 +307,12 @@ class StreamController(object):
 
         self.stream_panel.enable(True)
 
+    def pauseStream(self):
+        """ Pause (deactivate and stop updating) the stream """
+        if self.stream.should_update.value:
+            self.stream.is_active.value = False
+            self.stream.should_update.value = False
+
     def enable(self, enabled):
         """ Enable or disable all SettingEntries
         """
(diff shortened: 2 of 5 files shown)


src/odemis/gui/cont/acquisition.py
```diff
@@ -1573,6 +1575,11 @@ def on_acquisition_done(self, future):
             self._reset_acquisition_gui("Acquisition failed (see log panel).", level=logging.WARNING)
             return
 
+    def _on_storage_dir(self, sdir):
+        sdir = sdir.split('@')[-1]  # remove username/password
+        sdir = sdir.split('/', maxsplit=1)[-1]  # remove host
+        self._tab_panel.txt_destination.SetValue("%s/%s" % (sdir, self.path))
```
Kleijwegt, line 1581: Probably the setting of the external storage is already logged somewhere else. But in the case it isn't I think at least a debug/info log should be made.

src/odemis/gui/cont/acquisition.py
```diff
@@ -1397,9 +1397,11 @@ def __init__(self, tab_data, tab_panel, projectbar_ctrl, calibration_ctrl):
 
         # Path to the acquisition
         # <External storage>/<date>/<project name>/<roa name>
-        # TODO: display proper path to external storage (if mounted)
         self.path = datetime.today().strftime('%Y-%m-%d')
-        self._tab_panel.txt_destination.SetValue("<External Storage>/%s" % self.path)
+        if self._main_data_model.asm:
+            self._main_data_model.asm.externalStorageURL.subscribe(self._on_storage_dir, init=True)
+        else:
```
Sabrina85, line 1403: Do we still need the else statement for any case? We always have the ASM component in all backends and simulators right? And I think it is not possible to start the backend without specifying the directory on the external storage (you could do a quick try by commenting it out) as it is also read in the technolution driver for the .externalStorageURL VA.

src/odemis/gui/cont/acquisition.py
```diff
@@ -1397,9 +1397,11 @@ def __init__(self, tab_data, tab_panel, projectbar_ctrl, calibration_ctrl):
 
         # Path to the acquisition
         # <External storage>/<date>/<project name>/<roa name>
```
Sabrina85, line 1399: To have it more generic, how about <External storage> -> <Path to external storage> as it is not hardcoded anymore.

Sabrina85 (changes requested):
Looks fine, I some small things. Is it possible to also add a test case for checking that the path is updated, when the external storage image directory has been changed?

src/odemis/gui/cont/acquisition.py
philipwinklerdelmic: I was mostly concerned with the simulators, especially the ones in this repository under usr/bin/share/odemis/sim. We don't want to start the external ASM simulator here. But I agree that it's a bit annoying to have a special case just for this. How about I just add a comment that normally, there should be an ASM and the corner case is for the simulator?

philipwinklerdelmic:
I think it's a bit tricky to make a testcase for the label, at least I don't know any easy way. @Sabrina85

philipwinklerdelmic:
Updated with suggestions on storage path, added commit to pause overview stream when changing tabs.

src/odemis/gui/cont/acquisition.py
Sabrina85: True, I forgot about this file. It is the only one without an ASM simulator. Yes, maybe a comment is a good solution for now. I don't have a better idea for now.

Sabrina85:
> I think it's a bit tricky to make a testcase for the label, at least I don't know any easy way. @Sabrina85

I was thinking if we should check that when you update the externalStorageURL VA it is reflected in the path. But that would be more integration testing I guess and only works if the technolution driver is actually used (which needs the ASM simulator). So leave it :)

src/odemis/gui/cont/acquisition.py
```diff
@@ -1396,10 +1396,13 @@ def __init__(self, tab_data, tab_panel, projectbar_ctrl, calibration_ctrl):
...
         # Path to the acquisition
-        # <External storage>/<date>/<project name>/<roa name>
-        # TODO: display proper path to external storage (if mounted)
+        # <Path to External storage>/<date>/<project name>/<roa name>
         self.path = datetime.today().strftime('%Y-%m-%d')
-        self._tab_panel.txt_destination.SetValue("<External Storage>/%s" % self.path)
+        if self._main_data_model.asm:
+            self._main_data_model.asm.externalStorageURL.subscribe(self._on_storage_dir, init=True)
+        else:
+            # Should never happen on a real system, however, it's useful for the simulators
```
Sabrina85, line 1404: This is not completely true. The ASM simulator will return a path. Only the simulators without the asm component will need this hack, which at the moment is the one in usr/bin/share/odemis/sim. So I suggest to say, 
`# Should never happen on a real system, however, it's useful for simulators without an asm component`

philipwinklerdelmic: done

Sabrina85 (approved):
Looks good to me!

Sabrina85 (approved):

Sabrina85 (approved):

src/odemis/gui/cont/acquisition.py
```diff
@@ -1396,10 +1396,13 @@ def __init__(self, tab_data, tab_panel, projectbar_ctrl, calibration_ctrl):
...
-        # <External storage>/<date>/<project name>/<roa name>
-        # TODO: display proper path to external storage (if mounted)
+        # <Path to External storage>/<date>/<project name>/<roa name>
         self.path = datetime.today().strftime('%Y-%m-%d')
-        self._tab_panel.txt_destination.SetValue("<External Storage>/%s" % self.path)
+        if self._main_data_model.asm:
+            self._main_data_model.asm.externalStorageURL.subscribe(self._on_storage_dir, init=True)
+        else:
+            # Should never happen on a real system, however, it's useful for the simulators without an asm component
+            self._tab_panel.txt_destination.SetValue("<External Storage>/%s" % self.path)
```
Sabrina85, line 1405: We changed the behavior of the technolution wrapper a bit (there is currently an open pr for this). The .filename VA now contains the sub-directories on the external storage plus the megafield id (so all the adjustable part). The externalStorageURL VA now only contains the fixed part including the fixed directory "asm_service", which we cannot change. So I think we can simplify here and only display the date directory:
`self._tab_panel.txt_destination.SetValue(self.path)`
For now the user cannot change anything on this path, but that is ok. Then also the problem with the simulator is solved for now (so still possible to start the GUI without the asm component).

src/odemis/gui/cont/acquisition.py
```diff
@@ -1573,6 +1575,13 @@ def on_acquisition_done(self, future):
             self._reset_acquisition_gui("Acquisition failed (see log panel).", level=logging.WARNING)
             return
 
+    def _on_storage_dir(self, sdir):
+        sdir = sdir.split('@')[-1]  # remove username/password
+        sdir = sdir.split('/', maxsplit=1)[-1]  # remove host
+        sdir = "%s/%s" % (sdir, self.path)
+        logging.debug("New storage directory for image data: %s.", sdir)
+        self._tab_panel.txt_destination.SetValue(sdir)
```
pieleric, line 1583: For now, externalStorageURL is readonly, so this shouldn't be necessary. If useful to display the whole path on the FTP (I'm personally not convinced), you could listen to only asm.filename (which contains the subfolders and the file name).

src/odemis/gui/cont/streams.py
```diff
@@ -3383,6 +3403,14 @@ def _on_button(self, evt):
             roc_ctrl = self.roc_ctrls[num]
         roc_ctrl.fit_view_to_bbox()
 
+    @call_in_wx_main
+    def _on_active_scintillators(self, evt):
```
pieleric, line 3407: This is called when the active_scintillators VA changes, right? Then the argument is not "evt". It should be active_sincillators (or shorter), and you can use it, instead of reading the value again in the loop.

pieleric (changes requested):

src/odemis/gui/cont/acquisition.py
Sabrina85: Yes, I agree, this function can now go as we do not need to listen to the external storage anymore and the implementation as described above will be a lot simpler.

philipwinklerdelmic:
Pushed changes

Sabrina85 (approved):
The changes look fine to me.

pieleric (approved):
Looks good to me. Could you just squash you changes into 4 commits, then I can merge it as-is. (I could squash all together, but that'd be a pity as you already have separate commits and they are independent from each other)

philipwinklerdelmic:
done

PR #1857, opened 2021-08-31
FIB stream implementation

Kleijwegt:
This merge is needed to work on the alignment tab so that I can use the correct streams (FIB/SEM).
For testing the of the IP adress of the xt adapter is changed, before merging I will change this back.


install/linux/usr/share/odemis/hwtest/xt-only.odm.yaml
```diff
@@ -11,16 +11,23 @@ SEM: {
   class: xt_client.SEM,
   role: null,
   init: {
-    address: "PYRO:Microscope@192.168.31.150:4242",
+    address: "PYRO:Microscope@0.0.0.0:4242",
```
Kleijwegt, line 14: I will remove this change before merging

src/odemis/driver/test/xt_client_test.py
```diff
@@ -58,7 +58,7 @@
                 }
 CONFIG_FOCUS = {"name": "focuser", "role": "ebeam-focus"}
 CONFIG_DETECTOR = {"name": "detector", "role": "se-detector"}
-CONFIG_SEM = {"name": "sem", "role": "sem", "address": "PYRO:Microscope@192.168.31.162:4242",
+CONFIG_SEM = {"name": "sem", "role": "sem", "address": "PYRO:Microscope@0.0.0.0:4242",
```
Kleijwegt, line 61: I will remove these changes before merging

src/odemis/driver/test/xt_client_test.py
```diff
@@ -67,14 +67,14 @@
               }
 
 CONFIG_FIB_SCANNER = {"name": "fib-scanner", "role": "ion", "channel": "ion2"}
-CONFIG_FIB_SEM = {"name": "sem", "role": "sem", "address": "PYRO:Microscope@192.168.31.162:4242",
+CONFIG_FIB_SEM = {"name": "sem", "role": "sem", "address": "PYRO:Microscope@0.0.0.0:4242",
```
Kleijwegt, line 70: I will remove these changes before merging

src/odemis/driver/test/xt_client_test.py
```diff
@@ -67,14 +67,14 @@
...
                   "children": {"fib-scanner": CONFIG_FIB_SCANNER,
                                "stage": CONFIG_STAGE,
                                "detector": CONFIG_DETECTOR,
                                }
                   }
 
-CONFIG_DUAL_MODE_SEM = {"name": "sem", "role": "sem", "address": "PYRO:Microscope@192.168.31.162:4242",
+CONFIG_DUAL_MODE_SEM = {"name": "sem", "role": "sem", "address": "PYRO:Microscope@0.0.0.0:4242",
```
Kleijwegt, line 77: I will remove these changes before merging

install/linux/usr/share/odemis/hwtest/xt-only.odm.yaml
tepals: Now that all the tests are actually working with the adapter_sim, I think it actually makes sense to change the address to `0.0.0.0`. The `192.168.31.150` is needed for the simulator VM, but that the test cases are only occasionally run with the simulator VM.  But it might be better to do it in a separate PR. 
@pieleric do you agree? I can create a separate PR

src/odemis/driver/test/xt_client_test.py
pieleric: Normally this is already handled by setting the environment variable TEST_NOHW = "sim", so you shouldn't need to change the source code.

install/linux/usr/share/odemis/hwtest/xt-only.odm.yaml
pieleric: This is in "hwtest" directory, normally used for testing on real hardware. So I'd suggest to leave it as-is. You could add a comment to indicate that it should be replaced by "localhost:4242" for the simulator.
BTW, I wonder why "0.0.0.0" works. It's not really a valid address.
BTW2, maybe we could skip the "PYRO:Microscope@". It seems to leak some internal implementation details. We could automatically add it when building the URL.

src/odemis/gui/cont/tabs.py
```diff
@@ -988,6 +999,13 @@ def _create_views(self, main_data, viewports):
                     v["fov_hw"] = main_data.ebeam
                     vp.canvas.fit_view_to_next_image = False
 
+        if main_data.fib_scanner:
+            vpv[viewports[1]]['stream_classes'] = FIBStream
+            # Remove stage control since the pixel size/magnification is unknown for the FIB.
+            vpv[viewports[1]].pop('stage')
+            if CAN_DRAG in viewports[1].canvas.abilities:
+                viewports[1].canvas.disable_drag()
```
pieleric, line 1007: Let's not worry about this. Firstly, because moving the stage has actually effect on the FIB view, so it could still be useful (just not in the case of the ENZEL). Most importantly, that's just for internal testing, so let's not worry about it, and keep the code as simple as possible.

src/odemis/gui/cont/tabs.py
```diff
@@ -858,6 +858,17 @@ def __init__(self, name, button, panel, main_frame, main_data):
             else:
                 logging.error("No EM detector found")
 
+        if hasattr(tab_data, 'fibState'):
```
pieleric, line 861: Replace by `if tab_data.ion_beam:`

src/odemis/gui/cont/tabs.py
```diff
@@ -73,12 +73,12 @@
...
     PixelTemporalSpectrumProjection, SinglePointTemporalProjection, \
     ScannedTemporalSettingsStream, \
-    ARRawProjection, ARPolarimetryProjection, StaticStream, LiveStream
+    ARRawProjection, ARPolarimetryProjection, StaticStream, LiveStream, FIBStream
 from odemis.acq.move import LOADING, IMAGING, MILLING, COATING, UNKNOWN, LOADING_PATH, target_pos_str
 from odemis.acq.move import cryoSwitchSamplePosition, cryoTiltSample, getMovementProgress, getCurrentPositionLabel
 from odemis.util.units import decompose_si_prefix, readable_str
 from odemis.driver.actuator import ConvertStage
-from odemis.gui.comp.canvas import CAN_ZOOM
+from odemis.gui.comp.canvas import CAN_ZOOM, CAN_DRAG
```
pieleric, line 81: After the recommended changes, this can go away.

pieleric (changes requested):

install/linux/usr/share/odemis/hwtest/xt-only.odm.yaml
Kleijwegt: >"hwtest" directory, normally used for testing on real hardware. 
I don't think that IP adress corresponds with any of the real hardware, I think both FUMO-Beta (EA when in Delft Zuid) and Helios have another IP adress

src/odemis/gui/model/__init__.py
```diff
@@ -183,6 +183,7 @@ class MainGUIData(object):
         "tc-filter": "tc_filter",
         "slit-in-big": "slit_in_big",
         "sample-thermostat": "sample_thermostat",
+        "ion": "fib_scanner",
```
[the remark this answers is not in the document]
Kleijwegt: in the XT client we chose to name the child fib_scanner. I don't mind naming it here ion_beam but just so that you know.

src/odemis/model/_metadata.py
```diff
@@ -91,6 +91,7 @@
 MD_AT_SPECTRUM = "Spectrum"
 MD_AT_AR = "Angle-resolved"
 MD_AT_EM = "Electron microscope"
+MD_AT_FIB = "focused ion beam"
```
pieleric, line 94: Let's follow the other strings, and put an uppercase on the first word "Focused".

install/linux/usr/share/odemis/hwtest/xt-only.odm.yaml
pieleric: Still, 0.0.0.0 is the worst as it is not correct. I'm fine if you replace with "localhost", or if you leave the old IP.
Note that the fact you're adding the fib-scanner means that it'll probably fail to start with a VM too, as it doesn't simulate a FIB.

pieleric (changes requested):
Looks fine to me. Just a few very minor comments.

install/linux/usr/share/odemis/hwtest/xt-only.odm.yaml
Kleijwegt: @tepals Maybe for another PR we/you can think about what IP adress we want to actually write down in the test cases of xt_client_test? For now I resolve this issue

tepals (approved):

src/odemis/acq/stream/_live.py
```diff
@@ -296,13 +300,17 @@ def _applyROI(self):
         Note: should only be called when active (because it directly modifies
           the hardware settings)
         """
-        res, trans = self._computeROISettings(self.roi.value)
+        need_res = True if hasVA(self._emitter, "resolution") and not self._emitter.resolution.readonly else False
+        need_trans = True if hasVA(self._emitter, "translation") and not self._emitter.translation.readonly else False
```
pieleric, line 304: The `v = True if condition else False` is a bit verbose, to the point where I'm finding it a little bit confusing because I'd expect such a complex writing is a sign of a special trick, but there is no such trick.
Typically, you'd just write `need_res = hasVA(self._emitter, "resolution") and not self._emitter.resolution.readonly`.
If you find it confusing, you could also try `need_res = bool(hasVA(self._emitter, "resolution") and not self._emitter.resolution.readonly)`, what do you think?

src/odemis/acq/stream/_live.py
```diff
@@ -365,17 +373,40 @@ def _onActive(self, active):
...
+            kwargs["acq_type"] = model.MD_AT_EM
+        super().__init__(name, detector, dataflow, emitter, **kwargs)
+
+
+class FIBStream(ScannerStream):
+    def __init__(self, name, detector, dataflow, emitter, **kwargs):
+        if "acq_type" not in kwargs:
+            kwargs["acq_type"] = model.MD_AT_FIB
+        super().__init__(name, detector, dataflow, emitter, **kwargs)
+        self.scale = (1.0, 1.0)  # Explicitly define the scale which isn't present as a VA on the emitter
```
pieleric, line 409: We should be able to remove this line, now that the missing scale attribute is handled.

src/odemis/acq/stream/_live.py
```diff
@@ -284,7 +281,14 @@ def _computeROISettings(self, roi):
         # translation is distance from center (situated at 0.5, 0.5), can be floats
         trans = (shape[0] * (center[0] - 0.5), shape[1] * (center[1] - 0.5))
         # resolution is the maximum resolution at the scale in proportion of the width
-        scale = self._getEmitterVA("scale").value
+        try:
+            scale = self._getEmitterVA("scale").value
+        except AttributeError:
+            if hasattr(self, "scale"):
+                scale = self.scale
+            else:
```
pieleric, line 289: No need, just use the `else` case all the time.

pieleric (changes requested):
In addition to my small comments, could you also rebase against master, as there is apparently a conflict.

src/odemis/acq/stream/_live.py
Kleijwegt: I linke that implementation actually. I didn't think of it.
I have chosen for this one:
`need_res = hasVA(self._emitter, "resolution") and not self._emitter.resolution.readonly  # Boolean`

src/odemis/acq/stream/_live.py
Kleijwegt: Isn't it nice to still define it even if it isn't essential? So it is clear for whomever uses the FIB class where scale comes from (and that it isn't actually defined) and it also prevents quite some debug messages.
What do you think?

src/odemis/acq/stream/_live.py
```diff
@@ -425,8 +437,11 @@ def guessFoV(self):
             # control it from Odemis. In this case, the user has to type the
             # current magnification in, which directly sets the pixelSize.
             # The stream never has a local version of any of them.
-            base_pxs = self.emitter.pixelSize.value  # It's read-only and always present on the emitter
-            hpxs = base_pxs[0] * scale[0]
+            if hasVA(self.emitter, "pixelSize"):
+                base_pxs = self.emitter.pixelSize.value  # It's read-only and present on almost every emitter
+                hpxs = base_pxs[0] * scale[0]
+            else:
+                raise AttributeError("Failed to estimate the field-of-view, the emitter has no pixelSize VA.")
```
Kleijwegt, line 444: @pieleric I have added the things to deal with guessFoV in a separate commit. These lines are basically the only change. After that we could merge it. (I haven't tested it on Helios yet but I don't think that is necessary)

pieleric:
@Kleijwegt , you've resolved all my comments, but didn't push any more changes. Could you push them, so that we can merge this PR?

Kleijwegt:
Sadly, I didn't have time to refine my commit and clean it up completly because I don't have time on Enzel this sprint

PR #1859, opened 2021-09-02
technolution driver: add calculation of line scan time method

Sabrina85:


tepals (commented):
two small comments

tepals (approved):

src/odemis/driver/test/technolution_test.py
```diff
@@ -51,7 +51,7 @@
 logging.getLogger().setLevel(logging.DEBUG)
 
 # Export TEST_NOHW = 1 to prevent using the real hardware
-TEST_NOHW = (os.environ.get("TEST_NOHW", "0") != "0")  # Default to Hw testing
+TEST_NOHW = (os.environ.get("TEST_NOHW", "0") != "1")  # Default to Hw testing
```
tepals, line 54: Now you changed it to no hardware testing by default: if TEST_NOHW does not exist `os.environ.get("TEST_NOHW", "0") = "0"` and `("0" != "1") = True`

Sabrina85: Indeed that should not be like this. I forgot to change it back.

src/odemis/driver/technolution.py
```diff
@@ -1015,6 +991,11 @@ def __init__(self, name, role, parent, **kwargs):
 
         # The minimum of the cell resolution cannot be lower than the minimum effective cell size.
         self.cellCompleteResolution = model.ResolutionVA((900, 900), ((12, 12), (1000, 1000)))
+        self.cellCompleteResolution.subscribe(self._updateFrameDuration)
+        self.parent._ebeam_scanner.dwellTime.subscribe(self._updateFrameDuration)
+        # acquisition time for a single field image including overscanned pixels and flyback time
+        acq_time = self.getTotalFieldScanTime()
```
pieleric, line 997: Move this *after* creating `.frameDuration`, as  `_updateFrameDuration()` needs this attribute (so it's a little clearer). Also, you can avoid computing `acq_time` explicitly: initialize at 0, and in add an `init=True` in one of the subscribers.

pieleric (changes requested):

PR #1860, opened 2021-09-03
technolution wrapper: some updates and fixes

Sabrina85:
* illegal character -> actually allowed characters
* fix bug in characters: capital characters are also allowed for megafield id according to API
* add method to add subdirectories on external storage (such as project and date folders) - This is needed by the GUI, which will automatically create a folder with the current date to store all data. Additonally, the project name is stored on this VA by the GUI.
* add testcases for subdirectory
* raise ValueError when wrong characters are detected - let the caller decide how to handle the error
* raise ValueError for incorrect input when checking if a megafield exists. Otherwise, it is hidden whether now the megafield does not exist or just a wrong input was provided.
* improve test cases for ftp url: make sure every if statement is actuall triggered

I removed one of the checks for the ftp url - the general check on invalid characters. I couldn't find an example that triggers only this if statement, but none of the others. If I missed a case I can add it back.

### DIFF
--- a/src/odemis/driver/technolution.py
+++ b/src/odemis/driver/technolution.py
@@ -163,7 +163,7 @@ class AcquisitionServer(model.HwComponent):
                                                   externalStorage["host"],
                                                   externalStorage["directory"]),
                                                  setter=self._setURL, readonly=True)
-        self.externalStorageURL._set_value(self.externalStorageURL.value, force_write=True)
+        self.externalStorageURL._set_value(self.externalStorageURL.value, force_write=True)  # check URL ok
 
         # VA to switch between calibration and acquisition mode (megafield acquisition)
         self.calibrationMode = model.BooleanVA(False, setter=self._setCalibrationMode)
--- a/src/odemis/driver/technolution.py
+++ b/src/odemis/driver/technolution.py
@@ -56,11 +56,12 @@ RUNNING = "installation in progress"
 FINISHED = "last installation successful"
 FAILED = "last installation failed"
 
-ASM_USER_CHARS = r'[^A-Za-z0-9]'
-ASM_PASSWORD_CHARS = r'[^A-Za-z0-9]'
-ASM_HOST_CHARS = r'[^A-Za-z0-9.]'
-ASM_PATH_CHARS = r'[^A-Za-z0-9/_()-]'
-ASM_FILE_CHARS = r'[^A-Za-z0-9_()-]'
+ASM_USER_CHARS = r'[A-Za-z0-9]+'  # + -> should be at least one character
+ASM_PASSWORD_CHARS = r'[A-Za-z0-9]+'
+ASM_HOST_CHARS = r'[A-Za-z0-9.]+'
+ASM_PATH_CHARS = r'[A-Za-z0-9/_()-]+'
+ASM_SUBDIR_CHARS = r'[A-Za-z0-9/_()-]*'  # * -> subdirectories can also be empty string
+ASM_FILE_CHARS = r'[A-Za-z0-9_()-]+'
 
 
 def convertRange(value, value_range, output_range):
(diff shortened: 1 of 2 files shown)


src/odemis/driver/technolution.py
```diff
@@ -323,27 +333,26 @@ def system_checks(self):
...
+        :return: (bool) True if the mega field exists on external storage. False if it does not exist yet.
+        :raises: (ValueError) When the input arguments contain invalid characters.
         """
-        ASM_FILE_ILLEGAL_CHARS = r'[^a-z0-9_()-]'
-        ASM_PATH_ILLEGAL_CHARS = r'[^A-Za-z0-9/_()-]'
-        if re.search(ASM_PATH_ILLEGAL_CHARS, storage_dir):
-            logging.error("The specified storage directory contains invalid characters, cannot check if mega field "
-                          "exists (only the characters '%s' are allowed)." % ASM_FILE_ILLEGAL_CHARS[2:-1])
-            return False
+        if re.search(ASM_PATH_CHARS, path):
```
pieleric, line 345: The `ASM_PATH_CHARS` variable actually looks for characters which are not allowed. That's what the "^" is for. That's why the code raises an error in case there is a match. So the previous names with "ILLEGAL" were clearer. That said, I think it'd be even clearer to change the logic  by checking that *every* character is legal. You could write it like this:
```python
ASM_PATH_CHARS = r'[A-Za-z0-9/_()-]*'
ASM_FILE_CHARS = r'[A-Za-z0-9_()-]+' # + instead of * to force at least one character
if not re.fullmatch(ASM_PATH_CHARS, path):
    raise ...
:
```

src/odemis/driver/technolution.py
```diff
@@ -534,60 +543,48 @@ def _setCalibrationMode(self, mode):
...
+        if re.search(ASM_PATH_CHARS, url_parser.path):
+            raise ValueError("Path %s contains invalid characters. Only the following characters are allowed: "
+                             "'%s'." % (url_parser.path, ASM_PATH_CHARS[2:-1]))
 
         # TODO: Commented out because not present on EA
         # self.asmApiPostCall("/config/set_external_storage?host=%s&user=%s&password=%s" %
         #                     (urlparse(self.externalStorageURL.value).hostname,
         #                      urlparse(self.externalStorageURL.value).username,
         #                      urlparse(self.externalStorageURL.value).password), 204)
+
```
pieleric, line 587: I had never noticed, but it's actually not possible to change any of these values... So let's just mark it so by having the externalStorageURL VA as "read-only". You'd still need to validate the input, for the initial value, but the setter would never need to be called afterwards.

src/odemis/driver/technolution.py
```diff
@@ -1070,10 +1068,12 @@ def _assembleMegafieldMetadata(self):
         X_descan_setpoints = self._descanner.getXAcqSetpoints()
         Y_descan_setpoints = self._descanner.getYAcqSetpoints()
 
+        storage_directory = os.path.join(urlparse(self.parent.externalStorageURL.value).path, self.subdirectory.value)
```
pieleric, line 1071: Drop this. Just `storage_directory = self.subdirectory.value`. Then, it'll be possible to indicate the whole base URL in externalStorageURL.

pieleric (changes requested):

src/odemis/driver/technolution.py
Sabrina85: In the API we have this call, which is not implemented yet. 
```
  /config/set_external_storage:
    post:
      tags: ["config"]
      summary: "Configure the external storage FTP server. [NOT IMPLEMENTED] Is configured via configuration file on the ASM."
      description: "Configure the external FTP server for storage of the mega field on the server by providing the server address and necessary login credentials."
      ....
```
Do  you mean that we then never ever will make use of it? And that there is no need to implement this at all also in future? Why was it in the design of the API in first place?

pieleric: I think that if the API becomes available, then we can remove the "read-only" flag from the VA (in addition to actually calling the API function). 
Probably the intention is to allow the user choose a different storage sever without having to modify the ASM config (which only a service engineer can do). I'm fine with this not being implemented yet, but it's probably good to keep it mind and not wait too long to ask Technolution to implement it.
For now, if the value cannot be changed, it's better to indicate it, by using the "readonly" flag on the VA.

src/odemis/driver/technolution.py
```diff
@@ -144,14 +155,15 @@ def __init__(self, name, role, host, children, externalStorage, **kwargs):
...
+        # fixed url of the external storage configuration
         self.externalStorageURL = model.StringVA('ftp://%s:%s@%s/%s' %
                                                  (externalStorage["username"],
                                                   externalStorage["password"],
                                                   externalStorage["host"],
                                                   externalStorage["directory"]),
-                                                 setter=self._setURL)
-        self._setURL(self.externalStorageURL.value)  # Check and set the external storage URL to the ASM.
+                                                 setter=self._setURL, readonly=True)
+        self.externalStorageURL._set_value(self.externalStorageURL.value, force_write=True)
```
Kleijwegt, line 166: I do think the comment on why setting a value which has been set in the line before is helping here.

Kleijwegt (approved):

src/odemis/driver/test/technolution_test.py
```diff
@@ -1249,7 +1273,7 @@ def setUp(self):
         self.MPPC.dataContent.value = "empty"
 
         # Change megafield id to prevent testing on existing images/overwriting issues.
-        self.MPPC.filename.value = time.strftime("testing_megafield_id-%Y-%m-%d-%H-%M-%S")
+        self.MPPC.filename.value = time.strftime("test_images/project/project/testing_megafield_id-%Y-%m-%d-%H-%M-%S")
```
pieleric, line 1276: Is it on purpose that you have a double "project/"?

pieleric (approved):

src/odemis/driver/test/technolution_test.py
Sabrina85: It is fixed now

PR #1861, opened 2021-09-06
driver-piezomotor: change sleep time to 0.1, so timeout will not be t…

philipwinklerdelmic:
…riggered in simulator

* Timeout is calculated based on move length, small moves are not expected to take a second.

### DIFF
--- a/src/odemis/driver/piezomotor.py
+++ b/src/odemis/driver/piezomotor.py
@@ -955,6 +955,8 @@ class PMDSimulator(object):
         self.status = "0000"
         self.indexing = True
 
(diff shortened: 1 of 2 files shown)


pieleric (changes requested):
The code always waits "dur * 2 + 1", so it should be always > 1s. How come you hit such case? Is that for very small moves, and the overhead of the thread is more expensive that "dur * 2"?

The problem with this change is that the simulator is going to be even worse for large moves (eg, simulating an overview acquisition).
I'd suggest to "bite the bullet" and extend the function to accept as argument a distance and speed. Computing the duration is then easy.

PR #1863, opened 2021-09-13
scripts: new ar_to_csv script

pieleric:
This is a script to convert raw AR data to Theta/Phi data in CSV format.

Requested by a user, but also a nice example on how to open data, and
export it.

### DIFF
--- a/scripts/ar_to_csv.py
+++ b/scripts/ar_to_csv.py
@@ -1,0 +1,126 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+'''
+Created on 13 Sep 2021
+
+@author: Éric Piel
+
+This is a script to convert raw AR data to Theta/Phi data in CSV format
+
+run as:
+python ./scripts/ar_to_csv.py --background background.h5 ar_files_*.h5
+
+It will generate one CSV file per pixel of each AR stream.
+'''
+
+import argparse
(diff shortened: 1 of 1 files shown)


scripts/ar_to_csv.py
```diff
@@ -0,0 +1,126 @@
...
+                filenames.extend(glob.glob(fn))
+        else:
+            filenames = options.filenames
+
+        for fn in filenames:
+            export_ar_to_csv(fn, bkg)
+
+    except:
+        logging.exception("Unexpected error while performing action.")
+        return 127
```
Mahmood-B, line 119: Log the exception as well. It would be useful to understand more about the error in case it is raised.  
```
except Exception as exp:
        logging.exception("Unexpected error while performing action: %s" %exp)
        return 127
```

Mahmood-B (changes requested):
I do not have much to say. Looks find in general

scripts/ar_to_csv.py
pieleric: `logging.exception()` is special. It automatically appends the exception message and the whole backtrace. So it's not need to indicate the exception message explicitly.

Mahmood-B: Good to know

Mahmood-B (approved):

PR #1864, opened 2021-09-13
gui cont streams: don't fail to start/stop stream if no spotStream

pieleric:
Some plugins may use the SPARC stream, without having a spotStream.
Playing the stream for now causes an error. Instead, allow it, even if
there is not active SEM spot, so it's not super useful, it's better
than failing.

### DIFF
--- a/src/odemis/gui/cont/streams.py
+++ b/src/odemis/gui/cont/streams.py
@@ -1959,11 +1959,12 @@ class StreamBarController(object):
         # => need more ways to change current stream (at least pick one from the
         # current view?)
 
+        # Get the spot Stream, if the model has one, otherwise None
+        spots = getattr(self._tab_data_model, "spotStream", None)
         # Don't mess too much with the spot stream => just copy "should_update"
-        if hasattr(self._tab_data_model, "spotStream"):
-            if stream is self._tab_data_model.spotStream:
-                stream.is_active.value = updated
-                return
+        if stream is spots:
+            stream.is_active.value = updated
+            return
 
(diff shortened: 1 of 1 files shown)


src/odemis/gui/cont/streams.py
```diff
@@ -1960,10 +1960,10 @@ def _onStreamUpdate(self, stream, updated):
...
 
         # Don't mess too much with the spot stream => just copy "should_update"
-        if hasattr(self._tab_data_model, "spotStream"):
-            if stream is self._tab_data_model.spotStream:
-                stream.is_active.value = updated
-                return
+        spots = getattr(self._tab_data_model, "spotStream", None)
+        if stream is spots:
+            stream.is_active.value = updated
+            return
```
VictoriaDel, line 1966: Maybe a comment similar to the title of the pr, like "stream is still updated if there is no spotStream" would make it a bit clearer to me :)

src/odemis/gui/cont/streams.py
```diff
@@ -2040,9 +2038,8 @@ def _onStreamUpdate(self, stream, updated):
             self._tab_data_model.streams.value = l
         else:
             # Deactivate spot mode if spot stream was just paused
```
VictoriaDel, line 2040: What if spot stream is not present? Could we rephrase it?

VictoriaDel (changes requested):
It looks fine to me!! I would just like if there are a few comments explaining why you actually updated the code.

VictoriaDel (approved):

PR #1865, opened 2021-09-14
Ci testing

tepals:


.github/workflows/python-app.yml
```diff
@@ -0,0 +1,23 @@
+# This workflow will install Python dependencies, run tests and lint with a single version of Python
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
```
pieleric, line 2: Please update this comment to the actual behaviour of this command.

.github/workflows/python-app.yml
```diff
@@ -0,0 +1,23 @@
...
+
+name: Python application
```
pieleric, line 4: Let's pick a better name for this pipeline. Something like "Unit testing"

.github/workflows/python-app.yml
```diff
@@ -0,0 +1,23 @@
...
+    runs-on: ubuntu-18.04
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install dependencies
+      run: |
+        python3 -m pip install --upgrade pip
+        pip3 install -r requirements-py3.txt
+        sudo add-apt-repository ppa:delmic-soft/odemis
+        sudo apt install python-pyro4-delmic
```
pieleric, line 20: Preferably, I'd go for *only* the use of apt, instead of a mix pip+apt. 

I wonder if "sudo add-apt-repository ppa:delmic-soft/odemis" has any chance to work, as usually it asks you to confirm, which you cannot. Either find a way to avoid the question, or use the "low level style" which consist in adding a line in `/etc/apt/sources.list` (followed by `apt update`).
Also, for the apt line, you probably have to add a "-y". I think it's also recommended to keep using "apt-get" in scripts, as the command line interface is more stable.

requirements-py3.txt
```diff
@@ -2,16 +2,16 @@ Cython>=0.23.5
...
+#opencv>=2.4.11,<4.0
+#Pillow>=3.2.0
+#pycairo>=1.10.0
 PyInstaller>=3.6
 psutil>=4.3.0
 PyYAML>=3.11
 pyzmq>=15.2.0
 scipy>=0.15.1
-wxPython>=4.0.4
+#wxPython>=4.0.4
```
pieleric, line 16: All these dependencies are needed by Odemis, so there is no good reason to comment them out.

pieleric (changes requested):

tepals:
I've updated the code and got the tests in odemis.model.test running. The tests from `odemis.model.test.remote_test` give the following error: `OSError: Container creation timeout`, do you know if this has anything to do with running the tests in a VM? You can see the full log here: https://github.com/tepals/odemis/runs/3743791901?check_suite_focus=true

pieleric:
See this message:
```
ERROR:root:Directory /var/run/odemisd is not accessible, which is needed for creating the container testcont
```
This is actually an issue that we also have when we run the test case on a system which has never started Odemis yet (since reboot). You can recreate it with `sudo rm -rf /var/run/odemisd`. 

Normally, at startup, "odemisd" takes care of creating that directory (with the right permissions). So I think a simple way to take care of it is to automatically create the directory while setting up the VM, with the right ownership (ie, accessible in RW to the standard user). 
A more proper way, which will probably have to do one day, would be to support changing the odemis.model.BASE_DIRECTORY to something else at init. This would have the advantage of allowing multiple odemis backends simultaneously.

pieleric (changes requested):

.github/workflows/python-app.yml
```diff
@@ -0,0 +1,42 @@
...
+      run: |
+        export PYTHONPATH="/home/runner/work/odemis/odemis/src:$PYTHONPATH"
+        python3 -m unittest src/odemis/model/test/components_test.py
+        python3 -m unittest src/odemis/model/test/dataflow_test.py
+        python3 -m unittest src/odemis/model/test/futures_test.py
+        python3 -m unittest src/odemis/model/test/vattributes_test.py
+
+    - name: Run linting for xt_client driver
+      run: |
+        python3 -m flake8 src/odemis/driver/xt_client.py --max-line-length=120
```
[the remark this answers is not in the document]
tepals: I haven't figured out how to run linting only on changed files. We can also completely remove linting, until we've figured out how to have it only on changed/added lines.

src/odemis/driver/xt_client.py
```diff
@@ -1169,7 +1169,7 @@ def __init__(self, name, role, parent, hfw_nomag, channel="electron1", has_detec
             resolution = tuple(self.parent.get_resolution())
             res_choices = set(r for r in RESOLUTIONS
                               if (rng["x"][0] <= r[0] <= rng["x"][1] and rng["y"][0] <= r[1] <= rng["y"][1])
-                             )
+                              )
```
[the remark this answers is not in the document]
tepals: Agreed this shouldn't be here

.github/workflows/python-app.yml
```diff
@@ -0,0 +1,38 @@
...
+
+        sudo mkdir /var/run/odemisd
+        sudo chmod a+rw /var/run/odemisd
+
+    - name: Run tests from odemis.model
+      run: |
+        export PYTHONPATH="/home/runner/work/odemis/odemis/src:$PYTHONPATH"
+        python3 -m unittest src/odemis/model/test/components_test.py
+        python3 -m unittest src/odemis/model/test/dataflow_test.py
+        python3 -m unittest src/odemis/model/test/futures_test.py
```
tepals, line 36: @pieleric I just noticed that when one of these lines has a failure it doesn't execute the next line. There was a failure in futures_test, and it didn't execute remote_test and vattributes_test. Should I put it in separate run blocks? If we have to put it in separate run blocks the file will eventually become quite chaotic and long.

pieleric: Yes, I think that's the right approach.
Possibly, in the future we'd have a special version of the util/runtest.sh script to run in a pipeline (just the tests which are possible).

.github/workflows/python-app.yml
```diff
@@ -0,0 +1,34 @@
...
+        python3-decorator python3-xdg python3-suds python3-matplotlib python3-opencv \
+        python3-wxgtk4.0 python3-serial python3-evdev python3-notify2 python3-psutil \
+        graphicsmagick cgroup-bin authbind python3-setuptools cython python3-pyro4-delmic
+
+        sudo mkdir /var/run/odemisd
+        sudo chmod a+rw /var/run/odemisd
+
+    - name: Run tests from odemis.model
+      run: |
+        export PYTHONPATH="/home/runner/work/odemis/odemis/src:$PYTHONPATH"
```
pieleric, line 33: Is `/home/runner/work/odemis/odemis` the current directory? In Jenkins, the VM initial directory is the directory containing the project to test. I wouldn't be surprised it's the same on github.
If so, it might be more portable to write something like `PYTHONPATH="$PWD/src:$PYTHONPATH"`

pieleric (changes requested):

PR #1868, opened 2021-09-17
GenericComponent

Helsloot:


src/odemis/driver/simulated.py
```diff
@@ -339,3 +238,195 @@ def __init__(self, name, role, positions, has_pressure=False, **kwargs):
...
+
+            self._executor = model.CancellableThreadPoolExecutor(max_workers=1)
+
+            # RO, as to modify it the client must use .moveRel() or .moveAbs()
+            self.position = model.VigilantAttribute({}, unit="m", readonly=True)
+
+            self.speed = model.MultiSpeedVA(init_speed, (0., 10.), "m/s")
+
+        model.Actuator.__init__(self, name, role, axes=axes_def, **kwargs)
+
```
Helsloot, line 343: I attempted to delete the actuator specific methods and attributes here if no axes were defined, but this proved harder than expected.
Methods cannot be removed, since those are class specific, not instance specific. So they are either present for all instances or for none.
Attempting to delete attributes like self.axes, resulted in an AttributeError: cant' delete attribute.
I wasn't sure if deleting these attributes was worth diving deeper into, to track the origin of this exception and circumvent it somehow. (A quick google search yielded no result.)

Kleijwegt: Can you not overwrite a method with a function which returns a "NotImplementedError" in case these methods are really annoying? Or overwrite the method with a variable containing None?

I also found this solution, which raises an error whenever the method/attribute foo is called.
```
    def __getattr__(self, attr_name):
        if attr_name == 'foo': raise AttributeError
        return getattr(self._wrapped, attr_name)
```
From: https://stackoverflow.com/questions/231839/python-inheritance-how-to-disable-a-function

However it seems to be all a non-pythonic solution. For a nice implementation you should probably create a different parent class which doesn't contain that method....

src/odemis/driver/simulated.py
```diff
@@ -339,3 +238,195 @@ def __init__(self, name, role, positions, has_pressure=False, **kwargs):
...
                                          readonly=True)
+
+
+class GenericComponent(model.Actuator):
+    """
+    Simulated component, capable of simulating both axes and VA's.
+    This component allows for communication with it, asif it were a real component with axes and VA's, but does not do
+    anything else.
+    """
+    def __init__(self, name, role, vas=None, axes=None, **kwargs):
```
Kleijwegt, line 249: I don't get when/why you want to use this generic component. I also don't get why it is not based on HwComponent instead of Actuator. In principle the component is now a generic sub component of Actuator. Suggesting it is more a "generic actuator"?

src/odemis/driver/simulated.py
```diff
@@ -339,3 +238,195 @@ def __init__(self, name, role, positions, has_pressure=False, **kwargs):
         # TODO: set to None/None when the sample is ejected
         self.sampleHolder = model.TupleVA((PHENOM_SH_FAKE_ID, PHENOM_SH_TYPE_OPTICAL),
                                          readonly=True)
+
+
+class GenericComponent(model.Actuator):
+    """
+    Simulated component, capable of simulating both axes and VA's.
+    This component allows for communication with it, asif it were a real component with axes and VA's, but does not do
```
pieleric, line 246: asif -> as if

src/odemis/driver/simulated.py
```diff
@@ -339,3 +238,195 @@ def __init__(self, name, role, positions, has_pressure=False, **kwargs):
...
+            "unit" (str): optional, unit of the axis, defaults to "m"
+            "range" (float, float): optional, min/max of the axis, defaults to (-0.1, 0.1)
+            "choices" (dict): optional, alternative to ranges, these are the choices of the axis
+            "speed" (float, float): optional, allowable range of speeds, defaults to (0., 10.)
+
+        """
+        # Create desired VA's
+        if vas:
+            for n in vas:
+                if "value" in vas[n]:
```
pieleric, line 274: Let's just make "value" required. For now, if it's not there, it's replaced by 0.0, but the VA is a basic one. It's just confusing. Forcing a default value would also help to clarify the microscope files.

src/odemis/driver/simulated.py
```diff
@@ -339,3 +238,195 @@ def __init__(self, name, role, positions, has_pressure=False, **kwargs):
...
+                    vatype = "VigilantAttribute"
+                elif vavaluetype == "VA":  # and varangetype != "VA"
+                    vatype = varangetype + vavaluetype
+                else:
+                    vatype = vavaluetype + varangetype
+
+                if "value" not in vas[n]:
+                    vas[n]["value"] = 0.0  # default value
+                if "unit" not in vas[n]:
+                    vas[n]["unit"] = ""  # default unit
```
pieleric, line 298: No, please don't. In VAs we distinguish between unit=None (ie, it'd be meaningless) and unit="" (ie, it's a unit-less value, such as a ratio). Let's keep the standard behaviour which is to have the default unit=None.

src/odemis/driver/simulated.py
```diff
@@ -339,3 +238,195 @@ def __init__(self, name, role, positions, has_pressure=False, **kwargs):
...
+        axis property dict can contain the following keys:
+            "unit" (str): optional, unit of the axis, defaults to "m"
+            "range" (float, float): optional, min/max of the axis, defaults to (-0.1, 0.1)
+            "choices" (dict): optional, alternative to ranges, these are the choices of the axis
+            "speed" (float, float): optional, allowable range of speeds, defaults to (0., 10.)
+
+        """
+        # Create desired VA's
+        if vas:
+            for n in vas:
```
pieleric, line 273: Add a comment here that you're doing some hacky stuff to guess the VA type based on the value type and range/choices.

src/odemis/driver/simulated.py
```diff
@@ -339,3 +238,195 @@ def __init__(self, name, role, positions, has_pressure=False, **kwargs):
...
+                    vatype = varangetype + vavaluetype
+                else:
+                    vatype = vavaluetype + varangetype
+
+                if "value" not in vas[n]:
+                    vas[n]["value"] = 0.0  # default value
+                if "unit" not in vas[n]:
+                    vas[n]["unit"] = ""  # default unit
+
+                vaclass = getattr(model, vatype)
```
pieleric, line 300: Please catch KeyError and raise a proper error explaining which VA got detected as which type, which do not exist.

src/odemis/driver/simulated.py
```diff
@@ -339,3 +238,195 @@ def __init__(self, name, role, positions, has_pressure=False, **kwargs):
...
+                va = vaclass(**vas[n])
+                setattr(self, n, va)
+
+        # Create desired axes
+        axes_def = {}
+        if axes:
+            self._position = {}
+            init_speed = {}
+            for a in axes:
```
pieleric, line 309: Please use axes.items() to iterate, and avoid writting `axes[a]` everywhere.

src/odemis/driver/simulated.py
pieleric, line 273: Please use vas.items() to iterate more nicely without writting `vas[n]` everywhere.

src/odemis/driver/simulated.py
pieleric: Please add a small comment above to explain this. 
BTW, an option would be to have a GenericComponent(HwComponent) and a subclass GenericActuator(GenericComponent, Actuator), fix the possible diamond inheritance issues in the model, and use a __new__() method in GenericComponent() to generate a GenericActuator if there are axes... but that sounds a little over the top for such small class.

pieleric (changes requested):

src/odemis/driver/simulated.py
```diff
@@ -339,3 +238,197 @@ def __init__(self, name, role, positions, has_pressure=False, **kwargs):
...
+                init_speed[axisname] = 1.0  # we are fast!
+                if "range" not in axisprop and "choices" not in axisprop:  # if no range nor choices are defined
+                    axisprop["range"] = (-0.1, 0.1)  # use the default range
+                if "speed" not in axisprop:
+                    axisprop["speed"] = (0., 10.)  # default speed
+                axes_def[axisname] = model.Axis(**axisprop)
+                if "range" in axisprop:
+                    self._position[axisname] = (axisprop["range"][0] + axisprop["range"][1]) / 2  # start at the centre
+                else:
+                    self._position[axisname] = list(axisprop["choices"].values())[0]  # start at the first value
```
pieleric, line 321: The axes support two (!) kinds of choices: set and dict. In case of dict, it's position -> user-friendly name. So at least you should use .keys().
However, a more fancy way, which would also work in case of a set would be to write `next(iter(axisprop["choices"]))`. As a comment, to be more precise indicate "pick one value arbitrarily" (as set and dict are not ordered).

pieleric (changes requested):
One thing I realise is missing is test cases. Please add at least one test case which generates a GenericComponent with a couple of VAs without axes, and another one with axes.

src/odemis/driver/test/simulated_test.py
```diff
@@ -289,5 +290,29 @@ def test_wrong_moveAbs(self):
...
+    def test_creation_vas_only(self):
+        simulated.GenericComponent(name="test_component", role="test", vas={"vaRange":
+                                                                                {"value": 0.1, "readonly": True,
+                                                                                 "unit": "", "range": [0, 1]},
+                                                                            "vaChoices":
+                                                                                {"value": 1,
+                                                                                 "choices": set(range(0, 10))},
+                                                                            "vaBool": {"value": True}})
+
+
```
pieleric, line 316: Please check that the VAs are indeed there, with the intended values and attributes. It should also be possible to set them (excepted for the readonly one).

src/odemis/driver/test/simulated_test.py
```diff
@@ -289,5 +290,29 @@ def test_wrong_moveAbs(self):
...
+    def test_creation_complete(self):
+        simulated.GenericComponent(name="test_component", role="test", vas={"vaRange":
+                                                                                {"value": 0.1, "readonly": True,
+                                                                                 "unit": "", "range": [0, 1]},
+                                                                            "vaChoices":
+                                                                                {"value": 1,
+                                                                                 "choices": set(range(0, 10))},
+                                                                            "vaBool": {"value": True}},
+                                   axes={"x": {"range": (-0.2, 0.2), "unit": "m"},
+                                         "gripper": {"choices": {'open': False, 'closed': True}}})
```
pieleric, line 305: Please check that the axes are indeed there, and that it's possible to move them within the given range/choices.

pieleric (changes requested):

src/odemis/driver/test/simulated_test.py
```diff
@@ -278,17 +278,64 @@ def test_stop(self):
...
+    def test_creation_complete(self):
+        comp = simulated.GenericComponent(name="test_component", role="test", vas={"vaRange":
+                                                                                       {"value": 0.1, "readonly": True,
+                                                                                        "unit": "", "range": [0, 1]},
+                                                                                   "vaChoices":
+                                                                                       {"value": 1,
+                                                                                        "choices": set(range(0, 10))},
+                                                                                   "vaBool": {"value": True}},
+                                          axes={"x": {"range": (-0.2, 0.2), "unit": "m"},
+                                                "gripper": {"choices": {'open': False, 'closed': True}}})
```
pieleric, line 305: The "choices" are reversed: it should be actual position (True/False) -> user friendly string.
This way, the `assertTrue()` will actually check something, as currently the position is "open" or "closed", which both are considered True by Python.

src/odemis/driver/test/simulated_test.py
```diff
@@ -278,17 +278,64 @@ def test_stop(self):
...
+
+        f = comp.moveAbs({'x': 0.1})
+        f.result()
+        self.assertEqual(comp.position.value['x'], 0.1)  # position should be exact for simulated component
+        f = comp.moveAbs({'gripper': 'closed'})
+        f.result()
+        self.assertTrue(comp.position.value['gripper'])  # assert the axes are accessible
+
+    def test_creation_vas_only(self):
+        comp = simulated.GenericComponent(name="test_component", role="test", vas={"vaRange":
```
pieleric, line 323: Could you align `vas` on the next line, with the `(`, so that you have more space for the VAs content.

pieleric (changes requested):

PR #1870, opened 2021-09-21
driver: fix piezomotor simulator absolute move

philipwinklerdelmic:
For the absolute move, the steps need to be calculated based on the current position. Also, unlike the hardware only one axis can move at a time, so use twice the speed in the simulator.

### DIFF
--- a/src/odemis/driver/piezomotor.py
+++ b/src/odemis/driver/piezomotor.py
@@ -1032,22 +1032,26 @@ class PMDSimulator(object):
                 else:
                     raise ValueError()
             elif cmd == "T":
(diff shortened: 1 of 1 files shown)


pieleric (approved):
As a note, instead of making the move faster, it should be possible to simulate this better by raising the number of workers in the executor. However, this would require a extension of do_move() to be per axis, which is quite some extra work. So I'm fine with this simplification.

philipwinklerdelmic:
I had the same thought. In the end, I didn't want to spend too much time on the simulator, so I kept it simple. Let's merge it.

PR #1871, opened 2021-09-22
streams: syntax error in exception

philipwinklerdelmic:
',' should be '%'


pieleric:
Thanks for fixing my bug!

PR #1872, opened 2021-09-22
streams: only allow whitelisted characters for fastem ROAs

philipwinklerdelmic:
Also, names only have to be unique within a project.

### DIFF
--- a/src/odemis/gui/comp/text.py
+++ b/src/odemis/gui/comp/text.py
@@ -618,6 +618,78 @@ def _step_from_range(min_val, max_val):
         logging.exception(msg)
 
 
+class PatternValidator(ValidatorClass):
+
+    def __init__(self, pattern):
+        """ pattern (str): regex pattern of allowed entries """
+        super().__init__()
+        self.Bind(wx.EVT_CHAR, self.on_char)
+        self.entry_pattern = pattern
+
+    def _is_valid_value(self, val):
+        """ Validate the given value
+
+        Args:
+            val (str):
+
+        Returns:
+            (boolean): True if the given string is valid
+
+        """
+
+        entry_pattern = re.compile(self.entry_pattern)
+        return bool(re.match(entry_pattern, val)) and entry_pattern.match(val).end() == len(val)
+
+    def _get_str_value(self):
+        """ Return the string value of the wx.Window to which this validator belongs """
+
+        # Special trick in case we are validating a NumberTextCtrl, which has it's
+        # default 'GetValue' method replaced with one that returns number instances
+
+        fld = self.GetWindow()
+
+        if hasattr(fld, "get_value_str"):
+            val = fld.get_value_str()
+        else:
+            val = fld.GetValue()
+        return val
+
+    def on_char(self, event):
+        """ This method prevents the entry of illegal characters """
+        ukey = event.GetUnicodeKey()
+        # Allow control keys to propagate (most of them are WXK_NONE in unicode)
+        if ukey < wx.WXK_SPACE:
+            event.Skip()
+            return
+
+        field_val = str(self._get_str_value())
+        start, end = self.GetWindow().GetSelection()
+        field_val = field_val[:start] + chr(ukey) + field_val[end:]
+
+        entry_pattern = re.compile(self.entry_pattern)
+        # Make sure the entire string matches the pattern (i.e. the match length equals the string length)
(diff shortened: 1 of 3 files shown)


src/odemis/gui/cont/streams.py
```diff
@@ -3335,12 +3337,19 @@ def _on_combobox(self, _):
...
         txt = self.panel.txt_ctrl.GetValue()
         current_name = self.model.name.value
-        all_roas = [roa.name.value for project in self._tab_data.projects.value for roa in project.roas.value]
+
+        # Process input, make sure name is unique in project and complies with whitelisted characters of
+        # technolution driver
+        roa_project = [p for p in self._tab_data.projects.value if self.model in p.roas.value][0]
+        all_project_roas = [roa.name.value for roa in roa_project.roas.value]
+        txt = txt.replace(" ", "-")  # replace spaces with dash
+        txt = re.sub(r'[^A-Za-z0-9_()-]+', '', txt)  # only allow characters from whitelist
```
pieleric, line 3346: Possibly a more elegant way would be to use the notion of "Validator" on the wx.TextCtrl. Essentially, this would prevent the user from typing text which is not acceptable. 
It would be a matter of creating a Validator, and then calling textctrl.SetValidator(...).  You can have a look at https://www.wxpython.org/Phoenix/docs/html/validator_overview.html , and the class odemis.gui.comp.text._NumberValidator .

pieleric (changes requested):
As a comment, the GUI is starting to look very ad-hoc for the specific hardware components of the FASTEM. That said, for this project, it's relatively unlikely we're suddenly going to switch supplier, so that's fine.

src/odemis/gui/cont/streams.py
philipwinklerdelmic: done

src/odemis/gui/comp/text.py
```diff
@@ -618,6 +618,78 @@ def _step_from_range(min_val, max_val):
...
+        # Allow control keys to propagate (most of them are WXK_NONE in unicode)
+        if ukey < wx.WXK_SPACE:
+            event.Skip()
+            return
+
+        field_val = str(self._get_str_value())
+        start, end = self.GetWindow().GetSelection()
+        field_val = field_val[:start] + chr(ukey) + field_val[end:]
+
+        entry_pattern = re.compile(self.entry_pattern)
```
pieleric, line 669: You shouldn't compile the regex everytime you use it. Compile it at init, then use the pattern.

src/odemis/gui/comp/text.py
```diff
@@ -618,6 +618,78 @@ def _step_from_range(min_val, max_val):
...
+        # Make sure the entire string matches the pattern (i.e. the match length equals the string length)
+        if entry_pattern.match(field_val) and entry_pattern.match(field_val).end() == len(field_val):
```
pieleric, line 671: It feel you should call `_is_valid_value()`, instead of duplicating the code.

src/odemis/gui/comp/text.py
```diff
@@ -618,6 +618,78 @@ def _step_from_range(min_val, max_val):
...
+        """ Return the string value of the wx.Window to which this validator belongs """
+
+        # Special trick in case we are validating a NumberTextCtrl, which has it's
+        # default 'GetValue' method replaced with one that returns number instances
+
+        fld = self.GetWindow()
+
+        if hasattr(fld, "get_value_str"):
+            val = fld.get_value_str()
+        else:
```
pieleric, line 653: We don't need this trick for the 2 text controls that we use it on, right? It'd be only if we wanted to use the PatternValidator on numbers... but we already have a dedicated validator for that. => Drop the whole function and replace by  `self.GetWindow().GetValue()`.

src/odemis/gui/comp/text.py
```diff
@@ -618,6 +618,78 @@ def _step_from_range(min_val, max_val):
...
+        Args:
+            val (str):
+
+        Returns:
+            (boolean): True if the given string is valid
+
+        """
+
+        entry_pattern = re.compile(self.entry_pattern)
+        return bool(re.match(entry_pattern, val)) and entry_pattern.match(val).end() == len(val)
```
pieleric, line 641: You can use `pattern.fullmatch()` to check that. (Since Python 3.4, which is fine for us as we have Python 3.5+)

pieleric (changes requested):

philipwinklerdelmic:
Pushed changes.

PR #1874, opened 2021-09-27
Added Light, Focus and a start for Scanner

Helsloot:


src/odemis/driver/orsay.py
```diff
@@ -2614,3 +2641,247 @@ def terminate(self):
...
+        :param (dict {"z": value}) pos: value contains the desired new focus position in meter.
+        :param (int or float) timeout: the maximum number of seconds to wait until a TimeoutError is raised.
+        """
+        new_voltage = self.baseLensVoltage + self._metadata[model.MD_CALIB] * pos["z"]
+        self.parent._fib_beam.objectiveVoltage.value = new_voltage
+
+        start = time.time()
+        while not self.position.value["z"] == pos['z']:
+            if time.time() - start >= timeout:
+                raise TimeoutError("Timed out changing the objective lens voltage")
```
pieleric, line 2853: There are quite a few problems with this loop. First, it's an active loop. It'll use the CPU at 100% while waiting. Second, if the .position is not computed exactly identical value it'll block until the timeout. Third, it'd be nice to indicate the timeout duration in the error message, the requested position, and current position.

=> Instead, subscribe to .position (before changing the objectiveVoltage), to a small listener function that sets an Event. Then, while the value is not `util.almost_equal()`, wait on the Event until the end time (computed at the beginning). You could also log every time the event is triggered but the position is concidered too far, to see if it's really blocking.

src/odemis/driver/test/orsay_test.py
```diff
@@ -46,6 +46,10 @@
 CONFIG_FIBVACUUM = {"name": "fib-vacuum", "role": "fib-vacuum"}
 CONFIG_FIBSOURCE = {"name": "fib-source", "role": "fib-source"}
 CONFIG_FIBBEAM = {"name": "fib-beam", "role": "fib-beam"}
+CONFIG_LIGHT = {"name": "light", "role": "light"}
+CONFIG_SCANNER = {"name": "scanner", "role": "scanner"}
+CONFIG_FOCUS = {"name": "focus", "role": "focus", "rng": (-2e-3, 2e-3),
+                "metadata": {model.MD_CALIB: 0.18e6}}  # MD_CALIB is in [V/m]
```
pieleric, line 52: No, change to a call to `updateMetadata`.

src/odemis/driver/orsay.py
```diff
@@ -2614,3 +2641,247 @@ def terminate(self):
...
+
+    def _doMoveAbs(self, pos, timeout=60):
+        """
+        Calculated the new voltage as V = baseLensVoltage + MD_CALIB * (current_position + delta).
+        Blocking until the new position is reached or it times out.
+        :param (dict {"z": value}) pos: value contains the desired new focus position in meter.
+        :param (int or float) timeout: the maximum number of seconds to wait until a TimeoutError is raised.
+        """
+        new_voltage = self.baseLensVoltage + self._metadata[model.MD_CALIB] * pos["z"]
+        self.parent._fib_beam.objectiveVoltage.value = new_voltage
```
pieleric, line 2848: Please add a log indicating the computed voltage based on the distance.

pieleric (changes requested):

src/odemis/driver/test/orsay_test.py
Helsloot: I don't fully understand how the metadata should be handled then. Should or shouldn't they be specified in the yaml file? If not, then where should the 0.18e6 be defined? And how should it end up in the metadata?

pieleric: The metadata will/should be in the microscope file. That's actually what you've done in the example microscope file.
When the backend reads the microscope file and instantiate it, it does two things (that matters for here). First, it instantiate the component, with the "init" as argument. Then, it calls `comp.updateMetadata()` with the metadata from the microscope file.
So in the test case, you should simulate the same behaviour: instantiate the component (without the "metadata" argument), and immediately after call `updateMetadata({model.MD_CALIB: 0.18e6})`.

pieleric: For the case that the user hasn't put some metadata in the yaml file, it's good that the component still behaves not completely weird. So you could start with an internal default value for model.MD_CALIB as `0.2e6` or `0.1e6` to provide a rough value.

install/linux/usr/share/odemis/hwtest/orsay.odm.yaml
```diff
@@ -87,9 +87,10 @@ Orsay: {
...
-        MD_CALIB: 0.18e6  # Volt per meter. Read the driver specifications for more details on this value
-    }
-}
\ No newline at end of file
+    init: {
+      rng: (-2e-3, 2e-3)  # range of the focus in meter
+    },
+#    metadata: {
+#        MD_CALIB: 0.18e6  # Volt per meter. Read the driver specifications for more details on this value
+#    }
```
pieleric, line 95: Put it back, that's fine.

src/odemis/driver/orsay.py
```diff
@@ -2799,24 +2760,32 @@ class Focus(model.Actuator):
...
+        # Event that get's set when the position VA changes value while performing a move using _doMoveAbs
+        self._position_changed = threading.Event()
+        self._position_changed.clear()
+
         self._executor = CancellableThreadPoolExecutor(max_workers=1)
 
+        self.updateMetadata({})
+
+    def updateMetadata(self, md):
+        md[model.MD_CALIB] = 0.18e6
```
pieleric, line 2786: This value should be in `md`. If not, too bad. You typically don't need to override `updateMetadata()`. If you want to be extra careful, you could override it to check that the new MD_CALIB is valid (eg, it's a float).

src/odemis/driver/orsay.py
```diff
@@ -2828,29 +2797,39 @@ def _updatePosition(self, value=None):
...
-    def _doMoveAbs(self, pos, timeout=60):
+    def _doMoveAbs(self, pos):
         """
         Calculated the new voltage as V = baseLensVoltage + MD_CALIB * (current_position + delta).
         Blocking until the new position is reached or it times out.
         :param (dict {"z": value}) pos: value contains the desired new focus position in meter.
-        :param (int or float) timeout: the maximum number of seconds to wait until a TimeoutError is raised.
         """
+        self.position.subscribe(self._position_changed_event_handler)
+
```
pieleric, line 2818: Also clear _position_changed here, just before changing the objectiveVoltage.

src/odemis/driver/test/orsay_test.py
```diff
@@ -1773,6 +1779,230 @@ def test_imageAreaStepping(self):
...
+            raise unittest.SkipTest(NO_SERVER_MSG)
+
+        cls.oserver = orsay.OrsayComponent(**CONFIG_ORSAY)
+        cls.datamodel = cls.oserver.datamodel
+        cls.ov = cls.datamodel.HVPSFloatingIon.ObjectiveVoltage  # should only be needed in simulation
+        for child in cls.oserver.children.value:
+            if child.name == CONFIG_FIBBEAM["name"]:
+                cls.fibbeam = child
+            elif child.name == CONFIG_FOCUS["name"]:
+                cls.focus = child
```
pieleric, line 1908: Call updateMetadata() here.

src/odemis/driver/orsay.py
```diff
@@ -2614,3 +2643,224 @@ def terminate(self):
...
+        Blocking until the new position is reached or it times out.
+        :param (dict {"z": value}) pos: value contains the desired new focus position in meter.
+        """
+        self.position.subscribe(self._position_changed_event_handler)
+
+        new_voltage = self.baseLensVoltage + self._metadata[model.MD_CALIB] * pos["z"]
+        self.parent._fib_beam.objectiveVoltage.value = new_voltage
+        logging.debug(f"Focus position set to {pos['z']}, setting objective voltage to {new_voltage}")
+
+        start = time.time()
```
pieleric, line 2823: To simplify a tiny bit, instead of storing `start` you can store `endt = time.time() + FOCUS_CHANGE_TIMEOUT `.

pieleric (changes requested):

PR #1875, opened 2021-09-29
gui: only clip stage position if stage is available

philipwinklerdelmic:
PR #1803 introduced a check for the stage limits in the recenter_buffer function in miccanvas. This only works if the view has a stage. To make sure it doesn't fail, this case needs to be handled in the clipping function of the view.

### DIFF
--- a/src/odemis/gui/model/__init__.py
+++ b/src/odemis/gui/model/__init__.py
@@ -1655,6 +1655,9 @@ class StreamView(View):
         :param pos (dict): Position to be clipped with keys "x" and "y"
         :return(dict): Position clipped to the stage limits with keys "x" and "y"
         """
+        if not self._stage:
+            return pos
+
         stage_limits = self._getStageLimitsXY()
         if not stage_limits["x"][0] <= pos["x"] <= stage_limits["x"][1]:
             pos["x"] = max(stage_limits["x"][0], min(pos["x"], stage_limits["x"][1]))


src/odemis/gui/model/__init__.py
```diff
@@ -1655,6 +1655,9 @@ def clipToStageLimits(self, pos):
         :param pos (dict): Position to be clipped with keys "x" and "y"
         :return(dict): Position clipped to the stage limits with keys "x" and "y"
         """
+        if not self._stage:
+            return pos
```
Kleijwegt, line 1659: I would at least add a comment here why we return the dict directly.
However, I think there are two better implementations:
If there is no stage defined the function _getStageLimitsXY() should return an empty dict.
Adjusting the two if statements to not do anything in that case or return when the dict is empty seems better to me.

Or we change the code where the call is made:  (Since if no stage is defined we shouldn't even want to limit its movement to that.

We could write in _recenter_buffer_:

> move_dict = {"x": phys_pos[0], "y": phys_pos[1]}
> if self.view._stage:
>     move_dict = self.view.clipToStageLimits(move_dict)

philipwinklerdelmic: I don't agree, here's why:

* ```recenter_buffer```a: In my first implementation, I did just that. But it's not a good idea to access the private ```._stage``` attribute of the view. Also, the function ```clipToStageLimits``` might be called from somewhere else where we get the same problem

* adding if-statement for empty dict returned by ```_getStageLimitsXY``` doesn't work because this function tries to access some attributes (like the axes) from ```._stage```. If ```._stage``` is None, this fails (that's how I noticed the error in the first place).


Having 
```
if not self._stage:
    return
```
is quite common in ```model.__init__```, see for example ```moveStageToView```, ```moveStageTo```, ```moveStageBy```,..., so I think it's the easiest and cleanest solution. Does that make sense?

pieleric: Note that there is a `.has_stage()` function to let the other classes find out if there is a stage.
In general, I would tend to prefer also having the check in the caller, like `if view.has_stage(): view.clipToStageLimits(...)`. However, the rest of the code already has taken the alternative way of always calling the stage functions, and making them do nothing in case there is no stage. Let's follow this here too.

PR #1879, opened 2021-09-30
model futures: improve updating end time when starting a ProgressiveFuture

pieleric:
Force the order of computation to reduce floating point error.

More importantly: fix test case which didn't pass anymore now that the
end time is updated immediately.

### DIFF
--- a/src/odemis/model/_futures.py
+++ b/src/odemis/model/_futures.py
@@ -489,7 +489,7 @@ class ProgressiveFuture(CancellableFuture):
             # Set start to current time and end to estimated endtime (difference between current end and start)
             start, end = self.get_progress()
             startt = time.time()
-            endt = startt + end - start
+            endt = startt + (end - start)  # Force the computation order to reduce floating point error
             self.set_progress(start=startt, end=endt)
 
         return running
--- a/src/odemis/model/test/futures_test.py
+++ b/src/odemis/model/test/futures_test.py
@@ -313,7 +313,8 @@ class TestFutures(unittest.TestCase):
         f.set_running_or_notify_cancel()
         startf, endf = f.get_progress()
         self.assertLessEqual(startf, time.time())
-        self.assertEqual(end, endf)
+        self.assertAlmostEqual(startf + 1, endf)
+        self.assertAlmostEqual(endf, time.time() + 1, delta=0.1)
         time.sleep(0.1)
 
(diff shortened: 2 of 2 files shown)


src/odemis/model/_futures.py
```diff
@@ -489,7 +489,8 @@ def set_running_or_notify_cancel(self):
             # Set start to current time and end to estimated endtime (difference between current end and start)
             start, end = self.get_progress()
             startt = time.time()
-            endt = startt + end - start
+            endt = startt + (end - start)
```
tepals, line 492: Can you add this as a comment here, so that nobody ends up removing the brackets at some point?

`Force the order of computation to reduce floating point error.`

src/odemis/model/_futures.py
```diff
@@ -489,7 +489,8 @@ def set_running_or_notify_cancel(self):
...
+            print(endt - startt)
```
tepals, line 493: should this print stay here?

pieleric: Nope :see_no_evil:

tepals (approved):

src/odemis/model/_futures.py
```diff
@@ -489,7 +489,7 @@ def set_running_or_notify_cancel(self):
             # Set start to current time and end to estimated endtime (difference between current end and start)
             start, end = self.get_progress()
             startt = time.time()
-            endt = startt + end - start
+            endt = startt + (end - start)  # Force the computation order to reduce floating point error
```
Kleijwegt, line 492: The brackets only add some clarity for the user (it doesn't change any mathematics) though that is fine.

Kleijwegt (approved):

philipwinklerdelmic (approved):

src/odemis/model/_futures.py
pieleric: No. As the comment indicates, it does change the value. From a simple mathematics perspective, it shouldn't. However, this is float computation, not perfect mathematics numbers. So there are floating point errors that get introduced.
You can recreate the issue with something like this:
```python
startt = 1633339732.6324744
endt = startt + 1
newt = startt + 125.78
print(newt + endt - startt)
1633339859.4124742
print(newt + (endt - startt))
1633339859.4124744
```

PR #1884, opened 2021-10-06
fastem: hardware configuration

philipwinklerdelmic:
Megafield, overview, and livestream acquisitions need different settings of the scanner. The new function sets these values based on keyword arguments which can be passed through a dictionary.

### DIFF
--- a/src/odemis/acq/fastem.py
+++ b/src/odemis/acq/fastem.py
@@ -35,10 +35,8 @@ from odemis.acq import stitching
 from odemis.acq.stitching import REGISTER_IDENTITY
 from odemis.acq.stream import SEMStream
 from odemis.util import TimeoutError
+from odemis.acq import fastem_conf
 
-
-SINGLE_BEAM_ROTATION_DEFAULT = 0  # 0 is typically a good guess (better than multibeam rotation of ~0.0157 rad)
-MULTI_BEAM_ROTATION_DEFAULT = 0
 
 # The executor is a single object, independent of how many times the module (fastem.py) is loaded.
 _executor = model.CancellableThreadPoolExecutor(max_workers=1)
@@ -167,7 +165,7 @@ class FastEMROC(object):
         # if None -> acquire, if not None, don't acquire again
 
 
-def acquire(roa, path, multibeam, descanner, detector, stage):
+def acquire(roa, path, scanner, multibeam, descanner, detector, stage):
     """
     Start a megafield acquisition task for a given region of acquisition (ROA).
 
@@ -177,6 +175,7 @@ def acquire(roa, path, multibeam, descanner, detector, stage):
                 path as specified in the component.
                 The ASM will create the directory on the external storage, including the parent directories,
                 if they do not exist.
+    :param scanner: (xt_client.Scanner) Scanner component connecting to the XT adapter.
     :param multibeam: (technolution.EBeamScanner) The multibeam scanner component of the acquisition server module.
     :param descanner: (technolution.MirrorDescanner) The mirror descanner component of the acquisition server module.
     :param detector: (technolution.MPPC) The detector object to be used for collecting the image data.
@@ -192,7 +191,7 @@ def acquire(roa, path, multibeam, descanner, detector, stage):
 
     # TODO: pass path through attribute on ROA instead of argument?
     # Create a task that acquires the megafield image.
-    task = AcquisitionTask(multibeam, descanner, detector, stage, roa, path, f)
+    task = AcquisitionTask(scanner, multibeam, descanner, detector, stage, roa, path, f)
 
     f.task_canceller = task.cancel  # lets the future know how to cancel the task.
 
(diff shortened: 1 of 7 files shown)


src/odemis/acq/fastem.py
```diff
@@ -167,6 +167,75 @@ def __init__(self, name, coordinates):
...
+
+    # => compute the scale needed in X, use the same one in Y, and then compute the Y resolution.
+    # => set resolution to full FoV, and then adjust
+    scale = scanner.shape[0] / resolution[0]
+    scanner.scale.value = (scale, scale)
+    if scanner.resolution.value != TILE_RES:
+        logging.warning("Unexpected resolution %s on e-beam scanner, expected %s",
+                        scanner.resolution.value, TILE_RES)
+
+    md = scanner.getMetadata()
```
Sabrina85, line 228: How does this work now. You deleted below the method that reads the rotation for multibeam acquisition on the scanner from the MD. Here, you now always set the single beam rotation right? Do you not need to pick the right MD depending on the mode (live, megafield acq or overview acq)?

src/odemis/acq/fastem.py
```diff
@@ -552,7 +611,7 @@ def _run_overview_acquisition(f, stream, stage, area, live_stream):
     """
     :returns: (DataArray)
     """
-    _configure_overview_hw(stream.emitter)
```
Sabrina85, line 555: You also need to drop the method itself in right?

src/odemis/acq/fastem.py
```diff
@@ -167,6 +167,75 @@ def __init__(self, name, coordinates):
...
+                        scanner.resolution.value, TILE_RES)
+
+    md = scanner.getMetadata()
+    if model.MD_SINGLE_BEAM_ROTATION in md:
+        scanner.rotation.value = md[model.MD_SINGLE_BEAM_ROTATION]
+    else:
+        scanner.rotation.value = SINGLE_BEAM_ROTATION_DEFAULT
+        logging.warning("Scanner doesn't have SINGLE_BEAM_ROTATION metadata, using %s rad.",
+                        scanner.rotation.value)
+    md[model.MD_ROTATION_COR] = -scanner.rotation.value
```
Sabrina85, line 235: This was only needed for overview image acquisition. If I remember correctly in order to display the overview image at the same angle as the megafield image (where we do not use MD_ROTATION_COR) in Odemis. So you would need to distinguish here between live/overview and megafield as well.
@pieleric Can you comment or confirm, so we can add a comment here for clarity?
Is this the MD that goes on the image?

src/odemis/acq/fastem.py
```diff
@@ -167,6 +167,75 @@ def __init__(self, name, coordinates):
...
+}
+
+
+SCANNER_CONFIG_LIVESTREAM = {
+    "multibeam_mode": False,
+    "external": False,
+    "blanker": None,  # automatic
+    "immersion": True,
+    "horizontal_fov": 2.2e-5,
+    "resolution": (1024, 884),
```
Sabrina85, line 196: Add a comment saying that resolution for megafield imaging is controlled by the acquisition server module (ASM).

Sabrina85 (changes requested):

src/odemis/acq/fastem.py
philipwinklerdelmic: I don't know. Is this something that is different for each mode? If so, how?

src/odemis/acq/fastem.py
```diff
@@ -167,6 +167,75 @@ def __init__(self, name, coordinates):
         # if None -> acquire, if not None, don't acquire again
 
 
+SCANNER_CONFIG_OVERVIEW = {
+    "multibeam_mode": False,
+    "external": False,
```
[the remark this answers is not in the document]
Sabrina85: No, overview and live stream are both using scan mode, so external = False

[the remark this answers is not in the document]
Sabrina85: `# Full frame` is correct - sorry.

philipwinklerdelmic:
Ok, it's getting a bit complicated
@tepals Can you have a look at the xt commit? I think there might be a bug with the autoblanking functionality, see commit message. I propose to completely change it and use an additional VA. But maybe I'm wrong.

Other than that, I moved the configure_scanner function to a different location to avoid ciruclar dependencies in the live stream. The livestream needs this function in .prepare to set up the correct values.

If everyone agrees with the changes, I need to clean up a bit.

@Sabrina85 why don't we use autoblanking on the megafield acquisition?

tepals:
@philipwinklerdelmic I think using an autoblank VA is fine. I based the current implementation on the Phenom driver, but that does not have an update settings function, so the blanker VA is not overwritten.

pieleric (changes requested):
Just some minor comments about the test case.

Once you've fixed them, it'd be nice if you could rewrite the history to have (at least) two commits: one for the xt_client changes, and one for the actual fastem_conf.

Sabrina85:
@pieleric Can you have a last look on the last commit please? Afterwards, I will squash everything.

pieleric (approved):
Looks fine to me. Could you squash into 2 commits?

Sabrina85:
> Could you squash into 2 commits?

Done

PR #1885, opened 2021-10-06
simsem: allow modification of .rotation VA

philipwinklerdelmic:
In the FastEM, we want change the .rotation VA. The simulator failed because the VA was readonly. We can allow it to be r/w. Note that the rotation is stored in the metadata in ._simulate_image.

### DIFF
--- a/src/odemis/driver/simsem.py
+++ b/src/odemis/driver/simsem.py
@@ -198,7 +198,7 @@ class Scanner(model.Emitter):
         # (float) in rad => rotation of the image compared to the original axes
         # TODO: for now it's readonly because no rotation is supported
         self.rotation = model.FloatContinuous(0, [0, 2 * math.pi], unit="rad",
-                                              readonly=True)
+                                              readonly=False)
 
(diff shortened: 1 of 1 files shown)


src/odemis/driver/simsem.py
```diff
@@ -198,7 +198,7 @@ def __init__(self, name, role, parent, aperture=100e-6, wd=10e-3, **kwargs):
         # (float) in rad => rotation of the image compared to the original axes
         # TODO: for now it's readonly because no rotation is supported
```
Sabrina85, line 199: Is this TODO then not also obsolete?

src/odemis/driver/simsem.py
```diff
@@ -198,7 +198,7 @@ def __init__(self, name, role, parent, aperture=100e-6, wd=10e-3, **kwargs):
...
         self.rotation = model.FloatContinuous(0, [0, 2 * math.pi], unit="rad",
-                                              readonly=True)
+                                              readonly=False)
```
Sabrina85, line 201: Can you not just remove the readonly as arg?

Sabrina85 (changes requested):

PR #1886, opened 2021-10-06
driver xt_client: handle pressure value of -1

philipwinklerdelmic:
HW returns -1 in case something went wrong. Catch this and log a warning, so it doesn't fail completely.

### DIFF
--- a/src/odemis/driver/xt_client.py
+++ b/src/odemis/driver/xt_client.py
@@ -438,10 +438,14 @@ class SEM(model.HwComponent):
             self.server.vent()
 
     def get_pressure(self):
-        """Returns: (float) the chamber pressure in pascal."""
+        """Returns: (float) the chamber pressure in pascal, raises IOError in case pressure could not be read."""
         with self._proxy_access:
             self.server._pyroClaimOwnership()
-            return self.server.get_pressure()
+            pressure = self.server.get_pressure()
+            if pressure == -1:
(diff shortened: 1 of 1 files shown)


src/odemis/driver/xt_client.py
```diff
@@ -1756,10 +1760,12 @@ def _refreshPressure(self):
...
         # Pressure
-        pressure = self.parent.get_pressure()
-        self.pressure._set_value(pressure, force_write=True)
-
-        logging.debug("Updated chamber pressure, %s Pa, vacuum state %s.", pressure, val["vacuum"])
+        try:
+            pressure = self.parent.get_pressure()
+            self.pressure._set_value(pressure, force_write=True)
+            logging.debug("Updated chamber pressure, %s Pa, vacuum state %s.", pressure, val["vacuum"])
+        except IOError:
```
tepals, line 1767: I think it makes sense here to say:
```
except IOError:
    if pressure == -1:
        logging.warning("Couldn't read pressure value.")
    else:
        raise
```

philipwinklerdelmic: The way I implement it now is that in case of -1 pressure, an IOError is raised (and .get_pressure will not return anything) so it doesn't work. We could make a specific exception just for that case, but I think it's a bit over the top;. I don't think it's likely an IOError will be raised becaseu of other issues.

philipwinklerdelmic:
done

PR #1893, opened 2021-10-14
Changing the order of the stage movements for Enzel to be safer

Kleijwegt:


src/odemis/acq/move.py
```diff
@@ -365,17 +365,19 @@ def _doCryoSwitchSamplePosition(future, stage, align, target):
...
             # Add the sub moves to perform the loading move
-            sub_moves.append((stage, filter_dict({'rx', 'rz'}, target_pos[LOADING])))
             if current_label is UNKNOWN and not stage_referenced:
                 # After referencing the stage could move near the maximum axes range,
                 # and moving single axes may result in an invalid/reachable position error,
                 # so all axes will moved together for this special case.
                 sub_moves.append((stage, filter_dict({'x', 'y', 'z'}, target_pos[LOADING])))
             else:
-                sub_moves.append((stage, filter_dict({'x', 'y'}, target_pos[LOADING])))
+                sub_moves.append((stage, filter_dict({'x'}, target_pos[LOADING])))
```
pieleric, line 374: I think this code should have a large comment explaining *why* it's done in this way. What does it solve compared to moving all axes at once, and why this order is better than other order.

If I understand correctly, moving one axis at a time helps because it's easier for the user to notice something is going wrong and pressing the stop (aka "cancel") button. Moving X first is a way to move it to a safe position, as it's not affected by the rx (and rz is typically always 0). Moreover, it's the largest move, and so it'll be "around" the loading position. Every other move is typically much smaller. In case the pivot point is wrong, then z and y might move incorrectly, but it's less likely to cause major damage.

src/odemis/acq/move.py
```diff
@@ -390,11 +392,21 @@ def _doCryoSwitchSamplePosition(future, stage, align, target):
...
+                sub_moves.append((stage, filter_dict({'y'}, target_pos[target])))
                 sub_moves.append((stage, filter_dict({'z'}, target_pos[target])))
-                sub_moves.append((stage, filter_dict({'x', 'y'}, target_pos[target])))
-            sub_moves.append((stage, filter_dict({'rx', 'rz'}, target_pos[target])))
+                sub_moves.append((stage, filter_dict({'rx', 'rz'}, target_pos[target])))
+
+        elif target in (COATING, SEM_IMAGING):
+            # TODO This should be implemented such that the Rx movement isn't done under the pole piece. Meaning
+            #  the 5DOF is first moved away and then moved back.
+            raise NotImplementedError
```
pieleric, line 408: We need to have an implementation for this case too.

src/odemis/acq/move.py
```diff
@@ -390,11 +392,21 @@ def _doCryoSwitchSamplePosition(future, stage, align, target):
             if current_label == LOADING:
                 # As moving from loading position requires re-referencing the stage, move all axes together to
                 # prevent invalid/reachable position error
-                sub_moves.append((stage, filter_dict({'x', 'y', 'z'}, target_pos[target])))
+                sub_moves.append((stage, filter_dict({'y'}, target_pos[target])))
+                sub_moves.append((stage, filter_dict({'z'}, target_pos[target])))
+                sub_moves.append((stage, filter_dict({'rx', 'rz'}, target_pos[target])))
+                sub_moves.append((stage, filter_dict({'x'}, target_pos[target])))
```
pieleric, line 398: Here we need a comment too. I guess the idea is essentially the same/reverse as for going to LOADING: do the small movements first near the loading position, and end with the large x move to get close to the pole-piece.

src/odemis/acq/move.py
```diff
@@ -390,11 +392,21 @@ def _doCryoSwitchSamplePosition(future, stage, align, target):
...
             else:
+                sub_moves.append((stage, filter_dict({'x'}, target_pos[target])))
+                sub_moves.append((stage, filter_dict({'y'}, target_pos[target])))
                 sub_moves.append((stage, filter_dict({'z'}, target_pos[target])))
-                sub_moves.append((stage, filter_dict({'x', 'y'}, target_pos[target])))
-            sub_moves.append((stage, filter_dict({'rx', 'rz'}, target_pos[target])))
+                sub_moves.append((stage, filter_dict({'rx', 'rz'}, target_pos[target])))
```
pieleric, line 403: Here, it's not clear why it was changed. Is the idea just that each movement should be separate? Then, as each move is small (and sort of undefined), we can just do them in alphabetical order? The rotation would be last, because it's assumed that previous it's 0, so it's safer to rotate last (as the pivot point is under the pole piece). 

Or should we check that and do the rotation earlier if the current rx > target rx? Shouldn't be X moved last, as it's most likely to get closer to the pole-piece, so we should do it only when the rest is in place.

pieleric (changes requested):
I understand that this code was tested on the real hardware (at least once). I think that the main need is to have comments explaining why it's done in this way.

pieleric:
This is finally implemented by PR #1948, so dropping this PR.

PR #1901, opened 2021-10-21
Meteor mpi visit tested

Mahmood-B:

PR #1880, opened 2021-10-01
Make adjustments to the localization tab 

Mahmood-B:
The adjustments are two parts. One is making the tiling more user friendly, and the second part is updating the focus position metadata. 
The tiling adjustments include adding widgets to the overview acquisition window to control the number of tiles along x and y directions. Also, adding a text to the overview window to display the requested tiling area size. Testing for some of the added code was made. 

Note: this branch was made out of the chamber tab branch. So only from the commit (64d3fca) onward, the commits are relevant.

PR #1904, opened 2021-10-27
Adjust overview window

Mahmood-B:
I made this separate branch out of master for the commits related to the overview acquisition. I cherry-picked the relevant commits from the PR #1880 and the PR  #1901. Basically, this PR could be merged separately. But  i think it would be a good idea to run the same test cases on meteor system once the PR of meteor chamber tab is merged.

### DIFF
--- a/src/odemis/acq/stitching/_tiledacq.py
+++ b/src/odemis/acq/stitching/_tiledacq.py
@@ -189,9 +189,6 @@ class TiledAcquisitionTask(object):
         # The size of the smallest tile, non-including the overlap, which will be
         # lost (and also indirectly represents the precision of the stage)
         reliable_fov = ((1 - self._overlap) * self._sfov[0], (1 - self._overlap) * self._sfov[1])
-        logging.debug("Would need tiles: nx= %s, ny= %s",
-                      abs(self._area_size[0] / reliable_fov[0]),
-                      abs(self._area_size[1] / reliable_fov[1]))
         nx = math.ceil(abs(self._area_size[0] / reliable_fov[0]))
         ny = math.ceil(abs(self._area_size[1] / reliable_fov[1]))
         logging.debug("Calculated number of tiles nx= %s, ny= %s" % (nx, ny))
--- a/src/odemis/acq/stitching/_tiledacq.py
+++ b/src/odemis/acq/stitching/_tiledacq.py
@@ -189,7 +189,9 @@ class TiledAcquisitionTask(object):
         # The size of the smallest tile, non-including the overlap, which will be
         # lost (and also indirectly represents the precision of the stage)
         reliable_fov = ((1 - self._overlap) * self._sfov[0], (1 - self._overlap) * self._sfov[1])
-
+        logging.debug("Would need tiles: nx= %s, ny= %s",
+                      abs(self._area_size[0] / reliable_fov[0]),
+                      abs(self._area_size[1] / reliable_fov[1]))
         nx = math.ceil(abs(self._area_size[0] / reliable_fov[0]))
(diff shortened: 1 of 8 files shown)


src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -189,9 +189,11 @@ def _getNumberOfTiles(self):
...
         # lost (and also indirectly represents the precision of the stage)
         reliable_fov = ((1 - self._overlap) * self._sfov[0], (1 - self._overlap) * self._sfov[1])
-
-        nx = math.ceil(abs(self._area_size[0] / reliable_fov[0]))
-        ny = math.ceil(abs(self._area_size[1] / reliable_fov[1]))
+        logging.debug("Would need tiles: nx= %s, ny= %s",
+                      abs(self._area_size[0] / reliable_fov[0]),
+                      abs(self._area_size[1] / reliable_fov[1]))
+        nx = math.floor(abs(self._area_size[0] / reliable_fov[0]))
+        ny = math.floor(abs(self._area_size[1] / reliable_fov[1]))
```
pieleric, line 196: No. This doesn't make sense. If we need more than C tiles, we need C+1 tiles. We cannot round down. BTW, this would break several test cases.

The only thing that we could do is to slightly reduce the area_size (eg, `* 0.999` or `- reliable_fov * 0.01`), to make sure that if precisely N tiles are requested, it's not round up due to floating point error.

src/odemis/gui/win/acquisition.py
```diff
@@ -1008,6 +1061,21 @@ def on_acquisition_done(self, future):
...
         self.EndModal(wx.ID_OPEN)
 
+    def on_tiles_number(self, _=None):
+        """
+        Called when the user enters values for the tiles number in the GUI.
+        """
+        nx = self.tiles_nx.value 
+        ny = self.tiles_ny.value 
+        w = nx*self.fov[0]-(nx-1)*self.fov[0]*self.overlap
+        h = ny*self.fov[1]-(ny-1)*self.fov[1]*self.overlap
```
pieleric, line 1071: Did you check `nx * self.fov[0] * (1 - self.overlap)`? That should then be the same formula as tiledacq.

pieleric (changes requested):

src/odemis/gui/win/acquisition.py
Mahmood-B: This formula alone is not enough. As you said, the number of tiles is some number followed by `.00004` for example, then its taking the ceil value increases the number of tiles. So the suggestion of decreasing the size area seems effective

src/odemis/gui/win/test/acquisition_test.py
```diff
@@ -0,0 +1,145 @@
...
+        self.assertAlmostEqual(rect_pts[0], self.tiling_rng["x"][0])
+        self.assertAlmostEqual(rect_pts[1], self.tiling_rng["y"][0])
+        self.assertAlmostEqual(rect_pts[2], self.tiling_rng["x"][0] + 200e-6 + w/2)
+        self.assertAlmostEqual(rect_pts[3], self.tiling_rng["y"][0] + 200e-6 + h/2)
+
+    def test_clipping_is_not_performed_same_width_and_height(self):
+        # move the stage to some position away from the edges of the active
+        # range so that clipping is not needed. 
+        stage_grid1_fm = transformFromSEMToMeteor(self.stage_grid1_sem, self.stage_bare)
+        self.stage_bare.moveAbs(stage_grid1_fm).result()
```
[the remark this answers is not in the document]
Mahmood-B: I changed the logic a bit. I removed the reliance on the this transform function.

PR #1908, opened 2021-11-02
gui comp text: Fix number ctrl accepting - even if negative values are forbidden

pieleric:
We have code that check that "-" is not allowed, but due to a mistake in
the logic, it would basically always consider it allowed.

### DIFF
--- a/src/odemis/gui/comp/text.py
+++ b/src/odemis/gui/comp/text.py
@@ -438,7 +438,7 @@ class _NumberValidator(wx.Validator):
         }
 
         if (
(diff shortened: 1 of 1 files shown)


pieleric:
Actually the original code was fine. We need to treat min_val == None as min_val = -infinity.
There is a small problem with the regex, which allows to write non-negative floats as -123 because it could be the beginning of writing the exponent. I don't think it makes sense to spend time "fixing" this. In any case, on final value validation, such value will not be accepted.

PR #1909, opened 2021-11-02
acq stitching: don't acquire extra tiles when area is a multiple of the FoV

pieleric:
Due to floating point errors, sometimes when requesting the acquisition
of an area exactly the multiple of the FoV (eg, 5x5 FoV), it would
acquire one extra tile (eg, 6x6 FoV). That would typically be due to
floating point errors. However, to make it more generic, we just round
down anything less than 1% of a tile.

Useful for the new overview acquisition window which allows the user to
specify the acquisition area in terms of FoV.


Mahmood-B (approved):

lazem (approved):
Looks good.

PR #1910, opened 2021-11-03
Feature Bookmarking Fixes 

lazem:
Mostly fixes from Eric's review & MPI visit.


src/odemis/gui/win/acquisition.py
```diff
@@ -643,7 +643,7 @@ def __init__(self, parent, orig_tab_data):
 
         # The pattern to use for storing each tile file individually
         # None disables storing them
-        self.filename_tiles = create_filename(self.conf.last_path, "{datelng}-{timelng}-overview",
+        self.filename_tiles = create_filename(self.conf.pj_last_path, "{datelng}-{timelng}-overview",
```
pieleric, line 646: It's not that simple because this window can be called from the secom or delphi too, which don't have a `pj_last_path`. In such case, `last_path` is good. That's why OOP should help with, but I'm not sure how to handle it here nicely. An alternative would be to pass it as an argument to the OverviewAcquisitionDialog. Another way would be to check if it's enzel/meteor or other (but it sounds pretty easy to break).

pieleric (changes requested):

src/odemis/gui/cont/acquisition.py
```diff
@@ -558,6 +558,11 @@ def _on_acquire(self, _):
             self._acq_future = acqmng.acquire(
                 self._acquiStreams.value, self._tab_data.main.settings_obs)
 
+        # Pause live streams before acquiring
+        for s in self._acquiStreams.value:
+            s.is_active.value = False
+            s.should_update.value = False
```
pieleric, line 564: The streambar controller does already provide such function. I think it should be possible to call it like `self._tab.streambar_controller.pauseStreams()`. It works on the `.streams`, but if I understand correctly, `.acquisitionStreams` is a subset of .streams, right? If so, it should be fine.
Could you check if that works?

pieleric (changes requested):

src/odemis/gui/comp/viewport.py
```diff
@@ -336,6 +336,17 @@ def setView(self, view, tab_data):
...
             view.fov_hw.horizontalFoV.subscribe(self._on_hw_fov_change, init=True)
 
+        if hasattr(view, "showFeatures"):
+                view.showFeatures.subscribe(self._show_hide_feature_overlay)
+                self.bottom_legend.feature_toggle_va = view.showFeatures
+
+    def _show_hide_feature_overlay(self, va_val):
+        # show/hide feature overlay based on the legend toggle button
+        foverlay = next((ol for ol in self.canvas.world_overlays if isinstance(ol, CryoFeatureOverlay)), None)
+        if foverlay:
```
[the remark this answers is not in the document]
lazem: It's created from 2 different places: localization tab (via `canvas.cryofeature_overlay`) and the chamber tab from the `FeatureOverviewViewport. `

PR #1912, opened 2021-11-05
improve backend starter for testing

Mahmood-B:


src/odemis/util/test.py
```diff
@@ -43,15 +44,36 @@ def setlimits():
...
+    Start the backend by checking the currently running backend. Basically 2 cases/scenarios:
+        1. If no running backend => run the requested one.
+        2. If there is running backend =>
+            a. If the running backend is same as requested one => do nothing
+            b. If the running backend is different from the requested one => stop the running, and run the requested one.
+    """
+    # a backend running
+    if driver.get_backend_status() in (driver.BACKEND_RUNNING, driver.BACKEND_STARTING):
+        current_model = model.getMicroscope().model
+        req_model = modelgen.Instantiator(open(config)).ast
```
pieleric, line 57: This can raise all sort of errors. For instance, the file is missing, or the file cannot be parsed. You should decide what to do in such case. I'd tend to suggest to throw a ValueError with the original exception as message.

src/odemis/util/test.py
```diff
@@ -43,15 +44,36 @@ def setlimits():
...
+        if current_model == req_model:
+            return
```
pieleric, line 59: Please add a log.info("Backend for %s already running", config).

src/odemis/util/test.py
```diff
@@ -43,15 +44,36 @@ def setlimits():
...
+        else:
+            stop_backend()
```
pieleric, line 61: Please add a log to indicate that the backend is already running, but with a different model, so it will be stopped and restarted.

pieleric (changes requested):
As a separate commit, you could simplify src/odemis/gui/win/test/acquisition_test.py which used to do what start_backend now does.

src/odemis/util/test.py
```diff
@@ -43,15 +44,36 @@ def setlimits():
 
 def start_backend(config):
     """
-    Start the backend
+    config (str): path to the microscope config file.
+    Start the backend by checking the currently running backend. Basically 2 cases/scenarios:
```
tepals, line 48: Normally in the docstring the description is above the parameters, so I would update this docstring to:
```
"""
Start the backend by checking the currently running backend. Basically 2 cases/scenarios:
        1. If no running backend => run the requested one.
        2. If there is running backend =>
            a. If the running backend is same as requested one => do nothing
            b. If the running backend is different from the requested one => stop the running, and run the requested one.

config (str): path to the microscope config file.
"""
```

src/odemis/util/test.py
```diff
@@ -43,15 +44,36 @@ def setlimits():
...
+        1. If no running backend => run the requested one.
+        2. If there is running backend =>
+            a. If the running backend is same as requested one => do nothing
+            b. If the running backend is different from the requested one => stop the running, and run the requested one.
+    """
+    # a backend running
```
tepals, line 54: I don't understand this comment, do you mean: check if a backend is running?

tepals (approved):

src/odemis/util/test.py
```diff
@@ -49,16 +49,22 @@ def start_backend(config):
...
+        try:
+            req_model = modelgen.Instantiator(open(config)).ast
+        except Exception as exp:
+            raise ValueError(exp)
         if current_model == req_model:
+            logging.info("Backend for %s already running", config)
             return
         else:
+            logging.info("There is a backend running already, it will be turned off, and the backend \
+                                %s will be run instead.", config)
```
pieleric, line 67: In Python, you can write a string over muliple-lines just by using concatenation:
```python
            logging.info("There is a backend running already, it will be turned off, and the backend"
                                "%s will be run instead.", config)
```

PR #1915, opened 2021-11-08
Added (de-)scan gain and offset metadata for FastEM systems to odemis…

tepals:
… model

### DIFF
--- a/src/odemis/model/_metadata.py
+++ b/src/odemis/model/_metadata.py
@@ -219,3 +219,9 @@ TINT_RGB_AS_IS = "rgbasis"
 # Rotation for FastEM multi-beam and single-beam scanner
 MD_SINGLE_BEAM_ROTATION = "Single-beam rotation"
 MD_MULTI_BEAM_ROTATION = "Multi-beam rotation"
+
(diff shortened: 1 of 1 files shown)


Sabrina85 (approved):

src/odemis/model/_metadata.py
```diff
@@ -211,3 +211,8 @@
 # Rotation for FastEM multi-beam and single-beam scanner
 MD_SINGLE_BEAM_ROTATION = "Single-beam rotation"
 MD_MULTI_BEAM_ROTATION = "Multi-beam rotation"
+
+# Scan gains and offset for the FastEM scanner and descanner
+MD_SCAN_OFFSET = "Scan offset"
+MD_SCAN_GAIN = "Scan gain"
+MD_SCAN_GAIN_DYNAMIC = "Dynamic scan gain"
```
pieleric, line 218: Could you add a comment after each line describing the type. Is that a single float, or a list? Is there a fixed number of entries?

pieleric (changes requested):

src/odemis/model/_metadata.py
tepals: @pieleric I've updated it, can you have another look?

PR #1916, opened 2021-11-09
util saconfig: really set auto-zero option

pieleric:
We did compute the right value... but never store it.
So autozero behaved as a normal referencing.


lazem (approved):
Looks good.

PR #1919, opened 2021-11-12
remove cryo-secom role from test files

Mahmood-B:
After merging the meteor chamber PR, the roles become 'enzel' and 'meteor' explicitly, so having 'cryo-secom' role in the testing will always raise errors.


Mahmood-B:
A question. I am wondering if the integration tests cause issues in general. Because when i run these test cases in the commit for testing the backend starter, sometimes errors are raised when calling stop_backend(). Does this also happen in integration test?

PR #1920, opened 2021-11-12
auto update focuser position

Mahmood-B:
I ran the test cases, they ran without errors. 
I added one test case. Kind of a trick test. I am not sure if it is effective, and if it guarantees that the test on hardware will work fine.

### DIFF
--- a/src/odemis/driver/smaract.py
+++ b/src/odemis/driver/smaract.py
@@ -2644,6 +2644,8 @@ class MCS2(model.Actuator):
             else:
                 logging.warning("SA_CTL is not referenced. The device will not function until referencing occurs.")
 
+        self._focuser_pos_updater = RepeatingTimer(1.0, self._updatePosition)
+        self._focuser_pos_updater.start()
         self._updatePosition()
 
         self.speed = VigilantAttribute({}, unit="m/s", readonly=True)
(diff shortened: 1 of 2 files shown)


src/odemis/driver/smaract.py
```diff
@@ -2644,6 +2644,8 @@ def __init__(self, name, role, locator, ref_on_init=False, axes=None, speed=1e-3
             else:
                 logging.warning("SA_CTL is not referenced. The device will not function until referencing occurs.")
 
+        self._focuser_pos_updater = RepeatingTimer(1.0, self._updatePosition)
+        self._focuser_pos_updater.start()
```
pieleric, line 2648: You should also stop it in `terminate()`. You can do `self._update_position_timer.cancel()`.

src/odemis/driver/test/smaract_test.py
```diff
@@ -601,6 +601,17 @@ def test_reference_and_deactivate_move(self):
...
 
+    def test_auto_update_function(self):
+        self.dev.moveAbs({"x": 0.0, "y": 0.0, "z": 0.0}).result()
+        pos_before_move = self.dev.position.value
+        rel_move = {"z": 5e-6}
+        self.dev.moveRel(rel_move).result()
+        # the position updater function is called every 1 sec, wait a bit more.
+        time.sleep(2)
+        pos_after_move = self.dev.position.value
+        expected_pos = pos_before_move.copy()
```
pieleric, line 612: For safety, compute the expected_pos *before* calling `moveRel()`.

pieleric (changes requested):

PR #1922, opened 2021-11-15
driver semcomedi: run the garbage collector while waiting the for the acquisition

pieleric:
We used to run the garbage collector between two acquisitions.
For short acquisitions (eg, 10ms), as a GC run takes about 15 ms, that
would be quite costly. Especially, during a GC run, the entire Python
interpreter is non-responsive. So if during that time came a request
just to read or write any VA, that request would be blocked.

This caused the SPARC acquisition to get ~15ms overhead after every
acquisition.

=> Run the GC while we are waiting for the hardware. During that time,
we really have not much to do. Also limit the number of collections, and
if the acquisition is short, only do a simple collection.


VictoriaDel (approved):
It seems good to me ! :)

PR #1925, opened 2021-11-16
technolution driver: change definition of scanGain to scanAmplitude

Sabrina85:
The scanGain defined the end of the scanning ramp (offset + amplitude). So whenever the user wanted to change the amplitude, they needed to add the offset. Or whenever the offset was changed, the gain needed to be changed accordingly. Redefine the scanGain to scanAmplitude. Change the behavior of the code, so that the scanAmplitude is truely independent of the scanOffset.

* rename scanGain to scanAmplitude
* split up the test cases to cover a single item
* improved/fixed some tests, comments and doc strings
* fixed typo in EXTERNAL_STORAGE
* added tests for calculation of scanner offset and gradient as defined in the asm
* reordered imports
* made bit range for descanner also global variable similar to volt range for scanner
* improved plotting test cases and changed import for matplotlib to use GUI backend, so plots are shown
* removed the random loops in the tests and replaced by an example for floating point error for x setpoints; didn't find a case for y - could not identify the part of the method, that can introduce such a case
* adjusted the bit range: The range accepted by the ASM is `[-2**15, 2**15 -1]` - so no symmetrical around 0. 0 counts with the positive values. However, the user can select offset and amplitude symmetrically around 0. So what happend was that an amplitude of 0 a.u. was mapped to -0.5 bits and not 0 bits. If the user wants a flat line (no amplitude) there would have been always 1 bit being removed of whatever offset was choosen. So offset + amplitude in bits was not the same value as just the offset. Changed the behavior to map from a.u. [-1, 1] symmetrically to` [-2**15, 2**15]` and in case +1 is selected as amplitude the corresponding bit value 2**15 is reduced by 1 bit to match the requirements of the ASM. This should be no problem as one bit change in the descanner position is of no significance.

### DIFF
--- a/src/odemis/driver/technolution.py
+++ b/src/odemis/driver/technolution.py
@@ -597,9 +597,9 @@ class EBeamScanner(model.Emitter):
         # In x we typically scan from negative to positive centered around zero and
         # in y from positive to negative centered around zero.
         # the start of the sawtooth scanning signal
-        self.scanOffset = model.TupleContinuous((0.0, 0.0), range=((-1.0, -1.0), (1.0, 1.0)))
+        self.scanOffset = model.TupleContinuous((0.0, 0.0), range=((-1.0, -1.0), (1.0, 1.0)), cls=(int, float))
         # heights of the sawtooth scanner signal (it does not include the offset!)
-        self.scanAmplitude = model.TupleContinuous((0.1, -0.1), range=((-1.0, -1.0), (1.0, 1.0)))
+        self.scanAmplitude = model.TupleContinuous((0.1, -0.1), range=((-1.0, -1.0), (1.0, 1.0)), cls=(int, float))
         # FIXME add a check that offset + amplitude >! 2**15 - 1 and offset + amplitude <! -2**15
 
         # delay between the trigger signal to start the acquisition and the scanner to start scanning
@@ -883,9 +883,9 @@ class MirrorDescanner(model.Emitter):
         # direction of the executed descan
         self.rotation = model.FloatContinuous(0, range=(0, 2 * math.pi), unit='rad')
         # start of the sawtooth descanner signal
-        self.scanOffset = model.TupleContinuous((0.0, 0.0), range=((-1, -1), (1, 1)))
+        self.scanOffset = model.TupleContinuous((0.0, 0.0), range=((-1, -1), (1, 1)), cls=(int, float))
         # heights of the sawtooth descanner signal (it does not include the offset!)
-        self.scanAmplitude = model.TupleContinuous((0.008, 0.008), range=((-1, -1), (1, 1)))
+        self.scanAmplitude = model.TupleContinuous((0.008, 0.008), range=((-1, -1), (1, 1)), cls=(int, float))
         # FIXME add a check that offset + amplitude >! 2**15 - 1 and offset + amplitude <! -2**15
 
         clockFrequencyData = self.parent.asmApiGetCall("/scan/descan_control_frequency", 200)  # [1/sec]
--- a/src/odemis/driver/technolution.py
+++ b/src/odemis/driver/technolution.py
@@ -24,33 +24,33 @@ Odemis. If not, see http://www.gnu.org/licenses/.
 # multi-beam project
 from __future__ import division
 
-import math
-import os
-
-import numpy
 import base64
 import json
 import logging
+import math
+import os
 import queue
 import re
 import threading
 import time
-from PIL import Image
 from io import BytesIO
-from urllib.parse import urlparse, urlunparse
+from urllib.parse import urlparse
+
+import numpy
+from PIL import Image
 from requests import Session
 from scipy import signal
 
 from odemis import model
 from odemis.model import HwError
 from odemis.util import almost_equal
-
+from technolution_asm.models.calibration_loop_parameters import CalibrationLoopParameters
+from technolution_asm.models.cell_parameters import CellParameters
 from technolution_asm.models.field_meta_data import FieldMetaData
 from technolution_asm.models.mega_field_meta_data import MegaFieldMetaData
-from technolution_asm.models.cell_parameters import CellParameters
-from technolution_asm.models.calibration_loop_parameters import CalibrationLoopParameters
 
 VOLT_RANGE = (-10, 10)
+I16_SYM_RANGE = (-2 ** 15, 2 ** 15)  # Note: on HW range is not symmetrically (-2**15, 2**15 - 1)
 DATA_CONTENT_TO_ASM = {"empty": None, "thumbnail": True, "full": False}
 RUNNING = "installation in progress"
 FINISHED = "last installation successful"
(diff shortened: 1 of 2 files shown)


src/odemis/driver/technolution.py
```diff
@@ -604,10 +591,15 @@ def __init__(self, name, role, parent, **kwargs):
...
+        # In x we typically scan from negative to positive centered around zero and
+        # in y from positive to negative centered around zero.
         # the start of the sawtooth scanning signal
-        self.scanOffset = model.TupleContinuous((0.0, 0.0), range=((-1.0, -1.0), (1.0, 1.0)))
-        # the end (amplitude) of the sawtooth scanning signal
-        self.scanGain = model.TupleContinuous((0.3, 0.3), range=((-1.0, -1.0), (1.0, 1.0)))
+        self.scanOffset = model.TupleContinuous((0.0, 0.0), range=((-1.0, -1.0), (1.0, 1.0)), cls=(int, float))
+        # heights of the sawtooth scanner signal (it does not include the offset!)
+        self.scanAmplitude = model.TupleContinuous((0.1, -0.1), range=((-1.0, -1.0), (1.0, 1.0)), cls=(int, float))
+        # FIXME add a check that offset + amplitude >! 2**15 - 1 and offset + amplitude <! -2**15
```
pieleric, line 601: Preferably, implement it now. Or what's preventing you to do so? Maybe it's a little tricky to raise error when 2 VAs combined are incompatible. As seen before, it might be better to automatically adjust one of them instead of raising an error.
An alternative would be to just log a warning (though it's a lot less effective to notice that something is wrong).

src/odemis/driver/technolution.py
```diff
@@ -719,45 +710,72 @@ def getTicksDwellTime(self):
...
-        # Number of steps = number of setpoints - 1 = 3
-        # step size = (8 - 2) / 3 = 6 / 3 = 2
-        #################################################################
-        scan_offset_volts = convertRange(self.scanOffset.value, numpy.array(self.scanOffset.range)[:, 1], VOLT_RANGE)
-        scan_gain_volts = convertRange(self.scanGain.value, numpy.array(self.scanGain.range)[:, 1], VOLT_RANGE)
+        # --2-- start = offset (scan_offset)
+        # end = start + amplitude = 2 + 6 = 8
+        # center =  start + end = 2 + 8 = 5
+        ##################################################################
+        # Convert start of the scanning ramp from [a.u.] to [bits].
```
pieleric, line 730: The units are sort of clear, but just be aware that "bits" is not the normal usage. "bit" is normally a unit of storage capacity. Each value here is stored *on* 16 bits. I think that, to be very correct, the original unit is "arbitrary unit mapped between -1 and 1", and the end unit is "arbitrary unit mapped between -2¹⁵ and 2¹⁵".

src/odemis/driver/technolution.py
```diff
@@ -834,130 +852,168 @@ def __init__(self, name, role, parent, **kwargs):
...
-        return numpy.floor(setpoints).astype(int).tolist()
+        # Setpoints need to be integers when send to the ASM. First round down found setpoints to next integer
+        # then convert to int type.
+        # For consistency in the rounding use numpy.floor() to always round down (also for negative numbers):
+        # e.g. int(-3.4) = -3 vs. numpy.floor(-3.4) = -4.
+        setpoints = numpy.floor(setpoints).astype(int).tolist()  # [bits]
+
+        # mapping from a.u. to bits is symmetrically around 0, whereas the range in bits that is accepted by the ASM is
+        # not symetrically around 0 ([-32768, 32767]). Clip 2**15 = 32768 by one bit to 32767 bit.
+        setpoints = [p if p != BIT_RANGE[1] else (BIT_RANGE[1] - 1) for p in setpoints]
```
pieleric, line 937: As you have a numpy array just before, I'd suggest to make it faster by using a numpy operation:
```python
setpoints = numpy.floor(setpoints).astype(int)
setpoints = numpy.minimum(setpoints, BIT_RANGE[1] - 1)
return setpoints.lolist()
```

src/odemis/driver/technolution.py
```diff
@@ -834,130 +852,168 @@ def __init__(self, name, role, parent, **kwargs):
...
+
+        # Setpoints need to be integers when send to the ASM. First round down found setpoints to next integer
+        # then convert to int type.
+        # For consistency in the rounding use numpy.floor() to always round down (also for negative numbers):
+        # e.g. int(-3.4) = -3 vs. numpy.floor(-3.4) = -4.
+        setpoints = numpy.floor(setpoints).astype(int).tolist()  # [bits]
+
+        # mapping from a.u. to bits is symmetrically around 0, whereas the range in bits that is accepted by the ASM is
+        # not symetrically around 0 ([-32768, 32767]). Clip 2**15 = 32768 by one bit to 32767 bit.
+        setpoints = [p if p != BIT_RANGE[1] else (BIT_RANGE[1] - 1) for p in setpoints]
```
pieleric, line 969: Use `numpy.minimum()` for clipping here too.

src/odemis/driver/test/technolution_test.py
```diff
@@ -237,6 +225,8 @@ def test_assembleCalibrationMetadata(self):
         #  dwell_time is decreased. However, for the original maximum dwell time of 1e-4 seconds, this value
         #  needs to be increased on the ASM HW to a value above 9000.
 
+        # FIXME move checks that test the correct calculation of the calibration setpoints to separate test case
```
pieleric, line 228: Could you do it now? Maybe just a matter of adjusting the old `test_convertBits()`?

pieleric (changes requested):

src/odemis/driver/technolution.py
Sabrina85: I didn't do it because of the time that was originally planned for this task. And it took me already way longer. So far this check was also not there and would have also raised an error when acquiring an image. That's why I thought I only put the FIXME and do it in a separate pr. 
I agree, a warning is not useful as we only would notice when the acquisition is failing.
So we could do as you suggested:

1. raise an error (which informs the user what value range is possible with the current value of the other VA) and return the current value (which would be the easiest implementation)
2. return the max/min value that is still possible (but if offset = 1, and the amplitude = -0.5 -> what do you return if the user requests amplitude = 1? `-1 or 0`?) This option needs some more thought I think.

What do you think?

src/odemis/driver/technolution.py
```diff
@@ -24,33 +24,33 @@
...
-
+from technolution_asm.models.calibration_loop_parameters import CalibrationLoopParameters
+from technolution_asm.models.cell_parameters import CellParameters
 from technolution_asm.models.field_meta_data import FieldMetaData
 from technolution_asm.models.mega_field_meta_data import MegaFieldMetaData
-from technolution_asm.models.cell_parameters import CellParameters
-from technolution_asm.models.calibration_loop_parameters import CalibrationLoopParameters
 
 VOLT_RANGE = (-10, 10)
+BIT_RANGE = (-2**15, 2**15)  # Note: on HW range is not symmetrically (-2**15, 2**15 - 1)
```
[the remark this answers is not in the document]
Sabrina85: I like the `I16_SYM_RANGE`. The `VOLT_BITS_RANGE` is confusing, as we convert from `a.u.` to `volt` only for the scanner but from `a.u.` to `bits` for the descanner.

src/odemis/driver/technolution.py
Sabrina85: True. I will see if I can clarify this somewhere.

src/odemis/driver/test/technolution_test.py
Sabrina85: It needs a test case for each of the getCalibrationSetpoints on the scanner and descanner. I added them.

src/odemis/driver/technolution.py
Sabrina85: I will leave this for a separate pr.

pieleric:
@Sabrina85 , could you rebase your branch on master to fix the conflicts?

Sabrina85:
> @Sabrina85 , could you rebase your branch on master to fix the conflicts?

Yes will do. Please do not merge yet! I first need to make a corresponding pr for the fastem calibrations.

PR #1926, opened 2021-11-17
Add the Chamber component to the xt-only yaml file

tepals:


install/linux/usr/share/odemis/hwtest/xt-only.odm.yaml
```diff
@@ -18,6 +18,7 @@ SEM: {
     focus: "EBeam Focus",
     stage: "EBeam Stage",
     detector: "EBeam Detector",
+    chamber: "Chamber",
```
Sabrina85, line 21: You need to add the chamber also at the top in the list of children.

Sabrina85 (changes requested):
Btw I noticed that all the indentations are still wrong - the chamber is now the only correct one :) Leave it up to you if you want to fix it now or in a separate pr.

tepals:
> Btw I noticed that all the indentations are still wrong - the chamber is now the only correct one :) Leave it up to you if you want to fix it now or in a separate pr.

fixed

Sabrina85 (approved):

pieleric (approved):
Looks fine to me.
However, you've got conflicts against master, because very recently we've merged another change adding the fib. That said, I now wonder if it was a good idea to add the fib, because I don't know if a system without fib will work. So possibly, at the same time as you fix the conflicts, you could comment out the fib support.

tepals:
@pieleric I rebased and commented out the FIB scanner, can you have a last look and merge it?

install/linux/usr/share/odemis/hwtest/xt-only.odm.yaml
```diff
@@ -1,52 +1,57 @@
...
 # Delmic XTAdapter. Update the address to match that of the Support PC.
 "SEM EXT": {
-  class: xt_client.SEM,
-  role: null,
-  init: {
-    address: "PYRO:Microscope@localhost:4242",
-  },
-  children: {
-    scanner: "EBeam Scanner",
-    fib-scanner: "FIB Scanner",
```
pieleric, line 18: Please leave this line as comment too. "fib-scanner" is the important part actually.

pieleric (changes requested):

install/linux/usr/share/odemis/hwtest/xt-only.odm.yaml
tepals: done

PR #1927, opened 2021-11-17
technolution driver: only max of 50 characters allowed for filename

Sabrina85:


tepals (approved):

src/odemis/driver/technolution.py
```diff
@@ -1438,6 +1438,11 @@ def _setFilename(self, filename):
             raise ValueError("Filename %s contains invalid characters. Only the following characters are allowed: "
                              "'%s'." % (filename, ASM_FILE_CHARS[1:-2]))
 
+        # basename cannot be > 50 characters
+        if len(os.path.basename(filename)) > 50:
+            raise ValueError("Filename %s contains more than 50 characters. Maximum of 50 characters is allowed: "
+                             "'%s'." % (os.path.basename(filename), ASM_FILE_CHARS[1:-2]))
```
pieleric, line 1444: I don't think the second part of the error message is useful here. Instead, it should say how many characters the filename currently has.

pieleric (changes requested):

src/odemis/driver/technolution.py
Sabrina85: True, that was a stupid copy/paste mistake.

PR #1931, opened 2021-11-18
gui cont tabs CryoChamberTab: on METEOR also ask confirmation to move to SEM

pieleric:
We ask before moving to FM, let's do the same when requesting to move to SEM.

### DIFF
--- a/src/odemis/gui/cont/tabs.py
+++ b/src/odemis/gui/cont/tabs.py
@@ -3008,7 +3008,8 @@ class CryoChamberTab(Tab):
             self._end_pos = getTargetPosition(self._target_position, self._stage)
             if not self._end_pos:
                 return
-            if self._target_position == FM_IMAGING and current_position in [LOADING, SEM_IMAGING]:
+            if (self._target_position in [FM_IMAGING, SEM_IMAGING] and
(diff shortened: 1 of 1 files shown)


tepals (approved):

src/odemis/gui/cont/tabs.py
```diff
@@ -3008,7 +3008,8 @@ def _perform_switch_position_movement(self, target_button):
             self._end_pos = getTargetPosition(self._target_position, self._stage)
             if not self._end_pos:
                 return
-            if self._target_position == FM_IMAGING and current_position in [LOADING, SEM_IMAGING]:
+            if (self._target_position in [FM_IMAGING, SEM_IMAGING] and
+                 current_position in [LOADING, SEM_IMAGING, FM_IMAGING]):
                 if not self._display_meteor_pos_warning_msg(self._end_pos):
```
lazem, line 3013: Note that `_display_meteor_pos_warning_msg` has the caption "Moving to METEOR". So maybe it could be better changed to "Stage Movement" or something similar..

lazem (approved):

src/odemis/gui/cont/tabs.py
pieleric: Good catch. Updated to "Large stage move".

PR #1933, opened 2021-11-22
technolution driver: convert physical flyback time to VA

Sabrina85:
It turned out, that it is quite useful to adjust the flyback time during testing, debugging and also to find causes for the image artifacts. Make it a VA to be easy accessible. In practice later a "good" value will be found and will most likely be the same for all EA systems.

### DIFF
--- a/src/odemis/driver/technolution.py
+++ b/src/odemis/driver/technolution.py
@@ -843,7 +843,7 @@ class MirrorDescanner(model.Emitter):
         self.clockPeriod = model.FloatVA(1 / clockFrequencyData['frequency'], unit='s', readonly=True)
 
         # physical time for the mirror descanner to perform a flyback (moving back to start of a line scan)
-        self.physicalFlybackTime = 250e-6  # assumed constant [s]
+        self.physicalFlybackTime = model.FloatContinuous(250e-6, range=(0, 1e-3), unit='s')
 
     def getXAcqSetpoints(self):
         """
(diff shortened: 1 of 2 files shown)


src/odemis/driver/technolution.py
```diff
@@ -843,7 +843,7 @@ def __init__(self, name, role, parent, **kwargs):
         self.clockPeriod = model.FloatVA(1 / clockFrequencyData['frequency'], unit='s', readonly=True)
 
         # physical time for the mirror descanner to perform a flyback (moving back to start of a line scan)
-        self.physicalFlybackTime = 250e-6  # assumed constant [s]
+        self.physicalFlybackTime = model.FloatContinuous(250e-6, range=(0, 1e-3), unit='s')
```
tepals, line 846: Since you only want to change this value during debugging you could make it readonly. With readonly it's still possible to force write the value.

```
readonly (bool): if True, value setter will raise an exception. It's still
    possible to change the value by calling _set_value(..., force_write=True).
```

pieleric: Then it would only be possible to change it via a python terminal, and actually I have never tried it remotely. I guess it should work, but you want to double check.

Sabrina85: I was considering it, but then decided to make it a normal VA as it feels like less hassle for using it. We also made the overvoltage VA a normal VA, which also stays quite fixed and usually is also only used for debugging/testing.

Sabrina85: @pieleric Does `force_write` not work in a script such as the calibrations?

pieleric: I don't know! You can try and let me know 😉

PR #1934, opened 2021-11-22
technolution driver: increase the range for the scanner and acquisition delay

Sabrina85:


src/odemis/driver/technolution.py
```diff
@@ -708,10 +707,12 @@ def _calc_calibration_sampling_period(self, total_line_scan_time, system_clock_p
 
     def getTicksScanDelay(self):
         """
-        :return: Scan delay in multiple of ticks of the ebeam scanner clock frequency
+        Convert the scan delay in seconds into a multiple of the system clock period in ticks.
+        Convert the number of pre-scanned lines to integer.
+        :return (int, int): Scan delay in ticks. Number of line scans before starting the actual acquisition.
```
pieleric, line 712: It's not clear that the first sentence is for the first int, and the second sentence for the second int.
I'd suggest to write on multiple lines, like:
```python
return:
   x_delay (int): Scan delay in ticks.
   y_delay (int): Number of line scans before starting the actual acquisition.
```

src/odemis/driver/technolution.py
```diff
@@ -611,8 +611,7 @@ def __init__(self, name, role, parent, **kwargs):
         # delay between the trigger signal to start the acquisition and the scanner to start scanning
         # x: delay in starting a line scan; y: delay in scanning full lines (prescan lines)
         # TODO: y scan delay is y prescan lines which is currently unused an can probably be deleted.
-        # The scanDelay in x direction maximum (200e-6) is experimentally determined.
-        self.scanDelay = model.TupleContinuous((0.0, 0.0), range=((0.0, 0.0), (200e-6, 10.0)), unit='s',
+        self.scanDelay = model.TupleContinuous((0.0, 0.0), range=((0.0, 0.0), (1.0, 10.0)), unit='s', cls=(int, float),
```
pieleric, line 614: This is really confusing: the X is in seconds, but the Y in some "number of lines". However the unit is "s". In addition, Y is actually not implemented, but it's allowed to put up to 10. 

If it's not implemented, I'd suggest to just put as maximum range "0", to make it clear it's not possible to pass anything else (and expect it to be working). If you really want to support it, then I'd suggest to accept seconds, and convert from seconds to line scan in getTicksScanDelay().

src/odemis/driver/technolution.py
```diff
@@ -985,7 +986,7 @@ def __init__(self, name, role, parent, **kwargs):
         self.filename = model.StringVA("storage/images/date/project/megafield_id", setter=self._setFilename)
         self.dataContent = model.StringEnumerated('empty', DATA_CONTENT_TO_ASM.keys())
         # delay between the trigger signal to start the acquisition, and the start of the recording by the mppc detector
-        self.acqDelay = model.FloatContinuous(0.0, range=(0, 200e-6), unit='s', setter=self._setAcqDelay)
+        self.acqDelay = model.FloatContinuous(0.0, range=(0.0, 1.0), unit='s', setter=self._setAcqDelay)
```
pieleric, line 989: I'd suggest to add a comment here that acqDelay depends on scanDelay. If increasing it, scanDelay should be increased first.

pieleric (changes requested):

PR #1935, opened 2021-11-23
config: update meteor sim

pieleric:
Adjust positions so that the initial position (0,0,0,0,0) happens to be
the SEM position. That's not very realistic, but typically on a real
system, when Odemis start the stage is already in the SEM position.
Also much more handy for testing.

### DIFF
--- a/install/linux/usr/share/odemis/sim/meteor-sim.odm.yaml
+++ b/install/linux/usr/share/odemis/sim/meteor-sim.odm.yaml
@@ -4,10 +4,15 @@
 METEOR-Sim: {
     class: Microscope,
     role: meteor,
-    children: ["Optical Objective",
-               "Optical Focus",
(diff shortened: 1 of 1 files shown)


install/linux/usr/share/odemis/sim/meteor-sim.odm.yaml
```diff
@@ -189,14 +197,20 @@ METEOR-Sim: {
...
+                range: [-2.e-3, 20.e-3],
                 unit: 'm',
                 channel: 0,
             },
         },
     },
     metadata: {
-        FAV_POS_DEACTIVE: {'z': 0.002},
-        FAV_POS_ACTIVE: {'z': 0.00}
+        # Loading position to retract lense
```
lazem, line 209: Type in lense -> lens

lazem (approved):
Just a small type otherwise it looks good.

PR #1936, opened 2021-11-23
acq move: simplify _doCryoSwitchSamplePosition() and related functions

pieleric:
The code was very hard to confirm that every case would be handled.
Now, it's clear what happens.
In particular, it should now handle cases where the current position is
UNKNOWN.

Also merge METEOR and ENZEL constants.


lazem (approved):
Looks good.

PR #1937, opened 2021-11-23
acq move test: simplify back-end startup

pieleric:
Now that test.start_backend() automatically stop the current backend and
starts the right one, the code can be simplified a lot.

### DIFF
--- a/src/odemis/acq/test/move_test.py
+++ b/src/odemis/acq/test/move_test.py
@@ -17,22 +17,21 @@ Odemis. If not, see http://www.gnu.org/licenses/.
 """
 import copy
 import logging
-import os
-import time
-import unittest
-
-import scipy, numpy
-
-import odemis
+import numpy
 from odemis import model
 from odemis import util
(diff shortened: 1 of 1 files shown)


src/odemis/acq/test/move_test.py
```diff
@@ -49,15 +48,7 @@ class TestCryoMove(unittest.TestCase):
...
-        try:
-            test.start_backend(ENZEL_CONFIG)
-        except LookupError:
-            logging.info("A running backend is already found, skipping tests")
-            cls.backend_was_running = True
-            return
-        except IOError as exp:
-            logging.error(str(exp))
-            raise
+        test.start_backend(ENZEL_CONFIG)
```
Kleijwegt, line 51: Just a note:
In case the correct back-end is already running this is logged by _src/odemis/util/test.py_, although not very explicit this should be enough for the user of this function I think.

Kleijwegt (approved):

PR #1946, opened 2021-11-29
Smaract update 5DOF simulator range

lazem:
As the `FakeMC_5DOF_DLL` range is hard coded, changes to the simulator range might result in "Position not in range error".

Plus some miscellaneous refactoring for `MC_5DOF, FakeMC_5DOF_DLL & MCS2` drivers

### DIFF
--- a/src/odemis/driver/smaract.py
+++ b/src/odemis/driver/smaract.py
@@ -1804,14 +1804,15 @@ class FakeMC_5DOF_DLL(object):
         self._pivot = SA_MC_Vec3()
 
         # Specify ranges
-        # Mirroring actual ranges
-        self._range = {}
-        self._range['x'] = (-1.6e-2, 1.6e-2)
-        self._range['y'] = (-1.5e-2, 1.5e-2)
-        self._range['z'] = (-1.e-2, 0.002)
-        self._range['rx'] = (-28, 28)
-        self._range['ry'] = (0, 0)
-        self._range['rz'] = (-28, 28)
+        # Make sure these ranges mirroring the simulator ranges (ie on the yaml file)
+        self._range = {
+            'x': (-1.6e-2, 1.6e-2),
+            'y': (-1.5e-2, 1.5e-2),
+            'z': (-1.0e-2, 0.003),
+            'rx': (-28, 28),
+            'ry': (0, 0),
+            'rz': (-28, 28),
+        }
 
         self.stopping = threading.Event()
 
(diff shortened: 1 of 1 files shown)


src/odemis/driver/smaract.py
```diff
@@ -1195,11 +1195,7 @@ def __init__(self, name, role, locator, axes, ref_on_init=False, linear_speed=0.
         pos_deactive_after_ref (bool): if True, will move to the deactive position
             defined in metadata after referencing
         """
-        if locator != "fake":
-            self.core = MC_5DOF_DLL()
-        else:
-            self.core = FakeMC_5DOF_DLL()
-
+        self.core = MC_5DOF_DLL() if locator != "fake" else FakeMC_5DOF_DLL()
```
pieleric, line 1198: It's probably a matter of taste, and I a bit old-school, but I don't think this helps reading.

src/odemis/driver/smaract.py
```diff
@@ -1804,14 +1798,15 @@ def __init__(self):
...
-        self._range['rz'] = (-28, 28)
+        # Make sure these ranges mirroring the simulator ranges (ie on the yaml file)
+        self._range = {
+            'x': (-1.6e-2, 1.6e-2),
+            'y': (-1.5e-2, 1.5e-2),
+            'z': (-1.0e-2, 0.003),
+            'rx': (-28, 28),
+            'ry': (0, 0),
+            'rz': (-28, 28),
+        }
```
pieleric, line 1809: Instead of increasing the range to adjust to the ENZEL, we could automatically adjust it to fit the range requested.
Take a "rng" (optional) argument. Then update the _range to increase it if needed. Something like:
```python
for a, r in rng.items():
    orig_rng = self._range[a]
    self._range[a] = (min(orig_rng[0], r[0]), max(orig_rng[1], r[1]))
```
The MC_5DOF class can pass the range when it's a simulator. It's a bit cheating, but that's what the simulator is all about anyway ;-)

src/odemis/driver/smaract.py
```diff
@@ -3171,7 +3162,10 @@ def _doReference(self, future, axes):
                         # TODO: Raise some error here
 
                 # if referenced, move to the safe position (if requested)
-                all_axes_referenced = all([self.referenced._value[a] for a in self._axis_map.keys()])
+                all_axes_referenced = all(
+                    self.referenced._value[a] for a in self._axis_map.keys()
```
pieleric, line 3166: You can actually also drop the `.keys()`. Then it would likely fit on a single line.

pieleric (changes requested):

pieleric (approved):

PR #1948, opened 2021-11-30
Reorder ENZEL sub-moves

lazem:
Reorder sample stage & aligner sub-movement for safer positioning.
Plus refactor the sub-moves sequence code to make it more readable.


Kleijwegt (approved):
Assuming Daan has checked this the order of movements in this code, I think it looks all good and ready to merge.

pieleric (approved):

PR #1955, opened 2021-12-08
METEOR: Fix mirroring equation using dict axis keys instead of index number

lazem:
Fix issue introduced in PR #1929


pieleric:
I don't get it. Shouldn't the MD_POS be a tuple? that was the point of the previous PR. If it's a dict, fix the microscope file.

lazem:
> I don't get it. Shouldn't the MD_POS be a tuple? that was the point of the previous PR. If it's a dict, fix the microscope file.

That's correct, but `pos` is the actual position dict, the bug was changing both variables to arrays while it should have been only `pos_cor`. @pieleric

PR #1957, opened 2021-12-08
Support 3 axes on ConvertStage

lazem:
Adjust ConvertStage so its 2 axes logic can be overridden from child class Convert3DStage with an additional axis Z.

Follows the [Euler angle](https://www.wikiwand.com/en/Euler_angles) rotation system. 
Reference for rotation matrices: https://nl.mathworks.com/help/phased/ref/rotx.html#btq5fj0-5


src/odemis/driver/actuator.py
```diff
@@ -760,10 +760,119 @@ def stop(self, axes=None):
...
+                self.speed = model.VigilantAttribute({}, readonly=True)
+                self._dependency.speed.subscribe(self._updateSpeed, init=True)
+            else:
+                logging.info("Axes %s of dependency are missing from .speed, so not providing it",
+                             set(axes) - speed_axes)
+
+    def _get_rot_matrix(self, invert=False):
+        # Overrides parent class method with rotation matrices for the 3 angles
+        # NB: Rotation is counterclockwise
+        # TODO: handle rotation for the 3 angles at the same time
```
pieleric, line 824: Shouldn't the dot matrix multiplication be enough? What currently doesn't work with multiple angles?

src/odemis/driver/test/actuator_test.py
```diff
@@ -669,6 +669,278 @@ def test_move_abs(self):
...
+        test.assert_pos_almost_equal(stage.position.value, {"x": 0, "y": 0, "z": 0})
+        f = stage.moveRel({"x": 1e-06, "y": 2e-06, "z": 0})
+        f.result()
+        self.assertEqual(stage.position.value, {"x": 1e-06, "y": 2e-06, "z": 0})
+        test.assert_pos_almost_equal(dependency.position.value, {"x": -2e-06, "y": 1e-06, "z": 0})
+        f = stage.moveRel({"x": -1e-06, "y": -2e-06})
+        f.result()
+        test.assert_pos_almost_equal(stage.position.value, {"x": 0, "y": 0, "z": 0})
+        test.assert_pos_almost_equal(dependency.position.value, {"x": 0, "y": 0, "z": 0})
+
```
pieleric, line 732: That's fine, but could you add another one testing another dimension, to check it does the right thing also in "3D"? At least, just ry or rx. Ideally, you could check multiple axes...

src/odemis/driver/actuator.py
```diff
@@ -760,10 +760,119 @@ def stop(self, axes=None):
...
+        assert len(axes) == 3
+        if len(dependencies) != 1:
+            raise ValueError("ConvertStage needs 1 dependency")
+
+        self._dependency = list(dependencies.values())[0]
+        self._axes_dep = {"x": axes[0], "y": axes[1], "z": axes[2]}
+        if scale is None:
+            scale = (1, 1, 1)
+        if translation is None:
+            translation = (0, 0, 0)
```
pieleric, line 788: Why not directly indicating these default values in the signature too? As a tuple cannot be changed, it should be safe.

pieleric (changes requested):
It looks fine. I only have minor comments. Indeed using inheritance allows to separate well the 2D vs the 3D cases.
One thing that could be possible is to use `__new__()` so that ConvertStage automatically provides a Convert3DStage if passed 3 axes. This would hide the implementation detail to the user. What do you think?

src/odemis/driver/test/actuator_test.py
lazem: There are tests in `test_ab_rotation` for the 3 angles. I have also added them on `test_move_rel`.

src/odemis/driver/actuator.py
lazem: I tried to use more than 1 angle at the same time, the issue I have is that the convert stage doesn't reach the intended position. I have tried to use the rotation matrix as it is, as well as trying to invert the order this way, but had the same issue:
```
 if invert:
    rx_mat @ ry_mat @ rz_mat
```
	
Btw, the issue I was referring to in our talk that could be related is called Gimbal lock and it's described here:
https://omar-shehata.medium.com/how-to-fix-gimbal-lock-in-n-dimensions-f2f7baec2b5e

pieleric: Strange. I've just tried, and for me it looked like it works.
```
      if invert:
            rotation = tuple(r*-1 for r in rotation)
:
      if invert: 
          return rx_mat @ ry_mat @ rz_mat
      else:
          return rz_mat @ ry_mat @ rx_mat
```

Even if this doesn't work, I'd suggest to leave it like that, because that should be the closest to the "correct way".

pieleric (changes requested):

src/odemis/driver/actuator.py
```diff
@@ -594,6 +594,13 @@ class ConvertStage(model.Actuator):
     the two stages, as for each SEM stage move it is able to perform the
     corresponding “compensate” move in objective lens.
     """
+    def __new__(cls, *args, **kwargs):
+        # Automatically create Convert3DStage object if the number of axes = 3
+        axes = kwargs['axes']
+        if len(axes) == 2:
+            return super(ConvertStage, cls).__new__(cls)
+        elif len(axes) == 3:
+            return super(ConvertStage, cls).__new__(Convert3DStage)
```
pieleric, line 603: We need to raise a ValueError in other cases explaining the number of axis is incorrect. Otherwise `__new__()` returns None and so `None` would be used as instance, creating a lot of confusion on what is happening.

pieleric (changes requested):
Could you also update the enzel  (or Meteor) simulator to use this new class. This way we already have a user :-)

pieleric (approved):

effting:
The transform classes in `odemis.util.transform` currently only support 2D geometric transforms. It would be nice if `ConvertStage` would make use of it instead of duplicating code. We would need to extend the transform classes to be able to support 3D rotations if we want `ConvertStage3D` to follow the same route.

src/odemis/driver/actuator.py
```diff
@@ -716,53 +727,160 @@ def _updateReferenced(self, dep_refd):
...
+    """
+    Extends original ConvertStage with an additional axis Z
+    """
+    def __init__(self, name, role, dependencies, axes,
+                 rotation=(0, 0, 0), scale=(1, 1, 1), translation=(0, 0, 0), **kwargs):
+        """
+        dependencies (dict str -> actuator): name to objective lens actuator
+        axes (list of 3 strings): names of the axes for x, y and z
+        scale (None tuple of 3 floats): scale factor from exported to original position
+        rotation (tuple of 3 floats): rotation factor rz, ry, rx (in radians)
```
effting, line 784: For `rotation` please mention that these are Tait–Bryan angles using extrinsic rotations

See: https://en.wikipedia.org/wiki/Euler_angles

src/odemis/driver/actuator.py
```diff
@@ -716,53 +727,160 @@ def _updateReferenced(self, dep_refd):
...
+
+        axes_def = {"x": self._dependency.axes[axes[0]],
+                    "y": self._dependency.axes[axes[1]],
+                    "z": self._dependency.axes[axes[2]]}
+        model.Actuator.__init__(self, name, role, dependencies=dependencies, axes=axes_def, **kwargs)
+
+        self._metadata[model.MD_POS_COR] = translation
+        self._metadata[model.MD_ROTATION_COR] = rotation
+        self._metadata[model.MD_PIXEL_SIZE_COR] = scale
+        self._updateConversion()
```
effting, line 804: Aren't (or shouldn't) MD_ROTATION_COR and MD_PIXEL_SIZE_COR be properties of the imaging system instead of the stage?

lazem: I think the reason of using them in the stage is just reuse the metadata that has `correction` in its name, to not recreate new metadata, is that correct @pieleric ?

lazem:
Thanks for the review@effting! I replaced the 2D rotation matrix with the one found in `util.transform` as you advised. I think extending the module with 3D transform would make another interesting maintenance task. :)

src/odemis/driver/actuator.py
pieleric: We reuse these metadata (and the metadata API in general) as a way to reconfigure dynamically the wrapper. It's actually not something introduced by Bassim, it was already there for a long while...
Not that typically the values are not changed frequently, it's just during calibration, at system install, or during alignment.

PR #1960, opened 2021-12-13
driver andorcam2: fix camera constantly acquiring and refactor acquisition loop

pieleric:
Separate the "Stopped" and "Acquiring" states into two functions.
Most importantly, this simplifies the code to stop the camera, and fixes
the fact that the camera never stopped acquiring due to the wrong
indentation level for the stop code.

### DIFF
--- a/src/odemis/driver/andorcam2.py
+++ b/src/odemis/driver/andorcam2.py
@@ -2176,118 +2176,137 @@ class AndorCam2(model.DigitalCamera):
         """
         try:
             self.select()  # Make sure we are using the right camera
-            has_hw_lock = False
-
             while True: # Waiting/Acquiring loop
                 # Wait until we have a start (or terminate) message
                 self._acq_wait_start()
 
-                # Preparation & acquisition loop (until stop requested)
-                logging.debug("Preparing acquisition")
-                need_reconfig = True  # Always reconfigure when new acquisition starts
-                failures = 0
-                while True:  # Acquire one image
-                    if self.request_hw: # TODO: check if we still need this code.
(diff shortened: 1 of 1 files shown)


src/odemis/driver/andorcam2.py
```diff
@@ -2176,118 +2176,132 @@ def _acquire(self):
...
-                        failures = 0
 
-                    logging.debug("image acquired successfully after %g s", time.time() - tstart)
-                    self.data.notify(self._transposeDAToUser(array))
-                    del cbuffer, array
+                    im_res, duration, trigger_mode = self._update_settings()
+                    if not has_hw_lock:
+                        self.hw_lock.acquire()
+                        has_hw_lock = True
+                    if trigger_mode != TRIG_FAKE:  # TRIG_FAKE uses StartAcquisition to simulate trigger
```
Sabrina85, line 2228: Can you remind me why you call startAcquisition when the trigger mode is not fake trigger? The comment suggests the opposite and below in the related elif statement in line 2245 you call the same but when trigger mode equals fake trigger.
Why was it not enough to only start the acquisition (after reconfiguration) in the next block?

pieleric: Normally, we call StartAcquisition() just after reconfiguration, only. As it's in continuous mode, the images will keep coming.
With TRIG_FAKE, we set it to single frame, and use StartAcquisition() to acquire a single image, at the moment the trigger event is received.

I've updated the comment a little to clarify.

Sabrina85 (approved):
Looks good now :)

PR #1963, opened 2021-12-14
WIP: Full workflow test branch

Kleijwegt:


pieleric (changes requested):

src/odemis/gui/cont/tabs.py
```diff
@@ -4398,7 +4404,447 @@ def _on_align_move_done(self, f=None):
...
+            ),
+        ])
+        self.view_controller = viewcont.ViewPortController(tab_data, panel, vpv)
+
+        # TODO K.K. check this : Someone once said don't include a focusser because it could ruin the alignment. This
+        #  is not as straightforward as it seems. The focus should not mess with the 5DOF alignment but it should
+        #  focus using the 5 DOF (not the 3DOF). Plus it is akward for a user that you can control both stages in a
+        #  mode. (The idea was that each mode is focussed to one part of the alignment and doesn't move multiple
+        #  parts.) Daan Boltje has a video on this an knows more about this. Please discuss this with both Daan
+        #  Boltje, Caspar and maybe check this with Kornee before merging.
```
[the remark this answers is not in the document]
Kleijwegt: I think nothing is the best. During the testing of the workflow it didn't seem like a big issue. However before removing this comment I will check if that is correct with Marre and Caspar.

src/odemis/driver/scanner.py
```diff
@@ -86,17 +90,28 @@ def __init__(self, name, role, dependencies, **kwargs):
...
                 setattr(self, vaname, va)
             elif model.hasVA(self._external, vaname):
                 va = getattr(self._external, vaname)
                 setattr(self, vaname, va)
 
+        self.blanker = model.VAEnumerated(
+            None,
+            setter=self._setBlanker,
+            choices={True: 'blanked', False: 'unblanked', None: 'auto'})
+
```
pieleric, line 105: You should do the blanker only if the "internal" has a blanker. Possibly, you could even check if "external" has already a blanker with None, in which case all this is not needed, and .blanker should be just a copy of external.blanker.

src/odemis/driver/scanner.py
```diff
@@ -86,17 +90,28 @@ def __init__(self, name, role, dependencies, **kwargs):
...
         # TODO: if blanker has True/False (only), add a None (=auto), which
         # automatically put the underlying value based on the detector acquisition.
```
pieleric, line 107: You can drop this TODO, as you're implementing it.

src/odemis/acq/stream/_live.py
```diff
@@ -84,14 +88,30 @@ def _onActive(self, active):
...
         if active:
-            if not self._prepared:
-                logging.debug("Preparing stream before activating it as it wasn't prepared")
-                self._prep_future = self._prepare()
-                self._prep_future.add_done_callback(self._startAcquisition)
+            if self.inactive_on_next_image.value:
+                # Check if the stream itself has a dedicated get single frame method
+                if hasattr(self, "getSingleFrame"):  # TODO K.K. extra check if this method is callable
+                    single_frame_func = self.getSingleFrame
+                else:
```
pieleric, line 96: It would be nice if you could avoid this check. Maybe by using inheritance, you could avoid that?

pieleric (changes requested):

src/odemis/driver/scanner.py
Kleijwegt: I think I fixed this but I'm not sure @pieleric can you check?

src/odemis/gui/cont/tabs.py
```diff
@@ -4389,7 +4393,481 @@ def _on_align_move_done(self, f=None):
...
+                view.Shown = False
+        self.view_controller._grid_panel.set_shown_viewports(top['viewport'], bottom['viewport'])
+        self.tab_data_model.visible_views.value = [top['viewport'].view, bottom['viewport'].view]
+
+        # Alter the order of the view_ports so that the correct stream is displayed at top/bottom
+        viewport_list = self.panel.pnl_two_streams_grid.viewports
+        viewport_list.remove(top['viewport'])
+        viewport_list.insert(0, top['viewport'])
+        viewport_list.remove(bottom['viewport'])
+        viewport_list.insert(1, bottom['viewport'])
```
lazem, line 4820: This could be rewritten as `viewport_list = [top['viewport'], bottom['viewport'], viewport_list[2]]`

src/odemis/gui/cont/tabs.py
```diff
@@ -4389,7 +4393,481 @@ def _on_align_move_done(self, f=None):
...
+    def _flm_alignment_controls(self, axis, proportion, evt):
+        """
+        Links the FLM alignments mode control buttons to the stage and updates the correct metadata..
+
+        :param axis (str): axis which is controlled by the button.
+        :param proportion (int): direction of the movement represented by the button (-1/1)
+        :param evt (GenButtonEvent): Clicked event
+        """
+        # Only proceed if there is no currently running target_position
+        if hasattr(self, "_flm_move_future") and self._flm_move_future._state == RUNNING:
```
lazem, line 4769: You could define `_flm_move_future` and its similar others in `_init_` with None, that way you don't need to check its existence every time.

src/odemis/gui/cont/tabs.py
```diff
@@ -4389,7 +4393,481 @@ def _on_align_move_done(self, f=None):
...
+        """
+        beamShiftVA = self._sem_stream.emitter.beamShift
+        try:
+            if "x" in shift:
+                beamShiftVA.value = (beamShiftVA.value[0] + shift["x"], beamShiftVA.value[1])
+        except IndexError:
+            show_message(wx.GetApp().main_frame,
+                         "Reached the limits of the beam shift, cannot move any further in x direction.",
+                         timeout=3.0, level=logging.WARNING)
+            logging.error("Reached the limits of the beam shift, cannot move any further in x direction.")
```
lazem, line 4726: Does the error caught in this IndexError is really about reaching the x limit, or it's using invalid index `x`?

src/odemis/gui/cont/tabs.py
```diff
@@ -4389,7 +4393,481 @@ def _on_align_move_done(self, f=None):
...
+
+        try:
+            if "y" in shift:
+                beamShiftVA.value = (beamShiftVA.value[0], beamShiftVA.value[1] + shift["y"])
+        except IndexError:
+            show_message(wx.GetApp().main_frame,
+                         "Reached the limits of the beam shift, cannot move any further in y direction.",
+                         timeout=3.0, level=logging.WARNING)
+            logging.error("Reached the limits of the beam shift, cannot move any further in y direction.")
```
lazem, line 4735: Same code block as above except in x & y, maybe combine them as function?

src/odemis/gui/cont/tabs.py
```diff
@@ -4389,7 +4393,481 @@ def _on_align_move_done(self, f=None):
...
+        """
+        self.stream_bar_controller.pauseAllStreams()
+        self._set_top_and_bottom_stream_and_settings(self._FLM_view_and_control,
+                                                     self._SEM_view_and_control)
+
+        doc_path = pkg_resources.resource_filename("odemis.gui", "doc/enzel_flm_alignment.html")
+        self.panel.html_alignment_doc.LoadPage(doc_path)
+
+        self.panel.controls_step_size_slider.SetRange(100e-9, 50e-6)
+        self.panel.controls_step_size_slider.set_position_value(10e-6)
```
lazem, line 4753: Maybe put these numerical values as constant?

lazem (changes requested):

src/odemis/gui/cont/tabs.py
```diff
@@ -4389,7 +4393,474 @@ def _on_align_move_done(self, f=None):
...
+                                                        self.tab_data_model,
+                                                        view=bottom['viewport'].view)
+        new_bottom_stream_controller.stream_panel.show_remove_btn(False)
+        self._stream_controllers.append(new_bottom_stream_controller)
+
+        for v in self.tab_data_model.visible_views.value:
+            if hasattr(v, "stream_classes") and isinstance(top["stream"], v.stream_classes):
+                v.addStream(top["stream"])
+                new_top_stream_controller.stream_panel.show_visible_btn(False)
+            elif hasattr(v, "stream_classes") and isinstance(bottom["stream"], v.stream_classes):
```
Kleijwegt, line 4843: @pieleric Do I need to check if each view has the attribute stream_classes or can I assume they have and remove "hasattr(view, "stream_classes")" ?

src/odemis/gui/cont/tabs.py
Kleijwegt: It should be the limit, but I can test that again on the microscope to verify this.

src/odemis/gui/comp/grid.py
```diff
@@ -158,10 +187,11 @@ def on_size(self, _):
         """ Grab the child windows and perform layout when the size changes """
         if self.viewports is None:
             self.viewports = list(self.Children)
-            if len(self.viewports) < 4:
-                # Cannot put them on a grid => default to full screen for the 1st one
+            if len(self.viewports) == 1 or len(self.viewports) == 3:
```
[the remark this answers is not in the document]
Kleijwegt: Actually for 5, 7 etc. only the first four are displayed. I have updated the comment.

src/odemis/gui/cont/tabs.py
Kleijwegt: Then it would actually crash. Currently the if statement assumes the future is containing a real future and None would fail the code. I could check then if the attribute is defined as a future and not as None but that wouldn't decrease lines of code.

lazem: Right, you can do `if self._flm_move_future` but it won't decrease much..

src/odemis/driver/scanner.py
```diff
@@ -112,3 +124,211 @@ def updateMetadata(self, md):
...
+        if not isinstance(intnl, ComponentBase):
+            raise ValueError("Dependency internal is not a component.")
+        if not hasattr(extnl, "data") or not hasattr(extnl, "parent"):
+            # Note: the internal component doesn't need to provide pixelSize
+            raise ValueError("Dependency internal is not a Detector component.")
+        self._internal = intnl
+
+        comp_scanner = dependencies["comp_scanner"]
+        if not isinstance(comp_scanner, ComponentBase):
+            raise ValueError("Dependency internal_scanner is not a component.")
```
pieleric, line 172: Not `internal_scanner`: `comp_scanner` (or scanner).

src/odemis/driver/scanner.py
```diff
@@ -112,3 +124,211 @@ def updateMetadata(self, md):
...
+    '''
+    def __init__(self, name, role, dependencies, **kwargs):
+        '''
+        dependencies (dict string->model.HwComponent): the dependencies
+            There must be exactly two dependencies "external" and "internal".
+        Raise:
+          ValueError: if the dependencies are not compatible
+        '''
+        # we will fill the set of dependencies with Components later in ._dependencies
+        model.Emitter.__init__(self, name, role, dependencies=dependencies, **kwargs)
```
pieleric, line 152: Please use `super()`.

src/odemis/driver/scanner.py
```diff
@@ -112,3 +124,211 @@ def updateMetadata(self, md):
...
+    detector while manipulating the "internal" detector.
+    '''
+    def __init__(self, name, role, dependencies, **kwargs):
+        '''
+        dependencies (dict string->model.HwComponent): the dependencies
+            There must be exactly two dependencies "external" and "internal".
+        Raise:
+          ValueError: if the dependencies are not compatible
+        '''
+        # we will fill the set of dependencies with Components later in ._dependencies
```
pieleric, line 151: This comment doesn't match what you're doing. The next line actually creates a .dependencies VA.

src/odemis/driver/scanner.py
```diff
@@ -112,3 +124,211 @@ def updateMetadata(self, md):
...
+        if not hasattr(extnl, "data") or not hasattr(extnl, "parent"):
+            # Note: the internal component doesn't need to provide pixelSize
+            raise ValueError("Dependency internal is not a Detector component.")
+        self._internal = intnl
+
+        comp_scanner = dependencies["comp_scanner"]
+        if not isinstance(comp_scanner, ComponentBase):
+            raise ValueError("Dependency internal_scanner is not a component.")
+        if not hasattr(comp_scanner, "shape"):
+            # Note: the internal component doesn't need to provide pixelSize
```
pieleric, line 174: Wrong comment?

src/odemis/driver/scanner.py
```diff
@@ -112,3 +124,211 @@ def updateMetadata(self, md):
...
+            raise ValueError("Dependency external is not a Detector component.")
+        self._external = extnl
+
+        intnl = dependencies["internal"]
+        if not isinstance(intnl, ComponentBase):
+            raise ValueError("Dependency internal is not a component.")
+        if not hasattr(extnl, "data") or not hasattr(extnl, "parent"):
+            # Note: the internal component doesn't need to provide pixelSize
+            raise ValueError("Dependency internal is not a Detector component.")
+        self._internal = intnl
```
pieleric, line 168: That internal detector shouldn't be needed. Actually typically the internal scanner has no detector available.

src/odemis/driver/scanner.py
```diff
@@ -112,3 +124,211 @@ def updateMetadata(self, md):
...
+          ValueError: if the dependencies are not compatible
+        '''
+        # we will fill the set of dependencies with Components later in ._dependencies
+        model.Emitter.__init__(self, name, role, dependencies=dependencies, **kwargs)
+
+        # Check the dependencies
+        extnl = dependencies["external"]
+        if not isinstance(extnl, ComponentBase):
+            raise ValueError("Dependency external is not a component.")
+        if not hasattr(extnl, "data") and not hasattr(extnl, "shape"):
```
pieleric, line 158: You can make a single `if`, which check all 3 properties, and fails with "...is not a Detector component".
Also note that hasattr() doesn't really work with Pyro, because Pyro always claims there is something. So you'd need to check that they are present *and* they are of the right type (data -> DataFlowBase, shape -> tuple).

src/odemis/driver/scanner.py
```diff
@@ -112,3 +124,211 @@ def updateMetadata(self, md):
...
+        if not hasattr(comp_scanner, "shape"):
+            # Note: the internal component doesn't need to provide pixelSize
+            raise ValueError("Dependency internal_scanner is not a Emitter component.")
+        if not model.hasVA(comp_scanner, "external"):
+            # Note: the internal component doesn't need to provide pixelSize
+            raise ValueError("Dependency internal_scanner doesn't contain the necessary external VA")
+        self._comp_scanner = comp_scanner
+
+        # Special event to request software unblocking on the scan
+        self.softwareTrigger = self._external.softwareTrigger
```
pieleric, line 182: Only `if model.hasVA(self._external, "softwareTrigger")`.

pieleric (changes requested):

src/odemis/gui/cont/tabs.py
```diff
@@ -4391,7 +4395,475 @@ def _on_align_move_done(self, f=None):
...
+        self._step_size_controls_va_connector = VigilantAttributeConnector(tab_data.step_size,
+                                                                           self.panel.controls_step_size_slider,
+                                                                           events=wx.EVT_SCROLL_CHANGED)
+
+        # Bind the buttons to the button move to custom alignment
+        def __on_click_custom_alignment(evt):
+            # First move the stage then the objective
+            future_stage_move = self._stage.moveAbs(self._stage.getMetadata()[model.MD_FAV_POS_ACTIVE])
+            move_alginer = lambda future: self._aligner.moveAbs(self._aligner.getMetadata()[model.MD_FAV_POS_ACTIVE])
+            future_stage_move.add_done_callback(move_alginer)
```
pieleric, line 4565: Please make this function a standard method. Do not start with 2 `_`. One is enough.

src/odemis/gui/cont/tabs.py
```diff
@@ -4391,7 +4395,475 @@ def _on_align_move_done(self, f=None):
...
+        panel.btn_custom_alignment.Bind(wx.EVT_BUTTON, __on_click_custom_alignment)
+
+        self._align_mode = VAEnumerated(panel.btn_align_z,
+                                        self._align_btn_functions,
+                                        setter=self._set_align_mode)
+        self._set_align_mode(self._align_mode.value)  # Start with the first mode as default
+
+        # Bind the buttons to the align mode VA
+        def __on_click_align_mode_buttons(evt):
+            self._align_mode.value = evt.theButton
```
pieleric, line 4576: Make this a standard method. Start the name with a single `_`.

src/odemis/gui/cont/tabs.py
```diff
@@ -4391,7 +4395,475 @@ def _on_align_move_done(self, f=None):
...
+    def __init__(self, name, button, panel, main_frame, main_data):
+        tab_data = guimod.EnzelAlignGUIData(main_data)
+        super(EnzelAlignTab, self).__init__(name, button, panel, main_frame, tab_data)
+        self.set_label("ALIGNMENT")
+        self._stream_controllers = []
+        self._stage = main_data.stage
+        self._stage_global = main_data.stage_global
+        self._aligner = main_data.aligner
+
+        # Check stage FAV positions in its metadata, and store them in respect to their movement
```
pieleric, line 4417: I don't think the second part of the comment is actually done, is it?

src/odemis/gui/cont/tabs.py
```diff
@@ -4391,7 +4395,475 @@ def _on_align_move_done(self, f=None):
...
+        # Vigilant attribute connectors for the slider
+        self._step_size_controls_va_connector = VigilantAttributeConnector(tab_data.step_size,
+                                                                           self.panel.controls_step_size_slider,
+                                                                           events=wx.EVT_SCROLL_CHANGED)
+
+        # Bind the buttons to the button move to custom alignment
+        def __on_click_custom_alignment(evt):
+            # First move the stage then the objective
+            future_stage_move = self._stage.moveAbs(self._stage.getMetadata()[model.MD_FAV_POS_ACTIVE])
+            move_alginer = lambda future: self._aligner.moveAbs(self._aligner.getMetadata()[model.MD_FAV_POS_ACTIVE])
```
pieleric, line 4564: I don't see when is the aligner FAV_POS_ACTIVE position updated. If it's not updated during alignment then it actually doesn't move back to the previous alignment, but to the factory position, right?

pieleric (changes requested):

install/linux/usr/share/odemis/sim/enzel-sim-xt-analog-digital-composite.odm.yaml
```diff
@@ -0,0 +1,440 @@
...
+    FAV_POS_ACTIVE: { 'z': 0.002 }
+  },
+  persistent: {
+    metadata: [ FAV_POS_ACTIVE ],  # Store "good focus" position between run
+  },
+}
+
+# Special component to only provide the X/Y axis of the optical lens to align it with the e-beam lens
+"Objective Aligner": {
+  class: actuator.MultiplexActuator,
```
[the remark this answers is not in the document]
Kleijwegt: I haven't actually tested on the real Enzel with the 3D convertStage, do we want to implement this (and merge it) without testing it on the real system? (Maybe we should then also update the startup file of the real system)

src/odemis/gui/cont/tabs.py
Kleijwegt: in the methods _flm_alignment_controls and _z_alignment_controls in the (nested)function move_done MD_FAV_POS_ACTIVE is set with these lines:
            

> self._aligner.updateMetadata({model.MD_FAV_POS_ACTIVE: self._aligner.position.value})

and
            

> self._stage.updateMetadata({model.MD_FAV_POS_ACTIVE: updated_stage_pos})

src/odemis/gui/cont/tabs.py
Kleijwegt: No the storing part isn't done and actually should be removed from the comment.

src/odemis/gui/cont/tabs.py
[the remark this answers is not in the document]
Kleijwegt: Since these movements where considered to be always small (because we are already at least in the factory alignment). We (Caspar, Daan and I) decided that we could do the moves in all direction all at once. Of course if something fails the expection should still be logged. can I just add a try - except statement for that or isn't that enough?

src/odemis/driver/scanner.py
```diff
@@ -86,16 +90,24 @@ def __init__(self, name, role, dependencies, **kwargs):
...
             if model.hasVA(self._internal, vaname):
                 va = getattr(self._internal, vaname)
                 setattr(self, vaname, va)
             elif model.hasVA(self._external, vaname):
                 va = getattr(self._external, vaname)
                 setattr(self, vaname, va)
 
-        # TODO: if blanker has True/False (only), add a None (=auto), which
-        # automatically put the underlying value based on the detector acquisition.
+        if model.hasVA(self._external, "blanker") and self._external.blanker.value is None:
```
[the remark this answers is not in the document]
Kleijwegt: Can we check if we solved this comment together?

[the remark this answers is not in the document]
Kleijwegt: Can we check if we solved this comment together?

src/odemis/driver/scanner.py
```diff
@@ -86,16 +90,24 @@ def __init__(self, name, role, dependencies, **kwargs):
...
+            self.blanker = self._external.blanker.value
+        elif model.hasVA(self._internal, "blanker"):
+            self.blanker = model.VAEnumerated(
+                None,
+                setter=self._setBlanker,
+                choices={True: 'blanked', False: 'unblanked', None: 'auto'})
+
+    def _setBlanker(self, blank):
+        self._internal.blanker.value = blank
```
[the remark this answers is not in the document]
Kleijwegt: Can we check if we solved this comment together?

src/odemis/driver/scanner.py
```diff
@@ -112,3 +124,211 @@ def updateMetadata(self, md):
...
     def getMetadata(self):
         return self._external.getMetadata()
+
+    def claimBeam(self, claim):
+        """
+        Used to claim the beam by blanking the beam on the scanner of the XT client when the blanker
+        mode is set to automatic (blanker.value = None)
+        :param claim (boolean): True for unblanking and False for blanking
+        """
+        if self.blanker.value is None:
```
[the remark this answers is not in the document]
Kleijwegt: Can we check if we solved this comment together?

src/odemis/driver/scanner.py
Kleijwegt: So you mean like this:

>         if not isinstance(extnl, ComponentBase)\
>                 and not hasattr(extnl, "data") and isinstance(extnl.data, DataFlowBase)\
>                 and not hasattr(extnl, "shape") and isinstance(extnl.shape, tuple):
>                 raise ValueError("Dependency external is not a Detector component.")

src/odemis/driver/scanner.py
Kleijwegt: In this case the internal scanner has a detector (for the FIB).
But maybe that component is indeed not needed.

src/odemis/gui/cont/tabs.py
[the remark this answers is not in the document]
pieleric: try/except work... if you put them around the `future.result()`. You can also use future.exception() to directly check, which might be easier in a callback function.
For an example, you can look at gui.model.ActuatorGUIData._on_axis_move_done()

src/odemis/driver/scanner.py
pieleric: Approximately like that, but I think you need to tweak the logic to something like that:
```
if not isinstance(extnl, ComponentBase)\
            or not (hasattr(extnl, "data") and isinstance(extnl.data, DataFlowBase))\
            or not (hasattr(extnl, "shape") and isinstance(extnl.shape, tuple)):
            raise ValueError("Dependency external is not a Detector component.")
```

src/odemis/acq/stream/_live.py
```diff
@@ -480,6 +503,20 @@ def __init__(self, name, detector, dataflow, emitter, **kwargs):
...
+        via dataflow.get()
+        """
+        # If multiple emitters are connected to a detector switch to the correct one (e.g. both a FIB and a SEM)
+        if hasVA(self._detector, "scanner") and self._detector.scanner.value != self._emitter.name:
+            self._detector.scanner.value = self._emitter.name
+
+        # update Hw settings to our own ROI
+        self._applyROI()
+        return self._dataflow.get()
+
```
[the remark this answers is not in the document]
Kleijwegt: Nice, I didn't saw that this would simplify it that much.

src/odemis/gui/cont/tabs.py
Kleijwegt: I think it doesn't reduce the number of lines that much plus it would actually reduce the readability of the code.

src/odemis/gui/cont/tabs.py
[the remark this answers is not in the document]
Kleijwegt: I'm a bit confused on how to implement this without having the entire GUI geing stuck on waiting while the move is finished. Can we look at this together?

src/odemis/gui/cont/tabs.py
```diff
@@ -4391,7 +4395,475 @@ def _on_align_move_done(self, f=None):
...
+            future_stage_move.add_done_callback(move_alginer)
+
+        panel.btn_custom_alignment.Bind(wx.EVT_BUTTON, __on_click_custom_alignment)
+
+        self._align_mode = VAEnumerated(panel.btn_align_z,
+                                        self._align_btn_functions,
+                                        setter=self._set_align_mode)
```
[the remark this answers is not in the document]
Kleijwegt: After removing the knowledge about the buttons how do I then make the connection between the mode--> buttons? Via subscribing on this VA? Then how do I link the button press to the setter on the VA mode? Via a dictonary?
I see some solutions but cant come up with something very elegant one, I think you might have already some ideas about this.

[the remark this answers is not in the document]
pieleric: Yes, via a separate dictionary. For an example, see `cont.tabs.Sparc2Align._alignbtn_to_mode`.

src/odemis/driver/scanner.py
Kleijwegt: Git annotate says you wrote this comment in 2016, but maybe I miss something?

src/odemis/gui/cont/tabs.py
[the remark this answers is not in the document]
Kleijwegt: @pieleric I think I have fixed this issue but I'm not sure if I have fixed it as you intended/correctly, can you check that and close this conversion if all is ok?

src/odemis/gui/model/__init__.py
```diff
@@ -194,6 +195,8 @@ class MainGUIData(object):
         "ebeam-shift": "beamshift",
         "diagnostic-ccd": "ccd",
         "det-rotator": "det_rotator",
+        "se-detector-ion": "ion_sed",
+        "stage-global": "stage_global",
```
pieleric, line 199: You need to add these 2 variables to the __init__ too. (eg, line 273). This way, if the components are not present, they will be set to `None` (otherwise, the attributes are not present).

src/odemis/gui/model/__init__.py
```diff
@@ -1042,6 +1045,12 @@ def __init__(self, main):
         # For dichotomic mode
         self.dicho_seq = model.ListVA()  # list of 4 enumerated for each corner
 
+class EnzelAlignGUIData(ActuatorGUIData):
+    def __init__(self, main):
+        ActuatorGUIData.__init__(self, main)
+        self.viewLayout.value = VIEW_LAYOUT_VERTICAL
+        self.step_size = model.FloatVA(1e-6)
+        self.align_mode = StringEnumerated("Z alignment", choices=set(("Z alignment", "SEM alignment", "FLM alignment")))
```
pieleric, line 1053: Be aware it's a little tricky to write the strings like this because it's easy to make a typo, and Python will not notice (until really late). So that's why in such case, I usually go for constants (eg `Z_ALIGN = "Z alignment"`). This way, a typo is detected early. For the SPARC2 alignment, that's not what I did though, I just kept the strings all lower case and without space, to reduce the chances of typo.

src/odemis/gui/model/__init__.py
```diff
@@ -1042,6 +1045,12 @@ def __init__(self, main):
         # For dichotomic mode
         self.dicho_seq = model.ListVA()  # list of 4 enumerated for each corner
 
+class EnzelAlignGUIData(ActuatorGUIData):
+    def __init__(self, main):
+        ActuatorGUIData.__init__(self, main)
+        self.viewLayout.value = VIEW_LAYOUT_VERTICAL
+        self.step_size = model.FloatVA(1e-6)
```
pieleric, line 1052: Please make it a FloatContinuous, with a range and a unit. If you don't know the range, pick something "safe" like (1e-9 -> 1e-3).

src/odemis/gui/model/__init__.py
```diff
@@ -561,7 +564,7 @@ def __init__(self, main):
         # See class docstring for more info.
         self.focussedView = VigilantAttribute(None)
 
-        layouts = {VIEW_LAYOUT_ONE, VIEW_LAYOUT_22, VIEW_LAYOUT_FULLSCREEN}
+        layouts = {VIEW_LAYOUT_ONE, VIEW_LAYOUT_22, VIEW_LAYOUT_VERTICAL, VIEW_LAYOUT_FULLSCREEN}
```
pieleric, line 567: Let's not change the possible layouts for all the tabs automatically. You can redefine the viewLayout in the EnzelAlignGUIData.

src/odemis/gui/model/__init__.py
```diff
@@ -1042,6 +1045,12 @@ def __init__(self, main):
         # For dichotomic mode
         self.dicho_seq = model.ListVA()  # list of 4 enumerated for each corner
 
+class EnzelAlignGUIData(ActuatorGUIData):
+    def __init__(self, main):
+        ActuatorGUIData.__init__(self, main)
+        self.viewLayout.value = VIEW_LAYOUT_VERTICAL
```
pieleric, line 1051: Here, you can redefine it, with `self.viewLayout = model.IntEnumerated(VIEW_LAYOUT_22, choices={VIEW_LAYOUT_ONE, VIEW_LAYOUT_VERTICAL})` 
(Not sure if VIEW_LAYOUT_ONE is even possible, in which case, drop it).

src/odemis/gui/cont/tabs.py
```diff
@@ -4455,7 +4469,496 @@ def _on_align_move_done(self, f=None):
...
+    def _on_stage_pos(self, pos):
+        """
+        Called when the stage is moved, enable the tab if position is imaging mode, disable otherwise
+
+        :param pos: (dict str->float or None) updated position of the stage
+        """
+        guiutil.enable_tab_on_stage_position(self.button, self._stage, pos, target=[ALIGNMENT, IMAGING, THREE_BEAMS])
+        # Display a tooltip whenever the tab cannot be clicked.
+        if getCurrentPositionLabel(self._stage.position.value, self._stage) not in (ALIGNMENT, IMAGING, THREE_BEAMS):
+            self.button.SetToolTip("Alignment can only be performed in the imaging mode")
```
pieleric, line 4955: Should it be "imaging" -> "Three beams"?

src/odemis/gui/cont/tabs.py
```diff
@@ -4455,7 +4469,496 @@ def _on_align_move_done(self, f=None):
...
+        else:
+            self.button.SetToolTip(None)
+
+    @classmethod
+    def get_display_priority(cls, main_data):
+        if main_data.role in ("enzel"):
```
pieleric, line 4961: `("enzel")` -> ("enzel",)
(Otherwise it's not a tuple).

src/odemis/gui/cont/tabs.py
```diff
@@ -4455,7 +4469,496 @@ def _on_align_move_done(self, f=None):
...
+    def _z_alignment_controls(self, axis, proportion, evt):
+        """
+        Links the Z alignment mode control buttons to the stage and updates the correct metadata.
+
+        :param axis (str): axis which is controlled by the button.
+        :param proportion (int): direction of the movement represented by the button (-1/1)
+        :param evt (GenButtonEvent): Clicked event
+        """
+        # Only proceed if there is no currently running target_position
+        if hasattr(self, "_z_move_future") and self._z_move_future._state == RUNNING:
```
pieleric, line 4737: To check the state of a future, use `not self._z_move_future.done()` .
You sould also set _z_move_future at init as None, so that you can more nicely write `if self.z_move_future ...`

src/odemis/gui/cont/tabs.py
```diff
@@ -4455,7 +4469,496 @@ def _on_align_move_done(self, f=None):
...
+        for btn in self._align_btn_functions:
+            btn.Bind(wx.EVT_BUTTON, self._click_align_mode_button)
+
+        # Start with the first mode as default
+        self._align_btn_functions[panel.btn_align_z]()
+        panel.btn_align_z.SetToggle(True)
+
+        # TODO K.K. Before merging uncomment line about stage condition check.
+        # Set the stage condition on which the tab is enabled.
+        # self._stage.position.subscribe(self._on_stage_pos, init=True)
```
pieleric, line 4647: Please uncomment.

src/odemis/gui/cont/tabs.py
```diff
@@ -4455,7 +4469,496 @@ def _on_align_move_done(self, f=None):
...
+        """
+        Sets the SEM alignment mode with the correct streams, stream controllers, instructions and alignment controls.
+        """
+        self.stream_bar_controller.pauseAllStreams()
+        self._set_top_and_bottom_stream_and_settings(self._SEM_view_and_control,
+                                                     self._FIB_view_and_control)
+
+        # Adjust the overlay to a normal crosshair in the FIB viewport
+        for viewport in self.panel.pnl_two_streams_grid.viewports:
+            if viewport.view.stream_classes is acqstream.FIBStream:
```
pieleric, line 4769: That feels a little indirect.
You could write `fib_viewport = self._FIB_view_and_control["viewport"]`, and you'd be done.

pieleric (changes requested):

src/odemis/gui/cont/tabs.py
```diff
@@ -4455,7 +4469,497 @@ def _on_align_move_done(self, f=None):
...
+            except CancelledError:
+                logging.warning("Aligner move cancelled")
+                return
+            except Exception as ex:
+                logging.error("Aligner move failed: %s", ex)
+                return
+
+        future_stage_move.add_done_callback(move_aligner)
+
+    def _click_align_mode_button(self, clicked_align_mode):
```
pieleric, line 4685: The argument is an wx.Event. By convention, it's always named `evt`.

src/odemis/gui/cont/tabs.py
```diff
@@ -4455,7 +4469,497 @@ def _on_align_move_done(self, f=None):
...
+
+
+        panel.btn_custom_alignment.Bind(wx.EVT_BUTTON, self._on_click_custom_alignment)
+
+        # Bind the align mode buttons
+        for btn in self._align_btn_functions:
+            btn.Bind(wx.EVT_BUTTON, self._click_align_mode_button)
+
+        # Start with the first mode as default
+        self._align_btn_functions[panel.btn_align_z]()
```
pieleric, line 4644: I'd suggest to handle this a little bit differently:
* When one of the alignment button is pressed, set .align_mode to the corresponding mode (you can do this in just a couple of lines with a dict).
* Have a subscriber to the .align_mode (you can do that here, and pass `init=True`). Then, this function does 2 things: update the toggle buttons to the right state ( as currently done in `_click_align_mode_button()`), and set the tab to the proper state, essentially by calling the corresponding `_set_XXX_alignment_mode()` function. You can dispatch either with a if/elif/else block, or a dict. As there are only 3 functions, I feel that an if/elif block is fine.

pieleric (changes requested):

src/odemis/gui/cont/tabs.py
```diff
@@ -4766,12 +4767,11 @@ def _set_sem_alignment_mode(self):
...
-            if viewport.view.stream_classes is acqstream.FIBStream:
-                for FIB_line_overlay in viewport.canvas.view_overlays:
-                    if isinstance(FIB_line_overlay, CenteredLineOverlay):
-                        FIB_line_overlay.shape = CROSSHAIR
-                        break
+        fib_viewport = self._FIB_view_and_control["viewport"]
+        for FIB_line_overlay in fib_viewport.canvas.view_overlays:
+            if isinstance(FIB_line_overlay, CenteredLineOverlay):
+                FIB_line_overlay.shape = CROSSHAIR
+                break
```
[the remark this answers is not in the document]
Kleijwegt: It is actually, whenever z mode alignment is set a horizontal line is used and otherwise a crosshair.
```
if isinstance(FIB_line_overlay, CenteredLineOverlay):
    FIB_line_overlay.shape = HORIZONTAL_LINE
    break
```
This part should also be updated such as https://github.com/delmic/odemis/pull/1963#discussion_r845027466 actually.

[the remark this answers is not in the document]
pieleric: Sorry, I missed that part. It's all fine then!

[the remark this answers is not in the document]
Kleijwegt: it is actually changed. So I leave it as is.

src/odemis/gui/cont/tabs.py
```diff
@@ -4655,6 +4653,45 @@ def terminate(self):
...
+        :param align_mode (string): Current mode (Z_ALIGN, SEM_ALIGN, FLM_ALIGN)
+        """
+        # Un/toggle all the buttons.
+        for mode, btn in self._align_modes.items():
+            if mode is not align_mode:
+                btn.SetToggle(False)
+            elif mode is align_mode:
+                btn.SetToggle(True)
+            else:
+                logging.error("Incorrect alignment mode is set.")
```
pieleric, line 4686: The two tests before check every possibilities: either `mode` is `align_mode` or it's not. So this else will never happen. What you could check is that align_mode is in self._align_modes (but you'd do it outside of the loop). Alternatively, you could put an `else` in the `if` block below.

Anyway, for this loop, you can keep it simple with a code like:
```python
        for mode, btn in self._align_modes.items():
                btn.SetToggle(mode == align_mode)
```

pieleric (changes requested):

src/odemis/gui/cont/tabs.py
Kleijwegt: Neat

pieleric (approved):

PR #1968, opened 2021-12-17
gui test comp_canvas_cairo: fix VA value out of range

pieleric:
test_reshape() wasn't testing anything, so remove it.
test_nanana() tests the mpp changes, which is sort of useful. Fix the VA
initialisation.

### DIFF
--- a/src/odemis/gui/test/comp_canvas_cairo_test.py
+++ b/src/odemis/gui/test/comp_canvas_cairo_test.py
@@ -341,7 +341,7 @@ class TestCanvas(test.GuiTestCase):
         logging.getLogger().setLevel(logging.ERROR)
 
     # @profile
-    def test(self):
+    def test_set_images(self):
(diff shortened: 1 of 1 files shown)


src/odemis/gui/test/comp_canvas_cairo_test.py
```diff
@@ -406,7 +406,7 @@ def test_nanana(self):
 
         # old_canvas = DraggableCanvas(self.panel)
         tab = self.create_simple_tab_model()
-        mpp = FloatContinuous(10e-6, range=(1e-3, 1), unit="m/px")
+        mpp = FloatContinuous(10e-6, range=(1e-6, 1), unit="m/px")
```
Kleijwegt, line 409: Not related to the change but maybe we should at a comment on why this test is called nanana at the start of this test.
Even after looking at the code it isn't 100% clear to me (yet)

Kleijwegt (approved):

PR #1969, opened 2021-12-17
acq stitching tiledacq: Add focusing_method parameter

pieleric:


src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -102,6 +114,9 @@ def __init__(self, streams, stage, area, overlap, settings_obs=None, log_path=No
 
         # To use in re-focusing acquired images in case they fell out of focus
         # TODO: make adjust focus optional
```
lazem, line 116: I guess that TODO can be removed now.

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -586,40 +620,73 @@ def _acquireTiles(self):
...
+            logging.debug("Current focus level: %s (good = %s)", current_focus_level, self._good_focus_level)
+
+            # Run autofocus if current focus got worse than permitted deviation,
+            # or it was very bad originally.
+            if not (self._good_focus_level == 0 or
+                    (self._good_focus_level - current_focus_level) / self._good_focus_level > FOCUS_FIDELITY
+                   ):
+                return das
+        elif self._focusing_method == FocusingMethod.ALWAYS:
+            pass
```
lazem, line 661: Didn't get the reason about this, but it possibly can be removed without effect.

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -586,40 +620,73 @@ def _acquireTiles(self):
...
+                self._good_focus_level = current_focus_level
+                return das
+
+            logging.debug("Current focus level: %s (good = %s)", current_focus_level, self._good_focus_level)
+
+            # Run autofocus if current focus got worse than permitted deviation,
+            # or it was very bad originally.
+            if not (self._good_focus_level == 0 or
+                    (self._good_focus_level - current_focus_level) / self._good_focus_level > FOCUS_FIDELITY
+                   ):
```
lazem, line 658: Maybe this can be easier as: 
```
if ( self._good_focus_level != 0
and (self._good_focus_level - current_focus_level)  self._good_focus_level
                <= FOCUS_FIDELITY):
```

lazem (changes requested):

install/linux/usr/share/odemis/sim/enzel-sim.odm.yaml
```diff
@@ -253,9 +254,6 @@ ENZEL-Sim: {
   metadata: {
     FAV_POS_ACTIVE: { 'z': 0.002 }
   },
-  persistent: {
-    metadata: [ FAV_POS_ACTIVE ],  # Store "good focus" position between run
-  },
```
pieleric, line 258: @Kleijwegt , this part is inconvenient for the test cases, because the image is not necessarily in focus (and it's hard to know the right focus position, if a bad one is stored). I looked the GUI code, and it doesn't seemed to be used anyway (only FAV_POS_ACTIVE on stage and align are used). Could you confirm that's the case?

PR #1972, opened 2021-12-17
driver pvcam test: skip every tests if TEST_NOHW

pieleric:
The static test doesn't need the real hardware, but it at least needs
libpvcam, which we don't have on normal computers (because only 32 bit
version exists)

### DIFF
--- a/src/odemis/driver/test/pvcam_test.py
+++ b/src/odemis/driver/test/pvcam_test.py
@@ -44,6 +44,15 @@ class StaticTestPVCam(VirtualStaticTestCam, unittest.TestCase):
     camera_type = CLASS
     camera_kwargs = CONFIG_PVCAM
 
+    @classmethod
+    def setUpClass(cls):
+        if TEST_NOHW:
+            # Technically, we don't need the hardware, but at least the lipvcam is needed,
+            # and it's not available on 64 bits.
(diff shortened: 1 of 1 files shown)


src/odemis/driver/test/pvcam_test.py
```diff
@@ -55,18 +63,7 @@ class TestPVCam(VirtualTestCam, unittest.TestCase):
...
-    @classmethod
-    def tearDownClass(cls):
-        if not TEST_NOHW:
-            super(TestPVCam, cls).tearDownClass()
-
-    def setUp(self):
-        if TEST_NOHW:
-            self.skipTest("No simulator available")
-        super(TestPVCam, self).setUp()
+        raise unittest.SkipTest("No simulator available")
```
Kleijwegt, line 66: will we always skip this test? Don't we need to define an if statement? Even if it is checking a constant defined at the top of the file. If you want to run the tests on a 32 bit system you just need to adjust one variable instead of commening two setUpClass methods.

pieleric: Indeed, my bad. I was so enthusiastic of removing the code, that I forgot we should still run the code if the hardware is available!

=> I've updated with a `if TEST_NOHW`.

Kleijwegt (approved):

PR #1975, opened 2021-12-22
Piezomotor when raising a hardware error add name of axis on which th…

tepals:
…e error was detected.


Sabrina85 (changes requested):
The f-string formatting was only introduced in python 3.6. We still need to support Python 3.5 or don't we? In general, I think it is nicer to keep the formatting the same within the same file at least. 
The changes themselves look fine.

pieleric:
For the development version of Odemis, it's OK to only support Python 3.6+. (we already have a lot of f-strings). We only need to be careful for Python 2.7+ support for code being backported to the stable version.

PR #1978, opened 2021-12-28
ENZEL: Replace chamber tab position buttons

lazem:
In order to avoid confusion between 3 beams & original alignment (factory reset) position buttons, hide alignment button in advanced panel, that way it's visible if the stage is currently on that position (the button would be green). 
Also put the sem imaging button on position panel.


Kleijwegt (commented):
If you add the lowered safer  position of the 3 and 5 beam via the front-end I approve the PR.

src/odemis/acq/move.py
```diff
@@ -47,13 +48,16 @@
...
+    GRID_2: "GRID 2",
+    THREE_BEAMS: "THREE BEAMS"
 }
 
 ATOL_LINEAR_POS = 100e-6  # m
 ATOL_ROTATION_POS = 1e-3  # rad (~0.5°)
 RTOL_PROGRESS = 0.3
-SCALING_FACTOR = 0.03 # m (based on fine tuning)
+SCALING_FACTOR = 0.03  # m (based on fine tuning)
+SAFETY_MARGIN_5DOF = 100e-6  # m
```
Kleijwegt, line 59: I updated this number which added me as a author to all the commits. All the credits go to @lazem however.

Kleijwegt (approved):

src/odemis/acq/move.py
lazem: I made it this way because ATOL_LINEAR_POS is also 100e-6, so there won't be a difference between 3 beams and active in this case.

PR #1980, opened 2022-01-03
Spot tests reduced accuracy in test asserts for grid testing, because…

tepals:
… this caused the tests to fail while the unit tests do not need t be this accurate


Kleijwegt (approved):
This test file still needs an update, the test TestSpotAlignment fails. But most likely this is due to an error elsewhere.

tepals:
> This test file still needs an update, the test TestSpotAlignment fails. But most likely this is due to an error elsewhere.

It's not failing for me locally. In your log file I can see that it failed because of the Backend failed to start, so that's most likely due to a problem somewhere else.

PR #1981, opened 2022-01-03
Changed timeshrt to timeshrt_hypen in test_create_filename, because t…

tepals:
…imeshrt is not supported in odemis.util.filename.create_filename and timeshrt_hypen is but was not tested


tepals:
@pieleric I've now updated the test case to test for `timeshrt_hyphen`. Is that correct, or do we just not want to support `timeshrt_hyphen` anymore and should I remove it as an option from `create_filename`?

Kleijwegt (approved):

pieleric:
I think that's the right way to doit. We removed "timeshrt", because it was too easily confused with just a number. `timeshrt_hyphen` can stay, so let's use it.

PR #1983, opened 2022-01-05
fastem acquisition: configure scanner test cases

Sabrina85:
This file seems to have gotten lost for pr #1884 (maybe when squashing commits).

### DIFF
--- a/src/odemis/acq/test/fastem_conf_test.py
+++ b/src/odemis/acq/test/fastem_conf_test.py
@@ -1,0 +1,113 @@
+# -*- coding: utf-8 -*-
+"""
+Created on 03 December 2021
+
+@author: Sabrina Rossberger, Éric Piel
+
+Copyright © 2021 Sabrina Rossberger, Delmic
+
+This file is part of Odemis.
+
+Odemis is free software: you can redistribute it and/or modify it under the terms
+of the GNU General Public License version 2 as published by the Free Software
+Foundation.
+
+Odemis is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+PURPOSE. See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with
+Odemis. If not, see http://www.gnu.org/licenses/.
+
+"""
+from __future__ import division
+
+import logging
+import math
+import os
+import unittest
+
+import odemis
+from odemis import model
+from odemis.acq import fastem_conf
+from odemis.acq.fastem_conf import OVERVIEW_MODE, LIVESTREAM_MODE, MEGAFIELD_MODE
+from odemis.util import img
+
+logging.getLogger().setLevel(logging.DEBUG)
+logging.basicConfig(format="%(asctime)s  %(levelname)-7s %(module)s:%(lineno)d %(message)s")
+
+# * TEST_NOHW = 1: not connected to anything => skip most of the tests
+# * TEST_NOHW = 0: connected to the real hardware or simulator
+# technolution_asm_simulator/simulator2/run_the_simulator.py
+TEST_NOHW = os.environ.get("TEST_NOHW", "0")  # Default is HW/simulator testing
+
+CONFIG_PATH = os.path.dirname(odemis.__file__) + "/../../install/linux/usr/share/odemis/"
+FASTEM_CONFIG = CONFIG_PATH + "sim/fastem-sim.odm.yaml"
+
+
+class TestFASTEMConfig(unittest.TestCase):
+    """Test the FASTEM scanner configuration."""
+
+    @classmethod
+    def setUpClass(cls):
+        if TEST_NOHW is True:
+            raise unittest.SkipTest("No simulator running or HW present. Skip fastem acquisition tests.")
+
+        cls.scanner = model.getComponent(role="e-beam")
+        cls.sed = model.getComponent(role="se-detector")
+        cls.multibeam = model.getComponent(role="multibeam")
+        cls.mppc = model.getComponent(role="mppc")
+
+    def test_configure_scanner(self):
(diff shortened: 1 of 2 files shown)


src/odemis/acq/test/fastem_conf_test.py
```diff
@@ -0,0 +1,113 @@
...
+logging.getLogger().setLevel(logging.DEBUG)
+logging.basicConfig(format="%(asctime)s  %(levelname)-7s %(module)s:%(lineno)d %(message)s")
+
+# * TEST_NOHW = 1: not connected to anything => skip most of the tests
+# * TEST_NOHW = 0: connected to the real hardware or simulator
+# technolution_asm_simulator/simulator2/run_the_simulator.py
+TEST_NOHW = os.environ.get("TEST_NOHW", "0")  # Default is HW/simulator testing
+
+CONFIG_PATH = os.path.dirname(odemis.__file__) + "/../../install/linux/usr/share/odemis/"
+FASTEM_CONFIG = CONFIG_PATH + "sim/fastem-sim.odm.yaml"
```
tepals, line 45: why don't you use these config paths?

src/odemis/acq/test/fastem_conf_test.py
```diff
@@ -0,0 +1,113 @@
...
+    def setUpClass(cls):
+        if TEST_NOHW is True:
+            raise unittest.SkipTest("No simulator running or HW present. Skip fastem acquisition tests.")
+
+        cls.scanner = model.getComponent(role="e-beam")
+        cls.sed = model.getComponent(role="se-detector")
+        cls.multibeam = model.getComponent(role="multibeam")
+        cls.mppc = model.getComponent(role="mppc")
+
+    def test_configure_scanner(self):
```
tepals, line 61: It might be nice to split this up into 3 test cases. Now if the overview mode test fails the rest doesn't get tested. While if you have a test case for each mode, everything is always tested.

src/odemis/acq/test/fastem_conf_test.py
Sabrina85: That is currently the only one we have, but it does not work with these tests as it does not have an asm component. There is a Jira task for this problem as we have more tests that depend on an asm component (e.g. technolution driver tests): SDEV-988. So far the weekly testing was run with TEST_NOHW = True, so all these tests here are skipped for now.

tepals: Can you then add a TODO here that these configuration files should be used once there is an asm component available in them?

src/odemis/acq/test/fastem_conf_test.py
```diff
@@ -0,0 +1,113 @@
...
+        self.assertTrue(self.scanner.multiBeamMode.value)
+        self.assertTrue(self.scanner.external.value)
+        self.assertFalse(self.scanner.blanker.value)
+        self.assertTrue(self.scanner.immersion.value)
+        self.assertLess(self.scanner.horizontalFoV.value, 0.1e-3)  # should be smaller for megafield than for overview
+        self.assertEqual(self.scanner.rotation.value, math.radians(10))
+        
+        # check that rotation correction is 0 for megafield imaging using the ASM/SAM
+        self.assertEqual(self.multibeam.getMetadata().get(model.MD_ROTATION_COR, 0), 0)  # if set to 0 or not existing
+        self.assertEqual(self.multibeam.getMetadata()[model.MD_ROTATION], math.radians(10))  # same as on rotation VA
```
tepals, line 109: I don't think these checks belong in `test_configure_scanner`, since you only configure the scanner not the multibeam.  Plus, if you remove them you can get rid of the asm component dependency.

Sabrina85: When you configure the scanner it affects the multibeam (set via affects in yaml file). As the affects was missing, we found out, that this MD was not updated, but it is essential that it gets updated. That's why this check is here.

pieleric: I agree that it's useful to check. However, Thera has a very interesting remark: that's the only place where the multibeam is needed. So when TEST_NOHW=1, instead of entirely giving up running the test cases, I'd suggest that we start the `FASTEM_CONFIG` (ie, the simulator without any external dependencies), and run everything *except* these 2 lines.

pieleric (changes requested):

src/odemis/acq/test/fastem_conf_test.py
Sabrina85: The configure scanner method has some specific multibeam related VAs e.g. it sets the `scanner.multibeamMode.value` VA. This VA is not present when starting the current backend. So I changed the backend to have a `mb-scanner` child, which makes sure that the `MultiBeamScanner` is actually used. However, the `simsem` seems to need to have a `scanner ` child - otherwise the backend does not start. Any ideas on how to proceed with this?

pieleric: Arghhh.... I had forgotten about this. So the fastem-sim cannot be used at all to acquire any image in the GUI? That's a big limitation. For this PR, I think then you can leave it as is, but let's try to make fastem-sim more useful.

I think that we have two options:
* Extend SimSEM to simulate better the xt_client with multibeam (it's essentially a matter of adding a few VAs). Maybe via some flag (eg a parameter "multibeam" on the scanner), or the use of "mb-scanner" instead of "scanner", or a generic approach like `simulated.GenericComponent(..., vas...)`. It still doesn't bring any component related to the ASM.
* Bite the bullet and say that to run the fastem-sim you need to have the xtlib-adpater and ASM simulator running on your computer.

pieleric (approved):

tepals (approved):

Sabrina85:
@pieleric @tepals Are you fine with merging it? I had added the yaml file after you already had approved.

tepals:
I apparently forgot to submit my review

tepals (approved):

pieleric (approved):

PR #1984, opened 2022-01-07
Fastem set immersion mode to True at the end of the overview image ac…

tepals:
…quisition to be able to do SEM focusing and use the system as normal.

Force beam blanking at the end of megafield acquisition and overview acquisition, because the auto function is not working properly.

### DIFF
--- a/src/odemis/acq/fastem.py
+++ b/src/odemis/acq/fastem.py
@@ -318,6 +318,9 @@ class AcquisitionTask(object):
             # Remove references to the megafield once the acquisition is finished/cancelled.
             self._fields_remaining.clear()
 
+            # Blank the beam after the acquisition is done. 
+            self._scanner.blanker = True
+
             # Finish the megafield also if an exception was raised, in order to enable a new acquisition.
             logging.debug("Finish megafield acquisition.")
             dataflow.unsubscribe(self.image_received)
@@ -587,4 +590,11 @@ def _run_overview_acquisition(f, stream, stage, area, live_stream):
 
     if len(das) != 1:
         logging.warning("Expected 1 DataArray, but got %d: %r", len(das), das)
+
+    # Switch immersion mode back on, so we can focus the SEM from the TFS GUI.
+    stream.emitter.immersion.value = True
+
+    # FIXME auto blanking not working properly, so force beam blanking after image acquisition for now.
+    stream.emitter.blanker = True
+
     return das[0]


src/odemis/acq/fastem.py
```diff
@@ -318,6 +318,9 @@ def run(self):
             # Remove references to the megafield once the acquisition is finished/cancelled.
             self._fields_remaining.clear()
 
+            # FIXME auto blanking not working properly, so force beam blanking after image acquisition for now.
```
Sabrina85, line 321: Autoblanking will never work in external mode as far as I know.

src/odemis/acq/fastem.py
```diff
@@ -587,4 +590,11 @@ def _pass_future_progress(sub_f, start, end):
 
     if len(das) != 1:
         logging.warning("Expected 1 DataArray, but got %d: %r", len(das), das)
+
+    # Switch immersion mode back on, so we can do use the system normally.
+    stream.emitter.immersion.value = True
+
+    # FIXME auto blanking not working properly, so force beam blanking after image acquisition for now.
```
Sabrina85, line 597: Why did you add this line? I thought the autoblanking works now.

Sabrina85 (changes requested):

src/odemis/acq/fastem.py
tepals: Wilco just let me know that for some reason it doesn't always work

Sabrina85: ok

src/odemis/acq/fastem.py
```diff
@@ -587,4 +590,11 @@ def _pass_future_progress(sub_f, start, end):
 
     if len(das) != 1:
         logging.warning("Expected 1 DataArray, but got %d: %r", len(das), das)
+
+    # Switch immersion mode back on, so we can do use the system normally.
+    stream.emitter.immersion.value = True
```
pieleric, line 595: This is a little odd/unclear to me. Typically, you want to configure the hardware with the right settings *before* you do an acquisition. After an acquisition, you don't need to change the setting. It's up to the *next* acquisition to do this work. (Here, the chances that the next acquisition is an another overview is actually quite high, so it's just extra work/extra chances to mess up with some other settings).

So my question is: what is the "normal use of the system"? Is that because for now the user still relies on the TFS GUI to do some tasks, and it's inconvenient to switch immersion mode in their software? Or is there some acquisition in Odemis which do no set the immersion mode properly?

src/odemis/acq/fastem.py
pieleric: Indeed, auto-blanking is not going to work with external mode. Actually, Kornee has some new wrapper component almost ready, for the ENZEL, to add a "None" option to the blanker VA in such case, but that relies on the detector behaving "normally" (ie, not the extended API of the ASM DataFlow). So don't count on it anytime soon. It's simpler to just handle it explicitly here.

pieleric (changes requested):

src/odemis/acq/fastem.py
tepals: Yes, with the current state of the system after acquiring an overview image the change of acquiring another one directly is quite low. You will need to do SEM focusing through the TFS GUI first and indeed it's inconvenient to switch immersion mode in their software. I will update the comment to explain this a bit better.

src/odemis/acq/fastem.py
tepals: Why will autoblanking never work in external mode? I can get a value for whether the beam is blanked or not, and I can blank the beam in external mode.

pieleric: Because autoblanking means "unblank" when the DataFlow linked to this emitter has subscribers. However, as the .blanker VA is on the xt_client driver and the DataFlow is on the ASM, there is currently no way to connect one to the other.
The approach taken by Kornee to handle this is to have a wrapper for the DataFlow, which takes care of switching the .blanker VA. With some extension, it could work with the special FASTEM/ASM DataFlow API, but as all this is so ad-hoc, I don't see much the point compared to just explicitly controlling the blanker in the acquisition code, as you do.

PR #1985, opened 2022-01-10
fastem acqu manager: change the direction of stage move to match the requirements

Sabrina85:
First image is top/left corner of megafield. As soon as stage-scan is used, movements need to be opposite order to fulfill this requirement.

### DIFF
--- a/src/odemis/acq/fastem.py
+++ b/src/odemis/acq/fastem.py
@@ -396,11 +396,15 @@ class AcquisitionTask(object):
         rel_move_hor = self.field_idx[0] * px_size[0] * field_res[0]  # in meter
         rel_move_vert = self.field_idx[1] * px_size[1] * field_res[1]  # in meter
 
-        # Move in the negative x direction, because the second field should be right of the first.
-        # Move in positive y direction, because the second field should be bottom of the first.
-        # TODO verify that this results in the images being taken in the correct order (left to right, top to bottom).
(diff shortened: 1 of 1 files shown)


tepals (approved):

src/odemis/acq/fastem.py
```diff
@@ -398,9 +398,11 @@ def get_abs_stage_movement(self):
 
         # Move in the negative x direction, because the second field should be right of the first.
         # Move in positive y direction, because the second field should be bottom of the first.
```
pieleric, line 400: These 2 lines should be updated, to match the line below (and maybe remind which stage you expect to move... if it's not stage-scan).

pieleric (changes requested):

src/odemis/acq/fastem.py
pieleric: Ok. I'd even clarify like:
```
With role="stage" move pos in x and neg in y.
With role="stage-scan" move neg in x and pos in y (because this stage follows a referential where X and Y move in the global coordinates + a small extra angle)
```

pieleric (changes requested):

src/odemis/acq/fastem.py
Sabrina85: I updated the comment to clarify. I left out the part in the brackets. I think the "stage-bare" moves in the global referential and the "stage-scan" in the multiprobe referential. That's how we defined it in the yaml file.

PR #1986, opened 2022-01-10
Actuator: Fix DualChannelPositionSensor referencing.

tepals:
The stage requires a set of axis to reference, while the axis were passed as dict keys. This commit ensures the axis are passed as a set.

### DIFF
--- a/src/odemis/driver/actuator.py
+++ b/src/odemis/driver/actuator.py
@@ -2998,12 +2998,12 @@ class DualChannelPositionSensor(model.HwComponent):
     def _doReference(self, axes, omit_referenced):
         # Reference all stage axes (also those which are not in "axes", e.g. z axis)
         if self.stage:
-            stage_axes = self.stage.axes.keys()
+            stage_axes = set(self.stage.axes.keys())  # turn dict keys into a set, because reference expects a set
             if omit_referenced:
                 # Skip axes which are already referenced or cannot be referenced
                 stage_axes = {ax for ax in stage_axes if not self.stage.referenced.value.get(ax, True)}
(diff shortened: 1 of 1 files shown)


src/odemis/driver/actuator.py
```diff
@@ -3003,7 +3003,7 @@ def _doReference(self, axes, omit_referenced):
                 # Skip axes which are already referenced or cannot be referenced
                 stage_axes = {ax for ax in stage_axes if not self.stage.referenced.value.get(ax, True)}
             logging.debug("Referencing stage axes %s.", stage_axes)
-            f = self.stage.reference(stage_axes)
+            f = self.stage.reference(set(stage_axes))
```
Sabrina85, line 3006: Can you add a comment why you need to do this please?

pieleric: reference() expects a set. So it sort of makes sense... but I'd suggest that a clearer way is to directly convert stage_axes when it's read: line 3001 becomes `stage_axes = set(self.stage.axes.keys())`.

pieleric (changes requested):

src/odemis/driver/actuator.py
tepals: Does that then also work with line 3004, or should that then also be updated to 

`stage_axes = set(ax for ax in stage_axes if not self.stage.referenced.value.get(ax, True))`

pieleric: Line 3004 returns a set already.
In Python, it's a little confusing, but `{1, 2}` is a set, and `{1: "a", 2: "b"}` is a dict.
So `{a for a in....}` is a set, and `{a: b for a, b in ....}` is a dict.

PR #1990, opened 2022-01-12
Fastem revert changes in overview imaging

tepals:
The changes in commit 8a07a870b5ef88e33f22112dbaecf4a3fa89d356 caused problems with overview imaging. The commit removed copying the fastem stream object into a sem stream object.
Because a fastem stream object was used the overview image settings were overwritten with the live view image settings, this turned on the immersion mode and only a quarter of the overview image was acquired.

### DIFF
--- a/src/odemis/acq/fastem.py
+++ b/src/odemis/acq/fastem.py
@@ -534,9 +534,13 @@ def acquireTiledArea(stream, stage, area, live_stream=None):
     if live_stream:
         raise NotImplementedError("live_stream not supported")
 
-    est_dur = estimateTiledAcquisitionTime(stream, stage, area)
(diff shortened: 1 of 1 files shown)


Sabrina85 (approved):
Do we need to have a TODO there? Which steps/actions should be taken in future to make this code unnecessary?

pieleric:
> Do we need to have a TODO there? Which steps/actions should be taken in future to make this code unnecessary?

I don't think there is anything else to do. It's fine as it is. We had introduced that stream duplication for other reasons. The "new" reason is a good one to keep having the duplication. It's not really costly, just a tiny bit more complex.

pieleric (approved):

PR #1995, opened 2022-01-14
fastem: Unblank beam before live view and overview image acquisition

tepals:
We pause the stream in between each acquired image, this already blanks the beam. Forcing blanking and subsequent unblanking of the beam caused it to blanked during acquisition sometimes.
In the fastem configuration set the blanking to False for live view and overview imaging, so it's unblanked during acquisition.

### DIFF
--- a/src/odemis/acq/fastem_conf.py
+++ b/src/odemis/acq/fastem_conf.py
@@ -40,7 +40,10 @@ SCANNER_CONFIG = {
     OVERVIEW_MODE: {
         "multiBeamMode": False,
         "external": False,  # fullframe mode; controlled by SEM itself
-        "blanker": None,  # automatic: unblanked when acquiring; blanked when not acquiring
+        # manual: unblank when acquiring and the beam is blanked after the acquisition. Note that autoblanking does not
+        # work reliably for the XTTKDetector, therefore (contrary to Odemis convention) we need to unblank
+        # the beam here.
+        "blanker": False,
         "immersion": False,  # disable to get a larger field of view
         "horizontalFoV": 1.5e-3,
         "resolution": (1024, 884),  # px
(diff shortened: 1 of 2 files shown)


tepals:
@Sabrina85 because of this PR your new test cases in https://github.com/delmic/odemis/pull/1983 will need to be updated. We can merge your code first and then I'll do a rebase and update the testcases.

Sabrina85:
> @Sabrina85 because of this PR your new test cases in #1983 will need to be updated. We can merge your code first and then I'll do a rebase and update the testcases.

merged!

pieleric (commented):
I'm confused. You're setting the blanker to False, because when it's set to `False` it's actually doing auto-blanking (ie, blanked when not acquiring). I'd say that if that's the case, the xt_client.blanker.value should be `None`, to match the standard Odemis definition.
Maybe, blanker.value is `True`, you'd need to call  `blank_beam()` explicitly just before starting the acquisition, to force the blanking, and similarly, if blanker.value is `False`, you'd need to call `unblank_beam()` after acquisition.

That said, maybe it's a little optimistic that we understand _reliably_ the behaviour of the XTTK. Then just "give up" trying to control it explicitly, and set the blanker to False as you do here. In such, comment very clearly the fact that the blanker doesn't behave as the convention in Odemis.

tepals:
@pieleric it's confusing, if you set the blanker to False it's doing autoblanking when you're in full frame mode, however if you then switch to external mode the beam is not blanked. 
I would say it's optimistic that we reliably understand the XTTK behavior, I'll add comments that the blanker does not behave as the convention in Odemis.

tepals:
@pieleric @Sabrina85 updated

pieleric (approved):
When merging, please make sure to update the title of this commit to something more appropriate.

Sabrina85 (approved):

PR #1996, opened 2022-01-14
Added pointSpreadFunctionSize VA to model.DigitalCamera

effting:
Provides a spotSize VA on a model.DigitalCamera instance. This is equivalent to the standard deviation of the Gaussian approximation of a fluorescence microscope point spread function, measured in no. of pixels of the camera.

The value can be used as a characteristic size parameter when filtering an image prior to spot detection.

Will be used to clean-up the code in `fastem-calibrations:/src/fastem_calibrations/pattern_calibration.py`

### DIFF
--- a/src/odemis/model/_components.py
+++ b/src/odemis/model/_components.py
@@ -639,7 +639,8 @@ class DigitalCamera(with_metaclass(ABCMeta, Detector)):
         # the Gaussian approximation of a fluorescence microscope point spread
         # function, measured in number of pixels of the camera. This value can
         # be used as a characteristic size parameter when filtering an image
-        # prior to spot detection.
+        # prior to spot detection. To convert this to the FWHM diameter of a
+        # spot multiply by `2 * sqrt(log(4)) ≈ 2.3548`
         self.spotSize = _vattributes.FloatContinuous(1, range=(0, 1e9),
                                                      unit="px", readonly=True)
 
--- a/src/odemis/model/_components.py
+++ b/src/odemis/model/_components.py
@@ -637,9 +637,9 @@ class DigitalCamera(with_metaclass(ABCMeta, Detector)):
 
         # Microscope spot size. This is equivalent to the standard deviation of
         # the Gaussian approximation of a fluorescence microscope point spread
-        # function, measured in no. of pixels of the camera. This value can be
-        # used as a characteristic size parameter when filtering an image prior
-        # to spot detection.
+        # function, measured in number of pixels of the camera. This value can
+        # be used as a characteristic size parameter when filtering an image
+        # prior to spot detection.
         self.spotSize = _vattributes.FloatContinuous(1, range=(0, 1e9),
                                                      unit="px", readonly=True)
 
--- a/src/odemis/model/_components.py
+++ b/src/odemis/model/_components.py
@@ -635,6 +635,11 @@ class DigitalCamera(with_metaclass(ABCMeta, Detector)):
         self.depthOfField = _vattributes.FloatContinuous(1e-6, range=(0, 1e9),
                                                          unit="m", readonly=True)
 
+        # Microscope spot size. This is equivalent to the standard deviation of
+        # the Gaussian approximation of a fluorescence microscope point spread
+        # function, measured in no. of pixels of the camera. This value can be
+        # used as a characteristic size parameter when filtering an image prior
+        # to spot detection.
         self.spotSize = _vattributes.FloatContinuous(1, range=(0, 1e9),
                                                      unit="px", readonly=True)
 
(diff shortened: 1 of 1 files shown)


pieleric (commented):
Can you explain more what the "spot size" mean? We actually have a spotSize VA on some e-beam, that correspond to e-beam diameter, usually in arbitrary unit.  I wonder if it makes sense they have the same name.

It'd be nice to explain a little bit its meaning at the definition of the VA. You can also put more explaination in the commit message, such as giving an example of where it's going to be used. (For now there is no client code).

src/odemis/model/_components.py
```diff
@@ -643,29 +653,37 @@ def __init__(self, name, role, **kwargs):
...
+        except TypeError:
+            logging.warning("Depth of field computed seems incorrect: %f m", dof)
+        except Exception:
+            logging.warning("Failure to update the depth of field", exc_info=True)
+
+        try:
+            sigma = (mag / pxs) * synthetic.psf_sigma_wffm(ri, na, l)
+            self.spotSize._set_value(sigma, force_write=True)
+        except TypeError:
+            logging.warning("Spot size computed seems incorrect: %f px", sigma)
```
pieleric, line 684: That warning is a little tricky because if the first line (in the try block) triggers a `TypeError` (which is not so unlikely as it can happen that values are passed as string instead of float, then sigma is undefined, which would trigger an unhandled exception. Moreover, if the value is out of range, a `IndexError` is raised, and I think it should also be handle here.

I think the easiest is to place `_set_value()` in a second try block, with just this `except`.

(same comment for the new depthOfField code.

pieleric (changes requested):
My main comment is: how about calling it `pointSize` and storing it in meters?

src/odemis/model/_components.py
```diff
@@ -634,6 +634,16 @@ def __init__(self, name, role, **kwargs):
...
                                                          unit="m", readonly=True)
+
+        # Microscope spot size. This is equivalent to the standard deviation of
+        # the Gaussian approximation of a fluorescence microscope point spread
+        # function, measured in number of pixels of the camera. This value can
+        # be used as a characteristic size parameter when filtering an image
+        # prior to spot detection. To convert this to the FWHM diameter of a
+        # spot multiply by `2 * sqrt(log(4)) ≈ 2.3548`
+        self.spotSize = _vattributes.FloatContinuous(1, range=(0, 1e9),
+                                                     unit="px", readonly=True)
```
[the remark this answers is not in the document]
effting: I'd suggest to use `PSFSize` as name then - I think this is more clear.

Similarly, I prefer to keep the unit in pixels. This way it does not matter whether someone wants to use the PSFSize in the 'object' or 'image' plane. If you provide it in meters, you always have to specify which observation plane you are referring to.

Thanks about the binning. It should change with binning. I suggest using `pxs = mdf[_metadata.MD_PIXEL_SIZE][0]` instead of `pxs = self.pixelSize.value[0]`. Note that this will then also change the behaviour of the `depthOfField` VA, but that should also be dependent on binning, correct?

src/odemis/model/_components.py
effting: That first line can only trigger a `TypeError` if the entries in `self._metadata` have the wrong type. That should not happen, right? Or phrased differently, would an unhandled exception in that case not be the desired behaviour?

src/odemis/model/_components.py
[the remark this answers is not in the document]
pieleric: Pretty much any name which is not already used in Odemis is fine for me. That said, to be picky, I'd suggest not using an abreviation. Could we call it `pointSpreadFunctionSize`? Or `pointSpreadSize`? 

I'm not sure what you mean by object or image plane. Do you mean that if it's in meters, it would not be clear as to whether it's converted to pixels based on sensor pixel size or image pixel size? I think that's indeed a little unclear. One way would be to document it. The advantage is that it's binning independent. Otherwise, indeed, we'd need to update whenever the binning changes.

The idea of the depthOfField is to represent a minimum distance to move the *focus* so that the image is "sensibly different". Its main use is to define the smallest step size for the autofocus. I guess it should in theory take into account binning, as a large pixel size would mean you notice less the difference. I *think* that it's be fine to update its behaviour, but we might want to double check that the autofocusing of some system don't go too wrong (I don't expect it, because we don't usually running autofocus with a large binning!!)

src/odemis/model/_components.py
pieleric: The metadata can be written by the user (via updateMetadata() or `odemis-cli --update-metadata`, or the microscope file). So it can be the wrong type. In such case, it's preferable to trigger a TypeError, not a `NameError("name sigma is undefined")` on the `logging.warning(...)` line, as that would be hard to understand for the user.
Note that for the specific metadata here, typically the user is supposed to update them via the VA on the lens, which is then copied into a metadata. So it's not so likely to happen, but still better to report properly.

pieleric (approved):

Sabrina85 (approved):
Looks fine to me.

PR #2000, opened 2022-01-14
driver semcomedi: report that the DAQ board has an error

pieleric:
Update the state if the DAQ board cannot be recovered.
This helps to adjusts the user expectations, instead of letting him/her
find out that no SEM image will every come.

Also try a gentle recovery first. Mostly because the current
reset_device() code seems to sometimes cause more damage than help.
Unfortunately, this type of comedi error happens very rarely, so it's
very hard to reproduce.

### DIFF
--- a/src/odemis/driver/semcomedi.py
+++ b/src/odemis/driver/semcomedi.py
@@ -338,6 +338,12 @@ class SEMComedi(model.HwComponent):
     # We could create a VA for the board temperature.
 
     def _reset_device(self):
+        """
+        Attempts to recover the DAQ board, if it's in a bad state
+        """
+        # TODO: there seems that there a cases where after reseting the device,
+        # a write command causes ComediError(16) Device or resource busy, but
+        # restarting the backend solves the issue. So we probably can do better.
         logging.info("Resetting device %s", self._device_name)
         try:
             comedi.close(self._device)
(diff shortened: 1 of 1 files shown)


src/odemis/driver/semcomedi.py
```diff
@@ -1705,11 +1724,16 @@ def _acquisition_run(self):
                         nfailures += 1
                         if nfailures == 5:
                             logging.exception("Acquisition failed %d times in a row, giving up", nfailures)
+                            self._scanner.state._set_value(
+                                model.HwError("DAQ board non responsive, restart the acquisition to try a new recovery."),
```
Kleijwegt, line 1728: I don't fully get this sentence. Should it be:
"DAQ board non responsive, try restarting the acquisition."?

pieleric: Well, restarting the acquisition is not necessarily going to fix the issue. It'll just force another attempt to use the DAQ board (and from the current experience, most likely fail again). Maybe I should be more honest/pessimistic. What do you think of:
"DAQ board non responsive, try restarting the acquisition or power-cycling the computer."

Kleijwegt: Sounds good maybe this is better (just slightly less pessimistic):
"DAQ board currently non responsive, try restarting the acquisition or power-cycling the computer."

PR #2003, opened 2022-01-17
Chamber tab: pump/vent button - display state instead of action as label

Sabrina85:
It is confusing, that when the chamber is pumped, an action "unload" is displayed as button label. It is more intuitive to display the state "loaded" when pumped and "unloaded" when vented. The ebeam button also indicates a state and not an action. The action for both buttons is defined in the corresponding tooltip.

### DIFF
--- a/src/odemis/gui/cont/microscope.py
+++ b/src/odemis/gui/cont/microscope.py
@@ -152,14 +152,14 @@ class ChamberButtonController(HardwareButtonController):
             self.btn.SetIcon(self._btn_icons['working'])
         elif state == CHAMBER_VACUUM:
             self.btn.SetIcon(self._btn_icons['vacuum'])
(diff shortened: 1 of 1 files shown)


Sabrina85:
Note for the review: This pr affects the FASTEMStateController and the SecomStateController.

tepals (approved):

pieleric (approved):
I do agree it's a little bit more logical. The text matches the button: it represents the current state.

This affects the SECOM in theory, but there are no SECOM which show this button currently.
However it also affects the DELPHI. 

BTW, it's best to write the commit log as title + empty line + description (instead of a long description).

Sabrina85:
> BTW, it's best to write the commit log as title + empty line + description (instead of a long description).

ah yes usually I do this :see_no_evil: but this time I somehow did not.

PR #2005, opened 2022-01-17
FASTEM: always show ROA size if ROA is selected/active

Sabrina85:
Usually the ROA size is only shown when creating an ROA or when resizing it. However, it is convenient to always show the ROA size of the currently active/selected ROA.

### DIFF
--- a/src/odemis/gui/comp/overlay/world.py
+++ b/src/odemis/gui/comp/overlay/world.py
@@ -41,7 +41,8 @@ from odemis.acq.feature import FEATURE_ACTIVE, FEATURE_ROUGH_MILLED, FEATURE_POL
 from odemis.acq.stream import UNDEFINED_ROI
 from odemis.gui import img
 from odemis.gui.comp.overlay.base import Vec, WorldOverlay, Label, SelectionMixin, DragMixin, \
-    PixelDataMixin, SEL_MODE_EDIT, SEL_MODE_CREATE, EDIT_MODE_BOX, EDIT_MODE_POINT, SpotModeBase, SEL_MODE_NONE
+    PixelDataMixin, SEL_MODE_EDIT, SEL_MODE_CREATE, EDIT_MODE_BOX, EDIT_MODE_POINT, SpotModeBase, SEL_MODE_NONE, \
+    SEL_MODE_DRAG
 from odemis.gui.comp.overlay.view import CenteredLineOverlay
 from odemis.gui.model import TOOL_RULER, TOOL_LABEL, TOOL_NONE, TOOL_FEATURE
 from odemis.gui.util import call_in_wx_main
@@ -3049,8 +3050,30 @@ class FastEMROAOverlay(FastEMSelectOverlay):
 
     def draw(self, ctx, shift=(0, 0), scale=1.0):
         """ Draw the selection as a rectangle. Exactly the same as parent function except that
-         it has an adaptive line width (wider if the overlay is active). """
+         it has an adaptive line width (wider if the overlay is active) and it always shows the
+         size label of the selected rectangle. """
         line_width = 5 if self.active.value else 2
+
+        # show size label if ROA is selected
+        if self.active.value:
+            # Important: We need to use the physical positions, in order to draw
+            # everything at the right scale.
+            offset = self.cnvs.get_half_buffer_size()
+            b_start_pos = self.cnvs.phys_to_buffer(self.p_start_pos, offset)
+            b_end_pos = self.cnvs.phys_to_buffer(self.p_end_pos, offset)
+            b_start_pos, b_end_pos = self._normalize_rect(b_start_pos, b_end_pos)
+
+            w, h = (abs(s - e) for s, e in zip(self.p_start_pos, self.p_end_pos))
+            w = units.readable_str(w, 'm', sig=2)
+            h = units.readable_str(h, 'm', sig=2)
+            size_lbl = u"{} x {}".format(w, h)
+
+            pos = Vec(b_end_pos.x - 8, b_end_pos.y + 5)
+
+            self.position_label.pos = pos
+            self.position_label.text = size_lbl
+            self._write_labels(ctx)
+
         super(FastEMROAOverlay, self).draw(ctx, shift, scale, line_width, dash=True)
 
 


src/odemis/gui/comp/overlay/world.py
```diff
@@ -3072,6 +3072,7 @@ def draw(self, ctx, shift=(0, 0), scale=1.0):
 
             self.position_label.pos = pos
             self.position_label.text = size_lbl
+            self.position_label.colour = self.colour  # same color as ROA
```
tepals, line 3075: I see you removed this in the next commit. I'm also not sure this will actually improve readability. Bright green or yellow text does not always work well on a gray background either. Let's think about this a bit more before changing it and maybe talk to Maxim or someone with a bit of design background.

src/odemis/gui/comp/overlay/world.py
```diff
@@ -41,7 +41,8 @@
 from odemis.acq.stream import UNDEFINED_ROI
 from odemis.gui import img
 from odemis.gui.comp.overlay.base import Vec, WorldOverlay, Label, SelectionMixin, DragMixin, \
-    PixelDataMixin, SEL_MODE_EDIT, SEL_MODE_CREATE, EDIT_MODE_BOX, EDIT_MODE_POINT, SpotModeBase, SEL_MODE_NONE
+    PixelDataMixin, SEL_MODE_EDIT, SEL_MODE_CREATE, EDIT_MODE_BOX, EDIT_MODE_POINT, SpotModeBase, SEL_MODE_NONE, \
+    SEL_MODE_DRAG
```
tepals, line 45: I think this is unused

src/odemis/gui/comp/overlay/world.py
```diff
@@ -3049,8 +3050,30 @@ def on_left_up(self, evt):
...
+            b_start_pos = self.cnvs.phys_to_buffer(self.p_start_pos, offset)
+            b_end_pos = self.cnvs.phys_to_buffer(self.p_end_pos, offset)
+            b_start_pos, b_end_pos = self._normalize_rect(b_start_pos, b_end_pos)
+
+            w, h = (abs(s - e) for s, e in zip(self.p_start_pos, self.p_end_pos))
+            w = units.readable_str(w, 'm', sig=2)
+            h = units.readable_str(h, 'm', sig=2)
+            size_lbl = u"{} x {}".format(w, h)
+
+            pos = Vec(b_end_pos.x - 8, b_end_pos.y + 5)
```
tepals, line 3071: why `-8` and `+5`?

tepals (changes requested):

src/odemis/gui/comp/overlay/world.py
Sabrina85: oh yes forgot this when trying to figure out how to do this.

src/odemis/gui/comp/overlay/world.py
Sabrina85: This is code copied from the inherited class. I think it is for not showing it exactly at the top/right corner of the ROA but a little bit shifted to the left.

src/odemis/gui/comp/overlay/world.py
Sabrina85: No it is not removed but added. This is the commit after the other one. How about we choose the Odemis blue or orange? That should be readable. I can also check with Maxim what color he would suggest.

pieleric: The labels with couloured text always have a "shadow" below them to help readability, but arguably, it can be difficult still sometimes. Another approach could be to do as we do for the "Value under cursor", which draws a whole grey box below. I think that ideally we'd have some thin black borders around the letters, like subtitles in movies, but I don't know how to do that with Cairo....

pieleric (approved):

src/odemis/gui/comp/overlay/world.py
tepals: I like the grey box, that does make it more readable and sounds like it's not that hard to implement.

Sabrina85: I found a way to do it, but in the end it took me way longer then thought :laughing: I had to make a change in the Label base class. Please try it out to see if you like it like it is now. I also do not know how the changes in the base class are affecting any other systems. 
I also found a way to draw black borders around letters. I got it working by changing the base class as well. However, finding a good trade off between border and filling was difficult with the small labels. But maybe there are other ways as well. This is what I found - it needs to replace `ctx.show_text("text")`.
```
ctx.set_source_rgb(1, 1, 1)
ctx.text_path(part)
ctx.fill_preserve()
ctx.set_source_rgb(0, 0, 0)
ctx.set_line_width(0.5)
ctx.stroke()
```

tepals (approved):

src/odemis/gui/comp/overlay/base.py
```diff
@@ -239,10 +239,10 @@ def draw(self, ctx, canvas_padding=None, view_width=None, view_height=None):
 
         # Draw background
         if self.background:
-            margin_x = 4  # margins for better representation of background
-            margin_y = 6
-            ctx.set_source_rgba(0.0, 0.0, 0.0, 0.7 * self.opacity)
-            rect = (x - margin_x, y + margin_y, self.text_size[0] + margin_x, -self.text_size[1] - margin_y)
+            margin_x = 6  # margins for better representation of background
+            margin_y = 10
+            ctx.set_source_rgba(*(self.background + (self.opacity,)))
```
pieleric, line 244: That makes sense. However, notice that overlay.view.PixelValueOverlay currently sets the background to 4 floats. So this change breaks it. I'd recommend to adjust PixelValueOverlay (line 240) to use `hex_to_frgb()`. 
Alternatively, you could make it even more clever and detect that there is a 4th value, and so use it to multiply with `self.opacity`. This way, the caller can still choose to have a semi-transparent background (as it was so far) while the letters are fully opaque.

pieleric (changes requested):

src/odemis/gui/comp/overlay/base.py
```diff
@@ -241,7 +241,10 @@ def draw(self, ctx, canvas_padding=None, view_width=None, view_height=None):
         if self.background:
             margin_x = 6  # margins for better representation of background
             margin_y = 10
-            ctx.set_source_rgba(*(self.background + (self.opacity,)))
+            if len(self.colour) == 4:
```
pieleric, line 244: That should be `self.background`, not `self.colour`.

src/odemis/gui/comp/overlay/base.py
```diff
@@ -241,7 +241,10 @@ def draw(self, ctx, canvas_padding=None, view_width=None, view_height=None):
...
+                ctx.set_source_rgba(*self.background[:-1], self.background[-1] * self.opacity)
+            else:
+                ctx.set_source_rgba(*self.background)
```
pieleric, line 247: Next to the else, you can add a comment mentioning that self.background should be len of 3.
In such case, you should use `*(self.background + (self.opacity,)`

pieleric (changes requested):

src/odemis/gui/comp/overlay/base.py
Sabrina85: I was thinking about this, but came to the conclusion that self.opacity is also used for the letters right? So do we want to have both foreground and background being multiplied by the same opacity value (if not specified as fourth value)?

pieleric: I think that if colour or background is only 3 values it should be interpreted as the alpha==1. Right now your code is going to fail as it passes 3 values to a function that expects 4.

As to how to use .opacity when mixed with colour/background having already an alpha channel: it's a little up to us. I think that right now it's not doing anything properly. For the colour, it uses opacity if the colour is 3 channels, and don't use it if it's 4 channels. For the background and shadow it's used always multiplied by 0.7. So it's not consistent.

An alternative would be to drop entirely .opacity, which actually doesn't seem every used. I would be in favour of this last option. Then the user could either set colours with 3 values, in which case opacity is 1, or 4 values, in which case the opacity is the alpha channel.

If we go this way, you could just keep your current code simple (eg, use self.background or self.background + 1). Later on we would create another PR to drop the .opacity attribute.

pieleric (approved):

src/odemis/gui/comp/overlay/base.py
Sabrina85: Yes dropping the .opacity seems to make most sense.

PR #2007, opened 2022-01-20
Added image translation pre-align and autofocus before ROA acquisition.

tepals:
Split off calculating the position of the first tile, so it does not get calculated for each field, but only once.

### DIFF
--- a/src/odemis/acq/fastem.py
+++ b/src/odemis/acq/fastem.py
@@ -30,6 +30,14 @@ from concurrent.futures import CancelledError
 
 import numpy
 
+try:
+    from fastem_calibrations import autofocus_multiprobe, image_translation_pre_align, configure_hw
+
+    fastem_calibrations = True
+except ImportError:
+    logging.info("fastem_calibrations package not found")
+    fastem_calibrations = False
+
 from odemis import model, util
 from odemis.acq import stitching
 from odemis.acq.align.spot import FindGridSpots
@@ -37,7 +45,6 @@ from odemis.acq.stitching import REGISTER_IDENTITY, FocusingMethod
 from odemis.acq.stream import SEMStream
 from odemis.util import TimeoutError
 from odemis.acq import fastem_conf
-
 
 # The executor is a single object, independent of how many times the module (fastem.py) is loaded.
 _executor = model.CancellableThreadPoolExecutor(max_workers=1)
@@ -166,7 +173,7 @@ class FastEMROC(object):
         # if None -> acquire, if not None, don't acquire again
 
 
-def acquire(roa, path, scanner, multibeam, descanner, detector, stage, ccd, beamshift, lens):
+def acquire(roa, path, scanner, multibeam, descanner, detector, stage, ccd, beamshift, lens, pre_calibrate=False):
     """
     Start a megafield acquisition task for a given region of acquisition (ROA).
 
@@ -185,6 +192,7 @@ def acquire(roa, path, scanner, multibeam, descanner, detector, stage, ccd, beam
     :param ccd: (model.DigitalCamera) A camera object of the diagnostic camera.
     :param beamshift: (tfsbc.BeamShiftController) Component that controls the beamshift deflection.
     :param lens: (static.OpticalLens) Optical lens component.
+    :param pre_calibrate: (bool) If True run pre-calibrations before each ROA acquisition.
 
     :return: (ProgressiveFuture) Acquisition future object, which can be cancelled. The result of the future is
              a tuple that contains:
@@ -195,7 +203,8 @@ def acquire(roa, path, scanner, multibeam, descanner, detector, stage, ccd, beam
 
     # TODO: pass path through attribute on ROA instead of argument?
     # Create a task that acquires the megafield image.
-    task = AcquisitionTask(scanner, multibeam, descanner, detector, stage, ccd, beamshift, lens, roa, path, f)
+    task = AcquisitionTask(scanner, multibeam, descanner, detector, stage, ccd, beamshift, lens, roa, path, f,
+                           pre_calibrate=pre_calibrate)
 
     f.task_canceller = task.cancel  # lets the future know how to cancel the task.
 
@@ -212,7 +221,8 @@ class AcquisitionTask(object):
     An ROA consists of multiple single field images.
     """
 
-    def __init__(self, scanner, multibeam, descanner, detector, stage, ccd, beamshift, lens, roa, path, future):
+    def __init__(self, scanner, multibeam, descanner, detector, stage, ccd, beamshift, lens, roa, path, future,
+                 pre_calibrate=False):
         """
         :param scanner: (xt_client.Scanner) Scanner component connecting to the XT adapter.
         :param multibeam: (technolution.EBeamScanner) The multibeam scanner component of the acquisition server module.
@@ -232,6 +242,7 @@ class AcquisitionTask(object):
                             (model.DataArray): The acquisition data, which depends on the value of the
                                                detector.dataContent VA.
                             (Exception or None): Exception raised during the acquisition or None.
+        :param pre_calibrate: (bool) If True run pre-calibrations before each ROA acquisition.
         """
         self._scanner = scanner
         self._multibeam = multibeam
@@ -245,10 +256,12 @@ class AcquisitionTask(object):
         self._roc = roa.roc  # region of calibration object
         self._path = path  # sub-directories on external storage
         self._future = future
+        self._pre_calibrate = pre_calibrate
 
         # Dictionary containing the single field images with index as key: e.g. {(0,1): DataArray}.
         self.megafield = {}
         self.field_idx = (0, 0)
+        self._pos_first_tile = None  # only calculate the position of the first tile when `run` is called.
 
         # TODO the .dataContent might need to be set somewhere else in future when using a live stream for
         #  display of thumbnail images -> .dataContent = "thumbnail"
@@ -276,6 +289,15 @@ class AcquisitionTask(object):
         :raise:
             Exception: If it failed before any single field images were acquired or if acquisition was cancelled.
         """
+        # Update the position of the first tile.
+        self._pos_first_tile = self.get_pos_first_tile()
+
+        # Move the stage to the first tile, to ensure the autofocus and image translation pre-align are
+        # done in the correct location and the correct position is stored in the megafield metadata yaml file.
+        self.move_stage_to_next_tile()
+
+        if self._pre_calibrate:
+            self.pre_calibrate()
 
         # set the sub-directories (<acquisition date>/<project name>) and megafield id
         self._detector.filename.value = os.path.join(self._path, self._roa.name.value)
@@ -347,6 +369,8 @@ class AcquisitionTask(object):
 
             self.move_stage_to_next_tile()  # move stage to next field image position
 
+            # Beamshift correction is not necessary on the first field, because the pattern was just centered on the
+            # mppc detector using the image translation pre-alignment.
             if field_idx != (0, 0):
                 self.correct_beam_shift()  # correct the shift of the beams caused by the parasitic magnetic field.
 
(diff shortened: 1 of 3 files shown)


tepals:
Still needs to be tested

src/odemis/acq/fastem.py
```diff
@@ -276,6 +277,31 @@ def run(self):
...
+
+        # Update the position of the first tile.
+        self._pos_first_tile = self.get_pos_first_tile()
+
+        # Move the stage to the first tile, to ensure the autofocus and image
+        # translation pre-align are done in the correct location.
+        self.move_stage_to_next_tile()
+
+        # Autofocus multiprobe ensures the beams are focused at the start of the ROA acquisition.
+        autofocus_multiprobe.run_autofocus(self._scanner,
```
tepals, line 290: I wonder if we need a try except around this and the run_image_translation_pre_align.

src/odemis/acq/fastem.py
```diff
@@ -276,6 +277,31 @@ def run(self):
         :raise:
             Exception: If it failed before any single field images were acquired or if acquisition was cancelled.
         """
+        self._beamshift.shift.value = (0, 0)
```
Sabrina85, line 280: Can you add a comment why it needs to be reset before each ROA?

src/odemis/acq/fastem.py
```diff
@@ -415,14 +440,25 @@ def get_abs_stage_movement(self):
         # therefore add half a field in the x-direction and subtract half a field in the y-direction.
         pos_first_tile = (xmin_roa + field_res[0]/2 * px_size[0],
                           ymax_roa - field_res[1]/2 * px_size[1])
+        return pos_first_tile
```
Sabrina85, line 443: Can you add an empty line before the return statement? I find this clearer arranged and matches the rest of the style of this module.

src/odemis/acq/fastem.py
```diff
@@ -276,6 +277,31 @@ def run(self):
...
+        # Autofocus multiprobe ensures the beams are focused at the start of the ROA acquisition.
+        autofocus_multiprobe.run_autofocus(self._scanner,
+                                           self._multibeam,
+                                           self._descanner,
+                                           self._detector,
+                                           self._detector.data,
+                                           self._ccd,
+                                           self._stage)
+
+        # Image translation pre-alignment ensures the image of the multiprobe is roughly centered on the detector.
```
Sabrina85, line 298: detector -> mppc detector

src/odemis/acq/fastem.py
```diff
@@ -276,6 +277,31 @@ def run(self):
...
             Exception: If it failed before any single field images were acquired or if acquisition was cancelled.
         """
+        self._beamshift.shift.value = (0, 0)
+
+        # Update the position of the first tile.
+        self._pos_first_tile = self.get_pos_first_tile()
+
+        # Move the stage to the first tile, to ensure the autofocus and image
+        # translation pre-align are done in the correct location.
+        self.move_stage_to_next_tile()
```
Sabrina85, line 287: This is actually super nice as it also solves another bug we have. We had the incorrect stage position in the MD stored as yaml file on the external storage with the megafield. As you call this now before the .subscribe, also this should be solved. Can you add a comment that this also ensures that the correct stage position is stored in the MD on the megafield (and test it?)?

src/odemis/acq/fastem.py
Sabrina85: I have the feeling we should do that. If the calibration fails, it is nice if the exception is also handled by the caller as for the ROA acquisition I guess and also that we can put the system back into a "good state" with a finally statement. If I am not mistaken the exception for the ROA is handled by the callback here and that would put the GUI also back into a "good state".
```
    def on_acquisition_done(self, future):
        """
        Callback called when the acquisition is finished (either successfully or
        cancelled)
        """

        try:
            future.result()
            self._reset_acquisition_gui()
        except CancelledError:
            self._reset_acquisition_gui()
            return
        except Exception as exp:
            # leave the gauge, to give a hint on what went wrong.
            logging.exception("Acquisition failed")
            self._reset_acquisition_gui("Acquisition failed (see log panel).", level=logging.WARNING)
            return
```

Sabrina85 (changes requested):
I looks already pretty good!

src/odemis/acq/fastem.py
```diff
@@ -166,7 +172,7 @@ def __init__(self, name, coordinates):
         # if None -> acquire, if not None, don't acquire again
 
 
-def acquire(roa, path, scanner, multibeam, descanner, detector, stage, ccd, beamshift, lens):
+def acquire(roa, path, scanner, multibeam, descanner, detector, stage, ccd, beamshift, lens, pre_calibrate=False):
```
Sabrina85, line 175: Can you add this to the doc string?

src/odemis/acq/fastem.py
```diff
@@ -195,10 +201,15 @@ def acquire(roa, path, scanner, multibeam, descanner, detector, stage, ccd, beam
...
     # Create a task that acquires the megafield image.
-    task = AcquisitionTask(scanner, multibeam, descanner, detector, stage, ccd, beamshift, lens, roa, path, f)
+    task = AcquisitionTask(scanner, multibeam, descanner, detector, stage, ccd, beamshift, lens, roa, path, f,
+                           pre_calibrate=pre_calibrate)
 
     f.task_canceller = task.cancel  # lets the future know how to cancel the task.
 
+    if pre_calibrate:
+        # NEW:
+        _executor.submitf(f, task.pre_calibrate)
```
Sabrina85, line 211: Does the executor take the order of calibration and acquisition into account? So that when you first submit the pre-calibrate and then the run, is that order kept?

src/odemis/acq/fastem.py
```diff
@@ -195,10 +201,15 @@ def acquire(roa, path, scanner, multibeam, descanner, detector, stage, ccd, beam
...
     # TODO: pass path through attribute on ROA instead of argument?
     # Create a task that acquires the megafield image.
-    task = AcquisitionTask(scanner, multibeam, descanner, detector, stage, ccd, beamshift, lens, roa, path, f)
+    task = AcquisitionTask(scanner, multibeam, descanner, detector, stage, ccd, beamshift, lens, roa, path, f,
+                           pre_calibrate=pre_calibrate)
 
     f.task_canceller = task.cancel  # lets the future know how to cancel the task.
 
+    if pre_calibrate:
+        # NEW:
```
Sabrina85, line 210: This comment is not finalized I assume?

src/odemis/acq/fastem.py
```diff
@@ -212,7 +223,8 @@ class AcquisitionTask(object):
     An ROA consists of multiple single field images.
     """
 
-    def __init__(self, scanner, multibeam, descanner, detector, stage, ccd, beamshift, lens, roa, path, future):
+    def __init__(self, scanner, multibeam, descanner, detector, stage, ccd, beamshift, lens, roa, path, future,
+                 pre_calibrate=False):
```
Sabrina85, line 227: Extend the doc string also here.

src/odemis/acq/fastem.py
```diff
@@ -266,6 +279,62 @@ def __init__(self, scanner, multibeam, descanner, detector, stage, ccd, beamshif
         # Threading event, which keeps track of when image data has been received from the detector.
         self._data_received = threading.Event()
 
+    def pre_calibrate(self):
```
Sabrina85, line 282: I would prefer to have the run function first after the init and then all the functions that are called in the run function. What do you think?

Sabrina85, line 282: Add doc string.

src/odemis/acq/fastem.py
```diff
@@ -266,6 +279,62 @@ def __init__(self, scanner, multibeam, descanner, detector, stage, ccd, beamshif
...
+        exception = None
+        if self._pre_calibrate and fastem_calibrations:
+            try:
+                fastem_conf.configure_scanner(self._scanner, fastem_conf.MEGAFIELD_MODE)
+                self._beamshift.shift.value = (0, 0)
```
Sabrina85, line 287: Does it not make sense to reset the beamshift also when you do not pre-calibrate? So that the magnetic field correction is also not running into out of range trouble?

src/odemis/acq/fastem.py
```diff
@@ -266,6 +279,62 @@ def __init__(self, scanner, multibeam, descanner, detector, stage, ccd, beamshif
...
+                logging.debug("Read initial Hw settings.")
+                asm_config_orig = configure_hw.get_config_asm(self._multibeam, self._descanner, self._detector)
+                # Autofocus multiprobe ensures the beams are focused at the start of the ROA acquisition.
+                logging.debug("Running optical autofocus before ROA acquisition.")
+                autofocus_multiprobe.run_autofocus(
+                    self._scanner, self._multibeam, self._descanner,
```
Sabrina85, line 293: Either both brackets in separate line or move this code line one up.

src/odemis/acq/fastem.py
```diff
@@ -266,6 +279,62 @@ def __init__(self, scanner, multibeam, descanner, detector, stage, ccd, beamshif
...
+                    self._detector, self._detector.data, self._ccd, self._stage)
+
+                configure_hw.configure_asm(self._multibeam,
+                                           self._descanner,
+                                           self._detector,
+                                           self._detector.data,
+                                           asm_config_orig,
+                                           upload=False)
+                # Image translation pre-alignment ensures the image of the multiprobe is roughly centered on the
```
Sabrina85, line 302: '...ensures that the image...'

src/odemis/acq/fastem.py
```diff
@@ -266,6 +279,62 @@ def __init__(self, scanner, multibeam, descanner, detector, stage, ccd, beamshif
...
+                # mppc detector.
+                logging.debug("Running image translation pre alignment before ROA acquisition.")
+                descanner_offset = image_translation_pre_align.run_image_translation_pre_align(self._scanner,
+                                                                                               self._multibeam,
+                                                                                               self._descanner,
+                                                                                               self._detector,
+                                                                                               self._detector.data,
+                                                                                               self._ccd)
+
+                asm_config_orig["descanner"]["scanOffset"] = descanner_offset
```
Sabrina85, line 312: Can you add a comment here please?

src/odemis/acq/fastem.py
```diff
@@ -266,6 +279,62 @@ def __init__(self, scanner, multibeam, descanner, detector, stage, ccd, beamshif
...
+                configure_hw.configure_asm(self._multibeam,
+                                           self._descanner,
+                                           self._detector,
+                                           self._detector.data,
+                                           asm_config_orig,
+                                           upload=False)
+
+            except Exception as ex:
+                # If image data was already acquired, just log a warning.
+                logging.warning("Exception during roa acquisition (after some data has already been acquired).",
```
Sabrina85, line 326: This is not finished yet right?

src/odemis/acq/fastem.py
```diff
@@ -286,28 +355,14 @@ def run(self):
...
+        exception = self.pre_calibrate()
 
-        # Image translation pre-alignment ensures the image of the multiprobe is roughly centered on the detector.
-        image_translation_pre_align.run_image_translation_pre_align(self._scanner,
-                                                                    self._multibeam,
-                                                                    self._descanner,
-                                                                    self._detector,
-                                                                    self._detector.data,
-                                                                    self._ccd)
+        if exception is not None:
```
Sabrina85, line 360: Can you add a comment here: "If calibration failed, don't acquire ROA" or similar

src/odemis/acq/fastem.py
```diff
@@ -373,8 +428,7 @@ def acquire_roa(self, dataflow):
 
             self.move_stage_to_next_tile()  # move stage to next field image position
 
-            if field_idx != (0, 0):
-                self.correct_beam_shift()  # correct the shift of the beams caused by the parasitic magnetic field.
+            self.correct_beam_shift()  # correct the shift of the beams caused by the parasitic magnetic field.
```
Sabrina85, line 431: So you do it also now for the first field? Why is that?

src/odemis/acq/test/fastem_test.py
```diff
@@ -535,6 +543,52 @@ def test_get_abs_stage_movement(self):
...
+                                      self.mppc, self.stage, self.ccd,
+                                      self.beamshift, self.lens,
+                                      roa, path=None, future=None, pre_calibrate=True)
+
+        self.descanner.updateMetadata({model.MD_SCAN_GAIN: (5000, 5000)})
+
+        task.pre_calibrate = MagicMock(name='method')
+        task.pre_calibrate.return_value = None
+
+        scan_offset = (0.5, 0.5)
```
Sabrina85, line 577: I wonder if we should use values that also make sense when you run this test on real HW, so that the calibration still works.

src/odemis/acq/test/fastem_test.py
```diff
@@ -535,6 +543,52 @@ def test_get_abs_stage_movement(self):
             actual_position = task.get_abs_stage_movement()  # [m]
             numpy.testing.assert_allclose(actual_position, expected_position)
 
+    def test_pre_calibrate(self):
+        """
+        Test the correct stage positions are returned for the corner fields of
+        ROAs consisting of a varying number of single fields.
```
Sabrina85, line 549: You only test now for one megafield right (which is enough I think)?

Sabrina85 (changes requested):

src/odemis/acq/fastem.py
```diff
@@ -346,23 +289,22 @@ def run(self):
...
-        # translation pre-align are done in the correct location.
+        # Move the stage to the first tile, to ensure the autofocus and image translation pre-align are
+        # done in the correct location and the correct position is stored in the megafield metadata yaml file.
         self.move_stage_to_next_tile()
 
-        exception = self.pre_calibrate()
-
-        if exception is not None:
-            return None, exception
+        # If we run the pre-calibrate check those exceptions, else initialize exceptions as None.
```
Sabrina85, line 299: I am confused by the comment. Where do you check the exceptions? You catch exception in the self.pre_calibrate(). If an exception is raised, it is handled by the caller right?
The exception for the ROA acquisition is always initialized with None. Independent of the pre-calibrations. I suggest to just remove the comment completely unless I misunderstood what you wanted to say.

src/odemis/acq/fastem.py
```diff
@@ -452,6 +397,77 @@ def acquire_roa(self, dataflow):
...
+            # set back the initial values on the hardware
+            configure_hw.configure_asm(self._multibeam,
+                                       self._descanner,
+                                       self._detector,
+                                       self._detector.data,
+                                       asm_config_orig,
+                                       upload=False)
+
+        except Exception as ex:
+            # If image data was already acquired, just log a warning.
```
Sabrina85, line 461: This comment seems to be left over from copy/paste.

src/odemis/acq/fastem.py
```diff
@@ -452,6 +397,77 @@ def acquire_roa(self, dataflow):
 
         return self.megafield
 
+    def pre_calibrate(self):
+        """
+        Run optical multiprobe autofocus and image translation pre-alignment before the ROA acquisition.
+        The image translation pre-alignment adjusts the descanner.Scanoffset VA such that the image of the
```
Sabrina85, line 403: descaner.scanOffset

src/odemis/acq/fastem.py
```diff
@@ -452,6 +397,77 @@ def acquire_roa(self, dataflow):
...
+                                                                                           self._detector.data,
+                                                                                           self._ccd)
+
+            # Set the descanner offset to the value calibrated with the image translation pre-alignment.
+            asm_config_orig["descanner"]["scanOffset"] = descanner_offset
+            logging.debug(f"Descanner offset set to: {descanner_offset}")
+            # Blank the beam to reduce beam damage on the sample.
+            self._scanner.blanker.value = True
+            # set back the initial values on the hardware
+            configure_hw.configure_asm(self._multibeam,
```
Sabrina85, line 453: Should this go into the finally? If the code fails beforehand, the system might be in a bad state right?

src/odemis/acq/test/fastem_test.py
```diff
@@ -545,8 +545,7 @@ def test_get_abs_stage_movement(self):
 
     def test_pre_calibrate(self):
         """
-        Test the correct stage positions are returned for the corner fields of
-        ROAs consisting of a varying number of single fields.
+        Test the ASM settings, except the descanner scan offset, are unchanged after running the pre-calibrations.
```
Sabrina85, line 548: Can you reorder the sentence and put the except part at the end?

src/odemis/acq/test/fastem_test.py
```diff
@@ -589,6 +583,36 @@ def test_pre_calibrate(self):
...
+                                   self.mppc)
+
+            task = fastem.AcquisitionTask(self.scanner, self.multibeam, self.descanner,
+                                          self.mppc, self.stage, self.ccd,
+                                          self.beamshift, self.lens,
+                                          roa, path=None, future=None)
+
+            pos_first_tile_actual = task.get_pos_first_tile()
+
+            # The first tile is expected to be to the bottom right of the top left corner of the ROA in the
```
Sabrina85, line 610: This doesn't make sense to me. Why "bottom right".
With first tile you mean "The position of the first tile"?
Isn't the position of the first tile the center position of the top/left corner tile of the ROA?

src/odemis/acq/fastem.py
```diff
@@ -452,6 +397,77 @@ def acquire_roa(self, dataflow):
...
         return self.megafield
 
+    def pre_calibrate(self):
+        """
+        Run optical multiprobe autofocus and image translation pre-alignment before the ROA acquisition.
+        The image translation pre-alignment adjusts the descanner.Scanoffset VA such that the image of the
+        multiprobe is roughly centered on the mppc detector. This function reads in the ASM configuration
+        and makes sure all values, except the descanner offset, are set back after the calibrations are run.
+
+        :return: (str) exception, if an exception is raised the exception message is returned, else None is returned.
```
Sabrina85, line 407: Can you add that cancelling is currently not supported.

Sabrina85 (changes requested):

src/odemis/acq/fastem.py
tepals: This was indeed left over from a previous commit

src/odemis/acq/fastem.py
tepals: that makes sense!

Sabrina85 (approved):

PR #2009, opened 2022-01-20
plugins timelapse: add support for ENZEL and METEOR

pieleric:
The main difference is that the name of the tab is different.
Also, the .acquisitionStream attribute is slightly different.

This is enough to get the plugin working.


Kleijwegt:
Looks good, though we currently can't get into the localisation tab for Enzel because of the stage positions tags.

Kleijwegt (approved):

PR #2012, opened 2022-01-24
technolution driver: update the flyback time

Sabrina85:
150us have proven to be enough for the flyback, so change the default value on the VA.

### DIFF
--- a/src/odemis/acq/test/fastem_test.py
+++ b/src/odemis/acq/test/fastem_test.py
@@ -146,9 +146,6 @@ class TestFastEMROA(unittest.TestCase):
     def tearDownClass(cls):
         pass
 
-    def setUp(self):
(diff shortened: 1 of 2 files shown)


src/odemis/acq/test/fastem_test.py
```diff
@@ -147,7 +147,7 @@ def tearDownClass(cls):
         pass
 
     def setUp(self):
-        self.descanner.physicalFlybackTime.value = 250e-6  # TODO why is this necessary??
+        self.descanner.physicalFlybackTime.value = 150e-6  # TODO why is this necessary??
```
tepals, line 150: I just tested this, and if I remove this setUp, all tests still behave as expected and the flyback time is correctly set. So I think this is not necessary and you can remove the whole setUp.

tepals (changes requested):

tepals (approved):

PR #2014, opened 2022-01-28
Fastem acquisition test cases: fixes and code clean up.

tepals:
Fixed test_estimate_acquisition_time, test_stage_movement and test_pre_calibrate because they failed due to changes in the code.
Added the use of the fastem-sim-asm yaml file, for tests that need an asm component to run.
Minor changes to make the coding style consistent over the entire module.

### DIFF
--- a/src/odemis/acq/test/fastem_test.py
+++ b/src/odemis/acq/test/fastem_test.py
@@ -39,16 +39,17 @@ from odemis import model
 from odemis.acq import fastem, stream
 from odemis.util import test, img
 
-# * TEST_NOHW = 1: not connected to anything => skip most of the tests
-# * TEST_NOHW = 0: connected to the real hardware or simulator
-# technolution_asm_simulator/simulator2/run_the_simulator.py
-TEST_NOHW = (os.environ.get("TEST_NOHW", "0") != "0")  # Default is HW/simulator testing
+# * TEST_NOHW = 1: connected to the simulator or not connected to anything
+# * TEST_NOHW = 0: connected to the real hardware, the backend should be running
+# technolution_asm_simulator/simulator2/run_the_simulator.sh
+TEST_NOHW = (os.environ.get("TEST_NOHW", "0") != "0")  # Default is hardware testing
 
 logging.getLogger().setLevel(logging.DEBUG)
 logging.basicConfig(format="%(asctime)s  %(levelname)-7s %(module)s:%(lineno)d %(message)s")
 
 CONFIG_PATH = os.path.dirname(odemis.__file__) + "/../../install/linux/usr/share/odemis/"
 FASTEM_CONFIG = CONFIG_PATH + "sim/fastem-sim.odm.yaml"
+FASTEM_CONFIG_ASM = CONFIG_PATH + "sim/fastem-sim-asm.odm.yaml"
 
 
 class TestFASTEMOverviewAcquisition(unittest.TestCase):
@@ -132,10 +133,9 @@ class TestFastEMROA(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
         if TEST_NOHW:
-            raise unittest.SkipTest("No simulator running or HW present. Skip fastem ROA tests.")
+            test.start_backend(FASTEM_CONFIG_ASM)
 
         # get the hardware components
-        cls.microscope = model.getMicroscope()
         cls.asm = model.getComponent(role="asm")
         cls.mppc = model.getComponent(role="mppc")
         cls.multibeam = model.getComponent(role="multibeam")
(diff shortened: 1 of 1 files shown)


src/odemis/acq/test/fastem_test.py
```diff
@@ -42,47 +42,30 @@
 # * TEST_NOHW = 1: not connected to anything => skip most of the tests
```
Sabrina85, line 42: This needs to be update to simulator testing.

src/odemis/acq/test/fastem_test.py
```diff
@@ -437,30 +427,37 @@ def on_progress_update(self, future, start, end):
...
-        else:
-            # get the hardware components from the simulators or hardware
+            try:
+                test.start_backend(FASTEM_CONFIG_ASM)
+            except Exception:
+                logging.info("Failed to start testing, using mocked components for testing.")
+                cls.mocked_testing = True  # If we cannot start the backend, run the mocked test cases.
+                # If we are testing without hardware we just need a few attributes to be set correctly.
+                cls.scanner = None
+                cls.asm = None
```
pieleric, line 444: It's quite odd to run a test case differently if the init fails. I'd suggest to have another test case which runs with the mock components. It can inherit from the TestFastEMAcquisitionTask, and override a few of the methods/tests that are not supported. Here, if starting the backend fails, it should just fail!

pieleric (changes requested):

src/odemis/acq/test/fastem_test.py
```diff
@@ -108,16 +109,18 @@ def test_overview_acquisition(self):
         self.assertGreaterEqual(bbox[3], scintillator5_area[3])  # Top
 
 
+@unittest.skip("Can't run ROA tests until all simulator dependencies have been installed on the test VM.")
```
Sabrina85, line 112: What is the reason to not do it now?

pieleric: Typically, I think we shouldn't care about the "test VM". If the test cases are expected to pass (and I hope they are!), then they should be run. If the "test VM" has some limitation, so be it, but we just have to interpret the results from the test VM, not skip the tests on every computer.

tepals: Ok, then I'll remove it.

> What is the reason to not do it now?
It will take some time install the simulator and update the test script to start the simulator and xtadapter. As there are already so many test cases that fail, I discussed with @Kleijwegt to first fix the rest of the test cases before adding other things.

Sabrina85 (approved):

src/odemis/acq/test/fastem_test.py
```diff
@@ -212,8 +206,10 @@ class TestFastEMAcquisition(unittest.TestCase):
 
     @classmethod
     def setUpClass(cls):
-        if TEST_NOHW is True:
-            raise unittest.SkipTest("No simulator running or HW present. Skip fastem acquisition tests.")
+        if TEST_NOHW:
+            test.start_backend(FASTEM_CONFIG_ASM)
+        else:
+            raise unittest.SkipTest("No hardware backend file available, skipping tests.")
```
pieleric, line 212: Don't skip, run with the real hardware.

src/odemis/acq/test/fastem_test.py
```diff
@@ -437,40 +428,30 @@ def on_progress_update(self, future, start, end):
...
-            cls.asm = model.getComponent(role="asm")
-            cls.mppc = model.getComponent(role="mppc")
-            cls.multibeam = model.getComponent(role="multibeam")
-            cls.descanner = model.getComponent(role="descanner")
-            cls.stage = model.getComponent(
-                role="stage")  # TODO replace with stage-scan when ROA conversion method available
-            cls.ccd = model.getComponent(role="diagnostic-ccd")
-            cls.beamshift = model.getComponent(role="ebeam-shift")
-            cls.lens = model.getComponent(role="lens")
+            raise unittest.SkipTest("No hardware backend file available, skipping tests.")
```
pieleric, line 438: Don't skip.

src/odemis/acq/test/fastem_test.py
```diff
@@ -437,40 +428,30 @@ def on_progress_update(self, future, start, end):
...
+        cls.mppc = model.getComponent(role="mppc")
+        cls.multibeam = model.getComponent(role="multibeam")
+        cls.descanner = model.getComponent(role="descanner")
+        cls.stage = model.getComponent(
+            role="stage")  # TODO replace with stage-scan when ROA conversion method available
+        cls.ccd = model.getComponent(role="diagnostic-ccd")
+        cls.beamshift = model.getComponent(role="ebeam-shift")
+        cls.lens = model.getComponent(role="lens")
+
+        cls.beamshift.updateMetadata({model.MD_CALIB: cls.scanner.beamShiftTransformationMatrix.value})
```
pieleric, line 452: Please add a comment why you have to do it "manually". Where is it done normally?

pieleric (changes requested):
I'm confused as to why the test cases are not executed when TEST_NOHW is not set. Typically, we assume that there is already a correct backend running, and we run all the tests.

tepals:
@pieleric can you have another look?

pieleric (approved):

PR #2016, opened 2022-01-31
gui: small fixes and improvements

pieleric:


src/odemis/gui/test/comp_foldpanelbar_test.py
```diff
@@ -302,33 +298,20 @@ def test_foldpanel_manipulation(self):
 
         new_labels = []
 
-        # Normally, 5 children
-        for dummy in range(5):
+        # Add 15 more children
```
tepals, line 301: 15 should be 16

tepals (changes requested):

PR #2019, opened 2022-02-02
Fixes for comp canvas tests.

tepals:
The test cases assumed that when a green and a blue stream were added the resulting image would be 1/3 green, 2/3 blue, while it is half green and half blue.
In test_pyramidal_zoom added a precision for checking the mpp. The precision still needs to be checked, because I am not sure what the correct prescision is, but the test only passes with places=6.

### DIFF
--- a/src/odemis/gui/test/comp_canvas_test.py
+++ b/src/odemis/gui/test/comp_canvas_test.py
@@ -24,6 +24,8 @@ Odemis. If not, see http://www.gnu.org/licenses/.
 from __future__ import division
 
 import logging
+import math
+
 import numpy
 from odemis import model
 from odemis.acq.stream import RGBStream
@@ -489,6 +491,11 @@ class TestDblMicroscopeCanvas(test.GuiTestCase):
         self.view.addStream(stream1)
         self.view.addStream(stream2)
 
+        # Ensure the merge ratio of the images is 0.5
+        ratio = 0.5
+        self.view.merge_ratio.value = ratio
+        self.assertEqual(ratio, self.view.merge_ratio.value)
+
         test.gui_loop(0.5)
 
         self.canvas.shift_view((-200.5, 199.5))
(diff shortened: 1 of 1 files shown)


src/odemis/gui/test/comp_canvas_test.py
```diff
@@ -600,11 +600,11 @@ def test_pyramidal_zoom(self):
...
         px2 = get_rgb(result_im, result_im.Width // 2, result_im.Height // 2)
-        # center pixel, 2/3 green, 1/3 blue. The green image is the largest image
-        self.assertEqual(px2, (0, 179, 76))
+        # center pixel, half green, half blue. The green image is the largest image
+        self.assertEqual(px2, (0, 127, 128))
         px2 = get_rgb(result_im, result_im.Width // 2 - 30, result_im.Height // 2 - 30)
-        # background of the images, 2/3 green, 1/3 red
-        self.assertEqual(px2, (76, 179, 0))
+        # background of the images, half green, half red
+        self.assertEqual(px2, (128, 127, 0))
```
Kleijwegt, line 607: I don't get why the order of the color of the pixel (0, 127, 128) is different here than in the test case above.

tepals: I can add some comments. `get_rgb` takes as input arguments `image, x, y`, it then returns the value of that pixel. The image is half green, half red and has a blue-green square in the middle. This `result_im.Width // 2 - 30, result_im.Height // 2 - 30` just gets the rgb value of a pixel 30 pixels to the top left of the center.

pieleric (changes requested):

src/odemis/gui/test/comp_canvas_test.py
```diff
@@ -659,7 +659,7 @@ def test_pyramidal_zoom(self):
         # only .mpp changes, but the image keeps centered
         exp_mpp = (mpp * im2.shape[0]) / self.canvas.ClientSize[0]
         # TODO: check the precision
-        self.assertAlmostEqual(exp_mpp, self.view.mpp.value)  # ,6
+        self.assertAlmostEqual(exp_mpp, self.view.mpp.value, places=6)  # ,6
```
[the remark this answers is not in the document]
tepals: I think I found the issue, the window fitted tightly in the vertical direction, but not in horizontal. If I do `exp_mpp = (mpp * im2.shape[1]) / self.canvas.ClientSize[1]` I get the correct result.

image

tepals:
@Kleijwegt and @pieleric I now actually found out that these tests were failing because the default merge ratio was changed from 0.3 to 0.5 in commit [d7927b253455baab02e4738572473c9e780712b2](https://github.com/delmic/odemis/commit/d7927b253455baab02e4738572473c9e780712b2#diff-66db26a9631fdc384a494312f30917a7f083bba98e21a7e5cded1769fdbd041bR1144), could it be that this was done by accident? Then I shouldn't update the test cases, but I should update the default merge ratio.

Kleijwegt:
> @Kleijwegt and @pieleric I now actually found out that these tests were failing because the default merge ratio was changed from 0.3 to 0.5 in commit [d7927b253455baab02e4738572473c9e780712b2](https://github.com/delmic/odemis/commit/d7927b253455baab02e4738572473c9e780712b2#diff-66db26a9631fdc384a494312f30917a7f083bba98e21a7e5cded1769fdbd041bR1144), could it be that this was done by accident? Then I shouldn't update the test cases, but I should update the default merge ratio.

Looking at the comments on that line in the PR, [Kleijwegt](https://github.com/delmic/odemis/pull/1313/files#r364641763):
> Is change because this gives a better look onto the live view because of the transparency of the live image

And the fact that it is even included in the name of the commit it suppose this change was made on purpose. Meaning the test case was never updated while it should have been.
What do you think @pieleric?

pieleric:
I think the commit made sense. The 0.3 default merge ratio was really just some odd arbitrary value. 0.5 (ie, 50%) is easier to understand as default value for showing 2 images.
It's unfortunate that we didn't notice that this breaks this test case until now... Let's just fix the test case. Possibly, you could explicitly set the .merge_ratio to 0.5 in the test case to indicate the link.

pieleric (approved):

Kleijwegt (approved):

PR #2020, opened 2022-02-02
TEST_NOHW updated the consistency of getting the environment variable.

tepals:
In some places it was defined as TEST_NOHW = (os.environ.get("TEST_NOHW", 0) != 0). This usually worked because often we set TEST_NOHW = "1".
However environment variables are always strings and if TEST_NOHW = "0", the above will evaluate to ("0" != 0 ) = True.
TEST_NOHW is now everywhere defined as TEST_NOHW = (os.environ.get("TEST_NOHW", "0") != "0")


pieleric (approved):
Naice!!!

Sabrina85 (approved):
Indeed nice that it is the same now everywhere. Less confusions :)

PR #2022, opened 2022-02-03
Added pytest test report functionality

Kleijwegt:
This outputs the test cases results a bit cleaner and only shows the failed testcases. It still needs some refinement (removing extra new lines, 1 corner case where the entire testcase is outputted) but in general it is faster to produce a readable test report this way.
The script is based on the script runtests.sh and as little as possible is changed to make the results as identical as possible.
This is the layout this test report is as follows:

```

Running /home/testing/development/odemis/src/odemis/driver/test/tmcm_test.py...
ERROR: test_cancel (__main__.TestActuatorRelEnc)
  File "/home/testing/development/odemis/src/odemis/driver/tmcm.py", line 2343, in _getCurrentPos
	pos = startp + (endp - startp) * (now - startt) / (endt - startt)
ZeroDivisionError: float division by zero

Running /home/testing/development/odemis/src/odemis/driver/test/andorcam2_test.py...
FAIL: test_synchronization_removal (__main__.TestSynchronized)
  File "/home/testing/development/odemis/src/odemis/driver/test/cam_test_abs.py", line 726, in test_synchronization_removal
	self.assertEqual(self.ccd_left, 0)
AssertionError: 1 != 0


Running /home/testing/development/odemis/src/odemis/driver/test/avantes_test.py...
FAIL: test_trigger (__main__.AvantesTest)
```

### DIFF
--- a/util/pytest_log_filter.py
+++ b/util/pytest_log_filter.py
@@ -1,0 +1,89 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+Created on 02 02 2022
+
+@author: Kornee Kleijwegt
+
+Copyright © 2022 Kornee Kleijwegt, Delmic
+
+This file is part of Odemis.
+
+Odemis is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License version 2 as published by the Free Software Foundation.
+
+Odemis is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with Odemis. If not, see http://www.gnu.org/licenses/.
+
+----------------------------------------------------------------------------------------------------
+
+This file allows to filter a pytest test report such that only the errors, failures etc. are outputted.
+This means that the part of the report regarding the passed tests and on the current progress are filtered out.
+
+Run the script by calling it with as first (and only) input argument the location of a text based file where the pytest test report can be found.
+    > ./pytest_log_filter.py ~/location/test_file.txt
+"""
+import re
+import sys
+
+
+def filter_test_log(log_txt):
+    """
+    Filters a log file to only include the parts with a failure warning etc.
+    Text regarding progress and passed tests are excluded.
+
+    :param log (str): String with the log of a single test file.
+    :return (str): filtered log
+
+    Example of filtering:
+    This function expects a pytest test report of single test file as input, a string such as the example below:
+
+        ============================= test session starts ==============================
+        platform linux -- Python 3.6.9, pytest-3.3.2, py-1.5.2, pluggy-0.6.0 -- /usr/bin/python3
+        cachedir: ../../../../.cache
+        rootdir: /home/kleijwegt/development/odemis, inifile:
+        collecting ... collected 9 items
+
+        comp_canvas_test.py::TestDblMicroscopeCanvas::test_basic_display PASSED  [ 11%]
+        comp_canvas_test.py::TestDblMicroscopeCanvas::test_basic_move PASSED     [ 22%]
+        comp_canvas_test.py::TestDblMicroscopeCanvas::test_conversion_functions PASSED [ 33%]
+        comp_canvas_test.py::TestDblMicroscopeCanvas::test_conversion_methods PASSED [ 44%]
+        comp_canvas_test.py::TestDblMicroscopeCanvas::test_crosshair PASSED      [ 55%]
+        comp_canvas_test.py::TestDblMicroscopeCanvas::test_pyramidal_3x2 FAILED  [ 66%]
+        comp_canvas_test.py::TestDblMicroscopeCanvas::test_pyramidal_one_tile FAILED [ 77%]
+        comp_canvas_test.py::TestDblMicroscopeCanvas::test_pyramidal_zoom FAILED [ 88%]
+        comp_canvas_test.py::TestDblMicroscopeCanvas::test_zoom_move PASSED      [100%]
+
+        =================================== FAILURES ===================================
+        /home/kleijwegt/development/odemis/src/odemis/gui/test/comp_canvas_test.py:742: AssertionError: Tuples differ: (0, 128, 127) != (0, 76, 179)
+        /home/kleijwegt/development/odemis/src/odemis/gui/test/comp_canvas_test.py:501: AssertionError: Tuples differ: (0, 128, 127) != (0, 76, 179)
+        /home/kleijwegt/development/odemis/src/odemis/gui/test/comp_canvas_test.py:604: AssertionError: Tuples differ: (0, 127, 128) != (0, 179, 76)
+        ====================== 3 failed, 6 passed in 7.43 seconds ======================
+
+    For a full test report only failures, errors etc. are of interest, therefore only this part is outputted, the rest is
+    filtered. An example of such an output string is the following example:
+
+        =================================== FAILURES ===================================
+        /home/kleijwegt/development/odemis/src/odemis/gui/test/comp_canvas_test.py:742: AssertionError: Tuples differ: (0, 128, 127) != (0, 76, 179)
+        /home/kleijwegt/development/odemis/src/odemis/gui/test/comp_canvas_test.py:501: AssertionError: Tuples differ: (0, 128, 127) != (0, 76, 179)
+        /home/kleijwegt/development/odemis/src/odemis/gui/test/comp_canvas_test.py:604: AssertionError: Tuples differ: (0, 127, 128) != (0, 179, 76)
+        ====================== 3 failed, 6 passed in 7.43 seconds ======================
+
+    """
+    test_results = re.search("\n\n===(.+?)((.|\n)*)==", log_txt).group()  # Filter to only get the usefull parts of the test report
+    test_results = test_results.lstrip("\n")  # Remove preceding empty lines
+    # Only display when there is more to tell than just passed test cases (meaning failures, warning etc.)
+    if "\n" in test_results:  # Only a message with multiple lines contains interesting information.
+        if len(test_results) > 3500:
+            test_results = test_results[0:3500] + \
+                           "\n \t--logging message is too long for the default summary report, it is stopped here -- \n\n"
+        return test_results
+
+
+if __name__ == "__main__":
+    # Run with as first argument the path to a log txt file of a single test file.
+    with open(sys.argv[1]) as f:
+        log_txt = f.read()
+    output = filter_test_log(log_txt)
+    if output:
+        print(output)
--- a/util/runtests.sh
+++ b/util/runtests.sh
@@ -1,4 +1,8 @@
 #!/bin/bash
+# Can be executed with the either of both options but not simultaneously both:
+#   --unit-test for the standard extended unit-testing report (default when no option is provided)
+#   --pytest for the shortened pytest summary report
+#
 # Should be run in the directory where the log files should be saved.
 # It should also have ../mic-odm-yaml/ directory, which contains all the microscopes files.
 # Ex:
@@ -10,6 +14,13 @@
 #  * unittest-full-$DATE.log : results of unit tests
 #  * integtest-full-$DATE.log: results of the integration tests
 #  * ./integtest-$DATE/ :logs of integration testing (2 files per conf)
+
+# Only when the first argument contains pytest the unittests are run using pytests and a pytest report is made.
+if [[ $1 = "--pytest" ]]; then
+  pytest=True
+else
+  pytest=False
+fi
 
 # Root path of the repo
 ODEMIS_DIR="$(readlink -m $(dirname $0)/../)"
@@ -58,9 +69,24 @@ export TEST_NO_SUPPORT_TICKET=1
 # and there are many other things that can affect execution order (eg, threads).
 export PYTHONHASHSEED=1567315
 
-TESTLOG=./unittest-full-$DATE.log
+mkdir -p ~/development/odemis-testing
+TESTLOG=~/development/odemis-testing/unittest-full-$DATE.log
 # make sure it is full path
 TESTLOG="$(readlink -m "$TESTLOG")"
+
+if [ $pytest = True ]; then
+  TESTSUMMARY=~/development/odemis-testing/pytest-summary-$DATE.log
+  # Remove any files which might already exist from a previously (expected incomplete) run of that day.
+  rm -f $TESTSUMMARY
+  touch $TESTSUMMARY
+  TESTSUMMARY="$(readlink -m "$TESTSUMMARY")"
+fi
+
+
+# The temporary files are used to save multi line strings (This had issues with bash)
+rm -f /tmp/filtered_test.txt /tmp/test_summary.txt
+touch /tmp/test_summary.txt /tmp/filtered_test.txt
+
 
 if [ ! -d /var/run/odemisd ] ; then
     echo  "Need /var/run/odemisd"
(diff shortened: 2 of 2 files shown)


util/pytest_log_filter.py
```diff
@@ -0,0 +1,16 @@
+import re
+import sys
+
+
+def filter_test_log(log):
```
tepals, line 5: can you add a docstring explaining what log is and what type? Is is the path to the log file?

util/runpytest.sh
```diff
@@ -0,0 +1,173 @@
+#!/bin/bash
+# Should be run in the directory where the log files should be saved.
+# It should also have ../mic-odm-yaml/ directory, which contains all the microscopes files.
+# Ex:
+# cd odemis-testing/
+# ../odemis/util/runtests.sh 2>&1 | tee test-$(date +%Y%m%d).log
```
tepals, line 6: shouldn't this now be runpytest.sh: `# ../odemis/util/runpytest.sh 2>&1 | tee test-$(date +%Y%m%d).log`?

util/runpytest.sh
```diff
@@ -0,0 +1,173 @@
...
+if [ ! -d /var/run/odemisd ] ; then
+    echo  "Need /var/run/odemisd"
+    sudo mkdir -m 777 /var/run/odemisd
+fi
+
+# stop the backend
+sudo odemis-stop
+
+# find the test scripts (should not contain spaces)
+testfiles="$(find "$ODEMIS_SRC" -wholename "*/test/*test.py")"
```
tepals, line 83: Line 30 reads `# Every file which is in the pattern /test/*_test.py`, so shouldn't this pattern be `"*/test/*_test.py"` instead of `"*/test/*test.py"`?

util/runpytest.sh
```diff
@@ -0,0 +1,173 @@
...
+
+#Warn if some files are misnamed
+skippedfiles="$(find "$ODEMIS_SRC" -wholename "*/test/*.py" -and -not -wholename "*/test/*test.py")"
```
tepals, line 86: same here: shouldn't this pattern be `"*/test/*_test.py"` instead of `"*/test/*test.py"`?

util/runpytest.sh
```diff
@@ -0,0 +1,173 @@
...
+        echo $f returned $status >> "$TESTLOG" 2>&1
+    popd > /dev/null
+    # Don't show test output if the file hasn't grown, as it'd be the previous test output
+    new_size=$(wc -l < "$TESTLOG")
+    if [[ "$new_size" == "$prev_size" ]]; then
+        echo "NOT RUN"
+    else
+        tail -n "+$prev_size" "$TESTLOG" > /tmp/latest_test_case_log.txt
+        # Filter the output to only print the summary
+        python3 ~/development/pytest/log_filter.py /tmp/latest_test_case_log.txt > /tmp/filtered_test.txt
```
tepals, line 123: Is `~/development/pytest/log_filter.py` the same as `~/development/odemis/util/pytest_log_filter.py`?

util/runpytest.sh
```diff
@@ -0,0 +1,173 @@
...
+echo "Running integration tests" | tee -a "$TESTLOG"
+python3 "$ODEMIS_DIR/util/run_intg_tests.py" --log-path "$INTEGLOGDIR" "$SIMPATH"/ >> "$TESTLOG" 2>&1
+ODMPATH="$ODEMIS_DIR/../mic-odm-yaml/" # extra microscope files
+if [ -d "$ODMPATH" ]; then
+    python3 "$ODEMIS_DIR/util/run_intg_tests.py" --log-path "$INTEGLOGDIR" "$ODMPATH"/*/ >> "$TESTLOG" 2>&1
+fi
+
+# TODO: run GUI standalone tests by trying to load every test data file that we have.
+
+kill %1 # Stops the "tail -f"
```
tepals, line 173: you forgot an empty line at the end of the file

tepals (changes requested):

util/runpytest.sh
Kleijwegt: I should still catch the same test cases. @pieleric Do you think it is safe to assume all test cases are named correctly and none forgot the "_"?

util/runpytest.sh
Kleijwegt: Nice catch. This only worked because I created a folder before adding it to odemis.

util/pytest_log_filter.py
```diff
@@ -0,0 +1,16 @@
+import re
```
pieleric, line 1: If it's a python file that can be executed, as it seems to be the case, please add a shebang line, and mark the file executable.

util/pytest_log_filter.py
```diff
@@ -0,0 +1,16 @@
...
+import sys
```
pieleric, line 2: Please add a comment explaining briefly what this program does and how to call it from the command line.

util/pytest_log_filter.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def filter_test_log(log):
+    with open(log) as f:
+        log_txt = f.read()
+    test_results = re.search("\n\n===(.+?)((.|\n)*)==", log_txt).group()  # Filter to only get the usefull parts of the info
```
pieleric, line 8: It's a good idea to put in a comment an example of input that the regex is supposed to catch. (Otherwise here, it's defined only as "usefull part", which is not very precise.

util/pytest_log_filter.py
```diff
@@ -0,0 +1,16 @@
...
+    test_results = test_results[test_results.find("="):]  # Remove preceding enters
```
pieleric, line 9: "enters" -> "empty lines".

util/pytest_log_filter.py
```diff
@@ -0,0 +1,16 @@
...
+    # Only display when there is more to tell than just passed test cases (meaning failures, warning etc.)
+    if test_results.count("\n") > 0:
```
pieleric, line 11: `.count()` is going to scan the entire string, which can be slow if the string is long. You could write `"\n" in test_results`, which will automatically stop the scan as soon as one `\n` is found.

util/runpytest.sh
pieleric: Don't hardcode the path. You can use `$ODEMIS_DIR/util/...`

util/runpytest.sh
pieleric: Use `_` everywhere.

pieleric (changes requested):
Please do not create a new `runpytest.sh` which provides a slightly better way to run testcases than `runtest.sh`. Just update `runtest.sh` with the new improved way.

util/runpytest.sh
Kleijwegt: @tepals you are right (I have discussed it with @pieleric ) I will also fix these comments

pieleric:
> I don't think this script actually replaces runtest.sh. The extended and detailed log of that script is still better. I don't think we need that very often but occasionally we might.

90% of the script is the same as runtest. Having two versions of the script would bring a lot of code duplication. I think it'd best if we only had one script to run all test cases. Is this new version not storing the whole log anymore? Possibly, you could adjust the script to store both the full log and the summary. Otherwise, you could pass a parameter, to select a different output format (eg `--summary` or `--shortlog`). When such parameter is passed to runtest.py, it would activate running the tests with your new code. With a very quicky look, I have a feeling that it should be relatively easy to put the test execution (and logging) in a bash function, and so based on the argument, you'd call a different function.

util/pytest_log_filter.py
Kleijwegt: Does marking it excecutable actually transfer via Git?

util/pytest_log_filter.py
```diff
@@ -0,0 +1,84 @@
...
+        /home/kleijwegt/development/odemis/src/odemis/gui/test/comp_canvas_test.py:742: AssertionError: Tuples differ: (0, 128, 127) != (0, 76, 179)
+        /home/kleijwegt/development/odemis/src/odemis/gui/test/comp_canvas_test.py:501: AssertionError: Tuples differ: (0, 128, 127) != (0, 76, 179)
+        /home/kleijwegt/development/odemis/src/odemis/gui/test/comp_canvas_test.py:604: AssertionError: Tuples differ: (0, 127, 128) != (0, 179, 76)
+        ====================== 3 failed, 6 passed in 7.43 seconds ======================
+
+    """
+    test_results = re.search("\n\n===(.+?)((.|\n)*)==", log_txt).group()  # Filter to only get the usefull parts of the test report
+    test_results = test_results[test_results.find("="):]  # Remove preceding empty lines
+    # Only display when there is more to tell than just passed test cases (meaning failures, warning etc.)
+    if "n" in test_results:
```
tepals, line 74: what do you check with `if "n" in test_results:`, why does there need to be an "n" in the test results for there to be a failure, warning, etc.?

util/runpytest.sh
```diff
@@ -0,0 +1,173 @@
...
+    sudo mkdir -m 777 /var/run/odemisd
+fi
+
+# stop the backend
+sudo odemis-stop
+
+# find the test scripts (should not contain spaces)
+testfiles="$(find "$ODEMIS_SRC" -wholename "*/test/*_test.py")"
+
+#Warn if some files are misnamed
```
tepals, line 85: Can you add a space between the # and the first word?

tepals (changes requested):

util/pytest_log_filter.py
Kleijwegt: This was supposed to be a "\n" I made a small error

tepals (changes requested):
two small comments

util/pytest_log_filter.py
```diff
@@ -0,0 +1,89 @@
...
+
+You should have received a copy of the GNU General Public License along with Odemis. If not, see http://www.gnu.org/licenses/.
+
+----------------------------------------------------------------------------------------------------
+
+This file allows to filter a pytest test report such that only the errors, failures etc. are outputted.
+This means that the part of the report regarding the passed tests and on the current progress are filtered out.
+
+Run the script by calling it with as first (and only) input argument the location of a text based file where the pytest test report can be found.
+    > ./pytest_log_filter.py ~/location/test_file.txt
```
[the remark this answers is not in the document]
Kleijwegt: That was still there for debugging in put the intended code back.

tepals, line 24: doesn't this need python3 in front of it? `python3 pytest_log_filter.py ~/location/test_file.txt`

util/runtests.sh
```diff
@@ -1,4 +1,8 @@
 #!/bin/bash
+# Can be executed with the options:
+#   --unit-test for the standard extended unit-testing report (default when no option is provided)
+#   --pytest for the shortened pytest summary report
```
tepals, line 4: Can you add a comment stating whether it is it also possible to run it with both options, or not.

tepals (changes requested):

util/pytest_log_filter.py
Kleijwegt: Because the shebang is written to execute via python this is not necessary.

util/runtests.sh
Kleijwegt: Done

tepals (approved):

util/runtests.sh
```diff
@@ -11,6 +15,13 @@
 #  * integtest-full-$DATE.log: results of the integration tests
 #  * ./integtest-$DATE/ :logs of integration testing (2 files per conf)
 
+# Only when the first argument contains pytest the unittests are run using pytests and a pytest report is made.
+if [[ $1 = *pytest* ]]; then
```
pieleric, line 19: Let's just write explicitly `if [[ $1 = "--pytest" ]];`
It's clearer what it does.

util/runtests.sh
```diff
@@ -58,10 +69,25 @@ export TEST_NO_SUPPORT_TICKET=1
...
-TESTLOG=./unittest-full-$DATE.log
+mkdir -p ~/development/odemis-testing
+TESTLOG=~/development/odemis-testing/unittest-full-$DATE.log
 # make sure it is full path
 TESTLOG="$(readlink -m "$TESTLOG")"
 
+if [ $pytest = True ]; then
+  TESTSUMMARY=~/development/odemis-testing/pytest-summary-$DATE.log
+  # Remove any files which might already exist from a previously (expected incomplete) run of that day.
+  rm $TESTSUMMARY
```
pieleric, line 80: `rm -f $TESTSUMMARY`
To avoid error message that the file doesn't exist.

util/runtests.sh
```diff
@@ -109,10 +140,18 @@ for f in $testfiles; do
...
         echo "NOT RUN"
     else
-        tail -n "+$prev_size" "$TESTLOG" | grep -E 'OK' | tail -1
-        tail -n "+$prev_size" "$TESTLOG" | awk "/^FAIL: /,/FAILED/"
-        tail -n "+$prev_size" "$TESTLOG" | awk "/^ERROR: /,/FAILED/"
-        #tail -n "+$prev_size" "$TESTLOG" | awk '/===/, /FAILED/'
+        if [ $pytest = True ]; then
+          tail -n "+$prev_size" "$TESTLOG" > /tmp/latest_test_case_log.txt
+          # Filter the output to only print the summary
+          python3 ~/development/odemis/util/pytest_log_filter.py /tmp/latest_test_case_log.txt > /tmp/filtered_test.txt
```
pieleric, line 146: Please use `$ODEMIS_DIR` instead of `~/development/odemis/`, this avoids hard-coding paths.

util/runtests.sh
```diff
@@ -58,10 +69,25 @@ export TEST_NO_SUPPORT_TICKET=1
...
+  TESTSUMMARY=~/development/odemis-testing/pytest-summary-$DATE.log
+  # Remove any files which might already exist from a previously (expected incomplete) run of that day.
+  rm $TESTSUMMARY
+  touch $TESTSUMMARY
+  TESTSUMMARY="$(readlink -m "$TESTSUMMARY")"
+fi
+
+
+# The temporary files are used to save multi line strings (This had issues with bash)
+rm /tmp/filtered_test.txt /tmp/test_summary.txt
```
pieleric, line 87: `rm -f`.

util/pytest_log_filter.py
```diff
@@ -0,0 +1,89 @@
...
+        ====================== 3 failed, 6 passed in 7.43 seconds ======================
+
+    """
+    test_results = re.search("\n\n===(.+?)((.|\n)*)==", log_txt).group()  # Filter to only get the usefull parts of the test report
+    test_results = test_results[test_results.find("="):]  # Remove preceding empty lines
+    # Only display when there is more to tell than just passed test cases (meaning failures, warning etc.)
+    if "\n" in test_results:  # Only a message with multiple lines contains interesting information.
+        if len(test_results) > 3500:
+            test_results = test_results[0:3500] + \
+                           "\n \t--logging message is to long for the default summary report, it is stopped here -- \n\n"
```
pieleric, line 79: "to long" -> "too long".

util/pytest_log_filter.py
```diff
@@ -0,0 +1,89 @@
...
+
+        =================================== FAILURES ===================================
+        /home/kleijwegt/development/odemis/src/odemis/gui/test/comp_canvas_test.py:742: AssertionError: Tuples differ: (0, 128, 127) != (0, 76, 179)
+        /home/kleijwegt/development/odemis/src/odemis/gui/test/comp_canvas_test.py:501: AssertionError: Tuples differ: (0, 128, 127) != (0, 76, 179)
+        /home/kleijwegt/development/odemis/src/odemis/gui/test/comp_canvas_test.py:604: AssertionError: Tuples differ: (0, 127, 128) != (0, 179, 76)
+        ====================== 3 failed, 6 passed in 7.43 seconds ======================
+
+    """
+    test_results = re.search("\n\n===(.+?)((.|\n)*)==", log_txt).group()  # Filter to only get the usefull parts of the test report
+    test_results = test_results[test_results.find("="):]  # Remove preceding empty lines
```
pieleric, line 74: If you just want to remove empty lines, then you can write simply:
test_results = test_results.lstrip("\n")

pieleric (changes requested):

Kleijwegt:
Resolved all comments, ready for review but need to be tested by running all the test cases before merging.

pieleric (approved):

PR #2025, opened 2022-02-09
fastem acquisition manager raised a timeout error for dwell times lar…

tepals:
…ger than 10us

The error was: odemis.util.TimeoutError: Timeout while waiting for field image.
After investigation I found out this only happened for the first image. Because the first image is actually acquired twice, the timeout for the first image needs to be increased.

### DIFF
--- a/src/odemis/acq/fastem.py
+++ b/src/odemis/acq/fastem.py
@@ -358,7 +358,11 @@ class AcquisitionTask(object):
         """
 
         total_field_time = self._detector.frameDuration.value
-        timeout = total_field_time + 5  # TODO what margin should be used?
+        # For a dwell time of 10us the total field time is approximately 7 seconds,
+        # therefore 7 seconds is a save margin for the timeout.
+        timeout = total_field_time + 7
+        # The first field gets acquired twice, therefore the field time is multiplied by 2.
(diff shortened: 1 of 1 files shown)


Sabrina85 (approved):

src/odemis/acq/fastem.py
```diff
@@ -358,7 +358,11 @@ def acquire_roa(self, dataflow):
         """
 
         total_field_time = self._detector.frameDuration.value
-        timeout = total_field_time + 5  # TODO what margin should be used?
+        # For a dwell time of 10us the total field time is approximately 7 seconds,
+        # therefore 7 seconds is a save margin for the timeout.
+        timeout = total_field_time + 7
+        # The first field gets acquired twice, therefore the field time is multiplied by 2.
+        timeout_first_field = 2 * total_field_time + 7
```
pieleric, line 365: IMHO, a timeout should be only there to prevent waiting forever. It's hard to know how long the hardware will actually take. What do you gain in timing out 7 second after the expected duration? You're saving a few seconds to tell the user the acquisition failed? I'd suggest to give a much larger margin. Something like `5 * total_field_time + 2`. You then don't need to "leak" here the information that the first field is special.

pieleric (changes requested):

PR #2027, opened 2022-02-10
fastem calibrations: implement calibration manager

Sabrina85:


src/odemis/acq/align/fastem.py
```diff
@@ -18,22 +20,282 @@
...
+
+                # In case the calibrations was cancelled by a client, before the future returned,
+                # raise cancellation error.
+                if self._cancelled:
+                    raise CancelledError()
+
+                # Update the time left for the calibrations remaining
+                self._future.set_progress(end=time.time() + self.estimate_calibration_time())
+
+        except CancelledError:  # raised in try statement
```
tepals, line 211: This comment `# raised in try statement` confuses me, if you have an except, it's always because the error was raised in the try statement, right? This comment makes it sound like this is a special case, but I do not see why it would be a special case.

src/odemis/acq/align/fastem.py
```diff
@@ -18,22 +20,282 @@
...
+    def cancel(self, future):
+        """
+        Cancels the calibrations.
+        :param future: (future) The calibration future.
+        :return: (bool) True if cancelled, TODO False?
+        """
+        self._cancelled = True
+
+        # FIXME Currently there is no subfuture implemented for each calibration. So currently,
+        #  when cancelling the calibration is still finished.
```
tepals, line 270: Can you clarify this sentence a bit more:

`So currently, when cancelling if a calibration has already started it will run until it is completely finished.`

src/odemis/acq/align/fastem.py
```diff
@@ -18,22 +20,282 @@
...
+            if calib == DESCAN_GAIN_STATIC:
+                tot_time += descan_gain.estimate_calibration_time()
+            if calib == IMAGE_ROTATION_PREALIGN:
+                tot_time += image_rotation_pre_align.estimate_calibration_time()
+            if calib == IMAGE_TRANSLATION_PREALIGN:
+                tot_time += image_translation_pre_align.estimate_calibration_time()
+            if calib == IMAGE_ROTATION_FINAL:
+                tot_time += image_rotation.estimate_calibration_time()
+            if calib == IMAGE_TRANSLATION_FINAL:
+                tot_time += image_translation.estimate_calibration_time()
```
tepals, line 299: I think this function could just call `estimate_calibration_time`:
```
tot_time = estimate_calibration_time(self._calibrations_remaining)
```

src/odemis/acq/align/test/fastem_test.py
```diff
@@ -0,0 +1,185 @@
...
+        cls.descanner = model.getComponent(role="descanner")
+        cls.stage = model.getComponent(
+            role="stage")  # TODO replace with stage-scan when ROA conversion method available
+        cls.ccd = model.getComponent(role="diagnostic-ccd")
+        cls.focuser = model.getComponent(role="diagnostic-cam-focus")
+        cls.beamshift = model.getComponent(role="ebeam-shift")
+        cls.det_rotator = model.getComponent(role="det-rotator")
+
+    def setUp(self):
+
```
tepals, line 73: this empty line can be removed

src/odemis/acq/align/test/fastem_test.py
```diff
@@ -0,0 +1,185 @@
...
+        self.good_focus = -70e-6  # position where the image of the multiprobe is displayed in focus [m]
+        self.good_mp_position = (650, 450)  # position for multiprobe where mp is aligned with mppc detector [px_dc]
```
tepals, line 75: I would say:
`# position of multiprobe on diagnostic camera where mp is aligned with mppc detector [px_dc]`

src/odemis/acq/align/test/fastem_test.py
```diff
@@ -0,0 +1,185 @@
...
+        self.ccd.updateMetadata(
+            {model.MD_FAV_POS_ACTIVE:
+                {"x": self.good_mp_position[0],
+                 "y": self.good_mp_position[1],
+                 "z": self.good_focus}}
+                                )
```
tepals, line 81: Since you're not using `self.good_mp_position` anywhere, if you don't want to explicitly set it and just use the values from the yaml file, you could update these lines to:
```
self.ccd.updateMetadata(model.MD_FAV_POS_ACTIVE.update({"z": self.good_focus})
```

src/odemis/acq/align/fastem.py
```diff
@@ -18,22 +20,282 @@
...
+            descan_gain.run_descan_gain_static(self._scanner, self._multibeam, self._descanner,
+                                               self._detector, self._dataflow, self._ccd)
+
+        if calibration == IMAGE_ROTATION_PREALIGN:
+            image_rotation_pre_align.run_image_rotation_pre_align(self._scanner, self._multibeam, self._descanner,
+                                                                  self._detector, self._dataflow, self._ccd,
+                                                                  self._det_rotator)
+
+        if calibration == IMAGE_TRANSLATION_PREALIGN:
+            image_translation_pre_align.run_image_translation_pre_align(self._scanner, self._multibeam,
```
tepals, line 249: this calibration (and I think a few others) returns the calibrated value, which then still needs to be set on the hardware. How do you deal with that?

src/odemis/acq/align/test/fastem_test.py
```diff
@@ -0,0 +1,185 @@
...
+
+        e = f.result(timeout=900)
+
+        self.assertIsNone(e)  # check no exceptions were returned
+
+        # get the calibrated descanner offset
+        descan_offset_calib = self.descanner.scanOffset.value
+
+        # check the calibrated descan offset is different from the previous offset
+        self.assertNotEqual(descan_offset_cur, descan_offset_calib)
```
tepals, line 128: If you run it on hardware the descanner offset should actually change, so will this test case then fail?

src/odemis/acq/align/fastem.py
```diff
@@ -18,22 +20,282 @@
...
+    :param stage: (actuator.ConvertStage) The stage in the corrected scan coordinate system, the x and y axes are
+        aligned with the x and y axes of the multiprobe and the multibeam scanner.
+    :param ccd: (model.DigitalCamera) A camera object of the diagnostic camera.
+    :param beamshift: (tfsbc.BeamShiftController) Component that controls the beamshift deflection.
+    :param det_rotator: (tmcm.CANController) K-mirror controller.
+    :param calibrations: (list of str) List of calibrations that should be run.
+
+    :returns: (ProgressiveFuture) Alignment future object, which can be cancelled.
+            The result of the future is:
+                (Exception or None): Exception raised during the acquisition or None.
```
tepals, line 75: I think this is only the case if the run function returns the exceptions, but your run function only raises exceptions. If you look at `odemis.acq.fastem.AcquisitionTask.run` general exceptions are caught and returned, only the canceled error is raised.

src/odemis/acq/align/test/fastem_test.py
```diff
@@ -0,0 +1,185 @@
...
+
+        f.add_update_callback(self.on_progress_update)  # callback executed every time f.set_progress is called
+        f.add_done_callback(self.on_done)  # callback executed when f.set_result is called (via bindFuture)
+
+        e = f.result()
+
+        self.assertIsNone(e)  # check no exceptions were returned
+        self.assertTrue(self.done)
+        # at least one update per calibration plus once at start of calibration, plus once at end of calibration
+        self.assertGreaterEqual(self.updates, 4)  # at least one update per
```
tepals, line 147: you forgot to finish the sentence:  `# at least one update per...`

tepals, line 147: why use the `assertGreaterEqual` and not `assertEqual`, can't you be sure that there are exactly 4 updates?

src/odemis/acq/align/test/fastem_test.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def test_cancel(self):
+        """Test if it is possible to cancel the optical autofocus calibration."""
+
+        # FIXME no subfuture available yet, which are cancelable
```
tepals, line 152: so does this test always fail?

tepals (changes requested):

src/odemis/acq/align/fastem.py
```diff
@@ -18,22 +20,282 @@
...
 """
+import logging
+import time
+from concurrent.futures import CancelledError
 
 from odemis import model
-import time
 
+try:
+    import fastem_calibrations
```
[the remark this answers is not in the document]
Sabrina85: Not sure. For some reason it did not work without. I will check again.

src/odemis/acq/align/fastem.py
```diff
@@ -18,22 +20,282 @@
...
+        try:
+            logging.debug("Starting calibration.")
+
+            # loop over calibrations in list (order in list is important!)
+            for calib in self.calibrations:
+                # TODO return a sub-future when implemented for calibrations
+                self.run_calibrations(calib)
+
+                # def _pass_future_progress(sub_f, start, end):
+                #     f.set_progress(start, end)
```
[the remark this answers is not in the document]
Sabrina85: We do not handle sub-futures yet. I copied this code from the overview image acquisition.

src/odemis/acq/align/fastem.py
Sabrina85: True, in the acquisition it was more hidden as it was raised in acquire_roa, which itself was called in the try statement. I will remove it.

src/odemis/acq/align/fastem.py
Sabrina85: Didn't think about this yet. Will do. I suggest similar as in the `run_calibrations.py` module?

src/odemis/acq/align/fastem.py
Sabrina85: True

src/odemis/acq/align/test/fastem_test.py
Sabrina85: No I cannot. The problem with these set_progress calls is that it is not only called in the calibration manager explicitly, but also when the future starts and ends I found out. So even if I never call set_progress here in the manager, it is still called. In other test cases I only find assertGreaterEqual so I would go for this to avoid failing tests. As long as progress is found we are happy I think.

src/odemis/acq/align/fastem.py
tepals: sounds good

src/odemis/acq/align/test/fastem_test.py
tepals: aha, I didn't read the "not" I read it as `self.assertEqual`.

src/odemis/acq/align/fastem.py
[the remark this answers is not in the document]
Sabrina85: I checked. If I don't do it, then the GUI crashes when I check whether fastem_calibrations is not None. Which makes sense. I do now the same as you do in the acquisition manager with True/False.

src/odemis/acq/align/fastem.py
```diff
@@ -18,22 +20,282 @@
...
+                    raise CancelledError()
+
+                # Update the time left for the calibrations remaining
+                self._future.set_progress(end=time.time() + self.estimate_calibration_time())
+
+        except CancelledError:  # raised in try statement
+            logging.debug("Calibration was cancelled.")
+            raise
+        except Exception as ex:
+            raise ex
```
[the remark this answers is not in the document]
Sabrina85: True. What would you suggest?

src/odemis/acq/align/fastem.py
```diff
@@ -18,22 +20,282 @@
...
+        self._cancelled = False
+
+        if model.MD_CALIB not in self._beamshift.getMetadata().keys():
+            self._beamshift.updateMetadata({model.MD_CALIB: self._scanner.beamShiftTransformationMatrix.value})
+
+    def run(self):
+        """
+        Runs a set of calibration procedures.
+        :returns:
+            (None) If calibrations successful return None.
```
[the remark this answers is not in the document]
Sabrina85: yes true, but I also like that it says None = successful. I shortened it and hope that avoids confusion. Let me know if still confusing.

src/odemis/acq/align/test/fastem_test.py
Sabrina85: True. Maybe in future we can test the image translation pre-alignment in a fancier way, where we want to use explicit values. I removed it for now.

src/odemis/acq/align/fastem.py
Sabrina85: Let's discuss the details for this when I am back. It really depends also a bit if we want to allow to run subsets of the calibrations. Maybe that part is also easier when you had looked into the additional MD for the calibrated values next week.

src/odemis/acq/align/fastem.py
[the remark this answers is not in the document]
tepals: I think you can remove these two lines, or log something before raising the exception

src/odemis/acq/align/fastem.py
```diff
@@ -18,22 +20,245 @@
...
+        :param descanner: (technolution.MirrorDescanner) The mirror descanner component of the acquisition server module.
+        :param detector: (technolution.MPPC) The detector object to be used for collecting the image data.
+        :param stage: (actuator.ConvertStage) The stage in the corrected scan coordinate system, the x and y axes are
+            aligned with the x and y axes of the multiprobe and the multibeam scanner.
+        :param ccd: (model.DigitalCamera) A camera object of the diagnostic camera.
+        :param beamshift: (tfsbc.BeamShiftController) Component that controls the beamshift deflection.
+        :param det_rotator: (tmcm.CANController) K-mirror controller.
+        :param calibrations: (list of str) List of calibrations that should be run.
+
+        :param future: (ProgressiveFuture) Acquisition future object, which can be cancelled.
```
tepals, line 127: In the input arguments, on line 114, you have the future first and then the calibrations. The order in the docstring should match the order of the input arguments. I think first the future then the calibrations makes sense, although I do not have a strong preference.

src/odemis/acq/align/fastem.py
```diff
@@ -18,22 +20,245 @@
...
+        self.calibrations = calibrations
+
+        # List of calibrations to be executed. Used for progress update.
+        self._calibrations_remaining = set(calibrations)
+
+        # keep track if future was cancelled or not
+        self._cancelled = False
+
+        if model.MD_CALIB not in self._beamshift.getMetadata().keys():
+            self._beamshift.updateMetadata({model.MD_CALIB: self._scanner.beamShiftTransformationMatrix.value})
```
tepals, line 150: Can you add a comment why this is necessary here, or will this be removed in another PR anyway?

Also just fyi, the `.keys()` is not necessary `if model.MD_CALIB not in self._beamshift.getMetadata():` also works, but use whatever you feel is more readable :)

src/odemis/acq/align/fastem.py
```diff
@@ -18,22 +20,245 @@
...
+        finally:
+            # Remove references to the calibrations once all calibrations are finished/cancelled.
+            self._calibrations_remaining.clear()
+            self._scanner.blanker.value = True  # always blank the beam to reduce beam damage on sample
+            logging.debug("Finish calibrations.")
+
+    def run_calibrations(self, calibration):
+        """
+        Run a calibration.
+        # Note: All calibrations can be run on bare scintillator.
```
tepals, line 209: The `#` is not necessary

src/odemis/acq/align/test/fastem_test.py
```diff
@@ -0,0 +1,180 @@
...
+
+        # move the stage so that the image is out of focus
+        self.center_position = -30e-6
+        self.focuser.moveAbs({"z": self.center_position}).result()
+
+        # Run auto focus
+        f = fastem.align(self.scanner, self.multibeam, self.descanner, self.mppc, self.stage, self.ccd,
+                         self.beamshift, self.det_rotator, calibrations)
+
+        # foc_pos, foc_lev = f.result(timeout=900)  # TODO make use of the focus level?
```
tepals, line 93: you could add an assert that the focus level is larger than 1e10, at least for the hardware it should be something like that, I don't know what to expect for the simulator.

src/odemis/acq/align/fastem.py
Sabrina85: Removed now as we have a separate pr that loads this MD at init of the beamshift component. #pr2031

src/odemis/acq/align/test/fastem_test.py
Sabrina85: I just noticed, that we do not return the focus level and focus position yet. The future only returns exceptions in the calibration manager and nothing else. I don't know if we can extend the calibration manager to return such values (and only for the autofocus then). We do return DataArrays for the acquisition. But those are captured via a threading event I think.

Sabrina85: I think for now let's not do this?

src/odemis/acq/align/fastem.py
```diff
@@ -18,22 +20,254 @@
...
-    time.sleep(2)
+def estimate_calibration_time(calibrations):
+    """
+    Computes the approximate time it will take to run all calibrations.
+    :param calibrations: (list of str) List of calibrations that should be run.
+    :return (0 <= float): The estimated time for the requested calibrations in s.
+    """
+    tot_time = 0
+    for calib in calibrations:
+        tot_time += calib.estimate_calibration_time()
```
pieleric, line 105: This can be written as `return sum(c.estimate_calibration_time() for c in calibrations)`. Though, this only works if the calibrations are functions. Otherwise, you'll need to use a dict.

Maybe I've done too much code golfing? Actually, I like using `sum()`.

src/odemis/acq/align/fastem.py
```diff
@@ -18,22 +20,254 @@
...
+    :param scanner: (xt_client.Scanner) Scanner component connecting to the XT adapter.
+    :param multibeam: (technolution.EBeamScanner) The multibeam scanner component of the acquisition server module.
+    :param descanner: (technolution.MirrorDescanner) The mirror descanner component of the acquisition server module.
+    :param detector: (technolution.MPPC) The detector object to be used for collecting the image data.
+    :param stage: (actuator.ConvertStage) The stage in the corrected scan coordinate system, the x and y axes are
+        aligned with the x and y axes of the multiprobe and the multibeam scanner.
+    :param ccd: (model.DigitalCamera) A camera object of the diagnostic camera.
+    :param beamshift: (tfsbc.BeamShiftController) Component that controls the beamshift deflection.
+    :param det_rotator: (tmcm.CANController) K-mirror controller.
+    :param calibrations: (list of str) List of calibrations that should be run.
```
pieleric, line 74: not a `list of str`.

src/odemis/acq/align/fastem.py
```diff
@@ -18,22 +20,254 @@
...
+
+    :returns: (ProgressiveFuture) Alignment future object, which can be cancelled.
+            The result of the future is: (None)
```
pieleric, line 77: The normal way of calling this in Python is that the future doesn't return any value.

src/odemis/acq/align/fastem.py
```diff
@@ -18,22 +20,254 @@
...
+
     return f
 
 
-def _run_fake_alignment():
-    time.sleep(2)
+def estimate_calibration_time(calibrations):
+    """
+    Computes the approximate time it will take to run all calibrations.
+    :param calibrations: (list of str) List of calibrations that should be run.
```
pieleric, line 100: Not a `list of str`...

src/odemis/acq/align/fastem.py
```diff
@@ -18,22 +20,254 @@
...
+        """
+        :param scanner: (xt_client.Scanner) Scanner component connecting to the XT adapter.
+        :param multibeam: (technolution.EBeamScanner) The multibeam scanner component of the acquisition server module.
+        :param descanner: (technolution.MirrorDescanner) The mirror descanner component of the acquisition server module.
+        :param detector: (technolution.MPPC) The detector object to be used for collecting the image data.
+        :param stage: (actuator.ConvertStage) The stage in the corrected scan coordinate system, the x and y axes are
+            aligned with the x and y axes of the multiprobe and the multibeam scanner.
+        :param ccd: (model.DigitalCamera) A camera object of the diagnostic camera.
+        :param beamshift: (tfsbc.BeamShiftController) Component that controls the beamshift deflection.
+        :param det_rotator: (tmcm.CANController) K-mirror controller.
```
pieleric, line 126: I'd suggest to not be that specific on the types. Just indicate they are Actuators, and which axes must be available. This way, it's clearer that we are not relying on anything special of this driver.

src/odemis/acq/align/fastem.py
```diff
@@ -18,22 +20,254 @@
...
+        :param future: (ProgressiveFuture) Acquisition future object, which can be cancelled.
```
pieleric, line 127: I'd suggest to pass the future as first argument. This helps to see it's a special object that handles a Future.

src/odemis/acq/align/fastem.py
```diff
@@ -18,22 +20,254 @@
...
+        total_calibration_time = self.estimate_calibration_time()
+
+        # No need to set the start time of the future: it's automatically done when setting its state to running.
+        self._future.set_progress(end=time.time() + total_calibration_time)  # provide end time to future
+        logging.info("Starting calibrations, with expected duration of %f s", total_calibration_time)
+
+        try:
+            logging.debug("Starting calibration.")
+
+            logging.debug("Read initial Hw settings.")
```
pieleric, line 173: Two logs for the same moment, that's a little redundant. => Pick one of them.

src/odemis/acq/align/fastem.py
```diff
@@ -18,22 +20,254 @@
...
     """
+
+    if not fastem_calibrations:
+        raise ModuleNotFoundError("fastem_calibration module missing. Cannot run calibrations.")
+
     f = model.ProgressiveFuture()
-    _executor.submitf(f, _run_fake_alignment)
+
+    # Create a task that runs the calibration and alignments.
+    task = CalibrationTask(scanner, multibeam, descanner, detector, stage, ccd, beamshift, det_rotator, f, calibrations)
```
pieleric, line 86: You should estimate the duration and store it in the future either here, or in the init of the CalibrationTask. This way, even if the task is not yet executed (eg, executor is a little slow, or doing something else) it's possible for the GUI to know how long this task will take.

src/odemis/acq/align/fastem.py
```diff
@@ -18,22 +20,254 @@
...
+        self.calibrations = calibrations
+
+        # List of calibrations to be executed. Used for progress update.
+        self._calibrations_remaining = set(calibrations)
+
+        # keep track if future was cancelled or not
+        self._cancelled = False
+
+        # reset beamshift
+        self._beamshift.shift.value = (0, 0)
```
pieleric, line 151: Move this to the `run()`. There might be some task running before this one. So you cannot modify the hardware yet here.

src/odemis/acq/align/fastem.py
```diff
@@ -18,22 +20,254 @@
 You should have received a copy of the GNU General Public License along with
 Odemis. If not, see http://www.gnu.org/licenses/.
 """
+import logging
+import time
+from concurrent.futures import CancelledError
 
+from fastem_calibrations.configure_hw import get_config_asm, configure_asm
```
pieleric, line 27: This will fail if the fastem_calibrations module is not present, right? => Move it to the try/except too.

src/odemis/acq/align/fastem.py
```diff
@@ -18,22 +20,254 @@
...
+                                               self._dataflow, self._ccd, self._stage)
+
+        if calibration == SCAN_ROTATION_PREALIGN:
+            scan_rotation_pre_align.run_scan_rotation_pre_align(self._scanner, self._multibeam, self._descanner,
+                                                                self._detector, self._dataflow, self._ccd)
+
+        if calibration == SCAN_AMPLITUDE_PREALIGN:
+            self.asm_config["multibeam"]["scanOffset"], self.asm_config["multibeam"]["scanAmplitude"] = \
+                scan_amplitude_pre_align.run_scan_amplitude_pre_align(self._scanner, self._multibeam, self._descanner,
+                                                                      self._detector, self._dataflow, self._ccd)
```
pieleric, line 228: Please make it 3 lines, so that if an error happens, it's clearer what's the issue. Something like:
```
soff, samp = run_scan_amplitude_pre_align(...)
self.asm_config["multibeam"]["scanOffset"] = soff
self.asm_config["multibeam"]["scanAmplitude"] = samp
```

src/odemis/acq/align/fastem.py
```diff
@@ -18,22 +20,254 @@
...
+    def run_calibrations(self, calibration):
+        """
+        Run a calibration.
+        Note: All calibrations can be run on bare scintillator.
+        """
+        if calibration == OPTICAL_AUTOFOCUS:
+            autofocus_multiprobe.run_autofocus(self._scanner, self._multibeam, self._descanner, self._detector,
+                                               self._dataflow, self._ccd, self._stage)
+
+        if calibration == SCAN_ROTATION_PREALIGN:
```
pieleric, line 221: Could you put `elif` everywhere, to clarify only one of them will ever be executed per call.

src/odemis/acq/align/test/fastem_test.py
```diff
@@ -0,0 +1,178 @@
...
+
+import odemis
+from odemis import model
+from odemis.acq.align import fastem
+from odemis.acq.align.fastem import OPTICAL_AUTOFOCUS, IMAGE_TRANSLATION_PREALIGN
+from odemis.util import test
+
+# * TEST_NOHW = 1: use simulator (asm/sam and xt adapter simulators need to be running)
+# * TEST_NOHW = 0: connected to the real hardware (backend needs to be running)
+# technolution_asm_simulator/simulator2/run_the_simulator.py
```
pieleric, line 41: Let's move this line next to TEST_NOHW=1.

src/odemis/acq/align/test/fastem_test.py
```diff
@@ -0,0 +1,178 @@
...
+                         self.beamshift, self.det_rotator, calibrations)
+
+        f.add_update_callback(self.on_progress_update)  # callback executed every time f.set_progress is called
+        f.add_done_callback(self.on_done)  # callback executed when f.set_result is called (via bindFuture)
+
+        time.sleep(1)  # make sure it's started
+        self.assertTrue(f.running())
+        f.cancel()
+
+        self.assertRaises(CancelledError, f.result, 1)  # add timeout = 1s in case cancellation error was not raised
```
pieleric, line 162: Please use `with self.assertRaises()`.

Also, I don't get why you're using such a short timeout, especially right now the code actually doesn't cancel the current calibration, so you'd have to wait for the autofocus to end. It doesn't always take < 1s.

src/odemis/acq/align/test/fastem_test.py
```diff
@@ -0,0 +1,178 @@
...
+        """Test if it is possible to cancel the optical autofocus calibration."""
+
+        # FIXME no subfuture available yet, which are cancelable
+        #  when subfutures are implemented, add a check in this test case that the subfuture was also cancelled
+
+        self.end = None  # updated in callback on_progress_update
+        self.updates = 0  # updated in callback on_progress_update
+        self.done = False  # updated in callback on_done
+
+        calibrations = [OPTICAL_AUTOFOCUS]
```
pieleric, line 151: Try with more calibrations! Anyway, only the first one should be started anyway.

src/odemis/acq/align/fastem.py
```diff
@@ -18,22 +20,254 @@
...
+
+# TODO does it make sense to make this a list?
+OPTICAL_AUTOFOCUS = autofocus_multiprobe
+SCAN_ROTATION_PREALIGN = scan_rotation_pre_align
+SCAN_AMPLITUDE_PREALIGN = scan_amplitude_pre_align
+DESCAN_GAIN_STATIC = descan_gain
+IMAGE_ROTATION_PREALIGN = image_rotation_pre_align
+IMAGE_TRANSLATION_PREALIGN = image_translation_pre_align
+IMAGE_ROTATION_FINAL = image_rotation
+IMAGE_TRANSLATION_FINAL = image_translation
```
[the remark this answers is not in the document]
pieleric: Note that if you're using an Enum, you don't need to start the names with a common word (as they will be part of the Enum namespace).

src/odemis/acq/align/test/fastem_test.py
Sabrina85: We can only try with 2, as all the others rely on HW. I will add the second one.

src/odemis/acq/align/fastem.py
Sabrina85: Then we should also change this in the acquisition manager right?

src/odemis/acq/align/fastem.py
```diff
@@ -39,25 +39,43 @@
...
-IMAGE_ROTATION_PREALIGN = image_rotation_pre_align
-IMAGE_TRANSLATION_PREALIGN = image_translation_pre_align
-IMAGE_ROTATION_FINAL = image_rotation
-IMAGE_TRANSLATION_FINAL = image_translation
-
 # The executor is a single object, independent of how many times the module (fastem.py) is loaded.
 _executor = model.CancellableThreadPoolExecutor(max_workers=1)
 
 
+class Calibrations(Enum):
```
Sabrina85, line 55: @pieleric Is there a better way to pass None when the module is not available?

pieleric: I think that you could set explicitly each variable that you import as `None` in the except handler. eg `autofocus_multiprobe = None` ... 
The alternative would be to use the Enum as a standard set of constants (eg, 1, 2, 3...), possibly using `auto()` and use a dict to match constant -> function.

src/odemis/acq/align/fastem.py
```diff
@@ -18,22 +20,289 @@
...
 
 
-def align(main_data):
+class Calibrations(Enum):
     """
-    :param main_data (odemis.gui.model.FastEMMainGUIData):
-    :returns: (ProgressiveFuture): acquisition future
+    Connect each calibration to a unique constant name.
+    """
+    if fastem_calibrations == True:
```
pieleric, line 59: Please just write `if fastem_calibrations`, which is more pythonic.

pieleric (changes requested):

src/odemis/acq/align/fastem.py
Sabrina85: I will go for the first one. 
For the second one: Where would the dict go?

src/odemis/acq/align/fastem.py
```diff
@@ -18,22 +20,288 @@
...
+
     return f
 
 
-def _run_fake_alignment():
-    time.sleep(2)
+def estimate_calibration_time(calibrations):
+    """
+    Computes the approximate time it will take to run all calibrations.
+    :param calibrations: (list) List of calibrations that should be run.
```
tepals, line 117: Shouldn't this now be a Calibrations Enum object instead of a list?

src/odemis/acq/align/fastem.py
```diff
@@ -18,22 +20,288 @@
...
+        :param multibeam: (technolution.EBeamScanner) The multibeam scanner component of the acquisition server module.
+        :param descanner: (technolution.MirrorDescanner) The mirror descanner component of the acquisition server
+            module.
+        :param detector: (technolution.MPPC) The detector object to be used for collecting the image data.
+        :param stage: (actuator) The stage in the corrected scan coordinate system, the x and y axes are
+            aligned with the x and y axes of the multiprobe and the multibeam scanner. Must have x, y and z axes.
+        :param ccd: (model.DigitalCamera) A camera object of the diagnostic camera.
+        :param beamshift: (tfsbc.BeamShiftController) Component that controls the beamshift deflection.
+        :param det_rotator: (actuator) K-mirror controller. Must have a rotational (rz) axis.
+        :param calibrations: (list) List of calibrations that should be run.
```
tepals, line 144: Same here: Shouldn't this now be a Calibrations Enum object instead of a list?

Or is it a list of enumeration members from the Calibrations object? (I think that's the correct nomenclature, see: https://docs.python.org/3/library/enum.html#creating-an-enum )

src/odemis/acq/align/fastem.py
```diff
@@ -18,22 +20,288 @@
...
+                                                               self._detector, self._dataflow, self._ccd)
+            self.asm_config["descanner"]["scanOffset"] = d_offset
+            configure_asm(self._multibeam, self._descanner, self._detector, self._dataflow, self.asm_config,
+                          upload=False)
+
+    def cancel(self, future):
+        """
+        Cancels the calibrations.
+        :param future: (future) The calibration future.
+        :return: (bool) True if cancelled, TODO False?
```
tepals, line 292: I don't understand the TODO, can you add a bit more information?

tepals (changes requested):

src/odemis/acq/align/fastem.py
Sabrina85: The second one should be correct, as it is still a list that is passed.

Sabrina85: How about this:
`:param calibrations: (list enumeration members from the Calibrations object) List of calibrations that should be run.`

tepals: Thanks! Can you add this to the docstring?

tepals: sounds good

src/odemis/acq/align/fastem.py
Sabrina85: So far this function only ever returns True when called. I wonder if there is a case when it should return False. That would be the case when it is not possible to cancel right? But I cannot think about an event when it is not possible to cancel (maybe only at the very end?).

tepals: I think it's always possible to cancel, indeed maybe at the very end that it just doesn't make sense anymore to actually cancel.

src/odemis/acq/align/fastem.py
pieleric: I think the type can be just `list[Calibrations]` or `list of Calibrations`.

src/odemis/acq/align/fastem.py
pieleric: Yeah, typically you don't need to worry too much. The Future.cancel() is supposed to return `False` if the future cannot be cancelled anymore. Typically it's either because it's is already finished (but that's handled by the parent function) or if it's so late that it's impossible to cancel the very last details. Usually, I don't bother implementing this. Typically, at the very end of the main run function, I check whether the Future was cancelled, and raise a CancelledError, which handles every case. In addition, even if you get this behaviour incorrect nothing will break, because nothing in Odemis relies on this.

Sabrina85: Makes sense and we check at the end of the run function if a CancelledError was raised. Thanks for explaining!

tepals (approved):

src/odemis/acq/align/fastem.py
```diff
@@ -18,22 +20,287 @@
...
+
     return f
 
 
-def _run_fake_alignment():
-    time.sleep(2)
+def estimate_calibration_time(calibrations):
+    """
+    Computes the approximate time it will take to run all calibrations.
+    :param calibrations: (list) List of calibrations that should be run.
```
tepals, line 117: This should also be `list[Calibrations]` I think

PR #2028, opened 2022-02-10
fastem calibrations: connect calibrations to GUI

Sabrina85:


src/odemis/gui/cont/acquisition.py
```diff
@@ -2272,50 +2274,126 @@ def _on_ac_done(self, future):
...
+        self.is_aligned = tab_data.main.is_aligned
+
+        self.gauge_progress.Hide()  # hide progress bar
+        self._panel.Parent.Layout()
+        self.btn_align.Bind(wx.EVT_BUTTON, self.on_align)
+
+        # FIXME: How to differentiate between simulator and real HW?
+        #   The following two calibrations can also be run with the simulator.
+        #   However, the rest not. How make compartible with both cases?
+        # CALIBRATIONS_SIM and CALIBRATIONS as global var somehow?
```
tepals, line 2302: If you decide to make this distinction in this module and not in the test cases, to keep this class readable I would add `FastEMAlignmentControllerSimulator` here and not have it mixed in this code.

src/odemis/gui/cont/acquisition.py
```diff
@@ -2272,50 +2274,126 @@ def _on_ac_done(self, future):
...
+        self.calibrations = [OPTICAL_AUTOFOCUS, IMAGE_TRANSLATION_PREALIGN]
+
+        # check calibration state of system
+        # If backend was not restarted, but only GUI, then the system is in principle still calibrated.
+        # However, so far when closing the GUI the system is also not calibrated anymore.
+        # self.is_aligned.subscribe(self._on_align_state, init=True) # TODO needed?
```
tepals, line 2309: If this comment also applies to the two lines below, can you remove this empty line?

src/odemis/gui/cont/acquisition.py
```diff
@@ -2272,50 +2274,126 @@ def _on_ac_done(self, future):
...
 
-        self._panel.align_gauge_progress.Hide()
-        panel.btn_align.Bind(wx.EVT_BUTTON, self._on_btn_align)
-        tab_data.main.is_aligned.subscribe(self._on_align_state, init=True)
+        if not self.is_aligned.value:
+            self._reset_calibration_gui()  # display estimated calibration time
```
tepals, line 2311: I'm not sure I understand what this does. Does it reset the estimated calibration time?

src/odemis/gui/cont/acquisition.py
```diff
@@ -2272,50 +2274,126 @@ def _on_ac_done(self, future):
...
 
-    def _on_btn_align(self, evt):
-        # TODO: implement cancellation
+        self._future_connector = None  # connect the future to the progress bar and its label
```
tepals, line 2313: How does `self._future_connector = None` connect the future to the progress bar and its label?

tepals (changes requested):

src/odemis/gui/cont/acquisition.py
Sabrina85: That is indeed an idea. On the other hand all the rest of the code is the same. So I am a bit undecided. Is passing the calibrations in the init an option?

src/odemis/gui/cont/acquisition.py
Sabrina85: By default there are some random 30 seconds displayed, which I don't know where they come from. So I wanted here to actually set the calibration time needed. If we manage that the system is still calibrated after closing the GUI, we could have an else statement which resets the gui to show the label "calibrated".

src/odemis/gui/cont/acquisition.py
Sabrina85: I don't know if it is needed. I copied it from the acquisition. In on align it actually does this.

src/odemis/gui/cont/acquisition.py
tepals: that might work, if it's only one or two lines it's also fine to have it here, just try to keep readability in mind

src/odemis/gui/cont/acquisition.py
tepals: ok

src/odemis/gui/cont/acquisition.py
tepals: can you then update the comment to: `# is used to connect the future to the progress bar and its label`

src/odemis/gui/cont/acquisition.py
Sabrina85: @pieleric how would you suggest to solve this?

src/odemis/gui/cont/acquisition.py
pieleric: It shouldn't be in the `if` block. This is an attribute to store the ProgressiveFutureConnector. You can indicate this as comment.

src/odemis/gui/cont/acquisition.py
pieleric: There are two questions.
* How to know whether the current hardware is a simulator or not. I don't have any great idea. I think it should be some sort of flag on the microscope file. 
* In the GUI, I agree it's nice to be able to change just a attribute like `.calibrations` based on whether it's the simulator or not. If that's just a line, indeed, a single `if` is simplest to read. If it's more complex, let's use a subclass (of which .calibrations can be a class attribute).

src/odemis/gui/cont/acquisition.py
```diff
@@ -2272,50 +2274,126 @@ def _on_ac_done(self, future):
...
+
+        self._panel.Layout()
+
+    def on_cancel(self, evt):
+        """
+        Cancel the calibration task.
+        Called during calibration when pressing the cancel button.
+        :param evt: (GenButtonEvent) Button triggered.
+        """
+        align_fastem._executor.cancel()  # all the rest will be handled by on_alignment_done()
```
pieleric, line 2399: Any reason to make a whole function for this? This could be just a single line where the call happens.

pieleric (changes requested):

src/odemis/gui/cont/acquisition.py
Sabrina85: Not really. I wanted to make the code similar to the acquisition code. There it is also only one line.

src/odemis/gui/cont/acquisition.py
Sabrina85: It is not in the if block right?

src/odemis/gui/cont/acquisition.py
```diff
@@ -2272,50 +2274,126 @@ def _on_ac_done(self, future):
...
+        # FIXME: How to differentiate between simulator and real HW?
+        #   The following two calibrations can also be run with the simulator.
+        #   However, the rest not. How make compartible with both cases?
+        # CALIBRATIONS_SIM and CALIBRATIONS as global var somehow?
+        self.calibrations = [OPTICAL_AUTOFOCUS, IMAGE_TRANSLATION_PREALIGN]
+
+        # check calibration state of system
+        # If backend was not restarted, but only GUI, then the system is in principle still calibrated.
+        # However, so far when closing the GUI the system is also not calibrated anymore.
+        # self.is_aligned.subscribe(self._on_align_state, init=True) # TODO needed?
```
Sabrina85, line 2308: @pieleric can you comment on this please?

src/odemis/gui/cont/acquisition.py
```diff
@@ -2272,50 +2274,126 @@ def _on_ac_done(self, future):
...
         # make it look nice.
-        self._panel.align_spacer_panel.Hide()
-        self._panel.align_lbl_gauge.Hide()
-        self._panel.btn_align.SetLabel("Cancel")
+        # self._panel.align_lbl_gauge.Hide()  # hide label progress bar TODO put below gauge instead of hide?
+        self._panel.align_spacer_panel.Hide()  # hide space progress bar and progress bar label
+        self._panel.btn_align.SetLabel("Cancel")  # change button label
         self._panel.Layout()
 
+        # TODO loop over calibrations here and then use batch future?
```
Sabrina85, line 2338: @pieleric can you comment on this please?

tepals (approved):

src/odemis/gui/cont/acquisition.py
pieleric: What? The subscription to is_aligned? I don't really know what's the expected behaviour. Apparently, it was only used to show an indication that the system alignment is complete. I don't think this was completely required. If it's clear that the calibration is done (eg, the gauge is full), then it should be fine.
Can the system go out of calibration because (eg, some settings are changed, or the chamber was opened)? If so it should be clear to the user that the alignment has to be redone. Then, subscribing to is_aligned might be useful.

src/odemis/gui/cont/acquisition.py
pieleric: Well, `align.fastem.align()` already takes care of wrapping all the calibrations into a single future, so you don't need to do that.

pieleric (approved):

src/odemis/gui/cont/acquisition.py
Sabrina85: Yes indeed, it was just more a thought to keep it the same style as the acquisition, where we have the batch future and the `fastem.acquire()` only takes care of one ROA at a time.

src/odemis/gui/cont/acquisition.py
Sabrina85: Ok, then I suggest to remove it for now. In principle I think it is a good idea to realign the system when opening the chamber or pumping/venting. The optical autofocus will be off. On the other hand we run it anyways for now before each ROA. The other calibrations are more stable I think.

src/odemis/gui/cont/acquisition.py
```diff
@@ -2272,50 +2274,120 @@ def _on_ac_done(self, future):
...
-        self._panel.align_gauge_progress.Hide()
-        panel.btn_align.Bind(wx.EVT_BUTTON, self._on_btn_align)
-        tab_data.main.is_aligned.subscribe(self._on_align_state, init=True)
+        self.gauge_progress.Hide()  # hide progress bar
+        self._panel.Parent.Layout()
+        self.btn_align.Bind(wx.EVT_BUTTON, self.on_align)
+
+        # Check if we deal with a real or simulated microscope. If it is a simulator,
+        # we cannot run all calibrations yet.
+        if model.getComponent(role="mbsem") == "FASTEM-sim":  # it's a simulator
```
pieleric, line 2301: I doubt that you can compare a component with a string, and that it'd be ever return `True`. Probably you forgot `.name`. 
Also, there is an easier way to get the microscope component: `model.getMicroscope()`.

BTW, it feels really easy to break. We normally never user the name as a way to change the behaviour. On the other hand, I cannot think of a better way to detect that it's a simulator. Maybe write `.name.lower().endswith("sim")`?
Could you add something like a `# HACK warning: we don't have an official way to detect a component is simulated, but here, we really need to know that it's simulated otherwise we can never simulate an acquisition`.

pieleric (changes requested):

src/odemis/gui/cont/acquisition.py
Sabrina85: yes indeed I forgot the `.name`. Had only tested it in the terminal. I did not find anything better as well. That is why I went for this one. I will add the warning! Hopefully we have a simulation for all calibrations some time and then we can just get rid of this one here.

tepals (approved):

PR #2029, opened 2022-02-10
batch future: fix the progress update

Sabrina85:
When the batch future was used to represent the progress of multiple futures, only for the first future the progress in the overall progress was correctly shown. As soon as the first future was finished, the updated end of the batch future was incorrectly calculated. The same start of the future was used plus adding the remaining time for the not yet finished futures. Thus, the end time of the batch future moved closer to the start in time. As a result, the GUI displayed always "a few seconds left" for the remaining futures that still needed to be executed.

### DIFF
--- a/src/odemis/model/_futures.py
+++ b/src/odemis/model/_futures.py
@@ -502,21 +502,29 @@ class ProgressiveBatchFuture(ProgressiveFuture):
     """
     def __init__(self, futures):
         """
-        futures (ProgressiveFuture --> float): dict specifying futures and estimated times
+        :param futures (dict: ProgressiveFuture --> float): Keys are futures and values are the
+                        respective time estimates for the duration of the future.
         """
         self.futures = futures
-        start = time.time()
+        start = time.time()  # start time of the batch future
         super().__init__(start=start, end=start + sum(self.futures.values()))
         self.set_running_or_notify_cancel()
 
         for f in self.futures:
-            f.add_update_callback(self._on_future_update)
-            f.add_done_callback(self._on_future_done)
+            f.add_update_callback(self._on_future_update)  # called whenever set_progress is called
+            f.add_done_callback(self._on_future_done)  # called when future is done
 
     def _on_future_update(self, f, start, end):
-        if f.running():  # only care about future which are running, start/end estimates for others not reliable
-            self.futures[f] = end - start
-            self.set_progress(end=self._estimate_end())
+        """
+        Whenever progress on the single future is reported, the progress (or end) for the batch future
+        is updated accordingly.
+
+        :param start: (float) Start time of a single future in the batch future in seconds.
+        :param end: (float) End of a single future in the batch future in seconds.
+        """
+        if f.running():  # only care about the future which is currently running
+            self.futures[f] = end - start  # update the start and end of the single future
+            self.set_progress(end=self._estimate_end())  # set the progress for batch future
 
     def _on_future_done(self, f):
         self.set_progress(end=self._estimate_end())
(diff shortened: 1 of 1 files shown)


src/odemis/model/_futures.py
```diff
@@ -502,21 +502,29 @@ class ProgressiveBatchFuture(ProgressiveFuture):
...
         self.futures = futures
-        start = time.time()
+        start = time.time()  # start time of the batch future
         super().__init__(start=start, end=start + sum(self.futures.values()))
         self.set_running_or_notify_cancel()
 
         for f in self.futures:
-            f.add_update_callback(self._on_future_update)
-            f.add_done_callback(self._on_future_done)
+            f.add_update_callback(self._on_future_update)  # called whenever set_progress is called
```
tepals, line 514: is this when set_progress is called for the single future, or for the batch future?

Sabrina85: I think on the single future. Remember that the breakpoint in the `_on_future_update` was hit after every tile. And then `set_progress` on the batch future is called there.

pieleric: My expectation is that this is called whenever the set_progress of each "single" future is called. That might need clarification in the comment.

src/odemis/model/_futures.py
```diff
@@ -502,21 +502,29 @@ class ProgressiveBatchFuture(ProgressiveFuture):
...
-            self.set_progress(end=self._estimate_end())
+        """
+        Whenever progress on the single future is reported, the progress (or end) for the batch future
+        is updated accordingly.
+
+        :param start: (float) Start time of a single future in the batch future in seconds.
+        :param end: (float) End of a single future in the batch future in seconds.
+        """
+        if f.running():  # only care about the future which is currently running
+            self.futures[f] = end - start  # update the start and end of the single future
```
pieleric, line 526: I'd rephrase it as "update the time estimation of time left of that future".

src/odemis/model/_futures.py
```diff
@@ -538,8 +546,20 @@ def _on_future_done(self, f):
...
+        Estimate the end of the batch future based on the current time plus the
+        estimated time for all futures that still need to be executed.
+
+        returns: (float) Newly estimated end time based for the batch future.
+        """
+        duration = time.time()
+        for f, t in self.futures.items():
+            # only take futures into account that are not executed (finished) yet
+            if not f.done():
+                duration += t
```
pieleric, line 559: I liked the compact way of before. Actually, I would write:
`duration =  sum(t for f, t in self.futures.items() if not f.done())`

Note that duration implies that it's a difference of time between the start and the end. Maybe you could call it `time_left`.

src/odemis/model/_futures.py
```diff
@@ -538,8 +546,20 @@ def _on_future_done(self, f):
...
+        # FIXME: it does not update often enough in the GUI, the estimate seems reasonable, but
+        #  gauge does not execute 1s when showing 1 second has passed -> takes longer thought time would match well
+        return duration
```
pieleric, line 562: It's supposed to return the time at which it expects to end. So that variable shouldn't be called `duration`. I think it'd be fine to write: `return time.time() + time_left` .

src/odemis/model/_futures.py
```diff
@@ -538,8 +546,20 @@ def _on_future_done(self, f):
...
+
+        returns: (float) Newly estimated end time based for the batch future.
+        """
+        duration = time.time()
+        for f, t in self.futures.items():
+            # only take futures into account that are not executed (finished) yet
+            if not f.done():
+                duration += t
+        # FIXME: it does not update often enough in the GUI, the estimate seems reasonable, but
+        #  gauge does not execute 1s when showing 1 second has passed -> takes longer thought time would match well
```
pieleric, line 561: I don't get the issue that you're trying to describe in this comment.
Be aware that set_progress() is maybe not called so frequently, so it's not going to change the GUI directly. When the GUI doesn't receive any update on the estimated time, it assumes that the previous estimation was correct and update the displayed remaining time every 1s.

pieleric (changes requested):

src/odemis/model/_futures.py
Sabrina85: It was not updated every 1s put stayed at the same value for over 10 sec.

Sabrina85: For imaging of 2 overview images I see the following:
estimated time:  ~ 3 min 45sec (dwell time =  25ns)
start time acq: ~ 2 min 23sec updates once to 2 min 22 sec, then stays there over 10 sec before counting down
sec image: jumps to ~36sec

Sabrina85: Found that this issue is related to the bad estimated time for the tiled acquisition and has nothing to do with the changes here.

pieleric (approved):

PR #2031, opened 2022-02-10
beamshift TFS: add dependency on ebeam to get calibration matrix MD needed on beamshift component

Sabrina85:


src/odemis/driver/test/tfsbc_test.py
```diff
@@ -60,6 +67,23 @@ def setUpClass(cls):
...
+        """Test dependency on scanner."""
+        # TODO only works with xt simulator now - how to make generic? for xt simulator has special flag "sim"
+        if TEST_NOHW:
+            self.microscope = xt_client.SEM(**CONFIG_MB_SEM)
+            for child in self.microscope.children.value:
+                if child.name == CONFIG_MB_SCANNER["name"]:
+                    self.scanner = child
+
+            dependencies = {"scanner": self.scanner}
+            self.bc = tfsbc.BeamShiftController("DC Offset", None, PORT, "FT2OMDD5", dependencies)
```
tepals, line 80: why use `self.microscope` and `self.bc`, when you only use it in this test case?

src/odemis/driver/test/tfsbc_test.py
```diff
@@ -60,6 +67,23 @@ def setUpClass(cls):
...
+            # Note: for simulator matrix is only 0s
+            calib_md = self.bc.getMetadata()[model.MD_CALIB]
+            self.assertIsNotNone(calib_md)
+            self.assertIsInstance(calib_md, list)
```
tepals, line 84: I think it's fine to only have `self.assertIsInstance(calib_md, list)`, if it's a list it cannot be None.

src/odemis/driver/tfsbc.py
```diff
@@ -183,6 +185,9 @@ def __init__(self, name, role, port=None, serialnum=None, **kwargs):
                                            cls=(int, float), unit="m",
                                            setter=self._setShift)
 
+        if dependencies:
```
tepals, line 188: I would also check that scanner is in the dependencies:
```
if dependencies and "scanner" in dependencies:
```

tepals (changes requested):

src/odemis/driver/test/tfsbc_test.py
```diff
@@ -60,6 +67,23 @@ def setUpClass(cls):
...
+        # TODO only works with xt simulator now - how to make generic? for xt simulator has special flag "sim"
+        if TEST_NOHW:
+            self.microscope = xt_client.SEM(**CONFIG_MB_SEM)
+            for child in self.microscope.children.value:
+                if child.name == CONFIG_MB_SCANNER["name"]:
+                    self.scanner = child
+
+            dependencies = {"scanner": self.scanner}
+            self.bc = tfsbc.BeamShiftController("DC Offset", None, PORT, "FT2OMDD5", dependencies)
+            # Note: for simulator matrix is only 0s
```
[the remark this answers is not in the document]
Sabrina85: zeroes :D

src/odemis/driver/test/tfsbc_test.py
Sabrina85: Ah this was because I tried to add it directly in the setupclass method. However, that breaks all the other test cases as those rely on the calibration matrix that is hardcoded there. For the simulator we only get zeros returned for the calibration matrix. I think it is not the most elegant solution now, to instantiate the beamshift component in the setupclass and then here again, just with adding some dependency. Any suggestions how I can do this differently? The only idea I had was to create a new test class, but that seems also not right.

tepals: :thinking: I don't have a good idea right away, a new test class also seems a bit over the top, but I also cannot think of another way

src/odemis/driver/test/tfsbc_test.py
```diff
@@ -39,6 +39,13 @@
 else:
     PORT = "/dev/ttyUSB*"
 
+CONFIG_DETECTOR = {"name": "detector", "role": "se-detector"}
+CONFIG_MB_SCANNER = {"name": "mb-scanner", "role": "ebeam", "hfw_nomag": 1}
```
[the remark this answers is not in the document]
Sabrina85: True, I copied it from xt_client_test.py. You could consider changing it there then as well.

src/odemis/driver/test/tfsbc_test.py
Sabrina85: @pieleric do you have a good suggestions?

src/odemis/driver/test/tfsbc_test.py
```diff
@@ -60,6 +67,23 @@ def setUpClass(cls):
     def tearDownClass(cls):
         pass
 
+    def test_dependency_scanner(self):
+        """Test dependency on scanner."""
+        # TODO only works with xt simulator now - how to make generic? for xt simulator has special flag "sim"
+        if TEST_NOHW:
```
Sabrina85, line 73: How to handle this correctly. For xt there are 3 options for TEST_HW: 
0 (HW testing), "sim" (simulator testing), 1 (skip most tests). Here we only have 2 options. So if TEST_HW = 1, this test is run, but is then the xt adapter simulator not running?

pieleric: I think that the problem is you're relying on the xt_client connected to xtlib adpater (running in simulator mode on the same computer) to get a component which must now have a beamShiftTransformationMatrix VA.  It's a very big hammer for a tiny nail. I'd suggest to either create a mock component "manually" with the given VA (and you can then pass the right config!), or use a `simulated.GenericComponent`. You can have a fake "e-beam" in 2 lines!

pieleric (changes requested):

src/odemis/driver/test/tfsbc_test.py
Sabrina85: I used now the simulated.GenericComponent. Then I can use it directly in the setUpClass for all tests instead of manually adding the matrix to the MD. However, then for all tests the dependency is used. Is it ok to do this? Then we would not need the separate class and all tests pass as usual (I used the hard coded matrix for the simulated component).

src/odemis/driver/test/tfsbc_test.py
```diff
@@ -48,17 +48,24 @@ class TestBeamShiftController(unittest.TestCase):
...
+                                                   "unit": ""}})
+
+        dependencies = {"scanner": scanner}
+        cls.bc = tfsbc.BeamShiftController("DC Offset", None, PORT, "FT2OMDD5", dependencies)
+
+    def test_dependency_scanner(self):
+        """Test dependency on scanner."""
+        calib_md = self.bc.getMetadata()[model.MD_CALIB]
+        self.assertIsInstance(calib_md, list)
+        self.assertIsInstance(calib_md[0][0], float)
```
tepals, line 68: Now that you specify the exact values of the matrix in the set-up I would specifically check: `self.assertEqual(calib_md[0][0], -0.00027788219369730165)`

src/odemis/driver/tfsbc.py
```diff
@@ -160,12 +160,14 @@ class BeamShiftController(model.HwComponent):
...
-    def __init__(self, name, role, port=None, serialnum=None, **kwargs):
+    def __init__(self, name, role, port=None, serialnum=None, dependencies=None, **kwargs):
         """
         :param port (str): (e.g. "/dev/ttyUSB0") or pattern for port ("/dev/ttyUSB*"),
             "/dev/fake" will start the simulator
         :param serialnum (str): serial number of RS485 adapter
-        The connection can be specified by either port or serialnum, it's not needed to provide both.
+            The connection can be specified by either port or serialnum, it's not needed to provide both.
+        :param dependencies (dict str -> scanner):
+            scanner component -> name of the xt multibeam scanner component.
```
tepals, line 170: Because it's optional to provide the `dependencies`, can you update this with a short description of what happens when you don't provide the scanner dependency (`dependencies=None`)

tepals (changes requested):

tepals (approved):

pieleric (approved):

PR #2033, opened 2022-02-11
mdupdater-diagnostic-ccd

Sabrina85:


Sabrina85:
I did not manager to tune the magnification, refractive index and numerical aperture in such a way, that the test cases pass for me with a 1um accuracy for the simulator. I was able to reduce the depthOfField but I could not reduce the accuracy below 10um for the tests with this. I do not know what effect the refractive index had. It cannot be set below 1 anymore due to the recent code changes on the DigitalCamera component.

src/odemis/acq/align/test/clspots_optical_autofocus_test.py
```diff
@@ -83,13 +63,13 @@ def test_autofocus_opt(self):
...
+        self.stage.moveAbs({"z": center_position}).result()
         # Run auto focus
-        future_focus = align.autofocus.CLSpotsAutoFocus(self.diagnostic_cam, self.ofocus)
+        future_focus = align.autofocus.CLSpotsAutoFocus(self.diagnostic_cam, self.stage)
         foc_pos, foc_lev = future_focus.result(timeout=900)
         # Check that the auto focus converged to the correct position and the stage moved to the correct position
-        numpy.testing.assert_allclose(foc_pos, good_focus, atol=0.5e-6)
-        numpy.testing.assert_allclose(self.ofocus.position.value["z"], foc_pos, atol=1e-6)
+        numpy.testing.assert_allclose(foc_pos, good_focus, atol=10e-6)
+        numpy.testing.assert_allclose(self.stage.position.value["z"], foc_pos, atol=10e-6)
```
tepals, line 72: The range of the stage is `ranges: {"z": [0, 100.e-6]}`, shouldn't the autofocus be able to do better than a tolerance of 10e-6?

src/odemis/acq/align/test/clspots_optical_autofocus_test.py
```diff
@@ -44,34 +44,14 @@ class TestAutofocusSim(unittest.TestCase):
...
-        try:
-            test.start_backend(CLSPOTS_SIM_CONFIG)
-        except LookupError:
-            logging.debug("A running backend is already found, skipping tests")
-            cls.backend_was_running = True
-            return
-        except IOError as exp:
-            logging.error(str(exp))
-            raise
+        test.start_backend(CLSPOTS_SIM_CONFIG)
```
tepals, line 50: you don't want to run these tests if you're connected to the hardware, so shouldn't there be a 
```
if not TEST_NOHW:
    skip test
```

src/odemis/acq/align/test/clspots_optical_autofocus_test.py
pieleric: I also don't get why the tolerance has to be so big. The autofocus should finish with steps of DoF/2. I've checked, and until now the simulator had a depthOfField of ~0.1e-6 m (the sensor pixel size defaults to 10µm). 
When setting the ri to 1, you end up with a DoF of 0.78e-6m, which is still not that big. If you bump up the NA to 2, you'd have a DoF of 0.22e-6m, which might help in convincing the autofocus to find the position with more accuracy. Still, it's strange that we have such large inaccuracy given the DoF and the expected step size of the autofocus. Could you dig up a little more before making the test cases very weak?

pieleric (changes requested):

src/odemis/acq/align/test/clspots_optical_autofocus_test.py
Sabrina85: Why do we not want to run them with the HW?

src/odemis/acq/align/test/clspots_optical_autofocus_test.py
Sabrina85: I wasn't able to do better than 10us. Even for 5us it once in a while failed. I agree with you that it does not make sense based on your calculation. I do not really know anything about the autofocus procedure, so I need to find sometime for this.

Sabrina85: If I modify line `276` in `simcam`: `dist = abs(pos - self._metadata[model.MD_FAV_POS_ACTIVE]["z"]) / self.depthOfField.value` then I manage to see a clear blur when moving the stage by 1um. I managed then also that the test once in a while passes.  If it does not pass I get a timeout error on the simcam:
`        det.data.subscribe(receive_one_image)
        if not is_received.wait(timeout):
            det.data.unsubscribe(receive_one_image)
>           raise IOError("No data received after %g s" % (timeout,))
E           OSError: No data received after 3.02 s`

tepals (approved):

src/odemis/driver/simcam.py
```diff
@@ -273,7 +275,8 @@ def _generate(self):
         if self._focus:
             # apply the defocus
             pos = self._focus.position.value['z']
-            dist = abs(pos - self._metadata[model.MD_FAV_POS_ACTIVE]["z"]) / (self.depthOfField.value * 10)
+            # max blur of sqrt(50) pixels, else image generation takes too long
+            dist = math.sqrt(min(abs(pos - self._metadata[model.MD_FAV_POS_ACTIVE]["z"]) / self.depthOfField.value, 50))
```
pieleric, line 279: I think that sqrt(50) is a little too small  now (7px). It's nice to be able to get a really blurry image. I'd suggest to make it more like `min(sqrt(...), 30)`.  In that case, it's mostly clipped by the sqrt, but still limited to 30 (for values > 900).

src/odemis/acq/align/test/clspots_optical_autofocus_test.py
```diff
@@ -44,34 +44,17 @@ class TestAutofocusSim(unittest.TestCase):
...
-            test.start_backend(CLSPOTS_SIM_CONFIG)
-        except LookupError:
-            logging.debug("A running backend is already found, skipping tests")
-            cls.backend_was_running = True
-            return
-        except IOError as exp:
-            logging.error(str(exp))
-            raise
+        if not TEST_NOHW:
+            raise unittest.SkipTest("No simulator running. Skipping tests.")
```
pieleric, line 51: So these tests need the simulator to run. But it should still be fine to run the simulator backend if the computer is connected to the real hardware? So never skip, just always run the simulator.

pieleric (changes requested):

pieleric (approved):

pieleric:
Could you squash the commits to make them 3 commits (ie, yaml file, mdupdater, and simcam)?

PR #2034, opened 2022-02-11
component: fix message in warnings for point spread function size

Sabrina85:


effting (approved):

effting:
Perhaps also change line 672 'spot size' -> 'point spread function size'

pieleric:
Thanks Andries @effting for noticing the extra place to update.

PR #2047, opened 2022-02-25
Piezomotor ismoving fix and logging

tepals:


src/odemis/driver/piezomotor.py
```diff
@@ -593,9 +599,12 @@ def isMovingClosedLoop(self, axis):
         if d2 & 0b100:
             raise PMDError(6, "External limit reached.")
```
Kleijwegt, line 600: Adding the current positioning a debug message could help. Especially if the motor has two limits (lower and higher) since then you can find out on which end of the limit the motor currently is.

tepals: good one, I'll also add on which axis the limit was reached

pieleric (approved):

src/odemis/driver/piezomotor.py
```diff
@@ -425,26 +425,32 @@ def _waitEndMotion(self, f, shifts):
...
         while moving_axes:
             if f._must_stop.is_set():
                 for axname in moving_axes:
                     self.stopAxis(self._axis_map[axname])
                     raise CancelledError()
 
             if time.time() > end_time:
-                raise TimeoutError("Timeout after while waiting for end of motion on axes %s for %g s" % (moving_axes, max_dur))
+                raise TimeoutError(
+                    "Timeout while waiting for end of motion on axes %s for %g s" % (moving_axes, max_dur))
```
Kleijwegt, line 437: Why don't you also use f strings in the error message?

Kleijwegt (approved):

src/odemis/driver/piezomotor.py
tepals: I just reformatted this error so it would fit the line length. I used the f strings for new errors and log messages.

PR #2048, opened 2022-02-28
Acquisition fastem: add overlap argument to stage movement.

tepals:
During megafield acquisition it should be possible have the fields overlap by a certain factor. This commit adds an overlap argument to the ROA class.
calculate_field_indices has been updated to calculate the number fields when there is overlap. And get_abs_stage_movement has been updated to get the absolute stage movement for overlapping fields.

### DIFF
--- a/src/odemis/acq/fastem.py
+++ b/src/odemis/acq/fastem.py
@@ -125,6 +125,19 @@ class FastEMROA(object):
         :return: (list of nested tuples (col, row)) The column and row field indices of the field images in the order
                  they should be acquired. The tuples are re-ordered so that the single field images resembling the
                  ROA are acquired first rows then columns.
+
+        Note: The number of fields with overlap is calculated in the following way:
+        An ROA with overlap can be visualized as follows:
+        |------|--⁞----|--⁞----|--⁞
+        If you have 3 fields of 6400 pixels wide with 400 pixels overlap, the resulting megafield is
+        6000 * 3 + 400 pixels in width and the width is abs(r - l). The number of fields that fit in a
+        certain width can be derived from this:
+        abs(r - l) = 6000 * 3 + 400
+        (abs(r-l) - 400) / 6000 = 3 fields
+        with:
+        400 = 6400 * overlap = field_size * overlap
+        6000 = 6400 * (1 - overlap) = field_size * (1 - overlap)
+        => n_fields = (abs(r - l) - field_size * overlap) / (field_size * (1 - overlap))
         """
         l, t, r, b = self.coordinates.value  # tuple of floats: l, t, r, b coordinates in m
         px_size = self._multibeam.pixelSize.value
@@ -136,7 +149,7 @@ class FastEMROA(object):
 
         # Note: floating point errors here, can result in an additional row or column of fields (that's fine)
         # When fields overlap the number of fields is calculated by subtracting the size of the field that overlaps
-        # from the total roa size and dividing that by the non-overlapping field size.
+        # from the total roa size and dividing that by the non-overlapping field size. (see note in docstring)
         n_hor_fields = math.ceil((abs(r - l) - field_size[0] * self.overlap) / (field_size[0] * (1 - self.overlap)))
         n_vert_fields = math.ceil((abs(b - t) - field_size[1] * self.overlap) / (field_size[1] * (1 - self.overlap)))
         # Note: Megafields get asymmetrically extended towards the right and bottom.
--- a/src/odemis/acq/fastem.py
+++ b/src/odemis/acq/fastem.py
@@ -58,7 +58,7 @@ class FastEMROA(object):
     and detector.
     """
 
-    def __init__(self, name, coordinates, roc, asm, multibeam, descanner, detector):
+    def __init__(self, name, coordinates, roc, asm, multibeam, descanner, detector, overlap=0):
         """
         :param name: (str) Name of the region of acquisition (ROA). It is the name of the megafield (id) as stored on
                      the external storage.
@@ -70,6 +70,10 @@ class FastEMROA(object):
         :param multibeam: (technolution.EBeamScanner) The multibeam scanner component of the acquisition server module.
         :param descanner: (technolution.MirrorDescanner) The mirror descanner component of the acquisition server module.
         :param detector: (technolution.MPPC) The detector object to be used for collecting the image data.
+        :param overlap: (float), optional
+            The amount of overlap required between single fields. An overlap of 0.2 means that two neighboring fields
+            overlap by 20%. By default, the overlap is 0, this means there is no overlap and one field is exactly next
+            to the neighboring field.
         """
         self.name = model.StringVA(name)
         self.coordinates = model.TupleContinuous(coordinates,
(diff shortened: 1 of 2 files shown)


src/odemis/acq/test/fastem_test.py
```diff
@@ -216,6 +216,46 @@ def test_calculate_field_indices(self):
...
+        roa_name = time.strftime("test_megafield_id-%Y-%m-%d-%H-%M-%S")
+
+        field_size_x = res_x * px_size_x
+        field_size_y = res_y * px_size_y
+
+        for overlap in (0, 0.0625, 0.2, 0.5, 0.7):
+            # The coordinates of the ROA in meters.
+            xmin, ymin = (0, 0)
+            # The max field size is the number of fields multiplied with the field size including overlap, plus
+            # the extra overlap added to the end.
```
Sabrina85, line 236: What do you mean with "extra overlap to the end"?

src/odemis/acq/fastem.py
```diff
@@ -126,10 +131,14 @@ def _calculate_field_indices(self):
...
 
         # The size of a field consists of the effective cell images excluding overscanned pixels.
-        field_size = (field_res[0] * px_size[0], field_res[1] * px_size[1])
+        field_size = (field_res[0] * px_size[0],
+                      field_res[1] * px_size[1])
+
         # Note: floating point errors here, can result in an additional row or column of fields (that's fine)
-        n_hor_fields = math.ceil(abs(r - l) / field_size[0])
-        n_vert_fields = math.ceil(abs(b - t) / field_size[1])
+        # When fields overlap the number of fields is calculated by subtracting the size of the field that overlaps
```
Sabrina85, line 138: How did you get to this formula? It produces the correct values, but I am curious as it is not really intuitive :) .

Sabrina85 (commented):
It looks fine, just some questions.

src/odemis/acq/test/fastem_test.py
tepals: The way I understand it is that if we have overlap the fields within a megafield overlap. The last row and column do not overlap with any fields to there right or bottom, therefore they can actually use their full size.
Say you have 3 fields of 8 of these '-' and the overlap is 2, then the total size is 6 * 3 + 2. I tried to visualize it, where the left border of each field is drawn with a | and the right border of each field with a ⁞
|------|--⁞----|--⁞----|--⁞

src/odemis/acq/test/fastem_test.py
```diff
@@ -216,6 +216,46 @@ def test_calculate_field_indices(self):
...
+
+        roa_name = time.strftime("test_megafield_id-%Y-%m-%d-%H-%M-%S")
+
+        field_size_x = res_x * px_size_x
+        field_size_y = res_y * px_size_y
+
+        for overlap in (0, 0.0625, 0.2, 0.5, 0.7):
+            # The coordinates of the ROA in meters.
+            xmin, ymin = (0, 0)
+            # The max field size is the number of fields multiplied with the field size including overlap, plus
```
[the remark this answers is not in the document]
Sabrina85: Maybe I am confused as you calculate here the size of the ROA. How about:
`Calculate the ROA size for a well specified number of fields taking the specified overlap between fields into account.  As the fields of the last row and column do not have any neighbor to their right and bottom, do take their full size into account.`

src/odemis/acq/test/fastem_test.py
Sabrina85: I like the example :)  Please add it!

src/odemis/acq/test/fastem_test.py
```diff
@@ -216,6 +216,46 @@ def test_calculate_field_indices(self):
...
+                                   self.mppc,
+                                   overlap)
+
+            expected_indices = [(0, 0), (1, 0), (2, 0), (0, 1), (1, 1), (2, 1)]  # (col, row)
+            field_indices = roa._calculate_field_indices()
+
+            # Floating point errors can result in an extra field, which is fine.
+            # Check that maximum 1 extra field index in x and 1 in y is calculated.
+            self.assertLessEqual(max(field_indices)[0] - max(expected_indices)[0], 1)
+            self.assertGreaterEqual(max(field_indices)[0] - max(expected_indices)[0], 0)
```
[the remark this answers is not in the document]
Sabrina85: I agree with your statement. 
If you check the LessEqual I don't see how the GreaterEqual can ever go wrong. In LessEqual you check that the difference between expected and actual indices is either 0 or 1 (or can it ever be negative?). What do I miss?

Sabrina85 (changes requested):

src/odemis/acq/test/fastem_test.py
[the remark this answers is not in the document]
tepals: I'm checking that `max(field_indices)[0] - max(expected_indices)[0]` is less than 1 and greater than 0. If the field indices found are `[(1, 1), (2, 2)]` and the expected field indices are `[(1, 1), (2, 2), (3, 3)]` then `max(field_indices)[0] - max(expected_indices)[0] = (-1, -1)` and `(-1, -1)` is less than 1 but not greater than 0.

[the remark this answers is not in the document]
Sabrina85: Yes makes sense now. Thanks for explaining with an example.

src/odemis/acq/fastem.py
Sabrina85: Can you add this example as a comment? I think the next reader might also struggle here.

Sabrina85 (approved):

src/odemis/acq/fastem.py
```diff
@@ -120,16 +125,33 @@ def _calculate_field_indices(self):
...
+        |------|--⁞----|--⁞----|--⁞
+        If you have 3 fields of 6400 pixels wide with 400 pixels overlap, the resulting megafield is
+        6000 * 3 + 400 pixels in width and the width is abs(r - l). The number of fields that fit in a
+        certain width can be derived from this:
+        abs(r - l) = 6000 * 3 + 400
+        (abs(r-l) - 400) / 6000 = 3 fields
+        with:
+        400 = 6400 * overlap = field_size * overlap
+        6000 = 6400 * (1 - overlap) = field_size * (1 - overlap)
+        => n_fields = (abs(r - l) - field_size * overlap) / (field_size * (1 - overlap))
```
pieleric, line 140: Two notes.
Firstly, if the user requests a very tiny region, less than the overlap (eg, abs(r-l) == 300px), you end up with negative numbers, which eventually get rounded up to 0. It should probably handle such case. Maybe just use a "if width < field_size, then n == 1".

Secondly, be aware that in the tiledacq code we use a different formula. We use the simpler n_fields = abs(r - l) / (field_size * (1 - overlap)) . Basically, it considers that the first field is also smaller. The reason for this (in addition to simplifying a lot of code) is that the overlap is there to compensate for the stage (or other hw) inaccuracy. So you cannot really consider that the first field is perfectly overlapping the requested area. Therefore it's safer to also consider the first field smaller.

PR #2053, opened 2022-03-04
Acquisition fastem: perform pre-calibrations outside the region of ac…

tepals:
…quisition

* perform the pre-calibrations on the field with index (-1, -1) to limit beam damage in the actual acquired area.
* do not skip test_get_pos_first_tile in the mock test class, because get_pos_first_tile does work with mocking.
* in the mocked test cases update multibeam.resolution.value from (800, 800) to (6400, 6400), to match the actual system.

Will test on Monday

### DIFF
--- a/src/odemis/acq/fastem.py
+++ b/src/odemis/acq/fastem.py
@@ -320,12 +320,17 @@ class AcquisitionTask(object):
         # Update the position of the first tile.
         self._pos_first_tile = self.get_pos_first_tile()
 
-        # Move the stage to the first tile, to ensure the autofocus and image translation pre-align are
-        # done in the correct location and the correct position is stored in the megafield metadata yaml file.
-        self.move_stage_to_next_tile()
-
         if self._pre_calibrate:
+            # Move the stage to the tile with index (-1, -1), to ensure the autofocus and image translation pre-align
+            # are done to the top left of the first field, outside the region of acquisition to limit beam damage.
+            self.field_idx = (-1, -1)
+            self.move_stage_to_next_tile()
             self.pre_calibrate()
+
+        # Move the stage to the first tile, to ensure the correct position is
+        # stored in the megafield metadata yaml file.
+        self.field_idx = (0, 0)
+        self.move_stage_to_next_tile()
 
         # set the sub-directories (<acquisition date>/<project name>) and megafield id
         self._detector.filename.value = os.path.join(self._path, self._roa.name.value)
--- a/src/odemis/acq/fastem.py
+++ b/src/odemis/acq/fastem.py
@@ -321,11 +321,16 @@ class AcquisitionTask(object):
         self._pos_first_tile = self.get_pos_first_tile()
 
         if self._pre_calibrate:
+            # The pre-calibrations should run on a position that lies a full field
+            # outside the ROA, therefore temporarily set the overlap to zero.
+            overlap_init = self._roa.overlap
+            self._roa.overlap = 0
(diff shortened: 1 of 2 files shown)


tepals:
Testing was successful, pre-calibration is done outside the ROA.

src/odemis/acq/fastem.py
```diff
@@ -320,13 +320,18 @@ def run(self):
...
-        self.move_stage_to_next_tile()
-
         if self._pre_calibrate:
+            # Move the stage to the tile with index (-1, -1), to ensure the autofocus and image translation pre-align
+            # are done to the top left of the first field, outside the region of acquisition to limit beam damage.
+            self.field_idx = (-1, -1)
+            self.move_stage_to_next_tile()
             self.pre_calibrate()
 
+        # Move the stage to the first tile, to ensure the correct position is
```
Sabrina85, line 330: Does it not make sense to move line 320/321 here but outside of the if statement? Now you move to the first tile of the ROA, then you move outside of the ROA for pre-calibration and then back to the first tile. Instead you could only move to the (-1, -1) directly and then to (0,0). That would save one step I think.

src/odemis/acq/test/fastem_test.py
```diff
@@ -655,6 +655,18 @@ def test_get_abs_stage_movement_overlap(self):
             actual_position = task.get_abs_stage_movement()  # [m]
             numpy.testing.assert_allclose(actual_position, expected_position)
 
+            # Verify that for pre-calibrations compared to the top left corner of the ROA, the stage
```
Sabrina85, line 658: I suggest to make it a separate test case. This test is already extensive and you can just make a simple test for only checking this and only for one overlap.

Sabrina85 (changes requested):

src/odemis/acq/fastem.py
tepals: Line 321 does not move the stage, it just calculates the position of the first tile. To move to (-1, -1) the position of the first tile needs to have been calculated.

Sabrina85: Ah right this was needed as `move_stage_to_next_tile` needs to know the position of the first tile.

src/odemis/acq/fastem.py
```diff
@@ -320,12 +320,22 @@ def run(self):
...
 
-        # Move the stage to the first tile, to ensure the autofocus and image translation pre-align are
-        # done in the correct location and the correct position is stored in the megafield metadata yaml file.
-        self.move_stage_to_next_tile()
-
         if self._pre_calibrate:
+            # The pre-calibrations should be done a full field outside the ROA,
+            # therefore temporarily set the overlap to zero.
+            overlap_init = self._roa.overlap
+            self._roa.overlap = 0
```
tepals, line 327: This is a bit ugly, but I don't have a better idea of solving this.

src/odemis/acq/fastem.py
```diff
@@ -321,11 +321,16 @@ def run(self):
         self._pos_first_tile = self.get_pos_first_tile()
 
         if self._pre_calibrate:
+            # The pre-calibrations should be done a full field outside the ROA,
```
Sabrina85, line 324: missing word "on" ?

src/odemis/acq/fastem.py
Sabrina85: mhm I also do not have a better idea right now.

Sabrina85 (changes requested):

Sabrina85 (approved):

PR #2054, opened 2022-03-07
fastem-alignment-controller-input-changes

Sabrina85:


src/odemis/gui/main_xrc.py
```diff
@@ -5491,16 +5490,15 @@ def __init_resources():
                               <flag>wxTOP|wxBOTTOM|wxRIGHT</flag>
                               <border>14</border>
                             </object>
-
                             <orient>wxHORIZONTAL</orient>
                           </object>
                         </object>
-                        <label>ALIGNMENT</label>
+                        <label>CALIBRATION 1</label>
```
tepals, line 5496: I think this should be `<label>CALIBRATION 2</label>`, otherwise it doesn't match line 5510. And on line 5501 you also have `pnl_fastem_calibration_2`

tepals (changes requested):

src/odemis/gui/main_xrc.py
Sabrina85: This line is still referring to the item in line 5443. Item in line 5510 is referring to line 5501.

tepals (approved):

pieleric (changes requested):

tepals (approved):

pieleric (approved):

PR #2055, opened 2022-03-07
fastem overview tab: add calibration panel to run optical autofocus

Sabrina85:


Sabrina85:
pr is depending on pr #2054. Currently, label of button is also "Calibrate", label of panel is "CALIBRATION". Would be nice to integrate that we actually run the optical autofocus, but at the same time we can reuse the AlignmentController class. Maybe add another input arg? Tooltip should be integrated as well. Temporary solution could be: rename panel header to optical autofocus. In case other calibrations get added to this panel, implement a more flexible solution.

src/odemis/gui/main_xrc.py
```diff
@@ -5802,6 +5805,64 @@ def __init_resources():
                   <orient>wxVERTICAL</orient>
                   <object class="sizeritem">
                     <object class="FoldPanelBar" name="fpb_settings">
+                    <object class="FoldPanelItem">
```
pieleric, line 5808: All these changes should be done in main.xrc, and then generated into main_xrc.py, so you should always push changes to both files at the same time.

Sabrina85: I will fix it as soon as the other pr is merged :)

pieleric (approved):

tepals (approved):

PR #2057, opened 2022-03-09
fastem-fixes-acq-calib-managers

Sabrina85:


src/odemis/acq/fastem.py
```diff
@@ -223,7 +223,13 @@ def acquire(roa, path, scanner, multibeam, descanner, detector, stage, ccd, beam
...
                 (Exception or None): Exception raised during the acquisition or None.
     """
-    f = model.ProgressiveFuture()
+
+    est_dur = roa.estimate_acquisition_time()
+    if pre_calibrate:
+        est_dur += 60  # add some extra time for the pre-calibrations
+        # TODO replace with call to estimate calibration time in module for each calibrations
+        #  when code here replaced with calls to calibration manager
+    f = model.ProgressiveFuture(start=time.time(), end=time.time() + est_dur)
```
Sabrina85, line 232: The time estimate is currently overridden by line 1621 in gui/cont/acquisition.py. I don't have a good idea how to handle this yet. The ROA, which we get the time estimate from, does not know about whether pre calibrations need to be run or not.

tepals: You could add `pre_calibrate` as an attribute to the ROA class, and then update `roa.estimate_acquisition_time()` to add extra time when `pre_calibrate` is True. Then you could remove it as argument from `acquire` and `AcquisitionTask`, line 345 would then become `if self._roa.pre_calibrate:`

tepals: you will then also need to update `test_pre_calibrate`

Sabrina85: Yes that would make sense. I'll wait for the review from Éric whether I now have put setting the time in the right places and then I will change this.

src/odemis/acq/fastem.py
```diff
@@ -439,6 +450,8 @@ def pre_calibrate(self):
         if not fastem_calibrations:
             raise ModuleNotFoundError("Need fastem_calibrations repository to run pre-calibrations.")
 
+        asm_config_orig = None
+
```
pieleric, line 454: What's the point of initializing it to `None`? Is it so that even if get_config_asm() fails, the `finally` block doesn't fail? However, configure_asm() needs a dict properly field up, so passing None will fail anyway...

src/odemis/acq/fastem.py
pieleric: That sounds like a good plan to me too.

pieleric (changes requested):

src/odemis/acq/fastem.py
Sabrina85: Correct. I have an extension for the `configure_asm()` case ready. I will push a new commit for this here.

Sabrina85: Sorry, I will make a pr in the fastem-calibration repo for this not here.

tepals (changes requested):
One small change needed

Sabrina85:
> One small change needed

Which one? I don't see any comment?

src/odemis/acq/fastem.py
```diff
@@ -207,7 +214,7 @@ def __init__(self, name, coordinates):
         # if None -> acquire, if not None, don't acquire again
 
 
-def acquire(roa, path, scanner, multibeam, descanner, detector, stage, ccd, beamshift, lens, pre_calibrate=False):
+def acquire(roa, path, scanner, multibeam, descanner, detector, stage, ccd, beamshift, lens):
```
tepals, line 217: Since you remove it here you also need to remove it where this function is called, in fastem_acq on line 461.

tepals:
> > One small change needed
> 
> Which one? I don't see any comment?

it disappeared, I added it again

src/odemis/acq/fastem.py
Sabrina85: Ah that got added during the rebase again 🙈

tepals (approved):

pieleric (approved):

PR #2058, opened 2022-03-10
gui: improve overview acquisition and display on Cryo

pieleric:


src/odemis/gui/comp/miccanvas.py
```diff
@@ -650,6 +650,36 @@ def fit_view_to_content(self, recenter=None):
...
                 self.view.view_pos.value = self.requested_phys_pos
 
+    def fit_to_bbox(self, bbox):
+        """
+        Zoom in to the bounding box and recenter. Same as fit_view_to_content, but with explicit bounding box,
+        so we can zoom to a specific position.
+        bbox (None or 4 floats): bounding box to be shown, defined as minx, miny,
+           maxx, maxy positions in m.
+        """
+        if bbox[0] is None:
```
Kleijwegt, line 660: bbox seems to be defined as a tuple as 4 None's at first.
Shouldn't it be checked if all the values are not None?

pieleric: Good catch. Actually for this function, I think it makes sense to just assume we pass a correct bbox. Just I'll simplify the code.

src/odemis/gui/comp/miccanvas.py
```diff
@@ -650,6 +650,36 @@ def fit_view_to_content(self, recenter=None):
...
+            return  # no image => nothing to do
+
+        # compute mpp so that the bbox fits exactly the visible part
+        w, h = abs(bbox[2] - bbox[0]), abs(bbox[3] - bbox[1])  # m
+        if w == 0 or h == 0:
+            logging.warning("Weird image size of %fx%f m", w, h)
```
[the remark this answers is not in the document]
pieleric: That's copy-paste from the `fit_to_content()`. Here, we could recenter anyway, and also probably use the other dimension (if not both as 0). So I'm just going to change the code and convert 0 to "a very tiny value".

Kleijwegt (approved):

PR #2059, opened 2022-03-10
Fastem GUI: Added dwell time slider to control the multibeam dwell ti…

tepals:
…me in the acquisition panel.

### DIFF
--- a/src/odemis/gui/cont/tabs.py
+++ b/src/odemis/gui/cont/tabs.py
@@ -1799,11 +1799,11 @@ class FastEMAcquisitionTab(Tab):
         # Controller for acquistion settings panel
         self._acq_settings_controller = settings.SettingsController(
             panel.pnl_acq_settings,
-            "dwell time"
+            ""  # default message, which is shown if no setting is available
         )
-        dt_conf = {"range": [1e-6, 40e-6]}  # Limit the values the user can set the dwell time to.
+        dt_conf = get_hw_config(tab_data.main.multibeam, tab_data.main.hw_settings_config).get("dwellTime")
         self._acq_settings_controller.add_setting_entry(
-            "dwell time", tab_data.main.multibeam.dwellTime, tab_data.main.multibeam, dt_conf)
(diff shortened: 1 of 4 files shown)


Sabrina85 (approved):
Looks good! Really happy that you found this simple way to implement it :)

src/odemis/gui/cont/tabs.py
```diff
@@ -1796,6 +1796,15 @@ def __init__(self, name, button, panel, main_frame, main_data):
             calibrations
         )
 
+        # Controller for acquistion settings panel
+        self._acq_settings_controller = settings.SettingsController(
+            panel.pnl_acq_settings,
+            "dwell time"
```
pieleric, line 1802: That's the "default message", which is shown if no setting is available. `""` would be better here.

src/odemis/gui/cont/tabs.py
```diff
@@ -1796,6 +1796,15 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
+        )
+        dt_conf = {"range": [1e-6, 40e-6]}  # Limit the values the user can set the dwell time to.
```
pieleric, line 1804: Instead of putting this value here, I'd suggest to put it with the rest, in gui/conf/data.py/HW_SETTINGS_CONFIG["multibeam"]["dwellTime"]. 
Then, you'd access it with `get_hw_config(tab_data.main.multibeam).get("dwellTime")`. 
It's a little bit more verbose, but allows to keep that setting config in the same place as the others.

src/odemis/gui/cont/tabs.py
```diff
@@ -1796,6 +1796,15 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
+        self._acq_settings_controller.add_setting_entry(
+            "dwell time", tab_data.main.multibeam.dwellTime, tab_data.main.multibeam, dt_conf)
```
pieleric, line 1806: The first parameter is the name of the VA, which should be just the string "dwellTime". It will be converted automatically to something nice looking by the function.

pieleric (changes requested):

pieleric (approved):

PR #2060, opened 2022-03-10
fastem gui controller: refactor and move fastem specific controllers to own module

Sabrina85:
* rename alignmentController to calibration controller
* rename previous calibrationcontroller to calibrationsregionscontroller


src/odemis/gui/cont/project.py
```diff
@@ -0,0 +1,431 @@
...
+        cnvs.fit_to_bbox([l - 5 * sz[0], t - 5 * sz[1], r + 5 * sz[0], b + 5 * sz[1]])
+
+    def _on_coordinates(self, coords):
+        # Purely for logging
+        logging.debug("ROC '%s' coordinates changed to %s.", self.model.name.value, coords)
+
+
+class FastEMCalibrationRegionsController(object):
+    """
+    Listens to the calibration buttons and creates the FastEMROCControllers accordingly.
```
tepals, line 384: I know this is just copied from when it was still in streams.py, but the term "calibration buttons" is a bit vague. Maybe there it's also better to call them "calibration region buttons", especially now that we also have the Calibrate buttons.

Sabrina85: Would it be ok to do this in the next pr? I have might need to modify more and/or can make the doc strings etc also a bit more explanatory?

tepals: no problem!

tepals (approved):

pieleric (approved):

PR #2062, opened 2022-03-10
gui: show z pos feature in localization tab as meters with more digits

pieleric:
Need to show more digits, as the first digits are usually the same ones.
Also show in m, to get a nicer value (to avoid 0.00...).
However, this requires rellying properly on the UnitFloatCtrl value
reading. Howevever, this causes extra event order issues... which needs
to be worked around too.


src/odemis/gui/cont/features.py
```diff
@@ -189,9 +191,14 @@ def _on_ctrl_feature_z_change(self):
         Get the current feature Z ctrl value to set feature Z position
         :return: (tuple of 3 floats) the full position including the Z ctrl value
         """
+        # HACK: sometimes the event is first received by this handler and later
+        # by the UnitFloatCtrl. So the value is not yet computed => Force it, just in case.
+        self._panel.ctrl_feature_z.on_text_enter(None)
+        zpos = self._panel.ctrl_feature_z.GetValue()
```
Kleijwegt, line 197: I assume you checked this but is it correct that the call to this variable/attribute changed from Value --> GetValue()?

Kleijwegt (approved):

src/odemis/gui/cont/features.py
pieleric: Yes, that's on purpose. Because on this control .Value != .GetValue().

PR #2065, opened 2022-03-11
driver kryoz test: don't run when no hardware

pieleric:
The driver doesn't really work yet anyway... and especially there is no
simulator, so let's not try to run it when no hardware is available.


tepals (approved):
Looks good!

PR #2074, opened 2022-03-15
driver piezomotor: fix simulator to give the correct response for clo…

tepals:
…sed and open loop mode.

With the changes made in commit 4b74088751077d3286ec260f1e141fed24db7c8e isMovingClosedLoop now specifically checks whether closed loop mode is active.
The simulator did not know about the closed or open loop state and always returned the open loop status.
This commit fixes this and checks whether closed loop is active and if yes returns a status based on that.

### DIFF
--- a/src/odemis/driver/piezomotor.py
+++ b/src/odemis/driver/piezomotor.py
@@ -180,9 +180,6 @@ class PMD401Bus(Actuator):
         self._port = None  # port number
         min_axis = min(self._axis_map.values())
         self._serial = self._findDevice(port, min_axis)
-        # To be able to give the correct response the simulator needs to know which axes have closed loop active.
-        if port.startswith("/dev/fake"):
-            self._serial.closed_loop = {self._axis_map[k]: v for (k, v) in self._closed_loop.items()}
         self._recovering = False
 
         # Get version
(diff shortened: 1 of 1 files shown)


src/odemis/driver/piezomotor.py
```diff
@@ -180,6 +180,9 @@ def __init__(self, name, role, port, axes, inverted=None, param_file=None, **kwa
         self._port = None  # port number
         min_axis = min(self._axis_map.values())
         self._serial = self._findDevice(port, min_axis)
+        # To be able to give the correct response the simulator needs to know which axes have closed loop active.
+        if port.startswith("/dev/fake"):
+            self._serial.closed_loop = {self._axis_map[k]: v for (k, v) in self._closed_loop.items()}
```
pieleric, line 185: No. The real hardware doesn't need this info, so there is no reason the simulator needs it. Please, make the simulator behave like the real hardware: based on the commands it receives it switches between closed-loop and open-loop. I think for the simulator it should be enough to have this behaviour (not sure if the real hardware has extra rules):
 * runAbsTargetMove() and runRelTargetMove() set closed_loop to True.
 * stopAxis sets the closed_loop to False

pieleric (changes requested):

PR #2075, opened 2022-03-18
Added MD_FIELD_FREE_SHIFT to configure MS_POS_COR of the scanner base…

tepals:
…d on the status of the immersion mode VA.

MD_POS_COR shifts the single beam image. When immersion is off (field free on) shift the overview image by a calibrated factor with MD_POS_COR, then set back MD_POS_COR to [0, 0] during live stream imaging (immersion on).
This way the overview image is shifted compared to the live stream, since the live stream is somewhat calibrated with the megafield this also calibrates the overview image with the megafield.

### DIFF
--- a/src/odemis/acq/fastem_conf.py
+++ b/src/odemis/acq/fastem_conf.py
@@ -91,7 +91,7 @@ def configure_scanner(scanner, mode):
     # Immersion needs to be set before changing the horizontalFoV, as the range is updated
     scanner.immersion.value = conf["immersion"]
 
-    if scanner.immersion.value is False:
+    if scanner.immersion.value:
         # When the scanner is in immersion mode set the position correction to [0, 0]
         scanner.updateMetadata({model.MD_POS_COR: [0, 0]})
     else:
--- a/src/odemis/acq/fastem_conf.py
+++ b/src/odemis/acq/fastem_conf.py
@@ -91,6 +91,15 @@ def configure_scanner(scanner, mode):
     # Immersion needs to be set before changing the horizontalFoV, as the range is updated
     scanner.immersion.value = conf["immersion"]
 
+    if scanner.immersion.value is False:
+        # When the scanner is in immersion mode set the position correction to [0, 0]
+        scanner.updateMetadata({model.MD_POS_COR: [0, 0]})
+    else:
+        # Correct for shift in image between immersion mode and field free mode.
+        # If MD_FIELD_FREE_SHIFT is not provided fall back to a correction of [0, 0]
+        pos_cor = scanner.getMetadata().get(model.MD_FIELD_FREE_POS_SHIFT, [0, 0])
+        scanner.updateMetadata({model.MD_POS_COR: pos_cor})
+
     if "horizontalFoV" in conf:
         scanner.horizontalFoV.value = conf["horizontalFoV"]  # m
     else:
--- a/src/odemis/acq/test/fastem_conf_test.py
+++ b/src/odemis/acq/test/fastem_conf_test.py
@@ -123,6 +123,9 @@ class TestFASTEMConfig(unittest.TestCase):
         # check that rotation is the same as was specified for the scanner (e-beam) for megafield imaging
         self.assertEqual(self.multibeam.getMetadata()[model.MD_ROTATION], math.radians(10))
 
+        # check that the MD_POS_COR is set to [0, 0] for live stream imaging.
+        self.assertListEqual([0, 0], self.scanner.getMetadata()[model.MD_POS_COR])
+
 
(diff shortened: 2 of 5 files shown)


src/odemis/acq/fastem_conf.py
```diff
@@ -91,6 +91,15 @@ def configure_scanner(scanner, mode):
     # Immersion needs to be set before changing the horizontalFoV, as the range is updated
     scanner.immersion.value = conf["immersion"]
 
+    if scanner.immersion.value is False:
+        # Shift the overview image by a calibrated value, to align it with the live view.
```
Sabrina85, line 95: The comment is very specific for overview and live mode. However, here you just apply a shift in case the immersion mode is switched off (=field free mode). I suggest to rephrase in order to be more generic:
'Correct for shift in image between immersion mode and field free mode.'

Sabrina85 (changes requested):

src/odemis/acq/fastem_conf.py
```diff
@@ -91,6 +91,15 @@ def configure_scanner(scanner, mode):
     # Immersion needs to be set before changing the horizontalFoV, as the range is updated
     scanner.immersion.value = conf["immersion"]
 
+    if scanner.immersion.value is False:
```
pieleric, line 94: I think it'd simpler to read if you had `if scanner.immersion.value: ... else: ...`

src/odemis/acq/fastem_conf.py
```diff
@@ -91,6 +91,15 @@ def configure_scanner(scanner, mode):
...
+        # Correct for shift in image between immersion mode and field free mode.
+        # If MD_FIELD_FREE_SHIFT is not provided fall back to a correction of [0, 0]
+        pos_cor = scanner.getMetadata().get(model.MD_FIELD_FREE_SHIFT, [0, 0])
```
pieleric, line 97: A good name for this metadata is a bit tricky to find. It'd be nice that it's clearly related to `POS_COR` and also to "immersion". `FIELD_FREE_POS` is odd (as it'd sound like an absolute value). `FIELD_FREE_POS_COR` is odd too. But I feel that `FIELD_FREE_SHIFT` doesn't clearly indicate it's a shift in *position* . How about `FIELD_FREE_POS_SHIFT`?

pieleric (changes requested):

src/odemis/acq/fastem_conf.py
pieleric: That should now just be: `if scanner.immersion.value` (no `is False`).

pieleric (approved):

pieleric (changes requested):
BTW, I think there should be some test cases to extend.
Also, you could update the microscope file with a value for `MD_FIELD_FREE_POS_SHIFT`.

src/odemis/acq/test/fastem_conf_test.py
```diff
@@ -95,6 +101,10 @@ def test_configure_scanner_live(self):
         self.assertTrue(self.scanner.immersion.value)
         self.assertEqual(self.scanner.rotation.value, math.radians(5))
 
+        scanner_md = self.scanner.getMetadata()
+        # check that the MD_POS_COR is set to [0, 0] for live stream imaging.
+        self.assertListEqual([0, 0], scanner_md[model.MD_POS_COR])
+
```
Sabrina85, line 107: I think this test should also be part in the `configure_scanner_megafield` test to make sure that it is also set correctly in this mode.

src/odemis/acq/test/fastem_conf_test.py
```diff
@@ -74,8 +75,13 @@ def test_configure_scanner_overview(self):
         self.assertGreater(self.scanner.horizontalFoV.value, 1.e-3)  # should be big FoV for overview
         self.assertEqual(self.scanner.rotation.value, math.radians(5))
 
+        scanner_md = self.scanner.getMetadata()
+
+        # check that the MD_POS_COR is correctly set for overview imaging.
```
Sabrina85, line 80: Another test you could add is that the fallback is correctly working in case there is no MD provided. But maybe not absolutely necessary.

Sabrina85 (changes requested):
Looks fine. Only some small comment on the tests. Did both yaml files start properly?

src/odemis/acq/test/fastem_conf_test.py
tepals: I'm not able to remove the MD_FIELD_FREE_POS_SHIFT from the metadata, so I don't think there is a way to test this :(.

Sabrina85 (approved):

PR #2076, opened 2022-03-18
Super Z localisation function for Enzel/Meteor using astigmatism. Uses the jupyter notebook to calculate the calibration parameters.

Kleijwegt:


src/odemis/acq/align/test/meteor_super_z_test.py
```diff
@@ -0,0 +1,69 @@
...
+    def setUpClass(cls):
+        z_stack_step_size = 50
+        cls.precision = z_stack_step_size * 0.45  # Precision should be better than the step within a z stack
+
+    def test_determine_z_position(self):
+        """
+        Test for known data the outcome of the function determine_z_position
+        """
+        # Test on an image below focus
+        image = open_data("images/super_z_single_beed_aprox_500nm_under_focus.tif").content[0].getData()
```
pieleric, line 28: Note that for doing such simple thing, you can use the simpler API:
`image = read_data("images/super_z_single_beed_aprox_500nm_under_focus.tif")[0]`

pieleric, line 28: Note that for doing such simple thing, you can use the simpler API:
`image = read_data("images/super_z_single_beed_aprox_500nm_under_focus.tif")[0]`

src/odemis/acq/align/test/meteor_super_z_test.py
```diff
@@ -0,0 +1,69 @@
...
+        expected_outcome_image_1 = -592.5e-9  # Value determined using the function determine_z_position
+        z, warn_flag = determine_z_position(image, CALIB_DATA)
+        self.assertEqual(warn_flag, 0)
+        self.assertTrue(almost_equal(expected_outcome_image_1, z, atol=self.precision))
```
pieleric, line 32: You can write directly `self.assertAlmostEqual(expected_outcome_image_1, z, delta=self.precision)`

pieleric (changes requested):

src/odemis/acq/align/meteor_super_z.py
```diff
@@ -0,0 +1,155 @@
+import logging
```
pieleric, line 1: That file name "meteor_super_z" is a little too specific. It should be working also on the ENZEL and MIMAS. In addition, in the rest of the GUI, we've used less commercial terms, like "Z localization". I'd suggest to rename it "z_localization".

src/odemis/acq/align/meteor_super_z.py
```diff
@@ -0,0 +1,155 @@
...
+    :param z_stack_step_size (float): the step size in meters between two images in the z stack
+    :param fit_tol (float): factor for the precision of the Gaussian fit, range from 0 --> 1
+    :return:
+        z_position (float): determined z position of the feature in meter
+        warn_lvl (int): 0 = Success
+                        1 = from fmin_cg(scipy) max number of iterations exceeded
+                        2 = from fmin_cg(scipy) gradient and/or function calls were not changing, 3 = NaN result encountered
+                        4 = The Gaussian fit is not precise enough, probably because the image contains too much noise
+                        5 = The Gaussian fit found a feature to big for the current feature, the size > 85%
+                        6 = Outputted Z position is outside the defined maximum range, output is inaccurate
```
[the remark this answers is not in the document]
Kleijwegt: 0, 1, 2, and 3 are also raise by fmin_cg via warn_flag_int. But I could maybe comment a bit on that.

src/odemis/acq/align/meteor_super_z.py
Kleijwegt, line 1: Add author and delmic top part with also Daan Boltje as author

src/odemis/acq/align/meteor_super_z.py
```diff
@@ -0,0 +1,155 @@
...
+    warn_lvl = 0  # Set the warning level to zero, no warnings
+
+    # Fit a Gaussian to the feature in the image
+    popt, pcov = fit_gaussian_2D(image, theta=numpy.rad2deg(calibration_data["feature_angle"]), epsilon=1e-6)
+    sigma_x, sigma_y = popt[2:4]
+
+    # Ensure that for a normal distribution at least 95% falls within the set tollarance for the calculated standard
+    # deviation error (this means a confidence level of 1.96 time the standard deviation above and below the mean)
+    if (fit_tol / 0.98) * sigma_x < pcov[2] or (fit_tol / 0.98) * sigma_y < pcov[3]:
+        logging.warning("The image contains to much noise to determine the size of the feature accurately, "
```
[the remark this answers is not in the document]
Kleijwegt: Like this? Or even more info? I find it hard to estimate how much statistical info a user would like to have.

```logging.warning("The image contains too much noise to determine the size of the feature accurately, "
                        "the results may be inaccurate.\n"
                        "The standard deviation error for determined feature size in x is: %s and for y is %s.",
                        pcov[2], pcov[3])```

File: src/odemis/acq/align/z_localization.py [hunk 6]
```diff
@@ -0,0 +1,193 @@
...
+            KeyError if the calibration data is incomplete and does not include all the keys
+
+    """
+    warning = None  # Set the warning level to None, no warnings
+
+    if not psf_extractor:
+        raise ModuleNotFoundError("The psf_extractor module is not found. Cannot determine the Z position")
+
+    # Fit a Gaussian to the feature in the image
+    popt, pcov = psf_extractor.extractor.fit_gaussian_2D(image, theta=numpy.rad2deg(calibration_data["feature_angle"]), epsilon=1e-6)
```
pieleric, line 141: For single values, please use the standard Python function: `math.degrees(calibration_data["feature_angle"])`

src/odemis/acq/align/z_localization.py
```diff
@@ -0,0 +1,193 @@
...
+                                          args=(sigma_x, sigma_y, calibration_data, huang),
+                                          maxiter=2000, disp=False, full_output=True,
+                                          retall=True,
+                                          )
+
+    if warn_flag > 0:
+        logging.warning(f"Inaccuracy observed during when determining the Z position, the warning flag {warn_flag}"
+                        f"was raised. Current results may be inaccurate")
+        # fmin_cg output the warning flags 1, 2 and 3. The warnings of fmin_cg may be te result of previously found
+        # warning, therefore these are considered more important
```
pieleric, line 181: If you want `warning` to be set to the value of `warn_flag`, you have to explicitly do it (here).

src/odemis/acq/align/z_localization.py
```diff
@@ -0,0 +1,193 @@
...
+    popt, pcov = psf_extractor.extractor.fit_gaussian_2D(image, theta=numpy.rad2deg(calibration_data["feature_angle"]), epsilon=1e-6)
+    sigma_x, sigma_y = popt[2:4]
+
+    # Ensure that for a normal distribution at least 95% falls within the set tolerance for the calculated standard
+    # deviation error (this means a confidence level of 1.96 time the standard deviation above and below the mean)
+    if (fit_tol / 0.98) * sigma_x < pcov[2] or (fit_tol / 0.98) * sigma_y < pcov[3]:
+        logging.warning("The image contains too much noise to determine the size of the feature accurately, "
+                        "the results may be inaccurate.\n"
+                        "The standard deviation error for determined feature size in x is: %s and for y is %s.",
+                        pcov[2], pcov[3])
```
pieleric, line 150: This is fine. However, I don't think it's always worth to be very verbose. Honestly, the pcov values are only for us, when a user reports a problem that the "Z localization didn't work". So you could have kept it short like `std dev = {pcov[2]}, {pcov[3]}.`  Anyway, that's fine as it is.

src/odemis/acq/align/z_localization.py
```diff
@@ -0,0 +1,193 @@
...
+        # fmin_cg output the warning flags 1, 2 and 3. The warnings of fmin_cg may be te result of previously found
+        # warning, therefore these are considered more important
+        if not(warning):
+            warning = 4
+
+    z_position = round(zopt[0]*1e-9 - calibration_data["z_least_confusion"], 20)
+
+    if not(calibration_data["z_calibration_range"][0] < z_position < calibration_data["z_calibration_range"][1]):
+        # Always log this warning but only update this error if no other cause of the error is found.
+        logging.warning("The determined z position is out of the specified max_range."
```
pieleric, line 189: Please log the z position next to "z position" and  the "max_range" value next to "max_range". It'll be easier to debug.

pieleric (changes requested):

src/odemis/acq/align/z_localization.py
```diff
@@ -0,0 +1,193 @@
...
+    """
+    a = calibration_data["a"]
+    b = calibration_data["b"]
+    c = calibration_data["c"]
+    d = calibration_data["d"]
+    w_0 = calibration_data["w0"]
+    return w_0 * numpy.sqrt(
+                            1 + numpy.power(numpy.divide(z - c, d), 2) +
+                            a * numpy.power(numpy.divide(z - c, d), 3) +
+                            b * numpy.power(numpy.divide(z - c, d), 4)
```
pieleric, line 59: Is there an advantage in forcing numpy functions? Normally, all the operators are automatically switched to numpy if there is an array involved.

src/odemis/acq/align/z_localization.py
```diff
@@ -0,0 +1,193 @@
...
+    psf_extractor = None
+
+
+def huang(z, calibration_data):
+    """
+    Function for the expected features size in x/y direction in terms of z position for a certain degree of astigmatism.
+    This formula is based on the Huang algorithm as referenced below. The calibration data must be
+    made on this degree of astigmatism.
+
+    :param z (float): z position of the feature
```
pieleric, line 43: There is part of the code where you pass a numpy array instead of a single value. Please indicate that this is allowed.

src/odemis/acq/align/z_localization.py
```diff
@@ -0,0 +1,193 @@
...
+                        f"Using an image with more space around the feature might solve this problem."
+                        f"The detected feature size is {sigma_x} in x and {sigma_y} in y direction, while the total image has a "
+                        f"width of {image.shape[0]} pixels and a height of {image.shape[1]} pixels."
+                        f"The feature size should be < 85% in each direction\n"
+                        f"Current results may be inaccurate.")
+        warning = 6
+
+    # Determine the z position using the shape of the features (sigma_x/sigma_y)
+    max_range = calibration_data["z_calibration_range"][1] - calibration_data["z_calibration_range"][0]
+    fine_z = numpy.arange(0, max_range, z_stack_step_size*1e9)
```
pieleric, line 162: Instead of relying on the z_stack_step_size, maybe you could just arbitrarily take 10 or 20 values between 0 and max_range (ie, using linspace)

pieleric (changes requested):

src/odemis/acq/align/z_localization.py
Kleijwegt: I would say looking at the step size is better. Otherwise the precision of your initial guess is dependent on your max_range. That feels a bit odd.
Though the 'random value" I like. Maybe I was already thinking about removing the z_stack_step_size argument. Maybe I should just hardcode it and comment on it?

src/odemis/acq/align/z_localization.py
```diff
@@ -0,0 +1,193 @@
...
+                                    following keys:
+                                    x (dict) --> a, b ,c, d, w0 (floats), fit on the equation of Huang in x direction
+                                    y (dict) --> a, b ,c, d, w0 (floats), fit on the equation of Huang in y direction
+                                    feature_angle (float) angle of the ellipsoidal shapes w.r.t. the positive Y axis (anti-clockwise is positive)
+                                    upsample_factor (int) number with which the data is up sampled during the calibration
+                                    z_least_confusion (float) location in z in meters where the least confusion is present in the image (in focus)
+                                    z_calibration_range (tuple) --> (min, max) floats of the min and max z value in meters w.r.t. the z_least_confusion
+
+    :param z_stack_step_size (float): the step size in meters between two images in the z stack
+    :param fit_tol (float): factor for the precision of the Gaussian fit, range from 0 --> 1
```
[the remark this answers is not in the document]
Kleijwegt: It indeed affects the warning. The docstring+comments needs to be indeed slightly better. But it is quite hard to wrap your/my head around. Maybe we can discuss to see what is the smartest thing to do. We can then also discuss the 0.98 * 2 --> 1.96 statistical thing from this conversation: https://github.com/delmic/odemis/pull/2076#discussion_r834222193

src/odemis/acq/align/meteor_super_z.py
```diff
@@ -0,0 +1,155 @@
...
+
+try:
+    from skimage import io, exposure
+    from scipy.optimize import fmin_cg
+    from psf_extractor.extractor import fit_gaussian_2D
+
+    PSF_Extractor_module = True
+except ImportError:
+    logging.warning(
+        "PSF_Extractor module and required libraries not found. The function determine_z_position will not work.")
```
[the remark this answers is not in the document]
Kleijwegt: You mean like this isn't it?

```
except ImportError as exp:
    logging.warning("psf_extractor module and required libraries cannot be loaded."
                    "The function determine_z_position will not work.\n"
                    "The following exception was raised: \n %s", exp)```

File: src/odemis/acq/align/z_localization.py [hunk 15]
```diff
@@ -0,0 +1,193 @@
...
+    :param calibration_data (dict): contains the data from the calibration performed using the jupyter notebook with the
+                                    following keys:
+                                    x (dict) --> a, b ,c, d, w0 (floats), fit on the equation of Huang in x direction
+                                    y (dict) --> a, b ,c, d, w0 (floats), fit on the equation of Huang in y direction
+                                    feature_angle (float) angle of the ellipsoidal shapes w.r.t. the positive Y axis (anti-clockwise is positive)
+                                    upsample_factor (int) number with which the data is up sampled during the calibration
+                                    z_least_confusion (float) location in z in meters where the least confusion is present in the image (in focus)
+                                    z_calibration_range (tuple) --> (min, max) floats of the min and max z value in meters w.r.t. the z_least_confusion
+
+    :param z_stack_step_size (float): the step size in meters between two images in the z stack
```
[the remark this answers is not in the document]
Kleijwegt: Link the answer in this conversation:
https://github.com/delmic/odemis/pull/2076#discussion_r836118719

> I would say looking at the step size is better. Otherwise the precision of your initial guess is dependent on your max_range. That feels a bit odd.
Though the 'random value" I like. Maybe I was already thinking about removing the z_stack_step_size keyword argument. Maybe I should just hardcode it and comment on it?

src/odemis/acq/align/z_localization.py
pieleric: Well, my understanding of this code is that it just tries many values between 0 and max_range, and pick the one that seems to fit best. Currently it takes as many trial values as there was stacks in the calibration zstack. That's quite arbitrary too. If you hard-code the step size, it's probably going to work in the standard cases, because max_range is usually in the same order of magnitude anyway, but that could break in some extreme cases where max_range is particularly small. I think that just hard coding the number of values to try is the way to go.

pieleric (approved):

PR #2079, opened 2022-03-18
plugins: new update_rot plugin to connect ebeam rotation VA to metadata

pieleric:
Currently, when the SEM is scanned via a external card, but the controls
can be read/written via an API, the .rotation VA shows the rotation, but
the data doesn't have this information in the metadata.

That means that the image is always shown as if rotation is 0°. In many
cases that's fine. However, on the SECOM, that's not enough because the
fine alignment overlay looks at .rotation and assumes that MD_ROTATION
will be set. To compensate for it, it sets MD_ROTATION_COR on the
optical images. This causes the optical images to be incorrectly
rotated!

=> copy .rotation as MD_ROTATION & MD_ROTATION_COR. So the SEM image is
not rotated, and the fine alignement is aware of it, so it doesn't
rotate the optical image.

### DIFF
--- a/plugins/update_rot.py
+++ b/plugins/update_rot.py
@@ -1,0 +1,70 @@
+# -*- coding: utf-8 -*-
+'''
+Created on 18 Mar 2022
+
+@author: Éric Piel
+
+Sets the ebeam rotation metadata to the same value as the .rotation VA, so that
+the image is shown non-rotated, but that the fine alignment knows about it, and
+does not get confused (by rotating the optical image).
+
+This is free and unencumbered software released into the public domain.
+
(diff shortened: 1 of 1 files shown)


Kleijwegt (changes requested):
Even though this file is quite simple I think it is good to add unit tests and doc strings.

pieleric:
I've added a docstring to the `on_rotation()` method, and also added code to handle the case there is no backend. 
I didn't add test cases, mostly because we don't have any test cases for the plugins, and it's not super easy to test (as you'd need to load the plugin). Maybe one day we could add this, but let's not do that now...

Kleijwegt (approved):

PR #2080, opened 2022-03-21
fastem calibration 2: dark/gain calibration per scintillator

Sabrina85:
* add dark/gain to calibration manager
* make controller independent of each other
* make calibration controller using a prefix argument to be able to use same class for all calibration steps
* make the gauge having a fixed size in the GUI
* use short version for text label showing the progress
* fix: ROC size should be calculated based on multibeam resolution and pixelsize
* allow adding/removing ROCs
* synchronize gui with main model (active scintillators and calibration regions defined)
* remove unnecessary properties in tabs

### DIFF
--- a/src/odemis/acq/align/fastem.py
+++ b/src/odemis/acq/align/fastem.py
@@ -37,7 +37,9 @@ try:
         descan_gain,
         image_rotation_pre_align,
         image_rotation,
-        image_translation
+        image_translation,
+        dark_offset_correction,
+        digital_gain_correction
     )
     from fastem_calibrations.configure_hw import (
         get_config_asm,
@@ -73,9 +75,11 @@ class Calibrations(Enum):
     IMAGE_TRANSLATION_PREALIGN = image_translation_pre_align
     IMAGE_ROTATION_FINAL = image_rotation
     IMAGE_TRANSLATION_FINAL = image_translation
+    DARK_OFFSET = dark_offset_correction
+    DIGITAL_GAIN = digital_gain_correction
 
 
-def align(scanner, multibeam, descanner, detector, stage, ccd, beamshift, det_rotator, calibrations):
+def align(scanner, multibeam, descanner, detector, stage, ccd, beamshift, det_rotator, calibrations, stage_pos=None):
     """
     Start a calibration task for a given list of calibrations.
 
@@ -89,6 +93,8 @@ def align(scanner, multibeam, descanner, detector, stage, ccd, beamshift, det_ro
     :param beamshift: (tfsbc.BeamShiftController) Component that controls the beamshift deflection.
     :param det_rotator: (actuator) K-mirror controller. Must have a rotational (rz) axis.
     :param calibrations: (list[Calibrations]) List of calibrations that should be run.
+    :param stage_pos: (float, float) Stage position where the calibration should be run. If None,
+                      the calibration is run at the current stage position.
 
     :returns: (ProgressiveFuture) Alignment future object, which can be cancelled.
     """
@@ -100,7 +106,8 @@ def align(scanner, multibeam, descanner, detector, stage, ccd, beamshift, det_ro
     f = model.ProgressiveFuture(start=time.time(), end=time.time() + est_dur)
 
     # Create a task that runs the calibration and alignments.
-    task = CalibrationTask(f, scanner, multibeam, descanner, detector, stage, ccd, beamshift, det_rotator, calibrations)
+    task = CalibrationTask(f, scanner, multibeam, descanner, detector, stage, ccd, beamshift, det_rotator,
+                           calibrations, stage_pos)
 
     f.task_canceller = task.cancel  # lets the future know how to cancel the task.
 
@@ -127,7 +134,7 @@ class CalibrationTask(object):
     """
 
     def __init__(self, future, scanner, multibeam, descanner, detector, stage, ccd, beamshift, det_rotator,
-                 calibrations):
+                 calibrations, stage_pos):
         """
         :param future: (ProgressiveFuture) Acquisition future object, which can be cancelled.
                        (Exception or None): Exception raised during the calibration or None.
@@ -142,6 +149,8 @@ class CalibrationTask(object):
         :param beamshift: (tfsbc.BeamShiftController) Component that controls the beamshift deflection.
         :param det_rotator: (actuator) K-mirror controller. Must have a rotational (rz) axis.
         :param calibrations: (list[Calibrations]) List of calibrations that should be run.
+        :param stage_pos: (float, float) Stage position where the calibration should be run. If None,
+                      the calibration is run at the current stage position.
         """
         self.asm_config = None
         self._scanner = scanner
@@ -156,6 +165,7 @@ class CalibrationTask(object):
         self._future = future
 
         self.calibrations = calibrations
+        self.stage_pos = stage_pos
 
         # List of calibrations to be executed. Used for progress update.
         self._calibrations_remaining = set(calibrations)
@@ -165,9 +175,27 @@ class CalibrationTask(object):
 
     def run(self):
         """
-        Runs a set of calibration procedures.
+        Runs a set of calibration procedures and return the calibrated settings.
         :returns:
-            (None) Calibrations successful.
+            self.asm_config: (nested dict) A dictionary containing factory and/or calibrated settings. Settings
+            that are calibrated or are overwriting factory settings. The content of the dict is:
+            multibeam:
+                scanOffset: (tuple) The x and y start of the scanning movement (start of scan ramp) of the multibeam
+                            scanner in arbitrary units.
+                scanAmplitude: (tuple) The x and y heights of the scan ramp of the multibeam scanner in arbitrary units.
+                dwellTime: (float) The acquisition time for one pixel within a cell image in seconds.
+                resolution: (tuple) The effective resolution of a single field image excluding overscanned pixels
+                                    in pixels.
+            descanner:
+                scanOffset: (tuple) The x and y start of the scanning movement (start of scan ramp) of the descanner
+                            in arbitrary units.
+                scanAmplitude: (tuple) The x and y heights of the scan ramp of the descanner in arbitrary units.
+            mppc:
+                cellCompleteResolution: (tuple) The resolution of a cell image including overscanned pixels in pixels.
+                cellTranslation: (tuple of tuples of shape mppc.shape) The origin for each cell image within the
+                                 overscanned cell image in pixels.
+                cellDarkOffset: (tuple of tuples of shape mppc.shape) The dark offset correction for each cell image.
+                cellDigitalGain: (tuple of tuples of shape mppc.shape) The digital gain correction for each cell image.
         :raise:
             Exception: If a calibration failed.
             CancelledError: If the calibration was cancelled.
@@ -186,6 +214,11 @@ class CalibrationTask(object):
 
             # reset beamshift
             self._beamshift.shift.value = (0, 0)
+
+            if self.stage_pos:
+                # move to region of calibration (ROC) position
+                sf = self._stage.moveAbs({'x': self.stage_pos[0], 'y': self.stage_pos[1]})
+                sf.result()  # wait until stage is at correct position
 
             # loop over calibrations in list (order in list is important!)
             for calib in self.calibrations:
@@ -227,10 +260,11 @@ class CalibrationTask(object):
                 configure_asm(self._multibeam, self._descanner, self._detector, self._dataflow, self.asm_config)
             logging.debug("Calibrations finished.")
 
+        return self.asm_config
+
     def run_calibrations(self, calibration):
         """
         Run a calibration.
-        Note: All calibrations can be run on bare scintillator.
         """
         if calibration == Calibrations.OPTICAL_AUTOFOCUS:
             autofocus_multiprobe.run_autofocus(self._scanner, self._multibeam, self._descanner, self._detector,
@@ -287,6 +321,20 @@ class CalibrationTask(object):
             d_offset = image_translation.run_image_translation(self._scanner, self._multibeam, self._descanner,
                                                                self._detector, self._dataflow, self._ccd)
             self.asm_config["descanner"]["scanOffset"] = d_offset
+            configure_asm(self._multibeam, self._descanner, self._detector, self._dataflow, self.asm_config,
+                          upload=False)
+
+        elif calibration == Calibrations.DARK_OFFSET:
+            dark_offset = dark_offset_correction.run_dark_offset(self._scanner, self._multibeam, self._descanner,
+                                                                 self._detector, self._dataflow)
+            self.asm_config["mppc"]["cellDarkOffset"] = dark_offset
+            configure_asm(self._multibeam, self._descanner, self._detector, self._dataflow, self.asm_config,
+                          upload=False)
+
+        elif calibration == Calibrations.DIGITAL_GAIN:
+            digital_gain = digital_gain_correction.run_digital_gain(self._scanner, self._multibeam, self._descanner,
+                                                                    self._detector, self._dataflow)
+            self.asm_config["mppc"]["cellDigitalGain"] = digital_gain
             configure_asm(self._multibeam, self._descanner, self._detector, self._dataflow, self.asm_config,
                           upload=False)
 
--- a/src/odemis/acq/fastem_conf.py
+++ b/src/odemis/acq/fastem_conf.py
@@ -143,3 +143,22 @@ def configure_scanner(scanner, mode):
                             scanner.rotation.value)
     else:  # code should never be reached
         raise ValueError("Invalid mode %s." % mode)
+
+
+def configure_detector(detector, roc):
+    """
+    Configure the detector by setting the calibrated parameters as stored for
+    the provided region of calibration (ROC). If calibrated parameters are not
+    available, the current settings on the detector will stay as is.
+    :param detector: (technolution.MPPC) The detector to be configured.
+    :param roc: (FASTEMROC) The region of calibration as selected on the scintillator,
+                which stores the calibrated settings if calibration was performed.
+    """
+    if "cellDarkOffset" in roc.value.parameters:
+        detector.cellDarkOffset.value = roc.value.parameters["cellDarkOffset"]
+    else:
+        logging.warning("Region of calibration doesn't have dark offset parameters.")
+    if "cellDigitalGain" in roc.value.parameters:
+        detector.cellDigitalGain.value = roc.value.parameters["cellDigitalGain"]
+    else:
+        logging.warning("Region of calibration doesn't have digital gain parameters.")
--- a/src/odemis/gui/comp/miccanvas.py
+++ b/src/odemis/gui/comp/miccanvas.py
@@ -656,6 +656,9 @@ class DblMicroscopeCanvas(canvas.DraggableCanvas):
           but with explicit bounding box, so we can zoom to a specific position.
         bbox (4 floats): bounding box to be shown, defined as minx, miny, maxx,
           maxy positions in m.
+
+        Note: Should be called from main GUI thread. Make sure caller of this method
+        is running in the main GUI thread.
         """
         # compute mpp so that the bbox fits exactly the visible part
         w, h = abs(bbox[2] - bbox[0]), abs(bbox[3] - bbox[1])  # m
@@ -676,7 +679,7 @@ class DblMicroscopeCanvas(canvas.DraggableCanvas):
             self.view.mpp.value = self.view.mpp.clip(1 / self.scale)
             self.view.view_pos.value = c
 
-        wx.CallAfter(self.request_drawing_update)
+        self.request_drawing_update()
 
     def _on_view_mpp(self, mpp):
         """ Called when the view.mpp is updated """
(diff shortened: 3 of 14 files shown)


src/odemis/gui/cont/fastem_acq.py
```diff
@@ -581,32 +563,231 @@ def _on_alignment_done(self, future):
...
+        self._tab_data.main.is_acquiring.value = True  # make sure the acquire/tab buttons are disabled
+        self._tab_data.main.is_acquiring.subscribe(self._on_is_acquiring)
+
+        self._on_calibration_state()  # update the controls in the panel
+
+        futures = {}
+        for roc_num in sorted(self._tab_data.calibration_regions.value.keys()):
+            # check if calibration region (ROC) on scintillator is set
+            roc = self._tab_data.calibration_regions.value[roc_num]
+            if roc.coordinates.value != stream.UNDEFINED_ROI:
```
tepals, line 670: The comment on line 668 is not completely clear to me with respect line. Can you clarify that comment or add another comment here, explaining what `roc.coordinates.value != stream.UNDEFINED_ROI` means. Does it mean the ROC is not set?

src/odemis/gui/cont/fastem_acq.py
```diff
@@ -581,32 +563,231 @@ def _on_alignment_done(self, future):
...
+                                       self._main_data_model.descanner, self._main_data_model.mppc,
+                                       self._main_data_model.stage, self._main_data_model.ccd,
+                                       self._main_data_model.beamshift, self._main_data_model.det_rotator,
+                                       calibrations=self.calibrations, stage_pos=roc_center)
+                t = align.fastem.estimate_calibration_time(self.calibrations)
+                # also handles cancelling and exceptions
+                f.add_done_callback(partial(self._on_calibration_done, roc_num=roc_num))
+                futures[f] = t
+
+        # if futures:
```
tepals, line 692: what happens if futures is an empty dict?

src/odemis/gui/cont/fastem_acq.py
```diff
@@ -581,32 +563,231 @@ def _on_alignment_done(self, future):
...
+        # also handles cancelling and exceptions
+        calib_future.add_done_callback(self._on_batch_calibrations_done)
+        # connect the future to the progress bar and its label
+        self._future_connector = ProgressiveFutureConnector(calib_future, self.gauge, self.label, full=False)
+        # TODO needed?
+        # else:  # In case all calibrations failed to start
+        #     self._main_data_model.is_acquiring.value = False
+        #     self._reset_calibration_gui("Calibrations failed (see log panel).", level=logging.WARNING)
+
+    # @call_in_wx_main
```
tepals, line 703: why is this commented?

src/odemis/gui/cont/fastem_acq.py
```diff
@@ -581,32 +563,231 @@ def _on_alignment_done(self, future):
...
+            self._update_calibration_controls("Calibration successful")
+        except CancelledError:
+            self.is_aligned.value = False  # don't enable ROA acquisition
+            self._update_calibration_controls("Calibration cancelled")  # update label to indicate cancelling
+            # return
+        except Exception as ex:
+            self.is_aligned.value = False  # don't enable ROA acquisition
+            logging.exception("Calibration failed with %s.", ex)
+            self._update_calibration_controls("Calibration failed")
+            # return
```
tepals, line 747: I think all these `# return`'s can be removed right?

src/odemis/gui/cont/fastem_acq.py
```diff
@@ -581,32 +563,231 @@ def _on_alignment_done(self, future):
...
+        """
+        # check if any scintillators loaded
+        if not self._main_data_model.active_scintillators.value:
+            self._update_calibration_controls("No scintillator loaded (go to Chamber tab).", False)
         else:
-            duration = align.fastem.estimate_calibration_time(self.calibrations)
-            self.gauge_label.SetLabel(str(duration) + " seconds")
+            # check if at least one roc was selected
+            if all(roc.coordinates.value == stream.UNDEFINED_ROI
+                   for roc in self._tab_data.calibration_regions.value.values()):
```
tepals, line 760: The comment does not match the actual code. By reading the comment I would think `all(roc.coordinates.value == stream.UNDEFINED_ROI for roc in self._tab_data.calibration_regions.value.values()) = True` would mean that at least one roc was selected, but if I read the code it means that no roc was selected. You could update it to:
```
            # check if at least one roc was selected
            if any(roc.coordinates.value != stream.UNDEFINED_ROI
                   for roc in self._tab_data.calibration_regions.value.values()):
                self._update_calibration_controls()
            else:
                self._update_calibration_controls("No calibration region selected.", False)
```

src/odemis/gui/cont/project.py
```diff
@@ -336,100 +349,174 @@ def _on_text(self, evt):
...
+        self.calib_model.coordinates.subscribe(self._on_coordinates)
 
-        # Add overlay
-        self.overlay = view_ctrl.viewports[0].canvas.add_calibration_overlay(self.model.coordinates, number)
+        self.overlay = None
 
     def fit_view_to_bbox(self):
         """
-        Zoom in to calibration region. Calibration region is in the center and takes up 1/100 of viewport area.
+        Zoom into calibration region. Calibration region is centered in the scintillator with the
```
tepals, line 372: I think actually "Zoom in on calibration region" would fit better.

src/odemis/gui/cont/project.py
```diff
@@ -336,100 +349,174 @@ def _on_text(self, evt):
...
+        corresponding number and takes up approximately 1/100 of the viewport area.
         """
-        logging.debug("Zooming in to calibration region %s.", self.model.name.value)
+        logging.debug("Zooming into calibration region %s.", self.calib_model.name.value)
```
tepals, line 375: I think actually "Zoom in on calibration region" would fit better.

src/odemis/gui/cont/project.py
```diff
@@ -336,100 +349,174 @@ def _on_text(self, evt):
...
         cnvs = self._view_ctrl.viewports[0].canvas
-        l, t, r, b = self.model.coordinates.value
-        sz = (self._tab_data.main.ebeam.resolution.value[0] * self._tab_data.main.ebeam.pixelSize.value[0],
-              self._tab_data.main.ebeam.resolution.value[1] * self._tab_data.main.ebeam.pixelSize.value[1])
-        cnvs.fit_to_bbox([l - 5 * sz[0], t - 5 * sz[1], r + 5 * sz[0], b + 5 * sz[1]])
+        l, t, r, b = self.calib_model.coordinates.value
+        size = (self._tab_data.main.multibeam.resolution.value[0] * self._tab_data.main.multibeam.pixelSize.value[0],
+                self._tab_data.main.multibeam.resolution.value[1] * self._tab_data.main.multibeam.pixelSize.value[1])
+        cnvs.fit_to_bbox([l - 5 * size[0], t - 5 * size[1], r + 5 * size[0], b + 5 * size[1]])
```
tepals, line 380: Why do you do `5*size`?

src/odemis/gui/cont/project.py
```diff
@@ -336,100 +349,174 @@ def _on_text(self, evt):
...
 
+        tab_data.main.is_acquiring.subscribe(self._on_is_acquiring)  # enable/disable button
+
     def _on_button(self, evt):
-        btn = evt.GetEventObject()
-        btn.SetLabel("OK")
-        btn.SetForegroundColour(wx.GREEN)
+        """
+        Called when one of the 9 single region of calibration (ROC) buttons is triggered.
+        A ROC can be added or removed by clicking again. If a ROC is added, an overlay will be
```
tepals, line 439: Maybe a typo: I pronounce ROC as "arr ooo ceee" so I would write "an ROC", if you pronounce it as "rock" you can leave it as is.

https://www.quora.com/When-using-an-abbreviation-starting-with-the-letter-R-is-it-a-or-an-For-example-I-played-an-RTG-game#:~:text=It%20depends%20whether%20you%20pronounce,correct%20to%20use%20a%20instead.

src/odemis/gui/cont/tabs.py
```diff
@@ -1809,13 +1795,31 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
+        self._calib_1_controller = fastem_acq.FastEMCalibrationController(
             tab_data,
             panel,
-            calibrations
+            calib_prefix="calib_1",
+            calibrations=calibrations
         )
 
-        # Controller for acquistion settings panel
+        # FIXME instantiate FastEMCalibrationRegionsController in FastEMScintillatorCalibrationController inherit from?
```
tepals, line 1805: I do not understand this sentence, can you improve it?

tepals (changes requested):
I also pulled the code and the GUI looks good. I noticed that the calibration 2 button is enable, even when calibration 1 has not been run. Is that on purpose?

Sabrina85:
> 

yes that does not work yet. For this I need to make a fancier implementation for when the is_aligned. I will add a TODO somewhere.

src/odemis/acq/align/fastem.py
```diff
@@ -223,6 +252,8 @@ def run(self):
             configure_asm(self._multibeam, self._descanner, self._detector, self._dataflow, self.asm_config)
             logging.debug("Calibrations finished.")
 
+        return self.asm_config
+
     def run_calibrations(self, calibration):
```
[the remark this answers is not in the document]
Sabrina85: yes, but as I am already working on the field corrections I will remove it already now.

src/odemis/gui/comp/miccanvas.py
```diff
@@ -1855,6 +1855,14 @@ def add_calibration_overlay(self, coordinates, label, colour=gui.FG_COLOUR_WARNI
         overlay.active.value = False  # no need to activate by default
         return overlay
 
+    def remove_calibration_overlay(self, overlay):
+        """
+        overlay (FastEMROAOverlay): overlay to be deleted
+        """
+        overlay.active.value = False  # deactivating the overlay avoids weird behaviour
+        self.remove_world_overlay(overlay)
+        wx.CallAfter(self.request_drawing_update)
```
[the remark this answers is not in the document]
Sabrina85: I think you are right that we can reuse it. I just tested it and it seems to work fine (there are not dump questions ;) ).

src/odemis/gui/cont/fastem_acq.py
```diff
@@ -581,32 +563,231 @@ def _on_alignment_done(self, future):
...
+
+        futures = {}
+        for roc_num in sorted(self._tab_data.calibration_regions.value.keys()):
+            # check if calibration region (ROC) on scintillator is set
+            roc = self._tab_data.calibration_regions.value[roc_num]
+            if roc.coordinates.value != stream.UNDEFINED_ROI:
+
+                # calculate the center position of the ROC (half field right/bottom
+                # compared to top/left corner in view)
+                l, t, _, _ = roc.coordinates.value  # (l, t, r, b)
```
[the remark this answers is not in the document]
Sabrina85: Is top = ymin?

src/odemis/gui/cont/project.py
Sabrina85: It defines how much we zoom in on the ROC - so that there is still some space shown around the ROC. I added a comment.

src/odemis/gui/cont/fastem_acq.py
Sabrina85: `roc.coordinates.value != stream.UNDEFINED_ROI` means that the ROC is set, if it is undefined, it is not set.

src/odemis/gui/cont/fastem_acq.py
Sabrina85: I am not sure. 
The cancelling is not raised correctly for some reason atm. 
And in the ROA acquisition and overview acquisition there is a return. 
@pieleric can you comment please?

Sabrina85: What happens at the moment is that when I cancel the calibration, it raises the CancellationError in the callback of the single future, but then in the batch future it is not receiving the CancellationError.

src/odemis/gui/cont/fastem_acq.py
Sabrina85: The gui is blocked. I am just wondering if the dict can ever be empty.

src/odemis/gui/cont/fastem_acq.py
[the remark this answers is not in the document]
tepals: For the ROA the coordinates we get are in the sample carrier system, which corresponds to the component with role='stage'. There `xmin, ymin, xmax, ymax` is equal to `l, t, r, b`

src/odemis/gui/cont/fastem_acq.py
Sabrina85: @pieleric In which case would the dict be empty?

pieleric (changes requested):

src/odemis/gui/cont/project.py
```diff
@@ -336,100 +350,175 @@ def _on_text(self, evt):
...
+        for num, b in self.panel.buttons.items():
+            if num not in scintillators:
+                # reset coordinates for ROC to undefined
+                self.roc_ctrls[num].calib_model.coordinates.value = acqstream.UNDEFINED_ROI
+
+        # update ROC buttons
+        self._on_coordinates()
+
+    def _update(self):
+        pass
```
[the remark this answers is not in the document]
Sabrina85: Oh that should go. This was a left over from discussions.

src/odemis/gui/cont/fastem_acq.py
```diff
@@ -431,6 +430,7 @@ def _reset_acquisition_gui(self, text=None, level=None):
         else:
             self.update_acquisition_time()
 
+    @call_in_wx_main  # call in main thread as changes in GUI are triggered
```
pieleric, line 433: No, `on_acquisition()` is called when the "acquire" button is pressed. (see the "evt" argument). So it's already called in the main GUI thread. It shouldn't be called delayed.

src/odemis/gui/cont/project.py
```diff
@@ -291,6 +297,7 @@ def __init__(self, name, roc, colour, tab_data, project_panel, view_ctrl):
         self.overlay = canvas.add_roa_overlay(self.model.coordinates, colour)
         self.overlay.active.subscribe(self._on_overlay_active)
 
+    # FIXME how taken care of that his is run in the main gui thread?
```
pieleric, line 300: Add a note in the docstring "Should be called from the main GUI thread."  (and then make sure that the callers are running in the main GUI thread.

src/odemis/gui/cont/project.py
```diff
@@ -215,7 +221,7 @@ def _add_roa_ctrl(self, coords):
         # Enable buttons of project bar
         self._project_bar.enable_buttons(True)
 
-    # Should be called from GUI main thread
+    # already running in main GUI thread as it receives event from GUI
```
pieleric, line 224: Well, it's not directly receiving events from the GUI. Looking at the code, I see it's called from `_on_btn_remove()`, which is receiving a GUI event. I'd suggest to add a docstring saying "Must be called from within the main GUI thread".

src/odemis/gui/cont/project.py
```diff
@@ -368,6 +379,7 @@ def __init__(self, number, tab_data, view_ctrl):
 
         self.overlay = None
 
+    @call_in_wx_main  # call in main thread as changes in GUI are triggered
```
pieleric, line 382: Indeed, now I see that `canvas.fit_to_bbox()` does use some GUI stuff (while it also tries to not depend on GUI stuff!). So we could add a note in the `fit_to_bbox()` docstring that this function should be called within the main GUI thread.
An alternative to adding the decorator here would be to call `wx.CallAfter(cnvs.fit_to_bbox, [...])` (as it's the last function called, and we don't need the result)

pieleric (changes requested):

pieleric (approved):

src/odemis/gui/cont/project.py
Sabrina85: I tried the wx.CallAfter (and no decorator) but then I get this error:
`11:37:42 ERROR   main           : Traceback (most recent call last):
  File "/home/sabrina/development/odemis/src/odemis/gui/cont/project.py", line 485, in _on_button
    self.roc_ctrls[num].fit_view_to_bbox()  # Zoom to calibration region
  File "/home/sabrina/development/odemis/src/odemis/gui/cont/project.py", line 399, in fit_view_to_bbox
    wx.CallAfter(cnvs.fit_to_bbox([xmin - 5 * size[0], ymin - 5 * size[1], xmax + 5 * size[0], ymax + 5 * size[1]]))
  File "/usr/lib/python3/dist-packages/wx/core.py", line 3183, in CallAfter
    assert callable(callableObj), "callableObj is not callable"
AssertionError: callableObj is not callable`

pieleric: Please note the syntax. `CallAfter()` takes the function as first argument, and the arguments as following arguments. The way you wrote it immediately called `fit_to_bbox()`, then took the return value (None), and passed it to `CallAfter()`.
You need to write something like `wx.CallAfter(cnvs.fit_to_bbox, [...])`.

Sabrina85:
- replaced the viewport with the view only
- added test case for calibration manager: dark/gain calibration
- fixed/updated the wx calls in main thread
There is still an issue when you really quickly add and remove projects/ROAs it sometimes fails. I think it has to do with the update time of 1s.

src/odemis/gui/comp/miccanvas.py
```diff
@@ -656,6 +656,8 @@ def fit_to_bbox(self, bbox):
           but with explicit bounding box, so we can zoom to a specific position.
         bbox (4 floats): bounding box to be shown, defined as minx, miny, maxx,
           maxy positions in m.
+
+        Note: Should be called from main GUI thread.
```
tepals, line 660: If it is already called from the main GUI thread I would update this comment to: `Note: Is called from main GUI thread`. If it should be forced to be called from the main GUI thread, I think you need to add the `@call_in_wx_main` decorator to this function.

tepals (changes requested):

src/odemis/gui/cont/project.py
```diff
@@ -336,100 +350,176 @@ def _on_text(self, evt):
...
+
     def _on_button(self, evt):
-        btn = evt.GetEventObject()
-        btn.SetLabel("OK")
-        btn.SetForegroundColour(wx.GREEN)
+        """
+        Called when one of the 9 single region of calibration (ROC) buttons is triggered.
+        An ROC can be added or removed by clicking again. If a ROC is added, an overlay will be
+        created in the center of the corresponding viewport. Size of the ROC is one single field.
+        If a ROC is removed, the coordinates are reset to undefined.
```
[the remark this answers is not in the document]
Sabrina85: Yes indeed, this behavior we want in future.

src/odemis/gui/cont/project.py
```diff
@@ -336,100 +350,176 @@ def _on_text(self, evt):
...
 
-        # Add overlay
-        self.overlay = view_ctrl.viewports[0].canvas.add_calibration_overlay(self.model.coordinates, number)
+        self.overlay = None
 
     def fit_view_to_bbox(self):
         """
-        Zoom in to calibration region. Calibration region is in the center and takes up 1/100 of viewport area.
+        Zoom in on calibration region (ROC). Calibration region is centered in the scintillator with the
+        corresponding number and takes up approximately 1/100 of the viewport area.
```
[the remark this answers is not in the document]
Sabrina85: I rephrased:
`Zoom in on calibration region (ROC). Calibration region has the size of a single field
        image, which is approximately 1/100 of the viewport area (= size of a single scintillator)
        and is centered in the scintillator with the corresponding number.`

src/odemis/acq/fastem_conf.py
```diff
@@ -143,3 +143,22 @@ def configure_scanner(scanner, mode):
...
+    the provided region of calibration (ROC). If calibrated parameters are not
+    available, the current settings on the detector will stay as is.
+    :param detector: (technolution.MPPC) The detector to be configured.
+    :param roc: (FASTEMROC) The region of calibration as selected on the scintillator,
+                which stores the calibrated settings if calibration was performed.
+    """
+    if "cellDarkOffset" in roc.value.parameters:
+        detector.cellDarkOffset.value = roc.value.parameters["cellDarkOffset"]
+    else:
+        logging.warning("Region of calibration doesn't have dark offset parameters.")
```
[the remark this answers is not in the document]
Sabrina85: Not necessarily. It is two-fold. If it cannot find any parameters it indeed it will use the currently set values on the VAs. BUT, this method is called with two different ROCs. The first one will only have dark/gain parameters, so it will never have cell translation parameters and will always trigger the warning. However, it will not use the VA settings for cell translation parameters as the same function is also called with the ROC that actually contains the cell translation parameters. 
So I cannot add that the VA values are used. We can maybe change how the method is used by the 2 ROCs. But I don't have a good idea yet.

src/odemis/gui/comp/miccanvas.py
Sabrina85: I added this comment in concordance with Éric's review. The "should" indicates that every caller should already be running in the main GUI thread, so also potential future callers. I think that is a bit the difference to "Is" - this does not highlight the need that I need to make sure the caller is already running in the main GUI thread. 
I can extend the comment and add "Caller should run in main GUI thread".

I don't know if the decorator makes sense here. The part that changes the GUI is the drawing update, which is in a wx.CallAfter call. From what I understand this makes sure that this call is only executed after the current and pending event handlers have been completed. But I don't understand why we still need to make sure that the caller of this method runs in the main gui thread. @pieleric can you clarify?

tepals: Then can you maybe add what call makes sure that this is called from the main GUI thread?

src/odemis/acq/fastem_conf.py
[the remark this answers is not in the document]
Sabrina85: Will change input arg to list of ROCs in pr for calibration step 3.

tepals (approved):

PR #2081, opened 2022-03-22
Fastem acquisition: now that the pre-calibrations are done outside of…

tepals:
… the ROA the beamshift correction also needs to happen on the first field.


Sabrina85 (approved):

tepals:
> Is there a test that needs to be updated for this?

no

PR #2082, opened 2022-03-23
fastem-calib-3

Sabrina85:


src/odemis/gui/cont/fastem_acq.py
```diff
@@ -564,12 +548,11 @@ def on_align(self, evt):
...
 
-        f.add_done_callback(self._on_alignment_done)  # also handles cancelling and exceptions
+        f.add_done_callback(self._on_calibration_done)  # also handles cancelling and exceptions
         # connect the future to the progress bar and its label
-        self._future_connector = ProgressiveFutureConnector(f, self.gauge_progress, self.gauge_label)
+        self._future_connector = ProgressiveFutureConnector(f, self.gauge, self.label, full=False)
 
-    @call_in_wx_main
-    def _on_alignment_done(self, future):
+    def _on_calibration_done(self, future, _=None):
```
tepals, line 555: I added @call_in_wx_main above this function

tepals, line 555: I added @call_in_wx_main above this function

src/odemis/gui/cont/fastem_acq.py
```diff
@@ -581,32 +564,285 @@ def _on_alignment_done(self, future):
...
+            self._update_calibration_controls("Calibration cancelled")  # update label to indicate cancelling
         except Exception as ex:
             self.is_aligned.value = False  # don't enable ROA acquisition
             logging.exception("Calibration failed with %s.", ex)
-            self._reset_calibration_gui("Calibration failed")
-            return
+            self._update_calibration_controls("Calibration failed")
 
-    def _reset_calibration_gui(self, text=None):
+    def _on_calibration_state(self, _=None):
```
tepals, line 576: I added @call_in_wx_main above this function

tepals, line 576: I added @call_in_wx_main above this function

src/odemis/gui/cont/fastem_acq.py
```diff
@@ -581,32 +564,285 @@ def _on_alignment_done(self, future):
...
+        calib_future.add_done_callback(self._on_batch_calibrations_done)
+        # connect the future to the progress bar and its label
+        self._future_connector = ProgressiveFutureConnector(calib_future, self.gauge, self.label, full=False)
+        # TODO needed?
+        # else:  # In case all calibrations failed to start
+        #     self._main_data_model.is_acquiring.value = False
+        #     self._reset_calibration_gui("Calibrations failed (see log panel).", level=logging.WARNING)
+
+    # @call_in_wx_main
+    def _on_calibration_done(self, future, roc_num):
```
tepals, line 709: I added @call_in_wx_main above this function

tepals, line 709: I added @call_in_wx_main above this function

src/odemis/gui/cont/fastem_acq.py
```diff
@@ -581,32 +564,285 @@ def _on_alignment_done(self, future):
...
+            self.is_aligned.value = False  # don't enable ROA acquisition
+            self._update_calibration_controls("Calibration cancelled")  # update label to indicate cancelling
+            # return
+        except Exception as ex:
+            self.is_aligned.value = False  # don't enable ROA acquisition
+            logging.exception("Calibration failed with %s.", ex)
+            self._update_calibration_controls("Calibration failed")
+            # return
+
+    def _on_calibration_state(self, _=None):
```
tepals, line 754: I added @call_in_wx_main above this function

tepals, line 754: I added @call_in_wx_main above this function

src/odemis/gui/cont/fastem_acq.py
Sabrina85: Thanks. I am surprised that it needed to be added here as well. I understood that I only need to add it if something is adjusted in the GUI. Is this because I call here a function that eventually adjusts something in the GUI?

src/odemis/gui/model/__init__.py
```diff
@@ -1276,11 +1276,24 @@ def __init__(self, main):
...
+        # calibration regions for calibration step 3
+        self.regions_calib_3 = model.VigilantAttribute({})  # dict, number --> FastEMROC
+        for i in main.scintillator_positions:
+            self.regions_calib_3.value[i] = fastem.FastEMROC(str(i), acqstream.UNDEFINED_ROI)
+
         self.projects = model.ListVA([])  # list of FastEMProject
 
+        self.is_calib_1 = model.BooleanVA(False)
+        self.is_calib_2 = model.BooleanVA(False)
+        self.is_calib_3 = model.BooleanVA(False)
```
pieleric, line 1293: Could you add a comment explaining what True/False means? Is it running the calibration? Or is it done calibrating? Is it entirely calibrated for all the active scintillators?

src/odemis/gui/model/__init__.py
```diff
@@ -1293,6 +1306,10 @@ def __init__(self, main):
 
         self.selected_scintillators = model.ListVA([])  # set of ints, overview images to be acquired
 
+        self.is_calib = model.BooleanVA(False)
```
pieleric, line 1309: Maybe name that `is_calibrated` (assuming that's the meaning), to clarify better the difference with `is_calibrating`?

src/odemis/gui/model/__init__.py
```diff
@@ -1293,6 +1306,10 @@ def __init__(self, main):
...
+        # indicates whether the microscope is calibrating
+        self.is_calibrating = model.BooleanVA(False)
```
pieleric, line 1311: Note that there is some code in the GUI that looks at `is_acquiring` and prevents the user from closing the window if it's True (and tells the user that it should first stop the acquisition). Does the same thing need to happen if it's calibrating? (I'd tend to answer "no", because it's fine to close the GUI, in which case the calibration will eventually end, and next time the GUI is started, the calibration will have to be redone. Not a big deal as loosing acquisition data).

src/odemis/gui/cont/project.py
```diff
@@ -336,100 +356,192 @@ def _on_text(self, evt):
...
-    def __init__(self, tab_data, calibration_bar, view_ctrl):
+    def __init__(self, tab_data, calibration_panel, view_ctrl, calib_prefix):
         """
-        tab_data (FastEMAcquisitionGUIData): tab data model
-        calibration_bar (FastEMCalibrationBar): main calibration panel
-        view_ctrl (FastEMAcquisitionViewport): viewport controller
+        :param tab_data (FastEMAcquisitionGUIData): The tab data model.
+        :param calibration_panel (FastEMCalibrationPanelHeader): The main calibration panel including the 9 regions
+                of calibration (ROC) buttons, the calibrate button, the gauge and the label.
+        :param view_ctrl (FastEMAcquisitionViewport): The viewport controller. TODO replace with just view
```
pieleric, line 428: It seems that it would simplify quite a few lines, and seems rather straightforward to do... could you do it now?

src/odemis/gui/cont/project.py
```diff
@@ -336,100 +356,192 @@ def _on_text(self, evt):
...
+            self.roc_ctrls[roc_num] = FastEMROCController(roc_num, tab_data, view_ctrl, calib_prefix)
+            roc.coordinates.subscribe(self._on_coordinates)
+
         # Only enable buttons for scintillators which have been selected in the chamber tab
         tab_data.main.active_scintillators.subscribe(self._on_active_scintillators)
 
+        tab_data.main.is_acquiring.subscribe(self._on_is_acquiring)  # enable/disable calib button during acquisition
+        tab_data.is_calibrating.subscribe(self._on_is_acquiring)  # enable/disable calib button during calibration
+
+    # @call_in_wx_main FIXME crashes GUI!
```
pieleric, line 457: Add a comment explaining that it's called from the main GUI thread.

pieleric (changes requested):
I'm not sure how this PR will mix with the calib2 PR... I feel it's going to bring a lot of conflicts. Maybe you should merge these two PRs together to avoid too much conflict?

src/odemis/acq/fastem_conf.py
```diff
@@ -143,3 +143,26 @@ def configure_scanner(scanner, mode):
...
         raise ValueError("Invalid mode %s." % mode)
+
+
+def configure_detector(detector, roc):
+    """
+    Configure the detector by setting the calibrated parameter as stored for
+    the provided region of calibration (ROC). If calibrated parameters are not
+    available, the current settings on the detector will stay as is.
+    :param detector: (technolution.MPPC) The detector to be configured.
+    :param roc: (FASTEMROC) The region of calibration as selected on the scintillator,
```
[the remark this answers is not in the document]
Sabrina85: Maybe we should change it some time to all capital letters. FASTEM is the acronym for **F**ast **A**utomated **S**canning **T**ransmission **E**lectron **M**icroscope.

src/odemis/gui/model/__init__.py
Sabrina85: Good point. The calibrations are indeed themselves independent from the front-end. I am wondering about the part, when the calibration would return to the calibration manager and store the calibration values (which is ok to loose) and put the system back into a good state (which is for some calibrations important). So if the GUI is closed during a calibration that modifies the scan amplitude and usually at the end would put back the original value, that would not happen anymore right? On the other hand we indeed then run the calibrations again and those don't depend on VA settings but do use factory or calibrated values. So most likely it is not a problem indeed.

src/odemis/gui/model/__init__.py
Sabrina85: I had another look at this. I actually had named it "calib" or "calib_1" etc. to be able to make the controllers generic by using the prefix attribute. If I make it a new name, I don't see atm how to generalize.
`        # get attribute that keeps track of calibration status
        setattr(self, "is_" + calib_prefix, getattr(tab_data, "is_" + calib_prefix))`

src/odemis/gui/cont/fastem_acq.py
```diff
@@ -337,9 +350,12 @@ def _on_projects(self, projects):
...
         for roa in roas:
-            if roa not in self.roa_subscribers:
-                roa.roc.subscribe(self._on_va_change)
-                self.roa_subscribers.append(roa)
+            if roa not in self.roa_subscribers_2:
+                roa.roc_2.subscribe(self._on_va_change)
+                self.roa_subscribers_2.append(roa)
+            if roa not in self.roa_subscribers_3:
+                roa.roc_3.subscribe(self._on_va_change)
+                self.roa_subscribers_3.append(roa)
```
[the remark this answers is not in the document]
Sabrina85: I am now wondering for what the roa_subscriber attribute is used for. I didn't find it being used anywhere.

Sabrina85:
I updated the pr:
What does not work yet:

- When selecting one of the 9 ROCs in either step 2 or 3, the corresponding calibrate button will be already enabled, though the first step (and/or the second 2 for step 3) was not run yet.
- When moving a ROC, the parameters (e.g. dark offset) are not reset.
- When deselecting and then selecting the ROC for the same scintillator again, the parameters are also not reset (only the text label is reset).

src/odemis/acq/fastem.py
```diff
@@ -466,6 +470,16 @@ def pre_calibrate(self):
...
+            # temp fix: hardcode scan amplitude
+            asm_config["multibeam"]["scanAmplitude"] = configure_hw.CALIBRATED
+            asm_config["multibeam"]["scanOffset"] = configure_hw.CALIBRATED
+            configure_hw.configure_asm(self._multibeam,
+                                       self._descanner,
+                                       self._detector,
+                                       self._detector.data,
+                                       asm_config,
+                                       upload=False)
+
```
tepals, line 482: Do we want to have this on master, or should it just be a temp commit on EA2?

src/odemis/acq/fastem_conf.py
```diff
@@ -145,20 +145,28 @@ def configure_scanner(scanner, mode):
...
-    :param roc: (FASTEMROC) The region of calibration as selected on the scintillator,
-                which stores the calibrated settings if calibration was performed.
+    :param rocs: (list of FastEMROC) The region of calibration as selected on the scintillator,
+                 which stores the calibrated settings if calibration was performed.
     """
-    if "cellDarkOffset" in roc.value.parameters:
-        detector.cellDarkOffset.value = roc.value.parameters["cellDarkOffset"]
-    else:
+    # check all parameters are available
+    if not any(roc.parameters and "cellDarkOffset" in roc.parameters for roc in rocs):
```
tepals, line 158: doesn't this still log the warning when you only pass the cell translation ROC?

src/odemis/gui/cont/fastem_acq.py
```diff
@@ -594,7 +639,9 @@ def _update_calibration_controls(self, text=None, button_state=True):
         """
         self.button.Enable(button_state)  # enable/disable button
 
-        if self._tab_data.main.is_acquiring.value:
+        # TODO disable overlay, so it cannot be moved while calibrating!
```
tepals, line 642: Can you make this TODO more specific, what overlay do you want to disable?

src/odemis/gui/cont/project.py
```diff
@@ -312,6 +316,7 @@ def create_panel(self):
         self._project_panel.add_roa_panel(self.panel)
 
         self.panel.calibration_ctrl.Bind(wx.EVT_COMBOBOX, self._on_combobox)
+        # TODO add roc 3
```
tepals, line 319: Can you make this TODO more specific, where should roc 3 be added?

src/odemis/gui/cont/tabs.py
```diff
@@ -1811,15 +1811,47 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
-        self._calib_2_controller = fastem_acq.FastEMScintillatorCalibrationController(
+        self._calib_2_controller = fastem_acq.FastEMCalibration2Controller(
             tab_data,
             panel,
             calib_prefix="calib_2",
-            calibrations=[Calibrations.OPTICAL_AUTOFOCUS, Calibrations.IMAGE_TRANSLATION_PREALIGN,
-                          Calibrations.DARK_OFFSET, Calibrations.DIGITAL_GAIN]
+            calibrations=[Calibrations.OPTICAL_AUTOFOCUS, Calibrations.IMAGE_TRANSLATION_PREALIGN]
+            # calibrations=[Calibrations.OPTICAL_AUTOFOCUS, Calibrations.IMAGE_TRANSLATION_PREALIGN,
+            #               Calibrations.DARK_OFFSET, Calibrations.DIGITAL_GAIN]
```
tepals, line 1824: why is this commented?

src/odemis/gui/cont/tabs.py
```diff
@@ -1811,15 +1811,47 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
+            viewport=vp,
+            calib_prefix="calib_3",
+        )
+
+        # Controller for calibration panel 3
+        self._calib_3_controller = fastem_acq.FastEMCalibration3Controller(
+            tab_data,
+            panel,
+            calib_prefix="calib_3",
+            # calibrations=[Calibrations.IMAGE_TRANSLATION_PREALIGN]
```
tepals, line 1853: why is this commented?

tepals (changes requested):

src/odemis/acq/fastem.py
Sabrina85: yes let's make it a separate temporary commit.

src/odemis/acq/fastem_conf.py
Sabrina85: Yes correct - in that case you want to have the warning that the following acquisition does not have a calibrated dark offset I would assume. What do you think?

src/odemis/gui/cont/project.py
Sabrina85: I think I can remove this here. In the beginning I thought we also want the dropdown menu for the calibrations 3. However, there we should not give the user the possibility to select a different scintillator as this calibration is supposed to be scintillator specific.

src/odemis/gui/cont/tabs.py
Sabrina85: left over from testing

src/odemis/gui/cont/tabs.py
Sabrina85: left over from testing

src/odemis/acq/fastem_conf.py
tepals: Yes indeed, I thought this was somehow also called outside the acquisition. But that's not true, so this behavior is correct.

tepals (approved):

PR #2084, opened 2022-03-24
fastem-improvements-acq-calib

Sabrina85:
See descriptions of commits.

### DIFF
--- a/src/odemis/acq/fastem.py
+++ b/src/odemis/acq/fastem.py
@@ -464,11 +464,11 @@ class AcquisitionTask(object):
         if not fastem_calibrations:
             raise ModuleNotFoundError("Need fastem_calibrations repository to run pre-calibrations.")
 
-        asm_config_orig = None
+        asm_config = None
 
         try:
             logging.debug("Read initial Hw settings.")
-            asm_config_orig = configure_hw.get_config_asm(self._multibeam, self._descanner, self._detector)
+            asm_config = configure_hw.get_config_asm(self._multibeam, self._descanner, self._detector)
 
             fastem_conf.configure_scanner(self._scanner, fastem_conf.MEGAFIELD_MODE)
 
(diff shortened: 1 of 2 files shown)


src/odemis/acq/test/fastem_test.py
```diff
@@ -298,8 +298,8 @@ class TestFastEMAcquisition(unittest.TestCase):
 
     @classmethod
     def setUpClass(cls):
-        if TEST_NOHW:
-            test.start_backend(FASTEM_CONFIG_ASM)
+        # if TEST_NOHW:
+        #     test.start_backend(FASTEM_CONFIG_ASM)
```
tepals, line 302: I think this is from debugging?

Sabrina85: yes

tepals (approved):

src/odemis/acq/fastem.py
```diff
@@ -515,12 +511,17 @@ def pre_calibrate(self):
...
+            else:
+                configure_hw.configure_asm(self._multibeam, self._descanner, self._detector,
+                                           self._detector.dataflow, asm_config)
+            # put system back into state ready for next task and set calibrated settings
             configure_hw.configure_asm(self._multibeam,
                                        self._descanner,
                                        self._detector,
                                        self._detector.data,
-                                       asm_config_orig,
+                                       asm_config,
```
pieleric, line 524: Why do you do the same thing twice in a row?
Also `self._detector.dataflow` probably doesn't exist, it's `self._detector.data`.

pieleric (changes requested):

src/odemis/acq/fastem.py
Sabrina85: This indeed is incorrect.

pieleric (approved):

PR #2086, opened 2022-03-25
Fastem test cases: raise an error when trying to run the test cases w…

tepals:
…ith hardware and no backend running.

Previously when trying to run the test cases with TEST_NOHW=False and no backend running it would try to get the different components, and fail with a non descriptive error.
Raise a more informative error when there is no backend running, and thus it is impossible to get the components.

### DIFF
--- a/install/linux/usr/share/odemis/sim/fastem-sim.odm.yaml
+++ b/install/linux/usr/share/odemis/sim/fastem-sim.odm.yaml
@@ -209,11 +209,15 @@ FASTEM-sim: {
 "Beam Shift Controller": {
     class: tfsbc.BeamShiftController,
     role: ebeam-shift,
+    # The Multibeam Scanner XT can be set as a dependency to automatically retrieve
+    # the calibration data from the scanner, however this is not possible here,
+    # because the SimSem does not contain calibration data.
+    # dependencies: {"scanner": "MultiBeam Scanner XT"},
     init: {
         port: "/dev/fake",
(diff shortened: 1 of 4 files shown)


pieleric (approved):

Sabrina85 (approved):

install/linux/usr/share/odemis/sim/fastem-sim.odm.yaml
```diff
@@ -209,11 +209,12 @@ FASTEM-sim: {
 "Beam Shift Controller": {
     class: tfsbc.BeamShiftController,
     role: ebeam-shift,
+    dependencies: {"scanner": "MultiBeam Scanner XT"},
```
tepals, line 212: This actually doesn't work, because the simsem (used in this yaml file but not in the other), does not have the beamshift transformation matrix. Should I just delete this again? Or do you have another suggestion on how to add this?

Sabrina85: Is the matrix needed for any of the tests where we use this backend?

tepals: good question, it is not. I will comment this dependency here, and add a comment that this is not supported for the simsem

Sabrina85: Sounds good!

Sabrina85 (approved):

PR #2087, opened 2022-03-25
Model core: raise error in getMicroscope when the backend is stopped …

tepals:
…or dead.

When getMicroscope was called without a running backend you would get an uninformative error. This new error will give some more information on the status of the backend.


tepals:
I didn't properly test it and it wasn't working when trying to start a backend.

PR #2092, opened 2022-03-29
driver zeiss: increase magnification range

pieleric:
Some system report magnification up to 2M x. For now, on such system,
starting Odemis when the magnification is > 500K would fail.

=> Change the default range to 2M x, and allow to change the range
further just from the microscope file.

### DIFF
--- a/src/odemis/driver/zeiss.py
+++ b/src/odemis/driver/zeiss.py
@@ -37,8 +37,7 @@ import serial
 import threading
 import time
 
-
-MAGNIFICATION_RANGE = (5., 500.e3)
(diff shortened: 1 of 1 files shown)


src/odemis/driver/zeiss.py
```diff
@@ -37,8 +37,7 @@
 import threading
 import time
 
-
-MAGNIFICATION_RANGE = (5., 500.e3)
+MAGNIFICATION_RANGE = (5., 2e6)  # Doc says max 500k, but some microscopes has 2M
```
tepals, line 40: typo: some microscopes **have**

PR #2093, opened 2022-03-31
gui cont: refactor cryo acquire stream controller

pieleric:


src/odemis/gui/cont/streams.py
```diff
@@ -3083,3 +3084,124 @@ def _on_overview_streams(self, _):
             if isinstance(s, FastEMOverviewStream) and s not in tab_streams:
                 tab_streams.append(s)
                 canvas.view.addStream(s)
+
+
+class CryoAcquiredStreamsController(StreamBarController):
+    """
+    StreamBarController to control the display of the Static streams in the Cryo
+    Localization tab. It deals with two types of streams:
+     * Overview streams: shown only in the overview view, and always.
```
Kleijwegt, line 3093: This sentence ends weird, I think I get what you want to say but I can't formulate it any better.

src/odemis/gui/model/__init__.py
```diff
@@ -661,28 +661,43 @@ def add_new_feature(self, pos_x, pos_y, pos_z=None, f_name=None, milling_angle=D
...
+            return  # We are already good nothing else to do
+
+        # Find the closet feature (which is nearby)
+        try:
+            closest = min(self.main.features.value, key=dist_to_pos)
+            if dist_to_pos(closest) <= self.ATOL_FEATURE_POS:
+                self.main.currentFeature.value = closest
+                return
+        except ValueError:  # raise by min() when no features
+            pass
```
Kleijwegt, line 693: You pass here but I would expect the rest of the code in this except part. If it doesn't go into this expect it returns so that is actually clearer to follow.

Kleijwegt (changes requested):

src/odemis/gui/cont/streams.py
```diff
@@ -3083,3 +3084,124 @@ def _on_overview_streams(self, _):
...
+            if isinstance(stream, StaticStream):
+                for v in (self._feature_view, self._ov_view):
+                    if hasattr(v, "removeStream"):
+                        v.removeStream(stream)
+
+        if clear_model:
+            while self._tab_data_model.overviewStreams.value:
+                stream = self._tab_data_model.overviewStreams.value.pop()
+
+            # Typically, all the features will also be deleted, but that's not our job
```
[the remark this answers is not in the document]
pieleric: It's the job of "another controller". Each controller works independently so we don't need to care. Well, I've now indicated in the comment the function that takes care of the features.

src/odemis/gui/model/__init__.py
pieleric: The problem is that the rest of the code is both for the "except" part, and for the "else" of the if. That's why it's after.

src/odemis/gui/model/__init__.py
```diff
@@ -661,28 +661,43 @@ def add_new_feature(self, pos_x, pos_y, pos_z=None, f_name=None, milling_angle=D
...
+            return math.hypot(feature.pos.value[0] - current_position["x"],
+                              feature.pos.value[1] - current_position["y"])
+
+
+        if current_feature and dist_to_pos(current_feature) <= self.ATOL_FEATURE_POS:
+            return  # We are already good nothing else to do
+
+        # Find the closet feature (which is nearby)
+        try:
+            closest = min(self.main.features.value, key=dist_to_pos)
```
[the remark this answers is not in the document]
pieleric: I've improved the comment.

src/odemis/gui/cont/tabs.py
```diff
@@ -770,13 +665,33 @@ def display_acquired_data(self, data):
...
+            self.tab_data_model.streams.value.insert(0, s)  # TODO: let addFeatureStream do that
+            self._acquired_stream_controller.showFeatureStream(s)
+
+    def _on_stage_pos(self, pos):
+        """
+        Called when the stage is moved, enable the tab if position is imaging mode, disable otherwise
+        :param pos: (dict str->float or None) updated position of the stage
+        """
+        if self.main_data.role == "enzel":
+            # TODO: IMAGING position should not be used on ENZEL anymore
```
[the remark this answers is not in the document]
pieleric: I've started to update the move procedure to drop IMAGING. So soon it should not be an issue. Let's not worry about this for now.

src/odemis/gui/cont/features.py
```diff
@@ -74,6 +75,41 @@ def _on_btn_go_to_feature(self, _):
         self._main_data_model.stage.moveAbs({'x': pos[0], 'y': pos[1]})
         self._main_data_model.focus.moveAbs({'z': pos[2]})
 
+    def _enable_feature_ctrls(self, enable):
+        self._panel.cmb_feature_status.Enable(enable)
```
Kleijwegt, line 79: Please add the doctring:
       
       """
       enables/disables the feature controls

        :param (boolean) enable: enable/disable the feature controls 
        """

src/odemis/gui/cont/features.py
```diff
@@ -82,14 +118,15 @@ def _on_features_changes(self, features):
         if not features:
             # Clear current selections
             self._panel.cmb_features.Clear()
-            self._on_current_feature_changes(None)
+            # currentFeature should also be set to None, which will disable the other widgets
+            # self._on_current_feature_changes(None)
```
Kleijwegt, line 122: I think this line should be uncommented?

Kleijwegt: or removed maybe?

PR #2095, opened 2022-03-31
Fixing an error in the z localization test case due to failing import…

Kleijwegt:
… of a library

### DIFF
--- a/src/odemis/acq/align/test/z_localization_test.py
+++ b/src/odemis/acq/align/test/z_localization_test.py
@@ -40,6 +40,17 @@ z_stack_step_size = 50*10-9  # m
 PRECISION = z_stack_step_size * 0.45  # Precision should be better than the step within a z stack
 
 class TestDetermineZPosition(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        try:
+            from skimage import io, exposure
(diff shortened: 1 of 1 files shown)


src/odemis/driver/test/omicronxx_test.py
```diff
@@ -63,7 +63,7 @@ def test_scan_multi(self):
             dev.terminate()
 
 
-class TestGenericxX(object):
+class TestGenericxX(unittest.TestCase):
```
tepals, line 66: I think this might be a base test class. There is no setup class and if you read the test cases it requires a `self.dev` attribute, which is not available, so the test cases will fail. Now that you let it inherit from `unittest.TestCase` the unittest library will see it as a test class, but it will fail.

You could consider renaming this to something like BaseTestGenericxX, and removing your current changes.

tepals (changes requested):

src/odemis/driver/test/omicronxx_test.py
Kleijwegt: I thought these changes where still on another branch. But I will try if it fixes the problem from that other branch. Thanks!

pieleric (approved):

tepals (approved):

PR #2097, opened 2022-03-31
Acquisition tests: skip test cases that require the nikonc driver.

tepals:
The nikonc driver is not easily available for python3. When tests were run that required this driver the backend failed to start. In the setupclass it is now checked if the driver is available and otherwise the tests are skipped

### DIFF
--- a/src/odemis/acq/test/flim_test.py
+++ b/src/odemis/acq/test/flim_test.py
@@ -31,7 +31,6 @@ from odemis.util import test
 import os
 import time
 import unittest
-from unittest.case import skip
 
 logging.getLogger().setLevel(logging.DEBUG)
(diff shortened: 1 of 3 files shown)


Kleijwegt (approved):

src/odemis/acq/test/flim_test.py
```diff
@@ -54,6 +53,12 @@ class TestFlim(unittest.TestCase):
 
     @classmethod
     def setUpClass(cls):
+        try:
+            # The nikonc driver only works for python2.
```
pieleric, line 57: No, that's not really true. Let's write "The nikonc driver needs omniorb which is not packaged in Ubuntu anymore".

pieleric (changes requested):

pieleric (approved):

PR #2101, opened 2022-04-05
config: update authors and copyright date

pieleric:
Also add a new CITATION.cff file for making it easier to indicate in a
paper that Odemis was used. (for whoever feels like, there is no
requirement to do so!)


CITATION.cff
```diff
@@ -0,0 +1,62 @@
...
+    given-names: Éric
+    affiliation: Delmic B.V.
+    email: piel@delmic.com
+  - family-names: de Laat
+    given-names: Rinze
+    affiliation: Delmic B.V.
+  - family-names: Tsitsikas
+    given-names: Kimon
+    affiliation: Delmic B.V.
+  - family-names: Winkle
```
Sabrina85, line 16: WInkler

PR #2102, opened 2022-04-08
bugreporter: allow to select topic and small fixes

pieleric:


src/bugreporter/test/odemis_bugreporter_test.py
```diff
@@ -171,6 +171,8 @@ def test_compress_file(self):
         os.remove(self.bugreporter.zip_fn)
 
     def test_window(self):
+        # Note: UIActionSimulator.Text() doesn't seem to work on wxPython 4.0.7,
+        # but it works again on wxPython 4.1.
```
tepals, line 175: Would it make sense to check the wx version here and do some logging, or do all our systems use 4.1?

src/bugreporter/test/odemis_bugreporter_test.py
```diff
@@ -184,7 +186,7 @@ def test_window(self):
         # Fill up the form
         gui.name_ctrl.SetFocus()
         # TODO: how to simulate typing non ascii-characters? .Char() + modifiers?
-        # sim.Text(u"Tstingteam member")
+        # sim.Text(b"Tstingteam member")
         gui.name_ctrl.SetValue(u"TÉstingteam member")
```
tepals, line 190: doesn't this one need to be a bytes

src/bugreporter/odemis_bugreporter.py
```diff
@@ -66,6 +71,16 @@
 BACKEND_STOPPED = 2
 BACKEND_STARTING = 3
 
+# Topic IDs are used to indicate which "department" will be taking care of the ticket.
+# They can be found from the urls at https://support.delmic.com/scp/helptopics.php
```
tepals, line 75: The link doesn't bring me to the helptopics, it just redirects to https://support.delmic.com/scp/index.php. And does it make sense to add a fastem topic?

src/bugreporter/odemis_bugreporter.py
```diff
@@ -469,21 +488,51 @@ def __init__(self, controller):
...
+        topic_str = config.get('REPORT', 'topic', fallback=str(TOPIC_ID_DEFAULT))
+        try:
+            try:
+                topic_id = TOPIC_IDS[topic_str]
+            except KeyError:  # Not a known department
+                # Maybe it's just an int?
+                val = int(topic_str)
+                if val <= 0:
+                    raise ValueError("Topic ID must be > 0")
+                topic_id = val
```
tepals, line 530: I'm personally not a fan of nested try except loops, it makes it a bit hard to read. I would go for something like:
```python
topic_id = TOPIC_IDS.get(topic_str, None)
if topic_id is None:  # Not a known department
    # Maybe it's just an int?
    val = int(topic_str)
    if val <= 0:
        raise ValueError("Topic ID must be > 0")
    topic_id = val
```

Do whatever you think is most readable!

tepals (changes requested):

src/bugreporter/odemis_bugreporter.py
tepals: looks good now

PR #2105, opened 2022-04-21
progressive batch future: fix start time in progress update

Sabrina85:
When the progress was set the start time was not correctly updated.
If the sub-future was cancelled, also cancel the batch future.
Add test cases for batch future.

### DIFF
--- a/src/odemis/model/_futures.py
+++ b/src/odemis/model/_futures.py
@@ -509,35 +509,56 @@ class ProgressiveBatchFuture(ProgressiveFuture):
         start = time.time()  # start time of the batch future
         super().__init__(start=start, end=start + sum(self.futures.values()))
         self.set_running_or_notify_cancel()
+        self.task_canceller = self._cancel_all  # takes care of cancelling the task (=all sub-futures)
 
         for f in self.futures:
-            f.add_update_callback(self._on_future_update)  # called whenever set_progress of any single future is called
-            f.add_done_callback(self._on_future_done)  # called when future is done
+            f.add_update_callback(self._on_future_update)  # called whenever set_progress of a sub-future is called
+            f.add_done_callback(self._on_future_done)  # called when a sub-future is done
 
     def _on_future_update(self, f, start, end):
         """
-        Whenever progress on the single future is reported, the progress (or end) for the batch future
+        Whenever progress on the single sub-future is reported, the progress (or end) for the batch future
         is updated accordingly.
 
+        :param f: (ProgressiveFuture) A single sub-future.
         :param start: (float) Start time of a single future in the batch future in seconds.
         :param end: (float) End of a single future in the batch future in seconds.
         """
         if f.running():  # only care about the future which is currently running
-            self.futures[f] = end - start  # update the time estimation of the time left of that future
-            self.set_progress(end=self._estimate_end())  # set the progress for batch future
+            # For a running future it is of interest how much time is still needed for execution until finished.
+            # If the future is running, it is no longer of interest when the future has started (which is now in the
+            # past), but how much time is still needed between now and the estimated end.
+            now = time.time()
+            # Update the time estimation for the running future with the time still needed to finish this future
+            self.futures[f] = end - max(start, now)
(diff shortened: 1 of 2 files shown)


src/odemis/model/test/futures_test.py
```diff
@@ -260,44 +266,65 @@ def testCancelPostEnd(self):
 
         self.assertEqual(self.cancelled, 0)
 
-    def testProgressiveFuture(self):
+    # TODO create own class for progressive future
```
tepals, line 269: Why not just do that directly?

src/odemis/model/test/futures_test.py
```diff
@@ -310,28 +337,205 @@ def testPF_get_progress(self):
...
+        # check estimated time for batch future is sum of estimated time for sub-futures
+        self.assertEqual(self.end - self.start, fs[f1] + fs[f2])
+
+        # "start" the sub-tasks; the batch task is already started at creation
+        f1.set_running_or_notify_cancel()
+
+        now = time.time()
+        f1.set_progress(end=now + 2)  # update the progress on one sub-future
+        # check that the estimated end time of the batch futures has been updated
+        self.assertTrue(now + fs[f2] + 1.9 <= self.end <= now + fs[f2] + 2.1)
```
tepals, line 394: This is not very readable, I would change it to something like this:
```
expected_end = now + fs[f2] + 2 
self.assertTrue(expected_end - 0.1 <= self.end <= expected_end + 0.1)
```

tepals: same for the other assertions in this test

tepals (changes requested):

src/odemis/model/test/futures_test.py
Sabrina85: You would not have seen what I changed and what was already there :) I will do it now with the update.

src/odemis/model/_futures.py
```diff
@@ -509,35 +509,56 @@ def __init__(self, futures):
...
+            f.add_done_callback(self._on_future_done)  # called when a sub-future is done
 
     def _on_future_update(self, f, start, end):
         """
-        Whenever progress on the single future is reported, the progress (or end) for the batch future
+        Whenever progress on the single sub-future is reported, the progress (or end) for the batch future
         is updated accordingly.
 
+        :param f: (ProgressiveFuture) A single sub-future.
         :param start: (float) Start time of a single future in the batch future in seconds.
```
Kleijwegt, line 524: Shouldn't this be:

> :param start: (float) Start time of a single future in the batch future in seconds since the epoch.

src/odemis/model/test/futures_test.py
```diff
@@ -1,25 +1,31 @@
 # -*- coding: utf-8 -*-
```
Kleijwegt, line 1: Not your fault but I think this file is missing the shebang while it should have one:

> #!/usr/bin/env python3

src/odemis/model/test/futures_test.py
Sabrina85: I agree. I just need to check then all other tests and change it there as well as it is nicer to have the same style everywhere.

src/odemis/model/_futures.py
Sabrina85: I can add it here. There are more places in the futures module where this is not clarified. If I am looking at the code it looks like that the whole progress update functionality is only added in the progessive future. And there everything seems to depend on start/end values that are with respect to the epoch. But maybe I read this incorrectly.

src/odemis/model/test/futures_test.py
Sabrina85: Should it be there in every file? I found a few more where we do not have it.

src/odemis/model/_futures.py
```diff
@@ -1,19 +1,26 @@
+#!/usr/bin/env python3
```
tepals, line 1: I think this one is not necessary here, you just need it at the top of the test cases.

tepals (approved):

PR #2106, opened 2022-04-22
crosshair: increase thickness crosshair to increase visibility

Sabrina85:
Especially on an image the crosshair is not that clearly visible. Increase the thickness by 1 pixel to increase visibility.


Sabrina85:
Note, this will make the crosshair thicker for all systems. @pieleric is this ok?

tepals (approved):

tepals:
You could also make the linewidth an input argument of the `draw_crosshair` function with default value 1. `draw_crosshair` is called in `odemis.gui.comp.overlay.world.CurrentPosCrossHairOverlay.draw()` which should then also take it as an argument. This function is called in `odemis.gui.comp.viewport.FastEMOverviewViewport.setView()` and in `odemis.gui.comp.viewport.FastEMAcquisitionViewport.setView()`, so you could then add the value 2 there. That way the crosshair will not be thicker in the other systems.

src/odemis/gui/comp/overlay/view.py
```diff
@@ -73,7 +73,7 @@ def _draw_vertical_line(self, ctx, center, size, colour):
         top = center[1] - size
         bottom = center[1] + size
 
-        ctx.set_line_width(1)
+        ctx.set_line_width(2)
```
Kleijwegt, line 76: Do you intended to increase the line width of all cross_hairs on all systems? It never annoyed me that the crosshair was to small.
If it is mainly needed for fast-em this could also be an attribute on the crosshair class which can be set by a setter maybe?

Sabrina85: > Note, this will make the crosshair thicker for all systems. @pieleric is this ok?

Yes, that is with the current implementation the case. It is an open question if this would be useful also on other systems. We got this feedback from user/app and therefore changed it. It is mainly an issue if there is already data acquired and the crosshair is overlaid on the data. There, the visibility is less good. This case would apply for all systems.

Sabrina85:
> You could also make the linewidth an input argument of the draw_crosshair function with default value 1. draw_crosshair is called in odemis.gui.comp.overlay.world.CurrentPosCrossHairOverlay.draw() which should then also take it as an argument. This function is called in odemis.gui.comp.viewport.FastEMOverviewViewport.setView() and in odemis.gui.comp.viewport.FastEMAcquisitionViewport.setView(), so you could then add the value 2 there. That way the crosshair will not be thicker in the other systems.

Yes, I thought about this and it is an option, but also maybe over the top? Therefore, I went first for the easy change. If we decide it makes also sense for other systems, we do  not need to make this effort. I can imagine that for other systems it is the same problem. If there is data and the crosshair is overlaid, it will be also less visible.

src/odemis/gui/comp/overlay/view.py
Kleijwegt: Sorry I just saw that you included this in a message when you opened this PR. I only looked at the code but skipped that part.

Sabrina85: No worries. Do you have it in Cryo anywhere that you need to move the stage while already having an image where the crosshair is there as overlay? For CL I don't think we have that use-case.

Sabrina85:
I changed it using the composition approach. The test cases making use of the two overlays still pass.

pieleric (approved):
Looks good to me!

PR #2107, opened 2022-04-26
FASTEM: use calibration manager also during acquisition

Sabrina85:
When acquiring an ROA we run a set of calibrations before each ROA. So far there was own code for this in the acquisition manager. However, now that we have the calibration manager, make use of it.

### DIFF
--- a/src/odemis/acq/fastem.py
+++ b/src/odemis/acq/fastem.py
@@ -30,6 +30,8 @@ from concurrent.futures import CancelledError
 
 import numpy
 
+from odemis.acq.align.fastem import Calibrations, align, estimate_calibration_time
+
 try:
     from fastem_calibrations import (
         autofocus_multiprobe,
@@ -61,8 +63,8 @@ class FastEMROA(object):
     and detector.
     """
 
-    def __init__(self, name, coordinates, roc_2, roc_3, asm, multibeam, descanner, detector, overlap=0,
-                 pre_calibrate=False):
+    def __init__(self, name, coordinates, roc_2, roc_3, asm, multibeam, descanner, detector, overlap=0):
+
         """
         :param name: (str) Name of the region of acquisition (ROA). It is the name of the megafield (id) as stored on
                      the external storage.
@@ -81,7 +83,6 @@ class FastEMROA(object):
             The amount of overlap required between single fields. An overlap of 0.2 means that two neighboring fields
             overlap by 20%. By default, the overlap is 0, this means there is no overlap and one field is exactly next
             to the neighboring field.
-        :param pre_calibrate: (bool) If True run pre-calibrations before each ROA acquisition.
         """
         self.name = model.StringVA(name)
         self.coordinates = model.TupleContinuous(coordinates,
@@ -99,7 +100,6 @@ class FastEMROA(object):
         # Automatically updated when the coordinates change.
         self.field_indices = []
         self.overlap = overlap
-        self.pre_calibrate = pre_calibrate
         self.coordinates.subscribe(self.on_coordinates, init=True)
 
         # TODO need to check if megafield already exists, otherwise overwritten, subscribe whenever name is changed,
@@ -121,11 +121,6 @@ class FastEMROA(object):
         """
         field_time = self._detector.frameDuration.value + 1.5  # there is about 1.5 seconds overhead per field
         tot_time = (len(self.field_indices) + 1) * field_time  # +1 because the first field is acquired twice
-
-        if self.pre_calibrate:
-            # TODO replace with call to estimate calibration time in module for each calibrations
-            #  when code here replaced with calls to calibration manager
-            tot_time += 60  # add some extra time for the pre-calibrations
 
         return tot_time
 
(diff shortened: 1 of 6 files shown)


Sabrina85:
@tepals The not sure about the progress update yet. The estimated time takes the pre-calibrations into account. What we do not have is any feedback during the pre-calibrations on the progress as the calibrations themselves are running in a future themselves, but there is no progress update reported while the calibration is running. So far we only report progress between calibrations. However, now we have a batch future which takes the acquisition futures per ROA into account. But it does not know about the sub-future in the acquisition future, which is the alignment/calibration future I believe. 
It might be out of scope for this pr anyways. But I have the feeling if we want to handle progress correctly, the batch future needs to be able to deal with two layers of sub-futures. 

On the other hand, as we do only report progress for calibrations in between two calibrations at the moment and we only run two pre-calibrations, it might be something to consider in future only when there is a progress update during the calibration available.

src/odemis/acq/fastem.py
```diff
@@ -461,73 +484,20 @@ def pre_calibrate(self):
...
-                # put system into state ready for next task
-                # upload=False: it is enough to set calibrated values on HW during following acquisition
-                configure_hw.configure_asm(self._multibeam,
-                                           self._descanner,
-                                           self._detector,
-                                           self._detector.data,
-                                           asm_config,
-                                           upload=False)
-
-            logging.debug("Finish pre-calibration.")
```
tepals, line 530: I think some logging that we are in the pre-calibrations is nice to keep. I would leave this one, and add another `"starting pre-calibration"` at the beginning of this function.

src/odemis/acq/fastem.py
```diff
@@ -461,73 +484,20 @@ def pre_calibrate(self):
...
+        :param stage_pos: (float, float) Stage position where the calibration should be run in meter.
         """
         if not fastem_calibrations:
             raise ModuleNotFoundError("Need fastem_calibrations repository to run pre-calibrations.")
 
-        asm_config = None
+        f = align(self._scanner, self._multibeam,
+                  self._descanner, self._detector,
+                  self._stage, self._ccd,
+                  self._beamshift, self._det_rotator,
```
tepals, line 497: It's not necessary to actually have the detector rotator defined here, you can just pass `None`. That will save a lot of code in the test cases as well. It also feels a bit strange that we would need to always pass the detector rotator to the `AcquisitionTask`, even if we do not even pre-calibrate.

I think this should work:
```
    f = align(self._scanner, self._multibeam,
              self._descanner, self._detector,
              self._stage, self._ccd,
              self._beamshift, None,
              calibrations=pre_calibrations, 
              stage_pos=stage_pos)
              
```

src/odemis/acq/fastem.py
```diff
@@ -349,16 +372,16 @@ def run(self):
...
             # The pre-calibrations should run on a position that lies a full field
             # outside the ROA, therefore temporarily set the overlap to zero.
             overlap_init = self._roa.overlap
             self._roa.overlap = 0
             # Move the stage to the tile with index (-1, -1), to ensure the autofocus and image translation pre-align
             # are done to the top left of the first field, outside the region of acquisition to limit beam damage.
             self.field_idx = (-1, -1)
-            self.move_stage_to_next_tile()
-            self.pre_calibrate()
+            stage_pos_x, stage_pos_y = self.get_abs_stage_movement()
```
tepals, line 383: you could also move this to inside the pre_calibrate function. If I then update it to have the image_translation_prealign at the center of the megafield it will be easier to add. My code in the pre_calibrate function will then be something like this:

```
    self.field_idx = (-1, -1)
    stage_pos = self.get_abs_stage_movement()
    f = align(self._scanner, self._multibeam,
              self._descanner, self._detector,
              self._stage, self._ccd,
              self._beamshift, self._det_rotator,
              calibrations=[Calibrations.OPTICAL_AUTOFOCUS], 
              stage_pos=stage_pos)


    f.result()  # wait for the calibrations to be finished

    self.field_idx = ....... # half of the max of the megafield
    stage_pos = self.get_abs_stage_movement()

    f = align(self._scanner, self._multibeam,
              self._descanner, self._detector,
              self._stage, self._ccd,
              self._beamshift, self._det_rotator,
              calibrations=[Calibrations.IMAGE_TRANSLATION_PREALIGN], 
              stage_pos=stage_pos)

    f.result()  # wait for the calibrations to be finished

```

src/odemis/acq/fastem.py
```diff
@@ -349,16 +372,16 @@ def run(self):
...
+            self.pre_calibrate(self._pre_calibrations, (stage_pos_x, stage_pos_y))
```
tepals, line 384: If you do not use my previous comment you can at least update these two lines. I think this should also work:

```
stage_pos = self.get_abs_stage_movement()
self.pre_calibrate(self._pre_calibrations, stage_pos)
```

tepals (changes requested):

src/odemis/acq/fastem.py
Sabrina85: With "this" you refer to only the stage movement right? I had thought about it before as well. For the field index and overlap I thought it will be easier to leave it outside the pre calibration function to provide more flexibility. But that you had also in mind I guess?

Sabrina85: We would need to define `self._pos_first_tile` in the test case then. Otherwise, the test on the pre-calibration fails as this attribute is still None.

tepals: > With "this" you refer to only the stage movement right? I had thought about it before as well. For the field index and overlap I thought it will be easier to leave it outside the pre calibration function to provide more flexibility. But that you had also in mind I guess?

Yes, that's what I had in mind.

> We would need to define self._pos_first_tile in the test case then. Otherwise, the test on the pre-calibration fails as this attribute is still None. What do you think?

I think we already define the `self._pos_first_tile` in a few test cases.

src/odemis/acq/test/fastem_test.py
```diff
@@ -323,6 +326,7 @@ def setUpClass(cls):
             role="stage")  # TODO replace with stage-scan when ROA conversion method available
         cls.ccd = model.getComponent(role="diagnostic-ccd")
         cls.beamshift = model.getComponent(role="ebeam-shift")
+        cls.det_rotator = model.getComponent(role="det-rotator")
```
tepals, line 329: This one and the one on line 569 can now go, right?

tepals (commented):
one last comment, the rest looks good

src/odemis/acq/test/fastem_test.py
```diff
@@ -562,6 +565,7 @@ def setUpClass(cls):
...
```
tepals, line 568: this one can also go I think

tepals (approved):

PR #2108, opened 2022-04-28
Getting the orsay test cases to pass again

Kleijwegt:
Two TODO are left because they should be improved in the future.

### DIFF
--- a/src/odemis/driver/orsay.py
+++ b/src/odemis/driver/orsay.py
@@ -67,6 +67,18 @@ NO_ERROR_VALUES = (None, "", "None", "none", 0, "0", "NoError")
 
 INTERLOCK_DETECTED_STR = "Interlock event detected"
 
+def recursive_getattr(obj, attr):
+    """
+    Get a named attribute from an object; getattr(x, 'y.z') is equivalent to x.y.z.
+    If a "." is included in the attribute name, recursively it will find the sub-attribute.
+
+    :param obj: Object whose which attribute is to be acquired.
+    :param attr (string): The (sub-) attribute.
+    :return: the value of the specified (sub-) attribute from the specified object.
+    """
+    for a in attr.split("."):
+        obj = getattr(obj, a)
+    return obj
 
 def get_orsay_param_connectors(obj):
     """
@@ -248,6 +260,14 @@ class OrsayComponent(model.HwComponent):
         Update the VA's. Should be called after reconnection to the server
         """
         self._updateProcessInfo()
+        # TODO to consistently pass the some tests an update of all the VAs in the orsay server is added. This
+        #  suggests that the subscription doesn't work properly. Maybe this should be checked or the update_VAs
+        #  should be run in a thread continuously
+        # for child in self.children.value:
+        #     try:
+        #         child.update_VAs()
+        #     except AttributeError:  # if the child does not have an update_VAs() method
+        #         pass  # no need to do anything
 
     def _connection_monitor(self):
         """
@@ -304,10 +324,6 @@ class OrsayComponent(model.HwComponent):
         """
         if parameter is None:
             parameter = self.datamodel.HybridPlatform.ProcessInfo
-        if parameter is not self.datamodel.HybridPlatform.ProcessInfo:
-            raise ValueError("Incorrect parameter passed to _updateProcessInfo. Parameter should be "
-                             "datamodel.HybridPlatform.ProcessInfo. Parameter passed is %s"
-                             % parameter.Name)
         if attr_name != "Actual":
             return
         currentProcessInfo = str(parameter.Actual)
@@ -405,10 +421,6 @@ class pneumaticSuspension(model.HwComponent):
         """
         if parameter is None:
             parameter = self._valve
-        if parameter is not self._valve:
-            raise ValueError("Incorrect parameter passed to _updatePower. Parameter should be "
-                             "datamodel.HybridPlatform.ValvePneumaticSuspension.IsOpen. Parameter passed is %s"
-                             % parameter.Name)
         if attr_name != "Actual":
             return
         valve_state = int(parameter.Actual)
@@ -434,10 +446,6 @@ class pneumaticSuspension(model.HwComponent):
         """
         if parameter is None:
             parameter = self._gauge
-        if parameter is not self._gauge:
-            raise ValueError("Incorrect parameter passed to _updatePressure. Parameter should be "
-                             "datamodel.HybridPlatform.Manometer2.Pressure. Parameter passed is %s"
-                             % parameter.Name)
         if attr_name != "Actual":
             return
         self.pressure._set_value(float(parameter.Actual), force_write=True)
@@ -450,12 +458,6 @@ class pneumaticSuspension(model.HwComponent):
         :param (Orsay Parameter) parameter: the parameter on the Orsay server to use to update the VA
         :param (str) attr_name: the name of the attribute of parameter which was changed
         """
-        if parameter is not self.parent.datamodel.HybridPlatform.ValvePneumaticSuspension.ErrorState and parameter is \
-                not self.parent.datamodel.HybridPlatform.Manometer2.ErrorState and parameter is not None:
-            raise ValueError("Incorrect parameter passed to _updateErrorState. Parameter should be "
-                             "datamodel.HybridPlatform.ValvePneumaticSuspension.ErrorState or "
-                             "datamodel.HybridPlatform.Manometer2.ErrorState or None. Parameter passed is %s"
-                             % parameter.Name)
         if attr_name != "Actual":
             return
         eState = ""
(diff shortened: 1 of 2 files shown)


src/odemis/driver/test/orsay_test.py
```diff
@@ -2057,46 +2060,55 @@ def connector_test(test_case, va, parameters, valuepairs, readonly=False, hw_saf
         for i in range(len(parameters)):
             for a in attributes:
                 setattr(parameters[i], a, par_value1[i])
+        # TODO to consistently pass the some tests an update of all the VAs in the orsay server is added. This
+        #  suggests that the subscription doesn't work properly. Maybe this should be checked or the update_VAs
+        #  should be runned in a thread continuously
+        test_case.oserver.update_VAs()  # Ensure the VAs will be updated
```
pieleric, line 2066: I don't think that's the right way to deal with it. Basically, this test was supposed to check that when the Orsay parameter is changed, the change is detected, and the VA automatically updated. By adding this call, you're explicitly forcing the VA to be updated. So the test passes, but we are not checking the main behaviour anymore!

I'd suggest to dig more into this issue, and understand when the parameter callbacks are not called. It could be a change in the ConsoleClient API, or even a bug there...

Kleijwegt, line 2066: Change TODO such that the default is to not ensure updating and then uncomment updating line whenever development/debugging

src/odemis/driver/test/orsay_test.py
```diff
@@ -152,6 +151,10 @@ def test_updateProcessInfo(self):
         self.assertEqual(self.oserver.processInfo.value, test_string)
         self.datamodel.HybridPlatform.ProcessInfo.Actual = init_state  # return to value from before test
 
+    # TODO Improve test to have no side effects currently this class should restart the back-end after completing
+    #  this test. This side effects suggests that the reconnection doesn't go completely smooth. --> change "setUp" &
```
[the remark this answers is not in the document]
Kleijwegt: Indeed I mean reinitializing/connection the component, currently if the reconnection test fails also the test_updateProcessInfo test fails.
I have altered the reconnection test such that it does fail and I will add a TODO to the other test to note the cause of the fail.

src/odemis/driver/orsay.py
```diff
@@ -1530,12 +1556,15 @@ def __init__(self, va, parameter, attr_name="Actual", conversion=None, factor=No
...
+                        new_range[1][i] = self._parameter_to_VA_value(upperbound)  # copy it to the va
+                    else:
+                        new_range[1] = self._parameter_to_VA_value(upperbound)  # copy it to the va
 
             if self._va_is_tuple:
-                new_range = (new_range[0][0], new_range[1][0])
+                new_range = (new_range[0], new_range[1])
             else:
-                new_range = (tuple(new_range[0]), tuple(new_range[1]))
+                new_range = tuple(new_range)
```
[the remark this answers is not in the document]
Kleijwegt: In my opinion the rest of this method/class isn't particularly simple actually. Because of the in-explicit distinction between ranges with 1 and 2 limits  ( like [-1, +1] vs [ [-1, -2], [1, 2] ] ) it took my a while to figure out how the exact cause/fix for all the failing test cases.
Even though my solution means more lines of code, I think that the readability increases. The explicit distinction between both type of ranges helps me to understand what is happening. It shows that the same thing is applied for both type of ranges but acknowledges the differences explicitly.

Both implementations work so I in the end it doesn't matter much though.

src/odemis/driver/test/orsay_test.py
Kleijwegt: Maybe we should plan a pair programming session for this. The test useally pass but irregulary fail. I don't really know what the cause is actually this seemed to fix it a bit but to be honest I don't know why exactly and it is not the full fix.

src/odemis/driver/orsay.py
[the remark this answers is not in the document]
pieleric: I do agree that the original version is bit hard to understand. Let's go with your implementation then.

src/odemis/driver/orsay.py
```diff
@@ -2605,7 +2612,7 @@ def _changePower(self, goal):
         :param (float) return: goal
         """
         logging.debug("Turning Chamber light %s." % "on" if goal[0] else "off")
-        self._parameter.Target = goal[0] != 0.0
+        self._parameter.Target = int(goal[0])
```
pieleric, line 2615: The parameter is apparently a boolean, so it'd make more sense to pass a boolean. Also, `_changePower()` is a setter, so it should return the actual value set. So I'd suggest this code:
```python
power = bool(goal[0])
logging.debug("Turning Chamber light %s." % "on" if power else "off")
self._parameter.Target = power
return [1.0 if power else 0.0]
```

src/odemis/driver/test/orsay_test.py
```diff
@@ -674,10 +675,14 @@ def test_updatePositionReservoir(self):
 
         self.gis._reservoirPar.Target = init_flow  # return to value from before test
 
+    # TODO Currently not tested on the real hardware because for now it seems to be unsafe to move the GIS
     def test_moveAbs(self):
         """
         Test movement of the gis to working position and parking position
         """
+        if not TEST_NOHW == "sim":
```
pieleric, line 683: -> `if TEST_NOHW != "sim":`

src/odemis/driver/test/orsay_test.py
```diff
@@ -1199,8 +1210,12 @@ def test_interlockOutSEDTriggered(self):
 
     def test_columnPumpOn(self):
         """Check that the columnPumpOn VA is updated correctly"""
+        settletime=5
+        if not TEST_NOHW == "sim":
```
pieleric, line 1214: Maybe invert the logic:
```
settletime=60
if TEST_NOHW == "sim":
       settletime=5
```

src/odemis/driver/test/orsay_test.py
```diff
@@ -31,6 +31,10 @@
 from odemis.driver import orsay
 from odemis.model import HwError
 from odemis import model
+from odemis.util import timeout
+
+logging.getLogger().setLevel(logging.DEBUG)
+logging.basicConfig(format="%(asctime)s  %(levelname)-7s %(module)s:%(lineno)d %(message)s")
 
 TEST_NOHW = (os.environ.get("TEST_NOHW", "0") != "0")  # Default to Hw testing
```
pieleric, line 39: I think this whole part is broken. It converts to a boolean, and then it's compared to a string.
I'd suggest to copy the same as in `xt_client_test.py`:
```python
TEST_NOHW = os.environ.get("TEST_NOHW", "0")  # Default to Hw testing
if TEST_NOHW == "sim":
    pass
elif TEST_NOHW == "0":
    TEST_NOHW = False
elif TEST_NOHW == "1":
    TEST_NOHW = True
else:
    raise ValueError("Unknown value of environment variable TEST_NOHW=%s" % TEST_NOHW)
```

In addition, there are several checks like `if TEST_NOHW == 1`. It can never happen. It should be just `if TEST_NOHW`.

src/odemis/driver/test/orsay_test.py
```diff
@@ -1269,67 +1284,78 @@ def test_gunOn(self):
...
-            self.fib_source._hvps.HeaterState.Target = orsay.HEATER_RISING
-            sleep(0.5)
-            self.assertTrue(self.fib_source.heater.value)
-
-            self.fib_source._hvps.HeaterState.Target = orsay.HEATER_ERROR
-            sleep(0.5)
-            self.assertFalse(self.fib_source.heater.value)
-            self.assertIsInstance(self.fib_source.state.value, HwError)
-            self.assertIn("FIB source forced to shut down", str(self.fib_source.state.value))
+        if not TEST_NOHW == 1:
```
pieleric, line 1302: Shouldn't it be `if TEST_NOHW == "sim":` ? The comment just below seems to indicate you want to skip when in simulator mode.

src/odemis/driver/test/orsay_test.py
```diff
@@ -1823,7 +1864,7 @@ def setUpClass(cls):
         if TEST_NOHW == "sim":  # Actual stays 0 in simulation, so set it directly
             cls.ov.Actual = 10000
         else:
-            cls.fibbeam.objectiveVoltage.value = 10000
+            cls.fibbeam.objectiveVoltage.value = 14000
```
pieleric, line 1867: Why 14kV? The comment just above says it's the same as the baseLensVoltage, but now it's not anymore. 
Also, why do you set the VA only in the real server? Could we always set this VA, and afterwards, in case it's the simulator, also change the `ov.Actual` to simulate the change of voltage. Probably this last command will update the VA, so it'd have the same behaviour, but that would be clearer.

pieleric (changes requested):

src/odemis/driver/orsay.py
Kleijwegt: It actually is a string of an integer. But setting it with an int also works. Only the values 0 or 1 are allowed. The values True/False actually don't work. So you suggestion doesn't work.

src/odemis/driver/test/orsay_test.py
pieleric: Ok. Then just update the comment to be correct, and it's fine :-)

src/odemis/driver/test/orsay_test.py
```diff
@@ -243,7 +248,7 @@ def test_errorstate(self):
         """
         Check that the state VA is updated properly
         """
-        if not TEST_NOHW == "sim":
+        if TEST_NOHW == 1:
```
pieleric, line 251: that should be `is True`. It works because in Python True == 1, but it looks odd.

pieleric (changes requested):

PR #2110, opened 2022-05-04
Fastem acquisition manager: add and update logging to have more info …

tepals:
…during debugging.

### DIFF
--- a/src/odemis/acq/fastem.py
+++ b/src/odemis/acq/fastem.py
@@ -344,7 +344,10 @@ class AcquisitionTask(object):
 
         # No need to set the start time of the future: it's automatically done when setting its state to running.
         self._future.set_progress(end=time.time() + total_roa_time)  # provide end time to future
-        logging.info("Starting acquisition of mega field, with expected duration of %f s", total_roa_time)
+        logging.info(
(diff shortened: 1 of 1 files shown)


src/odemis/acq/fastem.py
```diff
@@ -452,6 +454,8 @@ def acquire_roa(self, dataflow):
             expected_time = len(self._fields_remaining) * total_field_time
             self._future.set_progress(start=time.time(), end=time.time() + expected_time)
 
+        logging.debug("Finished acquiring all fields of ROA.")
```
Sabrina85, line 457: This one seems to be redundant to the log in line 410.

Sabrina85 (approved):

src/odemis/acq/fastem.py
tepals: I wanted to have one to be sure that all fields were acquired successfully, the one on line 410 is also logged when the acquisition was canceled or an exception was raised. I could update this one to `"Successfully acquired all fields of ROA."`

Sabrina85: Yes I think this one is nice.

PR #2113, opened 2022-05-06
Github workflow: added test cases from dataio and acq.drift and remov…

tepals:
…ed tests from model because they kept failing.

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -28,7 +28,12 @@ jobs:
         sudo mkdir /var/run/odemisd
         sudo chmod a+rw /var/run/odemisd
 
-    - name: Run tests from odemis.model
+    - name: Run tests from odemis.dataio
       run: |
         export PYTHONPATH="$PWD/src:$PYTHONPATH"
-        python3 -m unittest discover -s src/odemis/model/test -p "*_test.py" --verbose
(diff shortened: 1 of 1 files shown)


tepals:
@pieleric I tried a few other test cases but I'm with most of them I still got the backend failed to start errors. If you have suggestions for test cases to add here, let me know!

pieleric (approved):

pieleric:
> I tried a few other test cases but I'm with most of them I still got the backend failed to start errors. If you have suggestions for test cases to add here, let me know!

Indeed, most of the test cases somehow rely on the backend... or a screen (for the GUI). We could add quite a few test cases from odemis.util (but would need to list them explicitly). The best would still need to find out what's going on with this backend timeout when stopping it.

tepals:
> > I tried a few other test cases but I'm with most of them I still got the backend failed to start errors. If you have suggestions for test cases to add here, let me know!
> 
> Indeed, most of the test cases somehow rely on the backend... or a screen (for the GUI). We could add quite a few test cases from odemis.util (but would need to list them explicitly). The best would still need to find out what's going on with this backend timeout when stopping it.

yes, I also thought about listing some tests from odemis.util explicitly, but I still hope we find out what's going on with the backend timeout.

PR #2117, opened 2022-05-11
FASTEM: some improvements to the calibration manager run method

Sabrina85:


src/odemis/acq/align/fastem.py
```diff
@@ -273,95 +273,74 @@ def run(self):
...
             translation = calibration.value.run(self._scanner, self._multibeam, self._descanner, self._detector,
                                                 self._dataflow, plot_cell_translation=True)
             self.asm_config["mppc"]["cellTranslation"] = translation
-            configure_asm(self._multibeam, self._descanner, self._detector, self._dataflow, self.asm_config,
-                          upload=False)
+
+        else:
+            raise ValueError(f"Unknown calibration {calibration.name}.")
+
+        configure_asm(self._multibeam, self._descanner, self._detector, self._dataflow, self.asm_config, upload=False)
```
pieleric, line 343: BTW, why do you reconfigure the ASM after every calibration? Is it just in case we would need to run an acquisition afterwards? If so, shouldn't the acquisition do that step? Or is it to store some of the new calibration values?

pieleric (approved):

src/odemis/acq/align/fastem.py
Sabrina85: Good question. We needed to do so to set the new values on the VAs, which the next calibration was relying on. However, we also changed our implementation to make use of special MD that stores nowadays the values that should be used during the calibration to not have this dependency anymore. However, this line still makes sure that the calibrated values are set on the respective VAs and everything that is currently set on the VAs will be also used for the acquisition. So I think it needs to stay. @tepals what do you think?

I will add a comment to clarify why it is needed.

tepals: I think it is indeed in case we need to run an acquisition afterwards. I'm not completely sure anymore why we left it like this and not move it into the acquisition code.

pieleric: If you're sure that each calibration is independent of the current VA value, then it should be safe.
In the code I see that also after running all the calibrations, `configure_asm()` is called one last time. So now it should be enough.

If you wanted to actually apply the settings just before the acquisition, then indeed this asm_config dict would need to obtained in some way. If I understand correctly, the "good" values are now stored as metadata. So it should be possible to have a similar function as `get_config_asm()`, but which would read the values from the metadata, instead of the VAs.

tepals: I am a bit concerned with updating it now, especially since we do not have that much time anymore.

pieleric: I'd suggest to then leave it at the end of the function, and add a TODO indicating that it's probably not useful anymore (as each calib should be independent of the settings).

Sabrina85: ok will do this. I think the idea to obtain it from the MD sounds like a good idea.

tepals (approved):

PR #2118, opened 2022-05-11
driver xt_client: allow to set the brightness and contrast

tepals:
For the overview imaging it is necessary to adjust the brightness and contrast beforehand.
Add a brightness and a contrast VA to allow setting these values.

### DIFF
--- a/src/odemis/driver/test/xt_client_test.py
+++ b/src/odemis/driver/test/xt_client_test.py
@@ -545,6 +545,38 @@ class TestMicroscope(unittest.TestCase):
         self.assertEqual(self.detector.brightness.value, init_brightness + 0.01)
         self.detector.brightness.value = init_brightness
 
+    @unittest.skipIf(TEST_NOHW, 'Contrast and brightness do not change on the simulated hardware.')
+    def test_brightness_contrast(self):
+        """Test that changing the contrast and brightness updates the image correctly."""
+        # Test that for brightness and contrast 1 all values in the image are equal to 255
+        init_brightness = self.detector.brightness.value
+        init_contrast = self.detector.contrast.value
+        self.detector.brightness = 1
+        self.detector.contrast = 1
+        image = self.microscope.get_latest_image(self.scanner.channel)
+        numpy.testing.assert_array_equal(image, 255)
+
+        # Test that decreasing the brightness results in a darker image
+        self.detector.brightness = 0.5
+        darker_image = self.microscope.get_latest_image(self.scanner.channel)
+        self.assertGreater(numpy.sum(image), numpy.sum(darker_image))
+
+        # Test that for brightness and contrast 0 all values in the image are equal to 0
+        self.detector.brightness = 0
+        self.detector.contrast = 0
+        image = self.microscope.get_latest_image(self.scanner.channel)
+        numpy.testing.assert_array_equal(image, 0)
+
+        # Test that increasing the brightness and contrast results in a lighter image
+        self.detector.brightness = 0.5
+        self.detector.contrast = 0.5
+        lighter_image = self.microscope.get_latest_image(self.scanner.channel)
+        self.assertGreater(numpy.sum(lighter_image), numpy.sum(image))
+
+        # Set back initial values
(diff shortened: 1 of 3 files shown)


src/odemis/driver/xt_client.py
```diff
@@ -1381,6 +1481,14 @@ def _setHorizontalFoV(self, fov):
         self.magnification.notify(mag)
         return fov
 
+    def _setBrightness(self, brightness):
+        self.parent.set_brightness(brightness)
```
pieleric, line 1485: No channel name?

src/odemis/driver/xt_client.py
```diff
@@ -1272,6 +1364,14 @@ def _updateSettings(self):
                 self.magnification._value = mag
                 self.horizontalFoV.notify(fov)
                 self.magnification.notify(mag)
+            brightness = self.parent.get_brightness()
```
pieleric, line 1367: No channel_name? `self.channel`?

pieleric (changes requested):
Looks alright. As the unit and range are received from the xtlib-adapter, I can't see what exactly they have.  Typically, the only other driver that already provides these VAs has values within [0, 1], with unit "". (empty). Is that the same? In particular, I'd rather avoid "%" as unit (which is sort of implied by the docstring of `set_brightness()`).

Another remark: the VAs are on the e-beam, which is a little odd as it's a detector setting. That said, `applyAutoContrastBrightness()` is also there, so at least there is some consistency. Now that the `Detector` has a .scanner, it should be possible to move everything to the detector.

One missing part of your PR is one or two new test cases that check the new VAs.

tepals:
> Looks alright. As the unit and range are received from the xtlib-adapter, I can't see what exactly they have. Typically, the only other driver that already provides these VAs has values within [0, 1], with unit "". (empty). Is that the same? In particular, I'd rather avoid "%" as unit (which is sort of implied by the docstring of `set_brightness()`).

It is a percentage in the xtadapter as that is what xtlib required, I'll update the xtadapter to accept a value between 0 and 1.

> Another remark: the VAs are on the e-beam, which is a little odd as it's a detector setting. That said, `applyAutoContrastBrightness()` is also there, so at least there is some consistency. Now that the `Detector` has a .scanner, it should be possible to move everything to the detector.
Indeed, I put these values on the scanner because that's where `applyAutoContrastBrightness()` is. I'll move everything to the detector.

> One missing part of your PR is one or two new test cases that check the new VAs.
will add them!

tepals:
Added an accompanying PR on bitbucket, still need to do more thorough testing before we merge this.

pieleric (changes requested):

src/odemis/driver/xt_client.py
```diff
@@ -2093,7 +2210,6 @@ def applyAutofocus(self, detector):
         This call is non-blocking.
 
         :param detector (str): Role of the detector.
```
pieleric, line 2212: That's incorrect. It's not a str, but actually the Detector itself.

pieleric (changes requested):

src/odemis/driver/test/xt_client_test.py
```diff
@@ -995,6 +1009,34 @@ def test_beamlet_index_info(self):
...
+        new_brightness = init_brightness + 0.01
+        brightness_range = self.microscope.brightness_info()["range"]
+        if new_brightness < brightness_range[0]:
+            new_brightness = brightness_range[0] + 0.01
+        elif new_brightness > brightness_range[1]:
+            new_brightness = brightness_range[1] - 0.01
+        self.microscope.set_brightness(new_brightness)
+        self.assertAlmostEqual(new_brightness, self.microscope.get_brightness())
+        # Set dwell time back to initial value
+        self.microscope.set_brightness(init_brightness)
```
Kleijwegt, line 1038: I already told you via chat but otherwise it might look as if I don't want to review your PR:
Maybe also a test should also be included (on the real Hw) which tests if the brightness of acquired images changes when you change the contrast/brightness

PR #2119, opened 2022-05-11
Acq align: move FindEbeamCenter to a separate module.

tepals:
In odemis.acq.align.fastem the fastem calibrations module was imported which eventually led to the import: from odemis.acq.align import FindEbeamCenter.
Moving FindEbeamCenter to a separate module prevents the circular import.


tepals:
Not sure this is the best way to do this, so if there is a better way: let me know!

src/odemis/acq/align/__init__.py
```diff
@@ -22,102 +22,8 @@
 
 from __future__ import division
 
-import logging
-from odemis.model import MD_PIXEL_SIZE
 from .autofocus import AutoFocus, AutoFocusSpectrometer, Sparc2AutoFocus
 from .light import turnOnLight
 from .find_overlay import FindOverlay
 from .spot import AlignSpot, FindSpot
 from .fastem import align
```
pieleric, line 29: Wouldn't the circular dependency also be solved by not importing this function? (ie, deleting this line).

Typically we have these imports in order to hide the fact that the functions are placed in submodules. However, typically these "hidden" submodules have names starting with a `_`. Anyway, the `align()` function has a too generic name to just be loaded as `acq.align.align()`.

pieleric (changes requested):

src/odemis/acq/align/__init__.py
tepals: I was sure I had tried that, but you are correct it works! And I agree that the `align` function has a too generic name to be loaded like that.

I will move back FindEbeamCenter and remove the line `from .fastem import align`

pieleric (approved):

PR #2120, opened 2022-05-11
gui cont tabs: allows to select multiple files to open in analysis tab

pieleric:
It's possible to open multiple files simultaneously, but until now the user had to
select each file one at a time using the "From file..." button.
Now, it's directly possible to select multiple files. Either from the
"open" button, or from the "From file...".

Handy when trying to open quite a few files at the same time, like
tiles.

### DIFF
--- a/src/odemis/gui/cont/tabs.py
+++ b/src/odemis/gui/cont/tabs.py
@@ -3401,7 +3401,7 @@ class AnalysisTab(Tab):
                                message="Choose a file to load",
                                defaultDir=path,
                                defaultFile="",
-                               style=wx.FD_OPEN | wx.FD_FILE_MUST_EXIST,
+                               style=wx.FD_OPEN | wx.FD_FILE_MUST_EXIST | wx.FD_MULTIPLE,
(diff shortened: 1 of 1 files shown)


src/odemis/gui/cont/tabs.py
```diff
@@ -3401,24 +3401,25 @@ def select_acq_file(self, extend=False):
...
-        # popup.show_message(self.main_frame, "Opening file")
-        self.load_data(filename, fmt, extend=extend)
+        for filename in dialog.GetPaths():
+            if extend:
+                logging.debug("Extending the streams with file %s", filename)
+            else:
+                logging.debug("Current file set to %s", filename)
+
+            self.load_data(filename, fmt, extend=extend)
+            extend = True  # If multiple files loaded, the first one is considered main one
```
Sabrina85, line 3421: Now extend is always set to True no matter what it was just in the lines before where you either load additional data or load new data. Is that what you want?

Sabrina85 (approved):

PR #2124, opened 2022-05-16
acq drift test: improve reliability

pieleric:
Force the "randomness" to always be the same one so that the tests
always pass or fail, independent of when you're running them.

### DIFF
--- a/src/odemis/acq/drift/test/shift_test.py
+++ b/src/odemis/acq/drift/test/shift_test.py
@@ -23,13 +23,11 @@ from __future__ import division
 
 import math
 from numpy import fft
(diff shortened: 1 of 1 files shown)


tepals:
Should we take this into consideration? https://towardsdatascience.com/stop-using-numpy-random-seed-581a9972805f

pieleric:
I wasn't aware of it. I think for the test cases it's a tiny bit over the top, but in the other hand it's good to have an example usage, so I'll switch the code to it.

pieleric:
@tepals , new version ready for review.

tepals (approved):

PR #2125, opened 2022-05-16
FASTEM: some clean-ups and improved logging

Sabrina85:


src/odemis/gui/cont/project.py
```diff
@@ -468,8 +467,8 @@ def __init__(self, tab_data, calibration_panel, viewport, calib_prefix):
         # Only enable buttons for scintillators which have been selected in the chamber tab
         tab_data.main.active_scintillators.subscribe(self._on_active_scintillators)
 
-        tab_data.main.is_acquiring.subscribe(self._on_is_acquiring)  # enable/disable calib button during acquisition
-        tab_data.is_calibrating.subscribe(self._on_is_acquiring)  # enable/disable calib button during calibration
+        tab_data.main.is_acquiring.subscribe(self._on_is_calibrating)  # enable/disable calib button during acquisition
```
tepals, line 470: shouldn't this be `self._on_is_acquiring`? Or is the comment incorrect?

src/odemis/gui/cont/fastem_acq.py
```diff
@@ -787,10 +791,12 @@ def _on_calibration_done(self, future, roc_num):
         try:
             config = future.result()  # wait until the calibration is done
             self.save_calibrated_settings(roc_num, config)  # save calibrated
+            logging.debug("Finished calibration step %s successfully for ROC number %s", self._calib_prefix, roc_num)
         except CancelledError:
             pass  # nothing to do here, callback of batch future takes care of everything
         except Exception as ex:
-            logging.exception("Calibration failed with %s.", ex)  # callback of batch future takes care of the rest
+            logging.exception("Calibration step %s failed for ROC number %s with %s.",
```
tepals, line 798: I would say `... ROC number %s with exception: %s."`

tepals (changes requested):

src/odemis/gui/cont/project.py
Sabrina85: Here, we check whether an acquisition (`is_acquiring`) or a calibration (`is_calibrating`) is ongoing. If so, in both cases `_on_is_calibrating` is called and the calibration panels get enabled or disabled depending on the state (`is_acquiring`/`is_calibrating`). The acquisition panel (button to acquire ROAs) is handled elsewhere.

tepals (approved):

PR #2126, opened 2022-05-17
util conversion: wave2rgb refactored to become more readable and redu…

tepals:
…ce the number of lines.

### DIFF
--- a/src/odemis/util/conversion.py
+++ b/src/odemis/util/conversion.py
@@ -31,49 +31,38 @@ from odemis import model
 import re
 import yaml
 
-# Inspired by code from:
-# http://codingmess.blogspot.nl/2009/05/conversion-of-wavelength-in-nanometers.html
-# based on:
-# http://www.physics.sfasu.edu/astro/colour/spectra.html
-def wave2rgb(wavelength):
-    """
-    Convert a wavelength into a (r,g,b) value
-    wavelength (0<float): wavelength in m
-    return (3-tupe int in 0..255): RGB value
-    """
-    w = wavelength * 1e9
-    # outside of the visible spectrum, use fixed colour
-    w = min(max(w, 350), 780)
 
-    # colour
-    if 350 <= w < 440:
-        r = -(w - 440) / (440 - 350)
-        g = 0
-        b = 1
-    elif 440 <= w < 490:
-        r = 0
-        g = (w - 440) / (490 - 440)
-        b = 1
-    elif 490 <= w < 510:
-        r = 0
-        g = 1
-        b = -(w - 510) / (510 - 490)
-    elif 510 <= w < 580:
-        r = (w - 510) / (580 - 510)
-        g = 1
-        b = 0
-    elif 580 <= w < 645:
-        r = 1
-        g = -(w - 645) / (645 - 580)
-        b = 0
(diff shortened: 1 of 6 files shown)


src/odemis/util/conversion.py
```diff
@@ -31,49 +31,36 @@
...
-        r = 1
-        g = -(w - 645) / (645 - 580)
-        b = 0
-    elif 645 <= w <= 780:
-        r = 1
-        g = 0
-        b = 0
+    w *= 1e9
+
+    x = lambda w, l, h: round((w - h) / abs((h - l)) * 255)  # maps w between 0 and +/- 1
```
pieleric, line 49: Could you give this function a better name? You don't have to limit yourself to one character ;-)
Also you could add some explanation on what it returns. It seems sometimes it returns negative values.
Moreover, l and h sounds like meaning low/high. But when low < high, it returns negative values... which is odd.

pieleric (changes requested):

src/odemis/util/conversion.py
```diff
@@ -31,49 +31,36 @@
...
-        b = 0
-    elif 580 <= w < 645:
-        r = 1
-        g = -(w - 645) / (645 - 580)
-        b = 0
-    elif 645 <= w <= 780:
-        r = 1
-        g = 0
-        b = 0
+    w *= 1e9
```
Kleijwegt, line 47: Could you add a comment that we go from wavelength in meters to wavelength in nm?

src/odemis/util/conversion.py
```diff
@@ -31,49 +31,36 @@
 import re
 import yaml
 
-# Inspired by code from:
-# http://codingmess.blogspot.nl/2009/05/conversion-of-wavelength-in-nanometers.html
-# based on:
-# http://www.physics.sfasu.edu/astro/colour/spectra.html
-def wave2rgb(wavelength):
+
+def wave2rgb(w):
```
Kleijwegt, line 35: I think we should also update the name of this function to wavelenght2rgb

Kleijwegt (changes requested):

tepals:
I did not mean to add the formatting changes to conversion.py. but since there are not too many I'll leave them (if you guys don't mind).

src/odemis/util/conversion.py
```diff
@@ -31,49 +31,38 @@
...
-    return (3-tupe int in 0..255): RGB value
+    Convert a wavelength into a (r, g, b) value
+    wavelength (0 < float): wavelength in m
+    return (3-tuple int in 0..255): RGB value
+
+    Notes:
+    Inspired by code from:
+    http://codingmess.blogspot.nl/2009/05/conversion-of-wavelength-in-nanometers.html
+    based on:
+    http://www.physics.sfasu.edu/astro/color/spectra.html
```
Kleijwegt, line 45: if this link doesn't work maybe we should hunt down a working one or write a note that it doesn't work.

Kleijwegt (approved):

src/odemis/util/conversion.py
tepals: I already did, and this is the working one

pieleric (approved):

PR #2130, opened 2022-05-19
util units: simplify round_down_significant and improve docstring

pieleric:
From code golf session (only kepts the non-crazy part)

### DIFF
--- a/src/odemis/util/units.py
+++ b/src/odemis/util/units.py
@@ -66,20 +66,22 @@ def round_significant(x, n):
     return round(x, int(n - math.ceil(math.log10(abs(x)))))
 
 
-def round_down_significant(x, n):
+def round_down_significant(x: float, n: int) -> float:
     """
-    Round a number to n significant figures making sure it's smaller
+    Round a number to n significant figures making sure it's smaller.
+    It rounds towards 0, so for negative numbers the absolute value is also reduced.
+    x: the value to round
+    n (>=1): number of significant figures to keep
+    returns: the rounded value
+    Example:
+        round_down_significant(136, 2) -> 130
+        round_down_significant(-0.136, 2) -> -0.13
     """
     if x == 0:
         return 0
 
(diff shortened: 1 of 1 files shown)


src/odemis/util/units.py
```diff
@@ -66,20 +66,19 @@ def round_significant(x, n):
...
         return 0
 
     exp = n - math.ceil(math.log10(abs(x)))
-    if x > 0:
-        ret = math.floor(x * 10 ** exp) / (10 ** exp)
-    else:
-        ret = math.ceil(x * 10 ** exp) / (10 ** exp)
-    # assert(abs(ret) <= abs(x))
-    return ret
+    return math.trunc(x * 10 ** exp) / (10 ** exp)
```
BasHolweg, line 81:
```
>>> n = 2
>>> x =136
>>> exp = n - math.ceil(math.log10(abs(x)))
>>> math.trunc(x * 10 ** exp) / (10 ** exp)
130.0
```
For n = 2, x = 136. The resulting value is a float with more significant numbers then x. Is this expected behaviour. If yes could this be commented in the docstring. If not perhaps make it an interger if first significant decimal number is zero.

BasHolweg (changes requested):

src/odemis/util/units.py
pieleric: That's per design. It *always* returns a float. Actually, in Python you can almost always abstract float vs integer, so we don't really care. The default print() of Python show them differently, but the Odemis GUI doesn't use the default print function ;-) Here, the final number has only two significant figures: 1 & 3. zeros are used to replace non-significant numbers.

tepals: Could you add an example like this to the docstring? It's also not completely clear to me, just by reading the docstring.

pieleric: Added 2 examples to the docstring.

tepals (approved):

PR #2131, opened 2022-05-19
Technolution add custom calibration metadata

tepals:
and PEP 8 clean up

### DIFF
--- a/src/odemis/acq/fastem.py
+++ b/src/odemis/acq/fastem.py
@@ -20,7 +20,7 @@ PURPOSE. See the GNU General Public License for more details.
 You should have received a copy of the GNU General Public License along with
 Odemis. If not, see http://www.gnu.org/licenses/.
 """
-
+import json
 import logging
 import math
 import os
@@ -30,7 +30,7 @@ from concurrent.futures import CancelledError
 
 import numpy
 
-from odemis.acq.align.fastem import Calibrations, align, estimate_calibration_time
+from odemis.acq.align.fastem import align, estimate_calibration_time
 
 try:
     from fastem_calibrations import (
@@ -53,6 +53,36 @@ from odemis.acq import fastem_conf
 
 # The executor is a single object, independent of how many times the module (fastem.py) is loaded.
 _executor = model.CancellableThreadPoolExecutor(max_workers=1)
+
+# TODO: Normally we do not use component names in code, only roles. Store in the roles in the SETTINGS_SELECTION,
+#  and at init lookup the role -> name conversion (using model.getComponent(role=role)).
+# Selection of components, VAs and values to save with the ROA acquisition, structured: {component: {VA: value}}
+SETTINGS_SELECTION = {
+    'Beam Shift Controller': ['shift', ],
+    'Detector Rotator': ['position',
+                         'referenced',
+                         'speed', ],
+    'Mirror Descanner': ['clockPeriod',
+                         'physicalFlybackTime',
+                         'rotation',
+                         'scanAmplitude',
+                         'scanOffset', ],
+    'MultiBeam Scanner': ['clockPeriod',
+                          'dwellTime',
+                          'rotation',
+                          'scanAmplitude',
+                          'scanDelay',
+                          'scanOffset', ],
+    'MultiBeam Scanner XT': ['accelVoltage',
+                             'beamShiftTransformationMatrix',
+                             'multiprobeRotation',
+                             'patternStigmator',
+                             'power',
+                             'rotation', ],
+    'Sample Stage': ['position',
+                     'referenced',
+                     'speed']
+}
 
 
 class FastEMROA(object):
@@ -231,7 +261,8 @@ def estimate_acquisition_time(roa, pre_calibrations=None):
     return tot_time
 
 
-def acquire(roa, path, scanner, multibeam, descanner, detector, stage, ccd, beamshift, lens, pre_calibrations=None):
+def acquire(roa, path, scanner, multibeam, descanner, detector, stage, ccd, beamshift,
+            lens, pre_calibrations=None, settings_obs=None):
     """
     Start a megafield acquisition task for a given region of acquisition (ROA).
 
@@ -254,11 +285,15 @@ def acquire(roa, path, scanner, multibeam, descanner, detector, stage, ccd, beam
     :param lens: (static.OpticalLens) Optical lens component.
     :param pre_calibrations: (list[Calibrations]) List of calibrations that should be run before the ROA acquisition.
                              Default is None.
+    :param settings_obs: (SettingsObserver) VAs of all components of which some will be
+                         integrated in the acquired ROA as metadata. Default is None,
+                         if None the metadata will not be updated.
 
     :return: (ProgressiveFuture) Acquisition future object, which can be cancelled. The result of the future is
              a tuple that contains:
                 (model.DataArray): The acquisition data, which depends on the value of the detector.dataContent VA.
                 (Exception or None): Exception raised during the acquisition or None.
+
     """
 
     est_dur = estimate_acquisition_time(roa, pre_calibrations)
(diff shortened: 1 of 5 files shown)


Sabrina85 (changes requested):
Can you add a test case?

src/odemis/acq/fastem.py
```diff
@@ -285,7 +286,7 @@ class AcquisitionTask(object):
     """
 
     def __init__(self, scanner, multibeam, descanner, detector, stage, ccd, beamshift, lens, roa, path,
-                 pre_calibrations, future):
+                 pre_calibrations, future, settings_obs=None):
```
Sabrina85, line 289: Update the doc string.

src/odemis/acq/fastem.py
```diff
@@ -231,7 +231,8 @@ def estimate_acquisition_time(roa, pre_calibrations=None):
     return tot_time
 
 
-def acquire(roa, path, scanner, multibeam, descanner, detector, stage, ccd, beamshift, lens, pre_calibrations=None):
+def acquire(roa, path, scanner, multibeam, descanner, detector, stage, ccd, beamshift,
+            lens, pre_calibrations=None, settings_obs=None):
```
Sabrina85, line 235: Update the doc string.

src/odemis/acq/fastem.py
Sabrina85: Is it needed that it is None here? The acquire method already makes sure it is either None or an object right?

src/odemis/driver/technolution.py
```diff
@@ -1175,44 +1202,64 @@ def _assembleMegafieldMetadata(self):
...
+    def _get_calibration_metadata(self):
+        """
+        Get the calibration metadata based on the SETTINGS_SELECTION.
+
+        :return: (dict)
+            Nested dictionary containing the current components, VAs and values: {component: {VA: value}}
+        """
+        all_settings = self.getMetadata().get(model.MD_CALIB, None)
+        if all_settings is None:
+            return "Failed getting calibration metadata"
```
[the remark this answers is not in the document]
tepals: yes, this will then be displayed in the custom_data field.

src/odemis/acq/fastem.py
tepals: you are correct, it doesn't matter here

Sabrina85 (approved):

Sabrina85:
Will you add some test?

Sabrina85 (approved):

tepals:
> Will you add some test?

Done! Had missed your comment.

src/odemis/driver/test/technolution_test.py
```diff
@@ -1682,6 +1683,29 @@ def test_assemble_megafield_metadata(self):
...
+        self.MPPC.updateMetadata({model.MD_CALIB: all_settings})
+        calibration_metadata = self.MPPC._get_calibration_metadata()
+        for k in SETTINGS_SELECTION.keys():
+            self.assertTrue(k in calibration_metadata.keys())
+
+        # Test that if the calibration metadata is None, the correct string is returned.
+        self.MPPC.updateMetadata({model.MD_CALIB: None})
+        calibration_metadata = self.MPPC._get_calibration_metadata()
+        self.assertEqual(calibration_metadata,
+                         "Failed getting calibration metadata, something might have gone wrong.")
```
Sabrina85, line 1705: Then hopefully the string never changes :D

tepals: I would hope that is someone updates the code, they also re-run and update the test cases :)

src/odemis/driver/technolution.py
```diff
@@ -1841,7 +1845,7 @@ def __init__(self, url, response, expected_status):
             elif hasattr(response, "content"):
                 self._errorMessageResponse(url, status_code, reason, expected_status, status_code, response.content)
             else:
-                self._emptyResponse(url, status_code, reason, expected_status,)
+                self._emptyResponse(url, status_code, reason, expected_status, )
```
pieleric, line 1895: I think that a proper clean-up is to just drop the comma (same thing line 1922)

src/odemis/driver/technolution.py
```diff
@@ -1172,44 +1202,64 @@ def _assembleMegafieldMetadata(self):
...
+    def _get_calibration_metadata(self):
+        """
+        Get the calibration metadata based on the SETTINGS_SELECTION.
+
+        :return: (dict)
+            Nested dictionary containing the current components, VAs and values: {component: {VA: value}}
+        """
+        all_settings = self.getMetadata().get(model.MD_CALIB, None)
+        if all_settings is None:
+            return "Failed getting calibration metadata, something might have gone wrong."
```
pieleric, line 1254: This is not a dict, while your docstring claims to return a dict.
I wonder whether it's really a good idea anyway to put such string. Maybe you've already discussed this during specification, in which case it's fine. Otherwise, I'd suggest to double check it's a really useful for the user to get such string in the metadata. Wouldn't an empty metadata convey the same information (and be easier to interpret for the computer)? Or just return was used to be set: "no custom data".

src/odemis/driver/technolution.py
```diff
@@ -63,6 +63,33 @@
 ASM_SUBDIR_CHARS = r'[A-Za-z0-9/_()-]*'  # * -> subdirectories can also be empty string
 ASM_FILE_CHARS = r'[A-Za-z0-9_()-]+'
 
+SETTINGS_SELECTION = {
```
pieleric, line 66: Please add a comment above about what it is for and its format.

src/odemis/driver/technolution.py
```diff
@@ -1172,44 +1202,64 @@ def _assembleMegafieldMetadata(self):
...
+            Nested dictionary containing the current components, VAs and values: {component: {VA: value}}
+        """
+        all_settings = self.getMetadata().get(model.MD_CALIB, None)
+        if all_settings is None:
+            return "Failed getting calibration metadata, something might have gone wrong."
+
+        selected_settings = {}
+        for comp, vas in SETTINGS_SELECTION.items():
+            if all_settings.get(comp):
+                selected_settings[comp] = {va: all_settings[comp].get(va) for va in vas}
```
pieleric, line 1259: I'm not convinced this is the right place. The driver shouldn't need to know about the whole microscope file, or about the VAs of other components. Hard-coding the names of the components here feels odd. I'd suggest to have this metadata already contain the info as it should be passed (ie, filtered). The GUI would take care of that.

src/odemis/driver/technolution.py
```diff
@@ -1172,44 +1202,64 @@ def _assembleMegafieldMetadata(self):
...
         return megafield_metadata
 
+    def _get_calibration_metadata(self):
+        """
+        Get the calibration metadata based on the SETTINGS_SELECTION.
+
+        :return: (dict)
+            Nested dictionary containing the current components, VAs and values: {component: {VA: value}}
+        """
+        all_settings = self.getMetadata().get(model.MD_CALIB, None)
```
pieleric, line 1252: `MD_CALIB` is normally used to pass calibration data to the component. Here, it's just a dict that will be stored as custom data. How about using `MD_EXTRA_SETTINGS`, I think it'd be a better fit. Typically it's directly a string (containing a JSON string). So it could be directly used as-is (and the GUI would take care of the filtering + json.dumps().

pieleric (changes requested):
As discussed in my comments, I'd recommend changing a little bit the structure. Move the data filtering to the acquisition code, and the driver code could be as dumb as possible.

src/odemis/acq/fastem.py
```diff
@@ -311,6 +342,9 @@ def __init__(self, scanner, multibeam, descanner, detector, stage, ccd, beamshif
                             (model.DataArray): The acquisition data, which depends on the value of the
                                                detector.dataContent VA.
                             (Exception or None): Exception raised during the acquisition or None.
+        :param settings_obs: (SettingsObserver) VAs of all components of which some will be
+                             integrated in the acquired ROA as metadata.
+
```
pieleric, line 347: Shouldn't this parameter be between `pre_calibrations` and `future`? That last parameter is a bit special as it's an extra parameter compared to `acquire()`, to track the progress and return the result.

src/odemis/acq/fastem.py
```diff
@@ -259,6 +288,8 @@ def acquire(roa, path, scanner, multibeam, descanner, detector, stage, ccd, beam
              a tuple that contains:
                 (model.DataArray): The acquisition data, which depends on the value of the detector.dataContent VA.
                 (Exception or None): Exception raised during the acquisition or None.
+    :param settings_obs: (SettingsObserver) VAs of all components of which some will be
```
pieleric, line 291: Please indicate that None is allowed, and what happens in such case (ie, metadata not updated).

pieleric (changes requested):

pieleric (approved):

PR #2133, opened 2022-05-31
gui cont streams: fix display order of HW Detector VAs

pieleric:
HW VAs are mixed for the emitter and detector, but until now we only
used the emitter order. That meant that if the VAs came from the
detector, the order would be roughly random (except for the VAs who
share the same name).
=> use both orders, one after the other. It's not an issue if a VA is
listed several times, the first occurence will be used.

### DIFF
--- a/src/odemis/gui/cont/streams.py
+++ b/src/odemis/gui/cont/streams.py
@@ -347,7 +347,8 @@ class StreamController(object):
 
         # Process the hardware VAs first (emitter and detector hardware VAs are combined into one
         # attribute called 'hw_vas'
-        vas_names = util.sorted_according_to(list(self.stream.hw_vas.keys()), list(emitter_conf.keys()))
+        vas_names = util.sorted_according_to(list(self.stream.hw_vas.keys()),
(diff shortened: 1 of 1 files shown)


src/odemis/gui/cont/streams.py
```diff
@@ -347,7 +347,8 @@ def _add_hw_setting_controls(self):
 
         # Process the hardware VAs first (emitter and detector hardware VAs are combined into one
         # attribute called 'hw_vas'
-        vas_names = util.sorted_according_to(list(self.stream.hw_vas.keys()), list(emitter_conf.keys()))
+        vas_names = util.sorted_according_to(list(self.stream.hw_vas.keys()),
+                                             list(emitter_conf.keys()) + list(detector_conf.keys()))
```
Kleijwegt, line 351: sorted_according_to actually works also when the second list contains doubles.

Kleijwegt (approved):
@LarsvanRoemburg @BasHolweg
@tepals Has the theory that on average every PR (irrespectable of size) has 10 comments. I challege you to make it up to the 10 comments😏

PR #2137, opened 2022-05-31
gui comp: increase the text fields in height to ensure all characters are shown

pieleric:
By default the text fields are too small, at least with the default
Ubuntu theme. The bottom of letters like p or µ was hidden.

So reduce the margin and increase the text height. It
should actually take a few pixels *less* in total now.


Kleijwegt:
@pieleric  What simulator should I use to test this? I think we should look on multiple computers how these changes look (hint @LarsvanRoemburg 😁

pieleric:
You can just use the Viewer, and open the Test data/pretty data/20140708-134153-spectrum-round-antenna.h5
Looking at the SEM survey, the integration time shows 7.65µs. It should show the whole "µ", like this:
image

In the "path" entry at the top, the "p" of "pretty data" should also be fully visible.

Kleijwegt (approved):

PR #2140, opened 2022-06-02
util units: add also µsec display to the time display

pieleric:
Avoids having values < 1ms represented as "0 s". The µs can still matter
a lot, depending on which time frame we care about.


src/odemis/util/units.py
```diff
@@ -366,6 +371,9 @@ def readable_time(seconds, full=True):
     if msec:
         result.append(u"%d ms" % msec)
 
+    if usec:
```
LarsvanRoemburg, line 374: Was it a deliberate choice to not include a full word option for the ms and us? For the rest it looks good!

PR #2141, opened 2022-06-02
driver andorcam2: add hardware trigger support

pieleric:


src/odemis/driver/andorcam2.py
```diff
@@ -422,6 +431,7 @@ def __getitem__(self, name):
 20099: "DRV_BINNING_ERROR",
 20100: "DRV_INVALID_AMPLIFIER",
 20101: "DRV_INVALID_COUNTCONVERT_MODE",
+20102: "DRV_USB_INTERRUPT_ENDPOINT_TIMEOUT",
```
Kleijwegt, line 434: With such a long list it maybe makes sense to create a file which contains just constants like this one? (ini or just python)
I don't think it required, and maybe it is not that Pythonic, @pieleric what is your view on this?

src/odemis/model/test/dataflow_test.py
```diff
@@ -333,6 +334,26 @@ def test_df_raise_error(self):
...
+    def test_hw_trigger(self):
+        self.df = SynchronizableDataFlow()
+
+        simpleEvent = model.Event()
+
+        # Not allowed to use a HwTrigger to notify (via software call)
+        with self.assertRaises(ValueError):
+            self.df.hwTrigger.notify()
+
+        # This should be fine (but cannot be tested more)
```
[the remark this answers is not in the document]
Kleijwegt: Looking at remote_test.py that is actually done there, is that correct? And if yes I think this comment should refer to that test case

src/odemis/driver/andorcam2.py
```diff
@@ -2938,6 +3058,9 @@ def IsTriggerModeAvailable(self, mode):
         if _val(mode) not in self._supported_triggers:
             raise AndorV2Error(20078, "DRV_INVALID_MODE")
 
+    def SetTriggerInvert(self, inverted):
+        return
```
Kleijwegt, line 3062: This does nothing, is a pass not better instead of a return? Why does it do nothing? Maybe that can be commented

src/odemis/driver/andorcam2.py
```diff
@@ -3031,6 +3154,9 @@ def AbortAcquisition(self):
         self.status = AndorV2DLL.DRV_IDLE
         self.acq_aborted.set()
 
+    def GetOldestImage16(self, cbuffer, size):
+        return self.GetMostRecentImage16(cbuffer, size)
```
Kleijwegt, line 3158: GetOldestImage16 uses the call to mostRecentImage without any minus, inverse or something to annotate that it is actually not getting the most recent image? How does that work?

src/odemis/driver/test/andorcam2_test.py
```diff
@@ -24,15 +24,18 @@
 from __future__ import division
 
 import logging
+from odemis import model
```
Kleijwegt, line 27: I have the feeling more test cases should have be added to this file.
For example I cannot find any test cases for these functions:

```
GetTemperatureRange
GetTriggerLevelRange
_setStaticSettings
```

But looking at the amount of changes I think there are more functions which might deserve some updates for there test cases.

Kleijwegt (changes requested):

src/odemis/driver/test/andorcam2_test.py
pieleric: These functions are private. They shouldn't be called from outside, so I prefer to not test them. We test the public interface, which indirectly relies on these private functions.

src/odemis/model/test/dataflow_test.py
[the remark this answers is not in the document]
pieleric: I added a test to check that the DataFlow waits for the hw trigger (and indeed doesn't generate any data, as it never receive any trigger).

src/odemis/driver/andorcam2.py
pieleric: It does nothing because we don't need to simulate the inversion of the hardware trigger, as it's all simulated. I've added a comment.

src/odemis/driver/andorcam2.py
pieleric: They behave the same because the simulator just generate one image on the fly, there is no image buffer. I'm added a comment.

src/odemis/driver/andorcam2.py
Kleijwegt: I agree, sounds good.

src/odemis/driver/test/andorcam2_test.py
Kleijwegt: The first of these two aren't private isn't it?

Kleijwegt (approved):

PR #2142, opened 2022-06-02
plugins ext_spec_map: new plugin for spectrum acquisition controlled via external e-beam

pieleric:


plugins/ext_spec_map.py
```diff
@@ -0,0 +1,540 @@
...
+        returns:
+            da: the acquired spectrum cube, or None if it was cancelled before
+               any image was received.
+            cancelled: True if the user cancelled the acquisition before the end
+        """
+        cancelled = False
+
+        rep = self.stream.repetition.value
+        num_acqs = rep[0] * rep[1]
+        # frame_dur = self.spectrometer.framePeriod.value
```
tepals, line 232: why is this commented?

plugins/ext_spec_map.py
```diff
@@ -0,0 +1,540 @@
...
+        time.sleep(0.5)
+
+        try:
+            # Indicate it's ready
+            # On the first image received, we will show a progress bar
+            dlg.setAcquisitionInfo("Ready to acquire", logging.WARNING)
+            wx.CallAfter(dlg.gauge_progress.SetValue, 0)
+            wx.CallAfter(dlg.gauge_progress.SetRange, num_acqs)
+
+            # Acquire (even if it was live, to be sure it's the data is up-to-date)
```
tepals, line 259: "it's" can be removed

plugins/ext_spec_map.py
```diff
@@ -0,0 +1,540 @@
...
+
+                # Check it's cancelled (which ever of the 2 signals arrives first)
+                if spec is None or future.cancelled():  # Cancelled?
+                    logging.debug("Stopping acquisition, as it was cancelled")
+                    cancelled = True
+                    break
+
+                specs.append(spec)
+
+                # left_time = (num_acqs - len(specs)) * frame_dur + 0.1
```
tepals, line 272: why is this comment here?

plugins/ext_spec_map.py
```diff
@@ -0,0 +1,540 @@
...
+        # Show the gauge, with the cancel button
+        # dlg.current_future = None
+        self._show_gauge_panel(dlg)
+
+        try:
+            try:
+                da, cancelled = self._acquire_spec(dlg, f)
+            except Exception as e:
+                logging.exception("Failed to acquire CL data: %s", e)
+                return
```
tepals, line 320: now it doesn't reach the finally, right? Don't you still want to execute those commands?

tepals (changes requested):

plugins/ext_spec_map.py
pieleric: Just some left over... removed.

plugins/ext_spec_map.py
pieleric: It *does* reach the `finally`. It's actually sort of the point of `finally`: it's always executed, even if you have a return or an exception... or a return in an exception.

PR #2144, opened 2022-06-07
util test: stop_backend updated to call odemis-stop when it fails to …

tepals:
…stop the odemis backend.

### DIFF
--- a/src/odemis/util/test.py
+++ b/src/odemis/util/test.py
@@ -136,7 +136,16 @@ def stop_backend():
     model._core._microscope = None  # force reset of the microscope for next connection
 
     if status != driver.BACKEND_STOPPED:
-        raise IOError("Backend failed to stop, now %s" % status)
+        logging.warning("Backend failed to stop, will run 'sudo odemis-stop' to kill it.")
+        # TODO this is a temporary fix for stopping the back-end. When a better solution is found, remove this.
+        try:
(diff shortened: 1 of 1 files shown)


src/odemis/util/test.py
```diff
@@ -136,7 +136,15 @@ def stop_backend():
     model._core._microscope = None  # force reset of the microscope for next connection
 
     if status != driver.BACKEND_STOPPED:
-        raise IOError("Backend failed to stop, now %s" % status)
+        logging.warning("Backend failed to stop, will run odemis-stop to kill it.")
```
Kleijwegt, line 139: I think it is useful to mention the fact that you use sudo for this call.
logging.warning("Backend failed to stop, will run 'sudo odemis-stop' to kill it.")

src/odemis/util/test.py
```diff
@@ -136,7 +136,15 @@ def stop_backend():
...
+        try:
+            ret = subprocess.run(["sudo", "odemis-stop"], timeout=60)
```
Kleijwegt, line 141: I think there should be a TODO that this is a temporary solution for stopping the back-end but that a better solution should be found one day.

Kleijwegt:
@pieleric Thera and I discussed that we can merge this PR so it is included in the "weekly" run of the unit tests of this afternoon.

PR #2146, opened 2022-06-09
driver xt_client: prepareForScan was only added to the FIBScanner and…

tepals:
… not to the normal Scanner.

This caused the acquisition code in the detector to fail. This commit adds the prepareForScan to the Scanner class, and moves beam blanking to that function, because the FIBScanner does not support beam blanking.

There were also a few test cases which overwrote the blanker VA with a boolean value, which is also fixed.

### DIFF
--- a/src/odemis/driver/test/xt_client_test.py
+++ b/src/odemis/driver/test/xt_client_test.py
@@ -35,7 +35,9 @@ from odemis.driver import xt_client
 from odemis.model import ProgressiveFuture, NotSettableError
 from odemis.util import test
 
-logging.basicConfig(level=logging.INFO)
+logging.basicConfig(level=logging.DEBUG,
+                    format="%(asctime)s  %(levelname)-7s %(module)s:%(lineno)d %(message)s",
+                    force=True)
 
 # Accept three values for TEST_NOHW
 # * TEST_NOHW = 1: not connected to anything => skip most of the tests
@@ -352,9 +354,14 @@ class TestMicroscope(unittest.TestCase):
         im = self.detector.data.get()
         duration = time.time() - start
         self.assertEqual(im.shape, self.scanner.resolution.value[::-1])
-        self.assertGreaterEqual(duration, expected_duration,
-                                "Error execution took %f s, less than exposure time %d." % (
-                                    duration, expected_duration))
+
+        if TEST_NOHW != 'sim':
+            # The basic simulator does not simulate the right execution time.
+            self.assertGreaterEqual(
+                duration,
+                expected_duration,
+                "Error execution took %f s, less than exposure time %d." % (duration, expected_duration)
+            )
         self.assertIn(model.MD_DWELL_TIME, im.metadata)
         # Set back dwell time to initial value
         self.scanner.dwellTime.value = init_dwell_time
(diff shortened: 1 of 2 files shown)


src/odemis/driver/xt_client.py
```diff
@@ -1524,15 +1539,25 @@ def __init__(self, name, role, parent, channel="ion2", **kwargs):
...
     def prepareForScan(self):
         """
-        Make sure the scan mode is in "full_frame" and not in "external"
+        Make sure the scan mode is in "full_frame" and not in "external" before starting to scan.
         """
+        # Currently the FIB blanker is not supported via the XT interface,
+        # so we cannot blank the beam at the start and end of the scan.
+
+        # Note: ideally this would be based on an external VA, which when put to
+        # 'auto' would switch to full frame at the start of a scan.
```
Kleijwegt, line 1548: I think the FIB was actually autblanked via XT, otherwise we couldn't scan with it. It would be unblanked when swithcing to FLM.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -35,7 +35,9 @@
 from odemis.model import ProgressiveFuture, NotSettableError
 from odemis.util import test
 
-logging.basicConfig(level=logging.INFO)
+logging.basicConfig(level=logging.DEBUG,
```
Kleijwegt, line 38: Did you check if the test file scanner_test.py with the composited scanner/detector still passes? It might influence that test case

src/odemis/driver/xt_client.py
```diff
@@ -1621,14 +1644,10 @@ def _acquire(self):
...
+                    # TODO When switching e-beam <--> FIB, handle calling finishScan on the old scanner and
+                    #  prepareForScan on the new scanner.
+                    self._scanner.prepareForScan()
                     self.parent.set_channel_state(self._scanner.channel, True)
-                    if hasattr(self._scanner, "blanker"):
-                        # TODO: When the acquisition ends, if blanker is set to automatic (None), we need to activate
-                        #  the blanker again. That should also happen when switching e-beam <--> FIB, but currently
-                        #  the FIB blanker is not supported via the XT interface.
-                        if self._scanner.blanker.value is None:
-                            self.parent.unblank_beam()
```
tepals, line 1631: The beam blanking is now done before the channel state is set to True. I tested with the simulator VM and this properly unblanks the beam.

src/odemis/driver/test/xt_client_test.py
tepals: just checked, and that still passes

src/odemis/driver/xt_client.py
tepals: I think your comment should be for line 1544 and 1545? I based that comment on what was there previously:
```
# TODO: When the acquisition ends, if blanker is set to automatic (None), we need to activate
#  the blanker again. That should also happen when switching e-beam <--> FIB, but currently
#  the FIB blanker is not supported via the XT interface.
```
Reading that comment I concluded that FIB blanking is not supported at the moment.

I also think XT does some autoblanking for the FIB. If you are sure you know when it blanks and unblanks I can add that to the comment. Otherwise I can update the comment to:

```
# Currently the XT interface autoblanks the FIB, control of the FIB blanker is not supported
# so we do not blank the beam at the start and end of the scan.
```

pieleric (approved):

PR #2147, opened 2022-06-10
Supress the warning that numpy.log10 divides by zero in _projection.py

Kleijwegt:


src/odemis/acq/stream/_projection.py
```diff
@@ -679,7 +679,9 @@ def _project2RGBPolar(self, ebeam_pos, pol_pos, cache_raw):
                 if pol_pos in [MD_POL_EPHI, MD_POL_ETHETA, MD_POL_EX, MD_POL_EY, MD_POL_EZ]:
                     data = numpy.abs(cache_raw[pol_pos])
                     try:
-                        plotorder = int(numpy.log10(data.max()))
+                        # numpy.log10 will give a warning for the case that plotorder = 0, supress this warning
```
pieleric, line 682: Technically, that's when `data.max() == 0`. That's a single value. Then log10() makes it a -inf, and int() raises an OverflowError. Instead of suppressing the warning + catching the error, I'd suggest to write:
```python
max_val = data.max()
if max_val > 0:
    plotorder = int(math.log10(max_val))
else:
    # max_val == 0, because bg image and image are the same
    # Let's not make too much fuss about it.
    plotorder = 0
```

pieleric (changes requested):

src/odemis/acq/stream/_projection.py
Kleijwegt: I agree this does make more sense. I didn't want to change to much code but that wasn't the best choice in this case..

pieleric (approved):

BasHolweg (approved):

PR #2148, opened 2022-06-10
If two viewports are visible, show these two viewports above each oth…

Kleijwegt:
…er, also test this correctly.

### DIFF
--- a/src/odemis/gui/comp/grid.py
+++ b/src/odemis/gui/comp/grid.py
@@ -111,7 +111,7 @@ class ViewportGrid(wx.Panel):
             except AttributeError:
                 # Should never happen, unless it's not really a ViewPort.
                 # If so, let's not go completely bad
-                logging.exception("View has no view")
+                logging.exception("View %s has no view", v)
                 viewports.append(v)
             if len(viewports) >= 4:
                 break
@@ -124,8 +124,6 @@ class ViewportGrid(wx.Panel):
         return ([ViewPort])
         """
         viewports = []
-        if [view.Shown for view in self.viewports].count(True) > 2:
-            logging.warning("Found more than two viewports to show, only the first 2 are displayed in this layout.")
 
         if len(self.visible_viewports) == 2:
             logging.debug("Only two viewports are visible so these are shown.")
@@ -140,9 +138,10 @@ class ViewportGrid(wx.Panel):
                 except AttributeError:
                     # Should never happen, unless it's not really a ViewPort.
                     # If so, let's not go completely bad
-                    logging.exception("Viewport has no view")
+                    logging.exception("Viewport %s has no view", v)
                     viewports.append(v)
                 if len(viewports) >= 2:
+                    logging.warning("Found more than two viewports to show, only the first 2 are displayed in this layout.")
                     break
 
             return viewports
--- a/src/odemis/gui/comp/grid.py
+++ b/src/odemis/gui/comp/grid.py
@@ -127,19 +127,25 @@ class ViewportGrid(wx.Panel):
         if [view.Shown for view in self.viewports].count(True) > 2:
             logging.warning("Found more than two viewports to show, only the first 2 are displayed in this layout.")
 
-        for v in self.viewports:
-            try:
-                if v.view is not None and v.Shown:
-                    viewports.append(v)
-            except AttributeError:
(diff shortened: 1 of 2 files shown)


src/odemis/gui/comp/grid.py
```diff
@@ -127,19 +127,25 @@ def get_2_viewports(self):
...
+
+        else:
+            # If more/less than 2 viewports are visible use the first 2 viewports with a view
+            for v in self.viewports:
+                try:
+                    if v.view is not None and v.Shown:
+                        viewports.append(v)
+                except AttributeError:
+                    # Should never happen, unless it's not really a ViewPort.
+                    # If so, let's not go completely bad
```
BasHolweg, line 142: I'm unclear on what going completely bad means in this context

src/odemis/gui/comp/grid.py
```diff
@@ -127,19 +127,25 @@ def get_2_viewports(self):
         if [view.Shown for view in self.viewports].count(True) > 2:
             logging.warning("Found more than two viewports to show, only the first 2 are displayed in this layout.")
 
-        for v in self.viewports:
-            try:
-                if v.view is not None and v.Shown:
+        if len(self.visible_viewports) == 2:
```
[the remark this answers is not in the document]
Kleijwegt: @pieleric Somehow it is possible to to have 3 viewports which are shown but only 2 visbleviewports. So a viewport can be invissible and shown. I lost track of the relation between all these parts here, can you maybe explain how that is possible?

src/odemis/gui/comp/grid.py
```diff
@@ -127,19 +127,25 @@ def get_2_viewports(self):
...
+        else:
+            # If more/less than 2 viewports are visible use the first 2 viewports with a view
+            for v in self.viewports:
+                try:
+                    if v.view is not None and v.Shown:
+                        viewports.append(v)
+                except AttributeError:
+                    # Should never happen, unless it's not really a ViewPort.
+                    # If so, let's not go completely bad
+                    logging.exception("Viewport has no view")
```
pieleric, line 143: Can you add a `"%s", viewport` as info to the log message so that it's a little bit clearer what happened in such case.
Ideally, could you also update the log messages in the other cases. (for instance line 114, which is even more confusing for now)

src/odemis/gui/comp/grid.py
pieleric, line 130: I think there is something wrong with this code. The `show_2_vert_stacked_viewports()` uses `get_2_viewports()` to *update* `.visible_viewports`. So you cannot just return `.visible_viewports`... that would be too easy.

src/odemis/gui/comp/grid.py
```diff
@@ -127,19 +127,25 @@ def get_2_viewports(self):
         if [view.Shown for view in self.viewports].count(True) > 2:
             logging.warning("Found more than two viewports to show, only the first 2 are displayed in this layout.")
```
pieleric, line 128: You could move this warning to line 146.

src/odemis/gui/comp/grid.py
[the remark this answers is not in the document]
pieleric: Honestly, I didn't write this class, and I have to say it's a little confusing. It's not clear what is the interface (method/attributes) and how one should use it. I looked a little bit and to me it seems that all one should do is to call `set_visible_viewports()`, and `.show_*_viewports()`. In that case, the expectation is just that there are at least as many viewports as requested (2 or 4) with a view. Set visible viewports will select which viewport to show and reorganise the order of the viewports, and the other two methods make sure that just the first 2/4 are shown.

I'm quite confused by `show_viewport()` and `hide_viewport()`. I guess it tries to adjust the layout based on the new number of visible viewports. But 3 viewports is just unexpected.

pieleric (changes requested):

src/odemis/gui/comp/grid.py
Kleijwegt: I actually think we can, but only **if** the number of visible viewports is 2, in another cause we need to do different stuff. That is the **else**
I have updated it according to you other comment which does simplify the reading the code.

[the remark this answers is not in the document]
Kleijwegt: @pieleric I think we can close this issue isn't it?

PR #2153, opened 2022-06-23
config: fix secom-pretty sim to match the images again

pieleric:
Some time ago the CCD simulator must have adjusted the metadata, and
this caused the images to not overlay properly anymore.

It's now easier, we just put the original lens magnification.


Kleijwegt (approved):
I don't have any detailed knowledge about this simulator.
For the rest these changes look fine.

PR #2154, opened 2022-06-27
util test: improve error message in assert_pos_almost_equal()

pieleric:
It's convenient to quickly know which axis is not equal.


Kleijwegt (approved):
I would like to have the word axis included. But any way I think this is a good commit.

PR #2155, opened 2022-06-27
driver smaract: update SmarPod

pieleric:


src/odemis/driver/smaract.py
```diff
@@ -291,119 +295,203 @@ def __str__(self):
...
-            on initialization
-        actuator_speed: (double) the default speed (in m/s) of the actuators
+                usb:sn:<serialnumber>
+            where <serialnumber> is the serial number of an MCS2 controller.
+            Be aware that on Linux, a udev rules must be added to disable the standard
+            tty driver from connecting to the device.
+            If the controller has a TCP/IP connection, use one of:
+                network:<ipv4>
+                network:sn:<serialnumber>
+        hwmodel (int): the hardware model code (typically between 10000 and 10100)
```
tepals, line 319: For the other arguments in this function the docstring format is `name: (type)` instead of `name (type):`

src/odemis/driver/smaract.py
```diff
@@ -291,119 +295,203 @@ def __str__(self):
...
             
         # Not to be mistaken with axes which is a simple public view
         self._axis_map = {}  # axis name -> axis number used by controller
         axes_def = {}  # axis name -> Axis object
-        self._locator = c_char_p(locator.encode("ascii"))
-        self._options = c_char_p("".encode("ascii"))  # In the current version, this must be an empty string.
 
-        for axis_name, axis_par in axes.items():
+        for axis_name, param in axes.items():
+            logging.debug("a %s = %s", axis_name, param)
```
tepals, line 350: why write `a` and not `axis`?

src/odemis/driver/smaract.py
```diff
@@ -291,119 +295,203 @@ def __str__(self):
...
+
         logging.debug("Successfully connected to SmarPod Controller ID %d", self._id.value)
-        self.core.Smarpod_SetSensorMode(self._id, SmarPodDLL.SMARPOD_SENSORS_ENABLED)
+        self.core.Smarpod_SetSensorMode(self._id, SmarPodDLL.SENSORS_ENABLED)
 
         model.Actuator.__init__(self, name, role, axes=axes_def, **kwargs)
 
         # Add metadata
-        self._swVersion = self.GetSwVersion()
+        # TODO: GetDLLVersion
```
tepals, line 383: What does this TODO mean? Or what needs to be done?

src/odemis/driver/smaract.py
```diff
@@ -291,119 +295,203 @@ def __str__(self):
...
         self._executor = CancellableThreadPoolExecutor(1)  # one task at a time
 
-        referenced = c_int()
-        self.core.Smarpod_IsReferenced(self._id, byref(referenced))
+        # Use a default actuator speed
+        self.SetSpeed(speed)
+        self.speed = VigilantAttribute({}, unit="m/s", readonly=True)
+        self._updateSpeed()
+        self.SetAcceleration(accel)
+        self._accel = self.GetAcceleration()
```
tepals, line 403: why isn't this a VA?

src/odemis/driver/smaract.py
```diff
@@ -291,119 +295,203 @@ def __str__(self):
...
+
         # should be safe to close the device multiple times if terminate is called more than once.
-        self.core.Smarpod_Close(self._id)
         super(SmarPod, self).terminate()
+
+    def updateMetadata(self, md):
+        if model.MD_PIVOT_POS in md:
+            pivot = md[model.MD_PIVOT_POS]
+            if not (isinstance(pivot, dict) and set(pivot.keys()) == {"x", "y", "z"}):
+                raise ValueError("Invalid metadata, should be a coordinate dictionary but got %s." % (pivot,))
```
tepals, line 438: I would specify this even further: `"Invalid metadata, should be a coordinate dictionary containing 'x', 'y' and 'z' but got %s." % (pivot,)"`

src/odemis/driver/smaract.py
```diff
@@ -415,46 +503,47 @@ def GetMoveStatus(self):
...
         status = c_uint()
         self.core.Smarpod_GetMoveStatus(self._id, byref(status))
-        return status
+        return status.value
 
     def Move(self, pos, hold_time=0, block=False):
         """
-        Move to pose command.
+        Move to pose command. It is possible and safe to call it if a previous
+           movement that has been started, called in non-blocking mode.
```
tepals, line 520: This sentence is a bit confusing, should it be:
`It is possible and safe to call it if a previous movement has already been started. Called in non-blocking mode.`
or:
`It is possible and safe to call it if a previous movement has been called in non-blocking mode.`?

src/odemis/driver/smaract.py
```diff
@@ -415,46 +503,47 @@ def GetMoveStatus(self):
...
-        hold_time: (float) specify in seconds how long to hold after the move.
+        hold_time: (float >=0) specify in seconds how long to hold after the move.
             If set to float(inf), will hold forever until a stop command is issued.
         block: (bool) Set to True if the function should block until the move completes
 
         Raises: SmarPodError if a problem occurs
         """
-        # convert into a smartpad pose
-        newPose = dict_to_pose(pos, self.GetPose())
+        # convert into a SmarpodPose
```
tepals, line 529: Is it useful to add here:
```
if hold_time < 0:
    raise ValueError(f"hold_time should be < 0, is {hold_time}")
```

src/odemis/driver/smaract.py
```diff
@@ -465,41 +554,58 @@ def Stop(self):
...
+        else:
+            # the second argument (1) turns on speed control.
+            self.core.Smarpod_SetSpeed(self._id, c_int(1), c_double(value))
 
-    def _get_speed(self):
+    def GetSpeed(self) -> Optional[float]:
         """
-        Returns (double) the speed of the SmarPod motion
+        Returns (float or None) the speed of the SmarPod motion.
+          If speed control is disable, it returns None.
```
tepals, line 576: typoe: disable -> disabled

src/odemis/driver/smaract.py
```diff
@@ -465,41 +554,58 @@ def Stop(self):
...
+            return speed.value
+        else:
+            return None
 
-    def SetAcceleration(self, value):
+    def SetAcceleration(self, value: Optional[float]):
         """
         Set the acceleration of the SmarPod motion
-        value: (double) indicating acceleration for all axes
+        value: (float or None) indicating acceleration for all axes
```
tepals, line 589: can you add what happens if you pass `None`?

src/odemis/driver/smaract.py
```diff
@@ -465,41 +554,58 @@ def Stop(self):
...
+            self.core.Smarpod_SetAcceleration(self._id, c_int(0), c_double(0))
+        else:
+            # Passing 1 enables acceleration control.
+            self.core.Smarpod_SetAcceleration(self._id, c_int(1), c_double(value))
 
-    def GetAcceleration(self):
+    def GetAcceleration(self) -> Optional[float]:
         """
-        Returns (double) the acceleration of the SmarPod motion
+        Returns (float or None) the acceleration of the SmarPod motion
```
tepals, line 600: can you add what return value `None` means?

src/odemis/driver/test/smaract_test.py
```diff
@@ -44,10 +44,10 @@
     }
 
 CONFIG_SMARTPOD = {"name": "SmarPod",
-        "role": "",
-        "ref_on_init": True,
-        "actuator_speed": 0.1,  # m/s
-        "locator": "usb:ix:0",
+        "role": "test",
+        # "locator": "usb:ix:0",
```
tepals, line 48: why is this commented?

src/odemis/driver/test/smaract_test.py
```diff
@@ -91,8 +95,12 @@ def setUpClass(cls):
 
     @classmethod
     def tearDownClass(cls):
+
```
tepals, line 98: why the extra blank line?

src/odemis/driver/test/smaract_test.py
```diff
@@ -167,25 +176,52 @@ def test_move_cancel(self):
         self.assertNotEqual(round(difference, 4), 0)
 
     def test_move_rel(self):
+        self.dev.reference(self.dev.axes.keys()).result()
+        # return
```
tepals, line 180: why is this comment here?

install/linux/usr/share/odemis/hwtest/smaract-smarpod.yaml
```diff
@@ -0,0 +1,55 @@
...
+    class: Microscope,
+    role: brightfield,
+    children: [Stage],
+}
+
+"Stage": {
+    class: smaract.SmarPod,
+    role: stage,
+    init: {
+#        locator: "fake",
```
tepals, line 11: why is this commented?

tepals (changes requested):

src/odemis/driver/smaract.py
```diff
@@ -58,7 +58,7 @@ def add_coord(pos1, pos2):
     return ret
 
 
-class Pose(Structure):
+class Smarpod_Pose(Structure):
```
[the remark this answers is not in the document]
pieleric: Indeed, the goal was to just follow exactly the same name as in the C file:
``` c
typedef struct
{
    double positionX;
    double positionY;
    double positionZ;
    double rotationX;
    double rotationY;
    double rotationZ;
}Smarpod_Pose;
```

src/odemis/driver/smaract.py
pieleric: I've updated the doctring to either not have the type anymore (for the types which are already fully defined in the type annotation), or formatted as `name (type):`.

src/odemis/driver/smaract.py
pieleric: Already did it... but forgot to remove the TODO.

src/odemis/driver/smaract.py
pieleric: Mostly because we don't have a formal VA in actuators for the acceleration. It wouldn't be hard to think of it, as it'd be the same as the `.speed`. That said here, the accel (like the speed) is so weirdly defined that I don't think we should care too much.

tepals (approved):

PR #2157, opened 2022-06-30
gui: fix minor errors with ENZEL

pieleric:


src/odemis/gui/model/__init__.py
```diff
@@ -338,8 +338,10 @@ def __init__(self, microscope):
             required_roles = []
             if self.role in ("secom", "delphi", "enzel"):
                 required_roles += ["e-beam", "light", "stage", "focus"]
-                if self.role == "secom":
+                if self.role in ("secom", "enzel"):
```
Kleijwegt, line 341: This mean we can no longer use an enzel outside of the microscope isn't it?
On second thought I think the application/installation engineers still do this quite often. Maybe there should be some option to still use it (like role:enzel_door_only)

install/linux/usr/share/odemis/sim/enzel-sim-xt.odm.yaml
```diff
@@ -6,7 +6,7 @@ ENZEL-Sim: {
         "Sample Thermostat", "Cooler Controller",
         "SEM E-beam Full", "SEM Detector Full",
         "SEM Detector", "EBeam Control",
-        "EBeam Focus", "EBeam Detector",
+        "EBeam Focus", "FIB ETD",
```
Kleijwegt, line 9: ETD? What do that stand for?

Kleijwegt (changes requested):
I think we should check this with Marre/Daan before we merge this:
https://github.com/delmic/odemis/pull/2157/files#r911387683

src/odemis/gui/model/__init__.py
pieleric: No. This code doesn't really change the behaviour of the GUI, it just makes it "nicer" to the user. Currently, if starting a microscope file with role "enzel" missing some component, the GUI will just fail to start with some weird error like "object of NoneType has no attribute beamShift". Now, it will also fail, but saying "Microscope is missing ion-beam". 

There are two ways to start a system with less hardware:
* simulate some of the components (that's often the most useful way)
* use a different microscope type (eg, optical)

install/linux/usr/share/odemis/sim/enzel-sim-xt.odm.yaml
pieleric: That's another name for "secondary electron detector". It stands for Everhart–Thornley detector. Maybe I could rename it to "FIB Detector" or "FIB SE Detector". Any preference or suggestion?

src/odemis/gui/model/__init__.py
Kleijwegt: Ok sounds great

Kleijwegt (approved):

PR #2158, opened 2022-07-05
acq/gui: Fix position display for ENZEL/METEOR

pieleric:
Do not change the position (mode) while the stage is moving. In
particular:
* The alignment and localization tab don't get temporarily enabled
  half-way through a movement
* The position switch buttons don't get enabled during movement

Also change the METEOR buttons to turn orange during the move, and green
at the end of the move.

Also, on ENZEL, always report THREE_BEAMS instead of IMAGING.

Also, in case of cancellation, don't log an error.

Also clean-ups:
* ENZEL cryo tilt is not needed anymore
* ENZEL IMAGING is not needed anymore
* Don't use internal _state of Future, but just .done() or .running()

In general the change shouldn't have any effect on movement, but only on
the display of the current position.

### DIFF
--- a/src/odemis/gui/cont/tabs.py
+++ b/src/odemis/gui/cont/tabs.py
@@ -51,7 +51,7 @@ from odemis.gui.comp.popup import show_message
 from odemis.gui.cont.features import CryoFeatureController
 from odemis.gui.util.wx_adapter import fix_static_text_clipping
 from odemis.gui.win.acquisition import ShowChamberFileDialog
-from odemis.model import getVAs, VAEnumerated
+from odemis.model import getVAs, VAEnumerated, InstantaneousFuture
 from odemis.util.filename import guess_pattern, create_projectname, create_filename
 
 import odemis.acq.stream as acqstream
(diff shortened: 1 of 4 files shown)


src/odemis/gui/cont/tabs.py
```diff
@@ -476,13 +472,17 @@ def __init__(self, name, button, panel, main_frame, main_data):
             sem_stream_cont = self._streambar_controller.addStream(sem_stream, add_to_view=True)
             sem_stream_cont.stream_panel.show_remove_btn(False)
 
+        # Only enable the tab when the stage is at the right position
         if self.main_data.role == "enzel":
             self.stage = self.tab_data_model.main.stage
+            self._allowed_targets = [THREE_BEAMS, SEM_IMAGING]
```
Kleijwegt, line 478: Don't you think we should also allow this when we are in ALIGNMENT?
It is a really small chance that someone wants to do the factory setting and go in the localization tab but it is not dangerous or anything.

pieleric: Indeed. I think that in the "ideal" version the ALIGNMENT position shouldn't be a reported position, but just a "specific position" within the THREE_BEAMS posture. However, for now that's not how it works. I'll add ALIGNMENT.

Kleijwegt:
> Normally, from any _known_ position you can now reach any other position. However, if you stop the stage halfway during a move, there is a high chance that the position is considered "unknown". In this special case, you can only go to loading, and I would say this is the right behaviour. There is just an extra twist: it's able to detect that the stage is in between loading and three-beams (we added this mostly because that's the most common path). In this case, it's allowed to go both to Loading and Three-beams. So your observations seem consistent with the expected behaviour.

I have checked with Daan Boltje and he also can find himself in this answer so lets keep it this way.

Kleijwegt (approved):

PR #2159, opened 2022-07-06
Fixes for enzel pre UQ test

Kleijwegt:


src/odemis/acq/move.py
```diff
@@ -643,6 +643,18 @@ def _doCryoSwitchSamplePosition(future, target):
                     if future._task_state == CANCELLED:
                         logging.info("Cancelling aligner movement...")
                         raise CancelledError()
+
+                # If the stage rotates a lot, move it first to a safe position (Moving it in X and RZ is sufficient)
+                if abs(current_pos["rx"] - target_pos[target]["rx"]) > numpy.deg2rad(2):
```
pieleric, line 648: Please use the standard `math.radians()` function instead of the numpy one (we typically use numpy functions only when it needs to handle numpy arrays).

src/odemis/gui/cont/tabs.py
```diff
@@ -2939,33 +2939,18 @@ def _milling_angle_changed(self, pos):
         Updates the milling control with the changed angle position.
         :param pos: (float) value of rx
        """
-        rx_value = math.degrees(pos)
-        self.panel.ctrl_rx.Value = readable_str(rx_value, unit="°", sig=3)
+        current_angle = math.degrees(pos)
```
pieleric, line 2942: Let's add a comment about the us of `HasFocus()`. Something like: "When the user is typing a new value in, it's not handy that the text changes, so don't update in such case".

src/odemis/acq/move.py
```diff
@@ -643,6 +643,18 @@ def _doCryoSwitchSamplePosition(future, target):
...
+                    sub_move_dict = filter_dict({"rz"}, target_pos[LOADING])
+                    logging.debug("Moving %s to a safe rotation position in RZ axis, to %s.", stage.name, sub_move_dict)
+                    run_sub_move(future, stage, sub_move_dict)
+
+                    sub_move_dict = filter_dict({"x"}, target_pos[LOADING])
+                    logging.debug("Moving %s to a safe position in X axis, to %s.", stage.name, sub_move_dict)
+                    run_sub_move(future, stage, sub_move_dict)
+
+
```
pieleric, line 657: One empty line is enough.

pieleric (changes requested):

pieleric (approved):

PR #2156, opened 2022-06-30
Acquisition stitching: Cleaned up weaver code and test cases.

tepals:
Split up getFullImage() into three functions: get_bounding_boxes, weave_tiles and update_md.
get_bounding_boxes and update_md are the same for each different Weaver class, therefore I moved them to the base class.
Each class now has a weave_tiles function, which only contains the code that is actually specific for that class.
This clean up will make it easier to update the code and add support for rotated tiles.

PR #2160, opened 2022-07-08
Rotated tiles overview image

tepals:
This PR is based on https://github.com/delmic/odemis/pull/2156 , so that PR should be reviewed first

### DIFF
--- a/src/odemis/acq/stitching/test/weaver_test.py
+++ b/src/odemis/acq/stitching/test/weaver_test.py
@@ -21,6 +21,7 @@ Odemis. If not, see http://www.gnu.org/licenses/.
 """
 
 import logging
+import math
 import os
 import random
 import time
@@ -33,7 +34,7 @@ from odemis import model
 from odemis.acq.stitching import CollageWeaver, MeanWeaver, CollageWeaverReverse, WEAVER_MEAN, WEAVER_COLLAGE_REVERSE, \
     WEAVER_COLLAGE
 from odemis.dataio import find_fittest_converter
-from odemis.util.img import ensure2DImage
+from odemis.util.img import ensure2DImage, rotate_img_metadata
 from stitching_test import decompose_image
 
 logging.getLogger().setLevel(logging.DEBUG)
@@ -66,6 +67,7 @@ class WeaverBaseTest:
             model.MD_EXP_TIME: 1.2,  # s
             model.MD_IN_WL: (500e-9, 520e-9),  # m
             model.MD_DIMS: "YX",
+            model.MD_ROTATION: 0,
         }
         intile = model.DataArray(img12, md)
 
@@ -92,6 +94,7 @@ class WeaverBaseTest:
             model.MD_POS: (10e-3, 30e-3),  # m
             model.MD_DWELL_TIME: 1.2e-6,  # s
             model.MD_DIMS: "YX",
+            model.MD_ROTATION: 0,
         }
         intile = model.DataArray(img8, md8)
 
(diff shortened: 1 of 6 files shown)


pieleric (changes requested):
I'd suggest to also extend the meteor-sim.odm.yaml with the ROTATION metadata on the CCD. Probably the nicest would be to also set ~5°, but you'd need to check that it doesn't mess up with some test cases (probably not). Otherwise, you can just set it to 0°.

Also, please add the changes needed for the overview acquisition.

src/odemis/acq/stitching/test/tiledacq_test.py
```diff
@@ -415,6 +416,28 @@ def test_cancel(self):
         self.assertTrue(self.done)
         self.assertTrue(f.cancelled())
 
+    def test_rotated_tiles(self):
+        """Test that all tiles contain the correct rotation, when a rotation is specified on the ccd metadata."""
```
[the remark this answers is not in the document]
tepals: I've removed the test

src/odemis/acq/stitching/test/weaver_test.py
```diff
@@ -189,6 +192,41 @@ def test_synthetic_perfect_overlap(self):
...
+                        weaver_rot = CollageWeaverReverse()
+                    elif self.weaver_type == WEAVER_MEAN:
+                        weaver = MeanWeaver()
+                        weaver_rot = MeanWeaver()
+
+                    for t in tiles:
+                        t.metadata[model.MD_ROTATION] = rotation
+                        weaver.addTile(t)
+                        t2 = rotate_image(t, rotation, (0, 0))
+                        weaver_rot.addTile(t2)
```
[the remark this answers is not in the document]
tepals: I don't fully understand it anymore either so I've removed it.

src/odemis/util/test/img_test.py
```diff
@@ -21,22 +21,23 @@
...
+import time
+import unittest
+from builtins import range
+
 import numpy
+
 from odemis import model
 from odemis.dataio import tiff
-from odemis.util import img, get_best_dtype_for_acc
+from odemis.util import img, get_best_dtype_for_acc, test
```
tepals, line 35: this should be testing now

Kleijwegt (approved):

src/odemis/util/test/img_test.py
```diff
@@ -1095,8 +1096,42 @@ def getSubData(dast, zoom, rect):
         
         os.remove(FILENAME)
 
+
+class TestRotateImage(unittest.TestCase):
+
+    def test_rotate_image(self):
```
pieleric, line 1102: Let's update the name too: "test_rotate_img_metadata".

src/odemis/util/test/img_test.py
```diff
@@ -1095,8 +1096,42 @@ def getSubData(dast, zoom, rect):
...
+            # pos:  {center of rot, rotation,  expected position}
+            {"pos": (0, 0), "cor": (0, 0), "rot": 10, "exp_pos": (0, 0)},
+            {"pos": (0, 1), "cor": (0, 0), "rot": 90, "exp_pos": (-1, 0)},
+            {"pos": (0, 1), "cor": (0, 0), "rot": -90, "exp_pos": (1, 0)},
+            {"pos": (123e-6, 24e-7), "cor": (123.0e-6, 2.4e-6), "rot": 5.7, "exp_pos": (123.0e-6, 2.4e-6)},
+            {"pos": (123e-6, 24e-7), "cor": (10.0e-6, 24e-6), "rot": 5.7, "exp_pos": (1.24586586e-4, 1.37299314e-05)},
+            {"pos": (123e-6, 24e-7), "cor": (10.0e-6, 24e-6), "rot": -5.7, "exp_pos": (1.20295973e-4, -8.7163320e-06)},
+        ]
+
+        data = numpy.random.rand(51, 76)
```
pieleric, line 1121: The data content doesn't matter, right? So let's just just `numpy.ones(51, 76)`, because "random" make it sound like we will have different behaviour depending on the random seed, which is not set...

src/odemis/acq/stitching/test/weaver_test.py
```diff
@@ -189,6 +192,33 @@ def test_synthetic_perfect_overlap(self):
...
+            weaver_class = CollageWeaverReverse
+        elif self.weaver_type == WEAVER_MEAN:
+            weaver_class = MeanWeaver
+
+        for img in IMGS:
+            conv = find_fittest_converter(img)
+            img = conv.read_data(img)[0]
+            img = ensure2DImage(img)
+            for n in numTiles:
+                [tiles, _] = decompose_image(img, overlap, n, "horizontalZigzag", False)
```
pieleric, line 212: Could you indicate somewhere that the content of the image will be bad looking, because of the rotation is not taken into account when decomposing the image (and that doesn't matter because we don't check for the image content).

pieleric (changes requested):

pieleric (approved):
Looks good to me. Could you squash/rebase against master to fix the conflicts?

PR #2174, opened 2022-07-19
gui model: allow SPARCv1 to not have lens

pieleric:
We had some odd SPARCv1 without lens in some configuration... let's not
break that.

### DIFF
--- a/src/odemis/gui/model/__init__.py
+++ b/src/odemis/gui/model/__init__.py
@@ -347,7 +347,10 @@ class MainGUIData(object):
             elif self.role == "meteor":
                 required_roles += ["light", "stage", "focus"]
             elif self.role in ("sparc", "sparc2"):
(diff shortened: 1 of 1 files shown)


src/odemis/gui/model/__init__.py
```diff
@@ -347,7 +347,9 @@ def __init__(self, microscope):
             elif self.role == "meteor":
                 required_roles += ["light", "stage", "focus"]
             elif self.role in ("sparc", "sparc2"):
-                required_roles += ["e-beam", "mirror", "lens"]
+                required_roles += ["e-beam", "mirror"]
```
Kleijwegt, line 350: I would add this comment since this is a semi odd configuration. The rest I approve:
`required_roles += ["e-beam", "mirror"]  # SPARCv1 can also work without a lens.`

PR #2184, opened 2022-07-22
gui comp stream: better handle long stream names

pieleric:
Make more space for the name (by removing extra margin on the right).
If the name still doesn't fit, show an ellipsis, to indicate it's
trunctated.
Always show the full name by hovering over the name.

### DIFF
--- a/src/odemis/gui/comp/stream.py
+++ b/src/odemis/gui/comp/stream.py
@@ -133,7 +133,7 @@ class StreamPanelHeader(wx.Control):
         self.btn_update = self._add_update_btn() if self.Parent.options & OPT_BTN_UPDATE else None
 
         # Add spacer for creating padding on the right side of the header panel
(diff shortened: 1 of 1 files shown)


src/odemis/gui/comp/stream.py
```diff
@@ -165,7 +165,10 @@ def _add_suggest_ctrl(self):
 
     def _add_label_ctrl(self):
         """ Add a label control to the header panel """
-        label_ctrl = wx.StaticText(self, -1, self.Parent.stream.name.value)
+        label_ctrl = wx.StaticText(self, wx.ID_ANY, self.Parent.stream.name.value,
+                                   style=wx.ST_ELLIPSIZE_END)
+        # In case the name is too long, as least we can see it full with a mouse hover
```
tepals, line 170: typo: as -> at

tepals (changes requested):

PR #2187, opened 2022-08-03
add check on importing required packages in technolution_test

FCanberk:
Added a check if technolution packages are available, if not skip the test.

### DIFF
--- a/src/odemis/driver/test/technolution_test.py
+++ b/src/odemis/driver/test/technolution_test.py
@@ -43,11 +43,17 @@ import matplotlib.pyplot as plt
 import numpy
 
 from odemis import model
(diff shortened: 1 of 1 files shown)


src/odemis/driver/test/technolution_test.py
```diff
@@ -103,6 +109,10 @@ class TestAcquisitionServer(unittest.TestCase):
 
     @classmethod
     def setUpClass(cls):
+        if not technolution_available:
+            raise unittest.SkipTest(f"Skipping the technolution tests, correct libraries to perform the tests are not"
+                                f" available.")
```
tepals, line 114: can you also add this to the other test classes? This way all test cases are skipped if technolution_asm is not available

tepals (changes requested):

tepals (approved):

PR #2189, opened 2022-08-05
add missing tests

FCanberk:
Added test cases for the functions that does not have tests yet in util.__init__.py

### DIFF
--- a/src/odemis/util/test/util_test.py
+++ b/src/odemis/util/test/util_test.py
@@ -21,15 +21,17 @@ Odemis. If not, see http://www.gnu.org/licenses/.
 """
 from __future__ import division, print_function
 
+import inspect
 from functools import partial
 import os
+import numpy.random
 import odemis
 from odemis import model, util
 import logging
 import math
 from odemis.model import CancellableFuture
 from odemis.util import limit_invocation, TimeoutError, executeAsyncTask, \
-    perpendicular_distance, to_str_escape, test, driver, timeout
+    perpendicular_distance, to_str_escape, testing, driver, timeout
 import time
 import unittest
 import weakref
@@ -61,7 +63,7 @@ class TestLimitInvocation(unittest.TestCase):
         u = Useless()
         wku = weakref.ref(u)
         now = time.time()
-        end = now + 1.1 # a bit more than 1 s
+        end = now + 1.1  # a bit more than 1 s
         while time.time() < end:
             u.doit(time.time(), b=3)
             time.sleep(0.01)
@@ -255,6 +257,7 @@ def tp(trans, ps):
         i = (i + 1) % len(trans)
     return tuple(r)
 
+
 # First we define a bounding boxes, at different locations
 bounding_boxes = [(-2, -2, 0, 0),
                   (-1, -1, 1, 1),
@@ -263,6 +266,7 @@ bounding_boxes = [(-2, -2, 0, 0),
 
 # From this, we generate boxes that are situated all around these
 # bounding boxes, but that do not touch or overlap them.
+
 
 def relative_boxes(bb):
 
@@ -297,6 +301,7 @@ def relative_boxes(bb):
     overlap_boxes = overlap_left + overlap_top + overlap_right + overlap_bottom
 
     return outside_boxes, touching_boxes, overlap_boxes
+
 
 class CanvasTestCase(unittest.TestCase):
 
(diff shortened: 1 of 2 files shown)


src/odemis/util/test/util_test.py
```diff
@@ -396,54 +401,255 @@ def test_line_clipping(self):
...
+                                                                "the closest index to 5 in %s" % in_exp)
+
+
+class NormalizeRectTestCase(unittest.TestCase):
+
+    def test_already_normalized(self):
+        """ Test normalizing a rectangle with already normalized values """
+        rect = (1, 2, 3, 4)
+        normalized_rect = util.normalize_rect(rect)
+        self.assertEqual(normalized_rect, (1, 2, 3, 4), "Failed to normalize the rectangle")
```
tepals, line 506: You could replace  `(1, 2, 3, 4)` with `rect`, that makes it clear that the output is the same as the input

src/odemis/util/test/util_test.py
```diff
@@ -396,54 +401,255 @@ def test_line_clipping(self):
...
+
+
+class ExpandRectTestCase(unittest.TestCase):
+
+    def test_expand_rect(self):
+        """ Test expanding a rectangle """
+        rect = (5, 3, 10, 7)
+        expanded_rect = util.expand_rect(util.normalize_rect(rect), 5)
+        expected_rect = (0, -2, 15, 12)
+        self.assertEqual(expanded_rect, expected_rect,  "Failed to get correct output for rectangle %s with "
```
tepals, line 541: There's 2 spaces between `expected_rect,` and `"Failed...`

src/odemis/util/test/util_test.py
```diff
@@ -396,54 +401,255 @@ def test_line_clipping(self):
...
+
+class FindPlotContentTestCase(unittest.TestCase):
+
+    def test_find_plot_content(self):
+        """ Test finding left and right most non-zero values"""
+        xd = (10, 15, 20, 25, 30)
+        yd = (0, 0, 1, -2, 0)
+        plot_content = util.find_plot_content(xd, yd)
+        self.assertEqual(plot_content, (20, 25), "Failed to get correct output while finding plot content "
+                                                 "for %s with values  %s %s" % (plot_content, xd, yd))
```
tepals, line 553: I think it would also make sense to add a test where there is one or more zero in between the left and right most non-zero value. So something like this for xd and yd:
```
        xd = (10, 15, 16, 17, 20, 25, 30)
        yd = (0, 0, 1, 0, 0, -2, 0)
```

tepals: or should yd also be sorted?

src/odemis/util/test/util_test.py
```diff
@@ -396,54 +401,255 @@ def test_line_clipping(self):
...
+
+
+class WrapToMpiPpiTestCase(unittest.TestCase):
+
+    def test_positive_pi(self):
+        """ Test value is smaller than +pi value """
+        converted_angle = util.wrap_to_mpi_ppi(4.65)
+        self.assertLessEqual(converted_angle, math.pi, "Failed to convert the angle to a value between -pi and +pi "
+                                                  "with the converted value: %s" % converted_angle)
```
tepals, line 562: I'm not sure if it's a PEP8 rule or not, but it looks a bit nicer if you align the string with the string above it:
```
        self.assertLessEqual(converted_angle, math.pi, "Failed to convert the angle to a value between -pi and +pi "
                                                       "with the converted value: %s" % converted_angle)
```

src/odemis/util/test/util_test.py
```diff
@@ -396,54 +401,255 @@ def test_line_clipping(self):
...
+
+    def test_negative_pi(self):
+        """ Test value is bigger than -pi value """
+        converted_angle = util.wrap_to_mpi_ppi(-5.1)
+        self.assertGreaterEqual(converted_angle, -math.pi, "Failed to convert the angle to a value between -pi and +pi "
+                                                      "with the converted value: %s" % converted_angle)
```
tepals, line 568: same here, it looks a bit nicer if you align the string with the string above it

src/odemis/util/test/util_test.py
```diff
@@ -396,54 +401,255 @@ def test_line_clipping(self):
...
+
+        new_dict = util.recursive_dict_update(dict1, dict2)
+        self.assertEqual(new_dict, expected_dict, "Failed to get correct output while updating the dictionary "
+                                                  " %s" % new_dict)
+
+
+class PairwiseTestCase(unittest.TestCase):
+
+    def test_pairwise(self):
+        """ Test pairing """
```
tepals, line 601: I would say  `""" Test pairing two items in a tuple """`

src/odemis/util/test/util_test.py
```diff
@@ -396,54 +401,255 @@ def test_line_clipping(self):
...
+        test = util.pairwise(("item 1", "item 2"))
+        test_list = list(test)[0]
+        self.assertEqual(test_list, ("item 1", "item 2"), "Failed to pair.")
+
+
+class InspectGetMembersTestCase(unittest.TestCase):
+
+    def test_inspect_get_members(self):
+        """ Test inspect get members """
```
tepals, line 610: This one is a bit more complex. I would update this to:
`""" Test the builtin inspect.getmembers raises a TypeError for the dummy class, and that util.inspect_getmembers has the correct output."""`

src/odemis/util/test/util_test.py
```diff
@@ -396,54 +401,255 @@ def test_line_clipping(self):
...
+        testing.start_backend(ENZEL_CONFIG)
         # check if the role now is enzel instead of sparc
         role = model.getMicroscope().role
         self.assertEqual(role, "enzel")
 
 
+class FindClosestTestCase(unittest.TestCase):
+
+    def test_simple(self):
+        """ Finding the closest point in a single list """
```
tepals, line 457: It's not a hard rule but in general for one line docstrings we do not put spaces between the quotes and the text, so this one would become:
```
"""Finding the closest point in a single list"""
```

Feel free to leave it like you had it, or update it for all your test cases.

src/odemis/util/test/util_test.py
```diff
@@ -396,54 +401,255 @@ def test_line_clipping(self):
...
+        in_exp = {
+            tuple([-1, -5, -6]): 0,
+            tuple([1, 2, 3]): 2,
+            tuple([0, 5, 10]): 1,
+            tuple([-15, 30, -60]): 0
+        }
+
+        for testlist, hardcoded_closest_index in in_exp.items():
+            closest = util.index_closest(5, testlist)
+            self.assertEqual(hardcoded_closest_index, closest,  "Failed to get correct output for "
```
tepals, line 496: There's 2 spaces between `closest`, and `"Failed...`

src/odemis/util/test/util_test.py
```diff
@@ -396,54 +401,255 @@ def test_line_clipping(self):
...
+                                                         "the closest value to 5 in %s" % in_exp)
+
+
+class IndexClosestTestCase(unittest.TestCase):
+
+    def test_simple(self):
+        """ Test finding the index of the closest point in a list """
+        simple_list = [1, 5, 6]
+        closest_index = util.index_closest(1, simple_list)
+        self.assertEqual(closest_index, 0,  "Failed to get correct output for "
```
tepals, line 482: There's 2 spaces between `0`, and `"Failed...`

tepals (changes requested):
The test cases already look good! Most my comments are on the style :).

tepals (approved):

PR #2191, opened 2022-08-10
util dataio: refactored _split_planes and added test cases

tepals:
Refactored _splitplanes to improve readability and make the code shorter. This was part of a software knowledge coding exercise.
Added test cases to ensure the code still behaves as expected.

### DIFF
--- a/src/odemis/util/test/dataio_test.py
+++ b/src/odemis/util/test/dataio_test.py
@@ -22,7 +22,7 @@ from odemis import model
 from odemis.acq import stream
 from odemis.dataio import tiff
 from odemis.util.dataio import data_to_static_streams, open_acquisition, \
-    splitext
+    splitext, _split_planes
 import time
 import unittest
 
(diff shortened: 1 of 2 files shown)


src/odemis/util/test/dataio_test.py
```diff
@@ -223,6 +223,60 @@ def test_splitext(self):
...
+            self.assertEqual(da.metadata[model.MD_DIMS], "YX")
+            self.assertEqual(multi_dim_da.metadata[model.MD_DIMS], "CTZYX")
+
+    def test_multi_dim_pLane_xy(self):
+        """
+        Test that for an array with more than 2 dimensions and the x- and y-axes ordered as XY,
+        the input remains unchanged, and the output is correct.
+        """
+        multi_dim_da = self._input_image_3d()
+        multi_dim_da.metadata[model.MD_DIMS] = "CTZXY"
```
BasHolweg, line 271: Why is the MD_DIMS metatdata set to "CTZXY" here when this is already done in self._input_image_3d?

src/odemis/util/test/dataio_test.py
```diff
@@ -223,6 +223,60 @@ def test_splitext(self):
...
+        list_da = _split_planes(multi_dim_da)
+        self.assertIsInstance(list_da, list)
+        for da in list_da:
+            self.assertIsInstance(da, model.DataArray)
+            self.assertEqual(da.shape, (256, 512))
+            self.assertEqual(da.metadata[model.MD_DIMS], "XY")
+            self.assertEqual(multi_dim_da.metadata[model.MD_DIMS], "CTZXY")
```
BasHolweg, line 278: `self.assertEqual(multi_dim_da.metadata[model.MD_DIMS], "CTZXY")`

Does this need to be re-asserted for every instance in list_da?
i.e. why is it in the for loop

src/odemis/util/test/dataio_test.py
```diff
@@ -223,6 +223,60 @@ def test_splitext(self):
...
+
+    def _input_image_3d(self):
+        metadata = {model.MD_SW_VERSION: "1.0-test",
+                    model.MD_HW_NAME: "fake hw",
+                    model.MD_DESCRIPTION: "sem",
+                    model.MD_DIMS: "CTZYX",
+                    }
+
+        return model.DataArray(numpy.zeros((3, 8, 10, 256, 512), dtype=numpy.dtype("uint16")), metadata.copy())
+
```
BasHolweg, line 243: Should there not also be a test case for when the MD_DIMS <= 2 and not "XY" or "YX" since it should also directly return the data when `len(dims) <=2`.

src/odemis/util/test/dataio_test.py
```diff
@@ -223,6 +223,60 @@ def test_splitext(self):
             self.assertEqual(ao, eo, "Unexpected output for '%s': %s" % (inp, ao))
 
 
+class TestSplitPlanes(unittest.TestCase):
+    def _input_image_2d(self):
+        metadata = {model.MD_SW_VERSION: "1.0-test",
+                    model.MD_HW_NAME: "fake hw",
+                    model.MD_DESCRIPTION: "sem",
+                    model.MD_DIMS: "XY"}
```
Kleijwegt, line 231: Can the metadata not be saved as a class attribute? (except the DIMS which can be added/updated in the function)
That also allows to check if the metadata matches with the class atribute and prevents hard coding the same metadata twice.

src/odemis/util/test/dataio_test.py
```diff
@@ -223,6 +223,60 @@ def test_splitext(self):
...
+        Test that for an array with more than 2 dimensions and the x- and y-axes ordered as YX,
+        the input remains unchanged, and the output is correct.
+        """
+        multi_dim_da = self._input_image_3d()
+        list_da = _split_planes(multi_dim_da)
+        self.assertIsInstance(list_da, list)
+        for da in list_da:
+            self.assertIsInstance(da, model.DataArray)
+            self.assertEqual(da.shape, (256, 512))
+            self.assertEqual(da.metadata[model.MD_DIMS], "YX")
```
Kleijwegt, line 262: why is only one part of the metadata checked after the _split_planes func? Maybe the entire metadata should be checked/remain unchanged?

Kleijwegt (changes requested):

BasHolweg (approved):

Kleijwegt (approved):

FCanberk (approved):

PR #2194, opened 2022-08-18
driver xt_client: use the relative move call for relative moves

pieleric:
Don't convert relatives moves directly to absolute moves. Let the SEM
do that. This has a couple of advantages:
* for rotation moves, less complexity with dealing with (fake) out of range,
   and handling cases where going a small negative turn could be
   converted  to a large positive turn.
* No need to read the current position first (and mistakenly force the
move to each axis).
* Simplify the code with range checking, by letting the xtadapter do
that.

### DIFF
--- a/src/odemis/driver/test/xt_client_test.py
+++ b/src/odemis/driver/test/xt_client_test.py
@@ -200,13 +200,14 @@ class TestMicroscope(unittest.TestCase):
         self.assertAlmostEqual(self.stage.position.value["rx"],
                                (abs_pos["rx"] + rel_pos["rx"]), places=4)
 
-        # Relative move in R < 0
+        # Relative move in rz < 0 => it should "wrap around" and report a value near 2pi.
+        # (rx doesn't support full rotation, so we don't use it)
         f = self.stage.moveAbs({"rz": 0})
         rel_pos = {"rz": -0.5} # rad
         f = self.stage.moveRel(rel_pos)
         f.result()
         self.assertAlmostEqual(self.stage.position.value["rz"] % (2 * math.pi),
-                               (0 + rel_pos["rz"]) % (2 * math.pi), places=4)
+                               rel_pos["rz"] % (2 * math.pi), places=4)
 
         self.stage.moveAbs(init_pos)
 
--- a/src/odemis/driver/test/xt_client_test.py
+++ b/src/odemis/driver/test/xt_client_test.py
@@ -199,6 +199,15 @@ class TestMicroscope(unittest.TestCase):
                                (abs_pos["rz"] + rel_pos["rz"]) % (2 * math.pi), places=4)
         self.assertAlmostEqual(self.stage.position.value["rx"],
                                (abs_pos["rx"] + rel_pos["rx"]), places=4)
+
+        # Relative move in R < 0
+        f = self.stage.moveAbs({"rz": 0})
+        rel_pos = {"rz": -0.5} # rad
+        f = self.stage.moveRel(rel_pos)
+        f.result()
+        self.assertAlmostEqual(self.stage.position.value["rz"] % (2 * math.pi),
+                               (0 + rel_pos["rz"]) % (2 * math.pi), places=4)
+
         self.stage.moveAbs(init_pos)
 
     def test_hfov(self):
--- a/src/odemis/driver/xt_client.py
+++ b/src/odemis/driver/xt_client.py
@@ -2051,8 +2051,11 @@ class Stage(model.Actuator):
         # actually allows rotation in both directions)
         for an in ("rx", "rz"):
             rng = self.axes[an].range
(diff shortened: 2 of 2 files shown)


src/odemis/driver/xt_client.py
```diff
@@ -2047,19 +2047,29 @@ def _getPosition(self):
         pos = self.parent.get_stage_position()
         pos["rx"] = pos.pop("t")
         pos["rz"] = pos.pop("r")
+        # Make sure the full rotations are within the range (because the SEM
+        # actually allows rotation in both directions)
+        for an in ("rx", "rz"):
+            rng = self.axes[an].range
+            if util.almost_equal(rng[1] - rng[0], 2 * math.pi):
+                pos[an] = pos[an] % (2 * math.pi) - rng[0]
```
tepals, line 2055: are you sure this is correct? For example, if `pos[an] = 0.5*math.pi` and the range is `[-math.pi, +math.pi]` I think this line will update `pos[an]` to `1.5*math.pi` instead of staying `0.5*math.pi`

src/odemis/driver/test/xt_client_test.py
```diff
@@ -199,6 +199,15 @@ def test_move_stage_rot_tilt(self):
...
         self.assertAlmostEqual(self.stage.position.value["rx"],
                                (abs_pos["rx"] + rel_pos["rx"]), places=4)
+
+        # Relative move in R < 0
+        f = self.stage.moveAbs({"rz": 0})
+        rel_pos = {"rz": -0.5} # rad
+        f = self.stage.moveRel(rel_pos)
+        f.result()
+        self.assertAlmostEqual(self.stage.position.value["rz"] % (2 * math.pi),
+                               (0 + rel_pos["rz"]) % (2 * math.pi), places=4)
```
tepals, line 209: can you add a comment on why you add the zero in: `(0 + rel_pos["rz"])`

src/odemis/driver/xt_client.py
```diff
@@ -2097,22 +2106,11 @@ def _moveTo(self, future, pos, timeout=60):
...
-            if rng == (0, 2 * math.pi) and an in ("rx", "rz"):
-                pos[an] = pos[an] % (2 * math.pi)
-                target_pos[an] = target_pos[an] % (2 * math.pi)
-            p = target_pos[an]
-            if not rng[0] <= p <= rng[1]:
-                raise ValueError("Relative move would cause axis %s out of bound (%g m)" % (an, p))
-
-        self._moveTo(future, pos)
+        """
+        shift (dict): position in internal coordinates
```
tepals, line 2110: what are internal coordinates?

src/odemis/driver/xt_client.py
```diff
@@ -2097,22 +2106,11 @@ def _moveTo(self, future, pos, timeout=60):
...
+        """
+        # We don't check the target position fit the range, the xt-adapter will take care of that
```
tepals, line 2112: the range is checked in `moveRel` will that cause any problems?

tepals (changes requested):

src/odemis/driver/xt_client.py
pieleric: Only with `_checkMoveRel()`, which almost doesn't check anything. It only checks that the shift is less than the entire range of the stage (which would be an obvious sign that the shift is too big). So it's very weak. However, the xt-adapter checks more precisely already... and I expect the XTlib does too.

src/odemis/driver/xt_client.py
pieleric: The user can invert the direction of an axis (with the parameter `inverted`). That's what `applyInversion()` takes care of. "Internal coordinates" mean *without* the inversion applied, ie the hardware point-of-view.

src/odemis/driver/xt_client.py
pieleric: Indeed! I fixed it to `(pos[an] - rng[0]) % (2 * math.pi) + rng[0]`. It looks to me like the proper formula, right?

src/odemis/driver/test/xt_client_test.py
pieleric: I wanted to keep the same formula as the test above, which is old_pos + rel_pos. However, that's probably more confusing than helpful... will remove it.

src/odemis/driver/test/xt_client_test.py
```diff
@@ -199,6 +199,15 @@ def test_move_stage_rot_tilt(self):
...
+
```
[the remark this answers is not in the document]
pieleric: rx has a limited range. It cannot wrap around. I've added a comment.

src/odemis/driver/xt_client.py
tepals: yes that looks correct

tepals (approved):

PR #2199, opened 2022-08-22
Github actions workflow: update to use ubuntu 20.04 instead of ubuntu…

tepals:
… 18.04, because the ubuntu 18.04 actions runner image will soon be deprecated in the github workflow.

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -8,7 +8,7 @@ on:
 jobs:
   build:
 
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


.github/workflows/python-app.yml
```diff
@@ -18,12 +18,12 @@ jobs:
         sudo apt-get -y update
 
         sudo apt-get -y install python3-pil python3-yaml python3-serial python3-zmq \
-        python3-scipy python3-decorator cgroup-bin \
+        python3-scipy python3-decorator \
         python3-libtiff python3-comedilib python3-h5py libtiff5-dev python3-future \
         python3-pkg-resources python3-netifaces python3-cairo python3-numpy python3-scipy \
         python3-decorator python3-xdg python3-suds python3-matplotlib python3-opencv \
         python3-wxgtk4.0 python3-serial python3-evdev python3-notify2 python3-psutil \
-        graphicsmagick cgroup-bin authbind python3-setuptools cython python3-pyro4-delmic
```
pieleric, line 26: That one is now called "cgroup-tools". You probably want to explicitly install it. (it's need to start an odemis backend)

pieleric:
> the scipy.misc.bytescale function is deprecated. I think we can replace it with skimage.util.img_as_ubyte, do you agree?

We don't depend on skimage, so I'd rather avoid adding this dependency just for such a simple function. Could you check in OpenCV if there is something fitting? Otherwise, we could write it with numpy (like here: https://codereview.stackexchange.com/questions/185785/scale-numpy-array-to-certain-range) .

PR #2203, opened 2022-08-25
odemisd: extend the file mask to include the execute bit

pieleric:
The execute bit on the mask doesn't really mean much for new files.
However, when creating directories, that allows to add files to the
directory... which is a very useful feature.
No driver seemed to depend on this, however on Ubuntu 20.04, some
libraries depend on OpenMPI, which creates some directories in the
background.
This prevented the backend from being started as a normal user on Ubuntu
20.04.

### DIFF
--- a/src/odemis/odemisd/main.py
+++ b/src/odemis/odemisd/main.py
@@ -518,10 +518,12 @@ class BackendRunner(object):
         try:
             os.setgid(gid_base)
         except OSError:
+            # This can happen especially when running the backend as a standard user.
             logging.warning("Not enough permissions to run in group " + model.BASE_GROUP + ", trying anyway...")
 
-        # everything created after must be rw by group
-        os.umask(~(stat.S_IRUSR | stat.S_IWUSR | stat.S_IRGRP | stat.S_IWGRP))
+        # Everything created after should be (also) accessible by the group.
+        # Need the user execute bit, to allow directory creation (with files inside)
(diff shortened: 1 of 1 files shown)


src/odemis/odemisd/main.py
```diff
@@ -518,10 +518,12 @@ def set_base_group(self):
...
             os.setgid(gid_base)
         except OSError:
+            # This can happen especially when running the backend as a standard user.
             logging.warning("Not enough permissions to run in group " + model.BASE_GROUP + ", trying anyway...")
 
-        # everything created after must be rw by group
-        os.umask(~(stat.S_IRUSR | stat.S_IWUSR | stat.S_IRGRP | stat.S_IWGRP))
+        # Everything created after should be (also) accessible by the group.
+        # Need the user execute bit, to allow directory creation (with files inside)
+        os.umask(~(stat.S_IRWXU | stat.S_IRGRP | stat.S_IWGRP))
```
tepals, line 526: is it possible/does it make sense to add a test case for this, or is this already tested?

pieleric: Well, without it, a lot of test cases failed to run on Ubuntu 20.04, including most of the ones in odemisd/test/. So it was indirectly tested. There is nothing explicitly checking that the files created in /var/run/odemisd/ are forbidden to the non odemis group, but I don't think it's super important to check per-se.

PR #2212, opened 2022-09-05
csv export function refactor for readability of code

sneepstefan:
added test case for missing functionality

### DIFF
--- a/src/odemis/dataio/csv.py
+++ b/src/odemis/dataio/csv.py
@@ -31,7 +31,7 @@ CAN_SAVE_PYRAMID = False
 
 
 def export(filename, data):
-    '''
+    """
     Write a CSV file:
         - If the given data is spectrum data write it as series of wavelength->intensity
         - If the given data is time data write it as series of time->intensity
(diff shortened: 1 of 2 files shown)


src/odemis/dataio/csv.py
```diff
@@ -45,159 +45,260 @@ def export(filename, data):
...
+        logging.info("Exporting spectrum without wavelength information")
+        spectrum_tuples = data.reshape(data.shape[0], 1)
+        headers = ['# intensity']
+
+    with open(filename, 'w', newline='') as fd:
+        csv_writer = csv.writer(fd)
+        csv_writer.writerow(headers)
+        csv_writer.writerows(spectrum_tuples)
+
+    fd.close()
```
pieleric, line 119: You don't need to close the file. That's what `with open()` takes care of.
(same for all the new functions)

src/odemis/dataio/csv.py
```diff
@@ -45,159 +45,260 @@ def export(filename, data):
...
+    """
+    This will export the spectrum data to a CSV file using the type of microscope,
+    the acquired data and filename title in its creation.
+    :param acq_type: data.metadata.get(model.MD_ACQ_TYPE, None)
+    :param data: (model.DataArray)
+    :param filename: (unicode)
+    """
+
+    if acq_type != model.MD_AT_SPECTRUM:
+        logging.warning("Data seems to be spectrum, but acq_type is %s", acq_type)
```
pieleric, line 101: I think it's strange that not all functions have the same signature. Let's move this check back into the `if/elif` block, and drop the `acq_type` argument`. This way all the functions would take `(data, filename)` as arguments, only.

src/odemis/dataio/csv.py
```diff
@@ -45,159 +45,260 @@ def export(filename, data):
...
-            csv_writer.writerow(headers)
-            csv_writer.writerows(rows)
 
+def _export_spectrum_data(acq_type, data, filename):
+    """
+    This will export the spectrum data to a CSV file using the type of microscope,
+    the acquired data and filename title in its creation.
+    :param acq_type: data.metadata.get(model.MD_ACQ_TYPE, None)
+    :param data: (model.DataArray)
+    :param filename: (unicode)
```
pieleric, line 97: In Python3 this is now called "str". 
Actually, you can now also indicate the type directly in the function signature, which has the advantage of being parsed by more tools. Something like:
`_export_spectrum_data(data: model.DataArray, filename: str):`

While you're at it, please also update the signature of `export()` .

src/odemis/dataio/csv.py
```diff
@@ -45,159 +45,260 @@ def export(filename, data):
...
+    """
+    This will export the angular spectrum data to a CSV file using the type
+    of microscope, the acquired data and filename title in its creation.
+    :param acq_type:
+    :param data:
+    :param filename:
+    :return:
+    """
+    if acq_type != model.MD_AT_EK:
+        logging.warning("Data seems to be EK, but acq_type is %s", acq_type)
```
pieleric, line 283: Let's move that check also back to the if/elif block.

src/odemis/dataio/csv.py
```diff
@@ -45,159 +45,260 @@ def export(filename, data):
...
+    """
+    This will export the temporal spectrum data to a CSV file using the type
+    of microscope, the acquired data and filename title in its creation.
+    :param acq_type:
+    :param data:
+    :param filename:
+    :return:
+    """
+    if acq_type != model.MD_AT_TEMPSPECTRUM:
+        logging.warning("Data seems to be temporal spectrum, but acq_type is %s", acq_type)
```
pieleric, line 251: Let's move to the if/elif block too.

src/odemis/dataio/csv.py
```diff
@@ -45,159 +45,260 @@ def export(filename, data):
...
 
-        angle_range, unit_a = spectrum.get_angle_range(data)
-        if unit_a == "rad":
-            unit_a = "°"
-            angle_range = [math.degrees(theta) for theta in angle_range]  # Convert radians to degrees
-            assert len(angle_range) == len(data)
+    # Export AR Data
+    elif acq_type == model.MD_AT_AR:
+        logging.debug("Exporting AR data to CSV")
+        _export_ar_data(data, filename)
```
tepals, line 85: You moved this to the end of the `elif` statements, with `elif` order does matter. I'm not sure if it matters in this case, but I would move this back to in between "Angle-line data" and "Spectrum data"

src/odemis/dataio/csv.py
```diff
@@ -45,159 +45,260 @@ def export(filename, data):
...
-        with open(filename, 'w') as fd:
-            csv_writer = csv.writer(fd)
-            csv_writer.writerow(headers)
-            csv_writer.writerows(rows)
 
+def _export_spectrum_data(acq_type, data, filename):
+    """
+    This will export the spectrum data to a CSV file using the type of microscope,
+    the acquired data and filename title in its creation.
+    :param acq_type: data.metadata.get(model.MD_ACQ_TYPE, None)
```
tepals, line 95: in this case `data.metadata` is a dict and with `.get` you get an element from the dict. In this case the element is a string, or if it doesn't exist `None`. So this should be written as:
```
:param acq_type: (str or None)
```

src/odemis/dataio/csv.py
tepals, line 97: you can just use `(str)` instead of `(unicode)`

src/odemis/dataio/test/csv_test.py
```diff
@@ -240,6 +240,29 @@ def testExportSpectrumLineNoWL(self):
             raised = True
         self.assertFalse(raised, 'Failed to read csv file')
 
+    def testExportTemporalSpectrum(self):
+        """Try simple temporal spectrum export; note that size is made up and is not specific"""
+        size = (150, 340)
+        dtype = numpy.uint16
+        md = {model.MD_WL_LIST: numpy.linspace(536e-9, 650e-9, size[0]).tolist(),
```
pieleric, line 247: Actually, it should be `size[1]`, as the spectrum (C) is the second dimension. This error is also in `testExportSpectrumLine()`. Could you fix it too? (I think we should have some logging when this happens, because for now, it's silently dropped. I'll have a look at it).

src/odemis/dataio/test/csv_test.py
```diff
@@ -240,6 +240,29 @@ def testExportSpectrumLineNoWL(self):
...
+              model.MD_ACQ_TYPE: model.MD_AT_TEMPSPECTRUM,
```
pieleric, line 248: If you want to test the time column, you'll need another metadata with model.MD_TIME_LIST (typically with about 1 ps per pixel). It's not necessary, as the function must also work without that metadata, but the typical usage will provide that metadata...

src/odemis/dataio/csv.py
```diff
@@ -45,159 +45,260 @@ def export(filename, data):
...
+    fd.close()
+
+
+def _export_chronogram_data(data, filename):
+    """
+    This will export the chronogram data to a CSV file using the acquired data,
+    and filename title in its creation.
+    :param data:
+    :param filename:
+    :return:
```
tepals, line 128: there is no return, so you can remove this line. This is the same for your other functions

src/odemis/dataio/csv.py
```diff
@@ -45,159 +45,260 @@ def export(filename, data):
...
+    fd.close()
+
+
+def _export_spectrum_line_data(data, filename):
+    """
+    This will export the line data to a CSV file using the acquired data,
+    and filename title in its creation.
+    :param data:
+    :param filename:
+    :return:
```
pieleric, line 215: I'm not super convinced of having the parameters defined in the docstring, without actually any definition. I think you could drop them. At least, I'd suggest to drop the `return` as the function doesn't return anything.

pieleric (changes requested):
Overall, the code looks a lot nicer :smile:

src/odemis/dataio/csv.py
```diff
@@ -45,159 +45,260 @@ def export(filename, data):
...
-            csv_writer.writerows(rows)
 
+def _export_spectrum_data(acq_type, data, filename):
+    """
+    This will export the spectrum data to a CSV file using the type of microscope,
+    the acquired data and filename title in its creation.
+    :param acq_type: data.metadata.get(model.MD_ACQ_TYPE, None)
+    :param data: (model.DataArray)
+    :param filename: (unicode)
+    """
```
tepals, line 98: can you also update the parameters with a short description? Something like:
```
    :param acq_type: (str or None) type of acquisition 
    :param data: (model.DataArray) data that will be exported
    :param filename: (str) filename to save the data to
```
And can you also add this to the other docstrings?

tepals (changes requested):

src/odemis/dataio/csv.py
sneepstefan: You are right when I look more into detail regarding this statement I found that the __exit__() method is implemented in this statement closing the file IO afterwards automatically. I was not aware of that functionality in Python, thanks for pointing this out.

src/odemis/dataio/csv.py
sneepstefan: I totally agree with that, could you point out to me for which acquisition type I have to do a check regarding to data type?

pieleric: I think that all the acquisition types possible to check are already in the code. The other type of export (eg, angle data only) doesn't match an existing acquisition type (they are obtained by processing data from an acquisition, not a direct acquisition).

src/odemis/dataio/csv.py
sneepstefan: You are right I just copied the whole syntax to get an element from the dictionary this is a bit odd now that I see this myself. Resolved this by removing whole type parameter as Eric suggested in his comments.

src/odemis/dataio/csv.py
sneepstefan: Yes, I have updated this now for every method which was added. To be honest I should have done this beforehand but somehow it did slip through my attention.

src/odemis/dataio/csv.py
sneepstefan: Considered the comments of Thera I choose to keep all of the docstrings with added definitions now. Indeed all return parameters have been deleted as well as the acq_type parameters.

src/odemis/dataio/csv.py
sneepstefan: Yes you are right, elif order does matter but since this specific elif checks completely different parameters I thought putting it on a different spot would make it more stand out. Does this check not have dimensions to check for as parameters? For now I moved it back to its original position as I do not want the code to behave different.

tepals (approved):

pieleric (approved):
Looks good to me!

PR #2214, opened 2022-09-08
acq path: distinguish "ek" path and "ek-align" paths

pieleric:
For ek-align, we used to just use the "ek" path. However, we want to
force the grating and the filter to be in "good" positions.
That'd be ugly to do in the GUI, so explicitly place it in the OPM.
It will force the axes everytime the stream starts. The GUI still allows
the user to change to different (and odd) positions while playing the
stream.

For the standard EK streams, the GUI takes care of setting the default
position of these streams when it's added by the user (and the user can
later override it).

Also extend the path test cases, to actually test a EK simulator.

Also found a bug in the Sparc EK polarizer simulator.

### DIFF
--- a/src/odemis/acq/test/path_test.py
+++ b/src/odemis/acq/test/path_test.py
@@ -52,8 +52,46 @@ SPEC_CONFIG = CONFIG_PATH + "sim/sparc-sim-spec.odm.yaml"
 SPARC2_CONFIG = CONFIG_PATH + "sim/sparc2-sim.odm.yaml"
 SPARC2_EXT_SPEC_CONFIG = CONFIG_PATH + "sim/sparc2-ext-spec-sim.odm.yaml"
 SECOM_FLIM_CONFIG = CONFIG_PATH + "sim/secom-flim-sim.odm.yaml"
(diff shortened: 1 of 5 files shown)


src/odemis/acq/test/path_test.py
```diff
@@ -52,10 +52,49 @@
...
+    if isinstance(pos, tuple):  # multiple options
+        for p in pos:
+            # From the metadata?
+            if isinstance(p, str) and p.startswith("MD:"):
+                md_name = p[3:]
+                try:
+                    pos = md[md_name][axis]
+                    logging.debug("Found metadata %s[%s]= %s", md_name, axis, pos)
+                    break
+                except KeyError:  # no such metadata, or axis on the metadata
```
sneepstefan, line 78: Is it not necessary to do debug logging when a KeyError exception occurs?

sneepstefan (approved):

src/odemis/acq/test/path_test.py
pieleric: Not really, this is actually very common.
However, the logging.debug() above should go away. That is left over from my debugging, but will fill up the testing log unnecessarily.

PR #2223, opened 2022-09-16
driver FixedPositionActuator method improvement and addition of testcase

sneepstefan:
After moving to a position, the actual reported position might not be exactly the position requested. It’s can be because the stage has a quantized position (ie, number of steps) like the TMCM, or the stage has an encoder which keeps updating to positions very near-by.

When asking to move again to the “same” position, the FixedPositionActuator passes on the request. Usually, it has little effect, because it causes a very tiny move, which is either discarded, or very quick. However, in case of anti-backlash under it, there is 50% chance that it it thinks it’s in the “bad” direction, and causes a small step back, followed by a small step forward. That can be quite long.

=> In FixedPositionActuator, after moving to a position, remember where it arrived. Later on, when requested to move to a position, if the dependency is already at either the requested position, or the previous position, don’t do anything.

### DIFF
--- a/src/odemis/driver/actuator.py
+++ b/src/odemis/driver/actuator.py
@@ -29,6 +29,8 @@ import numbers
 import threading
 from concurrent.futures._base import CancelledError, FINISHED
 from concurrent import futures
+from typing import Dict, Union, Set
+
 import numpy
 from past.builtins import basestring
 
@@ -1307,13 +1309,14 @@ class FixedPositionsActuator(model.Actuator):
 
         self._cycle = cycle
         self._move_sum = 0
-        self._position = {}
-        self._referenced = {}
+        self._position: Dict[str, float] = {}
+        self._referenced: Dict[str, bool] = {}
         axis, dep = list(dependencies.items())[0]
         self._axis = axis
         self._dependency = dep
         self._caxis = axis_name
-        self._positions = positions
+        self._positions: Union[Set[float], Dict[float, str]] = positions
+        self._actual_positions: Dict[float, float] = {}  # one choice position -> dependency position
         # Executor used to reference and move to nearest position
         self._executor = CancellableThreadPoolExecutor(max_workers=1)  # one task at a time
 
@@ -1333,7 +1336,7 @@ class FixedPositionsActuator(model.Actuator):
 
         model.Actuator.__init__(self, name, role, axes=axes, dependencies=dependencies, **kwargs)
 
-        self._position = {}
+        # self._position = {}  # isn't this a double declaration of empty dict?
         self.position = model.VigilantAttribute({}, readonly=True)
 
         logging.debug("Subscribing to position of dependency %s", dep.name)
@@ -1399,7 +1402,7 @@ class FixedPositionsActuator(model.Actuator):
     def moveRel(self, shift):
         if not shift:
             return model.InstantaneousFuture()
-        self._checkMoveRel(shift)
+        # self._checkMoveRel(shift)
         raise NotImplementedError("Relative move on fixed positions axis not supported")
 
     @isasync
@@ -1408,6 +1411,7 @@ class FixedPositionsActuator(model.Actuator):
         Move the actuator to the defined position in m for each axis given.
         pos dict(string-> float): name of the axis and position in m
         """
+
         if not pos:
             return model.InstantaneousFuture()
         self._checkMoveAbs(pos)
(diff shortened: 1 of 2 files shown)


src/odemis/driver/actuator.py
```diff
@@ -1333,7 +1336,7 @@ def __init__(self, name, role, dependencies, axis_name, positions, cycle=None,
 
         model.Actuator.__init__(self, name, role, axes=axes, dependencies=dependencies, **kwargs)
 
-        self._position = {}
+        # self._position = {}  # isn't this a double declaration of empty dict?
```
pieleric, line 1339: You can drop the line entirely.

src/odemis/driver/actuator.py
```diff
@@ -1399,7 +1402,7 @@ def _updateReferenced(self):
     def moveRel(self, shift):
         if not shift:
             return model.InstantaneousFuture()
-        self._checkMoveRel(shift)
+        # self._checkMoveRel(shift)
```
pieleric, line 1405: Let's drop this line completely.

src/odemis/driver/actuator.py
```diff
@@ -1417,23 +1421,33 @@ def moveAbs(self, pos):
         return f
 
     def _doMoveAbs(self, pos):
-        axis, distance = list(pos.items())[0]
-        logging.debug("Moving axis %s (-> %s) to %g", self._axis, self._caxis, distance)
+        axis, req_pos = list(pos.items())[0]
+
+        # if the actuator is already at the requested position skip movement
+        # this to prevent the actuator to move due to difference dep pos and requested pos
+        # and en up in the same place again. For instance, this is typical when anti-backlash is used.
```
pieleric, line 1428: This comment needs a little bit of love. I'd write it like:
> If the actuator is already at the requested position, skip the move.
> This prevents the actuator from moving just due to tiny difference between actual pos and requested pos,
> in which case it typically ends up in the same place again. For instance, this can happen when anti-backlash is used.

src/odemis/driver/actuator.py
```diff
@@ -1307,13 +1309,14 @@ def __init__(self, name, role, dependencies, axis_name, positions, cycle=None,
...
-        self._referenced = {}
+        self._position: Dict[str, float] = {}
+        self._referenced: Dict[str, bool] = {}
         axis, dep = list(dependencies.items())[0]
         self._axis = axis
         self._dependency = dep
         self._caxis = axis_name
-        self._positions = positions
+        self._positions: Union[Set[float], Dict[float, str]] = positions
+        self._actual_positions: Dict[float, float] = {}  # one choice position -> dependency position
```
pieleric, line 1319: Let's add a comment above this variable to describe what it is. Something like "Latest position reached after move to given position was requested".

src/odemis/driver/actuator.py
```diff
@@ -1307,13 +1309,14 @@ def __init__(self, name, role, dependencies, axis_name, positions, cycle=None,
 
         self._cycle = cycle
         self._move_sum = 0
-        self._position = {}
-        self._referenced = {}
+        self._position: Dict[str, float] = {}
```
pieleric, line 1312: This variable is basically never used. We could entirely get rid of it by merging `updatePosition()` into  `_update_dep_position()`. This would especially avoid having the confusing `_positions` vs `_position`.

Could you clean it up?

src/odemis/driver/test/actuator_test.py
```diff
@@ -197,6 +197,63 @@ def tearDown(self):
...
+                                                 {"x": self.ab_stage},
+                                                 "x",
+                                                 # Fixed positions, which cannot be reached exactly with the ustepsize
+                                                 {0 + 2e-6: "pos0", 0.001 + 3e-6: "pos1",
+                                                  0.002 - 2e-6: "pos2", 0.003 - 6e-6: "pos3"}
+                                                 )
+        self.dev_cycle = FixedPositionsActuator("stage", "stage",
+                                                {"x": self.ab_stage}, "x", {0: "pos0", 0.01: "pos1",
+                                                                      0.02: "pos2", 0.03: "pos3",
+                                                                      0.04: "pos4", 0.05: "pos5"}, cycle=0.06)
```
pieleric, line 226: You never use that one => drop it, and call `dev_normal` just `dev`.

pieleric (changes requested):

src/odemis/driver/test/actuator_test.py
```diff
@@ -68,27 +68,25 @@ def test_speed(self):
...
-                                         port="/dev/fake3",
-                                         axes=["a", "b"],
-                                         ustepsize=[5.9e-9, 5.8e-9],
-                                         rng=[[-1e-3, 1e-3], [0, 1e-3]],
-                                         refproc="Standard")
+                                              port="/dev/fake3",
+                                              axes=["a", "b"],
+                                              ustepsize=[5.9e-9, 5.8e-9],
+                                              rng=[[-1e-3, 1e-3], [0, 1e-3]],
+                                              refproc="Standard")
```
pieleric, line 79: Please, please, don't reformat the entire file whenever you have some change in it. It hides your "real" changes.
If the format is really terrible, do the reformatting as a separate first commit.
For this PR, let's not worry about it, and leave it like that, but avoid doing it in the future.

pieleric (approved):

src/odemis/driver/test/actuator_test.py
sneepstefan: Yes I thought indeed that went wrong at some point. It was never my intention to reformat the whole file only the part that I did the modification in. Somehow my IDE suggested i'd reformat the whole file and I clicked that too quickly I guess.
Thanks for reviewing and approving eventually.

FCanberk (approved):

PR #2225, opened 2022-09-16
acq move: concidered ENZEL alignement position also part of the THREE_BEAMS position

pieleric:
It's all very near, we cannot really distinguish the positions.
So concider them equivalent.

Also fix some test cases.

### DIFF
--- a/src/odemis/acq/test/move_test.py
+++ b/src/odemis/acq/test/move_test.py
@@ -195,9 +195,6 @@ class TestEnzelMove(unittest.TestCase):
         Test getCurrentPositionLabel function behaves as expected
         """
         aligner = self.aligner
-        # at start the aligner wouldn't be in one of the predefined positions
-        pos_label = getCurrentAlignerPositionLabel(aligner.position.value, aligner)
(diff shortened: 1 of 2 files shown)


FCanberk (approved):

src/odemis/acq/test/move_test.py
```diff
@@ -252,23 +251,27 @@ def test_get_current_position(self):
...
         time.sleep(7)
         f.cancel()
         pos_label = getCurrentPositionLabel(stage.position.value, stage)
-        self.assertEqual(pos_label, LOADING_PATH)
+        # It's really hard to get the timing right, so also allow to be at loading
+        # or three-beams
+        self.assertIn(pos_label, (THREE_BEAMS, LOADING, LOADING_PATH))
 
         # Move to imaging position
         cryoSwitchSamplePosition(LOADING).result()
```
BasHolweg, line 263: Is it not necessary to check whether the CurrentPositionLabel correctly switched to LOADING?

pieleric: We already do that line 250. It's not really needed to check again.

PR #2226, opened 2022-09-19
Build support for multiple temperature control

FCanberk:
Built support to communicate a LakeShore350 device through ethernet cable.
Built support to control multiple temperatures for LakeShore350 devices.


src/odemis/driver/lakeshore.py
```diff
@@ -69,33 +74,23 @@ class Lakeshore(model.HwComponent):
...
-
         name: (str)
         role: (str)
         port: (str) port name. Can be a pattern, in which case all the ports
           fitting the pattern will be tried.
           Use /dev/fake for a simulator
-        sensor_input (str): The sensor input to use, typically 'A' or 'B'
-        output_channel: (int): The channel output to control, typically 1 or 2
+        sensor_input (str): The sensor input to use, typically 'A' or 'B' for model 335 and
+            'A', 'B', 'C' or 'D' for model 350
```
pieleric, line 83: It now has two ways to specify it. Please describe both ways. I'd write something like:
`sensor_input (str or dict str->str): either a single input name, or the name of the VigilantAttribute (to be followed by temperature) -> input name`

Same thing for the output_channel.

src/odemis/driver/lakeshore.py
```diff
@@ -69,33 +74,23 @@ class Lakeshore(model.HwComponent):
...
-        if not self._sensor_input in ('A', 'B'):
-            raise ValueError("Sensor input must be either 'A' or 'B'")
-        
-        self._output_channel = output_channel
-        if not self._output_channel in (1, 2):
-            raise ValueError("Invalid output channel. Should be an int of 1 or 2")
-        
+        self._accesser = None
+
         self._port = self._findDevice(port)  # sets ._serial and ._file
```
pieleric, line 92: Now it doesn't set _serial and _file, but instead "_accesser".

src/odemis/driver/lakeshore.py
```diff
@@ -108,11 +103,74 @@ def __init__(self, name, role, port, sensor_input='b', output_channel=2, **kwarg
...
-                                     setter=self._set_heating)
+        # set supported sensor input and output channel values
+        if md == "MODEL335":
+            supported_sensor_inputs = ('A', 'B')
+            supported_output_channels = (1, 2)
+        elif md == "MODEL350":
+            supported_sensor_inputs = ('A', 'B', 'C', 'D')
+            supported_output_channels = (1, 2, 3, 4)
+        else:
+            raise IOError("Incompatible device model detected.")
```
pieleric, line 114: Please add the name of the model in the error message.

src/odemis/driver/lakeshore.py
```diff
@@ -108,11 +103,74 @@ def __init__(self, name, role, port, sensor_input='b', output_channel=2, **kwarg
...
+
+        # Check if there's only one sensor input and the input is defined as string
+        if isinstance(sensor_input, str):
+            sensor_input = {"": sensor_input}
+
```
pieleric, line 119: Drop this empty line, as there is an `elif`.

src/odemis/driver/lakeshore.py
```diff
@@ -108,11 +103,74 @@ def __init__(self, name, role, port, sensor_input='b', output_channel=2, **kwarg
...
+        # Check if there's one or multiple sensor input(s) and defined as dictionary
+        elif not isinstance(sensor_input, dict):
+            raise ValueError("Invalid sensor input type. Sensor input must be either str or dict")
+
+        self._sensor_inputs = {k: v.upper() for k, v in sensor_input.items()}
+        for va_name, sensor_name in self._sensor_inputs.items():
+            if sensor_name not in supported_sensor_inputs:
+                raise ValueError(f"Sensor input must be one of the following: {supported_sensor_inputs}")
```
pieleric, line 127: Add the sensor_name in the error message.

src/odemis/driver/lakeshore.py
```diff
@@ -108,11 +103,74 @@ def __init__(self, name, role, port, sensor_input='b', output_channel=2, **kwarg
...
+                full_va_name = f"{va_name}Temperature"
+            setattr(self, full_va_name, va)
+
+        # Check if there's only one int type output channel
+        if isinstance(output_channel, int):
+            output_channel = {"": output_channel}
+        elif not isinstance(output_channel, dict):
+            raise ValueError("Invalid output channel type. Output channel must be either str or dict")
+
+        self._setters = []
```
pieleric, line 143: Add a comment to explain why it's there. Something like "Keep reference to the partial functions, to avoid them being garbage collected".

src/odemis/driver/lakeshore.py
```diff
@@ -163,45 +220,72 @@ def _openSerialPort(port, baudrate=57600):
...
+
+            manufacturer, md, _, _ = self.GetIdentifier()  # if value is incorrect, will throw an exception wile unpacking
+
+            if manufacturer != "LSCI":
+                raise IOError("Invalid device manufacturer")
+
+            compatible_devices = ["MODEL335", "MODEL350"]
+            if md not in compatible_devices:
+                raise IOError("Incompatible device model detected. "
+                              "Compatible models are %s." % (compatible_devices,))
```
pieleric, line 264: You should be able to replace this by `self._checkDeviceCompatibility()`.

src/odemis/driver/lakeshore.py
```diff
@@ -163,45 +220,72 @@ def _openSerialPort(port, baudrate=57600):
...
 
                 if manufacturer != "LSCI":
                     raise IOError("Invalid device manufacturer")
-                if md != "MODEL335":
-                    raise IOError("The model is %s, not MODEL335." % (md,))
+
+                compatible_devices = ["MODEL335", "MODEL350"]
+                if md not in compatible_devices:
+                    raise IOError("Incompatible device model detected. "
+                                  "Compatible models are %s." % (compatible_devices,))
```
pieleric, line 288: replace by a call to `_checkDeviceCompatibility()`.

src/odemis/driver/lakeshore.py
```diff
@@ -287,32 +355,32 @@ def GetIdentifier(self):
...
         """
         Get the temperature at the thermocouple junction
         """
         return float(self._sendQuery(b"TEMP?"))
 
-    def SetSetpoint(self, temp):
+    def SetSetpoint(self, output_channel, temp):
         """
         Set the temperature setpoint
         temp (float): the temperature to set, in Celsius
```
pieleric, line 368: Please update the docstrings of all these functions where you've added a argument, with the description of the argument.

src/odemis/driver/lakeshore.py
```diff
@@ -344,98 +411,220 @@ def HeaterSetup(self,
...
         """
-        return int(self._sendQuery(b'RANGE? %d' % (self._output_channel,)))
+        return int(self._sendQuery(b'RANGE? %d' % (output_channel,)))
 
     # Internal API functions
-    def _set_targetTemperature(self, value):
+    def _set_targetTemperature(self, value, output_channel):
         """
-        Setter for the targetTemperature VA
+        Setter for the targetTemperature VA.
```
pieleric, line 460: It now handles multiple VAs, so add a "s" after VA.

src/odemis/driver/lakeshore.py
```diff
@@ -344,98 +411,220 @@ def HeaterSetup(self,
...
         VA is in Celsius, but controller uses Kelvin
```
pieleric, line 461: Explain in the docstring what "output_channel" is for.

src/odemis/driver/lakeshore.py
```diff
@@ -344,98 +411,220 @@ def HeaterSetup(self,
...
         self.checkError()
         if abs(value - svalue) >= 0.01:
             logging.warning("Did not set new target temperature to %f", value)
         return svalue
 
-    def _set_heating(self, value):
+    def _set_heating(self, value, output_channel):
         """
         Setter for the heater range VA
         """
```
pieleric, line 475: Update the doctring here too.

src/odemis/driver/lakeshore.py
```diff
@@ -344,98 +411,220 @@ def HeaterSetup(self,
...
-            self.checkError()
+            for va_name, sensor_name in self._sensor_inputs.items():
+                temp = self.GetSensorTemperature(sensor_name)
+                if va_name == "":
+                    full_va_name = "temperature"
+                else:
+                    full_va_name = f"{va_name}Temperature"
+                va = getattr(self, full_va_name)
+                va._set_value(temp, force_write=True)
+                logging.debug(u"Lakeshore temperature: %f °C", va.value)
```
pieleric, line 498: You should extend the message with the sensor (or va) name. Otherwise we're going to get multiple identical log messages for different sensors.

src/odemis/driver/lakeshore.py
```diff
@@ -344,98 +411,220 @@ def HeaterSetup(self,
...
             logging.exception("Failed to read sensor temperature.")
 
 
+class SerialBusAccesser(object):
+    """
+    Manages connections to the low-level bus
+    """
+
+    def __init__(self, port: str):
+        # Ensure no one will talk to it simultaneously, and we don't talk to devices already in use
```
pieleric, line 511: That comment should go next to `ser_access`.

src/odemis/driver/lakeshore.py
```diff
@@ -344,98 +411,220 @@ def HeaterSetup(self,
...
+                ans += char
+
+            logging.debug("Received answer %s", to_str_escape(ans))
+
+            return ans.strip()
+
+
+class IPBusAccesser(object):
+    """
+    Manages connections to the low-level bus
```
pieleric, line 573: Please update this docstring and the one of SerialBusAccesser to clarify the difference between the 2 classes.

src/odemis/driver/lakeshore.py
```diff
@@ -344,98 +411,220 @@ def HeaterSetup(self,
...
+    """
+
+    def __init__(self, ip_address):
```
pieleric, line 576: Please add a docstring to specify the argument. BTW, for new code, you can use the typing annotations. Like `__init__(self, ip_address: str):`

src/odemis/driver/test/lakeshore_test.py
```diff
@@ -35,36 +35,82 @@
...
+    "name": "Lakeshore Test",
+    "role": "temperature-controller",
+    "port": "/dev/ttyUSB0/",
+    # "port": "/dev/fake",
+    "sensor_input": "B",
+    "output_channel": 2,
+}
+
+# arguments used for the creation of basic components
+CONFIG_IP_BUS_CONNECTION = {
```
pieleric, line 48: The naming "IP" suggest it's going to test the IP connection, but on the simulator it test the serial port. In addition, it test multiple temperatures. So maybe it'd make more sense to name them "CONFIG_SINGLE_TEMP` and `CONFIG_MULTI_TEMP`, and as a comment indicate it also test the IP connection ?

src/odemis/driver/test/lakeshore_test.py
```diff
@@ -35,36 +35,82 @@
...
+
+class LakeshoreBaseTest:
+    """Base class for testing different models of LakeShore temperature controller."""
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.dev.terminate()  # free up socket.
+
+    def test_temperature_poll(self):
+        return
```
pieleric, line 71: Drop that return, otherwise it's too easy.

src/odemis/driver/test/lakeshore_test.py
```diff
@@ -35,36 +35,82 @@
...
+        # Test that the temperature polls.
+        # Designed to work with the simulator - the temperature will not be exactly the same
+        logging.debug("Test stasis temperature polling")
+        temp1 = self.dev.temperature.value
+        time.sleep(3)
+        temp2 = self.dev.temperature.value
+        # self.assertNotEqual(temp1, temp2)
```
pieleric, line 78: I think we should uncomment because otherwise it doesn't test anything.
However, I can imagine it's not great, as it might randomly return the same value. So 10% of the time the test case will fail. Maybe instead, it could be a loop of 10, every second, and check that not all the readings are the same.

src/odemis/driver/test/lakeshore_test.py
```diff
@@ -35,36 +35,82 @@
...
+            self.skipTest("Simulator always pretends to be MODEL350.")
+
+        self.assertEqual(model, "MODEL335")
 
     def test_simple(self):
         """
         Test some simple functions
         """
-        return
+        # return
```
pieleric, line 113: Drop completely this line.

src/odemis/driver/lakeshore.py
```diff
@@ -163,45 +220,72 @@ def _openSerialPort(port, baudrate=57600):
...
+
+            if manufacturer != "LSCI":
+                raise IOError("Invalid device manufacturer")
+
+            compatible_devices = ["MODEL335", "MODEL350"]
+            if md not in compatible_devices:
+                raise IOError("Incompatible device model detected. "
+                              "Compatible models are %s." % (compatible_devices,))
+
+            return
```
pieleric, line 266: It's important that if the connection fails (eg, because the device is turned off, or not connected to the network), the function raises HwError(). See how it's done for the serial port case. This allows the backend to know that the device is not ready. It'll display the error to the user, who has the possibility to fix the issue, and automatically the backend will retry.

src/odemis/driver/lakeshore.py
```diff
@@ -163,45 +220,72 @@ def _openSerialPort(port, baudrate=57600):
...
+        compatible_devices = ["MODEL335", "MODEL350"]
+        if md not in compatible_devices:
+            raise IOError("Incompatible device model detected. "
+                          "Compatible models are %s." % (compatible_devices,))
+
+    def _findDevice(self, bus_pattern):
         """
         Look for a compatible device
-        ports (str): pattern for the port name
+        bus_pattern (str): pattern for the port name
```
pieleric, line 243: Update the docstring. Indicate that if it doesn't start with a "/" then it should be an IP address.

pieleric (changes requested):

src/odemis/driver/lakeshore.py
```diff
@@ -69,33 +74,23 @@ class Lakeshore(model.HwComponent):
     def __init__(self, name, role, port, sensor_input='b', output_channel=2, **kwargs):
         """
         A driver for the Lakeshore 310 temperature controller.
-
         name: (str)
         role: (str)
         port: (str) port name. Can be a pattern, in which case all the ports
           fitting the pattern will be tried.
           Use /dev/fake for a simulator
```
tepals, line 81: I think it would make sense to update this to include that this is either an IP address or the usb port, do you agree?

src/odemis/driver/lakeshore.py
```diff
@@ -108,11 +103,75 @@ def __init__(self, name, role, port, sensor_input='b', output_channel=2, **kwarg
...
-        # Vigilant attributes of the controller.
-        self.temperature = FloatVA(unit=u"°C", value=self.GetSensorTemperature(), readonly=True)
-        self.targetTemperature = FloatContinuous(value=self.GetSetpoint(), unit=u"°C", range=[-273, 50], setter=self._set_targetTemperature)
-        self.heating = IntEnumerated(value=self.GetHeaterRange(), choices={0: "Off", 1: "Low", 2: "Medium", 3: "High"},
-                                     setter=self._set_heating)
+        # set supported sensor input and output channel values
+        if md == "MODEL335":
+            supported_sensor_inputs = ('A', 'B')
+            supported_output_channels = (1, 2)
+        elif md == "MODEL350":
```
tepals, line 110: Could you add a small comment on line 107 and line 110 explaining MODEL335 and MODEL350?

src/odemis/driver/lakeshore.py
```diff
@@ -108,11 +103,75 @@ def __init__(self, name, role, port, sensor_input='b', output_channel=2, **kwarg
...
+            supported_sensor_inputs = ('A', 'B', 'C', 'D')
+            supported_output_channels = (1, 2, 3, 4)
+        else:
+            raise IOError("Incompatible device model detected. Supported models: MODEL350 and MODEL335")
+
+        # Check if there's only one sensor input and the input is defined as string
+        if isinstance(sensor_input, str):
+            sensor_input = {"": sensor_input}
+        # Check if there's one or multiple sensor input(s) and defined as dictionary
```
tepals, line 119: this comment does not completely match the line below. I would say something like: `Make sure the input is either a single string or a dictionary.`. It's also possible to remove this comment as it does not add information you do not get from reading the code.

src/odemis/driver/lakeshore.py
```diff
@@ -108,11 +103,75 @@ def __init__(self, name, role, port, sensor_input='b', output_channel=2, **kwarg
...
+            supported_output_channels = (1, 2)
+        elif md == "MODEL350":
+            supported_sensor_inputs = ('A', 'B', 'C', 'D')
+            supported_output_channels = (1, 2, 3, 4)
+        else:
+            raise IOError("Incompatible device model detected. Supported models: MODEL350 and MODEL335")
+
+        # Check if there's only one sensor input and the input is defined as string
+        if isinstance(sensor_input, str):
+            sensor_input = {"": sensor_input}
```
tepals, line 118: Currently this comment just describes what the code does, and not why the code does this. You could update it to something like: `If there's only one sensor input and the input is defined as string, convert it to a dict for compatibility with the multiple input code.`

src/odemis/driver/lakeshore.py
```diff
@@ -108,11 +103,75 @@ def __init__(self, name, role, port, sensor_input='b', output_channel=2, **kwarg
...
+
+        self._sensor_inputs = {k: v.upper() for k, v in sensor_input.items()}
+        for va_name, sensor_name in self._sensor_inputs.items():
+            if sensor_name not in supported_sensor_inputs:
+                raise ValueError(f"{sensor_name} is an invalid sensor name. "
+                                 f"Sensor input must be one of the following: {supported_sensor_inputs}")
+
+            # Vigilant attributes of the temperature.
+            va = FloatVA(unit=u"°C", value=self.GetSensorTemperature(sensor_name), readonly=True)
+            if va_name == "":
```
tepals, line 131: you could add a comment here: # in case of a single string as sensor input

src/odemis/driver/lakeshore.py
```diff
@@ -108,11 +103,75 @@ def __init__(self, name, role, port, sensor_input='b', output_channel=2, **kwarg
...
+                full_va_name = "temperature"
+            else:
+                full_va_name = f"{va_name}Temperature"
+            setattr(self, full_va_name, va)
+
+        # Check if there's only one int type output channel
+        if isinstance(output_channel, int):
+            output_channel = {"": output_channel}
+        elif not isinstance(output_channel, dict):
+            raise ValueError("Invalid output channel type. Output channel must be either str or dict")
```
tepals, line 141: Is it possible to have one int type output channel and multiple input sensors?

src/odemis/driver/lakeshore.py
```diff
@@ -344,98 +403,234 @@ def HeaterSetup(self,
...
+        self._socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        self._socket.settimeout(3)  # set 3 seconds timeout to receive data
+        try:
+            self._socket.connect((ip_address, MODEL_350_TCP_PORT))
+        except socket.timeout:
+            raise model.HwError("Connection is timed out.")
+
+        self._ser_access = threading.Lock()
+
+        # self._socket.recv.timeout(3)  # set 3 seconds timeout to receive data
```
tepals, line 585: why is this commented?

src/odemis/driver/lakeshore.py
```diff
@@ -344,98 +403,234 @@ def HeaterSetup(self,
...
+
+    def sendQuery(self, cmd):
+        """
+        cmd (byte str): command to be sent to device (without the LF, but with the ?)
+        returns (byte str): answer received from the device (without \n or \r)
+        raise:
+            IOError if no answer is returned in time
+        """
+
+        # self._socket.listen()
```
tepals, line 608: why is this commented?

src/odemis/driver/test/lakeshore_test.py
```diff
@@ -92,30 +139,106 @@ def test_simple(self):
...
+        cls.dev = lakeshore.Lakeshore(**CONFIG_MULTI_TEMP)
+
+    def test_device_model(self):
+        """
+        Test device model
+        """
+        manufacturer, model, _, _ = self.dev.GetIdentifier()
+
+        if TEST_NOHW:
+            self.skipTest("Simulator always pretends to be MODEL350.")
```
tepals, line 181: this is the test for model 350, so it should work right?

src/odemis/driver/test/lakeshore_test.py
```diff
@@ -34,37 +34,84 @@
...
+
+        self.assertEqual(model, "MODEL335")
+
+    def test_temperature_poll(self):
+        # Test that the temperature polls.
+        # Designed to work with the simulator - the temperature will not be exactly the same
+        logging.debug("Test stasis temperature polling")
```
tepals, line 102: typo: statis -> static

tepals (changes requested):

src/odemis/driver/lakeshore.py
pieleric: So far we have been very liberal in what the user specifies. Temperature readings and control are completely independent, like the hardware allows. It would make little sense to have control on a temperature we cannot read ( it not the other way around), so it could be good to add a check. I also wonder how it works with hardware... need to confirm with Wilco nothing else has to be configured.

src/odemis/driver/test/lakeshore_test.py
```diff
@@ -34,37 +34,84 @@
...
+        temp1 = []
+        temp2 = []
+        for testCtr in range(10):
```
pieleric, line 105: In such simple loop, especially where the variable is not used, the convention is to name the variable `i`.

src/odemis/driver/test/lakeshore_test.py
pieleric: To be fair, this is actually old code that was just duplicated. I don't know if "statis" was a typo originally, or it's just a fancy English term (it's really a word!). I don't think that "static temperature polling" means anything either. I'd suggest the following:
* rename the test name to `test_temperature_update`.
* Change the first comment to a docstring that says "Tests the the temperature VA is automatically updated".
* Drop that logging message

src/odemis/driver/lakeshore.py
```diff
@@ -507,16 +503,16 @@ def _poll(self):
...
     """
 
     def __init__(self, port: str):
-        # Ensure no one will talk to it simultaneously, and we don't talk to devices already in use
         """
         port: the full path to the serial port (eg: /dev/ttyUSB1)
         raises:
            IOError: if something went wrong opening the port
         """
+        # Ensure no one will talk to it simultaneously, and we don't talk to devices already in use
```
pieleric, line 512: That comment is a little bit incorrect or at least prone to misinterpretation. This lock just ensures that this odemis driver will not send multiple command simultaneously. It doest NOT prevent other programs from trying to talk to the device (that's what the file lock does).

pieleric (changes requested):

src/odemis/driver/lakeshore.py
FCanberk: That line of code moved a couple of lines above, i'll remove it completely.

src/odemis/driver/lakeshore.py
FCanberk: Yes, that update would make it clearer.

src/odemis/driver/lakeshore.py
FCanberk: Forgot to remove that, i'll remove it completely.

tepals (approved):

src/odemis/driver/lakeshore.py
```diff
@@ -163,45 +222,59 @@ def _openSerialPort(port, baudrate=57600):
...
+        # Connection via ethernet cable
+        if not bus_pattern.startswith("/"):
+            try:
+                self._accesser = IPBusAccesser(bus_pattern)
+            except (IOError, LakeshoreError) as e:
+                logging.info("Could not establish connection to the device through IP bus accessor: %s", e)
+
+            self._checkDeviceCompatibility()
+
+            return
```
pieleric, line 263: It should return the "name of the port" used. => `return bus_pattern`.

src/odemis/driver/lakeshore.py
```diff
@@ -344,98 +405,228 @@ def HeaterSetup(self,
...
+        with self._ser_access:
+            logging.debug("Sending command %s", to_str_escape(cmd))
+            self._socket.send(cmd)
+
+            ans = b''
+            while ans[-1:] != b'\n':
+                try:
+                    data = self._socket.recv(TCP_BUFFER_SIZE)
+                except socket.timeout:
+                    raise model.HwError("Connection is timed out.")
```
pieleric, line 616: Here, I'd suggest to just raise a IOError. HwError is just for certain class of recoverable errors.

pieleric (changes requested):

pieleric (approved):

PR #2234, opened 2022-09-30
driver orsay: rename beamShift -> shift

pieleric:
We had already a VA called .shift for the same meaning (eg, in phenom,
and simsem). Later on, came xt_client which introduced beamShift, and we
picked that one up.
However, as it's on the scanner component, typically "shift" is clear
enough. It's also the oldest name. So let's use that one.
That should ensure that all client code can work on any hardware.

(Still need to rename the xt_client VA)

### DIFF
--- a/src/odemis/driver/orsay.py
+++ b/src/odemis/driver/orsay.py
@@ -2149,7 +2149,7 @@ class FIBBeam(model.HwComponent):
         + mirrorImage: BooleanVA, True to mirror the retrieved image
         + imageFromSteerers: BooleanVA, True to image from Steerers, False to image from Octopoles
         + objectiveVoltage: FloatContinuous, unit="V", range=(0.0, 2e4)
-        + beamShift: TupleContinuous Float, unit=m, range=[(-1.0e-4, -1.0e-4), (1.0e-4, 1.0e-4)]
+        + shift: TupleContinuous Float, unit="m", range=[(-1.0e-4, -1.0e-4), (1.0e-4, 1.0e-4)]
         + horizontalFov: FloatContinuous, unit="m", range=(0.0, 1.0)
         + measuringCurrent: BooleanVA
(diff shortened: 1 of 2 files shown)


src/odemis/driver/orsay.py
```diff
@@ -2149,7 +2149,7 @@ def __init__(self, name, role, parent, **kwargs):
         + mirrorImage: BooleanVA, True to mirror the retrieved image
         + imageFromSteerers: BooleanVA, True to image from Steerers, False to image from Octopoles
         + objectiveVoltage: FloatContinuous, unit="V", range=(0.0, 2e4)
-        + beamShift: TupleContinuous Float, unit=m, range=[(-1.0e-4, -1.0e-4), (1.0e-4, 1.0e-4)]
+        + shift: TupleContinuous Float, unit=m, range=[(-1.0e-4, -1.0e-4), (1.0e-4, 1.0e-4)]
```
tepals, line 2152: very minor comment: just to keep it consistent with the row above and below, could you change the unit to "m"?

src/odemis/driver/orsay.py
```diff
@@ -2357,7 +2357,7 @@ def on_connect(self):
         self._objectiveVoltageConnector = OrsayParameterConnector(self.objectiveVoltage, self._hvps.ObjectiveVoltage,
                                                                   minpar=self._hvps.ObjectiveVoltage_Minvalue,
                                                                   maxpar=self._hvps.ObjectiveVoltage_Maxvalue)
-        self._beamShiftConnector = OrsayParameterConnector(self.beamShift, [self._ionColumn.ObjectiveShiftX,
+        self._shiftConnector = OrsayParameterConnector(self.shift, [self._ionColumn.ObjectiveShiftX,
                                                                             self._ionColumn.ObjectiveShiftY],
```
tepals, line 2361: now that you've shortened the name I think it's a bit nicer to re-align this line with the one above

PR #2236, opened 2022-10-10
Move the  shift test to shift test directory

K4rishma:
acq/drift/test/shift_test.py was initially placed here because it uses the same images as the images used in the drift test cases. Move shift_test.py to acq/align/test and make sure to update the path to the test images. (don’t forget to run the test cases and make sure they still pass)

Testing :- passes 12/12 test cases after refactoring


pieleric (changes requested):
The content of the changes look fine, however the history is confusing.
Please make sure to rebase your branch against upstream/master before submitting a PR.

K4rishma:
@pieleric  this PR automatically closed when there was an attempt to re-synchronize the private repository with upstream repository by pushing with zero commits.

PR #2239, opened 2022-10-12
driver tescan: improve move handling

pieleric:
Safer relative move by computing the target position based on the
position when starting the move, not when requesting the move (which
could be a lot before if another move is running).

After asking to move, first wait a bit, to ensure it doesn't think
the move already finished.

Not sure if any of these will help the fact that sometimes an
(absolute) move request has no effect.

### DIFF
--- a/src/odemis/driver/test/tescan_test.py
+++ b/src/odemis/driver/test/tescan_test.py
@@ -19,9 +19,11 @@ You should have received a copy of the GNU General Public License along with
 Odemis. If not, see http://www.gnu.org/licenses/.
 """
 import Pyro4
+from concurrent.futures import CancelledError
 import copy
 import logging
 from odemis import model
+from odemis.dataio import hdf5
 from odemis.driver import tescan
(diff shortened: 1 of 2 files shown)


src/odemis/driver/test/tescan_test.py
```diff
@@ -240,7 +241,104 @@ def test_external(self):
...
+        testing.assert_pos_almost_equal(self.stage.position.value, subpos)
+
+        self.stage.moveAbsSync({"x": p["x"]})
+        testing.assert_pos_almost_equal(self.stage.position.value, p)
+        time.sleep(6)
+        testing.assert_pos_almost_equal(self.stage.position.value, p)
+
+    def test_stop(self):
+        """
+        Check it's possible to move the stage
```
sneepstefan, line 304: Maybe update the docstring a little bit with some info about the testcode for stopping while moving the stage

FCanberk (approved):

PR #2242, opened 2022-10-17
Fix pytest log filter

tepals:
In commit 25fe610fbf532ab5097f91e9d65ec5298a7bbfdf the log filter was updated to include the short test summary. It was only tested on ubuntu 18.04, where the 'short test summary' is at the beginning of the log. In Ubuntu 20.04 the 'short test summary' is at the end, therefore everything was filtered out except for the 'short test summary'. Additionally this commit makes sure that when the message is too long it is cut up into 20 complete lines, instead of the first 1000 characters of the log.


pieleric (approved):

BasHolweg (approved):

util/pytest_log_filter.py
```diff
@@ -81,15 +81,20 @@ def filter_test_log(log_txt):
...
-                           "\n \t--Logging message is too long for the default summary report, the middle part is not displayed -- \n\n"
+            test_results = test_results.split("\n")
+            # Cut out the middle part of the log and leave the first and last 20 lines.
+            test_results = "\n".join(test_results[:20]) + \
+                           "\n\n \t--The middle part of the logging message is not displayed," \
+                           " because it is too long for the default summary report. -- \n\n" + \
+                           "\n".join(test_results[-20:]) + \
+                           "\n \t--Logging message is too long for the default summary report, " \
+                           "the middle part is not displayed -- \n\n"
         return test_results
```
pieleric, line 99: It'd help have an explicit `else: return None` here. I hadn't noticed beforehand that the function can return None!

pieleric (changes requested):

pieleric (approved):

PR #2243, opened 2022-10-17
Add and verify test cases in drift_test.py and leech_test.py

K4rishma:
Added test functions in the AnchoredEstimator for the following functions

1. estimateAcquisitionTime
2. acquire
3. estimate
4. _updateSEMSettings

**Simultaneously I am working on the following test cases which will be pushed later when completed.**

For AnchorDriftCorrector the function estimateAcquisitionTime is not tested. The functions with a * are used in the test case test_get_next_pixels, you have to verify whether they are only used in this test case or if their output is tested.

estimateAcquisitionTime
series_start*
series_complete*
start*
complete*

### DIFF
--- a/src/odemis/acq/drift/test/drift_test.py
+++ b/src/odemis/acq/drift/test/drift_test.py
@@ -17,8 +17,9 @@ You should have received a copy of the GNU General Public License along with Ode
 import itertools
 import logging
 import numpy
-from odemis.acq.drift import AnchoredEstimator, GuessAnchorRegion
+from odemis.acq.drift import AnchoredEstimator, GuessAnchorRegion, MIN_RESOLUTION, MAX_PIXELS
 from odemis.dataio import hdf5
+from odemis.driver import simsem
 import os
 import unittest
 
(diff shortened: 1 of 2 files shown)


K4rishma:
@tepals 
1.  I added the test function estimateAcquisitionTime() in leech_test.py and it works. Before I did not define one constant leading to the failure of the test. Now it works as intended. 
2. For verifying the output of other functions called in test_get_next_pixels() in leech_test.py, the output of only AnchorDriftCorrector.start() is checked. For series_start, complete and series_complete, the outputs are not verified.

src/odemis/acq/drift/test/drift_test.py
```diff
@@ -27,14 +29,86 @@
...
+
+        # cls.roi = model.TupleContinuous(UNDEFINED_ROI,
+        #                                  range=((0, 0, 0, 0), (1, 1, 1, 1)),
+        #                                  cls=(int, long, float))
+        #                                  # setter=self._setROI)
+        # cls.dwellTime = model.FloatContinuous(cls.scanner.dwellTime.range[0],
+        #                                        range=cls.scanner.dwellTime.range, unit="s")
+
+        cls.region = (0, 0, 0.1, 0.1)
+        cls.dwellTime = 5e-6
```
tepals, line 63: The `setUpclass` is only executed at the start of running the entire test class. You cannot be sure if the test cases are always executed in the same order, therefore if `self.region` or `self.dwellTime` is changed anywhere in a test case you might not have the same input when you rerun all the test cases. So I would suggest to set theses values in the test cases it self, or add a `setUp` and put it there. See for instance `odemis.driver.test.scanner_test.TestDetectorHw.setUp`

src/odemis/acq/drift/test/drift_test.py
```diff
@@ -27,14 +29,82 @@
...
+        # scanner FoV: half the ROI resolution (at scale 1).
+        margin = ((ac._res[0] * ac._scale[0]) // 2,
+                  (ac._res[1] * ac._scale[1]) // 2)
+        trans_rng = self.scanner.translation.range  # pixels
+        self._trans_range = ((trans_rng[0][0] + margin[0], trans_rng[0][1] + margin[1]),
+                             (trans_rng[1][0] - margin[0], trans_rng[1][1] - margin[1]))
+        translation = (self._trans_range[1][0] - self._trans_range[0][0],
+                       self._trans_range[1][1] - self._trans_range[0][1])
+        # assuming the x y origin is in top left corner of the image
+        # if not this case will produce junk values
```
tepals, line 83: what do you mean with this comment?

src/odemis/acq/drift/test/drift_test.py
```diff
@@ -27,14 +29,82 @@
...
+
     def test_estimateAcquisitionTime(self):
-        # TODO
-        pass
+        # min amount of acquired time should be more than 0
+        self.region = (0, 0, 0.1, 0.1)
+        self.dwellTime = 5e-6
+        ac = AnchoredEstimator(self.scanner, self.detector, self.region, self.dwellTime)
+        calculate_estimateAcquisitionTime = ac.estimateAcquisitionTime()
+        self.assertGreater(calculate_estimateAcquisitionTime, 0)
```
tepals, line 107: would it be possible to create a bit more detailed test. Could you for instance determine the amount of pixels based on the region, then multiply that with the dwell time, to get an estimate?

tepals (changes requested):
Can you add docstrings to your test cases, explaining in one sentence what is being tested?

src/odemis/acq/drift/test/drift_test.py
K4rishma: I was not sure how the coordinate axes are arranged. Now I know the digital imaging convention is followed i.e y axis increases downwards. I will erase this comments.

src/odemis/acq/drift/test/drift_test.py
```diff
@@ -27,14 +29,82 @@
...
+    def test_updateSEMSettings(self):
+        # the arbitrary update value should be registered and updated when the update function is called
+        self.region = (0, 0, 0.1, 0.1)
+        self.dwellTime = 5e-6
+        ac = AnchoredEstimator(self.scanner, self.detector, self.region, self.dwellTime)
+        ac._scale = (1, 1)
+        ac._res = (1022, 1022)
+        ac._dwell_time = 2.5e-6
+        ac._updateSEMSettings()
+        # always in this order # TODO check the order of update settings
```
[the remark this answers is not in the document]
K4rishma: In the function _updateSEMSettings, it is mentioned as a comment to update the values in this particular order. So I thought it would be a nice test to add. I am not sure why this particular order.

tepals:
> For verifying the output of other functions called in test_get_next_pixels() in leech_test.py, the output of only AnchorDriftCorrector.start() is checked. For series_start, complete and series_complete, the outputs are not verified.

Can you add a todo, that these still need to be tested?

src/odemis/acq/drift/test/drift_test.py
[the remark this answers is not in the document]
tepals: Can you update the comment to the following?
```
 # Check in the order defined in _updateSEMSettings
```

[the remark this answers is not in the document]
tepals: And I think you can remove the TODO

src/odemis/acq/drift/test/drift_test.py
```diff
@@ -17,24 +17,107 @@
 import itertools
 import logging
 import numpy
-from odemis.acq.drift import AnchoredEstimator, GuessAnchorRegion
 from odemis.dataio import hdf5
+from odemis.acq.drift import AnchoredEstimator, GuessAnchorRegion, MIN_RESOLUTION, MAX_PIXELS
+from odemis.driver import simsem
```
tepals, line 22: can you update this to alphabetical order?
```
from odemis.acq.drift import AnchoredEstimator, GuessAnchorRegion, MIN_RESOLUTION, MAX_PIXELS
from odemis.dataio import hdf5
from odemis.driver import simsem
```

tepals (changes requested):

tepals (approved):

PR #2244, opened 2022-10-19
Smaract driver parameter file update and testcases

sneepstefan:
The default behaviour of the SmarAct is to search for a reference mark in the positive direction, which is down (towards the sample). If the lens is in focus, this will reach very near the sample (there is a physical limit switch right before). It’s dangerous for the hardware.

=> change the referencing to start with negative direction.

Similarly to the tmcm.TMCLController, accept a (new) parameter param_file, which takes a filename as value. This file contains a list of parameters, their channel and their value.

The file extension should be .mcs2.tsv.  Similar format as for  the TMCM. Preferably, the values can be either decimal or hexadecimal (starts with 0x).

This file is read at init, and all the values are immediately set on the device. If setting some values fails, the initialisation of the driver should fail with ValueError.

Extend the code in driver/smaract.py/MCS2 . Copy as much as possible from tmcm.py/TMCLController. (see parse_tsv_config et al.)

Add a small .mcs2.tsv file to the test cases, and check they are properly applied (by adding it to the param_file option in the test cases).

### DIFF
--- a/src/odemis/driver/smaract.py
+++ b/src/odemis/driver/smaract.py
@@ -2863,17 +2863,20 @@ class MCS2(model.Actuator):
 
         logging.debug("Using SA_CTL library version %s to connect to %s", self._swVersion, self._hwVersion)
 
-        # Read param_file
+        # Read param_file and apply it
         if param_file:
             try:
                 filestream = open(param_file)
             except Exception as ex:
                 raise ValueError("Failed to open file %s: %s" % (param_file, ex))
             try:
+                # try to parse read the file line-by-line
                 prop_params = self.parse_tsv_config(filestream)
+                logging.debug("Extracted param file config: %s" % prop_params)
             except Exception as ex:
                 raise ValueError("Failed to parse file %s: %s" % (param_file, ex))
-            logging.debug("Extracted param file config: %s", prop_params)
+
+            # apply the new property values
             self.apply_config(prop_params)
 
         # set specific axis properties
@@ -2951,32 +2954,37 @@ class MCS2(model.Actuator):
     def apply_config(self, prop_params):
         """
         Configure the device according to the given 'user configuration' from a specified param_file.
-        :param prop_params (dict (int, int) -> (str, int)): channel index, property code -> type, value
+        :param prop_params (dict (int, int, str) -> int/str): channel index, property code, type -> value
         """
 
         # run the appropriate method for every dict item depending on value type
-        for (ch, p), (typ, v) in prop_params.items():
+        for (ch, p, typ), v in prop_params.items():
             if typ == "I32":
                 self.SetProperty_i32(p, ch, v)
             elif typ == "I64":
                 self.SetProperty_i64(p, ch, v)
             elif typ == "Str":
                 self.SetProperty_s(p, ch, v)
+            else:
+                raise ValueError(f"Property type not handled: {typ}, valid types are I32/I64/Str")
+
+        logging.debug("parameter file device configuration property changes applied.")
 
     @staticmethod
     def parse_tsv_config(filestream):
         """
         Parse a tab-separated value (.mcs2.tsv) file in the following format:
-            (file header) channel_index     property    value   # description
+            (file header) channel_index     property    value   type   # description
             channel_index can be a (Ch) number starting from 0 this also refers to a defined channel type
             property is a hexadecimal code representing a specific property
             value is a hexadecimal code representing independent property ﬂags
+            type is the type of the value used for this property (I32/I64/Str)
         filestream (File): opened filestream
         return:
-            prop_params_i32 (dict (int, int) -> (str, int)): channel index, property code -> type, value
+            prop_params_i32 (dict (int, int, str) -> int/str): channel index, property code, type -> value
         """
 
-        # MCS2 property types dedicated dicts:
+        # MCS2 property types dictionary:
         prop_params = {}
 
         # read the parameters "database" the file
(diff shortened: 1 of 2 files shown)


src/odemis/driver/smaract.py
```diff
@@ -2930,6 +2950,75 @@ def updateMetadata(self, md):
...
+                ch_num, prop_code, set_val, typ = int(m.group("num")), \
+                                                  int(m.group("code"), 0), \
+                                                  str(m.group("value")), \
+                                                  m.group("type")
+            else:
+                ch_num, prop_code, set_val, typ = int(m.group("num")), \
+                                                  int(m.group("code"), 0), \
+                                                  int(m.group("value"), 0), \
+                                                  m.group("type")
+
```
pieleric, line 3016: I'd suggest to only put the value conversion in the if. Something like that:
```python
ch_num = int(m.group("num"))
prop_code = int(m.group("code"), 0)  # detect the base from the content
typ = m.group("type")
if typ == "Str":
    set_val = m.group("value")
else: # I32 or I64
    set_val = int(m.group("value"), 0)
```

src/odemis/driver/test/smaract_test.py
```diff
@@ -649,6 +650,62 @@ def test_auto_update_function(self):
...
 
+    def test_param_file_nofile(self):
+        # check if forcing a ValueError is the result of a file that does not exist
+        with self.assertRaises(ValueError):
+            CONFIG_3DOF['param_file'] = "nonexistingfile.mcs2.tsv"
+            smaract.MCS2(**CONFIG_3DOF)
+
+    def test_parse_tsv_config(self):
+        # create a temporary .mcs2.tsv file
+        file_dir = "/tmp/testfile.mcs2.tsv"
```
pieleric, line 661: Please use `tempfile.mkstemp(suffix=...)`

src/odemis/driver/test/smaract_test.py
```diff
@@ -649,6 +650,62 @@ def test_auto_update_function(self):
...
+
+        CONFIG_3DOF['param_file'] = file_dir
+        dev = smaract.MCS2(**CONFIG_3DOF)
+
+        # open the target file again, this time for reading only
+        testfile = open(file_dir, 'r')
+        out_params = dev.parse_tsv_config(testfile)
+
+        # check if the output Tuple contains the right property input values
+        for index, item in enumerate(out_params.items()):
```
pieleric, line 695: I think you're lucky because the dict happens to be in the same order as it was created, but you shouldn't rely on that.
I'd suggest to iterate over param_dict, and check that `(num, code, type)` is in `out_params`, and then that the value is equal.

pieleric (changes requested):

src/odemis/driver/test/smaract_test.py
```diff
@@ -649,6 +652,56 @@ def test_auto_update_function(self):
...
+        self.assertEqual(len(out_params), len(param_dict))
+
+        # check if the output dict contains the right property input values
+        for item in param_dict:
+            key = (item['num'], item['code'], item['type'])
+            if key in out_params:
+                self.assertEqual(item['value'], out_params[key])
+
+        testfile.close()
+        os.unlink(file_dir)
```
pieleric, line 702: Please just use `os.remove()`, as it's easier to understand what it does. Or does it do something different?

src/odemis/driver/test/smaract_test.py
```diff
@@ -649,6 +652,56 @@ def test_auto_update_function(self):
...
 
+    def test_param_file_nofile(self):
+        # check if forcing a ValueError is the result of a file that does not exist
+        with self.assertRaises(ValueError):
+            CONFIG_3DOF['param_file'] = "nonexistingfile.mcs2.tsv"
+            smaract.MCS2(**CONFIG_3DOF)
+
+    def test_parse_tsv_config(self):
+        # create a temporary .mcs2.tsv file
+        file_id, file_dir = tempfile.mkstemp(suffix='mcs2.tsv', text=True)
```
pieleric, line 663: `file_dir` sounds like the "file directory", but it's actually the completely file path (including the file name). So let's call it differently. Maybe `file_path` or `file_name`?

pieleric (changes requested):

src/odemis/driver/test/smaract_test.py
sneepstefan: I think it does not do anything different.. unlink was more of a Linux way to remove files. But both can be used interchangeably.

pieleric (approved):

src/odemis/driver/smaract.py
```diff
@@ -2930,6 +2950,76 @@ def updateMetadata(self, md):
...
+            if not m:
+                # read line is not in the correct format
+                raise ValueError("Failed to parse line, unexpected line format '%s'" % readline.rstrip("\n\r"))
+
+            # assign the readline values to the appropriate variables
+            ch_num = int(m.group("num"))
+            prop_code = int(m.group("code"), 0)
+            typ = m.group("type")
+
```
BasHolweg, line 3012: In this case when m is None the code will still try m.group which should not be possible since m is None.

BasHolweg (changes requested):

src/odemis/driver/smaract.py
sneepstefan: Good point, but these line are positioned above this block:

            if not m:
                # read line is not in the correct format
                raise ValueError("Failed to parse line, unexpected line format '%s'" % readline.rstrip("\n\r"))

Therefore, when m is None it is handled before assigning m.group to these variables.

tepals (approved):

BasHolweg (approved):

PR #2247, opened 2022-10-20
Addition of testcases for img.py for all methods which did not yet have testcases

sneepstefan:
The following functions in odemis.util.img.py do not have a test case yet. Try to think of a simple test case for each one of them and add them to odemis/util/test/test_image.py. Note that there is also the module odemis.gui.util.img.py , this task is only about odemis.util.img.py.

tint_to_md_format
md_format_to_tint
getOutliers
guessDRange
getColorbar
tintToColormap
getYXFromZYX
ensure2DImage
Subtract
mergeMetadata
getTilesSize
getCenterOfTiles
getBoundingBox
assembleZCube

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,3 +76,7 @@ magic.odm.yaml
 /src/odemis/driver/jeol.py
 /src/odemis/driver/nikonc.py
 /.mypy_cache/
+
+# Custom ignores
+src/odemis/dataio/test/htmlcov/
+install/linux/usr/share/odemis/sim/meteor-reverse-sim.odm.yaml
--- a/.gitignore
+++ b/.gitignore
@@ -76,7 +76,3 @@ magic.odm.yaml
 /src/odemis/driver/jeol.py
 /src/odemis/driver/nikonc.py
 /.mypy_cache/
-
-# Custom ignores
-src/odemis/dataio/test/htmlcov/
-install/linux/usr/share/odemis/sim/meteor-reverse-sim.odm.yaml
--- a/src/odemis/util/test/img_test.py
+++ b/src/odemis/util/test/img_test.py
@@ -1255,7 +1255,7 @@ class TestMdFormatToTint(unittest.TestCase):
         tint_output = img.md_format_to_tint(TINT_FIT_TO_RGB)
         self.assertTrue(isinstance(tint_output, str))
 
-    def test_md_format_to_tint_unknown_str(self):
+    def test_md_format_to_tint_unexpected_str(self):
         """
         check if passing an unknown str user_tint type in saved
         metadata will lead to raising the correct exception
@@ -1277,7 +1277,7 @@ class TestMdFormatToTint(unittest.TestCase):
         tint_output = img.md_format_to_tint('autumn')
         self.assertTrue(isinstance(tint_output, colors.Colormap))
 
-    def test_md_format_to_tint_unknown_type(self):
+    def test_md_format_to_tint_unexpected_type(self):
         """
         check if passing an unknown user_tint type in saved
         metadata will lead to raising the correct exception
@@ -1310,12 +1310,10 @@ class TestGetOutliers(unittest.TestCase):
         # keep the value for outliers default
         output_range = img.getOutliers(data)
 
-        # set expected output values based on low_range high_range input
-        calc_output1 = 0.0
-        calc_output2 = high_range - 1.0
-
-        # assess the returned 2-valued tuple min/max range values
-        self.assertEqual(output_range[0] == calc_output1, output_range[1] == calc_output2)
+        # assess the returned obj for tuple instance
+        self.assertIsInstance(output_range, tuple)
+        # check if the returned tuple length equals 2
+        self.assertEqual(len(output_range), len((1.0, 1.0)))
 
     def test_getOutliers_discard_all(self):
         """
@@ -1335,20 +1333,17 @@ class TestGetOutliers(unittest.TestCase):
 
         output_range = img.getOutliers(data, outlier_num)
 
-        # set expected output values based on low_range high_range
-        # input and take the outlier value into account
-        calc_output1 = high_range * outlier_num
-        calc_output2 = float(high_range) - calc_output1
-
-        # assess the returned 2-valued tuple min/max range values
-        self.assertEqual(output_range[0] == calc_output1, output_range[1] == calc_output2)
+        # assess the returned obj for tuple instance
+        self.assertIsInstance(output_range, tuple)
+        # check if the returned tuple length equals 2
+        self.assertEqual(len(output_range), len((1.0, 1.0)))
 
 
 class TestGuessDRange(unittest.TestCase):
 
-    def test_guessDRange_zero_BPP(self):
+    def test_guess_DRange_zero_BPP(self):
         """
-        Not exactly sure what has to be tested here
+        Test if value of 0 for BPP returns the default min/max for uint8
         """
 
         size = (2000, 1000)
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -1,78 +0,0 @@
-# based on https://github.com/github/gitignore/blob/master/Python.gitignore
```
tepals, line 1: something went wrong here, you deleted the entire gitignore

sneepstefan: I was under the impression that the .gitignore was from transferred from my branch.. because we were getting rid of that yesterday in my branch and master.

sneepstefan: Its fixed now and .gitignore is now recovered again. Thanks alot @tepals !

FCanberk (changes requested):

FCanberk (approved):

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,873 @@ def test_rotate_img_metadata(self):
...
+        # create the other 3 tiles by swapping axis of tile1
+        # 4 tiles together make diamond shape pattern
+        tile2 = numpy.flip(tile1, axis=1)
+        tile3 = numpy.flip(tile1, axis=0)
+        tile4 = numpy.flip(tile3, axis=1)
+
+        # create a list of 4 tiles
+        tile_list = (tile1, tile2, tile3, tile4)
+        # create a list of 4 tile lists
+        tiles_list = (tile_list, tile_list, tile_list, tile_list)
```
tepals, line 1750: technically this is a tuple and not a list

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,873 @@ def test_rotate_img_metadata(self):
...
+        # reverse the order of the img_list
+        for index, num in enumerate(reversed(range(len(img_list)))):
+            rev_list[index] = img_list[num]
+
+        output_DA_after = img.assembleZCube(rev_list, z_list)
+
+        # check if the Z dimension has the right size
+        self.assertEqual(output_DA_before.shape[0], len(z_list) and output_DA_after.shape[0], len(z_list))
+        # check if the returned DA before order reverse is different than after order reverse
+        numpy.testing.assert_equal(numpy.any(numpy.not_equal(output_DA_before, output_DA_after)), True)
```
tepals, line 2050: I think I would go for:
```
self.assertFalse(numpy.all(output_DA_before == output_DA_after))
```

tepals (changes requested):
General comments:
At the beginning of most test classes you have some comments, that look more like notes to yourself instead of actual comments. If that's true, can you remove them? If you think they should stay, can you turn them into docstrings?

Not a hard rule, but in general we don't have an empty line inbetween the docsrting and the start of the code in Odemis.

You have calculations in quite a few of the assert statements, see line 1423. I would try to do this as little as possible.

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,873 @@ def test_rotate_img_metadata(self):
...
+    def test_assemble_zcube_normal(self):
+        """
+        Verify that using a 2D shaped image DA's in a list will return the right ZCube assembly.
+        normal use of function
+        """
+        img_list = []
+        z_list = [0.1, 0.2, 0.3, 0.4, 0.5, 0.5, 0.4, 0.3, 0.2, 0.1]
+
+        # create a list of DA
+        for x in range(10):
```
BasHolweg, line 1995: More dynamic if you use len(z_list).

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,873 @@ def test_rotate_img_metadata(self):
...
+            img_list.append(model.DataArray(numpy.random.random_integers(0, 220, self.size), self.md))
+
+        output_DA = img.assembleZCube(img_list, z_list)
+
+        # check if the imput DA shapes are different than the output DA
```
BasHolweg, line 2000: \# check if the i**n**put DA shapes are different than the output DA

BasHolweg (changes requested):

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,873 @@ def test_rotate_img_metadata(self):
...
+
+        # NumPy array of uint8 RGB tuples is returned
+        output_array = img.getColorbar(input_cm, input_height, input_width, input_alpha)
+
+        # test for tuple shape of 3
+        self.assertEqual(len(output_array.shape), 3)
+        # test for third element in tuple regarding additional alpha channel
+        self.assertEqual(output_array.shape[2], 3 + int(input_alpha))
+        # test for RGB uint8 dtype
+        self.assertTrue(numpy.all(output_array >= 0) and numpy.all(output_array <= 255))
```
pieleric, line 1454: Please, just explicitly check the output_array.dtype == numpy.uint8.

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,873 @@ def test_rotate_img_metadata(self):
...
+        input_cm = cm.get_cmap('cividis')
+        input_height = 16
+        input_width = 72
+        input_alpha = True
+
+        # NumPy array of uint8 RGB tuples is returned
+        output_array = img.getColorbar(input_cm, input_height, input_width, input_alpha)
+
+        # test for tuple shape of 3
+        self.assertEqual(len(output_array.shape), 3)
```
pieleric, line 1450: You have much strong expectations, lets check for them: shape == (input_height, input_width, 4)

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,873 @@ def test_rotate_img_metadata(self):
...
+        # test for third element in tuple regarding additional alpha channel
+        self.assertEqual(output_array.shape[2], 3 + int(input_alpha))
```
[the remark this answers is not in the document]
pieleric: I wouldn't recommend writing test cases in such "generic way". In this case, input_alpha is True, so just check that the shape[2] == 4. If you have another test where the input_alpha is False, explicitly check it's 3.

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,873 @@ def test_rotate_img_metadata(self):
...
+        with self.assertRaises(ValueError):
+            img.tintToColormap(input_tint)
+
+
+class TestGetYXFromZYX(unittest.TestCase):
+    # input 3D data array as picture index int for z-axis start
+    # output 2D data array
+    # test for single dimension
+    # test forcing TypeError/ValueError
+    # test for 2D pos
```
pieleric, line 1521: You can drop that.

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,873 @@ def test_rotate_img_metadata(self):
...
+
+    def test_get_yx_from_zyx_normal(self):
+        """
+        Verify that using a 3D DA picture will return the
+        right 2D DA picture.
+        normal use of function
+        """
+        data = model.DataArray(self.arr1, metadata=self.md)
+        # take a random z-slice value
+        z_slice = 160
```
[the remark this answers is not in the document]
pieleric: "random" -> "arbitrary"?

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,873 @@ def test_rotate_img_metadata(self):
...
+        size = (200, 399, 399)
+        self.md = {
+              model.MD_DIMS: 'YXZ',
+              model.MD_PIXEL_SIZE: (1e-6, 1e-6),
+              model.MD_BPP: 8,
+              model.MD_POS: (5.0, 7.0)
+             }
+
+        # create a mock 3D image array with white color values
+        self.arr1 = numpy.ones(size, dtype='uint8') * 255
```
pieleric, line 1534: Typically, you can just write:
```python
self.arr1 = numpy.empty(size, dtype='uint8')
self.arr1[...]= 255
```

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,873 @@ def test_rotate_img_metadata(self):
...
+        x_count, y_count = 0, 0
+        self.color_num = 160
+
+        # change values of the ndarray in a pyramid shape
+        # so that every z-slice contains a different amount
+        # of gray color values
+        for z in range(self.arr1.shape[0]):
+            for y in range(y_count, self.arr1.shape[1] - y_count):
+                for x in range(x_count, self.arr1.shape[2] - x_count):
+                    self.arr1[z, y, x] = self.color_num
```
pieleric, line 1545: Typically, the last loop, you'd write as a single line, like this:
```python
                    self.arr1[z, y, x_count:self.arr1.shape[2] - x_count] = self.color_num
```
It does the same, just a lot faster. That's not a big deal in a test case, but good to know.

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,873 @@ def test_rotate_img_metadata(self):
...
+    def test_ensure_2d_image_2d(self):
+        """
+        Verify that using a 5D CTZYX DA will return the
+        right 2D DA.
+        normal use of function
+        """
+        out_arr = numpy.asarray(img.ensure2DImage(self.data))
+
+        # check if the returned array shape is 2D
+        self.assertTrue(out_arr.ndim == 2)
```
pieleric, line 1631: The next check already checks it's 2D.

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,873 @@ def test_rotate_img_metadata(self):
...
+        arr = numpy.ones(size, dtype='uint8')
+        self.data = model.DataArray(arr)
+
+    def test_ensure_2d_image_2d(self):
+        """
+        Verify that using a 5D CTZYX DA will return the
+        right 2D DA.
+        normal use of function
+        """
+        out_arr = numpy.asarray(img.ensure2DImage(self.data))
```
pieleric, line 1628: Why using `numpy.asarray()`? Shouldn't you check directly the output of the function? (actually, I'm not sure whether `asarray()` has any effect here).

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,873 @@ def test_rotate_img_metadata(self):
...
+        self.data = numpy.asarray(self.data).reshape(1, 2, 50, 400, 400)
+
+        with self.assertRaises(ValueError):
+            img.ensure2DImage(self.data)
+
+
+class TestSubtract(unittest.TestCase):
+    # input data array
+    # output 2D data array
+    # test for dtype unsigned int, float, combi
```
pieleric, line 1649: drop

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,873 @@ def test_rotate_img_metadata(self):
...
+        # add white squares in the black pictures
+        pic1[200:500, 200:500] = 255
+        pic2[300:400, 300:400] = 255
+
+        # subtract the pictures
+        out_arr = img.Subtract(pic1, pic2)
+
+        # check if the subtraction worked by finding the smaller
+        # square of pic2 in the output picture
+        self.assertEqual(out_arr.shape, pic1.shape, pic2.shape)
```
pieleric, line 1671: You cannot check that all of them are the same shape like that. The 3rd argument is the message to display in case there is an error.
Anyway, just check the shape of out_arr and pic1 is enough.

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,873 @@ def test_rotate_img_metadata(self):
...
+        Verify that using a 5D DA will fail.
+        see if a ValueError exception is raised
+        """
+        self.data = numpy.asarray(self.data).reshape(1, 2, 50, 400, 400)
+
+        with self.assertRaises(ValueError):
+            img.ensure2DImage(self.data)
+
+
+class TestSubtract(unittest.TestCase):
```
pieleric, line 1646: Please add another test case checking that it doesn't underflow. Using two uint8,  10 - 11 returns 0 (and not 254!).

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,873 @@ def test_rotate_img_metadata(self):
...
+        # no subtraction should have happened due to different dtypes
+        self.assertTrue(numpy.any(out_arr[300:400, 300:400]))
+
+
+class TestGetTilesSize(unittest.TestCase):
+    # input tiles (tuple of tuple of DataArray)
+    # output tuple (h, w) height/width
+    # test for tile list regular (2-)shape
+    # test for tile list unexpected 3-shape
+    # test for single tile only
```
pieleric, line 1717: Drop

pieleric (changes requested):

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1185,873 @@ def test_rotate_img_metadata(self):
...
+
+    def test_tint_to_md_format_str(self):
+        """
+        check if passing a known tint str will lead
+        to returning the correct metadata format
+        """
+
+        # test str tint instance type
+        tint_output = img.tint_to_md_format(TINT_FIT_TO_RGB)
+        self.assertTrue(isinstance(tint_output, str))
```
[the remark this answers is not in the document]
sneepstefan: That is a nice suggestion indeed

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,873 @@ def test_rotate_img_metadata(self):
...
+             }
+
+        # check for dtype uint8
+        arr = numpy.arange(size[0] * size[1], dtype=numpy.uint8)
+        data = model.DataArray(arr, metadata=md)
+
+        # min/max values should be uint8 min/max
+        min_max_type = numpy.iinfo(numpy.uint8)
+
+        # raises the ValueError as well
```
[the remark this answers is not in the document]
sneepstefan: Yeah that comment was sort of ill chosen, I changed it to something more applicable

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,873 @@ def test_rotate_img_metadata(self):
...
+
+        # set the positions to check for
+        pos_x = new_da.metadata[model.MD_POS][0]
+        pos_y = new_da.metadata[model.MD_POS][1]
+
+        # check if all the positions are almost equal
+        self.assertAlmostEqual(output_pos[0], pos_x, places=2)
+        self.assertAlmostEqual(output_pos[1], pos_y, places=2)
+        self.assertAlmostEqual(output_pos[2], pos_x, places=2)
+        self.assertAlmostEqual(output_pos[3], pos_y, places=2)
```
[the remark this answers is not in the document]
sneepstefan: Eventually I was not happy with the testcase and rewrote most of it. Thanks for the detailed feedback though.

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1185,827 @@ def test_rotate_img_metadata(self):
...
+        # set the positions to check for
+        pos_x = new_da.metadata[model.MD_POS][0]
+        pos_y = new_da.metadata[model.MD_POS][1]
+
+        # TODO Explain why it can be almost equal
+        # check if all the positions are almost equal
+        self.assertAlmostEqual(output_pos[0], pos_x, places=2)
+        self.assertAlmostEqual(output_pos[1], pos_y, places=2)
+        self.assertAlmostEqual(output_pos[2], pos_x, places=2)
+        self.assertAlmostEqual(output_pos[3], pos_y, places=2)
```
tepals, line 1898: this test is not checking the correct thing. The function returns a bounding box and pos_x and pos_y are the centers of the DA and thus of the bounding box. You want to check the following:

```python
expected_pos_x_min = pos_x - self.size[-1] * self.md[model.MD_PIXEL_SIZE]
self.assertEqual(output_pos[0], expected_pos_x_min)

expected_pos_x_max = pos_x + self.size[-1] * self.md[model.MD_PIXEL_SIZE]
self.assertEqual(output_pos[2], expected_pos_x_max)
```

and do the same for y. Then you probably don't need the `assertAlmostEqual`, although there might be a floating point error.

tepals (changes requested):

tepals (approved):

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,822 @@ def test_rotate_img_metadata(self):
...
+        # force an IndexError due to wrong DA shape
+        with self.assertRaises(IndexError):
+            img.getBoundingBox(new_da)
+
+
+class TestAssembleZCube(unittest.TestCase):
+
+    def setUp(self):
+        # create a 2D size and metadata with necessary items
+        self.size = (3200, 3200)
```
pieleric, line 1929: Maybe pick something a little smaller.  For the test cases, I don't think the size matters much, but as you have quite a few Z, that's going to consume quite a bit of memory.
You should also pick something not square.
=> `size = 400, 300`

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,822 @@ def test_rotate_img_metadata(self):
...
+    def test_assemble_zcube_normal(self):
+        """
+        Verify that using a 2D shaped image DA's in a list will return the right ZCube assembly.
+        """
+        img_list = []
+        z_list = [0.1, 0.2, 0.3, 0.4, 0.5, 0.5, 0.4, 0.3, 0.2, 0.1]
+
+        # create a list of DA
+        for x in range(len(z_list)):
+            img_list.append(model.DataArray(numpy.random.random_integers(0, 220, self.size), self.md))
```
pieleric, line 1945: `random_integers` is deprecated, please use `numpy.random.randint` instead. Pretty much the same usage...

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,822 @@ def test_rotate_img_metadata(self):
...
+
+        output_DA = img.assembleZCube(img_list, z_list)
+
+        # check if the input DA shapes are different compared to the output DA
+        self.assertNotEqual(img_list[0].ndim, output_DA.ndim)
+        # check if the Z dimension has the right size
+        self.assertEqual(output_DA.shape[0], len(z_list))
```
pieleric, line 1952: Just check the shape explicitly instead of checking ndim. It should be Z,Y,X, so len(z_list), size[0], size[1].

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,822 @@ def test_rotate_img_metadata(self):
...
+
+    def test_assemble_zcube_3d(self):
+        """
+        Verify that using a 3D shaped image DA's in a list will fail the
+        assembly of a ZCube and see if a ValueError exception is raised.
+        """
+        # adjust the size for 3D DA
+        self.size = (10, 700, 700)
```
pieleric, line 1960: Why use and change `self.size` while you don't want to change the behaviour of other functions?
=> just use a local `size` .

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,822 @@ def test_rotate_img_metadata(self):
...
+            model.MD_PIXEL_SIZE: (2e-06, 4e-05),
+            model.MD_POS: (0.2, 0.2)
+        }
+
+    def test_assemble_zcube_normal(self):
+        """
+        Verify that using a 2D shaped image DA's in a list will return the right ZCube assembly.
+        """
+        img_list = []
+        z_list = [0.1, 0.2, 0.3, 0.4, 0.5, 0.5, 0.4, 0.3, 0.2, 0.1]
```
pieleric, line 1941: Note that normally the z values shouldn't be duplicated. The function currently doesn't check for that, but it'd be fair if it raise an error. So please, generate a sorted list without duplicates.

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,822 @@ def test_rotate_img_metadata(self):
...
+    def test_assemble_zcube_3d(self):
+        """
+        Verify that using a 3D shaped image DA's in a list will fail the
+        assembly of a ZCube and see if a ValueError exception is raised.
+        """
+        # adjust the size for 3D DA
+        self.size = (10, 700, 700)
+
+        img_list = []
+        z_list = [0.1, 0.2, 0.3, 0.4, 0.5, 0.5, 0.4, 0.3, 0.2, 0.1]
```
pieleric, line 1963: Please use a proper z_list.

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,822 @@ def test_rotate_img_metadata(self):
...
+
+        # create a list of DA
+        for x in range(len(z_list)):
+            img_list.append(model.DataArray(numpy.random.random_integers(0, 220, self.size), self.md))
+
+        output_DA = img.assembleZCube(img_list, z_list)
+
+        # check if the input DA shapes are different compared to the output DA
```
pieleric, line 1949: Something else to check is that afterward, MD_DIMS should have a Z (and actually should be "ZYX"). In addition, MD_PIXEL_SIZE and MD_POS should be of length 3.
You could even check that the MD_PIXEL_SIZE is (2e-06, 4e-05, 0.1), and for MD_POS, the z value should be (almost equal to) the mean of z_list.

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,822 @@ def test_rotate_img_metadata(self):
...
+            img_list.append(model.DataArray(numpy.random.random_integers(0, 220, self.size), self.md))
+
+        output_da_before = img.assembleZCube(img_list, z_list)
+
+        # create a temporary reversed list based on img_list
+        rev_list = img_list.copy()
+
+        # reverse the order of the img_list
+        for index, num in enumerate(reversed(range(len(img_list)))):
+            rev_list[index] = img_list[num]
```
pieleric, line 1992: You can do all this with `rev_list = list(reversed(img_list))`

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,822 @@ def test_rotate_img_metadata(self):
...
+
+        output_da_after = img.assembleZCube(rev_list, z_list)
+
+        # check if the Z dimension has the right size
+        self.assertEqual(output_da_before.shape[0], len(z_list) and output_da_after.shape[0], len(z_list))
+        # check if the returned DA before order reverse is different after order reverse
+        self.assertFalse(numpy.all(output_da_before == output_da_after))
```
pieleric, line 1999: Use `odemis.util.testing.assert_array_not_equal()`

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,822 @@ def test_rotate_img_metadata(self):
...
+        rev_list = img_list.copy()
+
+        # reverse the order of the img_list
+        for index, num in enumerate(reversed(range(len(img_list)))):
+            rev_list[index] = img_list[num]
+
+        output_da_after = img.assembleZCube(rev_list, z_list)
+
+        # check if the Z dimension has the right size
+        self.assertEqual(output_da_before.shape[0], len(z_list) and output_da_after.shape[0], len(z_list))
```
pieleric, line 1997: I'm pretty sure it's not as you would it to happen. The second argument is `len(z_list) and output_da_after.shape[0]`, which is computed by python as `output_da_after.shape[0]`. The third argument is shown in case of failure.
=> use 2 `self.assertEqual()`

pieleric (changes requested):
I've just reviewed TestAssembleZCube

src/odemis/util/test/img_test.py
sneepstefan: Hi Eric I disagree with the check for MD_DIMS, regardless what happens MD_DIMS always gets 'ZYX' directly assigned without any constraint so why test that?

BasHolweg (approved):

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,825 @@ def test_rotate_img_metadata(self):
...
+
+        # output shape should be the same as biggest input DA
+        self.assertEqual(out_arr.shape, pic1.shape)
+        # all values in this square should return false=0
+        self.assertTrue(numpy.all(out_arr[300:400, 300:400] == 0))
+
+    def test_subtract_uint_float(self):
+        """
+        Verify that using pictures with different dtypes will return the right DA.
+        Check if using 2 different dtype arrays will still work.
```
pieleric, line 1615: Actually, `Subctract()` docinfo says  "return : a - b, with same dtype [...] as a".
So uint8 - float32 should yield an uint8, but it returns a float.

=> For now write down that this doesn't work. I'd suggest to also just have explicit check on the dtype.
```python
        # FIXME: for now this doesn't pass because Subtract() always return a float
        out_arr = img.Subtract(pic1, pic2)
        # self.assertEqual(out_arr.dtype, pic1.dtype)

        out_arr = img.Subtract(pic2, pic1)
        self.assertEqual(out_arr.dtype, pic2.dtype)
```

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,825 @@ def test_rotate_img_metadata(self):
...
+        self.assertEqual(output_pos[1], expected_pos_y_min)
+
+        # check if the max position of the bounding box is the same as half the
+        # size of the DA calculating from the middle point on the 1-axis upwards
+        expected_pos_y_max = pos_y + self.size[-2] * self.md[model.MD_PIXEL_SIZE][1] / 2
+        self.assertEqual(output_pos[3], expected_pos_y_max)
+
+    def test_get_bounding_box_incomplete_metadata(self):
+        """
+        Verify that removing PIXEL_SIZE in the metadate DA will fail and a LookupError exception is raised.
```
pieleric, line 1898: Please also check that without MD_DIMS, it also works.

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,825 @@ def test_rotate_img_metadata(self):
...
+
+        # force a LookupError
+        with self.assertRaises(LookupError):
+            img.getBoundingBox(new_da)
+
+    def test_get_bounding_box_wrong_arrayshape(self):
+        """
+        Verify that using a wrong shaped DA will fail and if an IndexError exception is raised.
+        """
+        # create a wrong shaped DA with random values
```
pieleric, line 1915: Please indicate why is 4000x3500 a "wrong shape". (It's because it only has 2 dims, while MD_DIMS reports 5 dims)

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,825 @@ def test_rotate_img_metadata(self):
...
+        # create a temporary reversed list based on img_list
+        rev_list = list(reversed(img_list))
+
+        output_da_after = img.assembleZCube(rev_list, z_list)
+
+        # check if the Z dimension has the right size
+        self.assertEqual(output_da_before.shape[0], len(z_list))
+        self.assertEqual(output_da_after.shape[0], len(z_list))
+        # check if the returned DA before order reverse is different after order reverse
+        testing.assert_array_not_equal(output_da_before, output_da_after)
```
pieleric, line 2002: Please also check that passing the z in the other order will reverse the cube in Z.
Something like:
```python
# Passing Z in the reverse order should lead to a cube reversed in Z, with a positive pixel size
output_rev_z = img.assembleZCube(img_list, z_list[::-1])
self.assertGreater(output_rev_z.metadata[model.MD_PIXEL_SIZE][2], 0)
numpy.testing.assert_array_equal(output_da_after, output_rev_z)
```

pieleric (changes requested):

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,838 @@ def test_rotate_img_metadata(self):
...
+        z_list = [0.1, 0.2, 0.3, 0.4, 0.5]
+
+        # create a list of DA
+        for x in range(len(z_list)):
+            img_list.append(model.DataArray(numpy.random.random_integers(0, 220, self.size), self.md))
+
+        output_da_before = img.assembleZCube(img_list, z_list)
+
+        # use a reversed image order to force a different ZCube assembly
+        output_da_after = img.assembleZCube(list(reversed(img_list)), z_list)
```
pieleric, line 2004:
```suggestion
        output_da_after = img.assembleZCube(img_list[::-1], z_list)
```

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,838 @@ def test_rotate_img_metadata(self):
...
+    def test_get_bounding_box_incomplete_metadata(self):
+        """
+        Verify that removing PIXEL_SIZE in the metadate DA will fail and a LookupError exception is raised.
+        """
+        # create a DA with random values
+        new_da = numpy.random.random(self.size)
+        new_da = model.DataArray(new_da, self.md)
+
+        # remove the DIMS key and check if a return value is passed and no exception is raised
+        del self.md[model.MD_DIMS]
```
pieleric, line 1910: This does the same, but I feel it'd be more explicit:
```suggestion
        del new_da.metadata[model.MD_DIMS]
```

src/odemis/util/test/img_test.py
```diff
@@ -1181,8 +1183,838 @@ def test_rotate_img_metadata(self):
...
+        """
+        Verify that using a 3D shaped image DA's in a list will return a
+        different ZCube assembly if the order of the images is changed.
+        See if the differences with a list generated before and after reordering are picked up.
+        """
+        img_list = []
+        z_list = [0.1, 0.2, 0.3, 0.4, 0.5]
+
+        # create a list of DA
+        for x in range(len(z_list)):
```
pieleric, line 1998:
```suggestion
        for z in z_list:
```

PR #2249, opened 2022-10-24
gui util img: support SpectrumStreams without spectrum

pieleric:
SpectrumStreams are (now) misnamed as they support a lot of different
acquisition type. Including acquisitions without any spectrum. For
instance time-resolved (only) data is handled by the SpectrumStream too.

This confused the exported which cared about the spectrumBandwidth VA
(to update the metadata).
=> Don't use the type of the stream, and just explicitly check that the
VA is present.

### DIFF
--- a/src/odemis/gui/util/test/img_test.py
+++ b/src/odemis/gui/util/test/img_test.py
@@ -999,6 +999,40 @@ class TestSpatialExport(test.GuiTestCase):
         for s, md in zip(streams, orig_md):
             self.assertEqual(md, s.raw[0].metadata)
 
+    def test_tr(self):
+        """
+        Test exporting spatial view of a time-resolved data
+        """
+        data = numpy.ones((1, 306, 1, 200, 300), dtype="uint16")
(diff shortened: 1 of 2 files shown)


src/odemis/gui/util/test/img_test.py
```diff
@@ -999,6 +999,42 @@ def test_spec_float_pp(self):
...
+             model.MD_DESCRIPTION: "TR",
+             model.MD_ACQ_DATE: time.time(),
+             model.MD_BPP: 12,
+             model.MD_PIXEL_SIZE: (2e-6, 2e-6),  # m/px
+             model.MD_POS: (-0.001203511795256, -0.000295338300158),  # m
+             model.MD_EXP_TIME: 0.2,  # s
+             model.MD_TIME_LIST: tld,
+            }
+
+        tld = list(numpy.arange(data.shape[1]) * 0.1e-9)
```
sneepstefan, line 1021: Duplicate line?

tepals (approved):
I agree with Stefans comment, the rest looks good

PR #2252, opened 2022-11-03
driver tmcm: fix version number for CANopen device

pieleric:
There were a couple of small issues when reading the version number:
* the IDENTITY was incorrectly clipped to 8 bits instead of 16 bits
* the SW_VERSION of the TMCM seems to contain NULL followed by garbage
  (which is common on C code). => discard anything after the NULL
  character to avoid confusing the output of swVersion.


tepals (commented):
does it make sense to add a small test case to check for this?

effting (approved):

tepals (approved):

PR #2254, opened 2022-11-04
acq move: on large Rx rotations, move Rx away

pieleric:
Not sure whether is this was a typo, but until now, when Rx was large,
we would move Rz... which doesn't help. The idea is that we can move in
X (because it's not really much affected by the Rx, and then Rx. Rz has
nothing to do there.

In particular, this prevented from going to LOADING position in some
cases, because Rx was so large that moving in Y didn't succeed.


tepals (commented):
Only a comment on the commit message

> 
> acq move: on large Rx rotations, move **~Rx~** **(=> should be X, I think)** away
> Not sure whether **~is~** this was a typo, but until now, when Rx was large,
> we would move Rz... which doesn't help. The idea is that we can move in
> X (because it's not really much affected by the Rx **)**, and then **move** Rx. Rz has
> nothing to do there.
> 
> In particular, this prevented from going to LOADING position in some
> cases, because Rx was so large that moving in Y didn't succeed.
>

pieleric:
I did mean Rx. The idea is that when a large Rx move has to be done, we do it far away from the lenses (by moving in X). Until now we moved Rz.

FCanberk (approved):

tepals (approved):

PR #2257, opened 2022-11-08
gui comp legend: add support for FIB stream icon

pieleric:
Introduce a FIB icon for FIB streams.
Also prepare for more stream types, but the icons are not ready yet.

Also convert the stream type -> icon data to a dict, which simplifies
the code.


src/odemis/gui/comp/legend.py
```diff
@@ -245,16 +250,15 @@ def set_stream_type(self, side, acq_type):
...
+        try:
+            icon = self._type_to_icon[acq_type]
+        except KeyError as ex:
             # Don't fail too bad
             icon = img.getBitmap("icon/ico_blending_opt.png")
             if self.merge_slider.IsShown():
-                logging.warning("Failed to find icon for stream of type %s",
-                                acq_type)
+                # For KeyError, the "message" is just the key name
+                logging.warning("Failed to find icon for stream of type %s", ex)
```
tepals, line 260: This feels like you're mis-using the `KeyError`, why not do something like:
```python
        icon = self._type_to_icon.get(acq_type, None)
        if icon is None:
            # Don't fail too bad
            icon = img.getBitmap("icon/ico_blending_opt.png")
            if self.merge_slider.IsShown():
                logging.warning("Failed to find icon for stream of type %s",
                                acq_type)
```

src/odemis/gui/comp/legend.py
```diff
@@ -22,10 +22,13 @@
 
 """
 
-from collections.abc import Iterable, Mapping
 import cairo
+from collections.abc import Iterable, Mapping
+
```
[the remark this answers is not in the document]
pieleric: I didn't mean to change this part. It was automatically refactored by my IDE. I'd discard it.
That said, yes, normally there is a blank line between the 3 types of imports.

src/odemis/gui/comp/legend.py
pieleric: I disagree it's a mis-use of Exceptions. I'd even argue it's precisely what Exceptions are intended for! The normal code is in the `try` block. We expect that this code works. However, if for some reason we haven't handled it properly (eg, new stream type not yet added), we fallback gracefully in the except block.
Admittedly, it's a thin line. Especially in this code, I'm aware that there are cases where the stream is unknown... That should be only temporary though, while we make the icons for these streams.

tepals: Okay, I would say that at least according to realpython the usual solution is to use `.get()`, but your way is at least the general solution. For readability I still prefer `logging.warning("Failed to find icon for stream of type %s", acq_type)` over `logging.warning("Failed to find icon for stream of type %s", ex)`.

pieleric: Oh, indeed, using `acq_type` is much nicer... I don't know how, but I got convinced that it wasn't available and I had to take the info from the exception!

src/odemis/gui/comp/legend.py
```diff
@@ -60,18 +62,21 @@ def __init__(self, parent, wid=-1, pos=(0, 0), size=wx.DefaultSize,
...
+            # MD_AT_EK: img.getBitmap("icon/ico_blending_spec.png"),
+            MD_AT_EM: img.getBitmap("icon/ico_blending_sem.png"),
+            MD_AT_FIB: img.getBitmap("icon/ico_blending_fib.png"),
+            MD_AT_OVV_TILES: img.getBitmap("icon/ico_blending_map.png"),
+            MD_AT_OVV_FULL: img.getBitmap("icon/ico_blending_navcam.png"),
+            MD_AT_HISTORY: img.getBitmap("icon/ico_blending_history.png"),
+            MD_AT_CL: img.getBitmap("icon/ico_blending_opt.png"),
+            MD_AT_FLUO: img.getBitmap("icon/ico_blending_opt.png"),
+            MD_AT_SLIT: img.getBitmap("icon/ico_blending_slit.png"),
+        }
```
BasHolweg, line 79: For what reason are MD_AT_TEMPORAL and MD_AT_EK commented out?

BasHolweg (approved):

src/odemis/gui/comp/legend.py
pieleric: I don't have the right icons yet. Maxim will make them. (I've already indicated that in the commit description).

pieleric: I've now changed the file name for these icons to be the "real" file name when the file will exists... it's hopefully clearer that for now it's not going to work.

tepals (approved):

PR #2260, opened 2022-11-14
driver xt_client: fix xttoolkit image acquisition in the detector

tepals:
During fastem overview imaging it was noticed that for each tile the image was acquired twice (SDEV-1095) and the stage was still moving during image acquisition (SDEV-873). The cause of this was that the stop acquiring message was not sent in time. Therefore after acquiring a single image we would start scanning again, then while it was scanning trigger the stage move. After the stage move there would be a stop and start message queued. Move the stop message to before canceling scanning to fix this.


src/odemis/driver/xt_client.py
```diff
@@ -2674,14 +2675,16 @@ def _acquire(self):
                 self._acq_wait_start()
                 logging.debug("Preparing acquisition")
                 while True:
-                    if self._acq_should_stop():
+                    logging.debug("Start acquiring an image")
+                    if self._acq_should_stop(timeout=0.5):
```
tepals, line 2679: @pieleric this timeout was still needed, otherwise the GEN_STOP message was still too soon. This will slow down the live stream a bit, do you have any idea for a better fix?

pieleric: I think you mean that the GEN_STOP would still be sent after the `scan_image()` has been called (so more like "too late"), right? I guess what happens is that the first image is acquired, it's sent to the client (ie, the GUI), then the client unsubscribes, which calls stop_generate() and sends a GEN_STOP message. However, in the meantime, the next acquisition has already started. Then `stop_generate()` should cancel the acquisition, but maybe that doesn't work well? So this helps by waiting a little bit before starting the next acquisition, to "makes sure it's worthy starting the next acquisition". 

0.5 s sounds a lot of time (for the client to receive the image and decide to unsubscribe). Wouldn't 0.1 s be enough?

That said, all the behaviour so far is how the the detectors are supposed to behave: if you subscribe to them, they try to acquire as fast as possible. If you are not interested anymore, then you unsubscribe, and the last acquisition is cancelled. Of course here, as you keep acquiring-cancelling, it's at least inefficient, and maybe the current cancellation behaviour is still not proper (do you see a "scan image aborted" every time?).

I think the correct way to handle it using the Odemis API would be to use software trigger (aka "event"). This tells the detector that before starting a new acquisition, it should wait for a trigger. You'd need to extend the acquisition thread to support it (see andorcam2, andorcam3, or semcomedi for examples).

1. the client indicates that the detector should be "synchronized on" an event. (-> nothing happens yet)
2. the client subscribes to the detector -> the detector "gets ready" (not really much happening in your case here)
3. the client sends an event (aka software trigger) -> the acquisition starts
4. the acquisition thread receives an image and waits for another event (-> the client receives the image)
5. the client moves the stage, do more fancy stuff
6. repeat from step 3 until all tiles are acquired.
7. the client unsubscribe
8. the client "synchronize on" nothing.

src/odemis/driver/xt_client.py
```diff
@@ -2693,9 +2696,9 @@ def _acquire(self):
                             raise
 
                     # Acquire the image
-                    # =================
                     image = self.parent.get_latest_image(self.parent._scanner.channel)
 
+                    # Add metadata to the image
```
pieleric, line 2701: Normally we copy the metadata from just after the acquisition started. This way, if the settings are changed during acquisition (eg, dwell time), it's not the new value (which was not yet in use) that's recorded, but properly the original value.

src/odemis/driver/xt_client.py
```diff
@@ -2674,14 +2675,16 @@ def _acquire(self):
...
+                    if self._acq_should_stop(timeout=0.5):
+                        logging.debug("Image acquisition should stop, exiting loop")
                         break
 
                     if hasattr(self._scanner, "blanker"):
                         if self._scanner.blanker.value is None:
                             self.parent.unblank_beam()
 
-                    # Start a complete scan.
+                    logging.debug("Start a complete scan of an image")
```
pieleric, line 2687: Just before starting the acquisition you should note the time, and store this information in MD_ACQ_DATE. This is used in several codes to check that the image received is indeed new enough.

pieleric (changes requested):

src/odemis/driver/xt_client.py
tepals: So this will have to move to above `get_latest_image` and below `scan_image`?

src/odemis/driver/xt_client.py
tepals: > I think you mean that the GEN_STOP would still be sent after the scan_image() has been called (so more like "too late"), right? I guess what happens is that the first image is acquired, it's sent to the client (ie, the GUI), then the client unsubscribes, which calls stop_generate() and sends a GEN_STOP message. 

you are correct, I meant too late.

> However, in the meantime, the next acquisition has already started. Then stop_generate() should cancel the acquisition, but maybe that doesn't work well? So this helps by waiting a little bit before starting the next acquisition, to "makes sure it's worthy starting the next acquisition".

Indeed I think it's not possible to actually stop scanning, which the code in `stop_generate` was supposed to do. Therefore once `scan_image` has started it will finish a full scan.

> 0.5 s sounds a lot of time (for the client to receive the image and decide to unsubscribe). Wouldn't 0.1 s be enough?

I checked the logs and usually GEN_STOP is usually around 0.17 s after `logging.debug("Start acquiring an image")`, so 0.2 is probably safe.

src/odemis/driver/xt_client.py
pieleric: Yes, that's typically the best place for such code, as it'll run essentially while waiting for the acquisition to complete.

src/odemis/driver/xt_client.py
```diff
@@ -2651,6 +2651,7 @@ def __init__(self, name, role, parent, address, **kwargs):
 
     def stop_generate(self):
         logging.debug("Stopping image acquisition")
+        self._genmsg.put(GEN_STOP)
         with self._cancel_access:
             self.cancel_connection._pyroClaimOwnership()
             # Directly calling set_channel_state does not work with XTToolkit.
```
pieleric, line 2657: If that code, to cancel the current acquisition doesn't work, could you add a big warning in comment that it doesn't work (because reading the code, I had assumed the cancellation worked). Alternatively, you could try to find out how to get working, of course ;-)

src/odemis/driver/xt_client.py
```diff
@@ -2674,14 +2675,21 @@ def _acquire(self):
                 self._acq_wait_start()
                 logging.debug("Preparing acquisition")
                 while True:
-                    if self._acq_should_stop():
+                    logging.debug("Start acquiring an image")
+
+                    # The 0.2 timeout is needed to make sure it is worthy to start a new acquisition.
```
pieleric, line 2680: I think it'd be worthy to indicate this is a hack. Something like "HACK: as cancellation doesn't work, to avoid acquiring an extra image at the end of an acquisition, we wait a little bit before starting the next frame. This gives enough time for the client after receiving the previous frame to decide that this is enough. Especially worthy when acquiring a single image."

pieleric (changes requested):

pieleric (approved):

nandishjpatel (approved):

PR #2261, opened 2022-11-16
Add fastem support for non-rectangular in the acquisition manager.

BasHolweg:
Add the functionality to be able to acquire non-rectangular or rotated rectangular regions of acquisition (ROA's). For now the previous method is still used for acquisition since the GUI still requires compatibility. Two utility functions were added for flood filling a ndarray, which is used for finding the inside of the polygonal ROA's and for finding the bounding box of a set of polygon coordinates.

### DIFF
--- a/src/odemis/acq/fastem.py
+++ b/src/odemis/acq/fastem.py
@@ -44,12 +44,11 @@ except ImportError as err:
     fastem_calibrations = False
 
 from odemis import model, util
-from odemis.acq import stitching
+from odemis.acq import fastem_conf, stitching
 from odemis.acq.align.spot import FindGridSpots
 from odemis.acq.stitching import REGISTER_IDENTITY, FocusingMethod
 from odemis.acq.stream import SEMStream
-from odemis.util import TimeoutError
-from odemis.acq import fastem_conf
+from odemis.util import img, TimeoutError
 
 # The executor is a single object, independent of how many times the module (fastem.py) is loaded.
 _executor = model.CancellableThreadPoolExecutor(max_workers=1)
@@ -155,6 +154,12 @@ class FastEMROA(object):
         return tot_time
 
     def _calculate_field_indices(self):
+        # TODO: Old method is used since polygonal fields is not yet supported by the gui. This must be adjusted when
+        #  it is.
+        indices = self.get_square_field_indices(self.coordinates.value)
+        return indices
+
+    def get_square_field_indices(self, coordinates):
         """
         Calculates the number of single field images needed to cover the ROA (region of acquisition). Determines the
         corresponding indices of the field images in a matrix covering the ROA. If the ROA cannot be covered
@@ -179,13 +184,13 @@ class FastEMROA(object):
         6000 = 6400 * (1 - overlap) = field_size * (1 - overlap)
         => n_fields = (abs(r - l) - field_size * overlap) / (field_size * (1 - overlap))
         """
-        l, t, r, b = self.coordinates.value  # tuple of floats: l, t, r, b coordinates in m
-        px_size = self._multibeam.pixelSize.value
-        field_res = self._multibeam.resolution.value
+        l, t, r, b = coordinates  # tuple of floats: l, t, r, b coordinates in m
+        px_size = self._multibeam.pixelSize.value   # Size per pixel in m
+        field_res = self._multibeam.resolution.value    # Number of pixels per field
 
         # The size of a field consists of the effective cell images excluding overscanned pixels.
         field_size = (field_res[0] * px_size[0],
-                      field_res[1] * px_size[1])
+                      field_res[1] * px_size[1])  # [px] * [m/px] = [m]
 
         # Note: Megafields get asymmetrically extended towards the right and bottom.
         # When fields overlap the number of fields is calculated by subtracting the size of the field that overlaps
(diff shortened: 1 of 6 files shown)


src/odemis/acq/fastem.py
```diff
@@ -217,6 +222,143 @@ def _calculate_field_indices(self):
 
         return field_indices
 
+    # TODO: should not require input argument but use self.coordinates when the behaviour is adjusted to support
+    #  multiple coordinate points in stead of 4 separate values
```
nandishjpatel, line 226:
```suggestion
    #  multiple coordinate points instead of 4 separate values
```

src/odemis/acq/fastem.py
```diff
@@ -155,6 +154,12 @@ def estimate_acquisition_time(self):
         return tot_time
 
     def _calculate_field_indices(self):
+        # TODO: Old method is used since polygonal fields is not yet supported by the gui. This must be adjusted when
+        #  it is.
+        indices = self.get_square_field_indices(self.coordinates.value)
+        return indices
+
+    def get_square_field_indices(self, coordinates):
```
nandishjpatel, line 162: `:param coordinates:` left to be stated in docstring

src/odemis/acq/fastem.py
```diff
@@ -217,6 +222,143 @@ def _calculate_field_indices(self):
...
+        :return: (ndarray) Filled numpy array that represent the fields that need to be acquired.
+        """
+        # Get the bounding megafield shape by taking the maximum indices from the indices list and create a boolian
+        # numpy array to represent the megafield with True values for fields that need to be acquired
+        _, _, rmax, cmax = util.get_polygon_bbox(indices)
+        megafield_grid_shape = (rmax + 1, cmax + 1)
+        megafield_grid_rep = numpy.zeros(megafield_grid_shape, dtype=numpy.bool)
+        megafield_grid_rep[[r[0] for r in indices], [c[1] for c in indices]] = True
+
+        # Pad the array with zeros so the flood fill wil always surround the entire shape
```
nandishjpatel, line 347:
```suggestion
        # Pad the array with zeros so the flood fill will always surround the entire shape
```

src/odemis/util/img.py
```diff
@@ -1201,3 +1201,41 @@ def assembleZCube(images, zlevels):
...
+    overflow_counter = 0
+    while len(queue) > 0:
+        # To prevent the possibility of an infinite loop count the number of iterations. If it is larger than the
+        # total number of values in the array raise an error since this should not be possible.
+        if overflow_counter > max_area:
+            raise ValueError(f"Number of loop iterations is higher than maximum iterations possible ({max_area}) in "
+                             f"array of shape: {array.shape}")
+
+        row, col = queue.pop(0)  # Set the current pixel to the first element of the queue and remove it from the queue
+        if not array[row, col]:
```
nandishjpatel, line 1230: If I understood correctly, I think the below code can be optimized reference: https://www.baeldung.com/cs/flood-fill-algorithm#2-naive-implementation

src/odemis/acq/fastem.py
```diff
@@ -217,6 +222,143 @@ def _calculate_field_indices(self):
...
+
+    @staticmethod
+    def _create_and_fill_megafield_rep(indices):
+        """
+        Create a numpy array that represents the given indexes as True values and fills any enclosed False values.
+
+        :param indices: (list of nested tuples (row,column)) Index values of the shape that needs to be filled.
+        :return: (ndarray) Filled numpy array that represent the fields that need to be acquired.
+        """
+        # Get the bounding megafield shape by taking the maximum indices from the indices list and create a boolian
```
tepals, line 340:
```suggestion
        # Get the bounding megafield shape by taking the maximum indices from the indices list and create a boolean 
```

src/odemis/util/img.py
```diff
@@ -1201,3 +1201,41 @@ def assembleZCube(images, zlevels):
         ret = DataArray(ret, metadata3d)
 
         return ret
+
+
+def apply_flood_fill(array, start):
+    """
+    Flood fills a 2-Dimensional numpy array with truth values from a given start position with the 4-connected method.
+    The input array contains a shape represented with truth values for the borders of the shape.
+
```
tepals, line 1210: could you give a small example drawing? Something like

```python
# input array                output array
# [[0, 0, 0, 0, 0, 0,],     [[0, 0, 0, 0, 0, 0,],
#  [0, 0, 1, 1, 0, 0,],      [0, 0, 1, 1, 0, 0,],
#  [0, 1, 0, 0, 1, 0,],  ->  [0, 1, 1, 1, 1, 0,],
#  [0, 0, 1, 1, 1, 0,],      [0, 0, 1, 1, 1, 0,],
#  [0, 0, 0, 0, 0, 0,]]      [0, 0, 0, 0, 0, 0,]]
```

src/odemis/util/img.py
```diff
@@ -1201,3 +1201,41 @@ def assembleZCube(images, zlevels):
...
+    :param array: (ndarray(bool)) binary array of size mxn containing the shape to be filled
+    :param start: (tuple(int, int)) position from which to start flood fill
+    :return: (ndarray(bool)) array of size mxn containing the filled shape, with filled values True
+    """
+    if start >= array.shape:
+        raise ValueError(f"Start position of {start} does not lie within the array.")
+
+    array = array.copy()
```
tepals, line 1218: `array` is a python standard module, I think it's good to be a bit careful with naming things the same as other libraries. I would suggest to rename this, maybe to something like `input_array` or `shape_array`

src/odemis/util/img.py
```diff
@@ -1201,3 +1201,41 @@ def assembleZCube(images, zlevels):
...
+    max_area = array.shape[0] * array.shape[1]
+    queue = [start]
```
tepals, line 1220: `queue` is a library that's used in other places in Odemis, I would suggest to rename this, maybe to `check_pixel`

src/odemis/util/test/img_test.py
```diff
@@ -1181,6 +1181,40 @@ def test_rotate_img_metadata(self):
             numpy.testing.assert_array_equal(image, rotated_img)  # The data should not change, only the metadata.
 
 
+class TestFloodFill(unittest.TestCase):
+
+    def test_standard_fill(self):
+        """Test the flood_fill as it is expected to be used"""
```
tepals, line 1187: I would say"

```suggestion
        """Test the expected use of flood_fill"""
```

src/odemis/util/test/img_test.py
```diff
@@ -1181,6 +1181,40 @@ def test_rotate_img_metadata(self):
...
+        for i in range(min(a.shape)):
+            a[i, i] = True
+            a[-i - 1, i] = True
+
+        filled_array = img.apply_flood_fill(a, (3, 0))
+
+        self.assertTrue(filled_array[3, 1])
+        self.assertFalse(filled_array[3, -1])
+        self.assertFalse(filled_array[0, 3])
+        self.assertFalse(filled_array[-1, 3])
```
tepals, line 1198: I think you could do a more thorough check here. You know exactly what the filled array should look like, so you could just create the expected filled array and check that they are equal.

```python
expected_array = numpy.array([
       [1, 0, 0, 0, 0, 0, 1],
       [1, 1, 0, 0, 0, 1, 0],
       [1, 1, 1, 0, 1, 0, 0],
       [1, 1, 1, 1, 0, 0, 0],
       [1, 1, 1, 0, 1, 0, 0],
       [1, 1, 0, 0, 0, 1, 0],
       [1, 0, 0, 0, 0, 0, 1]
])
numpy.testing.assert_array_equal(expected_array, filled_array)
```

src/odemis/util/test/util_test.py
```diff
@@ -543,6 +543,20 @@ def test_expand_rect(self):
                                                        "expanded rectangle %s" % (expanded_rect, expected_rect))
 
 
+class GetMinMaxTupleTestCase(unittest.TestCase):
+
+    def test_minmax(self):
```
tepals, line 548: can you rename these to match the new function name?

src/odemis/util/test/util_test.py
```diff
@@ -543,6 +543,20 @@ def test_expand_rect(self):
...
+        """Test getting the minimum and maximum values from a list of tuples"""
```
tepals, line 549: can you add a test case that checks what happens when you give the function negative values?

src/odemis/util/__init__.py
```diff
@@ -389,6 +389,20 @@ def expand_rect(rect, margin):
     return minx - margin, miny - margin, maxx + margin, maxy + margin
 
 
+def get_polygon_bbox(coordinates):
+    """
+    Get the maximum and minimum values for a and b from a list of tuples
+    with shape: [(a1,b1), (a2,b2), ....., (an,bn)]
+
+    :param coordinates: (list of nested tuples (a,b))
```
tepals, line 397: In your implementation coordinates can also be a single tuple, for instance `coordinates = [(0, 2)]`. The output is then `[0, 2, 0, 2]`, I think it would make sense to raise an error when the input is a single coordinate.

Additionally the input tuples can also be more than 2 values long, I think it would also make sense to check for that or mention in the docstring that only the first two values of each tuple are used. For instance you can have `coordinates = [(0, 1, 2), (0, 3)]`, the output is then `(0, 1, 0, 3)`.

tepals (changes requested):

pieleric (changes requested):

src/odemis/util/__init__.py
BasHolweg: That functionality has unfortunately been added since python 3.10.

src/odemis/acq/test/fastem_test.py
```diff
@@ -200,9 +200,9 @@ def test_estimate_acquisition_time(self):
 
             self.assertAlmostEqual(estimated_roa_acq_time, roa_acq_time)
 
-    def test_calculate_field_indices(self):
+    def test_get_square_field_indices(self):
         """Check that the correct number and order of field indices is returned and that row and column are in the
-        correct order."""
+                correct order."""
```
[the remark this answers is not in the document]
BasHolweg: It was not a conscious change, I think pycharm reformatted it for me. Have changed it back now

src/odemis/acq/fastem.py
tepals: don't forget this comment

src/odemis/acq/fastem.py
```diff
@@ -217,6 +222,139 @@ def _calculate_field_indices(self):
...
+
+        :param indices: (list of nested tuples (row,column)) Index values of the shape that needs to be filled.
+        :return: (ndarray) Filled numpy array that represent the fields that need to be acquired.
+        """
+        # Get the bounding megafield shape by taking the maximum indices from the indices list and create a boolean
+        # numpy array to represent the megafield with True values for fields that need to be acquired
+        _, _, rmax, cmax = util.get_polygon_bbox(indices)
+        megafield_grid_shape = (rmax + 1, cmax + 1)
+        megafield_grid_rep = numpy.zeros(megafield_grid_shape, dtype=numpy.bool)
+        megafield_grid_rep[[r[0] for r in indices], [c[1] for c in indices]] = True
```
tepals, line 341: this would also work:
```
megafield_grid_rep[numpy.array(indices)[:, 0], numpy.array(indices)[:, 1] = True
```
pick whichever you prefer

tepals (commented):
looks good!

tepals (approved):

nandishjpatel (approved):

PR #2262, opened 2022-11-16
fix syntax errors in errors being raised in andorcam3 and xt_client

tepals:
In two places the errors were raised with a comment like this raise Error("comment %s", value), however this is not the correct syntax for raising errors and will not put the value after the comment.


pieleric (approved):
We actually have a regex in util/runtest.sh that should detect (most of) these issues... but we don't pay attention to it apparently.
Maybe we should think of moving it to CI check.

tepals:
yes, I noticed because I was running the that script :). Moving it to CI sounds like a good idea

BasHolweg (approved):

PR #2264, opened 2022-11-21
runtests update to only show the short test summary info in the test …

tepals:
…summary file

The test summary used to show anything accept the passed test cases. Because there are a lot of warnings and logs the test summary per module was often more than 40 lines long and then the middle part was cut out. This resulted in some of the failures being cut out of the summary as well. From now on we only show the short test summary in the summary file. This contains one line per failed test case. There is no cutting in the summary anymore, so even if there are more than 40 fails,they will still all be shown. The summary for each test module starts with 'Running ....', with the filepath of the test module on the dots.

The example test output from ubuntu 20.04 started with the wrong line, so that line was removed.

### DIFF
--- a/util/pytest_log_filter.py
+++ b/util/pytest_log_filter.py
@@ -34,7 +34,9 @@ def filter_test_log(log_txt, filter_type='summary'):
     If the output log is too long the middle part of the logged is not returned.
 
     :param log_txt: (str) String with the log of a single test file.
-    :return (str): filtered log, returns None if there is nothing interesting in log_txt
+    :param filter_type: (str) The type of message that we want to filter out. Optional, default 'summary'.
+        'summary' returns the short test summary output, and 'warning' returns the warnings summary.
+    :return: (str) filtered log, returns None if there is nothing interesting in log_txt
 
     Example of filtering:
     This function expects a pytest test report of single test file as input, a string such as the example below:
--- a/util/pytest_log_filter.py
+++ b/util/pytest_log_filter.py
@@ -27,7 +27,7 @@ import re
 import sys
 
 
-def filter_test_log(log_txt):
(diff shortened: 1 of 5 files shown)


util/pytest_log_filter.py
```diff
@@ -27,7 +27,7 @@
 import sys
 
 
-def filter_test_log(log_txt):
+def filter_test_log(log_txt, filter_type='summary'):
```
tepals, line 30: add `filter_type` to docstring

util/test/test_pytest_log_filter.py
```diff
@@ -9,20 +9,20 @@ class TestFilterTestLog(unittest.TestCase):
     """
 
     def test_sample_input_ubuntu_18_04(self):
```
tepals, line 11: update tests for warnigns

pieleric (changes requested):

util/pytest_log_filter.py
```diff
@@ -34,6 +34,8 @@ def filter_test_log(log_txt, filter_type='summary'):
     If the output log is too long the middle part of the logged is not returned.
 
     :param log_txt: (str) String with the log of a single test file.
+    :param filter_type: (str) The type of message that we want to filter out. Optional, default 'summary'.
+        'summary' returns the short test summary output, and 'warning' returns the warnings summary.
     :return (str): filtered log, returns None if there is nothing interesting in log_txt
```
K4rishma, line 39:
```suggestion
    :return: (str) filtered log, returns None if there is nothing interesting in log_txt
```

BasHolweg (approved):

pieleric (approved):

K4rishma (approved):

PR #2267, opened 2022-11-21
dataio hdf5: remove "past" dependency

pieleric:
Now that we don't support Python 2, we don't need to use these past
dependencies. This clarifies/simplifies a little bit the code.

### DIFF
--- a/src/odemis/dataio/hdf5.py
+++ b/src/odemis/dataio/hdf5.py
@@ -19,21 +19,20 @@ PURPOSE. See the GNU General Public License for more details.
 You should have received a copy of the GNU General Public License along with 
 Odemis. If not, see http://www.gnu.org/licenses/.
 '''
-from past.builtins import basestring, unicode
-from numpy.polynomial import polynomial
-
 from collections.abc import Iterable
-import h5py
 import json
 import logging
-import numpy
-from odemis import model
-import odemis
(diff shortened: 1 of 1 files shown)


BasHolweg (approved):

src/odemis/dataio/hdf5.py
```diff
@@ -19,21 +19,20 @@
...
 import time
 
+from numpy.polynomial import polynomial
+import h5py
+import numpy
+
+from odemis import model
+from odemis.util import fluo, img, spectrum
+from odemis.util.conversion import JsonExtraEncoder
+import odemis
```
sneepstefan, line 35: `import odemis` should go above `from odemis import model` according to our new coding standards.

src/odemis/dataio/hdf5.py
```diff
@@ -19,21 +19,20 @@
...
 
 # User-friendly name
 FORMAT = "HDF5"
```
sneepstefan, line 38: line 54 contains a typo; CTZXY should be CTZYX

src/odemis/dataio/hdf5.py
pieleric: I thought we had to separate "from"s and "import"s.... but apparently not. So I'll fix it!

PR #2268, opened 2022-11-22
gui util img: make calculate_ticks() symetrical

pieleric:
Whether the range is from 100 -> 1000 or 1000 -> 100, or -1000 -> -100,
the number of ticks and values should be the same (modulo the sign).

Also allow to have ticks near the border, as it's actually quite useful
when reading the graph, and now the code displaying the values do handle
properly the positioning of text next to the border.

### DIFF
--- a/src/odemis/gui/util/test/img_test.py
+++ b/src/odemis/gui/util/test/img_test.py
@@ -137,6 +137,24 @@ class TestCalculateTicks(unittest.TestCase):
             # ticks are made of 2 values
             self.assertEqual(len(ticks[0]), 2)
 
+    def test_neg_value(self):
+        """
+        Check that a range with negative values behaves the same as a range with positive values
+        """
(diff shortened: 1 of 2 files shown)


src/odemis/gui/util/test/img_test.py
```diff
@@ -137,6 +137,24 @@ def test_simple(self):
             # ticks are made of 2 values
             self.assertEqual(len(ticks[0]), 2)
 
+    def test_neg_value(self):
+        """
+        Check that a range with negative values behaves the same as a range with positive values
+        """
+        ranges = [(100, 1000), (0, 9500), (1e-3, 8e-3), (-5, 6), (-2, -1000)]
+        # ranges = [(0, 9500)]
```
pieleric, line 145: This line can go

tepals (approved):

PR #2269, opened 2022-11-28
Sdev 1181 implement updating xtadapter from ubuntu to windows

nandishjpatel:
This PR should be reviewed together with 
- https://bitbucket.org/delmic/xtlib-adapter/pull-requests/70
- https://bitbucket.org/delmic/ubuntu-delmic/pull-requests/5

Flow (xtadapter release management):
- [xt_client line](https://github.com/nandishjpatel/odemis/blob/e864631d800198517d7aff99fc379242fbaeb095/src/odemis/driver/xt_client.py#L217): get the xtadapter version, type, bitness from the running application on Windows PC
- [xt_client line](https://github.com/nandishjpatel/odemis/blob/e864631d800198517d7aff99fc379242fbaeb095/src/odemis/driver/xt_client.py#L224): check if a newer package is available under `/usr/share/xtadapter` i.e. the directory where the zip files will be made available from the xtadapater debian package
- [xt_client line](https://github.com/nandishjpatel/odemis/blob/e864631d800198517d7aff99fc379242fbaeb095/src/odemis/driver/xt_client.py#L231): if a newer package (zip file) is available transfer it in the form of bytes
- xt_client: [Done](https://github.com/nandishjpatel/odemis/blob/8d89e80b9629ac29713b9be0358d1e3912444a58/src/odemis/driver/xt_client.py#L267) a pop-up starting a newer version in available, please safety restart the xtadapter application on the Windows PC to install it
- [xt_adapter:base_adapter](https://bitbucket.org/delmic/xtlib-adapter/src/58c99cd1ebade55e9625ebd43eda935e4890e497/xtadapter/base_adapter.py#lines-1441) : from the bytes, create a zipfile in memory and extract all contents of the zipfile in `Downloads` directory of the Windows PC
- [xt_adapter:server](https://bitbucket.org/delmic/xtlib-adapter/src/58c99cd1ebade55e9625ebd43eda935e4890e497/xtadapter/server.py#lines-195) : when the user will restart the xtadpater application, a message box pop-us. If the user would like to update the application then the latest package installer will start

### DIFF
--- a/src/odemis/driver/test/xt_client_test.py
+++ b/src/odemis/driver/test/xt_client_test.py
@@ -23,6 +23,7 @@ http://www.gnu.org/licenses/.
 import logging
 import math
 import os
+import shutil
 import time
 import unittest
 
@@ -613,6 +614,16 @@ class TestMicroscopeInternal(unittest.TestCase):
         if self.microscope.get_vacuum_state() != 'vacuum':
             self.skipTest("Chamber needs to be in vacuum, please pump.")
         self.xt_type = "xttoolkit" if "xttoolkit" in self.microscope.swVersion.lower() else "xtlib"
+
+    def test_get_current_version(self):
+        """Test get current adapter version from software version"""
+        sw_version = "xt: 1.1.1; xtadapter: 1.2.1"
+        version = self.microscope.get_current_version(sw_version)
+        self.assertEqual(version, "1.2.1")
+
+        sw_version = "xt: 1.1.1; xtadapter: 11.2.1; xttoolkit"
+        version = self.microscope.get_current_version(sw_version)
+        self.assertEqual(version, "11.2.1")
 
     def test_acquisition(self):
         """Test acquiring an image."""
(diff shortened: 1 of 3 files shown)


src/odemis/driver/xt_client.py
```diff
@@ -67,6 +69,108 @@
...
 
+class Package(object):
+    """A class containing relevant information about a xtadapter package."""
+
+    def __init__(self):
+        self._adapter = None
+        self._bitness = None
+        self._name = None
+        self._path = None
+        self._version = None
```
tepals, line 81: can you add a small comment explaining the different attributes

src/odemis/driver/xt_client.py
```diff
@@ -67,6 +69,108 @@
...
+
+    @property
+    def adapter(self):
```
tepals, line 84: can you add typing information to all properties?

src/odemis/driver/xt_client.py
```diff
@@ -67,6 +69,108 @@
...
+                package.path = os.path.join(directory, entity)
+                # Filter using adapter type and bitness
+                if bitness == package.bitness and adapter == package.adapter:
+                    version_package[package.version] = package
+        versions = list(version_package)
+        if len(versions):
+            # Sort the available versions in reverse order i.e. the latest version is the first element
+            versions.sort(key=lambda s: [int(u) for u in s.split(".")], reverse=True)
+            if current_version < versions[0]:
+                print("The latest version is {}.\n".format(versions[0]))
```
tepals, line 184: why do you use a print statement here?

src/odemis/driver/xt_client.py
```diff
@@ -106,6 +210,33 @@ def __init__(self, name, role, children, address, daemon=None,
...
+            if current_version is not None:
+                package = check_latest_package(
+                    directory=install_dir,
+                    current_version=current_version,
+                    adapter=adapter,
+                    bitness=bitness,
+                    is_zip=True,
+                )
+            if package.version is not None:
+            # Open the package's zip file as bytes and transfer them
```
tepals, line 253: I would add another tab before this comment

src/odemis/driver/xt_client.py
```diff
@@ -164,6 +324,34 @@ def __init__(self, name, role, children, address, daemon=None,
...
+        :return: The element's value, if not found return None.
+
+        """
+        # Example information: xt: 1.1.1; xtadapter: 1.2.1
+        items = info.split(";")
+        for item in items:
+            if element in item.lower():
+                value = item.split(":")[1].strip()
+                return value
+        return None
```
tepals, line 342: So far all the functions on this class correspond exactly to a function on the xtadapter. I'm not sure if this function actually fits here. Also I think you can just replace this by a single regex, so then you might as well just put that on line 243 for instance for the xtadapter element
```
current_version = re.search("xtadapter: (\s*([\d.]+))", info)
current_version = current_version.group(1) if current_version is not None else None
```

src/odemis/driver/xt_client.py
```diff
@@ -67,6 +70,127 @@
...
+                else os.path.isdir(entity_path)
+                and len(
+                    [
+                        entity
+                        for entity in os.listdir(entity_path)
+                        if entity.endswith(".exe")
+                    ]
+                )
+                == 1
+            ):
```
tepals, line 163: this is very hard to read, can you somehow clean it up?

src/odemis/driver/test/xt_client_test.py
```diff
@@ -614,6 +615,20 @@ def setUp(self):
             self.skipTest("Chamber needs to be in vacuum, please pump.")
         self.xt_type = "xttoolkit" if "xttoolkit" in self.microscope.swVersion.lower() else "xtlib"
 
+    def test_get_value_from_software_version(self):
```
tepals, line 618: can you add a test for `transfer_latest_package` and maybe by default skip it, so we do not transfer a package every time we run these test cases

src/odemis/driver/test/xt_client_test.py
```diff
@@ -1462,5 +1477,256 @@ def test_apply_autostigmator(self):
...
+            "delmic-fastem-xtadapter-64bit-v1.11.2",    # has 1 .exe file
+            "delmic-xtadapter-32bit-v1.11.2",           # has 1 .exe file
+            "delmic-fastem-xtadapter-64bit-v2.11.2",    # has 0 .exe file
+            "delmic-xtadapter-32bit-v2.11.2",           # has 0 .exe file
+            "delmic-fastem-xtadapter-64bit-v3.11.2",    # has 2 .exe file
+            "delmic-xtadapter-32bit-v3.11.2",           # has 2 .exe file
+        ]
+        for dir_name in self.dir_names:
+            if not os.path.isdir(os.path.join(self.dir, dir_name)):
+                os.makedirs(os.path.join(self.dir, dir_name))
```
tepals, line 1503: I think you can also use `os.makedirs(os.path.join(self.dir, dir_name), exist_ok=True)`

src/odemis/driver/test/xt_client_test.py
```diff
@@ -1462,5 +1477,256 @@ def test_apply_autostigmator(self):
...
+            if "2.11.2" not in dir_name:
+                f1 = open(os.path.join(self.dir, dir_name, "test1.exe"), "w")
+                f1.close()
+                if "3.11.2" in dir_name:
+                    f2 = open(os.path.join(self.dir, dir_name, "test2.exe"), "w")
+                    f2.close()
```
tepals, line 1509: can you add a short comment explaining why you do this?

tepals (changes requested):

src/odemis/driver/xt_client.py
nandishjpatel: So that the user first gets to read the print in the application window, next the message box pops-up asking yes/no question for the update.

src/odemis/driver/test/xt_client_test.py
nandishjpatel: Sure, will do so

nandishjpatel: I added an additional check [here](https://github.com/nandishjpatel/odemis/blob/46cd9bf7c74b01ffd64e4645496909b7a3fbaba7/src/odemis/driver/xt_client.py#L256-L258)

src/odemis/driver/xt_client.py
```diff
@@ -106,6 +232,50 @@ def __init__(self, name, role, children, address, daemon=None,
...
+                if ret is None:
+                    # Open the package's zip file as bytes and transfer them
+                    with open(package.path, mode="rb") as f:
+                        data = f.read()
+                        self.transfer_latest_package(data)
+                        f.close()
+                        notify2.init("odemis.driver.xt_client")
+                        update = notify2.Notification(
+                            "Update Delmic XT Adapter", "Newer version {} is available on Thermo Fisher Windows PC.\n\n"
+                            "How to update?\n\n1. Safely close Odemis and Delmic XT Adapter.\n2. Restart the Delmic XT Adapter "
```
tepals, line 269: Do you really need to close Odemis? And do you then need to do a full stop, or just close the GUI?

pieleric: I guess that in some cases the xt_client could detect the connection is down, and automatically reconnect to the new version of xt-adapter, but that's looking for troubles. So it makes sense to do a full stop of Odemis.

For the message. " Thermo Fisher Windows PC" -> "ThermoFisher Support PC"  (no space between "thermo" and "fisher").

src/odemis/driver/xt_client.py
pieleric: Don't expect the driver's standard output to be visible (typically it's hidden). Just use `logging` for any output.

src/odemis/driver/xt_client.py
```diff
@@ -106,6 +232,50 @@ def __init__(self, name, role, children, address, daemon=None,
                           "uri is correct and XT server is"
                           " connected to the network. %s" % (address, err))
 
+        # Steps to transfer latest xtadapter package if available
+        # The transferred package will be a zip file in the form of bytes
+        try:
+            # xtadapter debian package installation directory which contains xtadapter's zip files
+            package = None
+            install_dir = "/usr/share/xtadapter"
```
pieleric, line 240: That should be a constant at the top of the file.

src/odemis/driver/xt_client.py
```diff
@@ -106,6 +232,50 @@ def __init__(self, name, role, children, address, daemon=None,
...
+                zip_file = zipfile.ZipFile(package.path)
+                ret = zip_file.testzip()
+                zip_file.close()
+                if ret is None:
+                    # Open the package's zip file as bytes and transfer them
+                    with open(package.path, mode="rb") as f:
+                        data = f.read()
+                        self.transfer_latest_package(data)
+                        f.close()
+                        notify2.init("odemis.driver.xt_client")
```
pieleric, line 266: Make it generic with `notify2.init(__name__)`?

src/odemis/driver/xt_client.py
```diff
@@ -106,6 +232,50 @@ def __init__(self, name, role, children, address, daemon=None,
...
+            if package is not None:
+                zip_file = zipfile.ZipFile(package.path)
+                ret = zip_file.testzip()
+                zip_file.close()
+                if ret is None:
+                    # Open the package's zip file as bytes and transfer them
+                    with open(package.path, mode="rb") as f:
+                        data = f.read()
+                        self.transfer_latest_package(data)
+                        f.close()
```
pieleric, line 265: Why not just ending this `with` block here? The rest of the block can be done outside of the `with`. Then you don't need to explicitly close the file.

src/odemis/driver/xt_client.py
```diff
@@ -106,6 +232,50 @@ def __init__(self, name, role, children, address, daemon=None,
...
+                    directory=install_dir,
+                    current_version=current_version,
+                    adapter=adapter,
+                    bitness=bitness,
+                    is_zip=True,
+                )
+            if package is not None:
+                zip_file = zipfile.ZipFile(package.path)
+                ret = zip_file.testzip()
+                zip_file.close()
```
pieleric, line 259: Add a comment to indicate that it just checks it's a proper zip file.

src/odemis/driver/xt_client.py
```diff
@@ -106,6 +232,50 @@ def __init__(self, name, role, children, address, daemon=None,
                           "uri is correct and XT server is"
                           " connected to the network. %s" % (address, err))
 
+        # Steps to transfer latest xtadapter package if available
+        # The transferred package will be a zip file in the form of bytes
+        try:
```
pieleric, line 237: Please move this whole block to a separate method.

src/odemis/driver/xt_client.py
```diff
@@ -164,6 +334,17 @@ def __init__(self, name, role, children, address, daemon=None,
                 self._detector = Detector(parent=self, daemon=daemon, **ckwargs)
             self.children.value.add(self._detector)
 
+    def transfer_latest_package(self, data: bytes) -> None:
+        """
+        Transfer the latest xtadapter package.
```
pieleric, line 339: Can you indicate that there is a size limit, coming from Pyro5.

src/odemis/driver/xt_client.py
```diff
@@ -21,9 +21,14 @@
 """
 import logging
 import math
+import notify2
```
pieleric, line 24: It will need to have `notification-daemon` installed, right? If so, you need to add it to the dependencies of Odemis, in `debian/control`.
I'm actually wondering if that extra package is really necessary... Especially this component is not supported in the latest version of Gnome (though we still have time to change as Ubuntu 22.04 still has it).

pieleric (changes requested):

src/odemis/driver/xt_client.py
nandishjpatel: @pieleric because xt_client.py is run with `sudo`, we need `notification-daemon` (i.e. a separate service which makes use of notification-daemon) for `notify2` to work. Should I add it to `debian/control`? or this [PR](https://bitbucket.org/delmic/ubuntu-delmic/pull-requests/5) will suffice?

src/odemis/driver/test/xt_client_test.py
tepals: is that the correct link, I don't see it there

nandishjpatel: I believe [this additional](https://github.com/nandishjpatel/odemis/blob/5b35cbdf82b67cc7527465255a36a921d90324ac/src/odemis/driver/xt_client.py#L335) check if it's a proper zip file overcomes the need for [transfer_latest_package](https://bitbucket.org/delmic/xtlib-adapter/src/9708c2ef2f5b15cee2fd9a0ff0688c261a4dfbc0/xtadapter/base_adapter.py#lines-1438) test-case because standard modules are being used. What do you suggest @tepals?

src/odemis/driver/xt_client.py
pieleric: Let's add it to debian/control. We'll deal with it differently once it's gone...

src/odemis/driver/xt_client.py
```diff
@@ -66,6 +71,130 @@
...
+                if not is_zip:
+                    for entity in os.listdir(package.path):
+                        if entity.endswith(".exe"):
+                            package.path = os.path.join(package.path, entity)
+                            break
+                # Filter using adapter type and bitness
+                if bitness == package.bitness and adapter == package.adapter:
+                    version_package[package.version] = package
+        versions = list(version_package)
+        if len(versions):
```
pieleric, line 185: Just write `if versions`.

src/odemis/driver/xt_client.py
```diff
@@ -66,6 +71,130 @@
...
+                # If not a zip file, assign the path of the executable
+                if not is_zip:
+                    for entity in os.listdir(package.path):
+                        if entity.endswith(".exe"):
+                            package.path = os.path.join(package.path, entity)
+                            break
+                # Filter using adapter type and bitness
+                if bitness == package.bitness and adapter == package.adapter:
+                    version_package[package.version] = package
+        versions = list(version_package)
```
pieleric, line 184: Can you add a separating line, and maybe a comment for this part like "Check if any version is newer than the current one".

src/odemis/driver/xt_client.py
```diff
@@ -66,6 +71,130 @@
...
+        if len(versions):
+            # Sort the available versions in reverse order i.e. the latest version is the first element
+            versions.sort(key=lambda s: [int(u) for u in s.split(".")], reverse=True)
```
pieleric, line 187: Instead of sorting in reverse order, and then taking the first one, you could just say `newest_version = max(versions, key=...)`. It should be more readable.

src/odemis/driver/xt_client.py
```diff
@@ -164,6 +297,63 @@ def __init__(self, name, role, children, address, daemon=None,
...
+                # Check if it's a proper zip file
+                zip_file = zipfile.ZipFile(package.path)
+                ret = zip_file.testzip()
+                zip_file.close()
+                if ret is None:
+                    # Open the package's zip file as bytes and transfer them
+                    with open(package.path, mode="rb") as f:
+                        data = f.read()
+                    self.transfer_latest_package(data)
+                    notify2.init(__name__)
```
pieleric, line 344: Please add a comment explaining this whole part is just to show a notification to the user.

pieleric, line 344: The backend uses "Odemis" as the name in the notifications. It might be better to use the same, so that notifications are grouped together. (or at least, there is a chance that this happens).

pieleric (changes requested):
I think this PR is almost ready to be merged. Just a few last tiny comments....

pieleric (approved):

tepals (approved):

PR #2270, opened 2022-11-29
Driver ueye crop image data

sneepstefan:
The sensor of the uEye is sometimes too big compared to the image received. That means that on the border, there is no image. While on the live image it’s not a big issue, for tiling, it causes black line between each tile. 

To avoid this, we should have a way to crop the image. Probably the best is to use the same as for the andorcam3 (Zyla): add a max_res parameter to the component that restrict the maximum resolution reported by the driver. It’s always centered.

=> Extend ueye.Camera to support max_res (int, int).
This can be done by using the is_AOI() function.

See specification here: Technology Development/Software/Odemis/Specification uEye max_res argument.docx

https://docs.google.com/document/d/1iTjxhw3LBRfkOiASk4N_D_p9gXiU_BSG/edit?pli=1

### DIFF
--- a/src/odemis/driver/test/ueye_test.py
+++ b/src/odemis/driver/test/ueye_test.py
@@ -24,6 +24,7 @@ Odemis. If not, see http://www.gnu.org/licenses/.
 
 import logging
 from odemis.driver import ueye
+from odemis import model
 import os
 import unittest
 from unittest.case import skip
@@ -37,7 +38,7 @@ logging.getLogger().setLevel(logging.DEBUG)
 TEST_NOHW = (os.environ.get("TEST_NOHW", "0") != "0")  # Default to Hw testing
 
 CLASS = ueye.Camera
-KWARGS = dict(name="camera", role="ccd", device=None, transp=[2, -1])
+KWARGS = dict(name="camera", role="ccd", device=None, max_res=None, transp=[2, -1])
 
 
 class StaticTestUEye(VirtualStaticTestCam, unittest.TestCase):
@@ -95,6 +96,55 @@ class TestSynchronized(VirtualTestSynchronized, unittest.TestCase):
         if TEST_NOHW:
             self.skipTest("No simulator available")
         VirtualTestCam.setUp(self)
+
+
+class TestUEyeCroppingTranspose(VirtualTestCam, unittest.TestCase):
+    """
+    Test setting up the UEye class with a transposed image and a set resolution to force cropping.
+    """
+    camera_type = CLASS
+    camera_kwargs = KWARGS
+
+    @classmethod
+    def setUpClass(cls):
+        if TEST_NOHW:
+            raise unittest.SkipTest('No camera HW present. Skipping tests.')
+
+        # set a fixed max_res smaller than the maximum supported camera resolution
+        cls.camera_kwargs['max_res'] = (768, 1000)
+
+        super(TestUEyeCroppingTranspose, cls).setUpClass()
+
+    def test_set_resolution(self):
+        """
+        Test if the max resolution and the transpose argument are passed and applied.
+        Also test if an acquired image is of the correct dimensions and the axes are right also with binning active.
+        """
+        # assign the right resolutions after initialization of the camera class
+        sensor_res_raw = self.camera._sensor_res
+        expected_res = self.camera_kwargs['max_res']
+        sensor_res_user = self.camera.getMetadata()[model.MD_SENSOR_SIZE]
+
+        # check if the set camera resolution matches the expected resolution
+        self.assertEqual(expected_res, self.camera.resolution.value)
+
+        # check if the resolution is not the same, own cam res should be higher
+        self.assertNotEqual(self.camera_kwargs['max_res'], sensor_res_raw)
+
+        # check if the transpose parameter is applied and axes are switched
+        self.assertEqual(sensor_res_raw, sensor_res_user[::-1])
+
+        # be sure to have binning set to base value
+        self.camera.binning.value = (1, 1)
+        # check if the acquired image is of the right size
+        im = self.camera.data.get()
+        self.assertEqual(im.shape, expected_res[::-1])
+
+        # check if the acquired image is of the right size with higher binning
+        self.camera.binning.value = (2, 4)
+        binned_resolution = (expected_res[0] // 2, expected_res[1] / 4)
+        im = self.camera.data.get()
+        self.assertEqual(im.shape, binned_resolution[::-1])
 
 
 if __name__ == '__main__':
--- a/src/odemis/driver/ueye.py
+++ b/src/odemis/driver/ueye.py
@@ -725,14 +725,17 @@ class Camera(model.DigitalCamera):
             sensor_res_user = self._transposeSizeToUser(self._sensor_res)
             if max_res is None:
                 max_res = sensor_res_user
+                forced_max_res = False
             else:
                 max_res = tuple(max_res)
-                # force image cropping of max_res is specified in the configuration of the camera
-                self._set_aoi(self._sensor_res, max_res)
+                forced_max_res = True
             if not all(1 <= mr <= r for mr, r in zip(max_res, sensor_res_user)):
-                raise ValueError("max_res has to be between 1, 1 and %s, but got %s", sensor_res_user, max_res)
+                raise ValueError(f"max_res has to be between (1, 1) and {sensor_res_user}, but got {max_res}")
             max_res_hw = self._transposeSizeFromUser(max_res)
 
+            # force image cropping of max_res is specified in the configuration of the camera
+            if forced_max_res:
+                self._set_aoi(self._sensor_res, max_res)
             self._metadata[model.MD_SENSOR_SIZE] = sensor_res_user
 
             # old
@@ -1157,7 +1160,7 @@ class Camera(model.DigitalCamera):
 
         # TODO use the return value to see if the command was successful??
         self._dll.is_AOI(self._hcam, AOI_IMAGE_SET_AOI, byref(rect_aoi), sizeof(rect_aoi))
-        logging.debug("Updated area of interest from %s to %s", sensor_res, max_res)
+        logging.debug("Updated area of interest from %s to %s" % (sensor_res, max_res))
 
         # the use of SetFrameRate and SetExposure afterwards is recommended by the programmer's guide
 
--- a/src/odemis/driver/ueye.py
+++ b/src/odemis/driver/ueye.py
@@ -717,29 +717,22 @@ class Camera(model.DigitalCamera):
 
             self._metadata[model.MD_DET_TYPE] = model.MD_DT_INTEGRATING
 
-            # old
-            # res = (sensorinfo.nMaxWidth, sensorinfo.nMaxHeight)
-
-            # new
+            # compare the max_res parameter and the resolution of the camera sensor
+            # if the max_res param is smaller image cropping will be set through AOI
             self._sensor_res = (sensorinfo.nMaxWidth, sensorinfo.nMaxHeight)
             sensor_res_user = self._transposeSizeToUser(self._sensor_res)
+
             if max_res is None:
                 max_res = sensor_res_user
-                forced_max_res = False
             else:
                 max_res = tuple(max_res)
-                forced_max_res = True
             if not all(1 <= mr <= r for mr, r in zip(max_res, sensor_res_user)):
                 raise ValueError(f"max_res has to be between (1, 1) and {sensor_res_user}, but got {max_res}")
             max_res_hw = self._transposeSizeFromUser(max_res)
 
-            # force image cropping of max_res is specified in the configuration of the camera
-            if forced_max_res:
-                self._set_aoi(self._sensor_res, max_res)
+            # setting a specific Area Of Interest will force image cropping
+            self._set_aoi(sensor_res_user, max_res)
             self._metadata[model.MD_SENSOR_SIZE] = sensor_res_user
-
-            # old
-            # self._metadata[model.MD_SENSOR_SIZE] = self._transposeSizeToUser(res)
 
             pxs = sensorinfo.wPixelSize * 1e-8  # m
             self.pixelSize = model.VigilantAttribute(self._transposeSizeToUser((pxs, pxs)),
@@ -1146,10 +1139,6 @@ class Camera(model.DigitalCamera):
 
         note: This function is not supported by the camera models USB 3 uEye XC and XS
         """
-        # example from _setExposureTime()
-        # exp = self.SetExposure(exp)  # can take ~2s
-        # logging.debug("Updated exposure time to %g s", exp)
-
         # sets the position and size of the image by using an object of the IS_RECT type.
         rect_aoi = IS_RECT()
         # make sure the AOI rect region is centered compared to the sensor resolution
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -53,6 +54,7 @@ magic.odm.yaml
 .project
 .pydevproject
 .settings
+/install/linux/usr/share/odemis/sim/meteor-reverse-sim.odm.yaml
```
pieleric, line 57: Let's drop the changes on .gitignore. The `htmlcov/` should already be there.

src/odemis/driver/ueye.py
```diff
@@ -646,14 +679,20 @@ def __getitem__(self, name):
...
+    Represents an IDS uEye camera.
     Currently, only greyscale mode is supported
     """
 
-    def __init__(self, name, role, device=None, **kwargs):
+    def __init__(self, name, role, device=None, max_res=None, **kwargs):
         """
         device (None or str): serial number (eg, 1020345) of the device to use
-          or None if any device is fine.
+        or None if any device is fine.
```
pieleric, line 689: Typically, in doctrings, we use two spaces indentation to indicate that it continues the line before.

src/odemis/driver/ueye.py
```diff
@@ -646,14 +679,20 @@ def __getitem__(self, name):
...
+
+        max_res (1 <= int, 1 <= int): maximum resolution possible.
```
pieleric, line 691: `((1 <= int, 1 <= int) or None)`

src/odemis/driver/ueye.py
```diff
@@ -1078,6 +1134,36 @@ def _buffer_as_array(self, mem, md):
...
+                self.rect_aoi.s32Width = max_res[0]
+                self.rect_aoi.s32Height = max_res[1]
+        else:
+            self.rect_aoi.s32X = (sensor_res[0] - max_res[0]) // 2
+            self.rect_aoi.s32Y = (sensor_res[1] - max_res[1]) // 2
+            self.rect_aoi.s32Width = max_res[0]
+            self.rect_aoi.s32Height = max_res[1]
+
+        # TODO use the return value to see if the command was successful??
+        self._dll.is_AOI(self._hcam, AOI_IMAGE_SET_AOI, byref(self.rect_aoi), sizeof(self.rect_aoi))
```
pieleric, line 1164: Please wrap this into a separate function `SetAOI(rect: IS_RECT)`. This way it's clearer what it does. You can place that function just after `SetFrameRate()`.

(and you can also place the `logging.debug()` there, with the content of rect.

src/odemis/driver/ueye.py
```diff
@@ -1078,6 +1134,36 @@ def _buffer_as_array(self, mem, md):
...
+                self.rect_aoi.s32Y = (sensor_res[1] - max_res[1]) // 2
+                self.rect_aoi.s32Width = max_res[0]
+                self.rect_aoi.s32Height = max_res[1]
+        else:
+            self.rect_aoi.s32X = (sensor_res[0] - max_res[0]) // 2
+            self.rect_aoi.s32Y = (sensor_res[1] - max_res[1]) // 2
+            self.rect_aoi.s32Width = max_res[0]
+            self.rect_aoi.s32Height = max_res[1]
+
+        # TODO use the return value to see if the command was successful??
```
pieleric, line 1163: No need. The UEyeDLL already takes care of checking the return value and convert it to a Exception (UEyeError).

src/odemis/driver/ueye.py
```diff
@@ -678,8 +717,25 @@ def __init__(self, name, role, device=None, **kwargs):
...
+            else:
+                max_res = tuple(max_res)
+            if not all(1 <= mr <= r for mr, r in zip(max_res, sensor_res_user)):
+                raise ValueError(f"max_res has to be between (1, 1) and {sensor_res_user}, but got {max_res}")
+            max_res_hw = self._transposeSizeFromUser(max_res)
+
+            # setting a specific Area Of Interest will force image cropping
+            if sensor_res_user != max_res:
+                self.rect_aoi = IS_RECT()  # create the RECT structure for setting the AOI
+                self._set_aoi(sensor_res_user, max_res)
```
pieleric, line 736: You shouldn't pass "user" resolution, but the hardware resolution. That's already converted/transposed in the proper format.
I'd suggest to directly compute rect_aoi here, and call SetAOI(). Something like that:
```python
            rect_aoi = IS_RECT() 
            rect_aoi.s32X = (self._sensor_res[0] - max_res_hw[0]) // 2
            rect_aoi.s32Y = (self._sensor_res[1] - max_res_hw[1]) // 2
            rect_aoi.s32Width = max_res_hw[0]
            rect_aoi.s32Height = max_res_hw[1]
            self.SetAOI(rect_aoi)
```

You're done, you can drop the complex `_set_aoi()`!

src/odemis/driver/ueye.py
```diff
@@ -1078,6 +1134,36 @@ def _buffer_as_array(self, mem, md):
 
         return model.DataArray(na, md)
 
+    def _set_aoi(self, sensor_res: tuple, max_res: tuple):
+        """
+        :param sensor_res: tuple(int width, int height) The maximum resolution of the camera sensor
+        :param max_res: tuple(int width, int height) The maximum requested resolution of the image
+
+        note: This function is not supported by the camera models USB 3 uEye XC and XS
```
pieleric, line 1142: Probably worthy to keep as a comment in SetAOI.

pieleric (changes requested):

src/odemis/driver/test/ueye_test.py
```diff
@@ -37,7 +38,7 @@
 TEST_NOHW = (os.environ.get("TEST_NOHW", "0") != "0")  # Default to Hw testing
 
 CLASS = ueye.Camera
-KWARGS = dict(name="camera", role="ccd", device=None, transp=[2, -1])
+KWARGS = dict(name="camera", role="ccd", device=None, max_res=None, transp=None)
```
pieleric, line 41: I think you could leave the transp as is. It's always good to test with `transp`, it finds extra bugs ;-)

src/odemis/driver/test/ueye_test.py
```diff
@@ -97,5 +98,47 @@ def setUp(self):
...
+    camera_type = CLASS
+    camera_kwargs = KWARGS
+
+    @classmethod
+    def setUpClass(cls):
+        if TEST_NOHW:
+            raise unittest.SkipTest('No camera HW present. Skipping tests.')
+
+        cls.camera_kwargs['max_res'] = (1024, 768)
+        # apply a transpose of 90 CW and mirror ax-0
```
pieleric, line 111: No, this just applies a 90° CW rotation. (no extra mirroring)

src/odemis/driver/test/ueye_test.py
```diff
@@ -97,5 +98,47 @@ def setUp(self):
...
+
+        # check if the resolution is not the same own cam res should be higher
+        self.assertNotEqual(self.camera_kwargs['max_res'], cam_resolution)
+        # check if the transpose parameter is applied and axes are switched
+        self.assertEqual((768, 1024), trans_resolution)
+        # check is the AOI is set with the right x, y, height and width values
+        self.assertEqual(self.camera.rect_aoi.s32X, 128)
+        self.assertEqual(self.camera.rect_aoi.s32Y, 128)
+        self.assertEqual(self.camera.rect_aoi.s32Width, 1024)
+        self.assertEqual(self.camera.rect_aoi.s32Height, 768)
```
pieleric, line 140: No need to check that. It's something private to the driver (and I've just asked you to remove it ;-) ).
However, you should acquire some images, and check they are the expected resolution.
See `self.camera.data.get()`.
Afterwards, set the binning to (for example) "2, 4", and check that the resolution of the returned image is X/2 and Y/4.

pieleric (changes requested):

.gitignore
sneepstefan: The PEP 8 standard for docstrings refers to PEP 257 and in there it describes multiline docstrings with an example:
def bar():
    """
    A multi-line
    docstring.
    """
So I am a little confused. In my mind I wanted to start that line just below the first part with the same indentation.
If you say we always have to use 2 spaces in this case, we might also add that to Karishma's PR for coding guidelines.

src/odemis/driver/ueye.py
sneepstefan: The PEP 8 standard for docstrings refers to PEP 257 and in there it describes multiline docstrings with an example:
def bar():
    """
    A multi-line
    docstring.
    """
So I am a little confused. In my mind I wanted to start that line just below the first part with the same indentation.
If you say we always have to use 2 spaces in this case, we might also add that to Karishma's PR for coding guidelines.

src/odemis/driver/ueye.py
```diff
@@ -678,8 +715,27 @@ def __init__(self, name, role, device=None, **kwargs):
...
+            max_res_hw = self._transposeSizeFromUser(max_res)
+
+            # compare the max_res parameter and the resolution of the camera sensor
+            # if the max_res param is smaller image cropping will be set through AOI
+            rect_aoi = IS_RECT()
+            rect_aoi.s32X = (self._sensor_res[0] - max_res_hw[0]) // 2
+            rect_aoi.s32Y = (self._sensor_res[1] - max_res_hw[1]) // 2
+            rect_aoi.s32Width = max_res_hw[0]
+            rect_aoi.s32Height = max_res_hw[1]
+            self.SetAOI(rect_aoi)
```
pieleric, line 736: Can you place this in a `if` block to only do it if it's needed. I'm a little afraid to break cases if SetAOI() raises an error even if passing the full resolution. (something like `if self._sensor_res != max_res_hw`).

src/odemis/driver/ueye.py
```diff
@@ -1078,6 +1146,35 @@ def _buffer_as_array(self, mem, md):
 
         return model.DataArray(na, md)
 
+    def _set_aoi(self, sensor_res: tuple, max_res: tuple):
+        """
```
pieleric, line 1150: This whole function can go away, right?

pieleric (changes requested):

src/odemis/driver/ueye.py
pieleric: I think that is more "my" style than the official one, indeed. I used to directly the parameter name, without `:param `, hence it was handy to have some indentation to separate the parameters.

nandishjpatel (approved):

src/odemis/driver/test/ueye_test.py
```diff
@@ -97,5 +98,63 @@ def setUp(self):
...
+    def tearDownClass(cls):
+        if TEST_NOHW:
+            return
+        super(TestUEyeCroppingTranspose, cls).tearDownClass()
+
+    def setUp(self):
+        if TEST_NOHW:
+            self.skipTest("No simulator available")
+
+        VirtualTestCam.setUp(self)
```
pieleric, line 129: You don't need these 2 overrides if you already have the SkipTest in the setUpClass(), or do you?

src/odemis/driver/test/ueye_test.py
pieleric: Please, leave `transp=[2, -1]` always there.

src/odemis/driver/test/ueye_test.py
```diff
@@ -97,5 +98,63 @@ def setUp(self):
...
+
+    def test_set_resolution(self):
+        """
+        Test if the max resolution and the transpose argument are passed and applied.
+        Also test if an acquired image is of the correct dimensions and the axes are right also with binning active.
+        """
+        # assign the right resolutions from the initialization of the camera class
+        cam_resolution = self.camera._sensor_res
+        trans_resolution = (self.camera_kwargs['max_res'][1], self.camera_kwargs['max_res'][0])
+        set_resolution = self.camera._metadata[model.MD_SENSOR_SIZE]
```
pieleric, line 139: Don't use `_metadata` but the official `getMetadata()`:
```python
set_resolution = self.camera.getMetadata()[model.MD_SENSOR_SIZE]
```

src/odemis/driver/test/ueye_test.py
```diff
@@ -97,5 +98,63 @@ def setUp(self):
...
+
+        VirtualTestCam.setUp(self)
+
+    def test_set_resolution(self):
+        """
+        Test if the max resolution and the transpose argument are passed and applied.
+        Also test if an acquired image is of the correct dimensions and the axes are right also with binning active.
+        """
+        # assign the right resolutions from the initialization of the camera class
+        cam_resolution = self.camera._sensor_res
```
pieleric, line 137: I'm a bit confused by the names of these 3 resolutions. I'd suggest:
* cam_resolution -> sensor_res_raw
* trans_resolution -> expected_res  (and don't swap it!, just self.camera_kwargs['max_res'])
* set_resolution -> sensor_res_user

src/odemis/driver/test/ueye_test.py
```diff
@@ -97,5 +98,63 @@ def setUp(self):
...
+        # check if the resolution is not the same, own cam res should be higher
+        self.assertNotEqual(self.camera_kwargs['max_res'], cam_resolution)
+        # check if the transpose parameter is applied and axes are switched
+        self.assertEqual(cam_resolution, tuple(reversed(set_resolution)))
+
+        # be sure to have binning set to base value
+        self.camera.binning.value = (1, 1)
+        # check if the acquired image is of the right size
+        im = self.camera.data.get()
+        self.assertEqual(im.shape, trans_resolution)
```
pieleric, line 150: The shape of a numpy array is always reversed compared to the user resolution, because the numpy array is encoded as YX (that's numpy's convention!).
So it's expected that the shape is reversed. I'd suggest that instead of `trans_resolution`, you use the expected resolution, compare like that:
```
self.assertEqual(im.shape, expected_res[::-1])
```

src/odemis/driver/test/ueye_test.py
```diff
@@ -97,5 +98,63 @@ def setUp(self):
...
+        trans_resolution = (self.camera_kwargs['max_res'][1], self.camera_kwargs['max_res'][0])
+        set_resolution = self.camera._metadata[model.MD_SENSOR_SIZE]
+
+        # check if the resolution is not the same, own cam res should be higher
+        self.assertNotEqual(self.camera_kwargs['max_res'], cam_resolution)
+        # check if the transpose parameter is applied and axes are switched
+        self.assertEqual(cam_resolution, tuple(reversed(set_resolution)))
+
+        # be sure to have binning set to base value
+        self.camera.binning.value = (1, 1)
```
pieleric, line 147: Please first check that the resolution == expected_res:
```
self.assertEqual(expected_res, self.camera.resolution.value)
```

src/odemis/driver/test/ueye_test.py
```diff
@@ -97,5 +98,63 @@ def setUp(self):
...
+        """
+        # assign the right resolutions from the initialization of the camera class
+        cam_resolution = self.camera._sensor_res
+        trans_resolution = (self.camera_kwargs['max_res'][1], self.camera_kwargs['max_res'][0])
+        set_resolution = self.camera._metadata[model.MD_SENSOR_SIZE]
+
+        # check if the resolution is not the same, own cam res should be higher
+        self.assertNotEqual(self.camera_kwargs['max_res'], cam_resolution)
+        # check if the transpose parameter is applied and axes are switched
+        self.assertEqual(cam_resolution, tuple(reversed(set_resolution)))
```
pieleric, line 144: Instead of `tuple(reversed())`, you  should be able to just write `...[::-1]`.

src/odemis/driver/test/ueye_test.py
```diff
@@ -97,5 +98,63 @@ def setUp(self):
...
+    camera_type = CLASS
+    camera_kwargs = KWARGS
+
+    @classmethod
+    def setUpClass(cls):
+        if TEST_NOHW:
+            raise unittest.SkipTest('No camera HW present. Skipping tests.')
+
+        # set a fixed max_res smaller than the maximum supported camera resolution
+        cls.camera_kwargs['max_res'] = (1000, 1000)
```
pieleric, line 114: Please pick two different values for X  and Y. Otherwise, you don't notice if the transp is used or not. => 1000, 896.

pieleric (changes requested):

pieleric (approved):

PR #2271, opened 2022-12-01
Fix for passing z_localization_test.py

K4rishma:
This is the fix for the Jira issue [SWM-60](https://delmic.atlassian.net/browse/SWM-60?atlOrigin=eyJpIjoiYWRkZGUzMDM2YTNjNDQ2OTg5ODBkM2IyYTlkYjE1OTMiLCJwIjoiaiJ9) which is due to the import error for psf_extractor module.  

The psf_extractor works with numpy version 1.20.3.


tepals (approved):

requirements.txt
```diff
@@ -4,7 +4,7 @@ future>=0.16.0
 h5py>=2.5.0
 pylibtiff>=0.4.2
 matplotlib>=1.4.3,<3.0
-numpy>=1.11.0,<1.18
+numpy>=1.11.0,<1.26
```
pieleric, line 7: This file is not really magical. It's not going to get a different numpy version installed on the testing computer just by upping the version number. We haven't tested Odemis with numpy > 1.18, so I'm a little hesitant. Note that Ubuntu 22.04 ships with numpy 1.21, so we should test with such version soon, in any case!

It's actually possible to install numpy 1.20 without changing this file. So I'd suggest to just fix the installation document only.

pieleric (changes requested):

PR #2272, opened 2022-12-01
Swm 63 test reference cancel failing

nandishjpatel:
Ticket: https://delmic.atlassian.net/browse/SWM-63


src/odemis/driver/smaract.py
```diff
@@ -4870,6 +4870,9 @@ def SA_SI_Cancel(self, id):
         if self.waiting:
             self.cancel_referencing = True
             self.cancel_event = True
+            # Wait for the future to be finished
+            self.executor.cancel()
+
```
pieleric, line 4875: No need for an extra line.

pieleric (changes requested):

src/odemis/driver/smaract.py
nandishjpatel, line 4875:
```suggestion
```

pieleric (approved):
@nandishjpatel , could you squash your first and third commits? Then we can merge this PR :-)

PR #2274, opened 2022-12-01
gui test: skip tests using UIActionSimulator on wxPython 4.0.7

pieleric:
On wxPython 4.0.7, the simulator seems to only work when it's run from
within Eclipse. From the terminal the tests always fail, because the
simulator doesn't actually do anything.

Should be working when switching to wxPython 4.1

### DIFF
--- a/src/odemis/gui/test/comp_text_test.py
+++ b/src/odemis/gui/test/comp_text_test.py
@@ -24,11 +24,13 @@ Odemis. If not, see http://www.gnu.org/licenses/.
 # Test module for Odemis' gui.comp.text module
 # ===============================================================================
 import locale
+import logging
 import unittest
-import wx
-import odemis.gui.test as test
-from odemis.gui.comp.text import FloatTextCtrl, IntegerTextCtrl, UnitFloatCtrl, UnitIntegerCtrl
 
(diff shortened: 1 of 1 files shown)


src/odemis/gui/test/comp_text_test.py
```diff
@@ -88,13 +90,22 @@ def test_unit_float(self):
...
+        wx_ver = tuple(int(v) for v in wx.__version__.split("."))
+        if wx_ver < (4, 1, 0):
+            self.skipTest("Test case is known to fail on wxPython < 4.1 due to buggy UIActionSimulator")
+
     def test_int_txt_ctrl(self):
 
         ctrl = IntegerTextCtrl(self.panel, value=123456789)
         self.add_control(ctrl, label=ctrl.__class__.__name__, flags=wx.EXPAND | wx.ALL)
 
-        test.goto_manual()
```
tepals, line 96: why did you remove this?

pieleric: It's already done for the whole file at line 35. So this line has no effect.
Actually, I think we could probably change the `gui.test` so that this function is automatically done whenever it's imported (it just read an environment variable). But that's for another PR.

PR #2276, opened 2022-12-02
fix failing find_hole_center and find_lens_center testcases for the Delphi

sneepstefan:
These testcases were failing in a Ubuntu 20.04 environment probably due to opencv/cv2 version difference.
Fixed the testcases by adding opencv version checks. For test_find_hole_center it was required to add a different outcome.
For test_find_lens_center only setting a cv2 version boolean upfront was enough.

### DIFF
--- a/src/odemis/acq/align/test/delphi_test.py
+++ b/src/odemis/acq/align/test/delphi_test.py
@@ -20,6 +20,7 @@ You should have received a copy of the GNU General Public License along with
 Odemis. If not, see http://www.gnu.org/licenses/.
 '''
 import logging
+import math
 import numpy
 from odemis import model
 import odemis
@@ -104,8 +105,15 @@ class TestCalibration(unittest.TestCase):
         data[0].shape = Y, X
 
         hole_coordinates = delphi.FindCircleCenter(data[0], 0.02032, 6, darkest=True)
-        expected_coordinates = (0.0052705, -0.0018415)  # (391.5, 257.5) px
-        numpy.testing.assert_almost_equal(hole_coordinates, expected_coordinates)
+
+        # set fixed expected coordinates for the center also set a fixed difference
+        max_dist = 0.7e-3
+        expected_coordinates = (0.0050, -0.0024)
+
+        # measure the distance between the two centers and see if distance is less than the set max distance
+        dist = math.hypot(hole_coordinates[0] - expected_coordinates[0],
+                          hole_coordinates[1] - expected_coordinates[1])
+        self.assertLessEqual(dist, max_dist)
 
     @unittest.expectedFailure
     def test_find_sh_hole_center(self):
@@ -127,8 +135,6 @@ class TestCalibration(unittest.TestCase):
         data = hdf5.read_data(os.path.join(TEST_IMAGE_PATH, "navcam-calib2.h5"))
         imgs = img.RGB2Greyscale(img.ensureYXC(data[0]))
 
-        # lens_coordinates = delphi.FindCircleCenter(data[0][0], delphi.LENS_RADIUS, 5)
-        # expected_coordinates = (-5.9703947e-05, 1.5257675e-04)  # (451.5, 445.5) px
         lens_coordinates = delphi.FindRingCenter(imgs)
         expected_coordinates = (-1.6584835e-05, 1.3084411e-04)  # 454.75, 446.1) px
         numpy.testing.assert_almost_equal(lens_coordinates, expected_coordinates)
--- a/src/odemis/acq/align/delphi.py
+++ b/src/odemis/acq/align/delphi.py
@@ -1214,13 +1214,11 @@ def FindRingCenter(image):
     edge_image = (image > threshold).astype(numpy.uint8) * 255
     edge_image = cv2.medianBlur(edge_image, 5)
 
-    # tiff.export("test_contour.tiff", model.DataArray(edge_image))
-
     # Convert the edges into points
-    if opencv_v2:
-        contours, _ = cv2.findContours(edge_image, cv2.RETR_TREE, cv2.CHAIN_APPROX_NONE)
-    else:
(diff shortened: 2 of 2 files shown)


src/odemis/acq/align/test/delphi_test.py
```diff
@@ -127,8 +133,10 @@ def test_find_lens_center(self):
         data = hdf5.read_data(os.path.join(TEST_IMAGE_PATH, "navcam-calib2.h5"))
         imgs = img.RGB2Greyscale(img.ensureYXC(data[0]))
 
-        # lens_coordinates = delphi.FindCircleCenter(data[0][0], delphi.LENS_RADIUS, 5)
-        # expected_coordinates = (-5.9703947e-05, 1.5257675e-04)  # (451.5, 445.5) px
+        # set opencv_v2 before the FindRingCenter to force unpacking findContours into 2 parameters
+        if delphi.cv2.__version__ != "3.2":
+            delphi.opencv_v2 = True
```
tepals, line 138: It doesn't look like good idea to update `delphi.opencv_v2` here, this might interfere with other test cases. There is actually also a nicer solution, that will require you to update the code in FindRingCenter.
As explained in the link below you can adjust the `findContours` code in `FindRingCenter` to work with any version of opencv. 
https://stackoverflow.com/a/56142875

you would need to update these lines
```
    # Convert the edges into points
    if opencv_v2:
        contours, _ = cv2.findContours(edge_image, cv2.RETR_TREE, cv2.CHAIN_APPROX_NONE)
    else:
        _, contours, _ = cv2.findContours(edge_image, cv2.RETR_TREE, cv2.CHAIN_APPROX_NONE)
```
to:
```
# Convert the edges into points
contours, _ = cv2.findContours(edge_image, cv2.RETR_TREE, cv2.CHAIN_APPROX_NONE)[-2:]
```

tepals (changes requested):

src/odemis/acq/align/test/delphi_test.py
```diff
@@ -104,7 +104,13 @@ def test_find_hole_center(self):
         data[0].shape = Y, X
 
         hole_coordinates = delphi.FindCircleCenter(data[0], 0.02032, 6, darkest=True)
-        expected_coordinates = (0.0052705, -0.0018415)  # (391.5, 257.5) px
+
+        # depending on the cv2 version which is related to the Ubuntu version the expected outcome is different
+        if delphi.cv2.__version__ == "3.2":
```
tepals, line 109: This is not going to work, `delphi.cv2.__version__` will most likely return something like `3.2.x` so you want to check that the first two values are 3.2

K4rishma (approved):

src/odemis/acq/align/test/delphi_test.py
```diff
@@ -104,7 +104,13 @@ def test_find_hole_center(self):
         data[0].shape = Y, X
 
         hole_coordinates = delphi.FindCircleCenter(data[0], 0.02032, 6, darkest=True)
-        expected_coordinates = (0.0052705, -0.0018415)  # (391.5, 257.5) px
+
+        # depending on the cv2 version which is related to the Ubuntu version the expected outcome is different
+        if delphi.cv2.__version__[:3] == "3.2":
```
tepals, line 109: a bit nit-picky but this will also return True for version 3.21, you can do something like:
```
if ".".join(delphi.cv2.__version__.split(".")[:2]) == "3.2"
```

tepals (approved):

src/odemis/acq/align/test/delphi_test.py
```diff
@@ -104,7 +104,13 @@ def test_find_hole_center(self):
         data[0].shape = Y, X
 
         hole_coordinates = delphi.FindCircleCenter(data[0], 0.02032, 6, darkest=True)
-        expected_coordinates = (0.0052705, -0.0018415)  # (391.5, 257.5) px
+
+        # depending on the cv2 version which is related to the Ubuntu version the expected outcome is different
+        if ".".join(delphi.cv2.__version__.split(".")[:2]) == "3.2":
+            expected_coordinates = (0.0052705, -0.0018415)  # (391.5, 257.5) px
+        else:
+            expected_coordinates = (0.0046355, -0.0028575)  # (386.5, 265.5) px
```
pieleric, line 112: If  both results are correct, because we don't need to be very precise, then I would suggest to explicitly check that the returned coordinates are "not too far" from the expected center. Hardcoding the values to 2 very precise tuples is confusing. Please have one "expected coordinates" and check the distance of the measured coordinates is less than N meters (where you have to pick N a small constant).

pieleric (changes requested):

src/odemis/acq/align/test/delphi_test.py
```diff
@@ -104,8 +104,18 @@ def test_find_hole_center(self):
...
+        const_diff = 6e-4
+        expected_coordinates = (0.0050, -0.0024)
+
+        # check if the outcome is between 0.0044 and 0.0056 for the first element
+        self.assertGreaterEqual(hole_coordinates[0], expected_coordinates[0] - const_diff)
+        self.assertLessEqual(hole_coordinates[0], expected_coordinates[0] + const_diff)
+
+        # check if the outcome is between -0.0030 and -0.0018 for the second element
+        self.assertGreaterEqual(hole_coordinates[1], expected_coordinates[1] - const_diff)
+        self.assertLessEqual(hole_coordinates[1], expected_coordinates[1] + const_diff)
```
pieleric, line 118: You can measure the distance between the two centers, to be a little bit more concise.
```suggestion
        # set fixed expected coordinates for the center also set a fixed difference
        max_dist = 0.5e-3 # m
        expected_coordinates = (0.0050, -0.0024)

        # check if the outcome is between 0.0044 and 0.0056 for the first element
        dist = math.hypot(hole_coordinates[0] - expected_coordinates[0],
                           hole_coordinates[1] - expected_coordinates[1])
        self.assertLessEqual(dist, max_dist)
```

pieleric (changes requested):

pieleric (approved):

PR #2278, opened 2022-12-05
odemisd modelgen: fix YAML parsing when a dict has multiple !extend

pieleric:
The code to replace !extend by the content of the file referenced worked fine...
as long as there was only one extend. With more than one extend, the
wrong keys would be modified, and some "!extend" would be left, causing
some really confusing errors.


nandishjpatel (approved):
Looks good to me!

PR #2279, opened 2022-12-05
Create a FIB drift correction testing script

K4rishma:
Completes two related/overlapping tasks as follows
1) [Create a simulator of the FIB that has drift correction and beam shift](https://delmic.atlassian.net/browse/MIMAS-909?atlOrigin=eyJpIjoiOGU5NjE5NDM1MjM0NDgyYTlmODZlYzJkNDQ0NTYwMDMiLCJwIjoiaiJ9)
2) [Implement example script drift correction](https://delmic.atlassian.net/browse/MIMAS-910?atlOrigin=eyJpIjoiMTkwMDg1ZjU3NDU5NDU5Y2IxMGU3YTVlZTg5NTU5ZTAiLCJwIjoiaiJ9)

@pieleric  this is a draft pull request. My questions are added as comments. After this request is approved, I will create the real PR where there will be two separate commits. Changes to simsem driver will be there in the second commit.


scripts/drift_corr_ion.py
```diff
@@ -0,0 +1,155 @@
+"""
+Created on
+
+@author:
+
+Copyright © , Delmic
```
K4rishma, line 6: Hello Eric do I need to add these details? If yes, what would be the entries here?

src/odemis/driver/simsem.py
```diff
@@ -143,7 +143,7 @@ def __init__(self, name, role, parent, aperture=100e-6, wd=10e-3, **kwargs):
 
         # pixelSize is the same as MD_PIXEL_SIZE, with scale == 1
         # == smallest size/ between two different ebeam positions
-        pxs = fake_img.metadata[model.MD_PIXEL_SIZE]
+        pxs = fake_img.metadata.get(model.MD_PIXEL_SIZE, (1e-6, 1e-6))
```
K4rishma, line 146: I will create a separate commit in the actual PR for the changes to simsem driver.

scripts/drift_corr_ion.py
pieleric: Yes, put your name there :-)

src/odemis/acq/drift/__init__.py
```diff
@@ -244,20 +249,23 @@ def estimateCorrectionPeriod(period, t, repetitions):
 
         return itertools.cycle(acq_dc_period)
 
-    def _updateSEMSettings(self):
+    def _updateBeamSettings(self):
```
pieleric, line 252: Indeed, "SEM" is not well fitting anymore. However, "beam" usually refers to settings like voltage, current, etc. So I'd suggest `_updateScannerSettings()`.

src/odemis/acq/align/shift.py
```diff
@@ -19,167 +19,158 @@
...
+    by reducing the amplitude of noise and enhancing the shift amplitude. The coordinates
+    of the maximum of the cross-correlation define the shift vector between the two images.
+    The implementation of precision > 1 is based on the "Efficient subpixel image registration by
     cross-correlation" by Manuel Guizar, for the corresponding matlab code see
     http://www.mathworks.com/matlabcentral/fileexchange/
     18401-efficient-subpixel-image-registration-by-cross-correlation.
 
+    For e.g.
+    When the current image moves to top right corner relatively to the previous image.
+    The returned shift value is positive in both x and y respectively.
```
pieleric, line 104: Are you sure about this? Normally the convention is that X,Y positive (in pixels) mean a move towards *bottom* right. I looked quickly at the caller code and it seems to follow that convention...

src/odemis/acq/align/shift.py
```diff
@@ -19,167 +19,158 @@
...
-                                (arange(0, noc) - coff)[:, None]
-                       )
+        maxima = numpy.stack(maxima).astype(float_dtype, copy=False)
+        maxima -= dftshift
 
-    kernr = numpy.power(math.e, (-z * 2 * math.pi / (nr * precision)) *
-                                (fft.ifftshift(arange(0, nr))[:, None] - nr // 2) *
-                                ((arange(0, nor)[:, None]).T - roff)
-                       )
+        shifts += maxima / precision_factor
```
pieleric, line 174: This whole block about the maximum location should be also correct for the precision == 1, right? You'd just need to have  `precision_factor=1`. Then you could drop the special "locate peak" part of precision == 1, and always use this code.

src/odemis/acq/align/shift.py
```diff
@@ -19,167 +19,158 @@
...
-    of upsampled pixels
-    precision (int): Calculate drift within 1/precision of a pixel
-    roff, coff (ints): Row and column offsets, allow to shift the output array
-                    to a region of interest on the DFT
-    returns (tuple of floats): Drift in pixels
-    """
-    z = 1j  # imaginary unit
-    nr, nc = data.shape
+    else:
+        precision_factor = numpy.array(precision, dtype=float_dtype)
```
pieleric, line 155: Can you add a short comment explaining the trick of this line? It seems to just take the Python float and convert it to a numpy float of the "right" type. What's the advantage? Wouldn't numpy do the right thing otherwise?(typically numpy is fine with doing numpy.array * python_float)

src/odemis/acq/align/shift.py
```diff
@@ -19,167 +19,158 @@
...
+    if cloc > nd2:
+        col_shift = cloc - n
+    else:
+        col_shift = cloc
 
-    return col_shift, row_shift
+    shifts = [row_shift, col_shift]
 
+    if precision == 1:
+        pass
```
pieleric, line 152: You can just write `if precision > 1` (no `else`)

src/odemis/acq/align/shift.py
```diff
@@ -19,167 +19,158 @@
...
-        if cloc > nd2:
-            col_shift = cloc - n
-        else:
-            col_shift = cloc
-
+    image_product = previous_fft * current_fft.conj()
+
+    # Cross-correlation computation
+    eps = numpy.finfo(image_product.real.dtype).eps
+    image_product /= numpy.maximum(numpy.abs(image_product), 100 * eps)
```
pieleric, line 124: Please explain the magical part of this block, or at least put a reference to where this can be explained.

pieleric (changes requested):

src/odemis/acq/align/shift.py
K4rishma: Sorry, When the current image moves to **top left corner** relatively to the previous image.
    The returned shift value is positive in both x and y respectively. It is when object in the second image moves to top left corner w.r.t to first image, shift in col and row is measured in positive sign. Do you see the same way?

install/linux/usr/share/odemis/sim/fib-sim.odm.yaml
```diff
@@ -0,0 +1,234 @@
...
+#        slits: {
+#            1: "slit-in", # used to be called "slit"
+#        }
+#    },
+#    affects: ["Spectral Camera", "Spectrometer"],
+#}
+
+"Mirror Actuators": {
+    class: pigcs.FakeBus,
+    role: mirror,
```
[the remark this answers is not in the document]
K4rishma: I could not file the exact file mentioned, but I think i need to add this 
```suggestion
   SimFIB: {
    class: Microscope,
    role: sem,
    children: ["SEM E-beam", "SEM Detector", "Sample Stage"],
}
```

src/odemis/acq/align/shift.py
```diff
@@ -19,167 +19,158 @@
...
     else:
-        mlarge, nlarge = m * 2, n * 2
-
-        # Upsample by factor of 2 to obtain initial estimation and
-        # embed Fourier data in a 2x larger array
-        CC = numpy.zeros((mlarge, nlarge), dtype=numpy.complex)
-        CC[m - m // 2:m + 1 + (m - 1) // 2,
-           n - n // 2:n + 1 + (n - 1) // 2] = (fft.fftshift(previous_fft) *
-                                               fft.fftshift(current_fft).conj()
-                                              )
```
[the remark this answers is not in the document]
K4rishma: I think this step in the earlier version was only done to calculate shift. As the shift for precision 1 is already calculated we did not need this code in the old/earlier version. This calculation seems redundant to me.

src/odemis/driver/simsem.py
K4rishma: I will create a new PR for this

src/odemis/acq/align/shift.py
K4rishma: I have removed this line as Python does convert the result between float and int to float. I also do not think this line is necessary. Probably in other languages like C this is an important step. I have tested without it  and it works similar.

K4rishma:
@pieleric  I have worked on the requested changes.

FCanberk (approved):

src/odemis/acq/drift/__init__.py
```diff
@@ -44,23 +44,32 @@ class AnchoredEstimator(object):
...
     """
-    def __init__(self, scanner, detector, region, dwell_time):
+
+    def __init__(self, scanner, detector, region, dwell_time, max_pixels=MAX_PIXELS, follow_drift=True):
         """
         scanner (Emitter)
         detector (Detector)
         region (4 floats)
         dwell_time (float)
+        #TOdo
```
pieleric, line 54: Drop the "#TODO" line

src/odemis/acq/drift/__init__.py
```diff
@@ -143,7 +152,7 @@ def acquire(self):
                 self.raw = self.raw[0:2]
             self.raw.append(data)
         finally:
-            # Restore SEM settings
+            # Restore beam settings
```
pieleric, line 155: Let's call that "scanner settings".

src/odemis/acq/drift/__init__.py
```diff
@@ -44,23 +44,32 @@ class AnchoredEstimator(object):
...
         detector (Detector)
         region (4 floats)
         dwell_time (float)
+        #TOdo
+        max_pixels (int): the maximum number of pixels that the anchor region will use. If the region
+          acquired is large, then the scale is adjusted to maximum pixels
+        follow_drift (bool): If True, the anchor region is adjusted based on drift measure. It is useful with drift
+          compensation is based on updating anchor region. If False, the anchor
+          region is acquired in the same location. It is useful when drift compensation is based on
+          beam shift or stage movement
```
pieleric, line 60: I'd slightly adjust like the following:
`If True, the anchor region position is adjusted based on the drift measured. It is useful when drift
 compensation is done by adjusting the scanner settings. If False, the anchor
 region is fixed. It is useful when drift compensation is based on beam shift or stage movement.`

src/odemis/acq/drift/__init__.py
```diff
@@ -44,23 +44,32 @@ class AnchoredEstimator(object):
...
+
+    def __init__(self, scanner, detector, region, dwell_time, max_pixels=MAX_PIXELS, follow_drift=True):
         """
         scanner (Emitter)
         detector (Detector)
         region (4 floats)
         dwell_time (float)
+        #TOdo
+        max_pixels (int): the maximum number of pixels that the anchor region will use. If the region
+          acquired is large, then the scale is adjusted to maximum pixels
```
pieleric, line 56: "then the scale is adjusted to maximum pixels" -> "then the scanner scale is adjusted so that the anchor region contains at most max_pixels pixels."

scripts/drift_corr_ion.py
```diff
@@ -0,0 +1,163 @@
...
+    p = Path(options.base)
+    nb_images = options.nb_images
+    delay_time = options.delay_time
+    dwell_time = options.dwell_time  # s
+    drift_est = drift.AnchoredEstimator(escan, detector,
+                                        roi,
+                                        dwell_time, max_pixels=512 ** 2, follow_drift=False)
+    logging.info('Dwell time(s): %s', dwell_time)
+    logging.info('roi: %s', roi)
+    logging.info('Hfov(m): %s', escan.horizontalFoV.value)
```
pieleric, line 108: Usually it's abbreviated as "HFW".

scripts/drift_corr_ion.py
```diff
@@ -0,0 +1,163 @@
...
+    for i in range(nb_images):
+        logging.debug("IMAGE %s", i)
+
+        # Acquire an image at the given location (RoI)
+        drift_est.acquire()
+        da = drift_est.raw[-1]
+
+        # Save the Acquired Image
+        # (For debugging purposes, have an option to store the image, with current date and time in the filename)
+        timestr = time.strftime("%m%d-%H%M") + "_" + str(i)
```
pieleric, line 129: Just compute it once before the loop. Otherise, as the time changes, the base name changes, which is inconvenient.

scripts/drift_corr_ion.py
```diff
@@ -0,0 +1,163 @@
...
+        filename = "{0}{2}{1}".format(p.stem, p.suffix, timestr)
```
pieleric, line 130: It'd be easier to read if you ordered it 0, 1, 2. If you use the f-string, it'll be even easier to read.

scripts/drift_corr_ion.py
```diff
@@ -0,0 +1,163 @@
...
+    logging.info('file name by: %s', time.strftime("%m%d-%H%M"))
+    logging.info('Max images: %s', nb_images)
+    logging.info('Pixel size(m): %s', escan.pixelSize.value)
+
+    i = 0
+    row_drift = []
+    col_drift = []
+
+    for i in range(nb_images):
+        logging.debug("IMAGE %s", i)
```
pieleric, line 121: Could you put that whole loop in a separate function? That would clarify between the argument parsing, and the actual drift correction. The function could return the col_drift and row_drift.

pieleric (changes requested):

K4rishma:
@pieleric I have added the requested changes.

pieleric (approved):

PR #2284, opened 2022-12-12
bugreporter: add hostname as installation name

pieleric:
There is a new field on the tickets available in our osTicket database:
"installation". Let's make use of it. The hostname should normally
correspond to the installation name, so that's straightforward.

### DIFF
--- a/src/bugreporter/odemis_bugreporter.py
+++ b/src/bugreporter/odemis_bugreporter.py
@@ -187,13 +187,17 @@ class OdemisBugreporter(object):
         """
         Create ticket on osTicket server.
         :arg api_key: (String) API-Key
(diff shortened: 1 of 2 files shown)


sneepstefan:
Where exactly do we need the host name for?
And why is the attachments field overridden by this host name information is the another (easier) way to add attachments?
Just to understand the underlying motivation of the changes.

pieleric:
Our ticket database now supports storing the "installation name". To provide this information, the bug reporter uses the hostname (as at installation we explicitly set the hostname to be the installation name). 

There is no direct relationship with attachments. They are both sent. However, the test cases had a bug when calling `create_ticket()`. It also passed an "attachment" field, which was later overridden. As I had to change these calls, I took the opportunity to fix them at the same time.

PR #2286, opened 2022-12-13
fix failing test cases

FCanberk:
fixed 5 failing test cases under `class ProxyOfProxyTest(unittest.TestCase):`

set the `BASE_DIRECTORY` using `tempfile.gettempdir()` to ensure the folder exist on both windows and linux operating systems.

### DIFF
--- a/src/odemis/model/_core.py
+++ b/src/odemis/model/_core.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-'''
+"""
 Created on 18 Jun 2012
 
 @author: Éric Piel
@@ -18,18 +18,19 @@ PURPOSE. See the GNU General Public License for more details.
 
 You should have received a copy of the GNU General Public License along with
 Odemis. If not, see http://www.gnu.org/licenses/.
-'''
-import tempfile
-
-from past.builtins import basestring
-import Pyro4
-from Pyro4.core import oneway
-from collections.abc import Mapping
+"""
 import logging
 import multiprocessing
 import os
+import tempfile
 import threading
-from future.moves.urllib.parse import quote
+from collections.abc import Mapping
+from urllib.parse import quote
+
+import Pyro4
+from Pyro4.core import oneway
+from past.builtins import basestring
+
 from odemis.util import inspect_getmembers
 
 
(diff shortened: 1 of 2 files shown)


src/odemis/model/_core.py
```diff
@@ -53,14 +55,15 @@
 INIT_TIMEOUT = 300  # s
 CALL_TIMEOUT = 30  # s
 
-# TODO needs a different value on Windows
-# TODO try a user temp directory if /var/run/odemisd doesn't exist (and cannot be created)
-BASE_DIRECTORY="/var/run/odemisd"
-BASE_GROUP="odemis" # user group that is allowed to access the backend
-
+# Set the base directory regardless of the os
+BASE_DIRECTORY = tempfile.gettempdir() + "/odemisd"
```
tepals, line 59: to make it completely os independent you should use:
```
BASE_DIRECTORY = os.path.join(tempfile.gettempdir(), "odemisd")
```

src/odemis/model/_core.py
```diff
@@ -19,6 +19,8 @@
 You should have received a copy of the GNU General Public License along with
 Odemis. If not, see http://www.gnu.org/licenses/.
 '''
+import tempfile
+
 from past.builtins import basestring
 import Pyro4
```
tepals, line 25: If you want you could add another commit to update the import order according to the new coding guidelines and replace the `'''` with `"""` on line 2 and line 21

tepals (changes requested):
few small extra comments

pieleric (changes requested):

src/odemis/model/test/remote_test.py
```diff
@@ -300,7 +300,7 @@ def test_dataflow_unsub(self):
 
         comp2.sub(comp.data)
         time.sleep(0.5)
-#         comp2.unsub()
+        comp2.unsub()
```
pieleric, line 303: This should not be uncommented. It's to show where the unsub() call would be happening, but it's not happening and still things should work.... What you can do is to add a comment explaining that.

pieleric (changes requested):

tepals (approved):

PR #2289, opened 2022-12-13
driver avantes: fix simulator when start/stopping multiple times

pieleric:
The simulator didn't properly handle stopping. If a measurement was
stopped and just after a new one started, the first one would somehow
"continue"... in a very weird way, as all the rest of the code assumes
that only one measurement is running at a time. It could even happen if
starting a new measurement just after one finished, as happened in the
test case test_acquisition().

=> Stop using a RepeatingTimer, and introduce a dedicated class to
simulate measurement, using a command queue and FSM. It might be
slightly over-kill but works nicely.


src/odemis/driver/avantes.py
```diff
@@ -1076,6 +1075,128 @@ def _val(obj):
...
 
+# Start: tuple (number of measurements (int), period (float))
+SIM_CMD_STOP = "E"
+SIM_CMD_TERMINATE = "T"
+
+class RepeatingMeasurement:
+    """
+    Simulates a measurement loop.
+    Can be started/stopped at any time.
+    A callback is called one a measurement is completed.
```
nandishjpatel, line 1086:
```suggestion
    A callback is called once a measurement is completed.
```

src/odemis/driver/avantes.py
```diff
@@ -22,7 +22,9 @@
 # Driver for the Avantes spectrometers. The wavelengths are fixed, so no
 # spectrograph (actuator) component is provided, just the detector component.
 
-from ctypes import *
+import typing  # for typing.Union
+from typing import Tuple
+from ctypes import *  # also provides a Union
 import ctypes
 import logging
 import numpy
```
K4rishma, line 30:
```suggestion
import ctypes
from ctypes import *  # also provides a Union
import logging
import typing  # for typing.Union
from typing import Tuple

import numpy
```

tepals (approved):

PR #2290, opened 2022-12-13
 fix warning of wxPyDeprecationWarning in wxPython module

K4rishma:
Resolves wxPython warning ([SWM-73](https://delmic.atlassian.net/browse/SWM-73?atlOrigin=eyJpIjoiMmYwZGM1YjgwMGRmNGRlNjljNjE1MjljZTUyNDM3YTAiLCJwIjoiaiJ9)). The warnings were from the tests in odemis/src/odemeis/gui/test

Solution:-
Install PyPubSub package which is actually what wx.lib.pubsub was based on
Then replace 
`from wx.lib.pubsub import pub with from pubsub import pub`

Warning Message:-
`/usr/lib/python3/dist-packages/wx/lib/pubsub/__init__.py:23
  /usr/lib/python3/dist-packages/wx/lib/pubsub/__init__.py:23: wxPyDeprecationWarning: wx.lib.pubsub has been deprecated, plese migrate your code to use pypubsub, available on PyPI.
    warnings.warn('wx.lib.pubsub has been deprecated, plese migrate your '`

### DIFF
--- a/src/odemis/gui/cont/streams.py
+++ b/src/odemis/gui/cont/streams.py
@@ -34,7 +34,7 @@ from builtins import str
 import numpy
 import wx
 from past.builtins import basestring, long
-from wx.lib.pubsub import pub
+from pubsub import pub
 
 import odemis.acq.stream as acqstream
 import odemis.gui.model as guimodel
--- a/src/odemis/gui/cont/streams.py
+++ b/src/odemis/gui/cont/streams.py
@@ -258,6 +258,10 @@ class StreamController(object):
 
         self.entries = []
 
+        # Tell control bar to also discard all references to the stream
+        if self._sb_ctrl:
+             self._sb_ctrl.remove_stream(self.stream)
+
         gc.collect()
 
     def _display_metadata(self):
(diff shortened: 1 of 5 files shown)


K4rishma:
@Eric I solved it by installing PyPubSub package to resolve this warning. How should I add this installation in this PR?

src/odemis/gui/cont/streams.py
```diff
@@ -258,6 +258,10 @@ def _on_stream_panel_destroy(self, _):
 
         self.entries = []
 
+        # Tell control bar to also discard all references to the stream
+        if self._sb_ctrl:
+             self._sb_ctrl.remove_stream(self.stream)
+
```
pieleric, line 264: The idea of this code is that it's probably sufficient to not use pubsub at all.
So all the three places where there is `pubsub` imported should go away.
Of course, you'll need to make sure that this new code works... I am pretty sure there is no test case. So you need to manually test that when removing a stream (either in the localization tab or analysis tab), the "remove_stream()" function is indeed called.

pieleric (changes requested):

PR #2297, opened 2022-12-20
[fix] remove blocking function which keeps gui open after the test ca…

nandishjpatel:
```
Running /home/testing/development/odemis/src/odemis/gui/test/comp_legend_test.py:
FAIL comp_legend_test.py::LegendTestCase::test_legend
```

remove blocking function

### DIFF
--- a/src/odemis/gui/test/comp_legend_test.py
+++ b/src/odemis/gui/test/comp_legend_test.py
@@ -43,7 +43,6 @@ class LegendTestCase(test.GuiTestCase):
     frame_class = test.test_gui.xrccanvas_frame
 
     def test_legend(self):
-        test.goto_manual()
         self.frame.SetSize((400, 60))
         leg = legend.AxisLegend(self.panel)
         leg.SetBackgroundColour(wx.RED)


pieleric:
That's really odd. That function (`goto_manual()`) is explicitly supposed to do that... if the `NOMANUAL` environment variable is not set. Are you running the test with `export NOMANUAL=1` ? Does it also block/fail in such case?

nandishjpatel:
> That's really odd. That function (`goto_manual()`) is explicitly supposed to do that... if the `NOMANUAL` environment variable is not set. Are you running the test with `export NOMANUAL=1` ? Does it also block/fail in such case?

You are correct! with `export NOMANUAL=1` it works fine. Is this environment variable set in the test VM script @tepals?

pieleric:
It should be set at the beginning of `runtests.sh`....

tepals:
The line is there, I don't know why it doesn't work:
https://github.com/delmic/odemis/blob/master/util/runtests.sh#L57
```
# This environment variable makes the GUI test cases automatically close the test frames
export NOMANUAL=1
```

PR #2298, opened 2022-12-21
Technolution driver: increased default overvoltage on mppc

tepals:


nandishjpatel (approved):

pieleric:
@nandishjpatel , @tepals , note that to change the default value, you could also just update the microscope file and use a "properties" section for the component. The backend will then set that value just after initialization. The main difference is that you have to update the microscope file for every installation.

PR #2299, opened 2022-12-21
checks for pixel size metadata in simsem.py

K4rishma:
without raising error if not present

### DIFF
--- a/src/odemis/driver/simsem.py
+++ b/src/odemis/driver/simsem.py
@@ -143,7 +143,11 @@ class Scanner(model.Emitter):
 
         # pixelSize is the same as MD_PIXEL_SIZE, with scale == 1
         # == smallest size/ between two different ebeam positions
-        pxs = fake_img.metadata[model.MD_PIXEL_SIZE]
+        pxs = fake_img.metadata.get(model.MD_PIXEL_SIZE, (1e-6, 1e-6))
+        if not isinstance(pxs, tuple) or len(pxs) != 2:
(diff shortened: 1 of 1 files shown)


src/odemis/driver/simsem.py
```diff
@@ -143,7 +143,7 @@ def __init__(self, name, role, parent, aperture=100e-6, wd=10e-3, **kwargs):
 
         # pixelSize is the same as MD_PIXEL_SIZE, with scale == 1
         # == smallest size/ between two different ebeam positions
-        pxs = fake_img.metadata[model.MD_PIXEL_SIZE]
+        pxs = fake_img.metadata.get(model.MD_PIXEL_SIZE, (1e-6, 1e-6))
```
nandishjpatel, line 146: As per [here](https://github.com/K4rishma/odemis/blob/70ac8e6eb52ad5d3cd8e88c09164ba7487876be5/src/odemis/model/_metadata.py#L31) the default value can be tuple of 2 or 3 elements

pieleric: That's true indeed. If the image is 3D (XYZ), then the pixel size may have 3 values. You could add a check that the pixel size is only 2 dimensions, and fail otherwise. 
In this context, loading the simulated image for a camera simulator, passing a 3D image would be the user a little bit looking for troubles. I think it's fair to just say it's not supported.

pieleric (approved):

PR #2304, opened 2023-01-11
Extend Odemis Zeiss driver to handle rotation axes

sneepstefan:
Currently the Zeiss driver only supports X/Y/Z linear movements. That’s enough for the SPARC. For the METEOR, we’ll need at least Rx and Rz.

This is supported via the C95 command (6 axes move). See Remcon32.pdf p.34.

Testing should be first done a simulator. Then on a real hardware when we have the opportunity.


src/odemis/driver/zeiss.py
```diff
@@ -513,66 +560,101 @@ def _refreshPosition(self):
...
-
-        target_pos = self._applyInversion({"x": x * 1e-3, "y": y * 1e-3, "z": z * 1e-3})
-        # Check range (for the axes we are moving)
+            target_pos["z"] = z * 1e-3
+        if "rx" in shift:
+            t += shift["rx"]
+            target_pos["rx"] = t
+        if "rz" in shift:
+            r += shift["rz"]
+            target_pos["rz"] = r
```
pieleric, line 590: Convert to rad, and then back to degree for target_pos.
I think it might help/clarify if we had two functions to convert a pos (dict) from/to "user view" (ie, m/rad with inversion) and "hw view" (ie, mm/deg without inversion). Something like `pos_to_user()` and `pos_to_hw()`.

src/odemis/driver/zeiss.py
```diff
@@ -954,17 +1035,18 @@ def __init__(self, timeout=1, *args, **kwargs):
         self.pixelSize = 5
         self.blanker = 0
         self.external = 0
-        self.pos = numpy.array([20, 20, 5])  # X, Y, Z in mm
+        self.pos = numpy.array([20, 20, 5, 0, 0, 0])  # X, Y, Z in mm ,T, R, in radians, M in mm
```
pieleric, line 1038: TR are in degrees!

src/odemis/driver/zeiss.py
```diff
@@ -985,13 +1069,48 @@ def _run_move(self):
...
+                if self.pos[-2] < 0:
+                    # remainder should be subtracted from 360°
+                    bound_pos = 2 * math.pi - (0 - self.pos[-2])
+
+                elif self.pos[-2] > 2 * math.pi:
+                    # remainder should be added to 0°
+                    bound_pos = 0 + (self.pos[-2] - (2 * math.pi))
+
+                if bound_pos:
+                    self.pos[-2] = bound_pos
```
pieleric, line 1084: I think you can just write it like that:
```suggestion
                # R is allowed to do any rotation, but always stays within 0 -> 360°
                self.pos[-2] %= 360
```

src/odemis/driver/zeiss.py
```diff
@@ -985,13 +1069,48 @@ def _run_move(self):
...
+
             if not self._stage_stop.is_set():
                 self.pos = self.target_pos
+                if bound_pos:
+                    self.pos[-2] = bound_pos
+
```
pieleric, line 1090: You could also use the `%` trick.

pieleric (changes requested):

src/odemis/driver/test/zeiss_test.py
```diff
@@ -37,15 +37,17 @@
 # arguments used for the creation of basic components
 CONFIG_SCANNER = {"name": "scanner", "role": "ebeam", "hfw_nomag": 1}
 CONFIG_STAGE = {"name": "stage", "role": "stage",
-                "rng": {"x": (5.e-3, 152.e-3), "y": (5.e-3, 152.e-3)},  # skip one axis to see if default works
-                "inverted": ["x"],
-               }
+                "rng": {"x": (5.e-3, 152.e-3),
+                        "y": (5.e-3, 152.e-3),  # skip one axis to see if default works
```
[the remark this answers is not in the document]
sneepstefan: The fact is that it is actually present the moment the Stage object is created due to _updatePosition() being fired just before the initialization finished._updatePosition is followed up by _getStagePosition() and in there GetStagePosition6() is used always. So that makes it difficult to test if it is not present in stage.position. I am unsure why "z" is left out here in the testcase in the first place. I moved the comment (which was already there) because I changed the definition of the dict.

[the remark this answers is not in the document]
pieleric: Yes, indeed. I realised that later when reading further the code. Omitting the axis from "rng" doesn't means it's not present. We need to keep that behaviour (to not break the current systems). So we need a different way to indicate an axis is *not* present.

src/odemis/driver/zeiss.py
```diff
@@ -513,66 +560,101 @@ def _refreshPosition(self):
...
-        move by the shift
-        shift (float): unit m
+        prepare the stage for a relative move using the values in shift
+        :param future: a future that can be used to manage a move
+        :param shift (dict): contains float position values in m
         """
-        x, y, z, _ = self.parent.GetStagePosition()
+        # set the values of these axes to 0.0
+        # if none of these are used in the move the command for 3 axes will be used
+        t = r = m = 0.0
```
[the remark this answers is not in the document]
sneepstefan: Yes that's kind of silly right? Think its a remnant from the previous version of the method where there was also GetStagePositon on 3 axes present. It is not necessary here anymore indeed.

src/odemis/driver/test/zeiss_test.py
```diff
@@ -229,6 +231,108 @@ def test_move(self):
         dur = time.time() - tstart
         self.assertGreaterEqual(dur, 1.1, "1 mm move took only %g s" % dur)
 
+    def test_move_axes6_relative(self):
+        """
+        Test if it's possible to execute a relative move of the stage using 6 axes
+        Test if moving below 0° or over 360° using a full rotational axis applies correction of under- and overflow
```
K4rishma, line 237: Not sure if the overflow condition is checked. If it means something like this, i.e. 360° + 0.1.

src/odemis/driver/test/zeiss_test.py
```diff
@@ -150,9 +152,9 @@ def test_scan_rotation(self):
         self.assertAlmostEqual(orig_rot, ebeam.rotation.value)
 
     # @skip("skip")
-    def test_move(self):
+    def test_move_axes3(self):
         """
-        Check it's possible to move the stage
+        Check it's possible to move the stage on 3 axes
         """
```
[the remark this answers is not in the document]
sneepstefan: You are right, this sounds confusing I changed it to a more clear description.

src/odemis/driver/zeiss.py
```diff
@@ -513,66 +560,101 @@ def _refreshPosition(self):
...
         if "x" in pos:
             x = pos["x"] * 1e3
         if "y" in pos:
             y = pos["y"] * 1e3
         if "z" in pos:
             z = pos["z"] * 1e3
+        if "rx" in pos:
+            t = pos["rx"]
+        if "rz" in pos:
+            r = pos["rz"]
```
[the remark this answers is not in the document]
sneepstefan: I've changed the input in the testcases from radians to degree already. So this can stay this way, like it should be for the real case command which expects degree positions.

src/odemis/driver/test/zeiss_test.py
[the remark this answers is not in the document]
sneepstefan: I have changed this back to removing axes in the __init__ if they are not specified through the yaml file. So this way we 'preserve' the behavior the driver currently had in the past.

src/odemis/driver/zeiss.py
```diff
@@ -513,66 +560,101 @@ def _refreshPosition(self):
...
+            target_pos["m"] = m * 1e-3
+
+        target_pos = self._applyInversion(target_pos)
+        # Check range for axes which have finite ranges (for the axes we are moving)
         for an in shift.keys():
             rng = self.axes[an].range
             p = target_pos[an]
-            if not rng[0] <= p <= rng[1]:
-                raise ValueError("Relative move would cause axis %s out of bound (%g m)" % (an, p))
+            if not util.almost_equal(rng[1] - rng[0], 2 * math.pi):
```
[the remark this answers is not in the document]
sneepstefan: Changing 2pi to 360 neutralized this problem for linear axes. Using radians here was not necessary.

src/odemis/driver/zeiss.py
sneepstefan: There is no degree/rad conversion necessary anymore.

K4rishma (approved):

src/odemis/driver/zeiss.py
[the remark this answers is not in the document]
sneepstefan: I have simplified this to only check is an is not "rz" due to rz being the only axis which supports full rotations.

src/odemis/driver/zeiss.py
```diff
@@ -297,6 +301,20 @@ def GetStagePosition(self):
...
+        Read the absolute position of the stage for all 6 axes.
+        Non-existing axes get a float value of 0.0 (cannot be left out).
+        :return: float, float, float, float, float, float, bool: x, y, z in mm and t, r in deg and m in mm,
+        stage moving as bool.
+        """
+        # Return something like:
+        # 65.60162 64.75846 38.91104 0.23 259.22 0.0 0.0
+        # stage moving is either 0.0 or 1.0
+        s = self._SendCmd(b'C95?')
+        vals = s.split(b' ')
```
nandishjpatel, line 315:
```suggestion
        vals = s.strip().split(b' ')
```
Maybe better?

src/odemis/driver/zeiss.py
```diff
@@ -419,15 +437,26 @@ def Abort(self):
...
         self._SendCmd(c)
 
+    def MoveStage6(self, move_list):
+        """
+        Absolute move. Non-blocking.
+        Use GetStagePosition6() to check the move status.
+        :param move_list (floats): list with 6 absolute target positions in mm or degree
+        """
+        x, y, z, t, r, m = move_list
+        c = b'C95 %f %f %f %f %f %f' % (x, y, z, t, r, m)
```
nandishjpatel, line 457: This is tricky, here one may loose some precision. I don't how important the precision is.

For example, trying following in python:
```
>>> c = b'C95 %f' % (1.123456789)
>>> c
b'C95 1.123457'
```

One workaround is to count the precision for each of `x, y, z, t, r, m` and then use something like follows

```
>>> c = b'C95 %.9f' % (1.123456789)
>>> c
b'C95 1.123456789'
```

src/odemis/driver/zeiss.py
```diff
@@ -985,13 +1071,38 @@ def _run_move(self):
...
+        pos[3:5] = [math.radians(pos[3]), math.radians(pos[4])]
+        self.target_pos = numpy.array(pos)
+        self._start_move = time.time()
+        dist = numpy.sqrt((self.pos - self.target_pos) ** 2)
+        dur = sum(dist) / self._speed
+        self._end_move = time.time() + dur
+        self.dur = dur
+        self._is_moving = True
+        self._stage_stop.clear()
+        self._mover = threading.Thread(target=self._run_move)
```
nandishjpatel, line 1103: Just a question, how often will `_move_to_pos()` be called during runtime?

nandishjpatel (approved):
Some suggestions and questions, otherwise looks good to me

src/odemis/driver/zeiss.py
pieleric: That's a good remark. I've looked at some logs. STG? (ie, the position as given by the SEM) returns values like that: `45.26179 58.16799 33.24926`. So 5 digits after the decimal point. As it's in mm, that's 10nm, which is probably the best of the stage encoder precision (and a lot better than a move). So the standard `%f`, with 6 decimals should be fine.
For rotations, we don't have logs, but as it's in degrees, and I don't expect a precision more than 0.01°, that should be always fine. 

=> %f is fine, but please add a comment to explain it.

src/odemis/driver/zeiss.py
pieleric: In some worst cases, about once per second. Typically, it's more like every 10s or less. 
I think you're worried about the cost of creating a new thread every time? I'd say it's fine for this case. It's also only for the simulator. On real systems, this code is not used.

src/odemis/driver/test/zeiss_test.py
```diff
@@ -229,6 +231,122 @@ def test_move(self):
...
 
+    def test_move_axes6_relative(self):
+        """
+        Test if it's possible to execute a relative move of the stage using 6 axes
+        Test if moving below 0° using a full rotational axis applies correction of underflow
+        Test if moving with an inverted axis works like expected
+        """
+        self.stage.axes["rx"] = model.Axis(unit="rx", range=(0, math.radians(90)))
+        self.stage.axes["rz"] = model.Axis(unit="rz", range=(0, math.radians(360)))
+        self.stage.axes["m"] = model.Axis(unit="m", range=(0, 10.e-3))
```
pieleric, line 242: Humm, that's not supposed to be used like that. `.axes` is supposed to be read-only (and it is, if you're using it via the backend). The axes can only be added at init. Please just extend CONFIG_STAGE, as you used to do.

src/odemis/driver/zeiss.py
```diff
@@ -438,28 +467,34 @@ class Stage(model.Actuator):
...
+            rng["rx"] = (0, 0)
+        if "rz" not in rng:
+            rng["rz"] = (0, 0)
+        if "m" not in rng:
+            rng["m"] = (0, 0)
+
+        axes_def = {}
+        for ax, r in rng.items():
+            rng_ax = (r[0], r[1])  # Force a tuple, and check the user passed 2 values
+            if rng_ax != (0, 0):  # if range is not specified exclude the axis
```
pieleric, line 495: That behaviour sounds good to me, *except* that `(0, 0)` is not really nice to indicate the axis isn't supported. Please, replace by the special value `None` (which will be written `null` in YAML).
Also, indicate this behaviour in the docstring about `rng`.

src/odemis/driver/zeiss.py
```diff
@@ -483,20 +518,37 @@ def terminate(self):
...
+        Read the stage position, using either GetStagePosition of GetStagePosition6 where the first supports 3
+        linear axes and the latter 6 axes including rx and rz rotational axes.
+        :return position dict str -> float: (in m/rad) with the axes supported in .axes 
+                is_moving (bool)
         """
-        if raw_pos is None:
-            x, y, z, _ = self.parent.GetStagePosition()
+        pos = {}
+
+        if len(self.axes) > 3:
```
pieleric, line 530: Now it's actually possible to disable any of the axes, including "x", "y", and "z". So you cannot rely just on the number of axes. You should check if rx, rz, or m are in the axes, and in that case use the 6 version.

src/odemis/driver/zeiss.py
```diff
@@ -483,20 +518,37 @@ def terminate(self):
...
+            pos["rx"] = math.radians(t)
+            pos["rz"] = math.radians(r)
+            pos["m"] = m * 1e-3
         else:
-            x, y, z = raw_pos["x"], raw_pos["y"], raw_pos["z"]
+            x, y, z, is_moving = self.parent.GetStagePosition()
+
+        pos["x"] = x * 1e-3
+        pos["y"] = y * 1e-3
+        pos["z"] = z * 1e-3
```
pieleric, line 540: You should only copy the axes to pos **if** they are in self.axes!
For instance, if "z" is not in self.axes, pos shouldn't contain it.

Note that one option would be to do the code as you have it here, and just after add a line to remove the axes which are not present.

src/odemis/driver/zeiss.py
```diff
@@ -513,66 +565,99 @@ def _refreshPosition(self):
...
     def _doMoveRel(self, future, shift):
         """
-        move by the shift
-        shift (float): unit m
+        prepare the stage for a relative move using the values in shift
+        :param future: a future that can be used to manage a move
+        :param shift (dict): contains float position values in m or deg
         """
-        x, y, z, _ = self.parent.GetStagePosition()
+        if len(set(shift.keys()).intersection({"rx", "rz", "m"})) > 0:
```
pieleric, line 572: This works fine, as any non-empty set is considered `True`:
```suggestion
        if set(shift.keys()).intersection({"rx", "rz", "m"}):
```

src/odemis/driver/zeiss.py
```diff
@@ -513,66 +565,99 @@ def _refreshPosition(self):
...
         with future._moving_lock:
             try:
                 if future._must_stop.is_set():
                     raise CancelledError()
-                logging.debug("Moving to position (%s, %s, %s)", x, y, z)
-                self.parent.MoveStage(x, y, z)
-                # documentation suggests to wait 1s before calling
-                # GetStagePosition() after MoveStage()
+                # use explicitly the 3-axes or the 6-axes move command
+                if len(move_list) > 3:
```
pieleric, line 645: Check the rx, rz, m axes are there.

src/odemis/driver/zeiss.py
```diff
@@ -985,13 +1071,38 @@ def _run_move(self):
                                      1)
                 traveled_dist = total_dist * traveled_ratio
                 self.pos = orig_pos + traveled_dist
+
+                # R is allowed to do any rotation, but always stays within 0 -> 360°
+                self.pos[-2] %= 2 * math.pi
+
             if not self._stage_stop.is_set():
                 self.pos = self.target_pos
+                self.pos[-2] %= 2 * math.pi
```
pieleric, line 1080: pos is in degrees. So it should be `%= 360`.

pieleric (changes requested):

src/odemis/driver/zeiss.py
sneepstefan: Thanks for clearing that up Eric.

src/odemis/driver/zeiss.py
sneepstefan: I think it is fine for now, since we already used the command GetStagePosition for a while and the output from this is reliable. The GetStagePosition6 is derived from this command.

src/odemis/driver/test/zeiss_test.py
sneepstefan: Yeah I did actually try to add them to CONFIG_STAGE but then I get problems running the testcases for the 3 axes and 6 axis in the same TestSEM object. In this case the test_move_axes3 is using 6-axes commands for setting and getting due to all the 6 axes registered in self.axes.

src/odemis/driver/zeiss.py
sneepstefan: Actually the elements for the rotational axes in pos are converted to radians just before firing _run_move in _move_to_pos. This is to keep the readonly self._speed constant usable for the rotational axes. Therefore I have to check for 2π instead of 360°.

src/odemis/driver/test/zeiss_test.py
sneepstefan: Do you have any suggestions for a 'smart' way to combine these testcases in one testobject? Or is it just a matter of splitting up 3-axes and 6-axes versions of testcases?

pieleric: Yes, just split up the 2 types. Just create a new TestCase class.

src/odemis/driver/zeiss.py
pieleric: Ah, thanks for explaining. I now see the conversions. So the advantage is that the realistic speed corresponding to mm/s also corresponds to a realistic speed in rad/s. 
With this in mind, I'd recommend to adjust the simulator to use degrees only. I think that the amount of code will stay about the same, but it'll be clearer that the rotations are always in degrees (in the simulator) and that the speed for rotation and linear movements is not the same. 
I'd suggest to have two speeds: `_speed_lin = 0.1 # mm/s ` and `_speed_rot = 10 # deg/s`. Then, you never need to do any conversion for the rotation axes... but you need to compute the move time slightly differently. It should be relatively straightforward. I'd suggest something like that:
```python
dur_lin = sum(numpy.sqrt((self.pos[[0,1,2,5]] - self.target_pos[[0,1,2,5]]) ** 2)) / self._speed_lin
dur_rot = sum(abs(self.pos[[3,4]] - self.target_pos[[3,4]])) / self._speed_rot
```

src/odemis/driver/zeiss.py
```diff
@@ -513,66 +572,99 @@ def _refreshPosition(self):
...
             p = target_pos[an]
-            if not rng[0] <= p <= rng[1]:
-                raise ValueError("Relative move would cause axis %s out of bound (%g m)" % (an, p))
+            if an != "rz":  # only rz has full rotational support and shouldn't be checked
+                if not rng[0] <= p <= rng[1]:
+                    raise ValueError("Relative move would cause axis %s out of bound (%g m)" % (an, p))
 
-        self._moveTo(future, x, y, z)
+        # pass move_list without stage_moving value
+        self._moveTo(future, move_list[:-1])
```
pieleric, line 616: Please discard the `stage_moving` value at the very beginning (eg, line 583), so that `move_list` really only contain axes move.

pieleric (changes requested):

src/odemis/driver/test/zeiss_test.py
```diff
@@ -36,41 +36,58 @@
...
-CONFIG_SEM = {"name": "sem", "role": "sem", "port": "/dev/ttyUSB*",  # "/dev/fake*"
+CONFIG_SEM3 = {"name": "sem", "role": "sem", "port": "/dev/ttyUSB*",  # "/dev/fake*"
               "children": {"scanner": CONFIG_SCANNER,
                            "focus": CONFIG_FOCUS,
-                           "stage": CONFIG_STAGE,
-                          }
-              }
+                           "stage": CONFIG_STAGE3, }
+               }
+CONFIG_SEM6 = {"name": "sem", "role": "sem", "port": "/dev/ttyUSB*",  # "/dev/fake*"
```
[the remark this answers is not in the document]
sneepstefan: Since the "stage" dict is nested into the CONFIG just copying it and change the stage will still override it in the CONFIG_SEM version.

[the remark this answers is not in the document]
pieleric: Good remark... you can deal with it using `copy.deepcopy()`.

src/odemis/driver/zeiss.py
```diff
@@ -513,66 +572,99 @@ def _refreshPosition(self):
...
+        # Check range for axes which have finite ranges (for the axes we are moving)
+        # for full rotational axes moving freely is allowed
         for an in shift.keys():
             rng = self.axes[an].range
             p = target_pos[an]
-            if not rng[0] <= p <= rng[1]:
-                raise ValueError("Relative move would cause axis %s out of bound (%g m)" % (an, p))
+            if an != "rz":  # only rz has full rotational support and shouldn't be checked
+                if not rng[0] <= p <= rng[1]:
+                    raise ValueError("Relative move would cause axis %s out of bound (%g m)" % (an, p))
```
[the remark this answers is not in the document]
sneepstefan: Yeah but we should drop "rx" since that is not an axis which supports full rotation, only 0-90°.
So then line 612 would become:
if an in ("rz") and util.almost_equal(rng[1] - rng[0], 2 * math.pi)
But this feels like a double check right?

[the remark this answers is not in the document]
sneepstefan: Would in this case not only:

if an != "rz":

suffice?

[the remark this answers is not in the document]
pieleric: No, keep "rx". Let's be open-minded. The first SEM we are targetting doesn't support full rotation, but that doesn't mean the next ones will not.

src/odemis/driver/zeiss.py
```diff
@@ -438,28 +469,37 @@ class Stage(model.Actuator):
...
+        if "rx" not in rng:
+            rng["rx"] = None
+        if "rz" not in rng:
+            rng["rz"] = None
+        if "m" not in rng:  # Extra Z axis, in the Rx referential, to adjust eucentric height.
+            rng["m"] = None
+
+        axes_def = {}
+        for ax, r in rng.items():
+            if r:  # if range is not specified exclude the axis
```
pieleric, line 499: Please mark the comparison to `None` explicitly. It's not required by Python, but that makes things clearer when reading the code, and avoid also supporting `0`, `""` and `[]`.
```suggestion
            if r is not None:  # if range is not specified exclude the axis
```

src/odemis/driver/zeiss.py
```diff
@@ -513,66 +574,112 @@ def _refreshPosition(self):
...
                     raise CancelledError()
-                logging.debug("Moving to position (%s, %s, %s)", x, y, z)
-                self.parent.MoveStage(x, y, z)
-                # documentation suggests to wait 1s before calling
-                # GetStagePosition() after MoveStage()
+                # use explicitly the 3-axes or the 6-axes move command
+                if set(self.axes.keys()).intersection({"rx", "rz", "m"}):
+                    self.parent.MoveStage6(move_list)
+                else:
+                    self.parent.MoveStage(move_list)
```
pieleric, line 670: Let's use `*move_list` here.

pieleric (changes requested):
The radians and degrees separation looks fine now!
I just a few tiny comments, and it should be ready to merge.

src/odemis/driver/zeiss.py
```diff
@@ -419,15 +437,28 @@ def Abort(self):
...
         return self._SendCmd(b'ABO')
 
-    def MoveStage(self, x, y, z):
+    def MoveStage(self, move_list):
         """
-        Absolute move. Non blocking.
+        Absolute move. Non-blocking.
         Use GetStagePosition() to check the move status.
-        x, y, z (floats): absolute target position in mm
+        :param move_list (floats): list with 3 absolute target positions in mm
```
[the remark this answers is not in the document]
sneepstefan: Like this example for MoveStage6:
:param move_list (floats): x, y, z, t, r, m absolute target positions in mm or degree

pieleric (approved):

PR #2307, opened 2023-01-13
SDEV-1247 fix calibration region overlay appearing when the button was deselected

tepals:
When selecting and deselecting a region of calibration the coordinates of active but deselected scintillators can update unexpectedly. When the coordinates change all active scintllators are checked and re-drawn based on the coordinates. Because the coordinates had updated even the deselected boxes would be re-drawn. This fix makes sure that boxes are only drawn based on whether the calibration region box has been selected or de-selected.

### DIFF
--- a/src/odemis/gui/cont/project.py
+++ b/src/odemis/gui/cont/project.py
@@ -387,6 +387,7 @@ class FastEMROCController(object):
         self.calib_model.coordinates.subscribe(self._on_coordinates)
 
         self.overlay = None
+        self.is_selected = False
 
     def fit_view_to_bbox(self):
         """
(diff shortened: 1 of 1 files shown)


nandishjpatel (approved):

src/odemis/gui/cont/project.py
```diff
@@ -508,27 +512,21 @@ def _on_coordinates(self, _=None):
...
-            else:
+            if num in active_scintillators:
+                b.Enable(True)  # always enable the button when the scintillator is active
+                if self.roc_ctrls[num].is_selected:
+                    b.SetLabel("OK")
+                    if self._calib_prefix == "calib_2":
+                        b.SetForegroundColour(FG_COLOUR_WARNING)
+                    else:
+                        b.SetForegroundColour(wx.GREEN)  # default to green
+                elif not self.roc_ctrls[num].is_selected:
```
pieleric, line 526: is_selected is a boolean, so it should be that simple:
```suggestion
                else:
```

pieleric (approved):

PR #2311, opened 2023-01-20
driver: document some hardware behaviour better

pieleric:


FCanberk (approved):

src/odemis/driver/lakeshore.py
```diff
@@ -154,6 +154,8 @@ def __init__(self, name, role, port, sensor_input='b', output_channel=2, **kwarg
             if output_channel_nr not in supported_output_channels:
                 raise ValueError(f"Sensor input must be one of the following: {supported_output_channels}")
 
+            # TODO: confirm that there is an input with the same name and that
+            # the device is configured accordingly (OUTMODE? reports which input is linked to the output)
```
tepals, line 158: I don't understand what needs to be done in this TODO, can you make it more clear?

pieleric: I've now implemented half of the TODO, and clarified the other half. Let me know if you think it's not yet clear enough.

tepals (approved):

PR #2316, opened 2023-01-24
driver smaract SmarPod: add option to rename the axes

pieleric:
There are other ways to rename the axes, especially by using the
MutliplexActuator. However, this doesn't support the MD_PIVOT_POS.
=> Directly support renaming the axes to whichever name.

This also keeps the microscope files simpler.

### DIFF
--- a/src/odemis/driver/smaract.py
+++ b/src/odemis/driver/smaract.py
@@ -320,8 +320,9 @@ class SmarPod(model.Actuator):
                 network:sn:<serialnumber>
         hwmodel: the hardware model code (typically between 10000 and 10100)
         axes (dict str (axis name) -> dict (axis parameters)):
-            Typically, axes are x, y, z, rx, ry, rz
+            Axes can be x, y, z, rx, ry, rz
             axis parameters: {
+                name: (str), name of the axis as seen by the user in Odemis (default to the axis name)
                 range: [float, float], default is -1 -> 1
                 unit: (str) default will be set to 'm'
             }
@@ -331,6 +332,7 @@ class SmarPod(model.Actuator):
         accel: the maximum acceleration (in m/s²) of a point on the stage
         hold_time (float>=0): the hold time, in seconds, for the actuator after the target position is reached.
             Default is infinite (float('inf') in Python, .inf in YAML). Can be also set to 0 to disable hold.
+        inverted: (list of str) name of the axes (user names) that should be inverted
         """
         if not axes:
             raise ValueError("Needs at least 1 axis.")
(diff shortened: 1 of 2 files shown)


src/odemis/driver/test/smaract_test.py
```diff
@@ -68,14 +70,15 @@
                 'unit': 'rad',
             },
             'ry': {
-                'range': [-0.35, 0.35],
+                'range': [-0.45, 0.35],
```
K4rishma, line 73: Is it different i.e. assymetrical rotation in ry?

K4rishma: same remark for x

src/odemis/driver/smaract.py
```diff
@@ -1443,7 +1499,6 @@ def __init__(self, name, role, locator, axes, ref_on_init=False, linear_speed=0.
         else:
             self.core = FakeMC_5DOF_DLL(axes)
         # Not to be mistaken with axes which is a simple public view
```
K4rishma, line 1501: Maybe this comment was for the next deleted line?

src/odemis/driver/test/smaract_test.py
pieleric: I made it different so that it's easy to check which axis got which range in the test cases. There is no physical change in the system... and that doesn't matter much because it's mostly for the simulator. However, I realise that this test case could run on the real hardware, so I'm going to adjust the axes to be smaller than the actual hardware range.

K4rishma (approved):

src/odemis/driver/smaract.py
```diff
@@ -344,26 +346,38 @@ def __init__(self, name: str, role: str, locator: str, hwmodel: int=10001,
...
+                logging.info("Axis %s has no range. Assuming (-1, 1)", axis_hw)
                 axis_range = (-1, 1)
 
             try:
                 axis_unit = param['unit']
             except KeyError:
-                logging.info("Axis %s has no unit. Assuming m", axis_name)
-                axis_unit = "m"
+                axis_unit = "rad" if axis_hw.startswith("r") else "m"
+                logging.info("Axis %s has no unit. Assuming %s", axis_unit, axis_hw)
```
sneepstefan, line 373: Shouldn't axis_ unit and axis_hw not be switched in this case?
```suggestion
                logging.info("Axis %s has no unit. Assuming %s", axis_hw, axis_unit)
```

src/odemis/driver/smaract.py
```diff
@@ -421,6 +438,34 @@ def __init__(self, name: str, role: str, locator: str, hwmodel: int=10001,
...
+
+    def _to_hw_axes(self, pos: dict) -> dict:
+        """
+        Convert the axis names of a position from user to hardware names.
+        It also invert the axes which are marked as "inverted".
+        :param pos: dict (str -> value), the position, as seen by the user.
+        Not all existing axes have to be provided.
+        :return: the position with the names of the axes as the hardware/SmarPod DLL
+        expects them
+        KeyError: if an axis doesn't exist in the hardware
```
sneepstefan, line 463:
```suggestion
        KeyError: if an axis doesn't exist in the user defined axes definition.
```

src/odemis/driver/smaract.py
```diff
@@ -436,14 +481,17 @@ def terminate(self):
     def updateMetadata(self, md):
         if model.MD_PIVOT_POS in md:
             pivot = md[model.MD_PIVOT_POS]
-            if not (isinstance(pivot, dict) and set(pivot.keys()) == {"x", "y", "z"}):
-                raise ValueError("Invalid metadata, should be a coordinate dictionary with x, y, z keys but got %s." % (pivot,))
+            if not isinstance(pivot, dict):
+                raise ValueError("Invalid metadata, should be a coordinate dictionary but got %s." % (pivot,))
+            pivot_hw = self._to_hw_axes(pivot)
+            if not set(pivot_hw.keys()) == {"x", "y", "z"}:
+                raise ValueError("Invalid metadata, should have x, y, z keys (in hardware names) but got %s." % (pivot_hw,))
```
sneepstefan, line 488: You could actually simplify the format string variable to % pivot and % pivot_hw on line 485 and 488.

src/odemis/driver/smaract.py
```diff
@@ -782,29 +831,36 @@ def moveAbs(self, pos):
...
         raise:
             SmarPodError: if the controller reported an error
             CancelledError: if cancelled before the end of the move
         """
         last_upd = time.time()
+        pos_hw = self._to_hw_axes(pos)
         dur = self._estimateMoveDuration(pos)
         end = time.time() + dur
-        max_dur = dur * 2 + 1
+        max_dur = dur * 5 + 3
```
sneepstefan, line 863: Could you very briefly explain where the values in dur * 5 + 3 are based on?

src/odemis/driver/smaract.py
pieleric: Good catch!

src/odemis/driver/smaract.py
pieleric: No and yes. For `pivot`, we cannot, because we don't know the type. We just know it's *not* a dict. If it's a tuple `1, 2, 3`, and you have the format "blabla %s" % pivot, it will try to map `1` to the `%s` and then fail because there is nothing for `2` and `3`! That's actually why I have the habit of *always* passing a tuple, even when there is only one argument.
For `pivot_hw`, as we know it's a dict, it's technically not necessary, indeed. But I prefer to *always* pass a tuple.

The conclusion is: use f-strings ;-)

src/odemis/driver/smaract.py
pieleric: Done: "The move estimation is really rough, so add some big margin to be on the safe side (we should *never* timeout before a successful move is completed)"

sneepstefan (approved):

PR #2317, opened 2023-01-27
model: do not automatically fallback to a different BASE_DIRECTORY

pieleric:
Commit 6064bb32afe (model: if base directory is not available use a
temporary one) changed the way BASE_DIRECTORY is set.

This introduced two issues:
* As soon as odemis.model is imported, code is run, which is not a good behaviour.
In particular, it would call logging, possibly very early at startup,
which causes some odd logging configuration (everything is displayed in
the terminal).
* The standard BASE_DIRECTORY (/var/run/odemisd) *never* exists when
starting the backend. So it would always create a temporary directory.
Normally, create that directory is handled by the backend.

=> Remove the smartness of the BASE_DIRECTORY constant.
The backend takes care of creating that directory.
=> Create a new function testing.use_fake_backend_directory() for the
cases where we need a fake base directory... However I couldn't
reproduce the issue in ProxyOfProxy test cases. So for now it's unused.

### DIFF
--- a/src/odemis/model/test/remote_test.py
+++ b/src/odemis/model/test/remote_test.py
@@ -20,22 +20,22 @@ PURPOSE. See the GNU General Public License for more details.
 You should have received a copy of the GNU General Public License along with 
 Odemis. If not, see http://www.gnu.org/licenses/.
 '''
-import Pyro4
-from concurrent import futures
-from concurrent.futures import CancelledError
 import gc
 import logging
-import numpy
-from odemis import model
-from odemis.model import roattribute, oneway, isasync, VigilantAttributeBase
-from odemis.util import mock, timeout, executeAsyncTask
 import os
 import pickle
-import sys
 import threading
 import time
 import unittest
+from concurrent import futures
 from multiprocessing import Process
+
+import numpy
+import Pyro4
+
+from odemis import model
+from odemis.model import VigilantAttributeBase, isasync, oneway, roattribute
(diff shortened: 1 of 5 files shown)


FCanberk (approved):

src/odemis/model/test/remote_test.py
```diff
@@ -157,12 +158,17 @@ def test_mock(self):
 
 # @unittest.skip("simple")
 class ProxyOfProxyTest(unittest.TestCase):
-# Test sharing a shared component from the client
+    # Test sharing a shared component from the client
+
+    # @classmethod
+    # def setUpClass(cls) -> None:
+    #     testing.use_fake_backend_directory()
```
tepals, line 165: why don't you add a test case that uses the fake backend directory or add a comment why it is unused here.

src/odemis/model/test/remote_test.py
```diff
@@ -20,23 +20,24 @@
...
 import sys
 import threading
 import time
 import unittest
+from concurrent import futures
+from concurrent.futures import CancelledError
 from multiprocessing import Process
 
+import numpy
+import Pyro4
```
tepals, line 36: I think this should be a separate group

src/odemis/model/test/remote_test.py
pieleric: Created test case for it... it turned out to be a really good idea, as there was a bug in the code!

tepals (approved):

PR #2318, opened 2023-01-28
add milling module using Orsay API

FCanberk:


src/odemis/acq/test/orsay_milling_test.py
```diff
@@ -0,0 +1,282 @@
...
+        f.result(timeout=40)
+        milling_state = self._get_milling_state()
+        self.assertEqual(milling_state, '0')
+
+        f.result()
+        self.assertTrue(f.done())
+
+    def test_milling_inverted_rect(self):
+        """
+        Test milling a rectangle with inverted values. X1>X2 and Y1>Y2 in this case.
```
K4rishma, line 151: I think this test is overarching test for inverted vertical line and inverted horizontal line. Is that the reason we are are testing inverted lines?

src/odemis/acq/orsay_milling.py
```diff
@@ -0,0 +1,240 @@
...
+            if time.time() > milling_activation_timeout:
+                raise TimeoutError("Milling failed to start up after trying for 10 seconds.")
+
+        logging.debug("Now milling a rectangle %sm by %sm for %s seconds",
+                      obj_to_mill.primaryAxisX*2, obj_to_mill.secondaryAxisLength, duration)
+
+        milling_execution_timeout = time.time() + duration * iteration + 5
+        while miller.MillingActivationState.Actual != '0':
+            time.sleep(0.1)
+            if time.time() > milling_execution_timeout:
```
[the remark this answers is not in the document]
FCanberk: In that case it would log the "Milling started" message every 0.1-ish second until the milling process is finished.

src/odemis/acq/test/orsay_milling_test.py
FCanberk: Ideally the X2 and Y2 values are greater than X1 and Y1 values respectively. This one is already an edge-case that does overarch the lines cases. So yes, that's the reason we are not testing inverted lines.

K4rishma (approved):

src/odemis/acq/test/orsay_milling_test.py
```diff
@@ -0,0 +1,282 @@
...
+        self.assertTrue(f.running())
+        milling_state = self._get_milling_state()
+        self.assertEqual(milling_state, '1')
+
+        # the milling process should be finished within about 20 seconds
+        f.result(timeout=30)
+        milling_state = self._get_milling_state()
+        self.assertEqual(milling_state, '0')
+
+        f.result()
```
K4rishma, line 172: There are two f.result(), maybe one will work

src/odemis/acq/test/orsay_milling_test.py
```diff
@@ -0,0 +1,282 @@
...
+                             "focus": CONFIG_FOCUS,
+                             "detector": CONFIG_DETECTOR},
+                }
+
+NO_SERVER_MSG = "TEST_NOHW is set. No server to contact."
+CANT_FORCE_ACTUAL_MSG = "TEST_NOHW is not set to sim, cannot force data on Actual parameters of Orsay server outside " \
+                        "of simulation. "
+
+CONFIG_PATH = os.path.dirname(odemis.__file__) + "/../../install/linux/usr/share/odemis/"
+MIMAS_CONFIG = CONFIG_PATH + "sim/mimas-sim.odm.yaml"
```
pieleric, line 59: The mimas-sim.odm.yaml will not work, as it doesn't use the Orsay server. 
We have three options: 
* Rely on a back-end to be already running. If so, make it clear at the top of the file.
* Start the backend `install/linux/usr/share/odemis/hwtest/orsay.odm.yaml` when TEST_NOHW is False
* Start the backend `install/linux/usr/share/odemis/hwtest/orsay.odm.yaml` when TEST_NOHW=sim (because this file actually contains by default the IP address to the VM), and rely on the back-end to be already running if TEST_NOHW is False.

For now the Orsay simulator doesn't support the milling. So I'd go for option 1, which is the simplest.

src/odemis/acq/test/orsay_milling_test.py
```diff
@@ -0,0 +1,282 @@
...
+
+# TEST_NOHW = os.environ.get("TEST_NOHW", "0")  # Default to Hw testing
+TEST_NOHW = "0"
+
+if TEST_NOHW == "0":
+    TEST_NOHW = False
+elif TEST_NOHW == "1":
+    CONFIG_ORSAY["host"] = "192.168.56.101"  # IP address of the simulated Orsay Physics Control Server
```
pieleric, line 67: No need for that.

src/odemis/acq/test/orsay_milling_test.py
```diff
@@ -0,0 +1,282 @@
...
+        # start the backend
+        cls.scanner = model.getComponent(role="ion-beam")
+        cls.scanner.horizontalFoV.value = 100e-6  # m
+        cls.host = cls.scanner.parent.host
+
+    def _get_milling_state(self):
+        """
+        Gets current status of the milling process
+        :params miller: HybridPatternCreator in Orsay API
+        :return: current milling state
```
pieleric, line 92: As it's always "0" or "1", could you convert the return value to an int? It'll be nicer to read the callers.

pieleric (changes requested):

src/odemis/acq/orsay_milling.py
```diff
@@ -0,0 +1,240 @@
...
+
+        milling_activation_timeout = time.time() + 10
+        # wait for milling action to start
+        while miller.MillingActivationState.Actual != '1':
+            time.sleep(0.1)
+            if time.time() > milling_activation_timeout:
+                raise TimeoutError("Milling failed to start up after trying for 10 seconds.")
+
+        logging.debug("Now milling a rectangle %sm by %sm for %s seconds",
+                      obj_to_mill.primaryAxisX*2, obj_to_mill.secondaryAxisLength, duration)
```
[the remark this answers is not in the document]
FCanberk: The secondary axis is actually secondary axis **length** that's why it is not multiplied by 2. I'll add a comment about the secondary axis length and update the duration part indeed.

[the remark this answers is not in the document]
pieleric: No, it's half of the length. At least, that's how it's described in the documentation: "the length of the object’s center to the object’s edge". It's also how the formulas to compute the value, just above in the code seem to expect it.

src/odemis/acq/orsay_milling.py
```diff
@@ -0,0 +1,244 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+Created on 13 Oct 2021
```
pieleric, line 4: 2021 -> 2022?

pieleric (changes requested):

src/odemis/acq/orsay_milling.py
```diff
@@ -0,0 +1,243 @@
...
+        obj_to_mill.centerOfMassX = ((rect[2] + rect[0]) / 2 - 0.5) * fov
+        obj_to_mill.centerOfMassY = (0.5 - (rect[3] + rect[1]) / 2) * fov
+
+        lay = Layer()
+        lay.probeSize = probe_size
+        lay.AddForm(0, obj_to_mill)
+
+        proc = Procedure()
+        proc.AddLayer(0, lay)
+        proc.name = f"Odemis rectangle milling {time.time()}"
```
pieleric, line 179: Please add a comment that the procedure name may not include `:` (or whatever character you found out is not allowed)

src/odemis/acq/orsay_milling.py
```diff
@@ -0,0 +1,243 @@
...
+        while miller.MillingActivationState.Actual != '0':
+            time.sleep(0.1)
+            if time.time() > milling_execution_timeout:
+                raise TimeoutError("Milling operation timed out. Please check whether the milling is done correctly.")
+        if self.future.cancelled():
+            raise futures.CancelledError()
+
+        logging.debug("Milling completed.")
+
+        miller.MillingActivationState.Unsubscribe(self.on_milling_state)
```
pieleric, line 209: That line must *always* be called after the end of the function, even if an exception happens.
=> place that line in a `finally` block, with a `try:` starting after line 135.

pieleric (changes requested):

pieleric (approved):

PR #2321, opened 2023-02-02
[update] now making use of calibration class instead of the module in…

nandishjpatel:
… the Calibration enum class, made further changes accordingly

### DIFF
--- a/src/odemis/acq/align/fastem.py
+++ b/src/odemis/acq/align/fastem.py
@@ -37,7 +37,6 @@ try:
     from fastem_calibrations.image_rotation_pre_align import ImageRotationPreAlign
     from fastem_calibrations.image_translation import ImageTranslation
     from fastem_calibrations.image_translation_pre_align import ImageTranslationPreAlign
-    from fastem_calibrations.pattern_stigmation import PatternStigmation
     from fastem_calibrations.scan_amplitude import ScanAmplitude
     from fastem_calibrations.scan_amplitude_pre_align import ScanAmplitudePreAlign
     from fastem_calibrations.scan_rotation import ScanRotation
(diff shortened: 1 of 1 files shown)


src/odemis/acq/align/fastem.py
```diff
@@ -233,9 +240,11 @@ def run(self):
 
             # loop over calibrations in list (order in list is important!)
             for calib in self.calibrations:
-                logging.debug("Starting calibration %s", calib)
                 # TODO return a sub-future when implemented for calibrations
```
tepals, line 243: I think this todo should now go to line 247

tepals (approved):

tepals (approved):

src/odemis/acq/align/fastem.py
```diff
@@ -279,76 +288,36 @@ def run_calibration(self, calibration):
...
-            self.asm_config["mppc"]["cellTranslation"] = translation
-            self.asm_config["multibeam"]["resolution"] = resolution
-            self.asm_config["mppc"]["cellCompleteResolution"] = cell_complete_resolution
-
-        else:
-            raise ValueError(f"Unknown calibration {calibration.name}.")
+        calibration.run()
+
+        if isinstance(calibration, Calibrations.SCAN_AMPLITUDE_PREALIGN.value):
+            self.asm_config["multibeam"]["scanOffset"] = calibration.orig_config["multibeam"]["scanOffset"]
```
pieleric, line 294: This is fine... but I think we could consider doing something even more fancy.
Now that the run() is always the same, all that is different is the values to store. Could the calibration class provide this information too, so that we can have one generic code here?
Maybe something like `calibration.updated_settings` containing a list of tuples [str,str] (component role, VA name).
Then you'd do:
```
for comp, va in calibration.updated_settings:
      self.asm_config[comp][va] = calibration.orig_config[comp][va]
```

src/odemis/acq/align/fastem.py
```diff
@@ -233,9 +240,11 @@ def run(self):
 
             # loop over calibrations in list (order in list is important!)
             for calib in self.calibrations:
-                logging.debug("Starting calibration %s", calib)
+                calib_cls = calib.value
+                logging.debug("Starting calibration %s", calib_cls.__name__)
+                calib_cls = calib_cls(components)
```
pieleric, line 245: It's not really nice to override a class by an instance. Can you store the instance in a different variable? Like `calib_runner` (but a better name!) .

pieleric (changes requested):

pieleric (approved):

PR #2322, opened 2023-02-02
odemisd: add test for mdupdater

pieleric:
Simple test for the lens observer... more tests can come later.

### DIFF
--- a/src/odemis/odemisd/test/mdupdater_test.py
+++ b/src/odemis/odemisd/test/mdupdater_test.py
@@ -18,14 +18,17 @@ PURPOSE. See the GNU General Public License for more details.
 You should have received a copy of the GNU General Public License along with 
 Odemis. If not, see http://www.gnu.org/licenses/.
 """
+
+import unittest
+import unittest.mock
+
 import numpy
+
 from odemis import model
 from odemis.driver import static
 from odemis.model import Microscope
 from odemis.odemisd.mdupdater import MetadataUpdater
 from odemis.util import mock
-import unittest
-import unittest.mock
 
 
 class MDUpdaterTest(unittest.TestCase):
@@ -40,7 +43,8 @@ class MDUpdaterTest(unittest.TestCase):
 
         # Create Lens affects CCD (the role matters)
         lens = static.OpticalLens("Lens", "lens", mag=0.51, pole_pos=(458, 519), focus_dist=0.5e-3,
-                                  mirror_pos_top=[600.5, 0.2], mirror_pos_bottom=(-200, 0.3))
+                                  mirror_pos_top=[600.5, 0.2],  # Try with list, as when initialized from YAML it'll be a list
+                                  mirror_pos_bottom=(-200, 0.3))
         lens.affects.value = [ccd.name]
         
         # Mock model.getComponent()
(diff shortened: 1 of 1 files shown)


src/odemis/odemisd/test/mdupdater_test.py
```diff
@@ -0,0 +1,86 @@
...
+Odemis. If not, see http://www.gnu.org/licenses/.
+"""
+import numpy
+from odemis import model
+from odemis.driver import static
+from odemis.model import Microscope
+from odemis.odemisd.mdupdater import MetadataUpdater
+from odemis.util import mock
+import unittest
+import unittest.mock
```
nandishjpatel, line 28:
```suggestion
import unittest
import unittest.mock

import numpy

from odemis import model
from odemis.driver import static
from odemis.model import Microscope
from odemis.odemisd.mdupdater import MetadataUpdater
from odemis.util import mock
```

src/odemis/odemisd/test/mdupdater_test.py
```diff
@@ -0,0 +1,86 @@
...
+        # Create microscope (.Alive)
+        mic = Microscope("Fake mic", "secom")
+
+        # Create a CCD (the role matters)
+        img = model.DataArray(numpy.empty((512, 768), dtype=numpy.uint16))
+        ccd = mock.FakeCCD(img)  # role is always "ccd"
+
+        # Create Lens affects CCD (the role matters)
+        lens = static.OpticalLens("Lens", "lens", mag=0.51, pole_pos=(458, 519), focus_dist=0.5e-3,
+                                  mirror_pos_top=[600.5, 0.2], mirror_pos_bottom=(-200, 0.3))
```
nandishjpatel, line 43:
```suggestion
                                  mirror_pos_top=(600.5, 0.2), mirror_pos_bottom=(-200, 0.3))
```

nandishjpatel (approved):

src/odemis/odemisd/test/mdupdater_test.py
pieleric: This is on purpose, because when reading a YAML file, all the "tuples" are lists. So we need to make sure passing a list works fine. I'll add a comment.

src/odemis/odemisd/test/mdupdater_test.py
```diff
@@ -0,0 +1,86 @@
...
+        def fake_get_component(name):
+            for c in comps:
+                if c.name == name:
+                    return c
+            raise LookupError(f"no component {name}") 
+
+        with unittest.mock.patch.object(model, "getComponent", fake_get_component):
+
+            # Create a MetadataUpdater
+            mdup = MetadataUpdater("MDUpdater", mic)
```
sneepstefan, line 57: mdup is not used as a variable right?

```suggestion
            MetadataUpdater("MDUpdater", mic)
```

src/odemis/odemisd/test/mdupdater_test.py
```diff
@@ -0,0 +1,86 @@
...
+            mic.alive.value = {ccd, lens}
+
+            # Now, the lens should be observed: the VAs values should be copied on the CCD metadata
+            mdccd = ccd.getMetadata()
```
sneepstefan, line 61: I would use md_ccd as variable name, bit more clear and PEP8 friendly.

src/odemis/odemisd/test/mdupdater_test.py
pieleric: Actually, although we don't explicitly use it, we (theoretically) need to hold the instance, because it's working in the background (whenever a VA changes). So (theoretically), if we immediately discard the return value (ie, the instance), Python could delete the object. In practice, it's quite unlikely because the instance probably has some self references, and so it's need to go through the garbage collector, which, compared to the amount of code executed after, rarely runs. 

Anyway... I've found a even better way: I've added an explicitly call to `mdup.terminate()` at the end of the test. So we test more code, and it also makes this variable used.

sneepstefan: That's great Eric, and thanks for the explanation!

sneepstefan (approved):

PR #2324, opened 2023-02-03
Sdev 1294 autofocus calibration displays successful while the calibration failed

nandishjpatel:


src/odemis/acq/align/autofocus.py
```diff
@@ -462,15 +463,18 @@ def _DoBinaryFocus(future, detector, emt, focus, dfbkg, good_focus, rng_focus):
...
+            confidence = 0.1
         elif (best_fm - worst_fm) < best_fm * 0.5:
             # We can be confident of the data if there is a "big" (50%) difference
             # between the focus levels.
             logging.info("Auto focus indecisive but picking level %g @ %g m (lowest = %g)",
                          best_fm, best_pos, worst_fm)
+            confidence = 0.2
         else:
             logging.info("Auto focus found best level %g @ %g m", best_fm, best_pos)
+            confidence = 0.8
```
tepals, line 475: why not have `confidence = 1` here?

nandishjpatel: Maybe because the function can still be improved xP?

pieleric: When I suggested this value to Nandish it's because I have in mind that maybe later we can provide a more "fine-grain" confidence, going more smoothly between 0 and 1. I don't want code now to be written as `if confidence != 1: fail`. Giving an arbitrary value a little below 1 should hopefully avoid such code.
Also, as Nandish said, even at this point, we are not 100% certain that the focus is perfect, so let's not say 1.

src/odemis/acq/align/autofocus.py
```diff
@@ -249,6 +249,7 @@ def _DoBinaryFocus(future, detector, emt, focus, dfbkg, good_focus, rng_focus):
     returns:
         (float): Focus position (m)
         (float): Focus level
+        (float): Focus confidence (0<=f<=1)
```
pieleric, line 252: Could you add that 0 means it's certainly not in focus, and 1 means it's certainly the best focus possible.

pieleric (changes requested):

src/odemis/acq/align/autofocus.py
tepals: sounds good

tepals (approved):

pieleric (approved):

PR #2326, opened 2023-02-06
Bugfix in odemis.acq.align.FindGridSpots

effting:
Returned translation value was the initial estimation, not the refined value.


tepals:
I ran the test cases and they all still pass.

tepals (approved):

PR #2327, opened 2023-02-06
driver orsay: support changing the probeCurrent

pieleric:
The probe current cannot be changed directly. Instead, this is done by
loading a specific preset. To know what preset corresponds to which
probe current, we rely on a name following a specific template.

### DIFF
--- a/src/odemis/driver/orsay.py
+++ b/src/odemis/driver/orsay.py
@@ -20,20 +20,24 @@ You should have received a copy of the GNU General Public License along with
 Odemis. If not, see http://www.gnu.org/licenses/.
 """
 import collections.abc
-from odemis import model, util
-from odemis.model import isasync, CancellableThreadPoolExecutor, HwError, InstantaneousFuture, roattribute, \
-    MD_PIXEL_SIZE
-from odemis.util import almost_equal
-from ConsoleClient.Communication.Connection import Connection
-
-from functools import partial
+import inspect
+import logging
+import math
+import re
+import socket
 import threading
 import time
-import logging
-import inspect
+from functools import partial
 from math import pi
-import math
+from typing import Optional, Dict, Tuple
+
 import numpy
+from ConsoleClient.Communication.Connection import Connection
+
+from odemis import model, util
+from odemis.model import (MD_PIXEL_SIZE, CancellableThreadPoolExecutor,
+                          HwError, InstantaneousFuture, isasync, roattribute)
+from odemis.util import almost_equal
 
 VALVE_UNDEF = -1
 VALVE_TRANSIT = 0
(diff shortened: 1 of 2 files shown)


src/odemis/driver/orsay.py
```diff
@@ -434,6 +438,113 @@ def terminate(self):
...
+            raise
+
+        while not self._is_at_target_preset(preset_content, preset_mask_content):
+            time.sleep(0.5)
+            if time.time() > tstart + timeout:
+                raise TimeoutError(f"Setting preset {preset_name} timed out after {timeout} s")
+
+        logging.debug("Preset %s loaded", preset_name)
+
+    def _is_at_target_preset(self, preset, preset_mask) -> bool:
```
nandishjpatel, line 487: Unused argument `preset`

src/odemis/driver/orsay.py
```diff
@@ -434,6 +438,113 @@ def terminate(self):
...
+        for d in preset_mask.findall("Device"):
+            # sometimes the mask contains "Name" and sometime "name", so accept both
+            device_name = d.attrib['Name'] if 'Name' in d.attrib else d.attrib['name']
+            device = getattr(self.datamodel, device_name)
+            for p in d.findall("Parameter"):
+                param_name = p.attrib['name']
+                param = getattr(device, param_name)
+
+                if not bool(param.AtTarget):
+                    logging.debug(f"Preset {device_name}.{param_name} not yet at target: %s")
```
nandishjpatel, line 509:
```suggestion
                    logging.debug(f"Preset {device_name}.{param_name} not yet at target.")
```

src/odemis/driver/orsay.py
```diff
@@ -434,6 +438,113 @@ def terminate(self):
 
             super(OrsayComponent, self).terminate()
 
+    # Helper functions to handle presets (on top of the preset manager)
+    # They can be used by other class of this module, but otherwise are considered private
+    def _load_preset(self, preset_name: str, preset_mask: str, timeout: float=300):
```
nandishjpatel, line 443:
```suggestion
    def _load_preset(self, preset_name: str, preset_mask: str, timeout: float = 300):
```

src/odemis/driver/orsay.py
```diff
@@ -434,6 +438,113 @@ def terminate(self):
...
+        """
+        Lookup the names of the existing presets on the server
+        :return: name of every presets
+        """
+        # Contrary to its name, it doesn't return all the presets, but just a
+        # list with the preset names.
+        preset_summary = self.preset_manager.GetAllPresets()
+        return tuple(preset.get("name") for preset in preset_summary.iter("Preset"))
+
+    def _get_preset_setting(self, preset, sub_comp:str, setting:str, tag:str="Target") -> str:
```
nandishjpatel, line 524:
```suggestion
    def _get_preset_setting(self, preset, sub_comp: str, setting: str, tag: str = "Target") -> str:
```

src/odemis/driver/orsay.py
```diff
@@ -2871,113 +2983,166 @@ def __init__(self, name, role, parent, **kwargs):
...
-                               )
+        # It's not possible to change the probeCurrent directly. Instead, we
+        # rely on presets. They are created by the service engineer on the Orsay
+        # server, with specific names <current>pA_.... . Each preset with such
+        # name corresponds to one probe current. Whenever a new probe current is
+        # selected the preset is loaded. For the generic case, it's a little bit
+        # more complicated, because after changing the preset, the aperture should
+        # be checked that it's not worn out, and if it is, a new one should be
+        # used. That part is not handled in the driver, but at a higher level in
+        # Odemis (and for now, not at all... we just expect the user to manual
```
nandishjpatel, line 2994:
```suggestion
        # Odemis (and for now, not at all... we just expect the user to manually
```

src/odemis/driver/orsay.py
```diff
@@ -2871,113 +2983,166 @@ def __init__(self, name, role, parent, **kwargs):
...
+            current_str = preset_name.split("_")[1]
+            current = float(current_str[:-2]) * 1e-12
+            if current in current_presets:
+                logging.warning("Probe current %s pA defined in preset %s overridden by %s",
+                                current * 1e12, current_presets[current], preset_name)
+            current_presets[current] = preset_name
+
+        if not current_presets:
+            logging.warning("No probe current presets found. Make sure that the presets name are "
+                            "formatted as follows: CURRENTpA_EXTRA_INFO (e.g. '20pA_20um_25200V')")
```
nandishjpatel, line 3070:
```suggestion
                            "formatted as follows: NUMBER_CURRENTpA_EXTRA_INFO (e.g. '0002_20pA_20um25200V'")
```

src/odemis/driver/orsay.py
pieleric: Yes, I'm aware of it, but preset+preset_mask are typically always handled together. I consider the fact that we don't need the preset just an implementation detail. There could be slightly different implementation that would need the preset, so I prefer to keep the interface as it is.

K4rishma (approved):

nandishjpatel (approved):

PR #2328, opened 2023-02-07
`to_physical_space` and `to_pixel_index` to accept GeometricTransform

effting:
When [PR-2325](https://github.com/delmic/odemis/pull/2325) is merged, the updated `estimate_grid_orientation` and `estimate_grid_orientation_from_img` now return a GeometricTransform that works on image pixel coordinates. It is desirable to be able to convert this returned transform to work on physical coordinates. This pull request adds this functionality to `to_physical_space` and `to_pixel_index` in `odemis.util.transform`.

### DIFF
--- a/src/odemis/util/transform.py
+++ b/src/odemis/util/transform.py
@@ -121,10 +121,15 @@ def polar_to_cartesian(rho: numpy.ndarray, theta: numpy.ndarray) -> numpy.ndarra
 
 
 def to_physical_space(
-    ji: Union[Tuple[float, float], List[Tuple[float, float]], numpy.ndarray],
+    ji: Union[
+        Tuple[float, float],
+        List[Tuple[float, float]],
+        numpy.ndarray,
+        "GeometricTransform",
+    ],
     shape: Optional[Tuple[int, int]] = None,
     pixel_size: Optional[Union[float, Tuple[float, float]]] = None,
-) -> numpy.ndarray:
+) -> Union[numpy.ndarray, "AffineTransform"]:
     """
     Converts an image pixel index into a coordinate in physical space.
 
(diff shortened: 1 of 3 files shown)


src/odemis/util/transform.py
```diff
@@ -275,6 +302,17 @@ def to_pixel_index(
     array([ 0. ,  0.])
 
     """
+    if isinstance(xy, GeometricTransform):
```
tepals, line 305: If I understood correctly a `GeometricTransform` can be in physical and in pixel space, it just depends on what input it has. Does this code just assume the caller only calls `to_pixel_index` for geometric transforms that had coordinates in physical space as input and vice versa?

src/odemis/util/transform.py
```diff
@@ -224,10 +243,15 @@ def to_physical_space(
 
 
 def to_pixel_index(
```
tepals, line 245: I previously liked that these two functions `to_physical_space` and `to_pixel_index` were really clear and simple; the input and output are both a list of coordinates. 

We could also consider having a to_physical_space and a to_pixel_index method on the `GeometricTransform` class.

src/odemis/util/transform.py
effting: Typically a `GeometricTransform` is created using its `.from_poinset(src, dst)` method. Consider the following example:
```
src_ji = ...
dst_ji = ...
src_xy = to_physical_space(src_ji, shape, pixel_size)
dst_xy = to_physical_space(dst_ji, shape, pixel_size)
tform_ji = AffineTransform.from_pointset(src_ji, dst_ji)
tform_xy = AffineTransform.from_pointset(src_xy, dst_xy)
```
Then in general `tform_ji` is not equivalent to `tform_xy`. The modified code allows to convert between these two, and indeed assumes you only call `to_physical_space` on something in pixel indices.

src/odemis/util/transform.py
effting: Then I'd rather make it separate functions. Or just not merge this PR for now and see if it is really that necessary

pieleric: I agree with Thera that the original simple version of the function is nicer to read. Object-oriented pattern would indeed suggest to put such function as a method on the class.
That said, it might not be super easy due to (cycle-)dependencies. I'd say it's also fine to create new functions named something like `tform_to_physical_space` and `tform_to_pixel_index`, for just this purpose.

pieleric (changes requested):

src/odemis/util/transform.py
tepals: Érics idea of having two functions like `tform_to_physical_space` and `tform_to_pixel_index`, sounds like a good idea to me.

effting:
Decision is to not implement this in the current PR. In case of new implementation, a new PR will be made.

PR #2329, opened 2023-02-07
Update callers of FindGridSpots

effting:
Initial commit - to be tested and discussed

### DIFF
--- a/scripts/spot-grid.py
+++ b/scripts/spot-grid.py
@@ -9,9 +9,7 @@ This script provides a command line interface for displaying a video with a spot
 
 """
 import argparse
-import functools
 import logging
-import operator
 import os
 import sys
 import threading
@@ -71,13 +69,10 @@ class VideoDisplayerGrid(VideoDisplayer):
         grid = unit_gridpoints(self.gridsize, mode="ji")
         self.app.spots = tform_ji.apply(grid)
 
-        tform_xy = functools.reduce(
-            operator.matmul,
-            (
-                to_physical_space_transform(data.shape, self.app.pixel_size),
-                tform_ji,
-                to_pixel_index_transform(),
-            ),
+        tform_xy = (
+            to_physical_space_transform(data.shape, self.app.pixel_size)
+            @ tform_ji
+            @ to_pixel_index_transform()
         )
 
         self.app.translation = tform_xy.translation
--- a/scripts/spot-grid.py
+++ b/scripts/spot-grid.py
@@ -63,12 +63,12 @@ class VideoDisplayerGrid(VideoDisplayer):
         """
         tform_ji, _ = estimate_grid_orientation_from_img(
             data,
-            (8, 8),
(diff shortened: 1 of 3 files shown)


scripts/spot-grid.py
```diff
@@ -53,12 +58,22 @@ def new_image(self, data):
...
+            data,
+            (8, 8),
+            AffineTransform,
+            sigma=1.45,
+            threshold_rel=0.5,
+        )
+        grid = unit_gridpoints((8, 8), mode="ji")
+        self.app.spots = tform_ji.apply(grid)
+
+        tform_xy = to_physical_space(tform_ji, data.shape, pxsize)
```
tepals, line 71: `to_physical_space` needs to be imported and pxsize is not defined here, can be solved by setting self.pxsize in the init

tepals: I think there is also still a bug in the `to_physical_space` for a transform. I checked this code and `tform_xy.scale` is exactly equal to `tform_ji.scale`

effting:
I've updated the PR to something that works. Please give it a review.
I suggest we also check this on EA2 before merging

scripts/spot-grid.py
```diff
@@ -53,12 +61,30 @@ def new_image(self, data):
...
         """
-        self.app.spots, trans, scale, rot, shear = FindGridSpots(data, self.gridsize)
-        self.app.translation = trans[0], data.shape[0] - trans[1]
-        self.app.scale = scale
-        self.app.rotation = -rot
-        self.app.shear = shear
-
+        tform_ji, _ = estimate_grid_orientation_from_img(
+            data,
+            (8, 8),
```
tepals, line 66: you can use `self.gridsize` instead of (8, 8), to allow compatibility with images that have grids with different sizes.

tepals (changes requested):

scripts/spot-grid.py
```diff
@@ -53,12 +61,30 @@ def new_image(self, data):
...
+        self.app.spots = tform_ji.apply(grid)
+
+        tform_xy = functools.reduce(
+            operator.matmul,
+            (
+                to_physical_space_transform(data.shape, self.app.pixel_size),
+                tform_ji,
+                to_pixel_index_transform(),
+            ),
+        )
```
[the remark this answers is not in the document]
pieleric: The whole point of operators is to write them, like `@`. Now, for this current code, you have to check how `reduce` work, and read the whole `matmul` word, and realise that the next 3 arguments are actually 3 transforms. The explicit version is much easier to read.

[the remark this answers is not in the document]
effting: OK - changed it

scripts/spot-grid.py
effting: thanks

pieleric (approved):

tepals (approved):

PR #2331, opened 2023-02-10
fastem gui: fix for pop-up of calibration regions after de-selecting …

tepals:
…the scintillator

selecting a calibration region in the acquisition tab and afterwards deselecting the scintillator in the chamber tab, caused the calibration region to pop up randomly in the acquisition tab. With this fix, when a scintillator is removed from the list of active scintillators it will also toggle the is_selected attribute to False.


tepals:
see https://delmic.atlassian.net/browse/SDEV-1304

pieleric (approved):

nandishjpatel (approved):

PR #2332, opened 2023-02-10
Add support for moving to defined position in MIMAS, including the optical path manager

pieleric:
Add the 4 defined positions in MIMAS to the acq.move module, similarly to ENZEL and METEOR.
Then also add support for the OPM to switch automatically between FM and FIB based on the streams.

We also introduce a fix for the SmarPod simulator, and add a new MIMAS simulator using the Orsay simulator.

PR #2334, opened 2023-02-13
Milling manager

K4rishma:
Milling manager script with its test cases

### DIFF
--- a/src/odemis/acq/millmng.py
+++ b/src/odemis/acq/millmng.py
@@ -8,6 +8,7 @@ Assumptions:
 """
 import logging
 import math
+import os
 import threading
 import time
 from concurrent import futures
@@ -20,10 +21,10 @@ from odemis import model
 from odemis.acq.acqmng import acquire
 from odemis.acq.drift import AnchoredEstimator
 from odemis.acq.orsay_milling import mill_rectangle
-
 from odemis.acq.feature import CryoFeature
 from odemis.acq.stitching._tiledacq import MOVE_SPEED_DEFAULT
 from odemis.acq.stream import UNDEFINED_ROI
+from odemis.dataio import find_fittest_converter
 from odemis.util import executeAsyncTask, dataio
 
 ANCHOR_MAX_PIXELS = 512 ** 2  # max number of pixels for the anchor region
@@ -40,7 +41,7 @@ class MillingSettings(object):
                  dc_roi, dc_period, dc_dwell_time, dc_current):
         """
        Settings class for milling a rectangle.
-       
+
        :param name: (str) name for the given milling setting
        :param current: (float) Probe current, as available in the presets, based on the preset naming. The ion beam current in A
        :param horizontal_fov:(float) width of the FoV when milling (to be set on the ion-beam settings), keep it fixed below 45um
@@ -54,20 +55,16 @@ class MillingSettings(object):
        :param dc_period: (float) time of milling before running the drift correction in s
        :param dc_dwell_time: (float) dwell time to be used during anchor region acquisition
        :param dc_current: (float) drift correction current
-       :return: 
+       :return:
        """
-        current_choices = {20e-12, 100e-12, 350e-12, 760e-12}
-        dwell_time_choices = {1e-6, 10e-6}  # for drift correction
-        self.name = model.StringVA(name)
-        self.current = model.FloatEnumerated(current, unit="A",
-                                             choices=current_choices)
         self.horizontalFoV = model.FloatContinuous(horizontal_fov, unit="m", range=(5e-06, 45e-06))
+        self.current = model.FloatContinuous(current, unit="A", range=(0.5e-12, 3000e-12))
+        self.name = model.StringVA(name)
         self.roi = model.TupleContinuous(roi,
                                          range=((0, 0, 0, 0), (1, 1, 1, 1)),
                                          cls=(int, float))
         self.pixelSize = model.VigilantAttribute(pixel_size, unit="m")
-        # angles ot of this range would be dangerous for the hardware
-        # and probably not useful for the user
+        # angles ot of this range would be dangerous for the hardware and probably not useful for the user
         self.beamAngle = model.FloatContinuous(beam_angle, unit="rad", range=(math.radians(6), math.radians(10)))
         self.duration = model.FloatVA(duration, unit="s")
         # drift correction settings
@@ -75,10 +72,8 @@ class MillingSettings(object):
                                            range=((0, 0, 0, 0), (1, 1, 1, 1)),
                                            cls=(int, float))
         self.dcPeriod = model.FloatContinuous(dc_period, unit="s", range=(1, 180))
-        self.dcDwellTime = model.FloatEnumerated(dc_dwell_time, unit="s", choices=dwell_time_choices)
-        self.dcCurrent = model.FloatEnumerated(dc_current, unit="A",
-                                               choices=current_choices)
-
+        self.dcDwellTime = model.FloatContinuous(dc_dwell_time, unit="s", range=(1e-09, 1e-03))
+        self.dcCurrent = model.FloatContinuous(dc_current, unit="A", range=(0.5e-12, 1500e-12))
 
 def load_config(yaml_filename):
     """
@@ -96,47 +91,13 @@ def load_config(yaml_filename):
     return settings_dict
 
 
-def mill_features(millings: list, sites: list, feature_post_status, acq_streams, ebeam, sed, stage,
-                  aligner) -> futures.Future:
-    """
-    Mill features on the sample.
-    :param millings: (list of MillingSettings) Settings corresponding to each milling, to be milled in order
-    :param sites: (list of Features) Each Feature to be milled
-    :param feature_post_status: (str) value to set on the Feature at the end of a complete milling series
-    :param acq_streams: type of acquisition streams to be used for milling
-    :param ebeam: model component for the scanner
-    :param sed: model component for the detector
-    :param stage: model component for the stage
-    :param aligner: model component for the objective
-    :return: ProgressiveFuture
-    """
-    # Create a progressive future with running sub future
-    future = model.ProgressiveFuture()
-    future.running_subf = model.InstantaneousFuture()
-    future._task_lock = threading.Lock()
-    # create acquisition task
-    milling_task = MillingRectangleTask(future, millings, sites, feature_post_status, acq_streams, ebeam, sed, stage,
-                                        aligner)
-    # add the ability of cancelling the future during execution
-    future.task_canceller = milling_task.cancel
-
-    # set the progress of the future
-    total_duration = milling_task.estimate_milling_time()
-    future.set_end_time(time.time() + total_duration)
-
-    # assign the acquisition task to the future
-    executeAsyncTask(future, milling_task.run)
-
-    return future
-
-
 class MillingRectangleTask(object):
     """
     This class represents a milling Task for milling rectangular regions on the sample.
     """
 
     def __init__(self, future: futures.Future, millings: list, sites: list, feature_post_status: str, acq_streams,
-                 ebeam, sed, stage, aligner):
+                 ebeam, sed, stage, aligner, log_path=None):
         """
         Constructor
         :param future: (ProgressiveFuture) the future that will be executing the task
@@ -147,6 +108,8 @@ class MillingRectangleTask(object):
         :param ebeam: model component for the scanner
         :param sed: model component for the detector
         :param stage: model component for the stage
+        :param aligner: model component for the aligner
+        :param log_path: (str) path to the log anchor region acquisition used in drift correction
         """
         self._stage = stage
         self._scanner = ebeam
@@ -158,6 +121,15 @@ class MillingRectangleTask(object):
         self._feature_status = feature_post_status  # site and feature means the same
         self.sites = sites
         self.streams = acq_streams
+
+        self._log_path = log_path
+        if log_path:
+            filename = os.path.basename(self._log_path)
+            if not filename:
+                raise ValueError("Filename is not found on log path.")
+            self._exporter = find_fittest_converter(filename)
+            self._fn_bs, self._fn_ext = dataio.splitext(filename)
+            self._log_dir = os.path.dirname(self._log_path)
 
         # internal class requirements between functions
         self._pass_duration = []  # one value per milling setting, duration(s) after which drift acquisition starts
@@ -188,8 +160,10 @@ class MillingRectangleTask(object):
                 self._iterations.append(int(1))
                 self._pass_duration.append(setting.duration.value)
 
+        prev_stage_pos_ref = None  # previous stage position used for reference in estimating time
         for site in self.sites:
-            stage_time += self.estimate_stage_movement_time(site)
+            stage_time += self.estimate_stage_movement_time(site, prev_stage_pos_ref)
+            prev_stage_pos_ref = site.pos.value
 
         self._milling_time_estimate = time_estimate + drift_acq_time  # time_estimate per feature
         self._remaining_stage_time = stage_time  # total stage movement time
@@ -232,17 +206,20 @@ class MillingRectangleTask(object):
 
         return total_duration
 
-    def estimate_stage_movement_time(self, site: CryoFeature) -> float:
+    def estimate_stage_movement_time(self, site: CryoFeature, stage_pos_ref=None) -> float:
         """
         Estimation the time taken by the stage to move from current position to the location of given site
         :param site: (CryoFeature) consists X,Y,Z coordinates for stage location in m.
         :return: (float) time to move the stage between two points in s.
         """
         # current position from x,y,z of stage position and eliminating rx,ry,rz
-        current_pos = [p[1] for p in self._stage.position.value.items() if len(p[0]) == 1]
+        if stage_pos_ref is None:
+            current_pos = [p[1] for p in self._stage.position.value.items() if len(p[0]) == 1]
+        else:
+            current_pos = stage_pos_ref
         target_pos = site.pos.value  # list
         diff = [abs(target - current) for target, current in zip(target_pos, current_pos)]
-        stage_time = math.hypot(diff[0], diff[1], diff[2]) / self._move_speed
+        stage_time = math.sqrt(sum(d ** 2 for d in diff)) / self._move_speed
 
         return stage_time
 
@@ -252,15 +229,31 @@ class MillingRectangleTask(object):
         :param setting: (MillingSettings) milling settings
         return (float): estimated time to acquire 1 anchor area
         """
-        drift_acq_time = 0
         if setting.dcRoi.value == UNDEFINED_ROI:
             return 0
 
-        nb_anchor_scanning = math.ceil(setting.duration.value / setting.dcPeriod.value)  # repeat?
+        nb_anchor_scanning = math.ceil(setting.duration.value / setting.dcPeriod.value)
         drift_estimation = AnchoredEstimator(self._scanner, self._detector,
                                              setting.dcRoi.value, setting.dcDwellTime.value, ANCHOR_MAX_PIXELS,
                                              follow_drift=False)
         return drift_estimation.estimateAcquisitionTime() * nb_anchor_scanning
+
+    def _restore_beam_shift(self, current: float):
+        """
+        Restore the beam shift to previous shift after changing the beam current
+        :param current: New beam current
+        """
+        # check if scanner current is different from the one used for drift correction
+        if current != self._scanner.probeCurrent.value:
+            # store the scanner shift value used in previous current
+            previous_shift = self._scanner.shift.value
+            # set the scanner current to the one used for drift correction
+            self._scanner.probeCurrent.value = current
+            # previous and new current
+            logging.debug("Ion-beam current changed from %s to %s", self._scanner.probeCurrent.value,
+                          current)
+            # restore the scanner shift value
+            self._scanner.shift.value = previous_shift
 
     def _milling_setting_iterator(self, site: CryoFeature):
         """
@@ -278,18 +271,22 @@ class MillingRectangleTask(object):
             except TimeoutError:
                 logging.exception("Failed to tilt the stage for feature %s within %s s", site.name.value, t)
                 raise
-
-            # Configure current
-            self._scanner.probeCurrent.value = milling_settings.current.value
-
             # Configure HoV
             self._scanner.horizontalFoV.value = milling_settings.horizontalFoV.value
+
+            # Make the blanker automatic (ie, disabled when acquiring)
+            self._scanner.blanker.value = None
+
+            # set the milling current
+            self._restore_beam_shift(milling_settings.current.value)
 
             # Initialize the drift corrector only if it is requested by the user
             # max_pixels can change according to drift correction approach
             if milling_settings.dcRoi.value == UNDEFINED_ROI:
                 pass
             else:
+                # Change the current to the drift correction current and keep the beam shift value unchanged
+                self._restore_beam_shift(milling_settings.dcCurrent.value)
                 drift_est = AnchoredEstimator(self._scanner, self._detector,
                                               milling_settings.dcRoi.value,
                                               milling_settings.dcDwellTime.value,
@@ -298,6 +295,9 @@ class MillingRectangleTask(object):
                 drift_est.acquire()
                 da = drift_est.raw[-1]
                 pxs_anchor = da.metadata[model.MD_PIXEL_SIZE]
+                if self._log_path:
+                    fn = f"{self._fn_bs}-dc-{site.name.value}-{milling_settings.name.value}-0{self._fn_ext}"
+                    self._exporter.export(os.path.join(self._log_dir, fn), [da])
 
             # Compute the probe size and overlap, based on the requested pixel size
             # For the probe size, use the pixel size in the largest dimension (typically X)
@@ -323,9 +323,6 @@ class MillingRectangleTask(object):
             trans = [hori_t, vert_t]
 
             waiting_log_time = milling_settings.duration.value + self.estimate_drift_time(milling_settings)
-
-            # Make the blanker automatic (ie, disabled when acquiring)
-            self._scanner.blanker.value = None
 
             logging.debug(f"Milling setting: {milling_settings.name.value} for feature: {site.name.value}")
             logging.info(
@@ -353,6 +350,9 @@ class MillingRectangleTask(object):
                     if self._future._task_state == CANCELLED:
                         raise CancelledError()
 
+                    # restore the milling current
+                    self._restore_beam_shift(milling_settings.current.value)
+
                     self._future.running_subf = mill_rectangle(milling_settings.roi.value, self._scanner,
                                                                iteration=1,
                                                                duration=self._pass_duration[setting_nb],
@@ -365,6 +365,8 @@ class MillingRectangleTask(object):
                             f"Milling setting: {milling_settings.name.value} for feature: {site.name.value} at iteration: {itr} failed: {exp}")
                         raise
 
+                    # Change the current to the drift correction current and keep the beam shift value unchanged
+                    self._restore_beam_shift(milling_settings.dcCurrent.value)
                     # Acquire an image at the given location (RoI)
                     drift_est.acquire()
 
@@ -375,10 +377,12 @@ class MillingRectangleTask(object):
                     previous_shift = self._scanner.shift.value
                     logging.debug("Ion-beam shift in m : %s", previous_shift)
                     self._scanner.shift.value = (pxs_anchor[0] * drift_est.drift[0] + previous_shift[0],
-                                                 -(pxs_anchor[1] * drift_est.drift[1]) + previous_shift[
-                                                     1])  # shift in m - absolute position
+                                                 -(pxs_anchor[1] * drift_est.drift[1]) + previous_shift[1])  # shift in m - absolute position
                     logging.debug("New Ion-beam shift in m : %s", self._scanner.shift.value)
                     logging.debug("pixel size in m : %s", pxs_anchor)
+                    if self._log_path:
+                        fn = f"{self._fn_bs}-dc-{site.name.value}-{milling_settings.name.value}-{itr + 1}{self._fn_ext}"
+                        self._exporter.export(os.path.join(self._log_dir, fn), drift_est.raw[-1])
 
             self._scanner.blanker.value = True
 
@@ -388,8 +392,8 @@ class MillingRectangleTask(object):
     def _acquire_feature(self, site: CryoFeature) -> CryoFeature:
         """
         Acquire the data for the given feature.
-        :param site:
-        :return:
+        :param site: (CryoFeature) The feature to acquire.
+        :return: (CryoFeature) The feature with the acquired data.
         """
         data = []
         try:
@@ -415,7 +419,6 @@ class MillingRectangleTask(object):
             site.streams.value.extend(new_streams)
             logging.info("The acquisition for stream %s for feature %s is done", self.streams, site.name)
 
-        return site
 
     def run(self):
         """
@@ -472,7 +475,7 @@ class MillingRectangleTask(object):
                 logging.debug(f"The milling of feature {site.name.value} completed")
 
                 # Acquire the stream of milled site
-                site = self._acquire_feature(site)
+                self._acquire_feature(site)
 
                 # Store the actual time during milling one site computed after moving the stage
                 actual_time_per_site = time.time() - start_time
@@ -489,3 +492,36 @@ class MillingRectangleTask(object):
             # state that the future has finished
             with self._future._task_lock:
                 self._future._task_state = FINISHED
+
+def mill_features(millings: list, sites: list, feature_post_status, acq_streams, ebeam, sed, stage,
+                  aligner, log_path=None) -> futures.Future:
+    """
+    Mill features on the sample.
+    :param millings: (list of MillingSettings) Settings corresponding to each milling, to be milled in order
+    :param sites: (list of Features) Each Feature to be milled
+    :param feature_post_status: (str) value to set on the Feature at the end of a complete milling series
+    :param acq_streams: type of acquisition streams to be used for milling
+    :param ebeam: model component for the scanner
+    :param sed: model component for the detector
+    :param stage: model component for the stage
+    :param aligner: model component for the objective
+    :return: ProgressiveFuture
+    """
+    # Create a progressive future with running sub future
+    future = model.ProgressiveFuture()
+    future.running_subf = model.InstantaneousFuture()
+    future._task_lock = threading.Lock()
+    # create acquisition task
+    milling_task = MillingRectangleTask(future, millings, sites, feature_post_status, acq_streams, ebeam, sed, stage,
+                                        aligner, log_path)
+    # add the ability of cancelling the future during execution
+    future.task_canceller = milling_task.cancel
+
+    # set the progress of the future
+    total_duration = milling_task.estimate_milling_time()
+    future.set_end_time(time.time() + total_duration)
+
+    # assign the acquisition task to the future
+    executeAsyncTask(future, milling_task.run)
+
+    return future
--- a/src/odemis/acq/millmng.py
+++ b/src/odemis/acq/millmng.py
@@ -200,7 +200,7 @@ class MillingRectangleTask(object):
                 return False
             future._task_state = CANCELLED
             future.running_subf.cancel()
-            logging.debug("Milling procedure cancelled.")  # add line number
+            logging.debug("Milling procedure cancelled.")
         return True
 
     def estimate_milling_time(self, sites_done: int = 0, actual_time_per_site: float = None) -> float:
@@ -210,11 +210,6 @@ class MillingRectangleTask(object):
         :param actual_time_per_site: actual milling time measured for a single site
         :return: (float > 0): the estimated time is in seconds
         """
-        if len(self.sites) == 0:
-            logging.debug("Site location is not provided by the user, cancelling the milling")
-            self._future._task_state = CANCELLED
-            raise CancelledError()
-
         remaining_sites = (len(self.sites) - sites_done)
 
         if actual_time_per_site:
@@ -270,7 +265,6 @@ class MillingRectangleTask(object):
         target_pos = {"x": site.pos.value[0],
                       "y": site.pos.value[1],
                       "z": site.pos.value[2]}
-        # TODO check if it is wise to move in Z direction without affecting the OPTICAL-FIB alignment
         logging.debug("For feature %s moving the stage to %s m", site.name.value, target_pos)
         self._future.running_subf = self._stage.moveAbs(target_pos)
         stage_time = self.estimate_stage_movement_time(site)
@@ -402,12 +396,12 @@ class MillingRectangleTask(object):
         :param site: (CryoFeature) The feature to acquire.
         """
         data = []
-        # check if cancellation happened while the acquiring future is working
-        with self._future._task_lock:
-            if self._future._task_state == CANCELLED:
-                raise CancelledError()
         try:
-            self._future.running_subf = acquire(self.streams)
+            # check if cancellation happened while the acquiring future is working
+            with self._future._task_lock:
+                if self._future._task_state == CANCELLED:
+                    raise CancelledError()
+                self._future.running_subf = acquire(self.streams)
             data, exp = self._future.running_subf.result()
             if exp:
                 logging.warning(
--- a/src/odemis/acq/millmng.py
+++ b/src/odemis/acq/millmng.py
@@ -8,53 +8,25 @@ Assumptions:
 """
 import logging
 import math
-import os
 import threading
 import time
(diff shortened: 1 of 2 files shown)


src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,504 @@
...
+#     Returns
+#     -------
+#     dict
+#         Dictionary containing user input settings.
+#     """
+#     with open(yaml_filename, "r") as f:
+#         settings_dict = yaml.safe_load(f)
+#
+#     # settings = Settings(**settings_dict)  # convert to python dataclass
+#     return settings_dict
```
FCanberk, line 122:
```suggestion
# def load_config(yaml_filename):
#     """Load user input from yaml settings file.
#     Parameters
#     ----------
#     yaml_filename : str
#         Filename path of user configuration file.
#     Returns
#     -------
#     dict
#         Dictionary containing user input settings.
#     """
#     with open(yaml_filename, "r") as f:
#         settings_dict = yaml.safe_load(f)
#
#     # settings = Settings(**settings_dict)  # convert to python dataclass
#     return settings_dict
```

def load_config(yaml_filename):
    """Load user input from yaml settings file.
    Parameters
    ----------
    yaml_filename : str
        Filename path of user configuration file.
    Returns
    -------
    dict
        Dictionary containing user input settings.
    """
    try:
        with open(yaml_filename, "r") as f:
            settings_dict = yaml.safe_load(f)

    except yaml.YAMLError as exc:
        logging.error("Syntax error in milling settings yaml file: %s", exc)

    # settings = Settings(**settings_dict)  # convert to python dataclass
    return settings_dict

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,504 @@
...
+                raise CancelledError()
+            for site_idx, site in enumerate(self.sites):
+            # for/ site_idx in range(len(self.sites)):
+            #     site = self.sites[site_idx]
+                # Update progress on the milling sites left
+                start_time = time.time()
+                remaining_t = self.estimate_milling_time(site_idx, actual_time_per_site)
+                self._future.set_end_time(time.time() + remaining_t)
+
+                # .sites[idx].status already instantiated as FEATURE_ACTIVE
```
pieleric, line 323: We don't really know what the .status is, but that doesn't matter. => drop this comment.

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,504 @@
...
+
+                    try:
+                        t = 15  # s
+                        self._future.running_subf.result(timeout=t)
+                    except TimeoutError:
+                        logging.warning("Failed to tilt the stage for feature %s within %s s", site.name.value, t)
+                        self._future.running_subf.cancel()
+                        logging.debug("Cancelling the milling")
+                        self._future._task_state = CANCELLED
+                        raise CancelledError()
```
pieleric, line 355: Don't cancel => raise.

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,504 @@
...
+                    # Tilt stage angle
+                    stage_rx = self._ion_beam_angle - milling_settings.beamAngle.value
+                    logging.debug("Tilting stage to %f °", math.degrees(stage_rx))
+                    self._future.running_subf = self._stage.moveAbs({"rx": stage_rx})
+
+                    try:
+                        t = 15  # s
+                        self._future.running_subf.result(timeout=t)
+                    except TimeoutError:
+                        logging.warning("Failed to tilt the stage for feature %s within %s s", site.name.value, t)
```
pieleric, line 351: Should be `error()` or `exception()`.

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,504 @@
...
+                        self._future.running_subf.cancel()
+                        logging.debug("Cancelling the milling")
+                        self._future._task_state = CANCELLED
+                        raise CancelledError()
+
+                    # Configure current
+                    self._scanner.current.value = milling_settings.current.value
```
pieleric, line 358: On the scanner it's called `probeCurrent`.

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,504 @@
...
+
+                    # Configure HoV
+                    self._scanner.horizontalFoV.value = milling_settings.horizontalFoV.value
+
+                    # Initialize the drift corrector
+                    # max_pixels can change according to drift correction approach
+                    drift_est = AnchoredEstimator(self._scanner, self._detector,
+                                                  milling_settings.dcRoi.value,
+                                                  milling_settings.dcDwellTime.value,
+                                                  max_pixels=512 ** 2, follow_drift=False)
```
pieleric, line 368: Please place the max_pixels value (512 ** 2) into a constant at the top of the file (something like `ANCHOR_MAX_PIXELS = 512 ** 2`.

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,504 @@
...
+                    # acquire an image at the given location (RoI)
+                    drift_est.acquire()
+                    da = drift_est.raw[-1]
+                    pixel_size = da.metadata[model.MD_PIXEL_SIZE]
```
pieleric, line 372: That's not a very specific name... especially at the next variable is called `pxs`! 
I'd suggest `pxs_drift` or `pxs_anchor`.

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,504 @@
...
+
+                    # Compute the probe size and overlap, based on the requested pixel size
+                    # For the probe size, use the pixel size in the largest dimension (typically X)
+                    # To adjust the pixel size in Y, compute the overlap so that probe_size * (1-overlap)) == pixelSize.
+                    # X is horizontal axis, Y is vertical axis
+                    # TODO: does it work if pxs[0] < pxs[1] ?
```
pieleric, line 378: Yes, because now you're using `max()`. So you can drop this comment.

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,504 @@
...
+
+                    # Make the blanker automatic (ie, disabled when acquiring)
+                    self._scanner.blanker.value = None
+
+                    logging.debug(f"Milling setting: {milling_settings.name.value} for feature: {site.name.value}")
+                    logging.info(
+                        f"Milling a rectangle of dimensions: {dims} m which is {trans} m away from the center will "
+                        f"take approximately {waiting_log_time} s")
+
+                    if milling_settings.dcRoi.value != UNDEFINED_ROI:
```
pieleric, line 408: Please inverse the if condition, so that the "simple" version is first.

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,504 @@
...
+
+                # Check on the acquired data
+                if not data:
+                    logging.warning("The acquired data array in stream %s for feature %s is empty", self.streams,
+                                    site.name)
+                else:
+                    # Convert the data to StaticStreams, and add them to the feature
+                    new_streams = dataio.data_to_static_streams(data)
+                    site.streams.value.extend(new_streams)
+                    logging.info("The acquisition for stream %s for feature %s is done", self.streams, site.name)
```
pieleric, line 488: To keep the function a little shorter, I'd suggest to move this part to a separate function `_acquire_feature(future, feature, streams)`.

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,504 @@
...
+                        raise CancelledError()
+
+                    # Configure current
+                    self._scanner.current.value = milling_settings.current.value
+
+                    # Configure HoV
+                    self._scanner.horizontalFoV.value = milling_settings.horizontalFoV.value
+
+                    # Initialize the drift corrector
+                    # max_pixels can change according to drift correction approach
```
pieleric, line 364: Only do that if drift correction is requested.

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,504 @@
...
+                t = stage_time + 5  # adding extra margin
+                try:
+                    self._future.running_subf.result(t)  # or 2 min should be fast enough to move anywhere
+                except TimeoutError:
+                    logging.warning("Failed to move the stage for feature %s within %s s", site.name.value, t)
+                    self._future.running_subf.cancel()
+                    continue  # Try the next feature
+
+                # For each milling setting mill a rectangle for the given site
+                for setting_nb, milling_settings in enumerate(self._settings):
```
pieleric, line 341: To keep the function short, you could consider moving the main body of the for loop to a separate function.

src/odemis/acq/test/millmng_test.py
```diff
@@ -0,0 +1,351 @@
...
+from odemis.acq.feature import CryoFeature, FEATURE_ROUGH_MILLED
+from odemis.acq.millmng import mill_features, MillingSettings, MillingRectangleTask
+from odemis.acq.move import _isNearPosition
+from odemis.acq.stream import UNDEFINED_ROI
+from odemis.util import testing
+
+logging.basicConfig(format="%(asctime)s  %(levelname)-7s %(module)-15s: %(message)s")
+logging.getLogger().setLevel(logging.DEBUG)
+
+TEST_NOHW = os.environ.get("TEST_NOHW", "1")  # Default to simulation
```
pieleric, line 42: It should default to 0, that's the standard for everywhere in Odemis.

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,504 @@
...
+                self._future.running_subf = self._stage.moveAbs(target_pos)
+                stage_time = self.estimate_stage_movement_time(site)
+                t = stage_time + 5  # adding extra margin
+                try:
+                    self._future.running_subf.result(t)  # or 2 min should be fast enough to move anywhere
+                except TimeoutError:
+                    logging.warning("Failed to move the stage for feature %s within %s s", site.name.value, t)
+                    self._future.running_subf.cancel()
+                    continue  # Try the next feature
+
```
pieleric, line 339: Please add a `move.cryoSwitchSamplePosition(MILLING)`, so that we are certain the hardware is in the right position (in case of the FM streams changing to FM_IMAGING). 
For now, you can leave it commented out.

src/odemis/acq/test/millmng_test.py
```diff
@@ -0,0 +1,351 @@
...
+        total_time_with_drift = milling_setting_1.duration.value + milling_setting_2.duration.value + min_drift_time
+        self.assertGreaterEqual(time_estimated, total_time_with_drift,
+                                "estimated time with drift correction is less than the minimum time set to mill two features")
+
+        # check milling time without drift correction is less than milling with drift correction
+        total_est_time_with_drift = task.estimate_milling_time(site_idx=0)
+        self.assertLessEqual(total_est_time_without_drift, total_est_time_with_drift,
+                             "milling time with drift correction is less than milling time without drift correction")
+
+    @patch.object(odemis.acq.orsay_milling.OrsayMilling, 'do_milling', fake_do_milling)
```
pieleric, line 259: Ideally the patch should be done only if TEST_NOHW. To do so, you probably can place them in setUp() and tearDown()... but I don't know exactly how ;-)

pieleric (changes requested):

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,504 @@
...
+        # internal class requirements between functions
+        self._pass_duration = []  # one value per milling setting, duration(s) after which drift acquisition starts
+        self._iterations = []  # one value per milling setting, numbers of scans of .roi
+        stage_md = stage.getMetadata()
+        #todo ion beam angle not defined
+
+        # self._ion_beam_angle = stage_md[model.MD_ION_BEAM_TO_SAMPLE_ANGLE]
+        self._ion_beam_angle = math.radians(6)
+        self._move_speed = MOVE_SPEED_DEFAULT  # stage speed used from -tiledacq.py
```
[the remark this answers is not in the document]
K4rishma: Why do we add this TODO? I do not understand what does it suggest as TODO. In my comment " # stage speed used from -tiledacq.py" I meant that I used MOVE_SPEED_DEFAULT value from _tiledacq.py

src/odemis/acq/millmng.py
K4rishma: Should I cancel the sub_future in the except block? As it other sub-futures we are only using "raise" and logging commands.

src/odemis/acq/millmng.py
K4rishma: Used function as  _acquire_feature(self, site:CryoFeature) -> CryoFeature: in the class of the milling task. Would that work?

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,504 @@
...
+                self._iterations.append(int(numpy.ceil(setting.duration.value / setting.dcPeriod.value)))
+                self._pass_duration.append(setting.duration.value / self._iterations[setting_nb])
+
+                drift_acq_time += self.estimate_drift_time(setting)
+            else:
+                self._iterations.append(int(1))
+                self._pass_duration.append(setting.duration.value) 
+
+        for site in self.sites:
+            stage_time += self.estimate_stage_movement_time(site)
```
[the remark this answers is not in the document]
K4rishma: Good catch, I thought I was doing it as suggested i.e.  from the current position but it does not behave like that if I use it in the init like this. I will modify it.

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,504 @@
...
+        self._remaining_stage_time = stage_time  # total stage movement time
+
+        # set the FIB mode with the retracted state of the objective
+        # get the metadata to retract the objective
+        align_md = self._aligner.getMetadata()
+        deactive_pos = align_md[model.MD_FAV_POS_DEACTIVE]
+
+        # create a future and update the appropriate controls after it is called
+        logging.debug(f"Retracting the objective to set the imaging in FIB mode")
+        self._future.running_subf = self._aligner.moveAbs(deactive_pos)
```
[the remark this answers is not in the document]
K4rishma: will add a TODO

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,504 @@
...
+       :param beam_angle: 
+       :param duration: 
+       :param dc_roi: 
+       :param dc_period: 
+       :param dc_dwell_time:  do not ask user, depends on dc_current
+       :param dc_current:  do not ask user, keep it same as milling current
+       :return: 
+       """
+        current_choices = {20e-12, 100e-12, 350e-12, 760e-12}
+        dwell_time_choices = {1e-6, 10e-6} # for drift correction
```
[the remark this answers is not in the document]
K4rishma: I tried to use the second method like below. It appears that that the proxy VA has limited enumerated values for the probe current. Should I extend the proxyVA for probe current to include current_choices = {20e-12, 100e-12, 350e-12, 760e-12, 2.6e-09} ? The following error can then be resolved,   
 File "/home/dev/development/odemis/src/odemis/acq/test/millmng_test.py", line 158, in test_milling_settings
    milling_setting_1 = MillingSettings(name='rough_milling_1', current=20e-12, horizontal_fov=35e-6,
  File "/home/dev/development/odemis/src/odemis/acq/millmng.py", line 64, in __init__
    self.current.value = current
  File "/home/dev/development/odemis/src/odemis/model/_vattributes.py", line 203, in _set_value
    self._check(value) # we allow the setter to even put illegal value, it's the master
  File "/home/dev/development/odemis/src/odemis/model/_vattributes.py", line 1032, in _check
    Enumerated._check(self, value)
  File "/home/dev/development/odemis/src/odemis/model/_vattributes.py", line 985, in _check
    raise IndexError("Value %s is not part of possible choices: %s." %
IndexError: Value 2e-11 is not part of possible choices: 1.3e-09, 1e-10, 1.1564e-08, 2.6e-09, 3.4e-09, 2.3e-08.

install/linux/usr/share/odemis/sim/mimas-sim.odm.yaml
```diff
@@ -151,6 +151,7 @@ MIMAS-Sim: {
         inverted: ['y'],  # TODO Need to check that the rotations go counter-clockwise
     },
     metadata: {
+        ION_BEAM_TO_SAMPLE_ANGLE: 0.12217,
```
K4rishma, line 154: @pieleric What is the correct value to add here?

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,533 @@
...
+        if setting.dcRoi.value == UNDEFINED_ROI:
+            return 0
+
+        nb_anchor_scanning = math.ceil(setting.duration.value / setting.dcPeriod.value)
+        drift_estimation = AnchoredEstimator(self._scanner, self._detector,
+                                             setting.dcRoi.value, setting.dcDwellTime.value, ANCHOR_MAX_PIXELS,
+                                             follow_drift=False)
+        return drift_estimation.estimateAcquisitionTime() * nb_anchor_scanning
+
+    def _restore_beam_shift(self, current: float):
```
K4rishma, line 247: I added this function which changes current and restores the previous shift. The previous shift should already be drift compensated due to which I do not separately compensate for the drift. This approach is a little bit different from your suggestion to add drift. Is this way good to go?

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,533 @@
...
+            self._scanner.blanker.value = True
+
+            if self._future._task_state == CANCELLED:
+                raise CancelledError()
+
+    def _acquire_feature(self, site: CryoFeature) -> CryoFeature:
+        """
+        Acquire the data for the given feature.
+        :param site:
+        :return:
```
FCanberk, line 402: Also a brief explanation about the return value maybe

install/linux/usr/share/odemis/sim/mimas-sim.odm.yaml
pieleric: I already have a PR (#2332) which adds this metadata. In general, any value between 0 and 90° could be correct. On the current MIMAS, it's 8°.

FCanberk (approved):

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,527 @@
...
+            #     setting = self._settings[setting_nb]
+            time_estimate += setting.duration.value
+            # if self._settings[i]["dcROI"] != UNDEFINED_ROI:
+            if setting.dcRoi.value != UNDEFINED_ROI:
+                self._iterations.append(int(numpy.ceil(setting.duration.value / setting.dcPeriod.value)))
+                self._pass_duration.append(setting.duration.value / self._iterations[setting_nb])
+
+                drift_acq_time += self.estimate_drift_time(setting)
+            else:
+                self._iterations.append(int(1))
```
pieleric, line 160: `int(1)` -> `1`

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,527 @@
...
+            else:
+                for itr in range(self._iterations[setting_nb]):
+                    if self._future._task_state == CANCELLED:
+                        raise CancelledError()
+
+                    # restore the milling current
+                    self._restore_beam_shift(milling_settings.current.value)
+
+                    self._future.running_subf = mill_rectangle(milling_settings.roi.value, self._scanner,
+                                                               iteration=1,
```
pieleric, line 357: Please just write `self._iterations[setting_nb]` , to make it clear that it does nothing really special, and especially make it easy to change to a different number of iterations if we want to.

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,527 @@
...
+                       milling_settings.pixelSize.value[0]
+            vert_dim = (milling_settings.roi.value[3] - milling_settings.roi.value[1]) * \
+                       milling_settings.pixelSize.value[1]
+            dims = [hori_dim, vert_dim]
+
+            hori_t = ((milling_settings.roi.value[2] - milling_settings.roi.value[0]) / 2 - 0.5) * \
+                     milling_settings.pixelSize.value[0]
+            vert_t = ((milling_settings.roi.value[3] - milling_settings.roi.value[1]) / 2 - 0.5) * \
+                     milling_settings.pixelSize.value[1]
+            trans = [hori_t, vert_t]
```
pieleric, line 323: The correct formula should use the horizontalFoV instead of pixelSize. However, I'd suggest to drop all of this, and the logging.info(). `mill_rectangle()` is going to log almost the same thing. 
Possibly, you could move this logging to `do_fake_milling()` in the test cases, as that's a case where it would be handy.

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,527 @@
...
+                                             follow_drift=False)
+        return drift_estimation.estimateAcquisitionTime() * nb_anchor_scanning
+
+    def _restore_beam_shift(self, current: float):
+        """
+        Restore the beam shift to previous shift after changing the beam current
+        :param current: New beam current
+        """
+        # check if scanner current is different from the one used for drift correction
+        if current != self._scanner.probeCurrent.value:
```
pieleric, line 247: You can make it a little more readable by inverting the if:
```python
if current == self._scanner.probeCurrent.value:
    return
```
Then all the rest of the code has no indentation.

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,527 @@
...
+                    continue  # Try the next feature
+
+                # For each milling setting mill a rectangle for the given site
+                self._milling_setting_iterator(site)
+
+                # Update the status of milling for the given site
+                site.status = self._feature_status
+                logging.debug(f"The milling of feature {site.name.value} completed")
+
+                # Acquire the stream of milled site
```
pieleric, line 477:
```suggestion
                # Acquire the streams of the milled site
```

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,527 @@
...
+
+    def _acquire_feature(self, site: CryoFeature) -> CryoFeature:
+        """
+        Acquire the data for the given feature.
+        :param site: (CryoFeature) The feature to acquire.
+        :return: (CryoFeature) The feature with the acquired data.
+        """
+        data = []
+        try:
+            self._future.running_subf = acquire(self.streams)
```
pieleric, line 400: Add a check for cancellation, just before starting the acquisition.

pieleric (changes requested):

src/odemis/acq/millmng.py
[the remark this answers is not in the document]
K4rishma: @pieleric Do I need to update the probe current simsem.py according to the continuous float type used in milling settings? 
image

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,528 @@
...
+            raise ValueError("Ion beam angle not defined in stage metadata")
+        self._move_speed = MOVE_SPEED_DEFAULT  # stage speed used from _tiledacq.py
+
+        # estimate milling time from the list of milling settings-> millings
+        time_estimate = 0
+        drift_acq_time = 0
+        stage_time = 0
+        for setting_nb, setting in enumerate(self._settings):
+            # for setting_nb in range(len(self._settings)): # enumerate does not take a single-element list
+            #     setting = self._settings[setting_nb]
```
pieleric, line 150: Please drop these commented lines.

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,528 @@
...
+        self._remaining_stage_time -= self.estimate_stage_movement_time(self.sites[site_idx])
+        total_duration = milling_time + self._remaining_stage_time
+
+        return total_duration
+
+    def estimate_stage_movement_time(self, site: CryoFeature, stage_pos_ref=None) -> float:
+        """
+        Estimation the time taken by the stage to move from current position to the location of given site
+        :param site: (CryoFeature) consists X,Y,Z coordinates for stage location in m.
+        :return: (float) time to move the stage between two points in s.
```
pieleric, line 213: Add doc for stage_pos_ref

pieleric (changes requested):

src/odemis/acq/millmng.py
```diff
@@ -372,12 +383,14 @@ def _milling_setting_iterator(self, site: CryoFeature):
...
+                    self._drift = (pxs_anchor[0] * drift_est.drift[0],
+                                   -(pxs_anchor[1] * drift_est.drift[1]))
 
                     # Move FIB to compensate drift
                     previous_shift = self._scanner.shift.value
                     logging.debug("Ion-beam shift in m : %s", previous_shift)
-                    self._scanner.shift.value = (pxs_anchor[0] * drift_est.drift[0] + previous_shift[0],
-                                                 -(pxs_anchor[1] * drift_est.drift[1]) + previous_shift[1])  # shift in m - absolute position
+                    self._scanner.shift.value = (self._drift[0] + previous_shift[0],
+                                                 self._drift[1] + previous_shift[1])  # shift in m - absolute position
```
pieleric, line 393: There is something fishy here. `self._drift` is used  in `_restore_beam_shift()` to set the shift from the base shift (aka 0) to the current shift. However, here, scanner.shift should already contain the shift until the previous acquisition. So using drift_est.drift is fine here, but we should add up all the drifts since the beginning for the  `_restore_beam_shift()`.

Also, another difficulty is when the same current is used as the previous milling (task). In that case the current is not changed, so the shift says "good" (and the drift is only updated by a little bit). However, as soon as we change current, the shift is reset. => we need to reset the beam shift at the end of the task so that if another task is started after (without probe current change), the initial shift is the base shift.

src/odemis/acq/millmng.py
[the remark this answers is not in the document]
pieleric: Let's not change the simsem. Now the current is just a float continuous, and that should fit everything. For the test cases, we might need to be a little bit smarter by picking values for the milling settings based on the ion-beam.proveCurrent.choices (eg, min()).

src/odemis/acq/millmng.py
pieleric: I have been thinking a little bit more about the issue that the beam shift at the beginning might not be the base shift. A possible simple way to deal with it is to have a "reset_beam_shift()" function that would be called at the very beginning and ensure that the beam shift is at the base value (by changing to a different probe current and changing back to the previous probe current).

Another alternative would be to hide all these difficulties in the orsay driver. The beam shift VA would be (only) the extra value we add on top of the base shift... Then all these tricks to handle shift reset would go away. We would still provide the base shifts control, but via a metadata. That feel a little late to do such a change... But maybe that's the best way to handle it. Let me know how you feel about it.

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,519 @@
...
+                    raise CancelledError()
+            # set the FIB mode with the retracted state of the objective
+            # get the metadata to retract the objective
+            align_md = self._aligner.getMetadata()
+            deactive_pos = align_md[model.MD_FAV_POS_DEACTIVE]
+
+            # create a future and update the appropriate controls after it is called
+            # TODO: use move.cryoSwitchSamplePosition(MILLING)
+            logging.debug(f"Retracting the objective to set the imaging in FIB mode")
+            self._future.running_subf = self._aligner.moveAbs(deactive_pos)
```
pieleric, line 444: Please use move.cryoSwitchSamplePosition(). It's now available on master.

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,519 @@
...
+                start_time = time.time()
+                remaining_t = self.estimate_milling_time(site_idx, actual_time_per_site)
+                self._future.set_end_time(time.time() + remaining_t)
+
+                # For one given site, move the stage at the site and
+                # do milling with all milling settings for the given site
+                self._mill_all_settings(site)
+                # If timeout exception is raised during stage movement, continue milling the next site
+                if Exception == TimeoutError:
+                    continue
```
pieleric, line 464: That code makes no sense. `Exception` is a class. `TimeoutError` is a class. They are never equal.
I don't have a really nice way to handle the special case that the stage failed to move, so we just skipped the milling. Maybe make mill_all_settings() return True/False. Or crease a special exception, like "MoveError", to indicate this specific failure type (and use a try/except block).

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,519 @@
...
+            self._scanner.blanker.value = None
+
+            # set the milling current
+            self._scanner.probeCurrent.value = milling_settings.current.value
+
+            # Initialize the drift corrector only if it is requested by the user
+            # max_pixels can change according to drift correction approach
+            if milling_settings.dcRoi.value == UNDEFINED_ROI:
+                pass
+            else:
```
pieleric, line 302:
```suggestion
            if milling_settings.dcRoi.value != UNDEFINED_ROI:
```

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,519 @@
...
+            self._scanner.horizontalFoV.value = milling_settings.horizontalFoV.value
+
+            # Make the blanker automatic (ie, disabled when acquiring)
+            self._scanner.blanker.value = None
+
+            # set the milling current
+            self._scanner.probeCurrent.value = milling_settings.current.value
+
+            # Initialize the drift corrector only if it is requested by the user
+            # max_pixels can change according to drift correction approach
```
pieleric, line 299: I don't think that this `max_pixels` comment is very useful. For now it never changes. So let's drop this comment, or mark a TODO with an explanation what would need to be done.

pieleric (changes requested):

src/odemis/acq/millmng.py
K4rishma: Adding a special class MoveError(Exception) at line 94

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,526 @@
...
+        # Todo check if it is wise to move in Z direction without affecting the OPTICAL-FIB alignment
+        logging.debug("For feature %s moving the stage to %s m", site.name.value, target_pos)
+        self._future.running_subf = self._stage.moveAbs(target_pos)
+        stage_time = self.estimate_stage_movement_time(site)
+        t = stage_time + 5  # adding extra margin
+        try:
+            self._future.running_subf.result(t)  # or 2 min should be fast enough to move anywhere
+        except TimeoutError:
+            self._future.running_subf.cancel()
+            move.cryoSwitchSamplePosition(MILLING)
```
pieleric, line 271: Why starting to move to MILLING in case it fails to move? Drop.

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,526 @@
...
+        :param site: (CryoFeature) The site to move to.
+        """
+        target_pos = {"x": site.pos.value[0],
+                      "y": site.pos.value[1],
+                      "z": site.pos.value[2]}
+        # Todo check if it is wise to move in Z direction without affecting the OPTICAL-FIB alignment
+        logging.debug("For feature %s moving the stage to %s m", site.name.value, target_pos)
+        self._future.running_subf = self._stage.moveAbs(target_pos)
+        stage_time = self.estimate_stage_movement_time(site)
+        t = stage_time + 5  # adding extra margin
```
pieleric, line 266: That's really really not enough. So far the stage move is estimated based on an arbitrary speed. Not even the speed reported by the stage! So at least you should multiply the `stage_time * 10` for a timeout.

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,526 @@
...
+        nb_anchor_scanning = math.ceil(setting.duration.value / setting.dcPeriod.value)
+        drift_estimation = AnchoredEstimator(self._scanner, self._detector,
+                                             setting.dcRoi.value, setting.dcDwellTime.value, ANCHOR_MAX_PIXELS,
+                                             follow_drift=False)
+        return drift_estimation.estimateAcquisitionTime() * nb_anchor_scanning
+
+    def _move_to_site(self, site: CryoFeature):
+        """
+        Move the stage to the given site.
+        :param site: (CryoFeature) The site to move to.
```
pieleric, line 257: Please indicate it can raise MoveError.

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,526 @@
...
+
+        # internal class requirements between functions
+        self._pass_duration = []  # one value per milling setting, duration(s) after which drift acquisition starts
+        self._iterations = []  # one value per milling setting, numbers of scans of .roi
+        stage_md = stage.getMetadata()
+        try:
+            self._ion_beam_angle = stage_md[model.MD_ION_BEAM_TO_SAMPLE_ANGLE]
+        except KeyError:
+            raise ValueError("Ion beam angle not defined in stage metadata")
+        self._move_speed = MOVE_SPEED_DEFAULT  # stage speed used from _tiledacq.py
```
pieleric, line 155: Contrarily to what is suggested by the comment, that's *not* how tiledacq.py does estimate the move speed. Please take the whole code from there. It starts by trying to get the speed reported by the stage. MOVE_SPEED_DEFAULT is only a fallback.

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,526 @@
...
+        prev_stage_pos_ref = None  # previous stage position used for reference in estimating time
+
+        for site in self.sites:
+            stage_time += self.estimate_stage_movement_time(site, prev_stage_pos_ref)
+            prev_stage_pos_ref = site.pos.value
+
+        self._milling_time_estimate = time_estimate + drift_acq_time  # time_estimate per feature
+        self._remaining_stage_time = stage_time  # total stage movement time
+
+    def cancel(self, future):
```
pieleric, line 182: Add return type `-> bool`.

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,526 @@
...
+                return False
+            future._task_state = CANCELLED
+            future.running_subf.cancel()
+            logging.debug("Milling procedure cancelled.")  # add line number
+        return True
+
+    def estimate_milling_time(self, site_idx: int = 0, actual_time_per_site: float = None) -> float:
+        """
+        Estimates the milling time for the given feature.
+        :param site_idx: (int) index of the site to be milled
```
pieleric, line 199: "index of the site" is badly worded. Yes, that sort of is true, if you call this before you actually mill the site, and you index the sites from 0. I'd suggest to just describe as:
```suggestion
        :param sites_done: (int) number of sites already milled
```

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,526 @@
...
+            # X is horizontal axis, Y is vertical axis
+            pxs = milling_settings.pixelSize.value
+            probe_size = max(pxs)
+            overlap = [1 - (pxs[0] / probe_size), 1 - (pxs[1] / probe_size)]  # Always >= 0
+
+            # check if cancellation happened
+            with self._future._task_lock:
+                if self._future._task_state == CANCELLED:
+                    raise CancelledError()
+
```
pieleric, line 333: Drop that, as you're checking it immediately after again.

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,526 @@
...
+                        logging.debug(
+                            f"Milling setting: {milling_settings.name.value} for feature: {site.name.value} at iteration: {itr} failed: {exp}")
+                        raise
+
+                    # Change the current to the drift correction current
+                    self._scanner.probeCurrent.value = milling_settings.dcCurrent.value
+                    # Acquire an image at the given location (RoI)
+                    drift_est.acquire()
+
+                    # Estimate the drift
```
pieleric, line 380:
```suggestion
                    # Estimate the drift since last correction
```

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,526 @@
...
+                    drift = (pxs_anchor[0] * drift_est.drift[0],
+                             -(pxs_anchor[1] * drift_est.drift[1]))
+
+                    # Move FIB to compensate drift
+                    previous_shift = self._scanner.shift.value
+                    logging.debug("Ion-beam shift in m : %s", previous_shift)
+                    self._scanner.shift.value = (drift[0] + previous_shift[0],
+                                                 drift[1] + previous_shift[1])  # shift in m - absolute position
+                    logging.debug("New Ion-beam shift in m : %s", self._scanner.shift.value)
+                    logging.debug("pixel size in m : %s", pxs_anchor)
```
pieleric, line 391: That's really not so useful info => drop.

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,526 @@
...
+                    drift_est.estimate()
+                    drift = (pxs_anchor[0] * drift_est.drift[0],
+                             -(pxs_anchor[1] * drift_est.drift[1]))
+
+                    # Move FIB to compensate drift
+                    previous_shift = self._scanner.shift.value
+                    logging.debug("Ion-beam shift in m : %s", previous_shift)
+                    self._scanner.shift.value = (drift[0] + previous_shift[0],
+                                                 drift[1] + previous_shift[1])  # shift in m - absolute position
+                    logging.debug("New Ion-beam shift in m : %s", self._scanner.shift.value)
```
pieleric, line 390: Can you merge the logging lfor the previous and new ion-beam shifts into a single logging line? That'd make it more compact in the logs.

src/odemis/acq/test/millmng_test.py
```diff
@@ -0,0 +1,332 @@
...
+        # reset to current stage position
+        self.stage.moveAbs(self.target_position[0]).result()
+
+    def test_milling_settings(self):
+        """
+        Test the values in milling settings with predefined values
+        """
+        milling_setting_1 = MillingSettings(name='rough_milling_1', current=self.probe_current, horizontal_fov=35e-6,
+                                            roi=(0.5, 0.5, 0.8, 0.8),
+                                            pixel_size=(3.41796875e-08, 3.41796875e-08), beam_angle=self.beam_angle,
```
pieleric, line 153: Why such value for the pixel size? You decide it, and it's quite arbitrary (it actually depends on the probe size, which depends on the probe current). One thousands for the FoV would probably be fine for here => 35e-9, or even 10e-9. Something more round brings the feeling that it's arbitrary.

src/odemis/acq/test/millmng_test.py
```diff
@@ -0,0 +1,332 @@
...
+                                            duration=20,
+                                            dc_roi=(0, 0.3, 0.4, 0.7), dc_period=10, dc_dwell_time=10e-6,
+                                            dc_current=self.probe_current)
+
+        milling_setting_2 = MillingSettings(name='rough_milling_2', current=self.probe_current, horizontal_fov=35e-6,
+                                            roi=(0.5, 0.5, 0.8, 0.8),
+                                            pixel_size=(3.41796875e-08, 3.41796875e-08), beam_angle=self.beam_angle,
```
pieleric, line 217: Please try some milling settings with non-square pixel_size. Typically, for the MIMAS the Y should be smaller that X (like, 6 times smaller). However, it'd be good to try also when X is smaller than Y.

pieleric (changes requested):

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,504 @@
...
+        :return:
+        """
+        """
+        Estimates the milling time for the given feature.
+        :return (Float > 0): the estimated time
+        """
+        if len(self.sites) == 0:
+            logging.debug("Site location is not provided by the user, cancelling the milling")
+            self._future._task_state = CANCELLED
+            raise CancelledError()
```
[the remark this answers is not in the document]
pieleric: Please drop this whole block of code.

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,512 @@
...
+        :return: True if it successfully cancelled (stopped) the future
+        """
+        logging.debug("Canceling milling procedure...")
+
+        with future._task_lock:
+            if future._task_state == FINISHED:
+                return False
+            future._task_state = CANCELLED
+            future.running_subf.cancel()
+            logging.debug("Milling procedure cancelled.")  # add line number
```
pieleric, line 203: What is this comment about? logging automatically indicates the line number if the logging format requests so.
=> drop this comment?

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,512 @@
...
+    def _move_to_site(self, site: CryoFeature):
+        """
+        Move the stage to the given site.
+        :param site: (CryoFeature) The site to move to.
+        :raises MoveError: if the stage failed to move to the given site.
+        """
+        target_pos = {"x": site.pos.value[0],
+                      "y": site.pos.value[1],
+                      "z": site.pos.value[2]}
+        # TODO check if it is wise to move in Z direction without affecting the OPTICAL-FIB alignment
```
pieleric, line 273: Yes, that's the point of moving in Z: to keep the optical-FIB alignment.
=> Drop this comment.

src/odemis/acq/millmng.py
```diff
@@ -0,0 +1,512 @@
...
+    def _acquire_feature(self, site: CryoFeature):
+        """
+        Acquire the data for the given feature and add to the given site.
+        :param site: (CryoFeature) The feature to acquire.
+        """
+        data = []
+        # check if cancellation happened while the acquiring future is working
+        with self._future._task_lock:
+            if self._future._task_state == CANCELLED:
+                raise CancelledError()
```
pieleric, line 408: You should move this lock down, around the line 410 (inside the try block)

pieleric (changes requested):

pieleric (approved):

PR #2339, opened 2023-02-17
util registration: Add error metric in estimate_grid_orientation, to …

tepals:
…give an estimate for the quality of the grid estimation.


src/odemis/util/registration.py
```diff
@@ -531,11 +535,15 @@ def estimate_grid_orientation_from_img(
...
 
     """
     if num_spots is None:
         num_spots = shape[0] * shape[1]
     elif num_spots == 0:
         num_spots = None
     ji = find_spot_positions(image, sigma, threshold_abs, threshold_rel, num_spots)
-    return estimate_grid_orientation(ji, shape, transform_type)
+    tform, error_metric = estimate_grid_orientation(ji, shape, transform_type)
+    return tform, error_metric
```
effting, line 549: No need to modify? the old `return estimate_grid_orientation(ji, shape, transform_type)` would still work here.

tepals: Good catch, I'll change it back

effting (approved):

nandishjpatel (approved):

PR #2340, opened 2023-02-17
[hotfix] remove unused import from tiledacq that caused fastem-calibr…

tepals:
…ations CI to fail

### DIFF
--- a/src/odemis/acq/stitching/_tiledacq.py
+++ b/src/odemis/acq/stitching/_tiledacq.py
@@ -34,7 +34,6 @@ from odemis.acq.stitching._simple import register, weave
 from odemis.acq.stream import Stream, EMStream, ARStream, \
     SpectrumStream, FluoStream, MultipleDetectorStream, util, executeAsyncTask, \
     CLStream
-from odemis.gui import conf
 from odemis.gui.util.raster import point_in_polygon
 from odemis.model import DataArray
 from odemis.util import dataio as udataio, img, linalg


nandishjpatel (approved):

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -34,7 +34,6 @@
 from odemis.acq.stream import Stream, EMStream, ARStream, \
     SpectrumStream, FluoStream, MultipleDetectorStream, util, executeAsyncTask, \
     CLStream
-from odemis.gui import conf
 from odemis.gui.util.raster import point_in_polygon
```
pieleric, line 37: Actually, that import is also a bit fishy: normally the .acq should not have any import from .gui . If we need this function, we probably could move gui.util.raster to util.raster.

pieleric (approved):

pieleric:
Could you explain in the commit log how an import can cause the CI to fail?

src/odemis/acq/stitching/_tiledacq.py
tepals: good point. I don't really see a reason why `raster.py` is in `gui.util` and it is a very useful function to have, so I will move `gui.util.raster` to `util.raster`

tepals: should I do this in a separate PR?

pieleric: Yes, that's probably the best.

PR #2342, opened 2023-02-17
Test cases to support autofocus functions

K4rishma:
add test cases to support autofocus overview acquisition

test cases for roi_autofocus_test.py
test cases for finding autofocus in _tiledacq.py
test cases to support plane fitting in linalg.py
add metadata in enzel sim microscope file

### DIFF
--- a/src/odemis/acq/align/test/roi_autofocus_test.py
+++ b/src/odemis/acq/align/test/roi_autofocus_test.py
@@ -26,18 +26,8 @@ from odemis.util import testing
 logging.getLogger().setLevel(logging.DEBUG)
 logging.basicConfig(format="%(asctime)s  %(levelname)-7s %(module)s:%(lineno)d %(message)s")
 
-TEST_NOHW = os.environ.get("TEST_NOHW", "1")  # Default to simulation
-
-if TEST_NOHW == "0":
-    TEST_NOHW = False
-elif TEST_NOHW == "1":
-    TEST_NOHW = True
-else:
-    raise ValueError("Unknown value of environment variable TEST_NOHW=%s" % TEST_NOHW)
-
-# CONFIG_PATH = os.path.dirname(odemis.__file__) + "/../../install/linux/usr/share/odemis/"
-# METEOR_CONFIG = CONFIG_PATH + "sim/meteor-sim.odm.yaml"
-METEOR_CONFIG = "/home/dev/development/odemis/install/linux/usr/share/odemis/sim/meteor-sim.odm.yaml"
+CONFIG_PATH = os.path.dirname(odemis.__file__) + "/../../install/linux/usr/share/odemis/"
+METEOR_CONFIG = CONFIG_PATH + "sim/meteor-sim.odm.yaml"
 
 
 class RoiAutofocusTestCase(unittest.TestCase):
@@ -47,8 +37,7 @@ class RoiAutofocusTestCase(unittest.TestCase):
 
     @classmethod
     def setUpClass(cls):
-        if TEST_NOHW is True:
-            testing.start_backend(METEOR_CONFIG)
+        testing.start_backend(METEOR_CONFIG)
 
         # create some streams connected to the backend
         # cls.microscope = model.getMicroscope()
@@ -121,7 +110,8 @@ class RoiAutofocusTestCase(unittest.TestCase):
 
         bbox = (xmin, ymin, xmax, ymax)
 
-        f = autofocus_in_roi(bbox, self.stage, self.ccd, self.focus, self.focus_range, n_tiles[0], n_tiles[1], 0) # TODO add proper confidence value
+        f = autofocus_in_roi(bbox, self.stage, self.ccd, self.focus, self.focus_range, n_tiles[0], n_tiles[1],
+                             0)  # TODO add proper confidence value
 
         # Test cancelling of autofocus in roi
         time.sleep(2)
--- a/src/odemis/acq/align/test/roi_autofocus_test.py
+++ b/src/odemis/acq/align/test/roi_autofocus_test.py
@@ -1,0 +1,147 @@
+# -*- coding: utf-8 -*-
+"""
+Copyright © 2020 Delmic
+This file is part of Odemis.
+Odemis is free software: you can redistribute it and/or modify it under the terms
+of the GNU General Public License version 2 as published by the Free Software
+Foundation.
+Odemis is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+PURPOSE. See the GNU General Public License for more details.
+You should have received a copy of the GNU General Public License along with
+Odemis. If not, see http://www.gnu.org/licenses/.
+"""
+import logging
+import os
+import time
+import unittest
+from concurrent import futures
+
+import odemis
+from odemis import model
+from odemis.acq.align.autofocus import estimateAutoFocusTime
+from odemis.acq.align.roi_autofocus import autofocus_in_roi, estimate_autofocus_in_roi_time
+from odemis.util import testing
+
+logging.getLogger().setLevel(logging.DEBUG)
+logging.basicConfig(format="%(asctime)s  %(levelname)-7s %(module)s:%(lineno)d %(message)s")
+
+TEST_NOHW = os.environ.get("TEST_NOHW", "1")  # Default to simulation
+
+if TEST_NOHW == "0":
+    TEST_NOHW = False
+elif TEST_NOHW == "1":
+    TEST_NOHW = True
+else:
+    raise ValueError("Unknown value of environment variable TEST_NOHW=%s" % TEST_NOHW)
+
+# CONFIG_PATH = os.path.dirname(odemis.__file__) + "/../../install/linux/usr/share/odemis/"
+# METEOR_CONFIG = CONFIG_PATH + "sim/meteor-sim.odm.yaml"
+METEOR_CONFIG = "/home/dev/development/odemis/install/linux/usr/share/odemis/sim/meteor-sim.odm.yaml"
+
+
+class RoiAutofocusTestCase(unittest.TestCase):
+    """
+    Unit test for the roi_autofocus module
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        if TEST_NOHW is True:
+            testing.start_backend(METEOR_CONFIG)
+
+        # create some streams connected to the backend
+        # cls.microscope = model.getMicroscope()
+        cls.ccd = model.getComponent(role="ccd")
+        cls.focus = model.getComponent(role="focus")
+        cls.stage = model.getComponent(role="stage")
+
+        # Assumes the stage is referenced
+        cls.init_pos = (cls.stage.position.value["x"], cls.stage.position.value["y"])
+        cls.width, cls.height = cls.ccd.shape[:2]
+        cls.px_size = cls.ccd.getMetadata().get(model.MD_PIXEL_SIZE)
+
+        # Assume initial focus position is good
+        # The initial position comes from manual focussing the centre of overview image
+        current_focus = cls.focus.position.value["z"]  # assume manual focus before running the script
+        # Allow a range of +/- 30 µm around the focused position
+        cls.focus_range = (current_focus - 30.0e-6, current_focus + 30.0e-6)
+
+    @classmethod
+    def tearDownClass(cls):
+        logging.debug("moving back to initial position")
+        cls.stage.moveAbs({"x": cls.init_pos[0], "y": cls.init_pos[1]}).result()
+
+    def test_autofocus_in_roi(self):
+        """
+        Tests the autofus in roi is running and returns the focus points:
+        """
+        init_pos = self.init_pos
+        px_size = self.px_size
+        width = self.width
+        height = self.height
+        n_tiles = (3, 3)
+        overlap = 0.2
+
+        xmin = init_pos[0] - (1 - overlap) * n_tiles[0] / 2 * px_size[0] * width
+        ymin = init_pos[1] - (1 - overlap) * n_tiles[1] / 2 * px_size[1] * height
+        xmax = init_pos[0] + (1 - overlap) * n_tiles[0] / 2 * px_size[0] * width
+        ymax = init_pos[1] + (1 - overlap) * n_tiles[1] / 2 * px_size[1] * height
+
+        bbox = (xmin, ymin, xmax, ymax)
+
+        f = autofocus_in_roi(bbox, self.stage, self.ccd, self.focus, self.focus_range, n_tiles[0], n_tiles[1], 0)
+
+        # Test if the autofocus in roi is running
+        time.sleep(2)
+        self.assertTrue(f.running())
+
+        # The returned focus points should be minimum three
+        focus_points = f.result()
+        self.assertGreaterEqual(len(focus_points), 3)
+
+        # Test if the autofocus task is done
+        self.assertTrue(f.done())
+
+    def test_cancel_autofocus_in_roi(self):
+        """
+        Tests cancelling of autofocus
+        """
+        init_pos = self.init_pos
+        px_size = self.px_size
+        width = self.width
+        height = self.height
+        n_tiles = (3, 3)
+        overlap = 0.2
+
+        xmin = init_pos[0] - (1 - overlap) * n_tiles[0] / 2 * px_size[0] * width
+        ymin = init_pos[1] - (1 - overlap) * n_tiles[1] / 2 * px_size[1] * height
+        xmax = init_pos[0] + (1 - overlap) * n_tiles[0] / 2 * px_size[0] * width
+        ymax = init_pos[1] + (1 - overlap) * n_tiles[1] / 2 * px_size[1] * height
+
+        bbox = (xmin, ymin, xmax, ymax)
+
+        f = autofocus_in_roi(bbox, self.stage, self.ccd, self.focus, self.focus_range, n_tiles[0], n_tiles[1], 0) # TODO add proper confidence value
+
+        # Test cancelling of autofocus in roi
+        time.sleep(2)
+        self.assertTrue(f.running())
+
+        # cancel the autofocussing after 5 seconds
+        time.sleep(5)
+        f.cancel()
+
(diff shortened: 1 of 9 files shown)


src/odemis/acq/stitching/test/tiledacq_test.py
```diff
@@ -311,6 +311,119 @@ def test_whole_procedure(self):
...
+        self.assertGreaterEqual(tiled_acq_task._zlevels[0], zmin)
+        self.assertLessEqual(tiled_acq_task._zlevels[-1], zmax)
+
+        # Test 2 when z level is provided
+        zlevels = [focus_points[0][2], focus_points[1][2], focus_points[2][2]]
+
+        with self.assertRaises(NotImplementedError):
+            tiled_acq_task = TiledAcquisitionTask(self.fm_streams, self.stage,
+                                                  area=area, overlap=overlap, future=model.InstantaneousFuture(),
+                                                  zlevels=zlevels, focus_points=focus_points)
```
tepals, line 350: we actually need `zlevels` and `focus_points` to be defined for the `refocus` method to work. The `NotImplementedError` gets raised because you do not set the focusing_method to FocusingMethod.MAX_INTENSITY_PROJECTION. So maybe the code on line 161 of _tiledacq.py should have an additional check:

```
if len(self._zlevels) > 1 and focusing_method != FocusingMethod.MAX_INTENSITY_PROJECTION:
    raise NotImplementedError("Multiple zlevels currently only works with focusing method MAX_INTENSITY_PROJECTION")
elif len(self._zlevels) < 1 and focus_points is not None:
    raise  NotImplementedError("If focus_points are used initial zlevels must also be provided")

```

src/odemis/acq/align/test/roi_autofocus_test.py
```diff
@@ -0,0 +1,137 @@
...
+
+        xmin = init_pos[0] - (1 - overlap) * n_tiles[0] / 2 * px_size[0] * width
+        ymin = init_pos[1] - (1 - overlap) * n_tiles[1] / 2 * px_size[1] * height
+        xmax = init_pos[0] + (1 - overlap) * n_tiles[0] / 2 * px_size[0] * width
+        ymax = init_pos[1] + (1 - overlap) * n_tiles[1] / 2 * px_size[1] * height
+
+        bbox = (xmin, ymin, xmax, ymax)
+
+        f = autofocus_in_roi(bbox, self.stage, self.ccd, self.focus, self.focus_range, n_tiles[0], n_tiles[1],
+                             0)  # TODO add proper confidence value
```
tepals, line 114: Since you are only testing that it is running I think it is fine to use a confidence_level of 0. It just means that none of the focus points will be discarded

src/odemis/acq/stitching/test/tiledacq_test.py
```diff
@@ -311,6 +311,119 @@ def test_whole_procedure(self):
...
+                        [0.0, 0.0001485167330631956, -8.124999e-06],
+                        [0.00019889999999999998, 0.0001485167330631956, 4.765627e-06]]
+        focus_range = self.focus.getMetadata().get(model.MD_POS_ACTIVE_RANGE)
+        axis_range = self.focus.axes['z'].range
+
+        # Test 1 when z level is not provided
+        tiled_acq_task = TiledAcquisitionTask(self.fm_streams, self.stage,
+                                              area=area, overlap=overlap, future=model.InstantaneousFuture(),
+                                              focus_points=focus_points)
+        # TODO need the following metadata in mimas-sim and meteor-sim
```
tepals, line 333: I think it would be nice to just add this in this PR. I'm pretty sure the TODO will get lost here :)

src/odemis/acq/stitching/test/tiledacq_test.py
tepals: This comment is not correct, sorry. We can also do refocus when the zlevels are not defined. Then it just refocuses the single point.

But I think there should also be a test case that checks with zlevels and focus points and focusing_method = FocusingMethod.MAX_INTENSITY_PROJECTION

src/odemis/acq/stitching/test/tiledacq_test.py
```diff
@@ -311,6 +311,119 @@ def test_whole_procedure(self):
...
+        tiled_acq_task = TiledAcquisitionTask(self.fm_streams, self.stage,
+                                              area=area, overlap=overlap, future=model.InstantaneousFuture(),
+                                              focus_points=focus_points)
+        # Test a point that is on the plane
+        x = 0
+        y = 0
+        z = tiled_acq_task._get_z_on_focus_plane(x, y)
+        self.assertEqual(z, tiled_acq_task._focus_plane["gamma"])
+
+        # Test a point that is not on the plane
```
tepals, line 373: why is this point not on the plane?

src/odemis/util/test/linalg_test.py
```diff
@@ -71,5 +72,54 @@ def testUnity(self):
             self.assertTrue(numpy.allclose(tri_inv(c), c))
 
 
+class PlaneFittingTestCase(unittest.TestCase):
+    def test_fit_plane_lstsq(self):
+        """
+        Test fitting a plane to a set of 3D coordinates using least-squares fitting.
+        """
+        # Create a plane with equation z = 2x + 3y + 4
```
tepals, line 80: I think you can combine this comment with the one on line 82, because this equation is not valid anymore for the second part of this test case, or is it?

tepals (changes requested):

src/odemis/acq/align/test/roi_autofocus_test.py
```diff
@@ -0,0 +1,137 @@
...
+        ymin = init_pos[1] - (1 - overlap) * n_tiles[1] / 2 * px_size[1] * height
+        xmax = init_pos[0] + (1 - overlap) * n_tiles[0] / 2 * px_size[0] * width
+        ymax = init_pos[1] + (1 - overlap) * n_tiles[1] / 2 * px_size[1] * height
+
+        bbox = (xmin, ymin, xmax, ymax)
+
+        f = autofocus_in_roi(bbox, self.stage, self.ccd, self.focus, self.focus_range, n_tiles[0], n_tiles[1], 0)
+
+        # Test if the autofocus in roi is running
+        time.sleep(2)
```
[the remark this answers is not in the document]
K4rishma: I am not sure how much time it takes. If I remove it, the test fails. Do I need to check proper time?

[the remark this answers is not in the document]
pieleric: Yes, you need a sleep() as immediately after calling it could have not yet started (as the thread needs a bit of time to start). However it should typically take less than 0.1s.

src/odemis/acq/stitching/test/tiledacq_test.py
pieleric: Just a tiny comment here: NotImplementedError should be used for arguments which are correct but the code is actually not (yet) implemented. However, if the argument (combination) is wrong, it's more typical to raise ValueError. I have a feeling here it's the second case, but I let you judge.

src/odemis/acq/stitching/test/tiledacq_test.py
K4rishma: In mimas and enzel, the metadata of POS_ACTIVE_RANGE can be added, in meteor SAFE_REL_RANGE can be added in respective microscope files. @pieleric should I change the three microscope files or leave them as it is. Irrespective of adding the metadata in the microscope file, I will make sure no error arises when the metadata is not defined.

src/odemis/acq/align/test/roi_autofocus_test.py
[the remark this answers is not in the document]
K4rishma: sleep(0.1) works

src/odemis/acq/stitching/test/tiledacq_test.py
K4rishma: @tepals in the following check 
`elif len(self._zlevels) < 1 and focus_points is not None: raise NotImplementedError("If focus_points are used initial zlevels must also be provided")`
Why is it required? As even if z levels are not provided and focus points are provided, the following piece of code in _tileacq.py takes care
`    def _get_zstack_levels(self, focus_value):
        """
        Calculate the zstack levels from the current focus position and zsteps value
        :returns (list(float) or None) zstack levels for zstack acquisition.
          return None if only one zstep is requested.
        """
        # When initial z level is one or None, use the current focus value
        if len(self._init_zlevels) <= 1:
            return [focus_value, ]`
 Following up question, in autofocus_roi.py, I do not see where does the code depends on the manual focus value before autofocus begins. Could you guide me if I need to incorporate that or is it already incorporated?

src/odemis/acq/stitching/test/tiledacq_test.py
K4rishma: deleting this test, does not work as intended. The preliminary idea was to check a 3d point outside the plane and make sure the z value from the plane calculation does not coincide with the z of 3d point.

K4rishma: added better test

src/odemis/util/test/linalg_test.py
K4rishma: The points calculated in the first and second part both lie on this plane. First part tests when points are linear, many plane equations can fit through them. Check that c is not infinity. Second part tests the plane equation through non-planar points,  which would give a unique plane equation.

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -696,19 +698,28 @@ def _get_zstack_levels(self, focus_value):
...
+        else:
+            comp_range = self._focus_stream.focuser.axes['z'].range
+
+        # The number of zlevels will remain the same, but the range will be adjusted
+        zmin = max(min(zlevels), min(comp_range))
+        zmax = min(max(zlevels), max(comp_range))
+        zlevels = numpy.linspace(zmin, zmax, len(zlevels)).tolist()
+
+        # Create focus zlevels from the given zsteps number
         zlevels = numpy.linspace(zmin, zmax, len(zlevels)).tolist()
```
pieleric, line 719: Twice the same line.

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -696,19 +698,28 @@ def _get_zstack_levels(self, focus_value):
...
-        zmax = min(max(zlevels), max(axis_range), max(focus_range))
+        # Check which focus range is available
+        if self._focus_range is not None:
+            comp_range = self._focus_range
+        else:
+            comp_range = self._focus_stream.focuser.axes['z'].range
+
+        # The number of zlevels will remain the same, but the range will be adjusted
+        zmin = max(min(zlevels), min(comp_range))
+        zmax = min(max(zlevels), max(comp_range))
```
pieleric, line 715: That's not really useful. The distance between the zlevels should be fixed. Please use the same computation as in gui.win.acquistion._get_zstack_levels().
Essentially, it keeps the number of zlevels and the distance between zlevels. (by always adjusting both zmin and zmax simultaneously). Of course, this only works if (zmax-zmin) < (comp_range[1]-comp_range[0]), but that should always be the case.

src/odemis/acq/align/test/roi_autofocus_test.py
```diff
@@ -0,0 +1,136 @@
...
+
+        f = autofocus_in_roi(bbox, self.stage, self.ccd, self.focus, self.focus_range, n_tiles[0], n_tiles[1], 0)
+
+        # Test if the autofocus in roi is running
+        time.sleep(0.1)
+        self.assertTrue(f.running())
+
+        # The returned focus points should be minimum three
+        focus_points = f.result()
+        self.assertGreaterEqual(len(focus_points), 3)
```
pieleric, line 90: What should be the format of each focus point? (x,y, z)? Please also check that. 

BTW, is there a maximum amount of focus points that should be measured? (eg 9), if so, you could check that too.

pieleric (changes requested):

src/odemis/acq/stitching/_tiledacq.py
K4rishma: @tepals, to keep distance between the z stack levels same, I will have to use the code suggested above. Is there anything I should keep in mind related to above changes?

src/odemis/acq/align/test/roi_autofocus_test.py
K4rishma: At the moment we have not defined maximum number of focus points required for autofocus. The 9 points we take initially can be reduced to a lower number if the confidence level when the focus point is below a confidence level. (conf_level = 0.8). More than 9 focus points can be included in future depending on the number of tiles. In both the cases, minimum number of focus points can be checked.  

I have added the following
`        # The returned focus points should be minimum three
        # One focus point has three coordinates x,y and z
        # The returned focus points should be minimum three to fit a plane for the autofocus`

src/odemis/acq/stitching/test/tiledacq_test.py
tepals: You are correct, so raising the error is not necessary. 

autofocus_roi.py takes the current position of the focus component as the starting position, it might be useful to clarify this in the autofocus_roi docstring.

src/odemis/acq/stitching/_tiledacq.py
tepals: there's nothing to worry about

src/odemis/util/linalg.py
```diff
@@ -209,3 +211,18 @@ def get_point_on_plane(x: float, y: float, tr: tuple) -> float:
...
 
     return z
+
+
+def are_collinear(p1: numpy.ndarray, p2: numpy.ndarray, p3: numpy.ndarray) -> bool:
+    """
+    Check if three points are collinear.
+    :param p1: x,y,z coordinates of the first point
+    :param p2: x,y,z coordinates of the second point
+    :param p3: x,y,z coordinates of the third point
```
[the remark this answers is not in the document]
K4rishma: Forgot to add the dependency on z coordinate, I have updated the code to add it

src/odemis/util/linalg.py
```diff
@@ -209,3 +212,20 @@ def get_point_on_plane(x: float, y: float, tr: tuple) -> float:
...
+    :return: True if the points are on same line, False otherwise
+    """
+    x1, y1, z1 = p1
+    x2, y2, z2 = p2
+    x3, y3, z3 = p3
+
+    return abs((x2 - x1) * (y3 - y1) - (x3 - x1) * (y2 - y1)) < 1e-12 and \
+        abs((x2 - x1) * (z3 - z1) - (x3 - x1) * (z2 - z1)) < 1e-12 and \
+        abs((y2 - y1) * (z3 - z1) - (y3 - y1) * (z2 - z1)) < 1e-12
```
tepals, line 231: can you add a comment explaining what you are checking here?

tepals (changes requested):

src/odemis/acq/align/test/roi_autofocus_test.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def test_autofocus_in_roi(self):
+        """
+        Tests the autofus in roi is running and returns the focus points:
+        """
+        init_pos = self.init_pos
+        px_size = self.px_size
+        width = self.width
+        height = self.height
+        n_tiles = (3, 3)
```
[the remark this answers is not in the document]
K4rishma: https://github.com/delmic/odemis/pull/2349/files#diff-bdc729b3559ed5ec902586ae195a9dae2c02d84f7414193d8a73f45a035afaa1 , I  have added those changes in the mentioned PR.

src/odemis/acq/align/test/roi_autofocus_test.py
```diff
@@ -0,0 +1,142 @@
...
+        time.sleep(0.1)
+        self.assertTrue(f.cancelled())
+        with self.assertRaises(futures.CancelledError):
+            f.result()
+
+    def test_estimate_autofocus_in_roi(self):
+        """
+        Tests time estimation of autofocus in roi
+        """
+        n_tiles = (3, 3)
```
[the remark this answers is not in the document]
K4rishma: I will update the test files in https://github.com/delmic/odemis/pull/2349/files#diff-bdc729b3559ed5ec902586ae195a9dae2c02d84f7414193d8a73f45a035afaa1

src/odemis/acq/stitching/test/tiledacq_test.py
```diff
@@ -311,6 +312,136 @@ def test_whole_procedure(self):
...
+                        [-0.0001989, 0.0001485, -1.781e-05],
+                        [0.0, 0.0001485, -8.125e-06],
+                        [0.0001989, 0.0001485, 4.766e-06]]
+        zlevels = [focus_points[0][2], focus_points[1][2], focus_points[2][2]]
+        focus_range = self.focus.getMetadata().get(model.MD_POS_ACTIVE_RANGE, None)
+        axis_range = self.focus.axes['z'].range
+        if focus_range is None:
+            comp_range = axis_range
+        else:
+            comp_range = focus_range
```
tepals, line 334: wouldn't the following also work?
```python
        axis_range = self.focus.axes['z'].range
        comp_range = self.focus.getMetadata().get(model.MD_POS_ACTIVE_RANGE, axis_range)
```

src/odemis/acq/stitching/test/tiledacq_test.py
```diff
@@ -311,6 +312,136 @@ def test_whole_procedure(self):
...
+                                              area=area, overlap=overlap, future=model.InstantaneousFuture(),
+                                              zlevels=zlevels, focus_points=focus_points,
+                                              focusing_method=FocusingMethod.MAX_INTENSITY_PROJECTION)
+        tiled_acq_task._refocus()
+        # min with none values
+        zmin = min(comp_range)
+        zmax = max(comp_range)
+        # Test the min and max range of the zstack
+        self.assertGreaterEqual(tiled_acq_task._zlevels[0], zmin)
+        self.assertLessEqual(tiled_acq_task._zlevels[-1], zmax)
```
tepals, line 346: shouldn't this be equal? not GreaterEqual/LessEqual

tepals (changes requested):

src/odemis/acq/stitching/test/tiledacq_test.py
K4rishma: I will add that code.

PR #2343, opened 2023-02-20
add functional chamber tab buttons

FCanberk:
Add 4 buttons(namely: LOADING, COATING, OPTICAL and FIB) to the Chamber tab and make the button clicks work, including the color changes on button click.

Additionally hide all the tab-dependent buttons initially and make them appear only in the correct tab.


pieleric (changes requested):
The changes look fine. However there are many files that come from a separate PR. Please only include the changes in gui/cont/tabs.py and gui/main.xrc(.py).

BTW, did you test on the METEOR and the ENZEL? The chamber tab needs to still work as before.

FCanberk:
> The changes look fine. However there are many files that come from a separate PR. Please only include the changes in gui/cont/tabs.py and gui/main.xrc(.py).
> 
> BTW, did you test on the METEOR and the ENZEL? The chamber tab needs to still work as before.


- I removed the extra files from the separate PR.
- METEOR and ENZEL are already tested and works as before.

pieleric (approved):

PR #2344, opened 2023-02-21
fastem: update beamshift correction to use estimate_grid_orientation_…

tepals:
…from_img instead of FindGridSpots

### DIFF
--- a/src/odemis/acq/fastem.py
+++ b/src/odemis/acq/fastem.py
@@ -508,7 +508,7 @@ class AcquisitionTask(object):
         self._pre_calibrations = pre_calibrations
         self._settings_obs = settings_obs
 
-        self.path = fastem_util.create_image_dir("beam_shift_cor")
+        self.path = fastem_util.create_image_dir("beam-shift-correction")
 
         # Dictionary containing the single field images with index as key: e.g. {(0,1): DataArray}.
         self.megafield = {}
@@ -799,7 +799,8 @@ class AcquisitionTask(object):
         # Determine the shift of the spots, by subtracting the good multiprobe position from the average (center)
         # spot position.
         fav_pos_active = self._ccd.getMetadata()[model.MD_FAV_POS_ACTIVE]
-        # If i and j are not in the metadata, use x and y instead.
+        # FIXME: If i and j are not in the metadata, use x and y instead.
+        #  Old yaml files used x and y, support for x and y can be removed when all yaml files are updated.
         try:
             i = fav_pos_active["i"]
         except KeyError:
@@ -819,7 +820,7 @@ class AcquisitionTask(object):
         shift_m *= -1
         beam_shift_cor = shift_m / magnification  # [m]
         # Convert the shift from pixels to meters
-        logging.debug("Beam shift adjustment required due to stage magnetic field: {} [m]".format(beam_shift_cor))
+        logging.debug("Beam shift adjustment required: {} [m]".format(beam_shift_cor))
 
         cur_beam_shift_pos = numpy.array(self._beamshift.shift.value)
         logging.debug("Current beam shift: {} [m]".format(self._beamshift.shift.value))
@@ -828,7 +829,7 @@ class AcquisitionTask(object):
         logging.debug("New beam shift m: {}".format(self._beamshift.shift.value))
 
         ccd_image = self._ccd.data.get(asap=False)
-        fastem_util.save_image(self.path, f"{self.field_idx}_after", ccd_image)
+        fastem_util.save_image(self.path, f"{self.field_idx}_after.tiff", ccd_image)
 
     def _create_acquisition_metadata(self):
         """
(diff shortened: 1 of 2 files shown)


src/odemis/acq/fastem.py
```diff
@@ -783,31 +785,33 @@ def correct_beam_shift(self):
...
         centered on the mppc detector.
         """
+        pixel_size = self._ccd.pixelSize.value
+        magnification = self._lens.magnification.value
+        refractive_index = self._lens.refractiveIndex.value
+        numerical_aperture = self._lens.numericalAperture.value
+        wavelength = 550e-9
+        sigma = (magnification / pixel_size) * synthetic.psf_sigma_wffm(
+            refractive_index, numerical_aperture, wavelength
+        )
```
tepals, line 795: @effting is this the correct way of calculating sigma?

effting: It should be better to use `ccd.pointSpreadFunctionSize` from `src/odemis/model/_components.py`

effting: Could also be `self._ccd.spotSize`. Need to check which one to use

src/odemis/acq/fastem.py
```diff
@@ -783,31 +785,33 @@ def correct_beam_shift(self):
...
         # asap=False: wait until new image is acquired (don't read from buffer)
         ccd_image = self._ccd.data.get(asap=False)
-
-        # Find the location of the spots on the diagnostic camera.
-        spot_coordinates, *_ = FindGridSpots(ccd_image, (8, 8))
-
-        # Transform the spots from the diagnostic camera coordinate system to a right-handed coordinate
-        # system with the origin in the bottom left.
-        spot_coordinates[:, 1] = ccd_image.shape[1] - spot_coordinates[:, 1]  # [px]
+        tform = estimate_grid_orientation_from_img(ccd_image, (8, 8), AffineTransform, sigma)
```
tepals, line 798: @effting we were using an `AffineTransform` in `FindGridSpots`, I see that in your code you use a `SimilarityTransform`. Did you try both for the tests with missing spots, and was one performing "better"?

src/odemis/acq/fastem.py
effting: Indeed, should be `self._ccd.pointSpreadFunctionsize.value`

src/odemis/acq/fastem.py
effting: The main goal of the beamshift correction is to control (limit) cross-talk. We need to project the image of the multi-probe pattern on the MPPC sensor array. We can only control the scale, rotation, and translation so it makes sense to me to not fit the anisotropic scaling and shear here.

src/odemis/acq/fastem.py
```diff
@@ -506,6 +508,8 @@ def __init__(self, scanner, multibeam, descanner, detector, stage, ccd, beamshif
         self._pre_calibrations = pre_calibrations
         self._settings_obs = settings_obs
 
+        self.path = fastem_util.create_image_dir("beam_shift_cor")
```
nandishjpatel, line 511:
```suggestion
        self.path = fastem_util.create_image_dir("beam-shift-correction")
```

nandishjpatel (approved):

src/odemis/acq/fastem.py
```diff
@@ -783,34 +787,49 @@ def correct_beam_shift(self):
...
+        fav_pos_active = self._ccd.getMetadata()[model.MD_FAV_POS_ACTIVE]
+        # If i and j are not in the metadata, use x and y instead.
+        try:
+            i = fav_pos_active["i"]
+        except KeyError:
+            i = fav_pos_active["x"]
+        try:
+            j = fav_pos_active["j"]
+        except KeyError:
+            j = ccd_image.shape[1] - fav_pos_active["y"]
```
pieleric, line 810: Why wouldn't it be there, while x/y would be there? Old version of the calibration function? If so, please add this info in the comment, and indicate the x/y code can be dropped soon.

pieleric (changes requested):

src/odemis/acq/fastem.py
```diff
@@ -783,34 +787,49 @@ def correct_beam_shift(self):
...
+        #  on the diagnostic camera. Therefore we need to invert the shift.
+        #  When the beamshift component is fixed, this should be removed.
+        shift_m *= -1
+        beam_shift_cor = shift_m / magnification  # [m]
         # Convert the shift from pixels to meters
-        pixel_size = self._ccd.pixelSize.value
-        magnification = self._lens.magnification.value
-        shift_m = shift * pixel_size / magnification  # [m] pixel size diagnostic camera divided by 40x magnification
-        logging.debug("Beam shift adjustment required due to stage magnetic field: {} [m]".format(shift_m))
+        logging.debug("Beam shift adjustment required due to stage magnetic field: {} [m]".format(beam_shift_cor))
```
effting, line 822: You can drop the "due to stage magnetic field"

effting (approved):

nandishjpatel (approved):

PR #2345, opened 2023-02-21
acq orsay_milling: detect better end of milling

pieleric:
MillingActivationState sometimes goes at init to 1, then 0, and then
back to 1. Finally, at the end it becomes 0 again.
That meant that sometimes the detection of the end would be triggered
just before the actual acquisition.
=> Use .Target (as recommanded by Orsay Physics) and CurrentMillingPoint
to be really certain the milling is stated & finished.

Also generalise the parameter logger.

### DIFF
--- a/src/odemis/acq/orsay_milling.py
+++ b/src/odemis/acq/orsay_milling.py
@@ -122,7 +122,6 @@ class OrsayMilling:
         """
         Function that handles milling action
         """
-        # for now the ip address is hardcoded, it will be replaced with scanner.parent.host
         rect = self.rect
         iteration = self.iteration
         duration = self.duration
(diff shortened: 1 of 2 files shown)


src/odemis/acq/orsay_milling.py
```diff
@@ -181,37 +184,77 @@ def do_milling(self):
...
-        # make sure the Unsubscribe is always called even if any exception happens
+            if self.future.cancelled():
+                raise futures.CancelledError()
         finally:
-            miller.MillingActivationState.Unsubscribe(self.on_milling_state)
+            # make sure the Unsubscribe is always called even if any exception happens
+            miller.MillingActivationState.Unsubscribe(self.param_logger)
+            scanner.MillingStatus.Unsubscribe(self.param_logger)
+            # scanner.CurrentSubList.Unsubscribe(self.param_logger)
+            # scanner.CurrentMillingPoint.Unsubscribe(self.param_logger)
```
FCanberk, line 256: Do we need to keep these commented out lines?

pieleric: Yes, because at the top I also left the Subscribe commented out. It's for hard-core debugging. It's very verbose, but can still be handy if we notice the milling process doesn't work properly (yet).

PR #2351, opened 2023-03-01
fix for yaml.representer.RepresenterError raised exception when an ob…

nandishjpatel:
In the `self._persistent_data` dict for FAST-EM, ['Mirror Descanner']['properties']['scanOffset'] is a tuple containing numpy.float64 elements.

This raises a yaml exception while dumping the dictionary. A work-around is inheriting yaml `SafeRepresenter` class and handling `numpy.floating` datatype by converting into float. Finally, `CustomSafeDumper` makes use of `CustomSafeRepresenter` and is used as the dumper during `yaml.dump()`.

### DIFF
--- a/src/odemis/odemisd/main.py
+++ b/src/odemis/odemisd/main.py
@@ -23,24 +23,32 @@ Odemis. If not, see http://www.gnu.org/licenses/.
 
 import argparse
 import grp
-from logging import FileHandler
 import logging
-from logging.handlers import WatchedFileHandler
-from odemis import model
-import odemis
-from odemis.model import ST_UNLOADED, ST_STARTING
-from odemis.odemisd import modelgen
-from odemis.odemisd.mdupdater import MetadataUpdater
-from odemis.util.driver import BACKEND_RUNNING, BACKEND_DEAD, BACKEND_STOPPED, \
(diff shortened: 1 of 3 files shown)


src/odemis/odemisd/main.py
```diff
@@ -50,6 +58,36 @@
                     BACKEND_STARTING: 3,
                     }
 
+
+class CustomSafeRepresenter(SafeRepresenter):
+
+    def represent_numpy_float(self, data):
+        data = float(data)
+        self.represent_float(data)
```
tepals, line 66: can you add a comment to explain that this is because yaml does not support dumping numpy floats

tepals (changes requested):

pieleric (changes requested):
This class would be a lot less lonely if it had a test case for it ;-)
See util.test.conversion_test.test_json_numpy() for inspiration.

nandishjpatel:
> This class would be a lot less lonely if it had a test case for it ;-) See util.test.conversion_test.test_json_numpy() for inspiration.

Haha, sure I can write a testcase :)

pieleric (approved):

tepals (approved):

PR #2352, opened 2023-03-02
[update] fastem: blank the beam when the stage moves during acquisitions

tepals:


tepals:
see SDEV-1163

src/odemis/acq/fastem.py
```diff
@@ -658,6 +659,7 @@ def acquire_roa(self, dataflow):
             # Update the time left for the acquisition.
             expected_time = len(self._fields_remaining) * total_field_time
             self._future.set_progress(start=time.time(), end=time.time() + expected_time)
+            self._scanner.blanker.value = True  # blank the beam
```
nandishjpatel, line 662: Line 662 can go to the line above 639?

tepals: good point, that also works

pieleric: If use use the blanker, do it fully, and only unblank while acquiring. So I'd say move this line just after line 650.

pieleric (approved):

nandishjpatel (approved):

PR #2354, opened 2023-03-06
GeometricTransform __matmul__ method

effting:
Alternative implementation to allow easy modification of GeometricTransform instances by daisy-chaining.

Specifically adds a `__matmul__` method on `GeometricTransform` that allows to do for example:
```
tform = tform_a @ tform_b @ tform_a.inverse()
```


src/odemis/util/transform.py
```diff
@@ -738,6 +739,15 @@ def __init__(
             numpy.zeros(2, dtype=float) if translation is None else translation
         )
 
+    def __matmul__(self, other):
```
pieleric, line 742: No typing info? other should be a GeometricTransform, and it'll return also a GeometricTransform, right?

src/odemis/util/transform.py
```diff
@@ -738,6 +739,15 @@ def __init__(
...
+        """Overloads the `@` operator to combine two transforms into one."""
```
pieleric, line 743: It'd be nice to clarify the order in the documentation. Something like:
"(A @ B).apply(p) is equivalent to A.apply(B.apply(p))"

pieleric (changes requested):

tepals (approved):
looks good, can you fix the merge conflict?

pieleric (approved):

effting:
Thanks. I've resolved the merge conflict, so should be go

PR #2355, opened 2023-03-09
gui comp: add SampleBackgroundOverlay for MIMAS

pieleric:
Shows the (theoretical) position of the samples based on the YAML
metadata into the view, as blue wireframe.

It could almost work for METEOR, but the metadata is in the SEM stage
(stage-bare) referential, while we need to have it in the FM referential
(stage).

### DIFF
--- a/src/odemis/gui/comp/overlay/world.py
+++ b/src/odemis/gui/comp/overlay/world.py
@@ -376,7 +376,12 @@ class SampleBackgroundOverlay(WorldOverlay):
                                                 )
 
     def draw(self, ctx, shift=(0, 0), scale=1.0):
-        """ Draw the background image by displaying all circles. """
+        """
+        Draw the background image by displaying all circles
+        :param ctx: cairo context from the canvas
+        :param shift (float, float): physical coordinates of the center of the canvas buffer
+        :param scale (float > 0): the ratio between the size of a feature in pixels and its actual size, in px/m 
+        """
         if not self.show:
             return
 
--- a/src/odemis/gui/comp/overlay/world.py
+++ b/src/odemis/gui/comp/overlay/world.py
@@ -25,27 +25,31 @@ This file is part of Odemis.
 import logging
 import math
 from abc import ABCMeta, abstractmethod
+from typing import Dict, Tuple
 
 import cairo
-import wx
-from future.utils import with_metaclass
-
 import odemis.gui as gui
 import odemis.gui.img as guiimg
 import odemis.util.conversion as conversion
 import odemis.util.units as units
+import wx
+from future.utils import with_metaclass
 from odemis import model, util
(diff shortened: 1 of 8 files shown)


src/odemis/gui/cont/tabs.py
```diff
@@ -382,6 +381,11 @@ def __init__(self, name, button, panel, main_frame, main_data):
         panel.vp_secom_tl.canvas.remove_view_overlay(panel.vp_secom_tl.canvas.play_overlay)
         panel.vp_secom_tr.canvas.remove_view_overlay(panel.vp_secom_tr.canvas.play_overlay)
 
+        # Add a sample overlay to each view (if we have information)
+        if main_data.sample_centers:
```
nandishjpatel, line 385:
```suggestion
        if main_data.sample_centers and main_data.sample_radius:
```
maybe a better check

src/odemis/gui/comp/viewport.py
```diff
@@ -650,6 +652,27 @@ def _set_fov_from_mpp(self):
...
+        the arguments have no effect.
+        sample: Dict of sample name -> center position (X/Y in m)
+        radius: the radius of a sample in m (all samples are shown with the same radius)
+        """
+        if not self._sample_overlay:
+            self._sample_overlay = overlay.world.SampleBackgroundOverlay(self.canvas, samples, radius)
+        self.canvas.add_world_overlay(self._sample_overlay)
+        wx.CallAfter(self.canvas.request_drawing_update)
+
+    def hide_sample_overlay(self):
```
nandishjpatel, line 668:
```suggestion
    def remove_sample_overlay(self):
```
a better naming?

src/odemis/gui/comp/overlay/world.py
```diff
@@ -347,6 +351,67 @@ def set_icon(feature_icon):
...
+        self._labels = {}
+        for name in self._samples:
+            self._labels[name] = self.add_label(name,
+                                                align = wx.ALIGN_CENTRE_HORIZONTAL | wx.ALIGN_CENTER_VERTICAL,
+                                                colour=self._colour[:3],  # no transparency, to support .opacity
+                                                background=(0, 0, 0, 0.5)  # Black semi-transparent
+                                                )
+
+    def draw(self, ctx, shift=(0, 0), scale=1.0):
+        """ Draw the background image by displaying all circles. """
```
FCanberk, line 379: Would it be good to add parameter explanations in the doc string or do you think it is already self-explanatory?

nandishjpatel (approved):

src/odemis/gui/cont/tabs.py
```diff
@@ -339,7 +339,6 @@ def get_display_priority(cls, main_data):
 AUTOFOCUS_BINNING = (8, 8)
 AUTOFOCUS_HFW = 300e-06  # m
 
-
 class LocalizationTab(Tab):
```
FCanberk, line 342: It's a bit of nitpicking but I think we should keep the line break before the class definition as PEP-8 states: https://peps.python.org/pep-0008/#blank-lines

PR #2356, opened 2023-03-10
Added some small changes to the testcases for Zeiss driver for 6-axes movement

sneepstefan:
Added some small changes to the test cases for the 6-axes movement testing of the stage. Notably adding a starting position to return to later and changed moves to under- and overflow moves for Rm. Also renamed axis Rz to Rm and added a small correction in relative move regarding under- and overflow.

### DIFF
--- a/src/odemis/driver/zeiss.py
+++ b/src/odemis/driver/zeiss.py
@@ -487,8 +487,8 @@ class Stage(model.Actuator):
         # To keep compatibility with old configuration files, by default the extra axes are not used.
         if "rx" not in rng:
             rng["rx"] = None
-        if "rz" not in rng:
-            rng["rz"] = None
+        if "rm" not in rng:
+            rng["rm"] = None
         if "m" not in rng:  # Extra Z axis, in the Rx referential, to adjust eucentric height.
             rng["m"] = None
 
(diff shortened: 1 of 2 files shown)


nandishjpatel (approved):

K4rishma (approved):

src/odemis/driver/zeiss.py
```diff
@@ -598,9 +598,11 @@ def _doMoveRel(self, future, shift):
...
             move_list[3] += shift["rx"]
             target_pos["rx"] = move_list[3]
-        if "rz" in shift:
-            move_list[4] += shift["rz"]
-            target_pos["rz"] = move_list[4]
+        if "rm" in shift:
+            move_list[4] += shift["rm"]
+            move_list[4] = move_list[4] % (2 * math.pi)
+            # correct for full rotation under and overflow
+            target_pos["rm"] = move_list[4] % (2 * math.pi)
```
K4rishma, line 605: Though the result is same, just easy to read
```suggestion
            target_pos["rm"] = move_list[4] 
```

sneepstefan: Yeah good point that actually a remainder of quick programming where the second use of (2 * math.pi) is redundant.
Changed it

PR #2357, opened 2023-03-10
[Draft] Meteor stage test cases for FLM-SEM switching

K4rishma:
This branch is to check for new changes, Please do not merge.


src/odemis/model/_metadata.py
```diff
@@ -201,6 +201,7 @@
 MD_FM_IMAGING_RANGE = "FM imaging range"  # dict str → [float, float] defining the volume of the FM imaging area, along x, y and z axes.
 MD_FAV_FM_POS_ACTIVE = "Favourite FM position active"  # dict str->float representing the position required for FM imaging
 MD_FAV_SEM_POS_ACTIVE = "Favourite SEM position active"  # dict -> float representing the position required for SEM imaging
+MD_CAL_VAL = "Calibration values for stage switching" # dict str -> float representing the calibration values to find axes coordinates during stage switching
```
pieleric, line 204: We have a MD_CALIB already. It's for this sort of "arbitrary pot-luck values for calibration". So let's use that one instead of introducing a new metadata.

PR #2358, opened 2023-03-10
util comp: fix use of MD_POS_ACTIVE_RANGE in generate_zlevels()

pieleric:
MD_POS_ACTIVE_RANGE is a dict of the format axis -> tuple. However, the
code assumed that it would just be a tuple.

=> Look explicitly for the "z" axis.

### DIFF
--- a/src/odemis/util/comp.py
+++ b/src/odemis/util/comp.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-'''
+"""
 Created on 31 May 2018
 
 @author: Éric Piel
@@ -13,7 +13,7 @@ Odemis is free software: you can redistribute it and/or modify it under the term
 Odemis is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
 
 You should have received a copy of the GNU General Public License along with Odemis. If not, see http://www.gnu.org/licenses/.
-'''
+"""
 import copy
 import math
 from typing import List, Tuple
@@ -115,6 +115,7 @@ def get_fov_rect(comp, fov):
 
 MAX_ZLEVELS = 500
 
+
 def generate_zlevels(focuser: "Actuator", zrange: Tuple[float, float], zstep: float) -> List[float]:
     """
     Calculates the zlevels for a zstack acquisition, using the zmax, zmin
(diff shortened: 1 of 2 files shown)


tepals (approved):

pieleric:
@tepals , I've extended the PR with a test case... and then found more issues in the code. So it now a lot bigger than the original 2 lines change... Could you re-review?

src/odemis/util/comp.py
```diff
@@ -113,64 +115,67 @@ def get_fov_rect(comp, fov):
...
-    if focuser_pos + zrange[0] < focuser_rng[0]:
-        zrange[0] = focuser_rng[0] - focuser_pos
-    if zrange[0] == zrange[1]:
+    if sw_rng and "z" in sw_rng:
+        focuser_rng = (max(focuser_rng[0], sw_rng["z"][0]),
+                       min(focuser_rng[1], sw_rng["z"][1]))
+
+    # clip the zMax and zMin to the actuator limits if necessary
+    zrange_abs = (min(max(focuser_rng[0], zrange_abs[0]), focuser_rng[1]),
+                  min(max(focuser_rng[0], zrange_abs[1]), focuser_rng[1]))
```
tepals, line 154: This always breaks my head a bit, but shouldn't it be:
```suggestion
    zrange_abs = (min(max(focuser_rng[0], zrange_abs[0]), focuser_rng[1]),
                  max(min(focuser_rng[1], zrange_abs[1]), focuser_rng[0]))
```

src/odemis/util/comp.py
```diff
@@ -113,64 +115,67 @@ def get_fov_rect(comp, fov):
...
+
+    if zrange_abs[0] == zrange_abs[1]:
         return focuser_pos + zrange[0]
 
     # find number of samples
```
tepals, line 159: this is number of zlevels right instead of number of samples? Could you clean this up as well?

src/odemis/util/comp.py
```diff
@@ -113,64 +115,67 @@ def get_fov_rect(comp, fov):
...
-    n = (zrange[1] - zrange[0]) / abs(zstep) + 1
+    n = (zrange_abs[1] - zrange_abs[0]) / abs(zstep) + 1
     if n > MAX_ZLEVELS:
-        raise IndexError("The number of zlevels, %s, is too large. The max allowable number of zlevels is %s. Reduce the zstep value." %(n, MAX_ZLEVELS))
+        raise IndexError(f"The number of zlevels, {n}, is too large. Reduce the zstep value to < {MAX_ZLEVELS}.")
+
     # try the floor and ceil values for the number of samples, and 
     # take the one that give smaller error.
     # note: if n is 1 -> division by zero and step_c will always be picked,
     # so just choose any large value for step_f (e.g infinite)
```
tepals, line 167: this comment is incorrect, if n is exactly 1 there will also be a division by zero error for the step_c, which is then not handled at all.

tepals: you could also write it like this:
```python
    n_steps_f = math.floor(n) - 1
    n_steps_c = math.ceil(n) - 1

    if n_steps_f == 0:
        n = math.ceil(n)
    else:
        # errors
        ef = abs((zrange_abs[1] - zrange_abs[0]) / n_steps_f - abs(zstep))
        ec = abs((zrange_abs[1] - zrange_abs[0]) / n_steps_c - abs(zstep))
        n = math.floor(n) if ef < ec else math.ceil(n)
```

src/odemis/util/test/comp_test.py
```diff
@@ -235,6 +235,18 @@ def test_clipping_zmax_on_actuator_upper_limit(self):
         actual = generate_zlevels(self.focus, zrange, zStep)
         numpy.testing.assert_array_almost_equal(expected, actual)
 
+    def test_clipping_axis_range(self):
+        self.focus.updateMetadata({model.MD_POS_ACTIVE_RANGE: {"z": [-100e-6, 100e-6]}})
+        try:
+            self.focus.moveAbsSync({"z": 80e-6})
+            zrange = (-50e-6, 50e-6)
+            zStep = 10e-6
+            expected = numpy.arange(30e-6, 100e-6 + 1e-12, 10e-6)
```
tepals, line 244:
```suggestion
            expected = numpy.arange(30e-6, 100e-6 + 1e-12, zStep)
```

src/odemis/util/test/comp_test.py
```diff
@@ -235,6 +235,18 @@ def test_clipping_zmax_on_actuator_upper_limit(self):
         actual = generate_zlevels(self.focus, zrange, zStep)
         numpy.testing.assert_array_almost_equal(expected, actual)
 
+    def test_clipping_axis_range(self):
+        self.focus.updateMetadata({model.MD_POS_ACTIVE_RANGE: {"z": [-100e-6, 100e-6]}})
+        try:
+            self.focus.moveAbsSync({"z": 80e-6})
+            zrange = (-50e-6, 50e-6)
+            zStep = 10e-6
```
tepals, line 243: I would suggest to keep the variable name consistent with `zrange`
```suggestion
            zstep = 10e-6
```

src/odemis/util/comp.py
```diff
@@ -15,8 +15,10 @@
 You should have received a copy of the GNU General Public License along with Odemis. If not, see http://www.gnu.org/licenses/.
 '''
```
tepals, line 16: could you update this to triple double quotes, while your cleaning up the document anyway?

tepals (changes requested):

src/odemis/util/comp.py
pieleric: No. Here, the two lines do clipping of the zrange value between the extrema of focuser_rng. So both lines should be identical, except of the zrange index.
Clipping is done as `min(max(low, x), high)`.
Maybe instead of open-coding it every time, we should add it to odemis.util.clip (we already have a version for the VAs, and numpy has its own version, but we don't have a generic one).

src/odemis/util/comp.py
pieleric: I think that n can never be exactly 1 because it's defined as `(zrange_abs[1] - zrange_abs[0]) / abs(zstep) + 1` and we already handle the case of zrange_abs being null. 
Anyway, I think your version is clearer. I've adjusted it a tiny bit more.... and clarified the comments.

tepals (approved):

K4rishma (approved):

PR #2360, opened 2023-03-13
driver orsay: hide ion-beam.shift complexity

pieleric:
The actual beam shift has to
be adjusted for each probe current, so that the center of the FoV
corresponds to the same point. Also, every time the probe current is
changed, it's reset. We don't want to pass this trouble to the higher
level code. So the "base" shift for each probe current is stored (and
shared via MD_SCAN_OFFSET_CALIB), and it's hidden in the ion-beam.shift, so that a
shift of N µm always stays, even if the current changes.

The fib_beam.shift still allows to observe/affect the low-level value of
the shift.

### DIFF
--- a/src/odemis/driver/orsay.py
+++ b/src/odemis/driver/orsay.py
@@ -2981,8 +2981,6 @@ class Scanner(model.Emitter):
         self.scale.subscribe(self._updatePixelSize)
         self._fib_beam.horizontalFoV.subscribe(self._updatePixelSize, init=True)
 
-        self.shift = self._fib_beam.shift
-
         # It's not possible to change the probeCurrent directly. Instead, we
         # rely on presets. They are created by the service engineer on the Orsay
         # server, with specific names <current>pA_... . Each preset with such
@@ -3039,11 +3037,37 @@ class Scanner(model.Emitter):
         self.probeCurrent = model.VAEnumerated(current, unit="A",
                                                choices=self._probe_current_presets.keys() | {None},
                                                setter=self._setProbeCurrent)
+        self.probeCurrent.subscribe(self._onProbeCurrent, init=True)
 
         # TODO: it seems that the standard client store the latest preset loaded in
         # datamodel.IonColumnMCS.ConditionName. Setting .Target = PresetName
         # seems to not bother anything else. So maybe we should do that, and rely
         # on it to guess the preset in use (and so probe current).
+
+        # The .shift represents fib_beam.shift but with the value being
+        # independent from the probe current. Changing the probe current in the
+        # Orsay API automatically resets the beam to the preset values (which
+        # are calibrated so that the center of the FoV corresponds to the same
+        # point). We hide this confusing behaviour by only storing the shift
+        # which is extra to the preset shift (aka "base" shift). Every time the
+        # probe current is changed, the fib_beam.shift is adjusted to keep the
+        # extra shift correct. That means the user can set a shift of (x, y) µm,
+        # and it will always stay (x, y) µm, even if the current changes. The
+        # "base" shifts for each probe current are read from the presets and
+        # stored in MD_SCAN_OFFSET_CALIB. Note: if the Orsay beam shift is
+        # modified on the Orsay side, this shift is not updated (while
+        # fib_beam.shift is updated). Making it automatically update would be a
+        # lot more work/complexity to make sure that we 100% are certain it's
+        # due to a manual change and not just reset after changing probe
+        # current. Also, the probeCurrent also doesn't update automatically, so
+        # it's the same behaviour.
+        base_shifts = {c: self._get_base_shift(pn) for c, pn in self._probe_current_presets.items()}
+        self._metadata[model.MD_SCAN_OFFSET_CALIB] = base_shifts
+        self.shift = model.TupleContinuous((0.0, 0.0), unit="m",
+                                           cls=(int, float),
+                                           range=self._fib_beam.shift.range,
+                                           setter=self._set_shift)
+        self._set_shift(self.shift.value)  # ensure it's up-to-date
 
     def _get_all_probe_current_presets(self) -> Dict[float, str]:
         """
@@ -3098,10 +3122,6 @@ class Scanner(model.Emitter):
             except Exception:
                 logging.exception("Failed to retrieve preset %s", preset_name)
                 raise
-            # For now the ConsoleClient hides errors (eg, preset not existing), but
-            # it returns None, which is a good hint something went wrong.
-            if preset is None:
-                raise LookupError(f"Failed to retrieve preset {preset_name}")
 
             # Check every required parameter
             for device_name, param_name in required_parameters:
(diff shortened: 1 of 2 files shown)


src/odemis/driver/orsay.py
```diff
@@ -3039,12 +3037,34 @@ def __init__(self, name, role, parent, **kwargs):
...
         # on it to guess the preset in use (and so probe current).
 
+        # The .shift represents fib_beam.shift but with the value being independent
+        # from the probe current. The actual fib_beam.shift has to
+        # be adjusted for each probe current, so that the center of the FoV
+        # corresponds to the same point. Also, every time the probe current is
+        # changed, it's reset. We don't want to pass this trouble to the higher
+        # level code. So the "base" shift for each probe current is stored (and
+        # shared via MD_SCAN_OFFSET_CALIB), and it's hidden here, so that a
+        # shift of N µm always stays, even if the current changes.
```
tepals, line 3054: what do you mean with "a shift of N µm always stays"?

tepals (commented):
I don't understand what you mean with "a shift of N µm always stays", which is making it hard to review :), could you explain?

src/odemis/driver/orsay.py
pieleric: I mean that if the user change .shift to a value (x, y) (that I called `N`), and then changes other settings on the FIB, like the current, the shift value will stay (x, y). Maybe that sounds obvious, but so far, the behaviour was that the shift would be reset after changing the current... to some arbitrary values which correspond to "no shift".

pieleric: here is my attempt to make it clearer:
```
        # The .shift represents fib_beam.shift but with the value being
        # independent from the probe current. The actual fib_beam.shift has to
        # be adjusted for each probe current, so that the center of the FoV
        # corresponds to the same point. Changing the probe current in the Orsay
        # API, automatically reset the beam to the preset values (which should
        # correspond to a shift of 0, 0). We hide this confusing behaviour by
        # only storing the shift which is extra to the preset shift (aka "base"
        # shift). Every time the probe current is changed, the fib_beam.shift is
        # adjusted to keep the extra shift correct. That means the user can set
        # a shift of (x, y) µm, and it will always stay, even if the current
        # changes. The "base" shifts for each probe current are read from the
        # presets and stored in MD_SCAN_OFFSET_CALIB. Note: if the Orsay beam
        # shift is modified on the Orsay side, this shift is not updated (while
        # fib_beam.shift is updated). Making it automatically update would be a
        # lot more work/complexity to make sure that we 100% are certain it's
        # due to a manual change and not just reset after changing probe
        # current. Also, the probeCurrent also doesn't update automatically, so
        # it's the same behaviour.
 ```

Comment (tepals) [reply src/odemis/driver/orsay.py:3054]: Thanks I understand it now. Maybe you could make it even more explicit:
```
        That means the user can set a shift of (x, y) µm,
        and it will always stay (x, y) µm, even if the current changes.
```

File: src/odemis/driver/orsay.py [hunk 2]
```diff
@@ -3141,8 +3161,76 @@ def _setProbeCurrent(self, current: Optional[float]) -> float:
...
+        shift_clipped = act_shift[0] - base_shift[0], act_shift[1] - base_shift[1]
+        if shift != shift_clipped:
+            logging.info("Requested shift %s clipped to %s", shift, shift_clipped)
+        return shift_clipped
+
+    def _get_base_shift(self, preset_name: str) -> Tuple[float, float]:
+        """
+        Find the default beam shift for a given preset.
+        Note that this value is allowed to change over time (ie, the user can update
+        the preset), so the preset always read again.
```
tepals, line 3204:
```suggestion
        Note that this value is allowed to change over time (ie, the user can update
        the preset), so the preset is always read again.
```

tepals (changes requested):

src/odemis/driver/orsay.py
```diff
@@ -3141,8 +3161,72 @@ def _setProbeCurrent(self, current: Optional[float]) -> float:
...
+
+        # Add the base shift, and make sure it still fits the range
+        act_shift = base_shift[0] + shift[0], base_shift[1] + shift[1]
+        act_shift = self._fib_beam.shift.clip(act_shift)
+        self._fib_beam.shift.value = act_shift
+
+        shift_clipped = act_shift[0] - base_shift[0], act_shift[1] - base_shift[1]
+        if shift != shift_clipped:
+            logging.info("Requested shift %s clipped to %s", shift, shift_clipped)
+        return shift_clipped
```
K4rishma, line 3198: Will not the clipped shift have an an impact on drift correction? In such a case where we have clipped shift, when it compensates for drift, it will not able to compensate because it is at the boundary limit. In that case, we maybe can stop the milling and ask the user to resume milling after the shift has been reset by Odemis.

src/odemis/driver/test/orsay_test.py
```diff
@@ -1930,11 +1954,47 @@ def test_shift(self):
...
+        self.scanner.horizontalFoV.value = 100e-6  # m, a little large FoV, for the large shifts to work
+
+        base_shifts = self.scanner.getMetadata()[model.MD_SCAN_OFFSET_CALIB]
+        logging.info("MD_SCAN_OFFSET_CALIB = %s", base_shifts)
+        # There is one more choice in the currents that the base shift, which is "None"
+        self.assertEqual(set(base_shifts.keys()), set(self.scanner.probeCurrent.choices) - {None})
+
+        orig_shift = self.scanner.shift.value
+
+        # Changing current doesn't change the shift (but it may do that on fib_beam)
```
K4rishma, line 1975: It is bit unclear for me, does the comment wants to say the following - when the current change of the beam is requested -> only the current changes and the shift of the fib beam remains unchanged?

src/odemis/driver/orsay.py
pieleric: Indeed, that's quite true in general with drift correction, if we reach the limit of the shift we need to "gracefully" handle it. For now, on the SPARC, we generate a warning, and continue hoping for the best (which usually is not happening). For milling that could be even worse (because you can re-acquire after a failed acquisition, but you cannot undo a blurring hole). So we probably should detect that the shift reached its limit, and stop the milling of the given pattern (an alternative being to compensate with stage move, but that's for the future). 

I'd suggest to do the following two things in the milling:
* at the beginning of a new feature (aka site), reset the shift to 0,0
* Every time the shift is applied, check it's within bounds, and accepted. If not, log an error, and move to the next site.

src/odemis/driver/test/orsay_test.py
pieleric: Yes, that's correct. The shit.value remains the same, and the center of the FoV remains the same (assuming the calibration is good!).

src/odemis/driver/orsay.py
K4rishma: How do we track this issue? For e.g. through a TODO or a task.

K4rishma (approved):

PR #2361, opened 2023-03-13
Sdev 922 overview image implement cancelling

nandishjpatel:
Ticket: https://delmic.atlassian.net/browse/SDEV-922

### DIFF
--- a/src/odemis/acq/fastem.py
+++ b/src/odemis/acq/fastem.py
@@ -968,7 +968,7 @@ class OverviewAcquisition(object):
     def __init__(self) -> None:
         self.sf = None
 
-    def cancel_acquisition(self, future):
+    def _cancel_acquisition(self, future):
         if self.sf:
             self.sf.cancel()
 
@@ -1005,7 +1005,9 @@ class OverviewAcquisition(object):
         # FIXME removed to provide better progress update in GUI
         #  When _tiledacq.py has proper time update implemented, add this line here again.
         # sf.add_update_callback(_pass_future_progress)
-        f.task_canceller = self.cancel_acquisition
+        # Go one level deeper to cancel the future of the underlying acquisition
+        # TiledAcquisitionTask._cancelAcquisition will hence be called
+        f.task_canceller = self._cancel_acquisition
         das = self.sf.result()
 
         if len(das) != 1:
--- a/src/odemis/acq/fastem.py
+++ b/src/odemis/acq/fastem.py
@@ -915,7 +915,8 @@ def acquireTiledArea(stream, stage, area, live_stream=None):
 
     # Connect the future to the task and run it in a thread.
     # _run_overview_acquisition is executed by the executor and runs as soon as no other task is executed
-    _executor.submitf(f, _run_overview_acquisition, f, sem_stream, stage, area, live_stream)
+    overview_acq = OverviewAcquisition()
+    _executor.submitf(f, overview_acq.run, f, sem_stream, stage, area, live_stream)
 
     return f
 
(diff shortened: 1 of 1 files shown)


tepals (approved):

src/odemis/acq/fastem.py
```diff
@@ -962,49 +963,60 @@ def estimateTiledAcquisitionTime(stream, stage, area):
...
+    """Class to run the acquisition of one overview image (typically one scintillator)."""
+    def __init__(self) -> None:
+        self.sf = None
 
-def _run_overview_acquisition(f, stream, stage, area, live_stream):
-    """
-    Runs the acquisition of one overview image (typically one scintillator).
+    def _cancel_acquisition(self, future):
+        if self.sf:
+            self.sf.cancel()
```
pieleric, line 973: A canceller is suppose to return a bool. Please indicate it in the function signature, and in general, you can `return True` (meaning the cancelling succeeded).

src/odemis/acq/fastem.py
```diff
@@ -962,49 +963,60 @@ def estimateTiledAcquisitionTime(stream, stage, area):
...
 
-    :param f: (ProgressiveFuture) Acquisition future object.
-    :param stream: (SEMstream) The stream used for the acquisition.
-    :param stage: (actuator.MultiplexActuator) The stage in the sample carrier coordinate system.
-        The x and y axes are aligned with the x and y axes of the ebeam scanner.
-    :param area: (float, float, float, float) xmin, ymin, xmax, ymax coordinates of the overview region.
-    :param live_stream: (StaticStream or None): StaticStream to be updated with each tile acquired,
-           to build up live the whole acquisition. NOT SUPPORTED YET.
+    def run(self, f, stream, stage, area, live_stream):
```
pieleric, line 975: Instead of getting the future as argument here, I'd suggest to get it in the `__init__()`. For such "Task" class, which represent one asynchronous execution, it's the common way to do it, and I feel it's more logical.

src/odemis/acq/fastem.py
```diff
@@ -962,49 +963,60 @@ def estimateTiledAcquisitionTime(stream, stage, area):
...
+        # Note, for debugging, it's possible to keep the intermediary tiles with log_path="./tile.ome.tiff"
+        self.sf = stitching.acquireTiledArea([stream], stage, area, overlap, registrar=REGISTER_IDENTITY,
+                                        focusing_method=FocusingMethod.NONE)
+        # Connect the progress of the underlying future to the main future
+        # FIXME removed to provide better progress update in GUI
+        #  When _tiledacq.py has proper time update implemented, add this line here again.
+        # sf.add_update_callback(_pass_future_progress)
+        # Go one level deeper to cancel the future of the underlying acquisition
+        # TiledAcquisitionTask._cancelAcquisition will hence be called
+        f.task_canceller = self._cancel_acquisition
```
pieleric, line 1010: Let's put this at the init. It will be clearer that placing it in the middle of the code.

src/odemis/acq/fastem.py
```diff
@@ -962,49 +963,60 @@ def estimateTiledAcquisitionTime(stream, stage, area):
...
 
-    # Switch immersion mode back on, so we can focus the SEM from the TFS GUI.
-    stream.emitter.immersion.value = True
+        # Switch immersion mode back on, so we can focus the SEM from the TFS GUI.
+        stream.emitter.immersion.value = True
 
-    # FIXME auto blanking not working properly, so force beam blanking after image acquisition for now.
-    stream.emitter.blanker.value = True
+        # FIXME auto blanking not working properly, so force beam blanking after image acquisition for now.
+        stream.emitter.blanker.value = True
```
pieleric, line 1020: These lines should actually be in a (try)/finally block. So even if the sub future fails or is cancelled the hardware is still reset to a proper state. (That should have been done in the original code, but now it's even more important, as cancelling is going to be much more common that acquisition failure, hopefully!)

pieleric (changes requested):

src/odemis/acq/fastem.py
```diff
@@ -995,28 +995,22 @@ def run(self, f, stream, stage, area, live_stream):
...
-            f.set_progress(start, end)
-
         # Note, for debugging, it's possible to keep the intermediary tiles with log_path="./tile.ome.tiff"
-        self.sf = stitching.acquireTiledArea([stream], stage, area, overlap, registrar=REGISTER_IDENTITY,
+        self._sub_future = stitching.acquireTiledArea([stream], stage, area, overlap, registrar=REGISTER_IDENTITY,
                                         focusing_method=FocusingMethod.NONE)
-        # Connect the progress of the underlying future to the main future
-        # FIXME removed to provide better progress update in GUI
-        #  When _tiledacq.py has proper time update implemented, add this line here again.
-        # sf.add_update_callback(_pass_future_progress)
```
pieleric: It's worth to keep. Eventually that should be the right thing to do. So let's keep track of it. 
It just needs to keep the a reference to the `future` in `__init__()`, so that this is still possible.

src/odemis/gui/cont/fastem_acq.py
```diff
@@ -251,12 +251,8 @@ def on_acquisition_done(self, future, num):
         try:
             da = future.result()
         except CancelledError:
-            self._reset_acquisition_gui()
             return
         except Exception:
-            # leave the gauge, to give a hint on what went wrong.
```
pieleric, line 257: Please add a comment that we do nothing, because it's all taken care by `full_acquisition_done()`.

src/odemis/acq/fastem.py
```diff
@@ -962,49 +963,54 @@ def estimateTiledAcquisitionTime(stream, stage, area):
...
-    stream.emitter.blanker.value = True
+            # FIXME auto blanking not working properly, so force beam blanking after image acquisition for now.
+            stream.emitter.blanker.value = True
 
-    return das[0]
+        if len(das) == 1:
+            return das[0]
+        else:
+            logging.warning("Expected 1 DataArray, but got %d: %r", len(das), das)
+            return []
```
pieleric, line 1016: If there are too many DataArrays, it's nice to at least pass one, instead of discarding everything. You can do that so:
```suggestion
            return das[:1]
```

pieleric (changes requested):

pieleric (changes requested):
The code looks good. However, there should be a test case for that. 
See TestFASTEMOverviewAcquisition. You can add a new test case that starts an acquisition and cancels it a few seconds later. 
(Sorry for not thinking of this comment before!!)

nandishjpatel:
> The code looks good. However, there should be a test case for that. See TestFASTEMOverviewAcquisition. You can add a new test case that starts an acquisition and cancels it a few seconds later. (Sorry for not thinking of this comment before!!)

Added the testcase :)

src/odemis/acq/test/fastem_test.py
```diff
@@ -130,6 +131,34 @@ def test_estimateTiledAcquisitionTime(self):
...
+        scintillator5_area = (-0.007, -0.007, 0.007, 0.007)  # l, b, r, t
+
+        # Future to acquire the overview image
+        f = fastem.acquireTiledArea(self.stream, self.stage, scintillator5_area)
+        f.add_done_callback(self.on_acquisition_done)
+
+        # Wait a bit for the acquisition to happen and then cancel it
+        time.sleep(5)
+        f.cancel()
+        # Assert that the acquisition was cancelled
```
pieleric, line 149: As it runs in a separate thread, the task might take some time to cancel, and eventually call the `on_acquisition_done()`. So I'd suggest to insert here a `time.sleep(1)` to make sure the test passes on all systems.

src/odemis/gui/cont/fastem_acq.py
```diff
@@ -251,12 +251,10 @@ def on_acquisition_done(self, future, num):
         try:
             da = future.result()
         except CancelledError:
-            self._reset_acquisition_gui()
+            # Just return because the error is handled in full_acquisition_done
```
pieleric, line 254: You can drop the special exception handling for `CancelledError`, as now it does the same as the `Exception`.

pieleric (changes requested):

pieleric (approved):

PR #2364, opened 2023-03-14
gui win acquisition: extend overview acquisition to support MIMAS

pieleric:


K4rishma (approved):

src/odemis/gui/main.xrc
```diff
@@ -9010,15 +9010,15 @@
...
                           <fg>#DDDDDD</fg>
-                          <XRCED>
-                            <assign_var>1</assign_var>
-                          </XRCED>
                           <font>
                             <size>9</size>
                           </font>
+                          <XRCED>
+                              <assign_var>1</assign_var>
+                          </XRCED>
```
tepals, line 9021: out of curiosity what does this do (I guess it sets the default value to 1)? And why does it need to be after setting the font size?

src/odemis/gui/win/acquisition.py
```diff
@@ -27,29 +27,34 @@
...
-import wx
 
 import odemis.gui.model as guimodel
-from odemis import model, dataio
-from odemis.acq import stream, path, acqmng, stitching
-from odemis.acq.stitching import FocusingMethod, WEAVER_MEAN, REGISTER_IDENTITY, acquireOverview
-from odemis.acq.stream import EMStream, NON_SPATIAL_STREAMS, OpticalStream, ScannedFluoStream, LiveStream
-from odemis.gui.acqmng import presets, preset_as_is, apply_preset, \
-    get_global_settings_entries, get_local_settings_entries
+import wx
```
tepals, line 35: why is `import wx` here and not grouped with numpy? If there is a specific reason, can you add a comment?

src/odemis/gui/win/acquisition.py
```diff
@@ -894,7 +923,11 @@ def update_acquisition_time(self):
         if self.acquiring:
             return
 
-        if not self.area:
+        areas = self._get_areas()
+
+        if not areas[0]:
+            # This can happen if the stage is situated outside of the active range
+            # (and there is no sample in use
```
tepals, line 930: you forgot the closing bracket

src/odemis/gui/win/acquisition.py
```diff
@@ -904,11 +937,26 @@ def update_acquisition_time(self):
...
+                                                  registrar=REGISTER_IDENTITY,
+                                                  zlevels=zlevels,
+                                                  focusing_method=focus_mtd)
+
+            else:
+                acq_time = stitching.estimateTiledAcquisitionTime(streams,
+                                                              self._main_data_model.stage,
+                                                              self.area, self.overlap,
+                                                              zlevels=zlevels,
+                                                              focusing_method=focus_mtd)
```
tepals, line 959: Can you align the input arguments?
```python
            if self.autofocus_roi_ckbox.value:
                acq_time = stitching.estimateOverviewTime(streams,
                                                          self._main_data_model.stage,
                                                          areas,
                                                          .......
                                                          zlevels=zlevels,
                                                          focusing_method=focus_mtd)

            else:
                acq_time = stitching.estimateTiledAcquisitionTime(streams,
                                                                  self._main_data_model.stage,
                                                                  self.area, self.overlap,
                                                                  zlevels=zlevels,
                                                                  focusing_method=focus_mtd)
```

src/odemis/gui/win/acquisition.py
```diff
@@ -1033,6 +1081,28 @@ def _fit_view_to_area(self):
         fov = (self.area[2] - self.area[0], self.area[3] - self.area[1])
         self.pnl_view_acq.set_mpp_from_fov(fov)
 
+    def _get_areas(self) -> List[Tuple[float]]:
+        """
+        return: the bounding boxes (in physical coordinates) of all areas to acquire
```
tepals, line 1086: can you add the order of the boundingbox, like: xmin, ymin, xmax, ymax

src/odemis/gui/win/acquisition.py
```diff
@@ -1033,6 +1081,28 @@ def _fit_view_to_area(self):
...
+
+        # TODO: for now this is only for the MIMAS, but eventually, on the METEOR,
+        # which often supports 2 grids, this should also be possible to use.
+
+        # .sample_centers contains the center position of each area
+        # .sample_rel_bbox contains the relative bounding box on an area
+        rel_bbox = self._main_data_model.sample_rel_bbox
+
+        # TODO: once there are toggle buttons, filter to only the button toggled on
+        # Sort the centers along X, and for centers with the same X, along the Y.
```
tepals, line 1099: is this the order in which they will be acquired?

tepals (changes requested):

src/odemis/gui/main.xrc
pieleric: It tells the xrc-to-python generator to create an attribute on the class with the name of this widget (so that it's easy to access in the code). 
The position doesn't matter. I put it at the end because it was easier for me to check it matches the right XML element (XML is just hard to read).

src/odemis/gui/win/acquisition.py
pieleric: Indeed. The actual order doesn't matter much (maybe just saves a few seconds of stage movement at best), but what's important is that it's always the same order, to avoid "surprises" to the user. I'll add this to the comment.

src/odemis/gui/win/acquisition.py
pieleric: No, it just got automatically reordered by my IDE when adding the new import. I'll put it back where it belongs.

tepals (approved):

PR #2365, opened 2023-03-15
metadata: add MD_ERROR to store errors corresponding to calibrations …

tepals:
…in a dictionary.


nandishjpatel (approved):

pieleric:
Could you explain how you're going to use it? Or point me to a specification document? It looks like you could use log for that...

nandishjpatel:
> Could you explain how you're going to use it? Or point me to a specification document? It looks like you could use log for that...

@tepals indeed had logged this error. The reason to have this metadata is for Zabbix monitoring. An example how we get the `FACTORY_CALIBRATION` and `CALIBRATED` metadata out of a component is [here](https://bitbucket.org/delmic/fastem-calibrations/src/64fe6a224c2f3396548c0d8263e612116c5cafa8/src/fastem_calibrations/health_monitoring/calibration_info.py#lines-32). Do let is know your opinion, if it make sense to have this as metadata.

PR #2366, opened 2023-03-15
add yaml loader tests

FCanberk:


pieleric (changes requested):
That's not enough. See the spec of https://delmic.atlassian.net/browse/MIMAS-984 .
`load_config()` should eventually return a list of `MillingSettings`. That is the format that `mill_features()` takes as input. So you have to modify/extend `load_config()` to convert the dict read from YAML into `MillingSettings` (note that there should be clever ways to do it assuming that the names of the VAs are identical to the keys in YAML. 
Be careful to raise a proper error if a value is not the right format. (For instance, the `pixelSize` is a single value instead of a tuple).

src/odemis/acq/millmng.py
```diff
@@ -323,7 +350,7 @@ def _mill_all_settings(self, site: CryoFeature):
             # X is horizontal axis, Y is vertical axis
             pxs = milling_settings.pixelSize.value
             probe_size = max(pxs)
-            overlap = [1 - (pxs[0] / probe_size), 1 - (pxs[1] / probe_size)]  # Always >= 0
+            overlap = [1 - (float(pxs[0]) / float(probe_size)), 1 - (float(pxs[1]) / float(probe_size))]  # Always >= 0
```
pieleric, line 353: What are you trying to do here with the `float()`? If pxs and probe_size are numbers, I don't think this will have any effect. Could this be that probe_size is a string? If you rebase against master you should get my change that should help catch such issue early.

src/odemis/odemisd/test/milling-series-test-missing-attributes.mill.yaml
```diff
@@ -0,0 +1,14 @@
+# Milling settings for testing purposes. Duration attribute is missing in this example
+"Missing milling duration": {
+    scanner: "FIB Scanner",
+    iteration: 5,
+    probe_size: 5e-6,
```
pieleric, line 5: "probe_size" -> "pixel_size" + tuple.

src/odemis/odemisd/test/milling-series-test-working-config.mill.yaml
```diff
@@ -0,0 +1,31 @@
...
+    current: 100e-12,
+    horizontal_fov: 35e-6, # m
+    roi: [0, 0, 1, 1],
+    pixel_size: [4e-9, 8e-10],  # note that the 4e-9 / 5 notation is interpreted as str and cannot be converted to float
+    duration: 300, # s
+    beam_angle: 0.15, # rad
+    dc_roi: [0.05, 0.2, 0.32, 0.8],
+    dc_period: 60,
+    dc_dwell_time: 5e-6,
+    dc_current: 100e-12,
```
pieleric, line 14: Make these values float.

src/odemis/acq/test/millmng_test.py
```diff
@@ -71,6 +73,8 @@ class MillingManagerTestCase(unittest.TestCase):
 
     @classmethod
     def setUpClass(cls):
+        cls.millings = load_config(MILLING_CONFIG_PATH + "/milling-series-test-working-config.mill.yaml")
+
```
pieleric, line 77: Please move this call to each function that test `load_config()`. We typically don't want setUpClass to test the code.

src/odemis/acq/test/millmng_test.py
```diff
@@ -334,6 +338,34 @@ def on_progress_update(self, future, start, end):
         self.end = end
         self.updates += 1
 
+    def test_yaml_loader_attribute_types(self):
+        """
+        Test the data types of the milling attributes
+        """
+        for milling in self.millings:
```
pieleric, line 345: You can also check that there are 2 millings.

src/odemis/odemisd/test/milling-series-test-missing-attributes.mill.yaml
```diff
@@ -0,0 +1,14 @@
+# Milling settings for testing purposes. Duration attribute is missing in this example
+"Missing milling duration": {
+    scanner: "FIB Scanner",
```
pieleric, line 3: This attribute doesn't exist.
Please make sure this YAML file is wrong for only 1 reason.  (the dicts are not inside a list, etc)

pieleric (changes requested):

PR #2368, opened 2023-03-17
Stage switched to FM imaging mode

K4rishma:
Fixing failing test case @sneepstefan 

roi_autofocus_test.py::RoiAutofocusTestCase::test_autofocus_in_roi on Ubuntu 20

### DIFF
--- a/src/odemis/acq/align/test/roi_autofocus_test.py
+++ b/src/odemis/acq/align/test/roi_autofocus_test.py
@@ -21,6 +21,8 @@ import odemis
 from odemis import model
 from odemis.acq.align.autofocus import estimateAutoFocusTime
 from odemis.acq.align.roi_autofocus import autofocus_in_roi, estimate_autofocus_in_roi_time
+from odemis.acq.move import cryoSwitchSamplePosition, FM_IMAGING
+from odemis.driver import simsem
 from odemis.util import testing
 
 logging.getLogger().setLevel(logging.DEBUG)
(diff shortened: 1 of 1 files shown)


tepals (approved):

pieleric (approved):
OK, but we need a little more explanation, at least in the commit log, about *why* 0.8 -> 0.2. 
It's because in the simulator the autofocus returns a low confidence, so it discards too many points? 
However, the meteor sim does simulate the focus. So why would the autofocus return a low confidence?

src/odemis/acq/align/test/roi_autofocus_test.py
```diff
@@ -39,12 +41,12 @@ class RoiAutofocusTestCase(unittest.TestCase):
     def setUpClass(cls):
         testing.start_backend(METEOR_CONFIG)
 
-        # create some streams connected to the backend
-        # cls.microscope = model.getMicroscope()
         cls.ccd = model.getComponent(role="ccd")
         cls.focus = model.getComponent(role="focus")
         cls.stage = model.getComponent(role="stage")
 
+        cryoSwitchSamplePosition(FM_IMAGING).result()
```
pieleric, line 48: Can you add a comment explaining why this is needed? I think that's because in this mode, the focus position will be near its "good" position and so the simulator will simulate (almost) in focused images, so the autofocus will work .

pieleric (changes requested):
Looks like a much better solution.

tepals (approved):

PR #2371, opened 2023-03-20
acq millmng: fix milling with drift correction milling too long

pieleric:
The milling is cut in N passes. For each pass, the milling would mill
for N * duration. It should be just 1 * duration.

### DIFF
--- a/src/odemis/acq/millmng.py
+++ b/src/odemis/acq/millmng.py
@@ -355,7 +355,7 @@ class MillingRectangleTask(object):
 
                         self._scanner.probeCurrent.value = milling_settings.current.value
                         self._future.running_subf = mill_rectangle(milling_settings.roi.value, self._scanner,
-                                                                   iteration=self._iterations[setting_nb],
+                                                                   iteration=1,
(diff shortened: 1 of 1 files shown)


FCanberk (approved):

src/odemis/acq/millmng.py
```diff
@@ -355,7 +355,7 @@ def _mill_all_settings(self, site: CryoFeature):
 
                         self._scanner.probeCurrent.value = milling_settings.current.value
                         self._future.running_subf = mill_rectangle(milling_settings.roi.value, self._scanner,
-                                                                   iteration=self._iterations[setting_nb],
```
K4rishma, line 358: This will also provide iteration 1 as it is calculated from the init of this class.

K4rishma (approved):

src/odemis/acq/millmng.py
pieleric: No. This is the case where there is a dritf correction. It's used line 351 for the `for` loop. It's not 1.

PR #2375, opened 2023-03-21
Zeiss stage FLM-SEM switch implementation

K4rishma:
New coordinates namely x,y,z,rx,rm and m are used for the switch. Extended the test cases of Meteor move in move_test.py,

Following tasks are completed
- renaming rz to rm
- conversion formulas for SEM-FLM switch
- add meteor-zeiss-sim.odm.yaml file with different metadata values and new coordinate axes
- extend the test cases of Meteor stage movement in move_test.py
    - tested on simulation yaml file
    - tested on Zeiss hardware without the Meteor
 - test the code using GUI in simulation mode


K4rishma:
@pieleric I would like to create a PR to a different branch in Odemis. This PR is wrong. I will try to correct it.

plugins/meteor_engage_warn.py
```diff
@@ -68,7 +68,7 @@ def _display_meteor_pos_warning_msg(self, end_pos: Dict[str, float]) -> bool:
         return: True if the user accepts, False if the move should be cancelled.
         """
         pos_str = []
-        for axis in ("x", "y", "z", "rx", "ry", "rz"):
+        for axis in ("x", "y", "m", "rx", "ry", "rm"):
```
sneepstefan, line 71: `for axis in ("x", "y", "m", "rx", "ry", "rm"):`
Take into account that the driver>zeiss.py works with axes definition of "z" instead of "m". I am not sure if you run into issues regarding that.

src/odemis/driver/test/zeiss_test.py
```diff
@@ -1,3 +1,4 @@
+
```
sneepstefan, line 1: Could you remove this blank line and so keep the zeiss_test.py file unchanged?

src/odemis/driver/zeiss.py
```diff
@@ -1116,9 +1116,9 @@ def _move_to_pos(self, pos):
         pos = numpy.append(pos, self.pos[len(pos):])
         self.target_pos = numpy.array(pos)
         self._start_move = time.time()
-        dur_lin = sum(numpy.sqrt((self.pos[[0, 1, 2, 5]] - self.target_pos[[0, 1, 2, 5]]) ** 2)) / self._speed_lin
-        dur_rot = sum(abs(self.pos[[3, 4]] - self.target_pos[[3, 4]])) / self._speed_rot
-        dur = dur_lin + dur_rot
+        dur_lin = max(numpy.sqrt((self.pos[[0, 1, 2, 5]] - self.target_pos[[0, 1, 2, 5]]) ** 2)) / self._speed_lin
```
sneepstefan, line 1119: Could you comment on this in the code, why did you switch to max? Does this better represent the real situation for example or were you forced to change this due to increasing the linear move speed?

src/odemis/gui/cont/tabs.py
```diff
@@ -2595,8 +2595,6 @@ def __init__(self, name, button, panel, main_frame, main_data):
                 raise ValueError("The stage misses the 'SEM_IMAGING_RANGE' metadata.")
             if not model.MD_FM_IMAGING_RANGE in stage_metadata:
                 raise ValueError("The stage misses the 'FM_IMAGING_RANGE' metadata.")
-            if not model.MD_POS_COR in stage_metadata:
-                raise ValueError("The stage misses the 'POS_COR' metadata.")
```
sneepstefan, line 2599: Does this not break backwards compatibility? What about the METEOR's in Martinsreid using POS_COR for example?

src/odemis/gui/cont/tabs.py
```diff
@@ -3170,7 +3168,7 @@ def _display_meteor_pos_warning_msg(self, end_pos) -> bool:
         return: True if the user accepts, False if the move should be cancelled.
         """
         pos_str = []
-        for axis in ("x", "y", "z", "rx", "ry", "rz"):
+        for axis in ("x", "y", "m", "rx", "ry", "rm"):
```
sneepstefan, line 3171: Same comment here as in [meteor_engage_warn.py].

src/odemis/acq/move.py
```diff
@@ -576,6 +586,7 @@ def _doCryoSwitchSamplePosition(future, target):
 
     try:
         if role == "enzel":
+            # TODO enzel code broken as rm is used in other functions insteas of rm
```
sneepstefan, line 589: is this comment line as it should? referring 2 times to 'rm'? There is also a typo here: `insteas of rm` should be `instead of rm`

src/odemis/acq/move.py
```diff
@@ -735,30 +746,67 @@ def _doCryoSwitchSamplePosition(future, target):
...
 
             if target in (GRID_1, GRID_2):
                 # The current mode doesn't change. Only X/Y/Z should move (typically
                 # only X/Y).
-                sub_moves.append((stage, filter_dict({'x', 'y', 'z'}, target_pos)))
+                # It could be both SEM and FM position # TODO check
+                # TODO How to work Linked YM position and its associated not used METADATA POS_COR, CALIB_VAL
+                sub_moves.append((stage, filter_dict({'x', 'y', 'm', 'z'}, target_pos)))
+                sub_moves.append((stage, filter_dict({'rx', 'rm'}, target_pos)))
+                # x, y, m
```
sneepstefan, line 768: `# x, y, m`
What is this comment for?

src/odemis/acq/move.py
```diff
@@ -875,40 +923,81 @@ def _doCryoSwitchSamplePosition(future, target):
...
-# The rotation angles are constant existing in "FM_POS_ACTIVE" metadata,
-# but the translation are calculated based on the current position and some
-# correction/shifting parameters existing in metadata "FM_POS_ACTIVE". 
-# This correction parameters can change every session. They are calibrated
-# at the beginning of each run.
+
+def check_keys_in_md_calib(calibrated_values: dict) -> bool:
+    """
+    Check the required keys in the calibrated values.
+    :param calibrated_values:
```
sneepstefan, line 930: short description of the param calibrated_values

src/odemis/acq/move.py
```diff
@@ -875,40 +923,81 @@ def _doCryoSwitchSamplePosition(future, target):
...
+    pos (dict str->float): the current stage position. The position has to have x, y, m, z, rx and rm axes,
         otherwise error would be raised.
     stage (Actuator): the stage component. The stage has to have metadata "MD_POS_COR" and "MD_FAV_FM_POS_ACTIVE"
-    return (dict str->float): the transformed position. It returns the updated axes x, y, rx, rz. The axis z is same as the input.
+    return (dict str->float): the transformed position. It returns the updated axes x, y, m, z, rx, rm. The axis z is same as the input.
     """
-    if not {"x", "y", "rz", "rx"}.issubset(stage.axes):
-        raise KeyError("The stage misses 'x', 'y', 'rx' or 'rz' axes")
+    if not {"x", "y", "m", "z", "rx", "rm"}.issubset(stage.axes):
+        raise KeyError("The stage misses 'x', 'y', 'm', 'z', 'rx' or 'rm' axes")
```
sneepstefan, line 951:
```suggestion
    if not {"x", "y", "m", "z", "rx", "rm"}.issubset(stage.axes):
        raise KeyError("The stage misses %s axes" %({'x', 'y', 'm', 'z', 'rx', 'rm'} - stage.axes.keys()))
```

src/odemis/acq/move.py
```diff
@@ -875,40 +923,81 @@ def _doCryoSwitchSamplePosition(future, target):
...
     stage_md = stage.getMetadata()
     transformed_pos = pos.copy()
-    pos_cor = stage_md[model.MD_POS_COR]
```
sneepstefan, line 898: does this not have an effect on backwards compatibility?

src/odemis/acq/move.py
```diff
@@ -917,17 +1006,49 @@ def transformFromMeteorToSEM(pos, stage):
     stage (Actuator): the stage component
     returns (dict str->float): the transformed stage position. 
     """
-    if not {"x", "y", "rx", "rz"}.issubset(stage.axes):
-        raise KeyError("The stage misses 'x', 'y', 'rx' or 'rz' axes")
+    if not {"x", "y", "m", "z", "rx", "rm"}.issubset(stage.axes):
+        raise KeyError("The stage misses 'x', 'y', 'm', 'z', 'rx' or 'rm' axes")
```
sneepstefan, line 1007:
```suggestion
        raise KeyError("The stage misses %s axes" %({'x', 'y', 'm', 'z', 'rx', 'rm'} - stage.axes.keys()))
```

src/odemis/acq/move.py
```diff
@@ -917,17 +1006,49 @@ def transformFromMeteorToSEM(pos, stage):
...
     stage_md = stage.getMetadata()
     transformed_pos = pos.copy()
-    pos_cor = stage_md[model.MD_POS_COR]
```
sneepstefan, line 924: same comment as before..

src/odemis/acq/move.py
K4rishma: It does not works well with the calculation in the message, so I have done a bit differently
`    if not {"x", "y", "m", "z", "rx", "rm"}.issubset(stage.axes):
        missed_axes = {'x', 'y', 'm', 'z', 'rx', 'rm'} - stage.axes.keys()
        raise KeyError("The stage misses %s axes" % missed_axes)`

src/odemis/acq/test/move_test.py
```diff
@@ -710,6 +897,55 @@ def test_lin_rot_axes(self):
...
+        progress = getMovementProgress(current_point, start_point, end_point)
+        self.assertTrue(util.almost_equal(progress, 0.5, rtol=RTOL_PROGRESS))
+
+        current_point = {'x': .998, 'y': .999, 'z': .999}  # slightly off the line
+        progress = getMovementProgress(current_point, start_point, end_point)
+        self.assertTrue(util.almost_equal(progress, 0.5, rtol=RTOL_PROGRESS))
+
+        # Same in the other direction
+        act_rot_error = _getDistance(point1, point2)
+        self.assertAlmostEqual(exp_rot_error, act_rot_error)
```
sneepstefan, line 916: Where is exp_rot_error assigned?

src/odemis/acq/test/move_test.py
```diff
@@ -710,6 +897,55 @@ def test_lin_rot_axes(self):
...
+    def test_rotation_axes_missing_axis(self):
+        point1 = {'rx': numpy.radians(30), 'rm': numpy.radians(30)}  # 30 degree
+        # No rx => doesn't count it
+        point2 = {'rm': numpy.radians(60)}  # 60 degree
+        exp_rot_dist = ROT_DIST_SCALING_FACTOR * numpy.radians(30)
+        act_rot_dist = _getDistance(point2, point1)
+        self.assertAlmostEqual(exp_rot_dist, act_rot_dist)
+
+        # Same in the other direction
+        act_rot_dist = _getDistance(point2, point1)
```
sneepstefan, line 927: Could you point out how the same values passed to _getDistance will verify 'other' direction?

sneepstefan (changes requested):

install/linux/usr/share/odemis/sim/meteor-zeiss-sim.odm.yaml
```diff
@@ -0,0 +1,237 @@
...
+"Stage": {
+    role: stage-bare,
+    init: {
+        rng: {
+            # For Zeiss XB550
+            "x": [0, 102.5e-3],
+            "y": [0, 100.e-3],
+            "z": [0, 50.e-3],
+            "rx": [-0.06981317007977318, 0.7853981633974],  # rad (-4° -> 45°) # T tilt , -4° is the minimum
+            "rm": [0.0, 6.283185307179586], # rad (0° -> 360°) # R , rz, rotation
```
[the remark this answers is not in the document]
sneepstefan: That's a valid intention indeed

src/odemis/acq/test/move_test.py
```diff
@@ -281,30 +287,101 @@ def test_smaract_stage_fallback_movement(self):
...
         # get the stage components
         cls.stage = model.getComponent(role="stage-bare")
+        cls.linked_ym_stage = model.getComponent(role="stage")
 
         # get the metadata
         stage_md = cls.stage.getMetadata()
         cls.stage_grid_centers = stage_md[model.MD_SAMPLE_CENTERS]
         cls.stage_loading = stage_md[model.MD_FAV_POS_DEACTIVE]
 
+    def test_transformFromSEMToMeteor(self):
```
pieleric, line 321: Please add a test that takes a SEM position, and convert it to FM, and convert it back to SEM. That final position should be the same as the original position. You can easily try with many positions.
In some way it's a weaker test than the two current tests, but first, it checks that the two functions are opposite, and secondly, it's less "brittle". Even if you change a bit the formula, you don't need to update the test case.

pieleric (changes requested):

src/odemis/acq/move.py
```diff
@@ -875,40 +919,84 @@ def _doCryoSwitchSamplePosition(future, target):
...
+        missed_axes = {'x', 'y', 'm', 'z', 'rx', 'rm'} - stage.axes.keys()
+        raise KeyError("The stage misses %s axes" % missed_axes)
+
     stage_md = stage.getMetadata()
     transformed_pos = pos.copy()
-    pos_cor = stage_md[model.MD_POS_COR]
+
+    # Call out calibrated values and stage tilt and rotation angles
+    calibrated_values = stage_md[model.MD_CALIB]
+    check_keys_in_md_calib(calibrated_values)
```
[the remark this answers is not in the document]
K4rishma: The value error raises with a message in the function that a certain key is missing

src/odemis/acq/move.py
```diff
@@ -875,40 +919,84 @@ def _doCryoSwitchSamplePosition(future, target):
...
-# correction/shifting parameters existing in metadata "FM_POS_ACTIVE". 
-# This correction parameters can change every session. They are calibrated
-# at the beginning of each run.
+
+def check_keys_in_md_calib(calibrated_values: dict) -> bool:
+    """
+    Check the required keys in the calibrated values.
+    :param calibrated_values:Is a dictionary containing the calibrated values to carry out the transformation
+        from SEM to FM.
+    :return: True if all keys are in the calibrated values, False otherwise.
```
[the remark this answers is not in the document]
K4rishma: I have change the doc string as we need to raise instead of returning False. This function is modified to include no return value. The function just raises if the keys are not proper.

src/odemis/acq/test/move_test.py
K4rishma: Sorry this is not my code, I will align with the master

plugins/meteor_engage_warn.py
K4rishma: We need both z and m axes for the movement. I checked in the config, all the relevant axes are defined.  It is an interesting point.

install/linux/usr/share/odemis/sim/meteor-zeiss-sim.odm.yaml
```diff
@@ -0,0 +1,237 @@
...
+    },
+}
+# not sure if all z will be changed to m below, for now m is used
+"Meteor Stage": {
+    class: actuator.MultiplexActuator,
+    role: stage,
+    dependencies: { "x": "Stage", "y": "Linked YM", "z": "Linked YM", },
+    init: {
+        axes_map: { "x": "x", "y": "x", "z": "y",},
+    },
```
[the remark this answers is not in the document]
pieleric: In some way, it'd be better to actually use that argument, to make sure it works (but I've already tried, it does). On the other hand it'd make the testing a little bit less handy has the all the values of the stage vs stage-bare will be inverted. So let's say it's good enough.

src/odemis/acq/move.py
```diff
@@ -125,14 +128,18 @@ def getCurrentGridLabel(current_pos, stage):
...
     if current_pos_label == SEM_IMAGING:
+        grid1_pos.update(stage_md[model.MD_FAV_SEM_POS_ACTIVE])
+        grid2_pos.update(stage_md[model.MD_FAV_SEM_POS_ACTIVE])
         distance_to_grid1 = _getDistance(current_pos, grid1_pos)
         distance_to_grid2 = _getDistance(current_pos, grid2_pos)
-        return GRID_2 if distance_to_grid1 > distance_to_grid2 else GRID_1 
+        return GRID_2 if distance_to_grid1 > distance_to_grid2 else GRID_1
     elif current_pos_label == FM_IMAGING:
+        grid1_pos.update(stage_md[model.MD_FAV_FM_POS_ACTIVE])
+        grid2_pos.update(stage_md[model.MD_FAV_FM_POS_ACTIVE])
```
pieleric, line 138: Sorry, I mistook. I now see that this code is here used to differentiate between grid1 and grid2 position. So we need to make sure that the gridN_pos constains the x,y,z values from SAMPLE_CENTERS. Therefore, that should also be done inverted, as above. BTW, you can do it before the `if`, to avoid duplication, in line 128.

src/odemis/acq/move.py
```diff
@@ -576,6 +586,7 @@ def _doCryoSwitchSamplePosition(future, target):
 
     try:
         if role == "enzel":
+            # TODO enzel code broken as rm is used in other functions insteasd of rz
```
pieleric, line 589: Well, let's start by not modifying the enzel code! All the lines below which are in this block should be reverted.

src/odemis/acq/move.py
```diff
@@ -592,14 +603,14 @@ def _doCryoSwitchSamplePosition(future, target):
...
+            required_axes = {'x', 'y', 'z', 'rx', 'rm'}
             for stage_position in target_pos.values():
                 if not required_axes.issubset(stage_position.keys()):
                     raise ValueError("Stage %s metadata does not have all required axes %s." % (
-                    list(stage_md.keys())[list(stage_md.values()).index(stage_position)], required_axes))
+                        list(stage_md.keys())[list(stage_md.values()).index(stage_position)], required_axes))
             current_pos = stage.position.value
             # To hold the sub moves to run if normal ordering failed
-            fallback_submoves = [{'x', 'y', 'z'}, {'rx', 'rz'}]
+            fallback_submoves = [{'x', 'y', 'z'}, {'rx', 'rm'}]
```
pieleric, line 613: Drop this change (ENZEL).

src/odemis/acq/move.py
```diff
@@ -735,30 +746,63 @@ def _doCryoSwitchSamplePosition(future, target):
             # If at some "weird" position, it's quite unsafe. We consider the targets
             # LOADING and SEM_IMAGING safe to go. So if not going there, first pass
             # by SEM_IMAGING and then go to the actual requested position.
+            # logging.info("Current position is %s", stage.position.value)
```
pieleric, line 749: Unused logging -> drop.

src/odemis/acq/move.py
```diff
@@ -875,40 +919,82 @@ def _doCryoSwitchSamplePosition(future, target):
...
-# The rotation angles are constant existing in "FM_POS_ACTIVE" metadata,
-# but the translation are calculated based on the current position and some
-# correction/shifting parameters existing in metadata "FM_POS_ACTIVE". 
-# This correction parameters can change every session. They are calibrated
-# at the beginning of each run.
+
+def check_stage_metadata(calibrated_values: dict):
+    """
+    Check the required keys in the stage metadata based on given stage type in the microscope file.
+    :param calibrated_values: calibrated values to carry out stage transformation.
```
pieleric, line 926: I think it'd be nicer to either take the "stage" as input. The function can then decided which metadata to look at. You could even not take *any* hardware, and the function would itself call `.getComponent()` to get all the components that needs to be checked (ex: stage, stage-bare, focus).

src/odemis/acq/move.py
```diff
@@ -982,7 +1103,8 @@ def run_sub_move(future, component, sub_move):
 
             logging.debug("Performing sub move %s -> %s", component.name, sub_move)
             future._running_subf = component.moveAbs(sub_move)
-        future._running_subf.result(timeout=MAX_SUBMOVE_DURATION)
+        # future._running_subf.result(timeout=MAX_SUBMOVE_DURATION)
+        future._running_subf.result()
```
pieleric, line 1107: Why is that? Was it for debugging some issue? It probably should be reverted. Or then increase the MAX_SUBMOVE_DURATION.

pieleric (changes requested):

pieleric (approved):

sneepstefan (approved):

PR #2376, opened 2023-03-22
util conversion: Add **kwargs to the init of YamlExtraDumper, to fix …

tepals:
…passing along sort_keys

In 18140fbad328520717036585bdaad625deca7b15 the keyword argument sort_keys was removed for compatibility with ubuntu 18. However this caused the test case and starting of odemis to fail on ubuntu 20, because the class is called with the argument sort_keys in yaml.dump_all(). This fix adds **kwargs to the input arguments of YamlExtraDumper class and does not pass the keyword arguments along, for compatibility with both ubuntu 18 and 20.

### DIFF
--- a/src/odemis/util/conversion.py
+++ b/src/odemis/util/conversion.py
@@ -502,17 +502,17 @@ CustomSafeRepresenter.add_multi_representer(
 class YamlExtraDumper(Emitter, Serializer, CustomSafeRepresenter, Resolver):
 
     def __init__(self, stream,
-            default_style=None, default_flow_style=False,
-            canonical=None, indent=None, width=None,
-            allow_unicode=None, line_break=None,
-            encoding=None, explicit_start=None, explicit_end=None,
-            version=None, tags=None):
+                 default_style=None, default_flow_style=False,
+                 canonical=None, indent=None, width=None,
+                 allow_unicode=None, line_break=None,
+                 encoding=None, explicit_start=None, explicit_end=None,
+                 version=None, tags=None, **kwargs):
         Emitter.__init__(self, stream, canonical=canonical,
-                indent=indent, width=width,
-                allow_unicode=allow_unicode, line_break=line_break)
+                         indent=indent, width=width,
+                         allow_unicode=allow_unicode, line_break=line_break)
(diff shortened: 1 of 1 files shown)


K4rishma (approved):

nandishjpatel (approved):

src/odemis/util/conversion.py
```diff
@@ -502,17 +502,17 @@ def represent_numpy_floating(self, data):
...
+                         indent=indent, width=width,
+                         allow_unicode=allow_unicode, line_break=line_break)
         Serializer.__init__(self, encoding=encoding,
-                explicit_start=explicit_start, explicit_end=explicit_end,
-                version=version, tags=tags)
+                            explicit_start=explicit_start, explicit_end=explicit_end,
+                            version=version, tags=tags)
         CustomSafeRepresenter.__init__(self, default_style=default_style,
-                default_flow_style=default_flow_style)
+                                       default_flow_style=default_flow_style)
```
pieleric, line 517: I think we should still pass "sort_keys" if it's received. Either just put a `**kwargs` here and hope for the best, or explicitly look in kwargs like that:
```python
kwargs_representer = {}
if "sort_keys" in kwargs:
    kwargs_representer["sort_keys"] = kwargs["sort_keys"]
CustomSafeRepresenter(..., **kwargs_representer)
```

src/odemis/util/conversion.py
```diff
@@ -502,17 +502,17 @@ def represent_numpy_floating(self, data):
...
-            default_style=None, default_flow_style=False,
-            canonical=None, indent=None, width=None,
-            allow_unicode=None, line_break=None,
-            encoding=None, explicit_start=None, explicit_end=None,
-            version=None, tags=None):
+                 default_style=None, default_flow_style=False,
+                 canonical=None, indent=None, width=None,
+                 allow_unicode=None, line_break=None,
+                 encoding=None, explicit_start=None, explicit_end=None,
+                 version=None, tags=None, **kwargs):
```
pieleric, line 509: Let's add a comment that it's supposed to catch "sort_keys" on newer versions of PyYaml.

PR #2377, opened 2023-03-22
fastem: fix the text size of project and ROA names on ubuntu 20, with…

tepals:
… the default size the underscores where not visible

### DIFF
--- a/src/odemis/gui/comp/stream.py
+++ b/src/odemis/gui/comp/stream.py
@@ -1651,7 +1651,7 @@ class FastEMProjectPanelHeader(wx.Control):
     def _add_text_ctrl(self, name):
         """ Add a label control to the header panel """
         txt_ctrl = wx.TextCtrl(self, wx.ID_ANY, name, style=wx.TE_PROCESS_ENTER | wx.BORDER_NONE,
-                               validator=PatternValidator(r'[A-Za-z0-9_()-]+'), size=(35, 35))
+                               validator=PatternValidator(r'[A-Za-z0-9_()-]+'), size=(-1, 35))
         txt_ctrl.SetBackgroundColour(self.Parent.GetBackgroundColour())
         txt_ctrl.SetForegroundColour(FG_COLOUR_MAIN)
         self._add_ctrl(txt_ctrl, stretch=True)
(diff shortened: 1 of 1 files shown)


tepals:
For example see: https://delmic.atlassian.net/browse/SDEV-1319

src/odemis/acq/test/fastem_test.py
```diff
@@ -44,7 +44,7 @@
 # * TEST_NOHW = 1: connected to the simulator or not connected to anything
 # * TEST_NOHW = 0: connected to the real hardware, the backend should be running
 # technolution_asm_simulator/simulator2/run_the_simulator.sh
-TEST_NOHW = (os.environ.get("TEST_NOHW", "0") != "0")  # Default is hardware testing
+TEST_NOHW = 1  #(os.environ.get("TEST_NOHW", "0") != "0")  # Default is hardware testing
```
pieleric, line 47: That shouldn't be here.

src/odemis/acq/test/fastem_test.py
```diff
@@ -439,7 +439,7 @@ def setUpClass(cls):
         cls.multibeam = model.getComponent(role="multibeam")
         cls.descanner = model.getComponent(role="descanner")
         cls.stage = model.getComponent(
-            role="stage")  # TODO replace with stage-scan when ROA conversion method available
+            role="stage-scan")  # TODO replace with stage-scan when ROA conversion method available
```
pieleric, line 442: That change also shouldn't be here. Maybe it's a useful change too though... and you'd need to drop the TODO.

pieleric (changes requested):

src/odemis/acq/test/fastem_test.py
tepals: I accidentally committed this,  unfortunately it's a bit more work to fix this, it will also need to be replaced and tested in the acquisition code.

src/odemis/gui/comp/stream.py
```diff
@@ -1651,7 +1651,7 @@ def _add_remove_btn(self):
     def _add_text_ctrl(self, name):
         """ Add a label control to the header panel """
         txt_ctrl = wx.TextCtrl(self, wx.ID_ANY, name, style=wx.TE_PROCESS_ENTER | wx.BORDER_NONE,
-                               validator=PatternValidator(r'[A-Za-z0-9_()-]+'))
+                               validator=PatternValidator(r'[A-Za-z0-9_()-]+'), size=(35, 35))
```
pieleric, line 1654: You're only interested in increasing the vertical size, right? If so, I think that the first 35 (x) is a red herring. It should be possible to replace it by `-1`, which indicates to wxPython that the current/default value shouldn't be changed.

pieleric (changes requested):

PR #2387, opened 2023-03-30
add yaml loader tests

FCanberk:


src/odemis/acq/millmng.py
```diff
@@ -325,7 +352,7 @@ def _mill_all_settings(self, site: CryoFeature):
             # X is horizontal axis, Y is vertical axis
             pxs = milling_settings.pixelSize.value
             probe_size = max(pxs)
-            overlap = [1 - (pxs[0] / probe_size), 1 - (pxs[1] / probe_size)]  # Always >= 0
+            overlap = [1 - (float(pxs[0]) / float(probe_size)), 1 - (float(pxs[1]) / float(probe_size))]  # Always >= 0
```
pieleric, line 355: Please drop, unless you have a really good reason to force a float. I think you've probably added it because some of the "float" YAML values were detected by YAML as string.

pieleric (changes requested):

pieleric (approved):

PR #2390, opened 2023-04-06
Fixed a NoneType error in overlay selection base.py and world.py

sneepstefan:
Fixed an NoneType error in src > odemis > gui > comp > overlay > base.py and world.py. If both of the values returned by get_width and get_height is None the max() method in stop_selection will raise a NoneType error.

Now an int() conversion is applied with a default int value of 0.

### DIFF
--- a/src/odemis/gui/comp/overlay/base.py
+++ b/src/odemis/gui/comp/overlay/base.py
@@ -620,7 +620,7 @@ class SelectionMixin(DragMixin):
 
         logging.debug("Stopping selection")
 
-        if max(self.get_height(), self.get_width()) < gui.SELECTION_MINIMUM:
+        if max(self.get_height() or 0, self.get_width() or 0) < gui.SELECTION_MINIMUM:
             logging.debug("Selection too small")
             self.clear_selection()
         else:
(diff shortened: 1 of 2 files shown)


src/odemis/gui/comp/overlay/base.py
```diff
@@ -620,7 +620,7 @@ def stop_selection(self):
 
         logging.debug("Stopping selection")
 
-        if max(self.get_height(), self.get_width()) < gui.SELECTION_MINIMUM:
+        if max(int(self.get_height() or 0), int(self.get_width() or 0)) < gui.SELECTION_MINIMUM:
```
tepals, line 623: I don't think you need the `int()`, the following should also work. And can you add a comment explaining that you need the `or zero` in the height or width is None

```suggestion
        if max(self.get_height() or 0, self.get_width() or 0) < gui.SELECTION_MINIMUM:
```

FCanberk (approved):

tepals (approved):

pieleric:
BTW, it's a good practice to explain in the commit log how to reproduce the error that you fixed, and what the error message looks like. 
This helps to see if the bug is fixed when someone else encounters the same issue.
In this case, I have no idea how much I should prioritise this fix. Should I back port it to the stable version, or is it triggered only by new code? When this happens, what's the consequence? Does it prevent the user to do some needed action, or is that just avoiding logging an error during a corner case?

PR #2391, opened 2023-04-11
spot-grid script: when the beam is off or blanked there are no spots …

tepals:
…on the diagnostic camera image, this caused the script to become un-responsive.

SDEV-1334 You would need to restart the script after turning on the beam to see the live display update. This fix ensures that the script stays responsive and there is no need to restart the script.

### DIFF
--- a/scripts/spot-grid.py
+++ b/scripts/spot-grid.py
@@ -59,27 +59,32 @@ class VideoDisplayerGrid(VideoDisplayer):
         at ratio 1:1)
         data (numpy.ndarray): an 2D array containing the image (can be 3D if in RGB)
         """
-        tform_ji, _ = estimate_grid_orientation_from_img(
-            data,
-            self.gridsize,
-            AffineTransform,
-            sigma=1.45,
-            threshold_rel=0.5,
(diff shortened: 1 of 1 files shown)


nandishjpatel (approved):

pieleric (approved):

nandishjpatel (approved):

pieleric (commented):
@effting, it'd be quite helpful if the functions in odemis.util.registration, such as `estimate_grid_orientation_from_img()`, had a description of which exception is raised in which case.

For now, it's quite unreliable to say "I've run the code with various weird images and I think I've seen all the different type of exceptions that it can raise".

pieleric:
@tepals , I think you should be able to drop the IndexError handling now that PR #2396 is merged.

effting:
> @tepals , I think you should be able to drop the IndexError handling now that PR #2396 is merged.

indeed, that IndexError was a bug that's now properly handled. There should be no need to catch an IndexError here (and if there is one, it is another bug)

pieleric (approved):

PR #2396, opened 2023-04-14
Bugfix of estimate_grid_orientation

effting:
`estimate_grid_orientation` and `estimate_grid_orientation_from_img` no longer raise an IndexError when providing (an image of) a single point as input

### DIFF
--- a/src/odemis/util/graph.py
+++ b/src/odemis/util/graph.py
@@ -172,6 +172,27 @@ class GraphBase(UserList, metaclass=ABCMeta):
                     continue
                 yield (vertex, neighbor)
 
+    def size(self, directed: Optional[bool] = None) -> int:
+        """
+        Return the number of edges in the graph.
+
+        Parameters
+        ----------
+        directed : bool, optional
+            If `True` the edges `(j, i)` and `(i, j)` will be considered as
+            being two separate edges. If `False`, these are counted as a single
+            edge. If not specified, uses `self.directed`.
+
+        Returns
+        -------
+        size : int
+            The number of edges in the graph.
+
+        """
+        counter = itertools.count()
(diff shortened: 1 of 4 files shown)


src/odemis/util/graph.py
```diff
@@ -172,6 +172,27 @@ def iter_edges(self, directed: Optional[bool] = None) -> Iterator[Edge]:
...
+
+        Returns
+        -------
+        size : int
+            The number of edges in the graph.
+
+        """
+        counter = itertools.count()
+        collections.deque(zip(self.iter_edges(directed), counter), maxlen=0)
+        return next(counter)
```
pieleric, line 194: I think it's simpler to write:
```suggestion
        return sum(1 for _ in self.iter_edges(directed))
```

(and I think I found the stackoverflow article where it stated the deque option was a tiny bit faster, but that's not even correct for Python 3 anymore. `sum()` is now both simpler to read and faster)

src/odemis/util/registration.py
```diff
@@ -358,6 +363,13 @@ def _cluster_edges(
         matched centroids.
 
     """
+    n_edges = graph.size(False)
```
pieleric, line 366: Let's make it more explicit:
```suggestion
    n_edges = graph.size(directed=False)
```

pieleric (changes requested):

src/odemis/util/graph.py
tepals: I agree that Érics suggestion is a lot more readable

effting: done

src/odemis/util/registration.py
effting: done

pieleric (approved):

tepals (approved):

PR #2400, opened 2023-04-17
util spectrum: make get_angle_per_pixel() really able to fall back to recreating the metadata

pieleric:
If MD_THETA_LIST is not present would try to recreate the metadata...
but it would typically generate the wrong shape as we pass data of CAZYX
while the function expected AC.

Also detect that MD_THETA_LIST with only NaNs is wrong, so that the rest
of Odemis can fall-back to show just pixel values.

Also add test cases!

### DIFF
--- a/src/odemis/util/spectrum.py
+++ b/src/odemis/util/spectrum.py
@@ -259,7 +259,7 @@ def get_angle_per_pixel(da):
         # Try to build the MD_THETA_LIST from the mirror information metadata
         # ExtractThetaList() needs the shape as AC, so need to
         # 1. remove extra dimension (by picking just one point)
-        # 2, reorder the dimensions
+        # 2. reorder the dimensions
         try:
             ci = dims.index("C")  # get index of dimension C
         except ValueError:
--- a/src/odemis/util/spectrum.py
+++ b/src/odemis/util/spectrum.py
@@ -26,10 +26,11 @@ Odemis. If not, see http://www.gnu.org/licenses/.
 # Note that the spectrum is normally contained on the C dimension, which is
 # by convention the first of the 5 dimensions of a DataArray (CTZYX).
 
-from builtins import range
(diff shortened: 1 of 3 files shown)


src/odemis/util/spectrum.py
```diff
@@ -234,29 +235,53 @@ def get_angle_per_pixel(da):
...
+        # Sanity check: it should contain at least *some* angles (if the computation went wrong, it may only return NaN's)
+        if not [v for v in thetal if math.isfinite(v)]:
+            logging.warning("Discarding MD_THETA_LIST as it doesn't contain any proper angle")
+            thetal = None
+
+    if thetal is None:
+        # Try to build the MD_THETA_LIST from the mirror information metadata
+        # ExtractThetaList() needs the shape as AC, so need to
+        # 1. remove extra dimension (by picking just one point)
+        # 2, reorder the dimensions
```
nandishjpatel, line 262:
```suggestion
        # 2. reorder the dimensions
```

src/odemis/util/test/spectrum_test.py
```diff
@@ -49,7 +50,82 @@ def test_wl_list(self):
...
+
+class TestGetAngle(unittest.TestCase):
+
+    def test_angle_per_pixel_simple(self):
+        """
+        Test get_angle_per_pixel when all metadata is good
+        """
+        shape = (512, 256, 1, 5, 4)
+        dtype = numpy.dtype("uint16")
+        wl_orig = (400e-9 + numpy.arange(shape[0]) * 10e-9).tolist()
```
nandishjpatel, line 63: These three list can go to the `setUpClass()` if needed as they are common for the test-cases. Just a suggestion.

nandishjpatel (approved):

sneepstefan (approved):

PR #2405, opened 2023-05-01
Overview acquisition with Autofocus tested on hardware

K4rishma:
The autofocus algorithm is integrated with overview acquisition. When overview acquisition button is pressed in Odemis, first the autofocus algorithm works on a set number of focus points and afterwards the tiled acquisition works. 

The implementation for scaling down the number of focus points as per the given area of acquisition is also included. 

The following tasks are remaining: 
1. The progress bar shows underestimated time of the overview acquisition. For e.g. the progress bar shows "few seconds remaining" when tiled acquisition is beginning to acquire second row.
2. In win/acquisition.py line 1240, the z levels calculation is not placed needs to be changed at better location in the repository.


tepals (changes requested):

src/odemis/util/linalg.py
```diff
@@ -230,3 +230,74 @@ def are_collinear(p1: Iterable[float], p2: Iterable[float], p3: Iterable[float])
     return abs((x2 - x1) * (y3 - y1) - (x3 - x1) * (y2 - y1)) < 1e-12 and \
         abs((x2 - x1) * (z3 - z1) - (x3 - x1) * (z2 - z1)) < 1e-12 and \
         abs((y2 - y1) * (z3 - z1) - (y3 - y1) * (z2 - z1)) < 1e-12
+
+
+def find_focus_points(maximum_area: float, given_area_coords: Iterable[float]) -> Iterable[tuple]:
```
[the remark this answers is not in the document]
pieleric: With such suggested name, this function sounds like making sense in this file (linear algebra). (Alternatively, we could move this function to somewhere in acq.align.roi_autofocus.)

pieleric (changes requested):

src/odemis/acq/align/test/roi_autofocus_test.py
```diff
@@ -135,7 +140,7 @@ def test_estimate_autofocus_in_roi(self):
         """
         Tests time estimation of autofocus in roi
         """
-        n_focus_points = (3, 3)
-        min_time = n_focus_points[0] * n_focus_points[1] * estimateAutoFocusTime(self.ccd, None)
+        n_focus_points = 9
```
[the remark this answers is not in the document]
K4rishma: The focus points vary based on the selected roi by the user. It can vary from 4 to 16 focus points.  To have fast execution of the test cases, I think 9 or less than that should be sufficient for the testing.

src/odemis/acq/align/autofocus.py
```diff
@@ -348,7 +348,7 @@ def _DoBinaryFocus(future, detector, emt, focus, dfbkg, good_focus, rng_focus):
             good_focus = focus.position.value["z"]
             image = AcquireNoBackground(detector, dfbkg, timeout)
             fm_good = Measure(image)
-            logging.debug("Focus level at %.7g is %.7g", good_focus, fm_good)
+            logging.debug("Focus level at %.7g is %.7g, good focus known", good_focus, fm_good)
```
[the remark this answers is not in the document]
K4rishma: I will reword the sentence. The logging prints the good focus value already known to the function before starting binary focus algorithm.

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -1045,13 +1059,18 @@ def estimate_time(self, roi_idx=0, actual_time_per_roi=None) -> float:
         :return: (float) the estimated time for the rest of the acquisition
         """
         remaining_rois = (len(self.areas) - roi_idx)
-
+        # TODO progress bar does not calculate proper time, it underestimates the time
```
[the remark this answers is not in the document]
K4rishma: There is always time under estimation. I need to fix the time estimation.

src/odemis/util/linalg.py
```diff
@@ -230,3 +230,74 @@ def are_collinear(p1: Iterable[float], p2: Iterable[float], p3: Iterable[float])
...
+        y_points = [ymin, ymin, ymax, ymax, (ymax + ymin) / 2]
+    else:
+        num_points = 4
+        x_points = [xmin, xmax, xmin, xmax]
+        y_points = [ymin, ymin, ymax, ymax]
+
+    # Distribute the points
+    points = []
+    for i in range(num_points):
+        points.append((x_points[i], y_points[i]))
```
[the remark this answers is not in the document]
K4rishma: In order to have tuple of (x,y) coordinates from separate x and y points arrays.

src/odemis/util/linalg.py
```diff
@@ -230,3 +230,56 @@ def are_collinear(p1: Iterable[float], p2: Iterable[float], p3: Iterable[float])
...
+    """
+    # TODO Calculate points distribution for rectangular region where one side >> second side
+    # in the above case, points along the shorter side will be closer to each other and
+    # will not be evenly distributed throughout the given area
+    xmin, ymin, xmax, ymax = given_area_coords
+
+    # Avoid points exactly on the border of the given area, find points delta distance
+    # away from the border of the given area
+    delta_x = (1 / 20) * (xmax - xmin)
+    delta_y = (1 / 20) * (ymax - ymin)
```
pieleric, line 255: 1/20th is a little bit arbitrary. What I'd suggest is to make it so that the delta is about (but a little bit smaller than) max_dist. Essentially, that consists in first computing the number of points needed for the whole area, and then removing the distance needed for 1 point. One advantage in addition to being less arbitrary is that it's exactly this algorithm that we use in Odemis for the placement of points in a grid in general (like for the AR positions on the SPARC, or the SEM beam positions in an RoI).
```
points_x = math.floor(length_x/max_dist) 
step_x = length_x / points_x
xmin = xmin + step_x/2
xmax = xmax - step_x/2
```
(the y dimension is the same)

pieleric (changes requested):

src/odemis/util/test/linalg_test.py
```diff
@@ -127,6 +128,36 @@ def test_are_collinear(self):
...
+        length_1_start = 0
+        length_1_end = 1.77e-03
+        length_2_start = 0
+        length_2_end = 1.77e-03
+
+        # given area is 100% of maximum area (1.77mm x 1.77mm)
+        given_area_coords = (length_1_start, length_2_start, length_1_end, length_2_end)
+        focus_points = generate_triangulation_points(max_dis, given_area_coords)
+        self.assertEqual(len(focus_points), 16)
+        self.assertEqual(len(focus_points[0]), 2)  # (x,y)
```
pieleric, line 145: You could try more explicit and easy checks for the things that are claimed in the docstring of the function. In particular:
* every focus point is within the the given area.
* for any point in the area, there is always a focus point which is closer than the max distance. (you can check the smallest distance to point p with `min(math.hypot(f[0]-p[0], f[1]-p[1] for f in focus_points))`). You could just hard-code a couple of "typical points" to test.

src/odemis/util/test/linalg_test.py
```diff
@@ -127,6 +128,36 @@ def test_are_collinear(self):
...
+
+        # given area is 90 % maximum area
```
pieleric, line 147: Make it a separate test case/function.

src/odemis/util/test/linalg_test.py
```diff
@@ -127,6 +128,36 @@ def test_are_collinear(self):
...
+        red_length = 1.77e-03*numpy.sqrt(0.9)
```
pieleric, line 148: I don't know why it's called "red". It sounds like a colour, but I don't get why you'd use a colour.
I'd suggest to just explicitly reduce one (or 2 dimensions) to 0 -> 3.01e-6, and indicate as a comment that it should always round up to 4x4.

BTW, I'd recommend to *not* use square area in general in the tests. Because if your code somehow inverses x and y, or has a copy-paste error where x is used twice, you'll not notice it. => Use an area with 4x5 points for instance.

src/odemis/gui/win/acquisition.py
```diff
@@ -1233,14 +1244,19 @@ def on_acquire(self, evt):
...
+        zlevels, focus_ref = self._get_zstack_levels()
         focus_mtd = FocusingMethod.MAX_INTENSITY_PROJECTION if zlevels else FocusingMethod.NONE
 
         if self.filename_tiles:
             logging.info("Acquisition tiles logged at %s", self.filename_tiles)
             os.makedirs(os.path.dirname(self.filename_tiles), exist_ok=True)
 
         if self.autofocus_roi_ckbox.value:
+            areas = self._get_areas()
+            # calculate relative range of z levels
```
pieleric, line 1256: Yes. That's quite clear, from the code. What you should explain in this comment is **why** do you want the relative position (while you just had computed the absolute position).

src/odemis/gui/win/acquisition.py
```diff
@@ -1187,23 +1191,30 @@ def _get_zstack_levels(self):
...
+            shift = focus_range[0] - zmin
+            zmin += shift
+            zmax += shift
 
         # Create focus zlevels from the given zsteps number
         zlevels = numpy.linspace(zmin, zmax, zsteps).tolist()
-        return zlevels
+        return zlevels, focus_value
+
+    def _fit_view_to_area(self):
```
pieleric, line 1208: How does this work? line 1232 calls this method with an argument. There is no `.area` on `self`.
I think maybe it's a incorrect revert from my own changes to this class?

pieleric (changes requested):

src/odemis/util/linalg.py
```diff
@@ -230,3 +230,41 @@ def are_collinear(p1: Iterable[float], p2: Iterable[float], p3: Iterable[float])
...
+    Finds (x,y) positions on the given area. The points are evenly distributed from each other in the given area.
+    The distance between points is decided by maximum distance.
+
+    :param max_dist: the maximum distance allowed between two (x,y) positions.
+    :param given_area_coords: [xmin, ymin, xmax, ymax] the top right and bottom left (x,y) coordinates in meters.
+    :return: List of (x,y) coordinates in the given area.
+    """
+    # TODO Calculate points distribution for rectangular region where one side >> second side
+    # in the above case, points along the shorter side will be closer to each other and
+    # will not be evenly distributed throughout the given area
```
pieleric, line 246: That TODO probably can go away now... or is it still valid?

src/odemis/util/linalg.py
```diff
@@ -230,3 +230,41 @@ def are_collinear(p1: Iterable[float], p2: Iterable[float], p3: Iterable[float])
...
+    points_y = numpy.floor(length_y/max_dist) + 1
+    xmin = xmin + max_dist / 2
+    xmax = xmax - max_dist / 2
+    ymin = ymin + max_dist / 2
+    ymax = ymax - max_dist / 2
+    total_points = int(points_x * points_y)
+
+    # Create a centre point when total number of points is <= 3
+    if total_points <= 3:
+        points = [((xmax + xmin)/2, (ymax + ymin)/2)]
```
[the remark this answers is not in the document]
K4rishma: Sorry I should have been careful

src/odemis/util/test/linalg_test.py
```diff
@@ -127,6 +128,36 @@ def test_are_collinear(self):
         non_linear_points = numpy.array([[1, 2, 12], [-1, -2, -4], [3, 4, 22]])
         self.assertFalse(are_collinear(non_linear_points[0], non_linear_points[1], non_linear_points[2]))
 
+    def test_generate_triangulation_points(self):
+        """
+        Test the number of focus points when given area changes.
+        """
+        max_dis = 450e-06
+        length_1_start = 0
+        length_1_end = 1.77e-03
```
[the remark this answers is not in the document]
pieleric: I understand where these values come from. Still for unit tests, I think it's better to use simple values, which are easy to compute mentally.
If you want, you could also add another test with these values, explicitly saying it's a typical values, and we *know* that we should have 4x4 points.

src/odemis/gui/win/acquisition.py
pieleric: Copy-paste the function as it is in the master. It will "magically" disappear from the PR.

src/odemis/util/test/linalg_test.py
K4rishma: I used it as a short for reduced length that became red_length. I will explicitly mention the values and explain in the comments for better transparency.

src/odemis/acq/align/roi_autofocus.py
```diff
@@ -53,34 +53,34 @@ def do_autofocus_in_roi(
...
-                foc_pos, foc_lev, conf = f._running_subf.result(timeout=900)
-                if conf >= conf_level:
-                    focus_positions.append([stage.position.value["x"],
-                                            stage.position.value["y"],
-                                            focus.position.value["z"]])
-                    logging.debug(f"Added focus with confidence of {conf} at position: {focus_positions[-1]}")
-                else:
-                    logging.debug(f"Focus level is not added due to low confidence of {conf} at position: {x, y}.")
+
+        for points in focus_points:
```
pieleric, line 65: "points" is odd with a "s" as it should contain a single point. Actually, I'd suggest to just directly write x & y, like that: `for (x, y) in focus_points:`
You can then drop lines 69 & 70.

pieleric (approved):

src/odemis/util/linalg.py
```diff
@@ -230,3 +230,38 @@ def are_collinear(p1: Iterable[float], p2: Iterable[float], p3: Iterable[float])
...
+    :return: List of (x,y) coordinates in the given area.
+    """
+    xmin, ymin, xmax, ymax = area_coords
+
+    # Avoid points exactly on the border of the given area, find points delta distance
+    # away from the border of the given area
+    length_x = abs(xmax - xmin)
+    length_y = abs(ymax - ymin)
+    points_x = numpy.floor(length_x/max_dist) + 1
+    points_y = numpy.floor(length_y/max_dist) + 1
```
tepals, line 251:
```suggestion
    points_x = numpy.floor(length_x / max_dist) + 1
    points_y = numpy.floor(length_y / max_dist) + 1
```

src/odemis/util/linalg.py
```diff
@@ -230,3 +230,38 @@ def are_collinear(p1: Iterable[float], p2: Iterable[float], p3: Iterable[float])
...
+    xmin = xmin + max_dist / 2
```
tepals, line 252: You are now using a max distance of 450µm, I think a tile in the overview image for meteor is around 80µm. This means that your first focus point is 3 tiles away from the border, is that not a bit much?

src/odemis/util/test/linalg_test.py
```diff
@@ -128,5 +129,82 @@ def test_are_collinear(self):
...
+        # 4x4 focus points when given area is a square
+        max_dis = 1e-06
+        area_coords = (0,  0, 3.9e-06,  3.9e-06)
+        focus_points = generate_triangulation_points(max_dis, area_coords)
+        x_points = [i[0] for i in focus_points]
+        y_points = [i[1] for i in focus_points]
+        xmin = min(x_points)
+        ymin = min(y_points)
+        xmax = max(x_points)
+        ymax = max(y_points)
```
tepals, line 196: Another way of doing this is:

```suggestion
        xmin, ymin = numpy.min(focus_points, axis=0)
        xmax, ymax = numpy.max(focus_points, axis=0)
```

pick whichever one you like best

tepals (changes requested):

src/odemis/util/linalg.py
```diff
@@ -230,3 +230,38 @@ def are_collinear(p1: Iterable[float], p2: Iterable[float], p3: Iterable[float])
...
         abs((x2 - x1) * (z3 - z1) - (x3 - x1) * (z2 - z1)) < 1e-12 and \
         abs((y2 - y1) * (z3 - z1) - (y3 - y1) * (z2 - z1)) < 1e-12
+
+
+def generate_triangulation_points(max_dist: float, area_coords: Iterable[float]) -> Iterable[Tuple[float]]:
+    """
+    Finds (x,y) positions on the given area. The points are evenly distributed from each other in the given area.
+    The distance between points is decided by maximum distance.
+
+    :param max_dist: the maximum distance allowed between two (x,y) positions.
```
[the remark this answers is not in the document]
K4rishma: With the comment on line 1041 of _tiledacq.py, I would like to convey that `MAX_DISTANCE_FOCUS_POINTS ` controls the separation between two adjacent focus points. When `generate_triangulation_points` function is used with this input parameter(MAX_DISTANCE_ ...), the function makes sure that the distance calculated between the adjacent focus points is less than or equal to the max_dist.

I hope I was able to clarify your doubt. With the slight changes in the comment, will the following clarify the usage of `MAX_DISTANCE_FOCUS_POINTS `
```
# Maximum distance used to separate two focus points in overview acquisition using autofocus
# Increasing this distance may result in out of focus overview acquisition image
```

src/odemis/util/linalg.py
K4rishma: 450µm is the maximum distance between two adjacent focus points. With the overview acquisition hardware testing, the focus points separated by this distance could provide focused image. This idea was then extended to set distance from the border. If x is the max_distance between two focus points then x/2 would be the focus range of a single focus point.

src/odemis/acq/align/roi_autofocus.py
```diff
@@ -135,8 +134,9 @@ def autofocus_in_roi(
         ccd: model.Component,
         focus: model.Component,
         focus_range: tuple,
-        n_focus_points: tuple = (3, 3),
+        focus_points: Iterable[tuple],
         conf_level: float = 0
+
```
FCanberk, line 139: remove the empty line

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -33,7 +33,7 @@
 
 from odemis import model, dataio
 from odemis.acq import acqmng
-from odemis.acq.align.autofocus import MeasureOpticalFocus, AutoFocus, MTD_EXHAUSTIVE
+from odemis.acq.align.autofocus import MeasureOpticalFocus, AutoFocus, MTD_EXHAUSTIVE, MeasureSpotsFocus
```
FCanberk, line 36:
```suggestion
from odemis.acq.align.autofocus import AutoFocus, MeasureOpticalFocus, MeasureSpotsFocus, MTD_EXHAUSTIVE
```

src/odemis/util/linalg.py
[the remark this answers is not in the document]
K4rishma: @tepals updated the docstring

src/odemis/acq/align/roi_autofocus.py
```diff
@@ -160,5 +160,5 @@ def autofocus_in_roi(
     f.task_canceller = _cancel_autofocus_bbox
 
     executeAsyncTask(f, do_autofocus_in_roi,
-                     args=(f, bbox, stage, ccd, focus, focus_range, n_focus_points[0], n_focus_points[1], conf_level))
-    return f
+                     args=(f, bbox, stage, ccd, focus, focus_range, focus_points, conf_level))
+    return f
```
tepals, line 164: Can you put the empty line back at the end of this file?

src/odemis/util/linalg.py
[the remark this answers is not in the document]
tepals: looks good

tepals (changes requested):
one last small comment

tepals (approved):

FCanberk (approved):

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -491,11 +504,14 @@ def estimateTime(self, remaining=None):
 
         acq_time = 0
         for stream in self._streams:
-            acq_stream_time = acqmng.estimateTime([stream])
+            # add 1s to account for stage/objective movement time in z direction
+            acq_stream_time = acqmng.estimateTime([stream]) + 1
```
nandishjpatel, line 508: @K4rishma I think this line is wrong. I think it should be as below. What do you think?

```
        acq_time = 0
        for stream in self._streams:
            acq_stream_time = acqmng.estimateTime([stream])
            if stream.focuser is not None and len(self._zlevels) > 1:
                # Acquisition time for each stream will be multiplied by the number of zstack levels
                zlevels = [item for item in self._zlevels if item is not None]
                # add 1s to account for stage/objective movement time in z direction
                acq_stream_time = (acq_stream_time + 1) * len(zlevels)
            # add 2 seconds to account for switching from one tile to next tile
            acq_time += acq_stream_time + 2
```

PR #2408, opened 2023-05-03
Sdev 1373 implement calibration button changes

nandishjpatel:
Parent ticket: https://delmic.atlassian.net/browse/SDEV-1343
Child ticket: https://delmic.atlassian.net/browse/SDEV-1373

### DIFF
--- a/src/odemis/gui/cont/streams.py
+++ b/src/odemis/gui/cont/streams.py
@@ -215,10 +215,6 @@ class StreamController(object):
         if hasattr(stream, "repetition"):
             self._add_repetition_ctrl()
 
-        if tab_data_model.main.role == "mbsem" and isinstance(stream, acqstream.SEMStream):  # don't show for CCD stream
-            # It's a FastEM
-            self._add_fastem_ctrls()
-
         # Set the visibility button on the stream panel
         if view:
             vis = stream in view.stream_tree
@@ -1368,50 +1364,6 @@ class StreamController(object):
 
         # Make sure the current value is selected
         self._rep_ctrl.SetSelection(choices.index(rep))
-
-    def _add_fastem_ctrls(self):
-        self.stream_panel.add_divider()
-
-        # FIXME uncomment autostigmation, when the functionality is working correctly.
-        _, btn_autofocus = self.stream_panel.add_run_btn("Autofocus")
-        _, btn_autobc = self.stream_panel.add_run_btn("Auto-brightness/contrast")
-        # _, btn_autostigmation = self.stream_panel.add_run_btn("Autostigmation")
-
-        btn_autofocus.Bind(wx.EVT_BUTTON, self._on_btn_autofocus)
-        btn_autobc.Bind(wx.EVT_BUTTON, self._on_btn_autobc)
-        # btn_autostigmation.Bind(wx.EVT_BUTTON, self._on_btn_autostigmation)
-
-    @call_in_wx_main
-    def _on_btn_autofocus(self, _):
-        self.stream_panel.Enable(False)
-        self.pause()
-        self.pauseStream()
-        f = self.stream.focuser.applyAutofocus(self.stream.detector)
-        f.add_done_callback(self._on_autofunction_done)
-
-    @call_in_wx_main
-    def _on_btn_autobc(self, _):
-        self.stream_panel.Enable(False)
-        self.pause()
-        self.pauseStream()
-        f = self.stream.detector.applyAutoContrastBrightness()
-        f.add_done_callback(self._on_autofunction_done)
-
-    @call_in_wx_main
-    def _on_btn_autostigmation(self, _):
-        self.stream_panel.Enable(False)
-        self.pause()
-        self.pauseStream()
-        f = self.stream.emitter.applyAutoStigmator(self.stream.detector)
-        f.add_done_callback(self._on_autofunction_done)
-
-    @call_in_wx_main
-    def _on_autofunction_done(self, f):
-        self.stream_panel.Enable(True)
-        self.resume()
-        # Don't automatically resume stream, autofunctions can take a long time.
-        # The user might not be at the system after the functions complete, so the stream
-        # would play idly.
 
 
 class StreamBarController(object):
--- a/src/odemis/gui/cont/tabs.py
+++ b/src/odemis/gui/cont/tabs.py
@@ -51,6 +51,7 @@ from odemis.util.filename import guess_pattern, create_projectname
 
 from odemis import dataio
 from odemis import model
+from odemis.acq.align import fastem
 import odemis.acq.stream as acqstream
 import odemis.gui
 import odemis.gui.cont.acquisition as acqcont
@@ -1932,9 +1933,9 @@ class FastEMOverviewTab(Tab):
         # Acquisition Controller
         #   Takes care of the acquisition and acquisition selection buttons.
 
-        tab_data = guimod.FastEMOverviewGUIData(main_data)
+        self.tab_data = guimod.FastEMOverviewGUIData(main_data)
 
-        super(FastEMOverviewTab, self).__init__(name, button, panel, main_frame, tab_data)
+        super(FastEMOverviewTab, self).__init__(name, button, panel, main_frame, self.tab_data)
         self.set_label("OVERVIEW")
 
         # Flag to indicate the tab has been fully initialized or not. Some initialisation
@@ -1952,7 +1953,7 @@ class FastEMOverviewTab(Tab):
               "stream_classes": EMStream,
               }),
         ])
-        self.view_controller = viewcont.ViewPortController(tab_data, panel, vpv)
+        self.view_controller = viewcont.ViewPortController(self.tab_data, panel, vpv)
 
         # Single-beam SEM stream
         vanames = ("resolution", "scale", "dwellTime", "horizontalFoV")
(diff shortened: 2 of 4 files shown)


src/odemis/gui/cont/tabs.py
```diff
@@ -2000,6 +2007,84 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
+                                   self.tab_data.main.descanner, self.tab_data.main.mppc,
+                                   self.tab_data.main.stage, self.tab_data.main.ccd,
+                                   self.tab_data.main.beamshift, self.tab_data.main.det_rotator,
+                                   calibrations=[Calibrations.OPTICAL_AUTOFOCUS])
+
+            f.add_done_callback(self._on_calibration_done)  # also handles cancelling and exceptions
+            # connect the future to the progress bar and its label
+            self._future_connector = ProgressiveFutureConnector(f, self.gauge, full=False)
+            self.btn_sem_autofocus.SetLabelText("Cancel")
+        elif label_text == "Cancel":
```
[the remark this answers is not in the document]
nandishjpatel: My bad, I make the changes

sneepstefan (changes requested):

src/odemis/gui/cont/tabs.py
```diff
@@ -2000,6 +2007,83 @@ def __init__(self, name, button, panel, main_frame, main_data):
         self.tb.add_tool(TOOL_ACT_ZOOM_FIT, self.view_controller.fitViewToContent)
         # TODO: add tool to zoom out (ie, all the scintillators, calling canvas.zoom_out())
 
+    def _on_btn_optical_autofocus(self, _):
+        """
+        Start or cancel the calibration when the button is triggered.
+        """
+        # from odemis.gui.comp.buttons import ImageTextButton
+        # a = ImageTextButton()
+        # a.SetLabelText
```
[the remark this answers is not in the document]
nandishjpatel: Indeed not needed

src/odemis/gui/cont/tabs.py
```diff
@@ -2000,6 +2007,83 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
+        label_text = self.btn_optical_autofocus.GetLabelText()
+        if label_text == "Run...":
```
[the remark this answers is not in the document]
nandishjpatel: Will name it to "Run". Since there might be a need to run the calibrations again, the workaround for the "status" is to change the text label for the calibration and indicate it there

src/odemis/gui/cont/tabs.py
```diff
@@ -1968,28 +1969,34 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
+        self.btn_autobc = panel.btn_autobrigtness_contrast
+
+        self.btn_optical_autofocus.Bind(wx.EVT_BUTTON, self._on_btn_optical_autofocus)
+        self.btn_sem_autofocus.Bind(wx.EVT_BUTTON, self._on_btn_sem_autofocus)
+        self.btn_autobc.Bind(wx.EVT_BUTTON, self._on_btn_autobc)
+        # self.btn_autostigmation.Bind(wx.EVT_BUTTON, self._on_btn_autostigmation)
+
+        # For SEM Autofocus calibration
+        self._future_connector = None  # attribute to store the ProgressiveFutureConnector
+        self.gauge = wx.Gauge()
```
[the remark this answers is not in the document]
nandishjpatel: Maybe its better to create the object and just pass `wx.Gauge()` [here](https://github.com/delmic/odemis/blob/f26af1d024799934a8adb59cf02c094b4d2602cc/src/odemis/gui/cont/tabs.py#L2029)

It's a workaround, as this calibration just needs a button. Else I would have created the gauge in the main.xrc

nandishjpatel:
I will combine 2 Jira tickets, since both are related to button changes and will be easier to review.

Second ticket: https://delmic.atlassian.net/browse/SDEV-1374 (Immersion mode button)

src/odemis/gui/comp/stream.py
```diff
@@ -813,6 +813,18 @@ def _add_side_label(self, label_text, tooltip=None):
                           flag=wx.ALL | wx.ALIGN_CENTER_VERTICAL, border=5)
         return lbl_ctrl
 
+    @control_bookkeeper
+    def add_immersion_mode_button(self):
+        """ Create and return immersion mode button."""
```
pieleric, line 818: I'm a little late at the party, but I wonder about this option: why a toggle button? In Odemis, typically booleans VAs are shown as checkboxes. Why not doing that here too? The rest of the buttons in the GUI are more for actions. This is just a "mode".

src/odemis/gui/cont/streams.py
```diff
@@ -1371,47 +1371,13 @@ def is_compatible(c):
...
-
-    @call_in_wx_main
-    def _on_autofunction_done(self, f):
-        self.stream_panel.Enable(True)
-        self.resume()
-        # Don't automatically resume stream, autofunctions can take a long time.
-        # The user might not be at the system after the functions complete, so the stream
-        # would play idly.
+        # Store a setting entry for the immersion mode button
+        se = SettingEntry(name="immersion_mode", va=self.stream.emitter.immersion,
```
pieleric, line 1378: So this is just for live stream? But as soon as the actual overview acquisition starts, the immersion mode is force on. Is that really how it should behave? Wouldn't it be less confusing for the user if that setting was used during the overview too?

src/odemis/gui/cont/tabs.py
```diff
@@ -2000,6 +2008,116 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
+        logging.debug("Starting Optical Autofocus calibration")
+        # Start alignment
+        f = fastem.align(self.tab_data.main.ebeam, self.tab_data.main.multibeam,
+                               self.tab_data.main.descanner, self.tab_data.main.mppc,
+                               self.tab_data.main.stage, self.tab_data.main.ccd,
+                               self.tab_data.main.beamshift, self.tab_data.main.det_rotator,
+                               calibrations=[Calibrations.OPTICAL_AUTOFOCUS])
+        f.add_done_callback(self._on_calibration_done)  # also handles cancelling and exceptions
+        # connect the future to the progress bar and its label
+        self._future_connector = ProgressiveFutureConnector(f, wx.Gauge(), full=False)
```
pieleric, line 2035: Where will this wx.Gauge() be shown? Typically it's created at init (via the main.xrc), and shown/hidden. Is that something you will implement in a separate PR? If so, please mark as a comment.

Also, the `full` only matters if you pass a `label`, so here it should be fine to drop.

pieleric (changes requested):

src/odemis/gui/comp/stream.py
nandishjpatel: I didn't know that typically boolean VAs are shown as checkboxes, @tepals do you agree with this suggestion?

src/odemis/gui/cont/streams.py
nandishjpatel: Good catch, I didn't know this. It is not the expected behavior. So then I should make is of a `VigilantAttributeConnector` right @pieleric ?

src/odemis/gui/cont/tabs.py
nandishjpatel: The `Optical Autofocus` calibration in the Overview tab does not need a guage to be shown as per the mock-up. I can create a gauge and then hide it if you say so @pieleric.

src/odemis/gui/comp/stream.py
tepals: sounds fine

src/odemis/gui/cont/streams.py
tepals: I'm not sure I completely understand both your comments, so here is the explanation of the desired behavior:
During live view we want the user to be able to switch between immersion and non-immersion mode, to have full control of the stream. During the overview imaging the immersion mode should always be set to False (this allows us to have a larger field of view). At the end of the overview imaging the immersion mode should be set back to whatever is selected in the GUI.

src/odemis/gui/comp/stream.py
pieleric: If you go this route, it might be easier to make it part of the stream, as a "emitter_va" (+ an update to the gui.conf.data.STREAM_SETTINGS_CONFIG). You can also do it the current way.

src/odemis/gui/cont/streams.py
pieleric: Ok, if it's the expected behaviour, that's at least good code-wise (I don't know if it's obvious to the user). FYI, acq.fastem.OverviewAcquisition automatically disables the immersion mode, via `fastem_conf.configure_scanner()`.

It also automatically enables the immersion mode back on afterwards. It might be something you will want to change (ex, store the current setting and restore it at the end of the overview... but maybe it's not very efficient if multiple scintillators are acquired at once).

sneepstefan (approved):

src/odemis/gui/cont/tabs.py
pieleric: The only point of a `ProgressiveFutureConnector` is to connect the gauge (and text) to the (Progressive) future. If you don't show a gauge, you don't need the connector at all.

tepals (changes requested):

src/odemis/gui/cont/tabs.py
```diff
@@ -2000,6 +2007,113 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
+        f = self.sem_stream_cont.stream.emitter.applyAutoStigmator(self.sem_stream_cont.stream.detector)
+        f.add_done_callback(self._on_autofunction_done)
+
+    @call_in_wx_main
+    def _on_autofunction_done(self, f):
+        self.sem_stream_cont.stream_panel.Enable(True)
+        self.sem_stream_cont.resume()
+        # Don't automatically resume stream, autofunctions can take a long time.
+        # The user might not be at the system after the functions complete, so the stream
+        # would play idly.
```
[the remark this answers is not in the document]
tepals: I just checked and `self.sem_stream_cont.resume()` does the following: `Resume SettingEntry related control updates`, can you add a comment  on line 2119, explaining that that line resumes the SettingEntry related control updates of the stream?
Then you can leave the comment on line 2120

tepals (approved):

pieleric (approved):

PR #2409, opened 2023-05-04
Add limit to the number of milling points per milling object

FCanberk:
Not having such limit was causing issues due to the limitation by Orsay Physics.
- A warning added when "advised" number of milling points exceeded.
- A value error is raised when maximum allowed number of points exceeded.


src/odemis/acq/orsay_milling.py
```diff
@@ -177,6 +179,17 @@ def do_milling(self):
             obj_to_mill.centerOfMassX = ((rect[2] + rect[0]) / 2 - 0.5) * fov
             obj_to_mill.centerOfMassY = (0.5 - (rect[3] + rect[1]) / 2) * fov
 
+            step_distance = probe_size * (1-overlap)
+            points_to_be_scanned = (obj_to_mill.primaryAxisX/step_distance + 1) * \
+                                   (obj_to_mill.secondaryAxisLength/step_distance + 1)
+
+            if points_to_be_scanned >= ADVISED_MAX_POINTS_TO_SCAN:
+                logging.warning(f"It is advised to keep objects under 1 million points."
+                                f"Currently points to be scanned: {points_to_be_scanned}")
```
tepals, line 188: the 1 million will be incorrect if you ever update `ADVISED_MAX_POINTS_TO_SCAN`

tepals (approved):

src/odemis/acq/orsay_milling.py
```diff
@@ -177,6 +180,17 @@ def do_milling(self):
             obj_to_mill.centerOfMassX = ((rect[2] + rect[0]) / 2 - 0.5) * fov
             obj_to_mill.centerOfMassY = (0.5 - (rect[3] + rect[1]) / 2) * fov
 
+            step_distance = probe_size * (1 - overlap)
```
pieleric, line 183: This is different for X & Y. `overlap` is a tuple.
```suggestion
            step_distance = (probe_size * (1 - overlap[0]),
                             probe_size * (1 - overlap[1]))
```
(And then update the line below.

pieleric (changes requested):
Please also add a test case which does a `assertRaises(ValueError)` when the function is called with a tiny probe size a large area to mill.

src/odemis/acq/orsay_milling.py
```diff
@@ -33,6 +33,9 @@
 from MillingPattern.Layer import Layer
 from MillingPattern.Procedure import Procedure
 
+# The advised limit and maximum points to scan are limitations set by Orsay Physics
+MAX_POINTS_TO_SCAN = 1024*1024*16
+ADVISED_MAX_POINTS_TO_SCAN = 1000000
```
pieleric, line 38: To make it clearer, you can write `1_000_000`.

src/odemis/acq/orsay_milling.py
```diff
@@ -33,6 +33,9 @@
 from MillingPattern.Layer import Layer
 from MillingPattern.Procedure import Procedure
 
+# The advised limit and maximum points to scan are limitations set by Orsay Physics
```
pieleric, line 36: Could you order the constants in the same way as in this sentence?

pieleric (approved):

PR #2410, opened 2023-05-04
Fix for parsing extend yaml files which consists of a single component.

sneepstefan:
Problem occurs when an !extend yaml file parsed contains just one component. If the component name is absent or incorrect the error that is generated is not concise. With this fix it will point out the file in which the error was generated.

This is an example of the error generated when the component name is absent without the fix:
`ERROR	main:97:	When instantiating file /home/stefan/development/mic-odm-yaml/muenster-sparc-3056/sparc2-muenster-swd-uv-pmt-sim.odm.yaml
AttributeError: 'str' object has no attribute 'get'
2023-05-04 14:56:55,307	ERROR	main:780:	Unexpected error at instantiation`

This is an example of the state error with the fix:
`ERROR	main:777:	Error while parsing file /home/stefan/development/mic-odm-yaml/muenster-sparc-3056/sparc2-muenster-swd-uv-pmt-sim.odm.yaml:
Parsing of file 'lens-swd-mirror.odm.yaml' using the '!extend' key failed with the error:
wrong or missing component name in file /home/stefan/development/mic-odm-yaml/muenster-sparc-3056/lens-swd-mirror.odm.yaml`

### DIFF
--- a/src/odemis/odemisd/modelgen.py
+++ b/src/odemis/odemisd/modelgen.py
@@ -120,6 +120,14 @@ class SafeLoader(yaml.SafeLoader):
                     with open(filename, 'r') as f:
                         try:
                             node_extension = SafeLoader(f).get_single_node()  # Get the data from the external file.
+
+                            # if the node is the only component in the file raise a ParseError
+                            # if the component name is absent or incorrect
+                            if node_extension.value[0][0].value in {"class", "role", "creator"}:
+                                error = f"wrong or missing component name in file {filename}"
+                                raise ParseError(
+                                    "Parsing of file '%s' using the '!extend' key failed with the error:\n%s"
+                                    % (key_node.value, error))
                         except yaml.parser.ParserError as error:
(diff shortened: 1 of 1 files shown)


pieleric (changes requested):
Let's try to not mix up the "microscope" model level with the YAML parser. In particular, I would prefer not to define exactly how `!extend` is used. This new code would forbid to use `!extend` for extending a dict with keys "class", "role" or "creator". I'd rather keep the possibility to use `!extend` in other ways in the future.

 I think we can check that each component is indeed a str -> dict. That would make the error message better. I have a tentative code that does this. For now it's not as strong as your code as it's not able to point back to the specific "extended" file, though.

sneepstefan:
Yeah indeed my fist idea was also to just check for str / dict combination but could not find a valid spot to check for that. So eventually I ended up putting a check after get_single_node().
Good that you have a better fix!

PR #2411, opened 2023-05-05
Sdev 1126 add contrast brightness sliders for overview acquisition

nandishjpatel:
Ticket: https://delmic.atlassian.net/browse/SDEV-1126

### DIFF
--- a/src/odemis/gui/cont/streams.py
+++ b/src/odemis/gui/cont/streams.py
@@ -346,7 +346,8 @@ class StreamController(object):
 
         # Process the hardware VAs first (emitter and detector hardware VAs are combined into one
         # attribute called 'hw_vas'
-        # sort the list of hardware VAs first in case it contains items that are not in the emitter and detector conf, they will be appended at the end
+        # sort the list of hardware VAs first in case it contains items that are not in the
+        # emitter and detector conf, they will be appended at the end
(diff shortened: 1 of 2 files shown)


src/odemis/gui/cont/streams.py
```diff
@@ -346,7 +346,8 @@ def _add_hw_setting_controls(self):
 
         # Process the hardware VAs first (emitter and detector hardware VAs are combined into one
         # attribute called 'hw_vas'
-        vas_names = util.sorted_according_to(list(self.stream.hw_vas.keys()),
+        # sort the la first in case lb does not contain an item from la, it will then be appended in the end
```
tepals, line 349: what are la and lb?

nandishjpatel: https://github.com/delmic/odemis/blob/f62b7d2dbcb1a8d7e1739a1b1c3326ec145163c7/src/odemis/util/__init__.py#L201

tepals: I would suggest to make it a bit more user readable
```suggestion
        # sort the list of hardware VAs first in case it contains items that are not in the emitter and detector conf, they will be appended at the end
```

sneepstefan (approved):

pieleric (approved):

PR #2412, opened 2023-05-08
odemisd: log a better error if component name is missing

pieleric:
With the "!extend" syntax it's relatively easy to forget the name of the
component (if using it to insert a whole component).
It would fail in very confusing way. This was noticed by Stefan Sneep.
=> Add a check to more clearly indicate the error.

Note that ideally we'd link back the error to a specific "extend" file,
however it's tricky because technically the YAML is correct, and by the
name we've loaded it, we have discarded the information.

This is an example of the error generated when the component name is absent without the fix:
```
2023-05-08 11:19:28,415 ERROR   main:97:        When instantiating file ../mic-odm-yaml/muenster-sparc-3056/sparc2-muenster-lwd-polarizer-sim.odm.yaml
  File "/home/piel/development/odemis/src/odemis/odemisd/main.py", line 88, in __init__
    self._instantiator = modelgen.Instantiator(model_file, settings_file, self,
  File "/home/piel/development/odemis/src/odemis/odemisd/modelgen.py", line 263, in __init__
    self._preparate_microscope()
  File "/home/piel/development/odemis/src/odemis/odemisd/modelgen.py", line 287, in _preparate_microscope
    microscopes = [(n, a) for n, a in self.ast.items() if a.get("class") == "Microscope"]
  File "/home/piel/development/odemis/src/odemis/odemisd/modelgen.py", line 287, in <listcomp>
    microscopes = [(n, a) for n, a in self.ast.items() if a.get("class") == "Microscope"]
AttributeError: 'str' object has no attribute 'get'
2023-05-08 11:19:28,416 ERROR   main:780:       Unexpected error at instantiation
```

This is an example with the fix:
```
2023-05-08 11:58:21,547 ERROR   main:777:       Error while parsing file ../mic-odm-yaml/muenster-sparc-3056/sparc2-muenster-lwd-polarizer-sim.odm.yaml:
Definition of component "class" should be a dict but got a str
```

Note that this also adds checks for a few other errors that would cause
odd error messages (eg, if a component name is not a string).

Also fixes the test cases to run from any directory.

### DIFF
--- a/src/odemis/odemisd/modelgen.py
+++ b/src/odemis/odemisd/modelgen.py
@@ -260,6 +260,7 @@ class Instantiator(object):
         self.create_sub_containers = create_sub_containers # flag for creating sub-containers
         self.dry_run = dry_run # flag for instantiating mock version of the components
 
+        self._check_structure()
         self._preparate_microscope()
 
         # update/fill up the model with implicit information
@@ -276,6 +277,30 @@ class Instantiator(object):
         # TODO: check here that each class is loadable.
 
         # TODO: check there is no cyclic dependencies on the parents/children
+
+    def _check_structure(self):
+        """
+        The AST should be a dict[str, dict[str, ...]]: name of component -> definition
+        Each component definition must have at least the "role" key
+        """
+        if not isinstance(self.ast, dict):
+            raise ParseError(f"Microscope file must be a dictionary of component names -> definition, "
+                             f"but got a {type(self.ast).__name__}")
+
+        for n, a in self.ast.items():
+            if not isinstance(n, str):
+                raise ParseError(f"Component name should be string but got: {n}")
+            if not isinstance(a, dict):
+                raise ParseError(f"Definition of component \"{n}\" should be a dict, "
+                                 f"but got a {type(a).__name__}")
+
+            # Check inside the component definition
+            for arg in a.keys():
+                if not isinstance(arg, str):
+                    raise ParseError(f"Definition of component \"{n}\" should only have keys of type str, "
+                                     f"but got a {type(arg).__name__}")
+            if "role" not in a:
+                raise SemanticError(f"Definition of component \"{n}\" is missing the mandatory key 'role'.")
 
     def _preparate_microscope(self):
         """
@@ -334,7 +359,7 @@ class Instantiator(object):
 
         # For each component which is created by delegation (= no class):
         # * if no creator specified, use its parent (and error if multiple parents)
-        # * if creator specified, check it's one of the parents
+        # * if creator specified, check it's one of the parents (deprecated, since the use of dependencies)
         for name, comp in self.ast.items():
             if "class" in comp:
                 continue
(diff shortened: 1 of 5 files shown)


nandishjpatel (approved):

src/odemis/odemisd/modelgen.py
```diff
@@ -277,6 +278,27 @@ def __init__(self, inst_file, settings_file=None, container=None, create_sub_con
...
+            if not isinstance(n, str):
+                raise ParseError(f"Component name should be string but got: {n}")
+            if not isinstance(a, dict):
+                raise ParseError(f"Definition of component \"{n}\" should be a dict but got a {type(a).__name__}")
+
+            # Check inside the component definition
+            for arg in a.keys():
+                if not isinstance(arg, str):
+                    raise ParseError(f"Definition of component \"{n}\" should only have keys of type str but got a {type(arg).__name__}")
+            if not "role" in a:
```
sneepstefan, line 299:
```suggestion
            if "role" not in a:
```
more towards PEP8 standard

src/odemis/odemisd/modelgen.py
```diff
@@ -277,6 +278,27 @@ def __init__(self, inst_file, settings_file=None, container=None, create_sub_con
 
         # TODO: check there is no cyclic dependencies on the parents/children
 
+    def _check_structure(self):
+        """
+        The AST should be a dict[str, dict[str, ...]]: name of component -> definition
+        Each component definition must have at least the "role" key
+        """
+        if not isinstance(self.ast, dict):
+            raise ParseError(f"Microscope file must be a dictionary of component names -> definition, but got a {type(self.ast).__name__}")
```
sneepstefan, line 287: Line is a bit too long
```suggestion
            raise ParseError(f"Microscope file must be a dictionary of component names -> definition, "
                             f"but got a {type(self.ast).__name__}")
```

src/odemis/odemisd/modelgen.py
```diff
@@ -277,6 +278,27 @@ def __init__(self, inst_file, settings_file=None, container=None, create_sub_con
...
+        for n, a in self.ast.items():
+            if not isinstance(n, str):
+                raise ParseError(f"Component name should be string but got: {n}")
+            if not isinstance(a, dict):
+                raise ParseError(f"Definition of component \"{n}\" should be a dict but got a {type(a).__name__}")
+
+            # Check inside the component definition
+            for arg in a.keys():
+                if not isinstance(arg, str):
+                    raise ParseError(f"Definition of component \"{n}\" should only have keys of type str but got a {type(arg).__name__}")
```
sneepstefan, line 298: Line a bit too long as well.
```suggestion
                    raise ParseError(f"Definition of component \"{n}\" should only have keys of type str "
                                     f"but got a {type(arg).__name__}")
```

src/odemis/odemisd/modelgen.py
```diff
@@ -277,6 +278,27 @@ def __init__(self, inst_file, settings_file=None, container=None, create_sub_con
...
+            if not "role" in a:
+                raise SemanticError(f"Definition of component \"{n}\" is missing a key 'role'.")
```
sneepstefan, line 300: You could point out that 'role' is a mandatory/required key.
```suggestion
                raise SemanticError(f"Definition of component \"{n}\" is missing the mandatory key 'role'.")
```

sneepstefan (changes requested):

PR #2417, opened 2023-05-15
added stage position legend

nandishjpatel:
Ticket: https://delmic.atlassian.net/browse/SDEV-1377

### DIFF
--- a/src/odemis/gui/comp/legend.py
+++ b/src/odemis/gui/comp/legend.py
@@ -125,6 +125,12 @@ class InfoLegend(wx.Panel):
         self.zPos_text.SetToolTip("Z Position")
         self.zPos_text.Hide()
 
+        # Stage position
+        self.stage_pos_text = wx.TextCtrl(self, style=wx.NO_BORDER | wx.CB_READONLY)
+        self.stage_pos_text.SetBackgroundColour(parent.GetBackgroundColour())
+        self.stage_pos_text.SetForegroundColour(parent.GetForegroundColour())
+        self.stage_pos_text.SetToolTip("Current Stage Position")
+
         # Feature show/hide button
         self._feature_toggle_va = None
         bmp = img.getBitmap("menu/btn_feature_toggle_off.png")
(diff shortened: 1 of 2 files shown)


pieleric (changes requested):
For now it adds a big extra info to an already crowded legend... for every viewport of the GUI. Even for the viewports which are not connected a real stage (such as in the analysis tab). I think for now it should be limited to the FASTEM GUI.

nandishjpatel:
> For now it adds a big extra info to an already crowded legend... for every viewport of the GUI. Even for the viewports which are not connected a real stage (such as in the analysis tab). I think for now it should be limited to the FASTEM GUI.

Should I add the stage position subscriber to FAST-EM classes [here](https://github.com/delmic/odemis/blob/aaab22a46d8a061266098765b7cbd1bc6e226172/src/odemis/gui/comp/viewport.py#L832-L884) instead of MicroscopeViewport?

src/odemis/gui/comp/legend.py
```diff
@@ -125,6 +125,12 @@ def __init__(self, parent, wid=-1, pos=(0, 0), size=wx.DefaultSize,
         self.zPos_text.SetToolTip("Z Position")
         self.zPos_text.Hide()
 
+        # Stage position
+        self.stage_pos_text = wx.TextCtrl(self, style=wx.NO_BORDER | wx.CB_READONLY)
+        self.stage_pos_text.SetBackgroundColour(parent.GetBackgroundColour())
+        self.stage_pos_text.SetForegroundColour(parent.GetForegroundColour())
+        self.stage_pos_text.SetToolTip("Current Stage Position")
+
```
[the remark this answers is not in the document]
nandishjpatel: Yes, I didn't know the consequences

pieleric (changes requested):

src/odemis/gui/comp/viewport.py
```diff
@@ -324,6 +324,11 @@ def setView(self, view, tab_data):
         if model.hasVA(tab_data, "zPos"):
             tab_data.zPos.subscribe(self._on_zPos_change, init=True)
 
+        # Listen to the stage position, if the view has the stage and the stage position TextCtrl
+        # has been set to be shown (by default it is hidden)
+        if hasattr(view, "stage_pos") and self.bottom_legend.stage_pos_text.IsShown():
+            view.stage_pos.subscribe(self._on_stage_pos_change, init=True)
+
```
pieleric, line 331: That feels reversed organization. Now the child class needs to make sure the stage_pos_text is shown, before calling the parent setView. Moreover the idea of `set_stage_pos_label()` is that it takes care of hidding/showing the internal widgets.
I'd suggest to entirely remove the code from this class. Then, in FastEM*Viewport classes, override `.setView()` to add:
```python
if hasattr(view, "stage_pos"):
        view.stage_pos.subscribe(self._on_stage_pos_change, init=True)
```
(the `if` might not be necessary, as normally there is always a stage on the FASTEM)

pieleric (changes requested):

src/odemis/gui/comp/viewport.py
nandishjpatel: Ahh, I misunderstood your previous comment

pieleric (approved):

PR #2420, opened 2023-05-16
gui cont: protect all callbacks of is_acquiring which use the GUI

pieleric:
All functions which are listening to is_acquiring and change the GUI
should be protected with @call_in_wx_main .
Otherwise, the GUI could crash at the end of an acquisition.

This might affect most of the microscope types, and has the potential to
solve some unresolved "crashes after acquisition".


tepals (approved):
I checked your code with the fastem simulator and it at least does not seem to break anything, so lets hope it fixes things!

nandishjpatel (approved):

PR #2425, opened 2023-05-22
driver smaract SmarPod: simulator should explicitly go to different position after referencing

pieleric:
It's not very realistic that after referencing the position is the same.
=> Go to some arbitrary position. That makes it easier to see in the GUI
if after the referencing some axes are left as-is.

### DIFF
--- a/src/odemis/driver/smaract.py
+++ b/src/odemis/driver/smaract.py
@@ -1090,7 +1090,7 @@ class FakeSmarPodDLL(object):
             self.referenced = False
             raise SmarPodError(SmarPodDLL.STOPPED_ERROR)
         else:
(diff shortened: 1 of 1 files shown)


src/odemis/driver/smaract.py
```diff
@@ -1090,6 +1090,13 @@ def Smarpod_FindReferenceMarks(self, id):
             self.referenced = False
             raise SmarPodError(SmarPodDLL.STOPPED_ERROR)
         else:
+            # Put some abritrary values representing the new position
```
sneepstefan, line 1093:
```suggestion
            # Put some arbitrary values representing the new position
```

nandishjpatel (approved):

PR #2440, opened 2023-06-07
Sdev 1457 implement overview tab changes based on feedback after testing on ea1

nandishjpatel:
Ticket: https://delmic.atlassian.net/browse/SDEV-1457

Changes:

- Have the Contrast slider above the Brightness slider (same setup as the TFS UI)

- Either display, “System not calibrated, please run Optical Autofocus“ or remove this warning and allow the user to acquire Overview Image without running Optical Autofocus

- Handle disabling of calibration buttons when necessary. Disable the calibration buttons when acquiring Overview Image, and when running each calibration disable the other calibration buttons

### DIFF
--- a/src/odemis/gui/cont/tabs.py
+++ b/src/odemis/gui/cont/tabs.py
@@ -2033,6 +2033,9 @@ class FastEMOverviewTab(Tab):
             fastem._executor.cancel()
             return
 
+        # Disable other calibration buttons
+        self.btn_sem_autofocus.Enable(False)
+        self.btn_autobc.Enable(False)
         # calibrate
         self.tab_data.is_calibrating.unsubscribe(self._on_is_acquiring)
         # Don't catch this event (is_calibrating = True) - this would disable the button,
@@ -2063,7 +2066,10 @@ class FastEMOverviewTab(Tab):
         a calibration or acquisition is already ongoing or not.
         :param mode: (bool) Whether the system is currently acquiring/calibrating or not acquiring/calibrating.
         """
+        # TODO also include btn_autostigmation once autostigmation is working
         self.btn_optical_autofocus.Enable(not mode)
+        self.btn_sem_autofocus.Enable(not mode)
+        self.btn_autobc.Enable(not mode)
 
     @call_in_wx_main
     def _on_optical_autofocus_done(self, future, _=None):
@@ -2107,6 +2113,10 @@ class FastEMOverviewTab(Tab):
 
     @call_in_wx_main
     def _on_btn_sem_autofocus(self, _):
+        # Disable other calibration buttons
+        # TODO also disable btn_autostigmation once autostigmation is working
+        self.btn_optical_autofocus.Enable(False)
+        self.btn_autobc.Enable(False)
         self.sem_stream_cont.stream_panel.Enable(False)
         self.sem_stream_cont.pause()
         self.sem_stream_cont.pauseStream()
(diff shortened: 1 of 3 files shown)


src/odemis/gui/cont/streams.py
```diff
@@ -78,13 +78,15 @@ class StreamController(object):
...
     def __init__(self, stream_bar, stream, tab_data_model, show_panel=True, view=None,
-                 sb_ctrl=None):
+                 sb_ctrl=None, hw_settings_display_order=None):
         """
         view (MicroscopeView or None): Link stream to a view. If view is None, the stream
         will be linked to the focused view. Passing a view to the controller ensures 
         that the visibility button functions correctly when multiple views are present.
         sb_ctrl (StreamBarController or None): the StreamBarController which (typically)
           created this StreamController. Only needed for ROA repetition display.
+        hw_settings_display_order (None or list): A custom display order for the hardware settings.
```
K4rishma, line 88: By following the none placements from the other input arguments.
```suggestion
        hw_settings_display_order (list or None): A custom display order for the hardware settings.
```

K4rishma (approved):

src/odemis/gui/cont/tabs.py
```diff
@@ -1961,12 +1961,15 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
         # Single-beam SEM stream
-        hwemt_vanames = ("resolution", "scale", "horizontalFoV")
-        emt_vanames = ("dwellTime")
-        hwdet_vanames = ("brightness", "contrast")
+        hwemt_vanames = ["horizontalFoV", "scale", "resolution"]
+        hwdet_vanames = ["contrast", "brightness"]
+        emt_vanames = ["dwellTime"]
+        # VA names as per the display order
+        # Process order is first hardware VAs, then emitter VAs and finally detector VAs
+        hw_settings_display_order = hwemt_vanames + hwdet_vanames + emt_vanames
```
pieleric, line 1969: I'm not convinced by this approach. There is already a mechanism to sort the VAs. `get_hw_config()` returns the names in a order specified in HW_SETTINGS_CONFIG. Right now, the e-beam order is set to: horizontalFoV, dwellTime, scale, resolution. You want to change to horizontalFoV, scale, resolution, dwellTime. Actually, I don't think the current order is really logical. The other microscopes would also benefit from having dwellTime away, though I feel that it should be at the top. You probably could move dwellTime *up* in the HW_SETTINGS_CONFIG. As in the case of the FASTEM it's the only one which is a "local" VA, it will be at the bottom any way

The order for the `se-detector` is just not set at all. It would be easy to add an entry with contrast and brightness. You'd just need to indicate that this order is nice because it's the same as in the TFS software (so that no one  modifies it lightly). 

So, conclusion is that probably you can drop all the changes with `hw_settings_display_order`, and just tweak HW_SETTINGS_CONFIG to obtain the right thing. Can you try?

pieleric (changes requested):

src/odemis/gui/cont/tabs.py
nandishjpatel: I tried adding the order for `se-detector` in `HW_SETTINGS_CONFIG` and it works

src/odemis/gui/cont/streams.py
```diff
@@ -346,9 +346,7 @@ def _add_hw_setting_controls(self):
 
         # Process the hardware VAs first (emitter and detector hardware VAs are combined into one
         # attribute called 'hw_vas'
-        # sort the list of hardware VAs first in case it contains items that are not in the
-        # emitter and detector conf, they will be appended at the end
-        vas_names = util.sorted_according_to(sorted(list(self.stream.hw_vas.keys())),
+        vas_names = util.sorted_according_to(list(self.stream.hw_vas.keys()),
```
pieleric, line 349: Why remove that part? It shouldn't matter for your goal, right? In general, the idea was to make sure that the VAs always are in the same order, even the ones which are not in the HW_CONFIG.

pieleric (changes requested):

PR #2441, opened 2023-06-12
remove the need for Xvfb if the DISPLAY is set to :0

nandishjpatel:


.github/workflows/self_hosted_runner.yml
```diff
@@ -38,18 +38,13 @@ jobs:
...
-      # Sleep for 5s to give Xvfb some time to start
       # Start the VirtualBox Manager
       # Sleep for 5s to give VirtualBox Manager some time to start
-      # Pass the created DISPLAY to the Virtual Machine while starting it
+      # Pass the DISPLAY to the Virtual Machine while starting it
       # Sleep for 1.5m to give the Virtual Machine time to start and automatically login
       # Set runpytest_wrapper.sh execution timeout of 8 hours, i.e. 8 * 60 * 60 * 1000 = 28800000 milliseconds
       run: |
         export DISPLAY=:0
-        Xvfb "$DISPLAY" -screen 0 1920x1280x24 -ac -noreset > /dev/null 2>&1 &
```
pieleric, line 50: What's the plan here? Did you just connect a display to the computer? Or you're planning to run Xvfb in the background all the time, so that script doesn't need to care to start/stop it? 
Could you add information about this in this runner file? If Xvfb needs to run all the time, I'd indicate at least in a comment here how to start it.

pieleric (changes requested):

.github/workflows/self_hosted_runner.yml
nandishjpatel: I updated the comments, we don't need `Xvfb` anymore

tepals (approved):

pieleric (approved):

PR #2448, opened 2023-06-22
acq test move: fix mimas move to coating failure

pieleric:
We changed the behaviour of COATING, which now just engages the GIS arm.
However, the test was left as-is.

=> Remove the "COATING" position from the microscope file, and just
check that moving to COATING engages the GIS.

### DIFF
--- a/src/odemis/acq/test/move_test.py
+++ b/src/odemis/acq/test/move_test.py
@@ -494,7 +494,6 @@ class TestMimasMove(unittest.TestCase):
 
         cls.stage_active = cls.stage.getMetadata()[model.MD_FAV_POS_ACTIVE]
         cls.stage_deactive = cls.stage.getMetadata()[model.MD_FAV_POS_DEACTIVE]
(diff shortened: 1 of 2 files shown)


FCanberk (approved):

src/odemis/acq/test/move_test.py
```diff
@@ -541,7 +543,7 @@ def test_sample_switch_procedures(self):
         self.assertEqual(pos_label, FM_IMAGING)
 
         # Test the progress update
-        progress_fm = getMovementProgress(stage.position.value, self.stage_coating, self.stage_active)
+        progress_fm = getMovementProgress(stage.position.value, self.stage_deactive, self.stage_active)
```
tepals, line 546: why use stage_deactive here, while on line 528 stage_coating is replaced by stage_active

sneepstefan (approved):

src/odemis/acq/test/move_test.py
pieleric: It's because now that coating is the same as active, we need to test something else. Like the move between deactive and active. However, leaving this check here is confusing. I'll move it.

PR #2449, opened 2023-06-22
remove additional time, set the stage to loading before milling

K4rishma:
SWM-80 Fix failing milling manager test cases

### DIFF
--- a/src/odemis/acq/test/millmng_test.py
+++ b/src/odemis/acq/test/millmng_test.py
@@ -33,7 +33,7 @@ from odemis.acq.feature import (FEATURE_ACTIVE, FEATURE_ROUGH_MILLED,
                                 CryoFeature)
 from odemis.acq.millmng import (MillingRectangleTask, MillingSettings,
                                 load_config, mill_features)
-from odemis.acq.move import _isNearPosition
+from odemis.acq.move import _isNearPosition, cryoSwitchSamplePosition, LOADING
 from odemis.acq.stream import UNDEFINED_ROI
 from odemis.util import testing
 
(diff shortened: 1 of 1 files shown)


src/odemis/acq/test/millmng_test.py
```diff
@@ -407,7 +410,6 @@ def test_milling_settings_from_yaml(self):
         f = mill_features(millings, self.sites, self.feature_post_status, self.acq_streams, self.ion_beam,
                           self.sed, self.stage, self.aligner)
 
-        time.sleep(13)  # Long enough to make sure it's started
```
FCanberk, line 410: why the sleeps are removed here?

tepals: because the mill_features already starts the future, after 13 seconds the future had already stopped so that's why the test case failed

tepals (approved):

FCanberk (approved):

src/odemis/acq/test/millmng_test.py
pieleric: 13 s was probably really long, but now you risk having test case failures because the future might not have started yet... I think a `sleep(0.1)` would be useful here.

tepals: We tested it and it was failing with a sleep of 0.1s, it would pass with a sleep of 0.00001s but I don't know if it makes sense to add that.

pieleric: Well, then it sounds to me like something is really broken. It should be milling. Milling should take several seconds.

PR #2455, opened 2023-06-22
fix RuntimeWarning: numpy.ndarray size changed, may indicate binary i…

nandishjpatel:
…ncompatibility by filtering it

### DIFF
--- a/src/odemis/acq/align/test/find_overlay_test.py
+++ b/src/odemis/acq/align/test/find_overlay_test.py
@@ -19,16 +19,18 @@ PURPOSE. See the GNU General Public License for more details.
 You should have received a copy of the GNU General Public License along with 
 Odemis. If not, see http://www.gnu.org/licenses/.
 '''
-from concurrent import futures
 import logging
-from odemis import model, dataio
-from odemis.acq import align
-from odemis.driver import semcomedi, andorcam2
-from odemis.util import testing
 import os
+import re
 import time
 import unittest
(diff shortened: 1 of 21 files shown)


tepals (approved):

FCanberk (approved):

pieleric (approved):
In the commit message, could you indicate how to reproduce this warning? Is it on Ubuntu 18.04? I don't recall seeing it on 20.04... Or does it only happen if a new version of numpy is installed with pip?

nandishjpatel:
> In the commit message, could you indicate how to reproduce this warning? Is it on Ubuntu 18.04? I don't recall seeing it on 20.04... Or does it only happen if a new version of numpy is installed with pip?

It's happening on Ubuntu 20.04 (). It could be because of step no 28. in the odemis testing document (https://docs.google.com/document/d/1pG7fviVLdISrF6BNnnd5kChT09DVfNz_/edit). Do you think this is a proper fix, or we can solve this in a better way?

pieleric:
Yes, that's most likely because a newer version of numpy is installed with pip. As discussed before, that also means that our whole CI runs with a version of numpy that most of the user don't have, just for one feature which is under (slow) development. Now it also causes warnings.
Let's leave this as is, with the code to hide the warning. We should try to find a better solution for the long term...

pieleric:
@nandishjpatel Can you update the commit message with all this info?

PR #2456, opened 2023-06-22
fix H5pyDeprecationWarning: other_ds.dims.create_scale(ds, name) is d…

nandishjpatel:
…eprecated. Use ds.make_scale(name) instead with backwards compatibility for Ubuntu 18

### DIFF
--- a/src/odemis/dataio/hdf5.py
+++ b/src/odemis/dataio/hdf5.py
@@ -335,8 +335,12 @@ def _add_image_info(group, dataset, image):
             _h5svi_set_state(group["DimensionScaleY"], ST_REPORTED)
 
             # Attach the scales to each dimensions (referenced by their label)
-            dataset.dims.create_scale(group["DimensionScaleX"], "X")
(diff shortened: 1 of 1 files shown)


K4rishma (approved):

src/odemis/dataio/hdf5.py
```diff
@@ -335,8 +335,12 @@ def _add_image_info(group, dataset, image):
             _h5svi_set_state(group["DimensionScaleY"], ST_REPORTED)
 
             # Attach the scales to each dimensions (referenced by their label)
-            dataset.dims.create_scale(group["DimensionScaleX"], "X")
-            dataset.dims.create_scale(group["DimensionScaleY"], "Y")
+            if hasattr(dataset, "make_scale"):
```
pieleric, line 338: Could you add at least at one place (ex, this line), when can we remove the `else` block? Indicate the first version of h5py which support this new function.

pieleric (changes requested):

pieleric (approved):

PR #2457, opened 2023-06-22
Fix test_resize

tepals:
The test machine has a small virtual monitor, therefore increasing size of the window by a factor 2 automatically decreased the window again because it did not fit the monitor size. This fix uses 1.5 as a scaling factor so that the inceased window size fits the monitor.


K4rishma (approved):

pieleric (approved):
Looks fine. Did you check this actually solves the issue? I guess you could try to just run on your computer when setting a small screen resolution...

PR #2463, opened 2023-06-22
[FIX] driver piezomotor: fix the current position not updating proper…

tepals:
…ly in the simulator

_do_move updated the current position by deep copying self.target_pos, sometimes this caused that the wrong position got stored on self.current_pos. Pass the target_pos to _do_move so that after the move the current_pos gets updated with the correct target_pos.

### DIFF
--- a/src/odemis/driver/piezomotor.py
+++ b/src/odemis/driver/piezomotor.py
@@ -19,22 +19,23 @@ PURPOSE. See the GNU General Public License for more details.
 You should have received a copy of the GNU General Public License along with
 Odemis. If not, see http://www.gnu.org/licenses/.
 """
-from concurrent.futures import CancelledError
 import copy
 import glob
 import logging
+import os
+import re
+import threading
+import time
+from concurrent.futures import CancelledError
+from threading import Thread
+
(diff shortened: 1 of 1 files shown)


src/odemis/driver/piezomotor.py
```diff
@@ -1159,5 +1159,5 @@ def _do_move(self, steps):
                 return
             else:
                 time.sleep(0.1)
-        self.current_pos = copy.deepcopy(self.target_pos)
+        self.current_pos = target_pos
```
pieleric, line 1162: I can imagine that passing `target_pos` helps because maybe `self.target_pos` has changed while the move is happening. However, I don't think we can drop the `deepcopy()` . Now both variables are going to point to the same dict. So every call to "E" will directly update the current_pos (as well as target_pos). I doubt this is right.

pieleric (changes requested):

src/odemis/driver/piezomotor.py
tepals: I will put back the `deepcopy`

PR #2464, opened 2023-06-22
Numpy deprecation int float bool

tepals:


nandishjpatel (approved):

src/odemis/driver/picoquant.py
```diff
@@ -2207,7 +2207,7 @@ def _generate(self):
 
         # Read data and make it a DataArray
         d = self.parent.GetCountRate(self._channel)
-        nd = numpy.array([d], dtype=numpy.int)
+        nd = numpy.array([d], dtype=int)
```
pieleric, line 2210: I think we should actually specify the int. Need to check what the GetCountRate() C function returns, and use the same type. Most like numpy.uint32.

tepals: This has been fixed in: https://github.com/delmic/odemis/pull/2467

PR #2465, opened 2023-06-23
Modularize stage movement

K4rishma:
Extend acq.move to detect which type of Meteor is used for moving
METEOR-624
Task -> https://delmic.atlassian.net/browse/METEOR-624?atlOrigin=eyJpIjoiMWM5YWZjODQ4Yjk0NDVkNWJjYzFlZjJhYjVlODJiZTEiLCJwIjoiaiJ9

### DIFF
--- a/plugins/meteor_engage_warn.py
+++ b/plugins/meteor_engage_warn.py
@@ -21,7 +21,7 @@ see http://www.gnu.org/licenses/.
 
 import logging
 import math
-from odemis.acq.move import FM_IMAGING, SEM_IMAGING, getCurrentPositionLabel
+from odemis.acq.move import FM_IMAGING, SEM_IMAGING, MicroscopePostureManager
 from odemis.gui.plugin import Plugin
 from odemis.util.units import readable_str
 from typing import Dict
@@ -47,7 +47,7 @@ class MeteorEngageWarnPlugin(Plugin):
 
     def __init__(self, microscope, main_app):
         super().__init__(microscope, main_app)
-
+        self.meteor_manager = MicroscopePostureManager(microscope)
         self.main_frame = main_app.main_frame
 
         # It only makes sense if the METEOR chamber tab is present
@@ -77,7 +77,7 @@ class MeteorEngageWarnPlugin(Plugin):
         pos_str = "\n". join(pos_str)
 
         # Guess (back) which position the users wants to go to
-        target_pos = getCurrentPositionLabel(end_pos, self._chamber_tab._stage)
+        target_pos = self.meteor_manager.getCurrentPostureLabel(end_pos)
         if target_pos == FM_IMAGING:
             warn_msg = FM_WARN_MSG
         elif target_pos == SEM_IMAGING:
--- a/plugins/meteor_engage_warn.py
+++ b/plugins/meteor_engage_warn.py
@@ -47,8 +47,6 @@ class MeteorEngageWarnPlugin(Plugin):
 
     def __init__(self, microscope, main_app):
         super().__init__(microscope, main_app)
-        self.meteor_manager = MicroscopePostureManager(microscope)
-        self.main_frame = main_app.main_frame
 
         # It only makes sense if the METEOR chamber tab is present
         try:
@@ -77,7 +75,7 @@ class MeteorEngageWarnPlugin(Plugin):
         pos_str = "\n". join(pos_str)
 
         # Guess (back) which position the users wants to go to
-        target_pos = self.meteor_manager.getCurrentPostureLabel(end_pos)
+        target_pos = self.main_app.main_frame.getCurrentPostureLabel(end_pos)
         if target_pos == FM_IMAGING:
             warn_msg = FM_WARN_MSG
         elif target_pos == SEM_IMAGING:
--- a/src/odemis/acq/move.py
+++ b/src/odemis/acq/move.py
@@ -334,6 +334,8 @@ class MeteorTFS1PostureManager(MeteorPostureManager):
         # Check required metadata used during switching
         self.required_keys.add(model.MD_POS_COR)
         self.check_stage_metadata(required_keys=self.required_keys)
+        if not {"x", "y", "rz", "rx"}.issubset(self.stage.axes):
+            raise KeyError("The stage misses 'x', 'y', 'rx' or 'rz' axes")
 
     def getTargetPosition(self, target_pos_lbl: int) -> Dict[str, float]:
         """
@@ -418,12 +420,9 @@ class MeteorTFS1PostureManager(MeteorPostureManager):
         """
         Transforms the current stage position from the SEM imaging area to the
         meteor/FM imaging area.
-        :param pos: (dict str->float) the initial stage position. The position has to have x, y, rz, and rx axes,
-         otherwise error would be raised.
-        :return: (dict str->float) the transformed position. It returns the updated axes x, y, rx, rz. The axis z is same as the input.
+        :param pos: (dict str->float) the initial stage position.
+        :return: (dict str->float) the transformed position.
         """
-        if not {"x", "y", "rz", "rx"}.issubset(self.stage.axes):
-            raise KeyError("The stage misses 'x', 'y', 'rx' or 'rz' axes")
         stage_md = self.stage.getMetadata()
         transformed_pos = pos.copy()
         pos_cor = stage_md[model.MD_POS_COR]
@@ -443,11 +442,9 @@ class MeteorTFS1PostureManager(MeteorPostureManager):
         """
         Transforms the current stage position from the meteor/FM imaging area
         to the SEM imaging area.
-        :param pos: (dict str->float) the initial stage position
+        :param pos: (dict str->float) the initial stage position.
         :return: (dict str->float) the transformed stage position.
         """
-        if not {"x", "y", "rx", "rz"}.issubset(self.stage.axes):
-            raise KeyError("The stage misses 'x', 'y', 'rx' or 'rz' axes")
         stage_md = self.stage.getMetadata()
         transformed_pos = pos.copy()
         pos_cor = stage_md[model.MD_POS_COR]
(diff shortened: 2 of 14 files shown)


src/odemis/gui/util/__init__.py
```diff
@@ -292,7 +292,9 @@ def enable_tab_on_stage_position(button, stage, pos, target, aligner=None, toolt
     :param aligner: (Actuator) the align component
     :param tooltip: (str or None) Tooltip message to show when disabled
     """
-    within_target = getCurrentPositionLabel(pos, stage, aligner) in target
+    microscope = model.getMicroscope()
+    microscope_manager = MicroscopePostureManager(microscope)
+    within_target = microscope_manager.getCurrentPostureLabel(pos) in target
```
pieleric, line 297: No, don't create a new posture manager everytime. Instead, use the one from the GUI, by taking it as an argument. Note also that the stage, pos and aligner arguments are not used anymore.
=> change the signature to `button, posture_manager, target, tooltip=None`.
Don't forget to update every

pieleric (changes requested):

src/odemis/acq/move.py
```diff
@@ -52,685 +53,735 @@
...
+            elif stage_version == "tfs_1":
+                return super().__new__(MeteorTFS1PostureManager)
+            else:
+                raise ValueError(f"Stage version {stage_version} is not supported")
+        if role == "mimas":
+            return super().__new__(MimasPostureManager)
+        else:
+            return super().__new__(cls)  # default
+
+    def __init__(self, microscope):
```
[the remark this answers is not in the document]
K4rishma: @pieleric Do I add abstract method in other methods of this class? There are also not used for e.g. getCurrentPostureLabel(self, current_pos)

src/odemis/acq/move.py
```diff
@@ -52,685 +53,735 @@
...
+    def getTargetPosition(self, target_pos_lbl):
+        """
+        Returns the position that the stage would go to.
+        target_pos_lbl (int): a label representing a position (SEM_IMAGING, FM_IMAGING, GRID_1 or GRID_2)
+        returns (dict->float): the end position of the stage
+        raise ValueError: if the target position is not supported
+        """
+        pass
+
+    def getCurrentGridLabel(self, current_pos):
```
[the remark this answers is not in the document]
K4rishma: current_pos is required in cases where it checks the target posture label based on a target position. In that case, it needs the stage position as the input argument.

src/odemis/acq/move.py
```diff
@@ -52,685 +53,735 @@
...
+    def __init__(self, microscope):
+        # Load components and favorite position metadata
+        # Check required metadata using check_stage_metadata(MD_CALIB)
+        role = microscope.role
+        self.stage = model.getComponent(role='stage-bare')
+
+    def getCurrentPostureLabel(self, current_pos):
+        """
+        Determine where lies the current stage position
+        :param current_pos: (dict str->float) Current position of the stage
```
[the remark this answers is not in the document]
K4rishma: current_pos is required in cases where it checks the target posture label based on a target position. In that case, it needs the stage position as the input argument.

src/odemis/acq/move.py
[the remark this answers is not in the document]
K4rishma: Here it can be dropped

src/odemis/acq/move.py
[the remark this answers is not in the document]
K4rishma: image

src/odemis/acq/move.py
```diff
@@ -52,685 +55,728 @@
...
+        # set linear axes and rotational axes used
+        self.axes = self.stage.axes
+        self.linear_axes = set(key for key in self.axes.keys() if key in {'x', 'y', 'z', 'm'})
+        self.rotational_axes = set(key for key in self.axes.keys() if key in {'rx', 'ry', 'rz', 'rm'})
+        # required keys that must be present in the stage metadata
+        self.required_keys = {
+            model.MD_FAV_POS_DEACTIVE, model.MD_FAV_SEM_POS_ACTIVE, model.MD_FAV_FM_POS_ACTIVE,
+            model.MD_SAMPLE_CENTERS}
+
+    def getCurrentPostureLabel(self, pos: Dict[str, float]) -> int:
```
pieleric, line 283: Please drop the `pos` argument as well. You can read it (now) directly with `self.stage.position.value`. 

I don't think so, but if there is some caller that requires the name of different position than the current one, then you can leave this argument, but only as optional (and typically, it'd be `None`, in which case the position is read from self.stage).

src/odemis/acq/move.py
```diff
@@ -52,685 +55,728 @@
...
-                    target_name))
-
-            # Add the sub moves to perform the imaging/alignment move
-            # Moving Y axis first downwards so Z move upwards it doesn't hit the 5DOF stage
-            sub_moves = [{'y'}, {'z'}, {'x'}]
+    def getTargetPosition(self, target_pos_lbl: int) -> dict:
+        """
+        Returns the position that the stage would go to.
+        target_pos_lbl (int): a label representing a position (SEM_IMAGING, FM_IMAGING, GRID_1 or GRID_2)
+        returns (dict->float): the end position of the stage
```
pieleric, line 307:
```suggestion
        :returns (dict str->float): the target position of the stage
```

src/odemis/acq/move.py
```diff
@@ -52,685 +55,728 @@
...
-
-        elif target in (ALIGNMENT, IMAGING, THREE_BEAMS):
-            if current_label is UNKNOWN:
-                raise ValueError("Unable to move aligner to {} while current position is unknown.".format(
-                    target_name))
-
-            # Add the sub moves to perform the imaging/alignment move
-            # Moving Y axis first downwards so Z move upwards it doesn't hit the 5DOF stage
-            sub_moves = [{'y'}, {'z'}, {'x'}]
+    def getTargetPosition(self, target_pos_lbl: int) -> dict:
```
pieleric, line 303:
```suggestion
    def getTargetPosition(self, target_pos_lbl: int) -> Dict[str, float]:
```

src/odemis/acq/move.py
```diff
@@ -767,14 +845,103 @@ def _doCryoSwitchSamplePosition(future, target):
...
+class MimasPostureManager(MicroscopePostureManager):
+    def __init__(self, microscope):
+        # Load components
+        self.stage = model.getComponent(role="stage")
+        self.align = model.getComponent(role="align")
+        self.gis = model.getComponent(role="gis")
+        # set linear axes and rotational axes used
+        self.axes = self.stage.axes
+        self.linear_axes = set(key for key in self.axes.keys() if key in {'x', 'y', 'z', 'm'})
+        self.rotational_axes = set(key for key in self.axes.keys() if key in {'rx', 'ry', 'rz', 'rm'})
```
pieleric, line 871: Please add a few check for the metadata. At least, from the code it's clear these 3 metadata are required: MD_FAV_POS_ACTIVE, MD_FAV_POS_DEACTIVE, MD_POS_ACTIVE_RANGE.
The aligner also has some metadata requirements.

src/odemis/acq/move.py
```diff
@@ -834,160 +1001,410 @@ def _doCryoSwitchSamplePosition(future, target):
...
+class EnzelPostureManager(MicroscopePostureManager):
+    def __init__(self, microscope):
+        # Load components
+        self.stage = model.getComponent(role='stage')
+        self.align = model.getComponent(role='align')
+        # set linear axes and rotational axes used
+        self.axes = self.stage.axes
+        self.linear_axes = set(key for key in self.axes.keys() if key in {'x', 'y', 'z', 'm'})
+        self.rotational_axes = set(key for key in self.axes.keys() if key in {'rx', 'ry', 'rz', 'rm'})
+
```
pieleric, line 1055: Need to check the metadata too. See the gui.cont.tabs for the (minimal) list of metadata required.

src/odemis/acq/test/move_test.py
```diff
@@ -307,185 +315,187 @@ def setUpClass(cls):
...
-    Test move functions on the MIMAS
+    Test the MimasPostureManager functions
     """
     backend_was_running = False
 
     @classmethod
     def setUpClass(cls):
         testing.start_backend(MIMAS_CONFIG)
+        cls.microscope = model.getMicroscope()
+        cls.posture_manager = MicroscopePostureManager(microscope=cls.microscope)
```
pieleric, line 498: For all these `Test*Move` classes, I'd suggest to add a check that the posture manager is the correct type. Something like `cls.assertIsInstance(cls.posture_manager, MimasPostureManager)`.

src/odemis/gui/cont/tabs.py
```diff
@@ -4953,6 +4962,7 @@ def __init__(self, name, button, panel, main_frame, main_data):
         # Disable the tab when the stage is not at the right position
         main_data.is_acquiring.subscribe(self._on_acquisition, init=True)
 
+
```
pieleric, line 4965: Please drop this extra line.

pieleric (changes requested):

src/odemis/acq/move.py
```diff
@@ -834,160 +1006,414 @@ def _doCryoSwitchSamplePosition(future, target):
...
+        if isNearPosition(current_pos, align_deactive, self.align.axes):
+            return LOADING
+
+        # Anywhere around POS_ACTIVE, is THREE_BEAMS
+        # As POS_ACTIVE is updated every time the aligner is moved, it's typically
+        # exactly at POS_ACTIVE.
+        # TODO: should have a POS_ACTIVE_RANGE to define the whole region
+        if (isNearPosition(current_pos, align_active, self.align.axes) or
+                isNearPosition(current_pos, align_alignment, self.align.axes) or
+                isNearPosition(current_pos, three_beams, self.align.axes)
```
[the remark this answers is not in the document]
pieleric: I think you can solve it more by placing the last `):` as a separate line, with the `)` aligned on the first `(`.

plugins/meteor_engage_warn.py
```diff
@@ -77,7 +77,7 @@ def _display_meteor_pos_warning_msg(self, end_pos: Dict[str, float]) -> bool:
         pos_str = "\n". join(pos_str)
 
         # Guess (back) which position the users wants to go to
-        target_pos = getCurrentPositionLabel(end_pos, self._chamber_tab._stage)
+        target_pos = self.meteor_manager.getCurrentPostureLabel(end_pos)
```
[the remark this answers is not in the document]
K4rishma: @pieleric due to this function getCurrentPostureLabel(stage_position) requires one input argument that consists of target stage position. As a result,  I am not able to remove this input argument in other files like path.py line 1042.  How do I resolve this?

src/odemis/acq/move.py
pieleric: Yes, please add this more complete test to the PostureManager.

src/odemis/gui/cont/tabs.py
```diff
@@ -3305,7 +3287,7 @@ def _perform_switch_position_movement(self, target_button):
 
         # define the start position
         self._start_pos = current_pos = self._stage.position.value
-        current_position = getCurrentPositionLabel(current_pos, self._stage, self._aligner)
+        current_position = self.posture_manager.getCurrentPostureLabel()
 
         if self._role == 'enzel':
```
pieleric, line 3292: I think that we could change the ENZEL code here to the same as the METEOR: it would use `self.posture_manager.getTargetPosition()`, instead of using its own dictionary. That would save a lot of code, as the dict is used just for this place.

We could also do that separately, as I think this code could be refactored even further...

pieleric (changes requested):

FCanberk (approved):

PR #2466, opened 2023-06-23
util pituner: add master option

pieleric:
Needed to work with the latest C885 controller, which now uses master = 1.

Also fix an issue in Python 3.

### DIFF
--- a/util/pituner
+++ b/util/pituner
@@ -241,9 +241,9 @@ def main(args):
     parser.add_argument('--controller', dest="addr", type=int,
                         help="Controller address (if controller needs it)")
     parser.add_argument('--master', dest="master", type=int,
-                    help="The address of the 'master' controller when connecting over "
-                         "TCP/IP to multiple controllers. It is unused when connecting over serial port."
-                         "Default is 254 if applicable.")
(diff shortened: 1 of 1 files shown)


util/pituner
```diff
@@ -239,6 +240,10 @@ def main(args):
                         help="Port name (ex: /dev/ttyUSB0, autoip, or 192.168.95.5)")
     parser.add_argument('--controller', dest="addr", type=int,
                         help="Controller address (if controller needs it)")
+    parser.add_argument('--master', dest="master", type=int,
+                    help="The address of the 'master' controller when connecting over "
+                         "TCP/IP to multiple controllers. It is unused when connecting over serial port."
+                         "Default is 254 if applicable.")
```
tepals, line 246:
```suggestion
    parser.add_argument('--master', dest="master", type=int,
                        help="The address of the 'master' controller when connecting over "
                             "TCP/IP to multiple controllers. It is unused when connecting over serial port."
                             "Default is 254 if applicable.")
```

tepals (changes requested):

tepals (approved):

sneepstefan (approved):

PR #2469, opened 2023-06-27
Extend the TESCAN driver to support rotational axes

sneepstefan:
Since we do not yet support installing a METEOR on a TECAN SEM it is necessary to adjust the current driver to be able to handle controlled stage rotation to be able to align METEOR and the CCD properly.

Right now there is no support for axis other than for linear axis therefore the existing driver should be extended. The addition to the existing driver is only support for moving with the rotational axes.
Basic requirements are that the user will be able to control the stage through Odemis for moving all possible axes installed.

In addition test cases are written to test the  added functionality with moving the 2 newly added axes rx and rz.

### DIFF
--- a/src/odemis/driver/tescan.py
+++ b/src/odemis/driver/tescan.py
@@ -1039,6 +1039,8 @@ class Stage(model.Actuator):
     def __init__(self, name, role, parent, **kwargs):
         self._position = {}
         axes_def = {}
+        # limits are always returned as a list with min and max values in the order of x,y,z,r,t
+        # the list should always return 10 values and if an axis is not motorized, limits are zero
         axes_rng = parent._device.StgGetLimits(1)  # using the soft limits parameter here
         axes_motorized = parent._device.StgGetMotorized()
 
@@ -1050,12 +1052,15 @@ class Stage(model.Actuator):
                     # make a distinction between linear axes and rotational ones
                     # rotational axes will need their range converted to radians
                     if ax.startswith("r"):
+                        if axes_rng[:2] == [-360, 360]:  # Full rotation: report as 0 -> 360°
+                            axes_rng[0] = 0
                         axes_def[ax] = model.Axis(unit="rad",
-                                                  range=(self._degrees_to_rad(axes_rng[0]),
-                                                         self._degrees_to_rad(axes_rng[1])))
+                                                  range=(math.radians(axes_rng[0]),
+                                                         math.radians(axes_rng[1])))
                     else:
                         # convert the axis min and max range to meters
                         axis_m = (axes_rng[0] * 1e-3, axes_rng[1] * 1e-3)
+                        # note that the linear axes range is switched and inverted by default (for historical reasons)
                         axes_def[ax] = model.Axis(unit="m", range=(-axis_m[1], -axis_m[0]))
                 # slice the axes_rng list for the next iteration
                 axes_rng = axes_rng[2::]
@@ -1104,61 +1109,49 @@ class Stage(model.Actuator):
         self._position["y"] = -y * 1e-3
         self._position["z"] = -z * 1e-3
 
-        if set(self.axes.keys()).intersection({"rz", "rx"}):
-            self._position["rz"] = self._degrees_to_rad(rz)
-            self._position["rx"] = self._degrees_to_rad(rx)
+        if "rz" in self.axes:
+            self._position["rz"] = math.radians(rz)
+        if "rx" in self.axes:
+            self._position["rx"] = math.radians(rx)
 
         # it's read-only, so we change it via _value
         pos = self._applyInversion(self._position)
         self.position._set_value(pos, force_write=True)
         logging.debug("Updated stage position to %s", pos)
 
-    def _degrees_to_rad(self, val: float) -> float:
-        """
-        change the value of val from degrees to radians
-        :param val: float value in degrees to convert
-        :return: float value in radians
-        """
-        return math.radians(val)
-
-    def _rad_to_degrees(self, val: float) -> float:
-        """
-        change the value of val from radians to degrees
-        :param val: float value in radians to convert
-        :return: float value in degrees
-        """
-        return math.degrees(val)
-
     def _doMoveAbs(self, pos: dict):
         """
         move to the requested (absolute) position
-        :param pos: positions of linear axes in mm or rotational axes in radians
+        :param pos (dict[str, float]): positions of linear axes in m or rotational axes in radians
         """
-        # TODO: support cancelling (= call StgStop)
+        # TODO: support cancelling (= call StgStop) will be addressed in separate PR
         with self.parent._acq_progress_lock:
             logging.debug("Requesting stage move to %s", pos)
 
-            # If not all axes requested, fill up the rest with the current position
-            if pos.keys() < {"x", "y", "z", "rz", "rx"}:
-                self._updatePosition()
-                current_pos = self._position.copy()
-                current_pos.update(pos)
-                pos = current_pos
+            x, y, z, rz, rx = self.parent._device.StgGetPosition()
+            current_pos = {"x": x, "y": y, "z": z, "rz": rz, "rx": rx}
 
-            # Position from m to mm and inverted for the linear axes, from degree to radians for the rotational axes
-            # if any rotational axis movement is requested send a 5 param move command otherwise a 3 param command.
-            if set(self.axes.keys()).intersection({"rz", "rx"}):
-                self.parent._device.StgMoveTo(-pos["x"] * 1e3,
-                                              -pos["y"] * 1e3,
-                                              -pos["z"] * 1e3,
-                                              self._rad_to_degrees(pos["rz"]),
-                                              self._rad_to_degrees(pos["rx"]),
-                                              )
-            else:
-                self.parent._device.StgMoveTo(-pos["x"] * 1e3,
-                                              -pos["y"] * 1e3,
-                                              -pos["z"] * 1e3,
-                                              )
+            req_pos = {}
+
+            for axis in {"x", "y", "z", "rz", "rx"}:
+                if axis in pos:
+                    # convert from m to mm and invert for the linear axes
+                    if axis in {"x", "y", "z"}:
+                        req_pos[axis] = -pos[axis] * 1e3
+                    # convert from radians to degrees for the rotational axes
+                    elif axis in {"rz", "rx"}:
+                        req_pos[axis] = math.degrees(pos[axis])
+
+            current_pos.update(req_pos)
+
+            # always issue a move command containing values for all 5 axes so
+            # there is no separate code needed for backward compatibility
+            self.parent._device.StgMoveTo(current_pos["x"],
+                                          current_pos["y"],
+                                          current_pos["z"],
+                                          current_pos["rz"],
+                                          current_pos["rx"],
+                                          )
 
             # a very small delay before checking if the stage is busy
             time.sleep(0.1)
@@ -1173,19 +1166,19 @@ class Stage(model.Actuator):
     def _doMoveRel(self, shift: dict):
         """
         Request the position in absolute coordinates and create new positions after applying shift values
-        :param shift: shift of linear axes in mm or rotational axes in radians
+        :param shift (dict[str, float]): shift of linear axes in m or rotational axes in radians
         """
         self._updatePosition()
         pos = self._position.copy()
 
         # add the requested change to the current position
         for axis, change in shift.items():
-            if axis not in {"rz"}:
-                pos[axis] += change
-            # different approach for the Rz axis as it can fully turn back and forth
-            else:
-                # match against full rotation value to cover over- and underflow
-                pos[axis] = (pos[axis] + change) % math.radians(360)
+            pos[axis] += change
+            if axis in ("rx", "rz"):
+                rng = self.axes[axis].range
+                if abs(rng[1] - rng[0]) >= 2 * math.pi:
+                    # for full rotational axes this check maps values outside the range to between 0 and 2pi
+                    pos[axis] = (pos[axis] - rng[0]) % (2 * math.pi) + rng[0]
 
         self._checkMoveAbs(self._applyInversion(pos))
         self._doMoveAbs(pos)
--- a/src/odemis/driver/tescan.py
+++ b/src/odemis/driver/tescan.py
@@ -1054,7 +1054,9 @@ class Stage(model.Actuator):
                                                   range=(self._degrees_to_rad(axes_rng[0]),
                                                          self._degrees_to_rad(axes_rng[1])))
                     else:
-                        axes_def[ax] = model.Axis(unit="m", range=(axes_rng[0], axes_rng[1]))
+                        # convert the axis min and max range to meters
+                        axis_m = (axes_rng[0] * 1e-3, axes_rng[1] * 1e-3)
+                        axes_def[ax] = model.Axis(unit="m", range=(-axis_m[1], -axis_m[0]))
                 # slice the axes_rng list for the next iteration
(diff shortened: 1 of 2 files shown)


src/odemis/driver/test/tescan_test.py
```diff
@@ -189,6 +372,7 @@ def test_probe_current(self):
         time.sleep(6)  # Wait for value refresh
         self.assertAlmostEqual(orig_probe_current, ebeam.probeCurrent.value)
 
+    # @skip
```
FCanberk, line 375: Is this line necessary? If so, please add a comment line to explain why it is there and commented out.

src/odemis/driver/test/tescan_test.py
```diff
@@ -133,6 +145,177 @@ def test_pickle(self):
...
+        Test if it's possible to stop the stage while it's moving by using a big move.
+        """
+        start_pos = self.stage.position.value.copy()
+
+        # check a big move and stop it
+        f = self.stage.moveRel({"rz": math.radians(180)})
+        f.result()
+
+        # wait for 1 second and stop the movement
+        time.sleep(0.2)
```
FCanberk, line 313: In the comment it says "wait for 1 second" but the sleep duration is 0.2 seconds. The mismatch needs to be fixed.

src/odemis/driver/test/tescan_test.py
```diff
@@ -299,10 +483,12 @@ def test_move_abs(self):
         time.sleep(6)
         testing.assert_pos_almost_equal(self.stage.position.value, p)
 
+    @skip  # not working with the newer simulator Tescan Essence v1.2.2
```
tepals, line 486: You can specify why you are skipping something:
```suggestion
    @skip("not working with the newer simulator Tescan Essence v1.2.2")
```

src/odemis/driver/test/tescan_test.py
sneepstefan: It was a workaround for running it with the Amber simulator, since this class was written for the Mira simulator. With the Mira simulator running it actually just works fine. For now I will add a comment to it as you suggested!

src/odemis/driver/tescan.py
```diff
@@ -1127,12 +1170,22 @@ def _doMoveAbs(self, pos):
...
+        """
         self._updatePosition()
         pos = self._position.copy()
+
+        # add the requested change to the current position
         for axis, change in shift.items():
-            pos[axis] += change
+            if axis not in {"rz"}:
+                pos[axis] += change
+            # different approach for the Rz axis as it can fully turn back and forth
```
pieleric, line 1185: As you use the stage limits given by the SEM software, you cannot have so much expectations here.
I'd suggest to use the same code as we've used in other drivers (eg, xt_client):
```
if axis in ("rx", "rz"):
    rng = self.axes[axis].range
    if  util.almost_equal(rng[1] - rng[0], 2 * math.pi):
                pos[axis] = (pos[an] - rng[0]) % (2 * math.pi) + rng[0]
```

src/odemis/driver/test/tescan_test.py
```diff
@@ -299,10 +485,12 @@ def test_move_abs(self):
         time.sleep(6)
         testing.assert_pos_almost_equal(self.stage.position.value, p)
 
+    @skip("not working with the newer simulator Tescan Essence v1.2.2")
     def test_stop(self):
         """
         Check it's possible to stop the stage while it's moving.
         """
+        # TODO: make this method work again with Tescan Essence v1.2.2.
```
pieleric, line 493: I think now is the good time to look at this. Especially, the code really doesn't look like doing anything especially odd. Maybe it just tries to move too far (50mm)? I think there is no good reason to use such a large value. 5mm should be fine for testing.

src/odemis/driver/tescan.py
```diff
@@ -1086,38 +1099,68 @@ def _updatePosition(self):
...
+        """
+        return math.radians(val)
+
+    def _rad_to_degrees(self, val: float) -> float:
+        """
+        change the value of val from radians to degrees
+        :param val: float value in radians to convert
+        :return: float value in degrees
+        """
+        return math.degrees(val)
```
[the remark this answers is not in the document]
sneepstefan: Yeah I agree on that. At first I though it would make the code much better readable and in addition if these methods ever need any more code they would already be there. But for now indeed lets stick to directly using the math functions.

src/odemis/driver/test/tescan_test.py
sneepstefan: I rather keep this as a TODO, I tried to play around with the move value and with the time.sleep setting but it seems this would not be a  quick fix. There are more parts in this method failing for instance the CancelledError which is also not raised. The latter might have something to do with cancelling not supported yet. I think we should focus on basic functionality for now.

src/odemis/driver/tescan.py
```diff
@@ -1086,38 +1102,55 @@ def _updatePosition(self):
...
+                        req_pos[axis] = math.degrees(pos[axis])
 
-            # Position from m to mm and inverted
-            self.parent._device.StgMoveTo(-pos["x"] * 1e3,
-                                          -pos["y"] * 1e3,
-                                          -pos["z"] * 1e3)
+            current_pos.update(req_pos)
 
-            # TODO: does it need a bit of time before checking it's busy?
+            # always issue a move command containing values for all 5 axes
```
tepals, line 1145: can you add why?

src/odemis/driver/tescan.py
```diff
@@ -1127,12 +1160,25 @@ def _doMoveAbs(self, pos):
             logging.debug("Stage move to %s completed", pos)
             self._updatePosition()
 
-    def _doMoveRel(self, shift):
-        # Compute the position in absolute coordinates
+    def _doMoveRel(self, shift: dict):
+        """
+        Request the position in absolute coordinates and create new positions after applying shift values
+        :param shift: shift of linear axes in m or rotational axes in radians
```
tepals, line 1166: can you add what the keys and values are of this dict? I guess the axes names are the keys and the value is the corresponding shift.

src/odemis/driver/test/tescan_test.py
```diff
@@ -60,17 +62,27 @@
                            "focus": CONFIG_FOCUS,
                            # "camera": CONFIG_CM,
                            "pressure": CONFIG_PRESSURE},
-              "host": "192.168.1.175"
+              # "host": "192.168.1.175"
```
tepals, line 65: why leave this here as a comment?

src/odemis/driver/test/tescan_test.py
```diff
@@ -133,6 +145,177 @@ def test_pickle(self):
...
+        f = self.stage.moveAbs(new_pos)
+        f.result()
+        testing.assert_pos_almost_equal(self.stage.position.value, new_pos)
+
+        # move exactly on the edge of the max range of the rz axis
+        new_pos["rz"] = math.radians(360)
+        f = self.stage.moveAbs(new_pos)
+        f.result()
+        new_pos["rz"] = 0.0
+        testing.assert_pos_almost_equal(self.stage.position.value, new_pos)
```
tepals, line 200: can you add a comment explaining why moving to 360 degrees, actually moves to zero degrees?

src/odemis/driver/test/tescan_test.py
```diff
@@ -133,6 +145,177 @@ def test_pickle(self):
...
+    def test_move_out_of_range(self):
+        """
+        test if a relative move and an absolute move out of the range generate ValueErrors
+        """
+        start_pos = self.stage.position.value.copy()
+
+        # Try a relative move outside the range (less than min)
+        axes = self.stage.axes
+        with self.assertRaises(ValueError):
+            self.stage.moveRel({"x": axes["x"].range[1] - axes["x"].range[0] + 10e-6})
```
tepals, line 234: usually the range[0] < range[1], so this would move it to more than the max range, not less than min, which is mentioned in the comment.

src/odemis/driver/tescan.py
```diff
@@ -1127,12 +1160,25 @@ def _doMoveAbs(self, pos):
...
+
+        # add the requested change to the current position
         for axis, change in shift.items():
             pos[axis] += change
+            if axis in ("rx", "rz"):
+                rng = self.axes[axis].range
+                # check if any of the range values are 360°, range values for full rotational axis
+                # requested with command StgGetLimits() always returns (-360, 360)
+                if (2 * math.pi) in [abs(ele) for ele in rng]:
+                    # this check is for full rotational axes to handle under- and overflow
```
tepals, line 1179: This comment is not completely clear, if I read the code correctly it maps values outside of the range to between 0 and 2pi? Then I would say something like
```suggestion
                    # for full rotational axes this check maps values outside the range to between 0 and 2pi
```

src/odemis/driver/test/tescan_test.py
```diff
@@ -189,6 +372,8 @@ def test_probe_current(self):
         time.sleep(6)  # Wait for value refresh
         self.assertAlmostEqual(orig_probe_current, ebeam.probeCurrent.value)
 
+    # skip this method if running in a simulator not specifically for
+    # the Mira apparently the wait time of 6s is quite critical here
```
tepals, line 376: shouldn't you then add a `@unittest.skip`?

tepals (changes requested):

src/odemis/driver/tescan.py
```diff
@@ -1127,12 +1160,25 @@ def _doMoveAbs(self, pos):
...
+                    if util.almost_equal(abs(rng[0]) - abs(rng[1]), 0.0):
+                        pos[axis] = (pos[axis] - abs(rng[0])) % (2 * math.pi)
```
pieleric, line 1181: This is odd code. It seems the 2 `if`'s are checking for the same thing: whether the rotational axis supports full rotation or not. I'd suggest to just write the equivalent of "if the rotation range is >= 360°, it's considered to support full rotation". You can write it this way: `if abs(rng[1] - rng[0]) > 2 * math.pi`

pieleric (changes requested):
Looks good. Just one small comment about rotational axes range check.

src/odemis/driver/tescan.py
pieleric: Indeed, typically you could specify the type as `Dict[str, float]`.

sneepstefan: Thanks I also added this specification to the _doMoveAbs method.

src/odemis/driver/test/tescan_test.py
sneepstefan: Good point, changed the comment to a more descriptive one

src/odemis/driver/test/tescan_test.py
sneepstefan: I dont know what changed but now this test is passing constantly so I removed the whole comment eventually.

FCanberk (approved):

src/odemis/driver/tescan.py
```diff
@@ -1127,12 +1161,22 @@ def _doMoveAbs(self, pos):
...
+        """
         self._updatePosition()
         pos = self._position.copy()
+
+        # add the requested change to the current position
         for axis, change in shift.items():
             pos[axis] += change
+            if axis in ("rx", "rz"):
+                rng = self.axes[axis].range
+                if abs(rng[1] - rng[0]) > 2 * math.pi:
```
pieleric, line 1177: I think you want to have a `>=`, to also include the (common) case where it's exactly reported as 360°.

src/odemis/driver/tescan.py
```diff
@@ -1127,12 +1161,22 @@ def _doMoveAbs(self, pos):
...
+                    # for full rotational axes this check maps values outside the range to between 0 and 2pi
+                    pos[axis] = (pos[axis] - abs(rng[0])) % (2 * math.pi)
```
pieleric, line 1179: This code does not look correct. Unless rng[0] == 0 (which is probably very common), it will not adjust back the value to fit the range. I think this should work (same code as in xt_client):
```suggestion
                    pos[axis] = (pos[axis] - rng[0]) % (2 * math.pi) + rng[0]
```

pieleric (changes requested):

src/odemis/driver/tescan.py
pieleric: I understand. That code indeed somehow assumes that the total is range is 2π. However, with the hardware, you end up with 4π . I guess we could adjust it to use the actual range, using something like that: 
`(pos[axis] - rng[0]) % abs(rng[1] - rng[0]) + rng[0]`.
However, this might cause other weird behaviour if the hardware reports incorrect range which is not exactly a multiple of 2π. 

I'm not sure why they support such range in the first place. As they don't support relative moves, I could imagine that this is to allow indicating the direction of the rotation. For example, if the stage is at 0°, and you ask to move to -20°, it goes counter-clockwise, while if you ask to go to 340°, it would go to the same place but clockwise. At the end of the move it would always indicate 340°. However, to fully support this feature you'd need a range from -360° to 720°, as to support also going from 340° to 10° clockwise (this is what we actually do on some axes in Odemis). 
-360° to 360° seems too small to be useful to me. 

So maybe it does nothing special. In that case, a simpler way to handle it in Odemis is in the init, to check explicitly for the range -360° -> 360°, and replace it with just 0->2π. (Note that it's almost the same that we do in the xt_client, because there the SEM reports a range of 0->0 for axes that support full rotation).
If you do the init this way, then here you can more safely write: `(pos[axis] - rng[0]) % (2 * math.pi) + rng[0]`

sneepstefan: Allright thanks for sparring a bit over this matter, I chose the second option of your suggestion. Specifically because Tescan Essence reports values of the rz axis only in the 0 -> 360° range as well so I think we will be safe with this choice.

src/odemis/driver/tescan.py
```diff
@@ -1034,25 +1034,42 @@ def _waitSync(self):
...
+
+        for num, ax in enumerate(["x", "y", "z", "rz", "rx"]):
+            # axes ranges are always returned in pairs of 2 values (min, max)
+            if len(axes_rng) >= 2:
+                # if an axis is not motorized do not add it to the axes definition
+                if axes_motorized[num]:
+                    # make a distinction between linear axes and rotational ones
+                    # rotational axes will need their range converted to radians
+                    if ax.startswith("r"):
+                        axes_rng[0] = 0 if axes_rng[0] == -360 and axes_rng[1] == 360 else axes_rng[0]
```
pieleric, line 1055: I think this would be more readable (but untested!):
```suggestion
                        if axes_rng[:2] == [-360, 360]:  # Full rotation: report as 0 -> 360°
                             axes_rng[0] = 0
```

src/odemis/driver/test/tescan_test.py
```diff
@@ -299,10 +484,12 @@ def test_move_abs(self):
         time.sleep(6)
         testing.assert_pos_almost_equal(self.stage.position.value, p)
 
+    # @skip("not working with the newer simulator Tescan Essence v1.2.2")
     def test_stop(self):
         """
         Check it's possible to stop the stage while it's moving.
         """
+        # TODO: make this method work again with Tescan Essence v1.2.2.
```
pieleric, line 492: Let's drop this TODO if this now works?

pieleric (changes requested):

src/odemis/driver/test/tescan_test.py
sneepstefan: I have created a task for this in the sw maintenance board to address this.
https://delmic.atlassian.net/browse/SWM-89?atlOrigin=eyJpIjoiYTM0Y2RhMzg0NWE4NGNjMjk4NDMxZTdlNTRkOTBjNTciLCJwIjoiaiJ9

PR #2471, opened 2023-06-29
Fix simstreakcamtest andorcam2 test

tepals:


src/odemis/driver/test/simstreakcam_test.py
```diff
@@ -60,7 +58,8 @@ class TestSimStreakCamGenericCam(VirtualTestCam, unittest.TestCase):
     camera_type = CLASS_STREAKCAM
     camera_kwargs = KWARGS_STREAKCAM
 
-    def setUp(self):
```
nandishjpatel, line 63: Wouldn't just adding the line in `def setUp(self):` work?

`super().setUp()`

tepals: yes, but there is not really a reason to have this in the setUp, or is there? The advantage of having it in the setUpClass is that it only gets called once.

nandishjpatel (approved):

src/odemis/driver/test/simstreakcam_test.py
```diff
@@ -69,9 +68,6 @@ def setUp(self):
             "ignore", category=RuntimeWarning, message=re.escape("numpy.ndarray size changed")
         )
 
-    @classmethod
-    def setUpClass(cls):
-
         super(TestSimStreakCamGenericCam, cls).setUpClass()
```
pieleric, line 71: Typically, it's best to call super as the first line, unless there is good reason not too (I don't think there is).

src/odemis/driver/test/andorcam2_test.py
```diff
@@ -78,24 +76,26 @@ class TestFake(VirtualTestCam, unittest.TestCase):
...
+    @classmethod
+    def setUpClass(cls):
         # Ignore RuntimeWarning: numpy.ndarray size changed, may indicate binary incompatibility.
         # Expected 80 from C header, got 88 from PyObject
         # This warning is not caused by the code explicitly changing the array size but rather
         # by an inconsistency between different versions of NumPy.
         warnings.filterwarnings(
             "ignore", category=RuntimeWarning, message=re.escape("numpy.ndarray size changed")
         )
+        cls.camera = cls.camera_type(**cls.camera_kwargs)
```
pieleric, line 88: That's what super() does, right? So just replace this line by a call to `super()`. (and preferably place it as first line of the function).

pieleric (changes requested):

PR #2474, opened 2023-06-30
[FIX] millmng_test: test_milling_settings_from_yaml updated to actual…

tepals:
…ly test milling

There was a sleep of 13 seconds before checking if the future was running, this got removed in commit 9b2c12cb, because after 13 seconds the future was already done running. It turns out that the future was done running because the milling had failed, because the stage was not in loading position. This fix moves the stage to loading position and puts back a sleep of 0.1s to be sure the future is running at the check.

### DIFF
--- a/src/odemis/acq/test/millmng_test.py
+++ b/src/odemis/acq/test/millmng_test.py
@@ -340,6 +340,10 @@ class MillingManagerTestCase(unittest.TestCase):
         self.updates = 0
         self.done = False
 
+        # Milling does not start from an unknown stage position
+        # Set the stage to loading position
+        cryoSwitchSamplePosition(LOADING).result()
+
         milling_setting_1 = MillingSettings(name='rough_milling_1', current=self.probe_current, horizontal_fov=35e-6,
                                             roi=(0.5, 0.5, 0.8, 0.8),
                                             pixel_size=(3.5e-08, 3.5e-08), beam_angle=self.beam_angle,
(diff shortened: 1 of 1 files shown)


K4rishma (approved):

FCanberk (approved):

src/odemis/acq/test/millmng_test.py
```diff
@@ -405,11 +405,15 @@ def test_milling_settings_from_yaml(self):
...
+        # Milling does not start from an unknown stage position
+        # Set the stage to loading position
+        cryoSwitchSamplePosition(LOADING).result()
+
         millings = load_config(MILLING_CONFIG_PATH + "/milling-series-test-working-config.mill.yaml")
         logging.debug(millings)
         f = mill_features(millings, self.sites, self.feature_post_status, self.acq_streams, self.ion_beam,
                           self.sed, self.stage, self.aligner)
-
+        time.sleep(0.1)
```
pieleric, line 416: I would say that you could increase a bit more, to at least 1s, to make sure it's really running.

src/odemis/acq/test/millmng_test.py
```diff
@@ -405,11 +405,15 @@ def test_milling_settings_from_yaml(self):
         """
         Test mill features function using the milling settings from a YAML file
         """
+        # Milling does not start from an unknown stage position
```
pieleric, line 408: test_cancel() should also be adjusted, as there also used to be a sleep there.

pieleric (changes requested):

src/odemis/acq/test/millmng_test.py
tepals: I understand where you are coming from, but it also feels like we are testing the wrong thing if 1s is good, but 0.1s is not good. Is there another way we can test whether the milling has actually started, for instance can we check that a certain setting changed? In addition to checking `f.running()`.

pieleric: The problem is that in the simulator, we use `FakeOrsayMilling`. This class is very simple. It essentially just does a `sleep()`. Hence, there is not much else to check.

PR #2476, opened 2023-07-04
Fix developer manual compilation for bionic

nandishjpatel:


debian/changelog
pieleric, line 1: Please keep the changelog changes separated. These changes should only go into the v3.3-branch.

pieleric (changes requested):
Could you update the commit description to indicate that the ligature characters failed only on Ubuntu 18.04, that's why we never noticed.

pieleric (approved):

sneepstefan (approved):

tepals (approved):

tepals:
I saw the characters got added by @sneepstefan do we know how they got added, so we can avoid this in the future?

sneepstefan:
@tepals I am not sure we have to worry about it not happening in the future. I think it got in there because I copied the lines from the SPARC Manual from the default reader into the .rst file directly. We could however avoid doing these copy/paste actions or at least after copying from somewhere clear the text formatting before pasting.
In principle I didn't even notice the slight change in text format regarding those words.

pieleric:
>  I am not sure we have to worry about it not happening in the future. I think it got in there because I copied the lines from the SPARC Manual from the default reader into the .rst file directly. We could however avoid doing these copy/paste actions or at least after copying from somewhere clear the text formatting before pasting. In principle I didn't even notice the slight change in text format regarding those words.

Good to know where it comes from: copy-pasting the PDF. That means that if we type directly in a text editor, it shouldn't happen. Next time someone uses copy-paste, we should be on the lookout for such character. @tepals , they are relatively easy to spot. I actually had seen them in the PR from Stefan, but as it compiled fine on 20.04, I assumed it was fine.

PR #2478, opened 2023-07-04
added physical position in addition to pixel position for the PixelVa…

nandishjpatel:
…lueOverlay

### DIFF
--- a/src/odemis/gui/comp/overlay/view.py
+++ b/src/odemis/gui/comp/overlay/view.py
@@ -312,10 +312,11 @@ class PixelValueOverlay(ViewOverlay):
 
         # a rough estimation of the width and height margins in order to get the position of the pixel raw value display
         margin_w, margin_h = 5, 20
-        for stream in streams:
-            if self._v_pos:
-                self._p_pos = self.cnvs.view_to_phys(self._v_pos, self.cnvs.get_half_buffer_size())
 
+        if self._v_pos:
+            self._p_pos = self.cnvs.view_to_phys(self._v_pos, self.cnvs.get_half_buffer_size())
+            self._label.colour = self.colour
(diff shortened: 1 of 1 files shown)


src/odemis/gui/comp/overlay/view.py
```diff
@@ -343,6 +343,14 @@ def draw(self, ctx):
                     margin_r = 5
                     margin_h += (self._label.text_size[1] + margin_r)
 
+            # Display physical position at the top
+            pos = units.readable_str(self._p_pos, "m", sig=6)
+            text = f"Position: {pos}"
+            view_pos = self.view_width - margin_w, self.view_height - margin_h
+            self._label.pos = Vec(view_pos[0], view_pos[1])
+            self._label.text = text
+            self._label.align = wx.ALIGN_RIGHT
```
pieleric, line 352: This line is actually also always the same. So could you move it with the colour, line 319? (and also drop the line 338).

pieleric (changes requested):

src/odemis/gui/comp/overlay/view.py
```diff
@@ -312,10 +312,11 @@ def draw(self, ctx):
 
         # a rough estimation of the width and height margins in order to get the position of the pixel raw value display
         margin_w, margin_h = 5, 20
-        for stream in streams:
-            if self._v_pos:
-                self._p_pos = self.cnvs.view_to_phys(self._v_pos, self.cnvs.get_half_buffer_size())
 
+        if self._v_pos:
```
nandishjpatel, line 316:
```suggestion
        if self.v_pos:
```

K4rishma (approved):

pieleric (approved):

PR #2479, opened 2023-07-07
[fix] Convert float to int type in numpy.linspace

K4rishma:
6Z /home/testing/.local/lib/python3.8/site-packages/numpy/core/function_base.py:120: in linspace num = operator.index(num) TypeError: 'numpy.float64' object cannot be interpreted as an integer

### DIFF
--- a/src/odemis/util/linalg.py
+++ b/src/odemis/util/linalg.py
@@ -43,6 +43,7 @@ CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
 ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
 THE POSSIBILITY OF SUCH DAMAGE.
 """
+import math
 from typing import Iterable, Tuple
 
 import numpy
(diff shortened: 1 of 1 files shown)


src/odemis/util/linalg.py
```diff
@@ -252,8 +252,8 @@ def generate_triangulation_points(max_dist: float, area_coords: Iterable[float])
     # away from the border of the given area
     length_x = abs(xmax - xmin)
     length_y = abs(ymax - ymin)
-    points_x = numpy.floor(length_x / max_dist) + 1
-    points_y = numpy.floor(length_y / max_dist) + 1
```
sneepstefan, line 256: Maybe add a small comment as to what issue rises when float values are used instead of integers

K4rishma: in line 267, numpy.linspace expects the third input argument to be int. should I add a comment to explain it?

sneepstefan: Ah no I think that is pretty straightforward 'as is'. You can ignore my comment.

sneepstefan (approved):

pieleric (changes requested):

tepals (approved):

PR #2482, opened 2023-07-13
[CSDM-1059] add tabs enabler plugin

FCanberk:
Add plugin to enable tabs and add the function to enable alignment tab on enzel, sparc, secom and mimas through "Enable alignment tab" button under Help > Development on the top bar.

https://delmic.atlassian.net/browse/CSDM-1059

### DIFF
--- a/plugins/tabs_enabler.py
+++ b/plugins/tabs_enabler.py
@@ -1,0 +1,50 @@
+# -*- coding: utf-8 -*-
+"""
+Created on 10 July 2023
+
+@author: Canberk Akın
+
+Gives ability to enable the disabled tabs via a button under Help > Development.
+
+Copyright © 2023 Canberk Akın, Delmic
+
+This file is part of Odemis.
+
+Odemis is free software: you can redistribute it and/or modify it under the terms of the GNU
+General Public License version 2 as published by the Free Software Foundation.
+
+Odemis is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even
+the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General
+Public License for more details.
+
+You should have received a copy of the GNU General Public License along with Odemis. If not,
+see http://www.gnu.org/licenses/.
+"""
+
+from odemis.gui.plugin import Plugin
+
+class TabEnablerPlugin(Plugin):
+    name = "Tab Enabler"
+    __version__ = "1.0"
+    __author__ = "Canberk Akın"
+    __license__ = "Public domain"
+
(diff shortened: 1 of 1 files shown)


plugins/tabs_enabler.py
```diff
@@ -0,0 +1,47 @@
+# -*- coding: utf-8 -*-
+"""
+Created on 10 April 2017
```
tepals, line 3: that's a long time ago :p

plugins/tabs_enabler.py
```diff
@@ -0,0 +1,47 @@
...
+
+@author: Canberk Akın
+
+Gives ability to automatically place a EM image so that it's aligned with
+another one (already present).
```
tepals, line 8: I think this needs to be updated

tepals (changes requested):

plugins/tabs_enabler.py
```diff
@@ -0,0 +1,46 @@
...
+You should have received a copy of the GNU General Public License along with Odemis. If not,
+see http://www.gnu.org/licenses/.
+"""
+
+from odemis.gui.plugin import Plugin
+
+class TabEnablerPlugin(Plugin):
+    name = "Tab Enabler"
+    __version__ = "1.0"
+    __author__ = u"Canberk Akın"
```
pieleric, line 29: You can (now) drop the `u` from the string.

plugins/tabs_enabler.py
```diff
@@ -0,0 +1,46 @@
...
+    __license__ = "Public domain"
+
+    def __init__(self, microscope, main_app):
+        super(TabEnablerPlugin, self).__init__(microscope, main_app)
```
pieleric, line 33: Now, you can just write:
```suggestion
        super().__init__(microscope, main_app)
```

plugins/tabs_enabler.py
```diff
@@ -0,0 +1,46 @@
...
+        self.addMenu("Help/Development/Enable alignment tab", self.enable_alignment_tab)
```
pieleric, line 34: Please only add this menu if it's possible to click this option. Especially, the viewer (which is also available on all the computers, and doesn't connect to the back-end). The simplest way is probably to do the same check as in `enable_alignment_tab()`: check whether `btn_tab_align_enzel` or `btn_tab_align` are shown.

pieleric (changes requested):

pieleric:
Could you please follow the new convention for naming the PRs, and add the Jira task ID in the PR name.

pieleric (approved):

tepals (approved):

PR #2485, opened 2023-07-20
[config] Add Zeiss API connection test microscope file

pieleric:
Almost all SEM APIs have a corresponding test microsope file, which only
uses the SEM. The Zeiss didn't have one yet.

### DIFF
--- a/install/linux/usr/share/odemis/hwtest/zeiss-only.odm.yaml
+++ b/install/linux/usr/share/odemis/hwtest/zeiss-only.odm.yaml
@@ -1,0 +1,119 @@
+# For Zeiss Gemini SEM
+# Connection via analog for e-beam scan & serial port for e-beam column settings
+SEM: {
+    class: Microscope,
+    role: sem,
+    children: ["SEM E-beam full", "EBeam Focus", "Sample Stage",
+               "SEM Detector"],
+}
+
+"SEM E-beam full": {
+    class: scanner.CompositedScanner,
+    role: e-beam,
+    dependencies: {
+        external: "SEM E-beam",
+        internal: "EBeam control"
+    },
+#    children: {
+#        "detector": "SEM Detector full",
+#    },
+    init: {},
+    properties: {
+        scale: [8, 8], # (ratio) : start with a pretty fast scan
+        dwellTime: 10.e-6, # s
+    },
+    affects: ["SEM Detector"]
+}
+
+# To use the "external" control from the API instead of the analog
+# Merges the analog and digital detectors (zeiss one and the one from the DAQ board), to provide an "external" signal via the API
+#"SEM Detector full": {
+#    role: se-detector,
+#    dependencies: {
+#        external: "SEM Detector",
+#    },
+#    init: {},
+#}
+
+"SEM Scan Interface": {
+    class: semcomedi.SEMComedi,
+    role: null,
+    init: {device: "/dev/comedi0"},
+    # more detectors can be added, if necessary
+    children: {
+        scanner: "SEM E-beam",
+        detector0: "SEM Detector",
+    }
+}
+
+# Connect:
+# X -> AO 0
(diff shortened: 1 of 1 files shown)


FCanberk (approved):

install/linux/usr/share/odemis/hwtest/zeiss-only.odm.yaml
```diff
@@ -0,0 +1,127 @@
...
+
+"EBeam Focus": {
+    role: ebeam-focus,
+}
+
+"Sample Stage": {
+    role: stage,
+    affects: ["SEM E-beam full"],
+    init: {
+        inverted: ["x", "y"], # TODO: z ??
```
sneepstefan, line 123: Does this todo still need to be in?

pieleric: I guess with the METEOR work, we should now know whether the Z goes up or down. Normally the convention in Odemis is that an increase in Z causes the sample to be closer to the e-beam source. 
Do you know?

sneepstefan: Good question we have to check with Jochem and Karisma for that.

sneepstefan: z is positive upwards I just heard from Karishma so I guess we can leave inversion of z out.

sneepstefan, line 123:
```suggestion
        inverted: ["x", "y"],
```

pieleric:
@sneepstefan , I fixed the comment about Z, and many other comments. Could you give a second look?

install/linux/usr/share/odemis/hwtest/zeiss-only.odm.yaml
```diff
@@ -0,0 +1,119 @@
...
+
+# Connect:
+# X -> AO 0
+# Y -> AO 1
+# Ground -> AO GND
+"SEM E-beam": {
+    role: null,
+    init: {
+        channels: [0, 1],
+        # On Delmic scanning box v2, the voltage is x2, so need to specify twice smaller values than needed.
```
sneepstefan, line 57: This comment does not seem to be belong to the next line?

install/linux/usr/share/odemis/hwtest/zeiss-only.odm.yaml
```diff
@@ -0,0 +1,119 @@
...
+        # 1 = Open drain output (Y0.0)
+        # 2 = Digital Out 1
+        # 3 = Digital Out 0
+        # 4 = Status led
+        # output ports -> True (indicate scanning) or False (indicate parked)
+        scanning_ttl: {
+            3: [True, "external"],
+            4: True
+        },
+        settle_time: 120.e-6, # s
```
sneepstefan, line 72: 0.12 ms sounds extremely low.. not sure if this is usual

install/linux/usr/share/odemis/hwtest/zeiss-only.odm.yaml
pieleric: Good point.

install/linux/usr/share/odemis/hwtest/zeiss-only.odm.yaml
pieleric: That's the time needed for the flyback after scanning a whole line. With short dwell time of 1µs, which is relatively common, that's equivalent to 120 pixels. So, no, it's not really small. Actually some SEM even supports settle times as small as 10µs.

pieleric:
@sneepstefan , I've fixed the last comment. Do you mind to do one more review?

sneepstefan (approved):

PR #2486, opened 2023-07-20
[CSDM-1068] acq calibration: if multiple background data, average should be the same dtype as raw data

pieleric:
If a background data file contains multiple data, now, the average is
returned (it used to return just the first data). However, numpy.mean()
automatically uses float64 when averaging integers. That clashes with
the rest of the background subtraction code which expect the background
data type to be the same as the actual data.

So in such case, the user would get such error message:
"Incompatible encoding of background data float64 with the angular resolved encoding uint16."

=> convert the average value back to the original data type. It's a tiny
bit less precise, but that's typically really not important as the data
has a large range of values. (and it's still better than returning just
the first background data)

### DIFF
--- a/src/odemis/acq/test/calibration_test.py
+++ b/src/odemis/acq/test/calibration_test.py
@@ -93,13 +93,14 @@ class TestAR(unittest.TestCase):
              model.MD_LENS_MAG: 60,  # ratio
             }
         calib = model.DataArray(dcalib, md)
-        calib2 = model.DataArray(dcalib + 1, md)
+        calib2 = model.DataArray(dcalib + 10, md)
 
         # Give one DA, the correct one, so expect to get it back
         out = calibration.get_ar_data([calib, calib2])
         # The average (of calib=0 and calib2=1)
+        self.assertEqual(calib.dtype, out.dtype)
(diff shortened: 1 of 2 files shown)


tepals (approved):

src/odemis/acq/test/calibration_test.py
sneepstefan, line 1: I cannot completely catch how the 10-fold change is related to the dtype change. Is this to prevent a forced float dtype?

pieleric: Do you mean `dcalib + 1` -> `dcalib + 10`?
This change in the test is necessary indeed because of the dtype. The test code meant that calibration would average between 0  and 1. So the excepted result was 0.5. That works fine if it's a float. However, now that the dtype is an int, that doesn't work anymore (because 0.5 would be rounded down to 0)!
So, to test that now, I've changed the code to force the calibration to average between 0 and 10. The result is 5, which fits in an int.

sneepstefan: Perfectly clear, thanks for that

sneepstefan (approved):

PR #2491, opened 2023-07-26
[CSDM-730] Compensate chromatic aberration of angular spectrum data

pieleric:
The angular spectrum raw data is typically not straight because of
chromatic aberration. That means the angular values actually (slightly)
varies depending the on the wavelength.
The aberration is already measured at alignment, and stored in the metadata.
However, so far the GUI did nothing with this information. The angles
displayed corresponded to the angles at the center wavelength.

This adds support for compensation, by adjusting the data for each
wavelength independently. This is done just once when loading the data
(and every time a background or coefficient compensation file is
loaded).
This also indirectly means that the 2D view angle vs wavelength is now
linear on the angle axis (instead of being some sort of logarithmic
scale). That makes the graph easier to read.

### DIFF
--- a/src/odemis/acq/stream/_static.py
+++ b/src/odemis/acq/stream/_static.py
@@ -621,15 +621,16 @@ class StaticSpectrumStream(StaticStream):
             # cached list of angle or timestamps for each position in the second dimension
             if dims[1] == "A":
                 theta_list, unit_theta = spectrum.get_angle_range(image)
-                # TODO: handle case where theta list is only NAN
-                # Only keep valid values (ie, not the NaN)
-                # Note, the .calibrated data will have the same columns removed
-                self._thetal_px_values = numpy.array([theta for theta in theta_list if not math.isnan(theta)])
-                min_theta, max_theta = min(self._thetal_px_values), max(self._thetal_px_values)
+                # Only keep min/max (while discarding the NaNs)
+                # Note, the .calibrated data will have (almost) the same min/max
+                # but with values differently (regularly) distributed.
+                theta_list_np = numpy.array(theta_list)
+                min_theta, max_theta = numpy.nanmin(theta_list_np), numpy.nanmax(theta_list_np)
+                self._calibrated_theta_list = None  # To be filled when calibrated image is computed
 
                 # Allows to select the angle as any value within the range, and the
                 # setter will automatically "snap" it to the closest existing theta stamp
-                self.selected_angle = model.FloatContinuous(self._thetal_px_values[0],
+                self.selected_angle = model.FloatContinuous(min_theta,
                                                             range=(min_theta, max_theta),
                                                             unit=unit_theta,
                                                             setter=self._setAngle)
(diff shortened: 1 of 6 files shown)


src/odemis/acq/stream/_static.py
```diff
@@ -837,8 +838,19 @@ def _updateCalibratedData(self, data=None, bckg=None, coef=None):
             self.calibrated.value = None
             return
 
-        # If MD_THETA_LIST, the length of the A dimension might be reduced
         calibrated = calibration.apply_spectrum_corrections(data, bckg, coef)
+
+        # If angular spectrum, the length of the A dimension might have changed
+        if hasattr(self, "selected_angle"):  # update the list of angles
+            theta_list, _ = spectrum.get_angle_range(calibrated)
+            # Update the range, as it it might has slightly changed
```
nandishjpatel, line 846:
```suggestion
            # Update the range, as it might have slightly changed
```

nandishjpatel (approved):

src/odemis/util/angleres.py
```diff
@@ -619,25 +627,66 @@ def ExtractThetaList(data):
...
+    and the list of angles is derived.
+
+    :param wl > 0: The wavelength, in m, at which these theta values are computed.
+    :param line_bottom: px, px/mm, cf MD_AR_MIRROR_BOTTOM
+    :param line_top: px, px/mm, cf MD_AR_MIRROR_TOP
+    :param focus_dist: m, cf MD_AR_FOCUS_DISTANCE
+    :param x_max: m, cf MD_AR_XMAX
+    :param parabola_f: m, cf MD_AR_PARABOLA_F
+    :param length > 0: the number of angles returned
+    :returns:
```
K4rishma, line 681: same here

src/odemis/util/test/angleres_test.py
```diff
@@ -441,10 +440,85 @@ def test_uint16_input_rect2polar(self):
...
+        numpy.testing.assert_array_almost_equal(da_lin, 0)
+
+
 if __name__ == "__main__":
     # for debug:
     # import sys;sys.argv = ['', 'TestPolarConversionOutput.test_2000x2000']
     unittest.main()
     # suite = unittest.TestLoader().loadTestsFromTestCase(TestPolarConversionOutput)
     # unittest.TextTestRunner(verbosity=2).run(suite)
-
```
K4rishma, line 450: Should we omit an extra line at the end?

K4rishma (approved):

src/odemis/util/angleres.py
pieleric: Every function in this file is documented with `returns`. I guess it's automatically completed like that by PyCharm. I'm not going to update them. Just the functions I've touched.

src/odemis/util/test/angleres_test.py
pieleric: Yes, I think so. I think the rule is that the last line should have a `\n` at the end. This file had two `\n` at the end, showing up as one empty line. No need to have an empty line. Github will show a special read icon if the file is missing `\n` at the end, which is not the case here.

PR #2493, opened 2023-07-27
[feat] Update technolution driver to support ASM API v3

tepals:
Updates necessary to support the new features and depricated arguments of the ASM API v3.
* /scan/field get:Version 2 of this API returned a base64 encoded image, version 3 returns a binary TIFF.
* removed /scan/check_mega_field
* MegaFieldMetaData ** mega_field_id -> stack_id
** storage_directory updated -> with these new changes the storage directory will be: /asm_service/[user]/[Project-ID]/[ROA-ID]/[Zpos]_[DATE-TIME]/*.tiff *** since the date time is now automatically added on the ASM, adding it has been removed from the odemis code. ** info added
** z_position added
** pixel_size added
** y_prescan_lines removed
** x_ and y_eff_field_size added for stitching in post-processing

To support this the following metadata has been added:
* MD_USER = "Username"
* MD_SLICE_IDX = "Index of slice in volume stack"
* MD_FIELD_SIZE = "Average field of view of a megafield"

SDEV-1535


src/odemis/driver/technolution.py
```diff
@@ -47,6 +48,11 @@
 from technolution_asm.models.field_meta_data import FieldMetaData
 from technolution_asm.models.mega_field_meta_data import MegaFieldMetaData
 
+supported_version = "3.0.0"
+if StrictVersion(technolution_asm.__version__) < StrictVersion(supported_version):
+    raise ImportError(f"Version {technolution_asm.__version__} for technolution_asm not supported,"
+                      f"version {supported_version} or higher is expected")
```
nandishjpatel, line 54:
```suggestion
SUPPORTED_VERSION = "3.0.0"
if StrictVersion(technolution_asm.__version__) < StrictVersion(SUPPORTED_VERSION):
    raise ImportError(f"Version {technolution_asm.__version__} for technolution_asm not supported,"
                      f"version {SUPPORTED_VERSION} or higher is expected")
```

src/odemis/driver/technolution.py
```diff
@@ -1167,50 +1147,56 @@ def _assembleMegafieldMetadata(self):
...
-                # Reshape cell parameters such that the order of the cells is from left to right, then top to
-                # bottom. So the cells in the upper line are numbered 0..7, the bottom line 56..63.
-                cell_parameters=[CellParameters(translation[0], translation[1], darkOffset, digitalGain)
-                                 for translation, darkOffset, digitalGain in
-                                 zip(cellTranslation, cellDarkOffset, cellDigitalGain)],
-                sensor_over_voltage=self.overVoltage.value
-            )
+        md = self.getMetadata()
+        custom_data = md.get(model.MD_EXTRA_SETTINGS, "")  # if no custom metadata, pass an empty string
+        info = md.get(model.MD_USER_NOTE, None)  # if no user note, pass nothing. TODO support in GUI
```
nandishjpatel, line 1159: `MD_USER_NOTE` is of type string, see [here](https://github.com/delmic/odemis/blob/4fa33e8a91e8abd2c7463b45df31b28beb5b64d6/src/odemis/model/_metadata.py#L160)

src/odemis/acq/fastem.py
```diff
@@ -567,15 +570,17 @@ def run(self):
             self.pre_calibrate(self._pre_calibrations)
             self._roa.overlap = overlap_init  # set back the overlap to the initial value
 
+        # set the sub-directories (<user>/<project-name>/<roa-name>)
+        # FIXME use username from GUI when that is implemented
+        username = self._detector.getMetadata().get(model.MD_USER, "fastem-user")
```
pieleric, line 575: I don't see where it's defined in the driver. For now, nowhere? Or only in the microscope file? Is it just used to define the root directory of the filename? If so, please update the fastem microscope files examples. If it's only used in the GUI, why do you need to use a metadata on the detector?
The externalStorage argument already comes with a username key, used to define the FTP credentials. What's the difference?

src/odemis/driver/technolution.py
```diff
@@ -31,6 +30,7 @@
 import re
 import threading
 import time
+from distutils.version import StrictVersion
```
pieleric, line 33: So far we didn't have a dependency on distutils. Could we avoid it? It just one tiny function. You could either use `pkg_resources.parse_version()`, as in gui.util.updater, or do as in xt_client, and just "open code it" as ` [int(u) for u in s.split(".")]`.

src/odemis/driver/technolution.py
```diff
@@ -1167,50 +1147,56 @@ def _assembleMegafieldMetadata(self):
...
-                cell_parameters=[CellParameters(translation[0], translation[1], darkOffset, digitalGain)
-                                 for translation, darkOffset, digitalGain in
-                                 zip(cellTranslation, cellDarkOffset, cellDigitalGain)],
-                sensor_over_voltage=self.overVoltage.value
-            )
+        md = self.getMetadata()
+        custom_data = md.get(model.MD_EXTRA_SETTINGS, "")  # if no custom metadata, pass an empty string
+        info = md.get(model.MD_USER_NOTE, None)  # if no user note, pass nothing. TODO support in GUI
+        z_position = md.get(model.MD_SLICE_IDX, 0)  # if slice number is not provided use 0. TODO support in GUI
+        eff_field_size = md.get(model.MD_FIELD_SIZE, (6016, 6016))  # if not provided use 6% overlap
```
pieleric, line 1161: That default value only works if the full field is 6400px. I think it would be simpler to default to the effective field size being the same size as the full field. If that info is required to get a good final image, then we should really expect that the acquisition code always sets MD_FIELD_SIZE, and if not, we should notice it easily.

src/odemis/driver/technolution.py
```diff
@@ -1287,7 +1273,7 @@ def _acquire(self):
                                  str(DATA_CONTENT_TO_ASM[dataContent]).lower()),
                                 200, raw_response=True, stream=True)
                             resp.raw.decode_content = True  # handle spurious Content-Encoding
-                            img = Image.open(BytesIO(base64.b64decode(resp.raw.data)))
+                            img = Image.open(BytesIO(resp.raw.data))
```
pieleric, line 1276: Could you add a comment to indicate that it's expected to be a TIFF.

src/odemis/driver/technolution.py
```diff
@@ -1746,6 +1732,24 @@ def _updateFrameDuration(self, _):
...
+                if eff_field_size[0] % 32 <= 16:
+                    sug_field_size = eff_field_size[0] - eff_field_size[0] % 32
+                else:
+                    sug_field_size = eff_field_size[0] + 32 - eff_field_size[0] % 32
+                suggested_overlap = 1 - sug_field_size / self._scanner.resolution.value[0]
+                raise ValueError(
+                    f"Overlap must result in the value of the effective field size being a multiple of 32. "
+                    f"Effective field size is: {eff_field_size}, suggested overlap: {suggested_overlap * 100}%"
+                )
+        model.HwComponent.updateMetadata(self, md)
```
pieleric, line 1751: Please use super() for calling the method of a parent class. It handles better complex class hierarchy, and is easier to copy-paste.
```suggestion
        super().updateMetadata(md)
```

src/odemis/gui/cont/tabs.py
```diff
@@ -1824,15 +1824,28 @@ def __init__(self, name, button, panel, main_frame, main_data):
             calib_prefix="calib_2",
         )
 
+        if model.getMicroscope().name.lower().endswith("sim"):  # it's a simulator
```
pieleric, line 1827: The GUI already has a pointer to the microscope, let's use it:
```suggestion
        if main_data.microscope.name.lower().endswith("sim"):  # it's a simulator
```

pieleric (changes requested):

src/odemis/driver/technolution.py
tepals: We could avoid it, but distutils is in the standard library, what is the disadvantage of adding dependencies from the standard library?

src/odemis/driver/technolution.py
tepals: If we pass `info = None` nothing at all gets stored in the metadata.yaml file. If we pass `info=""` if does get stored in the metadata.yaml file. That's why I decided to pass `None`, I will update the comment to reflect this.

src/odemis/acq/fastem.py
tepals: The username of the externalStorage is the FTP username. The users want to be able to define a username for a certain project on the system, so that it gets stored as `/asm_service/[user]/[Project-ID]/[ROA-ID]/[Zpos][DATE-TIME]/*.tiff`. A single user can then find back all their projects in `/asm_service/[user]/`. 

I will update the fastem microscope files to include this

nandishjpatel (approved):

src/odemis/driver/technolution.py
pieleric: Oh... then in general it would be completely fine. However, I just saw in the doc that `distutils` is deprecated and will be removed in Python 3.12! So let's not depend on it now :confused:

src/odemis/driver/technolution.py
pieleric: You could. I don't think it's very necessary though. For updateMetadata, I think it should be pretty safe. Be careful because in general that syntax is not necessary exactly identical. If the class inherits from a class which itself inherits from a parent class, it can also be a way to skip the parent class and directly call the grand-parent class. So not every case can be replaced by a `super()`.

src/odemis/driver/technolution.py
tepals: Then I will replace it

tepals:
@nandishjpatel and @pieleric I realized I forgot to update the test cases for the new megafield metadata, I added that in the last commit. 
I also noticed the test cases of `odemis.driver.test.technolution_test.Test_ASMDataFlow` are failing, I checked and they were already failing before this PR. I will investigate it and try to fix it in a new Pr.

pieleric (approved):

tepals:
@pieleric The test cases are failing with the simulator. I agree we should try to run them regularly

PR #2498, opened 2023-08-10
[fix] METEOR-770: Make tiff files compatible with ImageJ

K4rishma:
https://delmic.atlassian.net/browse/METEOR-770?atlOrigin=eyJpIjoiYzBmMDM3YTdlZmZiNDU3YTk3Y2FiYmE3NjIxMmFhYTEiLCJwIjoiaiJ9

Make Odemis export functions for tiff and ome.tiff formats compatible with the widely used Life Sciences post processing software, ImageJ


scripts/imagej_libtiff.py
```diff
@@ -0,0 +1,110 @@
...
+    metadata_all = image_data[0].metadata
+    pixel_size_x = metadata_all['Pixel size'][0] / 1e-06
+    pixel_size_y = metadata_all['Pixel size'][1] / 1e-06
+    tif.SetField('Xresolution', pixel_size_x)
+    tif.SetField('Yresolution', pixel_size_y)
+
+    # Add ome data in image description
+    ometxt = _convertToOMEMD(image_data)
+    description = (f"ImageJ=1.11a\nimages={num_slices * num_channels}\nchannels={num_channels}\nslices={num_slices}\nunit=um\n"
+                   f"hyperstack=true\n")#mode=composite\n")
```
FCanberk, line 63:
```suggestion
                   f"hyperstack=true\n")  # mode=composite\n")
```

scripts/imagej_libtiff.py
```diff
@@ -0,0 +1,110 @@
...
+    tif.SetField('ImageDescription', description.encode('ascii')+ometxt)
```
FCanberk, line 64:
```suggestion
    tif.SetField('ImageDescription', description.encode('ascii') + ometxt)
```

FCanberk (changes requested):

src/odemis/dataio/tiff.py
```diff
@@ -1795,13 +1795,22 @@ def _saveAsMultiTiffLT(filename, ldata, thumbnail, compressed=True, multiple_fil
         # Only get the corresponding data for this file
         ldata = sorted_x[file_index][1]
 
+    # Extract ImageJ compatible Image description
```
pieleric, line 1798: Please place this in a dedicated function.

src/odemis/dataio/tiff.py
```diff
@@ -1795,13 +1795,22 @@ def _saveAsMultiTiffLT(filename, ldata, thumbnail, compressed=True, multiple_fil
...
+    num_channels = len(ldata)
+    try:
+        _, _, num_slices, height, width = ldata[0].shape
+    except:
+        num_slices = 1
```
pieleric, line 1803: That is very very rough. I would suggest to at least ensure that it's true (ie, that all the channels have the same shape.
At ImageJ metadata cannot describe all the type of data we support, you have (at least) two options:
* Describe only the first images for which this is correct
* Only put ImageJ metadata if it's completely correct

I'd go for the first option.
I'd suggest that in the code, you check the ZYX dimension of the first DataArray, and you add as "Channels" the other DataArray which have the same ZYX dimension. You could go more fancy and if there is a C dimension, you report it as channels.

src/odemis/dataio/tiff.py
```diff
@@ -1795,13 +1795,22 @@ def _saveAsMultiTiffLT(filename, ldata, thumbnail, compressed=True, multiple_fil
...
+    imagej_description = (f"ImageJ=1.11a\nimages={num_slices * num_channels}\nchannels={num_channels}\nslices="
```
pieleric, line 1804: I think it'd be nicer to read (and easier to extend) if you first make it a dict, and then convert it to a string. Something like that:
```
md = {
    "ImageJ": "1.11a",
    "images": num_slices * num_channels,
:
}
imagej_description = "\n".join(f"{k}={v}" for k, v in md) + "\n"
```

src/odemis/dataio/tiff.py
```diff
@@ -2359,9 +2368,14 @@ def _getOMEXML(self, tfile):
         tfile.SetDirectory(0)
         desc = tfile.GetField(T.TIFFTAG_IMAGEDESCRIPTION)
 
-        if (desc and ((desc.startswith(b"<?xml") and b"<ome " in desc.lower()) or
-                      desc[:4].lower() == b'<ome')):
+        if (desc and ((b"<?xml" in desc) and b"<ome " in desc.lower() or
+                desc[:4].lower() == b'<ome')):
```
pieleric, line 2372: The last part (` desc[:4].lower()`) would deserve more love. It used to say: either start with xml + has ome, or start with ome. Now it says has xml or has ome or start with ome.

I think you could start by converting desc to lower(), and then checking for a line starting with xml/ome + it should contain ome.

You can do that with a single regex match, which you use later to find out the index.

src/odemis/dataio/tiff.py
```diff
@@ -2359,9 +2368,14 @@ def _getOMEXML(self, tfile):
...
             try:
+                pattern = b'<?xml version="1.0" encoding="UTF-8"?>'
```
pieleric, line 2374: Be more flexible in the string you search. I think that just "<?xml" is enough. Maybe you could also be more strict, in checking that it's the first word of a line. Also, if it starts just with "<ome" that will not work...

pieleric (changes requested):

src/odemis/dataio/tiff.py
K4rishma: @pieleric That is true indeed, I have done slight changes to the code as below    
`    try:
        num_slices = ldata[0].shape[-3]
    except:
        num_slices = 1`
I am not able to gauge, in which cases will the above code will fail to incorporate your suggestion to improve this block.

src/odemis/dataio/tiff.py
```diff
@@ -2358,10 +2383,14 @@ def _getOMEXML(self, tfile):
...
         tfile.SetDirectory(0)
         desc = tfile.GetField(T.TIFFTAG_IMAGEDESCRIPTION)
-
-        if (desc and ((desc.startswith(b"<?xml") and b"<ome " in desc.lower()) or
-                      desc[:4].lower() == b'<ome')):
+        if (desc and ((b"<?xml" and b"<ome " in desc.lower() and not desc.lower().startswith(b"<ome ")) or
+                desc[:4].lower() == b'<ome')):
             try:
+                pattern = b'<?xml'
+                start_index = desc.find(pattern)
```
pieleric, line 2390: Did you try using a regex? I think it'd be more powerful and cleaner to just use a regex, to search for it, and then use the `if` if there is a match. This avoids parsing the description string 4 times.
Also, it should work if it starts directly with `<ome`. Please, also only match if it's at the beginning of a line.

src/odemis/dataio/test/tiff_test.py
pieleric, line 1: Please add a test case that checks the ImageJ metadata is present.

scripts/imagej_libtiff.py
pieleric, line 1: Honestly, this script is quite a mess. There quite a lot of assumptions that will only make it work if the input data is precisely a Z stack with data type uint16. There are also a few things that are different from the standard behaviour of Odemis, which is confusing.

We could make it more generic, and cleaner, but that's going to take quite some time. However I don't know if it's ever going to be used anyway. So I'd suggest to not add it to Odemis repo, and just add it to our `cryo-engineering` internal repo. Is that fine with you?

pieleric (changes requested):

src/odemis/dataio/tiff.py
```diff
@@ -1833,6 +1837,62 @@ def _saveAsMultiTiffLT(filename, ldata, thumbnail, compressed=True, multiple_fil
...
+    Create ImageJ compatible metadata that is added to the ImageDescription identifier for the tiff files.
+    :param ldata: (list of DataArray) list of 2D data of int or float. Should have at least one array
+    :return (str): metadata compatible with ImageJ
+    """
+    num_channels = 0
+    num_frames = 0  # Time frames
+    num_slices = 0  # zlevels
+
+    # Check the ZYX dimensions of the first DataArray
+    Y, X = ldata[0].shape[-2:]
```
pieleric, line 1851: uppercases are typically reserved for constants. I think here you could just write `res = ...`

src/odemis/dataio/tiff.py
```diff
@@ -1833,6 +1837,62 @@ def _saveAsMultiTiffLT(filename, ldata, thumbnail, compressed=True, multiple_fil
...
+    if len(ldata[0].shape) >= 3:
+        Z = ldata[0].shape[-3]
+    else:
+        Z = 1
+
+    # Add to "channels" and "frames" the other DataArray which have the same ZYX dimension.
+    for n in range(len(ldata)):
```
pieleric, line 1858: You can write a little bit more directly `for data in ldata`. This way, instead of writing `ldata[n]`, you can just write `data`.

src/odemis/dataio/tiff.py
```diff
@@ -1833,6 +1837,62 @@ def _saveAsMultiTiffLT(filename, ldata, thumbnail, compressed=True, multiple_fil
...
+    :param ldata: (list of DataArray) list of 2D data of int or float. Should have at least one array
+    :return (str): metadata compatible with ImageJ
+    """
+    num_channels = 0
+    num_frames = 0  # Time frames
+    num_slices = 0  # zlevels
+
+    # Check the ZYX dimensions of the first DataArray
+    Y, X = ldata[0].shape[-2:]
+    if len(ldata[0].shape) >= 3:
```
pieleric, line 1852: Note that in python it's possible to have a short code that convert from tuple of length N to a tuple with a fixed length, filled with `1` at the beginning. I think something like: `(1,) * (5 - len(s)) + s`, but it should handle cases where len(s) > 5.

pieleric (changes requested):

src/odemis/dataio/test/tiff_test.py
```diff
@@ -459,8 +459,21 @@ def testMetadata(self):
 
         # check OME-TIFF metadata
         omemd = imo.GetField("ImageDescription").decode('utf-8')
-        self.assertTrue(omemd.startswith('<?xml') or omemd[:4].lower() == '<ome')
+        self.assertTrue(("<?xml" in omemd) or (omemd[:4].lower() == '<ome'))
```
FCanberk, line 462:
```suggestion
        self.assertTrue(("<?xml" in omemd) or (omemd[:4].lower() == "<ome"))
```

src/odemis/dataio/tiff.py
```diff
@@ -2358,15 +2402,23 @@ def _getOMEXML(self, tfile):
         # It's OME TIFF, if it has a valid ome-tiff XML in the first T.TIFFTAG_IMAGEDESCRIPTION
         tfile.SetDirectory(0)
         desc = tfile.GetField(T.TIFFTAG_IMAGEDESCRIPTION)
+        if desc:
+            desc_lower = desc.lower()
 
-        if (desc and ((desc.startswith(b"<?xml") and b"<ome " in desc.lower()) or
-                      desc[:4].lower() == b'<ome')):
+        if (desc and ((b"<?xml" in desc_lower and b"<ome " in desc_lower and not desc_lower.startswith(b"<ome ")) or
+                desc_lower[:4] == b'<ome')):
```
FCanberk, line 2409:
```suggestion
        if (desc and ((b"<?xml" in desc_lower and b"<ome " in desc_lower and not desc_lower.startswith(b"<ome ")) or
                desc_lower[:4] == b"<ome")):
```

src/odemis/dataio/tiff.py
```diff
@@ -1833,6 +1837,46 @@ def _saveAsMultiTiffLT(filename, ldata, thumbnail, compressed=True, multiple_fil
...
+    # Check the CTZYX dimensions of the first DataArray
+    size = ldata[0].shape
+    res = (1,) * (5 - len(size)) + size
+
+    # Add to "channels" if other DataArray(s) have the same TZYX dimension.
+    for data in ldata:
+        next_shape = data.shape
+        next_res = (1,) * (5 - len(next_shape)) + next_shape
+        if next_res == res:
+            if len(data.shape) == 5:
```
pieleric, line 1857: I think you should just use `next_res[-5]`, then no need to check that the shape is big enough.

src/odemis/dataio/tiff.py
```diff
@@ -1833,6 +1837,46 @@ def _saveAsMultiTiffLT(filename, ldata, thumbnail, compressed=True, multiple_fil
...
+
+    # Check the CTZYX dimensions of the first DataArray
+    size = ldata[0].shape
+    res = (1,) * (5 - len(size)) + size
+
+    # Add to "channels" if other DataArray(s) have the same TZYX dimension.
+    for data in ldata:
+        next_shape = data.shape
+        next_res = (1,) * (5 - len(next_shape)) + next_shape
+        if next_res == res:
```
pieleric, line 1856: IMHO, this is a little odd behaviour in case multiple images have shape C > 1, with the same C. 
For instance, if you have:
* [40, 1, 10, 300, 400]
* [40, 1, 10, 300, 400]

=> It will create a hyperstack of shape [80, 1, 10, 300, 400].

However if it has:
* [40, 1, 10, 300, 400]
* [50, 1, 10, 300, 400]

It will only keep the first images.

I think that you could either do:
* add all the C dimensions together. Example: 40 + 1 + 50...
* only use the first image if C > 1, and if C == 1, add them (as long as the whole shape is the same)

pieleric (changes requested):

pieleric (approved):

FCanberk (approved):

PR #2499, opened 2023-08-11
[SDEV-1553] move scintillator parameters from source code to microscope yaml file

nandishjpatel:
Ticket: https://delmic.atlassian.net/browse/SDEV-1553

The [scintillator parameters](https://github.com/delmic/odemis/blob/f4ed9ea0cae903984b13e0ad771d274cbb777aba/src/odemis/gui/model/__init__.py#L1398) in FastEMMainGUIData have been moved to stage's metadata by creating 2 additional metadata keys namely `MD_SAMPLE_SIZES` and `MD_SAMPLE_BACKGROUND`


src/odemis/gui/util/conversion.py
```diff
@@ -127,3 +130,29 @@ def change_brightness(colour, weight):
     new_fcol = tuple(f(c * (1 - weight) + lim * weight, lim) for c in _col[:3])
 
     return new_fcol + (_alpha,) if _alpha is not None else new_fcol
+
+
+def sample_positions_to_layout(sample_centers: Dict[str, Tuple[float, float]]) -> List[List[Optional[str]]]:
```
pieleric, line 135: Please remove the duplicate from gui.win.acquisition.py .

pieleric (changes requested):

src/odemis/gui/model/__init__.py
```diff
@@ -1433,24 +1402,47 @@ def __init__(self, microscope):
...
-        carrier_params = self.SAMPLE_CARRIER_3x3
+        # SAMPLE_CENTERS contains the center position of the scintillators from bottom-left
+        # position with a sample
+        if model.MD_SAMPLE_CENTERS not in md:
+            raise KeyError("Stage has no MD_SAMPLE_CENTERS metadata.")
+        centers = md[model.MD_SAMPLE_CENTERS]
+        layout = sample_positions_to_layout(centers)
+        for row_idx, row in enumerate(layout):
+            for column_idx, name in enumerate(row):
+                scintillator_number = int(name.split()[-1])
```
pieleric, line 1413: In case `int()` fails, please give a nice error message indicating that the name of the sample must have a number in second place.

src/odemis/gui/model/__init__.py
```diff
@@ -1433,24 +1402,47 @@ def __init__(self, microscope):
...
+        self.scintillator_sizes = sizes
+        self.scintillator_positions = {}
+        for name, center in centers.items():
+            scintillator_number = int(name.split()[-1])
+            self.scintillator_positions[scintillator_number] = (minx + center[0], miny + center[1])
+        self.scintillator_layout = layout
         self.background = []
-        for rect in carrier_params["background"]:
+        for rect in background:
             self.background.append((minx + rect[0], miny + rect[1], minx + rect[2], miny + rect[3]))
```
pieleric, line 1446: Please catch any error here and report as an error in  SAMPLE_BACKGROUND. One of the typical error that could happen is that a list is too short , or that a number is entered like `1e-6` and interpreted as a string. Such error coming from the GUI will be very hard to understand by the installation engineer if there is not a nice explanation!

pieleric (changes requested):

src/odemis/gui/model/__init__.py
```diff
@@ -1407,27 +1408,44 @@ def __init__(self, microscope):
...
-        for row_idx, row in enumerate(layout):
-            for column_idx, name in enumerate(row):
-                scintillator_number = int(name.split()[-1])
-                layout[row_idx][column_idx] = scintillator_number
         # SAMPLE_SIZES contains the sizes of the scintillators
         if model.MD_SAMPLE_SIZES not in md:
             raise KeyError("Stage has no MD_SAMPLE_SIZES metadata.")
         sample_sizes = md[model.MD_SAMPLE_SIZES]
+        # Handle error cases for sample centers and sizes
+        if set(centers.keys()) != set(sample_sizes.keys()):
```
pieleric, line 1416: .keys() are equivalent to sets. I think it's fine to just write:
`if centers.keys() != sample_sizes.keys():`

src/odemis/gui/model/__init__.py
```diff
@@ -1407,27 +1408,44 @@ def __init__(self, microscope):
...
+            raise KeyError("MD_SAMPLE_CENTERS and MD_SAMPLE_SIZES metadata should have "
+                "the same keys.")
+        if not any(isinstance(i, float) for i in chain.from_iterable(centers.values())):
```
pieleric, line 1419: That means "if there is not a single value which is float", then it's an error. I think you want to replace `not any` by `not all` (or `any(not isinstance...`

pieleric (changes requested):

src/odemis/gui/model/__init__.py
nandishjpatel: Good catch! fixed

pieleric (approved):

PR #2500, opened 2023-08-18
[feature] driver scanner: support multiple detectors

pieleric:


src/odemis/driver/scanner.py
```diff
@@ -45,8 +47,8 @@ def __init__(self, name, role, children=None, dependencies=None, daemon=None, **
 
         if children is None:
             children = {}
-        elif set(children.keys()) - {"detector"}:
-            raise ValueError("Only a detector child is supported")
+        # elif set(children.keys()) - {"detector"}:
```
K4rishma, line 50: Should not these lines be removed?

src/odemis/driver/scanner.py
```diff
@@ -101,10 +101,12 @@ def __init__(self, name, role, children=None, dependencies=None, daemon=None, **
         self._va_external_ctrl = None
         self._va_blanker_ctrl = None
 
-        if "detector" in children:
-            ckwargs = children["detector"]
-            self._detector = CompositedDetector(parent=self, daemon=daemon, **ckwargs)
-            self.children.value.add(self._detector)
+        for cname, ckwargs in children.items():
+            if not cname.startswith("detector"):
+                raise ValueError(f"Only children detector* are supported, but got '{cname}'")
```
sneepstefan, line 106:
```suggestion
                raise ValueError(f"Only children of type detector are supported, but got '{cname}'")
```
What exactly is the purpose of the * in this str?

src/odemis/driver/scanner.py
```diff
@@ -130,7 +132,7 @@ def __init__(self, name, role, children=None, dependencies=None, daemon=None, **
             else:
                 logging.debug("No external VA supported for the Composited Scanner.")
 
-        else:
+        if not self.children.value:  # If no automatic blanker/external, just duplicate the original ones
```
sneepstefan, line 135: This comment is not very fitting for this line could you move this just above (or behind) this line `va_names += ["blanker", "external"]`

src/odemis/driver/scanner.py
```diff
@@ -201,30 +209,36 @@ def updateMetadata(self, md):
     def getMetadata(self):
         return self._external_scanner.getMetadata()
 
-    def claimBeam(self, claim):
+    def claimBeam(self, claim: bool, user: object):
```
sneepstefan, line 212: I really appreciate this addition of parameter type, makes it much better readable and forces anyone to use the proper type upfront.

sneepstefan (changes requested):

src/odemis/driver/scanner.py
pieleric: It's to mean that the role doesn't have to be "detector" exactly. Anything that starts with "detector" is fine.
I'll try to phrase that in English... without being too verbose.

sneepstefan (approved):

PR #2501, opened 2023-08-18
[refactor] remove basestring usage

pieleric:
basestring was used for support of Python 2&3, to represent str and
unicode, and bytes.
Now that we only support Python 3, we can usually replace by just "str".
It has to be a little bit carefully done because in some cases that
meant "str or bytes".

### DIFF
--- a/src/odemis/gui/comp/settings.py
+++ b/src/odemis/gui/comp/settings.py
@@ -19,8 +19,6 @@ This file is part of Odemis.
     Odemis. If not, see http://www.gnu.org/licenses/.
 
 """
-from builtins import str
-from past.builtins import basestring
 import wx
 from decorator import decorator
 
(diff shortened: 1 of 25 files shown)


nandishjpatel (approved):

src/odemis/gui/comp/settings.py
```diff
@@ -19,11 +19,11 @@
     Odemis. If not, see http://www.gnu.org/licenses/.
 
 """
-from builtins import str
-from past.builtins import basestring
 import wx
 from decorator import decorator
 
+import logging
```
FCanberk, line 25: Is this import necessary?

pieleric: No. I only used it temporarily to test my changes. Good catch!

PR #2503, opened 2023-08-22
[feat] add autofocus verifier plugin

FCanberk:
A 'Verify autofocus' button is placed under development menu in the GUI, if the plugin is installed. 
The button runs a set of automated processes to speed up the FAT process.


FCanberk:
This plugin will be moved to hw-tester, so another PR will be created for it on bitbucket on another repo.

PR #2507, opened 2023-08-25
[fix] driver simstreakcam: avoid error when stopping the acquisition

pieleric:
In some cases, when stopping an acquisition, such error would happen:
ERROR   __init__:707:   Failure while calling repeating timer 'SimCam image generator'
  File "/home/piel/development/odemis/src/odemis/util/__init__.py", line 701, in run
    self.callback()
  File "/home/piel/development/odemis/src/odemis/util/weak.py", line 36, in __call__
    return self.f(ins, *arg, **kwargs)
  File "/home/piel/development/odemis/src/odemis/driver/simstreakcam.py", line 314, in _generate
    exp = timer.period
AttributeError: 'NoneType' object has no attribute 'period'

This is not a big deal, as it's in a thread that's supposed to end when
stopping the acquisition. So it'd be restarted automatically on next
acquisition.
Still, it's nicer to not fail => detect such corner case, and just end
the thread nicely.

### DIFF
--- a/src/odemis/driver/simstreakcam.py
+++ b/src/odemis/driver/simstreakcam.py
@@ -289,6 +289,10 @@ class ReadoutCamera(model.DigitalCamera):
         """
         Generates the fake output image based on the resolution.
         """
+        timer = self._generator  # might be replaced by None afterwards, so keep a copy
+        if timer is None:
+            return
+
         gen_img = self._getNewImage()
 
         # Processes the fake image based on resolution and binning.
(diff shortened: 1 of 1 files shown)


FCanberk (approved):

nandishjpatel (approved):

src/odemis/driver/simstreakcam.py
```diff
@@ -311,7 +314,7 @@ def _generate(self):
         self._mergeMetadata(metadata)
 
         # update fake output metadata
-        exp = timer.period
+        exp = timer.period# FIXME
```
pieleric, line 317: Remove comment

PR #2517, opened 2023-09-08
[perf] changes during zeiss-upenn testing for SEM-FM switching behavior

K4rishma:
During stage switching between SEM and FM, all the axes can move together safely.


src/odemis/acq/test/move_test.py
```diff
@@ -310,12 +311,18 @@ def setUpClass(cls):
```
pieleric, line 311: If I understand correctly, the idea is that the test case runs by default with a METEOR TFS-1, but if you've already started a simulator of a different METEOR that still works? It's strange, because `start_backend()` should detect this situation and automatically stop the current back-end and start the METEOR TFS-1 version.

I'd suggest to refactor this test case to have 2 different versions: one for TFS-1 and one for Zeiss-1. I think you can (again) use sub-classes to do that elegantly. You'd have a parent class which does the TFS-1, but it would use a few class constants like `.MIC_CONFIG` and `.ROTATION_AXES`, which can be overridden by the child classes. One of the child class would be the Zeiss-1 . It can also override some methods, to check more things on the Zeiss.

pieleric (changes requested):

src/odemis/acq/test/move_test.py
```diff
@@ -329,6 +336,51 @@ def test_moving_to_grid1_in_sem_imaging_area_after_loading_1st_method(self):
...
+        self.linked_ym_stage.moveRel({"y": 1e-3}).result()
+        new_stage_pos = self.stage.position.value
+        new_linked_ym_pos = self.linked_ym_stage.position.value
+
+        self.assertAlmostEqual(old_linked_ym_pos["y"] + 1e-3, new_linked_ym_pos["y"], places=5)
+        self.assertTrue(old_stage_pos["y"] < new_stage_pos["y"])
+
+        if self.stage_version == "zeiss_1":
+            # the stage moved in the right direction if the pre-tilt angle was maintained at -26-degrees
+            beta = 0.4537856055185  # -26-degrees in radians
```
sneepstefan, line 380: This is not -26° but +26°. And why not use math.radians for that? `beta = math.radians(26)`

src/odemis/acq/test/move_test.py
```diff
@@ -329,6 +336,51 @@ def test_moving_to_grid1_in_sem_imaging_area_after_loading_1st_method(self):
...
+            estimated_beta = math.atan2(new_stage_pos["m"] - old_stage_pos["m"], new_stage_pos["y"] - old_stage_pos["y"])
+            self.assertAlmostEqual(beta, estimated_beta, places=5, msg="The stage moved in the wrong direction in "
+                                                                       "the FM imaging grid 1 area.")
```
sneepstefan, line 383:
```suggestion
            estimated_beta = math.atan2(new_stage_pos["m"] - old_stage_pos["m"], 
                                        new_stage_pos["y"] - old_stage_pos["y"])
            self.assertAlmostEqual(beta, estimated_beta, places=5, 
                                   msg="The stage moved in the wrong direction in the FM imaging grid 1 area.")
```

src/odemis/acq/test/move_test.py
```diff
@@ -353,6 +412,14 @@ def test_moving_to_grid1_in_fm_imaging_area_after_loading(self):
...
         self.assertEqual(position_label, FM_IMAGING)
         self.assertEqual(grid_label, GRID_1)
+        # check the values of tilt and rotation
+        fm_angles = self.stage.getMetadata()[model.MD_FAV_FM_POS_ACTIVE]
+        if self.stage_version == "zeiss_1":
+            self.assertAlmostEqual(self.stage.position.value["rx"], fm_angles["rx"], places=5)
+            self.assertAlmostEqual(self.stage.position.value["rm"], fm_angles["rm"], places=5)
+        else:
+            self.assertAlmostEqual(self.stage.position.value["rx"], fm_angles["rx"], places=5)
+            self.assertAlmostEqual(self.stage.position.value["rz"], fm_angles["rz"], places=4)
```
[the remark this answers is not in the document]
K4rishma: Relevant rotational axes values in yaml file has 4 decimal places. In simulation the stage moves to the specified decimal places as in yaml file and the higher decimal places are given random non-zero values, I think. 

Now I also see the test cases for linear axes movement supporting until 4 decimal places.  

To tackle it, need to maybe modify the functions in tmcm.py in driver/ like _doMoveRel, etc. Is that correct @pieleric ?

src/odemis/acq/test/move_test.py
```diff
@@ -490,6 +556,44 @@ def test_unknown_label_at_initialization(self):
         self.assertEqual(current_grid, None)
 
 
+class TestMeteorZeiss1Move(TestMeteorMove):
+    """
+    """
```
pieleric, line 561: That docstring needs a little bit more words!

src/odemis/acq/test/move_test.py
```diff
@@ -490,6 +556,44 @@ def test_unknown_label_at_initialization(self):
...
+    MIC_CONFIG = METEOR_ZEISS1_CONFIG
+    ROTATION_AXES = {'rx', 'rm'}
+
+    def test_moving_in_grid1_fm_imaging_area_after_loading(self):
```
pieleric, line 565: If all the beginning is the same as the whole function of the parent class, you call it, instead of duplicating the code:
`super().test_...()`

src/odemis/acq/test/move_test.py
```diff
@@ -302,14 +304,17 @@ class TestMeteorMove(unittest.TestCase):
     """
```
pieleric, line 304: Could you rename the class to indicate it's for the TFS1?

pieleric (changes requested):
Looks pretty good. Just a few minor comments.

Please not that when merging I will want to keep the commits separated, as they are pretty much unrelated. So you will have to squash the "review" commits into the original commits. Also the title of each commit should be clear by itself.

src/odemis/acq/test/move_test.py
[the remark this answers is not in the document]
sneepstefan: Rx is also a rotational axis right? Why is that line also 5 places?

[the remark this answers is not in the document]
K4rishma: I have changed the code to check upto 4 places. It looks sufficient to check upto 4 places. Now it is uniform.

K4rishma:
@pieleric Now there are 4 separate commits

src/odemis/driver/actuator.py
```diff
@@ -666,6 +667,41 @@ def _get_rot_matrix(self, invert=False):
...
+            new_coordinates.append(vpos)
+
+        # Update default axes range and increase by buffer
+        # to account for movements at the extreme location
+        new_range = numpy.array(new_coordinates, dtype=float)
+        buff = 2.e-03
+        self.axes["x"].range = (min(new_range[:, 0] - buff),
+                                max(new_range[:, 0]) + buff)
+        self.axes["y"].range = (min(new_range[:, 1] - buff),
+                                max(new_range[:, 1] + buff))
```
pieleric, line 702: Could you please have all the `min/max(...) +/- buff` written the same? You currently have one with different parenthesis. I think you can always first compute the max/min and then do the buffer.

src/odemis/driver/actuator.py
```diff
@@ -666,6 +667,41 @@ def _get_rot_matrix(self, invert=False):
...
+        # Calculate range based on conversion matrix
+        new_coordinates = []
+        for val in coordinates:
+            vpos = self._convertPosFromdep(val)
+            new_coordinates.append(vpos)
+
+        # Update default axes range and increase by buffer
+        # to account for movements at the extreme location
+        new_range = numpy.array(new_coordinates, dtype=float)
+        buff = 2.e-03
```
pieleric, line 698: That's a quite arbitrary value! Why do you need extra room to handle movements near the end in the first place? Can you explain in the comment? 
Also, the value means that for now every axis get 4mm extra. Some hardware have a range of a couple of mm, so 4 mm is huge. Maybe add a percentage instead? (like 1%) Or can you completely remove this extra margin?

src/odemis/driver/actuator.py
```diff
@@ -666,6 +667,41 @@ def _get_rot_matrix(self, invert=False):
...
+        """
+        # Dep's range
+        range_y = self.axes["y"].range
+        range_x = self.axes["x"].range
+
+        # Calculate rectangular border points on the given range
+        min_x = range_x[0]
+        max_x = range_x[1]
+        min_y = range_y[0]
+        max_y = range_y[1]
```
pieleric, line 683: You can directly write:
```python
    min_x, max_y = self.axes["x"].range
    min_y, max_y = self.axes["y"].range
```

src/odemis/driver/test/actuator_test.py
```diff
@@ -548,6 +548,49 @@ def test_metadata(self):
 
         testing.assert_pos_almost_equal(stage.speed.value, {"x": 10e-6, "y": 20e-6})
 
+    def test_get_axes_range(self):
+        """
+        Test the range calculation in ConvertStage
+        """
+        # Test based on given range of Zeiss stage
+        dependency = simulated.Stage("stage", "test", axes=["y", "m"],
+                                     ranges={"y": [0.e-3, 100.e-3], "m": [0.e-3, 13.e-3]})
```
pieleric, line 557: I understand it's based on a real world example, however for the test case, we don't need all this extra info. You can keep it abstract and simple. So you can drop the comment that it's based on Zeiss, or that it's a "Meteor stage". Also, can you simplify and name the axes more simpler values, like a & b, or u & v?

src/odemis/driver/test/actuator_test.py
```diff
@@ -548,6 +548,49 @@ def test_metadata(self):
...
+        y_value = first_value * math.cos(tilt)
+        m_value = first_value * math.sin(tilt)
+        testing.assert_pos_almost_equal(dependency.position.value, {"y": y_value, "m": m_value})
+        # Move the stage in FM imaging range at different points
+        # and check if ConvertStage coordinates fall in between the ConvertStage range
+        try:
+            f = dependency.moveAbs({"y": 30.e-3, "m": 0.e-3})
+            f.result()
+        except:
+            self.fail('Riased unexpected exception')
```
pieleric, line 578: No need to catch every exception and convert them to a failure. The unitest framework already does this.

src/odemis/driver/test/actuator_test.py
```diff
@@ -548,6 +548,49 @@ def test_metadata(self):
...
+        # Test based on given range of Zeiss stage
+        dependency = simulated.Stage("stage", "test", axes=["y", "m"],
+                                     ranges={"y": [0.e-3, 100.e-3], "m": [0.e-3, 13.e-3]})
+
+        # Transform into Meteor stage , rotated by 26 degrees --> 0.4537856055185 rad
+        tilt = math.radians(26)
+        stage = ConvertStage("conv", "stage", {"orig": dependency},
+                             axes=["y", "m"], rotation=tilt)
+        first_value = 2e-03
+        second_value = 0
```
pieleric, line 564: Could you try to move to some places which are impossible within the "dependency" range. For instance, a negative value in x?

pieleric (changes requested):
I only have comments on the last commit. If you want you could open a separate PR for that commit, as the rest has been already reviewed and ready to merge. (in case it helps for making another PR).

src/odemis/driver/actuator.py
pieleric: Long story short, I think you can pass the range just as it is converted, without "buffer".

src/odemis/acq/test/move_test.py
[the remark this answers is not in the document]
sneepstefan: It would be nice if you could provide a small description how you came to the conclusion that it looks sufficient to check up to 4 places. I mean for the  rotational axis I completely get that as 5 is already measuring in a tenth of millidegrees in which the difference might not even be measured. But for linear axis which are measured in m 4 is measuring in a tenth of a mm which is in relative terms quite a lot right?

src/odemis/acq/test/move_test.py
```diff
@@ -490,6 +555,29 @@ def test_unknown_label_at_initialization(self):
...
+    def test_moving_in_grid1_fm_imaging_area_after_loading(self):
+        """Check if the stage moves in the right direction when moving in the fm imaging grid 1 area."""
+        super().test_moving_in_grid1_fm_imaging_area_after_loading()
+
+        # move in the same imaging mode using linked YM stage
+        old_stage_pos = self.stage.position.value
+        self.linked_stage.moveRel({"y": 1e-3}).result()
+        new_stage_pos = self.stage.position.value
+
+        # the stage moved in the right direction if the pre-tilt angle was maintained at -26-degrees
```
sneepstefan, line 574: -26 = 26? This comment look not aligned with the line of code below it.

src/odemis/driver/actuator.py
```diff
@@ -666,6 +667,41 @@ def _get_rot_matrix(self, invert=False):
             rotation *= -1
         return RigidTransform(rotation=rotation).matrix
 
+    def _getAxesRange(self):
+        """
+        Calculate the axes range for position VA based on the conversion matrix
+        from dep's positon.
```
sneepstefan, line 673: If I read this docstring it is not entiry clear how to interpret it, what is dep's position for example? And the conversion matrix?
There is also a typo in there: positon -> position

src/odemis/driver/actuator.py
```diff
@@ -666,6 +667,41 @@ def _get_rot_matrix(self, invert=False):
...
+        coordinates = [[min_x, min_y],
+                       [min_x, max_y],
+                       [max_x, min_y],
+                       [max_x, max_y]]
+
+        # Calculate range based on conversion matrix
+        new_coordinates = []
+        for val in coordinates:
+            vpos = self._convertPosFromdep(val)
+            new_coordinates.append(vpos)
```
sneepstefan, line 693:
```suggestion
        new_coordinates = list(map(self._convertPosFromdep, coordinates))
```

sneepstefan: nice little example of a one-liner! :)

src/odemis/driver/test/actuator_test.py
```diff
@@ -548,6 +548,49 @@ def test_metadata(self):
...
+        Test the range calculation in ConvertStage
+        """
+        # Test based on given range of Zeiss stage
+        dependency = simulated.Stage("stage", "test", axes=["y", "m"],
+                                     ranges={"y": [0.e-3, 100.e-3], "m": [0.e-3, 13.e-3]})
+
+        # Transform into Meteor stage , rotated by 26 degrees --> 0.4537856055185 rad
+        tilt = math.radians(26)
+        stage = ConvertStage("conv", "stage", {"orig": dependency},
+                             axes=["y", "m"], rotation=tilt)
```
sneepstefan, line 562:
```suggestion
        stage = ConvertStage("conv", "stage", {"orig": dependency},
                             axes=["y", "m"], rotation=math.radians(26))
```

src/odemis/driver/test/actuator_test.py
sneepstefan, line 564: Can you leave a short comment as to what these values are supposed to represent?

src/odemis/driver/test/actuator_test.py
sneepstefan, line 578:
```suggestion
            self.fail('Raised an unexpected exception')
```

src/odemis/driver/test/actuator_test.py
```diff
@@ -548,6 +548,49 @@ def test_metadata(self):
...
+            f = dependency.moveAbs({"y": 30.e-3, "m": 0.e-3})
+            f.result()
+        except:
+            self.fail('Riased unexpected exception')
+        # Move to maximum value in m
+        try:
+            f = dependency.moveAbs({"y": 30.e-3, "m": 13.e-3})
+            f.result()
+        except:
+            self.fail('Riased unexpected exception')
```
sneepstefan, line 584: same as previous comment

src/odemis/driver/actuator.py
K4rishma: Works well, Thank you

src/odemis/driver/actuator.py
K4rishma: I have updated the doc string.

src/odemis/acq/test/move_test.py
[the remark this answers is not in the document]
K4rishma: It is checking the rotational axes with 4 decimal places( you have provided good explanation). It is not checking directly the linear axes by using assert, hence there is no decimal check for linear axes. The linear axes movement is checked indirectly by checking if the linear position is in desired imaging mode (SEM, FM, LOADING) or not.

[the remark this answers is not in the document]
K4rishma: All the axes with prefix r, are rotational axes.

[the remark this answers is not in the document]
sneepstefan: I see and I think I got a little dragged onto this comment of yours, assuming there were also assert checks for positions for the linear axes:

> Now I also see the test cases for linear axes movement supporting until 4 decimal places.

So I guess for now its fine like it is done now. Thanks for you explanation.

src/odemis/driver/test/actuator_test.py
```diff
@@ -548,6 +548,45 @@ def test_metadata(self):
...
+        new_first_value = first_value * math.cos(tilt)
+        new_second_value = first_value * math.sin(tilt)
+        testing.assert_pos_almost_equal(dependency.position.value, {"a": new_first_value, "b": new_second_value})
+        # Move the stage in FM imaging range at different points
+        # and check if ConvertStage coordinates fall in between the ConvertStage range
+        f = dependency.moveAbs({"a": 30.e-3, "b": 0.e-3})
+        f.result()
+        # Move outside the given range of b, an error from ConvertStage must raise
+        with self.assertRaises(ValueError):
+            f = dependency.moveAbs({"a": 30.e-3, "b": 15.e-3})
```
pieleric, line 583: The comment says an error from the ConvertStage should come, but that cannot be, and indeed the code  doesn't get such error. The code only access the dependency and gets an error from the dependency.
I don't think it's useful to test this (for this test case about the range conversion). One thing you could test is to move the dependency to a=100mm, b=0, and check that the x/y position of the convert stage is still within range (by comparing the x to the range of x and y to the range of y).

src/odemis/driver/actuator.py
```diff
@@ -666,6 +667,27 @@ def _get_rot_matrix(self, invert=False):
...
+
+        # Calculate range based on the conversion
+        new_coordinates = list(map(self._convertPosFromdep, coordinates))
+
+        # Update default axes range
+        new_range = numpy.array(new_coordinates, dtype=float)
+        self.axes["x"].range = (min(new_range[:, 0]), max(new_range[:, 0]))
+        self.axes["y"].range = (min(new_range[:, 1]), max(new_range[:, 1]))
+
+
```
pieleric, line 690: Normally there is only 1 line between each method => delete one empty line.

src/odemis/driver/test/actuator_test.py
pieleric: No, that's still too easy, because first_value and second_value (which you could just call x_pos and y_pos) are also "legal" values as a & b in the dependency. Please use values which would be forbidden as a & b *but allowed* as x & y. For instance something like -2mm/10mm. (You need to check these values, I don't know if that's correct!).

src/odemis/driver/test/actuator_test.py
```diff
@@ -548,6 +548,45 @@ def test_metadata(self):
...
+            f.result()
+
+        # Reset the position of dependency
+        f = stage.moveAbs({"x": first_value, "y": second_value})
+        f.result()
```
pieleric, line 588: Why do you do that? What is it testing? Typically, it's always going to work, as you've just tried it a few lines before.

pieleric (changes requested):

src/odemis/driver/actuator.py
K4rishma: Sorry for this mistake

src/odemis/driver/test/actuator_test.py
pieleric: Ahhhhhhhh! So there is something wrong happening in the ConvertStage code! It changes the range of the dependency, which it should not. I guess you can add another check: check that the dependency range is as you specified at init.

src/odemis/driver/test/actuator_test.py
K4rishma: This is no longer required. I will remove it. I used it when I was running the test repeatedly and using moveRel instead of moveAbs. In that case, the absolute stage position was changing, making it difficult to debug.

src/odemis/driver/actuator.py
```diff
@@ -666,6 +667,27 @@ def _get_rot_matrix(self, invert=False):
             rotation *= -1
         return RigidTransform(rotation=rotation).matrix
 
+    def _getAxesRange(self):
```
pieleric, line 670: Actually, it's not a really good name, because by convention functions that are called `get*` are supposed to return a value... and not change anything else. I'd suggest to call it `_updateAxesRange`.

src/odemis/driver/actuator.py
```diff
@@ -666,6 +667,27 @@ def _get_rot_matrix(self, invert=False):
...
+                       [max_x, min_y],
+                       [max_x, max_y]]
+
+        # Calculate range based on the conversion
+        new_coordinates = list(map(self._convertPosFromdep, coordinates))
+
+        # Update default axes range
+        new_range = numpy.array(new_coordinates, dtype=float)
+        self.axes["x"].range = (min(new_range[:, 0]), max(new_range[:, 0]))
+        self.axes["y"].range = (min(new_range[:, 1]), max(new_range[:, 1]))
```
pieleric, line 688: I think this is modifying the original range too because of the definition l 630. You need to make a copy. 
Also, the comment line 627 can now be removed, as you've implemented this TODO.

pieleric (changes requested):

src/odemis/driver/test/actuator_test.py
K4rishma: > Could you try to move to some places which are impossible within the "dependency" range. For instance, a negative value in x?

I performed other test to check movements outside the a/b range. It starts on line 568.

src/odemis/driver/actuator.py
K4rishma: I ll make a copy, this was one of my main worries that it directly affects the dependency and alters its original range.

src/odemis/driver/actuator.py
K4rishma: I have renamed the function.

src/odemis/driver/actuator.py
K4rishma: image
The first line of the comment has been done, I am not sure what the second and third line means and if needs to be still executed.

sneepstefan (approved):

src/odemis/driver/test/actuator_test.py
K4rishma: Added the test 

> No, that's still too easy, because first_value and second_value (which you could just call x_pos and y_pos) are also "legal" values as a & b in the dependency. Please use values which would be forbidden as a & b _but allowed_ as x & y. For instance something like -2mm/10mm. (You need to check these values, I don't know if that's correct!).

src/odemis/driver/test/actuator_test.py
```diff
@@ -548,6 +546,67 @@ def test_metadata(self):
 
         testing.assert_pos_almost_equal(stage.speed.value, {"x": 10e-6, "y": 20e-6})
 
+    def test_update_axes_range(self):
+        """
+        Test the range calculation in ConvertStage
+        """
+        [a_min, a_max] = [0.e-3, 100.e-3]
+        [b_min, b_max] = [0.e-3, 13.e-3]
```
pieleric, line 554: This is more "pythonic":
```suggestion
        a_min, a_max = (0.e-3, 100.e-3)
        b_min, b_max = (0.e-3, 13.e-3
```

src/odemis/driver/test/actuator_test.py
```diff
@@ -548,6 +546,67 @@ def test_metadata(self):
...
+        dependency = simulated.Stage("stage_", "test", axes=["a", "b"],
+                                     ranges={"a": [a_min, a_max], "b": [b_min, b_max]})
+        # Map dependency onto a given rotation
+        tilt = math.radians(26)
+        stage = ConvertStage("conv", "align", {"orig": dependency},
+                             axes=["a", "b"], rotation=tilt)
+        # get the range of stage
+        x_min, x_max = stage.axes["x"].range
+        y_min, y_max = stage.axes["y"].range
+
```
pieleric, line 564: Can you check that (now) the dependency range is the one that was passed at init? Something like:
```python
self.assertEqual(dependency.axes["a"].range, (a_min, a_max))
...
```

src/odemis/driver/test/actuator_test.py
```diff
@@ -548,6 +546,67 @@ def test_metadata(self):
...
+        # Move outside the dep's range
+        with self.assertRaises(ValueError):
+            f = dependency.moveAbs({"a": a_max, "b": 2 * b_max})
+            f.result()
+        # Move dependency in its range
+        [a_pos, b_pos] = [100.e-3, 0.e-3]
+        self.assertTrue(a_min <= a_pos <= a_max)
+        self.assertTrue(b_min <= b_pos <= b_max)
+        f = dependency.moveAbs({"a": a_pos, "b": b_pos})
+        f.result()
```
pieleric, line 608: I find this whole testing block odd because it's testing the move only of the dependency. It should have no reason to fail. At least, that's not what you want to test in this unit test. (I guess you could argue that you want to check that the ConvertStage code didn't mess up with the dependency component, as actually it used to do, but I think that just checking the range explicitly should be really enough).

pieleric (changes requested):
Looks almost good. Just a few minor comments. I think afterwards it should be good to merge!

pieleric (approved):

PR #2518, opened 2023-09-11
[fix] add rotational axes for Thermofisher when switching using grid mode 

K4rishma:
The rotational axes are updated when target is moved using GRID 1 and GRID 2.

With the old workflow. when the stage is moved from Loading to GRID 1 position (default SEM mode), then the resultant stage position will not be updated in rx and rz but only in x, y, z. The current changes, update rx and rz.

### DIFF
--- a/src/odemis/acq/move.py
+++ b/src/odemis/acq/move.py
@@ -503,6 +503,7 @@ class MeteorTFS1PostureManager(MeteorPostureManager):
                 # The current mode doesn't change. Only X/Y/Z should move (typically
                 # only X/Y).
                 sub_moves.append((self.stage, filter_dict({'x', 'y', 'z'}, target_pos)))
+                sub_moves.append((self.stage, filter_dict({'rx', 'rz'}, target_pos)))
             elif target in (LOADING, SEM_IMAGING, FM_IMAGING):
                 # Park the focuser for safety
                 if not isNearPosition(self.focus.position.value, focus_deactive, self.focus.axes):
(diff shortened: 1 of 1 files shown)


src/odemis/acq/move.py
```diff
@@ -503,6 +503,7 @@ def _doCryoSwitchSamplePosition(self, future, target):
                 # The current mode doesn't change. Only X/Y/Z should move (typically
                 # only X/Y).
```
pieleric, line 504: This comment now disagree with the code. Can you adjust it to explain why we also move rx&rz? 
If I understand correctly, that's to handle the cases when the current rx/rz are not correct compared to the current values, for instance when in LOADING. However, I don't think that was never supposed to happen in the first place. My understanding was that this move can only happen if either in SEM of FM imaging. 

So I think that maybe instead of implementing this, we should raise an error if the current position is neither SEM nor FM imaging. That might also require some adjustments of the GUI, to make sure it's not possible to press the GRID buttons when neither SEM or FM is active. 
Note that I've been thinking of further refactoring the code to separate the change of posture with the change of grid. Such change would maybe help to avoid this issue... but I don't think it's necessary to do that now, unless you feel super motivated.

pieleric (changes requested):

src/odemis/acq/move.py
K4rishma: Indeed the gui would require additional changes to implement the idea from this comment. Maybe I can just edit this comment for now?

pieleric: Yes, let's do so for now. Add a comment, with a `TODO: probably a better way would be to forbid grid switch if not in SEM/FM imaging posture".

pieleric: Please, still update this comment explaining why rx/rz are used while GRID 1&2 should always have the same rx/rz.

pieleric (changes requested):

PR #2520, opened 2023-09-12
[SDEV-1518] dwell time slider does not work for overview imaging - UMCG

nandishjpatel:
[fix] additional scanner VA values to be set for configuring scanner can now be passed as kwargs

Having the option to pass kwargs, the dwellTime VA value can be correctly set for OVERVIEW_MODE after the external VA is set

### DIFF
--- a/src/odemis/driver/xt_client.py
+++ b/src/odemis/driver/xt_client.py
@@ -1607,6 +1607,10 @@ class Scanner(model.Emitter):
 
     def _setDwellTime(self, dwell_time):
         self.parent.set_dwell_time(dwell_time)
+        # Cannot set the dwell_time on the parent if the scan mode is 'external'
+        # hence return the requested value itself
+        if self._isExternal():
+            return dwell_time
         return self.parent.get_dwell_time()
 
     def _setVoltage(self, voltage):
@@ -1691,6 +1695,12 @@ class Scanner(model.Emitter):
         """
         scan_mode = "external" if external else "full_frame"
         self.parent.set_scan_mode(scan_mode)
+        # The dwellTime VA setter can only reflect changes on the SEM server side (parent)
+        # after the external VA is set to False i.e. 'full_frame'
+        if not external:
+            if self.dwellTime.value != self.parent.get_dwell_time():
+                # Set the VA value again to reflect changes on the parent
+                self.dwellTime.value = self.dwellTime.value
         return external
 
     def prepareForScan(self):
--- a/src/odemis/driver/xt_client.py
+++ b/src/odemis/driver/xt_client.py
@@ -1695,12 +1695,15 @@ class Scanner(model.Emitter):
         """
         scan_mode = "external" if external else "full_frame"
         self.parent.set_scan_mode(scan_mode)
-        # The dwellTime VA setter can only reflect changes on the SEM server side (parent)
+        # The dwellTime and scale VA setter can only reflect changes on the SEM server side (parent)
         # after the external VA is set to False i.e. 'full_frame'
         if not external:
             if self.dwellTime.value != self.parent.get_dwell_time():
                 # Set the VA value again to reflect changes on the parent
                 self.dwellTime.value = self.dwellTime.value
+            if self.resolution.value != tuple(self.parent.get_resolution()):
+                # Set the VA value again to reflect changes on the parent
+                self.scale.value = self.scale.value
         return external
 
     def prepareForScan(self):
--- a/src/odemis/driver/xt_client.py
+++ b/src/odemis/driver/xt_client.py
@@ -1511,10 +1511,21 @@ class Scanner(model.Emitter):
         """
         logging.debug("Updating SEM settings")
         try:
-            dwell_time = self.parent.get_dwell_time()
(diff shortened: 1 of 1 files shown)


src/odemis/acq/fastem_conf.py
```diff
@@ -86,6 +87,8 @@ def configure_scanner(scanner, mode):
     scanner.multiBeamMode.value = conf["multiBeamMode"]
     scanner.external.value = conf["external"]
     scanner.blanker.value = conf["blanker"]
+    if "dwellTime" in kwargs:
+        scanner.dwellTime.value = kwargs["dwellTime"]
```
pieleric, line 91: Why do you have to do it here? Is there something special related to the "external" VA? If so, it should be explained. But alternatively, it should be fixed in the driver.

nandishjpatel: > But alternatively, it should be fixed in the driver.

This sound like a better fix, I will discuss it with @tepals

src/odemis/driver/xt_client.py
```diff
@@ -1607,6 +1608,7 @@ def _updatePixelSize(self):
 
     def _setDwellTime(self, dwell_time):
         self.parent.set_dwell_time(dwell_time)
+        self._current_dwell_time = dwell_time
         return self.parent.get_dwell_time()
```
pieleric, line 1612: It's still going to indicate that the dwell time hasn't changed, as you returned the "accepted" value. Or does the server return an acceptable value already?  If not, that's confusing as the user could set the dwell time to a new value, the old value would be returned, and later when changing to external, suddenly the requested value is set.
I think it should just return the value that is "about to be set as soon as external will be disabled". (ie, check if external is True, and if so, don't ask the server the new value.

src/odemis/driver/xt_client.py
```diff
@@ -1691,6 +1693,11 @@ def _setExternal(self, external):
         """
         scan_mode = "external" if external else "full_frame"
         self.parent.set_scan_mode(scan_mode)
+        # The dwellTime VA setter can only reflect changes on the SEM server side (parent)
+        # after the external VA is set to False i.e. 'full_frame'
+        if not external:
+            if self._current_dwell_time != self.parent.get_dwell_time():
+                self.dwellTime.value = self._current_dwell_time
```
pieleric, line 1700: Is there the same issue with resolution? If so, it'd make sense to fix it at the same time.

pieleric (changes requested):

src/odemis/driver/xt_client.py
nandishjpatel: > I think it should just return the value that is "about to be set as soon as external will be disabled". (ie, check if external is True, and if so, don't ask the server the new value.

Thanks, this makes sense

src/odemis/driver/xt_client.py
nandishjpatel: The resolution is a readonly VA [see here](https://github.com/delmic/odemis/blob/fb0fe2ad614bb03847647717748046dde4bdcbe1/src/odemis/driver/xt_client.py#L1485), which is used for calculation and setting  the scale VA (in the driver and [fastem_conf](https://github.com/delmic/odemis/blob/fb0fe2ad614bb03847647717748046dde4bdcbe1/src/odemis/acq/fastem_conf.py#L114)). I think there is no need to fix this, what do you think?

pieleric: Indeed... but then what about the scale?

src/odemis/driver/xt_client.py
pieleric: Just for safety, could you still call `self.parent.set_dwell_time(dwell_time)`. This way, if for some systems, it has effect, it will still behave as before.

pieleric (changes requested):
Looks good. Just a tiny (last?) remark.

src/odemis/driver/xt_client.py
nandishjpatel: @tepals can the scale be set in the TFS UI when in `external` mode?

nandishjpatel: Possible fixes:

```
        if not external:
            if self.dwellTime.value != self.parent.get_dwell_time():
                # Set the VA value again to reflect changes on the parent
                self.dwellTime.value = self.dwellTime.value
            if self.resolution.value != tuple(self.parent.get_resolution()):
                # Set the VA value again to reflect changes on the parent
                self.scale.value = self.scale.value
```

In below we don't call `self.parent.set_resolution(res)`
```
        if not external:
            if self.dwellTime.value != self.parent.get_dwell_time():
                # Set the VA value again to reflect changes on the parent
                self.dwellTime.value = self.dwellTime.value
            self._updateResolution()
```

tepals: The scale is specific to Odemis, it updates the resolution and that is what we set on the TFS interface.

nandishjpatel: Thanks Thera, fix 1. is needed then

pieleric (approved):

pieleric (approved):

tepals (approved):

PR #2521, opened 2023-09-12
[perf] StaticSpectrumStream: avoid running the calibration correction multiple times with the same input

pieleric:
At init, the background and coeficient are initialised to None, but
typically the GUI sets them immediately after to the current value,
which is often None too.
In such case, the calibration correction is done up to 3 times in a
row with the same input (and generates the same output).

For some data, like large EK data, the correction can take ~30s. So this
helps opening some files faster.

### DIFF
--- a/src/odemis/acq/stream/_static.py
+++ b/src/odemis/acq/stream/_static.py
@@ -24,26 +24,26 @@ see http://www.gnu.org/licenses/.
 # they were initialised with.
 
 from past.builtins import long
-from collections.abc import Iterable
 import copy
 import gc
 import logging
 import math
-import numpy
-from odemis import model, util
-from odemis.acq import calibration
-from odemis.model import MD_POS, MD_POL_MODE, VigilantAttribute, MD_POL_S0
-from odemis.util import img, conversion, spectrum, find_closest, almost_equal
 import threading
 import time
 import weakref
+from collections.abc import Iterable
 
-from ._base import Stream, POL_POSITIONS_RESULTS, POL_POSITIONS
-
+import numpy
 try:
     import arpolarimetry
 except ImportError:
     arpolarimetry = None
+
+from odemis import model, util
+from odemis.acq import calibration
+from odemis.model import MD_POL_MODE, MD_POL_S0, MD_POS, VigilantAttribute
+from odemis.util import almost_equal, conversion, find_closest, img, spectrum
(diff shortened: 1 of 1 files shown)


src/odemis/acq/stream/_static.py
```diff
@@ -826,13 +826,20 @@ def _updateCalibratedData(self, data=None, bckg=None, coef=None):
...
           compatible. In that case the current calibrated data is unchanged.
         """
         if data is None:
             data = self.raw[0]  # only one image in .raw for spectrum, temporal spectrum and chronograph
+            prev_bckg, prev_coef = self._calib_parameters
+            # Compare numpy arrays with just "is" to avoid comparing the internal,
+            # which is slow, and because we can assume that arrays are used as read-only.
+            if bckg is prev_bckg and coef is prev_coef:
+                logging.debug("Skipping re-calibration with the same parameters")
+                return
```
tepals, line 842: I thought comparing numpy arrays with "is" does not work. I might just be misunderstanding the code and/or the comment
image

tepals (commented):
One comment, the rest looks good

src/odemis/acq/stream/_static.py
pieleric: That's what I tried to explained in the comment. We don't compare the array's content. We only check if it's the same "block of memory" that was passed before. For this code, it's very unlikely that we receive a different array, but which turns out to contain the same data. In that very unlikely case, then it'll be a little less quick. No big deal.
Any suggestion for a better comment?

tepals: From the comment I did not understand that you were trying to check if it had not updated at all. Now that I understand it, the comment makes sense, which makes it harder to suggest an improvement :grimacing:. How about:
```
# Because we can assume that arrays are used as read-only, compare numpy arrays with just "is" to 
# check that they are exactly the same object. Using "is" is faster than comparing the internal array.
```

K4rishma (approved):

src/odemis/acq/stream/_static.py
tepals: you resolved the comment, but I don't see any changes, is that correct?

pieleric: Sorry... forgot to push :cry: 
Thanks for reminding me!

tepals (approved):

PR #2524, opened 2023-09-14
[SWM-113] Ubuntu22 fixes GUI

sneepstefan:
All GUI configurations are tested on Ubuntu 22.04 and found bugs were addressed.
Tested configurations are:
- SECOM
- ENZEL
- METEOR
- SPARCv1
- SPARCv2
- FAST-EM
- MIMAS
- DELPHI

### DIFF
--- a/src/odemis/gui/comp/overlay/world.py
+++ b/src/odemis/gui/comp/overlay/world.py
@@ -899,7 +899,7 @@ class RepetitionSelectOverlay(WorldSelectOverlay):
 
                 blit = horz_dc.Blit
                 for i in range(buf_rep_x):
-                    x = i * step_x + half_step_x
+                    x = int(i * step_x + half_step_x)
                     blit(x, 0, 3, 3, point_dc, 0, 0)
 
                 total_dc = wx.MemoryDC()
@@ -910,7 +910,7 @@ class RepetitionSelectOverlay(WorldSelectOverlay):
 
                 blit = total_dc.Blit
                 for j in range(buf_rep_y):
-                    y = j * step_y + half_step_y
+                    y = int(j * step_y + half_step_y)
                     blit(0, y, int(end_x - start_x), 3, horz_dc, 0, 0)
 
                 self._bmp.SetMaskColour(wx.BLACK)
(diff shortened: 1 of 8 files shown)


src/odemis/gui/util/widgets.py
```diff
@@ -362,7 +362,7 @@ def _update_progress(self):
             self._last_update = now
 
         # Update progress bar
-        self._bar.Value = int(PROGRESS_RANGE * ratio)
+        self._bar.Value = int(round(PROGRESS_RANGE * ratio))
```
K4rishma, line 365:
```suggestion
        self._bar.Value = round(PROGRESS_RANGE * ratio)
```

src/odemis/gui/comp/overlay/world.py
```diff
@@ -899,7 +899,7 @@ def _draw_points(self, ctx):
 
                 blit = horz_dc.Blit
                 for i in range(buf_rep_x):
-                    x = i * step_x + half_step_x
+                    x = round(i * step_x + half_step_x)
```
K4rishma, line 902: int would be nicer as this code is expecting int and the test cases would be expecting int as well. Round can lead to error in test cases because of it odd and even behavior.

src/odemis/gui/comp/overlay/world.py
```diff
@@ -910,7 +910,7 @@ def _draw_points(self, ctx):
 
                 blit = total_dc.Blit
                 for j in range(buf_rep_y):
-                    y = j * step_y + half_step_y
+                    y = round(j * step_y + half_step_y)
```
K4rishma, line 913: similar explanation to choose int over round

src/odemis/gui/cont/streams.py
```diff
@@ -3323,7 +3323,8 @@ def clear_feature_streams(self):
             self._stream_bar.remove_stream_panel(sc.stream_panel)
             if hasattr(v, "removeStream"):
                 v.removeStream(sc.stream)
-            self.stream_controllers.remove(sc)
+            else:
```
pieleric, line 3326: I don't think that's relate to whether the view can remove the stream or not.
I'd suggest to just explicitly check for whether the element is in the list:
`if sc in self.stream_controllers:`

pieleric (changes requested):

pieleric:
@sneepstefan , afterwards, could you merge the review fixes back into each commit they correspond to? I'd like to merge each commit independently, as they have no direct relationship with each other.

pieleric (approved):

PR #2528, opened 2023-09-20
[SWM-129] Fix img_test.py/test_tint_to_colormap_unexpected_input

sneepstefan:
Test case test_tint_to_colormap_unexpected_input in /util/test/img_test.py was failing on Ubuntu 18.04. 
The expected ValueError was not raised in this case.

On Ubuntu 18.04 RGB values in tuples higher than 255 are somehow accepted or forced in between the range of 0-255.
The root of this reason is not easily traced as the code is distributed over multple .py files and/or classes.

To make sure the ValueError exception is raised a check is added to /util/conversion.py rgb_to_frgb method to make 
sure all the values in the tuple are within the range of 0-255.

### DIFF
--- a/src/odemis/util/conversion.py
+++ b/src/odemis/util/conversion.py
@@ -106,10 +106,10 @@ def rgb_to_frgb(rgb):
     """
     # check if the tuple consists of three values
     if len(rgb) != 3:
-        raise ValueError("Illegal RGB colour %s" % rgb)
+        raise ValueError("Illegal RGB colour %s" % (rgb,))
     # check if the tuple values are valid RGB values
-    if not numpy.all(tuple(0 <= c <= 255 for c in rgb)):
-        raise ValueError("Not all values are valid RGB values (0-255)")
+    if not all(0 <= c <= 255 for c in rgb):
+        raise ValueError("Not all values are valid RGB values (0-255) of passed colour %s" % (rgb,))
     return tuple(v / 255.0 for v in rgb)
 
 
--- a/src/odemis/util/conversion.py
+++ b/src/odemis/util/conversion.py
@@ -104,9 +104,12 @@ def rgb_to_frgb(rgb):
     :return: (float, float, float)
 
     """
-
+    # check if the tuple consists of three values
     if len(rgb) != 3:
         raise ValueError("Illegal RGB colour %s" % rgb)
+    # check if the tuple values are valid RGB values
+    if not numpy.all(tuple(0 <= c <= 255 for c in rgb)):
+        raise ValueError("Not all values are valid RGB values (0-255)")
     return tuple(v / 255.0 for v in rgb)
 
 


sneepstefan:
@tepals could you test this on Ubuntu 18 to verify this fix as well please?

src/odemis/util/conversion.py
```diff
@@ -104,9 +104,12 @@ def rgb_to_frgb(rgb):
     :return: (float, float, float)
 
     """
-
+    # check if the tuple consists of three values
     if len(rgb) != 3:
         raise ValueError("Illegal RGB colour %s" % rgb)
+    # check if the tuple values are valid RGB values
+    if not numpy.all(tuple(0 <= c <= 255 for c in rgb)):
+        raise ValueError("Not all values are valid RGB values (0-255)")
```
pieleric, line 112: Please include the actual value of `rgb` in the message (to make the debugging easier).

pieleric (changes requested):
The detection of the value out of range was done in the matplotlib color function. It's not really hard to track if you look at the traceback 😉
Now, please be aware that a loooot of code of the GUI called this rgb_to_frgb conversion function. Some of it might call it with values (slightly) out of range. I think there is like 20% chance this happens. So please check well the GUI with this change applied.

src/odemis/util/conversion.py
```diff
@@ -104,9 +104,12 @@ def rgb_to_frgb(rgb):
     :return: (float, float, float)
 
     """
-
+    # check if the tuple consists of three values
     if len(rgb) != 3:
         raise ValueError("Illegal RGB colour %s" % rgb)
+    # check if the tuple values are valid RGB values
+    if not numpy.all(tuple(0 <= c <= 255 for c in rgb)):
+        raise ValueError("Not all values are valid RGB values (0-255) of passed colour %s" % rgb)
```
pieleric, line 112: You need to write `(rgb,)`.

pieleric (changes requested):

src/odemis/util/conversion.py
sneepstefan: yeah I guess I was a little to quick with my changes..

sneepstefan:
> Now, please be aware that a loooot of code of the GUI called this rgb_to_frgb conversion function. Some of it might call it with values (slightly) out of range. I think there is like 20% chance this happens. 
That does not sound too well actually. In this case we could better think of a different approach? If its only slightly we could accept a percentage of the value under or over the range maybe?

pieleric:
> That does not sound too well actually. In this case we could better think of a different approach? If its only slightly we could accept a percentage of the value under or over the range maybe?

Well, let's not also do too relaxed behaviour. Any RGB value out of 0-255 will at some point behave oddly. It's better to properly catch it. I'm just say you should run the GUI with a couple of back-end types to check nothing unexpected happens.

src/odemis/util/conversion.py
```diff
@@ -104,9 +104,12 @@ def rgb_to_frgb(rgb):
     :return: (float, float, float)
 
     """
-
+    # check if the tuple consists of three values
     if len(rgb) != 3:
-        raise ValueError("Illegal RGB colour %s" % rgb)
+        raise ValueError("Illegal RGB colour %s" % (rgb,))
+    # check if the tuple values are valid RGB values
+    if not numpy.all(tuple(0 <= c <= 255 for c in rgb)):
```
pieleric, line 111: You don't need to use `numpy.all()`, you can just use `all()`. You also don't need to convert to a tuple then.
```suggestion
    if not all(0 <= c <= 255 for c in rgb):
```

pieleric (changes requested):

nandishjpatel (approved):

pieleric (approved):

tepals:
> @tepals could you test this on Ubuntu 18 to verify this fix as well please?

I don't have a 18.04 VM ready, @nandishjpatel could you check on the CI server?

nandishjpatel:
> > @tepals could you test this on Ubuntu 18 to verify this fix as well please?
> 
> I don't have a 18.04 VM ready, @nandishjpatel could you check on the CI server?

```
testing@testing:~/development/odemis/src/odemis/util/test$ python3 -m unittest conversion_test.py 
ERROR:root:Syntax error: while parsing a flow node
expected the node content, but found '<stream end>'
  in "<unicode string>", line 1, column 4:
    {5:
       ^
ERROR:root:Syntax error: while parsing a flow sequence
  in "<unicode string>", line 1, column 1:
    [5.3
    ^
expected ',' or ']', but got '<stream end>'
  in "<unicode string>", line 1, column 5:
    [5.3
        ^
............
Ran 12 tests in 0.009s

OK
```
@sneepstefan here is the test result

sneepstefan:
@nandishjpatel Could you run test_tint_to_colormap_unexpected_input from src/odemis/util/test/img_test.py instead?

nandishjpatel:
> test_tint_to_colormap_unexpected_input

here you go

```
testing@testing:~/development/odemis/src/odemis/util/test$ python3 -m unittest img_test.TestTintToColormap.test_tint_to_colormap_unexpected_input
.
Ran 1 test in 0.002s

OK
```

PR #2533, opened 2023-09-29
[CDSM-1094] updated the StaticARStream to contain emission metadata

sneepstefan:
Currently only StaticCLStream supported the read-only emission field which show the band wavelength used in the analysis. When analyzing AR in the analysis tab the user would find the right MD under the button metadata. Therefore  the StaticARStream emission metadata is now updated to show as a RO VA on the stream panel.

### DIFF
--- a/src/odemis/acq/stream/_static.py
+++ b/src/odemis/acq/stream/_static.py
@@ -466,6 +466,19 @@ class StaticARStream(StaticStream):
 
         super(StaticARStream, self).__init__(name, list(self._pos.values()), *args, **kwargs)
 
+        # show the selected band filter wavelength in the emission field
+        try:
+            em_range = data[0].metadata[model.MD_OUT_WL]
+            if isinstance(em_range, str):
+                unit = None
+            else:
+                unit = "m"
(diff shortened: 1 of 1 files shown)


src/odemis/acq/stream/_static.py
```diff
@@ -466,6 +466,19 @@ def dis_bbtl(v):
...
+            em_range = data[0].metadata[model.MD_OUT_WL]
+            if isinstance(em_range, str):
+                unit = None
+            else:
+                unit = "m"
+            self.emission = VigilantAttribute(em_range, unit=unit,
+                                              readonly=True)
+
+        except KeyError:
+            logging.warning("No emission wavelength for CL stream")
```
pieleric, line 480: CL stream -> AR stream ?

pieleric (changes requested):

src/odemis/acq/stream/_static.py
pieleric: Also, it should not be a warning, but just info. It's very common to not have this metadata, as many systems don't have a filter wheel.

tepals (approved):

pieleric (approved):

PR #2534, opened 2023-09-29
[fix] hdf5: decode all strings from hdf5 files when they are bytes

tepals:
Sometimes when reading strings from hdf5 files bytes are returned instead of a string. This fix ensures that if bytes are returned they are always converted. The code has been refactored to move the function read_str out of the for-loop it was in and to it's own function.

On Ubuntu 22.04 the following test case failed:

FAIL: testMetadata (hdf5_test.TestHDF5IO)
checks that the metadata is saved with every picture ---------------------------------------------------------------------- Traceback (most recent call last):
  File "/home/runner/work/odemis/odemis/src/odemis/dataio/test/hdf5_test.py", line 423, in testMetadata
    self.assertEqual(metadata[model.MD_DESCRIPTION], desc)
AssertionError: 'tÉst' != b't\xc3\x89st'

### DIFF
--- a/src/odemis/dataio/hdf5.py
+++ b/src/odemis/dataio/hdf5.py
@@ -24,6 +24,7 @@ import json
 import logging
 import os
 import time
+from typing import Union
 
 import h5py
 import numpy
@@ -618,9 +619,7 @@ def _read_image_info(group):
                     if isinstance(group["COffset"], (bytes, str)):
                         # Only to support some files saved with Odemis 2.3-alpha
                         logging.warning("COffset is a string, not officially supported")
-                        out_wl = group["COffset"]
-                        if not isinstance(out_wl, str):
-                            out_wl = out_wl.decode("utf-8", "replace")
+                        out_wl = convert_to_str(group["COffset"])
                         md[model.MD_OUT_WL] = out_wl
                     else:
                         # read polynomial: first is C offset, second coefficient polynomial first order (linear)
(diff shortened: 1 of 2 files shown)


sneepstefan (approved):

src/odemis/dataio/hdf5.py
```diff
@@ -1663,3 +1647,14 @@ def read_thumbnail(filename):
     # TODO: support filename to be a File or Stream
 
     return _thumbFromHDF5(filename)
+
+
+def read_str(s):
+    """
+    Make sure we can read both bytes (HDF5 ascii) and string (HDF5 utf8) metadata
+    That's important because sometimes strings are stored as bytes (ie ascii).
+    This forces both ascii and utf8 to be stored as strings (unicode)
```
pieleric, line 1656: Reading this makes me quite confused. I'd suggest to clarify in this way:
```suggestion
    HDF5 has different types of strings: ASCII vs UTF8, and fixed- vs variable-length.
    h5py convert some to "unicode" (aka string) and some to bytes. Which one is converted
    to what depends on the version of h5py! To keep some sanity, when it is known that a
    data is a unicode text, use this function to ensure it.
```

src/odemis/dataio/hdf5.py
```diff
@@ -1663,3 +1647,14 @@ def read_thumbnail(filename):
     # TODO: support filename to be a File or Stream
 
     return _thumbFromHDF5(filename)
+
+
+def read_str(s):
```
pieleric, line 1652: As the function doesn't really read, but just converts, it's a little bit misnamed. I'd suggest `ensure_str` or `convert_to_str`
Also, please add typing info:
`(s: Union[bytes, str]) -> str:`

src/odemis/dataio/hdf5.py
```diff
@@ -1663,3 +1647,14 @@ def read_thumbnail(filename):
...
+    """
+    Make sure we can read both bytes (HDF5 ascii) and string (HDF5 utf8) metadata
+    That's important because sometimes strings are stored as bytes (ie ascii).
+    This forces both ascii and utf8 to be stored as strings (unicode)
+    """
+    if isinstance(s, bytes):
+        s = s.decode('utf8')
```
pieleric, line 1659: Please add `, "replace"` argument. That ensures that if some character is not a proper UTF8 bytes, it doesn't totally fail, but still pass to the user the basic content of the string.

pieleric (changes requested):

pieleric (approved):

PR #2535, opened 2023-09-29
[ci] add some gui related test cases to the cloud hosted CI

tepals:


pieleric (approved):

.github/workflows/unit_tests_cloud.yml
```diff
@@ -157,3 +151,27 @@ jobs:
...
+    - name: Run tests from odemis.gui.conf
+      if: success() || failure()
+      run: |
+        export PYTHONPATH="$PWD/src:$PYTHONPATH"
+        python3 -m unittest src.odemis.gui.conf.test.conf_test.AcquisitionConfigTest --verbose
+        python3 -m unittest src.odemis.gui.conf.test.conf_test.CalibrationConfigTest --verbose
+        python3 -m unittest src.odemis.gui.conf.test.conf_test.GeneralConfigTest.test_save --verbose
+        python3 -m unittest src.odemis.gui.conf.test.conf_test.GeneralConfigTest.test_save_unicode --verbose
+        python3 -m unittest src.odemis.gui.conf.test.conf_test.GeneralConfigTest.test_simple --verbose
+        # test_default fails on CI, because the manual path does not exist on the server
```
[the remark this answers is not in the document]
tepals: That doesn't seem to work, so I'll just leave it as is for now. 
```
touch: cannot touch '/usr/share/doc/odemis/user-guide.pdf': No such file or directory
```

.github/workflows/unit_tests_cloud.yml
```diff
@@ -157,3 +151,27 @@ jobs:
       run: |
         export PYTHONPATH="$PWD/src:$PYTHONPATH"
         python3 -m unittest src/odemis/acq/test/feature_test.py --verbose
+
+    - name: Run tests from odemis.gui.conf
+      if: success() || failure()
```
FCanberk, line 156: probably just a personal preference but those `success() || failure()`s in the if statements gives me the feeling that it would run the tests in any case. 

would it be too much of a hassle to change them as `if: '!cancelled()'`?

tepals: When I implemented everything originally I tried using `if: always()` but that doesn't have the desired result, I now see they added the use `if: '!cancelled()'` to the documentation, I don't think that description was there previously. I'll update the code.
https://docs.github.com/en/actions/learn-github-actions/expressions#always

tepals:
Don't merge yet, I added a failing test case to check the if !cancelled()

.github/workflows/unit_tests_cloud.yml
tepals: I've updated the code and I followed the suggestion of the github documentation, so I wrote `if: ${{ !cancelled() }}` instead of your suggested `if: '!cancelled()'`. Let me know if you agree or not

sneepstefan (approved):

.github/workflows/unit_tests_cloud.yml
FCanberk: looks good!

FCanberk (approved):

PR #2537, opened 2023-10-06
[fix] gui SPARC2 alignment tab: show mirror control in lens 1 alignment again

pieleric:
This is an error introduced in commit "gui SPARC2 alignment tab:
completly hide panels which cannot be used". It hided too many panels.

Also fix an issue introduced in commit "merge SPARC2 mirror alignment
into lens alignment": now that the stream allows for background
subtraction, the mirror alignment also has the background subtracted. So
it needs to have a way to reset it too. Especially, the image is not the
same as for the lens/mirror alignement.

### DIFF
--- a/src/odemis/gui/cont/tabs.py
+++ b/src/odemis/gui/cont/tabs.py
@@ -6560,7 +6560,7 @@ class Sparc2AlignTab(Tab):
                 self._ccd_stream.should_update.value = True
             if self._mirror_settings_controller:
                 self._mirror_settings_controller.enable(False)
-            self.panel.pnl_mirror.Show(False)
+            self.panel.pnl_mirror.Show(True)
             self.panel.pnl_lens_mover.Show(True)
             self.panel.pnl_lens_mover.Enable(False)  # Will be enabled once the lens is at the correct place
             self.panel.pnl_lens_switch.Show(False)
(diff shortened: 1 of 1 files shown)


FCanberk (approved):

src/odemis/gui/cont/tabs.py
```diff
@@ -6618,6 +6618,7 @@ def _onAlignMode(self, mode):
         elif mode == "center-align":
             self.tab_data_model.focussedView.value = self.panel.vp_align_center.view
             self._ccd_stream.should_update.value = True
+            # TODO: allow to reset background. Also reset automatically when switching between modes
```
pieleric, line 6621: Now the code allows to reset => change TODO to just "automatically reset" (or "use a different stream"?)

sneepstefan: I think "automatically reset" is fine is this case although I am not completely sure how both options differ. Where is the new code that allows to reset to be found?

pieleric: The new code is the `self.panel.btn_bkg_acquire.Enable(True)`. It shows the "Acquire background" button if there is no background yet, and a "Remove background" if the background is already acquired.

sneepstefan: Alright, could you in that case delete the TODO comment (its done right) and add a comment to line 6638 describing the choice to either remove or acquire the background based on the current state?

PR #2539, opened 2023-10-09
[SDEV-1516] Implement Picoscale driver using the Python SDK

nandishjpatel:
the driver is created making use of the Python SDK: https://drive.google.com/drive/folders/1AKxHZFxjX4sHoBNZrRswnp6-Hnoqo7bs

In addition PR to be reviewed with this: https://bitbucket.org/delmic/ubuntu-delmic/pull-requests/16


src/odemis/driver/smaract.py
```diff
@@ -4913,6 +4922,406 @@ def _cancelReference(self, future):
         return True
 
 
+class PicoscalePythonSDK(model.HwComponent):
```
pieleric, line 4925: Let's not keep the Picoscale class, which is broken. I'd suggest you remove the old class and rename this class as `Picoscale`. The arguments are compatible with the old class, right? (so the microscope files don't need to be updated).

src/odemis/driver/smaract.py
```diff
@@ -4913,6 +4922,406 @@ def _cancelReference(self, future):
...
+                outdated and the reported position values might not be accurate.
+            * False / "never": Never reference. This means that the device might not be able to produce position data,
+                if it was not previously referenced.
+        precision_mode: (0 <= int <= 5) strength of digital lowpass filter, a higher level corresponds to higher
+            precision, but lower velocity. Not available on all systems.
+        """
+        model.HwComponent.__init__(self, name, role, *args, **kwargs)
+
+        if not smaract_python_sdk:
+            raise ModuleNotFoundError("Smaract Python SDK modules are missing. Cannot run PicoScale driver.")
```
pieleric, line 4974: When running the backend without the python SDK installed, does it bring up a good error message? (I'm mostly worried that the `ModuleNotFoundError` convinces the backend to show a generic error message, preventing to see the actual error. In this case you could raise just a OSError). 
Also, please add to the message the line to type to install the packages. `sudo apt install python3-smaract-picoscale...`.

src/odemis/driver/smaract.py
```diff
@@ -4913,6 +4922,406 @@ def _cancelReference(self, future):
...
+
+        :param error: (si.EventType or ps.EventType) The event.
+        :return: (str) The readable event name.
+
+        """
+        ev_name = "(0x{:04X})".format(event.type)
+        if event.type in set(ev.value for ev in si.EventType):
+            ev_name = "si.EventType." + si.EventType(event.type).name + " " + ev_name
+        elif event.type in set(ev.value for ev in ps.EventType):
+            ev_name = "ps.EventType." + ps.EventType(event.type).name + " " + ev_name
```
pieleric, line 5209: For this sort of code, I think it's better to think of it in an "optimistic" way. Try to get the si.EventType, and let python find out if it exists or not. If not, it will raise an exception. In this end, it's less code executed.
Something like that (and you could update `getReadableError` similarly.
```suggestion

        try:
            ev_name = "si.EventType." + si.EventType(event.type).name + " " + ev_name
        except ValueError:
            try:
                  ev_name = "ps.EventType." + ps.EventType(event.type).name + " " + ev_name
            except ValueError:
                  ev_name = "(0x{:04X})".format(event.type)

```

src/odemis/driver/smaract.py
```diff
@@ -4913,6 +4922,406 @@ def _cancelReference(self, future):
...
+
+            # The adjustment state should be "disabled" after automatic adjustment. If the referencing
+            # procedure was stopped prematurely, right after it was set to "manual", it might be in the
+            # wrong state --> make sure it's set to disabled here.
+            if si.GetProperty_i32(self.id, si.EPK(ps.Property.AF_ADJUSTMENT_STATE, 0, 0)) != ps.AdjustmentState.DISABLED:
+                si.SetProperty_i32(self.id, si.EPK(ps.Property.AF_ADJUSTMENT_STATE, 0, 0), ps.AdjustmentState.DISABLED)
+                self._wait_for_progress_event(ps.AdjustmentState.DISABLED, timeout=600)  # state will be DISABLED when done
+            si.SetProperty_i32(self.id, si.Property.EVENT_NOTIFICATION_ENABLED << 16 | ps.EventType.AF_ADJUSTMENT_PROGRESS, si.DISABLED)
+
+    def _getReadableSIError(self, error):
```
pieleric, line 5182: The code returns either a SI or a PS error name, so I'd suggest to rename it `getReadableError`, or maybe `getErrorName`.

src/odemis/driver/smaract.py
```diff
@@ -4913,6 +4922,406 @@ def _cancelReference(self, future):
...
+            si.Close(self.id)
+            self._executor = None
+        super(PicoscalePythonSDK, self).terminate()
+
+    @roattribute
+    def axes(self):
+        """ dict str->Axis: name of each axis available -> their definition."""
+        return self._axes
+
+    def _wait_for_progress_event(self, end_state, timeout=float("inf")):
```
pieleric, line 5275: Actually, with Python 3 you can write `math.inf`, which is nicer to read. Could you update the code to use this writing?

src/odemis/driver/smaract.py
```diff
@@ -4913,6 +4922,406 @@ def _cancelReference(self, future):
...
+                # Switch to automatic adjustment
+                si.SetProperty_i32(self.id, si.EPK(ps.Property.AF_ADJUSTMENT_STATE, 0, 0), ps.AdjustmentState.AUTO_ADJUST)
+            self._wait_for_progress_event(ps.AdjustmentState.DISABLED, timeout=600)  # state will be DISABLED when done
+            logging.debug("Finished referencing.")
+
+            # We could save the referencing parameters to memory here. This could be useful for the validation,
+            # because it initializes the system by loading the stored referencing parameters. However, we don't
+            # currently use validation, so there is no need to store the values every time we reference.
+        except si.Error as error:
+            logging.debug(self._getReadableSIError(error))
```
pieleric, line 5142: Drop. Maybe, if that's not in `str(error)`, also include the error name in line 5145.

pieleric (changes requested):

src/odemis/driver/smaract.py
nandishjpatel: > The arguments are compatible with the old class, right?

That's right

> I'd suggest you remove the old class and rename this class as `Picoscale`

I let the old class remain because it can be used a `fake` simulator. Should I create a fake simulator for the Python SDK?

pieleric:
BTW, the title of this PR is very unclear. I guess it comes from the title of the task. However, the title should be about what the change is.

src/odemis/driver/smaract.py
```diff
@@ -4218,20 +4225,29 @@ def __init__(self, name, role, locator, channels, ref_on_init=False, precision_m
         """
         model.HwComponent.__init__(self, name, role, *args, **kwargs)
 
+        if not smaract_python_sdk:
```
tepals, line 4228: I can't comment on the lines above, but I think it might be a good idea to mention that connecting using the serial number has caused problems.

``` 
            The device can also be addressed by its serial number. Note: this has caused connection issue on some of the devices.
                usb:sn:<serial_number>
                network:sn:<serial_number>
```

src/odemis/driver/smaract.py
```diff
@@ -4218,20 +4225,29 @@ def __init__(self, name, role, locator, channels, ref_on_init=False, precision_m
...
+
         # Connection
         if locator == "fake":
-            self.core = FakePicoscale_DLL()
+            self.core = FakePicoscale()
         else:
-            self.core = SA_SIDLL()
-        self._locator = locator
-        self._id = self._openConnection(locator)
+            self.core = si
```
tepals, line 4236: ~I don't see `si` defined anywhere, should it be part of the keyword arguments?~

tepals: I just noticed that it comes from:
```
    import smaract.si as si
    import smaract.picoscale as ps
```
In some places you use `si.` and in others `self.core.`, what's the difference?

src/odemis/driver/smaract.py
```diff
@@ -4754,139 +4416,76 @@ def _doReference(self, future):
...
-        """
-        Callback function for referencing/validation future in __init__.
         """
         try:
-            future.result()
-            self.state._set_value(model.ST_RUNNING, force_write=True)
-        except Exception as e:
-            self.state._set_value(e, force_write=True)
-            logging.exception(e)
+            ev_name = "si.EventType." + si.EventType(event.type).name + " " + "0x{:04X}".format(si.EventType(event.type).value)
```
tepals, line 4468: Why not use f-string here?
```suggestion
            ev_name = f"si.EventType.{si.EventType(event.type).name} 0x{si.EventType(event.type).value:04X}"
```

src/odemis/driver/smaract.py
```diff
@@ -4754,139 +4416,76 @@ def _doReference(self, future):
...
+        except ValueError:
+            try:
+                ev_name = "ps.EventType." + ps.EventType(event.type).name + " " + "0x{:04X}".format(ps.EventType(event.type).value)
```
tepals, line 4471: same here?
```suggestion
                ev_name = f"ps.EventType.{ps.EventType(event.type).name} 0x{ps.EventType(event.type).value:04X}"
```

src/odemis/driver/smaract.py
```diff
@@ -4754,139 +4416,76 @@ def _doReference(self, future):
...
+            except ValueError:
+                ev_name = "0x{:04X}".format(event.type)
```
tepals, line 4473: and here
```suggestion
                ev_name = f"0x{event.type:04X}"
```

tepals (changes requested):

src/odemis/driver/smaract.py
nandishjpatel: `self.core.` has the methods (from si module itself or from the simulator), while the use of `si.` is for constants

src/odemis/driver/smaract.py
nandishjpatel: We are in contact with smaract and it seems the issue can be resolved

tepals (approved):

pieleric (changes requested):

src/odemis/driver/smaract.py
```diff
@@ -4244,23 +4260,19 @@ def __init__(self, name, role, locator, channels, ref_on_init=False, precision_m
...
+        self.core.SetProperty_i32(self.id, si.EPK(ps.Property.SYS_FULL_ACCESS_CONNECTION, 0, 0), si.ENABLED)
         # If referencing is still running, cancel it
-        self.core.SA_SI_Cancel(self._id)
+        self.core.Cancel(self.id)
         # Reset referencing state if necessary (in case referencing has been stopped improperly)
-        state = self.GetProperty_i32(SA_PS_AF_ADJUSTMENT_STATE_PROP)
-        if state != SA_PS_ADJUSTMENT_STATE_DISABLED:
-            self.SetProperty_i32(SA_PS_AF_ADJUSTMENT_STATE_PROP, SA_PS_ADJUSTMENT_STATE_DISABLED)
-        try:
-            self._load_configuration()
```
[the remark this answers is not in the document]
nandishjpatel: After testing, reading developers manual and checking examples from the Python SDK I found its better to not have this. `_load_configuration` loads a previous adjustment result and then waits until the system becomes stable. An adjustment result is saved when we setup the hardware for the first time using Picoscale Windows software. But, in our driver we are not saving an adjustment (referencing) result after completion (I tried it doesn't seem to make a difference). Also the referencing procedure on the device runs much quicker (<10 mins which is the timeout in `_load_configuration`).

src/odemis/driver/smaract.py
```diff
@@ -4754,139 +4415,75 @@ def _doReference(self, future):
...
             raise
         finally:
-            # If the state of the device is stable and all channels are valid, the referencing procedure
-            # succeeded. Typically, all channels become valid at the same time during the reference procedure.
-            # We still check all of them individually to be sure.
-            if self.IsStable():
-                for ch, num in self._channels.items():
-                    self.referenced._value[ch] = self.IsValid(num)
+            # Update the referenced values for the channels
+            for ch, num in self._channels.items():
```
pieleric, line 4419: In the original version, it wouldn't claim that the axes are referenced if the system is not stable. Why changing this?
Also, if the system is not stable, don't you want to still raise an exception, as you used to do?

pieleric (changes requested):

src/odemis/driver/smaract.py
nandishjpatel: Good catch, I made the changes

pieleric (approved):

PR #2541, opened 2023-10-11
[METEOR-520] [fix] Hide hardware settings in the Localization tab when viewing a static stream 

K4rishma:
Hide hardware settings in the Localization tab when viewing a static stream and show in live stream

### DIFF
--- a/src/odemis/gui/cont/tabs.py
+++ b/src/odemis/gui/cont/tabs.py
@@ -392,6 +392,7 @@ class LocalizationTab(Tab):
 
         # Default view is the Live 1
         tab_data.focussedView.value = panel.vp_secom_bl.view
+        tab_data.focussedView.subscribe(self._on_view, init=True)
 
         self._view_selector = viewcont.ViewButtonController(
             tab_data,
@@ -497,6 +498,13 @@ class LocalizationTab(Tab):
                 panel,
                 self
             )
+
+    def _on_view(self, view):
+        """Hide hardware related sub-panels on the right when not in live stream"""
+        live = issubclass(view.stream_classes, odemis.acq.stream._live.LiveStream)
+        self.panel.fp_settings_secom_optical.Show(live)
+        self.panel.fp_secom_streams.Show(live)
+        self.panel.fp_acquisitions.Show(live)
 
     @property
     def settingsbar_controller(self):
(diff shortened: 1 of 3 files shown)


src/odemis/gui/cont/tabs.py
```diff
@@ -498,6 +500,17 @@ def __init__(self, name, button, panel, main_frame, main_data):
                 self
             )
 
+    def _onView(self, view):
+        """Hide hardware related sub-panels on the right when not in live stream"""
+        if "Live" in view.name.value:
```
pieleric, line 505: In general, it's best to not depend on text for user. If one day we decide to adjust this text (ex: call it "Acquisition" instead of "Live"), suddenly this feature will stop working. So, I'd rather have the check rely on the view.stream_classes.  as we discussed previously. I think you can do something like `if is_subclass (view.stream_classes, LiveStream)`.  Or are there any reason not to?

src/odemis/gui/cont/tabs.py
```diff
@@ -498,6 +500,17 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
+            self.panel.fp_settings_secom_optical.Show(True)
+            self.panel.fp_secom_streams.Show(True)
+            self.panel.fp_acquisitions.Show(True)
+        else:
```
pieleric, line 509: For this kind of `if/else ... "bool"`, you can change to a shorter code like:
```
shown = ... my check...
a.Show(shown)
b.Show(shown)
:
```

pieleric (changes requested):
Please also adjust the name of the PR to contain the Jira task ID.

src/odemis/gui/cont/tabs.py
```diff
@@ -498,6 +500,13 @@ def __init__(self, name, button, panel, main_frame, main_data):
                 self
             )
 
+    def _onView(self, view):
```
tepals, line 503: Most of the functions in this class are snake_case, so I suggest to also use that here.
```suggestion
    def _on_view(self, view):
```

tepals (approved):

pieleric (changes requested):

pieleric (approved):

PR #2555, opened 2023-11-02
[fix] Show relevant static streams in stream panel based on the view

K4rishma:
Hide and show streams based on the view selected by the user in the Localisation Tab


src/odemis/gui/cont/streams.py
```diff
@@ -2534,6 +2533,37 @@ def _updateMicroscopeStates(self):
...
 
 
+class CryoStreamsController(StreamBarController):
+    """
+    Controls the display of stream panels without affecting the actual streams
+    """
+
+    def _onView(self, view):
+        """ Handle the changing of the focused view """
+        # super()._onView(view)
```
nandishjpatel, line 2543: remove this line?

nandishjpatel (approved):

src/odemis/gui/cont/streams.py
```diff
@@ -2534,6 +2533,36 @@ def _updateMicroscopeStates(self):
...
+    def _onView(self, view):
+        """ Handle the changing of the focused view """
+        if not view or self.ignore_view:
+            return
+
+        # hide/show the stream panels which are compatible with the view
+        allowed_classes = view.stream_classes
+        for e in self._stream_bar.stream_panels:
+            ov_streams = self._tab_data_model.overviewStreams
+
```
pieleric, line 2550: Please add a comment to explain what your `if` does. I think it tries to separate between different views, but which ones (there are 3 types)?

src/odemis/gui/cont/streams.py
```diff
@@ -2534,6 +2533,36 @@ def _updateMicroscopeStates(self):
...
+            if view.__class__ == guimodel.FeatureView and ov_streams.value:
```
pieleric, line 2551: You should typically never use `__class__`. Most often you wan to use `isinstance(view, ...)`, but may you actually don't want to catch the subclasses. In this case you can write `type(view) == ...`.

pieleric (changes requested):

src/odemis/gui/cont/streams.py
```diff
@@ -2534,6 +2533,39 @@ def _updateMicroscopeStates(self):
...
+            return
+
+        # hide/show the stream panels which are compatible with the view
+        allowed_classes = view.stream_classes
+        for e in self._stream_bar.stream_panels:
+            e.Show(isinstance(e.stream, allowed_classes))
+            ov_streams = self._tab_data_model.overviewStreams
+            static = issubclass(view.stream_classes, StaticStream)
+            # Check view type between Acquired and Overview
+            if static and ov_streams.value:
```
pieleric, line 2553: Why do you have to check for `ov_streams`? It cannot be None. It's just an empty list. So I don't think it's required. If it is, I think it's worth a comment.

src/odemis/gui/cont/streams.py
```diff
@@ -2534,6 +2533,39 @@ def _updateMicroscopeStates(self):
...
+        if not view or self.ignore_view:
+            return
+
+        # hide/show the stream panels which are compatible with the view
+        allowed_classes = view.stream_classes
+        for e in self._stream_bar.stream_panels:
+            e.Show(isinstance(e.stream, allowed_classes))
+            ov_streams = self._tab_data_model.overviewStreams
+            static = issubclass(view.stream_classes, StaticStream)
+            # Check view type between Acquired and Overview
```
pieleric, line 2552: Please explain what do you want to *attain* here. The final goal. I think you have like 3 options: l
* live stream -> live views
* acquired (static) stream related to a feature -> "Acquired" view
* (static) overviews -> "Overview" view

Is that correct?
Then, to distinguish between the 2 views related to StaticStream, it's not possible to only depend on the "stream_classes", so you depend on the class type (which is FeatureOverviewView only for the Overview) + the fact that the stream is in the "overviewStreams". Is that correct? If so, please write all this down in a comment here!

src/odemis/gui/cont/streams.py
```diff
@@ -2534,6 +2533,39 @@ def _updateMicroscopeStates(self):
...
+            if static and ov_streams.value:
+                if type(view) == guimodel.FeatureOverviewView:
```
pieleric, line 2554: Actually, the views are either FeatureOverviewView (which inherits from FeatureView), or FeatureView. So it should be fine to just write the "standard" `if isinstance(view, FeatureOverviewView)`.

pieleric (changes requested):

pieleric (approved):

PR #2556, opened 2023-11-06
[METEOR-829] TFS metadata parser, stitching improvements

patrickcleeve2:
This PR adds:

- [x] Extension to TIFF parser to read TFS / FEI metadata
- [x] Refactor of stitching functionality
- [x] Feature for loading tilesets in the GUI


src/odemis/dataio/tiff.py
```diff
@@ -2613,3 +2624,121 @@ def _iterDirectories(tiff_file):
...
+    md[model.MD_HW_NAME] = fei_md["System"]["SystemType"]+"-"+fei_md["System"]["Dnumber"]    
+    md[model.MD_SW_VERSION] = fei_md["System"]["Software"]
+    md[model.MD_PIXEL_SIZE] = (float(fei_md["Scan"]["PixelWidth"]), 
+                                  float(fei_md["Scan"]["PixelHeight"]))
+    
+    md[model.MD_DWELL_TIME] = float(fei_md["Scan"]["Dwelltime"])
+    md[model.MD_DIMS] = "YX"
+    md[model.MD_POS] = (-float(fei_md["Stage"]["StageX"]), 
+                           -float(fei_md["Stage"]["StageY"]),
+                        #    -float(fei_md["Stage"]["StageZ"])
```
pieleric, line 2678: Normally the MD_POS (and MD_PIXEL_SIZE) metadata should have a 3rd dimension if the dataset has a "Z" dimension. So probably you never need it here (as I don't think users will read "Z stacks" from a TFS SEM, at least in the short/mid term).

src/odemis/dataio/tiff.py
```diff
@@ -2613,3 +2624,121 @@ def _iterDirectories(tiff_file):
             tiff_file.ReadDirectory()
             dir_index += 1
             yield dir_index
+
+
+ 
+### FEI Metadata parsing
+def _parseINIMetadataAsDict(metadata: str, delimeter: str = "\r\n"):
```
pieleric, line 2631: Isn't it possible to just use the `configparser` module from the standard library?

src/odemis/dataio/tiff.py
```diff
@@ -2613,3 +2624,121 @@ def _iterDirectories(tiff_file):
...
+            "y": float(fei_md["Stage"]["StageY"]), 
+            "z": float(fei_md["Stage"]["StageZ"]),
+            "rx": float(fei_md["Stage"]["StageR"]),
+            "rz": float(fei_md["Stage"]["StageT"]),}
+            ,"m"
+    ]
+
+    # beam
+    beam_type = fei_md["Beam"]["Beam"]
+    scan_type = fei_md["Beam"]["Scan"]
```
pieleric, line 2718: duplicate lines -> remove

src/odemis/dataio/tiff.py
```diff
@@ -2613,3 +2624,121 @@ def _iterDirectories(tiff_file):
...
+        [float(fei_md[scan_type]["PixelWidth"]), 
+         float(fei_md[scan_type]["PixelHeight"])], "m"
+    ]
+    extra_md[beam_type]["power"] = [True, None]
+    extra_md[beam_type]["resolution"]= [
+        [int(fei_md["Image"]["ResolutionX"]), 
+         int(fei_md["Image"]["ResolutionY"])], "px"
+    ]
+    extra_md[beam_type]["probeCurrent"] = [float(fei_md[beam_type]["BeamCurrent"]), "A"]
+    extra_md[beam_type]["rotation"] = [float(fei_md[beam_type]["ScanRotation"]), "rad"]
```
pieleric, line 2736: You can store it in MD_ROTATION.

src/odemis/dataio/tiff.py
```diff
@@ -2613,3 +2624,121 @@ def _iterDirectories(tiff_file):
...
+    }
+    md[model.MD_ACQ_TYPE] = acq_map[beam_type]
+
+    # detector
+    md[model.MD_DET_TYPE] = (model.MD_DT_NORMAL 
+                                if fei_md["Detectors"]["Name"] in ["ETD", "TLD"] 
+                                else model.MD_DT_INTEGRATING)
+    
+    # system components
+    extra_md = {}
```
pieleric, line 2703: Note that this EXTRA_SETTINGS metadata is normally only used as info for the user (containing detailed settings about the hardware properties during an acquisition). It's never used for an actual functionality in Odemis). So it would be fine to completely skip it. Alternatively, you could also just copy-paste the whole "INI" string into it and be done. 
I don't think that trying to convert it into something that looks like it was acquired by Odemis is useful.

src/odemis/util/dataio.py
```diff
@@ -234,3 +234,21 @@ def splitext(path):
 
     root = path[:len(path) - len(ext)]
     return root, ext
+
+
+def _import_images_and_stitch(filenames: list):
```
pieleric, line 239: I think that it's a good proof-of-concept code for the current use-case (ie, loading TFS TIFF tiles and make a big image), but we probably shouldn't merge it as-is. 
Note in particular we already have something similar in the `odemis-convert` utility which also allows to select the registration and weaving methods. See the code in `odemis.cli.convert.stitch()`. 

I'm honestly not sure yet how this should be done in Odemis for the case of the METEOR. Maybe in the analysis tab, in the "Add stream" button, in addition to the "From file..." option you could add a "Add tile series..." which would allow the user to select multiple files, and they would be "weaved" into a stream. Eventually, in the METEOR, we'd call this function from a different place (somewhere in the localization tab) to load the SEM overview tiles. I'm happy to hear other options!

src/odemis/dataio/tiff.py
```diff
@@ -2613,3 +2624,121 @@ def _iterDirectories(tiff_file):
...
+            md[subsystem] = {}
+        else:
+            k, v = item.split("=")   
+            md[subsystem][k] = v
+    return md
+
+
+def _convertFEItoOdemisMD(metadata):
+    """
+    Converts FEI metadata to odemis metadata format
```
pieleric, line 2653: Could you add that this is to parse data from ThermoFisher Scientific SEMs? (I guess "FEI" is going to be less and less known).

src/odemis/dataio/tiff.py
```diff
@@ -2613,3 +2624,121 @@ def _iterDirectories(tiff_file):
             tiff_file.ReadDirectory()
             dir_index += 1
             yield dir_index
+
```
pieleric, line 2627: 2 empty lines are enough.

src/odemis/dataio/tiff.py
```diff
@@ -222,6 +226,13 @@ def _readTiffTag(tfile):
         except (OverflowError, ValueError):
             logging.info("Failed to parse date '%s'", val)
 
+    try:
+        fei_md = tfile.GetField(34682)
+        fei_md = fei_md.decode("latin1") # latin1 -> every char is acceptable
+        md.update(_convertFEItoOdemisMD(fei_md))
+    except Exception as e:
```
pieleric, line 233: Can you distinguish between "the tag is not present" (which is going to be 99% of the cases, and which doesn't even require a log message) and "there is FEI metadata but it looked weird", in which case this log is useful.

pieleric (changes requested):

src/odemis/acq/stitching/_weaver.py
```diff
@@ -57,6 +57,10 @@ def addTile(self, tile):
         """
         # Merge the correction metadata inside each image (to keep the rest of the
         # code simple)
+        if isinstance(tile, tuple):
+            tile = tile[0]
+        if isinstance(tile, model.DataArrayShadow):
```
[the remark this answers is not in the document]
patrickcleeve2: Moved the data load to the caller, added a type check explaining how to convert from DataArrayShadow -> DataArray. I think it would be better if it just loaded it automatically, but agree should be consistent with other functionality

src/odemis/dataio/tiff.py
patrickcleeve2: Ok, removing it for clarity

src/odemis/dataio/tiff.py
patrickcleeve2: Initially, I had MD_POS as 3D, but it isn't compatible with tiling which requires 2D POS, so I made it 2D. Alternatively I can change the transform to support mismatched dimensions (only take YX). 

Error from weave with 3D MD_POS

```bash
Exception has occurred: ValueError
matmul: Input operand 1 has a mismatch in its core dimension 0, with gufunc signature (n?,k),(k,m?)->(n?,m?) (size 3 is different from 2)
  File "/home/odemis/development/odemis/src/odemis/util/transform.py", line 969, in inverse
    translation = -numpy.matmul(matrix, self.translation)
  File "/home/odemis/development/odemis/src/odemis/util/img.py", line 1012, in rotate_img_metadata
    p1 = transl.inverse().apply(pos)
  File "/home/odemis/development/odemis/src/odemis/acq/stitching/_weaver.py", line 79, in getFullImage
    tiles.append(img.rotate_img_metadata(tile, -rotation, center_of_rot))
  File "/home/odemis/development/odemis/src/odemis/acq/stitching/_simple.py", line 111, in weave
    stitched_image = weaver.getFullImage()
  File "/home/odemis/development/odemis/src/odemis/util/dataio.py", line 252, in _import_images_and_stitch
    stitched_image = weave(images)
  File "/home/odemis/development/odemis/dev/main.py", line 31, in main
    stitched_image = _import_images_and_stitch(filenames)
  File "/home/odemis/development/odemis/dev/main.py", line 45, in <module>
    main()
ValueError: matmul: Input operand 1 has a mismatch in its core dimension 0, with gufunc signature (n?,k),(k,m?)->(n?,m?) (size 3 is different from 2)
```

src/odemis/dataio/tiff.py
patrickcleeve2: I attempted this initially, but couldn't get the parser to return anything except None. Tried different encodings and removing the returnline chars, but not sure. Open to suggestions

```python
import configparser
cfg = configparser.ConfigParser()
metadata = metadata.decode("latin1") # latin1 -> every char is acceptable
tfs_metadata = cfg.read_string(metadata)
```
tfs_metadata is None

src/odemis/dataio/tiff.py
patrickcleeve2: changed to guard (fei_md is None if tag is not present). Will only log if error while parsing / converting metadata

src/odemis/util/dataio.py
patrickcleeve2: OK, removed this from PR. Will test cli version. Agree on having separate gui functionality for loading tileset

patrickcleeve2: Able to replicate functionality with:

```bash
odemis-convert --tiles $(ls /path/to/data/) --output /home/odemis/development/test.tiff --registrar identity
```

src/odemis/dataio/tiff.py
```diff
@@ -2613,3 +2624,121 @@ def _iterDirectories(tiff_file):
...
+    date = datetime.strptime(datetime_str, "%m/%d/%Y %I:%M:%S %p")
+    md[model.MD_ACQ_DATE] = date.timestamp()
+    
+    md[model.MD_HW_NAME] = fei_md["System"]["SystemType"]+"-"+fei_md["System"]["Dnumber"]    
+    md[model.MD_SW_VERSION] = fei_md["System"]["Software"]
+    md[model.MD_PIXEL_SIZE] = (float(fei_md["Scan"]["PixelWidth"]), 
+                                  float(fei_md["Scan"]["PixelHeight"]))
+    
+    md[model.MD_DWELL_TIME] = float(fei_md["Scan"]["Dwelltime"])
+    md[model.MD_DIMS] = "YX"
```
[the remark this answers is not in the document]
pieleric: Yes, on 2D data MD_DIMS is not set, as it's the default. It's only necessary for RGB because RGB images are 3D (Y, X, 3), and Odemis assumes by default that 3D is ZYX.

src/odemis/cli/convert.py
```diff
@@ -210,12 +210,13 @@ def add_acq_type_md(das):
     return das
 
 
-def stitch(infns, registration_method, weaving_method):
+def stitch(infns: list, registration_method: int = REGISTER_IDENTITY, weaving_method: int = WEAVER_MEAN) -> list:
```
pieleric, line 213: Please move this function to `odemis.util.dataio` as something like `open_and_stitch()`

src/odemis/cli/convert.py
```diff
@@ -210,12 +210,13 @@ def add_acq_type_md(das):
...
     """
     Stitches a set of tiles.
-    infns: file names of tiles
-    method: weaving method (WEAVER_MEAN or WEAVER_COLLAGE)
-    returns list of data arrays containing the stitched images for every stream
+    :infns: file names of tiles
```
pieleric, line 216: Now the notation is like `:param infns:`

src/odemis/dataio/tiff.py
```diff
@@ -2613,3 +2627,127 @@ def _iterDirectories(tiff_file):
...
+def _convertTFStoOdemisMD(metadata: str) -> dict:
+    """
+    Converts ThermoFisher Scientific (TFS) metadata to odemis metadata format
+    :param: metadata : metadata string as ini (TFS format)
+    :return: metadata as dictionary (odemis format)
+    :raises KeyError: if the metadata cannot be parsed
+    """
+
+    # parse metadata as dictionary
+    fei_md = _parseINIMetadataAsDict(metadata)
```
pieleric, line 2671: Could you try `configparser` again? (Did you see my comment about it?)

pieleric (changes requested):

src/odemis/util/dataio.py
```diff
@@ -234,3 +239,93 @@ def splitext(path):
...
+    # If AR Stream is present, multiple data arrays are created. The data_to_static_streams
+    # function returns a single ARStream, so in this case many data arrays will not be assigned a
+    # stream and therefore also don't have an acquisition type.
+    for da in das:
+        if model.MD_ACQ_TYPE not in da.metadata:
+            da.metadata[model.MD_ACQ_TYPE] = model.MD_AT_AR
+    return das
+
+
+def open_files_and_stitch(infns: list, registration_method: int = REGISTER_IDENTITY, weaving_method: int = WEAVER_MEAN) -> list:
```
pieleric, line 274: Please add a test case in util/test/dataio_test.py . I'd suggest the code automatically creates fake tiles as TIFF files, and then calls `open_files_and_stitch()` and check the output is a single DataArray.

pieleric (changes requested):
It's starting to look pretty complete. I think the main missing parts are the test cases. Please also give a second try to configparser, to see if we can avoid the INI parser function totally. 

Eventually it'd be good to have just 3 commits: one for the fixes/improvements to acq.stitching, one for the `open_files_and_stitch` refactoring, and one for the dataio.tiff extension.

src/odemis/dataio/tiff.py
patrickcleeve2: resolved, was silly mistake on my end. now using configparser -> dict

src/odemis/dataio/test/tiff_test.py
```diff
@@ -2359,7 +2359,7 @@ def testConvertTFStoOdemisMD(self):
 
         # assert metadata
         md = {
-            model.MD_ACQ_DATE: 1699621755.0,
+            # model.MD_ACQ_DATE: 1699621755.0, breaks in CI?
```
pieleric, line 2362: There is no reason it breaks on the CI but not in a "normal" computer. So it's worth inverstigating. Do you have logs on how it failed (what was the actual value of MD_ACQ_DATE)? As it's a float, maybe you would need to use `assertAlmostEqual`.

patrickcleeve2: @pieleric I believe it is a timezone issue. The timestamps are exactly 1hr apart. This is because the TFS metadata doesn't have any timezone information, just a Date, Time string. Therefore, it just uses the default timezone of wherever the code is running, which is utc in CI, and whatever the timezone is set for on the Odemis computer. I can explicitly set the timezone in the parser so the tests pass, but that might be confusing to users rather than seeing it in their local timezone. What do you think?


CI Test
```bash
======================================================================
FAIL: testConvertTFStoOdemisMD (tiff_test.TestTiffIO)
  File "/home/runner/work/odemis/odemis/src/odemis/dataio/test/tiff_test.py", line 2384, in testConvertTFStoOdemisMD
AssertionError: 1699621755.0 != 1699625355.0
```

Local Test
``` bash
odemis@odemis-sim:~/development/odemis/src/odemis/dataio/test$ python3 -m unittest tiff_test.TestTiffIO.testConvertTFStoOdemisMD
WARNING:root:Failed to load optimised functions, slow version will be used.
INFO:root:Failed to use the OME data (No OME XML data found), will use standard TIFF
timestamp: 1699621755.0
.
Ran 1 test in 0.021s

OK
```

Explictly setting tz=utc
```bash

odemis@odemis-sim:~/development/odemis/src/odemis/dataio/test$ python3 -m unittest tiff_test.TestTiffIO.testConvertTFStoOdemisMD
WARNING:root:Failed to load optimised functions, slow version will be used.
INFO:root:Failed to use the OME data (No OME XML data found), will use standard TIFF
timestamp: 1699625355.0
timezone: UTC
.
Ran 1 test in 0.009s

OK
```


Checking Timestamps manually
```python
In [1]: timestamp = 1699625355.0

In [2]: import datetime

In [3]: date = datetime.datetime.fromtimestamp(timestamp)

In [4]: date
Out[4]: datetime.datetime(2023, 11, 11, 1, 9, 15)

In [5]: timestamp2 = 1699621755.0

In [6]: date = datetime.datetime.fromtimestamp(timestamp2)

In [7]: date
Out[7]: datetime.datetime(2023, 11, 11, 0, 9, 15)

In [8]: timestamp - timestamp2
Out[8]: 3600.0

```

pieleric: From the data bar, it seems that the time is stored as the local time (as they match, and this image was acquired in Australia, so if the time was stored as UTC internally it would be a quite different value!). 
So I think your code in tiff.py does the right thing. We just need to find the right way to test it.
One solution is to not check MD_ACQ_DATE... but it's a little sad.
The alternative is to just explicitly encode the time as a string (in the test case) and convert it to the local time.

patrickcleeve2: Ok, I've push a fix to explicitly encode the string in the test case, so it now works (CI still failing on 20.04 due to unrelated speed test)

pieleric (changes requested):
Looks almost ready to merge. Mostly just small comments on the tests.

src/odemis/dataio/tiff.py
```diff
@@ -2613,3 +2628,116 @@ def _iterDirectories(tiff_file):
...
+    subsystem:
+        k: v,
+        ...
+    }
+    """
+    md = {}
+    try:
+        cfg = configparser.ConfigParser()
+        cfg.read_string(metadata)
+        md = {s:dict(cfg.items(s)) for s in cfg.sections()}
```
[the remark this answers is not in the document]
patrickcleeve2: I've inlined the function, but think it should still be converted to a dict as it makes the access clearer. Otherwise, you need to index the lists, which is confusing to read. 

When converted to dict
``` python
date = fei_md["User"]["date"]

```

Using cfg directly
```python
#  (cfg.items("User") -> [('date', '05/09/2023'), ('time', '04:29:30 PM'), ('user', 'Supervisor'), ('usertext', ''), 
# ('usertextunicode', '')])
date = cfg.items("User")[0][1]

```

src/odemis/util/dataio.py
```diff
@@ -24,10 +24,14 @@
...
 from odemis import model
-from odemis.acq import stream
+from odemis.acq import stream, stitching
 from odemis.model import MD_WL_LIST, MD_TIME_LIST, MD_THETA_LIST
 import os
 
+from odemis.acq.stream import StaticSEMStream, StaticCLStream, StaticSpectrumStream, \
+                              StaticARStream, StaticFluoStream
+from odemis.acq.stitching import WEAVER_MEAN, WEAVER_COLLAGE, WEAVER_COLLAGE_REVERSE, \
+                                REGISTER_SHIFT, REGISTER_IDENTITY, REGISTER_GLOBAL_SHIFT
```
K4rishma, line 34:
```suggestion
from odemis.acq.stitching import REGISTER_IDENTITY, REGISTER_GLOBAL_SHIFT, REGISTER_SHIFT, \
                              WEAVER_COLLAGE, WEAVER_COLLAGE_REVERSE, WEAVER_MEAN
from odemis.acq.stream import StaticARStream, StaticCLStream, StaticFluoStream, StaticSEMStream,  \ 
                              StaticSpectrumStream
                               
```

src/odemis/util/dataio.py
```diff
@@ -234,3 +238,93 @@ def splitext(path):
...
+    :param registration_method: method used for registration
+    :param weaving_method: method used for weaving
+    :return: list of data arrays containing the stitched images for every stream
+    """
+
+    def leader_quality(da):
+        """
+        Function for sorting different streams. Use largest EM stream first, then other EM streams,
+        then other types of streams sorted by their size.
+        return int: The bigger the more leadership
```
K4rishma, line 286:
```suggestion
        return: (int) The bigger the more leadership
```

src/odemis/util/dataio.py
```diff
@@ -234,3 +238,93 @@ def splitext(path):
...
+    st_data = []
+    for s in range(len(das_registered[0])):
+        streams = []
+        for da in das_registered:
+            streams.append(da[s])
+        da = stitching.weave(streams, weaving_method)
+        da.metadata[model.MD_DIMS] = "YX"
+        st_data.append(da)
+
+    return st_data
```
K4rishma, line 330: To add a blank line at the end

src/odemis/util/test/dataio_test.py
```diff
@@ -19,15 +19,17 @@
 import time
 import unittest
 import warnings
+import os
```
K4rishma, line 22: It can be added before line 18

src/odemis/util/test/dataio_test.py
```diff
@@ -19,15 +19,17 @@
...
 import numpy
 
 from odemis import model
 from odemis.acq import stream
 from odemis.dataio import tiff
 from odemis.util.dataio import (_split_planes, data_to_static_streams,
-                                open_acquisition, splitext)
-
+                                open_acquisition, splitext, open_files_and_stitch)
+from odemis.util import testing
```
K4rishma, line 31: before line 29

src/odemis/util/test/dataio_test.py
```diff
@@ -19,15 +19,17 @@
...
 
 import numpy
 
 from odemis import model
 from odemis.acq import stream
 from odemis.dataio import tiff
 from odemis.util.dataio import (_split_planes, data_to_static_streams,
-                                open_acquisition, splitext)
-
+                                open_acquisition, splitext, open_files_and_stitch)
```
K4rishma, line 30: alphabetic order after import

src/odemis/util/dataio.py
K4rishma, line 286:
```suggestion
        return: (int) The bigger the more leadership
```

K4rishma:
Just adding minor formatting suggestions, rest looks good

patrickcleeve2:
Thanks @K4rishma, i've resolved those issues. 

@pieleric I think this should be ready to merge now

src/odemis/dataio/tiff.py
```diff
@@ -2613,3 +2628,93 @@ def _iterDirectories(tiff_file):
...
+    """
+
+    # parse ini metadata as dictionary
+    fei_md = {}
+    try:
+        cfg = configparser.ConfigParser()
+        cfg.read_string(metadata)
+        fei_md = cfg
+    except Exception as e:
+        logging.debug("Failed to parse metadata as ini: %s", e)
```
pieleric, line 2649: Probably worthy more than a debug message. I'd go for a  warning, or even exception.

Also, please immediately `return md`, so that this is not followed by every metadata read raising an exception.

src/odemis/gui/cont/tabs.py
```diff
@@ -3560,7 +3561,7 @@ def __init__(self, name, button, panel, main_frame, main_data):
     def stream_bar_controller(self):
         return self._stream_bar_controller
 
-    def select_acq_file(self, extend=False):
+    def select_acq_file(self, extend=False, tileset: bool = False):
         """ Open an image file using a file dialog box
 
         extend (bool): if False, will ensure that the previous streams are closed.
```
pieleric, line 3567: Please update the docstring with info about the "tileset" parameter.

pieleric (changes requested):
Found 2 tiny issues. I think after they are fixed it will be fully ready to merge.

pieleric (approved):

PR #2557, opened 2023-11-06
[fix] Don't round numbers in plot MarkingLine overlay

pieleric:
That overlay allows the user to read precisely the value of a plot.
However, it used to round large number to sometimes just 3 digits.
So numbers like 123456 would be shown "123000". That's good usage
for the axis ticks, but not in the plot where we want as much info as
possible.
In practice this means that for int's the value will now be always fully
shown.
For floats (which is rare), the value will be either rounded to the
closest int, or if its a small number, show at least 4 digits.

### DIFF
--- a/src/odemis/gui/comp/overlay/view.py
+++ b/src/odemis/gui/comp/overlay/view.py
@@ -24,11 +24,9 @@ This file is part of Odemis.
 import cairo
 import logging
 import math
-import numbers
 import numpy
 from odemis import util
 from odemis.gui.comp.overlay.base import Label, Vec, ViewOverlay
-from odemis.gui.util import img
 from odemis.gui.util.conversion import change_brightness
 from odemis.util import peak
 import wx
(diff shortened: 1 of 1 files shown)


src/odemis/gui/comp/overlay/view.py
```diff
@@ -616,10 +616,12 @@ def draw(self, ctx):
                 h_draw = True
                 v_draw = True
 
-            self.x_label = units.readable_str(val[0], self.cnvs.unit_x,
-                                              img.guess_sig_num_rng(self.cnvs.range_x, val[0]))
-            self.y_label = units.readable_str(val[1], self.cnvs.unit_y,
-                                              img.guess_sig_num_rng(self.cnvs.range_y, val[1]))
```
FCanberk, line 622: there isn't any usage of the `img` anymore, since the only usage is removed here. Can we also remove the import from the file(line 31)?

patrickcleeve2 (approved):
works as expected

PR #2558, opened 2023-11-06
[ci] Fix unit test util.img.test_fast failing 50% of the time

pieleric:
The CI test cases didn't build the cython files. So the img_fast was not available. As util.img automatically uses the standard numpy implementation if the fast implementation is not available, the test case still ran... but it ended comparing twice the same implementation. So 50% of the time it failed.

Two different ways to fix the unit test failures:
* Actually compile the fast version so that it's possible to compare fast vs normal
* Detect that the fast version is not available, and skip the test.

We do both, to make sure it's really not going to come back!

### DIFF
--- a/src/odemis/util/test/img_test.py
+++ b/src/odemis/util/test/img_test.py
@@ -451,6 +451,11 @@ class TestDataArray2RGB(unittest.TestCase):
 
     def test_fast(self):
         """Test the fast conversion"""
+        try:
+            import odemis.util.img_fast
+        except ImportError as ex:
+            self.skipTest(f"img_fast not available ({ex}), cannot test it")
+
         data = numpy.ones((251, 200), dtype="uint16")
         data[:, :] = numpy.arange(200)
         data[2, :] = 56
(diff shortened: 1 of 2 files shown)


src/odemis/util/test/img_test.py
```diff
@@ -451,6 +451,11 @@ def test_irange(self):
 
     def test_fast(self):
         """Test the fast conversion"""
+        try:
+            import odemis.util.img_fast
+        except ImportError as ex:
+            self.skipTest(f"img_fast not available ({ex}), cannot test it")
+
         data = numpy.ones((251, 200), dtype="uint16")
```
patrickcleeve2, line 459: test_speed is also intermittently failing on 20.04, is this related? 

```
======================================================================
FAIL: test_speed (src.odemis.util.test.img_test.TestFindOptimalRange)
  File "/home/runner/work/odemis/odemis/src/odemis/util/test/img_test.py", line 124, in test_speed
    self.assertLessEqual(dur_sc, dur_full)
AssertionError: 0.23967909812927246 not less than or equal to 0.20863556861877441

```

patrickcleeve2 (approved):
cython compiles, tests passed. Added a question about maybe related failure

nandishjpatel (approved):

src/odemis/util/test/img_test.py
pieleric: That's annoying... all these speed related tests are quite unreliable (especially in the "cloud"). That issue is not directly related as the "fast" and "slow" paths it tests are both written in numpy. So let's keep it for another time...

PR #2561, opened 2023-11-09
[feature] add odemis-start-turret to automatically turn on and change turret of a spectrograph

pieleric:
Useful to workaround if a spectrograph fails to read the turret ID.
Note that shrkconfig --turret alsmost does the same thing, but it's not
able to turn on the spectrograph in a SPARCv2. This script does that
trick too.

### DIFF
--- a/scripts/odemis-start-turret.py
+++ b/scripts/odemis-start-turret.py
@@ -1,0 +1,63 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+# Starts the SPARC spectrograph and select a specific turret
+# To start automatically with Odemis, run with:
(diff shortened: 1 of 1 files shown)


scripts/odemis-start-turret.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+# Starts the SPARC spectrograph and select a specific turret
+# To start automatically with Odemis, run with:
+# sh -c "env PYTHONPATH=$HOME/development/odemis/src/ $HOME/development/odemis/scripts/odemis-start-turret.py 1 && odemis-start"
```
FCanberk, line 5: to follow the standardized docstrings; could you  add a docstring with creation date, author etc?

FCanberk (approved):

PR #2564, opened 2023-11-10
[perf][METEOR-590] save z axis of meteor stage in feature position

K4rishma:
Now z axis of meteor stage will be used to move to a selected feature along with x, y position of meteor stage and z position of focus

### DIFF
--- a/src/odemis/acq/feature.py
+++ b/src/odemis/acq/feature.py
@@ -12,19 +12,23 @@ class CryoFeature(object):
     Model class for a cryo interesting feature
     """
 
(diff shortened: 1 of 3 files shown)


pieleric (commented):
Let's try to see if there is a better way to adjust the stage Z with the focus Z (see Jira task).
Also, I'm a bit worried about what will this code do on the MIMAS and ENZEL as there the focus is changed by directly moving the stage Z. So the 2 values are going to be 100% redundant. Can we make sure they never disagree with each other? Is it going to be a problem to ask moving twice to the same position in two different ways?

PR #2565, opened 2023-11-10
[perf][METEOR-527] save tilt and rotation of sem position

K4rishma:
during stage switching from fm to sem imaging mode, previous rx value in sem imaging is restored

### DIFF
--- a/plugins/meteor_engage_warn.py
+++ b/plugins/meteor_engage_warn.py
@@ -21,7 +21,9 @@ see http://www.gnu.org/licenses/.
 
 import logging
 import math
-from odemis.acq.move import FM_IMAGING, SEM_IMAGING, MicroscopePostureManager
+
+from odemis import model
+from odemis.acq.move import FM_IMAGING, SEM_IMAGING
 from odemis.gui.plugin import Plugin
 from odemis.util.units import readable_str
 from typing import Dict
@@ -75,16 +77,29 @@ class MeteorEngageWarnPlugin(Plugin):
         pos_str = "\n". join(pos_str)
 
         # Guess (back) which position the users wants to go to
-        target_pos = self.main_app.main_frame.getCurrentPostureLabel(end_pos)
+        target_pos = self.main_app.main_data.posture_manager.getCurrentPostureLabel(end_pos)
+        stage = self.main_app.main_data.stage_bare
         if target_pos == FM_IMAGING:
             warn_msg = FM_WARN_MSG
+            # Check the deviation in rotation angle when
+            # switching from SEM to FM,
+            # give a warning message if switching is done from a different rotation angle
+            stage_md = stage.getMetadata()
+            fav_angles = stage_md[model.MD_FAV_SEM_POS_ACTIVE]
+            axis_name = "rz" if "rz" in fav_angles else "rm"
+            current_angle = stage.position.value[axis_name]
+            fav_angle = fav_angles[axis_name]
+
+            if not math.isclose(current_angle, fav_angle):
+                warn_msg = ("The current rotation value is different from the desired value. The switching behavior"
+                            " may not be proper.\n\n") + FM_WARN_MSG
         elif target_pos == SEM_IMAGING:
             warn_msg = SEM_WARN_MSG
         else:
             logging.warning("Unexpected target position %s", target_pos)
             warn_msg = SEM_WARN_MSG
 
-        box = wx.MessageDialog(self.main_frame,
+        box = wx.MessageDialog(self.main_app.main_frame,
                                warn_msg % (pos_str,),
                                caption="Large move of the stage",
                                style=wx.YES_NO | wx.ICON_QUESTION | wx.CENTER)
--- a/src/odemis/acq/move.py
+++ b/src/odemis/acq/move.py
@@ -508,6 +508,12 @@ class MeteorTFS1PostureManager(MeteorPostureManager):
                 sub_moves.append((self.stage, filter_dict({'x', 'y', 'z'}, target_pos)))
                 sub_moves.append((self.stage, filter_dict({'rx', 'rz'}, target_pos)))
             elif target in (LOADING, SEM_IMAGING, FM_IMAGING):
+                # save rotation and tilt in SEM before switching to FM imaging
+                # to restore rotation and tilt while switching back from FM -> SEM
+                if current_label == SEM_IMAGING and target == FM_IMAGING:
+                    current_value = self.stage.position.value
+                    self.stage.updateMetadata({model.MD_FAV_SEM_POS_ACTIVE: {'rx': current_value['rx'],
+                                                                             'rz': current_value['rz']}})
                 # Park the focuser for safety
                 if not isNearPosition(self.focus.position.value, focus_deactive, self.focus.axes):
                     sub_moves.append((self.focus, focus_deactive))
@@ -828,6 +834,11 @@ class MeteorZeiss1PostureManager(MeteorPostureManager):
                         sub_moves.append((stage, filter_dict({'rx'}, target_pos)))
 
                     if current_label == SEM_IMAGING:
+                        # save rotation and tilt in SEM before switching to FM imaging
+                        # to restore rotation and tilt while switching back from FM -> SEM
+                        current_value = self.stage.position.value
+                        self.stage.updateMetadata({model.MD_FAV_SEM_POS_ACTIVE: {'rx': current_value['rx'],
+                                                                                 'rm': current_value['rm']}})
                         # when switching from SEM to FM
                         # move in the following order :
                         sub_moves.append((stage, filter_dict({'rx', 'rm', 'x', 'y', 'm', 'z'}, target_pos)))
@@ -1140,6 +1151,11 @@ class MeteorTescan1PostureManager(MeteorPostureManager):
                         sub_moves.append((stage, filter_dict({'z'}, target_pos)))
 
                     if current_label == SEM_IMAGING:
+                        # save rotation and tilt in SEM before switching to FM imaging
+                        # to restore rotation and tilt while switching back from FM -> SEM
+                        current_value = self.stage.position.value
+                        self.stage.updateMetadata({model.MD_FAV_SEM_POS_ACTIVE: {'rx': current_value['rx'],
+                                                                                 'rz': current_value['rz']}})
                         # when switching from SEM to FM
                         # move in the following order :
                         sub_moves.append((stage, filter_dict({'z'}, target_pos)))
--- a/src/odemis/gui/cont/tabs.py
+++ b/src/odemis/gui/cont/tabs.py
@@ -3255,8 +3255,24 @@ class CryoChamberTab(Tab):
                 else:
                     pos_str.append(f"{axis} = " + readable_str(end_pos[axis], "m", 4))
         pos_str = "\n". join(pos_str)
+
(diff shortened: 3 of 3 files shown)


src/odemis/acq/move.py
```diff
@@ -368,6 +368,9 @@ def getTargetPosition(self, target_pos_lbl: int) -> Dict[str, float]:
                     sem_grid1_pos = stage_md[model.MD_SAMPLE_CENTERS][POSITION_NAMES[GRID_1]]
                     fm_target_pos = self._transformFromSEMToMeteor(sem_grid1_pos)
                 elif current_position == SEM_IMAGING:
+                    current_value = self.stage.position.value
+                    self.stage.updateMetadata({model.MD_FAV_SEM_POS_ACTIVE: {'rx': current_value['rx'],
+                                                                             'rz': current_value['rz']}})
```
pieleric, line 373: I'm not very enthusiastic in putting this code here because the function is called `getTargetPosition()`. It implies to reads/compute data, but doesn't change it. Admittedly, there are enough checks (ie, current pos is in SEM_IMAGING, and target_pos is FM_IMAGING) that it should be safe... but that doesn't feel right. 
I've looked at the `transformFromSEMToMeteor` methods: none of them use `MD_FAV_SEM_POS_ACTIVE`. So it's not really needed to update the metadata so "early" anyway. 

Hence, I think it'd be safer/clearer to place that metadata update to just before moving to the FM_IMAGING posture, inside `_doCryoSwitchSamplePosition()`. There 3 versions, but I think they all have a path `if target == FM_IMAGING:` -> `if current_label == SEM_IMAGING:`. That would go right in this block.

src/odemis/acq/move.py
```diff
@@ -368,6 +368,9 @@ def getTargetPosition(self, target_pos_lbl: int) -> Dict[str, float]:
                     sem_grid1_pos = stage_md[model.MD_SAMPLE_CENTERS][POSITION_NAMES[GRID_1]]
                     fm_target_pos = self._transformFromSEMToMeteor(sem_grid1_pos)
                 elif current_position == SEM_IMAGING:
```
pieleric, line 370: Please add a comment on top of this code explaining why it's a good idea to copy the current position and update the metadata with it. (I think the rational is that this allows to remember the SEM position (tilt) and restore it when going back from FM_IMAGING to SEM_IMAGING.

pieleric (changes requested):
Please change the title of this PR to contain the Jira task ID.

src/odemis/acq/move.py
K4rishma: I feel _doCryoSwitchSamplePosition() moves to a target position in a set order without changing the values. The transform_to_() functions do the calculation. For clarity, will it be useful to add this update metadata in the transform functions, if not in _getTargetPosition?

src/odemis/acq/move.py
```diff
@@ -828,6 +834,11 @@ def _doCryoSwitchSamplePosition(self, future, target):
                         sub_moves.append((stage, filter_dict({'rx'}, target_pos)))
 
                     if current_label == SEM_IMAGING:
+                        # save rotation and tilt in SEM before switching to FM imaging
+                        # to restore rotation and tilt while switching back from FM -> SEM
+                        current_value = self.stage.position.value
+                        self.stage.updateMetadata({model.MD_FAV_SEM_POS_ACTIVE: {'rx': current_value['rx'],
+                                                                                 'rz': current_value['rz']}})
```
patrickcleeve2, line 841: Zeiss uses "rm" instead of "rz"

plugins/meteor_engage_warn.py
```diff
@@ -75,9 +77,26 @@ def _display_meteor_pos_warning_msg(self, end_pos: Dict[str, float]) -> bool:
...
+            # switching from SEM to FM,
+            # give a warning message if switching is done from a different rotation angle
+            stage_md = stage.getMetadata()
+            fav_angles = stage_md[model.MD_FAV_SEM_POS_ACTIVE]
+            for angle in ("rz", "rm"):
+                if angle in fav_angles:
+                    current_angle = stage.position.value[angle]
+                    fav_angle = fav_angles[angle]
+                    break
+            if current_angle != fav_angle:
```
patrickcleeve2, line 97: Wouldn't directly check floats like this, insteasd would use math.isclose() / np.isclose(). Otherwise it can fail for very small numerical differences

plugins/meteor_engage_warn.py
```diff
@@ -75,9 +77,26 @@ def _display_meteor_pos_warning_msg(self, end_pos: Dict[str, float]) -> bool:
...
+        current_angle = 0
+        fav_angle = 0
+        angle = ''
         if target_pos == FM_IMAGING:
             warn_msg = FM_WARN_MSG
+            # Check the deviation in rotation angle when
+            # switching from SEM to FM,
+            # give a warning message if switching is done from a different rotation angle
+            stage_md = stage.getMetadata()
+            fav_angles = stage_md[model.MD_FAV_SEM_POS_ACTIVE]
```
patrickcleeve2, line 91: Would avoid using a loop here, instead maybe something like:

```python
fav_angles = stage_md[model.MD_FAV_SEM_POS_ACTIVE]
key = "rx" if "rx" in fav_angles else "rm"
current_angle = stage.position.value[key]
fav_angle = fav_angles[key]
```

src/odemis/gui/cont/tabs.py
```diff
@@ -3255,10 +3255,35 @@ def _display_meteor_pos_warning_msg(self, end_pos) -> bool:
...
+        if target_label == FM_IMAGING:
+            stage_md = self._stage.getMetadata()
+            fav_angles = stage_md[model.MD_FAV_SEM_POS_ACTIVE]
+            for angle in ("rz", "rm"):
+                if angle in fav_angles:
+                    current_angle = self._stage.position.value[angle]
+                    fav_angle = fav_angles[angle]
+                    break
+
+        if current_angle != fav_angle and target_label == FM_IMAGING:
```
patrickcleeve2, line 3275: Same comment as above about comparing floats, and avoiding the loop

patrickcleeve2 (changes requested):

src/odemis/gui/cont/tabs.py
```diff
@@ -3255,10 +3255,33 @@ def _display_meteor_pos_warning_msg(self, end_pos) -> bool:
...
+        if target_label == FM_IMAGING:
+            stage_md = self._stage.getMetadata()
+            fav_angles = stage_md[model.MD_FAV_SEM_POS_ACTIVE]
+            key = "rz" if "rz" in fav_angles else "rm"
+            current_angle = self._stage.position.value[key]
+            fav_angle = fav_angles[key]
+
+        if not math.isclose(current_angle, fav_angle) and target_label == FM_IMAGING:
+            box = wx.MessageDialog(self.main_frame,
+                                   "The current " + angle + " value is different from the desired value. "
```
patrickcleeve2, line 3275: angle doesnt get set correctly, should be 'key'

src/odemis/gui/cont/tabs.py
```diff
@@ -3255,10 +3255,33 @@ def _display_meteor_pos_warning_msg(self, end_pos) -> bool:
...
+                                   "The switching behavior may not be proper." + "\n\nThe stage will move to "
+                                   "this position:\n" + pos_str + "\n\n Is it safe?",
+                                   caption="Large move of the stage",
+                                   style=wx.YES_NO | wx.ICON_QUESTION | wx.CENTER)
+        elif math.isclose(current_angle, fav_angle):
```
patrickcleeve2, line 3280: This always goes off if the target isn't FM_IMAGING is that expected?. current_angle, fav_angle are 0

K4rishma: That is expected behavior. To make it clear, I think I should just put else and not elif.

patrickcleeve2 (approved):

src/odemis/gui/cont/tabs.py
```diff
@@ -3255,10 +3255,33 @@ def _display_meteor_pos_warning_msg(self, end_pos) -> bool:
...
+            current_angle = self._stage.position.value[key]
+            fav_angle = fav_angles[key]
+
+        if not math.isclose(current_angle, fav_angle) and target_label == FM_IMAGING:
+            box = wx.MessageDialog(self.main_frame,
+                                   "The current " + key + " value is different from the desired value. "
+                                   "The switching behavior may not be proper." + "\n\nThe stage will move to "
+                                   "this position:\n" + pos_str + "\n\n Is it safe?",
+                                   caption="Large move of the stage",
+                                   style=wx.YES_NO | wx.ICON_QUESTION | wx.CENTER)
```
pieleric, line 3279: I think it's nicer to use the same structure as in the plugin: in the `if` block, set a `warn_msg` string, dependent on the angle, and after the block, do one `box = wx.MessageDialog(self.main_frame, warn_msg,...)`.

pieleric (changes requested):

pieleric (changes requested):

src/odemis/gui/cont/tabs.py
```diff
@@ -3254,9 +3260,25 @@ def _display_meteor_pos_warning_msg(self, end_pos) -> bool:
                     pos_str.append(f"{axis} = " + readable_str(math.degrees(end_pos[axis]), "°", 4))
                 else:
                     pos_str.append(f"{axis} = " + readable_str(end_pos[axis], "m", 4))
-        pos_str = "\n". join(pos_str)
+        pos_str = """\n""". join(pos_str)
+
+        # Check the deviation in rotation angle when switching from SEM to FM,
+        # give a warning message if switching is done from a different rotation angle
+        target_label = self.tab_data_model.main.posture_manager.getCurrentPostureLabel(end_pos)
+        warn_msg = STAGE_WARN_MSG
```
K4rishma, line 3268: @pieleric I have made a global variable for the the multiline message. If the same multiline message needs to be included in this function, it looks like below
image

src/odemis/gui/cont/tabs.py
```diff
@@ -3254,9 +3260,25 @@ def _display_meteor_pos_warning_msg(self, end_pos) -> bool:
                     pos_str.append(f"{axis} = " + readable_str(math.degrees(end_pos[axis]), "°", 4))
                 else:
                     pos_str.append(f"{axis} = " + readable_str(end_pos[axis], "m", 4))
-        pos_str = "\n". join(pos_str)
+        pos_str = """\n""". join(pos_str)
```
pieleric, line 3263: Why triple quotes? 
```suggestion
        pos_str = "\n".join(pos_str)
```

plugins/meteor_engage_warn.py
```diff
@@ -75,9 +77,22 @@ def _display_meteor_pos_warning_msg(self, end_pos: Dict[str, float]) -> bool:
         pos_str = "\n". join(pos_str)
 
         # Guess (back) which position the users wants to go to
-        target_pos = self.main_app.main_frame.getCurrentPostureLabel(end_pos)
+        target_pos = self.main_app.main_data.posture_manager.getCurrentPostureLabel(end_pos)
+        stage = self.main_app.main_data.get_component("stage")
```
pieleric, line 81: Does this work?! Did you test this plugin? I don't see any `get_component()` method.
Typically it should be:
`stage = self.main_app.main_data.stage`

pieleric (changes requested):

PR #2567, opened 2023-11-16
[fix] auto brightness/contrast and histrogram display could sometimes return shifted values

pieleric:
This fixes display issues related to the histrogram in the GUI, and should also speed up a little bit the histrogram computation (not that it was really noticeable anyway).
Also makes some test cases pass again.

### DIFF
--- a/src/odemis/acq/stream/_base.py
+++ b/src/odemis/acq/stream/_base.py
@@ -872,8 +872,8 @@ class Stream(object):
                         # => Two options:
                         # * add 10% margin on each side
                         # * round to the nearest power of 2 on each side
-                        # Take whichever is the smallest... and also ensure that the length is a multiple of 256 because
-                        # the histogram computation is much faster in such case.
+                        # Take whichever is the smallest... and also ensure that the length (mx - mn + 1)
+                        # is a multiple of 256 because the histogram computation is much faster in such case.
                         width10 = max(127, int((mx - mn) / 10))  # 10%, and at least 8 bits
 
                         # Min: only change the range if the previous one doesn't fit anymore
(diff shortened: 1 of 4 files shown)


src/odemis/acq/stream/_base.py
```diff
@@ -886,7 +886,6 @@ def _updateDRange(self, data=None):
                                 mn_p2 = 0
                             elif mn > 0:
                                 mn_p2 = 2 ** int(math.floor(math.log(mn + 1, 2)))
```
pieleric, line 888: This still causes some issues... when there is only one value, which is a multiple of 256 (mn > mx).

src/odemis/acq/stream/_base.py
```diff
@@ -897,26 +896,32 @@ def _updateDRange(self, data=None):
                             mn = prev_drange[0]
 
                         # Max: only change the range if the previous one doesn't fit anymore
+                        # Round to 256, and subtract 1, so that the length of the range is a multiple of 256.
                         if prev_drange is None or mx > prev_drange[1]:
                             mx10 = mx + width10
-                            mx10 = math.ceil(mx10 / 256) * 256
+                            mx10 = math.ceil(mx10 / 256) * 256 - 1  # so that the length a multiple of 256
```
sneepstefan, line 902:
```suggestion
                            mx10 = math.ceil(mx10 / 256) * 256 - 1  # to make sure the length is a multiple of 256
```

src/odemis/util/test/img_test.py
```diff
@@ -312,6 +312,23 @@ def test_compact(self):
...
+        test the compactHistogram()
+        """
+        # Histogram 1 pixel "too" long -> 1025px -> bins of 5 px -> length of 205px (multiple of bin size)
+        hist = numpy.ones(1025, dtype="uint16")
+        # make it compact
+        chist = img.compactHistogram(hist, 256)
+        self.assertEqual(len(chist), 205)
+        self.assertEqual(numpy.sum(chist), numpy.sum(hist))
+
+        # Hisogram really not the same size: 401px -> bins of 2px -> length of 201px
```
sneepstefan, line 326:
```suggestion
        # Histogram deliberately not the same size: 401px -> bins of 2px -> length of 201px
```

sneepstefan (changes requested):

sneepstefan (approved):

patrickcleeve2 (approved):

PR #2570, opened 2023-11-17
[fix] xtclient remove aperture index as it is not supported by TFS 

tepals:
aymore.

TFS removed reading and setting the aperture index from their interface. As we do not use it anyway, it is save to remove it. Also removed the TODO about the beamlet index updating when changing the multibeam, because the beamlet index is not updated when the multibeam mode is changed.

### DIFF
--- a/src/odemis/driver/xt_client.py
+++ b/src/odemis/driver/xt_client.py
@@ -1148,36 +1148,6 @@ class SEM(model.HwComponent):
             self.server._pyroClaimOwnership()
             return self.server.mpp_orientation_info()
 
-    def get_aperture_index(self):
-        """
-        Get the current aperture index.
-
-        :return (int): Aperture index (typical range 0 - 14)
-        """
-        with self._proxy_access:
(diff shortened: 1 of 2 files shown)


src/odemis/driver/xt_client.py
```diff
@@ -1148,36 +1148,6 @@ def mpp_orientation_info(self):
             self.server._pyroClaimOwnership()
             return self.server.mpp_orientation_info()
 
-    def get_aperture_index(self):
```
K4rishma, line 1151: There are some tests related to aperture_index in xt_client_test.py such as def test_get_aperture_index(self):. That will be removed I think.

tepals: good catch, I fixed it

K4rishma (approved):

sneepstefan:
Title and description were splitted in a unlogical way, can you fix that?

sneepstefan (approved):

PR #2573, opened 2023-11-20
[refactor] Simplify comp.GridViewport

pieleric:
This component had two main issues:
* A lot of the functions in the interface were never used in Odemis
* It tried to be clever, but was clever differently from what the rest
  of the GUI wanted. So the GUI had to find tricks to make it behave
  as it really wanted it.

Also, it used to support 2,3,and 4 viewports in a grid, with some larger
than others. However, as it wasn't fitting the GUI, the code was
protected and never used.

=> Simplify the code to just show the viewports requested, in the order
requested.

Also simplify the test cases.

### DIFF
--- a/src/odemis/gui/comp/grid.py
+++ b/src/odemis/gui/comp/grid.py
@@ -22,204 +22,114 @@
 
 """
 import logging
+from typing import Sequence
+
 import wx
 
 from odemis.gui.util import AttrDict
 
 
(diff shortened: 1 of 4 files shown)


patrickcleeve2 (approved):
tested on meteor sim

FCanberk (approved):

PR #2576, opened 2023-11-23
[fix] [METEOR-598] hide the overview images in chamber tab

K4rishma:
when overview images are made invisible in the localisation tab and, when new project is created in the chamber tab


src/odemis/gui/cont/tabs.py
```diff
@@ -499,6 +501,29 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
+        # All the overview streams
+        acquired_streams = set()
+        acquired_streams.update(self.tab_data_model.overviewStreams.value)
+        # Visible overview streams
+        streams = self.tab_data_model.views.value[0].stream_tree.streams
+        list_streams = []
+        for st in streams:
+            list_streams.append(st.stream)
+        # Invisible overview streams
+        unused_streams = acquired_streams.difference(set(list_streams))
```
patrickcleeve2, line 518: I would probably rename these to for clarity

```python
all_ov_streams
visible_ov_streams
invisible_ov_streams

```

src/odemis/gui/cont/tabs.py
```diff
@@ -499,6 +501,29 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
+        the overview streams in the Chamber tab.
+        """
+        # All the overview streams
+        acquired_streams = set()
+        acquired_streams.update(self.tab_data_model.overviewStreams.value)
+        # Visible overview streams
+        streams = self.tab_data_model.views.value[0].stream_tree.streams
+        list_streams = []
+        for st in streams:
+            list_streams.append(st.stream)
```
patrickcleeve2, line 516: Seeing as we cast this to a set immediatly after, would recommend just creating the set from the beginning

```python
list_streams = set()
for st in streams:
    list_streams.add(st.stream)

```

or use a comprehension, which will give you the same thing:

```python
list_streams = {st.stream for st in streams}
```

src/odemis/gui/cont/tabs.py
```diff
@@ -499,6 +501,29 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
+        # Invisible overview streams
+        unused_streams = acquired_streams.difference(set(list_streams))
+
+        chamber_tab = self.main_data.getTabByName("cryosecom_chamber")
+        # Hide the invisible overview streams
```
patrickcleeve2, line 521: This removes / adds the stream to the chamber tab, not just toggles the visible. It might be better to add another function to just toggle the visibility in the chamber overview?

src/odemis/gui/cont/tabs.py
```diff
@@ -2877,6 +2902,9 @@ def _create_new_dir(self):
 
     def _reset_project_data(self):
         try:
+            streams = self.tab_data_model.views.value[0].stream_tree.streams
```
patrickcleeve2, line 2905: I think we can just use the existing, 
```python 
self._get_overview_view()
``` 
to get the required view

patrickcleeve2 (changes requested):
Works as described, except that the overviews are removed/ added from the chamber tab when the visibility is toggled. It might be more efficient to only toggle their visibility.

src/odemis/gui/cont/tabs.py
K4rishma: I have changed the names of visible and invisible streams, it brings more clarity than before. For all_ov_streams for acquired_streams, I think as acquired_streams is used in function names and variable names in the other part of the file, consistency wise acquired_streams would bring more clarity. What do you think?

src/odemis/gui/cont/tabs.py
K4rishma: added a new function which is used at this place and also in line 2911 at reset_project_data

src/odemis/gui/cont/tabs.py
```diff
@@ -2877,6 +2910,8 @@ def _create_new_dir(self):
 
     def _reset_project_data(self):
         try:
+            streams = self._get_overview_view().stream_tree.streams
```
patrickcleeve2, line 2913: Maybe move this line inside remove_overview_streams

patrickcleeve2 (approved):

src/odemis/gui/cont/tabs.py
K4rishma: I used remove_overview_streams at one more place i.e. in line 504 of function _on_overview_visible where it takes list of invisible streams as input.

patrickcleeve2: missed that 👍

src/odemis/gui/cont/tabs.py
```diff
@@ -2762,6 +2783,18 @@ def remove_overview_stream(self, overview_stream):
         except AttributeError:  # No overview view
             pass
 
+    def remove_overview_streams(self, streams):
```
pieleric, line 2786: This is very similar to `remove_overview_stream()`. That function is called just once. I'd suggest to remove that function, and update the caller to do `remove_overview_streams([s])`. This way you'll avoid having twice almost the same code.

src/odemis/gui/cont/tabs.py
```diff
@@ -499,6 +501,25 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
             )
 
+    def _on_overview_visible(self, val):
+        """
+        Apply the visibility status of overview streams in the Localisation tab to
+        the overview streams in the Chamber tab.
+        """
+        # All the overview streams
+        acquired_streams = set()
+        acquired_streams.update(self.tab_data_model.overviewStreams.value)
```
pieleric, line 511: Let's call them `ov_streams`, as acquired is a misnomer (because there are other acquired streams which are not in this list).
Also you can simply write:
`ov_streams = set(self.tab_data_model.overviewStreams.value)`

src/odemis/gui/cont/tabs.py
```diff
@@ -499,6 +501,25 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
+        # Visible overview streams
+        streams = self.tab_data_model.views.value[0].stream_tree.streams
+        visible_ov_streams = {st.stream for st in streams}
```
pieleric, line 514: This works because all the overview images use Projections, for now (that's why you have to write `st.stream`). However, it would be safer to handle both Stream and Projection. The View actually already has a function for that. So you can replace these 2 lines by:
`visible_ov_streams = set(self.tab_data_model.views.value[0].getStreams())`

src/odemis/gui/cont/tabs.py
```diff
@@ -2762,6 +2783,18 @@ def remove_overview_stream(self, overview_stream):
         except AttributeError:  # No overview view
             pass
 
+    def remove_overview_streams(self, streams):
+        """
+       Remove the overview static stream from the view with the given list of acquired static streams
+       :param streams: (list of StaticStream) the newly acquired static streams from the localization tab
+       """
```
pieleric, line 2790: The indentation is off by 1 space.

src/odemis/gui/cont/tabs.py
pieleric, line 2913:
```suggestion
            streams = self._get_overview_view().getStreams()
```

pieleric (approved):

PR #2577, opened 2023-11-24
[SDEV-1356] fix fastem test cases

tepals:
[fix] fastem test cases fixed after being broken due to overlap and scan-stage updates
    
Commit 0c32be368bd491b416d5973abefba227b62367da fixed that the overlap is not allowed to be zero in the technolution API. The commit also updated the default value of the overlap to be 0.06 instead of 0.0. The test cases were written for an overlap of 0.0 and failed because of this.
Commit 9f40d9af143c305871a172ee0ff5581ffa397836 added the scan-stage as an input argument for the acquire function and the AcquisitionTask, but the test cases that called these functions were not updated.
Because of these commits the test cases were failing, this has now been fixed by using an overlap of 0.06 in the test cases and using the scan-stage.

[add] fastem test case for expected stage location with rotation correction
    
9f40d9af143c305871a172ee0ff5581ffa397836 added the movement of the stage in the scan-stage direction, but there was no test case added.

[ci] fastem mocked test cases added to cloud hosted CI

### DIFF
--- a/src/odemis/acq/test/fastem_test.py
+++ b/src/odemis/acq/test/fastem_test.py
@@ -32,7 +32,6 @@ from unittest.mock import Mock
 import numpy
 
 import odemis
-from fastem_calibrations import configure_hw
 from odemis import model
 from odemis.acq import fastem, stream
 from odemis.acq.acqmng import SettingsObserver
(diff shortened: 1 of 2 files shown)


src/odemis/acq/test/fastem_test.py
```diff
@@ -684,6 +701,58 @@ def test_stage_movement(self):
...
+                           self.stage, self.scan_stage, self.ccd, self.beamshift, self.lens)
+        data, e = f.result()
+
+        self.assertIsNone(e)  # check no exceptions were returned
+
+        # get the last stage position (it is the center of the last field)
+        cur_position = (self.stage.position.value['x'], self.stage.position.value['y'])
+        # For a single field and a positive 45 degree rotation correction it is expected that the stage moves only in x.
+        # It should move by half the diagonal size of a field in x.
+        exp_pos_x = math.sqrt((res_x / 2. * px_size_x) ** 2 + (res_y / 2. * px_size_y) ** 2)
```
pieleric, line 750: For info this can be written like:
```suggestion
        exp_pos_x = math.hypot(res_x / 2 * px_size_x, res_y / 2 * px_size_y)
```

pieleric (approved):

PR #2578, opened 2023-11-24
[fix] driver zeiss: adjust move speed test for new simulator speed

pieleric:
Commit 9a75b494b5b (make changes in zeiss.py to make stage simulater faster)
made the simulated moves faster... which resulted in the move speed test
to fail.
=> Just reduce the expected duration.

### DIFF
--- a/src/odemis/driver/test/zeiss_test.py
+++ b/src/odemis/driver/test/zeiss_test.py
@@ -233,17 +233,20 @@ class TestSEM3Axes(unittest.TestCase):
         testing.assert_pos_almost_equal(self.stage.position.value, p)
 
         # Check that a long move takes time (ie, that it waits until the end of the move)
-        # It's tricky, because it always waits at least 1s.
+        # It's tricky, because it always waits at least 1s, so we need to do a big move (more than 1s).
+        # The simulator goes at 2mm/s (see ._speed_lin), so we need to move more than 2mm.
+        # 4mm should take 2s, so that give a good margin. Also, moving 4 mm in the X axis on a real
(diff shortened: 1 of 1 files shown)


src/odemis/driver/test/zeiss_test.py
```diff
@@ -238,12 +238,12 @@ def test_move_axes(self):
         tstart = time.time()
         self.stage.moveRelSync({"x": 1e-3})
         dur = time.time() - tstart
-        self.assertGreaterEqual(dur, 1.1, "1 mm move took only %g s" % dur)
+        self.assertGreaterEqual(dur, 1.01, "1 mm move took only %g s" % dur)
```
K4rishma, line 241: Could we show how we got to 1.01, if there is a formula ? In that way if there are any changes to the simulator, these cases can be independent.

K4rishma (approved):

src/odemis/driver/test/zeiss_test.py
pieleric: Well... it's not great. The simulator goes at 2mm/s . This requests a move of 1mm, so it should take 0.5s. But, the driver wait always at least 1s. So we are stuck to 1s (or a tiny bit more) unless the move is > 2mm. Maybe we should move by 3 mm....

patrickcleeve2 (approved):

src/odemis/driver/test/zeiss_test.py
pieleric: I've now changed the test case to move by 4mm. So it should take 2s. I've also added a comment explaining how to compute this value.

PR #2579, opened 2023-11-24
[fix] acq stiching registrar: fix detection of too large shift

pieleric:
Surprisingly, the global shift registrar and the shift registrar don't
compute the shift between tiles in the same way.
But both functions had issues in the code deciding whether the shift is
too large to be correct.

Typically the overlapping areas are very elongated. So the max distance
in X and Y are very different. The shift vector should be compared
independently in X & Y.
The global shift method had an especially bad bug mixing Y & X, causing
it to estimate the overlap sometimes very small, which caused it to
almost always discard the estimated shift.


src/odemis/acq/stitching/_registrar.py
```diff
@@ -648,10 +651,10 @@ def _get_shift(self, prev_tile, tile):
         # Normalized cross correlation (values between -1 and 1)
         ncc = (covar / (stDev[0] * stDev[1]))
 
-        overlap = min(numpy.abs(numpy.subtract(tile.shape, numpy.abs(exp_shift))))
-        if max(numpy.abs(shift)) > overlap:
-            logging.info("Calculated shift is larger than the overlap size, using expected position "
-                         "instead.")
+        overlap = numpy.abs(numpy.subtract(tile.shape[::-1], numpy.abs(exp_shift)))
```
patrickcleeve2, line 654: I would add a comment here explaining the .shape[::-1] to flip the xy dimenisons

patrickcleeve2 (approved):

tepals (approved):

PR #2581, opened 2023-11-24
[feature] cli convert: add filename expansion to tile names

pieleric:
The --tiles argument can take many filenames. That can be a lot for the
command line. In some cases, it's also not handy to pass the whole list
(ex, when debugging, or running on Windows).

So detect that there is only one argument, and in this case, attempt to
expend it using the standard glob rules.
In the case it was a normal file, just that file is returned. So the
current behaviour is not changed, just "expended".

### DIFF
--- a/src/odemis/cli/convert.py
+++ b/src/odemis/cli/convert.py
@@ -22,6 +22,7 @@ You should have received a copy of the GNU General Public License along with Ode
 # convert --input file-as.hdf5 --output file-as.ome.tiff
 
 import argparse
+import glob
 from gettext import ngettext
 import logging
 import numpy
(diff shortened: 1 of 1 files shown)


src/odemis/cli/convert.py
```diff
@@ -249,6 +250,12 @@ def main(args):
                      len(data), ngettext("image", "images", len(data)),
                      len(thumbs), ngettext("thumbnail", "thumbnails", len(thumbs)))
     elif tifns:
+        if len(tifns) == 1:
+            file_pat = tifns[0]
```
sneepstefan, line 254: Consider changing file_pat to file_path to make it clearer?

tepals (approved):
Small comment, in the title and in the description I think you mean expansion/expand instead of expension/expend.

sneepstefan (approved):

PR #2587, opened 2023-11-29
[fix] fastem remove scan amplitude pre-align from calib 1

tepals:
The scan amplitude calibrations where disabled, because they gave strange values every now and then. We have now monitored the calibrations for over 2 months and it has been consistend over that period. This commit removes the scan amplitude pre-alignment from Calibration set 1 and the user will need to run this calibration every month.

### DIFF
--- a/src/odemis/gui/model/__init__.py
+++ b/src/odemis/gui/model/__init__.py
@@ -1546,7 +1546,6 @@ class FastEMAcquisitionGUIData(MicroscopyGUIData):
                                     Calibrations.IMAGE_ROTATION_PREALIGN,
                                     Calibrations.SCAN_ROTATION_PREALIGN,
                                     Calibrations.DESCAN_GAIN_STATIC,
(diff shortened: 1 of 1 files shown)


nandishjpatel (approved):

pieleric:
Out of curiosity, how will the user run the calibration? Is there a menu or button in the GUI for that? Is there a reminder? Shouldn't it be automatically done by Odemis, based on the last time the calibration was done?

tepals:
> Out of curiosity, how will the user run the calibration? Is there a menu or button in the GUI for that? Is there a reminder? Shouldn't it be automatically done by Odemis, based on the last time the calibration was done?

We're still thinking about how to exactly do this. For now the idea is that this will be a monthly service action, this can either be done by us or by the user. For now we/they will just run it as a script. If we know more about the stability of the calibration, and how many of these monthly actions there are we can decide if we want a button or a reminder or something else.

pieleric (approved):

nandishjpatel (approved):

PR #2588, opened 2023-11-30
[feature] add release-odemis utility

pieleric:
This script is really ad-hoc for releasing Odemis on Ubuntu, but as it's
exactly what we do with Odemis, that's handy.

### DIFF
--- a/util/release-odemis
+++ b/util/release-odemis
@@ -1,24 +1,31 @@
 #!/bin/bash
 # Release a new Odemis release
+#
+# Prerequisites are that you can build the Odemis package and upload it to the PPA.
+# That means your computer has the build tools installed (devscripts), configured
+# a signature, and are allowed to upload packages to the odemis PPA (on launchpad.net)
+# The odemis git repo is clone in ~/development/odemis, and the main github repo is
+# called "upstream".
+#
 # Beforehand, make sure you have already:
 #  1. switched to the right branch (eg, v2.6-branch)
-#  2. updated setup.py with the correct version number
-#  3. updated the changelog with dch -i
+#  2. updated setup.py with the correct version number (eg, 2.6.1)
+#  3. updated the changelog with dch -i (version should be like "2.6.1-1")
 #  4. all is commited
-# Call with "release-odemis"
+# Call with "release-odemis".
 #
 # Note: for a new major release (eg v2.7), it does NOT create a new branch
-# Create it with:
+# Create it (after updating the changelog) with:
 # git checkout -b v2.7-branch $commit-number-to-branch-from$
 #
 # It will:
 #  1. Tag the git repo with the changelog as description
 #  2. Push to the upstream remote
 #  3. Switch to ../pkg-native/odemis/ and update the repo
-#  4. Build for xenial (16.04), and then bionic (18.04).
+#  4. Build for bionic (18.04), focal (20.04) and jammy (22.04).
 
 if [ ! -d ~/development/pkg-native/odemis ]; then
-     echo "Need directory ~/development/pkg-native/odemis with a clone of the odemis repository"
+     echo "Need directory ~/development/pkg-native/odemis with a clone of the odemis repository. Creating one now..."
      # Do so with:
      mkdir -p ~/development/pkg-native/
      cd ~/development/pkg-native/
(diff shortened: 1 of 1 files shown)


util/release-odemis
```diff
@@ -0,0 +1,123 @@
...
+fi
+git archive --prefix=odemis/ -o ../odemis_${RELVER}.orig.tar.gz HEAD || exit 1
+debuild -S || exit 1
+
+# TODO: if the version ends by ".0", push to proposed
+dput ppa:delmic-soft/odemis ../odemis_${PVERSION}_source.changes
+
+# Also rebuild the package to "focal" (Ubuntu 20.04)
+# Just add to debian changelog
+sed -i "1s/^/odemis (${PVERSION}focal1) focal; urgency=medium\n\n  * Rebuild for focal\n\n -- Éric Piel (Delmic) <piel@delmic.com>  $(date -R)\n\n/" debian/changelog
```
tepals, line 105: does it make sense to always do this with your account?

util/release-odemis
```diff
@@ -0,0 +1,123 @@
...
+# Beforehand, make sure you have already:
+#  1. switched to the right branch (eg, v2.6-branch)
+#  2. updated setup.py with the correct version number
+#  3. updated the changelog with dch -i
+#  4. all is commited
+# Call with "release-odemis"
+#
+# Note: for a new major release (eg v2.7), it does NOT create a new branch
+# Create it with:
+# git checkout -b v2.7-branch $commit-number-to-branch-from$
```
tepals, line 12: I do not understand how step 1 and this note relate to each other. Step 1 seems to suggest that you should already be on the right branch, and the note seems to suggest that a new branch will be created.

util/release-odemis
```diff
@@ -0,0 +1,123 @@
...
+#
+# It will:
+#  1. Tag the git repo with the changelog as description
+#  2. Push to the upstream remote
+#  3. Switch to ../pkg-native/odemis/ and update the repo
+#  4. Build for xenial (16.04), and then bionic (18.04).
```
tepals, line 18: is this correct?

tepals (changes requested):
I remember @nandishjpatel  also had some comments on this script, so lets wait until he is back next week before merging.

util/release-odemis
pieleric: Nope. We need a way to find the signature name and email address. It should be possible, because `dch -i` does it. Maybe I should look more into `dch` and see if I can directly call that utility...

util/release-odemis
pieleric: No, the note indicates that it does *NOT* create the branch. I just meant that there is nothing special for the "major" release.

util/release-odemis
pieleric: No. Fixed.

tepals (approved):

sneepstefan (approved):

util/release-odemis
```diff
@@ -0,0 +1,130 @@
...
+fi
+
+cd ~/development/odemis
+
+# Uncomment this line to run line-per-line
+# trap 'read -p "run: $BASH_COMMAND"' DEBUG
+
+# Get the version number from the debian/changelog file
+PVERSION=$(dpkg-parsechangelog -S Version) # should be like "3.1.2-1"
+if ! [[ "$PVERSION" =~ .*-1$ ]]; then
```
nandishjpatel, line 42: So far the odemis changelog says that we always have a first successful build, but what if the first build fails on the PPA and we need to re-build locally with a new build number?

pieleric: This script just doesn't handle a rebuild after a build failure. If this happens, this will have to be done manually.

util/release-odemis
```diff
@@ -0,0 +1,130 @@
...
+
+read -p "Build odemis $RELVER and upload it? [Yn] " answer
+if [[ $answer =~ ^[nN].* ]]; then
+    exit 2
+fi
+git archive --prefix=odemis/ -o ../odemis_${RELVER}.orig.tar.gz HEAD || exit 1
+debuild -S || exit 1
+
+# TODO: if the version ends by ".0", push to proposed
+dput ppa:delmic-soft/odemis ../odemis_${PVERSION}_source.changes
```
nandishjpatel, line 108: Just a comment, no need to address it. In future if we drop `bionic` then the default build will be for `focal` and so on. Maybe its better to get the `Series` from the (changelog)[https://github.com/delmic/odemis/blob/f2c0a35f5e5fd8a62a3d4c45396db9987c691d9e/debian/changelog#L1] and then build for next Ubuntu versions?

nandishjpatel (approved):

util/release-odemis
pieleric: Indeed. For now, I just manually update it. It takes about 5 min every 2 years... so no rush to automate that ;-)

PR #2589, opened 2023-11-30
[feat] added syntax_check_cloud.yml to workflows

sneepstefan:
To automatically check for syntax errors and old style classes with GH instead of only echo these messages in the terminal.

### DIFF
--- a/.github/workflows/syntax_check_cloud.yml
+++ b/.github/workflows/syntax_check_cloud.yml
@@ -1,0 +1,37 @@
+# This workflow will check for exceptions to raise properly and do not generate syntax errors.
+# Also old code style using parenthesis in class definitions will be checked
+
+name: Syntax checking
+
+on:
+  [push, pull_request]
+jobs:
+  build:
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-20.04, ubuntu-22.04]
+
+    steps:
+    - uses: actions/checkout@v4
+    - name: Check exceptions
+      # Not related to tests, but to QA in general: Exceptions usually take only 1 argument
+      # So a comma is probably a sign of syntax error and should be replaced by a %
+      run: |
+        grep -IrE --colour 'raise .+".*%.*",' --include=*.py ./src/odemis ./scripts/ ./plugins/ || exit 0
+        if [[ $? -eq 0 ]]; then
+           echo "Found the above syntax errors in exceptions, this is a sign that a comma should be replaced by a %."        
+           exit 1
+        fi
+
+    - name: Check old style classes
+      if: ${{ !cancelled() }}
+      # Only use parenthesis in class definition if there is inheritance (code-style)
+      run: |        
+        grep -IrE --colour "class .+\(\).*:" --include=*.py ./src/odemis || exit 0
+        if [[ $? -eq 0 ]]; then
+           echo "The above syntax errors were found, only use parenthesis in class definition if there is inheritance (code-style)."
+           exit 1  
+        fi


.github/workflows/syntax_check_cloud.yml
```diff
@@ -0,0 +1,31 @@
...
+
+    steps:
+    - uses: actions/checkout@v4
+    - name: Check exceptions
+      # Not related to tests, but to QA in general: Exceptions usually take only 1 argument
+      # So a comma is probably a sign of syntax error and should be replaced by a %
+      run: |
+        export ODEMIS_DIR=.
+        export ODEMIS_SRC="$ODEMIS_DIR/src/odemis"
+        grep -IrE --colour 'raise .+".*%.*",' --include=*.py "$ODEMIS_SRC" "$ODEMIS_DIR"/scripts/ "$ODEMIS_DIR"/plugins/
```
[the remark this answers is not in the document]
pieleric: grep returns 0 if it finds nothing and 1 if it finds something.
Doesn't it work if it's the last command run?
Otherwise you could just write `exit $?`

pieleric (changes requested):

.github/workflows/syntax_check_cloud.yml
```diff
@@ -0,0 +1,34 @@
...
+      # Not related to tests, but to QA in general: Exceptions usually take only 1 argument
+      # So a comma is probably a sign of syntax error and should be replaced by a %
+      run: |
+        grep -IrE --colour 'raise .+".*%.*",' --include=*.py ./src/odemis ./scripts/ ./plugins/
+        if [[ $? -eq 0 ]]; then
+           exit 1
+        fi
+
+    - name: Check old style classes
+      # Only use parenthesis in class definition if there is inheritance (code-style)
```
tepals, line 29: Can you add the following line between `- name:` and `- run:`? This will ensure this check runs, even when the previous check fails.
```
if: ${{ !cancelled() }}
```

.github/workflows/syntax_check_cloud.yml
```diff
@@ -0,0 +1,34 @@
...
+      run: |        
+        grep -IrE --colour "class .+\(\).*:" --include=*.py ./src/odemis
+        if [[ $? -eq 0 ]]; then
+           exit 1
+        fi
```
tepals, line 34: can you add an empty line at the end?

tepals (changes requested):

.github/workflows/syntax_check_cloud.yml
```diff
@@ -0,0 +1,35 @@
...
+
+    - name: Check old style classes
+      if: ${{ !cancelled() }}
+      # Only use parenthesis in class definition if there is inheritance (code-style)
+      run: |        
+        grep -IrE --colour "class .+\(\).*:" --include=*.py ./src/odemis
+        if [[ $? -eq 0 ]]; then
+           exit 1
```
tepals, line 34: Same here:
```suggestion
           echo "The above syntax errors were found, only use parenthesis in class definition if there is inheritance (code-style)."
           exit 1
```

sneepstefan:
Exceptions are still failing when there are none..

.github/workflows/syntax_check_cloud.yml
```diff
@@ -0,0 +1,37 @@
...
+      matrix:
+        os: [ubuntu-20.04, ubuntu-22.04]
+
+    steps:
+    - uses: actions/checkout@v4
+    - name: Check exceptions
+      # Not related to tests, but to QA in general: Exceptions usually take only 1 argument
+      # So a comma is probably a sign of syntax error and should be replaced by a %
+      run: |
+        grep -IrE --colour 'raise .+".*%.*",' --include=*.py ./src/odemis ./scripts/ ./plugins/
```
pieleric, line 23: > Exceptions are still failing when there are none..

Noooooo!
I guess you can change to:
```suggestion
        grep -IrE --colour 'raise .+".*%.*",' --include=*.py ./src/odemis ./scripts/ ./plugins/ || exit 0
        echo "Found the above syntax errors in exceptions, this is a sign that a comma should be replaced by a %."
        exit 1
```
(Same for the next test)

pieleric (changes requested):

.github/workflows/syntax_check_cloud.yml
```diff
@@ -0,0 +1,41 @@
...
+        if [[ $? -eq 0 ]]; then
+           echo "Found the above syntax errors in exceptions, this is a sign that a comma should be replaced by a %."        
+           exit 1
+        else
+           exit 0
```
pieleric, line 28: No, that will not work, because the script automatically stops when `grep` returns 1. (That's the `-e` flag of bash, which stops the script as soon as a line reports an error. Hence you need the `||` just after the `grep`.

sneepstefan: Alright, cool thanks for you input, I applied you suggestion and it seems to work nicely

tepals:
@sneepstefan I created a PR for the old style classes errors https://github.com/delmic/odemis/pull/2602
Once that is merged, you can rebase and we can merge this PR.

tepals (approved):

sneepstefan:
Thank you @tepals

pieleric (approved):

PR #2593, opened 2023-12-06
[FEAT] [METEOR-856]: Add zeiss metadata reader + test + example image

patrickcleeve2:
This pull request add a parser for converting ZEISS metadata to odemis format. This allows the import of ZEISS SEM/FIB images with the correct metadata for display in the GUI. 

In addition it also adds:
- test case and example image
- libtiff warnings / error suppression


src/odemis/dataio/test/tiff_test.py
```diff
@@ -2387,7 +2387,49 @@ def testConvertTFStoOdemisMD(self):
             else:
                 self.assertEqual(value, data.content[0].metadata[key])
 
+    def testConvertZEISStoOdemisMD(self):
```
pieleric, line 2390: This lower and uppercase mix makes the function hard to read. I'd suggest: `testConvertZeissToOdemisMetadata`

Alternatively you could go all_lower_case, as we have decided for the new functions.

src/odemis/dataio/tiff.py
```diff
@@ -79,12 +79,28 @@
...
+ext = T.add_tags([tag_fei, tag_fei_xml, tag_zeiss_md_1, tag_zeiss_md_2]) # Note: if you ever dereference ext, libtiff will crash
+
+# Hack: they are officially defined as "ASCII" in the file, but they are not
+# null terminated. So read them "raw", clip to the count, and then convert from bytes to the right string format.
+# However, for now pylibtiff doesn't support reading variable length tags via add_tags(),
+# So we have to explicitly change the format to indicate it'll be variable (and GetField will work fine)
+T.tifftags[TIFFTAG_ZEISS_MD_1] = ((T.ctypes.c_uint32, T.ctypes.c_char), lambda d: d[1][:d[0]])
+T.tifftags[TIFFTAG_ZEISS_MD_2] = ((T.ctypes.c_uint32, T.ctypes.c_char), lambda d: d[1][:d[0]])
+
+# suppress warnings and errors from libtiff
```
pieleric, line 100: Please add why: because they are written to stderr (by default).

src/odemis/dataio/tiff.py
```diff
@@ -237,6 +254,19 @@ def _readTiffTag(tfile):
...
+    # zmd2 = tfile.GetField(TIFFTAG_ZEISS_MD_2) # NOTE: unused
+
+    if zmd is not None:
+        try:
+            zmd = zmd.decode("latin1")
+            md.update(_convertZEISStoOdemisMD(zmd))
+        except Exception as e:
+            logging.info(f"Failed to parse Zeiss metadata: {e}")
+
+
```
pieleric, line 268: Only one empty line is enough.

src/odemis/dataio/tiff.py
```diff
@@ -237,6 +254,19 @@ def _readTiffTag(tfile):
         except Exception as e:
             logging.info(f"Failed to parse ThermoFisher metadata: {e}")
 
+    # parse zeiss metadata
+    zmd = tfile.GetField(TIFFTAG_ZEISS_MD_1)
+    # zmd2 = tfile.GetField(TIFFTAG_ZEISS_MD_2) # NOTE: unused
```
pieleric, line 259: Please add why it's unused: it's exactly the same content, but UTF16 encoded.

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2752,210 @@ def _convertTFStoOdemisMD(metadata: str) -> dict:
         logging.warning(f"Failed to parse ThermoFisher metadata: {model.MD_DET_TYPE} - {e}")
 
     return md
+
+def _parseZEISSMetadata(metadata) -> dict:
```
pieleric, line 2756: Let's drop the "ZEISS" in all capital (in function names), it's really annoying to read.

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2752,210 @@ def _convertTFStoOdemisMD(metadata: str) -> dict:
...
+        # Stage at X = 48.5090 mm
+
+        # TODO: try to do it similar to tifffile to extract header information too            
+        try:
+            # NOTE: 
+            # Date, Time are separated by : 
+            # everything else is separated by =
+            try:
+                k, v = line.split("=")
+            except:
```
pieleric, line 2780: Can you be specific in terms of exception?

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2752,210 @@ def _convertTFStoOdemisMD(metadata: str) -> dict:
...
+            # Date, Time are separated by : 
+            # everything else is separated by =
+            try:
+                k, v = line.split("=")
+            except:           
+                k, v = line.split(":", maxsplit=1) 
+            k, v = k.strip(), v.strip()
+            zmd[k] = v
+        except Exception as e:
+            continue
```
pieleric, line 2785: Please, at least log it, with logging.warning(... exc_info=True).
Or is it most probably a header? In that case maybe you could explicitly detect such "one word", and skip it.

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2752,210 @@ def _convertTFStoOdemisMD(metadata: str) -> dict:
...
+        else:
+            raise ValueError(f"Unknown unit: {unit}")
+
+        md[model.MD_EBEAM_VOLTAGE] = hv
+    except Exception as e:
+        logging.warning(f"Failed to parse ZEISS metadata: {model.MD_EBEAM_VOLTAGE} - {e}")
+
+    # spot size
+    try:
+        # NB: this is not actual spot size, actual spot diam is in xml metadata
```
pieleric, line 2947: I don't think this comment is correct (that's for TFS).

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2752,210 @@ def _convertTFStoOdemisMD(metadata: str) -> dict:
...
+
+        md[model.MD_DWELL_TIME] = dwell_time
+    except Exception as e:
+        logging.warning(f"Failed to parse ZEISS metadata: {model.MD_DWELL_TIME} - {e}")
+    try:
+        unit = zmd["Stage at X"].split(" ")[-1]
+        x = float(zmd["Stage at X"].split(" ")[0])
+        y = float(zmd["Stage at Y"].split(" ")[0])
+
+        # TODO: investigate using src/odemis/util/units.py
```
pieleric, line 2855: Could you investigate now? I think it would be just a matter of something like that:
```
str_val, si_prefix, unit = decompose_si_prefix(str_x, "m")
x = si_scale_val(float(str_val), si_prefix)
```
This would make sure that if x and y are different units, it still reads the correct values.

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2752,210 @@ def _convertTFStoOdemisMD(metadata: str) -> dict:
...
+    :return: (dict) metadata as dictionary
+    :raises: (KeyError): if the metadata cannot be parsed
+    """
+
+    # parse as dict
+    zmd = _parseZEISSMetadata(metadata)
+
+    md = {}
+
+    # convert to odemis format
```
pieleric, line 2802: I think you can drop this comment. This is essentially what the whole function does!

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2752,210 @@ def _convertTFStoOdemisMD(metadata: str) -> dict:
...
+            continue
+
+    return zmd
+
+
+def _convertZEISStoOdemisMD(metadata: str) -> dict:
+    """Converts ZEISS metadata to odemis metadata format
+    :param metadata: metadata string
+    :return: (dict) metadata as dictionary
+    :raises: (KeyError): if the metadata cannot be parsed
```
pieleric, line 2794: Is that correct? I don't see any KeyError raised

src/odemis/dataio/tiff.py
pieleric, line 2756: Can you add the type of `metadata`?

pieleric (changes requested):

src/odemis/dataio/tiff.py
patrickcleeve2: - added a check for headers (.isupper() as all headers are single word all caps)
- differentiated ValueErrors from other errors. if both "=", and ":" delimeters dont return kv pairs, the line is a single unused metadata value, and can be ignored. 
- Other errors: something else has gone wrong -> log

src/odemis/dataio/tiff.py
```diff
@@ -79,12 +79,28 @@
...
+
+# Hack: they are officially defined as "ASCII" in the file, but they are not
+# null terminated. So read them "raw", clip to the count, and then convert from bytes to the right string format.
+# However, for now pylibtiff doesn't support reading variable length tags via add_tags(),
+# So we have to explicitly change the format to indicate it'll be variable (and GetField will work fine)
+T.tifftags[TIFFTAG_ZEISS_MD_1] = ((T.ctypes.c_uint32, T.ctypes.c_char), lambda d: d[1][:d[0]])
+T.tifftags[TIFFTAG_ZEISS_MD_2] = ((T.ctypes.c_uint32, T.ctypes.c_char), lambda d: d[1][:d[0]])
+
+# suppress warnings and errors from libtiff (written to stderr by default)
+T.suppress_errors()
```
pieleric, line 101: Let's keep the error active for now (you can just leave the line, but comment it out). So far, the annoying ones were the warnings. It can still be handy to see an error if there is one.

src/odemis/dataio/tiff.py
```diff
@@ -237,6 +254,17 @@ def _readTiffTag(tfile):
         except Exception as e:
             logging.info(f"Failed to parse ThermoFisher metadata: {e}")
 
+    # parse zeiss metadata
+    zmd = tfile.GetField(TIFFTAG_ZEISS_MD_1)
```
pieleric, line 258: Could you add a `ignore_undefined_tag=True` to the call, so that if it's not found (which is very common), it will not generate a warning. (Granted, now the warnings are not shown, but if we ever show them again, it's nice to know that this tag missing is really nothing to worry about). You could also do the same for the TFS tag.

pieleric (changes requested):
Looks good. I have just very tiny comments. Also, I guess you'll have to rebase against master as we've already merged the renaming FEI -> TFS.

pieleric (approved):

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2750,188 @@ def _convert_thermo_fisher_to_odemis_metadata(metadata: str) -> dict:
...
+
+
+def _convert_zeiss_to_odemis_metadata(metadata: str) -> dict:
+    """Converts ZEISS metadata to odemis metadata format
+    :param metadata: metadata string
+    :return: (dict) metadata as dictionary
+    """
+
+    # parse as dict
+    # from pprint import pprint
```
tepals, line 2801: can this be removed?

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2750,188 @@ def _convert_thermo_fisher_to_odemis_metadata(metadata: str) -> dict:
...
+        md[model.MD_ACQ_DATE] = date.timestamp()
+    except Exception as e:
+        logging.warning(f"Failed to parse ZEISS metadata: {model.MD_ACQ_DATE} - {e}")
+    try:
+        md[model.MD_HW_NAME] = zmd["Serial No."]
+    except Exception as e:
+        logging.warning(f"Failed to parse ZEISS metadata: {model.MD_HW_NAME} -  {e}")
+    try:
+        md[model.MD_SW_VERSION] = zmd["Version"]
+    except Exception as e:
```
tepals, line 2819: Why do you use generic exceptions here and not for instance a `KeyError`?

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2750,188 @@ def _convert_thermo_fisher_to_odemis_metadata(metadata: str) -> dict:
...
+        str_dwell_time = zmd["Dwell Time"]
+        str_val, si_prefix, unit = units.decompose_si_prefix(str_dwell_time, "s")
+        dwell_time = units.si_scale_val(float(str_val), si_prefix)
+
+        md[model.MD_DWELL_TIME] = dwell_time
+    except Exception as e:
+        logging.warning(f"Failed to parse ZEISS metadata: {model.MD_DWELL_TIME} - {e}")
+    try:
+
+        str_x = zmd["Stage at X"]
```
tepals, line 2846: can you move the empty line to above the `try:`?
```suggestion
        logging.warning(f"Failed to parse ZEISS metadata: {model.MD_DWELL_TIME} - {e}")

    try:
        str_x = zmd["Stage at X"]
```

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2750,188 @@ def _convert_thermo_fisher_to_odemis_metadata(metadata: str) -> dict:
...
+    # detector
+    try:
+        md[model.MD_DET_TYPE] = (model.MD_DT_NORMAL 
+                                    if zmd["Signal A"] in ["InLens", "SE2"] 
+                                    else model.MD_DT_INTEGRATING)
+    except Exception as e:
+        logging.warning(f"Failed to parse ZEISS metadata: {model.MD_DET_TYPE} - {e}")
+
+
+    return md
```
tepals, line 2937: can you add an empty line at the bottom of the file?

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2750,188 @@ def _convert_thermo_fisher_to_odemis_metadata(metadata: str) -> dict:
...
+            "SEM": "EHT",
+            "FIB": "FIB EHT",
+        } 
+        hv_key = hv_map[beam_type]
+
+        str_hv = zmd[hv_key]
+        str_val, si_prefix, unit = units.decompose_si_prefix(str_hv, "V")
+        hv = units.si_scale_val(float(str_val), si_prefix)
+        
+
```
tepals, line 2917: can you remove one empty line?

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2750,188 @@ def _convert_thermo_fisher_to_odemis_metadata(metadata: str) -> dict:
...
+    # detector
+    try:
+        md[model.MD_DET_TYPE] = (model.MD_DT_NORMAL 
+                                    if zmd["Signal A"] in ["InLens", "SE2"] 
+                                    else model.MD_DT_INTEGRATING)
+    except Exception as e:
+        logging.warning(f"Failed to parse ZEISS metadata: {model.MD_DET_TYPE} - {e}")
+
+
```
tepals, line 2936: can you remove one empty line?

tepals (changes requested):
sorry for the late review, I already had a look last week and forgot to finish it.

src/odemis/dataio/tiff.py
patrickcleeve2: I've change the above two to KeyError but think the rest should stay as generic exceptions because there are multiple ways the parser can fail, e.g. key error, formatting, additional chars.

patrickcleeve2: I've changed the two above to KeyError, but think the rest should stay as generic exceptions because there are multiple things that can fail for those keys.

pieleric (approved):

tepals (approved):

patrickcleeve2:
@pieleric @tepals I think this can be merged now

PR #2595, opened 2023-12-06
[fix] util.dataio: stitching HDF5 data failed because DataArrays always have 5 dimensions

pieleric:
When reading from HDF5, the DataArray always have 5 dimensions. If it's
a 2D image (XY), then the first 3 dimensions have length 1. That's
fine... except that this confuses the stitching functions which for now
only handle data with dimensions YX.

=> Remove the extra dimensions before passing the DataArray to the
stitching algorithm

### DIFF
--- a/src/odemis/util/dataio.py
+++ b/src/odemis/util/dataio.py
@@ -32,7 +32,8 @@ from odemis.acq.stitching import REGISTER_IDENTITY, REGISTER_GLOBAL_SHIFT, REGIS
                               WEAVER_COLLAGE, WEAVER_COLLAGE_REVERSE, WEAVER_MEAN
 from odemis.acq.stream import StaticARStream, StaticCLStream, StaticFluoStream, StaticSEMStream, \
                               StaticSpectrumStream
-                               
+from odemis.util import img
+
 
 def data_to_static_streams(data):
     """ Split the given data into static streams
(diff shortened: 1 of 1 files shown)


src/odemis/util/dataio.py
```diff
@@ -309,6 +310,14 @@ def leader_quality(da):
...
         # Make sure they are not DataArrayShadow, because we need the data to do the stitching
         das = [da.getData() if isinstance(da, model.DataArrayShadow) else da for da in das]
+        # For now the stitching doesn't handle more than 2 dimensions, so force the data to be 2D
+        das_2d = []
+        for da in das:
+            try:
+                das_2d.append(img.ensure2DImage(da))
+            except ValueError:
+                logging.info("Skipping %s because it is not 2D", da)
+        das = das_2d
```
tepals, line 320: So if there are no 2D data arrays this will just be an empty list, is that the expected behavior?

pieleric: Good remark... Actually there is code above that does already some filtering on the das, so this could have already happened. For now in such case, it fails with IndexError. We could catch this case and make it a little bit more understandable.

Honestly this function is not super robust. We probably should also check that all the files have the same amount of streams, as otherwise it's also going to go bad...

tepals (approved):

patrickcleeve2 (approved):

PR #2596, opened 2023-12-08
[FEAT] [METEOR-857] Add parser for TESCAN images

patrickcleeve2:
This pull request add a parser for converting TESCAN metadata to odemis format. This allows the import of TESCAN SEM/FIB images with the correct metadata for display in the GUI.

In addition it also adds:
- test case and example image


src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2744,168 @@ def _convertTFStoOdemisMD(metadata: str) -> dict:
...
+    try:
+        # NOTE: due to metadata corruption, we can't directly read the beam type
+        # we check for the columntilt key which is only present in FIB metadata
+        # Once we can read the metadata correctly, we can remove this
+        # and directly read the beam type        
+        # acq_map =  {
+        #     "SEM": model.MD_AT_EM,
+        #     "FIB": model.MD_AT_FIB,
+        # }
+        # md[model.MD_ACQ_TYPE] = acq_map[beam_type]
```
pieleric, line 2872: Is that correct? I couldn't find this info in the example metadata.

pieleric (changes requested):

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2744,168 @@ def _convert_thermo_fisher_to_odemis_metadata(metadata: str) -> dict:
...
+            k, v = (line.split("="))[-2:]
+
+            # NOTE: the first keys for each beam types metadata 
+            # have corrupt binary chars so we need to remove 
+            # these manually
+            for sk in BEAM_MD_START_KEYS:
+                if sk in k:
+                    k = k[k.find(sk):]
+
+        except Exception as e:
```
tepals, line 2783: shouldn't it be possible to use a less generic exception here? It looks like it should always be a `KeyError`?

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2744,168 @@ def _convert_thermo_fisher_to_odemis_metadata(metadata: str) -> dict:
...
+            logging.warning(f"Error: {line} - {e}")
```
tepals, line 2784:
```suggestion
            logging.warning(f"Error parsing metadata line: {line} - {e}")    
```

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2744,168 @@ def _convert_thermo_fisher_to_odemis_metadata(metadata: str) -> dict:
...
+    # 1. Change to configparser to parse the metadata correctly
+    # 2. Correctly read the beam type from section headers
+    # 3. Remove the beam type check below (acq_type)
+
+    # parse ini metadata as dictionary 
+    # NOTE: not correctly ini formatted due to corruption
+    try:
+       tmd = _parse_tescan_metadata(metadata)
+    except Exception as e:
+        raise Exception(f"Failed to parse TESCAN metadata: {e}")
```
tepals, line 2815: It would be nice to not raise a generic exception here, so we already have a bit more information on what's going wrong.

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2744,168 @@ def _convert_thermo_fisher_to_odemis_metadata(metadata: str) -> dict:
...
+    try:
+        md[model.MD_DWELL_TIME] = float(tmd["dwelltime"])
+    except Exception as e:
+        logging.warning(f"Failed to parse TESCAN metadata: {model.MD_DWELL_TIME} - {e}")
+    try:
+        # MD_POS should match image dimensions, e.g. only XY for 2D, XYZ for 3D, etc.
+        md[model.MD_POS] = (float(tmd["stagex"]), 
+                           float(tmd["stagey"]),
+                           )
+    except Exception as e:
```
tepals, line 2859: I think in all the exceptions from line 2827 to here you can use a KeyError instead of a generic Exception: `except KeyError as e:`

tepals: same from line 2878 to line 2909

tepals (changes requested):

src/odemis/dataio/tiff.py
patrickcleeve2: Yeah, the columntilt key is only in FIB images, so it doesnt appear in SEM images. The example image is SEM

src/odemis/dataio/tiff.py
patrickcleeve2: changed to value error

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2744,168 @@ def _convertTFStoOdemisMD(metadata: str) -> dict:
...
+## TESCAN
+
+def _parse_tescan_metadata(metadata: str) -> dict:
+    """
+    Parses TESCAN metadata (ini) to dictionary
+    :param metadata: metadata string as ini (TESCAN format)
+    :return: (dict) metadata as dictionary (TESCAN format)
+    """
+    # NOTE:
+    # The metadata should be formatted as ini, but seems to be corrupted
```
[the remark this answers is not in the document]
patrickcleeve2: yeah, you are correct good call. i've changed the parser to account for this (skip jpeg2000 metadata, then read chunks + headers  sequentially). It now can be read directly with configparser so that also removes the manual parsing to dict.

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2744,168 @@ def _convertTFStoOdemisMD(metadata: str) -> dict:
...
+    # and therefore not parsed correctly by configparser
+    # Therefore we need to parse the metadata manually
+    # The metadata should contain two sections, MAIN and SEM/FIB.
+    # The SEM/FIB section header allows us to determine the beam type,
+    # but it is corrupt and not parsed correctly by configparser.
+
+    # Therefore we need to parse the metadata manually, 
+    # the valid part starts with AccFrames= and ends with a newline
+    valid_md = metadata[metadata.find("AccFrames="):].split("\n")[:-1]
```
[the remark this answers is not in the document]
patrickcleeve2: see above comment

src/odemis/dataio/test/tescan_example_sem.tif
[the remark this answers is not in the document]
patrickcleeve2: I've changed the image to the smaller image, but I think it would be a good idea to also test on a fib image as the parser has gotten pretty complicated. But given the size of the images, might be better to do that on the self hosted runner, rather than include in the repo?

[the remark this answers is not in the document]
pieleric: Yes, I would really prefer keeping the size of this git repo small. What I have been thinking is to agree on a "official" path on the computer to the "Test data" directory on Google Drive. Then, we could add test cases that use files from this directory. If that directory is not available (eg, on the CI server), the test would just be skipped.

tepals (approved):

src/odemis/dataio/test/tescan_example_sem.tif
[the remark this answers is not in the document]
patrickcleeve2: yep, that sounds good. I will put together the images I have for the different manufacturers

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2742,188 @@ def _convert_thermo_fisher_to_odemis_metadata(metadata: str) -> dict:
...
+        chunk_start = _LENGTH_BYTES+_HEADER_BYTES
+        chunk_end = chunk_start + next_chunk_len - _END_BYTES
+        next_chunk = metadata[chunk_start:chunk_end].decode("latin1")
+
+        # read the final 3 bytes (unused)
+        # NOTE: 3 bytes at the end of each chunk, not sure what for?
+        # standard values were 768, 512, 1208, so maybe something to do with display?        
+        next_chunk_final = metadata[chunk_end:chunk_end+_END_BYTES]
+        
+        # add the chunk data to the metadata string
```
pieleric, line 2811: Skip if `next_chunk_header` not in `header_map` (to handle other metadata gracefully)

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2742,188 @@ def _convert_thermo_fisher_to_odemis_metadata(metadata: str) -> dict:
...
+        vmd += _header_map[next_chunk_header] + next_chunk
```
pieleric, line 2812: Why isn't there a `"\n"` between the header_map and the actual chunk data?

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2742,188 @@ def _convert_thermo_fisher_to_odemis_metadata(metadata: str) -> dict:
...
+        2: "[MAIN]",
+        3: "[SEM]",
+        5: "[FIB]"
+    }
+
+    # skip jpeg2000 metadata
+    next_chunk_len = int.from_bytes(metadata[:4], byteorder="little", signed=False)
+    
+    # loop through the two chunks, extract the valid data
+    for i in range(2):
```
pieleric, line 2789: Note: this works fine for our needs. However, if you want to be more generic by just continuously scanning until either the `len(metadata)` < 6, or if encountering the type 0. Skip the chunks if the type is not in the header_map.

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2742,188 @@ def _convert_thermo_fisher_to_odemis_metadata(metadata: str) -> dict:
...
+    # general
+    try:
+        datetime_str = tmd["MAIN"]["date"] + " " + tmd["MAIN"]["time"]
+        date = datetime.strptime(datetime_str, "%Y-%m-%d %H:%M:%S")
+        md[model.MD_ACQ_DATE] = date.timestamp()
+    except Exception as e:
+        logging.warning(f"Failed to parse TESCAN metadata: {model.MD_ACQ_DATE} - {type(e)}: {e}")
+
+    try:
+        md[model.MD_HW_NAME] = tmd["MAIN"]["device"] + "-" + tmd["MAIN"]["serialnumber"]
```
pieleric, line 2846: I don't get how "serialnumber" matches the "SerialNumber" key. Is the ConfigParser case insensitive? Or is that an error (due to previously using `lower()`)?

pieleric (changes requested):

src/odemis/dataio/tiff.py
patrickcleeve2: ConfigParser converts them to case-insensitive by default
https://stackoverflow.com/questions/19359556/configparser-reads-capital-keys-and-make-them-lower-case

Can turn the behaviour off with the following if required 
``` python
self.config = ConfigParser.ConfigParser()
self.config.optionxform = str
```

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2742,192 @@ def _convert_thermo_fisher_to_odemis_metadata(metadata: str) -> dict:
...
+            break # end of metadata
+
+        # read the chunk data as string (next n-3 bytes)
+        chunk_start = LENGTH_BYTES+HEADER_BYTES
+        chunk_end = chunk_start + chunk_len - END_BYTES
+        chunk = metadata[chunk_start:chunk_end].decode("latin1")
+
+        # read the final 3 bytes (unused)
+        # NOTE: 3 bytes at the end of each chunk, not sure what for?
+        # standard values were 768, 512, 1208, so maybe something to do with display?
```
pieleric, line 2811: I don't really see which values you're referring to. Also each bytes should be < 256, so what do you mean by "standard values"? All I can see is `0d 0a 00`, which is the standard line return (\r\n) followed by the null byte (which you could remove if you detect it's a string, a supported header)

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2742,192 @@ def _convert_thermo_fisher_to_odemis_metadata(metadata: str) -> dict:
...
+    while len(metadata) > LENGTH_BYTES + HEADER_BYTES:
+        # skip to next chunk
+        metadata = metadata[LENGTH_BYTES+chunk_len:] # skip to next chunk
+        
+        # read the length of the next chunk (first 4 bytes)
+        chunk_len = int.from_bytes(metadata[:LENGTH_BYTES], 
+                                            byteorder="little", signed=False)
+        
+        # read the type of the chunk (next 2 bytes)
+        chunk_type = int.from_bytes(metadata[LENGTH_BYTES:LENGTH_BYTES+HEADER_BYTES],
```
pieleric, line 2798: Please put spaces around `+`. (that's the code style)

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2742,192 @@ def _convert_thermo_fisher_to_odemis_metadata(metadata: str) -> dict:
...
+
+    # header type map
+    header_map ={
+        2: "[MAIN]",
+        3: "[SEM]",
+        5: "[FIB]"
+    }
+
+    # skip jpeg2000 metadata
+    chunk_len = int.from_bytes(metadata[:LENGTH_BYTES], byteorder="little", signed=False)
```
pieleric, line 2786: You can drop that now. Just need to invert the reading of chunk_len and metadata in the `while` block.

pieleric (changes requested):

src/odemis/dataio/tiff.py
patrickcleeve2: I think I must have done something wrong with those prev values, because I'm not getting it anymore. Im not getting those at the end, (chunk_final) getting e.g. b'\x00\x99\x03' or b'\x00\x02\x00' instead.

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2742,192 @@ def _convert_thermo_fisher_to_odemis_metadata(metadata: str) -> dict:
...
+    # METADATA FORMAT
+    # JPEG2000 metadata
+    #   4 bytes: length of the next chunk (n)
+    #   2 bytes: header of the next chunk (type)
+    #   n bytes: chunk data
+    # MAIN metadata
+    #   4 bytes: length of the next chunk (n)
+    #   2 bytes: header of the next chunk (type)
+    #   n - 3 bytes: valid chunk data
+    #   3 bytes: unsure?
```
pieleric, line 2763: Drop. It's just n bytes.

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2742,192 @@ def _convert_thermo_fisher_to_odemis_metadata(metadata: str) -> dict:
...
+    # SEM/FIB metadata
+    #   4 bytes: length of the next chunk (n)
+    #   2 bytes: header of the next chunk (type)
+    #   n - 3 bytes: valid chunk data
+    #   3 bytes: unsure?
```
pieleric, line 2768: Drop

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2742,192 @@ def _convert_thermo_fisher_to_odemis_metadata(metadata: str) -> dict:
...
+
+    # constants
+    LENGTH_BYTES = 4
+    HEADER_BYTES = 2
+    END_BYTES = 3
```
pieleric, line 2773: Drop.

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2742,192 @@ def _convert_thermo_fisher_to_odemis_metadata(metadata: str) -> dict:
...
+        # NOTE: 3 bytes at the end of each chunk, not sure what for?     
+        chunk_final = metadata[chunk_end:chunk_end + END_BYTES]
+
+        # add the chunk data to the metadata string (if valid)
+        if chunk_type in header_map:
+            vmd += f"{header_map[chunk_type]}\n" + chunk
+            print(chunk_type, chunk_final, metadata[chunk_end - 10:chunk_start + chunk_len ])
+        
+        # skip to next chunk
+        metadata = metadata[LENGTH_BYTES + chunk_len:] # skip to next chunk
```
pieleric, line 2813: I think there is a couple of errors with the indices. This code works fine, and there is not "end bytes" trick needed:
```suggestion
        # read the chunk data as string
        chunk_start = LENGTH_BYTES + HEADER_BYTES
        chunk_end = LENGTH_BYTES + chunk_len
        chunk = metadata[chunk_start:chunk_end]

        # add the chunk data to the metadata string (if valid), not including the null byte at the end of the string
        if chunk_type in header_map:
            vmd += header_map[chunk_type] + "\n" + chunk[:-1].decode("latin1")

        # skip to next chunk
        metadata = metadata[chunk_end:]  # skip to next chunk
```

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2742,192 @@ def _convert_thermo_fisher_to_odemis_metadata(metadata: str) -> dict:
...
+        assert beam_type in tmd.sections()
+            
+        try:
+            md[model.MD_DWELL_TIME] = float(tmd[beam_type]["dwelltime"])
+        except KeyError as e:
+            logging.warning(f"Failed to parse TESCAN metadata: {model.MD_DWELL_TIME} - {e}")
+        try:
+            # MD_POS should match image dimensions, e.g. only XY for 2D, XYZ for 3D, etc.
+            md[model.MD_POS] = (float(tmd[beam_type]["stagex"]), 
+                            float(tmd[beam_type]["stagey"]),
```
pieleric, line 2885: Please adjust the indentation. Typically you want the two elements (`float`...) to be aligned.

pieleric (changes requested):

src/odemis/dataio/tiff.py
```diff
@@ -2722,3 +2742,192 @@ def _convert_thermo_fisher_to_odemis_metadata(metadata: str) -> dict:
...
+        chunk = metadata[chunk_start:chunk_end].decode("latin1")
+
+        # read the final 3 bytes (unused)
+        # NOTE: 3 bytes at the end of each chunk, not sure what for?     
+        chunk_final = metadata[chunk_end:chunk_end + END_BYTES]
+
+        # add the chunk data to the metadata string (if valid)
+        if chunk_type in header_map:
+            vmd += f"{header_map[chunk_type]}\n" + chunk
+            print(chunk_type, chunk_final, metadata[chunk_end - 10:chunk_start + chunk_len ])
```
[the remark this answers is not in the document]
patrickcleeve2: Yep correct, sorry about that

src/odemis/dataio/tiff.py
patrickcleeve2: much cleaner, thanks 👍

pieleric (approved):

PR #2600, opened 2023-12-12
[refactor] Add comment to simulated yaml files, explaining that these…

tepals:
… can only be used for the backend.


pieleric (approved):

sneepstefan (approved):

sneepstefan:
I think making it clear this way with a comment is a good solution for now.

PR #2601, opened 2023-12-12
[fix] [METEOR-848] update the power slider range when excitation is changes

K4rishma:
power slider updates its range even when the stream is paused and shows the correct power range for the selected excitation.


pieleric (approved):

pieleric:
Note that this fixes the bug also on SECOM and DELPHI.

tepals (approved):

PR #2604, opened 2023-12-14
[METEOR-810] Manual overlay of FLM and SEM images (Correlation Tab)  

patrickcleeve2:
This PR adds:

- Correlation Tab
- Manual Overlay controls for correlation between images
- Export controller for exporting overlaid images.

### DIFF
--- a/src/odemis/gui/cont/correlation.py
+++ b/src/odemis/gui/cont/correlation.py
@@ -1,0 +1,298 @@
+
+# -*- coding: utf-8 -*-
+
+"""
+@author Patrick Cleeve
+
+Copyright © 2023, Delmic
+
+Handles the controls for correlating two (or more) streams together.
+
+This file is part of Odemis.
+
+Odemis is free software: you can redistribute it and/or modify it under the
+terms of the GNU General Public License version 2 as published by the Free
+Software Foundation.
+
+Odemis is distributed in the hope that it will be useful, but WITHOUT ANY
+WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+PARTICULAR PURPOSE. See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with
+Odemis. If not, see http://www.gnu.org/licenses/.
+
+"""
+import logging
+import math
+import wx
+# IMPORTANT: wx.html needs to be imported for the HTMLWindow defined in the XRC
+# file to be correctly identified. See: http://trac.wxwidgets.org/ticket/3626
+# This is not related to any particular wxPython version and is most likely permanent.
+import wx.html
+import odemis.acq.stream as acqstream
+import odemis.gui.model as guimod
+from odemis import model
+from odemis.acq.stream import StaticStream
+from odemis.gui.util import call_in_wx_main
+
+# TODO: move to more approprate location
+def update_image_in_views(s: StaticStream, views: list) -> None:
+    """Force update the static stream in the selected views
+    :param s: (StaticStream) the static stream to update
+    :param views: (list[View]) the list of views to update"""
+    v: guimod.View 
+    for v in views:
+        for sp in v.stream_tree.getProjections():  # stream or projection
+            if isinstance(sp, acqstream.DataProjection):
+                st = sp.stream
+            else:
+                st = sp
+            if st is s:
+                sp._shouldUpdateImage()
+
+
+class CorrelationController(object):
+
+    def __init__(self, tab_data, panel, tab, viewports):
+        """
+        :param tab_data: (MicroscopyGUIData) the representation of the microscope GUI
+        :param panel: (wx._windows.Panel) the panel containing the UI controls
+        :param tab: (Tab) the tab which should show the data
+        :param viewports: (ViewPorts) the tab view ports
+        """
+        
+        self._tab_data_model = tab_data
+        self._main_data_model = tab_data.main
+        self._panel = panel
+        self._tab = tab
+        self._viewports = viewports
+
+        # disable if no streams are present
+        self._tab_data_model.streams.subscribe(self._on_correlation_streams_change, init=True)
+        self._tab_data_model.streams.subscribe(self._update_correlation_cmb, init=True)
+
+        # connect the correlation streams to the tab data
+        self._panel.cmb_correlation_stream.Bind(wx.EVT_COMBOBOX, self._on_selected_stream_change)
+
+        # reset correlation data
+        self._panel.btn_reset_correlation.Bind(wx.EVT_BUTTON, self.reset_correlation_pressed)
+
+        # enable correlation controls
+        self._panel.ctrl_enable_correlation.SetValue(True) # enable by default
+
+        # bind mouse and keyboard events for correlation controls
+        for vp in self._viewports:
+            vp.canvas.Bind(wx.EVT_CHAR, self.on_char)
+            vp.canvas.Bind(wx.EVT_LEFT_DOWN, self.on_mouse_down)
+    
+    @call_in_wx_main
+    def _update_correlation_cmb(self, streams: list) -> None:
+        """update the correlation combo box with the available streams
+        :param streams: (list[StaticStream]) the streams to add to the combo box"""
+        # keep the combobox in sync with streams
+        self._panel.cmb_correlation_stream.Clear()
+        for s in streams:
+            self._panel.cmb_correlation_stream.Append(s.name.value, s)
+        
+        # select the first stream, if available
+        if len(streams) > 0:
+            self._panel.cmb_correlation_stream.SetSelection(0)
+    
+    @call_in_wx_main
+    def _on_correlation_streams_change(self, streams: list) -> None:
+        """hide/show the correlation panel if there are no streams
+        :param streams: (list[StaticStream]) the streams in the correlation tab"""
+        visible = len(streams) != 0
+        self._panel.fp_meteor_correlation.Show(visible)
+
+    @call_in_wx_main
+    def _on_selected_stream_change(self, evt: wx.Event) -> None:
+        """change the selected stream to the one selected in the combo box
+        :param evt: (wx.Event) the event"""
+        idx = self._panel.cmb_correlation_stream.GetSelection()
+        self._tab_data_model.selected_stream.value = self._panel.cmb_correlation_stream.GetClientData(idx)
+        logging.debug(f"Selected Stream Changed to {idx}: {self._tab_data_model.selected_stream.value.name.value}")
+    
+    def reset_correlation_pressed(self, evt: wx.Event) -> None:
+        """"Reset the correlation data for the selected stream, and re-draw
+        :param evt: (wx.Event) the event"""
+        s = self._tab_data_model.selected_stream.value
+        self._reset_stream_correlation_data(s)
+        update_image_in_views(s, self._tab_data_model.views.value)
+
+    def _reset_stream_correlation_data(self, s: StaticStream) -> None:
+        """reset the stream position to the original position / rotation / scale
+        :param s: (StaticStream) the stream to reset"""
+        s.raw[0].metadata[model.MD_POS_COR] = (0, 0)
+        s.raw[0].metadata[model.MD_ROTATION_COR] = 0
+        s.raw[0].metadata[model.MD_PIXEL_SIZE_COR] = (1, 1)       
+
+    def add_streams(self, streams: list) -> None:
+        """add streams to the correlation tab
+        :param streams: (list[StaticStream]) the streams to add"""
+
+        # NOTE: we only add streams if they are not already in the correlation tab
+        # we will not remove streams from the correlation tab from outside it, 
+        # as the user may still want to correlate them, 
+        # even if they have been removed from another tab, e.g. 'localization'
+
+        # add streams to correlation tab
+        logging.debug(f"Adding {len(streams)} streams to correlation tab")
+        for s in streams:
+            
+            # skip existing streams
+            if s in self._tab_data_model.streams.value:
+                continue 
+
+            # reset the stream correlation data, and add to correlation streams
+            self._reset_stream_correlation_data(s)
+            self._tab_data_model.streams.value.append(s)          
+            
+            # add stream to streambar
+            sc = self._tab.streambar_controller.addStream(s, add_to_view=True, play=False)
+            sc.stream_panel.show_remove_btn(True)  
+
+    def clear_streams(self) -> None:
+        """clears streams from the correlation tab"""
+        self._tab.streambar_controller.clear()
+
+    def correlation_enabled(self) -> bool:
+        """return if correlation controls are enabled and a stream is selected"""
+        logging.debug(f"correlation enabled: {self._panel.ctrl_enable_correlation.IsChecked()}")
+        logging.debug(f"selected stream: {self._tab_data_model.selected_stream.value}")
+        return (self._panel.ctrl_enable_correlation.IsChecked() and
+                self._tab_data_model.selected_stream.value is not None)
+
+    def on_mouse_down(self, evt: wx.Event) -> None:
+        """handle mouse down events
+        :param evt: (wx.Event) the event"""
+        active_canvas = evt.GetEventObject()
+        logging.debug(f"mouse down event, canvas: {active_canvas}")
+ 
+        # check if shift is pressed, and if a stream is selected
+        if evt.ShiftDown() and self.correlation_enabled():
+                                    
+            # get the position of the mouse, convert to physical position
+            pos = evt.GetPosition()
+            p_pos = active_canvas.view_to_phys(pos, active_canvas.get_half_buffer_size())
+            logging.debug(f"shift pressed, mouse_pos: {pos}, phys_pos: {p_pos}")
+
+            # move selected stream to position
+            self._move_stream_to_pos(p_pos)
+    
+        else:
+            logging.debug("invalid correlation event, passing event to canvas")
+            active_canvas.on_left_down(evt)       # super event passthrough      
+
+    def on_char(self, evt: wx.Event) -> None:
+        """handle key presses
+        :param evt: (wx.Event) the event"""
+
+        if not self.correlation_enabled():
+            logging.debug("correlation not enabled, passing event to canvas")
+            active_canvas = evt.GetEventObject()
+            active_canvas.on_char(evt)            # super event passthrough
+
+        key = evt.GetKeyCode()
+        shift_mod = evt.ShiftDown()
+
+        ### CONTROLS ##############################
+        # SHIFT + LEFT CLICK -> MOVE_TO_POSITION
+        # LEFT, RIGHT -> TRANSLATION X
+        # UP, DOWN -> TRANSLATION Y
+        # SHIFT + LEFT, RIGHT -> ROTATION
+        # SHIFT + UP, DOWN -> SCALE 
+        ###########################################
+        dx, dy, dr, dpx  = 0, 0, 0, 0
+
+        # correlation control modifiers
+        if shift_mod:
+            dr = math.radians(self._panel.dr_step_cntrl.GetValue()) 
+            dpx = self._panel.dpx_step_cntrl.GetValue() / 100
+        else:
+            dx = self._panel.dx_step_cntrl.GetValue()
+            dy = self._panel.dy_step_cntrl.GetValue()
+
+        logging.debug(f"key: {key}, shift: {shift_mod}")
+        logging.debug(f"dx: {dx}, dy: {dy}, dr: {dr}, dpx: {dpx}")
+
+        if key == wx.WXK_LEFT:
+            self._move_stream(-dx, 0, -dr, 0)
+        elif key == wx.WXK_RIGHT:
+            self._move_stream(dx, 0, dr, 0)
+        elif key == wx.WXK_UP:
+            self._move_stream(0, dy, 0, dpx)
+        elif key == wx.WXK_DOWN:
+            self._move_stream(0, -dy, 0, -dpx)
+    
+    def _move_stream(self, dx: float, dy: float , dr: float = 0, dpx: float = 0) -> None:
+        """move the selected stream by the specified amount. the stream is forced 
+        to update in the views.
+        :param dx: (float) the change in x translation (metres)
+        :param dy: (float) the change y translation (metres)
+        :param dr: (float) the change in rotation (radians)
+        :param dpx: (float) the change in scale of the pixelsize (percentage)
+        """
+        if self._tab_data_model.selected_stream.value is None:
+            return
+        
+        s = self._tab_data_model.selected_stream.value
+    
+        logging.debug(f"move stream {s.name.value}: {dx}, {dy}, {dr}, {dpx}")
+
+        # translation
+        p = s.raw[0].metadata[model.MD_POS_COR]
+        if len(p) == 2:
+            x, y = p
+            s.raw[0].metadata[model.MD_POS_COR] = (x - dx, y - dy) # correlation direction is reversed
+        else:
+            x, y, z = p
+            s.raw[0].metadata[model.MD_POS_COR] = (x - dx, y - dy, z)
+
+        # rotation
+        rotation = s.raw[0].metadata.get(model.MD_ROTATION_COR, 0)
+        s.raw[0].metadata[model.MD_ROTATION_COR] = (rotation + dr)
+        
+        # scale (pixel size)
+        scalecor = s.raw[0].metadata.get(model.MD_PIXEL_SIZE_COR, (1, 1))
+        s.raw[0].metadata[model.MD_PIXEL_SIZE_COR] = (scalecor[0] + dpx, scalecor[1] + dpx)
+
+        # TODO: split x, y scale, add shear?
+
+        # update the image in the views
+        update_image_in_views(s, self._tab_data_model.views.value)
+        
+        # fit the source viewports to the content, as the image may have moved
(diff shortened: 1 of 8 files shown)


src/odemis/gui/cont/tabs.py
```diff
@@ -1755,6 +1756,284 @@ def get_display_priority(cls, main_data):
...
+        """
+        guiutil.enable_tab_on_stage_position(
+            self.button,
+            self.main_data.posture_manager,
+            self._allowed_targets,
+            tooltip=self.DISABLED_TAB_TOOLTIP.get(self.main_data.role)
+        )
+
+    def terminate(self):
+        super(CorrelationTab, self).terminate()
```
pieleric, line 2015: Since Python 3, you can write simply `super().terminate()`. Please do so.

src/odemis/gui/cont/tabs.py
```diff
@@ -1755,6 +1756,284 @@ def get_display_priority(cls, main_data):
...
+        self._stage.position.unsubscribe(self._on_stage_pos)
+        # make sure the streams are stopped
+        for s in self.tab_data_model.streams.value:
+            s.is_active.value = False
+
+    @classmethod
+    def get_display_priority(cls, main_data):
+        if main_data.role in ("enzel", "meteor", "mimas"):
+            return 2
```
pieleric, line 2024: 1 should be enough. Essentially it should never be the first one.

src/odemis/gui/cont/tabs.py
```diff
@@ -1755,6 +1756,284 @@ def get_display_priority(cls, main_data):
...
+        else:
+            return None
+
+    def Show(self, show=True):
+        assert (show != self.IsShown())  # we assume it's only called when changed
+        super(CorrelationTab, self).Show(show)
```
pieleric, line 2030: -> `super()`

src/odemis/gui/cont/tabs.py
```diff
@@ -1755,6 +1756,284 @@ def get_display_priority(cls, main_data):
...
+
+        if not show:  # if localization tab is not chosen
```
pieleric, line 2032: localization -> correlation (or "this")

src/odemis/gui/cont/tabs.py
```diff
@@ -1755,6 +1756,284 @@ def get_display_priority(cls, main_data):
...
+            # pause streams when not displayed
+            self._streambar_controller.pauseStreams()
```
pieleric, line 2034: That shouldn't be needed, as it only shows static streams. => remove the whole `if` block.

src/odemis/gui/cont/tabs.py
```diff
@@ -1755,6 +1756,284 @@ def get_display_priority(cls, main_data):
...
+    def terminate(self):
+        super(CorrelationTab, self).terminate()
+        self._stage.position.unsubscribe(self._on_stage_pos)
+        # make sure the streams are stopped
+        for s in self.tab_data_model.streams.value:
+            s.is_active.value = False
+
+    @classmethod
+    def get_display_priority(cls, main_data):
+        if main_data.role in ("enzel", "meteor", "mimas"):
```
pieleric, line 2023: ENZEL  and MIMAS too? I think for now it should be only for METEOR. The other systems are already physically aligned, so they shouldn't be the same need.

src/odemis/gui/cont/tabs.py
```diff
@@ -1755,6 +1756,284 @@ def get_display_priority(cls, main_data):
...
+            return 2
+        else:
+            return None
+
+    def Show(self, show=True):
+        assert (show != self.IsShown())  # we assume it's only called when changed
+        super(CorrelationTab, self).Show(show)
+
```
pieleric, line 2031: If it's shown, and there is no stream yet, should it automatically load the FM overview streams from the localization tab? That would avoid the user having to locate them again in the file system.

src/odemis/gui/cont/correlation.py
```diff
@@ -0,0 +1,242 @@
...
+            self._move_stream_to_pos(p_pos)
+    
+        else:
+            active_canvas.on_left_down(evt)       # super event passthrough      
+
+    def _on_char(self, evt):
+        """handle key presses"""
+        _key = evt.GetKeyCode()
+        _shift_mod = evt.ShiftDown()
+        _ctrl_mod = evt.ControlDown()
```
pieleric, line 140: unused?

src/odemis/gui/model/__init__.py
```diff
@@ -876,6 +876,26 @@ def _on_project_path_change(self, _):
...
+    def __init__(self, main):
+        super().__init__(main)
+
+        # Current tool selected (from the toolbar)
+        tools = {TOOL_NONE}
+        # Update the tool selection with the new tool list
+        self.tool.choices = tools
+
+        # for export tool
+        self.acq_fileinfo = VigilantAttribute(None) # a FileInfo
```
pieleric, line 896: I think this is to display "the" file information (name/date) in the analysis tab. Here, there are typically at least 2 files opened, and the filename doesn't really mater. => Drop the display (if there is) and drop this VA?

src/odemis/gui/cont/correlation.py
```diff
@@ -0,0 +1,242 @@
...
+        
+        self._tab_data_model = tab_data
+        self._main_data_model = tab_data.main
+        self._panel = panel
+        self._tab = tab
+        self._viewports = viewports
+
+        # the correlation streams, and selected stream
+        self.correlationStreams = model.ListVA()
+        self.selected_stream = model.VigilantAttribute(None)
```
pieleric, line 71: That represent the state of the GUI, so it should go on the model: CryoCorrelationGUIData.
Note that all tab models have a `.streams` VA which should be enough to use, instead of `correlationStreams`. 
For `selected_stream`, I'd recommend the casing `selectedStream`, to match the rest of that file.

src/odemis/gui/cont/correlation.py
```diff
@@ -0,0 +1,242 @@
...
+            vp.canvas.Bind(wx.EVT_CHAR, self._on_char)
+            vp.canvas.Bind(wx.EVT_LEFT_DOWN, self._on_mouse_down)
+       
+        # TODO: add toggle for enablling moving the streams. 
+        # If not enabled, the streams dont move, use standard controls for mouse / keyboard events...
+        # TODO: add reset button to reset the stream position to the original position
+
+    def _on_correlation_streams_change(self, streams):
+        """hide/show the correlation panel if there are no streams"""
+        _visible = len(streams) != 0
```
pieleric, line 89: No need to use `_` for local variables.

src/odemis/gui/cont/correlation.py
```diff
@@ -0,0 +1,242 @@
...
+
+        for vp in self._viewports:
+            vp.canvas.Bind(wx.EVT_CHAR, self._on_char)
+            vp.canvas.Bind(wx.EVT_LEFT_DOWN, self._on_mouse_down)
+       
+        # TODO: add toggle for enablling moving the streams. 
+        # If not enabled, the streams dont move, use standard controls for mouse / keyboard events...
+        # TODO: add reset button to reset the stream position to the original position
+
+    def _on_correlation_streams_change(self, streams):
```
pieleric, line 87: Typically, every VA listener which calls a wxPython function should be protected with `@call_in_wx_main()`. Otherwise it may sometimes crash the whole GUI. (because the VA might be changed from any thread, not necessarily the main GUI thread)

pieleric (changes requested):

src/odemis/gui/cont/tabs.py
patrickcleeve2: Would we want to always add the overview streams here? not just when nothing is shown

src/odemis/gui/model/__init__.py
patrickcleeve2: This is required to use the existing export tool, it checks for this variable for a potential filename, but if its None, just uses the datetime. So needs to exist, but can be None

https://github.com/patrickcleeve2/odemis/blob/372028261d6b29a5dbab538856a7e1ffdf556dfa/src/odemis/gui/cont/export.py#L141-L142

src/odemis/gui/cont/tabs.py
pieleric: As discussed, let's make it the same behaviour as the chamber tab: automatically add all the acquired overview.

We probably could automatically empty the streams in the correlation tab when the user changes project (as for the localization tab).

src/odemis/gui/cont/correlation.py
```diff
@@ -0,0 +1,247 @@
+
+# -*- coding: utf-8 -*-
+
+"""
+@author Patrick Cleeve
+
+Copyright © 2023, Delmic
+
+Handles the correlation .
```
tepals, line 9: what correlation does it handle?

src/odemis/gui/cont/correlation.py
```diff
@@ -0,0 +1,247 @@
...
+import wx
+# IMPORTANT: wx.html needs to be imported for the HTMLWindow defined in the XRC
+# file to be correctly identified. See: http://trac.wxwidgets.org/ticket/3626
+# This is not related to any particular wxPython version and is most likely permanent.
+import wx.html
+from odemis import model
+import odemis.acq.stream as acqstream
+import odemis.gui.model as guimod
+from odemis.acq.stream import StaticStream
+from odemis.gui.util import call_in_wx_main
```
tepals, line 36: can you put the imports first, and then the from odemis import? See: https://github.com/delmic/odemis/blob/master/doc/develop/coding-guidelines.rst#import-order

src/odemis/gui/cont/correlation.py
```diff
@@ -0,0 +1,247 @@
...
+
+        # the offset is the difference between the current cor position and the new position
+        # that is how much we need to move the current correlation position by.
+        dx = cx - nx
+        dy = cy - ny
+        
+        logging.debug(f"cur_pos: {x}, {y}, cor_pos: {cx}, {cy}, new_pos: {nx}, {ny}, offset_pos: {dx}, {dy}")
+
+        # move the stream using the correlation position offset
+        self._move_stream(dx=dx, dy=dy, dr=0, dpx=0)
```
tepals, line 247: can you add an empty line at the end of the file?

src/odemis/gui/cont/tabs.py
```diff
@@ -1756,6 +1757,300 @@ def get_display_priority(cls, main_data):
...
+        super(CorrelationTab, self).__init__(
+            name, button, panel, main_frame, tab_data)
+    
+        # panel = tab
+        self.main_data = main_data
+
+        # First we create the views, then the streams
+        vpv = self._create_views(main_data, panel.pnl_secom_grid.viewports)
+        
+        # Order matters!
```
tepals, line 1780: in what way does order matter?

src/odemis/gui/cont/tabs.py
```diff
@@ -1756,6 +1757,300 @@ def get_display_priority(cls, main_data):
...
+                (panel.vp_secom_tl, panel.lbl_secom_view_tl)),
+            (panel.btn_secom_view_tr,
+                (panel.vp_secom_tr, panel.lbl_secom_view_tr)),
+            (panel.btn_secom_view_bl,
+                (panel.vp_secom_bl, panel.lbl_secom_view_bl)),
+            (panel.btn_secom_view_br,
+                (panel.vp_secom_br, panel.lbl_secom_view_br)),
+        ])
+
+        # remove the play overlay from all streams
```
tepals, line 1801: why?

src/odemis/gui/main.py
```diff
@@ -189,6 +189,12 @@ def init_gui(self):
                     "controller": tabs.LocalizationTab,
                     "button": self.main_frame.btn_tab_localization,
                     "panel": main_xrc.xrcpnl_tab_localization
+                },
+                                {
```
tepals, line 193: can you align this bracket with the one above?

tepals (changes requested):
Looks good, I had a quick look, did not test it

src/odemis/gui/cont/correlation.py
```diff
@@ -0,0 +1,269 @@
...
+    def _add_correlation_stream(self, s: StaticStream): 
+        """add a stream to the correlation controller"""
+        # reset the stream correlation data
+        self._reset_stream_correlation_data(s)
+
+        # add the stream to the correlation streams
+        self._tab_data_model.streams.value.append(s)
+        
+
+    def _update_correlation_streams(self, streams):
```
pieleric, line 127: It seems this function is the same as `_add_correlation_steam` but supports passing multiple streams. Is that correct, or are you going to change it to also remove streams?

In general, the docstring is missing so I'm not sure what your intention was, but I feel the name is not matching the behaviour. If it's "update" then probably it should also removes the streams which are not listed in the argument? Also why "correlation" ? Is that because it resets the correction metadata?

src/odemis/gui/cont/correlation.py
```diff
@@ -0,0 +1,269 @@
...
+        self._tab_data_model.streams.subscribe(self._update_correlation_cmb, init=True)
+
+        # connect the correlation streams to the tab data
+        self._panel.cmb_correlation_stream.Bind(wx.EVT_COMBOBOX, self._on_selected_stream_change)
+
+        for vp in self._viewports:
+            vp.canvas.Bind(wx.EVT_CHAR, self._on_char)
+            vp.canvas.Bind(wx.EVT_LEFT_DOWN, self._on_mouse_down)
+       
+    def _update_correlation_cmb(self, streams):
```
pieleric, line 89: Needs to be protected by `@call_in_wx_main`.

src/odemis/gui/cont/correlation.py
patrickcleeve2: changed to add_streams, added docstring and note explaining intended behaviour

src/odemis/gui/cont/correlation.py
```diff
@@ -0,0 +1,280 @@
...
+import odemis.gui.model as guimod
+from odemis import model
+from odemis.acq.stream import StaticStream
+from odemis.gui.util import call_in_wx_main
+
+# TODO: move to more approprate location
+def update_image_in_views(s: StaticStream, views: list):
+    """force update the static stream in the selected views
+    param: s (StaticStream) the static stream to update
+    param: views (list[View]) the list of views to update"""
```
K4rishma, line 42: docstring format from the coding guidelines
```
    def get_equality(arg1, arg2):
        """
        Summary line.
        Extended description of function.
        :param arg1: (int) Description of arg1.
        :param arg2: (str) Description of arg2.
        :raise: ValueError if arg1 is equal to arg2
        :return: (bool) Description of return value
        """
        if arg1 == arg2:
            raise ValueError('arg1 must not be equal to arg2')
        return True
```

src/odemis/gui/cont/correlation.py
```diff
@@ -0,0 +1,280 @@
...
+                sp._shouldUpdateImage()
+
+
+# TODO: add toggle for enablling moving the streams. 
+# If not enabled, the streams dont move, use standard controls for mouse / keyboard events...
+class CorrelationController(object):
+
+    def __init__(self, tab_data, panel, tab, viewports):
+        """
+        tab_data (MicroscopyGUIData): the representation of the microscope GUI
```
K4rishma, line 60: docstring format same as before

src/odemis/gui/cont/correlation.py
```diff
@@ -0,0 +1,280 @@
...
+
+    def _reset_stream_correlation_data(self, s: StaticStream):
+        """reset the stream position to the original position / rotation / scale"""
+        s.raw[0].metadata[model.MD_POS_COR] = (0, 0)
+        s.raw[0].metadata[model.MD_ROTATION_COR] = 0
+        s.raw[0].metadata[model.MD_PIXEL_SIZE_COR] = (1, 1)       
+
+    def add_streams(self, streams: list):
+        """add streams to the correlation tab
+        param: streams (list[StaticStream]) the streams to add"""
```
K4rishma, line 128: docstring format

src/odemis/gui/model/__init__.py
```diff
@@ -876,6 +876,26 @@ def _on_project_path_change(self, _):
...
+
+
```
K4rishma, line 898: 4 line space between two classes, I think generally there are two

src/odemis/gui/cont/correlation.py
```diff
@@ -0,0 +1,280 @@
...
+
+            # move selected stream to position
+            self._move_stream_to_pos(p_pos)
+    
+        else:
+            active_canvas.on_left_down(evt)       # super event passthrough      
+
+    def _on_char(self, evt):
+        """handle key presses"""
+        _key = evt.GetKeyCode()
```
pieleric, line 177: Please drop the `_` before the name of the variable. It's local, so it never needs to be "private".

src/odemis/gui/cont/correlation.py
```diff
@@ -0,0 +1,280 @@
...
+        _shift_mod = evt.ShiftDown()
+
+        # correlation control modifiers
+        dx = self._panel.dx_step_cntrl.GetValue() * 1e-6
+        dy = self._panel.dy_step_cntrl.GetValue() * 1e-6
+        dr = numpy.deg2rad(self._panel.dr_step_cntrl.GetValue())
```
pieleric, line 183: For single values, typically we avoid using numpy. You can just use `math.radians()` here.

src/odemis/gui/main.xrc
```diff
@@ -9085,6 +9103,481 @@
...
+                                      <flag>wxALIGN_CENTRE_VERTICAL</flag>
+                                      <border>10</border>
+                                  </object>
+                                  <object class="sizeritem">
+                                    <object class="UnitFloatCtrl" name="dy_step_cntrl">
+                                      <value>1</value>
+                                      <key_step>0.000001</key_step>
+                                      <min>0</min>
+                                      <max>10</max>
+                                      <unit>um</unit>
```
pieleric, line 9434: The base unit should just be `m`.

src/odemis/gui/cont/correlation.py
```diff
@@ -0,0 +1,280 @@
...
+            active_canvas.on_left_down(evt)       # super event passthrough      
+
+    def _on_char(self, evt):
+        """handle key presses"""
+        _key = evt.GetKeyCode()
+        _shift_mod = evt.ShiftDown()
+
+        # correlation control modifiers
+        dx = self._panel.dx_step_cntrl.GetValue() * 1e-6
+        dy = self._panel.dy_step_cntrl.GetValue() * 1e-6
```
pieleric, line 182: Normally all the controls are in the base unit (eg, "m"), and the control itself knows how to display it nicely. So if the value is 1e-6, it'll show "1 µm". You then don't need to do the conversion (and the user can type any number and it'll be displayed properly).

src/odemis/gui/cont/correlation.py
```diff
@@ -0,0 +1,280 @@
...
+        # LEFT, RIGHT -> TRANSLATION X
+        # UP, DOWN -> TRANSLATION Y
+        # SHIFT + LEFT, RIGHT -> ROTATION
+        # SHIFT + UP, DOWN -> SCALE 
+        ###########################################
+
+        if _shift_mod:
+            dx, dy = 0, 0
+        else:
+            dr, dpx = 0, 0
```
pieleric, line 197: I feel it's an inverted logic. Can you start by default all the values are set to 0, and based on shift, read just the right values from the widgets?

src/odemis/gui/cont/correlation.py
```diff
@@ -0,0 +1,280 @@
...
+        # TODO: split x, y scale, add shear?
+
+        # update the image in the views
+        views = self._tab_data_model.views.value
+        update_image_in_views(s, views)
+        
+        # fit to content (TODO: these are not generic, might be different for other tabs)
+        try:
+            self._panel.vp_secom_tl.canvas.fit_view_to_content()
+            self._panel.vp_secom_tr.canvas.fit_view_to_content()
```
pieleric, line 246: These names have changed, right? Need to update them.

For now your code only handles one tab, so let's not try to handle the cases for the other tabs. => drop the exception handler.

src/odemis/gui/cont/correlation.py
```diff
@@ -0,0 +1,280 @@
...
+        except AttributeError as e:
+            logging.debug(f"Requested views {e} are not availble to update...")
+    
+    def _move_stream_to_pos(self, pos: tuple):
+        """move the selected stream to the position pos"""
```
pieleric, line 251: Please add a description of the `pos` parameter. In particular, indicate the unit.

src/odemis/gui/cont/correlation.py
```diff
@@ -0,0 +1,280 @@
...
+        if _key == wx.WXK_LEFT:
+            self._move_stream(-dx, 0, -dr, 0)
+        elif _key == wx.WXK_RIGHT:
+            self._move_stream(dx, 0, dr, 0)
+        elif _key == wx.WXK_UP:
+            self._move_stream(0, dy, 0, dpx)
+        elif _key == wx.WXK_DOWN:
+            self._move_stream(0, -dy, 0, -dpx)
+    
+    def _move_stream(self, dx, dy, dr=0, dpx=0):
```
pieleric, line 211: Please add the type and description of the parameters.

src/odemis/gui/cont/correlation.py
```diff
@@ -0,0 +1,280 @@
...
+        # the difference between the clicked position, and the position in metadata
+        # is the offset (be careful because correlation is sign flipped)
+        s = self.selectedStream.value
+        
+        # the cur_pos is the realspace position of the image
+        p = s.raw[0].metadata[model.MD_POS]
+        if len(p) == 2:
+            x, y = p
+        else:
+            x, y, z = p
```
pieleric, line 261: If you're never going to use `z`, you can replace this code by just:
```suggestion
        x, y = p[:2]
```

src/odemis/gui/cont/correlation.py
```diff
@@ -0,0 +1,280 @@
...
+        
+        # the correlation pos is the change in position in the viewer
+        cx, cy = s.raw[0].metadata[model.MD_POS_COR]
+
+        # the new position is the different between the clicked position and the realspace position
```
pieleric, line 266: different -> difference.

src/odemis/gui/cont/correlation.py
```diff
@@ -0,0 +1,280 @@
...
+        """
+        
+        self._tab_data_model = tab_data
+        self._main_data_model = tab_data.main
+        self._panel = panel
+        self._tab = tab
+        self._viewports = viewports
+
+        # selected stream
+        self.selectedStream = model.VigilantAttribute(None)
```
pieleric, line 73: That VA should be an attribute of `CryoCorrelationGUIData` (so it's on `tab_data`).

src/odemis/gui/cont/tabs.py
```diff
@@ -1756,6 +1762,204 @@ def get_display_priority(cls, main_data):
...
+        # correlation controller
+        self._correlation_controller = CorrelationController(
+            tab_data,
+            panel, 
+            self, 
+            panel.pnl_correlaton_grid.viewports 
+        )
+     
+        # export controller
+        # TODO: export button must currently be called btn_secom_export,
```
pieleric, line 1826: We probably could rename everything to `btn_export`.

pieleric (changes requested):

src/odemis/gui/cont/correlation.py
patrickcleeve2: Is this what you mean? I think it should stay here, as it is only used inside this widget. It makes it transferable to other tabs in the future.

src/odemis/gui/cont/correlation.py
```diff
@@ -0,0 +1,282 @@
...
+        for sp in v.stream_tree.getProjections():  # stream or projection
+            if isinstance(sp, acqstream.DataProjection):
+                st = sp.stream
+            else:
+                st = sp
+            if st is s:
+                sp._shouldUpdateImage()
+
+
+# TODO: add toggle for enablling moving the streams.
```
pieleric, line 54: Maybe the correlation stream combo-box could have a special entry "None", in which case no stream is selected, and the special keyboards shortcuts are not active?

src/odemis/gui/cont/correlation.py
patrickcleeve2: OK, understand. I've moved it into GUIData

src/odemis/gui/cont/correlation.py
patrickcleeve2: Added a checkbox that enables/disables correlation controls. If disabled, events are passed through

src/odemis/gui/cont/correlation.py
```diff
@@ -0,0 +1,294 @@
...
+import odemis.gui.model as guimod
+from odemis import model
+from odemis.acq.stream import StaticStream
+from odemis.gui.util import call_in_wx_main
+
+# TODO: move to more approprate location
+def update_image_in_views(s: StaticStream, views: list) -> None:
+    """Force update the static stream in the selected views
+    param s: (StaticStream) the static stream to update
+    param views: (list[View]) the list of views to update"""
```
tepals, line 42: Can you use `:param [name]:` in all docstrings, you do it in most of them, but missed a few.
```suggestion
    :param s: (StaticStream) the static stream to update
    :param views: (list[View]) the list of views to update"""
```

patrickcleeve2: I think i've updated them all now

pieleric (approved):

K4rishma (approved):

PR #2605, opened 2023-12-15
[METEOR-880] Add support for stage movement with no rotation

patrickcleeve2:
Adds support for stage movement between SEM and METEOR without rotation for Thermo FIsher systems. 

**Implementation**
- Check if the current and target rotation (rz) are close to each other (~2.5  deg  tolerance)
- If they are close, move with translation only.
- If they are not close, move with previous rotation based movement. 

**Adds**
- Checks for required transformation to use
- Implementation of no rotation move
- Calibration script for manually adjusting the calibration data while odemis is running.

### DIFF
--- a/scripts/meteor-recalibrate-movement.py
+++ b/scripts/meteor-recalibrate-movement.py
@@ -1,0 +1,139 @@
+#!/usr/bin/env python3
+'''
+@author: Patrick Cleeve
+
+This script is intended only for METEOR on TFS FIBSEMs. 
+This script is to update the calibration metadata for the stage-bare component while odemis is running.
+
+The intended use for the script is to easily update the calibrate values for stage movements, when the meteor is
+moving with no rotation, i.e. moving between the FIB <-> METEOR positions, rather than between SEM <-> METEOR positions.
+ 
+The script will ask the user for the distance between the FIB and METEOR positions in the x and y directions. It will then convert these values
+to the calibration metadata format and update the stage-bare component with the new values.
+
+run as:
+python3 ./scripts/meteor-recalibrate-movement.py 
+
+and follow the instructions in the terminal.
+
+'''
+
+# TODO: update the yaml file with these values too
+# TODO: create a plugin for this
+
+import math
+from odemis import model
+from odemis.acq.move import isNearPosition, ATOL_ROTATION_TRANSFORM
+from odemis.util.driver import isInRange
+from Pyro4.errors import CommunicationError
+
+def main():
+
+    print(f"METEOR Stage Movement Calibration Updater")
+    print(f"-" * 20)
+    
+    # get stage-bare component
+    try:
+        stage = model.getComponent(role="stage-bare")
+    except CommunicationError as e:
+        print("ERROR: Could not connect to odemis. Is odemis running?")
+        return
+
+    print(f"Current stage-bare component: {stage}")    
+    print(f"-" * 20)
+    
+    # check which movement mode the calibration data is intended for, if not FIB <-> METEOR, warn user.
+    stage_md = stage.getMetadata()
+    fm_pos_active = stage_md[model.MD_FAV_FM_POS_ACTIVE]
+    sem_pos_active = stage_md[model.MD_FAV_SEM_POS_ACTIVE]
+
+    # check if configured for  FIB <-> METEOR movement
+    # SEM <-> METEOR movement has rz axis movement, FIB <-> METEOR movement does not
+    has_rz = not isNearPosition(sem_pos_active, fm_pos_active, {"rz"}, atol_rotation=ATOL_ROTATION_TRANSFORM)
+
+    if has_rz:
+        sem_rz = f"{math.degrees(sem_pos_active['rz']):02f}"
+        flm_rz = f"{math.degrees(fm_pos_active['rz']):02f}"
+        err_msg = f"""ERROR: 
+        The current calibration data is for SEM <-> METEOR movement (movement with rotation).
+        The current rotation positions are: FIBSEM: {sem_rz} deg, METEOR: {flm_rz} deg
+        This script is intended for FIB <-> METEOR movement (movement with no rotation). 
+        If you wish to continue, please update the calibration data in the yaml file to FIB <-> METEOR movement. 
+        You will need to restart odemis for the changes to take effect.
+        """
+        print(err_msg)
+        return
+
+    print(f"Valid calibration data found for FIB <-> METEOR movement. Continuing...")
+    print('-' * 20)
+
+    print(f"Calibration values are the distance between the FIB and METEOR positions in the x and y directions.")
+    print(f"Current calibration values: {stage_md[model.MD_POS_COR]}")
+
+    # get imaging ranges
+    sem_active_range = stage_md[model.MD_SEM_IMAGING_RANGE]
+    flm_active_range = stage_md[model.MD_FM_IMAGING_RANGE]
+    working_range = {"x": [0, 0], "y": [0, 0]}
+
+    for axis in ["x", "y"]: 
+        margin = (sem_active_range[axis][1] - sem_active_range[axis][0]) * 0.05 # 5% of SEM imaging range
+        x0 = flm_active_range[axis][0] - sem_active_range[axis][0] - margin     # minimum move sem -> flm
+        x1 = flm_active_range[axis][1] - sem_active_range[axis][0] + margin     # maximum move sem -> flm
+        working_range[axis] = [x0, x1]
+ 
+    print(f"""Calibration data should be within the ranges: 
+        x range: {working_range['x']}
+        y range: {working_range['y']} 
+        """)
+
+    # get x, y values from user
+    x = input(f"Enter new x calibration value (m): ")
+    y = input(f"Enter new y calibration value (m): ")
+    print("-" * 20)
+
+    # check they are floats
+    try:
+        x = float(x) 
+        y = float(y)
+    except  ValueError:
+        print("Invalid input. Please enter a float. Calibration data has not been updated.")
+        return
+    
+    # validate range
+    x_stage_range = stage.axes["x"].range
+    y_stage_range = stage.axes["y"].range
+    # check within stage axes range
+    if not isInRange({"x": x}, {'x': x_stage_range}, {'x'}):
+        print(f"Invalid input. x value {x} is outside of the range of the stage {x_stage_range}. Calibration data has not been updated.")
+        return
+    
+    if not isInRange({"y": y}, {'y': y_stage_range}, {'y'}):
+        print(f"Invalid input. y value {y} is outside of the range of the stage {y_stage_range}. Calibration data has not been updated.")
+        return
+
+    # check if within imaging range
+    if not isInRange({"x": x, "y": y}, working_range, {'x', 'y'}):
+        print(f"""Invalid input. Requested calibration would put stage outside of fm imaging range.
+              The calibration should be within these ranges:  
+            x: {x} (range: {working_range['x']}.
+            y: {y} (range: {working_range['y']}. 
+            Calibration data has not been updated.""")
+        return
+
+    # get user confirmation for values
+    print(f"Change to new calibration values: {[x, y]} m")
+
+    ret = input(f"Confirm? (y/N): ")
+
+    if ret.lower() == "y": 
+
+        md_pos_cor = {model.MD_POS_COR: [x, y]}
+        stage.updateMetadata(md_pos_cor)
+
+        print(f"Calibration values updated to: {stage.getMetadata()[model.MD_POS_COR]}")
+    else:
+        print("Calibration aborted. The calibration data was not updated.")
+
+
+if __name__ == "__main__":
+    main()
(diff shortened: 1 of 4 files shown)


pieleric (commented):
Looks good. I'd suggest to use only the "implicit" method. It should then be documented... "somewhere". (maybe in the METEOR installation manual).

scripts/update_metadata.py
pieleric, line 1: That's a quite too generic name. I'd suggest more something like `meteor-recalibrate-center`. (also, as it's a script, and can be run directly as `./meteor-recalibrate-center.py`, it's preferable to use `-` over `_`, as it's the convention in Linux.

scripts/update_metadata.py
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env python3
+'''
+@author: Patrick Cleeve
+
+This is a script to update the MD_POS_COR for the stage-bare component while odemis is running.
```
pieleric, line 5: You should at least indicate that it's for the METEOR, on TFS SEMs. Maybe also explain a bit why someone would want to change that value.

scripts/update_metadata.py
```diff
@@ -0,0 +1,55 @@
...
+
+and follow the instructions in the terminal.
+'''
+
+
+from odemis import model
+
+def main():
+
+    print(f"Calibration metadata updater")
```
pieleric, line 18: -> "METEOR calibration updater"?

"metedata" is really an implementation detail because in Odemis we started to call this sort of data "metadata", but that doesn't match the common English term of "metadata".

scripts/update_metadata.py
```diff
@@ -0,0 +1,55 @@
...
+
+    stage = model.getComponent(role="stage-bare")
+    print(f"-"*20)
```
pieleric, line 21: Please add spaces around the `*` (everywhere in the file)

scripts/update_metadata.py
```diff
@@ -0,0 +1,55 @@
...
+        y = float(y)
+    except  ValueError:
+        print("Invalid input. Please enter a float. Calibration data has not been updated.")
+        return
+    
+
+    # get user confirmation for values
+    print(f"Change to new calibration metadata: {[x, y]}")
+
+    ret = input(f"Confirm? (y/n): ")
```
pieleric, line 42: The default is "n", so the convention is to indicate it by using an uppercase. -> "y/N".

scripts/update_metadata.py
```diff
@@ -0,0 +1,55 @@
...
+
+    if ret.lower() == "y": 
+
+        md_pos_cor = {model.MD_POS_COR: [x, y]}
+        stage.updateMetadata(md_pos_cor)
+
+        print(f"Calibration metadata updated to: {stage.getMetadata()[model.MD_POS_COR]}")
```
pieleric, line 49: Please drop "metadata" from the user messages.

scripts/update_metadata.py
```diff
@@ -0,0 +1,55 @@
...
+
+    # get x , y values from user
+    x = input(f"Enter new x calibration value: ")
+    y = input(f"Enter new y calibration value: ")
+    print("-"*20)
+
+    # check they are floats
+    try:
+        x = float(x)
+        y = float(y)
```
pieleric, line 33: Wasn't there a plan to divide by 2?

scripts/update_metadata.py
```diff
@@ -0,0 +1,55 @@
...
+    except  ValueError:
+        print("Invalid input. Please enter a float. Calibration data has not been updated.")
+        return
+    
+
+    # get user confirmation for values
+    print(f"Change to new calibration metadata: {[x, y]}")
```
pieleric, line 40: Please add the unit at the end of the message: `m`.

scripts/update_metadata.py
```diff
@@ -0,0 +1,55 @@
...
+    print(f"Current calibration metadata: {stage.getMetadata()[model.MD_POS_COR]}")
+
+    # get x , y values from user
+    x = input(f"Enter new x calibration value: ")
+    y = input(f"Enter new y calibration value: ")
+    print("-"*20)
+
+    # check they are floats
+    try:
+        x = float(x)
```
pieleric, line 32: How bad could things go if the value is incorrect? I'm a little worried the user types the values in mm instead of m. Could the stage go suddenly somewhere really bad? Maybe at least check that the values are "reasonable". For instance, check they are within the axes ranges, using `stage_bare.axes["x"].range`

pieleric (changes requested):

K4rishma (approved):

scripts/update_metadata.py
patrickcleeve2: added, it might be a good idea to add something similar before the actual move too, as these values might be edited from the yaml too?

scripts/update_metadata.py
patrickcleeve2: I think we should change this calibration value to just be the distance between the FIB and METEOR, rather than half it. The reason it was half was to match the other rotation calibration, which used the radius and then multipled by two, but there is no real reason for them to be the same? 

I think it being half makes it really confusing for the user either way because they either have to measure the distance then divide by two (for some unknnown reason to them), or they enter the actual distance (e.g. 4) and then get a message saying there calibration value has been set to 2. 


I think the real solution will be to add an additional calibration parameter to separate the calibration for rotated movements, and non-rotated moves, or somehow unify them. That would avoid this issue entirely, and make a better user experience because they could move from SEM <>METEOR or FIB <> METEOR without having to redo calib?
Im happy to implement that change in this PR, or merge this and add it separately?

patrickcleeve2: @pieleric  I've changed the calibration for the no rotation move to not multiple by two (so dont need to divide the distance by two). I think this will be much less confusing for users. 

I've added a check in the calibration script to check that the values are within the stage axes (should prevent wildly bad values due to units etc), and to check the transform values are within the ranges defined for SEM <> FM imaging. It also checks that the yaml is setup for no rotation movements, and won't let you change the values if not. I've added a warning on the actual transformation calc if the transformation puts the stage outside the imaging range, which can be changed to raise an error if we want to be stricter there.  

I think this should prevent the worst errors for now as its only for 1 system, and we can re open this as part of the new movement system.

pieleric (changes requested):

src/odemis/acq/move.py
```diff
@@ -1798,3 +1839,22 @@ def _getCurrentAlignerPositionLabel(self) -> int:
...
         # None of the above -> unknown position
         return UNKNOWN
+
+### utils   
+def get_movement_mode(current_pos: Dict[str, float], target_pos: Dict[str, float]) -> str:
+    """Get the movement mode for the requested stage movement. Movement mode refers to 
+    whether the rotation axis is to be moved or not. Movement mode determines the transformation used.
+    :param current_pos: (dict str->float) the current position of the stage
+    :param target_pos: (dict str->float) the target position of the stage
+    :return: (str) the movement mode ["rz", "no_rz"]"""
```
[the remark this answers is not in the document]
patrickcleeve2: changed to bool has_rz

src/odemis/acq/move.py
```diff
@@ -1798,3 +1839,22 @@ def _getCurrentAlignerPositionLabel(self) -> int:
             return LOADING_PATH
         # None of the above -> unknown position
         return UNKNOWN
+
+### utils   
+def get_movement_mode(current_pos: Dict[str, float], target_pos: Dict[str, float]) -> str:
```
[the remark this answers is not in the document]
patrickcleeve2: I've added a more general helper, has_axis_move which will check for a desired movement on the specified axes. I've also added a test for it, because it the previous function has unintuitive default behaviour that needs to be kept for now. 

```python
has_rz: bool = has_axis_move(current_pos, target_pos, "rz", tolerance)
```

scripts/meteor-recalibrate-movement.py
```diff
@@ -0,0 +1,132 @@
...
+        return
+
+    print(f"Current stage-bare component: {stage}")    
+    print(f"-" * 20)
+    
+    # check which movement mode the calibration data is intended for, if not FIB <-> METEOR, warn user.
+    stage_md = stage.getMetadata()
+    fm_pos_active = stage_md[model.MD_FAV_FM_POS_ACTIVE]
+    sem_pos_active = stage_md[model.MD_FAV_SEM_POS_ACTIVE]
+    has_rz = has_axis_move(sem_pos_active, fm_pos_active, "rz") # returns False if FIB <-> METEOR movement
```
pieleric, line 49: The comment is odd. -> `False if  FIB<>METEOR switch does a 180° rotation`?

scripts/meteor-recalibrate-movement.py
```diff
@@ -0,0 +1,132 @@
...
+
+    for axis in ["x", "y"]: 
+        margin = (sem_active_range[axis][1] - sem_active_range[axis][0]) * 0.05 # 5% of SEM imaging range
+        x0 = flm_active_range[axis][0] - sem_active_range[axis][0] - margin     # minimum move sem -> flm
+        x1 = flm_active_range[axis][1] - sem_active_range[axis][0] + margin     # maximum move sem -> flm
+        working_range[axis] = [x0, x1]
+ 
+    print(f"""Calibration data should be within the ranges: 
+          \nx range: {[f'{v:4f}' for v in working_range['x']]}
+          \ny range: {[f'{v:4f}' for v in working_range['y']]}
```
pieleric, line 80: Please don't mix `\n` and triple quote strings. The whole point of triple quote strings is that you don't need `\n`, and you can write down directly as it'll appear. So just explicity add an empty line, or drop the `\n`.

scripts/meteor-recalibrate-movement.py
```diff
@@ -0,0 +1,132 @@
...
+    has_rz = has_axis_move(sem_pos_active, fm_pos_active, "rz") # returns False if FIB <-> METEOR movement
+
+    if has_rz:
+        sem_rz = f"{math.degrees(sem_pos_active['rz']):02f}"
+        flm_rz = f"{math.degrees(fm_pos_active['rz']):02f}"
+        print(f"""ERROR: The current calibration data is for SEM <-> METEOR movement.
+              \nThe current rotation positions are: FIBSEM: {sem_rz} deg, METEOR: {flm_rz} deg
+              \nThis script is intended for FIB <-> METEOR movement (movement with no rotation). 
+              \nIf you wish to continue, please update the calibration data in the yaml file to FIB <-> METEOR movement. 
+              \nYou will need to restart odemis for the changes to take effect.""")
```
pieleric, line 58: Please drop the `\n`'s.

src/odemis/acq/move.py
```diff
@@ -61,6 +61,9 @@
 SAFETY_MARGIN_5DOF = 100e-6  # m
 SAFETY_MARGIN_3DOF = 200e-6  # m
 
+# tolerance for the movement axes checks
+ATOL_MOVEMENT_ROTATION = 0.04  # rad ~2.5 deg
+ATOL_MOVEMENT_TRANSLATION = 0.1e-6 # 0.1 um
```
pieleric, line 66: Be aware we already have such constants in odemis.util.driver `ATOL_*_POS`.  They are quite different though... but I'm not sure there is a good reason for it. They are used by `isNearPosition()`, which indicates where two positions are similar.
I think you could probably use `isNearPosition(... {"rz"})` to replace `has_axis_move`. Possibly extend that function to accept an optional argument `atol`.
Alternatively, you'd need to extend a little bit the comment to indicate they should not be mixed up with the util.driver.ATOL_* constants used to

src/odemis/acq/test/move_test.py
```diff
@@ -939,6 +939,40 @@ def test_get_progress_lin_rot(self):
...
+        
+        self.assertTrue(has_axis_move(start, end, 'x'))
+        self.assertTrue(has_axis_move(start, end, 'y'))
+        self.assertTrue(has_axis_move(start, end, 'z'))
+        self.assertTrue(has_axis_move(start, end, 'rx'))
+        self.assertTrue(has_axis_move(start, end, 'rz'))
+
+        # negative tests
+        start = {'x': 0.023, 'y': 0.032, 'z': 0.01, "rx": 0, "rz": 0}
+        end = {'x': 0.023, 'y': 0.032, 'z': 0.01, "rx": 0, "rz": 0}
```
pieleric, line 963: If the values are identical, it's a little too easy. Please add some epsilon to at least a couple of axes!

pieleric (changes requested):

src/odemis/acq/move.py
patrickcleeve2: ah, i didnt see isNearPosition, i've changed this to use it now, and added the optional tolerance arg (one for linear, one for rot that default to driver.ATOL_*_POS). I've removed has_axis_move

the reasons for the additional constants, was that the rotation tolerance was 0.5 deg which is too strict for the leica stage, needed around 2.5deg. So i added these just for determining the transform, so they aren't as strict

pieleric (changes requested):

src/odemis/acq/test/move_test.py
```diff
@@ -562,6 +563,18 @@ def test_unknown_label_at_initialization(self):
         current_grid = self.posture_manager.getCurrentGridLabel()
         self.assertEqual(current_grid, None)
 
+    def test_transformFromSEMToMeteor(self):
+
+        # assert that raises error when no rz
```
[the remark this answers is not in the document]
patrickcleeve2: added explanation of previous behaviour, and changes we are checking for

pieleric (approved):

PR #2607, opened 2023-12-18
[SDEV-1761] odemis plugin to import export csv file with definitions of roas

nandishjpatel:
Ticket: https://delmic.atlassian.net/browse/SDEV-1761

### DIFF
--- a/plugins/fastem_roa_import_export.py
+++ b/plugins/fastem_roa_import_export.py
@@ -4,7 +4,7 @@ Created on 13 December 2023
 
 @author: Nandish Patel
 
-Gives ability to use Import Export ROAs tool under Help > Development.
+Gives ability to use Import ROAs, Export ROAs tool under Help > Development.
 
 Copyright © 2023 Nandish Patel, Delmic
 
@@ -24,50 +24,34 @@ see http://www.gnu.org/licenses/.
 import csv
 from enum import Enum
 import logging
+from typing import List, Union
 import wx
-from wx.lib.embeddedimage import PyEmbeddedImage
 
 from odemis.gui.plugin import Plugin
-
-
-IMPORT_IMAGE = PyEmbeddedImage(
-    b"iVBORw0KGgoAAAANSUhEUgAAABgAAAAVCAYAAABc6S4mAAAAqUlEQVQ4T2NkwA3+45FDlmoA"
-    b"chqBGKSeEV0PhgCSAnIsAGlXB+JbMHNoYQGK2bS0AGQRI60twIwUCuIAa5qghg/AQYErxVHL"
-    b"ApwpmhwLbgBN0yAyj5AVBzBHEZVPyPEBsh6CllBqASiklgBxNL5IxuUKfJZjMw+rOSBDaG4B"
-    b"Lt+RYjHOuKBGHOCNaEosaAd6vYJQfiDXAoLJk5r1AV5PkOMDQqGCIk9zCwB4bSsHCFUhIAAA"
-    b"AABJRU5ErkJggg=="
-)
-
-
-EXPORT_IMAGE = PyEmbeddedImage(
-    b"iVBORw0KGgoAAAANSUhEUgAAABgAAAAVCAYAAABc6S4mAAAAjUlEQVQ4T9WUgQqAIAxE51f3"
-    b"DX11qbRYhd7NHFQgBJ4+79xMIrLlgb50CIpW/9GaOl/EXoCuCwXQkFEHenoY11sAdOIFULlb"
-    b"EQuAJ22RPYDeHk1nswFLJq2jETEOtKfO6prtwDZthUQC6v4MADVT96n5BAA1l3V4d0NFNAoI"
-    b"veTQMr0UBXPJnoge2v8DdsqAJBTcjQywAAAAAElFTkSuQmCC"
-)
 
 
 class CSVFieldnames(Enum):
     PROJECT_NAME = "project_name"
     ROA_NAME = "roa_name"
-    ROA_COORDINATES_L = "roa_coordinates_l"
-    ROA_COORDINATES_T = "roa_coordinates_t"
-    ROA_COORDINATES_R = "roa_coordinates_r"
-    ROA_COORDINATES_B = "roa_coordinates_b"
+    ROA_COORDINATES_L = "roa_coordinates_l"  # xmin (left)
+    ROA_COORDINATES_T = "roa_coordinates_t"  # ymin (top)
+    ROA_COORDINATES_R = "roa_coordinates_r"  # xmax (right)
+    ROA_COORDINATES_B = "roa_coordinates_b"  # ymax (bottom)
     ROC_2_NAME = "roc_2_name"
-    ROC_2_COORDINATES_L = "roc_2_coordinates_l"
-    ROC_2_COORDINATES_T = "roc_2_coordinates_t"
-    ROC_2_COORDINATES_R = "roc_2_coordinates_r"
-    ROC_2_COORDINATES_B = "roc_2_coordinates_b"
+    ROC_2_COORDINATES_L = "roc_2_coordinates_l"  # xmin (left)
+    ROC_2_COORDINATES_T = "roc_2_coordinates_t"  # ymin (top)
+    ROC_2_COORDINATES_R = "roc_2_coordinates_r"  # xmax (right)
+    ROC_2_COORDINATES_B = "roc_2_coordinates_b"  # ymax (bottom)
     ROC_3_NAME = "roc_3_name"
-    ROC_3_COORDINATES_L = "roc_3_coordinates_l"
-    ROC_3_COORDINATES_T = "roc_3_coordinates_t"
-    ROC_3_COORDINATES_R = "roc_3_coordinates_r"
-    ROC_3_COORDINATES_B = "roc_3_coordinates_b"
+    ROC_3_COORDINATES_L = "roc_3_coordinates_l"  # xmin (left)
+    ROC_3_COORDINATES_T = "roc_3_coordinates_t"  # ymin (top)
+    ROC_3_COORDINATES_R = "roc_3_coordinates_r"  # xmax (right)
+    ROC_3_COORDINATES_B = "roc_3_coordinates_b"  # ymax (bottom)
 
     @staticmethod
-    def list():
-        return list(map(lambda c: c.value, CSVFieldnames))
+    def tolist() -> List[str]:
+        """CSVFieldnames enum class's list of values."""
+        return [c.value for c in CSVFieldnames]
 
 
 class ImportExportROAPlugin(Plugin):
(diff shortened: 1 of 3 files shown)


plugins/fastem_roa_import_export.py
```diff
@@ -0,0 +1,341 @@
...
+import csv
+from enum import Enum
+import logging
+import wx
+from wx.lib.embeddedimage import PyEmbeddedImage
+
+from odemis.gui.plugin import Plugin
+
+
+IMPORT_IMAGE = PyEmbeddedImage(
```
pieleric, line 33: The "source" of these images (ie, png) should be part of this PR too. You can add them to the `images` folder.

plugins/fastem_roa_import_export.py
```diff
@@ -0,0 +1,341 @@
...
+            "Import ROAs from a CSV file",
+            wildcard="CSV files (*.csv)|*.csv",
+            style=wx.FD_OPEN | wx.FD_FILE_MUST_EXIST,
+        ) as fileDialog:
+            if fileDialog.ShowModal() == wx.ID_CANCEL:
+                return  # the user changed their mind
+
+            # Proceed loading the file chosen by the user
+            pathname = fileDialog.GetPath()
+        try:
```
pieleric, line 220: Please separate the part that asks the user for a filename from the part that read the file, from the part that take the data and applies it to the GUI. So you should have 3 methods for the import and 3 methods for the export.
That will make writing test cases easier. It's also just easier to read.

plugins/fastem_roa_import_export.py
```diff
@@ -0,0 +1,341 @@
...
+            with open(pathname, newline="") as csvfile:
+                reader = csv.DictReader(csvfile)
+                data = {}
+                # Firstly extract the csv file data in a structured way
+                # {"project_name": [roa]} where roa is dict containing
+                # roa, roc_2, roc_3 name and its coordinates respectively
```
pieleric, line 226: Why using CSV format? It's a very limited format, so you end up having to use complex code to parse it (special Enum). I'd suggest to use the same format that we used to save the "CryoFeatures`: JSON. See odemis.acq.features. It's much more adapted to store the sort of structure which is used here.

Using JSON should also be much more future-proof in case the data changes. It'll be easier to detect the file uses an older format.

plugins/fastem_roa_import_export.py
```diff
@@ -0,0 +1,341 @@
...
+        self.addMenu("Help/Development/Import Export ROAs", self.import_export_roa)
+
+    def import_export_roa(self):
+        """Help/Development/Import Export ROAs callback."""
+        ImportExportTool().show(
+            project_list_controller=self._acquisition_tab._project_list_controller
+        )
+
+
+class ImportExportTool(object):
```
pieleric, line 100: That looks a very complex class, to just be initialized, called .show() once, and immediately be discarded.
I think you can move the `init()` to the plugin `__init__()`. Then, you can move the code of `show()` to `import_export_roa()`.

plugins/fastem_roa_import_export.py
```diff
@@ -0,0 +1,341 @@
...
+                logging.exception(ex)
+                wx.MessageBox(
+                    f"Exporting {pathname} failed, please see log.", "Error", wx.OK | wx.ICON_ERROR
+                )
+
+    def on_close(self, evt):
+        evt.Skip()
+        if not self:
+            return
+        if self.Parent:
```
pieleric, line 339: Can you explain what this does?

plugins/fastem_roa_import_export.py
```diff
@@ -0,0 +1,341 @@
...
+                            writer.writerow(data)
+            except Exception as ex:
+                logging.exception(ex)
+                wx.MessageBox(
+                    f"Exporting {pathname} failed, please see log.", "Error", wx.OK | wx.ICON_ERROR
+                )
+
+    def on_close(self, evt):
+        evt.Skip()
+        if not self:
```
pieleric, line 337: Can you explain what this does? I think this make sure that if the frame was already destroyed, it doesn't do anything anymore on it.

plugins/fastem_roa_import_export.py
```diff
@@ -0,0 +1,341 @@
...
+                            roc_3 = roa.roc_3.value
+                            data[CSVFieldnames.ROC_3_NAME.value] = roc_3.name.value
+                            roc_3_coordinates = roc_3.coordinates.value
+                            data[CSVFieldnames.ROC_3_COORDINATES_L.value] = roc_3_coordinates[0]
+                            data[CSVFieldnames.ROC_3_COORDINATES_T.value] = roc_3_coordinates[1]
+                            data[CSVFieldnames.ROC_3_COORDINATES_R.value] = roc_3_coordinates[2]
+                            data[CSVFieldnames.ROC_3_COORDINATES_B.value] = roc_3_coordinates[3]
+                            writer.writerow(data)
+            except Exception as ex:
+                logging.exception(ex)
```
pieleric, line 330: `logging.exception` already shows the exception, so no need to pass `ex`. You can add a text more like "Failure when exporting RoAs".

plugins/fastem_roa_import_export.py
```diff
@@ -0,0 +1,341 @@
...
+                            int(roc_2_name)
+                        ].coordinates.value = roc_2_coordinates
+                        project_ctrl.regions_calib_3.value[int(roc_3_name)].name.value = roc_3_name
+                        project_ctrl.regions_calib_3.value[
+                            int(roc_3_name)
+                        ].coordinates.value = roc_3_coordinates
+                        roa_ctrl = project_ctrl._on_btn_roa(None, name=roa_name)
+                        roa_ctrl.model.coordinates.value = roa_coordinates
+        except Exception as ex:
+            logging.exception(ex)
```
pieleric, line 277: No need to pass `ex` here.

plugins/fastem_roa_import_export.py
```diff
@@ -0,0 +1,341 @@
...
+            wx.MessageBox(
+                f"Importing {pathname} failed, please see log.", "Error", wx.OK | wx.ICON_ERROR
```
pieleric, line 279: You *could* pass "ex" here, so that the failure is less mysterious. Maybe it means nothing to the user... or maybe it already give them a clue. I don't think it can be too much information.

plugins/fastem_roa_import_export.py
```diff
@@ -0,0 +1,341 @@
...
+    This class is the frame that holds the Import Export ROAs tool. The toolbar is managed here.
+
+    :param project_list_controller: (FastEMProjectListController) class which creates/removes new
+            FastEM projects.
+
+    """
+
+    def __init__(self, title="Import Export ROAs Tool", project_list_controller=None, *args, **kw):
+        kw["title"] = title
+        wx.Frame.__init__(self, *args, **kw)
```
pieleric, line 180: Typically it's best to write `super().__init__(*args, **kw)` (that works better with copy-paste, and that handles multiple inheritance).

plugins/fastem_roa_import_export.py
```diff
@@ -0,0 +1,341 @@
...
+        # It only makes sense if the FASTEM acquisition tab is present
+        try:
+            self._acquisition_tab = main_app.main_data.getTabByName("fastem_acqui")
+        except LookupError:
+            logging.debug(
+                "Not loading Import Export ROAs tool since acquisition tab is not present"
+            )
+            return
+
+        self.addMenu("Help/Development/Import Export ROAs", self.import_export_roa)
```
pieleric, line 91: Is that really a user request to have an extra floating window to be able to press 2 buttons? Why not just having two menu entries "Import ROAs" & "Export ROAs"? About 50% of the code wouldn't be needed anymore...

pieleric (changes requested):

plugins/fastem_roa_import_export.py
tepals: The idea of using csv is that the user can easily copy and paste ROAs and adjust the position. A possible workflow is that the user draws 3 ROAs on a ribbon, then exports the data to csv, extrapolates the positions of these 3 ROAs to define the rest of the ribbon, imports it again and only has to move the new ROAs a bit.

pieleric: I see. If the choice of CSV was deliberate, and works better for the user, let's go for it!

plugins/fastem_roa_import_export.py
nandishjpatel: @tepals what's your opinion about the above comment?

tepals: I think Érics suggestion is good, it would be nice to have less code :)

src/odemis/gui/cont/project.py
```diff
@@ -178,20 +182,21 @@ def _on_btn_roa(self, _):
...
         self._current_roa_ctrl = roa_ctrl
 
         roa_ctrl.model.coordinates.subscribe(self._add_roa_ctrl)
+        return roa_ctrl
 
     # already running in main GUI thread as it receives event from GUI
     def _on_btn_remove(self, _, roa_ctrl):
         self.remove_roa_ctrl(roa_ctrl)
 
-    @call_in_wx_main  # call in main thread as changes in GUI are triggered
```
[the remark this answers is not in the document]
nandishjpatel: `_add_roa_ctrl` is only a one time subscription i.e. for the first coordinates value change it will be called. Running this in main thread makes it blocking, as in if I am importing 3 ROAs in the plugin loop, somehow only the third ROA panel will be created and drawn.

tepals (approved):

src/odemis/gui/cont/project.py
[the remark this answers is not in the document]
pieleric: Indeed, `call_in_wx_main` means the function is not immediately called. The call directly returns, but the actual function is put on a queue. That's why it's not blocking.

Thera is right, it's quite dangerous. At least, you should add a note in the docstring about the fact that this function should only be called from the main GUI thread. However, as it's a call back to a VA, that's hard to verify.

In addition the `_on_btn_roa()` is supposed to be called when the button is pressed. But now it's also called by your plugin when loading the RoA data. 

I think the proper way to deal with it would be to reorganise the code to have a proper separation between the GUI events, and the creation of a RoA. One approach could be to have a dedicated function to create a new RORController, and a second function to set the coordinates of a RoA. They would expect to run in the main GUI thread. Your plugin can call them. Separately, there would be two functions here to listen to the "Add ROA button" and to listen when the user has defined the coordinates on the first time. They would call each of the raw functions.

plugins/fastem_roa_import_export.py
```diff
@@ -0,0 +1,250 @@
...
+    ROC_3_COORDINATES_T = "roc_3_coordinates_t"  # ymin (top)
+    ROC_3_COORDINATES_R = "roc_3_coordinates_r"  # xmax (right)
+    ROC_3_COORDINATES_B = "roc_3_coordinates_b"  # ymax (bottom)
+
+    @staticmethod
+    def list():
+        return list(map(lambda c: c.value, CSVFieldnames))
+
+
+def get_csv_filepath_from_filedialog(parent, message: str, style: int) -> Union[str, None]:
```
pieleric, line 56: Why do you make it a separate function? I think it'd feel more logical/connected if it was just another method of the plugin. It could then use `self._parent` instead of receiving `parent` explicitly.

plugins/fastem_roa_import_export.py
```diff
@@ -0,0 +1,250 @@
...
+    ROC_2_COORDINATES_R = "roc_2_coordinates_r"  # xmax (right)
+    ROC_2_COORDINATES_B = "roc_2_coordinates_b"  # ymax (bottom)
+    ROC_3_NAME = "roc_3_name"
+    ROC_3_COORDINATES_L = "roc_3_coordinates_l"  # xmin (left)
+    ROC_3_COORDINATES_T = "roc_3_coordinates_t"  # ymin (top)
+    ROC_3_COORDINATES_R = "roc_3_coordinates_r"  # xmax (right)
+    ROC_3_COORDINATES_B = "roc_3_coordinates_b"  # ymax (bottom)
+
+    @staticmethod
+    def list():
```
pieleric, line 52: Let's use the same name as in numpy: `tolist()`. This especially clarifies that does some sort of conversion.

Please also indicate the (return) type in the signature and add a docstring.

pieleric (changes requested):

src/odemis/gui/cont/project.py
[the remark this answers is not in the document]
nandishjpatel: Thanks Éric for the comment, the restructuring/separation seems to me some work. I have a applied a quick workaround for now, let me know if that will suffice. The change is in `plugins/fastem_roa_import_export.py` from line 164. While trying a few things I found that we cannot have a VA subscribed to `functools.partial`, is that corrrect?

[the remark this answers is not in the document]
pieleric: Yes, it's better, as it doesn't drop the protection.
> I found that we cannot have a VA subscribed to functools.partial, is that corrrect?

It's "partially" correct (punt intended). You can pass a partial function to subscribe(). However, when you subscribe to a VA, it uses a weak reference. This means that if on the caller side you don't hold a reference to it anymore, the function you pass to the VA will be free'd (like any other Python object). In this case, the VA will not be able to call it. Typically, this is what happens with a partial (or a lambda function) at the end of the scope (ie, the method). To work around it, you'll need to keep a reference to the function you pass to the VA. Usually, we just add this partial function to a private list on the caller class (something like: `self._subscribers.append(my_func)` ).

pieleric (approved):

src/odemis/gui/cont/project.py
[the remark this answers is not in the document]
nandishjpatel: I managed to used `functools.partial` and make it a bit cleaner. Thanks Éric!

src/odemis/gui/cont/project.py
```diff
@@ -230,8 +235,9 @@ def remove_roa_ctrl(self, roa_ctrl):
         roa_ctrl.panel.Destroy()
         self.panel.fit_panels()
 
-        # Remove controller from .roa_ctrl list
-        self.roa_ctrls = {key: val for key, val in self.roa_ctrls.items() if val != roa_ctrl}
+        # Remove roa controller and its subscriber callback
+        self.roa_ctrls.pop(roa_ctrl)
```
pieleric, line 239: If you don't care about the value associated, and are certain the key is present, you can just write `del self.roa_ctrls[roa_ctrl]`.
If you're not certain it is still present, then you should do `.pop(..., None)`.

pieleric (changes requested):

pieleric (approved):

PR #2608, opened 2023-12-19
[TRCL-55] add folding of the streams

FCanberk:
Add folding of the CL intensity and EBIC streams for SPARC if the dwell time, repetition, roi values are the same and the detectors are different

Add an info text on the GUI as well that lets the user know the acquisitions will be folded(merged)

Add test cases with different dwell time, repetition, roi combinations for the folding

Lastly, fix small PEP-8 incompatibilities

### DIFF
--- a/src/odemis/acq/acqmng.py
+++ b/src/odemis/acq/acqmng.py
@@ -233,7 +233,8 @@ class ZStackAcquisitionTask(object):
                     except CancelledError:
                         raise
                     except Exception as e:
-                        logging.exception("The acquisition failed at the %s-th zlevel of the stream %s, because %s" %(i+1, stream, e))
+                        logging.exception("The acquisition failed at the %s-th zlevel of the stream %s, because %s" % (
+                        i + 1, stream, e))
                         # TODO handle zstack assembling in case of error 
                         return acquired_data, e
                         
@@ -273,7 +274,7 @@ def estimateZStackAcquisitionTime(streams, zlevels):
         zs = zlevels.get(s, [0])  # concider that streams without zlevels are acquired at a single z
         if len(zs) > 1 and s.focuser:
             zstep = abs(zs[0] - zs[1])
-            tot_zstep_time = (len(zs) - 1) *  guessActuatorMoveDuration(s.focuser, axis="z", distance=zstep)
+            tot_zstep_time = (len(zs) - 1) * guessActuatorMoveDuration(s.focuser, axis="z", distance=zstep)
             acq_time += tot_zstep_time
 
     return acq_time
@@ -305,8 +306,8 @@ def foldStreams(streams, reuse=None):
     reuse (list of Streams, or None): list of streams which was previously output
       by this function. If it's present, the streams will be reused when possible,
       for optimisation.
-    return (list of Streams): The list of streams, with the ones that can be
-      folded replaced by a MD streams, the other streams are pass as-is.
+    return (set of Streams): The set of streams, with the ones that can be
+      folded replaced by an MD streams, the other streams are pass as-is.
     """
     # TODO: support SPARC streams
     if reuse is None:
@@ -317,6 +318,7 @@ def foldStreams(streams, reuse=None):
 
     folds = set()
     scan_fluos = []  # List of sets of ScannedFluoStream with compatible settings
+    scan_semmds = []  # List of sets of SEMMDStream with compatible settings
     for s in streams:
         if isinstance(s, ScannedFluoStream):
             # Store it for folding
@@ -337,6 +339,24 @@ def foldStreams(streams, reuse=None):
         elif isinstance(s, ScannedRemoteTCStream):
             # Don't add extra FLIM streams
             continue
+        elif isinstance(s, SEMMDStream):
+            # Fold streams if roi, repetition and dwellTime are the same, but they all have different detectors
+            # (except for the e-beam)
+            for smds in scan_semmds:
+                compatible = True
+                for smd in smds:
+                    if not (smd._dwellTime.value == s._dwellTime.value and
+                            smd.repetition.value == s.repetition.value and
+                            smd.roi.value == s.roi.value and
+                            # as of now, each MDStream has only 2 streams, and the first one is always the SEM stream.
+                            smd.streams[1].detector.name != s.streams[1].detector.name
+                    ):
+                        compatible = False
+                if compatible:
+                    smds.add(s)
+                    break
+            else:  # No existing group of SEMMDStreams compatible => create a new one
+                scan_semmds.append({s})
         else:
             folds.add(s)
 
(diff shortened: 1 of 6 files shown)


src/odemis/acq/acqmng.py
```diff
@@ -337,6 +339,26 @@ def foldStreams(streams, reuse=None):
...
+                compatible = True
+                for smd in smds:
+                    # "compatible" means: same emitter/scanner/excitation
+                    if (smd._dwellTime.value == s._dwellTime.value and
+                            smd.repetition.value == s.repetition.value and
+                            smd.roi.value == s.roi.value and
+                            # as of now, each MDStream has only 2 streams
+                            smd.streams[1].detector.name != s.streams[1].detector.name
+                    ):
+                        compatible = True
```
pieleric, line 354: What this loop does is to look at a group of SEMMDStreams that are going to be folded, and see if it's possible to also add this (`smd`) Stream to the group. For this to be possible it should be compatible with all the other streams. 
So you shouldn't set `compatible` to True as soon as one stream is compatible. You should only do the `else` part (set to False). => Invert the `if` with a `not`.

src/odemis/gui/cont/acquisition.py
```diff
@@ -1395,10 +1395,15 @@ def update_acquisition_time(self):
             lvl = logging.WARN
         else:
             streams = self._tab_data_model.acquisitionStreams
-            acq_time = acqmng.estimateTime(streams)
-            acq_time = math.ceil(acq_time)  # round a bit pessimistic
-            txt = u"Estimated time is {}."
-            txt = txt.format(units.readable_time(acq_time))
+            folds = tuple(acqmng.foldStreams(self._tab_data_model.acquisitionStreams))
+            if len(streams) > len(folds):
+                txt = u"EBIC and CL streams acquired simultaneously"
```
pieleric, line 1400: That's true now, but if there are ever different detectors added, this might become incorrect. I think either you detect the name of the streams being folded (which is quite hard) or just keep it generic like "Some streams will be acquired simultaneously".

src/odemis/gui/cont/acquisition.py
```diff
@@ -1395,10 +1395,15 @@ def update_acquisition_time(self):
...
+                lvl = logging.INFO
+            else:
+                acq_time = acqmng.estimateTime(streams)
+                acq_time = math.ceil(acq_time)  # round a bit pessimistic
+                txt = u"Estimated time is {}."
+                txt = txt.format(units.readable_time(acq_time))
```
pieleric, line 1406: The main issue with this approach is that now if some streams are folded, it's not possible to see the estimated time. I guess the nicest way would be to have two separate texts: the info/warning message, and the time estimation. One above the other one, so that they can be both seen. If the info/warning message is empty, then the widget is not shown, to not take any space (and so the panel takes less space). Could you have a look at this?
Or do you have any idea for handling this differently?

src/odemis/gui/cont/acquisition.py
```diff
@@ -1500,6 +1505,13 @@ def on_acquisition(self, evt):
         self._show_status_icons(None)
         self._tab_panel.Layout()  # to put the gauge at the right place
 
+        # fold the streams if possible
+        folds = tuple(acqmng.foldStreams(self._tab_data_model.acquisitionStreams))
+
+        if len(self._tab_data_model.acquisitionStreams) > len(folds):
+            self.lbl_acqestimate.SetLabel("EBIC and CL streams acquired simultaneously")
+            self._tab_data_model.acquisitionStreams = folds
```
pieleric, line 1513: Isn't this going to cause issues afterwards when deleting the streams, or acquiring new streams? 
I think it's simpler to not change `.acquisitionStreams` pass `folds` to `acqmng.acquire()`

pieleric (changes requested):

src/odemis/acq/test/acq_test.py
```diff
@@ -564,6 +565,129 @@ def on_progress_update(self, future, start, end):
...
+        self.assertEqual(len(folds[0].streams), 2)
+        self.assertEqual(len(folds[1].streams), 2)
+
+    def test_folding_pairs_of_two(self):
+        """
+        Test folding of the streams that have the same repetition and dwell time values but different
+        roi and detector names
+        """
+        # Create the streams
+        sems = stream.SEMStream("test sem", self.sed, self.sed.data, self.ebeam,
```
K4rishma, line 662: Would it be nice to add the stream initialisation in the init as it is same for all the test cases or do we need flexibility to change them for future test cases?

FCanberk: I don't think the sem streams would change in the future so moving it to the init is indeed valid. But my intention was to keep all the streams next to each other and felt the code is more readable that way.

src/odemis/gui/cont/acquisition.py
```diff
@@ -1500,8 +1509,19 @@ def on_acquisition(self, evt):
...
+        folds = tuple(acqmng.foldStreams(self._tab_data_model.acquisitionStreams))
+
+        if len(self._tab_data_model.acquisitionStreams) > len(folds):
+            self.lbl_acqestimate.SetLabel("EBIC and CL streams acquired simultaneously")
+            # start acquisition
+            self.acq_future = acqmng.acquire(folds, self._main_data_model.settings_obs)
+        else:
+            # start acquisition
+            self.acq_future = acqmng.acquire(self._tab_data_model.acquisitionStreams,
+                                             self._main_data_model.settings_obs)
```
pieleric, line 1522: That's not what I meant. You can always use `acquire(folds)`, because either folds is the same as `acquisitionStreams`, or it's the right content.
```suggestion
        if len(self._tab_data_model.acquisitionStreams) > len(folds):
            self.lbl_acqestimate.SetLabel("EBIC and CL streams acquired simultaneously")
        # start acquisition
        self.acq_future = acqmng.acquire(folds, self._main_data_model.settings_obs)
```

src/odemis/gui/cont/acquisition.py
```diff
@@ -1395,6 +1397,13 @@ def update_acquisition_time(self):
             lvl = logging.WARN
         else:
             streams = self._tab_data_model.acquisitionStreams
+            folds = tuple(acqmng.foldStreams(self._tab_data_model.acquisitionStreams))
+            if len(streams) > len(folds):
+                self.lbl_fold_acq.Show()
+                self.bmp_fold_acq_info.Show()
+            else:
+                self.lbl_fold_acq.Hide()
+                self.bmp_fold_acq_info.Hide()
```
pieleric, line 1406: You can write it this way:
```suggestion
            has_folds = len(streams) >  len(acqmng.foldStreams(streams))
            self.lbl_fold_acq.Show(has_folds)
            self.bmp_fold_acq_info.Show(has_folds)
```

src/odemis/acq/test/acq_test.py
```diff
@@ -564,6 +565,129 @@ def on_progress_update(self, future, start, end):
...
+        cls.repetition.value = (1024, 1024)
+
+        ebics.roi.value = (0, 0, 1, 1)
+        ebics.repetition.value = (1024, 1024)
+
+        folds = acqmng.foldStreams([sms_ebic, sms_cl])
+        folds = tuple(folds)
+        str_cnt = len(folds)
+
+        self.assertEqual(str_cnt, 1)
```
pieleric, line 619: `str_cnt` sounds like "string" so it's a confusing name. I think you can avoid this variable totally though:
```suggestion
        self.assertEqual(str_cnt, len(folds))
```

pieleric (changes requested):

src/odemis/gui/cont/acquisition.py
```diff
@@ -1500,8 +1507,15 @@ def on_acquisition(self, evt):
...
         self._tab_panel.Layout()  # to put the gauge at the right place
 
-        # start acquisition + connect events to callback
-        self.acq_future = acqmng.acquire(self._tab_data_model.acquisitionStreams, self._main_data_model.settings_obs)
+        # fold the streams if possible
+        folds = tuple(acqmng.foldStreams(self._tab_data_model.acquisitionStreams))
+
+        if len(self._tab_data_model.acquisitionStreams) > len(folds):
+            self.lbl_acqestimate.SetLabel("EBIC and CL streams acquired simultaneously")
+            # start acquisition
```
pieleric, line 1515: This should have one indentation less.

pieleric (changes requested):
One very tiny small remark, other everything looks fine. Almost ready to merge!

pieleric (approved):

PR #2609, opened 2023-12-21
[METEOR-881] Add test case for multiple images from multiple manufacturers

patrickcleeve2:
This pull request adds:

- A writer for writing image metadata to test case format.
- A test case for testing the parsers for multiple manufacturers on multiple images

### DIFF
--- a/src/odemis/dataio/test/generate_test_metadata.py
+++ b/src/odemis/dataio/test/generate_test_metadata.py
@@ -1,0 +1,94 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+'''
+Created on 23 Jan 2024
+
+@author: Patrick Cleeve
+
+Copyright © 2024, Delmic
+
+This file is part of Odemis.
+
+Odemis is free software: you can redistribute it and/or modify it under the terms 
+of the GNU General Public License version 2 as published by the Free Software 
+Foundation.
+
+Odemis is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; 
+without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR 
+PURPOSE. See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with 
+Odemis. If not, see http://www.gnu.org/licenses/.
+'''
+import argparse
+import datetime
+import glob
+import json
+import logging 
+import os
+
+from odemis import model
+from odemis.util.dataio import open_acquisition
+
+logging.basicConfig(level=logging.DEBUG)
+
+
+"""The test case in multi_test_tiff.py requires additional test data to be downloaded from Drive/Software Engineering/Test data/metor/test_images
+and the associated metadata to be generated using this generate_test_metadata.py script.
+
+To generate the test metadata, run the following command from the root of the repository:
+
+python -m odemis.dataio.test.generate_test_metadata --path <path to test data>
+
+This will generate a test_metadata.json file in the test data directory. This file is used by the test case in multi_tiff_test.
+"""
+
+def write_test_metadata(path: str):
+    """Write test metadata to json file.
+    Use this function to generate test metadata in the required from a directory of *.tif images.
+    It will write the test_metadata.json file to the same directory as the test data.
+    :param path: (str) path to test data (directory of *.tif images)"""
+
+    filenames = glob.glob(os.path.join(path, "**/*.tif*"), recursive=True) 
+    test_metadata = {}
+
+    for fname in filenames:
+
+        try:
+            data = open_acquisition(fname)[0].getData()
+
+            # format metadata for tests
+            tmd = data.metadata
+            tmd["filename"] = os.path.basename(fname).replace(".tif", "")
+            tmd["shape"] = data.shape
+            tmd["dtype"] = str(data.dtype)
+            tmd["length"] = 1
+
+            # convert timestamp into datetime
+            ts = data.metadata[model.MD_ACQ_DATE]
+            tmd["timestamp"] = ts
+            tmd["datetime"] = datetime.datetime.fromtimestamp(ts).strftime('%Y-%m-%d %H:%M:%S')
+
+            test_metadata[tmd["filename"]] = tmd
+
+        except Exception as e:
+            logging.warning(f"Could not parse {fname}. Exception: {e}")
+
+    # save test_metadata as json
+    with open(os.path.join(path, "test_metadata.json"), "w") as f:
+        json.dump(test_metadata, f)
+    
+    return test_metadata
+
+def main():
+
+    parser = argparse.ArgumentParser(description='Generate test metadata from a directory of *.tif images.')
+    parser.add_argument('--path', type=str, help='Path to test data (directory of *.tif images).')
+    args = parser.parse_args()
+
+    # write test metadata
+    write_test_metadata(args.path)
+
+
+if __name__ == '__main__':
+    main()
--- a/src/odemis/dataio/test/multi_tiff_test.py
+++ b/src/odemis/dataio/test/multi_tiff_test.py
@@ -1,0 +1,92 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+'''
+Created on 23 Jan 2024
+
+@author: Patrick Cleeve
+
+Copyright © 2024, Delmic
+
+This file is part of Odemis.
+
+Odemis is free software: you can redistribute it and/or modify it under the terms 
+of the GNU General Public License version 2 as published by the Free Software 
+Foundation.
+
+Odemis is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; 
+without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR 
+PURPOSE. See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with 
+Odemis. If not, see http://www.gnu.org/licenses/.
+'''
+import glob
+import json
+import os
+import unittest
+
+import numpy
+
+from odemis import model
+from odemis.dataio import tiff
+
+
+"""This test case requires additional test data to be downloaded from Drive/Software Engineering/Test data/metor/test_images 
+and the associated metadata to be generated using the generate_test_metadata.py script.
+
+To generate the test metadata, run the following command from the root of the repository:
+
+python -m odemis.dataio.test.generate_test_metadata --path ~/test_data/meteor/test_images
+
+This will generate a test_metadata.json file in the test data directory. This file is used by the test case below. 
+You will need to re-run this script if you add new test data.
+
+"""
+
+class TestMultiTiffIO(unittest.TestCase):
+
+  def test_manufacturer_metadata_parsers(self):
+    """Test open_data function with multiple manufacturers and metadata parsers."""
+
+    # test data path (only available on hosted runner atm)
+    # To setup or update: copy data from Drive/Software Engineering/Test data/metor/test_images to home directory
+    TEST_PATH = os.path.join(os.path.expanduser('~'), "test_data/meteor/test_images")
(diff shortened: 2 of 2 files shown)


FCanberk:
we typically end the name of the test files with `_test` so the file name in this case would be `multi_tiff_eval_test.py`

src/odemis/dataio/test/multi_tiff_eval.py
```diff
@@ -0,0 +1,94 @@
...
+import datetime
+import glob
+import logging
+import os
+import json
+import unittest
+import numpy 
+from odemis import model
+from odemis.dataio import tiff
+from odemis.util.dataio import open_acquisition
```
FCanberk, line 11:
```suggestion
import datetime
import glob
import json
import logging 
import os
import unittest

import numpy

from odemis import model
from odemis.dataio import tiff
from odemis.util.dataio import open_acquisition
```

src/odemis/dataio/test/multi_tiff_eval.py
```diff
@@ -0,0 +1,94 @@
...
+    return test_metadata
+
+
+class TestTiffIO(unittest.TestCase):
+
+  def test_manufacturer_metadata_parsers(self):
+    """Test open_data function with multiple manufacturers and metadata parsers."""
+
+    # TEST_PATH = "../data/TEST/TEST-IMAGES/" # TODO: update
+    TEST_PATH = "/home/patrick/development/data/TEST/TEST-IMAGES"
```
FCanberk, line 57: the TEST_PATH should be generic instead of /home/**patrick**/...

FCanberk (changes requested):

tepals:
> we typically end the name of the test files with `_test` so the file name in this case would be `multi_tiff_eval_test.py`

This is important to do, because we only run test modules that end in `_test.py`.

patrickcleeve2:
I've moved the data onto the ci machine, will update here once it is running successfully

src/odemis/dataio/test/multi_tiff_test.py
```diff
@@ -0,0 +1,102 @@
...
+
+import numpy
+
+from odemis import model
+from odemis.dataio import tiff
+from odemis.util.dataio import open_acquisition
+
+logging.basicConfig(level=logging.DEBUG)
+
+# write test metadata to file
```
tepals, line 17: this can be the docstring

src/odemis/dataio/test/multi_tiff_test.py
```diff
@@ -0,0 +1,102 @@
...
+            tmd["timestamp"] = ts
+            tmd["datetime"] = datetime.datetime.fromtimestamp(ts).strftime('%Y-%m-%d %H:%M:%S')
+
+            test_metadata[tmd["filename"]] = tmd
+
+        except:
+            continue # ignore files that cannot be read for now
+
+    # save test_metadata as json
+    import json
```
tepals, line 46: I don't think it's necessary to do `import json` here, it's already at the top.

src/odemis/dataio/test/multi_tiff_test.py
```diff
@@ -0,0 +1,102 @@
...
+
+            # convert timestamp into datetime
+            ts = data.metadata[model.MD_ACQ_DATE]
+            tmd["timestamp"] = ts
+            tmd["datetime"] = datetime.datetime.fromtimestamp(ts).strftime('%Y-%m-%d %H:%M:%S')
+
+            test_metadata[tmd["filename"]] = tmd
+
+        except:
+            continue # ignore files that cannot be read for now
```
tepals, line 43: Does it make sense to log the files that cannot be read for now?

src/odemis/dataio/test/multi_tiff_test.py
```diff
@@ -0,0 +1,102 @@
...
+    return test_metadata
+
+
+class TestMultiTiffIO(unittest.TestCase):
+
+  def test_manufacturer_metadata_parsers(self):
+    """Test open_data function with multiple manufacturers and metadata parsers."""
+
+    # test data path (only available on hosted runner atm)
+    TEST_PATH = os.path.join(f"{os.path.expanduser('~')}", "test_data/meteor/test_images")
```
tepals, line 59: `os.path.expanduser()` returns a string, so there is no need to make a string out of it

```suggestion
    TEST_PATH = os.path.join(os.path.expanduser('~'), "test_data/meteor/test_images")
```

src/odemis/dataio/test/multi_tiff_test.py
```diff
@@ -0,0 +1,102 @@
...
+    IGNORED_KEYS = ["shape", "dtype", "length", "filename", "timestamp", "datetime"]
+
+    # if test path doesnt exist, we are on github actions, just ignore for now
+    # TODO: download test data for github actions
+    if not os.path.exists(TEST_PATH):
+        logging.warning(f"Test data not found at {TEST_PATH}. Skipping test.")
+        return
```
tepals, line 66: You can also skip the test, then it will show up as skipped in the report, instead of as passed.
```suggestion
        raise unittest.SkipTest(f"Test data not found at {TEST_PATH}. Skipping test.")
```

src/odemis/dataio/test/multi_tiff_test.py
```diff
@@ -0,0 +1,102 @@
...
+            if key in [model.MD_PIXEL_SIZE, model.MD_POS]: # array equal
+                numpy.testing.assert_array_equal(value, data.content[0].metadata[key])
+            elif key in [model.MD_DWELL_TIME]:
+                numpy.testing.assert_almost_equal(value, data.content[0].metadata[key])
+            else:
+                self.assertEqual(value, data.content[0].metadata[key])
+
+
+if __name__ == '__main__':
+    unittest.main()
```
tepals, line 102: can you add an empty line at the end of the file?

src/odemis/dataio/test/multi_tiff_test.py
```diff
@@ -0,0 +1,102 @@
...
+
+        for key, value in md.items():
+            if key in IGNORED_KEYS:
+                continue # skip
+            if key in [model.MD_PIXEL_SIZE, model.MD_POS]: # array equal
+                numpy.testing.assert_array_equal(value, data.content[0].metadata[key])
+            elif key in [model.MD_DWELL_TIME]:
+                numpy.testing.assert_almost_equal(value, data.content[0].metadata[key])
+            else:
+                self.assertEqual(value, data.content[0].metadata[key])
```
tepals, line 98: I think you can combine the two loops

```suggestion
        for key, value in md.items():
            if key in IGNORED_KEYS:
                continue # skip
            self.assertIn(key, data.content[0].metadata)
            if key in [model.MD_PIXEL_SIZE, model.MD_POS]: # array equal
                numpy.testing.assert_array_equal(value, data.content[0].metadata[key])
            elif key in [model.MD_DWELL_TIME]:
                numpy.testing.assert_almost_equal(value, data.content[0].metadata[key])
            else:
                self.assertEqual(value, data.content[0].metadata[key])
```

tepals (changes requested):

FCanberk (approved):

src/odemis/dataio/test/multi_tiff_test.py
```diff
@@ -0,0 +1,98 @@
...
+from odemis.dataio import tiff
+from odemis.util.dataio import open_acquisition
+
+logging.basicConfig(level=logging.DEBUG)
+
+def write_test_metadata(path: str):
+    """Write test metadata to json file.
+    :param path: (str) path to test data (directory of *.tif images)"""
+
+    filenames =  glob.glob(os.path.join(path, "**/*.tif*"), recursive=True)
```
pieleric, line 21:
```suggestion
    filenames = glob.glob(os.path.join(path, "**/*.tif*"), recursive=True) 
```

src/odemis/dataio/test/multi_tiff_test.py
```diff
@@ -0,0 +1,98 @@
...
+    return test_metadata
+
+
+class TestMultiTiffIO(unittest.TestCase):
+
+  def test_manufacturer_metadata_parsers(self):
+    """Test open_data function with multiple manufacturers and metadata parsers."""
+
+    # test data path (only available on hosted runner atm)
+    TEST_PATH = os.path.join(os.path.expanduser('~'), "test_data/meteor/test_images")
```
pieleric, line 59: I'm confused. Where is this directory? In Google Drive, the closest I can see is: Software Engineering/Test data/meteor-examples/

patrickcleeve2: It's on the self hosted server

src/odemis/dataio/test/multi_tiff_test.py
```diff
@@ -0,0 +1,98 @@
...
+
+import numpy
+
+from odemis import model
+from odemis.dataio import tiff
+from odemis.util.dataio import open_acquisition
+
+logging.basicConfig(level=logging.DEBUG)
+
+def write_test_metadata(path: str):
```
pieleric, line 17: This function doesn't seem used. Can you add a comment why it's useful... or delete that function?

pieleric (changes requested):

src/odemis/dataio/test/multi_tiff_test.py
```diff
@@ -0,0 +1,101 @@
...
+from odemis import model
+from odemis.dataio import tiff
+from odemis.util.dataio import open_acquisition
+
+logging.basicConfig(level=logging.DEBUG)
+
+
+def write_test_metadata(path: str):
+    """Write test metadata to json file.
+    Use this function to generate test metadata in the required from a directory of *.tif images.
```
pieleric, line 20: So, if I understand correctly, you'd need to call that function again every time you add new data to the test directory? How would you call it?
Could you add a comment at the top of this file saying something like:
```
This test file requires extra data. To install it do ....
If you want to extend the data, do ....
```

src/odemis/dataio/test/multi_tiff_test.py
```diff
@@ -0,0 +1,101 @@
+
```
pieleric, line 1: Please add the standard boilerplate for the test cases. It should have the shebang (as it's supposed to be executable by itself) + the copyright notice.

pieleric (changes requested):

src/odemis/dataio/test/multi_tiff_test.py
patrickcleeve2: I've moved the test metadata generation to a separate script, and added instructions for running it.

src/odemis/dataio/test/multi_tiff_test.py
```diff
@@ -0,0 +1,92 @@
...
+from odemis import model
+from odemis.dataio import tiff
+
+
+"""This test case requires additional test data to be downloaded from Drive/Software Engineering/Test data/metor/test_images 
+and the associated metadata to be generated using the generate_test_metadata.py script.
+
+To generate the test metadata, run the following command from the root of the repository:
+
+python -m odemis.dataio.test.generate_test_metadata --path <path to test data>
```
pieleric, line 39: That `<path to test data>` is hard-coded to `~/test_data/meteor/test_images`, so let's explicitly write that.

src/odemis/dataio/test/generate_test_metadata.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+"""The test case in multi_test_tiff.py requires additional test data to be downloaded from Drive/Software Engineering/Test data/metor/test_images
+and the associated metadata to be generated using this generate_test_metadata.py script.
+
+To generate the test metadata, run the following command from the root of the repository:
+
+python -m odemis.dataio.test.generate_test_metadata --path <path to test data>
+
+This will generate a test_metadata.json file in the test data directory. This file is by the test case in multi_tiff_test.
```
pieleric, line 43:
```suggestion
This will generate a test_metadata.json file in the test data directory. This file is used by the test case in multi_tiff_test.
```

pieleric (changes requested):

pieleric (approved):

PR #2612, opened 2024-01-03
[SDEV-1767] disable failure of roa acquisition when beamshift correction fails

nandishjpatel:
[fix] catch the raised ValueError when setting beamshift VA value goes out of range and instead log the error for now

By doing so the ROAs acquisition can still continue

### DIFF
--- a/src/odemis/acq/fastem.py
+++ b/src/odemis/acq/fastem.py
@@ -680,7 +680,7 @@ class AcquisitionTask(object):
             try:
                 self.correct_beam_shift()  # correct the shift of the beams caused by the parasitic magnetic field.
             except Exception:
-                logging.exception("Correcting the beam shift failed.")
+                logging.exception("Correcting the beam shift failed, check if the image quality is still good.")
                 # In case of failure save the ccd image
                 ccd_image = self._ccd.data.get(asap=False)
                 fastem_util.save_image(self.path, f"{self.field_idx}_after.tiff", ccd_image)
--- a/src/odemis/acq/fastem.py
+++ b/src/odemis/acq/fastem.py
@@ -674,7 +674,16 @@ class AcquisitionTask(object):
 
             self.move_stage_to_next_tile()  # move stage to next field image position
             self._scanner.blanker.value = False  # unblank the beam
-            self.correct_beam_shift()  # correct the shift of the beams caused by the parasitic magnetic field.
+            # TODO remove the below temporary fix when a proper solution is found
+            # The temporary fix has been applied so that the ROAs acquisition can still continue upon failure
+            # of correcting the beam shift
+            try:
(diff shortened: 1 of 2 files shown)


src/odemis/acq/fastem.py
```diff
@@ -875,7 +875,13 @@ def correct_beam_shift(self):
 
         cur_beam_shift_pos = numpy.array(self._beamshift.shift.value)
         logging.debug("Current beam shift: {} [m]".format(self._beamshift.shift.value))
-        self._beamshift.shift.value = (cur_beam_shift_pos + beam_shift_cor)
+        # TODO remove the below temporary fix when a proper solution is found
```
tepals, line 878: the acquisition will also fail when there is no pattern, then on line 846 an error will be raised. Could you also add a catch for that?

src/odemis/acq/fastem.py
```diff
@@ -875,7 +875,13 @@ def correct_beam_shift(self):
...
+        # The temporary fix has been applied so that the ROAs acquisition can still continue upon failure
+        # of setting beamshift VA value
+        try:
+            self._beamshift.shift.value = (cur_beam_shift_pos + beam_shift_cor)
+        except ValueError:
+            logging.exception("Setting beam shift value failed.")
```
tepals, line 884: can you also log the value it tries to set?

K4rishma, line 884:
```suggestion
            logging.exception("Setting beam shift value %s failed", cur_beam_shift_pos + beam_shift_cor )
```

K4rishma (approved):

src/odemis/acq/fastem.py
nandishjpatel: Should the try statement be on line 677 then? instead of two inside the function

tepals: I think it is good to know the specific case for which it failed, so you can keep it inside the function. Or try to catch both independently on line 677.

src/odemis/acq/fastem.py
```diff
@@ -674,7 +674,16 @@ def acquire_roa(self, dataflow):
...
             self.move_stage_to_next_tile()  # move stage to next field image position
             self._scanner.blanker.value = False  # unblank the beam
-            self.correct_beam_shift()  # correct the shift of the beams caused by the parasitic magnetic field.
+            # TODO remove the below temporary fix when a proper solution is found
+            # The temporary fix has been applied so that the ROAs acquisition can still continue upon failure
+            # of correcting the beam shift
+            try:
+                self.correct_beam_shift()  # correct the shift of the beams caused by the parasitic magnetic field.
+            except Exception:
+                logging.exception("Correcting the beam shift failed.")
```
tepals, line 683:
```suggestion
                logging.exception("Correcting the beam shift failed, check if the image quality is still good.")
```

tepals (approved):

PR #2613, opened 2024-01-03
[SDEV-1692] Fix and refactor registrar classes

tepals:
[fix] ShiftRegistrar._compute_registration a position in pixels was compared to a position in meters
    [refactor] Renaming of registered_positions to registered_positions_m and addition of registered_positions_px
    [refactor] Let GlobalShiftRegistrar inherit from the ShiftRegistrar class.
    [fix] shift registrar synthetic test case

### DIFF
--- a/src/odemis/acq/stitching/_registrar.py
+++ b/src/odemis/acq/stitching/_registrar.py
@@ -140,7 +140,7 @@ class ShiftRegistrar(object):
     def getPositions(self):
         """
         returns:
-        tile_positions (list of N tuples): the adjusted position in X/Y for each tile, in the order they were added
+        tile_positions (list of N tuples): the adjusted position in X/Y for each tile, in the order they were added [m]
         dep_tile_positions (list of N tuples of K tuples of 2 floats): for each tile, it returns
         the adjusted position of each dependent tile (in the order they were passed)
         """
--- a/src/odemis/acq/stitching/_registrar.py
+++ b/src/odemis/acq/stitching/_registrar.py
@@ -28,14 +28,16 @@ content, in other words performing "image registration".
 
 """
 
-from odemis.acq.drift import MeasureShift
-import numpy
-import math
-from odemis import model
 import logging
+import math
+from collections import deque
+
+import numpy
 from scipy.sparse import csr_matrix
 from scipy.sparse.csgraph import minimum_spanning_tree
-from collections import deque
+
+from odemis import model
+from odemis.acq.drift import MeasureShift
 
 GOOD_MATCH = 0.9  # consider all registrations with match > GOOD_MATCH
 LEFT_TO_RIGHT = 1
@@ -97,16 +99,16 @@ class ShiftRegistrar(object):
         self.ovrlp = 0
 
         # List of 2D indices for grid positions in order of acquisition
-        self.acqOrder = []
+        self.acq_order = []
 
         # Shift between main tile and dependent tiles
         self.shift_tile_dep_tiles = []  # N x K list. N: number of tiles, K: number of dep_tiles.
 
         self.px_size = None  # Tuple of 2 ints. (X,Y) pixel size of the current tile in m/px.
         self.size = None  # Tuple of 2 ints. (Y,X) shape of the current tile in px.
-        self.posX = None  # int. Position of the current tile in the grid. (0,0) is the top left, posX
+        self.pos_x = None  # int. Position of the current tile in the grid. (0,0) is the top left, pos_x
         # increases when moving to the right
-        self.posY = None  # int. Y position, increases when moving down.
+        self.pos_y = None  # int. Y position, increases when moving down.
         self.pos_prev_x = None  # int. X coordinate of previous tile. Used to determine whether
         # the current tile should be compared to the right or to the left tile.
         self.osize = None  # int. Overlap size in pixels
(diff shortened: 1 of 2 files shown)


patrickcleeve2 (approved):

src/odemis/acq/stitching/_registrar.py
```diff
@@ -96,16 +99,16 @@ def __init__(self):
...
-        self.acqOrder = []
+        self.acq_order = []
 
         # Shift between main tile and dependent tiles
         self.shift_tile_dep_tiles = []  # N x K list. N: number of tiles, K: number of dep_tiles.
 
         self.px_size = None  # Tuple of 2 ints. (X,Y) pixel size of the current tile in m/px.
         self.size = None  # Tuple of 2 ints. (Y,X) shape of the current tile in px.
-        self.posX = None  # int. Position of the current tile in the grid. (0,0) is the top left, posX
+        self.pos_x = None  # int. Position of the current tile in the grid. (0,0) is the top left, posX
```
nandishjpatel, line 109:
```suggestion
        self.pos_x = None  # int. Position of the current tile in the grid. (0,0) is the top left, pos_x
```

src/odemis/acq/stitching/_registrar.py
```diff
@@ -124,100 +127,105 @@ def addTile(self, tile, dependent_tiles=None):
...
+        return self.registered_positions_m, dep_tile_positions
+
+    def _insert_tile_to_grid(self, tile):
+        """
+        Stores the tile at the proper place in the grid. If necessary, the grid is
+        extended to accommodate the new tile.
+
+        :param tile: (DataArray) tile to be inserted
+        :returns: (int, int) row, col grid position of the tile
+        :updates self.tiles, self.posX, self.posY, self.ovrlp, self.px_size, self.size:
```
nandishjpatel, line 171:
```suggestion
        :updates self.tiles, self.pos_x, self.pos_y, self.ovrlp, self.px_size, self.size:
```

src/odemis/acq/stitching/_registrar.py
```diff
@@ -614,52 +603,57 @@ def _get_shift(self, prev_tile, tile):
...
+            logging.info(f"Calculated shift {meas_tile_dist_px} is larger than the overlap size {overlap}, "
                          f"using expected position instead.")
-            return exp_shift, 0
+            return exp_tile_dist_px, 0
 
-        return shift_total, ncc
+        return shift_px, ncc
 
-    def _compute_registration(self, row, col):
+    def _compute_registration(self, tile, row, col):
```
nandishjpatel, line 656: Is this param used?

src/odemis/acq/stitching/_registrar.py
```diff
@@ -124,100 +127,105 @@ def addTile(self, tile, dependent_tiles=None):
...
+
+        # Insert new tile either to the right or to the bottom of the
+        # closest tile.
+        ver_diff = md_pos[1] - md_pos_prev[1]
+        hor_diff = md_pos[0] - md_pos_prev[0]
+        if abs(ver_diff) > abs(hor_diff):
+            if ver_diff < 0:
+                self.pos_y = pos_prev[0] + 1
+                self.pos_x = pos_prev[1]
+                if len(self.registered_positions_px) <= self.pos_y:
```
nandishjpatel, line 201:
```suggestion
                if len(self.registered_positions_px[1]) <= self.pos_y:
```

src/odemis/acq/stitching/_registrar.py
```diff
@@ -124,100 +127,105 @@ def addTile(self, tile, dependent_tiles=None):
...
+                    self._updateGrid("y")  # extend grid in y direction
+            else:
+                self.pos_y = pos_prev[0] - 1
+                self.pos_x = pos_prev[1]
+        else:
+            if hor_diff > 0:
+                self.pos_y = pos_prev[0]
+                self.pos_x = pos_prev[1] + 1
+                if len(self.registered_positions_px[0]) <= self.pos_x:
```
nandishjpatel, line 210: This check is a bit different from line 201?

src/odemis/acq/stitching/_registrar.py
```diff
@@ -87,7 +89,8 @@ def __init__(self):
         # initialize grid to 1x1. The size will increase as new tiles are
         # added.
         self.nx = 1  # (int) total number of cols
-        self.registered_positions = [[None]]  # (None or tuple of 2 floats) list of calculated positions
+        self.registered_positions_px = [[None]]  # (None or tuple of 2 floats) list of calculated shifts in px
+        self.registered_positions_m = [[None]]  # (None or tuple of 2 floats) list of calculated positions in m
```
nandishjpatel, line 93: Do you actually need this? Is this for better understanding for the user? I think returning `tile_positions` will just do fine in line 162

nandishjpatel (approved):

src/odemis/acq/stitching/_registrar.py
pieleric: Indeed, this doesn't seem used any where. Also, the comment seem to imply the difference between the two is just a coefficient factor, but one is relative to the center of the first tile, while the other one is in absolute coordinates. Maybe worth mentioning?

src/odemis/acq/stitching/_registrar.py
```diff
@@ -284,9 +291,6 @@ def _estimateMatch(self, imageA, imageB, shift):
         Returns an estimation of the similarity between the given images
         when the second is shifted by the shift value. It is used to assess 
         the quality of a shift measurement by giving the shifted image.
-        imageA:
-        imageB:
-        shift: shift after phase correlation
```
pieleric: Why removing this doc? I'd suggest to at least document their types.

src/odemis/acq/stitching/_registrar.py
pieleric: I think the idea is that this method has the same signature as the parent class.

pieleric (changes requested):

src/odemis/acq/stitching/_registrar.py
tepals: that's indeed the idea. But this also allows me to remove line 666, which I will do.

src/odemis/acq/stitching/_registrar.py
tepals: That is correct. It is a list of lists, therefore `len(self.registered_positions_px)` gives the amount of vertical tiles, and `len(self.registered_positions_px[0])` gives the number of horizontal tiles in the first line

nandishjpatel (approved):

src/odemis/acq/stitching/test/registrar_test.py
```diff
@@ -607,46 +612,61 @@ class TestGlobalShiftRegistrar(unittest.TestCase):
...
-                    start = numpy.add((int(0.2 * size[0]), 0), shift)
-                    end = (start[0] + tsize[0], start[1] + tsize[1])
-                    tile2 = img[start[0]:end[0], start[1]:end[1]]
-
-                    px_size = (1e-6, 2e-5)
+                    # create a black image
+                    img = numpy.zeros(img_size, dtype=dtype)
+                    for _ in range(int(img_size[0] / 3)):
+                        # Randomly draw vertical and horizontal lines
+                        is_vertical = numpy.random.choice([True, False])
```
pieleric, line 633: No need to use numpy for that. `random.choice()` works as fine.

src/odemis/acq/stitching/test/registrar_test.py
```diff
@@ -607,46 +612,61 @@ class TestGlobalShiftRegistrar(unittest.TestCase):
...
+                        if is_vertical:
+                            # Draw a vertical line
+                            x = numpy.random.randint(0, img_size[0])
+                            color = numpy.random.randint(1, 256)
```
pieleric, line 637: `random.randint() should as fine here.

src/odemis/acq/stitching/test/registrar_test.py
```diff
@@ -607,46 +612,61 @@ class TestGlobalShiftRegistrar(unittest.TestCase):
...
+                            img[x:x + 2, :] = color
+                        else:
+                            # Draw a horizontal line
+                            y = numpy.random.randint(0, img_size[1])
+                            color = numpy.random.randint(1, 256)
```
pieleric, line 642: just use `random.randint()` here too.

src/odemis/acq/stitching/test/registrar_test.py
```diff
@@ -311,46 +301,61 @@ class TestShiftRegistrar(unittest.TestCase):
 
     def setUp(self):
         random.seed(1)
+        numpy.random.seed(10)
```
pieleric, line 304: I think you don't need anything specific from numpy.random. In this case, you could drop this line (and its clone in the other test case).

pieleric (changes requested):

src/odemis/acq/stitching/_registrar.py
```diff
@@ -124,100 +127,105 @@ def addTile(self, tile, dependent_tiles=None):
...
+                                   first_position_m[1] - shift[1] * self.px_size[1]))
 
         # Return positions for dependent tiles
         for t, sdts in zip(tile_positions, self.shift_tile_dep_tiles):
             dts = []  # dependent tiles for tile
             for sdt in sdts:
                 dts.append((t[0] + sdt[0], t[1] + sdt[1]))
             dep_tile_positions.append(dts)
+        self.registered_positions_m = tile_positions
+        return self.registered_positions_m, dep_tile_positions
```
tepals, line 162: this should still be removed

src/odemis/acq/stitching/test/registrar_test.py
tepals: thanks, good catch.

PR #2616, opened 2024-01-08
[METEOR-525] [feat] Make filenames of saved features more intuitive 

K4rishma:


K4rishma:
@tepals , the unit test cases are fixed

src/odemis/util/filename.py
```diff
@@ -103,20 +103,21 @@ def guess_pattern(fn):
...
+    # cnt_ptn = r'\d{1,5}'
+    # cnt_m = None
+    # for m in re.finditer(cnt_ptn, fn_ptn):
+    #     cnt_m = m
+    # # if multiple numbers are present, use last
+    # if cnt_m:
+    #     cnt = cnt_m.group()
+    #     fn_ptn = fn_ptn[:cnt_m.start()] + "{cnt}" + fn_ptn[cnt_m.end():]
+    # else:
+    #     cnt = "001"  # will be used in case cnt pattern is added afterwards
```
tepals, line 115: why is this commented?

src/odemis/gui/cont/acquisition.py
```diff
@@ -696,6 +694,30 @@ def _on_export_data_done(self, future):
...
         self._display_acquired_data(data)
 
+    def _add_feature_info_filename(self):
+        """
+        Add feature name, feature status and the counter at the end of the filename.
+        """
+        local_tab = self._tab_data.main.getTabByName("cryosecom-localization")
+        local_tab.tab_data_model.select_current_position_feature()
+        feature_pattern = (local_tab.tab_data_model.main.currentFeature.value.name.value +
+                 '-' + local_tab.tab_data_model.main.currentFeature.value.status.value)
```
tepals, line 704: Small suggestion, do with it what you want:

This is not the nicest to read and it is hard to see what the pattern looks like, I would suggest something like:
```suggestion
        feature_name = local_tab.tab_data_model.main.currentFeature.value.name.value
        feature_status = local_tab.tab_data_model.main.currentFeature.status.name.value
        feature_pattern = f"{feature_name}-{feature_status}"
```

tepals (changes requested):

tepals (approved):

src/odemis/gui/cont/acquisition.py
```diff
@@ -696,6 +694,31 @@ def _on_export_data_done(self, future):
...
+        # during the creation of filename
+        self._config.fn_count = '0'
+
+        if 'Feature' in self._config.fn_ptn:
+            self._config.fn_ptn = re.sub(r'(Feature).*', feature_pattern + '-{cnt}', self._config.fn_ptn)
+        else:
+            self._config.fn_ptn = re.sub('-{cnt}', '-' + feature_pattern + '-{cnt}', self._config.fn_ptn)
+
+        # update the filename for the current acquisition
+        self._filename.value = create_filename(self._config.pj_last_path, self._config.fn_ptn,
```
pieleric, line 718: I don't think that's the right abstraction model. You are fighting against `create_filename()` to make something that matches your expectations. I think you should keep it more straightforward: take the current filename as it is, add the "feature name + -001" at the end of it (and before the extension), and call `make_unique()` with the list of files in the directory. (or just compute the first count which is not existing yourself. 
=> No need to touch `util.filename`.

pieleric (changes requested):

src/odemis/util/filename.py
```diff
@@ -102,21 +102,7 @@ def guess_pattern(fn):
...
-    if cnt_m:
-        cnt = cnt_m.group()
-        fn_ptn = fn_ptn[:cnt_m.start()] + "{cnt}" + fn_ptn[cnt_m.end():]
-    else:
-        cnt = "001"  # will be used in case cnt pattern is added afterwards
-
-    # If neither time, nor count are specified, add count, if only short time (h, min)
-    # specified and no count, add seconds to make filename unique
-    # Doesn't behave properly if user enters terms with curly braces like {{cnt}}.
+    cnt = "001"
```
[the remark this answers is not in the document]
K4rishma: If the end number in the filename is a count, then adding a {cnt} separately at the end is not a good method, as you mentioned, "But having "test-123' converted to 'test-123-001' instead of 'test-124' doesn't sound good". In cases where the users have a special name for their files like "bacteria_na_08.tiff", "virus_50x", then in this changing the filename to  "bacteria_na_09.tiff", would not be the best option. I feel the naming by the users can vary across different users. To tackle it, if we do not touch their given name and append the counter at the end, it might be straightforward? For the users,  the order of the acquired images will be clear when the counter is starting from 1 ?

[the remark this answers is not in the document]
K4rishma: Should I leave it like this for time-being till we know the issue or reset my changes?

src/odemis/gui/cont/acquisition/cryo_acq.py
```diff
@@ -274,6 +276,7 @@ def _display_acquired_data(self, data):
         """
         # get the localization tab
         local_tab = self._tab_data.main.getTabByName("cryosecom-localization")
+        # local_tab.tab_data_model.select_current_position_feature()
```
pieleric, line 279: Drop?

src/odemis/gui/cont/acquisition/cryo_acq.py
```diff
@@ -294,6 +320,8 @@ def _export_data(self, data, thumb_nail):
         """
         filename = self._filename.value
         if data:
+            self._add_feature_info_filename()
+            filename = self._filename.value
```
pieleric, line 324: I think it'd best so that it doesn't change the VA:
```suggestion
            filename = self._add_feature_info_filename(filename)
```

src/odemis/util/filename.py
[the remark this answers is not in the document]
pieleric: Please drop your changes to `guess_pattern()`, and handle separately the feature number in the `_add_feature_info_filename()` function.

src/odemis/gui/cont/acquisition/cryo_acq.py
```diff
@@ -510,6 +538,17 @@ def _on_filename(self, name):
...
+        if self.overview_acqui_controller._tab_data_model.main.currentFeature.value is None:
+            self._config.fn_count = '0'
+            self._filename.value = create_filename(
+                self._config.pj_last_path,
+                self._config.fn_ptn,
+                self._config.last_extension,
+                self._config.fn_count,
+            )
+            name = self._filename.value
+
```
pieleric, line 551: I don't think you need that. `create_filename()` will automatically look at the current files and increase the counter if needed.

src/odemis/gui/cont/acquisition/cryo_acq.py
```diff
@@ -286,6 +289,29 @@ def _on_export_data_done(self, future):
...
+            base_name = re.sub(r'(Feature).*', f'{feature_pattern}-0', self._config.fn_ptn)
+        else:
+            # If feature pattern not present in filename add it
+            # Add the feature pattern and count at the end of the given filename
+            base_name = re.sub('{cnt}', f'-{feature_pattern}-0', self._config.fn_ptn)
+
+        # update the filename for the current acquisition
+        list_names = os.listdir(self._config.pj_last_path)
+        name = make_unique_name(f'{base_name}{self._config.last_extension}', list_names)
+        self._filename.value = os.path.join(self._config.pj_last_path, name)
```
pieleric, line 313: I think that instead of basing yourself on the pattern, you should take the expected filename (as argument), compute the basename and extension (using util.split_ext) and then create a new name by creating a pattern which only has fixed elements and `{cnt}`, like: `f"{basename}-{feature_name}-{feature_status}-\{cnt\}"`. Then  util.filename.create_filename() will automatically find the right count based on the current files in the directory.

You can return the filename to the caller.

src/odemis/gui/cont/acquisition/cryo_acq.py
```diff
@@ -286,6 +289,29 @@ def _on_export_data_done(self, future):
...
+        """
+        local_tab = self._tab_data.main.getTabByName("cryosecom-localization")
+        feature_name = local_tab.tab_data_model.main.currentFeature.value.name.value
+        feature_status = local_tab.tab_data_model.main.currentFeature.value.status.value
+        feature_pattern = f"{feature_name}-{feature_status}"
+
+        if 'Feature' in self._config.fn_ptn:
+            # If feature pattern is present in filename
+            # Replace Feature and everything after with the current feature pattern and the updated count
+            base_name = re.sub(r'(Feature).*', f'{feature_pattern}-0', self._config.fn_ptn)
```
pieleric, line 304: I'd rather drop this feature. It's too hidden as-is for the user to discover it. For now it's fine to *always* have the same filename modification, irrespective of the filename content.

pieleric (changes requested):

K4rishma:
ready for review

src/odemis/gui/cont/acquisition/cryo_acq.py
tepals: did you forget to drop it?

src/odemis/util/filename.py
```diff
@@ -30,16 +30,16 @@
...
+    Generates a filename pattern form a given filename. The function detects certain
     formats of dates, time, and counters. It returns a string with replacement fields which
     can be used in create_filename to provide an updated version of the filename.
-    input
-        fn (String): filename (may include path and extension)
-    returns 
-        fn_ptn (String): filename pattern (without path and extension)
-        cnt (String): counter
+    :param fn: (String) filename (may include path and extension)
+    :param is_count: (Boolean) by default True, if count is the name is to be detected, else False
```
tepals, line 39:
```suggestion
    :param fn: (string) filename (may include path and extension)
    :param is_count: (boolean) by default True, if count in the name is to be detected, else False
```

src/odemis/util/filename.py
```diff
@@ -30,16 +30,16 @@
...
+    return:
+        fn_ptn: (String) filename pattern (without path and extension)
+        cnt: (String) counter
```
tepals, line 42:
```suggestion
        fn_ptn: (string) filename pattern (without path and extension)
        cnt: (string) counter
```

tepals (changes requested):

tepals (approved):

src/odemis/util/filename.py
```diff
@@ -30,16 +30,16 @@
 from datetime import datetime, timedelta
 
 
-def guess_pattern(fn):
+def guess_pattern(fn, is_count=True):
```
pieleric, line 33: `is_count` doesn't sound proper English and make it a little unclear. I'd suggest `detect_count`.

src/odemis/util/filename.py
```diff
@@ -103,29 +103,29 @@ def guess_pattern(fn):
...
-        if '{timeshrt_colon}' in fn_ptn:
-            fn_ptn = re.sub('{timeshrt_colon}', '{timelng_colon}', fn_ptn)
-        elif ("{timelng}" in fn_ptn or
-              "{timelng_colon}" in fn_ptn or
-              "{timelng_hyphen}" in fn_ptn):
-            pass
-        else:
-            fn_ptn = fn_ptn + '-{cnt}'
+    cnt = "001"  # will be used in case cnt pattern is added afterwards
+    if is_count:
```
pieleric, line 107: Please add (at least) one test case, which calls the function with `is_count=False`.

pieleric (changes requested):

PR #2623, opened 2024-01-17
[refactor] odemis.gui.cont.tabs.py file

nandishjpatel:
[refactor] create a tabs subpackage in odemis.gui.cont, all the classes earlier in odemis.gui.cont.tabs.py have been split into individual python files under the odemis.gui.cont.tabs subpackage.

The refactoring will result in a better file structure.


src/odemis/gui/cont/tabs/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+"""
+
+from odemis.gui.model import TOOL_ZOOM, TOOL_ROI, TOOL_ROA, TOOL_RO_ANCHOR, \
+    TOOL_POINT, TOOL_LINE, TOOL_SPOT, TOOL_ACT_ZOOM_FIT, TOOL_RULER, TOOL_LABEL, \
+    TOOL_FEATURE
+
+# The constant order of the toolbar buttons
+TOOL_ORDER = (TOOL_ZOOM, TOOL_ROI, TOOL_ROA, TOOL_RO_ANCHOR, TOOL_RULER, TOOL_POINT,
+              TOOL_LABEL, TOOL_LINE, TOOL_SPOT, TOOL_ACT_ZOOM_FIT, TOOL_FEATURE)
+
```
pieleric, line 33: I'd suggest to move these constants to `odemis.gui.__init__.py` (also the part which is in `gui.model.__init__.py`).

src/odemis/gui/cont/tabs/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+# Preferable autofocus values to be set when triggering autofocus in delphi
+# Used in SecomStreamsTab
+AUTOFOCUS_BINNING = (8, 8)
+AUTOFOCUS_HFW = 300e-06  # m
```
pieleric, line 37: If the constants are only used by a single class, move them to the same file.
=> Move to `secom_streams_tab.py`

src/odemis/gui/cont/tabs/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+
+# Different states of the mirror stage positions
+# Used in ChamberTab
+MIRROR_NOT_REFD = 0
+MIRROR_PARKED = 1
+MIRROR_BAD = 2  # not parked, but not fully engaged either
+MIRROR_ENGAGED = 3
```
pieleric, line 44: Move to `sparc2_chamber_tab.py`

src/odemis/gui/cont/tabs/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+
+# Position of the mirror to be under the e-beam, when we don't know better
+# Note: the exact position is reached by mirror alignment procedure
+# Used in Sparc2AlignTab, ChamberTab
+MIRROR_POS_PARKED = {"l": 0, "s": 0}  # (Hopefully) constant, and same as reference position
+MIRROR_ONPOS_RADIUS = 2e-3  # m, distance from a position that is still considered that position
```
pieleric, line 50: I think it's better to not have these constants duplicated, but also they shouldn't be in __init__.py. You can move them to a _constants.py file, which is imported by the classes which need it.

src/odemis/gui/cont/tabs/correlation_tab.py
```diff
@@ -0,0 +1,253 @@
...
+import odemis.gui.cont.export as exportcont
+import odemis.gui.cont.streams as streamcont
+import odemis.gui.cont.views as viewcont
+import odemis.gui.model as guimod
+import odemis.gui.util as guiutil
+from odemis.acq.stream import OpticalStream, EMStream, StaticStream
+from odemis.gui.cont.correlation import CorrelationController
+from odemis.gui.model import TOOL_ACT_ZOOM_FIT
+from odemis.gui.util import call_in_wx_main
+from odemis.gui.cont.tabs import TOOL_ORDER
```
pieleric, line 51: You really want to avoid importing `.tabs`, as `.tabs` imports this module. It's a cyclic import. Python sometimes manages to handle that (and that's the probably the case here), but it's very brittle. Just moving the import order can break it and cause a lot of hair pulling. 
=> no constants used by these sub-modules should be directly defined in `__init__.py`.

pieleric (changes requested):

src/odemis/gui/cont/tabs/tab_bar_controller.py
```diff
@@ -0,0 +1,188 @@
...
+Odemis. If not, see http://www.gnu.org/licenses/.
+
+"""
+
+import logging
+import wx
+# IMPORTANT: wx.html needs to be imported for the HTMLWindow defined in the XRC
+# file to be correctly identified. See: http://trac.wxwidgets.org/ticket/3626
+# This is not related to any particular wxPython version and is most likely permanent.
+import wx.html
```
pieleric, line 31: This only needs to be done *once* somewhere in Odemis before loading a XRC file. 
I'd suggest to move it somewhere else. Maybe `gui.main.py` is the best place. Alternatively, we could "hide" it a little bit more by placing it in `gui.xmlh.__init__.py`. I don't think it helps to hide things though...

tepals: Might be good to add to clarify in the comment that this only needs to be done once somewhere in Odemis before loading a XRC file.

src/odemis/gui/cont/tabs/__init__.py
nandishjpatel: Can I move the `TOOL_ORDER` constant to `gui.model.__init__.py` instead?

tepals (approved):

src/odemis/gui/cont/tabs/__init__.py
pieleric: You could. I find it a little bit less matching because this is about the presentation (order) and the model should not really care about that. That's why I would suggest to just move everything to gui.init.py . But if you want to change this refactoring light, it's fine to move to model, and later we could move it again...

nandishjpatel: > But if you want to change this refactoring light, it's fine to move to model, and later we could move it again...

Indeed I want to keep this refactoring light.

pieleric (approved):

PR #2626, opened 2024-01-19
[METEOR-811] Add CLEM map to localization tab for feature selection

patrickcleeve2:
Features:
- Sync correlation and localization streams
- Allows selection of feaures in combined clem map

### DIFF
--- a/src/odemis/gui/cont/correlation.py
+++ b/src/odemis/gui/cont/correlation.py
@@ -32,8 +32,9 @@ import wx.html
 import odemis.acq.stream as acqstream
 import odemis.gui.model as guimod
 from odemis import model
-from odemis.acq.stream import StaticStream
+from odemis.acq.stream import StaticStream, StaticFluoStream, StaticSEMStream
 from odemis.gui.util import call_in_wx_main
+from odemis.gui.cont.tabs.localization_tab import LocalizationTab
 
 # TODO: move to more approprate location
 def update_image_in_views(s: StaticStream, views: list) -> None:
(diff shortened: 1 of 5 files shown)


src/odemis/gui/cont/correlation.py
```diff
@@ -70,10 +71,23 @@ def __init__(self, tab_data, panel, tab, viewports):
...
 
         # connect the correlation streams to the tab data
         self._panel.cmb_correlation_stream.Bind(wx.EVT_COMBOBOX, self._on_selected_stream_change)
 
+        # reference frames
+        self.ref_frames = ["METEOR", "FIBSEM", "No Reference"]
+        self.ref_stream_map = {"METEOR": StaticFluoStream, 
+                               "FIBSEM": StaticSEMStream, 
+                               "No Reference": type(None)} # match everything except None 
+        self.ref_stream = self.ref_stream_map[self.ref_frames[0]]
```
tepals, line 84: Since you are always accessing `self.ref_stream_map` with an index, why not just only have a list? Or have a dict with indices as the keys?

```suggestion
        self.ref_stream_map = [
            StaticFluoStream,  # METEOR
            StaticSEMStream,  # FIBSEM
            type(None)  # match everything except None  (No reference type)
        ]
        self.ref_stream = self.ref_stream_map[0]
```

or
```suggestion
        self.ref_stream_map = {0: StaticFluoStream,  # METEOR
                               1: StaticSEMStream,  # FIBSEM
                               2: type(None)} # match everything except None (No reference type)
        self.ref_stream = self.ref_stream_map[0]
```

patrickcleeve2: changed it too a dictionary, but keeping the self.ref_frames (renamed to self.ref_frame_names) because they are used for display

nandishjpatel (approved):

tepals (approved):

patrickcleeve2:
Fixed conflict with ui refactor

PR #2629, opened 2024-01-19
[refactor] odemis gui comp stream.py file

nandishjpatel:
[refactor] all the classes earlier in odemis.gui.comp.stream.py have been split into individual python files.

The stream file has been split into stream_bar and stream_panel files. The fastem specific code unrelated to the stream has also been sperated.


src/odemis/gui/comp/fastem.py
pieleric, line 1: `gui.comp.fastem` sounds too generic, especially compared to the other files like "stream_panel" et al. 
Maybe you could name it `fastem_panel` or  break it into `fastem_project` for the first couple of classes, and then `fastem_calib` for the last 3 classes?

src/odemis/gui/comp/_constants.py
```diff
@@ -0,0 +1,33 @@
...
+"""
+
+import wx.lib.newevent
+
+# Used by StreamPanel and StreamBar
+stream_remove_event, EVT_STREAM_REMOVE = wx.lib.newevent.NewEvent()
+
+# Used by FastEMProjectPanelHeader and StreamPanelHeader
+CAPTION_PADDING_RIGHT = 5
+ICON_WIDTH, ICON_HEIGHT = 16, 16
```
pieleric, line 33: Let's just duplicate these constants in the 2 files they are used. Especially, I can easily imagine that as the implementation diverges these values will be different.

Especially, the advantage is that you can drop this entire file now.

pieleric (changes requested):

src/odemis/gui/comp/fastem.py
nandishjpatel: Can I leave this as it is now? I will give more meaningful naming during VolumeEM GUI development?

src/odemis/gui/comp/stream_panel.py
```diff
@@ -46,9 +48,8 @@
 from odemis.gui.conf.data import COLORMAPS
 import matplotlib.colors as colors
 
-stream_remove_event, EVT_STREAM_REMOVE = wx.lib.newevent.NewEvent()
-stream_visible_event, EVT_STREAM_VISIBLE = wx.lib.newevent.NewEvent()
-stream_peak_event, EVT_STREAM_PEAK = wx.lib.newevent.NewEvent()
+StreamVisibleEvent, EVT_STREAM_VISIBLE = wx.lib.newevent.NewEvent()
+StreamPeakEvent, EVT_STREAM_PEAK = wx.lib.newevent.NewEvent()
```
pieleric, line 52: Please move these ones to `odemis.gui.evt` too.

pieleric (changes requested):

src/odemis/gui/comp/fastem.py
pieleric: OK.

tepals (approved):

pieleric (approved):

PR #2631, opened 2024-01-22
[refactor] odemis.gui.model.__init__.py file

nandishjpatel:
[refactor] all the classes earlier in odemis.gui.model.__init__.py have been split into individual python files.


pieleric (changes requested):
Please don't change what `odemis.gui.model` provides. It should import all the classes and constants it used to contain.

Also, don't break every single class into a separate file. I'd suggest the following files:
* `main_gui_data.py`
* `tab_gui_data.py`, which contains  all the MicroscopyGUIData classes
* `file_info.py`
* `stream_view.py`
* `_constants.py`

nandishjpatel:
> Please don't change what `odemis.gui.model` provides. It should import all the classes and constants it used to contain.
> 
> Also, don't break every single class into a separate file. I'd suggest the following files:
> 
>     * `main_gui_data.py`
> 
>     * `tab_gui_data.py`, which contains  all the MicroscopyGUIData classes
> 
>     * `file_info.py`
> 
>     * `stream_view.py`
> 
>     * `_constants.py`

Understood, I make the changes

pieleric (approved):

PR #2636, opened 2024-01-23
[refactor] main.xrc file

nandishjpatel:


src/odemis/gui/xmlh/generate_xrc.py
```diff
@@ -169,12 +174,19 @@ def GenerateInitResourcesEmbedded(self, resourceFilename, resourceDocument):
...
-        args = sys.argv[1:]
-
     inputFiles = []
-    for arg in args:
-        inputFiles += glob.glob(arg)
+
+    if args:
+        outputFilename = os.path.splitext(args[0])[0] + "_xrc.py"
+        for arg in args:
+            inputFiles += glob.glob(arg)
```
pieleric, line 182: So this will keep working for `src/odemis/gui/win/dialog.xrc` and other xrc files, right?

src/odemis/gui/xmlh/generate_xrc.py
pieleric, line 1: Why moving this script there? I think it's nice that it's easy to find.
In any case, we should update the developer manual, in `doc/develop/`.

src/odemis/gui/main_xrc.py
pieleric, line 1: What's the reason for not breaking this file into sub-files too? I think it would then be possible to skip some loading for the unused tabs (as `get_resources()`) would be called per tab. But that will most likely not be very significant, as the XML data is only parsed for the tabs which are used, already. However, I imagine it'd be a little clearer which `_xrc.py` contains what.

pieleric (changes requested):

tepals:
Would it be possible to add a single line explanation to each file? I don't know what `fr_` stands for, so it is still hard to understand what is in those files.

src/odemis/gui/xmlh/generate_xrc.py
nandishjpatel: Correct! it works, see below
```
patel@laptop:~/development/odemis/src/odemis/gui/xmlh$ ./generate_xrc.py ../win/dialog.xrc 
Resources written to ../win/dialog_xrc.py.
```

src/odemis/gui/main_xrc.py
nandishjpatel: I just wanted the current way of `main_xrc.py` file to work (it a auto-generated file, so I feel no need to look into it). I will give it a try!

nandishjpatel: The functions in https://github.com/delmic/odemis/blob/master/src/odemis/gui/xmlh/__init__.py is another reason for not breaking into sub-files.

nandishjpatel:
> Would it be possible to add a single line explanation to each file? I don't know what `fr_` stands for, so it is still hard to understand what is in those files.

`<object class="wxDialog" name="fr_acq">` the object class is mentioned alongside the name. Should I name the xrc file as `dialog_acq` and also the name? @pieleric do you have any suggestions?

pieleric:
> `<object class="wxDialog" name="fr_acq">` the object class is mentioned alongside the name. Should I name the xrc file as `dialog_acq` and also the name? @pieleric do you have any suggestions?
I would be careful with adding comments in these files because they are (were) supposed to be edited by an external tool, which might not preserve the comments.

Admittedly the names are not great (and that's not @nandishjpatel 's fault!). FYI, I beleive `fr_` stands for "frame" as it's how a window is called in wxWindow... however most of them are actually "Dialogs", which is a slightly different type of window. So definitely not great names! So I agree that `dialog_*` would be a better name (and `frame_main` or `window_main`).  You could also consider renaming the XRC objects... but that might cause too much changes for a single refactoring.

pieleric (approved):

K4rishma (approved):

tepals (approved):

sneepstefan (approved):

src/odemis/gui/main_xrc.py
patrickcleeve2, line 1: When I run `./gui/xmlh/generate_xrc.py` even with no changes to any of the xrcs? Is that normal or  should this be regenerated before merging?

patrickcleeve2 (approved):

src/odemis/gui/xmlh/generate_xrc.py
```diff
@@ -169,12 +174,19 @@ def GenerateInitResourcesEmbedded(self, resourceFilename, resourceDocument):
...
+    else:
+        outputFilename = MAIN_XRC
+        for filename in os.listdir(XMLH_RESOURCES_DIR):
```
pieleric, line 185:
```suggestion
        for filename in sorted(os.listdir(XMLH_RESOURCES_DIR)):
```

src/odemis/gui/main_xrc.py
pieleric: That's a very good point. We don't want that every time we change a line of a XRC file, the `main_xrc.py` looks like it has been completely re-shuffled. I think it's easy to fix it by just sorting the file by name. See my suggestion.

pieleric (changes requested):

pieleric (approved):

PR #2639, opened 2024-01-25
[SDEV-1770] refactor: Move focus metrics from acq/align/autofocus.py 

tepals:
to avoid circular import

MeasureSpotsFocus will need to be re-used in the fastem-calibrations code, but because that gets imported in odemis/acq/align/fastem.py it is causing a circular import. Since the functions are more utility functions of the autofocus code, I moved them to odemis/util/focus.py

relates to: https://bitbucket.org/delmic/fastem-calibrations/pull-requests/143

### DIFF
--- a/src/odemis/util/focus.py
+++ b/src/odemis/util/focus.py
@@ -1,0 +1,150 @@
+# -*- coding: utf-8 -*-
+"""
+Created on 12 Apr 2013
+
+@author: Rinze de Laat
+
+Copyright © 2013-2024 Éric Piel, Rinze de Laat, Thera Pals, Delmic
+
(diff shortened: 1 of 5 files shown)


tepals:
@pieleric is this the right way of solving this issue?

pieleric (approved):
That looks like a good way to solve the dependency issue.

src/odemis/util/focus.py
```diff
@@ -0,0 +1,129 @@
+import logging
```
nandishjpatel, line 1: Can you add the author and license?

nandishjpatel (approved):

PR #2640, opened 2024-01-26
[FIX][METEOR-933] Prevent move to feature while under SEM

patrickcleeve2:
Currently, users can press the go to feature button while under the SEM and odemis will move directly to the stage + objective position without any safety controls or checks which are implemented as part of the CryoSwitchPosition method. We should prevent this, until we explicitly support moving to the feature under the SEM

### DIFF
--- a/src/odemis/gui/cont/features.py
+++ b/src/odemis/gui/cont/features.py
@@ -3,6 +3,7 @@ import logging
 import wx
 
 from odemis.acq.feature import FEATURE_ACTIVE, FEATURE_ROUGH_MILLED, FEATURE_DEACTIVE, save_features, FEATURE_POLISHED
+from odemis.acq.move import  FM_IMAGING, POSITION_NAMES
 from odemis.gui.model import TOOL_FEATURE
 from odemis.gui.util.widgets import VigilantAttributeConnector
 from odemis.gui.util import call_in_wx_main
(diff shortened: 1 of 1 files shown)


src/odemis/gui/cont/features.py
```diff
@@ -3,6 +3,7 @@
 import wx
 
 from odemis.acq.feature import FEATURE_ACTIVE, FEATURE_ROUGH_MILLED, FEATURE_DEACTIVE, save_features, FEATURE_POLISHED
+from odemis.acq.move import UNKNOWN, SEM_IMAGING, FM_IMAGING, POSITION_NAMES
```
pieleric, line 6: You don't need to import all of these, do you?

src/odemis/gui/cont/features.py
```diff
@@ -87,9 +88,29 @@ def _on_btn_go_to_feature(self, _):
...
         pos = feature.pos.value
+
+        # get current position
+        pm = self._tab_data_model.main.posture_manager
+        current_label = pm.getCurrentPostureLabel()
+        logging.debug(f"Current posture: {POSITION_NAMES[current_label]}")
+
+        if current_label != FM_IMAGING: # TODO: @patrick remove this once SEM move is supported 
+            self._display_go_to_feature_warning()
+            logging.warning(f"Currently under {POSITION_NAMES[current_label]}, moving to feature position is not yet supported.")
```
pieleric, line 99: Please move the warning to above the dialog box.
Also lower its level to "info". Warning level is for potential issues detected that we don't handle. It will turn the log button to orange to attract the attention of the user. Now that the move is stopped, it's not that bad, and the user is fully aware already, so "info" is enough.

pieleric (changes requested):

K4rishma (approved):

tepals (approved):

PR #2641, opened 2024-01-28
[METEOR-589][feature] display the most recently used settings of stream panels 

K4rishma:
Display most recently used settings (MRU) in the combo box of add stream button in the localisation tab.

Things left to do:-
1. **Add the tint settings** - It is a bit tricky to add tint. When it is in tuple format of 3 floats, it is easy to save it. But, when the tint value is of type LinearSegmentedColormap which is a custom color map, I am still exploring how to save such a type and read it. - **DONE**
2. **Test cases for the gui** - the testing for the full process i.e. save the setting of the stream controller and reading to checking the values in the json file and also vice-versa i.e. check the values of the recently used setting with the settings displayed in the gui. For both the workflows, I am not sure if it can be tested in the unit testing. 
3. **Review of Specifications document** -  Document incomplete for the review stage

@pieleric and @patrickcleeve2 taggig you to update you with the progress on this task. 

@pieleric I have for now directly saved the stream values in the json file as it is easy to read it and write it. I checked with the the functions in preset.py. The  input argument of the function apply_preset(preset) , consists of stream settings saved as dict of SettingEntries -> value.  In order to save such type in the json file, I have to decode SettingEntries in json format. As we just need the value of the vigilant attribute, I am not sure if saving the value_ctrl and label_ctrl in SettingEntries will be advantageous or needed. Is it a good idea to save directly the values without using preset.py functions?

### DIFF
--- a/src/odemis/acq/stream_settings.py
+++ b/src/odemis/acq/stream_settings.py
@@ -1,0 +1,177 @@
+# -*- coding: utf-8 -*-
+"""
+Created on 24 Jan 2024
+
+@author: Karishma Kumar
+
+Copyright © 2024 Karishma Kumar, Delmic
+
+This file is part of Odemis.
+
+Odemis is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License version 2 as published by the Free Software Foundation.
+
+Odemis is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with Odemis. If not, see http://www.gnu.org/licenses/.
+"""
+import json
+import logging
+import os
+from typing import List, Dict, Any
+
+from odemis import model, util
+from odemis.acq.stream import FluoStream
+from odemis.util.img import md_format_to_tint, tint_to_md_format
+
+# Power value depends on selected excitation
+# Emission value depends on selected excitation
+# Tint can automatically change if excitation and emission are changed
+# Therefore, the below VAs are changed first in the below order and then the rest of stream VAs
+SETTINGS_ORDER = ["excitation", "power", "emission"]
+# The below VAs are not important for loading the settings of a specific stream
+NON_SETTINGS_VA = ['acquisitionType', 'auto_bc', 'background', 'histogram', 'image', 'intensityRange', 'is_active',
+                   'roi', 'should_update', 'single_frame_acquisition', 'status']
+
+
+def get_settings_order(stream):
+    """
+    Get the VAs for loading a stream setting and sort them so it can applied in the given order while loading
+    a stream later.
+    :param stream: stream object that consists of VAs
+    """
+    stream_vas = model.getVAs(stream)
+    settings = set(stream_vas.keys()).difference(NON_SETTINGS_VA)
+    settings_order = util.sorted_according_to(settings, SETTINGS_ORDER)
+
+    return settings_order
+
+
+class StreamSettingsConfig:
+    """
+    Read and write a list of most recently used (mru) settings from and to a JSON file respectively. The first element
+    in the list will be the latest used setting and the last element will be the oldest used setting. The list will
+    consist of finite number of entries which will update in the above mention order each time the list of stream
+    settings is updated.
+    """
+
+    def __init__(self, file_path, max_entries):
+        """
+        :param file_path: full directory path to save the JSON file
+        :param max_entries: number of entries to save in the JSON file
+        """
+        self.max_entries = max_entries
+        self.config_data: List[Dict[str, Any]] = []
+        self.file_path = file_path
+        try:
+            self._read()  # update the config_data by reading the JSON file
+        except Exception as e:
+            logging.exception(f"Error in reading the stream settings: {e}")
+
+    def _read(self):
+        logging.debug(f"Reading the most recently used stream settings from {self.file_path}")
+        if os.path.exists(self.file_path):
+            with open(self.file_path, 'r') as jsonfile:
+                loaded_data = json.load(jsonfile)
+            if isinstance(loaded_data, list) and all(isinstance(item, dict) for item in loaded_data):
+                self.config_data = loaded_data
+            else:
+                # Log a warning and reset config_data to an empty list
+                logging.warning("The loaded JSON data is not of the expected format.")
+                self.config_data = []
+
+    def _write(self):
+        logging.debug(f"Writing the most recently used stream settings in {self.file_path}")
+        if self.config_data:
+            with open(self.file_path, 'w') as jsonfile:
+                json.dump(self.config_data, jsonfile, indent=2)
+
+    @property
+    def entries(self):
+        return [e["name"] for e in self.config_data]
+
+    def _get_config_index(self, config_data: list, name: str) -> int:
+        """
+        Get the index of the config entry with the given name
+        :param config_data: list of config entries
+        :param name: name of config entry
+        :return: index of config entry with given name
+        """
+        try:
+            index = [x["name"] for x in config_data].index(name)
+        except ValueError:
+            return None
+        except TypeError:
+            return None
+        return index
+
+    def update_data(self, new_entries: List[Dict[str, Any]]):
+        """
+        Update the saved stream setting with the list of new_entries. First element in JSON file
+        is the latest entry while the last element is the oldest stream setting. The file updates
+        its entries based on the new_entries while maintaining a fixed length of entries in the file.
+        Each entry must have a key "name".
+        :param new_entries: Each element in the list represent the setting
+         of the stream saved as a list from oldest to latest stream setting
+        """
+        for e in new_entries:
+            # find if the new entry name exists in the saved self.config_data
+            index = self._get_config_index(self.config_data, e["name"])
+            # pop the entry from the list if the stream name is same to avoid duplication
+            if index is not None:
+                self.config_data.pop(index)
+            # insert the new entry at the first position
+            self.config_data.insert(0, e)
+            # length of list of entries should be constant
+            self.config_data = self.config_data[:self.max_entries]
+
+        try:
+            self._write()
+        except Exception as e:
+            logging.exception(f"Error in writing the stream settings: {e}")
+
+    def update_entries(self, streams: list):
+        """
+        Update the streams settings from the list of streams to update the entries in the JSON file.
+        :param streams: (list of streams) List of vigilant attributes of each stream settings
+        """
+        local_entries = []
+
+        for s in streams:
+            settings_order = get_settings_order(s)
+            entries = {}
+            # set the settings for the stream from a settings json file
+            # self.stream_settings.get_stream_settings(self.stream_controllers)
+            for attr_name in settings_order:
+                # tint requires special handling
+                if attr_name == "tint":
+                    entries["tint"] = tint_to_md_format(s.tint.value)
+                else:
+                    entries.update({attr_name: getattr(s, attr_name).value})
+
+            local_entries.append(entries)
+
+        # read and update the stream settings in the JSON file
+        self.update_data(local_entries)
+
+    def apply_settings(self, stream: FluoStream, name: str):
+        """
+        Apply the values from the saved data of the given name to the stream
+        controller settings.
+        :param stream: FluoStream that needs to be set with the values from the JSON file
+        :param name: Name of the saved setting in the JSON file
+        """
+        index = self._get_config_index(self.config_data, name)
+        if index is not None:
+            settings_order = get_settings_order(stream)
+            prev_setting = self.config_data[index]
+            # Follow the order for setting the values as listed in setting_keys
+            for key in settings_order:
+                current_value = prev_setting[key]  # Get the previous value if it exists
+                # tint requires special handling
+                if key == "tint":
+                    current_value = md_format_to_tint(prev_setting["tint"])
+                # Special handling for attributes that need to be converted to tuples
+                elif isinstance(current_value, list) and all(isinstance(x, (int, float)) for x in current_value):
+                    current_value = tuple(current_value)
+                # Set the attribute value
+                getattr(stream, key).value = current_value
--- a/src/odemis/acq/stream_settings.py
+++ b/src/odemis/acq/stream_settings.py
@@ -64,8 +64,11 @@ class StreamSettingsConfig:
         self.file_path = file_path
         try:
             self._read()  # update the config_data by reading the JSON file
+        except FileNotFoundError:
+            # Log that the file doesn't exist, which is expected on the first run
+            logging.info("Stream settings file does not exist.")
         except Exception as e:
-            logging.exception(f"Error in reading the stream settings: {e}")
+            logging.exception("Error in reading the stream settings.")
 
     def _read(self):
         logging.debug(f"Reading the most recently used stream settings from {self.file_path}")
@@ -126,8 +129,8 @@ class StreamSettingsConfig:
 
         try:
             self._write()
-        except Exception as e:
-            logging.exception(f"Error in writing the stream settings: {e}")
+        except Exception:
+            logging.exception(f"Error in writing the stream settings")
 
     def update_entries(self, streams: list):
         """
--- a/src/odemis/gui/cont/stream_bar.py
+++ b/src/odemis/gui/cont/stream_bar.py
@@ -24,6 +24,7 @@ Odemis. If not, see http://www.gnu.org/licenses/.
 import functools
 import gc
 import logging
+import os
 from collections import OrderedDict
 import threading
 import time
@@ -33,11 +34,12 @@ import wx
 import odemis.acq.stream as acqstream
 import odemis.gui.model as guimodel
 from odemis import model
+from odemis.acq.stream_settings import StreamSettingsConfig
 from odemis.acq.stream import StaticStream, FastEMOverviewStream
 from odemis.gui.conf.data import get_local_vas
 from odemis.gui.cont.stream import StreamController
 from odemis.gui.model import TOOL_NONE, TOOL_SPOT
-from odemis.gui.util import call_in_wx_main
+from odemis.gui.util import call_in_wx_main, get_home_folder
 
 # There are two kinds of controllers:
 # * Stream controller: links 1 stream <-> stream panel (cont/stream/StreamPanel)
@@ -1198,6 +1200,53 @@ class SecomStreamsController(StreamBarController):
     Controls the streams for the SECOM and DELPHI live view
     """
 
+    def __init__(self, *args, **kwargs):
+        dir = os.path.join(get_home_folder(), u".config/odemis")
+        file_path = os.path.abspath(os.path.join(dir, f"stream_settings.json"))
+        max_entries = 10  # Maximum number most recent settings to display
+        # Initialize a file to save acquired streams
+        self.stream_settings = StreamSettingsConfig(file_path, max_entries)
+        # self.stream_settings must be defined before super() to add new streams
+        super().__init__(*args, **kwargs)
+
+    def update_stream_settings(self):
+        """
+        Update the recently used stream setting when the stream settings are updated during an acquisition
+        """
+        streams = [s.stream for s in self.stream_controllers]
+        self.stream_settings.update_entries(streams)
+
+        if self.stream_settings.config_data:
+            acq_names = sorted(self.stream_settings.entries)
+            # remove the old list of settings and then update the menu list with the updated stream settings
+            for index, label in enumerate(list(self.menu_actions)):
+                self.remove_action(label)
+            for label in acq_names:
+                call_label = functools.partial(self._addFluoWithSettings, label)
+                self.add_action(label, call_label)
+
+    def _addFluoWithSettings(self, label, **kwargs):
+        """
+        Set the Fluostream attributes to the selected acquired stream panel setting
+        :param label: the name of stream used to locate the stream settings in the settings file
+        """
+        stream = acqstream.FluoStream(
+            label,
+            self._main_data_model.ccd,
+            self._main_data_model.ccd.data,
+            self._main_data_model.light,
+            self._main_data_model.light_filter,
+            focuser=self._main_data_model.focus,
+            opm=self._main_data_model.opm,
+            detvas={"exposureTime"},
+        )
+        # set the settings for the stream from a saved settings in the JSON file
+        self.stream_settings.apply_settings(stream, label)
+
+        self._ensure_power_non_null(stream)
+
+        return self._add_stream(stream, **kwargs)
+
     def _createAddStreamActions(self):
         """ Create the compatible "add stream" actions according to the current microscope.
 
@@ -1225,7 +1274,15 @@ class SecomStreamsController(StreamBarController):
 
             # TODO: how to know it's _fluorescent_ microscope?
             # => multiple source? filter?
-            self.add_action("Filtered colour", self._userAddFluo, fluor_capable)
+            # Add most recently used settings from a JSON file
+            if self.stream_settings.config_data:
+                acq_names = sorted(self.stream_settings.entries)
+                for label in acq_names:
+                    call_label = functools.partial(self._addFluoWithSettings, label)
+                    self.add_action(label, call_label, fluor_capable)
+            else:
+                # Add default preset option if no old setting is available
+                self.add_action("Filtered colour", self._userAddFluo, fluor_capable)
 
         # Bright-field & Dark-field are almost identical but for the emitter
         def brightfield_capable():
--- a/src/odemis/gui/cont/tabs/localization_tab.py
+++ b/src/odemis/gui/cont/tabs/localization_tab.py
@@ -422,8 +422,9 @@ class LocalizationTab(Tab):
                 chamber_tab = self.main_data.getTabByName("cryosecom_chamber")
                 chamber_tab.remove_overview_streams([st])
 
-        # Update and save the used stream settings
-        self._streambar_controller.update_stream_settings()
+        # Update and save the used stream settings on acquisition
+        if acquired_streams:
+            self._streambar_controller.update_stream_settings()
 
     @call_in_wx_main
     def display_acquired_data(self, data):
--- a/src/odemis/gui/cont/tabs/localization_tab.py
+++ b/src/odemis/gui/cont/tabs/localization_tab.py
@@ -422,6 +422,8 @@ class LocalizationTab(Tab):
                 chamber_tab = self.main_data.getTabByName("cryosecom_chamber")
                 chamber_tab.remove_overview_streams([st])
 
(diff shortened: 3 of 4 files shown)


src/odemis/acq/acq_stream_settings.py
```diff
@@ -0,0 +1,101 @@
...
+                    data.append(e)
+
+            if len(data) >= MAX_LENGTH:
+                # pop the entry from the list to insert it at the top
+                if index is not None:
+                    data.pop(index)
+                # insert the popped dictionary at the first position
+                data = [e] + data[:MAX_LENGTH - 1]
+                # list of entries should be constant
+                data = data[:MAX_LENGTH]
```
patrickcleeve2, line 98: Instead of checking the length of the list, I would recommend you add a "last_updated" timestamp to the json. That way you can just add/update the config entries, and then sort by last updated and limit the list to 10 entries.


```python

def get_config_index(config_data: list, name: str) -> int:
    """Get the index of the config entry with the given name
    :param config_data: list of config entries
    :param name: name of config entry
    :return: index of config entry with given name
    """
    try:
        index = list(map(lambda x: x["name"], config_data)).index(name)
    except:
        index = None
    
    return index

def add_and_sort_config_entries(config_data: list, new_entries: list, limit: int = 10) -> list[dict]:
    """Add new entries to config_data and sort by last_updated, limit to 10
    :param config_data: list of config entries
    :param new_entries: list of new config entries
    :param limit: limit of config entries
    :return: sorted list of config entries"""

    # add new entries to config_data
    for entry in new_entries:
        index = get_config_index(config_data, name=entry["name"])
        if index is None:
            config_data.append(entry)
        else:
            config_data[index] = entry

    # sort by last_updated (timestamp)
    sorted_config_data = sorted(config_data, key=lambda x: x["last_updated"], reverse=True)

    return sorted_config_data[:limit] # limit to 10

config_data = add_and_sort_config_entries(config_data, new_entries, limit=10)


```

src/odemis/gui/cont/stream_bar.py
```diff
@@ -1198,6 +1199,42 @@ class SecomStreamsController(StreamBarController):
...
+            name,
+            self._main_data_model.ccd,
+            self._main_data_model.ccd.data,
+            self._main_data_model.light,
+            self._main_data_model.light_filter,
+            focuser=self._main_data_model.focus,
+            opm=self._main_data_model.opm,
+            detvas={"exposureTime"},
+        )
+        index = next((i for i, k in enumerate(self.acq_settings.config_data) if k["name"] == name), None)
```
patrickcleeve2, line 1217: This iterator is kind of hard to read, I would recommend replacing it with a helper function, as it is used multiple times, see previous comment

src/odemis/gui/cont/tabs/localization_tab.py
```diff
@@ -419,6 +423,17 @@ def _on_acquired_streams(self, streams):
             correlation_tab = self.main_data.getTabByName("meteor-correlation")
             correlation_tab.correlation_controller.add_streams(streams)
 
+        # Save the used streams in .config/odemis/acquired_stream_mru.json
+        used_streams = acquired_streams.intersection(set(streams))
+        if len(used_streams) > 0:
+            local_entries = [
+                {attr_name: getattr(s.stream, attr_name).value for attr_name in self.acquired_settings.setting_keys}
+                for s in self._streambar_controller.stream_controllers
+            ]
```
patrickcleeve2, line 432: This is a pretty long list comprehension, maybe split it out

src/odemis/acq/acq_stream_settings.py
```diff
@@ -0,0 +1,116 @@
...
+    :param config_data: list of config entries
+    :param new_entries: list of new config entries
+    :param limit: limit of config entries
+    :return: sorted list of config entries
+    """
+
+    # add new entries to config_data
+    for entry in new_entries:
+        index = get_config_index(config_data, name=entry["name"])
+        entry["last_updated"] = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
```
pieleric, line 56: The idea of storing the settings as a list was that you don't need to store the date. So either you store the whole thing as a dict (stream name -> settings), or you store it as a list, and use the order of the list to drop the oldest ones. 
As you know that the oldest ones are at the end of the list, you can just take the original list, remove the entries which are old version of the new entries, insert the new entries at the beginning, and finally the list to `limit-len(new_entries)`.

src/odemis/acq/acq_stream_settings.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+class AcquiredStreamSettings:
+    """
+    Read and write a list of most recently used (mru) settings from and to a JSON file respectively. The first element in the list
+    will be the latest used setting and the last element will be the oldest used setting. The list will consist of
+    finite number of entries which will update in the above mention order each time the list of settings is updated and
+    consequently pushing out the oldest entry from the list.
+    """
+    file_name = "acquired_stream_mru.json"
```
pieleric, line 75: Let's move this to the GUI. I'd suggest more like `stream_settings.json`, or `stream_settings_mru.json`

src/odemis/acq/acq_stream_settings.py
```diff
@@ -0,0 +1,116 @@
+# -*- coding: utf-8 -*-
+"""
+Created on 24 Jan 2024
+
+@author: Karishma Kumar
+
+Copyright © 2024-2025 Karishma Kumar, Delmic
```
pieleric, line 7: Please drop the "2025". Hopefully you're done with this feature before this date :wink:

src/odemis/acq/test/acq_stream_settings_test.py
```diff
@@ -0,0 +1,93 @@
+import logging
```
pieleric, line 1: Unit test files also need a header (name + copyright)

src/odemis/acq/test/acq_stream_settings_test.py
```diff
@@ -0,0 +1,93 @@
...
+                 {'name': 'Filtered Colour 10', 'excitation': [0, 1, 2],
+                  'power': 0.0000000094, 'emission': [0.00001, 1.234e-07],
+                  'detExposureTime': 0.00123456789, 'auto_bc_outliers': 0.12345},
+                 # Add tint
+                 {'name': 'Filtered Colour 1', 'excitation': [0, 1, 2],
+                  'power': 0.0000000095, 'emission': [0.00001, 1.234e-07], "tint": "red",
+                  'detExposureTime': 0.00123456789, 'auto_bc_outliers': 0.12345}]
+
+    @classmethod
+    def setUpClass(cls):
```
pieleric, line 51: Once you have functions in the `stream_settings` module to get/set the settings of one stream, you should add test cases for these functions here too.

pieleric (changes requested):

src/odemis/acq/acq_stream_settings.py
K4rishma: "As you know that the oldest ones are at the end of the list, you can just take the original list, remove the entries which are old version of the new entries, insert the new entries at the beginning, and finally the list to limit-len(new_entries)." This works nicely when the new filter names are different from the names already stored in the list. When we have a duplication i.e. a same filter name, we search that filter name in the list and move it to top i.e. the most recently used.

src/odemis/acq/acq_stream_settings.py
```diff
@@ -0,0 +1,116 @@
...
+class AcquiredStreamSettings:
+    """
+    Read and write a list of most recently used (mru) settings from and to a JSON file respectively. The first element in the list
+    will be the latest used setting and the last element will be the oldest used setting. The list will consist of
+    finite number of entries which will update in the above mention order each time the list of settings is updated and
+    consequently pushing out the oldest entry from the list.
+    """
+    file_name = "acquired_stream_mru.json"
+
+    def __init__(self):
```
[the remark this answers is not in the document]
pieleric: You should pass the full path of the file.

src/odemis/acq/acq_stream_settings.py
pieleric: In my experience, if you try to modify the list at different places, you're easily going to mess up with the index. That is the point of my suggested algorithm: do each change to the list as separate independent steps:
* remove the duplicates from the current list => you're certain the list doesn't contain duplicates
* insert the new entries at the beginning => you're certain all the new entries are at the beginning
* crop => you're certain the list is not too big.

BTW, to insert an element at the beginning of a list you can use `.insert(0, e)`

patrickcleeve2: I don't think we know if new_entries is sorted when it is passed in though? depending how it is updated from the ui, mru might not always be last entry in the list, so we need the date to sort. 

Agree, think it would be simpler if it was a dict, and then we can remove get_index, and just update config_data[name] directly.

pieleric: The `config_data` is sorted (from latest used to oldest). The new_entries are not sorted, but the assumption is that they have *just* been used (that's why they are being added to the list). They all have the same time: right now. So the new_entries always go to the beginning of the list.

patrickcleeve2 (approved):

src/odemis/acq/stream_settings.py
```diff
@@ -0,0 +1,174 @@
...
+                data.pop(index)
+            # insert the new entry at the first position
+            data.insert(0, e)
+            # length of list of entries should be constant
+            data = data[:self.max_entries]
+
+        self.config_data = data
+        self._write()
+
+    def get_settings_order(self, stream: FluoStream):
```
pieleric, line 115: A "get" function has to return something. That's the convention. I think here you're doing some sort of caching, which should really not happen anyway. => just always run the code, and return the order.

Also, it would then become a simple function, which can be moved outside of this class.

src/odemis/acq/stream_settings.py
```diff
@@ -0,0 +1,174 @@
...
+import json
+import logging
+import os
+from typing import List, Dict, Any
+
+from odemis import model, util
+from odemis.acq.stream import FluoStream
+from odemis.util.img import md_format_to_tint, tint_to_md_format
+
+SETTINGS_ORDER = ["name", "excitation", "power", "emission", "tint", "detExposureTime", "auto_bc_outliers"]
```
pieleric, line 26: Please add a comment explaining why the order matters. 
Actually, I thought the order is just to handle the fact that changing the excitation & emission can affect the tint. So only these 3 should be listed. Or do the name also has an effect? (I think it's just a GUI thing, so changing the name directly on the stream wouldn't have any effect.

src/odemis/acq/stream_settings.py
```diff
@@ -0,0 +1,174 @@
...
+NON_SETTINGS_VA = ['acquisitionType', 'auto_bc', 'background', 'histogram', 'image', 'intensityRange', 'is_active',
```
pieleric, line 27: Please add a comment to explain what are these VAs, so that we know when to add a new one (if someone extends a Stream class)

src/odemis/acq/stream_settings.py
```diff
@@ -0,0 +1,174 @@
...
+def get_config_index(config_data: list, name: str) -> int:
+    """
+    Get the index of the config entry with the given name
+    :param config_data: list of config entries
+    :param name: name of config entry
+    :return: index of config entry with given name
+    """
+    try:
+        index = list(map(lambda x: x["name"], config_data)).index(name)
+    except:
```
pieleric, line 40: That's very wide. Can you just explicitly indicate which exceptions you're expecting here?

src/odemis/acq/stream_settings.py
```diff
@@ -0,0 +1,174 @@
...
+
+def get_config_index(config_data: list, name: str) -> int:
+    """
+    Get the index of the config entry with the given name
+    :param config_data: list of config entries
+    :param name: name of config entry
+    :return: index of config entry with given name
+    """
+    try:
+        index = list(map(lambda x: x["name"], config_data)).index(name)
```
pieleric, line 39: Often a list comprehension is easier to read than `map`. Here, I think this is easier:
```suggestion
        index = [x["name"] for x in config_data].index(name)
```

src/odemis/acq/stream_settings.py
```diff
@@ -0,0 +1,174 @@
...
+         stream entry. The number of entries is defined by max_entries.
+        """
+        self.max_entries = max_entries
+        self.config_data = config_data
+        self.file_path = file_path
+        self._read()  # update the config_data by reading the JSON file
+        self.settings_order = None  # the ordered list of setting parameters used for loading any given stream
+
+    def _read(self):
+        logging.debug("Reading the most recently used stream settings in JSON")
```
pieleric, line 69: Please indicate the file path in the log entry.

src/odemis/acq/stream_settings.py
```diff
@@ -0,0 +1,174 @@
...
+        """
+        Update the saved stream setting with the list of new_entries. First element in JSON file
+        is the latest entry while the last element is the oldest stream setting. The file updates
+        its entries based on the new_entries while maintaining a fixed length of entries in the file.
+        :param new_entries: (list of stream controller settings) Each element in the list represent the setting
+         of the stream saved as a list from oldest to latest stream setting
+        """
+        # Convert config data to list to enable list operations
+        if self.config_data is None:
+            self.config_data = []
```
pieleric, line 98: Why not immediately starting with this at init? It'd simplify.

src/odemis/acq/stream_settings.py
```diff
@@ -0,0 +1,174 @@
...
+from odemis import model, util
+from odemis.acq.stream import FluoStream
+from odemis.util.img import md_format_to_tint, tint_to_md_format
+
+SETTINGS_ORDER = ["name", "excitation", "power", "emission", "tint", "detExposureTime", "auto_bc_outliers"]
+NON_SETTINGS_VA = ['acquisitionType', 'auto_bc', 'background', 'histogram', 'image', 'intensityRange', 'is_active',
+                   'roi', 'should_update', 'single_frame_acquisition', 'status']
+
+
+def get_config_index(config_data: list, name: str) -> int:
```
pieleric, line 31: I don't think it makes sense to provide this function for outside of this module, as it's just a helper to the StreamSettings class. => mark it private with a `_`.

src/odemis/acq/stream_settings.py
```diff
@@ -0,0 +1,174 @@
...
+        Update the saved stream setting with the list of new_entries. First element in JSON file
+        is the latest entry while the last element is the oldest stream setting. The file updates
+        its entries based on the new_entries while maintaining a fixed length of entries in the file.
+        :param new_entries: (list of stream controller settings) Each element in the list represent the setting
+         of the stream saved as a list from oldest to latest stream setting
+        """
+        # Convert config data to list to enable list operations
+        if self.config_data is None:
+            self.config_data = []
+        data = self.config_data
```
pieleric, line 99: You're just manipulating references here (not copying it). So it's not helpful to store in `data` and then write `data` back in `config_data`, they are always the same object. 
I think the simplest is to always write `self.config_data` explicitly all the time.

src/odemis/acq/stream_settings.py
```diff
@@ -0,0 +1,174 @@
...
+                logging.warning(f"Error in reading the stream settings: {e}")
+
+    def _write(self):
+        logging.debug("Writing the most recently used stream settings in JSON")
+        try:
+            if self.config_data:
+                with open(self.file_path, 'w') as jsonfile:
+                    json.dump(self.config_data, jsonfile, indent=2)
+        except Exception as e:
+            logging.warning(f"Error in writing the stream settings: {e}")
```
pieleric, line 86: The name of this function is `_write()`. If `write()` fails to write, it's a big deal typically! So I would rather have it pass the exception. The *caller* can decide that failure to write is not a big deal, and just log it, but not the function itself. => move the try/except to the caller. 

Also, I am pretty sure that there is no good reason it didn't work, so please just make it at `log.exception()`.

Same thing with `read()`. Though the failure to read at init could be just because there is no file existing yet, so it would just be log.info if the file doesn't exist.

src/odemis/acq/stream_settings.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def set_stream_settings(self, stream: FluoStream, name: str):
+        """
+        Given the name of the stream, get the values from the saved data of the specific stream and set the values.
+        :param stream: FluoStream that needs to be set with the values from the JSON file
+        :param name: Name of the saved setting in the JSON file
+        """
+        index = get_config_index(self.config_data, name)
+        if index is not None:
+            #
```
pieleric, line 159: ?

src/odemis/acq/stream_settings.py
```diff
@@ -0,0 +1,174 @@
...
+            for key in self.settings_order:
+                current_value = prev_setting[key]  # Get the previous value if it exists
+                # tint requires special handling
+                if key == "tint":
+                    current_value = md_format_to_tint(prev_setting["tint"])
+                # Special handling for attributes that need to be converted to tuples
+                elif isinstance(current_value, list) and all(isinstance(x, (int, float)) for x in current_value):
+                    current_value = tuple(current_value)
+                # Set the attribute value
+                # Exclude the setting the name again
```
pieleric, line 172: Why not? What's the problem?

pieleric (changes requested):
Partial review

src/odemis/gui/cont/stream_bar.py
```diff
@@ -1198,6 +1200,51 @@ class SecomStreamsController(StreamBarController):
     Controls the streams for the SECOM and DELPHI live view
     """
 
+    def __init__(self, *args, **kwargs):
+        dir = os.path.join(get_home_folder(), u".config/odemis")
+        file_path = os.path.abspath(os.path.join(dir, f"stream_settings.json"))
+        max_entries = 10  # Maximum number most recent settings to display
+        # Initialize a file to save acquired streams
+        self.stream_settings = StreamSettings(file_path, max_entries)
+        super().__init__(*args, **kwargs)
```
pieleric, line 1209: Typically, it is suggested to call the super function as first thing in the function, unless there is a specific reason not to. I think here you can call it first.

src/odemis/acq/stream_settings.py
```diff
@@ -0,0 +1,174 @@
...
+    """
+    try:
+        index = list(map(lambda x: x["name"], config_data)).index(name)
+    except:
+        index = None
+
+    return index
+
+
+class StreamSettings:
```
pieleric, line 46: This name sort of implies it's a "settings", but there are other places where the "settings" are the Dict[str, Any]. So it's confusing. I'd suggest to call this class differently, something like `StreamSettingsConfig`, or `StreamSettingsFile`.

src/odemis/acq/stream_settings.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def get_streams_settings(self, streams: list):
+        """
+        Get the streams settings from the stream controller to update the entries in the JSON file.
+        :param streams: (list of streams) List of vigilant attributes of each stream settings
+        """
+        local_entries = []
+
+        if len(streams) > 0:
+            self.get_settings_order(streams[0])
```
pieleric, line 134: You should get the order per stream. Right now you always call this function with the same type of stream, so that works, but it's best to no rely on it. => then you can remove the `if len(...) > 0`.

src/odemis/gui/cont/stream_bar.py
```diff
@@ -1198,6 +1200,51 @@ class SecomStreamsController(StreamBarController):
...
+
+    def update_stream_settings(self):
+        """
+        Update the recently used stream setting when the stream settings are updated during an acquisition
+        """
+        streams = [s.stream for s in self.stream_controllers]
+        self.stream_settings.get_streams_settings(streams)
+
+        if self.stream_settings.config_data:
+            acq_names = sorted([e["name"] for e in self.stream_settings.config_data])
```
pieleric, line 1219: It's not required, but I feel it'd a nicer interface if you provided the list of entries (sorted in MRU) on the `StreamSettingsConfig` class, as a read-only attribute. It's just a matter of adding something like:
```
@property
def entries(self):
   return [e["name"] for e in self.stream_settings.config_data]
```

src/odemis/acq/stream_settings.py
```diff
@@ -0,0 +1,174 @@
...
+        Get the VAs for loading a stream setting and sort them so it can applied in the given order while loading
+        a stream later.
+        :param stream: stream object that consists of VAs
+        """
+        if self.settings_order is None:
+            stream_vas = model.getVAs(stream)
+            settings = set(stream_vas.keys()).difference(NON_SETTINGS_VA)
+            self.settings_order = util.sorted_according_to(settings, SETTINGS_ORDER)
+
+    def get_streams_settings(self, streams: list):
```
pieleric, line 126: The method names `get_stream_settings()`/`set_stream_settings()` are confusing. Especially the `get` doesn't return any value, but instead write something in a file! The `set` reads a file and modifies a stream based on that content. 

I think you have two options. The most straightforward is to keep these 2 methods, but adjust the names. I'd suggest something like `update_entries(streams)` and `apply_settings(stream, entry)`.

A second option would be to separate the notion of reading/writing the config file from the conversion stream<>settings (that's what I had in mind originally in my first review).
So in this class, there would be just two methods `update_data(settings)` (already there, above) and `get_entry(name) -> Dict[str, Any]` (which just returns the settings for the requested entry.
Separately, as simple functions of the module, there would be a `extract_settings_from_stream(stream)` and `apply_settings_to_stream(settings, stream)`.

I like the second option more because it feels like there is a better separation of concerns. So maybe more easily reused. However, the first approach might just be more straightforward when using it now. So I'm fine with whichever option.

src/odemis/acq/stream_settings.py
```diff
@@ -0,0 +1,174 @@
...
+                    json.dump(self.config_data, jsonfile, indent=2)
+        except Exception as e:
+            logging.warning(f"Error in writing the stream settings: {e}")
+
+    def update_data(self, new_entries: List[Dict[str, Any]]):
+        """
+        Update the saved stream setting with the list of new_entries. First element in JSON file
+        is the latest entry while the last element is the oldest stream setting. The file updates
+        its entries based on the new_entries while maintaining a fixed length of entries in the file.
+        :param new_entries: (list of stream controller settings) Each element in the list represent the setting
```
pieleric, line 93: "stream controller settings" is not correct -> drop the "type" as it's already defined in the signature.

src/odemis/acq/stream_settings.py
```diff
@@ -0,0 +1,174 @@
...
+         of the stream saved as a list from oldest to latest stream setting
```
pieleric, line 94: Please indicate that each entry *must* have a "name" -> str key.

src/odemis/acq/stream_settings.py
```diff
@@ -0,0 +1,174 @@
...
+        self.file_path = file_path
+        self._read()  # update the config_data by reading the JSON file
+        self.settings_order = None  # the ordered list of setting parameters used for loading any given stream
+
+    def _read(self):
+        logging.debug("Reading the most recently used stream settings in JSON")
+        if os.path.exists(self.file_path):
+            try:
+                with open(self.file_path, 'r') as jsonfile:
+                    self.config_data = json.load(jsonfile)
```
pieleric, line 73: There is a lot of JSON files that would not be of the right structure. Please check that config_data is a List[Dict[str,Any]], and if not, log a warning and reset config_data to `[]`.

src/odemis/gui/cont/stream_bar.py
```diff
@@ -477,7 +479,7 @@ def addStream(self, stream, **kwargs):
         """
         return self._add_stream(stream, **kwargs)
 
-    def _add_stream(self, stream, add_to_view=False, visible=True, play=None):
+    def _add_stream(self, stream, add_to_view=False, visible=True, play=None, **kwargs):
```
pieleric, line 482: Drop this `**kwargs`, it is not used by the function.

src/odemis/gui/cont/stream_bar.py
```diff
@@ -1198,6 +1200,51 @@ class SecomStreamsController(StreamBarController):
     Controls the streams for the SECOM and DELPHI live view
```
pieleric, line 1200: Did you check that this works fine also on the DELPHI and SECOM GUI?

src/odemis/gui/cont/stream_bar.py
```diff
@@ -1198,6 +1200,51 @@ class SecomStreamsController(StreamBarController):
...
+        if self.stream_settings.config_data:
+            acq_names = sorted([e["name"] for e in self.stream_settings.config_data])
+            # remove the old list of settings and then update the menu list with the updated stream settings
+            for index, label in enumerate(list(self.menu_actions)):
+                self.remove_action(label)
+            for label in acq_names:
+                call_label = functools.partial(self._acqAddFluo, label)
+                self.add_action(label, call_label)
+
+    def _acqAddFluo(self, label="Filtered colour", **kwargs):
```
pieleric, line 1227: I think it's a pity there is a `addFluo()` function that does almost exactly the same as this function, but you don't use it.
I think you could extend `addFluo()` so that it can handle also what this function does. The menu would call the current function _acqAddFluo(), which would read the settings, and then call `addFluo(label, settings, **kwargs).
If `label=None`, it would do the normal behaviour (of choosing a "Filtered colour" name, and other would use the label as-is. Same with `settings=None`, it would not do any, but otherwise would apply the settings on the newly created stream.

pieleric, line 1227: `acqAddFluo` doesn't really reflect what's special about this function -> `addFluoWithSettings()`?

pieleric (changes requested):

src/odemis/acq/stream_settings.py
K4rishma: In def _read(self) functions, it checks for file path before reading the file. This function is called at init of the class, do I log the exception like below for read function ?
```suggestion
        try:
            self._read()  # update the config_data by reading the JSON file
        except Exception as e:
            logging.exception(f"Error in reading the stream settings: {e}")
```

src/odemis/gui/cont/stream_bar.py
K4rishma: Needs to be checked, not checked. Should I add the microscope files for DELPHI and SECOM in unit tests?

src/odemis/acq/stream_settings.py
pieleric: I think it should be considered normal that the file doesn't exist (as this happens on the first run). So I'd recommand to catch such error separately and just indicate at "info" level. For the other exceptions, "exception" is fine. You just don't need to include the content in the message, as that's what "exception()" does already.

src/odemis/gui/cont/stream_bar.py
pieleric: No, I don't think the unit tests need to be extended. Just manually check that nothing weird happens. Either it's like the feature doesn't exist, or it shows the streams from the history (and also the other types of streams of the SECOM like SEM....)

K4rishma:
ready for review

pieleric (approved):

PR #2644, opened 2024-01-30
[feature] New SPARC Mirror actuator interlock plugin

pieleric:
Check for interlock affecting the SPARC parabolic mirror. If the interlock is triggered, the
"ENGAGE MIRROR" button is disabled. If the mirror is engaged when the interlock is triggered, the
mirror is automatically parked.

Obviously, it needs a interlock. So add a simulated interlock to a SPARC
simulator. Activate the interlock with:
odemis set-attr bsd-interlock interlockTriggered True

### DIFF
--- a/plugins/sparc_mirror_interlock.py
+++ b/plugins/sparc_mirror_interlock.py
@@ -1,0 +1,130 @@
+# -*- coding: utf-8 -*-
+"""
+Created on 30 Jan 2024
+
+@author: Éric Piel
+
+Check for interlock affecting the SPARC parabolic mirror. If the interlock is triggered, the
+"ENGAGE MIRROR" button is disabled. If the mirror is engaged when the interlock is triggered, the
+mirror is automatically parked.
+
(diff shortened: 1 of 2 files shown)


FCanberk (approved):

plugins/sparc_mirror_interlock.py
```diff
@@ -0,0 +1,130 @@
...
+                continue
+            logging.info("Will use %s as interlock for the parabolic mirror", c.name)
+            self._interlocks.append(c)
+            c.interlockTriggered.subscribe(self._on_interlock_change, init=True)
+
+        self._init = False
+
+    def terminate(self):
+        for c in self._interlocks:
+            c.interlockTriggered.subscribe(self._on_interlock_change)
```
sneepstefan, line 78: Do you not rather want to unsubscribe here on termination?

sneepstefan (changes requested):

sneepstefan (approved):

PR #2646, opened 2024-01-30
[config] add config file for starting the delphi GUI

pieleric:
Pretends the simulator stage is alrady calibrated.

### DIFF
--- a/install/linux/usr/share/odemis/sim/calibration.config
+++ b/install/linux/usr/share/odemis/sim/calibration.config
@@ -1,0 +1,18 @@
+# For the DELPHI simulator, so that the sample holder is considered already calibrated
+# To be copied in ~/.config/odemis/
+
+[delphi-499602d2] # fake sample holder for simulator
+top_hole_x = -0.500000000000000
+top_hole_y = 0.000001000000000
+bottom_hole_x = 0.5000000000000000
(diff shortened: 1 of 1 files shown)


install/linux/usr/share/odemis/sim/calibration.config
```diff
@@ -0,0 +1,18 @@
...
+# To be copied in ~/.config/odemis/
+
+[delphi-499602d2] # fake sample holder for simulator
+top_hole_x = -0.500000000000000
+top_hole_y = 0.000001000000000
+bottom_hole_x = 0.5000000000000000
+bottom_hole_y = -0.000001551580000
+stage_trans_x = -0.000474737007096
+stage_trans_y = -0.000142894820906
+stage_scaling_x = 1
```
sneepstefan, line 11: Is this done deliberately that stage_scaling variables are integers at assignment instead of floats?

sneepstefan (approved):

tepals (approved):

install/linux/usr/share/odemis/sim/calibration.config
pieleric: It's not really deliberate. 1 is a good number. That said, maybe you could say that it's even better to test an integer ;-)

PR #2648, opened 2024-01-30
[TRCL-23] support hamamatsu C16910 streak camera

FCanberk:
extend the hamamatsu streak camera driver to support C16910 model.

update hamamatsu streak camera driver to make it less strict on the delaybox. 
with the new model on new system(s) it should be optional.

### DIFF
--- a/src/odemis/driver/hamamatsurx.py
+++ b/src/odemis/driver/hamamatsurx.py
@@ -14,6 +14,7 @@ Odemis is distributed in the hope that it will be useful, but WITHOUT ANY WARRAN
 
 You should have received a copy of the GNU General Public License along with Odemis. If not, see http://www.gnu.org/licenses/.
 """
+
 import collections
 import functools
 import threading
@@ -597,6 +598,7 @@ class StreakUnit(model.HwComponent):
         # [Volt] Input and indication of the trigger level for the vertical sweep.
         parent.DevParamSet(self.location, "Trig. level", 1)  # TODO check what value needed regarding HW
         parent.DevParamSet(self.location, "Trig. slope", "Rising")
+        parent.DevParamSet(self.location, "Shutter", "Closed")
 
         # parent.DevParamGet(self.location, "Trig. status")  # read only
 
@@ -634,6 +636,9 @@ class StreakUnit(model.HwComponent):
         # is set in the setter of the timeRange VA
         self.timeRangeFactor = None
 
+        shutter = self.GetShutter()
+        self.shutter = model.BooleanVA(shutter, setter=self._setShutter)
+
         # read-only VAs TODO: Trig. Mode, Trig. level, Trig. slope?
 
     def GetPLLMode(self):
@@ -669,6 +674,34 @@ class StreakUnit(model.HwComponent):
         else:
             self.parent.DevParamSet(self.location, "PLL Mode", "Unlocked")
         logging.debug("Phase-locked loop mode is set to: %s", value)
+
+        return value
+
+    def GetShutter(self) -> bool:
+        """
+        Get the current value from the trigger delay HW (RemoteEx: delay A).
+        :return: (bool) current state of the shutter
+        """
+        shutter_raw = self.parent.DevParamGet(self.location, "Shutter")  # returns a list
+        if shutter_raw[0] == "Open":
+            shutter = True
+            self.state._set_value(model.ST_RUNNING, force_write=True)
+        elif shutter_raw[0] == "Closed":
+            shutter = False
+        else:
+            logging.warning("Unexpected streak mode %s", shutter_raw)
+            shutter = False  # safer!
+
+        return shutter
+
+    def _setShutter(self, value: bool) -> bool:
+        """
+        Updates the shutter state VA.
+        :param value: value to be set
+        :return: (bool) shutter state
+        """
+        self.parent.DevParamSet(self.location, "Shutter", "Open" if value else "Closed")
+        logging.debug("Reporting mechanical shutter state: %s.", value)
 
         return value
 
--- a/src/odemis/driver/hamamatsurx.py
+++ b/src/odemis/driver/hamamatsurx.py
@@ -600,6 +600,11 @@ class StreakUnit(model.HwComponent):
         parent.DevParamSet(self.location, "Trig. slope", "Rising")
         parent.DevParamSet(self.location, "Shutter", "Closed")
 
+        # Refresh regularly the values, from the hardware, starting from now
+        self._updateSettings()
+        self._va_poll = util.RepeatingTimer(5, self._updateSettings, "Streak unit settings polling")
+        self._va_poll.start()
+
         # parent.DevParamGet(self.location, "Trig. status")  # read only
 
         # Ready: Is displayed when the system is ready to receive a trigger signal.
@@ -640,6 +645,36 @@ class StreakUnit(model.HwComponent):
         self.shutter = model.BooleanVA(shutter, setter=self._setShutter)
 
         # read-only VAs TODO: Trig. Mode, Trig. level, Trig. slope?
+
+    def _updateSettings(self) -> None:
+        """
+        Read all the current streak unit settings from the RemoteEx and reflect them on the VAs
+        """
+        logging.debug("Updating streak unit settings")
+        try:
+            timeRange = self.GetTimeRange()
+            if timeRange != self.timeRange.value:
+                self.timeRange._value = timeRange
+                self.timeRange.notify(timeRange)
+
+            gain = self.GetMCPGain()
+            if gain != self.MCPGain.value:
+                self.MCPGain._value = gain
+                self.MCPGain.notify(gain)
+                logging.debug("MCP Gain synced with RemoteEx.")
+
+            mode = self.GetStreakMode()
+            if mode != self.streakMode.value:
+                self.streakMode._value = mode
+                self.streakMode.notify(mode)
+
+            pllMode = self.GetPLLMode()
+            if pllMode != self.phaseLocked.value:
+                self.phaseLocked._value = pllMode
+                self.phaseLocked.notify(pllMode)
+
+        except Exception:
+            logging.exception("Unexpected failure when polling streak unit settings")
 
     def GetPLLMode(self):
         """
@@ -885,6 +920,11 @@ class DelayGenerator(model.HwComponent):
         # Note: trigger rate (repetition rate) corresponds to the ebeam blanking frequency (read only in RemoteEx)
         self._metadata[model.MD_TRIGGER_RATE] = int(self.parent.DevParamGet(self.location, "Repetition Rate")[0])
 
+        # Refresh regularly the values, from the hardware, starting from now
+        self._updateSettings()
+        self._va_poll = util.RepeatingTimer(5, self._updateSettings, "Delay generator settings polling")
+        self._va_poll.start()
+
         # VAs
         self._subscribers = []
         devparamslist = parent.DevParamsList()
@@ -975,6 +1015,25 @@ class DelayGenerator(model.HwComponent):
         range_time = (min_time, max_time)
 
         return range_time
+
+    def _updateSettings(self) -> None:
+        """
+        Read all the current delay generator settings from the RemoteEx and reflect them on the VAs
+        """
+        logging.debug("Updating streak unit settings")
+        try:
+            delaynameremoteex = ["Delay A", "Delay B", "Delay C", "Delay D", "Delay E", "Delay F", "Delay G", "Delay H"]
+            delayvas = ["triggerDelay", "delayB", "delayC", "delayD", "delayE", "delayF", "delayG", "delayH"]
+            for i in range(0, 7):
+                if hasattr(self, delayvas[i]):
+                    delayva = getattr(self, delayvas[i])
+                    delayre = self.GetDelayByName(delaynameremoteex[i])
+                    if delayva._value != delayre:
+                        delayva = getattr(self, delayva)
+                        delayva._value = delayre
+
+        except Exception:
+            logging.exception("Unexpected failure when polling delay generator settings")
 
 
 class StreakCamera(model.HwComponent):
--- a/src/odemis/driver/hamamatsurx.py
+++ b/src/odemis/driver/hamamatsurx.py
@@ -51,6 +51,9 @@ PARAM_TYPE_STRING = 3
 PARAM_TYPE_EXPTIME = 4
 PARAM_TYPE_DISPLAY = 5
 
+DELAY_NAME_REMOTE_EX = ["Delay A", "Delay B", "Delay C", "Delay D", "Delay E", "Delay F", "Delay G", "Delay H"]
+DELAY_NAME_VAS = ["triggerDelay", "delayB", "delayC", "delayD", "delayE", "delayF", "delayG", "delayH"]
+
 
 class RemoteExError(IOError):
 
@@ -125,16 +128,17 @@ class ReadoutCamera(model.DigitalCamera):
         self._metadata[model.MD_SW_VERSION] = self._swVersion
         self._metadata[model.MD_DET_TYPE] = model.MD_DT_INTEGRATING
 
+        # TODO: Check what values to set
         # Set parameters readout camera
-        parent.CamParamSet("Setup", "TimingMode", "Internal timing")  # TODO external check displayed command in GUI
-        # parent.CamParamSet("Setup", "TriggerMode", 'Edge trigger')
-        # parent.CamParamSet("Setup", "TriggerSource", 'BNC')
-        # parent.CamParamSet("Setup", "TriggerPolarity", 'neg.')
-        parent.CamParamSet("Setup", "ScanMode", 'Subarray')
-        parent.CamParamSet("Setup", "Binning", '2 x 2')
-        parent.CamParamSet("Setup", "VWidth", '1016')
-        parent.CamParamSet("Setup", "HWidth", '1344')
-        parent.CamParamSet("Setup", "ShowGainOffset", 'True')
+        # parent.CamParamSet("Setup", "TimingMode", "Internal timing")  # TODO external check displayed command in GUI
+        # # parent.CamParamSet("Setup", "TriggerMode", 'Edge trigger')
+        # # parent.CamParamSet("Setup", "TriggerSource", 'BNC')
+        # # parent.CamParamSet("Setup", "TriggerPolarity", 'neg.')
+        # parent.CamParamSet("Setup", "ScanMode", 'Subarray')
+        # parent.CamParamSet("Setup", "Binning", '2 x 2')
+        # # parent.CamParamSet("Setup", "VWidth", '1016')
+        # # parent.CamParamSet("Setup", "HWidth", '1344')
+        # parent.CamParamSet("Setup", "ShowGainOffset", 'True')
 
         # sensor size (resolution)
         # Note: sensor size of OrcaFlash is actually much larger (2048px x 2048px)
@@ -585,21 +589,21 @@ class StreakUnit(model.HwComponent):
         self.parent = parent
         self.location = "Streakcamera"  # don't change, internally needed by HPDTA/RemoteEx
 
-        self._hwVersion = parent.DevParamGet(self.location, "DeviceName")[0]   # needs to be a string
+        self._hwVersion = parent.DevParamGet(self.location, "DeviceName")[0]  # needs to be a string
         self._metadata[model.MD_HW_VERSION] = self._hwVersion
 
         # Set parameters streak unit
-        parent.DevParamSet(self.location, "Time Range", "1 ns")
+        # parent.DevParamSet(self.location, "Time Range", "1 ns")  # TODO: check what value to set
         parent.DevParamSet(self.location, "MCP Gain", 0)
         # Switch Mode to "Focus", MCPGain = 0 (implemented in RemoteEx and also here in the driver).
         parent.DevParamSet(self.location, "Mode", "Focus")
         # Resets behavior for a vertical single shot sweep: Automatic reset occurs after each sweep.
-        parent.DevParamSet(self.location, "Trig. Mode", "Cont")
+        # parent.DevParamSet(self.location, "Trig. Mode", "Cont")  # TODO: check what value to set
         # [Volt] Input and indication of the trigger level for the vertical sweep.
-        parent.DevParamSet(self.location, "Trig. level", 1)  # TODO check what value needed regarding HW
-        parent.DevParamSet(self.location, "Trig. slope", "Rising")
+        # parent.DevParamSet(self.location, "Trig. level", 1)  # TODO check what value needed regarding HW
+        # parent.DevParamSet(self.location, "Trig. slope", "Rising")  # TODO: check what value to set
 
-        devparamslist = parent.DevParamsList()
+        devparamslist = parent.DevParamsList("TD")
         # only add the shutter parameter if it is possible to control the shutter via software
         if "Shutter" in devparamslist:
             parent.DevParamSet(self.location, "Shutter", "Closed")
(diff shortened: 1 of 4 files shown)


K4rishma (approved):

src/odemis/driver/hamamatsurx.py
```diff
@@ -833,9 +845,40 @@ def updateMetadata(self, md):
...
+        return pllMode
+
+    def _setPLLMode(self, value):
+        """
+        Updates the streakMode VA.
+        :parameter value: (bool) value to be set
+        :return: (bool) current streak mode
+        """
+        if value:
+            self.phaseLocked.value = True
```
pieleric, line 871: This is the VA itself. Don't change it in the setter function. The return value will set the VA value.

src/odemis/driver/hamamatsurx.py
```diff
@@ -833,9 +845,40 @@ def updateMetadata(self, md):
...
+            pllMode = True
+        elif pllMode_raw[0] == "Unlocked":
+            pllMode = False
+        else:
+            logging.warning("Unexpected phase-locked loop mode %s", pllMode_raw)
+            pllMode = True
+
+        return pllMode
+
+    def _setPLLMode(self, value):
```
pieleric, line 864: Please indicate types in the signature (in every new functions).

src/odemis/driver/hamamatsurx.py
```diff
@@ -807,15 +807,27 @@ def __init__(self, name, role, parent, daemon=None, **kwargs):
...
         parent.DevParamSet(self.location, "Delay B", 0.00000002)
         parent.DevParamSet(self.location, "Burst Mode", "Off")
+        parent.DevParamSet(self.location, "PLL Mode", "Locked")  # Set phase locked loop mode, only used in synchroscan
 
         self._metadata[model.MD_TRIGGER_DELAY] = self.GetTriggerDelay()
         # Note: trigger rate (repetition rate) corresponds to the ebeam blanking frequency (read only in RemoteEx)
         self._metadata[model.MD_TRIGGER_RATE] = int(self.parent.DevParamGet(self.location, "Repetition Rate")[0])
 
+        # VAs
+        pllMode = self.GetPLLMode()
```
pieleric, line 817: It should only add the `phaseLocked` VA if the parameter exists.

src/odemis/driver/hamamatsurx.py
```diff
@@ -807,15 +807,27 @@ def __init__(self, name, role, parent, daemon=None, **kwargs):
         parent.DevParamSet(self.location, "Delay A", 0)
         parent.DevParamSet(self.location, "Delay B", 0.00000002)
         parent.DevParamSet(self.location, "Burst Mode", "Off")
+        parent.DevParamSet(self.location, "PLL Mode", "Locked")  # Set phase locked loop mode, only used in synchroscan
```
pieleric, line 810: Don't change it by default. As it's available as a VA, if it turns out that it's convenient to have it to a specific value by default, we can set it from the microscope file.

src/odemis/driver/hamamatsurx.py
```diff
@@ -807,15 +807,27 @@ def __init__(self, name, role, parent, daemon=None, **kwargs):
...
         triggerDelay = self.GetTriggerDelay()
         range_trigDelay = self._getTriggerDelayRange()
         self.triggerDelay = model.FloatContinuous(triggerDelay, range_trigDelay, setter=self._setTriggerDelay, unit="s")
 
+        # triggerDelay is Delay A, that's why delays start with Delay B
+        delays= ["Delay B", "Delay C", "Delay D", "Delay E", "Delay F", "Delay G", "Delay H"]
+        for d in delays:
+            delay = self.GetDelayByName(d)
+            range_delay = self._getDelayRangeByName(d)
+            self.delay = model.FloatContinuous(delay, range_delay, setter=self._setDelayByName, unit="s")
```
pieleric, line 829: You're overriding the VA for every parameter! There should be one VA per parameter, named `delayB`, `delayC`..., `delayH`. => use `setattr()`.
It should also only happen on the component which have such parameters. (So, no on the synchroscan delay unit)

src/odemis/driver/hamamatsurx.py
```diff
@@ -833,9 +845,40 @@ def updateMetadata(self, md):
...
+        Get the current value from the streak unit HW.
+        :return: (bool) current streak mode value
+        """
+        pllMode_raw = self.parent.DevParamGet(self.location, "PLL Mode")  # returns a list
+        if pllMode_raw[0] == "Locked":
+            pllMode = True
+        elif pllMode_raw[0] == "Unlocked":
+            pllMode = False
+        else:
+            logging.warning("Unexpected phase-locked loop mode %s", pllMode_raw)
```
pieleric, line 859: You mentioned it has 5 options right? Some related to errors. You should expect them as well and handle them nicely. As discussed, the driver should report the errors a .state error (+ PLL mode).

To do so, I think you have 2 options:
* You have 2 functions: one which is a one-to-one mapping for the "PLL Mode" parameter, and a second "higher level" function which interprets this return value as boolean + state. 
* You have a single function which directly read this parameter and converts to boolean + state.
There should be a

src/odemis/driver/hamamatsurx.py
```diff
@@ -879,6 +922,42 @@ def _updateTriggerRate(self):
         triggerRate = float(triggerRate_raw[0])
         self._metadata[model.MD_TRIGGER_RATE] = triggerRate
 
+    def GetDelayByName(self, delay_name: str):
```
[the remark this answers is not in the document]
pieleric: I think it should be named `_getDelayByName`, to follow the same style as the other two functions.

src/odemis/driver/hamamatsurx.py
```diff
@@ -833,9 +845,40 @@ def updateMetadata(self, md):
...
+        if value:
+            self.phaseLocked.value = True
+            self.parent.DevParamSet(self.location, "PLL Mode", "Locked")
+        else:
+            self.parent.DevParamSet(self.location, "PLL Mode", "Unlocked")
+        logging.debug("Phase-locked loop mode is set to: %s", value)
+
+        return value
+
     def GetTriggerDelay(self):
```
pieleric, line 879: It should be possible to drop these 3 specific functions for `triggerDelay`, and only use the `*DelayByName`.

src/odemis/driver/test/hamamatsurx_test.py
```diff
@@ -108,6 +115,96 @@ def tearDownClass(cls):
...
+            if tr2d:
+                key = util.find_closest(timeRange, tr2d.keys())
+                # check that the corresponding trigger delay is set when changing the .timeRange VA
+                md_triggerDelay = tr2d[key]
+                self.assertAlmostEqual(self.delaybox.triggerDelay.value, md_triggerDelay)
+
+        # request value, which is not in choices of VA
+        with self.assertRaises(IndexError):
+            self.streakunit.timeRange.value = 0.000004  # 4us
+
```
pieleric, line 207: Add a test case for delays B to H.
(Maybe, skip it if the VAs are missing, so that it works with the synchroscan delay unit)

src/odemis/driver/test/hamamatsurx_test.py
```diff
@@ -108,6 +115,96 @@ def tearDownClass(cls):
...
+            tr2d = self.delaybox._metadata.get(model.MD_TIME_RANGE_TO_DELAY)
+            if tr2d:
+                key = util.find_closest(timeRange, tr2d.keys())
+                # check that the corresponding trigger delay is set when changing the .timeRange VA
+                md_triggerDelay = tr2d[key]
+                self.assertAlmostEqual(self.delaybox.triggerDelay.value, md_triggerDelay)
+
+        # request value, which is not in choices of VA
+        with self.assertRaises(IndexError):
+            self.streakunit.timeRange.value = 0.000004  # 4us
```
pieleric, line 206: Add a test case for the PLL mode.
(And skip if the VA is not present)

src/odemis/driver/test/hamamatsurx_test.py
```diff
@@ -108,6 +115,96 @@ def tearDownClass(cls):
...
+                5.e-3: 7.88e-3,
+                10.e-3: 15.4e-3,
+            }
+        })
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.streakcam.terminate()
+
+    def test_error(self):  # TODO more testcases
```
pieleric, line 166: Add more test cases and drop this `TODO`.

pieleric (changes requested):

src/odemis/driver/hamamatsurx.py
```diff
@@ -879,6 +922,42 @@ def _updateTriggerRate(self):
         triggerRate = float(triggerRate_raw[0])
         self._metadata[model.MD_TRIGGER_RATE] = triggerRate
 
+    def GetDelayByName(self, delay_name: str):
+        """
+        Get the current value from the trigger delay HW (RemoteEx: delay D).
+        :parameter delay_name: name of the delay e.g. (Delay H)
```
K4rishma, line 928: same here :param

src/odemis/driver/hamamatsurx.py
```diff
@@ -879,6 +922,42 @@ def _updateTriggerRate(self):
...
+        """
+        delay_raw = self.parent.DevParamGet(self.location, delay_name)  # returns a list
+        delay = float(delay_raw[0])
+
+        return delay
+
+    def _setDelayByName(self, delay_name: str, value):
+        """
+        Updates the trigger delay VA.
+        :parameter delay_name: (string) name of the delay e.g. (Delay A)
```
K4rishma, line 939: here

K4rishma (approved):

src/odemis/driver/hamamatsurx.py
```diff
@@ -833,9 +845,40 @@ def updateMetadata(self, md):
...
+        else:
+            logging.warning("Unexpected phase-locked loop mode %s", pllMode_raw)
+            pllMode = True
+
+        return pllMode
+
+    def _setPLLMode(self, value):
+        """
+        Updates the streakMode VA.
+        :parameter value: (bool) value to be set
```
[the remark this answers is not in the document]
FCanberk: I typed parameter instead of param because all the doc strings were written that way in this module, but let me fix all of them, that would be better :)

src/odemis/driver/hamamatsurx.py
```diff
@@ -597,6 +598,7 @@ def __init__(self, name, role, parent, daemon=None, **kwargs):
         # [Volt] Input and indication of the trigger level for the vertical sweep.
         parent.DevParamSet(self.location, "Trig. level", 1)  # TODO check what value needed regarding HW
         parent.DevParamSet(self.location, "Trig. slope", "Rising")
+        parent.DevParamSet(self.location, "Shutter", "Closed")
```
pieleric, line 601: This needs to work with the old hardware too, which doesn't have a Shutter (controllable in software).  I guess in this case, the shutter parameter will not be available. Please check with the (old) simulator. Ideally, the driver should detect the Shutter is not controllable, and just not provide a VA.

src/odemis/driver/test/hamamatsurx_test.py
```diff
@@ -418,6 +418,21 @@ def test_TriggerDelay(self):
         with self.assertRaises(IndexError):
             self.delaybox.triggerDelay.value = -1
 
+    def test_Shutter(self):
+        """Test Shutter state VA of streak unit."""
+        # set shutter VA
+        self.streakunit.shutter.value = False
+        time.sleep(0.5)  # give it some time to actually change the value
```
pieleric, line 425: That shouldn't be necessary. The setter is blocking. The VA value will be updated once the setter returns.

src/odemis/driver/hamamatsurx.py
```diff
@@ -672,6 +677,34 @@ def _setPLLMode(self, value: bool):
...
+        return shutter
+
+    def _setShutter(self, value: bool) -> bool:
+        """
+        Updates the shutter state VA.
+        :param value: value to be set
+        :return: (bool) shutter state
+        """
+        self.parent.DevParamSet(self.location, "Shutter", "Open" if value else "Closed")
+        logging.debug("Reporting mechanical shutter state: %s.", value)
```
pieleric, line 739: I don't think this sentence is matching. I'd just write "Setting shutter to ...".

src/odemis/driver/hamamatsurx.py
```diff
@@ -672,6 +677,34 @@ def _setPLLMode(self, value: bool):
...
+
+        return shutter
+
+    def _setShutter(self, value: bool) -> bool:
+        """
+        Updates the shutter state VA.
+        :param value: value to be set
+        :return: (bool) shutter state
+        """
+        self.parent.DevParamSet(self.location, "Shutter", "Open" if value else "Closed")
```
pieleric, line 738: Typically it's best for the rest of Odemis to be certain that the shutter is actually closed/opened by the time this setter function returns.
For the acquisition to work reliably, we need to make sure the shutter is actually opened otherwise, Odemis might start acquisition while the shutter is opening. That said, typically the exposure time is quite long (> 1s), and the shutter opening time is most likely < 0.1s, so that might have very little effect. 

In the other direction, it's safer to make sure the shutter is closed, before doing "something" else, which might involve a lot of light.
So, please have a look at whether the function returns after the move is complete, or you should add a small `sleep()`.

src/odemis/driver/hamamatsurx.py
```diff
@@ -641,6 +646,36 @@ def __init__(self, name, role, parent, daemon=None, **kwargs):
...
+
+            mode = self.GetStreakMode()
+            if mode != self.streakMode.value:
+                self.streakMode._value = mode
+                self.streakMode.notify(mode)
+
+            pllMode = self.GetPLLMode()
+            if pllMode != self.phaseLocked.value:
+                self.phaseLocked._value = pllMode
+                self.phaseLocked.notify(pllMode)
```
pieleric, line 674: This should be done only if the phaseLocked VA is present.

src/odemis/driver/hamamatsurx.py
```diff
@@ -641,6 +646,36 @@ def __init__(self, name, role, parent, daemon=None, **kwargs):
...
+            if gain != self.MCPGain.value:
+                self.MCPGain._value = gain
+                self.MCPGain.notify(gain)
+                logging.debug("MCP Gain synced with RemoteEx.")
+
+            mode = self.GetStreakMode()
+            if mode != self.streakMode.value:
+                self.streakMode._value = mode
+                self.streakMode.notify(mode)
+
```
pieleric, line 670: Please update all the VAs, so also the shutter.

src/odemis/driver/hamamatsurx.py
```diff
@@ -641,6 +646,36 @@ def __init__(self, name, role, parent, daemon=None, **kwargs):
...
+    def _updateSettings(self) -> None:
+        """
+        Read all the current streak unit settings from the RemoteEx and reflect them on the VAs
+        """
+        logging.debug("Updating streak unit settings")
+        try:
+            timeRange = self.GetTimeRange()
+            if timeRange != self.timeRange.value:
+                self.timeRange._value = timeRange
+                self.timeRange.notify(timeRange)
```
pieleric, line 658: For several VAs, you'd need to update the metadata (see line 617 - 619). Probably the simplest would be to separate the setter from the metadata update by placing the metadata update in a subscriber.

src/odemis/driver/hamamatsurx.py
```diff
@@ -833,51 +961,79 @@ def updateMetadata(self, md):
...
-        triggerRate_raw = self.parent.DevParamGet(self.location, "Repetition Rate")  # returns a list
-        triggerRate = float(triggerRate_raw[0])
-        self._metadata[model.MD_TRIGGER_RATE] = triggerRate
+        logging.debug("Updating streak unit settings")
+        try:
+            delaynameremoteex = ["Delay A", "Delay B", "Delay C", "Delay D", "Delay E", "Delay F", "Delay G", "Delay H"]
+            delayvas = ["triggerDelay", "delayB", "delayC", "delayD", "delayE", "delayF", "delayG", "delayH"]
+            for i in range(0, 7):
+                if hasattr(self, delayvas[i]):
+                    delayva = getattr(self, delayvas[i])
```
pieleric, line 1029:
```suggestion
            for i, delayva in enumerate(delayvas):
                if hasattr(self, delayva):
```

pieleric (changes requested):
This is a partial review.

src/odemis/driver/hamamatsurx.py
```diff
@@ -625,15 +640,122 @@ def __init__(self, name, role, parent, daemon=None, **kwargs):
...
+            timeRange = self.GetTimeRange()
+            if timeRange != self.timeRange.value:
+                self.timeRange._value = timeRange
+                self.timeRange.notify(timeRange)
+
+            gain = self.GetMCPGain()
+            if gain != self.MCPGain.value:
+                self.MCPGain._value = gain
+                self.MCPGain.notify(gain)
+                logging.debug("MCP Gain synced with RemoteEx.")
```
pieleric, line 668: Let's drop this log, as the other settings don't have a log entry, and I don't think it's really important per se to know.

pieleric (changes requested):

src/odemis/driver/hamamatsurx.py
```diff
@@ -709,39 +868,40 @@ def GetTimeRange(self):
...
-                logging.warning("Time range %s is not a key in MD for time range to "
-                                "trigger delay calibration" % value)
+        # tr2d = self.parent._delaybox._metadata.get(model.MD_TIME_RANGE_TO_DELAY)
+        # if tr2d:
+        #     key = util.find_closest(value, tr2d.keys())
+        #     if util.almost_equal(key, value):
+        #         self.parent._delaybox.triggerDelay.value = tr2d[key]
+        #     else:
+        #         logging.warning("Time range %s is not a key in MD for time range to "
+        #                         "trigger delay calibration" % value)
```
pieleric, line 886: That shouldn't be disabled.

src/odemis/driver/hamamatsurx.py
```diff
@@ -583,19 +589,32 @@ def __init__(self, name, role, parent, daemon=None, **kwargs):
         self.parent = parent
         self.location = "Streakcamera"  # don't change, internally needed by HPDTA/RemoteEx
 
-        self._hwVersion = parent.DevParamGet(self.location, "DeviceName")[0]   # needs to be a string
+        self._hwVersion = parent.DevParamGet(self.location, "DeviceName")[0]  # needs to be a string
         self._metadata[model.MD_HW_VERSION] = self._hwVersion
 
         # Set parameters streak unit
-        parent.DevParamSet(self.location, "Time Range", "1 ns")
+        # parent.DevParamSet(self.location, "Time Range", "1 ns")  # TODO: check what value to set
```
pieleric, line 596: We probably can skip this, as this is handled by a VA.

src/odemis/driver/hamamatsurx.py
```diff
@@ -625,15 +640,159 @@ def __init__(self, name, role, parent, daemon=None, **kwargs):
         timeRange = util.find_closest(timeRange, choices)  # make sure value is in choices
         self.timeRange = model.FloatEnumerated(timeRange, choices, setter=self._setTimeRange, unit="s")
```
pieleric, line 641: If I understand correctly, there is a difficult here, because the synchroscan unit doesn't provide a time range of the same unit as the single sweet unit. While the later one provides it in s, the former provides in "arbitrary values" 1,2,3,4... I've discussed with Herman, and apparently, for a given system, the arbitrary values can be fixed to a specific time. So, we could add an argument to the component, which allows to pass for each value the actual time range. Something like `time_ranges = {1: 2e-12, 2: 5e-12, 3: 8.9e-12 ...}`. It's only accepted if the time range has no unit. Then, we can have the VA always in s, and when it's used with the synchroscan, it's converted back and forth with the arbitrary values.

src/odemis/driver/hamamatsurx.py
```diff
@@ -753,7 +913,10 @@ def _getStreakUnitTimeRangeChoices(self):
         choices = []
         for choice in choices_raw:
             choice_raw = choice.split(" ")
-            choices.append(self.parent.convertUnit2Time(choice_raw[0], choice_raw[1]))
+            if len(choice_raw) == 1:
+                choices.append(int(choice_raw[0]))
+            else:
+                choices.append(self.parent.convertUnit2Time(choice_raw[0], choice_raw[1]))
```
pieleric, line 919: Please explain that the len==1 means there is no unit.
=> Need to convert to the corresponding (real) time value afterwards.

src/odemis/gui/cont/settings.py
```diff
@@ -212,11 +211,11 @@ def add_metadata(self, key, value):
             else:
                 # Still try to beautify a bit if it's a number
                 if (
-                    isinstance(value, (int, long, float)) or
```
pieleric, line 215: You shouldn't have these changes related to "long". They are already in the latest master. (But I guess git can handle them fine so not need to worry too much)

src/odemis/gui/cont/settings.py
```diff
@@ -897,17 +896,17 @@ def __init__(self, tab_panel, tab_data):
...
+        #                                           self.streak_delay.triggerDelay,
+        #                                           self.streak_delay,
+        #                                           conf={"control_type": odemis.gui.CONTROL_FLT,
+        #                                                 "label": "Trigger delay",
+        #                                                 "tooltip": "Change the trigger delay value to "
+        #                                                            "center the image."},
+        #                                           change_callback=self._onUpdateTriggerDelayMD)
+        #
+        # entry_triggerDelay.value_ctrl.SetBackgroundColour(odemis.gui.BG_COLOUR_PANEL)
+        # self.ctrl_triggerDelay = entry_triggerDelay.value_ctrl
```
pieleric, line 909: Should be put back.

src/odemis/gui/cont/settings.py
```diff
@@ -929,48 +928,48 @@ def __init__(self, tab_panel, tab_data):
...
-
-    def _onUpdateTriggerDelayGUI(self, text, colour=odemis.gui.FG_COLOUR_EDIT):
-        """
-        Updates the GUI elements regarding the new trigger delay value.
-        :parameter text (str): the text to show
-        :parameter colour (wx.Colour): the colour to use
-        """
-        self.panel.txt_StreakCalibFilename.Value = text
-        self.panel.txt_StreakCalibFilename.SetForegroundColour(colour)
+    # def _onUpdateTriggerDelayMD(self, evt):
```
pieleric, line 931: Should be put back.

src/odemis/gui/cont/settings.py
```diff
@@ -996,26 +995,26 @@ def _onOpenCalibFile(self, event):
...
-
-        # update triggerDelay shown in GUI
-        cur_timeRange = self.streak_unit.timeRange.value
-        # find the corresponding trigger delay
-        key = util.find_closest(cur_timeRange, tr2d.keys())
-        # Note: no need to check almost_equal again as we do that already when loading the file
-        self.streak_delay.triggerDelay.value = tr2d[key]  # set the new value
-
-        self._onUpdateTriggerDelayGUI(filename)  # update txt displayed in GUI
+        # try:
```
pieleric, line 998: Should be put back.

src/odemis/gui/cont/settings.py
```diff
@@ -1046,9 +1045,9 @@ def _onSaveCalibFile(self, event):
             filename += ".csv"
             path += ".csv"
 
-        # get a copy of the triggerDelay dict from MD
-        tr2d = self.streak_delay.getMetadata()[model.MD_TIME_RANGE_TO_DELAY]
-        calibration.write_trigger_delay_csv(path, tr2d)
-
-        # update txt displayed in GUI
-        self._onUpdateTriggerDelayGUI(filename)
+        # # get a copy of the triggerDelay dict from MD
```
pieleric, line 1048: Should be put back.

pieleric (changes requested):

pieleric:
Extended by PR #2824

PR #2656, opened 2024-02-05
[feature] Add MODEL_STRIC_SYNTAX to the backend settings

pieleric:
Allows to force the backend to only accept new syntax for microscope
files. This is handy to check that new files don't rely on deprecated
syntax.

For now, only uses one (new) flag: --strict-children, which forces the
proper distinction between children and dependencies.


src/odemis/odemisd/modelgen.py
```diff
@@ -340,21 +344,41 @@ def _fill_creator(self):
         Add the "parents" field (= reverse of children) and update the creator
          field for the components that don't have it explicitly set.
         """
-        # update the children by adding a "parents" attribute
         for name, comp in self.ast.items():
-            references = list(comp.get("children", {}).values()) + list(comp.get("dependencies", {}).values())
-            for ref in references:
+            # Check the dependencies are not crazy
```
FCanberk, line 348: maybe a bit more explanatory comment rather than crazy? :laughing:

FCanberk (approved):

FCanberk (approved):

src/odemis/odemisd/modelgen.py
```diff
@@ -245,6 +245,9 @@ def __init__(self, inst_file, settings_file=None, container=None, create_sub_con
           model without actually any driver contacting the hardware. It will also
           be stricter, and some issues which are normally just warnings will be
           considered errors.
+        strict_children: If True, only can use "children" for components whithout class and created
```
nandishjpatel, line 248:
```suggestion
        strict_children: If True, only can use "children" for components without class and created
```

PR #2657, opened 2024-02-06
[feat] Allow setting the flyback of the descanner at init

tepals:
The value for the flyback might be different for different systems, this allows us to set it in the yaml file


pieleric:
If it's already a VA, you typically don't need to pass as argument. you can just use the `properties` option in the microscope file to automatically change the VA value just after initialization. Like:
```
    properties: {
        physicalFlybackTime: 170.e-6,  # s
    },
```

tepals:
I knew I was forgetting something, I'll close this PR then.

PR #2658, opened 2024-02-06
[SDEV-1768] feature fastem: add option to save full cell images

tepals:


src/odemis/gui/plugin/__init__.py
```diff
@@ -221,7 +221,7 @@ def terminate(self):
         pass
 
     @call_in_wx_main
-    def addMenu(self, entry, callback):
+    def addMenu(self, entry, callback, check_item=False):
```
nandishjpatel, line 224:
```suggestion
    def addMenu(self, entry, callback, entry_kind=wx.ITEM_NORMAL):
```

nandishjpatel: Docstring needs to be updated

src/odemis/gui/plugin/__init__.py
```diff
@@ -282,7 +282,8 @@ def addMenu(self, entry, callback):
 
         # TODO: if adding for the first time to standard menu, first add a separator
         # Add the menu item
-        menu_item = curr_group.Append(wx.ID_ANY, path[-1])
+        item_kind = wx.ITEM_CHECK if check_item else wx.ITEM_NORMAL
+        menu_item = curr_group.Append(wx.ID_ANY, path[-1],  kind=item_kind)
```
nandishjpatel, line 286:
```suggestion
        menu_item = curr_group.Append(wx.ID_ANY, path[-1],  kind=entry_kind)
```

plugins/fastem_save_full_cell_images.py
```diff
@@ -0,0 +1,64 @@
...
+            logging.debug(
+                "Not loading Save full cell images tool since acquisition tab is not present."
+            )
+            return
+
+        self._acquisition_controller = self._acquisition_tab._acquisition_controller
+
+        self.addMenu("Help/Development/Save full cell images",
+                     self._save_full_cell_images,
+                     check_item=True)
```
nandishjpatel, line 52:
```suggestion
                     entry_kind=wx.ITEM_CHECK)
```

src/odemis/gui/plugin/__init__.py
```diff
@@ -294,6 +295,7 @@ def menu_callback_wrapper(evt):
                 logging.exception("Error when processing menu entry %s of plugin %s",
                                   path[-1], self)
         main_frame.Bind(wx.EVT_MENU, menu_callback_wrapper, id=menu_item.Id)
+        self.menu_group = curr_group
```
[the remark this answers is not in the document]
tepals: I tried to do that, but then it just returns None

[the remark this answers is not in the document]
pieleric: I totally agree that it's not a good solution. Actually, reading the code  you don't really care about the group. You care about the menu_item itself, right? As this function is decorated by `@call_in_wx_main`, it's called asynchronously, and you cannot receive its return value.
Having the menu_group or menu_item directly on the Plugin doesn't work well because this function can be called multiple times, to create multiple menu entries. 
I can think of two alternatives:
* Have a attribute on Plugin `.menu_items: Dict[str, MenuItem]` which maps the requested (full path) menu name to the actual menu item object.
* Have a optional argument to the function `pass_menu_item=False`, so that if it's True, the menu_item will be pass as argument to the callback.

I think I like a little bit more the second alternative, but I let you decide. Or do you have something else to suggest?

plugins/fastem_save_full_cell_images.py
```diff
@@ -0,0 +1,64 @@
...
+            self._acquisition_tab = fastem_main_tab.acquisition_tab
+        except LookupError:
+            logging.debug(
+                "Not loading Save full cell images tool since acquisition tab is not present."
+            )
+            return
+
+        self._acquisition_controller = self._acquisition_tab._acquisition_controller
+
+        self.addMenu("Help/Development/Save full cell images",
```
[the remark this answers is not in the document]
tepals: Yes for now it is only for debugging, and for expert users to experiment with. For now we don't know if we want to offer this to everyone.

pieleric (approved):
Looks good to me. Could you squash your last commit into the first 3 commits, so that we can merge a clean history?

nandishjpatel (approved):

PR #2664, opened 2024-02-07
[FIX] PARMA stage scanning fixes

sneepstefan:
After recent testing of the stage scanning feature on a real SEM with a stage scan wrapper it turned out a few fixes were needed:
- there needed to be made a few distinctions for automatic moving conditions of the stage, in general with a real stage we do not want to move when starting Odemis or to go back to the center position.
- correction for stage shift when acquiring
- added shift stage translation fake data to simsem detector class

### DIFF
--- a/src/odemis/acq/stream/_sync.py
+++ b/src/odemis/acq/stream/_sync.py
@@ -1737,15 +1737,22 @@ class SEMCCDMDStream(MultipleDetectorStream):
                     if self._acq_state == CANCELLED:
                         raise CancelledError()
 
-                    # MD_POS default to the center of the sample stage, but it needs to be the position
-                    # of the sample stage + e-beam + scan stage translation (without the drift cor)
                     # TODO: here the code is different compared to _runAcquisitionEbeam
-                    raw_pos = self._acq_data[0][-1].metadata[MD_POS]
-                    strans = spos[0] - orig_spos["x"], spos[1] - orig_spos["y"]
-                    # check if the scan stage is the SEM stage itself
-                    # if it is an 'independent' stage MD_POS (raw_pos) is added to the translation
-                    cor_pos = raw_pos[0] + strans[0], raw_pos[1] + strans[1]
-                    logging.debug("Updating pixel pos from %s to %s", raw_pos, cor_pos)
+                    if scan_stage_is_stage:
+                        # Use the theoretical position of the stage. We could use the stage position as reported by the
+                        # hardware, which could be more accurately representing the current position, but that would
+                        # cause each position to be slightly differently misaligned with the grid, potentially causing
+                        # issues during the display.
+                        cor_pos = spos
+                    else:
+                        # MD_POS default to the center of the sample stage, but it needs to be the position
+                        # of the sample stage + e-beam + scan stage translation (without the drift cor)
+                        raw_pos = self._acq_data[0][-1].metadata[MD_POS]
+                        strans = spos[0] - orig_spos["x"], spos[1] - orig_spos["y"]
+                        # if it is an 'independent' stage MD_POS (raw_pos) is added to the translation
+                        cor_pos = raw_pos[0] + strans[0], raw_pos[1] + strans[1]
+                        logging.debug("Updating pixel pos from %s to %s", raw_pos, cor_pos)
+
                     # In practice, for the e-beam data, it's only useful to
                     # update the metadata for the first pixel.
                     for adas in self._acq_data:
--- a/src/odemis/acq/stream/_sync.py
+++ b/src/odemis/acq/stream/_sync.py
@@ -31,11 +31,13 @@ from functools import partial
 import logging
 import math
 import numpy
+
 from odemis import model, util
 from odemis.acq import drift
 from odemis.acq import leech
 from odemis.acq.leech import AnchorDriftCorrector
 from odemis.acq.stream._live import LiveStream
+from odemis.util.driver import guessActuatorMoveDuration
 from odemis.model import MD_POS, MD_DESCRIPTION, MD_PIXEL_SIZE, MD_ACQ_DATE, MD_AD_LIST, \
     MD_DWELL_TIME, MD_EXP_TIME, MD_DIMS, MD_THETA_LIST, MD_WL_LIST, MD_ROTATION, \
     MD_ROTATION_COR
@@ -275,11 +277,9 @@ class MultipleDetectorStream(Stream, metaclass=ABCMeta):
 
         if hasattr(self, "useScanStage") and self.useScanStage.value:
             if self._sstage:
-                # It's pretty hard to estimate the move time, as the speed is
-                # only the maximum speed and what takes actually most of the time
-                # is to stop to the next point.
-                # TODO: just get the output of _getScanStagePositions() and add
-                # up distances?
+                # It's pretty hard to estimate the move time, as the speed is only the maximum speed
+                # and what takes actually most of the time is to stop to the next point.
+                # TODO: just get the output of _getScanStagePositions() and add up distances?
                 repetition = tuple(self.repetition.value)
                 roi = self.roi.value
                 width = (roi[2] - roi[0], roi[3] - roi[1])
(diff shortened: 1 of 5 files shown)


sneepstefan:
Ok something went wrong with picking the commits, please don't reply on this or review

FCanberk (approved):

src/odemis/acq/stream/_sync.py
```diff
@@ -296,13 +296,20 @@ def estimateAcquisitionTime(self):
 
                 # Count 2x as we need to go back and forth
                 tot_dist = (phy_width[0] * repetition[1] + phy_width[1]) * 2
```
pieleric, line 298: I don't think it makes sense to compute the total distance in X+Y to estimate the duration along X, and also along Y. Also, a long move is much faster than many small moves. So it should be more precise to compute the duration of every X moves and multiply by the number of moves, and same for Y.
```
pxs = self._getPixelSize()
move_time_x =  guessActuatorMoveDuration(self._sstage, "x", pxs[0]) * repetition[0]  * repetition[1]  + guessActuatorMoveDuration(self._sstage, "x", pxs[0] * repetition[0])
move_time_y =  guessActuatorMoveDuration(self._sstage, "y", pxs[1]) * repetition[1]  + guessActuatorMoveDuration(self._sstage, "y", pxs[1] * repetition[1]) 
```

src/odemis/acq/stream/_sync.py
```diff
@@ -1735,13 +1742,17 @@ def _runAcquisitionScanStage(self, future):
...
-                    # TODO: here different from _runAcquisitionEbeam
-                    # sample stage + e-beam + scan stage translation (without the drift cor)
+                    # MD_POS default to the center of the sample stage, but it needs to be the position
+                    # of the sample stage + e-beam + scan stage translation (without the drift cor)
+                    # TODO: here the code is different compared to _runAcquisitionEbeam
                     raw_pos = self._acq_data[0][-1].metadata[MD_POS]
                     strans = spos[0] - orig_spos["x"], spos[1] - orig_spos["y"]
-                    cor_pos = raw_pos[0] + strans[0], raw_pos[1] + strans[1]
+                    # check if a scan-stage wrapper (has stage dependency) is used
+                    # if so only use the translated position and don't add to raw
```
pieleric, line 1751: Check whether the scan stage is just the standard stage (ie, the scan-stage component "affects" the stage),
or it's a completely independent stage (in which case the "raw" MD_POS must be added to the translation.

src/odemis/acq/stream/_sync.py
```diff
@@ -1591,17 +1597,18 @@ def _runAcquisitionScanStage(self, future):
                                           "acquisitions.")
 
         sstage = self._sstage
+        if not sstage:
+            raise ValueError("Cannot acquire with scan stage, as no stage was provided")
+        orig_spos = sstage.position.value  # TODO: need to protect from the stage being outside of the axes range?
+        keep_stage_pos = model.getComponent(role="stage").name in sstage.affects.value
```
pieleric, line 1603: maybe rename to `scan_stage_is_stage` (or something like that). This way, you can use this condition also line 1752.

src/odemis/gui/cont/tabs/sparc_acquisition_tab.py
```diff
@@ -290,11 +290,13 @@ def __init__(self, name, button, panel, main_frame, main_data):
         # add "Use scan stage" check box if scan_stage is present
         sstage = main_data.scan_stage
         if sstage:
-            # Move the scan stage to the center (so that scan has maximum range)
             ssaxes = sstage.axes
             posc = {"x": sum(ssaxes["x"].range) / 2,
                     "y": sum(ssaxes["y"].range) / 2}
-            sstage.moveAbs(posc)
+            # In case of a 'real' scan stage, move the scan stage
```
pieleric, line 296: "real" -> independent.

src/odemis/gui/cont/tabs/sparc_acquisition_tab.py
```diff
@@ -290,11 +290,13 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
+            # to the center (so that scan has maximum range)
+            if main_data.stage.name not in sstage.affects.value:
+                sstage.moveAbs(posc)
```
pieleric, line 299: It seems posc is not used any where else, so you can move it inside the `if` block.

src/odemis/gui/cont/tabs/sparc_acquisition_tab.py
```diff
@@ -303,8 +305,9 @@ def __init__(self, name, button, panel, main_frame, main_data):
                 get_stream_settings_config()[acqstream.SEMStream]["useScanStage"]
             )
 
-            # Draw the limits on the SEM view
             tab_data.useScanStage.subscribe(self._on_use_scan_stage, init=True)
+
+            # In case of a 'real' scan stage, draw the limits on the SEM view
```
pieleric, line 310: You do this always, right? So no need to say "in case of real scan stage".

pieleric (changes requested):

src/odemis/acq/stream/_sync.py
sneepstefan: Why did you add repetition[1] in the move time computation of x in your suggestion?

src/odemis/gui/cont/tabs/sparc_acquisition_tab.py
sneepstefan: Well it is actually used in line 311-314 as well or do I not interpret you comment correctly?

src/odemis/acq/stream/_sync.py
pieleric: Because there will be X*Y spots. There is (roughly) as many moves in X than number of spots (actually, I think it's more like `(X-1) * Y`, so we could slightly improve that formula.

The second part of each formula corresponds to the return move at the end of the acquisition. Actually I don't think it's necessarily going back to the first spot. Maybe we should just drop that part, as it shouldn't be a large amount, and would keep it simpler.

src/odemis/gui/cont/tabs/sparc_acquisition_tab.py
pieleric: Sorry, missed that.

src/odemis/acq/stream/_sync.py
sneepstefan: I improved the formula as you suggested and also renamed repetition -> rep for better readability.

pieleric (changes requested):

pieleric (approved):
Looks good to me. @sneepstefan , should we merge it now, or do you have already more fixes to come to this PR?

sneepstefan:
Yes indeed @pieleric I have just one additional fix to be included soon. Will keep you updated.

src/odemis/acq/stream/_sync.py
```diff
@@ -1737,15 +1737,19 @@ def _runAcquisitionScanStage(self, future):
...
-                    # of the sample stage + e-beam + scan stage translation (without the drift cor)
                     # TODO: here the code is different compared to _runAcquisitionEbeam
-                    raw_pos = self._acq_data[0][-1].metadata[MD_POS]
-                    strans = spos[0] - orig_spos["x"], spos[1] - orig_spos["y"]
-                    # check if the scan stage is the SEM stage itself
-                    # if it is an 'independent' stage MD_POS (raw_pos) is added to the translation
-                    cor_pos = raw_pos[0] + strans[0], raw_pos[1] + strans[1]
-                    logging.debug("Updating pixel pos from %s to %s", raw_pos, cor_pos)
+                    if scan_stage_is_stage:
+                        # just make the MD_POS of cor_pos the same as spos neglecting the enabling of drift correction
```
pieleric, line 1742: That comment is really not clear. 
I'd suggest: "Use the theoretical position of the stage. We could use the stage position as reported by the hardware, which could be more accurately representing the current position, but that would cause each position to be slightly differently misaligned with the grid, potentially causing issues during the display."

pieleric (changes requested):

PR #2666, opened 2024-02-08
[ci] add minimal linting, will only fail for errors on changed files

tepals:


pieleric:
Are you sure the test will fail if you change a file with whitespace? Can you add a whitespace to a python file in this PR to check it fails?

tepals:
> Are you sure the test will fail if you change a file with whitespace? Can you add a whitespace to a python file in this PR to check it fails?

yes, see this run: 
https://github.com/tepals/odemis/actions/runs/7832176957/job/21370243934

.github/workflows/linting.yml
```diff
@@ -0,0 +1,44 @@
...
+        # W391 Blank line at end of file
+        # E999 SyntaxError
+        run: flake8 . --select W291,W292,W293,W391,E999
+
+      - name: Run flake8 on changed files
+        run: |
+          for changed_file in ${{ steps.changed-files.outputs.all_changed_files }}; do
+            if [[ $changed_file == *.py ]]
+            then
+              flake8 $changed_file --select W291,W292,W293,W391,E999
```
pieleric, line 42: This is a bash script. I think that inside `[[ ]]` you don't need `"` to protect the variable expansions. However, the argument to `flake8` needs to be written as `"$changed_file"`  (in case the file name contains odd characters like space).

pieleric (changes requested):

pieleric (approved):

FCanberk (approved):

PR #2667, opened 2024-02-09
[feature] update release-odemis to push major/minor releases to beta channel

pieleric:
First version of a branch typically goes to the beta channel. In theory,
the first N versions might go to the beta channel, until we are
confident it's stable enough. In practice, for now, we've always managed
to be confident enough from .1 onwards, so let's keep the script simple.

### DIFF
--- a/util/release-odemis
+++ b/util/release-odemis
@@ -34,6 +34,11 @@ fi
 
 cd ~/development/odemis
 
+if ! git remote get-url upstream > /dev/null ; then
+    echo "Need to have a upstream remote (typically pointing to git@github.com:delmic/odemis.git)"
+    exit 1
+fi
+
 # Uncomment this line to run line-per-line
 # trap 'read -p "run: $BASH_COMMAND"' DEBUG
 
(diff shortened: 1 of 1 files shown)


util/release-odemis
```diff
@@ -97,15 +110,15 @@ if [[ "$BPVERSION" != "$PVERSION" ]]; then
...
 
-read -p "Build odemis $RELVER and upload it? [Yn] " answer
+read -p "Build odemis $RELVER and upload it to $PPA_NAME? [Yn] " answer
 if [[ $answer =~ ^[nN].* ]]; then
     exit 2
 fi
 git archive --prefix=odemis/ -o ../odemis_${RELVER}.orig.tar.gz HEAD || exit 1
 debuild -S || exit 1
 
 # TODO: if the version ends by ".0", push to proposed
```
tepals, line 120: you can now remove this TODO

util/release-odemis
```diff
@@ -97,15 +110,15 @@ if [[ "$BPVERSION" != "$PVERSION" ]]; then
     exit 1
```
tepals, line 110: There is a TODO on line 96 (github just won't let me comment there): `# TODO: for a minor release, this is sufficient:`
That should now be easy to implement, although I'm not sure if it actually matters. Could you either implement the comment, or remove the TODO?

tepals (changes requested):

util/release-odemis
pieleric: Actually, I think this is more like "the first time you do a release of this branch on this computer". As it's not hurting to always explicitly do "--set-upstream upstream" everytime, and it might not always be the same person/computer that does the release, I've dropped the TODO, and adjusted the comment.

tepals (approved):

K4rishma (approved):

PR #2668, opened 2024-02-09
[fix] xt_client external VA failed when no detector is used

pieleric:
Changing the external VA if there is no detector connected was broken by
commit 47b9b1d4b (dwell time slider does not work for overview imaging).

This commit fixed reading the resolution VA when external is True.
However, if there are no detector, there is no resolution VA.
=> need to handle such case.


tepals (approved):

tepals:
Does it make sense to add a test case for the situation were you have scanner without a detector?

nandishjpatel (approved):

PR #2659, opened 2024-02-07
[SDEV-1812] add ruler to fastem toolbar

nandishjpatel:

PR #2669, opened 2024-02-12
[SDEV-1813] toolbox tool to draw rectangles

nandishjpatel:
NOTE: https://github.com/delmic/odemis/pull/2659 needs to be merged first before reviewing this PR

### DIFF
--- a/src/odemis/gui/comp/overlay/rectangle.py
+++ b/src/odemis/gui/comp/overlay/rectangle.py
@@ -73,21 +73,6 @@ class Rectangle(object):
             cls=cls,
             unit=unit,
         )
-        self.coordinates.subscribe(self._on_coordinates, init=True)
-
-    def _on_coordinates(self, coordinates):
-        """
-        Update the overlay with the new data of the .coordinates VA.
-        coordinates (tuple of 4 floats): left, top, right, bottom position in m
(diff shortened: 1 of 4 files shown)


src/odemis/gui/comp/overlay/rectangle.py
```diff
@@ -0,0 +1,356 @@
...
+        return None
+
+    def on_left_down(self, evt):
+        """Start drawing a rectangle if the overlay is active and there is no selected rectangle."""
+        if not self.active.value:
+            return super().on_left_down(evt)
+
+        self._selected_rectangle = self._get_rectangle(evt)
+        if self._selected_rectangle is None:
+            rectangle = RectangleOverlay(self.cnvs)
```
pieleric, line 316: As discussed, I think that you can mostly get it working by using Overlays. However it can be tricky to properly handle hover (with a correct icon). It might also make things quite complicated if you want to introduce groups.
So I'd recommend having a separate class, not directly an Overlay, and having this RectangesOverlay class call the draw of each instance of Rectangle/Shape.

pieleric (changes requested):

pieleric (changes requested):

PR #2671, opened 2024-02-15
[NSRPARMA-36] New driver for supporting the Diligent MCC DAQ 1208LS

sneepstefan:
Driver support for the Diligent MCC DAQ device, type 1208LS based on the Linux driver written by Warren Jasper.
https://github.com/wjasper/Linux_Drivers


sneepstefan:
Just to be clear this is not my own code, its written by Warren. Just implemented it for use with Odemis. Did a few small adjustments only.

pieleric:
Please also include in this PR the actual odemis driver that uses this file.

sneepstefan:
I just thought of splitting those due to the amount line of code added to make PR's like this get merged faster. In principle this is stand-alone driver code right?

pieleric:
There essentially nothing to review here, because it *should* be just a copy-paste from the original code.
However it's not a complete picture of why you need this change. It's not used any where in Odemis, and cannot be used as-is by any user. That's why to me I think it only makes sense to review as a part of the "pwrmccdaq" driver. I would rather review all at once. (The GUI code can be separate).

patrickcleeve2 (approved):

nandishjpatel (approved):

src/odemis/driver/_mccdaq.py
```diff
@@ -0,0 +1,700 @@
...
+
+
+class usb_1208LS(usb_1208):
+    def __init__(self, serial=None):
+        self.productID = 0x0007a  # USB-1208LS
+        usb_1208.__init__(self)
+        try:
+            self.h.open(0x09db, self.productID, serial)
+        except Exception:
+            raise HwError
```
pieleric, line 688: You should raise an *instance* of an exception. Also, it might be handy to know what caused the exception.
```
        except Exception as ex:
            raise HwError("Failed to find USB device {self.productID}, serial {serial}: {ex}")
```

src/odemis/driver/_mccdaq.py
```diff
@@ -0,0 +1,700 @@
...
+                         Port A     =   0x01
+                         Port B     =   0x04
+        Bug: need leading 0 in command string.
+        '''
+
+        self.h.write([0x0, self.DIN, port_number, 0, 0, 0, 0, 0, 0])
+        try:
+            value = self.h.read(8, 500)
+        except:
+            print('DIn: error in reading.')
```
pieleric, line 255: This type of except + print is really not helpful. The print will be lost in the darkness of backend stdout, and immediately after `value[0]` will fail as `value` is not defined, and hide the actual error. I'd rather just drop the whole try/except (for all the commands).

src/odemis/driver/_mccdaq.py
```diff
@@ -0,0 +1,700 @@
...
+           range:   the gain and input mode setting
+        '''
+
+        if (gain == self.SE_10_00V):
+            mode = 0  # single ended
+        else:
+            mode = 1  # differential
+
+        if (channel > 3 and mode == 1):
+            print('AIn: channel out of range for differential mode.')
```
pieleric, line 322: All `print()` should be changed to logging.error(), otherwise we'll never see them.

src/odemis/driver/_mccdaq.py
```diff
@@ -0,0 +1,700 @@
...
+    def SetID(self, id):
+        '''
+        This command stores an identifier on the unit.  Values of 0-255 are valid.  Note that 0
+        is used to flag uninitialized units.
+
+        id: user id number (0-255)
+        '''
+
+        if id < 0 or id > 255:
+            print('SetID: id out of range')
```
pieleric, line 630: `raise ValueError(...)`

src/odemis/driver/_mccdaq.py
```diff
@@ -0,0 +1,700 @@
...
+            return
+        try:
+            self.h.write([self.SET_ID, id, 0, 0, 0, 0, 0, 0])
+        except:
+            print('Error in writing id')
```
pieleric, line 635: `raise IOError(f'Error in writing id: {ex}')`

src/odemis/driver/_mccdaq.py
```diff
@@ -0,0 +1,700 @@
+#! /usr/bin/python3
```
pieleric, line 1: No need for shebang, as it's not an executable. => drop  this line.

src/odemis/driver/_mccdaq.py
```diff
@@ -0,0 +1,700 @@
...
+
+
+class UnderrunError(Error):
+  ''' Raised when underrun on AOutScan'''
+  pass
+
+
+class SaturationError(Error):
+  ''' Raised when DAQ is saturated '''
+  pass
```
pieleric, line 45: Unused? => drop.

pieleric (changes requested):

src/odemis/driver/_mccdaq.py
sneepstefan: I'd rather not tinker too much on GH code, but I attempted to remove a few.

sneepstefan:
Hey guys I have updated the driver file _mccdaq.py (it looks new due to complete copy/paste of different file) and added pwrmccdaq.py for the implementation and simulation part of the driver as well.
It is getting really big, I hope you don't mind reviewing it again.
Test cases will follow.

src/odemis/driver/pwrmccdaq.py
```diff
@@ -0,0 +1,420 @@
+import logging
```
nandishjpatel, line 1: author and license is missing

src/odemis/driver/pwrmccdaq.py
```diff
@@ -0,0 +1,420 @@
...
+
+
+class MCCDevice(HwComponent):
+    """
+    Basic version of the Class for MCC DAQ device functionality with support for digital input status.
+    This class can be inherited for more functionality to support more complex components.
+    """
+
+    def __init__(self, name: str, role: str, mcc_device: str,
+                 di_channels: Dict[int, Tuple[str, bool]] = None, **kwargs):
```
nandishjpatel, line 23:
```suggestion
                 di_channels: Dict[int, Tuple[str, bool]] = {}, **kwargs):
```

src/odemis/driver/pwrmccdaq.py
```diff
@@ -0,0 +1,420 @@
...
+            key is the channel number(int), value is a list with VA name(str) and tll high is true(bool),
+                this means that the VA will be considered True or False when the TTL signal is high.
+            with this approach it is actually possible to keep track of customized status changes.
+        """
+        super().__init__(name, role, **kwargs)
+        # force add a trailing 0 or just add a comment to config file?
+        # if device and len(device) < 8:
+        #     device = "0" + device
+        self._name = name
+        self._di_channels = di_channels if di_channels else {}
```
nandishjpatel, line 39:
```suggestion
        self._di_channels = di_channels
```

src/odemis/driver/pwrmccdaq.py
```diff
@@ -0,0 +1,420 @@
...
+                raise ValueError(f"Invalid channel value for status {self._di_channels[channel]}")
+
+            # create a VA with False as default
+            va = model.BooleanVA(False, readonly=True)
+            # append a new dict with the added VA
+            self._channel_vas[channel] = va_name, ttl_high, va
+            setattr(self, va_name, va)  # set the class VA variable name
+            logging.info(f"{va_name} status activated for component {self._name} on channel {channel}")
+
+        if self._di_channels:
```
nandishjpatel, line 73: Should log a warning or HwError if no `_di_channels`? not sure what's the right approach

src/odemis/driver/pwrmccdaq.py
```diff
@@ -0,0 +1,420 @@
...
+
+
+class MCCDeviceLight(Emitter, MCCDevice):
+    """
+    Class to support laser emitter control of one or more lasers through the use of a MCCDevice.
+    Inherits from Emitter for ComponentProxy support and from MCCDevice for power and interlock control.
+    """
+
+    def __init__(self, name: str, role: str, mcc_device: str, ao_channels: List[int], do_channels: List[int],
+                 spectra, pwr_curve, di_channels: Dict[int, Tuple[str, bool]] = None, **kwargs):
```
nandishjpatel, line 159:
```suggestion
                 spectra, pwr_curve, di_channels: Dict[int, Tuple[str, bool]] = {}, **kwargs):
```

src/odemis/driver/pwrmccdaq.py
```diff
@@ -0,0 +1,420 @@
...
+                                 "[int, [str, [boolean]]], but got %s" % (va_properties,))
+
+            # check the di_channel values
+            try:
+                self.channel_to_port(channel)
+            except ValueError:
+                raise ValueError(f"Invalid channel value for status {self._di_channels[channel]}")
+
+            # create a VA with False as default
+            va = model.BooleanVA(False, readonly=True)
```
nandishjpatel, line 67:
```suggestion
            va = model.BooleanVA(ttl_high, readonly=True)
```
Maybe I am wrong

src/odemis/driver/pwrmccdaq.py
```diff
@@ -0,0 +1,420 @@
...
+
+    def __init__(self, name: str, role: str, mcc_device: str,
+                 di_channels: Dict[int, Tuple[str, bool]] = None, **kwargs):
+        """
+        :param mcc_device (str or None): serial number or None (null in yaml file) for auto-detect.
+            the device is considered a USB-powered MCC DAQ 1208LS device.
+        :param di_channels (dict -> int, list(str, bool)):
+            the DIO channel used for TTL signal status change through the MCC device.
+            for example di_channels: {2: ["interlockTriggered", False], 7: ["leftSwitch", True]}.
+            key is the channel number(int), value is a list with VA name(str) and tll high is true(bool),
```
nandishjpatel, line 30:
```suggestion
            key is the channel number(int), value is a list with VA name(str) and flag stating if ttl high is True(bool),
```

src/odemis/driver/pwrmccdaq.py
```diff
@@ -0,0 +1,420 @@
...
+            # check the di_channel values
+            try:
+                self.channel_to_port(channel)
+            except ValueError:
+                raise ValueError(f"Invalid channel value for status {self._di_channels[channel]}")
+
+            # create a VA with False as default
+            va = model.BooleanVA(False, readonly=True)
+            # append a new dict with the added VA
+            self._channel_vas[channel] = va_name, ttl_high, va
```
nandishjpatel, line 69: This can work, but I think its better to create a channel class something like a `struct`

src/odemis/driver/pwrmccdaq.py
```diff
@@ -0,0 +1,420 @@
...
+    A really basic and simple interface to simulate a USB-1208LS device with
+    support for DIO pin read/write commands and a single AO write command.
+    """
+
+    def __init__(self):
+        # initialize values
+        self.productID = 0x0007a  # USB-1208LS
+        # to keep track of the individual bits and values
+        self.port_a_bit_config = {"0": 0, "1": 0, "2": 0, "3": 0, "4": 0, "5": 0, "6": 0, "7": 0}
+        self.port_b_bit_config = {"0": 1, "1": 1, "2": 1, "3": 1, "4": 1, "5": 1, "6": 1, "7": 1}
```
pieleric, line 310: Instead of a dict, just use a list of 8 integers (or bool). They are already accessible as config[0] to config[7].

Alternatively, you could just store a single int, and do the bit arithmetic in the function. ;-)

src/odemis/driver/pwrmccdaq.py
```diff
@@ -0,0 +1,420 @@
...
+        """
+        :param port_number: AUXPORT = 0x10 | Port A = 0x01 | Port B = 0x04
+        :return: the value seen at the port pins
+        """
+        ret_val = "0b"
+        DIO_port = self._return_port_config(port_number)
+
+        # use a sorted dict to be sure of the key order
+        for bit in list(sorted(DIO_port.keys()))[::-1]:
+            ret_val += str(DIO_port[bit])
```
pieleric, line 351: Assuming it's a list, you can just write:
`return sum(v << i for i, v in enumerate(DIO_port))`

src/odemis/driver/pwrmccdaq.py
```diff
@@ -0,0 +1,420 @@
...
+        :param channel: selects output channel (0 or 1)
+        :param value: value (uint16) in counts to output [10-bits 0-5V]
+        """
+        channel = 0 if channel > 1 or channel < 0 else None
+        # force automatic clipping
+        if (value > 0x3ff):
+            value = 0x3ff
+        if (value < 0):
+            value = 0
+        self.AO_channels[str(channel)] = value
```
pieleric, line 413: Why converting the channel to a string? You can use directly the channel number (int) as key of the dict.

src/odemis/driver/test/pwrmccdaq_test.py
```diff
@@ -0,0 +1,175 @@
...
+
+from odemis import model
+from odemis.driver import pwrmccdaq
+
+logging.getLogger().setLevel(logging.DEBUG)
+
+CONFIG_MCC = {
+                "name": "Interlock test", "role": "bsd-interlock",
+                "mcc_device": "fake",
+                "di_channels": {2: ["interlockTriggered", False]},
```
pieleric, line 36: Please make it extra complicated by having a second channel on port B. (eg, `9: ["rightFlag", True]`)

src/odemis/driver/test/pwrmccdaq_test.py
```diff
@@ -0,0 +1,175 @@
...
+
+        self.mcc_device.terminate()
+        time.sleep(0.5)  # give the thread some time to suspend
+
+        # check if polling of the DI port bits has suspended
+        self.assertTrue(self.mcc_device._status_thread.terminated)
+        self.assertFalse(self.mcc_device._status_thread.is_alive())
+
+    def tearDown(self):
+        if self.mcc_device._status_thread:
```
pieleric, line 147: Drop that `if`. It should always be safe to call `.terminate()`. If it's not, the MCCDevice component should be fixed.

src/odemis/driver/test/pwrmccdaq_test.py
```diff
@@ -0,0 +1,175 @@
...
+    # check subscribing and unsubscribing from power VA
+    # check right analog output voltage and if DIO is set accordingly
+
+    def setUp(self):
+        pass
+
+    def tearDown(self):
+        pass
+
+    def test_power(self):
```
pieleric, line 164: I think you can mostly copy paste the tests from other drivers, like the pwrcomedi_test.py ones.

pieleric (changes requested):

src/odemis/driver/pwrmccdaq.py
sneepstefan: I changed the position of the check for _di_channels being not empty so it will cover _channel_vas as well

src/odemis/driver/pwrmccdaq.py
sneepstefan: I guess its quite confusing but we like the VA default value to start with False. The ttl_high boolean represents the interpretation of the BooleanVA as in which case to set it to true. When ttl_high is true the BooleanVA should still be false when the signal bit is 0. When the bit changes to 1, the ttl signal is high and the BooleanVA set to true. This way the bit can also be 1 and the BooleanVA be false but then ttl_high should be false of course.

src/odemis/driver/pwrmccdaq.py
```diff
@@ -0,0 +1,420 @@
...
+            the DIO channel used for TTL signal status change through the MCC device.
+            for example di_channels: {2: ["interlockTriggered", False], 7: ["leftSwitch", True]}.
+            key is the channel number(int), value is a list with VA name(str) and tll high is true(bool),
+                this means that the VA will be considered True or False when the TTL signal is high.
+            with this approach it is actually possible to keep track of customized status changes.
+        """
+        super().__init__(name, role, **kwargs)
+        # force add a trailing 0 or just add a comment to config file?
+        # if device and len(device) < 8:
+        #     device = "0" + device
```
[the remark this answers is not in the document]
sneepstefan: It should have been preceding instead of trailing, changed it in the docstring

src/odemis/driver/test/pwrmccdaq_test.py
sneepstefan: Did that in test_multiple_di_channels()

patrickcleeve2 (approved):

src/odemis/driver/pwrmccdaq.py
```diff
@@ -0,0 +1,453 @@
+# -*- coding: utf-8 -*-
+'''
+Created on 23 Feb 2024
+
+@author: Stefan Sneep
+
+Copyright © 2016 Éric Piel, Delmic
```
nandishjpatel, line 7: The copyright name can be yours @sneepstefan ;)

nandishjpatel (approved):

src/odemis/driver/pwrmccdaq.py
```diff
@@ -0,0 +1,453 @@
...
+        """
+        if value < 0 or value > 255:
+            raise ValueError("Value to set is not between 0 and 255")
+
+        DIO_port = self._return_port_config(port_number)
+        in_val = format(value, '#010b')
+        in_val = in_val.replace("0b", "")
+
+        for num, c in enumerate(in_val[::-1]):
+            DIO_port[num] = int(c)
```
pieleric, line 401: Please use bit manipulation instead of strings. Also it should only modify the bits that are not masked in Dconfig.

I think in bitwise operation, you would write `dio_port ^=  ((dio_port ^ value) & dio_config)`. But that would require they are all ints.
Here, you can convert `value` to a list of booleans `list_val`, and then do
```
for i in range(8):
    if dio_config[i]
        dio_port[i] = list_val[i]
```

src/odemis/driver/pwrmccdaq.py
```diff
@@ -0,0 +1,453 @@
...
+        :param channel: selects output channel (0 or 1)
+        :param value: value (uint16) in counts to output [10-bits 0-5V]
+        """
+        if channel > 1 or channel < 0:
+            channel = 0
+        # force automatic clipping
+        if (value > 0x3ff):
+            value = 0x3ff
+        if (value < 0):
+            value = 0
```
pieleric, line 445: Clipping can be done in 1 line:
```suggestion
        value  = min(max(0, value), 0x3ff)
```

src/odemis/driver/test/pwrmccdaq_test.py
```diff
@@ -0,0 +1,322 @@
...
+                    },
+                    {
+                        0: 0,
+                        5: 0.06,  # 60mW light emitter
+                    }],
+                "di_channels": {10: ["interlockTriggered", False]},
+}
+
+
+# django implementation of tag for use as a decorator in unittest (from django.test.utils)
```
pieleric, line 60: "django" ?! It's good that you indicate where this code comes from, but please first explain what this code does.

src/odemis/driver/test/pwrmccdaq_test.py
```diff
@@ -0,0 +1,322 @@
...
+class TestMCCDevice(unittest.TestCase):
+    def setUp(self):
+        method = getattr(self, self._testMethodName)
+        tags = getattr(method, 'tags', {})
+        if 'skip_setup' in tags:
+            self.mcc_device = None
+            return  # setUp skipped
+
+        self.mcc_device = pwrmccdaq.MCCDevice(**CONFIG_MCC)
+        time.sleep(1)
```
pieleric, line 80: I think this whole thing could be simplified by having just a function called like `_create_device()` which would be called from each function which needs the setup.

src/odemis/driver/test/pwrmccdaq_test.py
```diff
@@ -0,0 +1,322 @@
...
+        mcc_device.device.DConfig(0x01, 0xff)
+        mcc_device.device.DConfig(0x04, 0xff)
+        time.sleep(0.15)  # wait a bit to update the port bit status
+
+        # trigger bits 2 and 3 for port A -> binary 0b11110011 -> hex 0x4d
+        mcc_device.device.DOut(0x01, 0xf3)
+
+        # trigger bits 1, 4, 5 and 7 for port B -> binary 0b01001101 -> hex 0x4d
+        mcc_device.device.DOut(0x04, 0x4d)
+        time.sleep(0.15)  # wait a bit to update the port bit status
```
pieleric, line 180: Does this code work on a real device or just the simulator? Is just the simulator, indicate so. 

I am a little bit confused by the usage of DConfig. I thought it's to indicate which port is used for DO or DI. It should be done by the driver, right?

src/odemis/driver/test/pwrmccdaq_test.py
```diff
@@ -0,0 +1,322 @@
...
+class TestMCCDeviceLight(TestMCCDevice):
+    def setUp(self):
+        method = getattr(self, self._testMethodName)
+        tags = getattr(method, 'tags', {})
+        if 'skip_setup' in tags:
+            self.mcc_device = None
+            return  # setUp skipped
+
+        self.mcc_device = pwrmccdaq.MCCDeviceLight(**CONFIG_LIGHT_DEVICE)
+        time.sleep(1)
```
pieleric, line 228: Please add a comment why a sleep is needed.

pieleric (changes requested):

src/odemis/driver/pwrmccdaq.py
```diff
@@ -0,0 +1,453 @@
...
+If not, see http://www.gnu.org/licenses/.
+'''
+# Uses a Diligent MCC DAQ board with analog output which can be used to control
+# the power of an emitter.
+# This is designed for the USB-1208LS board, but it can be modified to support
+# other board types of this brand. It is mainly used in single-ended mode pinout.
+# Note, with the USB-1208LS board, only 2 analogue channels can be used as output
+# you should use pins 13 and 14 for channels 0->1.
+# For using the DIO ports set the port bit to 1 for digital output and to 0 for
+# digital input. Use pins 21->28 for channels 0->7 and pins 32->39 for channels 8->15.
```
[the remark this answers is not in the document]
sneepstefan: That is true indeed, and if it needs changes it is possible to set it differently in the code. For now we will keep all the bits on 0 when the driver is initiated.

src/odemis/driver/_mccdaq.py
```diff
@@ -0,0 +1,674 @@
...
+
+        self.device_type, self.device_sn = self.h.get_product_string(), self.h.get_serial_number_string()
+        logging.debug(f"Successfully connected with device type {self.device_type} with s/n {self.device_sn}")
+
+        # enable non-blocking mode
+        self.h.set_nonblocking(1)
+
+        # set default configuration
+        self.DConfig(self.DIO_PORTA, 0x00)  # Port A output
+        self.DConfig(self.DIO_PORTB, 0x00)  # Port B input
```
pieleric, line 669: The comments don't seem to match the code. Both ports are set to "input" (0) at init, right?

src/odemis/driver/pwrmccdaq.py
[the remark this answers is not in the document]
pieleric: After reading the rest of the driver, I think I'm getting it. I'd suggest to clarify it by rephrasing like: "For selecting the direction of the DIO ports, the DConfig register needs to be set to 1 for output and 0 for input, for each bit."

Note that you're not actually doing this in the code! The low-level driver indeed set everything to 0. That means all ports are for input. So the `do_channels` are not going to work. Did you check the code with the real device recently? I guess you'll need to call DConfig with the do_channel bit mask.

pieleric (changes requested):

src/odemis/driver/pwrmccdaq.py
```diff
@@ -0,0 +1,453 @@
...
+
+from dataclasses import dataclass
```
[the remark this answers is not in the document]
sneepstefan: Thanks for pointing that out. Might we need that on Ubuntu 18.04 systems we can always bump the Python version to 3.7 right?

src/odemis/driver/pwrmccdaq.py
```diff
@@ -0,0 +1,453 @@
...
+    def DConfig(self, port_number, bit_mask):
+        """
+        This command sets the direction of the digital bits for a port.
+        :param port_number: AUXPORT = 0x10 | Port A = 0x01 | Port B = 0x04
+        :param bit_mask (int:bit value): 0 = output | 1 = input
+        """
+        if self.productID == 0x0075 and port_number == usb_1208LS.DIO_AUXPORT:
+            bit_mask = ((bit_mask ^ 0xff) & 0xff)
+
+        self.DOut(port_number, bit_mask)
```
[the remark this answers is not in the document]
sneepstefan: Does it matter for the simulator what the real code does in this case? In addition I must admit I am unsure what the bit mask does and what it is for. Your explanation is not clear to me as well.

src/odemis/driver/test/pwrmccdaq_test.py
sneepstefan: Yes good point, at first it seemed quite handy to make a distinction between tests that use the setUp() and others that don't, and still keep them in the same class. Your suggestion below to use a separate private function to create the MCCDevice class fits this scenario better and I can leave out the "django" wrapper.

src/odemis/driver/pwrmccdaq.py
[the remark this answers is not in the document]
pieleric: It'd be very really hard. We could always upgrade the computer. That should be much easier!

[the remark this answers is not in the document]
sneepstefan: Ah that is also an option I guess.
I though when reading this: https://linuxize.com/post/how-to-install-python-3-7-on-ubuntu-18-04/
Upgrading from source would be not so hard.

src/odemis/driver/pwrmccdaq.py
```diff
@@ -0,0 +1,464 @@
...
+    """
+    A really basic and simple interface to simulate a USB-1208LS device with
+    support for DIO pin read/write commands and a single AO write command.
+    """
+    def __init__(self):
+        # initialize values
+        self.productID = 0x0007a  # USB-1208LS
+        # to keep track of the individual bits and values
+        # self.port_a_bit_config = [0, 0, 0, 0, 0, 0, 0, 0]
+        # self.port_b_bit_config = [0, 0, 0, 0, 0, 0, 0, 0]
```
pieleric, line 366: Please drop this.

src/odemis/driver/pwrmccdaq.py
```diff
@@ -0,0 +1,464 @@
...
+        :param port_number: AUXPORT = 0x10 | Port A = 0x01 | Port B = 0x04
+        :param value: value to write to the port (0-255)
+        """
+        if value < 0 or value > 255:
+            raise ValueError("Value to set is not between 0 and 255")
+
+        if port_number == usb_1208LS.DIO_PORTA:
+            self.port_a_bit_config = [bool(value & 1 << i) for i in range(8)]
+        else:
+            self.port_b_bit_config = [bool(value & 1 << i) for i in range(8)]
```
pieleric, line 414: DOut and DIn should not write/read on the `config` register. They should affect a different memory. That said, it's a simulator, so in practice, if you write somewhere, no one will ever really know. So you could just do nothing here, and it'd be fine. 
Still, if you want to be able to call DOut and followed by DIn, to simulate a change in the value read by DIn, then this current is fine... but use a separate register from `config`.

src/odemis/driver/pwrmccdaq.py
```diff
@@ -0,0 +1,464 @@
...
+        :param port_number: AUXPORT = 0x10 | Port A = 0x01 | Port B = 0x04
+        :param bit: the bit to read (0-7)
+        :return (int): value 0 or 1
+        """
+        if bit < 0 or bit > 7:
+            raise ValueError("Bit value is not between 0 and 7")
+
+        DIO_port = self._return_port_config(port_number)
+
+        return DIO_port[bit]
```
pieleric, line 429: For DBitIn and DBitOut, as it's a simulator, the values read/written are not really meaningful. You could just always return 0 (or 1, whatever is more convenient for now).
What you could also do on these single bit function is to check the config register and warn if it hasn't been set to the right value beforehand (eg, 1 for BitIn and 0 for BitOut).

src/odemis/driver/pwrmccdaq.py
```diff
@@ -0,0 +1,464 @@
...
+        # force automatic clipping
+        value = min(max(0, value), 0x3ff)
+
+        self.AO_channels[channel] = value
+
+    def _return_port_config(self, port_number):
+        if port_number == usb_1208LS.DIO_PORTB:
+            return self.port_b_bit_config
+        else:
+            # Port A as default
```
pieleric, line 463: Please explicitly handle port A, and raise an exception for anything else. That will help to detect more errors in the caller.

src/odemis/driver/test/pwrmccdaq_test.py
```diff
@@ -0,0 +1,318 @@
...
+        # set all DI ports to HIGH
+        mcc_device.device.DConfig(0x01, 0xff)
+        mcc_device.device.DConfig(0x04, 0xff)
+        time.sleep(0.15)  # wait a bit to update the port bit status
+
+        # trigger bits 2 and 3 for port A -> binary 0b11110011 -> hex 0x4d
+        mcc_device.device.DOut(0x01, 0xf3)
+
+        # trigger bits 1, 4, 5 and 7 for port B -> binary 0b01001101 -> hex 0x4d
+        mcc_device.device.DOut(0x04, 0x4d)
```
pieleric, line 164: Does this code works on the real hardware? I would expect that it doesn't pass, because DOut doesn't even have any effect with Dconfig=0xff. Even if DConfig was temporarily set to 0x00, afterwards, I don't think the value you output on this port is going to be "recorded" when reading again.

You could make this test case skipped if not simulated (something like `if not TEST_NOHW: self.skipTest(...)` ).
Or you could remove completely this part of the test.
=> I'd suggest to

src/odemis/driver/test/pwrmccdaq_test.py
```diff
@@ -0,0 +1,318 @@
...
+import unittest
+
+from odemis import model
+from odemis.driver import pwrmccdaq
+
+logging.getLogger().setLevel(logging.DEBUG)
+
+CONFIG_MCC = {
+                "name": "Interlock test", "role": "bsd-interlock",
+                "mcc_device": "fake",
```
pieleric, line 35: Typically we put by default code that test the real hardware, and set the simulator mode when TEST_NOHW env variable == 1. Could you do the same here?

src/odemis/driver/pwrmccdaq.py
```diff
@@ -0,0 +1,464 @@
...
+    """
+    Basic version of the Class for MCC DAQ device functionality with support for digital input status.
+    This class can be inherited for more functionality to support more complex components.
+    """
+    def __init__(self, name: str, role: str, mcc_device: Optional[str],
+                 di_channels: Dict[int, Tuple[str, bool]] = {}, **kwargs):
+        """
+        :param mcc_device (str or None): serial number or None (null in yaml file) for auto-detect.
+            the device is considered a USB-powered MCC DAQ 1208LS device. the serial number should
+            contain an added 0 to the beginning of the s/n.
```
pieleric, line 70: Please add that "fake" makes use of a simulator.

src/odemis/driver/test/pwrmccdaq_test.py
sneepstefan: I have chosen to merge the test cases of the mcc device with the ones for the light (also using a mcc device). This way the test cases can be ran with real hw or a simulator and also interlocks can be tested this way.

pieleric (approved):

PR #2672, opened 2024-02-15
[ci] driver xt_client: add test cases for the case of no detectors

pieleric:
It's very often used with only the stage + ebeam control, but no
detector, as support to the analog scanning.
So it's good to have a test case that check this type of setup.


src/odemis/driver/test/xt_client_test.py
```diff
@@ -594,6 +608,57 @@ def test_brightness_contrast(self):
...
+            elif child.name == CONFIG_FOCUS["name"]:
+                cls.efocus = child
+            elif child.name == CONFIG_STAGE["name"]:
+                cls.stage = child
+            elif child.name == CONFIG_CHAMBER["name"]:
+                cls.chamber = child
+
+    # Disable tests for anything related to the detector
+    def test_acquire(self):
+        pass
```
tepals, line 636: I would personally prefer to explicitly skip these tests. If you just use `pass` it will show up as a successful test in the logs.

```suggestion
        self.skipTest("skip detector related test")
```

tepals (approved):

nandishjpatel (approved):

PR #2673, opened 2024-02-16
[refactor] drop last usage of Python 2 "future" package

pieleric:
The "import future" are long time gone, but we still have a few "import past" usage. This was to support Python 2.
=> Let's get rid of them.


src/odemis/driver/andorcam3.py
```diff
@@ -354,7 +353,6 @@ def __init__(self, name, role, device=None, bitflow_install_dirs=None,
             tran_rng = ((-uh_shape[0], -uh_shape[1]),
                         (uh_shape[0], uh_shape[1]))
             self.translation = model.ResolutionVA((0, 0), tran_rng, unit="px",
-                                                  cls=(int, long),
```
pieleric, line 357: Note: we can completely drop the "cls" argument, because the default value does the same job (better).

tepals (approved):

nandishjpatel (approved):

PR #2674, opened 2024-02-19
[fix] Increase timeout of xt autostigmation

tepals:
Commit 2fc1f5b84722c45ffb3e0fa9caf97e61d51033b1 added autostigmation, however autostigmation takes longer than 30 seconds, which is the default timeout. This fix increases the timeout when running autostigmation and sets it back to 30 seconds.

### DIFF
--- a/src/odemis/driver/xt_client.py
+++ b/src/odemis/driver/xt_client.py
@@ -1252,7 +1252,7 @@ class SEM(model.HwComponent):
     def start_autostig(self, n_images=17, sweep_range=0.015, dwell_time=3e-6, resolution=(1536, 1024),
                        reduced_area_size=0.5, save_results=False, results_fp=None):
         """
-        Preform the autostig routine for determining the optimal stigmator setting.
+        Preform the autostig routine for determining the optimal stigmator setting. This is a blocking call.
         NOTE: will be available from xtadapter version 1.11.5 and higher.
 
         Parameters
@@ -1275,7 +1275,7 @@ class SEM(model.HwComponent):
 
         Returns
         -------
-        optimal_stigmation
+        optimal_stigmation : float
             Optimal stigmator setting.
         """
         with self._proxy_access:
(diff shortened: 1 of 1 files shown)


src/odemis/driver/xt_client.py
```diff
@@ -1275,13 +1275,24 @@ def start_autostig(self, n_images=17, sweep_range=0.015, dwell_time=3e-6, resolu
...
+        optimal_stigmation
             Optimal stigmator setting.
         """
         with self._proxy_access:
             self.server._pyroClaimOwnership()
-            return self.server.start_autostig(n_images, sweep_range, dwell_time, resolution,
-                                              reduced_area_size, save_results, results_fp)
+            self.server._pyroTimeout = 300  # seconds
+            try:
+                optimal_stigmation = self.server.start_autostig(n_images, sweep_range, dwell_time, resolution,
```
pieleric, line 1285: Could you add in the docstring, that this is blocking (and maybe also imply that every other call is blocked).

src/odemis/driver/xt_client.py
```diff
@@ -1275,13 +1275,24 @@ def start_autostig(self, n_images=17, sweep_range=0.015, dwell_time=3e-6, resolu
...
+                                                                reduced_area_size, save_results, results_fp)
+            except TimeoutError:
+                logging.warning(
+                    "Autostigmation timed out after %s s. Check the xt user interface for the current status.",
+                    self.server._pyroTimeout
+                )
+                optimal_stigmation = None
```
pieleric, line 1292: I'm not convinced. Now the caller has to check whether the return is a float or None. In the case it's None, what should it do? Raise an exception? I think it'd be simpler to always return a float, or raise a TimeoutError in case of timeout. It's worth indicating that the "other side" might still be running the autostigmation as an extra warning.

pieleric (changes requested):

src/odemis/driver/xt_client.py
```diff
@@ -1275,13 +1275,24 @@ def start_autostig(self, n_images=17, sweep_range=0.015, dwell_time=3e-6, resolu
 
         Returns
```
patrickcleeve2, line 1276: The docstrings don't match the current format https://github.com/delmic/odemis/blob/c837e587c6356559fc6f4ace9a6d764cdfd0cd05/doc/develop/coding-guidelines.rst

patrickcleeve2 (approved):

src/odemis/driver/xt_client.py
tepals: This whole file contains a mix of different docstring formats, I'll open a new PR to make sure everything matches the current format.

pieleric (approved):

PR #2675, opened 2024-02-20
[ci] actually validate the Python syntax of all python files

pieleric:


tepals (approved):

pieleric:
Drop this whole idea... it appears that flake8 already checks for the syntax, as E999

PR #2677, opened 2024-02-20
[refactor] Clean up xt_client docstrings to make them match with the …

tepals:
…current guidelines

Additionally make use of typing and remove unused returns.

### DIFF
--- a/src/odemis/driver/xt_client.py
+++ b/src/odemis/driver/xt_client.py
@@ -27,14 +27,15 @@ import re
 import threading
 import time
 import zipfile
+from concurrent import futures
 from concurrent.futures import CancelledError
-from typing import Optional
+from typing import Any, Dict, Optional, Set, Tuple, Union
 
 import msgpack_numpy
 import notify2
 import numpy
-import Pyro5.api
 import pkg_resources
+import Pyro5.api
 from Pyro5.errors import CommunicationError
 
 from odemis import model
@@ -90,6 +91,7 @@ class Package(object):
         version: The version of the xtadapter.
 
     """
+
     def __init__(self):
         self._adapter = None
         self._bitness = None
@@ -139,7 +141,7 @@ class Package(object):
 
 
 def check_latest_package(
-    directory: str, current_version: str, adapter: str, bitness: str, is_zip: bool
+        directory: str, current_version: str, adapter: str, bitness: str, is_zip: bool
 ) -> Optional[Package]:
     """
     Check if a latest xtadapter package is available.
@@ -166,8 +168,9 @@ def check_latest_package(
             result = re.search(regex, entity)
             entity_path = os.path.join(directory, entity)
             if result and (
-                entity.endswith(".zip") if is_zip
-                else os.path.isdir(entity_path) and len([entity for entity in os.listdir(entity_path) if entity.endswith(".exe")]) == 1
+                    entity.endswith(".zip") if is_zip
+                    else os.path.isdir(entity_path) and len(
+                        [entity for entity in os.listdir(entity_path) if entity.endswith(".exe")]) == 1
             ):
                 package = Package()
                 package.adapter = result.group(1)
@@ -213,15 +216,11 @@ class SEM(model.HwComponent):
     be set to "xttoolkit".
     """
 
-    def __init__(self, name, role, children, address, daemon=None,
-                 **kwargs):
+    def __init__(self, name: str, role: str, children: Dict[str, dict], address: str, daemon=None,
+                 **kwargs) -> None:
         """
-        Parameters
-        ----------
-        address: str
+        :param address: (str)
             server address and port of the Microscope server, e.g. "PYRO:Microscope@localhost:4242"
-        timeout: float
-            Time in seconds the client should wait for a response from the server.
         """
 
         model.HwComponent.__init__(self, name, role, daemon=daemon, **kwargs)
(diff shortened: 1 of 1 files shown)


src/odemis/driver/xt_client.py
```diff
@@ -2133,7 +2072,7 @@ def __init__(self, name, role, parent, **kwargs):
 
         self._executor = CancellableThreadPoolExecutor(max_workers=1)
 
-    def stop(self, axes=None):
+    def stop(self, axes: Union[str, None] = None) -> None:
```
pieleric, line 2075: Union[..., None] can be simplified to `Optional[...]`

src/odemis/driver/xt_client.py
```diff
@@ -2291,7 +2230,7 @@ def _getPosition(self):
                 pos[an] = (pos[an] - rng[0]) % (2 * math.pi) + rng[0]
         return pos
 
-    def _moveTo(self, future, pos, rel=False, timeout=60):
+    def _moveTo(self, future: futures.Future, pos: dict, rel: bool = False, timeout: float = 60) -> None:
```
pieleric, line 2233: `Dict[str, float]`

src/odemis/driver/xt_client.py
```diff
@@ -2411,7 +2346,7 @@ def moveAbs(self, pos):
         f = self._executor.submitf(f, self._doMoveAbs, f, pos)
         return f
 
-    def stop(self, axes=None):
+    def stop(self, axes: str = None) -> None:
```
pieleric, line 2349: axes is `Optional[Set[str]]`

src/odemis/driver/xt_client.py
```diff
@@ -2584,9 +2524,9 @@ def _doMoveAbs(self, foc):
             self._updatePosition()
 
     @isasync
-    def moveRel(self, shift):
+    def moveRel(self, shift: dict) -> futures.Future:
```
pieleric, line 2527: `Dict[str, float]`

src/odemis/driver/xt_client.py
```diff
@@ -2597,9 +2537,9 @@ def moveRel(self, shift):
         return f
 
     @isasync
-    def moveAbs(self, pos):
+    def moveAbs(self, pos: dict) -> None:
```
pieleric, line 2540: `Dict[str, float]`

src/odemis/driver/xt_client.py
```diff
@@ -2609,7 +2549,7 @@ def moveAbs(self, pos):
         f = self._executor.submit(self._doMoveAbs, foc)
         return f
 
-    def stop(self, axes=None):
+    def stop(self, axes=None) -> None:
```
pieleric, line 2552: axes type: `Optional[Set[str]]`

src/odemis/driver/xt_client.py
```diff
@@ -1937,13 +1871,14 @@ def stop_acquisition(self):
             # Stopping it twice does a full stop.
             self.parent.set_channel_state(self._scanner.channel, False)
 
-    def _acq_should_stop(self, timeout=None):
+    def _acq_should_stop(self, timeout: Union[float, None] = None) -> bool:
```
pieleric, line 1874: Use `Optional` instead of Union.

src/odemis/driver/xt_client.py
```diff
@@ -1247,36 +1205,34 @@ def start_autostigmating_flash(self):
...
             self.server._pyroClaimOwnership()
-            return self.server.start_autostigmating_flash()
+            self.server.start_autostigmating_flash()
 
-    def start_autostig(self, n_images=17, sweep_range=0.015, dwell_time=3e-6, resolution=(1536, 1024),
-                       reduced_area_size=0.5, save_results=False, results_fp=None):
+    def start_autostig(self, n_images: int = 17, sweep_range: float = 0.015, dwell_time: float = 3e-6,
+                       resolution: Tuple[int, int] = (1536, 1024),
+                       reduced_area_size: float = 0.5, save_results: bool = False,
+                       results_fp: Optional[Union[Path, str]] = None) -> Tuple[float, float]:
```
pieleric, line 1213: Can you really send a `Path` over Pyro5?

pieleric, line 1213: The return seems to be just float.

src/odemis/driver/xt_client.py
```diff
@@ -1099,133 +1065,125 @@ def get_dc_coils(self):
...
 
-        :return (float): Multi probe orientation in degrees
+        :return: Multi probe orientation in degrees
         """
         with self._proxy_access:
             self.server._pyroClaimOwnership()
             return self.server.get_mpp_orientation()
 
-    def mpp_orientation_info(self):
+    def mpp_orientation_info(self) -> Dict[str, list]:
```
pieleric, line 1100: Is that always a `list` for values? The docstring seems to imply it can be `'degrees'`. So maybe just write `Any`?

src/odemis/driver/xt_client.py
```diff
@@ -1029,68 +1005,58 @@ def secondary_stigmator_info(self):
...
+        :param x: x value of stigmator, unitless.
+        :param y: y value of stigmator, unitless.
         """
         with self._proxy_access:
             self.server._pyroClaimOwnership()
-            return self.server.set_pattern_stigmator(x, y)
+            self.server.set_pattern_stigmator(x, y)
 
-    def pattern_stigmator_info(self):
+    def pattern_stigmator_info(self) -> dict:
```
pieleric, line 1041: `Dict[str, Any]`?

src/odemis/driver/xt_client.py
```diff
@@ -651,334 +648,317 @@ def vent(self):
...
+
+        :param free_working_distance: free working distance in meters.
         """
         with self._proxy_access:
             self.server._pyroClaimOwnership()
             self.server.set_free_working_distance(free_working_distance)
 
-    def fwd_info(self):
-        """Returns the unit and range of the free working distance."""
+    def fwd_info(self) -> dict:
```
pieleric, line 730: `Dict[str, Any]`?

pieleric (changes requested):

src/odemis/driver/xt_client.py
```diff
@@ -2793,28 +2735,28 @@ def _updateSettings(self):
         except Exception:
             logging.exception("Unexpected failure when polling XTtoolkit settings")
 
-    def _setDeltaPitch(self, delta_pitch):
+    def _setDeltaPitch(self, delta_pitch: float) -> float:
         self.parent.set_delta_pitch(delta_pitch * 1e6)  # Convert from meters to micrometers.
         return self.parent.get_delta_pitch() * 1e-6
 
-    def _setBeamStigmator(self, beam_stigmator_value):
+    def _setBeamStigmator(self, beam_stigmator_value: Tuple[float, float]) -> float:
```
patrickcleeve2, line 2742:
```suggestion
    def _setBeamStigmator(self, beam_stigmator_value: Tuple[float, float]) -> Tuple[float, float]:
```

src/odemis/driver/xt_client.py
```diff
@@ -2793,28 +2735,28 @@ def _updateSettings(self):
...
         self.parent.set_stigmator(*beam_stigmator_value)
         return self.parent.get_stigmator()
 
     def _setPatternStigmator(self, pattern_stigmator_value):
```
patrickcleeve2, line 2746:
```suggestion
    def _setPatternStigmator(self, pattern_stigmator_value: Tuple[float, float]) -> Tuple[float, float]:
```

patrickcleeve2 (changes requested):

FCanberk (changes requested):

pieleric:
BTW, to test your typing, you can install `beartype` with pip3, and then add at the top of the xt_client.py file:
```
from beartype.claw import beartype_this_package 
beartype_this_package()                              
```
Then, run the test cases.
It will raise an exception if a function is ever called with a value or returns a value which does not fit the type.

src/odemis/driver/xt_client.py
```diff
@@ -350,273 +349,269 @@ def check_and_transfer_latest_package(self) -> None:
...
-                    update.set_timeout(10000)    # 10 seconds
+                    update.set_timeout(10000)  # 10 seconds
                     update.show()
                 else:
                     logging.warning("{} is a bad file in {} not transferring latest package.".format(ret, package.path))
         except Exception as err:
             logging.exception(err)
 
-    def list_available_channels(self):
+    def list_available_channels(self) -> dict:
```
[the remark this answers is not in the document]
tepals: I just noticed that this function is not there anymore in the xtadapter, so I will completely remove it. Should I do that as a separate commit in this PR or in a new one?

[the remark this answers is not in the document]
pieleric: There are actually no user? If so, the removal can be very simple, and just as a separate commit of this PR. Otherwise, just keep it for later, as a separate PR.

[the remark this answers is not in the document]
tepals: There is no actual user, I'll then do it as a separate commit on this PR.

src/odemis/driver/xt_client.py
tepals: It is a tuple of x and y, I've updated the docstring to clarify

patrickcleeve2 (approved):

PR #2678, opened 2024-02-20
[SDEV-1807] configure multibeam

tepals:
Now that we are not running the scan amplitude pre-alignment in calibration 1, but have moved it to monthly maintenance, the calibrated value is not automatically set on the VA. This PR ensures that we are running the acquisition with the calibrated scan offset and amplitude.

@nandishjpatel I'm not sure if this is the best way to fix this, maybe we should rethink how to store calibrated values, now that we want to move some stuff to monthly maintenance.

### DIFF
--- a/src/odemis/acq/fastem_conf.py
+++ b/src/odemis/acq/fastem_conf.py
@@ -178,3 +178,12 @@ def configure_detector(detector, roc2, roc3):
         logging.warning("Region of calibration doesn't have cell translation parameters.")
     else:
         detector.cellTranslation.value = cell_translation
+
+
+def configure_multibeam(multibeam):
+    """
+    :param multibeam: (technolution.EBeamScanner) The multibeam scanner component that needs to be configured.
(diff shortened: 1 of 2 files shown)


src/odemis/acq/fastem_conf.py
```diff
@@ -178,3 +178,12 @@ def configure_detector(detector, roc2, roc3):
         logging.warning("Region of calibration doesn't have cell translation parameters.")
     else:
         detector.cellTranslation.value = cell_translation
+
+
+def configure_multibeam(multibeam):
```
nandishjpatel, line 183: The arguments of the for each of this configure functions is not the same, so it will be difficult. Otherwise a single configuration yaml file can possibly be used.

tepals: I'll just leave it like this then

nandishjpatel (approved):

PR #2679, opened 2024-02-20
[feature] driver semnidaq: support multiple ports for fast TTLs

pieleric:
Until now we only supported 0 or 1 digital out port for each type of fast TTL. It would be possible to connect physically several cables to each port. However, in some cases, it's more convenient to have the hardware directly output the same signal on multiple ports, and connect one cable per port.

This is on top of PR #2655


pieleric:
@tepals @FCanberk , please only look at the *last* commit.

FCanberk (approved):

tepals (approved):

PR #2680, opened 2024-02-21
[fix] autostig period for idx 0 behaved incorrectly

tepals:
For idx 0 even if the autostig period was set to 0 it would still run the extended pre-calibrations. Additionally it should not run for idx 0, only for subsequent indices.

### DIFF
--- a/src/odemis/gui/cont/acquisition/fastem_acq.py
+++ b/src/odemis/gui/cont/acquisition/fastem_acq.py
@@ -32,7 +32,6 @@ import math
 import os
 from builtins import str
 from concurrent.futures._base import CancelledError
-from datetime import datetime
 from functools import partial
 
 import wx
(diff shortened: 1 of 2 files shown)


src/odemis/gui/cont/acquisition/fastem_acq.py
```diff
@@ -523,7 +523,10 @@ def on_acquisition(self, evt):
 
         for p in self._tab_data_model.projects.value:
             for idx, roa in enumerate(p.roas.value):
-                pre_calib = pre_calib_plus if idx % autostig_period == 0 else pre_calibrations
+                if idx == 0 or idx % autostig_period != 0:
+                    pre_calib = pre_calibrations
+                elif idx % autostig_period == 0:
```
pieleric, line 528: Should just be `else`. It should be clear that there are only 2 cases.

pieleric (changes requested):

pieleric (approved):

nandishjpatel (approved):

PR #2685, opened 2024-02-22
[fix] GUI issue when plot data is just 0's

pieleric:
Found some small GUI issues when testing a monochromator simulator which constantly returns 0.

### DIFF
--- a/src/odemis/gui/comp/overlay/marking_line.py
+++ b/src/odemis/gui/comp/overlay/marking_line.py
@@ -150,8 +150,14 @@ class MarkingLineOverlay(base.ViewOverlay, base.DragMixin):
 
             # If it's a big number: use at least all the characters needed to show the number (IOW, round to an int).
             # If it's a small number: show 4 significant digits.
-            sig_x = max(4, math.ceil(math.log10(abs(val[0]))))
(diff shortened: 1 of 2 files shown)


nandishjpatel (approved):

src/odemis/gui/comp/overlay/marking_line.py
```diff
@@ -150,8 +150,14 @@ def draw(self, ctx):
 
             # If it's a big number: use at least all the characters needed to show the number (IOW, round to an int).
             # If it's a small number: show 4 significant digits.
-            sig_x = max(4, math.ceil(math.log10(abs(val[0]))))
-            sig_y = max(4, math.ceil(math.log10(abs(val[1]))))
+            def sig_digits(val):
```
K4rishma, line 153: It does the same thing as 153, 154, right?. Do we make a function to make it readable?

K4rishma (approved):

src/odemis/gui/comp/overlay/marking_line.py
pieleric: No. See commit message. This function now handles correctly `val == 0`, which is a pretty common "corner" case!

PR #2690, opened 2024-03-01
[fix] correlation tab should passthrough arrow controls when correlation enabled is unchecked

patrickcleeve2:
Correctly disable the arrow key based correlation shortcuts when flag is toggled. Previously did not disable, and caused repeated calls to canvas callback. Now keyboard events are passed through (skipped).

### DIFF
--- a/src/odemis/gui/cont/correlation.py
+++ b/src/odemis/gui/cont/correlation.py
@@ -238,13 +238,15 @@ class CorrelationController(object):
         """handle key presses
         :param evt: (wx.Event) the event"""
 
-        if not self.correlation_enabled():
-            logging.debug("correlation not enabled, passing event to canvas")
-            active_canvas = evt.GetEventObject()
-            active_canvas.on_char(evt)            # super event passthrough
-
+        # event data
         key = evt.GetKeyCode()
         shift_mod = evt.ShiftDown()
+
+        # pass through event, if not a valid correlation key or enabled
+        valid_keys = [wx.WXK_LEFT, wx.WXK_RIGHT, wx.WXK_UP, wx.WXK_DOWN]
+        if key not in valid_keys or not self.correlation_enabled():
+            evt.Skip()
+            return
 
         ### CONTROLS ##############################
         # SHIFT + LEFT CLICK -> MOVE_TO_POSITION


K4rishma (approved):

src/odemis/gui/cont/correlation.py
```diff
@@ -238,14 +238,22 @@ def on_char(self, evt: wx.Event) -> None:
...
-            active_canvas.on_char(evt)            # super event passthrough
-
+        # event data
         key = evt.GetKeyCode()
         shift_mod = evt.ShiftDown()
 
+        # pass event to canvas, if not a valid correlation key
+        valid_keys = [wx.WXK_LEFT, wx.WXK_RIGHT, wx.WXK_UP, wx.WXK_DOWN]
+        if evt.GetKeyCode() not in valid_keys:
+            logging.debug(f"invalid correlation key, passing event to canvas, key was: {key}")
```
pieleric, line 248: I'd drop this logging message. Essentially it's any key which is not an arrow... so pretty much one message for every key press. Too much.

src/odemis/gui/cont/correlation.py
```diff
@@ -238,14 +238,22 @@ def on_char(self, evt: wx.Event) -> None:
...
+            active_canvas = evt.GetEventObject()
+            active_canvas.on_char(evt)
```
pieleric, line 250: Looking at this code again, I don't think this is the proper way. It works because all the canvases have a `on_char` method. But I think the right way in wxWidgets is to use evt.Skip(). It tells wxWidget that this event handler is not interested, and the next event handler should take care of it (which will be the canvas, if it has bound to that event).

src/odemis/gui/cont/correlation.py
```diff
@@ -238,14 +238,22 @@ def on_char(self, evt: wx.Event) -> None:
...
+            return
+
+        if not self.correlation_enabled():
```
pieleric, line 253: So in this tab, the arrow keys will never move the canvas the standard way? Either they move a stream, or they have no effect. Is that the intention? If so, can you add a comment to clarify.

src/odemis/gui/cont/correlation.py
```diff
@@ -238,14 +238,22 @@ def on_char(self, evt: wx.Event) -> None:
...
+            logging.debug("valid correlation key, but correlation not enabled, returning...")
```
pieleric, line 254: Drop this log.

pieleric (changes requested):

pieleric (changes requested):
I don't think the commit message is very clear. Reading it, I'd assume that it modifies Odemis so that the arrow keys cannot effect the correlation. Looking at the code, that seems pretty much the inverse.
Also, it's good to give some context about which part of odemis is touched. I'd suggest something more like:
"[fix] correlation tab should never allow panning with the arrow keys"

src/odemis/gui/cont/correlation.py
patrickcleeve2: You're right, I've moved this into the above check and done the passthrough with evt.Skip

pieleric (approved):

PR #2691, opened 2024-03-01
[feat] allow user to control automatic overlay resizing

patrickcleeve2:


src/odemis/gui/cont/correlation.py
```diff
@@ -93,6 +93,7 @@ def __init__(self, tab_data, panel, tab, viewports):
 
         # enable correlation controls
         self._panel.ctrl_enable_correlation.SetValue(True) # enable by default
+        self._panel.ctrl_auto_resize_view.SetValue(False) # disable auto resize by default
```
nandishjpatel, line 96: Why not bind this with `wx.EVT_CHECKBOX`? and handle the option viewports fit to content in the callback
```
self._panel.vp_correlation_bl.canvas.fit_view_to_content()
self._panel.vp_correlation_br.canvas.fit_view_to_content()
```

patrickcleeve2: The resize doesn't happen when the user checks the box, but when they adjust the correlation so I don't think binding the event makes sense. 

I'll add a tooltip, good suggestion.

nandishjpatel (approved):

pieleric (approved):

PR #2692, opened 2024-03-01
[fix] disable scan rotation metadata for tfs

patrickcleeve2:
Users reporting all images are flipped 180 degrees. This is due to scan rotation metadata being inconsistent with idemis display. 

This PR disables (set to zero) scan rotation metadata to make displayed odemis images consistent with acquired images.

### DIFF
--- a/src/odemis/dataio/tiff.py
+++ b/src/odemis/dataio/tiff.py
@@ -2743,7 +2743,8 @@ def _convert_thermo_fisher_to_odemis_metadata(metadata: str) -> dict:
     try:
         # beam metadata
         beam_type = tfs_md["Beam"]["beam"]
-        md[model.MD_ROTATION] = float(tfs_md[beam_type]["scanrotation"])
+        # tmp: disable scan rotation metadata, as it is inconsistent with how odemis displays images
+        md[model.MD_ROTATION] = 0  # float(tfs_md[beam_type]["scanrotation"])
         md[model.MD_EBEAM_VOLTAGE] = float(tfs_md[beam_type]["hv"])
         md[model.MD_EBEAM_CURRENT] = float(tfs_md[beam_type]["beamcurrent"])
 


src/odemis/dataio/tiff.py
```diff
@@ -2743,7 +2743,8 @@ def _convert_thermo_fisher_to_odemis_metadata(metadata: str) -> dict:
     try:
         # beam metadata
         beam_type = tfs_md["Beam"]["beam"]
-        md[model.MD_ROTATION] = float(tfs_md[beam_type]["scanrotation"])
+        # tmp: disable scan rotation metadata, as it is inconsistent with how odemis displays images
+        md[model.MD_ROTATION] = 0  # float(tfs_md[beam_type]["scanrotation"])
```
tepals, line 2747: If it is consistently flipped by 180 degrees, why not just add 180 degrees to `float(tfs_md[beam_type]["scanrotation"])`?

patrickcleeve2: To be more specific, the display rotation should depend on both scan rotation and stage rotation. But stage rotation = 0 isnt the flat to sem rotation, depends on the specific microscope. So we'd need to read the configuration metadata (which isn't really something we'd want to depend on at image loading time, because it might be different (e.g. different microscope)). So for now its just easier to disable the rotation, so that users see the same images as they took

tepals (approved):

src/odemis/dataio/tiff.py
pieleric: Yes, I've seen similar issues before. The problem is that the way the SEM image should be displayed depends a lot on the rest of the system. Sometimes we really want to see it rotated, sometime it should be shown at 0°, because the scan rotation is there precisely to compensate for some other rotation! 
I think it's fine to "play dumb" and just not report the scan rotation. That's how it's shown in the SEM software anyway.

pieleric (approved):

patrickcleeve2:
I think this can be merged @tepals @pieleric

PR #2696, opened 2024-03-04
[fix] driver nfpm: SetVelocity used str instead of bytes

pieleric:
Some left-over from Python 2 times. Missed a "b" in the conversion.

It was not a big deal, as in practice no code in Odemis never changes
the speed. Still, better to fix.


patrickcleeve2 (approved):
Is there a test associated with this?

sneepstefan (approved):

pieleric:
> Is there a test associated with this?

I've just added a test case.

PR #2698, opened 2024-03-05
[fix] allow duplicate metadata keys in tfs metadata

patrickcleeve2:
Specific TFS SEM images were failing to parse correctly due the duplicate metadata keys in the metadata (e..g MultiGISGas parameters). These keys are not used by odemis, and can be discarded, but we need to disable strict parsing mode to do so.

### DIFF
--- a/src/odemis/dataio/tiff.py
+++ b/src/odemis/dataio/tiff.py
@@ -2696,11 +2696,11 @@ def _convert_thermo_fisher_to_odemis_metadata(metadata: str) -> dict:
     # parse ini metadata as dictionary
     tfs_md = {}
     try:
-        cfg = configparser.ConfigParser()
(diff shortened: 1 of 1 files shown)


src/odemis/dataio/tiff.py
```diff
@@ -2696,7 +2696,7 @@ def _convert_thermo_fisher_to_odemis_metadata(metadata: str) -> dict:
     # parse ini metadata as dictionary
     tfs_md = {}
     try:
-        cfg = configparser.ConfigParser()
+        cfg = configparser.ConfigParser(strict=False) # allow duplicate keys
```
nandishjpatel, line 2699: Can you update the warning in line `2703` stating now that parsing duplicate keys was allowed?

nandishjpatel (approved):

tepals (approved):

PR #2700, opened 2024-03-06
[fix] self_hosted_runner.yml remove display env variable

sneepstefan:
Due to the removal of the HDMI dongle from the server it is not needed to pass this environment variable to VBoxManage anymore.

This is currently causing this issue:
  Qt WARNING: could not connect to display :0
  Qt FATAL: This application failed to start because no Qt platform plugin
  could be initialized. Reinstalling the application may fix this problem.


nandishjpatel (approved):

nandishjpatel:
@tepals can you revert this PR? The changes didn't make it work https://github.com/delmic/odemis/actions/runs/8176007083/job/22354439422#step:5:38

PR #2701, opened 2024-03-07
[SDEV-1819] feature copy / paste and rotate shapes

nandishjpatel:


nandishjpatel:
This PR also contains the copy / paste shape feature

pieleric:
@nandishjpatel , could you rebase against master (instead merge), to clean-up the commits that github sees in this PR.

pieleric:
> This PR also contains the copy / paste shape feature

Please update the PR description.

nandishjpatel:
> @nandishjpatel , could you rebase against master (instead merge), to clean-up the commits that github sees in this PR.

Done, the commit history is now clean

src/odemis/gui/comp/overlay/shapes.py
```diff
@@ -164,41 +168,76 @@ def _get_shape(self, evt):
                     return shape
         return None
 
+    def _create_new_shape(self):
+        """Create a new shape."""
```
pieleric, line 172: This is a odd function which create a shape *as side-effect*. At least, I'd recommend to return the shape that was just returned. 
Maybe then the caller could do `self._selected_shape = self._create_new_shape()`, which would make things a little less "magical" when reading the code.

src/odemis/gui/cont/tabs/fastem_main_tab.py
```diff
@@ -205,6 +210,27 @@ def _toggle_user_settings_panel(self, _):
...
+        :param mode: (bool) whether the system is currently acquiring.
+        """
+        self.tab_data_model.tool.value = TOOL_NONE
+        self.tb.enable(not mode)
+
+    def _on_tool(self, selected_tool):
+        """Pause the SEM stream for a specific set of tools."""
+        if selected_tool in (TOOL_RECTANGLE, TOOL_ELLIPSE, TOOL_POLYGON):
+            if (
+                hasattr(self.tab_data_model, "semStream")
```
pieleric, line 228: Is there any reason that tab_data_model doesn't have semStream? How come the fastem (only) tab wouldn't have a SEM stream?

src/odemis/gui/cont/tabs/fastem_main_tab.py
```diff
@@ -205,6 +210,27 @@ def _toggle_user_settings_panel(self, _):
...
+
+        :param mode: (bool) whether the system is currently acquiring.
+        """
+        self.tab_data_model.tool.value = TOOL_NONE
+        self.tb.enable(not mode)
+
+    def _on_tool(self, selected_tool):
+        """Pause the SEM stream for a specific set of tools."""
+        if selected_tool in (TOOL_RECTANGLE, TOOL_ELLIPSE, TOOL_POLYGON):
+            if (
```
pieleric, line 227: use `and` with the previous `if`, that'd be more direct.

src/odemis/gui/comp/overlay/world_select.py
```diff
@@ -239,3 +242,268 @@ def on_motion(self, evt):
             WorldOverlay.on_motion(self, evt)
 
     # END Event Handlers
+
+
+class RectanglePointsSelectOverlay(WorldOverlay, RectangleMixin):
```
pieleric, line 247: Can you add a docstring to this class to explain what it does, or why you'd want to use it?

src/odemis/gui/comp/overlay/world_select.py
```diff
@@ -239,3 +242,268 @@ def on_motion(self, evt):
...
+        WorldOverlay.__init__(self, cnvs)
+        RectangleMixin.__init__(self, colour, center)
+
+        self.p_point1 = None
+        self.p_point2 = None
+        self.p_point3 = None
+        self.p_point4 = None
+
+        self._side1_label = self.add_label("", colour=(0.8, 0.8, 0.8), align=wx.ALIGN_RIGHT)
+        self._side2_label = self.add_label("", colour=(0.8, 0.8, 0.8), align=wx.ALIGN_RIGHT)
```
pieleric, line 259: Please explain what side 1 & 2 refer to. (typically a rectangle has 4 sides, so it's a bit confusing).
It'd be also useful to indicate when these labels are used. They are created empty.

pieleric (changes requested):

src/odemis/gui/comp/overlay/base.py
```diff
@@ -1109,42 +1135,43 @@ def __init__(self, colour=gui.SELECTION_COLOUR, center=(0, 0)):
...
 
-        self.v_edges = {}
+        # Dict in which key is the gui.HOVER_* type and value is a list of EdgeBoundingBox
+        # associated to the hover type
+        self.v_edges: Dict[int, List[EdgeBoundingBox]] = {}
 
         # TODO: Move these to the super classes
         self.colour = conversion.hex_to_frgba(colour)
         self.highlight = conversion.hex_to_frgba(gui.FG_COLOUR_HIGHLIGHT)
         self.center = Vec(center)
```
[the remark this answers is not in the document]
nandishjpatel: Yeah the `.center` and `.rotation` attributes are used by external classes specifically during drawing the rotation label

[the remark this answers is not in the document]
pieleric: Which class is that? Isn't it some inheriting classes? As long as these are classes that inherit from the Mixin, it would be fine to mark these as private. The opposite would be to leave them "public" so that the rest of the GUI can depend on these attributes to know the center and rotation.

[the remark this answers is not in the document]
nandishjpatel: > The opposite would be to leave them "public" so that the rest of the GUI can depend on these attributes to know the center and rotation.
When I start working on the Project Manager, the center and rotation will mostly become a VA. Users can input the values and the shape will be shifted or rotated accordingly.

tepals (approved):

src/odemis/gui/comp/overlay/base.py
```diff
@@ -1079,6 +1106,444 @@ def _on_motion(self, evt):
...
+        dx = abs(point1.x - point2.x)
+        dy = abs(point1.y - point2.y)
+        return gui.HOVER_DIRECTION_EW if dy > dx else gui.HOVER_DIRECTION_NS
+
+    def get_hover(self, vpos) -> Tuple[int, Optional[int]]:
+        """
+        Check if the given position is on/near a selection edge or inside the selection.
+
+        :return: (tuple) the first element is False if not hovering, or the type of hover.
+            The second element is the index of the v_point if the edit mode is EDIT_MODE_POINT.
```
pieleric, line 1465: When returning multiple values, it's nicer to give a name to each variable, and explain independently.
```suggestion
        :return:
           hover_type (HOVER_*): the type/location of hover, or HOVER_NONE if not hovering, 
           edge: the index of the v_point involved, if the edit mode is EDIT_MODE_POINT. Otherwise None.
```

src/odemis/gui/comp/overlay/base.py
```diff
@@ -1107,6 +1572,7 @@ def __init__(self, colour=gui.SELECTION_COLOUR, center=(0, 0), edit_mode=EDIT_MO
         self.edit_v_start_pos = None  # The view port coordinates where a drag/edit originated
         self.edit_hover = None  # What edge is being edited (gui.HOVER_*)
         self.edit_mode = edit_mode
+        self.edit_v_point_idx = None
```
pieleric, line 1575: Please add type and short description in comment (something like "int, index of the edge being moved")

src/odemis/gui/comp/overlay/base.py
```diff
@@ -1205,35 +1694,73 @@ def update_projection(self, idx, b_pos, shiftscale):
             logging.debug("Updating view position of %s, shiftscale %s", idx, shiftscale)
             self.v_points[idx] = Vec(self.cnvs.buffer_to_view(b_pos))
 
+    def _calc_center(self):
+        """Calculate the center of selection."""
+        n = len(self.v_points)
+        sum_x = sum(p.x for p in self.v_points)
+        sum_y = sum(p.y for p in self.v_points)
+        centroid_x = sum_x / n
+        centroid_y = sum_y / n
```
pieleric, line 1703: Nowadays, you can use the `statistics` module from the standard library:
`centroid_x = statistics.mean(p.x for p in self.v_points)`

src/odemis/gui/comp/overlay/ellipse.py
```diff
@@ -78,37 +114,47 @@ def on_left_up(self, evt):
...
             self.cnvs.update_drawing()
-            if self.selected.value:
+            # The rotation point is outside the shape and cannot be captured by selection VA
+            # Update the points VA if the selection mode is SEL_MODE_ROTATION
+            if self.selected.value or is_rotation:
                 self.points.value = self._points
         WorldOverlay.on_left_up(self, evt)
 
-    def draw(self, ctx, shift=(0, 0), scale=1.0):
+    def draw(self, ctx, shift=(0, 0), scale=1.0, dash=False):
```
pieleric, line 123: Every other `draw` function with a `dash` argument have it default to `dash=True`. Please do the same here.
While at it, you can indicate the type is a `bool`.

src/odemis/gui/comp/overlay/ellipse.py
```diff
@@ -78,37 +114,47 @@ def on_left_up(self, evt):
...
-            ellipse_center_x = (b_start_pos.x + b_end_pos.x) / 2
-            ellipse_center_y = (b_start_pos.y + b_end_pos.y) / 2
-            # Semi-major axis
-            a = abs(b_start_pos.x - b_end_pos.x) / 2
-            # Semi-minor axis
-            b = abs(b_start_pos.y - b_end_pos.y) / 2
+            ellipse_center_x = (b_point1.x + b_point3.x) / 2
+            ellipse_center_y = (b_point1.y + b_point3.y) / 2
+            # Calculate the side lengths to find the semi-major and semi-minor axis
+            side_length1 = math.sqrt((b_point1.x - b_point2.x)**2 + (b_point1.y - b_point2.y)**2)
```
pieleric, line 150:
```suggestion
            side_length1 = math.hypot(b_point1.x - b_point2.x, b_point1.y - b_point2.y)
```

src/odemis/gui/comp/overlay/shapes.py
```diff
@@ -84,6 +77,24 @@ def draw(self, ctx, shift=(0, 0), scale=1.0):
...
+
+    @abstractmethod
+    def move_to(self, pos: Union[Tuple[float, float], Vec]):
+        """Move the shape's center to a physical position."""
+        pass
+
+    @abstractmethod
+    def refresh(self):
+        """Refresh the shape's points VA after copy or move_to operation."""
+        pass
```
pieleric, line 96: What's the reason for this method? Wouldn't it be better to automatically call it at the end of copy() and move_to()? For now that means that copy and move_to generate (temporary) invalid objects. Is it to save a bit of computation (by not doing it twice) ? If so, as it is used now, I would argue it's a not a big deal. This is done only when the user copy-paste a shape (which probably takes multiple seconds, and that saves a few µs. 
So I'd advocate of dropping this .refresh() method to keep the API simpler & safer.

src/odemis/gui/comp/overlay/ellipse.py
```diff
@@ -48,27 +47,64 @@ def __init__(self, cnvs, colour=gui.SELECTION_COLOUR):
...
+        :returns: (EllipseOverlay) a new instance of EllipseOverlay with necessary copied attributes.
+
+        """
+        shape = EllipseOverlay(self.cnvs)
+        shape.colour = self.colour
+        shape.p_point1 = self.p_point1
+        shape.p_point2 = self.p_point2
+        shape.p_point3 = self.p_point3
+        shape.p_point4 = self.p_point4
+        shape._points = self._points.copy()
```
pieleric, line 62: This whole `.points` vs `._points` is confusing. Isn't it possible to only use `.points`?  
Or at least, from the code  is seems that ._points is recomputed automatically by `.draw` from the p_points. So they are not so useful. Why should they be copied? Shouldn't `.points` (VA content) be copied instead? 
Or have an explicit function that computes the points based on the p_points? (Then `draw()` would use it to update the points, before converting them to buffer coordinates)

pieleric (changes requested):

src/odemis/gui/comp/overlay/ellipse.py
nandishjpatel: I agree, it can be confusing. I am using `._points` for any internal point manipulation. `points` VA value is set post any points manipulation and if the shape is selected. Another reason is to avoid some unnecessary calculation if VA value changes often. For rectangle I can definitely get rid of `._points` and also for ellipse, but then I need it for the polygon.

In `EditableShape` its written as a comment.

```
       # list of nested points (x, y) representing the shape and whose value will be used
        # during ROA acquisition
        # The points VA is set to _points if the shape is selected
        self.points = model.ListVA()
        # Useful for internal points manipulation
        self._points: List[Vec] = []
```

src/odemis/gui/comp/overlay/ellipse.py
```diff
@@ -48,27 +47,61 @@ def __init__(self, cnvs, colour=gui.SELECTION_COLOUR):
...
+        shape.points.value = shape._points
```
pieleric, line 63: Why don't you call `self._phys_to_view()`, as done after `move_to()`?

pieleric (approved):

pieleric (approved):

PR #2705, opened 2024-03-12
[feature] driver andorshrk: allow to not complain if an axis fails to move

pieleric:
Add "check_move" argument to disable the check of a move having
succeded. This is a workaround an issue on some spectrographs which
do report a bad position of the flipper, while the flipper is actually
at the right place. Interestingly, that check was introduced in commit
09e346b5be to handle other spectrographs which sometimes failed to move
properly.

### DIFF
--- a/src/odemis/driver/andorshrk.py
+++ b/src/odemis/driver/andorshrk.py
@@ -334,7 +334,7 @@ class Shamrock(model.Actuator):
           "led_prot", it will set the .protection to True any time that the
           slit leds could be turned on.
         :param check_move: Name of the axis -> bool: If True (default), after move, raise an error if
-        that the position is not the expected one. If False, if the position is not the expected one,
+        that position is not the expected one. If False, if the position is not the expected one,
         just log a warning. Note: for now, only some axes are actually checked (flipper and band)
         """
         # From the documentation:
--- a/src/odemis/driver/andorshrk.py
+++ b/src/odemis/driver/andorshrk.py
@@ -20,6 +20,8 @@ from ctypes import *
 import ctypes
 import logging
 import math
+from typing import Optional, Dict
+
 from odemis import model
 import odemis
 from odemis.driver import andorcam2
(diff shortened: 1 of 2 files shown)


src/odemis/driver/andorshrk.py
```diff
@@ -328,6 +333,9 @@ def __init__(self, name, role, device, camera=None, accessory=None,
         dependencies (None or dict str -> HwComponent): if the key starts with
           "led_prot", it will set the .protection to True any time that the
           slit leds could be turned on.
+        :param check_move: Name of the axis -> bool: If True (default), after move, raise an error if
```
tepals, line 336: While your adjusting this docstring, could you fix the order and formatting of the whole docstring? And remove the `inverted` argument from the docstring.

src/odemis/driver/andorshrk.py
```diff
@@ -328,6 +333,9 @@ def __init__(self, name, role, device, camera=None, accessory=None,
...
+        that the position is not the expected one. If False, if the position is not the expected one,
```
tepals, line 337:
```suggestion
        that position is not the expected one. If False, if the position is not the expected one,
```

tepals (changes requested):

src/odemis/driver/andorshrk.py
pieleric: Well, I'd rather do it separately. I already have another change (iris support) in the make, that will add yet another parameter. I can clean up the docstring when I create a PR for this change.
As for the `inverted` argument, normally every Actuator accepts it (via the `**kwargs`), but it doesn't work here. So it has to stay documented that this is *not* supported.

tepals: Okay! Although I would then add the part about the `inverted` argument all the way at the end as a note. I don't use the `inverted` argument often in other Actuators, so to me this description made no sense.

tepals (approved):

sneepstefan (changes requested):

src/odemis/driver/andorshrk.py
```diff
@@ -294,8 +296,11 @@ class Shamrock(model.Actuator):
     Note: we don't handle changing turret (live).
     """
     def __init__(self, name, role, device, camera=None, accessory=None,
-                 slitleds_settle_time=1e-3, slits=None, bands=None, rng=None,
-                 fstepsize=1e-6, drives_shutter=None, dependencies=None, **kwargs):
+                 slitleds_settle_time=1e-3, slits=None,
+                 bands=None, rng=None,
+                 fstepsize=1e-6, drives_shutter=None, dependencies=None,
```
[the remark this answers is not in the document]
pieleric: My coming change (separate from this PR) will actually "fix" it by placing one argument per line.

sneepstefan (approved):

PR #2706, opened 2024-03-14
[feat] extend metadata for fibsem specific parameters

patrickcleeve2:
This pull request extends the existing metadata to represent the full state of a fibsem system. This should allow decoupling of different metadata components, and separate the system metadata from the metadata used for display (e.g. MD_POS vs MD_STAGE_POSITION). All metadata represents the state of the system that was used to acquire the image.

- MD_STAGE_POSITION: The standard 5 dimensional stage position (x, y, z, r, t)
- MD_STAGE_POSITION_EXT: Extended stage position metadata for representing additional axes (e.g. zeiss m, alpha/beta tilt stages)
- MD_STAGE_PRE_TILT: The pre-tilt of the stage / shuttle used. 
- MD_BEAM_SCAN_ROTATION: the scan rotation of the beam
- MD_BEAM_STIGMATOR: the stigmator values of the beam
- MD_BEAM_SHIFT: the beam shift values of the beam
- MD_BEAM_WORKING_DISTANCE: the working distance of the beam
- MD_BEAM_FIELD_OF_VIEW: the field of view of the beam
- MD_BEAM_SCANNING_MODE: the scanning mode of the beam
- MD_BEAM_COLUMN_TILT: the tilt of the beam column

### DIFF
--- a/src/odemis/model/_metadata.py
+++ b/src/odemis/model/_metadata.py
@@ -61,11 +61,34 @@ MD_LENS_NA = "Lens numerical aperture"  # float (ratio), numerical aperture
 MD_LENS_RI = "Lens refractive index"  # float (ratio), refractive index
 MD_FILTER_NAME = "Filter name"  # str, product name of the light filter
 # TODO: might need to merge DWELL_TIME and EXP_TIME into INTEGRATION_TIME: the time each pixel receive energy
+
+
 # + SCANNED_DIMENSIONS: list of dimensions which were scanned instead of being acquired simultaneously
-MD_DWELL_TIME = "Pixel dwell time"  # s (float), time the electron beam spends per pixel
-MD_EBEAM_VOLTAGE = "Electron beam acceleration voltage"  # V (float), voltage used to accelerate the electron beam
-MD_EBEAM_CURRENT = "Electron beam emission current"  # A (float), emission current of the electron beam (typically, the probe current is a bit smaller and the spot diameter is linearly proportional)
-MD_EBEAM_SPOT_DIAM = "Electron beam spot diameter"  # m (float), approximate diameter of the electron beam spot (typically function of the current)
+MD_BEAM_SCAN_ROTATION = "Beam scan rotation"  # rad (float) rotation of the beam scan
+# scan rotation is a subset of MD_ROTATION which is also influenced by stage rotation
+MD_BEAM_STIGMATOR = "Beam stigmator"  # (m, m) stigmator values
+MD_BEAM_SHIFT = "Beam shift"  # (m, m) shift of the beam
+MD_BEAM_SOURCE_TILT = "Beam source tilt"  # (m, m) tilt of the beam source
+MD_BEAM_WORKING_DISTANCE = "Beam working distance"  # m (float), working distance of the beam
+MD_BEAM_FIELD_OF_VIEW = "Beam field of view"  # m (float), horizontal field of view of the beam
+MD_BEAM_SCANNING_MODE = "Beam scanning mode"  # str, mode of the beam scanning
+MD_BEAM_DWELL_TIME = "Pixel dwell time"  # s (float), time the beam spends per pixel
+MD_BEAM_VOLTAGE = "Electron beam acceleration voltage"  # V (float), voltage used to accelerate the beam
+MD_BEAM_CURRENT = "Electron beam emission current"  # A (float), emission current of the beam
+# (typically, the probe current is a bit smaller and the spot diameter is linearly proportional)
+MD_BEAM_SPOT_DIAM = "Electron beam spot diameter"  # m (float), approximate diameter of the beam spot
+MD_BEAM_COLUMN_TILT = "Beam column tilt"  # (rad) tilt of the beam column
+
+# deprecated: use MD_BEAM_* instead
+MD_DWELL_TIME = MD_BEAM_DWELL_TIME
+MD_EBEAM_VOLTAGE = MD_BEAM_VOLTAGE
+MD_EBEAM_CURRENT = MD_BEAM_CURRENT
+MD_EBEAM_SPOT_DIAM = MD_BEAM_SPOT_DIAM
+
+# FIB-SEM metadata
+# position of the stage (in m or rad) for each axis in the chamber (raw hardware values)
+MD_STAGE_POSITION_RAW = "Stage position raw"  # dict of str -> float,
+MD_SAMPLE_PRE_TILT = "Sample pre-tilt"  # (rad) pre-tilt of the sample stage / shuttle (tilt)
 
 MD_STREAK_TIMERANGE = "Streak Time Range"  # (s) Time range for one streak/sweep
 MD_STREAK_MCPGAIN = "Streak MCP Gain"  # (int) Multiplying gain for microchannel plate


pieleric (changes requested):
MD_STAGE_PRE_TILT: Sounds good, it can be useful for describing the systems better. However, I think it would be clearer to name it `MD_SAMPLE_PRE_TILT`, as it typically doesn't affect the stage axes, but it's more something between the stage and the sample.

MD_BEAM_*: In general, I'm fine with them. However, as we already have several MD_EBEAM_* metadata, I'd rather have them placed all together, and all called MD_EBEAM_* (although it's now generalized to the ion-beam too). You can indicate that in a comment.

MD_BEAM_COLUMN_TILT: is that the same as "source tilt"? Is it really stored as an angle? Isn't there typically 2 dimensions?

MD_BEAM_FIELD_OF_VIEW: normally we store just PIXEL_SIZE, and multiply by the array shape to know the field of view. I guess that if you scan a RoI, then it's not equivalent, but I don't really see the point to store this information explicitly. We also have MD_LENS_MAG to store the exact setting used in the SEM. => Do we really need that or can we avoid?

MD_STAGE_POSITION*: I'm not found of these two new metadata. It's just too confusing. MD_POS is supposed to indicate that. We also have MD_POS_COR to adjust the display. I'd rather avoid going with another 2 metadata! Instead, I'd suggest to use MD_EXTRA_SETTINGS with a JSON containing the stage.position dict. Then you can just store the known axes -> position. (so no need to store values for axes that don't exist) It would make it clearer that this is really "raw" data, and to be used only for very special cases.


    MD_STAGE_POSITION: The standard 5 dimensional stage position (x, y, z, r, t)
    MD_STAGE_POSITION_EXT: Extended stage position metadata for representing additional axes (e.g. zeiss m, alpha/beta tilt stages)

    MD_BEAM_SCAN_ROTATION: the scan rotation of the beam
    MD_BEAM_STIGMATOR: the stigmator values of the beam
    MD_BEAM_SHIFT: the beam shift values of the beam
    MD_BEAM_WORKING_DISTANCE: the working distance of the beam
    MD_BEAM_FIELD_OF_VIEW: the field of view of the beam
    MD_BEAM_SCANNING_MODE: the scanning mode of the beam
    MD_BEAM_COLUMN_TILT: the tilt of the beam column

patrickcleeve2:
> MD_STAGE_PRE_TILT: Sounds good, it can be useful for describing the systems better. However, I think it would be clearer to name it `MD_SAMPLE_PRE_TILT`, as it typically doesn't affect the stage axes, but it's more something between the stage and the sample.
> 
Agree, updated

> `MD_BEAM_*`: In general, I'm fine with them. However, as we already have several `MD_EBEAM_*` metadata, I'd rather have them placed all together, and all called `MD_EBEAM_*` (although it's now generalized to the ion-beam too). You can indicate that in a comment.
> 
I would think it would be better to deprecate MD_EBEAM in favour of MD_BEAM if they are going to represent both beams? We should keep the MD_EBEAM variables for backwards compat, but point them at the new equivalent MD_BEAM variables. I've updated the pr to show what I mean

> MD_BEAM_COLUMN_TILT: is that the same as "source tilt"? Is it really stored as an angle? Isn't there typically 2 dimensions?
> 
No, this is the actually hardware angle that the column is mounted at (SEM=0, FIB=52, 54, 55) depending on the manufacturer. It is required to determine the imaging perspective. I can also add source tilt to the MD_BEAM_ parameters

> MD_BEAM_FIELD_OF_VIEW: normally we store just PIXEL_SIZE, and multiply by the array shape to know the field of view. I guess that if you scan a RoI, then it's not equivalent, but I don't really see the point to store this information explicitly. We also have MD_LENS_MAG to store the exact setting used in the SEM. => Do we really need that or can we avoid?
> 
I don't know if its strictly required since it can be computed. But I find it's a more interpretable piece of information than pixel size * shape

> `MD_STAGE_POSITION*`: I'm not fond of these two new metadata. It's just too confusing. MD_POS is supposed to indicate that. We also have MD_POS_COR to adjust the display. I'd rather avoid going with another 2 metadata! Instead, I'd suggest to use MD_EXTRA_SETTINGS with a JSON containing the stage.position dict. Then you can just store the known axes -> position. (so no need to store values for axes that don't exist) It would make it clearer that this is really "raw" data, and to be used only for very special cases.
> 
Agree, I think this could be a dict with the axes mapping, rather than a list with pre-determined values, that would be simpler. But I don't think this represents the same thing as MD_POS, or is a special case. To properly place the images in space, all the stage axes need to be considered, and MD_POS just contains x, y, and assumes a flat plane perspective. For example, representing stage rotation and scan rotation independently isn't currently possible

src/odemis/model/_metadata.py
```diff
@@ -61,11 +61,41 @@
...
-MD_DWELL_TIME = "Pixel dwell time"  # s (float), time the electron beam spends per pixel
-MD_EBEAM_VOLTAGE = "Electron beam acceleration voltage"  # V (float), voltage used to accelerate the electron beam
-MD_EBEAM_CURRENT = "Electron beam emission current"  # A (float), emission current of the electron beam (typically, the probe current is a bit smaller and the spot diameter is linearly proportional)
-MD_EBEAM_SPOT_DIAM = "Electron beam spot diameter"  # m (float), approximate diameter of the electron beam spot (typically function of the current)
+MD_BEAM_SCAN_ROTATION = "Beam scan rotation"  # (rad) rotation of the beam scan
+MD_BEAM_STIGMATOR = "Beam stigmator"  # (m, m) stigmator values
+MD_BEAM_SHIFT = "Beam shift"  # (m, m) shift of the beam
+MD_BEAM_SOURCE_TILT = "Beam source tilt"  # (m, m) tilt of the beam source
+MD_BEAM_WORKING_DISTANCE = "Beam working distance"  # m, working distance of the beam
+MD_BEAM_FIELD_OF_VIEW = "Beam field of view"  # (m) horizontal field of view of the beam
```
tepals, line 75: This file is a bit of a mess, but I think in most cases only if there are multiple values for the unit, like `(m, m)`, the unit is written in brackets.

```suggestion
MD_BEAM_FIELD_OF_VIEW = "Beam field of view"  # m (float), horizontal field of view of the beam
```

src/odemis/model/_metadata.py
```diff
@@ -61,11 +61,41 @@
...
+MD_BEAM_SCANNING_MODE = "Beam scanning mode"  # str, mode of the beam scanning
```
tepals, line 76:
```suggestion
MD_BEAM_SCANNING_MODE = "Beam scanning mode"  # (str), mode of the beam scanning
```

src/odemis/model/_metadata.py
```diff
@@ -61,11 +61,41 @@
...
+# extended fibsem metadata
+MD_STAGE_POSITION = "Stage position"  # dict of str -> float, position of the stage (in m/rad) for each axis
+MD_SAMPLE_PRE_TILT = "Stage pre-tilt"  # (rad) pre-tilt of the stage / shuttle (tilt)
+
 # + SCANNED_DIMENSIONS: list of dimensions which were scanned instead of being acquired simultaneously
-MD_DWELL_TIME = "Pixel dwell time"  # s (float), time the electron beam spends per pixel
-MD_EBEAM_VOLTAGE = "Electron beam acceleration voltage"  # V (float), voltage used to accelerate the electron beam
-MD_EBEAM_CURRENT = "Electron beam emission current"  # A (float), emission current of the electron beam (typically, the probe current is a bit smaller and the spot diameter is linearly proportional)
-MD_EBEAM_SPOT_DIAM = "Electron beam spot diameter"  # m (float), approximate diameter of the electron beam spot (typically function of the current)
+MD_BEAM_SCAN_ROTATION = "Beam scan rotation"  # (rad) rotation of the beam scan
```
tepals, line 70: Could you extend the comment to explain how this is different than `MD_ROTATION`

src/odemis/model/_metadata.py
```diff
@@ -61,11 +61,41 @@
...
+# extended fibsem metadata
+MD_STAGE_POSITION = "Stage position"  # dict of str -> float, position of the stage (in m/rad) for each axis
+MD_SAMPLE_PRE_TILT = "Stage pre-tilt"  # (rad) pre-tilt of the stage / shuttle (tilt)
+MD_BEAM_SCAN_ROTATION = "Beam scan rotation"  # (rad) rotation of the beam scan
+MD_BEAM_STIGMATOR = "Beam stigmator"  # (m, m) stigmator values
+MD_BEAM_SHIFT = "Beam shift"  # (m, m) shift of the beam
+MD_BEAM_WORKING_DISTANCE = "Beam working distance"  # m, working distance of the beam
+MD_BEAM_FIELD_OF_VIEW = "Beam field of view"  # (m) horizontal field of view of the beam
+MD_BEAM_SCANNING_MODE = "Beam scanning mode"  # str, mode of the beam scanning
+MD_BEAM_COLUMN_TILT = "Beam column tilt"  # (rad) tilt of the beam column
```
tepals, line 98: these are all duplicates

tepals (changes requested):
Could you also fix the linting?

src/odemis/model/_metadata.py
```diff
@@ -61,11 +61,32 @@
 MD_LENS_RI = "Lens refractive index"  # float (ratio), refractive index
 MD_FILTER_NAME = "Filter name"  # str, product name of the light filter
 # TODO: might need to merge DWELL_TIME and EXP_TIME into INTEGRATION_TIME: the time each pixel receive energy
+
+# extended fibsem metadata
+MD_STAGE_POSITION = "Stage position"  # dict of str -> float, position of the stage (in m/rad) for each axis
```
pieleric, line 66: I still it's too confusing with MD_POS. It should be clear that it's very rarely useful information. I'd suggest to name it `MD_STAGE_POSITION_RAW = "Stage position raw` . 
Add in the comment that it's the position *in the chamber*, using the values as read from the hardware.

For the unit, can you change the wording from "m/rad" (which sounds like a unit) to "m or rad".

pieleric (changes requested):

tepals (approved):

src/odemis/model/_metadata.py
```diff
@@ -61,11 +61,32 @@
 MD_LENS_RI = "Lens refractive index"  # float (ratio), refractive index
 MD_FILTER_NAME = "Filter name"  # str, product name of the light filter
 # TODO: might need to merge DWELL_TIME and EXP_TIME into INTEGRATION_TIME: the time each pixel receive energy
+
+# extended fibsem metadata
```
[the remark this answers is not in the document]
patrickcleeve2: removed 'extended'

pieleric (approved):

PR #2712, opened 2024-03-19
[feature] plugin secom_cl: don't adjust the RoI when changing repetition

pieleric:
This is an advanced plugin for advanced acquisition. The pixel size
doesn't have to be square. It's more helpful to make repetition independent
of the RoI so that it's easy to select any value.

### DIFF
--- a/plugins/secom_cl.py
+++ b/plugins/secom_cl.py
@@ -457,8 +457,7 @@ class SECOMCLSettingsStream(acqstream.CCDSettingsStream):
             _, rep, pxs = self._updateROIAndPixelSize((0, 0, 1, 1), pxs)
             return roi, rep, pxs
 
-        # compute the ROI.
-        roi = self._fitROI(roi)
+        # Note: don't update the RoI, as in this plugin we want to keep it as fixed as possible.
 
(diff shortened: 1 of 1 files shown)


plugins/secom_cl.py
```diff
@@ -457,8 +457,8 @@ def _updateROIAndPixelSize(self, roi, pxs):
             _, rep, pxs = self._updateROIAndPixelSize((0, 0, 1, 1), pxs)
             return roi, rep, pxs
 
-        # compute the ROI.
-        roi = self._fitROI(roi)
+        # compute the ROI: disabled, as we want to keep the ROI independent of the repetition
+        # roi = self._fitROI(roi)
```
nandishjpatel, line 461: This function is also called on line `485` after first calling `_adaptROI` and also finding `rep` which is compatible to the hardware. Maybe instead of commenting out it is safe to remove?

nandishjpatel (approved):

sneepstefan (approved):

plugins/secom_cl.py
pieleric: Indeed, that was a little bit rushed. I've now replaced it by a comment explaining why we don't do that (while the standard version of this function does it).

PR #2716, opened 2024-03-19
[feature] odemis-park-mirror: support simulation

pieleric:
Allow to pass TEST_NOHW=1 to run in simulation mode. Useful for testing.

### DIFF
--- a/util/odemis-park-mirror
+++ b/util/odemis-park-mirror
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 # Attempt to park the mirror of the SPARCv2, even if no backend is running
-'''
+"""
 Created on October 2015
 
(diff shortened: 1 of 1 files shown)


util/odemis-park-mirror
```diff
@@ -135,4 +145,5 @@ def main(args):
 
 if __name__ == '__main__':
     ret = main(sys.argv)
+    time.sleep(5)
```
patrickcleeve2, line 148: Whats the reason for the sleep?

patrickcleeve2 (approved):

sneepstefan (approved):

util/odemis-park-mirror
pieleric: Good catch. It was for debug. I forgot it... fixed now!

PR #2717, opened 2024-03-19
[fix] at startup, always show the components in the same order

pieleric:
Until now, the component shown in the back-end startup window were in random order. Ideally, we would show them in the same order they are started... but it turns out it's pretty hard to know in advance. So at least, show them in alphabetical order, so that each run looks similar. It's less confusing for the users.

### DIFF
--- a/src/odemis/odemisd/start.py
+++ b/src/odemis/odemisd/start.py
@@ -570,7 +570,7 @@ class BackendStarter(object):
     def _on_ghosts(self, ghosts):
         """ Called when the .ghosts changes """
         # The components running fine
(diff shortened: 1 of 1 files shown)


sneepstefan (approved):

tepals:
Can you rebase on the latest master so the unit testing will most likely pass?

tepals (approved):

pieleric:
> Can you rebase on the latest master so the unit testing will most likely pass?

If you look in the details, the only failure is in the Fastem ROA test case. So yes, it will pass if I rebase. Please just disregard that test.

PR #2719, opened 2024-03-20
[fix] PixelAngularSpectrumProjection failed to project with Fit-to-RGB tint

pieleric:
The special tint Fit-to-RGB has no meaning for this projection.
So it wasn't handled... and prevented the data from being shown.

=> Do the same as in LineSpectrumProjection: use white tint in such
case.

### DIFF
--- a/src/odemis/acq/stream/_projection.py
+++ b/src/odemis/acq/stream/_projection.py
@@ -1577,7 +1577,11 @@ class PixelTemporalSpectrumProjection(RGBProjection):
         try:
             data = self._computeSpec()
             if data is not None:
-                self.image.value = self._project2RGB(data, self.stream.tint.value)
+                tint = self.stream.tint.value
+                if tint == TINT_FIT_TO_RGB:
+                    # Not supported for this view => fallback to greyscale
+                    tint = (255, 255, 255)
+                self.image.value = self._project2RGB(data, tint)
             else:
                 self.image.value = None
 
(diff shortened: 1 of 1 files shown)


src/odemis/acq/stream/_projection.py
```diff
@@ -1660,7 +1660,11 @@ def _updateImage(self):
         try:
             data = self._computeSpec()
             if data is not None:
-                self.image.value = self._project2RGB(data, self.stream.tint.value)
+                tint = self.stream.tint.value
+                if tint == TINT_FIT_TO_RGB:
+                    # Not supported for this view => fallback to greyscale
+                    tint = (255, 255, 255)
+                self.image.value = self._project2RGB(data, tint)
```
K4rishma, line 1667: self._project2RGB is used in other parts of this file, do we need this change only here and not in the rest?

src/odemis/acq/stream/_projection.py
```diff
@@ -1660,7 +1660,11 @@ def _updateImage(self):
         try:
             data = self._computeSpec()
             if data is not None:
-                self.image.value = self._project2RGB(data, self.stream.tint.value)
+                tint = self.stream.tint.value
+                if tint == TINT_FIT_TO_RGB:
+                    # Not supported for this view => fallback to greyscale
```
patrickcleeve2, line 1665: Would it make sense to log this info, so that it doesn't just change silently without any feedback for the user?

patrickcleeve2 (approved):

src/odemis/acq/stream/_projection.py
pieleric: Well, that's mostly due to a limitation/simplification in the GUI: the user can only select 1 type of colormap/tint per stream, although the stream has multiple projections (in this case, there are 4 possible projections, typically all shown simultaneously on the screen). So the GUI allows to select "Fit to RGB" as for one projection it's meaningful. For all the other ones, it doesn't make sense, and we just use the "standard" tint instead. I could add a log message, but no one is going to look at it. I could make it a actual pop up in the GUI and it's just going to annoying everyone... so I think doing nothing is fine.

K4rishma (approved):

src/odemis/acq/stream/_projection.py
pieleric: Good catch. "Fit-To-RGB" can be set on any spectrum stream. So we need to also protect `PixelTemporalSpectrumProjection` .

PR #2720, opened 2024-03-20
[fix][METEOR-978] correct the image order in ome tiff files based on ImageJ compa…

K4rishma:
…tibility

- change the ome information and writing of tiff files
- fix grouping of images of the same detector type
- add tiff test cases

### DIFF
--- a/src/odemis/dataio/tiff.py
+++ b/src/odemis/dataio/tiff.py
@@ -1902,7 +1902,7 @@ def _saveAsMultiTiffLT(filename, ldata, thumbnail, compressed=True, multiple_fil
                 data = numpy.moveaxis(data, 0, 2)  # move C axis near YX
             elif data.ndim == 5:
                 write_rgb = False
-                # CTZYX -> TZCYX  (for ImajeJ compatible ordering)
+                # CTZYX -> TZCYX  (for ImageJ compatible ordering)
                 data = numpy.moveaxis(data, 0, 2)
                 hdim = data.shape[:-2]
             else:  # YX
@@ -1922,15 +1922,13 @@ def _saveAsMultiTiffLT(filename, ldata, thumbnail, compressed=True, multiple_fil
                     c = compression
                 write_image(f, data[i], write_rgb=write_rgb, compression=c, pyramid=pyramid)
 
-        else:  # len(das) > 1 => list of DataArrays to represent the C dimension
-            # the shape of all das are same: 1TZYX
+        else:
+            # len(das) > 1 => list of DataArrays to represent the C dimension
+            # the shape of all das are same: 1TZYX, or ZYX, or TZYX
             # len(das) is to be used as C
 
-            data = das[0]
-            assert data.shape[0] == 1
             write_rgb = False
-            # 1TZYX
-            hdim = data.shape[:-2]  # 1TZ
+            hdim = das[0].shape[:-2]  # 1TZ or TZ or Z
 
             for i in numpy.ndindex(*hdim):
                 for data in das:  # for ImageJ compatible ordering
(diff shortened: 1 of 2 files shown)


patrickcleeve2:
@K4rishma do you have any test images for this?

K4rishma:
> @K4rishma do you have any test images for this?

There are test cases with synthetic images to check some of the compatibility with Image J. These synthetic images can also be used for opening both in Image J and Odemis viewer. I can send the real images by chat.

src/odemis/dataio/tiff.py
```diff
@@ -1208,14 +1208,15 @@ def _findImageGroups(das):
 
     for da in das:
         # check if it can be part of the current group (compare just to the previous DA)
+        dims = da.metadata.get(model.MD_DIMS, "CTZYX"[-da.ndim::])
         if (prev_da is None
-            or da.shape[0] != 1  # If C != 1 => not possible to merge (C is always first dimension)
+            or ("C" in dims and da.shape[dims.index("C")] != 1)  # If C != 1 => not possible to merge (C is always first dimension)
+            or (da.shape[-2] != 1 and len(da.shape) == 2)  # thumbnail image of format YX
```
pieleric, line 1214: Is that check necessary? I thought thumbnails are not included in this "groups" anyway. I fear it's going to exclude pretty much any array with shape == 2, while it should be allowed to store FM images as greyscale YX.

pieleric (changes requested):

patrickcleeve2:
Successfully acquire multi-channel z-stack in meteor sim, and loaded in imagej

src/odemis/dataio/tiff.py
```diff
@@ -1854,83 +1862,110 @@ def _saveAsMultiTiffLT(filename, ldata, thumbnail, compressed=True, multiple_fil
...
-                    f.SetField(key, val)
-                except Exception:
-                    logging.exception("Failed to store tag %s with value '%s'", key, val)
-            if data[i].dtype in [numpy.int64, numpy.uint64]:
-                c = None # libtiff doesn't support compression on these types
-            else:
-                c = compression
-            write_image(f, data[i], write_rgb=write_rgb, compression=c, pyramid=pyramid)
+            write_rgb = False
+            hdim = das[0].shape[:-2]  # 1TZ or TZ or Z
```
patrickcleeve2, line 1931: Do the tests cover all these possible dimensions?

pieleric:
> Successfully acquire multi-channel z-stack in meteor sim, and loaded in imagej. We should also test this on a real system before merge
IMHO, I don't think we need to test on a real METEOR to check it works. The only difference between a real METEOR and a simulator is the content of the images, but that doesn't affect how it's shown in ImageJ. Moreover, the current state on the master branch is that it doesn't work, the channels are mixed. So we cannot really make it worse!

patrickcleeve2:
> > Successfully acquire multi-channel z-stack in meteor sim, and loaded in imagej. We should also test this on a real system before merge
> 
> IMHO, I don't think we need to test on a real METEOR to check it works. The only difference between a real METEOR and a simulator is the content of the images, but that doesn't affect how it's shown in ImageJ. Moreover, the current state on the master branch is that it doesn't work, the channels are mixed. So we cannot really make it worse!

I think since this touches how we save images, it might pay to be a little paranoid as it might have unintended sideeffects.

src/odemis/dataio/tiff.py
K4rishma: I have reverted it to the original code and only updated the comment.

src/odemis/dataio/tiff.py
K4rishma: I will add the tests

pieleric (approved):

PR #2721, opened 2024-03-22
[fix] Error moving between grids while in FM imaging mode

patrickcleeve2:
- Fixes an issues where users couldn't swap between grids while in FM imaging mode. (https://support.delmic.com/scp/tickets.php?id=1453#reply)

### DIFF
--- a/src/odemis/acq/move.py
+++ b/src/odemis/acq/move.py
@@ -354,6 +354,9 @@ class MeteorTFS1PostureManager(MeteorPostureManager):
         current_position = self.getCurrentPostureLabel()
         end_pos = None
 
+        # Note: all grid positions need to have rx, rz axes to be able to transform
+        # this is not the case by default, and needs to be added in the metadata
+
         if target_pos_lbl == LOADING:
             end_pos = stage_md[model.MD_FAV_POS_DEACTIVE]
         elif current_position in [LOADING, SEM_IMAGING]:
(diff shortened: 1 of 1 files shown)


src/odemis/acq/move.py
```diff
@@ -370,16 +373,19 @@ def getTargetPosition(self, target_pos_lbl: int) -> Dict[str, float]:
...
                     sem_grid1_pos = stage_md[model.MD_SAMPLE_CENTERS][POSITION_NAMES[GRID_1]]
+                    sem_grid1_pos.update(stage_md[model.MD_FAV_SEM_POS_ACTIVE])
                     fm_target_pos = self._transformFromSEMToMeteor(sem_grid1_pos)
                 elif current_position == SEM_IMAGING:
                     fm_target_pos = self._transformFromSEMToMeteor(self.stage.position.value)
                 end_pos = fm_target_pos
         elif current_position == FM_IMAGING:
             if target_pos_lbl == GRID_1:
                 sem_grid1_pos = stage_md[model.MD_SAMPLE_CENTERS][POSITION_NAMES[GRID_1]]
+                sem_grid1_pos.update(stage_md[model.MD_FAV_SEM_POS_ACTIVE])
```
K4rishma, line 384: Are these changes only required for TFS and not for Tescan and Zeiss?

patrickcleeve2: Yes, the check is only supported in TFS (for now)

K4rishma: This support is already there for Tescan and Zeiss, my bad. The order followed there is updating the angle and then the grid centre coordinates, I think both give the same results. But can we make it consistent for all i.e. TFS, Tescan and zeiss regarding the above update?

K4rishma (approved):

tepals (approved):

src/odemis/acq/move.py
patrickcleeve2: Yep, you are correct its a different order for each manufacturer. Will update

PR #2724, opened 2024-03-28
[fix][METEOR] resolve C/C++ wrapper runtime error when selecting custom colormap

sneepstefan:
When switching to a different tint colormap after having acquired an overview a C/C++ runtime error is generated referring to wrapped object:

File "/usr/lib/python3/dist-packages/odemis/gui/comp/stream_panel.py", line 422, in _on_colormap_value
    self.combo_colormap.SetClientData(self._colormap_customtint_idx, colour)
RuntimeError: wrapped C/C++ object of type ColorMapComboBox has been deleted

This runtime error does not seem to affect the colormap selection or anything GUI related but does however show an Error in the Odemis-gui.log. This error is now being addressed in a way that if the stream panel would be destroyed the tint is unsubscribed.

### DIFF
--- a/src/odemis/gui/comp/stream_panel.py
+++ b/src/odemis/gui/comp/stream_panel.py
@@ -40,7 +40,7 @@ from odemis.gui.comp.slider import UnitFloatSlider, VisualRangeSlider, UnitInteg
 from odemis.gui.comp.stream_bar import StreamBar
 from odemis.gui.comp.text import SuggestTextCtrl, UnitFloatCtrl, FloatTextCtrl, UnitIntegerCtrl
 from odemis.gui.evt import StreamRemoveEvent, StreamVisibleEvent, StreamPeakEvent
-from odemis.gui.util import call_in_wx_main
+from odemis.gui.util import call_in_wx_main, ignore_dead
 from odemis.gui.util.widgets import VigilantAttributeConnector
 from odemis.model import TINT_FIT_TO_RGB, TINT_RGB_AS_IS
 import wx
@@ -406,6 +406,7 @@ class StreamPanelHeader(wx.Control):
             self.label_change_callback(self.ctrl_label.GetValue())
 
     @call_in_wx_main
+    @ignore_dead
     def _on_colormap_value(self, colour):
         """ Update the colormap selector to reflect the provided colour """
         # determine which value to select
@@ -807,6 +808,12 @@ class StreamPanel(wx.Panel):
         self.gb_sizer.Add(lbl_ctrl, (self.num_rows, 0),
                           flag=wx.ALL | wx.ALIGN_CENTER_VERTICAL, border=5)
         return lbl_ctrl
+
+    def Destroy(self):
+        super().Destroy()
+
+        if self.stream.tint:
+            self.stream.tint.unsubscribe(self._header._on_colormap_value)
 
     @control_bookkeeper
     def add_autobc_ctrls(self):


src/odemis/gui/comp/stream_panel.py
```diff
@@ -419,7 +419,12 @@ def _on_colormap_value(self, colour):
         else:  # Can't find the colour => it's custom tint
             if isinstance(colour, tuple):
                 self.colormap_choices[TINT_CUSTOM_TEXT] = colour
-                self.combo_colormap.SetClientData(self._colormap_customtint_idx, colour)
+                # sometimes switching to a custom color will throw a C/C++ wrapper runtime error
+                # this can be ignored and won't affect the colormap selected
+                try:
+                    self.combo_colormap.SetClientData(self._colormap_customtint_idx, colour)
+                except RuntimeError:
```
K4rishma, line 426: I discussed this error with @pieleric and checked the log https://support.delmic.com/scp/tickets.php?id=1453#reply 
From this log, there are dead objects around this error, it seems there is something deleted in the gui, like a stream panel etc but not taken into account correctly, leading to this runtime error.

sneepstefan: Indeed, but there is not stream panel deleted, only the color picker box is deleted when clicking ok, I guess this does not get handle properly. Indeed I have also seen the dead objects. Do you think we can address this issue/error in a better way?

K4rishma: Now we do not understand what behavior in the GUI(that is the clicks) throw this error. If we can point pot when this error is thrown out, we could catch it better?

sneepstefan: I can change the pass line into a logging.warning to log it either way and we could spend more time on it later on? I can in this case create a Jira task on the SWM board for it. @tepals do you think this is a valid temporary solution?

pieleric: I think it'd be better to find out how to reproduce this issue, and then properly fix it by unsubscribing to the tint.

pieleric, line 426: I think that you can keep it generic by just decorating the function with an (extra) `@ignore_dead` (between the call_in_wx_main and the function name).

pieleric: Please note that for now it's not pretty but it *only* causes an error message in the log. I'm pretty certain it doesn't cause any actual issue in the behaviour of the GUI, so not super urgent.

pieleric (changes requested):

patrickcleeve2:
@sneepstefan I added this to jira here: https://delmic.atlassian.net/browse/METEOR-995

sneepstefan:
I addressed this issue, reverted back to the old code and added a Destroy override method to the stream panel class. When its destroyed also unsubscribe from the tint update.
The error is no longer thrown. However there are still dead object references in the log and I am not sure where they are related to.

src/odemis/gui/comp/stream_panel.py
```diff
@@ -808,6 +808,10 @@ def _add_side_label(self, label_text, tooltip=None):
                           flag=wx.ALL | wx.ALIGN_CENTER_VERTICAL, border=5)
         return lbl_ctrl
 
+    def Destroy(self):
```
K4rishma, line 811: I am not sure it works, is it called somewhere?

pieleric: To clarify, `.Destroy()` is part of the standard API in wx objects. So it's automatically called by wxPython when the component is destroyed. This method overrides the standard version. 
@sneepstefan , I think to do it "properly" you should call `super().Destroy()` from here.

pieleric (changes requested):

pieleric (approved):

K4rishma (approved):

PR #2726, opened 2024-04-03
[SDEV-1827] undo redo shape drawing action

nandishjpatel:
NOTE: https://github.com/delmic/odemis/pull/2701 must be reviewed and merged first

### DIFF
--- a/src/odemis/gui/comp/overlay/ellipse.py
+++ b/src/odemis/gui/comp/overlay/ellipse.py
@@ -19,9 +19,8 @@ This file is part of Odemis.
     Odemis. If not, see http://www.gnu.org/licenses/.
 
 """
-from enum import IntEnum
 import math
-from typing import Dict, Any
+from typing import Optional
 
 import cairo
 
@@ -40,12 +39,14 @@ from odemis.gui.comp.overlay.rectangle import RectangleOverlay
 NUM_ARCS_FACTOR = 3
 
 
-class EllipseStateKeys(IntEnum):
-    P_POINT1 = 1
-    P_POINT2 = 2
-    P_POINT3 = 3
-    P_POINT4 = 4
-    POINTS_PRIVATE = 5
+class EllipseState:
+    def __init__(self, ellipse_overlay) -> None:
+        self.p_point1 = ellipse_overlay.p_point1
+        self.p_point2 = ellipse_overlay.p_point2
+        self.p_point3 = ellipse_overlay.p_point3
+        self.p_point4 = ellipse_overlay.p_point4
+        self._points = ellipse_overlay._points.copy()
+
 
 class EllipseOverlay(RectangleOverlay):
     """Overlay representing one ellipse."""
@@ -79,24 +80,21 @@ class EllipseOverlay(RectangleOverlay):
         self._phys_to_view()
         self.points.value = self._points
 
-    def get_state(self):
+    def get_state(self) -> Optional[EllipseState]:
         """Get the current state of the shape."""
-        state = {
-            EllipseStateKeys.P_POINT1: self.p_point1,
-            EllipseStateKeys.P_POINT2: self.p_point2,
-            EllipseStateKeys.P_POINT3: self.p_point3,
-            EllipseStateKeys.P_POINT4: self.p_point4,
-            EllipseStateKeys.POINTS_PRIVATE: self._points.copy(),
-        }
-        return state
+        # Only return the state if the ellipse creation is finished
+        # By doing so avoid storing an undo action during ellipse creation
+        if self.p_point1 != self.p_point3:
+            return EllipseState(self)
+        return None
 
-    def restore_state(self, state: Dict[EllipseStateKeys, Any]):
+    def restore_state(self, state: EllipseState):
         """Restore the shape to a given state."""
-        self.p_point1 = state[EllipseStateKeys.P_POINT1]
-        self.p_point2 = state[EllipseStateKeys.P_POINT2]
-        self.p_point3 = state[EllipseStateKeys.P_POINT3]
-        self.p_point4 = state[EllipseStateKeys.P_POINT4]
-        self._points = state[EllipseStateKeys.POINTS_PRIVATE]
+        self.p_point1 = state.p_point1
+        self.p_point2 = state.p_point2
+        self.p_point3 = state.p_point3
+        self.p_point4 = state.p_point4
+        self._points = state._points
         self._phys_to_view()
         self.points.value = self._points
 
--- a/src/odemis/gui/comp/overlay/ellipse.py
+++ b/src/odemis/gui/comp/overlay/ellipse.py
@@ -19,7 +19,9 @@ This file is part of Odemis.
     Odemis. If not, see http://www.gnu.org/licenses/.
 
 """
+from enum import IntEnum
 import math
+from typing import Dict, Any
 
 import cairo
 
@@ -38,6 +40,13 @@ from odemis.gui.comp.overlay.rectangle import RectangleOverlay
 NUM_ARCS_FACTOR = 3
 
 
+class EllipseStateKeys(IntEnum):
+    P_POINT1 = 1
+    P_POINT2 = 2
+    P_POINT3 = 3
+    P_POINT4 = 4
+    POINTS_PRIVATE = 5
+
 class EllipseOverlay(RectangleOverlay):
     """Overlay representing one ellipse."""
 
@@ -55,13 +64,7 @@ class EllipseOverlay(RectangleOverlay):
         """
         shape = EllipseOverlay(self.cnvs)
         shape.colour = self.colour
-        shape.p_point1 = self.p_point1
-        shape.p_point2 = self.p_point2
-        shape.p_point3 = self.p_point3
-        shape.p_point4 = self.p_point4
-        shape._points = self._points.copy()
-        shape._phys_to_view()
-        shape.points.value = shape._points
+        shape.restore_state(self.get_state())
         return shape
 
     def move_to(self, pos):
@@ -73,6 +76,27 @@ class EllipseOverlay(RectangleOverlay):
         self.p_point3 += shift
         self.p_point4 += shift
         self._points = [p + shift for p in self._points]
+        self._phys_to_view()
+        self.points.value = self._points
+
+    def get_state(self):
+        """Get the current state of the shape."""
+        state = {
+            EllipseStateKeys.P_POINT1: self.p_point1,
+            EllipseStateKeys.P_POINT2: self.p_point2,
+            EllipseStateKeys.P_POINT3: self.p_point3,
+            EllipseStateKeys.P_POINT4: self.p_point4,
+            EllipseStateKeys.POINTS_PRIVATE: self._points.copy(),
+        }
+        return state
+
+    def restore_state(self, state: Dict[EllipseStateKeys, Any]):
+        """Restore the shape to a given state."""
+        self.p_point1 = state[EllipseStateKeys.P_POINT1]
+        self.p_point2 = state[EllipseStateKeys.P_POINT2]
+        self.p_point3 = state[EllipseStateKeys.P_POINT3]
+        self.p_point4 = state[EllipseStateKeys.P_POINT4]
+        self._points = state[EllipseStateKeys.POINTS_PRIVATE]
         self._phys_to_view()
         self.points.value = self._points
 
--- a/src/odemis/gui/comp/overlay/polygon.py
+++ b/src/odemis/gui/comp/overlay/polygon.py
@@ -19,8 +19,10 @@ This file is part of Odemis.
     Odemis. If not, see http://www.gnu.org/licenses/.
 
 """
+from enum import IntEnum
 import logging
 import math
+from typing import Dict, Any
 
 import cairo
 import wx
@@ -31,6 +33,12 @@ from odemis.gui.comp.overlay.base import SEL_MODE_ROTATION, LineEditingMixin, Ve
 from odemis.gui.comp.overlay.shapes import EditableShape
 import odemis.util.units as units
 from odemis.util.raster import point_in_polygon
+
+
+class PolygonStateKeys(IntEnum):
+    V_POINTS = 1
+    FINISHED_PRIVATE = 2
+    POINTS_PRIVATE = 3
 
 
 class PolygonOverlay(WorldOverlay, LineEditingMixin, EditableShape):
@@ -56,11 +64,7 @@ class PolygonOverlay(WorldOverlay, LineEditingMixin, EditableShape):
         """
         shape = PolygonOverlay(self.cnvs)
         shape.colour = self.colour
-        shape.v_points = self.v_points.copy()
-        shape._finished = self._finished
-        shape._points = self._points.copy()
-        shape._phys_to_view()
-        shape.points.value = shape._points
+        shape.restore_state(self.get_state())
         return shape
 
     def move_to(self, pos):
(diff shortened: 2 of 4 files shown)


src/odemis/gui/comp/overlay/shapes.py
```diff
@@ -232,6 +273,12 @@ def on_char(self, evt):
                 self._shape_to_copy = self._selected_shape
                 self.cnvs.set_default_cursor(wx.CURSOR_BULLSEYE)
                 self.cnvs.request_drawing_update()
+            elif evt.GetKeyCode() == wx.WXK_CONTROL_Z:
+                self._selected_shape.undo()
+                self._selected_shape.refresh()
+            elif evt.GetKeyCode() == wx.WXK_CONTROL_Y:
+                self._selected_shape.redo()
+                self._selected_shape.refresh()
```
pieleric, line 281: That mean the undo/redo stack is per *shape*. Isn't that really odd for the user? Typically programs provide a single undo stack. Did you check with the users that it's what they want? I guess that once you're aware of it, it fulfil its need, but the fact that as soon as you select another shape you cannot undo the changes you've made by pressing Ctrl+Z is confusing.

src/odemis/gui/comp/overlay/shapes.py
```diff
@@ -45,6 +47,11 @@ def __init__(self, cnvs):
         # Useful for internal points manipulation
         self._points: List[Vec] = []
         self.cnvs = cnvs
+        # History of states
+        self.undo_stack = deque()
+        self.redo_stack = deque()
+        self.undo_limit = 25
+        self.redo_limit = 25
```
pieleric, line 54: Please put a constant at the top of the file `UNDO_STACK_DEPTH = 25`

src/odemis/gui/comp/overlay/ellipse.py
```diff
@@ -74,10 +84,36 @@ def move_to(self, pos):
...
         self.points.value = self._points
 
+    def save_state(self):
+        """Save the current state of the shape."""
+        state = {
+            EllipseStateKeys.P_POINT1: self.p_point1,
+            EllipseStateKeys.P_POINT2: self.p_point2,
+            EllipseStateKeys.P_POINT3: self.p_point3,
+            EllipseStateKeys.P_POINT4: self.p_point4,
+            EllipseStateKeys.POINTS_PRIVATE: self._points.copy()
```
pieleric, line 98: `._points` is computed based on the 4 p_points, so I'd argue it doesn't need to be stored as a state.

src/odemis/gui/comp/overlay/shapes.py
```diff
@@ -92,7 +99,41 @@ def move_to(self, pos: Union[Tuple[float, float], Vec]):
...
+            self.cnvs.request_drawing_update()
+        else:
+            logging.warning(
+                "No redo action for %s at %s.", self.__class__.__name__, hex(id(self))
+            )
+
+    @abstractmethod
+    def save_state(self):
+        """Save the current state of the shape."""
+        pass
```
pieleric, line 132: I think you should be able to simplify a little bit the code by having a simpler `get_state` function that just returns the state of the shape. Then `save_state()` could be generic here, calling `get_state()` and appending it to the the redo stack (with a depth limit).

It would also allow to simplify the .copy(s) function to just `self.restore_state(s.get_state())`.

src/odemis/gui/comp/overlay/shapes.py
```diff
@@ -232,6 +273,12 @@ def on_char(self, evt):
                 self._shape_to_copy = self._selected_shape
                 self.cnvs.set_default_cursor(wx.CURSOR_BULLSEYE)
                 self.cnvs.request_drawing_update()
+            elif evt.GetKeyCode() == wx.WXK_CONTROL_Z:
+                self._selected_shape.undo()
+                self._selected_shape.refresh()
```
pieleric, line 278: I'd argue the `refresh()` should be done automatically by the `.restore_state()`.

src/odemis/gui/comp/overlay/shapes.py
```diff
@@ -92,7 +99,41 @@ def move_to(self, pos: Union[Tuple[float, float], Vec]):
...
+
+    def redo(self):
+        """Redo the last undone action."""
+        if self.redo_stack:
+            state = self.redo_stack.pop()
+            self.undo_stack.append(state)
+            self.restore_state(state)
+            self.cnvs.request_drawing_update()
+        else:
+            logging.warning(
```
pieleric, line 125: "warning" means that the error log panel will turn orange. Just for Ctrl+Z or Ctrl+Y, that sounds a little over the top. I'd argue a "info" level is enough.

src/odemis/gui/comp/overlay/shapes.py
```diff
@@ -92,7 +99,41 @@ def move_to(self, pos: Union[Tuple[float, float], Vec]):
...
-        """Refresh the shape's points VA after copy or move_to operation."""
+        """Refresh the shape's points VA after copy, move_to or restore_state operation."""
+        pass
+
+    def undo(self):
+        """Undo the last action."""
+        if self.undo_stack:
+            state = self.undo_stack.pop()
+            self.redo_stack.append(state)
+            self.restore_state(state)
```
pieleric, line 110: Does this work properly? Isn't there some sort of "off by 1" error?
The current state is the "topest" state on the undo_stack, right? Then the first time you call it, nothing will be undone. 
I think you should restore the state to `self.undo_stack[-1]`

src/odemis/gui/comp/overlay/shapes.py
```diff
@@ -45,6 +47,11 @@ def __init__(self, cnvs):
         # Useful for internal points manipulation
         self._points: List[Vec] = []
         self.cnvs = cnvs
+        # History of states
+        self.undo_stack = deque()
+        self.redo_stack = deque()
```
pieleric, line 52: If you use deque, then you should `max_len` to automatically discard the oldest elements. Otherwise, using a list is as simple.

src/odemis/gui/comp/overlay/shapes.py
```diff
@@ -92,7 +99,41 @@ def move_to(self, pos: Union[Tuple[float, float], Vec]):
...
+            self.redo_stack.append(state)
+            self.restore_state(state)
+            self.cnvs.request_drawing_update()
+        else:
+            logging.warning(
+                "No undo action for %s at %s.", self.__class__.__name__, hex(id(self))
+            )
+
+    def redo(self):
+        """Redo the last undone action."""
```
pieleric, line 118: Shouldn't the redo stack be automatically emptied when `save_state` is called? Admittedly, I think that it works fine as it is, but my impression is that most software have this behaviour, so it would avoid "surprises" for the user.

pieleric (changes requested):

src/odemis/gui/comp/overlay/shapes.py
nandishjpatel: Right, I will change the implementation to `Typically programs provide a single undo stack.`

src/odemis/gui/comp/overlay/shapes.py
nandishjpatel: Good catch. I had this implemented and for some reason its not here in the PR.

src/odemis/gui/comp/overlay/shapes.py
```diff
@@ -20,13 +20,18 @@
...
+from enum import IntEnum
+import logging
+from typing import Tuple, List, Union, Dict, Any
 
 import wx
 
 from odemis import model, util
 from odemis.gui.comp.overlay.base import WorldOverlay, Vec
 
+UNDO_STACK_DEPTH = 25
```
tepals, line 33: can you add a short comment to explain what this value is? Something like: `the number of undo actions stored in the stack`

src/odemis/gui/comp/overlay/shapes.py
```diff
@@ -211,7 +232,13 @@ def on_char(self, evt):
         if not self.active.value:
             return super().on_char(evt)
 
-        if self._selected_shape:
+        if evt.GetKeyCode() == wx.WXK_CONTROL_Z:
+            self._undo_action = True
+            self.undo()
+        elif evt.GetKeyCode() == wx.WXK_CONTROL_Y:
```
tepals, line 238: A lot of software (at least the I use) supports both `ctrl+y` and `ctrl+shift+z`, I find `ctrl+shift+z` easier to use, would it be easy to also support that here?

nandishjpatel: Its not that easy
https://docs.wxpython.org/wx.KeyCode.enumeration.html#wx-keycode

tepals:
I tested the PR locally and I noticed that the creation of a shape is not undone when pressing `ctrl+z`. However the action is added to the stack. If I move a shape and then create two new shapes, I need to press `ctrl+z` three times before the movement action is undone. I would expect that either the movement action is undone with one press of `ctrl+z`, or that the creation of shapes is also undone. 


https://github.com/delmic/odemis/assets/46314774/e4ab21c6-ca28-462b-8f1b-34f3c1c47625

nandishjpatel:
> I tested the PR locally and I noticed that the creation of a shape is not undone when pressing `ctrl+z`. However the action is added to the stack. If I move a shape and then create two new shapes, I need to press `ctrl+z` three times before the movement action is undone. I would expect that either the movement action is undone with one press of `ctrl+z`, or that the creation of shapes is also undone.
> undo-redo.mp4

The creation of a shape and copying of a shape is also handled by the undo and redo action now.

tepals:
Should we also handle deleting shapes? If I delete a shape, nothing happens when I press `ctrl+z`

https://github.com/delmic/odemis/assets/46314774/d7a7c1d8-4076-495e-bd94-57c077054761

nandishjpatel:
> Should we also handle deleting shapes? If I delete a shape, nothing happens when I press `ctrl+z`
> Screencast.2024-05-22.17.41.13.mp4

Yeah we should, I forgot about it. By the way I could not play any of your uploaded videos, it says the file is corrupt.

nandishjpatel:
@tepals a suggestion, you should actually test the undo and redo in the software lag fix PR https://github.com/delmic/odemis/pull/2752. The algorithm works better compared this PR because there we let ShapesOverlay handle all mouse events. Just give it a try.

nandishjpatel:
The shape deletion is also handled by the undo and redo action now

src/odemis/gui/comp/overlay/shapes.py
```diff
@@ -237,9 +288,73 @@ def on_left_up(self, evt):
...
+            logging.info(
+                "No undo action for %s at %s.", self.__class__.__name__, hex(id(self))
+            )
+
+    def redo(self):
+        """Redo the last undone action."""
+        if self._redo_stack:
+            # If the undo stack is empty or if the previous action was undo
+            # correct the off by 1 shape's state by popping it and not restoring the state
+            if not self._undo_stack or self._undo_action:
```
pieleric, line 334: I don't think the `not undo_stack` is useful here. That's because undo/redo are not symmetrical: editing a shape reset just the redo_stack. So the only way to have an empty undo_stage is that the last action was an undo_action. Hence, it's not useful here.

src/odemis/gui/comp/overlay/shapes.py
```diff
@@ -237,9 +288,73 @@ def on_left_up(self, evt):
...
+                shape_state = ShapeState(self._selected_shape, state, self._is_new_shape, False)
+                if not self._undo_stack or self._undo_stack[-1] != shape_state:
+                    self._undo_stack.append(shape_state)
+                    self._redo_stack.clear()  # Clear redo stack when a shape's state is saved
         else:
             WorldOverlay.on_left_up(self, evt)
 
+    def undo(self):
+        """Undo the last action."""
+        if self._undo_stack:
```
pieleric, line 302: You could make the code  a tiny simpler to read by adjusting this `if` to
```
if not self._undo_stack:
     logging.warning(...)
     return
```
Then, the rest of the code is one indentation less, and become the "normal code".

src/odemis/gui/comp/overlay/polygon.py
```diff
@@ -71,6 +75,25 @@ def move_to(self, pos):
...
+            PolygonStateKeys.FINISHED_PRIVATE: self._finished,
+            PolygonStateKeys.POINTS_PRIVATE: self._points.copy(),
+        }
+        # Only return the state if the polygon creation is finished
+        # By doing so avoid storing an undo action during polygon creation
+        return state if self.right_click_finished else {}
+
+    def restore_state(self, state: Dict[PolygonStateKeys, Any]):
+        """Restore the shape to a given state."""
+        self.v_points = state[PolygonStateKeys.V_POINTS]
```
pieleric, line 91: `v_points` are just the `_points` in the current view coordinate. They are going to be recomputed based on _points in the call to `_phys_to_view()`. So what's the point to store them? I'd suggest to not store them, and just reset v_points to `[None] * len(_points)` to make `_phys_to_view()` happy.

src/odemis/gui/comp/overlay/ellipse.py
```diff
@@ -76,6 +79,29 @@ def move_to(self, pos):
...
+        # By doing so avoid storing an undo action during ellipse creation
+        return state if self.p_point1 != self.p_point3 else {}
+
+    def restore_state(self, state: Dict[EllipseStateKeys, Any]):
+        """Restore the shape to a given state."""
+        self.p_point1 = state[EllipseStateKeys.P_POINT1]
+        self.p_point2 = state[EllipseStateKeys.P_POINT2]
+        self.p_point3 = state[EllipseStateKeys.P_POINT3]
+        self.p_point4 = state[EllipseStateKeys.P_POINT4]
+        self._points = state[EllipseStateKeys.POINTS_PRIVATE]
```
pieleric, line 101: The first thing that happens in `draw()` is `self._points.clear()`. They are dependent only on the 4 p_points. So what's the point of storing them? Is it because storing them is easier than recomputing them? If so, please add a comment explaining this. Otherwise, don't store them, and force the recomputation.

pieleric (changes requested):

src/odemis/gui/comp/overlay/ellipse.py
nandishjpatel: I store them, just to set the `points` VA in line 103. Settings the points VA is important because it does the megafield computation in the subscribers callback. If I don't store them, I have to call `update_drawing()` first and then I can set the `points` VA. Doing this will call `update_drawing` twice, once here and then in ShapesOverlay's `undo` or `redo`

src/odemis/gui/comp/overlay/rectangle.py
```diff
@@ -66,6 +70,28 @@ def move_to(self, pos):
...
+        # Only return the state if the rectangle creation is finished
+        # By doing so avoid storing an undo action during rectangle creation
+        return state if self.p_point1 != self.p_point3 else {}
+
+    def restore_state(self, state: Dict[RectangleStateKeys, Any]):
+        """Restore the shape to a given state."""
+        self.p_point1 = state[RectangleStateKeys.P_POINT1]
+        self.p_point2 = state[RectangleStateKeys.P_POINT2]
+        self.p_point3 = state[RectangleStateKeys.P_POINT3]
+        self.p_point4 = state[RectangleStateKeys.P_POINT4]
```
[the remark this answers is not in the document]
nandishjpatel: Right, I changed it

pieleric (changes requested):

tepals (approved):

PR #2729, opened 2024-04-05
[MPM-45] SONA validate off-centre clipping with Andorcam3

sneepstefan:
This is only needed for the SONA camera due to mechanical differences.

When max_res is less than the full resolution of the sensor, still allow to set a .translation VA != 0,0 , so that the image can be shifted.
We should then validate that the METEOR GUI doesn’t reset/change .translation when binning  (or other settings) is changed. If it does, either the GUI should be adjusted, or we should find a different way to apply translation.

-> validation was done on a real SONA and did not need modifications to the code.
-> changes are only to the simcam to better simulate the real camera.

### DIFF
--- a/src/odemis/driver/simcam.py
+++ b/src/odemis/driver/simcam.py
@@ -109,7 +109,8 @@ class Camera(model.DigitalCamera):
         self.binning = model.ResolutionVA(self._binning,
                               ((1, 1), (16, 16)), setter=self._setBinning)
 
-        hlf_shape = (self._shape[0] // 2 - 1, self._shape[1] // 2 - 1)
+        # the Numpy dimension order of the DA is YX for greyscale and YXC for RGB images
+        hlf_shape = (self._img.shape[1] // 2 - 1, self._img.shape[0] // 2 - 1)
         tran_rng = [(-hlf_shape[0], -hlf_shape[1]),
                     (hlf_shape[0], hlf_shape[1])]
         self._translation = (0, 0)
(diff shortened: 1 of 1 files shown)


src/odemis/driver/simcam.py
```diff
@@ -109,7 +109,7 @@ def clip_max_res(img_res):
         self.binning = model.ResolutionVA(self._binning,
                               ((1, 1), (16, 16)), setter=self._setBinning)
 
-        hlf_shape = (self._shape[0] // 2 - 1, self._shape[1] // 2 - 1)
+        hlf_shape = (self._img.shape[-1] // 2 - 1, self._img.shape[-2] // 2 - 1)
```
patrickcleeve2, line 112: I think this works differently than the previous code for 2D images? I'm not sure if these images are ever 2d, but comment above indicates that grayscale is supported?

```suggestion
     dim0, dim1 = 0, 1
     if self._img.ndim == 3:
         dim0, dim1 = -1, -2
     hlf_shape = (self._img.shape[dim0] // 2 - 1, self._img.shape[dim1] // 2 - 1)
```

pieleric: Typically the images are greyscale (last two dimensions are YX), but it has to also handle RGB images, of dimensions YXC, so the change indeed doesn't work. However the proposed change by Patrick also doesn't work, as it would mix up the dimensions, in both cases.
I think just writing img.shape[1] ..... img.shape[0] would work. You'd want to add a comment to indicate that the numpy dimension order is YX(C).

FCanberk (approved):

pieleric (changes requested):
The code is fine, but the commit title is confusing. It changes the simcam driver, but the title only mentions andorcam3. Also, the "Sona" mention is there because the use-case is with the Andor Sona, but that's really not important for the simulator.

I'd suggest a title like "[feature] allow simcam off-centre translation when using max_res". In the description, you can then explain that this is to simulate better the andorcam3, as we'll need to do that with the Sona.

sneepstefan:
commit message changed as you suggested @pieleric

pieleric (approved):

PR #2730, opened 2024-04-08
[fix][METEOR-590] TFS Posture Manager which reads FM features at a fixed imaging plane

K4rishma:
Read raw z coordinate system in TFS systems and image on fixed plane when performing FM imaging.
- add new posture manager
- add new test functions in move_test.py

https://bitbucket.org/delmic/xtlib-adapter/pull-requests/83

Switch from reading to linked Z coordinated to stage-bare Z coordinate for TFS  xt_client.py driver.

Tested:
- The stage bare Z is always be read with these commit changes irrespective of how TFS software reads Z. The TFS software can read Z as linked Z or stage bare Z depending on the user's workflow.
- SEM to FM switching works
- FM to SEM does not work as shift in y is observed on hardware. Now it is corrected and tested on simulation, needs to be tested on hardware

### DIFF
--- a/install/linux/usr/share/odemis/sim/meteor-tfs2-sim.odm.yaml
+++ b/install/linux/usr/share/odemis/sim/meteor-tfs2-sim.odm.yaml
@@ -1,0 +1,224 @@
+# For now, this is esssentially just a FM microscope, as the FIB part is handled
+# separately by the SEM
+
+"METEOR TFSv2": {
+    class: Microscope,
+    role: meteor,
+}
+
+"SEM": {
+    class: simsem.SimSEM,
+    role: null,
+    init: {
+           image: "simsem-fake-output.h5", # any large 16 bit image is fine
+    },
+    children: {
+        scanner: "EBeam",
+        detector0: "SE Detector",
+        focus: "EBeam Focus",
+    }
+}
+
+"EBeam": {
+    # Internal child of SimSEM, so no class
+    role: e-beam,  # Not required for the meteor
+    init: {},
+    affects: ["SE Detector"],
+}
+
+"EBeam Focus": {
+    # role: ebeam-focus,
+    role: null,
+    init: {},
+    affects: ["EBeam"]
+}
+
+"SE Detector": {
+    # Internal child of SimSEM, so no class
+    # role: se-detector,
+    role: null,
+    init: {},
+}
+
+
+# Normally provided by the SEM
+"Stage": {
+    class: tmcm.TMCLController,
+    role: stage-bare,
+    init: {
+        port: "/dev/fake6",
+        address: null,
+        axes: ["x", "y", "z", "rx", "rz"],
+        ustepsize: [1.e-7, 1.e-7, 1.e-7, 1.2e-5, 1.2e-5], # unit/µstep
+        rng: [[-0.1, 0.1], [-0.05, 0.05], [-0.05, 0.1], [-2, 2], [0, 6.28]],
+        unit: ["m", "m", "m", "rad", "rad"],
+        refproc: "Standard",
+    },
+    metadata: {
+        # Loading position:
+        FAV_POS_DEACTIVE: { 'rx': 0, 'rz': 1.9076449, 'x': -0.01529, 'y': 0.0506, 'z': 29.e-3 },
+        # XYZ ranges for SEM & METEOR
+        SEM_IMAGING_RANGE: {"x": [-10.e-3, 10.e-3], "y": [-30.e-3, 25.e-3], "z": [15.e-3, 37.e-3]},
+        FM_IMAGING_RANGE: {"x": [0.040, 0.054], "y": [-30.e-3, 25.e-3], "z": [15.e-3, 37.e-3]},
+        # Grid centers in SEM range
+        # Adjusted so that at init (0,0,0), it's at the Grid 1.
+        SAMPLE_CENTERS: {"GRID 1": {'x': 0, 'y': 0, 'z': 32.e-3}, "GRID 2": {'x': 2.98e-3, 'y': 2.46e-3, 'z': 32.e-3}},
+        # dx and dy are mirroring values between SEM - METEOR, pre-tilt is 35°
+        CALIB: {"dx": 0.0506252, "dy": 0.0049832, "version": "tfs_2", "pre-tilt": 0.6108652381980153},
+        # Active tilting (rx) & rotation(rz) angles positions when switching between SEM & FM.
+        # Note: these values are calibrated at installation time.
+        FAV_FM_POS_ACTIVE: {"rx": 0.12213888553625313  , "rz":  1.91986}, # 7° - 270°
+        # Typically rz = 110°, but we make it 0 so that at init it looks like in SEM position
+        FAV_SEM_POS_ACTIVE: {"rx": 0.314159, "rz": 5.06145},    # Note that milling angle (rx) can be changed per session
+    },
+}
+
+"Linked YZ": {
+    class: actuator.ConvertStage,
+    role: null,
+    dependencies: {
+        "under": "Stage"
+    },
+    init: {
+      axes: [ "y", "z" ], # name of the axes in the dependency, mapped to x,y (if identity transformation)
+      rotation:  0.6108652381980153, # rad , 35° same as pre-tilt in Stage component
+    },
+}
+
+"Meteor Stage": {
+    class: actuator.MultiplexActuator,
+    role: stage,
+    dependencies: { "x": "Stage", "y": "Linked YZ", "z": "Linked YZ", },
+    init: {
+        axes_map: { "x": "x", "y": "x", "z": "y",},
+    },
+    affects: ["Camera", "EBeam"],
+    metadata: {
+        # Typically, x range is the same as FM_IMAGING_RANGE, and Y has to be converted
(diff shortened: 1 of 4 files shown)


pieleric:
Note we need to support all the current installations without having to update their microscope files. So xt_client should still read *by default* the NON raw coordinates. Either you'd change it via an argument of the stage (eg `raw=False`), or via a VA (but I don't think it's useful to be able to switch the coordinates "live", especially if the ranges change).

Also, for now the xtlib-adpater in master doesn't support this function.  Please indicate in the log which version of xtlib-adapter is required to support this option (probably v1.11.5? Or more). Here, in this PR, you can also add a link to the xtlib-adapter PR.

install/linux/usr/share/odemis/sim/meteor-tfs2-sim.odm.yaml
```diff
@@ -0,0 +1,238 @@
...
+    children: [
+        "Stage",
+        "Meteor Stage",
+        "Optical Objective",
+        "Optical Focus",
+        "Light Source",
+        "Filter Wheel",
+        "Camera",
+        "EBeam",
+    ],
```
pieleric, line 16: You don't need these anymore. Please drop.

install/linux/usr/share/odemis/sim/meteor-tfs2-sim.odm.yaml
```diff
@@ -0,0 +1,238 @@
...
+}
+
+"SEM": {
+    class: simsem.SimSEM,
```
pieleric, line 20: I think that in this case it'd make more sense to use the xt_client, with the assumption that the xt_adapter simulator is running on the localhost. This way, you'd show the way to activate raw coordinates too.

src/odemis/acq/test/move_test.py
```diff
@@ -617,6 +618,151 @@ def test_moving_in_grid1_fm_imaging_area_after_loading(self):
...
+    """
+    Test the MeteorPostureManager functions for TFS 2
+    """
+    MIC_CONFIG = METEOR_TFS2_CONFIG
+    ROTATION_AXES = {'rx', 'rz'}
+
+    def test_moving_in_grid1_fm_imaging_area_after_loading(self):
+        """Check if the stage moves in the right direction when moving in the fm imaging grid 1 area."""
+        super().test_moving_in_grid1_fm_imaging_area_after_loading()
+        meteor_stage_md = self.linked_stage.getMetadata()
```
patrickcleeve2, line 631: what is self.linked_stage? I cant find a reference to it

src/odemis/driver/xt_client.py
```diff
@@ -2147,7 +2156,10 @@ class Stage(model.Actuator):
     moving the TFS stage and updating the position.
     """
 
-    def __init__(self, name: str, role: str, parent: model.HwComponent, rng: dict = None, **kwargs) -> None:
+    def __init__(self, name: str, role: str, parent: model.HwComponent, rng: dict = None, raw: int = 0,
```
patrickcleeve2, line 2159: change the raw argument to a boolean

src/odemis/acq/move.py
```diff
@@ -614,6 +618,278 @@ def _doCryoSwitchSamplePosition(self, future, target):
...
+    def _transformFromSEMToMeteor(self, pos: Dict[str, float]) -> Dict[str, float]:
+        """
+        Transforms the current stage position from the SEM imaging area to the
+        meteor/FM imaging area.
+        :param pos: (dict str->float) the initial stage position.
+        :return: (dict str->float) the transformed position.
+        """
+
+        stage_md = self.stage.getMetadata()
+        self.sem_stage_pos_pre = pos
```
patrickcleeve2, line 787: setting this here will reset it every time this function is called, and its called frequently to check target position etc for a variety of positions (e.g. grid_1_pos). I don't think thats what we want to do.

src/odemis/acq/test/move_test.py
```diff
@@ -617,6 +618,151 @@ def test_moving_in_grid1_fm_imaging_area_after_loading(self):
...
+
+    def test_moving_from_grid1_to_grid2_in_fm_imaging_Area(self):
+
+        # todo check zfm in hardware
+        #-0.0017951474872928566 , -2.4472740404152816e-05, grid 2 zfm 0.0017706747468887037
+        meteor_stage_md = self.linked_stage.getMetadata()
+        self.linked_stage.updateMetadata({model.MD_FAV_POS_ACTIVE: {'z': 0.0017706747468887037}})
+        fav_pos_active_orig = self.linked_stage.getMetadata().get(model.MD_FAV_POS_ACTIVE, None)
+
+        f = self.posture_manager.cryoSwitchSamplePosition(LOADING)
```
patrickcleeve2, line 658: You should move to the SEM imaging first

K4rishma: Is it not safe to move to FM from Loading position? Is there another reason we should not switch to FM directly from Loading? From this code it looks safe 
image

K4rishma:
Latest code after the hardware testing on 26th June is added, separate PR for xt_client changes.

K4rishma:
@pieleric @patrickcleeve2  I added correction to switch from FM to SEM properly.

K4rishma:
Ready to review

src/odemis/acq/move.py
```diff
@@ -614,6 +618,437 @@ def _doCryoSwitchSamplePosition(self, future, target):
                 future._task_state = FINISHED
 
 
+class ConvertStage:
```
patrickcleeve2, line 621: As discussed, instead of adding the convert stage, move the transformation into the posture manger, like https://github.com/delmic/odemis/pull/2893

src/odemis/acq/move.py
```diff
@@ -614,6 +618,437 @@ def _doCryoSwitchSamplePosition(self, future, target):
...
+                                     end_pos["z"] * math.cos(self.rotation))}
+
+    # Note: this transformation consists of translation of along x and y
+    # axes, and 7 degrees rotation around rx, and 180 degree rotation around rz.
+    # The rotation angles are constant existing in "FM_POS_ACTIVE" metadata,
+    # but the translation are calculated based on the current position and some
+    # correction/shifting parameters existing in metadata "FM_POS_ACTIVE".
+    # This correction parameters can change every session. They are calibrated
+    # at the beginning of each run.
+    def _transformFromSEMToMeteor(self, pos: Dict[str, float]) -> Dict[str, float]:
```
patrickcleeve2, line 828: Instead of having two functions that do very similar thing (except one condition), merge _transformFromSEMToMeteor and calculate_pos_before_fixing_imaging_plane. Just pass an argument when you want to fix the imaging plane

install/linux/usr/share/odemis/sim/meteor-tfs2-sim.odm.yaml
```diff
@@ -0,0 +1,207 @@
...
+METEOR-Sim: {
+    class: Microscope,
+    role: meteor,
+}
+
+"SEM": {
+    class: xt_client.SEM,
+    role: null,
+    init: {
+           address: "PYRO:Microscope@localhost:4242", #"PYRO:Microscope@192.168.6.5:4242"
```
patrickcleeve2, line 13: Does this require running the simulated xt-adapter?

K4rishma: no  not needed, I will remove it and incorporate  your yaml file structure in https://github.com/delmic/odemis/pull/2893/files#diff-18ae95404cd6730746a0ed90397cca0f276ec2e213d950a06406e28a5a4d78e4 Then we can have one yaml file that works with both the PRs

K4rishma: I am not able to use raw_coordinates to true in stage-bare if I do not use xt-adaptor simulator. I think that should be fine.

K4rishma:
ready for review @patrickcleeve2

src/odemis/acq/move.py
```diff
@@ -614,6 +805,237 @@ def _doCryoSwitchSamplePosition(self, future, target):
...
+        if not ("rz" in pos and "rz" in sem_pos_active):
+            raise ValueError(f"The stage position does not have rz axis. pos={pos}, sem_pos_active={sem_pos_active}")
+
+        # whether we need to rotate around the z axis (180deg)
+        has_rz = not isNearPosition(pos, sem_pos_active, {"rz"},
+                                    atol_rotation=ATOL_ROTATION_TRANSFORM)
+
+        # compensate the impact of fixing the fm imaging plane on the current stage position
+        # based on the value of original imaging plane value saved before fixing the fm imaging plane
+        new_pos = pos.copy()
```
patrickcleeve2, line 924: I think this is missing transformed_pos in the case there no stage_pos_before_fav_fm.  And I don't think we can assume it's always there

patrickcleeve2 (changes requested):

patrickcleeve2 (approved):

patrickcleeve2:
Can you update the title of the PR as this now refers to the static imaging plane instead of raw coordiantes

pieleric (approved):

pieleric:
@K4rishma , could you resolve the conflicts of this PR against master? Then it should be ready to be merged!

K4rishma:
Tested on xt_client simulator, resolved merge conflict and squashed commits.

PR #2733, opened 2024-04-15
[fix][METEOR-1004] display how many streams will be exported

patrickcleeve2:
Users were getting confused as to why overlay wasn't exported, because the export is based on the currently selected view. The export button now displays how many streams will be exported based on the view / stream visibility.


FCanberk (approved):

src/odemis/gui/cont/export.py
```diff
@@ -115,6 +115,7 @@ def on_streams_change(self, streams):
         enabled = (len(streams) > 0)
         self._main_frame.menu_item_export_as.Enable(enabled)
         self._tab_panel.btn_export.Enable(enabled)
+        self._tab_panel.btn_export.SetLabel(f"Export Image ({len(streams)})")
```
pieleric, line 118: If there are no streams, it shows a disabled button with label "Export Image (0)". I think it sounds odd. Probably the label should just be "Export Image" in such cases.

pieleric (changes requested):
Now the label of the button in the analysis tab is "Export Image (N)". I think it is going to bring more questions from users, because many will not know what this N refers to. Is that minutes, or seconds? Is that the view number? Is the the number of files it will generate? 
Instead, we could have a label above the button, like for the SPARC acquisition, which would explicitly state something like "View contains N streams". We could also consider a tooltip, but then it'd be harder to discover, so it's less useful for unfamiliar users. 
BTW, we could also add this info to the notification message shown after the export... but that'd be *in addition* to this change, as it's too infrequent to provide a useful feedback.

patrickcleeve2:
Not a priority, closing for now

PR #2735, opened 2024-04-18
[fix] Change odemisd.modelgen default to no strict syntax

pieleric:
For now, most of the microscope files are still incorrect. Defaulting to
strict syntax made the util.testing functions fail to parse the
microscope files.
We could switch that function alone, but it makes more sense to just
change the default, as long as most of the microscope files have errors.


sneepstefan (approved):

sneepstefan:
Just out of curiosity, what do you mean with most of the microscope files are still incorrect? In what context? Is that something we could address as well?

FCanberk (approved):

PR #2737, opened 2024-04-18
[fix] dataio hdf5: don't warn it's not possible to save MD_WL_LIST if it's empty

pieleric:
It's not possible to remove metadata from the component. So if there is
no MD_WL_LIST, the best we can do is to make it None or a empty list.
So don't try to save the MD_WL_LIST if it's empty (ex, for AR data), and
so don't warn that it's not possible to store it.

### DIFF
--- a/src/odemis/util/spectrum.py
+++ b/src/odemis/util/spectrum.py
@@ -70,6 +70,10 @@ def get_wavelength_per_pixel(da):
     # MD_WL_LIST has priority
     try:
         wl = da.metadata[model.MD_WL_LIST]
+        if wl is None or len(wl) == 0:
+            # This is an indirect way to remove metadata (because it's not possible to delete keys
+            # from a component metadata)
+            raise KeyError("Empty MD_WL_LIST metadata")
 
         if len(wl) != da.shape[ci]:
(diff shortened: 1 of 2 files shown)


src/odemis/util/spectrum.py
```diff
@@ -70,6 +70,10 @@ def get_wavelength_per_pixel(da):
     # MD_WL_LIST has priority
     try:
         wl = da.metadata[model.MD_WL_LIST]
+        if wl is None or len(wl) == 0:
```
tepals, line 73: I'm a bit confused as to why this would even be None or empty, could you explain? (I don't completely understand the commit message)

pieleric: The MD_WL_LIST is updated when the spectrograph grating moves or changes. When the grating is set to a "standard" first order diffraction, then this metadata contains the wavelength for each (horizontal) pixel of the CCD. 
However, the grating can be either moved to the 0th order (aka 0mm), or even changed to a mirror grating. In this case, ideally we would remove the MD_WL_LIST metadata. However, it's not possible with `.updateMetadata()` to remove it. So instead it is set to `[]`.

tepals (approved):

sneepstefan (approved):

PR #2738, opened 2024-04-18
[fix] driver semnidaq: AO event buffer not called and incorrect DO period

pieleric:
Fix issue when the AO & DO buffers would be sligtly too small for the
6361, which caused the event callback from not being called. This meant
the AO & DO would not be pushed to the hardware, and it would starve.

=> Increase the minimum buffer size by 1, which seems to be enough (4095
to 4096). It's not very clear the relationship, and couldn't find any
mention of it in the doc, and no one answered on the forum... so let's
just keep this simple solution for now.

Also fix issue when DO period would be incorrectly accepted, due to a
copy-paste error.

### DIFF
--- a/src/odemis/driver/semnidaq.py
+++ b/src/odemis/driver/semnidaq.py
@@ -421,8 +421,8 @@ class AnalogSEM(model.HwComponent):
                         logging.debug("Checking period osr + 1: %s * %s = %s",
                                       ai_period_post, ai_sub_osr_accepted_post, ao_period_post)
                         if (ao_period_post <= ao_period
-                            and self._is_period_valid(ao_task, ao_period)
-                            and (not do_task.do_channels or self._is_period_valid(do_task, ao_period / 2))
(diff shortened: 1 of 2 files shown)


src/odemis/driver/semnidaq.py
```diff
@@ -576,7 +576,10 @@ def __init__(self, sem: AnalogSEM, scanner: "Scanner"):
         self._ao_task = None
         self._ao_data_next_sample = 0
         self._ao_data = None
-        self._min_ao_buffer_n = self._scanner.get_hw_buf_size()
+        # It seems that if the buffer used is <= hw buffer, the events are not triggered
+        # Though, it's not very clear if that's just coincidence, because we actually ask for events
+        # buffer size / 2
```
tepals, line 581: Could you update the comment to make it explicit that this is why you add the `+ 1`

tepals (changes requested):

nandishjpatel (approved):

PR #2740, opened 2024-04-19
[feature] model.DataFlow automatically keep all data when synchronized

pieleric:
Introduce a way to have the .max_discard attribute of the DataFlow be
directly seen by the DataFlowProxy.

With this functionality in place, it's now really worthy to adjust
max_discard when the DataFlow is synchronized, on the backend side
(only).

So introduce a generic .synchronizedOn() method on DataFlow that all
inheriting DataFlows (implemented by the drivers) can use. A lot of
drivers actually already did this, so this mostly simplifies the code.
The main change is that now the proxy really will NOT discard data
when the DataFlow is synchronized (as most of the discard happend on the
client side).

Note that this now "upgrade" the code to only support 0MQ v3+. That is
available since Ubuntu 16.04, so should really not be an issue!

### DIFF
--- a/src/odemis/model/_dataflow.py
+++ b/src/odemis/model/_dataflow.py
@@ -19,6 +19,7 @@ PURPOSE. See the GNU General Public License for more details.
 You should have received a copy of the GNU General Public License along with
 Odemis. If not, see http://www.gnu.org/licenses/.
 """
+
 # Provides data-flow: an object that can contain a large array of data regularly
 # updated. Typically it is used to transmit video (sequence of images). It does it
 # losslessly and with metadata attached (see _metadata for the conventional ones).
@@ -32,6 +33,7 @@ from odemis.util.weak import WeakMethod, WeakRefLostError
 import os
 import threading
 import time
+import weakref
 import zmq
 
 from . import _core
@@ -192,7 +194,7 @@ class DataFlowBase(object):
 
 # DataFlow object to create on the server (in a component)
 class DataFlow(DataFlowBase):
-    def __init__(self, max_discard=100): # XXX max_discard=100
+    def __init__(self, max_discard=100):
         """
         max_discard (int): mount of messages that can be discarded in a row if
                             a new one is already available. 0 to keep (notify)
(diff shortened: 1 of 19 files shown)


FCanberk (approved):

src/odemis/model/_dataflow.py
```diff
@@ -510,42 +559,33 @@ def __del__(self):
 
 
 class SubscribeProxyThread(threading.Thread):
-    def __init__(self, notifier, uri, max_discard, zmq_ctx):
+    def __init__(self, df_proxy, uri, zmq_ctx):
         """
-        notifier (callable): method to call when a new array arrives
+        # notifier (callable): method to call when a new array arrives
```
tepals, line 564: why the `# `?

tepals (changes requested):

src/odemis/model/_dataflow.py
```diff
@@ -19,6 +19,8 @@
 You should have received a copy of the GNU General Public License along with
 Odemis. If not, see http://www.gnu.org/licenses/.
 """
+import weakref
```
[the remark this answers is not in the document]
pieleric: Good catch. There is no reason for this. This was automatically done by Pycharm. Usually it does it correctly.

src/odemis/model/_dataflow.py
pieleric: This has to go away. I'll take care of this.

pieleric:
I've found an issue with this change. It still needs to be addressed.

src/odemis/model/_dataflow.py
```diff
@@ -580,34 +615,37 @@ def run(self):
...
-                        self.w_notifier(darray)
-                    except WeakRefLostError:
-                        return  # It's a sign there is nothing left to do
+        except ReferenceError:  # The DataFlow(Proxy) is gone
+            # => stop this thread too
+            logging.debug("Dataflow proxy %s is gone, stopping the subscription thread", self.uri)
+            return
         except Exception:
             if logging:
                 logging.exception("Ending ZMQ thread due to exception")
```
tepals, line 651: does it make sense to log the exception here?

tepals (approved):
one small comment

src/odemis/model/_dataflow.py
pieleric: Yes, it's quite useful, especially in a thread. Otherwise, any error will stop the thread execution, without *any* feedback.

pieleric:
Let's merge this *after* the 3.5 release.

src/odemis/model/_dataflow.py
tepals: I forgot that `logging.exception` also logs the raised exception, so you can disregard my comment.

PR #2741, opened 2024-04-22
[fix] handle better axes going just at the limit

pieleric:


nandishjpatel (approved):

install/linux/usr/share/odemis/sim/sparc2-time-correlator-sim.odm.yaml
```diff
@@ -164,7 +164,7 @@ SPARC2-VisNIR-TR: {
         address: null,
         axes: ["l1", "l2", "spec-sel", "fiby", "slit"],
         # These values only need to be roughly correct
-        ustepsize: [25.1e-9, 25.1e-9, 25.1e-9, 5.86e-9, 5.e-9], # m/µstep
+        ustepsize: [25.1e-6, 25.1e-6, 25.1e-6, 5.86e-6, 5.e-7], # m/µstep
```
nandishjpatel, line 167:
```suggestion
        ustepsize: [25.1e-6, 25.1e-6, 25.1e-6, 5.86e-6, 5.e-6], # m/µstep
```
Maybe?

install/linux/usr/share/odemis/sim/sparc2-time-correlator-sim.odm.yaml
```diff
@@ -421,8 +412,9 @@ SPARC2-VisNIR-TR: {
     init: {
         axis_name: "od",
         # position when referenced (adjusted to be a multiple of the actuator stepsize)
-        offset: -3.00000925, # od
-        ref_start: 2.8 #  a little before the reference (inverted axis)
+        offset: -2.98, # od
+        ref_start: 2.78, #  a little before the reference (inverted axis)
+        ref_period: null,  # never automatically reference the axis
```
sneepstefan, line 417: Could you update the comment with the unit of the reference period when the auto reference is needed?

src/odemis/acq/stream/_base.py
```diff
@@ -681,8 +681,13 @@ def _update_linked_position(self, act, pos):
...
             # update va
-            va.value = axpos
-            logging.info("Updating local axis %s to position %s", va_name, axpos)
+            axpos_clipped = va.clip(axpos)  # In case the axis went out of range
+            va.value = axpos_clipped
+            if axpos == axpos_clipped:
+                logging.info("Updating local axis %s to position %s", va_name, axpos)
+            else:
+                logging.warning("Updating local axis %s to clipped position %s (actual pos = %s)",
+                                va_name, axpos_clipped, axpos)
```
sneepstefan, line 690: Shouldn't the message be the other way around? e.g. if the axis position is the same as the axis position clipped -> Updating local axis %s to clipped position. Or do I have to read it differently?

src/odemis/driver/tmcm.py
```diff
@@ -417,7 +417,10 @@ def __init__(self, name, role, port, axes, ustepsize, address=None,
             if sw_rng is not None:
                 if not sw_rng[0] <= 0 <= sw_rng[1]:
                     raise ValueError("Range of axis %d doesn't include 0: %s" % (i, sw_rng))
-                phy_rng = (max(phy_rng[0], sw_rng[0]), min(phy_rng[1], sw_rng[1]))
+                # Make sure the range is a multiple of the ustepsize
```
sneepstefan, line 420: This sounds logical but why is that necessary? Probably due to issues in the field with the tmcm?

sneepstefan (changes requested):

src/odemis/acq/stream/_base.py
pieleric: If the clipped value is the same are the original value, it means the "clipping" had no effect, which means it wasn't clipped. So we say it's been clipped if the clipped value is different from the original value. The code is correct.

src/odemis/driver/tmcm.py
pieleric: Good catch. The comment needs to explain *why* (not *what*). Essentially, it's because if it's not rounded and the user moves to the limit, the hardware will round the value, and there is 50% chances it ends up a tiny bit outside the range... which confuses some code (and some users).

sneepstefan (approved):

PR #2745, opened 2024-04-24
[fix] Fastem SEM autofocus also pass se_detector and ebeam_focus for …

tepals:
…pre-calibrations

The SEM autofocus was added to the pre-calibrations. However the components used for running the SEM autofocus were not passed, this caused the acquisition to fail.

SDEV-1843

### DIFF
--- a/src/odemis/acq/fastem.py
+++ b/src/odemis/acq/fastem.py
@@ -387,6 +387,8 @@ def acquire(roa, path, scanner, multibeam, descanner, detector, stage, scan_stag
     :param ccd: (model.DigitalCamera) A camera object of the diagnostic camera.
     :param beamshift: (tfsbc.BeamShiftController) Component that controls the beamshift deflection.
     :param lens: (static.OpticalLens) Optical lens component.
+    :param se_detector: (model.Detector) single beam secondary electron detector.
+    :param ebeam_focus: (model.Actuator) SEM focus control.
     :param pre_calibrations: (list[Calibrations]) List of calibrations that should be run before the ROA acquisition.
                              Default is None.
     :param save_full_cells: (bool) If True save the full cell images instead of cropping them
@@ -441,6 +443,8 @@ class AcquisitionTask(object):
         :param ccd: (model.DigitalCamera) A camera object of the diagnostic camera.
         :param beamshift: (tfsbc.BeamShiftController) Component that controls the beamshift deflection.
         :param lens: (static.OpticalLens) Optical lens component.
+        :param se_detector: (model.Detector) single beam secondary electron detector.
+        :param ebeam_focus: (model.Actuator) SEM focus control.
         :param roa: (FastEMROA) The acquisition region object to be acquired (megafield).
         :param path: (str) Path on the external storage where the image data is stored. Here, it is possible
                     to specify sub-directories (such as acquisition date and project name) additional to the main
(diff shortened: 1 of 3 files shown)


tepals:
@nandishjpatel I think this fixes the issues related to the SEM autofocusing. However not all test cases pass, for instance `odemis.acq.test.fastem_test.TestFastEMAcquisition.test_acquire_ROA` fails because it expects 6 fields to be acquired, but actually we acquire 8 fields. I it seems to be because we're not using get_square_field_indices anymore, but now we use get_poly_indices. Could you look into that?

nandishjpatel:
> @nandishjpatel I think this fixes the issues related to the SEM autofocusing. However not all test cases pass, for instance `odemis.acq.test.fastem_test.TestFastEMAcquisition.test_acquire_ROA` fails because it expects 6 fields to be acquired, but actually we acquire 8 fields. I it seems to be because we're not using get_square_field_indices anymore, but now we use get_poly_indices. Could you look into that?

This is tricky! I run get_poly_field_indices() and get_square_field_indices() for the same points and they result in different (col, row) indices

```
INFO:root:polygon [(0, 0), (1, 0), (0, 1), (1, 1), (0, 2), (1, 2), (0, 3), (1, 3)] --> 2 columns, 4 rows
INFO:root:square [(0, 0), (1, 0), (2, 0), (0, 1), (1, 1), (2, 1)] --> 3 columns, 2 rows
```

@tepals should we just adjust the testcase for now? and then investigate more? I think the `get_poly_field_indices` needs to be refactored!

tepals:
> @tepals should we just adjust the testcase for now? and then investigate more? I think the `get_poly_field_indices` needs to be refactored!

Let's do that for now, although it is risky to adjust test cases to make the test pass, so we should make sure we do not forget that this is an issue. For now I think that can be done in a separate PR, could you then review this PR?

nandishjpatel (approved):

src/odemis/acq/fastem.py
```diff
@@ -366,7 +366,7 @@ def estimate_acquisition_time(roa, pre_calibrations=None):
 
 
 def acquire(roa, path, scanner, multibeam, descanner, detector, stage, scan_stage, ccd, beamshift, lens,
-            pre_calibrations=None, save_full_cells=False, settings_obs=None):
+            se_detector, ebeam_focus, pre_calibrations=None, save_full_cells=False, settings_obs=None):
     """
     Start a megafield acquisition task for a given region of acquisition (ROA).
```
sneepstefan, line 371: Could  you also add the description of these 2 additional parameters to the docstring?

src/odemis/acq/fastem.py
```diff
@@ -425,7 +425,8 @@ class AcquisitionTask(object):
     An ROA consists of multiple single field images.
     """
 
-    def __init__(self, scanner, multibeam, descanner, detector, stage, scan_stage, ccd, beamshift, lens, roa, path,
+    def __init__(self, scanner, multibeam, descanner, detector, stage, scan_stage, ccd, beamshift, lens,
+                 se_detector, ebeam_focus, roa, path,
                  pre_calibrations, save_full_cells, settings_obs, future):
         """
         :param scanner: (xt_client.Scanner) Scanner component connecting to the XT adapter.
```
sneepstefan, line 432: Could  you also add the description of these 2 additional parameters to the docstring?

sneepstefan (changes requested):

src/odemis/acq/fastem.py
tepals: good catch, fixed

sneepstefan (approved):

PR #2746, opened 2024-04-26
[feat] fastem: reference the stage before overview image as well as a…

tepals:
…fter reloading the sample holder

The stage was not referenced after reloading a sample, this caused the user to see a misalignment between the live view and the overview image. To make sure this does not happen again we will automatically reference the stage after reloading (and the chamber is thus in vacuum) the sample holder and before acquiring an overview image.

### DIFF
--- a/src/odemis/acq/fastem.py
+++ b/src/odemis/acq/fastem.py
@@ -1015,6 +1015,10 @@ class OverviewAcquisition(object):
 
         :returns: (DataArray) The complete overview image.
         """
+        logging.debug("Referencing stage axes x and y.")
+        f = stage.reference({"x", "y"})
+        f.result(timeout=180)
+
         # Get the current immersion mode value before configuring the scanner.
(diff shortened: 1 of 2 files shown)


K4rishma (approved):

src/odemis/acq/fastem.py
```diff
@@ -1015,6 +1015,10 @@ def run(self, stream, stage, area, live_stream):
 
         :returns: (DataArray) The complete overview image.
         """
+        logging.debug("Referencing stage axes x and y.")
+        f = stage.reference({"x", "y"})
+        f.result()
```
nandishjpatel, line 1020: Maybe good to have a timeout here?

nandishjpatel (approved):
1 comment to address otherwise it looks good

src/odemis/acq/fastem.py
tepals: Makes sense, I added it in both places.

PR #2750, opened 2024-05-01
[fix][METEOR-1015] Change remove stream icon from a dash to an X.

patrickcleeve2:
User's are getting confused about the remove stream functionality, as it is mismatched with the OS. Typically the - symbol means minimise, and X means close. This PR changes the icon to be consistent with the OS

### DIFF
--- a/src/odemis/gui/comp/stream_panel.py
+++ b/src/odemis/gui/comp/stream_panel.py
@@ -140,9 +140,9 @@ class StreamPanelHeader(wx.Control):
     def _add_remove_btn(self):
         """ Add a button for stream removal """
         btn_rem = buttons.ImageButton(self,
-                                      bitmap=img.getBitmap("icon/ico_rem_str.png"),
+                                      bitmap=img.getBitmap("icon/ico_clear.png"),
                                       size=self.BUTTON_SIZE)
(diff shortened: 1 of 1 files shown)


FCanberk (approved):

src/odemis/gui/comp/stream_panel.py
```diff
@@ -140,9 +140,9 @@ def __init__(self, parent, wid=wx.ID_ANY, pos=wx.DefaultPosition, size=wx.Defaul
     def _add_remove_btn(self):
         """ Add a button for stream removal """
         btn_rem = buttons.ImageButton(self,
-                                      bitmap=img.getBitmap("icon/ico_rem_str.png"),
```
nandishjpatel, line 143: If add stream button is a `+` shouldn't remove stream button be a `-`?

patrickcleeve2: Users are assuming that the - is minimise, because that's the OS symbol, and closing when they mean to minimise. So the idea is to match the OS where X is close/ remove.

nandishjpatel (approved):

patrickcleeve2:
I think this can be merged @pieleric

PR #2751, opened 2024-05-03
[feature][METEOR-590] Implement reading raw coordinates of TFS stage

K4rishma:
- read raw coordinates instead of linked coordinates
- correct for offset in x and y such that the coordinated read by TFS and odemis is same in raw coordinate system
- This change requires v1.12.0 of xtadapter on support PC

### DIFF
--- a/src/odemis/driver/xt_client.py
+++ b/src/odemis/driver/xt_client.py
@@ -31,11 +31,11 @@ from concurrent import futures
 from concurrent.futures import CancelledError
 from typing import Any, Dict, Optional, Set, Tuple, Union
 
+import Pyro5.api
 import msgpack_numpy
 import notify2
 import numpy
 import pkg_resources
-import Pyro5.api
 from Pyro5.errors import CommunicationError
 
 from odemis import model
@@ -377,7 +377,7 @@ class SEM(model.HwComponent):
             self.server._pyroClaimOwnership()
             self.server.move_stage(position, rel)
 
-    def set_raw_coordinate_system(self, raw_coordinates:bool = True):
+    def set_raw_coordinate_system(self, raw_coordinates: bool = True):
         """
         Read raw z coordinate or linked z coordinate when requesting stage coordinates
         :param raw_coordinates: (bool) If True, request raw z stage coordinates, otherwise request linked Z coordiantes
@@ -386,10 +386,11 @@ class SEM(model.HwComponent):
             with self._proxy_access:
                 self.server._pyroClaimOwnership()
                 self.server.set_raw_coordinate_system(raw_coordinates)
-        except AttributeError:
+        except OSError:
             if raw_coordinates:
                 raise NotImplementedError("Ensure the version of Delmic XT Adapter is >= 1.12.0")
-            # If non-raw coordinates are requested, proceed with the current coordinate system
+            # Do not raise an error if non-raw coordinates are requested, because non-raw is the default in old
+            # versions of the xtadapter
 
     def get_raw_stage_position(self):
         """Read the stage position in raw coordinate system"""
--- a/src/odemis/driver/xt_client.py
+++ b/src/odemis/driver/xt_client.py
@@ -380,21 +380,20 @@ class SEM(model.HwComponent):
     def set_raw_coordinate_system(self, raw_coordinates: bool = True):
         """
         Read raw z coordinate or linked z coordinate when requesting stage coordinates
-        :param raw_coordinates: (bool) If True, request raw z stage coordinates, otherwise request linked Z coordiantes
+        :param raw_coordinates: (bool) If True, request raw z stage coordinates, otherwise request linked Z coordinates
         """
-        try:
-            with self._proxy_access:
-                self.server._pyroClaimOwnership()
-                self.server.set_raw_coordinate_system(raw_coordinates)
-        except OSError:
-            if raw_coordinates:
-                raise NotImplementedError("Ensure the version of Delmic XT Adapter is >= 1.12.0")
-            # Do not raise an error if non-raw coordinates are requested, because non-raw is the default in old
-            # versions of the xtadapter
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            self.server.set_raw_coordinate_system(raw_coordinates)
 
-    def get_raw_stage_position(self):
-        """Read the stage position in raw coordinate system"""
-        pass
+    def get_raw_coordinate_system(self):
+        """
+        Get raw coordinates system of the stage
+        :return: (bool) True if stage is read in raw coordinates, otherwise, False
+        """
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            return self.server.get_raw_coordinate_system()
 
     def stage_is_moving(self) -> bool:
         """
@@ -2222,10 +2221,10 @@ class Stage(model.Actuator):
         self._pos_poll = util.RepeatingTimer(5, self._refreshPosition, "Stage position polling")
         self._pos_poll.start()
 
-    def _switch_coordinate_system(self, raw_coordinates) -> None:
+    def _switch_coordinate_system(self, raw_coordinates: bool) -> None:
         """Calculate the offset in linear stage axes x and y when the stage coordinates are read in raw coordinates such
          that it is equal to x and y when stage coordinates are read in linked coordinates.
-         :param raw_coordinates: (bool) True if stage coordinates are read in raw system else False
+         :param raw_coordinates: True if stage coordinates are read in raw system else False
          """
         if raw_coordinates:
             self.parent.set_raw_coordinate_system(False)
@@ -2243,7 +2242,13 @@ class Stage(model.Actuator):
                           f"system for raw stage coordinates: {pos}, non-raw stage coordinates: {pos_linked}")
         else:
             # Make sure the system is read in the linked coordinate system
-            self.parent.set_raw_coordinate_system(False)
+            try:
+                self.parent.set_raw_coordinate_system(False)
+            except OSError as error_msg:
+                logging.warning("Delmic XT Adapter >= 1.12.0 is required to set the raw coordinate system. "
+                                "Failed to call the raw coordinate system for the stage: %s", error_msg)
+                # Do not raise an error if non-raw coordinates are requested, because non-raw is the default in old
+                # versions of the xtadapter
 
     def _updatePosition(self) -> None:
         """
--- a/src/odemis/driver/xt_client.py
+++ b/src/odemis/driver/xt_client.py
@@ -358,6 +358,7 @@ class SEM(model.HwComponent):
                     update.show()
                 else:
                     logging.warning("{} is a bad file in {} not transferring latest package.".format(ret, package.path))
+
         except Exception:
             logging.warning("Failure during transfer latest xtadapter package (non critical)", exc_info=True)
 
@@ -375,6 +376,24 @@ class SEM(model.HwComponent):
         with self._proxy_access:
             self.server._pyroClaimOwnership()
             self.server.move_stage(position, rel)
+
+    def set_raw_coordinate_system(self, raw_coordinates:bool = True):
+        """
+        Read raw z coordinate or linked z coordinate when requesting stage coordinates
+        :param raw_coordinates: (bool) If True, request raw z stage coordinates, otherwise request linked Z coordiantes
+        """
+        try:
+            with self._proxy_access:
+                self.server._pyroClaimOwnership()
+                self.server.set_raw_coordinate_system(raw_coordinates)
+        except AttributeError:
+            if raw_coordinates:
+                raise NotImplementedError("Ensure the version of Delmic XT Adapter is >= 1.12.0")
+            # If non-raw coordinates are requested, proceed with the current coordinate system
+
+    def get_raw_stage_position(self):
+        """Read the stage position in raw coordinate system"""
+        pass
 
     def stage_is_moving(self) -> bool:
         """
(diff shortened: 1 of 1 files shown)


src/odemis/driver/xt_client.py
```diff
@@ -358,6 +358,16 @@ def check_and_transfer_latest_package(self) -> None:
...
                     logging.warning("{} is a bad file in {} not transferring latest package.".format(ret, package.path))
+
+                # This version supports reading raw stage coordinates
+                if current_version is not None and bitness is not None:
+                    if current_version <= "1.12.0":
+                        raise ValueError("Required Delmic XT Adapter version shoulb be >= 1.12.0",
+                                         "How to update?\n\n1. Full stop Odemis and close Delmic XT Adapter.\n"
+                                         "2. Uninstall the old Delmic XT Adapter and install the correct version.")
+                else:
+                    logging.warning(
```
K4rishma, line 369: Should this be an error ? How about if the Delmix XT adaptor is in development version?

src/odemis/driver/xt_client.py
```diff
@@ -2170,7 +2196,7 @@ def __init__(self, name: str, role: str, parent: model.HwComponent, rng: dict =
             "rz": model.Axis(unit=stage_info["unit"]["r"], range=rng["rz"]),
         }
 
-        model.Actuator.__init__(self, name, role, parent=parent, axes=axes_def,
+        model.Actuator.__init__(self, name, role, parent=self.parent, axes=axes_def,
```
pieleric, line 2199: This function takes care of setting `self.parent` as `parent`. So no need to do it early explicitly.

src/odemis/driver/xt_client.py
```diff
@@ -2147,10 +2166,17 @@ class Stage(model.Actuator):
     moving the TFS stage and updating the position.
     """
 
-    def __init__(self, name: str, role: str, parent: model.HwComponent, rng: dict = None, **kwargs) -> None:
+    def __init__(self, name: str, role: str, parent: model.HwComponent, rng: dict = None, raw: bool = False,
+                 **kwargs) -> None:
```
pieleric, line 2170: This needs a docstring. At least it should document the new "raw" argument. It'd be a good opportunity to document rng too.

src/odemis/driver/xt_client.py
```diff
@@ -2147,10 +2166,17 @@ class Stage(model.Actuator):
     moving the TFS stage and updating the position.
     """
 
-    def __init__(self, name: str, role: str, parent: model.HwComponent, rng: dict = None, **kwargs) -> None:
+    def __init__(self, name: str, role: str, parent: model.HwComponent, rng: dict = None, raw: bool = False,
```
pieleric, line 2169: "raw" a little bit generic word. I think that when encountering it in a microscope file, a lot of people wouldn't understand what it refers too. You can of course count on comments, but I'd suggest to name it `raw_coordinates` to make it clearer immediately.

src/odemis/driver/xt_client.py
```diff
@@ -2147,10 +2166,17 @@ class Stage(model.Actuator):
...
+                 **kwargs) -> None:
+        self.raw = raw
+        self.parent = parent
+        # define axes which needs offset correction such that TFS stage coordinates
+        # are equal to stage coordinates displayed through Odemis
```
pieleric, line 2174: Please explain more this comment. At least you should explain that 1) it's related to the raw coordinates, and 2) it'll be updated later in the init.

src/odemis/driver/xt_client.py
```diff
@@ -2147,10 +2166,17 @@ class Stage(model.Actuator):
...
+        self.offset = {"x": 0, "y": 0}
```
pieleric, line 2175: Let's a use a more specific name like `_raw_offset`.

src/odemis/driver/xt_client.py
```diff
@@ -2147,10 +2166,17 @@ class Stage(model.Actuator):
     moving the TFS stage and updating the position.
     """
 
-    def __init__(self, name: str, role: str, parent: model.HwComponent, rng: dict = None, **kwargs) -> None:
+    def __init__(self, name: str, role: str, parent: model.HwComponent, rng: dict = None, raw: bool = False,
+                 **kwargs) -> None:
+        self.raw = raw
```
pieleric, line 2171: It's private -> `_raw` (or `_raw_coordinates`)

src/odemis/driver/xt_client.py
```diff
@@ -2179,16 +2205,42 @@ def __init__(self, name: str, role: str, parent: model.HwComponent, rng: dict =
...
+            self.parent.set_raw_coordinate_system(self.raw)
+            pos = self._getPosition()
+            logging.debug(f"Raw stage coordinates {pos}")
+            self.parent.set_raw_coordinate_system(False)
+            pos_after = self._getPosition()
+            logging.debug(f"Linked stage coordinates {pos_after}")
+            for axis in self.offset.keys():
+                self.offset[axis] = pos_after[axis] - pos[axis]
+            # the offset should only be in linear axes, it is not expected in rotational axes
+            assert all(pos[axis] == pos_after[axis] for axis in ["rx", "rz"])
```
pieleric, line 2228: It's interesting to check, but let's not immediately completely fail in this case. Just log a warning.

src/odemis/driver/xt_client.py
```diff
@@ -2179,16 +2205,42 @@ def __init__(self, name: str, role: str, parent: model.HwComponent, rng: dict =
                                                 readonly=True)
         self._updatePosition()
 
+        # update the offset if raw stage coordinates are read
+        self._get_offset_raw_coordinate_system()
```
pieleric, line 2209: This function currently does actually 2 things: measure the offset, and change the raw coordinates. The name implies it just does one (and return the value, which it doesn't do).
I'd suggest to call it like `self._switch_coordinate_system(self._raw_coordinates)`. (it takes 1 argument)

src/odemis/driver/xt_client.py
```diff
@@ -2179,16 +2205,42 @@ def __init__(self, name: str, role: str, parent: model.HwComponent, rng: dict =
...
+
         # Refresh regularly the position
         self._pos_poll = util.RepeatingTimer(5, self._refreshPosition, "Stage position polling")
         self._pos_poll.start()
 
+    def _get_offset_raw_coordinate_system(self) -> None:
+        """Calculate the offset in linear stage axes x and y when the stage coordinates are read in raw coordinates such
+         that it is equal to x and y when stage coordinates are read in linked coordinates."""
+        if self.raw:
```
pieleric, line 2218: If not raw, it should also explicitly change the raw coordinate system, so that we are certain it's in the coordinate system we expect.
If it fails because the function doesn't exist, it's fine (as that's typically the standard value).

src/odemis/driver/xt_client.py
```diff
@@ -2179,16 +2205,42 @@ def __init__(self, name: str, role: str, parent: model.HwComponent, rng: dict =
...
+            self.parent.set_raw_coordinate_system(self.raw)
```
pieleric, line 2219: Instead of changing 3 times, start with the "non raw coordinates" system, and then switch to the raw version.

src/odemis/driver/xt_client.py
```diff
@@ -2179,16 +2205,42 @@ def __init__(self, name: str, role: str, parent: model.HwComponent, rng: dict =
...
         """
         update the position VA
         """
         old_pos = self.position.value
         pos = self._getPosition()
+        if self.raw:
+            # correct for the offset such that the stage coordinates displayed in
+            # TFS software is the same as Odemis
+            pos["x"] = pos["x"] + self.offset["x"]
+            pos["y"] = pos["y"] + self.offset["y"]
```
pieleric, line 2243:
```suggestion
            pos["x"] += self.offset["x"]
            pos["y"] += self.offset["y"]
```

src/odemis/driver/xt_client.py
```diff
@@ -2235,6 +2287,13 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
                     pos["t"] = pos.pop("rx")
                 if "rz" in pos.keys():
                     pos["r"] = pos.pop("rz")
+
+                # Correct for offset compensation in absolute movements
+                if self.raw and not rel:
+                    if "x" in pos.keys():
+                        pos["x"] = pos["x"] - self.offset["x"]
```
pieleric, line 2294:
```suggestion
                        pos["x"] -= self.offset["x"]
```
(same for y)

src/odemis/driver/xt_client.py
```diff
@@ -2179,16 +2205,42 @@ def __init__(self, name: str, role: str, parent: model.HwComponent, rng: dict =
...
+            pos = self._getPosition()
+            logging.debug(f"Raw stage coordinates {pos}")
+            self.parent.set_raw_coordinate_system(False)
+            pos_after = self._getPosition()
+            logging.debug(f"Linked stage coordinates {pos_after}")
+            for axis in self.offset.keys():
+                self.offset[axis] = pos_after[axis] - pos[axis]
+            # the offset should only be in linear axes, it is not expected in rotational axes
+            assert all(pos[axis] == pos_after[axis] for axis in ["rx", "rz"])
+            logging.debug(f"The offset values in x and y are {self.offset} when stage is in the raw coordinate system")
```
pieleric, line 2229: Please only put 1 logging message for this function. It should say the non-raw & raw values, and conclude on the offset, as a single sentence.

src/odemis/driver/xt_client.py
```diff
@@ -376,6 +386,15 @@ def move_stage(self, position: Dict[str, float], rel: bool = False) -> None:
             self.server._pyroClaimOwnership()
             self.server.move_stage(position, rel)
 
+    def set_raw_coordinate_system(self, raw: bool = True):
```
pieleric, line 389: Even if you don't use it now, can you add the `get_raw_coordinate_system()` function so that we know it exists?

src/odemis/driver/xt_client.py
```diff
@@ -358,6 +358,16 @@ def check_and_transfer_latest_package(self) -> None:
                     update.show()
                 else:
                     logging.warning("{} is a bad file in {} not transferring latest package.".format(ret, package.path))
+
+                # This version supports reading raw stage coordinates
+                if current_version is not None and bitness is not None:
+                    if current_version <= "1.12.0":
```
pieleric, line 364: FYI, you cannot compare versions with string comparison. For instance "9" > "10"! There are dedicated functions for that.

src/odemis/driver/xt_client.py
```diff
@@ -358,6 +358,16 @@ def check_and_transfer_latest_package(self) -> None:
                     update.show()
                 else:
                     logging.warning("{} is a bad file in {} not transferring latest package.".format(ret, package.path))
+
+                # This version supports reading raw stage coordinates
+                if current_version is not None and bitness is not None:
```
pieleric, line 363: Don't check explicitly for the version. Instead, I would suggest that if (and *only if*) the raw coordinates are requested, and the call to `set_raw_coordinate_system()`  fails *because the method doesn't exist*, then you raise a ValueError (or maybe, more appropriately, a NotImplementedError).

So, in particular, if the user asks raw_coordinates = False, and calling the `set_raw_coordinate_system(False)` fails, it should continue business as usual.

pieleric (changes requested):

K4rishma:
Ready for review

src/odemis/driver/xt_client.py
```diff
@@ -376,6 +377,24 @@ def move_stage(self, position: Dict[str, float], rel: bool = False) -> None:
...
+    def set_raw_coordinate_system(self, raw_coordinates:bool = True):
+        """
+        Read raw z coordinate or linked z coordinate when requesting stage coordinates
+        :param raw_coordinates: (bool) If True, request raw z stage coordinates, otherwise request linked Z coordiantes
+        """
+        try:
+            with self._proxy_access:
+                self.server._pyroClaimOwnership()
+                self.server.set_raw_coordinate_system(raw_coordinates)
+        except AttributeError:
```
tepals, line 389: are you sure this will raise an `AttributeError`? I thought Pyro5 translated everything into an `OSError`

src/odemis/driver/xt_client.py
```diff
@@ -376,6 +377,24 @@ def move_stage(self, position: Dict[str, float], rel: bool = False) -> None:
...
+            if raw_coordinates:
+                raise NotImplementedError("Ensure the version of Delmic XT Adapter is >= 1.12.0")
+            # If non-raw coordinates are requested, proceed with the current coordinate system
```
tepals, line 392:
```suggestion
                raise NotImplementedError("Ensure the version of Delmic XT Adapter is >= 1.12.0")
            # Do not raise an error if non-raw coordinates are requested, because non-raw is the default in old versions of the xtadapter
```

src/odemis/driver/xt_client.py
```diff
@@ -376,6 +377,24 @@ def move_stage(self, position: Dict[str, float], rel: bool = False) -> None:
...
+
+    def get_raw_stage_position(self):
+        """Read the stage position in raw coordinate system"""
+        pass
```
tepals, line 396: why is this here?

tepals (changes requested):

src/odemis/driver/xt_client.py
K4rishma: @tepals the reason of adding this function is explained here

tepals: I think what Éric meant (but correct me if I'm wrong) is that since you have the function on the xtadapter side, you might as well implement it here. Now you just have it implemented as an empty function, while you could also implement the working call to the xtadapter. So not just do `pass` but actually call the function.

src/odemis/driver/xt_client.py
K4rishma: I am not sure about it either, I have added it like below. In SEM class, I see mostly timeout errors and I see OS errors in Detector class
`        except OSError:
            if raw_coordinates:
                raise NotImplementedError("Ensure the version of Delmic XT Adapter is >= 1.12.0")`

src/odemis/driver/xt_client.py
K4rishma: based on Eric's remark before I added this function

src/odemis/driver/xt_client.py
pieleric: Yes, I confirm this interpretation from Thera :-)
You have to actually implement `get_raw_coordinate_system()`: call the xtadapter's function and return the returned value)

src/odemis/driver/xt_client.py
```diff
@@ -376,6 +377,25 @@ def move_stage(self, position: Dict[str, float], rel: bool = False) -> None:
...
+        """
+        try:
+            with self._proxy_access:
+                self.server._pyroClaimOwnership()
+                self.server.set_raw_coordinate_system(raw_coordinates)
+        except OSError:
+            if raw_coordinates:
+                raise NotImplementedError("Ensure the version of Delmic XT Adapter is >= 1.12.0")
+            # Do not raise an error if non-raw coordinates are requested, because non-raw is the default in old
+            # versions of the xtadapter
```
pieleric, line 393: This shouldn't be hidden for every call. Only at init, when the `raw_coordinates` argument is `False` (which is the default). => move this `try`/`except` to line 2246.

pieleric (changes requested):

pieleric (approved):

patrickcleeve2 (approved):

PR #2752, opened 2024-05-06
[SDEV-1871] investigate gui software lag find proper fix

nandishjpatel:
NOTE: https://github.com/delmic/odemis/pull/2726 must be reviewed first


src/odemis/gui/comp/overlay/shapes.py
```diff
@@ -182,6 +184,20 @@ def _activate_shapes(self, flag=True):
...
+            if shape.selected.value and shape != self._selected_shape:
+                shape.selected.value = False
+
+    def _unbind_shape_events(self, shape: EditableShape):
+        """Unbind shape's events bound to the canvas."""
+        shape.cnvs.Unbind(wx.EVT_LEFT_DOWN, handler=shape.on_left_down)
+        shape.cnvs.Unbind(wx.EVT_LEFT_UP, handler=shape.on_left_up)
+        shape.cnvs.Unbind(wx.EVT_RIGHT_DOWN, handler=shape.on_right_down)
+        shape.cnvs.Unbind(wx.EVT_RIGHT_UP, handler=shape.on_right_up)
+        shape.cnvs.Unbind(wx.EVT_MOTION, handler=shape.on_motion)
```
pieleric, line 199: This works... but it just adds complexity on top of complexity. You're essentially undoing the Overlay code. I'd suggest to just not consider the shapes as Overlays. Their .draw() function can be called from the ShapeOverlay.draw(). They don't need to be part of the canvas's world overlays. 
You can merge RectanglePointsSelectOverlay into RectangleOverlay (because it's the only class that uses it). Then, pretty much the only thing that is left from the "WorldOverlay" class is the `.active` VA, which you can duplicate... and probably it's not even necessary, but can just be passed as an argument to .draw() (by the ShapeOverlay).

pieleric (changes requested):

src/odemis/gui/comp/overlay/shapes.py
nandishjpatel: Do you foresee  in the future that someone will use a (+) button to draw a single `RectanglePointsSelectOverlay`, `PolygonOverlay`, `EllipseOverlay`, or `RectangleOverlay` @pieleric ? I still left those classes inherit the `WorldOverlay` because of it

pieleric: No, I don't think this will be a coming feature. Especially, even if it's a case, that can still be implemented by having a ShapeOverlay + a way to tell it to create a shape. (Like this is done for the GadgetOverlay.

src/odemis/gui/comp/overlay/shapes.py
```diff
@@ -180,39 +180,37 @@ def on_leave(self, evt):
...
+        Find a shape corresponding to the given view position.
         :return: the most appropriate shape.
             If no shape is found, it returns None.
         """
         if self._shapes:
-            pos = self.cnvs.view_to_phys(evt.Position, self.cnvs.get_half_buffer_size())
-            for shape in self._shapes[::-1]:
-                if shape.is_point_in_shape(pos):
-                    return shape
+            v_centers = numpy.array([shape.v_center for shape in self._shapes])
```
nandishjpatel, line 200: @pieleric and @tepals FYI, the performance improvement if here in `_get_shape` + letting `ShapesOverlay` handle the mouse events and not the `EditableShape`.

tepals (changes requested):

src/odemis/gui/comp/overlay/shapes.py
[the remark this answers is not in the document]
nandishjpatel: > would it make sense to store the v_centers? 

I also thought the same and had implemented in the `draw()` function, the v_centers are updated after drawing. Do you think it makes sense to have it there @tepals ? I was worried the draw method being called more frequently for example `on_motion`. For my laptop `_get_shape` function takes less than a millisecond for hundreds of ROAs

src/odemis/gui/comp/overlay/shapes.py
```diff
@@ -180,39 +180,37 @@ def on_leave(self, evt):
...
-            self.active.value = True
-            self._activate_shapes(True)
-        else:
-            self.active.value = False
-            self._activate_shapes(False)
-            self.cnvs.reset_default_cursor()
+        self.active.value = selected_tool == self.tool
 
-    def _get_shape(self, evt):
+    def _get_shape(self, v_pos: Tuple[float, float]):
```
pieleric, line 193: Please also indicate the return type.

src/odemis/gui/comp/overlay/shapes.py
```diff
@@ -180,39 +180,37 @@ def on_leave(self, evt):
...
             If no shape is found, it returns None.
         """
         if self._shapes:
-            pos = self.cnvs.view_to_phys(evt.Position, self.cnvs.get_half_buffer_size())
-            for shape in self._shapes[::-1]:
-                if shape.is_point_in_shape(pos):
-                    return shape
+            v_centers = numpy.array([shape.v_center for shape in self._shapes])
+            v_centers_kdtree = cKDTree(v_centers)
+            distances, indices = v_centers_kdtree.query(numpy.array([v_pos[0], v_pos[1]]), k=4)
```
pieleric, line 202: It should be fine to just write `numpy.array(v_pos)`

src/odemis/gui/comp/overlay/rectangle.py
```diff
@@ -88,18 +138,83 @@ def restore_state(self, state: RectangleState):
         self._points = self.get_physical_sel()
         self.points.value = self._points
 
-    def is_point_in_shape(self, point):
+    def is_point_in_shape(self, v_point):
         if self._points:
-            return point_in_polygon(point, self._points)
+            hover, _ = self.get_hover(v_point)
+            return bool(hover)
```
pieleric, line 144: This converts HOVER_* to a bool by relying on the fact that HOVER_NONE == 0. I think it'd be a lot more explicit and safe (for the long term) to write `return hover != HOVER_NONE`.

src/odemis/gui/comp/overlay/rectangle.py
```diff
@@ -88,18 +138,83 @@ def restore_state(self, state: RectangleState):
         self._points = self.get_physical_sel()
         self.points.value = self._points
 
-    def is_point_in_shape(self, point):
+    def is_point_in_shape(self, v_point):
         if self._points:
-            return point_in_polygon(point, self._points)
+            hover, _ = self.get_hover(v_point)
```
pieleric, line 143: This change is odd. Firstly, `get_hover()` does extra things, and then calls `point_in_polygon()`, so it can only be slower than before. While this PR is about making things faster. Also, `get_hover()` will return true for all the "edges", including the rotation knob, which is can be outside the shape. So now the function will return True even if the point is not within the shape. Either, this is what you want, and you should rename the function (and add a docstring), or it's not what you want, and actually should use only `point_in_polygon()`.

src/odemis/gui/comp/overlay/rectangle.py
```diff
@@ -88,18 +138,83 @@ def restore_state(self, state: RectangleState):
         self._points = self.get_physical_sel()
         self.points.value = self._points
 
-    def is_point_in_shape(self, point):
+    def is_point_in_shape(self, v_point):
```
pieleric, line 141: Please add type info (for input and output).

src/odemis/gui/comp/overlay/shapes.py
```diff
@@ -180,39 +180,37 @@ def on_leave(self, evt):
...
+            for distance, index in zip(distances, indices):
```
pieleric, line 203: If I understand correctly, it'll only look at the 4 shapes with the centers the closest to the v_pos. If none of them is over the point, then it's considered that no shape is over the point? Is that correct? 
If so, that's just an heuristic, relying on the fact that most shapes are about the same area. If there is one very large shape, this heuristic could easily fail, right?
If so, I'd suggest to at least document this limitation. Ideally, you'd ensure that you *always* find a shape if there is  one that covers the point.

Also, wouldn't it be simpler, instead of using KDTree on the centers, to keep track of the bounding box of each shape, and use that BBox to quickly discard far away shapes?

src/odemis/gui/comp/overlay/shapes.py
```diff
@@ -241,12 +238,13 @@ def on_left_down(self, evt):
...
         else:
-            self._selected_shape = self._get_shape(evt)
+            self._selected_shape = self._get_shape(evt.Position)
             if self._selected_shape is None:
                 self._selected_shape = self._create_new_shape()
                 self._is_new_shape = True
-            self._selected_shape.active.value = True
-            self._selected_shape.on_left_down(evt)
+            if hasattr(self._selected_shape, "on_left_down"):
+                self._selected_shape.on_left_down(evt)
```
pieleric, line 246: Please make all shapes have a `on_left_down` method, so it's always safe to call it. (It might do just nothing). It'll be faster than checking first. Same thing with the other methods below.

src/odemis/gui/comp/overlay/rectangle.py
```diff
@@ -37,16 +43,60 @@ def __init__(self, rectangle_overlay) -> None:
         self.p_point4 = rectangle_overlay.p_point4
 
 
-class RectangleOverlay(RectanglePointsSelectOverlay, EditableShape):
-    """Overlay representing one rectangle."""
+class RectangleOverlay(EditableShape, RectangleEditingMixin):
```
pieleric, line 46: Does RectangleOverlay still matches the "Overlay" interface (ie, it has all the expected attributes and methods). If not, I'd suggest to rename it to not include the word "Overlay". If yes, I'd suggest to have it still inherit from WorldOverlay, just don't call the `__init__()` if that's not helpful.

pieleric (changes requested):

src/odemis/gui/comp/overlay/shapes.py
```diff
@@ -180,39 +180,37 @@ def on_leave(self, evt):
...
-    def _get_shape(self, evt):
+    def _get_shape(self, v_pos: Tuple[float, float]):
         """
-        Find a shape corresponding to the given on_left_down event position.
+        Find a shape corresponding to the given view position.
         :return: the most appropriate shape.
             If no shape is found, it returns None.
         """
         if self._shapes:
-            pos = self.cnvs.view_to_phys(evt.Position, self.cnvs.get_half_buffer_size())
```
pieleric, line 205: Please indicate that this method is optimized for speed, and why (it's called for every mouse move?). Also indicate whether it's optimized for the "best case" (finding a shape) or "worst case" (no shape is below the given position.

src/odemis/gui/comp/overlay/rectangle.py
nandishjpatel: Yes, it still matches the overlay "Overlay" interface. As you said will not include the `__init__()`

src/odemis/gui/comp/overlay/rectangle.py
nandishjpatel: This is what you want, I will rename the function and add the docstring

src/odemis/gui/comp/overlay/shapes.py
nandishjpatel: I will mention the limitations

src/odemis/gui/comp/overlay/ellipse.py
```diff
@@ -97,47 +97,54 @@ def restore_state(self, state: EllipseState):
...
-            self.clear_drag()
-            self.selection_mode = SEL_MODE_NONE
-            self.edit_hover = None
-
-            # Set the points VA after drawing because draw() gathers the points
-            self.cnvs.update_drawing()
-            if self.selected.value:
-                self.points.value = self._points
-        WorldOverlay.on_left_up(self, evt)
+        # If the Diagonal points are not the same means the rectangle has been created
```
tepals, line 123: This code is for ellipses, is "rectangle" a copy paste error? Otherwise I don't think I understand the comment :)

Same for `rectangle_points` on line 127 and 128

src/odemis/gui/comp/overlay/shapes.py
```diff
@@ -75,14 +75,12 @@ def get_size(self) -> Tuple[float, float]:
...
+    def check_point_proximity(self, v_point: Tuple[float, float]) -> bool:
         """
-        Determine if the view point is in the shape.
+        Determine if the view point is in the proximity of the shape.
 
         :param: v_point: The point in view coordinates.
-
-        :returns: whether the view point is inside the shape or not.
-
+        :returns: whether the view point is in proximity the the shape.
```
tepals, line 83:
```
        :returns: whether the view point is in proximity of the shape.
```

src/odemis/gui/cont/tabs/fastem_main_tab.py
```diff
@@ -237,13 +237,18 @@ def _on_is_acquiring(self, mode):
...
-        """Pause the SEM stream for a specific set of tools."""
+        """Disable canvas drag and pause the SEM stream for a specific set of tools."""
         if (
             selected_tool in (TOOL_RECTANGLE, TOOL_ELLIPSE, TOOL_POLYGON)
-            and self.tab_data_model.semStream.should_update.value
         ):
-            self.tab_data_model.semStream.is_active.value = False
-            self.tab_data_model.semStream.should_update.value = False
+            # Only let ShapesOverlay handle the dragging
+            # by doing so avoid the weird situation where the canvas in a dragging state on tool change
```
tepals, line 245:
```suggestion
            # by doing so avoid the weird situation where the canvas is in a dragging state on tool change
```

src/odemis/gui/comp/overlay/rectangle.py
```diff
@@ -88,18 +138,93 @@ def restore_state(self, state: RectangleState):
         self._points = self.get_physical_sel()
         self.points.value = self._points
 
-    def is_point_in_shape(self, point):
+    def check_point_proximity(self, v_point):
+        """
+        Determine if the view point is in the proximity of the shape.
+
+        Proximity is defined as either:
+        - Near the edges of the shape.
```
tepals, line 146:
```suggestion
        - Near the edges of the shape or on the rotation knob.
```

src/odemis/gui/comp/overlay/ellipse.py
nandishjpatel: Well, this is correct. We get the rectangle points 1, 2, 3, 4 first. Then the ellipse points are drawn inside the rectangle in the `draw()` method. You see this happening in line 140 where an ellipse circumference points are drawn based on rectangles points 1, 2, 3, 4 and finally the `point VA` value is assigned

pieleric (approved):

nandishjpatel:
Can we merge this PR? I wanna re-base my vEM GUI branch

pieleric:
> Can we merge this PR? I wanna re-base my vEM GUI branch

It seems you've made the changes but pushed them as an amended commit. If you do extra changes to fix new comments, it's better to put them in a new commit, instead of amending the latest commit and rewriting the history. This helps to see only the most recent changes, and also I get notification in Github.

Also, please mark the comments as "resolved". That also hints me that something has changed!

tepals (approved):

PR #2753, opened 2024-05-07
[perf][Ticket #515524] add unittests for smaract.MS2 to check movement in one axis

K4rishma:
To test the hardware component of Optical Focus which is run by smaract.MS2, 1DOF test is added.

### DIFF
--- a/src/odemis/driver/test/smaract_test.py
+++ b/src/odemis/driver/test/smaract_test.py
@@ -873,17 +873,6 @@ class TestMCS2_1DOF(unittest.TestCase):
         # Check that at least the position changed
         self.assertNotEqual(pos_move["z"], pos_refd["z"])
 
-    def test_reference_and_deactivate_move(self):
-        # Set a deactive position and check to be sure that the controller moves to this location
-        # after a reference move
-        de_pos = {'z': REF_Z}
-        self.dev.updateMetadata({model.MD_FAV_POS_DEACTIVE: de_pos})
-
-        f = self.dev.reference(set(self.dev.axes.keys()))
-        f.result()
-
-        testing.assert_pos_almost_equal(self.dev.position.value, de_pos, **COMP_ARGS)
-
     def test_auto_update_function(self):
         """Check if position VA is updated when it is changed"""
         self.dev.moveAbs({"z": SAFE_Z}).result()
@@ -896,7 +885,6 @@ class TestMCS2_1DOF(unittest.TestCase):
         time.sleep(2)
         pos_after_move = self.dev.position.value
         testing.assert_pos_almost_equal(expected_pos, pos_after_move, **COMP_ARGS)
-
 
 
 CONFIG_Picoscale = {"name": "Stage Metrology",
--- a/src/odemis/driver/test/smaract_test.py
+++ b/src/odemis/driver/test/smaract_test.py
@@ -720,6 +720,7 @@ class TestMCS2(unittest.TestCase):
         os.remove(file_name)
 
 
+# Configuration to mostly test the Optical focus Component
 CONFIG_1DOF = {"name": "1DOF",
         "role": "stage",
         "ref_on_init": True,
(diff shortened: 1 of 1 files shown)


patrickcleeve2 (approved):

src/odemis/driver/test/smaract_test.py
```diff
@@ -720,6 +720,184 @@ def test_parse_tsv_config(self):
...
+        # after a reference move
+        de_pos = {'z': -13.0e-3}
+        self.dev.updateMetadata({model.MD_FAV_POS_DEACTIVE: de_pos})
+
+        f = self.dev.reference(set(self.dev.axes.keys()))
+        f.result()
+
+        testing.assert_pos_almost_equal(self.dev.position.value, de_pos, **COMP_ARGS)
+
+    def test_auto_update_function(self):
```
sneepstefan, line 887: Could you explain a little more what you mean with auto update function in this test?

sneepstefan (changes requested):

src/odemis/driver/test/smaract_test.py
```diff
@@ -720,6 +720,184 @@ def test_parse_tsv_config(self):
...
+        pos = {'z': -21e-3}
+        with self.assertRaises(ValueError):
+            self.dev.moveAbs(pos).result()
+
+    def test_move_abs(self):
+        pos1 = {'z': -13.0e-3}
+        pos2 = {'z': -13.01e-3}  # move by -10 um
+        pos3 = {'z': -13.02e-3}  # move by -20 um
+        pos4 = {'z': -12.09e-3}  # move by 10 um
+        pos5 = {'z': -12.08e-3}  # move by 20 um
```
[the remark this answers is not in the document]
K4rishma: I changed a comments a little bit, I will use variable instead of hard coded values. In this case it will give the reader an idea how much each position varies from the pos1 which is a safe position. It can come handy during the hardware tests.

sneepstefan (approved):

src/odemis/driver/test/smaract_test.py
```diff
@@ -720,6 +720,185 @@ def test_parse_tsv_config(self):
...
+    def test_reference_and_deactivate_move(self):
+        # Set a deactive position and check to be sure that the controller moves to this location
+        # after a reference move
+        de_pos = {'z': REF_Z}
+        self.dev.updateMetadata({model.MD_FAV_POS_DEACTIVE: de_pos})
+
+        f = self.dev.reference(set(self.dev.axes.keys()))
+        f.result()
+
+        testing.assert_pos_almost_equal(self.dev.position.value, de_pos, **COMP_ARGS)
```
pieleric, line 885: This whole test doesn't make sense. In the parameters `pos_deactive_after_ref=False`. So after referencing the actuator might be pretty much anywhere arbitrary, and that's expected. So you cannot expect here that it goes to "REF_Z". 
=> either set `pos_deactive_after_ref=True`, or remove this test.

src/odemis/driver/test/smaract_test.py
```diff
@@ -720,6 +720,185 @@ def test_parse_tsv_config(self):
...
+        pos_before_move = self.dev.position.value
+        rel_move = {"z": 5e-6}
+        expected_pos = pos_before_move.copy()
+        expected_pos["z"] += rel_move["z"]
+        self.dev.moveRel(rel_move).result()
+        # the position updater function is called every 1 sec, wait a bit more.
+        time.sleep(2)
+        pos_after_move = self.dev.position.value
+        testing.assert_pos_almost_equal(expected_pos, pos_after_move, **COMP_ARGS)
+
```
pieleric, line 899: Too many empty lines. There should be only 2.

pieleric (changes requested):

patrickcleeve2 (approved):

src/odemis/driver/test/smaract_test.py
```diff
@@ -756,7 +756,9 @@ class TestMCS2_1DOF(unittest.TestCase):
     """
     @classmethod
     def setUpClass(cls):
-        cls.dev = smaract.MCS2(**CONFIG_1DOF)
+        # cls.dev = smaract.MCS2(**CONFIG_1DOF)
+        cls.microscope = model.getMicroscope()
+        cls.dev = model.getComponent(role="focus")
```
pieleric, line 761: I don't think you can merge this as-is. I understand this is to test while the backend is running. But that's not how standard test cases are run.

src/odemis/driver/test/smaract_test.py
```diff
@@ -828,17 +830,18 @@ def test_move_cancel(self):
...
         new_pos = {'z': SAFE_Z + BIG_STEP}
-        f = self.dev.moveAbs(new_pos)
+        self.dev.moveAbs(new_pos)
         time.sleep(0.05)
         self.dev.stop()
 
         difference = new_pos['z'] - self.dev.position.value['z']
-        self.assertNotEqual(round(difference, 4), 0)
+        self.assertNotAlmostEqual(difference, 0, places=4)
+        # self.assertNotAEqual(round(difference, 4), 0)
```
pieleric, line 844: Drop. It even has a typo ("AEqual").

pieleric (changes requested):

PR #2755, opened 2024-05-13
[TRCL-213] add keysight awg 33600 series driver

FCanberk:
This is a draft PR, will be updated soon.

### DIFF
--- a/src/odemis/driver/keysight.py
+++ b/src/odemis/driver/keysight.py
@@ -1,0 +1,527 @@
+# -*- coding: utf-8 -*-
+"""
+Created on 29 Apr 2024
+
+@author: Canberk Akin
+
+Copyright © 2024 Canberk Akin, Delmic
+
+This file is part of Odemis.
+
+Odemis is free software: you can redistribute it and/or modify it under the terms
+of the GNU General Public License version 2 as published by the Free Software
+Foundation.
+
+Odemis is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+PURPOSE. See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with
+Odemis. If not, see http://www.gnu.org/licenses/.
+"""
+
+import logging
+import re
+import time
+import socket
+
+from odemis import model
+from odemis.model import HwError
+from odemis.util import to_str_escape
+from typing import List, Tuple, Dict
+
+FREQUENCY_MIN = 1e-6 # Hz. The minimum frequency is 1 microhertz according to the keysight 33600 user manual
+FREQUENCY_MAX = 100e6 # Hz. The maximum frequency is 100 megahertz according to the keysight 33600 user manual
+
+
+class TrueForm(model.Emitter):
+    def __init__(self, name, role, address,
+                 channel: int,
+                 limits: List[Tuple[float, float]],
+                 tracking: Dict[int, str] = None,
+                 **kwargs):
+        """ Initializes the Keysight 33600 series Arbitrary Waveform Generator.
+        :param name: (str) as in Odemis
+        :param role: (str) as in Odemis
+        :param address: “fake” (to start the simulator) or an IP address
+        :param channel: Channel which generate the waveform
+        :param tracking: Tracking mode: channel and tracking mode(ON, OFF or INV).
+                         Typically, channel 2 is inverted version of channel 1
+        :param limits: min/max V for each channel. if min>max, the polarity of the channel is inverted.
+        """
+        super().__init__(name, role, **kwargs)
+        # Connect to serial port
+        self._accesser = None
+
+        self._ip_address = self._findDevice(address)  # sets ._accesser
+        logging.info("Found Keysight waveform generator device on address %s", self._ip_address)
+
+        self._channel = channel
+        logging.debug("%r ,%s", channel, type(channel))
+        self._limits = limits
+        self._tracking = tracking or {}
+
+        power = False
+        self.power = model.BooleanVA(power, setter=self._set_power)
+
+        self.setWaveform(1, b"SQU")
+
+        # copy the all settings of channel 1 into channel 2 in INVerted mode
+        for c, t in self._tracking.items():
+            self.setTracking(c, t)
+
+        self._set_power(power)
+
+        dutyCycle = 0.5  # default duty cycle value is %50
+        self.dutyCycle = model.FloatContinuous(dutyCycle, range=(0.2, 0.8),
+                                               setter=self._set_duty_cycle, unit="percentage")
+
+        period = 25e-9  # s, the standard value that is used in the first system. = 40 MHz
+        self.period = model.FloatContinuous(period, range=(1 / FREQUENCY_MAX, 1 / FREQUENCY_MIN),
+                                            setter=self._set_period, unit="Hz")
+        delay = 0.0  # s
+        self.delay = model.FloatContinuous(delay, range=(0, 1000), setter=self._set_delay, unit="s")
+
+        self._set_period(period)
+        self._set_delay(delay)
+
+    def terminate(self):
+        if self._accesser:
+            self._set_power(False)
+            self._accesser.close()
+            self._accesser = None
+
+        super().terminate()
+
+    def _findDevice(self, address):
+        """
+        Look for a compatible device
+        address (None or int): the address of the Waveform Generator
+        return (serial, int): the (opened) serial port used, and the actual address
+        raises:
+            IOError: if no device are found
+        """
+        # Connection via ethernet cable
+        if not address.startswith("/"):
+            try:
+                self._accesser = IPBusAccesser(address)
+            except Exception as e:
+                logging.info("Could not establish connection to the device through IP bus accesser: %s", e)
+                raise
+
+            return address
+        else:
+            raise HwError("Failed to find a device on the address: '%s'. "
+                          "Check that the device is turned on and connected to "
+                          "the computer." % (address,))
+
+    def _sendCommand(self, cmd):
+        """
+        cmd (str): command to be sent to device
+        """
+        # Send the given command to the instrument
+        self._accesser.sendCmd(cmd)
+
+    def _sendQuery(self, q) -> str:
+        """
+        q (str): query to be sent to device
+        :return: (string) response of the query from the hardware.
+        """
+        # Send the given command to the instrument
+        logging.debug("Sending query %s", to_str_escape(q))
+        response = self._accesser.sendQuery(q)
+
+        return response
+
+    def setOutput(self, c, p: bool):
+        if p:
+            self._sendCommand(b"OUTP%d ON" % c)
+        else:
+            self._sendCommand(b"OUTP%d OFF" % c)
+
+    def setFrequency(self, c, f):
+        """
+        c (int, 1 or 2): the channel to set the frequency of.
+        f (float): the frequency value in Hertz
+        """
+        self._sendCommand(b"sour%d:freq %e" %(c, f))
+
+    def setVoltageMin(self, c, v):
+        """
+        c (int): the channel to set the low voltage. 1 or 2
+        v (float): the low voltage value
+        """
+        self._sendCommand(b"sour%d:volt:low %f" % (c, v))
+
+    def setVoltageMax(self, c, v):
+        """
+        c (int, 1 or 2): the channel to set the high voltage of.
+        v (float): the high voltage value
+        """
+        self._sendCommand(b"sour%d:volt:high %f" % (c, v))
+
+    def setVoltageLimitMin(self, c, v):
+        """
+        c (int, 1 or 2): the channel to set the minimum voltage limit of.
+        v (float): the minimum voltage limit
+        """
+        self._sendCommand(b"sour%d:volt:lim:low %f" % (c, v))
+
+    def setVoltageLimitMax(self, c, v):
+        """
+        c (int, 1 or 2): the channel to set the maximum voltage limit of.
+        v (float): the maximum voltage limit
+        """
+        self._sendCommand(b"sour%d:volt:lim:high %f" % (c, v))
+
+    def setTriggerDelay(self, c, d):
+        """
+        c (int, 1 or 2): the channel to set the delay of.
+        d (float): the trigger delay to set. between 0 and 1000 s
+        """
+        self._sendCommand(b"trig%d:del %e" % (c, d))
+
+    def setPhase(self, c: int, t: float):
+        """
+        c (1 or 2): the channel to set the delay of.
+        t: phase value in time (s). Can only be between -period and +period
+        """
+        self._sendCommand(b"sour%d:phas %e sec" % (c, t))
+        # self._sendCommand(b"phas 1e-6 SEC")  #DEBUG
+
+    def setDutyCycle(self, c, dc):
+        """
+        c (int, 1 or 2): the channel to set the duty cycle of.
+        dc (float): the duty cycle percentage. between 00.01 and 99.99
+        """
+        self._sendCommand(b"sour%d:func:squ:dcyc %.2f" % (c, dc))
+
+    def setTracking(self, c: int, t: str) -> None:
+        """
+        c (1 or 2): the channel to set the tracking mode of.
+        t (ON, OFF or INV): the tracking mode
+        """
+        if t not in ["ON", "OFF", "INV"]:
+            raise ValueError("Incorrect tracking mode: %s given" % (t,))
+        self._sendCommand(b"sour%d:trac %s" % (c, t.encode("ascii")))
+
+    def setWaveform(self, c, f):
+        """
+        c (int, 1 or 2): the channel to set the tracking mode of.
+        f (str SIN, SQU, RAMP or PULS): the wave form to generate
+        """
+        valid_waveforms = {b"SIN", b"SQU", b"RAMP", b"PULS"}
+        if f.upper() not in (waveform.upper() for waveform in valid_waveforms):
+            raise ValueError("Incorrect waveform: %s given" % (f,))
+
+        self._sendCommand(b"sour%d:appl:%s" % (c, f))
+
+    def _set_period(self, p: float) -> float:
+        f = 1/p
+        self.setFrequency(self._channel, f)
+        self.delay.value = min(self.delay.value, p)
+        return p
+
+    def _set_delay(self, d):
+        if not (0 <= d <= self.period.value):  #% (bs,)
+            raise ValueError("Delay must be < period (%s), got %s" % (self.period.value, d))
+
+        self.setPhase(self._channel, d)
+        return d
+
+    def _set_duty_cycle(self, dc):
+        duty_cycle = dc * 100
+        logging.debug("%r ,%s", duty_cycle, type(duty_cycle))
+        self.setDutyCycle(self._channel, duty_cycle)
+        return dc
+
+    def _set_power(self, p: bool) -> bool:
+        c = self._channel
+        if p:
+            lim = self._limits[c - 1]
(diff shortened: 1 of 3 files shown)


src/odemis/driver/keysight.py
```diff
@@ -0,0 +1,333 @@
...
+        self.delay = model.FloatContinuous(delay, setter=self._setTimeRange, unit="s")
+
+        dutyCycle = "DEFAULT VALUE"
+        self.dutyCycle = model.FloatContinuous(dutyCycle, setter=self.setDutyCycle, unit="s")
+
+        period = 4e-6  # default value
+        self.period = model.FloatContinuous(period, setter=self.setPeriod, unit="s")
+
+        power = "DEFAULT VALUE"
+        self.power = model.BooleanVA(power, setter=self._setTimeRange, unit="s")
```
pieleric, line 54: setter should be a dedicated function, which activates or stops the waveform generation.

src/odemis/driver/keysight.py
```diff
@@ -0,0 +1,333 @@
...
+            voltage_lim_max = float(msg.split()[1])
+            self._output_states['channel' + channel]['voltage_limit_high'] = voltage_lim_max
+        elif re.match(r"sour[1-2]:func:squ:dcyc ", msg):
+            channel = msg[4]
+            duty_cycle = float(msg.split()[1])
+            # duty cycle value is between 0 and 1
+            self._output_states['channel' + channel]['duty_cycle'] = float(duty_cycle / 100)
+
+        else:
+            logging.info("Invalid command: %s", msg)
```
pieleric, line 333: That should be higher than info, as it's a sign something is gone really unexpected. I'd say "error" the the proper level.
Do you know how the device react in such case? Does it return a specific value? Or just ignore the command? Do you have this in your logs?

src/odemis/driver/keysight.py
```diff
@@ -0,0 +1,333 @@
...
+        :param role: (str) as in Odemis
+        :param address: # “fake” (to start the simulator) or an IP address
+        :param channels: # Channels which generate the waveform
+        :param limits: # min/max V for each channel. if min>max, the polarity of the channel is inverted.
+        """
+        delay = "DEFAULT VALUE"
+        self.delay = model.FloatContinuous(delay, setter=self._setTimeRange, unit="s")
+
+        dutyCycle = "DEFAULT VALUE"
+        self.dutyCycle = model.FloatContinuous(dutyCycle, setter=self.setDutyCycle, unit="s")
```
pieleric, line 48: Indicate a `range`. Unit is `""` as it's a ratio.

src/odemis/driver/keysight.py
```diff
@@ -0,0 +1,333 @@
...
+        :param name: (str) as in Odemis
+        :param role: (str) as in Odemis
+        :param address: # “fake” (to start the simulator) or an IP address
+        :param channels: # Channels which generate the waveform
+        :param limits: # min/max V for each channel. if min>max, the polarity of the channel is inverted.
+        """
+        delay = "DEFAULT VALUE"
+        self.delay = model.FloatContinuous(delay, setter=self._setTimeRange, unit="s")
+
+        dutyCycle = "DEFAULT VALUE"
```
pieleric, line 47: Default value should be 0.5 .

src/odemis/driver/keysight.py
```diff
@@ -0,0 +1,333 @@
...
+    def __init__(self, name, role, address, channels: List[int],
+                 limits: List[Tuple[float, float]], **kwargs):
+        """ Initializes the Keysight 33600 series Arbitrary Waveform Generator.
+        :param name: (str) as in Odemis
+        :param role: (str) as in Odemis
+        :param address: # “fake” (to start the simulator) or an IP address
+        :param channels: # Channels which generate the waveform
+        :param limits: # min/max V for each channel. if min>max, the polarity of the channel is inverted.
+        """
+        delay = "DEFAULT VALUE"
```
pieleric, line 44: Default should be 0.

src/odemis/driver/keysight.py
```diff
@@ -0,0 +1,333 @@
...
+        self.delay = model.FloatContinuous(delay, setter=self._setTimeRange, unit="s")
+
+        dutyCycle = "DEFAULT VALUE"
+        self.dutyCycle = model.FloatContinuous(dutyCycle, setter=self.setDutyCycle, unit="s")
+
+        period = 4e-6  # default value
```
pieleric, line 50: "default value" as a comment is not helpful. You could instead indicate that it happens to be the standard value we use in our first system using this. You can also indicate the value in MHz.
That said, I thought we will use 40MHz by default, so that's odd.

src/odemis/driver/keysight.py
```diff
@@ -0,0 +1,333 @@
...
+        self.period = model.FloatContinuous(period, setter=self.setPeriod, unit="s")
```
pieleric, line 51: Please specify the range, based on the min/max sample frequency.

src/odemis/driver/keysight.py
```diff
@@ -0,0 +1,333 @@
...
+
+    def _sendCommand(self, cmd):
+        """
+        cmd (str): command to be sent to device
+        """
+        # Send the given command to the instrument
+        logging.debug("Sending command %s", to_str_escape(cmd))
+        self._accesser.sendCmd(cmd)  # TODO: check with real hw if a brief wait needed(like 50 ms of time.sleep)
+
+    def _sendQuery(self, q):
```
pieleric, line 90: The point of sendQuery, compared to sendCommand is that it *returns* something. => Indicate it in the signature, and actually return the value.

src/odemis/driver/keysight.py
```diff
@@ -0,0 +1,333 @@
...
+        """
+        q (str): query to be sent to device
+        """
+        # Send the given command to the instrument
+        logging.debug("Sending query %s", to_str_escape(q))
+        self._accesser.sendQuery(q)  # TODO: check with real hw if a brief wait needed(like 50 ms of time.sleep)
+
+    def setPeriod(self, c, p):
```
pieleric, line 98: Please map directly the functions to the functions supported by the device. As the device supports changing the *frequency*, it should be `setFrequency`.

src/odemis/driver/keysight.py
```diff
@@ -0,0 +1,333 @@
...
+        """
+        c (int): the channel to set the period of. 1 or 2
+        p (str): the period value
```
pieleric, line 101: str -> float.

src/odemis/driver/keysight.py
```diff
@@ -0,0 +1,333 @@
...
+        """
+        # default will be 4000000 Hz (40 MHz) = 4e-6 for period val
```
pieleric, line 103: No. 40MHz is not equivalent to 40e-6 s!

src/odemis/driver/keysight.py
```diff
@@ -0,0 +1,333 @@
...
+class TrueForm(model.Emitter):
+    def __init__(self, name, role, address, channels: List[int],
+                 limits: List[Tuple[float, float]], **kwargs):
+        """ Initializes the Keysight 33600 series Arbitrary Waveform Generator.
+        :param name: (str) as in Odemis
+        :param role: (str) as in Odemis
+        :param address: # “fake” (to start the simulator) or an IP address
+        :param channels: # Channels which generate the waveform
+        :param limits: # min/max V for each channel. if min>max, the polarity of the channel is inverted.
+        """
```
pieleric, line 43: You are still missing all the connection and initialization part. It should also check that channels and limits are valid, and report a nice error if not.

src/odemis/driver/keysight.py
```diff
@@ -0,0 +1,333 @@
...
+        self._sendCommand("sour%d:trac %f" % (c, t))
+
+
+class IPBusAccesser(object):
+    """
+    Manage TCP/IP connections over ethernet
+    """
+
+    def __init__(self, ip_addr=None, tcp_timeout=10.0, tcp_port=23):
+        """ Initialize this `TEWWAwg` instance.
```
pieleric, line 158: Wrong name

src/odemis/driver/keysight.py
```diff
@@ -0,0 +1,333 @@
...
+        """
+        c (int): the channel to set the period of. 1 or 2
+        p (str): the period value
+        """
+        self._sendCommand("sour%d:volt:lim:high %d" % (c, v))
+
+    def setDutyCycle(self, c, dc):
+        """
+        c (int): the channel to set the period of. 1 or 2
+        p (str): the period value
```
pieleric, line 140: Wrong comment

src/odemis/driver/keysight.py
```diff
@@ -0,0 +1,333 @@
...
+        """
+        self._sendCommand("sour%d:func:squ:dcyc %s" % (c, dc))
+
+    def setTracking(self, c, t):
+        """
+        c (int): the channel to set the period of. 1 or 2
+        p (str): the period value
```
pieleric, line 147: Wrong comment. -> bool?

src/odemis/driver/keysight.py
```diff
@@ -0,0 +1,333 @@
...
+        if ip_addr is not None:
+            self._ip_addr = ip_addr
+        if tcp_timeout is not None:
+            self._tcp_timeout = float(tcp_timeout)
+
+        # Open TCP-IP Socket:
+        self._tcp_sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM, socket.IPPROTO_TCP)
+        self._tcp_sock.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
+        self._tcp_sock.settimeout(self._tcp_timeout)
+        self._tcp_sock.connect((self._ip_addr, self._tcp_port))
```
pieleric, line 192: You have to handle the case when the IP address cannot be found. Either here, or in the caller. In such case, it should raise a `HwError` to tell the Odemis back-end that a non-fatal error happened (and pass a user-friendly message asking to check the device is connected and turned on).

src/odemis/driver/keysight.py
```diff
@@ -0,0 +1,333 @@
...
+    def sendCmd(self, cmd_str, wait_for_complete=True):
+        # Send the given command to the instrument.
+        if self._tcp_sock is not None:  # pass if not connected
+            cmd_str = cmd_str.rstrip()
+            cmd_str += '\n'
+            cmd_byte = cmd_str.encode()
+            self._tcp_sock.sendall(cmd_byte)  # send command
+
+
+class Keysight33622ASimulator(object):
```
pieleric, line 230: Drop the `(object)` (not needed with Python 3).

src/odemis/driver/keysight.py
```diff
@@ -0,0 +1,333 @@
...
+        self._input_buf = ""  # what we receive from the "host computer"
+
+        self._output_states = {
+            'channel1': {
+                'tracking': 'on',  # on, off, invert
+                'voltage_low': -5.0,  # V
+                'voltage_high': 5.0,  # V
+                'voltage_limit_low': -5.0,  # V
+                'voltage_limit_high': 5.0,  # V
+                'period': 150,  # 1/frequency in Hz (1 μHz to 50 MHz)
```
pieleric, line 247: Please store the frequency, as the device does.

src/odemis/driver/keysight.py
```diff
@@ -0,0 +1,333 @@
...
+            self._output_states['channel' + channel]['voltage_limit_low'] = voltage_lim_min
+        elif re.match(r"sour[1-2]:volt:lim:low ", msg):
+            channel = msg[4]
+            voltage_lim_max = float(msg.split()[1])
+            self._output_states['channel' + channel]['voltage_limit_high'] = voltage_lim_max
+        elif re.match(r"sour[1-2]:func:squ:dcyc ", msg):
+            channel = msg[4]
+            duty_cycle = float(msg.split()[1])
+            # duty cycle value is between 0 and 1
+            self._output_states['channel' + channel]['duty_cycle'] = float(duty_cycle / 100)
```
pieleric, line 330: From the manual, the device accepts values between 0.01% and 99.99%. So it should check the range here.
Also, no need to convert the duty cycle to a float. You can store it as-is (in percentage). That's closer to what the real device does.

src/odemis/driver/keysight.py
```diff
@@ -0,0 +1,333 @@
...
+        """
+        msg (str): the message to parse
+        return None: self._output_buf is updated if necessary
+        """
+        logging.debug("SIM: parsing %s", to_str_escape(msg))
+        msg = msg.decode("latin1").strip()  # remove leading and trailing whitespace
+        msg = "".join(msg.split())  # remove all space characters
+
+        if msg == "*IDN?":
+            self._sendAnswer("Agilent Technologies,33622A,MY59003187,A.02.03-3.15-03-64-0")
```
pieleric, line 297: Please return a string which clearly indicate it's a simulator. For instance, the serial number can be "FAKE".

src/odemis/driver/keysight.py
```diff
@@ -0,0 +1,333 @@
...
+            'channel1': {
+                'tracking': 'on',  # on, off, invert
+                'voltage_low': -5.0,  # V
+                'voltage_high': 5.0,  # V
+                'voltage_limit_low': -5.0,  # V
+                'voltage_limit_high': 5.0,  # V
+                'period': 150,  # 1/frequency in Hz (1 μHz to 50 MHz)
+                'duty_cycle': 50  # percentage
+            },
+            'channel2': {
```
pieleric, line 250: As all the channels have the same attributes, I think it would make more sense to create a dedicated class (or use namedtuple, if you like it better)

src/odemis/driver/keysight.py
```diff
@@ -0,0 +1,333 @@
...
+                'tracking': 'on',  # on, off, invert
+                'voltage_low': -5.0,  # V
+                'voltage_high': 5.0,  # V
+                'voltage_offset': 0.0,  # V
+                'voltage_limit_low': -5.0,  # V
+                'voltage_limit_high': 5.0,  # V
+                'period': 150,  # 1/frequency in Hz (1 μHz to 50 MHz)
+                'duty_cycle': 75  # percentage
```
pieleric, line 258: The device default is 50, please use the default.

src/odemis/driver/keysight.py
```diff
@@ -0,0 +1,333 @@
...
+    """
+
+    def __init__(self, timeout=1):
+        self.timeout = timeout
+        self._output_buf = ""  # what the commands sends back to the "host computer"
+        self._input_buf = ""  # what we receive from the "host computer"
+
+        self._output_states = {
+            'channel1': {
+                'tracking': 'on',  # on, off, invert
```
pieleric, line 242: Still missing the "phase" (used for the delay).

pieleric (changes requested):

src/odemis/driver/keysight.py
FCanberk: It does not return a specific value and times out eventually(so returns timeout error).

Additionally on the hardware screen it shows: `"Error generated by remote interface"` error message.

src/odemis/driver/keysight.py
FCanberk: Will update the comment and make it informative and smarter. 

The frequency is in Hertz. So the default value of the period is indeed missing a 0.

src/odemis/driver/keysight.py
FCanberk: the tracking mode can be ON, OFF or INV so should still be string but the `p` param in the docstring needs an update obviously.

src/odemis/driver/keysight.py
FCanberk: I wanted to make it as close to the real hardware as possible so I've placed the exact response from the hardware. 

But it makes sense to indicate that the response is coming from the simulator. Will update the string then!

src/odemis/driver/keysight.py
```diff
@@ -87,75 +105,111 @@ def _sendCommand(self, cmd):
...
-        self._sendCommand("sour%d:trac %f" % (c, t))
+        self._sendCommand(b"sour%d:appl:%s" % (c, f))
+
+    def _set_period(self, p: float) -> float:
+        f = 1/p
+        self.setFrequency(self._channel, f)
+        return p
+
+    def _set_delay(self, d):
+        self.setDelay(self._channel, d)
```
pieleric, line 194: What we call "delay" is what is called "phase" in the Keysight AWG. See the specification: "[The delay needs] to be converted to phase by as a ratio of 360 compared to  period. Note: because the channels are locked, the waveforms of both channels are always starting at the same time. The phase only changes the delay between the channels output and the “sync out” signal."

src/odemis/driver/test/keysight_test.py
```diff
@@ -0,0 +1,124 @@
...
+logging.basicConfig(format="%(asctime)s  %(levelname)-7s %(module)s:%(lineno)d %(message)s")
+
+# Export TEST_NOHW=1 to force using only the simulator and skipping test cases
+# needing real hardware
+TEST_NOHW = (os.environ.get("TEST_NOHW", "0") != "0")  # Default to HW testing
+
+# arguments used for the creation of basic components to test via IP connection
+CONFIG_AWG = {
+    "name": "SIMULATED Keysight Arbitrary Waveform Generator model:33622A,"
+            "FAKE_Serial_Number: AB12345678,",
```
pieleric, line 42: The name  should just be "Keysight AWG"

src/odemis/driver/test/keysight_test.py
```diff
@@ -0,0 +1,124 @@
...
+    "role": "blanker",
+    "address": "192.168.5.11",
+    "channel": 1,
+    "tracking": [2, "INV"],
```
pieleric, line 46: Dict

pieleric (changes requested):

src/odemis/driver/keysight.py
```diff
@@ -0,0 +1,496 @@
...
+            self._sendAnswer("SIMULATED Keysight Arbitrary Waveform Generator,33622A,"
+                             "FAKE_Serial_Number: AB12345678")
+        elif re.match(r"sour[1-2]:freq ", msg):
+            channel = msg[4]
+            frequency = float(msg.split()[1])
+            if frequency < 1.0E-6 or frequency > 1.0E6:
+                logging.warning("The frequency value is out of bounds: %s given", frequency)
+            else:
+                self._output_states['channel' + channel].frequency = frequency
+        elif re.match(r"sour[1-2]:trac ", msg):
```
pieleric, line 460: Please run your test cases with the simulator again. There are some (new) commands that the simulator doesn't support. You need to handle them too.

src/odemis/driver/keysight.py
```diff
@@ -0,0 +1,496 @@
...
+        if t not in ["ON", "OFF", "INV"]:
+            raise ValueError("Incorrect tracking mode: %s given" % (t,))
+        self._sendCommand(b"sour%d:trac %s" % (c, t.encode("ascii")))
+
+    def setWaveform(self, c, f):
+        """
+        c (int, 1 or 2): the channel to set the tracking mode of.
+        f (str SIN, SQU, RAMP or PULS): the wave form to generate
+        """
+        self._sendCommand(b"sour%d:appl:%s" % (c, f))
```
pieleric, line 213: Here would be good to check for the input value, to check it's only within the acceptable values.

src/odemis/driver/keysight.py
```diff
@@ -0,0 +1,496 @@
...
+
+    def _set_period(self, p: float) -> float:
+        f = 1/p
+        self.setFrequency(self._channel, f)
+        self.delay.value = min(self.delay.value, p)
+        return p
+
+    def _set_delay(self, d):
+        if not (0 <= d <= self.period.value):
+            raise ValueError("Delay must be < period (%s), got %s", self.period.value, d)
```
pieleric, line 223: See the GitHub checks, you cannot use arguments in exceptions, you should use %

pieleric (changes requested):

PR #2757, opened 2024-05-16
[FIX][METEOR-1090] Prevent Enzel / Mimas from accessing correlation tab

patrickcleeve2:


src/odemis/gui/cont/tabs/localization_tab.py
```diff
@@ -318,7 +318,7 @@ def load_overview_data(self, data):
             chamber_tab.load_overview_streams(streams)
 
         # sync overview streams with correlation tab
-        if len(streams) > 0:
+        if len(streams) > 0 and self.main_data.role == "meteor":
```
pieleric, line 321: FYI, you could have also been more explicit by writting `"meteor-correlation" in self.main_data.tab.choices`, but this is fine as-is.

pieleric (approved):

sneepstefan (approved):

PR #2759, opened 2024-05-16
[fix] correct image shape index in _setTranslation in simcam.py

sneepstefan:
This fix addresses the wrong index numbers of self._img.shape used in the _setTranslation function which causes the translation to be calculated wrong. The index of self._img.shape should be YXC so always select X and Y for the translation.

### DIFF
--- a/src/odemis/driver/simcam.py
+++ b/src/odemis/driver/simcam.py
@@ -207,8 +207,8 @@ class Camera(model.DigitalCamera):
         """
         # compute the min/max of the shift. It's the same as the margin between
         # the centered ROI and the border, taking into account the binning.
-        max_tran = ((self._img.shape[-1] - self._resolution[0] * self._binning[0]) // 2,
-                    (self._img.shape[-2] - self._resolution[1] * self._binning[1]) // 2)
+        max_tran = ((self._img_res[0] - self._resolution[0] * self._binning[0]) // 2,
+                    (self._img_res[1] - self._resolution[1] * self._binning[1]) // 2)
 
         # between -margin and +margin
         trans = (max(-max_tran[0], min(value[0], max_tran[0])),


src/odemis/driver/simcam.py
```diff
@@ -207,8 +207,8 @@ def _setTranslation(self, value):
         """
         # compute the min/max of the shift. It's the same as the margin between
         # the centered ROI and the border, taking into account the binning.
-        max_tran = ((self._img.shape[-1] - self._resolution[0] * self._binning[0]) // 2,
-                    (self._img.shape[-2] - self._resolution[1] * self._binning[1]) // 2)
+        max_tran = ((self._img.shape[-2] - self._resolution[0] * self._binning[0]) // 2,
+                    (self._img.shape[-3] - self._resolution[1] * self._binning[1]) // 2)
```
tepals, line 211: Why not use the following?
```suggestion
        max_tran = ((self._img.shape[0] - self._resolution[0] * self._binning[0]) // 2,
                    (self._img.shape[1] - self._resolution[1] * self._binning[1]) // 2)
```

sneepstefan: I was not sure if there can be more dimensions than only YXC, e.g. ZYXC.
I could change it, but then it will be like this right?
```suggestion
        max_tran = ((self._img.shape[1] - self._resolution[0] * self._binning[0]) // 2,
                    (self._img.shape[0] - self._resolution[1] * self._binning[1]) // 2)
```
Since X is on [1] and Y is on [0] index.

tepals: That sounds like a good reason.

tepals: So then just leave it like it was

tepals (approved):

src/odemis/driver/simcam.py
pieleric, line 211: I don't think any of this works. The problem is that when the image is RGB, the shape is YXC, while if it's greyscale, the shape is YX. I don't think the code supports passing images with more data (the init code would fail in this case).
The init code already defines a `self._img_res`, which should be the right info (X = img_res[0] and Y =img_res[1]). => Use that attribute.

pieleric (changes requested):

FCanberk (approved):

pieleric (approved):

PR #2763, opened 2024-05-16
[fix] calibration button label text and colour

nandishjpatel:
image

In Ubuntu 22.04 it looked like this

Now

image

### DIFF
--- a/src/odemis/gui/main_xrc.py
+++ b/src/odemis/gui/main_xrc.py
@@ -7622,9 +7622,7 @@ D\x02\x12\x0c/\x81\x10.\xc4\xcc\xb0\x8f\xa1\x9e\xa1\x81a/\x90\x05\x06\x8d\
                             <object class="sizeritem">
                               <object class="wxStaticText" name="lbl_optical_autofocus">
                                 <label>Optical Autofocus</label>
-                                <font>
-                                  <sysfont>wxSYS_DEFAULT_GUI_FONT</sysfont>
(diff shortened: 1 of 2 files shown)


tepals:
Does is still look good in ubuntu 20.04?

src/odemis/gui/cont/tabs/fastem_overview_tab.py
```diff
@@ -65,6 +66,9 @@ def __init__(self, name, button, panel, main_frame, main_data, vp, view_controll
         self.tab_data.semStream = sem_stream
 
         # Buttons of the calibration panel
+        panel.lbl_optical_autofocus.SetForegroundColour(FG_COLOUR_MAIN)
+        panel.lbl_sem_autofocus.SetForegroundColour(FG_COLOUR_MAIN)
+        panel.lbl_autobrightness_contrast.SetForegroundColour(FG_COLOUR_MAIN)
```
pieleric, line 71: Looking at the XRC file, it seems that quite a few widgets just have a `<fg>#DDDDDD</fg>` (which is the FG_COLOUR_MAIN). I think we should do the same for these 3 StaticText, instead of changing their colour in the code.

pieleric (changes requested):

nandishjpatel:
> Does is still look good in ubuntu 20.04?

Yes it does

pieleric (approved):

PR #2754, opened 2024-05-13
[FEAT][METEOR-960] Add support for FIBSEM control

patrickcleeve2:
Adds support for controlling FIBSEM, includes:
- Autoscript driver and configuration file
- Fibsem UI Tab
- SEM and FIB image acquisition
- Milling controls 

I'm opening smaller pull requests for each component, but this is the lot

PR #2767, opened 2024-05-22
[FEAT]: convert shapes list to va

patrickcleeve2:
Convert the _shapes list to a VA, allowing it to be subscribed to for when shapes are removed. Used for milling patterns

### DIFF
--- a/src/odemis/gui/comp/overlay/shapes.py
+++ b/src/odemis/gui/comp/overlay/shapes.py
@@ -113,7 +113,7 @@ class ShapesOverlay(WorldOverlay):
         self.new_shape = model.VigilantAttribute(None, readonly=True)
         self._selected_shape = None
         self._shape_to_copy = None
-        self._shapes = []
+        self.shapes = model.ListVA()
         if tool and tool_va:
             self.tool = tool
             tool_va.subscribe(self._on_tool, init=True)
@@ -121,13 +121,13 @@ class ShapesOverlay(WorldOverlay):
     def clear(self):
         """Remove all shapes and update canvas."""
         self._activate_shapes(False)
-        self._shapes.clear()
+        self.shapes.value.clear()
 
     def remove_shape(self, shape):
         """Remove the shape and update canvas."""
-        if shape in self._shapes:
+        if shape in self.shapes.value:
             shape.active.value = False
-            self._shapes.remove(shape)
+            self.shapes.value.remove(shape)
             self.cnvs.remove_world_overlay(shape)
             self.cnvs.request_drawing_update()
 
@@ -145,7 +145,7 @@ class ShapesOverlay(WorldOverlay):
 
     def _activate_shapes(self, flag=True):
         """Activate or de-activate the shapes."""
-        for shape in self._shapes:
+        for shape in self.shapes.value:
             shape.active.value = flag
 
     def _on_tool(self, selected_tool):
@@ -164,9 +164,9 @@ class ShapesOverlay(WorldOverlay):
         :return: the most appropriate shape.
             If no shape is found, it returns None.
         """
-        if self._shapes:
+        if self.shapes.value:
             pos = self.cnvs.view_to_phys(evt.Position, self.cnvs.get_half_buffer_size())
-            for shape in self._shapes[::-1]:
+            for shape in self.shapes.value[::-1]:
                 if shape.is_point_in_shape(pos):
                     return shape
         return None
@@ -174,7 +174,7 @@ class ShapesOverlay(WorldOverlay):
     def _create_new_shape(self):
         """Create a new shape."""
         shape = self.shape_cls(self.cnvs)
-        self._shapes.append(shape)
+        self.shapes.value.append(shape)
         self.cnvs.add_world_overlay(shape)
         self.new_shape._set_value(shape, force_write=True)
         return shape
@@ -183,7 +183,7 @@ class ShapesOverlay(WorldOverlay):
         """Copy a selected shape to a view position as the center."""
         # Copy the shape
         shape = self._shape_to_copy.copy()
-        self._shapes.append(shape)
+        self.shapes.value.append(shape)
         self.cnvs.add_world_overlay(shape)
         self.new_shape._set_value(shape, force_write=True)
         # Move the copied shape to a view position
@@ -242,7 +242,7 @@ class ShapesOverlay(WorldOverlay):
 
     def draw(self, ctx, shift=(0, 0), scale=1.0, dash=False):
         """Draw all the shapes."""
-        for shape in self._shapes:
+        for shape in self.shapes.value:
             shape.draw(
                 ctx,
                 shift,


pieleric (commented):
You're only providing one "side" of the changes. It shows how the current code is changed, but not what you use these changes for. So it's a bit hard to be review how helpful this change is.
That said, I would not recommend changing a private attribute to a VA in order to be able to observe it, because it's "private". You could make it "public". However, in this case, I think it would make more sense to have the VA on the model (ie, one of the classes in gui.model), and pass this VA to the overlay, at creation. Then the overlay can update/observe the VA based on user inputs. I think it would make sense to also have "selected_shape" also a VA on the model.

patrickcleeve2:
> You're only providing one "side" of the changes. It shows how the current code is changed, but not what you use these changes for. So it's a bit hard to be review how helpful this change is. That said, I would not recommend changing a private attribute to a VA in order to be able to observe it, because it's "private". You could make it "public". However, in this case, I think it would make more sense to have the VA on the model (ie, one of the classes in gui.model), and pass this VA to the overlay, at creation. Then the overlay can update/observe the VA based on user inputs. I think it would make sense to also have "selected_shape" also a VA on the model.

Context for change is in: https://github.com/delmic/odemis/pull/2754. Basically, there is currently no-way of knowing when shapes are removed by the user. There is already a specific VA for new_shape, so you can know when they are created. 

I agree this can probably be 'public'. I don't think it makes sense to move the va to the models, the shapes are properties of the overlays, and only ui elements.  They are only used for display of some other internal state (patterns for meteor, roa for fastem).

nandishjpatel:
> > You're only providing one "side" of the changes. It shows how the current code is changed, but not what you use these changes for. So it's a bit hard to be review how helpful this change is. That said, I would not recommend changing a private attribute to a VA in order to be able to observe it, because it's "private". You could make it "public". However, in this case, I think it would make more sense to have the VA on the model (ie, one of the classes in gui.model), and pass this VA to the overlay, at creation. Then the overlay can update/observe the VA based on user inputs. I think it would make sense to also have "selected_shape" also a VA on the model.
> 
> Context for change is in: #2754. Basically, there is currently no-way of knowing when shapes are removed by the user. There is already a specific VA for new_shape, so you can know when they are created.
> 
> I agree this can probably be 'public'. I don't think it makes sense to move the va to the models, the shapes are properties of the overlays, and only ui elements. They are only used for display of some other internal state (patterns for meteor, roa for fastem).

I also need the removed functionality in the future :) Using the ListVA do you use some kind of set comparison to find the removed shape @patrickcleeve2 ?

In one of my PR's I am gonna use a named tuple [ShapeState](https://github.com/nandishjpatel/odemis/blob/c13c3e013b352b702cfc46afb78c10f38af57c04/src/odemis/gui/comp/overlay/shapes.py#L36). Do you think this can be the `new_shape` VA value (most probably rename the VA). This named tuple has attributes `is_new`, `is_deleted` . What do you think?

patrickcleeve2:
> > > You're only providing one "side" of the changes. It shows how the current code is changed, but not what you use these changes for. So it's a bit hard to be review how helpful this change is. That said, I would not recommend changing a private attribute to a VA in order to be able to observe it, because it's "private". You could make it "public". However, in this case, I think it would make more sense to have the VA on the model (ie, one of the classes in gui.model), and pass this VA to the overlay, at creation. Then the overlay can update/observe the VA based on user inputs. I think it would make sense to also have "selected_shape" also a VA on the model.
> > 
> > 
> > Context for change is in: #2754. Basically, there is currently no-way of knowing when shapes are removed by the user. There is already a specific VA for new_shape, so you can know when they are created.
> > I agree this can probably be 'public'. I don't think it makes sense to move the va to the models, the shapes are properties of the overlays, and only ui elements. They are only used for display of some other internal state (patterns for meteor, roa for fastem).
> 
> I also need the removed functionality in the future :) Using the ListVA do you use some kind of set comparison to find the removed shape @patrickcleeve2 ?
> 
> In one of my PR's I am gonna use a named tuple [ShapeState](https://github.com/nandishjpatel/odemis/blob/c13c3e013b352b702cfc46afb78c10f38af57c04/src/odemis/gui/comp/overlay/shapes.py#L36). Do you think this can be the `new_shape` VA value (most probably rename the VA). This named tuple has attributes `is_new`, `is_deleted` . What do you think?

@nandishjpatel Yep, I check the list of patterns and find which has been removed. 

I'm having trouble visualising how that va would be used. Does it trigger an event every time you create / delete the shape, and then you implement a check of is_new or is_deleted? Would you be able to show an example?

nandishjpatel:
> > > > You're only providing one "side" of the changes. It shows how the current code is changed, but not what you use these changes for. So it's a bit hard to be review how helpful this change is. That said, I would not recommend changing a private attribute to a VA in order to be able to observe it, because it's "private". You could make it "public". However, in this case, I think it would make more sense to have the VA on the model (ie, one of the classes in gui.model), and pass this VA to the overlay, at creation. Then the overlay can update/observe the VA based on user inputs. I think it would make sense to also have "selected_shape" also a VA on the model.
> > > 
> > > 
> > > Context for change is in: #2754. Basically, there is currently no-way of knowing when shapes are removed by the user. There is already a specific VA for new_shape, so you can know when they are created.
> > > I agree this can probably be 'public'. I don't think it makes sense to move the va to the models, the shapes are properties of the overlays, and only ui elements. They are only used for display of some other internal state (patterns for meteor, roa for fastem).
> > 
> > 
> > I also need the removed functionality in the future :) Using the ListVA do you use some kind of set comparison to find the removed shape @patrickcleeve2 ?
> > In one of my PR's I am gonna use a named tuple [ShapeState](https://github.com/nandishjpatel/odemis/blob/c13c3e013b352b702cfc46afb78c10f38af57c04/src/odemis/gui/comp/overlay/shapes.py#L36). Do you think this can be the `new_shape` VA value (most probably rename the VA). This named tuple has attributes `is_new`, `is_deleted` . What do you think?
> 
> @nandishjpatel Yep, I check the list of patterns and find which has been removed.
> 
> I'm having trouble visualising how that va would be used. Does it trigger an event every time you create / delete the shape, and then you implement a check of is_new or is_deleted? Would you be able to show an example?

Here is an example which you can try:

```
from collections import namedtuple
from enum import Enum

from odemis import model

# Named tuple for elements stored in undo and redo stacks
ShapeState = namedtuple("ShapeState", ["shape", "state", "action"])
# Enum class to store ShapeState's action
Action = Enum("Action", ["EDIT", "CREATE", "DELETE"])

def on_new_shape(shape: ShapeState):
    print(f"{shape.shape} {shape.state} {shape.action}")

new_shape = model.VigilantAttribute(None, readonly=True)
new_shape.subscribe(on_new_shape)

shape = "Rectangle"

new_shape._set_value(ShapeState(shape=shape, state=1, action=Action.CREATE), force_write=True)
new_shape._set_value(ShapeState(shape=shape, state=1, action=Action.EDIT), force_write=True)
new_shape._set_value(ShapeState(shape=shape, state=1, action=Action.DELETE), force_write=True)
new_shape._set_value(ShapeState(shape=shape, state=2, action=Action.EDIT), force_write=True)
```

patrickcleeve2:
> > Context for change is in: #2754. Basically, there is currently no-way of knowing when shapes are removed by the user. There is already a specific VA for new_shape, so you can know when they are created.
> > I agree this can probably be 'public'. I don't think it makes sense to move the va to the models, the shapes are properties of the overlays, and only ui elements. They are only used for display of some other internal state (patterns for meteor, roa for fastem).
> 
> VAs are good for representing states. Using them to indicating events (or changes of the state) is tricky. Especially, in theory, the VA API doesn't promise that every change of value is reported to all subscribers. If the values changes too fast, there _might_ be some intermediary values not reported to the subscribers. In the context of the GUI, for now, I think every value is always reported so that's more just a "long term risk".

But since you can subscribe to the 'state', you get an event when that VA 'state' changes. So for this case, you effectively get an event every time a shape is added/removed. Is that different?

> In my opinion, the shapes that the user draw in the GUI (either for the FASTEM RoAs or the METEOR milling patterns) are part of the GUI state. So _ideally_ the information of the shape should be on the model, as VA(s), and the Overlays would listen/update the VAs according to the user interaction.

I don't really know which model you're referring to, can you give an example? I'm not sure about fastem but for milling patterns, they are also on the milling model (https://github.com/delmic/odemis/blob/57a23d36e323a178c70d45739d60cfe68fbe8a52/src/odemis/gui/comp/milling.py), because they can be changes from shape updates or text/data entries. You'd still need the overlay / shapes themselves to send state updates when they get deleted or added?  

> That's not to say you cannot use VAs in other cases. It's totally fine to use VAs outside of the model, and use them as "communication way" between objects.
> 
> To conclude, I'm fine with making this VA public. However, please also evaluate whether all the information shown by the GUI is in the model. It might be the better answer... I cannot really say until I get a better understanding of the "user" code.

nandishjpatel:
@patrickcleeve2 do you need this feature change right now? I will be finishing my VolumeEM GUI development in a month from now. Can this wait for a while? Also if this change go into master, there will be some merge conflicts.

Given the discussion so far, I think its best to the `shapes` ListVA in the model. I also foresee this suggestion from @pieleric  in my ongoing work.

patrickcleeve2:
Yeah it can wait for a while. In the meantime, I'll just make a different version of the overlay for milling, we can then pick this up when you're done.

PR #2768, opened 2024-05-22
[FEAT][METEOR-960] add autoscript driver

patrickcleeve2:
Add autoscript driver and configuration file

### DIFF
--- a/install/linux/usr/share/odemis/sim/meteor-fibsem-sim.odm.yaml
+++ b/install/linux/usr/share/odemis/sim/meteor-fibsem-sim.odm.yaml
@@ -1,0 +1,250 @@
+# To start the xtadapter in autoscript simulator mode:
+# cd path/to/xtlib-adapter/xtadapter
+# python server_sim_autoscript.py sim
+# the server will be started, and listening on localhost:4242
+
+METEOR-Sim: {
+    class: Microscope,
+    role: meteor,
+    children: [
+        "Meteor Stage",
+        "Optical Objective",
+        "Optical Focus",
+        "Light Source",
+        "Filter Wheel",
+        "Camera",
+        "FIBSEM"
+    ],
+}
+
+"FIBSEM": {
+  class: autoscript_client.SEM,
+  role: fibsem,
+  init: {
+    # address: "192.168.6.5",  # real
+    address: "localhost",    # simulator
+    port: "4242"
+  },
+  children: {
+    stage: "Stage",
+    sem-scanner: "Electron-Beam",
+    sem-detector: "Electron-Detector",
+    sem-focus: "Electron-Focus",
+    fib-scanner: "Ion-Beam",
+    fib-detector: "Ion-Detector",
+    fib-focus: "Ion-Focus",
+  }
+}
+
+
+"Electron-Beam": {
+    role: e-beam,
+    init: {
+        hfw_nomag: 0.25, # hfw when mag == 100x
+    }
+}
+
+"Electron-Detector": {
+    role: se-detector,
+    init: {}
+
+}
+
+"Electron-Focus": {
+    role: ebeam-focus,
+    init: {},
+    affects: ["Electron-Beam"]
+}
+
+"Ion-Beam": {
+    role: "ion-beam",
+    init: {
+        hfw_nomag: 0.25, # hfw when mag == 100x
+    }
+}
+
+"Ion-Detector": {
+    role: "se-detector-ion",
+    init: {}
+}
+
+"Ion-Focus": {
+    role: ion-focus,
+    init: {},
+    affects: ["Ion-Beam"]
+}
+# Normally provided by the SEM
+"Stage": {
+    role: stage-bare,
+    init: {
+    },
+    metadata: {
+        # Loading position:
+        FAV_POS_DEACTIVE: { 'rx': 0, 'rz': 1.9076449, 'x': -0.01529, 'y': 0.0506, 'z': 0.01975 },
+        # XYZ ranges for SEM & METEOR
+        SEM_IMAGING_RANGE: {"x": [-10.e-3, 10.e-3], "y": [-5.e-3, 10.e-3], "z": [-0.5e-3, 8.e-3]},
+        FM_IMAGING_RANGE: {"x": [0.040, 0.054], "y": [-5.e-3, 10.e-3], "z": [-0.5e-3, 8.e-3]},
+        # Grid centers in SEM range
+        # Adjusted so that at init (0,0,0), it's at the Grid 1.
+        SAMPLE_CENTERS: {"GRID 1": {'x': 0, 'y': 2.46e-3, 'z': 0}, "GRID 2": {'x': 2.98e-3, 'y': 2.46e-3, 'z': 0}},
+        # Mirroring values between SEM - METEOR
+        POS_COR: [0.02647, -0.000017],
+        # Active tilting (rx) & rotation (rz) angles positions when switching between SEM & FM, in radians.
+        # Note: these values are calibrated at installation time.
+        FAV_FM_POS_ACTIVE: {"rx": 0.12213888553625313  , "rz":  5.06145}, # 7° - 270°
+        # Typically rz = 110°, but we make it 0 so that at init it looks like in SEM position
+        FAV_SEM_POS_ACTIVE: {"rx": 0, "rz": 0}    # Note that milling angle (rx) can be changed per session 
+    },
+}
+
+"Linked YZ": {
+    class: actuator.ConvertStage,
+    role: null,
+    dependencies: {
+        "under": "Stage"
+    },
+    init: {
+      axes: [ "y", "z" ], # name of the axes in the dependency, mapped to x,y (if identity transformation)
+      rotation: -0.8034946943806713, # rad , -45°
+    },
+}
+
+"Meteor Stage": {
+    class: actuator.MultiplexActuator,
+    role: stage,
+    dependencies: { "x": "Stage", "y": "Linked YZ", "z": "Linked YZ", },
+    init: {
+        axes_map: { "x": "x", "y": "x", "z": "y",},
+    },
+    affects: ["Camera", "EBeam"],
+    metadata: {
+        # Typically, x range is the same as FM_IMAGING_RANGE, and Y has to be converted
+        # POS_ACTIVE_RANGE: {"x": [0.040, 0.054], "y": [-10.e-3, 20.e-3]}
+        POS_ACTIVE_RANGE: {"x":[37.e-3, 59.e-3], "y": [-15.e-3, 6.e-3]}
+    },
+}
+
+"Light Source": {
+    class: omicronxx.HubxX,
+    role: light,
+    init: {
+        port: "/dev/fakehub", # Simulator
+        #port: "/dev/ttyFTDI*",
+        },
+    affects: ["Camera"],
+}
+
+"Optical Objective": {
+    class: static.OpticalLens,
+    role: lens,
+    init: {
+        mag: 84.0, # ratio, (actually of the complete light path)
+        na: 0.85, # ratio, numerical aperture
+        ri: 1.0, # ratio, refractive index
+    },
+    affects: ["Camera"]
+}
+
+# Normally a IDS uEye or Zyla
+# Axes: X is horizontal on screen (going left->right), physical: far->close when looking at the door
+#       Y is vertical on screen (going bottom->top), physical: left->right when looking at the door
+"Camera": {
+    class: simcam.Camera,
+    role: ccd,
+    dependencies: {focus: "Optical Focus"},
+    init: {
+        image: "andorcam2-fake-clara.tiff",
+        transp: [-1, 2], # To swap/invert axes
+    },
+    metadata: {
+        # To change what the "good" focus position is on the simulator
+        # It's needed for not using the initial value, which is at deactive position.
+         FAV_POS_ACTIVE: {'z': 1.7e-3},  # good focus position
+         ROTATION: -0.099484,  # [rad] (=-5.7°)
+    },
+}
+
+# Controller for the filter-wheel
+# DIP must be configured with address 7 (= 1110000)
+"Optical Actuators": {
+    class: tmcm.TMCLController,
+    role: null,
+    init: {
+        port: "/dev/fake6", # Simulator
+        address: null, # Simulator
+        axes: ["fw"],
+        ustepsize: [1.227184e-6], # [rad/µstep]
+        rng: [[-14, 7]], # rad, more than 0->2 Pi, in order to allow one extra rotation in both direction, when quickly switching
+        unit: ["rad"],
+        refproc: "Standard",
+        refswitch: {"fw": 0}, #digital output used to switch on/off sensor
+        inverted: ["fw"], # for the filter wheel, the direction doesn't matter, as long as the positions are correct
+    },
+}
+
+"AntiBacklash for Filter Wheel": {
+    class: actuator.AntiBacklashActuator,
+    role: null,
+    init: {
+        backlash: {
+            # Force every move to always finish in the same direction
+            "fw": 50.e-3,  # rad
+        },
+    },
+    children: {"slave": "Optical Actuators"},
+}
+
+"Filter Wheel": {
+    class: actuator.FixedPositionsActuator,
+    role: filter,
+    dependencies: {"band": "AntiBacklash for Filter Wheel"},
+    init: {
+        axis_name: "fw",
+        # This filter-wheel is made so that the light goes through two "holes":
+        # the filter, and the opposite hole (left empty). So although it has 8
+        # holes, it only supports 4 filters (from 0° to 135°), and there is no
+        # "fast-path" between the last filter and the first one.
+        positions: {
+             # pos (rad) -> m,m
+             0.08: [414.e-9, 450.e-9], # FF01-432/36
+             0.865398: [500.e-9, 530.e-9], # FF01-515/30
+             1.650796: [579.5e-9, 610.5e-9], # FF01-595/31
+             2.4361944: [663.e-9, 733.e-9], # FF02-698/70
+        },
+        cycle: 6.283185, # position of ref switch (0) after a full turn
+    },
+    # TODO: a way to indicate the best filter to use during alignement and brightfield? via some metadata?
+    affects: ["Camera"],
+}
+
+
+# CLS3252dsc-1
+"Optical Focus": {
+    class: smaract.MCS2,
+    role: focus,
+    init: {
+        locator: "fake",
+        ref_on_init: True,
+        # TODO: check speed/accel
+        speed: 0.003,  # m/s
+        accel: 0.003,  # m/s²
+        #hold_time: 5 # s, default = infinite
+        # TODO: check the ranges, and the channel
+        axes: {
+            'z': {
+                # -11.5mm is safely parked (FAV_POS_DEACTIVE)
+                # 1.7mm is typically in focus (FAV_POS_ACTIVE)
+                range: [-15.e-3, 5.e-3],
+                unit: 'm',
+                channel: 0,
+            },
+        },
+    },
+    metadata: {
+        # Loading position to retract lens
+        FAV_POS_DEACTIVE: {'z': -11.5e-3},
+        # Initial active position (close from the sample, but not too close, for safety)
+        FAV_POS_ACTIVE: {'z': 1.69e-3}
+    },
+    affects: ["Camera"],
+}
--- a/src/odemis/driver/autoscript_client.py
+++ b/src/odemis/driver/autoscript_client.py
@@ -1,0 +1,2134 @@
+# -*- coding: utf-8 -*-
+"""
+Created on 20 Feb 2024
+
+@author: Patrick Cleeve
+
+Copyright © 2024 Delmic
+
+This file is part of Odemis.
+
+Delmic Acquisition Software is free software: you can redistribute it and/or modify it under the terms of the GNU
+General Public License as published by the Free Software Foundation, either version 2 of the License, or (at your
+option) any later version.
+
+Delmic Acquisition Software is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even
+the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+more details.
+
+You should have received a copy of the GNU General Public License along with Delmic Acquisition Software. If not, see
+http://www.gnu.org/licenses/.
+"""
+import logging
+import math
+import queue
+import threading
+import time
+from concurrent.futures import CancelledError
+from typing import Optional, Union, Dict, List, Tuple
+
+import msgpack # only used for debug information
+import msgpack_numpy
+import numpy
+import Pyro5.api
+import pkg_resources
+from Pyro5.errors import CommunicationError
+
+from odemis import model
+from odemis import util
+from odemis.model import (CancellableFuture, CancellableThreadPoolExecutor,
+                          DataArray, HwError, ProgressiveFuture,
+                          StringEnumerated, isasync)
+from odemis.driver.xt_client import check_and_transfer_latest_package
+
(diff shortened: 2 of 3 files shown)


src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2304 @@
...
+        if "rz" not in rng:
+            rng["rz"] = stage_info["range"]["R"]
+
+        axes_def = {
+            "x": model.Axis(unit=stage_info["unit"]["x"], range=rng["x"]),
+            "y": model.Axis(unit=stage_info["unit"]["y"], range=rng["y"]),
+            "z": model.Axis(unit=stage_info["unit"]["z"], range=rng["z"]),
+            "rx": model.Axis(unit=stage_info["unit"]["T"], range=rng["rx"]),
+            "rz": model.Axis(unit=stage_info["unit"]["R"], range=rng["rz"]),
+        } # TODO: make these axis and unit arguments consistent
```
pieleric, line 1945: What is this TODO about? All axes should be in m/rad. If not, they should be converted between autoscript units and SI.

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2304 @@
...
+        f._channel_name = detector._scanner.channel
+        return self._executor.submitf(f, self._applyAutofocus, f)
+
+    def _applyAutofocus(self, future):
+        """
+        Starts autofocussing and checks if the autofocussing process is finished for ProgressiveFuture.
+        :param future (Future): the future to start running.
+        """
+        pass
+        # channel_name = future._channel_name
```
pieleric, line 2195: If that is not supported, just remove entirely AutoFocus. Otherwise, make it work.

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2304 @@
...
+        try:
+            self._updatePosition()
+        except Exception:
+            logging.exception("Unexpected failure when updating position")
+
+    def _getPosition(self):
+        """Get position and translate the axes names to be Odemis compatible."""
+        pos = self.parent.get_stage_position()
+        pos["rx"] = pos.pop("t")
+        pos["rz"] = pos.pop("r")
```
pieleric, line 1986: uppercase or lowercase t & r? In the init it's uppercase... It might be worthy to add a comment explaining it's important.

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2304 @@
...
+        """Returns: (dict) the unit and range of the spotsize. Unit is None means the spotsize is unitless."""
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            return self.server.spotsize_info(channel)
+
+    def set_dwell_time(self, dwell_time: float, channel: str) -> None:
+        """
+        :param dwell_time: (float) the dwell time in seconds.
+        :param channel: (str) Name of the channel to set the dwell time for.
+
```
pieleric, line 583: drop the line

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2304 @@
...
+        if "stage" in children:
+            ckwargs = children["stage"]
+            self._stage = Stage(parent=self, daemon=daemon, **ckwargs)
+            self.children.value.add(self._stage)
+
+        # # create the chamber child, if requested
+        # if "chamber" in children:
+        #     ckwargs = children["chamber"]
+        #     self._chamber = Chamber(parent=self, daemon=daemon, **ckwargs)
+        #     self.children.value.add(self._chamber)
```
pieleric, line 292: If not supported, drop the code (including the class).

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2304 @@
...
+        # Check if at least one of the required scanner types is instantiated
+        scanner_types = ["scanner", "fib-scanner"]  # All allowed scanners types
+        if not any(scanner_type in children for scanner_type in scanner_types):
+            raise KeyError("FIBSEM was not given any scanner as child. "
+                           "One of 'scanner', 'fib-scanner' or 'mb-scanner' need to be included as child")
+
+        has_detector = "detector" in children
+        has_fib_detector = "fib-detector" in children
+
+        if "scanner" in children: # TODO: rename to sem-scanner
```
pieleric, line 272: Please do the TODO.

pieleric (changes requested):
I didnt' look too much in the details, but tried to point to the main parts.

src/odemis/driver/autoscript_client.py
patrickcleeve2: api returns lowercase r, t for stage position, and uppercase for axis limits. I've changed it so the server will now return lowercase for all

src/odemis/driver/autoscript_client.py
patrickcleeve2: changed all components to sem-* or fib-*

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2304 @@
...
+    (2048, 1768),
+    (4096, 3536),
+    (768, 512),
+    (1536, 1024),
+    (3072, 2048),
+    (6144, 4096),
+)
+
+# Xtadapter debian package installation directory which contains xtadapter's zip files
+XT_INSTALL_DIR = "/usr/share/xtadapter"
```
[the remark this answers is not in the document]
patrickcleeve2: I assume it will be different, but havent implemented that yet on server side

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2220 @@
...
+
+
+class Package(object):
```
nandishjpatel, line 75: This class and `check_latest_package` can be imported from [here in xt_client](https://github.com/delmic/odemis/blob/2419a9a7ec2f2e4ae309cdbc0d83c529bc958f0b/src/odemis/driver/xt_client.py#L82)

Is there a common place or can you recommend a place where these can go @pieleric ?

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2220 @@
...
+    To use this driver the autoscript adapter developed by Delmic should be running on the TFS PC. Communication to the
+    Microscope server is done via Pyro5.
+    """
+
+    def __init__(self, name, role, children, address, port: str ='4242', daemon=None,
+                 **kwargs):
+        """
+        Parameters
+        ----------
+        address: str
```
nandishjpatel, line 223: docstring needs to be updated

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2220 @@
...
+        if not any(scanner_type in children for scanner_type in scanner_types):
+            raise KeyError("FIBSEM was not given any scanner as child. "
+                           "One of 'sem-scanner', 'fib-scanner' need to be included as child")
+
+        has_sem_detector = "sem-detector" in children
+        has_fib_detector = "fib-detector" in children
+
+        if "sem-scanner" in children:
+            kwargs = children["sem-scanner"]
+            self._scanner = Scanner(parent=self, daemon=daemon, channel="electron", has_detector=has_sem_detector, **kwargs)
```
nandishjpatel, line 264:
```suggestion
            self._scanner = Scanner(parent=self, daemon=daemon, channel="electron", has_detector=True, **kwargs)
```

and line 259 can go?

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2220 @@
...
+            self.children.value.add(self._scanner)
+
+        if "fib-scanner" in children:
+            kwargs = children["fib-scanner"]
+            self._fib_scanner = Scanner(parent=self, daemon=daemon, channel="ion", has_detector=has_fib_detector, **kwargs)
```
nandishjpatel, line 269:
```suggestion
            self._fib_scanner = Scanner(parent=self, daemon=daemon, channel="ion", has_detector=True, **kwargs)
```

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2220 @@
...
+            https://pyro5.readthedocs.io/en/latest/tipstricks.html#binary-data-transfer-file-transfer
+
+        :param data: The package's zip file data in bytes.
+
+        """
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            return self.server.transfer_latest_package(data)
+
+    def check_and_transfer_latest_package(self) -> None:
```
nandishjpatel, line 315: I see this is not used

src/odemis/driver/autoscript_client.py
nandishjpatel: Ahh I see it is not used

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2220 @@
...
+
+    def pressure_info(self):
+        """Returns (dict): the unit and range of the pressure."""
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            return self.server.pressure_info()
+
+##### BEAM CONTROL
+
+    def set_scan_mode(self, mode: str, channel: str, value: Union[float, dict] = None) -> None:
```
nandishjpatel, line 499:
```suggestion
    def set_scan_mode(self, mode: str, channel: str, value: Union[float, dict]) -> None:
```

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2220 @@
...
+            detector_mode,
+            choices=set(self.parent.detector_mode_info(self.channel)["choices"]),
+            setter=self._setDetectorMode)
+
+        self._genmsg = queue.Queue()  # GEN_*
+        self._generator = None
+
+        # Refresh regularly the values, from the hardware, starting from now
+        self._updateSettings()
+        # self._va_poll = util.RepeatingTimer(5, self._updateSettings, "Settings polling detector")
```
nandishjpatel, line 1571: Does this need to be uncommented?

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2220 @@
...
+                        "to install it.".format(package.version))
+                    update.set_urgency(notify2.URGENCY_NORMAL)
+                    update.set_timeout(10000)    # 10 seconds
+                    update.show()
+                else:
+                    logging.warning("{} is a bad file in {} not transferring latest package.".format(ret, package.path))
+        except Exception as err:
+            logging.exception(err)
+
+    def list_available_channels(self) -> list:
```
tepals, line 358: is this necessary for autoscript? If I remember correctly this was xtlib specific

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2220 @@
...
+msgpack_file: {msgpack.__file__}
+
+This is likely a dependencies issue:
+make sure you have the following matching versions of msgpack and msgpack-numpy installed:
+msgpack==0.5.6, msgpack-numpy==0.4.4
+or
+msgpack==1.0.3 msgpack-numpy==0.4.8
+"""
+
+class FIBSEM(model.HwComponent):
```
tepals, line 211: Why not just name it `SEM`, this is not specific for FIBSEMs or is it? At least I think it would be nice to have autoscript support for regular SEMs as well

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2220 @@
...
+        """
+        Parameters
+        ----------
+        blank (bool or None): True if the the electron beam should blank, False if it should be unblanked,
+            None if it should be blanked/unblanked automatically.
+
+        Returns
+        -------
+        (bool or None): True if the the beam is blanked, False if it is unblanked. See Notes for edge case,
+            None if it should be blanked/unblanked automatically.
```
tepals, line 1447: The beam blanking is a bit annoying. If you have trouble with it during testing it might be good to have look at this https://drive.google.com/file/d/1f5xdaA16qMMYpCGvDacCrIpUKAhRsu3Y/view (I can also explain the diagram if necessary)

tepals (changes requested):
I did not have a chance to look at everything yet, but here are my comments so far

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2220 @@
...
+                        n_pixels = self._scanner.resolution.value[0] * self._scanner.resolution.value[1]
+                        est_acq_time = self._scanner.dwellTime.value * n_pixels
+                    else:
+                        # Acquisition time is unknown => assume it will be long
+                        est_acq_time = 5 * 60  # 5 minutes
+
+                    # Retrieve the image (scans image, blocks until the image is received)
+                    image = self.parent.acquire_image(self._scanner.channel)
+                    # time.sleep(est_acq_time) # simulate acquisition time # simulator only
+                    # FIXME: BUG: double scanning? why?
```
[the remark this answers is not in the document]
patrickcleeve2: I believe it is the same issue yeah, will test the fix thanks

src/odemis/driver/autoscript_client.py
patrickcleeve2: yeah autoscript has a similar channels concept for setting the beam/detectors

src/odemis/driver/autoscript_client.py
patrickcleeve2: Not every scan mode takes a value, so changing it to optional arg

src/odemis/driver/autoscript_client.py
patrickcleeve2: This causes the tfs ui to change every 5 seconds which causes problems so its disabled for now, see Note below.

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2220 @@
...
+
+    def __init__(self, name, role, children, address, port: str ='4242', daemon=None,
+                 **kwargs):
+        """
+        Parameters
+        ----------
+        address: str
+            server ip address for the microscope server (sim address is localhost)
+        port: str
+            server port of the Microscope server, default is '4242'
```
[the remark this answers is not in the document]
patrickcleeve2: I don't think you want to run them at the same time as they conflict with each other. So i think having them on the same port will cause an error if you try to run both, which I think is good rather than silently running both

src/odemis/driver/autoscript_client.py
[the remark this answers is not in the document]
patrickcleeve2: @tepals this seems to have fixed it out of the box, thanks

src/odemis/driver/autoscript_client.py
patrickcleeve2: Changed to SEM to be consistent with other driver, added a comment explaining what is supported

src/odemis/driver/autoscript_client.py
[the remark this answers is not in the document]
patrickcleeve2: This will be the same adapter as xt_client, so will be re-using the code. Configuration is done on the server side to specify which adapter should be run

src/odemis/driver/autoscript_client.py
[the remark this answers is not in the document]
tepals: great! Glad I remembered that fix :)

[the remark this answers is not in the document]
patrickcleeve2: tested on hw too, all working

nandishjpatel (approved):

src/odemis/driver/autoscript_client.py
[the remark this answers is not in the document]
pieleric: Ok. So the xtadapter is shared, and the client (in odemis.driver) is different. I'd suggest to share the code related to xtadapter update. I'd say the simplest might be to move all the related functions & contansts in xt_client independent from the SEM class, and then import them in autoscript_client. I *think* it's OK that autoscipt_client  imports xt_client (to used these functions). Otherwise you'd need to create a third module, to be shared.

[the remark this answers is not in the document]
patrickcleeve2: Made everything common between the two clients

src/odemis/driver/autoscript_client.py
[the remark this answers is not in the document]
tepals: Okay, then it would probably be nice to have a check here and in the xt_client to make sure that you are connected to the correct adapter. Then you can have a bit nicer error reporting (in the block below) in case the xtadapter is running while you need the autoscript adapter or vice versa.

```
        try:
            self.server = Pyro5.api.Proxy(f"PYRO:Microscope@{address}:{port}")
            self.server._pyroTimeout = 30  # seconds
            self._swVersion = self.server.get_software_version()
            self._hwVersion = self.server.get_hardware_version()
            logging.debug(
                f"Successfully connected to autoscript server with software version {self._swVersion} and hardware"
                f"version {self._hwVersion}")
        except CommunicationError as err:
            raise HwError("Failed to connect to autoscript server '%s'. Check that the "
                          "uri is correct and autoscript server is"
                          " connected to the network. %s %s" % (address, err, debug_connection_info))
        except OSError as err:
            raise HwError("XT server reported error: %s." % (err,))
```

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2122 @@
...
+                future._was_stopped = True
+                # Update the position, even if the move didn't entirely succeed
+                self._updatePosition()
+
+    def _doMoveRel(self, future, shift):
+        """
+        shift (dict): position in internal coordinates (ie, axes in the same
+           direction as the hardware expects)
+        """
+        # We don't check the target position fit the range, the xt-adapter will take care of that
```
tepals, line 1914:
```suggestion
        # We don't check the target position fit the range, the autoscript-adapter will take care of that
```

tepals (changes requested):

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2122 @@
...
+        self._genmsg = queue.Queue()  # GEN_*
+        self._generator = None
+
+        # Refresh regularly the values, from the hardware, starting from now
+        self._updateSettings()
+        # self._va_poll = util.RepeatingTimer(5, self._updateSettings, "Settings polling detector")
+        # self._va_poll.start()
+        # Note: using the repeated polling causes the view in xtUI to be changed every five seconds,
+        # this makes it pretty annoying to do anything on that side, and error prone.
+        # disabling this until a better solution is found
```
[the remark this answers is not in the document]
patrickcleeve2: I think if we do that it would be confusing, because user might 'select' the SEM view in odemis or just leave it on that view, then start using the xtui and it would automatically switch back which would be confusing/annoying. It would also happen if user leaves odemis running which would be even more confusing.

In practice, this only updates detector brightness, contrast, mode and type. mode and type are usually just set once, and not changed. And brightness / contrast change when the user changes them or runs auto-brightness-contrast, so we could just update them directly there? We can see how required the auto-updates are after that

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2122 @@
...
+        Turn the beam on or off.
+        :param state: (bool) True to turn the beam on and False to turn the beam off.
+        :param channel: (str) Name of the channel to turn the beam on or off for.
+        :return None
+        """
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            self.server.turn_beam_on(state, channel)
+
+    def beam_is_on(self, channel: str):
```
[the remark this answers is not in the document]
patrickcleeve2: Would it be worth updating xt-client side then?

install/linux/usr/share/odemis/sim/meteor-fibsem-sim.odm.yaml
```diff
@@ -0,0 +1,251 @@
+# For now, this is esssentially just a FM microscope, as the FIB part is handled
+# separately by the SEM
+
+METEOR-Sim: {
+    class: Microscope,
+    role: meteor,
+    children: [
+        # "Stage",
```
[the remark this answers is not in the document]
patrickcleeve2: Stage is now defined under FIBSEM as a child of that instead of the METEOR as a whole.

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2122 @@
...
+            https://pyro5.readthedocs.io/en/latest/tipstricks.html#binary-data-transfer-file-transfer
+
+        :param data: The package's zip file data in bytes.
+
+        """
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            return self.server.transfer_latest_package(data)
+
+    def check_and_transfer_latest_package(self) -> None:
```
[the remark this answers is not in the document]
patrickcleeve2: Done, the reference to xtadapter are because the entire package is called xtadapter, and the autoscript adapter is a sub-part of that

src/odemis/driver/autoscript_client.py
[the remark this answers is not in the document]
tepals: sounds good

tepals (approved):

pieleric (changes requested):

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2121 @@
...
+                        # Acquisition time is unknown => assume it will be long
+                        est_acq_time = 5 * 60  # 5 minutes
+
+                    # HACK: from xt_client to prevent double scanning
+                    if self._acq_should_stop(timeout=0.2):
+                        logging.debug("Image acquisition should stop, exiting loop")
+                        break
+
+                    # Retrieve the image (scans image, blocks until the image is received)
+                    image = self.parent.acquire_image(self._scanner.channel)
```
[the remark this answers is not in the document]
patrickcleeve2: The current acquisition is blocking so it waits until the end of the scan. I need hw access to test the non-blocking acquisition

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2121 @@
...
+            ckwargs = children["sem-detector"]
+            self._detector = Detector(parent=self, daemon=daemon, channel="electron", **ckwargs)
+            self.children.value.add(self._detector)
+
+        if "fib-detector" in children:
+            ckwargs = children["fib-detector"]
+            self._fib_detector = Detector(parent=self, daemon=daemon, channel="ion", **ckwargs)
+            self.children.value.add(self._fib_detector)
+
+    def transfer_latest_package(self, data: bytes) -> None:
```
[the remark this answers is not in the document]
patrickcleeve2: refactor to shared functional version in xt_client. Not enabled for xt_client until tested

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2121 @@
...
+
+        contrast_info = self.parent.contrast_info(self.channel)
+        self.contrast = model.FloatContinuous(
+            self.parent.get_contrast(self.channel),
+            contrast_info["range"],
+            unit=contrast_info["unit"],
+            setter=self._setContrast)
+
+        detector_type = self.parent.get_detector_type(self.channel)
+        self.detector_type = model.StringEnumerated(
```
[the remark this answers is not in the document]
patrickcleeve2: The names detector type and detector mode are what is used in the TFS software by users, so I think it should match here. I've shortened it to just .type and .mode, and added a comment for typical values

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2121 @@
...
+            return self.server.pressure_info()
+
+##### BEAM CONTROL
+
+    def set_scan_mode(self, mode: str, channel: str, value: Optional[Union[float, dict]] = None) -> None:
+        """
+        Set the scan mode.
+        :param mode: (str) Name of desired scan mode, one of: unknown, external, full_frame, spot, or line.
+        :param channel: (str) Name of the channel to set the scan mode for.
+        :param value: (float or dict) Value of the scan mode.
```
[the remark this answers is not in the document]
patrickcleeve2: Each different scan mode requires a different value type (or None), so this is just an overload of all the options. I've added individual functions for each mode with the respective types. get_scan_mode can only return the scan mode, not the value so its just a string.

patrickcleeve2:
Closing, outdated

PR #2769, opened 2024-05-22
 [performance] increased FoV from 1.5mm to 2mm for overview mode in the scanner config

nandishjpatel:


pieleric (approved):
Please note that the title is too long. It should just be a short line, and have an explanation in the second part of the commit message.

tepals (approved):

PR #2776, opened 2024-05-23
[fix] driver andorcam2: improve behaviour with HW trigger and high frame rate

pieleric:
This fixes and improve the andorcam2 driver (with the Andor Newton) for cases when the hardware trigger is used, and for cases when frame rates > 100 FPS happen. (They are linked, as we use the hardware trigger to reach high frame rates).

### DIFF
--- a/src/odemis/driver/andorcam2.py
+++ b/src/odemis/driver/andorcam2.py
@@ -595,6 +595,22 @@ class AndorCam2(model.DigitalCamera):
 
             self._hw_trigger_invert = hw_trigger_invert
 
+            # Queue to control the acquisition thread
+            self._genmsg = queue.Queue()  # GEN_* or float
+            # Queue of all synchronization events received (typically max len 1)
+            # This is in case the client sends multiple triggers before one image
+            # is received.
+            self._old_triggers = []
+            self._synchronized = False  # True if the acquisition must wait for an Event trigger
+            self._num_no_gc = 0  # how many times the garbage collector was skipped
+
+            # For temporary stopping the acquisition (kludge for the andorshrk
+            # SR303i which cannot communicate during acquisition)
+            self.hw_lock = threading.RLock()  # to be held during DRV_ACQUIRING (or shrk communicating)
+            # append None to request for a temporary stop acquisition. Like an
+            # atomic counter, but Python has no atomic counter and lists are atomic.
+            self.request_hw = []
+
             # setup everything best (fixed)
             # image (6 ints), exposure, readout, gain, shutter, synchronized
             self._prev_settings = [None, None, None, None, None, None]
@@ -647,7 +663,6 @@ class AndorCam2(model.DigitalCamera):
                                                  (self._transposeSizeToUser(min_res),
                                                   self._transposeSizeToUser(resolution)),
                                                  setter=self._setResolution)
-            self._setResolution(self._transposeSizeToUser(resolution))
 
             maxbin = self.GetMaximumBinnings(AndorV2DLL.RM_IMAGE)
             self.binning = model.ResolutionVA(self._transposeSizeToUser(self._binning),
@@ -665,9 +680,7 @@ class AndorCam2(model.DigitalCamera):
 
             # The total duration of a frame acquisition (IOW, the time between two frames).
             # It's a little longer than the exposure time. (Corresponds to "accumulate" on the Andor 2 SDK).
-            # WARNING: for now it's only updated when the camera is acquiring.
-            # TODO: We probably could do better by having the acquisition thread still
-            # updating the camera settings when not acquiring (ie, idle).
+            # WARNING: during acquisition, it's only updated at the start of the next frame.
             self.frameDuration = model.FloatVA(self._exposure_time, unit="s", readonly=True)
 
             # To control the acquisition thread behaviour when several new frames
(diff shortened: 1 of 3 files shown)


src/odemis/driver/andorcam2.py
```diff
@@ -1884,6 +1888,25 @@ def _getMaxBPP(self):
...
+        if self.hw_lock.acquire(timeout=0.1):
+            try:
+                if self.GetStatus() != AndorV2DLL.DRV_ACQUIRING:
+                    self._update_settings()
+            finally:
+                self.hw_lock.release()
+        # TODO: might still need to update the settings at the end of the acquisition, in case
+        # the VAs were just changed before ending the acquisition?
+        # TODO: or should we just try to guess? (ie, duplicate the behaviour of GetAcquisitionTimings())
+        # We could also try to cache the result of GetAcquisitionTimings() for such cases.
```
tepals, line 1954: what would be the advantage of guessing?

pieleric: It would provide some feedback to the user about which value is acceptable. Still, as it's a guess, we could pick a wrong one, which would be quite annoying, as the behaviour would be different whether the camera is playing or not. => Let's not do that. I've updated the comment.

tepals (approved):

src/odemis/driver/andorcam2.py
```diff
@@ -3226,6 +3335,31 @@ def AbortAcquisition(self):
         self.status = AndorV2DLL.DRV_IDLE
         self.acq_aborted.set()
 
+    def GetNumberNewImages(self, p_first, p_last):
```
sneepstefan, line 3338: Docstring for arguments at least?

src/odemis/driver/andorcam2.py
```diff
@@ -3226,6 +3335,31 @@ def AbortAcquisition(self):
...
+        last = _deref(p_last, c_uint32)
+
+        # Just pretend there is no image ready yet. For now, the driver always calls WaitForAcquisition()
+        # in such case, so this works.
+        if self._first_frame == self._last_frame:
+            raise AndorV2Error(20024, "No new data")  # DRV_NO_NEW_DATA
+        first.value = self._first_frame
+        last.value = self._last_frame - 1
+
+    def GetImages16(self, first, last, cbuffer, size, p_validfirst, p_validlast):
```
sneepstefan, line 3349: Docstring for arguments at least?

sneepstefan (changes requested):

src/odemis/driver/andorcam2.py
```diff
@@ -3127,6 +3225,10 @@ def SetTriggerInvert(self, inverted):
         # We don't really have a hardware trigger, so no need to really invert it
         pass
 
+    def SetFastExtTrigger(self, fast):
```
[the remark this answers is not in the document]
pieleric: Well, that's the Andor2 SDK API. I'd rather keep the names of the arguments, even if the simulator doesn't care about it. It can be useful just to understand how the function should be used.

src/odemis/driver/andorcam2.py
pieleric: This simulates the Andor2 SDK C API. The python types are quite messy (like `POINTER(c_uint32)`), and don't really match the API documentation, so I'd rather not use type annotation. I could add the official C types, like `long*` .

src/odemis/driver/andorcam2.py
pieleric: Put the C types

src/odemis/driver/andorcam2.py
```diff
@@ -3183,11 +3287,14 @@ def _WaitForAcquisition(self, timeout=None):
...
-        # elif self._trigger_mode == AndorV2DLL.TM_EXTERNAL:
-        #     # Pretend it never comes in
-        #     time.sleep(timeout)
-        #     raise AndorV2Error(20024, "No new data, simulated acquisition aborted")
+        elif self._trigger_mode == AndorV2DLL.TM_EXTERNAL:
+            # Pretends it immediately arrived
+            pass
+            # time.sleep(timeout)
+            # Pretend it never comes in
+            # raise AndorV2Error(20024, "No new data, simulated acquisition aborted")
```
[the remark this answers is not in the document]
pieleric: This is useful for simulating errors. I've adjusted the comment to clarify this.

src/odemis/driver/andorcam2.py
```diff
@@ -1930,6 +1999,10 @@ def _configure_trigger_mode(self, synchronized):
             logging.debug("Using HW trigger for acquisition")
             self.atcore.SetAcquisitionMode(AndorV2DLL.AM_VIDEO)
             self.atcore.SetTriggerMode(AndorV2DLL.TM_EXTERNAL)
+            # FastExtTrigger allows to have a shorter frame duration, but at the expense of
+            # potentially incorrect data, because the clean cycle might not have occurred. It also
+            # causes the reported frame duration to be much shorter. For our case, it doesn't help.
+            # self.atcore.SetFastExtTrigger(0)  # Disable
```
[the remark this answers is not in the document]
pieleric: Yes, indeed, that's the default, so no need to call that function every time. I've updated the comment to clarify.

sneepstefan (approved):

PR #2778, opened 2024-05-24
[refactor] Use acquisition config for FAST-EM spot estimation thresho…

tepals:
…ld, not hard coded value.

The value used for the treshold of the algorithm used in estimate_grid_orientation_from_img was hardcoded everywhere. During development and for very thick samples an expert user or developer might want to change this, this was for instance necessary in the PSI testing. With this refactor makes it possible to update the value in the acquisition.ini file, instead of hardcoding it.


tepals:
@pieleric I implemented it the way you suggested. But if I use this on the fastem-calibrations side the bitbucket CI raises an error, because with the import of the `AcquisitionConfig` wx also gets imported.
I can try to install wx in the bitbucket CI, but it feels like a code smell that wx needs to be imported to read out a value from a file. Do you have any other suggestions?

https://bitbucket.org/delmic/fastem-calibrations/pipelines/results/280

src/odemis/acq/fastem.py
```diff
@@ -33,6 +33,7 @@
 
 from odemis.acq.align.fastem import align, estimate_calibration_time
 from odemis.gui import FG_COLOUR_WARNING
+from odemis.gui.conf.file import AcquisitionConfig
```
pieleric, line 36: Ideally, nothing from odemis.acq depends on odemis.gui. So we might want to even reconsider the `FG_COLOUR_WARNING`.

pieleric:
I had no idea that there were that many places calling `estimate_grid_orientation_from_img`. Indeed, I don't think all the code should depend on odemis.gui. Typically, we want odemis.gui to be "at the top" of the dependency tree: it depends on a lot of other code, but nothing else depends on it. 
Maybe we could make `odemis.gui.conf.file` an exception and allow to import it from other places. But then it shouldn't  depend on `wx` (it should be relatively easy to fix as it seems to only indirectly use it due to its dependency on gui.util). 
Alternatively, we could use a separate code to read that configuration value when in `fastem-calibration` (as it's read-only, using ConfigParser would be super minimal anyway). 
Or you'd find a way for the GUI code to "pass" this configuration option to the calibration.
In any case, I still wouldn't recommend to rely on metadata in the microscope file...

src/odemis/acq/fastem.py
nandishjpatel: It seems to be used here: https://github.com/delmic/odemis/blob/2419a9a7ec2f2e4ae309cdbc0d83c529bc958f0b/src/odemis/gui/cont/project.py#L434

tepals: Since it is only used as a default value, I think it should be fine to update the default of the input argument to `"#FFA300"`. Do you agree @nandishjpatel? I can update this in a separate PR
```
def __init__(self, name, coordinates, colour="#FFA300"):
```

tepals:
@pieleric I ended up going for the last option "Or you'd find a way for the GUI code to "pass" this configuration option to the calibration.". If it then ends up that this is something that the users need to be able to set it through the GUI, this also becomes easier to implement.

tepals:
related to: https://bitbucket.org/delmic/fastem-calibrations/pull-requests/150

src/odemis/acq/fastem.py
nandishjpatel: > Since it is only used as a default value, I think it should be fine to update the default of the input argument to `"#FFA300"`. Do you agree @nandishjpatel? I can update this in a separate PR
> 
> ```
> def __init__(self, name, coordinates, colour="#FFA300"):
> ```

That's fine

nandishjpatel (approved):

PR #2779, opened 2024-05-24
[refactor] Retry FAST-EM pre-calibrations 3 times in case of failure.

tepals:
The pre-calibrations can fail due to some dirt on the sample, therefore retry it 3 times in a slightly different location, before failing completely.

### DIFF
--- a/src/odemis/acq/fastem.py
+++ b/src/odemis/acq/fastem.py
@@ -697,28 +697,37 @@ class AcquisitionTask(object):
         fi = numpy.array(self._roa.field_indices)
         # col, row => row 0 is the top of the ROA and the lowest column value is the most left field
         min_col = numpy.min(fi[fi[:, 1] == 0], axis=0)[0]
-        self.field_idx = (min_col - 1, 0)  # one to the left of the top-left field
 
         logging.debug("Start pre-calibration.")
-        pos_hor, pos_vert = self.get_abs_stage_movement()  # get the absolute position for the new tile
-
-        logging.debug(f"Moving to stage position x: {pos_hor}, y: {pos_vert}")
-        self._stage_scan.moveAbsSync({'x': pos_hor, 'y': pos_vert})
-
-        self._pre_calibrations_future = align(self._scanner, self._multibeam,
-                                              self._descanner, self._detector,
-                                              self._stage, self._ccd,
-                                              self._beamshift, None,  # no need for the detector rotator
-                                              self._se_detector, self._ebeam_focus,
-                                              calibrations=pre_calibrations)
-
         try:
-            self._pre_calibrations_future.result()  # wait for the calibrations to be finished
-        except CancelledError:
-            logging.debug("Cancelled acquisition pre-calibrations.")
-            raise
+            for i in range(3):  # try running the pre-calibrations 3 times
+                # Move 1/10th of a field to the top right
+                self.field_idx = (min_col - 1 + 0.1 * i, - i * 0.1)
+                pos_hor, pos_vert = self.get_abs_stage_movement()  # get the absolute position for the new tile
+                logging.debug(f"Moving to stage position x: {pos_hor}, y: {pos_vert}")
+                self._stage_scan.moveAbsSync({'x': pos_hor, 'y': pos_vert})
+                logging.debug(f"Will run pre-calibrations at field index {self.field_idx}")
+                try:
+                    self._pre_calibrations_future = align(self._scanner, self._multibeam,
+                                                          self._descanner, self._detector,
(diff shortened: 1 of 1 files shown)


src/odemis/acq/fastem.py
```diff
@@ -712,13 +712,35 @@ def pre_calibrate(self, pre_calibrations):
...
 
+        logging.debug(f"Will run pre-calibrations at field index {self.field_idx}")
         try:
-            self._pre_calibrations_future.result()  # wait for the calibrations to be finished
-        except CancelledError:
-            logging.debug("Cancelled acquisition pre-calibrations.")
-            raise
+            for i in range(3):  # try running the pre-calibrations 3 times
+                try:
+                    self._pre_calibrations_future = align(self._scanner, self._multibeam,
```
K4rishma, line 719: It is set in line 708, do we redefine it?

K4rishma (approved):

src/odemis/acq/fastem.py
nandishjpatel: I second this

src/odemis/acq/fastem.py
```diff
@@ -712,13 +712,35 @@ def pre_calibrate(self, pre_calibrations):
                                               self._se_detector, self._ebeam_focus,
                                               calibrations=pre_calibrations)
 
+        logging.debug(f"Will run pre-calibrations at field index {self.field_idx}")
```
nandishjpatel, line 715: This can go inside the for loop

src/odemis/acq/fastem.py
tepals: yes, actually I need to remove it from line 708, because it will not re-run the future if I just call `.result()` on the old future.

nandishjpatel (approved):

K4rishma (approved):

src/odemis/acq/fastem.py
```diff
@@ -705,20 +705,35 @@ def pre_calibrate(self, pre_calibrations):
...
+                except Exception as err:
+                    logging.warning(f"Pre-calibration failed for ROA {self._roa.name.value} with error {err}, "
+                                    f"will try again.")
+                    if i == 2:
+                        raise ValueError(f"Pre-calibrations failed 3 times, with error {err}")
+                    # Move 1/10th of a field to the top right
+                    self.field_idx = (self.field_idx[0] + 0.1, self.field_idx[1] - 0.1)
+                    pos_hor, pos_vert = self.get_abs_stage_movement()  # get the absolute position for the new tile
+                    logging.debug(f"Moving to stage position x: {pos_hor}, y: {pos_vert}")
+                    self._stage_scan.moveAbsSync({'x': pos_hor, 'y': pos_vert})
```
pieleric, line 732: Note that you could avoid duplication of this code (with the one just before the for loop) by placing this version just after the `for`:
```suggestion
                    # Move 1/10th of a field to the top right
                    self.field_idx = ((min_col - 1 + 0.1 * i,  - i * 0.1)
                    pos_hor, pos_vert = self.get_abs_stage_movement()  # get the absolute position for the new tile
                    logging.debug(f"Moving to stage position x: {pos_hor}, y: {pos_vert}")
                    self._stage_scan.moveAbsSync({'x': pos_hor, 'y': pos_vert})
```

src/odemis/acq/fastem.py
```diff
@@ -705,20 +705,35 @@ def pre_calibrate(self, pre_calibrations):
...
+                    continue
```
pieleric, line 733: This is not needed. It's the end of the for-loop, so it'll continue.

src/odemis/acq/fastem.py
```diff
@@ -705,20 +705,35 @@ def pre_calibrate(self, pre_calibrations):
...
+                                                          self._se_detector, self._ebeam_focus,
+                                                          calibrations=pre_calibrations)
+                    self._pre_calibrations_future.result()  # wait for the calibrations to be finished
+                    break  # if it successfully ran, do not try again
+                except CancelledError:
+                    logging.debug("Cancelled acquisition pre-calibrations.")
+                    raise
+                except Exception as err:
+                    logging.warning(f"Pre-calibration failed for ROA {self._roa.name.value} with error {err}, "
+                                    f"will try again.")
```
pieleric, line 725: This warning will also be shown when i == 2. That's a bit confusing, because it won't try again. Maybe place this warning as a `else` of the `if == 2`, to only show it in the first 2 trials.

pieleric (changes requested):

pieleric (approved):

PR #2783, opened 2024-05-27
[feature] add spectrograph iris support to SPARC GUI

pieleric:
Allow the user to control the "iris-in" axis from the spectrum and
temporal spectrum streams. This is to support the Kymera 328i, which has
an input iris.

### DIFF
--- a/src/odemis/gui/conf/data.py
+++ b/src/odemis/gui/conf/data.py
@@ -472,7 +472,12 @@ HW_SETTINGS_CONFIG = {
             ("grating", {}),
             ("slit-in", {
                 "label": "Input slit",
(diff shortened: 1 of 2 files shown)


src/odemis/gui/conf/data.py
```diff
@@ -472,7 +472,12 @@
             ("grating", {}),
             ("slit-in", {
                 "label": "Input slit",
-                "tooltip": u"Opening size of the spectrograph input slit.\nA wide opening means more light and worse resolution.",
+                "tooltip": "Opening size of the spectrograph input slit.\nA wide opening means more light and worse resolution.",
```
nandishjpatel, line 475: Do you also wanna correct the same Unicode string in lines 772, 907?

patrickcleeve2 (approved):

nandishjpatel (approved):

PR #2785, opened 2024-05-28
[feature] nfpm driver improvements

pieleric:


src/odemis/driver/nfpm.py
```diff
@@ -477,25 +477,30 @@ def moveAbs(self, pos):
     def stop(self, axes=None):
         self._executor.cancel()
 
+        # To be sure, stop explicitly all the axes (requested)
+        if axes is None:
+            self.AbortMotion()
+        else:
+            # Abort motion doesn't support per axis, so use "soft" stop instead
+            for an in axes:
+                aid = self._name_to_axis[an]
```
K4rishma, line 486: I assume aid is axis in digit, not completely clear what the full form is. self._name_to_axis  is taking in the axis name and converting it into integer, maybe a rename from self._name_to_axis to self._axis_to_number maybe more insightful ?

K4rishma (approved):

nandishjpatel (approved):

src/odemis/driver/nfpm.py
pieleric: Yes _name_to_axis is a dict str -> int. It's defined line 83. This attribute is used everywhere in the driver. Let's not change its name just for this small fix.

PR #2789, opened 2024-06-03
[feature] support for using modular files with METEOR configurations

sneepstefan:
The bash script odemis-yaml-selector should be placed in /usr/bin/ and should be fired from the 'exec=' line in de odemis.desktop file.


install/linux/usr/bin/odemis-yaml-selector
```diff
@@ -0,0 +1,43 @@
...
+counter=0
+
+# find all the files matching the search pattern and store them in an array
+files=($(find $search_dir -name $search_pat -exec basename {} \;))
+
+for fn in ${files[@]};
+do
+    # extract the microscope name by using this small Python script
+    # note: does not check for errors when parsing
+    #       spaces in the microscope names cannot be parsed
```
pieleric, line 18: It should handle spaces in names. 
In bash, it's a bit of a pain, but usually putting `"` before the variable tells bash you want to keep it a single word.
Here, you can write 
`mic_name="$(python3 -c "import sys; import yaml; from odemis.odemisd import modelgen; f = open('$search_dir$fn'); d = yaml.load(f, modelgen.SafeLoader); print(next(n for n, d in d.items() if d.get('class') == 'Microscope'))")"`

install/linux/usr/bin/odemis-yaml-selector
```diff
@@ -0,0 +1,43 @@
...
+    mic_name=$(python3 -c "import sys; import yaml; from odemis.odemisd import modelgen; f = open('$search_dir$fn');
+                          d = yaml.load(f, modelgen.SafeLoader); print(next(n for n, d in d.items() if d.get('class') == 'Microscope'))")
+    mic_list+=($mic_name)
+done
+
+# prompt the user with a list of microscope configurations to choose from
+selection=$(zenity --list --title "Config file selection" --list --text "Please select a METEOR model to start:" --column "" ${mic_list[@]} --hide-header --width=300 --height=400 2>/dev/null)
```
pieleric, line 25: You have to protect the mic_list from spaces. This is done as `"${mic_list[@]}"`.

install/linux/usr/bin/odemis-yaml-selector
```diff
@@ -0,0 +1,43 @@
...
+do
+    if test "$fn" = "$selection"
+    then
+        break
+    fi
+    let counter++
+done
+
+# merge the selected file name with the specified path
+selected_file=$search_dir${files[$counter]}
```
pieleric, line 38: You can completely avoid this by using two column in Zenity . The first one is the file name, and the second on the user friendly name. Then hide the first column, and ask Zenity to only output the first column.
This does the trick:
```
mic_list=()
for fn in "${files[@]}"; do
    # extract the microscope name by using this small Python script
    # note: does not check for errors when parsing
    #       spaces in the microscope names cannot be parsed
    mic_name="$(python3 -c "import sys; import yaml; from odemis.odemisd import modelgen; f = open('$search_dir$fn'); d = yaml.load(f, modelgen.SafeLoader); print(next(n for n, d in d.items() if d.get('class') == 'Microscope'))")"
    echo "$mic_name"
    mic_list+=("$fn")
    mic_list+=("$mic_name")
done

# prompt the user with a list of microscope configurations to choose from
selection="$(zenity --list --title "Config file selection" --list --text "Please select a METEOR model to start:" --column "" --column "" "${mic_list[@]}" --hide-header --width=300 --height=400 --hide-column 1 --print-column=1 2>/dev/null)"

if test "$selection" != ""; then
    echo odemis-start $selection
fi
```

install/linux/usr/bin/odemis-yaml-selector
pieleric, line 25: Please avoid the "METEOR" name, to keep the text more generic. I'd suggest "Please select a configuration to start".
For the title, I'd suggest to write "Odemis" somewhere to clarify for the user. Maybe something like "Odemis configuration selection"

pieleric (changes requested):

setup.py
```diff
@@ -94,7 +94,7 @@ def _get_version_git():
                'install/linux/usr/bin/odemis-edit-mic',
                'install/linux/usr/bin/odemis-hw-status',
                'install/linux/usr/bin/odemis-live-view',
-               'install/linux/usr/bin/odemis-mic-selector',
```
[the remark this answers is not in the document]
sneepstefan: Whoops indeed that line should not be deleted, that was never my intention although it happened. Will revert back!

install/linux/usr/bin/odemis-yaml-selector
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
```
tepals, line 2: can you add a description explaining what this script does?

install/linux/usr/bin/odemis-select-mic-start
```diff
@@ -0,0 +1,36 @@
...
+The script will search for configuration (.yaml) files with the used search directory
+and pattern. Once files are found microscope names are extracted and placed in a list.
+This list is presented to the user in the form of a (zenity) selection dialog window.
+After selection of the microscope configuration the right yaml file will be used to start Odemis.
+'
+# extract the arguments which were used to start this script and verify the length
+args=( "$@" )
+if test "${#args[@]}" != "2"; then
+        echo "Wrong number of arguments, expected 2, got ${#args[@]}"
+        echo "Syntax is: odemis-select-mic-start [search_dir] [search_pat]"
```
tepals, line 16: I would suggest to rename `search_pat` to `search_ptn` because `pat` is quite similar to `path`. It at least confused me, I was thinking: why is there a search directory and a search path?

install/linux/usr/bin/odemis-select-mic-start
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+: '
+This script is used when there are multiple modular configurations available for METEOR.
+It is recommended to combine it with the exec= command of the Odemis.desktop file.
+However it could be used stand-alone if it is required.
+The syntax to run this script -> odemis-select-mic-start [search_dir] [search_pat].
```
tepals, line 6: is it possible to run `odemis-select-mic-start --help` to get an explanation of the input arguments?

tepals (changes requested):

install/linux/usr/bin/odemis-select-mic-start
tepals: @sneepstefan is there a reason you did not update this, but resolved the conversation?

install/linux/usr/bin/odemis-select-mic-start
```diff
@@ -0,0 +1,52 @@
...
+args=( "$@" )
+
+for arg in ${args[@]}; do
+    if test "$arg" == "--help"; then
+        # Display Help
+        echo
+        echo "Syntax: odemis-select-mic-start [search_dir] [search_pat]"
+        echo "arguments:"
+        echo "search_dir     The directory to search for finding the main configuration files."
+        echo "search_pat     The search pattern to use to filter out specific file names."
```
tepals, line 22: I have some trouble to understand why you would need this, can you add a small example of a pattern?

install/linux/usr/bin/odemis-select-mic-start
```diff
@@ -0,0 +1,52 @@
...
+
+# check if exactly 2 arguments are used, any other combination is invalid
+if test "${#args[@]}" != "2"; then
+        echo "Wrong number of arguments, expected 2, got ${#args[@]}"
+        echo "Syntax is: odemis-select-mic-start [search_dir] [search_pat]"
+        exit 0
+fi
+
+# find all the files matching the search pattern and store them in an array
+files=("$(find ${args[0]} -name ${args[1]} -exec basename {} \;)")
```
pieleric, line 36: What I had in mind was to just pass directly the names of the yaml files as argument (full path). That would avoid fixing the search method inside the script. Also typically, I imagine you could just call the script as `odemis-select-mic-start /usr/share/odemis/*.yaml`, which is even simpler than the current way.

Or is there an advantage into explicitly passing the directory and search pattern?

install/linux/usr/bin/odemis-select-mic-start
sneepstefan: @tepals  Sorry my bad at the start of my modifications I changed it, then reverted back the whole file eventually and these changes were left out. Added them again.

tepals (approved):

install/linux/usr/bin/odemis-select-mic-start
sneepstefan: There needs to be a pattern to search for, specifically for finding the 'main' files. We don't want to any other yaml files as those would not contain a microscope name at all.
I have added a small example for searching METEOR files.

install/linux/usr/bin/odemis-select-mic-start
```diff
@@ -0,0 +1,52 @@
...
+    mic_name="$(python3 -c "import sys; import yaml; from odemis.odemisd import modelgen; f = open('${args[0]}$fn'); d = yaml.load(f, modelgen.SafeLoader); print(next(n for n, d in d.items() if d.get('class') == 'Microscope'))")"
+    mic_list+=("$fn")
+    mic_list+=("$mic_name")
+done
+
+# prompt the user with a list of microscope configurations to choose from
+selection="$(zenity --list --title "Odemis configuration starter" --list --text "Please select a configuration to start:" --column "" --column "" "${mic_list[@]}" --hide-header --width=300 --height=400 --hide-column 1 --print-column=1 2>/dev/null)"
+
+if test "$selection" != ""; then
+    odemis-start "${args[0]}$selection"
```
pieleric, line 51: I think that if the backend is already running, you should stop it first, with `sudo odemis-stop`.
Otherwise, odemis-start will detect it, and actually not start the backend again... starting the GUI for potentially a different backend.

Also, you might want at the beginning of the script check if the backend is running. In such case, you could ask the user whether to just start the GUI, or restart Odemis with a new backend. Otherwise, if the user closes the Odemis GUI window by mistake, it might be quite annoying that the only way to start it again is by restarting the whole backend. An alternative to that is to have the same behaviour as currently: if the backend is already running, just start the GUI. This can be done in the launcher with:
`if odemis check; then odemis-start; else odemis-select-mic-start; fi`

pieleric (changes requested):

install/linux/usr/bin/odemis-select-mic-start
```diff
@@ -0,0 +1,60 @@
...
+        echo "Trying to start the GUI"
+        odemis-gui
+        exit
+    else
+        exit
+    fi
+fi
+
+mic_list=()
+for fn in ${args[@]}; do
```
pieleric, line 47: I think you need to write `"${args[@]}"` to protect it from spaces. However, you can also simplify it and just write `for var in "$@"; do`  (no need to convert to an array).

Same thing for the --help search.

install/linux/usr/bin/odemis-select-mic-start
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+: '
+This script is used when there are multiple modular configurations available for METEOR.
+It is recommended to combine it with the exec= command of the Odemis.desktop file.
```
pieleric, line 4: `exec=` -> `Exec=` (case matters)

pieleric (changes requested):

pieleric (approved):

PR #2790, opened 2024-06-03
[TRCL-266] driver semnidaq: extend active state to be available on detectors too

pieleric:
For the counting PMT, we need to have a way to turn it on/off explicitly via a TTL port whenever we are acquiring.
It's essentially the same logic as for the scanner. So generalise the scanning state TTLs manager into a dedicated class, and introduce two arguments to be able to use it on the Detectors.

### DIFF
--- a/src/odemis/driver/semnidaq.py
+++ b/src/odemis/driver/semnidaq.py
@@ -75,7 +75,7 @@ import warnings
 import weakref
 from collections.abc import Iterable
 from enum import Enum
-from typing import Any, Dict, List, Optional, Tuple, Set, Union
+from typing import Any, Dict, List, Optional, Tuple, Set, Union, Callable
 
 import nidaqmx  # Tested with 0.9.0-dev (August 2023), modified to accept
 import numpy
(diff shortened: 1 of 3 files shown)


src/odemis/driver/semnidaq.py
```diff
@@ -1863,6 +1887,253 @@ def _downsample_pixels(cls, data: numpy.array,
...
+            raise ValueError("activation_delay %g s is not between 0 and 1000 s" % (activation_delay,))
+        self._activation_delay = activation_delay
+
+        self._ttls = {}  # NI Task -> high_auto (bool), high_enabled (bool), name (str)
+        self._ttl_setters = []  # To hold the partial VA setters
+        self._ttl_tasks = {}  # NI-DAQmx task to control the given D
+        self._ttl_lock = threading.Lock()  # Acquire to change the hw TTL state
+
+        # configure each channel for output
+        available_do_ports = sem.get_available_do_channels(port_number=0)
```
nandishjpatel, line 1919: Can you write a comment here, why use the port number as 0?

nandishjpatel (approved):

patrickcleeve2 (approved):

src/odemis/driver/semnidaq.py
pieleric: The hardware has multiple channels on each port. port 0 is the main one. It has up to 32 channels, and that's the only port we ever used so far. So for now it's hard-coded.
I've added a comment about it.

PR #2791, opened 2024-06-04
[FEAT][METEOR-1102] Add new positions for METEOR

patrickcleeve2:
Add support for the following positions:
- SEM IMAGING: perpendicular to the SEM
- FIB IMAGING: perpendicular to the FIB
- MILLING: at the milling tilt angle (SEM orientation)
- FM IMAGING: perpendicular to the FM

Add UI Elements:
- Buttons to move to FIB and MILL positions
- Interface for editing defined positions and transformations
- Disable these elements if metadata (yaml) hasn't been updated and show tooltip to user.


install/linux/usr/share/odemis/sim/meteor-sim.odm.yaml
```diff
@@ -71,14 +71,17 @@
...
+        POS_TRANS_COR: [48.8e-3, 0],    # [m] translation correction (x, y) (fib <-> meteor), standard value is x=48.8mm
+        SAMPLE_PRE_TILT: 0.61,          # [rad] pre-tilt angle for the sample holder, standard value is 35°
         # Active tilting (rx) & rotation (rz) angles positions when switching between SEM & FM, in radians.
         # Note: these values are calibrated at installation time.
-        FAV_FM_POS_ACTIVE: {"rx": 0.12213888553625313  , "rz":  5.06145}, # 7° - 270°
-        # Typically rz = 110°, but we make it 0 so that at init it looks like in SEM position
-        FAV_SEM_POS_ACTIVE: {"rx": 0, "rz": 0}    # Note that milling angle (rx) can be changed per session 
+        FAV_FM_POS_ACTIVE: {"rx": 0.2968  , "rz":  3.1415}, # 17° - 180°, FM position (perpendicular to the fm)
+        FAV_FIB_POS_ACTIVE: {"rx": 0.2968, "rz": 3.1415},   # 17° - 180°, FIB position (perpendicular to the fib)
+        FAV_SEM_POS_ACTIVE: {"rx": 0.61, "rz": 0},          # 35° - 0° sem position (perpendicular to the sem)
```
nandishjpatel, line 83: Have 4 decimal place precision like above? similarly for `FAV_MILL_POS_ACTIVE`, `rotation`

install/linux/usr/share/odemis/sim/meteor-sim.odm.yaml
```diff
@@ -71,14 +71,17 @@
         FM_IMAGING_RANGE: {"x": [0.040, 0.054], "y": [-5.e-3, 10.e-3], "z": [-0.5e-3, 8.e-3]},
         # Grid centers in SEM range
         # Adjusted so that at init (0,0,0), it's at the Grid 1.
-        SAMPLE_CENTERS: {"GRID 1": {'x': 0, 'y': 0, 'z': 0}, "GRID 2": {'x': 2.98e-3, 'y': 2.46e-3, 'z': 0}},
+        SAMPLE_CENTERS: {"GRID 1": {'x': 0, 'y': 0, 'z': 0}, "GRID 2": {'x': 5.0e-3, 'y': 0, 'z': 0}},
         # Mirroring values between SEM - METEOR
-        POS_COR: [0.02447, -0.000017],
+        POS_COR: [0.02447, -0.000017],  # [m] rotation correction (x, y) (sem <-> meteor)
+        POS_TRANS_COR: [48.8e-3, 0],    # [m] translation correction (x, y) (fib <-> meteor), standard value is x=48.8mm
```
nandishjpatel, line 77: Given the comments, I am not sure if the metadata should be used like this. @pieleric can give a better answer

src/odemis/acq/test/move_test.py
```diff
@@ -469,6 +469,21 @@ def test_moving_from_grid2_to_grid1_in_sem_imaging_area(self):
...
+        # move the stage to the loading position
+        f = self.posture_manager.cryoSwitchSamplePosition(LOADING)
+        f.result()
+        # move the stage to grid2
+        f = self.posture_manager.cryoSwitchSamplePosition(MILLING)
+        f.result()
+        position_label = self.posture_manager.getCurrentPostureLabel()
+        grid_label = self.posture_manager.getCurrentGridLabel()
+        self.assertEqual(position_label, MILLING)
+        self.assertEqual(grid_label, GRID_1)
```
nandishjpatel, line 482: Should this be `GRID_1` and/or `GRID_2`? based on comment on line 476

src/odemis/acq/test/move_test.py
```diff
@@ -488,6 +503,83 @@ def test_moving_from_sem_to_fm(self):
...
+        # check the values of tilt and rotation
+        fib_angles = self.stage.getMetadata()[model.MD_FAV_FIB_POS_ACTIVE]
+        for axis in self.ROTATION_AXES:
+            self.assertAlmostEqual(self.stage.position.value[axis], fib_angles[axis], places=4)
+
+    def test_moving_from_fib_to_sem(self):
+        # move to loading position
+        f = self.posture_manager.cryoSwitchSamplePosition(LOADING)
+        f.result()
+        # move the stage to the sem imaging area
```
nandishjpatel, line 529: Update comment?

src/odemis/gui/comp/positions.py
```diff
@@ -0,0 +1,254 @@
+import logging
+import math
+import wx
+from odemis import gui, model
+from odemis.gui.comp.text import UnitFloatCtrl
+
+
+class EditMeteorPositionsDialog(wx.Dialog):
```
nandishjpatel, line 8: Adjust this class like https://github.com/delmic/odemis/blob/master/src/odemis/gui/comp/settings.py to have a cleaner init?

src/odemis/gui/cont/tabs/cryo_chamber_tab.py
```diff
@@ -683,6 +714,21 @@ def _on_aligner_btn(self, evt):
         self._show_warning_msg(None)
         self.panel.btn_cancel.Enable()
 
+    def _edit_meteor_positions(self, evt):
+        """Open a dialog for editing meteor stage positions and metadata"""
+
+        stage_md = self._stage.getMetadata()
+        dialog = EditMeteorPositionsDialog(self.main_frame, stage_md)
+        # dialog.SetSize(size=(400, 400))
```
nandishjpatel, line 722: Remove comment?

src/odemis/model/_metadata.py
```diff
@@ -227,6 +228,8 @@
 MD_FM_IMAGING_RANGE = "FM imaging range"  # dict str → [float, float] defining the volume of the FM imaging area, along x, y and z axes.
 MD_FAV_FM_POS_ACTIVE = "Favourite FM position active"  # dict str->float representing the position required for FM imaging
 MD_FAV_SEM_POS_ACTIVE = "Favourite SEM position active"  # dict -> float representing the position required for SEM imaging
+MD_FAV_FIB_POS_ACTIVE = "Favourite FIB positive active"  # dict -> float representing the position required for FIB imaging
```
nandishjpatel, line 231:
```suggestion
MD_FAV_FIB_POS_ACTIVE = "Favourite FIB position active"  # dict -> float representing the position required for FIB imaging
```

install/linux/usr/share/odemis/sim/meteor-sim.odm.yaml
patrickcleeve2: I'm not sure I understand? The metadata is already being used for this, this only adds another parameter

nandishjpatel (approved):

src/odemis/gui/comp/positions.py
```diff
@@ -0,0 +1,254 @@
...
+                    "rx": math.radians(ctrl["rx"].GetValue()),
+                    "rz": math.radians(ctrl["rz"].GetValue()),
+                }
+
+        logging.debug(f"Updated stage metadata: {self.stage_md}")
+
+        self._update_pressed = True
+        self.Close()
+
+        # ref: add_setting_entry
```
pieleric, line 254: What does this mean?

src/odemis/gui/cont/tabs/cryo_chamber_tab.py
```diff
@@ -683,6 +714,20 @@ def _on_aligner_btn(self, evt):
...
         self.panel.btn_cancel.Enable()
 
+    def _edit_meteor_positions(self, evt):
+        """Open a dialog for editing meteor stage positions and metadata"""
+
+        stage_md = self._stage.getMetadata()
+        dialog = EditMeteorPositionsDialog(self.main_frame, stage_md)
+        dialog.Center()
+
+        _ = dialog.ShowModal()
```
pieleric, line 724: If you don't need the return value, just write:
```suggestion
        dialog.ShowModal()
```

src/odemis/gui/comp/positions.py
```diff
@@ -0,0 +1,254 @@
...
+                self.stage_md[name_map[name]] = (
+                    (ctrl["x"].GetValue()),
+                    (ctrl["y"].GetValue()),
+                )
+            else:
+                self.stage_md[name_map[name]] = {
+                    "rx": math.radians(ctrl["rx"].GetValue()),
+                    "rz": math.radians(ctrl["rz"].GetValue()),
+                }
+
```
pieleric, line 248: There is now one dict + a for loop with 2 if blocks. Just change it to 2 dicts with 2 for loops.

src/odemis/gui/comp/positions.py
```diff
@@ -0,0 +1,254 @@
...
+    def _on_cancel(self, event):
+        logging.debug("Cancel button clicked")
+        event.Skip()
+        self.Close()
+
+    def _on_update(self, event):
+        logging.debug("Update button clicked")
+        event.Skip()
+
+        self.stage_md = {}
```
pieleric, line 226: I think it should just be initialised at `__init__` by the `stage_md` argument.

src/odemis/gui/comp/positions.py
```diff
@@ -0,0 +1,254 @@
...
+
+        self.SetForegroundColour(gui.FG_COLOUR_EDIT)
+        self.SetBackgroundColour(gui.BG_COLOUR_MAIN)
+        self.SetTitle("Edit METEOR Positions")
+
+        self.main_sizer = wx.BoxSizer(wx.VERTICAL)
+        self.SetSizer(self.main_sizer)
+
+        self.gb_sizer = wx.GridBagSizer()
+        self.num_rows = 0
```
pieleric, line 24: This doesn't need to be an attribute. It could be just a variable (discarded after the end of `__init__`).

src/odemis/gui/cont/tabs/cryo_chamber_tab.py
```diff
@@ -683,6 +714,20 @@ def _on_aligner_btn(self, evt):
...
+        if dialog._update_pressed:
```
pieleric, line 725: Depending on a private variable is not the right way. The typical way in wxPython is to check the return value of `ShowModal`. If it something like `wx.OK`, then it means the user expects the data to be used.
Alternatively, you could have a special value of `stage_md` when the user didn't accept the changes.

src/odemis/gui/comp/positions.py
```diff
@@ -0,0 +1,254 @@
...
+        self.Layout()
+
+    def _on_cancel(self, event):
+        logging.debug("Cancel button clicked")
+        event.Skip()
+        self.Close()
+
+    def _on_update(self, event):
+        logging.debug("Update button clicked")
+        event.Skip()
```
pieleric, line 224: Most of the time, you don't need to skip wx events. Is it necessary here? If so, please add a comment explaining why.
Same for `_on_cancel()`

src/odemis/gui/comp/positions.py
```diff
@@ -0,0 +1,254 @@
...
+        )
+        self.num_rows += 1
+
+        trans_cor = stage_md[model.MD_POS_TRANS_COR]
+        pos_cor = stage_md[model.MD_POS_COR]
+
+        conf = {"unit": "m", "accuracy": 3}
+
+        pos: dict = {
+            "Rotation": pos_cor,
```
pieleric, line 153: It might be of nitpicking, but seeing "rotation", I would expect an angle. But it takes X&Y in meters. Maybe name it "Rotation center" to make this less odd.

src/odemis/gui/comp/positions.py
```diff
@@ -0,0 +1,254 @@
...
+            border=5,
+        )
+        self.num_rows += 1
+
+        trans_cor = stage_md[model.MD_POS_TRANS_COR]
+        pos_cor = stage_md[model.MD_POS_COR]
+
+        conf = {"unit": "m", "accuracy": 3}
+
+        pos: dict = {
```
pieleric, line 152: In general, I wonder how safe is it to allow the user to tweak these values from the GUI. Typically, you can already tweak them from the command line (using `odemis --update-metadata stage ...`). I am afraid that allowing the user to change them from the GUI would mean they could easily shoot themselves in the foot by entering an incorrect value. Or maybe, the GUI could be used to explicitly avoid such case by adding specific validation checks to the values so that it's only possible to make minor tweaks to them (for example: 1mm change maximum compared to the previous value).

src/odemis/gui/cont/tabs/cryo_chamber_tab.py
```diff
@@ -220,6 +232,22 @@ def __init__(self, name, button, panel, main_frame, main_data):
                                                                      self.panel.ctrl_sample_target_tmp,
                                                                      events=wx.EVT_COMMAND_ENTER)
 
+    def _enable_advanced_movement_btns(self):
+        """Enable the advanced movement buttons based on the posture manager flag"""
+        if self._role != 'meteor':
+            return
+
+        adv_movement_enabled = self.posture_manager._flag_advanced_movement
```
pieleric, line 240: If you use `_flag_advanced_movement` outside of the PostureManager, then don't mark it private.

pieleric (changes requested):

install/linux/usr/share/odemis/sim/meteor-sim.odm.yaml
```diff
@@ -71,14 +71,17 @@
...
+        SAMPLE_PRE_TILT: 0.6109,          # [rad] pre-tilt angle for the sample holder, standard value is 35°
         # Active tilting (rx) & rotation (rz) angles positions when switching between SEM & FM, in radians.
         # Note: these values are calibrated at installation time.
-        FAV_FM_POS_ACTIVE: {"rx": 0.12213888553625313  , "rz":  5.06145}, # 7° - 270°
-        # Typically rz = 110°, but we make it 0 so that at init it looks like in SEM position
-        FAV_SEM_POS_ACTIVE: {"rx": 0, "rz": 0}    # Note that milling angle (rx) can be changed per session 
+        FAV_FM_POS_ACTIVE: {"rx": 0.2968  , "rz":  3.1415}, # 17° - 180°, FM position (perpendicular to the fm)
+        FAV_FIB_POS_ACTIVE: {"rx": 0.2968, "rz": 3.1415},   # 17° - 180°, FIB position (perpendicular to the fib)
```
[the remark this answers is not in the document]
patrickcleeve2: Yes, its the same rotation and tilt as meteor (FM) except under the FIB/SEM column

patrickcleeve2:
closing, not priority atm. Will re-open later

PR #2792, opened 2024-06-06
[FEAT] [METEOR-1103] Extend Linked YZ to work at the SEM Orientation

patrickcleeve2:
Based on [calculations](https://docs.google.com/presentation/d/1sCIGlAac6wHCH5gUjwc7_t73HQVkSv9IJIVLMB0fv54/edit#slide=id.g2e03b59b4c5_0_331) we need to invert the z direction for the yz linked stage when the stage is at the sem orientation (rotation)

TODO:
- [ ] Restrict to TFS_1
- [ ] Hardware test


pieleric (changes requested):
It seems this feature means the stage goes opposite direction when some other component (stage) is in a specific position. IMHO, it's a big risk to introduce lots of odd surprises. This wrapper typically is made to manipulate a set of axes in a different frame of reference.  I think if you need 2 frames of references, maybe the best would be to have 2 instances of this wrapper?

PR #2793, opened 2024-06-07
[feature] spectrometer wrapper force settings when synchronized

pieleric:
In order to know the duration a frame will take, the camera component
must have all the VAs set. However, normally the spectrometer driver
only sets the VA just before the DataFlow is started. That's too late.
So, as a workaround, at least force the settings "flush" to also happen
when calling "synchronizedOn".

This at least allows to support our main use case of frameDuration,
which is to know the duration of the CCD while it's synchronized with
the e-beam scanner.

### DIFF
--- a/src/odemis/driver/spectrometer.py
+++ b/src/odemis/driver/spectrometer.py
@@ -262,6 +262,11 @@ class SpecDataFlow(model.DataFlow):
         logging.debug("Spectrometer acquisition finished")
 
     def synchronizedOn(self, event):
+        # Trick: for the frameDuration VA, all the settings must be applied... but this normally
+        # only happens when starting the acquisition. So we force the settings to be applied when
+        # setting synchronization.
+        if event is not None:
+            self.component._applyCCDSettings()
         self._ccddf.synchronizedOn(event)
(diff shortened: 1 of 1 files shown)


src/odemis/driver/spectrometer.py
```diff
@@ -262,6 +262,11 @@ def stop_generate(self):
         logging.debug("Spectrometer acquisition finished")
 
     def synchronizedOn(self, event):
+        # Trick: for the frameDuration VA, all the settings must be applied... but this normally
+        # only happens when starting the acquisition. So we force the settings to be applied when
+        # setting synchronization.
+        if event is not None:
```
nandishjpatel, line 268:
```suggestion
        if event.GetEventObject():
```
Maybe?

nandishjpatel (approved):

tepals (approved):

src/odemis/driver/spectrometer.py
pieleric: This is not a wx.Event, but a odemis.model.Event *or None*. If passed `None`, it means that the synchronization is disabled. 
I'll add the type to the function signature to clarify.

PR #2800, opened 2024-06-13
[feature][METEOR-1094] Chromatic Aberration plugin

K4rishma:
Create a  plugin to do lateral chromatic correction in Odemis
 - user add streams in the localization tab
 - selects sub menu in Data correction
 - plugin finds transformation per channel w.r.t a reference channel
 - displays corrected streams in the Analysis tab
 - save the transformation values per channel in the yaml file
 - Get the output in the same format to be entered in the yaml file 
 -  find a good file location for yaml file storing the transformation values
-  Add test cases for find_corresponding_points


plugins/chromatic_correction.py
```diff
@@ -0,0 +1,222 @@
...
+            box.Destroy()
+            return
+
+        # Calculate Maximum Intensity projection for all the present channels in the stream panel
+        for i, da in enumerate(das):
+            mip_image = model.DataArray(numpy.amax(da, axis=0), copy.copy(da.metadata))
+            dataio.tiff.export(f"{da.metadata[model.MD_DESCRIPTION]}_{i}", mip_image)
+            # All the channels will be corrected according to a reference colored channel
+            # green channel is selected as reference channel, Why? -> Ask Deniz
+            if da.metadata[model.MD_USER_TINT] == (0, 255, 0):
```
patrickcleeve2, line 149: I think this should be based off the emission/excition rather than the colour tint. It might be best to let the user select the channel somehow?

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,222 @@
...
+        # save the image
+        dataio.tiff.export(filename, data)
+        # Read the saved images and display
+        im_read = dataio.tiff.read_data(filename)
+        aligned_stream = StaticFluoStream(im_read[0].metadata[model.MD_DESCRIPTION], im_read[0])
+        scont = analysis_tab.stream_bar_controller.addStream(aligned_stream, add_to_view=True)
+        scont.stream_panel.show_remove_btn(True)
+
+    def start(self):
+        tab_data = self.main_app.main_data.tab.value.tab_data_model
```
patrickcleeve2, line 121: I'm assuming this is meant to be run from the localization tab, so it might be better to make it explictly refer to that. 

``` python
self.localization_tab: LocalizationTab  = self._main_data_model.getTabByName("cryosecom-localization")
tab_data = self.localization_tab.tab_data_model
acq_streams = tab_data.acquisitionStreams.value
```

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,222 @@
...
+        self.new = []
+        self.ref = []
+        levels = generate_zlevels(tab_data.main.focus,
+                                  (-10e-06, 10e-06),
+                                  5e-06)
+
+        non_static = [s for s in tab_data.streams.value
+                      if isinstance(s, FluoStream)]
+        zlevels = {s: levels for s in non_static}
+        settings_observer = SettingsObserver(model.getComponents())
```
patrickcleeve2, line 131: You should have access to this already through: 

```
tab_data.main.settings_obs
```

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,222 @@
...
+        :param filename: filename of the data array
+        :param data: value and metadata of the given data array
+        """
+        # display the reference in the analysis tab
+        analysis_tab = self.main_app.main_data.getTabByName('analysis')
+        # save the image
+        dataio.tiff.export(filename, data)
+        # Read the saved images and display
+        im_read = dataio.tiff.read_data(filename)
+        aligned_stream = StaticFluoStream(im_read[0].metadata[model.MD_DESCRIPTION], im_read[0])
```
patrickcleeve2, line 116: You should be able to use the following helpers to convert the data to static stream directly, rather than reloading the file:

```python
from odemis.util.dataio import data_to_static_streams

```

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,222 @@
...
+            else:
+                self.new.append(mip_image)
+
+        # Save and display the reference channel in the analysis tab
+        self.show_static_stream(f"modified_{self.ref.metadata[model.MD_DESCRIPTION]}_ref", self.ref)
+
+        # Compute transformation between reference channel and other channels
+        trans_per_channel = {}  # stores transformation values per channel
+        _, centers_ref = process_image(self.ref)
+        centers1 = numpy.array(centers_ref)
```
patrickcleeve2, line 160: You can just override the name directly, rather than using centers1 which is a bad name or just convert to np array inside the function as it seems to always need to happen:

```suggestion
        centers_ref = numpy.array(centers_ref)

```

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,222 @@
...
+        for i, im in enumerate(self.new):
+            # Find corresponding points
+            _, centers_new = process_image(im)
+            centers2 = numpy.array(centers_new)
```
patrickcleeve2, line 164: Same naming comment as above
```suggestion
            centers_new = numpy.array(centers_new)
```

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,222 @@
...
+        # Read the saved images and display
+        im_read = dataio.tiff.read_data(filename)
+        aligned_stream = StaticFluoStream(im_read[0].metadata[model.MD_DESCRIPTION], im_read[0])
+        scont = analysis_tab.stream_bar_controller.addStream(aligned_stream, add_to_view=True)
+        scont.stream_panel.show_remove_btn(True)
+
+    def start(self):
+        tab_data = self.main_app.main_data.tab.value.tab_data_model
+        self.new = []
+        self.ref = []
```
patrickcleeve2, line 123: Use more descriptive names for these, something like reference_streams

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,222 @@
...
+                c, d = mat[1, 0], mat[1, 1]
+                shear = numpy.arctan2(a * c + b * d, scale[0] * scale[1])
+
+                # compute translation
+                tx = mat[0, 2]
+                ty = mat[1, 2]
+                pixel_size = im.metadata[model.MD_PIXEL_SIZE]
+                translation = (tx * pixel_size[0], -ty * pixel_size[1])
+
+                trans_per_channel[f"channel_{i}"] = {
```
patrickcleeve2, line 203: I would probably also store some more channel information here, such as the emission/excitation and the reference as well. That way we have all the information to re-create the calibration

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,222 @@
...
+            max_distance = BEAD_DIAMETER
+            corresponding_pairs = find_corresponding_points(centers1, centers2, max_distance)
+
+            # Extract the matching points
+            points1 = centers1[corresponding_pairs[:, 0]]
+            points2 = centers2[corresponding_pairs[:, 1]]
+
+            if len(corresponding_pairs) > 4:
+                # Estimate the transformation matrix
+                affine2 = AffineTransform.from_pointset(points1, points2)
```
patrickcleeve2, line 174:
```suggestion
                affine = AffineTransform.from_pointset(points1, points2)
                mat = numpy.eye(3)
                mat[:2, :2] = affine.matrix
                mat[:2, 2] = affine.translation

```

patrickcleeve2 (changes requested):

plugins/chromatic_correction.py
K4rishma: The idea here is save the file and read from the saved file to check the correction metadata is saved properly. Eric suggested another function that does the same from gui/plugin
`    def showAcquisition(self, filename):
        """
        Show the analysis (aka Gallery) tab and opens the given acquisition file.
        filename (str): filename of the file to open.
        """
        analysis_tab = self.main_app.main_data.getTabByName('analysis')
        self.main_app.main_data.tab.value = analysis_tab
        analysis_tab.load_data(filename)`

patrickcleeve2 (approved):

K4rishma:
Get the output in the same format to be entered in the yaml file

pieleric (changes requested):

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,228 @@
...
+    # Check the number of channels in the image
+    if len(image.shape) == 2:  # Grayscale image
+        gray = image
+    elif len(image.shape) == 3 and image.shape[2] == 3:  # BGR image
+        gray = cv2.cvtColor(image, cv2.COLOR_BGR2GRAY)
+    else:
+        raise ValueError("Unexpected image format!")
+
+    # Apply threshold
+    _, binary = cv2.threshold(gray, 200, 255, cv2.THRESH_BINARY)
```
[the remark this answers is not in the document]
K4rishma: It is based on the dataset provided on chromatic aberration. Tested their on the images. I can provide a formulation for this number based on the histogram, to avoid fixing it.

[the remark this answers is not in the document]
K4rishma: I am not able to find any formulation for this magic number, I tried what otsu thresholding provides and it does not match with this number. Thresholding with this number worked on the images provided by Deniz to calculate lateral chromatic shift such that after thresholding it keeps the size of beads is same in binary image as in gray scale image. When Otsu thresholding is used, the beads in binary image are dilated. The dilation would affect the operations using bead diameter. If it is fine, I can make it a global parameter  or a input parameter for the functions and use it like this for now. As it is tested more, we can find a better threshold if it does not work ?

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,228 @@
...
+    :param centers2: (np.ndarray) Mx2 array of points (x, y) in the second image.
+    :param max_distance: (float) Maximum allowable distance for point pairs.
+    :returns:
+        pairs: (list of tuples) Corresponding point pairs (index1, index2).
+    """
+    # Calculate pairwise distances
+    distances = cdist(centers1, centers2)
+
+    # Find pairs within the max_distance
+    pairs = numpy.argwhere(distances <= max_distance)
```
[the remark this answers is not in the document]
K4rishma: True, I should take multiple points pairing into account and sort them to get unique pairs.

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,228 @@
...
+from odemis.acq.acqmng import acquireZStack
+from odemis.acq.stream import FluoStream
+from odemis.gui.plugin import Plugin
+from odemis.util.comp import generate_zlevels
+from odemis.util.transform import AffineTransform
+
+BEAD_DIAMETER = 10 * numpy.sqrt(2)  # the length of the square that fits in the bead circle
+
+
+def process_image(image):
```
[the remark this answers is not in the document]
K4rishma: This function is advanced and works beautifully

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,228 @@
...
+                tx = mat[0, 2]
+                ty = mat[1, 2]
+                pixel_size = im.metadata[model.MD_PIXEL_SIZE]
+                translation = (tx * pixel_size[0], -ty * pixel_size[1])
+
+                # stream settings
+                input_wl = im.metadata[model.MD_IN_WL]
+                output_wl = im.metadata[model.MD_OUT_WL]
+
+                trans_per_channel[f"channel_{i}"] = {
```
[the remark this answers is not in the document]
K4rishma: It is a good idea, it will make the workflow easy for the service engineer. I will incorporate the comments. The yaml file icluding the include will be there in the other pull request. I have added more comments to explain the structure.

K4rishma:
Still need to add test cases for find_corresponding_points but resolved the previous comments. Ready to re-review.

src/odemis/model/_metadata.py
```diff
@@ -271,3 +271,5 @@
 # Fastem: Parameters used for stitching and reconstruction of 3D volumes
 MD_SLICE_IDX = "Index of slice in volume stack"  # int
 MD_FIELD_SIZE = "Average field of view of a megafield"  # tuple (px, px)
+
+MD_CHROMATIC_COR = "Chromatic cor"
```
pieleric, line 275: Please give a better string, it should be "user-friendly". Also add a comment of the format of the data it contains.

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,252 @@
...
+        # Save and display the reference channel in the analysis tab
+        self.show_static_stream(f"modified_{reference_mip_stream.metadata[model.MD_DESCRIPTION]}_ref.ome.tiff",
+                                reference_mip_stream)
+
+        # Compute transformation between reference channel and other channels
+        max_distance = ACTUAL_BEAD_DIAMETER * SCALE  # size of bead in the image
+        trans_per_channel = self.compute_transformation(reference_mip_stream, other_mip_streams, max_distance)
+
+        if trans_per_channel:
+            file_location = os.path.join(odemis.gui.conf.file.CONF_PATH, "lens-chromatic-correction.odm.yaml")
```
pieleric, line 250: I don't think it's a really convenient place, because it's in a hidden ".config" folder. (and that's not the folder where the microscope file is stored). I think it's nicer for the user to store it in `~/Documents/`.

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,252 @@
...
+                    im.metadata.update({model.MD_POS_COR: translation})
+                    im.metadata.update({model.MD_ROTATION_COR: rotation})
+                    im.metadata.update({model.MD_SHEAR_COR: shear})
+
+                    # save the image and display
+                    self.show_static_stream(f"modified_{im.metadata[model.MD_DESCRIPTION]}_{i}.ome.tiff", im)
+                except (ValueError, KeyError) as ex:
+                    logging.error("Failed to compute chromatic aberration correction yaml file due to: %s", ex)
+
+        return trans_per_channel
```
pieleric, line 204: Can you log `trans_per_channel`, as logging.debug(). This way, if something goes wrong and the file is updated/overridden, or sometimes the value is incorrectly computed, it's easy to check the different values computed.

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,252 @@
...
+                      if isinstance(s, FluoStream)]
+        zlevels = {s: levels for s in non_static}
+        settings_observer = tab_data.main.settings_obs
+        acqui_task = acquireZStack(non_static, zlevels, settings_obs=settings_observer)
+        das, e = acqui_task.result()
+
+        # Calculate Maximum Intensity projection for all the present channels in the stream panel
+        for i, da in enumerate(das):
+            mip_image = model.DataArray(numpy.amax(da, axis=0), copy.copy(da.metadata))
+            dataio.tiff.export(f"{da.metadata[model.MD_DESCRIPTION]}_{i}.ome.tiff", mip_image)
```
pieleric, line 233: Please don't save it "here" (which is a sort of random folder). Instead, explicitly select a folder. I'd recommend creating a folder dedicated for that like in `~/Document/Chromatic aberration correction" or the same in `~/Pictures`. Maybe also append the date, so there are less chances it overrides the previous (good) values by mistake.
The YAML file could then be stored in the same folder.

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,252 @@
...
+                except (ValueError, KeyError) as ex:
+                    logging.error("Failed to compute chromatic aberration correction yaml file due to: %s", ex)
+
+        return trans_per_channel
+
+    def start(self):
+        localization_tab = self.main_app.main_data.getTabByName("cryosecom-localization")
+        tab_data = localization_tab.tab_data_model
+        other_mip_streams = []
+        reference_mip_stream = []
```
pieleric, line 210: Should be `None`, as normally it just a single value.

pieleric, line 210: Let's not call them "stream" as they contains DataArrays. I'd suggest reference_mip_da and other_mip_das.

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,252 @@
...
+
+        if len(tab_data.streams.value) <= 1:
```
pieleric, line 212: It should check more carefully. If *either*:
* there is not "green" stream (aka ref stream)
* there is less than 2 streams
* Several streams have the same band

plugins/chromatic_correction.py
pieleric, line 233: As each stream has a different name, I think it'd be easier to find the files if you *start* the filename with `{i}`.

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,252 @@
...
+            dataio.tiff.export(f"{da.metadata[model.MD_DESCRIPTION]}_{i}.ome.tiff", mip_image)
+            # All the channels will be corrected according to a reference colored channel
+            # green channel is selected as reference channel, as it lies in the middle of the spectrum
+            if da.metadata[model.MD_USER_TINT] == (0, 255, 0):
+                reference_mip_stream = mip_image
+            else:
+                other_mip_streams.append(mip_image)
+
+        # Save and display the reference channel in the analysis tab
+        self.show_static_stream(f"modified_{reference_mip_stream.metadata[model.MD_DESCRIPTION]}_ref.ome.tiff",
```
pieleric, line 242: Technically, it's *not* modified. So I'd recommend to just name it "{i}_description_reference...".
The other streams should be named "{i}_description_corrected...".

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,252 @@
...
+            if len(corresponding_pairs) < 4:
+                txt = ('The chromatic correction did not happen for channel with %s. Skipping this channel.\n\n'
+                       'Alternatively, navigate to location with more beads or change the bead diameter in the script'
+                       % im.metadata[model.MD_DESCRIPTION])
+                box = wx.MessageDialog(self.main_app.main_frame,
+                                       txt, "Partial failure of chromatic correction", wx.OK)
+                box.ShowModal()
+                box.Destroy()
+
+            else:
```
pieleric, line 140: Instead of a `else`, place a `continue` at the end of the `if` block, and here, the "normal" code can be one indentation less, as if there was no error handling.

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,252 @@
...
+                # compute shear
+                a, b = mat[0, 0], mat[0, 1]
+                c, d = mat[1, 0], mat[1, 1]
+                shear = numpy.arctan2(a * c + b * d, scale[0] * scale[1])
+
+                # compute translation
+                tx = mat[0, 2]
+                ty = mat[1, 2]
+                pixel_size = im.metadata[model.MD_PIXEL_SIZE]
+                translation = (tx * pixel_size[0], -ty * pixel_size[1])
```
pieleric, line 169: Please move this to a separate function which takes points1 & points2, and returns scale, translation, rotation, and shear.

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,252 @@
...
+
+                # stream settings
+                input_wl = im.metadata[model.MD_IN_WL]
+                output_wl = im.metadata[model.MD_OUT_WL]
+
+                try:
+                    fl_band = im.metadata[model.MD_EXTRA_SETTINGS]['Filter Wheel']['position'][0]['band']
```
pieleric, line 176: Pulling it out of the MD_EXTRA_SETTINGS seems a little like pulling a rabbit out a hat... I'd suggest to get the "channels" as a `Dict[float, DataArray]`. The key is the band to store in the MD_CHROMATIC_COR.
It's up to the caller to find the info.
One (maybe better) option in the caller is to use  `s._emission_to_idx[s.emission.value]` (where `s` is a FluoStream). See `FluoStream._setup_emission()`.

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,252 @@
...
+                        "Pixel size cor": [float(scale[0]), float(scale[1])],
+                        "Centre position cor": [float(translation[0]), float(translation[1])],
+                        "Rotation cor": float(rotation),
+                        "Shear cor": float(shear),
+                        # Extra information saved for tracking/debugging purpose
+                        "tint": list(im.metadata[model.MD_USER_TINT]),
+                        "excitation": [float(input_wl[0]), float(input_wl[1])],
+                        "emission": [float(output_wl[0]), float(output_wl[1])],
+                        "exposure time": float(im.metadata[model.MD_EXP_TIME]),
+                        "light power": float(im.metadata[model.MD_LIGHT_POWER])
```
pieleric, line 190: You can also replace these keys by there standard MD_* version.

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,252 @@
...
+                    }
+
+                    # update metadata
+                    im.metadata.update({model.MD_PIXEL_SIZE_COR: scale})
+                    im.metadata.update({model.MD_POS_COR: translation})
+                    im.metadata.update({model.MD_ROTATION_COR: rotation})
+                    im.metadata.update({model.MD_SHEAR_COR: shear})
+
+                    # save the image and display
+                    self.show_static_stream(f"modified_{im.metadata[model.MD_DESCRIPTION]}_{i}.ome.tiff", im)
```
pieleric, line 200: Let's call them  "{i}_description_corrected...".

plugins/chromatic_correction.py
pieleric: I'd say it's not so bad, because using tint allows the user to select any channel (as the tint can be user selectable). As this plugin is to be used only internally, once per system, I think it's a good trade-off between development time and flexibility. 
Would be worth mentioning this in the comment.

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,252 @@
...
+    def compute_transformation(self, ref_channel, channels, max_distance):
+        """
+        Compute affine transformation between a reference channel and another coloured channel.
+        :param ref_channel: (model.DataArray) The reference channels to which other channels are aligned
+        :param channels: (list of model.DataArray) The channels are transformed to align with reference channel
+        :param max_distance: (float in meters) The maximum distance accepted to match two same locations in between
+         reference channel and channels
+        """
+        trans_per_channel = {model.MD_CHROMATIC_COR: {}}  # stores transformation values per channel
+        dist_pxl = int(max_distance / ref_channel.metadata[model.MD_PIXEL_SIZE][0])  # convert m to pixel units
```
pieleric, line 119: As it's just `max_distance` in a different unit, let's name it `max_distance_px`.
Also, you could add an assert max_distance_px > 1

pieleric (changes requested):

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,228 @@
...
+        # Compute transformation between reference channel and other channels
+        trans_per_channel = {}  # stores transformation values per channel
+        _, centers_ref = process_image(reference_mip_stream)
+        centers_ref = numpy.array(centers_ref)
+        for i, im in enumerate(other_mip_streams):
+            # Find corresponding points
+            _, centers_new = process_image(im)
+            centers_new = numpy.array(centers_new)
+            max_distance = BEAD_DIAMETER
+            corresponding_pairs = find_corresponding_points(centers_ref, centers_new, max_distance)
```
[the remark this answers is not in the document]
pieleric: I'd suggest to adjust compute_transformation() to accept 2 parameters: spot_diameter, and max_distance.
The first one you use in MaximaFind, and the second one in `find_corresponding_points()`.
Then add a comment just before calling `compute_transformation()`  to explain why you use spot_diameter as a max_distance too. Or maybe for now you can use spot_diameter * 1.1 to already show that it can be different... but for now you want to keep it small to handle better the clusters.

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,252 @@
...
+                # stream settings
+                input_wl = im.metadata[model.MD_IN_WL]
+                output_wl = im.metadata[model.MD_OUT_WL]
+
+                try:
+                    fl_band = im.metadata[model.MD_EXTRA_SETTINGS]['Filter Wheel']['position'][0]['band']
+                    # The below metadata dictionary will be added to the microscope file using input command
+                    # under the filter wheel component, only if the metadata has Pixel size cor, Centre position cor,
+                    # Rotation cor, and Shear cor for a minimum one filter wheel band position
+                    trans_per_channel[model.MD_CHROMATIC_COR][fl_band] = {
```
pieleric, line 180: You should also create an entry for the reference channel. It should set all the metadata to the default values. This way, when the user goes back to the "reference" channel, all the metadata is reset.

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,252 @@
...
+                        "Pixel size cor": [float(scale[0]), float(scale[1])],
+                        "Centre position cor": [float(translation[0]), float(translation[1])],
+                        "Rotation cor": float(rotation),
+                        "Shear cor": float(shear),
+                        # Extra information saved for tracking/debugging purpose
```
pieleric, line 185: The drawback with such "debugging" metadata is that it'll be on the microscope file, so the MD updater would read it, and copy it onto the other components (assuming you keep the code generic/simple). I think it'd be best if you could put them as comment, or at least as special metadata which will not be confused.

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,252 @@
...
+                try:
+                    fl_band = im.metadata[model.MD_EXTRA_SETTINGS]['Filter Wheel']['position'][0]['band']
+                    # The below metadata dictionary will be added to the microscope file using input command
+                    # under the filter wheel component, only if the metadata has Pixel size cor, Centre position cor,
+                    # Rotation cor, and Shear cor for a minimum one filter wheel band position
+                    trans_per_channel[model.MD_CHROMATIC_COR][fl_band] = {
+                        "Pixel size cor": [float(scale[0]), float(scale[1])],
+                        "Centre position cor": [float(translation[0]), float(translation[1])],
+                        "Rotation cor": float(rotation),
+                        "Shear cor": float(shear),
```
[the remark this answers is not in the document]
K4rishma: From the latest comments, Is it correct that I should not use MD_* and leave it as it is?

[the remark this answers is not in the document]
pieleric: For these "real" metadata, that you really want to pass to the "ccd", you should use `MD_*`.

plugins/chromatic_correction.py
K4rishma: Would it be a good idea to put datetime stamp instead of simple {i}? It would be indeed nice to start with i for proper sorting of multiple files. Then even if multiple streams have same name, that should not be an issue and it would be fine not to check it, right?

tmoerkerken:
When I run the plugin, it straight away closes the application:
```
python3: ../../src/xcb_io.c:374: poll_for_response: Assertion `!xcb_xlib_threads_sequence_lost' failed.
/usr/bin/odemis-gui: line 8: 24007 Aborted                 ${PYTHON_INTERPRETER:-/usr/bin/python3} -m $GUI "$@"
WARNING:root:The GUI closed with the error code 134.
```
Maybe introduced by the recent addition of threading. We can schedule some time to figure this out.

tmoerkerken (changes requested):

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,346 @@
...
+    """
+    Find corresponding points in two sets of points that are within a specified maximum distance.
+    :param centers_1: Nx2 array of points (x, y) in the first image.
+    :param centers_2: Mx2 array of points (x, y) in the second image.
+    :param max_distance: (float in pixels) Maximum allowable distance for point pairs.
+    :returns:
+        pairs: (numpy.ndarray) Corresponding point pairs (index1, index2).
+    """
+    # Calculate pairwise distances
+    distances = cdist(centers_1, centers_2)
```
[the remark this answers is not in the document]
tmoerkerken: We looked at this, and it is out of scope for now.

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,346 @@
...
+    :param points1: Nx2 (x, y) location of features
+    :param points2: Nx2 (x, y) location of the corresponding features
+    :returns:
+        tuple: A tuple containing correction parameters:
+            - scale_cor ((float, float)): Scaling factors along the x and y axes
+            - translation_cor ((float, float)): Translation components along the x and y axes
+            - rotation_cor (float): Rotation angle (in radians)
+            - shear_cor (float): Shear angle (in radians)
+    """
+    affine = AffineTransform.from_pointset(points1, points2)
```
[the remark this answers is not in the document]
tmoerkerken: We looked at this, and apparently the current AffineTransform module yields different results than this implementation. We should give it a critical look in the future.

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,346 @@
...
+class ChromaticCorrectionPlugin(Plugin):
+    name = "Chromatic Correction"
+    __version__ = "1.0"
+    __author__ = "Karishma Kumar"
+    __license__ = "GPLv2"
+
+    def __init__(self, microscope, main_app):
+        if "cryosecom-localization" in main_app.main_data.tab.choices.values():
+            super().__init__(microscope, main_app)
+        self.addMenu("Data correction/Lateral chromatic correction", self.start_with_thread)
```
tmoerkerken, line 197: Spawning a dialog first, like we discussed yesterday. Make use of the `@call_in_wx_main` decorator. Then there is also no use anymore for the other modal you are spawning (the one with the problems), since we can show problems beforehand. Something like this:
image

tmoerkerken (changes requested):

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,346 @@
...
+
+def update_chromatic_correction(im: DataArray, points_ref: numpy.ndarray, points_im: numpy.ndarray):
+    """
+    Update the transformation correction parameters that maps features given by points_im to set of features given
+    by point_ref.
+    :param points_ref: (x, y) location of features in reference data
+    :param points_im: (x, y) location of features in given data by im
+    :return: (Dict) transformation values of the given data
+    """
+    if numpy.array_equal(points_ref, points_im):
```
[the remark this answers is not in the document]
K4rishma: I will fix it. It is indeed odd.

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,346 @@
...
+
+    # Compute rotation
+    rotation_cor = numpy.arctan2(mat[1, 0] / sx, mat[0, 0] / sx)
+
+    # compute shear
+    a, b = mat[0, 0], mat[0, 1]
+    c, d = mat[1, 0], mat[1, 1]
+    shear_cor = numpy.arctan2(a * c + b * d, scale_cor[0] * scale_cor[1])
+
+    return scale_cor, translation_cor, rotation_cor, shear_cor
```
tepals, line 128: I don't understand why you can't just use `affine.translation, affine.scale, affine.rotation, affine.shear`. 
If this is really different, can you add a comment explaining why?

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,346 @@
...
+    file_location = os.path.join(dir_path, f"{current_date}_{filename}.{ext}")
+    return file_location
+
+
+def update_chromatic_correction(im: DataArray, points_ref: numpy.ndarray, points_im: numpy.ndarray):
+    """
+    Update the transformation correction parameters that maps features given by points_im to set of features given
+    by point_ref.
+    :param points_ref: (x, y) location of features in reference data
+    :param points_im: (x, y) location of features in given data by im
```
tepals, line 145: can you add the unit? I assume it's pixels

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,346 @@
...
+            transformation_dict = self.find_and_display_transformation(mip_das_dict, max_dist_pxl, ref_em_index)
+
+            # Save the transformation if it is not empty
+            if transformation_dict[model.MD_CHROMATIC_COR]:
+                file_location = get_file_location("lens_chromatic_correction", "odm.yaml")
+                with open(file_location, 'w') as yaml_file:
+                    yaml.dump(transformation_dict, yaml_file, default_flow_style=False, sort_keys=False)
+
+        except Exception as ex:
+            logging.error(" Chromatic Aberration Correction failed to: %s", ex)
```
tepals, line 342:
```suggestion
            logging.error("Chromatic Aberration Correction failed to: %s", ex)
```

tepals (changes requested):

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,346 @@
...
+    :param points2: Nx2 (x, y) location of the corresponding features
+    :returns:
+        tuple: A tuple containing correction parameters:
+            - scale_cor ((float, float)): Scaling factors along the x and y axes
+            - translation_cor ((float, float)): Translation components along the x and y axes
+            - rotation_cor (float): Rotation angle (in radians)
+            - shear_cor (float): Shear angle (in radians)
+    """
+    affine = AffineTransform.from_pointset(points1, points2)
+    mat = numpy.eye(3)
```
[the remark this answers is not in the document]
tmoerkerken: Similar to other comment, will close this one

plugins/chromatic_correction.py
K4rishma: This function does two things - calculate the affine transformation parameters and then does correction on those parameters such that Odemis can display it through the correction metadata. To see if the first method works, I checked how the Odemis display images. I did not dwell a lot on the mathematics. 

For the uncorrected affine parameters, these calculated values do not match with the attributes of affine transform. The closest values come from this function alt_transformation_matrix_to_implicit(affine.matrix, "RSU"), in which rotation and scale matches but not shear.  

As chromatic aberration is a setup to test the approach, in my opinion making a separate function will help in faster debugging and fixing. This way we can ensure that the already used functions in util/transform.py are separate and are not modified for this single purpose.

K4rishma: I have explained the same in the comment

plugins/chromatic_correction.py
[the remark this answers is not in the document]
tepals: I think this is a serious problem and an indication that one of the two is incorrect. So far I've never had a problem with the AffineTransform module, so it might just be a different interpretation of shear/scale/squeeze. I am a bit hesitant to approve this PR until it's been figured out what the origin of the difference is.

plugins/chromatic_correction.py
pieleric: It could be that `alt_transformation_matrix_to_implicit()` contains bugs. I don't think it was extensively validated, especially wrt the shear.
Please compare to `acq.stream._base.getPixelCoordinates()` (does the reverse), and `util.conversion.get_img_transformation_md()`. It might be possible to directly use this second function directly, by creating a 3x3 matrix from the AffineTransform.matrix + translation. (if that works, maybe you could check that your current code and get_img_transformation_md() always behave the same, in which case you could replace the content of get_img_transformation_md() with your current code, which looks very neat).

plugins/chromatic_correction.py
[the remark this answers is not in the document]
pieleric: I fully agree that this function should be moved to `odemis.util`. Either in util.conversion or util.transform.

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,392 @@
...
+        for da in da_list:
+            mip_image = model.DataArray(numpy.amax(da, axis=0), copy.copy(da.metadata))
+            file_location = get_file_location(f"{da.metadata[model.MD_DESCRIPTION]}", "ome.tiff")
+            dataio.tiff.export(file_location, mip_image)
+            if all(numpy.isclose(numpy.array(da.metadata[model.MD_OUT_WL]), self.ref_emission, rtol=1e-09, atol=1e-09)):
+                ref_em_pos = streams_list[0]._emission_to_idx[tuple(da.metadata[model.MD_OUT_WL])]
+                mip_das_dict[ref_em_pos] = mip_image
+            else:
+                em_index = streams_list[0]._emission_to_idx[tuple(da.metadata[model.MD_OUT_WL])]
+                mip_das_dict[em_index] = mip_image
```
pieleric, line 366: I think this would be easier to read:
```suggestion
            em_index = streams_list[0]._emission_to_idx[tuple(da.metadata[model.MD_OUT_WL])]
            mip_das_dict[em_index] = mip_image
            if all(numpy.isclose(numpy.array(da.metadata[model.MD_OUT_WL]), self.ref_emission, rtol=1e-09, atol=1e-09)):
                ref_em_pos = em_index

```

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,392 @@
...
+    :return: (Dict) transformation values of the given data
+    """
+    affine = AffineTransform.from_pointset(points_ref, points_im)
+    # Estimate the transformation matrix
+    scale, translation, rotation, shear = extract_transformation_parameters(affine)
+    # Convert translation from pixels to m/physical units
+    pixel_size = im.metadata[model.MD_PIXEL_SIZE]
+    translation = [translation[0] * pixel_size[0], translation[1] * pixel_size[1]]
+    # Correct the metadata of im
+    update_im_metadata(im, scale, translation, rotation, shear)
```
pieleric, line 193: It's a little unexpected that the function updates the metadata of the image. Typically, you don't want to change the input parameters. Either indicate it clearly in the docstring (that `im` is being updated), or let the caller code do it (it's just 1 line). I'd favour the second option, but up to you.

plugins/test/chromatic_correction_test.py
```diff
@@ -0,0 +1,87 @@
...
+        sy = numpy.linalg.norm(mat[:2, 1])
+        scale_cor = [1 / sx, 1 / sy]
+
+        # Compute rotation
+        rotation_cor = numpy.arctan2(mat[1, 0] / sx, mat[0, 0] / sx)
+
+        # Shear correction
+        a, b = mat[0, 0], mat[0, 1]
+        c, d = mat[1, 0], mat[1, 1]
+        shear_cor = numpy.arctan2(a * c + b * d, scale_cor[0] * scale_cor[1])
```
pieleric, line 81: This whole block is the code from `extract_transformation_parameters()`. Of course, the result will be equal. 
I think that instead of check the exact values, it'd make more sense to just check that the metadata looks correct (ie, it's present, and it's a float or a tuple of 2 floats).

As you have a specific dataset (FOV1 green & blue), you can also compare the values to a set of hardcoded values.

plugins/test/chromatic_correction_test.py
```diff
@@ -0,0 +1,87 @@
...
+        im = self.das[1][0]
+        im.metadata.update({model.MD_PIXEL_SIZE: (0.07738095984842844 * 1e-6, 0.07738095984842844 * 1e-6)})
+
+        # convert m to pixel units
+        max_dist_pxl = int(max_dist / im.metadata[model.MD_PIXEL_SIZE][0])
+
+        # Find transformation values per channel
+        points_ref = MaximaFind(ref, BEAD_QUANTITY, max_dist_pxl)
+        points_im = MaximaFind(im, BEAD_QUANTITY, max_dist_pxl)
+        _ = get_chromatic_correction_dict(im, points_ref, points_im)
```
pieleric, line 53: If you are testing this function, you shouldn't discard the return value. That return value should be checked.

pieleric, line 53: As second test could be to pass points_ref & points_ref as argument. The computed metadata should be the identity.

pieleric (changes requested):

K4rishma:
> When I run the plugin, it straight away closes the application:
> 
> ```
> python3: ../../src/xcb_io.c:374: poll_for_response: Assertion `!xcb_xlib_threads_sequence_lost' failed.
> /usr/bin/odemis-gui: line 8: 24007 Aborted                 ${PYTHON_INTERPRETER:-/usr/bin/python3} -m $GUI "$@"
> WARNING:root:The GUI closed with the error code 134.
> ```
> 
> Maybe introduced by the recent addition of threading. We can schedule some time to figure this out.
> 
> UPDATE, the ShowModal() call in the thread hangs and does not show a modal for me. We get this log though:
> 
> ```
> box.ShowModal()
> Evaluating: box.ShowModal() did not finish after 3.00 seconds.
> This may mean a number of things:
> - This evaluation is really slow and this is expected.
>     In this case it's possible to silence this error by raising the timeout, setting the
>     PYDEVD_WARN_EVALUATION_TIMEOUT environment variable to a bigger value.
> 
> - The evaluation may need other threads running while it's running:
>     In this case, it's possible to set the PYDEVD_UNBLOCK_THREADS_TIMEOUT
>     environment variable so that if after a given timeout an evaluation doesn't finish,
>     other threads are unblocked or you can manually resume all threads.
> 
>     Alternatively, it's also possible to skip breaking on a particular thread by setting a
>     `pydev_do_not_trace = True` attribute in the related threading.Thread instance
>     (if some thread should always be running and no breakpoints are expected to be hit in it).
> 
> - The evaluation is deadlocked:
>     In this case you may set the PYDEVD_THREAD_DUMP_ON_WARN_EVALUATION_TIMEOUT
>     environment variable to true so that a thread dump is shown along with this message and
>     optionally, set the PYDEVD_INTERRUPT_THREAD_TIMEOUT to some value so that the debugger
>     tries to interrupt the evaluation (if possible) when this happens.
> ```
> 
> UPDATE 2: After wrapping the gui altering functions with `@call_in_wx_main`, it starts to work for me, as explained [here](https://github.com/delmic/odemis/blob/master/doc/develop/extending.rst#adding-a-feature-to-the-graphical-user-interface).

Are you still getting this error?

tmoerkerken:
> > When I run the plugin, it straight away closes the application:
> > ```
> > python3: ../../src/xcb_io.c:374: poll_for_response: Assertion `!xcb_xlib_threads_sequence_lost' failed.
> > /usr/bin/odemis-gui: line 8: 24007 Aborted                 ${PYTHON_INTERPRETER:-/usr/bin/python3} -m $GUI "$@"
> > WARNING:root:The GUI closed with the error code 134.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Maybe introduced by the recent addition of threading. We can schedule some time to figure this out.
> > UPDATE, the ShowModal() call in the thread hangs and does not show a modal for me. We get this log though:
> > ```
> > box.ShowModal()
> > Evaluating: box.ShowModal() did not finish after 3.00 seconds.
> > This may mean a number of things:
> > - This evaluation is really slow and this is expected.
> >     In this case it's possible to silence this error by raising the timeout, setting the
> >     PYDEVD_WARN_EVALUATION_TIMEOUT environment variable to a bigger value.
> > 
> > - The evaluation may need other threads running while it's running:
> >     In this case, it's possible to set the PYDEVD_UNBLOCK_THREADS_TIMEOUT
> >     environment variable so that if after a given timeout an evaluation doesn't finish,
> >     other threads are unblocked or you can manually resume all threads.
> > 
> >     Alternatively, it's also possible to skip breaking on a particular thread by setting a
> >     `pydev_do_not_trace = True` attribute in the related threading.Thread instance
> >     (if some thread should always be running and no breakpoints are expected to be hit in it).
> > 
> > - The evaluation is deadlocked:
> >     In this case you may set the PYDEVD_THREAD_DUMP_ON_WARN_EVALUATION_TIMEOUT
> >     environment variable to true so that a thread dump is shown along with this message and
> >     optionally, set the PYDEVD_INTERRUPT_THREAD_TIMEOUT to some value so that the debugger
> >     tries to interrupt the evaluation (if possible) when this happens.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > UPDATE 2: After wrapping the gui altering functions with `@call_in_wx_main`, it starts to work for me, as explained [here](https://github.com/delmic/odemis/blob/master/doc/develop/extending.rst#adding-a-feature-to-the-graphical-user-interface).
> 
> Are you still getting this error?

No, the only deadlock I get is on cancel (see other comment). The correction process works, but just not the cancel part.

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,355 @@
...
+        # Acquire the Maximum Intensity Projection for the given streams
+        levels = generate_zlevels(tab_data.main.focus, Z_RANGE, Z_STEP_SIZE)
+        streams_list = [s for s in tab_data.streams.value
+                        if isinstance(s, FluoStream)]
+        zlevels = {stream: levels for stream in streams_list}
+        settings_observer = tab_data.main.settings_obs
+        self.acqui_task = acquireZStack(streams_list, zlevels, settings_obs=settings_observer)
+
+        try:
+            da_list, _ = self.acqui_task.result()
```
pieleric, line 322: Instead of blocking until the end of the acquisition, you can start this task from the GUI thread, and then `add_done_callback()` to call a separate function that will handle the end of the task (either completed, or cancelled).

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,355 @@
...
+        self.dlg.buttons[0].SetLabel("Cancelling...")
+        self.dlg.buttons[0].Disable()
+        self.dlg.lbl_description.SetLabel("Cancellation in progress\nMay take some time to "
+                                          "cancel the already running processes")
+        self.dlg.Destroy()
+        if self.acqui_task:
+            self.acqui_task.cancel()
+        logging.debug("Dialog closed")
+
+    def _on_start(self, evt):
```
pieleric, line 297: For the plugin buttons, the callback takes the AcquisitionDialog as argument (so `evt` -> `dlg`), and runs *in a separate thread*.
So you shouldn't do anything related with the GUI directly in this function, or protect it with `call_in_wx_main()`. That second option is probably what you want.

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,355 @@
...
+        """Start or cancel the correction method when the button is clicked"""
+        # If correction is running, cancel it, otherwise start one
+        if self.dlg.buttons[0].Label == "Start":
+            self.cancelled = False
+            self._start_chromatic_correction()
+        else:
+            self.on_close(evt)
```
pieleric, line 304: Instead of `evt`, you probably can pass `None`.

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,355 @@
...
+                   "3. No two streams should be the same.")
+            self.dlg.lbl_description.SetLabel(txt)
+            self.dlg.addButton("Manually fix")
+            self.dlg.ShowModal()
+            return
+        self.dlg.Bind(wx.EVT_CLOSE, self.on_close)
+        self.dlg.addButton("Start", self._on_start, face_colour='blue')
+        self.dlg.ShowModal()
+
+    def on_close(self, evt, *args, **kwargs):
```
pieleric, line 287: Why `*args, **kwargs`? I don't think they are needed. It seems the caller either is a event binder (only passes `evt`), or from `_on_start()` (no real argument at all).

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,355 @@
...
+                                     ("Ready to start Chromatic aberration on the given channel.\n"
+                                      "The process will take about 5 minutes\n\n"
+                                      "Z stack range is {} m\n"
+                                      "Z step size of {} m".format(Z_RANGE, Z_STEP_SIZE)))
+        if (len(tab_data.streams.value) <= 1) or (not check_ref_channel) or (not check_unique_channels):
+            txt = ("Problems\n1. Add minimum of two streams.\n2. One of the streams must be green channel.\n"
+                   "3. No two streams should be the same.")
+            self.dlg.lbl_description.SetLabel(txt)
+            self.dlg.addButton("Manually fix")
+            self.dlg.ShowModal()
```
pieleric, line 281: Can you explain why using the AcquisitionDialog here? Why not checking these conditions before creating the dialog, and to display an error just use `wx.MessageDialog`? (see remove_spike.py for an example)

plugins/test/chromatic_correction_test.py
```diff
@@ -0,0 +1,147 @@
...
+        mat = numpy.eye(3)
+        mat[:2, :2] = affine.matrix
+        mat[:2, 2] = affine.translation
+        tx = mat[0, 2]
+        ty = mat[1, 2]
+        # Projection image y-axis is upside down, add negative sign
+        translation = [tx * PIXEL_SIZE[0], -ty * PIXEL_SIZE[1]]
+
+        i = 0
+        for image in [ref, modified_im, modified_im, modified_im]:
```
pieleric, line 91: This `for` loop does 4 completely different things for each step. => Just unroll the loop as a single code!

pieleric (changes requested):

plugins/chromatic_correction.py
K4rishma: I looked into timelapse.py and modified the code. It takes dlg as input argument. Changed self.dlg to dlg, without the use of the decorator. Would that be a fine idea?

K4rishma: When I protect _on_start with call_in_wx_main(), it seems to block the functioning by popping up Odemis is not responding message box

K4rishma: I have added the decorator on on_close(), as it was losing reference of the dialog box

plugins/chromatic_correction.py
K4rishma: I have added the callback and decorated it.

K4rishma:
I get this warning and I am not able to dissolve it
2024-10-04 16:14:05,008 WARNING main:427:       /home/dev/.local/share/odemis/plugins/chromatic_correction.py:295: Warning: invalid (NULL) pointer instance
  self.dlg.ShowModal()

2024-10-04 16:14:05,009 WARNING main:427:       /home/dev/.local/share/odemis/plugins/chromatic_correction.py:295: Warning: g_signal_handler_disconnect: assertion 'G_TYPE_CHECK_INSTANCE (instance)' failed
  self.dlg.ShowModal()

plugins/chromatic_correction.py
K4rishma: added message dialog

src/odemis/util/linalg.py
```diff
@@ -262,3 +262,7 @@ def generate_triangulation_points(max_dist: float, area_coords: Iterable[float])
         points = matrix.tolist()
 
     return points
+
+
+def check_overlap(range1: list, range2: list) -> bool:
```
pieleric, line 267: Please specify the type of the list: `List[float]`. 

Also, add a docstring explaining what this function computes.

pieleric, line 267: "check" is not super clear. I think it'd be nicer to read the calling code if the function is named like `ranges_overlap()`

plugins/chromatic_correction.py
```diff
@@ -0,0 +1,372 @@
...
+    __version__ = "1.0"
+    __author__ = "Karishma Kumar"
+    __license__ = "GPLv2"
+
+    def __init__(self, microscope, main_app):
+        if "cryosecom-localization" in main_app.main_data.tab.choices.values():
+            super().__init__(microscope, main_app)
+            # Green channel is selected as reference channel, as it lies in the middle of the spectrum
+            # All the other channels will be corrected according to the reference colored channel
+            self.ref_emission = [500.e-9, 530.e-9]  # wavelength in meters, green as reference channel
```
pieleric, line 174: Please mark it as a constant, defined outside of the class.

pieleric (changes requested):

pieleric (approved):

plugins/test/chromatic_correction_test.py
```diff
@@ -0,0 +1,142 @@
...
+
+    @classmethod
+    def setUpClass(cls):
+        cls.das = read_data('FOV1_green.tif')
+
+    def test_chromatic_correction_with_fake_data_compare_methods(self):
+        """Given two images in which one is reference image and other is affine transformed image of reference image.
+         The calculated affine parameters should correct the transformed image parameters back to the reference image
+         Two methods namely RSU and RSL from odemis.util.transform.py are compared to show which method yields
+         good results"""
```
tepals, line 57: it should not be the task of a test case to show with method yields good results. What is the goal of this test case?

plugins/test/chromatic_correction_test.py
```diff
@@ -0,0 +1,142 @@
...
+from plugins.chromatic_correction import (BEAD_QUANTITY, SCALE, TOLERANCE,
+                                          find_corresponding_points, get_chromatic_correction_dict)
+
+PIXEL_SIZE = (1e-6, 1e-6)  # m
+# Affine parameters that are used on reference image
+# Translation can be any value. Scale is fixed to one if there is some rotation and shear else both scale, shear
+# can be changed by keeping the rotation to zero
+SHEAR = 0.9
+SCALE = [2, 3]
+ROTATION = numpy.radians(0)
```
tepals, line 35: why test with a rotation of 0?

plugins/test/chromatic_correction_test.py
tepals: and if it is to show which method yields good results, can you add an explanation as to how you can check which method yields good results?

tepals (changes requested):

plugins/test/chromatic_correction_test.py
```diff
@@ -0,0 +1,142 @@
...
+        modified_im.metadata.update({model.MD_SHEAR: SHEAR})
+
+        # Generate synthetic keypoints of reference and modified images based on given affine parameters
+        r_m = _rotation_matrix_from_angle(ROTATION)
+        # Make sure the given points are not in one line
+        points1 = numpy.array([[5, 5], [5, 10], [10, 8], [11, 11]])
+        shear_m = numpy.array([[1, SHEAR], [0, 1]])
+        points2 = TRANSLATION + SCALE * ((r_m @ (shear_m @ points1.transpose())).transpose())
+        print(f"points 1 {points1}")
+        print(f"points 2 {points2}")
```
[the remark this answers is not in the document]
K4rishma: yes, my bad

plugins/test/chromatic_correction_test.py
K4rishma: I have added two sets of affine parameters, one with rotation 0 and one with non-zero such that the the mix of shear, rotation, scale keeps it affine.

plugins/test/chromatic_correction_test.py
K4rishma: I have removed the comparison to compare the good methods, as rightly said by you that the test cases are not for the purpose of comparing which method is the best. With that, I have made the testing of chosen method better by testing it in two different  situations by altering the given affine parameters.

plugins/chromatic_correction.py
tepals: you resolved this, but I do not see the unit

plugins/test/chromatic_correction_test.py
```diff
@@ -0,0 +1,129 @@
...
+        # Set 1: change scale and sheer, keeping the rotation unchanged
+        # Set 2: change the rotation and sheer, keeping the scale unchanged
+        affine_sets = {"set_1": {"shear": 0.9, "scale": [2, 3], "rotation": numpy.radians(0), "translation": [2, 1]},
+                       "set_2": {"shear": 0.4, "scale": [1, 1], "rotation": numpy.radians(5), "translation": [2, 3]}}
+
+        for affine_set in affine_sets.values():
+            SHEAR = affine_set["shear"]
+            SCALE = affine_set["scale"]
+            ROTATION = affine_set["rotation"]
+            TRANSLATION = affine_set["translation"]
```
tepals, line 67: is there a reason to use all capital letters here and not `shear, scale, rotation, translation`?

plugins/test/chromatic_correction_test.py
```diff
@@ -0,0 +1,129 @@
...
+            ty = mat[1, 2]
+            # Projection image y-axis is upside down, add negative sign
+            translation = [tx * PIXEL_SIZE[0], -ty * PIXEL_SIZE[1]]
+            scale_u, rotation_u, shear_u = alt_transformation_matrix_to_implicit(affine.matrix, "RSU")
+            # Test 1, the correction metadata is same as the computed transformation values
+            scale_cor = [1 / scale_u[0], 1 / scale_u[1]]
+            modified_im.metadata.update({model.MD_ROTATION_COR: rotation_u})  # rotation corr
+            modified_im.metadata.update({model.MD_POS_COR: translation})  # translation correction
+            modified_im.metadata.update({model.MD_PIXEL_SIZE_COR: scale_cor})  # scale correction
+            modified_im.metadata.update({model.MD_SHEAR_COR: shear_u})  # shear corr
```
tepals, line 97: why are you updating the metadata here?

tepals (changes requested):

tepals (approved):

pieleric:
@K4rishma , could you squash the last commit into the previous one. Once done, I'll merge the PR.

PR #2802, opened 2024-06-19
[TRCL-339] support fslt in the alignment tab

sneepstefan:
The goal is to provide a nice UI with enough feedback for the user to adjust the alignment to one providing good acquisition conditions, and being able to confirm this alignment is (still) correct.
This is only for the “fine alignment”, which is to be done at regular intervals, either after system start, or before an acquisition. In particular, it does NOT aim to support the first rough alignment.


src/odemis/gui/cont/tabs/sparc2_align_tab.py
```diff
@@ -934,13 +1059,27 @@ def _onAlignMode(self, mode):
             self.panel.pnl_mirror.Show(False)
             self.panel.pnl_lens_mover.Show(False)
             self.panel.pnl_lens_switch.Show(False)
-            self.panel.pnl_focus.Show(True)
+            if main.spectrograph_ded and main.streak_ccd.name in main.spectrograph_ded.affects.value:
+                self.panel.pnl_focus_ext.Show(True)
+                self.panel.pnl_focus.Show(False)
+                self.panel.cmb_focus_detectors_ext.SetSelection(1)
```
nandishjpatel, line 1065:
```suggestion
                self.panel.cmb_focus_detectors_ext.SetSelection(0)
```
Maybe?

src/odemis/gui/cont/tabs/sparc2_align_tab.py
```diff
@@ -1327,6 +1533,65 @@ def _onManualFocus(self, event):
...
+        align_mode = self.tab_data_model.align_mode.value
+        gauge = self.panel.gauge_autofocus_ext
+        self._mf_future.cancel()  # In case it's still changing, immediately stop (the gauge)
+
+        if event.GetEventObject().GetValue():  # manual focus btn toggled
+            # Set the optical path according to the align mode
+            if align_mode == "streak-align":
+                opath = "streak-focus"
+            elif align_mode == "tunnel-lens-align":
+                opath = "spec-focus-ext"
```
pieleric, line 1552: For now the  Sparc2ManualFocus (or Sparc2AutoFocus) doesn't know about "spec-focus-ext". => add it.

src/odemis/acq/align/autofocus.py
```diff
@@ -834,6 +843,11 @@ def updateProgress(subf, start, end):
         ret = {}
         logging.debug("Running AutoFocusSpectrometer on %s, using %s, for the detectors %s, and using selector %s",
                       spgr, focuser, dets, selector)
+        # make sure the slit-in position is at the minimum when using an external spectrograph
+        if spgr.role == 'spectrograph-dedicated':
+            if spgr.position.value["slit-in"] != spgr.axes["slit-in"].range[0]:
+                spgr.moveAbsSync({"slit-in": spgr.axes["slit-in"].range[0]})
+
```
pieleric, line 850: The idea of the `spec-focus` alignment mode is that this sort of things is handled in the optical path manager. That's why there is a different mode from the "mirror-align". Similarly, here, instead of "tunnel-lens-align", I think the proper way is to have a "spec-ext-focus" mode in the path.py, which sets the hardware to the right position. I think it's mostly a copy of "tunnel-lens-align", but the slit-in is at 10e-6 m (like for "spec-focus").

Note this means you have to update the functions in the sparc2_align_tab, onAutoFocus() & onManualFocus().

pieleric (changes requested):

src/odemis/gui/cont/tabs/sparc2_align_tab.py
sneepstefan: I think I misunderstood 'focus_mode' and 'alignment_mode' which are essentially the same but named differently.  Removed spec-focus-ext completely and used tunnel-lens-align everywhere now.

src/odemis/acq/path.py
```diff
@@ -288,7 +290,20 @@
                  'light-aligner': {'x': "MD:" + model.MD_FAV_POS_ACTIVE,
                                    #'z': "MD:" + model.MD_FAV_POS_ACTIVE
                                    },
-                 }),
+                }),
+            'tunnel-lens-align': ("sp-ccd",
+                {'chamber-light': {'power': 'off'},
+                 'spec-selector': {'x': "MD:" + model.MD_FAV_POS_ACTIVE},
```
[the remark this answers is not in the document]
sneepstefan: Where exactly does it need to be 'double checked'? Cannot seem to find a spot where to do that in path.py or should that be done in sparc2_align_tab.py where the selectorsToPath() is called?

src/odemis/acq/align/autofocus.py
sneepstefan: I think I misunderstood 'focus_mode' and 'alignment_mode' which are essentially the same but named differently.  Added spec-focus-ext to path.py.

src/odemis/acq/path.py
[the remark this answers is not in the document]
pieleric: I meant that it should be removed, because it's already done automatically. However, after removing that line, you should run the simulator and confirm that indeed the spec-selector does go to the right position when selecting this optical mode (tunnel-lens-align).

PR #2803, opened 2024-06-23
[FEAT] add correlation tab to odemis viewer

patrickcleeve2:
Users have requested the ability to use the correlation tools (stitching and correlating) overviews in the odemis viewer without being connected to a microscope. This PR adds:

Features:
- Enable the use of the correlation tab in the odemis viewer.

### DIFF
--- a/src/odemis/gui/cont/menu.py
+++ b/src/odemis/gui/cont/menu.py
@@ -149,6 +149,15 @@ class MenuController(object):
         # /Help/About
         main_frame.Bind(wx.EVT_MENU, self._on_about, id=main_frame.menu_item_about.GetId())
 
+        # add a toggle for correlation tab in viewer mode
+        if main_data.is_viewer:
+            main_frame.Bind(wx.EVT_MENU, self._on_show_correlation, id=main_frame.menu_item_show_correlation.GetId())
+        else:
+            # remove the menu item
+            menu = main_frame.menu_item_show_correlation.GetMenu()
+            menu.Remove(main_frame.menu_item_show_correlation)
+            main_frame.menu_item_show_correlation.Destroy()
+
     def _on_update(self, evt):
         import odemis.gui.util.updater as updater
         u = updater.WindowsUpdater()
@@ -475,3 +484,18 @@ class MenuController(object):
         """
         # TODO: use event?
         self._main_data.debug.value = self._main_frame.menu_item_debug.IsChecked()
+
+    def _on_show_correlation(self, evt):
+        """ Toggle display of the correlation tab and tab buttons """
+
+        show = self._main_frame.menu_item_show_correlation.IsChecked()
+
+        if show:
+            self._main_frame.pnl_tabbuttons.Show()
+            tab = self._main_data.getTabByName("meteor-correlation")
+        else:
+            self._main_frame.pnl_tabbuttons.Hide()
+            tab = self._main_data.getTabByName("analysis")
+
+        self._main_data.tab.value = tab
+        self._main_frame.Layout()
--- a/src/odemis/gui/cont/tabs/correlation_tab.py
+++ b/src/odemis/gui/cont/tabs/correlation_tab.py
@@ -242,6 +242,8 @@ class CorrelationTab(Tab):
 
     @classmethod
     def get_display_priority(cls, main_data) -> Optional[int]:
+        if main_data.is_viewer:
+            return 0 # low priority in viewer
         if main_data.role == "meteor":
             return 1
         else:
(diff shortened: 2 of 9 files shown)


pieleric (changes requested):
The code per-se looks fine. However, this has a lot of extra implications on the main GUI level which are not handled. In particular, there has been a lot of work to make sure that the viewer, which has only one tab has the maximum screen real-estate, by hiding the tab bar. This is useful for users which have small laptop screens. This change automatically though this work away.
Also, *most* of the users of the viewer are not METEOR users. They have probably no idea what the correlation tab is for, and also likely no use of it. So this might end up being confusing & counter-productive for many use cases.

So I think it'd much better if we could find a more subtle way to introduce this functionality. Admittedly, I don't have any good idea on how this could be done, but probably one of these ways could be used as starting point:
* Have an option in the menu (or as a button) to show the correlation tab
* When loading a file which has both FM and EM data, show the correlation control buttons directly in the analysis tab (and hide them when the file content doesn't match)
* Have two launchers: one for the standard viewer, and one for the viewer with the correlation tab (but then it doesn't work well in case you open an OME-TIFF file, which runs the viewer)

src/odemis/gui/cont/tabs/correlation_tab.py
```diff
@@ -245,7 +245,9 @@ def get_display_priority(cls, main_data) -> Optional[int]:
         if main_data.microscope is None:
             # NOTE: microscope = None means the Odemis Viewer
             # Therefore, we want to enable the correlation tab
-            return 1
+            # hide tab by default
+
+            return 0
```
pieleric, line 250: `None` means "hidden". 0 is lowest priority. If you want the analysis tab to be higher priority that the correlation tab, you should raise the priority of the analysis tab to 1 (if it's a viewer?).

src/odemis/gui/cont/menu.py
```diff
@@ -149,6 +149,20 @@ def __init__(self, main_data, main_frame):
         # /Help/About
         main_frame.Bind(wx.EVT_MENU, self._on_about, id=main_frame.menu_item_about.GetId())
 
+        is_viewer = main_data.microscope is None
+        if is_viewer:
+            # add toggle for showing correlation tab
+            main_frame.Bind(wx.EVT_MENU, self._on_show_correlation, id=main_frame.menu_item_show_correlation.GetId())
```
pieleric, line 155: Otherwise, hide the menu entry.

pieleric (changes requested):
I think having a menu option would be fine.

patrickcleeve2:
@pieleric added a menu button for toggling correlation tab / tab button bar

pieleric (approved):
Looks fine. Patrick, did you test the correlation tab on Windows? If not yet, it should be done (we can already merge this part). I can build a Windows viewer for you if that helps... but it's probably best that you can start it from the source code, so that you can debug it.

patrickcleeve2:
> Looks fine. Patrick, did you test the correlation tab on Windows? If not yet, it should be done (we can already merge this part). I can build a Windows viewer for you if that helps... but it's probably best that you can start it from the source code, so that you can debug it.

@pieleric No I haven't tested on Windows, it would be great if you could test the build because I haven't done that before

src/odemis/gui/cont/menu.py
```diff
@@ -149,6 +149,15 @@ def __init__(self, main_data, main_frame):
...
+        if main_data.is_viewer:
+            main_frame.Bind(wx.EVT_MENU, self._on_show_correlation, id=main_frame.menu_item_show_correlation.GetId())
+        else:
+            # remove the menu item
+            menu = main_frame.menu_item_show_correlation.GetMenu()
+            menu.Remove(main_frame.menu_item_show_correlation)
+            main_frame.menu_item_show_correlation.Destroy()
+
     def _on_update(self, evt):
         import odemis.gui.util.updater as updater
```
K4rishma, line 162: Not a part of the review, but is it fine to have import statement in between the file?

K4rishma (approved):

src/odemis/gui/cont/menu.py
pieleric: It depends. Usually, you want to have all imports at the top of the file. However, if you want to only import "dynamically", just before a function is needed, then it's useful. I think that here the reasoning is that this function is only useful on Windows, so no need to import the module for the most common use cases (on Linux). IIRC, a long time ago, the import would even fail on Linux, hence it was placed here. Nowadays, that import should work on Linux, and be essentially a no-op, as long as you don't call any functions, so we could move it back to the top. (but it's fine here)

K4rishma: Thank you Eric!

PR #2804, opened 2024-06-24
[feature] keysight driver now check values

pieleric:


src/odemis/driver/test/keysight_test.py
```diff
@@ -107,15 +110,17 @@ def test_delay(self):
...
-        self.assertEqual(self.dev.delay.value, 1e-5)
+        self.assertAlmostEqual(self.dev.delay.value, 1e-5)
         self.dev.delay.value = 1e-6
-        self.assertEqual(self.dev.delay.value, 1e-6)
+        self.assertAlmostEqual(self.dev.delay.value, 1e-6)
+        self.dev.delay.value = -1e-6
+        self.assertAlmostEqual(self.dev.delay.value, -1e-6)
 
         with self.assertRaises(IndexError):
             self.dev.delay.value = self.dev.delay.range[1] + 1
```
K4rishma, line 120: if applicable, we can have a similar test for the other value of range -> delay.range[0]  - 1

src/odemis/driver/keysight.py
```diff
@@ -43,77 +60,125 @@ def __init__(self, name, role, address,
...
+        self._hwVersion = idn
+
+        # Empty the error queue, to avoid any error from the previous session
+        for i in range(32):
+            if self.getErrorState() == 0:
+                break
+        else:
+            logging.warning("Error queue is not empty, will continue anyway")
 
+        if channel not in {1, 2}:
```
nandishjpatel, line 85: `{1, 2}` can be a class constant? its also used in line 91

src/odemis/driver/keysight.py
```diff
@@ -43,77 +60,125 @@ def __init__(self, name, role, address,
...
+            raise HwError(f"Failed to find a device on the address '{address}'. "
+                          f"Check it is turned on and connected to the computer.")
+
+        try:
+            idn = self.getIdentification()
+            # Agilent Technologies,33622A,MY59002437,A.02.03-3.15-03-64-02
+            model_id = idn.split(",")[1]
+            if not re.match(r"33(5|6)...", model_id):
+                self._accesser.close()
+                raise LookupError()
```
nandishjpatel, line 176: Probably have a message using the `model_id` and the match?

src/odemis/driver/keysight.py
```diff
@@ -122,144 +187,259 @@ def _sendCommand(self, cmd):
...
+        act_p = self.getOutput(self._channel)
+        return act_p
+
+    def _update_settings(self):
+        """
+        Update the settings of the device.
+        Called by the repeating timer.
+        """
+        f = self.getFrequency(self._channel)
+        p = 1 / f
```
nandishjpatel, line 416: Will the frequency ever be `0`?

src/odemis/driver/keysight.py
```diff
@@ -459,69 +560,118 @@ def close(self):
...
+            # range when the frequency is high, we use 20-80%.
+            if duty_cycle < 20 or duty_cycle > 80:
                 logging.warning("The duy cycle value is out of bounds: %s given", duty_cycle)
-            else:
-                self._output_states['channel' + channel].duty_cycle = duty_cycle
-        elif re.match(r"sour[1-2]:volt:offs ", msg):
-            channel = msg[4]
-            voltage_offset = float(msg.split()[1])
-            self._output_states['channel' + channel].voltage_offset = voltage_offset
+                self._error = -222
```
nandishjpatel, line 634: I see 3 error values `0, -222, 32`

Is it good to create a custom `errno`?
https://docs.python.org/3/library/errno.html

src/odemis/driver/keysight.py
```diff
@@ -122,144 +187,259 @@ def _sendCommand(self, cmd):
...
+        if f.upper() not in {"SIN", "SQU", "RAMP", "PULS"}:
             raise ValueError("Incorrect waveform: %s given" % (f,))
 
-        self._sendCommand(b"sour%d:appl:%s" % (c, f))
+        self._sendCommand(b"sour%d:appl:%s" % (c, f.encode("ascii")))
+        self._checkError()
 
     def _set_period(self, p: float) -> float:
-        f = 1/p
+        f = 1 / p
```
K4rishma, line 361: To be safe, can we make sure p>0

src/odemis/driver/test/keysight_test.py
```diff
@@ -107,15 +110,21 @@ def test_delay(self):
         """
         self.dev.period.value = 1e-4
         self.dev.delay.value = 1e-5
-        self.assertEqual(self.dev.delay.value, 1e-5)
+        self.assertAlmostEqual(self.dev.delay.value, 1e-5)
         self.dev.delay.value = 1e-6
-        self.assertEqual(self.dev.delay.value, 1e-6)
+        self.assertAlmostEqual(self.dev.delay.value, 1e-6)
+        self.dev.delay.value = -1e-6
+        self.assertAlmostEqual(self.dev.delay.value, -1e-6)
```
K4rishma, line 117: if the delay time calculation depends on the sign of delay phase, maybe we can add test cases for that?

src/odemis/driver/keysight.py
pieleric: It's already taken care of by the VA. I'll add comment.

src/odemis/driver/keysight.py
pieleric: No, the device doesn't accept a frequency of 0, and never returns such value. I'll add a comment.

src/odemis/driver/keysight.py
pieleric: They are not relate to the standard C library error codes. They are the codes from the Keysight device. The manual has the complete list (~300 values). However, as there is a function that can return a "english translation" (`getSystemError()`), I would rather not fill up the driver with these codes. I'll just put a comment next to these numbers in the simulator to explain what they mean.

src/odemis/driver/test/keysight_test.py
pieleric: What do you mean more precisely? The phase is just proportional to the delay (+ period). It's automatically computed by the device. So it's pretty simple. This part of the test case checks that it's possible to put a negative value. What would you check more?

src/odemis/driver/keysight.py
```diff
@@ -122,144 +187,259 @@ def _sendCommand(self, cmd):
...
+        if p != self.period.value:
+            self.period._set_value(p)  # Avoid the setter
+
+        # Duty cycle
+        dc = self.getDutyCycle(self._channel) / 100
+        if dc != self.dutyCycle.value:
+            self.dutyCycle._set_value(dc)
+
+        # Delay
+        act_phase = self.getPhase(self._channel)
```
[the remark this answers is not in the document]
pieleric: A negative phase means a negative delay. So nothing special to do.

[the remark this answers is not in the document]
K4rishma: I am not sure what negative delay means. I understand the delay to be a positive value that adds up to delay the received signal by that much time. If the delay is negative, how is the timing of the signal interpreted ? I thought negative phase for e.g. -30 would be converted to positive value of 330. Then the delay will be based on 330. But I guess this is not the case.  

My comment in the test case was based on the above interpretation. If the formulation of positive and negative phase delay differs, we can test the formulation.

K4rishma (approved):

src/odemis/driver/test/keysight_test.py
K4rishma: I am not sure what negative delay means. I understand the delay to be a positive value that adds up to delay the received signal by that much time. If the delay is negative, how is the timing of the signal interpreted ? I thought negative phase for e.g. -30 would be converted to positive value of 330. Then the delay will be based on 330. But I guess this is not the case.

My comment in the test case was based on the above interpretation. If the formulation of positive and negative phase delay differs, we can test the formulation.

pieleric: You're right, the delay in sec in converted into a phase. You are also right that a phase of -30° == +330°.
The device supports both. It accepts -360° to 360°. The driver also supports both "directions" for the delay. So we just pass it as-is: if the user talks about a negative delay, we convert to a negative phase. If the user talks about positive delay, we convert to positive phase.

pieleric: I've added an extra comment next to the VA to clarify that both negative and positive values are fine. The docstrings of both getPhase() and setPhase() already indicate that both signs are valid.

nandishjpatel (approved):

PR #2805, opened 2024-06-24
[fix] omicronxx handle device reporting 0xff bytes in the GSI command

pieleric:
Normally the GSI answer looks like (wl in nm) § int (power in mW).
However, for one source of one device, the answer is:
(wl in nm) 0xff 0xff § int (power in mW)

That's not defined in the documentation, but the official Omicron program
handles it fine... so we now also handle it fine.

### DIFF
--- a/src/odemis/driver/omicronxx.py
+++ b/src/odemis/driver/omicronxx.py
@@ -418,10 +418,12 @@ class DevxX(object):
             subdev (set of int): subdevices available
         """
         ans = self._getValue("GSI")
-        # Expects something like:
-        # GSI [m63] (optional) int (wl in nm) § int (power in mW)
+        # Follows this pattern:
+        # GSI [m int] (available sources, optional) int (wl in nm) § int (power in mW)
+        # So something like: GSI[m63]550§1564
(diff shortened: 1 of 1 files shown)


pieleric:
@tepals , this should go to the stable version.

src/odemis/driver/omicronxx.py
```diff
@@ -420,8 +420,9 @@ def GetSpecInfo(self):
         ans = self._getValue("GSI")
         # Expects something like:
         # GSI [m63] (optional) int (wl in nm) § int (power in mW)
```
tepals, line 422: I tested both patterns with this example string, and no matches were found. Am I misinterpreting this string? I tried: `"GSI [m63] (optional) int (wl in nm) § int (power in mW)"`, `"GSI [m63] int 900 § int 900"` and `"GSI [m63] int (900) § int (900)"`

pieleric: No, it wasn't meant to be taken so literally. There is no space and "int" has to be replaced by a number. Like:
`GSI[m63]550§1564`
I can add this example, if you think that helps :-)

tepals (approved):

sneepstefan (approved):

PR #2806, opened 2024-06-26
[feature][meteor] show active microscope name in dialog window when backend is already started when using odemis-select-mic-start

sneepstefan:
If the backend is already running, show the current active configuration (microscope) name in the user dialog window.


pieleric (approved):
I like the `$'\t'`! (I had no idea about this)

nandishjpatel (approved):

PR #2807, opened 2024-06-26
[feat] Fastem acquisitions: Reduce the number of times the beamshift …

tepals:
…correction runs

The beamshift correction was running for every field, it takes about 0.46s to run, so reducing the number of times it runs increases the throughput significantly.

### DIFF
--- a/src/odemis/acq/fastem.py
+++ b/src/odemis/acq/fastem.py
@@ -627,12 +627,13 @@ class AcquisitionTask(object):
         Acquire the single field images that resemble the region of acquisition (ROA, megafield image).
         :param dataflow: (model.DataFlow) The dataflow on the detector.
         """
+        beam_shift_indices = self._calculate_beam_shift_cor_indices()
 
         total_field_time = self._detector.frameDuration.value + 1.5  # there is about 1.5 seconds overhead per field
         # The first field is acquired twice, so the timeout must be at least twice the total field time.
         # Use 5 times the total field time to have a wide margin.
         timeout = 5 * total_field_time + 2
-
+        beam_shift_failed = False
         # Acquire all single field images, which are automatically offloaded to the external storage.
         for field_idx in self._roa.field_indices:
             # Reset the event that waits for the image being received (puts flag to false).
@@ -642,16 +643,24 @@ class AcquisitionTask(object):
 
             self.move_stage_to_next_tile()  # move stage to next field image position
             self._scanner.blanker.value = False  # unblank the beam
-            # TODO remove the below temporary fix when a proper solution is found
-            # The temporary fix has been applied so that the ROAs acquisition can still continue upon failure
-            # of correcting the beam shift
-            try:
-                self.correct_beam_shift()  # correct the shift of the beams caused by the parasitic magnetic field.
-            except Exception:
-                logging.exception("Correcting the beam shift failed, check if the image quality is still good.")
-                # In case of failure save the ccd image
-                ccd_image = self._ccd.data.get(asap=False)
-                fastem_util.save_image(self.path, f"{self.field_idx}_after.tiff", ccd_image)
+
+            prev_beam_shift = self._beamshift.shift.value
+            if field_idx in beam_shift_indices or beam_shift_failed:
+                logging.debug(f"Will run beam shift correction for field index {field_idx}")
+                try:
+                    new_beam_shift = self.correct_beam_shift()
+                    # The difference in x or y should not be larger than 2 micrometers
+                    if any(map(lambda x, y: abs(x - y) > 2e-6, new_beam_shift, prev_beam_shift)):
+                        raise ValueError(
+                            "Difference in beam shift is larger than 2 µm, therefore it most likely failed."
+                        )
+                    beam_shift_failed = False
+                except Exception:
+                    logging.exception("Correcting the beam shift failed, check if the image quality is still good.")
+                    # In case of failure save the ccd image
+                    ccd_image = self._ccd.data.get(asap=False)
+                    fastem_util.save_image(self.path, f"{self.field_idx}_after.tiff", ccd_image)
+                    beam_shift_failed = True
 
             dataflow.next(field_idx)  # acquire the next field image.
 
@@ -838,6 +847,8 @@ class AcquisitionTask(object):
         beams are roughly centered on the mppc detector. Using the difference between the current beam positions and the
         good beam positions we calculate in what direction and how much to shift beams, such that they are always
         centered on the mppc detector.
+
+        :return: (float, float) the value of the beam shift after correction
         """
         pixel_size = self._ccd.pixelSize.value
         magnification = self._lens.magnification.value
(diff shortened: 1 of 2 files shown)


tepals:
@nandishjpatel the exact value for how often it needs to run still needs to be determined, but can you do an initial review already?

src/odemis/acq/fastem.py
```diff
@@ -632,7 +632,7 @@ def acquire_roa(self, dataflow):
         # The first field is acquired twice, so the timeout must be at least twice the total field time.
         # Use 5 times the total field time to have a wide margin.
         timeout = 5 * total_field_time + 2
-
+        bs_fail = False
```
pieleric, line 635: "bs" is not a great abbreviation to use. Let's explicitly write "beam_shift" or "beamshift".
I'd even call it "beam_shift_fail*ed*" to clarify a little further the meaning. Or you could even invert it and call it "force_beam_shift_update".

src/odemis/acq/fastem.py
```diff
@@ -645,13 +645,19 @@ def acquire_roa(self, dataflow):
...
-                self.correct_beam_shift()  # correct the shift of the beams caused by the parasitic magnetic field.
-            except Exception:
-                logging.exception("Correcting the beam shift failed, check if the image quality is still good.")
-                # In case of failure save the ccd image
-                ccd_image = self._ccd.data.get(asap=False)
-                fastem_util.save_image(self.path, f"{self.field_idx}_after.tiff", ccd_image)
+            # field indices are saved (col, row), run the beamshift correction for every 4th column.
+            # TODO value 4 needs to be updated based on experimental testing
+            if field_idx[0] % 4 == 0 or bs_fail:
+                logging.debug(f"Will run beamshift correction for field index {field_idx}")
```
pieleric, line 651: Please spell "beam shift", with or without a space, the same way everywhere (it's easier when searching the logs).

pieleric (changes requested):

src/odemis/acq/fastem.py
```diff
@@ -900,6 +906,32 @@ def _create_acquisition_metadata(self):
         self._detector.updateMetadata({model.MD_EXTRA_SETTINGS: json.dumps(selected_settings)})
         return selected_settings
 
+    def _calculate_beam_shift_cor_indices(self, n_beam_shifts=10):
```
nandishjpatel, line 909: I think this calculation can go to FastEMROA to reduce the overhead during acquisition

tepals: It's pretty fast so I decided to put it here to reduce overhead during drawing, and then it does not have to be recalculated when a shape is resized. Let me know if you still think that I should move it to FastEMROA

src/odemis/acq/fastem.py
```diff
@@ -900,6 +906,32 @@ def _create_acquisition_metadata(self):
...
+        """
+        Calculate for which indices to run the beam shift correction. The beam shift correction should run every
+        n sections, starting at the first field in a row. For polygonal sections there can be gaps in the indices,
+        the beam shift correction should then be run on the next possible section.
+        Example: If the beam shift correction should run every 3 sections for the indices [(1, 0), (2, 0), (8, 0)],
+        it should run for indices (1, 0) and (8, 0).
+
+        :param n_beam_shifts: (int) Every how many sections the beam shift correction should run.
```
K4rishma, line 917:
```suggestion
        :param n_beam_shifts: (int) Number of sections after which the beam shift correction should run.
```

K4rishma (approved):

K4rishma (approved):

src/odemis/acq/fastem.py
```diff
@@ -642,16 +646,24 @@ def acquire_roa(self, dataflow):
...
-                ccd_image = self._ccd.data.get(asap=False)
-                fastem_util.save_image(self.path, f"{self.field_idx}_after.tiff", ccd_image)
+
+            prev_beam_shift = self._beamshift.shift.value
+            if field_idx in beam_shift_indices or beam_shift_failed:
+                logging.debug(f"Will run beam shift correction for field index {field_idx}")
+                try:
+                    new_beam_shift = self.correct_beam_shift()
+                    # The difference in x or y should not be larger than 2 micrometers
+                    if any(map(lambda x, y: abs(x - y) > 2e-6, new_beam_shift, prev_beam_shift)):
```
pieleric, line 656: The names of the variables are confusing. It feels you're comparing X with Y. x & y refer to the new and previous values. This is computed for all axes, both X and Y axes. Right?
```suggestion
                    if any(map(lambda a, b: abs(a - b) > 2e-6, new_beam_shift, prev_beam_shift)):
```
(or maybe name them "n" and "p"?)

pieleric: Or maybe just compare the complete distance, with something like:
```suggestion
                    if math.hypot(new_beam_shift[0] prev_beam_shift[0], new_beam_shift[1] - prev_beam_shift[1]) >  2e-6:
```

src/odemis/acq/fastem.py
```diff
@@ -642,16 +646,24 @@ def acquire_roa(self, dataflow):
...
+                        raise ValueError(
+                            "Difference in beam shift is larger than 2 µm, therefore it most likely failed."
```
pieleric, line 658: Can you add the previous and new beam shift values to the log, so that it's easier to see how bad it went (or maybe there's something fishy in the values).

src/odemis/acq/fastem.py
```diff
@@ -900,6 +915,32 @@ def _create_acquisition_metadata(self):
...
+        :param n_beam_shifts: (int) Number of sections after which the beam shift correction should run.
+        """
+        # Sort indices by row first and then by column to enable row-wise processing
+        field_indices = sorted(self._roa.field_indices, key=lambda x: (x[1], x[0]))
+
+        beam_shift_indices = []
+        current_row = -1  # Initialize with an invalid row number to handle the first row properly
+        for idx in field_indices:
+            col, row = idx  # field indices are saved (col, row)
+            if row == current_row + 1:
```
pieleric, line 935: Can we just write `if row > current_row`? As the fields are sorted by row, and the whole megafield should be a single shape, I think that should be equivalent, but it feels like clearer to me, and also "safer" in (the impossible?) case the rows would "jump" by more than 1.

src/odemis/acq/fastem.py
```diff
@@ -900,6 +915,32 @@ def _create_acquisition_metadata(self):
...
+        Example: If the beam shift correction should run every 3 sections for the indices [(1, 0), (2, 0), (8, 0)],
+        it should run for indices (1, 0) and (8, 0).
+
+        :param n_beam_shifts: (int) Number of sections after which the beam shift correction should run.
+        """
+        # Sort indices by row first and then by column to enable row-wise processing
+        field_indices = sorted(self._roa.field_indices, key=lambda x: (x[1], x[0]))
+
+        beam_shift_indices = []
+        current_row = -1  # Initialize with an invalid row number to handle the first row properly
```
pieleric, line 932: "invalid row number" is not just that. It *has* to be -1, and the first row *must* be 0 (unless you use the ">" suggestion). If that's not the case, the `elif` will fail as `beam_shift_indices` is empty.

pieleric (changes requested):

src/odemis/acq/fastem.py
tepals: The names are indeed confusing, I'll leave it as a per axis check though, during testing we always see that the distance in one direction is high, and in the other not. So I think it is a safe check.

nandishjpatel (approved):

pieleric (approved):

PR #2809, opened 2024-06-28
[feature] add support to backend for child component to have depencencies or power supplier

pieleric:
A component can only be initialised if its dependencies are ready.
If it has children, the dependencies of the children must also be ready.
That wasn't taken into account.

Note that no microscope file ever had such complexity... until
yesterday.

=> Also check for the dependencies of children component (including the
"power supplier" which is just a special type of dependency)

### DIFF
--- a/src/odemis/odemisd/modelgen.py
+++ b/src/odemis/odemisd/modelgen.py
@@ -679,6 +679,12 @@ class Instantiator(object):
         if "power_supplier" in args:
             f = args["power_supplier"].supply({name: True})
             f.result()
+        # Also turn on the children
+        for child_args in args.get("children", {}).values():
+            if "power_supplier" in child_args:
+                child_name = child_args["name"]
+                f = child_args["power_supplier"].supply({child_name: True})
+                f.result()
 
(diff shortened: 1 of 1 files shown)


src/odemis/odemisd/modelgen.py
```diff
@@ -679,6 +679,12 @@ def _instantiate_comp(self, name):
         if "power_supplier" in args:
             f = args["power_supplier"].supply({name: True})
             f.result()
+        # Also turn on the children
```
patrickcleeve2, line 682: Is it always safe to turn on these components?

patrickcleeve2 (approved):

src/odemis/odemisd/modelgen.py
pieleric: Usually it is. But, in any case, that's the responsibility of the person writing the microscope file. In the microscope file, it is said "to be able to use this component, it must first be turned on using this power supplier". So this new code just ensures that we follow the rules laid down in the microscope file.

nandishjpatel (approved):

PR #2810, opened 2024-06-28
[ci] Add cloud hosted CI on Ubuntu 24.04

tepals:
The unit tests do not pass yet on Ubuntu 24.04. This ci-workflow can be used to figure out what we need to test and verify fixes. It will not send a notification for a failed run, but if you click on the specific github action you will see a failure for the specific build. I created a separate workflow, because it is not possible to integrate in the regular workflow such that you get a notification for failures on Ubuntu 20/22 but not on Ubuntu 24.

### DIFF
--- a/.github/workflows/unit_tests_cloud_ubuntu24.yml
+++ b/.github/workflows/unit_tests_cloud_ubuntu24.yml
@@ -1,0 +1,189 @@
+# This workflow will install Python dependencies, and run a selection of tests from the Odemis repository.
+# The selection is picked such that they do not require a back-end and are known to pass with this github workflow.
+# Test cases that will pass with this workflow can be added to this document.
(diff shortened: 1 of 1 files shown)


sneepstefan:
Do we keep a separate file for the cloud runner for Ubuntu 24.04 or do we eventually add 24.04 to the OS matrix at the beginning of the existing script?

sneepstefan (approved):

pieleric:
> Do we keep a separate file for the cloud runner for Ubuntu 24.04 or do we eventually add 24.04 to the OS matrix at the beginning of the existing script?

From what I understood from Thera, it's a separate file just because we don't want to be too much informed when this fails (ie, no email). However, once we officially support Ubuntu 24.04 (hopefully soon??), then this will be moved to the standard cloud runner file.

PR #2811, opened 2024-07-01
[ Ticket #515524][feature] Catch and raise faulty camera temperature of AndorCam3 driver

K4rishma:
# If approved, do the same for other versions of andorcam

### DIFF
--- a/src/odemis/driver/andorcam3.py
+++ b/src/odemis/driver/andorcam3.py
@@ -916,6 +916,12 @@ class AndorCam3(model.DigitalCamera):
             else:
                 status_str = ""
             logging.debug(u"Temp is %.2f°C%s", temp, status_str)
+
+            # Check if the status is "Fault" and set the state accordingly
+            if status == "Fault":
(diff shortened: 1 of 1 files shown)


src/odemis/driver/andorcam3.py
```diff
@@ -916,6 +916,12 @@ def updateTemperatureVA(self):
             else:
                 status_str = ""
             logging.debug(u"Temp is %.2f°C%s", temp, status_str)
+
+            # Check if the status is "Fault" and set the state accordingly
+            if status == "Fault":
+                self.state._set_value(HwError("Camera Temperature fault detected"), force_write=True)
```
pieleric, line 922: I'd slightly change the message, as logically I think it's more the *cooling* which has a fault than the *temperature*. So something like "Camera cooling fault detected".

pieleric (changes requested):
Did you test your code? 
One option is to add temporarily a second way to set the status to "fault", by detecting the presence of a file. See odemis.driver.simulated.Stage "stage.fail" . You also need to find a Zyla or Sona, but  we should have some around...

PR #2815, opened 2024-07-04
[feature][METEOR-1095] Integrate lateral chromatic correction metadata in Odemis

K4rishma:
In the yaml file - filter wheel has new metadata

### DIFF
--- a/src/odemis/odemisd/mdupdater.py
+++ b/src/odemis/odemisd/mdupdater.py
@@ -355,6 +355,21 @@ class MetadataUpdater(model.Component):
         def updateOutWLRange(pos, fl=filter, comp_affected=comp_affected):
             wl_out = fl.axes["band"].choices[pos["band"]]
             comp_affected.updateMetadata({model.MD_OUT_WL: wl_out})
+            # apply lateral chromatic correction to align with the reference channel
+            apply_transform = fl.getMetadata().get(model.MD_CHROMATIC_COR, None)
+            if apply_transform:
+                try:
+                    metadata_cor = apply_transform[fl.position.value["band"]]
+                    assert isinstance(metadata_cor, dict), "Expected a dictionary format"
+                    assert all(
+                        isinstance(key, str) for key in metadata_cor), "All keys should be strings"
+                    comp_affected.updateMetadata(metadata_cor)
+                except (AssertionError, KeyError) as exp:
+                    # Check if CHROMATIC_COR is a dictionary with filter band positions as keys and correction metadata
+                    # dictionary as values. For e.g. correction metadata dictionary for a given band position has
+                    # the following format:
+                    # {"Pixel size cor": [1, 1], "Centre position cor": [0, 0] , "Rotation cor": 0 , "Shear cor": 1}
+                    logging.error("Chromatic correction metadata was not updated due to %s", exp)
 
         filter.position.subscribe(updateOutWLRange, init=True)
         self._onTerminate.append((filter.position.unsubscribe, (updateOutWLRange,)))
--- a/install/linux/usr/share/odemis/sim/meteor-sim.odm.yaml
+++ b/install/linux/usr/share/odemis/sim/meteor-sim.odm.yaml
@@ -195,6 +195,14 @@
              1.650796: [579.5e-9, 610.5e-9], # FF01-595/31
              2.4361944: [663.e-9, 733.e-9], # FF02-698/70
         },
+        metadata: {
+            CHROMATIC_COR: {
+                0.08: {"Pixel size cor": [1, 1],"Centre position cor": [2, 3] ,"Rotation cor": 0.1 ,"Shear cor": 0.2},
+                0.865398: {"Pixel size cor": [1, 1],"Centre position cor": [0, 0] ,"Rotation cor": 0 ,"Shear cor": 1},
+                1.650796: {"Pixel size cor": [1, 1],"Centre position cor": [4, 5] ,"Rotation cor": 0.2 ,"Shear cor": 0.3},
+                2.4361944: {"Pixel size cor": [2, 3],"Centre position cor": [6, 7] ,"Rotation cor": 0 ,"Shear cor": 0.4},
+            },
+        },
         cycle: 6.283185, # position of ref switch (0) after a full turn
     },
     # TODO: a way to indicate the best filter to use during alignement and brightfield? via some metadata?


src/odemis/model/_metadata.py
```diff
@@ -271,3 +271,5 @@
 # Fastem: Parameters used for stitching and reconstruction of 3D volumes
 MD_SLICE_IDX = "Index of slice in volume stack"  # int
 MD_FIELD_SIZE = "Average field of view of a megafield"  # tuple (px, px)
+
+MD_TRANSFORM_PER_CHANNEL = "Transformation parameter per RGB channel"
```
nandishjpatel, line 275: new line missing at the end of file

src/odemis/odemisd/mdupdater.py
```diff
@@ -352,6 +352,18 @@ def observeFilter(self, filter, comp_affected):
...
+            if apply_transform is not None:
+                if fl.position.value["band"] in apply_transform.keys():
+                    comp_affected.updateMetadata(
+                        {model.MD_PIXEL_SIZE_COR: apply_transform[fl.position.value["band"]][model.MD_PIXEL_SIZE_COR]})
+                    comp_affected.updateMetadata(
+                        {model.MD_POS_COR: apply_transform[fl.position.value["band"]][model.MD_POS_COR]})
+                    comp_affected.updateMetadata(
+                        {model.MD_ROTATION_COR: apply_transform[fl.position.value["band"]][model.MD_ROTATION_COR]})
+                    comp_affected.updateMetadata(
+                        {model.MD_SHEAR_COR: apply_transform[fl.position.value["band"]][model.MD_SHEAR_COR]})
```
nandishjpatel, line 366:
```suggestion
            apply_transform = fl.getMetadata().get(model.MD_TRANSFORM_PER_CHANNEL, None)
            fl_pos_band = fl.position.value["band"]
            if apply_transform is not None and fl_pos_band in apply_transform.keys():
                comp_affected.updateMetadata(
                    {model.MD_PIXEL_SIZE_COR: apply_transform[fl_pos_band][model.MD_PIXEL_SIZE_COR]})
                comp_affected.updateMetadata(
                    {model.MD_POS_COR: apply_transform[fl_pos_band][model.MD_POS_COR]})
                comp_affected.updateMetadata(
                    {model.MD_ROTATION_COR: apply_transform[fl_pos_band][model.MD_ROTATION_COR]})
                comp_affected.updateMetadata(
                    {model.MD_SHEAR_COR: apply_transform[fl_pos_band][model.MD_SHEAR_COR]})
```

install/linux/usr/share/odemis/sim/meteor-sim-chr.odm.yaml
```diff
@@ -0,0 +1,250 @@
...
+          Pixel size cor: [1, 1]
+          Centre position cor: [0, 0]
+          Rotation cor: 1
+          Shear cor: 1
+        0.865398:
+          Pixel size cor: [1, 1]
+          Centre position cor: [0, 0]
+          Rotation cor: 1
+          Shear cor: 1
+        1.650796:
```
pieleric, line 212: Is it on purpose that you skipped a position? (2.43)

install/linux/usr/share/odemis/sim/meteor-sim-chr.odm.yaml
```diff
@@ -0,0 +1,250 @@
...
+    },
+    metadata:
+      TRANSFORM_PER_CHANNEL:
+        0.08:
+          Pixel size cor: [1, 1]
+          Centre position cor: [0, 0]
+          Rotation cor: 1
+          Shear cor: 1
+        0.865398:
+          Pixel size cor: [1, 1]
```
pieleric, line 208: The values should be different for each positions. Firstly, because that's more realistic. Even more importantly, because this will allow you to check the correct behaviour of the simulator (ie, the metadata changes when the user changes the band).

src/odemis/odemisd/mdupdater.py
pieleric: It should be even more straightforward: just copy *all* the metadata.
```suggestion
            apply_transform = fl.getMetadata().get(model.MD_TRANSFORM_PER_CHANNEL, None)
            fl_pos_band = fl.position.value["band"]
            if apply_transform is not None and fl_pos_band in apply_transform.keys():
                comp_affected.updateMetadata(apply_transform[fl_pos_band])

```

pieleric (changes requested):

nandishjpatel (approved):

src/odemis/odemisd/mdupdater.py
```diff
@@ -352,6 +352,11 @@ def observeFilter(self, filter, comp_affected):
         def updateOutWLRange(pos, fl=filter, comp_affected=comp_affected):
             wl_out = fl.axes["band"].choices[pos["band"]]
             comp_affected.updateMetadata({model.MD_OUT_WL: wl_out})
+            # apply lateral chromatic correction to align with the reference channel
+            apply_transform = fl.getMetadata().get(model.MD_CHROMATIC_COR, None)
```
pieleric, line 356: I'd rather keep this as "light" as possible when there is no such metadata. So I'd suggest to check here, early, if the `apply_transform is None`, and immediately leave if so, without reading `.position`. Each of these are a remote function call, so quite slow.

src/odemis/odemisd/mdupdater.py
```diff
@@ -352,6 +352,11 @@ def observeFilter(self, filter, comp_affected):
...
+            fl_pos_band = fl.position.value["band"]
+            if apply_transform is not None and fl_pos_band in apply_transform.keys():
```
pieleric, line 358: You don't need `.keys()` to check a key is in a dict. Just write `fl_pos_band in apply_transform`, or maybe you can do it in a "greedy" way: directly read `apply_transform[fl_pos_band]`, and if it's not there, it'll raise a KeyError that you can catch.

src/odemis/odemisd/mdupdater.py
```diff
@@ -352,6 +352,11 @@ def observeFilter(self, filter, comp_affected):
...
+                comp_affected.updateMetadata(apply_transform[fl_pos_band])
```
pieleric, line 359: As it's coming from a file that can be edited by a user (or service engineer), it'd be good to do some small checks on the content of the MD_CHROMATIC_COR metadata. I'd suggest to check it's a dict, and that `apply_transform[fl_pos_band]` is a dict with all keys of type string. If not, do a logging.error() with complete info about what's the expected format.
Without it, if there is any error in the microscope file, every change of filter will output some quite odd error, looking like it's coming from the filter wheel, potentially hard to diagnose.

pieleric (changes requested):

src/odemis/odemisd/mdupdater.py
```diff
@@ -355,6 +355,21 @@ def observeFilter(self, filter, comp_affected):
         def updateOutWLRange(pos, fl=filter, comp_affected=comp_affected):
             wl_out = fl.axes["band"].choices[pos["band"]]
             comp_affected.updateMetadata({model.MD_OUT_WL: wl_out})
+            # apply lateral chromatic correction to align with the reference channel
+            apply_transform = fl.getMetadata().get(model.MD_CHROMATIC_COR, None)
+            if apply_transform:
+                try:
+                    metadata_cor = apply_transform[fl.position.value["band"]]
+                    assert type(metadata_cor) == dict
```
pieleric, line 363: The pythonic way is to write `isinstance(metadata_cor, dict)`.

src/odemis/odemisd/mdupdater.py
```diff
@@ -355,6 +355,21 @@ def observeFilter(self, filter, comp_affected):
...
+                    assert all(isinstance(key, str) for key in metadata_cor)
+                    comp_affected.updateMetadata(metadata_cor)
+                except (AssertionError, KeyError) as exp:
+                    logging.error(
+                        "Chromatic correction metadata was not updated due to %s. Check if the CHROMATIC_COR is a "
```
pieleric, line 368: Passing the exception error message here doesn't work well currently because the "assert" raise an AssertionError with no message. => Add a message to the `assert` or use a `if ...: raise ValueError(...)`.

src/odemis/odemisd/mdupdater.py
```diff
@@ -355,6 +355,21 @@ def observeFilter(self, filter, comp_affected):
...
+                        "dictionary with filter band positions as keys and correction metadata dictionary as "
+                        "values. For e.g. correction metadata dictionary for a given band position should "
+                        "have the following format: {\"Pixel size cor\": [1, 1],\"Centre position cor\": [0, 0]"
+                        " ,\"Rotation cor\": 0 ,\"Shear cor\": 1}", exp)
```
pieleric, line 372: I wouldn't put an example in the error message. That causes the message to be very long, and it might be repeated a lot of time in the log. Possibly, you can include an example in comment here. If someone cares of fixing the error, they should be able to find the comment.

pieleric (changes requested):

PR #2816, opened 2024-07-04
[Bug] Investigate metadata delay in case western

K4rishma:
Testing on 9th July

Please let me know where I can add logging to understand the metadata issue at Case Western.

Objective:
1) understand the source or where the delay occurs in saving the tiled images
2) try to make the overview images work by adding time delay if source of metadata issue is not clear


patrickcleeve2:
Looks good to me. One note is that the additionally logging may slow things down enough that they are synced up, so might be hard to re-create. 

Also, instead of adding an artificial display, you can force update the metadata with the stage position
das[0].metadata[model.MD_POS] = stage.position.value["x"], stage.position.value["y"]

pieleric:
You'll need to also compare to the updated stage position by xt_client (in /var/log/odemis.log), see `Updated position to...`. And check the messages before, related to the stage movement. After the move it should *always* say `"Stage move completed"`.
You also might want to add a log in actuators.ConvertStage._updatePosition() (which is called when the raw stage position changes) and look for `"reporting position"` corresponding to the MultiplexActuator.
Finally, you want to look for "Updating position for component..." from the mdupdater that is supposed to copy the stage position onto the ccd metadata (as MD_POS). All of this should happen *before* the image acquisition starts (although I don't think there is any code that ensures this is the case...)

src/odemis/driver/xt_client.py
```diff
@@ -2271,6 +2281,55 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
...
+                        # # give enough time for the stage to update its position
+                        # timeout_update = dur_lin*10
+                    timeout_update = 10  # keep it to 10 seconds till the linear speed of stage is verified
+                    time_check = time.time()
+                    while check_pos:
+                        if time.time() > timeout_update + time_check:
+                            logging.warning(
+                                "Stage position after move + %s s is still not properly updated from original pos: "
+                                "%s. Giving up waiting", timeout_update, orig_pos_raw)
+
```
patrickcleeve2, line 2308: I think you need a break here to exit
```suggestion
                            logging.warning(
                                "Stage position after move + %s s is still not properly updated from original pos: "
                                "%s. Giving up waiting", timeout_update, orig_pos_raw)
                            break

```

src/odemis/driver/xt_client.py
```diff
@@ -2235,8 +2235,18 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
                     pos["t"] = pos.pop("rx")
                 if "rz" in pos.keys():
                     pos["r"] = pos.pop("rz")
+
                 self.parent.move_stage(pos, rel=rel)
-                time.sleep(0.1)  # It takes a little while before the stage is being reported as moving
+                time.sleep(100e-3)  # It takes a little while before the stage is being reported as moving
+
```
patrickcleeve2, line 2241: Recommend changing to a constant for re-use
```suggestion
                WAIT_DURATION = 100e-3
                time.sleep(WAIT_DURATION)  # It takes a little while before the stage is being reported as moving

```

src/odemis/driver/xt_client.py
```diff
@@ -2248,7 +2258,7 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
                 while moving:
                     # Take the opportunity to update .position (every 100 ms)
                     now = time.time()
-                    if now - last_pos_update > 0.1:
+                    if now - last_pos_update > 100e-3:
```
patrickcleeve2, line 2261:
```suggestion
                    if now - last_pos_update > WAIT_DURATION:
```

src/odemis/driver/xt_client.py
```diff
@@ -2271,6 +2281,55 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
...
+                        pos_raw = self.parent.get_stage_position()
+                        # Every axis requested to move should have moved (compared to the position before starting)
+                        if all(pos_raw[a] != op for a, op in orig_pos_raw.items()):
+                            all_axes_updated = True
+
+                            # For x, y, and z, be extra picky and wait until they are "almost" at the target
+                            for k in ["x", "y", "z"]:
+                                if k in pos:
+                                    if abs(pos_raw[k] - target_pos[k]) >= 1e-6:
+                                        all_axes_updated = False
```
patrickcleeve2, line 2324: We have a util to help with this check: ```odemis.util.driver.isNearPosition```
```suggestion
                            from odemis.util.driver import isNearPosition
                            # check if all axes have started moving
                            if not isNearPosition(orig_pos_raw, pos_raw, axes=set(orig_pos.raw.keys()):
                                # check if linear axes (x, y, z) are almost at target position
                                all_axes_updated =  isNearPosition(current_pos=pos_raw, target_position=target_pos, axes=["x", "y", "z"], atol_linear=1e-6)

File: src/odemis/driver/xt_client.py [hunk 5]
```diff
@@ -2271,6 +2281,55 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
...
+                                "Stage position after move + %s s is still not properly updated from original pos: "
+                                "%s. Giving up waiting", timeout_update, orig_pos_raw)
+
+                        moving = self.parent.stage_is_moving()
+                        if moving:
+                            logging.warning("Stage reported stopped but moving again, will wait longer")
+                            time.sleep(20e-3)
+                            continue
+
+                        pos_raw = self.parent.get_stage_position()
```
patrickcleeve2, line 2315: I'd probably call this ```current_pos``` to indicate that it is being updated

src/odemis/driver/xt_client.py
```diff
@@ -2271,6 +2281,55 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
...
+                            continue
+
+                    # The stage is not moving anymore, but we still want to wait until the position has been updated
+                    check_pos = True
+                    # check for rotation and tilt is not included as it was not tested on the hardware
+                    # do not check update if angular movements are present
+                    if 'r' in pos or 't' in pos:
+                        logging.debug("Position includes rx or rz. Breaking out of the loop.")
+                        check_pos = False
+                    else:
```
patrickcleeve2, line 2293: Dont need the else block

patrickcleeve2 (changes requested):
Some minor cleanup

tepals:
Can you fix the linting?

src/odemis/driver/xt_client.py
```diff
@@ -2235,9 +2236,15 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
                     pos["t"] = pos.pop("rx")
                 if "rz" in pos.keys():
                     pos["r"] = pos.pop("rz")
+ 
+                orig_pos_raw = self.parent.get_stage_position()
+
                 self.parent.move_stage(pos, rel=rel)
                 time.sleep(0.1)  # It takes a little while before the stage is being reported as moving
 
+                # Drop axes from the original position, which are not important becuase they will not move
```
tepals, line 2245:
```suggestion
                # Drop axes from the original position, which are not important because they have not moved
```

src/odemis/driver/xt_client.py
```diff
@@ -2279,6 +2286,51 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
                 # know the user that the move is not complete.
                 if future._must_stop.is_set():
                     raise CancelledError()
+
+                # The stage is not moving anymore, but we still want to wait until the position has been updated
+                # TODO: base the timeout on the stage movement time estimation (est. *10)
+                timeout = time.time() + 10  # s
```
tepals, line 2292: a timeout is typically a certain amount of time to wait before cancelling something. `time.time()` returns the current UTC time, so this will be a very long timeout. On line 2313 it looks like you are using this more as an expected end time, so I would suggest to rename it to something like `expected_end_time`

src/odemis/driver/xt_client.py
```diff
@@ -2279,6 +2286,51 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
...
+                not_updated = True
+                wait_duration = 20e-03  # s
+
+                # Check for presence of rx/t or rz/r keys in pos
+                # In FM mode, the stage will maintain its rotation and tilt, if it is requested to change
+                # it is a different imaging mode
+                # TODO: extend the behavior to other imaging modes
+                if 'r' in pos or 't' in pos:
+                    logging.debug("Position requested to move includes rx or rz. Reported position accuracy is not"
+                                  "checked.")
```
tepals, line 2302: this might get confusing during debugging, since you are checking for `'r'` and `'t'`, but the log reports that `rx` and `rz` are not present. 
```suggestion
                    logging.debug("Position requested to move to includes t (rx or rotation) or r (rz or tilt). Reported position accuracy is not"
                                  "checked.")
```

src/odemis/driver/xt_client.py
```diff
@@ -2279,6 +2286,51 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
...
+                    not_updated = False
```
tepals, line 2303: A double negation is confusing to read, I would update this to `pos_updated = True` and then the while becomes `while not pos_updated:`
Or even: `check_pos = False` and then the while loop becomes: `while check_pos:`

src/odemis/driver/xt_client.py
```diff
@@ -2279,6 +2286,51 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
...
+
+                # Find target position in absolute
```
tepals, line 2305: what do you mean with this comment?

src/odemis/driver/xt_client.py
```diff
@@ -2279,6 +2286,51 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
...
+
+                while not_updated:
+                    time.sleep(wait_duration)
+                    if time.time() > timeout:
+                        logging.warning("Stage position after move + %s s is still same as original pos: %s. "
+                                        "Giving up waiting", 10, orig_pos_raw)
+                        break
+
+                    # Every axis requested to move should have moved (compared to the position before starting)
+                    current_pos_raw = self.parent.get_stage_position()
```
tepals, line 2319: what does the `_raw` indicate?

src/odemis/driver/xt_client.py
```diff
@@ -2279,6 +2286,51 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
...
+                if not rel:
+                    target_pos = pos
+                else:
+                    target_pos = {ax: val + pos[ax] for ax, val in orig_pos_raw.items()}
+
+                while not_updated:
+                    time.sleep(wait_duration)
+                    if time.time() > timeout:
+                        logging.warning("Stage position after move + %s s is still same as original pos: %s. "
+                                        "Giving up waiting", 10, orig_pos_raw)
```
tepals, line 2315: can you make the 10s a variable, if you would now update the value on line 2292 it will not indicate it here

src/odemis/driver/xt_client.py
```diff
@@ -2279,6 +2286,51 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
                 # know the user that the move is not complete.
                 if future._must_stop.is_set():
                     raise CancelledError()
+
+                # The stage is not moving anymore, but we still want to wait until the position has been updated
+                # TODO: base the timeout on the stage movement time estimation (est. *10)
+                timeout = time.time() + 10  # s
+                not_updated = True
+                wait_duration = 20e-03  # s
```
tepals, line 2294: For readability it is nice to define variables close to the first place they are used, so I would move this to directly above the while-loop.

src/odemis/driver/xt_client.py
```diff
@@ -2279,6 +2286,51 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
...
+                        # Therefore, it is not important to check z
+                        all_axes_updated = isNearPosition(current_pos=current_pos_raw, target_position=target_pos,
+                                                          axes={"x", "y"}, atol_linear=1e-6)
+
+                        if all_axes_updated:
+                            logging.debug("Position has updated fully: from %s -> %s", orig_pos_raw,
+                                          current_pos_raw)
+                            break
+                        else:
+                            time.sleep(wait_duration)
```
tepals, line 2332: I think this can be removed, otherwise you are waiting twice

src/odemis/driver/xt_client.py
```diff
@@ -2279,6 +2286,51 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
...
+                                        "Giving up waiting", 10, orig_pos_raw)
+                        break
+
+                    # Every axis requested to move should have moved (compared to the position before starting)
+                    current_pos_raw = self.parent.get_stage_position()
+                    if not isNearPosition(orig_pos_raw, current_pos_raw, axes=set(orig_pos_raw.keys())):
+                        # For x, y, and z, be extra picky and wait until they are "almost" at the target
+                        # Due to y-z linkage, there is an equivalent change in z axis when y axis is changed
+                        # Therefore, it is not important to check z
+                        all_axes_updated = isNearPosition(current_pos=current_pos_raw, target_position=target_pos,
```
tepals, line 2324: why do you use keywords here, but not in line 2320?

src/odemis/driver/xt_client.py
```diff
@@ -2279,6 +2286,51 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
...
+                while not_updated:
+                    time.sleep(wait_duration)
+                    if time.time() > timeout:
+                        logging.warning("Stage position after move + %s s is still same as original pos: %s. "
+                                        "Giving up waiting", 10, orig_pos_raw)
+                        break
+
+                    # Every axis requested to move should have moved (compared to the position before starting)
+                    current_pos_raw = self.parent.get_stage_position()
+                    if not isNearPosition(orig_pos_raw, current_pos_raw, axes=set(orig_pos_raw.keys())):
```
tepals, line 2320: can you add a comment to explain why you check all the keys here, while in the next step you only check x and y?

tepals: And is it not necessary to define a atol_linear here? The default is 100µm, so if your move is less than 100µm, you will never go inside this if statement.

src/odemis/driver/xt_client.py
```diff
@@ -2279,6 +2286,51 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
...
+                        # For x, y, and z, be extra picky and wait until they are "almost" at the target
+                        # Due to y-z linkage, there is an equivalent change in z axis when y axis is changed
+                        # Therefore, it is not important to check z
+                        all_axes_updated = isNearPosition(current_pos=current_pos_raw, target_position=target_pos,
+                                                          axes={"x", "y"}, atol_linear=1e-6)
+
+                        if all_axes_updated:
```
tepals, line 2327: this is a bit of a misleading variable name, since you only checked x, y and z

tepals (changes requested):

src/odemis/driver/xt_client.py
K4rishma: not_updated/check_pos is used before the while_loop. With that , all the other variable are combined where check_pos is defined

src/odemis/driver/xt_client.py
K4rishma: removed raw, I confused the naming with the meteor functionality task

src/odemis/driver/xt_client.py
K4rishma: the target position of the requested move, the requested move can be relative and in absolute movement. If relative convert in absolute else keep it same

Changed the comment to # Get the target position in absolute coordinates

src/odemis/driver/xt_client.py
```diff
@@ -2279,6 +2286,52 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
                 # know the user that the move is not complete.
                 if future._must_stop.is_set():
                     raise CancelledError()
+
+                # The stage is not moving anymore, but we still want to wait until the position has been updated
+                # TODO: base the timeout on the stage movement time estimation (est. *10)
+                timeout = 10  #s
```
tepals, line 2292:
```suggestion
                timeout = 10  # s
```

src/odemis/driver/xt_client.py
```diff
@@ -2279,6 +2286,52 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
...
+                    check_pos = False
+
+                # Get the target position in absolute coordinates
+                if not rel:
+                    target_pos = pos
+                else:
+                    target_pos = {ax: val + pos[ax] for ax, val in orig_pos.items()}
+
+                while check_pos:
+                    time.sleep(wait_duration)
```
tepals, line 2313: why do you start the loop with a sleep?

src/odemis/driver/xt_client.py
```diff
@@ -2279,6 +2286,52 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
...
+                # TODO: base the timeout on the stage movement time estimation (est. *10)
+                timeout = 10  #s
+                expected_end_time = time.time() + timeout  # s
+                check_pos = True
+                wait_duration = 20e-03  # s
+
+                # Check for presence of rx/t or rz/r keys in pos
+                # In FM mode, the stage will maintain its rotation and tilt, if it is requested to change
+                # it is a different imaging mode
+                # TODO: extend the behavior to other imaging modes
```
tepals, line 2300: I have trouble understanding how you can extend stage movement to imaging modes, is this meteor specific?

src/odemis/driver/xt_client.py
```diff
@@ -2279,6 +2286,52 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
...
+                    target_pos = pos
+                else:
+                    target_pos = {ax: val + pos[ax] for ax, val in orig_pos.items()}
+
+                while check_pos:
+                    time.sleep(wait_duration)
+                    if time.time() > expected_end_time:
+                        logging.warning("Stage position after move + %s s is still same as original pos: %s. "
+                                        "Giving up waiting", timeout, orig_pos)
+                        break
```
tepals, line 2317: shouldn't this be directly after checking the position? Now you are sleeping for a duration and then break the loop if the time > expected_end_time, however you do not check if the stage position has updated after the sleep

tepals (changes requested):

src/odemis/driver/xt_client.py
K4rishma: I have added a better explanation in the code. 

To explain here - We are checking only x and y and not other axes because during overview acquisition x and y (also z which is indirectly checked due to y-z linkage) are repeatedly changed , keeping the rotation and tilt constant. For other imaging modes, like SEM there may be request to change tilt repeatedly or something similar. 

`                # TODO: extend the behavior if rotation, tilt axes also need to be checked.
                # So far checking the above is not required, as we have observed reporting stage position only in x/y
                # being compromised. The issue with case western system was that the stage metadata was not
                # updated in timely manner during the overview acquisition in FM mode (during which rotation, tilt are
                # not changed), resulting in misplaced tiles in the image display. Even when the stage has physically
                # moved to the target position but took time in updating its actual position through metadata
                # The issue is resolved by checking if only x and y stage metadata reports the target position properly`

src/odemis/driver/xt_client.py
K4rishma: good point, I have put this at the end of the while loop so it checks time after it checks stage position

src/odemis/driver/xt_client.py
```diff
@@ -2279,6 +2286,58 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
...
                 if future._must_stop.is_set():
                     raise CancelledError()
+
+                # TODO: extend the behavior if rotation, tilt axes also need to be checked.
+                # So far checking the above is not required, as we have observed reporting stage position only in x/y
+                # being compromised. The issue with case western system was that the stage metadata was not
+                # updated in timely manner during the overview acquisition in FM mode (during which rotation, tilt are
+                # not changed), resulting in misplaced tiles in the image display. Even when the stage has physically
+                # moved to the target position but took time in updating its actual position through metadata
+                # The issue is resolved by checking if only x and y stage metadata reports the target position properly
```
pieleric, line 2296: As described below, I think we can easily implement it, by loosing a little bit the requirements in the case of rotation. So this TODO could go away.
Also, in general, please refrain from mentioning specific names of installations in public code.

src/odemis/driver/xt_client.py
```diff
@@ -2279,6 +2286,58 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
...
+                if 'r' in pos or 't' in pos:
+                    logging.debug("Position requested to move to includes t (rx/tilt) or r (rz/rotation). "
+                                  "Reported position accuracy is not checked.")
+                    check_pos = False
+
+                # Get the target position in absolute coordinates
+                if not rel:
+                    target_pos = pos
+                else:
+                    target_pos = {ax: val + pos[ax] for ax, val in orig_pos.items()}
```
pieleric, line 2315: Can you invert the `if` so that it's more straightforward to read?

src/odemis/driver/xt_client.py
```diff
@@ -2279,6 +2286,58 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
...
+                # not changed), resulting in misplaced tiles in the image display. Even when the stage has physically
+                # moved to the target position but took time in updating its actual position through metadata
+                # The issue is resolved by checking if only x and y stage metadata reports the target position properly
+
+                # The stage is not moving anymore, but we still want to wait until the position has been updated
+                # TODO: base the timeout on the stage movement time estimation (est. *10)
+                timeout = 10  # s
+                expected_end_time = time.time() + timeout  # s
+                check_pos = True
+                wait_duration = 20e-03  # s
```
pieleric, line 2303: Can you move this up, and use it also in the `sleep`s of lines 2268 & 2277.
It be also clearer to use uppercases, to that it's clear it's a constant.

src/odemis/driver/xt_client.py
```diff
@@ -2279,6 +2286,58 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
...
+                            logging.debug("Position has updated fully: from %s -> %s", orig_pos,
+                                          current_pos)
+                            break
+
+                    if time.time() > expected_end_time:
+                        logging.warning("Stage position after move + %s s is moved to %s instead of target pos: %s. "
+                                        "Giving up waiting", current_pos, timeout, target_pos)
+                        break
+
+                    time.sleep(wait_duration)
```
pieleric, line 2339: You should add a check for cancellation here, before the sleep, so that it ends early if the stage was stopped early:
```
                if future._must_stop.is_set():
                    raise CancelledError()
```

src/odemis/driver/xt_client.py
```diff
@@ -2279,6 +2286,58 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
...
+
+                # TODO: extend the behavior if rotation, tilt axes also need to be checked.
+                # So far checking the above is not required, as we have observed reporting stage position only in x/y
+                # being compromised. The issue with case western system was that the stage metadata was not
+                # updated in timely manner during the overview acquisition in FM mode (during which rotation, tilt are
+                # not changed), resulting in misplaced tiles in the image display. Even when the stage has physically
+                # moved to the target position but took time in updating its actual position through metadata
+                # The issue is resolved by checking if only x and y stage metadata reports the target position properly
+
+                # The stage is not moving anymore, but we still want to wait until the position has been updated
```
pieleric, line 2298: Please give a little bit more background information here. I'd write something like:
"The stage is not moving anymore, however in some *rare* cases, the server still reports the old position for a short while (typically < 1s). This can cause all sorts of confusion in the calling code, as it would seem that the move didn't have any effect, and later on, after the position is eventually updated that the stage has moved unexpectedly. So, wait until the reported position is (1) not identical to the starting position and (2) not too far from the target position. "

src/odemis/driver/xt_client.py
```diff
@@ -2279,6 +2288,74 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
...
+                    target_pos = pos
+
+                while check_pos:
+                    current_pos = self.parent.get_stage_position()
+                    # Every axis requested to move should have moved (compared to the position before starting)
+                    # if there is some change w.r.t starting position, proceed to check the target position
+
+                    # TODO Check all axes and confirm it works on the hardware
+                    # if not isNearPosition(current_pos=current_pos, target_position=orig_pos, axes=set(orig_pos.keys()),
+                    #                       atol_linear=50e-9, atol_rotation=numpy.radians(0.05)):
```
pieleric, line 2326: I think the original code, which uses "precisely" not equal is fine. That's because here we just just want to know whether the position has been updated. Even a very tiny change is already a proof that it has been updated.

pieleric (approved):

src/odemis/driver/xt_client.py
pieleric: (no need to fix this here, you'll do that once you test that code)

K4rishma: I used the original code - pushed just now. To get clarity on the second comment, do we need to fix it during the testing as well or should it be fixed?

src/odemis/driver/xt_client.py
```diff
@@ -2358,6 +2367,73 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
...
+                    #     axes_updated = isNearPosition(current_pos=current_pos, target_position=target_pos,
+                    #                                   axes=set(orig_pos.keys()), atol_linear=1e-6,
+                    #                                   atol_rotation=numpy.radians(6))
+                    #     if axes_updated:
+                    #         logging.debug("Position has updated fully: from %s -> %s", orig_pos,
+                    #                       current_pos)
+                    #         break
+
+                    # TODO Remove the check only on x any y
+                    if all(current_pos[a] != op for a, op in orig_pos.items()):
```
patrickcleeve2, line 2414: Avoid direct float comparison
```suggestion
                    if all(math.isclose(current_pos[a], op, abs_tol=1e-6) for a, op in orig_pos.items()):
```

patrickcleeve2 (approved):
Just a minor comment

src/odemis/driver/xt_client.py
pieleric: No. There is a very special reason for this: we want to check whether the value has changed. That's mostly all we care about. If it has changed, it means the xt-client has received an updated metadata. Maybe it's very nearby, because the stage actually hasn't really moved (due to hardware issue), but we now know the new position. 
So, on purpose it's not using float comparison. 

That deserves a comment though...

PR #2817, opened 2024-07-08
[FIX] [METEOR-1126] Add support for importing RGB SEM Images

patrickcleeve2:
Currently Odemis assumes that all SEM images are 2D grayscale. However, recently we have encountered SEM images that are stored as 3D RGB images. This causes errors when they are imported in the correlation tab as they do not have the correct metadata or classification.

This PR adds support for converting these 3D RGB SEM Images into standard 2D Static SEM Streams to work with the correlation controls. This is a temporary fix until the full import pipeline can be updated to support this new data structure. 

More details: https://docs.google.com/document/d/1jtvdAykCm8G37IuthWWaEK0OSU1qRZIX-5LEJGtTh_4/edit#heading=h.xjiyqw33io4d

### DIFF
--- a/src/odemis/gui/cont/correlation.py
+++ b/src/odemis/gui/cont/correlation.py
@@ -22,6 +22,7 @@ You should have received a copy of the GNU General Public License along with
 Odemis. If not, see http://www.gnu.org/licenses/.
 
 """
+import copy
 import logging
 import math
 import wx
@@ -32,7 +33,7 @@ import wx.html
 import odemis.acq.stream as acqstream
 import odemis.gui.model as guimod
 from odemis import model
-from odemis.acq.stream import StaticStream, StaticFluoStream, StaticSEMStream
+from odemis.acq.stream import RGBStream, StaticStream, StaticFluoStream, StaticSEMStream
 from odemis.gui.util import call_in_wx_main
 from odemis.gui.cont.tabs.localization_tab import LocalizationTab
 
@@ -51,6 +52,31 @@ def update_image_in_views(s: StaticStream, views: list) -> None:
             if st is s:
                 sp._shouldUpdateImage()
 
+def convert_rgb_to_sem(rgb_stream: RGBStream) -> StaticSEMStream:
+    """Convert an RGB stream to a SEM stream
+    :param rgb_stream: (RGBStream) the RGB stream to convert
+    :return: (StaticSEMStream) the converted SEM stream
+    """
+    d = rgb_stream.raw[0]
+    if isinstance(d, model.DataArrayShadow):
+        d = d.getData()
+
+    # get dim order, and select the first channel (arbitrary choice)
+    dims = d.metadata[model.MD_DIMS]
+    if dims == "YXC":
+        d = d[:, :, 0]
+    elif dims == "CYX":
+        d = d[0, :, :]
+
+    # convert to sem stream
+    sem_stream = StaticSEMStream(rgb_stream.name.value, raw=d)
+
+    # update md
+    sem_stream.raw[0].metadata = copy.deepcopy(d.metadata)
+    sem_stream.raw[0].metadata[model.MD_ACQ_TYPE] = model.MD_AT_EM
+    sem_stream.raw[0].metadata[model.MD_DIMS] = "YX"
+
+    return sem_stream
 
 class CorrelationController(object):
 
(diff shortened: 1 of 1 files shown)


pieleric (changes requested):
See my comment on your google doc. I think we can adjust data_to_static_streams() in a "safe" way to automatically revert the RGB greyscale images to 2D images.

sneepstefan (approved):

src/odemis/util/dataio.py
```diff
@@ -120,7 +121,28 @@ def data_to_static_streams(data):
...
-            klass = stream.RGBStream
+
+            # Check if it's a SEM or FIB image, and if so, use the SEMStream
+            if acqtype in [model.MD_AT_EM, model.MD_AT_FIB]:
+                if isinstance(d, model.DataArrayShadow):
+                    d = d.getData()
+
+                # get dim order, and select the first channel (arbitrary choice)
+                dims = d.metadata.get(model.MD_DIMS, "CYX")
+                if dims == "YXC":
```
pieleric, line 132: We shouldn't just automatically discard data. *only* do it if it's actually duplicate data (ie, greyscale image encoded as RGB). So you need to check using something like: `if dims == "YXC" and numpy.array_equal(d[:, :, 0] , d[:, :, 1]) and  numpy.array_equal(d[:, :, 0] , d[:, :, 2])`
(Maybe there is a nicer way to write, for instance using `ci`, but I couldn't really think of it quickly)

src/odemis/util/dataio.py
```diff
@@ -20,6 +20,7 @@
 
 """
 
+import copy
```
pieleric, line 23: That's not needed, is it?

src/odemis/gui/cont/correlation.py
```diff
@@ -51,6 +52,34 @@ def update_image_in_views(s: StaticStream, views: list) -> None:
             if st is s:
                 sp._shouldUpdateImage()
 
+def convert_rgb_to_sem(rgb_stream: RGBStream) -> StaticSEMStream:
```
pieleric, line 55: Hopefully, if data_to_static_streams() is adjusted, this whole part can be dropped.

pieleric (changes requested):

src/odemis/util/dataio.py
patrickcleeve2: The issue is that the images sometimes have rgb content (overlay, metadata icons) so they aren't identical channels, i've added an example + comment in the doc. 
I think it is pretty safe to do this, as the only way the MD_AT_EM tag gets added is if we parse it as an SEM image, the rgb content is an edge case of that for these exported images with overlays.

src/odemis/gui/cont/correlation.py
patrickcleeve2: See comment in doc about stitched tilesets

patrickcleeve2:
@pieleric reverted to only correlation tab change

src/odemis/gui/cont/correlation.py
```diff
@@ -191,6 +217,12 @@ def add_streams(self, streams: list) -> None:
             if s in self._tab_data_model.streams.value or not isinstance(s, StaticStream):
                 continue
 
+            # if the user has loaded a rgb stream, assume it is meant to be a SEM stream
+            # (convert to 2D SEM stream, fix metadata, etc.)
+            if isinstance(s, RGBStream):
+                logging.debug(f"Converting RGB stream to SEM: {s.name.value}")
+                s = convert_rgb_to_sem(s)
+
```
pieleric, line 225: I agree this makes sense to only place it in the correlation tab, as it's a special case of what the user expectation, which is only valid for the correlation tab.
However, this is now called for every stream added (even from the localization tab). Also, it's a little inefficient, as it first creates a RGB stream, and then discards it for creating a new stream. Please move the whole code to `correlation_tab.load_stream()`. *Before* the call to `data_to_static_streams()`, you can check for each data, whether it is RGB, and if so, drop the the extra channels.

pieleric (changes requested):

src/odemis/gui/cont/correlation.py
patrickcleeve2: I disagree, I think since it is essentially error handling it should happen as late as possible. The streams from the localization tab will never be rgb (unless something has gone very wrong) so there isn't really any extra work being done there. 
In terms of efficiency, moving it before data_to_static_streams would mean going through each data another time, and performing a similar check that happens in data_to_static_streams, rather than a simple comparison after its loaded which seems less efficient, even if every once in a while the stream needs to be re-created.  I also think there is a higher chance of making a bad assumption if we modify the data before data_to_static_streams.

pieleric: OK.

pieleric (approved):

PR #2819, opened 2024-07-08
[NSRPARMA-52] Development of a driver for the Ephemeron MightyEBIC scan controller

sneepstefan:
Digital EBIC acquisition is needed specifically for the PARMA project. This basically consists of an extra/special stream for EBIC, a driver to communicate with their API for the controller and a modified SPARC acquisition procedure. The acquisition procedure must be simultaneous with the CL acquisition on our side. This is driven by TTL signaling which will trigger EBIC acquisition the moment we move to the pixel position of interest. Eventually the acquisition will be done and we would be able to request the data from the EBIC controller.

### DIFF
--- a/src/odemis/acq/stream/_helper.py
+++ b/src/odemis/acq/stream/_helper.py
@@ -1398,9 +1398,6 @@ class FastScanningDetector(RepetitionStream):
     def _onResolution(self, value):
         self._updateAcquisitionTime()
 
-    def _onSppChange(self, value):
-        self._detector.spp._set_value(value, force_write=True)
-
 
 class CLSettingsStream(FastScanningDetector):
     """
@@ -1474,10 +1471,15 @@ class DigitalEBICStream(FastScanningDetector):
             kwargs["acq_type"] = model.MD_AT_EBIC
         super().__init__(name, detector, dataflow, emitter, **kwargs)
 
-        if model.hasVA(self._detector, "spp"):
-            self.spp = model.IntContinuous(self._detector.spp.value, (1, 10))
-            self.spp.subscribe(self._onSppChange)
+        self._ebic = detector
 
+        self.repetition.subscribe(self.onRepetitionChange)
+        # if model.hasVA(self._detector, "spp"):
+        #     self.spp = model.IntContinuous(self._detector.spp.value, (1, 10))
+        #     self.spp.subscribe(self._onSppChange)
+
+    def onRepetitionChange(self, value):
+        self.detector.repetition.value = value
 
 # Maximum allowed overlay difference in electron coordinates.
 # Above this, the find overlay procedure will consider an error occurred and
--- a/src/odemis/acq/stream/_helper.py
+++ b/src/odemis/acq/stream/_helper.py
@@ -26,6 +26,8 @@ import math
 import time
 
 import numbers
+from typing import Tuple, Any, Dict
+
 import numpy
 
 from odemis import model
@@ -1448,11 +1450,15 @@ class EBICSettingsStream(FastScanningDetector):
         emtvas: don't put resolution or scale
         """
         if "acq_type" not in kwargs:
-            kwargs["acq_type"] = model.MD_AT_EM
+            kwargs["acq_type"] = model.MD_AT_EBIC
         super().__init__(name, detector, dataflow, emitter, **kwargs)
 
 
-class DigitalEBICStream(FastScanningDetector):
+SCANNER_POS_MD = {model.MD_POS, model.MD_POS_COR, model.MD_PIXEL_SIZE, model.MD_PIXEL_SIZE_COR,
+                  model.MD_SHEAR, model.MD_SHEAR_COR, model.MD_ROTATION, model.MD_ROTATION_COR}
+
+
+class IndependentEBICStream(FastScanningDetector):
     """
     A special EBIC stream, typically with a EBIC (current) as a detector and its own scanner.
     It's physically very similar to the SEM stream, but as we want to select just a region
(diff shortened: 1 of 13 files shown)


src/odemis/acq/stream/_sync.py
```diff
@@ -37,6 +37,7 @@
 from odemis.acq import leech
 from odemis.acq.leech import AnchorDriftCorrector
 from odemis.acq.stream._live import LiveStream
+from odemis.driver import ephemeron
```
pieleric, line 40: The acq module should know nothing about the drivers. If there are constants to care about, you can place them in odemis.model . That said, I don't think it's necessary to check the `scanState` anyway.

src/odemis/acq/stream/_sync.py
```diff
@@ -1953,6 +1961,20 @@ def _runAcquisition(self, future):
...
+                ebic_det = model.getComponent(role='ebic-detector')
+            except LookupError:
+                logging.debug("No EBIC detector found skipping synchronized acquisition of EBIC.")
+
+            if ebic_det:
+                if ebic_det.scanState.value != ephemeron.STATE_NAME_IDLE:
+                    logging.warning(f"Scan state of the EBIC scan controller is not {ephemeron.STATE_NAME_IDLE}")
+                else:
+                    # Set the EBIC scan controller state to running
+                    ebic_det.scanState.value = ephemeron.STATE_NAME_RUNNING
```
pieleric, line 1976: Is this to start the scanning? If so, this should be done by following the standard Odemis API. It consist in subscribing to the detector's DataFlow. The function already does that though... so normally there is nothing really special.

src/odemis/acq/stream/_sync.py
```diff
@@ -143,6 +144,13 @@ def __init__(self, name, streams):
                 self._sstage = s._sstage
                 break
 
+        # Check for an EBIC detector with its own scan box
+        for s in streams:
+            if hasattr(s, "_ebic") and s._ebic.role == "ebic-detector":
+                self.useEBIC = True  # bool for the presence of the EBIC
+                self._ebic = s._ebic  # the actual EBIC component
+                break
```
pieleric, line 152: We shouldn't have any special code for the "EBIC". I think it should just check whether the detector has also a scanner info (ie, .resolution and .dwellTime). If so, they should be set properly.

src/odemis/acq/stream/_helper.py
```diff
@@ -1451,6 +1451,36 @@ def __init__(self, name, detector, dataflow, emitter, **kwargs):
...
+        super().__init__(name, detector, dataflow, emitter, **kwargs)
+
+        self._ebic = detector
+
+        self.repetition.subscribe(self.onRepetitionChange)
+        # if model.hasVA(self._detector, "spp"):
+        #     self.spp = model.IntContinuous(self._detector.spp.value, (1, 10))
+        #     self.spp.subscribe(self._onSppChange)
+
+    def onRepetitionChange(self, value):
```
pieleric, line 1481: I think that the main part missing here is in live mode. The hardware is normally configured so that the SPARC scanner (only) controls the e-beam. So if we want to acquire an image, we have to acquire simultaneously a EBIC data, and run the "e-beam" component. This is typically done by acquiring on the se-detector. 
We need special code that handle this.

src/odemis/driver/ephemeron.py
```diff
@@ -0,0 +1,788 @@
...
+        self.scanState = model.StringVA("Disconnected")
+
+        try:
+            self._start_opcclient_thread()
+        except ConnectionError:
+            raise ConnectionError(self._error_msg)
+
+        # register the VA's
+        # higher spp will force a higher dwell_time as well as increasing scan_time
+        self.spp = model.IntEnumerated(1, set(range(1, 11)), setter=self.on_spp_change)
```
pieleric, line 150: Let's call this VA `samplesPerPixel`.

pieleric, line 150: Can you put the maximum number of samples per pixel as a constant at the top of the file. It'll be clearer (that it's 10).

src/odemis/driver/ephemeron.py
```diff
@@ -0,0 +1,788 @@
...
+        self.numberOfChannels = model.IntEnumerated(2, set(range(1, 9)), setter=self.on_chan_num_change)
```
pieleric, line 151: Please put the maximum number channels as a constant.

src/odemis/driver/ephemeron.py
```diff
@@ -0,0 +1,788 @@
...
+        self._channel = channel
+        self._namespace = namespace
+        self._client = Client(url=url)
+        self._opc_server = None
+        self._binning = (1, 1)
+        self._error_msg = None
+        self._simserver = None
+        self._connected = False
+        self._scan_time = 0.0
+        self._dwell_time = 0.0
```
pieleric, line 116: Every variable which is set as `None` should be "typed" to indicate what will be the real type of the variable.

src/odemis/driver/ephemeron.py
```diff
@@ -0,0 +1,788 @@
...
+        :param url:
+        :param namespace:
+        """
+        super().__init__(name, role, **kwargs)
+        # for debug, eventually -> get resolution from the controller/detector
+        max_res = (1024, 1024)
+        min_res = (1, 1)
+
+        self.device = device
+        self.port = port
```
pieleric, line 98: What are device and port? I think you can drop them, and just use the "url" + "namespace" arguments.

src/odemis/driver/ephemeron.py
```diff
@@ -0,0 +1,788 @@
...
+from scipy import ndimage
+
+import functions
+# from constants import *
+from odemis import model, dataio
+from odemis.model import Detector, oneway
+from odemis.util import img
+
+# constants
+URI = "opc.tcp://localhost:4840/freeopcua/server/"
```
pieleric, line 46: Unused -> drop

src/odemis/driver/ephemeron.py
```diff
@@ -0,0 +1,788 @@
...
+        self.states = {}
+        self.transitions = {}
+
+    async def setup(self) -> None:
+        """Call server async setup functions."""
+        await self.server.init()
+        self.server.set_endpoint("opc.tcp://localhost:4840/freeopcua/server/")
+
+        # set up our own namespace, not really necessary but should as spec
+        self.idx = await self.server.register_namespace("http://examples.freeopcua.github.io")
```
pieleric, line 555: You should use `NAMESPACE`, or `SIM_NAMESPACE`. And probably you want to use a namespace which is more going to match the actual device (I hope they don't keep using that example namespace!).

src/odemis/driver/ephemeron.py
```diff
@@ -0,0 +1,788 @@
...
+
+        self._evtq.put(time.time())
+
+    def _waitSync(self):
+        """
+        Block until the Event on which the dataflow is synchronised has been
+          received. If the DataFlow is not synchronised on any event, this
+          method immediately returns
+        """
+        if self._sync_event:
```
pieleric, line 760: I don't think you should try to handle software triggers (yet). => Just drop that code.
Possibly, you could still leave the `synchronizedOn()` function, to detect the user requested a hardware trigger (see andorcam2), and if that's not the case, raise an error at acquisition complaining that the device only works if synchronized (but I think it just makes our life harder when debugging... so I would not even bother).

src/odemis/driver/ephemeron.py
```diff
@@ -0,0 +1,788 @@
...
+                # for _all_ the detectors to be ready before scanning.
+                self.data._waitSync()
+                callback(self._simulate_image())
+        except Exception:
+            logging.exception("Unexpected failure during image acquisition")
+        finally:
+            logging.debug("Acquisition thread closed")
+            self._acquisition_must_stop.clear()
+
+    def _simulate_image(self):
```
pieleric, line 364: All the "simulation" code should go in the `MightyEbicSimulator`.

src/odemis/driver/ephemeron.py
```diff
@@ -0,0 +1,788 @@
...
+        try:
+            self._start_opcclient_thread()
+        except ConnectionError:
+            raise ConnectionError(self._error_msg)
+
+        # register the VA's
+        # higher spp will force a higher dwell_time as well as increasing scan_time
+        self.spp = model.IntEnumerated(1, set(range(1, 11)), setter=self.on_spp_change)
+        self.numberOfChannels = model.IntEnumerated(2, set(range(1, 9)), setter=self.on_chan_num_change)
+        self.oversampling = model.IntEnumerated(0, {0, 2, 4, 8, 16, 32, 64})
```
pieleric, line 152: I feel that "0" is odd. Isn't it more like "1" (and if the device wants to receive 0 in such case, we can convert). Actually I don't even understand the different between oversampling and samplesPerPixel. Can you add a comment explaining the concepts (as defined by the device)?

src/odemis/driver/ephemeron.py
```diff
@@ -0,0 +1,788 @@
...
+        """
+        var = await self._client.nodes.root.get_child(f"0:Objects/{self.idx}:{obj_name}/{self.idx}:{var_name}")
+        value = await var.read_value()
+        return value
+
+    def on_repetition_change(self, value):
+        return value
+
+    def on_spp_change(self, value):
+        return value
```
pieleric, line 526: My understanding is that the dwell time is defined by the samples per pixel, oversampling, and delay. However, in Odemis, typically we have just one setting the "dwellTime". Here, the driver *must* provide a dwellTime VA so that the acquisition code can copy the dwellTime of the standard e-beam scanner and let the EBIC know (and the closest dwellTime < the requested dwellTime should be used). 

Ideally, we would compute the best 3 settings based on the dwellTime. I don't know if that's possible. If not, we could allow the user to adjust the samples per pixel, oversampling, and the driver would pick the best delay for matching the dwellTime.

src/odemis/driver/ephemeron.py
```diff
@@ -0,0 +1,788 @@
...
+from typing import Any, Optional
+
+import asyncio
+from asyncua import Client, Server, Node, ua
+from asyncua.common.methods import uamethod
+from asyncua.common.statemachine import State, StateMachine, Transition
+import numpy
+from scipy import ndimage
+
+import functions
```
pieleric, line 39: That file is missing. That said, we don't want it as-is anyway. Maybe we can just copy paste the useful parts here? Otherwise, we could have a _ephemeron_functions.py file.

src/odemis/driver/ephemeron.py
```diff
@@ -0,0 +1,788 @@
...
+        # for debug, eventually -> get resolution from the controller/detector
+        max_res = (1024, 1024)
+        min_res = (1, 1)
+
+        self.device = device
+        self.port = port
+        self.acq_data = None
+        self.idx = None
+        self.sub_handler = SubHandler()
+        self.delay = 5e-8  # need to make this a VA on the stream panel as well?
```
pieleric, line 102: Yes, as a dwellTime.

src/odemis/driver/ephemeron.py
```diff
@@ -0,0 +1,788 @@
...
+        idt = numpy.iinfo(self.fake_img.dtype)
+        data_depth = idt.max - idt.min + 1
+        self._shape = (data_depth,)  # only one point
+
+        # 8 or 16 bits image
+        if data_depth == 255:
+            bpp = 8
+        else:
+            bpp = 16
+        self.bpp = model.IntEnumerated(bpp, {8, 16})
```
pieleric, line 140: All this simulated code should go away. The actual data is received as float64. However, I'm pretty certain the precision of the device is not that high. Maybe we can just hardcode it to float(2**32), which is more realistic, and also indicates it's a float.

src/odemis/driver/ephemeron.py
```diff
@@ -0,0 +1,788 @@
...
+        self.scanState = model.StringVA("Disconnected")
+
+        try:
+            self._start_opcclient_thread()
+        except ConnectionError:
+            raise ConnectionError(self._error_msg)
```
pieleric, line 146: You need to raise `HwError` if there is a problem in the connection which indicates that the device is not reachable. This way, the backend will show a nice warning message to the user, and try again later.

src/odemis/driver/ephemeron.py
```diff
@@ -0,0 +1,788 @@
...
+        # register the VA's
+        # higher spp will force a higher dwell_time as well as increasing scan_time
+        self.spp = model.IntEnumerated(1, set(range(1, 11)), setter=self.on_spp_change)
+        self.numberOfChannels = model.IntEnumerated(2, set(range(1, 9)), setter=self.on_chan_num_change)
+        self.oversampling = model.IntEnumerated(0, {0, 2, 4, 8, 16, 32, 64})
+        self.oversampling.subscribe(self.on_oversampling_change)
+        self.binning = model.ResolutionVA((1, 1), ((1, 1), (16, 16)))
+        # the resolution of the scanner of the EBIC controller
+        self.resolution = model.ResolutionVA(max_res, (min_res, max_res), unit="px")
+        self.repetition = model.TupleVA((1, 1), unit="px", setter=self.on_repetition_change)
```
pieleric, line 157: That shouldn't be needed. Only resolution.

src/odemis/driver/ephemeron.py
```diff
@@ -0,0 +1,788 @@
...
+        data_depth = idt.max - idt.min + 1
+        self._shape = (data_depth,)  # only one point
+
+        # 8 or 16 bits image
+        if data_depth == 255:
+            bpp = 8
+        else:
+            bpp = 16
+        self.bpp = model.IntEnumerated(bpp, {8, 16})
+        self.scanState = model.StringVA("Disconnected")
```
pieleric, line 141: I don't think you want to let the user change that, or even see that. The state should automatically be set based on the whether the DataFlow is "generating" or not.

src/odemis/driver/ephemeron.py
```diff
@@ -0,0 +1,788 @@
...
+        self.resolution = model.ResolutionVA(max_res, (min_res, max_res), unit="px")
+        self.repetition = model.TupleVA((1, 1), unit="px", setter=self.on_repetition_change)
+
+        self.data = EBICDataFlow(self)
+        self._acquisition_thread = None
+        self._acquisition_lock = threading.Lock()
+        self._acquisition_init_lock = threading.Lock()
+        self._acquisition_must_stop = threading.Event()
+        self._hwVersion = "Ephemeron EBIC controlbox S/N: 123456-SIM"
+        self._swVersion = "Firmware: simulated-device"
```
pieleric, line 165: Can you read that via OPC-UA? If not, don't touch them.

pieleric (changes requested):

src/odemis/model/_metadata.py
```diff
@@ -60,6 +60,7 @@
 MD_LENS_NA = "Lens numerical aperture"  # float (ratio), numerical aperture
 MD_LENS_RI = "Lens refractive index"  # float (ratio), refractive index
 MD_FILTER_NAME = "Filter name"  # str, product name of the light filter
+MD_DT_CORR = "Dwell time correction for EBIC"  # correction time needed for the EBIC component if not synchronized well
```
pieleric, line 63: I don't think that's needed.

pieleric (changes requested):

src/odemis/gui/cont/stream_bar.py
```diff
@@ -1658,38 +1658,41 @@ def addAR(self):
...
-                                  play=False
-                                  )
+        sem_ebic_stream = acqstream.SEMMDStream("SEM EBIC", [sem_stream, ebic_stream])
 
-        # With EBIC, often the user wants to get the whole area, same as the survey.
-        # But it's not very easy to select all of it, so do it automatically.
-        # (after the controller creation, to automatically set the ROA too)
-        if ebic_stream.roi.value == acqstream.UNDEFINED_ROI:
+        # Let the user select the ROI instead of using the full ROI automatically
+        if isinstance(ebic_stream, acqstream.EBICSettingsStream) and ebic_stream.roi.value == acqstream.UNDEFINED_ROI:
```
pieleric, line 1693: For now, we shouldn't distinguish between the types of EBIC streams. => always set the full RoI.

src/odemis/driver/ephemeron.py
sneepstefan: I have chosen a different approach.

src/odemis/driver/ephemeron.py
sneepstefan: Why does it matter that it is odd? It is just the way they created their MightyEBIC solution, defining if oversampling is unwanted a 0 should be passed. This is also contained in their example code of the client.
Regarding explaining the concepts of oversampling and samplesPerPixel I will ask Ephemeron and add that later (I'll put a todo there).

src/odemis/driver/ephemeron.py
sneepstefan: I made a shot at typing those ..

pieleric:
Replaced by PR #2920

PR #2820, opened 2024-07-09
[FIX] [METEOR-1127] Handle stitching tilesets with scan rotation

patrickcleeve2:
Users are reporting incorrect stitching of tilesets when using scan rotation. This PR adds support for identifying and handling scan rotated images in stitched tilesets


src/odemis/util/dataio.py
```diff
@@ -319,6 +320,17 @@ def leader_quality(da):
...
 
+        # handle scan rotation
+        dar = []
+        for da in das:
+            scan_rot = da.metadata.get(model.MD_BEAM_SCAN_ROTATION, 0)
+            if math.isclose(scan_rot, math.pi, abs_tol=0.1):
+                # rotate image content (Note: assumption is only scan rotation 180 is used.)
+                logging.debug(f"Found beam scan rotation ({math.degrees(scan_rot)}) rotating image content")
+                da = numpy.rot90(da, 2)
+            dar.append(da)
```
K4rishma, line 331: Can we add a note to flag rotation angle other than pi, and log warning that the stitching might be influenced due to that?

K4rishma (approved):

pieleric (changes requested):
We removed the rotation metadata (in commit 49de6ab6bd74) because that was showing the image rotated. Now we are trying to put it back "sometimes".

I think instead we should take a step back, and review how to handle the original issue.
At least, I think that all SEM metadata in TIFF should be handled the same way. So loading Zeiss, Tescan or OpenFIB images would work the same as TFS. 

We should also let the stitching function handle the rotation, as it appears to work well (for all cases, not just 180°). 

We still want that loaded SEM images are shown not rotated. Maybe only when loaded from the correlation tab? Or also in the "analysis" tab too?

I can think of a few alternatives:
* In the TIFF loader, store the scan rotation metadata as EBEAM_SCAN_ROTATION. In dataio.open_files_and_stitch(), detect this metadata, temporarily copy it on ROTATION, let the stitcher do its job, and then convert back ROTATION to EBEAM_SCAN_ROTATION afterwards.
* In the TIFF loader, store the scan rotation metadata as MD_ROTATION (as we used to do). Then, in the correlation tab, in `load_streams()`, so after the stitcher has run, if the acquisition type is EM, then convert the ROTATION metadata as EBEAM_SCAN_ROTATION.
* Let the TIFF loading store the scan rotation metadata as MD_ROTATION. In the correlation tab, after the user selects a file to open, have an "Import data" dialog box shown. It would allow the user to select how to interpret the metadata. Such as adjusting the rotation (and possibly set it to 0°).

What do you think? Do you have another alternative in mind?

patrickcleeve2:
> We removed the rotation metadata (in commit [49de6ab](https://github.com/delmic/odemis/commit/49de6ab6bd745cbae319420147e9cf157f9901a7)) because that was showing the image rotated. Now we are trying to put it back "sometimes".
> 
> I think instead we should take a step back, and review how to handle the original issue. At least, I think that all SEM metadata in TIFF should be handled the same way. So loading Zeiss, Tescan or OpenFIB images would work the same as TFS.
> 
> We should also let the stitching function handle the rotation, as it appears to work well (for all cases, not just 180°).
> 
> We still want that loaded SEM images are shown not rotated. Maybe only when loaded from the correlation tab? Or also in the "analysis" tab too?
> 
> I can think of a few alternatives:
> 
> * In the TIFF loader, store the scan rotation metadata as EBEAM_SCAN_ROTATION. In dataio.open_files_and_stitch(), detect this metadata, temporarily copy it on ROTATION, let the stitcher do its job, and then convert back ROTATION to EBEAM_SCAN_ROTATION afterwards.
> * In the TIFF loader, store the scan rotation metadata as MD_ROTATION (as we used to do). Then, in the correlation tab, in `load_streams()`, so after the stitcher has run, if the acquisition type is EM, then convert the ROTATION metadata as EBEAM_SCAN_ROTATION.
> * Let the TIFF loading store the scan rotation metadata as MD_ROTATION. In the correlation tab, after the user selects a file to open, have an "Import data" dialog box shown. It would allow the user to select how to interpret the metadata. Such as adjusting the rotation (and possibly set it to 0°).
> 
> What do you think? Do you have another alternative in mind?

I've fixed the other readers so they properly read MD_BEAM_SCAN_ROTATION and not MD_ROTATION. 

I've migrated the scan rotation handling inside the weaving (stitching) code, and its now just a single extra parameter read for scan rotation. I think it should be treated separately from MD_ROTATION because that has a lot of other consequences and not very keen on converting the metadata back and forth

src/odemis/dataio/tiff.py
```diff
@@ -2782,6 +2782,7 @@ def _convert_thermo_fisher_to_odemis_metadata(metadata: str) -> dict:
         md[model.MD_ROTATION] = 0  # float(tfs_md[beam_type]["scanrotation"])
```
pieleric, line 2782: Let's drop the MD_ROTATION entirely. The default is 0 anyway, and it highlights better that we don't really know. 
No need to keep the old version in a comment. We have git for that.

src/odemis/dataio/tiff.py
```diff
@@ -2911,6 +2912,7 @@ def _convert_zeiss_to_odemis_metadata(metadata: str) -> dict:
 
         # MD_POS should match image dimensions, e.g. only XY for 2D, XYZ for 3D, etc.
         md[model.MD_POS] = (x, y)
+        md[model.MD_ROTATION] = 0 # tmp: disable rotation metadata, as it is inconsistent with how odemis displays images
```
pieleric, line 2915: No need to add this. The default is 0.
You can put a comment on MD_BEAM_SCAN_ROTATION to indicate we don't use MD_ROTATION because we don't know the reason of the scan rotation in the first place.

src/odemis/dataio/tiff.py
```diff
@@ -3131,6 +3133,7 @@ def _convert_tescan_to_odemis_metadata(metadata: bytes) -> dict:
             md[model.MD_POS] = (float(tmd[beam_type]["stagex"]),
                                 float(tmd[beam_type]["stagey"]),
                             )
+            md[model.MD_ROTATION] = 0 # float(tmd[beam_type]["stager"])
```
pieleric, line 3136: Let's drop this. No need to (not?) associate it with "stage/r".
If you want, you could extend the metadata of about the stage position to be stored in MD_STAGE_POSITION_RAW (as a dict with "rz").

src/odemis/dataio/tiff.py
```diff
@@ -3233,6 +3236,7 @@ def _convert_openfibsem_to_odemis_metadata(metadata: str) -> dict:
             md[model.MD_POS] = (float(omd["microscope_state"]["stage_position"]["x"]),
                                 float(omd["microscope_state"]["stage_position"]["y"])
                                 )
+            md[model.MD_ROTATION] = 0  # float(omd["microscope_state"]["stage_position"]["r"])
```
pieleric, line 3239: Drop.

pieleric (changes requested):

pieleric (approved):

PR #2821, opened 2024-07-10
[fix] Add power metadata when using a spectrum stream with a light emitter

sneepstefan:
When using a spectrum stream with a light emitter connected which will control the power value of the emitter now this metadata is added to the stream as well.


patrickcleeve2 (approved):

pieleric (changes requested):
Normally the mdupdater can do this for you. You just need to put an affects on the light = [Camera, Spectrometer]. Can you check, and if this doesn't work, we can have a deeper look into it.

K4rishma (approved):

sneepstefan:
It seems the mdupdater indeed handles this as it should when the affects are correct. However since the stream is paused just before the acquisition starts, the actual power is changed to 0 due to safety choices -> When the stream is not playing the light source should be off.

Will close this PR and make new changes to the code responsible for handling the power VA on the SpectrumSettingsStream.

pieleric:
Fixed using the right "affects" in the microscope file. Added a test case in PR #2956 .

PR #2823, opened 2024-07-11
[feature] extend support for andor kymera 328i

pieleric:
The main goal is to add support for exchanging the turret on the Kymera 328. These changes also improve the support when having multiple Andor spectrographs (as the Kymera 328 is used in addition to the Kymera 193), and support reconnection even if the spectrograph power cannot be controlled (because we don't have power control for the Kymera 328).

### DIFF
--- a/src/odemis/driver/andorshrk.py
+++ b/src/odemis/driver/andorshrk.py
@@ -395,6 +395,8 @@ class Shamrock(model.Actuator):
         else:
             self._dll = ShamrockDLL()
 
+        self._removeNonImplementedFunctions()
+
         # Note: it used to need a "ccd" dependency, but not anymore
         self._camera = camera
         self._hw_access = HwAccessMgr(camera)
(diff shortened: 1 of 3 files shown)


K4rishma (approved):

src/odemis/driver/andorshrk.py
```diff
@@ -383,12 +386,17 @@ def __init__(self, name: str, role: str,
         if kwargs.get("inverted", None):
             raise ValueError("Axis of spectrograph cannot be inverted")
 
-        if device == "fake":
+        self._uses_shared_dll = (_dll is not None)
```
nandishjpatel, line 389:
```suggestion
        self._is_dll_shared = (_dll is not None)
```
Upto you which naming is better

src/odemis/driver/andorshrk.py
```diff
@@ -2072,6 +2156,115 @@ def scan():
...
+                    else:
+                        logging.info("Skipping Andor Shamrock with S/N %s", sn)
+
+                    # In (the unlikely) case we have found all the spectrographs we need, but there are
+                    # more spectrographs, stop early.
+                    if not shamrocks:
+                        break
+            except Exception:
+                # One of the child init went wrong, just clean up and pass on the bad news
+                self.Close()
```
nandishjpatel, line 2221:
```suggestion
                self.terminate()
```
Shouldn't you also terminate the appended children as a part of the clean-up?

src/odemis/driver/andorshrk.py
```diff
@@ -2072,6 +2156,115 @@ def scan():
...
+                raise
```
nandishjpatel, line 2222: Not sure, but shouldn't this be also some kind of HwError which is raised?

nandishjpatel (approved):
Minor comments

src/odemis/driver/andorshrk.py
pieleric: Sounds like  a better name, indeed. Thanks.

src/odemis/driver/andorshrk.py
pieleric: We `raise` to just pass on the exception. It might be a HwError, or something more "odd"... it's up to the Sharmrock class to decide.

PR #2826, opened 2024-07-16
[fix] Handle floating point error when computing axes ticks in gui.img.calculate_ticks()

pieleric:
For some very specific ranges, the calculation of the first tick value
would end up being very slightly less than the first value of the range
(eg, 1e-18 less). That would confuse the assertion later that all ticks
are within the range.
=> Make sure that the first tick is at least as big as the first value
in the range.

This avoids such error messages:
2024-07-12 15:39:59,308 ERROR   main:411:       Traceback (most recent call last):
  File "odemis/src/odemis/gui/comp/legend.py", line 495, in on_paint
    self._tick_list = calculate_ticks(self._value_range, csize, self._orientation, self._tick_spacing)
  File "odemis/src/odemis/gui/util/img.py", line 1077, in calculate_ticks
    pixel = value_to_pixel(tick_value, pixel_space, value_range, orientation)
  File "odemis/src/odemis/gui/util/img.py", line 899, in value_to_pixel
    assert value_range_finite[0] <= value <= value_range_finite[-1] or value_range_finite[-1] <= value <= value_range_finite[0]
AssertionError

### DIFF
--- a/src/odemis/gui/util/img.py
+++ b/src/odemis/gui/util/img.py
@@ -1055,7 +1055,7 @@ def calculate_ticks(value_range, client_size, orientation, tick_spacing):
                   value_step, value_space, tick_spacing)
 
     # Find the smallest value multiple of value_step, which is within the range
-    first_val = math.ceil(min_val / value_step) * value_step
+    first_val = max(min_val, math.ceil(min_val / value_step) * value_step)  # Clamp to min_val for floating point errors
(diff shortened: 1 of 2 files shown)


tepals (commented):
does it make sense to extend the unit tests to test for this?

nandishjpatel (approved):

tepals (approved):

PR #2832, opened 2024-07-18
[fix][METEOR-1026] Export OME-2016-06 compatible metadata + conversion script

patrickcleeve2:
The current metadata written by odemis was not valid ome, due to additional xml blocks and the imagej preamble in the image description.  This prevented other programs (imagej, tfs maps) from reading the metadata correctly, and therefore images were not displayed correctly (channel information, scale, position). 
This PR adds an option to save an additional image  with ome compatible metadata for use in other programs. It is enabled by default, and can be toggled in Menu/Development/Enable OME Compatibility. 

The current implementation is only enabled for meteor and will save an additional ome compatible image alongside the current image. Once tested by users and validated, it will be enabled as the default exporter for meteor.  

Added Dependencies (Python):
- ome_types
- tifffile

Add:
- [x] Conversion Script
- [x] Optional Exporter
- [x] Export full metadata
- [x] Reader for 2016-06 Metadata
- [x] Test Cases


src/odemis/dataio/tiff.py
```diff
@@ -2076,7 +2085,18 @@ def export(filename, data, thumbnail=None, compressed=True, multiple_files=False
     compressed (boolean): whether the file is compressed or not.
     multiple_files (boolean): whether the data is distributed across multiple
       files or not.
+    pyramid: whether to export data as pyramid
+    ome_compat: whether to export the file in OME-2016 format (in addition to standard metadata)
     '''
+
+    if ome_compat:
+        try:
+            new_fn = filename.replace(".ome.tiff", "-2016-06.ome.tiff")
```
pieleric, line 2094: Let's not change the file name. It should respect what is passed as input. Actually, I think we should just by default output TIFF file compliant with OME-TIFF 201606. 
There could be a separate "exporter" which output ImageJ compatible TIFF. That would be some ".ij.tiff" format, with an explicit extension defined. I think it could also contain OME-TIFF 201606 metadata (as for ImageJ, all that is needed is a short INI-style header, and it can be followed by an XML data).
The only thing we need to be careful is that the old OME-TIFF files generated by Odemis can still be read.

src/odemis/dataio/tiff.py
```diff
@@ -3306,3 +3332,290 @@ def _convert_openfibsem_to_odemis_metadata(metadata: str) -> dict:
...
+
+def reformat_ome_metadata(image_data: List[model.DataArray],
+                          filename: str,
+                          ) -> None:
+    """Save an odemis image with OME-XML format (2016-06)
+    :param image_data: (list of DataArray) image data
+    :param filename: (str) filename to save the image
+    """
+
+    import tifffile as tff
```
pieleric, line 3344: Please don't make a separate function for OME-TIFF 201606. It should be possible to just tweak the current code to "upgrade" from "201206++" to 201606. 
Here, you're introducing new dependencies, which is extra work to package and test, and especially, it's not clear what we are loosing. Does it support all the "special" metadata like MD_EXTRA_SETTINGS, or the metadata from the SPARC? Especially, does it support pyramidal imaging? We should aim at having one function that does things really well, instead of 2 equivalent functions with both their drawbacks and advantages.

pieleric (changes requested):

patrickcleeve2:
Closing, using #2890

PR #2833, opened 2024-07-18
[fix] [METEOR-1030] Prevent additional image groups in metadata being created for multi-channel z-stacks

patrickcleeve2:
When saving z-stacks additional images (groups) were being created in the metadata. This is because the check assumed all data was 5D (CTZYX) or 2D (YX) and only checked shape[0] assuming it was the channel dim, for if the group could be merged.

Since the z-stacks are 3D (ZYX) they were being added as a separate group, with incorrect channel information (TheC, FirstC) which caused problems with downstream readers (bio-formats).

When images were re-imported, they are reshaped into 5D, so when re-exporting them, the image was 'fixed'. This only occurred for multi-channel z-stacks, as the other checks caught other cases.

### DIFF
--- a/src/odemis/dataio/tiff.py
+++ b/src/odemis/dataio/tiff.py
@@ -1205,29 +1205,52 @@ def _findImageGroups(das):
     groups = dict()
     current_ifd = 0
     prev_da = None
+    plen_c = None
 
+    # required keys for merging
+    required_keys = [model.MD_IN_WL, model.MD_OUT_WL]
+    # metadata -> default value if missing
+    matching_keys = {
+        model.MD_HW_NAME: None,
+        model.MD_HW_VERSION: None,
+        model.MD_PIXEL_SIZE: None,
+        model.MD_POS: None,
+        model.MD_ROTATION: 0,
+        model.MD_SHEAR: 0,
+    }
     for da in das:
         # check if it can be part of the current group (compare just to the previous DA)
         dims = da.metadata.get(model.MD_DIMS, "CTZYX"[-da.ndim::])
-        if (prev_da is None
-            or da.shape[0] != 1  # If C != 1 => not possible to merge (C is always first dimension) or it is a
-                # greyscale YX image
-            or (model.MD_IN_WL not in da.metadata or model.MD_OUT_WL not in da.metadata)
-            or prev_da.shape != da.shape
-            or prev_da.metadata.get(model.MD_HW_NAME, None) != da.metadata.get(model.MD_HW_NAME, None)
-            or prev_da.metadata.get(model.MD_HW_VERSION, None) != da.metadata.get(model.MD_HW_VERSION, None)
-            or prev_da.metadata.get(model.MD_PIXEL_SIZE) != da.metadata.get(model.MD_PIXEL_SIZE)
-            or prev_da.metadata.get(model.MD_POS) != da.metadata.get(model.MD_POS)
-            or prev_da.metadata.get(model.MD_ROTATION, 0) != da.metadata.get(model.MD_ROTATION, 0)
-            or prev_da.metadata.get(model.MD_SHEAR, 0) != da.metadata.get(model.MD_SHEAR, 0)
-           ):
-            # new group
+
+        len_c = 1
+        if "C" in dims:
+            len_c = da.shape[dims.index("C")]
+
+        # conditions for merging:
+        # there is a previous data array
+        # there is one channel for both arrays
+        # the dimensions/shapes are the same
+        # it looks like an FM image (ie, WL_IN and WL_OUT are present in the metadata)
+        # all the required keys are present in the metadata
+        # all the matching keys have the same value in the metadata
+        # otherwise, create a new group
+        if (prev_da is not None
+            and (plen_c == 1 and len_c == 1)
+            and prev_da.shape == da.shape
+            and all(k in da.metadata for k in required_keys)
+            and all(prev_da.metadata.get(k, v) == da.metadata.get(k, v) for k, v in matching_keys.items())
+        ):
+            # merge with the existing group
+            pass
+        else:
+            # create a new group
             group_ifd = current_ifd
         groups.setdefault(group_ifd, []).append(da)
 
         # increase ifd by the number of planes
         current_ifd += _countNeededIFDs(da)
-        prev_da = da
+        prev_da, plen_c = da, len_c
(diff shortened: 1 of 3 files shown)


src/odemis/dataio/tiff.py
```diff
@@ -1221,13 +1224,20 @@ def _findImageGroups(das):
             or prev_da.metadata.get(model.MD_ROTATION, 0) != da.metadata.get(model.MD_ROTATION, 0)
             or prev_da.metadata.get(model.MD_SHEAR, 0) != da.metadata.get(model.MD_SHEAR, 0)
            ):
-            # new group
-            group_ifd = current_ifd
+            # don't add a new group if passed group check and it's z-stacks
+            if prev_da.metadata.get(model.MD_DIMS, None) == "ZYX" and dims == "ZYX" and da.ndim == 3 and prev_da.ndim == 3:
```
pieleric, line 1228: I think it's overly catching cases. Essentially, it's now merging any DataArray which is of shape ZYX. Including possibly SEM data, or ZYX of different sizes, or ZYX at different positions in XY.

I would suggest to just modify the original condition to be more specific and say "if previous image and current image both have C dimension == 1", which the default C dimension being of length 1. 
You can use something like this to read the C dimension:
```
        dims = da.metadata.get(model.MD_DIMS, "CTZYX"[-da.ndim::])
        len_c = 1
        if "C" in dims:
            len_c = da.shape[dims.index("C")]
```

src/odemis/dataio/tiff.py
```diff
@@ -1209,8 +1209,11 @@ def _findImageGroups(das):
     for da in das:
         # check if it can be part of the current group (compare just to the previous DA)
         dims = da.metadata.get(model.MD_DIMS, "CTZYX"[-da.ndim::])
-        if (prev_da is None
```
pieleric, line 1212: I feel that this code is extra difficult to read because of the inverted logic. Maybe you could inverted back, with something like:
```
if prev_da is not None and prev_da.shape == da.shape...:
    log("merging")
else:
    new group
```

pieleric (changes requested):
Please don't forget to add a test case. For this type of corner case, it's super important, otherwise, in a couple of years, we break it again without noticing.

src/odemis/dataio/tiff.py
```diff
@@ -1206,28 +1206,55 @@ def _findImageGroups(das):
...
+        # if there is no previous data array, create a new group
+        # if there is more than one channel, create a new group
+        # if the dimensions are different, create a new group
+        # if WL_IN or WL_OUT are not present (looks like FM image) (the metadata is missing), create a new group
+        # if the metadata is different, create a new group
+        if prev_da is None:
+            # create the first image group
+            group_ifd = current_ifd
+            logging.debug(f"creating initial image group: group id: {group_ifd}, current_ifd: {current_ifd}")
+
```
pieleric, line 1242: I don't think it's necessary to have a special `if` here. It just does the same as for the `elif`. => merge into a single `if` statement.

src/odemis/dataio/tiff.py
```diff
@@ -1206,28 +1206,55 @@ def _findImageGroups(das):
...
-            or (model.MD_IN_WL not in da.metadata or model.MD_OUT_WL not in da.metadata)
+
+
+        plen_c, len_c = 1, 1
+        if "C" in dims:
+            len_c = da.shape[dims.index("C")]
+        if prev_da is not None:
+            prev_dims = prev_da.metadata.get(model.MD_DIMS, "CTZYX"[-prev_da.ndim::])
+            if "C" in prev_dims:
+                plen_c = prev_da.shape[prev_dims.index("C")]
```
pieleric, line 1230: Instead of explicitly computing it, you can do the same as `prev_da`: start with `plen_c = None`, and at the end of the loop body, do a `plen_c = len_c`.

pieleric (changes requested):

K4rishma:
It would be a worthwhile to check if the test cases involving findImageGroups by using .export are happy. The relevant test  files are img_test.py, stream_test.py, calibration_test.py, dataio_test.py, stiff_test.py, and tiif_test.py.

K4rishma (approved):

src/odemis/dataio/tiff.py
```diff
@@ -1206,28 +1206,55 @@ def _findImageGroups(das):
...
+            # create a new group
             group_ifd = current_ifd
+            logging.debug(f"creating new image group: group id: {group_ifd}, current_ifd: {current_ifd}")
         groups.setdefault(group_ifd, []).append(da)
+        logging.debug(f"end of data array: group id: {group_ifd}, current_ifd: {current_ifd}, current groups: {groups.keys()}")
 
         # increase ifd by the number of planes
         current_ifd += _countNeededIFDs(da)
         prev_da = da
+    logging.debug(f"found {len(groups)} groups of images. final groups: {groups.keys()}")
```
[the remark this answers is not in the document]
patrickcleeve2: remove most, leaving one final message to assist future debugging

src/odemis/dataio/tiff.py
patrickcleeve2: I've inverted it, and update comments. should be easier to read

src/odemis/dataio/tiff.py
```diff
@@ -1206,28 +1206,50 @@ def _findImageGroups(das):
...
-            or prev_da.metadata.get(model.MD_HW_NAME, None) != da.metadata.get(model.MD_HW_NAME, None)
-            or prev_da.metadata.get(model.MD_HW_VERSION, None) != da.metadata.get(model.MD_HW_VERSION, None)
-            or prev_da.metadata.get(model.MD_PIXEL_SIZE) != da.metadata.get(model.MD_PIXEL_SIZE)
-            or prev_da.metadata.get(model.MD_POS) != da.metadata.get(model.MD_POS)
-            or prev_da.metadata.get(model.MD_ROTATION, 0) != da.metadata.get(model.MD_ROTATION, 0)
-            or prev_da.metadata.get(model.MD_SHEAR, 0) != da.metadata.get(model.MD_SHEAR, 0)
-           ):
-            # new group
+
+        plen_c, len_c = 1, 1
```
pieleric, line 1224: You shouldn't reset `plen_c` to 1 every time.
Just set it to `None` before the `for` loop, like `prev_da`.

pieleric (changes requested):

pieleric (approved):

PR #2840, opened 2024-07-24
[fix] Fast-EM optical autofocus tests

tepals:
Fix an issue in the binary autofocusing algorithm that caused the optical autofocus to fail.

### DIFF
--- a/src/odemis/acq/align/autofocus.py
+++ b/src/odemis/acq/align/autofocus.py
@@ -313,10 +313,6 @@ def _DoBinaryFocus(future, detector, emt, focus, dfbkg, good_focus, rng_focus):
             if future._autofocus_state == CANCELLED:
                 raise CancelledError()
 
-            if left == right:
-                logging.info("Seems to have reached minimum step size (at %g m)", 2 * step_factor * min_step)
-                break
-
             # if best focus was found at the center
             if i_max == 1:
                 step_factor /= 2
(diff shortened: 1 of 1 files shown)


src/odemis/driver/piezomotor.py
```diff
@@ -1138,7 +1138,8 @@ def _parseMessage(self, msg):
 
     def move(self, steps, target_pos):
         # simple move, same duration for every length, don't care about speed
-        self.executor.submit(self._do_move, steps, target_pos)
+        f = self.executor.submit(self._do_move, steps, target_pos)
+        f.result(timeout=900)
```
pieleric, line 1142: I don't think that's correct. Because now the command is blocking. Typically the hardware accepts the command, and immediately returns.

If the driver needs to wait until the move is complete. It should be done in the actual driver class, in `_waitEndMotion()`. Or maybe this part is not correctly simulated?

pieleric (changes requested):

src/odemis/driver/piezomotor.py
```diff
@@ -1139,6 +1139,8 @@ def _parseMessage(self, msg):
     def move(self, steps, target_pos):
         # simple move, same duration for every length, don't care about speed
         self.executor.submit(self._do_move, steps, target_pos)
+        # f = self.executor.submit(self._do_move, steps, target_pos)
+        # f.result(timeout=900)
```
pieleric, line 1143: Should be dropped?

src/odemis/driver/test/piezomotor_test.py
```diff
@@ -34,6 +34,7 @@
 # Export TEST_NOHW=1 to force using only the simulator and skipping test cases
 # needing real hardware
 TEST_NOHW = (os.environ.get("TEST_NOHW", "0") != "0")  # Default to Hw testing
+TEST_NOHW = 1
```
pieleric, line 37: Drop before merge.

pieleric (changes requested):

src/odemis/driver/test/piezomotor_test.py
tepals: yes, I converted the whole PR to a draft, since I did not manage to fix it yet. I think I will just close it, since it will be a while before I can work on this again.

pieleric (approved):

nandishjpatel (approved):

PR #2842, opened 2024-07-25
[FIX] Hide optical stream panel for meteor overview acq

patrickcleeve2:
Hide the optical stream panel for METEOR overview acquisition as it is never used and contains a misleading message "No streams available". 

I think it was accidentally re-enabled in the overview refactor

### DIFF
--- a/src/odemis/gui/win/acquisition.py
+++ b/src/odemis/gui/win/acquisition.py
@@ -660,9 +660,8 @@ class OverviewAcquisitionDialog(xrcfr_overview_acq):
         # feature flag to enable/disable FIBSEM mode (disabled until fibsem code is merged)
         self.fibsem_mode = False # isinstance(orig_tab_data, guimodel.CryoFIBSEMGUIData)
 
-        # hide optical settings / stream panel when in fibsem mode
+        # hide optical settings when in fibsem mode
         self.fp_settings_secom_optical.Show(not self.fibsem_mode)
-        self.pnl_opt_streams.Show(not self.fibsem_mode)
 
         self.filename = create_filename(save_dir, "{datelng}-{timelng}-overview",
                                               ".ome.tiff")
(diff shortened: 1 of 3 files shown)


src/odemis/gui/win/acquisition.py
```diff
@@ -658,9 +658,11 @@ def __init__(self, parent, orig_tab_data):
         # feature flag to enable/disable FIBSEM mode (disabled until fibsem code is merged)
         self.fibsem_mode = False # isinstance(orig_tab_data, guimodel.CryoFIBSEMGUIData)
 
-        # hide optical settings / stream panel when in fibsem mode
+        # hide optical settings when in fibsem mode
         self.fp_settings_secom_optical.Show(not self.fibsem_mode)
-        self.pnl_opt_streams.Show(not self.fibsem_mode)
+
+        # hide optical streams (never used for meteor)
+        self.pnl_opt_streams.Show(self._main_data_model.role != "meteor")
```
K4rishma, line 665: Is it used in Enzel and Secom?

K4rishma (approved):

src/odemis/gui/win/acquisition.py
patrickcleeve2: secom maybe, enzel don't think so

pieleric: I had a quick look, and it seems this stream panel is useless. The frame has a pnl_opt_streams, and a pnl_secom_streams. Only the second one seems ever used in the code. I guess it's due to copy/paste from another frame. (BTW, the names don't really make sense, as the SECOM is supposed to have SEM and optical streams... but that's something else to worry about). 

Could you double-check is not used anywhere (ie, SECOM and ENZEL too). Then, if it's indeed always empty, just delete the panel from the frame.

pieleric (changes requested):

pieleric (approved):

PR #2843, opened 2024-07-25
[perf] SPARC CL intensity with drift correction could run with many small non-uniform blocks

pieleric:


src/odemis/acq/drift/__init__.py
```diff
@@ -216,7 +216,7 @@ def estimateCorrectionPeriod(period, t, repetitions):
         :param t: (float) Time spend for acquiring one data at the fastest dimension.
         :param repetitions: (tuple of 2 ints) Number of iterations for the two fastest axes in the
                                               entire drift-corrected acquisition. First value is the
-                                              fastest dimension scanned. Usually it is the
+                                              slowest dimension scanned. Usually it is the
                                               number of px (ebeam) positions. Sometimes fastest dimension
                                               can be e.g. number of images to integrate.
```
tepals, line 221: I'm not sure if I understand the rest of the comment, should that now be the following?
```suggestion
                                              slowest dimension scanned. Usually the fastest dimension is the
                                              number of px (ebeam) positions, sometimes it
                                              can be e.g. number of images to integrate.
```

src/odemis/acq/stream/_sync.py
```diff
@@ -2324,6 +2324,38 @@ def _adjustHardwareSettings(self):
...
+    def _get_next_rectangle(self, rep: Tuple[int, int], acq_num: int, px_time: float,
+                            leech_np: List[Optional[int]]) -> Tuple[int, int]:
+        """
+        Get the next rectangle to scan, based on the leeches and the live update period.
+        :param rep: total number of pixels to scan (y, x)
+        :param acq_num: number of pixels scanned so far
+        :param px_time: time to acquire one pixel
+        :param leech_np: number of pixels to scan before the next run of each leech.
+        If None, the leech doesn't need to run.
+        :return: size of the rectangle to scan (y, x)
```
tepals, line 2336: can you specify if this is also in pixels?

src/odemis/acq/stream/_sync.py
```diff
@@ -2324,6 +2324,38 @@ def _adjustHardwareSettings(self):
...
+        """
+        tot_num = rep[0] * rep[1]
+        # The very max number of pixels to acquire
+        npixels2scan = min([np for np in leech_np if np is not None] + [tot_num - acq_num])
```
tepals, line 2340: I don't understand why you use `min` if it is supposed to be the max number of pixels to acquire

tepals (changes requested):

nandishjpatel (approved):

src/odemis/acq/drift/__init__.py
pieleric: Good catch, the whole text is actually about the "first dimension", which is the slowest, but I had updated only half of it. I've now updated it further.

tepals (approved):

PR #2847, opened 2024-08-02
[fix] acq.Stream.getPixelCoordinates() shouldn't detect pixel -0.5,-0.5 as inside the image

pieleric:
getPixelCoordinates() returns None if the coordinates are inside the
image. If going just out the image on the left or top, the X/Y value
could be between -1 and 0. However, in Python "int()" trunc the value to
0, and would claim this position is within the image, while it's just
outside.
=> Use floor() to properly detect it.

Note: it's most noticeable on data sets with just 1x1 pixel in X/Y.
Still no one had really noticed in a few years.

### DIFF
--- a/src/odemis/acq/stream/_base.py
+++ b/src/odemis/acq/stream/_base.py
@@ -15,13 +15,20 @@ Odemis is distributed in the hope that it will be useful, but WITHOUT ANY WARRAN
 You should have received a copy of the GNU General Public License along with Odemis. If not, see http://www.gnu.org/licenses/.
 '''
 
-from collections.abc import Iterable
 import functools
 import gc
 import logging
 import math
 import numbers
+import threading
+import time
+import weakref
+from collections.abc import Iterable
+from typing import Tuple, Optional
+
+import matplotlib
 import numpy
+
 from odemis import model
 from odemis.model import (MD_POS, MD_PIXEL_SIZE, MD_ROTATION, MD_ACQ_DATE,
                           MD_SHEAR, VigilantAttribute, VigilantAttributeBase,
@@ -31,11 +38,6 @@ from odemis.model import (MD_POS, MD_PIXEL_SIZE, MD_ROTATION, MD_ACQ_DATE,
                           MD_POL_EY, MD_POL_EZ, MD_POL_DOP, MD_POL_DOLP, MD_POL_DOCP, MD_POL_UP, MD_POL_DS1N,
                           MD_POL_DS2N, MD_POL_DS3N, MD_POL_S1N, MD_POL_S2N, MD_POL_S3N, TINT_FIT_TO_RGB, TINT_RGB_AS_IS)
 from odemis.util import img
-import threading
-import time
-import weakref
-import matplotlib
-
 # TODO: move to odemis.acq (once it doesn't depend on odemis.acq.stream)
 # Contains the base of the streams. Can be imported from other stream modules.
 # to identify a ROI which must still be defined by the user
@@ -1277,14 +1279,11 @@ class Stream(object):
 
         self._shouldUpdateImage()
 
-    def getPixelCoordinates(self, p_pos):
+    def getPixelCoordinates(self, p_pos: Tuple[float, float]) -> Optional[Tuple[int, int]]:
         """
         Translate physical coordinates into data pixel coordinates
-        Args:
-            p_pos(tuple float, float): the position in physical coordinates
-
-        Returns(tuple int, int or None): the position in pixel coordinates or None if it's outside of the image
-
+        :param p_pos: the position in physical coordinates (m)
+        :returns: the position in pixel coordinates or None if it's outside of the image
         """
         if not self.raw:
             raise LookupError("Stream has no data")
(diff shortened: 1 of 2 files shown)


src/odemis/acq/stream/_base.py
```diff
@@ -1302,7 +1302,8 @@ def getPixelCoordinates(self, p_pos):
         tform = AffineTransform(matrix, translation)
         pixel_pos_c = tform.inverse().apply(p_pos)
         # a "-" is used for the y coordinate because Y axis has the opposite direction in physical coordinates
-        pixel_pos = int(pixel_pos_c[0] + size[0] / 2), - int(pixel_pos_c[1] - size[1] / 2)
+        pixel_pos = (int(math.floor(pixel_pos_c[0] + size[0] / 2)),
+                     int(math.floor(- (pixel_pos_c[1] - size[1] / 2))))
         if 0 <= pixel_pos[0] < size[0] and 0 <= pixel_pos[1] < size[1]:
             return pixel_pos
```
tmoerkerken, line 1308: According to the docstring, the returned value could be a single `int`. I don't see how that can happen from the code (not related to your change by the way).

src/odemis/acq/stream/_base.py
```diff
@@ -1302,7 +1302,8 @@ def getPixelCoordinates(self, p_pos):
         tform = AffineTransform(matrix, translation)
         pixel_pos_c = tform.inverse().apply(p_pos)
         # a "-" is used for the y coordinate because Y axis has the opposite direction in physical coordinates
-        pixel_pos = int(pixel_pos_c[0] + size[0] / 2), - int(pixel_pos_c[1] - size[1] / 2)
+        pixel_pos = (int(math.floor(pixel_pos_c[0] + size[0] / 2)),
+                     int(math.floor(- (pixel_pos_c[1] - size[1] / 2))))
```
tmoerkerken, line 1306: Is `floor` not problematic for the positive scenario, where going outside of the image in the positive axes now makes it "in image"?

tmoerkerken (changes requested):

src/odemis/acq/stream/_base.py
nandishjpatel:
```suggestion
                     # Calculate pixel positions
                     x_pos_c = pixel_pos_c[0] + size[0] / 2
                     y_pos_c = -(pixel_pos_c[1] - size[1] / 2)

                    # Apply floor or ceil based on the sign
                    x_pos = math.floor(x_pos_c) if x_pos_c < 0 else math.ceil(x_pos_c)
                    y_pos = math.floor(y_pos_c) if y_pos_c < 0 else math.ceil(y_pos_c)
                    pixel_pos = (x_pos, y_pos)
```

Maybe

nandishjpatel (approved):

src/odemis/acq/stream/_base.py
tmoerkerken: So this would functionally alter the in range values (since `int` cast 'rounds' towards zero, but now the values are ceiled). To make if functionally equivalent, the default (only for this range, since doesn't hold for negative ranges) should be `floor`. I can also imagine that we want to `math.round` instead, before casting to `int`.

tmoerkerken, line 1306: Now that we are touching this code anyway, maybe explain with a comment why we are adding `shape / 2` in the mix here. I can imagine it is because we want the origin to be in the image center, but a comment could really help clarifying intentions.

tmoerkerken (changes requested):

src/odemis/acq/stream/_base.py
tmoerkerken, line 1308: I think this function yearns for a unit test :smile:

tmoerkerken (changes requested):

src/odemis/acq/stream/_base.py
pieleric: The docstring is supposed to mean it'll return either a tuple[int, int] or a None. However, the "," makes it confusing. I've changed it to use the Python typing format, which is a lot less ambiguous.

src/odemis/acq/stream/_base.py
pieleric: No. We just want to use `floor()` all the time. That's it.
To give an example, using a little extreme, but obvious values, imagine an image of resolution 1x1. In the "pixel" coordinates, the image goes from 0.0 -> 1.0 . And any position between 0.0 and 1.0 is on pixel 0. Position -0.001 is outside the image. Position 1.001 is outside the image. That's why `floor()` does exactly what we need.

pieleric: I added a comment. Indeed, it's because the image metadata refers to the center of the image, while the pixel coordinates are always relative to the top-left pixel.

tmoerkerken: I see now! Might also be helpful to add something similar as a comment, since this explanation was clarifying.

tmoerkerken (approved):

PR #2848, opened 2024-08-02
[testing] extend acq.leech test and run in CI

pieleric:
Add an explicit test for `acq.leech.get_next_rectangle()`, and also run the test cases for leech in the CI, as it doesn't need the back-end.


nandishjpatel (approved):

tmoerkerken (approved):
I'm not yet familiar with the leech functionality, but the test programmatically looks sane to me

PR #2853, opened 2024-08-07
[fix] also hide the tab buttons if there is only 1 tab or no tabs

nandishjpatel:


tepals (approved):

K4rishma (approved):

src/odemis/gui/cont/tabs/tab_bar_controller.py
```diff
@@ -192,8 +192,8 @@ def _create_needed_tabs(self, tab_defs, main_frame, main_data):
             if b not in buttons:
                 b.Hide()
 
-        # hide tab buttons for viewer
-        if main_data.is_viewer:
+        # Hide tab buttons if the microscope is a viewer or there is only 1 tab or no tabs
```
pieleric, line 195: This comment only explains what the code does, which is not really easier than just reading the code. Could you explain *why* instead? In particular, why checking for the viewer explicitly? Is that because it has 2 tabs, but one hidden?

nandishjpatel: @patrickcleeve2 what is the use of `is_viewer` flag? I see in the main gui data this is check: https://github.com/nandishjpatel/odemis/blob/6a5a9d93ed824dde9ce6907ee7efa7cf15014895/src/odemis/gui/model/main_gui_data.py#L394

patrickcleeve2: is_viewer is true if the app is opened as odemis viewer. When we added the correlation tab to the odemis viewer, we decided it should be disabled by default. So added check to hide the tab panels, until the user explicitly enables the correlation tab.

nandishjpatel: Thanks, I update the comment accordingly then

pieleric (approved):

PR #2798, opened 2024-06-12
[REFACTOR] [METEOR-1119] Refactor overview / tiled acquisition code

patrickcleeve2:
The previous overview acquisition code was fairly hardcoded for specific fm based imaging and coupled to the gui. The main components are:

**Utilities**:
- Move overview acquisition utils from gui methods into generic utilities allowing them to be used outside the user interface.
- Added test cases for tiling utilities.

**GUI**
- Added support for a acquiring overviews with FIB/SEM, and hide unused settings (e.g. z-stacks) for each mode.
- Allow using AcquireOverview task without autofocus
- Allow multi-area acquisitions without autofocus.
- Consolidate different code paths when using autofocus vs not.
- Removed direct call to AcquireTiledArea task, all paths now go through AcquireOverview.

PR #2855, opened 2024-08-08
[refactor] tiled area code to have more separation of concerns

tmoerkerken:
This PR proposes a solution to resolve the bounding box related bug found during integration testing. After inspection, the bug was always there, but it only reached the surface after Patrick's PR (https://github.com/delmic/odemis/pull/2798).  Changes:

- The `get_tiled_areas` function is now too convoluted in my opinion (it has two responsibilities), making it unnecessarily complex to read it and to unit test it. My proposal is to separate concerns, and split the original function into two functions and handle the paths on a higher level. 
- The original fallback was heuristic. I removed the fallback and raise an error instead, so the developer will be forced to investigate if the bounding box is applicable to the specific use case and set it accordingly (like @nandishjpatel did in `CryoMainGUIData`).
- Fixed a [bug](https://github.com/delmic/odemis/blob/6a5a9d93ed824dde9ce6907ee7efa7cf15014895/src/odemis/acq/stitching/_tiledacq.py#L1303C1-L1304C42) where the y tiling used the x repetition number. Probably never reached the surface, since a symmetrical grid is probably most often used. A regression test is added accordingly.
- Naming: the variable name `area` is ambiguous. It was now used for a bounding box, but it is easily mixed up with "surface area". I refactored most lower level functions to use `bbox` now.

TODO:
- [x] Fix lint
- [x] Update tests to reflect new functions
- [x] Publish PR
- [x] Check if Cryo workflow still works as intended. After testing: Even though Cryo ("meteor", "enzel", "mimas") has a potential path where the alternative custom bbox function is used, it is not tested by the simulator (self.stage.getMetadata().get(model.MD_SAMPLE_CENTERS) is None). It would be nice to have both paths tested in the GUI. At least we have unit tests.
- [x] Add regression test for non-square grid

### DIFF
--- a/src/odemis/acq/stitching/_tiledacq.py
+++ b/src/odemis/acq/stitching/_tiledacq.py
@@ -64,12 +64,6 @@ SAFE_REL_RANGE_DEFAULT = (-50e-6, 50e-6)  # m
 # Increasing this distance may result in out of focus overview acquisition image
 MAX_DISTANCE_FOCUS_POINTS = 450e-06  # in m
 
-# cryo-TEM grid sizes
-SAMPLE_RADIUS_TEM_GRID = 1.25e-3  # m, standard TEM grid size including the borders
-# Bounding-box relative to the center of a sample, corresponding to usable area
-# for imaging/milling. Used in particular for the overview image.
-hwidth = SAMPLE_RADIUS_TEM_GRID / math.sqrt(2)
-SAMPLE_USABLE_BBOX_TEM_GRID = (-hwidth, -hwidth, hwidth, hwidth)  # m, minx, miny, maxx, maxy
 DEFAULT_FOV = (100e-6, 100e-6) # m
 
 class FocusingMethod(Enum):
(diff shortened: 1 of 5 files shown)


tepals:
It looks good, typically we try to be hesitant with renaming functions, because it is an open source repository. But I think in this case it's fine, since it just got added and is not used in a lot of places.
@patrickcleeve2 can you have a look at this since it is mostly based on code you wrote.

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -1208,47 +1202,54 @@ def run(self):
...
+        fov=fov,
+        tiles_nx=tiles_nx,
+        tiles_ny=tiles_ny,
+        tiling_rng=tiling_rng,
+        overlap=overlap,
+    )
+
+
+def get_tiled_bboxes(
+    rel_bbox: tuple,
```
patrickcleeve2, line 1243: Add type annotation
```suggestion
    rel_bbox: Tuple[float],
```

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -1208,47 +1202,54 @@ def run(self):
...
+    sample_centers: Dict[str, Tuple[float]] = None,
```
patrickcleeve2, line 1244: remove default arg for sample_centers, not possible after refactor
```suggestion
    sample_centers: Dict[str, Tuple[float]],
```

src/odemis/gui/win/acquisition.py
```diff
@@ -905,19 +907,32 @@ def _get_areas(self) -> List[Tuple[float]]:
...
+                        streams=self.get_acq_streams(),
+                        tiles_nx=self.tiles_nx.value,
+                        tiles_ny=self.tiles_ny.value,
+                        overlap=self.overlap,
+                        tiling_rng=self._tiling_rng,
+            )
+            # Cast to list, to have a consistent format with the alternative path
+            areas = [area] if area is not None else []
+        else:
+            if not hasattr(self._main_data_model, "sample_rel_bbox"):
```
patrickcleeve2, line 922: This should check sample_centers, because currently every cryo app has sample_rel_bbox: https://github.com/tmoerkerken/odemis/blob/f32a634a1faa233fbdafc00f7dab2a1712871436/src/odemis/gui/model/main_gui_data.py#L501. The whole_grid_chkbox gets hidden/disabled based on the sample_centers too
```suggestion
            if not self._main_data_model.sample_centers:
```

src/odemis/gui/win/acquisition.py
```diff
@@ -905,19 +907,32 @@ def _get_areas(self) -> List[Tuple[float]]:
...
+                raise NotImplementedError(
+                    "The current data model does not have a relative bounding box defined",
```
patrickcleeve2, line 924: update comment based on above

patrickcleeve2 (changes requested):

src/odemis/gui/win/acquisition.py
```diff
@@ -905,19 +907,32 @@ def _get_areas(self) -> List[Tuple[float]]:
...
+                    "If a heuristic bounding box is desired, it can be implemented in ",
+                    "a specific MainGUIData model."
+                )
+
+            sample_centers = [pos for name, pos in self._main_data_model.sample_centers.items()
+                if name in self._selected_grids]
+
+            areas = get_tiled_bboxes(
+                sample_centers=sample_centers,
+                rel_bbox=self._main_data_model.rel_bbox,
```
tmoerkerken, line 934: Should be .sample_rel_bbox

src/odemis/gui/win/acquisition.py
tmoerkerken: Added

src/odemis/gui/win/acquisition.py
tmoerkerken: Fixed

patrickcleeve2 (approved):

patrickcleeve2:
@tmoerkerken can you rebase and squash the commits into one

tepals (approved):

PR #2860, opened 2024-08-19
[METEOR-1163][FEAT] Add changeable autofocus distance for overview acquisition

patrickcleeve2:
Allow the user to change the distance between autofocus for overview acquisition. 

Previously this was hardcoded at 450um, which meant that for overviews smaller than that only a single autofocus in the centre of the image was done. This was particularly bad on systems with 100x mag

### DIFF
--- a/src/odemis/gui/main_xrc.py
+++ b/src/odemis/gui/main_xrc.py
@@ -124,6 +124,8 @@ class xrcfr_overview_acq(wx.Dialog):
         self.selected_grid_pnl_holder = xrc.XRCCTRL(self, "selected_grid_pnl_holder")
         self.area_size_txt = xrc.XRCCTRL(self, "area_size_txt")
         self.autofocus_chkbox = xrc.XRCCTRL(self, "autofocus_chkbox")
+        self.focus_points_dist_lbl = xrc.XRCCTRL(self, "focus_points_dist_lbl")
+        self.focus_points_dist_ctrl = xrc.XRCCTRL(self, "focus_points_dist_ctrl")
         self.gauge_acq = xrc.XRCCTRL(self, "gauge_acq")
         self.lbl_acqestimate = xrc.XRCCTRL(self, "lbl_acqestimate")
         self.btn_cancel = xrc.XRCCTRL(self, "btn_cancel")
@@ -1931,7 +1933,7 @@ b\xeb\x85\x9f\xb6B\x1d\x0cK\x17\xac\xf0\x12\xfe\xa0\xe5\xee\xe03\xb1\xfa\
                   </object>
                   <object class="sizeritem">
                     <object class="wxCheckBox" name="autofocus_chkbox">
-                      <label>Run autofocus</label>
+                      <label>Run AutoFocus</label>
                       <fg>#E5E5E5</fg>
                       <XRCED>
                         <assign_var>1</assign_var>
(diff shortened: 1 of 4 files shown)


src/odemis/gui/xmlh/resources/dialog_overview_acq.xrc
```diff
@@ -317,6 +317,42 @@
                     <flag>wxLEFT</flag>
                     <border>10</border>
                   </object>
+                  <object class="sizeritem">
+                    <object class="wxBoxSizer">
+                      <orient>wxHORIZONTAL</orient>
+                      <object class="sizeritem">
+                        <object class="wxStaticText" name="autofocus_dist_lbl">
+                          <label>Distance between AutoFocus</label>
```
K4rishma, line 325: Maybe "Distance between Focus Points"

src/odemis/gui/main_xrc.py
```diff
@@ -1940,6 +1942,42 @@ def __init_resources():
                     <flag>wxLEFT</flag>
                     <border>10</border>
                   </object>
+                  <object class="sizeritem">
+                    <object class="wxBoxSizer">
+                      <orient>wxHORIZONTAL</orient>
+                      <object class="sizeritem">
+                        <object class="wxStaticText" name="autofocus_dist_lbl">
+                          <label>Distance between AutoFocus</label>
```
K4rishma, line 1950: Same comment here

K4rishma:
Maybe rename autofocus_dist to focus_points_dist? autofocus_dist seems to be setting the objective stage distance

src/odemis/gui/win/acquisition.py
```diff
@@ -783,6 +784,10 @@ def __init__(self, parent, orig_tab_data):
         self.btn_secom_acquire.Bind(wx.EVT_BUTTON, self.on_acquire)
         self.Bind(wx.EVT_CLOSE, self.on_close)
 
+        # range for the autofocus
+        self.autofocus_dist.SetValueRange(50e-6, 1000e-6)       # in m
```
K4rishma, line 788: Maybe good to make sure that the distance does not exceed the selected area for the overview acquisition

patrickcleeve2: good suggetion, done

nandishjpatel (approved):

src/odemis/gui/win/acquisition.py
tepals: can you also extend the comment to indicate what the values 50e-6 and 1000e-6 are based on?

src/odemis/gui/win/acquisition.py
```diff
@@ -1004,6 +1011,9 @@ def update_setting_display(self):
             area_size_str = util.readable_str(area_size, unit="m", sig=3)
             self.area_size_txt.SetLabel(area_size_str)
 
+            # limit focus point dist based on area size
+            self.focus_points_dist.SetValueRange(50e-6, max(area_size))
```
K4rishma, line 1012: area size is only defined in if and else loop, when if not areas is True, set value range will not work as intended

K4rishma: Another point, area size seems to be the addition of all the list of given areas. Should not the range depend on one area at a time ?

patrickcleeve2: Currently nothing uses the multi-area acquisition so this is always just a single area.

K4rishma (approved):

src/odemis/gui/win/acquisition.py
K4rishma: Might be safer to explicitly use one area in the code, to not break the functionality in future?

src/odemis/gui/win/acquisition.py
```diff
@@ -785,6 +785,10 @@ def __init__(self, parent, orig_tab_data):
         self.btn_secom_acquire.Bind(wx.EVT_BUTTON, self.on_acquire)
         self.Bind(wx.EVT_CLOSE, self.on_close)
 
+        # range for the autofocus (sensible defaults)
```
tepals, line 788: `sensible defaults` still does not tell me that much. Can you update the comment to explain why `(50e-6, 1000e-6)` covers what is needed? You sort of explain it in the description of the PR:

> Previously this was hardcoded at 450um, which meant that for overviews smaller than that only a single autofocus in the centre of the image was done. This was particularly bad on systems with 100x mag

src/odemis/gui/win/acquisition.py
patrickcleeve2: fair point, i've changed it to take the max of each area individually

src/odemis/gui/win/acquisition.py
patrickcleeve2: Added additional detail about specific values (half a grid square -> half a grid) for range

tepals (approved):

PR #2864, opened 2024-08-28
[fix] implement function to gradually move descanner axes to their full range

tepals:
The previous implementation moved them very quickly, this caused problems with the hardware. The new movement is sinusoidal to ensure gradual and complete coverage of the range.

This commit includes the following:
* add a function to the technolution descanner class to move the axes along the full range
* use that function at the start of the acquisition
* update MirrorDescanner.getCalibrationSetpoints to support a sine function in x as well as in y
* clean up test code plots, used for debugging the setpoints; ensure values used for physical flyback time and dwell time are within range and update plots to use latest matplotlib colors

### DIFF
--- a/src/odemis/driver/technolution.py
+++ b/src/odemis/driver/technolution.py
@@ -429,12 +429,12 @@ class AcquisitionServer(model.HwComponent):
 
         total_line_scan_time = self._mppc.getTotalLineScanTime()
 
-        # get the scanner setpoints
-        x_descan_setpoints, y_descan_setpoints = self._mirror_descanner.getCalibrationSetpoints(total_line_scan_time)
-
         # get the descanner setpoints
+        x_descan_setpoints, y_descan_setpoints = descanner.getCalibrationSetpoints(total_line_scan_time)
+
+        # get the scanner setpoints
         x_scan_setpoints, y_scan_setpoints, scan_calibration_dwell_time_ticks = \
-            self._ebeam_scanner.getCalibrationSetpoints(total_line_scan_time)
+            scanner.getCalibrationSetpoints(total_line_scan_time)
 
         calibration_data = CalibrationLoopParameters(descanner.rotation.value,
                                                      0,  # Descan X offset parameter unused.
@@ -987,17 +987,25 @@ class MirrorDescanner(model.Emitter):
 
         return setpoints.tolist()
 
-    def getCalibrationSetpoints(self, total_line_scan_time):
+    def getCalibrationSetpoints(self, total_line_scan_time, sine_in_x=True, sine_in_y=False):
         """
         Calculate the setpoints for the descanner during calibration mode. The setpoints resemble a sine shape
-        in x and a flat line at zero in y.
+        on one axis and a flat line on the other or a sine in both directions.
 
         :param total_line_scan_time: (float) Total line scanning time in seconds including
-                                     overscanned pixels and flyback time. TODO do we need the flyback?
+                                     overscanned pixels and flyback time.
+        :param sine_in_x: (bool) If True, the setpoints in the x-direction will follow a sine wave.
+                                 if False, the setpoints in the x-direction will be a flat line.
+        :param sine_in_y: (bool) If True, the setpoints in the y-direction will follow a sine wave.
+                                 if False, the setpoints in the y-direction will be a flat line.
         :return:
                 x_setpoints (list of ints): The calibration setpoints in x direction in bits.
                 y_setpoints (list of ints): The calibration setpoints in y direction in bits.
         """
+        if not sine_in_x and not sine_in_y:
+            raise ValueError("Calibration set points require a sine wave in either the x or y axis, or both, "
+                             "not in neither direction.")
+
         # The calibration frequency is the inverse of the total line scan time.
         calibration_frequency = 1 / total_line_scan_time  # [1/sec]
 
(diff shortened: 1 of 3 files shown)


src/odemis/driver/technolution.py
```diff
@@ -987,17 +987,25 @@ def getYAcqSetpoints(self):
...
+                                     overscanned pixels and flyback time.
+        :param sine_in_x: (bool) If True, the setpoints in the x-direction will follow a sine wave.
+                                 if False, the setpoints in the x-direction will be a flat line.
+        :param sine_in_y: (bool) If True, the setpoints in the y-direction will follow a sine wave.
+                                 if False, the setpoints in the y-direction will be a flat line.
         :return:
                 x_setpoints (list of ints): The calibration setpoints in x direction in bits.
                 y_setpoints (list of ints): The calibration setpoints in y direction in bits.
         """
+        if not sine_in_x and not sine_in_y:
```
nandishjpatel, line 1005:
```suggestion
        if not sine_in_x or not sine_in_y:
```

tepals: It should only raise an error if both are False

src/odemis/driver/technolution.py
```diff
@@ -1362,7 +1445,8 @@ def getNextField(self, field_num):
         """
         Puts the command 'next' field image scan on the queue with the appropriate field meta data model of the field
         image to be scanned. Can only be executed if it preceded by a 'start' mega field scan command on the queue.
-        The acquisition thread returns the acquired image to the provided notifier function added in the acquisition queue
+        The acquisition thread returns the acquired image to the provided notifier function added in the acquisition
+        queue
```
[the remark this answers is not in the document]
tepals: not really, I think it was autoformatting, the line length is too long for that line. Should I undo it?

src/odemis/driver/technolution.py
```diff
@@ -1039,12 +1048,86 @@ def getCalibrationSetpoints(self, total_line_scan_time):
...
+                                                     0.0,  # Scan X offset parameter unused.
+                                                     x_scan_setpoints,
+                                                     0.0,  # Scan Y offset parameter unused.
+                                                     y_scan_setpoints)
+
+        # Check if an acquisition is still in progress
+        if not self.parent._mppc.acq_queue.empty():
+            logging.error("There is still an unfinished acquisition in progress. "
+                          "Calibration mode cannot be started yet.")
+            return
```
nandishjpatel, line 1108: Does one need to `Restore original parameters and clear calibration parameters` before return?

src/odemis/driver/technolution.py
nandishjpatel: Right, my bad

src/odemis/driver/technolution.py
tepals: Yes, but I think it actually makes more sense to move this check to the start of this function

src/odemis/driver/technolution.py
```diff
@@ -1039,12 +1048,86 @@ def getCalibrationSetpoints(self, total_line_scan_time):
...
+        time.sleep(total_line_scan_time * 10)
+        self.parent.asmApiPostCall("/scan/stop_calibration_loop", 204)
+
+        # Restore original parameters and clear calibration parameters
+        self.parent._calibrationParameters = None
+        self.parent._ebeam_scanner.dwellTime.value = init_dwellTime
+        self.physicalFlybackTime.value = init_physicalFlybackTime
+        self.scanOffset.value = init_scanOffset
+        self.scanAmplitude.value = init_scanAmplitude
+        self.parent._mppc.cellCompleteResolution.value = init_cellCompleteResolution
```
pieleric, line 1129: Typically, you want to place this in a `try/finally`, so that even if the code fails the initial values are restored.

src/odemis/driver/technolution.py
```diff
@@ -1039,12 +1048,86 @@ def getCalibrationSetpoints(self, total_line_scan_time):
...
+        if self.parent.calibrationMode.value:
+            # Sending this command without the calibration loop being active might cause errors.
+            self.parent.asmApiPostCall("/scan/stop_calibration_loop", 204)
+
+        # Start the calibration loop with the new parameters
+        self.parent._calibrationParameters = calibration_data
+        self.parent.asmApiPostCall("/scan/start_calibration_loop",
+                                   204,
+                                   data=self.parent._calibrationParameters.to_dict())
+        time.sleep(total_line_scan_time * 10)
```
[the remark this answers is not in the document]
pieleric: Do you mean that the X & Y axes will move 10 times back and forth? If it's worth indicating it in the docstring too. 
Also, could indicate how long, roughly, the procedure takes (in the docstring). Is that 1s? 1 min?

src/odemis/driver/test/technolution_test.py
pieleric, line 1: Please add a test case that calls `moveFullRange()`. There is probably not much to validate, but at least it shouldn't fail.  You could check that the VA values haven't changed, and that it took at least *some* time.

src/odemis/driver/technolution.py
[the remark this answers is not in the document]
tepals: Yes, they will move 10 times back and forth, I'll clarify in the docstring

nandishjpatel (approved):

pieleric (approved):

PR #2866, opened 2024-09-02
[feature] Add desktop entrypoint for mic selector

tmoerkerken:
The goal of this PR is to make Odemis easier to access on the Application Specialist's PC. Therefore I introduce the following changes:
- Add Odemis version selector
- Progress bar for validating microscope files
- Added a `.desktop` application file, so that there is an easy-to-access shortcut to the script

Screen to select version of Odemis
image

Progress bar (old version, text is different now)
image

Follow up screen (`/usr/share/odemis/sim/sparc2*` argument)
image

### DIFF
--- a/install/linux/usr/bin/odemis-select-mic-start
+++ b/install/linux/usr/bin/odemis-select-mic-start
@@ -10,23 +10,86 @@ Using this class microscope names are extracted and placed in a list.
 This list is presented to the user in the form of a (zenity) selection dialog window.
 After selection of the microscope configuration the right (yaml) file will be used to start Odemis.
 '
-# extract the arguments which were used to start this script
-args=( "$@" )
+# Default values for options
+select_version=false
+target_files=()
 
-for arg in "${args[@]}"; do
-    if test "$arg" == "--help"; then
-        # Display help on the argument usage
-        echo
-        echo "Syntax: odemis-select-mic-start [target_files]"
-        echo "arguments:"
-        echo "target_files      A selection of configuration files from a specific location."
-        echo "                  (e.g. /usr/share/odemis/sim/*meteor-main*.yaml)."
-        exit 0
-    fi
+# Function to display help information
+display_help() {
+    echo
+    echo "Syntax: odemis-select-mic-start [options] [target_files]"
+    echo "Options:"
+    echo "  --help                  Display this help message."
+    echo "  --select-version        Show prompt for selection of Odemis version."
+    echo "  target_files            A selection of configuration files from a specific location."
+    echo "                         (e.g. /usr/share/odemis/sim/*meteor-main*.yaml)."
+    exit 0
+}
+
+# Parse command-line arguments
+while [[ $# -gt 0 ]]; do
+    case "$1" in
+        --help)
+            display_help
+            exit 0
+            ;;
+        --select-version)
+            select_version=true
+            ;;
+        --*)
+            echo "Warning: Unkown option: $1, ignoring."
+            display_help
+            ;;
+        *)
+            if [ -z "$1" ]; then
+                echo "Error: No valid target files provided"
+                display_help
+                exit 1
+            fi
+            target_files+=("$1")
+            ;;
+    esac
+    shift
 done
 
 odemis-cli --check
 status=$?
+
+if [ "$select_version" = true ]; then
+    # Navigate to odemis repository
+    REPO_PATH="$HOME/development/odemis"
+    if [[ ! -d $REPO_PATH ]]; then
+        echo "ERROR: Odemis development folder cannot be found on the system ($REPO_PATH)."
+        exit 1
+    fi
+    cd "$REPO_PATH" || exit
+
+    # Prompt for selecting branch
+    user_choice=$(zenity --list --title="Select Odemis Version" \
+        --text="Please select which Odemis version to use:" \
+        --radiolist \
+        --column "Select" --column "Version" \
+        FALSE "Latest Development Changes" \
+        FALSE "Latest Release" \
+        TRUE "Current Branch" \
+        --width=400 --height=250)
+
+    if [ "$user_choice" == "Latest Development Changes" ]; then
+        git fetch origin
+        git checkout master
+        git pull origin master
+    elif [ "$user_choice" == "Latest Release" ]; then
+        latest_tag=$(git describe --tags `git rev-list --tags --max-count=1`)
+        git fetch origin
+        git checkout "$latest_tag"
+        git pull origin "$latest_tag"
+    elif [ "$user_choice" == "Current Branch" ]; then
+        :  # Leave current branch as is
+    else
+        zenity --error --text="No valid selection made. Exiting."
+        exit 1
+    fi
+fi
 
 # check if the backend is already running or starting up
 if [ $status == 0 ] || [ $status == 3 ] ; then
@@ -56,15 +119,27 @@ mkdir $HOME/.local/share/odemis/log 2>/dev/null
 parse_logfile=$HOME/.local/share/odemis/log/odemis-select-mic-start-parse.log
 > $parse_logfile
 
-for fn in "${args[@]}"; do
-    # add a new header to the logfile
-    printf "$(date) Parsing file: $fn\n" >> $parse_logfile
-    # extract the microscope name by using this small Python script
-    mic_name="$(python3 -c "import sys; import yaml; from odemis.odemisd import modelgen; \
-        f = open('$fn'); \
-        d = yaml.load(f, modelgen.SafeLoader); \
-        print(next(n for n, d in d.items() if d.get('class') == 'Microscope'))" 2>>$parse_logfile)"
-    mic_list+=("$fn")
+process_file() {
+    local fn="$1"
+
+    # Add a new header to the logfile
+    printf "$(date) Parsing file: $fn\n" >> "$parse_logfile"
+
+    # Extract the microscope name and role using a small Python script
+mic_name=$(python3 <<EOF 2>>"$parse_logfile"
+import sys
+import yaml
+from odemis.odemisd import modelgen
+
+f = open('$fn')
+d = yaml.load(f, modelgen.SafeLoader)
+mic_name = next(n for n, d in d.items() if d.get('class') == 'Microscope')
+print(mic_name)
+EOF
+    )
+
+    mic_list+=("$fn")  # The dialog has a hidden column, which is the filepath
+
     if test "$mic_name" == ""; then
         # save the file name in the error list
         error_filenames+=("$(basename ${fn})")
@@ -72,10 +147,38 @@ for fn in "${args[@]}"; do
         # add a newline to the logfile
         echo "" >> $parse_logfile
     fi
+
     mic_list+=("$mic_name")
+}
+
+total_files=${#target_files[@]}
+current_file=0
+
+# Processing a large selection of microscope files can take quite a while,
+# so giving visual feedback of the progress gives the user clarity
+# that the script is still running. Using file descriptor so we are
+# still able to append to mic_list.
+exec 3> >(zenity --progress \
+           --title="Progress" \
+           --text="Analysing microscope configurations..." \
+           --percentage=0 \
+           --auto-close \
+           --auto-kill)
+
+
+for fn in "${target_files[@]}"; do
+    process_file "$fn"
+    current_file=$((current_file + 1))
+    # Calculate the progress percentage
+    progress=$((current_file * 100 / total_files))
+    # Update the progress bar
+    echo "$progress" >&3
 done
 
-# prompt the user with a list of microscope configurations to choose from
+# Close the file descriptor
+exec 3>&-
+
+# Prompt the user with a list of microscope configurations to choose from
 selection="$(zenity --list --title="Odemis configuration starter" \
     --list --text "Please select a configuration to start:" \
     --column "" --column "" "${mic_list[@]}" --hide-header \
@@ -85,7 +188,7 @@ selection="$(zenity --list --title="Odemis configuration starter" \
 # if one of the microscope names is selected which also contains an error, show the log file and stop the script
 for efn in "${error_filenames[@]}"; do
     if test "$efn" == "$(basename ${selection})"; then
-        gedit $parse_logfile   
+        gedit $parse_logfile
         exit
     fi
 done
--- a/install/linux/usr/share/odemis/sim/odemis-select-mic.desktop
+++ b/install/linux/usr/share/odemis/sim/odemis-select-mic.desktop
@@ -1,0 +1,14 @@
+# Special Odemis icon with an alternative Executor giving the option to
+# select a microscope file before starting Odemis.
+# Install with:
+# sudo desktop-file-install /usr/share/odemis/sim/odemis-select-mic.desktop
+# It should place the file in /usr/share/applications/, and update the Ubuntu menu.
+[Desktop Entry]
+Name=Odemis Microscope Simulator
+Comment=Open odemis with a microscope selector
+Keywords=Odemis;simulator;microscope;select
+Exec=bash -c "odemis-select-mic-start /usr/share/odemis/sim/*.odm.yaml --select-version"
+Type=Application
+StartupNotify=true
+Icon=odemis
+Categories=System;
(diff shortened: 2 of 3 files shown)


tmoerkerken:
@sneepstefan I know you initiated this script, so please let me know if the addition of these features are not to your liking. I can always create a separate script.

install/linux/usr/bin/odemis-select-mic-start
```diff
@@ -56,40 +86,102 @@ mkdir $HOME/.local/share/odemis/log 2>/dev/null
...
-    --width=300 --height=400 \
-    --hide-column 1 --print-column=1 2>/dev/null)"
+    # Format the setup name
+    setup_name="${fn#*-}"
+    setup_name="${setup_name%.odm.yaml}"
+    mic_list+=("$setup_name")
+}
+
+# Check if the input is a directory or a wildcard
+if [ -d "${args[0]}" ]; then
```
pieleric, line 120: Showing an extra selection step based on whether the argument is a file or a directory is a little bit odd. I'd rather link this behaviour to a specific argument like `--select-role` or `--select-type`.

install/linux/usr/bin/odemis-select-mic-start
```diff
@@ -56,40 +86,102 @@ mkdir $HOME/.local/share/odemis/log 2>/dev/null
...
+    # Input is a directory, list YAML files and extract unique prefixes
+    yaml_files=("${args[0]}"/*.yaml)
```
pieleric, line 122: Let's not support passing a directory, to keep things simple. The user has to pass filenames, always.

install/linux/usr/bin/odemis-select-mic-start
```diff
@@ -56,40 +86,102 @@ mkdir $HOME/.local/share/odemis/log 2>/dev/null
...
+        base_name="$(basename "$fn")"
+        prefix="${base_name%%-*}"
+        if [[ ! " ${prefixes[@]} " =~ " ${prefix} " ]]; then
+            prefixes+=("$prefix")
+        fi
+    done
+
+    # Prompt user to select a prefix
+    selected_prefix=$(zenity --list --title="Odemis configuration starter" \
+        --text="Please select a system:" \
```
pieleric, line 141: Let's clarify with "Please select a system type".

install/linux/usr/bin/odemis-select-mic-start
```diff
@@ -56,40 +86,102 @@ mkdir $HOME/.local/share/odemis/log 2>/dev/null
...
+else
+    # Input is a wildcard or single file, process directly
+    for fn in "${args[@]}"; do
+        process_file "$fn"
+    done
+fi
+
+selection=$(zenity --list --title="Odemis configuration starter" \
+    --text "Please select a configuration to start:" \
+    --column="Configuration" --column="System" --column="Variant" \
```
pieleric, line 165: Please don't change the display of this window. There are several systems which uses it, and it doesn't make sense to have 3 columns. The idea is that the selection is very simple: just one name per row.

If the problem is that in the case of the simulators, they all have identical names, then we should just fix the names of the simulator microscope files. So far, this name wasn't useful, so we never cared. That's the only reason.

install/linux/usr/share/applications/odemis-select-mic.desktop
pieleric, line 1: All the .desktop files in this folder are automatically installed on all the systems. 
On the standard systems (dedicated to one microscope), it doesn't make sense to have a way to select all sorts of simulators.
So please move this desktop file to a different folder. You could put it in `/usr/share/odemis/sim/` or `/usr/share/doc/odemis/` (You'll have to update in `setup.py`  `data_files`).

install/linux/usr/share/applications/odemis-select-mic.desktop
```diff
@@ -0,0 +1,9 @@
+[Desktop Entry]
+Name=Odemis with Mic Selector
```
pieleric, line 2: Odemis Microscope Simulator

install/linux/usr/share/applications/odemis-select-mic.desktop
```diff
@@ -0,0 +1,9 @@
...
+Comment=Open odemis with a microscope selector
+Keywords=odemis.conf
```
pieleric, line 4: Odemis;simulator;microscope

install/linux/usr/share/applications/odemis-select-mic.desktop
```diff
@@ -0,0 +1,9 @@
...
+Exec=gnome-terminal -- bash -c 'odemis-select-mic-start /usr/share/odemis/sim/; exec bash'
```
pieleric, line 5: You can find example of usage in the `mic-odm-yaml` repo already. It's good to have an example in the main odemis source anyway!
I think it should be `Exec=bash -c "/usr/bin/odemis-select-mic-start --select-odemis-version --select-type /usr/share/odemis/sim/*.yaml"

pieleric (changes requested):

install/linux/usr/bin/odemis-select-mic-start
tmoerkerken: With name, do you mean the "SPARC2-Fiber" part in this snippet from  `sparc2-ext-spec-sim-cntpmt.odm.yaml`?
```
SPARC2-Fiber: {
    class: Microscope,
    role: sparc2,
    children: ["SEM E-beam", "SEM Detector","Calibration Light", "Integrated Spectrograph",
               "Optical Path Properties", "Spectrometer Selector", "External Spectrograph",
               "Camera", "Spectrometer", "Fiber align", "Spectral Camera",
               "Spec CCD Flipper", "Spectrograph focus", "Monochromator",
               "Mirror Actuators", "Mirror Actuators in XY referential", "Slit",
               "Lens1 Mover", "Lens2 Switch"],
}
```

install/linux/usr/share/applications/odemis-select-mic.desktop
tmoerkerken: I needed a terminal to spawn alongside it, since sometimes `odemis-stop` or `odemis-start` requires a sudo password. With just the bash command it didn't spawn for me.

install/linux/usr/bin/odemis-select-mic-start
```diff
@@ -20,14 +20,44 @@ for arg in "${args[@]}"; do
...
+        echo "                  (e.g. /usr/share/odemis/sim/*meteor-main*.yaml or a folder"
+        echo "                   containing configuration files, like /usr/share/odemis/sim/)."
         exit 0
     fi
 done
 
 odemis-cli --check
 status=$?
 
+# Navigate to odemis repository
```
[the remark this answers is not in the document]
tmoerkerken: Will update

install/linux/usr/bin/odemis-select-mic-start
```diff
@@ -20,14 +20,44 @@ for arg in "${args[@]}"; do
...
+source /etc/odemis.conf
+REPO_PATH="$DEVPATH/odemis"
+cd "$REPO_PATH" || exit
```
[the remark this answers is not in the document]
tmoerkerken: Agree, I will make it $HOME instead and add error handling, but also, I will only call it when the --select-version argument is passed.

install/linux/usr/bin/odemis-select-mic-start
```diff
@@ -56,40 +86,102 @@ mkdir $HOME/.local/share/odemis/log 2>/dev/null
...
 
-# prompt the user with a list of microscope configurations to choose from
-selection="$(zenity --list --title="Odemis configuration starter" \
-    --list --text "Please select a configuration to start:" \
-    --column "" --column "" "${mic_list[@]}" --hide-header \
-    --width=300 --height=400 \
-    --hide-column 1 --print-column=1 2>/dev/null)"
+    # Format the setup name
+    setup_name="${fn#*-}"
+    setup_name="${setup_name%.odm.yaml}"
```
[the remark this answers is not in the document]
tmoerkerken: I will keep it simpler and just use the original name. Scrolling through the list is not too bad for now.

install/linux/usr/bin/odemis-select-mic-start
tmoerkerken: I will leave it as is, and will ask Apps Specialists to come up with a better name (will be out of scope for now).

install/linux/usr/bin/odemis-select-mic-start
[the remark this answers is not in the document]
tmoerkerken: I will abandon the extra selection window for now, since it is going to be too much effort

install/linux/usr/bin/odemis-select-mic-start
tmoerkerken: I will abandon the extra selection window for now, since it is going to be too much effort

install/linux/usr/share/applications/odemis-select-mic.desktop
tmoerkerken: I moved it to the sim folder and added an entry to data_files. Not sure if it is the correct way though.

install/linux/usr/share/applications/odemis-select-mic.desktop
tmoerkerken: This it what I get
```
DEBUG:root:Running: sudo odemisd --daemonize --log-level 2 --log-target /var/log/odemis.log /usr/share/odemis/sim/meteor-zeiss-sim.odm.yaml
[sudo] password for moerkerken: 
```
From the code, the `sudo` call is defined in `odemisd/start.py`:
```
        odemisd_cmd = ["sudo", "odemisd", "--daemonize",
                 "--log-level", self._config["LOGLEVEL"],
                 "--log-target", self._config["LOGFILE"],
                 modelfile]
```

sneepstefan:
Hi Tim, I did not knew you sort of hijacked my mic selector script, so next time would you involve or notify me prior to changing it?

About the changes, I don't really see the benefit of most of the changes or is this a special version for the Apps PC?
The script I have written is mainly for METEOR systems and we don't want users to be able to select different versions of Odemis, we want the published version at all times. I also don't see the benefit of a separate progress window..

Honestly I think the tool is good 'as-is' right now until we need more functionality, but that's just my opinion.
If this modified version is only for the App PC then I would say it is fine.

tmoerkerken:
> Hi Tim, I did not knew you sort of hijacked my mic selector script, so next time would you involve or notify me prior to changing it?
> 
> About the changes, I don't really see the benefit of most of the changes or is this a special version for the Apps PC? The script I have written is mainly for METEOR systems and we don't want users to be able to select different versions of Odemis, we want the published version at all times. I also don't see the benefit of a separate progress window..
> 
> Honestly I think the tool is good 'as-is' right now until we need more functionality, but that's just my opinion. If this modified version is only for the App PC then I would say it is fine.

My bad, I tried to tag you in the comment above, but I guess it did not reach you. Will message you next time.

The changes are for the Apps PC and by default it should not change the original functionality (apart from the progress bar). The version selector is optional (default false), but necessary since some features of interest only live on the unreleased development branch.

Since Apps Specialists want to select between a large set of setups, I added the progress bar, since it takes some time to go through all the different microscope files. 

If you think this setup is tricky, I can copy the script and revert all changes to the original one. Let me know what you like best!

install/linux/usr/bin/odemis-select-mic-start
```diff
@@ -56,26 +119,66 @@ mkdir $HOME/.local/share/odemis/log 2>/dev/null
...
+total_files=${#target_files[@]}
+current_file=0
+
+# Processing a large selection of simulators can take quite a while,
+# so giving visual feedback of the progress gives the user clarity
+# that the script is still running. Using file descriptor so we are
+# still able to append to mic_list.
+exec 3> >(zenity --progress \
+           --title="Validation" \
+           --text="Validating Simulators..." \
```
pieleric, line 163: They are not always simulators. These are just microscope files. Also, it's "validating" sounds like it's checking that the user didn't do a mistake. Typically, the files are fine, and the main reason we are reading them is to find out their name. So I'd say "Analysing microscope configurations..."

install/linux/usr/share/applications/odemis-select-mic.desktop
pieleric: There is a config in `sudo` so that `sudo /usr/bin/odemisd` doesn't require a password (precisely for this use case).
I think it's a sign that on your computer something is not configured according to the standard. Most likely, there is another `odemisd`, maybe from a manual installation? (see `type odemisd`).

pieleric:
> If you think this setup is tricky, I can copy the script and revert all changes to the original one. Let me know what you like best!

I think it's fine to extend the script. There are a lot of things in common. It's important to keep the original use case in mind (see some of the desktop files in mic-odm-yaml), to make sure this doesn't affect the current users.
Now, the only change for these users should be the progress bar at the beginning. I *guess* it's fine, as it's probably very short, but if that's annoying we could only show that progress bar if the number of files is large (for instance, > 10).

@sneepstefan , the more you're going to implement new features, the more you'll get have cases of "someone has hijacked my code!". I speak from experience! It's mostly a matter of accepting it and making sure the original workflow still works.

install/linux/usr/bin/odemis-select-mic-start
```diff
@@ -10,24 +10,87 @@ Using this class microscope names are extracted and placed in a list.
...
+    cd "$REPO_PATH" || exit
+
+    # Prompt for selecting branch
+    user_choice=$(zenity --list --title="Select Odemis Version" \
+        --text="Please select which Odemis version to use:" \
+        --radiolist \
+        --column "Select" --column "Version" \
+        TRUE "Latest Development Changes" \
+        FALSE "Latest Release" \
+        FALSE "Current Branch" \
```
pieleric, line 74: I think "Current branch" (= do nothing) should be the default.

install/linux/usr/share/odemis/sim/odemis-select-mic.desktop
```diff
@@ -0,0 +1,9 @@
+[Desktop Entry]
```
pieleric, line 1: Please add a header to explain when to use this file, and how to install it.
```
# Special Odemis icon with an alternative Executor giving the option to 
# select a microscope file before starting Odemis.
# Install with:
# sudo desktop-file-install /usr/share/odemis/sim/odemis-select-mic.desktop
# It should place the file in /usr/share/applications/, and update the Ubuntu menu.
```

pieleric (changes requested):

sneepstefan:
> My bad, I tried to tag you in the comment above, but I guess it did not reach you. Will message you next time.
> 
I am not so fond of simply 'tagging' someone when the changes could be discussed beforehand.
> The changes are for the Apps PC and by default it should not change the original functionality (apart from the progress bar). The version selector is optional (default false), but necessary since some features of interest only live on the unreleased development branch.
> 
> Since Apps Specialists want to select between a large set of setups, I added the progress bar, since it takes some time to go through all the different microscope files.
> 
> If you think this setup is tricky, I can copy the script and revert all changes to the original one. Let me know what you like best!

I don't think it is tricky at all, indeed if the version selector is off at default that would be fine.
Regarding the progress bar I do actually like the suggestion of @pieleric showing that if the number of files to parse is quite high.

pieleric (approved):

tmoerkerken:
> > My bad, I tried to tag you in the comment above, but I guess it did not reach you. Will message you next time.
> 
> I am not so fond of simply 'tagging' someone when the changes could be discussed beforehand.
> 
> > The changes are for the Apps PC and by default it should not change the original functionality (apart from the progress bar). The version selector is optional (default false), but necessary since some features of interest only live on the unreleased development branch.
> > Since Apps Specialists want to select between a large set of setups, I added the progress bar, since it takes some time to go through all the different microscope files.
> > If you think this setup is tricky, I can copy the script and revert all changes to the original one. Let me know what you like best!
> 
> I don't think it is tricky at all, indeed if the version selector is off at default that would be fine. Regarding the progress bar I do actually like the suggestion of @pieleric showing that if the number of files to parse is quite high.

I played around with it a little bit, and even for a smaller selection it takes a few seconds, and I would personally prefer a progress bar. To me it is not annoying so I'd suggest to keep it as is.

sneepstefan (approved):

PR #2867, opened 2024-09-03
[METEOR-1014][FEAT] add maximum intensity projection display mode

patrickcleeve2:
Users have requested the ability to display max intensity projections inside odemis. This is used in particular for 2D correlation, especially with multi-channel images.

This PR adds
- Standalone maximum intensity projection helper function
- User interface toggle for displaying z-stacks as maximum intensity projections (Static2DStream -> StreamView -> RGBSpatialProjection)
- Test case for helper function


src/odemis/util/img.py
```diff
@@ -1273,3 +1273,19 @@ def apply_flood_fill(input_array, start):
...
+    :param data: (DataArray) 3D image data (assumed to be ZYX)
+    :param axis: (int) axis along which to project the image. Default is 0 (Z-axis)
+    :return: (DataArray) 2D image data (YX) with the maximum intensity projection
+    """
+    if data.ndim != 3:
+        raise ValueError(f"Data must be 3D, but got {data.ndim} dimensions.")
+    if axis not in (0, 1, 2):
+        raise ValueError(f"Invalid axis {axis}. Must be 0, 1, or 2.")
+
+    proj = numpy.max(data, axis=axis)
```
tmoerkerken, line 1288: I think we can make this function a bit more flexible, also to facilitate the TODO you added about different projection types:

```suggestion
def intensity_projection(data: model.DataArray, projection_method: Callable[[numpy.ndarray], numpy.ndarray] = numpy.max, axis: int = 0) -> model.DataArray:
    """Intensity projection of a 3D image along a given axis.
    :param data: (DataArray) 3D image data (assumed to be ZYX)
    :param projection_method: (Callable) function to apply to data. Default is numpy.max
    :param axis: (int) axis along which to project the image. Default is 0 (Z-axis)
    :return: (DataArray) 2D image data (YX) with the intensity projection
    """
    if data.ndim != 3:
        raise ValueError(f"Data must be 3D, but got {data.ndim} dimensions.")
    if axis not in (0, 1, 2):
        raise ValueError(f"Invalid axis {axis}. Must be 0, 1, or 2.")

    proj = projection_method(data, axis=axis)
```

tmoerkerken (changes requested):
Looks like a nice addition. I still have one suggestion to make it a bit more flexible.

src/odemis/acq/stream/_static.py
```diff
@@ -216,6 +216,11 @@ def __init__(self, name, raw, *args, **kwargs):
             self.zIndex = model.IntContinuous(0, (0, raw[0].shape[0] - 1))
             self.zIndex.subscribe(self._on_zIndex)
 
+            # define a projection
+            # TODO: add support for multiple types of projections (max, min, avg, sum, ...)
```
pieleric, line 220: "projection" is very generic. Here, this refers to showing multiple Z levels simultaneously. 
=> I would change the comment to:
```
Option to see all Z indices at the same time using maximum intensity projection (MIP)".
TODO: support other methods for multi-focus fusion?
```

src/odemis/gui/conf/data.py
```diff
@@ -1020,6 +1020,10 @@
                 "label": "Z Index",
                 "control_type": odemis.gui.CONTROL_SLIDER,
             }),
+            ("max_projection", {
+                "label": "Max Projection",
```
pieleric, line 1024: "Max *p*rojection"
"tooltip": "Merge all Z indices into a single image using Maximum Intensity Projection"

src/odemis/util/img.py
pieleric: Please, let's not over-engineer. For now we only need "max".  If one day we see we have many similar functions which only differ on the operator, we can generalize, but not now.
I doubt it's likely to happen, as there are other methods to merge multiple Z layers into a single image, but they don't differ just on the numpy operator. Typically, the other methods are a lot more complex.

src/odemis/util/img.py
```diff
@@ -1273,3 +1273,19 @@ def apply_flood_fill(input_array, start):
...
+    md = data.metadata.copy()
+    md[model.MD_DIMS] = "ZYX"
```
pieleric, line 1290: If the data is 2D the MD_DIMS should be "YX".

src/odemis/util/test/img_test.py
```diff
@@ -2120,5 +2120,28 @@ def test_boundary_fill(self):
...
+        expected_array = numpy.zeros((10, 10), dtype=int)
+        expected_array[0, 0] = 1
+        expected_array[2, 2] = 2
+        expected_array[4, 4] = 3
+        expected_array[6, 6] = 4
+        expected_array[8, 8] = 5
+        eda = model.DataArray(expected_array, {model.MD_DIMS: 'YX'})
+
+        projection = img.max_intensity_projection(da, axis=0)
+        numpy.testing.assert_array_equal(eda, projection)
```
pieleric, line 2144: Also check the MD_DIMS == "YX" (as eda)

pieleric: Can you add a check the the output dtype is the same as the input dtype?

src/odemis/util/img.py
patrickcleeve2: it's a good suggestion, but we might just shelf it until we actually need other projections. other projection methods aren't really used anywhere as widely as max intensity

tmoerkerken (approved):

pieleric (changes requested):
Looks good. Especially, this is a feature which has been requested quite a few times.
To make the GUI behaviour clearer, the "Z Index" control should be disabled when the "Max projection" is checked. => Please extend odemis.gui.cont.StreamController to detect that max_projection VA is present, and in this case, disable the Z Index entry (ie, the SettingEntry which has `.vigilattr is stream.zIndex`).

Also, please update directly the code of `stitching._tiledacq` to replace the call to `numpy.amax()` by the more explicitly `max_intensity_projection()`.

nandishjpatel (approved):

src/odemis/gui/cont/stream.py
```diff
@@ -198,6 +198,12 @@ def __init__(self, stream_bar, stream, tab_data_model, show_panel=True, view=Non
             self.stream.zIndex.subscribe(self._on_z_index)
             self.tab_data_model.zPos.subscribe(self._on_z_pos, init=True)
 
+        if hasattr(stream, "zIndex") and hasattr(stream, "max_projection"):
+            for se in self.entries:
+                if se.vigilattr == self.stream.zIndex:
+                    self._zindex_se = se
```
pieleric, line 204: Comparison to find an instance should be just a `is` (it's faster).
Also, it's good to handle the unexpected case where it fails to find the zIndex.
```suggestion
            for se in self.entries:
                if se.vigilattr is self.stream.zIndex:
                    self._zindex_se = se
                    break
            else:
                 logging.warning("Stream has zIndex but no corresponding stream entry found")
```

src/odemis/gui/cont/stream.py
```diff
@@ -523,6 +529,10 @@ def _on_z_pos(self, zPos):
 
         self.stream.zIndex.subscribe(self._on_z_index)
 
+    def _on_max_projection(self, val):
```
pieleric, line 532: In theory, the VA might be changed via code which is not running in the main GUI thread. So to be safe, you have to decorate the function with `@call_in_wx_main`.

Note that the functions just above don't (directly) change the GUI, so they don't need to be protected.

pieleric (changes requested):

pieleric (approved):

PR #2868, opened 2024-09-04
[fix] SPARC2 GUI fail to start with streakcam if only one spectrograph available

pieleric:
The code somehow ended up assuming that there are always 2
spectrographs. Of course, that's not correct, and it possible that the
streakcam is connected on the only spectrograph present.


pieleric:
@tmoerkerken , this fixes the issue you've found when starting the streakcam simulator.

sneepstefan (approved):

tmoerkerken (approved):

PR #2872, opened 2024-09-13
[DRAFT][METEOR-1186] [REFACTOR] Store feature stage position and focus position separately

patrickcleeve2:
Previously for meteor, the stage position and focus position were stored as a single 3D tuple (stage_x, stage_y, focus_z)
This behaviour was different on enzel, where the pos was (stage_x, stage_y, stage_z). 

This difference in behaviour was confusing, and the meteor position lacked information required to transform stage positions (stage_z). 

This refactor separates the stage position and focus position explicitly, and stores them as dictionaries to match the underlying hw components more closely (e.g. stage.position). We should also store a coordinate system for the stage to retain which stage's coordinate it is (stage-fm, stage-bare)

### DIFF
--- a/src/odemis/gui/cont/acquisition/cryo_z_localization.py
+++ b/src/odemis/gui/cont/acquisition/cryo_z_localization.py
@@ -198,7 +198,7 @@ class CryoZLocalizationController(object):
         feature = self._tab_data.main.currentFeature.value
         if feature is None:
             raise ValueError("Select a feature first to specify the Z localization in X/Y")
-        pos = self._tab_data.main.posture_manager.from_dependant_position(feature.stage_position.value)
+        pos = self._tab_data.main.posture_manager.to_sample_stage_from_stage_position(feature.stage_position.value)
 
         # Disable the GUI and show the progress bar
         self._tab.streambar_controller.pauseStreams()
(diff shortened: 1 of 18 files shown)


src/odemis/gui/cont/acquisition/cryo_z_localization.py
```diff
@@ -248,8 +248,7 @@ def _on_measure_z_done(self, f):
 
             # Update the feature Z pos, and move there
             feature = self._tab_data.main.currentFeature.value
-            pos = feature.pos.value[:2]
-            feature.pos.value = pos + (zpos,)
+            feature.focus_pos.value["z"] = zpos
```
pieleric, line 251: Such assignment skips the VA setter. So it skips the checks (but they are none currently), and more importantly doesn't call the subscribers. Typically, you need to always write like:
`feature.focus_pos.value = {"z": zpos}`

pieleric (changes requested):
It's a bit difficult to give feedback on these changes as it's not clear how they are going to be used. The main point is that CryoFeature's XYZ coordinates were supposed to be some sort of "idealized" coordinates (representing a point inside a volume), which 2 dicts is now hiding.

It would help to read a specification about the overarching goal of these changes to understand better what they are for.

patrickcleeve2:
closing until tested

PR #2874, opened 2024-09-16
[METEOR-1205][FIX] Add position to getTargetPosition rather than always using current position

patrickcleeve2:
The current get target position either used a pre-defined metadata position (e.g. grid_1) or the current stage position. This meant that we could not calculate the transformed position for arbitrary stage positions (e.g. feature positions).

This PR adds a position arg to the getTargetPosition to allow for calculating these transforms.

### DIFF
--- a/src/odemis/acq/move.py
+++ b/src/odemis/acq/move.py
@@ -27,7 +27,7 @@ import threading
 from abc import abstractmethod
 from concurrent.futures import CancelledError
 from concurrent.futures._base import CANCELLED, RUNNING, FINISHED
-from typing import Dict, Union
+from typing import Dict, Union, Optional
 
 import numpy
 import scipy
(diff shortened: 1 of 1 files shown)


src/odemis/acq/move.py
```diff
@@ -104,11 +104,13 @@ def getCurrentPostureLabel(self, pos: Dict[str, float] = None) -> int:
         """
         pass
 
-    def cryoSwitchSamplePosition(self, target: int):
+    def cryoSwitchSamplePosition(self, target: int, pos: Dict[str, float] = None) -> model.CancellableFuture:
```
nandishjpatel, line 107:
```suggestion
    def cryoSwitchSamplePosition(self, target: int, pos: Optional[Dict[str, float]] = None) -> model.CancellableFuture:
```

tmoerkerken (approved):

src/odemis/acq/move.py
nandishjpatel: Can you update the type for all `pos` args in this PR?

nandishjpatel (approved):

pieleric (commented):
This looks fine to me. However, as discussed about the larger work on supporting multiple posture, I think eventually there should be a generic function that should take, in addition to the position, also the current posture.

patrickcleeve2:
Closing, replaced by #2893

PR #2875, opened 2024-09-16
[METEOR-1204][FEAT] Add stage coordinate system conversion for Meteor

patrickcleeve2:
Currently there is no convenient way to transform a stage position from stage-bare to stage-fm and vice-versa. The current method required getting the linked YZ stage from the backend, and remapping the coordinates manually.

This PR adds some helper functions to the convert stage, and posture manager to simplify converting between these coordinate systems.

### DIFF
--- a/src/odemis/acq/move.py
+++ b/src/odemis/acq/move.py
@@ -264,6 +264,7 @@ class MeteorPostureManager(MicroscopePostureManager):
         # Load components
         self.stage = model.getComponent(role='stage-bare')
         self.focus = model.getComponent(role='focus')
+        self.convert_stage = model.getComponent(name="Linked YZ")
         # set linear axes and rotational axes used
         self.axes = self.stage.axes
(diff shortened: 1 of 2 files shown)


pieleric (changes requested):
This breaks several "base conventions" of the backend: 
* the client (eg, GUI) should only access the components via their roles.
* the client shouldn't rely on a specific class to implement a component. It should rely only on standard methods provided by the type of component (Actuator, Detector, Emitter) 

Here, the code expects the component to be called "Linked YZ". Instead, it should rely on a role. Of course, I understand it's tricky, because for now this component has no role, as it was originally *on purpose* hidden, as a supporting component for the role="stage". In addition, it relies on new method `transform_stage_positions_*()`, which are only present on `ConvertStage`. 
I think we should take a step back and instead redefine how the pre-tilt angle is stored in the microscope file, and how the PostureManager is handling it.

patrickcleeve2:
closing, refactoring core code

PR #2878, opened 2024-09-20
[TRCL-436] [fix] Better behaviour of keysight beam blanker when blanking is disabled

pieleric:
Add an option so that when the power is "off", the output is not unterminated, but at a specific (constant) voltage.
Change the behaviour so that when the power is "on", it's really not possible to modify the setting on the front panel, but when it's off, the user is free to change as wished.

### DIFF
--- a/src/odemis/driver/keysight.py
+++ b/src/odemis/driver/keysight.py
@@ -23,16 +23,15 @@ Odemis. If not, see http://www.gnu.org/licenses/.
 import logging
 import math
 import re
+import socket
 import threading
 import time
-import socket
+from typing import List, Tuple, Dict, Optional
 
 from odemis import model
 from odemis.model import HwError
-from odemis.util import to_str_escape, RepeatingTimer
-from typing import List, Tuple, Dict, Optional
+from odemis.util import to_str_escape
 
-POLL_INTERVAL = 5  # s, how often to read the settings again
 
 FREQUENCY_MIN = 1e-6  # Hz. Based on user manual
 FREQUENCY_MAX = 100e6  # Hz. Based on user manual
@@ -59,6 +58,14 @@ class TrueFormError(OSError):
 
 
 class TrueForm(model.Emitter):
+    """
+    Control the Keysight 33600 series Arbitrary Waveform Generator to generate square waveforms
+    with the desired period, duty cycle, and delay.
+    When the power is "on" the waveform is active, and the device panel locked. When the power is
+    "off", the output is disabled, or a constant voltage is applied (according to the off_voltage).
+    In this mode, the device panel is unlocked, and the user can change the settings manually.
+    Setting the power back to "on" sets back the device to the last settings defined by the VAs.
+    """
     def __init__(self, name, role, address,
                  channel: int,
                  limits: List[Tuple[float, float]],
@@ -149,21 +156,11 @@ class TrueForm(model.Emitter):
         self._set_period(period)
         self._set_delay(delay)
 
-        # The device has a front panel which the user can use to manual change the settings.
-        # Normally, it'll show a warning that the "instrument in remote [control]", but it's
-        # still possible to change them. In case this happens, don't ignore them, and just update
-        # the VAs. The drawback of the regular polling is that it automatically sets back the device
-        # into "remote" mode, so it's really hard to change the settings manually.
-        self._poll_timer = RepeatingTimer(POLL_INTERVAL, self._update_settings, "Keysight settings update")
-        self._poll_timer.start()
-
     def terminate(self):
         if self._accesser:
             self._set_power(False)
             self._accesser.close()
             self._accesser = None
-
-        self._poll_timer.cancel()
 
         super().terminate()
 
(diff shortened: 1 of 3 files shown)


src/odemis/driver/keysight.py
```diff
@@ -85,37 +98,38 @@ def __init__(self, name, role, address,
...
+                raise ValueError(f"Tracking channel {c} not defined in limits")
             if t not in {"ON", "INV"}:
                 raise ValueError(f"Incorrect tracking mode: {t} given")
-            other_c = 3 - c
-            self.setTracking(other_c, t)
 
-        power = False
-        self.power = model.BooleanVA(power, setter=self._set_power)
-        self._set_power(power)
+        # Pretend the device is on, and that we now power if off, so that the settings are applied
```
tmoerkerken, line 130:
```suggestion
        # Pretend the device is on, and that we now power it off, so that the settings are applied
```

src/odemis/driver/keysight.py
```diff
@@ -420,44 +461,58 @@ def _set_duty_cycle(self, dc: float) -> float:
...
+                self.setVoltageMax(c + 1, lim[1])
+
+            # Reset all the settings, for 3 reasons:
+            # * if the VAs were changed while the power was off, they couldn't be applied, so we
+            # need to set them now
+            # * maybe the user has changed them in the meantime. So we need to put back the expected
+            # values.
+            # * it actually might be the first time the device is powered on, so we need to set them
+            # all.
+            self.dutyCycle.value = self.dutyCycle.value
```
tmoerkerken, line 488: This is very cryptic and the setter is also very vague to me. If the power is off, we still set the value. Is this value then different than what is on the machine? If the power is on, some data alteration takes place (multiplying by 1e2), which is also under-explained in my opinion.

It would require less comments if we bookkeep a shadow value (i.e. self.dutyCycleTarget) and keep the dutyCycle VA always in sync with the device.

tmoerkerken:
I think `test_duty_cycle` can be extended to test for different power states, since that code path is not trivial.

src/odemis/driver/test/keysight_test.py
tmoerkerken, line 1: I think `test_duty_cycle` can be extended to test for different power states, since that code path is not trivial.

tmoerkerken (changes requested):

sneepstefan:
Could you add a description to the PR describing why the Keysight bb needs better behavior and what behavior is expected? How does this PR contribute to better behavior?

src/odemis/driver/test/keysight_test.py
sneepstefan, line 1: Could you remove line 26 (util import) since it is not used?

sneepstefan (changes requested):

src/odemis/driver/keysight.py
pieleric: You're right, at this point, when changing the power from off -> on, the VA value is still "False". So calling the setter of dutyCycle will have no effect.

pieleric:
> Could you add a description to the PR describing why the Keysight bb needs better behavior and what behavior is expected? How does this PR contribute to better behavior?

Sorry about this. Each commit change is described separately, but I forgot to summarize them in the PR description. This is fixed.

sneepstefan (approved):

src/odemis/driver/keysight.py
```diff
@@ -620,9 +704,31 @@ def _parseMessage(self, msg):
         elif re.match(r"sour[1-2]:freq?", msg):
             channel = int(msg[4])
             self._sendAnswer(b"%+.15E" % (self._output_states[channel].frequency,))
-        elif re.match(r"sour[1-2]:appl:(sin|squ|ramp|puls)", msg):
+        elif re.match(r"sour[1-2]:appl:(sin|squ|ramp|puls|dc)", msg):
```
tmoerkerken, line 707: you can utilize the regex with group names to do the variable assignment for you.
Something in the lines of:
```
elif match := re.match(r"sour(?P<channel>[1-2]):appl:(?P<waveform>sin|squ|ramp|puls|dc)(?: (?P<params>.+))?", msg):
    channel = int(match.group("channel"))
    waveform = match.group("waveform")
    params = match.group("params").split(",") if match.group("params") else []
```

src/odemis/driver/test/keysight_test.py
```diff
@@ -80,6 +80,20 @@ def test_duty_cycle(self):
...
             self.dev.dutyCycle.value = 0.15
 
+        # Change the duty cycle while the power is off: it shouldn't have effect immediately, but
+        # it should be applied when the power is turned on.
+        prev_dev_duty_cycle = self.dev.getDutyCycle(1)  # low-level value (not the same unit)
+        self.dev.power.value = False
+        self.dev.dutyCycle.value = 0.5
+        self.assertEqual(self.dev.dutyCycle.value, 0.5)
+        dev_duty_cycle = self.dev.getDutyCycle(1)
+        self.assertEqual(prev_dev_duty_cycle, dev_duty_cycle)
```
tmoerkerken, line 90: I get how this works, but from variable names alone, this behavior is not trivial.

tmoerkerken (approved):

src/odemis/driver/keysight.py
pieleric: I'll not implement this for now. I just want to keep this change simple. Adding "dc" is enough.

src/odemis/driver/test/keysight_test.py
pieleric: Variables names are a good way to express behaviour, better than comments :-)

PR #2880, opened 2024-09-20
[fix] andorcam2: improve shutter handling on iXon Ultra

pieleric:
The iXon Ultra has a very perculiar way to use the "shutter".
Based on clarification from Andor, the shutter should just be left open
all the time, until we stop the program. So just consider it doesn't
have a shutter.

### DIFF
--- a/src/odemis/driver/andorcam2.py
+++ b/src/odemis/driver/andorcam2.py
@@ -822,27 +822,22 @@ class AndorCam2(model.DigitalCamera):
                                                     unit="m",
                                                     setter=self._setCountConvertWavelength)
 
-            # To control the shutter: select the maximum frequency, aka minimum
-            # period for the shutter. If it the acquisition time is below, the
-            # shutter stays open all the time. So:
-            # 0 => shutter always auto
-            # > 0 => shutter auto if exp time + readout > period, otherwise opened
-            # big value => shutter always opened
-            if self.hasShutter(shutter_times is not None):
+            if caps.CameraType == AndorCapabilities.CAMERATYPE_IXONULTRA:
+                # Special case for iXon Ultra -> leave it open (with 0, 0) (cf Section 5.6 "Shuttering" in hardware guide)
+                logging.debug("iXon Ultra camera, leaving shutter always open")
+                self.SetShutter(1, AndorV2DLL.SHUTTER_OPEN, 0, 0)
+                self._shutter_period = None
+            elif self.hasShutter(shutter_times is not None):
                 if shutter_times is None:
                     shutter_times = (0, 0)
                 elif not all(0 <= s < 10 for s in shutter_times):
                     raise ValueError("shutter_times must be between 0 and 10s")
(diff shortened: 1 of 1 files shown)


src/odemis/driver/andorcam2.py
```diff
@@ -828,21 +828,18 @@ def __init__(self, name, role, device, emgains=None, shutter_times=None,
             # 0 => shutter always auto
             # > 0 => shutter auto if exp time + readout > period, otherwise opened
             # big value => shutter always opened
-            if self.hasShutter(shutter_times is not None):
+            if caps.CameraType == AndorCapabilities.CAMERATYPE_IXONULTRA:
+                # Special case for iXon Ultra -> leave it open (with 0, 0) (cf p.77 hardware guide)
```
tepals, line 832: `(cf p.77 hardware guide)` can you mention a specific section? That will be helpful in case we get a new hardware guide

src/odemis/driver/andorcam2.py
```diff
@@ -2629,14 +2626,21 @@ def _acquire_images(self):
...
 
             # Force the shutter to be closed (in case it was forced open during acquisition)
+            sleep_time = 0
             if shutter_forced:
-                self.SetShutter(1, AndorV2DLL.SHUTTER_CLOSE, 0, 0)
+                try:
+                    self.SetShutter(1, AndorV2DLL.SHUTTER_CLOSE, 0, 0)
+                except Exception as ex:
+                    logging.warning("Failed to close the shutter: %s", ex)
                 # Wait enough time to be sure it's closed, before opening it again on next acquisition
```
tepals, line 2635: I think this comment can be moved to line 2640

src/odemis/driver/andorcam2.py
```diff
@@ -828,21 +828,18 @@ def __init__(self, name, role, device, emgains=None, shutter_times=None,
             # 0 => shutter always auto
             # > 0 => shutter auto if exp time + readout > period, otherwise opened
             # big value => shutter always opened
```
tepals, line 830: This comment doesn't really make sense to me, is it meant something like this:
```suggestion
            # To control the shutter: select the maximum frequency, aka minimum
            # period for the shutter. If it the acquisition time is below the minimum period, the
            # shutter stays open all the time. So:
            # minimum period = 0 => shutter always auto
            # minimum period > 0 => shutter auto if exp time + readout > period, otherwise opened
            # big value => shutter always opened
```

tepals (changes requested):

sneepstefan (approved):

src/odemis/driver/andorcam2.py
pieleric: Indeed, that was quite unclear. I've reworded the text. Especially, I dropped the examples with "= 0, >0, big value", because it is actually not how it is used in reality, so it doesn't help much to understand this setting.

tepals (approved):

PR #2887, opened 2024-10-01
[METEOR-1209][FIX] Invalid cached pyramids, broken correlation

patrickcleeve2:
User's reported that they were not able to correlate fm overview images, and that the images would not move. 

The data projection for overview images uses a cache to efficiently read/display tiled images. However, the projection assumed the image would never 'move', and there was no direct way to invalidate the cache to recompute the image. 
This PR adds a public method to force the projection to clear the cache and recompute the image. This allows user's to move the overviews (shadow-pyramids) correctly in the correlation tab.

- [Jira Issue](https://delmic.atlassian.net/browse/METEOR-1209)
- [Related](https://delmic.atlassian.net/browse/METEOR-1210)

### DIFF
--- a/src/odemis/acq/stream/_base.py
+++ b/src/odemis/acq/stream/_base.py
@@ -1100,8 +1100,8 @@ class Stream(object):
         return (DataArray): The merged image
         """
         # calculates the size of the merged image
-        width_zoomed = das.shape[1] / (2 ** z)
-        height_zoomed = das.shape[0] / (2 ** z)
+        width_zoomed = int(das.shape[1] / (2 ** z))
+        height_zoomed = int(das.shape[0] / (2 ** z))
         # calculates the number of tiles on both axes
         num_tiles_x = int(math.ceil(width_zoomed / das.tile_shape[1]))
         num_tiles_y = int(math.ceil(height_zoomed / das.tile_shape[0]))
@@ -1352,3 +1352,7 @@ class Stream(object):
         A list of metadata dicts is returned.
         """
         return [None if data is None else data.metadata for data in self.raw]
+
+    def force_image_update(self):
+        """Force the stream image to be updated"""
+        self._shouldUpdateImage()
--- a/src/odemis/acq/stream/_projection.py
+++ b/src/odemis/acq/stream/_projection.py
@@ -221,6 +221,17 @@ class RGBProjection(DataProjection):
         except Exception:
             logging.exception("Updating %s %s image", self.__class__.__name__, self.stream.name.value)
 
+    def force_image_update(self):
+        """Forces the image to be recomputed entirely, and invalidates the cache for tiled images"""
+
+        if hasattr(self, "_projectedTilesCache") and hasattr(self, "_rawTilesCache"):
+            # invalidate the raw and projectted tiles cache
+            self._projectedTilesCache = {}
+            self._rawTilesCache = {}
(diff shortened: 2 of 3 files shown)


src/odemis/acq/stream/_base.py
```diff
@@ -1358,3 +1359,9 @@ def getRawMetadata(self):
         A list of metadata dicts is returned.
         """
         return [None if data is None else data.metadata for data in self.raw]
+
+    def invalidate_cache(self):
```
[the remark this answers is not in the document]
patrickcleeve2: I've changed it to be the client responsibility, and renamed it to be 'force_image_update', because not all of the streams/projections have caches to invalidate. It works, but i need to update .rect for the tiledshadows so that fit_to_content works correctly, but this is proving more difficult because changing it changes the tile-read calculation, and it doesn't let me set it outside the current range (which is staticly calculated on load). Alternatively, it might be better to change how getBoundingBox works for RGBSpatialProjection, and just use the image content instead of the .rect.range?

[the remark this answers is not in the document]
patrickcleeve2: fixed the bounding box issue, now it's recalculated rather than fixed at .rect

src/odemis/acq/stream/_projection.py
```diff
@@ -934,13 +945,11 @@ def _on_max_projection(self, value):
     def getBoundingBox(self):
         '''
         Get the bounding box of the whole image, whether it`s tiled or not.
+        Since the image can now be updated, the bounding box can change
+        so we need to get it from the stream directly.
```
pieleric, line 949: As mentioned in PR #2888, I think that .rect.range should also be updated when `force_image_update()` is called. Then, this code can probably stay as-is.

src/odemis/acq/stream/_projection.py
```diff
@@ -934,13 +945,11 @@ def _on_max_projection(self, value):
...
         return (tuple of floats(minx, miny, maxx, maxy)): Tuple with the bounding box
         '''
-        if hasattr(self, 'rect'):
-            rng = self.rect.range
-            return rng[0][0], rng[0][1], rng[1][0], rng[1][1]
-        else:
-            return self.stream.getBoundingBox(self.image.value)
+        return self.stream.getBoundingBox()
```
pieleric, line 952: BTW, for now getBoundingBox doesn't take care of the rotation. However, for a *whole* image, that usually doesn't have huge effect, as rotation is centered on the image center. That means in the worst case, at 45° rotation, the bounding box is a little too small (but correctly centered).

src/odemis/acq/stream/_base.py
```diff
@@ -1100,8 +1100,8 @@ def _getMergedRawImage(self, das, z):
         return (DataArray): The merged image
         """
         # calculates the size of the merged image
-        width_zoomed = das.shape[1] / (2 ** z)
-        height_zoomed = das.shape[0] / (2 ** z)
+        width_zoomed = int(das.shape[1] / (2 ** z))
+        height_zoomed = int(das.shape[0] / (2 ** z))
```
pieleric, line 1104: This change should be dropped from this PR (it's handled in #2888).

pieleric (changes requested):

patrickcleeve2:
closed. content migrated to #2888

PR #2888, opened 2024-10-03
[METEOR-1210][FIX] Fix image pyramid tile read calculation and projection cache invalidation 

patrickcleeve2:
User's reported that they were not able to correlate fm overview images, and that the images would not move.  They also reported that some overviews wouldn't load. These two issues were interrelated:

1. [METEOR-1210](https://delmic.atlassian.net/browse/METEOR-1210) Tiled imaged reads were failing for specific resolution / tile size combinations due to inconsistencies and rounding issues in the tile calculation. Made all tile reads use the same code, and fixed the rounding issues. 

2. [ METEOR-1209](https://delmic.atlassian.net/browse/METEOR-1209) The data projection for overview images uses a cache to efficiently read/display tiled images. However, the projection assumed the image would never 'move', and there was no direct way to invalidate the cache to recompute the image. 
This PR adds a public method to force the projection to clear the cache and recompute the image. This allows user's to move the overviews (shadow-pyramids) correctly in the correlation tab. It also correctly now recalculates the image rect when moving the image.

Now when moving an image, the caller is required to call force_image_update which will:
1. Invalidate any cached tiles (projected and raw)
2. Recompute the image rect (view)
3. Request image re-computation

[METEOR-1210]: https://delmic.atlassian.net/browse/METEOR-1210?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/odemis/acq/stream/_projection.py
+++ b/src/odemis/acq/stream/_projection.py
@@ -28,6 +28,7 @@ import math
 import gc
 import numpy
 
+from typing import Tuple, Dict
 from odemis.acq.stream import POL_POSITIONS
 from odemis.model import TINT_FIT_TO_RGB
 
@@ -221,6 +222,15 @@ class RGBProjection(DataProjection):
         except Exception:
             logging.exception("Updating %s %s image", self.__class__.__name__, self.stream.name.value)
 
+    def force_image_update(self):
+        """Forces the image to be recomputed entirely, and invalidates the cache for tiled images"""
+
+        if hasattr(self, "_projectedTilesCache") and hasattr(self, "_rawTilesCache"):
+            # invalidate the raw and projected tiles cache
+            self._projectedTilesCache = {}
+            self._rawTilesCache = {}
+            self._update_rect()  # re-compute the image rect
+        self._shouldUpdateImage()
 
 class ARProjection(RGBProjection):
     """
@@ -849,6 +859,7 @@ class RGBSpatialProjection(RGBProjection):
             minx, miny, maxx, maxy = full_rect
             rect_range = ((minx, miny, minx, miny), (maxx, maxy, maxx, maxy))
             self.rect = model.TupleContinuous(full_rect, rect_range)
+            self.rect.clip_on_range = True
             self.mpp.subscribe(self._onMpp)
             self.rect.subscribe(self._onRect)
             # initialize the projected tiles cache
@@ -859,6 +870,15 @@ class RGBSpatialProjection(RGBProjection):
             self._projectedTilesInvalid = True
 
         self._shouldUpdateImage()
+
+    def _update_rect(self):
+        """Update the rect VA based on the full image bounding box"""
+        full_rect = self.stream.getBoundingBox()
+        minx, miny, maxx, maxy = full_rect
+        self.rect.unsubscribe(self._onRect)
+        self.rect.range = ((minx, miny, minx, miny), (maxx, maxy, maxx, maxy))
+        self.rect.value = full_rect
+        self.rect.subscribe(self._onRect)
 
     def _onMpp(self, mpp):
         self._shouldUpdateImage()
@@ -958,7 +978,8 @@ class RGBSpatialProjection(RGBProjection):
         return (tuple containing x1, y1, x2, y2): Rect on pixel coordinates where x1 < x2 and y1 < y2
         """
         das = self.stream.raw[0]
-        md = das.metadata
+        md = das.metadata.copy()
+        img.mergeMetadata(md)  # apply the corrections
         ps = md.get(model.MD_PIXEL_SIZE, (1e-6, 1e-6))
         pos = md.get(model.MD_POS, (0, 0))
         # Removes the center coordinates of the image. After that, rect will be centered on 0, 0
@@ -985,7 +1006,14 @@ class RGBSpatialProjection(RGBProjection):
             int(round(rect[1] / (-ps[1]) + img_shape[1] / 2)) - 1,
         )
 
-    def _getTile(self, x, y, z, prev_raw_cache, prev_proj_cache):
+    def _getTile(
+        self,
+        x: int,
+        y: int,
+        z: int,
+        prev_raw_cache: Dict[str, model.DataArray],
+        prev_proj_cache: Dict[str, model.DataArray],
+    ) -> Tuple[model.DataArray, model.DataArray]:
         """
         Get a tile from a DataArrayShadow. Uses cache.
         The cache for projected tiles and the cache for raw tiles has always the same tiles
@@ -1083,11 +1111,19 @@ class RGBSpatialProjection(RGBProjection):
         need_recompute = True
         while need_recompute:
             z = self._zFromMpp()
-            rect = self._rectWorldToPixel(self.rect.value)
-            # convert the rect coords to tile indexes
-            rect = [l / (2 ** z) for l in rect]
-            rect = [int(math.floor(l / das.tile_shape[0])) for l in rect]
-            x1, y1, x2, y2 = rect
+
+            # calculate the image pixels inside the view (rect)
+            view_rect = self._rectWorldToPixel(self.rect.value)
+
+            # convert the view rect coords to tile indexes
+            zoom_rect = img.apply_zoom_on_image_coordinates(view_rect, z)
+            # get the tile indexes
+            x1, y1, x2, y2 = img.get_tile_indices(zoom_rect, das.tile_shape)
+
+            # for debugging (cache value to check it changes)
+            rect = self.rect.value
+            rect_x, rect_y = rect[2] - rect[0], rect[3] - rect[1]
+
             # the 4 lines below avoids that lots of old tiles
             # stays in instance caches
             prev_raw_cache.update(self._rawTilesCache)
@@ -1130,6 +1166,9 @@ class RGBSpatialProjection(RGBProjection):
             except NeedRecomputeException:
                 # image changed
                 need_recompute = True
+
+        logging.debug(f"{self.stream.name.value} (rect size: {rect_x*1e3:.4f} x {rect_y*1e3:.4f} mm)")
+        logging.debug(f"read {x2-x1} x {y2-y1} tiles at zoom: {z}, mpp: {self.mpp.value} m/px")
 
         return tuple(raw_tiles), tuple(projected_tiles)
 
--- a/src/odemis/dataio/test/tiff_test.py
+++ b/src/odemis/dataio/test/tiff_test.py
@@ -21,25 +21,28 @@ You should have received a copy of the GNU General Public License along with
 Odemis. If not, see http://www.gnu.org/licenses/.
 '''
 # Don't import unicode_literals to avoid issues with external functions. Code works on python2 and python3.
-from PIL import Image
-import libtiff
+import json
 import logging
-import numpy
-from odemis import model
-import odemis
-from odemis.acq.stream import POL_POSITIONS, POL_POSITIONS_RESULTS
-from odemis.dataio import tiff
-from odemis.util import img
 import os
 import re
 import time
 import unittest
-from unittest.case import skip
-import json
-from datetime import datetime
-
-import libtiff.libtiff_ctypes as T # for the constant names
 import xml.etree.ElementTree as ET
+from datetime import datetime
+from unittest.case import skip
+
+import libtiff
+import libtiff.libtiff_ctypes as T  # for the constant names
+import numpy
+import odemis
+from odemis import model
+from odemis.acq.stream import POL_POSITIONS, POL_POSITIONS_RESULTS
+from odemis.acq.stream._projection import RGBSpatialProjection
+from odemis.dataio import tiff
+from odemis.dataio.tiff import export
+from odemis.util import img
+from odemis.util.dataio import data_to_static_streams, open_acquisition
+from PIL import Image
 
 logging.getLogger().setLevel(logging.DEBUG)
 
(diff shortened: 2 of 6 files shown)


src/odemis/acq/stream/_projection.py
```diff
@@ -1083,11 +1091,21 @@ class NeedRecomputeException(Exception):
         need_recompute = True
         while need_recompute:
             z = self._zFromMpp()
-            rect = self._rectWorldToPixel(self.rect.value)
+
+            # .rect was static, cant be updated when the image moves, we now use bounding box instead
+            # TODO: this works, but loads all tiles instead of ones in the view, might be good place to leave it
+            # TODO: get the bbox inside the .rect
+            view_rect = self._rectWorldToPixel(self.getBoundingBox())
```
pieleric, line 1098: This changes quite a lot the behaviour of this function. Instead of returning just the tiles needed, it returns all the tiles. At z==0, that means loading the entire data. So the point of pyramidal/tiled image is lost.

`.rect` is not static. It can be updated live. Both its value *and* its .range can be updated. See `.clip_on_range = True` on the VA to simplify the update. This way, you'd adjust it whenever `force_image_update()` is called.
I guess that when this happens, you could reset its value to the whole range (as it's initialized). Then, it's up to the caller to recompute the new value. (alternatively, you could try to use the value and range prior to the update, and adjust the value linearly to the new range... but that might make things more complicated for not help).

Possibly, the other part that this change "solved" is when the image has (large) rotation. So far, `_rectWorldToPixel()` assumes there is no rotation. To handle rotation, you'd need to adjust this function to properly compute the bounding box. I think this can be done easily by just using the `Stream.getPixelCoordinates()` function on the four corner coordinates. (that function didn't exist when we wrote `_rectWorldToPixel()`, hence it's not currently using it).

pieleric (changes requested):

src/odemis/util/img.py
```diff
@@ -1289,3 +1288,81 @@ def max_intensity_projection(data: model.DataArray, axis: int = 0) -> model.Data
...
+    :param rect: the pixel coordinates to fetch the image
+    :param tile_shape: the shape of the tiles
+    :return: the range of indices (start_x, start_y, end_x, end_y) of the tiles"""
+    
+    xmin, ymin, xmax, ymax = rect
+    tile_height, tile_width = tile_shape
+    start_tile_x = math.floor(xmin / tile_width)
+    start_tile_y = math.floor(ymin / tile_height)
+    end_tile_x = math.ceil(xmax / tile_width) - 1
+    end_tile_y = math.ceil(ymax / tile_height) - 1
```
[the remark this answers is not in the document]
patrickcleeve2: ceil/floor return ints not floats? https://docs.python.org/3/library/math.html#math.ceil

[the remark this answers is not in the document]
pieleric: Oh... sorry, indeed! On Python 2, it used to returned floats... and I never noticed this was fixed! Good to know :-)

src/odemis/dataio/test/tiff_test.py
```diff
@@ -2729,6 +2732,65 @@ def test_convert_openfibsem_to_odemis_metadata(self):
...
+        )
+
+        # export the dummy overview pyramid
+        export(PYRAMID_FILENAME, das, pyramid=True)
+
+        # open the pyramid and create a projection
+        data = open_acquisition(PYRAMID_FILENAME)
+        stream = data_to_static_streams(data)[0]
+        proj = RGBSpatialProjection(stream)
+        proj._getTilesFromSelectedArea()
```
pieleric, line 2771: What are you testing here? Is it just that _getTilesFromSelectedArea() doesn't fail? It's odd as it's a private function. If so, you should at least add a comment, and otherwise, try to trigger that call via an interaction using the public API.

Also, if this is to test the projection, this should go to src/odemis/acq/test/stream_test.py

src/odemis/dataio/test/tiff_test.py
```diff
@@ -2729,6 +2732,65 @@ def test_convert_openfibsem_to_odemis_metadata(self):
...
+                zoom_rect = img.apply_zoom_on_image_coordinates(image_rect, z)
+                tidx = img.get_tile_indices(zoom_rect, das.tile_shape)
+
+                xmin, ymin, xmax, ymax = tidx
+
+                # assert reading tiles doesn't raise an error
+                for x in range(xmin, xmax):
+                    for y in range(ymin, ymax):
+                        raw_tile = stream.raw[0].getTile(x, y, z)
+                        # logging.debug(f"Tile {x}, {y}, {z} shape: {raw_tile.shape}")
```
pieleric, line 2791: This tests more the `img` functions, so you could move this to util/test/img_test.py .

src/odemis/acq/stream/_projection.py
```diff
@@ -1083,11 +1112,19 @@ class NeedRecomputeException(Exception):
...
         while need_recompute:
             z = self._zFromMpp()
-            rect = self._rectWorldToPixel(self.rect.value)
-            # convert the rect coords to tile indexes
-            rect = [l / (2 ** z) for l in rect]
-            rect = [int(math.floor(l / das.tile_shape[0])) for l in rect]
-            x1, y1, x2, y2 = rect
+
+            # calculate the image pixels inside the view (rect)
+            view_rect = self._rectWorldToPixel(self.rect.value)
```
pieleric, line 1117: I think to support rotation properly, you'll need to adjust _rectWorldToPixel(). Instead of trying to itself compute the coordinates, it should just compute the coordinates of the 4 corners of .rect, and then pass them to Stream.getPixelCoordinates(). This then returns 4 points in the "world" referential, you can then get the bounding box of these 4 points.

src/odemis/util/img.py
```diff
@@ -1289,3 +1288,56 @@ def max_intensity_projection(data: model.DataArray, axis: int = 0) -> model.Data
     md = data.metadata.copy()
     md[model.MD_DIMS] = "YX"
     return model.DataArray(proj, md)
+
```
pieleric, line 1291: 2 lines, like in-between the other functions?

pieleric (changes requested):

src/odemis/dataio/test/tiff_test.py
patrickcleeve2: This tests the actual reads from the tiled shadow tiff, which is only for DataArrayShadowPyramidalTIFF so I think it makes sense here. There are tests for the individuail utils in img_test.py

src/odemis/dataio/test/tiff_test.py
patrickcleeve2: This is to test the tiled read still works correctly after updating the image. I think it makes more sense here, because it actually tests the tile-read from disk and the projection. the projection tests in stream_test use mocked tile reads so they dont pick up these errors.

src/odemis/acq/stream/_projection.py
patrickcleeve2: I'm not planning to support rotation properly at this time, because it's going to require too much time and these images are usually 0 degrees or minimially rotated. passing the corners of the rect to getPixelCoordinates returns None, likely because of a -1 difference or rounding differences again. We also need to rotate the tile index calculation so its going to be a a fair chunk of work that isn't a priority right now.

pieleric: I see. I actually had in mind the SEM overview images that would have 180° rotation... but I realise that for now this images are not stored in pyramidal format (as it done in TFS Maps typically). So we don't need to worry (at least until Odemis also supports SEM overview acquisition!)

The issue with `getPixelCoordinates()` returning `None` is... "interesting". I hadn't thought of it, but indeed that the danger of passing the coordinates just at the border of the data. We might need to break this function into a lower-level one which only does referential conversion, without checking the shape of the data.

src/odemis/acq/stream/_projection.py
```diff
@@ -221,6 +222,16 @@ def _updateImage(self):
         except Exception:
             logging.exception("Updating %s %s image", self.__class__.__name__, self.stream.name.value)
 
+    def force_image_update(self):
+        """Forces the image to be recomputed entirely, and invalidates the cache for tiled images"""
+
+        if hasattr(self, "_projectedTilesCache") and hasattr(self, "_rawTilesCache"):
+            # invalidate the raw and projectted tiles cache
+            self._projectedTilesInvalid = True
```
pieleric, line 230: This attribute doesn't seem used (anymore). -> drop this line.

pieleric (changes requested):
Looks good to me. Just a very tiny remark.

pieleric (approved):

src/odemis/dataio/test/tiff_test.py
```diff
@@ -2729,6 +2732,77 @@ def test_convert_openfibsem_to_odemis_metadata(self):
...
+            model.MD_DIMS: "YX",
+        }
+
+        # create a dummy overview pyramid with the same problematic shape
+        das = model.DataArray(
+            numpy.zeros((8194, 7143), dtype=numpy.uint16), metadata=md
+        )
+
+        # export the dummy overview pyramid
+        export(PYRAMID_FILENAME, das, pyramid=True)
```
tmoerkerken, line 2764: Is this just temporarily stored next to the actual .py file or root? Maybe it's neater to work with a temp folder (tempfile.TemporaryDirectory) that is instantiated in a setup class. Note that I'm not familiar with the I/O of the used functions, so maybe it's not trivial to implement if they don't accept file paths.

src/odemis/dataio/test/tiff_test.py
```diff
@@ -2729,6 +2732,77 @@ def test_convert_openfibsem_to_odemis_metadata(self):
...
+        # open the pyramid and create a projection
+        data = open_acquisition(PYRAMID_FILENAME)
+        stream = data_to_static_streams(data)[0]
+        proj = RGBSpatialProjection(stream)
+
+        # simulate 'moving' the stream
+        initial_rect = proj.rect.value
+        move = (25e-3, 10e-3)
+        stream.raw[0].metadata[model.MD_POS_COR] = move
+        proj.force_image_update()
```
tmoerkerken, line 2775: Is there no "subscribe" like mechanism on this property that automatically triggers `force_image_update`? This seems like a manual trigger is needed. Can that lead to unexpected behavior when people forget to call it?

src/odemis/util/img.py
```diff
@@ -1289,3 +1288,62 @@ def max_intensity_projection(data: model.DataArray, axis: int = 0) -> model.Data
...
+
+def get_merged_raw_image(das: model.DataArrayShadow, z: int) -> model.DataArray:
+    """
+    Returns the entire raw data of DataArrayShadow at a given zoom level
+    das (DataArrayShadow): shadow of the raw data
```
tmoerkerken, line 1296: I feel like also putting the type in the docstring while utilizing Python's typing mechanism is double the work and prone to desync between docstring and function definition over time.

tmoerkerken (changes requested):

src/odemis/acq/stream/_projection.py
```diff
@@ -1131,6 +1167,7 @@ class NeedRecomputeException(Exception):
                 # image changed
                 need_recompute = True
 
+        logging.debug(f"{self.stream.name.value} (rect size: {rect_x*1e3:.4f} x {rect_y*1e3:.4f} mm) read {x2-x1} x {y2-y1} tiles at zoom: {z}, mpp: {self.mpp.value} m/px")
```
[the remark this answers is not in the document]
pieleric: Probably passes our CI linter, as we made it super lax (ie, it only cares about spaces). Moreover, there is no "official" rule in the coding guidelines (probably just because we forgot to define it). Still, typically you want a line < 100 char, or at most < 120 chars. 
@patrickcleeve2, could you break this line, or maybe just drop it if you don't think this log is useful now that the bug is fixed.

src/odemis/util/img.py
pieleric: Indeed, the new "style" is to only put the types as hint, not in the docstring. Here, Patrick "upgraded" the function to include type hints, so he could drop the ones from the docstring.

src/odemis/dataio/test/tiff_test.py
patrickcleeve2: That's deliberate, the only way to move the image is modify the metadata which isn't a 'public' api so force the responsibility on the caller force-refresh

src/odemis/dataio/test/tiff_test.py
patrickcleeve2: It's a temp file stored in the same directory. It's removed when the class exits (tearDown)

src/odemis/dataio/test/tiff_test.py
tmoerkerken: Can be resolved then

src/odemis/acq/stream/_projection.py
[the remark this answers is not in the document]
patrickcleeve2: I've split the line, i'd like to leave the debug info in the log as it makes it easier to investigate any of these issues if they happen for customers.

src/odemis/util/img.py
patrickcleeve2: removed

src/odemis/dataio/test/tiff_test.py
tmoerkerken: I get it, but it just doesn't look as safe as using an actual temp file or dir via said library, which lives in the os's tmp dir in a randomly named dirname (thus works with multiple parallel instances of the same or similar test). It probably works "fine", but utilizing actual temp file utility is safer in terms of isolation, automatic cleanup, and concurrency.

tmoerkerken (approved):

PR #2891, opened 2024-10-10
[METEOR-1208][REFACTOR] Allow the user to select whether to create or load a project on init

patrickcleeve2:
Currently, when starting the odemis-gui for meteor it automatically creates a new project directory. User's have requested that we allow them to specify the new project or select an existing project to load.

### DIFF
--- a/src/odemis/gui/cont/tabs/cryo_chamber_tab.py
+++ b/src/odemis/gui/cont/tabs/cryo_chamber_tab.py
@@ -90,8 +90,11 @@ class CryoChamberTab(Tab):
         self.txt_projectpath = self.panel.txt_projectpath
         self.btn_load_project.Hide()
 
-        # Create new project directory on starting the GUI
-        self._create_new_dir()
+        # flag to indicate whether the user has successfully created or
+        # loaded a project when first opening the gui. other actions are blocked,
+        # via a dialog, until a project is created or loaded.
+        self._is_initial_project_ready: bool = False
+
         self._cancel = False
 
         self._current_position = UNKNOWN  # position of the sample (regularly updated)
@@ -255,20 +258,22 @@ class CryoChamberTab(Tab):
         except AttributeError:  # No overview view
             pass
 
-    def _on_change_project_folder(self, evt):
+    def _on_change_project_folder(self, evt) -> bool:
         """
         Shows a dialog to change the path and name of the project directory.
-        returns nothing, but updates .conf and project path text control
+        Updates .conf and project path text control when a new project is created.
+        :return: True if the project was successfully created, False otherwise
         """
-        # TODO: do not warn if there is no data (eg, at init)
-        box = wx.MessageDialog(self.main_frame,
-                               "This will clear the current project data from Odemis",
-                               caption="Reset Project", style=wx.YES_NO | wx.ICON_QUESTION | wx.CENTER)
+        # do not warn if there is no data (eg, at init)
+        if self._is_initial_project_ready:
+            box = wx.MessageDialog(self.main_frame,
+                                "This will clear the current project data from Odemis",
+                                caption="Reset Project", style=wx.YES_NO | wx.ICON_QUESTION | wx.CENTER)
 
-        box.SetYesNoLabels("&Reset Project", "&Cancel")
-        ans = box.ShowModal()  # Waits for the window to be closed
-        if ans == wx.ID_NO:
-            return
+            box.SetYesNoLabels("&Reset Project", "&Cancel")
+            ans = box.ShowModal()  # Waits for the window to be closed
+            if ans == wx.ID_NO:
+                return False
 
         prev_dir = self.conf.pj_last_path
 
@@ -277,7 +282,7 @@ class CryoChamberTab(Tab):
         np = create_projectname(root_dir, self.conf.pj_ptn, count=self.conf.pj_count)
         new_dir = ShowChamberFileDialog(self.panel, np)
         if new_dir is None: # Cancelled
-            return
+            return False
 
         logging.debug("Selected project folder %s", new_dir)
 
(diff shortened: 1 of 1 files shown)


src/odemis/gui/cont/tabs/cryo_chamber_tab.py
```diff
@@ -910,6 +924,37 @@ def _sample_heater_to_va(self):
...
         Tab.Show(self, show=show)
 
+        # the first time the tab is shown, ask the user to create or load a project
+        if self.IsShown() and not self._is_initial_project_ready:
+
+            # stay in the loop until the user successfully creates or loads a project
+            while True:
+
+                # Ask the user to create or load a project
+                box = wx.MessageDialog(
```
nandishjpatel, line 934: Maybe you can handle this dialog and the subsequent dialogues (steps -> pages (buttons) in Wizard) using https://docs.wxpython.org/wx.adv.Wizard.html

Just a suggestion!

patrickcleeve2: Interesting, didn't know about this! Probably not for this PR, but might be worth re-doing the create/load project workflow with this

src/odemis/gui/cont/tabs/cryo_chamber_tab.py
```diff
@@ -910,6 +924,37 @@ def _sample_heater_to_va(self):
...
+                    self.main_frame,
+                    message="Create or Load a Meteor Project?",
```
nandishjpatel, line 936:
```suggestion
                    message="Create or Load a Meteor Project",
```

src/odemis/gui/cont/tabs/cryo_chamber_tab.py
```diff
@@ -370,15 +382,16 @@ def _reset_project_data(self):
 
     def _load_project_data(self, evt: wx.Event):
```
nandishjpatel, line 383:
```suggestion
    def _load_project_data(self, evt: wx.Event) -> bool:
```

?

src/odemis/gui/cont/tabs/cryo_chamber_tab.py
```diff
@@ -260,15 +261,16 @@ def _on_change_project_folder(self, evt):
         Shows a dialog to change the path and name of the project directory.
         returns nothing, but updates .conf and project path text control
         """
-        # TODO: do not warn if there is no data (eg, at init)
```
nandishjpatel, line 263: `def _on_change_project_folder(self, evt) -> Optional[bool]:`

src/odemis/gui/cont/tabs/cryo_chamber_tab.py
```diff
@@ -353,6 +359,12 @@ def _create_new_dir(self):
 
         self._change_project_conf(np)
 
+    def _create_project(self):
```
nandishjpatel, line 362:
```suggestion
    def _create_project(self) -> Optional[bool]:
```

src/odemis/gui/cont/tabs/cryo_chamber_tab.py
```diff
@@ -910,6 +924,37 @@ def _sample_heater_to_va(self):
     def Show(self, show=True):
         Tab.Show(self, show=show)
 
+        # the first time the tab is shown, ask the user to create or load a project
+        if self.IsShown() and not self._is_initial_project_ready:
+
+            # stay in the loop until the user successfully creates or loads a project
+            while True:
```
nandishjpatel, line 931: Just wondering if you can do away with this while statement. If not, its fine!

nandishjpatel (approved):

src/odemis/gui/cont/tabs/cryo_chamber_tab.py
pieleric: I think it's supposed to be `bool`. Please add a docstring to explain what False & True mean.

Then, every place with just `return` should be `return False`.

src/odemis/gui/cont/tabs/cryo_chamber_tab.py
```diff
@@ -90,8 +90,9 @@ def __init__(self, name, button, panel, main_frame, main_data):
         self.txt_projectpath = self.panel.txt_projectpath
         self.btn_load_project.Hide()
 
-        # Create new project directory on starting the GUI
-        self._create_new_dir()
+        # indicates whether a project has been initially created or loaded
+        self._is_initial_project_ready: bool = False
```
pieleric, line 94: I don't really get what this means. Does this mean that the data is explicitly stored in a project selected by the user (in contrast to the data being stored in the default new project folder created at init)?
Could you clarify this in the comment (or the variable name).

pieleric (changes requested):

src/odemis/gui/cont/tabs/cryo_chamber_tab.py
```diff
@@ -910,6 +924,37 @@ def _sample_heater_to_va(self):
...
+                    style=wx.YES_NO | wx.ICON_QUESTION | wx.CENTER,
+                )
+
+                box.SetYesNoLabels("&Load Project", "&Create Project",)
+                ans = box.ShowModal()  # Waits for the window to be closed
+
+                if ans == wx.ID_NO:
+                    logging.debug("creating project")
+                    ret = self._create_project()
+                if ans == wx.ID_YES:
```
pieleric, line 947: `elif`

src/odemis/gui/cont/tabs/cryo_chamber_tab.py
patrickcleeve2: No, we want the user to stay here until they either create a new project or load an existing one. So if they fail to do so, they get put back to the start of the flow until they do so correctly.

src/odemis/gui/cont/tabs/cryo_chamber_tab.py
patrickcleeve2: It's a flag to indicate whether the user has initially created a new project or loaded an existing one. The rest of the tab is blocked until they succesffuly do that.

tmoerkerken (approved):

pieleric (approved):

PR #2893, opened 2024-10-11
[DRAFT][METEOR-1204][FEAT] Add coordinate transform to TFS2 Posture Manager

patrickcleeve2:


install/linux/usr/share/odemis/sim/meteor-tfs2-sim.odm.yaml
```diff
@@ -0,0 +1,236 @@
...
+        FAV_POS_DEACTIVE: { 'rx': 0, 'rz': 1.9076449, 'x': -0.01529, 'y': 0.0506, 'z': 29.e-3 },
+        # XYZ ranges for SEM & METEOR
+        SEM_IMAGING_RANGE: {"x": [-10.e-3, 10.e-3], "y": [-30.e-3, 25.e-3], "z": [15.e-3, 37.e-3]},
+        FM_IMAGING_RANGE: {"x": [0.040, 0.054], "y": [-30.e-3, 25.e-3], "z": [15.e-3, 34.e-3]},
+        # Grid centers in SEM range
+        # Adjusted so that at init (0,0,0), it's at the Grid 1.
+        SAMPLE_CENTERS: {"GRID 1": {'x': 0, 'y': 0, 'z': 32.e-3}, "GRID 2": {'x': 2.98e-3, 'y': 2.46e-3, 'z': 32.e-3}},
+        # Mirroring values between SEM - METEOR
+        POS_COR: [0.0253126, 0.0024916],
+        CALIB: {"version": "tfs_2",
```
K4rishma, line 78: this is nice :)

src/odemis/driver/actuator.py
```diff
@@ -3550,3 +3550,160 @@ def stop(self, axes=None):
...
+            else:
+                logging.info("Axes %s of dependency are missing from .speed, so not providing it",
+                             set(self.axes) - speed_axes)
+
+        # TODO: correctly modify the reference, speed and axes range attributes
+
+    # def _updateAxesRange(self):
+    #     """
+    #     Calculate the axes range of position VA from the given range of dep's.
+    #     """
```
pieleric, line 3617: I don't think you'll need this, as normally the calibration values are not going to change live.

src/odemis/driver/actuator.py
```diff
@@ -3550,3 +3550,160 @@ def stop(self, axes=None):
...
+
+        self.referenced._set_value(refd, force_write=True)
+
+    @isasync
+    def moveRel(self, shift: Dict[str, float], **kwargs):
+        """
+        :param shift: The relative shift to be made
+        :param **kwargs: Mostly there to support "update" argument
+        """
+        # missing values are assumed to be zero
```
pieleric, line 3681: You probably should put the "standard" move checks:
```
        if not shift:
        self._checkMoveRel(shift)
```

src/odemis/driver/actuator.py
```diff
@@ -3550,3 +3550,160 @@ def stop(self, axes=None):
...
+        logging.debug("converted relative move from %s to %s", shift, shift_stage)
+        return self._dependency.moveRel(shift_stage, **kwargs)
+
+    @isasync
+    def moveAbs(self, pos: Dict[str, float], **kwargs):
+        """
+        :param pos: The absolute position to be moved to
+        :param **kwargs: Mostly there to support "update" argument
+        """
+
```
pieleric, line 3692: I'd put the standard check here too:
```
        if not pos:
        self._checkMoveAbs(pos)
```

src/odemis/driver/actuator.py
```diff
@@ -3550,3 +3550,160 @@ def stop(self, axes=None):
...
+        self.referenced._set_value(refd, force_write=True)
+
+    @isasync
+    def moveRel(self, shift: Dict[str, float], **kwargs):
+        """
+        :param shift: The relative shift to be made
+        :param **kwargs: Mostly there to support "update" argument
+        """
+        # missing values are assumed to be zero
+        shift_stage = self.pm.from_sample_stage_to_stage_movement(shift)
```
pieleric, line 3682: I wonder if a relative move is actually possible to compute in every cases (TFS, probably yes, but I'm more concerned with Tescan and the complex transformations). It might be easier to convert to an absolute position, (as .position + shift) and then do the same as moveAbs().

pieleric (changes requested):

src/odemis/driver/actuator.py
patrickcleeve2: I don't understand why computing an absolute movement would be possible, but relative wouldn't?

src/odemis/driver/actuator.py
```diff
@@ -3550,3 +3550,160 @@ def stop(self, axes=None):
...
+    """
+
+    def __init__(self, name, role, dependencies, posture_manager: MicroscopePostureManager, **kwargs):
+        """
+        dependencies (dict str -> actuator): name to objective lens actuator
+        """
+        if len(dependencies) != 1:
+            raise ValueError("SampleStage needs 1 dependency")
+
+        self._dependency = list(dependencies.values())[0]
```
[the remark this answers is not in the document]
patrickcleeve2: ok, its a sole dependency on stage bare. I don't think it needs optical focus because this only relates to stage positions/postures.

patrickcleeve2:
closing until integrated test

PR #2896, opened 2024-10-11
[feat] Store more component information in hardware settings metadata

pieleric:
This ensures that every component with a role is stored in the MD_EXTRA_SETTINGS info (not just the ones that are explicitly controlled by the GUI). Also adds info about the microscope itself.

### DIFF
--- a/src/odemis/acq/test/acq_test.py
+++ b/src/odemis/acq/test/acq_test.py
@@ -221,7 +221,7 @@ class SECOMTestCase(unittest.TestCase):
         """
         Check if extra metadata are saved
         """
-        settings_obs = SettingsObserver(model.getComponents())
+        settings_obs = SettingsObserver(model.getMicroscope(), model.getComponents())
         self.ccd.binning.value = (1, 1)  # make sure we don't save the right metadata by accident
         detvas = {'exposureTime', 'binning', 'gain'}
         s1 = stream.FluoStream("fluo2", self.ccd, self.ccd.data,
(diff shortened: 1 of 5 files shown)


src/odemis/gui/model/main_gui_data.py
```diff
@@ -250,10 +250,11 @@ def __init__(self, microscope):
             for c in components:
                 if c.role is None:
                     continue
+
+                comps_with_role.append(c)
```
patrickcleeve2, line 254: Why is this now outside the try block?

src/odemis/acq/test/acq_test.py
```diff
@@ -324,21 +314,11 @@ def setUpClass(cls):
         cls.spec_det_sel = model.getComponent(role="spec-det-selector")
         cls.ar_spec_sel = model.getComponent(role="ar-spec-selector")
 
-    @classmethod
```
patrickcleeve2, line 327: Why is the teardown removed?

pieleric: That's a little bit unrelated, but since a couple of years, `start_backend()` is more clever and can automatically stop the backend if the wrong one is in use, and start the right one. So we don't need all this code to dance around the possibility that the wrong backend is used. It especially saves time when you run test cases multiple times in a row, as the backend isn't restarted every time.
Most of the test cases have been updated already, but this one wasn't.

src/odemis/gui/model/main_gui_data.py
pieleric: That's the core of the change: we now observe *every* component, not just the ones which are used by the GUI explicitly.

patrickcleeve2 (approved):

tepals (approved):

PR #2903, opened 2024-10-21
[perf] make use of ParallelThreadPoolExecutor so that stage moves in x and y simultaneously

nandishjpatel:
JIRA Ticket: https://delmic.atlassian.net/browse/SDEV-974

### DIFF
--- a/src/odemis/driver/piezomotor.py
+++ b/src/odemis/driver/piezomotor.py
@@ -33,7 +33,14 @@ import serial
 import serial.tools.list_ports
 
 from odemis import model
-from odemis.model import Actuator, CancellableThreadPoolExecutor, CancellableFuture, isasync, HwError
+from odemis.model import (
+    Actuator,
+    CancellableFuture,
+    CancellableThreadPoolExecutor,
+    HwError,
+    ParallelThreadPoolExecutor,
+    isasync,
+)
 from odemis.util import to_str_escape
 from odemis.util.driver import getSerialDriver
 
@@ -111,6 +118,7 @@ class PMD401Bus(Actuator):
         self._speed_steps = {}  # axis name (str) -> int, speed in steps per meter
         self._counts_per_meter = {}  # axis name (str) -> float
         self._steps_per_meter = {}  # axis name (str) -> float
+        self._encoder_resolution = {}  # axis name (str) -> float
         self._portpattern = port
 
         # Parse axis parameters and create axis
@@ -144,8 +152,10 @@ class PMD401Bus(Actuator):
 
             if 'encoder_resolution' in axis_par:
                 self._counts_per_meter[axis_name] = 1 / axis_par['encoder_resolution']
+                self._encoder_resolution[axis_name] = axis_par['encoder_resolution']
             else:
                 self._counts_per_meter[axis_name] = 1 / DEFAULT_ENCODER_RESOLUTION
+                self._encoder_resolution[axis_name] = DEFAULT_ENCODER_RESOLUTION
                 logging.info("Axis %s has no encoder resolution, assuming %s.",
                              axis_name, DEFAULT_ENCODER_RESOLUTION)
 
@@ -172,7 +182,7 @@ class PMD401Bus(Actuator):
             axes_def[axis_name] = ad
 
         Actuator.__init__(self, name, role, axes=axes_def, inverted=inverted, **kwargs)
-        self._executor = CancellableThreadPoolExecutor(max_workers=1)  # one task at a time
+        self._executor = ParallelThreadPoolExecutor()
         self._ser_access = threading.RLock()
 
         # Connect to hardware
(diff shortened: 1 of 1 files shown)


pieleric (approved):

nandishjpatel:
@tepals did more tests and collected logs. There are still some improvements which can be done to save time, please leave the PR open until then.

src/odemis/driver/piezomotor.py
```diff
@@ -523,16 +524,22 @@ def _check_axis_error(self, axis_id: int, status: list):
...
-            # TODO: if not in closed-loop, it's probably because there is no encoder, so we need a different way
-            pos[axname] = self.getEncoderPosition(axis)
+        for axname in self._axis_map:
+            if axes is None or axname in axes:
+                # TODO: if not in closed-loop, it's probably because there is no encoder, so we need a different way
+                pos[axname] = self.getEncoderPosition(axname)
         logging.debug("Reporting new position at %s", pos)
         pos = self._applyInversion(pos)
+        if axes is not None:
+            pos_full = self.position.value
```
pieleric, line 540: You need to make a copy (of the dict). Otherwise, afterwards `self.position.value` already contains the new position, and when setting the "new" value, the VA will not notify the subscribers.

src/odemis/driver/piezomotor.py
```diff
@@ -804,12 +810,14 @@ def _sendCommand(self, cmd):
         :returns: (str) response
         """
         cmd += EOL
+        resp = b""
         with self._ser_access:
             logging.debug("Sending command %s", to_str_escape(cmd))
             try:
                 self._serial.write(cmd)
+                resp = self._serial.read_until(EOL)
             # TODO: what kind of exception is raised? Needs to be more specific.
```
pieleric, line 819: Apparently you can  now drop this comment.

src/odemis/driver/piezomotor.py
```diff
@@ -819,26 +827,6 @@ def _sendCommand(self, cmd):
...
-                        raise
-                    else:
-                        self._tryRecover()
-                        # don't send command again
-                        raise IOError("Failed to read from PMT Control firmware, "
-                                      "restarted serial connection.")
-                if not char:
-                    raise IOError("Timeout after receiving %s" % to_str_escape(resp))
-                else:
-                    resp += char
```
pieleric, line 764: It seems that indeed `read_until()` can replace all the code above. However, you probably still want to keep this check. Now, instead of checking for an empty char, it should check whether the last byte is EOL. If not, that's due to a timeout => raise IOError

src/odemis/driver/piezomotor.py
```diff
@@ -592,18 +645,16 @@ def isMovingClosedLoop(self, axis):
...
+        status = self.getStatus(axis)
+        self._check_axis_error(axis, status)
 
         # Check d3 (third status value) bit 2 (targetLimit: position limit reached) and bit 0 (targetReached)
-        if not d3 & 0b010:  # closed loop not active, thus it is not moving in closed loop
+        if not status[2] & 2:  # closed loop not active, thus it is not moving in closed loop
             logging.debug(f"Closed loop not active, therefore not moving in closed loop on axis {axis}.")
             return False
-        elif d3 & 0b101:
+        elif status[2] & 5:
```
pieleric, line 657: Why changing this? If it's called "d3", then it's best to call it "d3", not "status[2]" (but I don't remember the doc, if the doc has a different name, then anything is fine). Also, binary values are much more readable than an int, for this sort of bit checks. So I think it's best to keep the values in `0b...` format.

pieleric (changes requested):

src/odemis/driver/piezomotor.py
nandishjpatel: The other status checks here: https://github.com/nandishjpatel/odemis/blob/faeec8a1943d7310e1e71e1998bb078acb914403/src/odemis/driver/piezomotor.py#L510 have ints. Should I also convert those to binary?

In the tech manual, there is use of d1, d2, d3 and d4 will adhere to them

nandishjpatel: Also d1, d2, d3, d4 are already converted to int by `getStatus`

pieleric: As long as the ints are a power of 2, the math is easy to do in the head. So I don't really mind (though I think that writing in binary is always clearer). When there are multiple bits set, binary is a lot faster to read.

pieleric: > Also d1, d2, d3, d4 are already converted to int by `getStatus`
Binary values are int. It's just a matter of how they are written 🙂

tepals (approved):

src/odemis/driver/piezomotor.py
```diff
@@ -829,6 +830,8 @@ def _sendCommand(self, cmd):
                                   "restarted serial connection.")
             logging.debug("Received response %s", to_str_escape(resp))
 
+            if not resp.endswith(EOL):
+                raise IOError("Response %s does not ends with %s, it's due to a timeout" % (to_str_escape(resp), EOL))
```
pieleric, line 834: "%s" % EOL gives quite an odd output `"b'\r'"`. I'd just slightly rephrase this message and hard-code EOL:
"Timeout: no EOL received after response '%s'"

src/odemis/driver/piezomotor.py
```diff
@@ -592,18 +647,16 @@ def isMovingClosedLoop(self, axis):
...
+        d1, d2, d3, d4 = self.getStatus(axis)
+        self._check_axis_error(axis, d1, d2, d3, d4)
 
         # Check d3 (third status value) bit 2 (targetLimit: position limit reached) and bit 0 (targetReached)
-        if not d3 & 0b010:  # closed loop not active, thus it is not moving in closed loop
+        if not d3 & 2:  # closed loop not active, thus it is not moving in closed loop
             logging.debug(f"Closed loop not active, therefore not moving in closed loop on axis {axis}.")
             return False
-        elif d3 & 0b101:
+        elif d3 & 5:
```
pieleric, line 659: Please, leave these constants in binary. Or add as a comment "bit 0 and 2"

pieleric (changes requested):

nandishjpatel:
I reverted the commit https://github.com/delmic/odemis/pull/2903/commits/a23a6e4286788e0a245f98581283393f3dc8bde9 because the simulator backend does not start because of it

pieleric (approved):

PR #2904, opened 2024-10-22
[fix] technolution driver: reduce flyback in move full range.

tepals:
The flyback was set to its maximum value. However it can be set to the minimum value, because the sine has the same start and end, therefore no flyback time is needed.


src/odemis/driver/technolution.py
```diff
@@ -1082,7 +1082,7 @@ def moveFullRange(self):
         try:
             # Set dwell time, flyback time, and resolution to maximum values to avoid moving the descanners too fast
```
nandishjpatel, line 1083: Can you change the comment here?

pieleric (approved):

nandishjpatel (approved):

PR #2905, opened 2024-10-22
[fix][METEOR-553] Make the time estimation for overview map correct

K4rishma:


src/odemis/gui/win/acquisition.py
```diff
@@ -784,12 +791,6 @@ def __init__(self, parent, orig_tab_data):
         self.btn_secom_acquire.Bind(wx.EVT_BUTTON, self.on_acquire)
         self.Bind(wx.EVT_CLOSE, self.on_close)
 
-        # range for the autofocus (in m)
-        # default min/max range to 50 µm (half a grid square) to 1 mm (half a grid)
-        # this range should cover all cases for different magnifications (e.g. 20x, 50x, 100x)
-        self.focus_points_dist_ctrl.SetValueRange(50e-6, 1000e-6)
```
nandishjpatel, line 790: I think you should set the range for both the VA and ctrl

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -524,7 +523,8 @@ def estimateTime(self, remaining=None):
 
         stitch_time = (self._nx * self._ny * max_pxs * self._overlap) / self.STITCH_SPEED
         try:
-            move_time = max(self._guessSmallestFov(self._streams)) * (remaining - 1) / self._move_speed
+            # Extra time to account for actual stage movement happening in more than 2 dimensions
+            move_time = max(self._guessSmallestFov(self._streams)) * (remaining - 1) / self._move_speed  + 0.3 * remaining
```
nandishjpatel, line 527:
```suggestion
            move_time = max(self._guessSmallestFov(self._streams)) * (remaining - 1) / self._move_speed + 0.3 * remaining
```

src/odemis/acq/align/autofocus.py
```diff
@@ -546,8 +546,7 @@ def estimateAutoFocusTime(detector, scanner=None, steps=MAX_STEPS_NUMBER):
     scanner (None or model.Emitter): In case of a SED this is the scanner used
     Estimates overlay procedure duration
     """
-    # Add 0.5s per step to account for the focus movement (very roughly approximated)
-    return steps * estimateAcquisitionTime(detector, scanner) + steps * 0.5
+    return steps * estimateAcquisitionTime(detector, scanner)
```
pieleric, line 549: That means the movement of the focus is instantaneous? That sounds a little bit optimistic. On the other hand, steps is MAX_STEPS_NUMBER, which is probably quite pessimistic. 
I can imagine it makes (by chance?) the estimation more realistic on the METEOR, but how many other cases did it worsen? 
I think you could try to make the focus movement more precise by:
* taking the `focus` component as argument
* try to guess the distance to move (on average)
* using the `estimateMoveDuration()`

I'd also suggest to use a smaller number than "steps". For instance "steps / 2", but it might be good to check typically how many steps are required. 
Hopefully this also gives you a "not too bad" estimate on the METEOR, and doesn't change too much the estimate on other systems.

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -524,7 +523,8 @@ def estimateTime(self, remaining=None):
 
         stitch_time = (self._nx * self._ny * max_pxs * self._overlap) / self.STITCH_SPEED
         try:
-            move_time = max(self._guessSmallestFov(self._streams)) * (remaining - 1) / self._move_speed
+            # Extra time to account for actual stage movement happening in more than 2 dimensions
```
pieleric, line 526: What do you mean by "more than 2 dimensions"? Isn't stage move always just X&Y?

src/odemis/acq/stitching/_tiledacq.py
pieleric, line 527: Maybe use estimateMoveDuration() here? 
Also, I'm not certain that "remaining - 1" is correct, but I would think it should be the same number as for the "0.3 -...". So either use on both sides "remaining - 1" or "remaining" or something more clever. 
Maybe the "- 1" was to handle the case when a 1x1 overview is acquired?! But that would not be really explaining it for the most standard cases....

pieleric (changes requested):

tepals:
@K4rishma and @pieleric the `estimateTime` is also used to estimate the time for an overview image on the Fast EM. There it is underestimating how long it takes to acquire an overview image, so Karishma let me know when you have incorporated the suggestions from Éric. Then I can also test it on the Fast EM.

src/odemis/acq/align/autofocus.py
K4rishma: "I can imagine it makes (by chance?) the estimation more realistic on the METEOR, but how many other cases did it worsen?" The parameter that I deleted was added by me in a previous commit to improve the acquisition timer before. With that in mind, the other systems should work as usual ?

src/odemis/acq/stitching/_tiledacq.py
K4rishma: The meteor stage moves in two dimensions x and y. As y is linked to y and z of the stage-bare, the movement in stitching the tiles occurs in 3 dimensions.

src/odemis/acq/align/autofocus.py
```diff
@@ -539,15 +541,23 @@ def estimateAcquisitionTime(detector, scanner=None):
...
+    # adjust to rng_focus if provided
+    rng = focus.axes["z"].range
+    if rng_focus:
+        rng = (max(rng[0], rng_focus[0]), min(rng[1], rng_focus[1]))
+    distance = rng[1] - rng[0]
+    # Optimally, the focus starts from middle to minimum, then maximum. Then it goes back to the middle.
+    # optimistic guess
+    move_time = guessActuatorMoveDuration(focus, "z", distance) + 2 * guessActuatorMoveDuration(focus, "z", distance/2)
+    # pessimistic guess
+    acquisition_time = MAX_STEPS_NUMBER * estimateAcquisitionTime(detector, emt)
```
pieleric, line 559: If I understand correctly, you did some tests and the moves where estimated so short that it's almost null? That's why you had to bump up the acquisition time to MAX_STEPS_NUMBER? 
Maybe the moves should be increased? We could consider doing `steps_number` of steps of size `distance/steps_numbers`? Maybe add 10ms per steps, to estimate the overhead? My small experience with the autofocus is that the move of the axis is a large part of the time, so it'd be logical to also estimate it in the same way.

pieleric (changes requested):

K4rishma:
@nandishjpatel @tepals I have removed +2 seconds while estimating tiled_acquisition during the z stack acquisition. I have used the function from acqmng to estimate time during acquisition of z stack. From the past hardware testing sessions, this function has proved to be quite accurate till the last second.

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -1095,12 +1127,23 @@ def estimate_time(self, roi_idx=0, actual_time_per_roi=None) -> float:
...
+                tiled_time += tiled_acq_time
+
+                if self._use_autofocus:
+                    self.time_per_task.extend([autofocus_time, tiled_acq_time])
+                else:
+                    self.time_per_task.append(tiled_acq_time)
+
 
-            logging.debug(f"Estimated autofocus time: {autofocus_time} s, Tiled acquisition time: {tiled_time} s")
+            logging.debug(f"Estimated autofocus time: {autofocus_time*len(self.areas)} s, Tiled acquisition time: {tiled_time} s")
```
pieleric, line 1138:
```suggestion
            logging.debug(f"Estimated autofocus time: {autofocus_time * len(self.areas)} s, Tiled acquisition time: {tiled_time} s")
```

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -1081,11 +1112,12 @@ def estimate_time(self, roi_idx=0, actual_time_per_roi=None) -> float:
             acquisition_time = actual_time_per_roi * remaining_rois
```
pieleric, line 1112: There are now 2 different ways to report the time left. The old "actual_time_per_roi" way, which assumes every RoI is going to take on average the same time, and the new "time_per_task" way, which updates the progress according to the subfutures progress. 
I don't think you need to keep the old way. Especially, it's possible it's going to work against the new way, causing weird fluctuation in the expected time. (Of course, right now this function is probably  never called with more than 1 RoI, so it's not noticeable).

=> remove everything related to the "actual_time_per_roi". That's going to simplify the code quite a bit, and avoid odd fluctuations.

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -895,6 +923,7 @@ def run(self):
             raise
         finally:
             logging.info("Tiled acquisition ended")
+            logging.info(f"The actual time taken per tile for each action is {self._save_time}")
```
pieleric, line 926: That's probably only useful for debugging. We shouldn't keep this in the final version.

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -503,33 +508,40 @@ def estimateTime(self, remaining=None):
...
+    def _correct_tiled_acq_time(self):
+        """Compute average time taken to complete tiled acquisition for one tile position"""
+        observed_time = 0
+        for k,v in self._save_time.items():
+            if not v:  # empty list => no history
+                return None
+            observed_time += numpy.mean(v)
+        if observed_time == 0:
+            return None
+        return observed_time
```
pieleric, line 545: Same comment as for "time_per_action". I don't see the point of storing the time into 4 different categories, and once per tile, when the only thing you do is to sum them all together.

=> Estimate the time per tile based on the starting time, and the number of tiles acquired so far.

pieleric (changes requested):

src/odemis/acq/stitching/_tiledacq.py
K4rishma: To get an idea how time is spent, it is saved in 4 different categories. It will be helpful incase the code for time estimation needs to be updated and actual_time_per_roi needs to be removed. I understand, saving it in 4 categories is not helpful for the calculation but it is for gaining insights later through the logs. Will that be a good way to go?

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -503,33 +508,40 @@ def estimateTime(self, remaining=None):
...
-
-        # Estimate stitching time based on number of pixels in the overlapping part
-        max_pxs = 0
-        for s in self._streams:
-            for sda in s.raw:
-                pxs = sda.shape[0] * sda.shape[1]
-                if pxs > max_pxs:
-                    max_pxs = pxs
-
-        stitch_time = (self._nx * self._ny * max_pxs * self._overlap) / self.STITCH_SPEED
```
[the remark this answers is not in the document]
K4rishma: It is almost equating to zero in Meteor, so I will leave it like that. I forgot this code is used outside Meteor as well, my bad

src/odemis/acq/stitching/_tiledacq.py
K4rishma: I understand that the two methods are working against each other. Should I be keeping it same as before and remove the new changes or keep the new changes?

pieleric: Remove the old way.

src/odemis/acq/stitching/_tiledacq.py
pieleric: I see. Logging in production is not very common, but I understand the need here. In that case, you should separate the logging code from the time estimation code. This way, it'll be easier to remove the logging when it's not needed. 
=> Do one single estimation of the time it took from the beginning and average based on the number of tiles acquired to estimate the time per tile (and estimate the time left).

Independently from that, record the 4 duration per tile. At the end of the function, in addition to logging the raw dict, I'd recommend also logging the average time for each type of action. This should help with interpreting the results.

nandishjpatel (approved):

pieleric (approved):

src/odemis/gui/win/acquisition.py
```diff
@@ -1067,6 +1071,7 @@ def update_acquisition_time(self):
             self.lbl_acqestimate.SetLabel("Add a stream area to acquire.")
             return
 
+        focus_points_dist = self.focus_points_dist_ctrl.GetValue()
```
pieleric, line 1074:
```suggestion
        focus_points_dist = self.focus_points_dist.value
```

src/odemis/gui/win/acquisition.py
```diff
@@ -766,6 +766,14 @@ def __init__(self, parent, orig_tab_data):
         for sc in self.streambar_controller.stream_controllers:
             self._orig_entries += get_local_settings_entries(sc)
 
+        # range for the autofocus (in m)
+        # default min/max range to 50 µm (half a grid square) to 1 mm (half a grid)
+        # this range should cover all cases for different magnifications (e.g. 20x, 50x, 100x)
+        self.focus_points_dist_ctrl.SetValueRange(50e-6, 1000e-6)
+        self.focus_points_dist= model.FloatContinuous(MAX_DISTANCE_FOCUS_POINTS, range=(50e-6, 1000e-6))
```
pieleric, line 773: missing space before =
```suggestion
        self.focus_points_dist = model.FloatContinuous(MAX_DISTANCE_FOCUS_POINTS, range=(50e-6, 1000e-6))
```

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -1067,39 +1106,44 @@ def cancel(self, future):
...
+            else:
+                self.time_per_task.append(tiled_acq_time)
+
+
+        logging.debug(f"Estimated autofocus time: {autofocus_time * len(self.areas)} s, Tiled acquisition time: {tiled_time} s")
+        acquisition_time = autofocus_time + tiled_time
 
         return acquisition_time
 
+    def _pass_future_progress(self, future, start, end):
```
pieleric, line 1143: Let's clarify the types for this callback.
```suggestion
    def _pass_future_progress(self, future, start: float, end: float) -> None:
```

pieleric (changes requested):

src/odemis/gui/win/acquisition.py
pieleric: You have to update the object! `focus_points_dist_ctrl` -> `focus_points_dist`

pieleric (changes requested):
Could you try one more time with the VA?

src/odemis/gui/win/acquisition.py
K4rishma: My bad, the VA works. I have updated the PR

PR #2907, opened 2024-10-23
[fix] keysight: don't get too confused if device complains when turning it on

pieleric:
Depending on the previous settings (which are random, as the user might
have changed them), when setting the device to generate a square waveform,
it might complain that some parameters are incompatible (and reset
them).
The driver used to then give up, and be confused because it states the
device is not on, but internally think it's on.

=> If this happens, try again, to confirm it wasn't a big deal.
If still an error happens, keep the internal state (_is_powered)
synchronized with the .power VA.

It used to look like that:

2024-10-09 18:40:47,717	DEBUG	keysight:599:	Sending command 'sour1:func SQU\n'
2024-10-09 18:40:47,717	DEBUG	keysight:577:	Sending query 'SYST:ERR?\n'
2024-10-09 18:40:47,741	DEBUG	keysight:591:	Received answer -221,"Settings conflict; frequency forced duty cycle change"\n

### DIFF
--- a/src/odemis/driver/keysight.py
+++ b/src/odemis/driver/keysight.py
@@ -481,56 +481,93 @@ class TrueForm(model.Emitter):
 
         # Immediately update the internal state, to let the setters know they should apply the settings
         self._is_powered = p
-        if p:
-            # Powered on means a square waveform
-            self.setWaveform(self._channel, "SQU")
-            for c, t in self._tracking.items():
-                other_c = 3 - c
-                self.setTracking(other_c, t)
-
-            for c, lim in enumerate(self._limits):
-                self.setVoltageMin(c + 1, lim[0])
-                self.setVoltageMax(c + 1, lim[1])
-
-            # Reset all the settings by "setting" the VAs to the current values. We need this for 3 reasons:
-            # * if the VAs were changed while the power was off, they couldn't be applied, so we
-            #   need to set them now.
-            # * maybe the user has changed them in the meantime. So we need to put back the expected values.
-            # * it actually might be the first time the device is powered on, so we need to set them all.
-            try:
-                self.dutyCycle.value = self.dutyCycle.value
-            except TrueFormError as ex:
-                logging.warning("Failed to set the duty cycle to %s after power on: %s", self.dutyCycle.value, ex)
-            try:
-                self.period.value = self.period.value  # also updates .delay, in a safe way
-            except TrueFormError as ex:
-                logging.warning("Failed to set the period to %s after power on: %s", self.period.value, ex)
-
-            # Ends by enabling the output
-            self.setOutput(self._channel, True)
-            for c in self._tracking.keys():
-                self.setOutput(c, True)
-
-            if not self.requestLock():
-                # It's not a big deal if we can't lock the device. At worse the user might
-                # be able to change the settings manually, which might be a little confusing.
-                logging.warning("Failed to lock the device front panel")
-        else:  # Off
-            # Either disable the output (if off_voltage is None), or set a DC voltage
-            # Note: when power is off, the sync is still sent.
-            for c, volt in enumerate(self._off_voltage):
-                if volt is None:
-                    self.setOutput(c + 1, False)
-                else:
-                    # Note: in DC mode, some settings cannot be changed (like the phase & duty cycle).
-                    # Also, tracking doesn't work in DC mode, so should be re-enabled when powered on.
-                    self.applyDC(c + 1, volt)
-            try:
-                self.releaseLock()
-            except TrueFormError as ex:
-                logging.warning("Failed to release the lock: %s", ex)
+        try:
+            if p:
+                self._power_device_on()
+            else:
+                self._power_device_off()
+        except Exception:
+            # If an error occurs, the device might be in some sort of inconsistent state. It's pretty
+            # hard to do "something clever". However, at least we should assume intenal state is synchronized
+            # with the VA value (which will not be updated). This way, try to set the VA again, will
+            # allow to try again doing the same thing.
+            self._is_powered = not p
+            raise
 
         return p
+
+    def _power_device_on(self):
+        """
+        Configure the device to "on" state
+        """
+        # Powered on means a square waveform
+        try:
+            self.setWaveform(self._channel, "SQU")
+        except TrueFormError as ex:
+            # It happens that when setting the waveform, some current settings have to be reset,
+            # which the keysight reports as an error, but it's not an issue, as we set all the
+            # settings again afterwards. A second error would be a sign of a real issue.
+            logging.warning("Failed to set the waveform to square, will retry: %s", ex)
+            self.setWaveform(self._channel, "SQU")
+
+        # If required, set the other channel to track the main channel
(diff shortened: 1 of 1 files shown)


src/odemis/driver/keysight.py
```diff
@@ -481,56 +481,91 @@ def _set_power(self, p: bool) -> bool:
...
+        except TrueFormError as ex:
+            # It happens that when setting the waveform, some current settings have to be reset,
+            # which the keysight reports as an error, but it's not an issue, as we set all the
+            # settings again afterwards. A second error would be a sign of a real issue.
+            logging.warning("Failed to set the waveform to square, will retry: %s", ex)
+            self.setWaveform(self._channel, "SQU")
+
+        # If required, set the other channel to track the main channel
+        for c, t in self._tracking.items():
+            other_c = 3 - c
```
tepals, line 515: can you add a comment on why the other channel is `3 - c`?

src/odemis/driver/keysight.py
```diff
@@ -481,56 +481,91 @@ def _set_power(self, p: bool) -> bool:
...
-                logging.warning("Failed to release the lock: %s", ex)
+                # Tracking a channel is similar to activating it, so it can have the same issue
+                # as setting the waveform.
+                logging.warning("Failed to set the channel %s to track %s, will retry: %s", t, other_c, ex)
+                self.setTracking(other_c, t)
 
-        return p
+        for c, lim in enumerate(self._limits):
+            self.setVoltageMin(c + 1, lim[0])
+            self.setVoltageMax(c + 1, lim[1])
```
tepals, line 526: Can you add a comment to explain what this does?

tepals (changes requested):

tepals (approved):

src/odemis/driver/keysight.py
```diff
@@ -481,56 +481,93 @@ def _set_power(self, p: bool) -> bool:
...
+            # hard to do "something clever". However, at least we should assume intenal state is synchronized
+            # with the VA value (which will not be updated). This way, try to set the VA again, will
+            # allow to try again doing the same thing.
+            self._is_powered = not p
+            raise
 
-            for c, lim in enumerate(self._limits):
-                self.setVoltageMin(c + 1, lim[0])
-                self.setVoltageMax(c + 1, lim[1])
+        return p
```
tmoerkerken, line 497: I wonder why you want to return something in a setter? It's not like the value will be different from the input parameter.

src/odemis/driver/keysight.py
```diff
@@ -481,56 +481,93 @@ def _set_power(self, p: bool) -> bool:
...
+        Configure the device to "on" state
+        """
+        # Powered on means a square waveform
+        try:
+            self.setWaveform(self._channel, "SQU")
+        except TrueFormError as ex:
+            # It happens that when setting the waveform, some current settings have to be reset,
+            # which the keysight reports as an error, but it's not an issue, as we set all the
+            # settings again afterwards. A second error would be a sign of a real issue.
+            logging.warning("Failed to set the waveform to square, will retry: %s", ex)
```
tmoerkerken, line 510: So an initial error causes the keysight hardware to reset, and a successive call might then actually work?

src/odemis/driver/keysight.py
```diff
@@ -481,56 +481,93 @@ def _set_power(self, p: bool) -> bool:
...
+            self.setWaveform(self._channel, "SQU")
+
+        # If required, set the other channel to track the main channel
+        for c, t in self._tracking.items():
```
tmoerkerken, line 514: To me this is a bit confusing. I get that items in _tracking should follow the main channel, but now the non-tracking channel will be set as tracking via `setTracking`. Furthermore, it seems that this module only works with two channels. That will be one main and one in tracking. That means that _tracking will always be a list of max one, so the for-loop seems redundant.

src/odemis/driver/keysight.py
```diff
@@ -481,56 +481,93 @@ def _set_power(self, p: bool) -> bool:
...
+            logging.warning("Failed to set the period to %s after power on: %s", self.period.value, ex)
+
+        # Ends by enabling the output
+        self.setOutput(self._channel, True)
+        for c in self._tracking.keys():
+            self.setOutput(c, True)
+
+        if not self.requestLock():
+            # It's not a big deal if we can't lock the device. At worse the user might
+            # be able to change the settings manually, which might be a little confusing.
```
tmoerkerken, line 551: Maybe comment/verify that `_update_settings` then is able to sync back all the settings

src/odemis/driver/keysight.py
```diff
@@ -481,56 +481,93 @@ def _set_power(self, p: bool) -> bool:
...
+
+    def _power_device_off(self):
+        """
+        Configure the device to "off" state
+        """
+        # Either disable the output (if off_voltage is None), or set a DC voltage
+        # Note: when power is off, the sync is still sent.
+        for c, volt in enumerate(self._off_voltage):
+            if volt is None:
+                self.setOutput(c + 1, False)
```
tmoerkerken, line 562: You can do `enumerate(self._off_voltage, 1)` so you don't have to `+ 1` in the `setOutput` call. Reads a bit nicer, since then c (channel) directly maps to it's channel number and not to its 0-based version of it.

tmoerkerken (changes requested):

src/odemis/driver/keysight.py
pieleric: This is the behaviour of setters, as defined by VAs. They take the value to be set, and return the "actual" value set. For a boolean, it's usually always the same, but for floats, like the exposure time, it can easily be different. You can find an example here with the period: the setter computes the frequency based on the period, then it's passed to the hardware, which rounds it to which ever frequency is supported, and then this frequency is converted back to the "actual" period accepted by the hardware.

src/odemis/driver/keysight.py
pieleric: Yes, I agree with both remark. `setTracking` is strangely defined, but that's how the device works, so we cannot do anything. The reason for the loop is that originally we decided to make the driver generic enough so that if we want to support more devices from the same manufacturer with more channels, it would be easy to support. That means we end up for now with a strange loop that only loops 0 or once. In retrospect, it's probably over-engineering, but let's keep it like this (and this PR doesn't actually change this code, just move it around).

src/odemis/driver/keysight.py
pieleric: Indeed. Good idea.

src/odemis/driver/keysight.py
pieleric: I don't really get your comment. `_update_settings()` doesn't exist (anymore). The point here is just that in this unlikely scenario, the user can make things confused, but only if they explicitly try to do so (so the user should be aware of it, so not a big deal).

tmoerkerken: My bad, I was looking at the old file where that function still was used. So now setting polling is removed since that was problematic (device ended up in _remote_ mode, adjusting settings manually was then not possible)? Then it's fine as is.

tmoerkerken (approved):

PR #2910, opened 2024-10-23
[fix] SPARC Temporal acquisition failure

pieleric:
Due to the API refactoring missed some code using get_center_pxs. This
broke the "old" SEMTemporalMDStream.

=> fix the last 2 places where _get_center_pxs() had not been updated.

### DIFF
--- a/src/odemis/acq/stream/_sync.py
+++ b/src/odemis/acq/stream/_sync.py
@@ -717,7 +717,13 @@ class MultipleDetectorStream(Stream, metaclass=ABCMeta):
 
         # start with the metadata from the first point
         md = data_list[0].metadata.copy()
-        center, pxs = self._get_center_pxs(rep, (1, 1), data_list[0])
+        center_0 = md[MD_POS]
+        pxs = self._getPixelSize()
+        shape_tl = data_list[0].shape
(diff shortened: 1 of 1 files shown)


src/odemis/acq/stream/_sync.py
```diff
@@ -774,7 +780,11 @@ def _assembleTiles(self, rep, data_list):
 
         # start with the metadata from the first point
         md = data_list[0].metadata.copy()
-        center, pxs = self._get_center_pxs(rep, (T, S), data_list[0])
+        center_tl = md[MD_POS]
+        pxs = md[MD_PIXEL_SIZE]
+        shape_tl = (S, T)
```
nandishjpatel, line 785:
```suggestion
        shape_tl = (T, S)
```
previously it was `(T, S)`, was it incorrect?

nandishjpatel (approved):

sneepstefan (approved):

src/odemis/acq/stream/_sync.py
pieleric: That's indeed correct. That was one of my changes in the refactoring of the API: all arguments are sorted X, Y. (No more mix of "numpy order" and "conventional order").

PR #2911, opened 2024-10-23
[feat] Add plugin to load FAST-EM factory settings from a dropdown menu

tepals:
During debugging or service actions it can be necessary to be able to load the factory settings of the descanner. This will make sure the pattern is visible on the diagnostic camera. This plugin adds an option in the help dropdown that will show a pop-up before loading the factory settings.

### DIFF
--- a/plugins/fastem_set_factory_values.py
+++ b/plugins/fastem_set_factory_values.py
@@ -60,9 +60,7 @@ class SetFactoryValues(Plugin):
         self._acquisition_controller = self._acquisition_tab._acquisition_controller
 
         self.addMenu("Help/Development/Set factory values",
-                     self._set_factory_values,
-                     item_kind=wx.ITEM_NORMAL,
-                     pass_menu_item=False)
+                     self._set_factory_values)
 
     def _set_factory_values(self):
         """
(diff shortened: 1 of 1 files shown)


tepals:
image

plugins/fastem_set_factory_values.py
```diff
@@ -0,0 +1,97 @@
...
+                self.descanner.scanAmplitude.value = descanner_md.get(model.MD_SCAN_AMPLITUDE)
+                upload = True
+            if upload:  # upload params to ASM
+                self.mppc.data.get(dataContent="empty")
+
+            logging.debug("<ove the z component of the stage to roughly a good focus position")
+            ccd_md = self.ccd.getMetadata()
+            focus_pos = ccd_md.get(model.MD_FAV_POS_ACTIVE).get("z")
+            if focus_pos:
+                self.stage.moveAbs({"z": focus_pos})
```
nandishjpatel, line 97:
```suggestion
                self.stage.moveAbs({"z": focus_pos}).result()
```
I don't know if you wanna block or not

plugins/fastem_set_factory_values.py
```diff
@@ -0,0 +1,97 @@
...
+                self.descanner.scanOffset.value = descanner_md.get(model.MD_SCAN_OFFSET)
+                upload = True
+            if model.MD_SCAN_AMPLITUDE in descanner_md:
+                logging.debug("Update descanner scan amplitude to factory calibration")
+                self.descanner.scanAmplitude.value = descanner_md.get(model.MD_SCAN_AMPLITUDE)
+                upload = True
+            if upload:  # upload params to ASM
+                self.mppc.data.get(dataContent="empty")
+
+            logging.debug("<ove the z component of the stage to roughly a good focus position")
```
nandishjpatel, line 93:
```suggestion
            logging.debug("Move the z component of the stage to roughly a good focus position")
```

nandishjpatel (approved):

plugins/fastem_set_factory_values.py
```diff
@@ -0,0 +1,97 @@
...
+        If yes is clicked, the pop-up remains open until the factory settings are uploaded (~4 seconds).
+        """
+        box = wx.MessageDialog(
+            self.main_frame,
+            "Do you want to load factory settings for descanner offset and amplitude,"
+            " and set the z-stage to the favorite position?",
+            style=wx.YES_NO | wx.ICON_QUESTION | wx.CENTER,
+        )
+        ans = box.ShowModal()  # Waits for the window to be closed
+        if ans == wx.ID_YES:  # only set values when "yes" has been selected
```
pieleric, line 79: It's usually nicer to read the "normal" code at the lowest indentation. So I'd inverse the if like that: 
```suggestion
        if ans != wx.ID_NO:  # only set values when "yes" has been selected
           return
```

plugins/fastem_set_factory_values.py
```diff
@@ -0,0 +1,97 @@
...
+            descanner_md = self.descanner.getMetadata()
```
pieleric, line 80: I don't know enough about the FASTEM, but shouldn't it be an error or warning if there is no "factory metadata" when the user asks to reset to factory settings?

plugins/fastem_set_factory_values.py
```diff
@@ -0,0 +1,97 @@
...
+                "Not loading set-factory-values tool since acquisition tab is not present."
+            )
+            return
+
+        self._acquisition_controller = self._acquisition_tab._acquisition_controller
+
+        self.addMenu("Help/Development/Set factory values",
+                     self._set_factory_values,
+                     item_kind=wx.ITEM_NORMAL,
+                     pass_menu_item=False)
```
pieleric, line 65: These are the default values. I'd just drop them.

pieleric (changes requested):

plugins/fastem_set_factory_values.py
tepals: I might be getting lost in the double negation, but I think your suggestion should be 
```
        if ans == wx.ID_NO:  # only set values when "yes" has been selected
           return
```

plugins/fastem_set_factory_values.py
tepals: yes, that makes sense

plugins/fastem_set_factory_values.py
pieleric: Yes, indeed!

pieleric: Actually, I think the safest is `if ans != wx.ID_YES` (because maybe it can also return values like ID_CLOSE, etc?)

plugins/fastem_set_factory_values.py
```diff
@@ -0,0 +1,97 @@
...
+            if model.MD_SCAN_OFFSET in descanner_md:
+                logging.debug("Update descanner scan offset to factory calibration")
+                self.descanner.scanOffset.value = descanner_md.get(model.MD_SCAN_OFFSET)
+                upload = True
+            if model.MD_SCAN_AMPLITUDE in descanner_md:
+                logging.debug("Update descanner scan amplitude to factory calibration")
+                self.descanner.scanAmplitude.value = descanner_md.get(model.MD_SCAN_AMPLITUDE)
+                upload = True
+            if upload:  # upload params to ASM
+                self.mppc.data.get(dataContent="empty")
```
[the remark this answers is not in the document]
tepals: It is not necessary to define the key, but I prefer to keep it like this for consistency reason, because that is how we have written it in other places as well.

nandishjpatel (approved):

K4rishma (approved):

pieleric (approved):

PR #2915, opened 2024-10-25
[METEOR-960][FEAT] Add autoscript client for openfibsem control

patrickcleeve2:
Adds an autoscript version of the xt-client for openfibsem based control of the fibsem

Server side: https://bitbucket.org/delmic/xtlib-adapter/pull-requests/84

### DIFF
--- a/src/odemis/driver/autoscript_client.py
+++ b/src/odemis/driver/autoscript_client.py
@@ -1,0 +1,2048 @@
+# -*- coding: utf-8 -*-
+"""
+Created on 20 Feb 2024
+
+@author: Patrick Cleeve
+
+Copyright © 2024 Delmic
+
+This file is part of Odemis.
+
+Delmic Acquisition Software is free software: you can redistribute it and/or modify it under the terms of the GNU
+General Public License as published by the Free Software Foundation, either version 2 of the License, or (at your
+option) any later version.
+
+Delmic Acquisition Software is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even
+the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+more details.
+
+You should have received a copy of the GNU General Public License along with Delmic Acquisition Software. If not, see
+http://www.gnu.org/licenses/.
+"""
+import logging
+import math
+import queue
+import threading
+import time
+from concurrent.futures import CancelledError, Future
+from typing import Any, Dict, List, Optional, Tuple, Union
+
+import msgpack  # only used for debug information
+import msgpack_numpy
+import numpy
+import pkg_resources
+import Pyro5.api
+from Pyro5.errors import CommunicationError
+
+from odemis import model, util
+from odemis.driver.xt_client import check_and_transfer_latest_package
+from odemis.model import (
+    CancellableFuture,
+    CancellableThreadPoolExecutor,
+    DataArray,
+    HwError,
+    isasync,
+)
+
+Pyro5.api.config.SERIALIZER = 'msgpack'
+msgpack_numpy.patch()
+
+# Acquisition control messages
+GEN_START = "S"  # Start acquisition
+GEN_STOP = "E"  # Don't acquire image anymore
+GEN_TERM = "T"  # Stop the generator
+
+# List of known supported resolutions.
+# Although the API only provide the min/max of resolution in X/Y, not every value
+# or combination works. Actually, setting X always changes Y to the corresponding
+# value. Note that they are not all the same aspect ratio. "Legacy" resolutions
+# are ~8/7 while the new ones are 3/2. The "legacy" resolutions end up with a
+# larger vertical field of view.
+# TODO: non-standard resolutions are available in autoscript, so we should eventually enable them
+RESOLUTIONS = (
+    (192, 128),
+    (512, 442),
+    (1024, 884),
+    (2048, 1768),
+    (4096, 3536),
+    (768, 512),
+    (1536, 1024),
+    (3072, 2048),
+    (6144, 4096),
+)
+DETECTOR_RNG = ((768, 512), (6144, 4096))
+
+# imaging acquisition states
+IMAGING_STATE_IDLE = "Idle"
+IMAGING_STATE_RUNNING = "Running"
+IMAGING_STATE_PAUSED = "Paused"
+IMAGING_STATE_ERROR = "Error"
+
+# milling states
+MILLING_STATE_IDLE = "Idle"
+MILLING_STATE_RUNNING = "Running"
+MILLING_STATE_PAUSED = "Paused"
+MILLING_STATE_ERROR = "Error"
+
+# information on compatible versions
+debug_connection_info = f"""PYRO 5: {pkg_resources.get_distribution('Pyro5').version},
+msgpack-numpy: {pkg_resources.get_distribution('msgpack-numpy').version},
+msgpack_numpy_file: {msgpack_numpy.__file__},
+msgpack version: {pkg_resources.get_distribution('msgpack').version}
+msgpack_file: {msgpack.__file__}
+
+This is likely a dependencies issue:
+make sure you have the following matching versions of msgpack and msgpack-numpy installed:
+msgpack==0.5.6, msgpack-numpy==0.4.4
+or
+msgpack==1.0.3 msgpack-numpy==0.4.8
+"""
+
+class SEM(model.HwComponent):
+    """
+    Driver to communicate with autoscript software on TFS microscopes. autoscript is the software TFS uses to control their microscopes.
+    To use this driver the autoscript adapter developed by Delmic should be running on the TFS PC. Communication to the
+    Microscope server is done via Pyro5. The component is a parent to the scanner, stage, focus, and detector components, and supports both SEM and FIB.
+    """
+
+    def __init__(self, name, role, children, address, port: str = '4242', daemon=None,
+                 **kwargs):
+        """
+        :param name: str, Name of the microscope.
+        :param role: str, Role of the microscope.
+        :param children: dict, Dictionary with the children of the microscope.
+            "sem-scanner": dict, SEM scanner child configuration (required).
+            "sem-focus": dict, SEM focus child configuration (optional).
+            "sem-detector": dict, SEM detector child configuration (optional).
+            "fib-scanner": dict, FIB scanner child configuration (required).
+            "fib-focus": dict, FIB focus child configuration (optional).
+            "fib-detector": dict, FIB detector child configuration (optional).
+            "stage": dict, Stage child configuration (optional).
+            Note: At least one of the required scanners types must be included as a child.
+        :param address: str, server ip address for the microscope server (sim address is localhost)
+        :param port: str, server port of the Microscope server, default is '4242'
+        :param daemon: Pyro4.Daemon (or None), as defined in HwComponent.
+        :param kwargs: dict, Additional keyword arguments.
+        """
+
+        model.HwComponent.__init__(self, name, role, daemon=daemon, **kwargs)
+        self._proxy_access = threading.Lock()
+        try:
+            self.server = Pyro5.api.Proxy(f"PYRO:Microscope@{address}:{port}")
+            self.server._pyroTimeout = 30  # seconds
+            self._swVersion = self.server.get_software_version()
+            self._hwVersion = self.server.get_hardware_version()
+            if "adapter: autoscript" not in self._swVersion:
+                raise HwError("The connected server is not an autoscript server. Please check the xt adapter configuration."
+                              "The server software version is '%s'." % self._swVersion)
+            logging.debug(
+                f"Successfully connected to autoscript server with software version {self._swVersion} and hardware"
+                f"version {self._hwVersion}")
+        except CommunicationError as err:
+            raise HwError("Failed to connect to autoscript server '%s'. Check that the "
+                          "uri is correct and autoscript server is"
+                          " connected to the network. %s %s" % (address, err, debug_connection_info))
+        except OSError as err:
+            raise HwError("XT server reported error: %s." % (err,))
+
+        # Transfer latest xtadapter package if available
+        # The transferred package will be a zip file in the form of bytes
+        # check_and_transfer_latest_package(self)  # TODO: enable this once the package is available
+
+        # Create the scanner type child(ren)
+        # Check if at least one of the required scanner types is instantiated
+        scanner_types = ["sem-scanner", "fib-scanner"]  # All allowed scanners types
+        if not any(scanner_type in children for scanner_type in scanner_types):
+            raise KeyError("SEM was not given any scanner as child. "
+                           "One of 'sem-scanner', 'fib-scanner' need to be included as child")
+
+        if "sem-scanner" in children:
+            kwargs = children["sem-scanner"]
+            has_detector = "sem-detector" in children
+            self._scanner = Scanner(parent=self, daemon=daemon, channel="electron", has_detector=has_detector, **kwargs)
+            self.children.value.add(self._scanner)
+
+        if "fib-scanner" in children:
+            kwargs = children["fib-scanner"]
+            has_detector = "fib-detector" in children
+            self._fib_scanner = Scanner(parent=self, daemon=daemon, channel="ion", has_detector=has_detector, **kwargs)
+            self.children.value.add(self._fib_scanner)
+
+        # create the stage child, if requested
+        if "stage" in children:
+            ckwargs = children["stage"]
+            self._stage = Stage(parent=self, daemon=daemon, **ckwargs)
+            self.children.value.add(self._stage)
+
+        # create a focuser, if requested
+        if "sem-focus" in children:
+            ckwargs = children["sem-focus"]
+            self._focus = Focus(parent=self, daemon=daemon, channel="electron", **ckwargs)
+            self.children.value.add(self._focus)
+
+        if "fib-focus" in children:
+            ckwargs = children["fib-focus"]
+            self._fib_focus = Focus(parent=self, daemon=daemon, channel="ion", **ckwargs)
+            self.children.value.add(self._fib_focus)
+
+        # create a detector, if requested
+        if "sem-detector" in children:
+            ckwargs = children["sem-detector"]
+            self._detector = Detector(parent=self, daemon=daemon, channel="electron", **ckwargs)
+            self.children.value.add(self._detector)
+
+        if "fib-detector" in children:
+            ckwargs = children["fib-detector"]
+            self._fib_detector = Detector(parent=self, daemon=daemon, channel="ion", **ckwargs)
+            self.children.value.add(self._fib_detector)
+
+    def get_software_version(self) -> str:
+        """Returns: (str) the software version of the microscope."""
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            return self.server.get_software_version()
+
+    def get_hardware_version(self) -> str:
+        """Returns: (str) the hardware version of the microscope."""
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            return self.server.get_hardware_version()
+
+    def move_stage_absolute(self, position: Dict[str, float]) -> None:
+        """ Move the stage the given position. This is blocking.
+        :param position: absolute position to move the stage to per axes.
+            Axes are 'x', 'y', 'z', 't', 'r'.
+            The unit is meters for axes 'x', 'y' and 'z', and radians for axes 't', 'r'.
+        """
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            self.server.move_stage_absolute(position)
+
+    def move_stage_relative(self, position: Dict[str, float]) -> None:
+        """
+        Move the stage by the given relative position. This is blocking.
+        :param position: relative position to move the stage to per axes in m.
+            Axes are 'x', 'y', 'z', 'r', 't'. The units are meters for x, y, z and radians for r, t.
+        """
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            self.server.move_stage_relative(position)
+
+    def stop_stage_movement(self):
+        """Stop the movement of the stage."""
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            self.server.stop_stage_movement()
+
+    def get_stage_position(self) -> Dict[str, float]:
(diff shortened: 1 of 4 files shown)


nandishjpatel:
@patrickcleeve2 is this PR upto-date and should I review it? sorry for late!

patrickcleeve2:
> @patrickcleeve2 is this PR upto-date and should I review it? sorry for late!

Yep, i've rebased it now too

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2161 @@
...
+            self.server._pyroClaimOwnership()
+            self.server.blank_beam(channel)
+
+    def unblank_beam(self, channel: str) -> None:
+        """Unblank the selected beam."""
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            self.server.unblank_beam(channel)
+
+    def beam_is_blanked(self, channel: str) -> bool:
```
nandishjpatel, line 519:
```suggestion
    def is_beam_blanked(self, channel: str) -> bool:
```
Just a comment

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2161 @@
...
+        """return: (bool) True if the beam is blanked and False if the beam is not blanked."""
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            return self.server.beam_is_blanked(channel)
+
+    def beam_is_installed(self, channel: str) -> bool:
```
nandishjpatel, line 525:
```suggestion
    def is_beam_installed(self, channel: str) -> bool:
```
Just a comment

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2161 @@
...
+        :param x: float, the x value of the stigmator, unitless
+        :param y: float, the y value of the stigmator, unitless
+        :param channel: str, name of the channel
+        :return None
+        """
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            self.server.set_stigmator(x, y, channel)
+
+    def stigmator_info(self, channel: str):
```
nandishjpatel, line 618:
```suggestion
    def stigmator_info(self, channel: str) -> dict:
```

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2161 @@
...
+        Turn the beam on or off.
+        :param state: (bool) True to turn the beam on and False to turn the beam off.
+        :param channel: (str) Name of the channel to turn the beam on or off for.
+        :return None
+        """
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            self.server.turn_beam_on(state, channel)
+
+    def get_beam_is_on(self, channel: str) -> bool:
```
nandishjpatel, line 684:
```suggestion
    def is_beam_on(self, channel: str) -> bool:
```
Just a comment

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2161 @@
...
+        """
+        Get the last acquired image from the detector.
+        :param channel: (str) Name of one of the channels.
+        :return: (numpy.ndarray) the last acquired image.
+        """
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            return self.server.get_last_image(channel, wait_for_frame)
+
+    def start_acquisition(self, channel:str) -> None:
```
nandishjpatel, line 858:
```suggestion
    def start_acquisition(self, channel: str) -> None:
```

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2161 @@
...
+        """
+        Stop the acquisition of images.
+        :param channel: (str) Name of one of the channels.
+        :param wait_for_frame: (bool) If True, the function will wait until the current frame is acquired.
+        """
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            self.server.stop_acquisition(channel, wait_for_frame)
+
+    def get_imaging_state(self, channel:str) -> str:
```
nandishjpatel, line 877:
```suggestion
    def get_imaging_state(self, channel: str) -> str:
```

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2161 @@
...
+            return self.server.get_scanning_filter_info(channel)
+
+#### AUTO FUNCTIONS
+
+    def run_auto_contrast_brightness(self, channel: str) -> None:
+        """Run auto contrast brightness function (blocking)
+        """
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            self.server.run_auto_contrast_brightness(channel, parameters={})
```
nandishjpatel, line 925: Why not have parameters as an arg with default value above?

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2161 @@
...
+    def set_default_application_file(self, application_file: str = "Si") -> None:
+        """
+        Set the default application file.
+        :param application_file: (str) Name of the default application file.
+        """
+        with self._proxy_access:
+            self.server._pyroClaimOwnership()
+            self.server.set_default_application_file(application_file)
+
+    def set_default_patterning_beam_type(self, channel: str) -> None:
```
nandishjpatel, line 1043: Is the arg for this function correct? I had expect `patterning_beam_type`

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2161 @@
...
+        """
+        :param name (str): name of the Scanner
+        :param role (str): role of the Scanner
+        :param parent (SEM): parent of the Scanner
+        :param hfw_nomag (float): horizontal field width at nominal magnification
+        :param channel (str): name of the electron channel
+        :param has_detector (bool): True if a Detector is also controlled. In this case,
+          the .resolution, .scale and associated VAs will be provided too.
+        """
+        self.parent: SEM  # for type hinting
```
nandishjpatel, line 1083:
```suggestion
        self.parent: SEM = parent  # for type hinting
```

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2161 @@
...
+            # pixelSize is the same as MD_PIXEL_SIZE, with scale == 1
+            # == smallest size/ between two different ebeam positions
+            pxs = (fov / self._shape[0],
+                   fov / self._shape[0])
+            # pixelsize is inferred indirectly via resolution and fov
+            self.pixelSize = model.VigilantAttribute(pxs, unit="m", readonly=True)
+
+            # scanning resolution
+            resolution = self.parent.get_resolution(self.channel)
+            res_choices = set(r for r in RESOLUTIONS)
```
nandishjpatel, line 1189:
```suggestion
            res_choices = set(RESOLUTIONS)
```

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2161 @@
...
+                                                   readonly=True)
+        # To provide some rough idea of the step size when changing focus
+        # Depends on the pixelSize, so will be updated whenever the HFW changes
+        self.depthOfField = model.FloatContinuous(1e-6, range=(0, 1e3),
+                                                  unit="m", readonly=True)
+        self._updateDepthOfField()
+
+        if has_detector:
+            rng = ((768, 512), (6144, 4096))
+            self._shape = (rng[1][0], rng[1][1])
```
nandishjpatel, line 1179: This can be a constant at top of the file

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2161 @@
...
+
+    def _setVoltage(self, voltage: float) -> float:
+        self.parent.set_high_voltage(voltage, channel=self.channel)
+        return self.parent.get_high_voltage(self.channel)   # return the actual value used
+
+    def _setCurrent(self, current: float) -> float:
+        self.parent.set_beam_current(current, channel=self.channel)
+        return self.parent.get_beam_current(self.channel)  # return the actual value used
+
+    def _setBlanker(self, blank: bool) -> bool:
```
nandishjpatel, line 1313:
```suggestion
    def _setBlanker(self, blank: Optional[bool]) -> Optional[bool]:
```

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2161 @@
...
+            if fov != self.horizontalFoV.value:
+                self.horizontalFoV._value = fov
+                mag = self._hfw_nomag / fov
+                self.magnification._value = mag
+                self.horizontalFoV.notify(fov)
+                self.magnification.notify(mag)
+        except Exception:
+            logging.exception("Unexpected failure when polling settings")
+
+    def _setScale(self, value):
```
nandishjpatel, line 1253: Indicate type for all `_set*` functions which return a value?

src/odemis/driver/autoscript_client.py
nandishjpatel: Not sure of this is needed because in the `Component` class there is a setter and getter

```
    @property
    def parent(self):
        if self._parent:
            return self._parent()
        else:
            return None

    @parent.setter
    def parent(self, p):
        if p:
            assert isinstance(p, Component)
            self._parent = weakref.ref(p)
        else:
            self._parent = None

```

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2161 @@
...
+        :param channel (str): name of the acquistion channel (electron, ion)
+        :param kwargs: additional keyword arguments
+        """
+        # The acquisition is based on a FSM that roughly looks like this:
+        # Event\State |    Stopped    |   Acquiring    | Receiving data |
+        #    START    | Ready for acq |        .       |       .        |
+        #    DATA     |       .       | Receiving data |       .        |
+        #    STOP     |       .       |     Stopped    |    Stopped     |
+        #    TERM     |     Final     |      Final     |     Final      |
+        self.parent: SEM  # for type hinting
```
nandishjpatel, line 1410: Not sure of this is needed because in the `Component` class there is a setter and getter

```
    @property
    def parent(self):
        if self._parent:
            return self._parent()
        else:
            return None

    @parent.setter
    def parent(self, p):
        if p:
            assert isinstance(p, Component)
            self._parent = weakref.ref(p)
        else:
            self._parent = None

```

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2161 @@
...
+        model.Detector.__init__(self, name, role, parent=parent, **kwargs)
+        self._shape = (256,)  # Depth of the image
```
nandishjpatel, line 1412: This can be a constant defined at top of the file

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2161 @@
...
+        except TerminationRequested:
+            logging.debug("Acquisition thread requested to terminate")
+        except Exception as err:
+            logging.exception("Failure in acquisition thread: {}".format(err))
+        finally:
+            self._generator = None
+
+    def start_acquisition(self):
+        """Start acquiring images"""
+        try: # TODO: remove
```
nandishjpatel, line 1573: Still a TODO?

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2161 @@
...
+        Note: it expects that the acquisition is running.
+
+        timeout (0<=float): how long to wait to check (use 0 to not wait)
+        return (bool): True if needs to stop, False if data is ready
+        raise TerminationRequested: if a terminate message was received
+        """
+        tend = time.time() + timeout
+        t = time.time()
+        logging.debug("Waiting for %g s:", tend - t)
+        while self.parent.get_imaging_state(self._scanner.channel) != "Idle":
```
nandishjpatel, line 1633: `"Idle"` can be a constant

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2161 @@
...
+            except Exception:
+                if future._must_stop.is_set():
+                    raise CancelledError()
+                raise
+            finally:
+                future._was_stopped = True
+                # Update the position, even if the move didn't entirely succeed
+                self._updatePosition()
+
+    def _doMoveRel(self, future: 'Future', shift: Dict[str, float]):
```
nandishjpatel, line 1934: Is `Future` valid typing?

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2161 @@
...
+        pressure = self.parent.get_pressure()
+        if pressure != -1:  # -1 is returned when the chamber is vented
+            self.pressure._set_value(pressure, force_write=True)
+            logging.debug("Updated chamber pressure, %s Pa, vacuum state %s.", pressure, val["vacuum"])
+        else:
+            pressure = 100e3  # ambient pressure, Pa
+            self.pressure._set_value(pressure, force_write=True)
+            logging.warning("Couldn't read pressure value, assuming ambient pressure %s.", pressure)
+
+    def terminate(self):
```
nandishjpatel, line 1784: Docstring is needed terminate different to stop

src/odemis/driver/autoscript_client.py
patrickcleeve2: it was named this way to match the same functions in the xt_client

src/odemis/driver/autoscript_client.py
patrickcleeve2: channel is the same as beam_type (electron or ion)

nandishjpatel (approved):

pieleric (changes requested):
There is one important thing to still provide: the test cases. That said, a lot of the xt_client test cases should be reusable. So it might be relatively easily to implement!

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2163 @@
...
+                    #         logging.debug("Stopping acquisition early")
+                    #         self.stop_acquisition(wait_for_frame=False)
+                    #         break
+                    # except TimeoutError as err:
+                    #     logging.error(err)
+                    #     self.stop_acquisition(wait_for_frame=False)
+                    #     break
+
+                    # # Retrieve the image
+                    # image = self.parent.get_last_image(self._scanner.channel, wait_for_frame=True)
```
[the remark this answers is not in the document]
patrickcleeve2: code for non-blocking acquisition (cancellable mid acq), not tested yet. Lower priority than everything else

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2163 @@
...
+        raise NotImplementedError("Relative movements are not implemented for vacuum control. Use moveAbs instead.")
+
+    @isasync
+    def moveAbs(self, pos):
+        self._checkMoveAbs(pos)
+        return self._executor.submit(self._changePressure, pos["vacuum"])
+
+    def _changePressure(self, target):
+        """
+        target (0, 1): 0 for pumping to vacuum, 1 for venting the chamber
```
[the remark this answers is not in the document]
patrickcleeve2: removing chamber component entirely, not tested yet

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2163 @@
...
+    def _doMoveRel(self, future: CancellableFuture, shift: Dict[str, float]) -> None:
+        """
+        shift (dict): position in internal coordinates (ie, axes in the same
+           direction as the hardware expects)
+        """
+        # We don't check the target position fit the range, the autoscript-adapter will take care of that
+        self._moveTo(future, shift, rel=True)
+
+    @isasync
+    def move_vertical(self, pos: Dict[str, float]) -> Union[model.InstantaneousFuture, CancellableFuture]:
```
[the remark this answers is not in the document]
patrickcleeve2: moving to SampleStage

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2065 @@
...
+
+        # dwell time
+        dwell_time_info = self.parent.dwell_time_info(self.channel)
+        self.dwellTime = model.FloatContinuous(
+            self.parent.get_dwell_time(self.channel),
+            dwell_time_info["range"],
+            unit=dwell_time_info["unit"],
+            setter=self._setDwellTime)
+        # when the range has changed, clip the current dwell time value to the new range
+        self.dwellTime.clip_on_range = True
```
pieleric, line 1111: Please move dwellTime to the "has_detector" part. It doesn't make sense to control the dwell time if there is no acquisition.

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2065 @@
...
+    def _updateSettings(self) -> None:
+        """
+        Read all the current settings from the SEM and reflects them on the VAs
+        """
+        logging.debug("Updating SEM settings")
+        try:
+            dwell_time = self.parent.get_dwell_time(self.channel)
+            if dwell_time != self.dwellTime.value:
+                self.dwellTime._value = dwell_time
+                self.dwellTime.notify(dwell_time)
```
pieleric, line 1232: Once you move the dwell time to "has_detector", this block must also be moved down into the "if has_detector" block.

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2065 @@
...
+            scale_choices = set((self._shape[0] / r[0],) * 2 for r in res_choices)
+            self.scale = model.VAEnumerated(scale, scale_choices, unit="", setter=self._setScale)
+            self.scale.subscribe(self._onScale, init=True)  # to update metadata
+
+            # Just to make some code happy
+            self.translation = model.TupleContinuous((0, 0), range=[(-512, -512), (512, 512)], unit="px", readonly=True)
+
+        # not yet implemented:
+        # beam on
+        # beam blank
```
pieleric, line 1214: You have the code for `.blanker`, and even code that relies on it. It should be pretty straightforward to add the VA. Or is there something that doesn't work with the blanker? If so, please remove all code related to .blanker then.

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2065 @@
...
+                        md[model.MD_BEAM_SHIFT] = self._scanner.shift.value
+                    if hasattr(self._scanner, "horizontalFoV"):
+                        md[model.MD_BEAM_FIELD_OF_VIEW] = self._scanner.horizontalFoV.value
+                    if hasattr(self._scanner, "_acq_type"):
+                        md[model.MD_ACQ_TYPE] = self._scanner._acq_type
+                    md[model.MD_ACQ_DATE] = time.time()
+                    md.update(self._metadata)
+
+                    # Estimated time for an acquisition is the dwell time times the total amount of pixels in the image.
+                    if hasattr(self._scanner, "dwellTime") and hasattr(self._scanner, "resolution"):
```
pieleric, line 1522: Are there cases where these VAs are missing? I feel it's just some left-over from the xt_client which has a very basic FIB scanner without dwellTime.

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2065 @@
...
+        self._detector.stop_generate()
+
+
+class Stage(model.Actuator):
+    """
+    This is an extension of the model.Actuator class. It provides functions for
+    moving the TFS stage and updating the position.
+    """
+
+    def __init__(self, name: str, role: str, parent: SEM, rng:Dict[str, Tuple[float]]=None, **kwargs):
```
pieleric, line 1745: If `rng` accepts `None`, then it should be marked as "optional". Also, the Tuple type normally has one entry per element.
```suggestion
    def __init__(self, name: str, role: str, parent: SEM, rng: Optional[Dict[str, Tuple[float, float]]] = None, **kwargs):
```

src/odemis/driver/autoscript_client.py
```diff
@@ -0,0 +1,2065 @@
...
+        if rng is None:
+            rng = {}
+        stage_info = parent.stage_info()
```
pieleric, line 1748: If I understand correctly, autoscript allows to select, like xt_client, between "raw" and "specimen" coordinate system. Don't we have to explicitly force it to "raw" coordinates, using `set_default_stage_coordinate_system()` or the special coordinate system key in the stage position dict? Or maybe we should even support selecting the coordinate system via an argument, like the xt_client.Stage?

src/odemis/driver/test/autoscript_client_test.py
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+Created on 10 Dec 2024
+
+Copyright © 2024 Delmic
+
+This file is part of Delmic Acquisition Software.
```
pieleric, line 8: Could you adjust the header to say "Odemis" instead of "Delmic Acquisition Software" (that was the very early name!)

src/odemis/driver/test/autoscript_client_test.py
```diff
@@ -0,0 +1,118 @@
...
+
+### BEAM
+
+### DETECTOR
+
+### AUTOFUNCTIONS
+
+### IMAGING
+
+### MILLING
```
pieleric, line 118: Yes please!

pieleric (changes requested):

src/odemis/driver/autoscript_client.py
patrickcleeve2: I don't agree with this, see other client PR

src/odemis/driver/autoscript_client.py
patrickcleeve2: The coordinate system is initialised as RAW on the server side, so there is no need to do anything here. Currently all the code assumes that the coordinate system is raw so there is no use case for it being configurable here.

src/odemis/driver/autoscript_client.py
pieleric: See my response here then: https://github.com/delmic/odemis/pull/2976#issuecomment-2577368504

I think that the resolution, scale, and translation would be as required as the the dwellTime, in any case. Because it wouldn't be useful to do a "blind scan" without defining where to scan and with which resolution.

That said, this detail is not very blocking. So for now if you need it somewhere in your code, leave it there. Eventually we'll need to find a proper way to do blind scans...

pieleric (approved):

src/odemis/driver/autoscript_client.py
patrickcleeve2: @pieleric I've moved it under the has_detector block, to keep it consistent with the other client. can be merged

PR #2917, opened 2024-10-28
[fix] add metadata arg to actuator to support new chromatic corrections

patrickcleeve2:
Filter wheel component doesn't support metadata assignment, so it is currently broken for yaml files with chromatic corrections


pieleric:
No, that's just the microscope file which is wrongly written. See PR #2919

PR #2920, opened 2024-10-30
[NSRPARMA-52] SPARC Ephemeron mightyEBIC driver

sneepstefan:
Digital EBIC acquisition is needed specifically for the PARMA project. This basically consists of an extra/special stream for EBIC, a driver to communicate with their API for the controller and a modified SPARC acquisition procedure. The acquisition procedure must be simultaneous with the CL acquisition on our side. This is driven by TTL signaling which will trigger EBIC acquisition the moment we move to the pixel position of interest. Eventually the acquisition will be done and we would be able to request the data from the EBIC controller.

### DIFF
--- a/src/odemis/driver/ephemeron.py
+++ b/src/odemis/driver/ephemeron.py
@@ -30,9 +30,9 @@ import threading
 import time
 import queue
 from asyncio import AbstractEventLoop
+from socketserver import ThreadingMixIn
+from threading import Thread
 from typing import Optional, Tuple
-
-from numpy.ma.testutils import assert_almost_equal, assert_equal
 
 from odemis import model
 from odemis.model import Detector, oneway, MD_ACQ_DATE, HwError
@@ -214,7 +214,7 @@ class MightyEBIC(Detector):
         with self._acquisition_lock:
             self._wait_acquisition_stopped()
             self._acquisition_thread = threading.Thread(target=self._acquire_thread,
-                                                        name="EBIC acquire flow thread")
+                                                        name="EBIC acquisition thread")
             self._acquisition_thread.start()
 
     def _acquire_thread(self) -> None:
@@ -233,20 +233,16 @@ class MightyEBIC(Detector):
             scan_time = self._opc_client.calculate_scan_time(self.dwellTime.value, res[0], res[1])
             end_time = time.time() + DEFAULT_TIMEOUT + scan_time
 
-            f = self._opc_client.start_trigger_scan(self.dwellTime.value, 0, 1, 0, res[0], res[1], bool(self._opc_server_sim))
-            time.sleep(1)  # wait just a second to start up the co-routine and wait for the state update
+            self._opc_client.start_trigger_scan(self.dwellTime.value, 0, 1, 0, res[0], res[1], True)
 
             # while the scan controller is busy, check for a time-out or stop signal
             while self._opc_client.controller_state == STATE_NAME_BUSY:
                 if time.time() > end_time:
                     raise TimeoutError()
-                if self._acquisition_must_stop.wait(0.1):
+                # TODO check if the interval is correct
+                if self._acquisition_must_stop.wait(1):
                     self._opc_client.stop_scan()
-                    f.cancel()
                     return
-
-            # if the scan is finished the result of the co-routine should be instant
-            f.result()
 
             # get the data from the last acquisition and notify subscribers
             da = asyncio.run(self.read_data(res, md))
@@ -335,6 +331,8 @@ class MightyEBICSimulator(Server):
         self._stop_scan = False
         self._dt = 1e-5
         self._server_exception: Optional[str] = None
+        self._points_fast: Optional[int] = None
+        self._points_slow: Optional[int] = None
 
         self._t_simserver = threading.Thread(target=self._start_opc_simserver)
         self._t_simserver.start()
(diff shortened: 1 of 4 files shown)


src/odemis/driver/test/ephemeron_test.py
```diff
@@ -0,0 +1,186 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+Created on 7 Feb 2014
+
+Copyright © 2014 Kimon Tsitsikas, Delmic
```
pieleric, line 6: Please put the proper date and author :-)

src/odemis/driver/ephemeron.py
```diff
@@ -0,0 +1,747 @@
...
+
+    async def _set_controller_state(self, new_state):
+        async with self.client:
+            await self._ebic_controller_node.call_method(f"{NAMESPACE_INDEX}:set_controller_state", new_state)
+
+    def get_scan_time(self, dt, res_fast, res_slow) -> float:
+        f = asyncio.run_coroutine_threadsafe(self._get_scan_time(dt, res_fast, res_slow) , self._loop)
+        return f.result()
+
+    async def _get_scan_time(self, dt, res_fast, res_slow) -> float:
```
pieleric, line 697: If it wraps "calculate_scan_time", it should be called "calculate_scan_time" (for the non async version and with a `_` for the async version)

src/odemis/driver/ephemeron.py
```diff
@@ -0,0 +1,747 @@
...
+        async with self.client:
+            # dt pf, ps
+            st = await self.ebic_info_node.call_method(f"{NAMESPACE_INDEX}:calculate_scan_time", dt, res_fast, res_slow)
+            return st
+
+    def start_scan(self, req_dt, oversampling, channels, delay, p_fast, p_slow, sim=False):
+        f = asyncio.run_coroutine_threadsafe(self._start_scan(req_dt, oversampling, channels, delay, p_fast, p_slow, sim), self._loop)
+        return f
+
+    async def _start_scan(self, req_dt, oversampling, channels, delay, p_fast, p_slow, sim=False):
```
pieleric, line 707: Please call this function `_start_trigger_scan`, as it's the name of the function it wraps.

src/odemis/driver/ephemeron.py
```diff
@@ -0,0 +1,747 @@
...
+        if "localhost" in self._url:
+            # if the device should be simulated, start a simulated server first
+            self._opc_server_sim = MightyEBICSimulator(self._url, self._namespace, self)
+            self._t_simserver = threading.Thread(target=self._start_opc_simserver)
+            self._t_simserver.start()
+            # start the simulated server threaded but wait for it to be ready (running)
+            while not self._opc_server_sim.ready:
+                time.sleep(0.1)
+                if self._server_exception:
+                    raise HwError(self._server_exception.args[0])
```
pieleric, line 181: It should be possible to move all of this (and the start_opc_simserver method) to the simulator (init). Then, only provide a .terminate() function on the simulator to stop it (ie, do what the .terminate() currently does).

src/odemis/driver/test/ephemeron_test.py
```diff
@@ -0,0 +1,186 @@
...
+                                                2,
+                                                "opc.tcp://192.168.56.2:4840/mightyebic/server/",
+                                                "http://opcfoundation.org/UA/")
+        else:
+            # EBIC detector with a simulated server
+            cls.ebic_det = ephemeron.MightyEBIC("EBIC Scan Controller",
+                                                "ebic-detector",
+                                                2,
+                                                "opc.tcp://localhost:4840/freeopcua/server/",
+                                                "http://examples.freeopcua.github.io")
```
pieleric, line 67: You should change your simulator so that that these commands be as close as possible as the real hardware. So use the name URL and same namespace.
Then, the namespace can be just hard-coded in a constant.

src/odemis/driver/ephemeron.py
```diff
@@ -0,0 +1,747 @@
...
+
+
+class MightyEBIC(Detector):
+    def __init__(self, name: str, role: str, channel: int, url: str, namespace: str, **kwargs):
+        """
+        Initialise the EBIC controller
+        :param name (str): The name of the device configured through the configuration file
+        :param role (str): The role of the device configured through the configuration file
+        :param channel (int): The channel the device should use for hw triggering.
+        :param url (str): The url address to use with the OPC UA protocol
```
pieleric, line 154: Please indicate a special string for running the simulator. It shouldn't be a standard URL, but more like "simulator" or "fake". If the code sees this, it should replace it by a hard-coded URL that points to localhost, and start the simulator.
This is more explicit for the user, but also allows to have the mighty EBIC server run on the same computer as Odemis, without being confused as a simulator.

src/odemis/driver/ephemeron.py
```diff
@@ -0,0 +1,747 @@
...
+        self._acquisition_thread: Optional[threading.Thread] = None
+        self._acquisition_lock = threading.Lock()
+        self._acquisition_must_stop = threading.Event()
+
+        # Special event to request software unblocking on the scan
+        self.softwareTrigger = model.Event()
+
+        self._metadata[model.MD_DET_TYPE] = model.MD_DT_NORMAL
+        self._metadata[model.MD_HW_VERSION] = self._hwVersion
+        self._metadata[model.MD_SW_VERSION] = self._swVersion
```
pieleric, line 207: if the versions are unknown, no need to store them in the metadata. 
Once you can read the version number from a OPC node, then you can add this (back).

pieleric (changes requested):

src/odemis/driver/ephemeron.py
```diff
@@ -0,0 +1,748 @@
...
+            self._ebic_state_node = await state_node.get_child(f"{NAMESPACE_INDEX}:CurrentState")
+
+            # Create a new event loop
+            self._loop = asyncio.new_event_loop()
+            self._loop_thread = threading.Thread(target=self.start_event_loop, args=(self._loop,), daemon=True)
+            self._loop_thread.start()
+
+    def start_event_loop(self, loop):
+        asyncio.set_event_loop(loop)
+        loop.run_forever()
```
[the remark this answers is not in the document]
sneepstefan: Yes, specifically the command loop.run_forever() has to be set in a thread due to its blocking nature. Passing loop.run_forever() as target argument does not work either.

src/odemis/driver/ephemeron.py
```diff
@@ -0,0 +1,748 @@
...
+            res = self.resolution.value
+            md = self.getMetadata()
+            md[MD_ACQ_DATE] = time.time()
+
+            scan_time = self._opc_client.get_scan_time(self.dwellTime.value, res[0], res[1])
+            time.sleep(0.5)
+            end_time = time.time() + DEFAULT_TIMEOUT + scan_time
+
+            f = self._opc_client.start_scan(self.dwellTime.value, 0, 1, 0, res[0], res[1], True)
+            time.sleep(10)  # wait just a second to start up the co-routine
```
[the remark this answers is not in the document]
sneepstefan: That's a remnant from debugging, changing back to 1

pieleric (changes requested):

src/odemis/driver/test/ephemeron_test.py
```diff
@@ -0,0 +1,183 @@
...
+    """
+    Test case to test the functionality of the driver for an Ephemeron MightyEBIC detector.
+    """
+    @classmethod
+    def setUpClass(cls):
+        if not TEST_NOHW:
+            cls.ebic_det = ephemeron.MightyEBIC("EBIC Scan Controller",
+                                                "ebic-detector",
+                                                2,
+                                                "opc.tcp://192.168.56.2:4840/mightyebic/server/")
```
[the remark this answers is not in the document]
sneepstefan: No we do not, I will add a comment that when used with a VM, the address will be this particular one.

src/odemis/driver/test/ephemeron_test.py
```diff
@@ -0,0 +1,183 @@
...
+        """
+        base_st = 0.0
+
+        for dt_num, dt in enumerate(self.dwell_time_values):
+            for res_num, res in enumerate(self.resolution_values):
+                st_req = self.ebic_det._opc_client.calculate_scan_time(dt, res[0], res[1])
+                if res_num == 0:  # take the first scan time as base
+                    base_st = st_req
+                else:
+                    self.assertGreater(st_req, base_st)
```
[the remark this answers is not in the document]
sneepstefan: Indeed dt_num became obsolete later in my changes, I have removed it as suggested. Also added tests for dwell times at equal resolutions.

sneepstefan:
Closing PR due to future implementation of this driver code is done by @pieleric in a separate PR with changes to the code from testing.

PR #2921, opened 2024-10-30
[fix] Avoid e-beam.external "flickering" during SPARC SEM/CCD stream acquisition

pieleric:
Commit f24d37d55 (Avoid e-beam.external "flickering" during SPARC SEM MD
stream acquisition) introduced the idea of forcing the external mode
during acquisition. It turns out that the SEM CCD streams need the same.

So a similar code in this type of stream as well. This ensures that the
SEM stays in exernal mode all the time, and also reduces the acquisition
time, as sometimes switching external mode can take quite a long time
(ex: 0.5s).

### DIFF
--- a/src/odemis/acq/stream/_sync.py
+++ b/src/odemis/acq/stream/_sync.py
@@ -1235,6 +1235,13 @@ class SEMCCDMDStream(MultipleDetectorStream):
             # CCD to be sure it is not slowing thing down.
             self._emitter.dwellTime.value = self._emitter.dwellTime.clip(exp + readout)
 
+        if model.hasVA(self._emitter, "external") and self._emitter.external.value is None:
+            # When the e-beam is set to automatic external mode, it would switch on/off for every
+            # block of acquisition. This is not efficient, and can disrupt the e-beam. So we force
+            # "external" while the acquisition is running.
+            self._orig_hw_values[self._emitter.external] = self._emitter.external.value
+            self._emitter.external.value = True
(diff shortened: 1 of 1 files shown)


src/odemis/acq/stream/_sync.py
```diff
@@ -1235,6 +1235,13 @@ def _adjustHardwareSettings(self):
             # CCD to be sure it is not slowing thing down.
             self._emitter.dwellTime.value = self._emitter.dwellTime.clip(exp + readout)
 
+        if model.hasVA(self._emitter, "external") and self._emitter.external.value is None:
+            # When the e-beam is set to automatic external mode, it would switch on/off for every
+            # block of acquisition. This is not efficient, and can disrupt the e-beam. So we force
+            # "external" while the acquisition is running.
+            self._orig_hw_values[self._emitter.external] = self._emitter.external.value
```
tmoerkerken, line 1242: I saw that in some microscope software, the user could explicitly set the control mode to external. Does this change make that unnecessary now?

tmoerkerken (approved):

src/odemis/acq/stream/_sync.py
pieleric: No. It's the same functionality. On some hardware Odemis is able to control it. On these hardware, on some SPARC acquisition modes, this external mode would switch on/off rapidly, corresponding to each pixel. That fixes it.

PR #2923, opened 2024-10-31
[feat] add a custom filter metadata to support special filters

K4rishma:


pieleric:
Can you explain more the background behind this? Odemis supports multi-band filters. Isn't it what you're trying to support? I feel like "custom" is just incredibly broad in meaning, and at the same time the implementation seems like it would only support one per system.

K4rishma:
@pieleric  your suggestion works. Thank you. Closing this.

K4rishma:
Closing

PR #2925, opened 2024-11-01
[FIX] Remove chromatic correction from default meteor sim configuration file

patrickcleeve2:
The current meteor simulator config contains chromatic corrections which distorts the simulator images, making it difficult to use. This removes the corrections (scale, shift, shear, rotation) so the channels line up again.


K4rishma (approved):

K4rishma:
Maybe explain the situation where it becomes difficult to test with correction applied?

install/linux/usr/share/odemis/sim/meteor-sim.odm.yaml
```diff
@@ -200,10 +200,10 @@
...
     metadata: {
         CHROMATIC_COR: {
-            0.08: {"Pixel size cor": [1, 1],"Centre position cor": [2, 3] ,"Rotation cor": 0.1 ,"Shear cor": 0.2},
-            0.865398: {"Pixel size cor": [1, 1],"Centre position cor": [0, 0] ,"Rotation cor": 0 ,"Shear cor": 1},
-            1.650796: {"Pixel size cor": [1, 1],"Centre position cor": [4, 5] ,"Rotation cor": 0.2 ,"Shear cor": 0.3},
-            2.4361944: {"Pixel size cor": [2, 3],"Centre position cor": [6, 7] ,"Rotation cor": 0 ,"Shear cor": 0.4},
+            0.08: {"Pixel size cor": [1, 1],"Centre position cor": [0, 0] ,"Rotation cor": 0 ,"Shear cor": 0},
+            0.865398: {"Pixel size cor": [1, 1],"Centre position cor": [0, 0] ,"Rotation cor": 0 ,"Shear cor": 0},
+            1.650796: {"Pixel size cor": [1, 1],"Centre position cor": [0, 0] ,"Rotation cor": 0 ,"Shear cor": 0},
+            2.4361944: {"Pixel size cor": [1, 1],"Centre position cor": [0, 0] ,"Rotation cor": 0 ,"Shear cor": 0},
```
pieleric, line 206: I agree that the previous values didn't make sense because they were way too big. However, it should be possible to test slightly different values, so that we can see they do change when the filter changes.
I'd suggest:
* pixel size cor = `[0.999, 1.001]`
* center position cor = `[1.0e-9, -2.0e-9]`
* rotation cor = (-)0.001
* shear cor = (-)0.001

You can mix & match for the different filters.

patrickcleeve2: I've added a commented out version of the metadata with chromatic corrections, but left the default as no corection. If somebody wants to test with corrections, they can uncomment.

pieleric (approved):

PR #2926, opened 2024-11-01
[feat] SPARC: move to favourite position of the spec-ded-aligner at start-up

pieleric:
The axes of the spec-ded-aligner which must be referenced should move to
the favourite position after startup. Do it at init in the GUI.

### DIFF
--- a/src/odemis/gui/cont/tabs/sparc2_align_tab.py
+++ b/src/odemis/gui/cont/tabs/sparc2_align_tab.py
@@ -28,7 +28,7 @@ import logging
 import math
 import os
 import time
-from concurrent.futures import CancelledError
+from concurrent.futures import CancelledError, Future
 from functools import partial
 
 import pkg_resources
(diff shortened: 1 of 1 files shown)


src/odemis/gui/cont/tabs/sparc2_align_tab.py
```diff
@@ -274,6 +274,13 @@ def __init__(self, name, button, panel, main_frame, main_data):
 
         ccd_focuser_ext = None
         if "tunnel-lens-align" in tab_data.align_mode.choices:
+            # if there is a favourite position, go to that position at init
+            if main_data.spec_ded_aligner:
+                md = main_data.spec_ded_aligner.getMetadata()
+                if model.MD_FAV_POS_ACTIVE in md:
```
tmoerkerken, line 280: How is this favorite position set?

tmoerkerken (approved):

K4rishma (approved):

src/odemis/gui/cont/tabs/sparc2_align_tab.py
pieleric: Indeed, that's set in the microscope file, based on the alignment done during installation.
The user has the possibility to adjust it afterwards from the alignment tab... At least in theory, because we've just noticed this part doesn't work!

PR #2930, opened 2024-11-04
[fix][METEOR] prevent race condition if acquiring while streams are live

patrickcleeve2:
There is a race condition on the is_acquiring/_on_acquisition callback, if the stream is playing, the callback has to wait for the main ui thread to be available to start, while the acquisition can just start. the callback will pause the stream, and the acquisition will timeout. We add a synchronous pause of the streams to prevent this from occurring. synchronously here. 

Currently this completely breaks imaging on the meteor, so needs to be fixed urgently

### DIFF
--- a/src/odemis/gui/cont/acquisition/cryo_acq.py
+++ b/src/odemis/gui/cont/acquisition/cryo_acq.py
@@ -200,7 +200,6 @@ class CryoAcquiController(object):
 
         # disable the streams settings while acquiring
         if is_acquiring:
-            self._tab.streambar_controller.pauseStreams()
             self._tab.streambar_controller.pause()
         else:
             self._tab.streambar_controller.resume()
@@ -217,6 +216,11 @@ class CryoAcquiController(object):
 
         # the acquisition is started
         self._tab_data.main.is_acquiring.value = True
+
+        # NOTE: acquisition manager expects all streams to be paused before starting.
+        # this cannot be done in the is_acquiring callback as it can be delayed
+        # until after the acquisition starts
+        self._tab.streambar_controller.pauseStreams()
 
         # acquire the data
         if self._zStackActive.value:
@@ -287,6 +291,11 @@ class CryoAcquiController(object):
 
         # the acquisition is started
         self._tab_data.main.is_acquiring.value = True
+
+        # NOTE: acquisition manager expects all streams to be paused before starting.
+        # this cannot be done in the is_acquiring callback as it can be delayed
+        # until after the acquisition starts
+        self._tab.streambar_controller.pauseStreams()
 
         zparams: dict = {}
         if self._zStackActive.value:


src/odemis/gui/cont/acquisition/cryo_acq.py
```diff
@@ -218,6 +218,12 @@ def _on_acquire(self, _):
         # the acquisition is started
         self._tab_data.main.is_acquiring.value = True
 
+        # there is a race condition on the is_acquiring/_on_acquisition callback
+        # if the stream is playing, the callback can start, and then the acquisition starts
+        # the callback will pause the stream, and the acquisition will timeout
+        # so we pause the stream synchronously here
+        self._tab.streambar_controller.pauseStreams()
```
pieleric, line 225: Indeed, that's needed here. The reason is that `is_acquiring.value = True` calls the `on_acquisition` callback, but as the callback is "@call_in_wx_main()` protected, it will only be run *after* the current function is done running (as it's called by the GUI event loop). Eventually, the acquisition starts in a separate thread, and it might or not start the first stream before `on_acquisition` is called.

So, you should also *remove* the call to `pauseStreams()` in `on_acquisition`, line 203, to be entirely sure it will behave correctly.  That function should only do display-related changes (ie, disable/enable buttons).

src/odemis/gui/cont/acquisition/cryo_acq.py
```diff
@@ -288,6 +294,13 @@ def _acquire_at_features(self, _: wx.Event):
         # the acquisition is started
         self._tab_data.main.is_acquiring.value = True
 
+        # there is a race condition on the is_acquiring/_on_acquisition callback
+        # if the stream is playing, the callback can start, and then the acquisition starts
+        # the callback will pause the stream, and the acquisition will timeout
+        # so we pause the stream synchronously here
+        # Note: it occurs less here because of the stage movement buffer time
```
pieleric, line 301: I don't think you need to explain the bug in the comment. This explanation should be in the commit message. Here, you can say just "The acquisition manager expects all streams to be paused before starting". If you want, you can add "Note that this cannot be done by .is_acquiring callback, because it might be delayed".

pieleric (changes requested):

pieleric (approved):

src/odemis/gui/cont/acquisition/cryo_acq.py
```diff
@@ -288,6 +292,11 @@ def _acquire_at_features(self, _: wx.Event):
         # the acquisition is started
         self._tab_data.main.is_acquiring.value = True
 
+        # NOTE: acquisition manager expects all streams to be paused before starting.
+        # this cannot be done in the is_acquiring callback as it can be delayed
+        # until after the acquisition starts
+        self._tab.streambar_controller.pauseStreams()
+
```
K4rishma, line 299: We need the same part in _on_acquire() function, basically wherever is_acquiring is True

K4rishma (approved):

pieleric:
> In the initialization is_acquiring is set to True, should we also initialize pause_streams wherever the is_acquiring is set to True ? line 170

No need. First, at init, `is_acquiring` is set to `False`. In line 170, the `True` means that we should call the callback immediately (at init).   At init, there are never any stream playing, so no need to pause them.

PR #2935, opened 2024-11-07
[fix] la_spec: also move lens 2 when using hardware triggered detectors

pieleric:
The lens 2 move happened in _adjustHardwareSettings(). However, when the
detectors can be hardware triggered, _adjustHardwareSettingsHwSync()
is called. So the lens 2 wasn't moved in such case. Similarly, if using
the scan stage, another function is called.

Instead of trying to override all these functions, do the same as the
LASpectrumSettingsStream: override the _prepare_opm(). It's actually
more logical, and should work in all cases.

Also avoid providing the stream on spectrographs which are not after the
lens 2. In such case, the acquisition would work... but it wouldn't be
a large FoV.

### DIFF
--- a/plugins/la_spec.py
+++ b/plugins/la_spec.py
@@ -52,7 +52,7 @@ class LASpectrumSettingsStream(SpectrumSettingsStream):
         analyzer (None or Actuator with "pol" axis): the polarization analyzer.
           It should have at least the 7 "standard" positions
         """
-        super(LASpectrumSettingsStream, self).__init__(name, detector, dataflow, emitter, **kwargs)
+        super().__init__(name, detector, dataflow, emitter, **kwargs)
 
         self.l2 = l2
         if l2:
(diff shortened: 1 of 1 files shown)


plugins/la_spec.py
```diff
@@ -166,29 +166,38 @@ class LASEMSpectrumMDStream(SEMSpectrumMDStream):
...
+        f = futures.Future()
+        executeAsyncTask(f, self._set_optical_path)
+        return f
+
+    def _set_optical_path(self):
+        f = super()._prepare_opm()
+        f.result()
+        # Take care of the axes as soon as the OPM is done
+        # Note: it's sub-optimal, as the OPM will explicitly move the axes away
+        # while we maybe end-up putting them back.
```
K4rishma, line 190: I am not sure if adding the done callback will be useful or not
`def _set_optical_path(self):
    # Run the parent’s OPM preparation asynchronously
    f = super()._prepare_opm()
    
    # When the future completes, handle the axes
    f.add_done_callback(self._handle_axes_after_opm)

def _handle_axes_after_opm(self, future):
    # This method runs when the OPM future is done
    future.result()  # This gets the result of the OPM call (now complete)
    self._sccd._changeLensAxes()  # Adjust the axes afterward
`

K4rishma (approved):

plugins/la_spec.py
pieleric: Ideally I would write it the way you proposed, however, the problem is that _prepare_opm() has to return a Future which is *completed* when all the axes are at the right place, because immediately after the acquisition starts. With your suggestion, the returned Future would be completed as soon as the movements are started. That's not good.

tmoerkerken (approved):

PR #2936, opened 2024-11-13
[fix] ENZEL alignment tab should pause stream when exiting the tab

pieleric:
It would pause the streams when changing mode inside the tab... but not
pause them when going to another tab, like the "localization" tab.
That's very confusing, because this leaves the detectors active, which
can cause lots of odd behaviour in the other procedures.

=> rename pauseAllStreams() to pauseStreams() of the stream bar
controller so that it has the same API as the other stream bar controllers.

### DIFF
--- a/src/odemis/gui/cont/tabs/enzel_align_tab.py
+++ b/src/odemis/gui/cont/tabs/enzel_align_tab.py
@@ -294,7 +294,7 @@ class EnzelAlignTab(Tab):
         """
         Sets the Z alignment mode with the correct streams, stream controllers, instructions and alignment controls.
         """
-        self.stream_bar_controller.pauseAllStreams()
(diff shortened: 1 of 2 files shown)


src/odemis/gui/cont/tabs/enzel_align_tab.py
```diff
@@ -65,7 +65,7 @@ def __init__(self, name, button, panel, main_frame, main_data):
         # Controls the stage movement based on the imaging mode
         self.posture_manager = main_data.posture_manager
 
-        # Check if the stage has FAV positions in its metadata
+        # Check if the stage ha s FAV positions in its metadata
```
K4rishma, line 68: small typo

K4rishma (approved):

tmoerkerken (approved):

PR #2945, opened 2024-11-20
[feat] make fastem beam blanking in between fields optional

tepals:
For some samples the beam damage is minimal when the beam is left unblanked during stage moves. As beam blanking and unblanking takes ~0.16 s, this feature makes beam blanking optional to increase throughput.

### DIFF
--- a/src/odemis/gui/xmlh/resources/panel_tab_fastem_acqui.xrc
+++ b/src/odemis/gui/xmlh/resources/panel_tab_fastem_acqui.xrc
@@ -73,6 +73,13 @@
                           <border>10</border>
                         </object>
                         <object class="sizeritem">
+                          <object class="wxCheckBox" name="chk_beam_blank">
+                            <label>Turn off beam blanking in between fields</label>
+                          </object>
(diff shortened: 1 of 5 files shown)


patrickcleeve2 (approved):

src/odemis/gui/xmlh/resources/panel_tab_fastem_acqui.xrc
```diff
@@ -72,6 +72,13 @@
                           <flag>wxALL|wxEXPAND</flag>
                           <border>10</border>
                         </object>
+                        <object class="sizeritem">
+                          <object class="wxCheckBox" name="chk_beam_blank_off">
+                            <label>Turn off beam blanking in between fields</label>
```
pieleric, line 77: It feels like double negation... which is negated if you don't check the box. It's a little confusing.
How about "Do not blank beam in between fields" or even "Blank beam in between fields" (in which case the check box has to be inverted)?

pieleric (changes requested):

pieleric (approved):

PR #2948, opened 2024-11-21
[fix] correct the check to find which axes are requested to move. 

K4rishma:
Now it correctly takes into account if the requested move is relative or not, to compare the axes

### DIFF
--- a/src/odemis/driver/xt_client.py
+++ b/src/odemis/driver/xt_client.py
@@ -2327,8 +2327,13 @@ class Stage(model.Actuator):
                 self.parent.move_stage(pos, rel=rel)
                 time.sleep(0.1)  # It takes a little while before the stage is being reported as moving
 
+                # Get the target position in absolute coordinates
+                if rel:
+                    target_pos = {ax: val + orig_pos[ax] for ax, val in pos.items()}
+                else:
+                    target_pos = pos
                 # Drop axes from the original position, which are not important because they have not moved
-                orig_pos = {a: orig_pos[a] for a, nv in pos.items() if nv != orig_pos[a]}
+                orig_pos = {a: orig_pos[a] for a, nv in target_pos.items() if nv != orig_pos[a]}
 
                 # Wait until the move is over.
                 # Don't check for future._must_stop because anyway the stage will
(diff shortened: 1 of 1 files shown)


patrickcleeve2 (approved):

src/odemis/driver/xt_client.py
```diff
@@ -2327,8 +2327,14 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
...
 
+                # Get the target position in absolute coordinates
+                if rel:
+                    target_pos = {ax: val + pos[ax] for ax, val in orig_pos.items()}
+                else:
+                    target_pos = {ax: pos[ax] for ax, val in orig_pos.items()}
+
                 # Drop axes from the original position, which are not important because they have not moved
-                orig_pos = {a: orig_pos[a] for a, nv in pos.items() if nv != orig_pos[a]}
+                orig_pos = {a: orig_pos[a] for a, nv in target_pos.items() if nv != orig_pos[a]}
```
nandishjpatel, line 2337: Do you really need this? At line 2327 one is already requesting a move to `pos` stating its `rel`. Maybe the `time.sleep(0.1)` is not enough in some cases?

K4rishma: We need orig position a bit later when we start checking if the stage have moved to the requested position. Does that answer your question?

nandishjpatel (approved):

pieleric (changes requested):
This code doesn't work. You need to run the test cases before making a pull-request!

To test, it's written in the xt_client_test:
Start the simulator with `python xtadapter/server_sim.py`
Then run the tests with:
`env TEST_NOHW=sim python3 src/odemis/driver/test/xt_client_test.py --verbose TestMicroscope.test_move_stage`

K4rishma:
> This code doesn't work. You need to run the test cases before making a pull-request!
> 
> To test, it's written in the xt_client_test: Start the simulator with `python xtadapter/server_sim.py` Then run the tests with: `env TEST_NOHW=sim python3 src/odemis/driver/test/xt_client_test.py --verbose TestMicroscope.test_move_stage`

Sorry, I have fixed and tested it

src/odemis/driver/xt_client.py
```diff
@@ -2327,8 +2327,13 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
                 self.parent.move_stage(pos, rel=rel)
                 time.sleep(0.1)  # It takes a little while before the stage is being reported as moving
 
+                # Get the target position in absolute coordinates
+                if rel:
+                    target_pos = {ax: val + pos[ax] for ax, val in pos.items()}
```
pieleric, line 2332: I think this should be `ax: orig_pos[ax] + val` . The current code just doubles the values in pos!

pieleric (changes requested):

PR #2949, opened 2024-11-22
[fix] add check if tescan stage moves to the requested position otherwise…

K4rishma:
… raise an error

Works on simulator
Tested on hardware

### DIFF
--- a/src/odemis/driver/tescan.py
+++ b/src/odemis/driver/tescan.py
@@ -20,21 +20,25 @@ You should have received a copy of the GNU General Public License along with
 Odemis. If not, see http://www.gnu.org/licenses/.
 '''
 
-import queue
 import gc
 import logging
 import math
+import queue
+import re
+import socket
+import threading
+import time
+import weakref
+from typing import Dict
+
 import numpy
+from tescan import sem, CancelledError
+
 from odemis import model, util
 from odemis.model import (HwError, isasync, CancellableThreadPoolExecutor,
                           roattribute, oneway)
 from odemis.util import TimeoutError
-import re
-import socket
-from tescan import sem, CancelledError
-import threading
-import time
-import weakref
+from odemis.util.driver import isNearPosition
 
 ACQ_CMD_UPD = 1
 ACQ_CMD_TERM = 2
@@ -1027,6 +1031,13 @@ class SEMDataFlow(model.DataFlow):
             self._evtq.get()
 
 
+STAGE_WAIT_DURATION = 20e-03  # s
+STAGE_WAIT_TIMEOUT = 5  # s
+STAGE_FRACTION_TOTAL_MOVE = 0.01  # tolerance of the requested stage movement
+STAGE_TOL_LINEAR = 1e-06  # in m, minimum tolerance
+STAGE_TOL_ROTATION = 0.00436  # in radians (0.25 degrees), minimum tolerance
+
+
 class Stage(model.Actuator):
     """
     This is an extension of the model.Actuator class. It provides functions for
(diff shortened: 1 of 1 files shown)


src/odemis/driver/tescan.py
```diff
@@ -1096,6 +1113,57 @@ def _pollXYZ(self):
...
+            movement_req_linear = [movement_req[ax] for ax in linear_axes_to_check]
+            tol_linear = max(min(movement_req_linear), STAGE_TOL_LINEAR)  # m
+
+        if rotational_axes_to_check:
+            movement_req_rotational = [movement_req[ax] for ax in rotational_axes_to_check]
+            tol_rotation = max(min(movement_req_rotational), STAGE_TOL_ROTATION)  # radians
+
+        axes_to_check = linear_axes_to_check | rotational_axes_to_check
+
+        while not isNearPosition(current_pos=current_pos, target_position=target_pos,
```
pieleric, line 1154: If you call this function at the end of the move, it doesn't make sense to keep waiting. Especially, it doesn't seem that the ._position has any chance to update...

pieleric (changes requested):
With so much "clean-up" mixed up with the actual change, it would be nicer to have all the clean-up as a separate commit.

src/odemis/driver/tescan.py
```diff
@@ -1096,6 +1106,58 @@ def _pollXYZ(self):
         except Exception:
             logging.exception("Unexpected failure during XYZ polling")
 
+    def _checkPosition(self, ref_pos: Dict[str, float], pos: Dict[str, float], rel: bool = False):
+        """
+        Checks if the current stage position is equal to the given position.
```
nandishjpatel, line 1111:
```suggestion
        Checks if the current stage position is near the given position.
```

nandishjpatel (approved):

K4rishma:
@pieleric it is ready to review

src/odemis/driver/tescan.py
```diff
@@ -1096,6 +1106,58 @@ def _pollXYZ(self):
...
+        :param ref_pos: stage position before movement
+        :param pos: requested stage position in relative or absolute coordinates
+        :param rel: By default False, requested stage position is in absolution position otherwise relative
+        """
```
pieleric, line 1115: Please add a :raise: docstring to indicate it will fail if the stage doesn't reach the target position within a given time.

src/odemis/driver/tescan.py
```diff
@@ -1096,6 +1106,58 @@ def _pollXYZ(self):
...
+        if rel:
+            target_pos = {ax: val + pos[ax] for ax, val in pos.items()}
+        else:
+            target_pos = {ax: pos[ax] for ax, val in pos.items()}
+
+        # Drop axes from the original position, which are not important because they have not moved
+        orig_pos = {a: ref_pos[a] for a, nv in target_pos.items() if nv != ref_pos[a]}
+
+        # TODO: base the timeout on stage movement time estimation, needs to be checked on hardware
+        wait_stage_move = 5  # s
```
pieleric, line 1126: Let's make this an argument (with default value  = 5s). That should make it clearer that the function will time-out.

src/odemis/driver/tescan.py
pieleric, line 1111: It doesn't just *check*, it also *waits*. Please indicate that in the docstring.

src/odemis/driver/tescan.py
```diff
@@ -1158,6 +1222,7 @@ def _doMoveAbs(self, pos: dict):
 
             logging.debug("Stage move to %s completed", pos)
             self._updatePosition()
+            self._checkPosition(pos_before_move, pos, rel=False)
```
pieleric, line 1225: Probably the most important part of this whole change is missing: explain *why* this is needed. If I understand correctly, that's because the stage might fail to move, but there is no way to detect it explicitly, so you need to detect a failure to move by just comparing the requested position to the final position? Or is it also like on the TFS stage, where it can report it's finished moving, but it's not?

pieleric (changes requested):

src/odemis/driver/tescan.py
K4rishma: Added to the comment, It is first option that we need to check explicitly that the stage has moved to the requested position as it does not give any error when the stage fails to move.

src/odemis/driver/tescan.py
```diff
@@ -1096,6 +1107,60 @@ def _pollXYZ(self):
         except Exception:
             logging.exception("Unexpected failure during XYZ polling")
 
+    def _checkPosition(self, orig_pos: Dict[str, float], move: Dict[str, float], rel: bool = False,
+                       timeout: float = STAGE_WAIT_TIMEOUT):
+        """
+        Checks and waits for the current stage position to report the requested stage position given by pos.
+        :param orig_pos: original stage position before the stage movement in absolute coordinates
+        :param move: requested stage position in relative or absolute coordinates
+        :param rel: By default False, requested stage position is in absolution position otherwise relative
```
pieleric, line 1116: This function is *always* called with `rel=False`. => drop this argument, and indicate that it's in absolute coordinates all the time.
Also, change the name of `move` to `target_pos`. Then, you can drop the first 4 lines of the function.

pieleric (approved):

pieleric (changes requested):
.

pieleric (approved):

PR #2950, opened 2024-11-22
[fix] Make the beamshift correction save path unique

tepals:
With the volume EM changes the ROA name does not have to be unique anymore. Since the beamshift correction save path was only based on the ROA name, this caused all the images of a project to be saved in the same directory, and fields with the same id got overwritten. This fix extends the save path to include the project name, the roa name and the slice idx.

### DIFF
--- a/src/odemis/acq/fastem.py
+++ b/src/odemis/acq/fastem.py
@@ -281,10 +281,15 @@ class AcquisitionTask(object):
         # save the initial multibeam resolution, because the resolution will get updated if save_full_cells is True
         self._old_res = self._multibeam.resolution.value
 
-        path = fastem_util.create_image_dir("beam-shift-correction")
-        # The path should be [image-dir]/beam-shift-correction/[timestamp]/[roa-name]
-        self.path = os.path.join(path, self._roa.name.value)
-        os.makedirs(self.path, exist_ok=True)
+        beam_shift_path = fastem_util.create_image_dir("beam-shift-correction")
+        # If there is a project name the path will be
+        # [image-dir]/beam-shift-correction/[timestamp]/[project-name]/[roa-name]_[slice-idx]
+        # if there is no project name it will be
+        # [image-dir]/beam-shift-correction/[timestamp]/[roa-name]_[slice-idx]
+        self.beam_shift_path = os.path.join(beam_shift_path,
+                                            path if path else "",  # project-name or empty
+                                            f"{self._roa.name.value}_{self._roa.slice_index.value}")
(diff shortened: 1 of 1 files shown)


src/odemis/acq/fastem.py
```diff
@@ -281,10 +281,12 @@ def __init__(self, scanner, multibeam, descanner, detector, stage, scan_stage, c
...
         self._old_res = self._multibeam.resolution.value
 
-        path = fastem_util.create_image_dir("beam-shift-correction")
-        # The path should be [image-dir]/beam-shift-correction/[timestamp]/[roa-name]
-        self.path = os.path.join(path, self._roa.name.value)
-        os.makedirs(self.path, exist_ok=True)
+        beam_shift_path = fastem_util.create_image_dir("beam-shift-correction")
+        # The path should be [image-dir]/beam-shift-correction/[timestamp]/[project-name]/[roa-name]_[slice-idx]
+        self.beam_shift_path = os.path.join(beam_shift_path,
+                                            path,  # project-name
```
nandishjpatel, line 287: Some of the `odemis/acq/test/fastem_test.py` test cases fail because `path=None` :(

src/odemis/acq/fastem.py
```diff
@@ -281,10 +281,12 @@ def __init__(self, scanner, multibeam, descanner, detector, stage, scan_stage, c
         # save the initial multibeam resolution, because the resolution will get updated if save_full_cells is True
         self._old_res = self._multibeam.resolution.value
 
-        path = fastem_util.create_image_dir("beam-shift-correction")
-        # The path should be [image-dir]/beam-shift-correction/[timestamp]/[roa-name]
-        self.path = os.path.join(path, self._roa.name.value)
-        os.makedirs(self.path, exist_ok=True)
+        beam_shift_path = fastem_util.create_image_dir("beam-shift-correction")
+        # The path should be [image-dir]/beam-shift-correction/[timestamp]/[project-name]/[roa-name]_[slice-idx]
```
tmoerkerken, line 285: it seems that the `path` argument can be anything, so the [timestamp]/[project-name] is not a given. If it should always be of that format, maybe introduce a custom datatype and loosely enforce it via typing.

tmoerkerken (approved):

src/odemis/acq/fastem.py
tepals: It really should only be a string or None, so for now I went for the simpler fix: `path if path else "",`. Let me know if you do not agree with that.

PR #2951, opened 2024-11-25
[METEOR-1189][FIX] Prevent circular updates when loading overview data for meteor projects

patrickcleeve2:
Prevent circular updates and sync issues when loading overview data

### DIFF
--- a/src/odemis/gui/cont/correlation.py
+++ b/src/odemis/gui/cont/correlation.py
@@ -202,6 +202,8 @@ class CorrelationController(object):
             s.raw[0].metadata[model.MD_ROTATION_COR] = 0
             s.raw[0].metadata[model.MD_PIXEL_SIZE_COR] = (1, 1)
 
+            self.update_localization_tab_streams_metadata(s)
+
     def add_streams(self, streams: list) -> None:
         """add streams to the correlation tab
         :param streams: (list[StaticStream]) the streams to add"""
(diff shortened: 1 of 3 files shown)


src/odemis/gui/cont/tabs/cryo_chamber_tab.py
```diff
@@ -464,6 +467,11 @@ def _load_project_data(self, evt: wx.Event) -> bool:
         logging.debug(f"{len(proj_data['overviews'])} overviews loaded.")
         logging.debug(f"{len(localization_tab.tab_data_model.streams.value)} streams loaded.")
 
+        # re-subscribe to stream updates
+        # NOTE: this must be called with @call_in_wx_main to force the call to happen after existing stream addition events
+        localization_tab._start_streams_subscriber()
+        correlation_tab.correlation_controller._start_streams_subscriber()
```
pieleric, line 473: As these functions must be called delayed, just in this specific context, I'd suggest to change it to explicitly do it here, and *not* in the function definition. It'd be this way:
```suggestion
         wx.CallAfter(localization_tab._start_streams_subscriber)
        wx.CallAfter(correlation_tab.correlation_controller._start_streams_subscriber)
```
=> Then drop the `@call_in_wx_main` decorators on these functions (which is more logical, as the functions don't directly touch the GUI).

src/odemis/gui/cont/correlation.py
```diff
@@ -233,6 +235,13 @@ def add_streams(self, streams: list) -> None:
...
             sc.stream_panel.show_remove_btn(True)
 
+    def _stop_streams_subscriber(self):
+        self._tab_data_model.streams.unsubscribe(self.update_localization_tab_streams)
+
+    @call_in_wx_main
+    def _start_streams_subscriber(self):
+        self._tab_data_model.streams.subscribe(self.update_localization_tab_streams, init=False)
+
     def update_localization_tab_streams(self, streams: list) -> None:
```
pieleric, line 245: Most likely this function should be decorated by `@call_in_wx_main`. In particular, it calls `showOverviewStream()`, which states that it must be running in the GUI thread.

pieleric (changes requested):

src/odemis/gui/cont/tabs/cryo_chamber_tab.py
patrickcleeve2: thanks, didnt know about wx.CallAfter

pieleric (approved):

PR #2957, opened 2024-11-28
[TRCL-517] [fix] SPARC alignment of tunnel lens did not save the aligned position

pieleric:
After the user adjust the lens position, it must be stored, so that
other optical mode which explicitly request going to the FAV_POS_ACTIVE
actually go to that latest position, and not back to default value.

### DIFF
--- a/src/odemis/gui/cont/tabs/sparc2_align_tab.py
+++ b/src/odemis/gui/cont/tabs/sparc2_align_tab.py
@@ -912,6 +912,9 @@ class Sparc2AlignTab(Tab):
                 main.spec_switch.position.unsubscribe(self._onSpecSwitchPos)
             if main.light_aligner:
                 main.light_aligner.position.unsubscribe(self._onLightAlignPos)
+        if mode != "tunnel-lens-align":
+            if main.spec_ded_aligner:
+                main.spec_ded_aligner.position.unsubscribe(self._on_spec_ded_aligner_pos)
 
         # This is running in a separate thread (future). In most cases, no need to wait.
         op_mode = self._mode_to_opm[mode]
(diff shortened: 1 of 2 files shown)


src/odemis/gui/cont/tabs/sparc2_align_tab.py
```diff
@@ -2037,6 +2054,26 @@ def _on_spec_ded_aligner_init_move(self, f: Future) -> None:
...
+        Called when the spec-ded-aligner is moved (and the tunnel align mode is active),
+          for updating the MD_FAV_POS_ACTIVE
+        """
+        if self.tab_data_model.main.tab.value != self:
+            # Should never happen, but for safety, we double check
+            logging.warning("Received active spec-ded-aligner position while outside of alignment tab")
+            return
+
+        # Update the axis which are supposed to be recorded: the ones already in MD_FAV_POS_ACTIVE
+        aligner = self.tab_data_model.main.spec_ded_aligner
```
K4rishma, line 2068: According to the docstring, should the code not be in mode == "tunnel-lens-align":

K4rishma (approved):

src/odemis/gui/cont/tabs/sparc2_align_tab.py
pieleric: Yes, that's what `_on_tunnel_lens_align_done()` does: it activates this function (ie, subscribe to the position updates of the spec_ded_aligner), only when tunel-lens-align is active.

src/odemis/gui/cont/tabs/sparc2_align_tab.py
```diff
@@ -2037,6 +2054,26 @@ def _on_spec_ded_aligner_init_move(self, f: Future) -> None:
...
+        fib_fav_pos = aligner.getMetadata().get(model.MD_FAV_POS_ACTIVE, {})
```
tmoerkerken, line 2069: From the looks of it, there is now a possibility that `fib_fav_pos` is an empty dict. In this case the position is never set. Is that desired?

pieleric: Yes, correct. This allows to define which position is recorded by Odemis by specifying it in the microscope file. The current "spec-ded-aligner" has actually 3 axes. But we only want to save the Z position. By specifying only the Z pos in FAV_POS_ACTIVE in the microscope file, we can force Odemis to only save that position.

PR #2958, opened 2024-11-29
[fix] stack overflow python error for ProgressiveBatchFuture having large number of sub-futures

nandishjpatel:
make use of _is_cancel_triggered flag to cancel only once the ProgressiveBatchFuture and all its sub-futures when an exception occurs or CancelledError is raised for a sub-future

### DIFF
--- a/src/odemis/model/_futures.py
+++ b/src/odemis/model/_futures.py
@@ -512,12 +512,15 @@ class ProgressiveBatchFuture(ProgressiveFuture):
         self.futures = futures
         start = time.time()  # start time of the batch future
         super().__init__(start=start, end=start + sum(self.futures.values()))
-        self.set_running_or_notify_cancel()
         self.task_canceller = self._cancel_all  # takes care of cancelling the task (=all sub-futures)
+        self._is_cancel_triggered = False
 
         for f in self.futures:
             f.add_update_callback(self._on_future_update)  # called whenever set_progress of a sub-future is called
             f.add_done_callback(self._on_future_done)  # called when a sub-future is done
+
+        self.add_done_callback(self._on_batch_future_done)
+        self.set_running_or_notify_cancel()
 
     def _on_future_update(self, f, start, end):
         """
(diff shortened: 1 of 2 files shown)


src/odemis/model/_futures.py
```diff
@@ -570,6 +573,11 @@ def _on_future_done(self, f):
             # we don't require the futures to be carried out sequentially
             self.set_result(None)
 
+    def _on_batch_future_done(self, _):
+        """Called whenever the ProgressiveBatchFuture is finished."""
+        logging.debug("ProgressiveBatchFuture done.")
+        self._is_cancel_triggered = False
```
pieleric, line 579: Is that helpful? A Future can be run only once. Once it's finished, it cannot be cancelled anymore. So if it's done, there is no need to be able to run/cancel it again.
Or can you add a comment to explain why you do that?

src/odemis/model/_futures.py
```diff
@@ -512,13 +512,16 @@ def __init__(self, futures):
         self.futures = futures
         start = time.time()  # start time of the batch future
         super().__init__(start=start, end=start + sum(self.futures.values()))
-        self.set_running_or_notify_cancel()
         self.task_canceller = self._cancel_all  # takes care of cancelling the task (=all sub-futures)
+        self._is_cancel_triggered = False
```
pieleric, line 516: Add a comment explaining what it is used for.

src/odemis/model/_futures.py
```diff
@@ -548,18 +551,18 @@ def _on_future_done(self, f):
...
-                self.set_exception(ex)
-                return
-        except CancelledError:
-            if self.cancel():  # if cancelling works
-                return
-            else:
-                self.set_exception(CancelledError())  # if cancelling fails
+        # If an exception occurs or CancelledError is raised for a sub-future, cancel the ProgressiveBatchFuture and
+        # all its sub-futures
+        if not self._is_cancel_triggered:
```
pieleric, line 556: I think it would be clearer to have an explicit "short-cut" like that:
```
if self._is_cancel_triggered:
    return
```
And then add a comment explaining why this code is needed.

pieleric (changes requested):
Could you extend the commit message to explain what was happening, including a example traceback if possible (so that someone who encounters the issue can directly see it's the one that is fixed by this commit), and a short summary of what your code does to fix it.

src/odemis/model/_futures.py
nandishjpatel: Right, this is not needed

src/odemis/model/_futures.py
nandishjpatel: I will explain a bit more in the comment. The explicit "short-cut" cannot be used here because line 570 still needs to be called to later set the ProgressiveBatchFuture's result

pieleric (approved):

PR #2960, opened 2024-12-02
[fix] estimate acquisition time calculation

nandishjpatel:
make use of acq_dwell_time arg to correctly calulate the fieldscan time using getTotalFieldScanTime()

In the vEM GUI projects can have their own acquisition dwell time. This is now taken into account and the acquisition time is estimated correctly for dwell time slider change.

### DIFF
--- a/src/odemis/acq/fastem.py
+++ b/src/odemis/acq/fastem.py
@@ -26,6 +26,7 @@ import math
 import os
 import threading
 import time
+from typing import Optional
 from concurrent.futures import CancelledError
 
 import numpy
@@ -131,7 +132,7 @@ class FastEMROC(object):
         self.colour = colour
 
 
-def estimate_acquisition_time(roa, pre_calibrations=None):
+def estimate_acquisition_time(roa, pre_calibrations=None, acq_dwell_time: Optional[float] = None):
     """
     Computes the approximate time it will take to run the ROA (megafield) acquisition including pre-calibrations
     if specified.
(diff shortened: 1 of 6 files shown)


patrickcleeve2 (approved):

src/odemis/acq/fastem.py
```diff
@@ -281,6 +284,9 @@ def __init__(self, scanner, multibeam, descanner, detector, stage, scan_stage, c
         self._settings_obs = settings_obs
         self._spot_grid_thresh = spot_grid_thresh
         self._blank_beam = blank_beam
+        self._total_roa_time = 0
+        if isinstance(future, model.ProgressiveFuture):
```
patrickcleeve2, line 288: Is this ever not a ProgressiveFuture?

nandishjpatel: In the unittest cases the future can be `None` or `Mock()`, that's why I added this check

src/odemis/gui/comp/fastem_roa.py
```diff
@@ -151,12 +151,14 @@ def on_points(self, points):
...
-    def estimate_acquisition_time(self):
+    def estimate_acquisition_time(self, acq_dwell_time: Optional[float] = None):
         """
         Computes the approximate time it will take to run the ROA (megafield) acquisition.
+
+        :param acq_dwell_time: (float or None) The acquisition dwell time.
         :return (0 <= float): The estimated time for the ROA (megafield) acquisition in s.
         """
-        field_time = self._detector.frameDuration.value + 1.5  # there is about 1.5 seconds overhead per field
+        field_time = self._detector.getTotalFieldScanTime(acq_dwell_time) + 1.5  # there is about 1.5 seconds overhead per field
```
tepals, line 161: We should keep in mind that this 1.5s might be too much with our throughput improvements. I would keep it for now, but if it turns out that the estimate is too high, we can tune this.

tepals (approved):

PR #2963, opened 2024-12-04
[SDEV-1937] support importing greater than 2000 roas

nandishjpatel:
[perf] optimize ProjectManagerImportExport's _apply_project_data function to import projects data faster

The following high level changes have been made:
- Extend tab_data shapes list and set the update value without notifying to avoid unnecessary subscriber's callback
- Add child to a node without sorting; instead call tab_data.projects_tree.sort_children_recursively() in the end
- Separate creating/gathering all projects data and adding grid rows/children of all projects, because the later takes more time because it has functions which have wx.PostEvent triggering GUI changes
- Created a smart way of switching ribbons, sections, roas grid for projects, the projects grid data is now stored in a A dict where key is the project name and value is the project grid's data, the project's grid data again is a dict where key is the name of the grid's tab and value is the grid object

### DIFF
--- a/src/odemis/gui/comp/fastem_project_manager_panel.py
+++ b/src/odemis/gui/comp/fastem_project_manager_panel.py
@@ -25,7 +25,8 @@ import math
 import random
 from colorsys import hls_to_rgb, rgb_to_hls
 from functools import partial
-from typing import List, Tuple, Union
+from typing import List, Tuple, Union, Dict
+import time
 
 import wx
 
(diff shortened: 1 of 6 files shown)


src/odemis/gui/comp/fastem_project_manager_panel.py
```diff
@@ -123,6 +125,25 @@ def is_unique(hls_color):
             return new_color_hex
 
 
+@dataclass
+class ImportData:
```
pieleric, line 129: This name is really generic. I'd suggest to name it maybe like `ImportedDataRow`... but feel free to find a better name!

tepals (approved):

PR #2964, opened 2024-12-05
[fix] project pnl and btn enabling in current_project and is_acquiring va callbacks

nandishjpatel:
current_project va callback
- Enable the project related pnls based on len of current project
- Enable the project related btns based on len of current project and active project tab not being project settings tab

is_acquiring va callback
- Enable the project related pnls based on acquiring flag
- Enable the project related btns based on acquiring flag and active project tab not being project settings tab

### DIFF
--- a/src/odemis/gui/comp/fastem_project_manager_panel.py
+++ b/src/odemis/gui/comp/fastem_project_manager_panel.py
@@ -576,10 +576,11 @@ class FastEMProjectManagerPanel:
 
         :param is_acquiring: (bool) Flag indicating if acquisition is in progress.
         """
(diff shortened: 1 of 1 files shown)


patrickcleeve2 (approved):
Is there ever a chance the user would want "" as a project name? Would it be better to set it to None rather than empty to be more explicit that it's invalid?

nandishjpatel:
> Is there ever a chance the user would want "" as a project name? Would it be better to set it to None rather than empty to be more explicit that it's invalid?

The current_project is a `StringVA`, so None will not work. The user is not allowed to set "" as a project name

tepals (approved):

PR #2965, opened 2024-12-05
[fix] limit how quickly turnOnLight() acquires images

pieleric:
turnOnLight() would continuously acquire images until it detects light.
This is a little over the top, as it can take several seconds and it's
really not in a rush. If the CCD is set to short exposure times, that
could lead to 100 fps, which causes high CPU usage, with no advantage.

=> wait at least 100ms per frame, so at most it will be 10fps.

As a side effect, that will help some test cases which were setting
the exposure time to the minimum, and which relied on andorcam2. That
driver seems to have a memory leak when doing many DataFlow.get() in a
row.

Also simplify the code of autofocus_test.

### DIFF
--- a/src/odemis/acq/align/light.py
+++ b/src/odemis/acq/align/light.py
@@ -19,12 +19,14 @@ PARTICULAR  PURPOSE. See the GNU General Public License for more details.
 You should have received a copy of the GNU General Public License along with
 Odemis. If not, see http://www.gnu.org/licenses/.
 """
-
-from concurrent.futures._base import CancelledError, CANCELLED, FINISHED, RUNNING
-from odemis.model import CancellableFuture, logging
-import numpy
-from odemis.util import executeAsyncTask, img
 import threading
+import time
+from concurrent.futures._base import CancelledError, CANCELLED, FINISHED, RUNNING
+
+import numpy
+
+from odemis.model import CancellableFuture, logging
+from odemis.util import executeAsyncTask, img
 
 
 def turnOnLight(bl, ccd):
@@ -80,6 +82,7 @@ def _doTurnOnLight(f, bl, ccd):
         # Turn the light on, full power!
         bl.power.value = bl.power.range[1]
         while True:
+            time.sleep(0.1)
             img2 = ccd.data.get()
             try:
                 new_img = img.Subtract(img2, img_light_off)
--- a/src/odemis/acq/align/test/autofocus_test.py
+++ b/src/odemis/acq/align/test/autofocus_test.py
@@ -56,20 +56,9 @@ class TestAutofocus(unittest.TestCase):
     """
     Test autofocus functions
     """
-    backend_was_running = False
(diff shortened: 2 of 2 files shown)


src/odemis/acq/align/light.py
```diff
@@ -80,6 +80,7 @@ def _doTurnOnLight(f, bl, ccd):
         # Turn the light on, full power!
         bl.power.value = bl.power.range[1]
         while True:
+            time.sleep(0.1)
```
tmoerkerken, line 83: Have you verified this is enough to resolve the test on Ubuntu 20/Python 3.8?

src/odemis/acq/align/test/autofocus_test.py
```diff
@@ -56,20 +56,9 @@ class TestAutofocus(unittest.TestCase):
...
-        try:
-            testing.start_backend(SECOM_CONFIG)
-        except LookupError:
-            logging.info("A running backend is already found, skipping tests")
-            cls.backend_was_running = True
-            return
-        except IOError as exp:
-            logging.error(str(exp))
-            raise
+        testing.start_backend(SECOM_CONFIG)
```
tmoerkerken, line 61: This refactor is unrelated and not described in the PR description. Is this error handling superseded by something internal of start backend?

src/odemis/acq/align/light.py
```diff
@@ -19,7 +19,7 @@
 You should have received a copy of the GNU General Public License along with
 Odemis. If not, see http://www.gnu.org/licenses/.
 """
-
+import time
```
tepals, line 22: while you are at it, can you clean up the imports?
```
import threading
import time
from concurrent.futures._base import CancelledError, CANCELLED, FINISHED, RUNNING

import numpy

from odemis.model import CancellableFuture, logging
from odemis.util import executeAsyncTask, img
```

tepals (changes requested):

src/odemis/acq/align/test/autofocus_test.py
pieleric: It's the last line of the PR description. It's also explicitly in a separate commit.

src/odemis/acq/align/light.py
pieleric: I could reproduce the issue (memory leak) on my computer, and this solved it. Anyway, for now it's just a good "side-effect". I am working separately on actually fixing the memory leak (and making a proper test case that checks for that). So I don't care too much on whether that will fix really the test case (yet) on a specific Ubuntu.

src/odemis/acq/align/test/autofocus_test.py
tmoerkerken: Ah my bad! All good :heavy_check_mark:

tmoerkerken (approved):

tepals (approved):

PR #2971, opened 2024-12-09
[fix][METEOR-1256] raise error message if focus position is outside the range

K4rishma:


src/odemis/util/comp.py
```diff
@@ -141,10 +141,15 @@ def generate_zlevels(focuser: "Actuator", zrange: Tuple[float, float], zstep: fl
         raise KeyError(f"The focus actuator {focuser} does not have z axis")
 
     focuser_pos = focuser.position.value["z"]
+    focuser_rng = focuser.axes["z"].range
+    if (focuser_pos > focuser_rng[1]) or (focuser_pos < focuser_rng[0]):
```
pieleric, line 145: The more "pythonic" way to write it is:
```suggestion
    if not focuser_rng[0] <= focuser_pos <= focuser_rng[1]:
```

src/odemis/util/comp.py
```diff
@@ -141,10 +141,15 @@ def generate_zlevels(focuser: "Actuator", zrange: Tuple[float, float], zstep: fl
...
+        raise ValueError(f"The current focus position {focuser_pos} is outside focus range {focuser_rng}. "
```
pieleric, line 146: Actually, from what I understand, lines 162-163 already *try* to handle this case. It's actually done in a more soft way, as it still allows to acquire one image (instead of forbidding acquisition here).
However, there, it return a float, instead of a list with 1 float. 
=> Drop your extra changes, and adjust line 163.

src/odemis/util/test/comp_test.py
```diff
@@ -247,6 +247,22 @@ def test_clipping_axis_range(self):
         finally:
             self.focus.updateMetadata({model.MD_POS_ACTIVE_RANGE: {}})
 
+    def test_focus_outside_range(self):
+        # focus position above the absolute range
+        focus_range = self.focus.axes["z"].range
+        self.focus.moveAbsSync({"z": focus_range[1] + 100e-6})
```
pieleric, line 253: This is a good test, but confusing as to why it works. Normally the actuator should fail if a move is requested outside the range.
What happens is that the "focus" component hasn't been referenced. In such case, Odemis is a little loose in the checks, and allows going up to rng * 2. That's why this move is still allowed.
=> Please add this info in comment, to explain the magic of this code :-)

pieleric (changes requested):

pieleric (approved):

PR #2975, opened 2024-12-13
[SDEV-1946] fix issues related to sem stream panel

nandishjpatel:
[fix] issues related to SEM stream panel

Fixed:
- Allow cancelling of Auto brightness / contrast and SEM autofocus buttons
- Correctly enable/disable buttons and panels when an autofunction or an acquisition is running is running
- Pause the stream at the start of an autofunction and replay it once its finished running if the stream was playing before the start

### DIFF
--- a/src/odemis/gui/cont/tabs/fastem_setup_tab.py
+++ b/src/odemis/gui/cont/tabs/fastem_setup_tab.py
@@ -22,6 +22,7 @@ Odemis. If not, see http://www.gnu.org/licenses/.
 """
 import logging
 from concurrent.futures import CancelledError
+from typing import Optional
 
 import wx
 
@@ -38,7 +39,7 @@ from odemis.gui.cont.stream import FastEMStreamController
 from odemis.gui.cont.stream_bar import FastEMStreamsBarController
 from odemis.gui.cont.tabs.tab import Tab
 from odemis.gui.util import call_in_wx_main, wxlimit_invocation
-from odemis.model import getVAs
+from odemis.model import getVAs, ProgressiveFuture
 
 
 class FastEMSetupTab(Tab):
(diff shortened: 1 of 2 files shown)


src/odemis/gui/cont/tabs/fastem_setup_tab.py
```diff
@@ -275,39 +266,35 @@ def _on_optical_autofocus_done(self, future, _=None):
...
-
-        self.sem_stream_cont.stream_panel.Layout()
-        self.sem_stream_cont.stream_panel.Refresh()
-
-    @call_in_wx_main
     def _on_btn_sem_autofocus(self, _):
         # Disable other calibration buttons
         # TODO also disable btn_autostigmation once autostigmation is working
+        if self.tab_data.is_calibrating.value:
+            fastem._executor.cancel()
```
tepals, line 279: where is `fastem` defined?

nandishjpatel: Here: https://github.com/delmic/odemis/blob/dd5d4cf75cdcf5fddd188f19127536d4cbdd8525/src/odemis/acq/fastem.py#L54

pieleric (changes requested):

tepals (approved):

pieleric (changes requested):

src/odemis/gui/cont/tabs/fastem_setup_tab.py
nandishjpatel: The `align.fastem.align()` future's cancel does not actually cancel the current running calibration but rather subsequent calibrations if any. I need to address a `FIXME` [here](https://github.com/delmic/odemis/blob/7f351beb0183fd7845b6cff7d33e307af833538e/src/odemis/acq/align/fastem.py#L329) to solve the problem that's why I am cancelling the executor for now.

pieleric: Ok... so let's merge it as-is, and hopefully it'll not stick around too long. Please keep in mind to clean that part later.

pieleric (approved):

PR #2976, opened 2024-12-13
[fix] xt_client: shouldn't provide dwellTime when there is no detector

pieleric:
When xt_client is used as a "beam control" in addition to the external
scan board, it doesn't make sense to provide access to the (internal)
dwell time.
=> Only provide .dwellTime if there is a detector child.

### DIFF
--- a/src/odemis/driver/xt_client.py
+++ b/src/odemis/driver/xt_client.py
@@ -1361,15 +1361,6 @@ class Scanner(model.Emitter):
         self._hfw_nomag = hfw_nomag
         self._has_detector = has_detector
 
-        dwell_time_info = self.parent.dwell_time_info()
-        self.dwellTime = model.FloatContinuous(
-            self.parent.get_dwell_time(),
-            dwell_time_info["range"],
-            unit=dwell_time_info["unit"],
-            setter=self._setDwellTime)
-        # when the range has changed, clip the current dwell time value to the new range
-        self.dwellTime.clip_on_range = True
-
         voltage_info = self.parent.ht_voltage_info()
         init_voltage = numpy.clip(self.parent.get_ht_voltage(), voltage_info['range'][0], voltage_info['range'][1])
         self.accelVoltage = model.FloatContinuous(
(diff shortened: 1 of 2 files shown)


tepals (approved):

patrickcleeve2:
I don't really agree with this change here (or on the other client PR), there are times where you would want to control the dwell time independently of detector presence. For example, you can repeatedly scan the SEM to discharge the sample without acquiring, or for FIB milling.

src/odemis/driver/xt_client.py
```diff
@@ -1361,15 +1361,6 @@ def __init__(self, name: str, role: str, parent: model.HwComponent, hfw_nomag: f
         self._hfw_nomag = hfw_nomag
         self._has_detector = has_detector
 
-        dwell_time_info = self.parent.dwell_time_info()
```
patrickcleeve2, line 1364: https://github.com/delmic/odemis/pull/2976#issuecomment-2552670083

patrickcleeve2 (changes requested):

pieleric:
> I don't really agree with this change here (or on the other client PR), there are times where you would want to control the dwell time independently of detector presence. For example, you can repeatedly scan the SEM to discharge the sample without acquiring, or for FIB milling.

I agree that on a (FIB)SEM, it's some time useful to scan the beam without actually acquiring any data. 
However, right now, in Odemis, with the current backend API, it's not possible to request a scan without having a detector. The only way to scan is to call `data.get()` or `data.subscribe()`. So no detector means no scan. 
Technically, it's probably possible to directly call the xt-adapter functions by calling their wrapper on SEM component, but that's not how the back-end is supposed to be used. Clients should only use the backend API, not other functions on the component, and the SEM component has typically no role, because it's not supposed to be used directly.
The right way would be to extend the backend API. I think either we could introduce a "dummy" detector for "blind scanning", or have a DataFlow directly on the scanner component (called something like `nodata` or `scan`...).

patrickcleeve2 (approved):

PR #2977, opened 2024-12-13
[legal] adjust name from "Delmic Acqusition Software" to Odemis

pieleric:
The name "Delmic Acqusition Software" was used for just a few months, at
the very beginning of the project... but it somehow managed to sneak in.
=> Use Odemis everywhere.


tepals (approved):
I guess this got copied in because the very first version of the xt code was based on very old code from you :')

PR #2978, opened 2024-12-17
[fix] DigitalCamera depthOfField formula error

pieleric:
The code references a website with a formula, but that formula was
incorrectly copied: "mag - na", instead of the correct "mag * na".
For large, standard values, like mag = 60, na = 0.9, this has
surprisingly little effect, as the first part of the formula is the
main contributor. However, for values found on the SPARC like mag = 0.3,
this can have very strong effect. In particular, this could even lead to
impossible negative values.


tepals (approved):

tepals:
This should be tested on hardware before pushing it to the stable release.

PR #2982, opened 2024-12-17
[SDEV-1947] estimating overview image tiled acquisition time

nandishjpatel:
Fix includes:
- correctly estimating the stage movement time and adding it to the total time
- keeping the overhead time separate to overview acquisition time and stage time

### DIFF
--- a/src/odemis/acq/fastem.py
+++ b/src/odemis/acq/fastem.py
@@ -826,8 +826,7 @@ def estimateTiledAcquisitionTime(stream, stage, area, dwell_time=None):
 
     :param stream: (SEMstream) The stream used for the acquisition.
     :param stage: (actuator.MultiplexActuator) The stage in the sample carrier coordinate system.
-        The x and y axes are aligned with the x and y axes of the ebeam scanner. UNUSED: Can be used to take
-        speed of axes into account for time estimation.
+        The x and y axes are aligned with the x and y axes of the ebeam scanner.
     :param area: (float, float, float, float) xmin, ymin, xmax, ymax coordinates of the overview region
         in the sample carrier coordinate system.
     :param dwell_time: (float) A user input dwell time to be used instead of the stream emitter's dwell time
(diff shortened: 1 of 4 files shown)


src/odemis/acq/fastem.py
```diff
@@ -845,18 +844,39 @@ def estimateTiledAcquisitionTime(stream, stage, area, dwell_time=None):
...
+
+    # Vertical movement: Time for repositioning to the next row
+    time_y_per_move = (fov[1] / stage_speed_y)  # Moving vertically between rows
+    time_y = time_y_per_move * (ny - 1)  # Moving (ny - 1) times
+
+    # Total stage movement time
+    stage_time = time_x + time_y
+
+    # Add overhead per tile
+    overhead_per_tile = 2  # [s]
```
tepals, line 873: this feels like quite a lot, did you measure this? If you measured it, can you update the comment to include that that is how you ended up at 2 s?

tepals (changes requested):

src/odemis/acq/fastem.py
nandishjpatel: For `"resolution": (1024, 884)`, `"horizontalFoV": 2.0e-3,` `dwellTime: 100 ns`, `speed: 0.005, # m/s` below are the calculations:

`overview_acq_time = 5.07`
`stage_time = 21.62` 
`overhead = 112`

As you can see the image acquisition cannot be completed just by adding `overview_acq_time + stage_time`. I just found the `overhead_per_tile` number by doing an acquisition on the actual system which had the updated `PMD401Bus` driver.

Can I write the above as a comment?

nandishjpatel: Also we reference the stage before starting an acquisition

nandishjpatel: I think, I have a found a better fix, will discuss with you

nandishjpatel: @tepals the better fix can be found in the last two commits. I will test this on the system and let you know if it works!

nandishjpatel: @tepals I tested this PR changes on EA1 and it works fine!

src/odemis/acq/fastem.py
```diff
@@ -837,26 +836,54 @@ def estimateTiledAcquisitionTime(stream, stage, area, dwell_time=None):
...
-    acq_time_tile = res[0] * res[1] * dwell_time_value + overhead  # [s]
-    # time for total overview image
-    acq_time_overview = nx * ny * acq_time_tile  # [s]
+    acq_time_tile = res[0] * res[1] * dwell_time_value
 
-    return acq_time_overview
+    # Total acquisition time for imaging (all tiles)
+    # add 2s to account for switching from one tile to next tile
+    # this time is added in TiledAcquisitionTask.estimateTime
+    acq_time = nx * ny * (acq_time_tile + 2)
```
tepals, line 857: 2s of additional overhead on top of the stage movement still sounds very high, did you also check on EA1 that this is a realistic value? If it is, do you understand where this is coming from? Is it for instance all in the overhead from transferring the images from the MPC to the Odemis PC?

nandishjpatel: I checked on EA1 and this value is realistic. Honestly, I don't know where this overhead is coming from. There are a lot of futures and sub-futures in tiled acquisition. This 2s overhead apparently seems to work also for MIMAS and METEOR, it was added by @K4rishma in `TiledAcquisitionTask.estimateTime`

tepals: Okay, let's keep it in mind them, maybe we want to invest this at some point

tepals (approved):

pieleric (approved):

pieleric:
@nandishjpatel how should we merge this PR? There are 4 commits. I think the second one is a little bit apart (z-stack). The last commit, refactoring the STITCH_SPEED could stay apart too, but would need to be first. I think the other 2 commits should be squashed together.
Do you want to make the history fancy... or should I squash all into a single commit? (I think this second option is also fine as the changes are still limited to just a few files).

nandishjpatel:
> @nandishjpatel how should we merge this PR? There are 4 commits. I think the second one is a little bit apart (z-stack). The last commit, refactoring the STITCH_SPEED could stay apart too, but would need to be first. I think the other 2 commits should be squashed together. Do you want to make the history fancy... or should I squash all into a single commit? (I think this second option is also fine as the changes are still limited to just a few files).

I did make the history fancy as you mentioned :)

pieleric (approved):

PR #2986, opened 2024-12-20
[CSDM-1083] FPLM CL spot overlay

tmoerkerken:
In a SPARC setup in combination with the FPLM module, the alignment process consists of multiple stages that have some dependency on each other. In the “LIGHT IN” stage, where the PL spot is aligned, the location of the CL spot is essential. However, currently, the CL spot is only visible in the “LENS” mode. Currently, the aligned position of the CL spot is physically marked on the display (often with tape). This PR adds CL CCD snapshots to the light in modes, so that the manual application of tape is not needed anymore.

Changes:
- Add button for new LIGHT IN AR mode for FPLM
- Rename old button (LIGHT IN -> LIGHT IN SPOT for FPLM)
- CENTERING mode has light alignment buttons removed for FPLM
- LIGHT IN AR mode has “Light Aligner” controls, just like LIGHT IN (SPOT)
- LIGHT IN SPOT and LIGHT IN AR modes enforces auto beam blanking (if possible)
- LIGHT IN SPOT and LIGHT IN AR modes show a warning if beam blanking was unsuccessful
- Add Static2DUpdatableStream, to facilitate static updatable snapshots
- LIGHT IN SPOT mode has the CL spot (from LENS mode) as a red overlay that blends with the blue slit overlay
- LIGHT IN AR mode has the CL pattern (from CENTERING mode) as a red overlay that blends with the blue slit overlay

Changes after testing in Parma:
- Better layouting of alignment buttons in all situations (few or many buttons).
- Now utilizes an optical stream for the slit and CL overlay visualizations, so black is rendered transparant instead of solid.

Here the message that shows when blanking was not successful:
image

Here the slit, cl overlay and light signal in one view (LIGHT IN SPOT mode). Note that due to simulation, it does not look anything like what we would expect in real life (slit is too wide, light signal should be a spot instead of the parabolic mirror), but you can at least see the different colors:
image

# Tests & verification
Testing this feature programmatically is complicated, since it’s mostly GUI work. Therefore I thoroughly tested the features manually:
- **Manually check for CL stream overlay in LIGHT IN SPOT mode:**
Ensure that in the LIGHT IN SPOT mode, the CL spot stream from LENS mode is correctly overlaid as a red stream on top of the blue slit image, and both are displayed without affecting their visibility (i.e., overlapping pixels appear purple).

- **Manually check for pattern stream overlay in LIGHT IN AR mode:**
Ensure that in the LIGHT IN AR mode, the  CL patter stream from CENTERING mode is correctly overlaid as a red stream on top of the blue slit image, and both are displayed 
- **Manually check for automatic blanker control:**
Check that when entering LIGHT IN SPOT or LIGHT IN AR modes, the blanker is set to None, the automatic mode, (if a blanker is available) and that the blanker is automatically set back to False when exiting these modes.
- **Manually check the FSLM test case and compare it to master.** There should not be visual changes.
- **Manually check the ELIM test case and compare it to master.** There should not be visible changes.


src/odemis/gui/cont/tabs/sparc2_align_tab.py
```diff
@@ -205,44 +225,116 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
+            # Settings for the lens alignment ccd stream
+            self._setFullFoV(self._ccd_stream , (2, 2))
 
-            # To activate the SEM spot when the CCD plays
-            ccd_stream.should_update.subscribe(self._on_ccd_stream_play)
+            self._addMoIEntries(self._ccd_spe.stream_panel)
+            self._ccd_stream.image.subscribe(self._onNewMoI)
+
+            # Settings for centering align mode.
+            # Change binning to differentiate from other stream
```
tepals, line 292: You've now set both binnings to `(2, 2)` how can you differentiate?

src/odemis/gui/cont/tabs/sparc2_align_tab.py
```diff
@@ -205,44 +225,116 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
 
-            self._addMoIEntries(self._ccd_spe.stream_panel)
-            ccd_stream.image.subscribe(self._onNewMoI)
+            # Settings for the lens alignment ccd stream
+            self._setFullFoV(self._ccd_stream , (2, 2))
 
-            # To activate the SEM spot when the CCD plays
-            ccd_stream.should_update.subscribe(self._on_ccd_stream_play)
+            self._addMoIEntries(self._ccd_spe.stream_panel)
+            self._ccd_stream.image.subscribe(self._onNewMoI)
```
tepals, line 289: Is it correct that you use `self._ccd_spe` on line 288 and `self._ccd_stream` on line 289?

src/odemis/gui/cont/tabs/sparc2_align_tab.py
```diff
@@ -205,44 +225,116 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
+
+            # Settings for centering align mode.
+            # Change binning to differentiate from other stream
+            self._setFullFoV(self._ccd_stream_center, (2, 2))
+
+            if "light-in-align" in tab_data.align_mode.choices :
```
tepals, line 295:
```suggestion
            if "light-in-align" in tab_data.align_mode.choices:
```

src/odemis/gui/cont/tabs/sparc2_align_tab.py
```diff
@@ -2194,6 +2353,17 @@ def get_display_priority(cls, main_data):
 
         return None
 
+    def enforce_blanker(self):
+        if self._blanker:
+            # Enforce auto beam blanking when available
```
tepals, line 2358: It feels a bit strange that you are enforcing the blanker, but then setting it to automatic, instead of always blanking. Is that correct?

src/odemis/gui/model/tab_gui_data.py
```diff
@@ -747,6 +748,15 @@ def __init__(self, main):
             if "lens-align" in amodes:
                 amodes.remove("lens-align")
 
+        # Special setup for FPLM
+        if (
+            not main.spec_switch
+            and main.mirror
+            and main.mirror.name in main.light_aligner.affects.value
```
tepals, line 755: since this is the third time you do this check, does it make sense to have this in a function somewhere? Something like `check_is_fplm`?

tepals (changes requested):

src/odemis/gui/model/tab_gui_data.py
tmoerkerken: That is a good one, and I actually discussed this with Éric already a little bit at one point. I'd like to actually, but these are heuristics and not total proof that it is an FPLM module (similar to ELIM and SFLM). Maybe we can come up with a variable name that reflects this, or at least at a comment at instantiation.

Now that I look at this code, I see how it can fail when there is no light aligner available (and no mirror, although that would be surprising for a SPARC system), so I need to improve it anyway.

src/odemis/acq/stream/_static.py
```diff
@@ -155,17 +155,23 @@ def _clean_raw(self, raw):
...
         """
+        raw = self._clean_raw(raw)
+        super().__init__(name, raw, *args, **kwargs)
+        self._set_tint_from_md()
+        self.tint.subscribe(self._onTint)
+
+    def _clean_raw(self, raw):
+        """Makes sure the raw data is converted to a format that works for a StaticStream"""
+        if raw is None:
+            return raw
```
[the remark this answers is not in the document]
tmoerkerken: If `self.raw` indeed is a DataArray as documented, then we cannot do a truth check like that at this stage, we will get `ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()`. I will try to rewrite it/document it, so it becomes more clear.

src/odemis/gui/cont/tabs/sparc2_align_tab.py
tmoerkerken: I don't want to "mess" with the auto blanking behavior (assuming it works). If I set it to "True" then there is no automatic mode anymore.

src/odemis/gui/cont/tabs/sparc2_align_tab.py
tmoerkerken: I had that difference earlier for debugging purposes. In real life, the data should be a lot more distinct than on the simulator. The comment should go.

src/odemis/gui/cont/tabs/sparc2_align_tab.py
tmoerkerken: Yes, it's the same as before (with the difference that the _ccd_stream variable now is being stored on self). _ccd_spe is the stream panel that is updated when the detector data changes.

src/odemis/gui/cont/tabs/sparc2_align_tab.py
```diff
@@ -2194,6 +2344,17 @@ def get_display_priority(cls, main_data):
 
         return None
 
+    def enforce_blanker(self):
```
pieleric, line 2347: Add a docstring to explain what this does.

src/odemis/gui/cont/tabs/sparc2_align_tab.py
pieleric: I think this does what you want... but it's really not obvious by looking at the code.
What it does now is: 
* if the blanker doesn't support "automatic mode" => force e-beam blanker (that's easy to understand)
* if the blanker supports "automatic mode" (most of the systems do), then set it to automatic (which is typically already the current state). That will work *because* `_on_ccd_stream_play()` automatically disables the SEM spot stream when the 2 new align modes are selected (because the `_ccd_stream_light` stream is used, so the others streams are not active). 

This needs to be explained in a comment! I think also the docstring should explain that it'll "enforce" the blanker, assuming that the SEM stream is not playing. (in general, that's very important assumption).

src/odemis/gui/cont/tabs/sparc2_align_tab.py
```diff
@@ -1112,8 +1209,58 @@ def _onAlignMode(self, mode):
...
+                self.panel.pnl_spec_switch.Show(True)
+                self.panel.pnl_spec_switch.Enable(False)  # Wait until the spec-switch is engaged
+            else:
+                self.panel.pnl_spec_switch.Show(False)
+
+            self.panel.pnl_moi_settings.Show(False)
+            f.add_done_callback(self._on_light_in_align_done)
+        # This mode is only present for FPLM
+        elif mode == "light-in-align-ar":
+            self.enforce_blanker()
```
pieleric, line 1253: Probably in every other modes, you need to unblank the beam, if it was blanked here. (Otherwise, the beam will stay blank all all modes).

src/odemis/acq/path.py
```diff
@@ -368,7 +368,8 @@
 
 ALIGN_MODES = {'mirror-align', 'lens2-align', 'ek-align', 'chamber-view',
                'fiber-align', 'streak-align', 'spec-focus', 'spec-fiber-focus',
-               'streak-focus', 'light-in-align', 'tunnel-lens-align', 'spec-focus-ext'}
+               'streak-focus', 'light-in', 'light-in-align', 'light-in-align-ar',
```
pieleric, line 371: `'light-in'` doesn't seem to be used anywhere -> drop.
`'light-in-align-ar'` should be defined above in `SPARC2_MODES`. Otherwise, going to that mode will leave the optical path as-is (and you shouldn't rely on that what is the current state of the hardware).

pieleric, line 371: Please add a test case in `path_test` that goes to `light-in-align-ar` mode.

pieleric (changes requested):

src/odemis/gui/cont/tabs/sparc2_align_tab.py
```diff
@@ -205,44 +225,111 @@ def __init__(self, name, button, panel, main_frame, main_data):
         spot_stream.should_update.value = True
         self._spot_stream = spot_stream
 
+        self._blanker = None
```
pieleric, line 228: Please see line ~559 `if main_data.ebeam_blanker:` . If one was to read the code now, he'd be quite confused. Can you add a comment above this code block (line 559) that the "ebeam_blanker" and the `ebeam.blanker` are not the same thing. The ebeam_blanker is a dedicated hardware for fast blanking (and typically used for pulsed e-beam). The `ebeam.blanker` is just the standard (& slow) blanker.

src/odemis/acq/stream/_static.py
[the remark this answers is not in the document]
tmoerkerken: I tried to explain it better with typing, docstrings, comments and slight re-organization of the code.

src/odemis/acq/path.py
tmoerkerken: Removed light-in and light-in-align-ar. I reuse light-in-align now

tmoerkerken: Not necessary since I reuse existing path

tepals (changes requested):

src/odemis/gui/cont/tabs/sparc2_align_tab.py
```diff
@@ -747,20 +836,33 @@ def _layoutModeButtons(self):
         Positions the mode buttons in a nice way: on one line if they fit,
          otherwise on two lines.
         """
-        # If 3 buttons or less, keep them all on a single line, otherwise,
-        # spread on two lines. (Works up to 6 buttons)
         btns = list(self._alignbtn_to_mode.keys())
+        gb_sizer = btns[0].Parent.GetSizer().GetChildren()[0].GetSizer()
+        # Update button sizes when content changed
+        gb_sizer.Layout()
```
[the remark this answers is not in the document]
tmoerkerken: This one is not so obvious. Let me try to explain:
Just for the FPLM module, we change the text of the LIGHT IN mode to LIGHT IN SPOT at runtime (to differentiate it from LIGHT IN AR). The mode is identical to the original light in mode, so that is why I opted for not adding another separate align mode. 

But when changing text like that (after initial rendering), the size property reflects the old text. I need a Layout() call to get the button sizes for the updated text. In the end everything is positioned and I don't necessarily need another Layout() call.

src/odemis/gui/cont/tabs/sparc2_align_tab.py
```diff
@@ -2194,6 +2344,17 @@ def get_display_priority(cls, main_data):
...
+    def enforce_blanker(self):
+        if self._blanker:
+            # Enforce auto beam blanking when available
+            if None in getattr(self._blanker, "choices", {}):
+                self._blanker.value = None
+            else:
+                self._blanker.value = True
+        else:
+            self.panel.pnl_blanker_status.Show(True)
+            logging.warning('Make sure the e-beam is blanked manually')
```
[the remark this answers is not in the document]
tmoerkerken: Removed it

src/odemis/gui/cont/tabs/sparc2_align_tab.py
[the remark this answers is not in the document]
tmoerkerken: I added a comment to clarify

tepals (approved):

pieleric (approved):

src/odemis/gui/cont/tabs/sparc2_align_tab.py
```diff
@@ -405,7 +493,7 @@ def __init__(self, name, button, panel, main_frame, main_data):
                 # Add the stream to the stream bar controller
                 # so that it's displayed with the default 0.3 merge ratio
                 self._stream_controller.addStream(focus_stream, visible=False,
-                                                  add_to_view=False)
+                                                  add_to_view=True)
```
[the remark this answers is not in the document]
pieleric: looks good to me.

PR #2989, opened 2025-01-09
[fix] simulated stage could take very long doing an absolute move to 0

pieleric:
When asking a moveAbs() to position 0, which is usually the center of
the axis, the code would detect it's not a float, and concider it as a
"arbitrary move". If the speed was set to something "reasonable" like
1 mm/s, that would cause a 100 s move.

=> detect it's a arbitrary move by the presence of .choices

### DIFF
--- a/src/odemis/driver/simulated.py
+++ b/src/odemis/driver/simulated.py
@@ -392,14 +392,18 @@ class GenericComponent(model.Actuator):
     def _doMoveAbs(self, pos):
         maxtime = 0
         for axis, new_pos in pos.items():
-            if isinstance(new_pos, float):
+            if hasattr(self.axes[axis], "choices"):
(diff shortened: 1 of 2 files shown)


src/odemis/driver/simulated.py
```diff
@@ -382,8 +382,8 @@ def _doMoveRel(self, shift):
                                  (axis, self._position[axis] + change, rng))
 
             self._position[axis] += change
-            logging.info("moving axis %s to %f", axis, self._position[axis])
             maxtime = max(maxtime, abs(change) / self.speed.value[axis] + 0.001)
+            logging.info("moving axis %s to %f", axis, self._position[axis])
```
tmoerkerken, line 386: Seems like a debug remnant, since it does not seem to be functionally different

tmoerkerken (approved):

PR #2990, opened 2025-01-09
[fix] acq_test: adjust acqmng.acquire() API usage

pieleric:
One call wasn't updated to the (long time done) API change on acqmng.

### DIFF
--- a/src/odemis/acq/test/acq_test.py
+++ b/src/odemis/acq/test/acq_test.py
@@ -234,6 +234,7 @@ class SECOMTestCase(unittest.TestCase):
         st = stream.StreamTree(streams=[s1, s2])
         f = acqmng.acquire(st.getProjections(), settings_obs=settings_obs)
         data, e = f.result()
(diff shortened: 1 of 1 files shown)


src/odemis/acq/test/acq_test.py
```diff
@@ -339,7 +339,8 @@ def test_metadata(self):
         specs2.detExposureTime.value = 0.05
 
         f = acqmng.acquire([sps, sps2], settings_obs)
-        data = f.result()
+        data, exp = f.result()
+        self.assertIsNone(exp)
```
K4rishma, line 343: I think same for line 236,  self.assertIsNone(exp) is missing

K4rishma (approved):

src/odemis/acq/test/acq_test.py
pieleric: Good catch! I'll add the check there too.

PR #2998, opened 2025-01-22
[METEOR-960] FIBSEM GUI Preparation

patrickcleeve2:
Minor GUI modifications in preparation for FIBSEM controls and automated workflow

### DIFF
--- a/src/odemis/gui/conf/data.py
+++ b/src/odemis/gui/conf/data.py
@@ -167,7 +167,11 @@ HW_SETTINGS_CONFIG = {
                 "event": wx.EVT_SCROLL_CHANGED  # only affects when it's a slider
             }),
             ("probeCurrent", {
-                "event": wx.EVT_SCROLL_CHANGED  # only affects when it's a slider
+                "label": "Beam Current",
+                "control_type": odemis.gui.CONTROL_SLIDER,
+                "type": "float",
+                "scale": "linear",
+                "event": wx.EVT_SCROLL_CHANGED
             }),
             ("spotSize", {
                 "tooltip": "Electron-beam Spot size",
@@ -234,6 +238,46 @@ HW_SETTINGS_CONFIG = {
             ("scanner", {
                 "control_type": odemis.gui.CONTROL_NONE,
             }),
+        )),
+        "ion-beam":
+        OrderedDict((
+            ("accelVoltage", {
+                "label": "Accel. Voltage",
+                "tooltip": "Accelerating voltage",
+                "event": wx.EVT_SCROLL_CHANGED  # only affects when it's a slider
+            }),
+            ("probeCurrent", {
+                "label": "Beam Current",
+                "control_type": odemis.gui.CONTROL_SLIDER,
+                "type": "float",
+                "scale": "linear",
+                "event": wx.EVT_SCROLL_CHANGED
+            }),
+            ("resolution", {
+                "label": "Resolution",
+                "control_type": odemis.gui.CONTROL_COMBO,
+                "tooltip": "Number of pixels in the image",
+                "choices": None,
+                "accuracy": None,  # never simplify the numbers
+            }),
+            ("dwellTime", {
+                "control_type": odemis.gui.CONTROL_SLIDER,
+                "tooltip": "Pixel integration time",
+                "type": "float",
+                "accuracy": 3,
+                "event": wx.EVT_SCROLL_CHANGED
+            }),
+            ("horizontalFoV", {
+                "label": "HFW",
+                "tooltip": "Horizontal Field Width",
+                "control_type": odemis.gui.CONTROL_COMBO,
+                "choices": util.hfw_choices,
+            }),
+            ("scale", {
+                # same as binning (but accepts floats)
+                "control_type": odemis.gui.CONTROL_NONE,
+            }),
+
         )),
     "ebeam-blanker":
         OrderedDict((
@@ -579,10 +623,47 @@ HW_SETTINGS_CONFIG = {
         # Keep the same `contrast` and `brigtness` slider order as in the TFS UI
         OrderedDict((
             ("contrast", {
+                "label": "Contrast",
                 "control_type": odemis.gui.CONTROL_SLIDER,
+                "tooltip": "Contrast of the electron detector",
             }),
             ("brightness", {
+                "label": "Brightness",
                 "control_type": odemis.gui.CONTROL_SLIDER,
+                "tooltip": "Brightness of the electron detector",
+            }),
+            ("mode", {
+                "label": "Detector Mode",
+                "control_type": odemis.gui.CONTROL_COMBO,
+                "tooltip": "Mode of the electron detector",
+            }),
+            ("type", {
+                "label": "Detector Type",
+                "control_type": odemis.gui.CONTROL_COMBO,
+                "tooltip": "Type of the electron detector",
+            }),
+        )),
+    "se-detector-ion":
+        OrderedDict((
+            ("brightness", {
+                "label": "Brightness",
+                "control_type": odemis.gui.CONTROL_SLIDER,
+                "tooltip": "Brightness of the ion detector",
+            }),
+            ("contrast", {
+                "label": "Contrast",
+                "control_type": odemis.gui.CONTROL_SLIDER,
+                "tooltip": "Contrast of the ion detector",
+            }),
+            ("mode", {
+                "label": "Detector Mode",
+                "control_type": odemis.gui.CONTROL_COMBO,
+                "tooltip": "Mode of the ion detector",
+            }),
+            ("type", {
+                "label": "Detector Type",
+                "control_type": odemis.gui.CONTROL_COMBO,
+                "tooltip": "Type of the ion detector",
             }),
         )),
 }
@@ -761,6 +842,20 @@ HW_SETTINGS_CONFIG_PER_ROLE = {
             },
         },
     },
+    "meteor" : {
+        "e-beam": {
+            "scale": {
+                "control_type": odemis.gui.CONTROL_NONE,
+            },
+            "resolution": {
+                "label": "Resolution",
+                "control_type": odemis.gui.CONTROL_COMBO,
+                "tooltip": "Number of pixels in the image",
+                "choices": None,
+                "accuracy": None,  # never simplify the numbers
+            }
+        },
+    }
 }
 
 # The sparc-simplex is identical to the sparc
--- a/src/odemis/gui/cont/stream_bar.py
+++ b/src/odemis/gui/cont/stream_bar.py
@@ -25,9 +25,9 @@ import functools
 import gc
 import logging
 import os
-from collections import OrderedDict
 import threading
 import time
+from collections import OrderedDict
 
 import wx
 
@@ -35,12 +35,13 @@ import odemis.acq.stream as acqstream
 import odemis.gui.conf.file
 import odemis.gui.model as guimodel
 from odemis import model
+from odemis.acq.stream import FastEMOverviewStream, StaticSEMStream, StaticStream
 from odemis.acq.stream_settings import StreamSettingsConfig
-from odemis.acq.stream import StaticStream, FastEMOverviewStream
 from odemis.gui.conf.data import get_local_vas
 from odemis.gui.cont.stream import StreamController
 from odemis.gui.model import TOOL_NONE, TOOL_SPOT
 from odemis.gui.util import call_in_wx_main
+from odemis.util.dataio import data_to_static_streams
 
 # There are two kinds of controllers:
 # * Stream controller: links 1 stream <-> stream panel (cont/stream/StreamPanel)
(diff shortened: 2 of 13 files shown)


src/odemis/gui/model/tab_gui_data.py
```diff
@@ -386,6 +397,49 @@ def _on_project_path_change(self, _):
...
+            config.pj_last_path, config.fn_ptn,
+            config.last_extension,
+            config.fn_count))
+        self.main.project_path.subscribe(self._on_project_path_change)
+
+        # milling patterns
+        self.patterns = model.ListVA()
+
+        self.view_posture = model.VigilantAttribute(SEM_IMAGING)
+        self.acqui_mode = AcquiMode.FIBSEM
```
pieleric, line 430: No, don't introduce such "mode" for distinguishing between the two tabs.  I've looked at the usage of this "constant", and if I understand correctly, this is to do some adjustment in classes depending on which tab they are used. The problem is that it makes the behaviour tricky to guess. The code creating the object looks identical everywhere, but the object behave differently. I think it would be much clearer to have explicit code that take this flag as input parameter.

For instance, now you have:
`feature_panel_controller = CryoFeatureController(tab_data, panel, self)`
It would be instead:
`feature_panel_controller = CryoFeatureController(tab_data, panel, self, mode=AcquisitionMode.FIBSEM)`
or
`feature_panel_controller = CryoFeatureController(tab_data, panel, self, mode=AcquisitionMode.FLM)`

src/odemis/gui/model/tab_gui_data.py
```diff
@@ -386,6 +397,49 @@ def _on_project_path_change(self, _):
...
+        self.is_sem_active_view: bool = False
+        self.is_fib_active_view: bool = False
```
pieleric, line 432: They seems to detect which view is active. You can already do that with `self.focussedView.value is self.panel.vp_secom_tl.view`. => drop.

src/odemis/gui/model/tab_gui_data.py
```diff
@@ -66,6 +69,10 @@
 from odemis.util.filename import create_filename, make_unique_name
 
 
+class AcquiMode(Enum): # TODO: move to better place
```
pieleric, line 72: I feel it's a very generic name for a very specific purpose... but I don't have really better proposal. Could you at least use a full words `AcquisitionMode`. 
Also, the right place for it should be in gui/model/_constant.py

An alternative could be to just use `model.MD_AT_FLUO` and `model.MD_AT_FIB`.

src/odemis/gui/xmlh/resources/panel_tab_localization.xrc
```diff
@@ -314,7 +314,7 @@
                             <object class="sizeritem">
                               <object class="wxBoxSizer">
                                 <object class="sizeritem">
-                                  <object class="wxStaticText">
+                                  <object class="wxStaticText" name="label_feature_z">
```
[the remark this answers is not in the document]
patrickcleeve2: yes it is accessed later

src/odemis/gui/model/tab_gui_data.py
patrickcleeve2: when this is accessed, you don't necessarily have direct access to the tab or panel, so it's clearer to have the property

src/odemis/gui/model/tab_gui_data.py
patrickcleeve2: ok, removed from guidata

src/odemis/gui/conf/data.py
```diff
@@ -235,6 +239,52 @@
...
+                "control_type": odemis.gui.CONTROL_COMBO,
+                "tooltip": "Number of pixels in the image",
+                "choices": None,
+                "accuracy": None,  # never simplify the numbers
+            }),
+            ("dwellTime", {
+                "control_type": odemis.gui.CONTROL_SLIDER,
+                "tooltip": "Pixel integration time",
+                # "range": (1e-9, 1),
+                # "scale": "log",
```
tepals, line 267: why is this commented?

src/odemis/gui/conf/data.py
```diff
@@ -235,6 +239,52 @@
...
+                "control_type": odemis.gui.CONTROL_SLIDER,
+                "type": "float",
+                "scale": "linear",
+                "event": wx.EVT_SCROLL_CHANGED
+            }),
+            ("resolution", {
+                "label": "Resolution",
+                "control_type": odemis.gui.CONTROL_COMBO,
+                "tooltip": "Number of pixels in the image",
+                "choices": None,
```
tepals, line 260: for the e-beam the resolution choices are `"choices": util.resolution_from_range,`, why not have that here?

src/odemis/gui/conf/data.py
```diff
@@ -235,6 +239,52 @@
...
+                "type": "float",
+                "accuracy": 3,
+                "event": wx.EVT_SCROLL_CHANGED
+            }),
+            ("horizontalFoV", {
+                "label": "HFW",
+                "tooltip": "Horizontal Field Width",
+                "control_type": odemis.gui.CONTROL_COMBO,
+                "choices": util.hfw_choices,
+                # "accuracy": 3,
```
tepals, line 277: same here, why is this commented?

src/odemis/gui/conf/data.py
```diff
@@ -235,6 +239,52 @@
...
+            }),
+            ("scale", {
+                # same as binning (but accepts floats)
+                "control_type": odemis.gui.CONTROL_NONE,
+                # "tooltip": "Pixel resolution preset",
+                # means will make sure both dimensions are treated as one
+                # "choices": util.binning_1d_from_2d,
```
tepals, line 284: same here, why is this commented?

src/odemis/gui/conf/data.py
```diff
@@ -594,6 +644,27 @@
...
+                "label": "Detector Type",
+            }),
+        )),
+    "se-detector-ion":
+        OrderedDict((
+            ("brightness", {
+                "label": "Brightness",
+            }),
+            ("contrast", {
+                "label": "Contrast",
```
tepals, line 660: is it not necessary to have a control type here?

src/odemis/gui/cont/stream_bar.py
```diff
@@ -2185,3 +2186,97 @@ def clear(self, clear_model=True):
...
+        self._stream_bar.fit_streams()
+
+        # Force a check of what can be garbage collected, as some of the streams
+        # could be quite big, that will help to reduce memory pressure.
+        gc.collect()
+
+    def clear(self, clear_model=True):
+        """
+        Remove all the streams, from the GUI (view, stream panels)
+        Must be called in the main GUI thread.
```
tepals, line 2264: I thought if something should be called in the main GUI thread, you should decorate it with `@call_in_wx_main`. But I see this comment comes back in some other places in Odemis and the decorator is not used in any of those places. Do you know why that's the case?

src/odemis/gui/cont/stream_bar.py
```diff
@@ -2185,3 +2186,97 @@ def clear(self, clear_model=True):
...
+        :param clear_model: unused, but required because of external api
```
tepals, line 2265: "external api" makes it sound like the api is external to Odemis, but this is part of the `StreamBarController` api, right?

src/odemis/gui/model/tab_gui_data.py
```diff
@@ -488,7 +532,12 @@ def __init__(self, main):
 
         self.stage_align_slider_va = model.FloatVA(1e-6)
         self.show_advaned = model.BooleanVA(False)
+        self.view_posture = VigilantAttribute(FM_IMAGING)
+        # self.main.posture_manager.current_posture.subscribe(self._on_posture_change, init=True) # TODO: enable once new pm is merged
```
tepals, line 536: what is "new pm"?

src/odemis/gui/win/acquisition.py
```diff
@@ -1405,18 +1405,49 @@ def ShowChamberFileDialog(parent, projectname):
...
+    message: str = "Choose a project directory to load",
+) -> Optional[str]:
     """
-    parent (wxframe): parent window
-    projectname (string): project name to propose by default
-    return (string or none): the project directory name to load (or the none if the user cancelled)
+    :param parent (wx.Frame): parent window
+    :param projectname (string): project name to propose by default
+    :param message (string): message to display in the dialog
+    :return (string or none): the project directory name to load (or the none if the user cancelled)
```
tepals, line 1417:
```suggestion
    :return (string or None): the project directory name to load (or None if the user cancelled)
```

src/odemis/gui/win/acquisition.py
```diff
@@ -1405,18 +1405,49 @@ def ShowChamberFileDialog(parent, projectname):
...
+        message=message,
+        defaultPath=projectname,
+        style=wx.DD_DEFAULT_STYLE | wx.DD_DIR_MUST_EXIST,
+    )
+
+    # Show the dialog and check whether is was accepted or cancelled
+    if dialog.ShowModal() != wx.ID_OK:
+        return None
+
+    # project path have been selected...
```
tepals, line 1431:
```suggestion
    # project path that has been selected...
```

src/odemis/gui/win/acquisition.py
```diff
@@ -1405,18 +1405,49 @@ def ShowChamberFileDialog(parent, projectname):
...
+def SelectFileDialog(
+    parent: wx.Frame,
+    message: str,
+    default_path: str,
+) -> Optional[str]:
+    """
+    :param parent (wx.Frame): parent window
+    :param message (string): message to display in the dialog
+    :param default_path (string): default path to open the dialog
+    :return (string or none): the selected file name (or the none if the user cancelled)
```
tepals, line 1443:
```suggestion
    :return (string or None): the selected file name (or None if the user cancelled)
```

tepals (changes requested):
I added a review, because Nandish will still be on holiday for 3 more weeks

src/odemis/gui/conf/data.py
patrickcleeve2: this resolution is set directly, so it has choices. so it can be used directly. whereas i think the other resolutions are set indirectly from scale / pixelsize

src/odemis/gui/model/tab_gui_data.py
patrickcleeve2: https://github.com/delmic/odemis/pull/3005

src/odemis/gui/conf/data.py
patrickcleeve2: updated, and added tooltips

src/odemis/gui/cont/stream_bar.py
patrickcleeve2: yeah, updated it

src/odemis/gui/cont/stream_bar.py
patrickcleeve2: I'm not sure either, but this it's the same in the above CryoAcquiredStreamController https://github.com/patrickcleeve2/odemis/blob/fbbd837d1c44c45c45ae077ac1b280320349668e/src/odemis/gui/cont/stream_bar.py#L2161 and in base StreamController: https://github.com/patrickcleeve2/odemis/blob/c1bb04bf7d87ab1e7f28d06a4a9a0fb12d207918/src/odemis/gui/cont/stream_bar.py#L900
My assumption is that the comment isn't actually true, or whatever calls it has the decorator.

pieleric: If it says "Must be called in the main GUI thread" that means the function is *not* protected (by the decorator). So the *caller* has to either be certain to be running in the main thread (ie, it's a GUI callback, or it's decorated), or it should call this function with `wx.CallLater()`.

It's useful to do it this way because the `@call_in_wx_main` decorator causes the call to be asynchronous, which means, it's slow, the control flow is hard to grasp, and especially it's not possible to get the return value.

src/odemis/gui/model/tab_gui_data.py
pieleric: The idea of the classes in the gui.model is that represent the state of the GUI. Their attributes should be "observable" (ie, they are VigilantAttributes) and information should be defined only once. 

So I'd recommend to move these out of the model.CryoFIBSEMGUIData . One alternative to my previous suggestion is to use the stream_classes of the view (which are SEMStream and FIBStream on the respective views). Something like:
`self.focussedView.value.is_compatible(SEMStream)` or `self.focussedView.value.is_compatible(FIBStream)`

pieleric (changes requested):
.

src/odemis/gui/model/tab_gui_data.py
patrickcleeve2: ok removed

src/odemis/gui/conf/data.py
tepals: makes sense

tepals (approved):

pieleric (approved):

PR #2999, opened 2025-01-22
[feat][METEOR-1128] TDCT Odemis Wrapper

patrickcleeve2:
Implementation of the 3DCT wrapper to allow running multi-point correlation from odemis. Includes a parser to read correlated position from file, and z-targeting utility for multi-channel images.

3DCT package now only has two dependencies for the core workflow (excluding user interface and image readers), which are already included in odemis (numpy, scipy). Packaging will be done later. [Source Repository](https://github.com/patrickcleeve2/3DCT/tree/refactor)

Subtasks: 
- [METEOR-1128](https://delmic.atlassian.net/browse/METEOR-1128?atlOrigin=eyJpIjoiZDU4MWM2ZGJiZjhjNDMxODhlZWU5NzZjM2IxZmNlYzQiLCJwIjoiaiJ9)
- [METEOR-1251](https://delmic.atlassian.net/browse/METEOR-1251?atlOrigin=eyJpIjoiNWZkYmUwNGNkZGZjNDlkMWFiMTk0NmQ5Zjk0NGExZGIiLCJwIjoiaiJ9)

Specifications:
- [Multi-Point Correlation](https://docs.google.com/document/d/1jBcJvvXyR41V0lCOaPBpqwRypYJ5xOTpkHsQYWiQa4c/edit?tab=t.0#heading=h.r7i4xprpd0oz)
- [FIB-FLM Correlation](https://docs.google.com/document/d/1k22BXNWQtyey1gxFHYHUkTC-5YmGuukFm7XAP_igSfw/edit?tab=t.0)

[METEOR-1128]: https://delmic.atlassian.net/browse/METEOR-1128?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[METEOR-1251]: https://delmic.atlassian.net/browse/METEOR-1251?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/odemis/acq/align/tdct.py
+++ b/src/odemis/acq/align/tdct.py
@@ -1,0 +1,193 @@
+"""
+Created on 8 Jan 2025
+
+Copyright © 2025 Delmic
+
+This file is part of Odemis.
+
+Odemis is free software: you can redistribute it and/or modify it under the
+terms  of the GNU General Public License version 2 as published by the Free
+Software  Foundation.
+
+Odemis is distributed in the hope that it will be useful, but WITHOUT ANY
+WARRANTY;  without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+PARTICULAR  PURPOSE. See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with
+Odemis. If not, see http://www.gnu.org/licenses/.
+"""
+
+import os
+import logging
+import sys
+from typing import Dict, List, Tuple, Any
+
+import numpy
+import yaml
+
+from odemis import model
+
+# install from: https://github.com/patrickcleeve2/3DCT/blob/refactor
+sys.path.append(f"{os.path.expanduser('~')}/development/3DCT")
+
+from tdct.correlation_v2 import run_correlation
+from tdct.util import multi_channel_get_z_guass
+
+def _convert_das_to_numpy_stack(das: List[model.DataArray]) -> numpy.ndarray:
+    """Convert a list of DataArrays to a numpy stack.
+    Channels are stored as list dimensions, rather than data array dimensions.
+    Therefore, multi-channel images are stored as list[CTZYX, CTZYX, ...] where C=1
+    and length of list is number of channels.
+    :param das: list of meteor data arrays (supports 5D CTZYX, 3D ZYX, 2D YX arrays)
+    :return the data arrays reshapes to a 4D numpy array (CZYX)"""
+    arr = []
+    for da in das:
+        if isinstance(da, model.DataArrayShadow):
+            da = da.getData()
+
+        # convert to 3D ZYX
+        if da.ndim == 5:
+            if da.shape[0] != 1 or da.shape[1] != 1:
+                logging.warning(f"Only the first channel and time dimension will be used for 5D data array: {da.shape}")
+            # remove the channel, time dimensions
+            da = da[0, 0, :, :, :]
+        elif da.ndim == 2:
+            # expand to 3D ZYX
+            da = numpy.expand_dims(da, axis=0)
+
+        assert da.ndim == 3, f"DataArray must be 3D ZYX, but is {da.shape}"
+        arr.append(da)
+
+    return numpy.stack(arr, axis=0)
+
+def get_optimized_z_gauss(das: List[model.DataArray], x: int, y: int, z: int, show: bool = False) -> float:
+    """Get the best fitting z-coordinate for the given x, y coordinates. Supports multi-channel images.
+    :param das: the data arrays (CTZYX, ZYX, or YX), all arrays must have the same shape
+    :param x: the x-coordinate
+    :param y: the y-coordinate
+    :param z: the z-coordinate (initial guess)
+    :param show: show the plot for debugging
+    :return: the z-coordinate (optimized)"""
+    prev_z = z
+    prev_x, prev_y = x, y
+
+    # fm_image  must be 4D np.ndarray with shape (channels, z, y, x)
+    fm_image = _convert_das_to_numpy_stack(das)
+
+    try:
+        # getzGauss can fail, so we need to catch the exception
+        zval, z, _ = multi_channel_get_z_guass(image=fm_image, x=x, y=y, show=show)
+        logging.info(f"Using Z-Gauss optimisation: {z}, previous z: {prev_z}")
+
+    except RuntimeError as e:
+        logging.warning(f"Error in z-gauss optimisation: {e}, using previous z: {prev_z}")
+        z = prev_z
+        x, y = prev_x, prev_y
+
+    return z
+
+def run_tdct_correlation(fib_coords: numpy.ndarray,
+                           fm_coords: numpy.ndarray,
+                           poi_coords: numpy.ndarray,
+                           fib_image: model.DataArray,
+                           fm_image: model.DataArray,
+                           path: str) -> Dict[str, Any]:
+    """Run 3DCT Multi-point correlation between FIB and FM images.
+    :param fib_coords: the FIB coordinates (n, (x, y)) (in pixels, origin at top left)
+    :param fm_coords: the FM coordinates (n, (x, y, z)) (in pixels, origin at top left)
+    :param poi_coords: the point of interest coordinates (1, (x, y, z)). Expects only one point of interest.
+    :param fib_image: the FIB image (YX)
+    :param fm_image: the FM image (CTZTX, CZYX or ZYX)
+    :param path: the path to save the results
+    :return: the correlation results
+        output:
+            error:
+                delta_2d: reprojection error between 3D and 2D coordinates
+                reprojected_3d: 3D coordinates reprojected to 2D
+                mean_absolute_error: mean absolute error of the transformation (x, y)
+                rms_error: root mean square error of the transformation
+            poi: list of transformed point of interest coordinates
+                image_px: coordinates in image pixels (0, 0 top left)
+                px:  coordinates in microscope image pixels (0, 0 image center)
+                px_um: coordinates in microscope image meters (0, 0 image center)
+            transformation:
+                rotation_eulers: transformation rotation (eulers)
+                rotation_quaternion: transformation rotation (quaternion)
+                scale: transformation scale
+                translation_around_rotation_center: transformation translation
+    """
+
+    # fib coordinates need to be x, y, z for 3DCT
+    if fib_coords.shape[-1] == 2:
+        fib_coords = numpy.column_stack((fib_coords, numpy.zeros(fib_coords.shape[0])))
+
+    # coordinates need to be float32 for 3DCT
+    fib_coords = fib_coords.astype(numpy.float32)
+    fm_coords = fm_coords.astype(numpy.float32)
+
+    # get first channel only, assume all channels are the same shape
+    if fm_image.ndim == 4:
+        fm_image = fm_image[0, :, :, :]
+    if fm_image.ndim == 5:
+        fm_image = fm_image[0, 0, :, :, :]
+
+    # get rotation center
+    halfmax_dim = int(max(fm_image.shape) * 0.5)
+    rotation_center = (halfmax_dim, halfmax_dim, halfmax_dim)
+
+    # get fib pixel size (meters)
+    fib_pixel_size = fib_image.metadata[model.MD_PIXEL_SIZE][0]
+
+    # fib image shape minus metadata, fib_pixelsize (microns), fm_image_shape
+    image_props = [fib_image.shape, fib_pixel_size * 1e6, fm_image.shape]
+
+    assert fm_coords.dtype == numpy.float32, "FM coordinates must be float32"
+    assert fib_coords.dtype == numpy.float32, "FIB coordinates must be float32"
+    assert fm_coords.shape[-1] == 3, "FM coordinates must be 3D (x, y, z)"
+    assert fib_coords.shape[-1] == 3, "FIB coordinates must be 3D (x, y, z)"
+    assert fib_coords.shape == fm_coords.shape, "FIB and FM coordinates must have the same shape"
+    assert fib_image.ndim == 2, "FIB Image must be 2D"
+    assert fm_image.ndim == 3, "FM Image must be 3D"
+    assert fib_pixel_size is not None, "FIB Pixel Size must be set"
+
+    logging.debug(
+        f"Running 3DCT correlation with FIB image shape: {fib_image.shape}, FM image shape: {fm_image.shape}"
+    )
+
+    # run correlation
+    correlation_results = run_correlation(
+        fib_coords=fib_coords,
+        fm_coords=fm_coords,
+        poi_coords=poi_coords,
+        image_props=image_props,
+        rotation_center=rotation_center,
+        path=path,
+    )
+
+    return correlation_results
+
+def get_reprojected_poi_coordinate(correlation_results: dict) -> Tuple[float, float]:
+    """Get the the point of interest coordinate from correlation data
+    and convert from micrometers to meters in the microscope image coordinate system.
+    The coordinate is centred at the image centre (x+ -> right, y+ -> up).
+    :param correlation_results: the correlation results
+    :return: the point of interest coordinate in meters
+    """
+    # get the point of interest coordinate (in microscope coordinates, in metres)
+    poi_coord = correlation_results["output"]["poi"][0]["px_um"]
+    poi_coord = (poi_coord[0] * 1e-6, poi_coord[1] * 1e-6)
+    return poi_coord
+
+def parse_3dct_yaml_file(path: str) -> Tuple[float, float]:
+    """Parse the 3DCT yaml file and extract the point of interest (POI)
+    in microscope image coordinates (um). Convert the coordinates to metres.
+    Note: only the first POI is extracted.
+    :param path: path to the 3DCT yaml file.
+    :return: The point of interest in microscope image coordinates (metres, centred at the image centre).
+    """
+    with open(path, "r") as f:
+        data = yaml.safe_load(f)
+
+        pt = get_reprojected_poi_coordinate(data["correlation"])
+
+    return pt
--- a/src/odemis/acq/align/test/tdct_test.py
+++ b/src/odemis/acq/align/test/tdct_test.py
@@ -1,0 +1,113 @@
+# -*- coding: utf-8 -*-
+"""
+Created on 12 Feb 2025
+
+Copyright © Delmic
+
+This file is part of Odemis.
+
+Odemis is free software: you can redistribute it and/or modify it under the terms
+of the GNU General Public License version 2 as published by the Free Software
+Foundation.
+
+Odemis is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+PURPOSE. See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with
+Odemis. If not, see http://www.gnu.org/licenses/.
+"""
+import os
+import unittest
+
+import numpy
+from odemis import model
+from odemis.acq.align import tdct
+
+RESULTS_PATH = os.path.join(os.getcwd(), "correlation_data.yaml")
+
+class TestTDCT(unittest.TestCase):
+
+    def tearDown(self):
+        if os.path.exists(RESULTS_PATH):
+            os.remove(RESULTS_PATH)
+
+    def test_convert_das_to_numpy_stack(self):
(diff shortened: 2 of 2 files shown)


src/odemis/acq/align/tdct.py
```diff
@@ -0,0 +1,146 @@
...
+def run_tdct_correlation(fib_coords: numpy.ndarray,
+                           fm_coords: numpy.ndarray,
+                           poi_coords: numpy.ndarray,
+                           fib_image: model.DataArray,
+                           fm_image: model.DataArray,
+                           path: str) -> Dict[str, Tuple[float, float]]:
+    """Run 3DCT Multi-point correlation between FIB and FM images.
+    :param fib_coords: the FIB coordinates (x, y, z)
+    :param fm_coords: the FM coordinates (x, y, z)
+    :param poi_coords: the point of interest coordinates
```
pieleric, line 83: Please specify shape.

src/odemis/acq/align/tdct.py
```diff
@@ -0,0 +1,146 @@
...
+        fm_image_filename="FM-Image",
+    )
+
+    return correlation_results
+
+def get_poi_coordinate(correlation_results: dict) -> Tuple[float, float]:
+    """Get the the point of interest coordinate from correlation data
+    and convert from micrometers to meters in the microscope image coordinate system.
+    :param correlation_results: the correlation results
+    :return: the point of interest coordinate in meters
```
pieleric, line 129: Trying to connect the dots, it seems that you can use this function on the result returned by `run_tdct_correlation()`. Is that correct? What's the origin of the coordinate system? top-left of the image?

src/odemis/acq/align/tdct.py
```diff
@@ -0,0 +1,146 @@
...
+    # get the point of interest coordinate (in microscope coordinates, in metres)
+    poi_coord = correlation_results["output"]["poi"][0]["px_um"]
+    poi_coord = (poi_coord[0] * 1e-6, poi_coord[1] * 1e-6)
+    return poi_coord
+
+def parse_3dct_yaml_file(path: str) -> Tuple[float, float]:
+    """Parse the 3DCT yaml file and extract the point of interest (POI)
+    in microscope image coordinates (um). Convert the coordinates to metres.
+    Note: only the first POI is extracted.
+    :param path: Path to the 3DCT yaml file."""
```
pieleric, line 140: Please add a docstring about the result.

src/odemis/acq/align/tdct.py
```diff
@@ -0,0 +1,146 @@
...
+    :param correlation_results: the correlation results
+    :return: the point of interest coordinate in meters
+    """
+    # get the point of interest coordinate (in microscope coordinates, in metres)
+    poi_coord = correlation_results["output"]["poi"][0]["px_um"]
+    poi_coord = (poi_coord[0] * 1e-6, poi_coord[1] * 1e-6)
+    return poi_coord
+
+def parse_3dct_yaml_file(path: str) -> Tuple[float, float]:
+    """Parse the 3DCT yaml file and extract the point of interest (POI)
```
pieleric, line 137: Is that function going to be useful? Why would we have a YAML file generated by 3DCT?

src/odemis/acq/align/tdct.py
```diff
@@ -0,0 +1,146 @@
...
+from odemis import model
+
+# tmp: append the path to the 3DCT module until installation solution
+sys.path.append("/home/patrick/development/openfibsem/3DCT")
+sys.path.append("/home/patrick/development/openfibsem/3DCT/tdct")
+from tdct.correlation_v2 import run_correlation
+from tdct.util import multi_channel_get_z_guass
+
+def convert_das_to_numpy_stack(das: List[model.DataArray]) -> numpy.ndarray:
+    """Convert a list of DataArrays to a numpy stack."""
```
pieleric, line 36: Please add a docstring about the input argument and return value.
Especially, what's the shape of these arrays.

src/odemis/acq/align/tdct.py
```diff
@@ -0,0 +1,146 @@
...
+    arr = []
+    for da in das:
+        if isinstance(da, model.DataArrayShadow):
+            da = da.getData()
+        if da.ndim == 5: # model.DataArray is always 5D
```
pieleric, line 41: Not true. They are *usually* 5D after you read them back from a TIFF or HDF5 file. If you directly get them from the acquisition manager, they might be just 2D or 3D.

src/odemis/acq/align/tdct.py
```diff
@@ -0,0 +1,146 @@
...
+            da = da[0, 0, :, :, :]
```
pieleric, line 42: Add at least a check here that the first two dimension are 1,1. If not, log a warning that the data in CT is being discarded.

src/odemis/acq/align/tdct.py
```diff
@@ -0,0 +1,146 @@
...
+
+from odemis import model
+
+# tmp: append the path to the 3DCT module until installation solution
+sys.path.append("/home/patrick/development/openfibsem/3DCT")
+sys.path.append("/home/patrick/development/openfibsem/3DCT/tdct")
+from tdct.correlation_v2 import run_correlation
+from tdct.util import multi_channel_get_z_guass
+
+def convert_das_to_numpy_stack(das: List[model.DataArray]) -> numpy.ndarray:
```
pieleric, line 35: I looks like a helper function for `get_optimized_z_guass()`. If it's not to be used by external code, add a `_` at the beginning of the name.

src/odemis/acq/align/tdct.py
```diff
@@ -0,0 +1,146 @@
...
+            da = da.getData()
+        if da.ndim == 5: # model.DataArray is always 5D
+            da = da[0, 0, :, :, :]
+        arr.append(da)
+
+    return numpy.stack(arr, axis=0)
+
+def get_optimized_z_guass(das: List[model.DataArray], x: int, y: int, z: int, show: bool = False) -> Tuple[float, float, float]:
+    """Get the best fitting z-coordinate for the given x, y coordinates. Supports multi-channel images.
+    :param das: the data arrays (one per channels)
```
pieleric, line 49: Please indicate the dimensions. It seems they all must have the same shape (including in Z), right?

src/odemis/acq/align/tdct.py
```diff
@@ -0,0 +1,146 @@
...
+    :param x: the x-coordinate
+    :param y: the y-coordinate
+    :param z: the z-coordinate (initial guess)
+    :param show: show the plot for debugging
+    :return: the z-coordinate and the corresponding correlation value"""
```
pieleric, line 54: The code looks like it's returning just 1 value. The typing info shows 3 values. This doc says 2 values. Can you fix the discrepancy?

src/odemis/acq/align/tdct.py
```diff
@@ -0,0 +1,146 @@
...
+from typing import Dict, List, Tuple
+
+import numpy
+import yaml
+
+from odemis import model
+
+# tmp: append the path to the 3DCT module until installation solution
+sys.path.append("/home/patrick/development/openfibsem/3DCT")
+sys.path.append("/home/patrick/development/openfibsem/3DCT/tdct")
```
pieleric, line 31: You should just assume that the PYTHONPATH contains 3DCT. If you want you can add a comment, with the URL to the github repo from where to get the code. Eventually, it should be installed via a debian package. In the meantime, it can be in ~/development/, and the path added to `odemis.conf` .

BTW, I think only the first path is needed.

pieleric (changes requested):
The main thing missing here is the unit test code. Please add a couple of unit test, to show at least how the main functions are to be called. You can just hard-code some number for input, and have some basic expectation on the output.

src/odemis/acq/align/tdct.py
```diff
@@ -0,0 +1,146 @@
...
+    for da in das:
+        if isinstance(da, model.DataArrayShadow):
+            da = da.getData()
+        if da.ndim == 5: # model.DataArray is always 5D
+            da = da[0, 0, :, :, :]
+        arr.append(da)
+
+    return numpy.stack(arr, axis=0)
+
+def get_optimized_z_guass(das: List[model.DataArray], x: int, y: int, z: int, show: bool = False) -> Tuple[float, float, float]:
```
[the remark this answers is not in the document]
patrickcleeve2: gaussian

src/odemis/acq/align/tdct.py
patrickcleeve2: This is so you can load the correlated point from external use of 3dct for other methods that odemis doesn't support. E.g. right now multi-point isn't available in odemis but is in 3dct. So users could technically use whatever correlation method they want, and then import the position into odemis and use the rest of the workflow

src/odemis/acq/align/tdct.py
```diff
@@ -0,0 +1,146 @@
...
+    # run correlation
+    correlation_results = run_correlation(
+        fib_coords=fib_coords,
+        fm_coords=fm_coords,
+        poi_coords=poi_coords,
+        image_props=image_props,
+        rotation_center=rotation_center,
+        path=path,
+        fib_image_filename="FIB-Image",
+        fm_image_filename="FM-Image",
```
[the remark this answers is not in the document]
patrickcleeve2: removed

src/odemis/acq/align/tdct.py
patrickcleeve2: it's centred at the image centre

src/odemis/acq/align/tdct.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+def run_tdct_correlation(fib_coords: numpy.ndarray,
+                           fm_coords: numpy.ndarray,
+                           poi_coords: numpy.ndarray,
+                           fib_image: model.DataArray,
+                           fm_image: model.DataArray,
+                           path: str) -> Dict[str, Tuple[float, float]]:
+    """Run 3DCT Multi-point correlation between FIB and FM images.
+    :param fib_coords: the FIB coordinates (x, y, z)
```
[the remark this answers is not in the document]
patrickcleeve2: xyz is just how the 3dct code expects the array. I've softened the requirement here, so you can pass x,y and it will add the empty z=0 to the array.

src/odemis/acq/align/tdct.py
```diff
@@ -0,0 +1,193 @@
...
+    )
+
+    return correlation_results
+
+def get_poi_coordinate(correlation_results: dict) -> Tuple[float, float]:
+    """Get the the point of interest coordinate from correlation data
+    and convert from micrometers to meters in the microscope image coordinate system.
+    The coordinate is centred at the image centre.
+    :param correlation_results: the correlation results
+    :return: the point of interest coordinate in meters
```
pieleric, line 174: If the coordinates are in meters, with origin on the image center, then, it's quite close from the "standard" world coordinates in Odemis, which I assume was the goal of this function! *However*, in Odemis, +Y in world coordinates goes *up*. Is that the case here, or does the "poi" output follows the convention that +Y goes down? I would suggest to ensure the result follows the Odemis convention, so that it's easy to directly use (ie, please check if you need to put a `-` before the `poi_coord[1]`)

pieleric (changes requested):
Looks good. Just one very minor comment.

src/odemis/acq/align/tdct.py
patrickcleeve2: It is actually the coordinate system used by the milling patterns, which is the same as world coordinates in odemis (by coincidence, not design). I've added the axes for the coordinates (x+ -> right, y+ -> up) in the docstring for clarity

pieleric (approved):

src/odemis/acq/align/tdct.py
```diff
@@ -0,0 +1,193 @@
...
+        fm_coords=fm_coords,
+        poi_coords=poi_coords,
+        image_props=image_props,
+        rotation_center=rotation_center,
+        path=path,
+    )
+
+    return correlation_results
+
+def get_poi_coordinate(correlation_results: dict) -> Tuple[float, float]:
```
K4rishma, line 169: It is not apparent that there is a difference between poi_coords which is used as input in run_tdct_correlation function and the output parameter of this function. 

For example to differentiate the two parameters I used the names as in the following structure - 

```
class CorrelationTarget:
    def __init__(self):
        self.fm_pois : List[Target] = []
        self.fm_fiducials : List[Target] = []
        self.fib_fiducials : List[Target] = []
        self.fib_surface_fiducial : Target = None

        self.correlation_result: float = None
        self.refractive_index_correction: bool = True
        self.fib_projected_pois: List[Target] = []
        self.fib_projected_fiducials: List[Target] = []

        self.fib_stream = None #:StaticSEMStream = None
        self.fm_streams = None #: List[StaticFluoStream] = []
        self.superz = None #: StaticFluoStream = None
        self.image_metadata: CorrelationMetadata = None

    def reset_attributes(self):
        # rest of the attributes is set to none except the streams
        self.correlation_result = None
        self.fib_projected_pois = None
        self.fib_projected_fiducials = None
```

src/odemis/acq/align/tdct.py
```diff
@@ -0,0 +1,193 @@
...
+    return z
+
+def run_tdct_correlation(fib_coords: numpy.ndarray,
+                           fm_coords: numpy.ndarray,
+                           poi_coords: numpy.ndarray,
+                           fib_image: model.DataArray,
+                           fm_image: model.DataArray,
+                           path: str) -> Dict[str, Any]:
+    """Run 3DCT Multi-point correlation between FIB and FM images.
+    :param fib_coords: the FIB coordinates (x, y) (in pixels, origin at top left)
```
tmoerkerken, line 96: plural, so I would expect a shape of (n_coords, x, y) (dito for the other coordinate arguments)

src/odemis/acq/align/tdct.py
```diff
@@ -0,0 +1,193 @@
...
+        fib_coords = numpy.column_stack((fib_coords, numpy.zeros(fib_coords.shape[0])))
+
+    # coordinates need to be float32 for 3DCT
+    fib_coords = fib_coords.astype(numpy.float32)
+    fm_coords = fm_coords.astype(numpy.float32)
+
+    # get first channel only, assume all channels are the same shape
+    if fm_image.ndim == 4:
+        fm_image = fm_image[0, :, :, :]
+    if fm_image.ndim == 5:
```
tmoerkerken, line 131: According to the docstring, fm_image has at most an ndim of 4, unless you forgot to mention CTZYX in there.

src/odemis/acq/align/tdct.py
```diff
@@ -0,0 +1,193 @@
...
+
+    return z
+
+def run_tdct_correlation(fib_coords: numpy.ndarray,
+                           fm_coords: numpy.ndarray,
+                           poi_coords: numpy.ndarray,
+                           fib_image: model.DataArray,
+                           fm_image: model.DataArray,
+                           path: str) -> Dict[str, Any]:
+    """Run 3DCT Multi-point correlation between FIB and FM images.
```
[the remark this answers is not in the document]
patrickcleeve2: 3D, cant have numbers to start package name

src/odemis/acq/align/tdct.py
```diff
@@ -0,0 +1,193 @@
...
+
+        assert da.ndim == 3, f"DataArray must be 3D ZYX, but is {da.shape}"
+        arr.append(da)
+
+    return numpy.stack(arr, axis=0)
+
+def get_optimized_z_gauss(das: List[model.DataArray], x: int, y: int, z: int, show: bool = False) -> float:
+    """Get the best fitting z-coordinate for the given x, y coordinates. Supports multi-channel images.
+    :param das: the data arrays (CTZYX, ZYX, or YX), all arrays must have the same shape
+    :param x: the x-coordinate
```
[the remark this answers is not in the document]
patrickcleeve2: It's meant to be generic, doesnt have to just be for this use case

src/odemis/acq/align/tdct.py
```diff
@@ -0,0 +1,193 @@
...
+    """Get the the point of interest coordinate from correlation data
+    and convert from micrometers to meters in the microscope image coordinate system.
+    The coordinate is centred at the image centre (x+ -> right, y+ -> up).
+    :param correlation_results: the correlation results
+    :return: the point of interest coordinate in meters
+    """
+    # get the point of interest coordinate (in microscope coordinates, in metres)
+    poi_coord = correlation_results["output"]["poi"][0]["px_um"]
+    poi_coord = (poi_coord[0] * 1e-6, poi_coord[1] * 1e-6)
+    return poi_coord
```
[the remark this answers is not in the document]
patrickcleeve2: this is the final output for the milling patterns, which have coordinates in metres

src/odemis/acq/align/tdct.py
```diff
@@ -0,0 +1,193 @@
...
+            error:
+                delta_2d: reprojection error between 3D and 2D coordinates
+                reprojected_3d: 3D coordinates reprojected to 2D
+                mean_absolute_error: mean absolute error of the transformation (x, y)
+                rms_error: root mean square error of the transformation
+            poi: list of transformed point of interest coordinates
+                image_px: coordinates in image pixels (0, 0 top left)
+                px:  coordinates in microscope image pixels (0, 0 image center)
+                px_um: coordinates in microscope image meters (0, 0 image center)
+            transfromation:
```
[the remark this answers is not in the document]
patrickcleeve2: transformation matricies are 'unitless'. translation is the same as the input dimensnions

src/odemis/acq/align/tdct.py
patrickcleeve2: changed it to get_reprojected_poi_coordinate

src/odemis/acq/align/tdct.py
patrickcleeve2: updated

src/odemis/acq/align/tdct.py
```diff
@@ -0,0 +1,193 @@
...
+
+def run_tdct_correlation(fib_coords: numpy.ndarray,
+                           fm_coords: numpy.ndarray,
+                           poi_coords: numpy.ndarray,
+                           fib_image: model.DataArray,
+                           fm_image: model.DataArray,
+                           path: str) -> Dict[str, Any]:
+    """Run 3DCT Multi-point correlation between FIB and FM images.
+    :param fib_coords: the FIB coordinates (x, y) (in pixels, origin at top left)
+    :param fm_coords: the FM coordinates (x, y, z) (in pixels, origin at top left)
```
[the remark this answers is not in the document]
patrickcleeve2: yes, but they don't have to be. it's a generic transformation algo

src/odemis/acq/align/tdct.py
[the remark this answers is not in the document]
patrickcleeve2: updated

src/odemis/acq/align/tdct.py
```diff
@@ -0,0 +1,193 @@
...
+        arr.append(da)
+
+    return numpy.stack(arr, axis=0)
+
+def get_optimized_z_gauss(das: List[model.DataArray], x: int, y: int, z: int, show: bool = False) -> float:
+    """Get the best fitting z-coordinate for the given x, y coordinates. Supports multi-channel images.
+    :param das: the data arrays (CTZYX, ZYX, or YX), all arrays must have the same shape
+    :param x: the x-coordinate
+    :param y: the y-coordinate
+    :param z: the z-coordinate (initial guess)
```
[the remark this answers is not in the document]
patrickcleeve2: I understand your point but disagree, almost always the handling will be to use the initial value. If it becomes an issue it can be changed later

src/odemis/acq/align/tdct.py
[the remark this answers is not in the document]
K4rishma: Maybe if most of the coordinates are in pixels, by default the coords can represent pixels and for m it can be represented by pos , while the function name can explicitly tell i.e. in meters ? For e.g. get_poi_pos() or get_poi_coords_in_meters()

src/odemis/acq/align/tdct.py
K4rishma: I would use get_projected_poi_coordinate because of two reasons. First, projected implies it is the same point in another coordinate system and secondly, I am using projected in my code which will provide a similar interface :).

K4rishma:
Just a quick question, how would the refractive index correction be handled? I am not sure if I have seen that code.

K4rishma (approved):

patrickcleeve2:
> Just a quick question, how would the refractive index correction be handled? I am not sure if I have seen that code.

That isn't implemented ye

PR #3001, opened 2025-01-22
[fix][METEOR-960] Minor client fixes from testing

patrickcleeve2:
Minor client fixes/updates from hardware testing

### DIFF
--- a/install/linux/usr/share/odemis/sim/meteor-fibsem-sim.odm.yaml
+++ b/install/linux/usr/share/odemis/sim/meteor-fibsem-sim.odm.yaml
@@ -37,7 +37,10 @@ METEOR-FIBSEM-Sim: {
 
 "Electron-Detector": {
     role: se-detector,
-    init: {}
+    init: {},
+    properties: {
+        medianFilter: 3,
+    }
 
 }
 
@@ -56,7 +59,10 @@ METEOR-FIBSEM-Sim: {
 
 "Ion-Detector": {
     role: "se-detector-ion",
-    init: {}
+    init: {},
+    properties: {
+        medianFilter: 3,
+    }
 }
 
 "Ion-Focus": {
@@ -84,7 +90,6 @@ METEOR-FIBSEM-Sim: {
                 "Sample pre-tilt":  0.6108652381980153},  # 35°
         FAV_FM_POS_ACTIVE: {"rx": 0.12213888553625313  , "rz":  3.141592653589793}, # 7° - 270°
         FAV_SEM_POS_ACTIVE: {"rx": 0.6108652381980153, "rz": 0},  # pre-tilt 35°
-        FAV_MILL_POS_ACTIVE: {"rx": 0.314159, "rz": 0},    # Note that milling angle (rx) can be changed per session  
     },
 }
 
--- a/src/odemis/driver/autoscript_client.py
+++ b/src/odemis/driver/autoscript_client.py
@@ -33,6 +33,7 @@ import numpy
 import pkg_resources
 import Pyro5.api
 from Pyro5.errors import CommunicationError
+from scipy import ndimage
 
 from odemis import model, util
 from odemis.driver.xt_client import check_and_transfer_latest_package
@@ -1244,6 +1245,10 @@ class Scanner(model.Emitter):
                 if dwell_time != self.dwellTime.value:
                     self.dwellTime._value = dwell_time
                     self.dwellTime.notify(dwell_time)
+                res = self.parent.get_resolution(self.channel)
+                if res != self.resolution.value:
+                    self.resolution._value = res
+                    self.resolution.notify(res)
                 self._updateResolution()
 
             voltage = self.parent.get_high_voltage(self.channel)
@@ -1254,6 +1259,10 @@ class Scanner(model.Emitter):
             if voltage != self.accelVoltage.value:
                 self.accelVoltage._value = voltage
                 self.accelVoltage.notify(voltage)
+            beam_current = self.parent.get_beam_current(self.channel)
+            if beam_current != self.probeCurrent.value:
+                self.probeCurrent._value = beam_current
+                self.probeCurrent.notify(beam_current)
             beam_shift = self.parent.get_beam_shift(self.channel)
             if beam_shift != self.shift.value:
                 self.shift._value = beam_shift
@@ -1454,6 +1463,9 @@ class Detector(model.Detector):
         # this makes it pretty annoying to do anything on that side, and error prone.
         # disabling this until a better solution is found
 
+        # median filter applied to the image (required for cryo data)
+        self.medianFilter = model.IntContinuous(0, range=(0, 9), setter=self._setMedianFilter)
+
     def terminate(self) -> None:
         if self._generator:
             self.stop_generate()
@@ -1520,6 +1532,9 @@ class Detector(model.Detector):
                     # TODO: use the metadata from the image acquisition _md once it's available
                     image, _md = self.parent.acquire_image(self._scanner.channel)
 
+                    # median filter to remove noise (required for cryo data)
+                    if self.medianFilter.value > 0:
+                        image = ndimage.median_filter(image, self.medianFilter.value)
                     # non-blocking acquisition (disabled until hw testing)
                     # logging.debug("Starting one image acquisition")
                     # # start the acquisition
@@ -1687,6 +1702,17 @@ class Detector(model.Detector):
     def _setDetectorType(self, detector_type: str) -> str:
         self.parent.set_detector_type(detector_type, self._scanner.channel)
         return self.parent.get_detector_type(self._scanner.channel)
+
+    def _setMedianFilter(self, value: int) -> int:
+        """Set the median filter value and update the metadata."""
+
+        # if value is 0, remove the filter from the metadata
+        if value == 0:
+            self.updateMetadata({model.MD_DATA_FILTER: None})
+        else:
+            self.updateMetadata({model.MD_DATA_FILTER: f"median-filter:{value}"})
+        return value
+
     # TODO: add support for auto functions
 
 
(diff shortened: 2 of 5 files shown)


src/odemis/driver/autoscript_client.py
```diff
@@ -33,7 +33,7 @@
 import pkg_resources
 import Pyro5.api
 from Pyro5.errors import CommunicationError
-
+from scipy import ndimage
 from odemis import model, util
```
tepals, line 37: Can you follow the import order from the guidelines? https://github.com/delmic/odemis/blob/master/doc/develop/coding-guidelines.rst#import-order
```suggestion
from Pyro5.errors import CommunicationError
from scipy import ndimage

from odemis import model, util
```

src/odemis/driver/autoscript_client.py
```diff
@@ -208,6 +208,14 @@ def get_hardware_version(self) -> str:
             self.server._pyroClaimOwnership()
             return self.server.get_hardware_version()
 
+    def list_available_channels(self) -> List[str]:
```
tepals, line 211: I think for the xt_client we left this out of the client side, since there was not really a use case for it. I don't see you using it here, is it necessary to have it?

src/odemis/driver/autoscript_client.py
```diff
@@ -267,7 +275,6 @@ def get_stage_coordinate_system(self) -> str:
         with self._proxy_access:
             self.server._pyroClaimOwnership()
             return self.server.get_stage_coordinate_system()
-
```
tepals, line 270: can you put back the empty line?

src/odemis/driver/autoscript_client.py
```diff
@@ -1519,7 +1534,7 @@ def _acquire(self) -> None:
                     # Retrieve the image (scans image, blocks until the image is received)
                     # TODO: use the metadata from the image acquisition _md once it's available
                     image, _md = self.parent.acquire_image(self._scanner.channel)
-
+                    image = ndimage.median_filter(image, 3)  # median filter to remove noise
```
tepals, line 1537: Why do you want to remove noise? I would expect the acquire method to just acquire and not do any image adjustments.

src/odemis/driver/autoscript_client.py
```diff
@@ -1770,12 +1792,31 @@ def terminate(self):
...
+        pos_linked = self._getPosition()
+        self.parent.set_default_stage_coordinate_system("RAW")
+        pos = self._getPosition()
+        for axis in self._raw_offset.keys():
+            self._raw_offset[axis] = pos_linked[axis] - pos[axis]
+        # the offset should only be in linear axes, it is not expected in rotational axes
+        if not all(pos[axis] == pos_linked[axis] for axis in ["rx", "rz"]):
+            logging.warning(
+                "During offset estimation in linear axes in raw coordinate system in x and y, unexpected offset is "
+                "found in rotation axes")
```
tepals, line 1807: I have a hard time understanding what this warning means, can you write it down in a more clear way?

src/odemis/driver/autoscript_client.py
pieleric, line 1537: This shouldn't be hard-coded in the driver. The driver should just pass the image as-is. Please move this to the acq/gui code, where de-noising is required.

src/odemis/driver/autoscript_client.py
```diff
@@ -1748,13 +1763,20 @@ def __init__(self, name: str, role: str, parent: SEM, rng: Optional[Dict[str, Tu
             "rz": model.Axis(unit=stage_info["unit"]["r"], range=rng["rz"]),
         }
 
+        # When raw coordinate system is selected, in theory just z should change
```
tepals, line 1766: Out of curiosity: why didn't you implement `_switch_coordinate_system` for the autoscript_client?

pieleric (changes requested):

tepals (changes requested):

src/odemis/driver/autoscript_client.py
patrickcleeve2: better?

src/odemis/driver/autoscript_client.py
patrickcleeve2: There's no use case for using linked coordinate system now

src/odemis/driver/autoscript_client.py
patrickcleeve2: """[tepals](https://github.com/tepals) [3 days ago](https://github.com/delmic/odemis/pull/3001#discussion_r1928682232)
Why do you want to remove noise? I would expect the acquire method to just acquire and not do any image adjustments.
"""

@tepals @pieleric responding to the same comment for clarity. 
Every cryo image needs to be median filtered because the data is bad, so i think the most natural place is in the driver. I've adjusted it so that it's a option in the metadata, so it's not hardcoded but configurable. 
If not here, where do you suggest? I'm not really sure of a suitable place

tepals: How would a user normally do this? Will they select median filtering by default in the XT UI?

install/linux/usr/share/odemis/sim/meteor-fibsem-sim.odm.yaml
```diff
@@ -84,7 +90,7 @@ METEOR-FIBSEM-Sim: {
                 "Sample pre-tilt":  0.6108652381980153},  # 35°
         FAV_FM_POS_ACTIVE: {"rx": 0.12213888553625313  , "rz":  3.141592653589793}, # 7° - 270°
         FAV_SEM_POS_ACTIVE: {"rx": 0.6108652381980153, "rz": 0},  # pre-tilt 35°
-        FAV_MILL_POS_ACTIVE: {"rx": 0.314159, "rz": 0},    # Note that milling angle (rx) can be changed per session  
+        # FAV_MILL_POS_ACTIVE: {"rx": 0.314159, "rz": 0},    # Note that milling angle (rx) can be changed per session
```
tepals, line 93: why is this commented?

src/odemis/driver/autoscript_client.py
```diff
@@ -1454,6 +1463,12 @@ def __init__(self, name: str, role: str, parent: SEM, channel: str, **kwargs):
         # this makes it pretty annoying to do anything on that side, and error prone.
         # disabling this until a better solution is found
 
+        # median filter applied to the image (required for cryo data)
+        self.median_filter = model.FloatContinuous(
+            value=self.getMetadata().get(model.MD_MEDIAN_FILTER, 3),
+            range=(0, 9),
+        )
```
tepals, line 1470: I would have it turned off by default. We will most likely also use this code for non-cryo data.
Also I think it should be an `int` not a `float`, since you are using it as the size input of the median filter function.

tepals (changes requested):

install/linux/usr/share/odemis/sim/meteor-fibsem-sim.odm.yaml
patrickcleeve2: The metadata item isn't added until another PR: https://github.com/delmic/odemis/pull/3005/files#diff-a2c9bc5144464d40c3b2bb45d972506662f62a97029ccb1dac0f0ac3ea68274f. It'll be uncommented then

src/odemis/driver/autoscript_client.py
patrickcleeve2: not sure if you can change it in the ui, but it would be expected

tepals (approved):

src/odemis/driver/autoscript_client.py
```diff
@@ -1770,12 +1795,31 @@ def terminate(self):
             self._pos_poll.cancel()
             self._pos_poll = None
 
+    def _get_coordinate_system_offset(self):
```
pieleric, line 1798: By convention, `get_*()` functions return a value. Here, the function only store the value. So let's pick a different name. For instance, `_store_coordinate_system_offset()`, or `_update_coordinate_system_offset()`.

src/odemis/model/_metadata.py
```diff
@@ -89,6 +89,7 @@
 # position of the stage (in m or rad) for each axis in the chamber (raw hardware values)
 MD_STAGE_POSITION_RAW = "Stage position raw"  # dict of str -> float,
 MD_SAMPLE_PRE_TILT = "pre-tilt"  # (rad) pre-tilt of the sample stage / shuttle (tilt)
+MD_MEDIAN_FILTER = "median-filter" # median filter applied to detector data (req for cryo)
```
pieleric, line 92: Let's drop this. If you go with the filter inside the driver, then introduce MD_DATA_FILTER = "Data filter"

pieleric (changes requested):

src/odemis/driver/autoscript_client.py
```diff
@@ -1454,6 +1463,12 @@ def __init__(self, name: str, role: str, parent: SEM, channel: str, **kwargs):
         # this makes it pretty annoying to do anything on that side, and error prone.
         # disabling this until a better solution is found
 
+        # median filter applied to the image (required for cryo data)
+        self.median_filter = model.IntContinuous(
+            value=self.getMetadata().get(model.MD_MEDIAN_FILTER, 0),
```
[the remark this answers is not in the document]
patrickcleeve2: I've changed it to properties field in the yaml file, and added the metadata field. I think it makes the most sense here, rather than the display. It is needed for both sem and fib, so it would have to be more generic than an individual stream type. I'm not sure if it is actually needed on other systems, but it is fairly easy to add.

src/odemis/driver/autoscript_client.py
```diff
@@ -1520,6 +1532,9 @@ def _acquire(self) -> None:
                     # TODO: use the metadata from the image acquisition _md once it's available
                     image, _md = self.parent.acquire_image(self._scanner.channel)
 
+                    # median filter to remove noise (required for cryo data)
```
pieleric, line 1535: Could you "extend" the `test_detector_component` test case to acquire an image, with the median filter set to 0, and another one with the filter set to 3? 
This would help testing all this code a little bit!

pieleric (changes requested):

src/odemis/driver/autoscript_client.py
patrickcleeve2: added to autoscript_client_test

pieleric (approved):

PR #3003, opened 2025-01-22
[fix] newlines in ar csv export (Windows)

tmoerkerken:
On Windows, the default newline character is "\n" for exporting rows with csv writer.
Explicitly overwrite the newline character to be empty. This behavior never occurred on Linux due to a different newline convention.

Example of current behavior in Windows: 
image

### DIFF
--- a/src/odemis/dataio/csv.py
+++ b/src/odemis/dataio/csv.py
@@ -276,7 +276,10 @@ def _export_angular_spectrum_data(data: model.DataArray, filename:str):
     headers = ["angle(" + unit_a + ")\\wavelength(" + unit_c + ")"] + spectrum_range
     rows = [(t,) + tuple(d) for t, d in zip(angle_range, data)]
 
-    with open(filename, 'w') as fd:
+    # Specifically set newline to an empty string, as described in the official documentation,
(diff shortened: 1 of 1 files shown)


K4rishma (approved):

src/odemis/dataio/csv.py
```diff
@@ -276,7 +276,7 @@ def _export_angular_spectrum_data(data: model.DataArray, filename:str):
     headers = ["angle(" + unit_a + ")\\wavelength(" + unit_c + ")"] + spectrum_range
     rows = [(t,) + tuple(d) for t, d in zip(angle_range, data)]
 
-    with open(filename, 'w') as fd:
+    with open(filename, 'w', newline='') as fd:
```
pieleric, line 279: That's indeed how the other functions also do it. However, none of them explains *why*. At first look, it's a really odd code, as it seems like it would prevent new lines. 
Looking into the doc, that's actually explained in the csv module: https://docs.python.org/3/library/csv.html 
It handles newlines by itself (to make sure it's OS-independent), so that's why you have to disable it in the `open()` function.

pieleric (changes requested):

pieleric (approved):

PR #3005, opened 2025-01-23
[feat][METEOR-996] Add TFS3 posture manager for multi-posture imaging

patrickcleeve2:
Adds:

- TFS3 posture manager for multi-posture imaging functionality.
- SampleStage virtual component

[JIRA: METEOR-996](https://delmic.atlassian.net/browse/METEOR-996?atlOrigin=eyJpIjoiMTZkZWQ3MzExZmRmNDE1ODk3ZDUxNTA0M2Q3NjIxN2YiLCJwIjoiaiJ9)
[Specification]( https://docs.google.com/document/d/1uDZrg1up0wrbu5ODU1uDm9iSeUZNYJitedxlhfMYfZk/edit?pli=1&tab=t.0#heading=h.b8qus0izsixj)

### DIFF
--- a/src/odemis/acq/move.py
+++ b/src/odemis/acq/move.py
@@ -25,22 +25,23 @@ import logging
 import math
 import threading
 from abc import abstractmethod
-from concurrent.futures import CancelledError
-from concurrent.futures._base import CANCELLED, RUNNING, FINISHED
-from typing import Dict, Union, List, Iterable
+from concurrent.futures import CancelledError, Future
+from concurrent.futures._base import CANCELLED, FINISHED, RUNNING
+from typing import Dict, Union, List, Iterable, Tuple
 
 import numpy
 import scipy
 
 from odemis import model, util
+from odemis.model import isasync
 from odemis.util import executeAsyncTask
-from odemis.util.driver import ATOL_ROTATION_POS, isNearPosition, isInRange
-from odemis.util.transform import RigidTransform
+from odemis.util.driver import ATOL_ROTATION_POS, isInRange, isNearPosition
+from odemis.util.transform import RigidTransform, get_rotation_transforms
 
 MAX_SUBMOVE_DURATION = 90  # s
 
 UNKNOWN, LOADING, IMAGING, ALIGNMENT, COATING, LOADING_PATH, MILLING, SEM_IMAGING, \
-    FM_IMAGING, GRID_1, GRID_2, THREE_BEAMS = -1, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10
+    FM_IMAGING, GRID_1, GRID_2, THREE_BEAMS, FIB_IMAGING = -1, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11
 POSITION_NAMES = {
     UNKNOWN: "UNKNOWN",
     LOADING: "LOADING",
@@ -53,7 +54,8 @@ POSITION_NAMES = {
     FM_IMAGING: "FM IMAGING",
     GRID_1: "GRID 1",
     GRID_2: "GRID 2",
-    THREE_BEAMS: "THREE BEAMS"
+    THREE_BEAMS: "THREE BEAMS",
+    FIB_IMAGING: "FIB_IMAGING"
 }
 
 RTOL_PROGRESS = 0.3
@@ -67,6 +69,13 @@ SAFETY_MARGIN_3DOF = 200e-6  # m
 ATOL_ROTATION_TRANSFORM = 0.04  # rad ~2.5 deg
 ATOL_LINEAR_TRANSFORM = 5e-6  # 5 um
 
+# roles that are affected by sample stage transformation
+COMPS_AFFECTED_ROLES = ["ccd", "e-beam", "ion-beam"]
+
+# fib column tilts, relative to sem column
+TFS_FIB_COLUMN_TILT = math.radians(52)
+TESCAN_FIB_COLUMN_TILT = math.radians(55)
+ZEISS_FIB_COLUMN_TILT = math.radians(54)
 
 class MicroscopePostureManager:
     def __new__(cls, microscope):
@@ -85,6 +94,8 @@ class MicroscopePostureManager:
                 return super().__new__(MeteorTFS1PostureManager)
             elif stage_version == "tfs_2":
                 return super().__new__(MeteorTFS2PostureManager)
+            elif stage_version == "tfs_3":
+                return super().__new__(MeteorTFS3PostureManager)
             elif stage_version == "tescan_1":
                 return super().__new__(MeteorTescan1PostureManager)
             else:
@@ -301,9 +312,57 @@ class MeteorPostureManager(MicroscopePostureManager):
         self._inv_transforms: Dict[int, numpy.ndarray] = {}  # inverse transforms (from-sample-stage)
         self._metadata = {}
         self._axes_dep = {}  # axes dependencies between different planes
+
+        # pre-tilt is required for milling posture, but not all systems have it
+        stage_md = self.stage.getMetadata()
+        md_calib = stage_md.get(model.MD_CALIB, {})
+        self.pre_tilt = md_calib.get(model.MD_SAMPLE_PRE_TILT, None)
+        self.fib_column_tilt = TFS_FIB_COLUMN_TILT
+
+        # feature flags, for features still in testing
+        # NOTE: use_linked_sem_focus_compensation:
+        # when true, the SEM focus is restored to the eucentric focus when moving the stage
+        # this is done on TFS systems to compensate for the SEM focus changing when the stage is moved (due to stage linking)
+        # NOTE: use_3d_transforms:
+        # when true, a 3D transformation matrix is used to convert between sample and stage positions.
+        # previously, only a 2D matrix was used, as the only transforms was around the x-axis (pre-tilt compensation). The x-axis coordinate
+        # was grafted onto the output of the 2D transform which only applied to the y/z axes. This is now done in a single matrix.
+        # NOTE: use_scan_rotation:
+        # when true, an additional rotation matrix is applied to the sample-stage transforms (rotation around z-axes of sample-stage),
+        # to account for scan rotation. The code checks that the scan rotation is the same for both the SEM/FIB before applying the transform.
+        # this requires the use_3d_transforms to be enabled. It is only expected to work for either 0 or 180 degree scan rotation, which in practice
+        # are the only values used.
+        # NOTE: use_3d_transforms and use_scan_rotation are still in testing, so are not enabled by default.
+        # NOTE: these flags will be removed, along with 2D transforms, once 3D transforms and scan rotation are fully tested.
+        self.use_linked_sem_focus_compensation: bool = md_calib.get("use_linked_sem_focus_compensation", False)
+        self.use_3d_transforms: bool = md_calib.get("use_3d_transforms", False)
+        self.use_scan_rotation: bool = md_calib.get("use_scan_rotation", False)
+
         # current posture va
         self.current_posture = model.VigilantAttribute(UNKNOWN)
         self.stage.position.subscribe(self._update_posture, init=True)
+
+
+        # set the transforms between different postures
+        self._posture_transforms = {
+            FM_IMAGING: {
+                SEM_IMAGING: self._transformFromMeteorToSEM,
+                MILLING: self._transformFromMeteorToMilling,
+            },
+            SEM_IMAGING: {
+                FM_IMAGING: self._transformFromSEMToMeteor,
+                MILLING: self._transformFromSEMToMilling,
+            },
+            MILLING: {
+                SEM_IMAGING: self._transformFromMillingToSEM,
+                FM_IMAGING: self._transformFromMillingToFM,
+                # milling position can be dynamically updated, so we need to support this recalculation
+                MILLING: self._transformFromSEMToMilling,
+            },
+            UNKNOWN: {
+                UNKNOWN: lambda x: x
+         }
+        }
 
     def getCurrentPostureLabel(self, pos: Dict[str, float] = None) -> int:
         """
@@ -324,9 +383,46 @@ class MeteorPostureManager(MicroscopePostureManager):
         if isInRange(pos, stage_fm_imaging_rng, self.linear_axes):
             return FM_IMAGING
         if isInRange(pos, stage_sem_imaging_rng, self.linear_axes):
+            if self.at_milling_posture(pos, stage_md):
+                return MILLING
             return SEM_IMAGING
         # None of the above -> unknown position
         return UNKNOWN
+
+    def at_milling_posture(self, pos: Dict[str, float], stage_md: Dict[str, float]) -> bool:
+        """Milling posture is not required for all meteor systems, so we need to
+        first check it's available
+        :param pos the stage position
+        :param stage_md the stage metadata
+        :param return True if the stage is at the milling posture, False if not (or not available)"""
+        if model.MD_FAV_MILL_POS_ACTIVE in stage_md:
+            stage_milling = self.get_posture_orientation(MILLING)
+            if isNearPosition(pos,
+                            stage_milling,
+                            self.rotational_axes,
+                            atol_rotation=math.radians(3)):
+                return True
+        return False
+
+    def get_posture_orientation(self, posture: int) -> Dict[str, float]:
+        """Get the orientation of the stage for the given posture
+        :param posture: the posture to get the orientation for
+        :return: a dict with the orientation of the stage for the given posture"""
+        stage_md = self.stage.getMetadata()
+        if posture == SEM_IMAGING:
+            return stage_md[model.MD_FAV_SEM_POS_ACTIVE]
+        elif posture == FM_IMAGING:
+            return stage_md[model.MD_FAV_FM_POS_ACTIVE]
+        elif posture == LOADING:
+            return stage_md[model.MD_FAV_POS_DEACTIVE]
+        elif posture == FIB_IMAGING:
+            return stage_md[model.MD_FAV_FIB_POS_ACTIVE]
+        elif posture == MILLING:
+            md = stage_md[model.MD_FAV_MILL_POS_ACTIVE]
+            rx = calculate_stage_tilt_from_milling_angle(milling_angle=md["rx"],
+                                                        pre_tilt=self.pre_tilt,
+                                                        column_tilt=self.fib_column_tilt)
+            return {"rx": rx, "rz": md["rz"]}
 
     def getTargetPosition(self, target_pos_lbl: int) -> Dict[str, float]:
         """
@@ -401,7 +497,7 @@ class MeteorPostureManager(MicroscopePostureManager):
 
     def _convert_sample_from_stage(self, val: List[float], absolute=True) -> List[float]:
         """
-        Convert values from stage to sample axes.
+        Convert values from stage to sample axes in the current posture.
         :param val: values of stage axes.
         :param absolute: if True, use absolute values otherwise relative.
         :return: values of the updated sample axes.
@@ -409,14 +505,15 @@ class MeteorPostureManager(MicroscopePostureManager):
         # stage-bare position
         Q = numpy.array(val, dtype=float)
         # Transform to coordinates in the reference frame of the sample plane
-        p = self._inv_transforms[FM_IMAGING].dot(Q)
+        posture = self.current_posture.value
+        p = self._inv_transforms[posture].dot(Q)
         if absolute:
             p -= self._O
         return p.tolist()
 
     def _convert_sample_to_stage(self, val: List[float], absolute=True) -> List[float]:
         """
-        Convert values from sample axes in FM posture to stage axes.
+        Convert values from sample axes in current posture to stage axes.
         :param val: values of sample axes.
         :param absolute: if True, use absolute values otherwise relative.
         :return: values of updated stage axes.
@@ -426,7 +523,8 @@ class MeteorPostureManager(MicroscopePostureManager):
         if absolute:
             P += self._O
         # Transform to coordinates in the reference frame of the stage-bare
-        q = self._transforms[FM_IMAGING].dot(P)
+        posture = self.current_posture.value
+        q = self._transforms[posture].dot(P)
         return q.tolist()
 
     def _update_conversion(self):
@@ -451,12 +549,54 @@ class MeteorPostureManager(MicroscopePostureManager):
         self._transforms[SEM_IMAGING] = scale_matrix @ shear_matrix @ self._get_rot_matrix(invert=True)
         self._inv_transforms[SEM_IMAGING] = numpy.linalg.inv(self._transforms[SEM_IMAGING])
 
+        # milling angle
+        self._transforms[MILLING] = self._transforms[SEM_IMAGING]
+        self._inv_transforms[MILLING] = self._inv_transforms[SEM_IMAGING]
+
+        # add unknown as same as SEM IMAGING
+        # NOTE: UKNOWN posture is not currently used for movement, but represents when the stage is
+        # not at one of the known postures. The transform is set to the same as SEM IMAGING, as that's the
+        # most common posture for the stage to be in when unknown. The transform is required to prevent the GUI
+        # from erroring when the stage is in an unknown posture when trying to display a position.
+        self._transforms[UNKNOWN] = self._transforms[SEM_IMAGING]
+        self._inv_transforms[UNKNOWN] = self._inv_transforms[SEM_IMAGING]
+
         # Offset between origins of the coordinate systems
         self._O = numpy.array(translation, dtype=float)
 
+        ###### 3D TRANSFORMS ######
+
+        # NOTE: transformations are defined as stage -> sample stage
+        # the inverse transformation is used for sample stage -> stage
+
+        # pre-tilt is rotation around the stage-bare x axis
+        rx = self.pre_tilt
+
+        # note: this is currently for tfs, which does not have scale, shear or translation
+        # rotation around x axis: fm = tf, sem = tf_inv
+        tf, tf_inv = get_rotation_transforms(rx=rx)
+
+        # NOTE: not yet implemented, see meteor-1100-fibsem-tab
+        sr = 0
+        if not self.use_scan_rotation:
+            sr = 0
+
+        # get scan rotation matrix (rz -> rx)
+        tf_sr, tf_inv_sr = get_rotation_transforms(rx=rx, rz=sr)
+        logging.debug(f"tf_matrix: {tf}, tf_sr: {tf_sr}")
+
+        self._transforms2 = {FM_IMAGING: tf,
+                             SEM_IMAGING: tf_inv_sr,
+                             MILLING: tf_inv_sr,
+                             UNKNOWN: tf_inv_sr}
+        self._inv_transforms2 = {FM_IMAGING: tf_inv,
+                                 SEM_IMAGING: tf_sr,
+                                 MILLING: tf_sr,
+                                 UNKNOWN: tf_sr}
+
     def _get_stage_pos(self, sample_val: Dict[str, float], absolute: bool = True) -> Dict[str, float]:
         """
-        Get stage coordinates from sample plane coordinates in FM posture.
+        Get stage coordinates from sample plane coordinates in current posture.
         :param sample_val: sample plane coordinates.
         :param absolute: if True, use absolute values otherwise relative.
         :return: stage coordinates.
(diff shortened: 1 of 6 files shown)


K4rishma:
Should not we merge TFS2 with TFS3 posture manager and convert into a single posture manager? TFS3 is advanced version TFS2. With this logic, we will not need TFS2.

src/odemis/acq/move.py
```diff
@@ -1007,6 +1376,92 @@ def _doCryoSwitchSamplePosition(self, future, target):
...
+                                        posture_manager=self)
+
+    def _transformFromSEMToMeteor(self, pos: Dict[str, float]) -> Dict[str, float]:
+        """
+        Transforms the current stage position from the SEM imaging area to the
+        meteor/FM imaging area.
+        :param pos: (dict str->float) the initial stage position.
+        :return: (dict str->float) the transformed position.
+        """
+        # NOTE: this transform now always rotates around the z axis (180deg)
```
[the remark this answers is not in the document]
patrickcleeve2: Combined with the raw-coordinates, saving the full position of the features (XYZRT) is enough to return to the same z-plane

patrickcleeve2:
> Should not we merge TFS2 with TFS3 posture manager and convert into a single posture manager? TFS3 is advanced version TFS2. With this logic, we will not need TFS2.

I'm not sure if any system is currently using TFS2. But technically, you need the code from the feature refactor + tfs3 to recreate the behaviour, so until that is merged too I would just keep it.

src/odemis/acq/move.py
```diff
@@ -324,10 +345,42 @@ def getCurrentPostureLabel(self, pos: Dict[str, float] = None) -> int:
...
+        if posture == SEM_IMAGING:
+            return stage_md[model.MD_FAV_SEM_POS_ACTIVE]
+        elif posture == FM_IMAGING:
+            return stage_md[model.MD_FAV_FM_POS_ACTIVE]
+        elif posture == LOADING:
+            return stage_md[model.MD_FAV_POS_DEACTIVE]
+        elif posture == FIB_IMAGING:
+            return stage_md[model.MD_FAV_FIB_POS_ACTIVE]
+        elif posture == MILLING:
+            md = stage_md[model.MD_FAV_MILL_POS_ACTIVE]
```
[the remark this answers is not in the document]
patrickcleeve2: Good point, I've added an additional explanation to the _metadata.py to explain the difference between it and the others. I think it is best to keep it defined as this, as it's waht the user expects to see (and the number they care about), as opposed to stage titl

src/odemis/acq/move.py
```diff
@@ -451,12 +506,70 @@ def _update_conversion(self):
...
+        logging.debug(f"tf_matrix: {tf}, sr_matrix: {self._sr_matrix}")
+
+        tf_sr = tf.dot(self._sr_matrix)
+        tf_inv_sr = self._sr_matrix_inv.dot(tf_inv)
+
+        self._transforms2 = {FM_IMAGING: tf,
+                             SEM_IMAGING: tf_inv_sr,
+                             MILLING: tf_inv_sr,
+                             UNKNOWN: tf_inv_sr}
+        self._inv_transforms2 = {FM_IMAGING: tf_inv,
```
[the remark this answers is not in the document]
patrickcleeve2: Yes, these will be removed once the old 2D transforms are not in use.

src/odemis/acq/move.py
```diff
@@ -324,10 +345,45 @@ def getCurrentPostureLabel(self, pos: Dict[str, float] = None) -> int:
...
+            if self.at_milling_posture(pos, stage_md):
+                return MILLING
             return SEM_IMAGING
         # None of the above -> unknown position
         return UNKNOWN
 
+    def at_milling_posture(self, pos: Dict[str, float], stage_md: Dict[str, float]) -> bool:
+        """Milling posture is not required for all meteor systems, so we need to
+        first check it's available
+        :pos the stage position
```
K4rishma, line 357: :param is missing, almost there :)

patrickcleeve2: :(

src/odemis/model/_metadata.py
```diff
@@ -228,6 +228,10 @@
 MD_FM_IMAGING_RANGE = "FM imaging range"  # dict str → [float, float] defining the volume of the FM imaging area, along x, y and z axes.
 MD_FAV_FM_POS_ACTIVE = "Favourite FM position active"  # dict str->float representing the position required for FM imaging
 MD_FAV_SEM_POS_ACTIVE = "Favourite SEM position active"  # dict -> float representing the position required for SEM imaging
+MD_FAV_FIB_POS_ACTIVE = "Favourite FIB position active"  # dict -> float representing the position required for FIB imaging
+MD_FAV_MILL_POS_ACTIVE = "Favourite Milling position active"  # dict -> float representing the position required for milling
+# NOTE: the MILL rx is defined as the angle between the FIB and the sample stage, not the stage tilt like the other metadata positions.
```
K4rishma, line 233: The milling angle is the angle measured in clockwise direction between the sample plane and FIB axis. Assuming sample plane axis is positive from left to right and FIB axis is positive from up to down.

K4rishma (approved):

src/odemis/acq/move.py
```diff
@@ -502,6 +620,253 @@ def constrain_stage_pos_axes(self, stage_val: Dict[str, float], fixed_sample_axe
         else:
             raise NotImplementedError(f"Posture {posture} not supported")
 
+    ###### DEPRECATE START #######
+    # These functions can be removed once 3D transforms are fully implemented
```
pieleric, line 624: Let's then complete the 3D transforms.

src/odemis/acq/move.py
```diff
@@ -2262,3 +2723,155 @@ def _getCurrentAlignerPositionLabel(self) -> int:
...
+        self._stage_bare = stage_bare
+
+        model.Actuator.__init__(self, name, role, dependencies={"under": stage_bare},
+                                axes=copy.deepcopy(self._stage_bare.axes), **kwargs)
+
+        # update related MDs
+        affects = ["ccd", "e-beam", "ion-beam"] # roles
+        # TODO: filter so only updates availble components
+        comps = [c.role for c in model.getComponents()]
+        self.affects = [a for a in affects if a in comps]
```
pieleric, line 2752: You should just build the list of affected components here:
```python
self.affected = []
for role in COMPS_AFFECTED_ROLES:
    try:
        self.affected.append(model.getComponent(role=role)
    except Lookup:
        pass
```

src/odemis/acq/move.py
```diff
@@ -2262,3 +2723,155 @@ def _getCurrentAlignerPositionLabel(self) -> int:
...
+            if key not in pos:
+                pos[key] = self.position.value[key]
+
+        # pos is a position, so absolute conversion
+        pos_stage = self.pm.from_sample_stage_to_stage_position(pos)
+        logging.debug("converted absolute move from %s to %s", pos, pos_stage)
+        return self._stage_bare.moveAbs(pos_stage, **kwargs)
+
+    @isasync
+    def move_vertical(self, pos: Dict[str, float]) -> Future:
```
pieleric, line 2846: Let's try to use a clearer name, following the same convention as the other methods. Something like `moveRelChamberReferential()`

src/odemis/acq/move.py
```diff
@@ -2262,3 +2723,155 @@ def _getCurrentAlignerPositionLabel(self) -> int:
...
+        """Move the stage vertically in the chamber. This is non-blocking.
+        From OpenFIBSEM"""
+        # TODO: account for scan rotation
+        theta = self._stage_bare.position.value["rx"] # tilt, in radians
+        dx = pos.get("x", 0)
+        pdy = pos.get("y", 0)
+
+        dy = pdy * math.sin(theta)
+        dz = pdy / math.cos(theta)
```
pieleric, line 2855: That  formula works for the TFS stage, but I guess it would depend on the exact stage construction (eg, depending on the relation between rx and z). It'd be worth adding a comment about this here.

pieleric (changes requested):

src/odemis/acq/move.py
```diff
@@ -25,22 +25,29 @@
...
 from odemis.util import executeAsyncTask
-from odemis.util.driver import ATOL_ROTATION_POS, isNearPosition, isInRange
-from odemis.util.transform import RigidTransform
+from odemis.util.driver import ATOL_ROTATION_POS, isInRange, isNearPosition
+from odemis.util.transform import RigidTransform, _get_transforms
+
+# feature flags
+USE_3D_TRANSFORMS = False
+USE_SCAN_ROTATION = False
+USE_LINKED_SEM_FOCUS_COMPENSATION = False
```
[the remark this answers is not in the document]
patrickcleeve2: ok

src/odemis/acq/move.py
```diff
@@ -25,22 +25,29 @@
...
+from odemis.model import isasync
 from odemis.util import executeAsyncTask
-from odemis.util.driver import ATOL_ROTATION_POS, isNearPosition, isInRange
-from odemis.util.transform import RigidTransform
+from odemis.util.driver import ATOL_ROTATION_POS, isInRange, isNearPosition
+from odemis.util.transform import RigidTransform, _get_transforms
+
+# feature flags
+USE_3D_TRANSFORMS = False
+USE_SCAN_ROTATION = False
```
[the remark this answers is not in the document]
patrickcleeve2: this is just a flag to enable the scan rotation correction. the code to calculate the transform reads the scan rotations directly

src/odemis/acq/move.py
```diff
@@ -2262,3 +2723,155 @@ def _getCurrentAlignerPositionLabel(self) -> int:
...
+                    comp.updateMetadata({
+                        model.MD_POS: md_pos,
+                        model.MD_STAGE_POSITION_RAW: pos_dep}
+                        )
+            except Exception as e:
+                logging.error("Failed to update %s with new position: %s", a, e)
+
+        # update the SEM focus position when the stage is moved to compensate for linked behavior
+        # TODO: update the self.sem_eucentric_focus when the user manually focuses.
+        if self.pm.use_linked_sem_focus_compensation:
```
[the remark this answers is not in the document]
patrickcleeve2: i've migrated to the md calib

src/odemis/acq/move.py
```diff
@@ -25,22 +25,29 @@
...
 from odemis import model, util
+from odemis.model import isasync
 from odemis.util import executeAsyncTask
-from odemis.util.driver import ATOL_ROTATION_POS, isNearPosition, isInRange
-from odemis.util.transform import RigidTransform
+from odemis.util.driver import ATOL_ROTATION_POS, isInRange, isNearPosition
+from odemis.util.transform import RigidTransform, _get_transforms
+
+# feature flags
+USE_3D_TRANSFORMS = False
```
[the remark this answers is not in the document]
patrickcleeve2: I don't think we can take such a simplified view of feature flags, especially when this is a smaller part of multiple pull requests to come. I've moved the flags into the MD_CALIB so they can be changed from the yaml file. The flags will be removed all togehter once all the features have been tested together more extensively

src/odemis/acq/move.py
patrickcleeve2: This will be removed alongside the feature flags

src/odemis/acq/move.py
```diff
@@ -2262,3 +2720,157 @@ def _getCurrentAlignerPositionLabel(self) -> int:
...
+
+        # pos is a position, so absolute conversion
+        pos_stage = self.pm.from_sample_stage_to_stage_position(pos)
+        logging.debug("converted absolute move from %s to %s", pos, pos_stage)
+        return self._stage_bare.moveAbs(pos_stage, **kwargs)
+
+    @isasync
+    def moveRelChamberReferential(self, pos: Dict[str, float]) -> Future:
+        """Move the stage vertically in the chamber. This is non-blocking.
+        From OpenFIBSEM"""
```
pieleric, line 2846: This function is quite difficult to grasp. It needs more info. In particular, when mentioning a *vertical* move, I think typically one would expect a "z" value. However, only "x" and "y" are used. This should be explained. Also, for relative moves, normally the argument is called "shift". Please use that name here too, because "pos" sounds too much like a "target position" / "absolute position".

Could you explain in the docstring how Y (in input) is related to a Z movement (in chamber referential)? Looking at the code, if rx = 0, then dy == 0 and dz == Y... which gives me the feeling that actually the two shifts input are x & z (in chamber referential). Is that correct? (if so, the key should be renamed "y" -> "z").

pieleric (changes requested):

src/odemis/acq/move.py
patrickcleeve2: I've updated it to take x,z input shift and added more info to the docstring. x,y refer to the fact that this movement is calculated form the image view. Givne where this is, it's clearer that it is x,z

src/odemis/acq/move.py
```diff
@@ -1008,6 +1379,93 @@ def _doCryoSwitchSamplePosition(self, future, target):
...
+        if not ("rz" in pos and "rz" in fm_pos_active):
+            raise ValueError(f"The stage position does not have rz axis pos={pos}, fm_pos_active={fm_pos_active}")
+
+        transformed_pos["x"] = md_calib["dx"] - pos["x"]
+        transformed_pos["y"] = md_calib["dy"] - pos["y"]
+        transformed_pos.update(fm_pos_active)
+
+        return transformed_pos
+
+    def _transformFromMeteorToSEM(self, pos: Dict[str, float]) -> Dict[str, float]:
```
tepals, line 1427: can you make the names of the methods in this class consistent? You now have `_transformFromMeteorToSEM` and `_transform_from_fm_to_fib`, so a mix of Meteor/FM and camel/snake case

src/odemis/acq/move.py
```diff
@@ -1008,6 +1379,93 @@ def _doCryoSwitchSamplePosition(self, future, target):
...
+        # NOTE: this transform now always rotates around the z axis (180deg)
+        # for pure translation, use FM -> FIB transform
+        stage_md = self.stage.getMetadata()
+        transformed_pos = pos.copy()
+        md_calib = stage_md[model.MD_CALIB]
+        sem_pos_active = stage_md[model.MD_FAV_SEM_POS_ACTIVE]
+
+        # check if the stage positions have rz axes
+        if not ("rz" in pos and "rz" in sem_pos_active):
+            raise ValueError(f"The stage position does not have rz axis. pos={pos}, sem_pos_active={sem_pos_active}")
```
tepals, line 1443: Just an fyi in case you're not aware of this feature, I think since python 3.8 (which is the lowest we support) you can also write this as 
```suggestion
            raise ValueError(f"The stage position does not have rz axis. {pos=:}, {sem_pos_active=:}")
```

tepals (changes requested):

src/odemis/acq/move.py
```diff
@@ -324,10 +342,47 @@ def getCurrentPostureLabel(self, pos: Dict[str, float] = None) -> int:
...
+            return stage_md[model.MD_FAV_FM_POS_ACTIVE]
+        elif posture == LOADING:
+            return stage_md[model.MD_FAV_POS_DEACTIVE]
+        elif posture == FIB_IMAGING:
+            return stage_md[model.MD_FAV_FIB_POS_ACTIVE]
+        elif posture == MILLING:
+            md = stage_md[model.MD_FAV_MILL_POS_ACTIVE]
+            rx = calculate_stage_tilt_from_milling_angle(milling_angle=md["rx"],
+                                                        pre_tilt=self.pre_tilt,
+                                                        column_tilt=math.radians(52))
```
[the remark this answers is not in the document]
patrickcleeve2: added constants for TFS, TESCAN, ZEISS

src/odemis/acq/move.py
```diff
@@ -451,12 +508,70 @@ def _update_conversion(self):
...
+        self._transforms2 = {FM_IMAGING: tf,
+                             SEM_IMAGING: tf_inv_sr,
+                             MILLING: tf_inv_sr,
+                             UNKNOWN: tf_inv_sr}
+        self._inv_transforms2 = {FM_IMAGING: tf_inv,
+                                 SEM_IMAGING: tf_sr,
+                                 MILLING: tf_sr,
+                                 UNKNOWN: tf_sr}
+
+    def _get_scan_rotation_matrix(self) -> Tuple[numpy.ndarray, numpy.ndarray]:
```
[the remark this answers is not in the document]
patrickcleeve2: added return values

src/odemis/acq/move.py
```diff
@@ -502,6 +617,253 @@ def constrain_stage_pos_axes(self, stage_val: Dict[str, float], fixed_sample_axe
...
+        """Convert a stage-bare position to a position in the target posture.
+        :param pos: stage position in the stage-bare coordinates
+        :param posture: (int) the target posture of the stage
+        :return: stage-bare position in the target posture"""
+
+        position_posture = self.getCurrentPostureLabel(pos)
+
+        logging.info(f"Position Posture: {POSITION_NAMES[position_posture]}, Target Posture: {POSITION_NAMES[posture]}")
+
+        self._posture_transforms = {
```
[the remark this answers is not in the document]
patrickcleeve2: I've moved it to init

src/odemis/acq/move.py
patrickcleeve2: Which way should it be? I thought new code should be snake case, and it was just old that was camel. I can change fm-> meteor

src/odemis/acq/move.py
```diff
@@ -2262,3 +2720,164 @@ def _getCurrentAlignerPositionLabel(self) -> int:
...
+        for comp in self.affects:
+            try:
+                if comp:
+                    md_pos = pos.get("x", 0), pos.get("y", 0)
+                    comp.updateMetadata({
+                        model.MD_POS: md_pos,
+                        model.MD_STAGE_POSITION_RAW: pos_dep}
+                        )
+            except Exception as e:
+                logging.error("Failed to update %s with new position: %s", a, e)
```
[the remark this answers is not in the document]
patrickcleeve2: typo

src/odemis/acq/move.py
tepals: ah I didn't realize `_transformFromMeteorToSEM` was from old code. I think we said new code is snake case, but match the style of code in a class or module if you're adjusting an existing class or module.

tepals (approved):

src/odemis/acq/move.py
```diff
@@ -301,10 +313,44 @@ def __init__(self, microscope):
...
+
+        # pre-tilt is required for milling posture, but not all systems have it
+        stage_md = self.stage.getMetadata()
+        md_calib = stage_md.get(model.MD_CALIB, {})
+        self.pre_tilt = md_calib.get(model.MD_SAMPLE_PRE_TILT, None)
+
+        # feature flags, for features still in testing
+        self.use_linked_sem_focus_compensation: bool = md_calib.get("use_linked_sem_focus_compensation", False)
+        self.use_3d_transforms: bool = md_calib.get("use_3d_transforms", False)
+        self.use_scan_rotation: bool = md_calib.get("use_scan_rotation", False)
```
pieleric, line 325: Could you add comment explaining more in details what each feature adds/changes? If I understand correctly the "scan_rotation" is to take into account the SEM/FIB scan rotation info (eg, 180°) to adjust the rotation of the sample-stage referential when in SEM/FIB posture. Is that correct? With this feature active, does the transforms SEM<>sample stage support any rotation, or only 0/180°?

The "3D transform", I don't understand. Does it mean that the "2D transform" doesn't take into account the stage tilt & sample pre-tilt? At least, in FM, this is required, and seems already handled (in `_transformFromMeteorToSEM()`), right? Does this mean that with "2D transform" in SEM & FIB views, the move along Y doesn't compensate the tilt? Or only that it doesn't handle the tilt if the sample is not perpendicular to the column? For now the code only uses Y & Z axes, which confuses me even more... How does it work currently?

src/odemis/acq/move.py
```diff
@@ -745,13 +1130,14 @@ def _doCryoSwitchSamplePosition(self, future, target):
...
                 sub_moves.append((self.stage, filter_dict({'rx', 'rz'}, target_pos)))
-            elif target in (LOADING, SEM_IMAGING, FM_IMAGING):
+            elif target in (LOADING, SEM_IMAGING, FM_IMAGING, MILLING):
                 # save rotation and tilt in SEM before switching to FM imaging
                 # to restore rotation and tilt while switching back from FM -> SEM
                 if current_label == SEM_IMAGING and target == FM_IMAGING:
-                    current_value = self.stage.position.value
-                    self.stage.updateMetadata({model.MD_FAV_SEM_POS_ACTIVE: {'rx': current_value['rx'],
-                                                                             'rz': current_value['rz']}})
+                    if not isinstance(self, MeteorTFS3PostureManager): # not required for TFS3
```
pieleric, line 1137: Can you extend the comment how this works for TFS3? SEM_IMAGING has a fixed rx and rz?

pieleric (changes requested):

src/odemis/acq/move.py
patrickcleeve2: updated to camelCase and fm->Meteor to match existing style

src/odemis/acq/move.py
patrickcleeve2: added a note for each flag, explain how/why its used and when it will be removed.

src/odemis/acq/move.py
```diff
@@ -2262,3 +2755,164 @@ def _getCurrentAlignerPositionLabel(self) -> int:
...
+        :param role: the role of the component (usually "stage")
+        :param stage_bare: the stage component to be wrapped
+        :param posture_manager: the posture manager to be used for conversion
+        :param **kwargs: additional arguments to be passed to the parent class
+        """
+
+        self._stage_bare = stage_bare
+
+        model.Actuator.__init__(self, name, role, dependencies={"under": stage_bare},
+                                axes=copy.deepcopy(self._stage_bare.axes), **kwargs)
```
pieleric, line 2778: This makes the sample stage provide 5 axes. I don't think this virtual stage should support rotations (in practice, I don't think the implementation supports any rotation). So it should only provide x, y, and z axes. 
(same for the `.position` and `.speed` VAs)

src/odemis/acq/move.py
```diff
@@ -2262,3 +2755,164 @@ def _getCurrentAlignerPositionLabel(self) -> int:
...
+        # update related MDs
+        self.affects = []
+        for role in COMPS_AFFECTED_ROLES:
+            try:
+                self.affects.append(model.getComponent(role=role))
+            except Exception:
+                pass
+
+        # posture manager to convert the positions
+        self.pm = posture_manager
```
pieleric, line 2789: Please mark this attribute as private -> `_pm`.

src/odemis/acq/move.py
```diff
@@ -2262,3 +2755,164 @@ def _getCurrentAlignerPositionLabel(self) -> int:
...
+        :param **kwargs: additional arguments to be passed to the parent class
+        """
+
+        self._stage_bare = stage_bare
+
+        model.Actuator.__init__(self, name, role, dependencies={"under": stage_bare},
+                                axes=copy.deepcopy(self._stage_bare.axes), **kwargs)
+
+        # update related MDs
+        self.affects = []
```
pieleric, line 2781: This attribute is normally provided on all `HwComponent` and should be a VA, containing the list of component *names* (not the actual component object). You're overriding it. I don't think any code uses `.affects` in the GUI, so this works... but better rename this attribute to `_affects` or `_affected_components`.

src/odemis/acq/move.py
```diff
@@ -2262,3 +2755,164 @@ def _getCurrentAlignerPositionLabel(self) -> int:
...
+                        model.MD_STAGE_POSITION_RAW: pos_dep}
+                        )
+            except Exception as e:
+                logging.error("Failed to update %s with new position: %s", comp, e)
+
+        # update the SEM focus position when the stage is moved to compensate for linked behavior
+        # TODO: update the self.sem_eucentric_focus when the user manually focuses.
+        if self.pm.use_linked_sem_focus_compensation:
+            try:
+                ebeam_focus = model.getComponent(role="ebeam-focus")
```
pieleric, line 2835: `getComponent()` is a very slow function (because it needs to connect to the back-end, possibly multiple times). So you'd want to get this component at init, and reuse it the rest of the time.

src/odemis/acq/move.py
```diff
@@ -2262,3 +2755,164 @@ def _getCurrentAlignerPositionLabel(self) -> int:
...
+        pos_stage = self.pm.from_sample_stage_to_stage_position(pos)
+        logging.debug("converted absolute move from %s to %s", pos, pos_stage)
+        return self._stage_bare.moveAbs(pos_stage, **kwargs)
+
+    @isasync
+    def moveRelChamberReferential(self, shift: Dict[str, float]) -> Future:
+        """Move the stage vertically in the chamber. This is non-blocking.
+        From OpenFIBSEM.
+        The desired input shift (x, z) is transformed to x, y, z axis components such that the
+        the stage moves in the vertical direction in the chamber. For TFS, the z-axis is
```
pieleric, line 2883: Please be even more specific about the behaviour. Is the "input shift" in the sample stage coordinates? After the move, where will appear (in the FoV) the point originally at the center of the FoV? It will be shifted along X by `x`, but what about `y` and `z`? Does it just behave as if there was a "z" axis perpendicular to the sample? If so, it should be part of the standard `moveRel()`...
Also, in which posture will this move work? SEM imaging only? FIB milling?

pieleric (changes requested):

src/odemis/acq/move.py
patrickcleeve2: the virtual stage can still rotate and tilt (even as a function of the underlying stage-bare), even if it can't do it itself. if you only have xyz, you can't fully define the position so i don't understand why we wouldn't include it?

src/odemis/acq/move.py
```diff
@@ -2262,3 +2755,164 @@ def _getCurrentAlignerPositionLabel(self) -> int:
...
+        attached to the tilt, so the tilt angle must be taken into account.
+        For Tescan systems, the z-axis is always vertical, so this function is not required.
```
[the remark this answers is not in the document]
patrickcleeve2: moved it inside the posture manager

src/odemis/acq/move.py
patrickcleeve2: updated

src/odemis/acq/move.py
```diff
@@ -570,19 +570,23 @@ def _update_conversion(self):
...
 
-        logging.debug(f"tf_matrix: {tf}, sr_matrix: {sr_matrix}")
+        # get scan rotation matrix (rz -> rx)
+        tf_sr, tf_inv_sr = get_rotation_transforms(rx=rx, rz=sr)
+        logging.debug(f"tf_matrix: {tf}, tf_sr: {tf_sr}")
 
-        tf_sr = tf.dot(sr_matrix)
-        tf_inv_sr = sr_matrix_inv.dot(tf_inv)
+        # tmp: assert that scan rotation matrix has no impact on standard tf, remove once scna rotation implemented
+        numpy.testing.assert_array_equal(tf, tf_sr)
```
pieleric, line 589: I don't think you need to check for this every time it's computed. The code uses explicitly `rz=0`. Looking at the code of `get_rotation_transforms()`, you can see it's converted into an identity matrix. So the 2 matrices will be identical, for sure. Let's drop this assert code.

src/odemis/acq/move.py
```diff
@@ -2871,31 +2885,29 @@ def moveAbs(self, pos: Dict[str, float], **kwargs) -> Future:
                 pos[key] = self.position.value[key]
 
         # pos is a position, so absolute conversion
-        pos_stage = self.pm.from_sample_stage_to_stage_position(pos)
+        pos_stage = self._pm.from_sample_stage_to_stage_position(pos)
         logging.debug("converted absolute move from %s to %s", pos, pos_stage)
         return self._stage_bare.moveAbs(pos_stage, **kwargs)
 
     @isasync
     def moveRelChamberReferential(self, shift: Dict[str, float]) -> Future:
```
pieleric, line 2893: Sorry, I found out that "referential" is not the usual term in English :see_no_evil: . It should be more like "reference frame" or "coordinate system". Let's rename that function to `moveRelChamberCoordinates`... or maybe you have a better suggestion?

src/odemis/acq/move.py
```diff
@@ -2871,31 +2885,29 @@ def moveAbs(self, pos: Dict[str, float], **kwargs) -> Future:
...
+        """Move the stage vertically in the chamber. This is non-blocking. From OpenFIBSEM.
         The desired input shift (x, z) is transformed to x, y, z axis components such that the
-        the stage moves in the vertical direction in the chamber. For TFS, the z-axis is
-        attached to the tilt, so the tilt angle must be taken into account.
-        For Tescan systems, the z-axis is always vertical, so this function is not required.
+        the stage moves in the vertical direction in the chamber.
+        The input shift is expected in the sample-stage coordinates. A feature in the FIB FoV is
+        expected to be moved by dx in the x-image coordinates, and dz in the y-image coordinates. The
+        feature will not be moved in the SEM FoV. To achieve this, we compute the stage-bare movement such
+        that the stage moves vertically in the chamber, resulting in the desired behavior in the SEM/FIB FoV.
```
pieleric, line 2900: Ok. Thanks for clarifying. I think this is now precise enough. Just for me, in practice for now, this function is only used to align the coincidental point for SEM/FIB, when in SEM imaging posture, right? Or also maybe in FIB imaging? Never in FLM posture, or even MILLING, right?

src/odemis/acq/move.py
pieleric: This actuator is supposed to be in the sample coordinates. So any rotation should happen relative to the standard rotation required for the posture. With the current implementation, that's not the case, the rotation is passed as-is to the stage-bare.  So I'd rather not support it than moving in the wrong way. Also, I don't think anything in the GUI requires rotation in the sample coordinates. Rotation should be need only to change posture (and for the milling angle, but that also should be handled via the Posture Manager/stage-bare). 

So I see it as a "safety feature" to *not* support rotation in the sample stage.

pieleric (changes requested):
Looking good. Just a few minor comments.

src/odemis/acq/move.py
patrickcleeve2: yes just used to align coincidence of sem/fib, and will work at any posture (SEM, FIB, MILL). it will do the same thing at FM, but theres no coincidence for meteor so in practice you'd never use it there.

src/odemis/acq/move.py
patrickcleeve2: ok, removed the rotation axes

pieleric (approved):

PR #3008, opened 2025-01-23
[fix] driver pwrmccdaq: simulate better connection of DIO ports A <-> B

pieleric:
The simulator simulates port A being connected to port B. So writting to
a bit on port A causes the equivalent bit on port B to read the same
value (and inversely too).

This worked, but it broke the possibility to read what value is
currently set. => write both to port A & B.

Also *document* this behaviour.

Also, as the test cases relied on such behaviour on the real hardware (by
physically connecting some pins together). The tests also work on the
simulator. So no need to disable it... However that test was broken, so
fix it too.

### DIFF
--- a/src/odemis/driver/pwrmccdaq.py
+++ b/src/odemis/driver/pwrmccdaq.py
@@ -69,13 +69,13 @@ class MCCDevice(HwComponent):
             the device is considered a USB-powered MCC DAQ 1208LS device. Note that the serial number
             should contain an added 0 to the beginning of the s/n. If "fake" is passed the simulator
             is used.
-        :param di_channels (dict -> int, list(str, bool)):
+        :param di_channels (dict: int -> list(str, bool)):
             the DIO channel used for TTL signal status change through the MCC device.
-            for example di_channels: {2: ["interlockTriggered", False], 7: ["leftSwitch", True]}.
-            key is the channel number(int), value is a list with VA name(str) and a flag stating
-                if ttl high is True(bool), this means that the VA will be considered True or False
-                when the TTL signal is high.
-            with this approach it is actually possible to keep track of customized status changes.
+            Example: {2: ["interlockTriggered", False], 7: ["leftSwitch", True]}.
+            key is the DI channel number.
+            value is a list with VA name (str) and a "TTL high" flag, stating the value to set when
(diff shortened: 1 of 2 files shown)


src/odemis/driver/pwrmccdaq.py
```diff
@@ -403,19 +408,16 @@ def DConfig(self, port_number, bit_mask):
         if bit_mask < 0x00 or bit_mask > 0xff:
             raise ValueError("Bit mask to set is not between 0 and 255")
 
-        if port_number == usb_1208LS.DIO_PORTA:
-            self.port_a_bit_config = [bool(bit_mask & 1 << i) for i in range(8)]
-        elif port_number == usb_1208LS.DIO_PORTB:
-            self.port_b_bit_config = [bool(bit_mask & 1 << i) for i in range(8)]
-        else:
-            raise ValueError()
+        if port_number not in self.port_config:
```
nandishjpatel, line 411: In the function's docstring `:param port_number: AUXPORT = 0x10 | Port A = 0x01 | Port B = 0x04`, should port number `AUXPORT` also be handled?

src/odemis/driver/pwrmccdaq.py
```diff
@@ -427,16 +429,13 @@ def DOut(self, port_number, value):
...
-        else:
-            raise ValueError()
+        DIO_port = self.port_status[port_number]
+        bit_config = self.port_config[port_number]
+        DIO_port[:] = [bool(value & 1 << i) if not bit_config[i] else DIO_port[i] for i in range(8)]
 
-        # when port A is requested, writing to port B is simulated and the other way around
+        # Simulate connection to the other port
+        other_port = usb_1208LS.DIO_PORTA if port_number == usb_1208LS.DIO_PORTB else usb_1208LS.DIO_PORTB
+        DIO_port = self.port_status[other_port]
```
K4rishma, line 438: Line 438 is overriding the value of line 434. I am not sure if that is intentional.

K4rishma (approved):

nandishjpatel (approved):

src/odemis/driver/pwrmccdaq.py
pieleric: Yes, it's the same code, to show we do the same thing (for simulation).

PR #3009, opened 2025-01-23
[NSRPARMA-131][feat] gui sparc: if laser interlock is triggered, also blank the beam

pieleric:
One of the most common reason for the laser interlock to be triggered
is that the user opened the SPARC ATM cover. The inside of the ATM can
receive X-rays caused by the e-beam. So, to be extra safe, also blank
the e-beam whenever the laser interlock is triggered.

### DIFF
--- a/src/odemis/gui/cont/acquisition/sparc_acq.py
+++ b/src/odemis/gui/cont/acquisition/sparc_acq.py
@@ -67,6 +67,12 @@ class SparcAcquiController(object):
         self._tab_panel = tab_panel
         self._streambar_controller = streambar_controller
         self._interlockTriggered = False  # local/private bool to track interlock status
+        if self._main_data_model.ebeam and model.hasVA(self._main_data_model.ebeam, "blanker"):
+            # blanker state before interlock (to restore it when interlock is reset)
+            self._ebeam_blanker = self._main_data_model.ebeam.blanker
+            self._pre_interlock_blanker = self._ebeam_blanker.value
+        else:
+            self._ebeam_blanker = None
 
         if model.hasVA(self._main_data_model.light, "interlockTriggered"):
             # subscribe to the VA and initialize the warning status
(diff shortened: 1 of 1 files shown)


src/odemis/gui/cont/acquisition/sparc_acq.py
```diff
@@ -229,19 +235,33 @@ def on_interlock_change(self, value):
...
+                ebeam_str = "and e-beam "
+
             popup.show_message(wx.GetApp().main_frame,
                                title="Laser safety",
-                               message=f"Laser was suspended automatically due to interlock trigger.",
+                               message=f"Laser {ebeam_str}were suspended automatically due to interlock trigger.",
                                timeout=10.0,
                                level=logging.WARNING)
         else:
+            # Reactivate the e-beam blanker, but only if it is still active (otherwise, it's a sign
```
nandishjpatel, line 255:
```suggestion
            # Unblank the e-beam blanker, but only if it is still active (otherwise, it's a sign
```

src/odemis/gui/cont/acquisition/sparc_acq.py
```diff
@@ -229,19 +235,33 @@ def on_interlock_change(self, value):
...
+            # that the user is manually handling it)
+            if self._ebeam_blanker and self._ebeam_blanker.value:
+                self._ebeam_blanker.value = self._pre_interlock_blanker
+                if self._pre_interlock_blanker != True:
+                    ebeam_str = " E-beam blanker was reactivated."
```
nandishjpatel, line 260:
```suggestion
                    ebeam_str = " E-beam was unblanked."
```
Maybe better

nandishjpatel (approved):

src/odemis/gui/cont/acquisition/sparc_acq.py
pieleric: "Reactivate the e-beam blanker" is not correct, indeed. However "unblank the [...] blanker" is confusing too!
It's tricky because the blanker has 3 states: e-beam blanked, e-beam unblanked, and "auto" (= e-beam blanked if not scanning). 
I'll try to find a better English wording....

src/odemis/gui/cont/acquisition/sparc_acq.py
```diff
@@ -229,19 +235,33 @@ def on_interlock_change(self, value):
...
+            # For the e-beam (to protect the user from X-rays), we force the blanker.
+            if self._ebeam_blanker:
+                self._pre_interlock_blanker = self._ebeam_blanker.value
+                self._ebeam_blanker.value = True
+                ebeam_str = "and e-beam "
+
             popup.show_message(wx.GetApp().main_frame,
                                title="Laser safety",
-                               message=f"Laser was suspended automatically due to interlock trigger.",
+                               message=f"Laser {ebeam_str}were suspended automatically due to interlock trigger.",
```
tepals, line 251: A minor comment, but if there is no `ebeam_str` the sentence should be `Laser was suspended` and not `Laser were suspended`. It's fine by me if you leave it like this, though.

tepals (approved):

src/odemis/gui/cont/acquisition/sparc_acq.py
pieleric: I've separated into 2 messages, depending whether the blanker is disabled or set to auto (None).

src/odemis/gui/cont/acquisition/sparc_acq.py
pieleric: I've stopped trying to create fancy sentences. It's not two different sentences, in a `if`.

tepals (approved):

PR #3011, opened 2025-01-23
[TRCL-265] [feat] acq SPARC partial acquisition and streakcam live protection

pieleric:
Every time an streakcam image is received, check that the intensity level is not too high. If too many pixels are too high, with a threshold defined in the microscope file, then the acquisition is stopped (and the data acquired so far is saved).

In order for the data to be still stored in case the acquisition is stopped in the middle, introduce for all SPARC acquisitions support for partial acquisition. => Change the API, in a similar way as the acqmng API: the Future always returns both the data acquired and the exception. If the acquisition went fine, the exception is None.
 
If an exception occurred, the GUI will show the error message, and store the data in a file named "filename-failed...", to make it clear it didn't work as expected.

### DIFF
--- a/src/odemis/acq/stream/_sync.py
+++ b/src/odemis/acq/stream/_sync.py
@@ -3127,6 +3127,7 @@ class SEMAngularSpectrumMDStream(SEMCCDMDStream):
 
         self._raw.extend(data)
 
+STREAK_CCD_INTENSITY_MAX_PX_COUNT = 10  # px, max number of pixels allowed above the threshold
 
 class SEMTemporalSpectrumMDStream(SEMCCDMDStream):
     """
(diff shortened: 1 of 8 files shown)


src/odemis/acq/stream/_sync.py
```diff
@@ -3142,13 +3143,41 @@ def _runAcquisition(self, future) -> Tuple[List[model.DataArray], Optional[Excep
...
+                # stream has exposure time
+                exp = self._sccd._getDetectorVA("exposureTime").value  # s
+
+            # Clip to the maximum value of the detector, otherwise we might never detect high intensity
+            # for long exposure times. Should be defined on the streak-ccd.metadata[MD_CALIB]["intensity_limit"]
+            # as count/s.
+            max_det_value = self._ccd.shape[2] - 1
+            ccd_md = self._ccd.getMetadata()
+
+            intensity_limit_s = ccd_md.get(model.MD_CALIB, {}).get("intensity_limit", 40000)
```
tepals, line 3161: why is this called `intensity_limit_s` if it is supposed to be specified in counts/s?

src/odemis/acq/stream/_sync.py
```diff
@@ -3142,13 +3143,41 @@ def _runAcquisition(self, future) -> Tuple[List[model.DataArray], Optional[Excep
...
+            self._intensity_limit = min(intensity_limit_s * exp, max_det_value)
```
tepals, line 3162: I find it a bit confusing that the metadata value is in counts/s, while this value (the value you store) is in counts.

src/odemis/acq/stream/_sync.py
```diff
@@ -3142,13 +3143,41 @@ def _runAcquisition(self, future) -> Tuple[List[model.DataArray], Optional[Excep
...
+            logging.debug("Streak CCD intensity threshold set to %s", self._intensity_limit)
```
tepals, line 3163: can you add a unit to this log?

tepals (changes requested):

src/odemis/driver/simstreakcam.py
```diff
@@ -278,7 +278,13 @@ def _getNewImage(self):
...
             mx = 1000
         else:
-            mx = image.max()
+            mx = image.max() * 5
+
+        # Increase the image intensity based on the MCP gain.
+        # 0 -> x1
+        # 63 (max) -> x6.4
+        gain_factor = (self.parent._streakunit.MCPGain.value + 1) / 10
+        numpy.multiply(image, gain_factor, out=image, casting="unsafe")
```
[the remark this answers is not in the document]
pieleric: Same behaviour, but less memory use. `out=image` reuse the already allocated array. `image = ...` allocates a new array (of the same size), and eventually the old array is freed.

src/odemis/acq/test/stream_test.py
```diff
@@ -4244,11 +4290,16 @@ def test_ebic_acq(self):
...
         # wait until it's over
-        data = f.result(timeout)
+        data, exp = f.result(timeout)
+        self.assertIsNone(exp)
         dc.series_complete(data)
         dur = time.time() - start
         logging.debug("Acquisition took %g s", dur)
         self.assertTrue(f.done())
+        for da in sms.raw:
+            print(f"raw shape {da.shape}, dtype {da.dtype}, metadata {da.metadata.get(model.MD_DESCRIPTION)}")
```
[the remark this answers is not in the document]
pieleric: I'll just drop these `print()`. They were just for debugging.

src/odemis/acq/stream/_sync.py
pieleric: Because it's "[counts per] s[econd]" . => Will change to  `intensity_limit_cps` & `intensity_limit_cpf`.

tepals (approved):

patrickcleeve2 (approved):

PR #3013, opened 2025-01-24
[fix] MeasureSpotsFocus() only worked with specific dtypes

pieleric:
MeasureSpotsFocus() uses OpenCV functions, which only support some
specific dtypes. In particular, they don't handle uint32, which some
detectors generates.

In such case, this error happens (during autofocus):
```
File "/usr/lib/python3/dist-packages/odemis/util/focus.py", line 123, in MeasureSpotsFocus
    sobelx = cv2.Sobel(image, cv2.CV_64F, 1, 0, ksize=5)
cv2.error: OpenCV(4.5.4) :-1: error: (-5:Bad argument) in function 'Sobel'
> Overload resolution failed:
>  - src data type = 6 is not supported
>  - Expected Ptr<cv::UMat> for argument 'src'
```

=> Detect the image dtype is not supported and convert to float64 before
calling OpenCV.

### DIFF
--- a/src/odemis/util/focus.py
+++ b/src/odemis/util/focus.py
@@ -70,6 +70,12 @@ def MeasureOpticalFocus(image):
         # TODO find faster/better solution
         image = _convertRBGToGrayscale(image)
 
+    # TODO: maybe switch to scipy.ndimage.laplace ?
+    # OpenCV only supports int of 8 & 16 bits and float32/64 images. So if we get anything else,
+    # we first convert to float64, to make sure it's compatible.
+    if image.dtype not in (numpy.int8, numpy.uint8, numpy.int16, numpy.uint16, numpy.float32, numpy.float64):
+         image = image.astype(numpy.float64)
+
     return cv2.Laplacian(image, cv2.CV_64F).var()
 
 
@@ -120,6 +126,11 @@ def MeasureSpotsFocus(image):
     image (model.DataArray): Optical image
     returns (float): The focus level of the image (higher is better)
     """
+    # TODO: maybe switch to scipy.ndimage.sobel ?
+    # OpenCV only supports int of 8 & 16 bits and float32/64 images. So if we get anything else,
+    # we first convert to float64, to make sure it's compatible.
+    if image.dtype not in (numpy.int8, numpy.uint8, numpy.int16, numpy.uint16, numpy.float32, numpy.float64):
+         image = image.astype(numpy.float64)
     sobelx = cv2.Sobel(image, cv2.CV_64F, 1, 0, ksize=5)
     sobely = cv2.Sobel(image, cv2.CV_64F, 0, 1, ksize=5)
     sobel_image = sobelx ** 2 + sobely ** 2
--- a/src/odemis/util/test/focus_test.py
+++ b/src/odemis/util/test/focus_test.py
@@ -1,0 +1,68 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+Created on 24 January 2025
+
+@author: Éric Piel
+
+Copyright © 2025 Éric Piel, Delmic
+
+This file is part of Odemis.
+
+Odemis is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License version 2 as published by the Free Software Foundation.
+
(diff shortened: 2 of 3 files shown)


src/odemis/util/focus.py
```diff
@@ -120,6 +120,13 @@ def MeasureSpotsFocus(image):
     image (model.DataArray): Optical image
     returns (float): The focus level of the image (higher is better)
     """
+    # OpenCV only supports int of 8 & 16 bits and float32/64 images. So if we get anything else,
+    # we first convert to float64, to make sure it's compatible.
+    if not (
+            (numpy.issubdtype(image.dtype, numpy.integer) and image.dtype.itemsize <= 2)
+            or image.dtype in (numpy.float32, numpy.float64)
+        ):
+         image = image.astype(numpy.float64)
```
tepals, line 129: I think this might also be a problem in `MeasureOpticalFocus`. If you agree could you add the check there as well?

src/odemis/util/test/focus_test.py
```diff
@@ -0,0 +1,57 @@
...
+Odemis is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License version 2 as published by the Free Software Foundation.
+
+Odemis is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with Odemis. If not, see http://www.gnu.org/licenses/.
+"""
+
+import unittest
+import numpy
+from odemis.util.focus import MeasureSpotsFocus
```
tepals, line 21:
```suggestion
import numpy
import unittest

from odemis.util.focus import MeasureSpotsFocus
```

src/odemis/util/test/focus_test.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+class TestMeasureSpotsFocus(unittest.TestCase):
```
tepals, line 24: Can you add running these test cases to the cloud hosted ci? https://github.com/delmic/odemis/blob/master/.github/workflows/unit_tests_cloud.yml

tepals (changes requested):

src/odemis/util/focus.py
pieleric: Indeed.
Or maybe we should switch to scipy? 
See:
scipy.ndimage.sobel()
scipy.ndimage.laplace()

At *first* look they seem equivalent, and available already on Ubuntu 20.04 (Scipy 1.3.3):
https://docs.scipy.org/doc/scipy-1.3.1/reference/generated/scipy.ndimage.laplace.html

pieleric: As we want to have it merged soon, I've just extended the change as-is to MeasureOpticalFocus(). We can later think of switching to scipy.

tepals: I vaguely remember that the scipy implementation was a lot slower. But this was 5 years ago, so maybe it has improved. 

I agree with leaving the change as is and thinking later of switching to scipy

tepals (approved):

src/odemis/util/focus.py
```diff
@@ -70,6 +70,15 @@ def MeasureOpticalFocus(image):
...
         image = _convertRBGToGrayscale(image)
 
+    # TODO: maybe switch to scipy.ndimage.laplace ?
+    # OpenCV only supports int of 8 & 16 bits and float32/64 images. So if we get anything else,
+    # we first convert to float64, to make sure it's compatible.
+    if not (
+            (numpy.issubdtype(image.dtype, numpy.integer) and image.dtype.itemsize <= 2)
+            or image.dtype in (numpy.float32, numpy.float64)
+        ):
+         image = image.astype(numpy.float64)
```
tmoerkerken, line 80: Seems a bit strange to check if dtype is float64 and than set the dtype to float64. Am I missing something?

pieleric: I think you skipped the `not` at the beginning of the `if`.

pieleric: Maybe I should just make the check "dumber", by listing all supported dtypes?
`if image.dtype not in (numpy.int8, numpy.uint8, numpy.int16, numpy.uint16, numpy.float32, numpy.float64)`

tmoerkerken: Correct, I didn't read it correctly, but it reads easier after your change :+1:

tmoerkerken (approved):

PR #3014, opened 2025-01-24
[SDEV-1958] extend tiled acquisition code to be able to acquire non rectangular shapes

nandishjpatel:
The code changes are similar to that of FASTEM ROA which is represented by points. One difference is that `TiledAcquisitionTask` supports both a bound box and points as an input region and is backwards compatible.

Attached an image for overview image acquisition of a rotated rectangle
Screenshot from 2025-01-24 14-25-46

### DIFF
--- a/src/odemis/acq/stitching/_tiledacq.py
+++ b/src/odemis/acq/stitching/_tiledacq.py
@@ -25,26 +25,45 @@ import statistics
 import threading
 import time
 from concurrent.futures import CancelledError, TimeoutError
-from concurrent.futures._base import RUNNING, FINISHED, CANCELLED
+from concurrent.futures._base import CANCELLED, FINISHED, RUNNING
 from enum import Enum
-from typing import List, Tuple, Optional, Dict
+from itertools import groupby
+from typing import Dict, List, Optional, Tuple, Union
 
 import numpy
 import psutil
+from scipy.ndimage import binary_fill_holes
 from scipy.spatial import Delaunay
+from shapely.geometry import Polygon, box
 
-from odemis import model, dataio
+from odemis import dataio, model
 from odemis.acq import acqmng
-from odemis.acq.align.autofocus import AutoFocus, MTD_EXHAUSTIVE
-from odemis.util.focus import MeasureOpticalFocus
-from odemis.acq.align.roi_autofocus import autofocus_in_roi, estimate_autofocus_in_roi_time
-from odemis.acq.stitching._constants import WEAVER_MEAN, REGISTER_IDENTITY, REGISTER_GLOBAL_SHIFT
+from odemis.acq.align.autofocus import MTD_EXHAUSTIVE, AutoFocus
+from odemis.acq.align.roi_autofocus import (
+    autofocus_in_roi,
+    estimate_autofocus_in_roi_time,
+)
+from odemis.acq.stitching._constants import (
+    REGISTER_GLOBAL_SHIFT,
+    REGISTER_IDENTITY,
+    WEAVER_MEAN,
+)
 from odemis.acq.stitching._simple import register, weave
-from odemis.acq.stream import Stream, EMStream, ARStream, \
-    SpectrumStream, FluoStream, MultipleDetectorStream, util, executeAsyncTask, \
-    CLStream
+from odemis.acq.stream import (
+    ARStream,
+    CLStream,
+    EMStream,
+    FluoStream,
+    MultipleDetectorStream,
+    SpectrumStream,
+    Stream,
+    executeAsyncTask,
+    util,
+)
 from odemis.model import DataArray
-from odemis.util import dataio as udataio, img, linalg, rect_intersect
+from odemis.util import dataio as udataio
+from odemis.util import img, linalg, rect_intersect
+from odemis.util.focus import MeasureOpticalFocus
 from odemis.util.img import assembleZCube
 from odemis.util.linalg import generate_triangulation_points
 from odemis.util.raster import point_in_polygon
@@ -81,13 +100,15 @@ class TiledAcquisitionTask(object):
     The goal of this task is to acquire a set of tiles then stitch them together
     """
 
-    def __init__(self, streams, stage, area, overlap, settings_obs=None, log_path=None, future=None, zlevels=None,
+    def __init__(self, streams, stage, region, overlap, settings_obs=None, log_path=None, future=None, zlevels=None,
                  registrar=REGISTER_GLOBAL_SHIFT, weaver=WEAVER_MEAN, focusing_method=FocusingMethod.NONE,
                  focus_points=None, focus_range=None):
         """
         :param streams: (list of Streams) the streams to acquire
         :param stage: (Actuator) the sample stage to move to the possible tiles locations
-        :param area: (float, float, float, float) left, top, right, bottom points of acquisition area
+        :param region: Either:
+            - (Tuple[float, float, float, float]) Bounding box as (xmin, ymin, xmax, ymax)
+            - (List[Tuple[float, float]]) List of (x, y) points defining a polygon
         :param overlap: (float) the amount of overlap between each acquisition
         :param settings_obs: (SettingsObserver or None) class that contains a list of all VAs
             that should be saved as metadata
@@ -111,19 +132,16 @@ class TiledAcquisitionTask(object):
         self._focus_range = focus_range
         # Average time taken for each tile acquisition
         self.average_acquisition_time = None
+        self._overlap = overlap
+        self._polygon = None
         if future is not None:
             self._future.running_subf: future.Future = model.InstantaneousFuture()
             self._future._task_lock = threading.Lock()
-        # Get total area as a tuple of width, height from ltrb area points
-        normalized_area = util.normalize_rect(area)
-        if area[0] != normalized_area[0] or area[1] != normalized_area[1]:
-            logging.warning("Acquisition area {} rearranged into {}".format(area, normalized_area))
 
-        self._area = normalized_area
-        height = normalized_area[3] - normalized_area[1]
-        width = normalized_area[2] - normalized_area[0]
-        self._area_size = (width, height)
-        self._overlap = overlap
+        # Normalize the region input into a polygon
+        self._polygon = self._normalize_region_to_polygon(region)
+        xmin, ymin, xmax, ymax = self._polygon.bounds
+        self._area_size = (xmax - xmin, ymax - ymin)
 
         # Note: we used to change the stream horizontalFoV VA to the max, if available (eg, SEM).
         # However, it's annoying if the caller actually cares about the FoV (eg,
@@ -138,7 +156,8 @@ class TiledAcquisitionTask(object):
         self._sfov = self._guessSmallestFov(streams)
         logging.debug("Smallest FoV: %s", self._sfov)
 
-        (self._nx, self._ny), self._starting_pos = self._getNumberOfTiles()
+        self._number_of_tiles, self._starting_pos, self._tile_indices = self._getNumberOfTiles()
+        logging.debug("Calculated number of tiles %s", self._number_of_tiles)
 
         # save actual time take for tile acquisition, stage movement from one tile to another
         # and stitching time for each tile. The keys tell the action, i.e. move or stitch and the time
@@ -222,6 +241,35 @@ class TiledAcquisitionTask(object):
         self._weaver = weaver
         self._focus_plane = {}
 
+    def _normalize_region_to_polygon(
+            self,
+            region: Union[Tuple[float, float, float, float], List[Tuple[float, float]]]
+        ) -> Polygon:
+        """
+        Normalize the region input into a polygon.
+        :param region: Either a bounding box (xmin, ymin, xmax, ymax) or a list of (x, y) points.
+        :return: A polygon representing the region.
+        :raise: ValueError if the region does not form a valid polygon.
+        :raise: ValueError if the region is not a bounding box or a list of points.
+        """
+        if isinstance(region, tuple) and len(region) == 4:
+            # Normalize the bounding box
+            xmin, ymin, xmax, ymax = util.normalize_rect(region)
+            if region[0] != xmin or region[1] != ymin:
+                logging.warning("Acquisition area %s rearranged into %s", region, (xmin, ymin, xmax, ymax))
+            points = [(xmin, ymin), (xmax, ymin), (xmax, ymax), (xmin, ymax)]
+        elif isinstance(region, list) and all(isinstance(point, tuple) and len(point) == 2 for point in region):
+            points = region
+        else:
+            raise ValueError(
+                "Region must either be a bounding box (xmin, ymin, xmax, ymax) or a list of (x, y) points."
+            )
+
+        polygon = Polygon(points)
+        if not polygon.is_valid:
+            raise ValueError("The provided region does not form a valid polygon.")
+        return polygon
+
     def _getFov(self, sd):
         """
         sd (Stream or DataArray): If it's a stream, it must be a live stream,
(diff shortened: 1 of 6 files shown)


src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -214,6 +232,35 @@ def __init__(self, streams, stage, area, overlap, settings_obs=None, log_path=No
         self._weaver = weaver
         self._focus_plane = {}
 
+    def _normalize_region_to_polygon(
+            self,
+            region: Union[Tuple[float, float, float, float], List[Tuple[float, float]]]
+        ) -> Polygon:
+        """
+        Normalize the region input into a polygon.
```
tepals, line 240: Can you extend this comment a bit. Looking it at the code it "normalizes" (the region if it is a rectangle and doesn't do anything if it's a polygon.

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -267,22 +315,85 @@ def _getNumberOfTiles(self):
...
 
         # We have a little bit more tiles than needed, we then have two choices
         # on how to spread them:
         # 1. Increase the total area acquired (and keep the overlap)
         # 2. Increase the overlap (and keep the total area)
         # We pick alternative 1 (no real reason)
-        center = (self._area[0] + self._area[2]) / 2, (self._area[1] + self._area[3]) / 2
+        xmin, ymin, xmax, ymax = self._polygon.bounds
+        center = ((xmin + xmax) / 2, (ymin + ymax) / 2)
         total_size = nx * reliable_fov[0], ny * reliable_fov[1]
```
tepals, line 326: Not completely sure if it matters here, but the total size should include one overlap as well. Since for the total width you have one full tile and (n-1) overlapping tiles
```suggestion
        total_size = nx * reliable_fov[0] + self._sfov[0] * self._overlap, 
                     ny * reliable_fov[1] + self._sfov[1] * self._overlap
```

src/odemis/acq/stitching/test/tiledacq_test.py
```diff
@@ -95,77 +111,54 @@ def test_get_number_of_tiles(self):
...
         # Smaller than FoV => 1x1
         tiled_acq_task = TiledAcquisitionTask(self.fm_streams, self.stage,
-                                              area=(0, 0, fov[0] / 2, fov[1] / 2),
+                                              (0, 0, fov[0] / 2, fov[1] / 2),
                                               overlap=0, future=model.InstantaneousFuture())
 
-        num_tiles, starting_pos = tiled_acq_task._getNumberOfTiles()
-        self.assertEqual(num_tiles, (1, 1))
+        num_tiles, _, _ = tiled_acq_task._getNumberOfTiles()
+        self.assertEqual(num_tiles, 1)
```
tepals, line 118: I think it would make sense for all the tests in this test case to actually test the indices returned.

src/odemis/acq/stitching/test/tiledacq_test.py
```diff
@@ -761,5 +754,108 @@ def test_get_zstack_levels(self):
         zlevels = get_zstack_levels(zsteps=10, zstep_size=2e-6, rel=False, focuser=focuser)
         self.assertAlmostEqual(zlevels[0], frange[0]) # minimum z level should be at minimum position
 
+
+class TiledAcquisitionTaskTestCase(unittest.TestCase):
```
tepals, line 758: Can you add running these test cases to the unit_test_cloud.yml?

tepals (changes requested):

src/odemis/acq/stitching/_tiledacq.py
nandishjpatel: I will rename the function to `_convert_region_to_polygon`

src/odemis/acq/stitching/_tiledacq.py
nandishjpatel: I am also not sure here. @pieleric can you help?

nandishjpatel:
@tepals and @pieleric I have resolved the merge conflicts in this PR, can you please give it another look?

src/odemis/acq/stitching/_tiledacq.py
pieleric: Indeed, Thera's remark sounds correct. For instance, with just 2 tiles, you'd have 2 full FoVs - 1 overlap. This correctly computed by Thera's formula.

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -275,22 +322,85 @@ def _getNumberOfTiles(self):
...
-        # Compute the top-left of the "bigger" area, and from it, shift by half
-        # the size of the smallest tile.
-        starting_pos = {'x': center[0] - total_size[0] / 2 + reliable_fov[0] / 2,  # left
-                        'y': center[1] + total_size[1] / 2 - reliable_fov[1] / 2}  # top
-
-        return (nx, ny), starting_pos
+        xmin = center[0] - total_size[0] / 2
+        ymax = center[1] + total_size[1] / 2
+
+        # Create an empty grid for storing intersected tiles
```
pieleric, line 337: Can you explain a tiny bit more what "intersected tile" mean? From the context I imagine it's the tiles which are within the polygon? Is that correct?

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -275,22 +322,85 @@ def _getNumberOfTiles(self):
...
+        tile_grid = numpy.zeros((ny, nx), dtype=bool)
+
+        def get_possible_intersected_tiles(row_pairs, col_pairs):
```
pieleric, line 340: This function is already long enough. No need to inline another function inside it. Can you move it as a normal method, or maybe move it somewhere into odemis.util.raster (or another module in util).

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -275,22 +322,85 @@ def _getNumberOfTiles(self):
...
+                    e2 = err * 2
+                    if e2 > -dy:
+                        err -= dy
+                        row1 += sx
+                    if e2 < dx:
+                        err += dx
+                        col1 += sy
+            return tiles
+
+        # Vectorized conversion of polygon points to grid coordinates
```
pieleric, line 368: I think this whole algorithm that computes which tile needs to be scanned should go into a dedicated function. That would make the intention much clearer (than in a function supposed to "just" count the number of tiles).

pieleric (changes requested):

src/odemis/acq/stitching/_tiledacq.py
nandishjpatel: That's correct, added an extra comment line

src/odemis/acq/stitching/_tiledacq.py
nandishjpatel: Renamed the function, and later find the number of tiles by using `len`

src/odemis/acq/stitching/_tiledacq.py
nandishjpatel: I have applied the suggested change. Given the change the starting position (center of top-left first tile) will change for all systems.

tepals (approved):
Looks good to me. One minor improvement you could still do is add the raster test cases to the cloud hosted CI

src/odemis/util/raster.py
```diff
@@ -21,9 +21,9 @@
 
 """
 
-from builtins import range  # For Python 2 & 3
-
 import math
+from builtins import range  # For Python 2 & 3
```
pieleric, line 25: This can be dropped. This was useful for Python 2 support.

src/odemis/util/raster.py
```diff
@@ -142,3 +142,45 @@ def point_in_polygon(p, polygon):
...
     return inside
+
+
+def get_possible_intersections(row_pairs: List[Tuple], col_pairs: List[Tuple], include_neighbors: bool = True) -> Set[Tuple]:
+    """
+    Uses Bresenham's line algorithm to determine the possible intersected fields/tiles.
+
+    :param row_pairs: (list of tuple) A list of (start_row, end_row) pairs defining line segments.
+    :param col_pairs: (list of tuple) A list of (start_col, end_col) pairs defining line segments.
+    :param include_neighbors: (bool) If True, includes adjacent neighbors to ensure complete coverage. Defaults to True.
```
pieleric, line 153: In the docstring, not need to indicate the type, if it's already fully defined in the type annotation, which is the case here.
Similarly, no need to indicate the default value, if it's clear by reading the method signature, which is the case here.

src/odemis/util/raster.py
```diff
@@ -142,3 +142,45 @@ def point_in_polygon(p, polygon):
         p1x, p1y = p2x, p2y
 
     return inside
+
+
+def get_possible_intersections(row_pairs: List[Tuple], col_pairs: List[Tuple], include_neighbors: bool = True) -> Set[Tuple]:
+    """
+    Uses Bresenham's line algorithm to determine the possible intersected fields/tiles.
```
pieleric, line 149: This needs a much clearer explanation. Please think of someone looking at this file without context. When would you want to call this function? Even knowing the context, I don't understand what it does! What's an "intersection"? What's a field/tile? Looking at the caller, it seems related to polygons? If so, why not receiving as input a "polygon" as a series of x,y coordinates?
Looking at the algorithm in detail, it seems that it determines the "pixels" that are "drawn" by drawing the series of segments passed as input (in a very odd format). Is that right?


On the other hand, the fact it uses Bresenham's line algorithm is an implementation detail. It can be worth mentioning if it ensures it's fast, or that it uses less memory, but that's a side note after explaining what it does.

src/odemis/util/raster.py
```diff
@@ -142,3 +142,45 @@ def point_in_polygon(p, polygon):
...
+
+    :param row_pairs: (list of tuple) A list of (start_row, end_row) pairs defining line segments.
```
pieleric, line 151: tuples of what? Is that ints, or can they be float? What are the min/max values? It seems it can handle any large/small value, right?

src/odemis/util/raster.py
```diff
@@ -142,3 +142,45 @@ def point_in_polygon(p, polygon):
         p1x, p1y = p2x, p2y
 
     return inside
+
+
+def get_possible_intersections(row_pairs: List[Tuple], col_pairs: List[Tuple], include_neighbors: bool = True) -> Set[Tuple]:
```
pieleric, line 147: Looking at the code, `include_neighbors` essentially draws with a pixel size of 3 pixels instead of 1. Why making it the default? I think it's just because your code uses this parameter as True. However, from a generic point of view, I think it'd be more logical to have it `False` by default.

src/odemis/util/raster.py
```diff
@@ -142,3 +142,45 @@ def point_in_polygon(p, polygon):
...
+        dy = abs(col2 - col1)
+        sx = 1 if row1 < row2 else -1
+        sy = 1 if col1 < col2 else -1
+        err = dx - dy
+
+        while True:
+            intersections.add((row1, col1))
+
+            if include_neighbors:
+                # Add adjacent neighbors to ensure complete coverage
```
pieleric, line 169: I don't really get the "ensure complete coverage" idea. Is the goal to ensure that *if it's used to draw a polygon*, and you use a specific fill algorithm, it cannot escape the polygon? 
I think this comment should go in the caller, explaining why it uses `include_neighbors`.

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -275,22 +322,61 @@ def _getNumberOfTiles(self):
...
+        points = numpy.array(self._polygon.exterior.coords)
+        rows = numpy.floor((ymax - points[:, 1]) / reliable_fov[1]).astype(int)
+        cols = numpy.floor((points[:, 0] - xmin) / reliable_fov[0]).astype(int)
+
+        # Vectorized calculation of intersected grid cells
+        row_pairs = numpy.vstack((rows, numpy.roll(rows, -1))).T
+        col_pairs = numpy.vstack((cols, numpy.roll(cols, -1))).T
+
+        intersected_tiles = get_possible_intersections(row_pairs, col_pairs)
```
pieleric, line 353: Can add the `include_neighbors` argument explicitly, it'll be a little clearer on what it does, and what to expect as result.

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -275,22 +322,61 @@ def _getNumberOfTiles(self):
...
+
+        for row, col in intersected_tiles:
```
pieleric, line 355: Please add a comment explaining what this loop does. If  I understand correctly, as the rasterization is done with a "pen of 3 pixels", it goes easily out of the actual polygon, so you need to remove any tile which actually doesn't intersect the polygon, by looking at the exact tile area. Right?

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -275,22 +322,61 @@ def _getNumberOfTiles(self):
...
+            if 0 <= row < ny and 0 <= col < nx:
```
pieleric, line 356: I think it'd clearer to write this `if` as if it's a special (optimisation) case:
```suggestion
            if not (0 <= row < ny and 0 <= col < nx):  # Certainly out of the polygon => not worthy to acquire
                continue
```
Then, the rest is the "standard case", with an indentation less.

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -251,22 +300,20 @@ def _guessSmallestFov(self, ss):
         return (min(f[0] for f in fovs),
                 min(f[1] for f in fovs))
 
-    def _getNumberOfTiles(self):
+    def _getTileCoverage(self):
```
pieleric, line 303: Please, use the `snake_case` style. Also while at it, add type annotation (about the return values).

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -251,22 +300,20 @@ def _guessSmallestFov(self, ss):
...
         """
-        Calculate needed number of tiles (horizontal and vertical) to cover the whole area,
-        and the adjusted position of the first tile.
+        Calculate the exact tiles required to cover the region.
         return:
-            nb (int, int): number of tile in X, Y
             starting_position (float, float): center of the first tile (at the top-left)
```
pieleric, line 307: The code returns a dict, not a tuple.

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -275,22 +322,61 @@ def _getNumberOfTiles(self):
...
+        tile_grid = numpy.zeros((ny, nx), dtype=bool)
+
+        # Vectorized conversion of polygon points to grid coordinates
+        points = numpy.array(self._polygon.exterior.coords)
+        rows = numpy.floor((ymax - points[:, 1]) / reliable_fov[1]).astype(int)
+        cols = numpy.floor((points[:, 0] - xmin) / reliable_fov[0]).astype(int)
+
+        # Vectorized calculation of intersected grid cells
+        row_pairs = numpy.vstack((rows, numpy.roll(rows, -1))).T
+        col_pairs = numpy.vstack((cols, numpy.roll(cols, -1))).T
```
pieleric, line 351: It seems that you have to convert a polygon format to a super odd format used by `get_possible_intersections()`.
I think it'd read much clearer if you passed to `get_possible_intersections()` a series of segments in the form `[((x1, y1), (x2,y2)), ((x2, y2), (x3, y3))...]` Or even just a series of points as `[(x1, y1), (x2, y2)...]`. Then that function would take care of adjusting the format for its own consumption.

src/odemis/acq/stitching/_tiledacq.py
```diff
@@ -306,31 +392,25 @@ def _cancelAcquisition(self, future):
...
-        # similar to writing/scanning convention, but move of just one unit
-        # every time.
-        # A-->-->-->--v
-        #             |
-        # v--<--<--<---
-        # |
-        # --->-->-->--Z
-        rep (int, int): X, Y number of tiles
-        return (generator of tuple(int, int)): x/y positions, starting from 0,0
+        Sort the given tile indices in a zigzag scanning order.
```
pieleric, line 397: I think it's worth leaving the ascii art, to explain what "zigzag" mean. You can say it's an example for the case of a 5x3 rectangle.

pieleric (changes requested):

src/odemis/acq/stitching/_tiledacq.py
nandishjpatel: Yes, you are right. The intersected tiles contains tiles which be inside the polygon, outside the polygon and tiles which actually intersect with the polygon. We only need the tiles which actually intersect the polygon, to binary fill the holes and get the tile indices.

pieleric (approved):

pieleric:
@nandishjpatel it looks fine. Could you squash your last commit into the first 3 commits? Then, we can merge this PR.

PR #3015, opened 2025-01-24
[NSRPARMA-120][feat] extend odemis-select-mic-start to ask for laser confirmation

pieleric:
For systems with multiple lasers, it's helpful to ask the user to
confirm before starting a configuration with a specific laser. Selecting
the wrong configuration could damage the detectors.

=> Introduce a new argument to the script, so that on such system
a confirmation dialog box is shown in such case.

### DIFF
--- a/install/linux/usr/bin/odemis-select-mic-start
+++ b/install/linux/usr/bin/odemis-select-mic-start
@@ -12,6 +12,7 @@ After selection of the microscope configuration the right (yaml) file will be us
 '
 # Default values for options
 select_version=false
+confirm_laser=false
 target_files=()
 
 # Function to display help information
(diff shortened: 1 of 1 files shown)


tmoerkerken:
So it only protects when the script is called with the newly added argument. Will that be safe enough in practice?

install/linux/usr/bin/odemis-select-mic-start
```diff
@@ -12,6 +12,7 @@ After selection of the microscope configuration the right (yaml) file will be us
 '
 # Default values for options
 select_version=false
+confirm_laser=false
```
tmoerkerken, line 15: So it only protects when the script is called with the newly added argument. Will that be safe enough in practice?

pieleric: It doesn't really "protect" the hardware per-se. It just shows an extra pop-up. To use this script as the Odemis icon, one need to override the default icon by writing a special launcher (`.desktop`). So once you're at this level of attention, it is fine to assume you can decide & test whether the extra pop-up is shown or not.

K4rishma (approved):

tmoerkerken (approved):

PR #3020, opened 2025-01-27
[METEOR-1238] [feature] Multi-point Correlation GUI Interface

K4rishma:
**Testing version (Not to be merged)**
- Button under the Localization tab
- TODOs added

Final version
- Button under the FIBSEM tab
- TODOs removed and implemented


src/odemis/gui/cont/multi_point_correlation.py
```diff
@@ -0,0 +1,815 @@
...
+    def _do_correlation(self):
+        """Run the correlation between the FIB and FM images."""
+        # Modify the input data to match the required format to run 3DCT
+        fm_das = [stream.raw[0] for stream in self.correlation_target.fm_streams]
+        fm_image = _convert_das_to_numpy_stack(fm_das)
+        fib_da = self.correlation_target.fib_stream.raw[0]
+        fib_coords = []
+        fm_coords = []
+        poi_coords = []
+        path = os.getcwd()
```
patrickcleeve2, line 439: Should be project_path

src/odemis/gui/cont/multi_point_correlation.py
```diff
@@ -0,0 +1,815 @@
...
+        for fib_coord in self.correlation_target.fib_fiducials:
+            fib_coord = self.correlation_target.fm_streams[0].getPixelCoordinates_alt(fib_coord.coordinates.value[0:2])
+            fib_coords.append(fib_coord)
+        fib_coords = numpy.array(fib_coords, dtype=numpy.float32)
+        for fm_coord in self.correlation_target.fm_fiducials:
+            fm_coord_2d = self.correlation_target.fm_streams[0].getPixelCoordinates_alt(fm_coord.coordinates.value[0:2])
+            fm_coords.append([fm_coord_2d[0], fm_coord_2d[1], fm_coord.coordinates.value[2]])
+        fm_coords = numpy.array(fm_coords, dtype=numpy.float32)
+        min_n = min(len(fib_coords), len(fm_coords)) # make fm coords and fib coords the same length
```
patrickcleeve2, line 448: Rather than clipping, don't do correlation if they're not the same length. It's more likely to be an error (user forgot a point, added one by mistake, etc) than be useful

src/odemis/gui/cont/multi_point_correlation.py
```diff
@@ -0,0 +1,815 @@
...
+        self.correlation_target.fib_projected_fiducials = []
+        points = self.correlation_target.correlation_result['output']['error']['reprojected_3d']
+        for n, i in enumerate(points[0]):
+            p_pos = self.correlation_target.fib_stream.getPhysicalCoordinates((points[0][n], points[1][n]))
+            target = Target(x=p_pos[0], y=p_pos[1], z=0, name="PP" + str(n + 1), type="ProjectedPoints", index=n + 1,
+                            fm_focus_position=0)
+            self._tab_data_model.projected_points.append(target)
+            self.correlation_target.fib_projected_fiducials.append(target)
+
+        projected_poi = get_reprojected_poi_coordinate(self.correlation_target.correlation_result)
```
patrickcleeve2, line 471: I think this might be one of the correlation errors, this projected_poi in in metres from center of image, don't think it can be directly used like this. You will probably need one of the other poi results image_px, or px

patrickcleeve2 (changes requested):

src/odemis/gui/cont/acquisition/overview_stream_acq.py
```diff
@@ -101,3 +101,65 @@ def open_acquisition_dialog(self):
...
+        self._tab = tab
+
+    def open_correlation_dialog(self):
+        """
+        return None or a list of DataArrays: the acquired images. None if it was
+          cancelled.
+        """
+        # Indicate we are acquiring, especially important for the SEM which
+        # need to get the external signal to not scan (cf MicroscopeController)
+        self._main_data_model.is_acquiring.value = True
```
pieleric, line 125: Is it really important here? The dialog is modal, and all the streams are paused. Do you also need this?

src/odemis/gui/cont/multi_point_correlation.py
```diff
@@ -0,0 +1,815 @@
...
+            if self.correlation_target.fib_fiducials and self.correlation_target.fib_stream:
+                targets.append(self.correlation_target.fib_fiducials)
+            #  As rhe correlation result is calculated quickly (no need to load it)
+            # if self.correlation_target.fib_projected_fiducials:
+            #     projected_points.append(self.correlation_target.fib_projected_fiducials)
+            # if self.correlation_target.fib_projected_pois:
+            #     projected_points.append(self.correlation_target.fib_projected_pois)
+            # projected_points = list(
+            #     itertools.chain.from_iterable([x] if not isinstance(x, list) else x for x in projected_points))
+            # self._tab_data_model.projected_points = projected_points
```
pieleric, line 163: Why is that all commented out? Isn't it useful?

src/odemis/gui/cont/multi_point_correlation.py
```diff
@@ -0,0 +1,815 @@
...
+            self.previous_group = group_key
+            if group_key:
+                streams_list = self._tab_data_model.main.currentFeature.value.streams.value
+                for key, indices in self.stream_groups.items():
+                    for index in indices:
+                        stream = streams_list[index]
+                        # find stream controller for the stream
+                        ssc = next(
+                            (sc for sc in  self._panel.streambar_controller.stream_controllers
+                             if sc.stream == stream), None)
```
pieleric, line 220: ssc will be `None` if no controller can be found. Probably very unlikely, but you should still explicitly handle such case, maybe just with a error log and ending the function.

src/odemis/gui/cont/multi_point_correlation.py
```diff
@@ -0,0 +1,815 @@
...
+            if (len(self.correlation_target.fib_fiducials) >= 4 and len(
+                    self.correlation_target.fm_fiducials) >= 4 and self.correlation_target.fm_pois and len(
+                self._tab_data_model.views.value[0].stream_tree) > 0) and self.correlation_target.fib_stream:
+                return True
+            else:
+                self.correlation_target.reset_attributes()
+                self._tab_data_model.projected_points = []
+                self.correlation_txt.SetLabel("Correlation RMS Deviation :")
+                for vp in self._viewports:
+                    if vp.view.name.value == "SEM Overview":
```
pieleric, line 392: Relying on the view name is quite easily broken. It'd better to explicitly use the viewport. Both viewports are assigned as attributes, so you should be able to directly do it. I guess it's `self._panel.vp_correlation_tr`?

src/odemis/gui/cont/multi_point_correlation.py
```diff
@@ -0,0 +1,815 @@
...
+
+        # pass through event, if not a valid correlation key or enabled
+        valid_keys = [wx.WXK_LEFT, wx.WXK_RIGHT, wx.WXK_UP, wx.WXK_DOWN]
+        if key not in valid_keys:
+            evt.Skip()
+            return
+
+        if ctrl_mode:
+            if key == wx.WXK_UP:
+                if self._tab_data_model.focussedView.value.name.value == "FLM Overview" or self._tab_data_model.focussedView.value.name.value == "SEM Overview":
```
pieleric, line 499: Don't rely on view name to identify the views. Either use the `stream_classes` of the view, or compare to the `viewport.view`.

Anyway, for this specific code, as there are only 2 views, probably  you don't even need a `if`?

src/odemis/gui/cont/multi_point_correlation.py
```diff
@@ -0,0 +1,815 @@
...
+        """
+        with self.lock:  # Ensure thread-safe access to latest_change
+            self.latest_change = True  # Overwrite with the latest change
+            if self.is_processing:  # If a process is running, it will handle the latest change.
+                logging.warning("Multipoint Correlation is running. It will handle the latest change.")
+                self.correlation_txt.SetLabel("Correlation RMS Deviation :  Calculating...")
+                return
+            # Start processing the latest change in a separate thread
+            self.process_thread = threading.Thread(target=self._process_latest_change)
+            self.process_thread.start()
```
pieleric, line 411: This is quite complex threading code. Typically, a simple way is to do the following:
* start a thread at init (which runs the whole time in the background, as "daemon")
* this thread waits for data coming from queue.Queue
* whenever new processing is required, it is placed on the queue.
* to make sure it only processes the latest request, after getting a request, it can check for more data on the queue, non-blocking (or just repeat until the queue is empty)
* to request the thread to stop, pass a special data, for instance `None`.

=> You don't need locks. You only start the thread once. You are certain there is only one computation running at a given time.

src/odemis/gui/win/acquisition.py
```diff
@@ -1311,6 +1317,164 @@ def on_acquisition_done(self, future):
...
+
+    def on_close(self, evt):
+        """Close event handler that executes various cleanup actions"""
+        assert self._correlation_points_controller.is_processing == False
+
+        self.EndModal(wx.ID_CANCEL)
+
+    def on_cancel(self, evt):
+        """Handle acquisition cancel button click"""
+        logging.info("Close button clicked, exiting 3DCT correlation")
```
pieleric, line 1471: I don't think there is any place that should mention "3DCT" to the user (because it's really unclear what this means). I'd just say "correlation window".

pieleric (changes requested):

src/odemis/acq/target.py
```diff
@@ -0,0 +1,39 @@
...
+You should have received a copy of the GNU General Public License along with
+Odemis. If not, see http://www.gnu.org/licenses/.
+"""
+from typing import List
+
+from odemis import model
+
+
+class Target:
+    def __init__(self, x,y,z, name:str, type:str, index: int,  fm_focus_position: float, size: float = None ):
```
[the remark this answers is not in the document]
K4rishma: I have added the docstring. Fm_focus_position is the z of the objective whereas z is the physical position of the target. Both can be used.

src/odemis/gui/cont/tabs/localization_tab.py
```diff
@@ -311,7 +311,7 @@ def load_overview_data(self, data: List[model.DataArray]):
         # sync overview streams with correlation tab
         if len(streams) > 0 and self.main_data.role == "meteor":
             correlation_tab = self.main_data.getTabByName("meteor-correlation")
-            correlation_tab.correlation_controller.add_streams(streams)
+            correlation_tab.correlation_controller.group_add_streams(streams)
```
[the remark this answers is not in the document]
K4rishma: Reverted it, good catch!

src/odemis/acq/feature.py
```diff
@@ -37,18 +38,63 @@
...
+
+    def __init__(self):
+        # Input parameters of 3DCT
+        self.fm_pois: List[Target] = []
+        self.fm_fiducials: List[Target] = []
+        self.fib_fiducials: List[Target] = []
+        self.fib_surface_fiducial : Target = None
+        self.fib_stream = None #:StaticSEMStream = None or StaticFIBStream = None
+        self.fm_streams: List[StaticFluoStream] = []
+        self.superz: StaticFluoStream = None
```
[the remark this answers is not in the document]
K4rishma: I think it is used to save the stream i.e. selected by the user for super z. I have renamed the parameter as per the suggestion.

src/odemis/gui/cont/multi_point_correlation.py
```diff
@@ -0,0 +1,815 @@
...
+                if isinstance(stream, StaticStream):
+                    # Add FIB stream
+                    ssc =  self._panel.streambar_controller.addStream(stream, play=False, add_to_view=True)
+                    ssc.stream_panel.show_remove_btn(True)
+                    if isinstance(stream, StaticFIBStream) or isinstance(stream, StaticSEMStream):
+                        # Check if the FIB stream is already loaded. If it is, replace the previous stream with the new
+                        if not self.correlation_target.fib_stream:
+                            self.correlation_target.fib_stream = stream
+                            # Load the FIB fiducials if they are not already loaded
+                            if all("FIB" not in target.name.value for target in
```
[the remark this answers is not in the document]
K4rishma: I understand the names are brittle and not good idea to use in condition. The above block is temporary and will be removed once the functionality is switched to FIBSEM tab. Though for other instances where I have used name as an identifier, I am not sure what other way I can distinguish between FM and FIB fiducial. Both their types are "Fiducial" and z coordinate of both can be zero. Any idea to make it better?

`    def _update_feature_correlation_target(self, surface_fiducial=False):
        """Populate the correlation target based on the latest changes to save it."""
        if not self.correlation_target:
            return

        if surface_fiducial:
            fib_surface_fiducial = self._tab_data_model.fib_surface_point.value
            self.correlation_target.fib_surface_fiducial = fib_surface_fiducial
        else:
            # Corner case: When fiducials are deleted and the indices are not continiuos. Then the fiducial pairs
            # will be incorrectly matched together.
            # TODO Handle the corner case
            fib_fiducials = []
            fm_fiducials = []
            self.correlation_target.fm_pois = []
            for target in self._tab_data_model.main.targets.value:
                if "FIB" in target.name.value:
                    fib_fiducials.append(target)
                elif "FM" in target.name.value:
                    fm_fiducials.append(target)
                elif "POI" in target.name.value:
                    self.correlation_target.fm_pois = target
            if fib_fiducials:
                fib_fiducials.sort(key=lambda x: x.index.value)
                self.correlation_target.fib_fiducials = fib_fiducials
            if fm_fiducials:
                fm_fiducials.sort(key=lambda x: x.index.value)
                self.correlation_target.fm_fiducials = fm_fiducials`

src/odemis/acq/target.py
```diff
@@ -0,0 +1,39 @@
...
+        self.coordinates = model.ListVA((x, y, z), unit="m")
```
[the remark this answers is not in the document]
K4rishma: I tried to implement a Dict VA as I thought it will be clearer but observed the following things
1. The subscription to DictVA coordinates did not work straight out of the box as for List VA
2. In the overlay classes, the canvas functions that change between the view and physical coordinates work with list. Changing it to Dict VA seems out of sync unless canvas functions also use a dict, in my opinion.

At the moment, I have changed it back to using the ListVA.

src/odemis/gui/model/tab_gui_data.py
```diff
@@ -405,7 +410,88 @@ def __init__(self, main):
...
+        elif self.focussedView.value.name.value == "SEM Overview":
+            if type == "SurfaceFiducial":
+                target = Target(x, y, z=0, name="FIB_surface", type=type, index=1,
+                                fm_focus_position=fm_focus_position)
+                self.fib_surface_point.value = target
+                return target
+            t_name = make_unique_name("FIB-1", existing_names)
+            index = int(t_name[-1])
+            target = Target(x, y, z=0, name=t_name, type=type, index=index,
+                            fm_focus_position=fm_focus_position)
```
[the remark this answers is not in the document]
K4rishma: I have a feeling to make a function if the code is repeated three times. I used the repetition twice, so thought to just use it directly without making a function

src/odemis/gui/comp/overlay/cryo_feature.py
```diff
@@ -267,3 +272,404 @@ def set_icon(feature_icon):
...
+        pass
+
+    def draw(self, ctx, shift=(0, 0), scale=1.0):
+        """
+        Draw all the targets, on their location, indicating their status.
+        """
+        pass
+
+
+class CryoCorrelationFmPointsOverlay(CryoCorrelationPointsOverlay):
```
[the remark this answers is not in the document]
K4rishma: Good point, this could be indeed clearer. I would have to change a quite a bit for this. I will maybe wait till the application is tested.

PR #3023, opened 2025-02-05
[FVEM-27] [feat] Add visibility toggle for calibration regions

tepals:
This feature adds toggle buttons for calibration regions, this allows the user to show and hide calibration regions

FVEM-27

### DIFF
--- a/src/odemis/gui/cont/acquisition/fastem_acq.py
+++ b/src/odemis/gui/cont/acquisition/fastem_acq.py
@@ -54,7 +54,9 @@ from odemis.gui import (
     FG_COLOUR_EDIT,
     FG_COLOUR_ERROR,
     FG_COLOUR_WARNING,
+    img,
 )
+from odemis.gui.comp import buttons
 from odemis.gui.comp.fastem_roa import FastEMROA
 from odemis.gui.comp.fastem_user_settings_panel import (
     DWELL_TIME_ACQUISITION,
@@ -1186,7 +1188,8 @@ class FastEMCalibrationController:
         self.calibration_panel = SettingsPanel(
             self._tab_panel.pnl_calib, size=(400, 80)
         )
-        self._calib_1_lbl, self._calib_1 = self.calibration_panel.add_checkbox_control(
+
+        self._calib_1_lbl, self._calib_1_vis_btn, self._calib_1 = self.add_calibration_control(
             CALIBRATION_1, value=False, pos_col=2, span=(1, 1)
         )
         self._calib_1_lbl.SetToolTip(
@@ -1198,7 +1201,8 @@ class FastEMCalibrationController:
         )
         self._calib_1.SetName(CALIBRATION_1)
         self._calib_1_lbl.SetForegroundColour(FG_COLOUR_BLIND_BLUE)
-        self._calib_2_lbl, self._calib_2 = self.calibration_panel.add_checkbox_control(
+
+        self._calib_2_lbl, self._calib_2_vis_btn, self._calib_2 = self.add_calibration_control(
             CALIBRATION_2, value=False, pos_col=2, span=(1, 1)
         )
         self._calib_2_lbl.SetToolTip(
@@ -1210,7 +1214,8 @@ class FastEMCalibrationController:
         )
         self._calib_2.SetName(CALIBRATION_2)
         self._calib_2_lbl.SetForegroundColour(FG_COLOUR_BLIND_ORANGE)
-        self._calib_3_lbl, self._calib_3 = self.calibration_panel.add_checkbox_control(
+
+        self._calib_3_lbl, self._calib_3_vis_btn, self._calib_3 = self.add_calibration_control(
             CALIBRATION_3, value=False, pos_col=2, span=(1, 1)
         )
         self._calib_3_lbl.SetToolTip(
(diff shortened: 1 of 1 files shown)


tepals:
This PR adds visibility toggles for the calibration regions. The Jira task contains a screen-cast demonstrating the feature.
https://delmic.atlassian.net/browse/FVEM-27

image

src/odemis/gui/cont/acquisition/fastem_acq.py
```diff
@@ -1197,6 +1205,91 @@ def __init__(
         # enable/disable calibration buttons if calibrating
         self._tab_data_model.is_calibrating.subscribe(self._on_is_calibrating)
 
+    def _on_visibility_btn(self, evt):
+        """Toggle the visibility of the calibration region based on the button state."""
+        current_sample = self._main_data_model.current_sample.value
+        focussed_view = self._main_tab_data.focussedView.value
+        # if current_sample and focussed_view:
```
nandishjpatel, line 1212: I think this if statement is needed and the below lines of code should be indented

tepals: ah yes, that was left over from debugging. Should I update it to:
```suggestion
        if not current_sample and focussed_view:
            return
```

nandishjpatel:
@tepals open questions related to this feature

If visibility toggle is OFF should the calibration checkbox be un-ticked and disabled?
If all three calibration visibility toggle is OFF should the calibration RUN button be disabled?
The calibrations checkbox have a hierarchy of checking and enabling them and together with the visilbity toggle its a bit more complex.

Reason for asking such questions is that I think there a connection between the toggle button and checkbox. Even if the visibility toggle is OFF the calibration box last coordinates are still in memory and the calibration can still be run with no box drawing in the canvas.

tepals:
> @tepals open questions related to this feature
> 
> If visibility toggle is OFF should the calibration checkbox be un-ticked and disabled? If all three calibration visibility toggle is OFF should the calibration RUN button be disabled? The calibrations checkbox have a hierarchy of checking and enabling them and together with the visilbity toggle its a bit more complex.
> 
> Reason for asking such questions is that I think there a connection between the toggle button and checkbox. Even if the visibility toggle is OFF the calibration box last coordinates are still in memory and the calibration can still be run with no box drawing in the canvas.

My intention was that the calibrations could still be run even if the boxes are hidden, so that's how it's implemented. But I will double check this with Marre next week.

nandishjpatel:
> My intention was that the calibrations could still be run even if the boxes are hidden, so that's how it's implemented. But I will double check this with Marre next week.

Well, if the user is allowed to run calibrations which are hidden, then at the start of calibration the boxes can be shown. This can be a simpler approach.

tepals:
> > My intention was that the calibrations could still be run even if the boxes are hidden, so that's how it's implemented. But I will double check this with Marre next week.
> 
> Well, if the user is allowed to run calibrations which are hidden, then at the start of calibration the boxes can be shown. This can be a simpler approach.

That makes sense!

src/odemis/gui/cont/acquisition/fastem_acq.py
```diff
@@ -1342,14 +1437,32 @@ def on_run(self, evt):
...
+                    self._calib_2_vis_btn.SetValue(True)
             if self._calib_3.IsChecked() and (
                 CALIBRATION_2 in calib_names or calib_2_done
             ):
                 calib_names.append(CALIBRATION_3)
+                if not self._calib_3_vis_btn.GetValue():
+                    # show the calibration region again when the user starts the calibration
+                    calibration = current_sample.scintillators[scintillator_num].calibrations[CALIBRATION_3]
+                    calibration.shape.active.value = True
+                    calibration.shape.cnvs.add_world_overlay(calibration.shape)
```
K4rishma, line 1464: If it is possible to make the visible button as an attribute of the calibration, then this function can be generalized by making a new function. The new function takes calibration as input and set the attributes based on the visibility.

src/odemis/gui/cont/acquisition/fastem_acq.py
```diff
@@ -1197,6 +1205,93 @@ def __init__(
...
+
+        lbl_ctrl, visibility_btn = self._add_side_label_with_toggle(label_text)
+        value_ctrl = wx.CheckBox(self.calibration_panel, wx.ID_ANY, style=wx.ALIGN_RIGHT | wx.NO_BORDER)
+        self.calibration_panel.gb_sizer.Add(value_ctrl, (self.calibration_panel.num_rows, pos_col), span=span,
+                                            flag=wx.EXPAND | wx.TOP | wx.BOTTOM, border=5)
+        value_ctrl.SetValue(value)
+
+        return lbl_ctrl, visibility_btn, value_ctrl
+
+    def _add_side_label_with_toggle(self, label_text):
```
K4rishma, line 1263: The function name seems like a general utility function, although the return value is specific. Maybe rename the function or the function contents  to make it either general or specific for clarity purpose.

src/odemis/gui/cont/acquisition/fastem_acq.py
tepals: The calibration is an instance of the `FastEMCalibration` class, which is part of the acquisition code, so it doesn't make sense to have the button as an attribute there. I will create a method on the current class, so I don't have to repeat the same code 3 times.

src/odemis/gui/cont/acquisition/fastem_acq.py
tepals: good point, I'll rename it to `_add_calibration_label_with_toggle`

tepals:
@K4rishma I fixed your comments, can you do another review? I can't click the re-review button because you already did that :grinning:

K4rishma (approved):

nandishjpatel (approved):

PR #3027, opened 2025-02-06
[TRCL-570][feat] introduce startScan Event to synchronize CCD with e-beam during SPARC acq

pieleric:
From the moment that the acquisition ask the e-beam scanner to start and the e-beam is actually at the right position, it takes *some* time.  It's in the order of a few ms. So far, the only e-beam scanner was  semcomedi, and it was (surprisingly) constant in terms of time. It would  take between 2 and 5 ms. So it was good enough to just always wait 5ms.

We now also have semnidaq, and the NI-DAQ library introduces a lot longer latency. While it can be as "fast" as 10ms, it sometimes takes 40ms to start a scan. There are probably ways to make it a little faster, but it's  unlikely it will be possible to always keep all scanners < 5ms.
    
=> Use a new Event, startScan, for the e-beam scanner to report the e-beam is ready. Then, we can use it to directly start a CCD acquisition.

### DIFF
--- a/src/odemis/acq/stream/_sync.py
+++ b/src/odemis/acq/stream/_sync.py
@@ -1086,7 +1086,7 @@ class SEMCCDMDStream(MultipleDetectorStream):
         self._sccd = s1
         self._ccd = s1._detector
         self._ccd_df = s1._dataflow
-        self._trigger = self._ccd.softwareTrigger
+        self._trigger = self._emitter.startScan  # to acquire a CCD image every time the SEM starts a new scan
         self._ccd_idx = len(self._streams) - 1  # optical detector is always last in streams
 
     def _supports_hw_sync(self):
@@ -1685,11 +1685,6 @@ class SEMCCDMDStream(MultipleDetectorStream):
             # retrigger, or unsynchronise/resynchronise just before the end of
             # last scan).
 
-            # prepare detector
-            self._ccd_df.synchronizedOn(self._trigger)
-            # subscribe to last entry in _subscribers (optical detector)
-            self._ccd_df.subscribe(self._subscribers[self._ccd_idx])
-
             # Instead of subscribing/unsubscribing to the SEM for each pixel,
             # we've tried to keep subscribed, but request to be unsynchronised/
             # synchronised. However, synchronizing doesn't cancel the current
@@ -1730,6 +1725,12 @@ class SEMCCDMDStream(MultipleDetectorStream):
             logging.debug("Scanning resolution is %s and scale %s",
                           self._emitter.resolution.value,
                           self._emitter.scale.value)
+
+            # Prepare CCD: acquire one frame every time the SEM starts scanning.
+            # The SEM may scan multiple times for each CCD frame.
+            self._ccd_df.synchronizedOn(self._trigger)
+            # Get the CCD ready to acquire
+            self._ccd_df.subscribe(self._subscribers[self._ccd_idx])
 
             last_ccd_update = 0
             start_t = time.time()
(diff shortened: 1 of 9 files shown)


src/odemis/driver/test/semnidaq_test.py
```diff
@@ -860,6 +888,10 @@ def test_flow(self):
         duration = stop_t - start_t
         self.assertGreaterEqual(duration, exp_tot_dur, f"Acquisition ended too early: {duration}s")
 
+        self.assertEqual(evt_counter.count, 1)
+        self.scanner.startScan.unsubscribe(evt_counter)
+
+
```
K4rishma, line 894: to remove the extra space between the functions

src/odemis/driver/simsem.py
```diff
@@ -540,6 +544,9 @@ def _acquire_thread(self, callback):
                 # as in Odemis the convention for SEM is that the ebeam waits
                 # for _all_ the detectors to be ready before scanning.
                 self.data._waitSync()
+                if first_frame:
+                    self.parent._scanner.startScan.notify()
+                    first_frame = False
```
K4rishma, line 549: Does this automatically stop notifying something which is similar to the unsubscribe used in the testcases when self._acquisition_must_stop.clear() is called in finally: section?

src/odemis/acq/stream/_sync.py
```diff
@@ -1887,26 +1888,10 @@ def _acquireImage(self, n, px_idx, img_time, sem_time,
                 raise CancelledError()
 
             # subscribe to _subscribers
+            # As soon as the e-beam start scanning (which can take a couple of ms), the
```
tmoerkerken, line 1891:
```suggestion
            # As soon as the e-beam starts scanning (which can take a couple of ms), the
```

src/odemis/acq/stream/_sync.py
```diff
@@ -1887,26 +1888,10 @@ def _acquireImage(self, n, px_idx, img_time, sem_time,
...
+            # startScan event is sent, which triggers the acquisition of one CCD frame.
             for s, sub in zip(self._streams[:-1], self._subscribers[:-1]):
```
tmoerkerken, line 1893: Can you comment on why the last element is excluded?

src/odemis/acq/stream/_sync.py
```diff
@@ -2119,12 +2112,11 @@ def _runAcquisitionScanStage(self, future):
                     if self._acq_state == CANCELLED:
                         raise CancelledError()
 
+                    # Start e-beam scan. As soon as it really start, a startScan event is sent, which
```
tmoerkerken, line 2115:
```suggestion
                    # Start e-beam scan. As soon as it really starts, a startScan event is sent, which
```

src/odemis/acq/stream/_sync.py
```diff
@@ -2008,9 +2001,9 @@ def _runAcquisitionScanStage(self, future):
         # The idea of the acquiring with a scan stage:
```
tmoerkerken, line 2001:
```suggestion
        # The idea of acquisition with a scan stage:
```

src/odemis/driver/semnidaq.py
```diff
@@ -2251,6 +2253,23 @@ def _set_state(self, active):
                 time.sleep(self._activation_delay)
 
 
+class EventOnce(model.Event):
```
tmoerkerken, line 2256: This is duplication of the function at the semcomedi driver. Maybe put it in a shared module?

tmoerkerken (changes requested):

src/odemis/acq/stream/_sync.py
```diff
@@ -2210,6 +2202,11 @@ def _runAcquisitionScanStage(self, future):
                                 sstage.moveAbsSync(orig_spos)
                                 prev_spos.update(orig_spos)
                             try:
+                                # Temporarily switch the CCD to a different event trigger, so that it
+                                # doesn't get triggered while the leech is running (because it could use the
+                                # e-beam, which would send a startScan event)
+                                self._ccd_df.synchronizedOn(self._ccd.softwareTrigger)
```
K4rishma, line 2208: Maybe an assert line can be used to check that the notification boolean value of startscan is False before reusing the real trigger? Same for _acquireImage

K4rishma (approved):

src/odemis/driver/semnidaq.py
pieleric: Indeed. moved to `driver.EventOnce`

src/odemis/driver/simsem.py
pieleric: No, this is different. It's a model.Event, an Odemis object that works in between Components. `_acquisition_must_stop` is a threading.Event, a standard Python object that only works within the component (because each component runs in a separate Python process).

tmoerkerken (approved):

PR #3030, opened 2025-02-11
[feat] SettingsObserver should also store axes "user-friendly" info if available

pieleric:
For some axes, the "position" is an abritrary number. For instance, on
the spectrograph, the band is 1->4. This is not helpful for the user.
However, the .axes contains the information to map that value (eg, 1) to
a user-friendly description (eg, "1200 l/mm (blaze: 500.0)").

Several SPARC users requested for this info to be stored too.

=> When available, also store the axis definition.

### DIFF
--- a/src/odemis/acq/acqmng.py
+++ b/src/odemis/acq/acqmng.py
@@ -29,7 +29,7 @@ from concurrent.futures import CancelledError
 import logging
 import os
 import threading
-from typing import Set
+from typing import Set, Dict
 
 from odemis import model
 from odemis.acq import _futures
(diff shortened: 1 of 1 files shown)


src/odemis/acq/acqmng.py
```diff
@@ -726,11 +726,31 @@ def __init__(self, microscope: model.HwComponent, components: Set[model.HwCompon
...
+                            for axis_name, p in value.items():
+                                if (axis_name in axes_def and hasattr(axes_def[axis_name], "choices")
+                                    and isinstance(axes_def[axis_name].choices, dict)
+                                    and p in axes_def[axis_name].choices
+                                   ):
+                                    value[axis_name] = f"{p} ({axes_def[axis_name].choices[p]})"
+                        except Exception:
+                            logging.exception("Failed to interpret user-friendly axis info for %s.position", comp_name)
+
+                        self._all_settings[comp_name][va_name][0] = value
```
K4rishma, line 746: I do not see anything under if contidition, is line 746 indexed properly?

pieleric: Yes. Depending on the if/else the `update_settings()` function is different.

tmoerkerken (approved):

PR #3036, opened 2025-02-20
[TRCL-543] [fix] odemisd: handle metadata for monochromator with both spectrograph and filter

pieleric:
With a monochromator, both the filter (in front of the spectrograph),
and the spectrograph center wavelength + output slit opening have effect
on which wavelength is observed. Until now, move on any of the axes
would update MD_OUT_WL... but only the latest move would define the
value.
That's especially annoying if the lastest move is to move the filter to
"pass-through" (very common). In this case, the MD_OUT_WL would be empty
(incorrectly).

The right way to compute MD_OUT_WL is to take into account all the axes
together. Essentially, taking the intersection between the filter
wavelength range and the output slit opening range.

### DIFF
--- a/src/odemis/odemisd/mdupdater.py
+++ b/src/odemis/odemisd/mdupdater.py
@@ -21,7 +21,8 @@ Odemis. If not, see http://www.gnu.org/licenses/.
 '''
 import collections
 import logging
-import numpy
+from typing import Optional, Tuple, Dict, Union, Iterable
+
 from odemis import model
 
 
@@ -54,6 +55,10 @@ class MetadataUpdater(model.Component):
         # All the components already observed
         # str -> set of str: name of affecting component -> names of affected
         self._observed = collections.defaultdict(set)
+
+        # To handle monochromator (aka detectors), which are affected both by the spectrograph and the filter
+        self._det_to_spectrograph : Dict[str, model.HwComponent] = {}  # Detector name -> Spectrograph that affects that detector
+        self._det_to_filter : Dict[str, model.HwComponent] = {}  # Detector name -> Filter that affects that detector
 
         microscope.alive.subscribe(self._onAlive, init=True)
 
@@ -102,9 +107,11 @@ class MetadataUpdater(model.Component):
                     # update the emitted light wavelength
                     observed = self.observeLight(a, d)
                 elif a.role and a.role.startswith("spectrograph"):  # spectrograph-XXX too
+                    self._det_to_spectrograph[dn] = a
                     # update the output wavelength range
                     observed = self.observeSpectrograph(a, d)
                 elif a.role in ("cl-filter", "filter"):
+                    self._det_to_filter[dn] = a
                     # update the output wavelength range
                     observed = self.observeFilter(a, d)
                 elif a.role == "quarter-wave-plate":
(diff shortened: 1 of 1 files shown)


src/odemis/odemisd/mdupdater.py
```diff
@@ -298,25 +305,106 @@ def updateLightPower(power, light=light, comp_affected=comp_affected):
...
+
+        if spec_bandwidth is None:  # Standard case, for everything except monochromators
+            bandwidth = filter_bandwidth  # None, str or tuple
+        elif filter_bandwidth is None:
+            bandwidth = spec_bandwidth
+        elif isinstance(filter_bandwidth, str):
+            # We don't support computing the intersection of a string & a tuple -> just trust the spectrograph
+            logging.info("Filter %s is not a range, will use spectrograph bandwidth for MD_OUT_WL", filter_bandwidth)
+            bandwidth = spec_bandwidth
+        else:  # Both are a wavelength range => take the intersection
```
tmoerkerken, line 385: These cases make for a good unit test. Not sure if that's easy to implement for this module.

tmoerkerken (approved):

src/odemis/odemisd/mdupdater.py
```diff
@@ -298,25 +305,106 @@ def updateLightPower(power, light=light, comp_affected=comp_affected):
...
+                 or a string if it's defined just as a name,
+                 or None if there is no filter.
+        """
+        if filter is None:
+            return None
+
+        pos = filter.position.value
+        bandwidth = filter.axes["band"].choices[pos["band"]]
+        if bandwidth == model.BAND_PASS_THROUGH:
+            return None
```
tepals, line 345: can you add a comment to explain why you return None in this case?

src/odemis/odemisd/mdupdater.py
```diff
@@ -298,25 +305,106 @@ def updateLightPower(power, light=light, comp_affected=comp_affected):
...
+        elif isinstance(bandwidth, str):  # Sometimes the filter is just a name like "red".
+            return bandwidth
+        # The 3rd case is a tuple of floats
+        elif not isinstance(bandwidth, (tuple, list)) and all(isinstance(v, float) for v in bandwidth):
```
tepals, line 349: I'm confused about when it would hit this `elif`, is that if bandwidth is a dict and all keys in the dict are floats? Or if bandwidth is a set? Or am I not reading it correctly?

src/odemis/odemisd/mdupdater.py
```diff
@@ -298,25 +305,106 @@ def updateLightPower(power, light=light, comp_affected=comp_affected):
...
+        """
+        if spectrograph is None:
+            return None
+
+        # For monochomators, we need to know the minimum and maximum wavelength
+        # detected based on the spectrograph settings. => Update MD_OUT_WL
+        pos = spectrograph.position.value
+        if 'slit-monochromator' not in pos:
+            logging.info("No 'slit-monochromator' axis was found, will not be able to compute monochromator bandwidth.")
+            wl = pos["wavelength"]
```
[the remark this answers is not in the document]
pieleric: yes, a "spectrograph" is an actuator with multiple axes. However, it doesn't move along the x, y, z dimensions. It has by definition, as defined in the developer manual, at least one axis "wavelength", and some optional ones like "grating" and "slit-in"... 
More precisely, wavelength is the center wavelength at which the spectrograph is currently configured in. It can also be "0", in which case the spectrograph is in a "0th order" mode, which basically means it's disabled and passes the image as-is.

src/odemis/odemisd/mdupdater.py
```diff
@@ -298,25 +305,106 @@ def updateLightPower(power, light=light, comp_affected=comp_affected):
...
+            if wl < 10e-9:
```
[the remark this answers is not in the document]
pieleric: Typically there is no chance to have anything below 200nm. < 10nm just means it's in 0th order mode. I'll add a comment to clarify.

src/odemis/odemisd/mdupdater.py
pieleric: Unfortunately, to test this specific case, I'd need to build a whole new base for the test cases. I'll skip it for this time.

tepals (approved):

PR #3037, opened 2025-02-21
[fix][METEOR-1332] Add stage-bare position to stitched image metadata

patrickcleeve2:
For SEM-FM guided correlation, we want to use the stage-bare metadata to pre-transform the images. However, overview (stitched) images do not correctly store this metadata (it is currently set as the first tile metadata).

This PR adds a calculation to get the stage-bare position for the overview image as a whole, and add it to the metadata.

### DIFF
--- a/src/odemis/acq/stitching/_weaver.py
+++ b/src/odemis/acq/stitching/_weaver.py
@@ -48,6 +48,7 @@ class Weaver(metaclass=ABCMeta):
         self.tbbx_px = None  # the bounding boxes of each tile in pixel coordinates
         self.gbbx_px = None  # the global bounding box of the weaved image in pixel coordinates
         self.gbbx_phy = None  # the global bounding box of the weaved image in physical coordinates
+        self.stage_bare_pos = None # the stage-bare position of the weaved image
 
     def addTile(self, tile):
         """
@@ -85,7 +86,7 @@ class Weaver(metaclass=ABCMeta):
             tiles.append(img.rotate_img_metadata(tile, -rotation, center_of_rot))
         self.tiles = tiles
 
-        self.tbbx_px, self.gbbx_px, self.gbbx_phy = self.get_bounding_boxes(self.tiles)
+        self.tbbx_px, self.gbbx_px, self.gbbx_phy, self.stage_bare_pos = self.get_bounding_boxes(self.tiles)
         im = self.weave_tiles()
         md = self.get_final_metadata(self.tiles[0].metadata.copy())
         weaved_image = img.rotate_img_metadata(model.DataArray(im, md), rotation, center_of_rot)
@@ -111,6 +112,7 @@ class Weaver(metaclass=ABCMeta):
         :return tbbx_px: (list of tuples) the ltrb bounding boxes of each tile in pixel coordinates
         :return gbbx_px: (list of tuples) the global ltrb bounding box of the weaved image in pixel coordinates
         :return gbbx_phy: (list of tuples) the global ltrb bounding box of the weaved image in physical coordinates
+        :return mean_stage_bare_pos: Dict[str, float] the mean stage-bare position of the weaved image, typically 5D
         """
 
         # Get a fixed pixel size by using the first one
@@ -119,6 +121,7 @@ class Weaver(metaclass=ABCMeta):
         pxs = tiles[0].metadata[model.MD_PIXEL_SIZE]
 
         tbbx_phy = []  # tuples of ltrb in physical coordinates
+        stage_bare_coords = []  # dict of stage-bare coordinates
         for t in tiles:
             c = t.metadata[model.MD_POS]
             w = t.shape[-1], t.shape[-2]
@@ -129,6 +132,17 @@ class Weaver(metaclass=ABCMeta):
                    c[0] + (w[0] * pxs[0] / 2), c[1] + (w[1] * pxs[1] / 2))
 
             tbbx_phy.append(bbx)
+
+            # add the stage-bare position
+            sbc = t.metadata.get(model.MD_STAGE_POSITION_RAW, None)
+            if sbc is not None:
+                stage_bare_coords.append(copy.deepcopy(sbc))
+
+        # get the mean of stage-bare coords for each axis
+        mean_stage_bare_pos = None
+        if stage_bare_coords:
+            axes = stage_bare_coords[0].keys()
+            mean_stage_bare_pos = {k: numpy.mean([r[k] for r in stage_bare_coords]) for k in axes}
 
         gbbx_phy = (min(b[0] for b in tbbx_phy), min(b[1] for b in tbbx_phy),
                     max(b[2] for b in tbbx_phy), max(b[3] for b in tbbx_phy))
@@ -153,7 +167,7 @@ class Weaver(metaclass=ABCMeta):
         if numpy.greater(gbbx_px[-2:], 4 * numpy.sum(tbbx_px[-2:])).any():
             # Overlap > 50% or missing tiles
             logging.warning("Global area much bigger than sum of tile areas")
-        return tbbx_px, gbbx_px, gbbx_phy
+        return tbbx_px, gbbx_px, gbbx_phy, mean_stage_bare_pos
 
     def get_final_metadata(self, md: dict) -> dict:
         """
@@ -170,6 +184,14 @@ class Weaver(metaclass=ABCMeta):
 
         md[model.MD_POS] = c_phy
         md[model.MD_DIMS] = "YX"
+
+        # add stage bare position to metadata if available
+        if self.stage_bare_pos is not None:
+            md[model.MD_STAGE_POSITION_RAW] = self.stage_bare_pos
+            try: # might not be present, but required for meteor
+                md[model.MD_EXTRA_SETTINGS]["Stage"]["position"][0] = self.stage_bare_pos
+            except KeyError:
+                pass
         return md
 
     def _adjust_brightness(self, tile, tiles):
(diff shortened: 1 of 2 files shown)


src/odemis/acq/stitching/_weaver.py
```diff
@@ -153,7 +166,7 @@ def get_bounding_boxes(tiles: list):
         if numpy.greater(gbbx_px[-2:], 4 * numpy.sum(tbbx_px[-2:])).any():
             # Overlap > 50% or missing tiles
             logging.warning("Global area much bigger than sum of tile areas")
-        return tbbx_px, gbbx_px, gbbx_phy
+        return tbbx_px, gbbx_px, gbbx_phy, mean_stage_bare_pos
```
pieleric, line 169: Please update the docinfo with a description for this extra return. Especially it's helpful to "remind" that the STAGE_POSITION_RAW can contain more than 2 axes (eg, 5).

pieleric (changes requested):
Just a minor comment on the code.

However, I don't fully get why you need to have the stage-bare coordinates on the overview images. I assumed that all images are in the sample coordinates. That should be enough to overlay SEM & FM images, right? Why do you need to (futher) "pre-transform" them?

patrickcleeve2:
> Just a minor comment on the code.
> 
> However, I don't fully get why you need to have the stage-bare coordinates on the overview images. I assumed that all images are in the sample coordinates. That should be enough to overlay SEM & FM images, right? Why do you need to (futher) "pre-transform" them?

Images are displayed in sample-stage coordinates (MD_POS) which is effectively 2D, but to transform between reference frames e.g. SEM->FM you need the stage-bare (MD_STAGE_POSITION_RAW, 5D) coordinates.

pieleric:
> Images are displayed in sample-stage coordinates (MD_POS) which is effectively 2D, but to transform between reference frames e.g. SEM->FM you need the stage-bare (MD_STAGE_POSITION_RAW, 5D) coordinates.

Can you clarify further. In my mind, the sample-stage referential are precisely defined so that one given point on the sample (when seen from a perpendicular point of view) has always the same X & Y, independent of the current posture (modulo the imprecision of the transform). So I assumed that the MD_POS of a tile is *really* in the same referential whether it's FM or SEM.  Is that not the case? Wasn't it the design goal? What's missing to make it work like that? What's the difference in MD_POS between a point in FM and SEM?

patrickcleeve2:
> > Images are displayed in sample-stage coordinates (MD_POS) which is effectively 2D, but to transform between reference frames e.g. SEM->FM you need the stage-bare (MD_STAGE_POSITION_RAW, 5D) coordinates.
> 
> Can you clarify further. In my mind, the sample-stage referential are precisely defined so that one given point on the sample (when seen from a perpendicular point of view) has always the same X & Y, independent of the current posture (modulo the imprecision of the transform). So I assumed that the MD_POS of a tile is _really_ in the same referential whether it's FM or SEM. Is that not the case? Wasn't it the design goal? What's missing to make it work like that? What's the difference in MD_POS between a point in FM and SEM?

That is the end goal, and this is a step towards that. Essentially, that assumes the SEM is the 'base' posture, and every stage-bare coordinate is transforms to that, and then projected to the sample-stage plane. 
E.g
1. image acquired at FM
2. stage-bare-sem = to_posture(stage-bare-fm, SEM)
3. sample-stage = to_sample_stage_from_stage_bare(stage-bare-sem) -> this is MD_POS you're taking about

But before any of that can happen, the basic building blocks need to be tested and understood.  We can't just ignore the practical inaccuracies of the current transform system. So for now, each image needs a stage-bare coordinate so it can be transformed into other postures, and each MD_POS is on the same 'plane' but at a different x/y position. Once that works, there is disabled code to 'auto transform' each stage-bare position to SEM posture.

pieleric (approved):

src/odemis/acq/stitching/_weaver.py
```diff
@@ -130,6 +133,17 @@ def get_bounding_boxes(tiles: list):
 
             tbbx_phy.append(bbx)
 
+            # add the stage-bare position
+            sbc = t.metadata.get(model.MD_STAGE_POSITION_RAW, None)
+            if sbc is not None:
+                stage_bare_coords.append(copy.deepcopy(sbc))
```
tmoerkerken, line 139:
```suggestion
                stage_bare_coords.append({**sbc})
```
Would be way more performant for this datastructure:
```
>>> meth1 = lambda lst, dct: lst.append(deepcopy(dct))
>>> meth2 = lambda lst, dct: lst.append({**dct})
>>> a = {"x": 0.0, "y": 1.0, "z": 2.0, "rx": 0.0, "rz": 0.0}
>>> dct1, dct2 = [], []
>>> timeit.timeit(functools.partial(meth1, dct1, a), number=10000)
0.2051957930016215
>>> timeit.timeit(functools.partial(meth2, dct2, a), number=10000)
0.008122990999254398
```

src/odemis/acq/stitching/_weaver.py
```diff
@@ -170,6 +184,14 @@ def get_final_metadata(self, md: dict) -> dict:
 
         md[model.MD_POS] = c_phy
         md[model.MD_DIMS] = "YX"
+
+        # add stage bare position to metadata if available
+        if self.stage_bare_pos is not None:
+            md[model.MD_STAGE_POSITION_RAW] = self.stage_bare_pos
+            try: # might not be present, but required for meteor
+                md[model.MD_EXTRA_SETTINGS]["Stage"]["position"][0] = self.stage_bare_pos
```
tmoerkerken, line 192: Why is this `position` property a list? Are there multiple stage positions stored? And if so, wouldn't a dict then by more explanatory?

src/odemis/acq/stitching/test/weaver_test.py
```diff
@@ -223,6 +229,87 @@ def test_rotated_tiles(self):
...
+                "rz": 0,
+            },
+            model.MD_PIXEL_SIZE: (px, px),
+        }
+        in4 = model.DataArray(img, md4)
+
+        # calculate mean stage-bare for each axes
+        stage_bares = []
+        for md in [md1, md2, md3, md4]:
+            stage_bares.append(copy.deepcopy(md[model.MD_STAGE_POSITION_RAW]))
```
tmoerkerken, line 290: could just use dict unpacking

src/odemis/acq/stitching/test/weaver_test.py
```diff
@@ -223,6 +229,87 @@ def test_rotated_tiles(self):
...
+        mean_stage_bare = {}
+        for key in stage_bares[0].keys():
+            mean_stage_bare[key] = numpy.mean([md[key] for md in stage_bares])
+
+        if self.weaver_type == WEAVER_COLLAGE:
```
tmoerkerken, line 295: This if/elif block is all over the codebase. Can't we just expand the constants.py a bit to to this mapping just once? 

```
WEAVER_MAPPING = {
    "WEAVER_COLLAGE": CollageWeaver,
    "WEAVER_COLLAGE_REVERSE": CollageWeaverReverse,
    "WEAVER_MEAN": MeanWeaver,
}
```
And then here:

`weaver = WEAVER_MAPPING.get(self.weaver_type)`

src/odemis/acq/stitching/test/weaver_test.py
```diff
@@ -223,6 +229,87 @@ def test_rotated_tiles(self):
...
+        weaver.addTile(in2)
+        weaver.addTile(in3)
+        weaver.addTile(in4)
+        outd = weaver.getFullImage()
+
+        # check that the mean stage-bare position is correct
+        for axes in mean_stage_bare.keys():
+            self.assertAlmostEqual(outd.metadata[model.MD_STAGE_POSITION_RAW][axes],
+                                   mean_stage_bare[axes],
+                                   places=3)
```
tmoerkerken, line 312: If the only deviation between the first and second argument is a potential floating point error, we can safely increase the `places` argument a little bit right?

tmoerkerken (changes requested):

src/odemis/acq/stitching/_weaver.py
patrickcleeve2: the extra settings are stored as val, unit in a list. [0] is the value, [1] is the unit

src/odemis/acq/stitching/test/weaver_test.py
patrickcleeve2: not relevant for this PR

src/odemis/acq/stitching/test/weaver_test.py
patrickcleeve2: for what purpose?

src/odemis/acq/stitching/_weaver.py
patrickcleeve2: It's not obvious what that unpacking function is doing, and if it is even a deep copy. prefer not to

src/odemis/acq/stitching/test/weaver_test.py
patrickcleeve2: need deepcopy, see above

src/odemis/acq/stitching/test/weaver_test.py
tmoerkerken: The values are in the 1e-9 range, so checking for 3 places feels a bit 'inaccurate', but I guess it's fine for now.

src/odemis/acq/stitching/_weaver.py
tmoerkerken: I see. A dict would have been more self-explanatory, but out of scope for now.

tmoerkerken (approved):

PR #3039, opened 2025-03-06
[fix] Independent EBIC detector might not have MD_POS

pieleric:
The MD_POS needs to be copied from the scanner metadata. However, the
acquisition restart function could easily stop the scanner *just* before
receiving the data (and metadata).
=> Wait explicitly for the scanner data before restarting the
acquisition.

Also fix a few issues in the test cases and simulator.

### DIFF
--- a/src/odemis/acq/stream/_helper.py
+++ b/src/odemis/acq/stream/_helper.py
@@ -1529,6 +1529,7 @@ class IndependentEBICStream(EBICSettingsStream):
 
         self._emitter_dataflow = emt_dataflow
         self._latest_emitter_md = {}  # the metadata of the emitter, to be copied onto the detector data
+        self._emitter_data_received = threading.Event()  # set when the emitter data (corresponding to one scan) is received
         self._acq_start_lock = threading.Lock()  # To be taken when starting/stopping acquisition
 
         # Change the standard dwell time setter to set both the detector and the emitter.
@@ -1550,6 +1551,11 @@ class IndependentEBICStream(EBICSettingsStream):
         """
         # Override standard method (on LiveStream), which checks whether the acquisition time has
         # changed "a lot", and in such case stop and restart the acquisition.
+        # Compared to the standard method, here, in adition to the detector, the scanner must also
+        # be restarted, which is more complex (see _restart_acquisition()). In addition, currently,
+        # this function is called just after active is set (although it's not stricly necessary), and
+        # this can trigger a restart while we are still starting the acquisition. All this is possible
+        # to handle, but for now, we keep it simple and just don't implement the functionality.
 
         prev_dur = self._prev_dur
         self._prev_dur = self.estimateAcquisitionTime()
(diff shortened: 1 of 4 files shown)


src/odemis/acq/stream/_helper.py
```diff
@@ -1550,6 +1551,9 @@ def _updateAcquisitionTime(self) -> None:
         """
         # Override standard method (on LiveStream), which checks whether the acquisition time has
         # changed "a lot", and in such case stop and restart the acquisition.
+        # Here, we would need to restart both the scanner and detector, which is more complex...
+        # especially because this function is called just after active is set, which would cause an
+        # immediate restart.
```
tepals, line 1556: This comment raises more questions for me than that it explains anything. For instance why would the scanner and detector need to be restarted, and compared to what is it more complex?

src/odemis/acq/stream/_helper.py
```diff
@@ -1611,22 +1619,29 @@ def _onNewData(self, dataflow: model.DataFlow, data: model.DataArray) -> None:
...
+    def _restart_acquisition(self, expect_emt_data: bool):
         """
         Stops and immediately restarts the acquisition.
         This is to force the detectors to start a new acquisition, synchronized with the scanner.
         """
         try:
             with self._acq_start_lock:
                 if self.is_active.value:
+                    # Make sure that the emitter data is received before restarting the acquisition,
+                    # otherwise, we might its metadata.
```
tepals, line 1634:
```suggestion
                    # otherwise, we might miss its metadata.
```

src/odemis/acq/stream/_helper.py
```diff
@@ -1611,22 +1619,29 @@ def _onNewData(self, dataflow: model.DataFlow, data: model.DataArray) -> None:
...
         # We cannot re-subscribe to the dataflow directly from here, as it's called from the dataflow,
         # which would cause a deadlock. So we do this in a separate thread.
-        t = threading.Thread(target=self._restart_acquisition)
+        t = threading.Thread(target=self._restart_acquisition, args=(True,))
         t.start()
 
-    def _restart_acquisition(self):
+    def _restart_acquisition(self, expect_emt_data: bool):
         """
         Stops and immediately restarts the acquisition.
```
tepals, line 1627: the "immediately" is now technically not true anymore

tepals (changes requested):

src/odemis/acq/test/stream_test.py
```diff
@@ -4185,7 +4185,7 @@ def test_ebic_live_stream(self):
         """
         ebics = stream.IndependentEBICStream("test ebic", self.ebic, self.ebic.data,
                                              self.ebeam, self.sed.data,
-                                             detvas={"dwellTime"})
+                                             emtvas={"dwellTime"})
```
tmoerkerken, line 4188: I don't see how these changes to this test are related to the PR topic (fixing setting of metadata). Would have expected a test where we actually check the metadata.

tmoerkerken (approved):

src/odemis/acq/test/stream_test.py
pieleric: You're right. It's actually a fix for the test cases. I found the test case were broken when running them after the changes... but it's not directly related. I can put these changes in a separate commit.

tepals (approved):

PR #3045, opened 2025-03-17
[fix] acquisition time in estimate_time test case for tiled acquisition

K4rishma:
_____________________ CRYOSECOMTestCase.test_estimate_time _____________________
tiledacq_test.py:499: in test_estimate_time
    self.assertTrue(min_total_time <= tiled_acq_task.estimateTime() <= 1.30 * min_total_time)
E   AssertionError: False is not true

### DIFF
--- a/src/odemis/acq/stitching/test/tiledacq_test.py
+++ b/src/odemis/acq/stitching/test/tiledacq_test.py
@@ -28,12 +28,13 @@ import numpy
 
 import odemis
 from odemis import model
-from odemis.acq import stream
+from odemis.acq import stream, acqmng
 from odemis.acq.acqmng import SettingsObserver
 from odemis.acq.stitching import WEAVER_COLLAGE_REVERSE, REGISTER_IDENTITY, \
     WEAVER_MEAN, acquireTiledArea, FocusingMethod
 from odemis.acq.stitching._tiledacq import (TiledAcquisitionTask, get_fov, get_zstack_levels, clip_tiling_bbox_to_range,
                                             get_stream_based_bbox, get_tiled_bboxes, get_fov_based_bbox)
+from odemis.acq.stream import FluoStream
 from odemis.util import testing, img
 from odemis.util.comp import compute_camera_fov, compute_scanner_fov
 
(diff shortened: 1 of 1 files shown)


tepals:
can you update the commit message and the description of this PR a bit better to describe which test case was failing and what the error message was?

src/odemis/acq/stitching/test/tiledacq_test.py
```diff
@@ -490,11 +491,14 @@ def test_estimate_time(self):
...
+        total_tiles = tiled_acq_task._nx * tiled_acq_task._ny
+        # Observed time duration during stage movement between tiles
+        min_stage_time = total_tiles * 0.3
+        # Calculated time duration during tile acquisition
+        zlevels_dict = {s: tiled_acq_task._zlevels for s in tiled_acq_task._streams
+                        if isinstance(s, (FluoStream))}
+        acq_time = acqmng.estimateZStackAcquisitionTime(tiled_acq_task._streams, zlevels_dict) * total_tiles
+        min_total_time = min_stage_time + acq_time
         # overhead of 30% on min_total_time is used to account for stream settings, like, exposure time etc.
         self.assertTrue(min_total_time <= tiled_acq_task.estimateTime() <= 1.30 * min_total_time)
```
tepals, line 503: while you are fixing this test case, can you also replace this `assertTrue`. `assertTrue` should be used when checking if a certain boolean is True, for this case `assertLess` and `assertGreater` work better (please double check if I didn't make a mistake in this suggestion). This will also make sure the error message is more informative
```suggestion
        estimated_time = tiled_acq_task.estimateTime()
        self.assertLessEqual(min_total_time, estimated_time )
        self.assertLessEqual(estimated_time, 1.30 * min_total_time)
```

src/odemis/acq/stitching/test/tiledacq_test.py
```diff
@@ -503,11 +507,14 @@ def test_estimate_time(self):
...
```
tepals, line 519: same here, can you replace `assertTrue`?

tepals (changes requested):

tmoerkerken (approved):

tepals (approved):

PR #3046, opened 2025-03-17
[fix] bypass the stage movement check if the stage is already close t…

K4rishma:
…o the requested position

Checked by running the GUI on xt_client sim server and test cases of xt_client


pieleric:
Could you check my take on this, in #3055 ? I think that if we don't care that the stage is still at the same position as orig_pos, then the code can be simplified quite a bit.

PR #3047, opened 2025-03-17
[TRCL-611][feat] SEMTemporalMDStream now use synchronizing event on the time-correlator

pieleric:
The picoquant driver used to not support synchronization, so the
acquisition code had to start/stop explicitly the time-correlator
acquisition for each pixel. Now that the driver supports
synchronization, we can use the same pattern as the standard SPARC
acquisition.

This brings 2 advantages:
* avoids opening/closing "things" like the shutter and blanker for each
  pixel. This used to cause ~1s overhead per pixel.
* makes the code closer to the standard code, so we are one (little)
  step closer to merging this code into the standard code.

Also adjust the test cases to still work, now that the acquisition is
much faster.

### DIFF
--- a/src/odemis/acq/test/stream_test.py
+++ b/src/odemis/acq/test/stream_test.py
@@ -4317,19 +4317,10 @@ class TimeCorrelatorTestCase(unittest.TestCase):
     """
     Tests the SEMTemporalMDStream.
     """
-    backend_was_running = False
 
     @classmethod
     def setUpClass(cls):
-        try:
-            testing.start_backend(TIME_CORRELATOR_CONFIG)
-        except LookupError:
-            logging.info("A running backend is already found, skipping tests")
-            cls.backend_was_running = True
-            return
-        except IOError as exp:
-            logging.error(str(exp))
-            raise
+        testing.start_backend(TIME_CORRELATOR_CONFIG)
 
         # Find CCD & SEM components
         cls.time_correlator = model.getComponent(role="time-correlator")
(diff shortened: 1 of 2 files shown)


src/odemis/acq/test/stream_test.py
```diff
@@ -4420,9 +4401,9 @@ def test_acquisition(self):
         self.assertIn(self.ebeam.dwellTime.value, (1, 1e-6))
         data, exp = f.result()
         self.assertIsNone(exp)
-        # Dwell time on detector and emitter should be back to normal
-        self.assertEqual(self.time_correlator.dwellTime.value, 2)
-        self.assertEqual(self.ebeam.dwellTime.value, 0.042)
+        # # Dwell time on detector and emitter should be back to normal
+        # self.assertEqual(self.time_correlator.dwellTime.value, 2)
+        # self.assertEqual(self.ebeam.dwellTime.value, 0.042)
```
tepals, line 4406: why is this commented?

pieleric: It should be completely dropped. We typically don't need the settings to be set back to the previous value. I've removed the code that does it in the stream, and so the test has to drop these checks.

tmoerkerken (approved):
Minor suggestions

src/odemis/acq/test/stream_test.py
```diff
@@ -4344,16 +4335,6 @@ def setUpClass(cls):
...
-    def tearDownClass(cls):
-        if cls.backend_was_running:
-            return
-        testing.stop_backend()
-
-    def setUp(self):
-        if self.backend_was_running:
-            self.skipTest("Running backend found")
-
     def test_acquisition(self):
```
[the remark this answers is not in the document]
pieleric: Unfortunately, it's not that easy, because the simulator (`sparc2-time-correlator-sim.odm.yaml`) doesn't simulate the sem API, so there is no blanker/external control available.
We could extend the simulator... but on the other hand this code is a duplicate of the one used on the "normal" SPARC acquisition, so I'm already pretty confident it works.

tepals (approved):

PR #3055, opened 2025-03-21
[fix] xt_client: bypass the stage movement check if the stage is already close

pieleric:
If requested to move somewhere very close to the current position, the
TFS stage doesn't move. However the code would be waiting very long
because it's still at its original position.
=> Just give up being clever, and only check that the stage is near the target position.

Not tested on real hardware!

### DIFF
--- a/src/odemis/driver/xt_client.py
+++ b/src/odemis/driver/xt_client.py
@@ -2329,8 +2329,6 @@ class Stage(model.Actuator):
                     target_pos = {ax: val + orig_pos[ax] for ax, val in pos.items()}
                 else:
                     target_pos = pos
-                # Drop axes from the original position, which are not important because they have not moved
-                orig_pos = {a: orig_pos[a] for a, nv in target_pos.items() if nv != orig_pos[a]}
 
                 # Wait until the move is over.
                 # Don't check for future._must_stop because anyway the stage will
@@ -2377,10 +2375,8 @@ class Stage(model.Actuator):
                 # The stage is not moving anymore, however in some rare cases, the server still reports the old
                 # position for a short while (typically < 1s). This can cause all sorts of confusion in the calling
                 # code, as it would seem that the move didn't have any effect, and later on, after the position
-                # is eventually updated that the stage has moved unexpectedly. So, wait until the reported position is
-                # (1) not identical to the starting position and (2) not too far from the target position.
+                # is eventually updated that the stage has moved unexpectedly. So, wait until close to the target position.
 
-                # The stage is not moving anymore, but we still want to wait until the position has been updated
                 # TODO: base the timeout on stage movement time estimation, needs to be checked on hardware
                 wait_stage_move = 10  # s
                 expected_end_time = time.time() + wait_stage_move  # s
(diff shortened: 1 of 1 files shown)


src/odemis/driver/xt_client.py
```diff
@@ -2404,40 +2400,31 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
...
-                    if all(current_pos[a] != op for a, op in orig_pos.items()):
-
-                        if axes_to_check:
-                            axes_updated = isNearPosition(current_pos=current_pos, target_position=target_pos,
-                                                          axes=axes_to_check, rot_axes=rotational_axes_to_check,
-                                                          atol_linear=tol_linear, atol_rotation=tol_rotation)
-                        else:
-                            axes_updated = True  # No axis to check => move is done
+                    if not axes_to_check:
+                        break
```
K4rishma, line 2404: Maybe we can put this as a condition in while, as while axes_to_check: instead of while True:

K4rishma (approved):

src/odemis/driver/xt_client.py
```diff
@@ -2402,42 +2398,29 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
...
-                            logging.debug("Waiting a little longer as position has not updated fully: %s != %s",
-                                          current_pos, target_pos)
-                    else:
-                        logging.debug("Waiting for position to update: %s == %s (for some axes)",
-                                      orig_pos, current_pos)
+                while axes_to_check:
+                    if isNearPosition(current_pos=current_pos, target_position=target_pos,
+                                      axes=axes_to_check, rot_axes=rotational_axes_to_check,
+                                      atol_linear=tol_linear, atol_rotation=tol_rotation):
+                        # Drop axes from the original position, which are not important because they have not moved
```
tepals, line 2405:
```suggestion
                        # Drop axes from the original position, which are not important because they have not been requested to move
```

src/odemis/driver/xt_client.py
```diff
@@ -2377,10 +2375,8 @@ def _moveTo(self, future: futures.Future, pos: Dict[str, float], rel: bool = Fal
                 # The stage is not moving anymore, however in some rare cases, the server still reports the old
                 # position for a short while (typically < 1s). This can cause all sorts of confusion in the calling
                 # code, as it would seem that the move didn't have any effect, and later on, after the position
-                # is eventually updated that the stage has moved unexpectedly. So, wait until the reported position is
-                # (1) not identical to the starting position and (2) not too far from the target position.
+                # is eventually updated that the stage has moved unexpectedly. So, wait until close to the target position.
 
-                # The stage is not moving anymore, but we still want to wait until the position has been updated
                 # TODO: base the timeout on stage movement time estimation, needs to be checked on hardware
                 wait_stage_move = 10  # s
```
tepals, line 2381: 10 seconds still feels quite long for the default timeout

tepals (changes requested):

src/odemis/driver/xt_client.py
pieleric: A timeout should be something that feels really long. It should be there as better behaviour than "waiting forever". Here, in this function, normally the new stage position is immediately provided. In some cases, we need to wait a little bit that it's provided (due to a bug in the TFS software). If that also fails, then we give up after 10s. I think for very long moves, there could be merit to wait *longer*.

Anyway, let's not change this part now. I'd rather do this in a separate task/PR.

tepals (approved):

PR #3056, opened 2025-03-21
[fix] [METEOR-1347] Tescan e-beam current API use

tmoerkerken:
Use SetBeamCurrent/GetBeamCurrent over the deprecated SetPCIndex API method. As a consequence, we don't have to do list index handling and we can have a continuous VA. Since the current range is device specific, a default value is introduced that is over-writable via the yaml.

And separately, I also introduced a log scale, since there was otherwise no differentiation on the slider for smaller values.

Tescan Sim
image
Odemis
image

### DIFF
--- a/src/odemis/driver/tescan.py
+++ b/src/odemis/driver/tescan.py
@@ -46,6 +46,7 @@ ACQ_CMD_TERM = 2
 # a spot with less than 100 integrations it gets an enormous delay to receive
 # new data from the server.
 TESCAN_PXL_LIMIT = 100
+PROBE_CURRENT_RANGE = (10e-12, 100e-9)
 
 
 class SEM(model.HwComponent):
@@ -539,7 +540,7 @@ class Scanner(model.Emitter):
     etc. Similarly it subscribes to the VAs of scale and magnification in order
     to update the pixel size.
     """
-    def __init__(self, name, role, parent, fov_range, **kwargs):
+    def __init__(self, name, role, parent, fov_range, current_range=PROBE_CURRENT_RANGE, **kwargs):
         # It will set up ._shape and .parent
         model.Emitter.__init__(self, name, role, parent=parent, **kwargs)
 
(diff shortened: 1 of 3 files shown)


src/odemis/driver/tescan.py
```diff
@@ -613,13 +613,10 @@ def __init__(self, name, role, parent, fov_range, **kwargs):
...
-        # Enumerated float with respect to the PC indexes of Tescan API
-        self._list_currents = self.GetProbeCurrents()
-        pc_choices = set(self._list_currents)
-        # We use the current PC
-        pc = self._list_currents[self.parent._device.GetPCIndex() - 1]
-        self.probeCurrent = model.FloatEnumerated(pc, pc_choices, unit="A",
-                                                  setter=self._setPC)
+        pc = self.parent._device.GetBeamCurrent() * 1e-12  # Convert from pA to A
+        # For limits of current, values from the Tescan UI are used, since the API did not specify any.
+        self.probeCurrent = model.FloatContinuous(pc, (10 * 1e-12, 100 * 1e-9), unit="A",
```
pieleric, line 618: The range is not going to be the same for all SEMs. I also don't see any function to get this info from the API. I'd suggest to have a default value (please make it a constant, like RNG_PROBE_CURRENT), and have a parameter `rng_current` in the init method to override it (from the microscope file).

src/odemis/gui/conf/data.py
```diff
@@ -168,7 +168,10 @@
             }),
             ("probeCurrent", {
                 "label": "Beam current",
-                "event": wx.EVT_SCROLL_CHANGED  # only affects when it's a slider
```
pieleric, line 171: This is not specific to the Tescan. I agree it's probably best to use a log scale. However, please put it in a separate commit (same PR).

pieleric (changes requested):

src/odemis/gui/conf/data.py
tmoerkerken: Done

src/odemis/driver/tescan.py
tmoerkerken: I introduced a default that is over-writable via the yaml (current_range, which is more in line with the already existing fov_range)

pieleric (approved):

PR #3060, opened 2025-03-26
[feat] FVEM-40 Disable offloading of calibration data to the technolution storage server

tepals:
Previously we set the filename of calibration data to .calibrations to ensure that the calibration data did not get post processed. In v3.0 Technolution added support for disabling the offloading of data to the storage server

Also removed 2 TODOs in a separate commit in this PR, since the TODOs have been implemented by now.

### DIFF
--- a/src/odemis/driver/technolution.py
+++ b/src/odemis/driver/technolution.py
@@ -1245,9 +1245,8 @@ class MPPC(model.Detector):
 
         md = self.getMetadata()
         custom_data = md.get(model.MD_EXTRA_SETTINGS, "")  # if no custom metadata, pass an empty string
-        # TODO support USER_NOTE in GUI
         info = md.get(model.MD_USER_NOTE, None)  # if no user note pass None, so it is not added to the metadata.yaml
-        z_position = md.get(model.MD_SLICE_IDX, 0)  # if slice number is not provided use 0. TODO support in GUI
+        z_position = md.get(model.MD_SLICE_IDX, 0)  # if slice number is not provided use 0.
         eff_field_size = md.get(model.MD_FIELD_SIZE, self._scanner.resolution.value)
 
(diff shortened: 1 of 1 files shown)


src/odemis/driver/technolution.py
```diff
@@ -1245,15 +1245,17 @@ def _assembleMegafieldMetadata(self):
 
         md = self.getMetadata()
         custom_data = md.get(model.MD_EXTRA_SETTINGS, "")  # if no custom metadata, pass an empty string
-        # TODO support USER_NOTE in GUI
         info = md.get(model.MD_USER_NOTE, None)  # if no user note pass None, so it is not added to the metadata.yaml
-        z_position = md.get(model.MD_SLICE_IDX, 0)  # if slice number is not provided use 0. TODO support in GUI
+        z_position = md.get(model.MD_SLICE_IDX, 0)  # if slice number is not provided use 0.
         eff_field_size = md.get(model.MD_FIELD_SIZE, self._scanner.resolution.value)
 
+        offload_disable = True if ".calibrations" in self.filename.value else False
```
nandishjpatel, line 1252: Is this line change also related to this TODO in fastem-calibrations here: https://bitbucket.org/delmic/fastem-calibrations/src/c9506cfbc9b95cb9cd85fc4ee7fb41441e1c91a6/src/fastem_calibrations/configure_hw.py#lines-199

tepals: It is somewhat related. The idea in that TODO is that there would be a function in the ASM such that we can upload settings without having to acquire an image. So for instance change the descanner offset. If we would have that functionality we would not need to acquire calibrations images and this change would not be necessary

nandishjpatel (approved):

K4rishma (approved):

PR #3064, opened 2025-03-28
[feat] Improve GUI workflow for simultaneous EBIC & CL acquisition

pieleric:
On the SPARC, users with an EBIC and PMT detectors can acquire both at the same time. This happens automatically if the dwell time and repetition settings of the streams are identical. However, the EBIC "independent detector" has limited dwell time options. This would make it very hard to select the same dwell time for the two streams.

=> Make sure the dwell time of the independent detector is always a valid dwell time for the hardware, and consider that the dwell times are the same if they are within 1%. Finally, make sure that in case the dwell times are not identical, they are "reconciled" at the start of the acquisition (instead of failing).


pieleric:
Closing, as it's a duplicate of PR #3051 !

PR #3029, opened 2025-02-10
[feat] Integrate OpenFIBSEM Milling

patrickcleeve2:
Integrate openfibsem milling manager for cross-platform milling
Requires:
- https://github.com/delmic/odemis/pull/3000

PR #3065, opened 2025-03-31
[FEAT][METEOR-1100] Meteor FIBSEM Tab

patrickcleeve2:
This is the pull request with the complete workflow.

Prerequisties:
- https://github.com/delmic/odemis/pull/3029
- https://github.com/delmic/odemis/pull/3067
- https://github.com/delmic/odemis/pull/3068
- https://github.com/delmic/odemis/pull/3037


src/odemis/driver/autoscript_client.py
```diff
@@ -1122,9 +1122,10 @@ def __init__(
         # beam current
         # NOTE: VA is named probeCurrent to match existing API
         beam_current_info = self.parent.beam_current_info(self.channel)
+        beam_current_range = beam_current_info["range"]
         self.probeCurrent = model.FloatContinuous(
```
pieleric, line 1126: Change already to a check if the length > 2, and if so, make an FloatEnumerated. Actually, to handle the corner cases when there are only 2 choices, it might be best to change the xtadapter to send in such case, instead of "range", a "choices".

pieleric:
At init, the GUI shows this error (in the log file):
```
2025-04-10 20:51:07,938 ERROR   main:417:       Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/wx/core.py", line 3282, in <lambda>
    lambda event: event.callable(*event.args, **event.kw) )
  File "/home/piel/development/odemis/src/odemis/gui/cont/milling.py", line 370, in draw_milling_tasks
    if not self.acq_cont.stream or not self.acq_cont.stream.raw:
AttributeError: 'CryoFIBAcquiredStreamsController' object has no attribute 'stream'
```
It does continue to load find afterwards though....

pieleric:
Superseded by PR #3078

PR #3068, opened 2025-03-31
[METEOR-1186] CryoFeature Refactor

patrickcleeve2:
CryoFeature Changes

- Changes to 5D stage position
- Add FM Focus position
- Add milling tasks
- Add reference image
- Add multi-posture positions
- Relevant GUI changes

Requires the following to be merged:
- https://github.com/delmic/odemis/pull/2998
- https://github.com/delmic/odemis/pull/3000
- https://github.com/delmic/odemis/pull/3001
- https://github.com/delmic/odemis/pull/3002
- https://github.com/delmic/odemis/pull/3005

PR #3075, opened 2025-04-08
[refactor][METEOR-1186] CryoFeature Refactor

pieleric:
Update of PR #3068 

CryoFeature Changes
* Changes to 5D stage position
* Add FM Focus position
* Add milling tasks
* Add reference image
* Add multi-posture positions
* Relevant GUI changes
* Extend all METEOR posture managers to support the 3D transformations


src/odemis/acq/move.py
```diff
@@ -1507,6 +1536,22 @@ def __init__(self, microscope):
             raise KeyError("The stage misses %s axes" % missed_axes)
         self.fib_column_tilt = ZEISS_FIB_COLUMN_TILT
 
+        if self.pre_tilt is None: # pre-tilt not available in the stage calib metadata
+            # First version of the microscope file had it hard-coded on the Linked YM wrapper component
+            comp = model.getComponent(name="Linked YM")
+            self.pre_tilt = comp.getMetadata()[model.MD_ROTATION_COR]
+
+        # Automatic conversion to sample-stage axes
+        self._initialise_transformation(axes=["y", "m"], rotation=self.pre_tilt)
```
tmoerkerken, line 1545: Quick check, is the 'm' axis correct? Never saw it before.

pieleric: Yes, it is. That is "another Z" axis, on the Zeiss... which makes a lot of code extra "fun"!

PR #3079, opened 2025-04-18
[ci] Remove testing on Ubuntu 20, as it is deprecated

tepals:


.github/workflows/syntax_check_cloud.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-20.04, ubuntu-22.04]
+        os: [ubuntu-22.04, ubuntu-24.04]
```
tepals, line 15: I guess we could also update this to only run on ubuntu-latest (latest LTS), it doesn't really make sense to do syntax checking on multiple Ubuntu versions. What do you guys think?

tepals: fixed! @tmoerkerken could you have a look at this PR, so we don't see the ubuntu 20 deprecation warning anymore on the other open PRs

pieleric (approved):

PR #3083, opened 2025-05-06
[fix] make use of the multi-beam tab given the detailed overview image gui layout

nandishjpatel:
Both single and multi-beam acquisition are now supported in the detailed overview image gui. The previous acquisition tab is now the multi-beam tab.

### DIFF
--- a/plugins/fastem_set_factory_values.py
+++ b/plugins/fastem_set_factory_values.py
@@ -50,14 +50,14 @@ class SetFactoryValues(Plugin):
             self.stage = main_data.stage
             self.mppc = main_data.mppc
             self.ccd = main_data.ccd
-            self._acquisition_tab = fastem_main_tab.acquisition_tab
+            self._multi_beam_tab = fastem_main_tab.acquisition_tab.multi_beam_tab
         except LookupError:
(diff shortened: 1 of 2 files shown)


tepals (approved):

plugins/fastem_set_factory_values.py
```diff
@@ -50,10 +50,10 @@ def __init__(self, microscope, main_app):
             self.stage = main_data.stage
             self.mppc = main_data.mppc
             self.ccd = main_data.ccd
-            self._acquisition_tab = fastem_main_tab.acquisition_tab
+            self._acquisition_tab = fastem_main_tab.acquisition_tab.multi_beam_tab
```
pieleric, line 53: It becomes odd that the `multi_beam_tab` is called `self._acquisition_tab`.
Actually, this variable isn't used outside of the function. So I'd recommend to rename it just `multi_beam_tab`. 
(comment valid for both plugins).

pieleric (changes requested):

K4rishma (approved):

pieleric (approved):

PR #3086, opened 2025-05-08
[feat] New plugin spectrum_laser_protection, for laser protection with SPARC FPLM

pieleric:
Changes the behaviour of the GUI, so that on a SPARC, it's not possible to turn on the laser
of a spectrum stream when the parabolic mirror is not engaged. The idea is that, although unlikely,
the user could go to the "acquisition" tab without having engaged the mirror (typically, it's not useful,
but the GUI allows this). In this case, the input laser (coming from via a FPLM module) could enter
the chamber and hit and damage some of the detectors.
The way it works in practice is that whenever the acquisition tab is shown, it checks the parabolic mirror
position. If it's not engaged it:
* sets the spectrum stream light power to 0, and disable the entry. So it's impossible to turn it on.
* when a new spectrum stream is added, it disables the power entry.
If the mirror is engaged, the power entry of all spectrum streams is re-enabled.

### DIFF
--- a/plugins/spectrum_laser_protection.py
+++ b/plugins/spectrum_laser_protection.py
@@ -1,0 +1,150 @@
+# -*- coding: utf-8 -*-
+"""
+Created on 7 March 2025
+
+@author: Éric Piel
+
+This file is part of Odemis.
+
+Odemis is free software: you can redistribute it and/or modify it under the
+terms of the GNU General Public License version 2 as published by the Free
+Software Foundation.
+
+Odemis is distributed in the hope that it will be useful, but WITHOUT ANY
+WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+PARTICULAR PURPOSE. See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with
+Odemis. If not, see http://www.gnu.org/licenses/.
+"""
+# Changes the behaviour of the GUI, so that on a SPARC, it's not possible to turn on the laser
+# of a spectrum stream when the parabolic mirror is not engaged. The idea is that, although unlikely,
+# the user could go to the "acquisition" tab without having engaged the mirror (typically, it's not useful,
+# but the GUI allows this). In this case, the input laser (coming from via a FPLM module) could enter
+# the chamber and hit and damage some of the detectors.
+# The way it works in practice is that whenever the acquisition tab is shown, it checks the parabolic mirror
+# position. If it's not engaged it:
+# * sets the spectrum stream light power to 0, and disables the entry. So it's impossible to turn it on.
+# * when a new spectrum stream is added, it disables the power entry.
+# If the mirror is engaged, the power entry of all spectrum streams is re-enabled (but left at 0).
+# For testing, use the sparc2-fplm-sim.odm.yaml microscope file.
+
+import functools
+import logging
+from typing import Dict, Callable
+
+from odemis.acq.stream import SpectrumSettingsStream
+from odemis.gui.cont.stream import StreamController
+from odemis.gui.cont.tabs import MIRROR_ENGAGED
+from odemis.gui.plugin import Plugin
+
+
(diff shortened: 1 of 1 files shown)


plugins/spectrum_laser_protection.py
```diff
@@ -0,0 +1,151 @@
...
+        if the parabolic mirror is not engaged.
+        :param name: name of the stream/action
+        :returns: the stream controller created
+        """
+        original_action = self._original_actions[name]
+        ret = original_action()
+        if not self._mirror_is_engaged:
+            self.disable_light_power()
+
+        return ret
```
tepals, line 151: can you add an empty line at the bottom?

plugins/spectrum_laser_protection.py
```diff
@@ -0,0 +1,151 @@
...
+    __version__ = "1.0"
+    __author__ = "Éric Piel"
+    __license__ = "GPLv2"
+
+    def __init__(self, microscope, main_app):
+        super().__init__(microscope, main_app)
+        # Can only be used with a SPARC with spectrometer(s)
+        main_data = self.main_app.main_data
+        if microscope and main_data.role.startswith("sparc"):
+            self._tab = main_data.getTabByName("sparc_acqui")
```
tepals, line 53: for readability in `on_tab_change`, I would suggest to rename `_tab` to `_acqui_tab`
```suggestion
            self._acqui_tab = main_data.getTabByName("sparc_acqui")
```

tepals (changes requested):

tepals (approved):

plugins/spectrum_laser_protection.py
```diff
@@ -0,0 +1,151 @@
...
+        Disable the light power of all spectrum streams
+        """
+        for sctrl in self._stb_ctrl.stream_controllers:
+            if isinstance(sctrl.stream, SpectrumSettingsStream):
+                for setting_entry in sctrl.entries:
+                    if setting_entry.name == "power":
+                        setting_entry.vigilattr.value = 0
+                        setting_entry.value_ctrl.Enable(False)
+
+
```
K4rishma, line 128: - extra line

K4rishma (approved):

plugins/spectrum_laser_protection.py
```diff
@@ -0,0 +1,151 @@
...
+        if the parabolic mirror is not engaged.
+        :param name: name of the stream/action
+        :returns: the stream controller created
+        """
+        original_action = self._original_actions[name]
+        ret = original_action()
+        if not self._mirror_is_engaged:
+            self.disable_light_power()
+
+        return ret
```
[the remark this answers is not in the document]
pieleric: What do you mean? I've already fixed the "no new line at the end". Would you expect even more empty line, or do you mean there should be less? I honestly think that the number of empty lines at the end is not very important... but it seems people care quite a lot!

plugins/spectrum_laser_protection.py
```diff
@@ -0,0 +1,151 @@
...
+# Changes the behaviour of the GUI, so that on a SPARC, it's not possible to turn on the laser
+# of a spectrum stream when the parabolic mirror is not engaged. The idea is that, although unlikely,
+# the user could go to the "acquisition" tab without having engaged the mirror (typically, it's not useful,
+# but the GUI allows this). In this case, the input laser (coming from via a FPLM module) could enter
+# the chamber and hit and damage some of the detectors.
+# The way it works in practice is that whenever the acquisition tab is shown, it checks the parabolic mirror
+# position. If it's not engaged it:
+# * sets the spectrum stream light power to 0, and disable the entry. So it's impossible to turn it on.
+# * when a new spectrum stream is added, it disables the power entry.
+# If the mirror is engaged, the power entry of all spectrum streams is re-enabled.
```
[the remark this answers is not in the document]
pieleric: Clarified that "re-enabled" only means that the user can change the value again, but it is left at 0 W.

PR #3096, opened 2025-05-13
[FVEM-135] Address V-EM GUI user feedback starting 2025

nandishjpatel:
Changes:

A-Z allowed in user profile name

image

ROIs have dashed line while ROAs have solid line

image

Columns ordered as per user convenience

image

image
 
Added scintillator column
  
image

Added reference stage button

image

User can enter custom HFW values in Single-Beam panel of Project Manager

image

### DIFF
--- a/src/odemis/gui/cont/stream.py
+++ b/src/odemis/gui/cont/stream.py
@@ -36,8 +36,10 @@ from odemis import model, util
 from odemis.acq.stream import MeanSpectrumProjection
 from odemis.gui import (CONTROL_COMBO, CONTROL_FLT, FG_COLOUR_DIS,
                         FG_COLOUR_ERROR, FG_COLOUR_WARNING)
+from odemis.gui.comp.buttons import ImageTextButton
 from odemis.gui.comp.foldpanelbar import FoldPanelBar
 from odemis.gui.comp.overlay.repetition_select import RepetitionSelectOverlay
+from odemis.gui.comp.toggle import GraphicalToggleButtonControl
 from odemis.gui.comp.stream_panel import (OPT_BTN_PEAK, OPT_BTN_REMOVE, OPT_BTN_SHOW,
                                           OPT_BTN_TINT, OPT_BTN_UPDATE, OPT_FIT_RGB,
                                           OPT_NAME_EDIT, OPT_NO_COLORMAPS,
@@ -1389,7 +1391,7 @@ class FastEMStreamController(StreamController):
 
     def _create_immersion_mode_cbox(self):
         # HACK manually create the checkbox using custom gb_sizer's pos and span
-        # The hack is needed because self.stream_paneladd_checkbox_control cannot be overridden,
+        # The hack is needed because self.stream_panel.add_checkbox_control cannot be overridden,
         # which has pos=(self.num_rows, 1) and span=(1, 2)
         _ = self.stream_panel._add_side_label("Immersion mode")
         cbox_immersion_mode = wx.CheckBox(self.stream_panel._panel, wx.ID_ANY,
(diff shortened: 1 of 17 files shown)


src/odemis/gui/cont/tabs/fastem_main_tab.py
```diff
@@ -248,8 +249,12 @@ def _on_focussed_view(self, focussed_view):
         for viewport in self.tab_data_model.viewports.value:
```
pieleric, line 249: I *think* this whole function could be moved into the viewport itself (FastEMMainViewport). By overriding .SetFocus(), each viewport will know when it becomes the focused view(port) or not. This would avoid the (explicit) `for` loop, and the `if` would become just `if focus`.

src/odemis/gui/cont/tabs/fastem_main_tab.py
```diff
@@ -248,8 +249,12 @@ def _on_focussed_view(self, focussed_view):
...
             if viewport.canvas.view.name.value == focussed_view.name.value:
                 viewport.cpol.active.value = True
+                colour = hex_to_frgba(SELECTION_COLOUR)
             else:
                 viewport.cpol.active.value = False
+                colour = (1.0, 1.0, 1.0)  # white
+            if viewport.canvas.bg_view_label:
```
pieleric, line 256: I find it odd that the canvas hold a reference to an object part of the text overlay. I think it'd be "cleaner" to not hold that `bg_view_label` explicitly, and here just access it via `canvas.bg_view_overlay.labels[0]` (as done for the `fps_ol`).

src/odemis/gui/cont/tabs/fastem_project_settings_tab.py
```diff
@@ -229,6 +232,7 @@ def _fill_hfw_choices(self):
         )
         hfw_unit = self.tab_data.main.ebeam.horizontalFoV.unit
         hfw_choices_formatted, choices_si_prefix = format_choices(horizontal_fw_choices)
+        hfw_choices_formatted = [(c, f) for c, f in hfw_choices_formatted if c > STAGE_PRECISION]
```
pieleric, line 235: It looks to me as you could filter the hfw_choices even earlier: just after computing `horizontal_fw_choices`, filter the list with `c > STAGE_PRECISION`. I think it'd be equivalent, but the code would be simpler (as it's just a list of floats, instead of a tuple). It could also lead to a nicer `choices_si_prefix`.

src/odemis/gui/cont/tabs/fastem_project_settings_tab.py
```diff
@@ -255,7 +260,45 @@ def on_evt_hfw_sb_ctrl(self, evt):
         if not ctrl:
             return
 
-        value = ctrl.GetClientData(ctrl.GetSelection())
+        ctrl_value = ctrl.GetValue()
+        i = ctrl.GetSelection()
+        # EVT_COMBOBOX is triggered when the user selects an item from the dropdown
+        if i != wx.NOT_FOUND and ctrl.Items[i] == ctrl_value:
```
pieleric, line 266: Can you add a comment explaining how this `if` conditions are linked to the "EVT_COMBOBOX"? Wouldn't it be simpler to check the event type (using `isinstance`), or even having each event bound to a different callback? 
Or does this `if` actually handle special corner cases which make one event having to be handled by the other event type handler? (eg, the user types in an existing value).

pieleric (changes requested):

src/odemis/gui/cont/tabs/fastem_main_tab.py
nandishjpatel: Nice suggestion, thanks!

src/odemis/gui/cont/tabs/fastem_project_settings_tab.py
nandishjpatel: If I filter them first, then for immersion mode I get all choices in `µm` and for non-immersion mode I get all choices in `mm`. For example `50 µm` hfw in immersion is `0.05 mm` hfw in non-immersion. I had like to keep the same ruler scale, so that for a switch between immersion and non-immersion I can also use FindString to find matching combobox values.

src/odemis/gui/cont/tabs/fastem_project_settings_tab.py
nandishjpatel: I just followed a similar approach to https://github.com/delmic/odemis/blob/c16892410729f7078b3e0403a480852aa20d54b8/src/odemis/gui/conf/util.py#L880

You are right, I can separate the 2 event callbacks.

pieleric (approved):
Looks good to me. Could you squash your last 2 commits into the relevant commits? Then we can merge this branch.

pieleric:
> Looks good to me. Could you squash your last 2 commits into the relevant commits? Then we can merge this branch.

Actually, I didn't realise that neither @tepals nor @tmoerkerken have reviewed this PR yet. So you might want to wait their feedback before squashing commits :-)

src/odemis/gui/cont/stream.py
```diff
@@ -1412,11 +1414,65 @@ def _create_immersion_mode_cbox(self):
...
+        self.stream_panel.num_rows += 1
+        return btn_reference_stage
+
     def _add_fastem_ctrls(self):
         self.stream_panel.add_divider()
         # Create the immersion mode button
         cbox_immersion_mode = self._create_immersion_mode_cbox()
         # Create the buttons
+        _, self.btn_reference_stage = self.stream_panel.add_run_btn(
+            label_text="Reference stage", button_label="Run"
```
tepals, line 1471: does it make sense to mention here that this is only x and y? Maybe as a tooltip?

tmoerkerken:
> > Looks good to me. Could you squash your last 2 commits into the relevant commits? Then we can merge this branch.
> 
> Actually, I didn't realise that neither @tepals nor @tmoerkerken have reviewed this PR yet. So you might want to wait their feedback before squashing commits :-)

I very briefly skimmed over a view commits, which looked good, but I only have time for a proper review after Wednesday.

PR #3098, opened 2025-05-14
[fix] mdupdater should use the filter value as is when there is no monochromator

pieleric:
Commit e4cc6aa51b (handle metadata for monochromator with both
spectrograph and filter) introduced a fancy way to compute the
bandwidth, useful of the SPARC.

However, it broke some basic cases on other systems, such as the METEOR
& SECOM, when the filter is "pass-through", which was then converted to
"None", which is a sort-of correct value, but less specific.

### DIFF
--- a/src/odemis/odemisd/mdupdater.py
+++ b/src/odemis/odemisd/mdupdater.py
@@ -20,8 +20,9 @@ You should have received a copy of the GNU General Public License along with
 Odemis. If not, see http://www.gnu.org/licenses/.
 '''
 import collections
+import itertools
 import logging
-from typing import Optional, Tuple, Dict, Union, Iterable
+from typing import Optional, Tuple, Dict, Union, Iterable, List
 
 from odemis import model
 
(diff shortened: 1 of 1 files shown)


K4rishma (approved):

src/odemis/odemisd/mdupdater.py
```diff
@@ -328,32 +329,55 @@ def getMonochromatorBandwidth(self, spectrograph: Optional[model.HwComponent]) -
             bandwidth = spectrograph.getOpeningToWavelength(width)  # always a tuple
             return bandwidth
 
-    def getFilterBandwidth(self, filter: Optional[model.HwComponent]) -> Union[Iterable[float], str, None]:
+    def get_filter_pos(self, filter: Optional[model.HwComponent]) -> Union[Iterable[float], str, None]:
```
tmoerkerken, line 332: I think the return type is not really matching the description now, should be:
```
Optional[Union[
    Tuple[float, float],
    List[Tuple[float, float]],
    str
]]
```

src/odemis/odemisd/mdupdater.py
```diff
@@ -328,32 +329,55 @@ def getMonochromatorBandwidth(self, spectrograph: Optional[model.HwComponent]) -
...
         """
         if filter is None:
             return None
 
         pos = filter.position.value
-        bandwidth = filter.axes["band"].choices[pos["band"]]
-        if bandwidth == model.BAND_PASS_THROUGH:  # "pass-through" == no filter
+        return filter.axes["band"].choices[pos["band"]]
+
+    def convert_filter_to_bandwidth(self, pos: Union[Iterable[float], str, None]) -> Optional[Tuple[float, float]]:
```
tmoerkerken, line 347: Same typing remark as above

tmoerkerken (changes requested):

src/odemis/odemisd/mdupdater.py
pieleric: Good catch. BTW, `Optional[...]` is the same as `Union[None,...]`, so I only use the `Union`.

tmoerkerken (approved):

PR #3100, opened 2025-05-16
[fix] change the port number for autoscript backend from 4242 to 4243

K4rishma:
Run backend of xtadapter and autoscript on different ports 4242, 4243 respectively.

Tested on the simulator for autoscript on port 4243
- one test is failing (working on the PR) but it is not related to change in the port number
image

### DIFF
--- a/install/linux/usr/share/odemis/sim/meteor-fibsem-sim.odm.yaml
+++ b/install/linux/usr/share/odemis/sim/meteor-fibsem-sim.odm.yaml
@@ -1,6 +1,6 @@
 # Simulates a METEOR on TFS SEM, with control of the FIB (milling)
 # It requires the separate xtadapter-autoscript simulator. To start it, do:
-# python3 path/to/xtlib-adapter/xtadapter/server_autoscript.py sim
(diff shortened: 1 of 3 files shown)


src/odemis/driver/test/xt_client_test.py
```diff
@@ -83,7 +83,7 @@
               }
 
 CONFIG_FIB_SCANNER = {"name": "fib-scanner", "role": "ion", "channel": "ion2"}
-CONFIG_FIB_SEM = {"name": "sem", "role": "sem", "address": "PYRO:Microscope@192.168.31.162:4242",
+CONFIG_FIB_SEM = {"name": "sem", "role": "sem", "address": "PYRO:Microscope@192.168.31.162:4243",
```
pieleric, line 86: That's incorrect. That "FIB_SEM" component uses xtlib to access the FIB image. So it should still be 4242... although on the testing VM, 4242 is used to run the *xttoolkit* simulator, which doesn't have a FIB component, so it's going to fail (it actually should already fail). 
In any case, better do nothing that changing to 4243, in the whole `xt_client_test` file.

pieleric (changes requested):

pieleric (approved):

PR #3104, opened 2025-05-20
[fix] add sample grid overlay for SEM and FM imaging modes in main_gui_data

K4rishma:
Tested on the simulator by adding feature and changing between different postures
Image 1 --> FIBSEM tab 
Image 2 --> Localisation tab 
Currently the sample centres are active only for FM and SEM imaging and for Meteor Posture Manager. 
image
image

### DIFF
--- a/src/odemis/gui/cont/tabs/fibsem_tab.py
+++ b/src/odemis/gui/cont/tabs/fibsem_tab.py
@@ -96,6 +96,11 @@ class FibsemTab(Tab):
         panel.vp_secom_bl.canvas.remove_view_overlay(panel.vp_secom_bl.canvas.play_overlay)
         panel.vp_secom_br.canvas.remove_view_overlay(panel.vp_secom_br.canvas.play_overlay)
 
+        # Add a sample overlay to each view (if we have information)
+        if main_data.get_sample_centers(SEM_IMAGING):
+            for vp in (panel.vp_secom_tl, panel.vp_secom_bl):
+                vp.show_sample_overlay(main_data.get_sample_centers(SEM_IMAGING), main_data.sample_radius)
+
         tab_data.focussedView.subscribe(self._on_view, init=True)
 
         self._view_selector = viewcont.ViewButtonController(
--- a/src/odemis/gui/cont/tabs/localization_tab.py
+++ b/src/odemis/gui/cont/tabs/localization_tab.py
@@ -107,9 +107,9 @@ class LocalizationTab(Tab):
         panel.vp_secom_tr.canvas.remove_view_overlay(panel.vp_secom_tr.canvas.play_overlay)
 
         # Add a sample overlay to each view (if we have information)
-        if main_data.sample_centers:
+        if main_data.get_sample_centers(FM_IMAGING):
             for vp in (panel.vp_secom_tl, panel.vp_secom_tr, panel.vp_secom_bl, panel.vp_secom_br):
-                vp.show_sample_overlay(main_data.sample_centers, main_data.sample_radius)
+                vp.show_sample_overlay(main_data.get_sample_centers(FM_IMAGING), main_data.sample_radius)
 
         # Default view is the Live 1
         tab_data.focussedView.value = panel.vp_secom_bl.view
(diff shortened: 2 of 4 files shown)


src/odemis/gui/cont/tabs/fibsem_tab.py
```diff
@@ -96,6 +96,11 @@ def __init__(self, name, button, panel, main_frame, main_data):
         panel.vp_secom_bl.canvas.remove_view_overlay(panel.vp_secom_bl.canvas.play_overlay)
         panel.vp_secom_br.canvas.remove_view_overlay(panel.vp_secom_br.canvas.play_overlay)
 
+        # Add a sample overlay to each view (if we have information)
```
tmoerkerken, line 99: This indenting shows not aligned to me in the comparison view, but it seems to be a Github bug

src/odemis/gui/cont/tabs/fibsem_tab.py
```diff
@@ -96,6 +96,11 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
+        if main_data.get_sample_centers(SEM_IMAGING):
```
tmoerkerken, line 100: You are repeating the getter, assigning it first and reusing it is more efficient.
```
sem_imaging_sample_centers = main_data.get_sample_centers(SEM_IMAGING)
```

src/odemis/gui/cont/tabs/localization_tab.py
```diff
@@ -107,9 +107,9 @@ def __init__(self, name, button, panel, main_frame, main_data):
         panel.vp_secom_tr.canvas.remove_view_overlay(panel.vp_secom_tr.canvas.play_overlay)
 
         # Add a sample overlay to each view (if we have information)
-        if main_data.sample_centers:
+        if main_data.get_sample_centers(FM_IMAGING):
```
tmoerkerken, line 110: same remark as above

src/odemis/gui/cont/tabs/localization_tab.py
```diff
@@ -107,9 +107,9 @@ def __init__(self, name, button, panel, main_frame, main_data):
...
             for vp in (panel.vp_secom_tl, panel.vp_secom_tr, panel.vp_secom_bl, panel.vp_secom_br):
-                vp.show_sample_overlay(main_data.sample_centers, main_data.sample_radius)
+                vp.show_sample_overlay(main_data.get_sample_centers(FM_IMAGING), main_data.sample_radius)
```
tmoerkerken, line 112: So now we add both the SEM and FM "circles" to the same viewports? If you zoom out, you would see 4 grid circles then? Could that be confusing?

src/odemis/gui/model/main_gui_data.py
```diff
@@ -562,6 +563,35 @@ def __init__(self, microscope):
         # Bounding-box of the "useful area" relative to the center of a grid
         self.sample_rel_bbox = self.SAMPLE_USABLE_BBOX_TEM_GRID
 
+    def get_sample_centers(self, posture_lbl: Optional[int] = None) -> Dict[str, Tuple[float]]:
```
tmoerkerken, line 566: I think technically the `Tuple[float]` typing is incorrect. It should be `Tuple[float, float, float, float, float]`, which looks maybe a little bit ridiculous. We could do something like:
```
FiveDPosition = Tuple[float, float, float, float, float]

def get_position(...) -> FiveDPosition:
```
But it's probably like this all over the code base, so maybe good to refactor it in one go later:
https://delmic.atlassian.net/browse/METEOR-1582

tmoerkerken (changes requested):

src/odemis/gui/cont/tabs/localization_tab.py
K4rishma: The grid ports are transported based on the current posture. Two at a time.

pieleric (changes requested):

K4rishma:
Temporarily, the sample centres are disabled as in https://github.com/delmic/odemis/pull/3123 
This PR of correcting the grid overlay will be taken up later

PR #3107, opened 2025-05-27
[FVEM-125] productization of service scripts

nandishjpatel:


pieleric (changes requested):

plugins/fastem_script_runner.py
```diff
@@ -0,0 +1,768 @@
...
+import threading
+import traceback
+from typing import List, Optional, Tuple
+
+import wx
+
+from odemis.gui.plugin import Plugin
+
+DEV_SCRIPT_BASE_PATH = os.path.expanduser("~/development/fastem-calibrations/")
+RELEASE_SCRIPT_BASE_PATH = "/usr/share/fastem-calibrations"
```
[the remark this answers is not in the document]
nandishjpatel: Right! this did not come to my mind

plugins/fastem_script_runner.py
```diff
@@ -0,0 +1,635 @@
...
+        )
+
+    def emit(self, record):
+        # The `record` object has all the log information
+        # `self.format(record)` will turn it into a string based on the handler's formatter
+        try:
+            msg = self.format(record) + "\n"  # Add a newline like a typical console log
+            self.wx_stream.write(msg)
+        except Exception:
+            self.handleError(
```
K4rishma, line 195: except Exception:
           # Default error handling (prints to original stderr)
            self.handleError(record)

plugins/fastem_script_runner.py
```diff
@@ -0,0 +1,627 @@
...
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+                text=True,
+                bufsize=1,
+                env=script_full_env,
+            )
+            self.is_script_running = True
+            if self.stop_button:
+                self.stop_button.Enable(True)
+            if self.input_ctrl:
```
pieleric, line 327: I don't understand these 2 `if` blocks. Aren't they always True?

plugins/fastem_script_runner.py
```diff
@@ -0,0 +1,627 @@
...
+            self.force_kill_timer = None
+
+        if self.is_script_running and self.process:
+            wx.CallAfter(self._append_output_safe, "\n--- Window closed. Attempting to terminate script process... ---\n")
+            try:
+                self.process.terminate()
+            except ProcessLookupError:  # Process already gone
+                pass
+            except Exception as e:
+                wx.CallAfter(self._append_output_safe, f"\n--- Error terminating script on close: {e} ---\n")
```
pieleric, line 438: Are this `CallAfter()` calls ever going to have an effect? It seems to me that it's show a text on a closed window! Maybe better to use `logging` anyway, in order to have this info in the log files.

plugins/fastem_script_runner.py
```diff
@@ -0,0 +1,627 @@
...
+                current_pythonpath_str = os.environ.get("PYTHONPATH", "")
+                current_pythonpath_list = [
+                    os.path.abspath(os.path.normpath(p))
+                    for p in current_pythonpath_str.split(os.pathsep)
+                    if p
+                ]
+
+                # Prepend the new base_path
+                new_path_list = [normalized_base_path] + current_pythonpath_list
+                self.script_env["PYTHONPATH"] = os.pathsep.join(new_path_list)
```
pieleric, line 556: You don't need to convert from string to list, and back to string.
```suggestion
                current_pythonpath = os.environ.get("PYTHONPATH", "")
                # Prepend the new base_path
                self.script_env["PYTHONPATH"] = normalized_base_path + os.pathsep + current_pythonpath
```

pieleric (changes requested):

plugins/fastem_script_runner.py
```diff
@@ -0,0 +1,610 @@
...
+                script_full_env.update(self.script_env)
+            script_full_env["PYTHONUNBUFFERED"] = "1"  # Force unbuffered output from Python script
+
+            self.process = subprocess.Popen(
+                command,
+                stdin=subprocess.PIPE,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+                text=True,
+                bufsize=1,
```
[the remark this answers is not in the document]
nandishjpatel: Good catch, I previously played around with it indeed to avoid the problem with `input()`, but in any case I still need the linesep. I made the `bufsize=0` for clarity.

tepals (approved):

PR #3118, opened 2025-06-13
[feat] Add "(development)" in the GUI window title when running the development version

pieleric:
Users should normally always run the stable (or release candidate) versions, unless there is a really good reason.
There have been some cases when for debugging purpose Odemis was switch "temporarily" to the development version, and left as-is afterwards.

To highlight that the development version is running (and so some things might be broken), explicitly indicate it in the window title.


tepals (approved):

tepals:
@pieleric would it make sense to also detect and display if they are running proposed?

ilyushkin (approved):

PR #3120, opened 2025-06-16
[fix] enable/disable reference stage button on sem stream panel calibration button clicks

nandishjpatel:


tepals (approved):
It might be nice to create a separate function `enable_calibration_buttons()` which allows you to enable and disable them all at once. I'll leave it up to you if you want to do that now.

pieleric (approved):

PR #3121, opened 2025-06-16
[MSD-23][style] remove u-prefix from string literals

ilyushkin:
Removed legacy u-prefixes from string literals in core application code and across secondary directories.
These prefixes are redundant in Python 3 and removed for code clarity.


src/odemis/gui/conf/file.py
```diff
@@ -161,17 +161,17 @@ def __init__(self):
...
         self.default.add_section("calibration")
-        self.default.set("calibration", "ar_file", u"")
-        self.default.set("calibration", "spec_file", u"")
-        self.default.set("calibration", "spec_bck_file", u"")
-        self.default.set("calibration", "temporalspec_bck_file", u"")
+        self.default.set("calibration", "ar_file", "")
+        self.default.set("calibration", "spec_file", "")
+        self.default.set("calibration", "spec_bck_file", "")
+        self.default.set("calibration", "temporalspec_bck_file", "")
         self.default.set("calibration", "angularspec_bck_file", "u")
```
nandishjpatel, line 175: `self.default.set("calibration", "angularspec_bck_file", "")`

Maybe

src/odemis/gui/model/stream_view.py
```diff
@@ -49,7 +49,7 @@ def __init__(self, name):
         self.lastUpdate = model.FloatVA(time.time(), unit="s")
 
     def __unicode__(self):
```
pieleric, line 51: On Python 3, `__unicode__()` is not used. => remove the entire method.

scripts/sem_bleaching_map.py
```diff
@@ -714,7 +714,7 @@ def save_data(self, data, fn):
         else:
             logging.info("Saving file '%s'", fn)
 
-        exporter.export(unicode(fn), data)
+        exporter.export(str(fn), data)
```
pieleric, line 717: You can pass `fn` directly. It's supposed to be a string already. 
Ideally, you'd also update the docstring, to just indicate `fn` is a `str`.

pieleric (changes requested):

PR #3133, opened 2025-06-24
[METEOR-1596][fix] test_cryo_feature_overlay in comp_overlay_test.py

K4rishma:
Mock backend components used in posture manager which is called by cryo_feature.py

comp_overlay_test.py:501: in test_cryo_feature_overlay
    cryofeature_overlay = CryoFeatureOverlay(cnvs, tab_mod)
../comp/overlay/cryo_feature.py:60: in __init__
    self.view_posture = self.tab_data.view_posture.value
E   AttributeError: 'CryoGUIData' object has no attribute 'view_posture'

### DIFF
--- a/src/odemis/gui/test/comp_overlay_test.py
+++ b/src/odemis/gui/test/comp_overlay_test.py
@@ -25,10 +25,10 @@ import copy
 import logging
 import math
 import re
-import time
 import unittest
 import warnings
 from builtins import range
+from unittest.mock import patch, Mock
 
 import numpy
 import wx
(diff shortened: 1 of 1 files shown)


src/odemis/gui/test/comp_overlay_test.py
```diff
@@ -489,6 +504,43 @@ def test_cryo_feature_overlay(self):
...
+            """
+            Mock function to return a scan rotation value without running the backend.
+            :return: float scan rotation value
+            """
+            return 0.0
+
+        # Mock getMicroscope to return FakeMicroscope
+        with patch('odemis.gui.test.comp_overlay_test.getMicroscope', return_value=FakeMicroscope()), \
+                patch('odemis.acq.move.model.getComponent', side_effect=getComponent), \
+                patch('odemis.acq.move.MeteorPostureManager._get_scan_rotation', return_value=get_scan_rotation()):
```
pieleric, line 539: I'm pretty sure you could simplify it by just replacing the call to `get_scan_rotation()` by just `0.0`.

PR #3135, opened 2025-06-25
[METEOR-1349][feat] andorcam3: better support for Sona and report blemish filter use

pieleric:
Add . frameRate VA to limit the frameRate, if necessary. One of such case is for sustaining a live stream at full res on the Sona, which can do 100Hz for a brief period, but only supports 40Hz for longer than 1 s.

Also improve a little bit recovery of the live stream when such timeout happen. It now takes only 1.5s to recover, instead of ~4s.

Also make the readout rate read-only on the Sona, as it is fixed, dependent on the current gain.

Finally, indicate in the metadata when the blemish filter is in use (which is activated by default on cameras supporting it).

### DIFF
--- a/src/odemis/driver/andorcam3.py
+++ b/src/odemis/driver/andorcam3.py
@@ -416,9 +416,18 @@ class AndorCam3(model.DigitalCamera):
         # Another advantage of global shutter allows to simplify if we need to
         # connect exposure to light.
 
+        filters = []
         if self.isImplemented("SpuriousNoiseFilter"):
             self.SetBool("SpuriousNoiseFilter", True)
-            self._metadata['Filter'] = "Spurious noise filter" # FIXME tag?
+            filters.append("Spurious noise")
+
+        if self.isImplemented("StaticBlemishCorrection"):
+            # It is activated by default on the Sona and Zyla, but just in case, explicitly set it.
+            self.SetBool("StaticBlemishCorrection", True)
+            filters.append("Blemish correction")
+
+        if filters:
+            self._metadata[model.MD_DATA_FILTER] = ", ".join(filters)
 
         if self.isImplemented("Overlap"):
             # No overlap, as it can introduce additional noise, and we don't need
--- a/src/odemis/driver/andorcam3.py
+++ b/src/odemis/driver/andorcam3.py
@@ -4,7 +4,7 @@ Created on 6 Mar 2012
 
 @author: Éric Piel
 
-Copyright © 2012-2016 Éric Piel, Delmic
+Copyright © 2012-2025 Éric Piel, Delmic
 
 This file is part of Odemis.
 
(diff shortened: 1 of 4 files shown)


src/odemis/driver/test/andorcam3_test.py
```diff
@@ -64,6 +66,13 @@ class TestAndorCam3(VirtualTestCam, unittest.TestCase):
     camera_type = CLASS
     camera_kwargs = KWARGS
 
+    def test_gain(self):
+        for g in self.camera.gain.choices:
```
K4rishma, line 70: A comment will help to understand what is the code checking

src/odemis/driver/test/cam_test_abs.py
```diff
@@ -322,6 +322,43 @@ def test_acquire_flow_no_drop(self):
...
+        start_t = time.time()
+        for i in range(number):
+            # end early if it's already finished
+            if self.left == 0:
+                break
+            time.sleep(0.01 + 1 / fr)  # Should be about the frame rate
+        stop_t = time.time()
+
+        dur = stop_t - start_t
+        self.assertGreaterEqual(dur, exp_dur)  # acquisition went too fast?
```
K4rishma, line 359: The acquisition is expected to go fast, should the question mark be there?

src/odemis/driver/test/cam_test_abs.py
```diff
@@ -322,6 +322,43 @@ def test_acquire_flow_no_drop(self):
...
+        self.assertEqual(self.left, 0)  # acquisition went too slow?
```
K4rishma, line 360: same comment as before

K4rishma (approved):

src/odemis/driver/andorcam3.py
```diff
@@ -315,15 +320,17 @@ def __init__(self, name, role, device=None, bitflow_install_dirs=None,
...
+            supported_fan_speeds = self.GetEnumStringImplemented("FanSpeed")
+            if len(supported_fan_speeds) > 1:
+                logging.info("Supported fan speeds: %s", supported_fan_speeds)
+                # Start at max speed
+                self.fanSpeed = model.FloatContinuous(1.0, (0.0, 1.0), unit="",
+                                                      setter=self.setFanSpeed) # ratio to max speed
+                self.setFanSpeed(1.0)
+
+        # binning, res, trans, exp time, readout rate, frame rate, gain, synchronised
+        self._prev_settings = [None, None, None, None, None, None, None, None]
```
nandishjpatel, line 333:
```suggestion
        self._prev_settings = [None] * 8
```

src/odemis/driver/andorcam3.py
```diff
@@ -943,47 +973,78 @@ def setFanSpeed(self, speed):
...
+        # ["Off", "Low", "On"]
         values = self.GetEnumStringImplemented("FanSpeed")
         if len(values) <= 1:
             return 0
         speed_index = int(round(speed * (len(values) - 1)))
         self.SetEnumString("FanSpeed", values[speed_index])
         return speed_index / (len(values) - 1)
 
-    def getReadoutRates(self):
+    def getReadoutRates(self, all_gains=True) -> List[Optional[float]]:
```
nandishjpatel, line 984:
```suggestion
    def getReadoutRates(self, all_gains: bool = True) -> List[Optional[float]]:
```

nandishjpatel (approved):

PR #3136, opened 2025-06-25
[perf] generate_zlevels in a more intuitive way

K4rishma:
This a new way of calculating the z levels

**New way** - the current focus level would be one of the z levels and ztep defined by the user would strictly dictate the difference between the consecutive calculated z levels

**Old way** - the above two were not strictly followed i.e. the zsteps could slightly vary and the current focus value might not be found in the calculated z levels.

Updated the function and test cases

Tested the working of modified implementation on
- comp_test.TestGenerateZlevels
- feature_acq_test.py
- acq_test.py

### DIFF
--- a/src/odemis/util/comp.py
+++ b/src/odemis/util/comp.py
@@ -118,8 +118,9 @@ MAX_ZLEVELS = 500
 
 def generate_zlevels(focuser: "Actuator", zrange: Tuple[float, float], zstep: float) -> List[float]:
     """
-    Calculates the zlevels for a zstack acquisition, using the zmax, zmin
-    and zstep, as well as the current focus z position.
+    Calculates the zlevels for a zstack acquisition such that zstep is the distance between two successive zlevels and
+    the zlevels are within the valid z range. The zlevels are based on the current focus position. In a rare case when
+    the when the zstep is larger than the z range, the z range itself is returned as a list of two elements.
     :param focuser: Actuator component with a "z" axis, to control the focus.
     :param zrange: contains the zmin and zmax, respectively. It's relative to the
     current position of the focuser. If the range would go out of the actuator
@@ -132,6 +133,11 @@ def generate_zlevels(focuser: "Actuator", zrange: Tuple[float, float], zstep: fl
     In that case IndexError is raised.
     :returns: list of zlevels, where each zlevel is absolute position for
     the focuser.
+    :raises ZeroDivisionError: if zstep is zero.
+    :raises ValueError: if the zrange is not correct, i.e. the first value is
+    larger than the second one.
+    :raises KeyError: if the focuser does not have a "z" axis.
+    :raises IndexError: if the number of zlevels exceeds MAX_ZLEVELS.
     """
     if zstep == 0:
         raise ZeroDivisionError("The step size 'zstep' can not be zero")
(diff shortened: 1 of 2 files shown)


src/odemis/util/test/comp_test.py
```diff
@@ -188,21 +188,40 @@ def test_normal_zlevels_output_with_rounding_down(self):
...
         zMax = 24e-6
         zMin = -24e-6
         zrange = [zMin, zMax]
         zStep = 17e-6
-        expected = numpy.asarray([-24e-6, -8e-6, 8e-6, 24e-6]) + \
-            self.focus.position.value["z"]
+        # the focus value should be one of the z levels and the difference in the levels should be z step
         actual = generate_zlevels(self.focus, zrange, zStep)
-        numpy.testing.assert_array_almost_equal(expected, actual)
+        # Use assertAlmostEqual or numpy's isclose for float comparison
```
pieleric, line 213: You can move these 2 generic checks to `test_large_number_of_levels()`.

pieleric (changes requested):
Please explain more what was the issue and how this code fixes it, in the commit message.

I don't see how the previous implementation didn't "have the zlevels around the focus". The main difference between your new code and the old one is that you enforce that the current focus position is one of the zlevels. I do agree it's a good idea... but that's a different change.

src/odemis/util/comp.py
```diff
@@ -162,22 +162,16 @@ def generate_zlevels(focuser: "Actuator", zrange: Tuple[float, float], zstep: fl
...
-        ef = abs((zrange_abs[1] - zrange_abs[0]) / (n_f - 1) - abs(zstep))
-        ec = abs((zrange_abs[1] - zrange_abs[0]) / (n_c - 1) - abs(zstep))
-        n = n_f if ef < ec else n_c
-
-    if zstep > 0:
-        return numpy.linspace(zrange_abs[0], zrange_abs[1], n).tolist()
-    elif zstep < 0:
-        return numpy.linspace(zrange_abs[1], zrange_abs[0], n).tolist()
+    if n_f <= 1:
+        return [zrange_abs[0], zrange_abs[1]]
```
tepals, line 169: In this case it's not based on the current focus position, right?

src/odemis/util/comp.py
```diff
@@ -162,22 +162,16 @@ def generate_zlevels(focuser: "Actuator", zrange: Tuple[float, float], zstep: fl
...
+
+    abs_zstep = abs(zstep)
+    focus_array_min = numpy.arange(focuser_pos, zrange_abs[0] - abs_zstep/10, -abs_zstep)
```
tepals, line 172: can you add a comment to explain these steps?

tepals, line 172: Note that there is a warning about using arange with floats in the numpy documentation, it's recommended to use linspace instead. Linspace also allows for a `endpoint=True` argument, making it unnecessary to do the `-abs_zstep/10`
https://numpy.org/doc/stable/reference/generated/numpy.arange.html

image

src/odemis/util/comp.py
```diff
@@ -162,22 +162,16 @@ def generate_zlevels(focuser: "Actuator", zrange: Tuple[float, float], zstep: fl
...
+    focus_array_max = numpy.arange(focuser_pos, zrange_abs[1] + abs_zstep/10, abs_zstep)
+    zlevels_list = numpy.unique(numpy.append(focus_array_min, focus_array_max)).tolist()
```
tepals, line 174: Why do you need `numpy.unique` here? If it is just for the focuser_pos, you could also write it something like this:
```suggestion
    zlevels_list = numpy.append(focus_array_min[1:][::-1], focus_array_max).tolist()
```
you will just have to make sure the focus_array_min is sorted

src/odemis/util/comp.py
tepals: I would suggest to write it like (note that the +1 for the number of steps is because you're including the endpoint):
```suggestion
    focus_array_min = numpy.linspace(zrange_abs[0], focuser_pos, (zrange_abs[0] + focuser_pos) / abs_zstep)
    focus_array_max = numpy.linspace(focuser_pos, zrange_abs[1], (zrange_abs[1] + focuser_pos) / abs_zstep + 1, endpoint=True)
```

pieleric: Note that arange and linspace don't give the same results. The previous implementation used linspace, but the problem is that it didn't respect the zstep. See such example:
```
$ numpy.arange(0, 10., 3)
 array([0., 3., 6., 9.])

$ numpy.linspace(0, 10, 4)
array([ 0.        ,  3.33333333,  6.66666667, 10.        ])
```
As I understand it, Karishma did handle the "floating point" issue by adding 10% of zstep . Of course, it should also be possible to write with linspace, but it'd be less explicit. I'd say the only thing missing here is to explain the trick of using `zstep/10`.

src/odemis/util/comp.py
pieleric: I don't think there is a lot better to do. Firstly, that's an extremely unlikely corner case (either the focuser can barely move, which is very odd, or the user asks for gigantic zstep, which is odd).
The alternatives would be to return just the current pos (quite sad for a zstack), or to return the extrema + current pos (3 points not at equal distance from each other, which is odd). Thera, would you suggest something else?

pieleric (changes requested):
Looks pretty fine to me. I'd just recommand explaining the use of arange, and pick the suggestion of Thera to avoid using `numpy.unique`.

src/odemis/util/comp.py
tepals: No, just that now the docstring mentions "The zlevels are based on the current focus position.", I would suggest to update that to mention that in unlikely cases for very large z-step or very small rang it returns the min/max range as z-levels.

PR #3152, opened 2025-07-09
[feat] add forcemd arg to StaticSEMStream to pass 'merge_ratio' as metadata to control the BitmapCanvas drawing

nandishjpatel:


src/odemis/acq/stream/_static.py
```diff
@@ -1053,6 +1053,11 @@ def update(self, raw):
...
-    # blending option ("blend screen" on non-overlapping positions = simple pasting without blending)
-    pass
+
+    def _find_metadata(self, md):
+        useful_md = super()._find_metadata(md)
+        if 'merge_ratio' in md:
+            useful_md['merge_ratio'] = md['merge_ratio']
+        if 'blend_mode' in md:
+            useful_md['blend_mode'] = md['blend_mode']
+        return useful_md
```
nandishjpatel, line 1063: There is already a mechanism in LiveStream that could be used for here too: forcemd. It allows to apply some extra metadata on top of the metadata in the raw data. Please use this mechanism here too. Either just in this class, or in StaticStream, as it could be useful in other cases too.

src/odemis/gui/comp/canvas.py
```diff
@@ -873,7 +873,11 @@ def set_images(self, im_args):
                 md['dc_shear'] = shear
                 md['dc_flip'] = flip
                 md['dc_keepalpha'] = keepalpha
-                md['blend_mode'] = blend_mode
+                # Stream._find_metadata discards 'blend_mode' from the metadata as it is considered not useful
+                # If 'blend_mode' is present in the metadata, it has been set by the user, in that case no need
+                # to update it with the value from DblMicroscopeCanvas._get_ordered_images logic
```
nandishjpatel, line 878: That comment is confusing. You've just added code that makes find_metadata not discard blend_mode. Also the "user" here probably doesn't refer to the microscope/app user, but to the calling code, right? Please adjust this comment to make it clearer.

src/odemis/gui/cont/acquisition/fastem_acq.py
```diff
@@ -314,6 +315,12 @@ def _on_overview_img_load_ctrl(self, _):
             da = open_acquisition(fn)
             s = data_to_static_streams(da)[0]
             s = FastEMOverviewStream(s.name.value, s.raw[0])
+            data = s.raw[0]
+            # Set 'blend_mode' and 'merge_ratio' metadata to be used by BitmapCanvas draw
+            # By setting the metadata override the default value set by BitmapCanvas.set_images
+            # and BitmapCanvas._draw_merged_images logic
+            data.metadata["blend_mode"] = BLEND_DEFAULT
```
nandishjpatel, line 322: Until now the almost only use of FastEMOverviewStream was to force the blend_mode to BLEND_SCREEN, because standard StaticSEMStream use BLEND_DEFAULT. So now, we have code that detects the FastEMOverviewStream, override blend mode to BLEND_SCREEN, and then extra code that override this override...

Instead, I'd suggest the following:

    Remove the special class FastEMOverviewStream
    Everywhere in the GUI where it uses FastEMOverviewStream, use StaticSEMStream. That should be also enough to distinguish the LiveSEMStream from the overview streams.
    Use forcemd to either override the blend_mode to BLEND_SCREEN (on the full overviews), or override the merge_ratio (to 1).

Actually, the BLEND_SCREEN was a trick to make sure the multiple overviews would display all without transparency when multiple ones are shown simultaneously. Is that even still needed? Are there still multiple large overviews shown per canvas now? And even if so, wouldn't they be shown properly too by using BLEND_DEFAULT + merge_ratio = 1?

nandishjpatel: @pieleric 
> Actually, the BLEND_SCREEN was a trick to make sure the multiple overviews would display all without transparency when multiple ones are shown simultaneously. Is that even still needed?

@tepals this is no longer needed right?

> Are there still multiple large overviews shown per canvas now?

Each scintillator has its own viewport and canvas, so there is only going to be 1 large overview image. There will be smaller detailed overview images but they can be shown / hidden using the toggle button

> And even if so, wouldn't they be shown properly too by using BLEND_DEFAULT + merge_ratio = 1?

Yes, I am looking for BLEND_DEFAULT + merge_ratio = 1. with your suggestions of using StaticSEMStream, I just have to pass `forcemd={"merge_ratio": 1}`

tepals: indeed that is no longer needed

src/odemis/gui/comp/canvas.py
```diff
@@ -983,7 +983,9 @@ def _draw_merged_images(self, ctx, interpolate_data=False):
                     #     1         0      1
                     # TODO: for now, this only works correctly if the background
                     # is black (otherwise, the background is also mixed in)
-                    if bm_last == BLEND_SCREEN:
+                    if 'merge_ratio' in md:
+                        merge_ratio = md['merge_ratio']
```
pieleric, line 987: It is worthy a comment, because no code in miccanvas or dblmiccanvas sets `'merge_ratio'`. So it can look a little bit mysterious on when this is used. So please explain this is some sort of hack to override the general `.merge_ratio` per image.

src/odemis/gui/comp/canvas.py
```diff
@@ -983,7 +983,9 @@ def _draw_merged_images(self, ctx, interpolate_data=False):
                     #     1         0      1
                     # TODO: for now, this only works correctly if the background
                     # is black (otherwise, the background is also mixed in)
-                    if bm_last == BLEND_SCREEN:
+                    if 'merge_ratio' in md:
```
pieleric, line 986: There is another function that implement the same logic: `gui.util.img.images_to_export_data()`. This is used when calling "Export". Admittedly, it's not used on the FASTEM, so not a big deal, but please update that function as well, so that it keeps doing the same thing (and we don't get surprise later).

pieleric (changes requested):
The approach looks good. Just small remarks.

src/odemis/gui/comp/canvas.py
nandishjpatel: Didn't know this function. Updated!

tepals (approved):

PR #3157, opened 2025-07-10
[FVEM-143] autofocus add argument to select a focus metric

tepals:
[feat] autofocus add argument to select a focus metric

Add an input argument that allows to select which focus metric to use. If None the focus metric used is based on the detector.

[fix] Add typing to autofocus module and PEP8 clean up

    - removed unused imports
    - ordered imports
    - enforce max line length

### DIFF
--- a/src/odemis/acq/align/autofocus.py
+++ b/src/odemis/acq/align/autofocus.py
@@ -26,7 +26,7 @@ import time
 from collections.abc import Iterable
 from concurrent.futures import TimeoutError, CancelledError
 from concurrent.futures._base import CANCELLED, FINISHED, RUNNING
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Optional, Tuple, Union, Callable
 
 import numpy
 
@@ -122,7 +122,8 @@ def _DoBinaryFocus(
         focus: model.Actuator,
         dfbkg: Optional[model.DataFlow],
         good_focus: Optional[float],
-        rng_focus: Optional[Tuple[float, float]]
+        rng_focus: Optional[Tuple[float, float]],
+        measure_func: Optional[Callable] = None
 ) -> Tuple[float, float, float]:
     """
     Iteratively acquires an optical image, measures its focus level and adjusts
@@ -136,6 +137,9 @@ def _DoBinaryFocus(
       taken into consideration while autofocusing
     rng_focus: if provided, the search of the best focus position is limited
       within this range
+    measure_func: function to measure the focus level on the image,
+      for instance MeasureSEMFocus. If None the focus metric used is based
+      on the detector.
     returns:
         (float): Focus position (m)
         (float): Focus level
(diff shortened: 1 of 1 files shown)


src/odemis/acq/align/autofocus.py
```diff
@@ -1003,19 +1092,25 @@ def AutoFocus(detector, emt, focus, dfbkg=None, good_focus=None, rng_focus=None,
...
 
 
-def AutoFocusSpectrometer(spectrograph, focuser, detectors, selector=None, streams=None):
+def AutoFocusSpectrometer(
+        spectrograph: model.Actuator,
+        focuser: model.Actuator,
+        detectors: Union[model.Detector, List[model.Detector]],
+        selector: Optional[model.Actuator] = None,
+        streams: Optional[List[Any]] = None
+) -> Dict[model.ProgressiveFuture, dict]:
```
tepals, line 1101: @pieleric I'm not sure if this is correct, this was the best I could come up with based on the docstring

pieleric: `streams` is a list of acq.stream.Stream`. Not sure how easy is it to add a dependency to the streams though. Maybe it's simpler to just write `"Stream"`.

For the return value, it's just a `model.ProgressiveFuture` . The `->` in the docstring meant that this Future will return a value.

src/odemis/acq/align/autofocus.py
```diff
@@ -526,12 +549,14 @@ def _CancelAutoFocus(future):
     return True
 
 
-def estimateAcquisitionTime(detector, scanner=None):
+def estimateAcquisitionTime(
+        detector: Union[model.DigitalCamera, model.Detector],
```
pieleric, line 570: DigitalCamera is a subtype of Detector. So you could replace this Union by just `model.Detector`.

src/odemis/acq/align/autofocus.py
```diff
@@ -201,19 +216,23 @@ def _DoBinaryFocus(future, detector, emt, focus, dfbkg, good_focus, rng_focus):
...
-                logging.debug("Using Spot method to estimate focus")
-                Measure = MeasureSpotsFocus
-            elif detector.resolution.value[1] == 1:
-                logging.debug("Using 1d method to estimate focus")
-                Measure = Measure1d
+        if measure_func is None:
+            if len(detector.shape) > 1:
+                if detector.role == 'diagnostic-ccd':
+                    logging.debug("Using Spot method to estimate focus")
+                    Measure = MeasureSpotsFocus
```
tmoerkerken, line 223: What's up with the capitalization?

src/odemis/acq/align/autofocus.py
```diff
@@ -417,19 +449,23 @@ def _DoExhaustiveFocus(future, detector, emt, focus, dfbkg, good_focus, rng_focu
...
         # TODO: is this working as expected? Alternatively, we could check
         # MD_DET_TYPE.
-        if len(detector.shape) > 1:
-            if detector.role == 'diagnostic-ccd':
-                logging.debug("Using Spot method to estimate focus")
-                Measure = MeasureSpotsFocus
-            elif detector.resolution.value[1] == 1:
-                logging.debug("Using 1d method to estimate focus")
-                Measure = Measure1d
+        if measure_func is None:
```
tmoerkerken, line 452: It looks like this block was already defined in `_DoBinaryFocus`. Does it make sense to make a reusable utility method for it?

src/odemis/acq/align/autofocus.py
```diff
@@ -201,19 +216,23 @@ def _DoBinaryFocus(future, detector, emt, focus, dfbkg, good_focus, rng_focus):
...
+                    Measure = MeasureSpotsFocus
+                elif detector.resolution.value[1] == 1:
+                    logging.debug("Using 1d method to estimate focus")
+                    Measure = Measure1d
+                else:
+                    logging.debug("Using Spot method to estimate focus")
+                    Measure = MeasureSpotsFocus
             else:
-                logging.debug("Using Spot method to estimate focus")
-                Measure = MeasureSpotsFocus
```
tepals, line 213: this change should be reverted

src/odemis/acq/align/autofocus.py
```diff
@@ -417,19 +449,23 @@ def _DoExhaustiveFocus(future, detector, emt, focus, dfbkg, good_focus, rng_focu
...
```
tepals, line 429: this change should be reverted

src/odemis/acq/align/autofocus.py
tepals: It was already there, so I decided to leave it like this. I will take a deeper dive into autofocusing towards the end of the year and I'll further clean up this module then.
There is also a lot of duplicated code, so as an answer to your question below: I would like to handle that all at the same point, but if you think I should make a method for the changes in this PR now, I'm also fine with it

pieleric (approved):

PR #3166, opened 2025-07-18
[doc] Minor updates & fixes about setting a development environment on Linux

pieleric:
Now using Ubuntu 22.04.
The odemis.conf syntax is a little more specific now ($HOME is not
supported).
Drop Eclipse description, it seems no one uses it anymore.
Explain briefly how to start & stop Odemis from the command line
(although there is a dedicated section about this later, a lot of people
didn't read up to there)
Also update a couple of URLs.

### DIFF
--- a/doc/develop/extending.rst
+++ b/doc/develop/extending.rst
@@ -27,7 +27,7 @@ Odemis is written almost entirely in Python language. So in theory, a simple
 text editor could be enough to edit Odemis. However in order to execute, debug,
 test, and edit efficiently Odemis, we recommend the following environment:
 
- * `Ubuntu <http://www.ubuntu.com>`_ 20.04 (x86 64 bits)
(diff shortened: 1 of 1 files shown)


tepals (approved):

doc/develop/extending.rst
```diff
@@ -355,8 +342,8 @@ with the python version (3.6.13) on which Odemis has been well tested.
     conda activate odemis
     conda config --append channels conda-forge
     conda install --name odemis --file requirements-conda.txt
```
tmoerkerken, line 344: This requirements file does not exist anymore. One should use `conda env create -f environment.yml -n odemis`, which can replace the first step and the line in this step

tmoerkerken (changes requested):

tmoerkerken (approved):

PR #3170, opened 2025-07-25
[refactor] Change use of depecated ENZEL sim to METEOR sim

pieleric:
Several test cases used the ENZEL simulator. However the ENZEL is not
supported anymore. So adjust these test cases to use a METEOR simulator
instead.

### DIFF
--- a/src/odemis/acq/stitching/test/tiledacq_test.py
+++ b/src/odemis/acq/stitching/test/tiledacq_test.py
@@ -34,6 +34,7 @@ import odemis
 from odemis import model
 from odemis.acq import acqmng, stream
 from odemis.acq.acqmng import SettingsObserver
+from odemis.acq.move import MicroscopePostureManager, FM_IMAGING
 from odemis.acq.stitching import (
     REGISTER_IDENTITY,
     WEAVER_COLLAGE_REVERSE,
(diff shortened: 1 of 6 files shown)


K4rishma:
If we can add stigmator related components to fibsem-sim, it will also be useful in testing the simulator. As multipoint correlation is incomplete without super z.

pieleric:
> Just noticed that the line 69 in the microscope file shows milling button in the gui which can be commented

Removed.

src/odemis/acq/stitching/test/tiledacq_test.py
```diff
@@ -55,16 +56,17 @@
...
-ENZEL_CONFIG = CONFIG_PATH + "sim/enzel-sim.odm.yaml"
 METEOR_CONFIG = CONFIG_PATH + "sim/meteor-sim.odm.yaml"
+METEOR_FIBSEM_CONFIG = CONFIG_PATH + "sim/meteor-fibsem-sim.odm.yaml"
 
 
 class CRYOSECOMTestCase(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
-        testing.start_backend(ENZEL_CONFIG)
+        testing.start_backend(METEOR_FIBSEM_CONFIG)
```
K4rishma, line 69: Maybe a thought for another PR but not this one. Should the class be called SECOMTestCase?

K4rishma (approved):

src/odemis/acq/stitching/test/tiledacq_test.py
pieleric: Cryo SECOM is the old name for ENZEL. I could rename it "METEORTestCase"... but there is already such test case. So we need to find a better name. I gave up for now, but if you have a good suggestion (for the 2 test cases), I'm happy to adjust the names :-)

K4rishma: Maybe CryoAcqTestCase , just one name I can think

tmoerkerken (approved):

PR #3175, opened 2025-07-31
Csdm 1148 create a plugin script to acquire full dataset for neural network training

nandishjpatel:
https://github.com/user-attachments/assets/c0111e17-f288-4383-9c74-cda9872f39a5


pieleric (changes requested):

plugins/sparc2_alignment_data_collector.py
```diff
@@ -0,0 +1,621 @@
...
+                                "s": self.mirror.position.value["s"],
+                                "x": self.mirror_xy.position.value["x"],
+                                "y": self.mirror_xy.position.value["y"],
+                                "z": self.stage.position.value["z"],
+                            }
+                            wx.CallAfter(self.status_lbl.SetLabel, "Status: Resumed")
+                            logging.debug(
+                                "Resumed. New aligned position: %s", self._aligned_pos
+                            )
+                        self.stage.moveAbs({"z": z}).result()
```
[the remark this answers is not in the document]
nandishjpatel: I am not sure about this, will ask Herman

plugins/sparc2_alignment_data_collector.py
```diff
@@ -0,0 +1,621 @@
...
+                        "s": self.mirror.position.value["s"],
+                        "x": self.mirror_xy.position.value["x"],
+                        "y": self.mirror_xy.position.value["y"],
+                        "z": self.stage.position.value["z"],
+                    }
+                    wx.CallAfter(self.status_lbl.SetLabel, "Status: Resumed")
+                    logging.debug(
+                        "Resumed. New aligned position: %s", self._aligned_pos
+                    )
+                self.mirror.moveAbs({"l": l}).result()
```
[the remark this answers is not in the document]
nandishjpatel: For now I wanna leave it as it is

plugins/sparc2_alignment_data_collector.py
```diff
@@ -0,0 +1,699 @@
...
+                        }
+                        logging.debug(f"Acquiring l={l} s={s} z={z}")
+                        filepath = os.path.join(
+                            path, f"{idx}_snapshot_{l}_{s}_{z}" + hdf5.EXTENSIONS[0]
+                        )
+                        self.tab._ccd_spe.pauseStream()
+                        self.tab._ccd_spe.pause()
+                        data = self.tab._ccd_stream.getSingleFrame()
+                        data.metadata.update(forcemd)
+                        self.tab._ccd_stream._onNewData(self.tab._ccd_stream._dataflow, data)
```
pieleric, line 629: This should work. 
That said, it'd probably more correct to use `single_frame_acquisition.value = True` and use `is_active.value = True` to trigger one frame acquisition. That's the "standard" way to use it.

plugins/sparc2_alignment_data_collector.py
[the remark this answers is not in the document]
pieleric: I'm pretty sure about it. Typically, in the workflow the user will start a little far in Z, focus the e-beam there, and then move the stage up. So we want to have a nice focused e-beam for the image corresponding to a Z too low. Also, maybe the user doesn't adjust the e-beam focus at every Z stage move, but if so, that's probably just because it's annoying to do, not that it's not useful... Let's double check with Herman !

plugins/sparc2_alignment_data_collector.py
```diff
@@ -0,0 +1,699 @@
...
+                self.mirror.moveAbs({"l": l}).result()
+                for s in s_values:
+                    if self.cancel_requested:
+                        raise CancelledError("Cancelled by user.")
+                    self.mirror.moveAbs({"s": s}).result()
+                    for z in z_values:
+                        if self.cancel_requested:
+                            raise CancelledError("Cancelled by user.")
+                        if self._pause_requested:
+                            wx.CallAfter(self.status_lbl.SetLabel, "Status: Paused")
```
pieleric, line 579: Shouldn't it first move to the previously known aligned position? This way, if you do pause/resume, you don't automatically mess up the acquisition by having a wrong "new" aligned position.
Also, you might want to make it clearer to the user that after resuming the position has to be *aligned*. Maybe set the status to "Status: paused on aligned position"?

pieleric (changes requested):

plugins/sparc2_alignment_data_collector.py
nandishjpatel: Right, will change the status to "Status: paused on aligned position"

The call in line 580 `wx.CallAfter(self._restore_position)` does move it to the previously know aligned position and then waits for resume in line 584 `self._resume_requested.wait()`

plugins/sparc2_alignment_data_collector.py
```diff
@@ -0,0 +1,726 @@
...
+        try:
+            idx = 0
+            # Pause the ccd stream
+            self.tab._ccd_stream.is_active.value = False
+            self.tab._ccd_stream.should_update.value = False
+            # Pausing the ccd stream causes the e-beam to be blanked and mode as internal
+            # make the mode as external and unblank the beam to get valid ccd data
+            self.ebeam.external.value = True
+            self.ebeam.blanker.value = False
+            for l in l_values:
```
nandishjpatel, line 600: @pieleric , @tepals  there is also a possibility to make this loop "Starts at the aligned center and moves outwards in expanding shells". This can help solve the drift issue over time in `l and s` and acquire good initial data at the aligned position first. Maybe this approach has its own issue and can take more time.

tepals: Sounds like a good approach! It's probably a good idea to have at least one iteration of the prediction, so having more accurate data closer to alignment will improve the final iteration.
And +1 for the 3D hilbert curve :grinning:

nandishjpatel:
@tepals can you please take a second look at the PR from a pure software perspective after the algorithm change (second commit)? Last Friday, Herman and me could not acquire good data using this plugin. This new algorithm start acquiring from a position closest to the aligned position and moving from inner shells to outward shells. The issue, when the acquisition starts this first move which is close to the aligned position results in the spot shape changing a lot. In the meantime I have some acquired files to debug it.

plugins/sparc2_alignment_data_collector.py
```diff
@@ -0,0 +1,891 @@
...
+        try:
+            progress_idx = 0
+
+            # Generate the optimized traversal path
+            wx.CallAfter(self.status_lbl.SetLabel, "Status: Calculating optimal acquisition path...")
+            acquisition_path, shell_end_indices = self._generate_acquisition_path(l_values, s_values, z_values)
+            self.n_images = len(acquisition_path) + len(shell_end_indices)
+            self.progress.SetRange(self.n_images)
+            wx.CallAfter(self.status_lbl.SetLabel, "Status: Path calculated. Starting acquisition...")
+
```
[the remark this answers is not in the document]
pieleric: @nandishjpatel , that's indeed needed

src/odemis/gui/cont/acquisition/snapshot.py
```diff
@@ -133,14 +134,16 @@ def _get_snapshot_info(self, dialog=False):
 
         return tab, filepath, exporter
 
-    def snapshot_viewport(self, tab, filepath, exporter, anim):
+    def snapshot_viewport(self, tab, filepath, exporter, anim, forcemd: Optional[dict] = None):
```
pieleric: Please adjust the plugin such that it doesn't need to add this option to the standard GUI code. This option is very unlikely going to be used by the standard code. In addition, it should be possible to find other way to do the same behaviour directly in the plugin. I think the best would be to let the plugin directly export the .raw data. Alternatively, it could update the .metadata by itself. No need to do it here.

plugins/sparc2_alignment_data_collector.py
```diff
@@ -0,0 +1,951 @@
...
+        }
+        if l_target is not None:
+            extra_settings_md.update({"l_target": l_target, "dl_target": l_target - l_current})
+        if s_target is not None:
+            extra_settings_md.update({"s_target": s_target, "ds_target": s_target - s_current})
+        if z_target is not None:
+            extra_settings_md.update({"z_target": z_target, "dz_target": z_target - z_current})
+        self.snapshot_controller.snapshot_viewport(
+            self.tab, filepath, self._exporter, anim=False, forcemd={model.MD_EXTRA_SETTINGS: extra_settings_md}
+        )
```
pieleric: Please avoid using the high level `snapshot_viewport` to save an image. The code below should work as well. Assuming the stream is playing, you don't have to call `_onNewData()`, and `raw_data = self.tab._ccd_stream.raw` should return the lastest acquire data. (then, you can update its metadata, as you currently do in `snapshot_viewport`)

PR #3182, opened 2025-08-11
[fix] GUI METEOR: handle streams with an empty name

pieleric:
Odemis tries to provide a list of the last streams created previously.
However, if one stream was created with an empty name (""), that empty
label would end up confusing wx.MenuItem(), showing such error:
  File "/home/piel/development/odemis/src/odemis/gui/cont/stream_bar.py", line 128, in __init__
    self._createAddStreamActions()
  File "/home/piel/development/odemis/src/odemis/gui/cont/stream_bar.py", line 1289, in _createAddStreamActions
    self.add_action(label, call_label, fluor_capable)
  File "/home/piel/development/odemis/src/odemis/gui/cont/stream_bar.py", line 192, in add_action
    self._stream_bar.btn_add_stream.add_choice(title, callback, check_enabled)
  File "/home/piel/development/odemis/src/odemis/gui/comp/buttons.py", line 1057, in add_choice
    menu_item = wx.MenuItem(self.menu, menu_id, label)
wx._core.wxAssertionError: C++ assertion ""Assert failure"" failed at /tmp/pip-install-eoie9qdr/wxpython_1b4ae2e90f734464912c8577e97b58ec/ext/wxWidgets/src/common/stockitem.cpp(219) in wxGet
StockLabel(): invalid stock item ID

=> Detect such case and replace by a generic name.

Also detect that PopupImageButton.add_choice() was called with an empty
label, and raise a less confusing error. That will save some time next
time.

## Summary by CodeRabbit

- Bug Fixes
  - Prevents creation of empty or duplicate menu choices; adding a choice without a label now raises an error. Menu items are enabled/disabled when shown, action handlers stop after a single match, and the control is disabled when no choices remain.

- New Features
  - Stream action labels are auto-filled with a readable fallback when missing and made unique for display while preserving underlying stream settings.

### DIFF
--- a/src/odemis/gui/comp/buttons.py
+++ b/src/odemis/gui/comp/buttons.py
@@ -30,7 +30,7 @@
 
 import logging
 import math
-from typing import Any, List
+from typing import Any, List, Tuple, Dict, Callable, Optional
 
 import numpy
 import wx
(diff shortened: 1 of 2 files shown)


K4rishma (approved):

pieleric:
@K4rishma , there was a bug in my original version (as caught by CodeRabbit). I fixed it. It requires quite a lot more changes... Do you mind to give a second look?

K4rishma (approved):

PR #3190, opened 2025-08-14
[ci] Add logging timeouts and other errors to runtests.sh

tepals:
The status of calling the test case from bash was logged in the full test report, but nothing was done with it. Now if the status is not 0 (OK) it is also copied to the summary. Especially for timeouts this is usefull to know, because if test cases time out failures can get hidden.

### DIFF
--- a/util/runtests.sh
+++ b/util/runtests.sh
@@ -85,13 +85,30 @@ if [ $pytest = True ]; then
   # Remove any files which might already exist from a previously (expected incomplete) run of that day.
   rm -f $SHORTSUMMARY
   touch $SHORTSUMMARY
+  echo "================== Short summary for $DATE =================" > "$SHORTSUMMARY"
   SHORTSUMMARY="$(readlink -m "$SHORTSUMMARY")"
 
   WARNINGSUMMARY=/tmp/pytest-warning-summary.log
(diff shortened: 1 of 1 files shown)


util/runtests.sh
```diff
@@ -146,6 +163,27 @@ for f in $testfiles; do
 
         status=$?
         echo $f returned $status >> "$TESTLOG" 2>&1
+        if [ "$pytest" = True ]; then
+          if [ "$status" -eq 124 ]; then
+              echo "Test $f timed out after $MAXTIME seconds" >> "$TESTLOG"
+              echo "Test $f timed out after $MAXTIME seconds" >> "$TIMEOUTSUMMARY"
```
pieleric, line 169: Maybe it'd be easier to parse for later scripts if the output looks more like the pytest short summary.
Eg: 
```
FAILED src/odemis/driver/test/piezomotor_test.py::TestPMD401OpenLoop::test_queued - OSError: F...
```
So you could output something like:
```
TIMEOUT $f
```
Then, eventually it has to be picked up by the "email writer". I don't know precisely where is this code...

pieleric: I think it'd still be easier parse if the output is something like "TIMEOUT $f" or "FAILED $f" , but I leave it up to you. You can make a separate PR for updating the filter to read this part of the summary.

PR #3195, opened 2025-08-14
[ci] Fix the pytest log filter when there is no summary

tepals:
When the test case times out before the short test summary was written, the failures where not reported.


tmoerkerken (approved):

K4rishma (approved):

tepals:
I updated the code to make sure all lines start with FAILED, this will help with future parsing by other scripts.

PR #3198, opened 2025-08-14
[refactor] Introduce new util.mock.SimulatedAxis to simulate axes

pieleric:
A generic, good, simulator for axes. Already replacing the TMCM and Piezomotor simulators. The hope is that it improves the behaviour of the Piezomotor simulator. More simulators could be converted, later on.

### DIFF
--- a/src/odemis/driver/piezomotor.py
+++ b/src/odemis/driver/piezomotor.py
@@ -19,7 +19,6 @@ PURPOSE. See the GNU General Public License for more details.
 You should have received a copy of the GNU General Public License along with
 Odemis. If not, see http://www.gnu.org/licenses/.
 """
-import copy
 import glob
 import logging
 import os
@@ -42,7 +41,7 @@ from odemis.model import (
     ParallelThreadPoolExecutor,
     isasync,
 )
-from odemis.util import to_str_escape
+from odemis.util import to_str_escape, mock
 from odemis.util.driver import getSerialDriver
 
 BAUDRATE = 115200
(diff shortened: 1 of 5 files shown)


tmoerkerken:
Do you think this is autoscript/xt compatible? Do you think we can introduce some optional inaccuracies, like the stage not exactly ending up at the target, or some randomness in the position. We are facing stage issues with TFS (https://delmic.atlassian.net/browse/NCT-386) and I wondered if we could simulate the stage more accurately, to use in developing a fix.

src/odemis/driver/piezomotor.py
```diff
@@ -1128,53 +1127,57 @@ def _parseMessage(self, msg):
...
+                    self._axes[axis].speed = speed_enc
+                    self._axes[axis].move_abs(target_pos)
                 else:
                     raise ValueError()
-            elif cmd == "S":  # stop axis
-                self.is_moving = False
-                self.closed_loop[axis] = False
-            elif cmd == "C":
-                # Relative move
+            elif cmd == "C":  # Relative move
```
nandishjpatel, line 1146: Is the command `elif cmd == "S":  # stop axis` in line 1149 not needed anymore?

nandishjpatel (approved):

pieleric:
> Do you think this is autoscript/xt compatible? Do you think we can introduce some optional inaccuracies, like the stage not exactly ending up at the target, or some randomness in the position. We are facing stage issues with TFS (https://delmic.atlassian.net/browse/NCT-386) and I wondered if we could simulate the stage more accurately, to use in developing a fix.

@tmoerkerken Yes, we could add a "accuracy" parameter to indicate a range within which the move ends (picked randomly when the move is completed). By default it would be 0, but for simulating more correctly a real stage, we could use for instance `1e-6`.

PR #3201, opened 2025-08-20
[fix] FIBSEM tab posture switch handling

tmoerkerken:


tmoerkerken:
I showed the implementation to Deniz, and she thinks we might need to move in SEM view, when in MILLING posture

src/odemis/gui/cont/tabs/fibsem_tab.py
```diff
@@ -349,7 +354,20 @@ def _stop_streams_subscriber(self):
     def _start_streams_subscriber(self):
         self.tab_data_model.streams.subscribe(self._on_acquired_streams)
 
-    def on_dbl_click(self, evt):
+    def on_sem_stage_movement(self, evt):
+        active_canvas = evt.GetEventObject()
+        logging.debug(f"mouse down event, canvas: {active_canvas}")
+
+        if not active_canvas._tab_data_model.tool.value and self.pm.current_posture.value != SEM_IMAGING:
```
pieleric, line 361: The tab_data_model should be the same for all the controls in this tab, so you can just use `self.tab_data_model` (and hence, no need to call `GetEventObject()`.

Also, `tool.value` is more like an enumeration, so it's confusing to use `not`. I think it'd be clearer to write ` tool.value == TOOL_NONE`.

src/odemis/gui/cont/tabs/fibsem_tab.py
```diff
@@ -349,7 +354,20 @@ def _stop_streams_subscriber(self):
...
+            logging.warning(("Moving the stage from the SEM stream is currently disabled. Please set the stage to the "
```
pieleric, line 362: No need to have double parenthesis.

src/odemis/gui/cont/tabs/fibsem_tab.py
```diff
@@ -349,7 +354,20 @@ def _stop_streams_subscriber(self):
...
+                           "SEM posture to proceed. You can still use the FIB view for movement."
+                            "Warning"))
```
pieleric, line 364: Can you add a comment on why it's find to still propagate the event? Some other code will also detect this state and drop (silently) the move?

pieleric (changes requested):

tmoerkerken:
Pia at Jülich reported that there was an issue placing features from the localization tab. I need to test this more extensively before this is being merged.

tmoerkerken:
Closing, because the PR introduces multiple features of which a few are undesired. I will make small feature branches, based on priority.

PR #3202, opened 2025-08-26
[fix] FM imaging fixed z on TFS

tmoerkerken:
Introduce an option to use a fixed FM imaging plane. Snapshots the position before doing so, so switching back to another posture will restore the previous position. Added unit tests to verify the changes.

### DIFF
--- a/src/odemis/acq/move.py
+++ b/src/odemis/acq/move.py
@@ -613,14 +613,17 @@ class MeteorPostureManager(MicroscopePostureManager):
         ppos[self._axes_dep["y"]] = pinv[2]
         return ppos
 
-    def from_sample_stage_to_stage_position(self, pos: Dict[str, float]) -> Dict[str, float]:
+    def from_sample_stage_to_stage_position(self, pos: Dict[str, float], posture: Optional[int] = None) -> Dict[str, float]:
         """
         Get stage position coordinates from sample stage coordinates (sample-stage -> stage-bare).
         :param pos: position in the sample-stage coordinates
+        :param posture: The posture to use for the transformation. If None, uses the current posture.
+            Valid values include SEM_IMAGING, FM_IMAGING, MILLING, FIB_IMAGING, etc.
         :return: position in the stage-bare coordinates
         """
         q = numpy.array([pos["x"], pos["y"], pos["z"]])
-        posture = self.current_posture.value
+        if posture is None:
+            posture = self.current_posture.value
         pinv = self._transforms[posture] @ q
 
         # add orientation (rx, rz)
@@ -633,14 +636,17 @@ class MeteorPostureManager(MicroscopePostureManager):
         ppos.update(orientation)
         return ppos
 
-    def to_sample_stage_from_stage_position(self, pos: Dict[str, float]) -> Dict[str, float]:
+    def to_sample_stage_from_stage_position(self, pos: Dict[str, float], posture: Optional[int] = None) -> Dict[str, float]:
         """
         Get sample stage coordinates from stage coordinates. (stage-bare -> sample stage)
         :param pos: position in the stage-bare coordinates
+        :param posture: The posture to use for the transformation. If None, uses the current posture.
+            Valid values include SEM_IMAGING, FM_IMAGING, MILLING, FIB_IMAGING, etc.
         :return: position in the sample-stage coordinates
         """
         p = numpy.array([pos["x"], pos[self._axes_dep["x"]], pos[self._axes_dep["y"]]])
-        posture = self.current_posture.value
+        if posture is None:
+            posture = self.current_posture.value
         q = self._inv_transforms[posture] @ p
 
         qpos = {}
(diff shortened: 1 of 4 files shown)


src/odemis/acq/move.py
```diff
@@ -1092,6 +1108,16 @@ def _transformFromSEMToMeteor(self, pos: Dict[str, float]) -> Dict[str, float]:
         transformed_pos["y"] = md_calib["dy"] - pos["y"]
         transformed_pos.update(fm_pos_active)
 
+        if fix_fm_plane:
+            # transformed_pos -> sample_stage
+            sample_stage_pos = self.to_sample_stage_from_stage_position(transformed_pos, posture=FM_IMAGING)
+            # store original working distance snapshot for recovery
+            self.fm_sample_z_snapshot = sample_stage_pos["z"]
```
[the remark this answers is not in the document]
tmoerkerken: Yeah, I agree the name was not ideal, but I couldn't think of something better. Will use one of your suggestions.

src/odemis/acq/test/move_test.py
```diff
@@ -589,6 +589,62 @@ def test_stage_to_chamber(self):
...
+    def test_revert_from_fixed_fm_z(self):
+        """
+        Test that when requesting a fixed fm plane posture switch, we also revert to the previous stage bare position.
+        """
+        translational = self.stage_grid_centers[POSITION_NAMES[GRID_1]]
+        rotational = self.stage_bare.getMetadata().get(model.MD_FAV_SEM_POS_ACTIVE)
+        position_base = {**translational, **rotational}
+
+        # Go from SEM posture to METEOR, and back. Check that we end up at the same spot as before
+        position_initial = {**position_base, "z": 0.025}
```
tepals, line 631: very minor comment: maybe it's a bit more clear if this is `position_initial_sem`

tmoerkerken: Fair point, will change it to `position_sem_initial`, since that is more in line with the `position_fm_initial` I already had.

src/odemis/acq/move.py
```diff
@@ -1092,6 +1109,16 @@ def _transformFromSEMToMeteor(self, pos: Dict[str, float]) -> Dict[str, float]:
...
 
+        if fix_fm_plane:
+            # transformed_pos -> sample_stage
+            sample_stage_pos = self.to_sample_stage_from_stage_position(transformed_pos, posture=FM_IMAGING)
+            # store original working distance snapshot for recovery
+            self.fm_sample_z_starting_point = sample_stage_pos["z"]
+            # sample_stage_pos[z] override
+            sample_stage_pos["z"] = self.fixed_fm_sample_z
+            # overridden sample_stage_pos -> stage bare
+            transformed_pos = self.from_sample_stage_to_stage_position(sample_stage_pos, posture=FM_IMAGING)
```
pieleric, line 1120: We shouldn't store and restore the Z position of the SEM every time this function is called. This function is called for various reasons, much more than just before switching position. For instance it can be called when converting the position of a Feature (at creation), or during the correlation step. See calls to `to_posture()` in the code of Odemis. 
So, instead, I'd suggest to only store explicitly the Z position when switching from SEM to FM in `_doCryoSwitchSamplePosition()` .

When going (back) from FM to SEM, it should check if such Z position is known, and use it. That should happen in `getTargetPosition()`, which is called only just before switching position. The Z of the SEM stage can be stored similarly to the Z of the FM stage: as a Z of the sample stage (when in SEM posture). 

Also, I think these (2) values should be stored in the metadata of the stage-bare, see the other comment.

src/odemis/acq/move.py
```diff
@@ -1108,12 +1135,22 @@ def _transformFromMeteorToSEM(self, pos: Dict[str, float]) -> Dict[str, float]:
...
 
+        if self.fm_sample_z_starting_point is not None:
+            # transformed_pos -> sample_stage
+            sample_stage_pos = self.to_sample_stage_from_stage_position(transformed_pos, posture=FM_IMAGING)
+            # sample_stage_pos[z] override
+            sample_stage_pos["z"] = self.fm_sample_z_starting_point
+            # overridden sample_stage_pos -> stage bare
+            transformed_pos = self.from_sample_stage_to_stage_position(sample_stage_pos, posture=FM_IMAGING)
+            # Unset the snapshotted position
+            self.fm_sample_z_starting_point = None
```
pieleric, line 1146: As mentioned in my previous comment, this part should be in in `getTargetPosition()`, when current posture is FM, and target is SEM.

src/odemis/acq/move.py
```diff
@@ -1064,18 +1070,29 @@ def __init__(self, microscope):
         if model.MD_FAV_FIB_POS_ACTIVE in stage_md:
             self.postures.append(FIB_IMAGING)
 
+        # For FM imaging, fix the imaging plane so that all features always lie on the same plane.
+        # Pick a sane default for the sample stage z using grid 1.
+        sem_grid1_pos = stage_md[model.MD_SAMPLE_CENTERS][POSITION_NAMES[GRID_1]]
+        sem_grid1_pos.update(stage_md[model.MD_FAV_SEM_POS_ACTIVE])
+        sem_grid1_pos_fm = self._transformFromSEMToMeteor(sem_grid1_pos, fix_fm_plane=False)
+        self.fixed_fm_sample_z = self.to_sample_stage_from_stage_position(sem_grid1_pos_fm, posture=FM_IMAGING)["z"]
+        self.fm_sample_z_starting_point = None  # allocate a variable for storing the sample z before fixing it later on
```
pieleric, line 1079: Instead of storing these 2 values on the class, I'd suggest to store them on the `stage-bare`, as extra metadata. This has 2 advantages: allow to observe them more easily (for debugging in case of issues), and if Odemis GUI is closed/opened, the state is still stored.
The new metadata names could be something like: SEM_POS_SAMPLE_ACTIVE and FM_POS_SAMPLE_ACTIVE. They contain a "position" (so dict[str, float]), in *sample stage* coordinates, which normally is either empty (ie, undefined position) or has a "z" axes. The other axes are unused.

It's fine if the stage-bare at init doesn't have any of these metadata defined (this way it's compatible with the current microscope files). However, if FM_POS_SAMPLE_ACTIVE is defined at init, then it is directly used, instead of doing the computation above.
For SEM_POS_SAMPLE_ACTIVE, it will be stored just before a switch from SEM to FM, and used when switching from FM to SEM.

pieleric (changes requested):

src/odemis/acq/move.py
```diff
@@ -1063,26 +1086,41 @@ def __init__(self, microscope):
             self.postures.append(MILLING)
         if model.MD_FAV_FIB_POS_ACTIVE in stage_md:
             self.postures.append(FIB_IMAGING)
+        # If there is no known fixed sample z for FM, compute it here and store it for later use
+        if not stage_md.get(model.MD_FM_POS_SAMPLE_ACTIVE):
+            # For FM imaging, fix the imaging plane so that all features always lie on the same plane.
+            # Pick a sane default for the sample stage z using grid 1.
+            sem_grid1_pos = dict(stage_md[model.MD_SAMPLE_CENTERS][POSITION_NAMES[GRID_1]])
+            sem_grid1_pos.update(stage_md[model.MD_FAV_SEM_POS_ACTIVE])
+            sem_grid1_pos_fm = self._transformFromSEMToMeteor(sem_grid1_pos)
```
pieleric, line 1095:
```suggestion
            sem_grid1_pos_fm = self._transformFromSEMToMeteor(sem_grid1_pos, fix_fm_plane=False)
```
Mostly in case we restart Odemis GUI and the GRID1 has changed, we want to use this new one.

src/odemis/acq/move.py
```diff
@@ -1063,26 +1086,41 @@ def __init__(self, microscope):
...
+    def _transformFromSEMToMeteor(self, pos: Dict[str, float], fix_fm_plane: bool = True) -> Dict[str, float]:
         """
         Transforms the current stage position from the SEM imaging area to the
         meteor/FM imaging area.
-        :param pos: (dict str->float) the initial stage position.
-        :return: (dict str->float) the transformed position.
+        :param pos: the initial stage position.
+        :param fix_fm_plane: If True, maintains a consistent FM imaging plane by keeping the z-coordinate
+            in sample-stage space constant. The original z value is stored in fm_sample_z_starting_point for
+            restoring the z height in sem posture later. If False, performs direct transformation without z-adjustment.
```
pieleric, line 1114:
```suggestion
        :param fix_fm_plane: If True, maintains a consistent FM imaging plane by keeping the z-coordinate
            in sample-stage space constant. It uses the z value is stored in MD_FM_POS_SAMPLE_ACTIVE. 
            If False, performs direct transformation without z-adjustment.
```

pieleric (changes requested):

PR #3214, opened 2025-09-05
[fix] test cases in move_test.py for zeiss

K4rishma:
Two tests were not modified for the ZeissPosture manager

src/odemis/acq/test/move_test.py:563 (TestMeteorZeiss1Move.test_unknown_label_at_initialization)
self = <move_test.TestMeteorZeiss1Move testMethod=test_unknown_label_at_initialization>

    def test_unknown_label_at_initialization(self):
        arbitrary_position = {"x": 0.0, "y": 0.0, "z": -3.0e-3}
>       self.stage.moveAbs(arbitrary_position).result()

src/odemis/acq/test/move_test.py:566: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/lib/python3/dist-packages/Pyro4/futures.py:315: in result
    return self.__get_result()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <ClientFuture at 0x7f3f1d713910 for Proxy of Component 'AntiBacklash Stage'>

    def __get_result(self):
        if self._exception:
>           raise self._exception
E           ValueError: Position -0.003 for axis z outside of range 0.000000->0.050000

/usr/lib/python3/dist-packages/Pyro4/futures.py:299: ValueError

----------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_transformFromSEMToMeteor(self):
 def _transformFromSEMToMeteor(self, pos: Dict[str, float]) -> Dict[str, float]:
        """
        Transforms the current stage position from the SEM imaging area to the
        meteor/FM imaging area.
        :param pos: (dict str->float) the current stage position.
        :return: (dict str->float) the transformed position.
        """
        transformed_pos = pos.copy()
    
        # Call out calibrated values and stage tilt and rotation angles
        calibrated_values = stage_md[model.MD_CALIB]
        fm_pos_active = stage_md[model.MD_FAV_FM_POS_ACTIVE]
        sem_pos_active = stage_md[model.MD_FAV_SEM_POS_ACTIVE]
    
        # Define values that are used more than once
>       rx_sem = pos["rx"]  # Current tilt angle (can differ per point of interest)
E       KeyError: 'rx'

src/odemis/acq/move.py:1308: KeyError

### DIFF
--- a/src/odemis/acq/test/move_test.py
+++ b/src/odemis/acq/test/move_test.py
@@ -345,7 +345,9 @@ class TestMeteorZeiss1Move(TestMeteorTFS1Move):
     ROTATION_AXES = {'rx', 'rm'}
 
     def test_moving_in_grid1_fm_imaging_area_after_loading(self):
-        """Check if the stage moves in the right direction when moving in the fm imaging grid 1 area."""
+        """
+        Check if the stage moves in the right direction when moving in the fm imaging grid 1 area.
+        """
         super().test_moving_in_grid1_fm_imaging_area_after_loading()
 
         # move in the same imaging mode using linked YM stage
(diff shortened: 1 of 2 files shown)


tepals (approved):

src/odemis/acq/test/move_test.py
```diff
@@ -359,6 +361,39 @@ def test_moving_in_grid1_fm_imaging_area_after_loading(self):
...
+        # update the stage with rotation axes
+        grid1_pos.update(stage_md[model.MD_FAV_SEM_POS_ACTIVE])
+        grid2_pos.update(stage_md[model.MD_FAV_SEM_POS_ACTIVE])
+
+        # check if no error is raised (test fails if error is raised)
+        try:
+            self.posture_manager._transformFromSEMToMeteor(grid1_pos)
+            self.posture_manager._transformFromSEMToMeteor(grid2_pos)
+        except Exception as e:
+            self.fail(f"_transformFromSEMToMeteor raised error when it shouldn't: {e}")
```
pieleric, line 387: In test cases, you really don't need to catch exceptions, and convert them to a failure. That's automatically done by the testing framework.

Here, just get the result of the _transformFromSEMToMeteor() and check (in a simple way) that the result is sane. For instance, just check that axes (ie, the keys of the dict) are the ones you expect.

src/odemis/acq/test/move_test.py
```diff
@@ -359,6 +361,39 @@ def test_moving_in_grid1_fm_imaging_area_after_loading(self):
...
+
+    def test_unknown_label_at_initialization(self):
+        arbitrary_position = {"x": 0.0, "y": 0.0, "z": 0.0e-3}
```
pieleric, line 390: It's not just "arbitrary", it's mainly outside of all ranges. Can you add a comment explaining this.

PR #3234, opened 2025-10-03
[fix] e-k solid angle compenstation issues

nandishjpatel:
Support ticket: https://support.delmic.com/scp/tickets.php?id=2194

If no background image is provided just remove a constant noise estimated from the corners of the image. This helps in better visualisation, given project_angular_spectrum_to_grid_5d now carries out the solid angle correction.

Now apply the angle range as a post-processing step, rather than previously using it during interpolation. Crop/extend the interpolated data onto a new theta grid.

Avoid the padding (left, right) during interpolation as it creates stair like pattern in the data.

### DIFF
--- a/src/odemis/acq/test/calibration_test.py
+++ b/src/odemis/acq/test/calibration_test.py
@@ -394,7 +394,8 @@ class TestSpectrum(unittest.TestCase):
         wl_calib = 400e-9 + numpy.arange(dcalib.shape[0]) * 10e-9
         calib = model.DataArray(dcalib, metadata={model.MD_WL_LIST: wl_calib})
 
-        compensated = calibration.apply_spectrum_corrections(spec, coef=calib)
+        dummy_bckg = model.DataArray(numpy.zeros_like(spec), spec.metadata)
+        compensated = calibration.apply_spectrum_corrections(spec, bckg=dummy_bckg, coef=calib)
 
         self.assertEqual(spec.shape, compensated.shape)
         numpy.testing.assert_equal(spec.metadata[model.MD_WL_LIST],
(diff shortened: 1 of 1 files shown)


src/odemis/util/angleres.py
pieleric, line 859: Shouldn't `data_lin` be filled by `nan` instead of zeros at init?, to be certain there are no "left-over" pixel at the end? Or alternatively, if you're certain the data is entirely replaced (which actually seems to be the case looking at the code), it should be set as `empty` (which is a tiny bit faster, but also a statement that the initial values don't matter).

src/odemis/util/angleres.py
```diff
@@ -874,9 +874,32 @@ def project_angular_spectrum_to_grid(
         omega_list_short = omega_list_raw[first_px: last_px + 1]
         d = data[first_px: last_px + 1, wli]
         d_corrected = d / omega_list_short
-        # Pad with 0.0 for points outside (left and right) the source data's angular range.
+        # Pad with numpy.nan for points outside (left and right) the source data's angular range.
         # This prevents creating artificial signal at the edges from clamping.
-        data_lin[:, wli] = numpy.interp(theta_list_linear, theta_list_short, d_corrected, left=0.0, right=0.0)
+        data_lin[:, wli] = numpy.interp(theta_list_linear, theta_list_short, d_corrected, left=numpy.nan, right=numpy.nan)
+
```
pieleric, line 880: Please add a comment here explaining the goal of this whole part. Alternatively, you could even put this part into its own function (as it's pretty much independent from the rest of the variables) and put the comment as the docstring of the function.

pieleric (changes requested):

PR #3243, opened 2025-10-16
[fix] "groom" new PNGs

pieleric:
Can save some disk space by compressing the PNG images better
(lossless).
Moreover, some of the metadata in PNG can confuse Windows. Removing it
helps.


K4rishma (approved):

tepals (approved):
It might be overkill, but I feel like we could turn this into a github action. Especially since we don't add png's that often, it might be nice to not forget about doing this.

PR #3259, opened 2025-11-03
[feat] util img: dedicated function for converting RGB data to RGB8

pieleric:
Converting from RGB to RGB should be simple... and it used to be handled
by essentially copying the data. Just a basic code to handle tint.
However, this works well only if the data is already uint8.
For uint16, this didn't work. Also, the range wasn't computed.

=> introduce projectYXC2RGB8() to handle explicitly such method, and use
in when the data is in "YXC" format.
In particular, this helps opening Tescan RGB 16-bits images.

### DIFF
--- a/src/odemis/util/img.py
+++ b/src/odemis/util/img.py
@@ -33,7 +33,7 @@ from odemis.model import DataArray
 from odemis.model import MD_DWELL_TIME, MD_EXP_TIME, TINT_FIT_TO_RGB, TINT_RGB_AS_IS
 from odemis.util import get_best_dtype_for_acc, transform
 from odemis.util.conversion import get_img_transformation_matrix, rgb_to_frgb
-from typing import Tuple, List, Union
+from typing import Tuple, List, Optional
 import matplotlib.colors as colors
 from matplotlib import cm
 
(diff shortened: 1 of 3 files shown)


src/odemis/util/img.py
```diff
@@ -486,6 +486,98 @@ def DataArray2RGB(data, irange=None, tint=(255, 255, 255)):
...
+    """
+    :param data: 2D image (shape YXC) with RGB or RGBA channels
+    :param irange: (None or tuple of 2 values) min/max intensities mapped
+        to black/white
+        None => auto (min, max are from the data);
+        0, max val of data => whole range is mapped.
+        min must be < max, and must be of the same type as data.dtype.
+    :param tint: (3-tuple of 0 < int <256) RGB colour of the final image (each
+        pixel is multiplied by the value. Default is white.
+    :return: (numpy.ndarray of 3*shape of uint8) converted image in RGB with the
```
tmoerkerken: The dimensionality is a bit confusing to me. And also, it could be RGBA as well.

tmoerkerken (changes requested):

ilyushkin (approved):

tmoerkerken (approved):

PR #3300, opened 2025-12-09
[feat] CLI: add auto completion script for fish

pieleric:
We already have completion for bash, but for developers (and advanced
users) which use fish, it's nice to also have auto-completion.


tepals (approved):
I think the comments from coderabbit and copilot are correct, please look at them. Rest looks good

tmoerkerken (approved):

PR #3302, opened 2025-12-12
[feat] omicronxx: optimize speed when changing the light intensity

pieleric:
The most typical change is to change just one channel. However, until
now, the driver would always update all channels. Changing a channel can
take ~50ms. It's not so long, but on systems with 4 channels, when
turning on and off one channel quickly, that would end up costing around
300ms overhead.

=> Only send commands for the channel that changed


tmoerkerken (approved):
A few unit tests would be nice to prove it works :)

K4rishma (approved):

PR #3349, opened 2026-02-06
[fix] PostureManager: adjust orientation of sample stage in SEM imaging posture

pieleric:
On TFS1 and Zeiss, the position of the features was shown rotated by 180° in the SEM imaging posture.
    
* Rewrite the transformation matrices formula to be clearer
* FM imaging direction should be 180° opposite to the SEM imaging, as rx is set to +180°
* Similarly to Tescan, assume that if the SEM has scan rotation, then CCD is also rotated, to match SEM orientation
* Also introduce a new default scan rotation. It's 180° on TFS and Zeiss.

In practice, this means that systems with e-beam and ion-beam components with 180° scan rotation, there is not change. On systems with these components, but 0° scan rotation, the sample stage in FM posture will be reversed in XY. There shouldn't be any such system, but if that's the case, the fix is to either use a 180° scan rotation, or adjust "transp" on the CCD.
On systems without these components, there will be no change in FM, but in SEM posture, the feature will be in the same orientation.

### DIFF
--- a/src/odemis/acq/move.py
+++ b/src/odemis/acq/move.py
@@ -327,6 +327,10 @@ class MeteorPostureManager(MicroscopePostureManager):
         self._metadata = {}
         self._axes_dep = {}  # axes dependencies between different planes
 
+        # Assumed SEM scan rotation if it cannot be read from the hardware (e-beam)
+        # Used for displaying the features and moving the stage in SEM imaging posture in the right orientation.
+        self._default_scan_rotation = 0  # rad
+
         # pre-tilt is required for milling posture, but not all systems have it
         stage_md = self.stage.getMetadata()
         md_calib = stage_md.get(model.MD_CALIB, {})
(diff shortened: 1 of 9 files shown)


src/odemis/acq/move.py
```diff
@@ -555,43 +559,65 @@ def _initialise_transformation(
...
-        self._transforms = {FM_IMAGING: tf,
-                            FIB_IMAGING: tf_sr,
-                             SEM_IMAGING: tf_inv_sr,
-                             MILLING: tf_inv_sr,
-                             UNKNOWN: tf_inv_sr}
+        self._transforms = {
+            FM_IMAGING: tf_fm,
+            SEM_IMAGING: tf_sem,
+            FIB_IMAGING: tf_fib_im,
+            MILLING: tf_sem,
```
K4rishma: For milling, should we take into account milling angle for the sample plane?

K4rishma (approved):

tepals (approved):

PR #3351, opened 2026-02-06
[fix] driver xt_client: also handle raw coordinate offset on rotation axes

pieleric:
Normally, the stages have no offset between raw and specimen coordinate systems on the rotation axes (rx & rz)... except we found one system where the rz offset is 180.2°. In such case, the rz value didn't match the one reported on the TFS GUI, which is quite confusing.

This is the equivalent change as already done for autoscript in commit 2442b63f679b3 .

### DIFF
--- a/src/odemis/driver/xt_client.py
+++ b/src/odemis/driver/xt_client.py
@@ -2197,11 +2197,14 @@ class Stage(model.Actuator):
          systems.
         """
         self._raw_coordinates = raw_coordinates
-        # When raw coordinate system is selected, in theory just z should change
-        # but in practice x and y change by a fixed offset. When raw coordinate system is used,
-        # offset correction is applied (in the later part of init)
-        # such that all axes apart from z, have same values
-        self._raw_offset = {"x": 0, "y": 0}
+        # We use a "hybrid" coordinate system: the main advantage of the raw coordinate system
+        # is that the Z axis is never "linked" to the working distance. When the Z axis is linked,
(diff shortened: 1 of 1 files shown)


src/odemis/driver/xt_client.py
```diff
@@ -2197,11 +2197,14 @@ def __init__(self, name: str, role: str, parent: model.HwComponent, rng: dict =
...
-        # such that all axes apart from z, have same values
-        self._raw_offset = {"x": 0, "y": 0}
+        # We use a "hybrid" coordinate system: the main advantage of the raw coordinate system
+        # is that the Z axis is never "linked" to the working distance. When the Z axis is linked,
+        # it's oriented in the other direction (0 is at the pole-piece, and an increase goes down),
+        # and is dependent on the SEM working distance (focus).
+        # In raw coordinates, Z always means the same thing, independent of what the user has done
+        # in the GUI. However, for the other axes, there might be an offset, but as it's convenient
+        # to show the same value as in the TFS GUI, we compensate for that offset.
+        self._raw_offset = {"x": 0, "y": 0, "rx": 0, "rz": 0}
```
K4rishma: There is a possibility for z to have an offset in the same raw coordinate system, as we have seen offsets already in x, y and rz.  I think there is no harm in adding z if we are adding all the other 4 axes.

K4rishma (approved):

ilyushkin (approved):

PR #3355, opened 2026-02-10
[test] PostureManager test: avoid running TFS tests also in the Zeiss and Tescan tests

pieleric:
The Zeiss and Tescan tests inherits from the TFS tests, to avoid code
duplication.
However, because the files imported directly the test classes (eg, TestMeteorTFS1Move),
these classes were concidered part of the file, and so also run during
the testing.

=> Import only the module containing the classes, so that the test classes are
not directly in the main module, and so they won't be automatically run
by pytest (or unittest)

### DIFF
--- a/src/odemis/acq/test/move_tescan_test.py
+++ b/src/odemis/acq/test/move_tescan_test.py
@@ -23,9 +23,9 @@ import unittest
 
 import odemis
 from odemis import model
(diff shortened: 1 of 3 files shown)


tepals (approved):
I think ideally there would be a base test module, with a base testing class `MoveBaseTest` that tescan , zeiss and tfs all inherit from. If that's too much work for now you can leave it as is.

PR #3374, opened 2026-02-20
[MSD-376][feat] Plugin to acquire a z-stack by moving the stage

pieleric:
z-stack is normally acquired by moving the focus (lens). In some
experiments, it's handy to move the stage (in Z) instead.
This, very basic, plugin adds a menu option "Acquisition/Stage Z-Stack..."
on the METEOR to run the current acquisition with z-stack options, using
the stage.

### DIFF
--- a/plugins/zstack_stage.py
+++ b/plugins/zstack_stage.py
@@ -1,0 +1,257 @@
+# -*- coding: utf-8 -*-
+"""
+Created on 12 Feb 2026
+
+@author: Éric Piel
+
+Plugin for the METEOR (and MIMAS).
+Acquire a z-stack using the sample stage z axis instead of the focus actuator.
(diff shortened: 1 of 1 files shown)


plugins/zstack_stage.py
```diff
@@ -0,0 +1,257 @@
...
+                dlg.ShowModal()
+                dlg.Destroy()
+
+            # Save the data automatically
+            self._save_data(data)
+
+            logging.info("Stage Z-Stack acquisition completed successfully")
+
+        except Exception as e:
+            logging.exception("Stage Z-Stack acquisition failed: %s", e)
```
nandishjpatel:
```suggestion
            logging.exception("Stage Z-Stack acquisition failed")
```

nandishjpatel (approved):

PR #3393, opened 2026-03-04
[feat] add goffset andorshrk

yxkdejong:


src/odemis/driver/andorshrk.py
```diff
@@ -2064,6 +2090,45 @@ def _doSetFlipper(self, flipper: int, pos: int, check: bool):
...
+                          current_grat_offset, current_det_offset)
+
+            if grating == 1:
+                if not allow_grating_offset:
+                    logging.debug("Grating offset update disabled (grating=1, target=%d)",target_offset,)
+                else:
+                    grating_offset = target_offset - current_det_offset
+                    self.SetGratingOffset(grating, grating_offset)
+
+            elif grating > 1:
```
pieleric: Just `else`. Then you're certain to handle every cases.

src/odemis/driver/andorshrk.py
```diff
@@ -496,7 +501,8 @@ def __init__(self, name: str, role: str,
 
             axes = {"wavelength": model.Axis(unit="m", range=wl_range,
                                              speed=(max_speed, max_speed)),
-                    "grating": model.Axis(choices=gchoices)
+                    "grating": model.Axis(choices=gchoices),
+                    "goffset": model.Axis(unit=None, range=((GRAT_OFFSET_MIN+DET_OFFSET_MIN), (GRAT_OFFSET_MAX+DET_OFFSET_MAX)))
```
pieleric: Please adjust the formatting. Make it 2 lines, and with spaces around `+`.

pieleric (changes requested):

pieleric (changes requested):

pieleric (changes requested):

tepals (changes requested):

tepals (approved):

src/odemis/driver/andorshrk.py
pieleric: This whole section is duplicate of the previous code (+ the range check which will never be triggered). I guess it should be dropped.

src/odemis/driver/andorshrk.py
```diff
@@ -2064,6 +2097,122 @@ def _doSetFlipper(self, flipper: int, pos: int, check: bool):
...
+            input_ports.append(self.GetFlipperMirror(INPUT_FLIPPER))
+
+        output_ports = [DIRECT_PORT]
+        if "flip-out" in self.axes:
+            output_ports.append(self.GetFlipperMirror(OUTPUT_FLIPPER))
+
+        all_detectors = [(i, o) for i in input_ports for o in output_ports]
+
+        current_detector = (flip_in_pos, flip_out_pos)
+        reference_detector = all_detectors[0]
```
pieleric: Note that on the Kymera 328, there are both input and output ports. However, in the (2) systems that we have, the input port is always set to SIDE. So grating offset calibration will always move the detector offset (for all 4 gratings)... and not do the right thing :-( 
I'm not certain what is the right approach though... . For now the alignment code doesn't support this "spectrograph-dedicated" component either, so it might be just fine to add a "WARNING" comment here explaining the limitation, and we'll think of it again once this support is required. Or maybe you have a good idea already?

pieleric (changes requested):

src/odemis/driver/andorshrk.py
yxkdejong: Can we not just say that it should only adjust the grating offset for one detector, regardless of whether the system sees this as the first or second detector? So that if the camera is connected to the side ports (and it's the only one) it will still adjust the grating offset? Then if there is a second detector available to it only adjusts the detector offset?

pieleric: The problem is that the driver & hardware don't know whether there is a detector or not.  All it knows is the position of the input and output selectors (requested by Odemis). Odemis reads the info from the microscope file, but this is not passed to the driver. 
I think the simplest for now is to default to the SIDE port on the input, if it is present.

src/odemis/driver/andorshrk.py
```diff
@@ -1883,6 +1883,9 @@ def moveAbs(self, pos):
             elif axis == "focus":
                 actions.append((axis, self._doSetFocusAbs, p))
             elif axis == "goffset":
+                if len(pos) > 1:
+                    logging.info("Ignoring 'goffset' in multi-axis move to preserve hardware calibration.")
+                    continue
```
pieleric: Humm... I guess this happened if a code tries to reset to "the original position", right? Can you let me know in which case precisely this happened?

Dropping the requested axis is a really odd behaviour. At least, the log should be raised to the warning level. I wonder if an alternative would be to just make sure this move happens as the very last one (so that the grating and flippers are the same as set when the goffset was read). If so, that would be a better solution.

pieleric (changes requested):

src/odemis/driver/andorshrk.py
yxkdejong: What happened was that when the detector offset was adjusted (for the second grating at Amolf) and we closed Odemis and restarted it, the grating offset for the second grating was changed to exactly the negative/positive value of the detector offset.

src/odemis/driver/andorshrk.py
```diff
@@ -1838,8 +1869,10 @@ def moveAbs(self, pos):
         self._checkMoveAbs(pos)
 
         # If grating needs to be changed, change it first, then the wavelength
-        ordered_axes = util.sorted_according_to(pos.keys(), ("grating", "wavelength"))
+        ordered_axes = util.sorted_according_to(pos.keys(), ("flip-in", "flip-out", "grating", "wavelength"))
```
pieleric: If you put "goffset" as the last one, then this should take care of the order completely, and you can remove the extra `goffset_action`.

src/odemis/driver/andorshrk.py
```diff
@@ -1926,6 +1964,10 @@ def _doSetGrating(self, g):
             self.SetGrating(g)
             # This is a trick, to immediately report the new position, in case
             # getPixelToWavelength() uses it. It's not notified.
+
+            # Allow residual mechanical vibrations to settle after the heavy turret rotates
+            # This time was needed when testing on hardware in the office and might be left out.
+            time.sleep(5)
```
pieleric: Let's move this to the very end of the function, after every move is finished. I think you can even put it after the updatePosition() call. 
BTW, the second part of the comment is odd. It's a bit hard to "leave it out" if it's hard-coded in the driver. Maybe just more explain the 5 seconds as "This duration has been tested on several systems and seems to be long enough for all cases".

src/odemis/driver/andorshrk.py
```diff
@@ -2064,6 +2106,124 @@ def _doSetFlipper(self, flipper: int, pos: int, check: bool):
...
+            input_ports.append(self.GetFlipperMirror(INPUT_FLIPPER))
+
+        output_ports = [DIRECT_PORT]
+        if "flip-out" in self.axes:
+            output_ports.append(self.GetFlipperMirror(OUTPUT_FLIPPER))
+
+        all_detectors = [(i, o) for i in input_ports for o in output_ports]
+
+        current_detector = (flip_in_pos, flip_out_pos)
+        reference_detector = all_detectors[0]
```
pieleric: I don't think this codes what it looks like it does. `input_ports` and `output_ports` can contain twice the same value (DIRECT_PORT).
`reference_detector` is *always* `DIRECT_PORT,DIRECT_PORT`
Eventually, all_detectors is only used by `is_multi_detector`, that always compute it as True, unless there is no flip-in and flip-out axes. So you could replace the whole code by:
`is_multi_detector =  "flip-out" in self.axes or "flip-out" in self.axes`

Let's do that, or did you have other things in mind?

pieleric (changes requested):

pieleric (approved):

PR #3394, opened 2026-03-04
[feat] added peak simulation

yxkdejong:


src/odemis/driver/simcam.py
```diff
@@ -413,6 +426,30 @@ def _simulate(self):
                      [int(round(ltrb[1] + i * binning[1])) for i in range(res[1])])
             sim_img = self._img[numpy.ix_(coord[1], coord[0])]  # copy
 
+        # spectrograph peak simulation
+        if getattr(self, "_spectrograph", None) is not None:
```
pieleric: The init code seems to always set `._spectrograph` (which makes sense). So you can just write `if self._spectograph:`

src/odemis/util/synthetic.py
```diff
@@ -128,6 +128,34 @@ def psf_gaussian(
...
+    if isinstance(shape, (int, numpy.integer)):
+        shape = (shape,)
+
+    # Determine resolution.y (height)
+    # If it's a 1D shape, resolution_y is 1
+    resolution_y = shape[0] if len(shape) > 1 else 1
+
+    # If an image is provided and it's a 2D request (y > 1)
+    if image is not None and resolution_y > 1:
+        return numpy.asanyarray(image).astype(dtype)
```
pieleric: This function is never called with `image != None` => remove the parameter and this block.

src/odemis/driver/simcam.py
```diff
@@ -413,6 +426,30 @@ def _simulate(self):
...
+            logging.info(f"DEBUG: x0_px={x0_px}, ltrb0={ltrb[0]}, result={peak_center_binned}")
+
+            width_binned = width_px/bin_x
+
+            peak = simulate_peak(amplitude=20000, x0=peak_center_binned, width=width_binned,
+                                shape=sim_img.shape, dtype=sim_img.dtype)
+
+            # set all values in sim_img to the minimal sim_img value
+            min_val = sim_img.min()
+            sim_img[...] = min_val
```
pieleric: Instead of getting the minimum of sim_img every time (which is potentially quite slow), and overriding sim_img every time, read it at init (based on `._img`), and keep it for the rest of the execution.

You can just then write *something like* `sim_img = peak + self._min_val`, which should be a lot faster... However, you should also pay attention to the overflow. Assuming you want to reproduce the same behaviour as the real hardware where the peak can easily saturate the CCD reading, then you should improve the way the base level (min) and the noise are added. You can look at the trick we use in _weaver.py `_adjust_brightness()` (line 223).

pieleric (changes requested):

tepals (changes requested):

tepals (approved):

src/odemis/driver/simcam.py
```diff
@@ -179,6 +181,22 @@ def clip_max_res(img_res):
...
             self._mirror = None
 
+        try:
+            spectrograph = dependencies["spectrograph"]
+            if not (
+                    isinstance(spectrograph, model.ComponentBase)
+                    and hasattr(spectrograph, "axes")
+                    and isinstance(spectrograph.axes, dict)
+                    and "goffset" in spectrograph.axes):
+                raise ValueError("spectrograph %s must have a 'goffset' attribute" % spectrograph)
```
pieleric: It's better to use f-strings: easier to read, and "safer". For instance, if "spectrograph" is a tuple (by mistake), the % will fail.
```suggestion
                raise ValueError(f"spectrograph {spectrograph} must have a 'goffset' attribute")
```

src/odemis/driver/simcam.py
```diff
@@ -195,6 +213,14 @@ def clip_max_res(img_res):
         self._error_creation_thread.daemon = True
         self._error_creation_thread.start()
 
+        self._min_val = self._img.min()
+        if numpy.issubdtype(self._img.dtype, numpy.integer):
+            self._dtype_max = numpy.iinfo(self._img.dtype).max
+        elif numpy.issubdtype(self._img.dtype, numpy.floating):
+            self._dtype_max = numpy.finfo(self._img.dtype).max
+        else:
+            self._dtype_max = numpy.inf
```
pieleric: Not used anywhere. => remove?

src/odemis/driver/simcam.py
```diff
@@ -413,12 +439,38 @@ def _simulate(self):
                      [int(round(ltrb[1] + i * binning[1])) for i in range(res[1])])
             sim_img = self._img[numpy.ix_(coord[1], coord[0])]  # copy
 
-        # Add some noise
+        # spectrograph peak simulation
+        if self._spectrograph:
+            current_offset = self._spectrograph.position.value["goffset"]
+
+            ccd_center_x = self._img_res[0]/ 2  # find the x-coordinate of the center of the ccd
```
pieleric: Add space around the operator:

```suggestion
            ccd_center_x = self._img_res[0] / 2  # find the x-coordinate of the center of the ccd
```

src/odemis/driver/simcam.py
```diff
@@ -413,12 +439,38 @@ def _simulate(self):
...
-        # There can still be some overflow, but let's just consider this "strong noise"
-        sim_img[sim_img > mx] = mx
+        noise_max = max(mx // 100, 10)
+
+            # generate noise and add directly
+        noise = numpy.random.randint(0, noise_max, sim_img.shape, dtype=self._img.dtype)
+        sim_img += noise
+
+            # final clamp to prevent overflow
+        sim_img = numpy.minimum(sim_img, mx)
```
pieleric: No. This doesn't handle overflow. For instance, if sim_img has 65530, and the noise is 10, it will overflow.
I think it should be:
```
sim_img += numpy.minimum(mx - sim_img, noise)
```
(or you can replace "mx" by dtype_max, to allow going above the maximum value in the image)

src/odemis/util/synthetic.py
```diff
@@ -128,6 +128,40 @@ def psf_gaussian(
...
+    :return: a numpy array containing the simulated peak, either as a 1D Gaussian or directly from the provided image
+    """
+
+    # Standardize shape to a tuple, so we can index it
+    if isinstance(shape, (int, numpy.integer)):
+        shape = (shape,)
+
+    # Generate the 1D Gaussian peak across the x-dimension
+    x = numpy.arange(shape[-1])
+    intensity = amplitude*(numpy.exp(-0.5 * ((x - x0) / width)**2))
```
pieleric:
```suggestion
    intensity = amplitude * (numpy.exp(-0.5 * ((x - x0) / width) ** 2))
```

src/odemis/util/synthetic.py
```diff
@@ -128,6 +128,40 @@ def psf_gaussian(
...
+
+    # Clip based on the actual dtype provided
+    info = numpy.iinfo(dtype)
```
pieleric: This only works if dtype is a int.

```suggestion
info = numpy.iinfo(dtype) if numpy.issubdtype(dtype, numpy.integer) else numpy.finfo(dtype)
```

pieleric (changes requested):
Please also have a look at code rabbit's comments.

src/odemis/util/synthetic.py
```diff
@@ -128,6 +128,44 @@ def psf_gaussian(
...
+    :return: a numpy array containing the simulated peak, either as a 1D Gaussian or directly from the provided image
+    """
+
+    # Standardize shape to a tuple, so we can index it
+    if isinstance(shape, (int, numpy.integer)):
+        shape = (shape,)
+
+    # Generate the 1D Gaussian peak across the x-dimension
+    if width <= 0:
+        raise ValueError("width should be positive")
```
pieleric: When raising errors about a value, it help a lot the debugging to know what was the value.
```suggestion
        raise ValueError(f"width ({width}) should be positive")
```

src/odemis/util/synthetic.py
```diff
@@ -128,6 +128,44 @@ def psf_gaussian(
...
+
+    x = numpy.arange(shape[-1], dtype=numpy.float64)
+    intensity = amplitude * (numpy.exp(-0.5 * ((x - x0) / width) ** 2))
+
+    # Clip based on the actual dtype provided
+    dtype = numpy.dtype(dtype)
```
pieleric: I don't think that line is necessary. At least, when testing with numpy 1.26, `numpy.iinfo()` works both with numpy.uint16 and numpy.array().dtype . 
If that line is required, please add a comment explaining what sort of corner case it handles.

src/odemis/driver/simcam.py
```diff
@@ -413,12 +433,38 @@ def _simulate(self):
...
-        # Clip, but faster than clip() on big array.
-        # There can still be some overflow, but let's just consider this "strong noise"
-        sim_img[sim_img > mx] = mx
+        noise_max = max(mx // 100, 10)
+
+            # generate noise and add directly
+        noise = numpy.random.randint(0, noise_max, sim_img.shape, dtype=self._img.dtype)
+
+        # final clamp to prevent overflow
+        dtype_max = numpy.iinfo(sim_img.dtype).max
```
pieleric: You need the same as in "synthetic":

```suggestion
        dt_info = numpy.iinfo(sim_img.dtype) if numpy.issubdtype(sim_img.dtype, numpy.integer) else numpy.finfo(sim_img.dtype)
        sim_img += numpy.minimum(dt_info.max - sim_img, noise)
```

pieleric (changes requested):

pieleric (approved):

PR #3395, opened 2026-03-04
[feat] add peak-finding algorithm

yxkdejong:


src/odemis/acq/align/goffset.py
```diff
@@ -0,0 +1,199 @@
...
+    f.task_canceller = _CancelSparcAutoGratingOffset
+
+    executeAsyncTask(f, _DoSparcAutoGratingOffset, args=(f, spgr, detector, align_grating, tolerance_px, max_it, gain))
+
+    return f
+
+def _DoSparcAutoGratingOffset(future: model.ProgressiveFuture,
+                              spgr: model.Actuator,
+                              detector: model.Detector,
+                              align_grating: bool,
```
pieleric: `align_grating` is not used anywhere. That actually makes sense as the selection between grating/detector is done at the driver level. => Remove this argument.

pieleric (changes requested):

tepals (changes requested):

src/odemis/acq/align/test/goffset_test.py
```diff
@@ -479,6 +479,70 @@ def test_peak_present_image(self):
...
+        fake_align_future = Future()
+        fake_align_future.set_result(True)
+        mock_sparc_offset.return_value = fake_align_future
+
+        # set up fake hardware
+        fake_spectrograph = MagicMock()
+
+        # Mock the initial position and the grating choices
+        fake_spectrograph.position.value = {"wavelength": 0, "grating": "grating1"}
+        fake_spectrograph.axes = {"grating": MagicMock(choices={"grating1": 0, "grating2": 1})}
```
pieleric: gratings are defined the other way around: int -> str

```
        fake_spectrograph.position.value = {"wavelength": 0, "grating": 0}
        fake_spectrograph.axes = {"grating": MagicMock(choices={0: "grating1", 1: "grating2"})}
```

src/odemis/acq/align/test/goffset_test.py
```diff
@@ -479,6 +479,70 @@ def test_peak_present_image(self):
...
+        fake_opm.setPath.assert_called_once_with("spec-focus", detector=fake_detector)
+
+        # did it try to align both gratings we mocked?
+        self.assertEqual(mock_sparc_offset.call_count, 2)
+
+        # did it clean up and turn the light off in the finally block?
+        self.assertEqual(fake_bl.power.value, 0)
+
+        # check the final returned results dict
+        self.assertEqual(results, {("grating1", "fake_ccd"): True, ("grating2", "fake_ccd"): True})
```
pieleric: If you change the gratings definition, this probably has to be updated too.

src/odemis/acq/align/goffset.py
```diff
@@ -658,6 +672,37 @@ def is_current_detector(d):
...
+            logging.warning("Brightlight did not confirm ON within 60s; continuing anyway")
+
+        _checkCancelled(future)
+
+        logging.info("Setting optical path to alignment mode: %s",align_mode)
+        future._subfuture = opm.setPath(align_mode,detector=first_detector)
+        future._subfuture.result()
+
+        _checkCancelled(future)
+        future._subfuture = InstantaneousFuture()
```
pieleric: No need to reset the subfuture. The previous one is fine (it's now in a "completed" state, which is the same as an InstantaneousFuture). => drop this line.

src/odemis/acq/align/test/goffset_alignment_test.py
```diff
@@ -0,0 +1,131 @@
...
+    # def test_cancel(self):
+    #     f = auto_align_grating_detector_offsets(spectrograph=self.spgr, detectors=[self.ccd],)
+    #     time.sleep(1)
+    #
+    #     cancelled = f.cancel()
+    #     self.assertTrue(cancelled)
+    #     self.assertTrue(f.cancelled())
+    #
+    #     with self.assertRaises(CancelledError):
+    #         f.result(timeout=900)
```
pieleric: Drop? There is another one similar just below.

src/odemis/acq/align/test/goffset_test.py
```diff
@@ -0,0 +1,561 @@
...
+                                     sparc_auto_grating_offset, auto_align_grating_detector_offsets,
+                                     _do_auto_align_grating_detector_offsets, log_detector_state)
+from odemis.dataio import hdf5
+from odemis.model import ProgressiveFuture
+from unittest.mock import patch, MagicMock
+
+logging.getLogger().setLevel(logging.DEBUG)
+
+HOME_PATH = os.path.expanduser("~") + "/"
+H5_FILE_2d_NO_PEAK = HOME_PATH + "development/odemis/grating 2 1024x256/"
```
pieleric: You're code shouldn't rely on files which are not in the source code (or in dependencies). How big are these files? If they are below 2Mb, you should add them to Odemis. Place them directly inside the same folder as this (test) file.
Then, make sure to compute the path as relative to odemis (ex: `DATA_DIR = Path(__file__).resolve().parent`).


Also, make sure the constant name is completely upper case (2d -> 2D).

src/odemis/acq/align/test/goffset_test.py
```diff
@@ -0,0 +1,561 @@
...
+        cls._original_goffset = cls.spgr.goffset
+        cls._original_position = cls.spgr.position.value.copy()
+
+    # @classmethod
+    # def tearDownClass(cls):
+    #     # restore original position
+    #     try:
+    #         cls.spgr.moveAbsSync(cls._original_position)
+    #     except Exception:
+    #         logging.exception("Failed restoring spectrograph position")
```
pieleric: Uncomment? Or remove?

src/odemis/acq/align/test/goffset_test.py
```diff
@@ -0,0 +1,561 @@
...
+    @classmethod
+    def setUpClass(cls):
+
+        cls.detector = model.getComponent(role="ccd")
+        cls.spgr = model.getComponent(role="spectrograph")
+        # cls.spccd = model.getComponent(role="sp-ccd")
+        cls.selector = model.getComponent(role="spec-det-selector")
+
+        cls._original_goffset = cls.spgr.goffset
+        cls._original_position = cls.spgr.position.value.copy()
```
pieleric, line 55: To run, this needs a backend. So it should start a (simulator) backend.
Probably something likes this would work:
```
SPARC2_CONFIG = CONFIG_PATH + "sim/sparc2-sim.odm.yaml"
:
testing.start_backend(SPARC2_CONFIG)
```

They are many backends. Maybe you need a specific one, just pick whichever you like!

src/odemis/acq/align/goffset.py
```diff
@@ -0,0 +1,793 @@
...
+    :return: estimated peak position in pixels (float)
+    :raises RuntimeError: if no significant peak is detected (SNR too low)
+    """
+
+    raw_data = numpy.asarray(data, dtype=float)
+    spectrum = numpy.squeeze(raw_data)
+
+    # maximum intensity projection
+    if spectrum.ndim > 1:
+        spectrum = spectrum.max(axis=0)
```
pieleric, line 61: Can you add a comment explaining why using MIP instead of the sum (or average)? I would have expected that the sum works better (ie, less noise), how come it doesn't?

src/odemis/acq/align/goffset.py
```diff
@@ -0,0 +1,793 @@
...
+        return False
+
+    mean = numpy.sum(x * w) / numpy.sum(w)
+    var = numpy.sum(w * (x - mean) ** 2) / numpy.sum(w)
+    width = numpy.sqrt(var)
+
+    snr = (peak_value - numpy.median(smoothed_data)) / (noise_std + 1e-6)
+    present = width_range[0] <= width <= width_range[1]
+
+    logging.debug("snr=%.2f width=%.2f present=%s", snr, width, present)
```
pieleric: Let's drop such logs, as it would fill up quite a bit the Odemis log, and I don't think it would help us much in case of issue.

src/odemis/acq/align/goffset.py
```diff
@@ -0,0 +1,793 @@
...
+        # avoid duplicate moves due to symmetry or float rounding
+        key = round(g, 3)
+        if key in tried:
+            continue
+        tried.add(key)
+
+        logging.debug("Coarse scan move attempt to goffset %.3f", g)
+        try:
+            spgr.moveAbsSync({"goffset": g})
+            time.sleep(2) # give hardware time to stabilize
```
pieleric: If that's required, it should go to the spectrograph driver. After the move, the hardware is supposed to be ready!

src/odemis/acq/align/goffset.py
```diff
@@ -0,0 +1,793 @@
...
+
+    # return back to start
+    spgr.moveRelSync({"goffset": -actual_delta})
+
+    # calculate goffset scale
+    scale = (p1 - p0) / actual_delta
+
+    logging.info(
+        "SCALE TRACKING | p0: %.1f | p1: %.1f | Delta: %.1f | Shift: %.1f | Result Scale: %.4f",
+        p0, p1, actual_delta, (p1 - p0), scale)
```
pieleric, line 288: Please write the text using standard upper & lower case. Otherwise, looking at the log will feel your code is shouting 😉

src/odemis/acq/align/goffset.py
```diff
@@ -0,0 +1,793 @@
...
+        # align each detector for the first grating
+        for d in detectors_sorted:
+            _checkCancelled(future)
+            logging.info("Starting alignment | Detector: %s | Grating: %s", d.name, g0)
+
+            if selector:
+                selector.moveAbsSync({selector_axes: detector_to_selector[d]})
+                time.sleep(stabilization_time)
+
+            gui_data = d.data.get(asap=False)
```
pieleric: "gui" seems a  little odd, as it comes from the CCD. Let's call this generically "ccd_data" or "det_data".

src/odemis/acq/align/goffset.py
```diff
@@ -0,0 +1,793 @@
...
+        future._subfuture.result()
+
+        _checkCancelled(future)
+        future._subfuture = InstantaneousFuture()
+
+        g0 = gratings[0]
+        logging.info("Starting alignment for initial grating: %s", g0)
+
+        spectrograph.moveAbsSync({"grating": g0, "wavelength": 0})
+        time.sleep(stabilization_time)
```
pieleric: This should go into the andorshrk driver, if stabilization is needed.
I'm very surprised that you need 10s... do you have any idea what takes so much time? My expectation would be that if something is still moving, it takes < 0.1s, or at most the duration of a CCD frame.

src/odemis/acq/align/goffset.py
```diff
@@ -0,0 +1,793 @@
...
+
+        if retries > 0:
+            return estimate_goffset_scale(spgr, detector, delta, retries - 1)
+
+        logging.warning("Scale estimation failed after retries, using default 0.5")
+        scale = 0.5
+
+    return scale, p0, p1
+
+def log_detector_state(caller: str, stage: str, detector: model.Detector, data: numpy.ndarray):
```
pieleric: Is that log still needed? It seems to be used only once in the code. If it's still useful, please at least lower the level  to INFO or DEBUG. (WARNING is used for something that is gone wrong and was recovered)

src/odemis/acq/align/goffset.py
```diff
@@ -0,0 +1,793 @@
...
+        logging.info("Preparing optical path and turning on light")
+
+        # make sure it's in 0th order (ie, show the image as-is), this also ensures the spectrograph
+        # is done with any previous actions.
+        spectrograph.moveAbsSync({"wavelength": 0})
+
+        # logging.info("Forcing brightlight ON") # bypassing sensor check for simulation
+        # bl.power.value = bl.power.range[1]
+        # time.sleep(1)
+
```
pieleric: In general, drop commented code.

src/odemis/acq/align/goffset_alignment.py
pieleric, line 1: Is that file an older version of `goffset`? If so, delete it.

src/odemis/acq/align/test/goffset_alignment_test.py
pieleric, line 1: Old version of goffset_test? Or does it contain different tests? If so, I think you should just merge the code into goffset_test.py

src/odemis/acq/align/goffset.py
```diff
@@ -0,0 +1,793 @@
...
+def _do_auto_align_grating_detector_offsets(future: model.ProgressiveFuture,
+                                            spectrograph: model.Actuator,
+                                            detectors: List[model.Detector],
+                                            opm: 'OpticalPathManager',
+                                            align_mode: str,
+                                            bl: model.Emitter,
+                                            selector: Optional[model.Actuator],
+                                            streams: List['Stream'],
+                                            stabilization_time: float = 10.0) -> Optional[Dict[Any, Any]]:
```
pieleric: I don't think this parameter is ever used, and the docstring reports a different default value.
Hopefully you can completely get rid of it by fixing the driver. If not, you should make this a constant (eg, `SPECTROGRAPH_STABILIZATION_TIME`)

src/odemis/gui/cont/tabs/sparc2_align_tab.py
```diff
@@ -1538,6 +1548,130 @@ def _on_align_mode_done(self, f):
...
+            return
+
+        # Reset progress bar
+        self.panel.gauge_auto_calibrate.SetValue(0)
+        self.panel.btn_auto_calibrate.SetLabel("Cancel")
+
+        wx.CallAfter(self._start_auto_calibration)
+
+        if not self._progress_timer.IsRunning():
+            self._progress_timer.Start(200)
```
pieleric: Please explain what 200 means and where it comes from. I guess it's in seconds? Is that just from experience, it takes about 200s?

src/odemis/acq/align/goffset.py
```diff
@@ -0,0 +1,793 @@
...
+    if len(detectors) > 1 and selector is None:
+        raise ValueError("No selector provided, but multiple detectors")
+
+    if streams is None:
+        streams = []
+
+    est_start = time.time() + 0.1
+    n_gratings = len(spectrograph.axes["grating"].choices)
+    n_detectors = len(detectors)
+    a_time = (_total_alignment_time(n_gratings, n_detectors) + 10)# estimated time to turn on light and close slit
```
pieleric: No need for parenthesis.

src/odemis/acq/align/goffset.py
```diff
@@ -0,0 +1,793 @@
...
+            gui_data = d.data.get(asap=False)
+            log_detector_state("GUI", "INITIAL", d, gui_data)
+
+            future._subfuture = sparc_auto_grating_offset(spectrograph, d)
+            success = future._subfuture.result()
+            results[(g0, d.name)] = success
+
+            # update progress
+            current_step += 1
+            future._progress = current_step / total_steps
```
pieleric: No, that's not how progress is supposed to be passed with `ProgressiveFuture`. 
You should convert this value to time left (one simple option is to  do  something like `(duration_until_now / current_step) * steps_left), and then pass it to the future: `future.set_progress(end=time.time() + time_left)`

You already compute the estimated total duration at init, so there is nothing else to do.

src/odemis/gui/cont/tabs/sparc2_align_tab.py
```diff
@@ -865,6 +865,16 @@ def add_axis(axisname, comp, label=None):
...
 
+        # Auto-calibration button
+        self.panel.btn_auto_calibrate.Bind(wx.EVT_BUTTON, self._on_btn_auto_calibrate)
+
+        # Auto-calibration state
+        self._auto_calibrate_future = model.InstantaneousFuture()
+
+        # Create progress timer ONCE
+        self._progress_timer = wx.Timer(self.panel)
+        self.panel.Bind(wx.EVT_TIMER, self._on_progress_timer, self._progress_timer)
```
pieleric: Please don't create your own timer and progress estimation. Instead, you can use the ProgressiveFuture, as described above, and then use a `ProgressiveFutureConnector(self._auto_calibrate_future, gauge)` when starting the alignment. You can check the autofocus code for example.

src/odemis/gui/cont/tabs/sparc2_align_tab.py
```diff
@@ -1538,6 +1548,130 @@ def _on_align_mode_done(self, f):
         else:
             logging.debug("Optical path was updated.")
 
+
+    # Auto-Calibration
+    def _on_btn_auto_calibrate(self, evt):
+
+        # Check if there's a process running, if so, cancel and reset
+        if hasattr(self, "_auto_calibrate_future") and not self._auto_calibrate_future.done():
```
pieleric: `_auto_calibrate_future` is defined at init, so no need for defensive code check for it.

src/odemis/gui/cont/tabs/sparc2_align_tab.py
```diff
@@ -1538,6 +1548,130 @@ def _on_align_mode_done(self, f):
...
+        # Reset progress bar
+        self.panel.gauge_auto_calibrate.SetValue(0)
+        self.panel.btn_auto_calibrate.SetLabel("Cancel")
+
+        wx.CallAfter(self._start_auto_calibration)
+
+        if not self._progress_timer.IsRunning():
+            self._progress_timer.Start(200)
+
+    def _start_auto_calibration(self):
```
pieleric: "auto calibration" is too generic. Let's rename this to something like "grating_calibration" of "goffset_calibration". Please update the function name and the log text too.

src/odemis/gui/cont/tabs/sparc2_align_tab.py
```diff
@@ -1538,6 +1548,130 @@ def _on_align_mode_done(self, f):
...
+        # Smooth UI trick:
+        # If we are behind the target, catch up quickly.
+        if current_val < target_val:
+            self.panel.gauge_auto_calibrate.SetValue(current_val + 1)
+
+        # If we are caught up, STILL allow the bar to creep up to 5% ahead of the target.
+        # This ensures the bar instantly starts moving to ~4% while waiting
+        # for the hardware's 10-second stabilization time.
+        elif current_val < (target_val + 4) and current_val < 95:
+            self.panel.gauge_auto_calibrate.SetValue(current_val + 1)
```
pieleric: If you use a `ProgressiveFutureConnector()`, this whole function can go away.

src/odemis/gui/cont/tabs/sparc2_align_tab.py
```diff
@@ -865,6 +865,16 @@ def add_axis(axisname, comp, label=None):
         # TODO: Auto remove the background when the image shape changes?
         # TODO: Use a toggle button to show the background is in use or not?
 
+        # Auto-calibration button
+        self.panel.btn_auto_calibrate.Bind(wx.EVT_BUTTON, self._on_btn_auto_calibrate)
```
pieleric: This button doesn't exist. I think you've modified the .xrc file but didn't add it to the commit.
BTW, be aware that you cannot directly modify main_xrc.py . You have to modify `panel_tab_sparc2_align.xrc` and then run `./src/odemis/gui/xmlh/generate_xrc.py` . See the developer manual for more info.

pieleric (changes requested):

src/odemis/gui/cont/tabs/sparc2_align_tab.py
yxkdejong: I think Nandish did this for me.. but I did indeed forget to add it to the commit

src/odemis/acq/align/goffset.py
yxkdejong: I don't think it's necessary anymore, I used this to see what was going on with the GUI vs the acquisition in odemis

src/odemis/acq/align/test/goffset_test.py
```diff
@@ -0,0 +1,495 @@
...
+
+        # pass the cleaned 1D array to the existing function
+        present = peak_is_present(spectrum, snr_threshold=10.0, width_range=(0.5, 12.0))
+
+        self.assertTrue(present, "Peak should have been detected in the cleaned spectrum.")
+
+    @patch('odemis.acq.align.goffset.sparc_auto_grating_offset')
+    @patch('odemis.acq.align.goffset.light.turnOnLight')
+    @patch('odemis.acq.align.goffset.time.sleep')
+
```
pieleric: Drop the empty line

src/odemis/acq/align/test/goffset_test.py
```diff
@@ -0,0 +1,495 @@
...
+        logging.info(f"Cleaned Spectrum Shape for Function: {spectrum.shape}")
+
+        # pass the cleaned 1D array to the existing function
+        present = peak_is_present(spectrum, snr_threshold=10.0, width_range=(0.5, 12.0))
+
+        self.assertTrue(present, "Peak should have been detected in the cleaned spectrum.")
+
+    @patch('odemis.acq.align.goffset.sparc_auto_grating_offset')
+    @patch('odemis.acq.align.goffset.light.turnOnLight')
+    @patch('odemis.acq.align.goffset.time.sleep')
```
pieleric: Does time.sleep() still need to be mocked? Was it to accelerate the test when there was extra time for waiting for the spectrograph?

src/odemis/gui/cont/tabs/sparc2_align_tab.py
```diff
@@ -865,6 +866,15 @@ def add_axis(axisname, comp, label=None):
         # TODO: Auto remove the background when the image shape changes?
         # TODO: Use a toggle button to show the background is in use or not?
 
+        # Auto-calibration button
+        self.panel.btn_auto_calibrate.Bind(wx.EVT_BUTTON, self._on_btn_grating_calibration)
+
+        # Auto-calibration state
+        self._grating_calibration_future = model.InstantaneousFuture()
+
+        # Create progress timer
```
pieleric: That comment doesn't really fit. I'd suggest "To hold the progressive future connector during calibration".

src/odemis/gui/xmlh/resources/panel_tab_sparc2_align.xrc
```diff
@@ -401,6 +401,71 @@
               <flag>wxBOTTOM|wxEXPAND</flag>
               <border>5</border>
             </object>
+            <object class="sizeritem">
+              <object class="wxPanel" name="pnl_calibrate">
+                <object class="wxBoxSizer">
+                  <orient>wxVERTICAL</orient>
+                  <object class="sizeritem">
+                    <object class="wxStaticText">
+                      <label>Calibrate</label>
```
pieleric: This is too generic. The entire tab is about alignment and calibration! Maybe "Grating" + button = "Auto centre"?  Or "Grating centre" + button="Auto calibration"/"Auto offset"? You could ask Noémie what she prefers.

src/odemis/gui/xmlh/resources/panel_tab_sparc2_align.xrc
```diff
@@ -401,6 +401,71 @@
...
+                        <sysfont>wxSYS_DEFAULT_GUI_FONT</sysfont>
+                      </font>
+                    </object>
+                    <flag>wxALL</flag>
+                    <border>5</border>
+                  </object>
+                  <object class="sizeritem">
+                    <object class="wxGridBagSizer">
+                      <object class="sizeritem">
+                        <object class="ImageTextButton" name="btn_auto_calibrate">
```
pieleric: It'd be clearer for the code to also rename the button to something like `btn_auto_grating_offset` or `btn_auto_grating_centre`.

src/odemis/acq/align/goffset.py
```diff
@@ -0,0 +1,757 @@
...
+    :param spgr: spectrograph
+    :param detector: detector
+    :param step: step size in goffset units for the coarse scan (default: 2000)
+    :return: tuple (actual_goffset, peak_pixel)
+    :raises RuntimeError: if no peak is found across the full goffset range
+    """
+
+    current = float(spgr.position.value["goffset"])
+    step = abs(step) if step != 0 else 2000.0
+    max_span = 200000.0  # limit how far we wander from the current valid position
```
pieleric: It might be more future-proof (in case we support other types of spectrographs) to compute these values based on the axis range. See `spgr.axes["goffset"].range`.

src/odemis/acq/align/goffset.py
```diff
@@ -0,0 +1,757 @@
...
+        return results
+
+    except CancelledError:
+        logging.info("Auto-alignment cancelled")
+        raise
+
+    finally:
+        logging.info("Turning off brightlight")
+        try:
+            bl.power.value = (bl.power.range[0])
```
pieleric: Drop the parenthesis.

src/odemis/acq/align/goffset.py
```diff
@@ -0,0 +1,757 @@
...
+        try:
+            future._subfuture.result(timeout=60)
+
+        except TimeoutError:
+            future._subfuture.cancel()
+            logging.warning("Brightlight did not confirm ON within 60s; continuing anyway")
+
+        except Exception as e:
+            # This catches any other errors the hardware might throw
+            logging.warning(f"Brightlight encountered an error: {e}. Continuing anyway.")
```
pieleric: Please don't. That could be all sort of errors. Also, this catches the CancelledError if the user cancels, which is confusing.

pieleric (changes requested):

src/odemis/acq/align/goffset.py
tepals: agreed

tepals (approved):

src/odemis/acq/align/goffset.py
```diff
@@ -189,12 +189,12 @@ def coarse_scan_goffset_for_peak(spgr, detector, future: model.ProgressiveFuture
     """
 
     current = float(spgr.position.value["goffset"])
-    step = abs(step) if step != 0 else 2000.0
-    max_span = 200000.0  # limit how far we wander from the current valid position
+    goffset_min, goffset_max = spgr.axes["goffset"].range
+    max_span = goffset_max - goffset_min  # Search far enough around the current position to cover the full axis range.
+    step = abs(step) if step != 0 else max_span / 100.0
```
pieleric: Step is now `None` by default, so this doesn't work. I'd just write:
```
if step is None:
     step = max_span / 100.0
```

src/odemis/gui/xmlh/resources/panel_tab_sparc2_align.xrc
```diff
@@ -400,6 +400,71 @@
...
+                <object class="wxBoxSizer">
+                  <orient>wxVERTICAL</orient>
+                  <object class="sizeritem">
+                    <object class="wxStaticText">
+                      <label>Calibrate</label>
+                      <font>
+                        <size>16</size>
+                        <sysfont>wxSYS_DEFAULT_GUI_FONT</sysfont>
+                      </font>
+                    </object>
```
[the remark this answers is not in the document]
pieleric: @yxkdejong that comment seems valid. Please add foreground too.

pieleric (changes requested):

pieleric (approved):

PR #3403, opened 2026-03-17
[refactor] add copyrights on files without one

pieleric:
Also rename new test case file which didn't finish with _test.py .


tepals (approved):
Looks good, just one comment about keeping a TODO in [src/odemis/gui/test/comp_canvas_cairo_test.py‎](https://github.com/delmic/odemis/pull/3403/changes#diff-27a9aab20e3874935354aeb78d8dadb796bc83d582f982b0ea836279909a4186), not sure if it really makes sense to keep a 12 year old TODO 😁

PR #3407, opened 2026-03-17
[MSD-350][feat] Ensure pyramidal tiff on import

tmoerkerken:
Support for on-the-fly conversion to pyramidal files when loading a file in tabs. 

Note: this PR changed quite a lot since the first version

## Changes
- Add mixin for file loading, overwriting the duplicate code in existing tabs
- Conditionally convert based on extension and size
- Converted files always end up next to source
- Add test suite that covers most scenarios

### DIFF
--- a/src/odemis/dataio/tiff.py
+++ b/src/odemis/dataio/tiff.py
@@ -2401,6 +2401,25 @@ def read_data(filename):
     return [acd.content[n].getData() for n in range(len(acd.content))]
 
 
+def is_pyramidal(filename: str) -> bool:
+    """
+    Check whether a TIFF file has SubIFD pyramid levels.
+
+    :param filename: Path to TIFF file
+    :return: True when TIFFTAG_SUBIFD contains one or more entries
+    :raises IOError: If the file cannot be opened as TIFF
+    """
+    try:
+        tiff_file = TIFF.open(filename, mode="r")
+        try:
+            sub_ifds = tiff_file.GetField(T.TIFFTAG_SUBIFD)
+            return sub_ifds is not None and len(sub_ifds) > 0
+        finally:
+            tiff_file.close()
+    except Exception as exc:
+        raise IOError("Failed to inspect TIFF pyramid status for %s: %s" % (filename, exc))
+
+
 def read_thumbnail(filename):
     """
     Read the thumbnail data of a given TIFF file.
--- a/src/odemis/dataio/tiff.py
+++ b/src/odemis/dataio/tiff.py
@@ -2363,6 +2363,7 @@ def export(
     :param pyramid: whether to export data as pyramid
     :param imagej: save the metadata in a format compatible with ImageJ
     """
+    filename = str(filename)
     if not isinstance(data, list):
         data = [data]
 
--- a/src/odemis/gui/cont/tabs/_constants.py
+++ b/src/odemis/gui/cont/tabs/_constants.py
@@ -34,3 +34,5 @@ MIRROR_NOT_REFD = 0
 MIRROR_PARKED = 1
 MIRROR_BAD = 2  # not parked, but not fully engaged either
 MIRROR_ENGAGED = 3
+
+PYRAMIDAL_CONVERSION_SUFFIX = "_pyramidal"
--- a/src/odemis/gui/cont/tabs/analysis_tab.py
+++ b/src/odemis/gui/cont/tabs/analysis_tab.py
@@ -27,6 +27,8 @@ import collections
 import gc
 import logging
 import os.path
+from pathlib import Path
+from typing import Union
 import wx
 # IMPORTANT: wx.html needs to be imported for the HTMLWindow defined in the XRC
 # file to be correctly identified. See: http://trac.wxwidgets.org/ticket/3626
@@ -55,11 +57,19 @@ from odemis.gui.comp.viewport import MicroscopeViewport, AngularResolvedViewport
     AngularSpectrumViewport, ThetaViewport
 from odemis.gui.conf import get_acqui_conf
 from odemis.gui.cont import settings
+from odemis.gui.cont.tabs import PYRAMIDAL_CONVERSION_SUFFIX
 from odemis.gui.cont.tabs.tab import Tab
 from odemis.gui.model import TOOL_POINT, TOOL_LINE, TOOL_ACT_ZOOM_FIT, TOOL_RULER, TOOL_LABEL, \
     TOOL_NONE
 from odemis.gui.util import call_in_wx_main
-from odemis.util.dataio import data_to_static_streams, open_acquisition, open_files_and_stitch
+from odemis.util.dataio import (
+    data_to_static_streams,
+    export_data_as_pyramidal,
+    open_acquisition,
+    open_files_and_stitch,
+    is_pyramidal,
+    convert_file_to_pyramidal,
+)
 
 
 class AnalysisTab(Tab):
@@ -93,6 +103,9 @@ class AnalysisTab(Tab):
         # displayed
         tab_data = guimod.AnalysisGUIData(main_data)
         super(AnalysisTab, self).__init__(name, button, panel, main_frame, tab_data)
+
+        self.main_data = main_data
+
         if main_data.role in ("sparc-simplex", "sparc", "sparc2"):
             # Different name on the SPARC to reflect the slightly different usage
             self.set_label("ANALYSIS")
@@ -318,13 +331,52 @@ class AnalysisTab(Tab):
     def _on_add_tileset(self):
         self.select_acq_file(extend=True, tileset=True)
 
+    def _get_pyramidal_path(self, filename: Union[str, Path]):
+        """
+        Returns the output path for a given input file, used in the pyramidal conversion process.
+
+        :param filename: The input filename of the file that is to be converted
+        :return: The adjusted output path
+        """
+        filename = Path(filename)
+        name_converted = f"{filename.stem}{PYRAMIDAL_CONVERSION_SUFFIX}.tif"
+
+        project_folder = (
+            None
+            if getattr(self.main_data, "is_viewer", False)
+            else self.main_data.project_path.value
+        )
+
+        if project_folder:
+            return Path(project_folder) / name_converted
+        else:
+            # For simplicity, always convert to tiff format
+            return filename.with_name(name_converted)
+
     def load_tileset(self, filenames, extend=False):
-        data = open_files_and_stitch(filenames) # TODO: allow user defined registration / weave methods
-        self.display_new_data(filenames[0], data, extend=extend)
+        data = open_files_and_stitch(filenames)  # TODO: allow user defined registration / weave methods
+        # For the lack of a better name, use the name of the first tile
+        source_filename = filenames[0]
+        filename_converted = self._get_pyramidal_path(source_filename)
+        try:
+            export_data_as_pyramidal(data, filename_converted, compressed=True)
+            self.display_new_data(filename_converted, open_acquisition(filename_converted), extend=extend)
+        except Exception:
+            logging.exception("Failed to export/reopen stitched pyramidal tileset")
+            self.display_new_data(filenames[0], data, extend=extend)
 
     def load_data(self, filename, fmt=None, extend=False):
-        data = open_acquisition(filename, fmt)
-        self.display_new_data(filename, data, extend=extend)
+        filename = Path(filename)
+        if is_pyramidal(filename):
+            self.display_new_data(filename, open_acquisition(filename, fmt), extend=extend)
+            return
+        filename_converted = self._get_pyramidal_path(filename)
+        try:
+            convert_file_to_pyramidal(filename, filename_converted, compressed=True)
+            self.display_new_data(filename_converted, open_acquisition(filename_converted, fmt), extend=extend)
+        except Exception as ex:
+            logging.exception("Failed to process TIFF import, loading original. Error: %s", ex)
+            self.display_new_data(filename, open_acquisition(filename, fmt), extend=extend)
 
     def _get_time_spectrum_streams(self, spec_streams):
         """
@@ -358,6 +410,7 @@ class AnalysisTab(Tab):
         extend (bool): if False, will ensure that the previous streams are closed.
           If True, will add the new file to the current streams opened.
         """
+        filename = str(filename)
         if not extend:
             # Remove all the previous streams
             self._stream_bar_controller.clear()
--- a/src/odemis/gui/cont/tabs/analysis_tab.py
+++ b/src/odemis/gui/cont/tabs/analysis_tab.py
@@ -27,6 +27,8 @@ import collections
 import gc
 import logging
 import os.path
+from typing import List, Union
+
 import wx
 # IMPORTANT: wx.html needs to be imported for the HTMLWindow defined in the XRC
 # file to be correctly identified. See: http://trac.wxwidgets.org/ticket/3626
@@ -55,6 +57,7 @@ from odemis.gui.comp.viewport import MicroscopeViewport, AngularResolvedViewport
     AngularSpectrumViewport, ThetaViewport
 from odemis.gui.conf import get_acqui_conf
 from odemis.gui.cont import settings
+from odemis.gui.cont.tabs.stream_load_mixin import StreamLoadMixin
 from odemis.gui.cont.tabs.tab import Tab
 from odemis.gui.model import TOOL_POINT, TOOL_LINE, TOOL_ACT_ZOOM_FIT, TOOL_RULER, TOOL_LABEL, \
     TOOL_NONE
@@ -62,7 +65,7 @@ from odemis.gui.util import call_in_wx_main
 from odemis.util.dataio import data_to_static_streams, open_acquisition, open_files_and_stitch
 
 
-class AnalysisTab(Tab):
+class AnalysisTab(StreamLoadMixin, Tab):
     """ Handle the loading and displaying of acquisition files
     """
     def __init__(self, name, button, panel, main_frame, main_data):
@@ -92,7 +95,7 @@ class AnalysisTab(Tab):
         # TODO: automatically change the display type based on the acquisition
         # displayed
         tab_data = guimod.AnalysisGUIData(main_data)
-        super(AnalysisTab, self).__init__(name, button, panel, main_frame, tab_data)
+        super().__init__(name, button, panel, main_frame, tab_data)
         if main_data.role in ("sparc-simplex", "sparc", "sparc2"):
             # Different name on the SPARC to reflect the slightly different usage
             self.set_label("ANALYSIS")
@@ -318,14 +321,6 @@ class AnalysisTab(Tab):
     def _on_add_tileset(self):
         self.select_acq_file(extend=True, tileset=True)
 
-    def load_tileset(self, filenames, extend=False):
-        data = open_files_and_stitch(filenames) # TODO: allow user defined registration / weave methods
-        self.display_new_data(filenames[0], data, extend=extend)
-
-    def load_data(self, filename, fmt=None, extend=False):
-        data = open_acquisition(filename, fmt)
-        self.display_new_data(filename, data, extend=extend)
-
     def _get_time_spectrum_streams(self, spec_streams):
         """
         Sort spectrum streams into the substreams according to types spectrum, temporal spectrum,
@@ -345,6 +340,17 @@ class AnalysisTab(Tab):
         # are considered temporal spectrum streams -> adapt StaticSpectrumStream
 
         return spectrum, temporalspectrum, timecorrelator, angularspectrum
+
+    @call_in_wx_main
+    def load_streams(self, das: List[Union[model.DataArray, model.DataArrayShadow]], filename, extend):
+        """
+        Method to streamline the loading/display of streams with other tabs, without breaking existing code.
+        :param das: the data to load
+        :param filename: name of the file containing the data, for information display purposes.
+        :param extend: if False, will ensure that the previous streams are closed.
+          If True, will add the new file to the current streams opened.
+        """
+        self.display_new_data(filename, das, extend)
 
     @call_in_wx_main
     def display_new_data(self, filename, data, extend=False):
--- a/src/odemis/gui/cont/tabs/correlation_tab.py
+++ b/src/odemis/gui/cont/tabs/correlation_tab.py
@@ -25,26 +25,33 @@ Odemis. If not, see http://www.gnu.org/licenses/.
 
 import collections
 import logging
-import os.path
-from typing import List, Optional
+import os
+from pathlib import Path
+from typing import List, Optional, Union
 import wx
-
 from odemis.gui import conf
-
 from odemis import dataio
 from odemis import model
 import odemis.gui
 import odemis.gui.cont.export as exportcont
-from odemis.gui.cont.stream_bar import StreamBarController
 import odemis.gui.cont.views as viewcont
 import odemis.gui.model as guimod
 import odemis.gui.util as guiutil
 from odemis.acq.stream import OpticalStream, EMStream, StaticStream
 from odemis.gui.cont.correlation import CorrelationController
+from odemis.gui.cont.stream_bar import StreamBarController
+from odemis.gui.cont.tabs import PYRAMIDAL_CONVERSION_SUFFIX
+from odemis.gui.cont.tabs.tab import Tab
 from odemis.gui.model import TOOL_ACT_ZOOM_FIT
 from odemis.gui.util import call_in_wx_main
-from odemis.gui.cont.tabs.tab import Tab
-from odemis.util.dataio import data_to_static_streams, open_acquisition, open_files_and_stitch
+from odemis.util.dataio import (
+    convert_file_to_pyramidal,
+    data_to_static_streams,
+    export_data_as_pyramidal,
+    is_pyramidal,
+    open_acquisition,
+    open_files_and_stitch,
+)
 
 
 class CorrelationTab(Tab):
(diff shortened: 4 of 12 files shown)


src/odemis/util/tiff_convert.py
```diff
@@ -0,0 +1,301 @@
...
+                progress_callback=progress_callback,
+            )
+        finally:
+            dlg.Destroy()
+    except IOError as exc:
+        logger.error("Failed to convert TIFF file %s: %s", filename, exc)
+        return filename
+
+
+def ensure_pyramidal_tiff_for_tileset_gui(filenames: list[str], parent: Any, main_data: Any) -> list[str]:
```
pieleric: The `Any` type(s) are not correct. You actually have specific expectations. I'd suggest to either explicitly import the correct types (but it might be annoying due to recursive dependencies it would create) or indicate the actual type as a string (less useful for the tools, but good for humans, and already better than `Any`)

src/odemis/util/tiff_convert.py
```diff
@@ -0,0 +1,301 @@
...
+
+You should have received a copy of the GNU General Public License along with
+Odemis. If not, see http://www.gnu.org/licenses/.
+"""
+
+import logging
+import os
+from typing import Any, Callable, Optional
+
+import wx
```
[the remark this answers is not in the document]
pieleric: @tmoerkerken I agree here that if this module relies on wx then it should not be here. I'd suggest to move it to odemis.gui.util. The alternative would be to have a more indirect way to report the progress: run the function in background, and return a ProgressiveFuture. Then, in the GUI side, create a progress dialog, and connect the ProgressiveFuture to the progress bar.

pieleric (changes requested):

src/odemis/gui/cont/tabs/_constants.py
```diff
@@ -34,3 +34,5 @@
 MIRROR_PARKED = 1
 MIRROR_BAD = 2  # not parked, but not fully engaged either
 MIRROR_ENGAGED = 3
+
+PYRAMIDAL_CONVERSION_SUFFIX = "_pyramidal"
```
pieleric: Please add a comment to indicate what this is used for (to give a hint on what changes/breaks if you change that constant).

src/odemis/gui/cont/tabs/correlation_tab.py
```diff
@@ -166,13 +173,53 @@ def _on_add_file(self) -> None:
...
+        try:
+            export_data_as_pyramidal(data, filename_converted, compressed=True)
+            self.load_streams(open_acquisition(filename_converted))
+        except Exception:
+            logging.exception("Failed to export/reopen stitched pyramidal tileset")
+            self.load_streams(data)
+
+
+    def load_data(self, filename: Union[str, Path], fmt: Optional[str] = None, extend: bool = False) -> None:
+        filename = Path(filename)
```
tepals: can you add a docstring here and to `load_tileset`? I don't know what fmt is

src/odemis/util/dataio.py
```diff
@@ -21,8 +21,9 @@
 """
 import json
 import logging
+from pathlib import Path
 import os
-from typing import Any, Optional
+from typing import Any, List, Optional, Union
```
tepals: Import guidelines are to have all the `import` statements first, then all the `from` statements.

src/odemis/util/dataio.py
```diff
@@ -225,6 +227,60 @@ def open_acquisition(filename, fmt=None):
     return data
 
 
+def is_pyramidal(filename: Union[str, Path]) -> bool:
+    """
+    Check whether a file is in pyramidal format.
+
+    :param filename: Path to file
+    :return: True if the file is pyramidal, False otherwise
+    :raises IOError: If the file cannot be read or converter doesn't support pyramid detection
```
tepals: the code on line 239 implies that this function will return `False` if the reader/converter does not support pyramids, or am I missing something?

src/odemis/gui/cont/tabs/correlation_tab.py
```diff
@@ -166,13 +173,53 @@ def _on_add_file(self) -> None:
     def _on_add_tileset(self) -> None:
         self.select_acq_file(extend=True, tileset=True)
 
-    def load_tileset(self, filenames: List[str], extend: bool = False) -> None:
-        data = open_files_and_stitch(filenames) # TODO: allow user defined registration / weave methods
-        self.load_streams(data)
+    def _get_pyramidal_path(self, filename: Union[str, Path]):
```
tepals: there are now 3 shared functions between the correlation and analysis tab, do you think there is any risk that someone will update one of these functions in one tab and forget to update it in the other?

tepals (changes requested):

src/odemis/gui/cont/tabs/correlation_tab.py
tmoerkerken: Yeah, I discussed that with Eric beforehand, since my intuition was to also make a reusable function. But there is minimal deviation in each function (in the stream visualization part), so that's why we kept it separate.

tepals: okay, didn't notice the difference, then it's fine

src/odemis/dataio/tiff.py
```diff
@@ -2401,6 +2401,25 @@ def read_data(filename):
...
+    Check whether a TIFF file has SubIFD pyramid levels.
+
+    :param filename: Path to TIFF file
+    :return: True when TIFFTAG_SUBIFD contains one or more entries
+    :raises IOError: If the file cannot be opened as TIFF
+    """
+    try:
+        tiff_file = TIFF.open(filename, mode="r")
+        try:
+            sub_ifds = tiff_file.GetField(T.TIFFTAG_SUBIFD)
```
pieleric: As I understand it, this checks only on the first data (aka page, aka directory). So, if the first image is for thumbnails, which we do in some cases, it will report it's not pyramidal, although the main image is. So you could go though each directory (with something like _iterDirectories()) and check the first one which is not a thumbnail (with _isThumbnail()).

src/odemis/gui/cont/tabs/analysis_tab.py
```diff
@@ -318,13 +331,52 @@ def _on_add_file(self):
     def _on_add_tileset(self):
         self.select_acq_file(extend=True, tileset=True)
 
+    def _get_pyramidal_path(self, filename: Union[str, Path]):
```
pieleric: Update type annotation to indicate the return type.

src/odemis/gui/cont/tabs/analysis_tab.py
```diff
@@ -318,13 +331,52 @@ def _on_add_file(self):
...
+        :return: The adjusted output path
+        """
+        filename = Path(filename)
+        name_converted = f"{filename.stem}{PYRAMIDAL_CONVERSION_SUFFIX}.tif"
+
+        project_folder = (
+            None
+            if getattr(self.main_data, "is_viewer", False)
+            else self.main_data.project_path.value
+        )
```
pieleric: Please use a normal `if/else` block, if you're going to use multiple lines anyway.

src/odemis/gui/cont/tabs/correlation_tab.py
pieleric: I do agree with Thera that this feels duplicated. I can't spot the difference, what is it? Is it possible to just have a single function that take an extra (optional?) argument so that a single function does both?

Otherwise, at least, add a comment somewhere that there is a (almost) identical function somewhere else, and what is the difference.

src/odemis/gui/cont/tabs/analysis_tab.py
```diff
@@ -318,13 +331,52 @@ def _on_add_file(self):
...
+            self.display_new_data(filename_converted, open_acquisition(filename_converted), extend=extend)
+        except Exception:
+            logging.exception("Failed to export/reopen stitched pyramidal tileset")
+            self.display_new_data(filenames[0], data, extend=extend)
 
     def load_data(self, filename, fmt=None, extend=False):
-        data = open_acquisition(filename, fmt)
-        self.display_new_data(filename, data, extend=extend)
+        filename = Path(filename)
+        if is_pyramidal(filename):
```
pieleric: There are several cases where it is not pyramidal and we don't want to convert it to pyramidal:
* The data is not TIFF (eg HDF5)
* The data cannot be converted to pyramidal format (eg, not 2D, such as Spectrum, AR...)
* The data is small (should be a constant defined somewhere, but anything with less than 2048² pixels is probably not worth the trouble)

Also is_pyramidal can raise exceptions. In such case, I'd suggest to not try harder, and open the file as-is too. 

Note that for now, as I read the code, if you load a HDF5 file, eventually, it will either fail, or output an exception and succeed. It should just succeed nicely.

BTW, maybe a different but more stream-lined approach could be to always call `open_acquisition(filename, fmt)`. And check the return data is DataArrayShadow with tiles, or not (ie, just DataArray, or DataArrayShadow without tiles). This way, you could get rid of `is_pyramidal` completely.

pieleric (changes requested):

tepals (approved):
approved, given that you fix the comments :)

src/odemis/gui/cont/tabs/test/stream_load_mixin_test.py
```diff
@@ -0,0 +1,117 @@
...
+    PYRAMIDAL_CONVERSION_SUFFIX,
+    StreamLoadMixin,
+)
+
+SMALL_SIDE = 10
+LARGE_SIDE = int(PYRAMIDAL_CONVERSION_MIN_RES ** 0.5) + 1
+
+
+class MockTab(StreamLoadMixin):
+    """Mocked tab to test the loading of data on"""
```
tepals: I think you forgot to finish this sentence

src/odemis/gui/cont/tabs/stream_load_mixin.py
```diff
@@ -0,0 +1,87 @@
...
+from odemis import dataio, model
+from odemis.dataio import tiff
+from odemis.util.dataio import open_files_and_stitch, open_acquisition
+from odemis.util.dataio import splitext
+
+PYRAMIDAL_CONVERSION_MIN_RES = 2048 ** 2  # Minimal resolution to trigger pyramidal conversion
+PYRAMIDAL_CONVERSION_SUFFIX = "_pyramidal"
+
+class StreamLoadMixin:
+    """Mixin for loading files in tabs. Note: it requires a load_streams method to be implemented on the parent"""
```
tepals, line 39: you could make `load_streams` an abstract method, or does that not work with mixin classes?

src/odemis/gui/cont/tabs/stream_load_mixin.py
```diff
@@ -0,0 +1,87 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def load_tileset(self, filenames, **kwargs):
```
tepals: can you add typing to the methods in this class?

tepals (approved):
some small comments, the rest looks good

src/odemis/gui/cont/tabs/analysis_tab.py
tmoerkerken: Tried to tackle your points. Only thing I'm not sure of is the second point, but I thought it was covered by the first.

src/odemis/dataio/tiff.py
tmoerkerken: removed it entirely

src/odemis/gui/cont/tabs/stream_load_mixin.py
tmoerkerken: Yes, it works, and it is a nice touch!

src/odemis/gui/cont/tabs/test/stream_load_mixin_test.py
tmoerkerken: Correct :)

src/odemis/gui/cont/tabs/analysis_tab.py
```diff
@@ -346,6 +339,10 @@ def _get_time_spectrum_streams(self, spec_streams):
 
         return spectrum, temporalspectrum, timecorrelator, angularspectrum
 
+    @call_in_wx_main
+    def load_streams(self, data, **kwargs):
+        self.display_new_data(kwargs["filename"], data, kwargs["extend"])
```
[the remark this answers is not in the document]
tmoerkerken: Hmm, seems outdated since I did that already

src/odemis/gui/cont/tabs/stream_load_mixin.py
```diff
@@ -0,0 +1,96 @@
...
+
+class StreamLoadMixin(ABC):
+    """Mixin for loading files in tabs. Note: it requires a load_streams method to be implemented on the parent"""
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    @abstractmethod
+    def load_streams(self, data: List[model.DataArray], **kwargs):
+        """
+        Placeholder for load_streams method on tab
```
pieleric: As an abstract method, it should describe what the method is used for, when it will be called, and all the parameters.

This `**kwargs` is really fishy. In practice, it seems you use it only with `filename` and `extend`, so it'd make the code much easier to understand if they were defined explicitly (as optional args).

src/odemis/gui/cont/tabs/stream_load_mixin.py
```diff
@@ -0,0 +1,96 @@
...
+        :param data: the data to load
+        """
+        pass
+
+    def load_tileset(self, filenames: List[os.PathLike], **kwargs):
+        """
+        Loads tileset, and converts it into pyramidal if fulfilling certain conditions
+        :param filenames: list of filenames to load
+        """
```
pieleric, line 61: Add a description for kwargs. Something like this would work: extra arguments to be passed to load_stream().

(same comment for the `load_data` function.

Also, please indicate the function returns nothing by adding a `-> None`.

src/odemis/gui/cont/tabs/stream_load_mixin.py
```diff
@@ -0,0 +1,96 @@
...
+        if data_format == tiff and _needs_conversion(data):
+            data, filename = _convert_to_pyramidal(data, filename)
+
+        self.load_streams(data, filename=filename, **kwargs)
+
+
+def _needs_conversion(data: List[model.DataArray]):
+    """
+    Check if not already pyramidal and check if resolution is large enough to validify the conversion.
+    # The 'maxzoom' property exists only when pyramidal, so it seems.
```
pieleric: This comment is an implementation detail, so it should be in the code, not in the docstring.

BTW, yet, for now it's defined this way. If a data has `.maxzoom`, then it has pyramidal support (aka `getTile`).

src/odemis/gui/cont/tabs/stream_load_mixin.py
```diff
@@ -0,0 +1,96 @@
...
+        data = open_acquisition(filename, fmt)
+        data_format = dataio.find_fittest_converter(filename, mode=os.O_RDONLY)
+        # Only potentially convert TIFFs for now, other data we just visualize as is
+        if data_format == tiff and _needs_conversion(data):
+            data, filename = _convert_to_pyramidal(data, filename)
+
+        self.load_streams(data, filename=filename, **kwargs)
+
+
+def _needs_conversion(data: List[model.DataArray]):
```
pieleric: "needs_conversion" is a little bit too generic name. I'd suggest to clarify that it's a conversion *to pyramidal format* (and not to HDF5, to RGB...). => something like `needs_conversion_to_pyramidal`

src/odemis/gui/cont/tabs/stream_load_mixin.py
```diff
@@ -0,0 +1,96 @@
...
+    :filename: filename to convert
+    :return: converted data and filename, or original data and filename if unsuccessful
+    """
+    try:
+        filename_converted = _get_converted_filename(filename)
+        # Try to get the raw data in two ways, due to the open_acquisition method having two loading paths
+        if hasattr(data[0], "data"):
+            raw_data = data[0]
+        elif hasattr(data[0], "getData"):
+            raw_data = data[0].getData()
```
pieleric: As coderabbit says, there are a couple of issues here. First, it only exports the first DataArray. Second, it can be simplified.
There are several other places in Odemis that require the actual data. We typically use such code:
`das = [da.getData() if isinstance(da, model.DataArrayShadow) else da for da in das]`

BTW, the comment "open_acquisition [has] two loading paths" is not really correct. It just can return data of two types: DataArray or DataArrayShadow. So, technically, all the `List[model.DataArray]` should be `List[Union[model.DataArray, model.DataArrayShadow]]`.

src/odemis/gui/cont/tabs/stream_load_mixin.py
```diff
@@ -0,0 +1,96 @@
...
+        tiff.export(filename_converted, raw_data, compressed=True, pyramid=True)
+        data_converted = open_acquisition(filename_converted)
+        return data_converted, filename_converted
+    except Exception as e:
+        logging.warning(f"Failed to convert to pyramidal: {e}")
+        return data, filename
+
+def _get_converted_filename(filename: os.PathLike):
```
pieleric: Maybe `get_pyramidal_filename` would be more specific?
Also, add return type annotation. `-> os.PathLike`

src/odemis/gui/cont/tabs/stream_load_mixin.py
```diff
@@ -0,0 +1,96 @@
...
+    Check if not already pyramidal and check if resolution is large enough to validify the conversion.
+    # The 'maxzoom' property exists only when pyramidal, so it seems.
+    :param data: data to check for pyramidal
+    :return: True if not yet pyramidal and resolution is large enough, False otherwise
+    """
+    for im in data:
+        if not hasattr(im, "maxzoom") and numpy.prod(im.shape[-2:]) >= PYRAMIDAL_CONVERSION_MIN_RES:
+            return True
+
+def _convert_to_pyramidal(data: List[model.DataArray], filename: os.PathLike):
```
pieleric: add return type annotation

src/odemis/gui/cont/tabs/stream_load_mixin.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def _needs_conversion(data: List[model.DataArray]):
+    """
+    Check if not already pyramidal and check if resolution is large enough to validify the conversion.
+    # The 'maxzoom' property exists only when pyramidal, so it seems.
+    :param data: data to check for pyramidal
+    :return: True if not yet pyramidal and resolution is large enough, False otherwise
+    """
+    for im in data:
```
pieleric: I'd be totally fine (and a little bit less worried of breaking some odd cases on the SPARC) if you had an extra check that there is only 1 data. That should cover all the use cases we want to handle (on the METEOR). 

I would suggest to be even more careful, by checking that the data is only 2D. To do so, you can check that MD_DIMS is only YX, CYX, or YXC. (the C is to accept RGB data).

src/odemis/util/dataio.py
pieleric: Please add type annotation to indicate that it accepts os.PathLike and returns 2 str.

src/odemis/gui/cont/tabs/stream_load_mixin.py
```diff
@@ -0,0 +1,96 @@
...
+from odemis.util.dataio import open_files_and_stitch, open_acquisition
+from odemis.util.dataio import splitext
+
+PYRAMIDAL_CONVERSION_MIN_RES = 2048 ** 2  # Minimal resolution to trigger pyramidal conversion
+PYRAMIDAL_CONVERSION_SUFFIX = "_pyramidal"
+
+class StreamLoadMixin(ABC):
+    """Mixin for loading files in tabs. Note: it requires a load_streams method to be implemented on the parent"""
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
pieleric: Typically, if your method only does call super(), then you can delete it. That should be the case here.

src/odemis/gui/cont/tabs/analysis_tab.py
```diff
@@ -346,6 +341,14 @@ def _get_time_spectrum_streams(self, spec_streams):
 
         return spectrum, temporalspectrum, timecorrelator, angularspectrum
 
+    @call_in_wx_main
+    def load_streams(self, data: List[model.DataArray], **kwargs):
```
pieleric: Note that in most of the rest of Odemis, we call such list of DataArrays/DataArrayShadows, a "das" (as in DataArrayS). I think `data` is also fine, but the lack of plural mark makes it harder to remember it's a list.

pieleric: `**kwargs` is not used as-is, but there are actually expectations for `filename` and `extend`. So just put these arguments explicitly `filename, extend` (+ add docstring).
Especially, in the docstring it would help to explain the relationship between filename and data. Here the filename is just "informational", right? IOW, you can pass any string and it'll just display it in the GUI, without failing because the file doesn't actually contain the data.

src/odemis/gui/cont/tabs/stream_load_mixin.py
```diff
@@ -0,0 +1,96 @@
...
+        # Try to get the raw data in two ways, due to the open_acquisition method having two loading paths
+        if hasattr(data[0], "data"):
+            raw_data = data[0]
+        elif hasattr(data[0], "getData"):
+            raw_data = data[0].getData()
+        tiff.export(filename_converted, raw_data, compressed=True, pyramid=True)
+        data_converted = open_acquisition(filename_converted)
+        return data_converted, filename_converted
+    except Exception as e:
+        logging.warning(f"Failed to convert to pyramidal: {e}")
```
pieleric: Please add `exc_info=True`, so that we can see the backtrace and more easily debug in case of issue.

src/odemis/gui/cont/tabs/stream_load_mixin.py
```diff
@@ -0,0 +1,96 @@
...
+from typing import List, Optional
+
+import numpy
+
+from odemis import dataio, model
+from odemis.dataio import tiff
+from odemis.util.dataio import open_files_and_stitch, open_acquisition
+from odemis.util.dataio import splitext
+
+PYRAMIDAL_CONVERSION_MIN_RES = 2048 ** 2  # Minimal resolution to trigger pyramidal conversion
```
pieleric: An image resolution is typically X*Y, but here it's just a big number. So I think it'd be clearer to name it `PYRAMIDAL_CONVERSION_MIN_PIXELS`

src/odemis/gui/cont/tabs/correlation_tab.py
```diff
@@ -218,9 +211,10 @@ def select_acq_file(self, extend: bool = False, tileset: bool = False):
                 extend = True  # If multiple files loaded, the first one is considered main one
 
     @call_in_wx_main
-    def load_streams(self, data: List[model.DataArray]) -> None:
+    def load_streams(self, data: List[model.DataArray], **kwargs) -> None:
         """ Load the data in the overview viewports
         :param data: (list[model.DataArray]) list of data arrays to load as streams
+        :param kwargs: For compatibility with mixin
```
pieleric: Instead of `**kwargs`, you could name it `**_`, to make it clear that it's not going to be used.

src/odemis/gui/cont/tabs/stream_load_mixin.py
```diff
@@ -0,0 +1,96 @@
...
+PYRAMIDAL_CONVERSION_SUFFIX = "_pyramidal"
+
+class StreamLoadMixin(ABC):
```
pieleric, line 38: As a note, my experience with Mixins is "mixed". Mainly because it brings multi-inheritance, and that makes code harder to follow, and harder to maintain. Here, I think you could have gone with a "controller" pattern (where the tab code instantiates a another class which takes care of some part of the widgets in the tab), or with just a "library" (or "util") pattern, where for instance `load_tileset()` would return the data & filename, after conversion if required.
Anyway, let's leave it as Mixin. For now, it's good enough!

src/odemis/gui/cont/tabs/test/stream_load_mixin_test.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def test_tiff_non_pyramidal_small(self):
+        """Test tiff data that is too small is not converted"""
+        filename = self.create_mock_tiff(shape=(SMALL_SIDE, SMALL_SIDE), pyramidal=False)
+        self.tab.load_data(filename)
+        self.assertNotIn(PYRAMIDAL_CONVERSION_SUFFIX, str(self.tab.loaded_filename))
+
+    def test_tiff_non_pyramidal_big(self):
+        """Test tiff data that is large enough is converted"""
+        filename = self.create_mock_tiff(shape=(LARGE_SIDE, LARGE_SIDE), pyramidal=False)
```
pieleric, line 101: Did you (manually) check that it works fine with all sort of data? I'm specifically thinking of RGB data (that many SEMs provide). I think you want to check with at least one dataset of TFS, Zeiss, Tescan, and JEOL (see in the SE Drive: Test Data/external/). It'd also be good to check that nothing goes wrong when opening a big z-stack, or SPARC data (see all the TIFF files in Test Data/good-data/sparc/)

pieleric (changes requested):

src/odemis/gui/cont/tabs/test/stream_load_mixin_test.py
tmoerkerken: I manually checked the external datasets from drive. I had to make some changes to make them all work.

src/odemis/gui/cont/tabs/stream_load_mixin.py
```diff
@@ -0,0 +1,150 @@
...
+        if im_len == 2:
+            dims = "YX"  # Assume regular, non-rgb, 2D data
+        elif im_len == 3:
+            min_shape = min(im.shape)
+            # Only handle rgb case
+            c_index = im.shape.index(min_shape)
+            if c_index == 0:
+                dims = "CYX"
+            elif c_index == 2:
+                dims = "YXC"
```
pieleric: This code is complicated because you default MD_DIMS to an empty string. The "good" default is the following: `dims = im.metadata.get(model.MD_DIMS, "CTZYX"[-im.ndim::])` 

It's explained in English in the _metadata.py file, but I think we could also add this "tip" code to the developer's manual. I'll look into this.

pieleric (changes requested):

src/odemis/gui/cont/tabs/stream_load_mixin.py
tmoerkerken: I agree. I replaced it and rechecked the tests.

src/odemis/gui/cont/tabs/stream_load_mixin.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+    if not len(das) == 1:
+        return False
+
+    im = das[0]
+    dims = im.metadata.get(model.MD_DIMS, "CTZYX"[-im.ndim:])
+
+    if (not hasattr(im, "maxzoom")  # The 'maxzoom' property exists only when pyramidal
+        and (im.shape[dims.index("Y")] * im.shape[dims.index("X")]) >= PYRAMIDAL_CONVERSION_MIN_PIXELS
+        and dims in ["YX", "CYX", "YXC"]  # Check if 2D
```
pieleric: There can be cases where the data has no X or no Y, which will make the second line of the `if` unhappy. => swap the second and third line. (and update the docstring to reflect the order?)

pieleric (changes requested):
Just a tiny more corner case I could think of...

pieleric (approved):

PR #3408, opened 2026-03-18
[fix] TestDetermineZPosition in z_localization_test.py

K4rishma:
Fix the z_stack_step_size value and add a folder path for images such that test cases can be run any path.


pieleric (changes requested):

pieleric (approved):
Looks good to me... assuming you implement the suggestions by Thera and copilot.

tmoerkerken (approved):

tepals (changes requested):
one small thing left

tepals (approved):

PR #3418, opened 2026-03-24
[MSD-494][feat] driver tmcm: if device fails to reply, immediately retry

pieleric:
In some very rare cases, mainly during referencing, the TMCM-6110 fails
to reply to instructions. When this happens, typically the next or
second next instruction work again. It's not clear why this happens, but instead
of completely failing the action (eg, referencing), detect such
communication issue and retry.

Example error that this solves:
```
  File "/home/meteor/development/odemis/src/odemis/driver/tmcm.py", line 2100, in _doReference
    self._waitReferencing(aid)  # block until it's over
  File "/home/meteor/development/odemis/src/odemis/driver/tmcm.py", line 1604, in _waitReferencingStd
    self._releaseRefSwitch(axis)
  File "/home/meteor/development/odemis/src/odemis/driver/tmcm.py", line 1507, in _releaseRefSwitch
    self.SetIO(2, refswitch, 0)
  File "/home/meteor/development/odemis/src/odemis/driver/tmcm.py", line 1071, in SetIO
    self.SendInstruction(14, port, bank, value)
  File "/home/meteor/development/odemis/src/odemis/driver/tmcm.py", line 928, in SendInstruction
    raise IOError("Received only %d bytes after %s" %
OSError: Received only 0 bytes after 8, 14, 0, 2, 0 (24)
```

### DIFF
--- a/src/odemis/driver/tmcm.py
+++ b/src/odemis/driver/tmcm.py
@@ -873,91 +873,135 @@ class TMCLController(model.Actuator):
 
         return axis_params, global_params, io_config
 
-    # TODO: finish this method and use where possible
-    def SendInstructionRecoverable(self, n, typ=0, mot=0, val=0):
-
-        try:
-            self.SendInstruction(n, typ, mot, val)
-
-        except IOError:
-            # TODO: could serial.outWaiting() give a clue on what is going on?
-
-
-            # One possible reason is that the device disappeared because the
-            # cable was pulled out, or the power got cut (unlikely, as it's
-            # powered via 2 sources).
-
-            # TODO: detect that the connection was lost if the port we have
-            # leads to nowhere. => It seems os.path.exists should fail ?
-            # or /proc/pid/fd/n link to a *(deleted)
-            # How to handle the fact it will then probably get a different name
-            # on replug? Use a pattern for the file name?
-
-            self._resynchonise()
-
-    def SendInstruction(self, n, typ=0, mot=0, val=0):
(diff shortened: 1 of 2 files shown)


nandishjpatel (approved):

src/odemis/driver/tmcm.py
```diff
@@ -873,91 +873,135 @@ def parse_tsv_config(f):
...
+                try:
+                    return self._receive_answer(n, msg)
+                except TimeoutError:
+                    # Only recover when 0 bytes were received, other errors are typically more severe,
+                    # so we just immediately fail the instruction without trying to recover.
+                    # Also, this most likely means that the device discarded the instruction,
+                    # so we can just retry it after recovery without worrying about side effects.
+                    logging.warning("Received 0 bytes after instruction %d, checking communication", n)
+
+                if attempt >= 2:  # Only try to recover once
```
tmoerkerken, line 986: Maybe something to expose into a constant at the beginning of the file (e.g. MAX_RETRY_ATTEMPTS = 1)

tmoerkerken (approved):

src/odemis/driver/tmcm.py
pieleric: I'm actually not super found of such constants, because it makes the code less obvious on how many times it's going to happen, and we don't get much benefits as it's only used once. It's the same for the "ping_attempt". => Let's only do that if the code gets duplicated.

PR #3420, opened 2026-03-24
[MSD-499][feat] FixedPositionsActuator: smart detection of the referencing position

pieleric:
Commit 159a10f81a (FixedPositionsActuator on rotation axis could take a long time to reference)
fixed a bug that caused the referencing from almost never happening on
cyclic actuators.

However, if the default ref_start value is on the wrong side of the
reference switch (for example because the actual axis is inverted), then
the referencing can take a long time. On some old METEOR, that can cause up to
50s delay to switch slot.

So adjust it in 2 ways:
* 2x less frequent referencing (to be more in par with the previous behaviour)
* "Smart" finding of the reference switch: the first 2 referencing are done
  from each sides of the reference switch, and afterwards it uses the fastest option.
   This "smart" behaviour is skipped if the position is given in the mic file.
   It can also detect that the first referencing was quick, and so never tries the other direction.

### DIFF
--- a/src/odemis/driver/actuator.py
+++ b/src/odemis/driver/actuator.py
@@ -27,6 +27,7 @@ import logging
 import math
 import numbers
 import threading
+import time
 from concurrent.futures._base import CancelledError, FINISHED
 from concurrent import futures
 from typing import Dict, Union, Set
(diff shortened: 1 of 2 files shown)


K4rishma (approved):

tmoerkerken (approved):

pieleric:
It's been tested this morning at a customers' site, and seems to work, but I'll wait until tomorrow morning to merge it, to be certain.

PR #3421, opened 2026-03-26
[S30-79] implement cl templates wireframe in the gui

nandishjpatel:
cl_acq_recipes_gui-2026-03-23_07 34 24-ezgif com-video-to-gif-converter

### DIFF
--- a/plugins/la_spec.py
+++ b/plugins/la_spec.py
@@ -30,15 +30,17 @@ from collections import OrderedDict
 from concurrent import futures
 import functools
 import logging
-import numpy
 from odemis import model
 from odemis.gui.model import TabName
 from odemis.acq.stream import SpectrumSettingsStream, POL_POSITIONS, SEMSpectrumMDStream
 import odemis.gui
 from odemis.gui.conf import data
 from odemis.gui.conf.data import get_local_vas
+from odemis.gui.cont.stream import StreamController
+from odemis.gui.cont.stream_bar import SparcStreamsController
+from odemis.gui.model.tab_gui_data import SparcAcquisitionGUIData
 from odemis.gui.plugin import Plugin
-from odemis.model import MD_POL_NONE, MD_DESCRIPTION
+from odemis.model import MD_POL_NONE
 from odemis.util import executeAsyncTask
 
 import odemis.gui.conf.util as confutil
@@ -267,15 +269,23 @@ class SpecExtraPlugin(Plugin):
             ))
         )
 
-    def addSpectrum(self, name, detector):
+    def addSpectrum(
+        self,
+        name: str,
+        detector: model.DigitalCamera,
+        tab_data: SparcAcquisitionGUIData,
+        stctrl: SparcStreamsController,
+        **kwargs
+    ) -> StreamController:
         """
-        name (str): name of the stream
-        detector (DigitalCamera): spectrometer to acquire the spectrum
+        :param name: (str) name of the stream
+        :param detector: (DigitalCamera) spectrometer to acquire the spectrum
+        :param tab_data: (SparcAcquisitionGUIData) the data of the acquisition tab
+        :param stctrl: (SparcStreamsController) the SPARC stream controller to which the stream should be added
+        :param kwargs: other options, can contain "settings_entries"
         """
         logging.debug("Adding spectrum stream for %s", detector.name)
-
-        main_data = self.main_app.main_data
-        stctrl = self._tab.streambar_controller
+        main_data = tab_data.main
 
         spg = stctrl._getAffectingSpectrograph(detector, default=main_data.spectrograph)
 
@@ -295,7 +305,7 @@ class SpecExtraPlugin(Plugin):
         axes = stctrl._filter_axes(axes)
 
         spec_stream = LASpectrumSettingsStream(
-            name,
+            stctrl.get_unique_stream_name(name),
             detector,
             detector.data,
             main_data.ebeam,
@@ -307,6 +317,9 @@ class SpecExtraPlugin(Plugin):
             detvas=get_local_vas(detector, main_data.hw_settings_config),
         )
         stctrl._set_default_spectrum_axes(spec_stream)
+
+        if kwargs.get("settings_entries"):
+            spec_stream.set_settings_entries(kwargs["settings_entries"])
 
         # Create the equivalent MDStream
         sem_stream = self._tab.tab_data_model.semStream
--- a/plugins/la_spec.py
+++ b/plugins/la_spec.py
@@ -30,14 +30,16 @@ from collections import OrderedDict
 from concurrent import futures
 import functools
 import logging
-import numpy
 from odemis import model
 from odemis.acq.stream import SpectrumSettingsStream, POL_POSITIONS, SEMSpectrumMDStream
 import odemis.gui
 from odemis.gui.conf import data
 from odemis.gui.conf.data import get_local_vas
+from odemis.gui.cont.stream import StreamController
+from odemis.gui.cont.stream_bar import SparcStreamsController
+from odemis.gui.model.tab_gui_data import SparcAcquisitionGUIData
 from odemis.gui.plugin import Plugin
-from odemis.model import MD_POL_NONE, MD_DESCRIPTION
+from odemis.model import MD_POL_NONE
 from odemis.util import executeAsyncTask
 
 import odemis.gui.conf.util as confutil
@@ -266,15 +268,23 @@ class SpecExtraPlugin(Plugin):
             ))
         )
 
-    def addSpectrum(self, name, detector):
+    def addSpectrum(
+        self,
+        name: str,
+        detector: model.DigitalCamera,
+        tab_data: SparcAcquisitionGUIData,
+        stctrl: SparcStreamsController,
+        **kwargs
+    ) -> StreamController:
         """
-        name (str): name of the stream
-        detector (DigitalCamera): spectrometer to acquire the spectrum
+        :param name: (str) name of the stream
+        :param detector: (DigitalCamera) spectrometer to acquire the spectrum
+        :param tab_data: (SparcAcquisitionGUIData) the data of the acquisition tab
+        :param stctrl: (SparcStreamsController) the SPARC stream controller to which the stream should be added
+        :param kwargs: other options, can contain "settings_entries"
         """
         logging.debug("Adding spectrum stream for %s", detector.name)
-
-        main_data = self.main_app.main_data
-        stctrl = self._tab.streambar_controller
+        main_data = tab_data.main
 
         spg = stctrl._getAffectingSpectrograph(detector, default=main_data.spectrograph)
 
@@ -294,7 +304,7 @@ class SpecExtraPlugin(Plugin):
         axes = stctrl._filter_axes(axes)
 
         spec_stream = LASpectrumSettingsStream(
-            name,
+            stctrl.get_unique_stream_name(name),
             detector,
             detector.data,
             main_data.ebeam,
@@ -306,6 +316,9 @@ class SpecExtraPlugin(Plugin):
             detvas=get_local_vas(detector, main_data.hw_settings_config),
         )
         stctrl._set_default_spectrum_axes(spec_stream)
+
+        if kwargs.get("settings_entries"):
+            spec_stream.set_settings_entries(kwargs["settings_entries"])
 
         # Create the equivalent MDStream
         sem_stream = self._tab.tab_data_model.semStream
--- a/plugins/la_spec.py
+++ b/plugins/la_spec.py
@@ -244,7 +244,9 @@ class SpecExtraPlugin(Plugin):
                 actname = "Large Area Spectrum"
             else:
                 actname = "Large Area Spectrum with %s" % (sptm.name,)
-            act = functools.partial(self.addSpectrum, name=actname, detector=sptm)
+            act = functools.partial(
+                self.addSpectrum, name=actname, detector=sptm, tab_data=self._tab.tab_data_model, stctrl=stctrl
+            )
             stctrl.add_action(actname, act)
 
         # We "patch" the gui.conf.data for our special stream
@@ -278,10 +280,10 @@ class SpecExtraPlugin(Plugin):
         **kwargs
     ) -> StreamController:
         """
-        :param name: (str) name of the stream
-        :param detector: (DigitalCamera) spectrometer to acquire the spectrum
-        :param tab_data: (SparcAcquisitionGUIData) the data of the acquisition tab
-        :param stctrl: (SparcStreamsController) the SPARC stream controller to which the stream should be added
+        :param name: name of the stream
+        :param detector: spectrometer to acquire the spectrum
+        :param tab_data: the data of the acquisition tab
+        :param stctrl: the SPARC stream controller to which the stream should be added
         :param kwargs: other options, can contain "settings_entries"
         """
         logging.debug("Adding spectrum stream for %s", detector.name)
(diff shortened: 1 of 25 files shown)


src/odemis/acq/stream/_base.py
```diff
@@ -1368,3 +1376,56 @@ def getRawMetadata(self):
     def force_image_update(self):
         """Force the stream image to be updated"""
         self._shouldUpdateImage()
+
+    def _get_settings_order(self):
```
pieleric: Add type annotation.

src/odemis/acq/stream/_base.py
```diff
@@ -1368,3 +1376,56 @@ def getRawMetadata(self):
...
+                isinstance(x, (int, float)) for x in value
+            ):
+                value = tuple(value)
+
+            try:
+                getattr(self, key).value = value
+            except Exception:
+                logging.exception(
+                    "Failed to set %s setting %s to value %s",
+                    entries.get("class", self.__class__.__name__),
```
pieleric: Let's not report the "class" entry, as it's discarded above. If the value is different from the actual class, it'll be more confusing than helpful.
Also, a warning is enough, as we are just skipping the VA. You can add exc_info=True if you think the backtrace is helpful.

src/odemis/acq/stream/_base.py
```diff
@@ -1368,3 +1376,56 @@ def getRawMetadata(self):
...
+            if key == "tint":
+                self.tint.value = img.md_format_to_tint(entries["tint"])
+            elif isinstance(value, list) and all(
+                isinstance(x, (int, float)) for x in value
+            ):
+                value = tuple(value)
+
+            try:
+                getattr(self, key).value = value
+            except Exception:
```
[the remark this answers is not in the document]
pieleric: The original code was correct. Now, it has this error.

src/odemis/dataio/tiff.py
```diff
@@ -1548,6 +1553,10 @@ def _addImageElement(root, das, ifd, rois, fname=None, fuuid=None):
         ad = ET.SubElement(ime, "AcquisitionDate")
         ad.text = time.strftime("%Y-%m-%dT%H:%M:%S", time.gmtime(globalAD))
 
+    if model.MD_ACQ_RECIPES in globalMD:
+        ar = ET.SubElement(ime, "AcquisitionRecipes")
+        ar.text = globalMD[model.MD_ACQ_RECIPES]
+
```
pieleric: With OME, you cannot add any tag you want where you want, because some software will decide it's not a OME XML compliant, and refuse to read it. See https://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2016-06/ome.html

So, instead, for all "non standard" metadata, you can add it as StructuredAnnotations/MapAnnotation, see code below (like for MD_EXTRA_SETTINGS).

src/odemis/dataio/hdf5.py
pieleric, line 1: Please add a test case for these changes

src/odemis/dataio/tiff.py
pieleric, line 1: Please add a test case for these changes

plugins/la_spec.py
```diff
@@ -267,15 +269,23 @@ def __init__(self, microscope, main_app):
...
+        stctrl: SparcStreamsController,
+        **kwargs
+    ) -> StreamController:
         """
-        name (str): name of the stream
-        detector (DigitalCamera): spectrometer to acquire the spectrum
+        :param name: (str) name of the stream
+        :param detector: (DigitalCamera) spectrometer to acquire the spectrum
+        :param tab_data: (SparcAcquisitionGUIData) the data of the acquisition tab
+        :param stctrl: (SparcStreamsController) the SPARC stream controller to which the stream should be added
```
tepals: typing info in the docstring is not necessary if you have it with the arguments already

src/odemis/gui/cont/tabs/sparc_acquisition_tab.py
```diff
@@ -24,29 +24,433 @@
...
+                if "BlSpectrumSettingsStream" == class_name:
+                    self._createBlSpectrumStreamActions()
+                elif "LASpectrumSettingsStream" == class_name:
+                    self._createLargeAreaSpectrumStreamActions()
+                elif "SpectrumRawSettingsStream" == class_name:
+                    self._createSpectrumRawStreamAction()
+                elif "SpectrumArbitraryOrderSettingsStream" == class_name:
+                    self._createSpectrumArbitraryOrderStreamAction()
+                elif "CL intensity on CCD" in name:
+                    self._createCLIntensityCCDStreamAction()
```
tepals: is there a way to not have the class names hardcoded here? Feels like someone could easily make a mistake

src/odemis/gui/cont/tabs/sparc_acquisition_tab.py
```diff
@@ -531,6 +941,193 @@ def on_hardware_protect(self) -> None:
...
+        """
+        if not self.tab_data_model.recipes:
+            self.current_recipes.SetValue("")
+            return
+        label = ", ".join(sorted(self.tab_data_model.recipes))
+        if is_modified:
+            label += " (modified)"
+        self.current_recipes.SetValue(label)
+
+    def _export_streams_settings(self, evt):
```
tepals: I have a hard time reviewing this function and `_import_streams_settings`, because they contain a lot of code. Would it be possible to split them into smaller functions?

tepals (approved):
I am not too familiar with this code, but overall it looks fine

src/odemis/gui/comp/foldpanelbar.py
```diff
@@ -251,16 +251,22 @@ def children_to_sizer(self):
...
+    or hide the associated panel. See panel_tab_sparc_acquis.xrc object cbar_acq_recipes.
     """
 
-    def __init__(self, parent, caption, collapsed):
+    def __init__(self, parent, caption, collapsed, foldable=False):
         """
         :param parent: Parent window (FoldPanelItem)
         :param caption: Header caption (str)
         :param collapsed: Draw the CaptionBar collapsed or not (boolean)
+        :param foldable: Whether the CaptionBar is foldable (boolean)
```
pieleric: Please add that it is only used if the parent is not a FoldPanelItem.

src/odemis/gui/comp/settings.py
```diff
@@ -585,4 +585,18 @@ def add_run_btn(self, label_text):
                          flag=wx.ALIGN_CENTRE_VERTICAL | wx.EXPAND | wx.TOP | wx.BOTTOM, border=5)
         return lbl_ctrl, run_btn
 
+    @control_bookkeeper
+    def add_btn(self, label_text, btn_label):
```
pieleric: Please add type annotation (and remove the type info from the docstring). All new functions should use this new format.

src/odemis/gui/cont/tabs/sparc_acquisition_tab.py
```diff
@@ -24,29 +24,433 @@
...
+            self.file_path_ctrl.Clear()
+            if self.file_paths:
+                for fp in self.file_paths:
+                    self.file_path_ctrl.Append(os.path.basename(fp))
+                self.file_path_ctrl.SetToolTip("\n".join(self.file_paths))
+            else:
+                self.file_path_ctrl.Append("No file selected")
+                self.file_path_ctrl.SetToolTip("")
+
+    def _get_filepaths_from_filedialog(self) -> Union[list[str], None]:
```
pieleric: Please use `List` and `Dict` from typing (everywhere), otherwise this breaks on Python 3.8.

src/odemis/gui/cont/tabs/sparc_acquisition_tab.py
```diff
@@ -24,29 +24,433 @@
...
+        stream_bar.SetSize((-1, h))
+        self._scrolled_win.SetVirtualSize((-1, h))
+        stream_bar.Layout()
+        self._scrolled_win.Layout()
+
+    def on_size(self, evt):
+        self._fit_streams()
+        evt.Skip()
+
+    def _createBlSpectrumStreamActions(self):
```
pieleric: For this function and all the similar functions below: they should not be there. You are hard-coding code for plugins inside the main code. Plugins should be independent. Especially, right now it means that if we decide to create a new plugin, it will not be possible, unless we change the code in Odemis itself, which defeats the purpose of the plugin.

So either you find a way to directly call the existing plugin code (eg, based on the "actions" available in "ADD STREAM", or you have to create a new dedicated plugin option (eg, a little bit similar to Plugin.addMenu(), but to add a new stream type/stream creation).

src/odemis/gui/cont/tabs/sparc_acquisition_tab.py
pieleric: I don't think we can have these class names hardcoded. They should be provided by the plugin, and then this code automatically picks all the ones (from the loaded plugins).

src/odemis/gui/cont/tabs/sparc_acquisition_tab.py
```diff
@@ -24,29 +24,433 @@
...
+
+def show_warning(message: str):
+    """
+    Show a warning message box.
+    :param message: The warning message to display.
+    """
+    wx.MessageBox(message, "Warning", wx.ICON_WARNING | wx.OK)
+
+
+class FilesDialog(WizardPageSimple):
```
pieleric: Please move the Wizard pages to a new separate file (somewhere in odemis.gui.cont). This file is for the acquisition tab, which is big enough.

src/odemis/gui/cont/tabs/sparc_acquisition_tab.py
pieleric: I think that the export and import functions could go on a dedicated "controller" class, which receives the two buttons at init. 

It should then be much easier to have multiple functions, without confusing them with the rest of the sparc acquisition tab code.

src/odemis/gui/cont/tabs/sparc_acquisition_tab.py
```diff
@@ -531,6 +941,193 @@ def on_hardware_protect(self) -> None:
...
+        Export the current stream settings entries to a .json file, using a wizard to let the user select which streams
+        to export and the file path.
+        """
+        def validate_wizard_page(evt):
+            """Validate the wizard page before changing to the next page."""
+            page = evt.GetPage()
+            if page and evt.GetDirection():
+                if not page.validate():
+                    # If validation fails, prevent page change
+                    evt.Veto()
```
pieleric: No need to be within the function, just move to a separate method.

src/odemis/gui/cont/tabs/sparc_acquisition_tab.py
```diff
@@ -531,6 +941,193 @@ def on_hardware_protect(self) -> None:
...
+
+        # "roi" is a non-settings VA by default. For SPARC the roi VA of any stream is relevant and we want to be able
+        # to export it, so we temporarily remove "roi" from the NON_SETTINGS_VA list. The stream_page.roi_cbox allows
+        # the user to choose whether to keep the roi settings or not.
+        if "roi" in NON_SETTINGS_VA:
+            NON_SETTINGS_VA.remove("roi")
```
pieleric: NO.
`NON_SETTINGS_VA` is a constant. You shouldn't change it live, ever. 
I think the best is to add a parameter to functions that need the NON_SETTINGS_VA info, to override the NON_SETTINGS_VA value.

pieleric (changes requested):

src/odemis/model/_metadata.py
```diff
@@ -25,6 +25,7 @@
 # This list of constants are used as key for the metadata
 MD_EXP_TIME = "Exposure time"  # s
 MD_ACQ_DATE = "Acquisition date"  # s since epoch
+MD_ACQ_RECIPES = "Acquisition recipes"  # str
```
pieleric: Please add a description to explain what this is for (when you have no context).

src/odemis/gui/cont/tabs/sparc_acquisition_tab.py
pieleric: Actually, looking at it a little further, it seems the NON_SETTINGS_VA was only used in acq.stream_settings. This is only used on the SECOM & METEOR (where the roi cannot be changed). So it might be fine to just never include the "roi" in that constant.

src/odemis/gui/cont/tabs/sparc_acquisition_tab.py
nandishjpatel: Its now split into smaller functions

src/odemis/gui/cont/recipes.py
```diff
@@ -0,0 +1,305 @@
...
+
+        # Setup UI
+        recipes_settings = SettingsPanel(self._pnl_acq_recipes, size=(400, 90))
+        _, self.current_recipes = recipes_settings.add_readonly_field("Current recipes", "")
+        _, self.import_btn = recipes_settings.add_btn(label_text="Read from file(s)", btn_label="Import...")
+        _, self.export_btn = recipes_settings.add_btn(label_text="Write to file", btn_label="Export...")
+        self.export_btn.Bind(wx.EVT_BUTTON, self._export_streams_settings)
+        self.import_btn.Bind(wx.EVT_BUTTON, self._import_streams_settings)
+
+        cbar = wx.FindWindowByName("cbar_acq_recipes", panel)
```
pieleric: That shouldn't be necessary, as the named widgets are available as attributes on the Panel instance (see main_xrc.py. 
```suggestion
        cbar = panel.cbar_acq_recipes
```

=> In such case you probably can drop your changes to xh_delmic.py

src/odemis/gui/cont/recipes.py
```diff
@@ -0,0 +1,305 @@
...
+
+    def _on_any_va(self, _):
+        """
+        Called whenever a VA of any stream is updated. Used to update the current recipes display,
+        as any change in the stream settings may change the recipe.
+        """
+        if self.current_recipes.GetValue():
+            self._update_current_recipes_label(is_modified=True)
+
+    def _on_streams(self, streams):
```
pieleric: Please add type annotation.

src/odemis/gui/cont/recipes.py
```diff
@@ -0,0 +1,305 @@
...
+
+    def validate_wizard_page(self, evt):
+        """Validate the wizard page before changing to the next page."""
+        page = evt.GetPage()
+        if page and evt.GetDirection():
+            if not page.validate():
+                # If validation fails, prevent page change
+                evt.Veto()
+
+    def _create_export_wizard(self):
```
pieleric, line 132: Add return type annotation.

src/odemis/gui/cont/recipes.py
```diff
@@ -0,0 +1,305 @@
...
+        except Exception as e:
+            logging.exception("An error occurred while exporting the stream settings.")
+            show_error(f"An error occurred while exporting the stream settings:\n{str(e)}")
+        finally:
+            self._tab_data.streams.subscribe(self._on_streams, init=True)
+            self._tab_data.streams.subscribe(self._on_any_va)
+            self._sem_stream.set_settings_entries(orig_sem_sse)
+            wizard.Destroy()
+
+    def before_import_wizard_page_changed(self, evt, files_page: FilesPage, streams_page: Sparc2StreamsPage):
```
pieleric: For the evt, you can add the specific event type. It would help reading the method's code.

pieleric (changes requested):
Looks pretty much ready to merge. Just some minor comments.

pieleric (approved):

PR #3430, opened 2026-04-01
[MSD-527][fix] Add callback on stream change

tmoerkerken:
The `Correlate FIB/FM` button was not enabling correctly in the FIBSEM tab, due to the recent lazy loading changes. 
We only enabled/disabled the button (based on some conditions) after a feature change. But now with lazy loading, the conditions can change _after_.

Add a callback on streams changing, to correctly enable/disable the `Correlate FIB/FM` button.


pieleric (approved):
Apart from the method signature to fix, it looks good.

ilyushkin (approved):

PR #3435, opened 2026-04-07
[SWM-223][refactor] tab names to Enum

tmoerkerken:
Changed all hard-coded instances of tab name strings to use a newly introduced Enum.


tepals (approved):
Looks good to me. As part of the CI server testing we run this script that tries to start all backends and frontends, but do not report on it https://github.com/delmic/odemis/blob/master/util/run_intg_tests.py . Could you either run this script before merging, or check the logs on the CI server after merging?

pieleric (changes requested):

tmoerkerken:
Again, the branch is broken, due to an intermediate merge with master before me pushing. Reopening a clean one.

PR #3437, opened 2026-04-08
[MSD-443][feat] 3D correlation improvements

tmoerkerken:
This PR introduces a variety of improvements to the 3D correlation workflow:
- Improve Refine Z method → Using more robust 3D method
- Improve Z value in table → Z column now shows slice index. This means that the automatically found position can now be verified using the slice slider when MIP is disabled.
- Improve timely rendering (sometimes needing click in viewport to update the scene)
→ Now using improved drawing calls that automatically update the viewports when fiducials change
- Remove usage of reflection images in refine method → Only using non-reflective streams now
- Remove auto refinement trigger
→ Auto refinement overwrote manual edits, which is undesirable
- Improve randomness → Making sure to reset random seed before every 3dct call, although unsure what the impact is at this point.
- Placing a FM fiducial when MIP is disabled should inherit the z index of the currently selected slice, but doesn't→ Fiducial now inherits slice index when only one non-reflective stream is visible when MIP is disabled.
- Fixed bug where numpad enter did not trigger 3dct algorithm

### Bad input
We see that the new algorithm is very good at recovering from bad user input. This makes placement of fiducials really simple. New method: Center of Mass (new), Gauss fit 1d (old).
debug_6_1
debug_6_2
### Other examples
See JIRA task for more examples

## Out of scope
I encountered various peculiar UX related things (listed in the JIRA ticket) that I'm not touching right now. We could improve upon them later.

### DIFF
--- a/src/odemis/gui/cont/multi_point_correlation.py
+++ b/src/odemis/gui/cont/multi_point_correlation.py
@@ -35,7 +35,7 @@ import wx
 # This is not related to any particular wxPython version and is most likely permanent.
 
 from odemis import model, util
-from odemis.acq.align.tdct import get_optimized_z_gauss, _convert_das_to_numpy_stack, run_tdct_correlation
+from odemis.acq.align.tdct import _convert_das_to_numpy_stack, run_tdct_correlation
 from odemis.acq.feature import save_features, FIBFMCorrelationData, Target, TargetType
 from odemis.acq.stream import StaticFluoStream, StaticSEMStream, StaticFIBStream, FluoStream
 from odemis.gui import conf
@@ -43,7 +43,7 @@ from odemis.gui.model import CryoGUIData
 from odemis.gui.util import call_in_wx_main
 from odemis.model import ListVA
 from odemis.util.dataio import data_to_static_streams
-from odemis.util.interpolation import interpolate_z_stack
+from odemis.util.img import get_bounding_box_slice, get_brightest_channel, compute_local_center_of_mass
 from odemis.util.units import readable_str
 
 # create an enum with column labels and position
@@ -61,12 +61,14 @@ GRID_PRECISION = 2  # Number of decimal places to display in the grid
 FIDUCIAL_PATTERN = r"^[^-]+-"
 RIM_COR_DEFAULT = 0.495  # See MD_RIM_COR. This value works fine for 50x objectives, which are common
 
-# Both functions getPixel3DCoordinates(args*, kwargs*) and getPhysical3DCoordinates(args*, kwargs*) need special
+# Both functions get_pixel_3d_coordinates(args*, kwargs*) and get_physical_3d_coordinates(args*, kwargs*) need special
 # conditions to convert between physical and pixel coordinate systems in order for multipoint correlation to operate.
 # For coordinate conversions, we assume the pixels in 3D are isosymmetric
 # i.e. size in pixel[0]=pixel[1]=pixel[2].
+COM_ROI_PADDING = 12  # Padding (pixels) for center of mass ROI extraction
+# TODO: Could adapt padding based on pixel spacing for more flexibility if needed
 
-def getPixel3DCoordinates(stream: FluoStream, p_pos: Tuple[float, float, float], check_bbox: bool = False) \
+def get_pixel_3d_coordinates(stream: FluoStream, p_pos: Tuple[float, float, float], check_bbox: bool = False) \
         -> Optional[Tuple[float, float, float]]:
     """
     Translate 3D physical coordinates into 3D pixel coordinates. The z coordinate is computed assuming iso-voxel
@@ -96,7 +98,7 @@ def getPixel3DCoordinates(stream: FluoStream, p_pos: Tuple[float, float, float],
     pixel_pos = (pixel_pos[0], pixel_pos[1], z)
     return pixel_pos
 
-def getPhysical3DCoordinates(stream: FluoStream, pixel_pos: Tuple[float, float, float])\
+def get_physical_3d_coordinates(stream: FluoStream, pixel_pos: Tuple[float, float, float])\
                              -> Optional[Tuple[float, float, float]]:
     """
     Translate 3D pixel coordinates into 3D physical coordinates. The z coordinate is computed assuming iso-voxel
@@ -175,18 +177,18 @@ class CorrelationPointsController:
         # Access the correlation points table (wxListCtrl)
         self.grid = self._panel.table_grid
 
-        # Access the Refine Z text (to check if refine_z is working or not)
+        # Access the Refine XYZ status text (to check if XYZ targeting is working or not)
         self.txt_refinez_active = self._panel.txt_refinez_active
         self.txt_refinez_active.Show(True)
 
-        # Access the Z-targeting button
+        # Access the XYZ-targeting button
         self.z_targeting_btn = self._panel.btn_z_targeting
         self.z_targeting_btn.Bind(wx.EVT_BUTTON, self._on_z_targeting)
         self.z_targeting_btn.Enable(False)
-        # Disable Z-targeting button if super z stream is available as Z-targeting is not required in that case
+        # Disable XYZ-targeting button if super z stream is available as XYZ-targeting is not required in that case
         self.refinez_active = True
         if self._tab_data_model.main.currentFeature.value.superz_stream_name:
-            self.z_targeting_btn.SetToolTip("Super Z information available, Refine Z disabled")
+            self.z_targeting_btn.SetToolTip("Super Z information available, Refine XYZ disabled")
             self.txt_refinez_active.SetLabel("Super Z information in use")
             self.refinez_active = False
 
@@ -479,11 +481,11 @@ class CorrelationPointsController:
             fib_coords.append(fib_coord)
         fib_coords = numpy.array(fib_coords, dtype=numpy.float32)
         for fm_coord in self.correlation_target.fm_fiducials:
-            fm_coord_px = getPixel3DCoordinates(self.correlation_target.fm_streams[0], fm_coord.coordinates.value)
+            fm_coord_px = get_pixel_3d_coordinates(self.correlation_target.fm_streams[0], fm_coord.coordinates.value)
             fm_coords.append(fm_coord_px)
         fm_coords = numpy.array(fm_coords, dtype=numpy.float32)
         poi_coord = self.correlation_target.fm_pois[0]
-        poi_coord_px = getPixel3DCoordinates(self.correlation_target.fm_streams[0], poi_coord.coordinates.value)
+        poi_coord_px = get_pixel_3d_coordinates(self.correlation_target.fm_streams[0], poi_coord.coordinates.value)
         poi_coords.append(poi_coord_px)
         poi_coords = numpy.array(poi_coords, dtype=numpy.float32)
         # Run the correlation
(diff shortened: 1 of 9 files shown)


src/odemis/gui/cont/multi_point_correlation.py
```diff
@@ -697,27 +697,41 @@ def _on_current_target_changes(self, target: Target) -> None:
...
             return
 
-        mip_enabled = any([stream.max_projection.value for stream in self.correlation_target.fm_streams])
+        # Only check for visible and non-reflective streams, since the refinement method is discarding other streams
+        # anyway. Doesn't make sense to check for MIP value of a stream that is not used in refining a fiducial.
+        visible_streams = [
+            s.stream for s in self._tab_data_model.views.value[0].stream_tree.flat.value
+            if s.stream.getRawMetadata()[0].get(model.MD_OUT_WL) != model.BAND_PASS_THROUGH
+        ]
+        mip_enabled = any([stream.max_projection.value for stream in visible_streams])
```
pieleric: It's a very minor optimisation, but you can drop the `[]` as `any()` works directly on iterators:

```suggestion
        mip_enabled = any(stream.max_projection.value for stream in visible_streams)
```

src/odemis/gui/cont/multi_point_correlation.py
```diff
@@ -803,40 +820,58 @@ def _on_target_changes(self, targets: List[Target]) -> None:
...
+            com_z = com[0]
+            com_y_crop = com[1] + roi[1].start
+            com_x_crop = com[2] + roi[2].start
+            # Map back to physical coordinates using optimized X, Y, and Z
+            physical_coords = get_physical_3d_coordinates(streams[0],(com_x_crop, com_y_crop, com_z))
+            # Update the model with the refined 3D coordinates
+            target_coords = self._tab_data_model.main.currentTarget.value.coordinates.value
+            target_coords[0] = physical_coords[0]
+            target_coords[1] = physical_coords[1]
+            target_coords[2] = physical_coords[2]
```
pieleric: I think this can be written as:
```suggestion
            target_coords[:] = physical_coords
```

src/odemis/util/img.py
```diff
@@ -1441,3 +1441,33 @@ def apply_zoom_on_image_coordinates(rect: List[int], z:int) -> List[int]:
...
+
+def compute_center_of_mass(image: numpy.ndarray,
+                           baseline_percentile: float = 95.0, roi_slice: tuple = None) -> Tuple[float, float, float]:
+    """
+    Computes center of mass with baselining.
+    Uses the brightest pixels (above the baseline percentile) as weights for COM calculation,
+    effectively filtering background noise.
+
+    :param image: Single-channel 3D sub-image as (Z, Y, X) array
+    :param baseline_percentile: Percentile threshold for background separation (0-100). Values lower than the baseline
```
pieleric: It's a little nitpicking, but in most of the project such value is passed as a ratio between 0 and 1 (eg, 0.95). So I think it'd be more in line with the other functions to pass a "baseline_ratio" (and convert to percentile inside the function).

pieleric (changes requested):
Only very minor comments.

src/odemis/util/img.py
```diff
@@ -1441,3 +1441,33 @@ def apply_zoom_on_image_coordinates(rect: List[int], z: int) -> List[int]:
...
+    Finds the index of the channel with the highest maximum intensity.
+    Useful for multi-channel imaging to select the brightest channel for analysis.
+
+    :param sub_image_multi: Multi-channel sub-image as (C, Z, Y, X) array
+    :returns: Index of the channel with the highest intensity
+    """
+    channel_maxes = numpy.max(sub_image_multi, axis=(1, 2, 3))
+    return int(numpy.argmax(channel_maxes))
+
+
```
K4rishma, line 1457: extra blank line to be removed

src/odemis/util/img.py
```diff
@@ -1441,3 +1441,33 @@ def apply_zoom_on_image_coordinates(rect: List[int], z: int) -> List[int]:
...
+    :param image: Single-channel 3D sub-image as (Z, Y, X) array
+    :param baseline_ratio: Ratio for background separation (0-1). Values lower than the baseline
+        will be discarded.
+    :returns: Tuple of (global_z, global_y, global_x) in pixel coordinates maintaining input order.
+    """
+    baseline = numpy.percentile(image, baseline_ratio * 100)
+    image_weights = numpy.where(image > baseline, image - baseline, 0)
+    com = scipy.ndimage.center_of_mass(numpy.asarray(image_weights))
+    return com
```
K4rishma, line 1473: empty line to be added at the end of the file

src/odemis/util/img.py
```diff
@@ -1441,3 +1441,33 @@ def apply_zoom_on_image_coordinates(rect: List[int], z: int) -> List[int]:
...
+    """
+    channel_maxes = numpy.max(sub_image_multi, axis=(1, 2, 3))
+    return int(numpy.argmax(channel_maxes))
+
+
+def compute_center_of_mass(image: numpy.ndarray,
+                           baseline_ratio: float = 0.95, roi_slice: tuple = None) -> Tuple[float, float, float]:
+    """
+    Computes center of mass with baselining.
+    Uses the brightest pixels (above the baseline percentile) as weights for COM calculation,
```
K4rishma: COM I would write the full form in the function description

src/odemis/util/img.py
```diff
@@ -1441,3 +1441,33 @@ def apply_zoom_on_image_coordinates(rect: List[int], z: int) -> List[int]:
     :return: (list of int) the rectangle in image pixel coordinates at the given zoom level
     """
     return [px // (2 ** z) for px in rect]
+
+
```
K4rishma, line 1445: extra line to be removed

src/odemis/util/img.py
```diff
@@ -1441,3 +1441,33 @@ def apply_zoom_on_image_coordinates(rect: List[int], z: int) -> List[int]:
...
+def get_brightest_channel(sub_image_multi: numpy.ndarray) -> int:
+    """
+    Finds the index of the channel with the highest maximum intensity.
```
K4rishma, line 1448: highest maximum -> maximum

src/odemis/util/img.py
```diff
@@ -1441,3 +1441,33 @@ def apply_zoom_on_image_coordinates(rect: List[int], z: int) -> List[int]:
...
+    Useful for multi-channel imaging to select the brightest channel for analysis.
+
+    :param sub_image_multi: Multi-channel sub-image as (C, Z, Y, X) array
+    :returns: Index of the channel with the highest intensity
```
K4rishma: highest intensity --> maximum intensity

K4rishma (approved):

src/odemis/gui/model/main_gui_data.py
```diff
@@ -386,6 +386,7 @@ def __init__(self, microscope):
                 # VAs for currently selected targets
                 self.targets = model.ListVA()
                 self.currentTarget = model.VigilantAttribute(None)
+                self.needs_target_refinement = model.BooleanVA(False)
```
pieleric: The MainGUIData is supposed to contain the state global to the entire GUI. I don't think that is useful here. You could move it "down" to the CryoGUIData, which represent the state just for a single tab. However, I think that what would be even more logical is to store the info directly on the `Target`. I think that what you'd want to indicate is "z is not defined yet (and needs to be computed based on data)".
So I'd suggest to extend `Target`, and either:
* support passing `None` in z, to indicate it's unknown and needs recomputation. That was my first feeling, but looking at the code, there are lots of expectation that z is a float, which are going to be annoying to lift.
* add an extra argument/attribute `z_needs_refinement` (bool, default = False)

Then, you can adjust `add_new_target()` to pass the new argument in the right cases, and `_on_current_target_changes` can directly look on its `target` argument.

src/odemis/gui/cont/multi_point_correlation.py
```diff
@@ -697,27 +697,45 @@ def _on_current_target_changes(self, target: Target) -> None:
...
+                if self._tab_data_model.main.needs_target_refinement.value:
+                    if mip_enabled:
+                        self._on_xyz_targeting(None)
+                    # When there is just one stream without MIP, use the currently selected z index
+                    elif len(visible_streams) == 1 and hasattr(visible_streams[0], "zIndex"):
+                        # We only care about z, so pass placeholders for x and y
+                        physical_coords = get_physical_3d_coordinates(
+                            visible_streams[0], (0, 0, visible_streams[0].zIndex.value))
+                        # Update the model with the refined 3D coordinates
+                        target_coords = self._tab_data_model.main.currentTarget.value.coordinates.value
```
pieleric: The function receives `self._tab_data_model.main.currentTarget.value` as argument, so you can simply write:
```suggestion
                        target_coords = target.coordinates.value
```

src/odemis/gui/cont/multi_point_correlation.py
```diff
@@ -807,34 +808,63 @@ def _on_target_changes(self, targets: List[Target]) -> None:
         if self.check_correlation_conditions():
             self._need_reprocessing()
 
-    def _on_z_targeting(self, evt) -> None:
+    def _on_xyz_targeting(self, evt) -> None:
```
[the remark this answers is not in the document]
pieleric: As this function is also called explicitly from other function, without any event, you could change the signature to be:

```suggestion
    def _on_xyz_targeting(self, evt: Optional[wx.Event] = None) -> None:
```
This way, you can call the function without argument.

pieleric (changes requested):

src/odemis/util/img.py
tmoerkerken: It's there, but it's not shown, since I'm not adding it (it was already there).

image

src/odemis/util/img.py
tmoerkerken: Yeah, you are correct. However, I did this on purpose, since all the methods in this module have two lines between methods. It felt a bit strange to suddenly use one blank line :sweat_smile: . Not sure what to do now @tepals, what do you think?

tepals: How it is done in this document is in line with PEP8, these are all top level functions and not method definitions inside a class https://peps.python.org/pep-0008/#blank-lines

src/odemis/util/img.py
tmoerkerken: I think Github never shows it in files, but shows a symbol when it is not there: image

tepals: @K4rishma if an empty line is missing at the end of  a file github will indicate this with a red circle with a line through it

Image
source: https://dan-niles.medium.com/why-github-prs-shows-that-red-circle-understanding-newlines-at-end-of-files-8a4869821694#:~:text=GitHub%20shows%20a%20red%20circle,control%20diffs%2C%20and%20Unix%20tools.

src/odemis/util/img.py
tmoerkerken: Same as other comment. I could make it one newline, but the rest of the module also has two newlines between methods.

src/odemis/util/img.py
tmoerkerken: I think it is correctly worded, but I could expand the text a bit to (try) make it more clear.

src/odemis/util/img.py
tmoerkerken: Each channel has a maximum intensity, and we pick the channel that has the highest maximum intensity (or argmax of max if you will).

src/odemis/util/img.py
tmoerkerken: Ah, I guess we have so many classes and class methods in Odemis that this is actually a rare occasion :laughing:

src/odemis/util/img.py
tmoerkerken: I think I should use the "highest maximum intensity" terminology from before, if that is not too confusing :sweat_smile:

src/odemis/gui/model/main_gui_data.py
tmoerkerken: I'm opting for the second option, since it is less involved with existing code. While technically only z truly needs refinement (since we have no clue at this point), we also refine the other two axes, so I'm going with `needs_refinement`. It would otherwise be unexpected that the other two values also change.

pieleric (approved):

PR #3438, opened 2026-04-09
[fix] Making sure feature is present before acquiring

tmoerkerken:
We found that due to recent lazy-loading changes, acquiring without a feature present caused issues. Making sure a feature is present before acquiring instead of after, resolves the problem it seems.

[Screencast from 2026-04-09 19-07-31.webm](https://github.com/user-attachments/assets/d28905bb-4c33-4536-8c28-aa09d0a988b4)


tmoerkerken:
@pieleric could you please try it again on your end? Maybe I tested a slightly different path.

pieleric (approved):
I've tested it on my computer, and it solves the encountered issue.

ilyushkin (approved):

PR #3440, opened 2026-04-10
[feat] stream: change z-stack default value to the middle Z

pieleric:
When acquiring a z-stack, often the most interesting (aka in focus) part
is near the center, while the z-edges are completely out of focus. So
starting with the lowest Z was easy in software, but not user-friendly.

=> select the middle z when opening a z-stack

### DIFF
--- a/src/odemis/gui/model/tab_gui_data.py
+++ b/src/odemis/gui/model/tab_gui_data.py
@@ -451,28 +451,8 @@ class CryoLocalizationGUIData(CryoGUIData):
 
         self.view_posture = model.VigilantAttribute(FM_IMAGING)
 
-    def _updateZParams(self):
-        # Calculate the new range of z pos
-        # NB: this is a copy of AnalysisGUIData._updateZParams
-        limits = []
-
-        for s in self.streams.value:
-            if model.hasVA(s, "zIndex"):
-                metadata = s.getRawMetadata()[0]  # take only the first
-                zcentre = metadata[model.MD_POS][2]
-                zstep = metadata[model.MD_PIXEL_SIZE][2]
-                limits.append(zcentre - s.zIndex.range[1] * zstep / 2)
-                limits.append(zcentre + s.zIndex.range[1] * zstep / 2)
-
-        if len(limits) > 1:
-            self.zPos.range = (min(limits), max(limits))
-            logging.debug("Z stack display range updated to %f - %f, ZPos: %f",
-                          self.zPos.range[0], self.zPos.range[1], self.zPos.value)
-        else:
-            self.zPos.range = (0, 0)
-
-    def _on_stream_change(self, _):
-        self._updateZParams()
+    def _on_stream_change(self, streams):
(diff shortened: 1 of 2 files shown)


tmoerkerken (approved):

src/odemis/gui/model/tab_gui_data.py
```diff
@@ -465,7 +465,15 @@ def _updateZParams(self):
                 limits.append(zcentre + s.zIndex.range[1] * zstep / 2)
 
         if len(limits) > 1:
-            self.zPos.range = (min(limits), max(limits))
+            zrng = (min(limits), max(limits))
+            if not zrng[0] <= self.zPos.value <= zrng[1] or self.zPos.range == (0, 0):
```
K4rishma: Do we need the second condidtion explicitly i.e. self.zpos.range == (0,0)?

src/odemis/gui/model/tab_gui_data.py
```diff
@@ -465,7 +465,15 @@ def _updateZParams(self):
...
+                # Completely new position => pick the middle as typically that's where the most interesting part is
+                new_z_pos = (zrng[0] + zrng[1]) / 2
+            else:
+                new_z_pos = None
+            self.zPos.range = zrng
+            if new_z_pos is not None:
+                self.zPos.value = new_z_pos
```
K4rishma: Maybe zPos.value can be updated on line 470  directly?

src/odemis/gui/model/tab_gui_data.py
```diff
@@ -706,7 +714,15 @@ def _updateZParams(self):
...
         if len(limits) > 1:
-            self.zPos.range = (min(limits), max(limits))
+            zrng = (min(limits), max(limits))
+            if not zrng[0] <= self.zPos.value <= zrng[1] or self.zPos.range == (0, 0):
+                # Completely new position => pick the middle as typically that's where the most interesting part is
+                new_z_pos = (zrng[0] + zrng[1]) / 2
+            else:
+                new_z_pos = None
+            self.zPos.range = zrng
+            if new_z_pos is not None:
```
K4rishma: Similar to above comment

K4rishma (changes requested):

src/odemis/gui/model/tab_gui_data.py
pieleric: No. it can only be updated after the range has been updated, otherwise it might not be valid.

src/odemis/gui/model/tab_gui_data.py
pieleric: Yes, because that's the special range when there is no z-stack. In such case, zPos.value == 0. This value might be valid, ie within the new range, so we have to detect this case specifically.

pieleric: I've added a comment to clarify it.

K4rishma (approved):

PR #3444, opened 2026-04-13
[feature][MSD-318] implement store annotated data framework

K4rishma:
Provides a thread-safe, non-blocking ``DataCollector.record()`` call that any Odemis module can invoke to capture a labelled data sample. CLI interface is implemented to download the data from the cloud storage.


src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,717 @@
...
+# Set explicitly only for custom S3-compatible storage.
+S3_ENDPOINT_URL = None
+S3_REGION = "eu-west-1"
+
+# Path to the S3 credentials key file (JSON with access_key / secret_key).
+_CREDENTIALS_PATH = "/usr/share/odemis/datacollector.key"
+
+# Default paths
+_CONF_DIR = os.path.join(os.path.expanduser("~"), ".config", "odemis")
+_DEFAULT_QUEUE_DIR = Path("/var/log/odemis/dc_queue")
```
pieleric: That directory cannot be created by the user. So it won't work if the data collection framework runs from the GUI (which is the most typical case). I'd suggest to use `~/.local/share/odemis/dc_queue`

src/odemis/gui/main.py
```diff
@@ -410,6 +414,27 @@ def on_level_va(self, log_level):
...
+    @call_in_wx_main
+    def _maybe_show_consent_dialog(self) -> None:
+        """Show consent dialog when consent is undecided and prompt is due."""
+        try:
+            if not self._data_collector.should_prompt_for_consent():
+                return
+
+            remind_days = self._data_collector.get_consent_remind_days()
+            dlg = ConsentDialog(parent=self.main_frame, remind_days=remind_days)
+            result = dlg.ShowModal()
```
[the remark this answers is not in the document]
K4rishma: It does define the method in datacollector

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,717 @@
...
+from dataclasses import dataclass, field
+from datetime import datetime, timedelta, timezone
+from pathlib import Path
+from typing import Optional
+
+try:
+    import boto3
+except ImportError:
+    logging.error("boto3 is required for S3 upload functionality; install with 'sudo apt install python3-boto3'")
+    raise
```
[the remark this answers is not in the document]
K4rishma: it is better to fail it, we will add it in the installation packages

debian/control
```diff
@@ -75,6 +75,8 @@ Depends: ${shlibs:Depends},
          python3-libusb1,
 # Needed for acq.fastem
          python3-shapely,
+# Needed for communicating with AWS S3
+         python3-boto3,
```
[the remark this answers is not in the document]
K4rishma: Not sure how to handle it

src/odemis/util/test/datacollector_test.py
```diff
@@ -0,0 +1,648 @@
...
+            had_pending = self._worker._process_pending_zips(self._queue_dir)
+        self.assertTrue(had_pending)
+        self.assertEqual(uploaded, ["oldest.zip", "middle.zip", "newest.zip"])
+        self.assertEqual(list(self._queue_dir.glob("*.zip")), [])
+
+
+class TestRealS3Integration(unittest.TestCase):
+    """Real S3 integration tests for Phase 2 upload workflow.
+
+    Tests use the credentials from ``_CREDENTIALS_PATH`` and upload to
```
tmoerkerken: No md in docstrings. I guess the rule is not picked up nicely @pieleric .

> # Style
> For Python code, always use type hints for function parameters and return types.
> Include docstrings for all functions and classes, following the reStructuredText style guide,
> but without the type information and **without using inline formatting markers or backticks.**

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+
+    event_name: str
+    schema_version: str
+    payload: dict
+    image_format: str = "TIFF"
+    submitted_at: float = field(default_factory=time.monotonic)
+
+
+def _serialize(item: _WorkItem, queue_dir: Path) -> Path:
+    """Serialise *item* into a ZIP archive and place it in *queue_dir*.
```
tmoerkerken: Markdown

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+def _upload(zip_path: Path, backend: S3UploadBackend) -> None:
+    """Upload *zip_path* with *backend* using the standard remote key."""
+    remote_key = f"{socket.gethostname()}/{zip_path.name}"
+    backend.upload(zip_path, remote_key)
+
+
+class _BackgroundWorker:
+    """Daemon thread that consumes :class:`_WorkItem` objects from a queue.
+
+    For each item it calls :func:`_enforce_queue_limit`, :func:`_serialize`,
```
tmoerkerken: Markdown

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+    ) -> None:
+        """Capture an annotated data sample at a software event.
+
+        Returns immediately (non-blocking). Serialisation and upload happen
+        asynchronously in a background thread.  If consent has not been
+        granted, this is a no-op.  This function never raises (beyond the
+        input validation below); all errors are logged and suppressed.
+
+        :param event_name: Human-readable event identifier, e.g.
+            ``"z_stack_acquired"``.  Must be a non-empty string.
```
tmoerkerken: Markdown

src/odemis/util/dc_fetch.py
```diff
@@ -0,0 +1,293 @@
...
+    )
+    parser.add_argument(
+        "--region",
+        dest="region",
+        help="Optional AWS region name for S3 client creation.",
+    )
+    return parser
+
+def iter_s3_objects(s3_client: Any, bucket: str, prefix: str) -> Iterator[Dict[str, Any]]:
+    """Iterate S3 objects under `prefix` using `list_objects_v2` pagination."""
```
tmoerkerken: Markdown

src/odemis/gui/win/consent.py
```diff
@@ -0,0 +1,70 @@
+# -*- coding: utf-8 -*-
```
tmoerkerken: Lacks license info

src/odemis/util/test/datacollector_test.py
```diff
@@ -0,0 +1,648 @@
+# -*- coding: utf-8 -*-
+"""
+Created on 11 March 2026
+
+Copyright © 2026 Delmic
+
```
tmoerkerken, line 8: Lacks author

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+    return {
+        "access_key": data["access_key"],
+        "secret_key": data["secret_key"],
+    }
+
+
+class DataCollectorConfig:
+    """Persistent configuration for the data-collection framework.
+
+    Backed by a ``configparser`` INI file at
```
tmoerkerken: Markdown

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+        return self._client
+
+    def upload(self, local_path: Path, remote_key: str) -> None:
+        """Upload *local_path* to *remote_key* in the configured bucket."""
+        client = self._get_client()
+        client.upload_file(str(local_path), self._bucket, remote_key)
+
+
+def _upload(zip_path: Path, backend: S3UploadBackend) -> None:
+    """Upload *zip_path* with *backend* using the standard remote key."""
```
tmoerkerken: Markdown

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+                "s3",
+                endpoint_url=self._endpoint_url,
+                region_name = self._region,
+                aws_access_key_id=self._access_key,
+                aws_secret_access_key=self._secret_key,
+            )
+        return self._client
+
+    def upload(self, local_path: Path, remote_key: str) -> None:
+        """Upload *local_path* to *remote_key* in the configured bucket."""
```
tmoerkerken: Markdown

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+        self._queue_dir = queue_dir
+        self._queue: queue.Queue = queue.Queue()
+        self._thread: Optional[threading.Thread] = None
+        self._lock = threading.Lock()
+        self._upload_backend: Optional[S3UploadBackend] = None
+        self._next_retry_at: float = 0.0
+        self._retry_delay: float = _INITIAL_RETRY_DELAY_SECONDS
+
+    def enqueue(self, item: _WorkItem) -> None:
+        """Add *item* to the processing queue and ensure the thread is alive.
```
tmoerkerken: Markdown

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+    def _pending_zip_paths(self, queue_dir: Path) -> list[Path]:
+        """Return pending ZIP files ordered oldest-first."""
+        if not queue_dir.exists():
+            return []
+        return sorted(queue_dir.glob("*.zip"), key=lambda p: p.stat().st_mtime)
+
+    def _process_pending_zips(self, queue_dir: Path) -> bool:
+        """Upload pending ZIP files from *queue_dir*.
+
+        Returns ``True`` when pending work existed (including backoff wait),
```
tmoerkerken: Markdown and wrong docstring style

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+        self._lazy_init()
+        if not self._init_ok:
+            return
+        self._config.consent = value  # type: ignore[union-attr]
+
+    def postpone_consent(self, remind_date: Optional[datetime] = None) -> None:
+        """Postpone consent prompt and clear current consent state.
+
+        :param remind_date: UTC datetime after which the consent prompt should
+            be shown again.  Defaults to ``_DEFAULT_CONSENT_REMIND_DELTA`` from
```
tmoerkerken: Markdown

tmoerkerken:
I think for such a big PR, a more elaborate description could help me review it a bit better.

src/odemis/gui/cont/menu.py
```diff
@@ -163,6 +169,32 @@ def __init__(self, main_data, main_frame):
             menu.Remove(main_frame.menu_item_show_correlation)
             main_frame.menu_item_show_correlation.Destroy()
 
+    def _append_data_sharing_menu_item(self, main_frame):
+        """Append and initialize Help menu checkbox for data sharing consent."""
```
tmoerkerken: No param, return and typing

src/odemis/gui/win/consent.py
```diff
@@ -0,0 +1,70 @@
...
+    def __init__(self, parent: wx.Window, remind_days: int) -> None:
+        """
+        Initialize the consent dialog asking the user to opt in, opt out, or be reminded later
+        about sharing anonymized data with Delmic.
+
+        :param parent: The parent window.
+        :param remind_days: Number of days to remind the user later.
+        """
+        title = "Share data with Delmic"
+        super().__init__(parent, wx.ID_ANY, title=title, size=(560, -1))
```
tmoerkerken: Positioning is now funky (never seen it like that in other software). I would expect the consent popup first (centered) and then any other popups, after you handled the consent popup.

Image

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+from dataclasses import dataclass, field
+from datetime import datetime, timedelta, timezone
+from pathlib import Path
+from typing import Optional
+
+try:
+    import boto3
+except ImportError:
+    logging.error("boto3 is required for S3 upload functionality; install with 'sudo apt install python3-boto3'")
+    raise
```
tmoerkerken, line 53: It does not fail graceful now. Odemis does not start when boto is not installed. This would be annoying when testing a dev branch on a costumer system.

src/odemis/gui/cont/menu.py
```diff
@@ -163,6 +169,32 @@ def __init__(self, main_data, main_frame):
...
+        help_menu = main_frame.menu_item_about.GetMenu()
+        if help_menu is None:
+            return None
+        help_menu.AppendSeparator()
+        item = help_menu.AppendCheckItem(wx.ID_ANY, "Share data with Delmic")
+        item.Check(self._data_collector.get_consent() is True)
+        return item
+
+    def refresh_consent_menu_item(self) -> None:
```
tmoerkerken: This method is now called one time. Since it is a single line, can't you just put the single line there where it is called?

src/odemis/gui/cont/menu.py
```diff
@@ -163,6 +169,32 @@ def __init__(self, main_data, main_frame):
...
+        without requiring a restart.
+        """
+        if self._consent_menu_item is not None:
+            self._consent_menu_item.Check(self._data_collector.get_consent() is True)
+
+    def _on_toggle_data_sharing(self, evt):
+        """Toggle data sharing consent from Help menu check item."""
+        enabled = self._consent_menu_item.IsChecked()
+        self._data_collector.set_consent(enabled)
+        evt.Skip()
```
tmoerkerken: What does this line do/prevent? Maybe comment?

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+Odemis is distributed in the hope that it will be useful, but WITHOUT ANY
+WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with
+Odemis. If not, see http://www.gnu.org/licenses/.
+
+Odemis Annotated Data Collection Framework.
+
+Provides a thread-safe, non-blocking ``DataCollector.record()`` call that any
```
tmoerkerken: markdown

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+# S3 bucket used for automated tests (not the production bucket).
+S3_TEST_BUCKET = "delmic-odemis-collect-test"
+
+# S3 endpoint URL — None means let boto3 resolve the regional endpoint automatically.
+# Set explicitly only for custom S3-compatible storage.
+S3_ENDPOINT_URL = None
+S3_REGION = "eu-west-1"
+
+# Path to the S3 credentials key file (JSON with access_key / secret_key).
+_CREDENTIALS_PATH = "/usr/share/odemis/datacollector.key"
```
tmoerkerken, line 73: Having this file unencrypted on disk (and showing this in our open source codebase) is not best practice. We could utilize system's keyring capabilities to store the credential encrypted. Or at least read the path from an environment variable. Both require installation steps to setup though.

src/odemis/util/dc_fetch.py
```diff
@@ -0,0 +1,293 @@
...
+from datetime import datetime, timezone
+from pathlib import Path
+from typing import Any, Dict, Iterator, List, Optional
+
+from odemis.util.datacollector import DataCollectorConfig, S3UploadBackend
+
+
+def parse_since_utc(value: str) -> datetime:
+    """Parse a date/datetime string to UTC-aware datetime.
+    :param value: ISO-8601 date (``YYYY-MM-DD``) or datetime
```
tmoerkerken: markdown

src/odemis/util/dc_fetch.py
```diff
@@ -0,0 +1,293 @@
...
+    # datetime.fromisoformat() does not accept the 'Z' suffix in Python < 3.11.
+    if text.endswith("Z"):
+        text = text[:-1] + "+00:00"
+    parsed = datetime.fromisoformat(text)
+    if parsed.tzinfo is None:
+        return parsed.replace(tzinfo=timezone.utc)
+    return parsed.astimezone(timezone.utc)
+
+def parse_key_timestamp_utc(key: str) -> Optional[datetime]:
+    """Parse `<event>-<YYYYMMDDTHHmmss>-<uuid>.zip` timestamp from key basename."""
```
tmoerkerken: no params/return docs

src/odemis/util/dc_fetch.py
tmoerkerken: no params/return docs

src/odemis/util/dc_fetch.py
```diff
@@ -0,0 +1,293 @@
...
+            yield item
+        if not response.get("IsTruncated"):
+            break
+        token = response.get("NextContinuationToken")
+        if not token:
+            logging.warning("S3 list_objects_v2 returned IsTruncated=True but no NextContinuationToken; stopping pagination.")
+            break
+
+def should_download_key(key: str, event_filter: Optional[str], since_utc: Optional[datetime]) -> bool:
+    """Return whether an S3 key should be downloaded by filters."""
```
tmoerkerken: no params/return docs

src/odemis/util/dc_fetch.py
```diff
@@ -0,0 +1,293 @@
...
+    hosts = [part.strip().strip("/") for part in value.split(",")]
+    return [host for host in hosts if host]
+
+def build_s3_client_from_config(
+    config: DataCollectorConfig,
+    bucket_override: Optional[str] = None,
+    endpoint_override: Optional[str] = None,
+    region_override: Optional[str] = None,
+) -> tuple[Any, str]:
+    """Build an S3 client using datacollector credentials with optional endpoint/bucket overrides."""
```
tmoerkerken: no params/return docs

src/odemis/util/dc_fetch.py
```diff
@@ -0,0 +1,293 @@
...
+def fetch_samples(
+    event_filter: Optional[str],
+    since_utc: Optional[datetime],
+    output_dir: Path,
+    host_filter: Optional[str] = None,
+    bucket_override: Optional[str] = None,
+    endpoint_override: Optional[str] = None,
+    region_override: Optional[str] = None,
+) -> Dict[str, int]:
+    """Fetch matching samples from S3 into output directory."""
```
tmoerkerken: no params/return docs

src/odemis/util/dc_fetch.py
```diff
@@ -0,0 +1,293 @@
...
+        "  odemis-dc-fetch --output ./downloads\n"
+        "  odemis-dc-fetch --event z_stack_acquired\n"
+        "  odemis-dc-fetch --since 2026-03-01\n"
+        "  odemis-dc-fetch --host meteor-5099\n"
+        "  odemis-dc-fetch --host meteor-5099,atlas-001,secom-22\n"
+        "  odemis-dc-fetch --bucket delmic-odemis-collect-test --region eu-west-1\n"
+        "  odemis-dc-fetch --since 2026-03-01T12:30:00 --event z_stack_acquired --output ./dc_samples"
+    )
+    parser = argparse.ArgumentParser(
+        description="Fetch data-collection ZIP samples from S3.",
```
tmoerkerken: I'm missing explanation on the intend of this script. Is it for Delmic personnel to inspect acquired data? How do we prevent customers using it to download other data? Is that explained somewhere?

util/odemis-dc-fetch.py
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
```
tmoerkerken, line 2: Missing license information

src/odemis/gui/cont/menu.py
```diff
@@ -42,14 +43,16 @@ class MenuController(object):
     tab controller.
     """
 
-    def __init__(self, main_data, main_frame):
+    def __init__(self, main_data, main_frame, data_collector):
```
tmoerkerken: Maybe a good moment to move typing into function definition

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+import numpy
+
+import odemis
+from odemis import model
+from odemis.dataio import hdf5
+from odemis.dataio import tiff
+
+
+# S3 bucket name — shared production bucket, created once by the dev team.
+S3_BUCKET = "delmic-odemis-collect"
```
tmoerkerken, line 62: I'm surprised we are locking ourselves into the Amazon infrastructure instead of going with a platform agnostic tool like [rclone](https://rclone.org/) ([python](https://pypi.org/project/rclone-python/)) for instance, which has a good track record on Linux. @tepals @pieleric.

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+    def __init__(self) -> None:
+        self.file_path = Path(_CONF_DIR) / self.file_name
+        self._cp = configparser.ConfigParser(interpolation=None)
+        self._lock = threading.Lock()
+        self._read()
+
+    def _read(self) -> None:
+        """Read the config file if it exists; otherwise leave defaults."""
+        if self.file_path.exists():
+            self._cp.read(str(self.file_path))
```
tmoerkerken: According to docs, you can just pass the pathlib object: https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.read

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+            consent_line = "consent = false"
+        else:
+            consent_line = "# consent = true"
+
+        if remind_val is not None:
+            remind_line = f"reminder_date = {remind_val.strftime('%Y-%m-%d')}"
+        else:
+            remind_line = "# reminder_date = "
+
+        content = (
```
tmoerkerken: This feels very manual and finicky, can't we use the configparser library for this?
```
config.set('general', '# Data sharing consent (true / false)')
config.set('general', 'consent', consent)

config.set('general', '# Date for reminder (YYYY-MM-DD)')
config.set('general', 'reminder_date', date)
```

Then we could also get rid of the if/else structure above.

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+            "#\n"
+            "# Date after which the consent dialog will be shown again (YYYY-MM-DD).\n"
+            f"{remind_line}\n"
+        )
+
+        with open(str(self.file_path), "w") as fh:
+            fh.write(content)
+        os.chmod(str(self.file_path), 0o600)
+
+    def _ensure_section(self, section: str) -> None:
```
tmoerkerken, line 192: docstrings

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+        """Ensure that *section* exists in the config, creating it if necessary."""
```
tmoerkerken: markdown

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+        if not self._cp.has_section(section):
+            self._cp.add_section(section)
+
+    @property
+    def consent(self) -> Optional[bool]:
+        """Return the consent state, or ``None`` if not yet set."""
```
tmoerkerken: no return docs adhering to our formatting rule, markdown in docstring

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+        try:
+            return self._cp.getboolean("general", "consent")
+        except (configparser.NoSectionError, configparser.NoOptionError):
+            return None
+        except ValueError:
+            return None
+
+    @consent.setter
+    def consent(self, value: bool) -> None:
```
tmoerkerken, line 214: docstring

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+    def clear_consent(self) -> None:
+        """Unset consent so it becomes undecided again."""
+        with self._lock:
+            self._ensure_section("general")
+            self._cp.remove_option("general", "consent")
+            self._write()
+
+    @property
+    def remind_date(self) -> Optional[datetime]:
+        """Return next reminder date as a UTC-aware datetime, or ``None`` when unset."""
```
tmoerkerken: Return param

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+                if parsed.tzinfo is None:
+                    return parsed.replace(tzinfo=timezone.utc)
+                return parsed.astimezone(timezone.utc)
+            except ValueError:
+                continue
+        return None
+
+    @remind_date.setter
+    def remind_date(self, value: Optional[datetime]) -> None:
+        """Persist next reminder UTC timestamp, or unset when ``None``."""
```
tmoerkerken: lacking param and has markdown

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+        elif remind_date.tzinfo is None:
+            remind_date = remind_date.replace(tzinfo=timezone.utc)
+        with self._lock:
+            self._ensure_section("general")
+            self._cp.remove_option("general", "consent")
+            self._cp.set("general", REMINDER_DATE_KEY, remind_date.astimezone(timezone.utc).isoformat())
+            self._write()
+
+    def should_prompt_for_consent(self) -> bool:
+        """Return whether the consent dialog should be shown now."""
```
tmoerkerken: return docstring

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+        if self.consent is not None:
+            return False
+        remind_after = self.remind_date
+        if remind_after is None:
+            return True
+        return datetime.now(timezone.utc) >= remind_after
+
+    def get_upload_backend(self) -> "S3UploadBackend":
+        """Return the configured upload backend instance."""
```
tmoerkerken: return docstring

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+    try:
+        payload_meta: dict = {}
+        extra_files: list = []  # list of (arcname, abs_path)
+
+        for key, value in item.payload.items():
+            if value is None or isinstance(value, (str, int, float, bool)):
+                payload_meta[key] = value
+
+            elif isinstance(value, numpy.ndarray):
+                if item.image_format.upper() == "HDF5":
```
tmoerkerken, line 344: I think you defined "HDF5" as a variable somewhere, but also, are you covering all combination of extensions?
```
>>> get_available_formats(os.O_WRONLY)
{'TIFF': ['.ome.tiff', '.ome.tif', '.tiff', '.tif'], 'Serialized TIFF': ['.0.ome.tiff'], 'HDF5': ['.h5', '.hdf5'], 'ImageJ Compatible TIFF': ['.ij.tiff']}
```

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+                extra_files.append((arc_name, abs_path))
+                payload_meta[key] = arc_name
+
+            else:
+                # Fallback: store string representation, guarding against
+                # __repr__/__str__ implementations that raise.
+                try:
+                    payload_meta[key] = str(value)
+                except Exception:
+                    # TODO add logging
```
tmoerkerken: Something to do for this PR or later?

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+                zf.write(str(abs_path), arc_name)
+
+        os.replace(str(tmp_zip), str(final_zip))
+        return final_zip
+
+    finally:
+        shutil.rmtree(str(tmp_dir), ignore_errors=True)
+
+
+def _enforce_queue_limit(queue_dir: Path) -> None:
```
tmoerkerken, line 414: Feels like the 10% should be a parameter

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+    For each item it calls :func:`_enforce_queue_limit`, :func:`_serialize`,
+    and :func:`_upload` in sequence.  After a successful upload the local ZIP
+    is deleted.  Exceptions are caught and logged so that the worker never
+    crashes the host application.
+
+    The thread is started lazily and restarted automatically if it dies.
+    """
+
+    def __init__(self, config: DataCollectorConfig, queue_dir: Path = _DEFAULT_QUEUE_DIR) -> None:
+        self._config = config
```
tmoerkerken, line 509: No docstring

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+        self._retry_delay = min(self._retry_delay * 2.0, _MAX_RETRY_DELAY_SECONDS)
+        logging.warning("DataCollector upload failed; retrying in %.0f s", delay)
+
+    def _reset_retry(self) -> None:
+        """Reset retry state after a successful upload."""
+        self._next_retry_at = 0.0
+        self._retry_delay = _INITIAL_RETRY_DELAY_SECONDS
+
+    def _pending_zip_paths(self, queue_dir: Path) -> list[Path]:
+        """Return pending ZIP files ordered oldest-first."""
```
tmoerkerken, line 562: No params

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+
+    def _process_work_item(self, item: _WorkItem) -> None:
+        """Serialize one work item and trigger upload processing."""
+        _enforce_queue_limit(self._queue_dir)
+        _serialize(item, self._queue_dir)
+        self._process_pending_zips(self._queue_dir)
+
+    def _run(self) -> None:
+        """Main loop: process items until the thread is stopped."""
+        while True:
```
tmoerkerken, line 610: Does this loop need to run without timeout, or is that built in some of the queue calls?

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+
+    def should_prompt_for_consent(self) -> bool:
+        """Return whether a consent dialog should be shown to the user."""
+        self._lazy_init()
+        if not self._init_ok:
+            return False
+        return self._config.should_prompt_for_consent()  # type: ignore[union-attr]
+
+    def set_consent(self, value: bool) -> None:
+        """Persist explicit user consent choice."""
```
tmoerkerken: Param

debian/control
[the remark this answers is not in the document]
pieleric: @K4rishma the code should stay Python 3.8 compatible. You should change the type annotation to be `List` instead of `list` and `Dict` instead of `dict`.

src/odemis/util/dc_fetch.py
pieleric: Indeed, this script is to be used only by the Delmic "data analysts". They have different access rights compared to the standard users. Standard users get a key (via the Odemis package, common to all users) that only give rights to upload data. They cannot list or read the data. 
The data analysts will have a different key (one per analyst) that gives full access to the data, and with which this script will work.

@K4rishma , I'd suggest to add documentation. I think you can add a chapter to the developer's documentation. It should describe the basic model of the data collection access rights, and explain how to use the odemis-dc-fetch tool. In particular, it should explain where the keys are stored, and in which format.

src/odemis/util/dc_fetch.py
```diff
@@ -0,0 +1,293 @@
...
+    if not name.endswith(".zip"):
+        return None
+    stem = name[:-4]
+    parts = stem.rsplit("-", 2)
+    if len(parts) != 3:
+        return None
+    return parts[0] or None
+
+def build_argument_parser() -> argparse.ArgumentParser:
+    """Build CLI argument parser for `odemis-dc-fetch`."""
```
pieleric: I think the separation between code and script is odd. It does make sense to keep some of the low level functions in a separate file (which makes it easier to test). However, all the argument parsing and main function should be in the `odemis-dc-fetch` script.

util/odemis-dc-fetch.py
pieleric, line 1: Don't place it in `scripts/` , which is a folder that contains example of how to write python scripts using Odemis. Instead, it should be in util/ .

src/odemis/gui/cont/menu.py
pieleric: Indeed, why letting the event "skip"? That's usually done when the event listener is not the main one, and you want the standard listeners to also receive the event. Here, that the only listener, so it doesn't seem useful.

src/odemis/gui/win/consent.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class ConsentDialog(wx.Dialog):
+    """Dialog asking user consent for data sharing."""
+
+    RESULT_OPT_IN = int(wx.NewIdRef())
+    RESULT_OPT_OUT = int(wx.NewIdRef())
+    RESULT_REMIND_LATER = int(wx.NewIdRef())
+
+    def __init__(self, parent: wx.Window, remind_days: int) -> None:
```
pieleric: I think that instead of creating your own window, you should use a standard MessageBox, with 3 buttons. It might also be easier to make sure it's properly centered.
Here is an example:
```
def show_custom_dialog(self, event):
        # 1. Initialize the standard MessageDialog with 3 buttons (Yes, No, Cancel)
        dialog = wx.MessageDialog(
            self, 
            message="What would you like to do with the modified document?", 
            caption="Unsaved Changes", 
            style=wx.YES_NO | wx.CANCEL | wx.ICON_QUESTION
        )
        
        # 2. Adjust the labels for the 3 buttons (Yes, No, Cancel order)
        dialog.SetYesNoCancelLabels(yes="Opt in", no="Opt out", cancel="Remind me in 30 days")
        
        # 3. Display the dialog box
        response = dialog.ShowModal()
        
        # 4. Handle the response using the standard built-in IDs
        if response == wx.ID_YES:
            ...
        elif response == wx.ID_NO:
            ...
        elif response == wx.ID_CANCEL:
            ...
```

src/odemis/util/datacollector.py
pieleric: We did ponder about this, and concluded that this was a good trade-off. Now, it's just a matter of copy-pasting the command, and you're done. This is only during the transition from to v3.8 to v3.9 (as boto3 will then be automatically installed when upgrading to v3.9). I think it's good enough.

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+    import boto3
+except ImportError:
+    logging.error("boto3 is required for S3 upload functionality; install with 'sudo apt install python3-boto3'")
+    raise
+import numpy
+
+import odemis
+from odemis import model
+from odemis.dataio import hdf5
+from odemis.dataio import tiff
```
pieleric:
```suggestion
from odemis.dataio import hdf5, tiff
```

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+    backend.upload(zip_path, remote_key)
+
+
+class _BackgroundWorker:
+    """Daemon thread that consumes :class:`_WorkItem` objects from a queue.
+
+    For each item it calls :func:`_enforce_queue_limit`, :func:`_serialize`,
+    and :func:`_upload` in sequence.  After a successful upload the local ZIP
+    is deleted.  Exceptions are caught and logged so that the worker never
+    crashes the host application.
```
pieleric: As it runs in a thread, the exceptions would never crash the application any way. I'd just say "Exceptions are caught and logged, but never directly shown to the user."

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+            logging.exception("DataCollector failed to initialize upload backend")
+            self._schedule_retry()
+            return True
+        for zip_path in pending:
+            try:
+                _upload(zip_path, backend)
+                zip_path.unlink(missing_ok=True)
+                self._reset_retry()
+            except Exception:
+                logging.exception("DataCollector upload failed for %s", zip_path.name)
```
pieleric: I don't think we should make it an "exception". The user has nothing to do with it. Let's make it a "warning" (and pass the backtrace with `exc_info=True`).

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+            try:
+                item = self._queue.get_nowait()
+            except queue.Empty:
+                item = None
+
+            if item is not None:
+                try:
+                    self._process_work_item(item)
+                except Exception:
+                    logging.exception(
```
pieleric: All these logging.exception should be converted to warnings, because it should be really not of concern to the user if some data collection failed.

src/odemis/util/datacollector.py
pieleric: Indeed, from Python 3.6, this should work. So you can assume it works.

src/odemis/util/datacollector.py
pieleric: Indeed. Just use the ConfigParser.

src/odemis/util/datacollector.py
pieleric: The idea is to provide the same key for all systems, part of the Odemis package. That's a special key, which can only upload data. So it's relatively safe. It's similar to what we already do with the support ticket API (special key which only allows to create a new bug report).

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,774 @@
...
+# Default paths
+_CONF_DIR = os.path.join(os.path.expanduser("~"), ".config", "odemis")
+_DEFAULT_QUEUE_DIR = Path("/var/log/odemis/dc_queue")
+
+_VALID_IMAGE_FORMATS = ("TIFF", "HDF5")
+_INITIAL_RETRY_DELAY_SECONDS = 30.0
+_MAX_RETRY_DELAY_SECONDS = 3600.0
+# Default delay before re-prompting for consent after postpone.
+_DEFAULT_CONSENT_REMIND_DELTA = timedelta(days=30)
+REMINDER_DATE_KEY = "reminder_date"
```
pieleric: No need to have a constant for this. That's the same at the other entries of the INI file.

src/odemis/util/datacollector.py
pieleric: That's a decision I've made. I prefer having a simple code that uses a single platform for now. The code relying on it is just in two files, and very short (see UploadBackend). If we decide to switch platform, we'll update that code. It's simpler than writing a generic code that support multiple platforms right now... while we don't even know how useful the entire "data collection" feature will be.

src/odemis/util/dc_fetch.py
```diff
@@ -0,0 +1,293 @@
...
+
+    return {
+        "listed": listed,
+        "matched": matched,
+        "downloaded": downloaded,
+        "skipped_existing": skipped_existing,
+        "failed": failed,
+    }
+
+def main(argv: Optional[List[str]] = None) -> int:
```
pieleric: Move to the script.

pieleric (changes requested):

src/odemis/util/datacollector.py
K4rishma: This is being used in 6 places, to avoid mistakes in writing the strings - a parameter identifier is added

src/odemis/util/datacollector.py
K4rishma: Yes, the queue calls have timing functions,
- it processes when only there is an item 
- there is a 1 second waiting time, if there is no item present to process
- there is repeated scheduling which increasingly sleeps between repeated tries

src/odemis/util/datacollector.py
K4rishma: That is true, I am not covering all the extensions. @pieleric do I need to cover all of them ?

src/odemis/util/datacollector.py
K4rishma: That is true, but for now keeping it hardcoded in MVP.

doc/develop/data-framework-setup-guide.rst
```diff
@@ -0,0 +1,431 @@
...
+   /usr/share/odemis/datacollector.key
+
+The file must contain exactly the following two keys:
+
+::
+
+   { "access\_key": "\<AWS\_ACCESS\_KEY\_ID\>","secret\_key":"\<AWS\_SECRET\_ACCESS\_KEY\>" }
+
+| Obtain the actual key values from a Delmic software engineer. Do NOT
+  commit
```
tmoerkerken: This newline seems off

doc/develop/data-framework-setup-guide.rst
```diff
@@ -0,0 +1,431 @@
...
+
+   | # Restrict access: readable only by root and the odemis process
+     user in bash (Can be copied together)
+   | sudo chmod 600 /usr/share/odemis/datacollector.key
+   | sudo chown root:root /usr/share/odemis/datacollector.key
+
+-——— TEST / DEVELOPER SETUP -———
+
+| Use the test IAM credentials (scoped to “delmic-odemis-collect-test”).
+  Ask a
```
tmoerkerken: This newline seems off

doc/develop/data-framework-setup-guide.rst
```diff
@@ -0,0 +1,431 @@
...
+| This guide covers:
+| Prerequisites
+| Installing the boto3 package
+| Creating and installing the credentials key file
+| Setting up the local queue directory (dc_queue)
+| Verifying the configuration file
+| Test-bucket vs production-bucket mode
+| Running the unit tests
+| Quick smoke-test with odemis-dc-fetch
+| Troubleshooting
```
tmoerkerken: This is not applying any formatting

doc/develop/data-framework-setup-guide.rst
```diff
@@ -0,0 +1,431 @@
...
+
+::
+
+   { "access\_key": "\<AWS\_ACCESS\_KEY\_ID\>","secret\_key":"\<AWS\_SECRET\_ACCESS\_KEY\>" }
+
+| Obtain the actual key values from a Delmic software engineer. Do NOT
+  commit
+| them to any repository or share them over unencrypted channels.
+
+-——— PRODUCTION SETUP -———
```
tmoerkerken: This is strange formatting, I think you should stick to https://docutils.sourceforge.io/docs/user/rst/quickstart.html#document-title-subtitle

doc/develop/data-framework-setup-guide.rst
```diff
@@ -0,0 +1,431 @@
...
+
+| Contact a software engineer who has access to the AWS IAM console
+| to get key pair.
+
+Steps:
+
+::
+
+   \# Create the file (replace the placeholder values with the real keys) line by line
+
```
tmoerkerken, line 90: The formatting renders very strange (I'm just checking with a standards rst viewer). This will not be simple to follow.

doc/develop/data-framework-setup-guide.rst
tmoerkerken, line 1: I commented on a few formatting problems, but I think this document has more of the same issues. Have a closer look with a rst viewer to see if you can fix them.

src/odemis/gui/cont/menu.py
```diff
@@ -163,6 +171,56 @@ def __init__(self, main_data, main_frame):
             menu.Remove(main_frame.menu_item_show_correlation)
             main_frame.menu_item_show_correlation.Destroy()
 
+    def _append_data_sharing_menu_item(self, main_frame: wx.Frame) -> wx.MenuItem | None:
```
tmoerkerken: I don't think union with `|` will work for Ubuntu 20.04

src/odemis/gui/main.py
```diff
@@ -30,8 +30,10 @@
 from odemis.gui.cont import acquisition
 from odemis.gui.cont.menu import MenuController
 from odemis.gui.cont.temperature import TemperatureController
+from odemis.gui.win.consent import ConsentDialog
```
tmoerkerken: This module does not seem to be part of your PR
```
    from odemis.gui.win.consent import ConsentDialog
ModuleNotFoundError: No module named 'odemis.gui.win.consent'
```

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,909 @@
...
+            self._ensure_section("general")
+            self._cp.set("general", "consent", "true" if value else "false")
+            self._cp.remove_option("general", REMINDER_DATE_KEY)
+            self._cp.remove_option("general", CONSENT_DATE_KEY)
+            self._write()
+
+    def clear_consent(self) -> None:
+        """
+        Unset consent so it becomes undecided again.
+         :return: None
```
tmoerkerken:
```suggestion
        :return: None
```

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,909 @@
...
+    try:
+        payload_meta: dict = {}
+        extra_files: list = []  # list of (arcname, abs_path)
+
+        for key, value in item.payload.items():
+            if value is None or isinstance(value, (str, int, float, bool)):
+                payload_meta[key] = value
+
+            elif isinstance(value, numpy.ndarray):
+                if item.image_format.upper() == "HDF5":
```
tmoerkerken, line 344: I think it is also more common to normalize with .lower() instead of .upper()

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,909 @@
...
+        :param event_name: Human-readable event identifier, e.g.
+         "z_stack_acquired".  Must be a non-empty string.
+        :param schema_version: Payload schema version string, e.g. "1.0".
+         Must be a non-empty string.
+        :param payload: Dict of arbitrary values.  Must be a dict.  Supported
+         value types:
+         - Python primitives (str, int, float, bool, None) — inlined in
+           metadata.json
+         - :class:odemis.model.DataArray / :class:numpy.ndarray —
+            exported as TIFF or HDF5 side-car
```
tmoerkerken: What does this mean? Never heard of it 😅

tmoerkerken (changes requested):

doc/develop/data-framework-setup-guide.rst
```diff
@@ -0,0 +1,431 @@
...
+   “<PRODUCTION_SECRET_ACCESS_KEY>”}
+
+   EOF
+
+   | # Restrict access: readable only by root and the odemis process
+     user in bash (Can be copied together)
+   | sudo chmod 600 /usr/share/odemis/datacollector.key
+   | sudo chown root:root /usr/share/odemis/datacollector.key
+
+-——— TEST / DEVELOPER SETUP -———
```
pieleric: This is now a standard RST formatting. Just use `---------------` under the line.

doc/develop/data-framework-setup-guide.rst
```diff
@@ -0,0 +1,431 @@
...
+| Use the test IAM credentials (scoped to “delmic-odemis-collect-test”).
+  Ask a
+| software engineer for the test key pair. The setup steps are
+  identical:
+
+::
+
+   \# Create the file (replace the placeholder values with the real keys) line by line
+
+3. | sudo tee /usr/share/odemis/datacollector.key > /dev/null << ‘EOF’
```
pieleric: This line (and the one above with sudo) seems completely off. I don't get why it starts with a number, why it has a `|`, and why the quotes are "fancy" quotes instead of being standard `'`.

doc/develop/data-framework-setup-guide.rst
```diff
@@ -0,0 +1,431 @@
...
+
+4. { “access_key”: “<TEST_ACCESS_KEY_ID>”,“secret_key”:
+   “<TEST_SECRET_ACCESS_KEY>”}
```
pieleric: Wrong quotes, weird line break.

```suggestion
{"access_key": "<TEST_ACCESS_KEY_ID>",
"secret_key":"<TEST_SECRET_ACCESS_KEY>"}
```
This probably could be written on a single line, and then the whole command would be simplified.

doc/develop/data-framework-setup-guide.rst
```diff
@@ -0,0 +1,431 @@
...
+| the PRODUCTION bucket by default. You must also set the environment
+  variable
+| described in Section 6 to redirect uploads to the test bucket.
+
+.. _section-1:
+
+Setting up the local queue directory (dc_queue)
+--------------------------------------------------
+
+| Skip for Production setup
```
pieleric: There is no reason to use `|` everywhere. You can write standard text.

doc/develop/data-framework-setup-guide.rst
```diff
@@ -0,0 +1,431 @@
...
+| -——— TEST / DEVELOPER SETUP -———
+| The framework stages serialised ZIP archives in a local directory
+  before
+| uploading them. The default path is:
+
+::
+
+   /.local/share/odemis/dc\_queue
```
pieleric: For most the paths, you're missing the `$HOME` (or `~`) part.
`~/.local/share/odemis/dc\_queue`

doc/develop/data-framework-setup-guide.rst
```diff
@@ -0,0 +1,431 @@
...
+  process.
+
+| For a standard Odemis installation /.local/share/odemis/ is already created
+  by the
+| package post-install script. If that is not the case, create it
+  manually:
+
+::
+
+   sudo mkdir \-p /.local/share/odemis/dc\_queue
```
pieleric: Why `\` before `-p`?

src/odemis/util/datacollector.py
pieleric: No need to support more than 2 formats. We are not here pleasing the user, just the Delmic engineers. Simpler is better here.
However, instead of having twice the same code for HDF5 and TIFF, you can do:
```
if item.image_format.upper() == "HDF5":
   exporter = hdf5
elif item.image_format.upper() == "TIFF":
    exporter = tiff
else:
   raise ValueError("Unknown exporter ...")

ext = exporter.EXTENSIONS[0]
arc_name = f"{key}.{ext}"
:
```

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,909 @@
...
+    limit = usage.total * 0.10  # 10 % of partition
+    total_size = sum(p.stat().st_size for p in zips)
+
+    while total_size > limit and zips:
+        oldest = zips.pop(0)
+        try:
+            size = oldest.stat().st_size
+            oldest.unlink()
+            total_size -= size
+            logging.warning("Queue limit exceeded: removed oldest sample %s", oldest.name)
```
pieleric: Shouldn't be a warning. Just `info`.

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,909 @@
...
+        except (configparser.NoSectionError, configparser.NoOptionError):
+            return None
+        except ValueError:
+            return None
+
+    @consent.setter
+    def consent(self, value: bool) -> None:
+        """
+        Set consent to true or false, and clear any reminder date and consent expiry.
+        :param value: True to opt in, False to opt out.
```
pieleric, line 218: the setter is not equivalent to the getter. The getter returns `None` when not set, and the setter set the value to `False` when receiving `None`. => Handle passing `None`, and do what is currently in `clear_content`.

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,909 @@
...
+                    parsed = datetime.fromisoformat(value)
+                if parsed.tzinfo is None:
+                    return parsed.replace(tzinfo=timezone.utc)
+                return parsed.astimezone(timezone.utc)
+            except ValueError:
+                continue
+        return None
+
+    @remind_date.setter
+    def remind_date(self, value: Optional[datetime]) -> None:
```
pieleric: Please remove all support for the remind date. For now, it's unused, and having 2 dates to handle make the whole code so confusing to read.

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,909 @@
...
+            return None
+        for fmt in ("%Y-%m-%d", None):
+            try:
+                if fmt:
+                    parsed = datetime.strptime(value, fmt)
+                else:
+                    parsed = datetime.fromisoformat(value)
+                if parsed.tzinfo is None:
+                    return parsed.replace(tzinfo=timezone.utc)
+                return parsed.astimezone(timezone.utc)
```
pieleric: Please, keep things simple. Only support a single format of date. It should be the same format as we use for the odemis advanced feature.

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,909 @@
...
+
+    def should_prompt_for_consent(self) -> bool:
+        """
+        Return whether a consent dialog should be shown to the user.
+        :return: True if the consent dialog should be shown, False otherwise.
+        """
+        self._lazy_init()
+        if not self._init_ok:
+            return False
+        return self._config.should_prompt_for_consent()  # type: ignore[union-attr]
```
pieleric: Drop. We don't prompt.

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,909 @@
...
+                logging.warning(
+                    "DataCollector failed to initialise; all record() calls will be no-ops."
+                )
+
+    def get_consent(self) -> Optional[bool]:
+        """Return current consent state from configuration."""
+        self._lazy_init()
+        if not self._init_ok:
+            return None
+        return self._config.consent  # type: ignore[union-attr]
```
pieleric: I think this function should be a "high level" function. It says whether *right now* the consent is allowed or not. So if the consent is True, it should also check the expiry date of consent, and automatically set it to false in such case. 
This way, all the other code that check for the expiry date wouldn't be needed.

src/odemis/util/datacollector.py
```diff
@@ -0,0 +1,909 @@
...
+        When days == 1, the remaining time at the moment record() is called
+        will be ≤ 1 day, so 100% collection probability applies throughout.
+        When days > 1, the default 10% probability applies until the final day.
+
+        :param days: Number of days the temporary consent is active.
+        """
+        self._lazy_init()
+        if not self._init_ok:
+            return
+        consent_date = datetime.now(timezone.utc) + timedelta(days=days)
```
pieleric: It would probably be simpler to store the date in local time, and round it to a day, instead of having to store the complete time.

pieleric (changes requested):

src/odemis/util/datacollector.py
K4rishma: This means additional data saved along the main files

K4rishma: Will just remove that work to avoid confusion

src/odemis/util/dc_fetch.py
```diff
@@ -0,0 +1,245 @@
...
+            break
+        token = response.get("NextContinuationToken")
+        if not token:
+            logging.warning("S3 list_objects_v2 returned IsTruncated=True but no NextContinuationToken; stopping pagination.")
+            break
+
+def should_download_key(key: str, event_filter: Optional[str], since_utc: Optional[datetime]) -> bool:
+    """
+    Return whether an S3 key should be downloaded by filters.
+    :param key: S3 object key.
```
pieleric: Please add an example of such "key". Its format matters very much in the rest of the code. As I understand it's the "path", right? An example would clarify a lot.

src/odemis/util/dc_fetch.py
```diff
@@ -0,0 +1,245 @@
...
+    :param endpoint_override: Optional S3 endpoint URL override.
+    :param region_override: Optional AWS region name override.
+    :return: Tuple of (Boto3 S3 client, bucket name).
+    """
+    backend = config.get_upload_backend()
+    if not isinstance(backend, S3UploadBackend):
+        raise RuntimeError("Only S3 backend is supported for retrieval.")
+    endpoint_url = endpoint_override or backend._endpoint_url  # pylint: disable=protected-access
+    bucket = bucket_override or backend._bucket  # pylint: disable=protected-access
+    import boto3
```
pieleric: Don't import here. Move to the top of the module.

doc/develop/data-framework-setup-guide.rst
```diff
@@ -0,0 +1,431 @@
...
+   python3 -c "import boto3; print(boto3.__version__)"
+
+If the command prints a version string (e.g. ``1.28.0``) the package is ready.
+If it raises ``ImportError``, re-run the ``apt install`` command above.
+
+
+Installing the Credentials Key File
+-------------------------------------
+
+The framework reads S3 credentials from a single JSON file at: /usr/share/odemis/datacollector.key
```
pieleric, line 61: We need to have a way to package this key. If I recall correctly, the idea was that we store this file in Google Drive, and add it to the package when building the debian file. As the debian file is open source, and freely available, it'll still be possible for "anyone" to access it, but it's less likely to be picked up by automatic crawlers on github code. 
Please add the code to the build scripts (`util/release-odemis` and `debian/`) to check that the file is present, and package it.

doc/develop/data-framework-setup-guide.rst
```diff
@@ -0,0 +1,431 @@
...
+
+The file must contain exactly the following two keys:
+
+.. code-block:: json
+
+   { "access_key": "<AWS_ACCESS_KEY_ID>", "secret_key": "<AWS_SECRET_ACCESS_KEY>" }
+
+Obtain the actual key values from a Delmic software engineer. Do **not** commit
+them to any repository or share them over unencrypted channels.
```
pieleric: I think it should be safe to share them on unencrypted channels. The keys only allowed to upload data, not retrieve it. It shouldn't really be worse than knowing our email address and sending us big attachment files. IIRC, the main reason for not committing them is that there are scrappers for public github code that look for such AWS key, and so they would still end constantly try (and fail) to read data. Better avoid it. Also, this might trigger some standard security check warnings in github.

doc/develop/data-framework-setup-guide.rst
```diff
@@ -0,0 +1,431 @@
...
+
+.. code-block:: bash
+
+   python3 -c "
+   from odemis.util.datacollector import DataCollectorConfig
+   cfg = DataCollectorConfig()
+   cfg.clear_consent()
+   print('Consent cleared.')
+   "
+
```
pieleric: Way too detailed. I don't expect anyone needs this, before this code is outdated. Please drop.

doc/develop/data-framework-setup-guide.rst
```diff
@@ -0,0 +1,431 @@
...
+If ``failed`` is non-zero, check the log output for ``AccessDenied`` or network
+errors and verify the key file credentials.
+
+
+Troubleshooting
+----------------
+
+**Problem:** ``boto3 ImportError`` when starting Odemis.
+
+**Solution:** ``sudo apt install python3-boto3``
```
pieleric: Drop: the error should be describing the fix already.

doc/develop/data-framework-setup-guide.rst
```diff
@@ -0,0 +1,431 @@
...
+
+**Cause:** Odemis crashed mid-write.
+
+**Solution:** Stop Odemis, then:
+
+.. code-block:: bash
+
+   rm -f ~/.local/share/odemis/dc_queue/*.tmp
+
+The framework also cleans these up automatically on the next startup.
```
pieleric: Drop: if the frameworks take care of it, then no need to describe how to fix it.

doc/develop/data-framework-setup-guide.rst
```diff
@@ -0,0 +1,431 @@
...
+----
+
+**Problem:** ``TEST_DATACOLLECTION=1`` is set but uploads still go to
+production.
+
+**Solution:** This cannot happen — the environment variable is read at the
+moment ``get_upload_backend()`` is called, which is lazy (first upload attempt).
+Verify the variable is exported in the same shell/environment that runs the
+Odemis process.
+
```
pieleric: Drop: this cannot happen.

doc/develop/data-framework-setup-guide.rst
```diff
@@ -0,0 +1,431 @@
...
+oldest-first once connectivity is restored.
+
+----
+
+**Problem:** ``Queue limit exceeded: removed oldest sample`` appears in the log.
+
+**Cause:** The queue directory has grown beyond 10 % of the partition.
+
+**Solution:** Check disk space with ``df -h ~/.local/share/odemis/dc_queue``.
+Consider moving ``~/.local/share/odemis/`` to a larger partition, or
```
pieleric: I don't think we should move this to a larger partition. Drop this advice.

src/odemis/util/dc_fetch.py
```diff
@@ -0,0 +1,245 @@
...
+    :param region_override: Optional AWS region name override.
+    :return: Dictionary with counts of listed, matched, downloaded, skipped, and failed samples.
+    """
+    cfg = DataCollectorConfig()
+    s3_client, bucket = build_s3_client_from_config(
+        cfg,
+        bucket_override=bucket_override,
+        endpoint_override=endpoint_override,
+        region_override=region_override,
+    )
```
pieleric, line 291: I don't think it makes sense to build the config from the Odemis config. The data analyst will have different credentials from upload part of Odemis. So they *cannot* share the same key/secret. Instead, I'd suggest to rely on a different `.ini` file, dedicated to dc_fetch. It should default to the `datacollector.S3_*` constants. For the secret and key, if it's not present, make sure the .ini is created with the right entries for key and secret, and report an error to the user explaining they have to edit the file (with the full path).

Hopefully, this should be almost as simple as `aws configure` to use for the first time:
* Run the script 
  -> it creates the .ini file, with the entries `key=` and `secret=` empty
  -> it tells the user to edit the file
* The user edits the file
* Run the script again
  -> it works

pieleric (changes requested):

src/odemis/util/dc_fetch.py
K4rishma: Should not this INI file be cleaned up after use if used on client's PC?

tmoerkerken:
During your presentation we identified two unknowns:

- [ ] Find out if we upload 10 or 20% of the cases
- [ ] Find out if the upload chance is per session or per record upload call

Did you already find out?

K4rishma:
> During your presentation we identified two unknowns:
> 
> * [ ]  Find out if we upload 10 or 20% of the cases
> * [ ]  Find out if the upload chance is per session or per record upload call
> 
> Did you already find out?

Hey Tim, these questions would be relevant for the implementation of this framework which is included in this PR https://github.com/delmic/odemis/pull/3451 

I have already noted them and will check out for that PR.

doc/develop/data-framework-setup-guide.rst
```diff
@@ -0,0 +1,395 @@
...
+.. code-block:: bash
+
+   # Create the file — replace placeholder values with the real keys
+   sudo tee /usr/share/odemis/datacollector.key > /dev/null << 'EOF'
+   { "access_key": "<PRODUCTION_ACCESS_KEY_ID>", "secret_key": "<PRODUCTION_SECRET_ACCESS_KEY>" }
+   EOF
+
+   # Restrict access: readable only by root and the odemis process user
+   sudo chmod 600 /usr/share/odemis/datacollector.key
+   sudo chown root:root /usr/share/odemis/datacollector.key
```
pieleric: No. That's incorrect. It should be accessible to the GUI. Drop this.

doc/develop/data-framework-setup-guide.rst
```diff
@@ -0,0 +1,395 @@
...
+.. code-block:: bash
+
+   # Create the file — replace placeholder values with the real keys
+   sudo tee /usr/share/odemis/datacollector.key > /dev/null << 'EOF'
+   { "access_key": "<TEST_ACCESS_KEY_ID>", "secret_key": "<TEST_SECRET_ACCESS_KEY>" }
+   EOF
+
+   # Restrict access: readable only by root and the odemis process user
+   sudo chmod 600 /usr/share/odemis/datacollector.key
+   sudo chown root:root /usr/share/odemis/datacollector.key
```
pieleric: Drop

doc/develop/data-framework-setup-guide.rst
```diff
@@ -0,0 +1,395 @@
...
+
+
+Test / Developer Setup
+~~~~~~~~~~~~~~~~~~~~~~
+
+Use the test IAM credentials scoped to ``delmic-odemis-collect-test``. Ask a
+software engineer for the test key pair. The setup steps are identical:
```
pieleric: I recall we use the same key for the testing. Isn't it the case? If so, this paragraph can be dropped, and all the info is in the `test-bucket-mode` section.

doc/develop/data-framework-setup-guide.rst
```diff
@@ -0,0 +1,395 @@
...
+
+Contact a software engineer who has access to the AWS IAM console to get the
+key pair, then run the following commands line by line:
+
+.. code-block:: bash
+
+   # Create the file — replace placeholder values with the real keys
+   sudo tee /usr/share/odemis/datacollector.key > /dev/null << 'EOF'
+   { "access_key": "<PRODUCTION_ACCESS_KEY_ID>", "secret_key": "<PRODUCTION_SECRET_ACCESS_KEY>" }
+   EOF
```
pieleric, line 89: Please add a copy of this file in the Google Drive, in this folder: https://drive.google.com/drive/u/0/folders/1fUZcoRVmtP-01bh_7uUFjYZSJg5gkL0q . This way it's easy to copy it.

util/release-odemis
```diff
@@ -34,6 +34,12 @@ fi
 
 cd ~/development/odemis
 
+# Ensure datacollector key is present before release operations start.
+if [ ! -f install/linux/usr/share/odemis/datacollector.key ]; then
```
pieleric: What's important is that the file is present in ~/development/pkg-native/odemis/ . The development/odemis/ folder doesn't need to have the file.
=> Do a single check, very early in the script, for ~/development/pkg-native/odemis/install/linux/usr/share/odemis/datacollector.key

util/release-odemis
```diff
@@ -106,6 +112,12 @@ git fetch
 git checkout $VBRANCH
 git pull
 
+# Check again in the build tree that the packaged key file exists.
+if [ ! -f install/linux/usr/share/odemis/datacollector.key ]; then
```
pieleric: Please add a line to .gitignore, to make sure no one add this file.

debian/rules
```diff
@@ -10,6 +10,10 @@ export PYBUILD_DISABLE=test
 	dh $@ --with python3 --with sphinxdoc --buildsystem=pybuild
 
 override_dh_auto_install:
+	@if [ ! -f install/linux/usr/share/odemis/datacollector.key ]; then \
+		echo "ERROR: Missing install/linux/usr/share/odemis/datacollector.key"; \
+		exit 1; \
+	fi
```
pieleric, line 15: We have a script on the CI that check that building the package works. However, now that the key is missing, the build will fail. It only runs on master, after a PR merge. So that's why it didn't fail yet.

So here, I'd suggest to add just a warning, and continue. Alternatively, you could edit `.github/workflows/build_debian.yml` to create a fake datacollector.key file.

pieleric (changes requested):

tmoerkerken (approved):

pieleric (approved):

PR #3448, opened 2026-04-20
[MSD-530][feat] driver tucsen: add support to switch between gains

pieleric:
We typically only need the HDR gain... but for testing/debugging it
turns out it's handy to have the other gains.

### DIFF
--- a/src/odemis/driver/test/cam_test_abs.py
+++ b/src/odemis/driver/test/cam_test_abs.py
@@ -201,6 +201,29 @@ class VirtualTestCam(metaclass=ABCMeta):
         logging.debug("Got res of %s", da.shape)
         self.assertEqual(da.shape, exp_res)
 
+    def test_change_gain(self):
+        """
+        Try various gains, and check acquisition works with each of them
+        """
+        if not model.hasVA(self.camera, "gain") or self.camera.gain.readonly:
+            self.skipTest("Camera doesn't support setting gain")
+
+        if hasattr(self.camera.gain, "range"):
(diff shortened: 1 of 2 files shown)


nandishjpatel (approved):

nandishjpatel (approved):

src/odemis/driver/test/cam_test_abs.py
```diff
@@ -201,6 +201,29 @@ def test_change_settings(self):
...
+            gain_values = [gain_range[0], gain_range[1]]
+        elif hasattr(self.camera.gain, "choices"):
+            gain_values = self.camera.gain.choices
+        else:
+            self.skipTest("Camera doesn't support setting gain")
+
+        self.camera.exposureTime.value = 0.09
+        for g in gain_values:
+            self.camera.gain.value = g
+            actual_gain = self.camera.gain.value
```
tmoerkerken, line 222: Do we expect the getter to return a different value than we used for the setter? Maybe comment on it, since otherwise I'd say this line is not needed if we adjust the loop to `for actual_gain in gain_values`.

tmoerkerken (approved):

src/odemis/driver/test/cam_test_abs.py
pieleric: Yes, generally, the hardware might not accept directly the requested value, and so the setter (slightly) adjust it to an acceptable value. That's actually not the case with the Tucsen, but this code is generic for all cameras, so I'm writing it a little bit defensively.

PR #3449, opened 2026-04-20
[MSD-548][fix] posture manager TFS3: bring back option to use old choreography

pieleric:
Some SEMs don't like going to a little low Z before moving. So, to keep
compatibility with the current systems, revert to use the old, more direct, choreography by
default.
The new choreograph can be activated by adding explicitly a z_low
in the CALIB of the TFSv3 stage-bare metadata.

### DIFF
--- a/install/linux/usr/share/odemis/sim/meteor-tfs3-sim.odm.yaml
+++ b/install/linux/usr/share/odemis/sim/meteor-tfs3-sim.odm.yaml
@@ -40,7 +40,7 @@
         SAMPLE_CENTERS: {"GRID 1": {'x': 0, 'y': 0, 'z': 32.e-3}, "GRID 2": {'x': 2.98e-3, 'y': 2.46e-3, 'z': 32.e-3}},
         CALIB: {
             "version": "tfs_3",
-            "z_low": 27.e-3,
+#            "z_low": 27.e-3,
             "dx": 0.0506252, "dy": 0.0049832,  # mirroring values between SEM - METEOR
             "pre-tilt": 0.6108652381980153,  # rad, 35°
             "use_linked_sem_focus_compensation": false,  # if True, use SEM-Eucentric-Focus
--- a/src/odemis/acq/move.py
+++ b/src/odemis/acq/move.py
@@ -898,10 +898,6 @@ class MeteorPostureManager(MicroscopePostureManager):
         return self._transformFromSEMToMilling(sem_pos)
 
 
-# Minimum stage bare z height required to switch between FM and SEM postures
-Z_LOW = 0.027  # m, safe value provided by TFS on a Hydra Bio CX system which should also be compatible on other systems
-
-
 class MeteorTFS1PostureManager(MeteorPostureManager):
     def __init__(self, microscope):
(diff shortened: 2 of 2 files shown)


src/odemis/acq/move.py
```diff
@@ -1141,7 +1141,7 @@ def _doCryoSwitchSamplePosition(self, future, target):
                 elif (type(self) == MeteorTFS3PostureManager):
                     stage_md = self.stage.getMetadata()
                     md_calib = stage_md[model.MD_CALIB]
-                    z_low = md_calib.get("z_low", Z_LOW)  # safe z to achieve before switching to SEM posture
+                    z_low: float = md_calib.get("z_low")  # safe z to achieve before switching to SEM posture
```
tmoerkerken: Technically it should be this I think:
```suggestion
                   z_low: Optional[float] = md_calib.get("z_low")  # safe z to achieve before switching to SEM posture
```

install/linux/usr/share/odemis/sim/meteor-tfs3-sim.odm.yaml
```diff
@@ -40,7 +40,7 @@
         SAMPLE_CENTERS: {"GRID 1": {'x': 0, 'y': 0, 'z': 32.e-3}, "GRID 2": {'x': 2.98e-3, 'y': 2.46e-3, 'z': 32.e-3}},
         CALIB: {
             "version": "tfs_3",
-            "z_low": 27.e-3,
+#            "z_low": 27.e-3,
```
tmoerkerken, line 43: The presence/absence of this value now entirely changes the choreography. My intuition would be that if there was no z_low provided, we just skip that step, but now we change the translation/rotation order in addition. Not quite sure how to improve, but this should be well documented.

pieleric: Yes, indeed, it's not intuitive. I have a separate task to update the (installation) manual.

src/odemis/acq/move.py
```diff
@@ -1154,31 +1154,43 @@ def _doCryoSwitchSamplePosition(self, future, target):
...
-                    sub_moves.append((self.stage, target_pos))
+                        if "z_low" in md_calib:
+                            # Stage switching based on Hydra Bio TFS assessment
+                            # If current z is more than safe value, move it to safe z value else skip it.
+                            # move x and y
+                            # move all to final position (outside if else block to take other switching postures combinations)
+                            if current_pos["z"] > z_low:
+                                sub_moves.append((self.stage, {"z": z_low}))
+                            sub_moves.append((self.stage, filter_dict({'x', 'y'}, target_pos)))
+                            sub_moves.append((self.stage, filter_dict({'rx', 'rz'}, target_pos)))
```
K4rishma: This line is not there in the new choreo with safe stage z, it is taken care all together in the next line

K4rishma (changes requested):

src/odemis/acq/move.py
pieleric: Thanks, I mistook in the copy-pasting. Good catch!

PR #3451, opened 2026-04-22
[feature][MSD-506] Meteor store annoted data

K4rishma:
Builds on PR of Building the data framework https://github.com/delmic/odemis/pull/3444 

**Overview**

Introduces per-project sampling for cryo feature data collection. When a project is opened or created, a single random decision is propagates from the data collection framework and applies uniformly to all features created during that session. Features loaded from a previous session are immediately excluded from collection.

---

**How it works**

**Single decision per project (`cryo_chamber_tab.py`)**  
`_change_project_conf()` is called every time a project is opened or created. It makes one random draw (`random.random() \< PROBABILITY`) and stores the result as `main.features_collectable` — a lightweight dynamic attribute on the main GUI data model (not a formal VA, never persisted to disk).

**All new features inherit the decision (`tab_gui_data.py`)**  
`CryoGUIData.add_new_feature()` reads `getattr(self.main, "features_collectable", False)` and passes it as `collect=` when constructing each `CryoFeature`. Every feature created in the same session therefore shares the same flag value — either all are eligible for collection or none are.

**Loaded features are excluded (`cryo_chamber_tab.py`)**  
In `_load_project_data()`, after features are read from `features.json`, every loaded feature is immediately reset to `collect=False` before being assigned to the model. Features from a prior session were either already collected or never selected; the new session's sampling decision applies only to features created going forward.

**`CryoFeature.collect` flag (`feature.py`)**  
The `collect` parameter on `CryoFeature.__init__` is a plain `bool`, defaulting to `False`. The flag is serialised into `features.json` and survives a save/load cycle. If the key is absent in loaded JSON (older data), it defaults to `False`.


src/odemis/acq/feature.py
```diff
@@ -471,6 +486,161 @@ def _create_fibsem_filename(filename: str, acq_type: str) -> str:
...
+    :param project_dir: Optional project directory path.  When provided and
+        feature.streams is empty, streams are loaded from disk first.
+    """
+    if not feature.collect:
+        return
+
+    try:
+        _dc = DataCollector()
+        if not _dc.get_consent():
+            return
```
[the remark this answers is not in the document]
pieleric: @K4rishma currently, gui/main.py creates a _data_collector, and keeps it for the rest of the GUI lifetime. I'd suggest to move it to the MainGUIData(). Then, pretty much every controller will be able to access it. You can have a single DataCollector instantiated. You can pass from the caller to this function.

src/odemis/acq/feature.py
```diff
@@ -396,6 +406,159 @@ def _create_fibsem_filename(filename: str, acq_type: str) -> str:
...
+        fm_zstacks: List = []
+        fm_images: List = []
+        for s in feature_streams:
+            if isinstance(s, StaticFluoStream):
+                if hasattr(s, "zIndex"):
+                    fm_zstacks.append(s)
+                else:
+                    fm_images.append(s)
+
+        def _get_raw(stream: Stream) -> Optional[model.DataArray]:
```
pieleric, line 486: I don't think there is a case where there is a stream with an image, but without .raw. Do you have something in mind? If so, it would help to add it as a comment. Otherwise, I think you can simplify and return .raw, and None if it's empty.

src/odemis/gui/cont/tabs/cryo_chamber_tab.py
```diff
@@ -373,6 +375,18 @@ def _change_project_conf(self, new_dir):
         self.conf.pj_ptn, self.conf.pj_count = guess_pattern(new_dir)
         self.txt_projectpath.Value = os.path.basename(self.conf.pj_last_path)
         self.tab_data_model.main.project_path.value = new_dir
+        # Decide once per project whether features created during this session are
+        # eligible for data collection.  Stored as a dynamic attribute — not part
+        # of the formal model — and read by add_new_feature via getattr.
+        probability = DataCollector().probability
```
pieleric: Use the global data collector.

src/odemis/gui/model/tab_gui_data.py
```diff
@@ -341,7 +341,8 @@ def add_new_feature(self, stage_position: Dict[str, float],
             else:
                 md = self.main.focus.getMetadata()
                 fm_focus_position = md[model.MD_FAV_POS_ACTIVE]
-        feature = CryoFeature(f_name, stage_position, fm_focus_position)
+        features_collectable = getattr(self.main, "features_collectable", False)
```
pieleric: It should be an attribute on CryoMainGUIData, and not require any protection. 
Maybe adjust the name to make it more clear like "should_collect_features".

src/odemis/gui/cont/tabs/cryo_chamber_tab.py
```diff
@@ -373,6 +375,18 @@ def _change_project_conf(self, new_dir):
...
+        self.tab_data_model.main.features_collectable = (
+            random.random() < probability
+        )
```
pieleric: Write it on a single line.

src/odemis/gui/cont/features.py
```diff
@@ -49,6 +53,10 @@
 
 SUPPORTED_POSTURES = [Posture.SEM_IMAGING, Posture.FM_IMAGING, Posture.MILLING, Posture.FIB_IMAGING, Posture.FIB_VIEW_FM]
 
+# Maximum distance (in metres) within which another feature is considered "nearby"
+# for the feature-deletion data-collection trigger.
+_NEARBY_FEATURE_DISTANCE_M = 100e-6
```
pieleric: Please don't put `_` before the name of the constants.

src/odemis/gui/cont/features.py
```diff
@@ -104,6 +115,9 @@ def __init__(self, tab_data, panel, tab, mode: guimod.AcquiMode):
             self._panel.btn_feature_save_position.Show(LICENCE_MILLING_ENABLED)
             self.pm.current_posture.subscribe(self._on_posture_change)
 
+        # Track previous posture so we can detect FM → SEM/FIB transitions.
+        self._prev_posture = self.pm.getCurrentPostureLabel() if self.pm else None
```
pieleric: No need to protect when `self.pm` is None. It's always present.

src/odemis/gui/cont/features.py
```diff
@@ -78,6 +86,9 @@ def __init__(self, tab_data, panel, tab, mode: guimod.AcquiMode):
         self._feature_status_va_connector = None
         self._feature_z_va_connector = None
 
+        # Feature whose status VA we are subscribed to for data-collection triggering.
+        self._status_collect_feature: Optional[CryoFeature] = None
```
pieleric: I find the name hard to understand. Maybe "_feature_for_collection", or something of this sort.

src/odemis/gui/cont/features.py
```diff
@@ -449,3 +478,92 @@ def _on_ctrl_feature_z_change(self):
...
+        overview_streams = self._tab_data_model.overviewStreams.value
+        project_dir = self._tab_data_model.conf.pj_last_path
+
+        def _run():
+            collect_feature_data(feature, overview_streams=overview_streams, project_dir=project_dir)
+
+        t = threading.Thread(target=_run, name="FeatureDataCollection", daemon=True)
+        t.start()
+
+    def _collect_eligible_features_in_thread(self) -> None:
```
pieleric: I think you could simplify this code by merging `_collect_eligible_features_in_thread()` and `_collect_feature_in_thread()` into a single function `_collect_features_in_thread(features)`. It takes a list of features, and skip the ones which are not marked for collection (ie, `feature.collect` is false).
* For _collect_eligible_features_in_thread(), replace by `_collect_features_in_thread( self._tab_data_model.main.features.value)`
* For _collect_feature_in_thread(), replace by `_collect_features_in_thread([f])`

src/odemis/gui/cont/features.py
```diff
@@ -449,3 +478,92 @@ def _on_ctrl_feature_z_change(self):
...
+        :returns: True when another feature is within the threshold distance.
+        """
+        pos = feature.stage_position.value
+        fx, fy = pos.get("x", 0.0), pos.get("y", 0.0)
+        for other in self._tab_data_model.main.features.value:
+            if other is feature:
+                continue
+            other_pos = other.stage_position.value
+            ox, oy = other_pos.get("x", 0.0), other_pos.get("y", 0.0)
+            dist = math.sqrt((fx - ox) ** 2 + (fy - oy) ** 2)
```
pieleric: use `math.dist`

src/odemis/util/datacollector.py
```diff
@@ -657,6 +656,9 @@ def __init__(self) -> None:
         self._worker: Optional[_BackgroundWorker] = None
         self._init_ok: bool = False
         self._init_lock = threading.Lock()
+        # Collect percentage of the acquired data based on the selected probability
+        # in order to keep the growth of the collected data in control.
```
pieleric: This comment is very hard to understand. Can you re-explain in your own words. I think the most important part is that it's a float between 0 -> 1 (right?). And a small value means less chance to collect.

src/odemis/util/datacollector.py
```diff
@@ -788,16 +790,9 @@ def record(
...
             else:
-                probability = _DEFAULT_COLLECTION_PROBABILITY
-
-            if random.random() >= probability:
-                logging.debug(
-                    "DataCollector: event '%s' not sampled (%.0f%% collection probability).",
-                    event_name, probability * 100,
-                )
-                return
+                self.probability = _DEFAULT_COLLECTION_PROBABILITY
```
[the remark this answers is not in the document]
pieleric: @K4rishma I agree here that it doesn't make sense to re-update .probability every time `.record()` is called. It should be done at init only then, or when the consent date changes.

src/odemis/gui/cont/features.py
```diff
@@ -449,3 +478,92 @@ def _on_ctrl_feature_z_change(self):
...
+
+    def _has_nearby_feature(self, feature: CryoFeature, distance_m: float = _NEARBY_FEATURE_DISTANCE_M) -> bool:
+        """Return True if any other feature is within distance_m of the given feature.
+
+        :param feature: The feature to check proximity for.
+        :param distance_m: Maximum distance in metres to be considered nearby.
+        :returns: True when another feature is within the threshold distance.
+        """
+        pos = feature.stage_position.value
+        fx, fy = pos.get("x", 0.0), pos.get("y", 0.0)
```
pieleric: Do we need such protection? All features should have x&y in the positions. For "safety" just catch all exceptions and claim that the feature is not near any other feature.

src/odemis/gui/cont/features.py
```diff
@@ -449,3 +478,92 @@ def _on_ctrl_feature_z_change(self):
         zpos = self._panel.ctrl_feature_z.GetValue()
 
         return {"z": zpos}
+
+    def _on_status_for_collection(self, _status: str) -> None:
```
pieleric: I'm not fond of this name... but can't find something really better. Maybe just "_on_feature_status"? Any other idea?

pieleric (changes requested):

src/odemis/gui/cont/tabs/cryo_chamber_tab.py
```diff
@@ -373,6 +375,18 @@ def _change_project_conf(self, new_dir):
         self.conf.pj_ptn, self.conf.pj_count = guess_pattern(new_dir)
         self.txt_projectpath.Value = os.path.basename(self.conf.pj_last_path)
         self.tab_data_model.main.project_path.value = new_dir
+        # Decide once per project whether features created during this session are
+        # eligible for data collection.  Stored as a dynamic attribute — not part
```
tmoerkerken:
```suggestion
        # eligible for data collection. Stored as a dynamic attribute — not part
```

src/odemis/acq/feature.py
```diff
@@ -205,6 +210,9 @@ def __init__(self, name: str,
             correlation_data = {}
         self.correlation_data = correlation_data
 
+        # Whether this feature is eligible for data collection
+        self.collect: bool = collect
```
tmoerkerken: I think this is not a great name. Collect is a verb and implies a method that does collection in my view. I'd rather go with something self-explanatory like `eligible_for_collection`, or a bit less self-explanatory, yet shorter `is_collectible`.

src/odemis/gui/cont/tabs/cryo_chamber_tab.py
```diff
@@ -373,6 +375,18 @@ def _change_project_conf(self, new_dir):
...
+        # of the formal model — and read by add_new_feature via getattr.
+        probability = DataCollector().probability
+        self.tab_data_model.main.features_collectable = (
```
tmoerkerken: Is this now also executed when there is no consent? Especially the log can be confusing then: no consent, but stuff eligible for collection.

src/odemis/gui/cont/features.py
tmoerkerken: Is there a reason we ignore z? Also, would using the sample stage coordinate space make more sense?

tmoerkerken (changes requested):

src/odemis/gui/cont/tabs/cryo_chamber_tab.py
K4rishma: The probability can either be 100percent or x percent. That is why I ask the class to provide the probability. Should this be not the case?

src/odemis/util/datacollector.py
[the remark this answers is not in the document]
K4rishma: I have removed the probability from this function (record). I am unclear on one thing. I am not using the probability directly in DataCollector class in the record method. I am using the probability attribute of this class, and calling record method when collecting the desired data. Is that a good approach? Or should I call record at every trigger and put probability check inside record method in order to decide if the data needs to be uploaded. 

I have for now chosen the first option to minimize the payload preparation

src/odemis/gui/model/tab_gui_data.py
K4rishma: renamed to collect_features

src/odemis/acq/feature.py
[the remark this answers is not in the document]
K4rishma: I have now made a common instance to share the data collector once created. It is in datacollector.py This function is used to get the shared instance. I did not put this main.gui.data because - as an external item (service)  it has it own skills and it is best to keep separate from main_gui_data according to AI as below. Let me know what you think

Why not put DataCollector inside MainGUIData?
It mixes state model and infrastructure service in one place.
It makes MainGUIData heavier and harder to test (tests now need/avoid a service with background behavior).
It encourages “grab from global model everywhere” coupling.

src/odemis/gui/cont/tabs/cryo_chamber_tab.py
K4rishma: It just calculates the probability, if the consent is not there (new method is added in datacollector.py to calculate probability), the probability will be zero which will not collect any data

tmoerkerken (approved):

src/odemis/gui/model/main_gui_data.py
```diff
@@ -274,6 +274,9 @@ def __init__(self, microscope):
         # Indicates whether the microscope is milling
         self.is_milling = model.BooleanVA(False)
 
+        # Controls if the acquired features should be collected by the DataCollector
+        self._collect_features = False
```
pieleric, line 278: It's called `.collect_features` in other places. Let's just use the "public" name everywhere. This attribute should be readable by other classes.

PR #3454, opened 2026-04-24
[TRCL-645][TRCL-694][feat] SPARC streakcam: improve settings control

pieleric:
Several small improvements to the GUI when using SPARC with streakcam.

* Can control the phase lock (on/off) of the Hamamatsu delay generator from Odemis
* User can control the trigger delay directly from the acquisition tab
* time range -> trigger delay connection was not set at init
* Add entries to write down the gun parameters in the metadata

Now the Temporal spectrum stream looks like that:
image
The gun exciter settings look like that:
image

### DIFF
--- a/src/odemis/driver/hamamatsurx.py
+++ b/src/odemis/driver/hamamatsurx.py
@@ -1043,13 +1043,26 @@ class DelayGenerator(model.HwComponent):
             # Create a VA, with a corresponding setter function
             delay_setter = functools.partial(self._setDelayByName, rx_param)
             self._delay_setters[rx_param] = delay_setter  # keep a ref to the partial so that it's not garbage collected
-            delay_va = model.FloatContinuous(delay, range_delay, setter=delay_setter, unit="s")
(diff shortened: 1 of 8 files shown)


tepals (approved):
Looks good to me, but I am very unfamiliar with this code

src/odemis/driver/hamamatsurx.py
```diff
@@ -1081,8 +1094,12 @@ def updateMetadata(self, md):
                                      "Please check the calibration file for the trigger delay."
                                      % (delay, timeRange, self.triggerDelay.range))
 
+        first_time = model.MD_TIME_RANGE_TO_DELAY not in self._metadata
```
nandishjpatel:
```suggestion
        is_first_time = model.MD_TIME_RANGE_TO_DELAY not in self._metadata
```

for readability that its a boolean, similarly in line 447 of src/odemis/driver/simstreakcam.py; line 124 of src/odemis/driver/streak.py

nandishjpatel (approved):

PR #3456, opened 2026-04-29
[fix] Import annotations from future for typing compatibility with python 3.8

tepals:
Using lower-case dict, list, and tuple for type hinting is not supported in Python 3.8

In Python 3.8, these are built-in collection types, but they do not support indexing (the [] syntax) for typing. That specific feature was introduced in Python 3.9


tmoerkerken (approved):
Nice one, it's a bit magical how it works though 😅

tepals:
> Nice one, it's a bit magical how it works though 😅

haha, agreed :sweat_smile:, that's why I added the comment. I also thoroughly tested it, because I did not fully trust it, but it works.

ilyushkin (approved):

PR #3458, opened 2026-05-05
[MSD-416][fix] race condition in overview streams

tmoerkerken:
Fixes a race condition causing certain streams not to show up after multi-channel overview acquisition.

## Out of scope
Sadly, I encountered more broken code in the syncing between tabs, which now causes one more additional (duplicate channel) to show up. I decided to create a backlog item, and leave this PR minimal, since this in fact fixes an obvious bug. But the whole Idea of altering the variable that triggered the callback, **within** the callback itself feels very fishy to me. This causes race conditions and extra triggers. Also the whole correlation.py code where streams from other tabs get deleted feels off. I think a tab should be responsible for its own visualization, and if another tab is altering something relevant to the project, this should go via the project file. Since we just introduced this concept of a project file, we could improve this a lot.

UPDATE: decided to patch it anyway. A proper fix will be part of the UI redesign.

### DIFF
--- a/src/odemis/gui/cont/stream_bar.py
+++ b/src/odemis/gui/cont/stream_bar.py
@@ -2197,7 +2197,7 @@ class CryoAcquiredStreamsController(CryoStreamsController):
 
     def showFeatureStream(self, stream) -> StreamController:
         """
-        Shows an Feature stream (in the Acquired view)
+        Shows a Feature stream (in the Acquired view)
         Must be run in the main GUI thread.
         """
         # TODO: don't delete/create stream controller every time? Instead, we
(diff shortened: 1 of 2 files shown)


src/odemis/gui/cont/stream_bar.py
```diff
@@ -2197,7 +2197,7 @@ def showOverviewStream(self, stream) -> StreamController:
 
     def showFeatureStream(self, stream) -> StreamController:
         """
-        Shows an Feature stream (in the Acquired view)
+        Shows a Feature stream (in the Acquired view)
```
tepals, line 2200: I'm not a big fan of adding files with only very tiny changes to a PR if it does not influence the PR. But if you want to keep this, I'm also fine with keeping it.

tepals: For the rest the PR looks good

tepals (approved):

src/odemis/gui/cont/stream_bar.py
tmoerkerken: I do agree it feels a bit out of place here. But maybe let's keep it for now, and discuss later how to go about these kind of fixes?

K4rishma (approved):

pieleric (approved):

PR #3464, opened 2026-05-18
[MSD-592] Pass milling preset name to fibsemOS

ilyushkin:
The Tescan API expects a preset name to be provided for the FIB column.
This change constructs a preset name from the voltage and current values for fibsemOS that passes it to the Tescan API.
The same `FibsemMillingSettings` class is also used for TFS microscopes, where the preset name is ignored.

### DIFF
--- a/src/odemis/acq/milling/fibsemos.py
+++ b/src/odemis/acq/milling/fibsemos.py
@@ -235,13 +235,67 @@ def _convert_microexpansion_pattern(p: MicroexpansionPatternParameters) -> 'Micr
         point=Point(x=p.center.value[0], y=p.center.value[1])
     )
 
+def _format_preset(voltage: float, current: float) -> str:
+    """
+    Format voltage (V) and current (A) into a preset name string like '30 keV; 150 pA'.
+    Voltage is shown in eV below 1 keV, keV otherwise. Current is shown in pA, nA, or uA.
+
+    This format is a convention for Tescan preset names. Presets in the Tescan
+    software must be named following this convention so that fibsemOS can match
+    them correctly.
+
+    :param voltage: beam voltage in volts (must be a positive finite number).
+    :param current: beam current in amperes (must be a positive finite number).
+    :raises ValueError: if voltage or current are not positive finite numbers.
+    """
+    if not math.isfinite(voltage):
+        raise ValueError(f"Voltage must be a finite number, got {voltage!r}")
+    if voltage <= 0:
+        raise ValueError(f"Voltage must be positive, got {voltage!r}")
+    if not math.isfinite(current):
+        raise ValueError(f"Current must be a finite number, got {current!r}")
+    if current <= 0:
+        raise ValueError(f"Current must be positive, got {current!r}")
+
+    # Voltage: choose eV or keV based on order of magnitude
+    if voltage < 1000:
+        # eV range: [0, 1 keV)
+        voltage_str = f"{voltage:g} eV"
+    else:
+        # keV range: [1 keV, ...)
+        voltage_str = f"{voltage / 1000:g} keV"
+
+    # Current: pA, nA, or uA based on order of magnitude
+    if current < 1e-9:
+        # pA range: [0, 1 nA)
+        current_val = current * 1e12
+        unit = "pA"
+    elif current < 1e-6:
+        # nA range: [1 nA, 1 uA)
+        current_val = current * 1e9
+        unit = "nA"
+    else:
+        # uA range: [1 uA, ...)
+        current_val = current * 1e6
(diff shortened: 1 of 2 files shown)


src/odemis/acq/milling/fibsemos.py
```diff
@@ -235,13 +235,57 @@ def _convert_microexpansion_pattern(p: MicroexpansionPatternParameters) -> 'Micr
...
+    """
+    if not math.isfinite(voltage):
+        raise ValueError(f"Voltage must be a finite number, got {voltage!r}")
+    if voltage <= 0:
+        raise ValueError(f"Voltage must be positive, got {voltage!r}")
+    if not math.isfinite(current):
+        raise ValueError(f"Current must be a finite number, got {current!r}")
+    if current <= 0:
+        raise ValueError(f"Current must be positive, got {current!r}")
+
```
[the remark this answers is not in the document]
ilyushkin: I would prefer to keep the logic as is and raise ValueError if either voltage or current is zero.

src/odemis/acq/milling/fibsemos.py
```diff
@@ -235,13 +235,57 @@ def _convert_microexpansion_pattern(p: MicroexpansionPatternParameters) -> 'Micr
...
+    # Voltage: choose eV or keV based on order of magnitude
+    if voltage < 1000:
+        # eV range: [0, 1 keV)
```
tmoerkerken, line 262: Is this `[..., ...)` notation intended? I'm not familiar with it 😅

src/odemis/acq/milling/fibsemos.py
```diff
@@ -235,13 +235,57 @@ def _convert_microexpansion_pattern(p: MicroexpansionPatternParameters) -> 'Micr
...
+    return f"{voltage_str}; {current_str}"
+
 def convert_milling_settings(s: MillingSettings) -> 'FibsemMillingSettings':
     """Convert Odemis milling settings to fibsemOS milling settings."""
     return FibsemMillingSettings(
         milling_current=s.current.value,
         milling_voltage=s.voltage.value,
         patterning_mode=s.mode.value,
         hfw=s.field_of_view.value,
+        preset=_format_preset(s.voltage.value, s.current.value)
```
tmoerkerken, line 298: I'm not entirely sure what the intend of passing this to FibsemOS is, but it is possible that the preset name does not match the voltage and current settings (since users can name their own presets). Just double checking this is fine.

tmoerkerken (approved):

src/odemis/acq/milling/fibsemos.py
ilyushkin: Yes, it is intended interval notation.

src/odemis/acq/milling/fibsemos.py
ilyushkin: You're right, this is just a workaround to make it work now with Tescan with minimal modifications. There is no guarantee that the preset name will match an actual preset available in Essence. At this point, it is the user's responsibility to ensure that the selected voltage/current combination exists in Essence. Eventually, this should be replaced with a per-feature preset picker in Odemis, similar to how this is already done for FIB acquisition.

src/odemis/acq/milling/fibsemos.py
tmoerkerken: Ah I see, cool!

src/odemis/acq/milling/fibsemos.py
pieleric, line 298: If I understand correctly, reading the fibsemOS code, the .preset is only used for Tescan, while "milling_current" and "milling_voltage" are only used for TFS. Can you add this info here, to clarify why we pass both all of them

src/odemis/acq/milling/fibsemos.py
```diff
@@ -235,13 +235,57 @@ def _convert_microexpansion_pattern(p: MicroexpansionPatternParameters) -> 'Micr
         point=Point(x=p.center.value[0], y=p.center.value[1])
     )
 
+def _format_preset(voltage: float, current: float) -> str:
+    """
+    Format voltage (V) and current (A) into a string like:
+    '30 keV; 150 pA', scaling units automatically.
+    Voltage is shown in eV below 1 keV, and in keV otherwise.
```
pieleric: I'd suggest to mention that this is a *convention* and that the user/support engineer is supposed to follow that convention in the Tescan software when creating presets. 

I think it should also also be indicated in the METEOR installation manual for Tescan. Could you add a paragraph about this?

pieleric (changes requested):

pieleric (approved):

PR #3466, opened 2026-05-19
[fix] saves partial z-stack during mid error

K4rishma:
Root cause of the original bug: a camera communication error could return an image with the wrong shape.  On NumPy < 1.24, numpy.array() on mixed-shape arrays silently produces an object-dtype array, which cannot be written to TIFF (WriteDirectory() → AssertionError: 0).

### DIFF
--- a/src/odemis/acq/acqmng.py
+++ b/src/odemis/acq/acqmng.py
@@ -166,6 +166,7 @@ class ZStackAcquisitionTask(object):
         """
         remaining_t = self.estimate_total_duration()
         acquired_data = []
+        exp = None
         # iterate through streams
         for stream in self._streams:
             zstack = []
@@ -201,6 +202,7 @@ class ZStackAcquisitionTask(object):
 
             else:
                 # for each stream, iterate through zlevels
+                z_exp = None  # tracks failure within this stream's z-stack
                 for i, z in enumerate(self._zlevels[stream]):
                     # move the focuser
                     self._actuator_f = stream.focuser.moveAbs({"z": z})
@@ -218,9 +220,10 @@ class ZStackAcquisitionTask(object):
                     try:
                         # acquire this single stream, and get the data
                         self._single_acqui_f = acquire([stream], self._settings_obs)
-                        data, exp = self._single_acqui_f.result()
-                        if exp:
-                            return acquired_data, exp
+                        data, z_exp = self._single_acqui_f.result()
+                        if z_exp:
+                            # exit the z-loop; partial z-stack will be assembled below
+                            break
                         # check if cancellation happened while the acquiring future is working
                         if self._future_state == CANCELLED:
                             raise CancelledError()
@@ -234,20 +237,53 @@ class ZStackAcquisitionTask(object):
                     except CancelledError:
                         raise
                     except Exception as e:
-                        logging.exception("The acquisition failed at the %s-th zlevel of the stream %s, because %s" % (
-                        i + 1, stream, e))
-                        # TODO handle zstack assembling in case of error
-                        return acquired_data, e
+                        logging.exception("The acquisition failed at the %s-th zlevel of the stream %s",
+                                          i + 1, stream.name.value)
+                        z_exp = e
+                        break  # exit the z-loop; partial z-stack will be assembled below
 
                     # only if there is data acquired
                     if data:
-                        zstack.append(data[0])
+                        da = data[0]
+                        # In case of partial error, the data might be empty
+                        if zstack and da.shape[-2:] != zstack[0].shape[-2:]:
+                            logging.error(
+                                "Z-level %d of stream %s has shape %s, inconsistent with "
+                                "first z-level shape %s; stopping z-stack acquisition",
+                                i + 1, stream.name.value, da.shape[-2:], zstack[0].shape[-2:]
+                            )
+                            z_exp = ValueError(
+                                "Z-level %d image shape %s is inconsistent with first "
+                                "z-level shape %s" % (i + 1, da.shape[-2:], zstack[0].shape[-2:])
+                            )
+                            break
+                        zstack.append(da)
+
                     # update the remaining time
                     remaining_t -= stream.estimateAcquisitionTime()
                     self._main_future.set_end_time(time.time() + remaining_t)
 
-                zcube = assembleZCube(zstack, self._zlevels[stream])
-                acquired_data.append(zcube)
+                # Assemble whatever z-levels were acquired (partial or complete).
+                if zstack:
+                    if len(zstack) < len(self._zlevels[stream]):
+                        logging.warning(
+                            "Partial z-stack for stream %s: assembling %d of %d levels",
+                            stream.name.value, len(zstack), len(self._zlevels[stream])
+                        )
+                    try:
+                        zcube = assembleZCube(zstack, self._zlevels[stream][:len(zstack)])
+                        acquired_data.append(zcube)
+                    except Exception as e:
+                        logging.exception("Failed to assemble z-stack for stream %s",
+                                          stream.name.value)
+                        if z_exp is None:
+                            z_exp = e
+                elif z_exp is None:
+                    logging.warning("All z-levels for stream %s returned empty data, skipping",
+                                    stream.name.value)
+
+                if z_exp is not None:
+                    return acquired_data, z_exp
 
         # state that the future has finished
         with self._future_lock:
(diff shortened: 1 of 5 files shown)


src/odemis/acq/test/acq_test.py
```diff
@@ -856,5 +858,153 @@ def test_settings_observer_metadata_with_zstack(self):
...
+def _make_sim_future(result=None):
+    """
+    Return a stdlib Future already completed with result.
+
+    :param result: value to store in the future
+    :return: completed concurrent.futures.Future
+    """
+    f = Future()
+    f.set_result(result)
+    return f
```
pieleric: We already have something similar. It's a class: create a `InstantaneousFuture()`.

src/odemis/acq/test/acq_test.py
```diff
@@ -856,5 +858,153 @@ def test_settings_observer_metadata_with_zstack(self):
...
+        """
+        zlevels_list = [0.0e-6, 1.0e-6, 2.0e-6]
+        s = _make_sim_stream("fluo")
+        task, _ = _make_sim_task(s, {s: zlevels_list})
+
+        good_img = _make_sim_data_array((64, 64))
+        bad_img = _make_sim_data_array((32, 64))  # truncated height — simulates camera error
+
+        acq_futures = [
+            _make_sim_future(([good_img], None)),
```
pieleric: `InstantaneousFuture([good_img])`

src/odemis/acq/acqmng.py
```diff
@@ -234,20 +237,57 @@ def run(self):
...
+                        zstack.append(da)
+
                     # update the remaining time
                     remaining_t -= stream.estimateAcquisitionTime()
                     self._main_future.set_end_time(time.time() + remaining_t)
 
-                zcube = assembleZCube(zstack, self._zlevels[stream])
-                acquired_data.append(zcube)
+                # Assemble whatever z-levels were acquired (partial or complete).
+                # This also handles the TODO: save partial z-stack data on failure.
```
pieleric: Drop this second line about handling the TODO. If it's done, it's done.

src/odemis/acq/test/acq_test.py
```diff
@@ -856,5 +858,153 @@ def test_settings_observer_metadata_with_zstack(self):
...
+        run() must stop, assemble only the valid z-levels, and report the error.
+
+        This is the primary regression test for the TIFF-crash bug.
+        """
+        zlevels_list = [0.0e-6, 1.0e-6, 2.0e-6]
+        s = _make_sim_stream("fluo")
+        task, _ = _make_sim_task(s, {s: zlevels_list})
+
+        good_img = _make_sim_data_array((64, 64))
+        bad_img = _make_sim_data_array((32, 64))  # truncated height — simulates camera error
```
pieleric: I don't think that simulates a camera error. In no cases will there be a truncated heigh. (it could happen only if the user manually change the binning while acquiring).  Typically, in case of camera error, you just get a timeout error and (maybe) an empty array as data.

src/odemis/util/img.py
```diff
@@ -1286,7 +1286,22 @@ def assembleZCube(images, zlevels):
...
         :param zlevels:  (list of float) list of focus positions
         :return: (DataArray of shape ZYX) the data array of the xyz cube
+        :raises ValueError: if images is empty or the images have inconsistent spatial shapes
         """
+        if not images:
+            raise ValueError("Cannot assemble z-cube from an empty image list")
+
+        # Validate that all images have the same spatial (YX) dimensions.
+        # With NumPy < 1.24, numpy.array() on a list of arrays with different shapes
+        # silently creates an object-dtype array, which cannot be written to TIFF.
```
pieleric: Drop the "which cannot be written to TIFF", and replace by "instead of generating the expected 3D array".

src/odemis/acq/acqmng.py
```diff
@@ -234,20 +237,57 @@ def run(self):
...
+                        break  # exit the z-loop; partial z-stack will be assembled below
 
                     # only if there is data acquired
                     if data:
-                        zstack.append(data[0])
+                        da = data[0]
+                        # Validate spatial shape consistency across z-levels.
+                        # A camera communication error can return an image of the wrong size.
+                        # With NumPy < 1.24, numpy.array() on mixed-shape images silently
+                        # creates an object-dtype array, which cannot be written to TIFF.
```
pieleric: No need to be so specific. Just indicate in case of partial failure, the data might not be the right shape, so better drop it than not being able to create a Z-stack at all.

pieleric (changes requested):
Nice way to write defensive code by handling the error at 3 different levels.

The whole code assumes that the issue is because one image was of different shape. Did you really really confirm this? Were you able to reproduce it and observe the shape, or see it in the log? To the best of my knowledge, a camera issue would just return no data at all. I've never seen a camera returning a part of a frame. There would be other ways to end up with a frame of different shape though, such as if the binning was changed by the user.

src/odemis/acq/acqmng.py
ilyushkin: I think using "not the right shape" wording here is not ideal as we only care about situations when one of the Z-stack layers is simply `None` or fully empty and has the `(0, 0)` shape. We just want to discard those empty arrays. I think this wording is misleading and that's the reason why in line 969 in acq_test.py Karishma also tried to simulate the truncated height.

ilyushkin (changes requested):

src/odemis/acq/test/acq_test.py
K4rishma: Removing the test case as it is not seen if that is actually case that the camera produces wrong shape

src/odemis/acq/test/acq_test.py
K4rishma: Rmeoved it

src/odemis/util/img.py
```diff
@@ -1286,7 +1286,22 @@ def assembleZCube(images, zlevels):
...
+        ref_shape = images[0].shape[-2:]
```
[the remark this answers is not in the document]
ilyushkin: I do not think we should require the number of acquired images to match the number of requested z-levels, since missing planes can be valid and should be skipped. The more important issue here is preserving the correct image-to-z-level mapping: if a missing image occurs in the middle of the stack, slicing the first `len(zstack)` z-levels will produce incorrect Z metadata. We should track the z-levels corresponding to the actually acquired images instead.

src/odemis/acq/acqmng.py
```diff
@@ -234,20 +237,53 @@ def run(self):
...
-                acquired_data.append(zcube)
+                # Assemble whatever z-levels were acquired (partial or complete).
+                if zstack:
+                    if len(zstack) < len(self._zlevels[stream]):
+                        logging.warning(
+                            "Partial z-stack for stream %s: assembling %d of %d levels",
+                            stream.name.value, len(zstack), len(self._zlevels[stream])
+                        )
+                    try:
+                        zcube = assembleZCube(zstack, self._zlevels[stream][:len(zstack)])
```
ilyushkin, line 274: This only works correctly if missing images occur at the end of the z-stack. If an image is missing in the middle, the Z metadata will be shifted. Should we keep a separate list of z-levels for the images that were actually acquired?

ilyushkin (changes requested):

src/odemis/util/test/img_test.py
```diff
@@ -2068,6 +2068,33 @@ def test_assemble_zcube_z_order(self):
...
+        assembleZCube() must raise ValueError when z-level images have different shapes.
+
+        On NumPy < 1.24, numpy.array() silently creates an
+        object-dtype array, instead of returning the expected 3D array.  The fix detects this
+        early and raises explicitly.
+        """
+        images = [
+            model.DataArray(numpy.zeros(self.size, dtype=numpy.uint16), self.md),
+            model.DataArray(numpy.zeros(self.size, dtype=numpy.uint16), self.md),
+            model.DataArray(numpy.zeros((self.size[0] // 2, self.size[1]), dtype=numpy.uint16), self.md),
```
pieleric, line 2089: I don't think it makes a lot of sense to test this, because there is no reason such issue ever happens. I'm pretty certain that the initial error reported by the user wasn't caused by the image changing resolution, but more that the camera failed to send an image. (ie, some sort of Timeout).

pieleric (changes requested):
Eventually, the error appeared with a backtrace from libtiff.TIFF.WriteDirectory. 
1. Please add the whole backtrace to the commit message, so that it's easier to find when another user reports the same issue.
2. Is that a way to add some early error report in dataio.tiff that catch when the dtype is not supported by libtiff, and output more detailed information (eg, the dtype and shape). This way, if this happens again for a different reason, we can more easily reproduce & fix.

PR #3467, opened 2026-05-21
[fix] testcases for ci run 20260521

nandishjpatel:


src/odemis/acq/move.py
pieleric, line 1: I have an almost identical PR to fix this issue (#3468), with just a tiny bit extra check. Could you drop this commit (and review my PR)?

src/odemis/driver/test/simsem_test.py
```diff
@@ -166,7 +165,7 @@ def test_acquire_full(self):
         duration = time.time() - start
 
         self.assertEqual(im.shape, self.size[::-1])
-        self.assertGreaterEqual(duration, expected_duration, "Error execution took %f s, less than exposure time %d." % (duration, expected_duration))
+        self.assertAlmostEqual(duration, expected_duration, delta=0.1, msg="Error execution took %f s, expected %f s." % (duration, expected_duration))
```
pieleric: No, I don't think that change is correct. The idea is that `expected_duration` is the theoretical minimum duration. So anything shorter than that is sign that something is incorrect.
I think the issue is more that `compute_expected_duration()` includes settle time, while the simulator doesn't. We can drop the settle time computation. (or add it to the simulator).

src/odemis/odemisd/test/start_test.py
```diff
@@ -81,7 +81,14 @@ def thread_call():
...
         self._gui_thread.start()
-        time.sleep(30)  # Give the back-end & GUI time to properly start
+
+        # Poll for the window to appear, rather than using a fixed sleep
+        timeout = 60  # s
+        end_time = time.time() + timeout
+        while time.time() < end_time:
+            if find_window("Odemis"):
+                break
+            time.sleep(10)
```
pieleric: I think you can change to `sleep(1)`, which should allow to make the tests run a tiny bit faster.

pieleric (changes requested):

src/odemis/gui/test/comp_text_test.py
```diff
@@ -90,16 +89,28 @@ def test_unit_float(self):
 class NumberTextCtrlTestCase(test.GuiTestCase):
     frame_class = test.test_gui.xrcbutton_frame
 
-    def _check_wxpython_41(self):
-        """
-        Skip the test if wxpython < 4.1.
+    def _simulate_type(self, ctrl, text):
```
tepals: the name is a bit confusing, at first sight I would think it has to do with simulating the type of a string. Maybe go for `_simulate_typing`

tepals (approved):

src/odemis/gui/test/comp_text_test.py
nandishjpatel: Updated!

pieleric (approved):

pieleric:
@nandishjpatel the changes look fine, but could you squash the last commit (pr fixes) into the 2 respective commits that it fix. I'd like to keep each fix independent from each other.

nandishjpatel:
> @nandishjpatel the changes look fine, but could you squash the last commit (pr fixes) into the 2 respective commits that it fix. I'd like to keep each fix independent from each other.

@pieleric each fix is now independent from each other. Thr PR is now ready to be merged.

PR #3469, opened 2026-05-22
[feat] plugin secom_user_sim: also support METEOR

pieleric:
METEOR is very similar to the SECOM. The "user simulator" also work on
it, assuming it knows which tab to look for.

Also add support to move the focus.


tmoerkerken (approved):

K4rishma (approved):

plugins/secom_user_sim.py
```diff
@@ -31,7 +31,7 @@
 
 class SecomUserPlugin(Plugin):
     name = "SECOM user simulator"
```
tepals: should this name be updated as well?

tepals (approved):
one small comment

PR #3474, opened 2026-05-28
[refactor] util: split __init__.py into geometry and concurrent modules

tepals:
Extract 2D geometry functions into a new `geometry.py` module:
 - Rectangle operations: rect_intersect, intersect, normalize_rect, expand_rect, is_point_in_rect, rotate_rect, separate_rect_rotation
 - Line/polygon operations: perpendicular_distance, clip_line, get_polygon_bbox, slope_of_line, intercept_of_line, project_point_on_line
 - Constants: INSIDE, LEFT, RIGHT, LOWER, UPPER

 Extract concurrency utilities into a new `concurrent.py` module:
 - Rate limiting: limit_invocation
 - Threading: RepeatingTimer, BackgroundWorker
 - Future helpers: executeAsyncTask, bindFuture
 - Decorators: timeout
 - Introspection: inspect_getmembers

General-purpose helpers remain in __init__.py

 All moved symbols are re-exported from `odemis.util` for backward compatibility.

 Co-authored-by: Copilot

### DIFF
--- a/src/odemis/util/geometry.py
+++ b/src/odemis/util/geometry.py
@@ -1,0 +1,417 @@
+# -*- coding: utf-8 -*-
+"""
+Utility functions for 2D geometry: rectangles, lines, polygons and projections.
+
+Copyright © 2013-2024 Delmic
+
+This file is part of Odemis.
+
+Odemis is free software: you can redistribute it and/or modify it under the terms
+of the GNU General Public License version 2 as published by the Free Software
+Foundation.
+
+Odemis is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+PURPOSE. See the GNU General Public License for more details.
+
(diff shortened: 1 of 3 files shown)


src/odemis/util/geometry.py
pieleric, line 1: How about naming it using the full word "geometry"? I see there is already a module in sympy called "geometry", which match this aspect, and I couldn't find any module called "geom" in the common python libraries.

pieleric (changes requested):
Looks to me like a good separation. I only have one small suggestion.

src/odemis/util/geometry.py
tepals: makes sense, done!

tmoerkerken (approved):
I think the suggested changes are fine, but we probably should move all remaining methods to submodules eventually. Also, I think it would be nicer, from now on, to not directly import from util anymore, but from the submodule (e.g. `util.concurrent`). But we could leave that for another round of refactoring.

pieleric (approved):

PR #3479, opened 2026-06-03
[MSD-612][feat] Meteor: MILLING <> FIB-VIEW FM posture switching (TESCAN/TFS)

tmoerkerken:
Add support for MILLING <> FIB-VIEW FM posture switching for Tescan and TFS:
- implement transforms for `milling_to_fib_view_fm` and `fib_view_fm_to_milling` (both Tescan and TFS transforms checked by Jochem)
- implement additional `fib_view_fm` transforms to facilitate sample stage computation
- update sample stage computation
- add new icons and update existing ones
- add controls to chamber tab
- update microscope files to include new focus metadata
- add new focus distance to choreography 
- move calculate milling angle <> stage tilt methods to posture manager base class
- add unit tests 

[Screencast from 2026-06-03 14-58-36.webm](https://github.com/user-attachments/assets/96b86852-9d08-4edf-8754-0e1a9c5aaf4f)

> [!NOTE]
> We decided to rename the label after making the video. It is now called FIB-VIEW FM and not FM MILLING.


src/odemis/acq/move.py
```diff
@@ -337,6 +339,7 @@ def __init__(self, microscope):
         md_calib = stage_md.get(model.MD_CALIB, {})
         self.pre_tilt = md_calib.get(model.MD_SAMPLE_PRE_TILT, None)
         self.fib_column_tilt = TFS_FIB_COLUMN_TILT
+        self.fm_column_tilt = self.pre_tilt + stage_md.get(model.MD_FAV_FM_POS_ACTIVE)["rx"]
         # use_linked_sem_focus_compensation: when True, the SEM focus is restored to the eucentric
```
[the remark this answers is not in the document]
tmoerkerken: agreed, moving this one to tfs3. For tescan I already handle it separately.

src/odemis/acq/move.py
```diff
@@ -897,6 +955,64 @@ def _transform_from_fib_to_milling(self, pos: Dict[str, float]) -> Dict[str, flo
...
+        # Overrides based on specific workflows
+        if source_posture == LOADING:
+            move_allowed = True
+        # For FM @ Milling angle we only allow to go to and from the milling posture.
+        if target_posture == FM_MILLING and source_posture != MILLING:
+            move_allowed =  False
+        elif source_posture == FM_MILLING and target_posture != MILLING:
+            move_allowed =  False
+
+        return move_allowed
```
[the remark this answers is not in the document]
tmoerkerken: Highlighted code is outdated, but in general I think it is a fair remark. I think we should do that eventually. For now, I like to restrict it to the chamber tab to reduce the risk of breaking existing workflows. When implemented correctly, it shouldn't though :sweat_smile: .

install/linux/usr/share/odemis/sim/meteor-tescan-fibsem-full-sim.odm.yaml
```diff
@@ -79,7 +79,7 @@
         port: "/dev/fake6",
         address: null,
         axes: ["x", "y", "z", "rx", "rz"],
-        ustepsize: [1.e-7, 1.e-7, 1.e-7, 1.2e-5, 1.2e-5], # unit/µstep
+        ustepsize: [1.e-6, 1.e-6, 1.e-6, 1.2e-4, 1.2e-4], # unit/µstep, 10x faster than real life values for simulation purposes
```
tepals, line 82: I think for development it is fine to make it 10x faster. However, we also use this microscope file for testing the GUI before a release, I think we don't want to have too fast then. I'm not sure what the best way to do this is, it would be nice to have a dev mode plugin that updates these values, but I think these values cannot be updated after starting.

Anyway I was also wondering about this unit: `unit/µstep`, looking at the docstring I think it should be `m/µstep` @pieleric do you know?
```
        ustepsize (list of float): size of a microstep in m (the smaller, the
          bigger will be a move for a given distance in m)
```

src/odemis/acq/test/move_tescan_test.py
```diff
@@ -307,6 +307,22 @@ def test_milling_angle_stable_pos(self):
...
+        """Test that milling <> fm milling switches work as expected"""
+        self.pm.cryo_switch_sample_position(MILLING).result()
+        self.assertEqual(self.pm.current_posture.value, MILLING)
+
+        self.pm.cryo_switch_sample_position(FM_MILLING).result()
+        resulting_posture = self.pm.get_current_posture_label()
+        self.assertEqual(resulting_posture, FM_MILLING)
+
+        # Verify that directly going from fm milling to fm imaging is not allowed
+        self.assertRaises(ValueError, lambda: self.pm.cryo_switch_sample_position(FM_IMAGING).result())
```
tepals: The lambda is a bit confusing to read here, I think typically we do 
```python
with self.assertRaises(ValueError):
    self.pm.cryo_switch_sample_position(FM_IMAGING).result()
```
but feel free to leave it the way you implemented it

src/odemis/gui/cont/tabs/cryo_chamber_tab.py
```diff
@@ -47,6 +47,7 @@
     LOADING_PATH,
     MILLING,
     FIB_IMAGING,
+    FM_MILLING,
```
tepals: can you alphabetize this list?

src/odemis/acq/move.py
```diff
@@ -337,6 +337,11 @@ def __init__(self, microscope):
         md_calib = stage_md.get(model.MD_CALIB, {})
         self.pre_tilt = md_calib.get(model.MD_SAMPLE_PRE_TILT, None)
         self.fib_column_tilt = TFS_FIB_COLUMN_TILT
+        if self.pre_tilt is not None:
+            self.fm_column_tilt = self.pre_tilt + stage_md.get(model.MD_FAV_FM_POS_ACTIVE)["rx"]
```
tepals, line 343: are `MD_FAV_FM_POS_ACTIVE` and `rx` always defined or should you add a default to the get?

src/odemis/acq/move.py
```diff
@@ -503,11 +518,51 @@ def at_fib_posture(self, pos: Dict[str, float], stage_md: Dict[str, float]) -> b
                 return True
         return False
 
+    def at_fm_posture(self, pos: Dict[str, float], stage_md: Dict[str, float]) -> bool:
+        """
+        Check if at fm posture
+
+        :param pos the stage position
```
tepals: I think the guideline is to do `:param pos: the stage position`

src/odemis/acq/move.py
```diff
@@ -503,11 +518,51 @@ def at_fib_posture(self, pos: Dict[str, float], stage_md: Dict[str, float]) -> b
                 return True
         return False
 
+    def at_fm_posture(self, pos: Dict[str, float], stage_md: Dict[str, float]) -> bool:
```
tepals: for readability, do you think it makes sense to specifically define `at_fm_imaging_posture` or is that clear by itself?

src/odemis/acq/move.py
```diff
@@ -897,6 +960,79 @@ def _transform_from_fib_to_milling(self, pos: Dict[str, float]) -> Dict[str, flo
...
+        """
+        # Always allow to "stay" at the posture
+        if source_posture == target_posture:
+            return True
+        # Overrides based on specific workflows
+        # From unknown position we never allow moves
+        if source_posture == UNKNOWN:
+            return False
+        # For FM @ Milling angle we only allow to go to and from the milling posture.
+        elif source_posture != MILLING and  target_posture == FM_MILLING:
```
tepals:
```suggestion
        elif source_posture != MILLING and target_posture == FM_MILLING:
```

tepals (changes requested):

install/linux/usr/share/odemis/sim/meteor-tescan-fibsem-full-sim.odm.yaml
tmoerkerken: Indeed, a dev mode is something that would be really nice. We should give it a good thought though. What other things would we like to change? Let's leave it out of scope for now.
I will change the unit and will add a note to the test procedure, since @pieleric made exactly this change for two other systems recently (https://github.com/delmic/odemis/commit/ea5e5880c3690ab3b4da1d1d30d3eafac5e786ca)

src/odemis/acq/move.py
tmoerkerken: yes, and also the fib one that is not in the current diff. It is way more intuitive to follow the already existing definitions, like you are proposing

src/odemis/acq/move.py
tmoerkerken: To me it should, since that's the one thing we want to do with a meteor system. It should be part of the installation procedure. There is more places where we directly access it without safeguards.

install/linux/usr/share/odemis/sim/meteor-tescan-fibsem-full-sim.odm.yaml
pieleric: @tepals 
I think a dev mode might be tricky per se, because that means that we'll almost always use it, and if that hides some issues, we'll only see them when testing. I'd rather find a balance between saving time during development, and making the system realistic. 
Or we should have an explicit "fast" mode, and only use it when actually developing a specific code path, but it should be disabled by default. 

In practice, as you said, these constants are difficult to update live. Maybe it could be just a commented out line. Or maybe we just make them a little less fast/more realistic (eg, 2x slower?). We don't gain a lot during manual testing in waiting 20s for the posture switch either!

As for the units. It depends on the axis: the linear axes are in m/µstep and the rotational ones are in rad/µstep. Hence the generic "unit/µstep".

install/linux/usr/share/odemis/sim/meteor-fibsem-sim.odm.yaml
```diff
@@ -294,7 +294,8 @@
         # Loading position to retract lens
         FAV_POS_DEACTIVE: {'z': -11.5e-3},
         # Initial active position (close from the sample, but not too close, for safety)
-        FAV_POS_ACTIVE: {'z': 1.69e-3}
+        FAV_POS_ACTIVE: {'z': 1.69e-3},
+        FAV_MILL_POS_ACTIVE: { "z": 1.00e-3 },
```
pieleric: Update to use wd.

src/odemis/acq/move.py
```diff
@@ -503,11 +536,53 @@ def at_fib_posture(self, pos: Dict[str, float], stage_md: Dict[str, float]) -> b
...
+        Check if at fm milling posture
+
+        :param pos: the stage position
+        :return: True if the stage is at the fm milling posture, False if not (or not available)
+        """
+        stage_md = self.stage.getMetadata()
+        focus_md = self.focus.getMetadata()
+
+        if model.MD_FAV_MILL_POS_ACTIVE not in focus_md:
+            return False
```
pieleric: Not needed anymore, right?

src/odemis/acq/move.py
```diff
@@ -418,6 +444,7 @@ def get_current_posture_label(self, pos: Dict[str, float] = None) -> int:
         :return: (int) a label LOADING, SEM_IMAGING, FM_IMAGING or UNKNOWN
         """
         stage_md = self.stage.getMetadata()
+        focus_md = self.focus.getMetadata()
```
pieleric: You don't need this anymore ,right?

src/odemis/acq/move.py
```diff
@@ -897,6 +1007,120 @@ def _transform_from_fib_to_milling(self, pos: Dict[str, float]) -> Dict[str, flo
...
+        :param target_posture: the target posture
+        :return: whether it is allowed to move from source to target posture
+        """
+        # Always allow to "stay" at the posture
+        if source_posture == target_posture:
+            return True
+        # Overrides based on specific workflows
+        # From unknown position we never allow moves
+        if source_posture == UNKNOWN:
+            return False
```
pieleric: It's allowed to move from UNKNOWN to LOADING.

src/odemis/acq/move.py
```diff
@@ -897,6 +1007,120 @@ def _transform_from_fib_to_milling(self, pos: Dict[str, float]) -> Dict[str, flo
...
+        # For FM @ Milling angle we only allow to go to and from the milling posture.
```
pieleric: It should be possible to short-cut a little bit the code with GRID1/GRID2. I think, except from UNKNOWN and LOADING, it's always possible to go to GRID1/GRID2.

Note: GRID 1/GRID 2 are really apart. They can never be a source posture. Eventually we should change the code to not mix them with a posture.

src/odemis/acq/move.py
```diff
@@ -1088,6 +1323,11 @@ def _do_cryo_switch_sample_position(self, future, target):
 
             # get the current label
             current_label = self.get_current_posture_label()
+            if not self.is_posture_switch_allowed(current_label, target):
+                raise ValueError(
+                    "Moving from posture {} to posture {} is not allowed.".format(current_label, target)
+                )
```
pieleric: In new code, please use f-strings. It's easier to read.

src/odemis/acq/move.py
```diff
@@ -1375,6 +1617,55 @@ def _transform_from_fib_to_sem(self, pos: Dict[str, float]) -> Dict[str, float]:
...
+
+    def _transform_from_milling_to_fm_milling(self, pos: Dict[str, float]) -> Dict[str, float]:
+        """
+        Transforms the stage position from milling to fm milling imaging area.
+        :param pos: (dict str->float) the current stage position
+        :return: (dict str->float) the transformed stage position.
+        """
+        stage_md = self.stage.getMetadata()
+        md_calib = stage_md[model.MD_CALIB]
+        # Define values that are used more than once
```
pieleric: Drop this comment. Typically, this is not correct, as `rx_fm_milling` is only once, but anyway, this is the basic of using variables, no need to explain that.

src/odemis/acq/move.py
```diff
@@ -1375,6 +1617,55 @@ def _transform_from_fib_to_sem(self, pos: Dict[str, float]) -> Dict[str, float]:
...
 
+    def _transform_from_fm_milling_to_milling(self, pos: Dict[str, float]) -> Dict[str, float]:
+        """
+        Transforms the stage position from fm milling to the milling imaging area.
+        :param pos: (dict str->float) the current stage position
+        :return: (dict str->float) the transformed stage position.
+        """
+        stage_md = self.stage.getMetadata()
+        md_calib = stage_md[model.MD_CALIB]
+        rx_milling = self.get_posture_orientation(MILLING)["rx"]
```
pieleric, line 1627: Could you add a note here that typically the ion-beam and FM objective are almost parallel so the angle should be very close (but not identical) to the current angle.

src/odemis/acq/move.py
```diff
@@ -2105,12 +2487,17 @@ def _do_cryo_switch_sample_position(self, future, target):
...
             # To hold the ordered sub moves list
             sub_moves = []  # list of tuples (component, position)
 
             # get the current label
             current_pos = self.stage.position.value
             current_posture = self.get_current_posture_label(current_pos)
+            if not self.is_posture_switch_allowed(current_posture, target):
+                raise ValueError(
+                    "Moving from posture {} to posture {} is not allowed.".format(current_posture, target)
+                )
```
pieleric: f-string

src/odemis/acq/move.py
```diff
@@ -897,6 +1016,122 @@ def _transform_from_fib_to_milling(self, pos: Dict[str, float]) -> Dict[str, flo
...
+        """
+        # Always allow to "stay" at the posture
+        if source_posture == target_posture:
+            return True
+        # Overrides based on specific workflows
+        # From unknown position we never allow moves, except to the loading position
+        if source_posture == UNKNOWN and target_posture != LOADING:
+            return False
+        elif source_posture == UNKNOWN and target_posture == LOADING:
+            return True
```
pieleric: I think it'd be clearer to write, because then it's very obvious you handle every case of UNKNOWN, and don't need to deal with it anymore:
```suggestion
        if source_posture == UNKNOWN:
            if target_posture == LOADING:
                return True
            else:
                return False
```

pieleric (changes requested):

src/odemis/acq/move.py
tmoerkerken: I moved the grid shortcut to the top and extended it to cover LOADING and UNKNOWN.

src/odemis/acq/move.py
tmoerkerken: I agree, there were 2 other instances where I added that line, and there are 6 other occasions in this module (it's left there after copy pasting). I removed my additions. I guess we could let AI pass all the comments and check for usefulness at some point. I'll add a tech debt item https://delmic.atlassian.net/browse/SWM-253.

src/odemis/acq/move.py
tmoerkerken: I tried to improve other parts in a similar manner as well

src/odemis/acq/move.py
```diff
@@ -897,6 +1016,123 @@ def _transform_from_fib_to_milling(self, pos: Dict[str, float]) -> Dict[str, flo
...
+        Check if it is allowed to move from source to target posture. This is a combination of availability of
+        transformation matrices between postures and workflow specific restrictions.
+        :param source_posture: the source posture
+        :param target_posture: the target posture
+        :return: whether it is allowed to move from source to target posture
+        """
+        # Always allow to "stay" at the posture
+        if source_posture == target_posture:
+            return True
+        # It is always accepted to go to the loading position.
```
pieleric, line 1111: I stand by my previous suggestion to have a dedicated `if target_posture in [GRID_1, GRID_2]:` at the beginning. These is a very different case from the rest, and so would simplify to have it explicitly handled apart. Also, we should eventually consider GRID switching as a different move from posture switching. 
So it'd be easier to update the code if GRID_* is not mixed with the rest.

src/odemis/gui/img/icon/ico_meteor_fib_view_fm.png
pieleric, line 1: Let's adjust the name too. How about `ico_meteor_fib_view_fm.png`?

pieleric (changes requested):

tepals (approved):

pieleric (approved):

PR #3484, opened 2026-06-11
[fix] comp text test for headless env

nandishjpatel:
directly invoke the ctrl's focus handler if SetFocus() didn't succeed in case of a headless display environment

### DIFF
--- a/src/odemis/gui/test/comp_text_test.py
+++ b/src/odemis/gui/test/comp_text_test.py
@@ -112,6 +112,18 @@ class NumberTextCtrlTestCase(test.GuiTestCase):
                     ctrl.WriteText(c)
             test.gui_loop(0.02)
 
+    def _set_focus(self, ctrl):
+        """Set focus; if focus can't be obtained (e.g. headless CI), synthesize a focus event to trigger selection."""
+        ctrl.SetFocus()
+        test.gui_loop(0.1)
+
+        if not ctrl.HasFocus():
+            # Headless fallback: directly invoke the focus handler
+            evt = wx.FocusEvent(wx.wxEVT_SET_FOCUS, ctrl.Id)
+            evt.SetEventObject(ctrl)
+            ctrl.GetEventHandler().ProcessEvent(evt)
+            test.gui_loop(0.1)
+
     def test_int_txt_ctrl(self):
 
         ctrl = IntegerTextCtrl(self.panel, value=123456789)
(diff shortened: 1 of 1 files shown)


tepals (approved):

src/odemis/gui/test/comp_text_test.py
```diff
@@ -112,6 +112,18 @@ def _simulate_typing(self, ctrl, text):
...
+        """Set focus; if focus can't be obtained (e.g. headless CI), synthesize a focus event to trigger selection."""
+        ctrl.SetFocus()
+        test.gui_loop(0.1)
+
+        if not ctrl.HasFocus():
+            # Headless fallback: directly invoke the focus handler
+            evt = wx.FocusEvent(wx.wxEVT_SET_FOCUS, ctrl.Id)
+            evt.SetEventObject(ctrl)
+            ctrl.GetEventHandler().ProcessEvent(evt)
+            test.gui_loop(0.1)
```
pieleric, line 125: Did you verify that this works? What do you mean by "headless"? As far as I understood, our VMs run with a simulated screen, with Gnome running. Why would it be "headless"? Do you know why such setup work differently from a normal computer?

nandishjpatel: I haven't tested this fix.

This fix is needed for Ubuntu 24, where `xvfb` a virtual display is needed to run the testcases. By headless I mean the testcases/programs run headless in the `xvfb` virtual display.

Ubuntu 24 testcases run command: https://bitbucket.org/delmic/odemis-testing/src/30268666bced16e45931ccaad349d5db38b6017c/ci/canonical_lxd/action.sh#lines-58

pieleric: @nandishjpatel I think that there are two options forward:
* we find a way to test this fix without merging it. Maybe one way is to temporarily use this branch as target for testing on the Ubuntu 24 run. Another option would be to start the Ubuntu 24 image on your computer and run the test suite (or just this GUI part).
* we merge this change, wait for the test cases to run, and if it doesn't help, we revert it.

I'd rather go with the first route. Can you try finding a way to do it?

nandishjpatel: I went for the first option and it seems to work (used this PR code in Ubuntu 24 CI server image and just ran single comp_text_test):

```
===============================================
Running unit tests in python3
Running /home/ubuntu/development/odemis/src/odemis/gui/test/comp_text_test.py...


All tests passed
```

Image

pieleric (approved):

PR #3487, opened 2026-06-11
[feat] save drift corrector images

nandishjpatel:


src/odemis/acq/drift/__init__.py
```diff
@@ -143,6 +153,15 @@ def acquire(self):
             if data.shape[::-1] != self._res:
                 logging.warning("Shape of data is %s instead of %s", data.shape[::-1], self._res)
 
+            # Save all the drift region scans for offline autocorrelation purposes
+            if self._save_images:
```
nandishjpatel: @pieleric The changes in src/odemis/acq/drift/__init__.py are made by @yxkdejong. I am not 100% sure if this is what AMOLF needs. Since you have previously written code for the drift corrector, can you tell?

src/odemis/acq/drift/__init__.py
```diff
@@ -20,20 +20,24 @@
...
 import cv2
 
 from odemis import model
 from odemis.acq.align.shift import MeasureShift
+from odemis.gui.util import get_picture_folder
 
 MIN_RESOLUTION = (20, 20)  # sometimes 8x8 works, but it's not reliable enough
 MAX_PIXELS = 128 ** 2  # px
 
+DRIFT_IMAGES_DIR = os.path.join(get_picture_folder(), "Drift Correction images")
```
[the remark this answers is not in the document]
tepals: @nandishjpatel can you look at this comment. You could also leave it completely up to the caller/plugin and add drift_images_dir as an argument to the `AnchoredEstimator` class

tepals (changes requested):

plugins/sparc_save_drift_corrector_images.py
```diff
@@ -0,0 +1,62 @@
...
+            logging.debug(
+                "Not loading save drift corrector images tool since SPARC acquisition tab is not present."
+            )
+            return
+
+        self._drift_corrector = sparc_acq_tab.tab_data_model.driftCorrector
+        self.addMenu("Help/Development/Save drift corrector images",
+                     self._save_drift_corrector_images,
+                     item_kind=wx.ITEM_CHECK,
+                     pass_menu_item=True)
```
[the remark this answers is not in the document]
pieleric: @nandishjpatel , for clarification, you could add a comment here that states the plugins are always initialised after the rest of the GUI, so the driftCorrector is always initialised.
Alternatively, for even more confidence, you could just store the tab controller, and explicitly access the driftCorrector inside `_save_drift_corrector_images()`

src/odemis/acq/drift/__init__.py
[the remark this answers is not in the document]
pieleric: That's actually my biggest issue with this original patch: it stores the drift data at a different place from the acquisition file, and mix it with the rest of the standard acquisitions. Instead, it could use the argument `log_path=` (as used in other functions), for indicating where to store (and could even include the beginning of the filename). If `None`, it doesn't record anything.
Then, we can implement once of the two:
* Always record in a dedicated folder in the same place as where the mirror alignment function stores the data ($HOME/odemis-status/)
* In the stream tab, when the filename (and path) is changed, the drift estimator leech `.log_path`  attribute is updated. (instead of `.save_images`)

I would probably go for the second option, as it seems more targetted at the user than the developers, but I don't have strong opinion.

pieleric (changes requested):

src/odemis/acq/drift/__init__.py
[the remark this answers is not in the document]
nandishjpatel: I went with the second option

plugins/sparc_save_drift_corrector_images.py
```diff
@@ -0,0 +1,65 @@
...
+        self._sparc_acq_tab = sparc_acq_tab
+        self.addMenu("Help/Development/Save drift corrector images",
+                     self._save_drift_corrector_images,
+                     item_kind=wx.ITEM_CHECK,
+                     pass_menu_item=True)
+
+    def _on_sparc_acq_ctrl_filename(self, filename):
+        self._sparc_acq_tab.tab_data_model.driftCorrector.log_path = filename
+
+    def _save_drift_corrector_images(self, menu_item):
```
pieleric, line 57: Please add type annotation to all methods.

src/odemis/acq/drift/__init__.py
```diff
@@ -143,6 +147,17 @@ def acquire(self):
...
                 logging.warning("Shape of data is %s instead of %s", data.shape[::-1], self._res)
 
+            # Save all the drift region scans for offline autocorrelation purposes
+            if self._log_path is not None:
+                filename = os.path.basename(self._log_path)
+                if not filename:
+                    raise ValueError("Filename is not found on log path.")
+                exporter = dataio.find_fittest_converter(filename)
+                path, base = os.path.split(filename)
+                fn  = f"drift_{self._image_counter:05d}_" + base
```
pieleric, line 157: If you add drift at the beginning of the filename, all the "drift_xxx" file images will be next to each other in the directory. I think it'd be nicer to use `"{filename}_drift_{self._image_counter:05d}_"`. You can use `dataio.splitext()` for finding the place just before the extension.

pieleric (changes requested):
The most important comments are from code rabbit.

src/odemis/acq/drift/__init__.py
```diff
@@ -143,6 +147,17 @@ def acquire(self):
             if data.shape[::-1] != self._res:
                 logging.warning("Shape of data is %s instead of %s", data.shape[::-1], self._res)
 
+            # Save all the drift region scans for offline autocorrelation purposes
+            if self._log_path is not None:
+                filename = os.path.basename(self._log_path)
+                if not filename:
+                    raise ValueError("Filename is not found on log path.")
+                exporter = dataio.find_fittest_converter(filename)
+                path, base = os.path.split(filename)
```
tepals, line 156: filename is the basename of log_path (see line 152), when you split that (line 156) the path variable is just going to be an empty string.
```suggestion
                path, filename = os.path.split(self._log_path)
                if not filename:
                    raise ValueError("Filename is not found on log path.")
                exporter = dataio.find_fittest_converter(filename)
```

tepals: If you do this, you also need to change `base` to `filename` on line 157

tepals (changes requested):

PR #3488, opened 2026-06-11
[fix] post refactoring get settings order from the stream itself

nandishjpatel:


tepals (approved):

pieleric (approved):

src/odemis/acq/test/stream_settings_test.py
```diff
@@ -132,7 +132,7 @@ def test_get_streams_settings(self):
         # get the saved data from the JSON file
         data = self.acq_settings.config_data[index]
         # check the values of the given stream with the values saved in the JSON file
-        settings_order = get_settings_order(fms)
+        settings_order = fms._get_settings_order()
```
K4rishma, line 135: is this function already present and that is why it is not part of this PR?

K4rishma (approved):

src/odemis/acq/test/stream_settings_test.py
pieleric: @nandishjpatel For such fix, it's best to indicate in the commit comment since when it's broken (which commit broke it). Then, it's easier to know whether to backport the fix to the stable version, and questions such as Karishma's one are easier to answer :-)

PR #3493, opened 2026-06-17
[GANIL-154] create a odemis backend component to control the nd filter

nandishjpatel:
[feat] created MCCDeviceDIOActuator which is an actuator with binary (ON/OFF) axes controlled via MCC DAQ digital outputs, with position feedback via separate digital inputs

### DIFF
--- a/src/odemis/driver/_mccdaq.py
+++ b/src/odemis/driver/_mccdaq.py
@@ -663,12 +663,3 @@ class usb_1208LS(usb_1208):
 
         # enable non-blocking mode
         self.h.set_nonblocking(1)
-
-        # set default configuration
-        self.DConfig(self.DIO_PORTA, 0x00)  # Port A output
-        self.DConfig(self.DIO_PORTB, 0x00)  # Port B output
-        # TODO check if these DOut calls are necessary
-        self.DOut(self.DIO_PORTA, 0x0)  # all ports set to 0
-        self.DOut(self.DIO_PORTB, 0x0)  # all ports set to 0
-        self.AOut(0, 0x0)
-        self.AOut(1, 0x0)
--- a/src/odemis/driver/pwrmccdaq.py
+++ b/src/odemis/driver/pwrmccdaq.py
@@ -364,6 +364,231 @@ class MCCDeviceLight(Emitter, MCCDevice):
                     self.device.DBitOut(port, bit, new_bit_value)
 
 
+class MCCDeviceDIOActuator(model.Actuator, MCCDevice):
+    """
+    Actuator with binary (ON/OFF) axes controlled via MCC DAQ digital outputs,
+    with position feedback via separate digital inputs.
+    Follows the same pattern as TMCM do_axes for binary axis control.
+    """
+
+    def __init__(self, name: str, role: str, mcc_device: Optional[str],
+                 do_axes: Dict[int, Tuple[str, str, str, float]],
+                 di_feedback: Dict[str, Tuple[int, bool]],
+                 di_channels: Dict[int, Tuple[str, bool]] = {},
+                 **kwargs):
+        """
+        :param mcc_device: serial number or None for auto-detect.
+            If "fake" is passed the simulator is used.
+        :param do_axes:
+            Digital output channel -> [axis_name, high_position_name, low_position_name, transition_duration_s].
+            Each entry defines a binary axis controlled via a DO channel.
+            Example: {0: ["nd-filter", "on", "off", 0.5]}
+            The axis will have two choices: "on" and "off" (or any two string values).
+            The transition_duration is the time (s) to wait for the physical mechanism to settle.
+        :param di_feedback:
+            axis_name -> [DI channel number, ttl_high].
+            The DI channel used to read back the actual position of the axis.
+            ttl_high=True means DI high corresponds to the "high" position of the axis.
+            Example: {"nd-filter": [8, True]}
+        :param di_channels: refer to MCCDevice parent.
+            Additional DI channels for general status monitoring (e.g., interlock).
+        """
+        # Validate do_axes
+        if not do_axes:
+            raise ValueError("do_axes must define at least one axis")
+
+        self._do_axes = {}  # int (channel) -> (axis_name, hpos, lpos, duration)
+        self._name_to_do_channel = {}  # str (axis_name) -> int (DO channel)
+        self._di_feedback = {}  # str (axis_name) -> (int DI channel, bool ttl_high)
+
+        axes_def = {}
+        for channel, axis_info in do_axes.items():
+            if len(axis_info) != 4:
+                raise ValueError(
+                    f"do_axes expects [name, high_pos, low_pos, duration] for channel {channel}, "
+                    f"got {axis_info}")
+            an, hpos, lpos, dur = axis_info
+            if an in self._name_to_do_channel:
+                raise ValueError(f"Axis '{an}' specified multiple times in do_axes")
+            if not 0 <= dur < 1000:
+                raise ValueError(f"Axis '{an}' duration {dur} should be between 0 and 1000 s")
+            if hpos == lpos:
+                raise ValueError(f"Axis '{an}' high and low positions must be different")
+
+            self._do_axes[channel] = (an, hpos, lpos, dur)
+            self._name_to_do_channel[an] = channel
+            axes_def[an] = model.Axis(choices={lpos, hpos})
+
+        # Validate di_feedback
+        for an, fb_info in di_feedback.items():
+            if an not in self._name_to_do_channel:
+                raise ValueError(f"di_feedback references unknown axis '{an}'")
+            if len(fb_info) != 2:
+                raise ValueError(
+                    f"di_feedback expects [channel, ttl_high] for axis '{an}', got {fb_info}")
+            di_ch, ttl_high = fb_info
+            self._di_feedback[an] = (di_ch, ttl_high)
+
+        # Merge feedback DI channels into di_channels so MCCDevice configures them as inputs.
+        # Use internal names that won't conflict with user-provided di_channels.
+        all_di_channels = dict(di_channels)
+        for an, (di_ch, ttl_high) in self._di_feedback.items():
+            if di_ch in all_di_channels:
+                raise ValueError(
+                    f"di_feedback channel {di_ch} for axis '{an}' conflicts with di_channels")
+            # Add a hidden DI channel for feedback; the VA name is internal
+            all_di_channels[di_ch] = [f"_feedback_{an}", ttl_high]
+
+        model.Actuator.__init__(self, name, role, axes=axes_def, **kwargs)
+        MCCDevice.__init__(self, name, role, mcc_device, all_di_channels)
+
+        self._executor = model.CancellableThreadPoolExecutor(max_workers=1)
+
+        # Set all outputs to low (safe state) at init
+        for channel, (an, hpos, lpos, dur) in self._do_axes.items():
+            port, bit = self.channel_to_port(channel)
+            with self._connection_lock:
+                self.device.DBitOut(port, bit, 0)
+
+        self.position = model.VigilantAttribute({}, readonly=True)
+        self._updatePosition()
+
+        self._metadata = {model.MD_HW_NAME: f"{self.device.getManufacturer()} "
+                                            f"{self.device.getProduct()} "
+                                            f"{self.device.getSerialNumber()}"}
+        self._swVersion = odemis.__version__
+        self._metadata[model.MD_SW_VERSION] = self._swVersion
+        self._metadata[model.MD_HW_VERSION] = self._hwVersion
+
+    def _updatePosition(self):
+        """Read the actual position from the DI feedback channels."""
+        pos = {}
+        for channel, (an, hpos, lpos, _) in self._do_axes.items():
+            if an in self._di_feedback:
+                di_ch, ttl_high = self._di_feedback[an]
+                port, bit = self.channel_to_port(di_ch)
+                with self._connection_lock:
+                    val = self.device.DBitIn(port, bit)
+                if bool(val) == ttl_high:
+                    pos[an] = hpos
+                else:
+                    pos[an] = lpos
+            else:
+                # No feedback channel: report based on what was last written
+                port, bit = self.channel_to_port(channel)
+                with self._connection_lock:
+                    val = self.device.DBitIn(port, bit)
+                pos[an] = hpos if val else lpos
+
+        self.position._set_value(pos, force_write=True)
+        logging.debug(f"Updated position: {pos}")
+
+    def _createMoveFuture(self):
+        """Return a CancellableFuture for a move."""
+        f = model.CancellableFuture()
+        f._moving_lock = threading.Lock()
+        f._must_stop = threading.Event()
+        f._was_stopped = False
+        f.task_canceller = self._cancelCurrentMove
+        return f
+
+    @model.isasync
+    def moveAbs(self, pos):
+        if not pos:
+            return model.InstantaneousFuture()
+        self._checkMoveAbs(pos)
+
+        f = self._createMoveFuture()
+        self._executor.submitf(f, self._doMoveAbs, f, pos)
+        return f
+
+    moveAbs.__doc__ = model.Actuator.moveAbs.__doc__
+
+    @model.isasync
+    def moveRel(self, shift):
+        raise NotImplementedError("Relative move is not supported on binary axes")
+
+    moveRel.__doc__ = model.Actuator.moveRel.__doc__
+
+    def stop(self, axes=None):
+        self._executor.cancel()
+
+    def _checkMoveAbs(self, pos):
+        for axis, val in pos.items():
+            if axis not in self.axes:
+                raise ValueError(f"Unknown axis '{axis}'")
+            axis_def = self.axes[axis]
+            if hasattr(axis_def, "choices") and val not in axis_def.choices:
+                raise ValueError(
+                    f"Unsupported position '{val}' for axis '{axis}', "
+                    f"choices are {axis_def.choices}")
+
+    def _doMoveAbs(self, future, pos):
+        """
+        Blocking and cancellable absolute move.
+        Sets the DO pin and waits for the transition duration.
+        """
+        with future._moving_lock:
+            end = 0
+            for an, v in pos.items():
+                channel = self._name_to_do_channel[an]
+                _, hpos, _, dur = self._do_axes[channel]
+
+                port, bit = self.channel_to_port(channel)
+                new_val = int(v == hpos)
+                logging.debug("Setting DO channel %d to %d for axis '%s' -> '%s'",
+                              channel, new_val, an, v)
+                with self._connection_lock:
+                    self.device.DBitOut(port, bit, new_val)
+                end = max(end, time.time() + dur)
+
+            # Wait for the transition, checking for cancellation
+            self._waitEndMove(future, end)
+
+        logging.debug("Move successfully completed")
+
+    def _waitEndMove(self, future, end):
+        """
+        Wait until the transition duration has elapsed or cancellation is requested.
+        """
+        while not future._must_stop.is_set():
+            left = end - time.time()
+            if left <= 0:
+                break
+            sleept = max(0.001, min(left / 2, 0.1))
+            future._must_stop.wait(sleept)
+        else:
+            future._was_stopped = True
+            raise model.CancelledError()
+
+        self._updatePosition()
+
+    def _cancelCurrentMove(self, future):
+        """Cancel the current move."""
+        future._must_stop.set()
+        with future._moving_lock:
+            if not future._was_stopped:
+                logging.debug("Cancelling current move")
+            return future._was_stopped
+
+    def terminate(self):
+        if self._executor:
+            self._executor.cancel()
+            self._executor.shutdown()
+
+        # Set all outputs to low (safe state) on shutdown
+        if self.device:
+            for channel in self._do_axes:
+                port, bit = self.channel_to_port(channel)
+                try:
+                    with self._connection_lock:
+                        self.device.DBitOut(port, bit, 0)
+                except Exception:
+                    logging.warning("Failed to reset DO channel %d on terminate", channel)
+
+        super().terminate()
+
+
 class MCCDeviceSimulator:
     """
     A really basic and simple interface to simulate a USB-1208LS device with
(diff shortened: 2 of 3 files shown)


src/odemis/driver/pwrmccdaq.py
```diff
@@ -364,6 +364,231 @@ def _update_power(self, value: List[float]):
                     self.device.DBitOut(port, bit, new_bit_value)
 
 
+class MCCDeviceDIOActuator(model.Actuator, MCCDevice):
```
pieleric: It shouldn't inherit from MCCDevice, because if you do this, that means it's the only component. However, we want to also have a `MCCDeviceLight` component. Instead, you should use the "children" mechanism: the MCCDecice(Light) will get a `children` argument with "actuator" (for instance) as "internal role" (and kwargs associated to it). When the MCCDevice code detects such children is requested, it takes care of instantiating the MCCDeviceDIOActuator (passing kwargs and itself as parent).

The rest of the code in MCCDeviceDIOActuator can stay almost identical, with the exception that it has to use self._parent for the lower level functions.
See the "zeiss" or "tescan" drivers for examples.

src/odemis/driver/pwrmccdaq.py
```diff
@@ -364,6 +364,231 @@ def _update_power(self, value: List[float]):
...
+            Example: {0: ["nd-filter", "on", "off", 0.5]}
+            The axis will have two choices: "on" and "off" (or any two string values).
+            The transition_duration is the time (s) to wait for the physical mechanism to settle.
+        :param di_feedback:
+            axis_name -> [DI channel number, ttl_high].
+            The DI channel used to read back the actual position of the axis.
+            ttl_high=True means DI high corresponds to the "high" position of the axis.
+            Example: {"nd-filter": [8, True]}
+        :param di_channels: refer to MCCDevice parent.
+            Additional DI channels for general status monitoring (e.g., interlock).
```
pieleric: If you use it as a child, this can be dropped.

src/odemis/driver/pwrmccdaq.py
```diff
@@ -364,6 +364,231 @@ def _update_power(self, value: List[float]):
...
+    """
+
+    def __init__(self, name: str, role: str, mcc_device: Optional[str],
+                 do_axes: Dict[int, Tuple[str, str, str, float]],
+                 di_feedback: Dict[str, Tuple[int, bool]],
+                 di_channels: Dict[int, Tuple[str, bool]] = {},
+                 **kwargs):
+        """
+        :param mcc_device: serial number or None for auto-detect.
+            If "fake" is passed the simulator is used.
```
pieleric: When used as a child, this should be replaced by "parent".

src/odemis/driver/pwrmccdaq.py
```diff
@@ -364,6 +364,231 @@ def _update_power(self, value: List[float]):
...
+            Digital output channel -> [axis_name, high_position_name, low_position_name, transition_duration_s].
+            Each entry defines a binary axis controlled via a DO channel.
+            Example: {0: ["nd-filter", "on", "off", 0.5]}
+            The axis will have two choices: "on" and "off" (or any two string values).
+            The transition_duration is the time (s) to wait for the physical mechanism to settle.
+        :param di_feedback:
+            axis_name -> [DI channel number, ttl_high].
+            The DI channel used to read back the actual position of the axis.
+            ttl_high=True means DI high corresponds to the "high" position of the axis.
+            Example: {"nd-filter": [8, True]}
```
pieleric: I think the advantage is that it's similar to the TMCM driver, but it is quite complicated. I'd suggest to use a different syntaxes:
```
"axes": {
    "nd-filder": {  # axis name
          "do": 0,  # channel number for setting the position
          "low": 0.0,  # position reported when low
          "high": 1.57,  # position reported when high
          "unit": "rad",  # optional: the unit of the axis
          "duration": 0.5,  # transition time
          "di": 8, # optional channel number for reading the position
          "di_inverted": False, # optional, whether the position read is the same as written, or inverted
    }
}
```

src/odemis/driver/pwrmccdaq.py
```diff
@@ -364,6 +364,231 @@ def _update_power(self, value: List[float]):
...
+                 di_feedback: Dict[str, Tuple[int, bool]],
+                 di_channels: Dict[int, Tuple[str, bool]] = {},
+                 **kwargs):
+        """
+        :param mcc_device: serial number or None for auto-detect.
+            If "fake" is passed the simulator is used.
+        :param do_axes:
+            Digital output channel -> [axis_name, high_position_name, low_position_name, transition_duration_s].
+            Each entry defines a binary axis controlled via a DO channel.
+            Example: {0: ["nd-filter", "on", "off", 0.5]}
```
pieleric: Typically, in Odemis we use floats for the position of the axes. If you want to associate a user-friendly name to a position, you can use a dict in .choices: float -> str. (eg: {0.0: "on", 1.57: "off"})

src/odemis/driver/pwrmccdaq.py
```diff
@@ -364,6 +364,231 @@ def _update_power(self, value: List[float]):
...
+        # Set all outputs to low (safe state) on shutdown
+        if self.device:
+            for channel in self._do_axes:
+                port, bit = self.channel_to_port(channel)
+                try:
+                    with self._connection_lock:
+                        self.device.DBitOut(port, bit, 0)
+                except Exception:
+                    logging.warning("Failed to reset DO channel %d on terminate", channel)
+
```
pieleric: No. it shouldn't move the axes on shutdown. That's the safest.

src/odemis/driver/pwrmccdaq.py
```diff
@@ -364,6 +364,231 @@ def _update_power(self, value: List[float]):
...
+        model.Actuator.__init__(self, name, role, axes=axes_def, **kwargs)
+        MCCDevice.__init__(self, name, role, mcc_device, all_di_channels)
+
+        self._executor = model.CancellableThreadPoolExecutor(max_workers=1)
+
+        # Set all outputs to low (safe state) at init
+        for channel, (an, hpos, lpos, dur) in self._do_axes.items():
+            port, bit = self.channel_to_port(channel)
+            with self._connection_lock:
+                self.device.DBitOut(port, bit, 0)
```
pieleric: No, don't move the axes at init. Just read the current position.

pieleric (changes requested):

src/odemis/driver/pwrmccdaq.py
nandishjpatel: Thanks @pieleric now I understand what you meant with `children`, I previously thought of it as software inheritance. Made the change.

src/odemis/driver/pwrmccdaq.py
```diff
@@ -130,6 +132,15 @@ def __init__(self, name: str, role: str, mcc_device: Optional[str],
             self._status_thread = MCCDeviceDIStatus(self.device, self._channel_vas, self._connection_lock)
             self._status_thread.start()
 
+        # create DIO actuator child
+        try:
```
pieleric: It'd nicer to read to just have a `if "actuator" in children`, instead of catching exceptions.

src/odemis/driver/test/pwrmccdaq_test.py
```diff
@@ -32,9 +32,25 @@
 logging.getLogger().setLevel(logging.DEBUG)
 TEST_NOHW = (os.environ.get("TEST_NOHW", "0") != "0")  # Default to Hw testing
 
+CONFIG_ND_FILTER = {
+    "name": "ND Filter",
+    "role": "nd-filter",
+    "axes": {
+        "nd-filter": {
```
pieleric: For the test, please add a second axis, without `di` information, and move that second axis. This should help cover more code.

src/odemis/driver/pwrmccdaq.py
pieleric: I don't think there is any advantage in doing this loop, instead of just `future._must_stop.wait(left)`.
In the TMCM, there is a regular update of the axis position, so that would make sense, but not here.

pieleric: If there is a `di`, you could wait until the `di` reports that the axis has really reached its position. Then, it should raise a TimeoutError if not reached after a while. In case you control a laser, it's quite helpful to know that the move failed!

Probably, we should also immediately log a warning if di contradicts the requested position after the waiting time.

src/odemis/driver/test/pwrmccdaq_test.py
```diff
@@ -257,6 +260,42 @@ def test_terminate_di_status(self):
...
+    def test_child_nd_filter_move(self):
+        """
+        Test the ND filter actuator child of the MCCDeviceLight.
+        """
+        self._create_device()
+        actuator = self.mcc_device._actuator
+
+        # Move to high
+        f = actuator.moveAbs({"nd-filter": 1.57})
+        f.result()
```
pieleric, line 356: Please check the duration of the move too. It should be > 0.1s. (for both cases)

pieleric (changes requested):

src/odemis/driver/pwrmccdaq.py
```diff
@@ -364,6 +377,193 @@ def _update_power(self, value: List[float]):
...
+            high = axis_cfg["high"]
+            unit = axis_cfg.get("unit", "")
+            duration = axis_cfg["duration"]
+
+            if low == high:
+                raise ValueError(f"Axis '{axis_name}' 'low' and 'high' positions must be different")
+            if not 0 <= duration < 1000:
+                raise ValueError(f"Axis '{axis_name}' duration {duration} should be between 0 and 1000 s")
+            try:
+                parent.channel_to_port(do_ch)
```
pieleric: I missed something in the first reviews: the device must have the (whole) port (A/B) configured explicitely either for reading or writing. This is done originally by this line `self.device.DConfig(port, val)`. However, until now, that was easy, because the code only needed reading. Now, we need to be extra careful, and make sure that a given port only has either DO or DI channels. It should also be documented in the docstring.

pieleric (changes requested):

src/odemis/driver/test/pwrmccdaq_test.py
```diff
@@ -32,9 +32,32 @@
...
+            "unit": "rad",
+            "duration": 0.5,
+        },
+    },
+}
+
 CONFIG_LIGHT_DEVICE = {
                 "name": "Laser Hub", "role": "light",
-                "mcc_device": None,
+                "mcc_device": "fake",
```
pieleric: Revert to None by default.

src/odemis/driver/pwrmccdaq.py
```diff
@@ -364,6 +379,266 @@ def _update_power(self, value: List[float]):
...
+                except ValueError:
+                    raise ValueError(f"Axis '{axis_name}' has invalid 'di' channel {di_ch}")
+                parent.dconfig[di_port] = 0xff  # set all channel to input
+                self._di_feedback[axis_name] = (di_ch, di_inverted)
+
+        di_ports |= di_ports_parent
+        shared_ports = do_ports & di_ports
+        if shared_ports:
+            raise ValueError(f"DO and DI channels cannot share the same port(s): {shared_ports}. "
+                             f"Port A/B can be used for either DO or DI, but not both.")
```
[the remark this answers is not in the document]
pieleric: It's right. When telling you that we didn't use the DO before, I was wrong. We already used it to control the power on/off of the laser, on the MCCDeviceLight. I've check the configurations: we always used port A for DO and port B for DI. So adding a check that they are on different port shouldn't break any existing system.

src/odemis/driver/pwrmccdaq.py
```diff
@@ -364,6 +383,270 @@ def _update_power(self, value: List[float]):
...
+                )
+
+            logging.debug(
+                "Waiting for DI feedback, current position: %s, target position: %s",
+                current_pos,
+                di_axes
+            )
+            time.sleep(sleept)
+
+    def _cancelCurrentMove(self, future):
```
pieleric: For all new functions/methods, please use the new style: `snake_case` names and type annotations.

src/odemis/driver/pwrmccdaq.py
```diff
@@ -364,6 +383,270 @@ def _update_power(self, value: List[float]):
...
+
+        axes_def = {}
+        do_ports = set()
+        di_ports = set()
+        do_ports_parent = set()
+        if hasattr(parent, "_do_channels"):
+            do_ports_parent = {parent.channel_to_port(ch)[0] for ch in parent._do_channels}
+        di_ports_parent = {parent.channel_to_port(ch)[0] for ch in parent._di_channels}
+        for axis_name, axis_cfg in axes.items():
+            for required in ("do", "low", "high", "duration"):
```
pieleric: You can also make a set of `missing_keys` defined as `{"do", "low", "high", "duration"} - set(axis_cfg.values())`, and immediately report all the missing keys, without a loop.

src/odemis/driver/_mccdaq.py
```diff
@@ -663,12 +663,3 @@ def __init__(self, serial=None):
...
         # enable non-blocking mode
         self.h.set_nonblocking(1)
-
-        # set default configuration
-        self.DConfig(self.DIO_PORTA, 0x00)  # Port A output
-        self.DConfig(self.DIO_PORTB, 0x00)  # Port B output
-        # TODO check if these DOut calls are necessary
-        self.DOut(self.DIO_PORTA, 0x0)  # all ports set to 0
-        self.DOut(self.DIO_PORTB, 0x0)  # all ports set to 0
-        self.AOut(0, 0x0)
```
pieleric, line 673: Just to be extra safe, please put this back in, maybe at the beginning of the MCCDevice driver. 
Now, the MCCDeviceLight does set all the configured AO to 0V. However, the non-configured AO are not reset. On some systems, there are multiple configurations, with each configuration controlling a single voltage. So the other voltages are temporarily not controlled, although still physically connected. The default voltage on the device is probably 0V anyway... but just to be 100% certain we don't break anything on the next update, let's explicitly put all the AOs to 0V.

src/odemis/driver/test/pwrmccdaq_test.py
```diff
@@ -243,6 +267,100 @@ def test_terminate_di_status(self):
...
+        MOD_CONFIG["children"]["actuator"]["axes"]["nd-filter1"]["di"] = 13
+        with self.assertRaises(ValueError):
+            pwrmccdaq.MCCDeviceLight(**MOD_CONFIG)
+
+    def test_child_nd_filter(self):
+        """
+        Test the ND filter actuator child of the MCCDeviceLight.
+        """
+        self._create_device()
+        self.assertTrue(hasattr(self.mcc_device, "_actuator"))
```
pieleric: No, it shouldn't be accessed this way, via a private attribute. The actuator should be on .children.value . It can be found based on the component name given to the Actuator (that's how the backend does it).

pieleric (changes requested):

src/odemis/driver/pwrmccdaq.py
```diff
@@ -364,6 +386,270 @@ def _update_power(self, value: List[float]):
...
+                    self.parent.device.DBitOut(port, bit, new_val)
+                durations.append(duration)
+
+            self._waitEndMove(future, max(durations), pos)
+
+        logging.debug("Move successfully completed")
+
+    def _waitEndMove(self, future, duration, pos):
+        """
+        Wait until the max transition duration has elapsed or cancellation is requested.
```
tmoerkerken, line 606: lacking docstrings for the parameters

src/odemis/driver/pwrmccdaq.py
```diff
@@ -93,8 +98,13 @@ def __init__(self, name: str, role: str, mcc_device: Optional[str],
...
                 raise HwError("Failed to open MCC DAQ device with s/n '%s'" % mcc_device)
 
+        self.init_dio_values = {}  # dict of channel number -> initial bit value read from the device at startup
+        for channel in range(16):
+            port, bit = self.channel_to_port(channel)
+            self.init_dio_values[channel] = self.device.DBitIn(port, bit)
+
         # Default to every channel is output (both ports set to 0)
-        dconfig = {
+        self.dconfig = {
```
tmoerkerken, line 107: Adding it to `self` seems unnecessary since it is only accessed inside `__init__`

tmoerkerken (approved):

src/odemis/driver/pwrmccdaq.py
nandishjpatel: It used in `MCCDeviceDIOActuator` or can be used in any children to check or assign port config, that's why added to `self`

src/odemis/driver/pwrmccdaq.py
```diff
@@ -123,9 +133,18 @@ def __init__(self, name: str, role: str, mcc_device: Optional[str],
                 setattr(self, va_name, va)  # set the class VA variable name
                 logging.info(f"{va_name} status registered for component {self._name} on channel {channel}")
 
-            for port, val in dconfig.items():
-                self.device.DConfig(port, val)
+        for port, val in self.dconfig.items():
+            self.device.DConfig(port, val)
+
+        # set default configuration
+        self.device.AOut(0, 0x0)
```
pieleric, line 140: Also `self.AOut(1, 0x0)`

src/odemis/driver/pwrmccdaq.py
```diff
@@ -364,6 +386,270 @@ def _update_power(self, value: List[float]):
...
+        pos = {}
+        for axis_name, (do_ch, low, high, _) in self._axes_cfg.items():
+            port, bit = self.parent.channel_to_port(do_ch)
+            val = self.parent.init_dio_values[do_ch]
+            pos[axis_name] = high if val else low
+            with self.parent._connection_lock:
+                self.parent.device.DBitOut(port, bit, val)
+        self.position._set_value(pos, force_write=True)
+
+    def _updatePosition(self):
```
pieleric: For *all* the private methods, make sure to:
* use snake_case for the name
* add type annotation (including return)
* add docstring for the parameters and return value

pieleric (changes requested):

src/odemis/driver/pwrmccdaq.py
```diff
@@ -233,6 +254,8 @@ def __init__(self, name: str, role: str, mcc_device: Optional[str], ao_channels:
             At least one pair should be provided. If no voltage is linked to 0W, then a 0V -> 0W mapping is used.
             The total curve should be monotonic.
         :param di_channels (dict -> int, list(str, bool)): refer to parent.
+        :param children: Known children are "actuator" for an MCCDeviceDIOActuator child component which can be used
+            to control binary axes through the digital outputs of the MCC device.
         """
         Emitter.__init__(self, name, role, **kwargs)
```
pieleric, line 262: If you add a `daemon` parameter in the function signature, it's automatically dropped from `kwargs`, so it needs to be explicitly passed here too. Without this parameter, the standard test cases will pass, because they don't depend on the backend, but starting a full system will fail.

src/odemis/driver/pwrmccdaq.py
```diff
@@ -364,6 +387,295 @@ def _update_power(self, value: List[float]):
...
+                )
+
+            logging.debug(
+                "Waiting for DI feedback, current position: %s, target position: %s",
+                current_pos,
+                di_axes
+            )
+            time.sleep(sleept)
+
+    def _cancel_current_move(self, future: model.CancellableFuture):
```
pieleric: Please add type annotation to each function here, including return. This one returns `bool`. Otherwise, if a method doesn't return anything, report `-> None`. 
Also add it to the moveRel, moveAbs functions.

pieleric (changes requested):

src/odemis/driver/pwrmccdaq.py
nandishjpatel: @pieleric, If I pass the `daemon` parameter here, it throws an error. The `Emitter` and `MCCDevice` both initialize the `HwComponent` with the `daemon` arg that's why?

```
2026-07-21 09:38:17,967	INFO	main:311:	Remote exception Traceback (most recent call last):
  File "/home/patel/development/Pyro4/src/Pyro4/core.py", line 861, in handleRequest
    data=method(*vargs, **kwargs)   # this is the actual method call to the Pyro object
  File "/home/patel/development/odemis/src/odemis/model/_core.py", line 281, in instantiate
    return self.daemon.instantiate(klass, kwargs)
  File "/home/patel/development/odemis/src/odemis/model/_core.py", line 362, in instantiate
    comp = klass(**kwargs)
  File "/home/patel/development/odemis/src/odemis/driver/pwrmccdaq.py", line 315, in __init__
    MCCDevice.__init__(self, name, role, mcc_device, di_channels, children=children, daemon=daemon)
  File "/home/patel/development/odemis/src/odemis/driver/pwrmccdaq.py", line 83, in __init__
    super().__init__(name, role, daemon=daemon, **kwargs)
  File "/home/patel/development/odemis/src/odemis/model/_components.py", line 294, in __init__
    Component.__init__(self, name, *args, **kwargs)
  File "/home/patel/development/odemis/src/odemis/model/_components.py", line 121, in __init__
    daemon.register(self, quote(name))
  File "/home/patel/development/Pyro4/src/Pyro4/core.py", line 959, in register
    raise errors.DaemonError("object (%s) already has a Pyro id (%s)" % (obj, obj._pyroId))
Pyro4.errors.DaemonError: object (MCCDeviceLight 'External Laser') already has a Pyro id (External%20Laser)
```

pieleric: Oh, I had forgotten it was inheriting from 2 classes... which inherit from the same grand-parent class. Such "diamond" inheritance is a terrible pain to get a simple feel on how it's behave, especially when mixing super() (and that's one reason so many languages forbid it). 
Anyway, you're right, the `MCCDevice.__init__()` takes care of it.

src/odemis/driver/pwrmccdaq.py
```diff
@@ -123,9 +133,19 @@ def __init__(self, name: str, role: str, mcc_device: Optional[str],
...
+            self.device.DConfig(port, val)
+
+        # set default configuration
+        self.device.AOut(0, 0x0)
+        self.device.AOut(1, 0x0)
+
+        # check if an actuator child is requested and create it
+        if "actuator" in children:
+            self._actuator = MCCDeviceDIOActuator(parent=self, daemon=daemon, **children["actuator"])
+            self.children.value.add(self._actuator)
```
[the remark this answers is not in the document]
pieleric: @nandishjpatel That's a valid remark. Add this code to the .terminate():
```
        for c in self.children.value:
            c.terminate()
```

pieleric (changes requested):

pieleric (approved):

PR #3494, opened 2026-06-17
[MSD-572] [feat] move milling patterns horizontally

K4rishma:
control + shift + arrow keys move short distance and control + arrow keys move large distance

### DIFF
--- a/src/odemis/gui/cont/milling.py
+++ b/src/odemis/gui/cont/milling.py
@@ -67,6 +67,9 @@ MILLING_COLOURS_CANONICAL = {
     "Microexpansion": "#FFA500",
     "Fiducial": "#FF69B4",
 }
+# Step sizes to move the milling patterns horizontally
+MOVE_DELTA_X_SHORT = 1  # px
+MOVE_DELTA_X_LONG = 5  # px
 
 def _get_milling_colour(task_name: str, idx: int) -> str:
     """Get the colour based on the task name or index"""
@@ -166,6 +169,7 @@ class MillingTaskController:
         )
         self.canvas.add_world_overlay(self.rectangles_overlay)
         self.canvas.Bind(wx.EVT_LEFT_DOWN, self.on_mouse_down) # bind the mouse down event
+        self.canvas.Bind(wx.EVT_CHAR, self.on_char)
 
         self.selected_tasks = model.ListVA([])  # List of strings, names of the selected milling tasks
         self._panel.milling_task_chk_list.Bind(wx.EVT_CHECKLISTBOX, handler=self._update_selected_tasks)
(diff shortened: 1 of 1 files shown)


src/odemis/gui/cont/milling.py
```diff
@@ -329,6 +333,73 @@ def on_mouse_down(self, evt):
...
+            if key == wx.WXK_LEFT:
+                view_dx = -step_px
+            elif key == wx.WXK_RIGHT:
+                view_dx = step_px
+            else:
+                view_dx = 0
+
+            # find the center position for any of the milling patterns as they have same central position
+            selected_center_rel = None
+            for task_name in self.selected_tasks.value:
```
tepals: I don't know about selected_tasks, but it feels like this could get slow if selected_tasks is a long list and only at the end there is a task that has patterns. Especially if you slam the arrow keys a lot to move faster

src/odemis/gui/cont/milling.py
```diff
@@ -329,6 +333,73 @@ def on_mouse_down(self, evt):
...
+                task = self.milling_tasks.get(task_name)
+                if task and task.patterns:
+                    selected_center_rel = task.patterns[0].center.value
+                    break
+
+            if selected_center_rel is None:
+                evt.Skip()
+                return
```
tepals: don't you want to look for a task where selected_center_rel is not None?

tepals (changes requested):

src/odemis/gui/cont/milling.py
K4rishma: the selected_tasks are milling patterns which all should have a centre_rel value. The idea is to get the first task, look for this value and break. In the comment above the loop, I have tried to explain about this. Should I remove the for loop?

src/odemis/gui/cont/milling.py
K4rishma: I think all the patterns should have the same centre value. @ilyushkin would that be a correct assumption?

ilyushkin: Currently, yes, but I would not rely on this assumption. I think it would be better to move each sub-pattern individually without supposing that they're all centered.

ilyushkin (commented):
The patterns might not be all centered eventually.

ilyushkin:
Is this action executed before updating the view position with arrow keys? I'm asking because it is also possible to move the view in the IV quadrant of the FIBSEM tab with Ctrl (Ctrl/Shift) + arrows. See odemis/src/odemis/gui/comp/canvas.py DraggableCanvas.on_char().

K4rishma:
> Is this action executed before updating the view position with arrow keys? I'm asking because it is also possible to move the view in the IV quadrant of the FIBSEM tab with Ctrl (Ctrl/Shift) + arrows. See odemis/src/odemis/gui/comp/canvas.py DraggableCanvas.on_char().

It overwrites the default on_char operation of the canvas

src/odemis/gui/cont/milling.py
```diff
@@ -329,6 +333,72 @@ def on_mouse_down(self, evt):
...
+        if key not in valid_keys:
+            evt.Skip()
+            return
+
+        active_canvas = evt.GetEventObject()
+        logging.debug(f"keyboard button pressed event, canvas: {active_canvas}")
+
+        feature = self._tab_data.main.currentFeature.value
+
+        # check if control is pressed, and if a stream is selected
```
tepals: This comment does not seem to match what you are checking, or do `self.allow_milling_pattern_move and feature and feature.reference_image is not None` mean that a stream is selected? Anyway I do not understand why a stream should be selected to move a pattern? Shouldn't the pattern need to be selected?

In general, for comments try to keep in mind that they explain why the code is doing something, not just what the code is doing.

src/odemis/gui/cont/milling.py
```diff
@@ -329,6 +333,72 @@ def on_mouse_down(self, evt):
...
+                step_px = MOVE_DELTA_X_SHORT
+            else:
+                step_px = MOVE_DELTA_X_LONG
+            ref_img = feature.reference_image
+            if key == wx.WXK_LEFT:
+                view_dx = -step_px
+            elif key == wx.WXK_RIGHT:
+                view_dx = step_px
+            else:
+                view_dx = 0
```
tepals: I think this can be removed, it should not happen, because you have the check on line 349

tepals (changes requested):

tepals (approved):

ilyushkin (approved):

PR #3495, opened 2026-06-17
[MSD-302][feat] milling patterns takes spot size input

K4rishma:
The “hole size” parameter of the Essence preset (in XML) currently contains the diameter of the orifice that will be milled out by FIB. The milling boxes displayed in Odemis should circumscribe the beam path to account for so-called beam tails.

### DIFF
--- a/src/odemis/acq/milling/tasks.py
+++ b/src/odemis/acq/milling/tasks.py
@@ -36,13 +36,15 @@ from odemis.acq.milling.patterns import (
 class MillingSettings:
     """Represents milling settings for a single milling task"""
 
-    def __init__(self, current: float, voltage: float, field_of_view: float, mode: str = "Serial", channel: str = "ion", align: bool = True):
+    def __init__(self, current: float, voltage: float, field_of_view: float, mode: str = "Serial",
+                 channel: str = "ion", align: bool = True, spot_size: float = 5.0e-09):
         self.current = model.FloatContinuous(current, unit="A", range=(20e-12, 120e-9))
         self.voltage = model.FloatContinuous(voltage, unit="V", range=(0, 30e3))
         self.field_of_view = model.FloatContinuous(field_of_view, unit="m", range=(50e-06, 960e-06))
         self.mode = model.StringEnumerated(mode, choices={"Serial", "Parallel"})
         self.channel = model.StringEnumerated(channel, choices={"ion"})
-        self.align = model.BooleanVA(align) # align at the milling current
+        self.align = model.BooleanVA(align)  # align at the milling current
+        self.spot_size = model.FloatContinuous(spot_size, unit="m", range=(5.0e-09, 1.0e-06))  # FIB beam spot size (diameter)
 
     def to_dict(self) -> dict:
         return {"current": self.current.value,
@@ -50,7 +52,8 @@ class MillingSettings:
                 "field_of_view": self.field_of_view.value,
                 "mode": self.mode.value,
                 "channel": self.channel.value,
-                "align": self.align.value}
+                "align": self.align.value,
+                "spot_size": self.spot_size.value}
 
     @staticmethod
     def from_dict(data: dict) -> "MillingSettings":
@@ -59,7 +62,8 @@ class MillingSettings:
                                 field_of_view=data["field_of_view"],
                                 mode=data.get("mode", "Serial"),
                                 channel=data.get("channel", "ion"),
-                                align=data.get("align", True)
+                                align=data.get("align", True),
+                                spot_size=data.get("spot_size", 5.0e-09)
                                 )
 
     def __repr__(self):
--- a/src/odemis/gui/comp/milling.py
+++ b/src/odemis/gui/comp/milling.py
@@ -67,6 +67,7 @@ class MillingTaskPanel(wx.Panel):
             "current": {"label": "Current", "accuracy": 2, "unit": "A"},
             "align": {"label": "Align at milling current"},
             "mode": {"label": "Milling mode"},
+            "spot_size": {"label": "Spot Size", "accuracy": 2, "unit": "m"},
             "width": {"label": "Width", "accuracy": 2, "unit": "m"},
             "height": {"label": "Height", "accuracy": 2, "unit": "m"},
             "depth": {"label": "Depth", "accuracy": 2, "unit": "m"},
@@ -83,8 +84,7 @@ class MillingTaskPanel(wx.Panel):
                 continue
 
             conf = CONFIG.get(param, {})
-            label = conf.get("label", param)
-            del conf["label"]
+            label = conf.pop("label", None)
 
             val = getattr(task.milling, param)
             self._add_value_field(label, val, conf, param=param)
@@ -97,8 +97,7 @@ class MillingTaskPanel(wx.Panel):
                 continue
 
             conf = CONFIG.get(param, {})
-            label = conf.get("label", param)
-            del conf["label"]
+            label = conf.pop("label", None)
 
             val = getattr(pattern, param)
             self._add_value_field(label, val, conf, param=param)
--- a/src/odemis/gui/cont/milling.py
+++ b/src/odemis/gui/cont/milling.py
@@ -100,13 +100,28 @@ def pos_to_absolute(pos: Tuple[float, float], ref_img: model.DataArray) -> Tuple
 def rectangle_pattern_to_shape(canvas,
                         ref_img: model.DataArray,
                         pattern: RectanglePatternParameters,
+                        spot_size: float,
                         colour: str = "#FFFF00",
                         name: str = None) -> EditableShape:
-    """Convert a rectangle pattern to a shape"""
-    rect = RectangleOverlay(cnvs=canvas, colour = colour, show_selection_points = False)
+    """
+    Convert a rectangle pattern to a shape, taking spot size of the beam into account.
+
+    :param canvas: Canvas object to draw shapes on
+    :param ref_img: Reference image on which the shapes are drawn
+    :param pattern: provide type of shape and its parameters
+    :param spot_size: spot size in m (beam diameter)
+    :param colour: colour of the shape
+    :param name: name of the shape, e.g. Rough Milling, Polishing etc.
+    """
+    rect = RectangleOverlay(cnvs=canvas, colour=colour, show_selection_points=False)
     width = pattern.width.value
     height = pattern.height.value
-    x, y = pos_to_absolute(pattern.center.value, ref_img) # image coordinates -> physical coordinates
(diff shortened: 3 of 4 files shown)


src/odemis/gui/cont/milling.py
```diff
@@ -101,12 +101,19 @@ def rectangle_pattern_to_shape(canvas,
                         ref_img: model.DataArray,
                         pattern: RectanglePatternParameters,
                         colour: str = "#FFFF00",
-                        name: str = None) -> EditableShape:
-    """Convert a rectangle pattern to a shape"""
-    rect = RectangleOverlay(cnvs=canvas, colour = colour, show_selection_points = False)
+                        name: str = None,
+                        spot_size: float = 0.0) -> EditableShape:
+    """
+    Convert a rectangle pattern to a shape, optionally expanding for beam spot.
```
tepals: Can you update the docstring to include the parameters? And can you explain the unit of spot_size and why the default is 0.0?

tepals (changes requested):

src/odemis/acq/milling/tasks.py
```diff
@@ -36,21 +36,24 @@
 class MillingSettings:
     """Represents milling settings for a single milling task"""
 
-    def __init__(self, current: float, voltage: float, field_of_view: float, mode: str = "Serial", channel: str = "ion", align: bool = True):
+    def __init__(self, current: float, voltage: float, field_of_view: float, mode: str = "Serial",
+                 channel: str = "ion", align: bool = True, spot_size: float = 50.0e-09):
```
pieleric: Put a small value by default, so that it doesn't change too much the current systems => use the smallest one (15e-9) or even smaller: 10e-9?

src/odemis/acq/milling/tasks.py
```diff
@@ -59,7 +62,8 @@ def from_dict(data: dict) -> "MillingSettings":
                                 field_of_view=data["field_of_view"],
                                 mode=data.get("mode", "Serial"),
                                 channel=data.get("channel", "ion"),
-                                align=data.get("align", True)
+                                align=data.get("align", True),
+                                spot_size=data.get("spot_size", 50.0e-09)
```
pieleric: Update the default value also here.

src/odemis/gui/comp/milling.py
```diff
@@ -84,7 +85,7 @@ def __init__(self, parent, task: MillingTaskSettings):
 
             conf = CONFIG.get(param, {})
             label = conf.get("label", param)
-            del conf["label"]
+            conf.pop("label", None)
```
pieleric: `pop()` is actually made for this sort of usage, so you can replace the 2 lines by just:
`label = conf.pop("label", param)`

Same for the line 102.

src/odemis/gui/cont/milling.py
```diff
@@ -101,12 +101,19 @@ def rectangle_pattern_to_shape(canvas,
...
+    """
+    rect = RectangleOverlay(cnvs=canvas, colour=colour, show_selection_points=False)
```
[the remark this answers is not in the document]
pieleric: The docstring should NOT use RST. It should just be text.

src/odemis/gui/cont/milling.py
```diff
@@ -224,9 +231,9 @@ def _update_pattern_panels(self) -> None:
 
         # create the setting panels, and connectors
         self.controls: Dict[str, MillingTaskPanel] = {}
-        pattern_parameters = ["width", "height", "depth", "spacing"] # TODO: add milling params
-        # milling params: current, voltage, field of view, mode
-        milling_parameters = ["current", "align", "mode"]
+        pattern_parameters = ["width", "height", "depth", "spacing"]  # TODO: add milling params
```
pieleric: Let's drop the TODO. It's clearly done.

pieleric (changes requested):

ilyushkin (approved):

src/odemis/gui/cont/milling.py
K4rishma: Actually it should not be zero, I have made it as non-default parameter. As and when odemis is updated, the default value of spot size will be added through the milling tasks configuration.

src/odemis/gui/cont/milling.py
```diff
@@ -100,13 +100,27 @@ def pos_to_absolute(pos: Tuple[float, float], ref_img: model.DataArray) -> Tuple
...
+    :param name: name of the shape, e.g. Rough Milling, Polishing etc.
+    """
+    rect = RectangleOverlay(cnvs=canvas, colour=colour, show_selection_points=False)
     width = pattern.width.value
     height = pattern.height.value
-    x, y = pos_to_absolute(pattern.center.value, ref_img) # image coordinates -> physical coordinates
+    # Expand each dimension by the spot size (diameter) to circumscribe the beam path
+    # Since spot_size is the beam diameter, we expand by spot_size/2 on each side
+    width_expanded = width + spot_size
+    height_expanded = height + spot_size
```
tepals, line 123: I would already divide by 2 here, then the code matches the comment and width_expanded/height_expanded are actually the expanded width and height.
```suggestion
    width_expanded = width + spot_size / 2
    height_expanded = height + spot_size / 2
```

src/odemis/gui/cont/milling.py
```diff
@@ -115,10 +129,10 @@ def rectangle_pattern_to_shape(canvas,
...
     # 4  -  3
 
-    rect.p_point1 = Vec(x - width / 2, y + height / 2)
-    rect.p_point2 = Vec(x + width / 2, y + height / 2)
-    rect.p_point3 = Vec(x + width / 2, y - height / 2)
-    rect.p_point4 = Vec(x - width / 2, y - height / 2)
+    rect.p_point1 = Vec(x - width_expanded / 2, y + height_expanded / 2)
+    rect.p_point2 = Vec(x + width_expanded / 2, y + height_expanded / 2)
+    rect.p_point3 = Vec(x + width_expanded / 2, y - height_expanded / 2)
+    rect.p_point4 = Vec(x - width_expanded / 2, y - height_expanded / 2)
```
tepals, line 136:
```suggestion
    rect.p_point1 = Vec(x - width_expanded, y + height_expanded)
    rect.p_point2 = Vec(x + width_expanded, y + height_expanded)
    rect.p_point3 = Vec(x + width_expanded, y - height_expanded)
    rect.p_point4 = Vec(x - width_expanded, y - height_expanded)
```

src/odemis/gui/cont/milling.py
tepals: Alternatively, you could already do this step on line 117/118, since `width` and `height` are not used anywhere else:

```python
    width_expanded = pattern.width.value + spot_size / 2
    height_expanded = pattern.height.value + spot_size / 2
```

src/odemis/gui/cont/milling.py
```diff
@@ -224,9 +238,9 @@ def _update_pattern_panels(self) -> None:
 
         # create the setting panels, and connectors
         self.controls: Dict[str, MillingTaskPanel] = {}
-        pattern_parameters = ["width", "height", "depth", "spacing"] # TODO: add milling params
-        # milling params: current, voltage, field of view, mode
-        milling_parameters = ["current", "align", "mode"]
+        pattern_parameters = ["width", "height", "depth", "spacing"]
+        # milling params: current, voltage, field of view, mode, spot_size
```
tepals: Why is this comment here and how does it relate to the code? Does "voltage" correspond to "align" and "field of view" to "mode"?

tepals (changes requested):

src/odemis/gui/cont/milling.py
K4rishma: Yes I can do that as well.

src/odemis/gui/cont/milling.py
K4rishma: I just added the spot size to preciously mentioned parameters. I do not know if voltage affects alignment, fov the mode. But according to the comment, it might be the case.

I can see that the comment is confusing nonetheless, I have decided to remove it as it does not bring any insights.

src/odemis/gui/cont/milling.py
K4rishma: Actually I cannot do it, because at this stage the total width of the pattern is being calculated. I have updated the comment to make it more clear

tepals: I don't understand, why is it not possible to write:
```python
    width_expanded = pattern.width.value + spot_size
    height_expanded = pattern.height.value + spot_size
```

src/odemis/gui/cont/milling.py
K4rishma: Similarly I cannot do this, x and y are central origin points of a rectangular pattern. To calculate the vertices, it needs to be shifted by pattern_length /2.

from the calcuations, if I simplify the equation by dividing and removing 2 as per your suggestion, I do not get the desired shifts because
width + spot_size/2 is not equal to (width + spot size) / 2

tepals: Sorry for the confusion, I meant to suggest that you would apply both suggestions

```python
    width_expanded = width + spot_size / 2
    height_expanded = height + spot_size / 2
    rect.p_point1 = Vec(x - width_expanded, y + height_expanded)
    rect.p_point2 = Vec(x + width_expanded, y + height_expanded)
    rect.p_point3 = Vec(x + width_expanded, y - height_expanded)
    rect.p_point4 = Vec(x - width_expanded, y - height_expanded)
```
is the same as

```python
    width_expanded = width + spot_size
    height_expanded = height + spot_size
    rect.p_point1 = Vec(x - width_expanded / 2, y + height_expanded / 2)
    rect.p_point2 = Vec(x + width_expanded / 2, y + height_expanded / 2)
    rect.p_point3 = Vec(x + width_expanded / 2, y - height_expanded / 2)
    rect.p_point4 = Vec(x - width_expanded / 2, y - height_expanded / 2)
```

K4rishma: I understood your comments that it needs to be added together, my response was indeed on above. Maybe I am missing something but

First version: expands width fully + half spot size
Second version: expands half width + half spot size

K4rishma:
Reason for converting to draft - 
During the testing, it was determined that the spot size is current dependent. If the spot size is known per current, it should be written to a config which will then be used during milling.

- spot size should not be shown in the GUI. Most users do not know the spot size. If the spot size is written in the config,  milling will take spot size into account
- Introducing the spot size should not change the displayed width of the milling patterns, the spot size is accounted such that the milling of specified dimensions is performed.
- Implement a way to get the values of spot size per current
- find a way to get the spot size per OEM

pieleric:
@K4rishma I agree with you that it should go back to the specification level. I've replied in the Jira task.

PR #3492, opened 2026-06-16
[MSD-613][feat] Meteor: localization for FIB-view FM posture 

tmoerkerken:
Add FIB-view FM support to the localization tab:
- Add enabling of localization tab when at FIB-view FM posture
- Add read-only indicator for current posture to the localization tab
- Add milling angle label for acquired images at milling angle
- Add disabled state logic for the "Acquire Overview" button

# Screenshots

image

Read-only indicator in the localization tab

image

When image is acquired at FM Milling, we indicate that via the "at <milling angle>" suffix

image

Overview acquisition disabled when at FIB-view FM posture

PR #3498, opened 2026-06-23
[MSD-616][feat] FIB-view FM correlation workflow

tmoerkerken:
> [!WARNING]
> Requires #3492 to be merged, so the diff is showing a lot of unrelated files. The most important change is [src/odemis/gui/cont/multi_point_correlation.py](https://github.com/delmic/odemis/pull/3498/changes#diff-b8d04bf6eaa94c63772703427594101df9cedb9a95ee5ebf0140e2e167e28d65)

The PR implements changes to facilitate a correlation workflow for FIB-view FM data:
- Differentiate milling workflow depending on FM type (Orthogonal FM vs FIB-view FM)
- Add indicator to show the FM type
- Introduce new simple 2d correlation method for FIB-view FM data
- Change grid table to not show z column for FIB-view FM
- Change point requirement depending on FM type (2 vs 4)
- Fix bug in stream deletion, causing to still show a deleted stream in the correlation popup.
- Change display of refinement controls to use a general method
- Disable refinement for FIB-view FM
- Improve refinement search range to use a physical unit (bit of a bonus, since I disabled refinement for FIB-view FM after all)

image
Example of the new 2d correlation method on synthetic data (ignore the content of the images). See how the projected point  at the bottom is located at a plausible location.

### DIFF
--- a/src/odemis/acq/move.py
+++ b/src/odemis/acq/move.py
@@ -379,7 +379,7 @@ class MeteorPostureManager(MicroscopePostureManager):
         self.has_fib_view_fm_objective = False
         try:
             lens = model.getComponent(role='lens')
-            if model.hasVA(lens, "workingDistance") and lens.workingDistance.value > MIN_LENS_WD_FIB_VIEW_FM:
+            if model.hasVA(lens, "workingDistance") and lens.workingDistance.value >= MIN_LENS_WD_FIB_VIEW_FM:
                 self.has_fib_view_fm_objective = True
         except LookupError:
             pass
@@ -1120,7 +1120,6 @@ class MeteorPostureManager(MicroscopePostureManager):
         elif source_posture == FIB_VIEW_FM and target_posture != MILLING:
             return False
         elif source_posture == LOADING:
-            # From loading, we always allow to go to SEM imaging
             if target_posture == SEM_IMAGING:
                 return True
             # When we are at the loading position, we internally work with the SEM imaging source posture
@@ -1128,7 +1127,7 @@ class MeteorPostureManager(MicroscopePostureManager):
             else:
                 source_posture = SEM_IMAGING
 
-        # Otherwise, check for availability in posture transforms
+        # Check for availability in posture transforms
         transform_available = bool(
             self._posture_transforms.get(source_posture, {}).get(target_posture, False)
         )
--- a/src/odemis/gui/cont/multi_point_correlation.py
+++ b/src/odemis/gui/cont/multi_point_correlation.py
@@ -23,6 +23,7 @@ Odemis. If not, see http://www.gnu.org/licenses/.
 from __future__ import annotations  # allows the use of Python3.9 style typing in Python3.8
 
 import logging
+import math
 import queue
 import re
 import threading
@@ -32,21 +33,20 @@ from typing import List, Union, Optional, Tuple, Dict, Set
 
 import numpy
 import wx
-# IMPORTANT: wx.html needs to be imported for the HTMLWindow defined in the XRC
-# file to be correctly identified. See: http://trac.wxwidgets.org/ticket/3626
-# This is not related to any particular wxPython version and is most likely permanent.
 
 from odemis import model, util
 from odemis.acq.align.tdct import _convert_das_to_numpy_stack, run_tdct_correlation
 from odemis.acq.feature import FIBFMCorrelationData, Target, TargetType
+from odemis.acq.move import POSITION_NAMES, FM_IMAGING, FIB_VIEW_FM
 from odemis.acq.stream import StaticFluoStream, StaticSEMStream, StaticFIBStream, FluoStream
-from odemis.gui import conf
-from odemis.gui.model import CryoGUIData
 from odemis.gui.cont.features import save_project
+from odemis.gui.img import getBitmap
+from odemis.gui.model import CryoGUIData
 from odemis.gui.util import call_in_wx_main
 from odemis.model import ListVA
 from odemis.util.dataio import data_to_static_streams
 from odemis.util.img import get_brightest_channel, compute_center_of_mass
+from odemis.util.transform import SimilarityTransform
 from odemis.util.units import readable_str
 
 # create an enum with column labels and position
@@ -68,8 +68,11 @@ RIM_COR_DEFAULT = 0.495  # See MD_RIM_COR. This value works fine for 50x objecti
 # conditions to convert between physical and pixel coordinate systems in order for multipoint correlation to operate.
 # For coordinate conversions, we assume the pixels in 3D are isosymmetric
 # i.e. size in pixel[0]=pixel[1]=pixel[2].
-COM_ROI_PADDING = 12  # Padding (pixels) for center of mass ROI extraction
-# TODO: Could adapt padding based on pixel spacing for more flexibility if needed
+REFINE_SEARCH_RANGE = 2.5e-6  # m, search range for fiducial refinement
+
+# If the milling angle of a FIB image and an FM z-stack are below the tolerance, they are considered to be matching.
+MILLING_ANGLE_TOLERANCE = math.radians(0.1)  # Now 0.1 degree, but this depends on stage accuracy, so might need to update this later.
+
 
 def get_pixel_3d_coordinates(stream: FluoStream, p_pos: Tuple[float, float, float], check_bbox: bool = False) \
         -> Optional[Tuple[float, float, float]]:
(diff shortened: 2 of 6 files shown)


K4rishma:
I have a general question. Why do we do 2d correlation from the multi point correlation dialog box instead of using correlation tab which is designed to do 2D? Maybe @pieleric knows it.

tmoerkerken:
> I have a general question. Why do we do 2d correlation from the multi point correlation dialog box instead of using correlation tab which is designed to do 2D? Maybe @pieleric knows it.

It's a decision I made, based on a discussion between Eric and Marit. Either would have been possible, but since I think point matching is more accurate, reusing the existing point matching widget was the quickest implementation. By the way, I do think we need to critically look at the two, once we do the redesign. They probably should be one tool.

Eric:
How is the overlay FM_MILLING <> FIB displayed? Should we use the 3D correlation window? That sounds a little odd, as it should only be 2D (that said, 2D is a subset of 3D! So it could work?). Should we have a separate window “Milling angle correlation”, which would be a simplified version of the 3D correlation window? Or should the 3D correlation window support switching the FM streams between FM and FM_MILLING (for the rare cases where a feature has both FM and FM_MILLING streams). Are the tools from the 3D correlation window good for such 2D alignment (ie, series of points to match in both images), or should we have tools like in the correlation tab (ie, manually shift, scale and rotate).

Marit:
Intuitively, the “2D correlation” option would make more sense. Typically, the rotation shift is 0, the translation default value is “not too bad”, and the scale almost correct (and might be improved by doing a calibration of the optical lens mag once). In addition, the 3D correlation uses gaussian fitting + Z option… which is actually counter helping (as the features that will be matched might not be spots, but milled arbitrary shapes). However, the ideal case would be to have a 2D view with an option to use “match 3+ points” as an extra help. On the other hand, in practical terms, it might be easier to re-use the 3D correlation window, and have a way to “dumb it down” by removing some features (gaussian fitting and Z position) when a FM_MILLING data is loaded.

src/odemis/gui/cont/tabs/localization_tab.py
```diff
@@ -440,9 +446,13 @@ def display_acquired_data(self, data):
...
         view = self.tab_data_model.views.value[1]
         self.tab_data_model.focussedView.value = view
+        pm = self.main_data.posture_manager
         for s in data_to_static_streams(data):
+            pos = s.getRawMetadata()[0].get(model.MD_STAGE_POSITION_RAW)
+            # Add milling angle suffix to stream's name if it was acquired at the milling angle.
+            if pm.at_fib_view_fm_posture(pos):
+                s.name.value = f"{s.name.value} at {math.degrees(pm.milling_angle.value):0.0f}°"
             if self.tab_data_model.main.currentFeature.value:
-                s.name.value = self.tab_data_model.main.currentFeature.value.name.value + " - " + s.name.value
```
K4rishma: Is this the intention, to remove the feature name from the stream name?

src/odemis/gui/cont/tabs/localization_tab.py
```diff
@@ -475,7 +485,10 @@ def _on_acquisition(self, is_acquiring):
 
     def _on_stage_pos(self, pos):
         """
-        Called when the stage is moved, enable the tab if position is imaging mode, disable otherwise
+        Called when the stage is moved, and perform the following actions:
+        - enable the tab if position is imaging mode, disable otherwise
```
K4rishma: We can maybe explicit which imaging mode we are talking about, is it overview acquisition?

src/odemis/gui/cont/multi_point_correlation.py
```diff
@@ -32,6 +33,14 @@
...
 import numpy
 import wx
+from odemis.util.transform import SimilarityTransform
+
+from odemis.acq.align.transform import CalculateTransform
+
+from odemis.gui.img import getBitmap
+
+from odemis.acq.move import POSITION_NAMES, FM_IMAGING, FIB_VIEW_FM
+
```
K4rishma: The blank lines should not be there between the imports

src/odemis/gui/cont/multi_point_correlation.py
```diff
@@ -68,9 +77,13 @@ class GridColumns(Enum):
 # conditions to convert between physical and pixel coordinate systems in order for multipoint correlation to operate.
 # For coordinate conversions, we assume the pixels in 3D are isosymmetric
 # i.e. size in pixel[0]=pixel[1]=pixel[2].
-COM_ROI_PADDING = 12  # Padding (pixels) for center of mass ROI extraction
+REFINE_SEARCH_RANGE = 2.5e-6  # m, search range for fiducial refinement
 # TODO: Could adapt padding based on pixel spacing for more flexibility if needed
```
K4rishma: Is this comment still valid?

src/odemis/gui/cont/tabs/localization_tab.py
tmoerkerken: Yes, that was the plan. We only show streams corresponding to the selected feature. So it's kind of redundant. Plus, we had to make a bit of room for the "at 10°" suffix.

src/odemis/gui/cont/tabs/localization_tab.py
tmoerkerken: Yeah, this docstring is not great. In the other PR I left a comment to improve it. Should be FM Imaging and FIB-view FM

src/odemis/gui/cont/multi_point_correlation.py
tmoerkerken: No, that was exactly what was implemented now, so should be gone :sweat_smile: .

src/odemis/gui/cont/tabs/localization_tab.py
K4rishma: I understand why it is removed, because as we display streams for the current feature, it does not make sense to add feature name. Although this is true only for the Localization Tab. The feature name is helpful to be added in the stream name when the the users view diffferent features together in the Gallery/Analysis/Odemis Viewer.

src/odemis/gui/cont/multi_point_correlation.py
```diff
@@ -209,6 +257,10 @@ def __init__(self, frame):
         self.grid.Bind(wx.EVT_KEY_DOWN, self._on_key_down_grid)
         self.grid.EnableEditing(True)
 
+        # Hide the z-column for the FIB-view FM workflow, since we only perform 2d correlation.
+        if self.at_fib_view_fm.value:
+            self.hide_grid_column(3)
```
K4rishma: to avoid using hard coded numbers
```suggestion
            self.hide_grid_column(GridColumns.Z.value)
```

src/odemis/gui/cont/multi_point_correlation.py
```diff
@@ -168,6 +181,41 @@ def __init__(self, frame):
         self._main_data_model = self._tab_data_model.main
         self._panel = frame
         self._viewports = frame.pnl_correlation_grid.viewports
+
+        streams = self._main_data_model.currentFeature.value.streams.value
```
K4rishma, line 175: There is a possibility that the current features has both the streams, aka, FM streams and FM at milling angle. If there are both streams then the lines from 210 to 218, will just focus on FM at milling angle, and not consider the FM view at all. Should this not be decided by the user, if they want to use milling angle FM or normal FM ? 

If the user had created z stack at FM view, they might want to use the information

src/odemis/gui/cont/stream_bar.py
```diff
@@ -2369,9 +2370,14 @@ def _display_feature_streams(self, feature: Optional[CryoFeature]) -> None:
...
             return
 
+        pm = self._tab_data_model.main.posture_manager
         # Add streams to the tab model
         for s in feature.streams.value:
             if s not in self._tab_data_model.streams.value:
+                pos = s.getRawMetadata()[0].get(model.MD_STAGE_POSITION_RAW)
+                # Add milling angle suffix to stream's name if it was acquired at the milling angle.
+                if pm.at_fib_view_fm_posture(pos):
+                    s.name.value = f"{s.name.value} at {math.degrees(pm.milling_angle.value):0.0f}°"
```
K4rishma: Why do we add this twice, once in the locatlzation tab and also here?

tmoerkerken: When freshly acquiring the data, we use a different display method, since the data is already in memory (odemis.gui.cont.tabs.localization_tab.LocalizationTab.display_acquired_data) then when (lazy) loading the data from disk (odemis.gui.cont.stream_bar.CryoAcquiredStreamsController._display_feature_streams). 

It does look a bit strange that we have a CryoAcquiredStreamsController but separately also handling it in the LocalizationTab. That's part of the reason I'm going to look into this for a while in upcoming sprints, since it all feels a bit incoherent.

src/odemis/gui/cont/multi_point_correlation.py
tmoerkerken: Indeed, we decided that FIB-view FM takes priority. But I do agree it can be confusing to the user. We made the decision also based on the knowledge that these are two different workflows, and in practice, having the two flavours of FM would be rare. We could however make use of the visibility, to have some form of control. I will have a look.

src/odemis/gui/cont/multi_point_correlation.py
tmoerkerken: Nice, much better

src/odemis/gui/cont/tabs/localization_tab.py
tmoerkerken: Yes, I can see how that is helpful. The current master code also does not show the feature name prefix in the Analysis/Odemis Viewer tabs, so maybe add it to the idea board?

src/odemis/gui/cont/multi_point_correlation.py
tmoerkerken: Alright, I made some changes (there was a bug in stream deletion). Now you can delete the undesired stream from the localization tab, and it will not use it for correlation. If I'm correct, this would be an edge-case though.

K4rishma (approved):

src/odemis/acq/move.py
```diff
@@ -368,6 +375,15 @@ def __init__(self, microscope):
             milling_angle, (MILLING_RANGE[0], MILLING_RANGE[1]), unit="rad", setter=self._set_milling_angle
         )
 
+        # Check if objective is FIB-view FM compatible
+        self.has_fib_view_fm_objective = False
+        try:
+            lens = model.getComponent(role='lens')
+            if model.hasVA(lens, "workingDistance") and lens.workingDistance.value > MIN_LENS_WD_FIB_VIEW_FM:
+                self.has_fib_view_fm_objective = True
```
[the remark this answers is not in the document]
pieleric: @tmoerkerken this is fair enough -> make it  `>=`.

src/odemis/gui/cont/multi_point_correlation.py
```diff
@@ -32,23 +33,27 @@
...
 from odemis.util.dataio import data_to_static_streams
 from odemis.util.img import get_brightest_channel, compute_center_of_mass
+from odemis.util.transform import SimilarityTransform
 from odemis.util.units import readable_str
 
+
+# IMPORTANT: wx.html needs to be imported for the HTMLWindow defined in the XRC
+# file to be correctly identified. See: http://trac.wxwidgets.org/ticket/3626
+# This is not related to any particular wxPython version and is most likely permanent.
+
```
pieleric: As mentioned by copilot, this comment doesn't make sense here. So please drop it (instead of moving it).

src/odemis/gui/cont/multi_point_correlation.py
```diff
@@ -460,8 +552,58 @@ def stop(self):
...
+            index_order.append(fib_coord.index.value)
+            fib_coord_px = self.correlation_target.fib_stream.getPixelCoordinates(fib_coord.coordinates.value[0:2], check_bbox=False)
+            fib_coords.append(fib_coord_px)
+
+        for fm_coord in self.correlation_target.fm_fiducials:
+            fm_coord_px = get_pixel_3d_coordinates(self.correlation_target.fm_streams[0], fm_coord.coordinates.value)[:2]
+            fm_coords.append(fm_coord_px)
+
+        fib_coords = numpy.array(fib_coords, dtype=numpy.float32)
+        fm_coords = numpy.array(fm_coords, dtype=numpy.float32)
```
pieleric: Is there any specific reason to use float32 (over the default float64)? I suspect it's just a copy-paste from the 3D correlation (which is a little faster in 32 bits). If there is no specific reason you can just write `dtype=float`.

src/odemis/gui/cont/multi_point_correlation.py
```diff
@@ -460,8 +552,58 @@ def stop(self):
...
+                            type=TargetType.ProjectedFiducial, index=index_order[i],
+                            fm_focus_position=0)
+            self._tab_data_model.projected_points.append(target)
+            self.correlation_target.fib_projected_fiducials.append(target)
+
+        rms = numpy.mean(((fib_coords - fm_coords_transformed) ** 2) ** 0.5)
+        # Fill in error metric in similar way to 3d correlation, for consistency
+        self.correlation_target.correlation_result = {
+            "output": {"error": {"rms_error": rms}},
+        }
```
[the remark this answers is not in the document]
pieleric: @tmoerkerken this looks like a good suggestion.

pieleric (changes requested):

pieleric (approved):

